Issue #9588, opened 2021-01-19
agent/cache: stop refreshing a cache after some period of constant errors


dnephin:
Related to #8541 (and other issues I can't find right now).

We've had a few reports of the logs being filled with errors caused by the background refresh of agent/cache entries. The agent/cache has a TTL of 72 hours. So when something is removed it can produce a lot of logging noise for days. This problem can also cause the agent to OOM. If many requests for non-existent data are made, then we start a goroutine for each. Eventually all those goroutine can OOM the agent.

Instead of attempting to refreshing the entry for 72 hours , and getting an error every time, we should stop refreshing after either some number of attempts, or after some period of time without a success.

mkeeler:
So made an attempt to cause ACL not found errors received to force a cache eviction and stopping of the background refresh. The thought was that if another request came later (such as in the case of proxycfg watches still being ongoing) that they would reissue the query and thus not impact the correctness of the data returned.

All was seeming to go well and I had made it so that the cache would be evicted and the query stopped but I ran into one problem. That problem is with how the results of a fetch get propagated back to all callers waiting for fetch completion. The fetch always happens in a background go routine and is single flighted. Callers then wait for a chan to be closed and then refetch from the cache to pull the results.

The problem here is that if we evict the cache upon certain errors then those waiting for the fetch have no way of seeing the fetches error. If our single flighting utilized something such as the [`x/sync/singleflight` package](https://pkg.go.dev/golang.org/x/sync/singleflight) then the actual result would also be sent directly back to the callers eliminating the need for refetching from the cache and allowing the forced eviction approach to work. As this would be a much larger change to the cache I have since stopped my efforts for now.

PR #9738, opened 2021-02-08
Stop background refresh of cached data for requests that result in ACL not found errors

mkeeler:
This is a partial fix to #9588 but is limited in scope. When the cache encounters an ACL not found error it will stop any background refresh which may be happening for that request. The assumption here is that the token was deleted and is not coming back.

There is however one case where a token may have yet to be replicated in a secondary DC that will prevent the background refresh of the request from happening when it may succeed. This might make the next request take a little longer but is preferable to spamming the logs for days with ACL not found errors.

Also of note is that cache watches for the request will cause the request to continually be made anyways and keep it in an always fetching state (or in exponential retry backoff sleeps).

### DIFF
--- a/.changelog/9738.txt
+++ b/.changelog/9738.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cache: Prevent spamming the logs for days when a cached request encounters an "ACL not found" error.
+```
--- a/agent/cache/cache.go
+++ b/agent/cache/cache.go
@@ -27,6 +27,7 @@ import (
 	"github.com/armon/go-metrics/prometheus"
 	"golang.org/x/time/rate"
 
+	"github.com/hashicorp/consul/acl"
 	"github.com/hashicorp/consul/lib"
 	"github.com/hashicorp/consul/lib/ttlcache"
 )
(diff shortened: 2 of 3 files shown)


banks (approved):
I think this makes sense. In the case where there is something blocking (or using Notify internally) you mention that we'll keep restarting the watch anyway. Does that produce the same level of log spam until the thing stops watching with an old ACL?

I think it's still better though because once the thing with the old ACL goes away, the logs will stop instead of continuing for another 3 days right?

mkeeler:
In the case where the watch is still around it will be 1 log per minute (1m is the max time between errors that the watch will invoke `getWithIndex`).

Before when the `fetch` method was invoked for the background refresh it would do so with an attempts parameter and the exponential backoff (utilizing the same `backOffWait` function) was handled in the internal go routine right before reissuing the next fetch in a new go routine. So basically both cases will ensure that in the event of failures we do the exact same exponential backoff so the log spam would be identical if the watch was ongoing as it would be with no watch before this fix.

The assumption here is that the token was a short lived one for a workload. In k8s maybe that workload is already gone and the token that was generated for it has been deleted or expired. In that case, the service definition should have been deregistered and all injected Envoy proxies would be gone. Therefore no watches utilizing that token should be ongoing as they would all be cancelled when the service was deregistered.

The problem previously was that even after all watches had ended we were continuing to refresh these entries for 72h and if you churn through workloads quickly enough you would accumulate many erroring background refreshes causing potentially millions of log entries for these ACL errors.

PR #9741, opened 2021-02-09
Add fields to the /acl/auth-methods endpoint.

markan:
Add fields to the /acl/auth-methods endpoint.

A GET of the /acl/auth-methods/:name endpoint returns the fieldsMaxTokenTTL and TokenLocality, while a LIST (/acl/auth-methods) does
not.

The list command returns a filtered subset of the full set. This is
somewhat deliberate, so that secrets aren't shown, but the TTL and
Locality fields aren't (IMO) security critical, and it is useful for
the front end to be able to show them.

For consistency these changes mirror the 'omit empty' and string
representation choices made for the GET call.

This includes changes to the gRPC and API code in the client.

The new output looks similar to this
curl 'http://localhost:8500/v1/acl/auth-methods' | jq '.'

  {
    "MaxTokenTTL": "8m20s",
    "Name": "minikube-ttl-local2",
    "Type": "kubernetes",
    "Description": "minikube auth method",
    "TokenLocality": "local",
    "CreateIndex": 530,
    "ModifyIndex": 530,
    "Namespace": "default"
  }
]

### DIFF
--- a/api/acl_test.go
+++ b/api/acl_test.go
@@ -3,6 +3,7 @@ package api
 import (
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/hashicorp/consul/sdk/testutil/retry"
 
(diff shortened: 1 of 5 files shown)


mkeeler (approved):
LGTM, I assume this was just an oversight when adding the locality and max ttl to the authmethod and agree that it would be useful in the list endpoint.

api/acl_test.go
```diff
@@ -657,6 +658,101 @@ func TestAPI_ACLToken_Clone(t *testing.T) {
...
+	defer s.Stop()
+
+	acl := c.ACL()
+	s.WaitForSerfCheck(t)
+
+	method1 := ACLAuthMethod{
+		Name:          "test_1",
+		Type:          "kubernetes",
+		Description:   "test 1",
+		MaxTokenTTL:   260 * 1000 * 1000 * 1000, // can't use underscores until go-13
```
rboyer, line 674: Why not something like `4 * time.Minute`?

rboyer:
I feel like the PR title should be about `/acl/auth-methods` endpoint

PR #9746, opened 2021-02-10
ui: DataSource Decorator

johncowen:
We use a `<DataSource @src={{url}} />` component throughout our UI for when we want to load data from within our components. The URL specified as the `@src` is used to map/lookup what is used in to retrieve data, for example we mostly use our repository methods wrapped with our Promise backed `EventSource` implementation, but DataSource URLs can also be mapped to EventTarget backed `EventSource`s and native `EventSource`s or `WebSockets` if we ever need to use those (for example these are options for potential streaming support with the Consul backend).

The URL to function/method mapping previous to this PR used a very naive humongous `switch` statement which was a temporary 'this is fine for the moment' solution, although we'd always wanted to replace with something more manageable.

Here we add `wayfarer` as a dependency - a very small (1kb), very fast, radix trie based router, and use that to perform the URL to function/method mapping.

This essentially turns every `DataSource` into a very small SPA - change its URL and the view of data changes. When the data itself changes, either the yielded view of data changes or the `onchange` event is fired with the changed data, making the externally sourced view of data completely reactive.

```javascript
// use the new decorator a service somewhere to annotate/decorate
// a method with the URL that can be used to access this method
@dataSource('/:ns/:dc/services')
async findAllByDatacenter(params) {
  // get the data
}

// can use with JS in a route somewhere
async model() {
  return this.data.source(uri => uri`/${nspace}/${dc}/services`)
}
```

```hbs
{{!-- or just straight in a template using the component --}}
<DataSource @src="/default/dc1/services" @onchange="" />
```

We did some investigation as to whether we could use Ember's router/route-recognizer for this but we hit a couple of problems:

1. We need to allow empty URL segments/params, for example for an empty namespace (`//dc-1/services`), it doesn't seem like Ember's Router supports this.
2. There doesn't seem to be a public way to get to Ember's router microlib (route-recognizer), we tried getting to it via private methods and then accessing the constructor of the instantiated instance of Ember's Router, which was all a bit ugh. We also briefly looked at separately depending on `route-recognizer` but this adds 4kb to our dependencies vs 1kb for `wayfarer` which also supports empty segments (see point 1)

Lastly, this probably isn't the final version of this:

We found that as Ember Services aren't statically imported using `import` but instead sourced by ember-cli and only executed on demand, the decorators used in the service aren't executed until you first access the service. So, we have a little bit of a chicken and egg problem - the decorator needs to be executed so we know which Service needs to be accessed, but the decorator isn't executed until you access the Service!

So, for the moment we still require a large list of services injecting into our datasource service in order to execute the decorators. I'm not entirely sure the best way to work around this as yet, but potentially we may still move away from the decorators and end up with a centralized place to configure these, similar to Ember's `router.js` file we use to configure our Ember Routes.

Whichever way we go the overall approach using a router to map URLs to methods will remain the same.

### DIFF
--- a/ui/packages/consul-ui/app/decorators/data-source.js
+++ b/ui/packages/consul-ui/app/decorators/data-source.js
@@ -1,0 +1,20 @@
+import wayfarer from 'wayfarer';
+import { singularize } from 'ember-inflector';
+
+const router = wayfarer();
+export default path => (target, propertyKey, desc) => {
+  router.on(path, function(params, owner) {
+    let name;
+    if (params.modelName) {
+      name = singularize(params.modelName);
+    } else {
+      name = target.getModelName.call();
+    }
+    const instance = owner.lookup(`service:repository/${name}`);
+    return configuration => desc.value.apply(instance, [params, configuration]);
+  });
+  return desc;
+};
+export const match = path => {
+  return router.match(path);
+};
--- a/ui/packages/consul-ui/app/mixins/token/with-actions.js
+++ b/ui/packages/consul-ui/app/mixins/token/with-actions.js
@@ -8,11 +8,11 @@ export default Mixin.create(WithBlockingActions, {
   actions: {
     use: function(item) {
       return this.repo
-        .findBySlug(
-          get(item, 'AccessorID'),
-          this.modelFor('dc').dc.Name,
-          this.modelFor('nspace').nspace.substr(1)
-        )
+        .findBySlug({
+          ns: this.modelFor('nspace').nspace.substr(1),
+          dc: this.modelFor('dc').dc.Name,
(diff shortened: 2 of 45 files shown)


kaxcode (approved):
LGTM

johncowen:
Just a quick note here, there are a few PRs in progress that will potentially need to be updated to following this, so I'm going to wait until they are merged first and rebase this to deal with any conflicts on this PR before finally merging this down.

johncowen:
I rebased this and made an additional PR here https://github.com/hashicorp/consul/pull/9801 to bring all of this into line with the work that was happening in tandem with this. Once that PR is merged onto here we can move this down onto `master`.

PR #9749, opened 2021-02-10
ui: [BUGFIX] Ensure we show the correct count of instances for each node

johncowen:
In our node listing page we show the a number for the amount of Service Instances on each node.

Screenshot 2021-02-10 at 15 22 28

Before this PR, this count included any proxies on the Node, essentially doubling the service count. This was probably correct in a previous version of the UI when we used to show all the proxies across the catalog as well as the services, but since we merged proxies into the services themselves, essentially making the an implementation detail and therefore invisible in the UI, this count should not include proxies.

This PR adds a `MeshServiceInstances` property to the node model, 'mesh' here meaning 'services that have a proxy assigned to them, and therefore not a proxy itself'. We then use this property to show the instance count, meaning we omit proxy instances from the final count.

We also noticed that we weren't showing an 'empty' icon for services where the check count for either service or node checks was zero. So fixed that here also 🐦 🐦 🥧 

This is how consul looks in the node listing after this fix (the consul service only has node checks, no service checks)

Screenshot 2021-02-10 at 15 25 26

### DIFF
--- a/.changelog/9749.txt
+++ b/.changelog/9749.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Exclude proxies when showing the total number of instances on a node.
+```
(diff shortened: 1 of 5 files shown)


kaxcode (approved):
:shipit:

PR #9752, opened 2021-02-10
ui: Remove any trailing fullstop/period DNS characters from Gateways UI API

johncowen:
Previous to this commit, the API response for getting information about the upstreams for a gateway (`/v1/internal/ui/gateway-services-nodes/:gateway`) would include Gateway Addresses in the form `domain.name.:8080`, which due to the addition of the port is probably not the expected format (the trailing `.` is DNS not Address:Port URLs).

This commit rightTrims any `.` characters from the end of the domain before formatting the address to include the port resulting in `domain.name:8080`

One thing I'm unsure of, whilst this 'fixes' things for displaying this in the UI, does this effect anything else backend related? I had a quick search to see is the `.Addresses()` method was being used elsewhere, and I couldn't see anything, but would be good to double check.

### DIFF
--- a/.changelog/9752.txt
+++ b/.changelog/9752.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+api: Remove trailing periods from the gateway internal HTTP API endpoint
+```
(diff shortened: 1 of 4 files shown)


crhino (approved):
LGTM, thanks for doing this!

crhino:
We should probably add a changelog entry for this as well.

johncowen:
Thanks @crhino! I've added a changelog so I'll merge this a little later on 👍

PR #9758, opened 2021-02-11
http: error if near is used with streaming

dnephin:
Follow up to https://github.com/hashicorp/consul/pull/9745#pullrequestreview-587002996

Error instead of silently falling back to blocking queries, also document the limitation.


banks (approved):
Thanks Daniel.

We'll have to review how big of a limitation this is for existing users. I wonder if it might be enough to prevent us from making streaming the default/only transport because people rely on this feature?

For now I think this is good an we can evaluate how many people would be impacted when considering GA.

PR #9759, opened 2021-02-11
streaming: default rpc.enable_streaming to true

dnephin:
So that all servers will start the grpc server used by streaming.

Not backporting this, it should only be included in the next release.


banks (approved):
LGTM!

PR #9761, opened 2021-02-12
ui: Use ember-on-resize-modifier to fix bug with sidebar

kaxcode:
🐛 The icons in the Topology View were not positioning themselves in the right place with the sidebar close/open action.

🔧 Use ember-on-resize-modifier to recalculate the positioning of the icons.

### DIFF
--- a/ui/packages/consul-ui/app/components/topology-metrics/index.js
+++ b/ui/packages/consul-ui/app/components/topology-metrics/index.js
@@ -68,21 +68,10 @@ export default class TopologyMetrics extends Component {
 
   // =actions
   @action
-  setHeight() {
-    if (document.querySelector('#downstream-container')) {
-      const downContainer = document.querySelector('#downstream-container').getBoundingClientRect();
-
-      document
-        .querySelector('#downstream-lines')
-        .setAttribute('style', `height:${downContainer.height}px`);
-    }
-
-    if (document.querySelector('#upstream-container')) {
-      const upContainer = document.querySelector('#upstream-container').getBoundingClientRect();
-
-      document
-        .querySelector('#upstream-lines')
-        .setAttribute('style', `height:${upContainer.height}px`);
+  setHeight(el, item) {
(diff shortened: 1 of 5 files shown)


johncowen (commented):
Thanks for looking at this!

I tried the preview site (🎉 ) and saw this scroll bar weird thing happening:

topo-scroll

Looks like there is some sort of recursively resizing thing happening somewhere?

kaxcode:
@johncowen I fixed that scroll issue with some css. Easiest fix 😂

ui/packages/consul-ui/app/components/topology-metrics/index.js
```diff
@@ -67,6 +67,27 @@ export default class TopologyMetrics extends Component {
...
+      const upContainer = document.querySelector('#upstream-container').getBoundingClientRect();
+
+      document
+        .querySelector('#upstream-lines')
+        .setAttribute('style', `height:${upContainer.height}px`);
+    }
+
+    this.calculate();
+  }
+
```
johncowen, line 90: If possible it would be good to use ember to get to these DOM elements instead of using `querySelector`. I'm thinking if you move the modifiers to the `#-container` elements themselves, that will pass the elements through as arguments meaning you can avoid using `querySelector` in order to find those.

Then I think if you use a `{{did-insert (set this 'upstream/downstreamLines')}}` on the lines DOM elements that should give you a reference on the component for each of the lines DOM elements, meaning you can avoid `querySelector` for those also.

Not a necessity but would make this a little nicer by depending less on ids. I think if thats ^ not possible for whatever reason we should at least cache the results of `querySelector` so we aren't doing it multiple times.

johncowen (commented):
Looks good! I left a comment on the approach here, let me know what you think of that before we merge this down.

johncowen (approved):
Thanks for looking at the DOM vs Ember stuff! Looks good to me!

PR #9763, opened 2021-02-12
cache: log a warning when Cache.Notify handles an error

dnephin:
Without these warnings, errors are silently ignored, which can make debugging problems more challenging.

### DIFF
--- a/agent/config/testdata/TestRuntimeConfig_Sanitize.golden
+++ b/agent/config/testdata/TestRuntimeConfig_Sanitize.golden
@@ -74,7 +74,8 @@
     "CAPath": "",
     "Cache": {
         "EntryFetchMaxBurst": 42,
-        "EntryFetchRate": 0.334
+        "EntryFetchRate": 0.334,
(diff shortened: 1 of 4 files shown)


i0rek (approved):
LGTM

agent/config/testdata/TestRuntimeConfig_Sanitize.golden
```diff
@@ -74,7 +74,8 @@
     "CAPath": "",
     "Cache": {
         "EntryFetchMaxBurst": 42,
-        "EntryFetchRate": 0.334
+        "EntryFetchRate": 0.334,
+        "Logger": null
```
dnephin, line 78: TODO: see if this json marshal still works with a non-nill Logger.

dnephin: I tried it out, and it serialized just fine. Prints `map[string]interface{}`, since the struct has no exported fields.

PR #9706, opened 2021-02-04
ui: Implement read-only KV/Session and Intention views based on ACLs

johncowen:
This PR is a continuation of https://github.com/hashicorp/consul/pull/9687.

Here we implement read-only views for KVs, Sessions and Intentions based on your ACL Tokens policies.

This means you can assign ACLs to users in the UI giving certain users read only access to areas like KV. This is currently at a global level using something like:

```hcl
key_prefix "" {
  policy = "read"
}
```

which would give the user with a token with the above policy read only access to all KVs.

Being able to configure this more finely per KV will be added in a further PR.

Further notes:

1. `{{disabled}}` modifier f0cc8cb938ed0aa0d24750c06dd1446b6d7314fa. `<fieldset>`s have a nice feature that lets you set a `<fieldset disabled>` on them which means all of the form elements within the fieldset are in a disabled state, meaning you only have to set the disabled-ness in one place. The downside is, if you want everything disabled apart from one form element, there doesn't seem to be a native way to do that. Using a modifier instead means we can add the functionality we want here. This is currently not super reactive (if you add further elements to the parent element after insertion these will not be set to the correct value for disabled). In order to improve this we'll probably require some `MutationObserver` work. As and when we need that we can look at that then.
2. Global configuration of protected routes 66c8cc94d453b728b471969d0bf5e3caa8c8d4c2 and e66c220302d67385116cba1870a1ade9b81f8406. We currently have a JSON-like object which is then used by `Router` in order to create embers routes. This means we can also use this information for other things, for example knowing whether a route is a wildcard route in order to add further url de/encoding logic to the params for this route. These commits further add to this configuration with an `abilities: []` property. The `abilities` property allows you to use `ember-can`s DSL to specify what abilities a user requires in order to access a route. If the user doesn't have all of the specified abilities then they see a 403 page instead. Using the central configuration means we can add the logic required for this to our Consul specific base `Route`, then its super simple to see all in one place, what routes have additional ability based restrictions.
3. Acceptance tests where added, along with a new step to allow us to set a users permissions during testing. Worthwhile noting that in order to set permissions during development the cookie names to use in WebInspector takes the form `CONSUL_RESOURCE_{singular-resource-name}_{access_type}`, so for example setting the cookie `CONSUL_RESOURCE_INTENTION_WRITE=false` will give you readonly intentions.

PR #9687, opened 2021-02-02
ui: Restrict the viewing/editing of certain UI elements based on the users ACLs

johncowen:
This PR builds on the minimal permission inspection we were using in order to decide whether to show the 'Manage Namespace' menu item for operators.

As we are likely to be doing this kind of thing a lot more often, we've used `ember-can` and followed the approach used by other products using this addon/functionality.

The user facing impact of this is that we now only show certain main menu items depending on whether you have read access to them or not. For example, if your ACL token doesn't have read access to KV, then we don't show you the KV navigation link in the main navigation.

We've added a very simple acceptance test only here as there are more upcoming tasks related to this which will probably mean us backing some of the work here with `ember-data` in order to follow the patterns used by the rest of the app, and I felt it would be a good idea to implement another feature related to this first before deciding on whether/how we should do this. Once this is done we'll probably add a little more testing around this also.

PR #9769, opened 2021-02-15
ui: Implement fine grained read-only views based on ACLs

johncowen:
This is a continuation of https://github.com/hashicorp/consul/pull/9687 and https://github.com/hashicorp/consul/pull/9706.

This PR performs a pre-request when viewing individual models that may be restricted via `service_prefix`/`key_prefix` etc. If the request tells us that the user doesn't have access to view the individual model, then we don't even make the request to try to fetch it (which would result in a 403 anyway).

If your access is restricted whilst you are viewing a model and we are told that during our blocking queries, then we show a message (similar to when a service is deregistered) but instead of keeping the information in the page, we wipe the page and show you our standard 403 error page (including a link to re-authenticate)

As only certain models require this pre-request check, we've added a `segmented` property to all of our `abilities`, if this property is `false` no check is ever made as the model cannot be 'segmented'/used with `_prefix` (such as ACLs).

### DIFF
--- a/ui/packages/consul-ui/app/abilities/acl.js
+++ b/ui/packages/consul-ui/app/abilities/acl.js
@@ -8,6 +8,7 @@ export default class ACLAbility extends BaseAbility {
   @service('env') env;
 
   resource = 'acl';
+  segmented = false;
 
   get canRead() {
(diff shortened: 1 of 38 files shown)


kaxcode (approved):
LGTM

PR #9771, opened 2021-02-15
ui: Adds a service for Class based service lookup for dataSource decorator

johncowen:
In https://github.com/hashicorp/consul/pull/9746 we added a new `@dataSource` decorator in order to annotate our `EventSource` functions with a URL for access using our `<DataSource />` component, the data then automatically updates live from the backend using our blocking query support:

```javascript
@dataSource('/:ns/:dc/services')
async findAllByDatacenter({ns, dc}) {
   return fetch(`/somewhere?ns=${ns}&dc=${dc}`)
}
```

```hbs
<DataSource @src="/default/dc1/services" @onchange="" />
```

A few potential improvements we thought we could make with that PR revolved around:

1. Ensuring decorators of 'ember resolved' (and therefore auto imported) classes are executed at app boot up.
2. Being able to lookup instances of a class using a DI container. Ember's default container supports string based identifier lookup. This means that we have to either generate an instance manually (losing singleton guarantees?), or figuring our the string identifier based on 'something'. In https://github.com/hashicorp/consul/pull/9746 we used our `getModelName` which despite working only 'worked by accident' and is really used for providing a centralized getter for model names not service names - accidentally both model and service names are generally the same.
3. Providing some debugging functions to aid development experience with an easy way to get a list of DatSource URLs and the functions they call.

This PR provides those improvements by using a new `container` Service to automatically execute/import certain services yet not execute them. This new service also provides a lookup that supports both standard ember DI lookup plus Class based lookup or these specific services. Lastly we also provide another debug function called `DataSourceRoutes()` which can be called from `console` which gives you a list of URLs and their mappings.

Overall this means the only Services we need to inject statically onto our DataSource service are now only the ones that don't have consistent URL/Service names (for example `oidc/` > `oidc-provider`). A last improvement here would be to remove the need to `reconcile` the data within the DataSource service, and at some point in the future we'll probably look to automatically reconcile the data with the backend from within the repositories themselves.

Notes:

We use an ember service called [ContainerDebugAdapter](https://api.emberjs.com/ember/release/classes/ContainerDebugAdapter) in order to inspect and retrieve classes automatically sourced, yet not executed/imported, by ember. This service has the 'debug' word in its name, so we checked to make sure this service exists during production (which it does), and its also publicly documented without mentioning anything about this being stripped from production builds (similar to other debug utilities). Just incase, we also looked into using the `ModuleRegistry` (which this service uses) that provides similar functionality but there didn't seem to be a public way to get to this class.

If using this `ContainerDebugAdapter` is ever problematic, we can see if theres another way to do this, or just hardcode in a list of the service names we want executing (we currently restrict this to execute only our repository classes)

One other worthwhile comment is that we no longer add a `@dataSource` to the abstract repository class for inherited methods, instead you have to override the inherited methods within the child class in order to decorate them with the correct URL, and in fact this is probably clearer to see what is going on, so is probably a positive here.

### DIFF
--- a/ui/packages/consul-ui/app/decorators/data-source.js
+++ b/ui/packages/consul-ui/app/decorators/data-source.js
@@ -1,16 +1,15 @@
+import { runInDebug } from '@ember/debug';
 import wayfarer from 'wayfarer';
-import { singularize } from 'ember-inflector';
 
 const router = wayfarer();
+const routes = {};
 export default path => (target, propertyKey, desc) => {
+  runInDebug(() => {
+    routes[path] = { cls: target, method: propertyKey };
+  });
   router.on(path, function(params, owner) {
-    let name;
-    if (params.modelName) {
-      name = singularize(params.modelName);
-    } else {
-      name = target.getModelName.call();
-    }
-    const instance = owner.lookup(`service:repository/${name}`);
+    const container = owner.lookup('service:container');
+    const instance = container.get(target);
     return configuration => desc.value.apply(instance, [params, configuration]);
   });
   return desc;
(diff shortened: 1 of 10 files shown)


kaxcode (approved):
LGTM

PR #9772, opened 2021-02-16
streaming: fix snapshot cache bug

dnephin:
Previously a snapshot created as part of a resume-stream request could have incorrectly cached the `newSnapshotToFollow` event. When the next subscribe request was received with `Index=0` the returned snapshot would cause clients to error because they received an unexpected framing event.

Fixed the bug by saving a snapshot without the framing event, then creating a new snapshot to prepend the framing event.


banks (approved):
Thanks LGTM!

i0rek (approved):
LGTM

PR #9773, opened 2021-02-16
Bump components to remove client-side md processing

zchsh:
This PR bumps `@hashicorp/react-vertical-text-block-list` and `@hashicorp/react-section-header` to recently published versions which do not process markdown within the component.

### Affected pages

- [`/community`](https://consul-git-zsbump-components.hashicorp.vercel.app/community), should be identical to [live site](https://www.consul.io/community), **except the link color should now be "Consul pink"**, as the current blue link colour is a minor visual bug


website/pages/community/index.jsx
```diff
@@ -18,27 +18,27 @@ export default function CommunityPage() {
           {
             header: 'Community Forum',
             body:
-              '[Consul Community Forum](https://discuss.hashicorp.com/c/consul)',
+              '<a href="https://discuss.hashicorp.com/c/consul">Consul Community Forum</a>',
```
zchsh, line 21: `vertical-text-block-list` no longer processes markdown internally, so we need to pass it HTML instead. This could be done automatically in `getStaticProps` if we feel it'd be worth it for authoring convenience.

website/pages/style.css
```diff
@@ -21,7 +21,6 @@
 @import '~@hashicorp/react-product-downloader/dist/style.css';
 @import '~@hashicorp/react-product-features-list/style.css';
 @import '~@hashicorp/react-search/style.css';
-@import '~@hashicorp/react-section-header/style.css';
```
zchsh, line 24: `section-header` now uses CSS modules, so this CSS import is no longer needed.

website/pages/community/style.css
```diff
@@ -1,4 +1,6 @@
 #p-community {
+  --brand: var(--consul);
```
zchsh, line 2: This ensures links within `vertical-text-block-list` are styled correctly.

zchsh:
Note: I need to add the `type/docs-cherrypick` label, but I don't have the permissions to do so

kylemac (approved):

Issue #9776, opened 2021-02-17
[Streaming] Results from nodes with streaming set header X-Consul-Knownleader: false


pierresouchay:
When Streaming is running, the header `X-Consul-Knownleader: false` is set to false.

Some SDKs such as Orbitz, use this field to throw an error, thus, it fails to retrieve data.

This header should be set to true with streaming activated

pierresouchay:
Cc @rickfast, @heruv1m, @yfouquet any opinion on this (and the linked PR and discussion)?

chuckyz:
@pierresouchay I encountered a situation in a very small production DC where `X-Consul-Knownleader: false` was appended only on a specific `X-Consul-Index` which ended up breaking all communication from a client perspective in one case (e.g.: `curl http://127.0.0.1/v1/health/service/foo?index=12345` returned a 500 but `curl http://127.0.0.1/v1/health/service/foo?index=12346` did not).  I've only ever seen this happen with grpc streaming.  Could this possibly be related?

pierresouchay:
@chuckyz Unlikely, in such case, the Consul agent does not return HTTP 500, it does return HTTP 200, but with such header wrongly set (which is interpreted by Orbitz as no-result). So, if you are sure it was an HTTP 500, it was not the same thing.

You could have however hit another but fixed (in our case) by https://github.com/hashicorp/consul/pull/9555 when a service is not changing too much (I mean less than every 10min), which is likely the case if you are using a vanilla Consul without any patch. See https://medium.com/criteo-engineering/consul-streaming-whats-behind-it-6f44f77a5175

dnephin:
Cross posting from #9778,

The [Consistency Modes](https://www.consul.io/api-docs/features/consistency) docs for this header say:

> can be used by clients to gauge the staleness of a result and take appropriate action

A false value for this header does not mean "there is no leader", it simply means that the leader status was not verified. There are plenty of places where we return `KnownLeader: false` simply because we didn't set the field, and when the response is returned from the cache we may set it to true, even though the leader status was not verified as part of that request.

It does not seem appropriate to error when this header is false.

pierresouchay:
Hello @dnephin ,

I highly disagree with your analysis of the doc and about who is wrong/right. From an implementor POV, this field means : the data you see is not correct, hence the choice of Orbitz to say: this data cannot be trusted. So it seems reasonable to make the choice of failing.

> There are plenty of places where we return KnownLeader: false simply because we didn't set the field

[...]

PR #9778, opened 2021-02-17
[Streaming] Set KnownLeader by default when streaming is activated

pierresouchay:
This is not a real fix but will avoid breaking clients relying on this header

Fix https://github.com/hashicorp/consul/issues/9776

I know the fix is not semantically correct (as we should carry the knowleader with streaming), but don't see how to do it as the data is not carried in anyway with GRPC on updates. Would probably require a refactoring.

### DIFF
--- a/.changelog/9778.txt
+++ b/.changelog/9778.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+streaming: return X-Consul-KnownLeader properly set when using streaming
+```
--- a/agent/cache-types/streaming_health_services.go
+++ b/agent/cache-types/streaming_health_services.go
@@ -156,12 +156,14 @@ func newHealthView(filterExpr string) (*healthView, error) {
 // (IndexedCheckServiceNodes) and update it in place for each event - that
 // involves re-sorting each time etc. though.
 type healthView struct {
-	state  map[string]structs.CheckServiceNode
-	filter filterEvaluator
+	state       map[string]structs.CheckServiceNode
+	filter      filterEvaluator
+	knownLeader bool
 }
 
 // Update implements View
 func (s *healthView) Update(events []*pbsubscribe.Event) error {
+	s.knownLeader = true
 	for _, event := range events {
 		serviceHealth := event.GetServiceHealth()
 		if serviceHealth == nil {
@@ -227,6 +229,9 @@ func (s *healthView) Result(index uint64) (interface{}, error) {
 		Nodes: make(structs.CheckServiceNodes, 0, len(s.state)),
 		QueryMeta: structs.QueryMeta{
 			Index: index,
+			// TODO: fill properly those fields, see https://github.com/hashicorp/consul/issues/9776
+			KnownLeader: s.knownLeader,
+			LastContact: 0,
 		},
 	}
 	for _, node := range s.state {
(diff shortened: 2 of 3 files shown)


dnephin (commented):
Thank you for the PR!

I was not familiar with this header, so I went looking in our docs. I found it referenced at the end of [Consistency Modes](https://www.consul.io/api-docs/features/consistency), but all it says is

> can be used by clients to gauge the staleness of a result and take appropriate action

As you say, reporting `true` in the case of streaming seems like it is a lie. I guess that it's also been a lie for any results from the cache. If the response came from a cache, the client doesn't actually know if there is an active leader.

I spoke with @kyhavlov and we think it might be better to omit this header instead of continuing to lie. From what I can, considering `X-Consul-KnownLeader=false` to be an error is incorrect. Maybe we should report this as a bug in Orbitz?

pierresouchay:
To me, interpretation of knowLeader=false being the data I am returning is not accurate and can't be trusted looks correct from an SDK implementors POV. To me it sounds like a reasonable choice. 

I though about removing the header, but it required more lines, and it sounds also more correct to me, but also more disruptive as we don't know if other implementors are not counting on that header even if missing from header (I mean being absent implies false) 

Maybe @yfouquet (an Orbitz contributor) has an opinion about this?

banks:
We'd still love to hear from users who'd be impacted by this change (especially Orbitz client contributors).

Currently though, the KnownLeader header doesn't really make a huge amount of sense for streaming. In order to make it behave in a way that is meaningful we'd need to add significant complexity to how streaming works (with regular updates about the server's state of whether it knows of a leader or not being sent as well as just actual result updates). And for now we're not sure there is much value in the feature - it's hard to see how a client will actually make use of that information since it can't force the agent to try a different server while using streaming.

As Daniel mentioned before it's not really correct or advisable to assume `KnownLeader = false` is an error state. So the only thing a client can do is decide to retry or not, but it the context of streaming, it can't even decide to use a different server to see if it gets a less stale result.

Even if we made some way for that to happen, from experience we've seen that while it sounds desirable, setting a limit on how stale things can be almost always causes more problems than it solves. It could in theory limit staleness of a read in some cases where a single server is partitioned from the rest of the servers but the clients can still speak to all of them - if the client retries with a way to steer directly to the leader, then they can "recover" from the stale server.

The problem is that that condition is _extremely_ rare - I've never heard of it happening on a real Consul cluster. What _does_ happen very often is that servers get "stale" or have "no leader" due to being overloaded and unstable. In this case, having all the clients after some timeout suddenly switch to making more expensive requests just increases the load on all servers and the leader but doesn't actually get a less stale result. In other words, it just makes the overload issues even worse.

So our inclination is to remove these headers from streaming (and probably cached) responses and note that in the upgrade guide.

We'd encourage any client code that uses it in logic to avoid assuming a missing header (or false value) means that there is no cluster leader - this is already incorrect behaviour as Daniel pointed out above.

dnephin:
Thank you again for making this change.

Unfortunately the logic changed in this PR has moved to `agent/rpcclient/health` so this commit can't be merged anymore.

Given the discussion above, and that we are already reporting `KnownLeader: false` in a bunch of places, and that it does not mean "no leader", I'm going to close this PR. Let's continue the discussion in #9776.

PR #8387, opened 2020-07-28
Introducing changelog-gen

i0rek:
This PR introduces [changelog-gen](https://github.com/hashicorp/go-changelog) for consul, with the goal to automatically generate our changelog in the future. This PR is more of an introduction though, it adds entries for the currently unreleased PRs and showcases how to generate the changelog for them.

There are two notable things:

1) the templates for individual entries and the changelog as a whole: `changelog.tmpl` and `note.tmpl`
2) a file for each PR, containing the changelog entry

When running `changelog-gen`, it looks like this:

```
$ changelog-build -last-release v1.8.0 -entries-dir .changelog/ -changelog-template changelog.tmpl -note-template note.tmpl -this-release 86b6b38faa7c69f26f1d4c71e271cd4285daadf9
```

```markdown
FEATURES:

* acl: Added ACL Node Identities for easier creation of Consul Agent tokens. [[GH-7970](https://github.com/hashicorp/consul/pull/7970)]

IMPROVEMENTS:

* acl: allow auth methods created in the primary datacenter to optionally create global tokens. [[GH-7899](https://github.com/hashicorp/consul/pull/7899)]
* agent: Allow to restrict servers that can join a given Serf Consul cluster. [[GH-7628](https://github.com/hashicorp/consul/pull/7628)]
* connect: Append port number to expected ingress hosts. [[GH-8190](https://github.com/hashicorp/consul/pull/8190)]
* connect: support Envoy v1.14.4, v1.13.4, v1.12.6. [[GH-8216](https://github.com/hashicorp/consul/pull/8216)]
* connect: various changes to make namespaces for intentions work more like for other subsystems. [[GH-8194](https://github.com/hashicorp/consul/pull/8194)]
* dns: Improve RCODE of response when query targets a non-existent datacenter. [[GH-8102](https://github.com/hashicorp/consul/issues/8102)] [[GH-8218](https://github.com/hashicorp/consul/pull/8218)]
* version: The `version` CLI subcommand was altered to always show the git revision the binary was built from on the second line of output. Additionally the command gained a `-format` flag with the option now of outputting the version information in JSON form. **NOTE** This change has the potential to break any parsing done by users of the `version` commands output. In many cases nothing will need to be done but it is possible depending on how the output is parsed. [[GH-8268](https://github.com/hashicorp/consul/pull/8268)]

BUG FIXES:

[...]

PR #9781, opened 2021-02-17
update changelog and website checks

alvin-huang:
This PR splits out the `website/` check to only run on PR `opened` while the changelog check runs on `opened`, `synchronize` (new commits) and `labeled` and also changes it to a status check that fails when there is no `pr/no-changelog` label applied nor is there a `.changelog/$pr_number.txt` file. 

I also changed the `on` trigger to be `pull_request` for the changelog check because it no longer needs a token secret to post to the PR so it doesn't need to only run on `pull_request_target`

### DIFF
--- a/.github/workflows/changelog-checker.yml
+++ b/.github/workflows/changelog-checker.yml
@@ -1,0 +1,36 @@
+# This workflow checks that there is either a 'pr/no-changelog' label applied to a PR
+# or there is a .changelog/<pr number>.txt file associated with a PR for a changelog entry
+
+on:
+  pull_request:
+    types: [opened, synchronize, labeled]
+    # Runs on PRs to master and all release branches
+    branches:
+      - master
+      - release/*
+
+jobs:
+  # checks that a .changelog entry is present for a PR
+  changelog-check:
+    # If there  a `pr/no-changelog` label we ignore this check
+    if: "!contains(github.event.pull_request.labels.*.name, 'pr/no-changelog')"
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


alvin-huang:
> 🤔 Double check that this PR does not require a changelog entry in the `.changelog` directory. [Reference](https://github.com/hashicorp/consul/pull/8387)

This will go away after this gets merged, it's running the old check in the `master` branch.

.github/workflows/changelog-checker.yml
```diff
@@ -0,0 +1,36 @@
...
+  changelog-check:
+    # If there  a `pr/no-changelog` label we ignore this check
+    if: "!contains(github.event.pull_request.labels.*.name, 'pr/no-changelog')"
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0 # by default the checkout action doesn't checkout all branches
```
dnephin, line 23: Can we omit this now that it uses `on: pull_request` instead of target?

dnephin (approved):
LGTM!

One question, but not blocking, probably just a minor optimization.

.github/workflows/changelog-checker.yml
alvin-huang: I can test but I don't think so, we still need to checkout all branches to compare the HEAD to the base ref (which wouldn't be checked out if we just checkout HEAD).

PR #9783, opened 2021-02-17
remove reference to docker/ path for old docker mirror

alvin-huang:
Followup to #9782, there was an old lingering `/docker` namespace that we don't need anymore.


mdeggies (approved):

alvin-huang:
`release/1.8.x` and `release/1.7.x` already had the old mirror address so it shouldn't make any changes but applying the backport label for posterity to capture this change.

PR #9792, opened 2021-02-19
command/kv: Add prefix option to kv import command

dzeban:
Currently when data is imported via `consul kv import` it overwrites
keys under the root key. Since `consul kv export` can retrieve data for
the given prefix, i.e. part of the KV tree, importing it under root may
be not what users want.

To mirror prefix behavior from export this PR adds prefix feature to the
import command that adds prefix to all keys that are imported.

### DIFF
--- a/.changelog/9792.txt
+++ b/.changelog/9792.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+cli: Add prefix option to kv import command
+```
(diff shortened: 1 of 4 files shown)


dzeban:
Please advise on the correct PR labels and whether I need to update changelog.

kyhavlov (approved):

kyhavlov:
Thanks for the PR @dzeban! I went ahead and added a changelog entry before merging.

PR #9793, opened 2021-02-19
ui: Add an optional environment variable to control how testem tests start

johncowen:
Here we look for a TESTEM_AUTOLAUNCH environment variable, which can be
set to either Chrome, Firefox or Safari, which will control which
browser to automatically start when running testem tests.

If the variable is set to anything else, then it will not automatically
start a browser in order to run the tests and you will need to visit the
tests manually. e.g.:

TESTEM_AUTOLAUNCH=0 make test-oss-view

Previously we only ever tested in Chrome and therefore there are no
specific settings for Firefox or Safari. If specific settings are
required for these browsers they can be added at a later date.


kaxcode (approved):
:shipit:

PR #9801, opened 2021-02-22
ui: Update authorization things to use new DataSource decorator/params interface

johncowen:
In a parallel timescale to https://github.com/hashicorp/consul/pull/9746 we added auth-methods and a new set of methods for authorization within repositories. Now that work is merged we've added the changes required to bring that work up to date with https://github.com/hashicorp/consul/pull/9746 here in a separate PR for keeping things easy to see those specific changes.

Few notes:

1. This mainly moves everything from using multiple arguments to using a params object. Previously we'd preferred multiple arguments to avoid having to remember the names of params/object properties when using the repository methods. Now we have our `@dataSource` decorator, the names of the properties of the `params` object are defined right alongside where they are used, and DataSource usage now uses URIs to 'call' specific methods (along with automatically generated documentation for these URLs). So this now gives us the best of both worlds.
2. The names of methods on repositories are now completely unimportant to the end user, seeing as you call them via URIs.  Additionally now we pass a params object through to the method, names like 'findBySlug' don't quite match up as this could just be 'findOne' or `findById`, the names are now almost unimportant apart from for documentation purposes. So there's likely to be some renaming at some point in the near future.
3. The `repository` approach that we've used from the very beginning of the rebuild of the UI to provide some abstraction over `ember-data` is now completely decoupled from the rest of the app. Our self-documenting DataSource usage and their URI based data/resource identification/retrieval means that the app could be completely unaware of the concept of 'repositories' and the method names used in order to retrieve data. There is still a little work to move over to DataSources everywhere (mainly within the ACLs area) but this is also likely to be updated to use DataSources in the near future.

### DIFF
--- a/ui/packages/consul-ui/app/adapters/permission.js
+++ b/ui/packages/consul-ui/app/adapters/permission.js
@@ -4,20 +4,20 @@ import { inject as service } from '@ember/service';
 export default class PermissionAdapter extends Adapter {
   @service('env') env;
 
-  requestForAuthorize(request, { dc, ns, permissions = [], index }) {
+  requestForAuthorize(request, { dc, ns, resources = [], index }) {
     // the authorize endpoint is slightly different to all others in that it
     // ignores an ns parameter, but accepts a Namespace property on each
     // resource. Here we hide this different from the rest of the app as
-    // currently we never need to ask for permissions/resources for mutiple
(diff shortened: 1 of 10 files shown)


kaxcode (approved):
LGTM

PR #9805, opened 2021-02-22
structs: remove EnterpriseMeta.GetNamespace

dnephin:
I added this recently without realizing that the method already existed and was named NamespaceOrEmpty. Replace all calls to GetNamespace with NamespaceOrEmpty or NamespaceOrDefault as appropriate.


banks (approved):
👍 Thanks for cleaning things up!

dnephin:
The unit test failure is a flake, and the integration test failure was fixed in #9853

PR #9807, opened 2021-02-22
test: omit envoy golden test files that differ from the latest version

rboyer:
Since we currently do no version switching this removes 75% of the PR
noise.

To generate all *.golden files were removed and then I ran:

    go test ./agent/xds -update

For quality of life this can be backported to 1.9, 1.8, and 1.7 but they'll all have to be manual.


freddygv (approved):
🙏 🙏

agent/xds/golden_test.go
```diff
@@ -45,14 +53,38 @@ func golden(t *testing.T, name, subname, got string) string {
...
+	// To trim down PRs, we only create per-version golden files if they differ
+	// from the latest version.
+	newestExpected := ""
+	if newestSubname != "" && subname != newestSubname {
+		newestGolden := filepath.Join("testdata", fmt.Sprintf("%s.%s.golden", name, newestSubname))
+		expected, err := ioutil.ReadFile(newestGolden)
+		require.NoError(t, err)
+
+		if string(expected) == got {
+			if *update && got != "" {
```
markan, line 66: I couldn't figure out what was going on here; the original optional writes the file, then returns the contents, which seems straightforward, and provides a path to regenerate the golden files. But I'm not sure why we delete the golden file, but don't do anything with the newest golden, or why this is guarded by the match with expected.

Perhaps a comment explaining the new workflow, in particular if the version under test isn't the newest.

markan (commented):
This looks great, I like the fact that a lot of mechanical duplication is being removed.

It might be worthwhile to document how the new flow works with .golden files, especially when they need to be recreated, or exceptions managed.

agent/xds/golden_test.go
rboyer: Yeah I'll add a bunch of comments.

The gist is that when you run the tests with the `-update` flag they will erase any golden test files for prior versions that are identical to the golden test file for the latest version of envoy.

During comparison mode, we check for a specific version file if one exists and use that for comparison, but if that does not exist we use the one from the latest version of envoy instead.

rboyer: @markan check the latest commit in the PR for the restructured function that should hopefully be way clearer.

markan (approved):

PR #9812, opened 2021-02-23
ui: Side navigation tweaks

johncowen:
This PR adds a few little amends to the new sidebar navigation:

1. Larger Consul logo
2. Use dark menus for header menus as well as side menus.
3. Remove app footer and move Consul version output to the Help menu
4. Revert unique ID usage so we can keep using the login modal id throughout the app to toggle it instead of having to pass around JS things.


kaxcode (approved):
:shipit:

jnwright:
Looks like the color of the version needs to be changed to be the same as the section headers (ui-gray-400). Font size also should be 12px

johncowen:
Hey @jnwright

Oh good spot! I knocked the gray on the version down a couple of notches so its equivalent to `ui-gray-400`.

We use relative sizes for font-sizing (not pixels), the version is `0.75rem` which, in our app, equates to `12px`. 

Lemme know if that makes sense.

PR #8779, opened 2020-09-30
ui: Outlet Loading

johncowen:
The Route loading logic in the UI has historically been spread out amongst various pieces of the application, in the AppView component (sometimes), the Application route and the Feedback service.

The sprawly nature of this along with the fact that you could easily forget to set certain properties when creating new routes meant it was more than straightforwards for loading to 'just work' especially in sub-routes and sub-sub-routes etc. We took a look at using model hooks and loading.hbs templates, but both options had their downsides, and both required extra things to be done in order for loading to 'just work' whether adding hooks or files.

We realized that if each `{{outlet}}` was  wrapped in a `<Outlet data-outlet="dc">{{outlet}}</Outlet>` component, `dc` being the name/identifier of the current `Route`/`template` (the `Route.routeName`), then we could automatically figure out which outlet in the Route hierarchy was to be loaded by traversing up the tree of outlets and comparing the clicked route name with the name of the outlet.

For example:

```
Clicking: dc.services.show.healthcheck

- Outlet: `application`
  - Outlet: `dc`
    - Outlet: `dc.services.show` <== dc.services.show.healthcheck - Match, this outlet will be loading
      - Outlet: `dc.services.show.intentions` <== dc.services.show.healthcheck - No Match

Clicking: dc.services.index

- Outlet: `application`
  - Outlet: `dc` <== dc.services.index - Match, this outlet will be loading
    - Outlet: `dc.services.show` <== dc.services.index - No Match
      - Outlet: `dc.services.show.intentions` <== dc.services.index - No Match
```

In our case we add a `data-state="loading"` attribute to the outlet when we find the match. Once the router has finished transitioning, we set the state of the loading outlet back to idle.

Using this attribute we can do all of out route loading visuals via CSS by hiding and showing the unloaded and subsequently loaded outlet, which massively reduces the spread of code used to control loading state in the application. Future work could involve a more subtle transition between loading pages, if that was something we wanted to do.

The one downside is that the only way to 'name' the outlets correctly is to have access to the `Route.routeName` property, something that is only available from within the Route. We tried various methods of retrieving this and passing it through to the template, and eventually settled on using a base Route class, which all of our Routes now inherit from. The approach means:

1. We aren't using reopen (not Octane friendly)
2. We don't have to manually set the routeName in every Route or template.
3. We have to remember to inherit from this base class for every route, AND wrap outlets in the Outlet component, something we can probably mitigate somewhat by adding custom blueprints in the future.

[...]

PR #9813, opened 2021-02-23
ui: Add Route component / routlet service

johncowen:
Back in https://github.com/hashicorp/consul/pull/8779 we added a new `<Outlet />` component which we use to automatically figure out the loading states for the hierarchy of templates/outlets as well as help us to provide unique CSS selectors to target individual outlets/routes in order to apply those tiny individual styling tweaks that are sometimes needed.

This PR builds upon that work by providing a new `<Route>` component and a new `routlet` service which ties both the `<Outlet />` and `<Route />` components together to provide various different pieces of functionality. All in all this combined work gives us:

1. Automatic loading/idle state of deeply nested templates/routes. (explained in https://github.com/hashicorp/consul/pull/8779)
2. Unique CSS selectors that can be used to target individual pages/routes/outlets in a very specific way (explained in https://github.com/hashicorp/consul/pull/8779)
3. A stricter approach to 'data down, actions up' and the ability to easily pass component yielded data and/or yielded component methods from a component or route in one template, through the 'routlet' into the scope of a more deeply nested route/template (added in this PR)
4. Add the ability to easily add visual transitions between 2 routes, for example if we want to add a very slight transition/animation when you navigate from one route to another (this PR)
5. Add the potential to provide a11y friendly route navigation by announcing page changes when you navigate between routes (a future PR)

Please see https://github.com/hashicorp/consul/pull/8779 for an explanation of points 1 and 2.

This PR adds the `Route` component and only uses it in one area of the catalog area of the UI just for usage reference for this PR, but the idea is that every single template will have a top level `Route` component, just as every single template with an outlet has an `Outlet` component.

### Data down

Once we have this we can achieve the following:

```hbs
{{!  the dc template }}
<Outlet @name="dc" @model={{hash dc="dc-1"}}>
{{! the dc.services.index template appears here }}
</Outlet>

{{! the dc.services.index template  }}
<Route @name="dc.service.index" as |route|>
{{route.model.dc}} === "dc-1"
</Route>
```

We currently use ember `Routes` and `modelFor` to achieve this functionality but this new 'routlet' approach has several advantages:

1. Using `this.modelFor('dc')` from an ember `Route` is essentially 'reaching up' through the route hierarchy to retrieve data rather than passing 'data down' as is usually recommended.
2. Using `this.modelFor('dc')` from an ember `Route` means the `Route` has access to anything it likes. Passing the required data down through the 'routlet' gives a little more control - the nested template only has access to the data that has been passed to it by its parent template.
3. We can access data that has been yielded/exported by other components and pass that down through the template hierarchy. The following example using a 'routlet' to cross the template hierarchy is not really possible via ember Routes and model hooks alone:

```hbs
{{! application.hbs }}
<HashicorpConsul as |loginModal|>
  <Outlet @model={{hash loginModal=loginModal}}>
  </Outlet>
</HashiCorpConsul>

{{! dc.hbs }}
<Route as |route|>
  <button
    {{on 'click' route.model.loginModel.open}}
  >
    Open Login
  </button>
</Route>

```
This means we can reduce the amount of code/Routes that only exist in order to 'reach up' to the parent route in order to make data available in the template.

For the times when we still need to access this data from an ember Route, we also provide a `@service('routlet')` so you can access the same data from an ember Route just as you would using the normal `modelFor` method:

```javascript
@service('routlet') routlet;
async model() {
  const dc = this.routlet.modelFor('dc').dc;
}
```

### Visual transitions

Whilst in this PR we don't add any visual transitions when navigating between routes, we add the machinery here to allow us to do this easily. `<Outlet>`s now have a `transitionend` event listener applied to them. So, only if they have a CSS transition duration, they will resolve a promise that is shared across the 'routlet'. This promise can be accessed via the `routlet` service:

```javascript
@service('routlet') routlet;
async model() {
  // request the data and start loading it whilst we are transitioning
  const data = this.startLoadingDataButDontAwaitItYet();
  // also wait/block until the transition out of the previous route has finished
  await this.routlet.ready();
  // we know the transition has finished so just wait until our data has loaded
  // if the data is loaded by now, then we won't even see a loader
  // if it hasn't we show a loader until it has
  return {
   data: await data
  }
}
```

We still need to decide whether this is something we actually want to do, so we don't use this here yet, but the machinery is there to be able to do this.

A small tangent to this was changing our `is-href` helper to 'activate' buttons/links when you click them rather than when the route has loaded. We achieve this using an event listener on the ember `router` service rather than an old style observer, which overall makes the UI feel a lot more responsive now that buttons stay highlighted when you click on them.

### A11y route navigation

Once every page/tempalte has a wrapping `<Route />` this means we can centralize our page title functionality, possibly like:

```hbs
<Route @title="Services" />

vs

{{page-title "Services"}}
```

This (hopefully) means we can potentially implement page navigation announcements in a centralized place in order to improve the a11y of the UI. This feature seems to be the number one priority for improving a11y generally for ember apps. 

There is already quite a lot of functionality all surrounding this 'routlet' concept, so we've left this for a future PR.

One final tiny little detail, I've found the spelling/typing of `routlet` vs `routelet` to be a tiny-teeny bit awkward sometimes, but I eventually decided on the `routlet` spelling for if you ever need to search across the project for `route` as opposed to `routlet`. Using the `routlet` spelling means `routlet` will be excluded from a search for `route`.

### DIFF
--- a/ui/packages/consul-ui/app/components/main-nav-horizontal/index.scss
+++ b/ui/packages/consul-ui/app/components/main-nav-horizontal/index.scss
@@ -9,7 +9,6 @@
   @extend %main-nav-horizontal-action;
 }
 %main-nav-horizontal .popover-menu [type='checkbox']:checked + label > *,
-%main-nav-horizontal > ul > li > a:active,
 %main-nav-horizontal > ul > li.is-active > a,
 %main-nav-horizontal > ul > li.is-active > label > * {
   @extend %main-nav-horizontal-action-active;
--- a/ui/packages/consul-ui/app/components/main-nav-vertical/index.scss
+++ b/ui/packages/consul-ui/app/components/main-nav-vertical/index.scss
@@ -5,7 +5,6 @@
 %main-nav-vertical > ul > li > span {
   @extend %main-nav-vertical-action;
 }
-%main-nav-vertical > ul > li > a:active,
 %main-nav-vertical > ul > li.is-active > a {
   @extend %main-nav-vertical-action-active;
 }
--- a/ui/packages/consul-ui/app/components/outlet/index.hbs
+++ b/ui/packages/consul-ui/app/components/outlet/index.hbs
@@ -1,9 +1,10 @@
-  {{did-insert this.connect}}
-  {{will-destroy this.disconnect}}
+{{did-insert this.connect}}
+{{will-destroy this.disconnect}}
 <section
+  {{did-insert (fn this.attributeChanged 'element')}}
   class="outlet"
   data-outlet={{@name}}
-  data-route={{this.route}}
+  data-route={{this.routeName}}
   data-state={{this.state.name}}
   data-transition={{concat this.previousState.name ' ' this.state.name}}
 >
--- a/ui/packages/consul-ui/app/components/outlet/index.hbs
+++ b/ui/packages/consul-ui/app/components/outlet/index.hbs
@@ -7,6 +7,7 @@
   data-route={{this.routeName}}
   data-state={{this.state.name}}
   data-transition={{concat this.previousState.name ' ' this.state.name}}
+  {{on 'transitionend' this.transitionEnd}}
 >
   {{yield (hash
     state=this.state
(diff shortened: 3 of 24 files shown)


kaxcode (changes requested):
LGTM ~whoops, accidentally requested changes.

kaxcode (approved):

PR #9817, opened 2021-02-23
example and default syslog CTS settings

eikenb:
Kept it simple and purposely left off 'enabled' to highlight option that it is auto-enabled if any other value is set.


lornasong (approved):
Looks good to me! I find the example of the non-default value syslog.facility helpful

PR #9819, opened 2021-02-24
ui: A11y modals

johncowen:
This PR uses the excellent `a11y-dialog` to implement our modal functionality across the UI.

This package covers all our a11y needs - overlay click and ESC to close, controlling `aria-*` attributes, focus trap and restore. It's also very small (1.6kb) and has good DOM and JS APIs and also seems to be widely used and well tested.

There is one downside to using this, and that is:

We made use of a very handy characteristic of the relationship between HTML labels and inputs in order to implement our modals previously. Adding a `for="id"` attributes to a label means you can control an `<input id="id" />` from anywhere else in the page without having to pass javascript objects around. It's just based on using the same string for the `for` attribute and the `id` attribute. This allowed us to easily open our login dialog with CSS from anywhere within the UI without having to manage passing around a javascript object/function/method in order to open the dialog.

We've PRed https://github.com/hashicorp/consul/pull/9813 which includes an approach which would make passing around JS `modal` object easier to do. But in the meantime we've added a little 'hack' here using an additional `<input />` element and a `change` listener which allows us to keep this label/input characteristic of our old modals. I'd originally thought this would be a temporary amend in order to wait on https://github.com/hashicorp/consul/pull/9813 but the more I think about it, the more I think its quite a nice thing to keep - so longer term we may/may not keep this.

Lastly we noticed that the modal CSS could be cleaned up slightly, and are considering changing the name of the component to something less tied to the word 'modal' (a11y-dialog explains and provides a difference between a modal dialog and a non-modal dialog) but we can think about this and do this is a future PR.


kaxcode (approved):
LGTM

johncowen:
I'm gonna merge this, but just a short note. I got a flakey test here for the first time in a long time, so we _may_ have a testing only timing issue here somewhere, which wouldn't massively surprise me considering this PR completely changes the way we do our modals (HTML based vs JS based). I'd rather merge this and keep an eye on it though, thought it'd be worthwhile to share here incase.

PR #9835, opened 2021-02-26
ui: Implement ACLs access based on ACLs

johncowen:
Continuing on from the work done in https://github.com/hashicorp/consul/pull/9687 this PR adds restrictions to everything within the ACLs (and nspaces) area based on your ACLs (including readonly views etc.)

An extra note here, in previous version of the UI we have a completely different method of gating access to the ACLs area dating from the beginning of the new UI. This is probably redundant now that we inspect ACLs globally using our new `authorize` endpoint. Whether it is or not is a little unclear as the previous approach also checks for existence of the old ACLs system and whether ACLs are enabled or not. I'll look a bit deeper into this an possibly completely remove all the old style of gating in a separate PR.

### DIFF
--- a/ui/packages/consul-ui/app/abilities/auth-method.js
+++ b/ui/packages/consul-ui/app/abilities/auth-method.js
@@ -1,7 +1,7 @@
 import BaseAbility from './base';
 import { inject as service } from '@ember/service';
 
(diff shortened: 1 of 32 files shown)


kaxcode (approved):
LGTM

PR #9845, opened 2021-03-05
ui: Create auth-method show page with General Info Tab

kaxcode:
✨  Creates the show page for auth-methods of type Kubernetes, OIDC, and JWT.

[Visit demo
](https://consul-ui-staging-kslwven8x-hashicorp.vercel.app/ui/dc1/services#CONSUL_ACLS_ENABLE=1)
auth-method-show-oidc


kaxcode:
@johncowen ready for review 😄

ui/packages/consul-ui/app/components/certificate/index.hbs
```diff
@@ -0,0 +1,14 @@
+<div class="certificate">
+  <CopyButton @value={{@item}} @name={{@name}} />
+  <button
+    type="button"
+    class={{concat "visibility" (if this.show " hide" " show")}}
+    {{on "click" this.setVisibility}}
```
johncowen, line 6: Could we use the `set` helper here?

ui/packages/consul-ui/app/components/certificate/index.js
```diff
@@ -0,0 +1,14 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+
+export default class TopologyMetrics extends Component {
```
johncowen, line 5: If you don't want to go with the `set` helper the name of this class should be `Certificate`.

ui/packages/consul-ui/app/components/certificate/index.scss
```diff
@@ -0,0 +1,26 @@
...
+    @extend %with-visibility-show-icon, %as-pseudo;
+  }
+  div.key {
+    background-color: var(--gray-050);
+    overflow-wrap: anywhere;
+  }
+  hr {
+    border: 3px dashed var(--gray-300);
+    background-color: $white;
+    width: 10vw;
```
johncowen, line 22: No biggie, wondered if this should be a fixed width so it doesn't change when I resize my window?

ui/packages/consul-ui/app/components/consul/auth-method/index.scss
```diff
@@ -1,9 +1,74 @@
...
+    flex-wrap: wrap;
+    > dt:last-of-type,
+    > dd:last-of-type {
+      border-bottom: 1px solid $gray-300 !important;
+    }
+    dt, dd {
+      padding: 12px 0;
+      margin: 0;
+      border-top: 1px solid $gray-300 !important;
+      color: $black !important;
```
johncowen, line 45: I think we decided that we were going to move all these grays to use CSS properties instead of SASS vars, then we don't have a mix of `$gray-***` and `var(--gray-***)`

ui/packages/consul-ui/app/components/consul/auth-method/view/index.hbs
```diff
@@ -0,0 +1,217 @@
...
+      <dt>CA Cert</dt>
+      <dd>
+        <Certificate @item={{@item.Config.CACert}} @name={{t 'models.auth-method.Config.CACert'}} />
+      </dd>
+      {{/if}}
+      {{#if @item.Config.ServiceAccountJWT}}
+      <dt>Service account JSON Web Token</dt>
+      <dd>
+        <CopyButton @value={{@item.Config.ServiceAccountJWT}} @name="Service account JSON Web Token" />
+        <span>
```
johncowen, line 30: I think there're some `t`s missed here, not sure if you meant to do all these or not? Seems like sometimes `t` is used and sometimes not?

ui/packages/consul-ui/app/components/consul/auth-method/view/index.hbs
```diff
@@ -0,0 +1,217 @@
...
+            <td>{{get entry 0}}</td>
+            <td>{{get entry 1}}</td>
+          </tr>
+          {{/each}}
+        </tbody>
+      </table>
+    </section>
+    {{/if}}
+  {{/if}}
+  </div>
```
johncowen, line 217: I think these two sections need empty states when they are empty. Would be good for another PR though.

ui/packages/consul-ui/mock-api/v1/acl/auth-method/_
```diff
@@ -14,17 +27,45 @@ ${
       case 'oidc':
         config = {
           OIDCDiscoveryURL: `https://${fake.internet.ip()}:8443`,
+          OIDCDiscoveryCACert: `-----BEGIN CERTIFICATE-----${fake.internet.password(1357)}-----END CERTIFICATE-----`,
```
johncowen, line 30: Do we know the format of how these come out of consul in the API?

'Generally' they have returns in them (which might mean you don't need to do the hard wrapping in the CSS to show these). If they come out without the returns then don't worry!

ui/packages/consul-ui/translations/en-us.yaml
```diff
@@ -180,3 +180,35 @@ components:
...
+      JWKSURL: JWKS URL
+      JWKSCACert: JWKS CA Cert
+      JWTValidationPubKeys: JWT validation pub keys
+      OIDCDiscoveryURL: Discovery URL
+      JWTSupportedAlgs: JWT supported algorithms
+      BoundAudiences: Bound audiences
+      BoundIssuer: Bound issuer
+      ExpirationLeeway: Expiration leeway
+      NotBeforeLeeway: Not before leeway
+      ClockSkewLeeway: Clocks skew leeway
```
johncowen, line 203:
```suggestion
      ClockSkewLeeway: Clock skew leeway
```

johncowen (approved):
Thanks for the Demo Link, I tried it out and all looks good!

I left a few minor things, if you wanted to look at those before merging, but otherwise all good!

ui/packages/consul-ui/app/components/certificate/index.scss
kaxcode: 💯

ui/packages/consul-ui/mock-api/v1/acl/auth-method/_
kaxcode: I assume the same way certs come from other endpoints. This is  why I'm mocking it out the same way.

ui/packages/consul-ui/app/components/consul/auth-method/view/index.hbs
kaxcode: Whoops! Looks like I forgot to fix up the 'kubernetes' section.

ui/packages/consul-ui/app/components/certificate/index.hbs
kaxcode: I'll keep the action.

PR #9847, opened 2021-03-08
ui: CSP Improvements

johncowen:
This PR allows folks to set a stricter CSP policy for the UI. Mainly by configuring `ember-auto-import` to not require `unsafe-eval` and by re-arranging things so we no longer need to use inline scripts for some of our stranger requirements (pre vendor polyfills and CodeMirror syntax highlight loading).

We also set our development-time CSP to the strictest policy we can currently use so we are nudged if anything doesn't adhere to policy.

More details in the commit messages.

There are still some minor (inline CSS related) improvements we can still make here, but we now support the strictest JS related CSP available:

```
default-src 'self' img-src 'self' data: ; style-src 'self' 'unsafe-inline'
```

Removal of the `unsafe-inline` for style/CSS is likely to be PRed at a later date


kaxcode (approved):
🔒 LGTM

PR #9855, opened 2021-03-09
Support Incremental xDS mode

rboyer:
This adds support for the [Incremental xDS](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#incremental-xds) protocol when using xDS v3. This is best reviewed commit-by-commit and will not be squashed when merged.

Union of all commit messages follows to give an overarching summary:

- xds: exclusively support incremental xDS when using xDS v3
    - Attempts to use SoTW via v3 will fail, much like attempts to use incremental via v2 will fail.
    - Work around a strange older envoy behavior involving empty CDS responses over incremental xDS.
-  xds: various cleanups and refactors that don't strictly concern the addition of incremental xDS support
    - Dissolve the connectionInfo struct in favor of per-connection ResourceGenerators instead.
    - Do a better job of ensuring the xds code uses a well configured logger that accurately describes the connected client.
-  xds: pull out checkStreamACLs method in advance of a later commit
-  xds: rewrite SoTW xDS protocol tests to use protobufs rather than hand-rolled json strings
   
    - In the test we very lightly reuse some of the more boring protobuf construction helper code that is also technically under test. The important thing of the protocol tests is testing the _protocol_. The actual inputs and outputs are largely already handled by the xds golden output tests now so these protocol tests don't have to do double-duty.
    
    - This also updates the SoTW protocol test to exclusively use xDS v2 which is the only variant of SoTW that will be supported in Consul 1.10.
- xds: default xds.Server.AuthCheckFrequency at use-time instead of construction-time

TODO

- [x] more tests: https://github.com/hashicorp/consul/issues/10125
- [ ] decide if the `logTrace*` function should be removed
- [ ] docs (likely deferred and rolled into other pending docs updates)


agent/xds/clusters.go
```diff
@@ -20,11 +20,10 @@ import (
 	"github.com/hashicorp/consul/agent/connect"
 	"github.com/hashicorp/consul/agent/proxycfg"
 	"github.com/hashicorp/consul/agent/structs"
-	"github.com/hashicorp/consul/logging"
 )
 
 // clustersFromSnapshot returns the xDS API representation of the "clusters" in the snapshot.
-func (s *Server) clustersFromSnapshot(_ connectionInfo, cfgSnap *proxycfg.ConfigSnapshot) ([]proto.Message, error) {
+func (s *ResourceGenerator) clustersFromSnapshot(cfgSnap *proxycfg.ConfigSnapshot) ([]proto.Message, error) {
```
rboyer, line 26: I figure after this lands I can do a big `s/\bs\b/g/g` rename of the receiver but I didn't want to touch all of the lines in this PR.

agent/xds/server.go
```diff
@@ -193,14 +210,20 @@ func (s *Server) process(stream ADSStream, reqCh <-chan *envoy_discovery_v3.Disc
 
 	// Loop state
 	var (
-		cfgSnap       *proxycfg.ConfigSnapshot
-		req           *envoy_discovery_v3.DiscoveryRequest
-		node          *envoy_config_core_v3.Node
-		proxyFeatures supportedProxyFeatures
```
rboyer, line 199: `proxyFeatures` moved into the `ResourceGenerator`

agent/xds/server.go
```diff
@@ -209,13 +232,13 @@ func (s *Server) process(stream ADSStream, reqCh <-chan *envoy_discovery_v3.Disc
...
 		EndpointType: {
+			generator: g,
 			typeURL:   EndpointType,
-			resources: s.endpointsFromSnapshot,
 			stream:    stream,
 		},
 		ClusterType: {
+			generator: g,
 			typeURL:   ClusterType,
-			resources: s.clustersFromSnapshot,
```
rboyer, line 218: This type-switched method call is now handled by a helper function and shared between sotw and delta.

test/integration/connect/envoy/consul-base-cfg/base.hcl
```diff
@@ -1 +1,2 @@
-primary_datacenter = "primary"
\ No newline at end of file
+primary_datacenter = "primary"
+log_level          = "trace"
```
rboyer, line 2: If we keep the `logTrace*` functions, this is super helpful because it'll log the full protocol chatter to TRACE.

test/integration/connect/envoy/case-wanfed-gw/secondary/server.hcl
```diff
@@ -7,6 +7,7 @@ primary_gateways = [
   "consul-primary:4431",
 ]
 primary_gateways_interval = "5s"
+retry_interval_wan        = "5s"
```
rboyer, line 10: This defaults to `30s` which is a great way to have the `TestEnvoy/case-wanfed-gw` test accidentally take long enough that the bats test framework assumes the test has failed.

test/integration/connect/envoy/case-wanfed-gw/primary/verify.bats
```diff
@@ -10,6 +10,18 @@ load helpers
   assert_upstream_has_endpoints_in_status 127.0.0.1:19000 secondary HEALTHY 1
 }
 
+@test "gateway-primary should have healthy endpoints for secondary servers" {
```
rboyer, line 13: Adding more explicit line-item steps here to aid in future stuck state debugging.

agent/xds/server_test.go
```diff
@@ -1,130 +1,95 @@
...
 	"github.com/hashicorp/consul/acl"
-	"github.com/hashicorp/consul/agent/cache"
 	"github.com/hashicorp/consul/agent/proxycfg"
 	"github.com/hashicorp/consul/agent/structs"
-	"github.com/hashicorp/consul/sdk/testutil"
 )
 
-// testManager is a mock of proxycfg.Manager that's simpler to control for
-// testing. It also implements ConnectAuthz to allow control over authorization.
-type testManager struct {
```
rboyer, line 25: These moved and were enhanced so that the sotw and delta protocol tests could share more.

agent/xds/resources.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/hashicorp/consul/agent/proxycfg"
+)
+
+// ResourceGenerator is associated with a single gRPC stream and creates xDS
+// resources for a single client.
+type ResourceGenerator struct {
+	Logger         hclog.Logger
+	CheckFetcher   HTTPCheckFetcher
+	CfgFetcher     ConfigFetcher
```
rboyer, line 17: aside: In a future PR I want to move `HTTPCheckFetcher` and `ConfigFetcher` over into `agent/proxycfg` so that the only blob of data we need in `agent/xds` is the `ConfigSnapshot`.

agent/xds/delta.go
```diff
@@ -0,0 +1,728 @@
...
+				node = req.Node
+				var err error
+				g.ProxyFeatures, err = determineSupportedProxyFeatures(req.Node)
+				if err != nil {
+					return status.Errorf(codes.InvalidArgument, err.Error())
+				}
+			}
+
+		case cfgSnap = <-stateCh:
+			newRes, err := g.allResourcesFromSnapshot(cfgSnap)
```
rboyer, line 187: This is the first big divergence from SoTW: instead of delaying xDS construction until send-time, we pre-generate it when the snapshot changes so we can compute versions to use when doing the differential sync.

agent/xds/delta.go
```diff
@@ -0,0 +1,728 @@
...
+			var pendingTypes []string
+			for typeUrl, handler := range handlers {
+				if !handler.registered {
+					continue
+				}
+				if len(handler.pendingUpdates) > 0 {
+					pendingTypes = append(pendingTypes, typeUrl)
+				}
+			}
+			if len(pendingTypes) > 0 {
```
rboyer, line 286: When we compute what is "wrong" in envoy's representation we build a list of clusters, endpoints, routes, and listeners to add/remove/update and send up to 4 responses at once in a particular order so as to not drop traffic.

Until ALL of the outstanding updates are {N,}ACKed by envoy we do not send additional changes yet so we don't get ahead of ourselves and break something. There are likely some subset of these changes that can go out simultaneously but that's tricky to figure out from these opaque blobs so this approach is the simplest for now.

agent/xds/delta.go
```diff
@@ -0,0 +1,728 @@
...
+		}
+	}
+
+	if !hasRelevantUpdates && t.sentToEnvoyOnce {
+		return nil, nil, nil
+	}
+
+	// now turn this into a disco response
+	resp := &envoy_discovery_v3.DeltaDiscoveryResponse{
+		// TODO(rb): consider putting something in SystemVersionInfo?
```
rboyer, line 611: If we send some string in `SystemVersionInfo` envoy will log it. This can help follow along over on the envoy side. When I was debugging something here I stuffed much of the state of the `xDSDeltaType` structure here which helped. Since our updates aren't officially versioned but versioned as the snapshot evolves there really isn't a useful human-readable string I could come up with to pass in general.

freddygv (commented):
Went through all commits except the main delta one, will go over that tomorrow. Left some initial comments inline.

test/integration/connect/envoy/helpers.bash
```diff
@@ -231,6 +231,33 @@ function get_envoy_http_filters {
...
+}
+
+function assert_envoy_dynamic_cluster_exists_once {
+  local HOSTPORT=$1
+  local NAME_PREFIX=$2
+  local EXPECT_SNI=$3
+  BODY="$(get_envoy_dynamic_cluster_once $HOSTPORT $NAME_PREFIX)"
+  [ -n "$BODY" ]
+
+  SNI="$(echo "$BODY" | jq --raw-output ".cluster.transport_socket.typed_config.sni | select(. | startswith(\"${EXPECT_SNI}\"))")"
```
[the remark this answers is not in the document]
rboyer: I didn't need a regular expression.

agent/xds/server.go
```diff
@@ -243,9 +243,14 @@ func (s *Server) process(stream ADSStream, reqCh <-chan *envoy_discovery_v3.Disc
 		},
 	}
 
+	var authCheckFrequency = s.AuthCheckFrequency
+	if authCheckFrequency == 0 {
+		authCheckFrequency = DefaultAuthCheckFrequency
+	}
```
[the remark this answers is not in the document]
rboyer: :+1:

agent/proxycfg/testing.go
```diff
@@ -10,14 +10,15 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/mitchellh/go-testing-interface"
```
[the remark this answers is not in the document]
rboyer: It's the `goimports` setting that separates the imports for the current project from the dependencies of the project.

```
let g:go_fmt_options = {
    \ 'goimports': '-local github.com/hashicorp/consul',
    \ }
```

agent/xds/server_test.go
```diff
@@ -1,130 +1,95 @@
...
 	mgr.DeliverConfig(t, sid, snap)
 
-	assertResponseSent(t, envoy.stream.sendCh, expectClustersJSON(snap, 1, 1))
+	convert := func(v3 *envoy_discovery_v3.DiscoveryResponse) *envoy_api_v2.DiscoveryResponse {
+		v2, err := convertDiscoveryResponseToV2(v3)
+		require.NoError(t, err)
+		return v2
+	}
+	expectClusterResponse := func(v, n uint64) *envoy_api_v2.DiscoveryResponse {
+		return convert(&envoy_discovery_v3.DiscoveryResponse{
```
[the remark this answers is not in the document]
rboyer: It didn't seem useful to fabricate v2 protobufs here when nowhere in the normal operation of the v2 SoTW endpoint do we ever create this protobuf outside of the v3/v2 conversion routines.

I can switch it if you think it's valuable.

agent/xds/server_test.go
```diff
@@ -211,176 +176,200 @@ func TestServer_StreamAggregatedResources_BasicProtocol(t *testing.T) {
...
+				makeTestListener_v2(t, snap, "http2:db"),
+				makeTestListener_v2(t, snap, "tcp:geo-cache"),
+			),
+		})
 	}
-}
 
-func assertResponseSent(t *testing.T, ch chan *envoy_discovery_v3.DiscoveryResponse, wantJSON string) {
-	t.Helper()
+	require.True(t, t.Run("no-rds", func(t *testing.T) {
```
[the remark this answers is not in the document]
rboyer: The normal behavior of `(*testing.T).Run` is to not terminate the enclosing test when they fail. In this case the next subtest depends upon the passing of the first subtest, so by asserting that `Run` succeeded we can avoid spurious test failures of subsequent subtests if the first one fails.

agent/xds/server_test.go
```diff
@@ -211,176 +176,200 @@ func TestServer_StreamAggregatedResources_BasicProtocol(t *testing.T) {
...
+			Resources: makeTestResources_v2(t,
+				makeTestListener_v2(t, snap, "tcp:public_listener"),
+				makeTestListener_v2(t, snap, "http2:db:rds"),
+				makeTestListener_v2(t, snap, "tcp:geo-cache"),
+			),
+		})
+	}
+
+	require.True(t, t.Run("with-rds", func(t *testing.T) {
+		// RESP: listeners (but also a stray update of the other registered types)
```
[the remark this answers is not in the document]
rboyer: I didn't want to change v2 SoTW behavior. I'd rather just wait to delete it wholesale.

agent/xds/server.go
```diff
@@ -194,14 +190,19 @@ func (s *Server) process(stream ADSStream, reqCh <-chan *envoy_discovery_v3.Disc
...
+		cfgSnap     *proxycfg.ConfigSnapshot
+		req         *envoy_discovery_v3.DiscoveryRequest
+		node        *envoy_config_core_v3.Node
+		ok          bool
+		stateCh     <-chan *proxycfg.ConfigSnapshot
+		watchCancel func()
+		proxyID     structs.ServiceID
+	)
+
+	g := newResourceGenerator(
```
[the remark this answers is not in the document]
rboyer: :+1:

agent/xds/server_test.go
```diff
@@ -149,7 +114,7 @@ func TestServer_StreamAggregatedResources_BasicProtocol(t *testing.T) {
 	envoy.SendReq(t, EndpointType, 1, 2)
 
 	// And should get a response immediately.
-	assertResponseSent(t, envoy.stream.sendCh, expectListenerJSON(snap, 1, 3))
+	assertResponseSent(t, envoy.stream.sendCh, expectListenerResponse(1, 3))
 
 	// Now send Route request along with next listener one
 	envoy.SendReq(t, RouteType, 0, 0)
```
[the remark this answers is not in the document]
rboyer: :+1:

agent/xds/server.go
```diff
@@ -129,21 +131,64 @@ type Server struct {
...
+	s := &Server{
+		Logger:       logger,
+		CfgMgr:       cfgMgr,
+		ResolveToken: resolveToken,
+		CheckFetcher: checkFetcher,
+		CfgFetcher:   cfgFetcher,
+	}
+
+	var authCheckFrequency = s.AuthCheckFrequency
+	if authCheckFrequency == 0 {
```
freddygv, line 166: This will always be true; same with L172

agent/xds/delta.go
```diff
@@ -0,0 +1,718 @@
...
+			case structs.ServiceKindConnectProxy:
+			case structs.ServiceKindTerminatingGateway:
+				generator.Logger = generator.Logger.Named(logging.TerminatingGateway)
+			case structs.ServiceKindMeshGateway:
+				generator.Logger = generator.Logger.Named(logging.MeshGateway)
+			case structs.ServiceKindIngressGateway:
+				generator.Logger = generator.Logger.Named(logging.IngressGateway)
+			}
+
+			generator.Logger.Trace("Got initial config snapshot")
```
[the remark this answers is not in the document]
rboyer: The entire `generator.Logger` is reconfigured to always add the `service_id` KV to logs up in `stateDeltaInit` which comes before this.

agent/xds/delta.go
```diff
@@ -0,0 +1,718 @@
...
+
+			generator.logTraceRequest("Incremental xDS v3", req)
+
+			if req.TypeUrl == "" {
+				return status.Errorf(codes.InvalidArgument, "type URL is required for ADS")
+			}
+
+			handler, ok := handlers[req.TypeUrl]
+			if !ok {
+				return nil // not a type we care about
```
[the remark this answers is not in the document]
rboyer: Nice catch. I think this used to be in a function called BY the select/case, rather than embedded IN the select/case.

agent/xds/server_test.go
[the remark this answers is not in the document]
rboyer: Ah the main thing the outer `convert` was doing was swapping in the TypeURL. I did that manually here and got rid of the convert calls.

agent/xds/delta.go
```diff
@@ -0,0 +1,728 @@
...
+				g.ProxyFeatures, err = determineSupportedProxyFeatures(req.Node)
+				if err != nil {
+					return status.Errorf(codes.InvalidArgument, err.Error())
+				}
+			}
+
+		case cfgSnap = <-stateCh:
+			newRes, err := g.allResourcesFromSnapshot(cfgSnap)
+			if err != nil {
+				return err
```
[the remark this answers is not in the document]
rboyer: Did that.

freddygv:
Looks like that 1.14.6 integration test is still failing. That's one we won't support in 1.10 if we move our matrix to start with Envoy 1.18, right?

Also were any of the TODOs in your main PR description for this PR?

rboyer:
> Looks like that 1.14.6 integration test is still failing. That's one we won't support in 1.10 if we move our matrix to start with Envoy 1.18, right?

Apparently it was a flake and rerunning the 1.14.6 one passed this time.

Also if we roll envoy 1.18 into support we'll roll 1.14 out of support

agent/xds/delta_test.go
```diff
@@ -0,0 +1,760 @@
...
+	mgr, errCh, envoy := scenario.mgr, scenario.errCh, scenario.envoy
+
+	sid := structs.NewServiceID("web-sidecar-proxy", nil)
+
+	// Register the proxy to create state needed to Watch() on
+	mgr.RegisterProxy(t, sid)
+
+	snap := newTestSnapshot(t, nil, "")
+
+	// Send initial cluster discover (empty payload)
```
freddygv, line 50: I think this comment is outdated, since the geocache version is sent

agent/xds/delta_test.go
```diff
@@ -0,0 +1,760 @@
...
+	// Envoy then tries to discover endpoints for those clusters.
+	envoy.SendDeltaReq(t, EndpointType, &envoy_discovery_v3.DeltaDiscoveryRequest{
+		// We'll assume we are testing a partial "reconnect"
+		InitialResourceVersions: mustMakeVersionMap(t,
+			makeTestEndpoints(t, snap, "tcp:geo-cache"),
+		),
+		ResourceNamesSubscribe: []string{
+			"db.default.dc1.internal.11111111-2222-3333-4444-555555555555.consul",
+			// "geo-cache.default.dc1.query.11111111-2222-3333-4444-555555555555.consul",
+			// see what happens if you try to subscribe to an unknown thing
```
freddygv, line 83: Was this a debug comment? Seems like we know what happens

agent/xds/delta_test.go
```diff
@@ -0,0 +1,760 @@
...
+			makeTestEndpoints(t, snap, "tcp:geo-cache"),
+		),
+	})
+
+	envoy.SendDeltaReqACK(t, EndpointType, 5, true, nil)
+
+	// And no other response yet
+	assertDeltaChanBlocked(t, envoy.deltaStream.sendCh)
+
+	// TODO: test NACK
```
freddygv, line 164: Is this for this PR?

agent/xds/delta_test.go
```diff
@@ -0,0 +1,760 @@
...
+				makeTestCluster(t, snap, "http2:db"),
+				makeTestCluster(t, snap, "tcp:geo-cache"),
+			),
+		})
+
+		// Envoy then tries to discover endpoints for those clusters.
+		// {
+		//   "typeUrl": "type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment",
+		//   "resourceNamesSubscribe": [
+		//     "api.default.dc1.internal.2902259a-31e7-62e0-fccc-0d482f347e98.consul"
```
freddygv, line 218: This SNI doesn't line up with the subscription below

freddygv (approved):
🚢 🇮🇹

agent/xds/delta_test.go
rboyer: in another one

agent/xds/delta_test.go
rboyer: Fixed.

agent/xds/delta_test.go
rboyer: I made it more obvious that this comment is about why "fake-service" shows up.

agent/xds/delta_test.go
rboyer: Removed the completely stale xDS json blobs from this.

agent/xds/envoy_versioning.go
```diff
@@ -28,6 +29,8 @@ type supportedProxyFeatures struct {
 	// add version dependent feature flags here
 
 	GatewaysNeedStubClusterWhenEmptyWithIncrementalXDS bool
```
freddygv, line 31: I think it would be useful to also add comments here for why these are used

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/359719.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 71d45a34601423abdfc0a64d44c6a55cf88fa2fc onto `release/1.10.x` succeeded!

PR #9859, opened 2021-03-10
Docs for helm values introduced by metrics

thisisnotashwin:
These were autogenerated by `make gen-docs` on `consul-helm`

### DIFF
--- a/website/content/docs/k8s/helm.mdx
+++ b/website/content/docs/k8s/helm.mdx
@@ -216,6 +216,24 @@ and consider if they're appropriate for your deployment.
       `<global.name>-federation` (if setting `global.name`), otherwise
       `<helm-release-name>-consul-federation`. Requires consul-k8s 0.15.0+.
 
+  - `metrics` ((#v-global-metrics)) - Configures metrics for Consul service mesh
+
+    - `enabled` ((#v-global-metrics-enabled)) (`boolean: false`) - Configures the Helm chart’s components
+      to expose Prometheus metrics for the Consul service mesh. By default
+      this includes gateway metrics and sidecar metrics.
+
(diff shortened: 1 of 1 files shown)


website/content/docs/k8s/helm.mdx
```diff
@@ -216,6 +216,24 @@ and consider if they're appropriate for your deployment.
...
+      global.metrics.enabled is true.
+
+    - `agentMetricsRetentionTime` ((#v-global-metrics-agentmetricsretentiontime)) (`string: 1m`) - Configures the retention time for metrics in Consul clients and
+      servers. This must be greater than 0 for Consul clients and servers
+      to expose any metrics at all.
+      Only applicable if global.metrics.enabled is true.
+
+    - `enableGatewayMetrics` ((#v-global-metrics-enablegatewaymetrics)) (`boolean: true`) - If true, mesh, terminating, and ingress gateways will expose their
+      Envoy metrics on port 20200 at the /metrics path and all gateway pods
+      will have Prometheus scrape annotations. Only applicable if global.metrics.enabled is true.
```
ndhanushkodi, line 235:
```suggestion
    - `enableGatewayMetrics` ((#v-global-metrics-enablegatewaymetrics)) (`boolean: true`) - If true, mesh, terminating, and ingress gateways will expose their
      Envoy metrics on port `20200` at the `/metrics` path and all gateway pods
      will have Prometheus scrape annotations. Only applicable if `global.metrics.enabled` is true.
```

I know this change would have to be made in the helm values :( but I think the docs render nicer with the formatting.

ndhanushkodi: There's a few other places in the values.yaml where we don't have the ticks for formatting as code too, so maybe we could do all those in one PR?

PR #9864, opened 2021-03-10
ui: Adds warning icon to side menu when ACLs are disabled

johncowen:
Shows a warning icon next to the Access Controls title of the side nav when ACLs are disabled to give an always on reminder that they are not enabled. We also show related unlinked navigation items which are visible but not clickable (Roles, Policies, Auth Methods). Tokens is still linked to give an area for us to provide further info.

Screenshot 2021-03-10 at 18 47 58

When ACLs are enabled there is no change to how it was previously.

Worthwhile noting that we added a `(can "use acls")` which should be used within templates moving forwards rather then `(env "CONSUL_ACLS_ENABLED")` - we should try to gradually phase out using the `env` helper from within templates, instead favouring `can` for permission/abilities and `t` for text/copy/URLs.

Both `can` and `t` helpers are used widely in other product UIs whereas `env` is not.

### DIFF
--- a/.changelog/9864.txt
+++ b/.changelog/9864.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: add permanently visible indicator when ACLs are disabled
+```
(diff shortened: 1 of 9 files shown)


kaxcode (approved):
LGTM

PR #9872, opened 2021-03-11
Create new types for service-defaults upstream cfg

freddygv:
This PR expands service-defaults such that most upstream configuration can be configured centrally. There is a map called upstream_configs keyed on `<optional-namespace/>service-name` which provides per-upstream configuration. There is also an `upstream_defaults` object for providing configuration that applies across all upstreams of the given service.

The fields in both of these were promoted out of  the opaque config map in `Upstream` so that they can be validated on config write. Note that calls to `ResolveServiceConfig` will return them in the existing opaque map so that clients on older versions can receive the new centrally-configured fields.

Below is an example config entry:

```
kind = "service-defaults"
name = "web"

connect {
  # These apply to the upstream key and override the value from upstream_defaults
  upstream_configs {
    "redis" {
      passive_health_check {
        max_failures = 3
      }
    }

    "finance/billing" {
      mesh_gateway {
        mode = "remote"
      }
    }
  }
  
  # These defaults apply across all upstreams of web
  upstream_defaults {
    connect_timeout_ms = "5s"
  }
}
```

In general this is the merge order for something like "protocol" for an upstream listener:
- `proxy-defaults.Config["protocol"]`
- `service-defaults` of the upstream
- `service-defaults.upstream_defaults` of the downstream
- `service-defaults.upstream_configs` of the downstream for the upstream
- Local upstream configuration (`Proxy.Upstreams[i].Config["protocol"]`)

--------------

TODO: 
- [x] Fixup some tests 
- [ ] Docs changes
- [ ] Changelog


agent/consul/config_endpoint.go
```diff
@@ -389,55 +397,121 @@ func (c *ConfigEntry) ResolveServiceConfig(args *structs.ServiceConfigRequest, r
...
+
+			// The goal is to flatten the mesh gateway mode in this order:
+			// 	0. Value from centralized upstream_defaults
+			// 	1. Value from local proxy registration
+			// 	2. Value from centralized upstream_configs
+			// 	3. Value from local upstream definition. This last step is done in the client's service manager.
+			var registrationMGConfig structs.MeshGatewayConfig
+
+			if args.ID != "" && args.NodeName != "" {
+				index, registration, err := state.NodeServiceWatch(ws, args.NodeName, args.ID, &args.EnterpriseMeta)
```
rboyer, line 456: This is not going to do what you expect. If the ultimate goal is to figure out what mesh gateway mode the user originally used in their explicit agent-service-register attempt, on the FIRST register this NodeServiceWatch is going to return `nil` because the catalog hasn't even seen the data yet:

order of operation:

1. user invokes agent service register
2. agent intercepts the agent/local state write (and mem write) on the client agent and uses service manager to pull resolved config data from the server via that RPC
3. when that RPC returns the agent smooshes the data from (1) with the data from (2) to make the REAL local service registration data
4. the intercept from (2) returns as if the smooshed data is what someone wrote into their operation in (1)
5. the agent writes this to agent/local state
6. time passes
7. agent does AE sync of the smooshed data only

The Servers never see what the user originally stated in (1). ServiceManager is a crazy man-in-the-middle attack on registration if that helps think about it.

The easier thing would be to pass the local mesh-gateway mode for the proxy up on the config merging RPC explicitly rather than trying to extract it from the state machine (where it doesn't exist yet).

api/config_entry.go
```diff
@@ -91,15 +91,91 @@ type ExposePath struct {
...
+	PassiveHealthCheck *PassiveHealthCheck `json:",omitempty" alias:"passive_health_check"`
+
+	// MeshGatewayConfig controls how Mesh Gateways are configured and used
+	MeshGateway MeshGatewayConfig `json:",omitempty" alias:"mesh_gateway" `
+}
+
+type PassiveHealthCheck struct {
+	// Interval between health check analysis sweeps. Each sweep may remove
+	// hosts or return hosts to the pool.
+	Interval time.Duration
```
rboyer, line 143: omitempty?

api/config_entry.go
```diff
@@ -91,15 +91,91 @@ type ExposePath struct {
...
+	// routing etc.
+	Protocol string
+
+	// ConnectTimeoutMs is the number of milliseconds to timeout making a new
+	// connection to this upstream. Defaults to 5000 (5 seconds) if not set.
+	ConnectTimeoutMs int `alias:"connect_timeout_ms"`
+
+	// Limits are the set of limits that are applied to the proxy for a specific upstream of a
+	// service instance.
+	Limits *UpstreamLimits
```
rboyer, line 130: omitempty?

api/config_entry.go
```diff
@@ -91,15 +91,91 @@ type ExposePath struct {
...
+	//
+	// Note: This escape hatch is NOT compatible with the discovery chain and
+	// will be ignored if a discovery chain is active.
+	ClusterJSON string `alias:"cluster_json"`
+
+	// Protocol describes the upstream's service protocol. Valid values are "tcp",
+	// "http" and "grpc". Anything else is treated as tcp. The enables protocol
+	// aware features like per-request metrics and connection pooling, tracing,
+	// routing etc.
+	Protocol string
```
rboyer, line 122: omitempty?

api/config_entry.go
```diff
@@ -91,15 +91,91 @@ type ExposePath struct {
...
+
+	// ConnectTimeoutMs is the number of milliseconds to timeout making a new
+	// connection to this upstream. Defaults to 5000 (5 seconds) if not set.
+	ConnectTimeoutMs int `alias:"connect_timeout_ms"`
```
rboyer, line 126: omitempty?

api/config_entry.go
```diff
@@ -91,15 +91,91 @@ type ExposePath struct {
...
+	// will be ignored if a discovery chain is active.
+	ListenerJSON string `json:",omitempty" alias:"listener_json"`
+
+	// ClusterJSON is a complete override ("escape hatch") for the upstream's
+	// cluster. The Connect client TLS certificate and context will be injected
+	// overriding any TLS settings present.
+	//
+	// Note: This escape hatch is NOT compatible with the discovery chain and
+	// will be ignored if a discovery chain is active.
+	ClusterJSON string `alias:"cluster_json"`
```
rboyer, line 116: omitempty?

api/config_entry.go
```diff
@@ -91,15 +91,91 @@ type ExposePath struct {
...
+	UpstreamDefaults UpstreamConfig `json:",omitempty" alias:"upstream_defaults"`
+}
+
+type UpstreamConfig struct {
+	// ListenerJSON is a complete override ("escape hatch") for the upstream's
+	// listener.
+	//
+	// Note: This escape hatch is NOT compatible with the discovery chain and
+	// will be ignored if a discovery chain is active.
+	ListenerJSON string `json:",omitempty" alias:"listener_json"`
```
rboyer, line 108: omitempty?

agent/structs/config_entry.go
```diff
@@ -592,13 +633,210 @@ func (r *ServiceConfigRequest) CacheInfo() cache.RequestInfo {
...
+func (chk PassiveHealthCheck) Validate() error {
+	if chk.Interval <= 0*time.Second {
+		return fmt.Errorf("passive health check interval must be greater than 0s")
+	}
+	return nil
+}
+
+// UpstreamLimits describes the limits that are associated with a specific
+// upstream of a service instance.
+type UpstreamLimits struct {
```
rboyer, line 797: should these all have `json:",omitempty"`?

agent/structs/config_entry.go
```diff
@@ -592,13 +633,210 @@ func (r *ServiceConfigRequest) CacheInfo() cache.RequestInfo {
...
+func ParseUpstreamConfig(m map[string]interface{}) (UpstreamConfig, error) {
+	cfg, err := ParseUpstreamConfigNoDefaults(m)
+
+	// Set defaults (even if error is returned)
+	cfg.Normalize()
+
+	return cfg, err
+}
+
+type PassiveHealthCheck struct {
```
rboyer, line 773: should these all have `json:",omitempty"`?

agent/structs/config_entry.go
```diff
@@ -592,13 +633,210 @@ func (r *ServiceConfigRequest) CacheInfo() cache.RequestInfo {
...
+		dst["limits"] = cfg.Limits
+	}
+	if cfg.PassiveHealthCheck != nil {
+		dst["passive_health_check"] = cfg.PassiveHealthCheck
+	}
+}
+
+func (cfg *UpstreamConfig) Normalize() {
+	if cfg.Protocol == "" {
+		cfg.Protocol = "tcp"
```
rboyer, line 710: normalizing this is going to hurt your mergability

agent/structs/config_entry.go
```diff
@@ -592,13 +633,210 @@ func (r *ServiceConfigRequest) CacheInfo() cache.RequestInfo {
 }
 
 type UpstreamConfig struct {
```
rboyer, line 635: should these all have `json:",omitempty"`?

rboyer (changes requested):
I think if you can rearrange things so the config flattening RPC doesn't need to know about nodes that'll be a big improvement. The only other substantial concern here was about some missing `api` and `structs` json `omitempty` struct tags and some possible merging problems down the line.

agent/agent.go
```diff
@@ -1948,6 +1948,7 @@ type addServiceLockedRequest struct {
 // agent using Agent.AddService.
 type AddServiceRequest struct {
 	Service               *structs.NodeService
+	nodeName              string
```
rboyer, line 1951: Probably don't need any of these nodeName plumbing changes now.

agent/consul/config_endpoint.go
```diff
@@ -329,31 +329,21 @@ func (c *ConfigEntry) ResolveServiceConfig(args *structs.ServiceConfigRequest, r
...
-				if !ok {
-					return fmt.Errorf("invalid service config type %T", serviceEntry)
-				}
-			}
 
-			// Use the default enterprise meta to look up the global proxy defaults. In the future we may allow per-namespace proxy-defaults
-			// but not yet.
+			// Pass the WatchSet to both the service and proxy config lookups. If either is updated during the
+			// blocking query, this function will be rerun and these state store lookups will both be current.
+			// We use the default enterprise meta to look up the global proxy defaults because their are not namespaced.
```
rboyer, line 336:
```suggestion
			// We use the default enterprise meta to look up the global proxy defaults because they are not namespaced.
```

agent/consul/config_endpoint.go
```diff
@@ -367,11 +357,29 @@ func (c *ConfigEntry) ResolveServiceConfig(args *structs.ServiceConfigRequest, r
...
+
+				// Extract the global protocol from proxyConf for upstream configs.
+				rawProtocol := proxyConf.Config["protocol"]
+				if rawProtocol != nil {
+					proxyConfGlobalProtocol, ok = rawProtocol.(string)
+					if !ok {
+						return fmt.Errorf("invalid protocol type %T", rawProtocol)
+					}
+				}
 			}
```
rboyer, line 369: If `proxyConfGlobalProtocol` is empty, should you initialize it to `tcp` here? Or is it totally fine for the protocol to end up as emptystring?

agent/consul/state/catalog.go
```diff
@@ -1136,6 +1136,24 @@ func (s *Store) NodeService(nodeName string, serviceID string, entMeta *structs.
 	return idx, service, nil
 }
 
+// NodeServiceWatch is used to retrieve a specific service associated with the given
+// node, and add it to the watch set.
+func (s *Store) NodeServiceWatch(ws memdb.WatchSet, nodeName string, serviceID string, entMeta *structs.EnterpriseMeta) (uint64, *structs.NodeService, error) {
```
rboyer, line 1141: You shouldn't need this anymore.

agent/consul/state/catalog.go
```diff
@@ -1150,6 +1168,21 @@ func getNodeServiceTxn(tx ReadTxn, nodeName, serviceID string, entMeta *structs.
 	return nil, nil
 }
 
+func getNodeServiceWatchTxn(tx ReadTxn, ws memdb.WatchSet, nodeName, serviceID string, entMeta *structs.EnterpriseMeta) (*structs.NodeService, error) {
```
rboyer, line 1171: or this

agent/service_manager.go
```diff
@@ -335,6 +338,7 @@ func makeConfigRequest(bd BaseDeps, addReq AddServiceRequest) *structs.ServiceCo
 		Name:           name,
 		Datacenter:     bd.RuntimeConfig.Datacenter,
 		QueryOptions:   structs.QueryOptions{Token: addReq.token},
+		MeshGateway:    ns.Proxy.MeshGateway,
```
rboyer, line 341: Before GA we should figure out if these changes on both sides of the wire for this RPC strongly implies an upgrade order when upgrading to 1.10. I think it's probably correct to do servers before clients given how this was updated, but it's worth writing that down.

rboyer (approved):
LGTM

agent/service_manager.go
freddygv: Our standard upgrade process already recommends upgrading servers before clients:
https://www.consul.io/docs/upgrading#standard-upgrades

Do you mean something more strongly worded?

agent/consul/config_endpoint.go
freddygv: That should be OK. If the proxy-defaults protocol gets overwritten by the higher precedence config like the upstream's service-defaults. If all of them are empty then the protocol key isn't added to the resolved opaque config map (which is the current behavior).

agent/service_manager.go
rboyer: Right. Sorry I was mixing up the servers/clients and primary-dc/secondary-dc upgrade order. The former is specified and the latter is not.

rboyer (approved):

markan (commented):
One minor suggestion around readability, but otherwise this looks good.

agent/consul/config_endpoint.go
```diff
@@ -329,31 +329,21 @@ func (c *ConfigEntry) ResolveServiceConfig(args *structs.ServiceConfigRequest, r
 		&reply.QueryMeta,
 		func(ws memdb.WatchSet, state *state.Store) error {
```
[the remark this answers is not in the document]
freddygv: That is a good point. I added a TODO here that I can take care of for the beta.

PR #9873, opened 2021-03-11
Add TransparentProxy opt to proxy definition

freddygv:
Branched off #9872

This PR adds a flag for TransparentProxy mode to: proxy-defaults, service-defaults, and proxy registrations.

The main bit of logic here is changing is the service manager's config merging. In TransparentProxy mode we don't expect explicitly defined upstreams, so we extend our local list of Upstreams with objects generated from central config. That is so centrally-configured upstream settings like circuit breakers can be applied to the appropriate Envoy filter chains.

Here are example config entries with the flag:

```
kind = "service-defaults"
name = "web"

Protocol = "http"
TransparentProxy = true
}
```

```
kind = "proxy-defaults"
name = "global"

TransparentProxy = true
}
```

--------------

TODO: 
- [ ] Docs changes


agent/service_manager.go
```diff
@@ -379,39 +379,76 @@ func mergeServiceConfig(defaults *structs.ServiceConfigResponse, service *struct
...
 		ns.Proxy.MeshGateway.Mode = defaults.MeshGateway.Mode
 	}
+	if !ns.Proxy.TransparentProxy {
+		ns.Proxy.TransparentProxy = defaults.TransparentProxy
+	}
+
+	// seenUpstreams stores the upstreams seen from the local registration so that we can also add synthetic entries.
+	// for upstream configuration that was defined via service-defaults.UpstreamConfigs. In TransparentProxy mode
+	// ns.Proxy.Upstreams will likely be empty because users do not need to define upstreams explicitly.
+	// So to store upstream-specific flags from central config, we add entries to ns.Proxy.Upstream with thosee values.
```
rboyer, line 389:
```suggestion
	// So to store upstream-specific flags from central config, we add entries to ns.Proxy.Upstream with those values.
```

agent/service_manager.go
```diff
@@ -379,39 +379,76 @@ func mergeServiceConfig(defaults *structs.ServiceConfigResponse, service *struct
 	if ns.Proxy.MeshGateway.Mode == structs.MeshGatewayModeDefault {
 		ns.Proxy.MeshGateway.Mode = defaults.MeshGateway.Mode
 	}
+	if !ns.Proxy.TransparentProxy {
+		ns.Proxy.TransparentProxy = defaults.TransparentProxy
+	}
+
+	// seenUpstreams stores the upstreams seen from the local registration so that we can also add synthetic entries.
```
rboyer, line 386:
```suggestion
	// remoteUpstreams stores the upstreams seen from the local registration so that we can also add synthetic entries.
```

rboyer: Hmm, the variable used to be called `seenUpstreams`, but below it is `remoteUpstreams`. The comment here though leans in the direction of this probably being a local thing not a remote thing so I'm unclear which it's supposed to be.

agent/service_manager.go
```diff
@@ -379,39 +379,76 @@ func mergeServiceConfig(defaults *structs.ServiceConfigResponse, service *struct
...
+	for _, us := range defaults.UpstreamIDConfigs {
+		parsed, err := structs.ParseUpstreamConfig(us.Config)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse upstream config map for %s: %v", us.Upstream.String(), err)
+		}
 
-	// Merge upstream defaults if there were any returned
+		// Delete the mesh gateway key since this is the only place it is read from an opaque map.
+		// Later reads use Proxy.MeshGateway.
+		delete(us.Config, "mesh_gateway")
```
rboyer, line 400: Would you also need to delete variations of this (camel, UPPER, etc) or is that safe to ignore since this is coming straight back from the RPC endpoint?

agent/service_manager.go
rboyer: Should this comment actually move down to the `localUpstreams` variable below?

agent/service_manager.go
```diff
@@ -379,39 +379,76 @@ func mergeServiceConfig(defaults *structs.ServiceConfigResponse, service *struct
...
+
+		remoteUpstreams[us.Upstream] = structs.Upstream{
+			DestinationNamespace: us.Upstream.NamespaceOrDefault(),
+			DestinationName:      us.Upstream.ID,
+			Config:               us.Config,
+			MeshGateway:          parsed.MeshGateway,
```
rboyer, line 406: Did you mean to lift the transparentproxy flag here as well?

agent/structs/connect_proxy_config.go
```diff
@@ -256,6 +265,10 @@ type Upstream struct {
 	// an ingress gateway. This cannot and should not be set by a user, it is
 	// used internally to store the association of hosts to an upstream service.
 	IngressHosts []string `json:"-" bexpr:"-"`
+
+	// CentrallyConfigured indicates whether the upstream was defined in a proxy
+	// instance registration or whether it was generated from a config entry.
+	CentrallyConfigured bool
```
rboyer, line 271: Does this need a `json:",omitempty"` so various catalog results don't start emitting a lot of `false` values immediately?

Should you exclude this from `bexpr` filtering?

rboyer (changes requested):
Found some minor things, especially when tproxy mode is not enabled.

antonio490102 (approved):

agent/service_manager.go
freddygv: Safe to ignore for the reason you just mentioned. That RPC endpoint and `UpstreamConfig.MergeInto()` are the only places that populate mesh gateway config in the opaque map.

Added a comment about that.

agent/service_manager.go
freddygv: Nope, the TransparentProxy flag on an upstream definition won't do anything. It only kicks in as a property of the downstream proxy's registration.

agent/structs/connect_proxy_config.go
freddygv: Yes and yes

agent/service_manager.go
```diff
@@ -379,39 +379,76 @@ func mergeServiceConfig(defaults *structs.ServiceConfigResponse, service *struct
...
-		if err := mergo.Merge(&us.Config, usCfg); err != nil {
+		if err := mergo.Merge(&us.Config, usCfg.Config); err != nil {
 			return nil, err
 		}
 	}
+
+	// Ensure upstreams present in central config are represented in the local configuration.
+	// This does not apply outside of TransparentProxy mode because in that situation every upstream needs to be defined
+	// explicitly and locally with a local bind port.
+	if ns.Proxy.TransparentProxy {
```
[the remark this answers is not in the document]
freddygv: If I understand what you're saying, this validation should handle that:
https://github.com/hashicorp/consul/pull/9873/files#diff-568e9f391e90a5291bcdc97b624a0da95b20a95ffc0edd1966cae6428d855e13R328

`LocalBindPort` must always be populated unless the Upstream is CentrallyConfigured, and that flag is only set in this function.

agent/service_manager.go
freddygv: Yep, I must have moved that map but missed moving the comment.

rboyer (approved):
LGTM

PR #9875, opened 2021-03-12
ui: Improves UI engineering docs

johncowen:
This PR iterates on our engineering docs for the UI.

1. Tweaks existing READMEs to use `preview-template` for live previews of components based on the snippets.
2. Adds glimmer or ember icons to each component in the nav bar so we can easily see what is left to be upgraded to glimmer.
3. Adds some more general documentation around the UI

We also added two helpers here:

1. `to-route` this uses `router.recognize` to convert URLs (generated by docfy) to routeNames so we can then pass this to our existing `is-href` helper.
2. `flatten-property` specific helper for helping us to flatten out the nested menu structure from docfy. docfy does have a `flat` option but we need a mix between the two.

Lastly whilst working through docs we spotted a couple of things we had to work around/fixup.

1. We disabled our 'select but don't fire click event' initializer when you are viewing docs.
2. We spotted a couple of typos in our `with-overlay` modifier, so we fixed that up here also.

Screenshot 2021-03-12 at 13 13 00


kaxcode (approved):
Love the use of the glimmer and ember icons  👍

PR #9877, opened 2021-03-15
Add RPC endpoint to compute upstreams from intentions

freddygv:
This PR adds an RPC and backing state store function that returns the upstreams of a service. Upstreams and downstreams here are inferred from intentions and do not consider proxy registrations.

If intentions allow a connection from the target to some candidate service, the candidate service is considered an upstream of the target.

The current implementation is a quick/naive pass for the alpha release:
1. Watch all service names
2. Determine whether a connection would be allowed between the two.

Optimizations will be added later after profiling. This will also need to be generalized into being able to compute both upstreams and downstreams, since that is needed for the service mesh visualization. Those changes are targeted for the beta.


agent/consul/state/catalog.go
```diff
@@ -744,7 +745,11 @@ func serviceListTxn(tx ReadTxn, ws memdb.WatchSet, entMeta *structs.EnterpriseMe
 	unique := make(map[structs.ServiceName]struct{})
 	for service := services.Next(); service != nil; service = services.Next() {
 		svc := service.(*structs.ServiceNode)
-		unique[svc.CompoundServiceName()] = struct{}{}
+		// TODO (freddy) This is a hack to exclude certain kinds.
+		//				 Need a new index to query by kind and namespace, have to coordinate with consul foundations first
+		if include != nil && include(svc) {
```
rboyer, line 750:
```suggestion
		if include == nil || (include != nil && include(svc)) {
```

Make sure this still works with no filter at all.

agent/consul/state/intention.go
```diff
@@ -936,3 +950,88 @@ func intentionMatchGetParams(entry structs.IntentionMatchEntry) ([][]interface{}
...
+	// That way we can check whether intentions from the source allow connections to upstream candidates.
+	matchType := structs.IntentionMatchSource
+	if downstreams {
+		matchType = structs.IntentionMatchDestination
+	}
+	entry := structs.IntentionMatchEntry{
+		Namespace: target.NamespaceOrDefault(),
+		Name:      target.Name,
+	}
+	index, intentions, err := s.IntentionMatchOne(ws, entry, matchType)
```
dnephin, line 971: same here, I imagine all the calls in this method would share the same transaction, right?

agent/consul/state/catalog.go
```diff
@@ -2859,16 +2864,20 @@ func (s *Store) ServiceTopology(
...
 
-	// The given service is the source relative to upstreams
-	sourceURI := connect.SpiffeIDService{
+	matchEntry := structs.IntentionMatchEntry{
 		Namespace: entMeta.NamespaceOrDefault(),
-		Service:   service,
+		Name:      service,
+	}
+	// The given service is a source relative to its upstreams
+	_, intentions, err := s.IntentionMatchOne(ws, matchEntry, structs.IntentionMatchSource)
```
dnephin, line 2872: Should this be calling a function that accepts the active transaction instead of creating a new transaction?

Same question about the call to `IntentionDecision` just below this.

agent/consul/intention_endpoint.go
```diff
@@ -684,19 +683,6 @@ func (s *Intention) Check(args *structs.IntentionQueryRequest, reply *structs.In
 		return fmt.Errorf("Invalid destination namespace %q: %v", query.DestinationNS, err)
 	}
 
-	// Build the URI
-	var uri connect.CertURI
-	switch query.SourceType {
-	case structs.IntentionSourceConsul:
```
rboyer, line 690: I think you need to retain at a bare minimum this part:
```
if query.SourceType != structs.IntentionSourceConsul {
  return ...
}
```

agent/consul/state/intention.go
```diff
@@ -792,6 +775,37 @@ func (s *Store) IntentionDecision(
 	return resp, nil
 }
 
+// AuthorizeIntentionTarget determines whether the destination is covered by the given intention
+// and whether the intention action allows a connection.
+func AuthorizeIntentionTarget(target, targetNS string, ixn *structs.Intention, matchType structs.IntentionMatchType) (bool, bool) {
```
rboyer, line 780: May want to note that this logic formerly was in `(agent/connect).SpiffeIDService.Authorize`. Is the former method used anymore in mainline consul+envoy code?

agent/consul/state/intention.go
```diff
@@ -778,7 +761,7 @@ func (s *Store) IntentionDecision(
 	if len(ixnMatch.Permissions) > 0 {
 		// If there are L7 permissions, DENY.
```
rboyer, line 762: This comment is now incorrect since it's selectable behavior in this PR.

agent/consul/helper_test.go
```diff
@@ -910,3 +910,138 @@ func registerTestTopologyEntries(t *testing.T, codec rpc.ClientCodec, token stri
 		require.NoError(t, msgpackrpc.CallWithCodec(codec, "ConfigEntry.Apply", &req, &out))
 	}
 }
+
+func registetIntentionUpstreamEntries(t *testing.T, codec rpc.ClientCodec, token string) {
```
rboyer, line 914:
```suggestion
func registerIntentionUpstreamEntries(t *testing.T, codec rpc.ClientCodec, token string) {
```

agent/consul/internal_endpoint.go
```diff
@@ -188,6 +188,49 @@ func (m *Internal) ServiceTopology(args *structs.ServiceSpecificRequest, reply *
 		})
 }
 
+// IntentionUpstreams returns the upstreams or downstreams of a service. Upstreams and downstreams are inferred from intentions.
```
rboyer, line 191: > IntentionUpstreams returns the upstreams or downstreams

Why does it return both if it's named for only upstreams?

agent/consul/state/catalog.go
```diff
@@ -2888,17 +2897,17 @@ func (s *Store) ServiceTopology(
 		maxIdx = idx
 	}
 
+	// The given service is a destination relative to its downstreams
+	_, intentions, err = s.IntentionMatchOne(ws, matchEntry, structs.IntentionMatchDestination)
```
rboyer, line 2901: I think to avoid an accident the `intentions` variable should be renamed here so the source and destination variants are kept separate to avoid an accident.

agent/consul/state/catalog.go
rboyer, line 2872: need to pass `tx` through here

agent/consul/state/catalog.go
```diff
@@ -2859,16 +2864,20 @@ func (s *Store) ServiceTopology(
...
+		Name:      service,
+	}
+	// The given service is a source relative to its upstreams
+	_, intentions, err := s.IntentionMatchOne(ws, matchEntry, structs.IntentionMatchSource)
+	if err != nil {
+		return 0, nil, fmt.Errorf("failed to query intentions for %s", sn.String())
 	}
 	for _, un := range upstreamNames {
-		decision, err := s.IntentionDecision(&sourceURI, un.Name, un.NamespaceOrDefault(), defaultAllow)
+		decision, err := s.IntentionDecision(un.Name, un.NamespaceOrDefault(), intentions, structs.IntentionMatchDestination, defaultAllow, false)
```
rboyer, line 2877: need to pass `tx` through here

agent/consul/state/catalog.go
rboyer, line 2901: need to pass `tx` through here

agent/consul/state/catalog.go
```diff
@@ -2888,17 +2897,17 @@ func (s *Store) ServiceTopology(
...
+	}
 	downstreamDecisions := make(map[string]structs.IntentionDecisionSummary)
 	for _, dn := range downstreamNames {
-		// Downstreams are the source relative to the given service
-		sourceURI := connect.SpiffeIDService{
-			Namespace: dn.NamespaceOrDefault(),
-			Service:   dn.Name,
-		}
-		decision, err := s.IntentionDecision(&sourceURI, service, entMeta.NamespaceOrDefault(), defaultAllow)
+		decision, err := s.IntentionDecision(dn.Name, dn.NamespaceOrDefault(), intentions, structs.IntentionMatchSource, defaultAllow, false)
```
rboyer, line 2907: need to pass `tx` through here

agent/consul/state/intention.go
```diff
@@ -936,3 +950,88 @@ func intentionMatchGetParams(entry structs.IntentionMatchEntry) ([][]interface{}
 	result = append(result, []interface{}{entry.Namespace, entry.Name})
 	return result, nil
 }
+
+// IntentionTopology returns the upstreams or downstreams of a service. Upstreams and downstreams are inferred from
+// intentions. If intentions allow a connection from the target to some candidate service, the candidate service is considered
+// an upstream of the target.
+func (s *Store) IntentionTopology(ws memdb.WatchSet,
```
rboyer, line 957: Need to start a transaction here

agent/consul/state/intention.go
rboyer, line 971: need to pass `tx` through here

agent/consul/state/intention.go
```diff
@@ -936,3 +950,88 @@ func intentionMatchGetParams(entry structs.IntentionMatchEntry) ([][]interface{}
...
+		//				 Maybe ixn.HasWildcardSource() && ixn.HasWildcardDestination()
+		if ixn.SourceName == structs.WildcardSpecifier && ixn.DestinationName == structs.WildcardSpecifier {
+			defaultDecision = acl.Allow
+			if ixn.Action == structs.IntentionActionDeny {
+				defaultDecision = acl.Deny
+			}
+		}
+	}
+
+	index, allServices, err := s.ServiceList(ws, func(svc *structs.ServiceNode) bool {
```
rboyer, line 1003: need to pass `tx` through here

agent/consul/state/intention.go
```diff
@@ -936,3 +950,88 @@ func intentionMatchGetParams(entry structs.IntentionMatchEntry) ([][]interface{}
...
+	if err != nil {
+		return index, nil, fmt.Errorf("failed to fetch catalog service list: %v", err)
+	}
+	if index > maxIdx {
+		maxIdx = index
+	}
+
+	result := make(structs.ServiceList, 0, len(allServices))
+	for _, candidate := range allServices {
+		decision, err := s.IntentionDecision(candidate.Name, candidate.NamespaceOrDefault(), intentions, matchType, defaultDecision, true)
```
rboyer, line 1022: need to pass `tx` through here

agent/consul/internal_endpoint.go
freddygv: Fixed. It will return both in the future, but I updated the comment to reflect its current state.

agent/consul/state/intention.go
freddygv: It was still being used in `Agent.ConnectAuthorize`. I don't really see the need to keep the old Authorize method, nor to keep it in the CertURI interface, so I removed it in favor of the new one.

agent/consul/state/intention.go
```diff
@@ -853,6 +837,16 @@ func (s *Store) IntentionMatchOne(
 	tx := s.db.Txn(false)
 	defer tx.Abort()
 
+	return compatIntentionMatchOneTxn(tx, ws, entry, matchType)
+}
+
+func compatIntentionMatchOneTxn(
```
rboyer, line 843: Why prefix this with `compat` if you're already suffixing it with `Txn`?

agent/consul/state/intention.go
```diff
@@ -936,3 +930,91 @@ func intentionMatchGetParams(entry structs.IntentionMatchEntry) ([][]interface{}
...
+	}
+	if index > maxIdx {
+		maxIdx = index
+	}
+
+	// Reset the matchType since next it is used for evaluating the upstreams or downstreams against a set of intentions.
+	// When evaluating upstreams, the match type is now destination because we are evaluating upstream candidates
+	// as eligible destinations for intentions that have the target service as a source.
+	// The reverse is true for downstreams.
+	matchType = structs.IntentionMatchDestination
```
rboyer, line 966: I wonder if this would be clearer if up front you just declared two vars like `matchTypeUp` and `matchTypeDown` and only did the `if downstreams { ... } else { ... }` conditional once rather than mutating a variable midway through.

agent/consul/state/intention.go
rboyer: Still need to do this.

agent/consul/state/intention.go
rboyer: Still need to do this.

agent/consul/state/catalog.go
freddygv: Yep that's right. Fixed that.

IntentionDecision doesn't need one because it doesn't query the state store anymore.

agent/consul/state/intention.go
freddygv: There's already an `intentionMatchOneTxn`. I first thought to rename that one to `legacyIntentionMatchOneTxn` since it does legacy things, but that also exists, and calls the former. So I named this one compat because it has the config/legacy split in it.

What do you think?

agent/consul/state/intention.go
freddygv: Forgot to comment, but I don't think that's needed. IntentionDecision no longer queries the state store now that it takes in a set of intentions.

agent/consul/state/intention.go
freddygv: Fixed.

agent/consul/state/intention.go
freddygv: I didn't fully understand what you were recommending. I ended up renaming these so that the matchType is prefixed with the intended use.

agent/consul/state/intention.go
rboyer: :+1:

rboyer (approved):
LGTM

PR #9452, opened 2020-12-22
config: make TestBuilder_BuildAndValidate_ConfigFlagsAndEdgecases easier to work with

dnephin:
Branched from #9448

Lots of context in the commits messages.

The main changes here are:

1. reduce the number of fields in the `testCase` struct, and rename others so their purpose is more obvious from the name
2. reduce the test logic to separate iteration from the actual test logic
3. replace the slice with individual calls to `run()` so that when a case fails, the failure output includes a line for both the test function and the testCase. Without this change only the line number of the test function is printed, which makes it extremely difficult to find the relevant testCase data, because there are thousands of lines with very similar names.

An alternative to 3 would be to split this one case into a few hundred test cases. That would likely be significantly more work for roughly the same gain. This change was made with find/replace, which made it much easier to apply.

Probably best viewed by commit, with whitespace off.

PR #9880, opened 2021-03-15
state: use runCase pattern for large test

dnephin:
Best viewed with whitespace off. This same pattern was used in #9452 for a large `agent/config` test.

This pattern is temporarily broken (see https://github.com/golang/go/issues/44887), but should be fixed in 1.16.x.

----

The TestServiceHealthEventsFromChanges function was over 1400 lines.
Attempting to debug test failures in test functions this large is
difficult. It requires scrolling to the line which defines the testcase
because the failure message only includes the line number of the
assertion, not the line number of the test case.

This is an excellent example of where test tables stop working well, and
start being a problem. To mitigate this problem, the runCase pattern can
be used. When one of these tests fails, a failure message will print the
line number of both the test case and the assertion. This allows a
developer to quickly jump to both of the relevant lines, signficantly
reducing the time it takes to debug test failures.

For example, one such failure could look like this:

    catalog_events_test.go:1610: case: service reg, new node
    catalog_events_test.go:1605: assertion failed: values are not equal

### DIFF
--- a/agent/consul/state/catalog_events_oss_test.go
+++ b/agent/consul/state/catalog_events_oss_test.go
@@ -1,7 +1,0 @@
-// +build !consulent
-
-package state
-
-func withServiceHealthEnterpriseCases(cases []serviceHealthTestCase) []serviceHealthTestCase {
-	return cases
-}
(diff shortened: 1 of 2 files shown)


kyhavlov (approved):
LGTM

PR #9882, opened 2021-03-16
state: use constants and add indexer tests for gateway tables

dnephin:
The mesh-topology and gateway-services tables already use indexers that are pretty close to the new functional index pattern.

This PR doesn't change the indexers, but makes the following changes:
* adds tests for the indexers
* uses the index name constants for all operations
* moves the `UpstreamDownstream` type from `structs` to `state`, and un-exports it. This type is only used to store things, it's not part of any API.
* removes `gatewayServices` and `serviceGateways` functions, and calls the `tx.Get` operations directly. This makes it easier to see what the function is doing from the caller.

### DIFF
--- a/agent/consul/state/catalog_oss_test.go
+++ b/agent/consul/state/catalog_oss_test.go
@@ -40,6 +40,51 @@ func testIndexerTableChecks() map[string]indexerTestCase {
 	}
 }
 
+func testIndexerTableMeshTopology() map[string]indexerTestCase {
+	obj := structs.UpstreamDownstream{
+		Upstream:   structs.ServiceName{Name: "UpStReAm"},
+		Downstream: structs.ServiceName{Name: "DownStream"},
+	}
+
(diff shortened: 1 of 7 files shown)


agent/consul/state/catalog_oss_test.go
```diff
@@ -25,6 +25,96 @@ func testIndexerTableChecks() map[string]indexerTestCase {
...
+			},
+			write: indexValue{
+				source:   obj,
+				expected: []byte("upstream\x00downstream\x00"),
+			},
+		},
+		indexUpstream: {
+			read: indexValue{
+				source: structs.ServiceName{Name: "UpStReAm"},
+
```
kyhavlov, line 51: extra whitespace line here, and in the index case below too

kyhavlov (approved):
LGTM 👍 , 1 minor whitespace comment

PR #9886, opened 2021-03-16
[SDK] change all cases of *testing.T to testing.TB

eikenb:
Using the interface opens up the use of all methods to benchmarks as well as tests.

Should be complete a drop in replacement.

Tested changes manually against both consul-terraform-sync's and consul-template's use of the sdk/testutil's TestServer to validate real-world use cases.


dnephin (approved):
LGTM!

I usually like to define a small local interface for just the methods that get used by the package or function, but using any interface is an improvement, so I think we should merge as-is.

PR #9888, opened 2021-03-17
ui: Add all tooltips to the default tabbing order in the page

johncowen:
This amends our tooltip modifier to automatically add a `tabindex="0"` to
all of our tooltips (if they don't have a tabindex already).

This means that all tooltips will automatically be
added to the natural tab order of the page. I'm pretty sure we don't
currently require the ability to disable this automatic functionality
but if we do at some point in the future we can add an option to disable
it, meaning all tooltips will be tabbable by default.


kaxcode (approved):
LGTM

PR #9894, opened 2021-03-17
Update proxycfg state management and xDS generation for transparent proxy

freddygv:
TODO: Fix service manager lifecycle test


agent/xds/endpoints.go
```diff
@@ -47,45 +47,46 @@ func (s *Server) endpointsFromSnapshotConnectProxy(cfgSnap *proxycfg.ConfigSnaps
 	resources := make([]proto.Message, 0,
 		len(cfgSnap.ConnectProxy.PreparedQueryEndpoints)+len(cfgSnap.ConnectProxy.WatchedUpstreamEndpoints))
 
+	for id, chain := range cfgSnap.ConnectProxy.DiscoveryChain {
+		// Prepared queries get handled in the second loop below
+		if conf, ok := cfgSnap.ConnectProxy.UpstreamConfig[id]; ok && conf.DestinationType == structs.UpstreamDestTypePreparedQuery {
```
rboyer, line 52: Do you also want to skip the escape hatch clusters here? I guess I am not entirely certain I follow why the abort conditions on these two loops are not inverses.

agent/xds/endpoints.go
```diff
@@ -47,45 +47,46 @@ func (s *Server) endpointsFromSnapshotConnectProxy(cfgSnap *proxycfg.ConfigSnaps
 	resources := make([]proto.Message, 0,
 		len(cfgSnap.ConnectProxy.PreparedQueryEndpoints)+len(cfgSnap.ConnectProxy.WatchedUpstreamEndpoints))
 
+	for id, chain := range cfgSnap.ConnectProxy.DiscoveryChain {
```
rboyer, line 50: Discovery chains do not exist for prepared queries, so this conditional is unnecessary.

agent/xds/routes.go
```diff
@@ -42,37 +42,29 @@ func (s *Server) routesFromSnapshot(cInfo connectionInfo, cfgSnap *proxycfg.Conf
...
-		var chain *structs.CompiledDiscoveryChain
-		if u.DestinationType != structs.UpstreamDestTypePreparedQuery {
-			chain = chains[upstreamID]
+	for id, chain := range chains {
+		if chain.IsDefault() {
+			continue
 		}
 
-		if chain == nil || chain.IsDefault() {
-			// TODO(rb): make this do the old school stuff too
```
freddygv, line 59: What old stuff was this?

agent/proxycfg/state.go
```diff
@@ -295,12 +297,36 @@ func (s *state) initWatchesConnectProxy() error {
...
+	}
+
 	// Watch for updates to service endpoints for all upstreams
 	for _, u := range s.proxyCfg.Upstreams {
+		// This can be true if the upstream is a synthetic entry populated from centralized upstream config.
+		// Watches should not be created for them.
+		if u.CentrallyConfigured {
+			continue
+		}
+		snap.ConnectProxy.UpstreamConfig[u.Identifier()] = &u
```
rboyer, line 320: Note you are taking the address of the loop variable here, not the spot in the slice.

agent/proxycfg/state.go
```diff
@@ -1515,25 +1603,35 @@ func makeUpstream(g *structs.GatewayService) structs.Upstream {
...
-		EvaluateInDatacenter: s.source.Datacenter,
-		EvaluateInNamespace:  u.DestinationNamespace,
-	}, "discovery-chain:"+u.Identifier(), s.ch)
+		Datacenter:             s.source.Datacenter,
+		QueryOptions:           structs.QueryOptions{Token: s.token},
+		Name:                   name,
+		EvaluateInDatacenter:   s.source.Datacenter,
+		EvaluateInNamespace:    namespace,
+		OverrideProtocol:       cfg.Protocol,
+		OverrideConnectTimeout: cfg.ConnectTimeout(),
```
rboyer, line 1619: Make sure these aren't the defaulted values that defaulted during parsing of the config map.

agent/xds/listeners.go
```diff
@@ -32,6 +32,10 @@ import (
 	"github.com/hashicorp/consul/logging"
 )
 
+const (
+	TProxyOutboundPort = 15001
```
rboyer, line 36: Might want to file a TODO here about possibly making this configurable somewhere.

agent/xds/listeners.go
```diff
@@ -54,35 +58,225 @@ func (s *Server) listenersFromSnapshot(cInfo connectionInfo, cfgSnap *proxycfg.C
 
 // listenersFromSnapshotConnectProxy returns the "listeners" for a connect proxy service
 func (s *Server) listenersFromSnapshotConnectProxy(cInfo connectionInfo, cfgSnap *proxycfg.ConfigSnapshot) ([]proto.Message, error) {
-	// One listener for each upstream plus the public one
-	resources := make([]proto.Message, len(cfgSnap.Proxy.Upstreams)+1)
+	resources := make([]proto.Message, 1)
```
rboyer, line 61: Is there a better guesstimate that we can use here now?

agent/xds/listeners.go
```diff
@@ -54,35 +58,225 @@ func (s *Server) listenersFromSnapshot(cInfo connectionInfo, cfgSnap *proxycfg.C
...
+				Name: "envoy.filters.listener.original_dst",
+			},
 		}
+	}
+
+	var hasChains bool
+
+	for id, chain := range cfgSnap.ConnectProxy.DiscoveryChain {
+		upstreamCfg := cfgSnap.ConnectProxy.UpstreamConfig[id]
+		if upstreamCfg != nil && upstreamCfg.DestinationType == structs.UpstreamDestTypePreparedQuery {
```
rboyer, line 91: Don't need this, since PQs can't have chains.

agent/proxycfg/state.go
freddygv: Yep, that is decoded without defaulting:
https://github.com/hashicorp/consul/blob/master/agent/proxycfg/state.go#L380

agent/xds/listeners.go
```diff
@@ -514,104 +708,126 @@ func (s *Server) makePublicListener(cInfo connectionInfo, cfgSnap *proxycfg.Conf
...
-			routeName:        "public_listener",
-			cluster:          LocalAppClusterName,
-			statPrefix:       "",
-			routePath:        "",
-			requestTimeoutMs: cfg.LocalRequestTimeoutMs,
-		}
+	// Override with bind port if one is set, otherwise default to
+	// proxy service's address
+	port := cfgSnap.Port
+	// if cfgSnap.Proxy.TransparentProxy {
```
rboyer, line 741: Did you want to put a TODO here for later?

agent/xds/listeners.go
```diff
@@ -514,104 +708,126 @@ func (s *Server) makePublicListener(cInfo connectionInfo, cfgSnap *proxycfg.Conf
...
 
-		filter, err := makeListenerFilter(opts)
+	filterOpts := listenerFilterOpts{
+		protocol:         cfg.Protocol,
+		filterName:       name,
+		routeName:        name,
+		cluster:          LocalAppClusterName,
+		requestTimeoutMs: cfg.LocalRequestTimeoutMs,
+	}
+	if useHTTPFilter {
```
rboyer, line 757: You could use that new finalize method here to avoid duplication.

agent/xds/listeners.go
freddygv: That should have been deleted, since there's no constant for the inbound port number. Fixed.

agent/xds/listeners.go
freddygv: There isn't a great one. If in TProxy mode there might only be 2 listeners, since it relies on per-upstream filter chains rather than per-upstream listeners. Outside of TProxy mode it could be the size of the DiscoverChain map.

Not sure if it's worth doing having a variable starting capacity based on whether TProxy is on or not.

agent/xds/listeners.go
```diff
@@ -973,6 +1189,109 @@ func (s *Server) makeMeshGatewayListener(name, addr string, port int, cfgSnap *p
...
+func (s *Server) makeUpstreamFilterChainForDiscoveryChain(
+	id string,
+	overrideCluster string,
+	protocol string,
+	u *structs.Upstream,
+	chain *structs.CompiledDiscoveryChain,
+	cfgSnap *proxycfg.ConfigSnapshot,
+	tlsContext *envoy_tls_v3.DownstreamTlsContext,
+) (*envoy_listener_v3.FilterChain, error) {
+	// TODO (freddy) Make this actually legible
```
rboyer, line 1201: I feel like there's some work we could do during ConfigSnapshot construction to alleviate some of this pain down here in the xDS code.

agent/xds/listeners.go
freddygv: Yep. Done.

agent/xds/listeners.go
```diff
@@ -54,35 +58,225 @@ func (s *Server) listenersFromSnapshot(cInfo connectionInfo, cfgSnap *proxycfg.C
...
+		}
+		filterChain.FilterChainMatch = &envoy_listener_v3.FilterChainMatch{
+			PrefixRanges: ranges,
+		}
+
+		// Only attach the filter chain if there are addresses to match on
+		if len(ranges) > 0 {
+			outboundListener.FilterChains = append(outboundListener.FilterChains, filterChain)
+		}
+		hasChains = true
```
rboyer, line 215: Please rename this to be something like `hasFilterChains` since my initial read at the top was _discovery_ chain.

agent/xds/listeners.go
```diff
@@ -54,35 +58,225 @@ func (s *Server) listenersFromSnapshot(cInfo connectionInfo, cfgSnap *proxycfg.C
...
+	}
+
+	// Only create the outbound listener when there are upstreams and filter chains are present
+	if outboundListener != nil && hasChains {
+		// Filter chains are stable sorted to avoid draining if the list is provided out of order
```
rboyer, line 220: :+1:

agent/proxycfg/state.go
rboyer: I think now you're taking the address of a copy of the slice item.

I think the fix would be `= &s.proxyCfg.Upstreams[i]`

rboyer (approved):
Left more comments, but LGTM for the alpha.

PR #9895, opened 2021-03-17
ui: Updates to auth-method show page

kaxcode:
- Update Certificate to be monospaced
- Add empty states for claim and list claim mappings
- Update the styling of empty state actions block

[Demo Link](https://consul-ui-staging-ndzws32au-hashicorp.vercel.app/ui/dc1/acls/auth-methods/capacitor-0/auth-method#CONSUL_ACLS_ENABLE=1)

Screen Shot 2021-03-18 at 9 32 11 AM

Screen Shot 2021-03-17 at 5 04 40 PM

### DIFF
--- a/ui/packages/consul-ui/app/styles/components/empty-state/skin.scss
+++ b/ui/packages/consul-ui/app/styles/components/empty-state/skin.scss
@@ -2,9 +2,6 @@
   color: $gray-500;
   background-color: $gray-010;
 }
-%empty-state > ul {
-  border-color: $gray-300;
-}
 %empty-state-header {
   border-bottom: none;
 }
--- a/ui/packages/consul-ui/app/components/certificate/index.hbs
+++ b/ui/packages/consul-ui/app/components/certificate/index.hbs
@@ -7,7 +7,7 @@
   >
   </button>
 {{#if this.show}}
-  <div class="key">{{@item}}</div>
+  <code>{{@item}}</code>
 {{else}}
   <hr />
 {{/if}}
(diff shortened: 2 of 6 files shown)


ui/packages/consul-ui/app/styles/components/empty-state/skin.scss
```diff
@@ -2,9 +2,6 @@
   color: $gray-500;
   background-color: $gray-010;
 }
-%empty-state > ul {
-  border-color: $gray-300;
-}
 %empty-state-header {
```
johncowen, line 5: Are we ok removing this line app wide? I had a look in the various figma files and it seems like sometimes this line is there and sometimes its not, all in all I'm not massively clear to me whether this needs removing app wide or only on certain empty states, @jnwright can you let us know whats required here?

johncowen (approved):
🔥 awesome

I left one comment but its mainly for Jasmine to resolve for us.

I suppose the last thing is, if it was me I'd add the carriage returns in the PEM certificate format to the mocked data so its a standard looking cert, that way you get a more realistic idea of what this will look like to users. AFAIK they generally have returns in them.

kaxcode:
> I suppose the last thing is, if it was me I'd add the carriage returns in the PEM certificate format to the mocked data so its a standard looking cert, that way you get a more realistic idea of what this will look like to users. AFAIK they generally have returns in them.

@johncowen I've updated the mock data and updated the screen shot in PR. [Demo Link](https://consul-ui-staging-bmp4y72eb-hashicorp.vercel.app/ui/dc1/acls/auth-methods/application-4/auth-method)

ui/packages/consul-ui/app/styles/components/empty-state/skin.scss
jnwright: It can be removed 👍🏼

PR #9897, opened 2021-03-18
Retroactive Chagelog update for header fix

banks:
This was fixed as part of a refactor in 1.9.0 but no changelog entry was added. It affected some users who used automated security scanning tools which detected violations on Consul's error pages because appropriate security-related headers were not being added.

@mikemorris @alvin-huang do you think it makes sense to backport this to the changelog in release/1.9.x too? It seems to to me so I added the label but let me know if there's a good reason not to.


mikemorris (approved):
LGTM, thanks for writing this up, 👍 to backporting to 1.9.x branch.

PR #9899, opened 2021-03-18
Add methods to check intention has wildcard src or dst

freddygv:
This is needed for an oss/ent split, where the enterprise side checks for wildcard name AND namespace.


mkeeler (approved):
LGTM

PR #9903, opened 2021-03-18
api: enable query options on agent endpoints

shoenig:
This PR adds support for setting `QueryOptions` on a few `Agent` API
endpoints. Nomad needs to be able to set the `Namespace` field on
these endpoints to:
 - query for services / checks in a namespace
 - deregister services / checks in a namespace
 - update TTL status on checks in a namespace

### DIFF
--- a/.changelog/9903.txt
+++ b/.changelog/9903.txt
@@ -1,0 +1,4 @@
+```release-note:improvement
+api: Enable setting query options on agent endpoints.
+```
+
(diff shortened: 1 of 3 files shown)


shoenig:
I'm not sure of where to edit the changelog, https://github.com/hashicorp/consul/blob/master/CHANGELOG.md seems really out of date. The test-integrations failure looks like some docker auth error.

dnephin (approved):
Thank you for the PR! LGTM with a changelog entry. The changelog entry can be added by adding the file `.changelog/9903.txt`, see https://github.com/hashicorp/consul/pull/8387 for some examples

PR #9907, opened 2021-03-19
ui: Dependency Upgrade

johncowen:
This PR upgrades the majority of our dependencies to the latest versions

Notes:

- `js-yaml` required a small change of method names
- `ember-page-title` stopped requiring `ember-cli-head` so we removed that component from our app.
- We pinned `ember-changeset-validations` (and its dependencies) to the version we were on due to a `validator is not a function` error that seemed to be coming from an upgrade `validated-changeset` in later versions. We can look deeper into that at a later date.
- `ember-href-to` also seemed problematic so I've left that temporarily also knowing we partly tweak this in our app code.

Anything related to Ember 3.20 we didn't upgrade to stick to the dependencies for 3.20, notably we also skipped upgrading ember lint related things due to the amount of errors and therefore churn it brings. We'll do this in a full ember upgrade PR.

All in all, all remaining upgrades (apart from `ember-changeset-validations` and `ember-href-to`) will be done with a full ember upgrade.


kaxcode (approved):
🎉 LGTM

PR #9908, opened 2021-03-19
ui: Add a README to the new Certificate component

kaxcode:


ui/packages/consul-ui/app/components/certificate/README.mdx
```diff
@@ -0,0 +1,31 @@
...
+
+| Argument | Type | Default | Description |
+| --- | --- | --- | --- |
+| `item` | `String` | | The certificate as a string to be displayed and copied |
+| `name` | `String` | | The 'Name' of the certificate to be displayed and copied. Mainly used for giving feedback to the user. |
+
+This component has the following:
+- a copy button
+- a visibility button to show and hide certificate
+- a hidden and visible display of the certificate
```
johncowen, line 24: Ah I previewed this and I can see what you were saying about the `ul`s - the bullets don't show as we have them 'turned off' by default on the UI (the docs currently take all the styles from the UI). You can overwrite this in `debug.scss` if you want so the bullets show up just in the docs pages.

johncowen (approved):
LGTM! There was one tiny thing I noticed with the obfuscating dots are over to the right a bit in the docs page, but! I'd much rather spend our time right now getting the docs pages together - if for some reason they look slightly different in the docs right now, its no big deal, so I wouldn't bother looking at this just yet, I'd just merge as is.

I'm hoping that once we get a largish break in feature work, we can go through the docs pages and tweak what needs tweaking into shape (i.e. any rendering differences, copy updates or whatever)

ui/packages/consul-ui/app/components/certificate/README.mdx
kaxcode: I think it's fine. When I read these files I read them raw. The bullets do show in GitHub preview so it's 💯

PR #9910, opened 2021-03-19
cli: Add new `consul connect redirect-traffic` command for applying traffic redirection rules when Transparent Proxy is enabled.

ishustava:
This PR proposes the following changes:

- Add new `consul connect redirect-traffic` command for applying traffic redirection rules when Transparent Proxy is enabled.
- Add new `iptables` package for applying traffic redirection rules with iptables.


sdk/iptables/iptables.go
```diff
@@ -0,0 +1,158 @@
...
+import (
+	"errors"
+	"fmt"
+	"os/exec"
+	"strconv"
+)
+
+const (
+	// Chain to intercept inbound traffic
+	ProxyInboundChain = "PROXY_INBOUND"
```
freddygv, line 12: What if we added a `CONSUL_` prefix to make it clear to users where these rules came from? Mostly for the VMs case where there may be other processes modifying `iptables` rules.

sdk/iptables/iptables.go
```diff
@@ -0,0 +1,158 @@
...
+	// IptablesProvider is the Provider that will apply iptables rules.
+	IptablesProvider Provider
+}
+
+// Provider is an interface for executing iptables rules.
+type Provider interface {
+	// AddRule adds a rule without executing it.
+	AddRule(name string, args ...string)
+	// ApplyRules executes rules that have been added via AddRule.
+	ApplyRules() error
```
freddygv, line 45: It could be good to document that this operation isn't atomic, and you could end up with some rules applied but not others if there's an error.

sdk/iptables/iptables.go
```diff
@@ -0,0 +1,158 @@
...
+	Rules() []string
+}
+
+// iptablesExecutor implements IptablesProvider using exec.Cmd.
+type iptablesExecutor struct {
+	commands []*exec.Cmd
+}
+
+func (i *iptablesExecutor) AddRule(name string, args ...string) {
+	i.commands = append(i.commands, exec.Command(name, args...))
```
freddygv, line 56: Out of curiosity, did you come across any iptables libraries? Wondering if there are alternatives to exec'ing these commands directly.

sdk/iptables/iptables.go
```diff
@@ -0,0 +1,158 @@
...
+		cfg.IptablesProvider.AddRule("iptables", "-t", "nat", "-A", ProxyOutputRedirectChain, "-p", "tcp", "-j", "REDIRECT", "--to-port", strconv.Itoa(cfg.ProxyOutboundPort))
+
+		// For outbound TCP traffic jump from OUTPUT chain to PROXY_OUTPUT chain.
+		cfg.IptablesProvider.AddRule("iptables", "-t", "nat", "-A", "OUTPUT", "-p", "tcp", "-j", ProxyOutputChain)
+
+		// Don't redirect Envoy traffic back to itself, return it to the next chain for processing.
+		cfg.IptablesProvider.AddRule("iptables", "-t", "nat", "-A", ProxyOutputChain, "-m", "owner", "--uid-owner", cfg.ProxyUserID, "-j", "RETURN")
+
+		// Skip localhost traffic, doesn't need to be routed via the proxy.
+		cfg.IptablesProvider.AddRule("iptables", "-t", "nat", "-A", ProxyOutputChain, "-d", "127.0.0.1/32", "-j", "RETURN")
```
freddygv, line 122: Have you seen whaat others do for IPv6 here? It's not something to change now necessarily, but to keep in mind.

sdk/iptables/iptables.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// Configure outbound rules.
+	{
+		// Redirects outbound TCP traffic hitting PROXY_REDIRECT chain to Envoy's outbound listener port.
+		cfg.IptablesProvider.AddRule("iptables", "-t", "nat", "-A", ProxyOutputRedirectChain, "-p", "tcp", "-j", "REDIRECT", "--to-port", strconv.Itoa(cfg.ProxyOutboundPort))
+
+		// For outbound TCP traffic jump from OUTPUT chain to PROXY_OUTPUT chain.
+		cfg.IptablesProvider.AddRule("iptables", "-t", "nat", "-A", "OUTPUT", "-p", "tcp", "-j", ProxyOutputChain)
+
+		// Don't redirect Envoy traffic back to itself, return it to the next chain for processing.
```
freddygv, line 118:
```suggestion
		// Don't redirect proxy traffic back to itself, return it to the next chain for processing.
```

This library doesn't need to be Envoy specific, so we can use a more generic word here. Same applies to the other Envoy references in this file.

sdk/iptables/iptables.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	return rules
+}
+
+// Setup will set up iptables interception and redirection rules
+// based on the configuration provided in cfg.
+// This implementation was inspired by
+// https://github.com/openservicemesh/osm/blob/650a1a1dcf081ae90825f3b5dba6f30a0e532725/pkg/injector/iptables.go
+func Setup(cfg Config) error {
```
freddygv, line 94: Is this idempotent? What happens if it's run twice? Either with the same or different Config.

Should we attempt to drop the chains before adding them?

website/content/commands/connect/envoy.mdx
```diff
@@ -1,8 +1,8 @@
 ---
 layout: commands
-page_title: 'Commands: Connect Proxy'
+page_title: 'Commands: Connect Envoy'
```
freddygv, line 3: Nice catch!

website/content/commands/connect/redirect-traffic.mdx
```diff
@@ -0,0 +1,73 @@
...
+  when using Connect in Transparent Proxy mode.
+---
+
+# Consul Connect Redirect Traffic
+
+Command: `consul connect redirect-traffic`
+
+The connect redirect-traffic command is used to apply traffic redirection rules to enforce
+all traffic to go through the [Envoy proxy](https://envoyproxy.io) when using [Consul
+Connect](/docs/connect/) in the Transparent Proxy mode.
```
freddygv, line 16:
```suggestion
all traffic to go through the [Envoy proxy](https://envoyproxy.io) when using [Consul
service mesh](/docs/connect/) in the Transparent Proxy mode.
```

website/content/commands/connect/redirect-traffic.mdx
```diff
@@ -0,0 +1,73 @@
...
+
+This command requires `iptables` command line utility to be installed,
+and as a result, this command can currently only run on linux.
```
freddygv, line 19: We should also add the permissions required to run the command.

sdk/iptables/iptables.go
```diff
@@ -0,0 +1,158 @@
+package iptables
```
freddygv, line 1: Since this will only support linux, we could set up alternative os-specific implementations and return an error if used in not-linux.

Here's an example where we have an alternative Windows implementation:
https://github.com/hashicorp/consul/blob/87f6617eecd23a64add1e79eb3cd8dc3da9e649e/agent/config/limits_windows.go
https://github.com/hashicorp/consul/blob/87f6617eecd23a64add1e79eb3cd8dc3da9e649e/agent/config/limits.go

freddygv (commented):
This is looking good @ishustava, nice work! I added questions/comments in line.

command/connect/redirecttraffic/redirect_traffic.go
```diff
@@ -0,0 +1,141 @@
...
+
+		svc, _, err := c.client.Agent().Service(c.proxyID, nil)
+		if err != nil {
+			return iptables.Config{}, fmt.Errorf("failed to fetch proxy service from Consul Agent: %s", err)
+		}
+
+		cfg.ProxyInboundPort = svc.Port
+
+		// todo: change once it's configurable
+		cfg.ProxyOutboundPort = xds.TProxyOutboundPort
```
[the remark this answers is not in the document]
rboyer: I was just coming here to say that the `command` package probably shouldn't import `agent/xds`

command/connect/redirecttraffic/redirect_traffic.go
```diff
@@ -0,0 +1,141 @@
...
+	if c.proxyID != "" {
+		var err error
+		if c.client == nil {
+			c.client, err = c.http.APIClient()
+			if err != nil {
+				return iptables.Config{}, fmt.Errorf("error creating Consul API client: %s", err)
+			}
+		}
+
+		svc, _, err := c.client.Agent().Service(c.proxyID, nil)
```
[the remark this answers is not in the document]
ishustava: Hmmm, I don't think so since the client has the namespace set in its config when we call `c.http.APIClient()` above.

sdk/iptables/iptables.go
```diff
@@ -0,0 +1,158 @@
...
+	ProxyInboundChain = "PROXY_INBOUND"
+
+	// Chain to redirect inbound traffic to the proxy
+	ProxyInboundRedirectChain = "PROXY_IN_REDIRECT"
+
+	// Chain to intercept outbound traffic
+	ProxyOutputChain = "PROXY_OUTPUT"
+
+	// Chain to redirect outbound traffic to the proxy
+	ProxyOutputRedirectChain = "PROXY_REDIRECT"
```
freddygv, line 21: This could be a place to add the default inbound and outbound ports.

For inbound it would be 21000, since that's the default for Connect proxies.

sdk/iptables/iptables.go
ishustava: Yeah, I like that a lot!

sdk/iptables/iptables.go
ishustava: Yep, will do! Although, I'll change it later to be atomic to better support VMs.

sdk/iptables/iptables.go
ishustava: Nope, all go libraries I came across used `exec` unfortunately.

sdk/iptables/iptables.go
ishustava: Yeah, that's a great call-out. I've been thinking about your comment above about documenting that `ApplyRules` is not atomic, and I think it makes sense to make that one atomic and this one idempotent. I'm thinking to address it in a follow-up PR because most likely I'll switch to using `iptables-save` and `iptables-restore` to do that.

sdk/iptables/iptables.go
ishustava: From what I've seen, only Istio supports IPv6. That's a great question though, and I think we should investigate whether we should and how to support IPv6. I'll add a task for that in the future.

command/connect/redirecttraffic/redirect_traffic.go
```diff
@@ -0,0 +1,141 @@
...
+		if err != nil {
+			return iptables.Config{}, fmt.Errorf("failed to fetch proxy service from Consul Agent: %s", err)
+		}
+
+		cfg.ProxyInboundPort = svc.Port
+
+		// todo: change once it's configurable
+		cfg.ProxyOutboundPort = xds.TProxyOutboundPort
+	} else {
+		cfg.ProxyInboundPort = c.proxyInboundPort
```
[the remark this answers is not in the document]
ishustava: Yep, after our zoom convo, changed the outbound port to default if not provided and inbound port will be required.

command/connect/redirecttraffic/redirect_traffic_test.go
```diff
@@ -0,0 +1,148 @@
...
+				proxyOutboundPort: 16000,
+			},
+			false,
+			iptables.Config{
+				ProxyUserID:       "1234",
+				ProxyInboundPort:  15000,
+				ProxyOutboundPort: 16000,
+			},
+			"",
+		},
```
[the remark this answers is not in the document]
ishustava: Added for the outbound port!

sdk/iptables/iptables.go
ishustava: Done!

command/connect/redirecttraffic/redirect_traffic.go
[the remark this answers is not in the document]
ishustava: 💯 Agree. Sorry, I think misunderstood how defaulting should work and thought this const will go away eventually and so was using it temporarily. After syncing with Freddy, I've moved it to the `iptables` package instead.

command/connect/redirecttraffic/redirect_traffic.go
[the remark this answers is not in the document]
freddygv: Ah yep, that should be taken care of then.

freddygv (approved):
🚀

PR #9911, opened 2021-03-19
state: convert ACLRoles policies index to new functional indexer pattern

dnephin:
Just like all the others.

This also changes the `acl-roles.polcies` index to use an encoded UUID, where as previously it used a string. This should save a bit of memory for each linked policy.


banks (approved):
LGTM!

PR #9917, opened 2021-03-23
Document agent token policy requirement for rexec

Ranjandas:
The Agent token policy when using rexec should have `write` on "_rexec" key prefix. Updated the exec command documentation to explicitly state this requirement.


dnephin (approved):
LGTM, thank you!

PR #9923, opened 2021-03-24
http: fix a bug that would cause runtimeConfig to be cached

dnephin:
This bug would result in the UI not having the correct settings in Consul enterprise, which could produce many warnings in the logs.

This bug occurred because the index page, which includes a map of configuration was rendered when the HTTPHandler is first created. This PR changes the UIServer to instead render the index page when the page is requested.

The rendering does not appear to be all that expensive, so rendering it when requested should not cause much extra latency.

Also includes a bit of cleanup to remove some unnecessary args, and unnecessary function wrapping.

This bug only exists in 1.9.x, so I think we only need one backport.

### DIFF
--- a/agent/uiserver/uiserver.go
+++ b/agent/uiserver/uiserver.go
@@ -12,9 +12,10 @@ import (
 	"sync/atomic"
 	"text/template"
 
+	"github.com/hashicorp/go-hclog"
+
 	"github.com/hashicorp/consul/agent/config"
 	"github.com/hashicorp/consul/logging"
-	"github.com/hashicorp/go-hclog"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


agent/uiserver/uiserver.go
```diff
@@ -123,55 +105,47 @@ func (h *Handler) ReloadConfig(newCfg *config.RuntimeConfig) error {
...
+		return cfg
 	}
 	return nil
 }
 
-func (h *Handler) serveUIMetricsProviders(resp http.ResponseWriter, req *http.Request) {
+func (h *Handler) serveUIMetricsProviders(resp http.ResponseWriter) {
 	// Reload config in case it's changed
-	state := h.getState()
+	cfg := h.getRuntimeConfig()
```
rboyer, line 138: may be nil

agent/uiserver/uiserver.go
```diff
@@ -50,71 +43,60 @@ type reloadableState struct {
...
-	newState := reloadableState{
-		cfg: &newCfg.UIConfig,
-	}
+	h.runtimeConfig.Store(newCfg)
+	return nil
+}
 
-	var fs http.FileSystem
+func (h *Handler) handleIndex() (http.Handler, error) {
+	cfg := h.getRuntimeConfig()
```
rboyer, line 88: same comment about this returning nil

agent/uiserver/uiserver.go
dnephin: Ya, some callers did explicitly panic with a custom message, and others did not. In practice it should never be nil, but if there is a bug, it will panic on the next line either way. A panic on line 140 should be unambiguous because the `UIConfig` field is not a pointer, so only `cfg` could be nil.

rboyer (approved):
LGTM

banks (approved):
LGTM! Thanks Daniel.

Issue #9891, opened 2021-03-17
Add telemetry and logging around expired certificates 


preetapan:
Consul has knowledge about various cerificates ( the agent certificates for TLS communication across clients/servers, Connect CA certificates). 

Currently, we don't emit warnings when certificates are about to expire. It's helpful for operators to know about this so that they can set up alerts within their monitoring systems. We could also emit warning logs with the name/type of certificate and how much time is left that trigger based on a window. 

This issue captures both needs (logs and adding metrics to the v1/metrics end point with seconds remaining for expiration). Can be split into multiple issues as needed.

ashwinkupatkar:
Hi @preetapan,

Any tentative timeline when this feature will be rolled out?

dnephin:
Hi @ashwinkupatkar , thank you for your interest in this issue! Hopefully we can get a metric into the next release.

I've opened a PR for one possible option in #9924. Would this work for your use case? If not, could you share more about what you would like to see?

Thank you!

ashwinkupatkar:
Hi @dnephin,

Thanks for taking a look.

Yes, this is one of the metric that is needed. Apart from this metric we also need metric for the consul server certificates and consul client certificate expiration data.

ashwinkupatkar:
Hello, @dnephin 

Just wanted to follow up on this feature. Any idea in which version this would be launched ?

ashwinkupatkar:
Hi @dnephin, any clue ... when this feature would be available ? thanks

dnephin:
Hi @ashwinkupatkar , I learned there were a few more certs that should be tracked, and I haven't had a chance to finish adding the metrics. It won't be for 1.10, so at the earliest 1.11.

ashwinkupatkar:
Hi @dnephin, I see the change has been merged to master. So should I expect it in GA of 1.10.0 ?
Thanks

dnephin:
For anyone watching this issue, the following PRs add new metrics:
* #9924
* #10504
* #10768

And #10770 adds logging when a cert is about to expire in the next 24h.

I believe this covers all the cases, but if there is something missing please do comment here.

dnephin:
For anyone watching this issue, the following PRs add new metrics:
* #9924
* #10504
* #10768

And #10770 adds logging when a cert is about to expire in the next 24h.

I believe this covers all the cases, but if there is something missing please do comment here.

dnephin:
[...]

PR #9924, opened 2021-03-24
connect: emit a metric for the seconds until root CA expiry

dnephin:
Implements the metrics portion of #9891

This PR adds a new metric, `consul.mesh.root-ca.expiry` which is a gauge. The metric is update every hour by a goroutine run from the leader. The value of the gauge is the number of seconds until the root CA expires. Someone running Consul should be able to setup an alert on this value, so that they get notified when the value drops below some threshold.

This is difficult to unit test so I might try and test it out manually.

TODO:
* [x] manual test (using `socat -d - udp6-listen:8125` as a statsd server)
* [x] update static metric definitions so this is populated properly when used with prometheus.
* [x] changelog


agent/consul/leader_connect.go
```diff
@@ -146,6 +148,47 @@ func (s *Server) pruneCARoots() error {
 	return nil
 }
 
+func emitCAExpirationMetrics(s *Server) func(ctx context.Context) error {
+	key := []string{"mesh", "root-ca", "expiry"}
```
dnephin, line 152: Thoughts on this metric name? I couldn't find any other CA specific metrics. We could use a different key.

kyhavlov (approved):
LGTM 👍 

I can't think of a better name for the metric - `mesh.root-ca.expiry` is concise and accurate.

website/content/docs/agent/telemetry.mdx
```diff
@@ -478,6 +478,7 @@ These metrics give insight into the health of the cluster as a whole.
...
 | `consul.catalog.connect.query-tags..` | Increments for each connect-based catalog query for the given service with the given tags.                                                                                                                                                                                                                                                                                                                                                | queries                                 | counter |
 | `consul.catalog.connect.not-found.`   | Increments for each connect-based catalog query where the given service could not be found.                                                                                                                                                                                                                                                                                                                                               | queries                                 | counter |
+| `consul.mesh.root-ca.expiry`          | The number of seconds until the root CA expires, updated every hour. | seconds | gauge |
```
freddygv, line 481: How about `active-root-ca` since multiple can be returned by `/connect/ca/roots`?

agent/consul/leader_metrics.go
```diff
@@ -0,0 +1,71 @@
...
+func rootCAExpiryMonitor(s *Server) certExpirationMonitor {
+	return certExpirationMonitor{
+		Key: metricsKeyMeshRootCAExpiry,
+		Labels: []metrics.Label{
+			{Name: "datacenter", Value: s.config.Datacenter},
+		},
+		Logger: s.logger.Named(logging.Connect),
+		Query: func() (time.Duration, error) {
+			state := s.fsm.State()
+			_, root, err := state.CARootActive(nil)
```
freddygv, line 33: Could the root be nil?

dnephin: I guess it could be nil. I added a case to handle that to prevent the panic.

dnephin:
Updated the metric name to `active-root-ca` and handled the case where the RootCA returns `nil`.

freddygv (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/389762.

PR #9932, opened 2021-03-25
Set default enterprise meta in test case

dnephin:
I merged #7324, and apparently thet test that had been lying dormant and not running needed an update for enterprise.


mikemorris (approved):
Makes sense in context that this wasn't being run previously.

PR #9934, opened 2021-03-26
Docs: add enterprise upgrade link

mikegreen:
easier to do this than rebase pull/8643

### DIFF
--- a/website/content/docs/upgrading/index.mdx
+++ b/website/content/docs/upgrading/index.mdx
@@ -18,7 +18,7 @@ keep in mind when using Consul.
 
 This page documents how to upgrade Consul when a new version is released.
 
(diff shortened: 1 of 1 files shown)


website/content/docs/upgrading/index.mdx
```diff
@@ -18,6 +18,8 @@ keep in mind when using Consul.
 
 This page documents how to upgrade Consul when a new version is released.
 
+~> For Consul Enterprise, see the [Automated Upgrades documentation](/docs/enterprise/upgrades).
```
im2nguyen, line 21: Is there a reason why you chose the yellow callout instead of `-> **Tip:**`?

mikegreen: Nope, copied from an old PR way back so prob didn't know the colors back then :-) just tweaked

jsosulska (approved):
LGTM!

PR #9935, opened 2021-03-26
ui: Auth Methods -  Namespace rules tab

kaxcode:
- Added mock data for NamespaceRules
- Created NamespaceTable with styling
- Created a Namespace rules tab with a flag to only show the tab if namespaces are enabled

[Demo Link
](https://consul-ui-staging-191qln2jp-hashicorp.vercel.app/ui/dc1/acls/auth-methods/alarm-6/namespace-rules#CONSUL_ACLS_ENABLE=1)

Screen Shot 2021-03-26 at 4 00 01 PM

### DIFF
--- a/ui/packages/consul-ui/app/components/consul/auth-method/index.scss
+++ b/ui/packages/consul-ui/app/components/consul/auth-method/index.scss
@@ -62,8 +62,8 @@
   }
 }
 
-
-.consul-auth-method-namespace-table {
+// Nspace List
+.consul-auth-method-nspace-list {
   thead {
     td {
       color: var(--gray-500) !important;
--- a/ui/packages/consul-ui/app/components/consul/auth-method/index.scss
+++ b/ui/packages/consul-ui/app/components/consul/auth-method/index.scss
@@ -24,9 +24,17 @@
         font-weight: $typo-weight-semibold;
         font-size: $typo-size-700;
       }
-      tbody td {
-        font-size: $typo-size-600;
-        color: $black;
+      tbody {
+        td {
+          font-size: $typo-size-600;
+          color: $black;
+        }
+        tr {
+          cursor: default;
+        }
+        tr:hover {
+          box-shadow: none;
+        }
       }
     }
   }
@@ -54,4 +62,26 @@
   }
 }
 
+
+.consul-auth-method-namespace-table {
+  thead {
+    td {
+      color: var(--gray-500) !important;
+      font-weight: $typo-weight-semibold !important;
+      font-size: $typo-size-700 !important;
+    }
+  }
+  tbody {
+    td {
+      font-size: $typo-size-600;
+      color: $black;
+    }
+    tr {
+      cursor: default;
+    }
+    tr:hover {
+      box-shadow: none;
+    }
+  }
+}
 
--- a/ui/packages/consul-ui/app/router.js
+++ b/ui/packages/consul-ui/app/router.js
@@ -199,8 +199,8 @@ export const routes = {
           'binding-rules': {
             _options: { path: '/binding-rules' },
           },
-          'namespace-rules': {
-            _options: { path: '/namespace-rules' },
+          'nspace-rules': {
+            _options: { path: '/nspace-rules' },
           },
         },
       },
--- a/ui/packages/consul-ui/app/router.js
+++ b/ui/packages/consul-ui/app/router.js
@@ -199,6 +199,9 @@ export const routes = {
           'binding-rules': {
             _options: { path: '/binding-rules' },
           },
+          'namespace-rules': {
+            _options: { path: '/namespace-rules' },
+          },
         },
       },
     },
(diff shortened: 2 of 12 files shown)


ui/packages/consul-ui/app/components/consul/auth-method/index.scss
```diff
@@ -24,9 +24,17 @@
         font-weight: $typo-weight-semibold;
         font-size: $typo-size-700;
       }
-      tbody td {
-        font-size: $typo-size-600;
-        color: $black;
+      tbody {
```
kaxcode, line 27: @johncowen I thought about making this into a styling component but this is an overwrite of the already existing %table. Later on I will take a look at what can be refactored in %table to delete all this. For now, I'm just overwriting to get this working.

ui/packages/consul-ui/app/templates/dc/acls/auth-methods/show.hbs
```diff
@@ -27,6 +27,9 @@
       compact
         (array
           (hash label="General info" href=(href-to "dc.acls.auth-methods.show.auth-method") selected=(is-href "dc.acls.auth-methods.show.auth-method"))
+    (if (env "CONSUL_NSPACES_ENABLED")
```
kaxcode, line 30: @johncowen Is this the right conditional? The demo site had it showing before I changed the scenario to be nspaces enabled. Does the demo site always have nspaces enabled?

ui/packages/consul-ui/app/components/consul/auth-method/index.scss
johncowen: Sounds great!

ui/packages/consul-ui/app/templates/dc/acls/auth-methods/show.hbs
johncowen: Yeah this is the right env var to use. You're right though, the demo site does have everything enabled by default. We should maybe change that in future now that we have our bookmarklets and URLs to be able to link to specific scenarios.

I mentioned a while back about using potentially using `can` for this type of feature flagging I saw some convo about others specifically not doing that, so maybe we should chat about it before we 100% decide.

TL;DR The conditional here is 💯

ui/packages/consul-ui/app/router.js
```diff
@@ -199,6 +199,9 @@ export const routes = {
           'binding-rules': {
             _options: { path: '/binding-rules' },
           },
+          'namespace-rules': {
+            _options: { path: '/namespace-rules' },
```
johncowen, line 203: Bit of a quibble this one. The URL here is great, although we refer to any Routes that are namespace related as `nspace`. So maybe we should call this one `nspace-rules`. 

I'm trying to imagine myself writing this in amongst some code and wondering which one I would write automatically without thinking `nspace-rules` or `namespace-rules` - as I've trained myself in this codebase to never write the word `namespace` I'd probably go with `nspace-rules`.

ui/packages/consul-ui/app/templates/dc/acls/auth-methods/show/namespace-rules.hbs
```diff
@@ -0,0 +1,21 @@
+
+<div class="tab-section">
+{{#if (gt item.NamespaceRules.length 0)}}
+  <p>A set of rules that can control which namespace tokens created via this auth method will be created within. Unlike binding rules, the first matching namespace rule wins.</p>
```
johncowen, line 4: I think this `<p>` tag needs to show whether you have rules or not, check with design maybe but in other places in the UI we still show the info paragraphs at the top if there are no results to show.

ui/packages/consul-ui/app/templates/dc/acls/auth-methods/show/namespace-rules.hbs
```diff
@@ -0,0 +1,21 @@
...
+  <Consul::AuthMethod::NamespaceTable @items={{item.NamespaceRules}} />
```
johncowen, line 5: Here's another thing that is a little difficult to give reasons/rules for naming things. 

Ideally this one would be `Consul::AuthMethod::NamespaceRule::List`. List here doesn't necessarily mean 'this is formatted using `li`s' more that this is multiple of them rather than a single one. Then if we ever need to change how we markup the `NamespaceRule::List` (so not use tables anymore), we don't have to rename all the areas in the code where we use the component.

johncowen (approved):
🔥 This is sooo good! 👏 

I left a few comments, and I think the only definite change we should make is to make that info paragraph of text show even when you have no rules.

Such a good PR, I went to look for the demo link in the GH Checks and saw you put me a nice demo link in the description, and I was thinking wow this PR gets even better. And then I clicked on it and saw you'd used the deep linking stuff we did to take me straight to the page AND enable the scenario/cookies I needed to see it - super good!

ui/packages/consul-ui/app/router.js
kaxcode: Renamed to be `nspace-rules`

PR #9938, opened 2021-03-27
Add Avi Network, AWS ALB and NS1 TF Registries and GitHub links

BeenzSyed:
* Add Avi Network GH link
* Add AWS ALB TF Registry and GH link
* Add NS1 TF Registry and GH link


devarshishah3 (approved):
Looks good

PR #9954, opened 2021-03-31
Add DeleteKV method to sdk testutil

sobczak-m:


jsosulska:
Hi @sobczak-m ,

Thanks for submitting this enhancement. Before merging this in, I wanted to check - what are you using the SDK for?
I ask because the SDK is more used as a test harness, than a fully fledged API. The maintainers would like to keep the changes here as small as possible, where possible. The [API package ](https://pkg.go.dev/github.com/hashicorp/consul/api)we offer is more fully featured for most development work, and I wonder if this fits your usecase more. 

Thanks!

sobczak-m:
Hi @jsosulska,
I use SDK testutil in integration tests for consul mock. I do something like this:
`consulMock.SetKV(t,"key",[]byte("value"))
someFunctionTriggeredByAddingKey()

consulMock.DeleteKV(t,"key")
someFunctionTriggeredByKeyRemoval()
`

jsosulska:
Hi @sobczak-m , 

I'm sorry for the long tail on this response. Thank you for your patience! After speaking with some folks internally, there was a suggestion made that would likely better serve your usecase with the [Consul API](https://pkg.go.dev/github.com/hashicorp/consul/api) package (and I learned some new ways of helping code this!)

First, the SDK exposes a [TestServer](https://pkg.go.dev/github.com/hashicorp/consul/sdk@v0.7.0/testutil#TestServer). You can then use a [client](https://pkg.go.dev/github.com/hashicorp/consul/api#NewClient) from the api module to model the calls you are trying to test. The client can call the [KV.Delete](https://pkg.go.dev/github.com/hashicorp/consul/api#KV.Delete) which will give you the needed functionality.

While this is a little bit more involved method of performing this functionality, the maintainers feel it more closely aligns with how Consul is intended to be used, and tested.
I am going to close this PR per feedback from the maintainers, and I hope the workflow above helps!

PR #9958, opened 2021-03-31
state: support additional test cases in indexer tests

dnephin:
And add a few additional cases.

### DIFF
--- a/agent/consul/state/schema_test.go
+++ b/agent/consul/state/schema_test.go
@@ -12,11 +12,18 @@ type indexerTestCase struct {
 	write      indexValue
 	prefix     []indexValue
 	writeMulti indexValueMulti
+	// extra test cases can be added if the indexer has special handling for
+	// specific cases.
+	extra []indexerTestCase
 }
 
 type indexValue struct {
 	source   interface{}
 	expected []byte
+	// expectedIndexMissing indicates that this test case should not produce an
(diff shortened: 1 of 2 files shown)


agent/consul/state/schema_test.go
```diff
@@ -54,42 +61,61 @@ func TestNewDBSchema_Indexers(t *testing.T) {
...
-			}
+	if i, ok := indexer.(memdb.MultiIndexer); ok {
+		valid, actual, err := i.FromObject(tc.writeMulti.source)
+		require.NoError(t, err)
+		require.True(t, valid)
+		require.Equal(t, tc.writeMulti.expected, actual)
+	}
+
+	for _, extra := range tc.extra {
+		t.Run("extra", func(t *testing.T) {
```
banks, line 117: Does it matter that these all end up with the same name? A cheap fix would be to include the slice index in the name like `extra[0], extra[1]` etc?

banks (approved):
Nice, this seems like a good addition!

One minor Q about the extra test names but optional as it's only one for now anyway.

agent/consul/state/schema_test.go
dnephin: This is exactly what Go does automatically for us! If there are duplicate names it adds `#01` , `#02`, etc.

PR #9959, opened 2021-03-31
catalog: add option to v1/health/service/:service endpoint to prefer returning connect capable nodes over underlying nodes if some are present

### LINKED ISSUE #9744: Provide the minimal connect setup in non-connect endpoints.
#### Feature Description

Provide the minimal connect setup in non-connect endpoints.

#### Use Case(s)

At integrating Consul-Connect natively in a load-balancer, I'm struggling with the existing endpoints.
* `/v1/health/service/foo-bar` returns a service but don't tell me whether it has a connect definition or not
* `/v1/health/connect/foo-bar` in fact returns me the equivalent of `/v1/health/service/foo-bar-sidecar-proxy` and actually returns nothing/empty if the service I'm requesting is not connect-enabled.

When watching endpoints to provision all services, including connect services, I have two options:
* watch concurrently /v1/health/service and /v1/health/connect, and manage all the pain associated to reconciliation of events, occurring on both endpoints, with potential delays between first and second event, etc.
* watch only /v1/health/service, force my users to add a "connect" tag, then add custom business logic on my side to deduce that I should fetch /v1/health/connect in addition, so that I can finally retrieve the connect IP+port (which is suboptimal).

I was wondering why all the connect config is [actually removed from the newly registered service](https://github.com/hashicorp/consul/blob/e7db6135f9a641952e660a8cb7dcd4ad7a67aa6f/agent/agent_endpoint.go#L980-L984).

To me, the connect endpoint gives a "connect-first" approach, where for ex. the Service.port is the connect one.
```
  "Service": {
      "Kind": "connect-proxy",
      "ID": "foo-bar-sidecar-proxy",
      "Service": "foo-bar-sidecar-proxy",
      "Port": 21001,
```
But still you have the downstream information available:
```
   "Proxy": {
        "DestinationServiceName": "foo-bar",
        "DestinationServiceID": "foo-bar",
        "LocalServiceAddress": "127.0.0.1",
        "LocalServicePort": 80,
        "MeshGateway": {},
        "Expose": {}
      },
```
#### Proposal

Can't we have the opposite as well with /v1/health/service? Such as:
```
   "Connect": {
        "ProxyServiceName": "foo-bar-sidecar-proxy",
        "ProxyServiceID": "foo-bar",
        "ProxyServiceAddress": "192.0.2.1",
        "ProxyServicePort": 21001,
      },
```
This would allow an event consumer to guess that a connect equivalent exists, and to grab its IP+port directly.
Maybe other properties would be interesting as well, but I've no use-case for now to identify which of them would make sense.

rboyer:
Adds a new parameter `?prefer-connect` to the existing endpoint `/v1/health/service/:service`

Fixes #9744

Ordinarily a request to `/v1/health/service/:service` will return standard catalog data for instances of `:service` (including the node each runs on, the definition of each, and their checks and check states). There is a similar API endpoint `/v1/health/connect/:service` that does something very similar, except it returns just connect-enabled endpoints for `:service`.

The former is what you would normally do for standard service discovery when you are not using Connect. The latter is what you do if you need to locate the ports of the proxies that front the real services when Connect is enabled.

If you run a mix of Connect and non-Connect workloads and wish to do some sort of general query to list all of your services and _figure out_ which ones are using Connect today you have to do both of these queries serially.

This PR handles doing this sort of operation under the covers so the caller only has to call `/v1/health/service/:service?prefer-connect` and the results will EITHER be those you would have gotten from `/v1/health/connect/:service` OR those you would have gotten from `/v1/health/service/:service`. You can tell the difference by checking the inner `Service.Kind` values. If they are unset that represents a "typical" service which is a plain service. If they are set to something like `connect-proxy` or `terminating-gateway` then you know that the results represent Connect enabled endpoints.

### DIFF
--- a/agent/cache-types/streaming_health_services.go
+++ b/agent/cache-types/streaming_health_services.go
@@ -14,6 +14,7 @@ import (
 	"github.com/hashicorp/consul/agent/cache"
 	"github.com/hashicorp/consul/agent/structs"
 	"github.com/hashicorp/consul/agent/submatview"
+	"github.com/hashicorp/consul/lib"
 	"github.com/hashicorp/consul/lib/retry"
 	"github.com/hashicorp/consul/proto/pbservice"
 	"github.com/hashicorp/consul/proto/pbsubscribe"
@@ -28,7 +29,8 @@ const (
 // catalog using the streaming gRPC endpoint.
 type StreamingHealthServices struct {
 	RegisterOptionsBlockingRefresh
-	deps MaterializerDeps
+	deps  MaterializerDeps
+	cache *cache.Cache // Cache that has StreamingHealthServices
 }
 
 // RegisterOptions returns options with a much shorter LastGetTTL than the default.
@@ -46,8 +48,8 @@ func (c *StreamingHealthServices) RegisterOptions() cache.RegisterOptions {
 
 // NewStreamingHealthServices creates a cache-type for watching for service
 // health results via streaming updates.
-func NewStreamingHealthServices(deps MaterializerDeps) *StreamingHealthServices {
-	return &StreamingHealthServices{deps: deps}
+func NewStreamingHealthServices(deps MaterializerDeps, cache *cache.Cache) *StreamingHealthServices {
+	return &StreamingHealthServices{deps: deps, cache: cache}
 }
 
 type MaterializerDeps struct {
@@ -63,11 +65,20 @@ type MaterializerDeps struct {
 // Fetch implements part of the cache.Type interface, and assumes that the
 // caller ensures that only a single call to Fetch is running at any time.
 func (c *StreamingHealthServices) Fetch(opts cache.FetchOptions, req cache.Request) (cache.FetchResult, error) {
+	srvReq, ok := req.(*structs.ServiceSpecificRequest)
+	if !ok {
+		return cache.FetchResult{}, fmt.Errorf(
+			"Internal cache failure: request wrong type: %T", req)
+	}
+
+	if srvReq.PreferConnect {
+		return c.preferConnectFetch(srvReq, opts.LastResult)
+	}
+
 	if opts.LastResult != nil && opts.LastResult.State != nil {
 		return opts.LastResult.State.(*streamingHealthState).Fetch(opts)
 	}
 
-	srvReq := req.(*structs.ServiceSpecificRequest)
 	newReqFn := func(index uint64) pbsubscribe.SubscribeRequest {
 		req := pbsubscribe.SubscribeRequest{
 			Topic:      pbsubscribe.Topic_ServiceHealth,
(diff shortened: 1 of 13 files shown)


agent/consul/health_endpoint.go
```diff
@@ -254,6 +257,7 @@ func (h *Health) ServiceNodes(args *structs.ServiceSpecificRequest, reply *struc
 		if args.Ingress {
 			key = "ingress"
 		}
+		// TODO(rb): handle PreferConnect here as well?
```
rboyer, line 260: Should this emit slightly different metrics when used this way?

agent/consul/state/catalog_events.go
```diff
@@ -358,6 +361,10 @@ func ServiceHealthEventsFromChanges(tx ReadTxn, changes Changes) ([]stream.Event
 	}
 	events = append(events, connectEvents...)
 
+	// Duplicate all of these into the preferconnect topic
```
rboyer, line 364: In the common case I would expect folks wouldn't use `?prefer-connect` so would this event duplication be too wasteful?

dnephin: Change events are emitted even when there are no subscribers. But when there are no subscribers the events will effectively be GC'ed immediately, so I think worst case scenario is extra unnecessary time spend GC'ing.

agent/cache-types/streaming_health_services.go
```diff
@@ -98,6 +109,64 @@ func (c *StreamingHealthServices) Fetch(opts cache.FetchOptions, req cache.Reque
...
+	var result cache.FetchResult
+	result.Value = plainResults
+	result.Index = maxIndex
+	return result, nil
+}
+
+func (c *StreamingHealthServices) fetchFromCache(req *structs.ServiceSpecificRequest) (uint64, *structs.IndexedCheckServiceNodes, error) {
+	// Background is fine here because this isn't a blocking query as no index is set.
+	// Therefore this will just either be a cache hit or return once the non-blocking query returns.
+	rawResult, rm, err := c.cache.Get(context.Background(), StreamingHealthServicesName, req)
```
rboyer, line 159: I cribbed this from how the ConnectCALeaf cache-type grabs the roots. I suspect there might be something fishy about this pattern used here.

banks:
>  If they are set to something like connect-proxy or terminating-gateway then you know that the results represent Connect enabled endpoints.

Presumably you would also need to check for Connect-native services too if they are used. So the full decision for "should I use Connect to talk to this instance would be something like: `service.Kind == "connect-proxy" || service.Kind == "terminating-gateway" || service.Connect.Native` right?

Not a big deal here but just checking understanding and we should make sure that's clear from docs too.

banks (commented):
Hmmm

I might be missing something, but this seems a lot more complicated than I was expecting.

Rather than have one cache type that includes lots of logic to produce results in two different modes including nested calls back into the cache, can we not just make a new cache type for `prefer-connect` streams and switch the cache type used in the endpoint code?

We already do have logic that switches between the streaming and non-streaming cache types in this endpoint based on query params, so it seems cleaner to keep all similar routing logic in that one place.

That way there wouldn't need to be extra complexity added to this cache type of the need to call in to other cache types.

The cache is already so complex I'd have a strong preference not to add yet another layer of caching/complexity into this cache type by having one cache Fetch rely on watching a different cache entity.

I guess one reason it's more complex than expected is that I was imagining that we would introduce a whole new topic for this mode. If we did that we'd end up with just another cache type with simple materializer logic on the client side and no ordering/watching two independent streams complexity.

Was there a reason you went away from that approach? This could technically be a little less work for servers on every catalog update but I don't think that is super significant as it's only constant work not per-client. The cost of the extra complexity in the client seems worse to me. But perhaps I'm missing some difficulty with that approach?

rboyer:
_(Summarizing reply from chat elsewhere)_

Originally I took a crack at creating another streaming topic of the union of the existing two streams. When it came time to actually synthesize the view off of the new event stream realized I would have to be doing two separate view synthesizations and switching between them (which this PR has to do anyway in its current form).

That coupled with the having to duplicate all of the events into a different topic all of the time for the rare time someone exercised the prefer-connect code path seemed very wasteful.

This current proposed solution just repurposes two topic/streams/views that already exist to toggle between them dynamically without having to invest a lot of changes deeper in the implementation. This variation is actually less complicated than all of the duplication required on both sides of the stream to do it the new-topic way.

Since the not-so-simple materializer logic on the consuming side requires following along with all of the events coming in and building a not-connect and a yes-connect state of the world off of the incoming events all the time so that when that last proxy is perhaps deregistered all of the things its presence hid from you are suddenly returned.

The further I got into all of the event rewriting and de-multiplexing at the view side the more wary I got that any of it was going to be correct and/or maintainably correct going forward

But if i stepped back ultimately this prefer-connect feature boils down to:

- receive data from both topics
- collect all of the data
- figure out which topic's data should provide the answer

If you do it with a new topic that is:

- duplicate all events destined for either topic, but relabel them
- subscribe to that single topic
- build two views and switch between them at request time

In the PR its:

- change nothing about topics
- subscribe to both topics separately (has the added benefit of being shareable for folks making plain health or connect queries for that on the agent)
- switch between the views at request time

So structurally it's very similar, but with far less net-new code involved.

pierrecdn:
Hi,
Does something prevent this from being reviewed / merged?

pierrecdn:
Hi there,
No news regarding this feature?

jmurret:

PR #9960, opened 2021-03-31
changelog: fixup release-note formatting

mikemorris:
So these changelog entries will be parsed successfully.

TODO from chatting with @i0rek - could we catch formatting issues like this in the CI changelog check instead of omitting entries silently when running `changelog-build` at release time?


i0rek (approved):
LGTM

i0rek:
We could run changelog-build with the PR as the rang and check if there is output. That at least covers the single entry case. Problems with multiple entries in a single file are not caught.

PR #9963, opened 2021-04-01
changelog: Auth Method feature release

kaxcode:


.changelog/9617.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+ui: ACLs now has a Auth Methods view
```
johncowen, line 2:
```suggestion
ui: ACLs now has an Auth Methods view
```

johncowen (approved):
I wonder if we should just put this as one changelog item? There wasn't a release with just one of these two additions, so we could just say `ui: Read-only ACL Auth Methods view` or something - both commit logs will still be in git history. I don't mind either way so LGTM - I'll leave it up to you.

Issue #9929, opened 2021-03-25
Update documentation that Service.Address in the service registration can be an IP or a string


ltagliamonte-dd:
Consul 1.8.8
consul server is recording in its catalog the following service:
```
cat <<EOF > payload.json
{
  "Node": "foobar",
  "Address": "192.168.10.10",
  "Service": {
    "ID": "deploy",
    "Service": "deployment-service-web",
    "Address": "192.168.1.400+what",
    "Port": 8000
  }
}
EOF

curl --request PUT --data @payload.json http://127.0.0.1:8500/v1/catalog/register

```

IP: `192.168.1.400+what` is not a valid IP.
<img width="470" alt="Screen Shot 2021-03-25 at 1 33 09 PM" src="https://user-images.githubusercontent.com/51684360/112539827-afc87100-8d6e-11eb-9a00-cd6ae5cafed9.png">

ltagliamonte-dd:
I chased the problem and the bug appears to be here:
https://github.com/hashicorp/consul/blob/master/ipaddr/ipaddr.go#L38-L40

jsosulska:
Hi @ltagliamonte-dd ,

Thanks for catching this. Would you mind creating a PR for the change, and adding some tests (like in your original post)? Feel free to tag me in it, and I'll give it a shot :)

ltagliamonte-dd:
@jsosulska to be host i briefly tried to "fix the code" but this [check](https://github.com/hashicorp/consul/blob/master/ipaddr/ipaddr.go#L24) from my code understanding it completely wrong
and it breaks even the consul server bootstrap.

jsosulska:
Thanks for taking a look at this @ltagliamonte-dd . I'll dig into it a bit as well. If the logic for validation lives in a downstream dependent library, we may need to make additional changes. I'll follow up on this one by start of next week :) Feel free to update this ticket with any information you may come across.

ltagliamonte-dd:
so i gave another read at the code and the IsAny just checks that isn't a ANY address.
So there is a total lack in IP validation. I opened https://github.com/hashicorp/consul/pull/9965

jsosulska:
Hi @ltagliamonte-dd ,
Bringing over the conversation from #9965 

> i don't know how the doc update works if you have some doc i can look into improving the doc.

[...]

PR #9965, opened 2021-04-02
add IP validation

ltagliamonte-dd:
Validate service IP.
Fixes https://github.com/hashicorp/consul/issues/9929

### DIFF
--- a/.changelog/9965.txt
+++ b/.changelog/9965.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+catalog: Validate service IP [[GH-9929](https://github.com/hashicorp/consul/issues/9929)]
+```
--- a/agent/consul/catalog_endpoint.go
+++ b/agent/consul/catalog_endpoint.go
@@ -108,6 +108,11 @@ func servicePreApply(service *structs.NodeService, authz acl.Authorizer) error {
 		return fmt.Errorf("Invalid service address")
 	}
 
+	// Check the service address to be a valid IP address
+	if !ipaddr.IsValid(service.Address) {
+		return fmt.Errorf("Invalid service address. Not a valid IP")
+	}
+
 	var authzContext acl.AuthorizerContext
 	service.FillAuthzContext(&authzContext)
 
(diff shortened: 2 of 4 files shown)


ltagliamonte-dd:
I had a look at the failing tests, they are failing because the registration request don't contain the `Address` field.
Do you want me to fix all the tests? Or `Address` is not a required field?

dnephin:
Thank you for contributing this PR!  We discussed this issue a bit internally and it sounds like the reason that this field allows any string is because hostnames can be used instead of IP addresses. I'm not sure how common it is to use hostnames in this field, but changing that would likely break someone.

There may be some way to warn when the address is not in an expected format, but it seems like the changes in this PR can not be merged as-is. Maybe we can discuss the problem and some options for fixing it on the linked github issue?

ltagliamonte-dd:
Thank you for the reply @dnephin, If consumed via the http API makes sense, I don't think that the DNS interface will work i can't find any CNAME support in [here](https://www.consul.io/docs/discovery/dns)

If hostnames can be used i don't know how can we do any sort of validation, at least I think we should call it out in the doc.

Since this field is a free text, do you think would be safer to "escape it" in order to protect from any malicious injection? I'm using sql terminology just for example here.

dnephin:
Documenting that the field can be an IP or a hostname would be good. We might also want some tests with a hostname. From what I've seen, most of the tests use an IP (probably because that is way easier to test).

I'm not sure how we would escape it. Maybe url encoding, but we definitely don't want to encode `.` or `:` as those are necessary in IP addresses. I don't think any injection is possible in Consul, but maybe something consuming the API would need to take care to encode the responses before it reads them?

ltagliamonte-dd:
i don't know who the doc update work if you have some doc i can look into improving the doc.

About the encoding/injection maybe still something to call out in doc... free test fields can always be used as threat vectors

jsosulska:
Thank you for your PR!
Replied in [thread](https://github.com/hashicorp/consul/issues/9929#issuecomment-824426560).
I will be closing this and updating the issue to reflect a docs update.

PR #9152, opened 2020-11-11
use streaming backend for connect service health

dnephin:
Should be merged after #9671

Enables the connect streaming topic.
Enables the use of the streaming cache in connect.
Also fixes a bug in the snapshot for the connect topic.

PR #9967, opened 2021-04-02
api: ensure v1/health/ingress/:service endpoint works properly when streaming is enabled

rboyer:
The streaming cache type for service health has no way to handle `v1/health/ingress/:service` queries as there is no equivalent topic that would return the appropriate data.

This PR ensures that attempts to use this endpoint will use the old cache-type for now so that they return appropriate data when streaming is enabled.

Manual backport to 1.9.x https://github.com/hashicorp/consul/pull/9968

### DIFF
--- a/agent/health_endpoint.go
+++ b/agent/health_endpoint.go
@@ -219,14 +219,19 @@ func (s *HTTPHandlers) healthServiceNodes(resp http.ResponseWriter, req *http.Re
 		return nil, nil
 	}
 
-	useStreaming := s.agent.config.UseStreamingBackend && args.MinQueryIndex > 0
+	useStreaming := s.agent.config.UseStreamingBackend && args.MinQueryIndex > 0 && !args.Ingress
 	args.QueryOptions.UseCache = s.agent.config.HTTPUseCache && (args.QueryOptions.UseCache || useStreaming)
 
 	if args.QueryOptions.UseCache && useStreaming && args.Source.Node != "" {
 		return nil, BadRequestError{Reason: "'near' query param can not be used with streaming"}
 	}
 
-	out, md, err := s.agent.rpcClientHealth.ServiceNodes(req.Context(), args)
+	healthClient := s.agent.rpcClientHealth
+	if args.Ingress {
+		healthClient = s.agent.rpcClientHealthNoStreaming
+	}
+
+	out, md, err := healthClient.ServiceNodes(req.Context(), args)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


agent/health_endpoint.go
```diff
@@ -219,14 +219,19 @@ func (s *HTTPHandlers) healthServiceNodes(resp http.ResponseWriter, req *http.Re
...
 
 	if args.QueryOptions.UseCache && useStreaming && args.Source.Node != "" {
 		return nil, BadRequestError{Reason: "'near' query param can not be used with streaming"}
 	}
 
-	out, md, err := s.agent.rpcClientHealth.ServiceNodes(req.Context(), args)
+	healthClient := s.agent.rpcClientHealth
+	if args.Ingress {
+		healthClient = s.agent.rpcClientHealthNoStreaming
+	}
```
dnephin, line 232: The purpose of the `rpcclient` is to abstract this logic from the different places that use it.  The DNS server also makes these queries and would not be fixed by this change.

I think what needs to happen is restoring the `CacheNameConnect` (or maybe `CacheNameNoConnect` instead) field the to rpc client. It was removed in https://github.com/hashicorp/consul/pull/9152/commits/1a764553c028be1078373fb51e71c71c4e87802a because we no longer needed it, but I guess we still do for this case.

dnephin (changes requested):
Thank you for making this change! I think this roughly works, but moving it into the `rpcclient/health.Client` will ensure it is fixed in the right place for all callers.

I think when we chatted about it I suggested making the change in this spot, but now I realize that because we're having to switch the underlying cache, that isn't the case.

agent/health_endpoint.go
rboyer: :+1:

rboyer: Updated

agent/health_endpoint_test.go
```diff
@@ -1418,13 +1418,22 @@ func TestHealthConnectServiceNodes(t *testing.T) {
...
+	t.Run("no streaming", func(t *testing.T) {
+		testHealthIngressServiceNodes(t, ` rpc { enable_streaming = false } use_streaming_backend = false `)
+	})
+	t.Run("cache with streaming", func(t *testing.T) {
+		testHealthIngressServiceNodes(t, ` rpc { enable_streaming = true } use_streaming_backend = true `)
+	})
+}
+
+func testHealthIngressServiceNodes(t *testing.T, agentHCL string) {
+	t.Helper()
```
dnephin, line 1430:
```suggestion
```

This function contains all the logic for the test, so I think this is a "test case" function not a test helper function. Using `t.Helper` in a test case does not help us identify the case, because we already know which case is which from the name. It can actually make the failures harder to debug since we won't get the line number for the failed assertion, and it's not always obvious which line failed.

dnephin (approved):
LGTM with the one suggestion

PR #9969, opened 2021-04-05
ui:  Reformat MaxTokenTTL to sortBy as an integer

kaxcode:
🐛 MaxTokenTTL comes in as a string from endpoint. We installed Duration.js package to reformat MaxTokenTTL as an integer and use the integer format to sortBy.

* No changelog since this feature has not been released.

### DIFF
--- a/ui/packages/consul-ui/app/models/auth-method.js
+++ b/ui/packages/consul-ui/app/models/auth-method.js
@@ -1,5 +1,7 @@
 import Model, { attr } from '@ember-data/model';
 import { or } from '@ember/object/computed';
+import Duration from '@icholy/duration';
+import { computed } from '@ember/object';
 
 export const PRIMARY_KEY = 'uid';
 export const SLUG_KEY = 'Name';
@@ -22,4 +24,10 @@ export default class AuthMethod extends Model {
   @attr('number') ModifyIndex;
   @attr() Datacenters; // string[]
   @attr() meta; // {}
+
+  @computed('MaxTokenTTL')
+  get TokenTTL() {
+    const d = new Duration(this.MaxTokenTTL);
+    return d.milliseconds();
+  }
 }
--- a/ui/packages/consul-ui/app/models/auth-method.js
+++ b/ui/packages/consul-ui/app/models/auth-method.js
@@ -1,6 +1,6 @@
 import Model, { attr } from '@ember-data/model';
 import { or } from '@ember/object/computed';
-import Duration from '@icholy/duration';
+import parse from 'parse-duration';
 import { computed } from '@ember/object';
 
 export const PRIMARY_KEY = 'uid';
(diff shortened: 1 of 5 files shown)


ui/packages/consul-ui/app/models/auth-method.js
```diff
@@ -22,4 +24,10 @@ export default class AuthMethod extends Model {
   @attr('number') ModifyIndex;
   @attr() Datacenters; // string[]
   @attr() meta; // {}
+
+  @computed('MaxTokenTTL')
+  get TTL() {
```
johncowen, line 29: Could we use a name that still references the fact that its the TTL of the token that is created by the auth method that we are talking about here, not the auth-method?

ui/packages/consul-ui/app/models/auth-method.js
```diff
@@ -1,5 +1,7 @@
 import Model, { attr } from '@ember-data/model';
 import { or } from '@ember/object/computed';
+import Duration from '@icholy/duration';
```
johncowen, line 3: Did you figure out what made the golang_compatible branch of this package golang compatible?

ui/packages/consul-ui/package.json
```diff
@@ -182,5 +182,8 @@
       "lib/commands",
       "lib/startup"
     ]
+  },
+  "dependencies": {
+    "@icholy/duration": "^5.1.0"
```
johncowen, line 187: Could we add this to `devDependencies` instead of `dependencies`?

johncowen (commented):
Looks good! I left a couple of questions but nothing major

ui/packages/consul-ui/app/models/auth-method.js
kaxcode: renamed to be `TokenTTL`

ui/packages/consul-ui/app/models/auth-method.js
kaxcode: no, it looks the same in both branches. The master branch does have newer updates.

johncowen: Hmm weird 🤷 

I'm kinda wondering whether we should use this one instead of Duration.js?

https://github.com/jkroso/parse-duration

Seems to be far more well used that Duration.js (100,000s of weekly downloads vs 1,000s) and slightly more featureful and simpler, what do you think?

It'd be a tiny switch the parser inside your computed property.

kaxcode: Updated to use `parse-duration`.

kaxcode:
@johncowen package update made. Ready for review again 😄

johncowen (approved):
Awesome, looks great, thanks for switching that out!

PR #9973, opened 2021-04-06
connect: add toggle to globally disable wildcard outbound network access when transparent proxy is enabled

rboyer:
This adds a new config entry kind "cluster" with a single special name "cluster" where this can be controlled.

TODO
- [ ] docs

### DIFF
--- a/agent/proxycfg/snapshot.go
+++ b/agent/proxycfg/snapshot.go
@@ -5,8 +5,9 @@ import (
 	"fmt"
 	"sort"
 
-	"github.com/hashicorp/consul/agent/structs"
 	"github.com/mitchellh/copystructure"
+
+	"github.com/hashicorp/consul/agent/structs"
 )
 
 // TODO(ingress): Can we think of a better for this bag of data?
@@ -59,6 +60,9 @@ type configSnapshotConnectProxy struct {
 	// intentions.
 	Intentions    structs.Intentions
 	IntentionsSet bool
+
+	ClusterConfig    *structs.ClusterConfigEntry
+	ClusterConfigSet bool
 }
 
 func (c *configSnapshotConnectProxy) IsEmpty() bool {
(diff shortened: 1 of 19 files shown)


agent/structs/config_entry_cluster_oss.go
```diff
@@ -0,0 +1,5 @@
+package structs
+
+func (e *ClusterConfigEntry) validateEnterpriseMeta() error {
```
rboyer, line 3: The enterprise version of this method will be lifted from the current version of `ProxyConfigEntry`

agent/proxycfg/snapshot.go
```diff
@@ -5,8 +5,9 @@ import (
 	"fmt"
 	"sort"
 
-	"github.com/hashicorp/consul/agent/structs"
 	"github.com/mitchellh/copystructure"
+
+	"github.com/hashicorp/consul/agent/structs"
```
freddygv, line 10: Should this have moved?

agent/xds/listeners_test.go
```diff
@@ -2,21 +2,23 @@ package xds
...
 	"bytes"
+	"path/filepath"
+	"sort"
+	"testing"
+	"text/template"
+	"time"
+
 	envoy_listener_v3 "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
+	testinf "github.com/mitchellh/go-testing-interface"
+	"github.com/stretchr/testify/require"
```
freddygv, line 13: Should these have moved?

command/config/write/config_write_test.go
```diff
@@ -8,11 +8,12 @@ import (
 
 	"github.com/hashicorp/consul/agent/structs"
 
+	"github.com/mitchellh/cli"
```
freddygv, line 11: Should these have moved?

freddygv (approved):
This LGTM, thanks!

Only had a few minor comments about imports moving around.

agent/proxycfg/snapshot.go
rboyer: Yeah I picked this tweak up from @dnephin for invoking `goimports`:
```
let g:go_fmt_options = {
    \ 'goimports': '-local github.com/hashicorp/consul',
    \ }
```

PR #9976, opened 2021-04-06
Fixup centralized upstream config resolution

freddygv:
This is a bit of an omnibus PR fixing primarily these issues:

- Centrally configured upstreams should not have ports validated, since the port will be empty.
- Update how we know whether the request was from an old client, since checking for explicit upstreamIDs won't be correct in transparent proxy mode when no explicit upstreams are configured.
- Ensure that upstream defaults can be applied to all upstreams from intentions, rather than only merging them against explicit upstreams.
- Ensure that overriding mesh gateway mode for disco chain watches works in transparent proxy mode.
- Ensure synthetic upstreams get cleaned up in local registrations

### DIFF
--- a/agent/proxycfg/state.go
+++ b/agent/proxycfg/state.go
@@ -807,6 +807,7 @@ func (s *state) handleUpdateConnectProxy(u cache.UpdateEvent, snap *ConfigSnapsh
 				wildcardSID := structs.NewServiceID(structs.WildcardSpecifier, structs.WildcardEnterpriseMeta())
 				defaults, ok := snap.ConnectProxy.UpstreamConfig[wildcardSID.String()]
 				if ok {
+					u = defaults
 					cfgMap = defaults.Config
 					snap.ConnectProxy.UpstreamConfig[svc.String()] = defaults
 				}
@@ -823,7 +824,18 @@ func (s *state) handleUpdateConnectProxy(u cache.UpdateEvent, snap *ConfigSnapsh
 				)
 			}
 
-			err = s.watchDiscoveryChain(snap, cfg, svc.String(), svc.Name, svc.NamespaceOrDefault())
+			meshGateway := s.proxyCfg.MeshGateway
+			if u != nil {
+				meshGateway = meshGateway.OverlayWith(u.MeshGateway)
+			}
+			watchOpts := discoveryChainWatchOpts{
+				id:          svc.String(),
+				name:        svc.Name,
+				namespace:   svc.NamespaceOrDefault(),
+				cfg:         cfg,
+				meshGateway: meshGateway,
+			}
+			err = s.watchDiscoveryChain(snap, watchOpts)
 			if err != nil {
 				return fmt.Errorf("failed to watch discovery chain for %s: %v", svc.String(), err)
 			}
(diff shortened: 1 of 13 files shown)


agent/structs/structs.go
```diff
@@ -1153,6 +1153,11 @@ func (s *NodeService) Validate() error {
 				"Proxy.DestinationServiceName must be non-empty for Connect proxy "+
 					"services"))
 		}
+		if strings.TrimSpace(s.Proxy.DestinationServiceName) == WildcardSpecifier {
```
dnephin, line 1156: For my own understanding, what is the reason for trimming spaces here? I don't think I've seen that done in other places.

agent/structs/connect_proxy_config.go
```diff
@@ -333,6 +333,9 @@ func (u *Upstream) Validate() error {
 	if u.DestinationName == "" {
 		return fmt.Errorf("upstream destination name cannot be empty")
 	}
+	if u.DestinationName == WildcardSpecifier && !u.CentrallyConfigured {
```
rboyer, line 336: Why gate this on CentrallyConfigured? A service name of `*` isn't going to result in SNI stuff in xDS that functions so it should be safe to prohibit all the time.

rboyer, line 336: So this can be wildcard when centrally configured?

dnephin (commented):
Nice! I had a quick read to learn a bit about the transparent proxy implementation. I think it looks good, but I will leave the approval to someone with more context.  I did have one question.

agent/proxycfg/state.go
```diff
@@ -333,6 +333,11 @@ func (s *state) initWatchesConnectProxy(snap *ConfigSnapshot) error {
 	for i := range s.proxyCfg.Upstreams {
 		u := s.proxyCfg.Upstreams[i]
 
+		// Store defaults keyed under wildcard so they can be applied to centrally configured upstreams
+		if u.DestinationName == structs.WildcardSpecifier {
+			snap.ConnectProxy.UpstreamConfig[u.DestinationID().String()] = &u
```
rboyer, line 338: Did you mean to have a `continue` here?

agent/proxycfg/state.go
```diff
@@ -807,6 +807,7 @@ func (s *state) handleUpdateConnectProxy(u cache.UpdateEvent, snap *ConfigSnapsh
 				wildcardSID := structs.NewServiceID(structs.WildcardSpecifier, structs.WildcardEnterpriseMeta())
 				defaults, ok := snap.ConnectProxy.UpstreamConfig[wildcardSID.String()]
 				if ok {
+					u = defaults
```
rboyer, line 810: Is this already a value type or should you clone these defaults every time to prevent any incidental bleedover?

agent/service_manager.go
```diff
@@ -326,7 +326,7 @@ func makeConfigRequest(bd BaseDeps, addReq AddServiceRequest) *structs.ServiceCo
 		// Also if we have any upstreams defined, add them to the request so we can
 		// learn about their configs.
 		for _, us := range ns.Proxy.Upstreams {
-			if us.DestinationType == "" || us.DestinationType == structs.UpstreamDestTypeService {
+			if us.DestinationType == "" || us.DestinationType == structs.UpstreamDestTypeService && !us.CentrallyConfigured {
```
rboyer, line 329: Could you please add some extra parentheses here? I usually find it helps me not have to go look up operator precedence rules every time to ensure I'm looking at an intended grouping instead of an accidental bug.

agent/structs/connect_proxy_config.go
freddygv: Yep, since the upstream defaults are stored until a wildcard name.

agent/structs/structs.go
freddygv: It was copied from the empty string check above, going to remove it since we're checking against "*".

agent/proxycfg/state.go
freddygv: Yep, fixed.

agent/service_manager.go
freddygv: This was resolved by not needing to exclude centrally configured upstreams since they don't get merged into the original NodeService

agent/proxycfg/state.go
freddygv: `u` and `defaults` are pointers. What do you mean by bleedover? Modifying the upstream config for the defaults after we assign it to `u`?

rboyer: Yeah.

rboyer (approved):
LGTM

PR #9977, opened 2021-04-07
streaming: Grpc tuning

banks:
These tuning parameters match the gRPC transport used by streaming to have the same keepalive behaviour our Yamux-based RPCs do.

While doing scaling experiments we saw issues with TCP connections dropping and not being detected until an event delivery is attempted since the default keepalive is only sent every 2 hours in gRPC.

This also tunes the retry backoff during streaming materializer failures. These parameters may have improved recovery times after TCP drops some, but in fact gRPC itself has it's own backoff policy that comes into play here too which was not investigated further. These settings seem reasonable though, especially given that gRPC is already doing backoff on reconnections lower down in the stack for the vast majority of cases where streaming clients would error.

### DIFF
--- a/agent/cache-types/streaming_health_services.go
+++ b/agent/cache-types/streaming_health_services.go
@@ -113,9 +113,13 @@ func newMaterializer(
 		Logger: deps.Logger,
 		Waiter: &retry.Waiter{
 			MinFailures: 1,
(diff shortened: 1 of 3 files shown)


dnephin (approved):
LGTM

PR #9978, opened 2021-04-07
cache: fix bug where TTLs were ignored leading to leaked memory in client agents

banks:
While testing Consul at scale we found this issue where agent caches might never expire.

It occurs for cache types that cancel requests and so return from `Fetch` when `Close` is called.

Currently, only streaming cache types have this behaviour but it is a supported behaviour for other future cache types too. The race here occurs very frequently (i.e. virtually all the time):

 1. Value expires in cache, the entry is removes and then `Close` is called on the cache type. 
 2. This immediately cancels a context which causes the goroutine running the current backround `Fetch` call to return. 
 3. The cache handles Fetch return in `fetch` method and immediately re-inserts the resulting error into the cache entry that was just removed, touches the TTL and lets it live again!

Never removing the cache type even after clients have stopped requesting the data effectively causes a permanent memory leak which could be exacerbated by heavy churn in watches or tokens used during a client agent's lifetime. 

Even if we move streaming types to not use the cache, it's worth fixing this so that `Close` behaviour is safe to use in future cache types without this leak.

### DIFF
--- a/.changelog/9978.txt
+++ b/.changelog/9978.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cache: fix a bug in the client agent cache where streaming could potentially leak resources. [[GH-9978](https://github.com/hashicorp/consul/pulls/9978)].
+```
--- a/.changelog/9978.txt
+++ b/.changelog/9978.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-cache: fix a bug in the client agent cache where streaming could potentially leak resources. [[GH-9978](https://github.com/hashicorp/consul/pulls/9978)].
+cache: fix a bug in the client agent cache where streaming could potentially leak resources. [[GH-9978](https://github.com/hashicorp/consul/pull/9978)].
 ```
--- a/agent/cache/cache.go
+++ b/agent/cache/cache.go
@@ -723,6 +723,22 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
 		// Set our entry
 		c.entriesLock.Lock()
 
+		if _, ok := c.entries[key]; !ok {
+			// This entry was evicted during our fetch. DON'T re-insert it or fall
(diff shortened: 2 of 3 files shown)


.changelog/9978.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+cache: fix a bug in the client agent cache where streaming could potentially leak resources. [[GH-9978](https://github.com/hashicorp/consul/pulls/9978)].
```
banks, line 2:
```suggestion
cache: fix a bug in the client agent cache where streaming could potentially leak resources. [[GH-9978](https://github.com/hashicorp/consul/pull/9978)].
```

agent/cache/cache.go
```diff
@@ -723,6 +723,22 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
 		// Set our entry
 		c.entriesLock.Lock()
 
+		if _, ok := c.entries[key]; !ok {
```
dnephin, line 726: I wonder if it is possible for a new entry to be added to the `entries` map while the fetch is happening. In that case we would still leak a goroutine right? We could compare the entry here, maybe by using the `Expiry` which is a pointer.

```suggestion
		if cachedEntry, ok := c.entries[key]; !ok || cachedEntry.Expiry != entry.Expiry {
```

Probably not a big deal since the follow up PR makes this effectively impossible.

dnephin (approved):
LGTM

agent/cache/cache.go
banks: 🤔 I _think_ if we get a double race and the entry expires and then is legitimately recreated by a new client `Get` during fetch then the behavior is still correct (i.e. not leaky) because by definition the client that recreated the entry is still interested in keeping it alive right? So it's not really being leaked in that case but rather kept because there is a client that cares about it still?

But maybe I'm missing a possible edge case?

PR #9979, opened 2021-04-07
cache: Fix bug where connection errors can cause early cache expiry

banks:
Based on #9978.

This fixes a bug in the agent cache where TTL expiry times are only updated when successful updates are received.

There are two different ways this can cause items to be prematurely evicted from the cache:

 1. If the agent is disconnected from servers for at least one TTL period, the waiting client `Get` requests will go round their loop seeing only errors and not updating the expiry time even though they are still interested (i.e. the value is still "alive").
 2. If no updates are sent for one entire TTL then the waiting Get's will never see a valid update response and so never touch the expiry time meaning the value will be evicted.

Currently we don't run into this because until streaming, all cache-types had long TTLs of 72 hours.

It's also self-recovering in that the active clients will just cause the cache value to be refetched anyway even though it was evicted. For typical RPC-based cache types the cost is minimal but for leaf certificates, streaming and other possible future types the eviction causes connections to drop and data to be lost and recomputed which negatively effects performance or requires unnecessary work.

For streaming, this causes dropped connections which then take some time to reconnect based on the timing of the client which can mean delayed event delivery.

### DIFF
--- a/.changelog/9978.txt
+++ b/.changelog/9978.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cache: fix a bug in the client agent cache where streaming could potentially leak resources. [[GH-9978](https://github.com/hashicorp/consul/pulls/9978)].
+```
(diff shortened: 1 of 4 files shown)


.changelog/9978.txt
```diff
@@ -1,3 +1,3 @@
 ```release-note:bug
-cache: fix a bug in the client agent cache where streaming could potentially leak resources. [[GH-9978](https://github.com/hashicorp/consul/pulls/9978)].
+cache: fix a bug in the client agent cache where streaming could potentially leak resources. [[GH-9978](https://github.com/hashicorp/consul/pull/9978)].
```
banks, line 2: I fixed this in the upstream PR branch too and reset base but it's still showing here 🤷

dnephin (approved):
LGTM, this seems like a really important fix.

PR #9980, opened 2021-04-07
command: when generating envoy bootstrap configs to stdout do not mix informational logs into the json

### LINKED ISSUE #9921: consul connect envoy -bootstrap can output logs intertwined with envoy bootstrap json
#### Overview of the Issue

`consul connect envoy -bootstrap` can output this log line right before the envoy json:

```
Defaulting to grpc port = 8502
{
  "admin": {
    "access_log_path": "/dev/null",
    "address": {
      "socket_address": {
        "address": "127.0.0.1",
        "port_value": 19000
```

#### Reproduction Steps

create svc.hcl
```
services {
  id = "test-id"
  name = "test"
  port = 8080
}

services {
  name = "test-sidecar-proxy"
  kind = "connect-proxy"
  port = 20000
  proxy {
    destination_service_name = "test"
    destination_service_id = "test-id"
  }
}
```

Start consul

```
consul agent -bootstrap -server -config-file svc.hcl -hcl 'connect { enabled = true} ' -data-dir /tmp/consul-data
```

In another tab run bootstrap command and note the `Defaulting to grpc port = 8502`:

```
consul connect envoy -proxy-id test-sidecar-proxy -bootstrap
    Defaulting to grpc port = 8502
{
  "admin": {
    "access_log_path": "/dev/null",
```

Now when you start envoy with the output it will error since it's invalid json.

Consul 1.9.4. 

Looks like it was introduced here: https://github.com/hashicorp/consul/commit/1a8ffec6a7fcb1f7bca67240e03a58db3f4ab6da

Perhaps we should consider a `-bootstrap-file` flag since it's going to be tough to know that this one command needs you to be careful about what you output to stdout.

rboyer:
Fixes #9921


command/connect/envoy/envoy.go
```diff
@@ -24,6 +24,12 @@ import (
 )
 
 func New(ui cli.Ui) *cmd {
+	if bui, ok := ui.(*cli.BasicUi); ok {
+		// Prevent any incidental use of Output/Info from writing to Stdout and
+		// corrupting the generated bootstrap file.
+		bui.Writer = os.Stderr
```
freddygv, line 30: Wouldn't this mean that this log message would go to Stderr now?
https://github.com/hashicorp/consul/blob/master/command/connect/envoy/envoy.go#L383

That seems a bit odd

rboyer: One of the semi-"standard" things about log messages on unix is that informational messages go to stderr and only the real output of a program goes to stdout so it can be used in a pipeline. It certainly doesn't seem out of place.

dnephin (approved):
I think this fix works as a quick bug fix. Maybe a longer term solution is to pass a struct with two `io.Writer` around to each command, instead of the `UI` interface. Then each command can choose to either use the UI struct, or do something more directly. 

That would allow us to slowly port commands away from `cli.UI`  where it makes sense, without having to change everything all at once.

command/connect/envoy/envoy.go
```diff
@@ -24,6 +24,12 @@ import (
 )
 
 func New(ui cli.Ui) *cmd {
+	if bui, ok := ui.(*cli.BasicUi); ok {
```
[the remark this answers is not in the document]
rboyer: I originally started going down this short list and I scrapped that and did it completely differently. I did not make an attempt to shrink the interface down mostly because that likely should happen uniformly across the whole CLI rather than in a one-off like this.

command/connect/envoy/envoy.go
```diff
@@ -47,9 +36,6 @@ type cmd struct {
 	http   *flags.HTTPFlags
 	help   string
 	client *api.Client
-	// DirectOut defaults to os.stdout but is a property to allow capture during
-	// tests to have more useful output.
-	directOut io.Writer
```
rboyer, line 52: If you don't mix up informational messages with output then the underlying OutputWriter construct is already doing this for us.

dnephin (approved):
Nice! LGTM

command/connect/envoy/envoy.go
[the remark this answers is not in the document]
dnephin: Sounds good, I like the approach you took here

command/connect/envoy/testdata/CONSUL_HTTP_ADDR-with-https-scheme-enables-tls.golden
```diff
@@ -190,3 +190,4 @@
     }
   }
 }
+
```
rboyer, line 193: This is a side effect of the internal plumbing for `ui.Output()` which seemed fine since trailing whitespace isn't semantically relevant for json.

Issue #9311, opened 2020-12-02
consul connect fails to boot envoy when health checks are exposed and ipv6 is disabled


timrosenblatt:
#### Overview of the Issue

Envoy refuses to create exposed health check listeners when ipv6 is disabled. Here’s a relevant line from the logs:
 
```
[2020-11-19 18:25:07.670][144682][warning][config] [external/envoy/source/common/config/grpc_subscription_impl.cc:101] gRPC config for type.googleapis.com/envoy.api.v2.Listener rejected: Error adding/updating listener(s) exposed_path_grpchealthv1HealthCheck:0.0.0.0:51700: IPv6 addresses are not supported on this machine: [::1]:0
```
 
The only place that seems to use an IPv6 address in the listener config is in the source_prefix_ranges:

``` 
"filter_chain_match": {
"source_prefix_ranges": [
  {
   "address_prefix": "127.0.0.1",
   "prefix_len": 8
  },
  {
   "address_prefix": "::1",
   "prefix_len": 128
  },
  {
   "address_prefix": "45.32.96.231",
   "prefix_len": 32
  }
]
}
```
 
It looks like there’s currently not a way to skip including the IPv6 address in [listeners.go](https://github.com/hashicorp/consul/blob/v1.8.4/agent/xds/listeners.go#L538)

There are no workarounds.

#### Reproduction Steps

The problem should be reproducible within a CentOS7 container with [ipv6 disabled](https://linuxhint.com/disable_ipv6_centos7/)

### Operating system and Environment details

We are using Consul 1.8, Envoy 1.14.4, and RHEL7.

PR #9987, opened 2021-04-07
Remove kube-dns resolution since clusterip will be a tagged addr

freddygv:
Upcoming changes to consul-k8s will sync the ClusterIP as a tagged address.

This PR adds every tagged address as a possible address to match on for transparent proxies.

### DIFF
--- a/agent/xds/listeners.go
+++ b/agent/xds/listeners.go
@@ -192,6 +192,12 @@ func (s *Server) listenersFromSnapshotConnectProxy(cInfo connectionInfo, cfgSnap
 				PrefixLen:     &wrappers.UInt32Value{Value: pfxLen},
 			})
 		}
+
+		// The match rules are stable sorted to avoid draining if the list is provided out of order
+		sort.SliceStable(ranges, func(i, j int) bool {
+			return ranges[i].AddressPrefix < ranges[j].AddressPrefix
+		})
+
 		filterChain.FilterChainMatch = &envoy_listener_v3.FilterChainMatch{
 			PrefixRanges: ranges,
 		}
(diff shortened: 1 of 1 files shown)


agent/xds/listeners.go
```diff
@@ -150,35 +150,23 @@ func (s *Server) listenersFromSnapshotConnectProxy(cInfo connectionInfo, cfgSnap
...
-					continue
+				for _, tagged := range e.Node.TaggedAddresses {
+					if tagged != "" {
+						uniqueAddrs[tagged] = struct{}{}
+					}
 				}
-				for _, addr := range resolved {
-					if _, ok := uniqueAddrs[addr]; !ok {
-						uniqueAddrs[addr] = struct{}{}
+				for _, tagged := range e.Service.TaggedAddresses {
```
freddygv, line 167: Not sure if this is susceptible to this issue if a tagged address is IPv6:
https://github.com/hashicorp/consul/issues/9311

I don't _think_ it should be, since it's an address in a filter chain match rather than a listener address.

rboyer: You might want to test it out in a quick throwaway VM. That's what I had to do in order to get the kernel to hard-disable the mere possibility of ipv6

rboyer (commented):
When you populate `ranges := make([]*envoy_core_v3.CidrRange, 0)` you should deterministically sort the final results before sending them back to envoy so that envoy doesn't see a different iteration order every time it asks for these listeners (which will cause a listener drain event)

rboyer (approved):
LGTM assuming the ipv6 thing doesn't require changes.

freddygv:
Tested listener creation with IPv6 filter chain match in a VM with IPv6 disabled and it worked fine.

PR #8802, opened 2020-10-01
lib/retry - extract a new package from lib/retry.go

dnephin:
This PR extracts a new package (`lib/retry`) from `lib`. Large packages like `lib` suffer from weak cohesion. They provide many unrelated things making them hard to use.

This PR also changes `Waiter` to reduce the number of exported methods, and make it easier to use.

Export fields so that the caller doesn't have to define constants that are only ever used once. For example:

```go
// it's not really clear what all of these numbers are doing
retry.NewWaiter(1, 0, 10*time.Minute, retry.NewJitter(25))
```

Some callers may choose to make these constants, but they are likely only ever used once. Instead we can have callers set fields on the struct. The field names document how the values will be used:

```go
w := &Waiter{
	MinFailures: 1,
	MinWait:     0,
	MaxWait:     10*time.Minute,
	Jitter:      retry.NewJitter(25),
}
```

There are many examples of this pattern in the stdlib, https://golang.org/pkg/os/exec/#Cmd is one of them.

The new type exports three methods: `Wait(context.Context) error`, `Reset()`, and `Failures() int`.

This new package is used in #8809

PR #9988, opened 2021-04-07
lib/retry: allow jitter to exceed max wait

dnephin:
I changed this in https://github.com/hashicorp/consul/pull/8802#pullrequestreview-500779357 because exceeding the MaxWait seemed wrong, but as other have pointed out, that behaviour is probably correct.

When multiple waiters hit the max value, we don't want them to converge, so restore the behaviour of allowing jitter to exceed max, and document it.


banks (approved):
Thanks LGTM!

PR #9344, opened 2020-12-08
ui: Change URI helper to a template based approach

johncowen:
This moves our `uri` helper to use a the template renderer we already have for rendering URLs.

Reasons behind this:

1. We had a bug a while ago where we were trying to construct a URL with an undefined nspace, this was resulting in a URL like `//dc/something`. This approach has a strict mode that will warn you whilst in development with a clear error if any values are `undefined`. This should prevent any such bugs in the future (we re-use the functionality we already had for rendering URLs without a strict mode).
2. Our template based datasources now look pretty much exactly the same as our javascript based ones, which additionally means it's easy to find places where we are using a specific data sources by searching through the project. This approach generally just looks easier to see whats happening rather than our previous concat approach.

Right now, there are dev time errors being thrown for a missing `@protocol` argument within the topology tab, so would be good to catch up to see what the 'default' value would be for a protocol before just assuming its an empty string.

PR #9989, opened 2021-04-08
ui: Return empty string  protocol for upstream/downstream metrics request

kaxcode:
Protocol is not a required field to make the metrics request for upstream/downstream. Quick fix with empty string.

### DIFF
--- a/ui/packages/consul-ui/app/components/topology-metrics/stats/index.hbs
+++ b/ui/packages/consul-ui/app/components/topology-metrics/stats/index.hbs
@@ -7,7 +7,7 @@
         dc=@dc
         endpoint=@endpoint
         service=@service
-        protocol=@protocol
+        protocol=(or @protocol '')
       )
     }}
     @onchange={{action 'statsUpdate'}}


kaxcode:
@johncowen does this need a changelog?

johncowen (approved):
Excellent!

Re: changelog - nope not right now. The error that currently appears in console is set to be dev time only - more info here https://github.com/hashicorp/consul/pull/9344

> We had a bug a while ago where we were trying to construct a URL with an undefined nspace, this was resulting in a URL like //dc/something. This approach has a strict mode that will warn you whilst in development with a clear error if any values are undefined. This should prevent any such bugs in the future (we re-use the functionality we already had for rendering URLs without a strict mode).

As the protocol argument was never used in the API calls this affects, there is no user facing change here.

Moving forwards though we should avoid these forced errors (even though they are dev time only) - theres one more like this I'm aware of, but I'll fix that up at some point.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/400958.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 84064f972ed2dc8cd44569151b22d6a3e1ba7a93 onto `release/1.9.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/400958)

PR #9991, opened 2021-04-09
Handle FSM.Apply errors in raftApply

### LINKED ISSUE #9970: Handle FSM.Apply errors in Server.raftApply to avoid missing the error
Almost everywhere we call `Server.raftApply` we have this same code to unpack an error out of the response:

```go
if respErr, ok := resp.(error); ok {
    err = respErr
}
```

This is because FSM errors are reported from `ApplyFuture.Response` and only raft errors come from `ApplyFuture.Error`.

In some cases we completely ignore the `response` from the future, which would result in missing any errors returned by the `FSM.Apply`. Generally we want to log or return these errors so that the user knows something failed.

By handling the response error in `Server.raftApply` we can ensure that any error from the apply is always returned in the `error` response value, and we remove the need to check the response in every caller. We may be able to completely remove the response value, if the only use is unpacking this error.

dnephin:
Fixes #9970

Previously we were inconsistently checking the response for errors. This
PR moves the response-is-error check into `raftApply`, so that all callers
can look at only the error response, instead of having to know that
errors could come from two places.

This should expose a few more errors that were previously hidden because
some calls to `raftApply` were ignoring the response return value.

Also handle errors more consistently. In some cases we would log the
error before returning it. This can be very confusing because it can
result in the same error being logged multiple times. Instead return
a wrapped error.

Also fixes a bug with `canRetry` and chunking errors. Previously it never would
have retried on those errors because it was looking at the wrong arg.

### DIFF
--- a/agent/consul/acl_endpoint.go
+++ b/agent/consul/acl_endpoint.go
@@ -13,16 +13,17 @@ import (
 
 	"github.com/armon/go-metrics"
 	"github.com/armon/go-metrics/prometheus"
+	"github.com/hashicorp/go-bexpr"
+	"github.com/hashicorp/go-hclog"
+	memdb "github.com/hashicorp/go-memdb"
+	uuid "github.com/hashicorp/go-uuid"
+
 	"github.com/hashicorp/consul/acl"
 	"github.com/hashicorp/consul/agent/consul/authmethod"
 	"github.com/hashicorp/consul/agent/consul/state"
 	"github.com/hashicorp/consul/agent/structs"
 	"github.com/hashicorp/consul/lib"
 	"github.com/hashicorp/consul/lib/template"
-	"github.com/hashicorp/go-bexpr"
-	"github.com/hashicorp/go-hclog"
-	memdb "github.com/hashicorp/go-memdb"
-	uuid "github.com/hashicorp/go-uuid"
 )
 
 const (
@@ -250,12 +251,8 @@ func (a *ACL) BootstrapTokens(args *structs.DCSpecificRequest, reply *structs.AC
 
 	req.Token.SetHash(true)
 
-	resp, err := a.srv.raftApply(structs.ACLBootstrapRequestType, &req)
+	_, err = a.srv.raftApply(structs.ACLBootstrapRequestType, &req)
 	if err != nil {
-		return err
-	}
-
-	if err, ok := resp.(error); ok {
 		return err
 	}
 
@@ -729,17 +726,13 @@ func (a *ACL) tokenSetInternal(args *structs.ACLTokenSetRequest, reply *structs.
 		req.ProhibitUnprivileged = true
 	}
 
-	resp, err := a.srv.raftApply(structs.ACLTokenSetRequestType, req)
+	_, err = a.srv.raftApply(structs.ACLTokenSetRequestType, req)
 	if err != nil {
 		return fmt.Errorf("Failed to apply token write request: %v", err)
 	}
 
 	// Purge the identity from the cache to prevent using the previous definition of the identity
 	a.srv.acls.cache.RemoveIdentity(tokenSecretCacheID(token.SecretID))
-
-	if respErr, ok := resp.(error); ok {
-		return respErr
-	}
 
 	// Don't check expiration times here as it doesn't really matter.
 	if _, updatedToken, err := a.srv.fsm.State().ACLTokenGetByAccessor(nil, token.AccessorID, nil); err == nil && updatedToken != nil {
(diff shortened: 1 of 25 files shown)


agent/consul/acl_endpoint.go
```diff
@@ -729,18 +726,14 @@ func (a *ACL) tokenSetInternal(args *structs.ACLTokenSetRequest, reply *structs.
...
 	if err != nil {
 		return fmt.Errorf("Failed to apply token write request: %v", err)
 	}
 
 	// Purge the identity from the cache to prevent using the previous definition of the identity
 	a.srv.acls.cache.RemoveIdentity(tokenSecretCacheID(token.SecretID))
 
-	if respErr, ok := resp.(error); ok {
-		return respErr
-	}
```
dnephin, line 742: This is one of a few places where we were treating these errors differently. As far as I can tell, if the FSM.Apply returns an error, we shouldn't consider the operation a success, so I suspect this new behaviour is actually more correct.

We keep the identity in the cache if the FSM.Apply errors, because the operation did not complete.

There are a bunch more of these cases in this file.

agent/consul/rpc.go
```diff
@@ -729,28 +730,34 @@ func (s *Server) keyringRPCs(method string, args interface{}, dcs []string) (*st
...
 }
 
-// raftApplyWithEncoder is used to encode a message, run it through raft,
-// and return the FSM response along with any errors. Unlike raftApply this
-// takes the encoder to use as an argument.
-func (s *Server) raftApplyWithEncoder(t structs.MessageType, msg interface{}, encoder raftEncoder) (interface{}, error) {
+// raftApplyWithEncoder encodes a message, and then calls raft.Apply with the
+// encoded message. Returns the FSM response along with any errors. If the
+// FSM.Apply response is an error it will be returned as the error return
+// value with a nil response.
```
dnephin, line 755: I updated the docs here   and deprecated the old `raftApply`.

agent/consul/rpc.go
```diff
@@ -789,17 +796,19 @@ func (s *Server) raftApplyWithEncoder(t structs.MessageType, msg interface{}, en
...
 			return nil, errors.New("unknown type of response back from chunking FSM")
 		}
-		// Return the inner wrapped response
-		return chunkedSuccess.Response, nil
+		resp = chunkedSuccess.Response
 	}
 
+	if err, ok := resp.(error); ok {
+		return nil, err
+	}
```
dnephin, line 811: This is the core of the change.

agent/consul/session_endpoint.go
```diff
@@ -160,10 +160,6 @@ func (s *Session) Apply(args *structs.SessionRequest, reply *string) error {
 		s.srv.clearSessionTimer(args.Session.ID)
 	}
 
-	if respErr, ok := resp.(error); ok {
-		return respErr
-	}
```
dnephin, line 165: Another case where we handled the errors differently.

agent/consul/rpc.go
```diff
@@ -789,17 +796,19 @@ func (s *Server) raftApplyWithEncoder(t structs.MessageType, msg interface{}, en
 			// apply function. Downstream client code expects to see any error
 			// from the FSM (as opposed to the apply itself) and decide whether
 			// it can retry in the future's response.
-			return ErrChunkingResubmit, nil
+			return nil, ErrChunkingResubmit
```
dnephin, line 799: Hmm, this `return nil, ErrChunkingResubmit` part of the change I need to look at. The handling of this in `canRetry` is strange.

dnephin: Ya, the handling of `ErrChunkingResubmit` seems completely broken. It's only ever referenced in one place here: https://github.com/hashicorp/consul/blob/master/agent/consul/rpc.go#L530-L539

But looking at the callers to `canRetry`, there's no way that an error is ever going to be passed as the `args` argument:

* https://github.com/hashicorp/consul/blob/a02245b75af59b8af8545639c79f2dd571b0e174/agent/consul/client.go#L289 - args is going to be the `Request` struct for the RPC
* https://github.com/hashicorp/consul/blob/a02245b75af59b8af8545639c79f2dd571b0e174/agent/consul/rpc.go#L607 - `info` is of type `RPCInfo`, not an error.


So it looks like this was always broken.

dnephin: I pushed another commit which should fix this problem.

agent/consul/rpc.go
```diff
@@ -534,9 +535,8 @@ func canRetry(args interface{}, err error) bool {
 		return true
 	}
 
-	// If we are chunking and it doesn't seem to have completed, try again
-	intErr, ok := args.(error)
-	if ok && strings.Contains(intErr.Error(), ErrChunkingResubmit.Error()) {
+	// If we are chunking and it doesn't seem to have completed, try again.
+	if err != nil && strings.Contains(err.Error(), ErrChunkingResubmit.Error()) {
```
dnephin, line 539: This is the buf fix for the second bug I noticed.

rboyer: Is there a test case that exercises this ErrChunkingResubmit flow to verify this is a-ok?

rboyer (approved):
LGTM (with small nonblocking comments)

agent/consul/rpc.go
dnephin: There must not be, because it was previously broken. I'm not sure how chunking works, I'll see if I can write a test for it.

dnephin: I'll work on a test in a follow up. I think there is more cleanup to do around `canRetry` to prevent similar bugs in the future.

PR #9999, opened 2021-04-12
Replace TransparentProxy flag with ProxyMode and config struct

freddygv:
This PR replaces the original boolean used to configure transparent
proxy mode. It was replaced with a string mode that can be set to:

- "": Empty string is the default for when the setting should be
defaulted from other configuration like config entries.
- "direct": Direct mode is how applications originally opted into the
mesh. Proxy listeners need to be dialed directly.
- "transparent": Transparent mode enables configuring Envoy as a
transparent proxy. Traffic must be captured and redirected to the
inbound and outbound listeners.

This PR also adds a struct for transparent proxy specific configuration.
Initially this is not stored as a pointer. Will revisit that decision
before GA.


agent/dns_test.go
```diff
@@ -6688,7 +6688,7 @@ func TestDNS_trimUDPResponse_NoTrim(t *testing.T) {
 		},
 	}
 
-	cfg := loadRuntimeConfig(t, `data_dir = "a" bind_addr = "127.0.0.1"`)
+	cfg := loadRuntimeConfig(t, `node_name = "test" data_dir = "a" bind_addr = "127.0.0.1"`)
```
rboyer, line 6691: stray changes?

agent/xds/clusters.go
```diff
@@ -56,10 +56,11 @@ func (s *Server) clustersFromSnapshotConnectProxy(cfgSnap *proxycfg.ConfigSnapsh
 	}
 	clusters = append(clusters, appCluster)
 
-	// In TransparentProxy mode there needs to be a passthrough cluster for traffic going to destinations
+	// In transparent proxy mode there needs to be a passthrough cluster for traffic going to destinations
 	// that aren't in Consul's catalog.
-	// TODO (freddy): Add cluster-wide setting that can disable this cluster and restrict traffic to catalog destinations.
-	if cfgSnap.Proxy.TransparentProxy {
+	if cfgSnap.Proxy.Mode == structs.ProxyModeTransparent &&
+		!cfgSnap.ConnectProxy.ClusterConfig.TransparentProxy.CatalogDestinationsOnly {
```
rboyer, line 62: Does this need rebasing?

api/agent.go
```diff
@@ -118,11 +118,12 @@ type AgentServiceConnectProxyConfig struct {
 	DestinationServiceID   string                 `json:",omitempty"`
 	LocalServiceAddress    string                 `json:",omitempty"`
 	LocalServicePort       int                    `json:",omitempty"`
+	Mode                   ProxyMode              `json:",omitempty"`
+	TransparentProxy       TransparentProxyConfig `json:",omitempty"`
```
rboyer, line 122: Reminder: `omitempty` doesn't apply to empty structs. This means that because this is not a pointer type every single API result is going to say `..., "TransparentProxy": {}, ...`

agent/dns_test.go
freddygv: It was unrelated, so I put it on its own commit. This test fails locally on a mac because of how the node name is defaulted.

agent/xds/clusters.go
freddygv: Nope, the listeners were gated on the ClusterConfig, but the passthrough cluster wasn't until now.

api/agent.go
freddygv: Yep, was just matching the other fields. Same applies to MeshGateway and Expose below.

Left a comment about this in the main PR description, going to leave this as not-a-pointer for now and going to revisit that for GA. Want to explore some alternatives like using custom JSON marshaling function.

rboyer (approved):
LGTM

PR #10000, opened 2021-04-12
Remove zero-value validation of upstream cfg structs

freddygv:
The zero value of these flags was already being excluded in the xDS
generation of circuit breaker/outlier detection config.

See: `makeThresholdsIfNeeded` and `ToOutlierDetection`.

This validation actually leads to users having to specify every field 
in the structs if any one of them is defined.


markan (approved):
Looks good to me

PR #10002, opened 2021-04-12
ui: TProxy Service Mesh Viz Beta Changes

kaxcode:
[Demo Link 
](https://consul-ui-staging-kplii2rrz-hashicorp.vercel.app/ui/dc1/services/service-0/topology#CONSUL_METRICS_PROXY_ENABLE=1;CONSUL_METRICS_PROVIDER=prometheus
)
✨ Created warning banner for permissive intentions (wildcard-intentions) and default-allow intentions
permissive_intention
default-allow

✨ Created topology-metrics/source-type component (label). Used in each Card.
label-defined

✨ Created consul/transparent-proxy component (label). Used in service instance show page. 
mode_label

✨ Created warning banner for not explicitly defining an upstream intention
undefine_banner

✨ Created a dashed line with a warning icon for card with upstream intention not explicitly defined
dashed_line

✨ Created a *(All Services) card as a placeholder for all the services our main service can talk to because of permissive intentions
Screen Shot 2021-04-13 at 12 20 14 PM

🤡 Updated service-topology mock data with new attributes: TransparentProxy, Source, and DefaultAllow
🤡 Updated connect mock data with new attribute: Mode
♻️ Refactored the topology-metrics/card component to be in its own folder structure


ui/packages/consul-ui/app/components/informed-action/layout.scss
```diff
@@ -2,6 +2,9 @@
   & {
     min-width: 190px;
   }
+  &.documentation {
```
kaxcode, line 5: I tried making the width longer but it broke the delete confirmation. I used class 'documentation' since it could be used in the future again for "Documentation" links.

ui/packages/consul-ui/app/components/topology-metrics/down-lines/index.hbs
```diff
@@ -90,6 +90,13 @@
       @item={{item}}
       @oncreate={{action @oncreate item @service}}
     />
+  {{else if (and item.Intention.Allowed (not item.TransparentProxy) (eq item.Source 'specific-intention'))}}
```
kaxcode, line 93: This is only in the down-lines because only downstreams would have "not defined upstreams explicitly."

ui/packages/consul-ui/app/components/topology-metrics/source-type/index.hbs
```diff
@@ -0,0 +1,6 @@
+<span
+  class="topology-metrics-source-type"
+  {{tooltip "This connection was defined in a proxy registration."}}
```
kaxcode, line 3: Here I need the `tooltip` modifier but I can't use the modifier with translations. LMK if you know how to make both work. @johncowen

ui/packages/consul-ui/mock-api/v1/internal/ui/service-topology/_
```diff
@@ -52,10 +52,26 @@ ${
     }
     fake.seed(index);
 
+
+    // Randomly make permissive intentions
+    let defaultAllow = false;
```
kaxcode, line 57: Here I'm trying to create a 33% chance of false/false, false/true, and true/false.

ui/packages/consul-ui/mock-api/v1/internal/ui/service-topology/_
```diff
@@ -70,11 +86,14 @@ ${
       "ChecksPassing":${fake.random.number({min: 1, max: env('CONSUL_CHECK_COUNT', fake.random.number(10))})},
       "ChecksWarning":${fake.random.number({min: 0, max: env('CONSUL_CHECK_COUNT', fake.random.number(10))})},
       "ChecksCritical":${fake.random.number({min: 0, max: env('CONSUL_CHECK_COUNT', fake.random.number(10))})},
+      "Source": "${fake.helpers.randomize(['proxy-registration', 'default-allow', 'wildcard-intention'])}",
```
kaxcode, line 89: I purposely left out 'specific-intention' because only downstreams would have "upstreams not explicitly defined."

ui/packages/consul-ui/translations/en-us.yaml
```diff
@@ -184,6 +184,46 @@ components:
             name: Precedence
             asc: Ascending
             desc: Descending
+    transparent-proxy: Transparent Proxy
```
kaxcode, line 187: @johncowen this file is getting really big now. 😬

ui/packages/consul-ui/app/components/topology-metrics/source-type/index.hbs
johncowen: You can nest helpers within things in handlebars with round brackets:

```hbs
{{!-- modifiers --}}
{{tooltip (t "the.identifier")}}

{{!-- block helpers --}}
{{#let (t "the.identifier") as |text|}}{{/let}}

{{!-- inline helpers --}}
{{another-helper (t "the.identifier")}}
```

https://guides.emberjs.com/v3.22.0/components/helper-functions/#toc_nested-helpers

ui/packages/consul-ui/app/templates/dc/services/show/topology.hbs
```diff
@@ -25,6 +25,15 @@ as |route|>
...
       {{/if}}
+      {{#if topology.DefaultAllow}}
+        <TopologyMetrics::Notice::DefaultAllow />
+      {{/if}}
+      {{#if topology.WildcardIntention}}
+        <TopologyMetrics::Notice::WildcardIntention />
+      {{/if}}
+      {{#if topology.undefinedIntention}}
+        <TopologyMetrics::Notice::UndefinedIntention />
+      {{/if}}
```
johncowen, line 36: Components/partials are usually 'write something once, and use multiple times' I think here we've done the reverse in that we have written multiple components and used each one a single time.

Is there a way we can do this less repetitively, especially if all the copy is done with `t`?

ui/packages/consul-ui/mock-api/v1/internal/ui/service-topology/_
johncowen: If its an equal split probability wise, why not just `({min: 1, max: 3})`, or I think you could also use

```javascript
.randomize([
  {
   defaultAllow: true,
   wildcardIntention: false
  }, {
   defaultAllow: false,
   wildcardIntention: true
  }, {
   defaultAllow: ...,
   wildcardIntention: ...
  }
])
```
if you like that better?

Is there anyway we can set up the mock data so we are guaranteed to get one of every type by default, right now I can see allow and deny but no undefined.

ui/packages/consul-ui/app/templates/dc/services/instance.hbs
```diff
@@ -58,6 +58,9 @@ as |route|>
             </h1>
             <Consul::ExternalSource @item={{item}} />
             <Consul::Kind @item={{item}} @withInfo={{true}} />
+            {{#if proxy.ServiceProxy.Mode}}
```
johncowen, line 61: Does any value for `proxy.ServiceProxy.Mode` mean TransparentProxy? Or just truthy? I think you mentioned also, but don't forget to add this to the mocks I don't think I see that here.

ui/packages/consul-ui/translations/en-us.yaml
johncowen: Yeah ideally the translations for components needs to be split up to live in the components folders with all the rest of the component code. There are other things we can do to split it up also, is that something you can look at?

ui/packages/consul-ui/mock-api/v1/internal/ui/service-topology/_
```diff
@@ -52,10 +52,26 @@ ${
     }
     fake.seed(index);
 
+
+    // Randomly make permissive intentions
+    let defaultAllow = false;
+    let wildwardIntention = false;
```
johncowen, line 58: `wildward` > `wildcard` in a few places below also.

ui/packages/consul-ui/app/components/topology-metrics/index.js
```diff
@@ -66,6 +66,21 @@ export default class TopologyMetrics extends Component {
...
   }
 
+  get upstreams() {
+    const items = [...get(this.args.topology, 'Upstreams')];
+    if (get(this.args.topology, 'DefaultAllow') || get(this.args.topology, 'WildcardIntention')) {
+      items.push({
+        Name: '*(All Services)',
+        Datacenter: '',
+        Namespace: '',
+        Intention: {
```
kaxcode, line 76: I had to leave this Intention block or else it creates a Deny icon.

ui/packages/consul-ui/app/components/topology-metrics/index.js
```diff
@@ -104,5 +117,17 @@ export default class TopologyMetrics extends Component {
 
     // Set Upstream Cards Positioning points
     this.upLines = this.drawUpLines(upCards);
+
+    // Calculate viewBox dimensions
+    this.downView = document.getElementById('downstream-lines').getBoundingClientRect();
+
+    const upstreamLines = document.getElementById('upstream-lines').getBoundingClientRect();
+    const upstreamColumn = document.getElementById('upstream-column').getBoundingClientRect();
+    this.upView = {
```
kaxcode, line 126: I had to make some height adjustments. I'll probably have to rework on this again for the GA dropdown.

ui/packages/consul-ui/app/components/topology-metrics/card/index.scss
```diff
@@ -0,0 +1,81 @@
...
+  display: block;
+  color: $gray-700;
+  overflow: hidden;
+  background-color: $white;
+  border-radius: $decor-radius-100;
+  border: 1px solid $gray-200;
+  p {
+    padding: 12px 12px 0 12px;
+    font-size: 16px;
+    font-weight: 600;
```
johncowen, line 16: Could these use our font variables?

ui/packages/consul-ui/app/components/topology-metrics/source-type/index.scss
```diff
@@ -0,0 +1,4 @@
+.topology-metrics-source-type {
+  margin: 6px 0 6px 12px;
+  display: table;
```
johncowen, line 3: This jumps out as strange to me? I don't think we use this for any other badge type things?

johncowen (commented):
Thanks for the notes! I left a few more, I think there are 4 main things this could do with before merging:

1. `undefined` is a bit of a weird name to use in javascript with it being so important in javascript, I can imagine various ways bugs could slip in here if something is accidentally undefined and that gets cast to a string. I think we should change this to something else so it doesn't end up getting stuck in here.
2. It would be googd to try and avoid the multiple single use components for the notices here if possible.
3. Add enough mocks so you can see all different types by default (and add the `mode` to the proxy endpoints if thats where they go)
4. Choose a solution for specifying paragraphs in the translations.

I think the rest could be done here, or at any point moving forwards.

Nice work!

ui/packages/consul-ui/translations/en-us.yaml
kaxcode: I tried splitting things into folder translation/components/transparent-proxy but it didn't work. Do I have to add the new files in settings?

ui/packages/consul-ui/app/components/topology-metrics/card/index.hbs
```diff
@@ -1,4 +1,4 @@
-<a class="card"
+<a class="topology-metrics-card"
```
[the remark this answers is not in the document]
kaxcode: 💯 I'm going skipping doing that in this PR to keep things simple.

ui/packages/consul-ui/app/components/topology-metrics/card/index.scss
```diff
@@ -0,0 +1,78 @@
+#upstream-container .topology-metrics-card:not(:last-child),
+#downstream-container .topology-metrics-card:not(:last-child) {
+  margin-bottom: 8px;
+}
+#upstream-container .topology-metrics-card,
+#downstream-container .topology-metrics-card {
```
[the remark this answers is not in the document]
kaxcode: Let me make a note to do this later. For GA I'm going to have to deal with the SVG drawings again and can touch up on this.

ui/packages/consul-ui/translations/en-us.yaml
johncowen: Oh sorry, we'd have to do some broccoli magic to make the component folder thing work, there are other ways to split it up though, take a look at the docs there's some guidance there.

ui/packages/consul-ui/app/components/topology-metrics/notice/default-allow.hbs
```diff
@@ -0,0 +1,23 @@
...
+as |notice|>
+    <notice.Header>
+    <h3>
+      {{t "components.consul.topology-metrics.notice.default-allow.header"}}
+    </h3>
+  </notice.Header>
+  <notice.Body>
+    <p>
+      {{t "components.consul.topology-metrics.notice.default-allow.body"}}
+    </p>
```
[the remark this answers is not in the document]
kaxcode: Wait, that looks the same as what I have 🤔 Did you mean using the `p0` `p1` ?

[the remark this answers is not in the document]
johncowen: We need to be able to define paragraphs of text in the translations file, not in the template (in these cases). Ideally we'd use markdown in the translations file, but I guess we should add something later. So for now I'd put the `<p>` tags here in the translations file.

I know this seems wrong (putting HTML in the translations file), but it would make more sense once it's markdown. To try help, imagine if we had a piece of text that has 2 paragraphs, you wouldn't want to specify each paragraph as a different translation key, which means we have to somehow specify simple formatting in the translations file. Let me know if that doesn't make sense and we can go over in a bit more detail.

ui/packages/consul-ui/app/components/topology-metrics/notice/undefined-intention.hbs
```diff
@@ -0,0 +1,23 @@
...
+    </h3>
+  </notice.Header>
+  <notice.Body>
+    <p>
+     {{t "components.consul.topology-metrics.notice.undefined-intention.body"}}
+    </p>
+  </notice.Body>
+  <notice.Footer>
+    <p>
+      <a href="{{env 'CONSUL_DOCS_URL'}}/connect/registration/service-registration#upstreams" rel="noopener noreferrer" target="_blank">
```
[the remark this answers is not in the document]
kaxcode: The `root=(env 'CONSUL_DOCS_URL')` didn't work. Maybe we could go over this together.

ui/packages/consul-ui/app/templates/dc/services/show/topology.hbs
kaxcode: I tried. It all comes down to making the footer dynamic. I ran into some issues and just left them split up.

ui/packages/consul-ui/app/templates/dc/services/instance.hbs
kaxcode: Mode is true/false. This `Mode` attribute is a new addition to the /connect endpoint. I've added that to the PR notes.

ui/packages/consul-ui/app/components/topology-metrics/source-type/index.scss
kaxcode: Yea, this is the first bad we have in the sandwiched between text. It needed a bit more spacing in the bottom and sides.

ui/packages/consul-ui/app/components/topology-metrics/notice/default-allow.hbs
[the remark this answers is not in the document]
kaxcode: In this scenario I just have 1 `<p>`. Use still want to use `default-allow.body.p0`?

ui/packages/consul-ui/app/components/informed-action/layout.scss
kaxcode: Removed 😄

ui/packages/consul-ui/translations/en-us.yaml
```diff
@@ -184,6 +184,52 @@ components:
...
+      popover:
+        l7:
+          header: Layer 7 permissions
+          body: Certain HTTP request info must be identified.
+          action: View
+        deny:
+          header: Connection Denied
+          body:
+            isExact: Change the action of this intention to allow.
+            notExact: Add an intention that allows these two services to connect.
```
johncowen, line 222: Can we prefer kebabcase for the translation identifiers/keys? (Obviously can't do this for `models:` properties)

Kinda tangent from this - its seems like we are starting to use different words for like `exact`/`explicit`/`specific`/`defined`/`wildcard` along with things like `hasExact`/`isExact`/`notExact` etc etc.

The typical 'naming is hard' thing - but in this case I think it's probably important we get a handle on this sooner rather than later. What do you think?

ui/packages/consul-ui/app/components/topology-metrics/popover/index.hbs
```diff
@@ -4,22 +4,21 @@
...
     <:body>
       <p>
         {{#if @item.Intention.HasExact}}
-          Change the action of this intention to allow.
+          {{t "components.consul.topology-metrics.popover.deny.body.isExact"}}
         {{else}}
-          Add an intention that allows these two services to connect.
+          {{t "components.consul.topology-metrics.popover.deny.body.notExact"}}
         {{/if}}
       </p>
```
johncowen, line 23: I couldn't reply to the previous comment about `<p>` tags as the code has gone now, so just moving it here:

> In this scenario I just have 1 <p>. Use still want to use default-allow.body.p0?

In this case:

```hbs
<:body>
  {{t "body" htmlSafe=true}}
</:body>
```

```yaml
body: |
  <p>Paragraph one</p>
  <p>Paragraph <em>two</em></p>
```

At some point in the future if we can do:

```yaml
body: |
  Paragraph one
  Paragraph _two_
```

..and shim a markdown parser into the `t` helper, that would be much better - but that's probably something for a later date.

johncowen (commented):
I gave this another pass and added a bunch of stuff, but theres quite a lot here that we should talk over. I don't want to block you on things though, so if we can get the tests passing here I'd be happy to approve as is, and then use the comments here as a basis for a follow up PR.

Did you think about whether you wanted to add some additional tests in here for what you've added?

kaxcode:
@johncowen all tests have passed and I've moved this PR out of draft. Ready for last ✅

johncowen (approved):
Nice, I know you are waiting on backend for this but 👍 from me!

ui/packages/consul-ui/app/templates/dc/services/instance.hbs
johncowen: I just spoke to @freddygv 

> Mode can be "", "direct", or "transparent"

ui/packages/consul-ui/app/models/topology.js
```diff
@@ -11,7 +12,30 @@ export default class Topology extends Model {
...
   @attr('boolean') FilteredByACLs;
+  @attr('boolean') TransparentProxy;
+  @attr('boolean') DefaultAllow;
+  @attr('boolean') WildcardIntention;
   @attr() Upstreams; // Service[]
   @attr() Downstreams; // Service[],
   @attr() meta; // {}
+
+  @computed('Upstreams', 'Downstreams')
+  get undefinedIntention() {
```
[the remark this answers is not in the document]
kaxcode: I'm using `undefinedIntention` not just `undefined`. `hasIntention` is the opposite of what I'm trying to figure out in this compute.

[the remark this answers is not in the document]
johncowen: In that case I think this would be preferable: `!item.hasIntention`/`(not item.hasIntention)`.

I'd say 'generally' boolean things like this should be positively named (mostly with `is` or `has`), and then use either `!` or `(not)` to flip it if/when you need to. Otherwise at some point you end up writing double negatives like `!item.undefinedIntention`/`(not item.undefinedIntention)` which is like saying "doesn't doesn't have an intention" or "doesn't not have an intention", and things can start to get confusing.

ui/packages/consul-ui/app/models/topology.js
```diff
@@ -11,7 +12,30 @@ export default class Topology extends Model {
...
+        item =>
+          item.Source === 'specific-intention' && !item.TransparentProxy && item.Intention.Allowed
+      ).length !== 0;
+
+    undefinedDownstream =
+      this.Downstreams.filter(
+        item =>
+          item.Source === 'specific-intention' && !item.TransparentProxy && item.Intention.Allowed
+      ).length !== 0;
+
```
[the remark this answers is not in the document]
johncowen: If you wanted to, you don't need to do the second filter on Downstreams if the first one gives you `true`

PR #10006, opened 2021-04-13
Convert new tproxy structs in api module into ptrs

freddygv:
This way we avoid serializing these when empty. Otherwise users of the
latest version of the api submodule cannot interact with older versions
of Consul, because the api client would send keys that Consul doesn't
recognize yet.

Related: https://github.com/golang/go/issues/11939


rboyer (approved):
LGTM

PR #10009, opened 2021-04-13
dns: properly trim response when EDNS is used

### LINKED ISSUE #9945: Incorrect DNS response size
## Context:
In our infrastructure we use [coredns](https://coredns.io/) in front of consul agent for DNS protocol.

Meanwhile investigating an issue we are facing with [coredns](https://github.com/coredns/coredns/issues/4547)
I found that the problem of the issue isn't in coredns but in Consul.

When EDNS is in use and buffsize is specified at 2048 bytes, Consul response size on the wire is 2058 bytes, and causes coredns to memory overflow.
The instructions, and tcpdump captures on how to replicate the bug, are all present in the first comment of this [issue](https://github.com/coredns/coredns/issues/4547)

## Consul code:
 Looking at the code (adding a few prints over the DNS request path) it appears that the problem is in how the [maxSize](https://github.com/hashicorp/consul/blob/566efad2e7eccce8b581150b61713ea065f0b43f/agent/dns.go#L1108) is calculated when EDNS is requested:
```
before truncation 4135
before extra 2039
after extra 2039
before setEdns 2039
before wire 2050
```

the truncation of the DNS response happens before [setEdns](https://github.com/hashicorp/consul/blob/566efad2e7eccce8b581150b61713ea065f0b43f/agent/dns.go#L500) and doesn't take  into account the 11 bytes that EDNS will add to the response before the wire.
As matter of fact if I modify [maxSize](https://github.com/hashicorp/consul/blob/566efad2e7eccce8b581150b61713ea065f0b43f/agent/dns.go#L1108) to be:
```
maxSize = int(size) - 11
```
The issue on coredns size is fixed.

dnephin:
Fixes #9945

This PR refactors parts of the `DNSServer` to allow us to set the response code and trim the response in a single place. This fixes the problem of the DNS response exceeding the max size.

I notice we never called trim in `handlePtr` or `handleRecurse`. I wonder if we should be.

TODO:
* [x] changelog
* [x] add a new test specifically for this case
* [x] figure out if the last remaining TODO needs to be resolved, which might require returning the error down through the calls

### DIFF
--- a/agent/dns.go
+++ b/agent/dns.go
@@ -609,7 +609,6 @@ func (d *DNSServer) parseDatacenter(labels []string, datacenter *string) bool {
 
 var errECSNotGlobal = fmt.Errorf("ECS response is not global")
 var errNameNotFound = fmt.Errorf("DNS name not found")
-var errQueryRefused = fmt.Errorf("query refused")
 
 // errNoAnswer is used to indicate that the response should set SOA, and the
 // success response code.
@@ -876,12 +875,9 @@ func rCodeFromError(err error) int {
 	case err == nil:
 		return dns.RcodeSuccess
 	case errors.Is(err, errNoAnswer):
-		// TODO: why do we return success if the answer is empty?
 		return dns.RcodeSuccess
 	case errors.Is(err, errECSNotGlobal):
 		return rCodeFromError(errors.Unwrap(err))
-	case errors.Is(err, errQueryRefused):
-		return dns.RcodeRefused
 	case errors.Is(err, errNameNotFound):
 		return dns.RcodeNameError
 	case structs.IsErrNoDCPath(err) || structs.IsErrQueryNotFound(err):
@@ -896,7 +892,7 @@ func (d *DNSServer) nodeLookup(cfg *dnsConfig, datacenter, node string, req, res
 	// Only handle ANY, A, AAAA, and TXT type requests
 	qType := req.Question[0].Qtype
 	if qType != dns.TypeANY && qType != dns.TypeA && qType != dns.TypeAAAA && qType != dns.TypeTXT {
-		return errQueryRefused
+		return nil
 	}
 
 	// Make an RPC request
--- a/agent/dns.go
+++ b/agent/dns.go
@@ -3,6 +3,7 @@ package agent
 import (
 	"context"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"net"
 	"regexp"
@@ -476,7 +477,7 @@ func (d *DNSServer) handleQuery(resp dns.ResponseWriter, req *dns.Msg) {
 	m.Authoritative = true
 	m.RecursionAvailable = (len(cfg.Recursors) > 0)
 
-	ecsGlobal := true
+	var err error
 
 	switch req.Question[0].Qtype {
 	case dns.TypeSOA:
@@ -496,10 +497,10 @@ func (d *DNSServer) handleQuery(resp dns.ResponseWriter, req *dns.Msg) {
 		m.SetRcode(req, dns.RcodeNotImplemented)
 
 	default:
-		ecsGlobal = d.dispatch(network, resp.RemoteAddr(), req, m, maxRecursionLevelDefault)
+		err = d.dispatch(network, resp.RemoteAddr(), req, m, maxRecursionLevelDefault)
 	}
 
-	setEDNS(req, m, ecsGlobal)
+	setEDNS(req, m, !errors.Is(err, errECSNotGlobal))
 
 	// Write out the complete response
 	if err := resp.WriteMsg(m); err != nil {
@@ -601,9 +602,12 @@ func (d *DNSServer) parseDatacenter(labels []string, datacenter *string) bool {
 	}
 }
 
+var errECSNotGlobal = fmt.Errorf("ECS response is not global")
+var errNameNotFound = fmt.Errorf("DNS name not found")
+
 // dispatch is used to parse a request and invoke the correct handler.
 // parameter maxRecursionLevel will handle whether recursive call can be performed
-func (d *DNSServer) dispatch(network string, remoteAddr net.Addr, req, resp *dns.Msg, maxRecursionLevel int) bool {
+func (d *DNSServer) dispatch(network string, remoteAddr net.Addr, req, resp *dns.Msg, maxRecursionLevel int) error {
 	// By default the query is in the default datacenter
 	datacenter := d.agent.config.Datacenter
 
@@ -643,11 +647,11 @@ func (d *DNSServer) dispatch(network string, remoteAddr net.Addr, req, resp *dns
 		}
 	}
 
-	invalid := func() bool {
+	invalid := func() error {
 		d.logger.Warn("QName invalid", "qname", qName)
 		d.addSOA(cfg, resp)
 		resp.SetRcode(req, dns.RcodeNameError)
-		return true
+		return errNameNotFound
 	}
 
 	switch queryKind {
@@ -684,7 +688,7 @@ func (d *DNSServer) dispatch(network string, remoteAddr net.Addr, req, resp *dns
 			lookup.Service = queryParts[0][1:]
 			// _name._tag.service.consul
 			d.serviceLookup(cfg, lookup, req, resp)
-			return true
+			return nil
 		}
 
 		// Consul 0.3 and prior format for SRV queries
@@ -699,7 +703,7 @@ func (d *DNSServer) dispatch(network string, remoteAddr net.Addr, req, resp *dns
 
 		// tag[.tag].name.service.consul
 		d.serviceLookup(cfg, lookup, req, resp)
-		return true
+		return nil
 
 	case "connect":
 		if len(queryParts) < 1 {
@@ -721,7 +725,7 @@ func (d *DNSServer) dispatch(network string, remoteAddr net.Addr, req, resp *dns
 		}
 		// name.connect.consul
 		d.serviceLookup(cfg, lookup, req, resp)
-		return true
+		return nil
 
 	case "ingress":
 		if len(queryParts) < 1 {
@@ -743,7 +747,7 @@ func (d *DNSServer) dispatch(network string, remoteAddr net.Addr, req, resp *dns
 		}
 		// name.ingress.consul
 		d.serviceLookup(cfg, lookup, req, resp)
-		return true
+		return nil
 
 	case "node":
 		if len(queryParts) < 1 {
@@ -757,7 +761,7 @@ func (d *DNSServer) dispatch(network string, remoteAddr net.Addr, req, resp *dns
 		// Allow a "." in the node name, just join all the parts
 		node := strings.Join(queryParts, ".")
 		d.nodeLookup(cfg, datacenter, node, req, resp, maxRecursionLevel)
-		return true
+		return nil
 
 	case "query":
 		// ensure we have a query name
@@ -772,7 +776,7 @@ func (d *DNSServer) dispatch(network string, remoteAddr net.Addr, req, resp *dns
 		// Allow a "." in the query name, just join all the parts.
 		query := strings.Join(queryParts, ".")
 		d.preparedQueryLookup(cfg, network, datacenter, query, remoteAddr, req, resp, maxRecursionLevel)
-		return false
+		return errECSNotGlobal
 
 	case "addr":
 		// <address>.addr.<suffixes>.<domain> - addr must be the second label, datacenter is optional
@@ -825,7 +829,7 @@ func (d *DNSServer) dispatch(network string, remoteAddr net.Addr, req, resp *dns
 				resp.Answer = append(resp.Answer, aaaaRecord)
 			}
 		}
-		return true
+		return nil
 	default:
 		return invalid()
 	}
@@ -1905,6 +1909,7 @@ func (d *DNSServer) resolveCNAME(cfg *dnsConfig, name string, maxRecursionLevel 
 		resp := &dns.Msg{}
 
 		req.SetQuestion(name, dns.TypeANY)
+		// TODO: handle error response
 		d.dispatch("udp", nil, req, resp, maxRecursionLevel-1)
 
 		return resp.Answer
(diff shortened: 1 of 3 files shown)


agent/dns.go
```diff
@@ -838,23 +863,33 @@ func (d *DNSServer) trimDomain(query string) string {
...
+// rCodeFromError return the DNS Error code an error
+func rCodeFromError(err error) int {
+	switch {
+	case err == nil:
 		return dns.RcodeSuccess
-	}
-	if structs.IsErrNoDCPath(err) || structs.IsErrQueryNotFound(err) {
+	case errors.Is(err, errNoAnswer):
+		// TODO: why do we return success if the answer is empty?
+		return dns.RcodeSuccess
```
dnephin, line 873: I don't think we need to resolve this TODO, but I added it because I don't fully understand this case. The existing comments say:

> If the answer is empty and the response isn't truncated, return not found

However, we were never setting the response code to `RcodeNameError`. Maybe the existing comment is misleading? The one thing we did for this case was set the SOA response, and we still do that, so this preserves existing behaviour.

Does anyone know why this we set SOA for this case, but not the response code?

agent/dns_test.go
```diff
@@ -545,6 +545,7 @@ func TestDNS_NodeLookup_CNAME(t *testing.T) {
 
 	m := new(dns.Msg)
 	m.SetQuestion("google.node.consul.", dns.TypeANY)
+	m.SetEdns0(8192, true)
```
dnephin, line 548: This test started to fail because we are now trimming every response instead of only some of them. So I had to `setEdns0` to increase the size of the response.

I guess this is probably expected behaviour, and is now more correct? I'm honestly not sure.

agent/dns.go
```diff
@@ -1896,6 +1900,7 @@ func (d *DNSServer) resolveCNAME(cfg *dnsConfig, name string, maxRecursionLevel
 		resp := &dns.Msg{}
 
 		req.SetQuestion(name, dns.TypeANY)
+		// TODO: handle error response
 		d.dispatch("udp", nil, req, resp, maxRecursionLevel-1)
```
dnephin: This is the last TODO to resolve. Looking at it again, I guess we only ever return the `resp.Answer` here, so ignoring the error might be ok, since the error only impacts the `MsgHdr` and we won't send that anyway.

agent/dns.go
```diff
@@ -838,23 +859,33 @@ func (d *DNSServer) trimDomain(query string) string {
...
 }
 
 // nodeLookup is used to handle a node query
-func (d *DNSServer) nodeLookup(cfg *dnsConfig, datacenter, node string, req, resp *dns.Msg, maxRecursionLevel int) {
+func (d *DNSServer) nodeLookup(cfg *dnsConfig, datacenter, node string, req, resp *dns.Msg, maxRecursionLevel int) error {
 	// Only handle ANY, A, AAAA, and TXT type requests
 	qType := req.Question[0].Qtype
 	if qType != dns.TypeANY && qType != dns.TypeA && qType != dns.TypeAAAA && qType != dns.TypeTXT {
-		return
+		return errQueryRefused
```
dnephin, line 888: This is also new. I guess previously we returned some partially incomplete response? Maybe it had a success Rcode? I'm not really sure, but I'm pretty sure it was wrong.  Now we should be returning a response with `RcodeRefused`, which seems better.

agent/dns.go
```diff
@@ -838,23 +859,33 @@ func (d *DNSServer) trimDomain(query string) string {
...
-	if err == nil {
+// rCodeFromError return the DNS Error code an error
+func rCodeFromError(err error) int {
+	switch {
+	case err == nil:
 		return dns.RcodeSuccess
-	}
-	if structs.IsErrNoDCPath(err) || structs.IsErrQueryNotFound(err) {
+	case errors.Is(err, errNoAnswer):
+		// TODO: why do we return success if the answer is empty?
```
dhiaayachi, line 868: This is to avoid caching of error response for other query types. if we return an error a recursor will cache it and use it even for other query types

agent/dns_test.go
dhiaayachi: The spec state that without eDNS we should trim to 512 by default and this is what we do now, So I think we are good

agent/dns.go
dhiaayachi: not sure, in that case it mean we have the wrong query type (if I understand right). Like I explained in the other comment we should return an empty success if the Q type is wrong but the node exist.

dhiaayachi (commented):
@dnephin I added the missing test and added an error check for the CNAME.
The only part I'm not sure about is the change in the error handling change in `rCodeFromError`.
dns error are a bit tricky because of how response caching is done in recursors and also the fact that a cached response could be returned to a different type of request (different Qtype for example)

agent/dns_test.go
```diff
@@ -6658,6 +6658,47 @@ func TestDNS_PreparedQuery_AgentSource(t *testing.T) {
...
+
+	{
+		m := new(dns.Msg)
+		m.SetQuestion("foo.query.consul.", dns.TypeSRV)
+		m.SetEdns0(2048, true)
+
+		c := new(dns.Client)
+		r, _, err := c.Exchange(m, a.DNSAddr())
+		require.NoError(t, err)
+		require.Len(t, r.Answer, 53)
```
dnephin, line 6698: I think if we call `r.Pack()` we can also test that the returned byte buffer is less than 2048, which might be good to verify as well in case of the size of the records changes somehow.

From my reading of `WriteMsg` the only other thing that can get added is `Server.TsigSecret`, which it looks like we don't set right now, so `Pack` should be accurate.

dhiaayachi: yes I like the idea, I will add this

dnephin:
I'm going to rebase this PR to get the verify-ci check working, just an FYI because it will require a force push.

agent/dns.go
dnephin: Makes sense now that I've learned a bunch about DNS this week! I'll remove this TODO.

agent/dns.go
dnephin: Yup, you are definitely right. I will fix this as well.

agent/dns.go
```diff
@@ -1358,13 +1361,11 @@ func (d *DNSServer) preparedQueryLookup(cfg *dnsConfig, network, datacenter, que
...
-	d.trimDNSResponse(cfg, network, req, resp)
-
 	// If the answer is empty and the response isn't truncated, return not found
-	if len(resp.Answer) == 0 && !resp.Truncated {
-		d.addSOA(cfg, resp)
-		return
+	if len(resp.Answer) == 0 {
+		return errNoAnswer
 	}
+	return nil
```
dnephin, line 1364: Looking at this part of the code changes, I'm wondering if the handling of SOA is still correct in this branch. There is a similar branch above in `serviceLookup`.

Previously we would only set SOA if we had no answers, where as now we'll set it in both cases (when the prepared query or service lookup returns no instances, and when it returns some service instances).

@blake do you know in what cases would we not want to set the `SOA` record on the response.NS section? Only for recurse, or are there other cases where setting the SOA record would be incorrect?

agent/dns.go
```diff
@@ -1899,9 +1899,13 @@ func (d *DNSServer) resolveCNAME(cfg *dnsConfig, name string, maxRecursionLevel
...
-		// TODO: handle error response
-		d.dispatch(nil, req, resp, maxRecursionLevel-1)
 
+		err := d.dispatch(nil, req, resp, maxRecursionLevel-1)
+		rCode := rCodeFromError(err)
+		if rCode == dns.RcodeNameError || errors.Is(err, errNoAnswer) {
+			d.addSOA(cfg, resp)
+		}
+		resp.SetRcode(req, rCode)
 		return resp.Answer
```
dnephin: @dhiaayachi  I think  the challenge with this TODO is that we only return `resp.Answer`.

`SetRcode and `addSOA` change different parts of the `resp`, they don't modify the `Answer` section, so when we return `Answer` here we throw away the `Rcode` and `NS` sections that we modified.

My TODO comment was not very complete, but I think the question we need to answer is: Do we need to care about the  error returned from `dispatch` ?

If we need to care, then `resolveCNAME` needs to return either the full `resp` or some other struct to include all of that information.

dnephin: Another possibility is that `dispatch` is just the wrong abstraction here, but if that's the case it'll be more work to clean up I think.

agent/dns.go
dhiaayachi: I changed the code a bit to avoid trimming the answer after adding SOA, we only add SOA in certain conditions + no trimming happened (otherwise we will bust the response size limit).

agent/dns.go
dhiaayachi: yes you are right, I also when through the call flow and we don't really care about dispatch failing. it end up returning a nil answer which mean that at the end we won't add any extra fields.

dnephin:
I rebased this PR with the latest main branch, and removed commit bb42aea23cab822f37c33d7384efe3d52e6aa9a3 (which was causing test failures), and merged the commit+revert to cancel those out.

agent/dns.go
dnephin: Ok, I looked into this again and I think it was mostly correct. The only problem I found was that my fix in the commit with title "dns: correct rcode for qtype not supported" I returned `errNoAnswer` (which sets SOA) instead of `nil` (which doesn't). They both use the same RCode, but only one will set SOA.

I have no idea if this is correct, but at least it matches the old behaviour. I've amended the commit which introduced the problem to properly return a `nil` error when the Rcode is not supported by `nodeLookup`.

agent/dns.go
```diff
@@ -849,23 +869,30 @@ func (d *DNSServer) trimDomain(query string) string {
 	return strings.TrimSuffix(query, shorter)
 }
 
-// computeRCode Return the DNS Error code from Consul Error
-func (d *DNSServer) computeRCode(err error) int {
-	if err == nil {
+// rCodeFromError return the DNS Error code an error
```
blake, line 872:
```suggestion
// rCodeFromError return the appropriate DNS response code for a given error
```

agent/dns.go
blake: For unsupported qtypes, I believe the better response is to send `errNoAnswer` (or `errNoData` as I suggested renaming to).

https://umbrella.cisco.com/blog/nxdomain-nodata-debugging-dns-dual-stacked-hosts offers a good explanation as to why a NODATA response should be returned if resource records exist for other query types. Section 2.2 from RFC 2308 (linked in another comment) explains why its recommended to add the SOA.

> @blake do you know in what cases would we not want to set the SOA record on the response.NS section? Only for recurse, or are there other cases where setting the SOA record would be incorrect?

It took me a bit of staring at the code before I realized that response.Ns actually refers  to the the authority section of the DNS response. [[1]] :-)

After reading through a few RFCs, the SOA should definitely be omitted in negative responses for non-authoritative resources (i.e., recursive queries). It is also recommended that the SOA be added to the authority section [[2]] of a negative response (NODATA or NXDOMAIN) for resources which the the server is authoritative [[3]].

> I have no idea if this is correct,…

Based on my understanding of the linked RFCs, this seems to be the correct behavior.

[1]: https://pkg.go.dev/github.com/miekg/dns#Msg
[2]: https://datatracker.ietf.org/doc/html/rfc2181#section-7.1
[3]: https://datatracker.ietf.org/doc/html/rfc1034#section-4.3.4

agent/dns.go
```diff
@@ -610,9 +607,37 @@ func (d *DNSServer) parseDatacenter(labels []string, datacenter *string) bool {
 	}
 }
 
-// doDispatch is used to parse a request and invoke the correct handler.
+var errECSNotGlobal = fmt.Errorf("ECS response is not global")
+var errNameNotFound = fmt.Errorf("DNS name not found")
+
+// errNoAnswer is used to indicate that the response should set SOA, and the
+// success response code.
+var errNoAnswer = fmt.Errorf("no DNS Answer")
```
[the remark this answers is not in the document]
dnephin: This is much better, thank you! Definitely agree we should make the domain terminology as much as possible.

agent/dns.go
```diff
@@ -1189,14 +1189,24 @@ func trimUDPResponse(req, resp *dns.Msg, udpAnswerLimit int) (trimmed bool) {
...
+	if trimmed {
+		if cfg.EnableTruncate {
+			resp.Truncated = true
+		}
+		d.logger.Debug("DNS response too large, truncated",
+			"protocol", network,
+			"question", req.Question,
+			"records", fmt.Sprintf("%d/%d", len(resp.Answer), originalNumRecords),
+			"size", fmt.Sprintf("%d/%d", resp.Len(), originalSize),
+		)
```
dnephin, line 1211: I made a small change to this log message:
1. to always emit it when trimmer, even if we aren't setting the truncated field. I believe technically we truncate in both cases, but we only indicate it in the response when the setting is enabled (seems like strange behaviour to me)
2. removed TCP from the log message and added a protocol field to indicate tcp or udp

dnephin (commented):
@dhiaayachi your changes LGTM! Thank you for all of your help on this

dhiaayachi (approved):
LGTM!! Thanks @dnephin !!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/412032.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 1c8ac9cd4bda5b265fad4d7776f9592fa72721a0 onto `release/1.10.x` succeeded!

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 1c8ac9cd4bda5b265fad4d7776f9592fa72721a0 onto `release/1.9.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/412032)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 1c8ac9cd4bda5b265fad4d7776f9592fa72721a0 onto `release/1.8.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/412032)

PR #10013, opened 2021-04-14
Move static token resolution into the ACLResolver

mkeeler:
This moves static token into the ACLResolver. Previously it was being handled at the Agent level for resolving the agent master token.

This will ensure that these tokens can be resolved properly for audit logging purposes in enterprise.

### DIFF
--- a/.changelog/10013.txt
+++ b/.changelog/10013.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+audit-logging: (Enterprise only) Fixed an issue that resulted in usage of the agent master token or managed service provider tokens from being resolved properly.
+```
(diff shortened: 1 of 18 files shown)


dnephin (approved):
LGTM, nice cleanup as well.

I think pretty soon we'll be able use the `ACLResolver` directly from other code instead of having it as part of the `delegate` interface.

mkeeler:
@dnephin I almost tried to go down the path of putting it in the base deps. But that was going to be a much larger refactoring as currently there are a number of client/server specific interfaces used as a delegate to the resolver itself which would be a little harder to decouple. I still think we should but this is the tactical steps necessary to fix the bugs for now.

mkeeler:
Also the nomad integration test failing is a known issue and others are working on it.

PR #10015, opened 2021-04-14
connect: update centralized upstreams representation in service-defaults

rboyer:


agent/structs/config_entry.go
```diff
@@ -688,6 +708,29 @@ type UpstreamConfig struct {
...
+func (cfg UpstreamConfig) Clone() UpstreamConfig {
+	cfg2 := cfg
+
+	cfg2.Limits = cfg.Limits.Clone()
+	cfg2.PassiveHealthCheck = cfg.PassiveHealthCheck.Clone()
+
+	return cfg2
+}
+
+func (cfg *UpstreamConfig) ServiceID() ServiceID {
```
freddygv, line 720: Is this used anywhere? Maybe we could just keep `ServiceName()`

agent/structs/config_entry.go
```diff
@@ -713,15 +756,48 @@ func (cfg UpstreamConfig) MergeInto(dst map[string]interface{}) {
...
 	cfg.Protocol = strings.ToLower(cfg.Protocol)
 
 	if cfg.ConnectTimeoutMs < 0 {
 		cfg.ConnectTimeoutMs = 0
 	}
+	return nil
 }
 
-func (cfg UpstreamConfig) Validate() error {
+func (cfg UpstreamConfig) ValidateWithoutName() error {
```
freddygv, line 781: Can you add validation/normalization tests to `structs` or a TODO for them?

freddygv (approved):
LGTM, left two minor comments

agent/structs/config_entry.go
rboyer: Yes, semantically an UpstreamConfig really should only discuss `ServiceName` things, since upstreams are to "all" instances instead of just _this one_, but `seenUpstreams` logic in the flattening RPC only operates on ServiceID which is why this is here.

agent/consul/config_endpoint.go
```diff
@@ -449,23 +450,23 @@ func (c *ConfigEntry) ResolveServiceConfig(args *structs.ServiceConfigRequest, r
...
+				// usConfigs stores the opaque config map for each upstream and is keyed on the upstream's ID.
+				usConfigs = make(map[structs.ServiceID]map[string]interface{})
 			)
-			if serviceConf != nil && serviceConf.Connect != nil {
-				if serviceConf.Connect.UpstreamDefaults != nil {
-					upstreamDefaults = serviceConf.Connect.UpstreamDefaults
+			if serviceConf != nil && serviceConf.UpstreamConfig != nil {
+				for _, override := range serviceConf.UpstreamConfig.Overrides {
+					if override.Name == "" {
+						continue // skip this impossible condition
```
markan, line 463: Why is it impossible? Feels like it either should be handled (at least logged) or not tested for....

markan (approved):
Other than the minor comment above, looks good. Took me a little bit to figure out what was going on in the config_endpoint code; maybe lack of context on my part.

agent/consul/config_endpoint.go
rboyer: Earlier validation before the config entry was accepted should require the name field to be set.

rboyer: logged

agent/structs/config_entry.go
rboyer: added some tests for the untested part

PR #10016, opened 2021-04-14
Update topology endpoint to account for transparent proxy

freddygv:
This PR ensures that we return:
- Upstreams and downstreams from intentions
- Whether all instances of a service's proxies are in transparent mode
- The default allow policy

TODO after beta:

- Refactor/break up the catalog function

### DIFF
--- a/agent/consul/state/catalog.go
+++ b/agent/consul/state/catalog.go
@@ -2853,6 +2853,8 @@ func checkProtocolMatch(tx ReadTxn, ws memdb.WatchSet, svc *structs.GatewayServi
 	return idx, svc.Protocol == protocol, nil
 }
 
+// TODO(freddy) Split this up. The upstream/downstream logic is very similar.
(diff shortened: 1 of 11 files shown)


agent/consul/state/catalog.go
```diff
@@ -2884,6 +2887,38 @@ func (s *Store) ServiceTopology(
...
+
+		idx, proxies, err := serviceNodesTxn(tx, ws, indexConnect, q)
+		if err != nil {
+			return 0, nil, fmt.Errorf("failed to fetch connect endpoints for service %s: %v", sn.String(), err)
+		}
+		if idx > maxIdx {
+			maxIdx = idx
+		}
+		if len(proxies) == 0 {
+			break
```
rboyer, line 2905: took me a minute to realize you were using this to skip over the rest of the switch/case block

rboyer (approved):
LGTM

PR #10023, opened 2021-04-14
Add content type headers to raw KV responses

picatz:
CVE-2020-25864

A vulnerability was identified in Consul and Consul Enterprise ("Consul") such that a specially crafted KV entry could be used to perform a XSS attack when viewed in the raw mode.

This PR adds content-type headers to raw KV responses to prevent that attack.


rboyer (approved):
LGTM

PR #10031, opened 2021-04-15
Split Upstream.Identifier() so non-empty namespace is always prepended in ent

freddygv:
Previously the Upstream.Identifier() key would not prepend the namespace
if it was the default namespace. On the other hand, elements in the
DiscoveryChain map always have the namespace prepended in enterprise.
This mismatch leads to not being able to fetch UpstreamConfig for a
given upstream's DiscoveryChain.

This PR preserves the original behavior in OSS, and adapts some
tests to account for the oss/ent split.


rboyer (approved):
LGTM

PR #10035, opened 2021-04-15
ui: Never respond with empty gateway addresses (mock-api)

johncowen:
We noticed that our mock API would sometimes respond with an empty array
of addresses - which resulted in an empty space in the gateway upstream
listing which looked as though it could be broken.

I checked with backend, and as this will never happen, I made the change
here also so the gateway upstream list is always fully populated with
addresses.


kaxcode (approved):
LGTM

PR #10039, opened 2021-04-15
ui: remove old nspace argument from the proxy instance repository

johncowen:
The extra argument meant that the blocking query configuration wasn't
being read properly, and therefore the correct ?index wasn't being sent
with the request.

The error was introduced after our last 1.9.x release (but it is in the 1.10-0-alpha), hence no backport.

@mikemorris are we adding changelog entries for things which were broken in 1.10.0-alpha and fixed here?

### DIFF
--- a/.changelog/10039.txt
+++ b/.changelog/10039.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: ensure proxy instance API requests perform blocking queries correctly
+```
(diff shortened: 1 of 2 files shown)


mikemorris:
To clarify, this issue was introduced in https://github.com/hashicorp/consul/pull/9746, which was not backported to the `release/1.9.x` branch (as opposed to having been backported but just not yet released), so no backport is necessary.

Changelog entry for beta1 would be good, we'll collapse those down to remove intermediate changes when doing the final 1.10.0 GA release, but helpful for now.

johncowen:
Changelog added, just needs an approval now

kaxcode (approved):
LGTM

PR #10046, opened 2021-04-15
Don't panic on nil UpstreamConfiguration.Clone()

freddygv:
Previously would panic when attempting to read `c.Overrides` just below.


rboyer (approved):
LGTM

PR #10049, opened 2021-04-16
Check for optionally prepended namespace in upstream assertion

freddygv:
Previously the Upstream identifier in enterprise avoided prepending the namespace if the upstream was in the default namespace. 

This changed in #10031, so this PR updates integration test assertions to account for a potential `default/` prefix.


markan (approved):
Looks reasonable to me.

PR #10050, opened 2021-04-16
ui: Add information regarding Host header for ingress gateways

johncowen:
Adds intro copy to the ingress gateways upstreams tab explaining the `Host` header for non-Consul usage.

Screenshot 2021-04-16 at 12 22 27

Notes:

1. Added an `%inline-code` CSS component in order to do this, plus a story for it, then used this component for the other place where we have inline code (KVs) for consistency.
2. Added the ability to reference certain environment variables from within our translations file to make it easier to insert root domain configurable URLs into translations.
3. We are currently inserting HTML into the translations file to allow us to sprinkle _minimal_ formatting and links into the translations. This is a small stop gap until we can add a minimal markdown formatter in-between the translation file and the hbs template (could be done in the same injected custom I18nService we've added here). Importantly we won't have to update out template files when we do this, only replace the HTML in the translations file with markdown.

### DIFF
--- a/ui/packages/consul-ui/app/components/inline-code/README.mdx
+++ b/ui/packages/consul-ui/app/components/inline-code/README.mdx
@@ -1,0 +1,22 @@
+---
+class: css
+---
+# InlineCode
+
+All `p code` within `main` and any `ModalLayer`s default to use the following
+inline code CSS component.
+
+```hbs preview-template
+<p>
+  This is to so we can highlight code <code>inline</code> in paragraphs and such-like.
+</p>
+```
+
+It can also be added to any additional elements using the following
+placeholder.
+
+```css
+p code {
+  @extend %inline-code;
+}
+```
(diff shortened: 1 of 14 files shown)


kaxcode (approved):
👍 LGTM, I just have one question.

Should the inline-code component be in the styles/components folder?

johncowen:
> Should the inline-code component be in the styles/components folder?

Nope, we've been moving all our old CSS components from the styles/components folder into the app/components folder - that way everything is in the same place whether its a template only, a javascript, or a CSS only component. Everything is a component, everything has `.mdx` documentation etc etc all contained within the one folder.

There are a few things that don't feel component-like (for example our typography file with `%h100`, `%h200`, `%p1` etc etc and our colors maybe) but I even think these will eventually move to a component folder.

Off the top of my head there _might_ be something that ends up staying in `styles` but we'll see what's left over when all the component-like things have been moved.

PR #10051, opened 2021-04-16
ui: Add Admin Partition feature flag

johncowen:
This adds a `PartitionEnabled`/`CONSUL_PARTITIONS_ENABLED` feature flag
that can be set during production form the consul binary, or
additionally during development/testing via cookies.

If this property is not set in the binary at all, then we default to `false`.

Also added this setting to our engineering docs (theres a little more change here just due to me moving that around a little, see commit messages for details)


kaxcode (approved):
👍

PR #10052, opened 2021-04-16
ui: Vertically stack methods listing in the L7 intentions form

johncowen:
Before:

Screenshot 2021-04-16 at 13 33 13

After:

Screenshot 2021-04-16 at 13 32 47


kaxcode (approved):
👍

PR #10067, opened 2021-04-19
structs: remove unnecessary name field from proxy-defaults

dnephin:
The `ProxyConfigEntry.Name` field should always be `global`. So instead of having to set that everywhere, and write code that checks that constraint everywhere, we can remove the field. It likely has to stay around for backwards compatibility, but making it an `interface{}` and deprecating it should indicate that it's not supposed to be used.


jmurret:

PR #10068, opened 2021-04-19
health: use blocking queries for near query parameter

dnephin:
Effectively reverts the changes in #9758

There's no reason to error here. We can still satisfy the request using blocking queries.

### DIFF
--- a/agent/rpcclient/health/health.go
+++ b/agent/rpcclient/health/health.go
@@ -86,3 +86,11 @@ func (c *Client) Notify(
 	}
 	return c.Cache.Notify(ctx, cacheName, &req, correlationID, ch)
 }
+
+func (c *Client) UseStreaming(req structs.ServiceSpecificRequest) bool {
+	if req.Ingress || req.Source.Node != "" {
+		return false
+	}
+
+	return req.QueryOptions.UseCache || req.QueryOptions.MinQueryIndex > 0
+}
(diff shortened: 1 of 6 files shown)


rboyer:
Would a test be feasible?

agent/rpcclient/health/health.go
```diff
@@ -81,8 +81,16 @@ func (c *Client) Notify(
 	ch chan<- cache.UpdateEvent,
 ) error {
 	cacheName := c.CacheName
-	if req.Ingress {
-		cacheName = c.CacheNameIngress
+	if req.Ingress || req.Source.Node != "" {
```
rboyer, line 84: It's not a straight replacement, but it would be neat if you could share this potentially variable conditional line with the same line in `UseStreaming` to avoid drift should either one pick up another clause.

agent/rpcclient/health/health.go
```diff
@@ -81,8 +81,16 @@ func (c *Client) Notify(
...
 	}
 	return c.Cache.Notify(ctx, cacheName, &req, correlationID, ch)
 }
+
+func (c *Client) UseStreaming(req structs.ServiceSpecificRequest) bool {
+	if req.Ingress || req.Source.Node != "" {
+		return false
+	}
+
+	return req.QueryOptions.UseCache || req.QueryOptions.MinQueryIndex > 0
```
rboyer, line 95: why is `args.QueryOptions.UseCache` here? It wasn't here in the pre-PR equivalent, and the place it was present is used differently when evaluating the chain of boolean expressions.

dnephin:
This change was merged as part of #10112, because that change made it possible to test this one.

PR #10069, opened 2021-04-19
Disallow * as service-defaults name

freddygv:
This PR disallows wildcards as the name for a `service-defaults` config entry. Using wildcards as service names clashes with how they're used by intentions.

Note that the wildcard specifier is already disallowed as the destination name of connect proxies.

### DIFF
--- a/.changelog/10069.txt
+++ b/.changelog/10069.txt
@@ -1,0 +1,3 @@
+```release-note:breaking-change
+connect: Disallow wildcard as name for service-defaults. 
+```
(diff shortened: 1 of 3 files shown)


rboyer:
Can someone that already has a `service-defaults/*` config entry persisted still delete it after this change?

freddygv:
@rboyer Yep, just tried that out in a single server setup. Delete is allowed and re-writing it is disallowed.

rboyer (approved):
LGTM

PR #10073, opened 2021-04-20
Add replication metrics

mkeeler:
This adds metrics around replication status and latest replicated index to be emitted for:

* ACL Tokens
* ACL Policies
* ACL Roles
* Legacy ACLs
* Federation States
* Config Entries
* Namespaces

The metrics will be emitted with the name: `consul.leader.replication.<type>.status` and `consul.leader.replication.<type>.index`

The `status` one is 0 if the last round of replication errored and 1 if it didn't. The `index` one is updated with the last replicated index.

As we don't have unit tests to ensure metrics were emitted correctly I tested this manually like so:

1. `make dev-docker` for this branch
2. `git clone github.com/mkeeler/consul-docker-test`
3. `cd acls` within the `consul-docker-test` source
4. `terraform init`
5. `terraform apply -var consul_image=consul-dev`
   * This will bring up a multi-dc setup with ACLs enabled with token replication and connect enabled which is enough to cause all the various replicators to be running.
6. In a browser go to http://localhost:3000/d/FanxUjuGk/replication?orgId=1
7. Alternatively navigate to http://localhost:9090 (prometheus ui) and search for the metrics:
   * `consul_leader_replication_acl_tokens_status`
   * `consul_leader_replication_acl_tokens_index`
   * `consul_leader_replication_acl_policies_status`
   * `consul_leader_replication_acl_policies_index`
   * `consul_leader_replication_acl_roles_status`
   * `consul_leader_replication_acl_roles_index`
   * `consul_leader_replication_config_entries_status`
   * `consul_leader_replication_config_entries_index`
   * `consul_leader_replication_federation_state_status`
   * `consul_leader_replication_federation_state_index`

### DIFF
--- a/agent/consul/leader.go
+++ b/agent/consul/leader.go
@@ -806,10 +806,19 @@ func (s *Server) runLegacyACLReplication(ctx context.Context) error {
 		}
 
 		if err != nil {
+			metrics.SetGauge([]string{"leader", "replication", "acl-legacy", "status"},
+				0,
+			)
 			lastRemoteIndex = 0
 			s.updateACLReplicationStatusError()
 			legacyACLLogger.Warn("Legacy ACL replication error (will retry if still leader)", "error", err)
 		} else {
+			metrics.SetGauge([]string{"leader", "replication", "acl-legacy", "status"},
+				1,
+			)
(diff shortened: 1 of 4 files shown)


chapmanc (approved):
LGTM

PR #10075, opened 2021-04-20
rpc: some cleanup of canRetry and ForwardRPC

dnephin:
Follow up to #9991

* Remove an unnecessary arg from ForwardRPC
* more cleanup of `canRetry`
* add tests for `canRetry`
* backport some ent changes to integration tests.

### DIFF
--- a/agent/consul/client.go
+++ b/agent/consul/client.go
@@ -10,6 +10,10 @@ import (
 
 	"github.com/armon/go-metrics"
 	"github.com/armon/go-metrics/prometheus"
+	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/serf/serf"
+	"golang.org/x/time/rate"
+
 	"github.com/hashicorp/consul/agent/pool"
 	"github.com/hashicorp/consul/agent/router"
 	"github.com/hashicorp/consul/agent/structs"
@@ -17,9 +21,6 @@ import (
 	"github.com/hashicorp/consul/logging"
 	"github.com/hashicorp/consul/tlsutil"
 	"github.com/hashicorp/consul/types"
-	"github.com/hashicorp/go-hclog"
-	"github.com/hashicorp/serf/serf"
-	"golang.org/x/time/rate"
 )
 
 var ClientCounters = []prometheus.CounterDefinition{
@@ -287,7 +288,10 @@ TRY:
 	)
 	metrics.IncrCounterWithLabels([]string{"client", "rpc", "failed"}, 1, []metrics.Label{{Name: "server", Value: server.Name}})
 	manager.NotifyFailedServer(server)
-	if retry := canRetry(args, rpcErr); !retry {
+
+	// Use the zero value for RPCInfo if the request doesn't implement RPCInfo
+	info, _ := args.(structs.RPCInfo)
+	if retry := canRetry(info, rpcErr); !retry {
 		return rpcErr
 	}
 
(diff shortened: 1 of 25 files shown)


agent/consul/client.go
```diff
@@ -287,7 +288,10 @@ TRY:
 	)
 	metrics.IncrCounterWithLabels([]string{"client", "rpc", "failed"}, 1, []metrics.Label{{Name: "server", Value: server.Name}})
 	manager.NotifyFailedServer(server)
-	if retry := canRetry(args, rpcErr); !retry {
+
+	// Use the zero value for RPCInfo if the request doesn't implement RPCInfo
+	info, _ := args.(structs.RPCInfo)
+	if retry := canRetry(info, rpcErr); !retry {
```
dnephin, line 294: callout

jsosulska (approved):
LGTM. Thanks for deduping here. This makes a lot of sense. I had one question wrt removing args for req, but that's non-blocking.

agent/consul/rpc.go
```diff
@@ -527,8 +525,8 @@ func (c *limitedConn) Read(b []byte) (n int, err error) {
 	return c.lr.Read(b)
 }
 
-// canRetry returns true if the given situation is safe for a retry.
-func canRetry(args interface{}, err error) bool {
+// canRetry returns true if the request and error indicate that a retry is safe.
+func canRetry(info structs.RPCInfo, err error) bool {
```
dnephin, line 529: callout, there are more interesting changes in this file below as well.

dnephin (commented):
Most of this change is a simple find/replace, but commenting on two places where the change is a bit more interesting

agent/consul/rpc.go
```diff
@@ -542,12 +540,7 @@ func canRetry(args interface{}, err error) bool {
 
 	// Reads are safe to retry for stream errors, such as if a server was
 	// being shut down.
-	info, ok := args.(structs.RPCInfo)
-	if ok && info.IsRead() && lib.IsErrEOF(err) {
-		return true
-	}
-
-	return false
+	return info != nil && info.IsRead() && lib.IsErrEOF(err)
```
rboyer, line 543: I'm getting a funny feeling that this may have the nil-not-nil behavior after the refactor. I could totally be wrong though.

agent/consul/rpc.go
```diff
@@ -790,11 +783,6 @@ func (s *Server) raftApplyWithEncoder(
 		// In this case we didn't apply all chunks successfully, possibly due
 		// to a term change; resubmit
 		if resp == nil {
-			// This returns the error in the interface because the raft library
```
rboyer, line 793: Removed because this error unification already happened in a prior refactor?

dnephin: That is correct, yes. I missed removing the comment earlier.

agent/consul/rpc.go
rboyer: Nah I think it's fine: https://play.golang.org/p/3oTfQl1EQ4G

rboyer (approved):

agent/consul/rpc.go
dnephin: Nice test! That's a good thing to confirm.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/365042.

PR #10085, opened 2021-04-21
lib/ttlcache: never decrease the expiry on update

dnephin:
I ran into this while working on a larger change.  This was not actually causing a bug, but it seems like it would have the potential to cause a bug if we ever had different sources use a different TTL value for the same entry.

Previously we allowed `Update` to decrease the expiry value. With this change the expiry must always increase.


rboyer (approved):
LGTM

PR #10086, opened 2021-04-21
sdk/retry: a few small debug improvements

dnephin:
On a few occasions I've had to dig through stack traces for tests that timeout
and noticed that `retry.Run` runs the function in a goroutine. This makes
debuging a timeout more difficult because the gourinte of the retryable
function is disconnected from the stack of the actual test. It requires
searching through the entire stack dump to find the other goroutine.

By using panic instead of runtime.Goexit() we remove the need for a
separate goroutine.

Also a few other small improvements:
* add `R.Helper` so that an assertion function can be used with both
  testing.T and retry.R.
* Pass t to `Retryer.NextOr`, and call `t.Helper` in a number of places
  so that the line number reported by `t.Log` is the line in the test
  where `retry.Run` was called, instead of some line in `retry.go` that
  is not relevant to the failure.
* improve the implementation of `dedup` by removing the need to iterate
  twice. Instad track the lines and skip any duplicate when writing to
  the buffer.

### DIFF
--- a/sdk/testutil/retry/retry.go
+++ b/sdk/testutil/retry/retry.go
@@ -17,7 +17,6 @@ import (
 	"fmt"
 	"runtime"
 	"strings"
-	"sync"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


rboyer (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/358697.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit dede7c3a60bf16f0032b49ef2d165d4fcc297b6f onto `release/1.10.x` succeeded!

PR #10089, opened 2021-04-22
CLI: Allow snapshot inspect to work on internal raft snapshots directly.

banks:
As documented, this extends `consul snapshot inspect` to allow it to automatically detect an attempt to open an internal snapshot direct from the raft data dir on a server rather than requiring a "packaged" snapshot to be generated and downloaded.

This is often needed by operators trying to debug issues on live Consul servers.

A note is added to docs to try to clarify the difference between external, on-demand, packaged snapshot files, and internal raw raft state snapshots, and an example usage added.

image


mkeeler (approved):
I think this needs a changelog entry added but otherwise this all looks good. Having this is going to be very useful for debugging.

.changelog/10089.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:improvement
+cli: snapshot inspect command can now inspect raw snapshots from a server's data
+dir. [[GH-10089](https://github.com/hashicorp/consul/pull/10089)]
```
banks, line 3:
```suggestion
dir.
```

PR #10093, opened 2021-04-22
ui: Splitting up the translations file into folder structure

kaxcode:


johncowen (approved):
Great stuff! Thanks for doing this.

I think moving forwards the only thing we need to be slightly cautious of is trying to make sure that the translation components folder matches up with the app components folder, so all the keys match up if we eventually move them to be all bundled together.

PR #10097, opened 2021-04-23
test: switch envoy integration tests to use pgrep+kill instead of ps+grep+awk+kill

rboyer:
It's less DIY now, and also for some reason the old way started breaking during #9855 tests.


dnephin (approved):
LGTM, seems like a good improvement.

I think `get_envoy_pid` is only ever used by `kill_envoy`, so potentially `kill_envoy` could be changed to use `pkill -TERM -f` with this same pattern.

rboyer:
Switched to the `pkill` variant.

markan (approved):
LGTM

i0rek (commented):
LGTM

PR #10107, opened 2021-04-23
Update changelog for #10073 and also update the telemetry docs which I forgot to before.

mkeeler:
This is a simple changelog + docs PR. The docs update was to add in information about the new metrics added in #10073


rboyer (approved):
LGTM

PR #10109, opened 2021-04-23
Add prometheus guage definitions for replication metrics.

mkeeler:
This fixes up not adding these definitions in #10073


dnephin (approved):
LGTM

PR #10110, opened 2021-04-23
sdk/retry: add support for ending the iteration early

dnephin:
This is a feature that exists in an another implementation of this pattern that I wrote a while ago (https://pkg.go.dev/gotest.tools/v3/poll#WaitOn). I have encountered a few tests where I need to be able to stop retrying if a particular error is received.

This PR adds support for stopping the `retry.Run` by using `r.Stop()`.

It also does a bit of cleanup that I missed in #10086. I realized while looking at this again that it is not necessary to pass anything to `Retryer.Continue` (formerly `Retryer.NextOr`). The return value from that function is all we need to decide to call `fail` or not.

### DIFF
--- a/sdk/testutil/retry/retry.go
+++ b/sdk/testutil/retry/retry.go
@@ -34,6 +34,7 @@ type Failer interface {
 // R provides context for the retryer.
 type R struct {
 	fail   bool
+	done   bool
 	output []string
 }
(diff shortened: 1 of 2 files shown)


dnephin:
This change was merged as part of #10112

PR #10112, opened 2021-04-23
streaming: replace agent/cache with submatview.Store

dnephin:
This PR replaces `agent/cache.Cache` in the streaming flow with a new `submatview.Store`. The new `Store` implements a similar interface, and consumes similar interfaces, but removes all the logic about fetching results, which is no longer necessary with streaming.

Some notable differences:
* instead of registering cache-types ahead of time, the new Store expects the incoming `Request` to specify the details about the type and how to initialize the `Materializer` if there is no existing entry.
* the cache eviction logic takes into account active requests. The TTL starts when the last active request ends, instead of each time a request starts. This is important for streaming because we expect the requests to stay active for longer than 10 minutes.
* stopping the `Materializer` (which shuts down the stream) is done synchronously with expiring the entry from the store. This prevents issues that we encountered in the cache where expiration would happen while background goroutines were still running.

This PR is unfortunately a bit large, but hopefully it is mostly tests. This PR includes #10110 (which can be rebased out if that merges), and #10068 because it was not really possible to test those changes without first making this change.

TODO:
* [x] port the 6 tests that moved from `cache-type/streaming_health_services_test.go` to `submatview/store_test.go`
* [x] fix the data race caused by initializing the gRPC logger in tests.
* [x] fix the existing test failures, I believe being caused by the changes to `Agent.New`.
* [x] resolve TODOs around error handling in `Store.Get` and `Store.Notify`
* [x] resolve TODOs about godoc (could be a follow up)
* [x] cleanup the git history a bit, there are some WIP commits that need to be squashed.
* [x] changelog entry


agent/consul/options.go
```diff
@@ -15,5 +16,9 @@ type Deps struct {
 	Tokens          *token.Store
 	Router          *router.Router
 	ConnPool        *pool.ConnPool
-	GRPCConnPool    *grpc.ClientConnPool
+	GRPCConnPool    GRPCClientConner
```
dnephin, line 19: This was done so that tests which use `Agent.New` but don't actually need a gRPC connection pool can use a no-op fake.

agent/health_endpoint.go
```diff
@@ -219,13 +219,6 @@ func (s *HTTPHandlers) healthServiceNodes(resp http.ResponseWriter, req *http.Re
 		return nil, nil
 	}
 
-	useStreaming := s.agent.config.UseStreamingBackend && args.MinQueryIndex > 0 && !args.Ingress
-	args.QueryOptions.UseCache = s.agent.config.HTTPUseCache && (args.QueryOptions.UseCache || useStreaming)
-
-	if args.QueryOptions.UseCache && useStreaming && args.Source.Node != "" {
-		return nil, BadRequestError{Reason: "'near' query param can not be used with streaming"}
-	}
```
dnephin, line 227: This logic has moved into `rpcclient/health.Client.useStreaming` so that all of the logic to determine which backend to use is in a single place.

agent/submatview/store.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+	defer s.releaseEntry(key)
+
+	ctx, cancel := context.WithTimeout(ctx, info.Timeout)
+	defer cancel()
+
+	result, err := e.materializer.getFromView(ctx, info.MinIndex)
+
+	// TODO: does context.DeadlineExceeded need to be translated into a nil error
+	// to match the old interface?
```
dnephin, line 103: Making a note of this TODO. "old interface" here refers to agent/cache.Cache

agent/submatview/store.go
```diff
@@ -0,0 +1,220 @@
...
+		index := info.MinIndex
+		for {
+			result, err := e.materializer.getFromView(ctx, index)
+			switch {
+			case ctx.Err() != nil:
+				return
+			case err != nil:
+				// TODO: cache.Notify sends errors on updateCh, should this do the same?
+				// It seems like only fetch errors would ever get sent along and eventually
+				// logged, so sending may not provide any benefit here.
```
dnephin, line 137: Making a note of this TODO. This is a small change in behaviour from `Cache.Notify` which might be significant.

agent/setup.go
```diff
@@ -123,6 +126,10 @@ func NewBaseDeps(configLoader ConfigLoader, logOut io.Writer) (BaseDeps, error)
 	return d, nil
 }
 
+// grpcLogInitOnce because the test suite will call NewBaseDeps in many tests and
+// causes data races when it is re-initialized.
+var grpcLogInitOnce sync.Once
```
rboyer, line 131: Doesn't this mean that these logs are useless during tests?

dnephin: I guess it depends on how the logs are setup. If the logs are setup to go to stdout/stderr then it should work fine, they'll be attributed to the active test. If the logs are sent via `t.Log` it might hide some.

Although not doing this doesn't really fix the issue either, because the gRPC logger is a global. So if multiple tests are run in parallel, we would send all of the logs to one test logger, and none to the other.

This is only the logs for the grpc client and server though, so probably not important for any tests of Consul behaviour.

agent/agent.go
```diff
@@ -373,15 +373,21 @@ func New(bd BaseDeps) (*Agent, error) {
 		cache:           bd.Cache,
 	}
 
-	cacheName := cachetype.HealthServicesName
-	if bd.RuntimeConfig.UseStreamingBackend {
-		cacheName = cachetype.StreamingHealthServicesName
+	// TODO: create rpcClientHealth in BaseDeps once NetRPC is available without Agent
+	conn, err := bd.GRPCConnPool.ClientConn(bd.RuntimeConfig.Datacenter)
+	if err != nil {
+		return nil, err
```
rboyer, line 379: What does it mean for this to error here? Should we add extra context? Would this fail if there was a connectivity issue with other consul agents. If so that seems bad to fail to startup.

agent/submatview/materializer.go
```diff
@@ -80,6 +79,18 @@ func NewMaterializer(deps Deps) *Materializer {
 		retryWaiter: deps.Waiter,
 		updateCh:    make(chan struct{}),
 	}
+	if deps.Waiter == nil {
```
rboyer, line 82:
```suggestion
	if v.retryWaiter == nil {
```
Technically the same thing, but this one seems less confusing should this ever get refactored and additional lines be added in between the field initial assignment and the later defaulting.

agent/submatview/store.go
```diff
@@ -0,0 +1,222 @@
...
+		timer := s.expiryHeap.Next()
+		s.lock.RUnlock()
+
+		select {
+		case <-ctx.Done():
+			timer.Stop()
+			return
+		case <-s.expiryHeap.NotifyCh:
+			timer.Stop()
+			continue
```
rboyer, line 59: Why is this case branch here, and why does it just resume the `for` loop? Is it because this is a notification that the timer on the heap has been updated and we need to recycle the timer?

I think these branches could benefit from some light commentary while it's fresh in your mind.

agent/agent.go
dnephin: Good question. As long as we don't use `WithBlock()` (we don't use and it seems unlikely that we ever would) then the only errors that `grpc.Dial` can return are errors related to the options we pass into `grpc.Dial`, or building the resolver. Our customer resolver builder does not return errors, so this seems ok.  I can definitely add some context to the error.

I'll see if there is another place that this could be done.

agent/submatview/store.go
```diff
@@ -0,0 +1,222 @@
...
+			case err != nil:
+				// TODO: cache.Notify sends errors on updateCh, should this do the same?
+				// It seems like only fetch errors would ever get sent along and eventually
+				// logged, so sending may not provide any benefit here.
+
+				s.logger.Warn("handling error in Store.Notify",
+					"error", err,
+					"request-type", req.Type(),
+					"index", index)
+				continue
```
rboyer, line 145: Wouldn't this busy-loop since the `index` value hasn't been bumped?

agent/submatview/store.go
```diff
@@ -0,0 +1,222 @@
...
+			}
+
+			index = result.Index
+			u := cache.UpdateEvent{
+				CorrelationID: correlationID,
+				Result:        result.Value,
+				Meta:          cache.ResultMeta{Index: result.Index},
+				Err:           err,
```
rboyer, line 153: Isn't this just `nil` as defined?

agent/submatview/store.go
```diff
@@ -0,0 +1,222 @@
...
+				return
+			}
+		}
+	}()
+	return nil
+}
+
+// getEntry from the store, and increment the requests counter. releaseEntry
+// must be called when the request is finished to decrement the counter.
+func (s *Store) getEntry(req Request) (string, entry, error) {
```
rboyer, line 167: I think this should return a materializer instead of an `entry` as return-value-2. The only field off of the entry we use is that one. Other fields on `entry` are misleading (`requests` (a copy)) or unsafe (`stop`) for callers to get access to.

agent/submatview/store.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	mat, err := req.NewMaterializer()
+	if err != nil {
+		return "", e, err
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	go mat.Run(ctx)
+
```
rboyer, line 187: Should `getEntry` remove the entry from the expiryHeap?

agent/submatview/store.go
dnephin: I don't believe so. This is an error case, so I think it's expected that `index` doesn't change. `getFromView` will block when it gets called again, waiting on either a new value, another error, or the timeout. 

In `Materializer` there is a retry with backoff on an error, so the `updateCh` that `getFromView` blocks on shouldn't get updated until that reports another error.

agent/submatview/store.go
dnephin: Ah yes, true. I guess depending on how the `TODO` above this is resolved, it may become non-nil with that change.

agent/submatview/store.go
dnephin: Ah yes, true! That looks like a good change

agent/submatview/store.go
dnephin: It could, but it's not necessary the way it is written now. `getEntry` increments the count of requests, and the expiry is ignored if `requests > 0`. Alternatively we could remove things from the expiry heap instead of tracking requests. I wonder if that would be more expensive because of the need to modify the heap more often. Where as incrementing the request count is pretty cheap since we have a reference to it from the map.

agent/submatview/store.go
dnephin: Right, exactly. The heap has changed so we need to start a new timer. I've added comments to these branches.

agent/submatview/store.go
dnephin: I've removed the TODO. From what I can tell there is no reason to report errors to the notify channel. Also removed this line since, as you noticed, `err` will always be `nil`.

agent/agent.go
dnephin: It looks like the alternative would be to pass the `GRPCConnPool` to `health.MaterializerDeps` instead of the `conn`. That would however force a separate connection for each Materializer in the Store, and makes closing the connections more difficult because the lifecycle is tied to the materialize. The current approach seems correct based on what I've found so far.

agent/submatview/store.go
```diff
@@ -0,0 +1,237 @@
...
+		return Result{}, err
+	}
+	defer s.releaseEntry(key)
+
+	ctx, cancel := context.WithTimeout(ctx, info.Timeout)
+	defer cancel()
+
+	result, err := materializer.getFromView(ctx, info.MinIndex)
+	// context.DeadlineExceeded is translated to nil to match the behaviour of
+	// agent/cache.Cache.Get.
```
banks, line 123: I wonder if this could mask underlying network timeouts too?

I assume gRPC client code also uses contexts for things like connection timeout, if it did and there was an actual error like unable to connect to servers which caused a timeout, could we mistakenly mask that error?

I wonder if there is a way to be sure this is the timeout based on the user's requested blocking timeout vs a downstream network timeout that's propagated directly back through the calls? Right now this might never be possible but it makes me wonder if this could ever result in a subtle bug later?

agent/submatview/store.go
banks: I think that sounds right, perhaps a comment here could clarify that assumption that `getFromView` will never return repeated errors without a delay/backoff so this is safe would be useful?

It's not needed here, but I was thinking if we did need to be defensive, adding a _rate limit_ in this error case of the loop rather than exponentially backing off errors again might be a nice approach. That way if something else is already doing a backoff we won't add additional compounded delays here, but if there is a condition introduced later where the materializer could return errors without backing of then the rate limit will kick in to prevent this from becoming a busy loop?

I don't think we need to do that now, just thinking aloud.

agent/submatview/store.go
banks: Yeah I like this approach where the expiry heap only contains inactive entries rather than needing to maintain a heap over all entries and constantly update it to keep nudging back expiry on the active ones.

99% of the time I suspect all entries will be active so not needing to maintain a heap in steady state seems like a cleaner approach.

agent/submatview/store.go
```diff
@@ -0,0 +1,237 @@
...
+// agent/cache.Cache, and allows a single Materliazer to fulfill multiple requests
+// as long as the requests are identical.
+// Store is used in place of agent/cache.Cache because with the streaming
+// backend there is no longer any need to run a background goroutine to refresh
+// stored values.
+type Store struct {
+	logger     hclog.Logger
+	lock       sync.RWMutex
+	byKey      map[string]entry
+	expiryHeap *ttlcache.ExpiryHeap
```
banks, line 26: Perhaps this would be a good place to note that this expiryHeap unlike the one in agent/cache only contains _inactive_ entries? Might help orient folks who are expecting this to work the same way?

agent/submatview/store_test.go
```diff
@@ -0,0 +1,458 @@
...
+	})
+
+	// wait for the entry to expire, with lots of buffer
+	time.Sleep(3 * ttl)
+
+	store.lock.Lock()
+	defer store.lock.Unlock()
+	require.Len(t, store.byKey, 0)
+	require.Equal(t, ttlcache.NotIndexed, e.expiry.Index())
+}
```
banks, line 451: I think these tests look great.

Possibly a TODO for later but what do you think about adding a slightly "fuzzy" test that starts up a bunch of concurrent `Notify` calls and asserts reasonable behaviour for each one. The value in the test isn't so much covering specific anomalies or cases but just verifying the concurrent behaviour works (i.e. all clients see the update), and when run with `-race` we have a better chance of catching any data race issues?

sdk/testutil/retry/retry.go
```diff
@@ -162,9 +177,9 @@ func ThreeTimes() *Counter {
 // Retryer provides an interface for repeating operations
 // until they succeed or an exit condition is met.
 type Retryer interface {
-	// NextOr returns true if the operation should be repeated.
-	// Otherwise, it calls fail and returns false.
-	NextOr(t Failer, fail func()) bool
+	// NextOr returns true if the operation should be repeated, otherwise it
```
banks, line 180:
```suggestion
	// Continue returns true if the operation should be repeated, otherwise it
```

banks (approved):
Looks awesome @dnephin 👏 

I have a few minor comments inline, none are blocking though so see what you think!

agent/submatview/materializer.go
```diff
@@ -32,7 +31,7 @@ type View interface {
 	// separately and passed in in case the return type needs an Index field
 	// populating. This allows implementations to not worry about maintaining
 	// indexes seen during Update.
-	Result(index uint64) (interface{}, error)
+	Result(index uint64) interface{}
```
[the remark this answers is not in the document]
dnephin: This interface is only for the view, which is effectively just a map with some filtering and sorting. There's no IO or network operations here. It seems unlikely that other `View` implementations would need an error return for this, but we can always add it back if we do need it.

Even the `error` return from `View.Update` is arguably not necessary. The one place we error there would be a logic bug, and could panic.

agent/submatview/store.go
dnephin: From what I can tell request timeouts in gRPC are always set by the context, so there's no difference. Connect timeouts are reported using this error: https://github.com/grpc/grpc-go/blob/24d03d9f769106b3c96b4145244ce682999d3d88/internal/transport/transport.go#L713, so would not be `context.DeadlineExceeded`.

I believe this is safe. I do think it is unfortunate, and it would be better to report this timeout to the user, but agent/cache masks these kinds of errors, so I thought for now it would be better to match the behaviour.

agent/submatview/store_test.go
dnephin: Ya, that is an option. In my mind, if we were to do that, we should use a real backend not a fake one. It seems like a fake will never be close enough to the real behaviour that any such test wouldn't be meaningful with a fake.

rboyer (approved):
LGTM

dnephin:
Test failures pass locally, so likely flakes.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/358627.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 9b344b3300e5fe8ee261149e72d8e3337683cf84 onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/358627)

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/358712.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 9b344b3300e5fe8ee261149e72d8e3337683cf84 onto `release/1.10.x` succeeded!

PR #10117, opened 2021-04-25
Fix strings.Replace->strings.ReplaceAll

haxandmat:
Function
`strings.Replace (version, "'", "", -1)`
replaced by
`strings.ReplaceAll (version, "'", "")`


Haxandmat:
Fix

PR #10121, opened 2021-04-26
ui: Add conditionals to Lock Session list items

kaxcode:
Adds conditionals to show ID if no Name is provided. Adds conditional to hide TTL icon if no TTL is provided. 

Before
Screen Shot 2021-04-27 at 10 04 58 AM

After
Screen Shot 2021-04-27 at 10 09 28 AM

### DIFF
--- a/ui/packages/consul-ui/app/components/consul/lock-session/list/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/lock-session/list/index.hbs
@@ -8,6 +8,7 @@
   {{/if}}
   </BlockSlot>
   <BlockSlot @name="details">
+    {{#if item.Name}}
     <dl>
       <dt>
         <CopyButton
@@ -17,6 +18,7 @@
       </dt>
       <dd>{{item.ID}}</dd>
     </dl>
+    {{/if}}
     <dl class="lock-delay">
       <dt>
         <Tooltip>
--- a/ui/packages/consul-ui/app/components/consul/lock-session/list/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/lock-session/list/index.hbs
@@ -33,16 +33,18 @@
       </dt>
       <dd data-test-session-delay>{{duration-from item.LockDelay}}</dd>
     </dl>
-    {{#if item.TTL}}
     <dl class="ttl">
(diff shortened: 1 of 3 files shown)


ui/packages/consul-ui/app/components/consul/lock-session/list/index.hbs
```diff
@@ -1,9 +1,14 @@
 {{#if (gt items.length 0)}}
 <ListCollection @items={{items}} class="consul-lock-session-list" as |item index|>
   <BlockSlot @name="header">
+  {{#if item.Name}}
     <span>{{item.Name}}</span>
+  {{else}}
+    <span>{{item.ID}}</span>
+  {{/if}}
```
johncowen, line 8: We probably still need the `ID` `CopyButton` if the Name isn't set

ui/packages/consul-ui/app/components/consul/lock-session/list/index.hbs
```diff
@@ -13,6 +18,7 @@
       </dt>
       <dd>{{item.ID}}</dd>
     </dl>
+    {{/if}}
```
johncowen, line 21: I know this is from ages ago, but could we move the CopyButton here out of the `dt` and into the `dd`?

johncowen (commented):
Thanks for sorting this! There are few other little things we could do here (like the API change/bug we found plus some other bits).

If it was me I'd bundle them all up into a Lock Sessions Improvements/Fixes PR (this PR), but up to you.

If you want to keep them here then lemme know and I can make some comments here with them on. Otherwise let's catch up when you get a sec and we can go through.

ui/packages/consul-ui/app/components/consul/lock-session/list/index.hbs
kaxcode: I'm moving away from using `dl` in the composite-row. It's caused a lot of issues. If I update this is to move everything into a `span`. We can discuss this.

kaxcode: Leaving any `dl` changes outside of this PR.

PR #10122, opened 2021-04-26
ui: Fix empty SVG height

kaxcode:
Adding a conditional that prevents the Topology view from breaking when there are no upstreams in the upstream column.

Before
Screen Shot 2021-04-27 at 9 57 52 AM

After
Screen Shot 2021-04-27 at 9 58 34 AM

### DIFF
--- a/ui/packages/consul-ui/app/components/topology-metrics/index.js
+++ b/ui/packages/consul-ui/app/components/topology-metrics/index.js
@@ -108,9 +108,8 @@ export default class TopologyMetrics extends Component {
     // Calculate viewBox dimensions
     this.downView = document.getElementById('downstream-lines').getBoundingClientRect();
     const upstreamLines = document.getElementById('upstream-lines').getBoundingClientRect();
-    if (document.getElementById('upstream-column')) {
-      const upstreamColumn = document.getElementById('upstream-column').getBoundingClientRect();
-
+    const upstreamColumn = document.getElementById('upstream-column').getBoundingClientRect();
+    if (upstreamColumn) {
       this.upView = {
         x: upstreamLines.x,
         y: upstreamLines.y,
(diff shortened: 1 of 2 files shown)


ui/packages/consul-ui/app/components/topology-metrics/index.js
```diff
@@ -108,14 +108,16 @@ export default class TopologyMetrics extends Component {
...
+    if (document.getElementById('upstream-column')) {
+      const upstreamColumn = document.getElementById('upstream-column').getBoundingClientRect();
+
+      this.upView = {
+        x: upstreamLines.x,
+        y: upstreamLines.y,
+        width: upstreamLines.width,
+        height: upstreamColumn.height + 10,
+      };
+    }
```
johncowen, line 120: Are there any ways we can tighten this up a little, for example the code refers to the same DOM element twice, but looks up it up twice also.

johncowen (commented):
Hey :wave:! 

I'm not 100% sure what this fixes, could you add something to the description for me?

Apart from that I left a little inline comment suggestion.

ui/packages/consul-ui/app/components/topology-metrics/index.js
kaxcode: Where is it looked up twice? 🤔

kaxcode:
> Hey 👋!
> 
> I'm not 100% sure what this fixes, could you add something to the description for me?
> 
> Apart from that I left a little inline comment suggestion.

@johncowen I've added a description.

ui/packages/consul-ui/app/components/topology-metrics/index.js
johncowen: On lines 111 and 112:

```javascript
if (document.getElementById('upstream-column')) {
      const upstreamColumn = document.getElementById('upstream-column').getBoundingClientRect();
```

johncowen:
Hey thanks for fixing this up. I was wondering if this is a good opportunity to add some additional testing around this somewhere. What do you think? We could test for 0 'streams in each column and make sure no errors are thrown?

kaxcode:
> Hey thanks for fixing this up. I was wondering if this is a good opportunity to add some additional testing around this somewhere. What do you think? We could test for 0 'streams in each column and make sure no errors are thrown?

@johncowen I'll have to leave this for GA. I'm going to be changing some things for that *(All Services) card.

johncowen (approved):
LGTM! One little thing, I think when we spoke you mentioned changing the changelog a little, not sure if you are going to do that or not.

PR #10124, opened 2021-04-26
ui: Update conditional for topology empty state

kaxcode:
If there are no upstreams or downstreams but the user has permissive intentions, like wildcard or default allow, then we still need to show banner and *(All Services) card, not empty state.

Before
Screen Shot 2021-04-27 at 10 00 24 AM

After
Screen Shot 2021-04-27 at 10 02 55 AM

### DIFF
--- a/ui/packages/consul-ui/app/templates/dc/services/show/topology.hbs
+++ b/ui/packages/consul-ui/app/templates/dc/services/show/topology.hbs
@@ -3,7 +3,7 @@
 as |route|>
   <EventSource @src={{topology}} />
   <div class="tab-section">
-    {{#if (and (eq topology.Upstreams.length 0) (eq topology.Downstreams.length 0))}}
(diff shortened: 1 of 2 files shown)


ui/packages/consul-ui/app/templates/dc/services/show/topology.hbs
```diff
@@ -3,7 +3,7 @@
 as |route|>
   <EventSource @src={{topology}} />
   <div class="tab-section">
-    {{#if (and (eq topology.Upstreams.length 0) (eq topology.Downstreams.length 0))}}
+    {{#if (and (eq topology.Upstreams.length 0) (eq topology.Downstreams.length 0) (not topology.DefaultAllow) (not topology.WildcardIntention))}}
```
johncowen, line 6: Does this cover for both when TProxy is enabled and when its disabled?

kaxcode: Let's discuss this.

johncowen (approved):
LGTM! Thanks for chatting me through that query I had 👍

PR #10126, opened 2021-04-26
Exclude node addresses when using tproxy on k8s

freddygv:
Each filter chain on the tproxy outbound listener needs unique addresses
to match on. However, if multiple upstream services are hosted on the
same node, then their node address would be duplicated across filter
chain matches. If the local node were to dial that upstream IP then
Envoy would not know how to disambiguate it.

In Kubernetes we do not expect services in the mesh to dial upstreams in
the mesh using the node's address.


freddygv:
Closing this in favor of a different set of changes in an upcoming PR.

PR #10127, opened 2021-04-26
Rename "cluster" config entry to "mesh"

freddygv:
This config entry is being renamed primarily because in k8s the name
cluster could be confusing given that the config entry applies across
federated datacenters.

Additionally, this config entry will only apply to Consul as a service
mesh, so the more generic "cluster" name is not needed.

### DIFF
--- a/agent/structs/config_entry_cluster_oss.go
+++ b/agent/structs/config_entry_cluster_oss.go
@@ -2,6 +2,6 @@
 
 package structs
 
-func (e *ClusterConfigEntry) validateEnterpriseMeta() error {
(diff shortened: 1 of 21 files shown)


agent/structs/config_entry_cluster_oss.go
```diff
@@ -2,6 +2,6 @@
 
 package structs
 
-func (e *ClusterConfigEntry) validateEnterpriseMeta() error {
```
rboyer, line 5: note: you'll have to fix one teeny file during the ent merge

rboyer (approved):
LGTM

Issue #9609, opened 2021-01-21
Improve observability around snapshot size vs trailing logs


banks:
## Large Snapshot Replication Failure Problem Statement

Currently, raft writes out a complete snapshot of the state store to disk for every 16k writes to the DB, checked every 30s. After a successful snapshot, we truncate the logs to prevent the logs growing unboundedly. Instead of truncating all logs up to the index we just wrote the snapshot at, we always ensure we have at least 10k logs left in the log store (these parameters can be tuned).

The reason for leaving these "trailing logs" is so that a follower who goes down for a restart just before the snapshot completes can continue replicating once it comes backup (which may take a few seconds).

If a follower does get behind (e.g. it's partitioned for a few minutes) it's possible that there will have been more than 10k updates and a snapshot+truncate on the leader in the interim meaning the follower is unable to continue replicating from the logs. In this case the leader will deliver it an entire new snapshot over the wire before resuming log-based replication.

In general this works well for typical uses of Consul, however it has a hard assumption that the time required to transmit the most recent snapshot to a peer, and have the peer restore it into it's state store ready to resume accepting logs is much smaller than the time taking to append 10k more log updates on the leader.

On clusters that have both a high write rate (e.g. 500 writes per second) and a large state snapshot (e.g. larger than 1GB) even on reasonably fast hardware that assumption can be violated resulting in a permanent error condition where followers can never catch up to the leader again.

A worked example:

* At 500 writes/second, the snapshot threshold is met within 32 seconds and with an interval of 30 seconds that means the leader will be **writing out a full snapshot (1GB) file every minute**.
* Every minute after the snapshot, logs are truncated to 10k entries which is only the last 20 seconds worth of updates.
* If sending 1GB snapshot and restoring it on a follower takes more than 20 seconds, **the follower will never be able to get healthy after a restart** unless the write volume drops.

Note that `raft_trailing_logs` can be increased, but since this requires a restart and only once the current leader has the new value will it improve things, it's often not possible to recover from this situation without downtime or somehow reducing the write rate on the cluster enough to allow followers to catch up and then be reconfigured with a rolling restart. Reducing snapshot size helps in the medium term but often isn't possible immediately and requires more writes to do it so is rarely an effective step that can be taken during an incident.

[...]

PR #10129, opened 2021-04-27
Make Raft trailing logs and snapshot timing reloadable

banks:
This is a partial fix for #9609.

The description in that issue gives lots of background although much of it is also present in the telemetry docs proposed in this PR too.

This PR:

 * Updates Raft and go-metrics to pull in new telemetry and hot reloadable features
 * Makes the raft options reloadable in Consul
 * Updates our pre-registered prometheus metrics to cover the new raft ones and the old ones I've not added to the telemetry guide as important
 * Updates telemetry and config docs with the changes in metrics/reload behaviour
   * Sorry, but I also re-organised the raft metrics to be alphabetical - I've made the mistake of thinking we missed metrics on two separate occasions because they were not grouped alphabetically and even not hierarchically (e.g. some `raft.fsm` metrics are not alongside others... Probably makes the diff more onerous than it should be.)
 * Adds a "Key Metric" section to the telemetry docs that describes this issue in more detail and how to use the new metrics to monitor for it.

## Feedback Requested

The Key Metrics section is verbose but it's pretty hard to simplify describing how to deal with this case without that detail. I feel like it would be better suited for more of a "operator runbook" section where we can go into a bit more detail on debugging common or serious failure modes of Consul, but we have no such thing yet and I wanted to get this available at least before taking on the project of starting a whole new type of content in the docs.

Would love feedback on whether it feels sufficient/appropriate/clear enough or if there are other ideas to improve. I'd probably defer any major changes to docs structure to later rather than take them on in this PR though as I want to get this into the next 1.10 beta.

## Testing

The reloadable stuff has tests here but I also spent a while trying this out for real to make sure the metrics actually work and are useful. For example:

image

Here the leader has an artificially low `raft_trailing_logs` of `1` initially. It is writing our 1GB snapshots roughly every 2 minutes and is handling about 200 writes a second.

At 12:01 I changed `raft_trailing_logs` to `100k` and used `consul_reload` the same leader (s1) continued to snapshot another couple of times but without truncating logs.

I also tested changes to snapshot threshold and timing.

## TODO
 - [x] Add changelogs for these changes, but also check for other fixes in Raft 1.3.0 that could impact Consul users and add those too


website/content/docs/agent/options.mdx
```diff
@@ -1762,29 +1762,50 @@ bind_addr = "{{ GetPrivateInterfaces | include \"network\" \"10.0.0.0/8\" | attr
...
+  1.1.0 and later this defaults to `30s`, and in prior versions it was set to
+  `5s`.
+  
+  Since Consul 1.10.0 this can be reloaded using `consul reload` or sending the
+  server a `SIGHUP` to allow tuning snapshot activity without a rolling restart
+  in emergencies.
+
+- `raft_trailing_logs` - This controls how many log entries are left in the log
+  store on disk after a snapshot is made. This should only be adjusted when
+  followers cannot catch up to the leader due to a very large snapshot size that
```
mkeeler, line 1799:
```suggestion
  followers cannot catch up to the leader due to a very large snapshot size
```

I think the sentence makes sense without the `that`

mkeeler (approved):
LGTM, found on stray word in some docs but otherwise things seem perfect.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/361903.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/361915.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 3ad754ca7b40d61d64d4674bced64820449ca45b onto `release/1.10.x` succeeded!

PR #10131, opened 2021-04-27
xds: ensure that all envoyproxy/go-control-plane protobuf symbols are linked into the final binary

rboyer:
This ensures that if someone does include some extension Consul does not currently make use of, that extension is actually usable. Without linking these envoy protobufs into the main binary it can't round trip the escape hatches to send them down to envoy.

Whenenver the go-control-plane library is upgraded next we just have to re-run 'make envoy-library'.


build-support/scripts/envoy-library-references.sh
```diff
@@ -0,0 +1,53 @@
...
+mkdir _envoy_tmp
+trap "rm -rf _envoy_tmp" EXIT
+(
+cd _envoy_tmp
+
+git clone git@github.com:envoyproxy/go-control-plane
+cd go-control-plane
+git co -b consul-temp "${LIBRARY_VERSION}"
+
+IFS=$'\n' candiates=($(find . -name *.pb.go -a -type f | sed 's@/[^/]*\.pb\.go$@@' | sort -u))
```
markan, line 29: Minor nitpick maybe rename candiates -> candidates?

markan (approved):
Looks reasonable in itself. 

Do we have a doc 'common maintenance tasks' or the like to note the process for these sorts of upgrades? I'm thinking it would be good to have somewhere written down a note for "how to upgrade envoyproxy"

build-support/scripts/envoy-library-references.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
```
dnephin, line 1: Some explanation of why this is necessary, and some breadcrumbs back to the original issue would be great. I imagine we will forget why this is necessary and explaining here in the script that generates it would probably be a good place.

Some mention of the call to `proto.RegisterFile` in package level `init()` functions would be a good clue.

dnephin (commented):
Would a test case that uses one of the unavailable types be possible? That would show that this fixes the reported issue.

rboyer:
@dnephin added a quick test that tests decoding something that previously was not decodable using `*any.Any`

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/359866.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 85a718da63abc61f4f5cdb335461305f144024de onto `release/1.10.x` succeeded!

PR #10133, opened 2021-04-27
ui: Update not-defined intention popover and banner

kaxcode:
Update the wording for the banner and the popover for a service with an upstream that is not explicitly defined. 

Before
Screen Shot 2021-04-27 at 2 41 58 PM

After
Screen Shot 2021-04-27 at 2 42 42 PM

### DIFF
--- a/ui/packages/consul-ui/app/components/topology-metrics/down-lines/index.hbs
+++ b/ui/packages/consul-ui/app/components/topology-metrics/down-lines/index.hbs
@@ -93,6 +93,7 @@
   {{else if (and item.Intention.Allowed (not item.TransparentProxy) (eq item.Source 'specific-intention'))}}
     <TopologyMetrics::Popover
       @type='not-defined'
+      @service={{@service}}
       @position={{find-by 'id' (concat this.guid item.Namespace item.Name) this.iconPositions}}
       @item={{item}}
       @oncreate={{action @oncreate item @service}}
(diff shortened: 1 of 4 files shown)


kaxcode:
@johncowen not sure if this should be considered a bug/feature. This is an update. LMK what you think so I can make the changelog.

ui/packages/consul-ui/app/components/topology-metrics/down-lines/index.hbs
```diff
@@ -93,6 +93,7 @@
   {{else if (and item.Intention.Allowed (not item.TransparentProxy) (eq item.Source 'specific-intention'))}}
     <TopologyMetrics::Popover
       @type='not-defined'
+      @service={{@service}}
```
johncowen, line 96: Just a little suggestion here. Now you are passing two very similar things into the Popover, they could be `@upstream` and `@downstream` instead of `@service` and `@item`? (I saw that you were using upstream and downstream for your variables for your text/copy which is 👌 ) All good either way.

johncowen (approved):
:shipit:

ui/packages/consul-ui/app/components/topology-metrics/down-lines/index.hbs
kaxcode: `Popover` component is also used by `UpLines`. The wording wouldn't work when using this component interchangeably.

Issue #7330, opened 2020-02-21
Retrieve health check definition


DanielCharczynski:
#### Feature Description

There is no possibility to get health definition.

https://discuss.hashicorp.com/t/is-there-a-way-to-retrieve-a-health-check-definition-with-the-api/6107

#### Use Case(s)

We would like to use consul as a single point of truth about services.
We save there health check definition.
Unfortunately there is no possibility to retrievie this definition.

There is possibillity to get service definition but not health check definition. I do not understand why....

Please add such feature. It should be quite easy but usefull.

pierresouchay:
I think it is a duplicate of #1680

There are security reasons why you don't want to do that (for instance, you might have a password in your checks settings).
In our infrastructure, we handle this using the "notes" attributes which is properly retrieved, thus allowing us to get this information.

For an HTTP check, you can put the same value as in your HTTP field (but hiding the possible user/passwords), so it is available from your UIs as information, but not for an attacker

jsosulska:
Hello,

As there has been no movement on this topic, and it is a duplicate as @pierresouchay pointed out, I'll be closing this issue. Please follow #1680 and 👍 if this is something you are looking for.

Happy Coding!

PR #10134, opened 2021-04-27
 Implement traffic redirection exclusion based on proxy config and user-provided values

ishustava:
* If `-proxy-id` is provided to the redirect-traffic command, exclude any listener ports from inbound traffic redirection. This includes `envoy_prometheus_bind_addr`, `envoy_stats_bind_addr`, and the `ListenerPort` from the `Expose` configuration.
* Allow users to provide additional inbound and outbound ports, outbound CIDRs and additional user IDs to be excluded from traffic redirection. This affects both the traffic-redirect command and the iptables SDK package.
* Additionally, use the proxy outbound port from the proxy config if it's set.


command/connect/redirecttraffic/redirect_traffic.go
```diff
@@ -48,6 +54,14 @@ func (c *cmd) init() {
 	c.flags.IntVar(&c.proxyInboundPort, "proxy-inbound-port", 0, "The inbound port that the proxy is listening on.")
 	c.flags.IntVar(&c.proxyOutboundPort, "proxy-outbound-port", iptables.DefaultTProxyOutboundPort,
 		"The outbound port that the proxy is listening on. When not provided, 15001 is used by default.")
+	c.flags.Var((*flags.AppendSliceValue)(&c.excludeInboundPorts), "exclude-inbound-port",
+		"Inbound port to exclude from traffic redirection. May be provided multiple times.")
+	c.flags.Var((*flags.AppendSliceValue)(&c.excludeOutboundPorts), "exclude-outbound-port",
+		"Outbound port to exclude from traffic redirection. May be provided multiple times.")
+	c.flags.Var((*flags.AppendSliceValue)(&c.excludeOutboundCIDRs), "exclude-outbound-cidr",
+		"Outbound CIDR to exclude from traffic redirection. May be provided multiple times.")
+	c.flags.Var((*flags.AppendSliceValue)(&c.excludeUIDs), "exclude-uid",
```
freddygv, line 63: Would it be valuable to also be able to exclude GIDs?

command/connect/redirecttraffic/redirect_traffic.go
```diff
@@ -132,23 +148,72 @@ func (c *cmd) generateConfigFromFlags() (iptables.Config, error) {
...
+		// Set the proxy's inbound port.
+		cfg.ProxyInboundPort = svc.Port
 		if trCfg.BindPort != 0 {
 			cfg.ProxyInboundPort = trCfg.BindPort
 		}
 
-		// todo: Change once it's configurable
+		// Set the proxy's outbound port.
 		cfg.ProxyOutboundPort = iptables.DefaultTProxyOutboundPort
+		if svc.Proxy.TransparentProxy.OutboundListenerPort != 0 {
```
freddygv, line 166: TransparentProxy should have a nil check here since it's a pointer. 

Consul currently always returns a non-nil struct even if it's empty, but that's likely going to change for GA.

command/connect/redirecttraffic/redirect_traffic.go
```diff
@@ -132,23 +148,72 @@ func (c *cmd) generateConfigFromFlags() (iptables.Config, error) {
...
+			_, port, err := net.SplitHostPort(trCfg.StatsBindAddr)
+			if err != nil {
+				return iptables.Config{}, fmt.Errorf("failed parsing host and port from envoy_stats_bind_addr: %s", err)
+			}
+
+			cfg.ExcludeInboundPorts = append(cfg.ExcludeInboundPorts, port)
+		}
+
+		// Exclude the ListenerPort from Expose configs from inbound traffic redirection.
+		for _, exposePath := range svc.Proxy.Expose.Paths {
```
freddygv, line 191: One thing I'm realizing is missing is this bit of magic:
https://www.consul.io/docs/connect/registration/service-registration#checks

If someone enables this, Consul will dynamically expose checks on a listener for each check:
https://github.com/hashicorp/consul/blob/c343544fef094a505d22cf60bd79635cf8e2553e/agent/agent.go#L3839

We don't expose what port is allocated in any of our APIs, similarly to how we don't expose check definitions via any of our APIs, only check statuses. See this comment for more information on why: https://github.com/hashicorp/consul/issues/7330#issuecomment-592015038

There are a few potential solutions to discover those ports off the top of my head:

1. Expand the `structs.HealthCheck` type to include the exposed path.
2. Add a Meta map to checks, like we have for services, and include the exposed path there.
3. Include all check definition details in `structs.HealthCheck`. That comes with the security risk mentioned above.

command/connect/redirecttraffic/redirect_traffic.go
```diff
@@ -132,23 +148,72 @@ func (c *cmd) generateConfigFromFlags() (iptables.Config, error) {
...
+			if exposePath.ListenerPort != 0 {
+				cfg.ExcludeInboundPorts = append(cfg.ExcludeInboundPorts, strconv.Itoa(exposePath.ListenerPort))
+			}
+		}
 	} else {
```
freddygv, line 196: What if the defaulting is pulled out the else? These lines are really far removed from the `if` condition so it's not obvious what case this is. 

```suggestion
	cfg := iptables.Config{
		ProxyUserID: c.proxyUID
		ProxyInboundPort: c.proxyInboundPort
		ProxyOutboundPort: c.proxyOutboundPort
	}
	
	if c.proxyID != "" {
		.....
	}
```

freddygv (commented):
Thanks for expanding this! I made a couple minor comments and one less trivial one about the Expose config. Depending on timing we could push the solution to that issue to after beta2 though.

command/connect/redirecttraffic/redirect_traffic.go
freddygv: So based on the discussion today we should go with #1.

The exposed ports are generated in a few places. 

One is where we add checks:
https://github.com/hashicorp/consul/blob/0c9555ec0e2366996cff09d39d1bcb40eaeee435/agent/agent.go#L2610
https://github.com/hashicorp/consul/blob/0c9555ec0e2366996cff09d39d1bcb40eaeee435/agent/agent.go#L2541

In `addCheck` it's easy to expand the HealthCheck type, since we already have the `check` pointer to it. 

And the other is where we add services with checks:
https://github.com/hashicorp/consul/blob/0c9555ec0e2366996cff09d39d1bcb40eaeee435/agent/agent.go#L3791
https://github.com/hashicorp/consul/blob/0c9555ec0e2366996cff09d39d1bcb40eaeee435/agent/agent.go#L3818

In these two other methods we'll need to query `a.State.Check`, I think, and modify that.

command/connect/redirecttraffic/redirect_traffic.go
ishustava: I think it could be. The main reason we're adding this is to better support the VM use case when you might want to exclude other processes/uids/gids from traffic redirection. I'm inclined to wait for users' feedback on this one because I'm not sure what would be valuable there.

command/connect/redirecttraffic/redirect_traffic.go
ishustava: Great catch! let me update it.

command/connect/redirecttraffic/redirect_traffic.go
ishustava: That's a great find! Let me address this in a separate PR so that this one can be included in beta2.

command/connect/redirecttraffic/redirect_traffic.go
ishustava: done!

freddygv (approved):
:shipit:

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/359312.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 8dffb8913105572af71f582bd204c6f473d4a915 onto `release/1.10.x` succeeded!

PR #10136, opened 2021-04-27
ui: Add TProxy Mode notice banner to service instance Upstreams tab

kaxcode:
Adding a "warning" notice at the top of the service instance upstream tab. This gives the user more information about the upstreams list rendered. 

Screen Shot 2021-04-27 at 4 27 51 PM

### DIFF
--- a/ui/packages/consul-ui/app/templates/dc/services/instance/upstreams.hbs
+++ b/ui/packages/consul-ui/app/templates/dc/services/instance/upstreams.hbs
@@ -35,6 +35,30 @@ as |route|>
           @filter={{filters}}
         />
       {{/if}}
+      {{#if (eq proxy.ServiceProxy.Mode 'transparent')}}
+      <Notice
+        @type="warning"
+      as |notice|>
+        <notice.Header>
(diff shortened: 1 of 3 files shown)


ui/packages/consul-ui/app/templates/dc/services/instance/upstreams.hbs
```diff
@@ -35,6 +35,15 @@ as |route|>
...
       {{/if}}
+      {{#if (eq proxy.ServiceProxy.Mode 'transparent')}}
+        <Consul::Notice
+          @type="warning"
+          @for="tproxy-mode"
+          @link="{{env 'CONSUL_DOCS_URL'}}/connect/transparent-proxy"
+          @internal={{false}}
+          @action={{true}}
+        />
+      {{/if}}
```
johncowen, line 46: I'd prefer to use a normal `<Notice />` component here, (then we keep the `h3` etc in the top level route template) would it be ok to change that out?

kaxcode: @johncowen Updated to use `<Notice /> `

johncowen (approved):
:shipit:

PR #10146, opened 2021-04-28
connect/proxy: fix a few problems with tests

dnephin:
We noticed that `TestUpstreamListener` would deadlock sometimes when run
with the race detector ([example](https://app.circleci.com/pipelines/github/hashicorp/consul/18113/workflows/d9d32c2f-9adb-4171-8957-734f50786d70/jobs/358395/tests)). While debugging this issue I found and fixed the
following problems.

1. the net.Listener was not being closed properly when Listener.Stop was
   called. This caused the Listener.Serve goroutine to run forever.
   Fixed by storing a reference to net.Listener and closing it properly
   when Listener.Stop is called.
2. call connWG.Add in the correct place. WaitGroup.Add must be called
   before starting a goroutine, not from inside the goroutine.
3. Set metrics config EnableRuntimeMetrics to `false` so that we don't
   start a background goroutine in each test for no reason. There is no
   way to shutdown this goroutine, and it was an added distraction while
   debugging these timeouts.
5. two tests were calling require.NoError from a goroutine.
   require.NoError calls t.FailNow, which MUST be called from the main
   test goroutine. Instead use t.Errorf, which can be called from other
   goroutines and will still fail the test.
6. `assertCurrentGaugeValue` was breaking out of a for loop, which
   would cause the `RWMutex.RUnlock` to be missed. Fixed by calling
   unlock before `break`.

The core issue of a deadlock  was fixed by armon/go-metrics#124. If that merges soon I will include the vendor update in this PR.

### DIFF
--- a/connect/proxy/listener.go
+++ b/connect/proxy/listener.go
@@ -10,10 +10,11 @@ import (
 	"time"
 
 	metrics "github.com/armon/go-metrics"
+	"github.com/hashicorp/go-hclog"
+
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/consul/connect"
 	"github.com/hashicorp/consul/ipaddr"
-	"github.com/hashicorp/go-hclog"
 )
 
 const (
@@ -44,6 +45,7 @@ type Listener struct {
 	// `connection refused`. Retry loops and sleeps are unpleasant workarounds and
 	// this is cheap and correct.
 	listeningChan chan struct{}
+	listener      net.Listener
 
 	logger hclog.Logger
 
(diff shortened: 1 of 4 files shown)


connect/proxy/listener.go
```diff
@@ -171,7 +175,6 @@ func (l *Listener) handleConn(src net.Conn) {
 
 	// Make sure Close() waits for this conn to be cleaned up. Note defer is
 	// before conn.Close() so runs after defer conn.Close().
-	l.connWG.Add(1)
 	defer l.connWG.Done()
```
rboyer, line 178: I think this needs to be the very first line of this function so it pairs well with the WaitGroup.Add

rboyer (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/358670.

PR #10154, opened 2021-04-29
docs(discovery/service): Clarify multiple service definitions

fhemberger:
Be more explicit that the definition of multiple services only works in config files,
not using the HTTP API.

Ref: https://discuss.hashicorp.com/t/register-multiple-services-via-put-request/

### DIFF
--- a/website/content/docs/discovery/services.mdx
+++ b/website/content/docs/discovery/services.mdx
@@ -27,8 +27,8 @@ To configure a service, either provide the service definition as a
 `-config-file` option to the agent or place it inside the `-config-dir` of the
 agent. The file must end in the `.json` or `.hcl` extension to be loaded by
 Consul. Check definitions can be updated by sending a `SIGHUP` to the agent.
-Alternatively, the service can be registered dynamically using the [HTTP
-API](/api).
+Alternatively, the service can be [registered dynamically](/api-docs/agent/service#register-service)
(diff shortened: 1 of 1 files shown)


website/content/docs/discovery/services.mdx
```diff
@@ -276,7 +276,8 @@ deprecated and has been removed as of Consul 1.1.
 ## Multiple Service Definitions
 
 Multiple services definitions can be provided at once using the plural
```
blake, line 278:
```suggestion
Multiple services definitions can be provided at once when registering services via the agent configuring by using using the plural `services` key.
```
What do you think about using this phrasing instead?

website/content/docs/discovery/services.mdx
```diff
@@ -275,8 +275,9 @@ deprecated and has been removed as of Consul 1.1.
 
 ## Multiple Service Definitions
 
-Multiple services definitions can be provided at once using the plural
-`services` key in your configuration file.
+Multiple services definitions can be provided at once when registering services
+via the agent configuring by using the plural `services` key (registering
```
blake, line 279:
```suggestion
via the agent configuration by using the plural `services` key (registering
```

blake (approved):
Requested one minor change. Otherwise, LGTM!

fhemberger:
Ok, fixed. Thanks for the review!

PR #10155, opened 2021-04-29
config-entry: remove Kind and Name field from Mesh config entry

dnephin:
Related to #10067, making this change to the new config entry before it is released makes it much easier to remove these fields.

No config entry needs a Kind field. It is only used to determine the Go type to target. As we introduce new config entries (like this one) we can remove the kind field and have the `GetKind` method return the single supported value.

In this case (similar to proxy-defaults) the Name field is also unnecessary. We always use the same value. So we can omit the name field entirely. In the future if we want to support separate names, we can introduce the field in a backwards compatible way, where empty string means the old default name.

TODO:
* [x] test marshalling the value to the HTTP API. I guess with this change we no longer have a `kind` field in the output. Marshaling will need to add the kind field based on the type.

### DIFF
--- a/agent/config_endpoint_test.go
+++ b/agent/config_endpoint_test.go
@@ -7,10 +7,11 @@ import (
 	"net/http/httptest"
 	"testing"
 
-	"github.com/hashicorp/consul/agent/structs"
-	"github.com/hashicorp/consul/testrpc"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/require"
+
+	"github.com/hashicorp/consul/agent/structs"
+	"github.com/hashicorp/consul/testrpc"
 )
 
 func TestConfig_Get(t *testing.T) {
@@ -45,6 +46,18 @@ func TestConfig_Get(t *testing.T) {
 				Config: map[string]interface{}{
 					"foo": "bar",
 					"bar": 1,
+				},
+			},
+		},
+		{
+			Datacenter: "dc1",
+			Entry: &structs.MeshConfigEntry{
+				TransparentProxy: structs.TransparentProxyMeshConfig{
+					CatalogDestinationsOnly: true,
+				},
+				Meta: map[string]string{
+					"key1": "value1",
+					"key2": "value2",
 				},
 			},
 		},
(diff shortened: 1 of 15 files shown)


agent/config_endpoint_test.go
```diff
@@ -95,6 +108,37 @@ func TestConfig_Get(t *testing.T) {
 		_, err := a.srv.Config(resp, req)
 		require.Error(t, errors.New("Must provide either a kind or both kind and name"), err)
 	})
+	t.Run("get the single mesh config", func(t *testing.T) {
+		req, _ := http.NewRequest("GET", "/v1/config/mesh/mesh", nil)
```
dnephin, line 112: Querying for the entry still required me to specify the "name" in the URL because we haven't special cased this type in the memdb indexer yet. An example of that can be seen in #10067. This would allow someone to query for `/v1/config/mesh/`

agent/config_endpoint_test.go
```diff
@@ -95,6 +108,37 @@ func TestConfig_Get(t *testing.T) {
...
+		require.NoError(t, err)
+
+		ce, ok := obj.(*structs.MeshConfigEntry)
+		require.True(t, ok, "wrong type %T", obj)
+		// Set indexes to expected values for assertions
+		ce.CreateIndex = 12
+		ce.ModifyIndex = 13
+
+		out, err := a.srv.marshalJSON(req, obj)
+		require.NoError(t, err)
```
dnephin, line 124: It is unfortunate that our API tests don't test against a marshalled response. That seems like a potentially large place for bugs to hide. Here I am calling the underlying function that our HTTP framework calls. It's probably an adequate simulation for now, but seems like it could break in the future.

Also not being able to test for the response code or response headers is really unfortunate.

freddygv: Alternatively you could add a case to `TestAPI_ConfigEntries` in the api module. Wouldn't that verify that the kind was serialized?

freddygv (approved):

freddygv (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/362571.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 347f3d212843d980de19c1c3c1da152111ed4f4f onto `release/1.10.x` succeeded!

PR #10162, opened 2021-04-30
Only consider virtual IPs for transparent proxies

freddygv:
Initially we were loading every potential upstream address into Envoy
and then routing traffic to the logical upstream service. The downside
of this behavior is that traffic meant to go to a specific instance
would be load balanced across ALL instances.

Traffic to specific instance IPs should be forwarded to the original
destination and if it's a destination in the mesh then we should ensure
the appropriate certificates are used.

This PR makes transparent proxying a Kubernetes-only feature for now 
since support for other environments requires generating virtual IPs, 
and Consul does not do that at the moment.

### DIFF
--- a/agent/xds/listeners.go
+++ b/agent/xds/listeners.go
@@ -146,8 +146,6 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
 			return nil, err
 		}
 
-		// For filter chains used by the transparent proxy, we need to match on multiple destination addresses.
-		// These might be: the ClusterIP in k8s, or any of the service instance addresses.
 		endpoints := cfgSnap.ConnectProxy.WatchedUpstreamEndpoints[id]
 		uniqueAddrs := make(map[string]struct{})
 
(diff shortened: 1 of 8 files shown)


agent/xds/listeners.go
```diff
@@ -32,6 +32,8 @@ import (
 	"github.com/hashicorp/consul/sdk/iptables"
 )
 
+const virtualIPTag = "virtual"
```
rboyer, line 35: Would this be better off in `agent/structs`?

rboyer (approved):
LGTM minus the one comment.

agent/xds/listeners.go
dnephin: If we don't need to export it now, I think it's desirable to be able to keep it locally unexported. We can always move it and export it later, but if we don't need to it's one less thing that has to be changed when we need to split up the `agent/structs` package.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/361369.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 2ca3f481f8d6fc54040683cce06fe6ccac168bb7 onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/361369)

Issue #10053, opened 2021-04-16
Transparent proxy doesn't work with service resolver redirect


tjhiggins:
#### Overview of the Issue

Tried out the hashicorp/consul:1.10.0-beta1 because we were excited about potentially using the transparent proxy when its released. It seems to work for explicit service upstreams, but doesn't work when refing a service resolver that has a redirect to a service. Switching from mode transparent to direct allows traffic again. 

#### Reproduction Steps

Steps to reproduce this issue, eg:

1. Install consul with image hashicorp/consul:1.10.0-beta1
1. Create service A
1. Create service resolver B that redirects to A
1. Create service C that refs B as an upstream destination

#### Log Fragments

The following is the envoy sidecar logs for service C

When mode is direct I see the listener added:
```console
[info][upstream] [source/server/lds_api.cc:79] lds: add/update listener 'public_listener:10.1.1.120:21000'
[info][upstream] [source/server/lds_api.cc:79] lds: add/update listener 'B:127.0.0.1:12345'
[info][upstream] [source/server/lds_api.cc:79] lds: add/update listener 'outbound_listener:127.0.0.1:15001'
```

However, when the mode is transparent no listener for B gets added:
```console
[info][upstream] [source/server/lds_api.cc:79] lds: add/update listener 'public_listener:10.1.1.120:21000'
[info][upstream] [source/server/lds_api.cc:79] lds: add/update listener 'outbound_listener:127.0.0.1:15001'
```

Using service A does work when mode is transparent, but we want to use service resolvers:
```console
[info][upstream] [source/server/lds_api.cc:79] lds: add/update listener 'public_listener:10.1.1.120:21000'
[info][upstream] [source/server/lds_api.cc:79] lds: add/update listener 'A:127.0.0.1:12345'
[info][upstream] [source/server/lds_api.cc:79] lds: add/update listener 'outbound_listener:127.0.0.1:15001'
```

jsosulska:
Hi @tjhiggins 

Thanks for reporting this, and welcome to the Consul Repo! The engineering team is looking into this more, and will link the fix to this issue when it is submitted.

Happy coding!

tjhiggins:
@freddygv thank you for fixing! I'll keep any eye out for the next beta release and try again.

freddygv:
@tjhiggins awesome! We should be releasing beta2 shortly. 

The main caveat is mentioned in the PR:

> NOTE: this implementation requires an intention to the redirecting
[...]

PR #10168, opened 2021-05-03
Fixup discovery chain handling in transparent mode

freddygv:
Previously we would associate the address of a discovery chain target
with the discovery chain's filter chain. This was broken for a few reasons:

- If the upstream is a virtual service, the client proxy has no way of
dialing it because virtual services are not targets of their discovery
chains. The targets are distinct services. This is addressed by watching
the endpoints of all upstream services, not just their discovery chain
targets.

- If multiple discovery chains resolve to the same target, that would
lead to multiple filter chains attempting to match on the target's
virtual IP. This is addressed by only matching on the upstream's virtual
IP.

**NOTE:** this implementation requires an intention to the redirecting 
virtual service and not just to the final destination. This is how 
we can know that the virtual service is an upstream to watch.

A later PR will look into traversing discovery chains when computing 
upstreams so that intentions are only required to the discovery chain 
targets.

TODO:
- [x] Add proxycfg/state tests
- [x] Add xds/listeners tests

Fixes: #10053

### DIFF
--- a/agent/proxycfg/state.go
+++ b/agent/proxycfg/state.go
@@ -1020,12 +1020,12 @@ func (s *state) resetWatchesFromChain(
 	}
 
 	var (
-		watchedChainSource bool
-		needGateways       = make(map[string]struct{})
+		watchedChainEndpoints bool
+		needGateways          = make(map[string]struct{})
 	)
 	for _, target := range chain.Targets {
 		if target.ID == chain.ID() {
-			watchedChainSource = true
+			watchedChainEndpoints = true
 		}
 
 		s.logger.Trace("initializing watch of target",
@@ -1069,7 +1069,7 @@ func (s *state) resetWatchesFromChain(
 	//
 	// Outside of transparent mode we only watch the chain target, B,
 	// since A is a virtual service and traffic will not be sent to it.
-	if !watchedChainSource && s.proxyCfg.Mode == structs.ProxyModeTransparent {
+	if !watchedChainEndpoints && s.proxyCfg.Mode == structs.ProxyModeTransparent {
 		s.logger.Trace("initializing watch of chain source",
 			"upstream", id,
 			"chain", chain.ServiceName,
--- a/agent/proxycfg/state.go
+++ b/agent/proxycfg/state.go
@@ -1023,8 +1023,10 @@ func (s *state) resetWatchesFromChain(
 		watchedChainEndpoints bool
 		needGateways          = make(map[string]struct{})
 	)
+
+	chainID := chain.ID()
 	for _, target := range chain.Targets {
-		if target.ID == chain.ID() {
+		if target.ID == chainID {
 			watchedChainEndpoints = true
 		}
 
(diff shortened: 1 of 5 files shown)


agent/structs/discovery_chain.go
```diff
@@ -88,6 +88,12 @@ func (c *CompiledDiscoveryChain) IsDefault() bool {
 	return target.Service == c.ServiceName && target.Namespace == c.Namespace
 }
 
+// ID returns an ID that endoces the service, namespace, and datacenter.
```
rboyer, line 91:
```suggestion
// ID returns an ID that encodes the service, namespace, and datacenter.
```

agent/proxycfg/state.go
```diff
@@ -1019,8 +1019,15 @@ func (s *state) resetWatchesFromChain(
 		cancelFn()
 	}
 
-	needGateways := make(map[string]struct{})
+	var (
+		watchedChainEndpoints bool
+		needGateways          = make(map[string]struct{})
+	)
 	for _, target := range chain.Targets {
+		if target.ID == chain.ID() {
```
rboyer, line 1027: trivial: You could compute `chain.ID()` outside of the loop since it's static.

agent/proxycfg/state.go
```diff
@@ -1054,6 +1061,39 @@ func (s *state) resetWatchesFromChain(
...
+			chain.Datacenter,
+			"",
+			&entMeta,
+		)
+		if err != nil {
+			cancel()
+			return err
+		}
+
+		snap.WatchedUpstreams[id][chain.ID()] = cancel
```
rboyer, line 1094: Could you move this into a helper of some form since it's nearly identical to the one in the loop immediately preceding?

agent/xds/listeners.go
```diff
@@ -146,19 +146,21 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
 			return nil, err
 		}
```
rboyer, line 148: It might be helpful to add a small comment around line 135 above (I can't click there in GH) above:
> `filterChain, err := s.makeUpstreamFilterChainForDiscoveryChain`
To imply that the rest of the for-loop after that point is just for tproxy.

rboyer (approved):
LGTM with comments

agent/proxycfg/state.go
freddygv: Tried a couple options and made some changes here but I'm not sure it's much better, thoughts @rboyer ?

rboyer (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/361968.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit ed1082510dc80523b1f2a3a740fa5a13c77594f9 onto `release/1.10.x` succeeded!

PR #10169, opened 2021-05-03
docs: add streaming to api-docs

dnephin:
Second draft of API docs for streaming


website/content/api-docs/health.mdx
```diff
@@ -197,9 +197,10 @@ the following selectors and filter operations being supported:
 | `ServiceTags` | In, Not In, Is Empty, Is Not Empty                 |
 | `Status`      | Equal, Not Equal, In, Not In, Matches, Not Matches |
 
-## List Nodes for Service
+((list-nodes-for-service))
```
dnephin, line 200: hmm, this did not work. I'm trying to find a way to keep the link text the same while changing the readable text.  I'll try something else.

blake:
```suggestion
## List Nodes for Service ((#list-nodes-for-service))
```
This format should do it.

website/content/api-docs/features/streaming.mdx
```diff
@@ -0,0 +1,43 @@
...
+description: |-
+  Some API endpoints support a feature known as "streaming". Change events
+  are published to a stream topic, allowing clients to build a localized
+  view, and removing the need to send full state-of-the-world responses
+  every time data changes.
+---
+
+# Streaming
+
+A few endpoints in Consul support a feature known as "streaming". With streaming
```
banks, line 13: Hmmm. I'm not sure if we should document streaming as a "feature" that is different from blocking queries 🤔 .

To me streaming is an implementation detail - an important one, but something more fitting for "internals" or "architecture" docs than as an "API feature".

The design goal of streaming was to be effectively transparent to the user - the public API wasn't intended to change at all, just the transport used to power it become much more efficient.

I think the fact that we have it only partially implemented with fallbacks in complex cases and slight backwards incompatibilites like different headers being returned or cache options supported when an endpoint is using "streaming" is a little unfortunate and I definitely think we should try to make it clear where that is the case and why.

But I wonder if it's more confusing to try to distinguish "streaming" and "blocking queries" as separate things from an API point of view? To the API consumer they are both still "blocking queries" IMO since they still use the same long-poll and get a fill response back every timeout.

What do you think about having a page with some of the internal details of streaming vs blocking transports under `Architecture` in the sidebar, and then annotate the specific endpoints with notes that "this endpoint uses streaming transport" or similar with a link the those docs and call out the different headers supported?

We could also maybe add a section to the Blocking Queries page under API Features to note that some endpoints support a more efficient streaming transport and those are clearly called out and might support different headers etc?

dnephin: I don't think architecture is the right place for these docs. From the users perspective Streaming is part of the API functionality and this section seems like the right place. 

I think we can use a single page for both blocking queries and streaming. What we are missing is a term to distinguish the long-polling from the HTTP API, from the backend used by the long polling.

banks (approved):
Nice! Thanks I think this looks great.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/383074.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit ac65f3adc537b8b27b26dfc4b990de1136f70ed4 onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/383074)

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/383123.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit ac65f3adc537b8b27b26dfc4b990de1136f70ed4 onto `release/1.10.x` succeeded!

PR #10174, opened 2021-05-04
ui: Adds ability to show a 'partial' list in list-collections

johncowen:
We have a component that we use quite a lot called `<ListCollection />`, we use it for our 'virtual/DOM-recycling scrolling' and just for normal 'native scrolling' lists.

This PR adds a `@partial` attribute to this component, letting you specify that you only want a certain number of rows to show initially. When this attribute is set, we also render a 'View more' button to let the user expand the list to show all the rows instead of just a 'partial' amount.

partial-list

We also took the opportunity here to move the `list-collection` styles into the `app/components/list-collection` folder, so everything is co-located in the one folder.

### DIFF
--- a/.changelog/10174.txt
+++ b/.changelog/10174.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: Only show a partial list of intention permissions, with the option to show all
+```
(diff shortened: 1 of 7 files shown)


kaxcode (approved):
Very cool 😎 LGTM

PR #10176, opened 2021-05-04
mog: automatically support arbitrary struct/pointer conversions

rboyer:


devtools/mog/generate.go
```diff
@@ -172,20 +164,39 @@ func sourceFieldMap(fields []fieldConfig) map[string]fieldConfig {
...
+	PtrTo, ToPtr, PtrToPtr *ast.FuncDecl
+
+	From                         *ast.FuncDecl
+	PtrFrom, FromPtr, PtrFromPtr *ast.FuncDecl
 
 	// TODO: RoundTripTest *ast.FuncDecl
 }
 
+func (g *generated) Decls() []ast.Decl {
+	return []ast.Decl{
```
rboyer, line 177: This will likely generate a bunch of dead code, unless the pointerness is inlined into the caller which we could improve upon later.

devtools/mog/generate.go
```diff
@@ -213,12 +224,155 @@ func generateToFunc(cfg structConfig, imports *imports) *ast.FuncDecl {
 	}
 }
 
-func funcNameTo(cfg structConfig) string {
-	return cfg.Source + "To" + cfg.FuncNameFragment
+func generatePtrToPtrFunc(cfg structConfig, imports *imports) *ast.FuncDecl {
```
rboyer, line 227: I suspect these variations would be slightly less absurd if we bootstrapped them off of pre-baked strings using:
```
	fset := token.NewFileSet()
	f, err := parser.ParseFile(fset, "", src, 0)
```

somehow

devtools/mog/config.go
```diff
@@ -62,6 +62,29 @@ type fieldConfig struct {
 	FuncFrom   string
 	FuncTo     string
 	// TODO: Pointer pointerSettings
+
+	cfg    structConfig // for dynamic
+	cfgSet bool
```
dnephin, line 67: I guess this is a flag to say that we found the referenced struct? Where as previously it would generate code that did not compile?

devtools/mog/config.go
```diff
@@ -176,7 +201,7 @@ func parseFieldAnnotation(field *ast.Field) (fieldConfig, error) {
 		case "target":
 			c.TargetName = value
 		case "pointer":
-			// TODO:
+			// TODO(rb): remove as unnecessary?
```
dnephin, line 204: The types of options that I expected here were things like:

1. how to deal with zero values when converting to a pointer. Should they be converted to `nil`, or a pointer to the zero value
2. if both the source and target pointers are of the same type, should they reference the same value, or be a copy. But I realize now that is more general than pointers. It should apply to maps and slices as well.

devtools/mog/generate.go
dnephin: Mentioned in slack, but I am wondering if we should change the default function definitions to something like the below, so that we can use a single set of functions for all of these:

```
func SourceFromTarget(t *Target, s *Source) {}
func SourceToTarget(s *Source, *Target) {}
```

devtools/mog/generate.go
```diff
@@ -172,20 +164,39 @@ func sourceFieldMap(fields []fieldConfig) map[string]fieldConfig {
...
 	targetType := &ast.SelectorExpr{
 		X:   &ast.Ident{Name: path.Base(imports.AliasFor(cfg.Target.Package))},
 		Sel: &ast.Ident{Name: cfg.Target.Struct},
 	}
 
+	funcName := funcNameTo(cfg, false, false)
+
 	return &ast.FuncDecl{
-		Name: &ast.Ident{Name: funcNameTo(cfg)},
+		Doc:  newComments(funcName + " is a struct-to-struct conversion that translates FROM a protobuf"),
```
dnephin, line 198: I don't think we should hard code anything about protobuf here. The tool could be used to convert between `api` types and `structs` types, or `config` types and `structs` types, or used in other projects that have similar needs.

devtools/mog/config.go
```diff
@@ -62,6 +62,29 @@ type fieldConfig struct {
 	FuncFrom   string
 	FuncTo     string
 	// TODO: Pointer pointerSettings
+
+	cfg    structConfig // for dynamic
```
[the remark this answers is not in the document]
rboyer: This is the structConfig corresponding to the struct definition for the type of the field, not the enclosing field.

devtools/mog/config.go
rboyer: It wouldn't be set if it wasn't actually pointing to a struct type (like `ID string` fields in the e2e test)

devtools/mog/generate.go
rboyer: Pushing a followup commit that does just that^^

devtools/mog/generate.go
rboyer: removed the comments both for this reason and because now there aren't 8 functions per type

dhiaayachi:
This now live in it's own repo: https://github.com/hashicorp/mog

PR #10180, opened 2021-05-04
website: support hidden pages in nav-data

zchsh:

PR #10197, opened 2021-05-06
website: support hidden pages in nav-data (#10180)

zchsh:
This PR cherry-picks commits from #10180.


zchsh:
I don't have write access, but I think this PR needs the following labels: `pr/no-changelog` `type/docs`

Issue #10011, opened 2021-04-13
Set Content-Type header for requests from Consul Go API


blake:
#### Feature Description

Consul's Go API does not set the Content-Type header in any of its requests (notably, in PUT and POST requests it makes). Consul itself handles this fine and it works as expected. Some proxies or firewalls will reject HTTP requests that do not have the Content-Type header set. If you have one of these between the client and Consul, the lack of this header will cause requests to fail.

In order to avoid this situation, the API package should set the appropriate Content-Type header when sending requests to the Consul API. [Section 3.5.1](https://tools.ietf.org/html/rfc7231#section-3.1.1.5) from RFC 7231, Hypertext Transfer Protocol (HTTP/1.1): Semantics and Content, suggests that:

> A sender that generates a message containing a payload body SHOULD generate a Content-Type header field in that message unless the intended media type of the enclosed representation is unknown to the sender.

The content-type header is not explicitly required to be sent, but doing so will help address the aforementioned scenario where the request is dropped by an intermediary firewall/proxy.

#### Use Case(s)

This surfaces most prominently when using the official Consul Terraform Provider to configure Consul. It uses the v1.5.2 Go Consul API to perform its API requests and this issue prevents the Terraform runs from completing successfully if you have a firewall or proxy rejecting requests because of the absence of the Content-Type header.

mikemorris:
Fixed in https://github.com/hashicorp/consul/pull/10204#issuecomment-847955418

PR #10204, opened 2021-05-07
Always set the Content-Type header when a body is present

remilapeyre:
Closes https://github.com/hashicorp/consul/issues/10011


api/operator_license.go
```diff
@@ -100,6 +100,7 @@ func (op *Operator) LicensePut(license string, opts *WriteOptions) (*LicenseRepl
 	r := op.c.newRequest("PUT", "/v1/operator/license")
 	r.setWriteOptions(opts)
 	r.body = strings.NewReader(license)
+	r.header.Set("Content-Type", "application/octet-stream")
```
banks, line 103: Great job finding these exceptions! How did you find them out of interest? Did tests break for these endpoints because of the wrong content type?

I can't think of any other exceptions but then I doubt I'd have remembered this one either so wondering if there is anything else we should do to be sure we caught all the non-json endpoints.

If you already went through exhaustively though then great!

banks (commented):
Thanks @remilapeyre this is a great PR.

I think it can probably be merged as-is I just had one Q about those non-JSON exceptions and whether we need to do any more diligence to make sure we didn't miss any other edge cases?

remilapeyre:
I just remembered those two and looked briefly at the documentation and thought I had found them all. Looking at the output of `rg -P 'curl.*?```' website/content/api-docs/ --multiline --multiline-dotall` actually finds more cases where the payload sent is not JSON. I've added them but I'm not sure if it's possible to be sure to have found them all without doing something more complicated.

banks (changes requested):
Oh nice idea!

I tried a few things in the code to pick up possible others:

```
$ rg -P 'io.Copy' agent/*endpoint.go
agent/operator_license_endpoint.go
37:	if _, err := io.Copy(buf, req.Body); err != nil {

agent/kvs_endpoint.go
219:	if _, err := io.Copy(buf, req.Body); err != nil {

agent/event_endpoint.go
49:		if _, err := io.Copy(&buf, req.Body); err != nil
```

Which you've already got all of anyway.

I also tried:

```
rg -P '\.Body' agent/*.go | rg -v 'decodeBody\(|DecodeJSON\(|_test.go'
agent/watch_handler.go:		defer resp.Body.Close()
agent/watch_handler.go:		io.Copy(output, resp.Body)
agent/txn_endpoint.go:	req.Body = http.MaxBytesReader(resp, req.Body, maxTxnLen)
agent/txn_endpoint.go:							Body:                           check.Definition.Body,
agent/snapshot_endpoint.go:		if err := s.agent.delegate.SnapshotRPC(&args, req.Body, resp, nil); err != nil {
agent/operator_license_endpoint.go:	if _, err := io.Copy(buf, req.Body); err != nil {
agent/namespace_endpoint.go:	if req.Body == nil {
agent/kvs_endpoint.go:	if _, err := io.Copy(buf, req.Body); err != nil {
agent/event_endpoint.go:		if _, err := io.Copy(&buf, req.Body); err != nil {
agent/http.go:	if req.Body == nil {
agent/http.go:	dec := json.NewDecoder(req.Body)
agent/acl_endpoint.go:	policyBytes, err := ioutil.ReadAll(req.Body)
agent/agent.go:				Body:            chkType.Body,
```

Which are all false-positives **except one**: https://www.consul.io/api-docs/acl#translate-rules expects a raw HCL ACL token policy as the body so it probably needs a different content type. So I think we need an update here too: https://github.com/hashicorp/consul/blob/cc5dba1e3b88cc670e75b1fafaf777536962e8c6/api/acl.go#L896-L918

I think with that we can merge!

Thanks for this Remi! Much appreciated.

remilapeyre:
Hi Paul I added the Content Type for the ACL and fixed the conflicts. I think it should be good now :)

banks (approved):
Perfect!

Thanks for getting this in Remi!

banks:
GitHub did something strange here when I merged - it errored on my several times but when I went to merge manually I found the merge actually worked and this is already in master: https://github.com/hashicorp/consul/commit/213524a6579bfe070015cc05970d82f335a94264

Despite that GH isn't marking this PR as merged and if I attempt to merge again it just creates another empty commit on master and fails again.

So I'll manually close this and go and do the 1.10 backport cherry-pick by hand.

PR #10212, opened 2021-05-10
ui: Support Route optional parameters/segments

johncowen:
This felt like a fairly significant change, so I did a fairly large write up:

This PR moves our URLs with 'optional namespace segment' into a separately abstracted 'optional URL segment' feature.

When Consul added namespace support we chose to detect 'optional' namespace parameters in the UIs URL, recognising a 'namespace' parameter via the presence of a `~` character at the beginning of the segment (a character that is not supported within DNS). This meant we could support URLs with the following format:

```
/ui/datacenter-1/services
/ui/~my-namespace/datacenter-1/services
```

Support for this led to various scattered areas in the codebase that had logic to support this, and also meant that there were two separate Routes (that we automatically created) for both of the above URLs just to support an 'optional parameter'.

In order to support our upcoming admin partition feature, we would like to use the same approach to allow an optional segment for admin partitions again by using a prefix that is not allowed via DNS `-`.

Instead of repeating what we had done for namespaces by duplicating and making the necessary code more complex, we decided to refactor the feature to only be about 'optional URL segments' rather than anything to do with namespaces or admin partitions.

## FSM Location

Ember's Router/Routing and related classes/services are rather complex and originate from the very beginning of ember, but they use the notion of 'Locations' to provide different ways of dealing with URL changes within your app. [Locations](https://api.emberjs.com/ember/release/classes/Location) are far more lightweight and easier to work with than working directly with Ember's routing classes, so we chose to write a set of custom locations in order to extract this feature into a more isolated approach.

Much of the work here is based on Ember's `history` location, and we've added 3 custom locations:

1. `fsm`: Based on the idea that routing could be seen as a finite state machine we've created a base FSM class which copies the interface of the browsers native History API and create a very simple custom Location
2. `fsm-with-optional`: This builds on our base `FSMLocation` to provide support for optional URL parameters throughout our application. We currently add our namespace support (and eventually our admin partition support) in here - further work will move this project/application specific code into one of our configuration files.
3. `fsm-with-optional-test`: Following the relationship between ember's `history` and `none` locations we add an equivalent `none` location here specifically for testing.

Whilst working on the above I noticed that the way that ember instantiates the initial route for your application differs between when you are running your app and when you are testing your app. Therefore there are come changes here so that how the initial route set is no longer different between prod/dev and test. As we do a lot of injection here, this was relatively straightforwards to do without too much churn. I also took the opportunity here to replace some jQuery-ness taken from ember-cli-page-object and replace it with what we already use in our app for query parameter processing.

These locations are now the centralized place to deal with anything URL related, therefore `href-to` now gathers what it needs in order to print URLs from the location also.

## href-mut

As its clearer now how our application deals with URLs we also removed our unfortunately named `href-mut` helper and gave `href-to` the ability to give you URLs from the same URL you are already on, but with different parameters. We can do this now be specifying `.` (meaning current URL) as the first argument (the routeName) to `href-to`. At this point we felt that it was time to remove the `ember-href-to` plugin. We still use a little code from `ember-href-to` and that will probably be reduced further at a later date. We removed the last instance we had of ember's old `observer` here also, so that's one less deprecation from our list.

## optionalParams()

Ideally we would have liked to have used `paramsFor` and the various `params` arguments in model hooks to access these 'optional' parameters. They are just 'params' just like all our other parameters in URLs, they just happen to be optional. Unfortunately this proved to be a little difficult to integrate into ember, so we (hopefully temporarily) went with a new `optionalParams` method that is available on all routes. `optionalParams` returns a hash/object of the configured optionalParams, defaulting each of them to an empty string (this decision was based on the experience of what we needed for our namespace support and felt like it made sense)

Overall our application code is much slimmer and our namespace/admin partition support is now not really specific to our application, and we no longer duplicate all of our Routes to cope for the presence of an extra parameter. The usage of 'special URL segment prefix characters' like `~` are now removed from the application as a whole, and longer term will (hopefully) just be part of a line of configuration in our routing configuration.


ui/packages/consul-ui/app/locations/fsm-with-optional.js
```diff
@@ -0,0 +1,316 @@
...
+
+  formatURL(url, optional, withOptional = true) {
+    if (url !== '') {
+      // remove trailing slashes if they exists
+      this.rootURL = this.rootURL.replace(/\/$/, '');
+      this.baseURL = this.baseURL.replace(/\/$/, '');
+    } else if (this.baseURL[0] === '/' && this.rootURL[0] === '/') {
+      // if baseURL and rootURL both start with a slash
+      // ... remove trailing slash from baseURL if it exists
+      this.baseURL = this.baseURL.replace(/\/$/, '');
```
johncowen, line 251: See if we can precompile and hoist these trailing slash regex's

ui/packages/consul-ui/app/locations/fsm-with-optional.js
```diff
@@ -0,0 +1,316 @@
...
+    // this.optional = {};
+    console.log(path, 'setURL');
+    this.changeURL('push', path);
+  }
+
+  replaceURL(path) {
+    console.log(path, 'replaceURL');
+    this.changeURL('replace', path);
+  }
+
```
johncowen, line 290: remove console.logs

ui/packages/consul-ui/app/components/consul/upstream/list/index.hbs
```diff
@@ -24,15 +24,22 @@ as |item index|>
         </Tooltip>
       </dd>
     </dl>
-    {{#if (and (env 'CONSUL_NSPACES_ENABLED') (not-eq item.Namespace @nspace))}}
-      <a data-test-service-name href={{href-to 'nspace.dc.services.show' (concat '~' item.Namespace) @dc item.Name }}>
+{{#if (and (env 'CONSUL_NSPACES_ENABLED') (not-eq item.Namespace @nspace))}}
```
johncowen, line 27: Does this need to use `can` now?

ui/packages/consul-ui/app/locations/fsm-with-optional.js
```diff
@@ -0,0 +1,316 @@
...
+    // remove trailing slashes if they exists
+    url = url || this.location.pathname;
+    this.rootURL = this.rootURL.replace(/\/$/, '');
+    this.baseURL = this.baseURL.replace(/\/$/, '');
+
+    // remove baseURL and rootURL from start of path
+    return url
+      .replace(new RegExp(`^${this.baseURL}(?=/|$)`), '')
+      .replace(new RegExp(`^${this.rootURL}(?=/|$)`), '')
+      .replace(/\/\//g, '/'); // remove extra slashes
```
johncowen, line 131: Another pre-compilable regex

ui/packages/consul-ui/app/locations/fsm-with-optional.js
```diff
@@ -0,0 +1,316 @@
...
+  hrefTo(routeName, params, hash) {
+    if (typeof hash.dc !== 'undefined') {
+      delete hash.dc;
+    }
+    if (typeof hash.nspace !== 'undefined') {
+      hash.nspace = `~${hash.nspace}`;
+    }
+    if (typeof hash.partition !== 'undefined') {
+      hash.partition = `-${hash.partition}`;
+    }
```
johncowen, line 192: Comment this out for now

johncowen (commented):
Adding review notes

ui/packages/consul-ui/app/components/consul/upstream/list/index.hbs
kaxcode: We decided we would make a new ability called `canSee()`.

ui/packages/consul-ui/app/routes/dc/nodes/show/index.js
```diff
@@ -7,6 +7,7 @@ export default class IndexRoute extends Route {
       .split('.')
       .slice(0, -1)
       .join('.');
+    model = this.modelFor(parent);
```
kaxcode, line 10: what's this about?

johncowen: Oooooh I can't quite remember, lemme take a look 👀

johncowen: No idea, must have just looked weird to me when I was in here, both model coming from the arguments and model coming from parent are the same (as we have no model hook here I guess?).

Happy to leave in or take out really, I suppose leaving it in means that if we do add a model hook here, this hook still works as expected, so I'd if it was me I'd go with leaving it in. 🤷

ui/packages/consul-ui/app/components/consul/upstream/list/index.hbs
johncowen: I know I said `canSee` yesterday and I know this is quibbling over words, but I was just about to do this and figured that it's not about whether you can see them or not, its about whether you can use them.. ha I just thought its about whether you have them or not, but maybe `canHaz()` `can('haz nspaces')` isn't the best 😆 

Think we can go with canAccess or canUse or something similar 🤔

johncowen: Oh look I just saw this:

https://github.com/hashicorp/consul/blob/ce958e72180d80fb8da054a9483a8172f83076ce/ui/packages/consul-ui/app/abilities/acl.js#L22-L24

I'll go with `canUse` then 👍

johncowen:
ok @kaxcode I made those couple of tweaks we discussed yesterday, so this is ready again now I think.

Quick note on the swap of `env 'CONSUL_NSPACES_ENABLED'` for `can 'use nspaces'` - There are probably a tonne of places where this would be best to swap, but I'm happy to leave this for whenever in a future PR, theres no change in functionality its just a preference to use `can` over `env` and then consistency - I might just do it as one of my 'take a break' PRs at some point.

kaxcode (approved):
I took another look and I believe I understand all the changes. LGTM - just needs a changelog.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/374912.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 2e4c9f5330da384c5763e6f0d8b8896b013adaef onto `release/1.10.x` succeeded!

Issue #10095, opened 2021-04-23
docs: Session Lock read response documentation update


johncowen:
The documentation for the read response for Session Locks uses the old pre-1.7 shape:

https://www.consul.io/api-docs/session#sample-response-2

This was changed in 1.7:

```javascript
// pre1.7
"Checks": ["serfHealth", ...]

//post-1.7
"NodeChecks": ["serfHealth", ...],
"ServiceChecks": null
```

Creating/updating sessions accepts both new style and old style payloads for back compat reasons, but guessing we should add something to the documentation around that also.

PR #9288, opened 2020-11-27
ui: Modifier based tooltips

johncowen:
This replaces the innards of our `<Tooltip />` component (which was based upon `ember-tooltip`) with a modifier based tooltip:

```
<div {{tooltip "Here's the tooltip"}} />
```

A modifier based approach feels a little better (add a tooltip to this DOM element), plus the underlying library for the modifier is tippy.js v6 which solves some bugbears with our previous Tooltip component:

1. Auto positioning functions much better.
2. Tooltips completely clean up after themselves (they don't leave any DOM lying around once the tooltip is hidden)
3. Prevents accidental insertion of Tooltips into elements where they don't belong (adding a tooltip component into a `dl` which results in `dl > div`.

I've rebuilt out `<Tooltip />` component to just use the tooltip modifier, but I would assume the majority of out usage of 'Tooltip' component could just be switched out to use the modifier.

Longer term, it would be nice to create a similar `{{popover}}` for our more generic popovers to avoid all of our z-index dancing.

Notes:

1. I removed a whole bunch of code that was only there to deal with tooltips, mainly CSS but also quite a bit of JS in the disco-chain component.
2. I tweaked the `{{dom-position}}` helper while I was here - I could remove some of our weird positioning in discovery-chain which seemed to affect this, so I took the opportunity to simplify this a little.
3. I tweaked the copy-button component to use this modifier approach rather than the old component tooltip approach.

Lastly, I would have like to make an additional `{{popover}}` modifier here, but I wanted to avoid too much change in this PR, so as well as adding this additional modifier, the tooltip modifier will probably change somewhat moving forwards.

PR #10225, opened 2021-05-11
ui: Miscellaneous Lock Session fixes

johncowen:
Whilst reviewing https://github.com/hashicorp/consul/pull/9930 and https://github.com/hashicorp/consul/pull/10121 we noticed a few bugs that required fixing up within Lock Sessions in the UI.

The first and most important was that the Lock Session APIs changed shape back in 1.7 but the documentation doesn't reflect this change (https://www.consul.io/api-docs/session#sample-response-3) so I've already added an issue for this here https://github.com/hashicorp/consul/issues/10095

The UI still expected the JSON response to be in the pre-1.7 shape so primarily this PR alters our mock API and the UI to the shape of the response as it is post-1.7.

Also:

Viewing a Session for a KV:

1. We noticed that the `Name` property is optional, so we only show this row if a Name is set for a session.
2. We noticed that this is the only place in the UI where we refer to a Node as an 'Agent' and is probably just a hangover from the original version of the UI, so we've changed this up here to use the word Node instead to match the rest of the UI.
3. There was a bug where we weren't showing the Delay or TTL here at all, so we fixed that here also.
4. We changed the order of the rows in the table to match the order of the information in our Lock Session listing page.

Lock Session Listing page/component:

1. We moved the CopyButton for the ID out of the `dt` and into the `dd` to make it visible/clickable.
2. Upgraded the tooltips to use a modifier style which we added in https://github.com/hashicorp/consul/pull/9288

### DIFF
--- a/.changelog/10225.txt
+++ b/.changelog/10225.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Reflect the change of Session API response shape for Checks in post 1.7 Consul 
+```
--- a/ui/packages/consul-ui/app/components/consul/lock-session/form/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/lock-session/form/index.hbs
@@ -16,30 +16,32 @@
           <a href="{{env 'CONSUL_DOCS_URL'}}/internals/sessions.html#session-design" rel="help noopener noreferrer" target="_blank">Lock Session</a>
         </h2>
         <dl>
+{{#if api.data.Name}}
           <dt>Name</dt>
           <dd>{{api.data.Name}}</dd>
-          <dt>Agent</dt>
+{{/if}}
+          <dt>ID</dt>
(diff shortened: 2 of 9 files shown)


kaxcode (approved):
👍

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/370670.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 39302041e9d65d2ab2c312d7583b9c54fc1ae83d onto `release/1.10.x` succeeded!

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 39302041e9d65d2ab2c312d7583b9c54fc1ae83d onto `release/1.9.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/370670)

PR #10228, opened 2021-05-11
Fix strings.Replace->strings.ReplaceAll

Haxandmat:
Function
`strings.Replace (version, "'", "", -1)`
replaced by
`strings.ReplaceAll (version, "'", "")`


markan (approved):
Looks reasonable to me.

PR #10229, opened 2021-05-11
mog: handle most type conversions automatically

rboyer:
This builds on #10176 but rewrites enough that I'm targeting the original mog base branch so you're not distracted by the interim solution.

These types are now handled:

1. basic (and pointers)
2. named structs (and pointers)
3. slices to (1) or (2)
4. maps with assignable keys and values of (1) or (2)
5. type aliases of the above

With these changes most of the `func-from` and `func-to` mog annotations in consul can be removed.


devtools/mog/testdata/TestE2E-expected-node_gen.go
```diff
@@ -2,19 +2,335 @@
```
rboyer, line 2: Here's the generated code.

devtools/mog/main.go
```diff
@@ -76,14 +76,14 @@ func runMog(opts options) error {
...
 	}
+
 	targets, err := loadTargetStructs(targetPackages(cfg.Structs))
 	if err != nil {
 		return fmt.Errorf("failed to load targets: %w", err)
 	}
 
 	cfg.Structs = applyAutoConvertFunctions(cfg.Structs)
 
-	log.Printf("Generating code for %d structs", len(cfg.Structs))
```
dnephin, line 86: I found this logging helpful. Could we restore this or does it not work anymore for some reason?

devtools/mog/main_e2e_test.go
```diff
@@ -1,13 +1,16 @@
...
 
 import (
+	"bufio"
+	"fmt"
 	"io/ioutil"
 	"os"
+	"strings"
 	"testing"
 
+	"gotest.tools/icmd"
```
dnephin, line 11:
```suggestion
	"gotest.tools/v3/icmd"
```

devtools/mog/main_e2e_test.go
```diff
@@ -33,5 +36,27 @@ func TestE2E(t *testing.T) {
 
 	actual, err := ioutil.ReadFile(output)
 	assert.NilError(t, err)
+
+	// t.Logf("OUTPUT\n%s\n", PrependLineNumbers(string(actual)))
```
dnephin, line 40: stray debug line? Can be removed I guess

dnephin (approved):
Nice! Thank you for working on this, it looks great!

I didn't look at everything super closely, but the generated code looks good, and I guess that's what really matters. Awesome to see even `type things []thing` are handled.

Mostly small nits, but I think the log line in `main.go` and the import of `gotest.tools/v3/icmd` should be restored.

devtools/mog/ast.go
```diff
@@ -0,0 +1,383 @@
...
+		panic("impossible")
+	}
+}
+
+func newAssignStmtSlice(
+	left ast.Expr,
+	leftType ast.Expr,
+	leftElemType ast.Expr,
+	right ast.Expr,
+	rightType ast.Expr, // UNUSED
```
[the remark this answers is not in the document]
rboyer: Every time I flipped the generation to reverse the mog flow it was helpful to be able to just flip the first and second half of the args (even if they weren't used). I'll go through and clean those up now that it's no longer necessary.

devtools/mog/main_e2e_test.go
rboyer: Restored

devtools/mog/main.go
rboyer: Restored

devtools/mog/main_e2e_test.go
rboyer: Instead I just made it more formal: `-print-gen=true`. When I was debugging `go vet` crashes in the E2E test I found the line numbers useless, so I was commenting out the vet lines and toggling this line to print the generated code with line numbers so i could understand the vet failure that had just happened.

Now if it fails you could rerun with `-vet-gen=false -print-gen=true`

devtools/mog/ast.go
```diff
@@ -0,0 +1,383 @@
...
+		// Pointer to Pointer
+		//
+		// <left> = <right>
+		return astAssign(left, right)
+	default:
+		panic("impossible")
+	}
+}
+
+func printTypeExpr(expr ast.Expr) string {
```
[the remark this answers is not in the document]
rboyer: It was used several times during mog debugging so I've added a comment about that use.

I remember trying `ast.Fprint` and IIRC abandoned it for this at some point but I don't remember specifically what was not great about it.

devtools/mog/generate.go
```diff
@@ -124,39 +129,155 @@ func generateConversion(cfg structConfig, t targetStruct, imports *imports) (gen
...
-	to.Body.List = append(to.Body.List, returnStmt)
+		mapErrFn := func(err error) {
+			if err == nil {
+				errs = append(errs, fmt.Errorf(
+					"struct %v field %v is not mappable to target",
+					cfg.Source, name,
+				))
+			} else {
+				errs = append(errs, fmt.Errorf(
+					"struct %v field %v is not mappable to target: %w",
```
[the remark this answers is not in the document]
rboyer: Updated terms.

dhiaayachi:
This now live in it's own repo: https://github.com/hashicorp/mog

PR #10230, opened 2021-05-12
docs: updated the standard upgrade process

karl-cardenas-coding:
Added a cross-reference link in the upgrade guides.  This resource https://www.consul.io/docs/upgrading/instructions/general-process including specific-version guides for breaking changes and a more detailed upgrade process, but it's not mentioned in the  https://www.consul.io/docs/upgrading#standard-upgrade overview page.


website/content/docs/upgrading/index.mdx
```diff
@@ -25,6 +25,7 @@ For upgrades we strive to ensure backwards compatibility. To support this,
 nodes gossip their protocol version and builds. This enables clients and
 servers to intelligently enable new features when available, or to gracefully
 fallback to a backward compatible mode of operation otherwise.
+Visit the [General Upgrade Process](https://www.consul.io/docs/upgrading/instructions/general-process) for a detailed upgrade guide.
```
banks, line 28: This link is a good addition. Do you think the instructions below are still valuable given the duplication with the much more detailed ones on that page?

I could see us modifying one place and not the other for example or folks not benefitting from the more detailed ones because they assume this is the place to come for the reference steps?

FYI, we've discussed recently how we only cover one type of upgrade (in place rolling update) in both of these locations and we support at least three different types (also "immutable" rolling upgrade and full blue-green with Enterprise Autoupgrade) but I think that's a separate issue to address. It would be messy though to end up fixing that in two different places!

karl-cardenas-coding: The way I see it, we can either remove them or keep them there but call it something like `Quick start` or a better indication of a brief instruction set. As for the later points, I agree that cleaner restructuring is needed and that's something I am hoping the tech writer can chime in on.

banks: That plan sounds good to me. I'll approve this as is as I think it's an improvement on it's own and those other suggestions can be follow up work.

Thanks Karl!

banks (approved):

website/content/docs/upgrading/index.mdx
karl-cardenas-coding: @banks Any concerns with the integration test failing? The `docker-upload` CI steps is marked as unauthorized.

PR #9675, opened 2021-02-01
ui: Add 'Scenario' debug function for easy saving debug scenarios

johncowen:
Following on from https://github.com/hashicorp/consul/pull/9666, this PR adds a `Scenario()` debug function to allow you to easily save off 'scenarios' for easy debugging or demoing purposes.

When called with no arguments (using WebInspector's `console`, or via a bookmarklet), the function will open a new tab with an easily copy/pastable URL (for sharing preview sites) and a draggable bookmarklet link containing the configuration for your current development/debug configuration settings (which we can set via the cookies panel in the WebInspector). Whilst this function only exists in non-production environments we also only emit cookies that begin with `CONSUL_` which we know are our own debug/dev cookies.

Scenarios are composable, so instead of having lots of multiple bookmarklets, you can save say one for enabling nspace/acls another for metrics - clicking one after the other will give you nspaces/acls plus metrics.

We haven't added this to the README as yet , we'll wait until https://github.com/hashicorp/consul/pull/9666 has been merged and also possibly decide a more central place for all these types of functions to live.

Directions below:

```
data:text/html,<a href="javascript:Scenario()">Save Current Scenario</a>
```

scenario

PR #9666, opened 2021-01-29
ui: Adds a `Routes()` router DSL dumping function only in dev mode

johncowen:
Back in https://github.com/hashicorp/consul/pull/5206 we fixed a URL encoding/decoding issues with URLs for wildcarded routes by extracting our routing configuration to a JSON-like JS pojo, before feeding this in Embers `Route.map`. This meant we could inspect the routing configuration from elsewhere to provide support for URL encoding/decoding. The one downside in doing this was that we were not using Embers Route DSL to code our Routing configuration.

At the time we wrote a little `dump` util in order to easily convert the JSON-like configuration back to Ember's Router DSL should we ever need to, but commented it out as we didn't want it ending up in the production source code.

We've since discovered `runInDebug`, which lets you specify inline code to be stripped from production builds.

This PR uses `runInDebug` to wrap the a utility function called `Routes()`, which will print out Embers Route DSL in a new tab should we need to. Passing a URL with a `%s` marker will pass the Route DSL to the URL, for example `http://url.com?routes=%s`, this lets you easily use tools like `ember-diagonal`. Additionally we've added a `DEBUG_ROUTES_ENDPOINT` development environment variable (set via cookies), so you can set this URL semi-permanently if you always want the DSL passed to a URL instead of opening in a new tab. We named it `Routes()` for now and hung it off of `window` for easy usage - we shouldn't have any problems with clashes and its only a development function so we feel this is ok.

To call this `Routes()` function, you can either type `Routes()` into your WebInspector console and hit enter, or make some bookmarklets if you find yourself wanting to do this often - see below for directions:

```
data:text/html,<a href="javascript:Routes()">Print Routing DSL</a><br /><a href="javascript:Routes('https://alexspeller.com/ember-diagonal/?routesInput=%s')">Show Routing DSL in Ember Diagonal</a>
```

bookmarklets

Lastly, there are probably a few other debug/dev functions I sometimes use that I could add in also, so I've added a new section in the README file to document this one and any future debug/dev functions.

PR #10237, opened 2021-05-13
ui: Use babel plugin to strip calls to runInDebug

johncowen:
A while ago we discovered `runInDebug` imported from `@ember/debug` and started using it to be able to add engineering utilities to the UI to help other engineers without compiling the required debug code into the final production build.

For example:

https://github.com/hashicorp/consul/pull/8658
https://github.com/hashicorp/consul/pull/9675
https://github.com/hashicorp/consul/pull/9666
https://github.com/hashicorp/consul/pull/9831
https://github.com/hashicorp/consul/pull/9771

During https://github.com/hashicorp/consul/pull/9831 we found out there was a change at some point in the ember docs that changed the documented behaviour of `runInDebug` from being 'stripped' to being a 'noop' instead.

> We also noticed whilst we were doing this that although the older ember documentation for runInDebug states `Uses of this method in Ember itself are stripped from the ember.prod.js build.`, but that is not actually the case 😬 (or at least not in later versions).

Here we use an additional babel plugin to strip any calls to `runInDebug` from the final production source code.

P.S. Be aware that GitHubs diff here is a little confusing by default (unless you expand the code), the stripping is only performed during ~`test` and~ `production`.

As we can't strip this for test runs (I'm assuming runInDebug is used deep down for test setup or something), that means anything that is affected by the removal/stripping of `runInDebug` is never tested fully.

Worth noting that as `runInDebug` is noop-ed in default ember usage, so even without this addition, anything that `runInDebug` affects is never tested fully either.

The only addition here is to strip it instead of turning it into a noop so we don't ship dead code that is never run.

### DIFF
--- a/ui/packages/consul-ui/ember-cli-build.js
+++ b/ui/packages/consul-ui/ember-cli-build.js
@@ -8,13 +8,19 @@ module.exports = function(defaults) {
 
   const env = EmberApp.env();
   const prodlike = ['production', 'staging'];
+  const sourcemaps = !['production'].includes(env);
 
   const trees = {};
   const addons = {};
   const outputPaths = {};
   let excludeFiles = [];
 
-  const sourcemaps = !['production'].includes(env);
+  const babel = {
+    plugins: [
+      '@babel/plugin-proposal-object-rest-spread',
+    ],
+    sourceMaps: sourcemaps ? 'inline' : false,
+  }
 
   // setup up different build configuration depending on environment
   if(!['test'].includes(env)) {
@@ -38,6 +44,9 @@ module.exports = function(defaults) {
       // exclude docfy
       '@docfy/ember'
     ];
+    babel.plugins.push(
+      ['strip-function-call', {'strip': ['Ember.runInDebug']}]
+    )
   } else {
     // add debug css is we are not in test or production environments
     outputPaths.app = {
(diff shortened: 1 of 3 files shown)


ui/packages/consul-ui/ember-cli-build.js
```diff
@@ -46,6 +52,13 @@ module.exports = function(defaults) {
       }
     }
   }
+  if(['production'].includes(env)) {
+    // everything apart from production is 'debug', including test
+    // which means this and everything it affects is never tested
+    babel.plugins.push(
+      ['strip-function-call', {'strip': ['Ember.runInDebug']}]
```
kaxcode, line 59: awesome 🎉

kaxcode (approved):
LGTM

ui/packages/consul-ui/ember-cli-build.js
johncowen: Worth noting that as `runInDebug` is noop-ed in default ember usage, so even without this addition, anything that `runInDebug` affects is never tested either.

The only addition here is to strip it instead of turning it into a noop so we don't ship dead code that is never run.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/369591.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit d1bbe0895fd7c5a61d15545057cfb64798839dc0 onto `release/1.10.x` succeeded!

PR #10239, opened 2021-05-13
server: ensure that central service config flattening properly resets the state each time

rboyer:
The prior solution to call reply.Reset() aged poorly since newer fields
were added to the reply, but not added to Reset() leading serial
blocking query loops on the server to blend replies.

This could manifest as a service-defaults protocol change from
default=>http not reverting back to default after the config entry
reponsible was deleted.

PR #10245, opened 2021-05-14
[1.9.x] server: ensure that central service config flattening properly resets the state each time

rboyer:
The prior solution to call reply.Reset() aged poorly since newer fields
were added to the reply, but not added to Reset() leading serial
blocking query loops on the server to blend replies.

This could manifest as a service-defaults protocol change from
default=>http not reverting back to default after the config entry
reponsible was deleted.

Backport of #10239 to 1.9.x

### DIFF
--- a/.changelog/10239.txt
+++ b/.changelog/10239.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+server: ensure that central service config flattening properly resets the state each time
+```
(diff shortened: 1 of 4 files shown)


markan (approved):
LGTM

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 3357a14339a6b74ddb0b87b506c75cd4af0a1097 onto `release/1.8.x` succeeded!

PR #10253, opened 2021-05-18
redirecttraffic: fix a flaky test

dnephin:
I noticed this test failure: https://app.circleci.com/pipelines/github/hashicorp/consul-enterprise/6408/workflows/e90b1140-daa2-458f-8197-d1821e3693e3/jobs/94694/tests#failed-test-0

Apparently the `ExcludeInboundPorts` slice can be in a different order sometimes, although I'm not sure how. Is it safe to ignore the order of these ports?

Also removes t.Parallel. We should only need to use t.Parallel in large packages with many slow tests. In this case there is only a single slow test so we don't get any benefit.

Also add struct field names to the testcases list, so that it is easier to see what each value is doing, and to make it easier to add new fields in the future. I learned GoLand has a "add keys to struct literal" helper in the context menu that made adding these easy.

### DIFF
--- a/command/connect/redirecttraffic/redirect_traffic_test.go
+++ b/command/connect/redirecttraffic/redirect_traffic_test.go
@@ -1,18 +1,18 @@
 package redirecttraffic
 
 import (
+	"sort"
 	"testing"
+
+	"github.com/mitchellh/cli"
+	"github.com/stretchr/testify/require"
 
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/consul/sdk/iptables"
 	"github.com/hashicorp/consul/sdk/testutil"
-	"github.com/mitchellh/cli"
-	"github.com/stretchr/testify/require"
 )
 
 func TestRun_FlagValidation(t *testing.T) {
-	t.Parallel()
-
 	cases := []struct {
 		name     string
 		args     []string
@@ -59,8 +59,6 @@ func TestRun_FlagValidation(t *testing.T) {
 }
 
 func TestGenerateConfigFromFlags(t *testing.T) {
-	t.Parallel()
-
 	cases := []struct {
 		name           string
 		command        func() cmd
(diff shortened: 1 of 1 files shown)


markan (commented):
The test change looks reasonable. My only hesitation is that for the life of me I can't spot where the non-determinism in the ordering is coming from in the first place; I've read the code a few times and nothing leaps out.

dnephin:
> My only hesitation is that for the life of me I can't spot where the non-determinism in the ordering is coming from in the first place; I've read the code a few times and nothing leaps out.

Same! It must be there somewhere, but I don't see the values going through any maps. I wondered if maybe there was an async operation that happened to write in a different order, but I couldn't find one.

freddygv (approved):
LGTM, thanks for fixing!

FYI this ChecksWithFilter call returns a map, and then we range over it to collect the ports: 
https://github.com/hashicorp/consul/blob/master/command/connect/redirecttraffic/redirect_traffic.go#L204

ishustava (approved):
Edit: @freddygv jinx 😄 

---
Looks great, thanks so much for making these changes!

I think the non-deterministic part comes from the fact that `ChecksWithFilter` called [here](https://github.com/hashicorp/consul/blob/a003265964f5b1ef6195ecbab1fb21d826bb597b/command/connect/redirecttraffic/redirect_traffic.go#L204) returns a map, and as a result the ports that are added to the array could be added in different orders.

dnephin:
Aha, thank you both!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/370813.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit dc73939861ae42e244579c2ddd848d0fdebdf8f4 onto `release/1.10.x` succeeded!

PR #9587, opened 2021-01-19
ui: Adds @NullValue attr decorator

johncowen:
There are many places in the API where we receive a property set to
`null` which can then lead to defensive code deeper in the app in order
to guard for this type of thing when usually we are expecting an array
or for the property to be undefined using omitempty on the backend.

Previously we had two places where we would deal with this in the
serializer using our 'remove-null' util (KV and Intentions).

This new decorator lets you declaritively define this type of data using
a decorator @NullValue([]) (which would replce a null value with [].

@NullValue in turn uses a more generic @replace helper, which we
currently don't need but would let you replace any value with another,
not just a null value.

An additional benefit here is that the guard/replacement is executed
lazily when we get the property instead of serializing all the values
when they come in via the API. On super large datasets, where we only
visualize part of the dataset (say in our scroll panes), this feels like
a good improvement on the previous approach.

Whilst we've not looked at the issue yet, this PR could help us to fix https://github.com/hashicorp/consul/issues/9535. Once we've looked at that we can decide whether to backport this to 1.9.x or not.

PR #10257, opened 2021-05-19
[Backport/1.9.x] Backport #10225

johncowen:
Manual backport of #10225

### DIFF
--- a/.changelog/10225.txt
+++ b/.changelog/10225.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Reflect the change of Session API response shape for Checks in post 1.7 Consul 
+```
--- a/ui/packages/consul-ui/app/components/consul/lock-session/form/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/lock-session/form/index.hbs
@@ -16,30 +16,32 @@
           <a href="{{env 'CONSUL_DOCS_URL'}}/internals/sessions.html#session-design" rel="help noopener noreferrer" target="_blank">Lock Session</a>
         </h2>
         <dl>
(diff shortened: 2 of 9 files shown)


kaxcode (approved):
:shipit:

johncowen:
@kaxcode found out a couple of interesting things here:

1. Luckily I noticed that CI tests didn't run here, I'm guessing it might have been because the branch isn't a `ui/*` branch. So in future I'm gonna try to remember to use `ui/backport/x.x.x/original-pr-number` for a branch name for these so we still get an automated tests. It's probably even more important to make sure the automated tests run for backports than it is for main branch things as its less likely for us to spot anything post merge.
2. As I'd noticed that the tests didn't run, I ran them locally before merging - and the tests failed! I looked into it and found that https://github.com/hashicorp/consul/pull/9587 didn't go into `1.9.x` which we used in the original 1.10 PR for this, so I removed this usage here, and checked that `union` doesn't require empty arrays in order to work (passing `undefined` as an argument is fine). Once I'd done that all tests pass locally so I'm gonna merge this now.

@alvin-huang a just incase FYI here incase you have any other suggestions re: point 1 ^

alvin-huang:
There isn't tooling in there to do that right now. We mostly assume that the backport will apply cleanly and work, but the tests should run once it's actually merged into the release branch and we could fix it then.

The right approach would be to create a backport branch of the backport commit(s) applied on the release branch and run tests on that, and only backport the commit if all those tests pass. This is similar to what we do for the oss-merge. 

The branch being called `ui/*` will help PRs like this where you have to create a manual backport but the automation described ^ would help for auto backports based on PR labels.

PR #10262, opened 2021-05-19
remove domain search from hostname when using it as node_name

dhiaayachi:
This is to fix the following issue:

when the machine running consul is part of subdomain the name returned by `os.Hostname()` include the subdomain which usually have `.`  (at least on MacOS)
Consul use `os.Hostname()` as default `node_name` value (if none provided) but do not accept `.` as part of the `node_name`
This will lead to consul by default not discoverable via DNS when starting in a machine that is part of sub domain and giving the following warning:

`2021-05-19T16:46:05.374-0400 [WARN]  agent: Node name "xxxxxxxxxxx.subdomain" will not be discoverable via DNS due to invalid characters. Valid characters include all alpha-numerics and dashes.`

The fix in this PR will take only that part before the first as node_name value, for the given example this will be `xxxxxxxxxxx`

### DIFF
--- a/command/services/config_test.go
+++ b/command/services/config_test.go
@@ -1,8 +1,6 @@
 package services
 
 import (
-	"os"
-	"strings"
 	"testing"
 	"time"
 
@@ -17,18 +15,14 @@ import (
 // We depend on this behavior for ServiesFromFiles so we want to fail
 // tests if that ever changes.
 func TestDevModeHasNoServices(t *testing.T) {
-	hostname, err := os.Hostname()
-	require.NoError(t, err)
 	devMode := true
-	opts := config.LoadOpts{DevMode: &devMode}
+	opts := config.LoadOpts{
+		DevMode: &devMode,
+		HCL:     []string{`node_name = "dummy"`},
+	}
 	result, err := config.Load(opts)
 	require.NoError(t, err)
-	if !strings.Contains(hostname, ".") {
-		require.Len(t, result.Warnings, 0)
-	} else {
-		require.Len(t, result.Warnings, 1)
-		require.Contains(t, result.Warnings[0], "will not be discoverable via DNS due to invalid characters. Valid characters include all alpha-numerics and dashes.")
-	}
+	require.Len(t, result.Warnings, 0)
 	require.Len(t, result.RuntimeConfig.Services, 0)
 }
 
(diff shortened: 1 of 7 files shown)


dnephin (commented):
@dhiaayachi and I chatted about this. We think changing the default could be a problem, because the node name gets used in ACL policies, in metric names, and the unique ID in serf. So a change in the default could cause problems on upgrades.

We think maybe we could apply this change only to the tests that fail, to allow the test suite to be run when the Hostname contains dots.

dhiaayachi:
@dnephin I reverted the initial change and implemented the changes to the tests as discussed, please review when you have some spare time.

command/services/config_test.go
```diff
@@ -15,10 +17,18 @@ import (
...
 func TestDevModeHasNoServices(t *testing.T) {
+	hostname, err := os.Hostname()
+	require.NoError(t, err)
 	devMode := true
-	result, err := config.Load(config.LoadOpts{DevMode: &devMode})
+	opts := config.LoadOpts{DevMode: &devMode}
+	result, err := config.Load(opts)
 	require.NoError(t, err)
-	require.Len(t, result.Warnings, 0)
+	if !strings.Contains(hostname, ".") {
```
dnephin, line 26: Generally I try to avoid any conditional assertions in tests, especially when they are for handling differences in the test environment.

I think you could handle it this way instead:

```go
opts := config.LoadOpts{
    DevMode: &devMode
    HCL: []string{`node_name = "dummy"`},
}

That should do about the same thing as the other tests.
```

dnephin (commented):
Thanks! I think this is a good thing to fix. One suggestion for one of the test changes

command/services/config_test.go
dhiaayachi: Should I add a test to validate the current behaviour when a node name is invalid and that we get the right warning?

dhiaayachi: fixed the test and added an extra test for invalid `node_name` use case

dnephin: I think we should already have test coverage for the warnings.  I wouldn't worry about adding a new one since this PR doesn't change that behaviour.

dnephin: Ah, I see you added one, that looks good too!

dnephin (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/378582.

PR #10267, opened 2021-05-20
Move some things around to allow for license updating via config reload

mkeeler:
The bulk of this commit is moving the LeaderRoutineManager from the agent/consul package into its own package: lib/gort. It also got a renaming and its Start method now requires a context. Requiring that context required updating a whole bunch of other places in the code.

### DIFF
--- a/agent/consul/leader.go
+++ b/agent/consul/leader.go
@@ -115,7 +115,7 @@ func (s *Server) monitorLeadership() {
 
 				if canUpgrade := s.canUpgradeToNewACLs(weAreLeaderCh != nil); canUpgrade {
 					if weAreLeaderCh != nil {
-						if err := s.initializeACLs(true); err != nil {
+						if err := s.initializeACLs(&lib.StopChannelContext{StopCh: weAreLeaderCh}, true); err != nil {
 							s.logger.Error("error transitioning to using new ACLs", "error", err)
 							continue
 						}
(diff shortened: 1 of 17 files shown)


agent/consul/leader.go
```diff
@@ -979,7 +979,7 @@ func (s *Server) startConfigReplication() {
 		return
 	}
 
-	s.leaderRoutineManager.Start(configReplicationRoutineName, s.configReplicator.Run)
+	s.leaderRoutineManager.Start(context.Background(), configReplicationRoutineName, s.configReplicator.Run)
```
freddygv, line 982: How come these don't use the context from parent functions like `establishLeadership`? Should these be `context.TODO()`?

freddygv (commented):
This generally looks good to me, have a couple questions.

Did you consider other names for the gort package? This abbreviation seems a bit unintuitive. What if it was called `routine`?

agent/consul/leader.go
mkeeler: I didn't look far enou

> Did you consider other names for the gort package? This abbreviation seems a bit unintuitive. What if it was called `routine`?

I started from goroutine and just shortened it. I would be happy to rename it to `routine`.

mkeeler:
> Did you consider other names for the gort package? This abbreviation seems a bit unintuitive. What if it was called `routine`?

I started from goroutine and just shortened it. I would be happy to rename it to `routine`.

markan (approved):
This generally looks good; I will note that 'gort' as a name threw me as well.

lib/routine/gort.go
```diff
@@ -0,0 +1,148 @@
...
+
+	m.logger.Debug("stopping routine", "routine", name)
+	instance.cancel()
+
+	delete(m.routines, name)
+
+	return instance
+}
+
+func (m *Manager) StopAll() {
```
freddygv, line 134: Can you add a test for this function?

freddygv (commented):
There's some stray references to gort remaining like the dir `/lib/gort`, also the filenames under `lib/routine`.

I also left one last request in line. 

Should be all all set after that.

lib/routine/gort.go
mkeeler: Added.

freddygv (approved):
:shipit:

banks (approved):
Nice LGTM too! 👍 on `routine` 😄

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/374007.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit da31e0449ea6323956f8ebc942a6d76332c52e27 onto `release/1.10.x` succeeded!

PR #10270, opened 2021-05-20
ui: Create and use collapsible notices component

kaxcode:
collapsible_banner

- Creates `CollapsibleNotices` component
- Restructure the acceptance tests folder for topology
- Adds testing for TProxy notices in Topology tab
- Use of `CollapsibleNotices` in Toplogy route when there are more than 2 `Notices`.

### DIFF
--- a/ui/packages/consul-ui/app/components/notice/layout.scss
+++ b/ui/packages/consul-ui/app/components/notice/layout.scss
@@ -22,3 +22,6 @@
   left: 0.6rem;
   font-size: 1rem;
 }
+%notice p.hide {
+  display: none;
+}
--- a/ui/packages/consul-ui/app/components/notice/layout.scss
+++ b/ui/packages/consul-ui/app/components/notice/layout.scss
@@ -22,6 +22,3 @@
   left: 0.6rem;
   font-size: 1rem;
 }
-%notice p.hide {
-  display: none;
-}
--- a/ui/packages/consul-ui/tests/acceptance/steps/dc/acls/auth-methods/index-steps.js
+++ b/ui/packages/consul-ui/tests/acceptance/steps/dc/acls/auth-methods/index-steps.js
@@ -1,4 +1,4 @@
-import steps from '../../../steps';
+import steps from '../../../../steps';
 
 // step definitions that are shared between features should be moved to the
 // tests/acceptance/steps/steps.js file


ui/packages/consul-ui/app/templates/dc/services/show/topology.hbs
```diff
@@ -22,43 +22,93 @@ as |route|>
         </BlockSlot>
       </EmptyState>
     {{else}}
+
+
+    {{#if (or (and topology.DefaultAllow topology.FilteredByACLs topology.undefinedIntention) (and topology.WildcardIntention topology.FilteredByACLs topology.undefinedIntention))  }}
```
kaxcode, line 27: 🙈 The conditional works but I had to duplicate the code in the `{{else}}`. LMK if you can think of a better way to format this.

johncowen: Theres an option here to maybe make a computedProperty on the topology model

ui/packages/consul-ui/app/models/topology.js
```diff
@@ -19,23 +19,27 @@ export default class Topology extends Model {
   @attr() Downstreams; // Service[],
   @attr() meta; // {}
 
-  @computed('Upstreams', 'Downstreams')
+  @computed('Downstreams')
```
kaxcode, line 22: When I did this computed function I forgot undefined intentions only exist for Downstream services.

johncowen (commented):
Sorry I had this sitting here this morning and I hadn't submitted the review, just realized you pushed the computed version! Sorry!

Anyway, read through and see what you think.

ui/packages/consul-ui/app/templates/dc/services/show/topology.hbs
kaxcode: The 2nd option does not take into account that I have to have >2 notices to show that button. That's what I'm trying to capture in the conditionals I wrote. Grouping them all together in the `or` does not guarantee that use case.

johncowen: Ahhh, of course yeah I forgot about that 🤔

johncowen: Wonder if you could do something with `compact` or `array_some` 🤔

johncowen: Maybe this?

```hbs
      {{#let (without false (array true true false false)) as |visibleNotices|}}
        {{if gt visibleNotices.length 2}}
          {{!-- make collapsable --}}
        {{/if}}
      {{/let}}

```

kaxcode: I've already moved the logic to a computed func in the model. Does that work? 
I also just pushed renaming `undefinedIntention` to be `notDefinedIntention`.

ui/packages/consul-ui/tests/pages/dc/services/show.js
```diff
@@ -26,6 +35,20 @@ export default function(visitable, clickable, attribute, collection, text, inten
     }),
     intentionList: intentions(),
   };
+  page.tabs.topologyTab = {
+    defaultAllowNotice: {
+      see: isPresent('[data-test-notice="default-allow"]'),
+    },
+    filteredByACLs: {
+      see: isPresent('[data-test-notice="filtered-by-ACLs"]'),
```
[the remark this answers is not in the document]
kaxcode: Whoops! I committed this change but it broke tests because the change needed to be made in the hbs file as well. I've committed this changes in a separate commit.

johncowen (approved):
LGTM! Especially, thanks for adding the docs for the new component!

There are a couple of things here we could potentially circle back and tweak, but fine to do that in a PR in the future at some point.

PR #9309, opened 2020-12-01
docs: Add example ACL policy for snapshot agent

blake:
Add example least privilege ACL policy for Consul snapshot agent.

PR #10271, opened 2021-05-20
docs: Add example ACL policy for snapshot agent

dhiaayachi:
This is to merge changes from https://github.com/hashicorp/consul/pull/9309

Add example least privilege ACL policy for Consul snapshot agent.

Co-Authored-By: Blake Covarrubias <blake@covarrubi.as>


blake:
@dhiaayachi Are these the same changes as in the original PR? If so, I'm happy to just rebase #9309 so that its able to be merged.

dhiaayachi:
@blake I had to change the alignment a bit because it was not rendering correctly, but yes the content is the same

dnephin:
Merged in #9309

PR #10273, opened 2021-05-20
debug: remove the CLI check for debug_enabled

### LINKED ISSUE #10208: cli: consul debug should not require `enable_debug` for profiling
Currently `consul debug` will only collect profiling data if the server it speaks to has `enable_debug = true` in it's configuration.

Ironically in the [same PR that added the command]() we actually agreed to change the behaviour to make pprof endpoints enabled automatically when ACLs are enabled since most user do run in production securely with ACLs.

So most users should be able to use `consul debug` to gather profiling data since the pprof endpoints are actually exposed provided they have a operator ACL token, however the command will refuse to unless `enable_debug` is true requiring a rolling restart of servers many cases to get profiling data to help debug an issue.

We should change the command to:
 - check if ACLs are enabled and if they are, attempt to fetch profile data anyway (provided a token is available via ENV/CLI args).
 - if ACLs are enabled but the profile endpoints return a 403, output a more accurate error that explains that the user may simply provide an operator ACL token rather than needing to restart servers with debug enabled.

dhiaayachi:
Fixes #10208

The API allows collecting profiles even when debug_enabled=false, as long as ACLs are enabled. This PR remove the check for `debug_enabled` from the CLI so that users do not need to set debug_enabled=true for no reason.

Also fixes the API client to return errors on non-200 status codes for debug endpoints and improves the failure messages when pprof data can not be collected.

The fix to the API client exposed the fact that most of these tests were actually failing silently. They were never checking the actual output, and all of these tests were hitting 500 errors on the server (see comment below for why). So this PR also fixes the tests.

### DIFF
--- a/.changelog/10273.txt
+++ b/.changelog/10273.txt
@@ -1,0 +1,5 @@
+```release-note:bug
+cli: removes the need to set debug_enabled=true to collect debug data from the CLI. Now
+the CLI behaves the same way as the API and accepts either an ACL token with operator:read, or
+debug_enabled=true.
+```
--- a/agent/http.go
+++ b/agent/http.go
@@ -238,18 +238,16 @@ func (s *HTTPHandlers) handler(enableDebug bool) http.Handler {
 			var token string
 			s.parseToken(req, &token)
 
+			// If enableDebug is not set, and ACLs are disabled, write
+			// an unauthorized response
+			if !enableDebug && s.checkACLDisabled(resp, req) {
+				return
+			}
+
 			rule, err := s.agent.delegate.ResolveTokenAndDefaultMeta(token, nil, nil)
 			if err != nil {
 				resp.WriteHeader(http.StatusForbidden)
 				return
-			}
-
-			// If enableDebug is not set, and ACLs are disabled, write
-			// an unauthorized response
-			if !enableDebug {
-				if s.checkACLDisabled(resp, req) {
-					return
-				}
 			}
 
 			// If the token provided does not have the necessary permissions,
--- a/api/debug.go
+++ b/api/debug.go
@@ -29,6 +29,10 @@ func (d *Debug) Heap() ([]byte, error) {
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode != 200 {
+		return nil, generateUnexpectedResponseCodeError(resp)
+	}
+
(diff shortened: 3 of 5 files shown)


dhiaayachi:
@dnephin I think this is now ready.
The reason tests were failing is that this kind of tests cannot run in parallel is that this use (among others) the `pprof` go package to collect debug data and this can be activated only once within a runtime. In case of parallel testing it gets activated multiple times within the same runtime (the test runtime). the straight forward solution is to not run tests in parallel in that case.

dnephin:
Ah yes, makes sense! In general I have been removing `t.Parlallel` as much as possible. I think we should only need it for tests in the `agent` and `agent/consul` package because there are   a large number of slow tests in those packages. Every other package should not need to use `t.Parallel`.

I'll add a changelog entry and then I think we should be ready for someone to review.

kyhavlov (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/376360.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/377975.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 4c7f5f31c718ea1ee00ef93545177c951376b51f onto `release/1.10.x` succeeded!

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 4c7f5f31c718ea1ee00ef93545177c951376b51f onto `release/1.9.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/377975)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 4c7f5f31c718ea1ee00ef93545177c951376b51f onto `release/1.8.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/377975)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 4c7f5f31c718ea1ee00ef93545177c951376b51f onto `release/1.7.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/377975)

PR #10279, opened 2021-05-21
generate a single debug file for a long duration capture

dhiaayachi:
Current behaviour when generating dynamic debug files is to create multiple files in subfolders that represent each capture interval that make it harder to correlate logs and debug

This PR is to: 
- change the behaviour to create a single file for each captured item and append all the data under that file, the single file will still be under a subfolder with the capture starting timestamp (need to be discussed)
- simplify the capture loop by using context to handle timeouts


dnephin (commented):
Nice, thank you for working on this!

This is a bit different then what I was imagining. I'm actually not sure if these file formats are safe for concatenation. Maybe they are?

I think this might work, but  collecting a single profile might also help us miss fewer things. I was thinking that we could split up `captureDynamic` and make a single API call for each of the profiles.

Looking at https://golang.org/pkg/net/http/pprof/ , we can specify the duration using the `seconds` parameter. That way we can specify `seconds=120` (or whatever duration is set using the command line flag).

I guess that will also require changes to the HTTP client in `api/debug`, because right now it looks like it is not accepting any parameters.

dhiaayachi:
Thank you @dnephin, 
I was not sure if it's the right way of doing it neither. I will  read more about the pprof api and attempt to do the change as you described it

dhiaayachi:
I went through the code to understand what is the current behaviour and my findings are the following:
-  3 different type of data in the `captureDynamic`  func are captured (logs, metrics, pprof)
- the capture is done for a time span (config parameter `interval`)
- we create a new file for each `interval` capture time and we do that for the `duration` parameter
- the `duration` parameter have a grace period of 2 seconds added to it

So if we do a capture with duration=200ms and an interval=100ms we will end up with 22 folders ((2s + 200ms)/100ms) each having a file with the data for each type of capture we specified (logs, metrics, pprof)

To achieve what is stated in this PR we need to change the logic because capturing everything in the same file defy the logic of having a duration and an interval.

one option would be to simplify the usage and have interval = duration if not specified so we capture a single set of files for the total duration specified (including the grace period)

@dnephin I will need your insights on this one as I'm not sure what we want

website/content/commands/debug.mdx
```diff
@@ -71,14 +71,14 @@ all targets for 2 minutes.
...
+| `cluster` | A list of all the WAN and LAN members in the cluster.                                                                                                                                                                                                                                 |
+| `metrics` | Metrics from the in-memory metrics endpoint in the target, captured at the interval.                                                                                                                                                                                                  |
+| `logs`    | `DEBUG` level logs for the target agent, captured for the interval.                                                                                                                                                                                                                   |
+| `pprof`   | Golang heap, CPU, goroutine, and trace profiling captured for `-duration` time. This information is not retrieved unless [`enable_debug`](/docs/agent/options#enable_debug) is set to `true` on the target agent or ACL are enable and an ACL token with `operator:read` is provided. |
```
dhiaayachi, line 81: I'm not sure if this is still accrurate

command/debug/debug.go
```diff
@@ -320,212 +323,218 @@ func (c *cmd) captureStatic() error {
...
+		case <-c.shutdownCh:
+			return errors.New("stopping collection due to shutdown signal")
 		}
+	}
+}
 
-		// Capture metrics
-		if c.configuredTarget("metrics") {
-			wg.Add(1)
+func capture(c *cmd) error {
```
freddygv, line 371: I think this could be renamed to something like captureShortLived or captureSnapshots, to contrast with captureLongRunning. What do you think?

freddygv (commented):
I think you're making good progress here. I only took a quick look but added some suggestions.

One question I have is: Do we intend to have some data be collected in intervals and some not? I think I'm seeing that logs, traces and profiles are long running while everything else is interval driven. 

I can see why that could be the case though. Goroutine dumps, stats gauges, and heap dumps could be more useful to see as multiple point-in-time intervals. Whereas CPU profiles and traces could be seen over the whole duration. 

Aside: looking at what we collect, I think we should consider adding `debug/pprof/allocs`. This endpoint samples all past allocations unlike heap which only shows live objects. (But that should be handled in a separate PR, I'll reply to the original issue with my reasoning)

dhiaayachi:
> I think you're making good progress here. I only took a quick look but added some suggestions.
> 
> One question I have is: Do we intend to have some data be collected in intervals and some not? I think I'm seeing that logs, traces and profiles are long running while everything else is interval driven.
> 
> I can see why that could be the case though. Goroutine dumps, stats gauges, and heap dumps could be more useful to see as multiple point-in-time intervals. Whereas CPU profiles and traces could be seen over the whole duration.
> 
> ~Aside: looking at what we collect, I think we should consider adding `debug/pprof/allocs`. This endpoint samples all past allocations unlike heap which only shows live objects. (But that should be handled in a separate PR, I'll reply to the original issue with my reasoning)~ Actually never mind that, apparently the difference between `heap` vs `allocs` profiles is the default view in the go tool. The data seems to be the same: https://golang.org/pkg/runtime/pprof/#Profile

Thank you for the review @freddygv 
Yes, that's the intend and the reason is exactly what you guessed.

command/debug/debug.go
dhiaayachi: I ran out of func names starting with capture for that point.
I like `captureShortLived`, I will make the change.

command/debug/debug.go
```diff
@@ -320,212 +323,218 @@ func (c *cmd) captureStatic() error {
...
 }
 
-// captureDynamic blocks for the duration of the command
+// captureInterval blocks for the duration of the command
 // specified by the duration flag, capturing the dynamic
 // targets at the interval specified
-func (c *cmd) captureDynamic() error {
-	successChan := make(chan int64)
-	errCh := make(chan error)
+func (c *cmd) captureInterval() error {
```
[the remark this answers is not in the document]
dhiaayachi: Yes it was hard to read, I hope I improved the readability of this part of the code a bit!

website/content/commands/debug.mdx
dhiaayachi: @dnephin this is still true but not fully accurate, do you think we should say more in the doc or this is the right level of detail?

command/debug/debug.go
```diff
@@ -320,212 +323,217 @@ func (c *cmd) captureStatic() error {
...
-	var wg sync.WaitGroup
+	err := captureShortLived(c)
+	if err != nil {
+		return err
+	}
+	for {
+		select {
+		case t := <-intervalChn.C:
+			intervalCount++
+			c.UI.Output(fmt.Sprintf("Capture successful %s (%d)", t.Local().String(), intervalCount))
```
banks, line 357: Maybe I'm missing something: why do we only output a success message after waiting? What if the `captureShortLived` call below fails?

Should this line maybe come after the call below succeeds?

command/debug/debug.go
```diff
@@ -320,212 +323,217 @@ func (c *cmd) captureStatic() error {
...
+			err := captureShortLived(c)
+			if err != nil {
+				return err
+			}
+		case <-durationChn:
+			intervalChn.Stop()
+			return nil
+		case <-c.shutdownCh:
+			return errors.New("stopping collection due to shutdown signal")
```
banks, line 366: nit-pick: Do we consider this an error case? I'd assume if we are shutting down due to user request they are not expecting to see that reported as an error. But maybe this is consistent with prior behaviour. Not a big deal either way.

command/debug/debug.go
```diff
@@ -320,212 +323,217 @@ func (c *cmd) captureStatic() error {
...
-				// the waitgroup would return on Wait() and repeat without
-				// waiting for the interval.
-				time.Sleep(c.interval)
-
-				wg.Done()
-			}()
-		}
+		g.Go(func() error {
+			return c.captureGoRoutines(timestampDir)
+		})
```
banks, line 386: 🤔 is there any possible downside to capturing these in parallel? for example the heap dump might show memory usage that was in fact only due to the concurrent goroutine dump being run. Both have some slightl overhead too so doing both at the same time potentially compounds the overhead on a production server... 

Honestly, I doubt either is a problem in practice but just made me wonder. Would it be safer just to do each in series here?

banks: Hmm looks like we used to go profile and trace in parallel with goroutines anyway and then heap separately. This probably is fine as it is!

website/content/commands/debug.mdx
banks: This is really the most detailed reference documentation we have for these features so I think this would be a good place to be precise about what users should expect. I don't think we need to go into implementation specifics but I think this would be a good place to describe that `profile` and `trace` will now be single captures for the full duration while `heap` and `goroutines` will be captured per-interval.

website/content/commands/debug.mdx
```diff
@@ -71,14 +71,14 @@ all targets for 2 minutes.
...
+| `host`    | Information about resources on the host running the target agent such as CPU, memory, and disk.                                                                                                                                                                                       |
+| `cluster` | A list of all the WAN and LAN members in the cluster.                                                                                                                                                                                                                                 |
+| `metrics` | Metrics from the in-memory metrics endpoint in the target, captured at the interval.                                                                                                                                                                                                  |
+| `logs`    | `DEBUG` level logs for the target agent, captured for the interval.                                                                                                                                                                                                                   |
```
banks, line 80: I think this should say "captured for the whole duration" now right?

banks (commented):
Hey Dhia, this looks really good to me!

I left a few questions inline which are mostly just thoughts I had than things that need to change.

Is this now complete and working? It looks like it does what was discussed which makes sense to me! The new tests look good too.

I think with a few tweaks to the docs as noted this is looking good to me!

.changelog/10279.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:improvement
+debug: Capture a single file when creating pprof debug capture, the capture will be for the duration parameter
```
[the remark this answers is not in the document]
dhiaayachi: 👍

command/debug/debug.go
dhiaayachi: the debug capture is user triggered, this will error when the application is stopped (by SIGINT or SIGTERM) while capturing debug logs. Could be useful in the case of an oom killer triggering the process to be killed.

command/debug/debug.go
dhiaayachi: This will never be printed of `captureShortLived` return an error. The logic is a bit inverted because we want to run `captureShortLived ` before waiting for an `interval` long, this is why `captureShortLived` is called before the loop start first.

command/debug/debug.go
dhiaayachi: previously goroutine and heap were run in series (each alone). Profile and traces in parallel. I can extract heap and profile in a separate routine that run before we run profile and trace captures and then run profile and trace in parallel with metrics and logs

website/content/commands/debug.mdx
dhiaayachi: I updated it with more details, let me know if this is ok

command/debug/debug.go
dhiaayachi: I investigated the changes needed for this one and it's hard to separate because we run the profile and trace for the whole duration without interruption now we can't stop them while doing the goroutine and heap dump. the other option I can think of is to introduce another level of granularity where we can specify the pprof metrics we want to capture any of (cpu, heap, goroutine and trace). We should do it in a separate PR, this one have already a lot of changes.

command/debug/debug.go
banks: Hmmm. Wouldn't it be clearer to the operator if they saw this right after the capture completes though?

I think I'd be confused if I was looking at debug logs and saw all the requests come in but nothing output for 30 seconds and then a message saying "success" right before the next interval and set of requests came in.

Perhaps this is how it worked before though? I don't think it's a huge deal it just puzzled me why we'd put it this way around rather than:

 1. initial `captureShortLived`
 2. Output success
 3. wait
 4. goto 1

So that the success message is shown right after the operation rather than at the end of the interval?

command/debug/debug.go
banks: Yeah I came to the conclusion it probably doesn't matter anyway - unless we see issues with this in practice I'd say leave it like this for now 👍

website/content/commands/debug.mdx
```diff
@@ -71,14 +71,14 @@ all targets for 2 minutes.
...
+| `logs`    | `DEBUG` level logs for the target agent, captured for the duration.                                                                                                                                                                                                                                                                                                                                                       |
+| `pprof`   | Golang heap, CPU, goroutine, and trace profiling. CPU and traces are captured for `duration` in a single file while heap and goroutine are snapshots for each `interval` captured in multiple files for `duration`. This information is not retrieved unless [`enable_debug`](/docs/agent/options#enable_debug) is set to `true` on the target agent or ACL are enable and an ACL token with `operator:read` is provided. |
```
banks, line 81:
```suggestion
| `pprof`   | Golang heap, CPU, goroutine, and trace profiling. CPU and traces are captured for `duration` in a single file while heap and goroutine are separate snapshots for each `interval`. This information is not retrieved unless [`enable_debug`](/docs/agent/options#enable_debug) is set to `true` on the target agent or ACLs are enable and an ACL token with `operator:read` is provided. |
```

Ha ha I know I said this was a good place for detail 😄 I think the "multiple files for duration" part might actually make it harder to understand without a little more clarification which might be overkill here. This is slightly _less_ specific but I think clear enough. What do you think?

banks (approved):
I think this looks good.

Not sure if other reviewers want to take another pass - might be worth a ping to see but I'll approve this 😄 .

There was one integration test failure that looks like a flake and I just re-ran it. If it fails again it might be worth double-checking but I expect it's unrelated as I don't think we even run this command in envoy integration tests anywhere.

command/debug/debug.go
dhiaayachi: I got your point, should be fixed now.

website/content/commands/debug.mdx
dhiaayachi: > I think this looks good.
> 
> Not sure if other reviewers want to take another pass - might be worth a ping to see but I'll approve this 😄 .
> 
> There was one integration test failure that looks like a flake and I just re-ran it. If it fails again it might be worth double-checking but I expect it's unrelated as I don't think we even run this command in envoy integration tests anywhere.

I will check if @dnephin or @freddygv  can take a quick look and merge it. 
Thank you 😄 !!

command/debug/debug.go
```diff
@@ -320,212 +323,218 @@ func (c *cmd) captureStatic() error {
...
+	if err != nil {
+		return err
+	}
+
+	err = ioutil.WriteFile(fmt.Sprintf("%s/%s.json", timestampDir, "metrics"), marshaled, 0644)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
dnephin, line 535: minor:

```suggestion
	return err
```

dnephin (approved):
LGTM, couple minor things, but nothing blocking

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/381978.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 005ad9e46de6223a9c1423301b057cb3dfb58fa6 onto `release/1.10.x` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 005ad9e46de6223a9c1423301b057cb3dfb58fa6 onto `release/1.9.x` succeeded!

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 005ad9e46de6223a9c1423301b057cb3dfb58fa6 onto `release/1.8.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/381978)

PR #10259, opened 2021-05-19
ui: change coloring of secondary navigation elements

johncowen:
Following our branding update, we noticed that the new `rose` color we use as a highlight on our secondary nav/tabs looked a little weird next to our `red-500` which we use quite a lot. Considering current `structure` uses a `blue-500` for the highlight for the tabs, we took the opportunity to bring our tab coloring into line with that. This was a one line change here https://github.com/hashicorp/consul/pull/10259/files#diff-e0f0d30e1061b08e14e2c0c0ad168b4ce7fc4c638576abbb7893df5777433590L16, but we also:

1. Added a gray bottom border to the tabs whilst hovering (something else I noticed we we differed slightly from current `structure`)
2. Moved all the grays to use CSS properties rather than SASS variables now we have those (confusing _only all our grays_ are currently back to CSS properties)
3. I removed a top border that a realized we no longer need and we overwrite here: https://github.com/hashicorp/consul/blob/7dc78b39c9c5f5120619e70ba168a2d5c12081f2/ui/packages/consul-ui/app/styles/components/app-view/skin.scss#L26-L31 (theres a upcoming PR (https://github.com/hashicorp/consul/pull/10265) to remove that also now we don't need it)
4. Added docs for `TabNav` and noted that it's due for an overhaul since I think it's the first component we ever had.

tabs

PR #10284, opened 2021-05-24
[Backport Revert 1.9.x] Revert of auto backport #10259

johncowen:
We mistakenly backported https://github.com/hashicorp/consul/pull/10259 to 1.9.x.

Probably best to read the above PR description to help understand the why, but tldr is:

1. 1.10.x changed the main navigation
2. We changed the branding in 1.9.x **and** 1.10.x but in 1.9.x whilst we change the logo, the logo color stays white, whereas in 1.10 we changed the main logo to the new `rose` color.
3. Changing to the `rose` color, meant that the new coloring clashed slightly with our `red-500`.
4. As this PR was about 're-branding' it seemed like it needed to go onto 1.9 and 1.10 - but as the rebrand is slightly different (one has color changes and one doesn't), actually it is only needed on 1.10.


kaxcode (approved):
👍 LGTM

PR #10289, opened 2021-05-24
Add readme for network area implementation

freddygv:
This provides some implementation details for the enterprise-only network areas.

TODO for this PR:
- [x] Mention changes in TLS configuration. If TLS is configured in the DC it also applies to network areas.
- [x] Mention the `consul.area.connections.outgoing` metrics which can be used to debug area connectivity.


contributing/network-areas/README.md
```diff
@@ -0,0 +1,54 @@
+# Network Areas (ENTERPRISE ONLY)
+
+## Overview
+
+### Description and Background
+Network areas define pairwise gossip pools over the WAN between Consul datacenters. Gossip traffic between Consul servers in an area is done over the server RPC port, and uses TCP connections. This is unlike Consul's primary WAN and LAN pools, which primarily rely on UDP-based probes. 
+
+TCP was used because it allows configuration with TLS certificates. The TLS encryption will be used if configured for the datacenter, it does not need to be configured for the network areas in particular. The overhead of the TCP protocol is limited since there is a small number of servers in an area. Note that the symmetric keys used by Consul's WAN and LAN pools are not used to encrypt traffic from network areas.
```
i0rek, line 8:
```suggestion
TCP was used because it allows configuration with TLS certificates. The TLS encryption will be used if [configured](https://www.consul.io/docs/security/encryption#rpc-encryption-with-tls) and [enabled](https://www.consul.io/api-docs/operator/area#usetls). Because gossip connects every node in that pool, there is also a connection between servers from the same datacenter. Connections between servers from the same datacenter default to using TLS if it is configured. The overhead of the TCP protocol is limited since there is a small number of servers in an area. Note that the symmetric keys used by Consul's WAN and LAN pools are not used to encrypt traffic from network areas.
```

i0rek (approved):
LGTM, left a comment trying to explain it more detailed when TLS needs to be enabled and when not.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/374594.

PR #10298, opened 2021-05-26
ui: Move all our icons to use CSS custom properties instead of SASS vars

johncowen:
We've been slowly but surely moving all of our colors (and other things) to use CSS custom properties instead of SASS vars  now that we have them available.

The good thing about SASS vars is, if you don't use them they get removed from the final build. Whereas with CSS we have no tree-shaking to get rid of unused CSS custom properties. We can mostly work around this and for some things like colors its no big deal if we have some hex-codes in the build that we don't use as hex-codes are relatively small.

We use CSS for all our decorative imagery including icons, and we would rather not include any icons we don't use, whilst still being able to move everything away from SASS vars.

This PR makes use of the 'tree-shaking' abilities of `@extend` to ensure that we only compile in the icons that we use.

This PR is mostly churn-less as we already use `@extend` for the majority of our icons, so generally there is zero change here for working on the UI, but I did spot one single place where we were using SASS vars instead of `@extend`. This now uses the new form (second commit)

Interestingly this reduces our CSS payload by ~2kb to ~53kb 🤔  (around 25kb of that is these icons)


kaxcode (approved):
🧹 LGTM

johncowen:
Theres been another update to `@hashicorp/structure-icons` since this was done, which now includes some of our icons that we have as custom icons from `iris`/`base` (https://github.com/hashicorp/structure-icons/pull/28). I'll probably update `@hashicorp/structure-icons` again and either add to the end of here, or put a PR ontop before I merge this.

johncowen:
I'd rather get this in than leave it hanging around any more just to do another update of structure-icons, we can always do that at a later date, so I'm gonna merge this now.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/390327.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 9eab715149f63705dfe10161e069be3e270a1995 onto `release/1.10.x` succeeded!

PR #10307, opened 2021-05-27
Support getting nil in testutil.NewTestServerConfigT()

remilapeyre:
Some Consul Enterprise users have asked the Consul backend of Terraform
to support storing the Terraform state in a Namespace. Since Terraform
uses an old version of the Consul SDK it needs to be updated to be able
to specify a namespace when interacting with the KV store.

Terraform uses testutil.NewTestServerConfig(cb) in the tests to start the
Consul server against it runs the tests at https://github.com/hashicorp/terraform/blob/abf7f3416b45c10f2e62e83d5e223a9fb9d6228b/internal/backend/remote-state/consul/backend_test.go#L21-L54
This functions has been replaced by testutil.NewTestServerConfigT(t, cb)
which takes both a *testing.T interface and a configuration callback.
Since testutil.NewTestServerConfig is called only once in TestMain there
is no *testing.T to give it.

testutil.NewTestServerConfigT has some support to receive nil instead of
the *testing.T object (https://github.com/hashicorp/consul/blob/4c7f5f31c718ea1ee00ef93545177c951376b51f/sdk/testutil/server.go#L231-L234)
but it is incomplete (https://github.com/hashicorp/consul/blob/4c7f5f31c718ea1ee00ef93545177c951376b51f/sdk/testutil/server.go#L253)
making the tests passing if they do this.

This just fixes this code path and add TestAPI_NewTestServerConfigT() to
exercise it so that it continues to work in the future.

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -1078,3 +1078,21 @@ func TestAPI_GenerateEnvHTTPS(t *testing.T) {
 
 	require.Equal(t, expected, c.GenerateEnv())
 }
+
+func TestAPI_NewTestServerConfigT(t *testing.T) {
+	configCallback := func(c *testutil.TestServerConfig) {
+		c.Bootstrap = true
+	}
+
+	// with *testing.T
+	server, err := testutil.NewTestServerConfigT(t, configCallback)
+	require.NoError(t, err)
+	server.WaitForLeader(t)
+	server.Stop()
+
+	// without *testing.T
+	server, err = testutil.NewTestServerConfigT(nil, configCallback)
+	require.NoError(t, err)
+	server.WaitForLeader(t)
+	server.Stop()
+}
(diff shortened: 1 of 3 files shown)


remilapeyre:
The failure in `go-tests` are TestLeader_SecondaryCA_IntermediateRefresh() and TestConnectCAConfig_GetSetForceNoCrossSigning() and don't look related to this change. I don't whether this change needs an entry in the changelog thought.

dnephin (commented):
Thank you for the PR! 

Accepting `nil` for `testing.T` can be a problem because it makes it difficult to use new features like `T.Cleanup`.

Since this function already accepts an interface `TestingTB`, it is possible to pass an implementation of that interface that doesn't use `testing.T`. There is a great example of an implementation here: https://github.com/hashicorp/consul-template/blob/870905de57f085588c3b718b779d8550aefc5dcf/test/helpers.go#L80 (and  https://pkg.go.dev/github.com/mitchellh/go-testing-interface#RuntimeT is another example, but does not handle cleanup properly).

Would this solution work for your use case?

remilapeyre:
Thanks for the feedback, I will just updated the tests.

PR #10275, opened 2021-05-21
ui: Colocate old base components into our app/component folder

johncowen:
Since https://github.com/hashicorp/consul/pull/7403 we've been trying to ~colocate~ chocolate as many component related files as possible into component folders, this includes:

1. The template hbs files (https://github.com/hashicorp/consul/pull/7403)
2. Our pageobject test-support files (https://github.com/hashicorp/consul/pull/8098)
3. Component related (other than component `class` files which are already chocolated) JS utilities (https://github.com/hashicorp/consul/pull/7892)
4. and finally, our component CSS files (https://github.com/hashicorp/consul/pull/8660)

Due to how things have gone, previous to the above PRs, the files where in other conventional ember places, and now our conventions have changed we've been slowly moving files over as we've worked on these files for other reasons.

As we've been doing this little by little, and some files just haven't needed to be touched for literally years, it's become a little confusing as to where files should go "Which component folder of the 3 we have? The javascript one, the styles one, or our pseudo addon one?". So we felt it was time to move the remaining files.

Our 'pseudo addon' in `styles/base`, is where we'd started adding generic components due to us having a `html` storybook install there where we could build/manually test out components in isolation. We now use the excellent `docfy` to build, document and catalog our components, so there is no reason to keep these components in the 'pseudo addon'.

This PR moves all of our 'pseudo addon' (`styles/base/components`) CSS component files into the app component folder. So from now on:

"Q: Where are the CSS files for this component? A: Right next to the javascript and hbs files in the same folder."

 ~Colocated~ Chocolated

There are still few more 'non-pseudo addon' CSS files in the old `styles/components` folder, so we aren't quite there yet, but there'll be a follow up PR to move those coming up soon also, as well as some sort of solution for moving our remaining component related files (our translations yaml files for components).

Lastly the remaining thing to do here will be to import the component CSS files along with the javascript/hbs files, so we don't have to keep adding to the growing list of imports in `components.scss`.

PR #10328, opened 2021-06-01
ui: Colocate remaining CSS components to the app/components folder

johncowen:
Second part of https://github.com/hashicorp/consul/pull/10275  

> There are still few more 'non-pseudo addon' CSS files in the old styles/components folder, so we aren't quite there yet, but there'll be a follow up PR to move those coming up soon also, as well as some sort of solution for moving our remaining component related files (our translations yaml files for components).

🍫

We still have translation files and this:

> Lastly the remaining thing to do here will be to import the component CSS files along with the javascript/hbs files, so we don't have to keep adding to the growing list of imports in components.scss. 

to do at some point.

There are a few CSS files still left in `styles` but these are pretty much 'tokens' and other global like layout things, open to suggestions about whether these should stay there or move (and if we should move them then where to?) Personally I'm good to leave them where they are for the minute at least.


kaxcode:
> There are a few CSS files still left in `styles` but these are pretty much 'tokens' and other global like layout things, open to suggestions about whether these should stay there or move (and if we should move them then where to?) Personally I'm good to leave them where they are for the minute at least.

Yea, let's keep global styles/layouts in the styles folder. @johncowen

kaxcode:
@johncowen looks like this PR needs to be rebased.

johncowen:
Thanks! Rebasing was a little weird, but I think this is good to go again 👍

kaxcode (approved):
"needs-love" 😂 but LGTM

johncowen:
😂 " 🎼 ...love is all you need 🎶 "

More seriously tho, as well as being fun, I figured it would be easy to search for that if/when we need to. It would be nice to be able to call out what things we would appreciate work on/help with from within our docs somehow - just not sure exactly how yet 🤔

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/394987.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/395034.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit c78f7ecb27175f50bc4d04cca261cf7379f4c0a6 onto `release/1.10.x` succeeded!

PR #10329, opened 2021-06-01
Add flag for transparent proxies to dial individual instances

freddygv:
This PR adds a new flag to proxy definitions: `Proxy.TransparentProxy.DialedDirectly`.

When enabled, services can opt into allowing transparent proxies to dial individual instances.

Typically transparent proxies only dial virtual IPs representing a logical service name. Dialing individual instances can be helpful in cases like stateful services, like a database cluster with a leader.

xDS configuration:
- There will be a passthrough [original destination cluster](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/listener_filters/original_dst_filter) for each upstream service name. The cluster name is prefixed with passthrough to distinguish from the non-passthrough cluster for the upstream.

- For listeners we will add a filter chain for each logical upstream that routes to the passthrough clusters mentioned above.

Limitations:
- The protocol and connection timeout of the upstream are not respected and default to tcp and 5s respectively.
- L7 rules like [RetryOnStatusCodes](https://www.consul.io/docs/connect/config-entries/service-router#retryonstatuscodes) would likely not be able to be supported with this approach. See [this comment](https://github.com/hashicorp/consul/pull/10329#discussion_r647005187)

Main Commits:
1. Adds the DialedDirectly option
2. Updates proxycfg and xds handling

TODO:
- [x] Add proxycfg and xDS tests


command/connect/envoy/bootstrap_config.go
```diff
@@ -352,40 +352,41 @@ func resourceTagSpecifiers(omitDeprecatedTags bool) ([]string, error) {
 		// - cluster.f8f8f8f8~pong.default.dc2.internal.e5b08d03-bfc3-c870-1833-baddb116e648.consul.bind_errors: 0
 		// - cluster.v2.pong.default.dc2.internal.e5b08d03-bfc3-c870-1833-baddb116e648.consul.bind_errors: 0
 		// - cluster.f8f8f8f8~v2.pong.default.dc2.internal.e5b08d03-bfc3-c870-1833-baddb116e648.consul.bind_errors: 0
+		// - cluster.passthrough.pong.default.dc2.internal.e5b08d03-bfc3-c870-1833-baddb116e648.consul.bind_errors: 0
```
freddygv, line 355: This was updated so we can continue to correctly label metrics for this upstream, in the presence of two distinct cluster names.

website/content/docs/connect/transparent-proxy.mdx
```diff
@@ -149,3 +153,16 @@ it can dial `sample-app.default.svc.cluster.local`, using
...
+configured using the [DialedDirectly](/docs/connect/config-entries/service-defaults#dialeddirectly) 
+option. 
+
+Individual instance addresses can then be discovered using DNS, and dialed through the transparent proxy.
+When this mode is enabled on the upstream, connect certificates will be presented for mTLS and
+intentions will be enforced at the destination.
+
+Note that when dialing individual instances HTTP routing rules configured with config entries 
+will **not** be considered. The transparent proxy acts as a TCP proxy to the original
+destination IP address.
```
freddygv, line 168: This is the main bit that gives me pause about the current implementation.

Configuration like retrying requests is applied on HTTP routes, and currently the network filter for passthrough is a TCP proxy. If we did attach an HTTP connection manager, what route config would it use? If it used the one for the upstream _name_, we would no longer route to the passthrough cluster and would instead load balance across the route destination endpoints.

I'm not sure that we can support L7 rules for headless services without creating a separate service registration for each upstream address. This seems really limiting.

agent/proxycfg/state.go
```diff
@@ -885,33 +887,38 @@ func (s *state) handleUpdateConnectProxy(u cache.UpdateEvent, snap *ConfigSnapsh
...
 
 	default:
-		return s.handleUpdateUpstreams(u, &snap.ConnectProxy.ConfigSnapshotUpstreams)
+		return s.handleUpdateUpstreams(u, snap)
 	}
 	return nil
 }
 
-func (s *state) handleUpdateUpstreams(u cache.UpdateEvent, snap *ConfigSnapshotUpstreams) error {
+func (s *state) handleUpdateUpstreams(u cache.UpdateEvent, snap *ConfigSnapshot) error {
```
rboyer, line 895: This change breaks ingress gateways.

agent/xds/clusters.go
```diff
@@ -176,6 +168,58 @@ func makeExposeClusterName(destinationPort int) string {
...
 }
 
+// In transparent proxy mode there are potentially multiple passthrough clusters added.
+// The first is for destinations outside of Consul's catalog. This is for a plain TCP proxy.
+// All of these use Envoy's ORIGINAL_DST listener filter, which forwards to the original
+// destination address (before the iptables redirection).
+// The rest are for destinations inside the mesh, which require certificates for mTLS.
+func makePassthroughClusters(cfgSnap *proxycfg.ConfigSnapshot) ([]proto.Message, error) {
+	// This size is a lower bound.
+	clusters := make([]proto.Message, 0, len(cfgSnap.ConnectProxy.PassthroughUpstreams))
```
rboyer, line 178:
```suggestion
	clusters := make([]proto.Message, 0, len(cfgSnap.ConnectProxy.PassthroughUpstreams)+1)
```

agent/xds/clusters.go
```diff
@@ -176,6 +168,58 @@ func makeExposeClusterName(destinationPort int) string {
...
+				Type: envoy_cluster_v3.Cluster_ORIGINAL_DST,
+			},
+			LbPolicy:       envoy_cluster_v3.Cluster_CLUSTER_PROVIDED,
+			ConnectTimeout: ptypes.DurationProto(5 * time.Second),
+		})
+	}
+
+	for _, passthrough := range cfgSnap.ConnectProxy.PassthroughUpstreams {
+		// Prefixed with passthrough to distinguish from non-passthrough clusters for the same upstream.
+		name := "passthrough." + passthrough.SNI
```
rboyer, line 195: What if someone had a subset of a cluster named `passthrough`? I think then you would have two clusters with the same name potentially generated and they'd conflict with each other (the tproxy passthrough, and the plain subset)

rboyer: You might go with the something like the tilde trick I used for the "customized hash" prefixes.

agent/xds/listeners.go
```diff
@@ -209,7 +219,7 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
 			}
 
 			filterChain, err := s.makeUpstreamFilterChainForDiscoveryChain(
-				"passthrough",
```
rboyer, line 212: Intentional?

agent/xds/listeners.go
```diff
@@ -1254,7 +1293,10 @@ func (s *ResourceGenerator) makeUpstreamFilterChainForDiscoveryChain(
 	if overrideCluster != "" {
 		useRDS = false
 		clusterName = overrideCluster
-		filterName = overrideCluster
+
+		if destination == "" {
+			filterName = overrideCluster
+		}
```
rboyer, line 1299: What is this change here for?

agent/xds/clusters.go
freddygv: Ah yep, swapping this out for a tilde as well.

agent/xds/listeners.go
freddygv: Actually that should be empty. That field is used for discovery chains IDs to get hooked up as the route name. None of that is relevant for this plain TCP-passthrough.

agent/xds/listeners.go
freddygv: If the filterName uses the overrideCluster we end up with a filterName that contains the full SNI. That filterName is then used in the statPrefix, which should only have `name.ns.dc`.

This is a hack to get it to use the one from L1287, which is what's expected in the filter's statPrefix. 

This gets back to this function needing to be rewritten because it is _very_ confusing. I'm going to do that before the RC release.

rboyer (approved):
LGTM

agent/xds/listeners.go
```diff
@@ -162,16 +160,9 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
 		if filterChain.FilterChainMatch != nil && len(filterChain.FilterChainMatch.PrefixRanges) > 0 {
 			outboundListener.FilterChains = append(outboundListener.FilterChains, filterChain)
 		}
-		hasFilterChains = true
 	}
 
-	// Only create the outbound listener when there are upstreams and filter chains are present
-	if outboundListener != nil && hasFilterChains {
```
freddygv, line 194: If no chains with matches on virtual IPs were created above then we would skip the block below, which was unintended.

agent/xds/listeners.go
```diff
@@ -197,7 +197,8 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
 
 		// Filter chains are stable sorted to avoid draining if the list is provided out of order
 		sort.SliceStable(outboundListener.FilterChains, func(i, j int) bool {
-			return outboundListener.FilterChains[i].Name < outboundListener.FilterChains[j].Name
+			return outboundListener.FilterChains[i].FilterChainMatch.PrefixRanges[0].AddressPrefix <
```
rboyer, line 200: And this makes sense because the `PrefixRanges` slice is already sorted stably by `AddressPrefix`, so taking the zeroth element to compare on is fine.

agent/proxycfg/state.go
```diff
@@ -926,10 +933,53 @@ func (s *state) handleUpdateUpstreams(u cache.UpdateEvent, snap *ConfigSnapshotU
...
+
+				svc := node.Service.CompoundServiceName()
+
+				// Overwrite the name if it's a connect proxy (as opposed to Connect native).
+				// We don't reference the proxy name directly for things like SNI, but rather the name
+				// of the destination. The enterprise meta of a proxy will always be the same as that of
+				// the destination service, so that remains intact.
+				if node.Service.Kind == structs.ServiceKindConnectProxy {
+					dst := node.Service.Proxy.DestinationServiceName
+					if dst == "" {
```
rboyer, line 957: isn't this backwards? I thought IDs were more specific than Names

rboyer (approved):
LGTM again with one small question

agent/proxycfg/state.go
freddygv: I think we want to prefer the name for the SNI, and only fall back if it doesn't exist.

Follows this: https://github.com/hashicorp/consul/blob/master/agent/structs/structs.go#L952

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/383864.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 429f9d8bb8fcbcb2c522294d77e32024c33d8a2d onto `release/1.10.x` succeeded!

Issue #9572, opened 2021-01-15
Ensure CA provider config changes result in certificate regeneration across all CA providers


preetapan:
Currently, changes to two different fields `private-key-bits` and `private-key-type` in the CA provider config for Consul connect, don't result in regenerating root certificates. 

All providers implement this logic differently, and we will need to create follow up issues per provider to make sure nuances like the intermediate mount point in the Vault CA provider is updated correctly.

tristanmorgan:
Hi @preetapan , should the changes in private-key-bits always result in the root CA (as opposed to the consul controlled intermediate CA) requiring a reconfiguration?

My example is: I have a Root CA that has EC, 521 bit keys for its certificate while my Intermediate still EC, used 256 bit keys and generates leaf certs with the same EC 256. 

With the changes in #10331 I now get the error of `cannot update the PrivateKeyBits field without choosing a new PKI mount for the root CA` when I try to use the same Root CA mount for Consul Connect.

dnephin:
Thank you for reporting this problem! I just wrote up #12246 which I believe may describe the problem you encountered. Given these problems, I think we may need to either change this validation, or remove it.

PR #9556, opened 2021-01-13
structs: Add more cache key completeness tests

dnephin:
related issue: #9437, this PR builds on #9480

This PR adds cache key completeness tests for the remaining types in `structs`. See the comment on `assertCacheInfoKeyIsComplete`, or the description in the linked issue, for how this works.

There are 2 or 3 structs outside of the `structs` package that also implement the `cache.Request` interface, but to test those we'll need to extract this assertion function to an `internal/testing/` helper package so we can export it.

PR #10330, opened 2021-06-01
connect/ca: ensure edits to the key type/bits for the connect builtin CA will regenerate the roots

rboyer:
progress on #9572

### DIFF
--- a/agent/connect/ca/provider_consul.go
+++ b/agent/connect/ca/provider_consul.go
@@ -60,6 +60,11 @@ type ConsulProviderStateDelegate interface {
 	ApplyCARequest(*structs.CARequest) (interface{}, error)
 }
 
+func hexStringHash(input string) string {
+	hash := sha256.Sum256([]byte(input))
+	return connect.HexString(hash[:])
+}
+
 // Configure sets up the provider using the given configuration.
 func (c *ConsulProvider) Configure(cfg ProviderConfig) error {
 	// Parse the raw config and update our ID.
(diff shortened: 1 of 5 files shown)


agent/connect/ca/provider_consul.go
```diff
@@ -63,8 +69,7 @@ func (c *ConsulProvider) Configure(cfg ProviderConfig) error {
 		return err
 	}
 	c.config = config
-	hash := sha256.Sum256([]byte(fmt.Sprintf("%s,%s,%v", config.PrivateKey, config.RootCert, cfg.IsPrimary)))
-	c.id = connect.HexString(hash[:])
+	c.id = hexStringHash(fmt.Sprintf("%s,%s,%s,%d,%v", config.PrivateKey, config.RootCert, config.PrivateKeyType, config.PrivateKeyBits, cfg.IsPrimary))
```
rboyer, line 72: This `id` is used to key the builtin proxy configuration in the state store (much like a cache-key). Adding the key type/bits here allows it to have two builtin proxy roots defined simultaneously during the rotation.

dnephin: At first I thought `RootCert` should have already been unique here, but after reading the tests I realize now that this input struct is a `ConsulCAProviderConfig` , not the `CARoots`, which means in most cases when `PrivateKey{Type,Bits}` are set, `RootCert` would be the zero value.

agent/consul/leader_connect_test.go
```diff
@@ -25,6 +25,156 @@ import (
...
+	}{
+		{connect.DefaultPrivateKeyType, connect.DefaultPrivateKeyBits},
+		{"ec", 256},
+		{"ec", 384},
+		{"rsa", 2048},
+		{"rsa", 4096},
+	}
+
+	for _, src := range types {
+		for _, dst := range types {
```
dnephin, line 45: I wish we could test these permutations without having to spin up a full server. Hopefully after a few changes like #10445 we'll be able to test this with a full persistence backend, but without serf and all the other parts that aren't used here.

dnephin (approved):
LGTM! This seems like it should fix the bug with the Consul CA provider.

I also wonder how we could prevent bugs like this in the future. We could write a test, similar to  #9556, which verifies that all the important fields are considered in the hash. We can ignore any existing fields that we don't need in the hash.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/408197.

PR #10340, opened 2021-06-02
usagemetrics: add cluster members to metrics API

pcewing:
This PR adds cluster members to the metrics API. The number of members per
segment are reported as well as the total number of members.

Tested by running a multi-node cluster locally and ensuring the numbers were
correct. Also added unit test coverage to add the new expected gauges to
existing test cases.


pcewing:
Realized that `members.servers` doesn't need to be reported on a per-segment basis because servers should always be in the default segment. Will get that fixed and also add an entry to the changelog to appease the checks later tonight!

agent/consul/usagemetrics/usagemetrics.go
```diff
@@ -77,9 +88,10 @@ type UsageMetricsReporter struct {
 	metricLabels   []metrics.Label
 	stateProvider  StateProvider
 	tickerInterval time.Duration
+	getMembersFunc getMembersFunc
 }
 
-func NewUsageMetricsReporter(cfg *Config) (*UsageMetricsReporter, error) {
+func NewUsageMetricsReporter(cfg *Config, getMembersFunc getMembersFunc) (*UsageMetricsReporter, error) {
```
mkeeler, line 94: Do you think it would be a little more consistent if the `getMembersFunc` was specified within the `Config` along with having a `WithMemberGetter` method for building the configuration?

agent/consul/usagemetrics/usagemetrics.go
```diff
@@ -137,4 +150,66 @@ func (u *UsageMetricsReporter) runOnce() {
 	}
 
 	u.emitServiceUsage(serviceUsage)
+
+	servers, clients := u.memberUsage()
+	u.emitMemberUsage(servers, clients)
+}
+
+func (u *UsageMetricsReporter) memberUsage() (int, map[string]int) {
+	if u.getMembersFunc == nil {
```
mkeeler, line 159: I wonder if it would be better to add a nil check in the reporters constructor and error out there instead of reporting 0 metrics. I am having a hard time thinking of a case where it would be okay to create a reporter without one.

agent/consul/usagemetrics/usagemetrics.go
```diff
@@ -137,4 +150,66 @@ func (u *UsageMetricsReporter) runOnce() {
...
+	clients := make(map[string]int)
+
+	for _, m := range mems {
+		if m.Status != serf.StatusAlive {
+			continue
+		}
+
+		switch m.Tags["role"] {
+		case "node":
+			clients[m.Tags["segment"]]++
```
mkeeler, line 179: Nice! I am glad it worked out for you where you didn't need a nested for loop over all segments and then over members of the segment.

agent/consul/usagemetrics/usagemetrics_oss_test.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/hashicorp/consul/agent/consul/state"
 	"github.com/hashicorp/consul/agent/structs"
 	"github.com/hashicorp/consul/sdk/testutil"
+	"github.com/hashicorp/serf/serf"
```
mkeeler, line 15: Just a small note that we will probably want to follow up with some tests in enterprise as well to ensure the right segment gets reported.

mkeeler (commented):
Overall this is great. I have a few minor nits but nothing major.

agent/consul/usagemetrics/usagemetrics.go
pcewing: I considered putting it in the `Config` type but it felt more like functionality and not configuration? Wouldn't argue with changing it though

agent/consul/usagemetrics/usagemetrics.go
pcewing: I agree, will error out in the constructor

agent/consul/usagemetrics/usagemetrics_oss_test.go
pcewing: Theoretically that could be tested in the OSS tests as well, all we'd need to add is `segment` tags to the items in the array returned by the `getMembersFunc` we test with

mkeeler: oooh, even better

agent/consul/usagemetrics/usagemetrics.go
pcewing: Went ahead and made this change, feels much cleaner. Thanks for the suggestions!

agent/consul/usagemetrics/usagemetrics_oss_test.go
pcewing: Just added this to the test. Now instead of 1 client there are two, each in their own segment (`"a"` and `"b"` respectively).

agent/consul/usagemetrics/usagemetrics.go
mkeeler: You are 100% right that its functionality and not configuration in the traditional sense. However everywhere else in the Consul code we would put both into the config struct. 

Really the only reason there is a Config type at all here is so we don't have to pass so many parameters into the constructor but can build it up. We could call it Parameters or something to indicate its real purpose too.

mkeeler (approved):
LGTM. I think we also need to update the telemetry docs in this table here: https://github.com/hashicorp/consul/blob/master/website/content/docs/agent/telemetry.mdx#metrics-reference

website/content/docs/agent/telemetry.mdx
```diff
@@ -296,6 +296,8 @@ This is a full list of metrics emitted by Consul.
...
+| `consul.members.clients`                                 | Measures the current number of client agents registered with Consul. It is only emitted by Consul servers. Added in v1.9.6.                                                                                                                                                                                                                                                                                    | number of objects    | gauge   |
+| `consul.members.servers`                                 | Measures the current number of server agents registered with Consul. It is only emitted by Consul servers. Added in v1.9.6.                                                                                                                                                                                                                                                                                    | number of objects    | gauge   |
```
mkeeler, line 300:
```suggestion
| `consul.members.clients`                                 | Measures the current number of client agents registered with Consul. It is only emitted by Consul servers. Added in v1.9.6.                                                                                                                                                                                                                                                                                    | number of clients    | gauge   |
| `consul.members.servers`                                 | Measures the current number of server agents registered with Consul. It is only emitted by Consul servers. Added in v1.9.6.                                                                                                                                                                                                                                                                                    | number of servers    | gauge   |
```

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/380096.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 42a51b1a2c9f7ffc325bae7790ccd1dd088ea870 onto `release/1.10.x` succeeded!

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 42a51b1a2c9f7ffc325bae7790ccd1dd088ea870 onto `release/1.9.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/380096)

Issue #8329, opened 2020-07-17
ci: Run unit tests with -race


dnephin:
I've attempted to enable 'go test -race', but there are many failures  ([example](https://circleci.com/gh/hashicorp/consul/204265)). Data races could cause flaky tests, and may even cause issues in production.

I'm starting work on this in #8269. This issue is to track packages that have failures so that we can fix them incrementally.

* [ ] agent/cache-types
* [ ] agent/local
* [ ] agent/xds
* [ ] connect - `dynamicTLSConfig.readyCh` is set to nil, which races with `ReadyWait`
* [ ] connect/proxy
* [ ] agent
* [ ] agent/consul
* [ ] snapshot (in enterprise only, some tests require a Consul binary, no races were found)
* [ ] (other packages not yet listed)

dnephin:
I'm seeing some of the `TestCheck*` tests  in `agent/checks` fail on master due to races (https://circleci.com/gh/hashicorp/consul/209863, https://circleci.com/gh/hashicorp/consul/209838).

Some of the stacks are missing (I think that is due to another bug in Go that I patched and will be in go1.15). I tried to reproduce locally and it appears to be a race with `t.Log` and the test shutting down. I guess the goroutines outlive the test, so maybe the root cause is a leaked go routine.

I think https://github.com/hashicorp/consul/pull/7948 may fix the immediate problem with the race.

```
WARNING: DATA RACE
Read at 0x00c000496b83 by goroutine 28:
  testing.(*common).logDepth()
      /usr/lib/go/src/testing/testing.go:675 +0x14b
  testing.(*common).log()
      /usr/lib/go/src/testing/testing.go:662 +0x77
  testing.(*common).Log()
      /usr/lib/go/src/testing/testing.go:694 +0x1d
  testing.(*T).Log()
      <autogenerated>:1 +0x61
  github.com/hashicorp/consul/sdk/testutil.(*testWriter).Write()
      /home/daniel/pers/code/consul/sdk/testutil/testlog.go:86 +0x319
  github.com/hashicorp/go-hclog.(*writer).Flush()
      /home/daniel/go/pkg/mod/github.com/hashicorp/go-hclog@v0.12.0/writer.go:29 +0x28c
  github.com/hashicorp/go-hclog.(*intLogger).log()
      /home/daniel/go/pkg/mod/github.com/hashicorp/go-hclog@v0.12.0/intlogger.go:140 +0x20c
[...]

PR #10342, opened 2021-06-03
remove a race condition in sdk testutil server stop

dgsb:
We have 2 threads calling `exec.Cmd.Wait` which is not reentrant. It causes some test case failure at our side when running the test suite with the `-race` flag.

I've just removed the `cmd.Wait` which looks useless.

Related to #8329

### DIFF
--- a/sdk/testutil/server.go
+++ b/sdk/testutil/server.go
@@ -342,7 +342,6 @@ func (s *TestServer) Stop() error {
 		return err
 	case <-time.After(10 * time.Second):
 		s.cmd.Process.Signal(syscall.SIGABRT)
-		s.cmd.Wait()
 		return fmt.Errorf("timeout waiting for server to stop gracefully")
 	}
 }
--- a/sdk/testutil/server.go
+++ b/sdk/testutil/server.go
@@ -342,6 +342,7 @@ func (s *TestServer) Stop() error {
 		return err
 	case <-time.After(10 * time.Second):
 		s.cmd.Process.Signal(syscall.SIGABRT)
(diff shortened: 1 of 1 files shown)


sdk/testutil/server.go
```diff
@@ -342,7 +342,6 @@ func (s *TestServer) Stop() error {
 		return err
 	case <-time.After(10 * time.Second):
 		s.cmd.Process.Signal(syscall.SIGABRT)
-		s.cmd.Wait()
```
dnephin, line 345: I think we do still want to wait for the process to exit in this case, because there is a defer at the top of this function, which is responsible for removing the temporary directory. We want to make sure the process has exited before we try and remove it to make sure the cleanup succeeds.

I think we can do that by blocking on a read from the `waitDone` channel instead of calling `s.cmd.Wait()`

```go
<-waitDone
return ...
```

dnephin (commented):
Thank you for the PR and for fixing this data race! We've been tracking the work to enable the race detector in #8329, but we had not seen this race yet.

I've made one suggestion below, but otherwise I think this looks correct.

sdk/testutil/server.go
dgsb: Hello @dnephin that totally makes sense. Can you check the last commit ?

dnephin (approved):
LGTM, thank you!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/383609.

PR #10349, opened 2021-06-04
ci: Go 1.16.5

mikemorris:


dnephin (approved):
LGTM when the build is green

dnephin:
Also, reminder to make this same change in consul-releases repo, so that we use the same version for testing and releasing.

PR #10352, opened 2021-06-04
docs/nia: Add service filter configuration, deprecate tag

mkam:


website/content/docs/nia/configuration.mdx
```diff
@@ -97,7 +97,8 @@ service {
 - `id` - (string) ID identifies the service for Consul-Terraform-Sync. This is used to explicitly identify the service config for a task to use. If no ID is provided, the service is identified by the service name within a [task definition](#task).
 - `name` - (string: required) The Consul logical name of the service (required).
 - `namespace` <EnterpriseAlert inline /> - (string: "default") The namespace of the service. If not provided, the namespace will be inferred from the Consul-Terraform-Sync ACL token, or default to the `default` namespace.
-- `tag` - (string) Tag is used to filter nodes based on the tag for the service.
+- `tag` - (string) **Deprecated** - Use `filter` with the `Service.Tags` selector instead. This field will be removed in a future version of Consul-Terraform-Sync. Tag is used to filter nodes based on the tag for the service.
```
lornasong, line 100: I was looking for [consul config examples of deprecated fields](https://www.consul.io/docs/agent/options#_non_voting_server). I kind of like in the example that it included the version that the field was deprecated. What would you think about something like adding in a version in the bolded deprecated message?

website/content/docs/nia/configuration.mdx
```diff
@@ -97,7 +97,8 @@ service {
...
+- `filter` - (string) Specifies the expression used to filter nodes for the service. For more details on supported filters, see the Consul documentation on [filtering service nodes](https://www.consul.io/api-docs/health#filtering-2).
```
lornasong, line 101: Within consul.io docs, I think links can be a relative path e.g. (...)[/api-docs/....] so that the link can point to the staging website (in case there's changes there too)

lornasong (approved):
looks good to me! left two non-blocking suggestions

website/content/docs/nia/configuration.mdx
mkam: Yeah, that seems helpful to have the version! I'll update the format to be more like that consul config.

PR #10353, opened 2021-06-04
Remove a few unused things from build-support

dnephin:
I believe these files, bash functions, and makefile targets are no longer used. Some are reaplced by other tools(ex: changelogfmt), the rest has moved to consul-releases.

I attempted to verify that all of this was used by using `git grep` in both this repo and in consul-releases. We have a couple copies of these scripts in consul-releases, so they should still all be available to anyone running a release.

There are a few more things that I think could be pruned, but this seemed like a relatively safe and quick first pass.


freddygv (approved):
LGTM, I don't see any stray references to these

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/411834.

PR #10355, opened 2021-06-04
stream: fix a bug with subscriptions and the latest item in the topic buffers.

dnephin:
While working on an end-to-end test for streaming I encountered this bug. The e2e test isn't quite ready yet, but I thought I would surface this bug fix separately.

This PR is best viewed by individual commit.

The head of the topic buffer was being ignored when creating a snapshot. This commit fixes the bug by ensuring that the head of the topic buffer is included in the snapshot before handing it off to the subscription.

### DIFF
--- a/agent/consul/stream/event_snapshot.go
+++ b/agent/consul/stream/event_snapshot.go
@@ -67,12 +67,12 @@ func (s *eventSnapshot) spliceFromTopicBuffer(topicBufferHead *bufferItem, idx u
 			return
 		}
 
-		next := item.NextNoBlock()
-		if next == nil {
+		next, ok := item.NextNoBlock()
+		if !ok {
 			// We reached the head of the topic buffer. We don't want any of the
 			// events in the topic buffer as they came before the snapshot.
 			// Append a link to any future items.
-			s.buffer.AppendItem(item.NextLink())
+			s.buffer.AppendItem(next)
 			return
 		}
 		// Proceed to the next item in the topic buffer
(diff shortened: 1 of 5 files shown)


agent/consul/stream/event_snapshot.go
```diff
@@ -49,34 +49,33 @@ func (s *eventSnapshot) appendAndSplice(req SubscribeRequest, fn SnapshotFunc, t
 func (s *eventSnapshot) spliceFromTopicBuffer(topicBufferHead *bufferItem, idx uint64) {
 	item := topicBufferHead
 	for {
-		next := item.NextNoBlock()
```
banks, line 52: If I understand correctly the location of this call is the source of the bug right?

So before with it here there is a case where `topicBufferHead` pointed to an actual event that should be delivered after the snapshot, but this first iteration skips over it on to the next one and so misses that event?

I guess the assumption here is that because we took a reference to `topicBufferHead` _before_ we made the snapshot, the item it points to is either a nil link-only item (because it was empty), or will be an event that must already be in the snapshot.

I'm interested to know under which circumstances that invariant can be broken and the `topicBufferHead` actually point to an event that is _after_ the snapshot index in your tests.

That said, the changes here all seem reasonable and there's no reason not to have this more robust case _even_ if the original invariant assumptions did hold.

banks (approved):
LGTM.

Approving because I think the minor refactor is reasonable and the bug fix is a solid change regardless. I do have Q inline about how exactly the bug is possible given the invariants I _though_ we held here but regardless of the answer it seems sensible to go with this more robust implementation!

Good catch from e2e tests though!

agent/consul/stream/event_snapshot.go
dnephin: Yup, that all sounds right. The test that triggered this condition returned an `index=1` which was older than the topic buffer head.  I'm looking around now to see if we document this invariant. I imagine it will be important for future topics to understanding what index they need to send back from the snapshot functions. I think you are right that this isn't a bug that can impact users today because the current snapshot functions will always return the index of the topic buffer head.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/382298.

PR #10361, opened 2021-06-07
Migrate from GitHub HashiBot to GitHub Actions

bflad:
GitHub HashiBot is being decommissioned. This migrates the crash issue labeling and waiting-reply label removal to GitHub Actions. The stale issue handling is not migrated as neither behavior appears to be actively in use.


jsosulska (approved):
LGTM!

PR #10368, opened 2021-06-09
improve monitor performance

### LINKED ISSUE #10347: monitor: consul monitor and consul debug are missing many logs
#### Overview of the Issue

`consul monitor` (and `consul debug` which uses the same monitoring) are supposed to show all of the consul logs. Instead they appear to only have logs for some loggers and not others. This problems exists on master, and I believe has been a problem for the last few major releases.

I suspect the issue may actually be in  https://github.com/hashicorp/go-hclog/, but filing the issue here in case it is only related to Consuls implementation.

Edit: removed the reproduction steps, which were not correct

dhiaayachi:
Fixes #10347

improve monitor performance to avoid filling the log channel and loose logs when running monitor or debug.

An issue was reported that we loose some logs when are running `consul monitor` or `consul debug` commands. 

the current underlying monitor implementation avoid blocking when writing logs to not impact performance and allow losing some logs when we overflow the log channel (size 512 lines of code). This PR is to enhance the performance of the log channel read so we can avoid loosing logs.

### DIFF
--- a/.changelog/10368.txt
+++ b/.changelog/10368.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+monitoring: optimize the monitoring endpoint to avoid losing logs when under high load
+```
--- a/.changelog/10368.txt
+++ b/.changelog/10368.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-monitoring: optimize the monitoring endpoint to avoid losing logs when under high load. [GH-10368](https://github.com/hashicorp/consul/pull/10368)
+monitoring: optimize the monitoring endpoint to avoid losing logs when under high load.
 ```
--- a/.changelog/10368.txt
+++ b/.changelog/10368.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-monitoring: optimize the monitoring endpoint to avoid losing logs when under high load
+monitoring: optimize the monitoring endpoint to avoid losing logs when under high load. [GH-10368](https://api.github.com/repos/hashicorp/consul/issues/10368)
 ```
--- a/.changelog/10368.txt
+++ b/.changelog/10368.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-monitoring: optimize the monitoring endpoint to avoid losing logs when under high load. [GH-10368](https://api.github.com/repos/hashicorp/consul/issues/10368)
+monitoring: optimize the monitoring endpoint to avoid losing logs when under high load. [GH-10368](https://github.com/hashicorp/consul/pull/10368)
 ```
--- a/agent/agent_endpoint.go
+++ b/agent/agent_endpoint.go
@@ -1211,10 +1211,10 @@ func (s *HTTPHandlers) AgentMonitor(resp http.ResponseWriter, req *http.Request)
 			if droppedCount > 0 {
 				s.agent.logger.Warn("Dropped logs during monitor request", "dropped_count", droppedCount)
 			}
+			flusher.Flush()
 			return nil, nil
 		case log := <-logsCh:
 			fmt.Fprint(resp, string(log))
-			flusher.Flush()
 		}
 	}
 }
--- a/agent/agent_endpoint.go
+++ b/agent/agent_endpoint.go
@@ -1205,6 +1205,7 @@ func (s *HTTPHandlers) AgentMonitor(resp http.ResponseWriter, req *http.Request)
 	flusher.Flush()
 	const flushDelay = 200 * time.Millisecond
 	flushTicker := time.NewTicker(flushDelay)
+	defer flushTicker.Stop()
 
 	// Stream logs until the connection is closed.
 	for {
--- a/agent/agent_endpoint.go
+++ b/agent/agent_endpoint.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-memdb"
@@ -1202,7 +1203,8 @@ func (s *HTTPHandlers) AgentMonitor(resp http.ResponseWriter, req *http.Request)
 	// a gzip stream it will go ahead and write out the HTTP response header
 	resp.Write([]byte(""))
 	flusher.Flush()
-
+	const flushDelay = 200 * time.Millisecond
+	t := time.AfterFunc(flushDelay, flusher.Flush)
 	// Stream logs until the connection is closed.
 	for {
 		select {
@@ -1215,6 +1217,11 @@ func (s *HTTPHandlers) AgentMonitor(resp http.ResponseWriter, req *http.Request)
 			return nil, nil
 		case log := <-logsCh:
 			fmt.Fprint(resp, string(log))
+			select {
+			case <-t.C:
+				t = time.AfterFunc(flushDelay, flusher.Flush)
+			default:
+			}
 		}
 	}
 }
--- a/agent/agent_endpoint.go
+++ b/agent/agent_endpoint.go
@@ -1204,7 +1204,8 @@ func (s *HTTPHandlers) AgentMonitor(resp http.ResponseWriter, req *http.Request)
 	resp.Write([]byte(""))
 	flusher.Flush()
 	const flushDelay = 200 * time.Millisecond
-	t := time.AfterFunc(flushDelay, flusher.Flush)
+	flushTicker := time.NewTicker(flushDelay)
+
 	// Stream logs until the connection is closed.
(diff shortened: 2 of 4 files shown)


agent/agent_endpoint.go
```diff
@@ -1211,10 +1213,15 @@ func (s *HTTPHandlers) AgentMonitor(resp http.ResponseWriter, req *http.Request)
...
 				s.agent.logger.Warn("Dropped logs during monitor request", "dropped_count", droppedCount)
 			}
+			flusher.Flush()
 			return nil, nil
 		case log := <-logsCh:
 			fmt.Fprint(resp, string(log))
-			flusher.Flush()
+			select {
+			case <-t.C:
+				t = time.AfterFunc(flushDelay, flusher.Flush)
```
dnephin, line 1222: `AfterFunc` calls `flusher.Flush` in a separate goroutine, which I'm not sure is necessarily safe. It also creates an unnecessary channel each time.

What do you think about using just a `time.Time` here?

```go
lastFlush := time.Now()

...

if time.Since(lastFlush) > flushDelay {
    flusher.Flush()
    lastFlush = time.Now()
}


```

dnephin (commented):
Thanks! I think this is going to be a good improvement. Left some suggestions below.

I also had a question about 1eeddf7ad801cb8f55cd088c275e1a59bd126209. Generally it is expected that once something is closed or Stopped it can't be started again. Was that behaviour causing a problem? Maybe we could document that a monitor is only safe for one use, instead of changing that behaviour?

dhiaayachi (commented):
> Thanks! I think this is going to be a good improvement. Left some suggestions below.
> 
> I also had a question about [1eeddf7](https://github.com/hashicorp/consul/commit/1eeddf7ad801cb8f55cd088c275e1a59bd126209). Generally it is expected that once something is closed or Stopped it can't be started again. Was that behaviour causing a problem? Maybe we could document that a monitor is only safe for one use, instead of changing that behaviour?

Yes the issue was that if we stop the monitor we can't start it again. I was testing and I came cross that issue. 
Why do we not expect the monitor to be started multiple times. This is in the server side so it's a plausible scenario to start and stop monitoring multiple times to debug an agent. am I missing something?

agent/agent_endpoint.go
dnephin: That is true, it does mean that when logs are infrequent we wouldn't get the last few logs until the monitor is closed. Since this is a debugging interface maybe that is ok? Generally a monitor is only going to be run for a few seconds or minutes. So even if we don't write logs often, when the user closes the monitor they should get the last few logs, right?

.changelog/10368.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+monitoring: optimize the monitoring endpoint to avoid losing logs when under high load. [GH-10368](https://github.com/hashicorp/consul/pull/10368)
```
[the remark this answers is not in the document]
dnephin: A tool that finds the right PR number by using the github API and creates the right filename and prompts for the type would be pretty cool!

logging/monitor/monitor.go
```diff
@@ -101,7 +103,7 @@ func (d *monitor) Start() <-chan []byte {
 			}
 		}
 	}()
-
+	d.logger.RegisterSink(d.sink)
```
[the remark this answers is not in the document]
dnephin: Ya, this is a good find! This should help a lot I think

agent/agent_endpoint.go
dhiaayachi: The challenge is more with the tests, we will need to rework a lot of tests for this. As we use the log interceptor to check if we got the logs we expect and once we stop the monitor the interceptor is not anymore available to check the remaining of the logs.

.changelog/10368.txt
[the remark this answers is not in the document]
dhiaayachi: yes that's exactly what I did :) I will be publishing it soon. I just wanted to see if go-changelog would accept it as part of the project or I need to publish it separately

agent/agent_endpoint.go
dhiaayachi: Added complexity is that I noticed the api tests fail on the pipeline but not locally when we miss/remove the flush. I think because locally we use localhost as interface so the flush is implicit.

dhiaayachi: I changed it to use a ticker, it do not avoid using extra resources but it solve the race issue. IMO we should avoid having data not flushed until we close the connection. The user experience is not ideal.

logging/monitor/monitor.go
```diff
@@ -35,7 +35,7 @@ type monitor struct {
 	droppedCount int
 
 	// doneCh coordinates the shutdown of logCh
-	doneCh chan struct{}
+	doneCh chan bool
```
dnephin, line 38: Your question over in https://github.com/hashicorp/consul/pull/9285#discussion_r650157725 made me think of this. I think this is actually not safe for the reason I outline there.

This needs to be a broadcast, because multiple goroutines need to be informed of the change (both the `Start` goroutine, and   whatever is calling `Write`). 

If we need to change a bunch of tests to account for the flushing change let me know, maybe I can help with that.

agent/agent_endpoint.go
```diff
@@ -1211,9 +1214,13 @@ func (s *HTTPHandlers) AgentMonitor(resp http.ResponseWriter, req *http.Request)
...
 				s.agent.logger.Warn("Dropped logs during monitor request", "dropped_count", droppedCount)
 			}
+			flusher.Flush()
 			return nil, nil
+
 		case log := <-logsCh:
 			fmt.Fprint(resp, string(log))
+
+		case <-flushTicker.C:
 			flusher.Flush()
```
dnephin, line 1224: Nice! I think I've seen this pattern work well before. This should work well to periodically flush.

agent/agent_endpoint.go
```diff
@@ -1202,6 +1203,8 @@ func (s *HTTPHandlers) AgentMonitor(resp http.ResponseWriter, req *http.Request)
 	// a gzip stream it will go ahead and write out the HTTP response header
 	resp.Write([]byte(""))
 	flusher.Flush()
+	const flushDelay = 200 * time.Millisecond
+	flushTicker := time.NewTicker(flushDelay)
```
dnephin, line 1207: The docs for NewTicker say:

> Stop the ticker to release associated resources.

so I think we need to do this to prevent leaking some things on each monitor:

```suggestion
	const flushDelay = 200 * time.Millisecond
	flushTicker := time.NewTicker(flushDelay)
	defer flushTicker.Stop()
```

logging/monitor/monitor.go
dhiaayachi: I reverted the change, anyway after reading the code closely it's not meant to be concurrent (we create a new monitor in the handler for each request, reusing the same monitor will not work on both start and stop)

logging/monitor/monitor.go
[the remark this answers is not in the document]
dnephin: I guess we still register the sink without knowing if the goroutine has started. We could create a `chStarted` and close it from the goroutine, and block on it from the main goroutine, to further delay the `RegisterSink` and be confident that the goroutine has started. 

Would also be fine as a follow up.

dnephin (approved):
LGTM! One idea for a small improvement, but not blocking

dhiaayachi:
@dnephin I added the suggested fix but went with a `WaitGroup` as it seems appropriate for this use case, any thought?

dnephin:
That works. I think it could potentially be confusing the the reader since generally wait groups are used for goroutines to end (not start), but functionally I think it's going to do the right thing.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/386958.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/386967.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit c8ba2d40fd8d6915c3e469bfa5bb252f7ab31b70 onto `release/1.10.x` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit c8ba2d40fd8d6915c3e469bfa5bb252f7ab31b70 onto `release/1.9.x` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit c8ba2d40fd8d6915c3e469bfa5bb252f7ab31b70 onto `release/1.8.x` succeeded!

PR #10391, opened 2021-06-14
Explicit upstreams in other dcs shouldn't be deleted

ndhanushkodi:
when they're not in the seen services list of explicit and implicit
upstreams within a datacenter.

Co-authored-by: Freddy Vallenilla <freddy@hashicorp.com>

* pulled out the fmt.Sprintf("%s?dc=dc2", n.String()) into a function that may not be in the right place and doesn't quite have the right interpolation since it hardcodes dc2.
* ran acceptance tests, and just the namespace ones failed since I only created an OSS image.


agent/structs/structs_oss.go
```diff
@@ -126,6 +127,10 @@ func (n ServiceName) String() string {
 	return n.Name
 }
 
+func (n ServiceName) DiscoveryChainID() string {
+	return fmt.Sprintf("%s?dc=dc2", n.String())
+}
```
dnephin, line 132: This appears to only be used by tests, so would probably be better as a function in `state_test.go` instead of a method on `ServiceName`.

dnephin (commented):
Thank you for the fix!

It looks like CI did not run on this PR, I think you may need to re-authenticate with CircleCI to have them run.

agent/structs/structs_oss.go
ndhanushkodi: Thanks @dnephin! I've moved it to the test file and left the hardcoding of the dc and named the function appropriately since it's the only case the function is used for. Let me know if you'd prefer a different style.

freddygv (commented):
Thank you!

freddygv (approved):

.changelog/10391.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+proxycfg: Explicit upstreams in other datacenters shouldn't be deleted when transparent proxy is enabled, and when they're not in the seen services list of explicit and implicit upstreams within a datacenter.
```
dnephin, line 2: not blocking, but I wonder if this might be too much detail for a user reading the changelog. They might not have enough context to understand what this means.

How did this bug manifest to users?

I'm not sure I understand this change well enough to suggest an improvement, but maybe something roughly like: "watch upstreams from remote datacenters to ..."

My assumption is that the "do not delete" means that we continue to watch them for changes?

dnephin (approved):
Nice! I'm not familiar enough with tproxy to understand the underlying bug, but I trust you and Freddy do.

The code change LGTM, but one suggestion for the changelog entry (not blocking a merge).

.changelog/10391.txt
freddygv: That assumption is correct.

Will update the message:
```suggestion
proxycfg: Ensure that endpoints for explicit upstreams in other datacenters are watched in transparent mode.
```

ndhanushkodi: Thank you that looks good!

dnephin (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/387160.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit b8b44419a0a9be639efb9126223badea05674330 onto `release/1.10.x` succeeded!

PR #10394, opened 2021-06-14
Relax validation for expose.paths config

freddygv:
Previously we would return an error if duplicate paths were specified.
This could lead to problems in cases where a user has the same path, 
say /healthz, on two different ports.

This validation was added to signal a potential misconfiguration.
Instead we will only check for duplicate listener ports, since that is
what would lead to ambiguity issues when generating xDS config.

In the future we could look into using a single listener and creating
distinct filter chains for each path/port.

### DIFF
--- a/agent/structs/structs_test.go
+++ b/agent/structs/structs_test.go
@@ -589,13 +589,6 @@ func TestStructs_NodeService_ValidateExposeConfig(t *testing.T) {
 			func(x *NodeService) { x.Proxy.Expose.Paths[0].ListenerPort = 65536 },
 			"invalid listener port",
 		},
-		"duplicate paths": {
-			func(x *NodeService) {
-				x.Proxy.Expose.Paths[0].Path = "/metrics"
-				x.Proxy.Expose.Paths[1].Path = "/metrics"
-			},
-			"duplicate paths exposed",
-		},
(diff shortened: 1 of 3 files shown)


agent/structs/structs_test.go
```diff
@@ -589,13 +589,6 @@ func TestStructs_NodeService_ValidateExposeConfig(t *testing.T) {
 			func(x *NodeService) { x.Proxy.Expose.Paths[0].ListenerPort = 65536 },
 			"invalid listener port",
 		},
-		"duplicate paths": {
```
rboyer, line 592: Wouldn't you rather keep the test case and change it to be a "this is explicitly allowed" style test instead?

freddygv: Yep, that would reflect the policy better. Updated.

rboyer (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/385952.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 33bd9b5be800b8a07c692a178da5101df26cabc0 onto `release/1.10.x` succeeded!

Issue #10320, opened 2021-05-31
debug: improvements to the debug command


dnephin:
This issue is to track a number of improvements to the data produced by [consul debug](https://www.consul.io/commands/debug). These improvements should make it easier to get useful data from `consul debug`.

* [ ] only capture a single cpu profile and single `trace` for the entire duration, instead of a separate one for each interval. A single cpu profile and trace should contain all the same data, and is easier to consume than 4 or 5 separate profiles.
* [ ] capture a single delta `heap`, and `goroutine` profile, instead of a separate one for each interval
* [ ] change the metrics endpoint to return a stream of metrics each time the window ends, instead of having to poll for metrics, which results in most metrics being missed.
* [ ] only capture logs once, instead of once per interval
* [ ] add tests for the log capture to show that it properly captures all logs
* [ ] rename `cluster.json` to `members.json` to better match the name used by the cli (`consul members`)

banks:
> capture a single delta heap, and goroutine profile, instead of a separate one for each interval

I'm interested in the thought process behind this one Daniel - in some cases seeing a few different point-in-time snapshots of these can help e.g. to see which object allocations are growing fastest from one time to the next. Or how the number of goroutines in a specific location is evolving over time. At any rate they are cheap so it seems like we don't loose much by having them there but maybe I'm missing some benefit of removing them?

The rest of these sound like solid improvements!

dnephin:
In general, my thinking is that having multiple copies of the same profile is a distraction and makes it harder to use the debug data. As much as possible we should attempt to move to a debug archive that has a single file of each type. I believe this proposal, once complete, would accomplish this goal. More details on the specifics below.

The heap profile includes 4 views:
```
  -inuse_space      Display in-use memory size
  -inuse_objects    Display in-use object counts
  -alloc_space      Display allocated memory size
  -alloc_objects    Display allocated object counts
```

My understanding is that by switching between the alloc and inuse view you can find most problems, without needing additional profiles. If we need additional data points, capturing them with more time in-between (ex: 10 minutes or 30 minutes) would probably provide a clearer signal, so even when we do need multiple snapshots, capturing separate debug dumps would be a better option than multiple snapshots that are only 30 seconds apart.

For the goroutine profile I'm not sure what we expect to see in the different snapshots. If we are worried about missing something in a single snapshot, then another option is to capture multiple snapshots, and then merge them into a single one in the archive.

PR #10399, opened 2021-06-14
debug: use the new metrics stream in debug command

dnephin:
Related to #10320

Tested by running a dev agent and collecting a `consul debug` archive. All the metrics are indent formatted in the `metrics.json`.

TODO:
* [x] changelog
* [x] manual test to see what the metrics look like in the debug archive
* [ ] add a test for the API endpoint
* [ ] verify and improve the assertions in the debug CLI test for the metrics json file.

### DIFF
--- a/agent/agent_endpoint.go
+++ b/agent/agent_endpoint.go
@@ -1,6 +1,7 @@
 package agent
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"strconv"
@@ -164,6 +165,54 @@ func (s *HTTPHandlers) AgentMetrics(resp http.ResponseWriter, req *http.Request)
 		return nil, nil
 	}
 	return s.agent.baseDeps.MetricsHandler.DisplayMetrics(resp, req)
+}
+
+func (s *HTTPHandlers) AgentMetricsStream(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
+	// Fetch the ACL token, if any, and enforce agent policy.
+	var token string
+	s.parseToken(req, &token)
+	rule, err := s.agent.delegate.ResolveTokenAndDefaultMeta(token, nil, nil)
+	if err != nil {
+		return nil, err
+	}
+	if rule != nil && rule.AgentRead(s.agent.config.NodeName, nil) != acl.Allow {
+		return nil, acl.ErrPermissionDenied
+	}
+
+	flusher, ok := resp.(http.Flusher)
+	if !ok {
+		return nil, fmt.Errorf("Streaming not supported")
+	}
+
+	resp.WriteHeader(http.StatusOK)
+
+	// 0 byte write is needed before the Flush call so that if we are using
+	// a gzip stream it will go ahead and write out the HTTP response header
+	resp.Write([]byte(""))
+	flusher.Flush()
+
+	enc := metricsEncoder{
+		logger:  s.agent.logger,
+		encoder: json.NewEncoder(resp),
+		flusher: flusher,
+	}
+	s.agent.baseDeps.MetricsHandler.Stream(req.Context(), enc)
+	return nil, nil
+}
+
+type metricsEncoder struct {
+	logger  hclog.Logger
+	encoder *json.Encoder
+	flusher http.Flusher
+}
+
+func (m metricsEncoder) Encode(summary interface{}) error {
+	if err := m.encoder.Encode(summary); err != nil {
+		m.logger.Error("failed to encode metrics summary", "error", err)
+		return err
+	}
+	m.flusher.Flush()
+	return nil
 }
 
 func (s *HTTPHandlers) AgentReload(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
--- a/agent/agent_endpoint.go
+++ b/agent/agent_endpoint.go
@@ -172,10 +172,10 @@ func (s *HTTPHandlers) AgentMetricsStream(resp http.ResponseWriter, req *http.Re
 	var token string
 	s.parseToken(req, &token)
 	rule, err := s.agent.delegate.ResolveTokenAndDefaultMeta(token, nil, nil)
-	if err != nil {
+	switch {
+	case err != nil:
 		return nil, err
-	}
-	if rule != nil && rule.AgentRead(s.agent.config.NodeName, nil) != acl.Allow {
+	case rule != nil && rule.AgentRead(s.agent.config.NodeName, nil) != acl.Allow:
 		return nil, acl.ErrPermissionDenied
 	}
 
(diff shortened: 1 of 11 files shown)


agent/agent_endpoint.go
```diff
@@ -158,6 +159,54 @@ func (s *HTTPHandlers) AgentMetrics(resp http.ResponseWriter, req *http.Request)
...
+}
+
+type metricsEncoder struct {
+	logger  hclog.Logger
+	encoder *json.Encoder
+	flusher http.Flusher
+}
+
+func (m metricsEncoder) Encode(summary interface{}) error {
+	if err := m.encoder.Encode(summary); err != nil {
```
dhiaayachi, line 202: If I understand right, the encoder is json encoding but later we are decoding the json and re encoding it, I think we should pass the file encoder directly to the stream or use a byte encoder here and encode the json only once later

command/debug/debug.go
```diff
@@ -515,20 +515,33 @@ func (c *cmd) captureLogs(timestampDir string) error {
...
-	return err
+	decoder := json.NewDecoder(stream)
+	encoder := json.NewEncoder(fh)
+	// TODO: is More() correct here?
+	for decoder.More() {
+		var raw interface{}
+		if err := decoder.Decode(&raw); err != nil {
+			return fmt.Errorf("failed to decode metrics: %w", err)
+		}
+		if err := encoder.Encode(raw); err != nil {
```
dhiaayachi, line 540: There is an open issue [7872](https://github.com/golang/go/issues/7872) about the json encoder buffering the whole file and not flushing until close time. That mean that we will not flush to disk before close, not sure if it's an issue in case of app crash debug

command/debug/debug.go
```diff
@@ -433,12 +425,20 @@ func (c *cmd) captureLongRunning() error {
 			return c.captureTrace(s, timestampDir)
 		})
 	}
-	// Capture logs
 	if c.configuredTarget("logs") {
 		g.Go(func() error {
 			return c.captureLogs(timestampDir)
```
dhiaayachi, line 430: We should definitely do something similar to logs it will be much cleaner, but probably in a separate PR

agent/agent_endpoint.go
```diff
@@ -158,6 +159,54 @@ func (s *HTTPHandlers) AgentMetrics(resp http.ResponseWriter, req *http.Request)
...
+		m.logger.Error("failed to encode metrics summary", "error", err)
+		return err
+	}
+	m.flusher.Flush()
```
dhiaayachi, line 206: The flush here is for each metric data we receive, I don't know how much this is memory consuming on the json encoder side

dhiaayachi (commented):
I saw that this touch the debug part and I took a look, I really like how the stream API work. It make it easier to handle long running requests. As commented below I think we should also change the log API to work the same way. 
I also had few comments/questions

agent/agent_endpoint.go
dnephin: Ya, we could write it directly to the file with an `io.Copy`. I did it this way to match what we were doing before.

I think if we marshal it here with indent, then we don't need to unmarshal/marshal it again in the CLI, we can write to disk. I think this one does need to be JSON though, in case someone else is using the API directly and not using the CLI.

command/debug/debug.go
dnephin: In this case I think it is safe because the API returns multiple JSON objects. I believe the encoder only buffers until it reaches an object, so this will write each one separately. Each call to `Encode` will reach a full JSON object, and write it to the file, then the loop will continue to wait for the next full JSON object.

This can be replaced with an `io.Copy` though, if we  make the other change.

agent/agent_endpoint.go
dnephin: The PR that added this is: https://github.com/armon/go-metrics/pull/125

The intervals end every 10s, and it writes a full summary, which is a pretty large object. I think flushing once per summary is about right.

agent/agent_endpoint.go
dhiaayachi: yes right I forgot that those are public APIs

dnephin:
I think the TODOs are resolved, and this should be ready for review

agent/agent_endpoint.go
```diff
@@ -166,6 +167,55 @@ func (s *HTTPHandlers) AgentMetrics(resp http.ResponseWriter, req *http.Request)
...
 
+func (s *HTTPHandlers) AgentMetricsStream(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
+	// Fetch the ACL token, if any, and enforce agent policy.
+	var token string
+	s.parseToken(req, &token)
+	rule, err := s.agent.delegate.ResolveTokenAndDefaultMeta(token, nil, nil)
+	if err != nil {
+		return nil, err
+	}
+	if rule != nil && rule.AgentRead(s.agent.config.NodeName, nil) != acl.Allow {
```
eculver, line 178: If the `rule` here is actually `nil`, the ACL check is skipped. Is there any case where this could actually happen? Would it make sense to add the `nil` check separately?

eculver (commented):
Looks good. I just had one question about an edge-case. Also, can we add a couple tests for this new endpoint?

agent/agent_endpoint.go
dnephin: Excellent question! This `rule != nil` has given me pause as well, and it's something I'd like to change. But I believe it is correct given the way the code works today. I think basically everywhere in the code base that checks ACLs does this very same check (`git grep 'rule != nil'` returns quite a few lines that look exactly like this).

The reason can be seen here: https://github.com/hashicorp/consul/blob/v1.10.0/agent/consul/acl.go#L1187-L1190

When ACLs are disabled we return a nil `acl.Authorizer` (`rule` in this code is an authorizer, I believe the name comes from the old ACL system, and we probably just left the variable names alone when we updated the ACL system).

I think it would be great to change `ResolveToken` to have it return the [AllowAll](https://github.com/hashicorp/consul/blob/v1.10.0/acl/static_authorizer.go#L229-L232) `Authorizer` instead of `nil`. That would remove the need for a `nil` check and I think would make the code less scary to read.

I think that's a change we could make pretty easily in another PR, and see what others think.

dnephin:
> can we add a couple tests for this new endpoint?

Ya, good idea. The change itself is covered by the tests for the CLI, although I could probably improve the checks in that test as well. The underlying behaviour is tested in the `go-metrics` library, but a test of the Consul integration wouldn't hurt.

dnephin:
Ok, I've rebased this PR and added a test for the HTTP endpoint, and improved the test coverage of the CLI.

I think this is ready for another review.

eculver (approved):
software of the highest quality.. plus tests!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/416873.

PR #10400, opened 2021-06-14
api: properly close the response body

dnephin:
https://golang.org/pkg/net/http/#Client.Do says this:

> If the Body is not both read to EOF and closed, the Client's underlying RoundTripper (typically Transport) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request.

In some places we use `json.Decoder.Decode` to read the response body, but the decoder does not necessarily read all bytes. If there is any trailing whitespace (or other characters) the response body may not be fully read, preventing the client from re-using the TCP connection.

By reading and discarding all the bytes we ensure that the connection can be re-used.

I've seen a similar patch applied in a few other code bases to ensure that tcp connections can be re-used.

### DIFF
--- a/api/acl.go
+++ b/api/acl.go
@@ -449,7 +449,7 @@ func (a *ACL) Bootstrap() (*ACLToken, *WriteMeta, error) {
 	if err != nil {
 		return nil, nil, err
 	}
-	defer resp.Body.Close()
+	defer closeResponseBody(resp)
 
 	wm := &WriteMeta{RequestTime: rtt}
 	var out ACLToken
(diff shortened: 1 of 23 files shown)


freddygv (approved):
LGTM, nice find.

Related Go issue I found about this: https://github.com/golang/go/issues/5645

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/386864.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 2f1f98088e63b00e1932da91a6489d12002f3a39 onto `release/1.10.x` succeeded!

PR #10401, opened 2021-06-15
return an empty record when asked for an addr dns with type other then A, AAAA and ANY

### LINKED ISSUE #10393: Consul Incorrectly returning A record responses for other record types  <hex-ip>.addr.consul.
When filing a bug, please include the following headings if possible. Any example text in this template can be deleted.

#### Overview of the Issue

While querying consul for an `<hex-ip>.addr.consul.` record but specifying any particular record type except for `SOA` and `NS`, an `A` record is returned when an empty answer with `NOERROR` is expected.

#### Reproduction Steps

1. Create a cluster with n client nodes n and n server nodes
1. Run `dig -t MX +nocmd 0b040cf9.addr.consul @127.0.0.1 -p 8600`
1. Repeat for TXT, AAAA, MX, etc 
1. Same results received when querying directly to servers.

Received Result:
```;dig -t MX +nocmd 0b040cf9.addr.consul @127.0.0.1 -p 8600
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 22838
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;0b040cf9.addr.consul.          IN      MX

;; ANSWER SECTION:
0b040cf9.addr.consul.   60      IN      A       11.4.12.249

;; Query time: 0 msec
;; SERVER: 127.0.0.1#8600(127.0.0.1)
;; WHEN: Mon Jun 14 13:21:54 MDT 2021
;; MSG SIZE  rcvd: 65
```

Expected result:
```;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 13343
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;0b040cf9.addr.consul.          IN      MX

;; AUTHORITY SECTION:
consul.                 1       IN      SOA     ns.consul. hostmaster.consul. 1623698585 3600 600 86400 1

;; Query time: 2 msec
;; SERVER: 127.0.0.1#8600(127.0.0.1)
;; WHEN: Mon Jun 14 13:23:05 MDT 2021
;; MSG SIZE  rcvd: 107

```

<details>
  <summary>Client info</summary>

```
agent:
        check_monitors = 4
        check_ttls = 0
        checks = 7
        services = 5
build:
        prerelease =
        revision = 3c1c2267
        version = 1.9.5
consul:
        acl = disabled
        known_servers = 5
        server = false
runtime:
        arch = amd64
        cpu_count = 4
        goroutines = 58
        max_procs = 4
        os = linux
        version = go1.15.8
serf_lan:
        coordinate_resets = 0
        encrypted = true
        event_queue = 0
        event_time = 2414
        failed = 0
        health_score = 0
        intent_queue = 0
        left = 0
        member_time = 46327
        members = 117
        query_queue = 0
        query_time = 1
```

</details>

<details>
  <summary>Server info</summary>

```
agent:
        check_monitors = 0
        check_ttls = 0
        checks = 0
        services = 0
build:
        prerelease =
        revision = 3c1c2267
        version = 1.9.5
consul:
        acl = enabled
        bootstrap = false
        known_datacenters = 3
        leader = true
        leader_addr = 10.3.103.17:8300
        server = true
raft:
        applied_index = 298115613
        commit_index = 298115613
        fsm_pending = 0
        last_contact = 0
        last_log_index = 298115613
        last_log_term = 1958
        last_snapshot_index = 298115609
        last_snapshot_term = 1958
        latest_configuration = [{Suffrage:Voter ID:ba057953-7829-686d-0963-1dcc66705798 Address:10.3.103.10:8300} {Suffrage:Voter ID:d12cba51-23c7-8803-c18f-cff55913ae9f Address:10.3.103.17:8300} {Suffrage:Voter ID:62d21446-d971-8f86-afe8-0b2017d151e2 Address:10.3.103.19:8300} {Suffrage:Voter ID:b481027d-84f8-700c-6f50-264af0805701 Address:10.3.103.12:8300} {Suffrage:Voter ID:564d2409-1975-044d-7a0b-27b80f619c97 Address:10.3.103.14:8300}]
        latest_configuration_index = 0
        num_peers = 4
        protocol_version = 3
        protocol_version_max = 3
        protocol_version_min = 0
        snapshot_version_max = 1
        snapshot_version_min = 0
        state = Leader
        term = 1958
runtime:
        arch = amd64
        cpu_count = 4
        goroutines = 665
        max_procs = 4
        os = linux
        version = go1.15.8
serf_lan:
        coordinate_resets = 0
        encrypted = true
        event_queue = 0
        event_time = 2414
        failed = 0
        health_score = 0
        intent_queue = 0
        left = 0
        member_time = 46327
        members = 117
        query_queue = 0
        query_time = 1
serf_wan:
        coordinate_resets = 0
        encrypted = true
        event_queue = 0
        event_time = 1
        failed = 0
        health_score = 0
        intent_queue = 0
        left = 0
        member_time = 6892
        members = 15
        query_queue = 0
        query_time = 1

```

</details>

servers running `CentOS Linux release 7.9.2009 (Core)` and using dig version: `DiG 9.11.4-P2-RedHat-9.11.4-26.P2.el7_9.5`

### Log Fragments

Include appropriate Client or Server log fragments. If the log is longer than a few dozen lines, please include the URL to the [gist](https://gist.github.com/) of the log instead of posting it in the issue. Use `-log-level=TRACE` on the client and server to capture the maximum log detail.

dhiaayachi:
This fix #10393
 
this is a bug for sure because we return a different record type then the one requested. That said we need to verify what is the intended behaviour:

- [x] is returning `A` or `AAAA` in case the origin query part is `IPv4` or `IPv6` respectively enough? ---> yes
- [x] What we do about `ANY` queries? ---> we should return an A or AAAA record
- [x] do this break any existing behaviour or users call flow? ---> no

### DIFF
--- a/.changelog/10401.txt
+++ b/.changelog/10401.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+return an invalid record when asked for an addr dns with type other then A and AAAA.
+```
--- a/.changelog/10401.txt
+++ b/.changelog/10401.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-dns: return an invalid record when asked for an addr dns with type other then A and AAAA.
+dns: return an empty answer when asked for an addr dns with type other then A and AAAA.
 ```
--- a/.changelog/10401.txt
+++ b/.changelog/10401.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-return an invalid record when asked for an addr dns with type other then A and AAAA.
+dns: return an invalid record when asked for an addr dns with type other then A and AAAA.
 ```
--- a/agent/dns.go
+++ b/agent/dns.go
@@ -783,8 +783,8 @@ func (d *DNSServer) doDispatch(network string, remoteAddr net.Addr, req, resp *d
 	case "addr":
 		// <address>.addr.<suffixes>.<domain> - addr must be the second label, datacenter is optional
 
-		//check if the query type is A for IPv4 or AAAA for IPv6
-		if req.Question[0].Qtype != dns.TypeA && len(queryParts[0])/2 == 4 || req.Question[0].Qtype != dns.TypeAAAA && len(queryParts[0])/2 == 16 {
+		//check if the query type is A for IPv4 or AAAA for IPv6 or ANY
+		if (req.Question[0].Qtype != dns.TypeANY) && (req.Question[0].Qtype != dns.TypeA || len(queryParts[0])/2 != 4) && (req.Question[0].Qtype != dns.TypeAAAA || len(queryParts[0])/2 != 16) {
 			return invalid()
 		}
 		if len(queryParts) != 1 {
--- a/agent/dns.go
+++ b/agent/dns.go
@@ -826,7 +826,6 @@ func (d *DNSServer) doDispatch(network string, remoteAddr net.Addr, req, resp *d
 				AAAA: ip,
 			}
 			if req.Question[0].Qtype != dns.TypeAAAA && req.Question[0].Qtype != dns.TypeANY {
-				//resp.SetRcode(req, dns.RcodeSuccess)
 				resp.Extra = append(resp.Extra, AAAARecord)
 			} else {
 				resp.Answer = append(resp.Answer, AAAARecord)
--- a/agent/dns.go
+++ b/agent/dns.go
@@ -783,7 +783,6 @@ func (d *DNSServer) doDispatch(network string, remoteAddr net.Addr, req, resp *d
 	case "addr":
 		// <address>.addr.<suffixes>.<domain> - addr must be the second label, datacenter is optional
 
-		//check if the query type is A for IPv4 or AAAA for IPv6 or ANY
 		if len(queryParts) != 1 {
 			return invalid()
 		}
@@ -795,6 +794,7 @@ func (d *DNSServer) doDispatch(network string, remoteAddr net.Addr, req, resp *d
 			if err != nil {
 				return invalid()
 			}
+			//check if the query type is  A for IPv4 or ANY
 			if req.Question[0].Qtype != dns.TypeA && req.Question[0].Qtype != dns.TypeANY {
 				resp.SetRcode(req, dns.RcodeSuccess)
 			} else {
@@ -814,6 +814,7 @@ func (d *DNSServer) doDispatch(network string, remoteAddr net.Addr, req, resp *d
 			if err != nil {
 				return invalid()
 			}
+			//check if the query type is  AAAA for IPv6 or ANY
 			if req.Question[0].Qtype != dns.TypeAAAA && req.Question[0].Qtype != dns.TypeANY {
 				resp.SetRcode(req, dns.RcodeSuccess)
 			} else {
--- a/agent/dns.go
+++ b/agent/dns.go
@@ -805,7 +805,6 @@ func (d *DNSServer) doDispatch(network string, remoteAddr net.Addr, req, resp *d
 				A: ip,
 			}
 			if req.Question[0].Qtype != dns.TypeA && req.Question[0].Qtype != dns.TypeANY {
-				//resp.SetRcode(req, dns.RcodeSuccess)
 				resp.Extra = append(resp.Answer, ARecord)
 			} else {
 				resp.Answer = append(resp.Answer, ARecord)
--- a/agent/dns.go
+++ b/agent/dns.go
@@ -796,7 +796,16 @@ func (d *DNSServer) doDispatch(network string, remoteAddr net.Addr, req, resp *d
 			}
 			//check if the query type is  A for IPv4 or ANY
 			if req.Question[0].Qtype != dns.TypeA && req.Question[0].Qtype != dns.TypeANY {
-				resp.SetRcode(req, dns.RcodeSuccess)
+				//resp.SetRcode(req, dns.RcodeSuccess)
+				resp.Extra = append(resp.Answer, &dns.A{
+					Hdr: dns.RR_Header{
+						Name:   qName + d.domain,
+						Rrtype: dns.TypeA,
+						Class:  dns.ClassINET,
+						Ttl:    uint32(cfg.NodeTTL / time.Second),
+					},
+					A: ip,
+				})
 			} else {
 				resp.Answer = append(resp.Answer, &dns.A{
 					Hdr: dns.RR_Header{
@@ -816,7 +825,16 @@ func (d *DNSServer) doDispatch(network string, remoteAddr net.Addr, req, resp *d
 			}
 			//check if the query type is  AAAA for IPv6 or ANY
 			if req.Question[0].Qtype != dns.TypeAAAA && req.Question[0].Qtype != dns.TypeANY {
-				resp.SetRcode(req, dns.RcodeSuccess)
+				//resp.SetRcode(req, dns.RcodeSuccess)
+				resp.Extra = append(resp.Extra, &dns.AAAA{
+					Hdr: dns.RR_Header{
+						Name:   qName + d.domain,
+						Rrtype: dns.TypeAAAA,
+						Class:  dns.ClassINET,
+						Ttl:    uint32(cfg.NodeTTL / time.Second),
+					},
+					AAAA: ip,
+				})
 			} else {
 				resp.Answer = append(resp.Answer, &dns.AAAA{
 					Hdr: dns.RR_Header{
--- a/agent/dns.go
+++ b/agent/dns.go
@@ -795,7 +795,7 @@ func (d *DNSServer) doDispatch(network string, remoteAddr net.Addr, req, resp *d
 				return invalid()
 			}
 			//check if the query type is  A for IPv4 or ANY
-			ARecord := &dns.A{
+			aRecord := &dns.A{
 				Hdr: dns.RR_Header{
 					Name:   qName + d.domain,
 					Rrtype: dns.TypeA,
--- a/agent/dns.go
+++ b/agent/dns.go
@@ -784,9 +784,6 @@ func (d *DNSServer) doDispatch(network string, remoteAddr net.Addr, req, resp *d
 		// <address>.addr.<suffixes>.<domain> - addr must be the second label, datacenter is optional
 
 		//check if the query type is A for IPv4 or AAAA for IPv6 or ANY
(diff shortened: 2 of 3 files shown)


.changelog/10401.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+return an invalid record when asked for an addr dns with type other then A and AAAA.
```
dnephin, line 2:
```suggestion
dns: return an invalid record when asked for an addr dns with type other then A and AAAA.
```

We generally prefix the message with a label of some kind. I don't think we have a good convention for how to choose that label yet, but `dns` is probably a same option

agent/dns.go
```diff
@@ -782,6 +782,11 @@ func (d *DNSServer) doDispatch(network string, remoteAddr net.Addr, req, resp *d
 
 	case "addr":
 		// <address>.addr.<suffixes>.<domain> - addr must be the second label, datacenter is optional
+
+		//check if the query type is A for IPv4 or AAAA for IPv6 or ANY
+		if (req.Question[0].Qtype != dns.TypeANY) && (req.Question[0].Qtype != dns.TypeA || len(queryParts[0])/2 != 4) && (req.Question[0].Qtype != dns.TypeAAAA || len(queryParts[0])/2 != 16) {
+			return invalid()
```
dnephin, line 788: `invalid()` is returning this:
```
RcodeNameError      = 3  // NXDomain  - Non-Existent Domain 
```

I wonder if that is the right response code. The issue reported said they exposed `NoError` (`RcodeSuccess`) with an empty body, but I wonder if `RcodeNotImplemented` would be appropriate.

agent/dns.go
```diff
@@ -782,6 +782,11 @@ func (d *DNSServer) doDispatch(network string, remoteAddr net.Addr, req, resp *d
 
 	case "addr":
 		// <address>.addr.<suffixes>.<domain> - addr must be the second label, datacenter is optional
+
+		//check if the query type is A for IPv4 or AAAA for IPv6 or ANY
+		if (req.Question[0].Qtype != dns.TypeANY) && (req.Question[0].Qtype != dns.TypeA || len(queryParts[0])/2 != 4) && (req.Question[0].Qtype != dns.TypeAAAA || len(queryParts[0])/2 != 16) {
```
dnephin, line 787: A switch/case on `Qtype` might make this a bit easier to read.

Is the `len(queryParts[0]/2)` part necessary, or would that be handled later on by `hex.DecodeString` ? We might want to the `Qtype` check separate from the length check so that we can return different error codes.

If we need to keep this all together it would be good to extract this into a function to avoid the really long line.

dnephin (commented):
Thank you for working on this! 

I think this should fix the reported issue, although I think we have a similar problem for the other queries (`service`, `connect`, `node`, etc). We could address the broader issue separately if you'd prefer to keep this fix focused on just the reported problem, and mitigate any potential breaking changes of the broader change.

I'm not familiar enough with the DNS protocol to be confident the test changes are correct, but they seem reasonable given my understanding.

agent/dns.go
dhiaayachi: No it's not, I need to add logic to check which error to return:
- if the domain is not correct return invalid
- if the type is not implemented, we should return `not implemented`

agent/dns.go
dhiaayachi: I agree, it's not readable. I will check. The reason I added the len check is for the case we have an IPv6 request with an IPv4 len.
Not sure if that's possible though.

agent/dns.go
rmchugh-qualtrics: You should only return a `NXDomain` if there is no valid record at all for that query.  Thus if the `<address>` section would return a valid `A` or `AAAA` record, you should return as listed `NoError` (`RcodeSuccess`) instead.  If you return a NXDomain then resolvers will cache / return that even if a client also asks for a correct / existing record type.

Example if requesting `0b040cf9.addr.consul.` `AAAA` -> `NoError` & empty body.
`0b040cf9.addr.consul.` `A` -> `NoError` & `11.4.12.249` in answer.  If you respond to the `AAAA` request with `NXDomain` and it comes in first, the nameserver will also cache that for the default TTL time in the SOA and thus return `NXDomain` for the `A` request coming in a few ms later.  
If however something like `frog.addr.consul` was asked, you should return a `NXDomain`  (invalid) for every record type as there will never be a valid answer.

agent/dns.go
```diff
@@ -793,32 +794,58 @@ func (d *DNSServer) doDispatch(network string, remoteAddr net.Addr, req, resp *d
...
-					Name:   qName + d.domain,
-					Rrtype: dns.TypeA,
-					Class:  dns.ClassINET,
-					Ttl:    uint32(cfg.NodeTTL / time.Second),
-				},
-				A: ip,
-			})
+			//check if the query type is  A for IPv4 or ANY
+			if req.Question[0].Qtype != dns.TypeA && req.Question[0].Qtype != dns.TypeANY {
+				//resp.SetRcode(req, dns.RcodeSuccess)
```
dnephin, line 799:
```suggestion
```

agent/dns.go
```diff
@@ -793,32 +794,58 @@ func (d *DNSServer) doDispatch(network string, remoteAddr net.Addr, req, resp *d
...
+			} else {
+				resp.Answer = append(resp.Answer, &dns.A{
+					Hdr: dns.RR_Header{
+						Name:   qName + d.domain,
+						Rrtype: dns.TypeA,
+						Class:  dns.ClassINET,
+						Ttl:    uint32(cfg.NodeTTL / time.Second),
+					},
+					A: ip,
+				})
```
dnephin, line 818: These two blocks are identical except for the field on the response where the value is set, is that right?

I think we could structure the code to make that more obvious:

```go

rr := &dns.A{...}
if req.Question[0].Qtype == dns.TypeA || req.Question[0].Qtype == dns.TypeANY {
    resp.Answer = append(resp.Answer, rr)
    return true
}

resp.Extra = append(resp.Extra, rr)
return true
```

(same below)

agent/dns_test.go
```diff
@@ -6044,7 +6044,7 @@ func TestDNS_AddressLookup(t *testing.T) {
 	}
 	for question, answer := range cases {
 		m := new(dns.Msg)
-		m.SetQuestion(question, dns.TypeSRV)
+		m.SetQuestion(question, dns.TypeA)
```
dnephin, line 6047: Leaving this comment for anyone else who may be reading this PR. We believe this test was previously incorrect because it was asking for a SRV record but getting back an A record. This change is to fix that bug and only return the type of record that is requested.

agent/dns_test.go
```diff
@@ -6069,6 +6072,88 @@ func TestDNS_AddressLookup(t *testing.T) {
...
+	}
+	for question, answer := range cases {
+		m := new(dns.Msg)
+		m.SetQuestion(question, dns.TypeANY)
+
+		c := new(dns.Client)
+		in, _, err := c.Exchange(m, a.DNSAddr())
+		if err != nil {
+			t.Fatalf("err: %v", err)
+		}
```
dnephin, line 6097: Not blocking, but I have been converting any tests that use this style over to use `require.{NoError, Equal, ...}`. Not a big deal, but many of these `t.Fatalf` end up not providing enough information to debug. So as we write new tests we should probably prefer using `require` assertions instead.

dnephin (commented):
Nice! Looks good, just one suggestion for the implementation

agent/dns.go
dnephin: This line got back in somehow

agent/dns.go
```diff
@@ -793,32 +794,44 @@ func (d *DNSServer) doDispatch(network string, remoteAddr net.Addr, req, resp *d
...
 					Name:   qName + d.domain,
 					Rrtype: dns.TypeAAAA,
 					Class:  dns.ClassINET,
 					Ttl:    uint32(cfg.NodeTTL / time.Second),
 				},
 				AAAA: ip,
-			})
+			}
+			if req.Question[0].Qtype != dns.TypeAAAA && req.Question[0].Qtype != dns.TypeANY {
+				//resp.SetRcode(req, dns.RcodeSuccess)
```
dnephin, line 830: same

.changelog/10401.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+dns: return an invalid record when asked for an addr dns with type other then A and AAAA.
```
dnephin, line 2:
```suggestion
dns: return an empty answer when asked for an addr dns with type other then A and AAAA.
```

agent/dns.go
```diff
@@ -793,32 +794,44 @@ func (d *DNSServer) doDispatch(network string, remoteAddr net.Addr, req, resp *d
 			if err != nil {
 				return invalid()
 			}
-
-			resp.Answer = append(resp.Answer, &dns.A{
+			//check if the query type is  A for IPv4 or ANY
+			ARecord := &dns.A{
```
dnephin, line 798:
```suggestion
			aRecord := &dns.A{
```

Generally local vars would be lowercase, even when they are acronyms or short forms I think.

agent/dns.go
dnephin, line 830:
```suggestion
```

agent/dns.go
```diff
@@ -793,32 +794,44 @@ func (d *DNSServer) doDispatch(network string, remoteAddr net.Addr, req, resp *d
...
 		// IPv6
 		case 16:
 			ip, err := hex.DecodeString(queryParts[0])
 			if err != nil {
 				return invalid()
 			}
-
-			resp.Answer = append(resp.Answer, &dns.AAAA{
+			//check if the query type is  AAAA for IPv6 or ANY
+			AAAARecord := &dns.AAAA{
```
dnephin, line 820:
```suggestion
			aaaaRecord := &dns.AAAA{
```

dnephin (approved):
LGTM with a few small edits

dnephin (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/397792.

PR #10424, opened 2021-06-17
connect/ca: cease including the common name field in generated certs

rboyer:
As part of this change, we ensure that the SAN extensions are marked as
critical when the subject is empty so that AWS PCA tolerates the loss of
common names well and continues to function as a Connect CA provider.

Parts of this currently hack around a bug in crypto/x509 and can be
removed after https://go-review.googlesource.com/c/go/+/329129 lands in
a Go release.

Note: the AWS PCA tests do not run automatically, but the following
passed locally for me:

    ENABLE_AWS_PCA_TESTS=1 go test ./agent/connect/ca -run TestAWS

The only remaining place that we use a CommonName in certs is for CA certs themselves, since they don't get SAN fields.


agent/connect/ca/provider_aws.go
```diff
@@ -28,10 +29,12 @@ const (
 	// IntermediateTemplateARN is the AWS-defined template we need to use when
 	// issuing an intermediate cert.
 	IntermediateTemplateARN = "arn:aws:acm-pca:::template/SubordinateCACertificate_PathLen0/V1"
+	// IntermediateTemplateARN = "arn:aws:acm-pca:::template/SubordinateCACertificate_PathLen0_CSRPassthrough/V1"
```
rboyer, line 32: TODO: remove these comments

agent/connect/x509_patch.go
```diff
@@ -0,0 +1,122 @@
+package connect
```
rboyer, line 1: This can all go away with https://github.com/golang/go/pull/46810

rboyer:
Parts of this can be cleaned up after https://github.com/golang/go/pull/46810 (or something like it) lands.

agent/connect/ca/provider_consul.go
```diff
@@ -487,8 +471,12 @@ func (c *ConsulProvider) SignIntermediate(csr *x509.CertificateRequest) (string,
 	effectiveNow := time.Now().Add(-1 * CertificateTimeDriftBuffer)
 	template := x509.Certificate{
```
rboyer, line 472: This blob threw me until I found it.

This spot in the code _accepts_ a CSR but does not pass that CSR off to the `crypto/x509` package or anything of the sort. The incoming CSR has select fields plucked out of it to assign to the newly generated intermediate cert.

I done an overmatch during cleanup and stripped the `Subject: csr.Subject,` line from this function which super broke intermediate certs (which only put data in the Subject/CN field). For now I modified this to handle this more like how a CSR should operate and copy the various SAN fields along with the Subject and any provided ExtraExtensions as-is.

We can discuss if the CA certs themselves should be using SAN fields, too, but the CN fields for those things are less awful to construct so it's fine to keep them for now.

agent/connect/x509_patch.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+const (
+	x509_nameTypeEmail = 1
+	x509_nameTypeDNS   = 2
+	x509_nameTypeURI   = 6
+	x509_nameTypeIP    = 7
+)
+
+func HackSANExtensionForCSR(template *x509.CertificateRequest) {
```
rboyer, line 31: This is basically attempting to repeat this blob of code from the stdlib ourselves:
https://github.com/golang/go/blob/0e67ce3d28320e816dd8e7cf7d701c1804fb977e/src/crypto/x509/x509.go#L1088

agent/connect/x509_patch.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	if x509_oidInExtensions(x509_oidExtensionSubjectAltName, template.ExtraExtensions) {
+		return
+	}
+
+	// Ensure we wipe the subject field completely since we're going to use the
+	// SAN field instead.
+	template.RawSubject = nil
+	template.Subject = pkix.Name{}
```
rboyer, line 48: This is my own amendment to the stdlib fix. I feel like at this point it's safe to remove this, but also if we aren't using the Subject fields why include them? Thoughts?

When I switch to the patched go toolchain and disable this entire hack everything works, so maybe i really don't need this.

agent/connect/x509_patch_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+			for _, ext := range csr.Extensions {
+				if ext.Id.Equal(x509_oidExtensionSubjectAltName) {
+					if tc.expectCritical {
+						if !ext.Critical {
+							t.Fatal("SAN extension is not critical")
+						}
+					} else {
+						if ext.Critical {
+							t.Fatal("SAN extension is critical now; maybe we don't need the hack anymore with this version of Go?")
```
rboyer, line 88: Tech debt automated debt collector!

agent/connect/ca/provider_consul.go
picatz: FWIW, if we're not going to use CN everywhere else, it probably makes sense to stop using it for the CA and reflect that in the TLS utilities. Is there a reason not to?

agent/connect/x509_patch.go
```diff
@@ -0,0 +1,122 @@
...
+		// subjectAltName extension ... is marked as critical”
+		//
+		// Since we just cleared the subject above, it's critical.
+		Critical: true,
+		Value:    value,
+	}
+	template.ExtraExtensions = append(template.ExtraExtensions, ext)
+}
+
+// oidNotInExtensions reports whether an extension with the given oid exists in
```
picatz, line 68: Small nit: `oidNotInExtensions` is not the real name of the function anymore.

```suggestion
// x509_oidInExtensions reports whether an extension with the given oid exists in
```

agent/connect/x509_patch.go
```diff
@@ -0,0 +1,122 @@
...
+func x509_oidInExtensions(oid asn1.ObjectIdentifier, extensions []pkix.Extension) bool {
+	for _, e := range extensions {
+		if e.Id.Equal(oid) {
+			return true
+		}
+	}
+	return false
+}
+
+// marshalSANs marshals a list of addresses into a the contents of an X.509
```
picatz, line 79: Small nit: `marshalSANs` is not the real name of the function anymore.

```suggestion
// x509_marshalSANs marshals a list of addresses into a the contents of an X.509
```

agent/connect/x509_patch.go
picatz: Should there be a comment for this function that includes the link as well?

agent/connect/testing_ca.go
```diff
@@ -84,6 +85,7 @@ func testCA(t testing.T, xc *structs.CARoot, keyType string, keyBits int, ttl ti
 	} else {
 		after = after.AddDate(10, 0, 0)
 	}
+	// TODO: should this have a SUBJECT?
```
rboyer, line 88: TODO

tlsutil/generate.go
```diff
@@ -107,7 +108,8 @@ func GenerateCA(opts CAOpts) (string, string, error) {
 			Locality:      []string{"San Francisco"},
 			StreetAddress: []string{"101 Second Street"},
 			Organization:  []string{"HashiCorp Inc."},
-			CommonName:    name,
+			// TODO(rb): should we stop inserting CommonNames here as well?
```
rboyer, line 111: TODO

agent/connect/x509_patch.go
rboyer: Removed

agent/connect/ca/provider_consul.go
rboyer: AWS PCA requires a CN field to be populated in order to create the Root CA.

agent/connect/x509_patch.go
rboyer: :+1:

agent/connect/ca/provider_consul.go
picatz: Interesting! Seems to align with that I'm finding in https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.6 and https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.4 where the Subject/CN is required.

.changelog/10424.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+connect/ca: cease including the common name field in generated certs
```
picatz, line 2: Might be worth clarifying somehow that generated CA certs still require/include a common name? Not totally happy with the wording, but maybe something like:
```suggestion
connect/ca: cease including the common name field in generated x509 certificates when not used as a CA
```

picatz (approved):
Small suggestion on release note, but otherwise LGTM! 👍

agent/connect/ca/provider_consul.go
```diff
@@ -447,6 +429,8 @@ func (c *ConsulProvider) Sign(csr *x509.CertificateRequest) (string, error) {
 // are met. It should return a signed CA certificate with a path length constraint
 // of 0 to ensure that the certificate cannot be used to generate further CA certs.
 func (c *ConsulProvider) SignIntermediate(csr *x509.CertificateRequest) (string, error) {
+	connect.HackSANExtensionForCSR(csr)
```
freddygv, line 432: Do we need this hack if the CommonName isn't being removed from intermediates?

agent/connect/x509_patch.go
```diff
@@ -0,0 +1,126 @@
...
+	switch {
+	case len(template.DNSNames) > 0:
+	case len(template.EmailAddresses) > 0:
+	case len(template.IPAddresses) > 0:
+	case len(template.URIs) > 0:
+	default:
+		return
+	}
+
+	if x509_oidInExtensions(x509_oidExtensionSubjectAltName, template.ExtraExtensions) {
```
freddygv, line 50: Should this check that the extension is marked as critical?

agent/connect/ca/provider_consul.go
rboyer: You have a point. I blanket added this before any of the relevant `crypto/x509` library calls in the spirit of something approximating monkeypatching. I also did all of this while trying to figure out why nothing was working after getting fixed (which was an unrelated bug that I later fixed). I did not revisit to see of the intermediates actually had this hack *activate* practically speaking.

rboyer: Tested removing this call from SignIntermediate functions in the providers and everything (even AWS) works.

agent/connect/x509_patch.go
rboyer: This is just doing whatever the stdlib will ultimately do. Basically if you bring your own extension it's assumed that you did it correctly.

freddygv (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/398201.

pierrecdn:
Hi,
We perform a native integration of Connect and noticed this during an upgrade to 1.12 (TL;DR: we used it for the leaf cert storage).
The RFC is clear on the fact that it is valid (if SAN extensions are marked as critical), but at reading

> Parts of this can be cleaned up after https://github.com/golang/go/pull/46810 (or something like it) lands.

I was wondering if you still plan to remove it at some point or not.
Thanks,

PR #10425, opened 2021-06-17
tlsutil: fix a possible panic, and make the package safer

dnephin:

As I have been reading through the `tlsutil` package I noticed a panic and a few other things that were potentially unsafe or made it more difficult to read. This PR fixes the issues.

### DIFF
--- a/tlsutil/config.go
+++ b/tlsutil/config.go
@@ -11,6 +11,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/hashicorp/go-hclog"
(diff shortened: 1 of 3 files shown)


tlsutil/config.go
```diff
@@ -252,7 +255,8 @@ func (c *Configurator) Update(config Config) error {
 	c.manual.cert = cert
 	c.manual.caPems = pems
 	c.caPool = pool
-	c.version++
+	atomic.AddUint64(&c.version, 1)
+	c.log("Update")
```
i0rek, line 259: This changes the behaviour slightly to only log `update` when it didn't error. Which is fine I guess.

i0rek (approved):
LGTM! Left a comment about a tiny change of behaviour.

tlsutil/config.go
```diff
@@ -319,29 +318,30 @@ func (c *Configurator) UpdateAutoTLS(manualCAPems, connectCAPems []string, pub,
...
-	c.version++
+	atomic.AddUint64(&c.version, 1)
+	c.log("UpdateAutoTLS")
 	return nil
 }
 
 func (c *Configurator) UpdateAreaPeerDatacenterUseTLS(peerDatacenter string, useTLS bool) {
-	c.Lock()
-	defer c.Unlock()
-	c.version++
```
rboyer, line 329: Contextually I'm guessing you removed this because it probably wasn't supposed to be here in the first place.

rboyer (approved):
LGTM

tlsutil/config.go
dnephin: Oh, hmm, no this might have been a mistake. I had tried some other refactors before restoring the `version` increments to the Update functions. I think I should restore it here.

The version is only used in trace logs, so not super critical, but I guess since this is an update function it would still be good to increment the version.

tlsutil/config.go
dnephin: Correct, I should have mentioned that. Seems like only logging the update in the success case is slightly more correct, although these are trace level logs, so we probably don't see them very often.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/389680.

PR #10431, opened 2021-06-18
Fixing SOA record to use alt domain when alt domain in use

jm4games:
So this PR fixes a bug in how SOA record is generated on a DNS query when using an alt domain. Essentially whenever the alt domain is in use and the SOA record is generated it always assumes the primary domain and never takes into account an alt domain. I'm not sure how critical this is, but it seems like this could be fairly problematic. Maybe people are not using alt domains? Right now I'm merging this to master, but my company really needs the fix on version v1.9.6, not sure what the process is for back porting. My company is using a custom build for now.

### DIFF
--- a/agent/dns.go
+++ b/agent/dns.go
@@ -373,7 +373,7 @@ func (d *DNSServer) handlePtr(resp dns.ResponseWriter, req *dns.Msg) {
 
 	// Only add the SOA if requested
 	if req.Question[0].Qtype == dns.TypeSOA {
-		d.addSOA(cfg, m)
+		d.addSOA(cfg, m, q.Name)
 	}
 
 	datacenter := d.agent.config.Datacenter
@@ -486,7 +486,7 @@ func (d *DNSServer) handleQuery(resp dns.ResponseWriter, req *dns.Msg) {
 	switch req.Question[0].Qtype {
 	case dns.TypeSOA:
 		ns, glue := d.nameservers(cfg, maxRecursionLevelDefault)
-		m.Answer = append(m.Answer, d.soa(cfg))
+		m.Answer = append(m.Answer, d.soa(cfg, q.Name))
 		m.Ns = append(m.Ns, ns...)
 		m.Extra = append(m.Extra, glue...)
 		m.SetRcode(req, dns.RcodeSuccess)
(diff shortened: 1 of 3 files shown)


agent/dns.go
```diff
@@ -505,18 +505,23 @@ func (d *DNSServer) handleQuery(resp dns.ResponseWriter, req *dns.Msg) {
 	}
 }
 
-func (d *DNSServer) soa(cfg *dnsConfig) *dns.SOA {
+func (d *DNSServer) soa(cfg *dnsConfig, questionName string) *dns.SOA {
+	domain := d.domain
+	if strings.HasSuffix(questionName, d.altDomain) {
+		domain = d.altDomain
+	}
```
dhiaayachi, line 512: A check if `d.altDomain` is not empty need to be done before, otherwise this will always be true.
Also it's safer if we do the check with a `.` prefix
```suggestion
	if d.altDomain !="" && strings.HasSuffix(questionName, "." + d.altDomain) {
		domain = d.altDomain
	}
```

dhiaayachi (commented):
Thank you @jm4games for adding this Fix!!
I added a comment about the `altDomain` check. I should be able to approve this as soon as it's resolved.

markan:
This looks good. Can you add a test and a 'bugfix' changelog entry?

jm4games:
Will do, try to get to it by this weekend.

markan:
Hi there! Checking in; @jm4games do you need any help with getting this sorted out?

jm4games:
Sorry getting the tests running was frustrating me and I kind of forgot about this. Let me see if I can give it another go.

jm4games:
@markan this should be good to go now I think.

dhiaayachi (approved):
LGTM!!
Thank you for working on this @jm4games

jm4games:
@dhiaayachi I guess someone else will have to merge this since I don't have write perms.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/464208.

jm4games:
@dhiaayachi thanks!

PR #10432, opened 2021-06-18
Hold onto the token store from the base deps in the Client

mkeeler:
This is needed in order to access the agent token in some enterprise code.

PR #10433, opened 2021-06-18
[1.8.x] Backport #10432

mkeeler:
This backports #10432. There was a conflict with because the `Client` in 1.8.x doesn't have the agent `BaseDeps` passed into it but rather has the token store provided in its options.


rboyer (approved):
LGTM

PR #10434, opened 2021-06-18
docs: updated FAQ to accommodate new requirements prior to upgrading

trujillo-adam:
This PR was to update the FAQ to include information about the upgrade requirements for 1.10.0: customers on Consul Enterprise 1.8+ or 1.9+ must upgrade Consul agents to the latest patch available (1.8.13 or 1.9.7) prior to upgrading the Consul server instances to 1.10.0.

### DIFF
--- a/website/content/docs/enterprise/license/faq.mdx
+++ b/website/content/docs/enterprise/license/faq.mdx
@@ -20,6 +20,11 @@ This check is part of the server boot-up process.
 In previous versions of HashiCorp enterprise products, one server could distribute a license to other servers via the Raft protocol.
 This will no longer work since each server must be able to find a valid license during the startup process.
 
+## Q: What are the upgrade requirements?
+
+If ACLs are enabled, customers on Consul Enterprise 1.8+ or 1.9+ must upgrade Consul agents to the latest patch available (1.8.13 or 1.9.7) prior to upgrading the Consul server instances to 1.10.0.
+Otherwise, the Consul agents will fail to retrieve the license with a valid agent token.
+
 ## Q: Is there a tutorial available for the license configuration steps?
 
 Please visit the [Enterprise License Tutorial](https://learn.hashicorp.com/tutorials/nomad/hashicorp-enterprise-license?in=consul/enterprise).
@@ -49,7 +54,7 @@ The in-storage license feature will not be supported starting with Consul Entepr
 
 ## Q: What is the impact on evaluation licenses due to this change?
 
-The 6-hour trial period for evaluation licenses will be deprecated as of Consul Enterprise 1.10.0.  
+The 6-hour trial period for evaluation licenses will be deprecated as of Consul Enterprise 1.10.0.
 This means that any clusters deployed with Consul 1.10.0+ent binaries will need to have a valid license on the disk (auto-loaded) or as an environment variable.
 Failure to provide a valid license key will result in the Consul server agent not starting.
 
@@ -61,13 +66,14 @@ As Consul Enterprise approaches the expiration date, warnings will be issued in 
 
 ## Q: Does this affect client agents?
 
-For existing clusters, if the clients agents are using ACLs and have a valid token, then they will be able to retrieve the license from the server.
+For existing clusters, if the clients agents are using ACLs and have a valid token, then they will be able to retrieve the license from the server if the upgrade requirements have been met. Please see the [upgrade requirements](faq#q-what-are-the-upgrade-requirements).
+
 If the client agents are not using ACLs, then the client agents will be need to have the license on-disk (auto-loading) or as an environment variable.
 For new Consul clusters using Consul 1.10.0+ent, customers must ensure that there is a valid license on-disk (auto-loaded) or as an environment variable.
 
 ## Q: Does this affect snapshot agents?
 
-Same behavior as Consul clients. See answer for [Does this affect snapshot agents? ](faq#q-does-this-affect-client-agents)
+Same behavior as Consul clients. See answer for [Does this affect client agents? ](faq#q-does-this-affect-client-agents)
 
 ## Q: What is the behavior if the license is missing?
 
@@ -108,6 +114,7 @@ When a customer deploys new clusters to a 1.10.0+ent release, they need to have 
 
 New Consul cluster deployments using 1.10.0+ent will need to have a valid license to successfully deploy.
 This valid license must be on-disk (auto-loaded) or as an environment variable.
+Please see the [upgrade requirements](faq#q-what-are-the-upgrade-requirements).
 
 ## Q: What is the migration path for customers who want to migrate from their existing license-as-applied-via-the-CLI flow to the license on disk flow?
 
@@ -127,6 +134,7 @@ This valid license must be on-disk (auto-loaded) or as an environment variable.
 1. Follow the [Kubernetes Upgrade Docs](/docs/k8s/upgrade) to upgrade. No changes to your `values.yaml` file are needed to enable enterprise autoloading since this support is built in to consul-helm 0.32.0 and greater.
 
 ~> **WARNING:** If you are upgrading the Helm chart but **not** upgrading the Consul version, you must set `server.enterpriseLicense.enableLicenseAutoload: false`. See [Kubernetes - Consul Enterprise](/docs/k8s/installation/deployment-configurations/consul-enterprise) for more details.
+
 ## Q: What is the migration path for customers who want to migrate from their existing perpetually-licensed binaries to the license on disk flow?
 
 ### VM
@@ -148,7 +156,9 @@ This valid license must be on-disk (auto-loaded) or as an environment variable.
 
 When downgrading to a version of Consul before 1.10.0+ent, customers will need to follow the previous process for applying an enterprise licenses to Consul Enterprise.
 
-## Q: Are there potential pitfalls when downgrading or upgrading?
+## Q: Are there potential pitfalls when downgrading or upgrading Consul server instances?
+
+~> Verify that you meet the [upgrade requirements](faq#q-what-are-the-upgrade-requirements).
 
 Assume a scenario where there are three Consul server nodes:
 
--- a/website/content/docs/enterprise/license/faq.mdx
+++ b/website/content/docs/enterprise/license/faq.mdx
@@ -25,6 +25,54 @@ This will no longer work since each server must be able to find a valid license 
 If ACLs are enabled, customers on Consul Enterprise 1.8+ or 1.9+ must upgrade Consul agents to the latest patch available (1.8.13 or 1.9.7) prior to upgrading the Consul server instances to 1.10.0.
 Otherwise, the Consul agents will fail to retrieve the license with a valid agent token.
 
+The following table provides an overview of 1.10.0 upgrade scenarios.
+
+| Client Version | Server Version | ACLs     | Snapshot Agent Version                             | Guidance                                                                                                                                      |
+| -------------- | -------------- | -------- | -------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
+| 1.9.x          | 1.9.x          | Enabled  | Do not upgrade Snapshot Agent to the latest 1.9.x. | See [Upgrade Path for Consul Clients and Servers 1.9.x - ACLs Enabled](faq#upgrade-path-for-consul-clients-and-servers-1-9-x-acls-enabled).   |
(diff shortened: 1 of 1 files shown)


mkeeler (approved):

mkeeler (commented):
I like the structure of this but some of the steps are a little out of order in all of the sections here.

In general upgrades are performed by upgrading servers first, then clients and then other stuff such as the snapshot agent.

So for example in the upgrade path from 1.9.x with ACLs enabled the steps should be more like

1. Upgrade Consul servers to the latest 1.9.x
2. Upgrade Consul clients to the latest 1.9.x
3. Upgrade Consul servers to 1.10
4. Upgrade Consul clients to 1.10
5. Upgrade snapshot agents to 1.10

The other sections would require similar ordering modifications.

trujillo-adam:
@mkeeler - Thanks for reviewing. Should the license/env var be configured after all upgrades, including the Snapshot Agent upgrade? Can configuring the license for all components be consolidated into a single step or is there specific order that must be followed? Would the following be correct, for example:

1. Upgrade Consul servers to the latest version of 1.9.x.
2. Upgrade Consul clients to the latest version of 1.9.x.
3. Upgrade Snapshot Agent to the latest version of 1.9.x. 
4. Configure the license file or set the license environment variable for Consul servers and clients.
5. Configure the license file or set the license environment variable for Snapshot Agents.
6. Upgrade Consul servers to 1.10.0.
7. Upgrade Consul clients to 1.10.0.
8. Upgrade Snapshot Agents to 1.10.0.

karl-cardenas-coding:
@trujillo-adam I think you can cross out `#5` as the license needs to be available before the servers are upgraded to 1.10.0

mkeeler:
@trujillo-adam I think typically changing the configuration is usually lumped in with the upgrade of the consul version to minimize the number of times restarts are required.

So in your case it would be:

1. Upgrade consul servers to the latest version of 1.9.x and set the license in the env/config
2. Upgrade consul clients to the latest version of 1.9.x and set the license in the env/config
3. Upgrade snapshot agents to latest and set the license in the env config
4. Upgrade consul servers to 1.10.0
5. Upgrade consul clients to 1.10.0
6. Upgrade snapshot agents to 1.10.0

If you were to upgrade to latest 1.9.x and have a later step to configure the license in the env then applying that configuration would require another round of restarts.

Does that make sense?

Also this looks like the 1.9.x no acls -> 1.10 no acls case. For ACLs, setting the license for servers could either be done during step 1 or step 4 and step 2/3 shouldn't require setting the license at all because clients/snapshot agents can auto-retrieve the license and the upgrade is just prepping them to send the right requests along.

website/content/docs/enterprise/license/faq.mdx
```diff
@@ -20,6 +20,59 @@ This check is part of the server boot-up process.
...
+| 1.9.x          | 1.9.x          | Enabled  | Do not upgrade Snapshot Agent to the latest 1.9.x. | See [Upgrade Path for Consul Clients and Servers 1.9.x - ACLs Enabled](faq#upgrade-path-for-consul-clients-and-servers-1-9-x-acls-enabled).   |
+| 1.9.x          | 1.9.x          | Disabled | Snapshot Agent affected by upgrade.                | See [Upgrade Path for Consul Clients and Servers 1.9.x - ACLs Disabled](faq#upgrade-path-for-consul-clients-and-servers-1-9-x-acls-disabled). |
+| 1.8.x          | 1.8.x          | Enabled  | Do not upgrade Snapshot Agent to the latest 1.8.x  | See [Upgrade Path for Consul Clients and Servers 1.8.x - ACLs Disabled](faq#upgrade-path-for-consul-clients-and-servers-1-8-x-acls-enabled).  |
+| 1.8.x          | 1.8.x          | Disabled | Snapshot Agent affected by upgrade.                | See [Upgrade Path for Consul Clients and Servers 1.8.x - ACLs Disabled](faq#upgrade-path-for-consul-clients-and-servers-1-8-x-acls-disabled). |
+
```
preetapan, line 36: I would pull out the snapshot agent column out of this table. Its mentioned explicitly below for each section where its clear that when ACLs are disabled, we do require upgrading both the snapshot agent and clients/servers to the latest in the 1.9.x series.

website/content/docs/enterprise/license/faq.mdx
```diff
@@ -20,6 +20,59 @@ This check is part of the server boot-up process.
 In previous versions of HashiCorp enterprise products, one server could distribute a license to other servers via the Raft protocol.
 This will no longer work since each server must be able to find a valid license during the startup process.
 
+## Q: What are the upgrade requirements?
+
+If ACLs are enabled, customers on Consul Enterprise 1.8+ or 1.9+ must upgrade Consul agents to the latest patch available (1.8.13 or 1.9.7) prior to upgrading the Consul server instances to 1.10.0.
```
preetapan, line 25: per our talk earlier today and what I understood, *any* upgrade to 1.10.0 enterprise must start with upgrading to the latest patch, so I would replace this sentence with

"Any customers on Consul Enterprise 1.8/1.9 should first upgrade their agents to the latest patch release. Additionally, the license file must be set in an environment variable or file path, and the upgrade process varies if ACLs are not enabled in the cluster. We have provided additional instructions below"

cc @mkeeler for additional review on what I said above

mkeeler: That all looks correct to me.

website/content/docs/enterprise/license/faq.mdx
```diff
@@ -20,6 +20,58 @@ This check is part of the server boot-up process.
...
+| 1.9.x          | 1.9.x          | Enabled  | See [Upgrade Path for Consul Clients and Servers 1.9.x - ACLs Enabled](faq#upgrade-path-for-consul-clients-and-servers-1-9-x-acls-enabled).   |
+| 1.9.x          | 1.9.x          | Disabled | See [Upgrade Path for Consul Clients and Servers 1.9.x - ACLs Disabled](faq#upgrade-path-for-consul-clients-and-servers-1-9-x-acls-disabled). |
+| 1.8.x          | 1.8.x          | Enabled  | See [Upgrade Path for Consul Clients and Servers 1.8.x - ACLs Disabled](faq#upgrade-path-for-consul-clients-and-servers-1-8-x-acls-enabled).  |
+| 1.8.x          | 1.8.x          | Disabled | See [Upgrade Path for Consul Clients and Servers 1.8.x - ACLs Disabled](faq#upgrade-path-for-consul-clients-and-servers-1-8-x-acls-disabled). |
+
+For Consul clients and servers older than 1.8, follow the instructions for [upgrading to specific versions](https://www.consul.io/docs/upgrading/upgrade-specific) to install Consul 1.8.X on clients and servers before performing one of the following upgrade scenarios.
```
blake, line 37:
```suggestion
For Consul clients and servers older than 1.8, follow the instructions for [upgrading to specific versions](/docs/upgrading/upgrade-specific) to install Consul 1.8.X on clients and servers before performing one of the following upgrade scenarios.
```
minor nit: By convention most URLs on the same property use relative paths.

website/content/docs/enterprise/license/faq.mdx
```diff
@@ -20,6 +20,58 @@ This check is part of the server boot-up process.
 In previous versions of HashiCorp enterprise products, one server could distribute a license to other servers via the Raft protocol.
 This will no longer work since each server must be able to find a valid license during the startup process.
 
+## Q: What are the upgrade requirements?
+
+All customers on Consul Enterprise 1.8/1.9 must first upgrade their agents to the latest patch release.
+The license file must also be set in an environment variable or file path. Otherwise,
+the Consul agents will fail to retrieve the license with a valid agent token.
+The upgrade process varies if ACLs are enabled or disabled in the cluster. The following table and corresponding sections describe common 1.10.0 upgrade scenarios.
+
```
preetapan, line 29: @mkeeler has a whole new section on 1.10.0 upgrades in https://github.com/hashicorp/consul/pull/10446 that go over each scenario in detail. Rather than repeating them here, I think its best to share the most relevant bit of information which is the first paragraph above until "The upgrade process varies if ACLs are enabled or disabled in the cluster", and then link to the new page @mkeeler added to our upgrade section of the docs that's landing in that PR

preetapan: Link to page from #10446 for 1.10.0 will be in `/docs/upgrading/instructions/upgrade-to-1-10-x`, to link to from here

trujillo-adam: Great, we'll remove the table and the subsections and add the link.

preetapan (approved):

website/content/docs/enterprise/license/faq.mdx
```diff
@@ -20,6 +20,13 @@ This check is part of the server boot-up process.
 In previous versions of HashiCorp enterprise products, one server could distribute a license to other servers via the Raft protocol.
 This will no longer work since each server must be able to find a valid license during the startup process.
 
+## Q: What are the upgrade requirements?
+
+All customers on Consul Enterprise 1.8/1.9 must first upgrade their agents to the latest patch release.
+The license file must also be set in an environment variable or file path, otherwise the Consul agents will fail to retrieve the license with a valid agent token.
```
blake, line 26:
```suggestion
All customers on Consul Enterprise 1.8/1.9 must first upgrade their client and server agents to the latest patch release.
During the upgrade the license file must also be configured on client agents in an environment variable or file path, otherwise the Consul agents will fail to retrieve the license with a valid agent token.
```

website/content/docs/enterprise/license/faq.mdx
```diff
@@ -61,13 +68,14 @@ As Consul Enterprise approaches the expiration date, warnings will be issued in
 
 ## Q: Does this affect client agents?
 
-For existing clusters, if the clients agents are using ACLs and have a valid token, then they will be able to retrieve the license from the server.
+For existing clusters, if the clients agents are using ACLs and have a valid token, then they will be able to retrieve the license from the server if the upgrade requirements have been met. Please see the [upgrade requirements](faq#q-what-are-the-upgrade-requirements).
+
```
blake, line 72: The upgrade requirements state that the license needs to be present on the host, yet this section states that the clients will be able to retrieve the license from the servers if they have a valid ACL token.

This is somewhat confusing, and I'm not sure how to re-word it.

website/content/docs/enterprise/license/faq.mdx
```diff
@@ -108,6 +116,7 @@ When a customer deploys new clusters to a 1.10.0+ent release, they need to have
 
 New Consul cluster deployments using 1.10.0+ent will need to have a valid license to successfully deploy.
 This valid license must be on-disk (auto-loaded) or as an environment variable.
```
blake, line 118:
```suggestion
New Consul cluster deployments using 1.10.0+ent will need to have a valid license on servers to successfully deploy.
This valid license must be on-disk (auto-loaded) or as an environment variable.
```

blake (commented):
I left a few comments/suggestions. Here's some additional suggestions which GitHub would allow me to leave on the relevant lines since they were out of scope from the latest changes on the PR.

* A couple of the URLs point to localhost under the "What resources are available" section. These should be modified to point to relative URL paths.
* Under "What is the behavior if the license is missing", the sentence about the Consul snapshot agent should either be a bulleted list, similar to the previous section about Consul clients, or it should be combined with the Consul client section given that the conditions are the same.
* Under "Does this need to happen every time a node restarts, or is this a one-time check?", new clusters should only have the license configured on disk if:
    * The agent is a server.
    * The agent is a client and ACLs are not enabled in the environment.

website/content/docs/enterprise/license/faq.mdx
karl-cardenas-coding: I simplified the sentence and instead relying on the previous information provided at the top to avoid confusion.

```md
There are upgrade requirements that affect Consul Enterprise clients.
Please review the [upgrade requirements](faq#q-what-are-the-upgrade-requirements) documentation.
```

website/content/docs/enterprise/license/faq.mdx
```diff
@@ -61,13 +68,15 @@ As Consul Enterprise approaches the expiration date, warnings will be issued in
 
 ## Q: Does this affect client agents?
 
-For existing clusters, if the clients agents are using ACLs and have a valid token, then they will be able to retrieve the license from the server.
+There are upgrade requirements that affect Consul Enterprise clients.
+Please review the [upgrade requirements](faq#q-what-are-the-upgrade-requirements) documentation.
+
 If the client agents are not using ACLs, then the client agents will be need to have the license on-disk (auto-loading) or as an environment variable.
 For new Consul clusters using Consul 1.10.0+ent, customers must ensure that there is a valid license on-disk (auto-loaded) or as an environment variable.
```
preetapan, line 75: I think this second sentence "For new clusters using .." can be removed, since it doesn't apply to *new* 1.10.0 clients and this sentence is in the client agent section, also the latest commit added the upgrade requirements above and linked to that already..

blake (approved):

PR #10441, opened 2021-06-21
Add ability to load a license from the configuration/environment

mkeeler:
This is mainly for forward compatibility with 1.10 where licensing requirements are changing.

For older releases we are adding the ability to load the license from a configuration/environment to facilitate a smoother upgrade process. For servers, we will allow the configuration to be set but it will not be used (a warning log is emitted saying as much). For client agents it will actually cause the license to be used in place of the auto-retrieval process.

Unlike with 1.10, client agents will not be able to update the license loaded via config with a reload. It is expected that this configuration is only used while on the path to upgrading to 1.10 and therefore the ability to reload the license should not be necessary.


rboyer (approved):
LGTM

PR #10442, opened 2021-06-21
[1.8.x] Backport #10441 - Add ability to load a license from the configuration/environment

mkeeler:
This is mainly for forwards compatibility with 1.10 and should not be relied on for a cluster staying on a 1.8.x/1.9.x version. This was also just a backport of #10441 to the release/1.8.x branch.

For older releases we are adding the ability to load the license from a configuration/environment to facilitate a smoother upgrade process. For servers, we will allow the configuration to be set but it will not be used (a warning log is emitted saying as much). For client agents it will actually cause the license to be used in place of the auto-retrieval process.

Unlike with 1.10, client agents will not be able to update the license loaded via config with a reload. It is expected that this configuration is only used while on the path to upgrading to 1.10 and therefore the ability to reload the license should not be necessary.

### DIFF
--- a/agent/config/runtime_test.go
+++ b/agent/config/runtime_test.go
@@ -4707,6 +4707,7 @@ func TestFullConfig(t *testing.T) {
 			},
 			"key_file": "IEkkwgIA",
 			"leave_on_terminate": true,
+			"license_path": "/path/to/license.lic",
 			"limits": {
 				"http_max_conns_per_client": 100,
 				"https_handshake_timeout": "2391ms",
@@ -5372,6 +5373,7 @@ func TestFullConfig(t *testing.T) {
 			}
 			key_file = "IEkkwgIA"
 			leave_on_terminate = true
+			license_path = "/path/to/license.lic"
 			limits {
 				http_max_conns_per_client = 100
 				https_handshake_timeout = "2391ms"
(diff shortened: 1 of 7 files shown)


agent/config/runtime_test.go
```diff
@@ -7194,11 +7196,11 @@ func TestSanitize(t *testing.T) {
 				"Enabled": false,
 				"AllowReuse": false,
 				"AuthMethod": {
-					"ACLAuthMethodEnterpriseFields": {},
+					"ACLAuthMethodEnterpriseFields": ` + entAuthMethodFieldsJSON + `,
```
mkeeler, line 7199: These tests couldn't pass for both oss and enterprise. Not sure how these could have passed before without these changes. The AutoConf config builder creates the AuthMethod [here](https://github.com/hashicorp/consul-enterprise/blob/5ff983fd34d7d643ee9b4083508d5a8af7512b50/agent/config/builder.go#L2261-L2280)

The EnterpriseMeta is always a concrete struct so it will always end up in here for enterprise but blank for OSS. As for the auth method enterprise fields I think it should be the same way due to the ent fields being embedded as a concrete instance in the ACLAuthMethod struct.

agent/consul/client_test.go
```diff
@@ -78,7 +79,9 @@ func testClientWithConfigWithErr(t *testing.T, cb func(c *Config)) (string, *Cli
 		t.Fatalf("err: %v", err)
 	}
 
-	client, err := NewClient(config, WithLogger(logger), WithTLSConfigurator(tlsConf))
+	store := &token.Store{}
+
+	client, err := NewClient(config, WithLogger(logger), WithTLSConfigurator(tlsConf), WithTokenStore(store))
```
mkeeler, line 84: Apparently we were running all test clients without a token store in 1.8.x and the recent additions caused plenty of test failures now that we expect to be able to retrieve a token from it.

mkeeler:
Nomad integration seems to be failing due to a failure to build nomad. So not our problem. The 0.8 integration still works just fine.

rboyer (approved):
LGTM

PR #10439, opened 2021-06-21
tlsutil: additional godoc and small cleanup

dnephin:
Best viewed by individual commit.

Adds some godoc to explain what the types/fields are used for.
Removes some unused things, and un-exports internal things.
Change a couple pointer fields to non-pointers to remove the need to point at different places in memory.

PR #10444, opened 2021-06-21
tlsutil: reduce interface provided to auto-config

dnephin:
Branched from #10439

Replace two methods with a single one that returns the cert. This moves more of the logic into the single caller (auto-config).

tlsutil.Configurator is widely used. By keeping it smaller and focused only on storing and returning TLS config, we make the code easier to follow.

These two methods were more related to auto-config than to tlsutil, so reducing the interface moves the logic closer to the feature that requires it.

### DIFF
--- a/agent/auto-config/run.go
+++ b/agent/auto-config/run.go
@@ -116,14 +116,17 @@ func (ac *AutoConfig) run(ctx context.Context, exit chan struct{}) {
 	// expires. The agent cache should be handling the expiration
 	// and renew it before then.
 	//
-	// If there is no cert, AutoEncryptCertNotAfter returns
-	// a value in the past which immediately triggers the
+	// If there is no cert, use a value which immediately triggers the
 	// renew, but this case shouldn't happen because at
 	// this point, auto_encrypt was just being setup
 	// successfully.
 	calcFallbackInterval := func() time.Duration {
-		certExpiry := ac.acConfig.TLSConfigurator.AutoEncryptCertNotAfter()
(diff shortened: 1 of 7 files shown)


i0rek (approved):
LGTM

agent/auto-config/run.go
```diff
@@ -122,8 +122,11 @@ func (ac *AutoConfig) run(ctx context.Context, exit chan struct{}) {
 	// this point, auto_encrypt was just being setup
 	// successfully.
 	calcFallbackInterval := func() time.Duration {
-		certExpiry := ac.acConfig.TLSConfigurator.AutoEncryptCertNotAfter()
-		return certExpiry.Add(ac.acConfig.FallbackLeeway).Sub(time.Now())
+		cert := ac.acConfig.TLSConfigurator.AutoEncryptCert()
+		if cert == nil {
+			return -1
+		}
+		return cert.NotAfter.Add(ac.acConfig.FallbackLeeway).Sub(time.Now())
```
freddygv, line 129: I think this line got harder to parse now, what if there was an intermediate variable here?

```suggestion
		renewalDeadline := cert.NotAfter.Add(ac.acConfig.FallbackLeeway)
		return renewalDeadline.Sub(time.Now())
```

freddygv (approved):
LGTM, made a couple minor suggestions

freddygv (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/393586.

PR #9009, opened 2020-10-22
connect: Fix an issue with updating CA config in a secondary datacenter

kyhavlov:
This PR fixes a case where updating the CA config in a secondary datacenter would cause an error when it should trigger the creation of a new intermediate certificate. 

The test passes but I think this still needs some extra logic to prevent races between the RPC endpoint and `secondaryCARootWatch`/`intermediateCertRenewalWatch`. Currently `secondaryCARootWatch` will set things right on its next iteration if there's a race where it gets overwritten by a call to the RPC endpoint, but since it spends most of its time waiting on a blocking query that could be up to 5-10 minutes with the wrong intermediate/root, which is a pretty long time to wait for things to converge.

Fixes #7009.

PR #10445, opened 2021-06-21
ca: isolate more of the CA logic in CAManager

dnephin:

This is in preparation for documenting the CA provider flows, and fixing some bugs related to CA provider in secondary DCs.


dnephin:
Looks like there is a lint error and a couple test failures to fix.

agent/consul/leader_connect_ca.go
```diff
@@ -9,35 +9,34 @@ import (
...
 const (
 	caStateUninitialized     caState = "UNINITIALIZED"
-	caStateInitializing              = "INITIALIZING"
-	caStateInitialized               = "INITIALIZED"
-	caStateRenewIntermediate         = "RENEWING"
-	caStateReconfig                  = "RECONFIGURING"
+	caStateInitializing      caState = "INITIALIZING"
+	caStateInitialized       caState = "INITIALIZED"
+	caStateRenewIntermediate caState = "RENEWING"
+	caStateReconfig          caState = "RECONFIGURING"
```
dnephin, line 29: Repeating the type only works with `iota`. In blocks like this without iota the type has to be explicit. This was changed so that tests can use `require.Equal` , instead of `require.EqualValue`

agent/consul/leader_connect_ca.go
```diff
@@ -100,13 +100,6 @@ func NewCAManager(delegate caServerDelegate, leaderRoutineManager *routine.Manag
 	}
 }
 
-func (c *CAManager) reset() {
-	c.state = caStateUninitialized
-	c.primaryRoots = structs.IndexedCARoots{}
-	c.actingSecondaryCA = false
```
dnephin, line 93: TODO: check to see if this should be done in `Stop`

dnephin, line 93: TODO: check to see if this should be done in `Stop`, should primaryRoots and actingSecondarCA be called to zero things out on leader loss.

dhiaayachi: I checked and I'm pretty sure it's needed to clean up the struc and the state. I added them as part of the stop

agent/consul/leader_connect_ca.go
```diff
@@ -386,6 +398,24 @@ func (c *CAManager) InitializeCA() (reterr error) {
 	return nil
 }
 
+// createProvider returns a connect CA provider from the given config.
+func (c *CAManager) newProvider(conf *structs.CAConfiguration) (ca.Provider, error) {
+	logger := c.logger.Named(conf.Provider)
+	switch conf.Provider {
+	case structs.ConsulCAProvider:
+		return &ca.ConsulProvider{Delegate: c.delegate, Logger: logger}, nil
```
freddygv, line 406: Maybe there should be a `NewConsulProvider` like with the other providers

agent/consul/server.go
```diff
@@ -473,7 +473,7 @@ func NewServer(config *Config, flat Deps) (*Server, error) {
 		return nil, fmt.Errorf("Failed to start Raft: %v", err)
 	}
 
-	s.caManager = NewCAManager(&caDelegateWithState{s}, s.leaderRoutineManager, s.loggers.Named(logging.Connect), s.config)
+	s.caManager = NewCAManager(&caDelegateWithState{s}, s.leaderRoutineManager, s.logger.ResetNamed("connect.ca"), s.config)
```
freddygv, line 476: There's a `logging.CA` constant, "ca", that could be used here

agent/consul/leader_connect_ca.go
freddygv: How was `reset()` being used originally? I can't find where it was being called from

freddygv (commented):
This mostly looks good to me, left a couple questions/comments.

The main one being the question about `CAManager.reset()`

agent/connect/ca/provider_aws_test.go
```diff
@@ -375,9 +376,7 @@ func TestAWSProvider_Cleanup(t *testing.T) {
 }
 
 func testAWSProvider(t *testing.T, cfg ProviderConfig) *AWSProvider {
-	p := &AWSProvider{}
-	logger := testutil.Logger(t)
-	p.SetLogger(logger)
+	p := &AWSProvider{logger: testutil.Logger(t)}
```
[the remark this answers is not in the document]
dnephin: Nope, I think I must have added the constructor after I had made this change and forgot it existed. We could do that.

agent/connect/ca/provider_consul.go
```diff
@@ -38,7 +37,7 @@ type ConsulProvider struct {
 	clusterID string
 	isPrimary bool
 	spiffeID  *connect.SpiffeIDSigning
-	logger    hclog.Logger
+	Logger    hclog.Logger
```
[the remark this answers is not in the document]
dnephin: Only because there is no constructor. If we add one we can unexport this too.

agent/consul/server.go
dnephin: This was to keep it consistent with the previous logger name.

IMO, the `logging` constants are not really valuable. I think we should aim to only create a logger with a given name in one place, which makes the name constants in some far away package more of a nuisance than something helpful. If we want a constant it could be un-exported in the local package.

agent/consul/leader_connect_ca.go
dnephin: It wasn't, which is why 7dae65c was able to remove the method without any other changes. While reviewing this I thought it might be worth checking to see if the `Stop` method was missing any of this logic, and I guess it was.

I believe an earlier change must have removed the call to `reset`, but I haven't looked into it.

dnephin:
Pushed another commit which makes the provider constructors more consistent.

agent/consul/leader_connect_ca.go
dnephin: `reset` seems like it was added in 0bfda4481f004fd40759b9afe99518dc5b37bc3a (#9009) but never used.

freddygv: Got it. I just saw that when reset was added it wasn't being called then either, looks like a stray change.

The important parts of `reset` were to set the state to uninitialized and set the provider to nil, and those were being handled separately in `revokeLeadership`.

It's definitely clearer to move those down into caManager.Stop() rather than spread around.

freddygv (approved):
LGTM, solid improvement!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/407297.

PR #10447, opened 2021-06-22
ui: Adds Partitions to the HTTP layer

johncowen:
This PR mainly adds `partition` to our HTTP adapter. Additionally and perhaps most importantly, we've also taken the opportunity to move our 'conditional namespaces' deeper into the app.

The reason for doing this was, we like that namespaces should be thought of as required instead of conditional, 'special' things and would like the same thinking to be applied to `partitions`.

Now, instead of using code ~throughout the app~ throughout the adapters to add/remove namespaces or partitions depending on whether they are enabled or not. As a UI engineer you just pretend that namespaces and partitions are always enabled, and we remove them for you deeper in the app, out of the way of you forgetting to treat these properties as a special case.

Notes:

1. Added a PartitionAbility while we were there (not used as yet)
2. Started to remove the CONSTANT variables we had just for property names. I prefer that our adapters are as readable and straightforwards as possible, it just looks like HTTP.
3. We'll probably remove our `formatDatacenter` method we use also at some point, it was mainly too make it look the same as our previous `formatNspace`, but now we don't have that, it instead now looks different!
4. We enable parsing of `partition` in the UIs URL, but this is feature flagged so still does nothing just yet.
5. All of the test changes were related to the fact that we were treating `client.url` as a function rather than a method, and now that we reference `this` in `client.url` (etc) it needs binding to `client`.

There are no additional tests added for `partition` as yet.

---

I've been a AFK for a while since this originally went up in draft, and there is some additional information worthwhile adding here before I take it out of draft.

I also found out that whilst partitions are almost the same feature as namespaces, they are 'slightly' different even in the way the HTTP API works.

Namespace related HTTP API lets us use a `?ns=` queryParam, a `X-Consul-Namespace` HTTP header, or for PUTing/POSTing a `Namespace` property in the HTTP body.

When building our namespace support we decided to stick as close to recognised REST-like standards as possible. So for all GET requests we used the `?ns=` queryParam, and for all PUT/POST requests we used the `Namespace` HTTP body property (that was part of the data we were PUTing/POSTing.

One of the little 'slightly different' facts out our Partition support is that its HTTP API interaction doesn't allow you to specify the partition in the HTTP body at all. All partition HTTP interaction MUST use either the `?paritition=` or the `X-Consul-Partition` HTTP header. This means even though (from our UIs perspective) namespaces and partitions are pretty much the same thing, just different levels of the hierarchy, the HTTP interaction would have been (previous to the work in this PR) slightly different. Namespaces PUTing/POSTing using HTTP body and Partitions PUTing/POSTing using the `?partition=` query param.

Considering everything, now seemed like a good time to change the previous Namespace HTTP interaction to only use `?ns=` even for PUTing/POSTing rather than have this weird little confusing difference between Namespaces and Partitions in our codebase forever.

Therefore this PR included moving all the the Namespace HTTP interaction to only ever use `?ns=`.

And then came the tests...

As we had moved things around a little during this PR, we already had to make some integration test amends to reflect the fact that code that was affecting the result of the integration tests had been moved to a different level/layer, and therefore the expected result changed. But with this change to '`?ns=` everywhere' meant we also had to update all our tests to expect this - everywhere.

All in all this was a similar level of tedium that we mentioned in the earlier piece of this PR, and as well as just sitting down to get the same tiny little thing done everywhere throughout our tests - you need to sit a really think about whether you were actually changing the test to ensure its testing what it should be. As 'a namespace' can depend on whether you are in OSS or not, and then what happens with an undefined namespace with no token, and undefined namespace with a token that is different from the current namespace etc etc to take into account, you have to be careful going through this. Added to this is the fact that our test suite is currently ~4000-5000 tests strong, which means it take around 20-25 minutes or so to run the entire test suite, and whilst I'd generally be just filtering the tests I was interested in at anyone time, it still took quite a while of backwards and forwards-ing and re-running tests to make sure I'd thought everything through properly (🤞 ). Luckily I've had plenty to work on apart from this so it hasn't meant much "just watching and waiting for a green".

Finally this was just about adding partition support to the HTTP layer (and making namespaces consistent with it), once this is merged application layer will be next up.

### DIFF
--- a/ui/packages/consul-ui/app/abilities/nspace.js
+++ b/ui/packages/consul-ui/app/abilities/nspace.js
@@ -16,7 +16,7 @@ export default class NspaceAbility extends BaseAbility {
   }
 
   get canChoose() {
-    return this.canUse && (this.nspaces.length || []) > 0;
+    return this.canUse && (this.nspaces || []).length > 0;
   }
 
   get canUse() {
--- a/ui/packages/consul-ui/app/abilities/partition.js
+++ b/ui/packages/consul-ui/app/abilities/partition.js
@@ -1,7 +1,7 @@
 import BaseAbility from './base';
 import { inject as service } from '@ember/service';
 
-export default class NspaceAbility extends BaseAbility {
+export default class PartitionAbility extends BaseAbility {
   @service('env') env;
 
   resource = 'operator';
@@ -16,10 +16,10 @@ export default class NspaceAbility extends BaseAbility {
   }
 
   get canChoose() {
-    return this.canUse && (this.nspaces.length || []) > 0;
+    return this.canUse && (this.partitions || []).length > 0;
   }
 
   get canUse() {
-    return this.env.var('CONSUL_NSPACES_ENABLED');
+    return this.env.var('CONSUL_PARTITIONS_ENABLED');
   }
 }
--- a/ui/packages/consul-ui/app/locations/fsm-with-optional.js
+++ b/ui/packages/consul-ui/app/locations/fsm-with-optional.js
@@ -1,9 +1,9 @@
 import { env } from 'consul-ui/env';
 const OPTIONAL = {};
-// if (true) {
-//   OPTIONAL.partition = /^-([a-zA-Z0-9]([a-zA-Z0-9-]{0,62}[a-zA-Z0-9])?)$/;
-// }
-//
+if (env('CONSUL_PARTITIONS_ENABLED')) {
+  OPTIONAL.partition = /^-([a-zA-Z0-9]([a-zA-Z0-9-]{0,62}[a-zA-Z0-9])?)$/;
+}
+
 if (env('CONSUL_NSPACES_ENABLED')) {
   OPTIONAL.nspace = /^~([a-zA-Z0-9]([a-zA-Z0-9-]{0,62}[a-zA-Z0-9])?)$/;
 }
@@ -193,9 +193,9 @@ export default class FSMWithOptionalLocation {
     if (typeof hash.nspace !== 'undefined') {
       hash.nspace = `~${hash.nspace}`;
     }
-    // if (typeof hash.partition !== 'undefined') {
-    //   hash.partition = `-${hash.partition}`;
-    // }
+    if (typeof hash.partition !== 'undefined') {
+      hash.partition = `-${hash.partition}`;
+    }
     if (typeof this.router === 'undefined') {
       this.router = this.container.lookup('router:main');
     }
--- a/ui/packages/consul-ui/app/services/client/http.js
+++ b/ui/packages/consul-ui/app/services/client/http.js
@@ -3,7 +3,11 @@ import { get } from '@ember/object';
 import { next } from '@ember/runloop';
 
 import { CACHE_CONTROL, CONTENT_TYPE } from 'consul-ui/utils/http/headers';
-import { HEADERS_TOKEN as CONSUL_TOKEN } from 'consul-ui/utils/http/consul';
+import {
+  HEADERS_TOKEN as CONSUL_TOKEN,
+  HEADERS_NAMESPACE as CONSUL_NAMESPACE,
+  HEADERS_DATACENTER as CONSUL_DATACENTER,
+} from 'consul-ui/utils/http/consul';
 
 import createURL from 'consul-ui/utils/http/create-url';
 import createHeaders from 'consul-ui/utils/http/create-headers';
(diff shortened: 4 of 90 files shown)


ui/packages/consul-ui/app/abilities/nspace.js
```diff
@@ -16,7 +16,7 @@ export default class NspaceAbility extends BaseAbility {
   }
 
   get canChoose() {
-    return this.canUse && this.nspaces.length > 0;
+    return this.canUse && (this.nspaces.length || []) > 0;
```
johncowen, line 19: This was another 'while I was there':

When I added the PartitionAbility I noticed that (whilst its never the case in our app) partition/nspaces here could be null, so I just guarded it for safety incase it ever is the case in our app.

ui/packages/consul-ui/app/abilities/partition.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  get canChoose() {
+    return this.canUse && (this.partitions || []).length > 0;
+  }
+
+  get canUse() {
+    return this.env.var('CONSUL_PARTITIONS_ENABLED');
+  }
+}
```
johncowen, line 25: This is just copypasta from the NsapceAbility for now. I need to double check that it requires the same `operator` resource.

ui/packages/consul-ui/app/locations/fsm-with-optional.js
```diff
@@ -190,9 +190,9 @@ export default class FSMWithOptionalLocation {
     if (typeof hash.nspace !== 'undefined') {
       hash.nspace = `~${hash.nspace}`;
     }
-    // if (typeof hash.partition !== 'undefined') {
-    //   hash.partition = `-${hash.partition}`;
-    // }
+    if (typeof hash.partition !== 'undefined') {
+      hash.partition = `-${hash.partition}`;
+    }
```
johncowen, line 195: These two changes allow `/-partition-name/` in URLs (both on the way in and out)

ui/packages/consul-ui/app/services/client/http.js
```diff
@@ -26,8 +26,9 @@ export const restartWhenAvailable = function(client) {
     throw e;
   };
 };
-const stringifyQueryParams = createQueryParams(encodeURIComponent);
-const parseURL = createURL(encodeURIComponent, stringifyQueryParams);
+const QueryParams = {
+  stringify: createQueryParams(encodeURIComponent),
+};
```
johncowen, line 31: I made this look like `JSON.stringify` as I vaguely remember doing the same elsewhere in the app. At somepoint we could probably move this to a separate import and use this interface in both places.

ui/packages/consul-ui/app/services/client/http.js
```diff
@@ -72,21 +73,34 @@ const parseBody = function(strs, ...values) {
...
+      }
+    }
+    if (!this.env.var('CONSUL_PARTITIONS_ENABLED')) {
+      delete obj.partition;
+    } else {
+      if (typeof obj.partition === 'undefined' || obj.partition === null || obj.partition === '') {
+        delete obj.partition;
+      }
+    }
+    return obj;
```
johncowen, line 103: This is the stuff that removes `?ns` and `?partition` if you don't have them enabled. When I did this I remember thinking it's a shame that its a bit `mutate-y` but 🤷 for now at least.

ui/packages/consul-ui/tests/integration/adapters/topology-test.js
```diff
@@ -22,16 +23,18 @@ module('Integration | Adapter | topology', function(hooks) {
...
         dc: dc,
       });
     });
   });
   test('requestForQueryRecord returns the correct body', function(assert) {
     return nspaceRunner(
       (adapter, serializer, client) => {
-        return adapter.requestForQueryRecord(client.body, {
+        const request = client.body.bind(client);
+        return adapter.requestForQueryRecord(request, {
```
johncowen, line 37: So, all this test editing was a bit tedious 😄 , got me thinking about how we could improve this somehow.

johncowen (commented):
Added some inline notes also.

ui/packages/consul-ui/app/abilities/nspace.js
johncowen: I fixed a little typo here, main comment still stands

```javascript
(this.nspaces.length || [])
// should be
(this.nspaces || []).length
```

similar to the equivalent partitions one below.

ui/packages/consul-ui/app/abilities/partition.js
johncowen: From speaking with others on the team sounds like there is a `mesh` resource to check to see whether I can actually 'manage' (i.e. CRUD) partitions. I'll confirm this with the rest of the team and do this as part of the application level changes (rather than the HTTP/model layer changes here)

johncowen (commented):
Adding extra inline comments

johncowen:
I've added some further description to the PR explaining the changes and reasons for work that has happened since the draft PR went up, so if you've been here before be sure to re-read the description. I think this is ready to come out of draft now 🎉

kaxcode (approved):
LGTM

I caught the following:
- adding `partitions` and removing `this.formatNspace(ns)` to all the adapters
- using `data.Datacenter`, `data.Namespace` , and `data.Partition` for updating, deleting, and creating records in certain adapters
- parsing of partitions in the URL - thanks for the comment
- Updating use of HTTP headers for dc and nspace.
- Tweaks to mock-data
- Updating tests to send `@!namespace` /`@namespace`,  updates for  headers, and cleanup for `client.requestParams.bind(client)`

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/445258.

PR #10440, opened 2021-06-21
contrib: add some docs for the DNS interface

dnephin:
Any other links or details I should include on this page? [Preview](https://github.com/hashicorp/consul/blob/dnephin/contib-docs-dns/contributing/service-discovery/dns.md)

PR #10452, opened 2021-06-22
add matrix for dns type vs kind

dhiaayachi:
This PR add a matrix to document the current DNS server behaviour per DNS query Type vs dns query kind

### DIFF
--- a/contributing/service-discovery/dns-matrix.md
+++ b/contributing/service-discovery/dns-matrix.md
@@ -1,0 +1,12 @@
+|                | service                 | connect (enterprise)    | ingress (enterprise)    | node                   | query           | addr                         |
+|----------------|-------------------------|-------------------------|-------------------------|------------------------|-----------------|------------------------------|
+| TypeSOA        | Supported               | Supported               | Supported               | Supported              | Supported       | Supported                    |
+| TypeNS         | Supported               | Supported               | Supported               | Supported              | Supported       | Supported                    |
+| TypeAXFR       | Not Implemented         | Not Implemented         | Not Implemented         | Not Implemented        | Not Implemented | Not Implemented              |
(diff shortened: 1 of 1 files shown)


jsosulska:
Hi @dhiaayachi - Is this something that we may want to add to our Docs site, instead of just the codebase?

contributing/service-discovery/dns-matrix.md
```diff
@@ -0,0 +1,12 @@
+|                | service                 | connect (enterprise)    | ingress (enterprise)    | node                   | query           | addr                         |
```
dnephin, line 1: A quick intro paragraph would be great to explain what this is, and maybe also to say that only a single `OpCode` is supported.

dnephin (commented):
Nice! Do you think it would make sense to include on the same page as the docs in #10440 ?

dnephin:
>  Is this something that we may want to add to our Docs site, instead of just the codebase?

I think our existing docs for the [DNS interface](https://www.consul.io/docs/discovery/dns) are probably in a better format for users. I would expect a table like this to mostly only be useful for someone looking to understand the code.

Ideally the code would be structured in a way that makes the matrix obvious, but until then the docs will help.

dhiaayachi:
> Nice! Do you think it would make sense to include on the same page as the docs in #10440 ?

Yes If OK I can bring it to the same PR

dhiaayachi:
> Hi @dhiaayachi - Is this something that we may want to add to our Docs site, instead of just the codebase? 

IMO it will be a high maintenance task to keep it in sync with the code specially in the current state. I would wait until the code is properly refactored and reevaluate after if still needed.

dhiaayachi:
I added the change in this PR in #10440

PR #10460, opened 2021-06-23
Describe how to run a subset of all unit tests

jkirschner-hashicorp:
Useful for contributors to check their work-in-progress by running only the tests which they expect to be affected by their changes, as the full test suite can take several minutes to execute.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -90,6 +90,16 @@ lives in the spot where the go cli tools are expecting to find it.
 
 ## Testing
 
+### During Development: Run Relevant Test(s)
+
+During development, it may be more convenient to check your work-in-progress by running only the tests which you expect to be affected by your changes, as the full test suite can take several minutes to execute. [Go's built-in test tool](https://golang.org/pkg/cmd/go/internal/test/) allows specifying a list of packages to test and the `-run` option to only include test names matching a regular expression.
+
+Examples (run from the repository root):
+- `go test -v ./connect` will run all tests in the connect package (see `./connect` folder)
+- `go test -v -run TestRetryJoin ./command/agent` will run all tests in the agent package (see `./command/agent` folder) with name substring `TestRetryJoin`
+
+### Before Submitting Changes: Run All Tests
+
 Before submitting changes, run **all** tests locally by typing `make test`. 
 The test suite may fail if over-parallelized, so if you are seeing stochastic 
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -90,6 +90,16 @@ lives in the spot where the go cli tools are expecting to find it.
 
 ## Testing
 
+### During Development: Run Relevant Test(s)
+
+During development, it may be more convenient to check your work-in-progress by running only the tests which you expect to be affected by your changes, as the full test suite can take several minutes to execute. [Go's built-in test tool](https://golang.org/pkg/cmd/go/internal/test/) allows specifying a list of packages to test and the `-run` option to only include test names matching a regular expression.
+
+Examples (run from the repository root):
+- `go test -v ./connect` will run all tests in the connect module (see `./connect` folder)
```
dnephin, line 98:
```suggestion
- `go test -v ./connect` will run all tests in the connect package (see `./connect` folder)
```

.github/CONTRIBUTING.md
```diff
@@ -90,6 +90,16 @@ lives in the spot where the go cli tools are expecting to find it.
...
+- `go test -v ./command/agent -run TestRetryJoin` will run all tests in the agent module (see `./command/agent` folder) with name substring `TestRetryJoin`
```
dnephin, line 99:
```suggestion
- `go test -v -run TestRetryJoin ./command/agent` will run all tests in the `command/agent` package with name substring `TestRetryJoin`
```

dnephin (approved):
Thanks! LGTM, but one suggestion for improving clarity. Modules in Go are how libraries are distributed. Every module contains one or more packages. The things being referred to here are packages.

.github/CONTRIBUTING.md
dnephin: I also changed the order of the args. I believe the `-run` arg must come before the package list.

jkirschner-hashicorp: Turns out it works either way - I had been running the commands this way while working on my first bugfix. But I can see why that wouldn't be expected (flags don't usually come at the end like that), so I 100% agree with this change.

jkirschner-hashicorp:
PR review feedback addressed in a commit that's been fixed-up with the original. Will merge given prior approval once tests pass in CI.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/417101.

PR #10461, opened 2021-06-23
regenerate expired root certs for aws consul and vault

### LINKED ISSUE #9574: CA re-initialization could result in invalid certs
We initialize our CA config on a fresh cluster, and also on leader elections. If any CA certificate in the chain has already expired (either root or intermediate), the leader loop currently doesn't regenerate them. 

Potential fix for this is to regenerate any expired CA certs in the leader loop. This logic is specific per provider and this issue may need to be broken up into smaller issues per provider.   

One option is to implement this (ie regenerate if expired) in the `GenerateRoot` or `GenerateIntermediate` methods (current behavior in these methods is to not regenerate if there's a non empty value).

dhiaayachi:
Fix #9574 
Check if cert is expired in `GenerateRoot` of each ca provider and regenerate it if so

### DIFF
--- a/.changelog/10461.txt
+++ b/.changelog/10461.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+added a check to the intermediate certificate expiration time before it's used to sign a leaf certificate.
+```
--- a/agent/connect/ca/common.go
+++ b/agent/connect/ca/common.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"crypto/x509"
 	"fmt"
+	"time"
 
 	"github.com/hashicorp/consul/agent/connect"
 )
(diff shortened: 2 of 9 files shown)


dnephin (commented):
Thanks for looking into this issue!

My understanding is that probably this needs to happen more in `agent/consul/leader_connect_ca.go`. I suspect we'll want to regenerate before the root expires to prevent any interruption in service.

dhiaayachi:
> Thanks for looking into this issue!
> 
> My understanding is that probably this needs to happen more in `agent/consul/leader_connect_ca.go`. I suspect we'll want to regenerate before the root expires to prevent any interruption in service.

Thanks for taking a look!!
The idea I was having is to add another routine in the same fashion of `intermediateCertRenewalWatch` to do the job for the root cert, That first change was just to make sure that the provider regenerate the cert in case it expire when `generateRoot` get called. 
I'm not sure about this though, I don't yet understand all the nuances of how cert generation works.

PR #10479, opened 2021-06-23
ca: move Server.SignIntermediate to CAManager

dnephin:
Branched from #10445
Best viewed by individual commit. Most of this diff is moving code around, but the first commit shows the bulk of the change.

This PR moves Server.SignIntermediate to CAManager, which further isolates the CA management logic into a single place and helps reduce the size of Server (by removing a both a field and a method).

As part of this change, also extract a few more functions from Server that were exclusively used as part of `AutoConfigBackend`. The move of `SignIntermediate` required introducing a new struct for `autoConfigBackend`, so the rest of the methods were moved over to this new struct at the same time. This backend pattern matches the one we used for the streaming `subscribe` endpoint, and other new RPC endpoints in enterprise.


agent/consul/leader_connect_ca.go
```diff
@@ -82,6 +85,29 @@ func (c *caDelegateWithState) ApplyCARequest(req *structs.CARequest) (interface{
...
+	// term goal is still to insert some metadata into raft about the certificates
+	// and use that raft index for the ModifyIndex. This is a partial step in that
+	// direction except that we only are setting an index and not storing the
+	// metadata.
+	req := structs.CALeafRequest{
+		Op:         structs.CALeafOpIncrementIndex,
+		Datacenter: c.Server.config.Datacenter,
+	}
+	resp, err := c.Server.raftApplyMsgpack(structs.ConnectCALeafRequestType|structs.IgnoreUnknownTypeFlag, &req)
+
```
dnephin, line 108: TODO: this is missing the err check before checking the resp

dhiaayachi: Fixed

agent/consul/leader_connect_ca.go
```diff
@@ -180,7 +180,8 @@ func (c *CAManager) getPrimaryRoots() structs.IndexedCARoots {
 // when setting up the CA during establishLeadership. The state should be set to
 // non-ready before calling this.
 func (c *CAManager) initializeCAConfig() (*structs.CAConfiguration, error) {
-	st := c.delegate.State()
+	delegate := c.delegate.(ca.ConsulProviderStateDelegate)
```
dnephin, line 186: I'm not sure about this change. Why assign `c.delegate` here to a variable?

I very frequently make the opposite change (remove temporary variables) because accessing the field directly makes code easier to trace. An IDE can show you all usages of a field very quickly.

Where as, if you store the value on a local variable, now you have to pull up two different lists of calls. Every time that is done it makes it harder to trace.

What problem was this fixing? The commit mentions fixing a lint error, but I'm not familiar with any linter warnings about this, and I didn't see any from CI in older commits.

dnephin (commented):
@dhiaayachi Could you share more context about the changes in 6356302b54f06c8f2dee8e59740409d49e84ef24 ?

I don't remember any linter errors and I can't tell what was fixed from the diff.

agent/consul/leader_connect_ca.go
dhiaayachi: The reason is that the linter type check is giving an error because of implicit cast.

dnephin: Thank you for the additional details! I had a look and I think the issue was actually a bad rebase. Commit 338e558ac00acdbe57a0f1aaaa2705613bd18e85 (which I think was introduced by rebase) added a duplicate commit, which caused the linter to error out because the build was failing. 

I reverted both 338e558ac00acdbe57a0f1aaaa2705613bd18e85 and 6356302b54f06c8f2dee8e59740409d49e84ef24, and fixed the import sorting in the latest commit. The linter seems happy with this change.

dhiaayachi: thanks, sorry for that confusion !!

freddygv (approved):
👍🏽

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/407626.

PR #9729, opened 2021-02-08
ui: Update browser targets

johncowen:
Consul UI pre 1.9 used a mainly ES5 style of javascript, this meant that the majority of the code did not need any ES6 to ES5 babel transpilation or shim code, and our javascript payload was roughly:

Consul UI: 116KB gzipped
Dependencies: 374KB gzipped

As of 1.9 we started moving towards using Ember's new Octane edition, which felt much more ES6-y. So as well as beginning to update all of our components to use glimmer components, we also updated the style of how we write the UI to use ES6 javascript rather than ES5. This meant that (with our existing browser targets file), much of this was being transpiled using babel into ES5 javascript, giving a rough javascript payload of:

Consul UI: 229KB gzipped
Dependencies: 460KB gzipped

That's roughly **200KB gzipped** of payload difference between Consul UI 1.8 and 1.9 split roughly 50/50 between app and dependencies, mainly due to moving to a ES6 style of writing javascript.

This PR alters our browser targets file to target newer browsers and transpile less.

We took a good while looking into this, basing the following on the fact that we generally use these ES6 features:

1. Template literals
2. Native classes
3. Arrow functions
4. Async/await
5. Default arguments/parameters
6. Destructuring
7. The odd generator

We found the newest feature here (surprisingly) was async/await, which landed in Chrome 55, FF 53, Safari 10, Edge 15, which "release date wise" takes us to around the end of 2016 for all of these 'evergreen' browsers, probably a reasonable date.

The slight caveat here is a garbage collection bug for tagged Template Literals in Safari 12, which considering our current usage of template literals wouldn't affect us.

This brings us to a browserlist of: (bumped Safari version to get non-transpiled template literals)

```
'Chrome 55', 'Firefox 53', 'Safari 13', 'Edge 15'
```

This brings us back down to:

Consul UI: 184KB gzipped
Dependencies: 431KB gzipped

A ~71KB reduction on 1.9, not close to the 200KB between 1.8 and 1.9 but a good reduction (although I'm still confused as to where we gained 130KB 🤷 )

We also took the opportunity to remove `ember-cli-autoprefixer` here as the above browserlist supports all the things that autoprefixer was prefixing for us (mainly MS flavoured grid and flexbox).

There's probably a little more we can do here but nothing quite like a 70KB reduction, so we'll probably leave that for future PRs.

Overall this frees us up to make use of some other nice JS dependencies for upcoming features with a clear conscience 😄 

### Consul UI this PR

now

PR #7651, opened 2020-04-15
ui: enable Babel source maps

mikemorris:
To display pre-transpilation content matching the source files when viewing in a browser debugger, instead of the post-transpilation ES5 source.

The only actual change here is adding the line below to `babel` config block, the other changes are just reformatting from the Prettier hook.
```
sourceMaps: sourcemaps ? 'inline' : false,
```

As referenced in https://github.com/babel/ember-cli-babel#enabling-source-maps

Before:
Screen Shot 2020-04-15 at 4 28 58 PM

After:
Screen Shot 2020-04-15 at 4 27 47 PM

PR #10491, opened 2021-06-24
ui: Allow disabling of sourcemaps via env var

johncowen:
The rebuild of the Consul UI used very minimal ES6 features for 2 reasons:

1. Ember at the time felt very ES5-like (we started on Ember 2.18), to a certain extent there are still things in Ember that feel ES5-like (no real support for non-syntax related ES6 features like `Sets` and `Maps` in handlebars templates for example)
2. Using less layers makes things simpler in lots of subtle ways, for example we didn't even realize that we had sourcemaps disabled for such a long time before https://github.com/hashicorp/consul/pull/7651

We had a period of time that we were we started moving to use more ES6 style syntax, but pretty soon after that we reduced our browser support (https://github.com/hashicorp/consul/pull/9729) by not supporting anything older than ~2016. This brought us back to being able transpile as little as possible (decorators are so necessary for ember work we can't really avoid having to transpile those until they are native), and thus removing extra layers again. This means that sourcemaps aren't that much of a necessity again (depending on what your preference is)

Additionally, for a good while now I've noticed that since we enabled sourcemaps, they don't really work correctly in Ember. I often find myself debugging something and the stack trace line numbers are completely wrong, often reporting line numbers of over 1000 when the file where the error is is only a few lines long, it even sometimes gives me minus numbers. This can sometimes make chasing down errors a little frustrating as it is essentially saying 'the error is....umm I dunno, somewhere in this file but I'm not sure where'.

Here's one I noticed today which points me to line 1336 in a file of only 85 lines.

Screenshot 2021-06-24 at 12 39 29

And another from a few days ago telling me the error was in a comment

Screenshot 2021-06-21 at 16 45 28

I've also heard both online and offline that there are problems with sourcemaps and ember, potentially related to ember-auto-import, and this might be the root of the problem. Its a strange one, as you don't notice at first, and the line numbers seem fine to begin with, but as time goes on the line numbers seem to get wildly out of kilter.

I've experimented a little with turning sourcemaps off again and I no longer seem to have problems with line numbers and my browser points me to exactly where the error is within the file. As we aren't really doing much transpilation, this is much easier for me to read and figure out what is happening.

Therefore (for the moment at least) I personally prefer not to have sourcemaps enabled, but I understand other folks might not. This PR makes enabling/disabling sourcemaps possible via a local environment variable, so I can choose what is best for me whilst not affecting anyone else and not have to remember to not commit change in our config files that I use just for myself.

We can revisit this when the new ember build tooling is ready for prime time, but if you don't set the environment variable then there is no real change here.

### DIFF
--- a/ui/packages/consul-ui/config/environment.js
+++ b/ui/packages/consul-ui/config/environment.js
@@ -13,29 +13,8 @@ const repositorySHA = utils.repositorySHA;
 const binaryVersion = utils.binaryVersion(repositoryRoot);
 
 module.exports = function(environment, $ = process.env) {
+  const env = utils.env($);
   // basic 'get env var with fallback' accessor
-  const env = function(flag, fallback) {
-    // a fallback value MUST be set
-    if (typeof fallback === 'undefined') {
-      throw new Error(`Please provide a fallback value for $${flag}`);
-    }
-    // return the env var if set
-    if (typeof $[flag] !== 'undefined') {
-      if (typeof fallback === 'boolean') {
-        // if we are expecting a boolean JSON parse strings to numbers/booleans
-        return !!JSON.parse($[flag]);
-      }
-      return $[flag];
-    }
-    // If the fallback is a function call it and return the result.
-    // Lazily calling the function means binaries used for fallback don't need
-    // to be available if we are sure the environment variables will be set
-    if (typeof fallback === 'function') {
-      return fallback();
-    }
-    // just return the fallback value
-    return fallback;
-  };
 
   let ENV = {
     modulePrefix: 'consul-ui',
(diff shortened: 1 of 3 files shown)


mikemorris:
This does sound like a bug upstream in Ember somewhere (looks like line numbers are pointing to the transpiled file but inspector is showing the pre-transpilation source) that should be fixed rather than just losing this functionality, but local env var to optionally disable feels like an okay middle ground for now 👍

kaxcode (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/404161.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit b0d69efcc16df65729948a17bd7b2aac566f4608 onto `release/1.10.x` succeeded!

PR #10495, opened 2021-06-24
docs: Ingress controllers configurations 

kschoche:
Add a document describing common configurations required for using an ingress controller with consul-k8s

Note to reviewers : this is a bit WIP at the moment, I'm trying to provide enough context and information  to be helpful without being too prescriptive about specific settings/platforms.
Ideally we could have individual blog posts showing "how to use x, y, z ingress controllers" and link them on this page?

### DIFF
--- a/website/content/docs/k8s/connect/ingress-controllers.mdx
+++ b/website/content/docs/k8s/connect/ingress-controllers.mdx
@@ -19,13 +19,13 @@ A few steps are generally required to enable an Ingress controller to join the m
 * Enable connect-injection via an annotation on the Ingress Controller's deployment: `consul.hashicorp.com/connect-inject` is `true`.
 
 * Using the following annotations on the Ingress controller's deployment, set up exclusion rules for its ports.
-  * [`consul.hashicorp.com/transparent-proxy-exclude-inbound-ports`](/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-inbound-ports) - Provides the ability to exclude a list of ports for
+   * [`consul.hashicorp.com/transparent-proxy-exclude-inbound-ports`](/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-inbound-ports) - Provides the ability to exclude a list of ports for
 inbound traffic that the service exposes from redirection. Typical configurations would require all inbound service ports
 for the controller to be included in this list.
-  * [`consul.hashicorp.com/transparent-proxy-exclude-outbound-ports`](/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-outbound-ports) - Provides the ability to exclude a list of ports for
+   * [`consul.hashicorp.com/transparent-proxy-exclude-outbound-ports`](/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-outbound-ports) - Provides the ability to exclude a list of ports for
 outbound traffic that the service exposes from redirection. These would be outbound ports used by your ingress controller
   which expect to skip the mesh and talk to non-mesh services.
-  * [`consul.hashicorp.com/transparent-proxy-exclude-outbound-cidrs`](/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-outbound-cidrs) - Provides the ability to exclude a list of CIDRs that
+   * [`consul.hashicorp.com/transparent-proxy-exclude-outbound-cidrs`](/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-outbound-cidrs) - Provides the ability to exclude a list of CIDRs that
 the service communicates with for outbound requests from redirection. It is somewhat common that an Ingress controller
 will expect to make API calls to the Kubernetes service for service/endpoint management. As such including the ClusterIP of the
 Kubernetes service is common.
--- a/website/content/docs/k8s/connect/ingress-controllers.mdx
+++ b/website/content/docs/k8s/connect/ingress-controllers.mdx
@@ -32,46 +32,45 @@ Kubernetes service is common.
 
 ~> Note: Depending on which ingress controller you use, these stanzas may differ in name and layout, but it is important to apply
 these annotations to the *pods* of your *ingress controller*.
-
-```yaml
-# An example list of pod annotations for an ingress controller, which need be applied to PODS for the controller, not the deployment itself.
-  podAnnotations:
-    consul.hashicorp.com/connect-inject: "true"
-    # Add the container ports used by your ingress controller
-    consul.hashicorp.com/transparent-proxy-exclude-inbound-ports: "80,8000,9000,8443"
-    # And the CIDR of your Kubernetes API: `kubectl get svc kubernetes -o jsonpath='{.spec.clusterIP}'
-    consul.hashicorp.com/transparent-proxy-exclude-outbound-cidrs: "10.108.0.1/32"
-```
+   ```yaml
+   # An example list of pod annotations for an ingress controller, which need be applied to PODS for the controller, not the deployment itself.
+     podAnnotations:
+       consul.hashicorp.com/connect-inject: "true"
+       # Add the container ports used by your ingress controller
+       consul.hashicorp.com/transparent-proxy-exclude-inbound-ports: "80,8000,9000,8443"
+       # And the CIDR of your Kubernetes API: `kubectl get svc kubernetes -o jsonpath='{.spec.clusterIP}'
+       consul.hashicorp.com/transparent-proxy-exclude-outbound-cidrs: "10.108.0.1/32"
+   ```
 
 * If the Ingress controller acts as a LoadBalancer and routes directly to Pod IPs instead of the ClusterIP of your Kubernetes Services
 a `ServiceDefault` CRD must be applied to *each backend service* allowing it to use the `dialedDirectly` features. By default this is disabled.
 
-```yaml
-# Example Service defaults config entry
-apiVersion: consul.hashicorp.com/v1alpha1
-kind: ServiceDefaults
-metadata:
-  name: backend
-spec:
-  transparentProxy:
-    dialedDirectly: true
-```
+   ```yaml
+   # Example Service defaults config entry
+   apiVersion: consul.hashicorp.com/v1alpha1
+   kind: ServiceDefaults
+   metadata:
+     name: backend
+   spec:
+     transparentProxy:
+       dialedDirectly: true
+   ```
 
 * An intention from the Ingress Controller to the backend application must also be applied, this could be an L4 or L7 intention:
 
-```yaml
-# example L4 intention, but an L7 intention can also be used to control access to specific routes.
-apiVersion: consul.hashicorp.com/v1alpha1
-kind: ServiceIntentions
-metadata:
-  name: ingress-backend
-spec:
-  destination:
-    name: backend
-  sources:
-    - name: ingress
-      action: allow
-```
+   ```yaml
+   # example L4 intention, but an L7 intention can also be used to control access to specific routes.
+   apiVersion: consul.hashicorp.com/v1alpha1
+   kind: ServiceIntentions
+   metadata:
+     name: ingress-backend
+   spec:
+     destination:
+       name: backend
+     sources:
+       - name: ingress
+         action: allow
+   ```
 
 ### Common Configuration Problems:
 - The Ingress Controller's ServiceAccount name and Service name differ by default in some platforms. Consul on Kubernetes requires the
--- a/website/content/docs/k8s/connect/ingress-controllers.mdx
+++ b/website/content/docs/k8s/connect/ingress-controllers.mdx
@@ -22,9 +22,9 @@ A few steps are generally required to enable an Ingress controller to join the m
   * [`consul.hashicorp.com/transparent-proxy-exclude-inbound-ports`](https://www.consul.io/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-inbound-ports) - Provides the ability to exclude a list of ports for
 inbound traffic that the service exposes from redirection. Typical configurations would require all inbound service ports
 for the controller to be included in this list.
-  * `consul.hashicorp.com/transparent-proxy-exclude-outbound-ports` - Provides the ability to exclude a list of ports for
+  * [`consul.hashicorp.com/transparent-proxy-exclude-outbound-ports`](https://www.consul.io/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-outbound-ports) - Provides the ability to exclude a list of ports for
 outbound traffic that the service exposes from redirection.
-  * `consul.hashicorp.com/transparent-proxy-exclude-outbound-cidrs` - Provides the ability to exclude a list of CIDRs that
+  * [`consul.hashicorp.com/transparent-proxy-exclude-outbound-cidrs`](https://www.consul.io/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-outbound-cidrs) - Provides the ability to exclude a list of CIDRs that
 the service communicates with for outbound requests from redirection. It is somewhat common that an Ingress controller
 will expect to make API calls to the Kubernetes service for service/endpoint management. As such including the ClusterIP of the
 Kubernetes service is common.
--- a/website/content/docs/k8s/connect/ingress-controllers.mdx
+++ b/website/content/docs/k8s/connect/ingress-controllers.mdx
@@ -19,7 +19,7 @@ A few steps are generally required to enable an Ingress controller to join the m
 * Enable connect-injection via an annotation on the Ingress Controller's deployment: `consul.hashicorp.com/connect-inject` is `true`.
 
 * Using the following annotations on the Ingress controller's deployment, setup exclusion rules for it's ports.
-  * `consul.hashicorp.com/transparent-proxy-exclude-inbound-ports` - Provides the ability to exclude a list of ports for
+  * [`consul.hashicorp.com/transparent-proxy-exclude-inbound-ports`](https://www.consul.io/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-inbound-ports) - Provides the ability to exclude a list of ports for
 inbound traffic that the service exposes from redirection. Typical configurations would require all inbound service ports
 for the controller to be included in this list.
   * `consul.hashicorp.com/transparent-proxy-exclude-outbound-ports` - Provides the ability to exclude a list of ports for
@@ -39,8 +39,9 @@ deployment:
     consul.hashicorp.com/transparent-proxy-exclude-outbound-cidrs: "10.108.0.1/32"
 ```
 
-* If the Ingress controller acts as a LoadBalancer and routes directly to Pod IPs instead of KubeDNS a `ServiceDefault` CRD
+* If the Ingress controller acts as a LoadBalancer and routes directly to Pod IPs instead of Kubernetes Services a `ServiceDefault` CRD
 must be applied to each backend service allowing it to use the `dialedDirectly` features:
+
 ```yaml
 apiVersion: consul.hashicorp.com/v1alpha1
 kind: ServiceDefaults
@@ -52,6 +53,7 @@ spec:
 ```
 
 * An intention from the Ingress Controller to the backend application must also be applied:
+
 ```yaml
 apiVersion: consul.hashicorp.com/v1alpha1
 kind: ServiceIntentions
@@ -71,4 +73,3 @@ ServiceAccount and Service to have the same name.
 
 - If the Ingress Controller does not have the correct inbound ports excluded it will fail to start and the Ingress'
 service will not get created, causing the controller to hang in the init container.
-
--- a/website/content/docs/k8s/connect/ingress-controllers.mdx
+++ b/website/content/docs/k8s/connect/ingress-controllers.mdx
@@ -30,8 +30,7 @@ will expect to make API calls to the Kubernetes service for service/endpoint man
 Kubernetes service is common.
 
 ```yaml
-# An Example from a Traefik helm chart
-# Annotations to add to the deployed services
+# An Example configuration showing some typical podAnnotations settings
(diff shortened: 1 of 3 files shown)


david-yu (approved):
LGTM to me. This doc is great!

website/content/docs/k8s/connect/ingress-controllers.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+# Configuring Ingress Controllers with Consul on Kubernetes
+
+-> This topic requires familiarity with [Ingress Controllers](https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/)
+
+-> The following information requires Consul 1.10+, Consul-k8s 0.26+, Consul-helm 0.32+ configured with
+[Transparent Proxy](/docs/connect/transparent-proxy) mode enabled.
+
+This page describes a general approach for integrating Ingress Controllers with Consul on Kubernetes to secure traffic from the Controller
+to the backend services. This allows Consul to transparently manage secure traffic from the ingress point through the entire traffic flow of the service.
```
blake, line 15:
```suggestion
to the backend services. This allows Consul to transparently secure traffic from the ingress point through the entire traffic flow of the service.
```

blake: This reads a bit strange to me. I think omitting "manage" flows a bit better, and still gets the point across.

website/content/docs/k8s/connect/ingress-controllers.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+A few steps are generally required to enable an Ingress controller to join the mesh and pass traffic through to a service:
+
+* Enable connect-injection via an annotation on the Ingress Controller's deployment: `consul.hashicorp.com/connect-inject` is `true`.
+
+* Using the following annotations on the Ingress controller's deployment, setup exclusion rules for it's ports.
+  * `consul.hashicorp.com/transparent-proxy-exclude-inbound-ports` - Provides the ability to exclude a list of ports for
```
blake, line 22:
```suggestion
  * [`consul.hashicorp.com/transparent-proxy-exclude-inbound-ports`](https://www.consul.io/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-inbound-ports) - Provides the ability to exclude a list of ports for
```
Should these link to the docs for each individual annotation so its clear what the expected input syntax should be?

website/content/docs/k8s/connect/ingress-controllers.mdx
```diff
@@ -0,0 +1,74 @@
...
+inbound traffic that the service exposes from redirection. Typical configurations would require all inbound service ports
+for the controller to be included in this list.
+  * `consul.hashicorp.com/transparent-proxy-exclude-outbound-ports` - Provides the ability to exclude a list of ports for
+outbound traffic that the service exposes from redirection.
+  * `consul.hashicorp.com/transparent-proxy-exclude-outbound-cidrs` - Provides the ability to exclude a list of CIDRs that
```
blake, line 27: This is not documented on the list of supported annotations. In fact, the list incorrectly documents the availability of a [`consul.hashicorp.com/transparent-proxy-exclude-inbound-cidrs`](https://www.consul.io/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-inbound-cidrs) annotation, which does not exist. That doc should be updated to reference this annotation for excluding outbound CIDRs.

IDK If you want to lump this in with your PR, or address that discrepancy separately.

website/content/docs/k8s/connect/ingress-controllers.mdx
```diff
@@ -0,0 +1,74 @@
...
+the service communicates with for outbound requests from redirection. It is somewhat common that an Ingress controller
+will expect to make API calls to the Kubernetes service for service/endpoint management. As such including the ClusterIP of the
+Kubernetes service is common.
+
+```yaml
+# An Example configuration showing some typical podAnnotations settings
+deployment:
+  podAnnotations:
```
blake, line 35: This syntax seems specific to a particular Helm chart. Should this example be changed to show how to add annotations to a generic Deployment resource?

website/content/docs/k8s/connect/ingress-controllers.mdx
```diff
@@ -0,0 +1,74 @@
...
+    consul.hashicorp.com/connect-inject: "true"
+    consul.hashicorp.com/transparent-proxy-exclude-inbound-ports: "80,8000,9000,8443"
+    # And the CIDR of the Kubernetes API svc
+    consul.hashicorp.com/transparent-proxy-exclude-outbound-cidrs: "10.108.0.1/32"
+```
+
+* If the Ingress controller acts as a LoadBalancer and routes directly to Pod IPs instead of KubeDNS a `ServiceDefault` CRD
+must be applied to each backend service allowing it to use the `dialedDirectly` features:
+```yaml
```
blake, line 44:
```suggestion
* If the Ingress controller acts as a LoadBalancer and routes directly to Pod IPs instead of Kubernetes Services a `ServiceDefault` CRD
must be applied to each backend service allowing it to use the `dialedDirectly` features:

```yaml
```

website/content/docs/k8s/connect/ingress-controllers.mdx
```diff
@@ -0,0 +1,74 @@
...
+kind: ServiceDefaults
+metadata:
+  name: backend
+spec:
+  transparentProxy:
+    dialedDirectly: true
+```
+
+* An intention from the Ingress Controller to the backend application must also be applied:
+```yaml
```
blake, line 55:
```suggestion
* An intention from the Ingress Controller to the backend application must also be applied:

```yaml
```

blake (commented):
Looks good overall. I left a few comments / suggestions.

website/content/docs/k8s/connect/ingress-controllers.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+This page describes a general approach for integrating Ingress Controllers with Consul on Kubernetes to secure traffic from the Controller
+to the backend services. This allows Consul to transparently secure traffic from the ingress point through the entire traffic flow of the service.
+
+A few steps are generally required to enable an Ingress controller to join the mesh and pass traffic through to a service:
+
+* Enable connect-injection via an annotation on the Ingress Controller's deployment: `consul.hashicorp.com/connect-inject` is `true`.
+
+* Using the following annotations on the Ingress controller's deployment, setup exclusion rules for it's ports.
+  * [`consul.hashicorp.com/transparent-proxy-exclude-inbound-ports`](https://www.consul.io/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-inbound-ports) - Provides the ability to exclude a list of ports for
```
joatmon08, line 22: It might help to call out that if your Ingress Controller uses a `Type: LoadBalancer` service, you must add container port exclusions to this list to allow the load balancer to communicate with the services.

website/content/docs/k8s/connect/ingress-controllers.mdx
kschoche: the idea was that we were hoping that it could be inferred via most of the helm charts having a `podAnnotations` section or `deployment` section, but yeah this is actually a traefik example, I'm struggling to find a more generic way to visualize it though. Any ideas?

website/content/docs/k8s/connect/ingress-controllers.mdx
kschoche: Fixed good catch!

website/content/docs/k8s/connect/ingress-controllers.mdx
david-yu: @Blake This is the particular pattern that I've seen most Ingress Controllers leverage annotations for their underlying pods and we know is tested and working.

joatmon08: It think the best thing we can do is add a comment inline that this example is specific to Traefik. At the very least, someone will examine this a few times before copying and pasting it.

website/content/docs/k8s/connect/ingress-controllers.mdx
```diff
@@ -0,0 +1,75 @@
...
+must be applied to each backend service allowing it to use the `dialedDirectly` features:
+
+```yaml
+apiVersion: consul.hashicorp.com/v1alpha1
+kind: ServiceDefaults
+metadata:
+  name: backend
+spec:
+  transparentProxy:
+    dialedDirectly: true
```
joatmon08, line 52: Should we add a note on why we don't set this by default for all services? I suspect that we may get more than a few questions on why you have to do it for each backend service.

website/content/docs/k8s/connect/ingress-controllers.mdx
```diff
@@ -0,0 +1,75 @@
...
+```
+
+* An intention from the Ingress Controller to the backend application must also be applied:
```
joatmon08, line 55: I think it might help to point out that this is an L4 intention and you can set L7 intentions to enforce traffic to specific routes.

website/content/docs/k8s/connect/ingress-controllers.mdx
blake: @david-yu My concern is that this configuration example is not a generic example on how to apply annotations to a Deployment. Its is a specific example on how to configure annotations for Traefik.

If I look at the Helm charts for a few of the top 5 ingress controller's (according to the [CNCF's 2020 Survey](https://www.cncf.io/wp-content/uploads/2020/11/CNCF_Survey_Report_2020.pdf)), they each provide different methods for configuring annotations on the pods of a deployment.

**NGINX**

```yaml
controller:
  pod:
    annotations: {}
```

**HAProxy** (Kong also uses this)

```yaml
podAnnotations: {}
```
**Traefik**

```yaml
deployment:
  podAnnotations: {}
```

I think it would be helpful to either document how to annotate a generic Deployment object, or add a sentence stating "The following example shows how to configure Pod annotations in the Traefik Ingress controller Helm chart. Refer to the documentation for your specific ingress on how to configure pod annotations for the ingress controller."

joatmon08:
Looks good to me! I left some individual suggestions that you can choose to add or not.

website/content/docs/k8s/connect/ingress-controllers.mdx
```diff
@@ -0,0 +1,79 @@
...
+  * [`consul.hashicorp.com/transparent-proxy-exclude-inbound-ports`](https://www.consul.io/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-inbound-ports) - Provides the ability to exclude a list of ports for
+inbound traffic that the service exposes from redirection. Typical configurations would require all inbound service ports
+for the controller to be included in this list.
+  * [`consul.hashicorp.com/transparent-proxy-exclude-outbound-ports`](https://www.consul.io/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-outbound-ports) - Provides the ability to exclude a list of ports for
+outbound traffic that the service exposes from redirection.
+  * [`consul.hashicorp.com/transparent-proxy-exclude-outbound-cidrs`](https://www.consul.io/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-outbound-cidrs) - Provides the ability to exclude a list of CIDRs that
+the service communicates with for outbound requests from redirection. It is somewhat common that an Ingress controller
+will expect to make API calls to the Kubernetes service for service/endpoint management. As such including the ClusterIP of the
+Kubernetes service is common.
```
ishustava, line 30: It'd be nice to mention for each of these when and why you'd need these. The example below doesn't exclude outbound ports, and it'd be nice to have some explanation for when you'd need to set it (if that's possible!).

kschoche: Good call! I don't think there are many cases for outbound port exclusions, at least none that I've found yet, aside from whats needed for kube api calls.
I can expand a bit for sure in the example. Just trying to keep it generic as possible.

ishustava: Makes sense! You have way more context on it, so if you think it won't makes sense because there are things that we don't know about ingress controllers, then feel free to ignore my suggestion! It was more of a nice to have 🙂

website/content/docs/k8s/connect/ingress-controllers.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+* Enable connect-injection via an annotation on the Ingress Controller's deployment: `consul.hashicorp.com/connect-inject` is `true`.
+
+* Using the following annotations on the Ingress controller's deployment, setup exclusion rules for it's ports.
+  * [`consul.hashicorp.com/transparent-proxy-exclude-inbound-ports`](https://www.consul.io/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-inbound-ports) - Provides the ability to exclude a list of ports for
+inbound traffic that the service exposes from redirection. Typical configurations would require all inbound service ports
+for the controller to be included in this list.
+  * [`consul.hashicorp.com/transparent-proxy-exclude-outbound-ports`](https://www.consul.io/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-outbound-ports) - Provides the ability to exclude a list of ports for
+outbound traffic that the service exposes from redirection.
+  * [`consul.hashicorp.com/transparent-proxy-exclude-outbound-cidrs`](https://www.consul.io/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-outbound-cidrs) - Provides the ability to exclude a list of CIDRs that
```
blake, line 27:
```suggestion
  * [`consul.hashicorp.com/transparent-proxy-exclude-inbound-ports`](/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-inbound-ports) - Provides the ability to exclude a list of ports for
inbound traffic that the service exposes from redirection. Typical configurations would require all inbound service ports
for the controller to be included in this list.
  * [`consul.hashicorp.com/transparent-proxy-exclude-outbound-ports`](/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-outbound-ports) - Provides the ability to exclude a list of ports for
outbound traffic that the service exposes from redirection.
  * [`consul.hashicorp.com/transparent-proxy-exclude-outbound-cidrs`](/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-outbound-cidrs) - Provides the ability to exclude a list of CIDRs that
```

blake: My apologies. In my earlier comment I believe my suggestion included an absolute URL. Our convention is actually to use relative URLs when pointing to content on the same domain.

website/content/docs/k8s/connect/ingress-controllers.mdx
```diff
@@ -0,0 +1,79 @@
...
+  name: backend
+spec:
+  transparentProxy:
+    dialedDirectly: true
+```
+
+* An intention from the Ingress Controller to the backend application must also be applied, this could be an L4 or L7 intention:
+
+```yaml
+# example L4 intention, but an L7 intention can also be used to enforce traffic to specific routes.
```
blake, line 61:
```suggestion
# example L4 intention, but an L7 intention can also be used to control access to specific routes.
```

blake (approved):
I had two minor suggestions, but otherwise LGTM!

website/content/docs/k8s/connect/ingress-controllers.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+-> The following information requires Consul 1.10+, Consul-k8s 0.26+, Consul-helm 0.32+ configured with
+[Transparent Proxy](/docs/connect/transparent-proxy) mode enabled.
+
+This page describes a general approach for integrating Ingress Controllers with Consul on Kubernetes to secure traffic from the Controller
+to the backend services. This allows Consul to transparently secure traffic from the ingress point through the entire traffic flow of the service.
+
+A few steps are generally required to enable an Ingress controller to join the mesh and pass traffic through to a service:
+
+* Enable connect-injection via an annotation on the Ingress Controller's deployment: `consul.hashicorp.com/connect-inject` is `true`.
```
lkysow, line 19: An example here that shows the annotation must be on the pod spec, not the deployment itself would be helpful

website/content/docs/k8s/connect/ingress-controllers.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+* Using the following annotations on the Ingress controller's deployment, setup exclusion rules for it's ports.
```
lkysow, line 21:
```suggestion
* Using the following annotations on the Ingress controller's deployment, set up exclusion rules for its ports.
```

website/content/docs/k8s/connect/ingress-controllers.mdx
lkysow: I think an example of a real spec here would be immensely useful.

website/content/docs/k8s/connect/ingress-controllers.mdx
```diff
@@ -0,0 +1,79 @@
...
+  * [`consul.hashicorp.com/transparent-proxy-exclude-inbound-ports`](https://www.consul.io/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-inbound-ports) - Provides the ability to exclude a list of ports for
+inbound traffic that the service exposes from redirection. Typical configurations would require all inbound service ports
+for the controller to be included in this list.
+  * [`consul.hashicorp.com/transparent-proxy-exclude-outbound-ports`](https://www.consul.io/docs/k8s/connect#consul-hashicorp-com-transparent-proxy-exclude-outbound-ports) - Provides the ability to exclude a list of ports for
```
lkysow, line 25: So if I was hitting the kube api on `10.0.0.1:443` could I use either `consul.hashicorp.com/transparent-proxy-exclude-outbound-ports: 443` or `consul.hashicorp.com/transparent-proxy-exclude-outbound-cidrs: 10.0.0.1`? Just want to know how exactly this works.

website/content/docs/k8s/connect/ingress-controllers.mdx
```diff
@@ -0,0 +1,79 @@
...
+~> Note: Depending on which ingress controller you use, these stanzas may differ in name and layout, but it is important to apply
+these annotations to the *pods* of your *ingress controller*.
+
+```yaml
+# An example list of pod annotations for an ingress controller.
+  podAnnotations:
+    consul.hashicorp.com/connect-inject: "true"
+    consul.hashicorp.com/transparent-proxy-exclude-inbound-ports: "80,8000,9000,8443"
+    # And the CIDR of your Kubernetes API svc
+    consul.hashicorp.com/transparent-proxy-exclude-outbound-cidrs: "10.108.0.1/32"
```
lkysow, line 41: blurb on how to get this CIDR would be nice. (kubectl get svc kubernetes)

website/content/docs/k8s/connect/ingress-controllers.mdx
lkysow: (the answer is to reduce load on Consul/Envoy by having to share all the pod ips all the time)

website/content/docs/k8s/connect/ingress-controllers.mdx
```diff
@@ -0,0 +1,79 @@
...
+kind: ServiceIntentions
+metadata:
+  name: ingress-backend
+spec:
+  destination:
+    name: backend
+  sources:
+    - name: ingress
+      action: allow
+```
```
lkysow, line 72: These need to be indented to be under the bullets. Be careful because on git commit there's a hook that will reformat. I think it needs to be indented 3 spaces so it doesn't get reformatted.

website/content/docs/k8s/connect/ingress-controllers.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+### Common Configuration Problems:
+- The Ingress Controller's ServiceAccount name and Service name differ by default in some platforms. Consul on Kubernetes requires the
+ServiceAccount and Service to have the same name.
```
lkysow, line 76: What's the resolution for this?

lkysow (approved):
This is an excellent addition. At a high-level I think this is going to be very complicated for people to follow for their specific ingress controller and they will likely get something in their config wrong. As a follow-up I think we must have dedicated docs for at least the NGINX ingress controller.

website/content/docs/k8s/connect/ingress-controllers.mdx
blake: Either would work to enable the Kube API to be directly accessed from the pod. Using `exclude-outbound-ports` would permit access to _any_ destination IP over port 443, whereas `exclude-outbound-cidrs` would permit access to 10.0.0.1 on any destination port.

Here's the rendered iptables rules for reference.

```shell
$ iptables --table nat --list-rules OUTPUT
-P OUTPUT ACCEPT
-A OUTPUT -p tcp -j CONSUL_PROXY_OUTPUT

$ iptables --table nat --list-rules CONSUL_PROXY_OUTPUT
-N CONSUL_PROXY_OUTPUT
-A CONSUL_PROXY_OUTPUT -m owner --uid-owner 996 -j RETURN
-A CONSUL_PROXY_OUTPUT -d 10.0.0.1/32 -j RETURN
-A CONSUL_PROXY_OUTPUT -p tcp -m tcp --dport 443 -j RETURN
-A CONSUL_PROXY_OUTPUT -d 127.0.0.1/32 -j RETURN
-A CONSUL_PROXY_OUTPUT -j CONSUL_PROXY_REDIRECT
```

david-yu:
Thanks @kschoche this PR looks good to merge to me based on most recent changes.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/402389.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 307f49dd319d68ed53717df9837bca9ee0d0d692 onto `stable-website` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 307f49dd319d68ed53717df9837bca9ee0d0d692 onto `release/1.10.x` succeeded!

PR #10497, opened 2021-06-24
tlsutil: unexport and remove methods

dnephin:
This PR makes a number of changes to try and reduce the size of tlsutil, and also make the logic a bit easier to trace by removing some functions.

* unexports a method that was only used internally
* removes methods whose sole purpose is to read a single field ("getter methods") and read the field directly
* removes some duplicate methods
* adds some slightly more descriptive docstrings
* inverts the logic in `outgoingRPCTLSDisabled` to remove double negatives.

### DIFF
--- a/tlsutil/config.go
+++ b/tlsutil/config.go
@@ -537,21 +537,12 @@ func (c *Configurator) VerifyIncomingRPC() bool {
 }
 
 // This function acquires a read lock because it reads from the config.
-func (c *Configurator) outgoingRPCTLSDisabled() bool {
+func (c *Configurator) outgoingRPCTLSEnabled() bool {
 	c.lock.RLock()
 	defer c.lock.RUnlock()
 
-	// if AutoEncrypt enabled, always use TLS
-	if c.base.AutoTLS {
-		return false
-	}
-
-	// if CAs are provided or VerifyOutgoing is set, use TLS
-	if c.base.VerifyOutgoing {
-		return false
-	}
-
-	return true
+	// use TLS if AutoEncrypt or VerifyOutgoing are enabled.
+	return c.base.AutoTLS || c.base.VerifyOutgoing
 }
 
 // MutualTLSCapable returns true if Configurator has a CA and a local TLS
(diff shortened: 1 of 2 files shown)


tlsutil/config_test.go
```diff
@@ -850,14 +850,6 @@ func TestConfigurator_VerifyIncomingRPC(t *testing.T) {
 	require.Equal(t, c.base.VerifyIncomingRPC, verify)
 }
 
-func TestConfigurator_VerifyIncomingHTTPS(t *testing.T) {
-	c := Configurator{base: &Config{
-		VerifyIncomingHTTPS: true,
-	}}
-	verify := c.verifyIncomingHTTPS()
-	require.Equal(t, c.base.VerifyIncomingHTTPS, verify)
-}
```
dnephin, line 859: `IncomingHTTPSConfig` (the one function which uses the replacement of the function this was testing) seems to be missing tests. I'll add some to cover this removal.

tlsutil/config.go
```diff
@@ -537,21 +537,12 @@ func (c *Configurator) VerifyIncomingRPC() bool {
...
 
-	// if AutoEncrypt enabled, always use TLS
-	if c.base.AutoTLS {
-		return false
-	}
-
-	// if CAs are provided or VerifyOutgoing is set, use TLS
-	if c.base.VerifyOutgoing {
-		return false
-	}
```
dnephin, line 563: This comment did not match the logic. I'm not sure if we should be looking at `c.caPool` here or not. But I updated the comment to match the code, assuming the code is right.

i0rek: yeah the code is correct. `VerifyOutgoing` decides what is going on. It is important to be able to load a ca without enabling outgoing TLS for the transition from unencrypted to encrypted communication. Which is why the comment is incorrect.

i0rek (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/398460.

PR #10500, opened 2021-06-25
check expiry date of the root/intermediate before using it to sign a leaf

### LINKED ISSUE #9573: Check validity of certificates before they are used in Consul Connect to sign leaf certs
It is possible for expired intermediate and root certs to be continued to be used to sign leaf certs. Right now, the fact that certs are expired ins't caught until TLS validation at request time. 

This affects both expired root certs that could be used to sign intermediate certs, and expired intermediate certs used to sign leaf certs.

Instead of waiting till request time for TLS verification, we should be checking for cert expiration/validity at certificate signing time. 

TBD - This behavior might be different across different CA providers, needs additional research to figure out which providers are affected.

dhiaayachi:
Fix #9573 
This need to be merged after #10479


agent/consul/leader_connect_ca.go
```diff
@@ -1511,3 +1517,17 @@ func (c *CAManager) SignCertificate(csr *x509.CertificateRequest, spiffeID conne
...
 }
+
+func checkExpired(pem string) error {
+	cert, err := connect.ParseCert(pem)
+	if err != nil {
+		return err
+	}
+	if cert.NotBefore.After(time.Now()) {
+		return fmt.Errorf("certificate start date is in the future %s", cert.NotBefore.String())
+	}
```
dhiaayachi, line 1528: remove the check for the start date

agent/consul/leader_connect_ca.go
```diff
@@ -1451,6 +1451,22 @@ func (c *CAManager) SignCertificate(csr *x509.CertificateRequest, spiffeID conne
 		defer c.caLeafLimiter.csrConcurrencyLimiter.Release()
 	}
 
+	// Append our local CA's intermediate if there is one.
```
kyhavlov, line 1454: This comment line should probably get moved down to 1484 (where the intermediate gets appended)

agent/consul/leader_connect_ca.go
```diff
@@ -1451,6 +1451,22 @@ func (c *CAManager) SignCertificate(csr *x509.CertificateRequest, spiffeID conne
...
+	}
+	root, err := provider.ActiveRoot()
+	if err != nil {
+		return nil, err
+	}
+
+	//check if the intermediate expired before using it to sign
+	err = checkExpired(inter)
+	if err != nil {
+		return nil, err
```
kyhavlov, line 1467: We should mention that this error pertains to the intermediate (or it might not be clear which cert is expired)

agent/consul/leader_connect_ca.go
```diff
@@ -1451,6 +1451,22 @@ func (c *CAManager) SignCertificate(csr *x509.CertificateRequest, spiffeID conne
...
+	inter, err := provider.ActiveIntermediate()
+	if err != nil {
+		return nil, err
+	}
+	root, err := provider.ActiveRoot()
+	if err != nil {
+		return nil, err
+	}
+
+	//check if the intermediate expired before using it to sign
```
kyhavlov, line 1464:
```suggestion
	// Check if the intermediate expired before using it to sign.
```
Tiny formatting nitpick

kyhavlov (commented):
Looking good, just a few comments around errors/formatting

agent/consul/leader_connect_ca.go
```diff
@@ -1451,6 +1451,22 @@ func (c *CAManager) SignCertificate(csr *x509.CertificateRequest, spiffeID conne
 		defer c.caLeafLimiter.csrConcurrencyLimiter.Release()
 	}
```
dhiaayachi, line 1453: check the limiter behaviour  (consul rate limiter)

dhiaayachi: We are rate limited using `CSRMaxPerSecond` and `CSRMaxConcurrent` even in the case of intermediate expired. That said we also should be safe because I don't see any loop that aggressively retry the leaf generation.

agent/consul/leader_connect_ca.go
```diff
@@ -1451,6 +1451,21 @@ func (c *CAManager) SignCertificate(csr *x509.CertificateRequest, spiffeID conne
...
 
+	inter, err := provider.ActiveIntermediate()
+	if err != nil {
+		return nil, err
+	}
+	root, err := provider.ActiveRoot()
+	if err != nil {
+		return nil, err
+	}
```
dnephin, line 1461: I forget if we discussed this already, but do we need to do this `root` lookup here? It is only used later on to compare them. Since we aren't checking the root expiration maybe this can stack where it was?

agent/consul/leader_connect_ca_test.go
```diff
@@ -334,6 +340,98 @@ func TestCAManager_UpdateConfigWhileRenewIntermediate(t *testing.T) {
...
+	args := []struct {
+		testName  string
+		notBefore time.Time
+		notAfter  time.Time
+		isError   bool
+		errorMsg  string
+	}{
+		{"intermediate valid", time.Now().AddDate(0, 0, -1), time.Now().AddDate(0, 0, 2), false, ""},
+		{"intermediate expired", time.Now().AddDate(-2, 0, 0), time.Now().AddDate(0, 0, -1), true, "certificate expired end date"},
+		//we dont' case about a cert in the future
```
dnephin, line 354:
```suggestion
		// a cert that is not yet valid is ok, assume it will be valid soon enough
```

agent/consul/leader_connect_ca_test.go
```diff
@@ -334,6 +340,98 @@ func TestCAManager_UpdateConfigWhileRenewIntermediate(t *testing.T) {
...
+				Bytes: x509.MarshalPKCS1PrivateKey(caPrivKey),
+			})
+			manager.providerShim = &mockCAProvider{
+				callbackCh: delegate.callbackCh,
+				rootPEM:    caPEM.String(),
+			}
+			initTestManager(t, manager, delegate)
+
+			// Wait half the TTL for the cert to need renewing.
+			time.Sleep(1000 * time.Millisecond)
```
dnephin, line 418: Any way we could avoid sleeping a while second in each test case?

Maybe we could add a `var timeNow = time.Now` shim for testing. `checkExpired` could use that var instead of calling `time.Now` directly, and then this test could patch the value of now, similar to how we patch the TTL?

dnephin (commented):
Nice! I think this is looking good. One question and one suggestion for speeding up the test.

agent/consul/leader_connect_ca.go
dhiaayachi: Actually you bring a good point, I'm not sure about that and may be we need input from other folks but root expiration need to be checked as well. My understanding is that any cert in the cert chain that expire invalidate the whole chain.

dhiaayachi: After discussion with @banks he confirmed my thoughts. I added the root expiry check.

freddygv (approved):
LGTM, left one minor suggestion

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/408262.

PR #10504, opened 2021-06-25
Add ca certificate metrics

dhiaayachi:
The PR is to add metrics around ca certificates. Ideally for each available certificate:
- Seconds until it expire
- Status of fetching it or creating (success/error)

For the time being this apply to root CA certs and intermediate CA certs, for leaf certs it's more challenging as we don't keep a local copy of the certs in the agent (need to be confirmed)


.changelog/10504.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
+added metrics to track certificates expiry.
```
dnephin, line 2:
```suggestion
telemetry: added metrics to track certificates expiry.
```

agent/consul/leader_connect.go
```diff
@@ -46,6 +47,8 @@ func (s *Server) stopConnectLeader() {
 	s.caManager.Stop()
 	s.leaderRoutineManager.Stop(intentionMigrationRoutineName)
 	s.leaderRoutineManager.Stop(caRootPruningRoutineName)
+	s.leaderRoutineManager.Stop(caRootMetricRoutineName)
```
dnephin, line 50: Thank you for adding this thing I missed!

agent/consul/leader_metrics.go
```diff
@@ -3,6 +3,7 @@ package consul
 import (
 	"context"
 	"fmt"
+	"github.com/hashicorp/consul/agent/connect"
```
dnephin, line 6: Do you use `goimports` to fix the imports? 

I have GoLand configured with a File Watcher that runs `goimports -w -local github.com/hashicorp/consul` so that it formats the imports in the correct order. It's not perfect, but it works most of the time. We could also look at one of the newer tools (ex: https://github.com/daixiang0/gci) which might be more reliable.

I think this import should be in the third block of imports.

agent/consul/leader_metrics.go
```diff
@@ -12,18 +13,25 @@ import (
 	"github.com/hashicorp/consul/logging"
 )
 
+var metricsKeyMeshCAExpiry = map[string][]string{
+	"root-cert":         {"mesh", "active-root-ca", "expiry"},
+	"intermediate-cert": {"mesh", "active-intermediate-ca", "expiry"},
+}
```
dnephin, line 19: I think a couple vars might work better than a map here. Since we have to use a string key for a map we don't get the benefits of the compiler checking correctness for us:

```suggestion

var metricsKeyMeshRootCAExpiry = []string{"mesh", "active-root-ca", "expiry"}
var metricsKeyMeshIntermediateCAExpiry = []string{"mesh", "active-intermediate-ca", "expiry"}

```

agent/consul/leader_metrics.go
```diff
@@ -43,6 +51,28 @@ func rootCAExpiryMonitor(s *Server) certExpirationMonitor {
...
+func intermediateCAExpiryMonitor(s *Server) certExpirationMonitor {
+	return certExpirationMonitor{
+		Key: metricsKeyMeshCAExpiry["intermediate-cert"],
+		Labels: []metrics.Label{
+			{Name: "datacenter", Value: s.config.Datacenter},
+		},
+		Logger: s.logger.Named(logging.Connect),
+		Query: func() (time.Duration, error) {
+			provider, _ := s.caManager.getCAProvider()
+			certPem, err := provider.ActiveIntermediate()
```
dnephin, line 63: I need to look into this more next week. The current implementation of the active-root-ca metric uses the state store to fetch the certificate, where as this is using the provider. I'm not sure if those could ever return different answers, or what the implications are of using one or the other. 

I'd like to understand that, and possibly have both metrics use the same source of data.

Despite the name of the method, `CARootActive` actually returns a `struct.CARoot` which also has a slice of Intermediate certs on it. So we could get both sets of data from the same source, I'm not sure which is more correct.

dnephin (commented):
Thanks for adding this! I think this is looking good, but I'd like to understand the data sources for these certs a bit better.

agent/consul/leader_metrics.go
dhiaayachi: Yes I have the same questions. I was also tempted to change the root metric to use the same data source for the root, also we need to handle the case were we use a Vault/AWS provider. In that case the root CA is not ours and we use an intermediate. My understanding is that the current root metric won't be accurate in that case.

agent/consul/leader_metrics.go
dhiaayachi: I added `goimports` as a watcher, let me see how reliable it is before trying another tool.

agent/consul/leader_metrics.go
dhiaayachi: I changed to use the same way to access data

agent/consul/leader_metrics.go
```diff
@@ -43,6 +58,83 @@ func rootCAExpiryMonitor(s *Server) certExpirationMonitor {
...
+	if err != nil {
+		return nil, err
+	}
+
+	// the CA used in a secondary DC is the active intermediate,
+	// which is the last in the IntermediateCerts stack
+	if len(root.IntermediateCerts) == 0 {
+		return nil, errors.New("no intermediate available")
+	}
+	cert, err := connect.ParseCert(root.IntermediateCerts[len(root.IntermediateCerts)-1])
```
dnephin, line 131: I'm not familiar enough with this code, but I wonder how safe this assumption is (that the last cert will be the important one). This may be fine just calling out as something I need to look into more to confirm.

dhiaayachi: I went through all the code that generate an intermediate and modify this and it always append to it. So the most recent cert will be always the last and this is the one we want to expose.

agent/consul/leader_metrics.go
```diff
@@ -43,6 +58,77 @@ func rootCAExpiryMonitor(s *Server) certExpirationMonitor {
...
+			state := s.fsm.State()
+			_, root, err := state.CARootActive(nil)
+			switch {
+			case err != nil:
+				return 0, fmt.Errorf("failed to retrieve root CA: %w", err)
+			case root == nil:
+				return 0, fmt.Errorf("no active root CA")
+			}
+
+			return time.Until(root.NotAfter), nil
```
dnephin, line 89: It might be nice to extract this as a function, and use it both here and in the existing active-roots metric, to make it more obvious that this metric reports the same value as the active-root metric in some cases.

agent/consul/leader_metrics.go
```diff
@@ -2,25 +2,40 @@ package consul
...
-		Help: "Seconds until the service mesh root certificate expires.",
+		Help: "Seconds until the service mesh root certificate expires. Updated every hour",
+	},
+	{
+		Name: metricsKeyMeshPrimaryCAExpiry,
+		Help: "Seconds until the service mesh primary DC certificate expires. Updated every hour",
+	},
+	{
+		Name: metricsKeyMeshSecondaryCAExpiry,
+		Help: "Seconds until the service mesh secondary DC certificate expires. Updated every hour",
```
dnephin, line 35: I think the only question I have left is if we should report two separate metrics, or a single one called `active-signing-ca`. I think either could work.The datacenter label should work to differentiate the values from different DCs.

I'm starting to like the idea of using a single metric. I expect it will be a little easier to set up monitoring if both primary and secondary use the same metric name. It might be worth it to see what others think about these two options.

dnephin (commented):
Thanks! Just one question/thought about using a single metric vs separate ones.

agent/consul/leader_metrics.go
dhiaayachi: Yes I think you're right a single metric that would return the right expiry will be easier to integrate and expose the user less to the implementation details
I will make the change

dnephin (approved):
LGTM!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/405375.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 6390e91be544c716f966924922d878f69360a70e onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/405375)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 6390e91be544c716f966924922d878f69360a70e onto `release/1.9.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/405375)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 6390e91be544c716f966924922d878f69360a70e onto `release/1.8.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/405375)

Issue #10250, opened 2021-05-17
rpc: add RPC method label to rpc.request metric


dnephin:
`consul.rpc.request` (and a a few others) provide metrics for the count of RPC requests, however none of them provide enough granularity to debug most common issues. There are many RPC methods, and an elevated count of requests doesn't help identify the source of the problem.

We should either add labels to some of the existing metrics, or add new metrics which those labels. An ideal label would be the full method name (ex: `Catalog.Register`) however if that is too high cardinality then simply `Catalog` would be an improvement, or one label for type (ex: `Catalog`) and one for operation (ex: `read` or `write`). 

While there are a large number of RPC methods, must clusters will only use a small number of those endpoints, so a label for method name might be ok, despite it being potentially high cardinality.

PR #10506, opened 2021-06-25
docs: correct some misleading telemetry docs

dnephin:
The query metrics are actually reported for all read queries, not only ones that use a `MinIndex` to block for updates.

Also clarify the raft.apply metric is only on the leader.

Issue #10250 would greatly improve these metrics, and would likely allow us to deprecate the misleading metrics about blocking queries.


rboyer (approved):
LGTM

website/content/docs/agent/telemetry.mdx
```diff
@@ -421,8 +421,8 @@ These metrics are used to monitor the health of the Consul servers.
...
+| `consul.rpc.queries_blocking`                       | The current number of in-flight read queries the server is handling.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       | queries                           | gauge   |
```
banks, line 425: Perhaps we could elaborate on this. The _value_ here is that for the most part this will be dominated by actually blocking requests at any given sample point rather than the in-flight reads that will return immediately.

So this is technically more accurate, but the distinction being made in this PR seems pretty subtle. Perhaps it would help operators interpret if we were more explicit? Maybe something like:

```suggestion
| `consul.rpc.queries_blocking`                       | The current number of in-flight read queries the server is handling. This counts all in-flight read requests whether they are blocked on updates or not, but in general will be dominated by blocked requests and can be used to monitor overall blocking query load on the server.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           | queries                           | gauge   |
```

dnephin: > for the most part this will be dominated by actually blocking requests at any given sample point rather than the in-flight reads that will return immediately

I don't think that is always a fair assumption. Some times it will be true, but it's also very possible that the server is handling many non-blocking queries and the few blocking queries are a small percentage of this value.

Instead of attempting to explain the metric maybe we should fix it? I think we could only report the metric when `MinQueryIndex > 0`. I'll make that change in this PR, and restore the text for `queries_blocking`.

dnephin: Oh, actually looking again, `queries_blocking` is correct, I was mislead by the `GOTO` I think. So I'll revert the docs changes to `queries_blocking`.

dnephin:
I reverted the changes to `queries_blocking`. That metric was already documented correctly.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/398878.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 2e1e80266aca3895e87912c67fdde147dbaa66c6 onto `stable-website` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 2e1e80266aca3895e87912c67fdde147dbaa66c6 onto `release/1.10.x` succeeded!

PR #10396, opened 2021-06-14
Fix some data races

dnephin:
Related to #8329
Fixes #9458

These commits are extracted from #10341. There's still a lot more work required to properly solve #9457 (ex: https://github.com/hashicorp/memberlist/pull/238), so I'd like to get some of the data race fixes merged to prevent them from rotting, and so that contributors can run the race detector locally with much less noise.

PR #10514, opened 2021-06-28
streaming: fix not being able to enable streaming

dnephin:
This PR adds a new `X-Consul-Query-Backend` response header to make it easier to determine when streaming was used for a query. Adds a check for this header to a few tests that are meant to check that streaming is being used.

And finally fix the bug that was preventing streaming from being enabled on the client.

One commit is cherry-picked from #10396. It removes a test that is both very racy and also no longer valid because the requests are made via streaming, not the cache. Since the cache rate limit functionality is already tested in the appropriate package (agent/cache) it seems like we can remove this failing/flaky test.

### DIFF
--- a/.changelog/10514.txt
+++ b/.changelog/10514.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+streaming: fix a bug that was preventing streaming from being enabled.
+```
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -394,6 +394,7 @@ func New(bd BaseDeps) (*Agent, error) {
 			Conn:   conn,
 			Logger: bd.Logger.Named("rpcclient.health"),
 		},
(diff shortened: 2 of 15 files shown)


banks (approved):
LGTM. Thanks Daniel!

banks (approved):
I see, this test is no longer valid now that streaming is actually default and so cache is not being used.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/399092.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 1041ba71de91fd87dc6213b9422e87a812759846 onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/399092)

PR #10546, opened 2021-07-01
Add support for returning ACL secret IDs for accessors with acl:write

eculver:
In order to support copying of secret IDs from the UI, we need to include them in the API response with the caveat of when the accessor has only has `acl:read` in which case the redacted secret ID will be returned instead.


command/acl/token/formatter_test.go
```diff
@@ -155,6 +155,7 @@ func TestFormatTokenList(t *testing.T) {
 			tokens: []*api.ACLTokenListEntry{
 				{
 					AccessorID:  "fbd2447f-7479-4329-ad13-b021d74f86ba",
+					SecretID:    "257ade69-748c-4022-bafd-76d27d9143f8",
```
mkeeler, line 158: I think the command/acl/token/formatter.go also needs updating to output the SecretIDs if they are not redacted. Then the golden files for the `pretty` format should also be updated.

mkeeler (commented):
Functionally this looks to do exactly what it set out to and is well done. There is one small bit missing from the implementation to get the `pretty` cli output format to show the secrets for the `consul acl token list` command but otherwise things look great.

I started to write a small bit about whether we need the `ACLTokenListStub` type anymore since it holds pretty much everything the full `ACLToken` does. While that is true I am not sure we could just get rid of the type and have the `ACL.TokenList` RPC return a `[]ACLToken` instead due to upgrade concerns. It might be possible but I am not positive.

As a side note I noticed a bug that I think I introduced a while back. Essentially the `ACLAuthMethodEnterpriseMeta` field from the token isn't being propagated to the stub. Not something to fix in this PR but its one of the things that got me thinking we would be better off without the dual types.

mkeeler:
The issue for the extra bug in the token list stub conversion is here: https://github.com/hashicorp/consul/issues/10557

eculver:
> The issue for the extra bug in the token list stub conversion is here: #10557

Ok, I can take a stab it this next.

command/acl/token/formatter_test.go
eculver: I added the `SecretID` to the formatter. The only caveat is that the `SecretID` field is always shown rather than being omitted when it is redacted. This matches the API's behavior but wanted to make sure this is what we want for the CLI too vs only showing it when it's not redacted (as you mentioned above).

mkeeler: Does that mean that it will show the secret ID as `<hidden>`? If so that seems fine. There probably isn't a good reason to hide the existence of the field all together in that case.

mkeeler (approved):
LGTM :shipit:

freddygv (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/406153.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 13bd86527bcba8bc25a6086688c3e02ec04ea7a3 onto `release/1.10.x` succeeded!

PR #10513, opened 2021-06-28
ui: Add intl debug helpers

johncowen:
We made a few PRs lately just trying to make our UI documentation and overall UI DX better. Having good documentation and developer experience make it much for myself and others to add new features to the UI.

Back in https://github.com/hashicorp/consul/pull/9399 we added `ember-intl` we've started slowly adding i18n support throughout the app. We don't translate anything just now at least, but its general good practice to externalize all of this copy so its easy for others to find and edit.

This PR adds a couple of debug utilities to help us to continue doing this:

1. We've added a `CONSUL_INTL_DEBUG` env/cookie variable to turn off variable interpolation within the `t` helper so you can see which variables are being interpolated.
2. We've added a `CONSUL_INTL_LOCALE` env/cookie which currently supports two 'pseudo-locales' - `la-fk` (fake latin) and `-` (just dashes) either of which will make it easier to see what has not been localized until we can add prettier rules to prevent adding any copy into templates at all. I would guess if we ever translated the app we would use this for looking at things whilst developing also - but as yet I've not adding anything for that here seeing as we don't translate anything.

Both variables are dev-time only and all code for this is removed from the production build.

`la-fk`

Screenshot 2021-06-28 at 14 54 29

`-`

Screenshot 2021-06-28 at 14 54 15

PR #10561, opened 2021-07-06
ui: [Backport 1.10.x] Add intl debug helpers #10513

johncowen:
This commit adds a couple of debug utilities to help us to continue slowly adding i18n support:

- We've added a CONSUL_INTL_DEBUG env/cookie variable to turn off variable interpolation within the t helper so you can see which variables are being interpolated.
- We've added a CONSUL_INTL_LOCALE env/cookie which currently supports two 'pseudo-locales' - la-fk (fake latin) and - (just dashes) either of which will make it easier to see what has not been localized until we can add prettier rules to prevent adding any copy into templates at all. I would guess if we ever translated the app we would use this for looking at things whilst developing also - but as yet I've not adding anything for that here seeing as we don't translate anything.
Both variables are dev-time only and all code for this is removed from the production build.

### DIFF
--- a/ui/packages/consul-ui/app/instance-initializers/debug.js
+++ b/ui/packages/consul-ui/app/instance-initializers/debug.js
@@ -1,4 +1,9 @@
 import ApplicationRoute from '../routes/application';
+import { I18nService, formatOptionsSymbol } from './i18n';
+import ucfirst from 'consul-ui/utils/ucfirst';
+
+import faker from 'faker';
+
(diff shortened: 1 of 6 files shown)


kaxcode (approved):
:shipit:

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/405222.

PR #10564, opened 2021-07-06
cli/sdk: Allow applying redirect-traffic rules in a provided Linux namespace

ishustava:
This PR proposes to apply traffic redirection rules in a given Linux network namespace:
- The `iptables` SDK package now accepts `NetNS` configuration which should point to the path of the network namespace.
- The `redirect-traffic` command accepts `netns` flag with the same.

**Note:** I did not find a meaningful enough way to unit test this change since it's just wrapping every `iptables` command with a call to `nsenter`, and that's why there aren't tests added. I did test it manually on a VM:

```
root@test2:/home/irynashustava# ip netns add foo
root@test2:/home/irynashustava# ip netns
foo
root@test2:/home/irynashustava# ./consul connect redirect-traffic -netns=/var/run/netns/foo -proxy-uid=1234 -proxy-inbound-port=15001
    Successfully applied traffic redirection rules
root@test2:/home/irynashustava# nsenter --net=/var/run/netns/foo -- iptables -t nat --list
Chain PREROUTING (policy ACCEPT)
target     prot opt source               destination
CONSUL_PROXY_INBOUND  tcp  --  anywhere             anywhere

Chain INPUT (policy ACCEPT)
target     prot opt source               destination

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination
CONSUL_PROXY_OUTPUT  tcp  --  anywhere             anywhere

Chain POSTROUTING (policy ACCEPT)
target     prot opt source               destination

Chain CONSUL_PROXY_INBOUND (1 references)
target     prot opt source               destination
CONSUL_PROXY_IN_REDIRECT  tcp  --  anywhere             anywhere

Chain CONSUL_PROXY_IN_REDIRECT (1 references)
target     prot opt source               destination
REDIRECT   tcp  --  anywhere             anywhere             redir ports 15001

Chain CONSUL_PROXY_OUTPUT (1 references)
target     prot opt source               destination
RETURN     all  --  anywhere             anywhere             owner UID match 1234
RETURN     all  --  anywhere             localhost
CONSUL_PROXY_REDIRECT  all  --  anywhere             anywhere

Chain CONSUL_PROXY_REDIRECT (1 references)
target     prot opt source               destination
REDIRECT   tcp  --  anywhere             anywhere             redir ports 15001
```

### DIFF
--- a/command/connect/redirecttraffic/redirect_traffic.go
+++ b/command/connect/redirecttraffic/redirect_traffic.go
@@ -44,6 +44,7 @@ type cmd struct {
 	excludeOutboundPorts []string
 	excludeOutboundCIDRs []string
 	excludeUIDs          []string
+	netNS                string
 }
 
 func (c *cmd) init() {
@@ -62,6 +63,8 @@ func (c *cmd) init() {
 		"Outbound CIDR to exclude from traffic redirection. May be provided multiple times.")
 	c.flags.Var((*flags.AppendSliceValue)(&c.excludeUIDs), "exclude-uid",
 		"Additional user ID to exclude from traffic redirection. May be provided multiple times.")
+	c.flags.StringVar(&c.netNS, "netns", "", "The network namespace where traffic redirection rules should apply."+
+		"This must be a path to the network namespace, e.g. /var/run/netns/foo.")
 
 	c.http = &flags.HTTPFlags{}
 	flags.Merge(c.flags, c.http.ClientFlags())
@@ -130,6 +133,7 @@ func (c *cmd) generateConfigFromFlags() (iptables.Config, error) {
 		ProxyUserID:       c.proxyUID,
 		ProxyInboundPort:  c.proxyInboundPort,
 		ProxyOutboundPort: c.proxyOutboundPort,
+		NetNS:             c.netNS,
 	}
 
 	// When proxyID is provided, we set up cfg with values
--- a/.changelog/10564.txt
+++ b/.changelog/10564.txt
@@ -1,0 +1,6 @@
+```release-note:feature
+cli: allow running `redirect-traffic` command in a provided Linux namespace.
+```
+```release-note:feature
+sdk: allow applying `iptables` rules in a provided Linux namespace.
+```
--- a/sdk/iptables/iptables.go
+++ b/sdk/iptables/iptables.go
@@ -6,16 +6,16 @@ import (
 )
 
 const (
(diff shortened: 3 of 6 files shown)


command/connect/redirecttraffic/redirect_traffic.go
```diff
@@ -62,6 +63,8 @@ func (c *cmd) init() {
 		"Outbound CIDR to exclude from traffic redirection. May be provided multiple times.")
 	c.flags.Var((*flags.AppendSliceValue)(&c.excludeUIDs), "exclude-uid",
 		"Additional user ID to exclude from traffic redirection. May be provided multiple times.")
+	c.flags.StringVar(&c.netNS, "netns", "", "The network namespace where traffic redirection rules should apply."+
+		"This must be a path to the network namespace, e.g. /var/run/netns/foo.")
```
freddygv, line 67: When would the network namespace in linux not be located in `/var/run`? Wondering if we could assume the path in the executor for simplicity.

It seems that when running commands like `ip netns exec foo <command>` you don't need to specify the path to the file.

freddygv (approved):
Awesome, thanks for getting this done!

Tested it out locally and it works as expected. I added one non-blocking question below.

command/connect/redirecttraffic/redirect_traffic.go
blake: Namespaces created from `ip netns` will have an entry under `/var/run/netns`, but namespaces created via other means are not guaranteed to be listed here. See https://bunn.cc/2017/nsenter_network_namespace/ for a better explanation and alternate solution for execing into a namespace.

ishustava: What Blake said! For example, on Kube, it'll live under `/run/netns` as well as under `/var/run/netns`. I think we'd need to support a different path because also that's what the [CNI](https://github.com/containernetworking/cni/blob/master/SPEC.md#attachment-parameters) (container networking interface) does.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/408104.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 95305881ce6cbe1f86b09b14540f0fcabd7b88f0 onto `release/1.10.x` succeeded!

PR #10574, opened 2021-07-07
telemetry: add a new rpc.call metric

### LINKED ISSUE #10250: rpc: add RPC method label to rpc.request metric

dnephin:
Closes #10250

This PR adds a new `rpc.call` metric. This metric should be able to replace both `rpc.request`, and `rpc.query`, as those metrics don't really provide enough data to answer most operational questions. Most of the diff is just updating vendor, so probably best viewed by individual commit.

This new metric is a timing metric, but most metric systems will also give you a count along with the timing percentiles. The only disadvantage to this approach is that the metric is not reported until the request ends. If this is a problem we could also add a count metric before performing the request, with mostly the same set of labels.

The new labels should allow us to answer many questions:
* how many requests are being handled on a given server, vs being forward to another DC, or forwarded to the leader?
* what is the error rate?
* which RPC methods are being called?

In order to implement this metric I had to create a fork of `net/rpc`, since the stdlib has frozen that library since 2016. You can see the changes I made to `net/rpc` here: https://github.com/hashicorp/go-msgpack/compare/91a94c4...hashicorp:dnephin/fork-net-rpc.  I put the fork into `go-msgpack` for now because both it and `net-rpc-msgpackrpc` must be updated to point at the same fork. They all reference the `net/rpc.Request` and `net/rpc.Response` types. I tried using a type alias for those two structs, but unfortunately the `net/rpc.Server` references an unexported field on these structs, so the type alias did not work.

This change was tested by running a fake statsd instance using `socat -d - udp6-listen:8125`, and running Consul built from this branch with `consul agent -dev -hcl='telemetry {statsd_address = "localhost:8125"}'`.

Questions:
* Do these labels seem sufficient? Any other labels that might be important?
* Should we also add a count metric or is waiting for the request to end ok in most cases?

TODO:
* [ ] get the `net/rpc` fork into the main branch of some library
* [ ] merge updates to `go-msgpack` and `net-rpc-msgpackrpc` to their main branches
* [ ] update the dependencies and vendor in this PR to point at released versions of those libraries
* [ ] deprecate some of the replaced rpc metrics
* [ ] add the new metric to the consul telemetry docs
* [ ] changelog entry


agent/consul/rpc.go
```diff
@@ -949,3 +952,58 @@ func (s *Server) consistentRead() error {
...
+	}
+	role := "follower"
+	if r.srv.IsLeader() {
+		role = "leader"
+	}
+	labels := []metrics.Label{
+		{Name: "method", Value: serviceMethod},
+		{Name: "datacenter", Value: r.srv.config.Datacenter},
+		{Name: "server", Value: role},
+		{Name: "error", Value: failed},
```
FFMMM, line 977: maybe using 0, 1 here would help instrument

agent/consul/rpc.go
```diff
@@ -949,3 +952,58 @@ func (s *Server) consistentRead() error {
...
+	}
+	if td, ok := argvi.(targetDC); ok {
+		labels = append(labels, metrics.Label{Name: "target-datacenter", Value: td.RequestDatacenter()})
+	}
+	metrics.MeasureSinceWithLabels(metricRPCRequest, start, labels)
+}
+
+var metricRPCRequest = []string{"rpc", "call"}
+
+var _ rpc.Interceptor = rpcMetricsInterceptor{}
```
FFMMM, line 1000: Thinking if it may be worth to break down interceptors by classes.

agent/consul/rpc.go
```diff
@@ -949,3 +952,58 @@ func (s *Server) consistentRead() error {
...
+	// TODO: accept a smaller interface
+	srv *Server
+}
+
+func (r rpcMetricsInterceptor) ServiceCall(fn func(), serviceMethod string, argv, replyv reflect.Value) {
+	start := time.Now()
+	fn()
+
+	failed := "false"
+	if _, ok := replyv.Interface().(error); ok {
```
FFMMM, line 966: i'm just leaving this as a note to myself:

`replyv` here might actually come out as a `*net.rpc.Reply` (ptr) ; so `.(error)` type conversion may panic/ not work as intended.

FFMMM:
Hey @dnephin thanks loads for trail blazing here. At this stage, a lot of this project and work done here has been completed 🥳 . 

I'll close this PR now.

PR #10576, opened 2021-07-08
tests: skip cases that depend on test server when -short flag set

### LINKED ISSUE #10575: testing: Skip cases that depend on the test server when invoked with `go test -short`
#### Feature Description

Many of the API and CLI tests depend on a running consul instance. In order to speed up the suite (and make `go test -short` more useful, we should add support for skipping those tests conditionally.

#### Use Case(s)

Faster, simpler test runs. For the 32-bit test cases that use `-short`

eculver:
For cases where `go test -short` is used to invoke tests, we want to skip those that involve invoking the `consul` binary. 

Fixes #10575.

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -61,6 +61,11 @@ func makeClientWithConfig(
 	var server *testutil.TestServer
 	var err error
 	retry.RunWith(retry.ThreeTimes(), t, func(r *retry.R) {
+		// Skip test when -short flag provided; any tests that create a test
+		// server will take at least 100ms which is undesirable for -short
+		if testing.Short() {
+			t.Skip("too slow for testing.Short")
+		}
 		server, err = testutil.NewTestServerConfigT(t, cb2)
 		if err != nil {
 			r.Fatalf("Failed to start server: %v", err.Error())
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -50,6 +50,11 @@ func makeClientWithConfig(
 	t *testing.T,
 	cb1 configCallback,
 	cb2 testutil.ServerConfigCallback) (*Client, *testutil.TestServer) {
+	// Skip test when -short flag provided; any tests that create a test
+	// server will take at least 100ms which is undesirable for -short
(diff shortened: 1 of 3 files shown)


command/connect/redirecttraffic/redirect_traffic_test.go
```diff
@@ -549,6 +549,12 @@ func TestGenerateConfigFromFlags(t *testing.T) {
 		t.Run(c.name, func(t *testing.T) {
 			cmd := c.command()
 			if c.consulServices != nil {
+				// Skip test when -short flag provided; any tests that create a test server
+				// will take at least 100ms which is undesirable for -short
+				if testing.Short() {
+					t.Skip("too slow for testing.Short")
+				}
```
dnephin, line 556: I believe this function `TestGenerateConfigFromFlags` already has this `t.Skip` logic as the first line of the test. I believe it was only added recently, but it seems to be working. So I suspect adding the skip here has no effect.

api/api_test.go
```diff
@@ -61,6 +61,11 @@ func makeClientWithConfig(
 	var server *testutil.TestServer
 	var err error
 	retry.RunWith(retry.ThreeTimes(), t, func(r *retry.R) {
+		// Skip test when -short flag provided; any tests that create a test
+		// server will take at least 100ms which is undesirable for -short
+		if testing.Short() {
+			t.Skip("too slow for testing.Short")
+		}
```
dnephin, line 68: Generally we put these as the first line to avoid doing any work if the test will be skipped. Here it's in a retry. I think it might be better placed as the first line of `makeClientWithConfig`.

dnephin (commented):
Thank you for adding these! I left a couple suggestions for the placement

dhiaayachi (approved):
LGTM!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/406726.

PR #10581, opened 2021-07-09
ui: Split up the socket mode from the socket path

kaxcode:
When there is no `LocalBindSocketMode`:
Screen Shot 2021-07-09 at 10 54 47 AM

With  a `LocalBindSocketMode`:
Screen Shot 2021-07-09 at 10 54 59 AM

### DIFF
--- a/ui/packages/consul-ui/app/components/consul/upstream-instance/list/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/upstream-instance/list/index.hbs
@@ -40,19 +40,19 @@
       {{#if item.LocalBindSocketPath}}
 {{#let (if item.LocalBindSocketMode
           (hash
-            label=(concat '(Local bind mode)')
-            value=(concat '(Mode:' item.LocalBindSocketMode ')')
+            label='Local bind mode'
+            value='item.LocalBindSocketMode'
           )
           (hash
-            label=""
-            value=""
+            label='Local bind mode'
+            value="-"
           )
         )
 as |mode|}}
           <dl class="local-bind-socket">
             <dt>
               <span>
-                Local bind socket {{mode.label}}
+                Local bind socket
               </span>
             </dt>
             <dd>
@@ -61,6 +61,16 @@ as |mode|}}
                 @name="Socket path"
               />
               {{item.LocalBindSocketPath}}
+            </dd>
+          </dl>
+          <dl class="local-bind-mode">
+            <dt>
+              <span>
+                Local bind mode
+              </span>
+            </dt>
+            <dd>
+              <span>mode:</span> {{mode.value}}
             </dd>
           </dl>
 {{/let}}
--- a/ui/packages/consul-ui/app/components/consul/upstream-instance/list/index.scss
+++ b/ui/packages/consul-ui/app/components/consul/upstream-instance/list/index.scss
@@ -11,4 +11,7 @@
   dl.nspace dt::before {
     @extend %with-folder-outline-mask, %as-pseudo;
   }
+  dl.local-bind-mode dd span {
+    font-weight: $typo-weight-semibold;
+  }
 }


johncowen (changes requested):
Looking at the screengrab things look a little out to me, not sure what's going on there?

kaxcode:
> Looking at the screengrab things look a little out to me, not sure what's going on there?

Yea, I'm not sure. I'm not making any changes to spacing in this PR though. @johncowen

kaxcode:
@johncowen fixed the issue and update the screenshot.

ui/packages/consul-ui/app/components/consul/upstream-instance/list/index.hbs
```diff
@@ -40,19 +40,19 @@
...
-            label=(concat '(Local bind mode)')
-            value=(concat '(Mode:' item.LocalBindSocketMode ')')
+            label='Local bind mode'
+            value=item.LocalBindSocketMode
           )
           (hash
-            label=""
-            value=""
+            label='Local bind mode'
+            value="-"
```
johncowen, line 48: I don't think we need to do any of this `let`ing now we are always showing a label. We should remove this and just do it inline instead.

ui/packages/consul-ui/app/components/consul/upstream-instance/list/index.hbs
```diff
@@ -63,6 +63,16 @@ as |mode|}}
...
             </dd>
           </dl>
+          <dl class="local-bind-mode">
+            <dt>
+              <span>
+                Local bind mode
+              </span>
+            </dt>
+            <dd>
+              <span>mode:</span> {{mode.value}}
```
johncowen, line 73: Maybe check in with design, but I think we generally uppercase the first letter of field/property names

johncowen (commented):
Thanks for sorting that issue and adding the updated screengrab.

I've added another couple of tiny comments.

Apart from that, I'm ok for you to look at the spacing here or in another PR. It's completely up to you, but from what I remember back when the team decided to squash commits for PRs, it was so we don't end up adding lots of tiny commits to the git history that probably weren't relevant to folks. I think it's handy for everything to go in under the same meaningful change/changelog but either way here I think is fine.

I wrote some docs on how to style these things here https://consul-ui-staging-hashicorp.vercel.app/ui/docs/components/horizontal-kv-list if it helps, if you hit any problems doing that let me know, I'd be happy to go through together.

ui/packages/consul-ui/app/components/consul/upstream-instance/list/index.hbs
kaxcode: Refactored!

ui/packages/consul-ui/app/components/consul/upstream-instance/list/index.hbs
kaxcode: According to the designers we follow a pattern of lowercase field names. Ex. Exposed Paths
Screen Shot 2021-07-16 at 9 54 37 AM

Following the same pattern here. No changes made.

ui/packages/consul-ui/app/components/consul/upstream-instance/list/index.hbs
```diff
@@ -63,7 +52,18 @@ as |mode|}}
...
+          <dl class="local-bind-mode">
+            <dt>
+              <span>
+                Local bind mode
+              </span>
+            </dt>
+            <dd>
+              <span>mode:</span>{{or item.LocalBindSocketMode '-'}}
+            </dd>
+          </dl>
```
johncowen, line 65: Looks like we have two titles here? `Local bind mode` and `mode`, could we just use one?

ui/packages/consul-ui/app/components/consul/upstream-instance/list/index.hbs
```diff
@@ -63,7 +52,18 @@ as |mode|}}
...
-{{/let}}
+        {{#if item.LocalBindSocketMode}}
+          <dl class="local-bind-mode">
+            <dt>
+              <span>
+                Local bind mode
+              </span>
+            </dt>
+            <dd>
+              <span>mode:</span>{{or item.LocalBindSocketMode '-'}}
```
johncowen, line 63: According to the code here, we will never show a `-`?

ui/packages/consul-ui/app/components/consul/upstream-instance/list/index.hbs
kaxcode: yea, I'll just use `mode`.

ui/packages/consul-ui/app/components/consul/upstream-instance/list/index.hbs
kaxcode: Oof, yea. I'll add another {{else}} block.

johncowen: I think here we can just remove the `if` entirely? I'm not 100% sure tho?

ui/packages/consul-ui/app/components/consul/upstream-instance/list/index.hbs
johncowen: Not sure if I'm making sense here, with that additional change we still have two titles `Mode` and `mode`, and I think we can keep the `or` and lose a conditional.

> I think here we can just remove the if entirely?

So think of it like this maybe:

```javascript
if(primaryProperty) {
  console.log(`Primary Property: ${primaryProperty}`);
  console.log(`Secondary Property: ${secondaryProperty || '-'}`);
}
```

which could output:

```sh
Primary Property: Hello
Secondary Property: -
# or
Primary Property: Hello
Secondary Property: There
# or
..nothing
```

Happy to sync up on this if thats easier?

johncowen (commented):
Can't see my last comment in the Conversation view, I tried un-resolving it but that didn't effect anything, here's a link to it instead

https://github.com/hashicorp/consul/pull/10581/files#r673248452

🤞 it shows up in the Convo

johncowen (approved):
Awesome, I think this looks great! I think the only thing that may help this PR now would be to make another edit to the description to add a new screengrab to show how we finished it up.

Other than that LGTM! :shipit:

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/425533.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit ab6a675209af9acd8a7e5f5bc3863501c0455771 onto `release/1.10.x` succeeded!

PR #10588, opened 2021-07-09
config: rename `ports.grpc` to `ports.xds`

dnephin:
Now that we have gRPC endpoints (streaming subscribe, and a new one coming in enterprise) that use the regular RPC port it seems likely that we will continue to use the RPC port for any "regular" RPC endpoints.

However, our config has a `ports.grpc` and `addresses.grpc` that apply only to the xDS interface. Since the consumer of the xDS interface is quite different from the RPC interface (expected to be localhost, and a Envoy proxy), it seems likely that this port will only ever be used for xDS, and all other things will continue to use the RPC port.

To make our config and documentation less confusing I've renamed this config setting to be `xds` instead of `grpc`. This change should be entirely backwards compatible. The old config setting and command line flag both continue to work, and the `consul connect envoy` command is still able to query for this port with the new name.

TODO:
* [x] the v2compat integration tests are failing because they use an older CLI with a newer agent. Need to restore the GRPC.Ports to the DebugConfig to keep that working.
* [x] changelog entry


command/agent/agent.go
```diff
@@ -201,7 +201,7 @@ func (c *cmd) run(args []string) int {
 	ui.Info(fmt.Sprintf("    Datacenter: '%s' (Segment: '%s')", config.Datacenter, segment))
 	ui.Info(fmt.Sprintf("        Server: %v (Bootstrap: %v)", config.ServerMode, config.Bootstrap))
 	ui.Info(fmt.Sprintf("   Client Addr: %v (HTTP: %d, HTTPS: %d, gRPC: %d, DNS: %d)", config.ClientAddrs,
```
rboyer, line 203:
```suggestion
	ui.Info(fmt.Sprintf("   Client Addr: %v (HTTP: %d, HTTPS: %d, xDS: %d, DNS: %d)", config.ClientAddrs,
```

agent/config/runtime.go
```diff
@@ -715,27 +715,27 @@ type RuntimeConfig struct {
 	// hcl: encrypt_verify_outgoing = (true|false)
 	EncryptVerifyOutgoing bool
 
-	// GRPCPort is the port the gRPC server listens on. Currently this only
+	// XDSPort is the port the xDS gRPC server listens on. This port only
 	// exposes the xDS and ext_authz APIs for Envoy and it is disabled by default.
 	//
 	// hcl: ports { grpc = int }
 	// flags: -grpc-port int
```
rboyer, line 722:
```suggestion
	// hcl: ports { xds = int }
	// flags: -xds-port int
```

agent/config/runtime.go
```diff
@@ -715,27 +715,27 @@ type RuntimeConfig struct {
...
 	// space separated list of ip addresses, UNIX socket paths and/or
 	// go-sockaddr templates. UNIX socket paths must be written as
-	// 'unix://<full path>', e.g. 'unix:///var/run/consul-grpc.sock'.
+	// 'unix://<full path>', e.g. 'unix:///var/run/consul-xds.sock'.
 	//
-	// If 'addresses.grpc' was not provided the 'client_addr' addresses are
+	// If 'addresses.xds' was not provided the 'client_addr' addresses are
 	// used.
 	//
 	// hcl: client_addr = string addresses { grpc = string } ports { grpc = int }
```
rboyer, line 737:
```suggestion
	// hcl: client_addr = string addresses { xds = string } ports { xds = int }
```

dnephin:
Test failures is a known flake and unrelated to this change.

website/content/docs/agent/options.mdx
```diff
@@ -242,8 +242,8 @@ The options below are all specified on the command-line.
   If it is provided after Consul has been initialized with an encryption key, then
   the provided key is ignored and a warning will be displayed.
 
-- `-grpc-port` ((#\_grpc_port)) - the gRPC API port to listen on. Default
```
rboyer, line 245: Since our docs aren't versioned yet, I think we need to keep the two old grpc lines in here, but mark them as deprecated and link them to the new names.

rboyer (approved):
LGTM pending the one minor suggestion about retaining the old config keys on the website.

website/content/docs/agent/options.mdx
dnephin: Unfortunate, but makes sense

rboyer: Usually when we deprecate stuff like this we also indicate which version of consul it was deprecated in so that folks using older versions can still figure out what's going on for their version.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/408416.

website/content/docs/agent/options.mdx
dnephin: Ah, sorry I missed this comment. I wasn't sure which version this would land in yet. I guess it will be 1.11, so I can add that.

dnephin: https://github.com/hashicorp/consul/pull/10618

PR #10589, opened 2021-07-09
Fix warning message typo

rileytg:
I still feel like it is not a very clear message but i dont really understand this well enough to articulate the logic...

related code: 
```golang
maxRemoval := (voters - 1) / 2
...
} else if maxRemoval < 1 {
...
```


dnephin:
Thank you for the PR! This warning message is in a library, so I think this change will need to be applied to https://github.com/hashicorp/raft-autopilot first. If you open a PR there, we can get it merged.

rileytg:
doh! `vendor/`... i didnt even check the path!

rileytg:
https://github.com/hashicorp/raft-autopilot/pull/11

PR #996, opened 2015-06-03
Bumps protocol version to 3 to get serf version 5.

slackpad:
This can't be merged until hashicorp/serf#293 has been merged. It ups the corresponding consul version which maps to the new serf version.

PR #10598, opened 2021-07-12
acls: Show `AuthMethodNamespace` when reading/listing ACL token meta

### LINKED ISSUE #10557: `ACLTokenListStub` is missing the `ACLAuthMethodEnterpriseMeta` from the `ACLToken`
#### Overview of the Issue

The [`ACLAuthMethodEnterpriseMeta` field](https://github.com/hashicorp/consul/blob/619eb5875d856c8f0f7adcc68e87086adcbe732d/agent/structs/acl.go#L281-L282) of the `ACLToken` structure is not being propagated to the `ACLTokenListStub` type along with the [`AuthMethod` field](https://github.com/hashicorp/consul/blob/619eb5875d856c8f0f7adcc68e87086adcbe732d/agent/structs/acl.go#L587).

This was an oversight on my part when adding the extra field and we should ensure our ACL token listings include this information.

#### Reproduction Steps

Requires Consul Enterprise

1. Create an auth method in the default namespace which generates tokens into a different namespace.
1. `consul login` using that auth method
1. `consul acl token list -ns <non default namespace>`

You should see the auth methods namespace (`default`) presented somehow but currently thats not possible due to the list stub not containing that information.

eculver:
Closes #10557.

All right, this is a fun one. The code is rather simple, but testing it was a challenge. The goal is to show auth method namespace in some capacity (see #10557). The updated CLI output looks like this:

```
$ consul acl token read -self -token-file=jwt-test.consul.token -meta
consul acl token read -self -token-file=jwt-test.consul.token -meta
AccessorID:       15a9579a-dfb9-1c69-260e-8e8ebe64e184
SecretID:         d324616b-18cf-0505-fac8-cd9f8cf48394
Namespace:        testing
Description:      token created via login
Local:            true
Auth Method:      jwt-test (Namespace: default)
Create Time:      2021-07-12 11:04:23.622906 -0700 PDT
Hash:             69b17b95cc30a91956202427641393816c5e8c5806fc2981260b387ea3e1ae18
Create Index:     21
Modify Index:     21
Service Identities:
   test-service (Datacenters: all)
```

The changes will only be visible to Enterprise. There is one small change to the Enterprise codebase that is also required ([consul-enterprise #996](https://github.com/hashicorp/consul-enterprise/pull/996)).

To test it, I wrote a little script that sets up a JWT auth method and a new `testing` namespace. The auth method is configured to mint tokens in the `testing` namespace. When configured, minting a new token yields the output above for both the `read` and `list` endpoints/commands.


freddygv (commented):
Nice work!

How come you chose to guard the new field behind the `meta` CLI flag? As opposed to always displaying `Auth Method Namespace:`

eculver:
> How come you chose to guard the new field behind the `meta` CLI flag? As opposed to always displaying `Auth Method Namespace:`

Since we are already showing the `AuthMethod` and this field is technically on that nested object, I thought it would be nice to keep existing output cleanly displayed by default. I could certainly be persuaded to show it by default though... we could even do something like:
```
AuthMethod: testing (Namespace: default)
```
So that the output is still in one line. Thoughts?

mkeeler:
In general the pretty output of our CLI is not considered an area for backwards compatibility. The JSON output is what any automation should use to ensure that fields can be pulled out consistently across versions.

Given that I think I would be in favor of outputting it by default.

freddygv (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/411039.

PR #10602, opened 2021-07-12
Set up Dependabot 

mmorel-35:


dnephin (commented):
Thank you for the PR! Could you please help us understand your motivation in making this change? 

I think we would rather manage dependency updates as part of the regular release cycle, instead of updating them immediate after they are released.

mmorel-35:
The intention os to help you keep dependencies and github actions up-to-date.
I'm not sure what you mean by 

> updating them immediate after they are released.

Dependabot will create a pr to update the dependency according to the conditions defined in the configuration file. 

So you can say that you want to receive the pr daily, weekly or even monthly. Then it's up to you to decide to merge them just before release, as you which. Anyway it's here so that you don't need to worry about checking for new versions of your dependencies manually.

dnephin:
Thank you for opening this PR and starting the discussion! We had someone internally add a similar config in #10727. Hopefully it will help us update our dependencies. We definitely have a lot that can be updated.

Sorry we did not get this PR merged. Since we have a config now, I am going to close this one.

PR #10608, opened 2021-07-13
ui: [BUGFIX] Ensure we use the ns query param name when requesting permissions

johncowen:
Previously when namespaces were enabled, we weren't requesting permission for the actively selected namespace, and instead always checking the permissions for the default namespace.

This PR ensures we request permissions for the actively selected namespace.

Notes:

All of the UI/API reading interaction uses a GET request with`?ns=` to specify the namespace, we have one API request which requires a POST with the request body containing the `Namespace`. This is so we can request various pieces of information required for the UI using a method we can customise for exactly what we need (there are more details in the comments in the changeset here). Unfortunately we were passing through a `nspace` parameter instead of a `ns` parameter which is the param name we are expecting deeper down, pretty much a typo.

An initial fix (https://github.com/hashicorp/consul/pull/10608/commits/6f00c81ca897febda1e528af166c770fc6edb249) meant  converting this from nspace to ns inside the repository itself. But then I realized that the entire 'repository' interface as of not too long ago now uses/expects `ns` not `nspace`, so in a follow up commit (https://github.com/hashicorp/consul/pull/10608/commits/5c868272faff4724c1fe7b6866341b02076c2cb7) I undid the first approach and made the fix consistent with the rest of the interface.

### DIFF
--- a/.changelog/10608.txt
+++ b/.changelog/10608.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: **(Enterprise only)** Ensure permissions are checked based on the actively selected namespace
+```
--- a/ui/packages/consul-ui/app/adapters/permission.js
+++ b/ui/packages/consul-ui/app/adapters/permission.js
@@ -7,7 +7,7 @@ export default class PermissionAdapter extends Adapter {
   requestForAuthorize(request, { dc, ns, resources = [], index }) {
     // the authorize endpoint is slightly different to all others in that it
     // ignores an ns parameter, but accepts a Namespace property on each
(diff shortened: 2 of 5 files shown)


kaxcode (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/410079.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 0762da3a629a6b50f887d5e6a3f440cdd99e5b52 onto `release/1.10.x` succeeded!

Issue #10545, opened 2021-07-01
armhv7: panic: unaligned 64-bit atomic operation


mrmstn:
When filing a bug, please include the following headings if possible. Any example text in this template can be deleted.

#### Overview of the Issue
Hi there

I'm using the release `consul_1.10.0_linux_armhfv6` on a raspery pi 4B 4G model. 

Sadly, the agent crashes always with the panic message: `unaligned 64-bit atomic operation`
```
2021-07-01T22:58:55.832+0200 [INFO]  agent.server: Handled event for server in area: event=member-join server=pi-rack1.dc1 area=wan
2021-07-01T22:58:55.834+0200 [INFO]  agent.server: Raft data found, disabling bootstrap mode
panic: unaligned 64-bit atomic operation

goroutine 70 [running]:
runtime/internal/atomic.panicUnaligned()

```

I made a downgrad to 1.9.7 and it worked out of the box

#### Reproduction Steps

Steps to reproduce this issue, eg:

1. Setup a raspberry PI 4B with the latest raspian OS
1. `wget https://releases.hashicorp.com/consul/1.10.0/consul_1.10.0_linux_armhfv6.zip`
1. extract and start `consul agent -dev`
1. View error

### Consul info for both Client and Server

<details>
  <summary>Client info</summary>

```
Error querying agent: Get "http://127.0.0.1:8500/v1/agent/self": dial tcp 127.0.0.1:8500: connect: connection refused
```

</details>

<details>
  <summary>Server info</summary>

```
Error querying agent: Get "http://127.0.0.1:8500/v1/agent/self": dial tcp 127.0.0.1:8500: connect: connection refused
```

</details>

### Operating system and Environment details

OS, Architecture, and any other information you can provide about the environment.

### Log Fragments
```
2021-07-01T22:58:55.831+0200 [INFO]  agent.server: Adding LAN server: server="pi-rack1 (Addr: tcp/192.168.5.63:8300) (DC: dc1)"
2021-07-01T22:58:55.832+0200 [WARN]  agent: grpc: addrConn.createTransport failed to connect to {192.168.5.63:8300 0 pi-rack1 <nil>}. Err :connection error: desc = "transport: Error while dialing dial tcp 192.168.5.63:8300: operation was canceled". Reconnecting...
2021-07-01T22:58:55.832+0200 [INFO]  agent.server: Handled event for server in area: event=member-join server=pi-rack1.dc1 area=wan
[...]

Issue #6519, opened 2019-09-20
ARMv7 Support


abrucker235:
Please search the existing issues for relevant feature requests, and use the reaction feature (https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to add upvotes to pre-existing requests.

#### Feature Description
Downloads supporting latest ARM versions.

#### Use Case(s) 
Homelab Raspberry PI4 Clusters or A Series machines on Amazon.
 
I got a simple working version compiling myself: GOOS=linux GOARM=7 GOARCH=arm go build

jwalzer:
Is this still unsolved?

https://github.com/hashicorp/docker-consul currently breaks, when building the container on a RasperryPi because no build for armv7 is available.

mikemorris:
This should be resolved in the upcoming 1.11.0 release with the merge of https://github.com/hashicorp/consul/pull/11417, as well as upcoming patch releases for [1.10.x](https://github.com/hashicorp/consul/pull/11454), [1.9.x](https://github.com/hashicorp/consul/pull/11452), [1.8.x](https://github.com/hashicorp/consul/pull/11455) once the corresponding PRs have been merged and we switch over to the new release process.

This new process will build `GOARCH=arm GOARM=5` 32-bit binaries (intended for broad compatibility) and `GOARCH=arm64` (ARMv8) 64-bit binaries, per the matrix at https://github.com/golang/go/wiki/GoArm

proffalken:
@mikemorris is there any more news on this?

Arch and Ubuntu are showing 1.11.1, whereas the Arm repos for rasbian etc. are only showing 1.10.x

Just wanted to make sure I'd not missed an updated release URI or similar?

blake:
Hi @proffalken,

> Arch and Ubuntu are showing 1.11.1, whereas the Arm repos for rasbian etc. are only showing 1.10.x

These repositories are not managed by HashiCorp. [HashiCorp's Linux repository](https://www.hashicorp.com/blog/announcing-the-hashicorp-linux-repository) currently only provides packages for x86_64/amd64 systems.

Binaries for 32-bit ARM can be found at <https://releases.hashicorp.com/consul/1.11.1/>.

proffalken:
Ok thanks @blake

blake:
Hi @proffalken,

> HashiCorp's Linux repository currently only provides packages for x86_64/amd64 systems.

[...]

PR #10610, opened 2021-07-13
ci: set CGO_ENABLED: 0 for arm/arm64 builds

mikemorris:
Refs https://github.com/hashicorp/consul/pull/7970 for context on the ARM crash in versions of Go prior to 1.14 that required enabling CGO.

As an open question, is the suggested approach in https://github.com/hashicorp/consul/issues/6519#issuecomment-547951176 sufficient for running on ARMv7 hardware @jwalzer @mrmstn? It seems that approach (aside from the reported bug) has been working for some users according to https://github.com/hashicorp/consul/issues/10545.

**TODO:** This change only affects our test suite, and will require a followup to our build pipelines if it looks good to merge.a

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -392,7 +392,7 @@ jobs:
       - image: *GOLANG_IMAGE
     environment:
       <<: *ENVIRONMENT
-      CGO_ENABLED: 1
+      CGO_ENABLED: 0
       GOOS: linux
     steps:
       - checkout


alvin-huang:
With https://github.com/golang/go/pull/34030 in Go 1.14, we should be able to compile arm/arm64 normally without having to download the gcc libraries IIRC. We just never switched it. 

I don't know if we just need a `linux_arm` (default ARMv6) and `linux_arm64` (aarch64/ARMv8) or if we need to support `GOARM=5|6|7`  as well.

mikemorris:
This is obsolete as of https://github.com/hashicorp/consul/pull/11417 merging, although we should probably remove the entire [`build-arm`](https://github.com/hashicorp/consul/blob/c61208146d40ea6fb7801071399b555bced2ddf7/.circleci/config.yml#L399-L429) and other binary build sections from CircleCI config as a followup, preferring to instead validate binary builds using the new GitHub actions workflows.

hc-github-team-consul-core:
Backport failed @zalimeni. Run: [this is a test to fix broken commenting, no link included]

PR #8882, opened 2020-10-08
Add rotate option for recursors

munali:
Closes #8807

PR #10611, opened 2021-07-14
Add DNS recursor strategy option

### LINKED ISSUE #8807: Add option to round robin the queries to upstream DNS servers
#### Feature Description

Add a configuration parameter (e.g. recursor_rotate) to have round-robin selection of name servers from among those listed as recursors. Similar to "option:rotate" in resolv.conf.

Proposed change is to agent/dns.go, handleRecurse. 
for _, recursor := range cfg.Recursors {

Instead keeping track of a counter per call to handleRecurse that is used to iterate over recursors. Not sure yet of the implementation...new to go.

#### Use Case(s)

In an environment where nodes will use Consul directly as the DNS server (aka no dnsmasq), queries for records outside of the "consul." domain will be resolved upstream. When migrating to using Consul for service registry many records are resolved upstream. Having an ability to spread the load across upstream servers is necessary to continue to have quick response times.

blake:
This change adds a new `dns_config.recursor_strategy` option which controls how Consul queries DNS resolvers listed in the `recursors` config option. The supported options are `sequential` (default), and `random`.
    
This is a continuation of hashicorp/consul#8882 with additional changes to address review feedback from that PR.
    
Closes #8807


dnephin (commented):
Thank you for working on getting this ready to merge! I think it's looking good, just a couple suggestions

dnephin (commented):
Looks good! I think the test failures should be easy to fix, one is failing because the  types are different (string vs RecusorStrategy), and the other can be fixed by running `go test ./agent/config -update` to update the golden file.

I guess we probably also want some test coverage of the new behaviour. I'm not sure what the best way is to handle that yet, but I'd be happy to work with you to figure that one, whenever you have a chance.

agent/dns.go
```diff
@@ -1844,7 +1847,11 @@ func (d *DNSServer) handleRecurse(resp dns.ResponseWriter, req *dns.Msg) {
 	var r *dns.Msg
 	var rtt time.Duration
 	var err error
-	for _, recursor := range cfg.Recursors {
+	for idx, random_idx := range rand.Perm(len(cfg.Recursors)) {
+		recursor := cfg.Recursors[idx]
+		if cfg.RecursorStrategy == "random" {
+			recursor = cfg.Recursors[random_idx]
+		}
```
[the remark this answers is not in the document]
blake: Thanks for the feedback. I've incorporated this into the latest version.

agent/config/runtime.go
```diff
@@ -270,6 +270,15 @@ type RuntimeConfig struct {
...
 	DNSOnlyPassing bool
 
+	// DNSRecursorStrategy controls the order in which DNS recursors are queried.
+	// 'sequential' queries recursors in the order they are listed under `recursors`.
+	// 'random' causes random selection of recursors which has the effect of
+	// spreading the query load among all listed servers, rather than having
+	// client agents try the first server in the list every time.
+	//
+	// hcl: dns_config { recursor_strategy = "(random|sequential)" }
+	DNSRecursorStrategy string
```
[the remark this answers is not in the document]
blake: Done. :-)

.changelog/10611.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
```
dnephin, line 1: Not sure if we would generally use `feature` or `improvement` for this kind of thing. I don't have a strange preference either way.

agent/dns_test.go
```diff
@@ -3832,6 +3832,45 @@ func TestDNS_Recurse_Truncation(t *testing.T) {
 	}
 }
 
+func TestDNS_Recursor_StrategyRandom(t *testing.T) {
```
dnephin, line 3835: Since the code this is testing is in `agent/dns/dns.go` , I guess these two tests should move to `agent/dns/dns_test.go`

agent/dns_test.go
```diff
@@ -3832,6 +3832,45 @@ func TestDNS_Recurse_Truncation(t *testing.T) {
...
+	configuredRecursors := []string{"1.1.1.1", "8.8.4.4", "8.8.8.8"}
+	recursorStrategy := agentdns.RecursorStrategy("random")
+
+	retries := func() *retry.Counter {
+		return &retry.Counter{Count: 5}
+	}
+
+	retry.RunWith(retries(), t, func(r *retry.R) {
```
dnephin, line 3843: minor: no need for this function I think, we can specify the value directly

```suggestion
	retry.RunWith(&retry.Counter{Count: 5}, t, func(r *retry.R) {
```

agent/dns_test.go
```diff
@@ -3832,6 +3832,45 @@ func TestDNS_Recurse_Truncation(t *testing.T) {
...
+		}
+
+		// Ensure the slices contain the same elements
+		require.ElementsMatch(t, configuredRecursors, recursorsToQuery)
+
+		if reflect.DeepEqual(configuredRecursors, recursorsToQuery) {
+			// Error if the elements are in the same order, and retry generating
+			// random recursor list
+			r.Fatal("dns recursor order is not randomized.")
+		}
```
dnephin, line 3856: I think this also works:

```go
require.Equal(r, configuredRecursors, recursorsToQuery)
```

dnephin (approved):
LGTM, thanks for adding these tests! Just a few minor suggestions on the test, otherwise I think this is ready to merge.

agent/dns_test.go
blake: I initially tried this approach, but when running the test under the debugger I did not see the retry loop being executed if the recursor order was equal. Using `require.Equal(r, configuredRecursors, recursorToQuery)` seems to cause the test to fail silently, and appear to `go test` as if the test passed, whereas `r.Fatal()` correctly triggers the retry loop.

dnephin: Oh, that's concerning. I made some changes to retry recently. I'll double check how it is working.

dnephin: Oh, right of course `require.Equal` is not at all what we want! It should be `require.NotEqual` to match this behaviour.

But I think what you have here is good too

dnephin (approved):

agent/dns_test.go
blake: You're right, I should've been using `require.NotEqual()`. 🤦‍♂️ I just updated the code to use this function and the retry is working as expected.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/412643.

PR #10612, opened 2021-07-14
acl: acl replication routine to report the last error message

### LINKED ISSUE #10521: ensure replication routines can report what the last error message was
Namely when the acl replication routines fail, they set `LastError` on the [acl replication status endpoint](https://www.consul.io/api-docs/acl#check-acl-replication). It would be nice to include something like `LastErrorMessage` captured the content of the `error` that caused the `LastError` field to be updated.

Possibly worth also including giving any replication routines their own replication status endpoints if they don't currently have one (like federation states, config entries, etc).

bigmikes:
This PR adds the new field `LastErrorMessage` to the `ACLReplicationStatus` endpoint. Whenever the Goroutine executing the `runACLReplicator` encounters an error, both `LastError` and `LastErrorMessage` field in the endpoint are updated. 

Test `TestACLReplication_TokensRedacted` is updated to check also the `LastErrorMessage` field.

Fixes #10521.

### DIFF
--- a/agent/consul/acl_replication.go
+++ b/agent/consul/acl_replication.go
@@ -484,11 +484,12 @@ func (s *Server) IsACLReplicationEnabled() bool {
 		s.config.ACLTokenReplication
 }
 
-func (s *Server) updateACLReplicationStatusError() {
+func (s *Server) updateACLReplicationStatusError(errorMsg error) {
 	s.aclReplicationStatusLock.Lock()
 	defer s.aclReplicationStatusLock.Unlock()
 
 	s.aclReplicationStatus.LastError = time.Now().Round(time.Second).UTC()
+	s.aclReplicationStatus.LastErrorMessage = errorMsg
 }
 
 func (s *Server) updateACLReplicationStatusIndex(replicationType structs.ACLReplicationType, index uint64) {
--- a/agent/consul/acl_replication.go
+++ b/agent/consul/acl_replication.go
@@ -484,12 +484,12 @@ func (s *Server) IsACLReplicationEnabled() bool {
 		s.config.ACLTokenReplication
 }
 
-func (s *Server) updateACLReplicationStatusError(errorMsg error) {
+func (s *Server) updateACLReplicationStatusError(errorMsg string) {
 	s.aclReplicationStatusLock.Lock()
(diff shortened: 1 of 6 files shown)


agent/consul/acl_replication.go
```diff
@@ -484,11 +484,12 @@ func (s *Server) IsACLReplicationEnabled() bool {
 		s.config.ACLTokenReplication
 }
 
-func (s *Server) updateACLReplicationStatusError() {
+func (s *Server) updateACLReplicationStatusError(errorMsg error) {
```
dnephin, line 487: Since we are already doing a `err != nil` check in the caller, we were thinking it would be safer to accept a string here instead of an error. That way no new callers can accidentally pass a `nil` error.

```suggestion
func (s *Server) updateACLReplicationStatusError(errorMsg string) {
```

which I guess will requires a change to the line below that uses `errorMsg`.

agent/structs/acl.go
```diff
@@ -1269,6 +1269,7 @@ type ACLReplicationStatus struct {
 	ReplicatedTokenIndex uint64
 	LastSuccess          time.Time
 	LastError            time.Time
+	LastErrorMessage     string
```
dnephin, line 1272: There is unfortunately another copy of this struct in `api/acl.go` which will need this field added to it as well. That copy of `ACLReplicationStatus` is used by the `api` client library package to deserialize the API response.

dnephin (commented):
Thank you for the PR! This looks really good. I left a couple comments below for a couple small improvements.

It would also be great to include a changelog entry. A changelog entry can be added in a file named `.changelog/10612.txt` (named using the PR number). You can see some other examples in that directory for the exact syntax to use.

agent/structs/acl.go
bigmikes: Done.

agent/consul/acl_replication.go
bigmikes: Done.

dnephin (approved):
LGTM, thank you!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/423688.

PR #10613, opened 2021-07-14
Feature: allow manipulation of HTTP headers in ingress and mesh routing

banks:
This PR adds HTTP header manipulation options to three new places:

 - `ingress-gateway` services
 - `service-router` route destinations
 - `service-splitter` split legs

The syntax in all three places is the same:

```hcl
request_headers {
	add {
		foo = "bar"
	}
	set {
		bar = "baz"
	}
	remove = ["qux"]
}
response_headers {
	add {
		foo = "bar"
	}
	set {
		bar = "baz"
	}
	remove = ["qux"]
}
```

Unit tests are updated to make this work and new integration tests cover all all of these new places.

The PR is largish mostly due to the golden unit test files but the Commits build the features out incrementally if that helps reviewers.

## TODO In this PR

 - [x] Verify clean merge with Enterprise
 - [x] Add `api` package support
 - [x] Add Changelog entry
 - [x] Add integration tests that exercise replacements using [envoy special variables](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#custom-request-response-headers) to be sure those work as expected.

## TODO in later PR

 - [ ] Document the new config-entry options
 - [ ] Document the slight changes to the discovery chain API response (new fields on splitter

### DIFF
--- a/agent/structs/config_entry_discoverychain.go
+++ b/agent/structs/config_entry_discoverychain.go
@@ -677,8 +677,8 @@ type ServiceSplit struct {
 // MergeParent is called by the discovery chain compiler when a split directs to
 // another splitter. We refer to the first ServiceSplit as the parent and the
 // ServiceSplits of the second splitter as its children. The parent ends up
-// "flattened" by the compiler, i.e. replaced with it's children recursively
-// with the weights modified as necessary.
+// "flattened" by the compiler, i.e. replaced with its children recursively with
+// the weights modified as necessary.
 //
 // Since the parent is never included in the output, any request processing
 // config attached to it (e.g. header manipulation) would be lost and not take
@@ -721,7 +721,7 @@ func (s *ServiceSplit) MergeParent(parent *ServiceSplit) (*ServiceSplit, error) 
 	}
 
 	// Merge any request handling from parent _unless_ it's overridden by us.
-	copy.RequestHeaders, err = s.RequestHeaders.MergeDefaults(parentReq)
+	copy.RequestHeaders, err = MergeHTTPHeaderModifiers(parentReq, s.RequestHeaders)
 	if err != nil {
 		return nil, err
 	}
@@ -735,7 +735,7 @@ func (s *ServiceSplit) MergeParent(parent *ServiceSplit) (*ServiceSplit, error) 
 	// time since responses flow the other way so the unflattened behavior would
 	// be that the parent processing happens _after_ ours potentially overriding
 	// it.
-	copy.ResponseHeaders, err = parentResp.MergeDefaults(s.ResponseHeaders)
+	copy.ResponseHeaders, err = MergeHTTPHeaderModifiers(s.ResponseHeaders, parentResp)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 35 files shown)


agent/structs/discovery_chain.go
```diff
@@ -191,6 +191,13 @@ type DiscoveryRoute struct {
 
 // compiled form of ServiceSplit
 type DiscoverySplit struct {
+	Definition *ServiceSplit `json:",omitempty"`
+	// Weight is not necessarily a duplicate of Definition.Weight since when
+	// multiple splits are compiled down to a single set of splits the effective
+	// weight of a split leg might not be the same as in the original definition.
+	// Proxies should use this compiled weight. The Definition is provided above
+	// for any other significant configuration that the proxy might need to apply
+	// to that leg of the split.
```
banks, line 200: Adding the definition is needed for the new config attached to split legs to propagate to xDS. It will also mean that the `/v1/discovery-chain` API will inherit this new data (which is good if anyone actually uses it). I've put a TODO to update this in API docs.

agent/structs/config_entry_gateways.go
```diff
@@ -186,6 +198,11 @@ func (e *IngressGatewayConfigEntry) Validate() error {
 			if s.NamespaceOrDefault() == WildcardSpecifier {
 				return fmt.Errorf("Wildcard namespace is not supported for ingress services (listener on port %d)", listener.Port)
 			}
+			sid := NewServiceID(s.Name, &s.EnterpriseMeta)
+			if _, ok := serviceNames[sid]; ok {
+				return fmt.Errorf("Service %s cannot be added multiple times (listener on port %d)", sid, listener.Port)
+			}
+			serviceNames[sid] = struct{}{}
```
banks, line 205: This is _technically_ a breaking change in the sense that we didn't error before if you had duplicate services in an ingress gateway config so it "worked" although it's not clear what it actually means.

I _think_ I'm OK with calling it a bug though rather than a feature - it seems like impact would be extremely low and very easy to mitigate, nothing would break initially, just config entries that validated before could possibly fail after this but can be fixes without any behaviour change.

Does that seem reasonable for a point release? I think it's OK.

The reason it now actually matters is that you could configure different header manip on two duplicate services with the same service name and it becomes ambiguous which rules should apply to traffic for that service.

We _could_ instead not make this an error but just document that if there duplicates we'll always ignore them after the first one or something?

banks:
There are a few legit test failures I'll go fix...

agent/structs/config_entry_discoverychain.go
```diff
@@ -1460,3 +1468,43 @@ func IsProtocolHTTPLike(protocol string) bool {
...
+		return true
+	}
+	return len(m.Add) == 0 && len(m.Set) == 0 && len(m.Remove) == 0
+}
+
+func (m *HTTPHeaderModifiers) Validate(protocol string) error {
+	if m == nil {
+		// Empty is always valid
+		return nil
+	}
```
rboyer, line 1501: This should be unnecessary since the IsZero method already returns true with a nil receiver.

rboyer:
There's a stray empty file committed named `ingress-with-chain-and-router-header-manip`.

banks:
Dealt with the stray file and redundant nil check @rboyer thanks! Are you happy to approve the rest of the changes or did you just have a quick glance while it was WIP?

agent/xds/routes.go
```diff
@@ -199,6 +241,36 @@ func (s *ResourceGenerator) routesForIngressGateway(
 	return result, nil
 }
 
+func makeHeadersValueOptions(vals map[string]string, add bool) []*envoy_core_v3.HeaderValueOption {
```
kisunji, line 244: Minor, but I think this is clearer
```suggestion
func makeHeaderValueOptions(vals map[string]string, append bool) []*envoy_core_v3.HeaderValueOption {
```

agent/structs/config_entry_discoverychain.go
```diff
@@ -1463,3 +1471,39 @@ func IsProtocolHTTPLike(protocol string) bool {
...
 }
+
+// HTTPHeaderModifiers is a set of rules for HTTP header modification that
+// should be performed by proxies as the request passes through them. It can
+// operate on either request or response headers depending on the context in
+// which it is used.
+type HTTPHeaderModifiers struct {
+	// Add is a set of name -> value pairs that should be appended to the request
+	// or response (i.e. allowing duplicates if the same header already exists).
+	Add map[string]string `json:",omitempty"`
```
kisunji, line 1482: Is there a use-case when we want to append multiple values to a single key:
```
Foo: bar, baz
```
It looks like a map wouldn't work if trying to register multiple appends to a single key?

banks:
Rebased based on the Enterprise changes needed.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/442994.

PR #10619, opened 2021-07-15
xds: ensure single L7 deny intention with default deny policy does not result in allow action (CVE-2021-36213)

rboyer:
Note to self:

- 1.10.x will cleanly backport
- 1.9.x requires a manual backport (which is not bad) : https://github.com/hashicorp/consul/pull/10620

PR #10620, opened 2021-07-15
(1.9.x) xds: ensure single L7 deny intention with default deny policy does not result in allow action (CVE-2021-36213)

rboyer:
Backport of #10619 to 1.9.x


dhiaayachi (approved):
same as #10619

PR #10623, opened 2021-07-15
Validate Subject Alternative Name for upstreams (1.9.x)

freddygv:
These changes ensure that the identity of services dialed is
cryptographically verified.

For all upstreams we validate against SPIFFE IDs in the format used by
Consul's service mesh:

spiffe://<trust-domain>/ns/<namespace>/dc/<datacenter>/svc/<service>


rboyer (approved):
LGTM pending changelog

PR #10630, opened 2021-07-16
defer setting the state before returning to avoid  stuck in `INITIALIZING` state

### LINKED ISSUE #10625: CA provider state stuck in `INITIALIZING` after provider initialization fail
#### Overview of the Issue
This happened in case when VAULT key expired and consul was restarted before reconfiguring CA

#### Reproduction Steps

- `InitializeCA` on startup fails because of Vault token issue
- background loop retries it and it keeps failing. Each time it leaves `CAManager.state = UNINITIALIZED`
- User reconfigures CA to fix the token which does it’s thing and eventually sets `CAManager.state = INITIALIZED`
- background loop retry’s one more time, exits early with no error because of this early return on state `INITIALIZED` 
- sets state back to `INITIALIZING` in the process
- Nothing else can perform any further `CAmanager` updates because of the state

The reason this happen is the following bug:
https://github.com/hashicorp/consul/blob/54813b2a085f45817a65ace6655966335d386323/agent/consul/leader_connect_ca.go#L361-L363

notice that there is a possibility to set the state to `INITIALIZING` and return before getting to the `defer` that put it back to either `INITIALIZED` or `UNINITIALIZED`

dhiaayachi:
fix #10625

### DIFF
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -154,14 +154,22 @@ func (c *CAManager) setState(newState caState, validateState bool) (caState, err
 	state := c.state
 
 	if !validateState ||
-		state == caStateInitialized ||
+		(state == caStateInitialized && newState != caStateInitializing) ||
 		(state == caStateUninitialized && newState == caStateInitializing) ||
 		(state == caStateUninitialized && newState == caStateReconfig) {
 		c.state = newState
 	} else {
-		return state, fmt.Errorf("CA is already in state %q", state)
+		return state, &caStateError{Current: state}
 	}
 	return state, nil
+}
+
+type caStateError struct {
+	Current caState
+}
+
+func (e *caStateError) Error() string {
+	return fmt.Sprintf("CA is already in state %q", e.Current)
 }
 
 // setPrimaryRoots updates the most recently seen roots from the primary.
@@ -360,8 +368,12 @@ func (c *CAManager) InitializeCA() (reterr error) {
 	}
 
 	// Update the state before doing anything else.
-	oldState, err := c.setState(caStateInitializing, true)
-	if err != nil {
+	_, err := c.setState(caStateInitializing, true)
+	var errCaState *caStateError
+	switch {
+	case errors.As(err, &errCaState) && errCaState.Current == caStateInitialized:
+		return nil
+	case err != nil:
 		return err
 	}
 
@@ -376,11 +388,6 @@ func (c *CAManager) InitializeCA() (reterr error) {
 			c.setState(caStateUninitialized, false)
 		}
 	}()
-
-	// if we were already in the initialized state then there is nothing to be done.
-	if oldState == caStateInitialized {
-		return nil
-	}
 
 	// Initialize the provider based on the current config.
 	conf, err := c.initializeCAConfig()
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -361,7 +361,6 @@ func (c *CAManager) InitializeCA() (reterr error) {
 
 	// Update the state before doing anything else.
 	oldState, err := c.setState(caStateInitializing, true)
-	// if we were already in the initialized state then there is nothing to be done.
 	if err != nil {
 		return err
 	}
@@ -378,6 +377,7 @@ func (c *CAManager) InitializeCA() (reterr error) {
 		}
 	}()
 
+	// if we were already in the initialized state then there is nothing to be done.
 	if oldState == caStateInitialized {
 		return nil
 	}
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -362,9 +362,6 @@ func (c *CAManager) InitializeCA() (reterr error) {
 	// Update the state before doing anything else.
 	oldState, err := c.setState(caStateInitializing, true)
 	// if we were already in the initialized state then there is nothing to be done.
-	if oldState == caStateInitialized {
-		return nil
-	}
 	if err != nil {
 		return err
 	}
@@ -380,6 +377,10 @@ func (c *CAManager) InitializeCA() (reterr error) {
 			c.setState(caStateUninitialized, false)
 		}
 	}()
+
+	if oldState == caStateInitialized {
+		return nil
+	}
 
 	// Initialize the provider based on the current config.
 	conf, err := c.initializeCAConfig()
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -141,7 +141,7 @@ func NewCAManager(delegate caServerDelegate, leaderRoutineManager *routine.Manag
 // setState attempts to update the CA state to the given state.
 // Valid state transitions are:
 //
-// caStateInitialized -> <any state>
+// caStateInitialized -> <any state except caStateInitializing>
 // caStateUninitialized -> caStateInitializing
 // caStateUninitialized -> caStateReconfig
 //
--- a/.changelog/10630.txt
+++ b/.changelog/10630.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ca: fixed a bug when ca provider fail and provider state is stuck in `INITIALIZING` state.
+```


agent/consul/leader_connect_ca.go
```diff
@@ -362,9 +362,6 @@ func (c *CAManager) InitializeCA() (reterr error) {
 	// Update the state before doing anything else.
 	oldState, err := c.setState(caStateInitializing, true)
 	// if we were already in the initialized state then there is nothing to be done.
```
dnephin, line 364: I guess this comment should be moved as well.

dnephin (commented):
The code being changed here was introduced in d9d4c492abb1f5ebf9675efe5e006e8bf5a811d2, and looking at the changelog entries for that commit it seems to suggest it was attempting to fix this very bug.

Maybe the user experiencing this problem was not running v1.10 yet (where this fix was first introduced) ?

I think this PR may fix the issue, but I'm not confident I understand all the repercussions of this change. It's possible we are reverting a fix for some other bug.

I think it might be valuable to create a state diagram. I suspect `setState` could be changed to better model valid states which would help make the correct behaviour easier to reason about.

dhiaayachi:
> The code being changed here was introduced in [d9d4c49](https://github.com/hashicorp/consul/commit/d9d4c492abb1f5ebf9675efe5e006e8bf5a811d2), and looking at the changelog entries for that commit it seems to suggest it was attempting to fix this very bug.
> 
> Maybe the user experiencing this problem was not running v1.10 yet (where this fix was first introduced) ?
> 
> I think this PR may fix the issue, but I'm not confident I understand all the repercussions of this change. It's possible we are reverting a fix for some other bug.
> 
> I think it might be valuable to create a state diagram. I suspect `setState` could be changed to better model valid states which would help make the correct behaviour easier to reason about.

Thank you for looking at this @dnephin 

My understanding is we should never leave `InitializeCA` method with the provider state other then `INITIALIZED` or `UNINITIALIZED` which make sense as InitializeCA is supposed to do all the ca initializing and set the state to `INITIALIZED` or fail and set the state to `UNINITIALIZED`. 
The current implementation have a possibility to leave the `InitializeCA` method with the state `INITIALIZING` which will break the state machine and keep it stuck in `INITIALIZING`, this happen in the following conditions:

- we start `InitializeCA` with state `INITIALIZED`
- set the state to `INITIALIZING` which return `oldState` = `INITIALIZED`
- leave `InitializeCA` in the if statement

banks:
@dnephin yeah I think it was the fixes for those other issues that introduced this.

The user was running 1.9.7 which contains the same lines https://github.com/hashicorp/consul/blob/01555fb292cef4e127537bdbd0ab0c9904d4327b/agent/consul/leader_connect_ca.go#L310-L318

I think this specific issue was not the same as any of the ones Matt found and was trying to fix there although it's a similar class of problem. In general, I think his fixes made this code much more robust and worked well, I think this one issue of returning before the defer has introduced this new broken state since we are breaking the invariant that we always exit `InitializeCA` in a valid state for further changes (i.e. either it worked and we should now be `Initialized` or it failed and we should be back to `uninitialized`.

dnephin:
I added some thoughts and a diagram in #10634.

I forgot that the commit would not show if the cherry-pick was needed for a backport, good call looking at the actual release branch.

I think on the surface this change does fix the problem, but is it the right fix? Looking at the model I put into #10634, it seems like this is one of the invalid states. Maybe `setState` should actually reject this state transition, in which case the code in `InitializeCA` would already be correct.

dhiaayachi:
> I added some thoughts and a diagram in #10634.
> 
> I forgot that the commit would not show if the cherry-pick was needed for a backport, good call looking at the actual release branch.
> 
> I think on the surface this change does fix the problem, but is it the right fix? Looking at the model I put into #10634, it seems like this is one of the "INVALID" state transitions. Maybe `setState` should actually reject this state transition, in which case the code in `InitializeCA` would already be correct.

I agree, we need to understand more what got us to call `InitializeCA` while we are at `INITIALIZED` state first and then we can decide if it's better to forbid that transition in `setState` and remove the if statement (because it should never happen) or just move the if statement

banks:
> I think on the surface this change does fix the problem, but is it the right fix? Looking at the model I put into #10634, it seems like this is one of the "INVALID" state transitions. Maybe setState should actually reject this state transition, in which case the code in InitializeCA would already be correct.

Hmm yeah, I think cleaning up the state machine is a great idea. One thing I'd say is that _even if_ `setState` disallowed this transition, I'd still prefer to move the return as in this PR - having it be "correct" to exit early before the defer only because of the implementation of another method somewhere else feels unnecessarily fragile when there are no downsides to moving it afterwards.

> we need to understand more what got us to call InitializeCA while we are at INITIALIZED state

The thing that causes it is the background retrying of Initialize due to a misconfiguration which will only be resolved by the operator reconfiguring the CA. If we fix this by changing the state machine rules to be more restrictive then we will need to be very careful that the background Initializer is modified too. Right now if you only change `setState` then I think the background init routine will get into an infinite retry backoff and key trying to initialize forever but fail because it's not allowed to initialize when the state is already "initialized". Provided that case is handled specially then it seems fine to me.

I'd still also be fine with merging the PR as it is and continuing state machine cleanup as a separate task though as I think the defer is meant to make it easier to reason about the exit state regardless of code path and this ordering violates that unnecessarily but see what you think @dhiaayachi.

dhiaayachi:
@banks @dnephin 
I went through the code again today taking into account the fact that the issue happen when the retry loop is running and we perform a reconfigure at the same time and I have 2 observations:
- I think the root issue is that we let a retry loop run while we do reconfigure concurrently, this can lead to multiple race conditions other then the one we caught (Could explain the fact that the key was not updated in raft in our case). In my opinion reconfigure should stop the retry loop before starting and restart it after if the reconfigure fail.
- Fixing the state machine is a bit risky for this PR specially with the fact that we have `state == caStateInitialized ||` at the validation check and removing it impact other use cases. also most of the calls use `validateState = false` which bypass the state machine check and remove any guaranty on the transitions

I propose to:
-  keep this PR as is  to fix that specific case while limiting the risk
- investigate the possible issues around having a retry loop run while we do reconfigure
- start a series of PRs to guarantee the consistency of the state machine as proposed by @dnephin in #10634 and remove completely the option to disable the validation check.

banks:
That plan sounds good to me Dhia.

I think to be clear the disable validation flag is intended to be used in
places like Defer callbacks where we can't actually do anything useful with
the error anyway and we'd rather be sure that the state ends up back at
something reasonable even if something strange has happened. It would be
great if there is a cleaner pattern for that though.

dhiaayachi:
My thinking is that we can have a state machine with a clear transition path combined with an init that return the state machine to an initial state if something bad happen and we want to get back to a known state.

agent/consul/leader_connect_ca.go
```diff
@@ -154,16 +154,24 @@ func (c *CAManager) setState(newState caState, validateState bool) (caState, err
...
 
 	if !validateState ||
-		state == caStateInitialized ||
+		(state == caStateInitialized && newState != caStateInitializing) ||
 		(state == caStateUninitialized && newState == caStateInitializing) ||
 		(state == caStateUninitialized && newState == caStateReconfig) {
 		c.state = newState
 	} else {
-		return state, fmt.Errorf("CA is already in state %q", state)
+		return state, &caStateError{Current: state}
```
dhiaayachi, line 162: returning the state 2 times here is confusing. The only reason we return the state as part of `SetState` is to check if it changed, if we decide that we will return an error if the state is not changing or if the state transition is illegal and that error have the state embedded, we should not return the state anymore, only an error struct. If the error is nil that mean that the `SetState` succeeded and the new state is the one we passed as parameter.

dhiaayachi (commented):
Thanks @dnephin  for changing this. I think this way of fixing the bug bring us closer to fixing the state machine and is less error prone.
I have a comment inline about the signature of `SetState` that became a bit weird.

agent/consul/leader_connect_ca.go
```diff
@@ -154,16 +154,24 @@ func (c *CAManager) setState(newState caState, validateState bool) (caState, err
 	state := c.state
 
 	if !validateState ||
-		state == caStateInitialized ||
+		(state == caStateInitialized && newState != caStateInitializing) ||
```
banks, line 157: nit: we should probably update the comment above that lists the allowed state transitions

banks (commented):
Thanks Daniel, that does look cleaner!

One minor nit inline and I think @dhiaayachi has a good point provided no other callers use that previous state argument that it's now unnecessary. I didn't check other callers though so there may be a reason to leave it?

dnephin:
> provided no other callers use that previous state argument that it's now unnecessary. I didn't check other callers though so there may be a reason to leave it?

There is another caller that uses it to restore the previous state on error, so I think we still need it in both places until we make a larger change.

> we should probably update the comment above that lists the allowed state transitions

That comment is pretty misleading because it doesn't account for all the calls with validateState=false, as we can see from the diagram in #10634, the actual state transitions are more involved. Maybe it would be better to delete the comment instead of having it be misleading.

dhiaayachi:
> There is another caller that uses it to restore the previous state on error, so I think we still need it in both places until we make a larger change.

What about doing something like this?

```go
func (c *CAManager) setState(newState caState, validateState bool) error {
	c.stateLock.Lock()
	defer c.stateLock.Unlock()
	state := c.state

	if !validateState ||
		(state == caStateInitialized && newState != caStateInitializing) ||
		(state == caStateUninitialized && newState == caStateInitializing) ||
		(state == caStateUninitialized && newState == caStateReconfig) {
		c.previousState = c.state
		c.state = newState
	} else {
		return &caStateError{Current: state}
	}
	return nil
}
func (c *CAManager) revertState(state caState) error {
	c.stateLock.Lock()
	defer c.stateLock.Unlock()
	if c.state != state {
		return &caStateError{Current: c.state}
	}
	previousState := c.previousState
	c.previousState = c.state
	c.state = previousState
	return nil
}
```
I think we should have the state managed internally through clear interfaces (set,reset,revert) and not let the state user meddle with it, and this would be a step toward it. 
That said, I agree to keep this as a future step in enhancing this code.

dnephin (approved):
LGTM

dnephin:
> What about doing something like this?

I think the interface proposed at the bottom of #10634 does a better job of demonstrating the operations, and lets us keep more of the logic in a centralized place. With this approach the caller doesn't need to know anything about the states, or what transitions are allowed. 

I think there are definitely similarities between the two proposals.

> I think we should have the state managed internally through clear interfaces 

Definitely agree with this, but I think `set`, `revert` is too low level. It forces the caller to be much more aware of the states. 

Instead of `validateState` we should provide the caller a function to restore to the appropriate state. That is called `Complete` in the proposal in #10634

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/422659.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/422703.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit b49503682319b25734e198c77954b6ff5b8d41e1 onto `release/1.10.x` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit b49503682319b25734e198c77954b6ff5b8d41e1 onto `release/1.9.x` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit b49503682319b25734e198c77954b6ff5b8d41e1 onto `release/1.8.x` succeeded!

PR #10632, opened 2021-07-16
acls: Update ACL authorizer to return meaningful permission when ACLs are disabled

eculver:
As part of the discussion in [#10399](https://github.com/hashicorp/consul/pull/10399#discussion_r670020574), we thought that we could clean up the main token resolution API to return a non-nil value for the case where ACLs are disabled so that users don't need to do counter-intuitive nil checks on the result.

By changing this API, we had to update a few cases that were making assumptions out of the `nil` return. I think this makes more sense and cleans up the code ever so slightly. Curious if others agree. 

Note: the primary change is the one in `agent/consul/acl.go` -- everything else is fallout.

### DIFF
--- a/agent/structs/intention.go
+++ b/agent/structs/intention.go
@@ -325,7 +325,7 @@ func (ixn *Intention) CanRead(authz acl.Authorizer) bool {
 }
 
 func (ixn *Intention) CanWrite(authz acl.Authorizer) bool {
-	if authz == nil {
+	if authz == nil || authz == acl.ManageAll() {
 		return true
 	}
 	var authzContext acl.AuthorizerContext
--- a/agent/acl.go
+++ b/agent/acl.go
@@ -75,6 +75,7 @@ func (a *Agent) vetServiceRegisterWithAuthorizer(authz acl.Authorizer, service *
 
 // vetServiceUpdate makes sure the service update action is allowed by the given
 // token.
+// TODO: move to test package
 func (a *Agent) vetServiceUpdate(token string, serviceID structs.ServiceID) error {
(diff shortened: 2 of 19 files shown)


agent/structs/intention.go
```diff
@@ -325,7 +325,7 @@ func (ixn *Intention) CanRead(authz acl.Authorizer) bool {
 }
 
 func (ixn *Intention) CanWrite(authz acl.Authorizer) bool {
-	if authz == nil {
+	if authz == nil || authz == acl.ManageAll() {
```
mkeeler, line 328: Can the `authz` somehow still be nil here?

Also curious as to what checking against a specific authz singleton provides for us here. IIUC without this check it should fall through and call `authz.IntentionWrite` on line 338 which would be allowed with the `ManageAll` authorizer anyways.

mkeeler (approved):
Overall, I think these changes make sense and definitely reduce all the boilerplate code quite a bit. I left one comment but its very minor as I think the code is still functionally correct as it is.

dnephin:
I think we could also potentially remove any other `authz == nil` checks in the code, but that could just as easily be done in a separate PR some other time.

agent/structs/intention.go
dnephin: I think that's right. I'm going to do a second pass after this to remove more of the nil checks that should no longer be necessary.

dnephin:
The one test failure is a flake so I'm going to merge this

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/418921.

PR #10645, opened 2021-07-19
Reset root prune interval after TestLeader_CARootPruning completes

freddygv:
Aims to address this failure:
https://app.circleci.com/pipelines/github/hashicorp/consul/20339/workflows/dcdee545-794f-4a80-8d42-f359b79fdd4f/jobs/412315/tests#failed-test-0

Failing assertion:
https://github.com/hashicorp/consul/blob/766d62e4918050d7576f530daa2eb50c32cab251/agent/consul/leader_connect_test.go#L494

Relevant log:

> TestLeader_SecondaryCA_IntermediateRefresh-node-377.server.connect: pruning old unused root CA: <id>

This test ran after TestLeader_CARootPruning, which sets the pruning interval to 200ms. Since it is set to run in parallel, there could have been a long enough delay between setting and reading the CA config to trigger a prune.

### DIFF
--- a/agent/consul/leader_connect_test.go
+++ b/agent/consul/leader_connect_test.go
@@ -1173,7 +1173,12 @@ func TestLeader_CARootPruning(t *testing.T) {
 	}
 
 	// Can not use t.Parallel(), because this modifies a global.
+	origPruneInterval := caRootPruneInterval
 	caRootPruneInterval = 200 * time.Millisecond
+	t.Cleanup(func() {
+		// Reset the value of the global prune interval so that it doesn't affect other tests
+		caRootPruneInterval = origPruneInterval
+	})
 
 	require := require.New(t)
 	dir1, s1 := testServer(t)
@@ -1229,9 +1234,6 @@ func TestLeader_CARootPruning(t *testing.T) {
 	require.Len(roots, 1)
 	require.True(roots[0].Active)
 	require.NotEqual(roots[0].ID, oldRoot.ID)
-
-	// Reset the value of the global prune interval so that it doesn't affect other tests
-	caRootPruneInterval = 1 * time.Hour
 }
 
 func TestLeader_PersistIntermediateCAs(t *testing.T) {
--- a/agent/consul/leader_connect_test.go
+++ b/agent/consul/leader_connect_test.go
@@ -592,8 +592,6 @@ func TestLeader_SecondaryCA_IntermediateRefresh(t *testing.T) {
 
 	t.Parallel()
 
-	require := require.New(t)
-
 	dir1, s1 := testServerWithConfig(t, func(c *Config) {
 		c.Build = "1.6.0"
 	})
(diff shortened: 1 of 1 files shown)


agent/consul/leader_connect_test.go
```diff
@@ -653,16 +651,16 @@ func TestLeader_SecondaryCA_IntermediateRefresh(t *testing.T) {
 		}
 		var reply interface{}
 
-		require.NoError(s1.RPC("ConnectCA.ConfigurationSet", args, &reply))
+		require.NoError(t, s1.RPC("ConnectCA.ConfigurationSet", args, &reply))
 	}
 
 	var updatedRoot *structs.CARoot
-	{
+	retry.Run(t, func(r *retry.R) {
```
dnephin, line 658: If I understand correctly, what we are waiting on is for the `UpdateConfiguration` to perform the root rotation. 

Do you know why that is not completed synchronously? I thought it should be complete by the time the `RPC` call above returned.

It might be worth it to comment on this line on what async operation we are waiting on.

dnephin (commented):
Thanks for working on fixing this flaky test!

agent/consul/leader_connect_test.go
freddygv: That's a great point. I thought there was some asynchronous handling but I'm not seeing any after looking closer.

I also just noticed this log line in the CI failure:

> TestLeader_SecondaryCA_IntermediateRefresh-node-377.server.connect: pruning old unused root CA: id=`5f:de:2d:bc:45:23:14:dd:48:85:40:e9:58:f8:6e:e2:ad:31:ef:ae`

Maybe the reason why there was only two roots is because the old one was pruned already, but the really strange thing there is that the `caRootPruneInterval` is set to an hour, so we shouldn't have hit the pruning function to begin with.

That aside, maybe the assertion here could be relaxed to check that the active root doesn't match the original, since we just set a new CA configuration.

What do you think?

dnephin: Aha! That's an excellent lead. It looks like `TestLeader_CARootPruning` sets the value to 200 milliseconds, but never resets the old value.  So any test that runs after `TestLeader_CARootPruning` could run into this pruning. That seems like a good thing to fix.

If we fix that pruning interval back to 1 hour, maybe we don't need to make any changes to this test? Relaxing the assertion might also be ok, but I generally like to be confident in the underlying cause of the flakiness. Relaxing the assertion would make the test less flaky, but could hide other bugs.

freddygv: Ah, good find. I reverted the retry and added a commit to reset that value to an hour.

freddygv: From looking at the CI logs, TestLeader_CARootPruning did run first

agent/consul/leader_connect_test.go
```diff
@@ -1173,7 +1173,12 @@ func TestLeader_CARootPruning(t *testing.T) {
 	}
 
 	// Can not use t.Parallel(), because this modifies a global.
+	origPruneInterval := caRootPruneInterval
 	caRootPruneInterval = 200 * time.Millisecond
+	t.Cleanup(func() {
```
freddygv, line 1178: That looks good, thanks!

dnephin (approved):
Thanks! I pushed another commit to move the reset into a `t.Cleanup` function. This way the reset will happen even when the test fails. I also had it capture the original value. Setting it back to 1 hour should be fine too, but I think capturing/resetting the original is generally safer. Doing it that way here should allow others to copy the pattern.

Assuming you're ok with that change this LGTM.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/416549.

PR #10511, opened 2021-06-28
ui: CopyButton amends

johncowen:
During https://github.com/hashicorp/consul/pull/10285 we noticed that our CopyButton needed a few tweaks, luckily we have also decided to update how this works slightly.

This PR refactors the copy-to-clipboard functionality out of our CopyButton component and into a 'non-visual' modifier where it probably belongs.

- [with-copyable](https://consul-ui-staging-defwazcw9-hashicorp.vercel.app/ui/docs/modifiers/with-copyable)
- [CopyButton](https://consul-ui-staging-defwazcw9-hashicorp.vercel.app/ui/docs/components/copy-button)

Benefits:

- We don't have to reuse our CopyButton component if we want a method of copying text that doesn't look the same as our CopyButton (design vs behaviour)
- Our CopyButton component (as it already uses our StateChart component to deal with behaviour/state) is now almost a lighter-weight templateOnly component, and we should be able to make it 100% template only pretty soon.

We also:

- Moved our styles for this from our 'all-the-buttons' CSS file into the copy-button component by itself (resisted tweaking this for the moment as we are looking to change the design/look and feel of it)
- Added all the copy to our translations
- Added documentation for both `{{with-copyable}}` and `<CopyButton />`, including some info on our acceptance test steps that we use specifically for testing copying into our documentation (more related things to come here 🤞 )

Redesign will be in a follow up PR.

PR #8117, opened 2020-06-16
ui: Token listing redesign

johncowen:
This PR is a redesign of the token listing page.

Screenshot 2020-06-16 at 13 01 15

This necessitated various things, and I'll be adding various self-review comments to try to aid review here:

1. Slight reworking of `%composite-list` to add a CSS `grid` layout to allow us to have a right hand grid cell for the actions button.
2. Mirrored the CSS grid layout with 3 new block slots for `ListComponent`
3. Migrated our `copy-button` component to use `StateChart` and `Tooltip` components. This also meant we could delete the `FeedbackDialog` component whose functionality was pretty much a state machine anyway.
4. Move all usages of `ListComponent` into higher order components named by their 'consul type' i.e. `ConsulSomethingList` and migrate those to use the new block slots.

Potential future work/improvements:

1. Tooltips seem to be slightly problematic in hard to realize areas, such as control over accessibility features and testing.
2. Add pageobjects and READMEs to all new presentational components.
3. Move all references to `%more-popover-menu` to the new CSS files specifically for this component.

I've PRed this now as the changeset was growing more than I'd like, but I'll likely to be looking into ^ at some point in the future.

PR #10655, opened 2021-07-20
ui: Use native functionality for clipboard functionality, but keep a fallback

johncowen:
We recently re-wrote our copy to clipboard functionality to use a non-visual modifier, more details in https://github.com/hashicorp/consul/pull/10511.

@nfagerlund pointed out that we could be using the newer native browser functionality here, so we decided to tweak this a little more to use that instead of ClipboardJS. Our targeted browsers still fall within versions that do not support the newer native browser functionality as yet, so we keep a ClipboardJS polyfill here just incase, but only load that code for folks without `window.navigator.clipboard` support. End user (engineer) usage is the same for both.

Additionally we noticed we actually ignored our tests around this due to ember-tooltip being problematic https://github.com/hashicorp/consul/pull/8117#discussion_r440800963

We got rid of ember-tooltip a while back and replaced it with our own tooltip modifier, so we un-ignored that test here also (`health-check` spelling had also changed since then)

We made the service injection slightly less brittle here seeing as we are now switching between 3 implementations instead of just 2, but we can probably make this even better if we move to this to an addon (addons have features/capabilities that aren't available to applications). There also seems to be some interest from other products/folks in using this so we'll leave this as a relatively soon "to be done"/ additional PR.

~I've just noticed one last thing that is potentially a non-user facing engineering bug also 👀 , but that shouldn't block review.~ Ah no it failed tests, awesome, will fix that up now.

P.S. just incase, I've added a no changelog label here also, but I'll add in the final PR where we actually change the user facing feature


kaxcode (changes requested):
tests are failing

kaxcode (approved):
LGTM

PR #10657, opened 2021-07-20
config raft apply silent error

### LINKED ISSUE #10656: Applying Connect CA configuration to raft is silently failing when an invalid `ModifyIndex` is provided
#### Overview of the Issue
When applying a configuration and providing a `ModifyIndex`, if the provided `ModifyIndex` do not much the one stored in raft the new configuration will not be applied but the error is never reported back to the API call.

The issue is related to the fact that we return a `nil` error in [here](https://github.com/hashicorp/consul/blob/c271308473dd04e13567b894867547d8d76fd670/agent/consul/state/connect_ca.go#L159-L161)

dhiaayachi:
Fix #10656

### DIFF
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -909,6 +909,9 @@ func (c *CAManager) UpdateConfiguration(args *structs.CARequest) (reterr error) 
 		if respErr, ok := resp.(error); ok {
 			return respErr
 		}
+		if respOk, ok := resp.(bool); ok && !respOk {
+			return errors.New("configuration not applied")
+		}
 
 		// If the config has been committed, update the local provider instance
 		cleanupNewProvider = false
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -909,9 +909,6 @@ func (c *CAManager) UpdateConfiguration(args *structs.CARequest) (reterr error) 
 		if respErr, ok := resp.(error); ok {
 			return respErr
 		}
-		if respOk, ok := resp.(bool); ok && !respOk {
-			return errors.New("configuration not applied")
-		}
 
 		// If the config has been committed, update the local provider instance
 		cleanupNewProvider = false
--- a/agent/consul/state/connect_ca.go
+++ b/agent/consul/state/connect_ca.go
@@ -143,7 +143,7 @@ func (s *Store) CASetConfig(idx uint64, config *structs.CAConfiguration) error {
 
 // CACheckAndSetConfig is used to try updating the CA configuration with a
 // given Raft index. If the CAS index specified is not equal to the last observed index
-// for the config, then the call is a noop,
+// for the config, then the call will return an error,
 func (s *Store) CACheckAndSetConfig(idx, cidx uint64, config *structs.CAConfiguration) (bool, error) {
 	tx := s.db.WriteTxn(idx)
 	defer tx.Abort()
--- a/agent/consul/state/connect_ca.go
+++ b/agent/consul/state/connect_ca.go
@@ -2,7 +2,9 @@ package state
 
 import (
 	"fmt"
+
 	"github.com/hashicorp/go-memdb"
+	"github.com/pkg/errors"
 
 	"github.com/hashicorp/consul/agent/structs"
 )
@@ -157,7 +159,7 @@ func (s *Store) CACheckAndSetConfig(idx, cidx uint64, config *structs.CAConfigur
 	// return early here.
 	e, ok := existing.(*structs.CAConfiguration)
 	if (ok && e.ModifyIndex != cidx) || (!ok && cidx != 0) {
-		return false, nil
+		return false, errors.Errorf("invalid index")
 	}
 
 	if err := s.caSetConfigTxn(idx, tx, config); err != nil {
--- a/agent/consul/state/connect_ca.go
+++ b/agent/consul/state/connect_ca.go
@@ -159,7 +159,7 @@ func (s *Store) CACheckAndSetConfig(idx, cidx uint64, config *structs.CAConfigur
 	// return early here.
 	e, ok := existing.(*structs.CAConfiguration)
 	if (ok && e.ModifyIndex != cidx) || (!ok && cidx != 0) {
-		return false, errors.Errorf("invalid index")
+		return false, errors.Errorf("ModifyIndex did not match existing")
 	}
 
 	if err := s.caSetConfigTxn(idx, tx, config); err != nil {
--- a/agent/consul/state/connect_ca_test.go
+++ b/agent/consul/state/connect_ca_test.go
@@ -4,6 +4,8 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/hashicorp/consul/sdk/testutil"
+
 	"github.com/hashicorp/go-memdb"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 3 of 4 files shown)


agent/consul/leader_connect_ca.go
```diff
@@ -909,6 +909,9 @@ func (c *CAManager) UpdateConfiguration(args *structs.CARequest) (reterr error)
 		if respErr, ok := resp.(error); ok {
 			return respErr
 		}
+		if respOk, ok := resp.(bool); ok && !respOk {
+			return errors.New("configuration not applied")
```
banks, line 913: Perhaps we could make this more descriptive so operators understand why not?
```suggestion
			return errors.New("CA configuration not applied as the given ModifyIndex is older than the current one")
```

agent/consul/state/connect_ca.go
```diff
@@ -157,7 +159,7 @@ func (s *Store) CACheckAndSetConfig(idx, cidx uint64, config *structs.CAConfigur
 	// return early here.
 	e, ok := existing.(*structs.CAConfiguration)
 	if (ok && e.ModifyIndex != cidx) || (!ok && cidx != 0) {
-		return false, nil
+		return false, errors.Errorf("invalid index")
```
banks, line 162: Hmm I prefer returning an explicit error although I think the wording could be a little different - this is somewhat 'expected' in that the user requested "Check and Set" so it's not that their index is _invalid_ so much as that their requested check has failed and their write may have conflicted with another one.

Perhaps a message like `CAS failure: ModifyIndex did not match existing` would be clearer and also easier for consumers to detect since their behviour in this case might be to retry the read-modify-write rather than this being a fatal error.

banks: Are there any other callers to this other than the `CAManager` one we modified above?

agent/consul/leader_connect_ca.go
banks: I wonder how this will affect the HTTP API - did you try it? I assume this error just gets returned to the user.

I notice that we don't document the Check-and-set behaviour at all in this specific API endpoint anyway (maybe another thing to fix) https://www.consul.io/api-docs/connect/ca#update-ca-configuration.

Could use something similar to the words in Config Entries: https://www.consul.io/api-docs/config#cas

Looking though all other endpoints that have optional CAS semantics do so with an explicit field like `cas` rather than just implicitly on the ModifyIndex. I wonder why we choose to do that in this case and then not document it?

Perhaps we should make it explicit instead and ignore `ModifyIndex` entirely which is probably what operators assume we would do and the same as other similar endpoints.

banks (commented):
Some thoughts inline, what do you think @dhiaayachi ?

agent/consul/state/connect_ca_test.go
```diff
@@ -59,8 +59,8 @@ func TestStore_CAConfigCAS(t *testing.T) {
 	ok, err := s.CACheckAndSetConfig(2, 0, &structs.CAConfiguration{
 		Provider: "static",
 	})
-	if ok || err != nil {
-		t.Fatalf("expected (false, nil), got: (%v, %#v)", ok, err)
+	if ok || err == nil {
+		t.Fatalf("expected (false, \"invalid index\"), got: (%v, %#v)", ok, err)
```
dnephin, line 63: I think it's always a good idea to at least check part of the error message. Otherwise when the code is changed the test could return an error for a different reason, and the code path that you expected to be testing is no longer covered.

```suggestion
require.False(t, ok)
testutil.RequireErrorContains(t, err, "ModifyIndex did not match existing")
```

agent/consul/leader_connect_ca.go
```diff
@@ -909,6 +909,9 @@ func (c *CAManager) UpdateConfiguration(args *structs.CARequest) (reterr error)
 		if respErr, ok := resp.(error); ok {
 			return respErr
 		}
+		if respOk, ok := resp.(bool); ok && !respOk {
```
dnephin, line 912: I think in practice this code path will not be hit because the FSM operation will return an error. Even during an upgrade this code probably won't be relevant because the leader handles both operations.

I think unnecessary code like this can be a distraction and can lead to other problems, so I suggest we do not add this handling.

dnephin (commented):
Thanks for working on this fix!

agent/consul/leader_connect_ca.go
dhiaayachi: I'm not sure that the error should be specific to the `ModifyIndex` in here, by returning an error in connect_ca.go this will not be triggered and act as a fail safe/generic error in case another case happen where we return a nil error with a false. My understanding is in this case the false response mean that we did not modify the config and need to be associated with an error that state the reason, this is the case that we did not write the config but for no reason, which should never happen.

I agree on being explicit it will make it clear and easier to document. 
Is there any concern with retro compatibility if we decide to add a `cas` field?

agent/consul/state/connect_ca.go
dhiaayachi: This is only triggered by leader_connect_ca.go in 3 places:
- UpdateConfiguration (the one we are fixing)
- initializeCAConfig which read the config from raft copy it and modify it and write it again
- initializeRootCA which add the root ca to the config and write it again

agent/consul/leader_connect_ca.go
dhiaayachi: The issue is that the FSM code is error prone, the fact that we return a flag and an error and the flag is not really used can lead to other errors as this one. 
Another possibility would be to return always false or true so it's explicit that the response is not used for that specific operation

agent/consul/state/connect_ca.go
dnephin: Don't forget that `Store.CACheckAndSetConfig` is also called by the `CAOpSetRootsAndConfig` operation, which is called in two other places:
* a second place in `UpdateConfiguration`
* `CAManager.persistNewRootAndConfig` called by `initializeSecondaryCA` and `RenewIntermediate`, although `RenewIntermediate` will use a `nil` config, so should always use the existing stored config and not hit this bug.

agent/consul/leader_connect_ca.go
dnephin: I think the `bool` return value is the problem. The error should be able to indicate if it was success or not. The bool is a distraction that we should either remove or ignore.

dhiaayachi: I will remove this from here and create another PR to refactor it to be less error prone.

agent/consul/leader_connect_ca.go
dhiaayachi: I ended up removing this, I opened #10660 to refactor and remove the bool response

dnephin (approved):
LGTM! Thanks for fixing this.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/414411.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit c6859b3fb07031e3ea1fcd0230bacb070315d93d onto `release/1.10.x` succeeded!

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit c6859b3fb07031e3ea1fcd0230bacb070315d93d onto `release/1.9.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/414411)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit c6859b3fb07031e3ea1fcd0230bacb070315d93d onto `release/1.8.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/414411)

Issue #10629, opened 2021-07-16
can not build ui target (make ui) on Linux


zhangguanzhang:
step:
```
$ git clone https://github.com/hashicorp/consul.git
$ cd consul
$ git checkout v1.8.14
$ export XC_OS=linux XC_ARCH=amd64
$ make static-assets-docker
Building Golang build container
sha256:266a702eeda31911c04861b120285d775694e1b9298817632466ea728dd3c931
==> Building Static Assets
Creating the Go Build Container with image: consul-build-go
Copying the sources from '/root/go/src/github.com/hashicorp/consul/(pkg/web_ui|GNUmakefile)' to /consul/pkg
tar: pkg/web_ui: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
Running build in container
fatal: not a git repository (or any of the parent directories): .git
bindata: Failed to stat input path 'pkg/web_ui': lstat pkg/web_ui: no such file or directory
Error: go-bindata:  exit status 1
make: *** [GNUmakefile:309: static-assets] Error 1
make: *** [static-assets-docker] Error 1

```

mikemorris:
`make static-assets-docker` is not meant to be invoked directly, try building the UI with `make ui` instead?

zhangguanzhang:
```
[root@centos7 consul]# export DOCKER_BUILD_QUIET=0
[root@centos7 consul]#  export XC_OS=linux XC_ARCH=amd64
[root@centos7 consul]# make ui
...
==> Building Static Assets
Creating the Go Build Container with image: consul-build-go
Copying the sources from '/root/go/src/github.com/hashicorp/consul/(pkg/web_ui|GNUmakefile)' to /consul/pkg
Running build in container
fatal: not a git repository (or any of the parent directories): .git   <======= error
bindata_assetfs.go
Copying back artifacts

```

jkirschner-hashicorp:
@mikemorris : when building on Ubuntu 20.04 LTS under Windows WSL, I get the same error message as @zhangguanzhang:

```
// I seem to need sudo because of Windows WSL to connect to the Docker daemon socket
$ sudo make ui 

...

==> UI Built with Version: 1.10.0-dev (5a7c39a01), Logo: oss
Building Golang build container
sha256:d36236a1d37aaa314794f22886a2f96bf4362aac80de5c85b6b0ef0408a63140
==> Building Static Assets
Creating the Go Build Container with image: consul-build-go
[...]

PR #10659, opened 2021-07-20
Remove misleading "not a git repo" build message

jkirschner-hashicorp:
**Context**
Resolves confusion in build output seen in #10629.

Some build targets previously generated a misleading build message even on
successful completion:
```
$ make ui
...
==> Building Static Assets
Creating the Go Build Container with image: consul-build-go
Copying the sources from '/root/go/src/github.com/hashicorp/consul/(pkg/web_ui|GNUmakefile)' to /consul/pkg
Running build in container
========> fatal: not a git repository (or any of the parent directories): .git  <======== error
bindata_assetfs.go
Copying back artifacts
```
The GNUmakefile uses git to extract repo status information for use in the
build. Because some builds occur within a docker container in which the .git
folder is not available, these git commands would fail and generate the error
message above, even though the build can succeed without this repo status
information.

For example, "make ui" causes "make static-assets" to be built inside a docker
container into which only the GNUmakefile and pkg/web_ui have been copied.

**Open Questions**
- [ ] What should the value of `GIT_DIRTY` be if the repo is pristine? It looks like the code meant to set it to `true`, but I think it's not using the `test` command correctly... shouldn't it be `|| echo true`, not `|| true`?
- [ ] What should the value of `GIT_DIRTY` be if building inside a docker container with no access to the `.git` folder? The value I currently have probably isn't right.
- [ ] Does this break any expectations of existing build targets for the values of the `GIT_*` variables?
- [ ] The code that sets `GIT_DIRTY` seems to also exist in `.circleci/bash_env.sh` and `build-support/functions/10-util.sh`. Is there a reason those shouldn't also be changed (and ideally made DRY)?
- [ ] I don't think this needs a changelog entry ... is that correct?

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -17,10 +17,25 @@ GOPATH=$(shell go env GOPATH)
 MAIN_GOPATH=$(shell go env GOPATH | cut -d: -f1)
 
 ASSETFS_PATH?=agent/uiserver/bindata_assetfs.go
-# Get the git commit
+
+# If running inside a git repository, store repo status info in environment variables
+IS_INSIDE_GIT_REPO="$(git rev-parse --is-inside-work-tree 2>/dev/null)"
+GIT_DIRTY_DEFAULT=true
+ifeq (${IS_INSIDE_GIT_REPO},1)
 GIT_COMMIT?=$(shell git rev-parse --short HEAD)
 GIT_COMMIT_YEAR?=$(shell git show -s --format=%cd --date=format:%Y HEAD)
-GIT_DIRTY?=$(shell test -n "`git status --porcelain`" && echo "+CHANGES" || true)
+GIT_DIRTY?=$(shell test -n "`git status --porcelain`" && echo "+CHANGES" || ${GIT_DIRTY_DEFAULT})
+else
+# In not running inside a git repository, define default repo status environment variables
+# This can happen if building a target inside a docker container in which this
+# GNUmakefile, but not the .git folder, are available. For example, "make ui"
+# will cause "make static-assets" to be built inside a docker container into
+# which only this file and pkg/web_ui have been copied.
+GIT_COMMIT?=
+GIT_COMMIT_YEAR?=
+GIT_DIRTY?=${GIT_DIRTY_DEFAULT}
+endif
+
 GIT_IMPORT=github.com/hashicorp/consul/version
 GOLDFLAGS=-X $(GIT_IMPORT).GitCommit=$(GIT_COMMIT)$(GIT_DIRTY)
 


GNUmakefile
```diff
@@ -17,10 +17,25 @@ GOPATH=$(shell go env GOPATH)
 MAIN_GOPATH=$(shell go env GOPATH | cut -d: -f1)
 
 ASSETFS_PATH?=agent/uiserver/bindata_assetfs.go
-# Get the git commit
+
+# If running inside a git repository, store repo status info in environment variables
+IS_INSIDE_GIT_REPO="$(git rev-parse --is-inside-work-tree 2>/dev/null)"
+GIT_DIRTY_DEFAULT=true
```
jkirschner-hashicorp, line 23: What should the value of `GIT_DIRTY` be if the repo is pristine? It looks like the code meant to set it to `true`, but I think it's not using the test command correctly... shouldn't it be `|| echo true`, not `|| true`?

If I'm correct, I'd need to fix my next version of this line as well to include the `echo` after the `||`.

 What should the value of GIT_DIRTY be if building inside a docker container with no access to the .git folder? The value I currently have probably isn't right.

GNUmakefile
```diff
@@ -17,10 +17,25 @@ GOPATH=$(shell go env GOPATH)
...
+GIT_DIRTY?=$(shell test -n "`git status --porcelain`" && echo "+CHANGES" || ${GIT_DIRTY_DEFAULT})
+else
+# In not running inside a git repository, define default repo status environment variables
+# This can happen if building a target inside a docker container in which this
+# GNUmakefile, but not the .git folder, are available. For example, "make ui"
+# will cause "make static-assets" to be built inside a docker container into
+# which only this file and pkg/web_ui have been copied.
+GIT_COMMIT?=
+GIT_COMMIT_YEAR?=
+GIT_DIRTY?=${GIT_DIRTY_DEFAULT}
```
jkirschner-hashicorp, line 36: What should the value of `GIT_DIRTY` be if building inside a docker container with no access to the .git folder? The value I currently have probably isn't right.

dnephin (commented):
Thanks for working on a fix for this problem!

I think this problem is a great example of why Makefiles, despite being widely used, are actually a really poor fit for this use case.  There are many reasons Makefiles are a bit fit, the particular reason in this case is that variables in Makefiles must be outside of directives, and that forces them to be execute every time. Most of the time they aren't needed.

Instead of adding yet more complexity to this Makefile, I think we should look at fixing the underlying problem. These `shell` commands are running shell scripts anyway. We could move the `GOLDFLAGS` and `test-docker` into shell scripts and then they would not be executed at the wrong time.

johncowen:
Guessing I'm a reviewer here as it affects me. Looks fine to me, but I'll leave it with daniel for proper review, thanks for the heads up though!

Umm I can give you my answer for:

> don't think this needs a changelog entry ... is that correct?

Yeah agree with that, I don't think it needs one, it's not user facing and doesn't affect anything user facing.

Lastly, I've not looked too closely at this, but it sounds like that fatal error has no real affect on anything anyway, so just redirecting that might be fine for folks also? You're right it confuses folks when they see it (not sure it it makes the target exit with an error actually 🤔 ?), so just making it not appear might be the solution here? Me personally, I hardly ever run this.

jkirschner-hashicorp:
@johncowen : redirecting the error output was my first approach, but I decided against it because it would hide any real errors in those commands (e.g., if they were ever changed, if `git` wasn't available when it should be, etc).

@dnephin : one advantage of defining the variables in the makefile is that they are always present (if there's a `.git` folder), so someone modifying the scripts doesn't have to remember to define them everywhere they are needed. That said, I still agree with you that simplifying the makefile is probably the better approach overall. I'm not very familiar with the various targets and how they depend (or not) on these GIT_* environment variables, and I also don't have a lot of spare cycles to revisit this right now, so this PR might be dormant for a bit if I'm the one pushing it forward.

jmurret:

PR #10660, opened 2021-07-20
refactor CACheckAndSetConfig to not return unnecessary bool value

dhiaayachi:
This PR is to refactor `CACheckAndSetConfig` to not return unnecessary bool value. 
The reason behind this change is the bug filed in #10656 and prevent other bugs like this to be introduced in `CACheckAndSetConfig`

### DIFF
--- a/agent/connect_ca_endpoint.go
+++ b/agent/connect_ca_endpoint.go
@@ -98,10 +98,10 @@ func (s *HTTPHandlers) ConnectCAConfigurationSet(req *http.Request) (interface{}
 			}
 		}
 		args.Op = structs.CAOpSetConfigCAS
-		args.Cas = casVal
+		args.ConfigCASIndex = casVal
 	} else {
 		args.Op = structs.CAOpSetConfig
-		args.Cas = 0
+		args.ConfigCASIndex = 0
 	}
 	if err := decodeBody(req.Body, &args.Config); err != nil {
 		return nil, BadRequestError{
--- a/agent/connect_ca_endpoint.go
+++ b/agent/connect_ca_endpoint.go
@@ -98,10 +98,10 @@ func (s *HTTPHandlers) ConnectCAConfigurationSet(req *http.Request) (interface{}
 			}
 		}
 		args.Op = structs.CAOpSetConfigCAS
-		args.Index = casVal
+		args.Cas = casVal
 	} else {
 		args.Op = structs.CAOpSetConfig
-		args.Index = 0
+		args.Cas = 0
 	}
 	if err := decodeBody(req.Body, &args.Config); err != nil {
 		return nil, BadRequestError{
--- a/agent/connect_ca_endpoint.go
+++ b/agent/connect_ca_endpoint.go
@@ -89,6 +89,15 @@ func (s *HTTPHandlers) ConnectCAConfigurationSet(req *http.Request) (interface{}
 	var args structs.CARequest
 	s.parseDC(req, &args.Datacenter)
 	s.parseToken(req, &args.Token)
+	if casStr := req.URL.Query().Get("cas"); casStr != "" {
+		casVal, err := strconv.ParseUint(casStr, 10, 64)
+		if err != nil {
+			return nil, BadRequestError{
+				Reason: fmt.Sprintf("Request decode failed: %v", err),
+			}
+		}
+		args.Cas = casVal
+	}
 	if err := decodeBody(req.Body, &args.Config); err != nil {
 		return nil, BadRequestError{
 			Reason: fmt.Sprintf("Request decode failed: %v", err),
--- a/agent/connect_ca_endpoint.go
+++ b/agent/connect_ca_endpoint.go
@@ -89,6 +89,7 @@ func (s *HTTPHandlers) ConnectCAConfigurationSet(req *http.Request) (interface{}
 	var args structs.CARequest
 	s.parseDC(req, &args.Datacenter)
 	s.parseToken(req, &args.Token)
+	// Check for cas value
 	if casStr := req.URL.Query().Get("cas"); casStr != "" {
 		casVal, err := strconv.ParseUint(casStr, 10, 64)
 		if err != nil {
@@ -96,7 +97,11 @@ func (s *HTTPHandlers) ConnectCAConfigurationSet(req *http.Request) (interface{}
 				Reason: fmt.Sprintf("Request decode failed: %v", err),
 			}
 		}
-		args.Cas = casVal
+		args.Op = structs.CAOpSetConfigCAS
+		args.Index = casVal
+	} else {
+		args.Op = structs.CAOpSetConfig
+		args.Index = 0
 	}
 	if err := decodeBody(req.Body, &args.Config); err != nil {
 		return nil, BadRequestError{
--- a/agent/consul/fsm/commands_oss.go
+++ b/agent/consul/fsm/commands_oss.go
@@ -423,7 +423,7 @@ func (c *FSM) applyConnectCAOperation(buf []byte, index uint64) interface{} {
 		[]metrics.Label{{Name: "op", Value: string(req.Op)}})
 	switch req.Op {
 	case structs.CAOpSetConfigCAS:
-		return c.state.CACheckAndSetConfig(index, req.Index, req.Config)
+		return c.state.CACheckAndSetConfig(index, req.Cas, req.Config)
 
 	case structs.CAOpSetConfig:
 		return c.state.CASetConfig(index, req.Config)
@@ -457,7 +457,7 @@ func (c *FSM) applyConnectCAOperation(buf []byte, index uint64) interface{} {
 			return act
 		}
 
-		return c.state.CACheckAndSetConfig(index, req.Index, req.Config)
+		return c.state.CACheckAndSetConfig(index, req.Cas, req.Config)
 	case structs.CAOpIncrementProviderSerialNumber:
 		sn, err := c.state.CAIncrementProviderSerialNumber(index)
 		if err != nil {
--- a/agent/consul/fsm/commands_oss.go
+++ b/agent/consul/fsm/commands_oss.go
@@ -424,12 +424,13 @@ func (c *FSM) applyConnectCAOperation(buf []byte, index uint64) interface{} {
 	switch req.Op {
 	case structs.CAOpSetConfig:
 		if req.Config.ModifyIndex != 0 {
-			act, err := c.state.CACheckAndSetConfig(index, req.Config.ModifyIndex, req.Config)
+			var emptyResp interface{}
+			err := c.state.CACheckAndSetConfig(index, req.Config.ModifyIndex, req.Config)
 			if err != nil {
 				return err
 			}
 
-			return act
+			return emptyResp
 		}
 
 		return c.state.CASetConfig(index, req.Config)
@@ -461,12 +462,12 @@ func (c *FSM) applyConnectCAOperation(buf []byte, index uint64) interface{} {
 		if !act {
 			return act
 		}
-
-		act, err = c.state.CACheckAndSetConfig(index, req.Config.ModifyIndex, req.Config)
+		var emptyResp interface{}
+		err = c.state.CACheckAndSetConfig(index, req.Config.ModifyIndex, req.Config)
 		if err != nil {
 			return err
 		}
-		return act
+		return emptyResp
 	case structs.CAOpIncrementProviderSerialNumber:
 		sn, err := c.state.CAIncrementProviderSerialNumber(index)
 		if err != nil {
--- a/agent/consul/fsm/commands_oss.go
+++ b/agent/consul/fsm/commands_oss.go
@@ -423,9 +423,9 @@ func (c *FSM) applyConnectCAOperation(buf []byte, index uint64) interface{} {
 		[]metrics.Label{{Name: "op", Value: string(req.Op)}})
 	switch req.Op {
 	case structs.CAOpSetConfig:
-		if req.Config.ModifyIndex != 0 {
+		if req.Cas != 0 {
 			var emptyResp interface{}
-			err := c.state.CACheckAndSetConfig(index, req.Config.ModifyIndex, req.Config)
+			err := c.state.CACheckAndSetConfig(index, req.Cas, req.Config)
 			if err != nil {
 				return err
 			}
@@ -463,7 +463,7 @@ func (c *FSM) applyConnectCAOperation(buf []byte, index uint64) interface{} {
 			return act
 		}
 		var emptyResp interface{}
-		err = c.state.CACheckAndSetConfig(index, req.Config.ModifyIndex, req.Config)
+		err = c.state.CACheckAndSetConfig(index, req.Cas, req.Config)
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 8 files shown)


agent/consul/state/connect_ca_test.go
```diff
@@ -79,11 +77,11 @@ func TestStore_CAConfigCAS(t *testing.T) {
...
 
 	// Do another CAS, this time with the correct index
-	ok, err = s.CACheckAndSetConfig(2, 1, &structs.CAConfiguration{
+	err = s.CACheckAndSetConfig(2, 1, &structs.CAConfiguration{
 		Provider: "static",
 	})
-	if !ok || err != nil {
-		t.Fatalf("expected (true, nil), got: (%v, %#v)", ok, err)
+	if err != nil {
+		t.Fatalf("expected (true, nil), got: (%#v)", err)
```
banks, line 84: Perhaps replace this if with `require.NoError(t, err)` instead? I was going to point out the message is a little wrong now as we don't expect a bool, so we should at least change that but I'd probably just move it to `require` at this point. Up to you though it might be odd to have a mix of assertion styles in this test 🤷 .

banks:
@dhiaayachi this looks good.

I'm wondering about docs: given that we've never documented the check and set behaviour here publicly I think we have a few options for how to improve that along with this fix.

 1. **Just document this in the API docs.** The one downside here is that it's a somewhat surprising and inconsistent API. In other places where we allow and optional CAS so the end-user can control concurrent updates we do so with an explicit parameter like `"cas": 1234` in the payload and ignore the `ModifyIndex`.
 2. **Change it so we ignore the `ModifyIndex` from user payload**. This probably means explicitly setting `ModifyIndex` to zero in the HTTP handler. I _don't think we should_ remove the check from the FSM code because I think the only reason it exists at all is because we rely on it in `InitializeCA` where we do direct read-modify-writes on the state store and we want to ensure those fail if anything races. That said, the locking mechanisms in `CAManager` should really also mean there is never any other code calling into this write path so maybe this option should be strip out the CAS code path entirely?
 3. **Add a `cas` field to the PUT endpoint and CLI.** So users can explicitly opt-in, also strip modify index. Then document it the same way we document KV and ConfigEntry API and CLI cas options.

I'm open to other options too though leaving this undocumented even if it's no longer buggy and silent feels a bit odd to me. Also open to deferring docs/API changes if we choose one of the options that is a little more work to a later PR so we at least fix the issue. Interested to hear what you think?

agent/consul/state/connect_ca_test.go
dhiaayachi: yes a `require.NoError(t, err)` is definitely nicer and we are converting as much as possible to use testify require. So, I will convert the whole test.

dhiaayachi:
> @dhiaayachi this looks good.
> 
> I'm wondering about docs: given that we've never documented the check and set behaviour here publicly I think we have a few options for how to improve that along with this fix.
> 
> 1. **Just document this in the API docs.** The one downside here is that it's a somewhat surprising and inconsistent API. In other places where we allow an optional CAS so the end-user can control concurrent updates we do so with an explicit parameter like `"cas": 1234` in the payload and ignore the `ModifyIndex`.
> 2. **Change it so we ignore the `ModifyIndex` from user payload**. This probably means explicitly setting `ModifyIndex` to zero in the HTTP handler. I _don't think we should_ remove the check from the FSM code because I think the only reason it exists at all is because we rely on it in `InitializeCA` where we do direct read-modify-writes on the state store and we want to ensure those fail if anything races. That said, the locking mechanisms in `CAManager` should really also mean there is never any other code calling into this write path so maybe this option should be strip out the CAS code path entirely?
> 3. **Add a `cas` field to the PUT endpoint and CLI.** So users can explicitly opt-in, also strip modify index. Then document it the same way we document KV and ConfigEntry API and CLI cas options.
> 
> I'm open to other options too though leaving this undocumented even if it's no longer buggy and silent feels a bit odd to me. Also open to deferring docs/API changes if we choose one of the options that is a little more work to a later PR so we at least fix the issue. Interested to hear what you think?

Thank you for taking a look at this @banks 

Solution 3 (Adding a `cas` field) is the one that sound right to me. It make the feature explicit and make it the same as other API/CLI in consul. That said, I'm not sure what is the approach to such an API change as I guess it will break compatibility and anybody relying on this API/CLI will have to make a change.

I think we should take a progressive approach toward it. The change in this PR and #10657 are a step toward fixing/avoiding a class of bug in this part of the code I can add a basic documentation into this PR so it can be released as part of the next minor release.

Next step would be to start a PR that make the API change (to use a `cas` field) and document it, this would be part of the next major release (assuming no deprecation needed)

What you think about this?

banks:
> That said, I'm not sure what is the approach to such an API change as I guess it will break compatibility and anybody relying on this API/CLI will have to make a change.

I think in this case it's less clear that this is a "breaking change". We never documented the CAS behaviour of Modify index field but more importantly it was completely broken anyway. If anyone had guessed and tried to use it, they would get a partial ca config change  and a success even though the new config didn't persist. In other words I'm pretty OK with changing the API to either remove ModifyIndex CAS or switch the field name in a point release as it's strictly less broken than now either way and isn't breaking any currently working functionality.

That said, if you would feel more comfortable deferring the cas change I'm fine with what you proposed too!

agent/consul/fsm/commands_oss.go
```diff
@@ -461,12 +462,12 @@ func (c *FSM) applyConnectCAOperation(buf []byte, index uint64) interface{} {
 		if !act {
 			return act
 		}
-
-		act, err = c.state.CACheckAndSetConfig(index+1, req.Config.ModifyIndex, req.Config)
+		var emptyResp interface{}
+		err = c.state.CACheckAndSetConfig(index+1, req.Config.ModifyIndex, req.Config)
```
dnephin, line 466: I believe this `index+1` is also a bug. It should be just `index`. We could fix it here at the same time, but also fine to leave it for a separate PR.

agent/consul/fsm/commands_oss.go
```diff
@@ -424,12 +424,13 @@ func (c *FSM) applyConnectCAOperation(buf []byte, index uint64) interface{} {
...
 		if req.Config.ModifyIndex != 0 {
-			act, err := c.state.CACheckAndSetConfig(index, req.Config.ModifyIndex, req.Config)
+			var emptyResp interface{}
+			err := c.state.CACheckAndSetConfig(index, req.Config.ModifyIndex, req.Config)
 			if err != nil {
 				return err
 			}
 
-			return act
+			return emptyResp
```
dnephin, line 433: I think this would be equivalent:

```go
return c.state.CACheckAndSetConfig(index, req.Config.ModifyIndex, req.Config)
```

agent/consul/fsm/commands_oss.go
```diff
@@ -461,12 +462,12 @@ func (c *FSM) applyConnectCAOperation(buf []byte, index uint64) interface{} {
...
 		}
-
-		act, err = c.state.CACheckAndSetConfig(index+1, req.Config.ModifyIndex, req.Config)
+		var emptyResp interface{}
+		err = c.state.CACheckAndSetConfig(index+1, req.Config.ModifyIndex, req.Config)
 		if err != nil {
 			return err
 		}
-		return act
+		return emptyResp
```
dnephin, line 470: same as above, I think this can be a single line.

dnephin (commented):
I think this is looking good. I agree with the discussion above that probably we want to also document this behaviour, and probably add an explicit `CAS` field as well.

Made some suggestions for the implementation below to reduce it a bit.

agent/consul/fsm/commands_oss.go
dhiaayachi: I was thinking of fixing this in the upstream branch so we can backport it.

agent/connect_ca_endpoint.go
```diff
@@ -89,6 +89,15 @@ func (s *HTTPHandlers) ConnectCAConfigurationSet(req *http.Request) (interface{}
...
 	s.parseDC(req, &args.Datacenter)
 	s.parseToken(req, &args.Token)
+	if casStr := req.URL.Query().Get("cas"); casStr != "" {
+		casVal, err := strconv.ParseUint(casStr, 10, 64)
+		if err != nil {
+			return nil, BadRequestError{
+				Reason: fmt.Sprintf("Request decode failed: %v", err),
+			}
+		}
+		args.Cas = casVal
```
dhiaayachi, line 99: not sure if we need to overrite ModifyIndex with `cas` value in here

agent/consul/leader_connect_ca.go
```diff
@@ -503,6 +504,7 @@ func (c *CAManager) initializeRootCA(provider ca.Provider, conf *structs.CAConfi
 		req := structs.CARequest{
 			Op:     structs.CAOpSetConfig,
 			Config: conf,
+			Cas:    conf.ModifyIndex,
```
dhiaayachi, line 507: I set `cas` to ModifyIndex in here because we are reading modifying and writing the config in one place, I'm not sure about this may be I need to set it to 0 to avoid the check.

agent/consul/leader_connect_ca.go
```diff
@@ -760,6 +762,7 @@ func (c *CAManager) persistNewRootAndConfig(provider ca.Provider, newActiveRoot
 		Index:  idx,
 		Roots:  newRoots,
 		Config: &newConf,
+		Cas:    newConf.ModifyIndex,
```
dhiaayachi, line 765: same here.
I set `cas` to ModifyIndex in here because we are reading modifying and writing the config in one place, I'm not sure about this may be I need to set it to 0 to avoid the check.

agent/consul/leader_connect_ca.go
```diff
@@ -1004,6 +1007,7 @@ func (c *CAManager) UpdateConfiguration(args *structs.CARequest) (reterr error)
 	args.Op = structs.CAOpSetRootsAndConfig
 	args.Index = idx
 	args.Config.ModifyIndex = confIdx
+	args.Cas = confIdx
```
dhiaayachi, line 1010: Same here
I set `cas` to ModifyIndex in here because we are reading modifying and writing the config in one place, I'm not sure about this may be I need to set it to 0 to avoid the check.

dhiaayachi (commented):
@banks @dnephin 
I added the code to add a `cas` parameter, I have some questions inline about how to set it (in the cases that it's not coming from the API.
Next, I will add the doc part.

banks:
I think this approach would work @dhiaayachi. I'd maybe consider following the way other similar places do this though just for the sake of consistency in the code. I think that would mean:

 * Remove the "implicit" CAS from FSM - the `CAOpSetConfig` should just always be a set regardless and ignore the `ModifyIndex` field if it's set on the argument.
 * Add an explicit `CAOpSetConfigCAS`
 * If the query param `?cas` is non-zero, then switch the Operation to `*CAS`
 * Re-use the `CARequest.Index` field which is documented as being specifically for CAS operations (this differs from ConfigEntrys which just uses the `ModifyIndex` directly but it makes sense to use a field that already exists for this purpose than introduce a different CAS pattern in the same request object.
     * That said, I see in the diff here that despite documenting `Index` as being the CAS index for CAS operations. we often set it to the current index which seems wrong. Perhaps we should check the actual behaviour and either update the docs or the existing call sites? I'd be open to deprecating it and naming it `CAS` if this is infact it's only use to avoid this confusion? 
 * Internal read-modify-writes can either use the new CAS operation to be defensive against concurrency bugs in our code, or can just use set without worrying about CAS and just rely on the fact that they are already serialized and the only component that writes this state in Consul currently. I'd be OK with either!

I think re-using the existing patterns of specifying CAS with a op and re-using the existing field make more sense but open to changing it.

dhiaayachi:
> * CAOpSetConfigCAS

Yes make sense to use the same pattern, I will implement the changes.
About the `CARequest.Index` it was used in internal operations in the following pattern:
- Read the config
- Modify the copy we read
- Write it and set the index to the Read config Index to make sure no one modified it in the mean time
So it's used for CAS but not for the op we are fixing (`CAOpSetConfig`)
I will make the change to use the same pattern for `CAOpSetConfig` (use index as `CAS`) and rename it to CAS to avoid confusion.
Doc will follow.

dhiaayachi:
> Is setting `Index: idx` correct above? If so why is that different to the line being removed? I know you didn't change this, I'm just trying to figure out if `Index` is actually being used properly above or if it was added by mistake because of the confusing name?

I just found out that while trying to fix the failing tests:
in the case of `CAOpSetRootsAndConfig` we do 2 CAS ops in 1 and this is why the `ModifyIndex` was used for `CACheckAndSetConfig`. Because in the case of CAOpSetRootsAndConfig we use `Index` for the `CARootSetCAS` operation so we can't use it for `CACheckAndSetConfig` because both have different indexes.
The last commit use `CASetConfig`  in the `CAOpSetRootsAndConfig` operation but I'm not sure it's ok because it's supposed to be a CAS operation.

dhiaayachi:
@banks @dnephin 
I think this is now ready for another round of reviews. I went back to the option where we have separate field to store the `CAOpSetConfigCAS` operation index

agent/structs/connect_ca.go
```diff
@@ -208,9 +209,12 @@ type CARequest struct {
 	// Datacenter is the target for this request.
 	Datacenter string
 
-	// Index is used by CAOpSetRoots and CAOpSetConfig for a CAS operation.
+	// Index is used by CAOpSetRoots for a CAS operation.
 	Index uint64
 
+	// Index is used by CAOpSetRootsAndConfig and CAOpSetConfigCAS for a CAS operation.
+	Cas uint64
```
dnephin, line 216: I think a separate field is necessary, but we could probably clarify the name here. Maybe `ConfigCASIndex`  ?

I think we should also clarify the comment
```suggestion
	// ConfigCASIndex is used as the check-and-set index for the CAOpSetConfigCAS operation, and for the
	// Config portion of the CAOpSetRootsAndConfig operation.
	ConfigCASIndex uint64
```

dnephin (commented):
I think this is looking good. Added one suggestion below for the field name and godoc string.

I think this PR will also need docs updates, to at least the api doc reference in `website/content/api-docs/connect/ca.mdx`. I'm not sure if anything else needs to be updated.

I'd like to do another detailed pass of the existing calls to make sure we've preserved the CAS behaviour from internal CA operations.

agent/consul/leader_connect_ca_test.go
```diff
@@ -85,26 +85,27 @@ func (m *mockCAServerDelegate) ApplyCARequest(req *structs.CARequest) (interface
...
 		if req.Config.ModifyIndex != 0 {
-			act, err := m.store.CACheckAndSetConfig(idx+1, req.Config.ModifyIndex, req.Config)
+			var emptyResp interface{}
+			err := m.store.CACheckAndSetConfig(idx+1, req.Config.ModifyIndex, req.Config)
 			if err != nil {
 				return nil, err
 			}
 
-			return act, nil
+			return emptyResp, nil
```
dnephin, line 94: This should mirror `FSM.applyConnectCAOperation` so that our tests match production (see the godoc comment I added to this method).

So I think this one should be simply:
```go
return c.state.CASetConfig(idx, req.Config)
```

and probably the cases below should also be updated to match `FSM.applyConnectCAOperation`

dhiaayachi:
Closing this as it's very stale and not likely to be merged

PR #10661, opened 2021-07-20
http: disable legacy ACL API endpoints

dnephin:
These 6 API endpoints will now return "410 Gone". 

A couple tests that used the legacy API endpoints had to be updated to use the new ACL endpoints.

I think we can add a single changelog entry for the whole set of changes related to removing the legacy ACL system, so omitting the changelog for now.


agent/consul/state/acl.go
```diff
@@ -563,6 +563,7 @@ func fixupRolePolicyLinks(tx ReadTxn, original *structs.ACLRole) (*structs.ACLRo
 }
 
 // ACLTokenSet is used to insert an ACL rule into the state store.
+// Deprecated (ACL-Legacy-Compat)
 func (s *Store) ACLTokenSet(idx uint64, token *structs.ACLToken, legacy bool) error {
```
rboyer, line 567: A+ for keeping these state store functions, since they'll be necessary to replay raft logs created from prior to the upgrade if someone was still using legacy operations right up until the 1.11 upgrade.

rboyer (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/428939.

PR #10664, opened 2021-07-21
ui: Give CopyButtons an idle tooltip as well as success/error ones

johncowen:
This PR builds upon https://github.com/hashicorp/consul/pull/10655 and adds a tooltip saying "Copy" to all our CopyButtons. 

The success notification is now simply "Copied" as we figured that copying something isn't something to get that excited about! 🎉 

Notes:

It would have been nice to try use a conditional modifier here to avoid double buttons. We began by making a single button with a single modifier with conditional arguments, but that didn't work for whatever reason. I quickly tried doubling up the button and using our State component, and that worked fine. So considering this is in one place and we might have the ability to do conditional modifiers soon (there's an RFC somewhere for `(modifer 'modifer-name')`), I decided to stick with it for this case. We can circle back here when/if the conditional modifier thing happens.

- [ ] Needs a changelog

### DIFF
--- a/ui/packages/consul-ui/app/components/copy-button/index.hbs
+++ b/ui/packages/consul-ui/app/components/copy-button/index.hbs
@@ -7,12 +7,25 @@
     ...attributes
   >
 {{#let (fn dispatch 'SUCCESS') (fn dispatch 'ERROR') (fn dispatch 'RESET') as |success error reset|}}
+<State @matches="idle">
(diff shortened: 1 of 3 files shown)


jnwright:
@johncowen @kaxcode Tooltip should be over the icon only

kaxcode (approved):
LGTM

PR #10671, opened 2021-07-21
subscribe: improve TestSubscribeBackend_IntegrationWithServer_DeliversAllMessages

dnephin:
`TestSubscribeBackend_IntegrationWithServer_DeliversAllMessages` has been flaking a few times. This commit cleans up the test a bit, and improves the failure output. Unfortunately it does not actually fix the flakiness, but at least it exposes more details about the problem.

The failure appears to be that the event with Port=0 is being sent in both the snapshot and as the first event after the EndOfSnapshot event. With the improved logging we can see that both events have the same index:

```
    subscribe_backend_test.go:409: subscriber 00004: saw canary in snapshot with port 0 at index 1019
    subscribe_backend_test.go:385: subscriber 00004: snapshot done at index 1019, expect next port to be 1
    subscribe_backend_test.go:409: subscriber 00000: saw canary in snapshot with port 0 at index 1019
    subscribe_backend_test.go:385: subscriber 00000: snapshot done at index 1019, expect next port to be 1
    subscribe_backend_test.go:409: subscriber 00001: saw canary in snapshot with port 0 at index 1019
    subscribe_backend_test.go:385: subscriber 00001: snapshot done at index 1019, expect next port to be 1
    subscribe_backend_test.go:409: subscriber 00002: saw canary in snapshot with port 0 at index 1019
    subscribe_backend_test.go:385: subscriber 00002: snapshot done at index 1019, expect next port to be 1
    subscribe_backend_test.go:409: subscriber 00003: saw canary in snapshot with port 0 at index 1019
    subscribe_backend_test.go:385: subscriber 00003: snapshot done at index 1019, expect next port to be 1
    subscribe_backend_test.go:318: 
                Error Trace:    subscribe_backend_test.go:318
                Error:          Received unexpected error:
                                subscriber 00004: at index 1019: expected port 1, got 0
                Test:           TestSubscribeBackend_IntegrationWithServer_DeliversAllMessages
```

### DIFF
--- a/agent/consul/subscribe_backend_test.go
+++ b/agent/consul/subscribe_backend_test.go
@@ -2,15 +2,16 @@ package consul
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"strings"
-	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/require"
+	"golang.org/x/sync/errgroup"
 	gogrpc "google.golang.org/grpc"
 
 	grpc "github.com/hashicorp/consul/agent/grpc"
@@ -303,31 +304,18 @@ func TestSubscribeBackend_IntegrationWithServer_DeliversAllMessages(t *testing.T
 	// Now start a whole bunch of streamers in parallel to maximise chance of
 	// catching a race.
 	n := 5
-	var wg sync.WaitGroup
+	var g errgroup.Group
 	var updateCount uint64
-	// Buffered error chan so that workers can exit and terminate wg without
-	// blocking on send. We collect errors this way since t isn't thread safe.
-	errCh := make(chan error, n)
 	for i := 0; i < n; i++ {
 		i := i
-		wg.Add(1)
-		go func() {
-			defer wg.Done()
-			verifyMonotonicStreamUpdates(ctx, t, streamClient, i, &updateCount, errCh)
-		}()
+		g.Go(func() error {
+			return verifyMonotonicStreamUpdates(ctx, t, streamClient, i, &updateCount)
+		})
 	}
 
 	// Wait until all subscribers have verified the first bunch of updates all got
 	// delivered.
-	wg.Wait()
-
-	close(errCh)
-
-	// Require that none of them errored. Since we closed the chan above this loop
-	// should terminate immediately if no errors were buffered.
-	for err := range errCh {
-		require.NoError(t, err)
-	}
+	require.NoError(t, g.Wait())
 
 	// Sanity check that at least some non-snapshot messages were delivered. We
 	// can't know exactly how many because it's timing dependent based on when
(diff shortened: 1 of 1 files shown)


banks:
Took a look locally to try to reproduce. I didn't manage but I think while we are touching this test it's worth making a big optimization. The loop that registers 1000 instances takes about 60 seconds on my laptop right now which makes this test really annoying to run a lot!

I replaces that whole loop locally with this:

```go
// Register a whole bunch of service instances so that the initial snapshot on
	// subscribe is big enough to take a bit of time to load giving more
	// opportunity for missed updates if there is a bug. This uses Txn RPC as it
	// takes ~60 second son my laptop to register 1000 in serial but only about
	// a few ms in one batch. The data is small enough this works.
	batch := make([]*structs.TxnOp, 0, 2000)
	for i := 0; i < 1000; i++ {
		nop := &structs.TxnOp{
			Node: &structs.TxnNodeOp{
				Verb: api.NodeSet,
				Node: structs.Node{
					Node:       fmt.Sprintf("node-redis-%03d", i),
					Address:    "3.4.5.6",
					Datacenter: "dc1",
				},
			},
		}
		batch = append(batch, nop)
		sop := &structs.TxnOp{
			Service: &structs.TxnServiceOp{
				Verb: api.ServiceSet,
				Node: fmt.Sprintf("node-redis-%03d", i),
				Service: structs.NodeService{
					ID:      fmt.Sprintf("redis-%03d", i),
					Service: "redis",
					Port:    11211,
				},
			},
		}
		batch = append(batch, sop)
	}

	r := &structs.TxnRequest{
		Datacenter: "dc1",
		Ops:        batch,
	}
	var txnOut structs.TxnResponse
	require.NoError(t, server.RPC("Txn.Apply", &r, &txnOut))
	require.NoError(t, txnOut.Error())

```

And it went from ~60second to about 200ms. I also added an event count for snapshot events into the verify func and output it in the "snapshot done" log message to check that the registrations are actually happening and creating the number of instances we expect:

```
    subscribe_backend_test.go:477: subscriber 00000: saw canary in snapshot with port 0
    subscribe_backend_test.go:450: subscriber 00000: snapshot done (1001 events), expect next port to be 1
```

That makes whole test take ~7s locally instead of ~67s.

The actual flake is a bug and I have no idea how it's possible but since I can't reproduce locally it's hard to debug further. For now since all our events are idempotent it's not _actually_ affecting the correctness of Consul so if we can't make progress on debugging, we could consider relaxing the test constraint to allow consecutive duplicates. It would still be effective in its original capacity of finding _gaps_ in the delivery like we previously had.

Did you find any way to make it fail this way more frequently @dnephin? With my optimized test I've run it over 100 times locally and only seen 4 failures but they were all to do with the server loosing leadership sometime later in the test run not this issue.

Other than this, I like the changes in the PR here for clarity and better information on failure.

banks:
FWIW, CI runs the test much quicker than on my laptop. I checked my SSD and it's pretty fast - 250k IOPS for random 4k writes, thought I guess fsyncs would slow things down tons.

I realise when I first wrote this test it may have been pre go 1.12 so boltdb was not actually doing a full fsync on macs?

If I put `$TMPDIR` in a RAM disk then it takes ~1.5 seconds with or without the Transaction optimization above but hitting my actually SSD this makes a huge difference.

I'm guessing CI runs TMPDIR on a RAM disk? Or there is some other factor we've not figured out yet - several other Consul devs on recent hardware get the same time I do.

PR #10673, opened 2021-07-21
Update ingress-gateways.mdx

srikrishmurthy:
Updated a note in the ingress-gateway YAML that the ingress gateway name must match with what's configured in the helm chart

### DIFF
--- a/website/content/docs/k8s/connect/ingress-gateways.mdx
+++ b/website/content/docs/k8s/connect/ingress-gateways.mdx
@@ -78,6 +78,7 @@ spec:
       services:
         - name: static-server
 ```
+~> **Note:** The value in the 'name' field for ingressgateway must match the name used in the helm chart. For example, in the above YAML, the name ingress-gateway" is the same name used in the helm chart YAML configuration.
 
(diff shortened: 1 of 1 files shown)


website/content/docs/k8s/connect/ingress-gateways.mdx
```diff
@@ -78,6 +78,7 @@ spec:
       services:
         - name: static-server
 ```
+~> **Note:** The value in the 'name' field for ingressgateway must match the name used in the helm chart. For example, in the above YAML, the name ingress-gateway" is the same name used in the helm chart YAML configuration.
```
blake, line 81:
```suggestion
~> **Note:** The 'name' field for the IngressGateway resource must match the name
specified when creating the gateway in the Helm chart. In the above example, the
name "ingress-gateway" is the [default name](https://www.consul.io/docs/k8s/helm#v-ingressgateways-gateways-name)
used by the Helm chart when enabling ingress gateways.
```

blake (commented):
I had just one suggestion, and then I think this is good to merge.

PR #10689, opened 2021-07-23
Log the correlation ID when blocking queries fire

freddygv:
Knowing that blocking queries are firing does not provide much
information on its own. If we know the correlation IDs we can
piece together which parts of the snapshot have been populated.

Some of these responses might be empty from the blocking
query timing out. But if they're returning quickly I think we
can reasonably assume they contain data.

### DIFF
--- a/.changelog/10689.txt
+++ b/.changelog/10689.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+proxycfg: log correlation IDs for proxy configuration snapshot blocking queries.
+```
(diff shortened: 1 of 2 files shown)


rboyer (approved):
LGTM

freddygv:
Knowing that blocking queries are firing does not provide much
information on its own. If we know the correlation IDs we can
piece together which parts of the snapshot have been populated.

Some of these responses might be empty from the blocking
query timing out. But if they're returning quickly I think we
can reasonably assume they contain data.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/415616.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/415632.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 19f6e1ca31fba1a59c57e6ce621c3b3f86ebf77a onto `release/1.10.x` succeeded!

Issue #10680, opened 2021-07-22
Service should be in Warning state before breaching failures_before_critical


jakubgs:
#### Feature Description

Currently if I set `failures_before_critical` to a value higher than `1` the service appears to be healthy even if checks are failing. We can see this in the implementation here:

https://github.com/hashicorp/consul/blob/c6859b3fb07031e3ea1fcd0230bacb070315d93d/agent/checks/check.go#L946-L952

What would make more sense would be either to:

* Update the service state as `Warning`, which will show the new check state in UI.
* Provide a `failures_before_warning` config to configure when it is updated to `Warning`.

#### Use Case(s)

Right now if I have a fleet of 3 hosts that I deploy an update to gradually - one host at a time - and that deployment is faulty, and the relevant service has a value of `failures_before_critical` higher than `1`(lets say it's a test fleet and we don't want to alarm too early), the state of the service will change only once the threshold is breached. This is an issue for 3 reasons:

1. It delays indication of the issue in the UI until threshold is breached.
2. Obscures the scale of impact, since the issue will appear only on one host initially.
3. It hides the current state of the checks on other hosts in the Web UI.

For someone debugging the issue live as they watch the deployment happen it might make them think at first that the issue is specific to just one host, rather than the whole fleet.

A more sensible approach would be to at least mark the service with a failing check as `Warning`, which would:

1. Indicate the issue in the services/nodes views in the Web UI.
2. Update the response from the check in the `Health Checks` view.

#### Options

The simplest possible fix is to just update the service state to be `Warning` on first failure if the `failures_before_critical` is not reached. But that would change the default behavior of all checks for all users.

The more advanced and more customizable solution would involve introducing a new setting called `failures_before_warning` which by default would be set to the same value as `failures_before_critical` so as to preserve the old behavior. If need be anyone could simply set it to a lower value than `failures_before_critical` and gain ability to see the changed state of the service earlier.

jakubgs:
@dnephin hey. I think this functionality would be really cool, and I could attempt to implement it, since it doesn't look that difficult to do, but I'd like to get the idea of what the Consul team thinks about it. I don't want to waste time implementing `failures_before_warning` only to find out that that's not the right way to do it.

If I could get some feedback on which option makes more sense to you and how you'd expect this to be done I could attempt it.

jkirschner-hashicorp:
[...]

PR #10701, opened 2021-07-27
agent: add failures_before_warning setting

jakubgs:
The new setting allows users to specify the number of check failures that have to happen before a service status us updated to be `warning`.  This allows for more visibility for detected issues without creating alerts and pinging administrators. Unlike the previous behavior, which caused the service status to not update until it reached the configured `failures_before_critical` setting, now Consul updates the Web UI view with the `warning` state and the output of the service check when `failures_before_warning` is breached.

The default value of `FailuresBeforeWarning` is the same as the value of `FailuresBeforeCritical`, which allows for retaining the previous default behavior of not triggering a warning.

When `FailuresBeforeWarning` is set to a value higher than that of `FailuresBeforeCritical it has no effect as `FailuresBeforeCritical` takes precedence.

Resolves: https://github.com/hashicorp/consul/issues/10680


jakubgs:
I've tested this one one of my hosts and it works:
```
[WARN]  agent: Check failed but has not reached failure threshold: check=service-abc status=critical failure_count=1 failure_threshold=5
[WARN]  agent: Check socket connection failed: check=service-abc error="dial tcp 127.0.0.1:30305: connect: connection refused"
[WARN]  agent: Check is now warning: check=service-abc
[INFO]  agent: Synced check: check=service-abc
```
Though I wonder if I should add a `warning_threshold` to that first log message.

jkirschner-hashicorp:
@jakubgs : good catch on the log message. We'll include our thoughts* on that in the review feedback (generally looks good, but need to spend a bit more time to make sure all the relevant test cases are covered).

*We may only want to modify the log output only if the `failures_before_warning` option is in use. That way, any users with systems that parse the existing message (from `failures_before_critical`) will be unaffected unless they opt into using this new flag.

jakubgs:
Actually, I noticed I had some bugs in the tests I modified, and I'm starting to wonder if the `>=` check makes sense.

The documentation states that:

>In Consul 1.7.0 and later, a check may be configured to become passing/warning/critical __only *after* a specified number of consecutive checks return passing/critical__.

The setting is called `failures_before_critical`, which suggests that it sets the number of failures that can happen __before__ the status of the service will change to `critical`, but the use of `>=` isntead of `>` makes it so that the status is changed one failure earlier than it would make sense.

It also means that the value of `1` has the same effect as the value of `0`. If `failuresCounter` is `0`, and a failure happens, the `failuresCounter` variable is incremented to `1`, and now the `s.failuresCounter >= s.failuresBeforeCritical` check is true. If I'm setting `failures_before_critical` to `1` I'd expect one failure to __not__ cause the service to go into critical state.

The same goes for `failures_before_warning`. It's a bit confusing if you ask me.

jakubgs:
I guess there is no point in changing it now, but it really would make more sense if the check was `>`.

jakubgs:
@jkirschner-hashicorp any news on this? Do I need to fix/change anything?

jkirschner-hashicorp:
@jakubgs : my understanding is that you'll receive review feedback in the next 36 hours. If there are any updates to that, I'll let you know. Thanks for checking in!

website/content/docs/discovery/checks.mdx
```diff
@@ -413,16 +413,20 @@ key in your configuration file.
 }
 ```
 
-## Success/Failures before passing/critical
+## Success/Failures before passing/warning/critical
 
-In Consul 1.7.0 and later, a check may be configured to become passing/critical
+In Consul 1.7.0 and later, a check may be configured to become passing/warning/critical
```
dnephin, line 418: I think we'll need to clarify this a bit, since only passing and critical are available in 1.7.0, and warning will be in 1.11.

For most of the settings we have a list of fields. If we did that here I think that would make it easier to document when the fields were added.

Maybe something like this?

> To prevent flapping health checks, and limit the load they cause on the cluster, a health check may be configured to become passing/warning/critical only after a specified of consecutive checks return passing/critical. The status will not transition states until the configured threshold is reached.

> - `success_before_passing` - ... Added in Consul 1.7.0.
> - `failures_before_warning` - ...  Added in Consul 1.11.0.
> - `failures_before_critical` - ... Added in Consul 1.7.0.

The text for these `...` can probably be copied from the API docs.

The text you added below (about the default value) should probably move into this list of fields section.

website/content/api-docs/agent/check.mdx
```diff
@@ -220,6 +220,10 @@ The table below shows this endpoint's support for
   results required before check status transitions to passing. Available for HTTP,
   TCP, gRPC, Docker & Monitor checks.
 
+- `FailuresBeforeWarning` `(int: 0)` - Specifies the number of consecutive unsuccessful
+  results required before check status transitions to warning. Defaults to the same
+  value as `FailuresBeforeCritical`. Available for HTTP, TCP, gRPC, Docker & Monitor checks.
```
dnephin, line 225: Looks good! We should add the "Added in Consul 1.11" here too. We're missing that for the other fields, but we could add that later .

agent/checks/check.go
```diff
@@ -907,17 +908,19 @@ type StatusHandler struct {
...
 	successBeforePassing   int
 	successCounter         int
+	failuresBeforeWarning  int
 	failuresBeforeCritical int
 	failuresCounter        int
 }
 
 // NewStatusHandler set counters values to threshold in order to immediatly update status after first check.
-func NewStatusHandler(inner CheckNotifier, logger hclog.Logger, successBeforePassing, failuresBeforeCritical int) *StatusHandler {
+func NewStatusHandler(inner CheckNotifier, logger hclog.Logger, successBeforePassing, failuresBeforeWarning, failuresBeforeCritical int) *StatusHandler {
```
dnephin, line 917: Accepting so many ints as function arguments is likely going to be error prone (easy to get them mixed up), but I think that's something we can address later. It should be fine for now.

dnephin (commented):
Thank you for working on this enhancement! This looks great! Mostly just a few comments about the docs and testing.

I think the two test changes you made look good, but while reviewing we thought of a couple other scenarios that we should test:

1. a test case where both warning and critical thresholds are set to the same value (the default) to show that it changes to critical and not warning.
2. what happens if the check goes to warning, then gets some updates to passing (but not enough to change its state), and some updates to critical (but not enough to change its state) ? I guess it would remain in warning, which seems like reasonable behaviour. A test case to show that would be great.

We were also thinking about the scenario where `failures_before_warning` is set to a non-zero value, but `failures_before_critical` is still zero. What should the behaviour be? Maybe that should be an error and refuse the registration?

website/content/docs/discovery/checks.mdx
jakubgs: Yeah, that makes sense. Sure.

website/content/api-docs/agent/check.mdx
jakubgs: Sure.

agent/checks/check.go
jakubgs: Yes. I also thought that this might be a bit much, but I don't know the codebase that well so I didn't want to get started on any refactoring in this change. I guess we could do a data structure to pass those around, but that could be done separately, as it would probably complicate this change considerably.

jakubgs:
>We were also thinking about the scenario where `failures_before_warning` is set to a non-zero value, but `failures_before_critical` is still zero. What should the behaviour be? Maybe that should be an error and refuse the registration? A test case for this would also be good.

Yes, I also considered this. Since `failures_before_warning` higher than `failures_before_critical` makes no logical sense there are essentially two options:

* Ignore the value of `failures_before_warning`, since `failures_before_critical` takes precedence
* Explicitly fail to accept a higher value of `failures_before_warning` when loading configuration

The first option is sane, but the second one is more clear about what the program is doing. I'll look at how this kind of check can be added.

jakubgs:
@dnephin I have addressed your comments. The second test case that you described has one caveat tho. Because of how `failuresCounter` is reset by a single passing check the `warning` status can be updated multiple times without actually changing:
https://github.com/status-im/consul/blob/18bc12546df2bce9d89b8bed9f7afbb25634cdda/agent/checks/check_test.go#L1017-L1033

Now, I'm not sure if this is an issue, since it works as intended, it just can generate a few more updates than expected.

This could be optimized if we compared the current state of the check in `updateCheck` before calling `s.inner.UpdateCheck()`, but not sure if that's a sensible solution.

agent/config/builder.go
```diff
@@ -1410,6 +1410,12 @@ func (b *builder) validate(rt RuntimeConfig) error {
 		if err := s.Validate(); err != nil {
 			return fmt.Errorf("service %q: %s", s.Name, err)
 		}
+		// Check for errors in the check definitions
+		for _, c := range s.Checks {
+			if err := c.Validate(); err != nil {
+				return fmt.Errorf("check %q: %s", c.Name, err)
+			}
+		}
```
dnephin, line 1418: Nice! This looks good for validating the checks in services. There can also be checks specified at the top level of the config (outside of services), and I believe we aren't validating those yet.

Unless this is happening somewhere else that I missed, I think we may need something like this:

```go
	for _, check := range rt.Checks {
		if err := check.CheckType().Validate(); err != nil {
			return fmt.Errorf("check %q: %w", check.Name, err)
		}
	}
```

agent/config/builder.go
```diff
@@ -1580,6 +1586,7 @@ func (b *builder) checkVal(v *CheckDefinition) *structs.CheckDefinition {
 		TTL:                            b.durationVal(fmt.Sprintf("check[%s].ttl", id), v.TTL),
 		SuccessBeforePassing:           intVal(v.SuccessBeforePassing),
 		FailuresBeforeCritical:         intVal(v.FailuresBeforeCritical),
+		FailuresBeforeWarning:          intValWithDefault(v.FailuresBeforeWarning, intVal(v.FailuresBeforeCritical)),
```
dnephin, line 1589: This default value looks correct for checks from the config, but unfortunately we have to handle it in different places for the API.

In `HTTPHandlers.AgentRegisterService` and `HTTPHandlers.AgentRegisterCheck` I think we'll need to have a line that sets those to a default value.

Alternatively, maybe we could set that default later on in `NewStatusHandler` ? I honestly don't have a good sense of the trade-offs for those two options. 

As you noticed, since a value of `1` is essentially the same as `0` (if I understand it correctly), maybe we could treat `0` as unset?

dnephin (commented):
Thank you again for your work on this! As I'm sure you have noticed, health checks is not the easiest area of Consul to change. 

While reviewing this PR I wrote up [this list of the 6 ways in which a check can be registered](https://gist.github.com/dnephin/37b1587478d6e968aa697e66f245a3b5), which helped me sort through all the areas that we needed to change.

I think you found most of them, but I left a couple comments about other areas that may have been missed.

I noticed CI was not running for some reason on the latest commit. I have made a small change to the generated proto (which I think will be required to pass CI checks), and pushed that up. I'm hoping that gets CI running.

I have not had a chance to read over and understand the tests yet, but it sounds like they are probably in good shape, so I'll have a look at those soon.

dnephin:
> I noticed CI was not running for some reason on the latest commit. I have made a small change to the generated proto (which I think will be required to pass CI checks), and pushed that up. I'm hoping that gets CI running.

Oh, apparently I don't have permission to push to that remote. The change I made was:

```diff
commit dbfb5452093c57e7b3aacb34cc9f28142f9ac4ac
Author: Daniel Nephin <dnephin@hashicorp.com>
Date:   Wed Aug 11 13:41:01 2021 -0400

    proto: restore previous import
    
    Not sure how this changed, but this diff is the result of me running
    'make protos', and I suspect CI will require this.

diff --git a/proto/pbservice/healthcheck.pb.go b/proto/pbservice/healthcheck.pb.go
index 673d5cd5e..b803ccf04 100644
--- a/proto/pbservice/healthcheck.pb.go
+++ b/proto/pbservice/healthcheck.pb.go
@@ -6,11 +6,11 @@ package pbservice
 import (
 	fmt "fmt"
 	_ "github.com/gogo/protobuf/gogoproto"
+	_ "github.com/gogo/protobuf/types"
 	github_com_gogo_protobuf_types "github.com/gogo/protobuf/types"
 	proto "github.com/golang/protobuf/proto"
 	pbcommon "github.com/hashicorp/consul/proto/pbcommon"
 	github_com_hashicorp_consul_types "github.com/hashicorp/consul/types"
-	_ "google/protobuf"
 	io "io"
 	math "math"
 	math_bits "math/bits"
```

Hopefully after you push another commit CI will start running again.

agent/config/builder.go
jakubgs: I don't know if I like using things like special/magic values, especially if they look "correct", like `0`.

I think it's better just to handle individual cases wherever they appear, to be more explicit closes to where they are read.

agent/config/builder.go
jakubgs: And indeed there are:
https://github.com/hashicorp/consul/blob/242b3a2dc574556e5dc34be1f679ac5f52463bf8/agent/config/runtime.go#L494
And I see you also added a call to `CheckType()`:
https://github.com/hashicorp/consul/blob/242b3a2dc574556e5dc34be1f679ac5f52463bf8/agent/structs/check_definition.go#L170-L173
I see the intention is to convert `CheckDefinition` into `CheckType`. Good call.

agent/config/builder.go
jakubgs: I can see that we are already validating checks inside of `HTTPHandlers.AgentRegisterService`:
https://github.com/hashicorp/consul/blob/710bd90ef76b1d1cf9178f58a7db7b42808021ae/agent/agent_endpoint.go#L990-L996
And we do it inside of `HTTPHandlers.AgentRegisterCheck` as well:
https://github.com/hashicorp/consul/blob/710bd90ef76b1d1cf9178f58a7db7b42808021ae/agent/agent_endpoint.go#L621-L628
So at least validation is covered since `CheckTypes()` does call `Validate()`:
https://github.com/hashicorp/consul/blob/710bd90ef76b1d1cf9178f58a7db7b42808021ae/agent/structs/service_definition.go#L127-L142

jakubgs: Truth be told I'm testing setting just `FailuresBeforeCritical` to a value of `6` and it doesn't seem to work at all.
My test service goes into `critical` state essentially right after creation:
```
2021-08-12T10:42:12.185Z [INFO]  agent: Synced check: check=conusl-check-api-call-test
2021-08-12T10:42:23.940Z [WARN]  agent: Check socket connection failed: check=conusl-check-api-call-test error="dial tcp 127.0.0.1:1234: connect: connection refused"
2021-08-12T10:42:23.940Z [WARN]  agent: Check is now critical: check=conusl-check-api-call-test
```
I'm not entirely sure this functionality works at all... it seems like `FailuresBeforeCritical` is being ignored.

jakubgs: It doesn't seem to return `FailuresBeforeCritical` from the `GET` query at all:
```
 > curl -s http://localhost:8500/v1/agent/checks | jq '."conusl-check-api-call-test"' 
{
  "Node": "node-01.do-ams3.eth.test",
  "CheckID": "conusl-check-api-call-test",
  "Name": "conusl-check-api-call-test",
  "Status": "critical",
  "Notes": "",
  "Output": "dial tcp 127.0.0.1:1234: connect: connection refused",
  "ServiceID": "",
  "ServiceName": "",
  "ServiceTags": [],
  "Type": "tcp",
  "Interval": "30s",
  "Timeout": "2s",
  "ExposedPort": 0,
  "Definition": {},
  "CreateIndex": 0,
  "ModifyIndex": 0
}
```

jakubgs: I've added some debugging prints inside of `StatusHandler.updateCheck` in `agent/checks/check.go`:
```go
		fmt.Printf("checkID.String(): %s\n", checkID.String())
		fmt.Printf("s.failuresCounter: %d\n", s.failuresCounter)
		fmt.Printf("s.failuresBeforeWarning: %d\n", s.failuresBeforeWarning)
		fmt.Printf("s.failuresBeforeCritical: %d\n", s.failuresBeforeCritical)
```
And this is what I got in logs:
```
2021-08-12T11:07:50.077Z [WARN]  agent: Check socket connection failed: check=conusl-check-api-call-test error="dial tcp 127.0.0.1:1234: connect: connection refused"
checkID.String(): conusl-check-api-call-test
s.failuresCounter: 6
s.failuresBeforeWarning: 0
s.failuresBeforeCritical: 6
2021-08-12T11:07:50.077Z [WARN]  agent: Check is now critical: check=conusl-check-api-call-test
```
It appears like `failuresCounter` starts off with value of `6`, which causes it to go into `critical` right away.
https://github.com/hashicorp/consul/blob/710bd90ef76b1d1cf9178f58a7db7b42808021ae/agent/checks/check.go#L921-L922
Not sure if this is the intended behavior but okay.

jakubgs: Okay, I see the comment, makes sense:
```go
// NewStatusHandler set counters values to threshold in order to immediatly update status after first check.
```

jakubgs: Because `AgentRegisterCheck` loads things into `structs.CheckDefinition` then we can't do what the config does:
https://github.com/hashicorp/consul/blob/8f5e2a440b3a858b224a519ba3b2860386f76f14/agent/structs/check_definition.go#L43-L44
Since the values are of type `int` and not `*int` we won't get `nil` for undefined values. So the only solution is indeed to threat the value of `0` as "undefined", and adjust the default based on that.

jakubgs: Also, I noticed that the `OutputMaxSize` attribute has a default:
https://github.com/hashicorp/consul/blob/8f5e2a440b3a858b224a519ba3b2860386f76f14/agent/config/builder.go#L1585
Which is described in the API docs:
image
But I do not see it being changed inside of `agent/agent_endpoint.go`. I will follow that trail then, must happen somewhere.

dnephin: It also quite possible that we don't properly default that field. It is known that because there are multiple paths we have a few bugs where different paths behave differently.

jakubgs: I see, this is being adjusted inside of `Agent.addCheck`:
https://github.com/hashicorp/consul/blob/525bc2d76366dc52fa464f72a406109447c2ccce/agent/agent.go#L2436
https://github.com/hashicorp/consul/blob/525bc2d76366dc52fa464f72a406109447c2ccce/agent/agent.go#L2465-L2471
So that's probably the right place to add this default.

jakubgs: There we go:
https://github.com/status-im/consul/blob/1df254c928e7a0135089ecfabb460fa3a110effe/agent/agent.go#L2468-L2471

jakubgs:
Thanks @dnephin for your detailed comments. I believe I've addressed all of them.

jakubgs:
Any news?

dnephin (commented):
Thanks, this looks great! I've got one request for a small changes to the test case inline.

Also, we recently update `gogo/protobuf`, which I think will changes the code generated in `proto/pbservice/healthcheck.pb.go`.  If we merge this without the updated protobuf it will start to fail on `main`.

If you rebase this branch onto the `main` branch, and `go install github.com/gogo/protobuf/protoc-gen-gofast@v1.3.2`, then re-run `make proto -B` that should update that one `healthcheck.pb.go` file for the new version.

My only remaining concern is that CI is still not running on this branch. By chance do you have CircleCI enabled on `status-im/consul`? I've noticed that if you enable CI on the fork, it won't run CI on the PR, and we can't see the results. I don't see it here (https://github.com/status-im/consul/tree/failures-before-warning) so maybe that's not the problem.

If we can't get CI running I will push these commits to a branch on the main remote and check CI that way.

agent/checks/check_test.go
```diff
@@ -920,19 +937,115 @@ func TestStatusHandlerResetCountersOnNonIdenticalsConsecutiveChecks(t *testing.T
...
+	statusHandler.updateCheck(cid, api.HealthCritical, "bar")
+	statusHandler.updateCheck(cid, api.HealthPassing, "bar")
+	statusHandler.updateCheck(cid, api.HealthCritical, "bar")
+	statusHandler.updateCheck(cid, api.HealthCritical, "bar")
+	statusHandler.updateCheck(cid, api.HealthCritical, "bar")
+	statusHandler.updateCheck(cid, api.HealthPassing, "bar")
+
+	// Two updates happen because the failuresCounter is reset
+	retry.Run(t, func(r *retry.R) {
+		require.Equal(r, 4, notif.Updates(cid))
```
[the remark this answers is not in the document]
jakubgs: Good point. I'll add the additional checks for the state change.

jakubgs:
I have tried rebasing, but it brought no new changes:
```
 > g rebase main
Successfully rebased and updated refs/heads/failures-before-warning.

 > go get github.com/gogo/protobuf/protoc-gen-gofast@v1.3.2       
go: found github.com/gogo/protobuf/protoc-gen-gofast in github.com/gogo/protobuf v1.3.2

 > make proto -B                 
Generating proto/pbconfig/config.proto into proto/pbconfig/config.pb.go and proto/pbconfig/config.pb.binary.go
Generating proto/pbconnect/connect.proto into proto/pbconnect/connect.pb.go and proto/pbconnect/connect.pb.binary.go
Generating proto/pbcommon/common.proto into proto/pbcommon/common.pb.go and proto/pbcommon/common.pb.binary.go
Generating proto/pbsubscribe/subscribe.proto into proto/pbsubscribe/subscribe.pb.go and proto/pbsubscribe/subscribe.pb.binary.go
Generating proto/pbautoconf/auto_config.proto into proto/pbautoconf/auto_config.pb.go and proto/pbautoconf/auto_config.pb.binary.go
Generating proto/pbservice/node.proto into proto/pbservice/node.pb.go and proto/pbservice/node.pb.binary.go
Generating proto/pbservice/healthcheck.proto into proto/pbservice/healthcheck.pb.go and proto/pbservice/healthcheck.pb.binary.go
Generating proto/pbservice/service.proto into proto/pbservice/service.pb.go and proto/pbservice/service.pb.binary.go
Generating proto/pbacl/acl.proto into proto/pbacl/acl.pb.go and proto/pbacl/acl.pb.binary.go
Generating agent/grpc/internal/testservice/simple.proto into agent/grpc/internal/testservice/simple.pb.go and agent/grpc/internal/testservice/simple.pb.binary.go
Generated all protobuf Go files

 > g status
On branch failures-before-warning
nothing to commit, working tree clean
```
But running `proto-gen.sh` directly breaks imports:
```
 > ./build-support/scripts/proto-gen.sh proto/pbservice/healthcheck.proto
Generating proto/pbservice/healthcheck.proto into proto/pbservice/healthcheck.pb.go and proto/pbservice/healthcheck.pb.binary.go

 > g diff
```
```diff
diff --git a/proto/pbservice/healthcheck.pb.go b/proto/pbservice/healthcheck.pb.go
index f1e53fb4f..3cda608ec 100644
--- a/proto/pbservice/healthcheck.pb.go
+++ b/proto/pbservice/healthcheck.pb.go
@@ -6,11 +6,11 @@ package pbservice
 import (
        fmt "fmt"
        _ "github.com/gogo/protobuf/gogoproto"
-       _ "github.com/gogo/protobuf/types"
        github_com_gogo_protobuf_types "github.com/gogo/protobuf/types"
        proto "github.com/golang/protobuf/proto"
        pbcommon "github.com/hashicorp/consul/proto/pbcommon"
        github_com_hashicorp_consul_types "github.com/hashicorp/consul/types"
+       _ "google/protobuf"
        io "io"
        math "math"
        math_bits "math/bits"
```
Because it causes:
```
 > make dev
==> Building Consul - OSes: linux, Architectures: amd64
Building sequentially with go install
--->   linux/amd64
proto/pbservice/healthcheck.pb.go:13:2: package google/protobuf is not in GOROOT (/nix/store/dqqw8sc1darh0ka6d9ii42afryxw50fw-go-1.16.7/share/go/src/google/protobuf)
ERROR: Failed to build Consul for linux/amd64
make: *** [GNUmakefile:149: dev-build] Error 1
```

>By chance do you have CircleCI enabled on status-im/consul?

I don't think we use CircleCI at all, I don't have an account. So I doubt it.

dnephin:
Thanks for making those test changes!

I'm not sure what's going on with CI. I'll take care of re-generating the protobuf and running CI by pushing to another branch. I might not get a chance to do this for a but, but I think this is ready to merge once I've done that.

jakubgs:
I have logged into CircleCI with GitHub just to verify we weren't running these CI jobs, and we don't:

image

jakubgs:
Any news?

jkirschner-hashicorp:
Hi @jakubgs ,

@dnephin and I were both out last week. We'll reconnect and get back to you!

dnephin:
I opened #10969 to get CI running.  The protobuf looks correct now, I think you must have fixed it and maybe I was looking at an old diff.

I fixed one of the test failures, but it looks like the addition of `Validate` caused a bunch more tests to fail. I'm not sure if this is a problem with the tests, or a problem with the new call to `Validate`. I'll need to dig more, but at least now we have the tests running.

jakubgs:
Any news?

jkirschner-hashicorp:
@jakubgs : @dnephin took your commits, added a few more related to protobuf conversion and additional tests, then created PR #10969 from it. PR #10969 now needs a review by someone other than @dnephin (can't review his own code). My understanding is that @dnephin was happy with your commits as they are, and that we just need someone to review/approve @dnephin 's changes - I'll check in with the team about doing that soon. Thanks for checking in!

kisunji:
@jakubgs I've approved and merged the related PR to main. Thank you for your contribution!

jakubgs:
Thanks!

PR #10707, opened 2021-07-27
streaming: set default query timeout

dnephin:
This PR fixes two problems:

1. the HTTP API would log cancelled requests as an error
2. the health endpoint, with the streaming backend, did not set the default timeout that was used with the blocking query backend

### DIFF
--- a/agent/config/runtime.go
+++ b/agent/config/runtime.go
@@ -1930,3 +1930,16 @@ func isFloat(t reflect.Type) bool { return t.Kind() == reflect.Float32 || t.Kind
 func isComplex(t reflect.Type) bool {
 	return t.Kind() == reflect.Complex64 || t.Kind() == reflect.Complex128
 }
+
+// ApplyDefaultQueryOptions returns a function which will set default values on
+// the options based on the configuration. The RuntimeConfig must not be nil.
+func ApplyDefaultQueryOptions(config *RuntimeConfig) func(options *structs.QueryOptions) {
+	return func(options *structs.QueryOptions) {
+		switch {
+		case options.MaxQueryTime > config.MaxQueryTime:
+			options.MaxQueryTime = config.MaxQueryTime
+		case options.MaxQueryTime == 0:
+			options.MaxQueryTime = config.DefaultQueryTime
+		}
+	}
+}
--- a/agent/http.go
+++ b/agent/http.go
@@ -432,12 +432,20 @@ func (s *HTTPHandlers) wrap(handler endpoint, methods []string) http.HandlerFunc
 		}
 
 		handleErr := func(err error) {
-			httpLogger.Error("Request error",
-				"method", req.Method,
-				"url", logURL,
-				"from", req.RemoteAddr,
-				"error", err,
-			)
+			if req.Context().Err() != nil {
+				httpLogger.Info("Request cancelled",
+					"method", req.Method,
+					"url", logURL,
+					"from", req.RemoteAddr,
+					"error", err)
+			} else {
+				httpLogger.Error("Request error",
+					"method", req.Method,
+					"url", logURL,
+					"from", req.RemoteAddr,
+					"error", err)
+			}
+
 			switch {
 			case isForbidden(err):
 				resp.WriteHeader(http.StatusForbidden)
--- a/.changelog/10707.txt
+++ b/.changelog/10707.txt
@@ -1,0 +1,6 @@
+```release-note:bug
+streaming: set the default wait timeout for health queries
+```
+```release-note:bug
+http: log cancelled requests as such at the INFO level, instead of logging them as errored requests.
+```
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -395,6 +395,7 @@ func New(bd BaseDeps) (*Agent, error) {
 			Logger: bd.Logger.Named("rpcclient.health"),
 		},
 		UseStreamingBackend: a.config.UseStreamingBackend,
+		QueryOptionDefaults: config.ApplyDefaultQueryOptions(a.config),
 	}
 
 	a.serviceManager = NewServiceManager(&a)
--- a/agent/rpcclient/health/health.go
+++ b/agent/rpcclient/health/health.go
@@ -17,6 +17,7 @@ type Client struct {
 	MaterializerDeps    MaterializerDeps
 	CacheName           string
 	UseStreamingBackend bool
+	QueryOptionDefaults func(options *structs.QueryOptions)
 }
 
 type NetRPC interface {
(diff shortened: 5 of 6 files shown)


agent/config/runtime.go
```diff
@@ -1930,3 +1930,16 @@ func isFloat(t reflect.Type) bool { return t.Kind() == reflect.Float32 || t.Kind
...
 	return t.Kind() == reflect.Complex64 || t.Kind() == reflect.Complex128
 }
+
+// ApplyDefaultQueryOptions returns a function which will set default values on
+// the options based on the configuration.
+func ApplyDefaultQueryOptions(config *RuntimeConfig) func(options *structs.QueryOptions) {
+	return func(options *structs.QueryOptions) {
+		switch {
+		case options.MaxQueryTime > config.MaxQueryTime:
+			options.MaxQueryTime = config.MaxQueryTime
```
kisunji, line 1940: For my own understanding, the point of this is to only allow apply MaxQueryTime if shorter than existing?

agent/config/runtime.go
```diff
@@ -1930,3 +1930,16 @@ func isFloat(t reflect.Type) bool { return t.Kind() == reflect.Float32 || t.Kind
 func isComplex(t reflect.Type) bool {
 	return t.Kind() == reflect.Complex64 || t.Kind() == reflect.Complex128
 }
+
+// ApplyDefaultQueryOptions returns a function which will set default values on
+// the options based on the configuration.
+func ApplyDefaultQueryOptions(config *RuntimeConfig) func(options *structs.QueryOptions) {
```
kisunji, line 1936: Is there a need to keep this a pointer? Otherwise we may need a nil-check on config

banks (commented):
Thanks Daniel. I think this looks like a good pattern for applying those defaults.

I had one comment inline which I _think_ should probably be changed to avoid creating a new confusing log line if ever we have context timeout errors which are not related to clients at all. What do you think?

agent/http.go
```diff
@@ -432,12 +432,20 @@ func (s *HTTPHandlers) wrap(handler endpoint, methods []string) http.HandlerFunc
 		}
 
 		handleErr := func(err error) {
-			httpLogger.Error("Request error",
-				"method", req.Method,
-				"url", logURL,
-				"from", req.RemoteAddr,
-				"error", err,
-			)
+			if req.Context().Err() != nil {
```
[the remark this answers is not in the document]
banks: Hmm, actually not sure about this. I wasn't thinking about the fact that this is the `req.Context` directly rarthe than a test against any `err` that might be returned by the handler.

In that case I don't know of any way the request context would error with `DeadlineExceeded` so this is probably equivalent, but what do you think about being explicit with the test given that this warning would be inappropriate if there ever was a way for http requests to inherently timeout?

banks (approved):
I realised I was not reading it right. This LGTM, see if you feel like the possible tweak to make this potentially more future proof but not a blocker.

agent/http.go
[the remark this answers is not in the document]
dnephin: Because of the context hierarchy, I think the only way for `req.Context().Err()` to be not nil is if either:
* the request itself was cancelled by the user making the request
* if [http.Server.BaseContext](https://pkg.go.dev/net/http#Server) is set, and something cancels that context

I don't think we use `BaseContext` yet, but even if we did I think this message would be appropriate. Generally the only reason for the BaseContext to be cancelled would be if the server was shutting down, in which case logging the request as cancelled would also be accurate.

agent/config/runtime.go
dnephin: I don't know if I fully understand the question, but I think so yes. 

`config.MaxQueryTime` could also be called `MaxMaxQueryTime`. It is maximum value we allow for the `QueryOptions.MaxQueryTime`. So this code limits the value that users can specify as API query parameters to the value from the config file. If the API query parameter value is greater than the configured maximum, we override the value and set it to the configured maximum.

agent/config/runtime.go
dnephin: Good question! I often try not to use pointers and instead use values when possible. I think it would be fine to use a value here, but I opted for a pointer because `RuntimeConfig` is a very large struct, and generally (because it is so large) we have been passing it around by pointer. I think we expect to only call this method on startup, so using a non-pointer might also be ok.

Regarding the nil-check, if this value were to come from a user then I think that would be very important. But the user can't directly set `RuntimeConfig`. It's expected the agent startup will always create one of these, and there should be no way that it could be nil.

So the only way for `RuntimeConfig` to be `nil` would be a "programming error". A panic is an appropriate way of handling programming errors because we expect our test suite to catch them. If we were to check for the nil, we'd either have to change the signature and return an error from this function, or we'd create a custom panic with more detail.

I suspect a detailed panic wouldn't really help here, since a nil pointer should hopefully be easy to debug. Changing the function signature to return an error would imply a runtime error, but there is no possible runtime error here (only programming errors). I generally try to avoid adding error returns unless there is a possible runtime error because it can force all the callers to also handle the error (which can be a distraction if the only cause is a possible programming error).

I think passing a non-pointer value could in some cases be worse. If we were to pass a zero value for `RuntimeConfig` it would force `MaxQueryTime` to be 0 always. Hopefully that kind of error would be noticed by tests, but a panic should be a lot more obvious.

In short, I think that's an excellent call out (checking for nil), but in this case I'm tempted to say the current implementation is reasonable given that we always expect a non-nil `RuntimeConfig`.

That said, it would be good to document these assumptions in the function docstring, so I will add a line about "RuntimeConfig must not be nil".

agent/http.go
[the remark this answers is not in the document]
dnephin: Thinking about it more, I guess because `BaseContext` is a function, that could act as the middleware, and it could set a deadline, so I think you're right it is possible. 

I guess in most cases the `err` here would include the context about it being a timeout or a cancel, so maybe this log line will be ok in that case as well? I think it's a bit hard to say until we start using `BaseContext`.

agent/config/runtime.go
kisunji: Thanks for the detailed answer! I definitely got more than I asked for :D 
Your point about programming error vs runtime is really interesting - I've never thought of it that way and have seen my fair share of "this change touches a billion files because something down the stack now returns an error"

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/418180.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit d2b58cd0d6e84f04eed08d233b8e8b396f6d7663 onto `release/1.10.x` succeeded!

PR #10709, opened 2021-07-28
Generalize base64 command usage in docs

jkirschner-hashicorp:
The base64 CLI utility has two different short flag arguments for decode depending on the platform: [-D](https://www.unix.com/man-page/osx/1/base64/) and [-d](https://linux.die.net/man/1/base64).

Previously, the docs used the -D flag exclusively with the base64 utility. Luckily, the long form of the flag is the same across platforms: --decode.

All uses of the base64 -D flag have been replaced with --decode. Said uses were identified with grep:
image


blake (approved):
LGTM!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/417579.

PR #10716, opened 2021-07-28
Remove issue-labeler github workflow

dnephin:
It has been failing every time, ex: https://github.com/hashicorp/consul/runs/3186335298?check_suite_focus=true

I guess the problem is that the config uses `@v2` and maybe it needs to be `@v2.0`. It has been failing since it was added in 9fd352a821377f42ab5650ea0b2ed0bc89fa436d.

From the deleted config, we can see that all this does is set the `crash` label when it sees `panic:`. I think we can probably do without that.


freddygv (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/418071.

PR #10735, opened 2021-07-30
ui: Add copy button for Secret ID in Tokens list page

kaxcode:
### ✨  Description:
Adds a copy button to each composite row in tokens list page, if Secret ID returns an actual ID. 
Reuses the [CopyButton](https://github.com/hashicorp/consul/blob/main/ui/packages/consul-ui/app/components/copy-button/README.mdx) component and some conditional logic to flag Secret ID returning an empty string or `'<hidden>'`.
Adds secretID as a common consul translation.

Note: The styling changes did not impact the other composite rows with actions.

[Demo Link](https://consul-ui-staging-jbjn83adm-hashicorp.vercel.app/ui/dc1/acls/tokens#CONSUL_ACLS_ENABLE=1
)

### 📸  Screenshots:
Screen Shot 2021-07-30 at 10 35 35 AM

### ⚡  Backend Changes:
Implemented in #10546.
In the backend, user permission are checked. 
If the user has the correct permission then the SecretID attribute returns an actual ID. 
If the user does not have the correct permissions `'<hiddlen>'` is returned.
It is rare but possible for SecretID to be an empty string. The endpoint would still return the attribute.

### 🤡 Updates to mock-api:
Updates made to /acl/tokens mock data.
Added SecretID attribute with 3 possible responses.
Updated the count of tokens and policies returned because it was making the composite row look very busy. 

Busy composite row:
Screen Shot 2021-07-30 at 10 34 56 AM

### 🧪 Testing: 
No added tests. Simple conditional logic used and the CopyButton component has testing.

### 📌  Before merging: 
- [x] Changelog
- [x] Demo Link

### DIFF
--- a/ui/packages/consul-ui/app/components/consul/token/list/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/token/list/index.hbs
@@ -30,7 +30,7 @@ as |item|>
         </dl>
     </BlockSlot>
     <BlockSlot @name="actions" as |Actions|>
-    {{#if (and (not-eq item.SecretID '') (not-eq item.SecretID '<hidden>'))}}
+    {{#if item.hasSecretID}}
       <CopyButton
         @value={{item.SecretID}}
         @name={{t "common.consul.secretID"}}
--- a/ui/packages/consul-ui/app/components/consul/token/list/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/token/list/index.hbs
@@ -33,9 +33,9 @@ as |item|>
     {{#if item.hasSecretID}}
       <CopyButton
         @value={{item.SecretID}}
-        @name={{t "common.consul.secretID"}}
+        @name={{t "components.consul.token.secretID"}}
       >
-        {{t "common.consul.secretID"}}
+        {{t "components.consul.token.secretID"}}
       </CopyButton>
     {{/if}}
       <Actions as |Action|>
--- a/ui/packages/consul-ui/app/components/consul/token/list/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/token/list/index.hbs
@@ -30,7 +30,14 @@ as |item|>
         </dl>
     </BlockSlot>
     <BlockSlot @name="actions" as |Actions|>
-
+    {{#if (and (not-eq item.SecretID '') (not-eq item.SecretID '<hidden>'))}}
+      <CopyButton
+        @value={{item.SecretID}}
+        @name={{t "common.consul.secretID"}}
+      >
+        {{t "common.consul.secretID"}}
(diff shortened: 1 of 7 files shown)


ui/packages/consul-ui/app/components/consul/token/list/index.hbs
```diff
@@ -30,7 +30,14 @@ as |item|>
         </dl>
     </BlockSlot>
     <BlockSlot @name="actions" as |Actions|>
-
+    {{#if (and (not-eq item.SecretID '') (not-eq item.SecretID '<hidden>'))}}
```
gsusmi, line 33: Granted that I'm not familiar with consul ui code, is it possible to add a `item.hasSecret` or a similar conditional method to do such string matching?

kaxcode: yea, I could do a computed func in the token model to check for a valid Secret ID.

ui/packages/consul-ui/translations/common/en-us.yaml
```diff
@@ -38,6 +38,7 @@ consul:
   destinationname: Destination Name
   sourcename: Source Name
   displayname: Display Name
+  secretID: Secret ID
```
gsusmi, line 41: Could [components/consul/en-us.yaml](https://github.com/hashicorp/consul/blob/e93e5ce149b50a336f22e592e7cb5d3fc8cca12d/ui/packages/consul-ui/translations/components/consul/en-us.yaml#L47) be a possible alternative for this to tag usage for a token copyable text? Something like: `{{t "components.consul.token.secretID"}}` or `{{t "components.consul.token.properties.secretID"}}`

kaxcode: 🤔 yes, it would be better under tokens. I thought SecretID was used in other ACL models, but looks like it's just in tokens. I'll move that over.

gsusmi (approved):
LGTM. Ty!

randallmorey (changes requested):
Recommend either changing the text, adding a title attribute, and/or adding an aria attribute that includes the _verb_ for the action.  This would better communicate to users what happens when they activate the button.  For example, "Copy Secret ID".

kaxcode:
> Recommend either changing the text, adding a title attribute, and/or adding an aria attribute that includes the verb for the action. This would better communicate to users what happens when they activate the button. For example, "Copy Secret ID".

@randallmorey PR #10664 addresses this.

randallmorey (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/418982.

PR #10736, opened 2021-07-30
config: add secure-defaults setting

dnephin:
This is a quick spike to demonstrate the idea of a `-secure-defaults` setting.

The setting would force enable a bunch of settings required for a secure setup (ex: ACLs enabled), and would make others required (ex: an agent TLS cert and key).


jmurret:

PR #10737, opened 2021-07-30
acl: remove authz == nil checks

dnephin:
Follow up to #10632

Remove more of the `authz == nil` checks now that resolving a token should never return a `nil`. There are a few more `auth != nil` checks to handle, but I'll do that in another PR to keep the diff small.

Also cleans up some agent methods that were only used by tests.

### DIFF
--- a/agent/connect_auth.go
+++ b/agent/connect_auth.go
@@ -132,14 +132,6 @@ func (a *Agent) ConnectAuthorize(token string,
 		return false, reason, &meta, nil
 	}
 
-	// No match, we need to determine the default behavior. We do this by
-	// fetching the default intention behavior from the resolved authorizer. The
-	// default behavior if ACLs are disabled is to allow connections to mimic the
-	// behavior of Consul itself: everything is allowed if ACLs are disabled.
-	if authz == nil {
-		// ACLs not enabled at all, the default is allow all.
-		return true, "ACLs disabled, access is allowed by default", &meta, nil
-	}
 	reason = "Default behavior configured by ACLs"
 	return authz.IntentionDefaultAllow(nil) == acl.Allow, reason, &meta, nil
 }
--- a/agent/consul/acl_endpoint_legacy.go
+++ b/agent/consul/acl_endpoint_legacy.go
@@ -169,7 +169,7 @@ func (a *ACL) Apply(args *structs.ACLRequest, reply *string) error {
 	// NOTE: We will not support enterprise authorizer contexts with legacy ACLs
 	if rule, err := a.srv.ResolveToken(args.Token); err != nil {
 		return err
-	} else if rule == nil || rule.ACLWrite(nil) != acl.Allow {
+	} else if rule.ACLWrite(nil) != acl.Allow {
 		return acl.ErrPermissionDenied
 	}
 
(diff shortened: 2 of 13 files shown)


agent/structs/intention.go
```diff
@@ -325,11 +322,12 @@ func (ixn *Intention) CanRead(authz acl.Authorizer) bool {
...
-	if authz == nil || authz == acl.ManageAll() {
+	if authz == acl.ManageAll() {
 		return true
 	}
 	var authzContext acl.AuthorizerContext
 
+	// TODO: this line seems to require checking 'authz == acl.ManageAll()' above
 	if ixn.DestinationName == "" {
 		return false
 	}
```
dnephin, line 333: Calling out this TODO. In all other cases we've found (so far) simply using the `acl.Authorizer` without checking its value is correct.

In this case it seems like it may not be , so we had to add an `authz == acl.ManangeAll()` check above (the `ManageAll` authorized is returned when acls are disabled).

Is this `ixn.DestinationName == ""` in the right place? Maybe it should be in `authz.IntentionWrite` (called below)?   I believe if we made that change we could remove the `authz == acl.ManangeAll()` the `acl.Authorizer` interface would be much safer to use in general.

dnephin: cc @freddygv @rboyer  Thoughts on this proposed change? Any concerns?

agent/consul/acl_endpoint_legacy.go
```diff
@@ -169,7 +169,7 @@ func (a *ACL) Apply(args *structs.ACLRequest, reply *string) error {
 	// NOTE: We will not support enterprise authorizer contexts with legacy ACLs
 	if rule, err := a.srv.ResolveToken(args.Token); err != nil {
```
freddygv, line 170: minor: What if rule here (and other places calling `ResolveToken*`) was renamed to authz and the variable name rule was reserved for binding rules?

freddygv (approved):

agent/connect_auth.go
```diff
@@ -132,14 +132,6 @@ func (a *Agent) ConnectAuthorize(token string,
 		return false, reason, &meta, nil
 	}
 
-	// No match, we need to determine the default behavior. We do this by
-	// fetching the default intention behavior from the resolved authorizer. The
-	// default behavior if ACLs are disabled is to allow connections to mimic the
-	// behavior of Consul itself: everything is allowed if ACLs are disabled.
-	if authz == nil {
```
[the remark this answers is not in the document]
dnephin: I think it's handled here: https://github.com/hashicorp/consul/pull/10738/files#diff-385793ff8e66adc4fd524432b9ba17d84e6bef2655696874e19664f39b3077a9 

But I'll do another pass along with the other suggestion to make sure

agent/consul/acl_endpoint_legacy.go
dnephin: I think that's a great idea, I'll do that as a follow up

agent/structs/intention.go
dnephin: Sounds good. I'll address this in a follow up as well

PR #10738, opened 2021-07-30
acl: remove the last of the authz == nil checks

dnephin:
Branched from #10737, this PR continues the work from #10632

This diff is larger because it removes some checks that were around larger blocks of code. You can use [hide whitespace changes](https://github.com/hashicorp/consul/pull/10738/files?diff=split&w=1) to reduce the diff significantly.

There is one last remaining `authz != nil` check here: https://github.com/hashicorp/consul/blob/v1.10.1/agent/consul/acl.go#L1239-L1241
That one has to remain because it can be `nil` in OSS. Since this is internal to the "ACL system" it seems like a reasonable exception. We could probably change it to use a new `acl.DefaultAll` authorizer, but I'm not sure its worth it for this one small case.

### DIFF
--- a/agent/consul/txn_endpoint.go
+++ b/agent/consul/txn_endpoint.go
@@ -81,9 +81,9 @@ func (t *Txn) preCheck(authorizer acl.Authorizer, ops structs.TxnOps) structs.Tx
 			}
 
 			service := &op.Service.Service
-			// This is intentionally nil as we will authorize the request
-			// using vetServiceTxnOp next instead of doing it in servicePreApply
-			if err := servicePreApply(service, nil); err != nil {
+			// acl.ManageAll is used here because the request will be authorized
+			// later using vetServiceTxnOp.
+			if err := servicePreApply(service, acl.ManageAll()); err != nil {
 				errors = append(errors, &structs.TxnError{
 					OpIndex: i,
 					What:    err.Error(),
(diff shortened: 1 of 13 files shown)


agent/consul/txn_endpoint.go
```diff
@@ -81,9 +81,9 @@ func (t *Txn) preCheck(authorizer acl.Authorizer, ops structs.TxnOps) structs.Tx
 			}
 
 			service := &op.Service.Service
-			// This is intentionally nil as we will authorize the request
-			// using vetServiceTxnOp next instead of doing it in servicePreApply
-			if err := servicePreApply(service, nil); err != nil {
+			// acl.ManageAll is used here because the request will be authorized
+			// later using vetServiceTxnOp.
+			if err := servicePreApply(service, acl.ManageAll()); err != nil {
```
dnephin, line 86: Calling out this change as it's slightly different from the rest.

The goal of this chain of PRs is to remove these nil checks, so we replace the `nil` we pass in here with `ManageAll` to achieve the same goal.

freddygv (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/422115.

PR #10741, opened 2021-07-30
acl: rename tokens to Root and AgentRoot

mikemorris:
Rough first pass at renaming the ACL `Master` and `AgentMaster` tokens to `Root` and `AgentRoot`. All the current changes are just simple renaming and website or docs updates, but I have a few outstanding questions before this is ready to be pulled out of a draft state:

- [ ] Legacy ACLs - it will likely be easier to wait for these to be removed entirely rather than actually renaming and figuring out any compatibility issues
- [ ] Is there a need to support any transitory state, such as upgrading from a 1.10 to 1.11 Consul release, and how might we best handle that?
- [ ] Are there concerns with changing protobuf wire format, e.g. should the existing fields be preserved/deprecated and new fields added instead of renaming?
- [ ] `consul/api` module - figure out how we might want to handle this change in a backwards compatible way


kisunji:
>    Are there concerns with changing protobuf wire format, e.g. should the existing fields be preserved/deprecated and new fields added instead of renaming?

Renames of existing fields are safe to do on pb definitions!

mikemorris:
Outdated, work completed in the following series of PRs:
- https://github.com/hashicorp/consul/pull/11665
- https://github.com/hashicorp/consul/pull/11669
- https://github.com/hashicorp/consul/pull/11744
- https://github.com/hashicorp/consul/pull/11746
- https://github.com/hashicorp/consul/pull/11751
- https://github.com/hashicorp/consul/pull/11764
- https://github.com/hashicorp/consul/pull/11827

PR #10742, opened 2021-07-30
acl: move agent/consul vet functions

dnephin:
While looking at possibly moving `ACLResolver` into a package under `acl` I noticed these "vet" functions in `agent/consul/acl.go`. Most of what's in `agent/consul/acl.go` is the `ACLResolver` and filtering. These vet functions are not really related to the `ACLResolver`, and they are only called from one place (either the catalog RPC endpoints or Txn RPC endpoints).

This PR moves the vet function to where they are used, which seems like a better place for them. It is done in hopes that moving the ACLResolver can be done as mostly a file rename without a large diff. But even if that doesn't happen, this change seems like an improvement. These functions are actually part of the catalog and txn components, not part of the ACL system.

Now that #10632 has merged, this change also removes the `authz == nil` checks in these functions, because those cases are no longer possible.


freddygv (approved):
LGTM, thanks for adding the `rule` var rename

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/422965.

PR #10744, opened 2021-07-31
Document tagged addresses

### LINKED ISSUE #6989: Confused about doc in 'Service Definition' part
I'm reading the doc of consul .The doc gives a [configuration](https://www.consul.io/docs/agent/services.html#service-definition) file for the service, parts of it are as follows:
```
    "tagged_addresses": {
      "lan": {
        "address": "192.168.0.55",
        "port": 8000,
      },
      "wan": {
        "address": "198.18.0.23",
        "port": 80
      }
    },
    "port": 8000
```
So why the ports in tagged_addresses are different, does this mean that the wan address uses a proxy. Or the tagged_addresses configuration item have any other meaning?

blake:
Resolves #6989


freddygv (approved):
Looks accurate from my perspective

website/content/docs/discovery/services.mdx
```diff
@@ -275,6 +275,255 @@ For Consul 0.9.3 and earlier you need to use `enableTagOverride`. Consul 1.0
...
+  }
+}
+```
+
+</CodeBlockConfig>
+</CodeTabs>
+
+#### Virtual
+
+Virtual tagged addresses are a logical address type which is used by
```
trujillo-adam, line 392:
```suggestion
Virtual tagged addresses are logical address types used by
```

website/content/docs/discovery/services.mdx
```diff
@@ -275,6 +275,255 @@ For Consul 0.9.3 and earlier you need to use `enableTagOverride`. Consul 1.0
...
+[Connect](/docs/connect) to route traffic to the associated destination service
+in the service mesh when:
```
trujillo-adam, line 394: Just to be clear, both conditions should be met to route traffic--not either or, correct? If so, consider making it more explicit.

website/content/docs/discovery/services.mdx
```diff
@@ -275,6 +275,255 @@ For Consul 0.9.3 and earlier you need to use `enableTagOverride`. Consul 1.0
...
+   [dialed directly](/docs/connect/config-entries/service-defaults#dialeddirectly).
+
+-> **Note**: Virtual addresses are not required to be routable IPs within the
+network. They are strictly a control plane construct used to provide a single
+address for the instances of a given logical service.
+
+##### Supported tags
+
+The following address tag can be configured for a service to advertise the
+logical address to be used to dial it over the service mesh.
```
trujillo-adam, line 407:
```suggestion
Use the following address tag to specify the logical address the service should advertise when dialing over the service mesh.
```
I am not certain that I correctly restated the part about "dialing over the service mesh".

website/content/docs/discovery/services.mdx
```diff
@@ -275,6 +275,255 @@ For Consul 0.9.3 and earlier you need to use `enableTagOverride`. Consul 1.0
...
+}
+```
+
+</CodeBlockConfig>
+</CodeTabs>
+
+#### WAN
+
+These addresses are intended to be accessible only when dialing a node or service
+from a remote WAN federated data center.
```
trujillo-adam, line 455: It's probably just me, but I'm having trouble understanding the type of data center at line 455. 

Do we mean "remote WAN-federated data center", which would mean that the DC is remote and federated according to WAN addresses?

Do we mean "remote, WAN, federated data center", which means that the federated DC is also remote and identified by WAN address? 

Feel free to ignore this comment if it's just a gap in my understanding.

trujillo-adam (approved):
Looks good. I added a few minor suggestions and a few questions.

website/content/docs/discovery/services.mdx
blake: Yes, both conditions are required to be met. I can add such phrasing to the previous sentence.

website/content/docs/discovery/services.mdx
blake:
```suggestion
Use the following address tag to specify the logical address at which the service can be reached by other services in the mesh.
```
Is this perhaps more clear?

website/content/docs/discovery/services.mdx
blake: Consul has a feature called [translated addresses](https://www.consul.io/api-docs#translated-addresses) which allow the catalog to return different addresses in the discovery response based on whether the requestor is in the same DC, or a different DC that is connected via WAN federation.

The docs for the [`translate_wan_addrs`](https://www.consul.io/docs/agent/options#translate_wan_addrs) config option offers a better explanation on this behavior.

> If set to true, Consul will prefer a node's configured WAN address when servicing DNS and HTTP requests for a node in a remote datacenter. This allows the node to be reached within its own datacenter using its local address, and reached from other datacenters using its WAN address, which is useful in hybrid setups with mixed networks. This is disabled by default.

These `wan` tagged addresses are the same as the `WAN address` mentioned in the previous paragraph.

website/content/docs/discovery/services.mdx
trujillo-adam: Yes, I think I was thrown off by the term 'dial' -- thinking it meant something more technical than literally just a service communicating with another another service

website/content/docs/discovery/services.mdx
trujillo-adam: OK, then I _believe_ the correct statement would be: 

"Define a WAN address when services need to remotely access a federated datacenter at a WAN address." 

Yeah?

trujillo-adam (approved):
Made a minor suggestion based on your explanation

website/content/docs/discovery/services.mdx
blake: We settled on the following phrasing.
> Define a WAN address for the service or node when it should be accessed at an alternate address by services in a remote datacenter.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/427248.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 8a396ae73f56c1a4f48c1b6a272a241e3c7be863 onto `stable-website` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 8a396ae73f56c1a4f48c1b6a272a241e3c7be863 onto `release/1.10.x` succeeded!

PR #10710, opened 2021-07-28
NIA Docs: Add Citrix TF module links

BeenzSyed:
Add Citrix Terraform Registry and GH links

PR #10765, opened 2021-08-04
Add Citrix TF module links to NIA docs (#10710)

BeenzSyed:
Add Citrix TF module links to NIA docs


lornasong (approved):
Thanks for linking the original pr that the cherry-picker wasn't able to port over! Looks good to me!

PR #10770, opened 2021-08-04
telemetry: add log message when certs are about to expire

### LINKED ISSUE #9891: Add telemetry and logging around expired certificates

dnephin:
Branched from #10768
Closes #9891

Adds a warning log when any of the monitored certificates are within 1 day of expiration. Ideally the metrics should be used to monitor the expiration, but these logs will help in case the metrics are not being monitored correctly.

### DIFF
--- a/agent/consul/leader_metrics.go
+++ b/agent/consul/leader_metrics.go
@@ -140,6 +140,12 @@ func (m CertExpirationMonitor) Monitor(ctx context.Context) error {
 				logger.Warn("failed to emit certificate expiry metric", "error", err)
 				continue
 			}
+
+			if d < 24*time.Hour {
+				logger.Warn("certificate will expire soon",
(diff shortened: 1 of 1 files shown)


hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/422828.

Issue #10762, opened 2021-08-03
connect/ca: Vault Provider: Document and make the root CA TTL configuration


dnephin:
Related to #10643

When the Vault root PKI is not already mounted, we mount it and set the `max_lease_ttl` to 1 year. I believe this effectively sets the Root CA expiration to 1 year. But we never document that it is set to this, and we don't allow the user to configure it.

1 year seems like a strange default, because we use 10 years for the Consul provider.

Related: `ttl` for generating the root https://www.vaultproject.io/api/secret/pki#ttl-2, which I believe is capped ta max_lease_ttl according to https://www.vaultproject.io/docs/secrets/pki#configure-a-ca-certificate

This issue is to:
1. document the current behaviour (which will likely remain the default)
2. allow the user to configure an expiration, and document that it is only used if the vault PKI is not already initialized.

dschaaff:
The work around for those who want a longer root CA TTL is to create the Vault PKI backend yourself with lease TTLs to your desired length and then configure Consul to use it.

FFMMM:
With https://github.com/hashicorp/consul/pull/11428 merged and coming in next https://github.com/hashicorp/consul/pull/11449, I'm going to close this issue.

Setting a ttl value for root certs will be in 1.11

Also the docs should be updated as part of https://github.com/hashicorp/consul/pull/11488

PR #10771, opened 2021-08-04
telemetry: improve cert expiry metrics

dnephin:
Related to some improvements suggested here: https://github.com/hashicorp/consul/pull/10768#discussion_r682871038

Emit the metric immediately so that after restarting an agent, the new expiry time will be
emitted. This is particularly important when this metric is being monitored, because we want
the alert to resovle itself immediately.

Fixed a bug that was exposed in one of these metrics. The CARoot can be nil, so we have
to handle that case.

Removed the labels. With the labels these metrics would not match the predefined metric, and the prometheus sink would expire them long before they were emitted again. It should be the responsibility of agent or tool that pulls these metrics to add node and dc tags.

Only declare these metrics on servers, and emit a NaN when a server starts so that it only has a value on leaders. Otherwise attempting to use this metric in a monitor would be challenging because followers would always report 0.

TODO:
* [x] handle "erasing" the metrics from non-leaders so that when a server transitions from leader to follower prometheus sees the correct data.
* [x] test the "emit NaN value" as a fix for above

### DIFF
--- a/agent/consul/leader_metrics.go
+++ b/agent/consul/leader_metrics.go
@@ -5,6 +5,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"math"
 	"strings"
 	"time"
 
@@ -156,6 +157,9 @@ func (m CertExpirationMonitor) Monitor(ctx context.Context) error {
 	for {
 		select {
 		case <-ctx.Done():
+			// "Zero-out" the metric on exit so that when prometheus scrapes this
+			// metric from a non-leader, it does not get a stale value.
+			metrics.SetGauge(m.Key, float32(math.NaN()))
 			return nil
 		case <-ticker.C:
 			fn()
--- a/agent/consul/leader_metrics.go
+++ b/agent/consul/leader_metrics.go
@@ -42,10 +42,7 @@ var AgentCertExpirationGauges = []prometheus.GaugeDefinition{
 
 func rootCAExpiryMonitor(s *Server) CertExpirationMonitor {
 	return CertExpirationMonitor{
-		Key: metricsKeyMeshRootCAExpiry,
-		Labels: []metrics.Label{
-			{Name: "datacenter", Value: s.config.Datacenter},
-		},
+		Key:    metricsKeyMeshRootCAExpiry,
 		Logger: s.logger.Named(logging.Connect),
 		Query: func() (time.Duration, error) {
 			return getRootCAExpiry(s)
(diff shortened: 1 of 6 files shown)


agent/consul/leader_metrics.go
```diff
@@ -130,24 +134,35 @@ func (m CertExpirationMonitor) Monitor(ctx context.Context) error {
 
 	logger := m.Logger.With("metric", strings.Join(m.Key, "."))
 
+	fn := func() {
```
FFMMM, line 137: can't think of a reason not to use a function literal here but a small nit was to see if we could rename the literal to something like `emitCertExpiryFunc` or some other descriptive name. Wdyt?

agent/consul/leader_metrics.go
```diff
@@ -130,24 +134,35 @@ func (m CertExpirationMonitor) Monitor(ctx context.Context) error {
...
+		expiry := d / time.Second
+		metrics.SetGaugeWithLabels(m.Key, float32(expiry), m.Labels)
+	}
+
+	// emit the metric immediately so that if a cert was just updated the
+	// new metric will be updated to the new expiration time.
+	fn()
+
 	for {
 		select {
```
FFMMM, line 158: Also, I may be a bit naive here but `select`s without a `default` block make me a bit uneasy.

Theoretically, they could block forever, right?

> Since communication on nil channels can never proceed, a select with only nil channels and no default case blocks forever.

https://golang.org/ref/spec#Select_statements

Just trying to learn what the rationale might have been behind not including it originally.

dnephin: I believe it is pretty common to have no `default` in `select`, but it is definitely important to have a "shutcase" case. Here the shutdown case is `ctx.Done()`. The context will be cancelled by the caller when it wants to shutdown the goroutine. At that time the goroutine will exit. Using `ctx.Done()` should be a very common pattern in Go for managing the lifecycle of goroutines. It's unfortunately not yet super common in Consul, but we have been using it a lot more over the last year or so.

I believe the key point in those docs is "with only nil channels". Here both cases (`ctx.Done()`, and `ticker.C`) can not be nil, so I don't think we are at a risk of blocking forever.

Adding a `default` case actually changes the behaviour quite a bit. It makes it so that the select doesn't actually wait on the channels, but instead immediately goes to the `default` if there are no values ready in any of the channels. If we were to add a `default` here it would change this logic from a rather efficient select that blocks on operations, to a busy for loop that is constantly consuming CPU.

FFMMM: thanks for the explanation 💯 !

FFMMM (approved):
🚀 LGTM! Just the small nit around naming !

FFMMM:
Also, just adding a blurb on testing.

I know we chatted in the very beginning of my research into https://github.com/hashicorp/consul/pull/11231 about testing the metrics emission.

While still under review, one e2e approach is [here](https://github.com/hashicorp/consul/pull/11231/files#diff-472e516d540a393b43170efc09098871a521b8b8ed4788089412829311ff81c5R12)

A more "elegant" way could be to query the `.MetricsHandler` and look at respective counters/ gauges but I wanted increased confidence in how the prom metrics showed up.

I think going about it the same way would be harder here since root CA max ttl is not configurable for now -- hoping I can work on https://github.com/hashicorp/consul/issues/10762 next which could help.

If this PR does go in without the specific metric emissions tests maybe it'd be worth backlogging the intent as an issue at least?

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/491083.

PR #10792, opened 2021-08-04
acl: use authz consistently as the variable name for an acl.Authorizer

dnephin:
Follow up to https://github.com/hashicorp/consul/pull/10737#discussion_r682147950

Renames all variables for `acl.Authorizer` to use `authz`. Previously some places used `rule` which I believe was an old name carried over from the legacy ACL system.

A couple places also used `authorizer`.

This commit also removes another couple of authorizer nil checks that are no longer necessary.


rboyer (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/428776.

PR #10795, opened 2021-08-05
cli: Test API access using /status/leader in consul watch

### LINKED ISSUE #9353: consul watch within namespace
trying to watch over a KV entry in a namespace, but gets permission denied errors when using the namespace-management token.

Еxpectation is to be able to do use consul watch command in a namespace.

When running a consul watch command in the default namespace, it executes normally:
consul watch -type=keyprefix -prefix=redis/
[
{
"Key": "redis/config/minconns",
"CreateIndex": 9601002,
"ModifyIndex": 9601002,
"LockIndex": 0,
"Flags": 0,
"Value": "NTY0Mw==",
"Session": "",
"Namespace": "default"
}
]
However, when this is run within a namespace (using namespace operator token), a 403 error is returned:
consul watch -type=keyprefix -prefix=redis/
Error querying Consul agent: Unexpected response code: 403 (Permission denied)

Looking at docs and source code, it seems that consul watch command does not support namespace functionality.

https://github.com/hashicorp/consul-enterprise/blob/master/command/watch/watch.go

Filing this bug report as requested from @mikemorris .

blake:
Replace call to /agent/self with /status/leader to verify agent reachability before initializing a watch. This endpoint is not guarded by ACLs, and as such can be queried by any unauthenticated API client.

Fixes #9353

### DIFF
--- a/.changelog/10795.txt
+++ b/.changelog/10795.txt
@@ -1,0 +1,4 @@
+```release-note:bug
+cli: Fix a bug which prevented initializing a watch when using a namespaced
+token.
+```
--- a/command/watch/watch.go
+++ b/command/watch/watch.go
@@ -158,13 +158,19 @@ func (c *cmd) Run(args []string) int {
 		return 1
 	}
 
-	// Create and test the HTTP client
+	// Create and test that the API is accessible before starting a blocking
+	// loop for the watch.
+	//
+	// Consul does not have a /ping endpoint, so the /status/leader endpoint
+	// will be used as a substitute since it does not require an ACL token to
+	// query, and will always return a response to the client, unless there is a
+	// network communication error.
 	client, err := c.http.APIClient()
 	if err != nil {
 		c.UI.Error(fmt.Sprintf("Error connecting to Consul agent: %s", err))
 		return 1
 	}
-	_, err = client.Agent().NodeName()
+	_, err = client.Status().Leader()
 	if err != nil {
 		c.UI.Error(fmt.Sprintf("Error querying Consul agent: %s", err))
 		return 1


dnephin (commented):
Thank you for working on this fix! I think this change would indeed fix the problem, but I wonder why this was done in the first place.

I looked through git history, but it seems like this was added as part of the first commit for this command, so there's no real context.

I have a vague recollection (which might be wrong) of a conversation about this, someone had suggested we did this kind of thing in a watch as a way of catching connection problems before going into the blocking loop.

I wonder if that kind of pre-check before starting a blocking loop would still be valuable. It's unfortunate that we don't have a `/v1/ping` endpoint that we can use to check basic connectivity.

I recently learned that `/v1/status/leader` (`Status().Leader()` from the api client) will never return an error unless there is a connection problem. That API endpoint also requires no authorization.  

Maybe we could replace the `Agent().NodeName()` call with `Status().Leader()` and document why we are using it. Something along the lines of:
* there is no ping endpoint
* we want a ping before going into the blocking loop
* /status/leader can act as a substitute for `/ping` because it requires no ACL permissions

blake:
Thanks for the suggestion, @dnephin. I've changed the code to query the `/status/leader` endpoint prior to attempting to initialize the watch.

dnephin (approved):
LGTM, thank you!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/423988.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 6a68bfc5e133111c054d52cb810bc974886954b8 onto `release/1.10.x` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 6a68bfc5e133111c054d52cb810bc974886954b8 onto `release/1.9.x` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 6a68bfc5e133111c054d52cb810bc974886954b8 onto `release/1.8.x` succeeded!

PR #10799, opened 2021-08-05
changelog: add KNOWN ISSUES note for streaming with mesh gateways

mikemorris:
Open to suggestions on improving this language, but hope this captures the issue well enough.

_meta note:_ should we add a `known-issue` type to https://github.com/hashicorp/consul/blob/main/.changelog/changelog.tmpl? thinking that would mostly make sense if we intend to sometimes ship releases with known issues (such as incomplete/preview features) but not if we're just adding notes like this post-release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 1.10.1 (July 15, 2021)
 
+KNOWN ISSUES:
+
+* The change to enable streaming by default uncovered an incompatibility between streaming and WAN federation over mesh gateways causing traffic to fall back to attempting a direct WAN connection rather than transiting through the gateways. We currently suggest not upgrading beyond 1.10.0 if using WAN federation over mesh gateways, and are working to address this issue in a future patch release.
+
 SECURITY:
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 1.10.1 (July 15, 2021)
 
+KNOWN ISSUES:
+
+* The change to enable streaming by default uncovered an incompatibility between streaming and WAN federation over mesh gateways causing traffic to fall back to attempting a direct WAN connection rather than transiting through the gateways. We currently suggest not upgrading beyond 1.10.0 if using WAN federation over mesh gateways, and are working to address this issue in a future patch release.
```
dnephin, line 7: Maybe we should recommend setting `use_streaming_backend=false` ([ref](https://www.consul.io/docs/agent/options#use_streaming_backend)) instead of staying on 1.10.0 ?  That would allow them to pickup other CVE and bug fixes from 1.10.1.

dnephin (approved):
SGTM! Thanks for adding this.

CHANGELOG.md
mikemorris: Good call, yea that would be preferable.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/423284.

PR #10804, opened 2021-08-06
debug: rename cluster.json -> members.json  and fix handling of Interrupt Signal

dnephin:
Related to #10320

Renames `cluster.json` to `members.json` to be more consistent with the API and internal method names we use for this data.

While working on this I noticed that handling of SIGINT was broken due to other changes we made for #10320. I fixed that problem by using a context, and changing the signal handler to cancel the context instead of only closing a channel.

### DIFF
--- a/api/debug.go
+++ b/api/debug.go
@@ -1,7 +1,9 @@
 package api
 
 import (
+	"context"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"strconv"
 )
@@ -68,6 +70,26 @@ func (d *Debug) Profile(seconds int) ([]byte, error) {
 	}
 
 	return body, nil
+}
+
+// PProf returns a pprof profile for the specified number of seconds. The caller
+// is responsible for closing the returned io.ReadCloser once all bytes are read.
+func (d *Debug) PProf(ctx context.Context, name string, seconds int) (io.ReadCloser, error) {
+	r := d.c.newRequest("GET", "/debug/pprof/"+name)
+	r.ctx = ctx
+
+	// Capture a profile for the specified number of seconds
+	r.params.Set("seconds", strconv.Itoa(seconds))
+
+	_, resp, err := d.c.doRequest(r)
+	if err != nil {
+		return nil, fmt.Errorf("error making request: %s", err)
+	}
+
+	if resp.StatusCode != 200 {
+		return nil, generateUnexpectedResponseCodeError(resp)
+	}
+	return resp.Body, nil
 }
 
 // Trace returns an execution trace
--- a/command/debug/debug.go
+++ b/command/debug/debug.go
@@ -12,8 +12,10 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"os/signal"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"time"
 
 	"golang.org/x/sync/errgroup"
@@ -55,7 +57,7 @@ const (
 	debugProtocolVersion = 1
 )
 
-func New(ui cli.Ui, shutdownCh <-chan struct{}) *cmd {
+func New(ui cli.Ui) *cmd {
 	ui = &cli.PrefixedUi{
 		OutputPrefix: "==> ",
 		InfoPrefix:   "    ",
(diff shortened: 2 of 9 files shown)


command/debug/debug.go
```diff
@@ -450,8 +442,7 @@ func (c *cmd) captureGoRoutines(timestampDir string) error {
 		return fmt.Errorf("failed to collect goroutine profile: %w", err)
 	}
 
-	err = ioutil.WriteFile(fmt.Sprintf("%s/goroutine.prof", timestampDir), gr, 0644)
-	return err
+	return ioutil.WriteFile(fmt.Sprintf("%s/goroutine.prof", timestampDir), gr, 0644)
```
dnephin, line 455: I fix these to use `filepath.Join` in a follow up commit.

api/debug.go
```diff
@@ -70,6 +72,26 @@ func (d *Debug) Profile(seconds int) ([]byte, error) {
...
+
+	_, resp, err := d.c.doRequest(r)
+	if err != nil {
+		return nil, fmt.Errorf("error making request: %s", err)
+	}
+
+	if resp.StatusCode != 200 {
+		return nil, generateUnexpectedResponseCodeError(resp)
+	}
+	return resp.Body, nil
```
kisunji, line 92: What's the use case for returning the `resp.Body` raw vs `ioutil.ReadAll`-ing and returning `[]byte`?

dnephin: Good question! That is indeed a change from the previous methods.

The nice thing about returning an `io.ReadCloser` is that we don't have to buffer the entire response in memory. We can write the response to a file as a stream, so the memory of the CLI process should stay relatively constant.

With `ioutil.ReadAll` we have to read the entire profile into the CLI process memory, which could be quite large. I don't know if that's actually a probably in this case, but in general using an `io.Reader` or `io.ReadCloser` will give the caller the option to either stream or buffer, where as returning a `[]byte` forces them to use the fully buffered response.

kisunji (approved):
LGTM!

markan (approved):
Looks reasonable to me. TIL; Hadn't known this interface existed.

command/debug/debug_test.go
```diff
@@ -63,6 +62,16 @@ func TestDebugCommand(t *testing.T) {
 	require.Equal(t, 0, code)
 	require.Equal(t, "", ui.ErrorWriter.String())
 
+	expected := fs.Expected(t,
+		fs.WithDir("debug",
+			fs.WithFile("agent.json", "", fs.MatchAnyFileContent),
+			fs.WithFile("host.json", "", fs.MatchAnyFileContent),
+			fs.WithFile("index.json", "", fs.MatchAnyFileContent),
+			fs.WithFile("members.json", "", fs.MatchAnyFileContent),
```
dhiaayachi, line 70: The target name is cluster, I think we should rename both for a better user experience. I'm not sure though if renaming a debug subcommand is considered a breaking change.
Also, I believe we have a reference to it as a cluster target in the documentation

dhiaayachi (approved):
Thank you for those improvements @dnephin. LGTM!!

I have a minor comment about the fact that the file and the subcommand have different names now and the possibility to rename the subcommand as well but feel free to merge it as is and save that for a future PR.

command/debug/debug_test.go
dnephin: Good question! I believe I have handled that in the second commit, specifically here: https://github.com/hashicorp/consul/pull/10804/commits/58af70c2207945ce7c465279f78cd2eec122d1fb#diff-e4c25813d9467f20b9374e5dd16394ccf77aaf1db8fee3d1c7bba52b716b3cdeR551-R552

The way it should work now is that:
* both `cluster` and `members` should be accepted values for the `-capture` flag, and they should both do the same thing
* the help text should say `members` now

So I believe this handles backwards compatibility and ensures the flag values match the filename. 

But good call on the docs! I missed an update to https://www.consul.io/commands/debug#capture-targets 

I'll add a commit to this PR to make that change.

dnephin:
test failures are known flakes, and I confirmed they pass locally.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/430041.

PR #10806, opened 2021-08-06
debug: use human readable dates for filenames and improve the tests

dnephin:
Branched from #10804

The unix timestamps that were used previously make the debug data a little bit more difficult to consume. By using human readable dates we can easily see when the profile data was collected.

Also moves `profile.prof`, `trace.out`, `consul.log`, and `metrics.json` out of the sub-directories. There is only one of each of these, so we can put them in the top-level directory.

Also improves the test coverage. Two test cases are removed and the assertions from those cases are moved to `TestDebugCommand`.

Now `TestDebugCommand` is able to validate the contents of all files. This change reduces the test runtime of the command/debug package by almost 50%, while also improving the test coverage by making much more strict assertions about the files and their contents.

### DIFF
--- a/command/debug/debug.go
+++ b/command/debug/debug.go
@@ -18,10 +18,9 @@ import (
 	"syscall"
 	"time"
 
-	"golang.org/x/sync/errgroup"
-
 	"github.com/hashicorp/go-multierror"
 	"github.com/mitchellh/cli"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/consul/command/flags"
@@ -86,8 +85,6 @@ type cmd struct {
 	// validateTiming can be used to skip validation of interval, duration. This
 	// is primarily useful for testing
 	validateTiming bool
-
-	index *debugIndex
 }
 
 // debugIndex is used to manage the summary of all data recorded
@@ -178,15 +175,6 @@ func (c *cmd) Run(args []string) int {
 	c.UI.Info(fmt.Sprintf("        Output: '%s'", archiveName))
 	c.UI.Info(fmt.Sprintf("       Capture: '%s'", strings.Join(c.capture, ", ")))
 
-	// Record some information for the index at the root of the archive
-	index := &debugIndex{
-		Version:      debugProtocolVersion,
-		AgentVersion: version,
-		Interval:     c.interval.String(),
-		Duration:     c.duration.String(),
-		Targets:      c.capture,
-	}
-
 	// Add the extra grace period to ensure
 	// all intervals will be captured within the time allotted
 	c.duration = c.duration + debugDurationGrace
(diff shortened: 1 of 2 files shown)


command/debug/debug_test.go
```diff
@@ -64,17 +71,75 @@ func TestDebugCommand(t *testing.T) {
...
-			fs.WithFile("host.json", "", fs.MatchAnyFileContent),
-			fs.WithFile("index.json", "", fs.MatchAnyFileContent),
-			fs.WithFile("members.json", "", fs.MatchAnyFileContent),
-			// TODO: make the sub-directory names predictable)
+			fs.WithFile("index.json", "", fs.MatchFileContent(validIndexJSON)),
+			fs.WithFile("agent.json", "", fs.MatchFileContent(validJSON)),
+			fs.WithFile("host.json", "", fs.MatchFileContent(validJSON)),
+			fs.WithFile("members.json", "", fs.MatchFileContent(validJSON)),
+			fs.WithFile("metrics.json", "", fs.MatchAnyFileContent),
+			fs.WithFile("consul.log", "", fs.MatchFileContent(validLogFile)),
```
dnephin, line 79: Another possible improvement here might be to ensure the files contain at least one line. We may need to run a goroutine to perform some operations to ensure that some logs and metrics are emitted. 

Metrics in particular are difficult to capture because we'd have to wait the 10s interval window, which makes the test a bit slower to run.

command/debug/debug.go
```diff
@@ -133,6 +139,9 @@ func (c *cmd) init() {
 	c.help = flags.Usage(help, c.flags)
 
 	c.validateTiming = true
+	c.timeNow = func() time.Time {
+		return time.Now().UTC()
```
markan, line 143: I see we use both UTC and Local times. It looks like the usage is to use Localtime for output and UTC for files, but I've always found it a bit jarring to mix and match like that. My preference is UTC throughout, but I know reasonable people differ on this. Thoughts?

markan (commented):
Generally looks good, but one question about timestamps

command/debug/debug.go
```diff
@@ -373,43 +373,32 @@ func (c *cmd) captureInterval(ctx context.Context) error {
...
 }
 
-func (c *cmd) createTimestampDir(timestamp int64) (string, error) {
-	// Make the directory that will store all captured data
-	// for this interval
-	timestampDir := fmt.Sprintf("%s/%d", c.output, timestamp)
-	err := os.MkdirAll(timestampDir, 0755)
-	if err != nil {
-		return "", err
+func makeTimestampDir(base string, now time.Time) (string, error) {
```
dhiaayachi, line 393: We should continue to get rid of `Timestamp`, what about `makeCaptureDir`?

command/debug/debug.go
```diff
@@ -105,10 +105,16 @@ type debugIndex struct {
...
+// underscores. This is to make it more convenient to untar these files, because
+// tar assumes colons indicate the file is on a remote host, unless --force-local
+// is used.
+const timeDateFormat = "2006-01-02T15-04-05Z0700"
+
 func (c *cmd) init() {
 	c.flags = flag.NewFlagSet("", flag.ContinueOnError)
 
-	defaultFilename := fmt.Sprintf("consul-debug-%d", time.Now().Unix())
+	defaultFilename := fmt.Sprintf("consul-debug-%v", time.Now().Format(timeDateFormat))
```
dhiaayachi, line 117: We are changing from a Unix time to a second precision time. Could we end up having duplicates, so overwriting other files? 
I'm thinking of the use case where the interval is under a second.

dhiaayachi (commented):
Thank you for cleaning this up @dnephin !!
I have a question about the time precision and a suggested cosmetic change.

command/debug/debug.go
dhiaayachi: yes that would work too. I also saw some mentions of timestamp as var names

command/debug/debug.go
dnephin: I've made this change. For the time we print to stdout I used `time.Now().UTC()` and not this `c.timeNow`, because the intent of this `c.timeNow` is to make filenames predictable in tests, and if we call it extra times for stdout it becomes harder to predict the filenames.

command/debug/debug.go
dnephin: Good catch, fixed those now as well

markan (approved):

dhiaayachi (approved):
LGTM!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/430133.

PR #10807, opened 2021-08-06
config: remove ACLDatacenter

dnephin:
The `ACLDatacenter` field in `agent/config.RuntimeConfig` and `agent/consul.Config` has been unnecessary for a while now. It was always set to the same value as `PrimaryDatacenter`. So we can remove the duplicate field and use `PrimaryDatacenter`
directly. The field remains in `agent/config.Config` which is necessary for backwards compatibility for now.

This change was made by GoLand refactor, which did most of the work for me.

Also , the constructor for Server is not at all the appropriate place to be setting default values for a config struct that was passed in. In production this value is always set from agent/config. In tests we should set the default in a test helper. The second commit removes this defaulting.

### DIFF
--- a/agent/acl_endpoint_legacy.go
+++ b/agent/acl_endpoint_legacy.go
@@ -19,7 +19,7 @@ func (s *HTTPHandlers) ACLDestroy(resp http.ResponseWriter, req *http.Request) (
 	}
 
 	args := structs.ACLRequest{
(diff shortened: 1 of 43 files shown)


rboyer (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/428753.

PR #10743, opened 2021-07-30
acl: decouple filtering from ACLResolver and remove a couple methods

dnephin:
Branched from #10742

Best viewed by individual commit. There is more context in the commit messages.

This PR is another small incremental step toward two goals: making the ACL system easier to understand (by making it easier to trace the code), and making the code easier to change (by improving the structure). No structural changes yet, but these small changes will make it easier to move `ACLResolver` out of the `agent/consul` package.

PR #10808, opened 2021-08-06
acl: isolate the config used by ACLResolver

dnephin:
Branched from #10807, and continuing the work in #10743 and #10742

This PR is another step toward decoupling the `ACLResolver` from the rest of the `agent/consul` package, to allow us to move the ACLResolver into an acl package.

This PR creates a new `ACLResolverSettings` (see note below on the name) which contains all the fields needed by the `ACLResolver`. All previous fields that existed on `agent/config.RuntimeConfig` and `agent/consul.Config` are replaced by this new struct. This removes the need to copy the fields, and allows us to pass the struct through from config to `ACLResolver`. The fields on `agent/config.Config` remain unmodified, as those are necessary for backwards compatibility.

Note: `ACLResolverSettings` is probably not the best name. Suggests for other names welcome. `Config` is sadly overloaded in this context. Once the `ACLResolver` is moved into its own package a name should be easier to find because the package will scope the name.


agent/agent.go
```diff
@@ -1114,21 +1115,6 @@ func newConsulConfig(runtimeCfg *config.RuntimeConfig, logger hclog.Logger) (*co
...
-	}
-	if runtimeCfg.ACLRoleTTL != 0 {
-		cfg.ACLRoleTTL = runtimeCfg.ACLRoleTTL
-	}
-	if runtimeCfg.ACLDefaultPolicy != "" {
-		cfg.ACLDefaultPolicy = runtimeCfg.ACLDefaultPolicy
-	}
-	if runtimeCfg.ACLDownPolicy != "" {
-		cfg.ACLDownPolicy = runtimeCfg.ACLDownPolicy
-	}
```
dnephin, line 1131: I notice that `ACLDisabledTTL` is missing from this list. I believe that means that the `acl.disabled_ttl` that we set in `agent/config.NonUserSource` (`120s`) is completely ignored. The value that is actually used is the one from `consul.DefaultConfig`, which is set to `30s`. I guess that's probably not a huge problem, just a potential source of confusion.

I can look at cleaning that up in a follow up.

dnephin: PR'ed this cleanup in #10814

agent/consul/acl.go
```diff
@@ -211,6 +212,40 @@ type ACLResolverConfig struct {
 	Tokens *token.Store
 }
 
+// TODO: rename the fields to remove the ACL prefix
+type ACLResolverSettings struct {
```
rboyer, line 216: Probably best to wait until this moves elsewhere to bother renaming. Then you'll have the package name to help prefix it nicely.

agent/consul/acl.go
```diff
@@ -211,6 +212,40 @@ type ACLResolverConfig struct {
...
+	ACLsEnabled  bool
+	Datacenter   string
+	NodeName     string
+	ACLPolicyTTL time.Duration
+	ACLTokenTTL  time.Duration
+	ACLRoleTTL   time.Duration
```
rboyer, line 222: Why not carry the comments over from their prior home?

agent/consul/acl.go
```diff
@@ -237,7 +272,8 @@ type ACLResolverConfig struct {
 //   upon.
 //
 type ACLResolver struct {
-	config *Config
+	// TODO: store the ACLResolverConfig as a field instead of copying all the fields onto ACLResolver.
```
rboyer, line 275:
```suggestion
	// TODO: store the ACLResolverSettings as a field instead of copying all the fields onto ACLResolver.
```

rboyer: Aren't you already doing what the comment says to do?

agent/consul/acl.go
dnephin: Oh ya, I should do that. I missed it for the first few fields I did. I'll do that now

agent/consul/acl.go
dnephin: oops, yes this is done. The comment was from an earlier revision and I forgot to remove it. Will remove the comment now.

rboyer (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/429043.

PR #10810, opened 2021-08-09
docs: Clarify ingress gateway's -address flag

### LINKED ISSUE #9849: Documentation around -address and Port for ingress gateways is not clear, plus an improvement
#### Problem Description
We have been struggling with the ingress gateways recently. I asked @blake and he was similarly unsure of the behavior, but admitted it was probably just a documentation issue. Here's what we read:

https://learn.hashicorp.com/tutorials/consul/service-mesh-ingress-gateways

https://www.consul.io/commands/connect/envoy#address

https://www.consul.io/docs/connect/config-entries/ingress-gateway#port

The tutorial specifies that the config entry for the ingress gateway will use `Port = 8080` but that the ingress gateway's envoy proxy should start with the argument `-address '{{ GetInterfaceIP "eth0" }}:8888'`

It's not clear why the two separate ports are specified, and port 8888 is never referenced again.

Through some digging and experimenting, I have found the following to be true:

##### -address
`-address` specifies the port that the envoy proxy will present a readiness port for the consul agent to check the health. I was able to deduce this by hitting envoy's /config_dump and seeing that the definition for the listener on the port specified in `-address` is defined simply as responding 200 OK to /ready and 404 to anything else (if you're curious, expand the following)

<details>
  <summary>Click to view the excerpt response from /config_dump</summary>
  
```
...
         "address": "10.1.28.102",
         "port_value": 8888
        }
       },
       "filter_chains": [
        {
         "filters": [
          {
           "name": "envoy.http_connection_manager",
           "hidden_envoy_deprecated_config": {
            "stat_prefix": "envoy_ready",
            "route_config": {
             "virtual_hosts": [
              {
               "routes": [
                {
                 "match": {
                  "path": "/ready"
                 },
                 "route": {
                  "cluster": "self_admin",
                  "prefix_rewrite": "/ready"
                 }
                },
                {
                 "direct_response": {
                  "status": 404
                 },
                 "match": {
                  "prefix": "/"
                 }
                }
               ],
               "domains": [
                "*"
               ],
               "name": "self_admin"
              }
             ],
             "name": "self_admin_route"
            },
            "codec_type": "HTTP1",
            "http_filters": [
             {
              "name": "envoy.router"
             }
            ]
           }
          }
...
```
</details>

##### ingress-gateway Port

The ingress-gateway's config entry Port is where envoy will listen for incoming traffic. Interestingly it seems to bind to the interface specified in `-address`, so you can't just omit `-address`

#### Requests

1. I think the documentation needs to be made clearer, though I'm not quite close enough to know the way to word things. Some key information I think needs to be called out:
    a. The inbound traffic destined for the ingress gateway should be sent to the port specified on the config entry, not the port listed for `-address`
    b. The port for `-address` *must* differ from the one specified in the config entry (we assumed they should be the same and tried setting them as such) or the /ready endpoint will clobber the ingress listener
    c. The interface that you use for `-address` will become the interface that the config entry-specified port will listen on. This one is particularly not obvious -- we thought "ok we'll just omit `-address` entirely" but then the ingress listener ends up binding to 127.0.0.1 which isn't helpful
2. I would love to see `-address` supplemented or replaced with an option *just* for specifying the interface that the ingress listener should bind to (or if there is an option for that already, better documentation pointing to that fact). As it stands, you are forced to configure the port for the `/ready` listener merely because you want to configure the interface used for the ingress listener, and it's not obvious that your one flag is being used to drive both things. I would rather see the `/ready` listener always bind to 127.0.0.1 and choose its own port automatically, ideally a non-conflicting one, like how sidecar port registration works. If there are use cases for explicitly specifying the interface and/or port then maybe renaming this argument to be more clear about its intended target

blake:
Clarify the function of `-address` flag when instantiating an ingress gateway.

Resolves #9849

### DIFF
--- a/website/content/commands/connect/envoy.mdx
+++ b/website/content/commands/connect/envoy.mdx
@@ -136,15 +136,16 @@ proxy configuration needed.
   `-register`. This takes the form of `<ip address>:<port>` but also supports go-sockaddr
   templates.
 
-  For terminating and mesh gateways, this is the IP and port on which
-  to receive traffic from services in the mesh. For ingress gateways, this is
-  the IP and port at which to instantiate a `/ready` HTTP endpoint which is used
-  by Consul for health checking the gateway. The specified IP will also be used
-  by the ingress gateway when instantiating user-defined listeners
-  configured in the
+  If Envoy is configured as a terminating or mesh gateway, traffic from services
+  within the mesh will be received at the specified IP and port.
+
+  If Envoy is configured as an ingress gateway, a `/ready` HTTP endpoint will be
+  instantiated at the specified IP and port. Consul uses `/ready` HTTP endpoints
+  to check gateway health. The specified IP will also be used by the ingress
+  gateway when instantiating user-defined listeners configured in the
   [ingress gateway](/docs/connect/gateways/ingress-gateway) configuration entry.
 
-  ~> **Note**: Ensure that user-defined ingress gateway listeners utilize a
+  ~> **Note**: Ensure that user-defined ingress gateway listeners use a
      different port than the port specified in `-address` so that they do not
      conflict with the health check endpoint.
 
@@ -153,7 +154,7 @@ proxy configuration needed.
   use `/dev/null`.
 
 - `-bind-address` - The bind address to use instead of the default binding rules.
-  Speciifed as `<name>=<ip>:<port>` pairs. This flag may be specified multiple times
+  Specified as `<name>=<ip>:<port>` pairs. This flag may be used multiple times
   to add multiple bind addresses.
 
 - `-expose-servers` - Expose the servers for WAN federation via this mesh
--- a/website/content/commands/connect/envoy.mdx
+++ b/website/content/commands/connect/envoy.mdx
@@ -136,12 +136,24 @@ proxy configuration needed.
   `-register`. This takes the form of `<ip address>:<port>` but also supports go-sockaddr
   templates.
 
+  For terminating and mesh gateways, this is the IP and port on which
+  to receive traffic from services in the mesh. For ingress gateways, this is
+  the IP and port at which to instantiate a `/ready` HTTP endpoint which is used
+  by Consul for health checking the gateway. The specified IP will also be used
+  by the ingress gateway when instantiating user-defined listeners
+  configured in the
+  [ingress gateway](/docs/connect/gateways/ingress-gateway) configuration entry.
+
+  ~> **Note**: Ensure that user-defined ingress gateway listeners utilize a
+     different port than the port specified in `-address` so that they do not
+     conflict with the health check endpoint.
+
 - `-admin-access-log-path` The path to write the access log for the administration
   server. If no access log is desired specify `/dev/null`. By default it will
   use `/dev/null`.
 
-- `-bind-address` - The bind address to use instead of the default binding rules
-  given as `<name>=<ip>:<port>` pairs. This flag may be specified multiple times
+- `-bind-address` - The bind address to use instead of the default binding rules.
+  Speciifed as `<name>=<ip>:<port>` pairs. This flag may be specified multiple times
   to add multiple bind addresses.
 
 - `-expose-servers` - Expose the servers for WAN federation via this mesh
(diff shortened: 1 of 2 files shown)


website/content/commands/connect/envoy.mdx
```diff
@@ -136,12 +136,24 @@ proxy configuration needed.
   `-register`. This takes the form of `<ip address>:<port>` but also supports go-sockaddr
   templates.
 
+  For terminating and mesh gateways, this is the IP and port on which
+  to receive traffic from services in the mesh. For ingress gateways, this is
+  the IP and port at which to instantiate a `/ready` HTTP endpoint which is used
+  by Consul for health checking the gateway. The specified IP will also be used
```
trujillo-adam, line 142:
```suggestion
  If Envoy is configured as a terminating or mesh gateway, traffic from services within the mesh will be received at the specified IP and port. 
  
  If Envoy is configured as an ingress gateway, a `/ready` HTTP endpoint will be instantiated at the specified IP and port. Consul uses `/ready` HTTP endpoints to check gateway health. The specified IP will also be used
```
We generally avoid passive voice, but in this instance I think using PV improves readability. Your original text was grammatically sound, so feel free to ignore this suggestion if you disagree.

website/content/commands/connect/envoy.mdx
```diff
@@ -136,12 +136,24 @@ proxy configuration needed.
...
+  by the ingress gateway when instantiating user-defined listeners
+  configured in the
+  [ingress gateway](/docs/connect/gateways/ingress-gateway) configuration entry.
+
+  ~> **Note**: Ensure that user-defined ingress gateway listeners utilize a
```
trujillo-adam, line 147:
```suggestion
  ~> **Note**: Ensure that user-defined ingress gateway listeners use a
```
Nothing wrong with "utilize"; "use" is just shorter.

website/content/commands/connect/envoy.mdx
```diff
@@ -136,12 +136,24 @@ proxy configuration needed.
...
+     conflict with the health check endpoint.
+
 - `-admin-access-log-path` The path to write the access log for the administration
   server. If no access log is desired specify `/dev/null`. By default it will
   use `/dev/null`.
 
-- `-bind-address` - The bind address to use instead of the default binding rules
-  given as `<name>=<ip>:<port>` pairs. This flag may be specified multiple times
+- `-bind-address` - The bind address to use instead of the default binding rules.
+  Speciifed as `<name>=<ip>:<port>` pairs. This flag may be specified multiple times
```
trujillo-adam, line 156:
```suggestion
  Specified as `<name>=<ip>:<port>` pairs. This flag may be used multiple times
```

trujillo-adam (approved):
Looks good. A few minor suggestions

blake:
> Looks good. A few minor suggestions

Thanks for the suggestions. These definitely improve the readability of the parameter descriptions.

website/content/docs/connect/config-entries/ingress-gateway.mdx
```diff
@@ -784,7 +784,11 @@ spec:
         {
           name: 'Port',
           type: 'int: 0',
-          description: 'The port that the listener should receive traffic on.',
+          description: `The port on which the listener should receive traffic.
```
chrisjohnson, line 787: A hint that the interface/IP used here is derived from the interface/IP supplied to `-address` would also be helpful I think.

blake: I've modified this paragraph to note the port will be bound to the IP specified in `-address`.

chrisjohnson:
LGTM, thanks!!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/427119.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 8aa89c2c12a56218ad24dbd8234f8a426111f308 onto `stable-website` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 8aa89c2c12a56218ad24dbd8234f8a426111f308 onto `release/1.10.x` succeeded!

PR #10811, opened 2021-08-09
api: expose upstream routing configurations in topology view

kisunji:
Some users are defining routing configurations that do not have associated services. In the UI, these configs are not surfaced in the topology visualization.

Todo:

- [ ] Frontend changes to show routing config on topology view
- [ ] Frontend changes to show a new routing config view (as opposed to services view)


agent/consul/state/catalog.go
```diff
@@ -2943,14 +2943,8 @@ func (s *Store) ServiceTopology(
 		maxIdx = idx
 	}
 
-	var (
-		seenUpstreams   = make(map[string]struct{})
```
kisunji, line 2947: I was trying to refactor upstream and downstream processing but it got too complex for this PR.
`seenUpstreams` and `seenDownstreams` were redundant here and can be safely removed

agent/ui_endpoint.go
```diff
@@ -343,6 +343,20 @@ RPC:
...
 		upstreamResp = append(upstreamResp, &sum)
 	}
+	for k, v := range out.ServiceTopology.UpstreamSources {
+		if v == structs.TopologySourceRoutingConfig {
+			sn := structs.ServiceNameFromString(k)
+			sum := ServiceTopologySummary{
+				ServiceSummary: ServiceSummary{
+					Name:           sn.Name,
+					EnterpriseMeta: sn.EnterpriseMeta,
+				},
```
kisunji, line 353: How this looks is WIP; will need to coordinate with FE to expose the fields they need to navigate to a routing-config view.

agent/ui_endpoint.go
```diff
@@ -343,6 +343,20 @@ RPC:
 		}
 		upstreamResp = append(upstreamResp, &sum)
 	}
+	for k, v := range out.ServiceTopology.UpstreamSources {
+		if v == structs.TopologySourceRoutingConfig {
+			sn := structs.ServiceNameFromString(k)
+			sum := ServiceTopologySummary{
+				ServiceSummary: ServiceSummary{
```
kisunji, line 350: I'm not a fan of pretending this is a service but I've created a dummy ServiceSummary for the response body, since intention and source get flattened into a ServiceTopologySummary object per "service". This is least friction/effort for frontend as well.

agent/consul/state/catalog.go
```diff
@@ -3077,7 +3081,7 @@ func (s *Store) ServiceTopology(
 		downstreamSources[svc.Name.String()] = source
 	}
 
-	idx, unfilteredDownstreams, err := s.combinedServiceNodesTxn(tx, ws, downstreamNames)
+	idx, unfilteredDownstreams, _, err := s.combinedServiceNodesTxn(tx, ws, downstreamNames)
```
kisunji, line 3084: Currently, services infer their downstreams _through_ routing configs (i.e. they are not aware of routing configs and point directly to their downstreams).
Not sure if we want to expose routing config downstreams here as well?

agent/consul/state/catalog.go
```diff
@@ -3142,32 +3146,45 @@ func (s *Store) ServiceTopology(
...
 // This enabled aggregating checks statuses across both.
-func (s *Store) combinedServiceNodesTxn(tx ReadTxn, ws memdb.WatchSet, names []structs.ServiceName) (uint64, structs.CheckServiceNodes, error) {
+// Returns a slice of ServiceNames not associated with a ServiceNode instance.
+func (s *Store) combinedServiceNodesTxn(tx ReadTxn, ws memdb.WatchSet, names []structs.ServiceName) (uint64, structs.CheckServiceNodes, []structs.ServiceName, error) {
 	var (
-		maxIdx uint64
-		resp   structs.CheckServiceNodes
+		maxIdx   uint64
+		resp     structs.CheckServiceNodes
+		notFound []structs.ServiceName
```
freddygv, line 3154: It would probably be preferable to move these checks for `notFound` services up to the ServiceTopology function, rather than returning them here:

- `combinedServiceNodesTxn` is called by our most heavily used endpoints (`/health`), and this new return value adds some overhead for a single less-common use-case.
- We should aim to not return more than 3 values from a function.

freddygv (commented):
Nice progress @kisunji! The synthetic ServiceSummary in the ui endpoint looks good to me. I left some comments below.

Let me know if you want to chat about the test changes needed for this.

agent/consul/state/catalog.go
kisunji: Sounds good! I wanted to avoid a `N ServiceNames * n CheckServiceNodes` operation to find unused ServiceNames by doing it in one pass here but it does clutter the func signature a bit.

kisunji: re: `combinedServiceNodesTxn` being used by `/health`, I only see this func being used in ServiceTopology-related endpoints. Is there an RPC call I'm missing?

agent/consul/state/catalog.go
```diff
@@ -3020,13 +3014,42 @@ func (s *Store) ServiceTopology(
 	if err != nil {
 		return 0, nil, fmt.Errorf("failed to query intentions for %s", sn.String())
 	}
+
+	var foundUpstreamNames = make(map[structs.ServiceName]bool)
```
kisunji, line 3018: @freddygv Would this be preferred over returning an array of notFoundUpstreamNames in the [previous commit](https://github.com/hashicorp/consul/pull/10811/commits/2a37b868ccf394a1217a3d06a20b8017859555da#r685576288)?

A downside to doing it here is that I'm weaving more specific logic for upstreams which might make refactoring more difficult in the future

agent/consul/state/catalog.go
freddygv: Ah you're right. I got this function confused with `parseCheckServiceNodes`. 

I think what you have now looks good though.

agent/consul/internal_endpoint_test.go
```diff
@@ -1897,6 +1898,71 @@ func TestInternal_ServiceTopology(t *testing.T) {
...
+				"routing-config": {DefaultAllow: true, Allowed: true},
+			}
+			require.Equal(r, expectUp, out.ServiceTopology.UpstreamDecisions)
+
+			expectUpstreamSources := map[string]string{
+				"counting":       structs.TopologySourceDefaultAllow,
+				"counting-v2":    structs.TopologySourceDefaultAllow,
+				"routing-config": structs.TopologySourceRoutingConfig,
+			}
+			require.Equal(r, expectUpstreamSources, out.ServiceTopology.UpstreamSources)
```
kisunji, line 1959: I'm a little surprised by this but I think it's due to default intentions? The UI endpoint ends up filtering these out

agent/ui_endpoint_test.go
```diff
@@ -1397,34 +1398,59 @@ func TestUIServiceTopology(t *testing.T) {
...
+			require.NotNil(t, obj)
+			result := obj.(ServiceTopology)
+			clearUnexportedFields(result)
+
+			require.Equal(r, *tc.want, result)
+		})
+	}
 
-			expect := ServiceTopology{
+	tcs := []testCase{
```
kisunji, line 1430: I refactored these tests to be table-based

agent/ui_endpoint_test.go
```diff
@@ -1397,34 +1398,59 @@ func TestUIServiceTopology(t *testing.T) {
...
-
-	t.Run("request with unsupported kind", func(t *testing.T) {
-		req, _ := http.NewRequest("GET", "/v1/internal/ui/service-topology/ingress?kind=not-a-kind", nil)
-		resp := httptest.NewRecorder()
-		obj, err := a.srv.UIServiceTopology(resp, req)
-		require.Nil(t, err)
-		require.Nil(t, obj)
-		require.Equal(t, `Unsupported service kind "not-a-kind"`, resp.Body.String())
-	})
+	type testCase struct {
```
freddygv, line 1401: An alternative is to use an anonymous struct for these, since the type is only used once:

```suggestion
	tcs := []struct {
        name    string
		httpReq *http.Request
		want    *ServiceTopology
		wantErr string
    }{
        {
        	...
        },
    } 
```

Will leave that decision up to you though.

freddygv (commented):
This is getting really close, nice work! I like the conversion to the table driven tests.

I left some stylistic comments and one bug from the 1.10 release. I think it would be great to fix that up here.

agent/ui_endpoint_test.go
```diff
@@ -1397,34 +1398,59 @@ func TestUIServiceTopology(t *testing.T) {
...
+		name    string
+		httpReq *http.Request
+		want    *ServiceTopology
+		wantErr string
+	}
 
-	t.Run("ingress", func(t *testing.T) {
+	run := func(t *testing.T, tc testCase) {
```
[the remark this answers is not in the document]
dnephin: Hey @freddygv , this layout for table tests is the same one I use. I wrote up a short explanation of it here: https://github.com/hashicorp/consul/pull/10814#discussion_r686221701

I quite like this style because I think it allows the test to be read easily from top-to-bottom, instead of having to scroll back and forth to match the test logic with the test cases.

agent/ui_endpoint_test.go
dnephin: With the pattern I mention below, it is required to have a named `type testCase struct` as the first section of the table test.

agent/ui_endpoint.go
kisunji: Looks like I'll need to provide datacenter as well, which is going to be tricky!

agent/ui_endpoint_test.go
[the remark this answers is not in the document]
kisunji: I tried following @dnephin's pattern of putting the test logic closer to the testcase type (described [here](https://github.com/hashicorp/consul/pull/10814#discussion_r686221701)), which I thought was neat and legible. I can move this back down if you don't agree or would like to discuss the style elsewhere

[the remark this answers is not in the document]
kisunji: Oh woops dan was faster than me :D

agent/ui_endpoint_test.go
```diff
@@ -1590,70 +1583,326 @@ func TestUIServiceTopology(t *testing.T) {
...
+		d.checks = nil
+	}
+}
+
+func TestUIServiceTopology_RoutingConfigs(t *testing.T) {
+	if testing.Short() {
+		t.Skip("too slow for testing.Short")
+	}
 
-	t.Run("redis", func(t *testing.T) {
```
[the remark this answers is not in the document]
kisunji: Good catch! Not intentional

agent/consul/internal_endpoint_test.go
```diff
@@ -1897,6 +1898,66 @@ func TestInternal_ServiceTopology(t *testing.T) {
...
+			}
+			var out structs.IndexedServiceTopology
+			require.NoError(r, msgpackrpc.CallWithCodec(codec, "Internal.ServiceTopology", &args, &out))
+			require.False(r, out.FilteredByACLs)
+			require.Equal(r, "http", out.ServiceTopology.MetricsProtocol)
+
+			require.Empty(r, out.ServiceTopology.Downstreams)
+
+			expectDown := map[string]structs.IntentionDecisionSummary{
+				"counting":    {DefaultAllow: true, Allowed: true},
```
freddygv, line 1933: These downstream decision summaries and sources should probably be excluded as well, since I don't believe either of them are transparent proxies

kisunji: Ah I misunderstood your last comment about downstreams. I added the `if hasTransparent` check to downstream intentions as well

agent/consul/state/catalog.go
```diff
@@ -3057,24 +3075,27 @@ func (s *Store) ServiceTopology(
...
-		var source string
-		switch {
-		case svc.Decision.HasExact:
-			source = structs.TopologySourceSpecificIntention
-		case svc.Decision.DefaultAllow:
-			source = structs.TopologySourceDefaultAllow
-		default:
-			source = structs.TopologySourceWildcardIntention
+	// Only transparent proxies have downstreams from intentions
+	if hasTransparent {
```
freddygv, line 3079: Sorry, I think I miscommunicated here.

A service being a transparent proxy only influences its upstreams, not its downstreams.

This means that a downstream is a downstream from intentions if the downstream is a transparent proxy. hasTransparent on the other hand is based on the target service.

I think what would clarify this is to look at the usage and comment for `tproxyMap` below:

> // Store downstreams with at least one instance in transparent proxy mode.
> // This is to avoid returning downstreams from intentions when none of the downstreams are transparent proxies.

agent/consul/state/catalog.go
```diff
@@ -3092,44 +3092,67 @@ func (s *Store) ServiceTopology(
...
+			default:
+				source = structs.TopologySourceWildcardIntention
+			}
+			upstreamSources[svc.Name.String()] = source
+		}
 	}
-	if idx > maxIdx {
-		maxIdx = idx
+
+	matchEntry := structs.IntentionMatchEntry{
```
kisunji, line 3133: Re-ordering function calls to match the changes in the downstream section

agent/consul/state/catalog.go
```diff
@@ -3226,6 +3244,26 @@ func (s *Store) ServiceTopology(
...
+		return 0, nil, fmt.Errorf("failed to query intentions for %s", sn.String())
+	}
+	for _, dn := range downstreamNames {
+		decision, err := s.IntentionDecision(dn.Name, dn.NamespaceOrDefault(), dstIntentions, structs.IntentionMatchSource, defaultAllow, false)
+		if err != nil {
+			return 0, nil, fmt.Errorf("failed to get intention decision from (%s) to (%s): %v",
+				dn.String(), sn.String(), err)
+		}
+		downstreamDecisions[dn.String()] = decision
+	}
```
kisunji, line 3265: Re-ordered this to be earlier so that the last line 
```
downstreamDecisions[dn.String()] = decision
```
can be deleted if the service name is not in `tproxyMap`

agent/ui_endpoint.go
```diff
@@ -343,6 +343,21 @@ RPC:
...
+					Datacenter:     args.Datacenter,
```
kisunji, line 351: Datacenter passed from `args` required for FE

PR #10822, opened 2021-08-10
Update namespaced replication token example

krarey:
This documentation patch updates the sample replication token policy provided for multi-cluster federation using namespaces.

As ACLs are namespace-scoped resources, the `acl = "write"` permission must be applicable to all namespaces. Otherwise secondary datacenters will fail to replicate ACL tokens and policies outside the `default` namespace.

### DIFF
--- a/website/content/docs/k8s/installation/multi-cluster/vms-and-kubernetes.mdx
+++ b/website/content/docs/k8s/installation/multi-cluster/vms-and-kubernetes.mdx
@@ -232,7 +232,6 @@ You'll need:
    If you're running Consul Enterprise you'll need the rules:
 
    ```hcl
-   acl = "write"
    operator = "write"
    agent_prefix "" {
      policy = "read"


dnephin (approved):

I believe this is correct. `acl = "write"` at the top level would only grant permission to `acl:write` in the default namespace. By moving into the `namespace_prefix ""` block, it now grants `acl:write` on all namespaces.

ACL replication will need to be able to write acls in all namespace, so this change is correct.

dnephin:
CI did not run, but this is a docs change, and Vercel deployed fine, so I think that's ok. I'm going to merge it anyway, and we'll get that issue sorted out as a follow up.

dnephin:
Oh, but because CI is not working , the backport to the website-stable branch will also probably will not work, so maybe we should get that sorted out before we merge.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/492778.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 7ed6e3b7901adf6716f7e92f4811761c3ea4a7b6 onto `stable-website` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 7ed6e3b7901adf6716f7e92f4811761c3ea4a7b6 onto `release/1.10.x` succeeded!

PR #10837, opened 2021-08-12
Improve documentation around using DNS with ACLs Enabled

### LINKED ISSUE #10723: Is it possible to access Consul DNS that has ACL enabled but without prepared queries?
#### Feature Description

Currently after troubleshooting I realized that when ACL is enforced, queries to `.service.consul` and others will fail with NXDOMAIN. Turning off ACL then makes Consul DNS works as expected. Unfortunately, after some reading, CMIIW, to access Consul DNS you need to first create a prepared query first using the prepared query API, and then access the service from `.query.consul` domain.

If this is true, does that mean I have to register all services first using API before my other external services can discover those using Consul DNS? Is there a way to just allow Consul DNS to be accessible everywhere but with ACL enabled? Seems to be a chore if I have to create prepared queries everytime, especially knowing that Nomad does not seem to create one for me either.

#### Use Case(s)

Well, there are cases when my DNS is accessible only through local network and I have to use DNS SRV records for service discovery. Even this document gives an example of this: [https://learn.hashicorp.com/tutorials/consul/load-balancing-haproxy](https://learn.hashicorp.com/tutorials/consul/load-balancing-haproxy). If I want to implement something like the document describes, how to do it with ACL enabled? Without having to create/delete prepared queries all the time?

jkirschner-hashicorp:
Closes #10723 

Documentation changes motivated by a practitioner's confusion around getting DNS working with ACLs. Ready for review.

Related changes have also been submitted for the [Learn site](https://github.com/hashicorp/learn/pull/3783). The Learn site PR will be merged first. This PR contains a small amount of shared text between the two sites. I want to make sure that shared text stays in sync, which is why I want to address these PRs in an explicit sequence.

**Note**
The image was taken from a Learn tutorial, so it's an image we're already using. This is what it looks like in context of the ACL system page:

image


website/content/docs/discovery/dns.mdx
```diff
@@ -285,6 +285,39 @@ caching of DNS results. However, there are many situations in which caching is
...
+- Otherwise, the built-in [`anonymous` token](/docs/security/acl/acl-system#builtin-tokens).
+  Because the anonymous token is used when any request is made to Consul without
+  explicitly specifying a token, production deployments should not apply policies
+  needed for DNS to this token.
+
+Consul will either authorize or revoke the request, depending on the token's
+policy rules:
+
+- [node](/docs/agent/acl-rules.html#node-rules) read access is
+  required on the node serving DNS requests
```
dnephin, line 309: I took a quick look at `dns.go` to see if I could answer this question for myself, but it wasn't immediately obvious, so I thought I'd ask:

The ACL permissions are checked by the server that receives the RPC request, right? 

Why does the token need `node:read` on the node _serving_  (handling?) the DNS request? I would expect it to need `node:read` for any node in the response, which is the "node being queried", not the node handling the request.

website/content/docs/security/acl/acl-system.mdx
```diff
@@ -22,6 +22,10 @@ To learn how to setup the ACL system on an existing Consul datacenter, use the [
 ## ACL System Overview
 
 The ACL system is designed to be easy to use and fast to enforce while providing administrative insight.
+The diagram below shows the relationships between most of the components of the ACL system:
+
+![ACL system component relationships](/img/acl-token-policy-rule-relationship.png)
```
dnephin, line 27: Nice diagram! I think this does a great job of showing the relationship. It doesn't include Roles, or the Service/Node Identities, but I think that's probably good. Trying to include it all up front is probably confusing.

This is the diagram I made for the contrib docs that includes most thing and their relationships: https://github.com/hashicorp/consul/tree/main/contributing/acl#acl-entities

I think that level of detail is good for contributors, but probably way too much for users.

dnephin (commented):
Looks great! Just the one question about the node ACL permissions

website/content/docs/security/acl/acl-system.mdx
jkirschner-hashicorp: Good artists copy, as they say. Glad this already exists on the Learn site. Just re-using! I noticed that it doesn't include roles, either because roles aren't mentioned in the Learn tutorial or because the tutorial was written before roles were a thing in Consul.

I was very tempted to add roles and service/node identities. Especially roles ... I think they are an important thing to surface. But to get 80% of the value in 2% of the time, I decided to forego that for now.

And I'm glad you created a more developer-focused diagram for contributors!

website/content/docs/discovery/dns.mdx
jkirschner-hashicorp: I'm the wrong person to answer that question. I was trying to make the documentation around which policies are needed clearer than what exists today, which is just:
> Depending on your use case, the token used for DNS may need policy rules for nodes, services, and prepared queries. 

... which is not very helpful. I wrote that clarification based on what I could gather from readings lots of docs. I haven't had enough practice tracing through the RPC code yet to understand what's going on fully.

Do you feel like you understand what's happening with reasonable confidence? Here are a few questions we should be able to answer:
- For service lookups, do you need node permissions for the nodes on which the service resides? Or just service permissions?
- For node lookups, I assume you need `node:read` on the node being looked up.
- For [connect](https://www.consul.io/docs/discovery/dns#connect-capable-service-lookups) and [ingress](https://www.consul.io/docs/discovery/dns#ingress-service-lookups) lookups ... do you need the same rules as for standard service lookups?

dnephin: All good questions. I think I would be able to answer them, but I'll need to dig through the RPC code a bit to do so. If no one else has the answer to this readily available I'll try and take some time soon to do that digging.

I suspect at the very least this should say something along the lines of:

```suggestion
- [node](/docs/agent/acl-rules.html#node-rules) read access is
  required for any nodes that may be returned in DNS responses, which may be all nodes.
```

As for connect/ingress lookups, I believe those also only use `service:read`.

jkirschner-hashicorp: > required for any nodes that may be returned in DNS responses

Does that mean just for any `*.node.consul` lookups? Or also any nodes which host services to be included in `*.service.consul` lookups?

blake: > Does that mean just for any *.node.consul lookups? Or also any nodes which host services to be included in *.service.consul lookups?

This applies to explicit node address lookups (`*.node.consul`) and service lookups (`*.service.consul`). The docs for the [`Address`](https://www.consul.io/api-docs/agent/service#address) parameter state:

> Specifies the address of the service. If not provided, the agent's address is used as the address for the service during DNS queries.

The token used for the initial request must have `node:read` permissions for the node(s) on which the service is running in order to be able to resolve and return the agent/node IP address.

jkirschner-hashicorp: Thanks @blake and @dnephin. I've updated the wording given the information you provided. I also restructured this list to be organized by type of lookup rather than by type of permission, as that felt easier to understand.

website/content/docs/discovery/dns.mdx
```diff
@@ -285,6 +285,50 @@ caching of DNS results. However, there are many situations in which caching is
 desirable for performance and scalability. This is discussed more in the tutorial
 for [DNS caching](https://learn.hashicorp.com/tutorials/consul/dns-caching).
 
+## DNS with ACLs
```
blake, line 288: The majority of this page highlights features provided by the DNS interface, whereas this section explains how DNS interacts with another area of Consul – the ACL system. What do you think about moving this to the last section of the page – after the feature sections – or to a separate page altogether?

website/content/docs/discovery/dns.mdx
```diff
@@ -285,6 +285,50 @@ caching of DNS results. However, there are many situations in which caching is
...
+
+After enabling [Access Control Lists (ACLs)](/docs/security/acl/acl-system),
+DNS requests will fail with error `NXDOMAIN` unless ACL tokens have been
+configured with the necessary rules.
```
blake, line 292:
```suggestion
In order to use the DNS interface when [Access Control Lists (ACLs)](/docs/security/acl/acl-system) are enabled, you must first create ACL tokens with the necessary policies.
```

website/content/docs/discovery/dns.mdx
```diff
@@ -285,6 +285,50 @@ caching of DNS results. However, there are many situations in which caching is
...
+
+When a DNS request is made to a Consul agent, the agent will use a previously
+configured token with the request (since there is no means for the initial DNS
+request to specify a token):
```
blake, line 296:
```suggestion
Consul agent's resolve DNS requests using one of two preconfigured tokens:
```

website/content/docs/discovery/dns.mdx
```diff
@@ -285,6 +285,50 @@ caching of DNS results. However, there are many situations in which caching is
...
+
+- The agent's [`default` token](/docs/agent/options#acl_tokens_default)
+  token, if configured.
+- Otherwise, the built-in [`anonymous` token](/docs/security/acl/acl-system#builtin-tokens).
+  Because the anonymous token is used when any request is made to Consul without
+  explicitly specifying a token, production deployments should not apply policies
+  needed for DNS to this token.
+
+Consul will either authorize or revoke the request, depending on the token's
+policy rules:
```
blake, line 306:
```suggestion
Consul will either authorize the request, depending on the token's
policy rules:
```

website/content/docs/discovery/dns.mdx
```diff
@@ -285,6 +285,50 @@ caching of DNS results. However, there are many situations in which caching is
...
+DNS requests will fail with error `NXDOMAIN` unless ACL tokens have been
+configured with the necessary rules.
+
+When a DNS request is made to a Consul agent, the agent will use a previously
+configured token with the request (since there is no means for the initial DNS
+request to specify a token):
+
+- The agent's [`default` token](/docs/agent/options#acl_tokens_default)
+  token, if configured.
+- Otherwise, the built-in [`anonymous` token](/docs/security/acl/acl-system#builtin-tokens).
```
blake, line 300:
```suggestion
- The built-in [`anonymous` token](/docs/security/acl/acl-system#builtin-tokens), if a default token is not configured.
```

website/content/docs/discovery/dns.mdx
```diff
@@ -285,6 +285,50 @@ caching of DNS results. However, there are many situations in which caching is
...
+  - [`node:read`](/docs/security/acl/acl-rules#node-rules) permission for nodes
+    on which the target service instances are running.
+- [Prepared query lookups](#prepared-query-lookups) (`*.query.consul`) are a
+  special type of service lookup that require:
+  - [`query:read`](/docs/security/acl/acl-rules#prepared-query-rules)
+    permission for the target prepared query, and
+  - the same permissions as needed for a service lookup of the query's
+    [target service](/api/query#service-1)
+    ([`service:read`](/docs/security/acl/acl-rules#service-rules),
+    [`node:read`](/docs/security/acl/acl-rules#node-rules)).
```
blake, line 324: This would probably be easier to read as a table.

website/content/docs/security/acl/acl-system.mdx
blake: minor suggestion: It might be better to place the image after the paragraph describing ACL Policies and Tokens so that the reader has an understanding of what these components are before viewing the image.

karl-cardenas-coding:
@jkirschner-hashicorp Let's get this merged and released before we update the Learn guide.  That way when we link back to the docs it makes sense.

website/content/docs/security/acl/acl-system.mdx
jkirschner-hashicorp: @trujillo-adam: do you have a preference on whether images like this should generally be just before or just after the content (when the content is more than a few sentences long)?

I lean towards having the image first to provide an overview so that the details in the text below are easier to understand / more likely to stick, but this isn't a strong opinion.

website/content/docs/discovery/dns.mdx
jkirschner-hashicorp: I'll try this out and see what it looks like. Then we can decide between the options.

trujillo-adam: It's not clear how this information completes the idea in the lead-in statement ("Consul will authorize the request depending on the token's policy rules"). I would expect more specific guidance on how to configure the policy rules to enable authorization. It takes a lot of mental work to put align the lookup options with the goal of request authorization. An example configuration based on some kind of real-world scenario would help.

website/content/docs/discovery/dns.mdx
```diff
@@ -285,6 +285,50 @@ caching of DNS results. However, there are many situations in which caching is
...
+request to specify a token):
+
+- The agent's [`default` token](/docs/agent/options#acl_tokens_default)
+  token, if configured.
+- Otherwise, the built-in [`anonymous` token](/docs/security/acl/acl-system#builtin-tokens).
+  Because the anonymous token is used when any request is made to Consul without
+  explicitly specifying a token, production deployments should not apply policies
+  needed for DNS to this token.
+
+Consul will either authorize or revoke the request, depending on the token's
```
trujillo-adam, line 305:
```suggestion
Consul will authorize or deny the request depending on the token's
```

website/content/docs/security/acl/acl-system.mdx
trujillo-adam: I could go either way. It's really a matter of which learning style do we want to appeal to. Some users appreciate having some kind of spatial representation for attaching the concepts they are about to read--even if it's not immediately clear what they mean. This specific image is actually great for the upfront approach because it uses concrete labels/parameters (rather than some of the glossier branded boxes). Other users may want to try to understand the concept syntactically and use the diagram to reinforce the understanding. The point might also be moot because I think that users are going to read the entire section more than once before it starts to coalesce. That said, I with @jkirschner-hashicorp on leaving it where it is.

trujillo-adam (approved):
I added a few suggestions/questions

website/content/docs/discovery/dns.mdx
jkirschner-hashicorp: I don't see a great place for a separate page in the nav:
image

I'm fine with moving this content to the bottom of the page. I thought I recall us including use with ACLs at the bottom of another page, but I can't find it at the moment. I think that's our best option.

Long-term, I think it's a good idea to help users understand how the pieces fit together better (e.g., Learn tutorials that give guidance on how to accomplish the same thing with ACLs, mentioning use with ACLs in the docs, etc).

website/content/docs/discovery/dns.mdx
jkirschner-hashicorp: Not against this change, but trying to understand it. The previous wording felt more concise to me and clear from the first word, whereas here it's longer and I have to read to the last clause to understand there's a precedence here.

website/content/docs/discovery/dns.mdx
jkirschner-hashicorp: Why "authorize" instead of "either authorize or deny" (as Adam suggested)?

website/content/docs/discovery/dns.mdx
jkirschner-hashicorp: @trujillo-adam : I agree, I'll see if I can come up with a real-world scenario + configuration to help. Regardless, I think these details are needed (either before or after the example).

@blake : I tested out reformatting this information as a table. What do you think?

image

karl-cardenas-coding: @jkirschner-hashicorp what are the questions marks in the title for ? 
I do like the table. We can probably do away with the parentheses around the code blocks.

jkirschner-hashicorp: I originally didn't have the question marks, so the title was just `Requires service:read`. Having an answer "yes" in a cell without an explicit question being asked felt weird, so I tried out adding a question mark to the title. The question mark look funny though, and it seems clear enough in context without it. I'll remove it.

It feels weird that `*.query.consul` is listed in two rows. It's a special case of service lookup. **Q:** What do we think about changing "Service Lookups" to "Standard Service Lookups" and "Prepared Query Lookups" to "Prepared Query Service Lookups"?

blake: In order to be consistent with how ACL policy requirements are expressed throughout the rest of the site, it might be better to collapse the permission requirements into a single "ACL Required" column. An additional "Description" or similar column can be added to explain why the policies are needed.

For example

| Lookup type | ACLs required | Description |
| --- | --- | --- |
| Node lookups (`*.node.consul`) | `node:read` | Allow resolving the IP address for the target node |
| Service lookups (`*.service.consul`) | `service:read` <br> `node:read` | Allow resolving the IP addresses for the target service and node on which the service instances are running. |

jkirschner-hashicorp: I just pushed a commit where I tried this out (see below). I think it's easier to read, though I'm not sure the connection between the ACL rules and the behavior is as clear as the previous table option.

image

Perhaps that can be overcome with an example like @trujillo-adam was suggesting?

blake: IMO the updated table has a reduced cognitive load compared to the previous table. 👍  for adding a code example to show different policy combinations.

website/content/docs/discovery/dns.mdx
blake: After reading this a second time, it should probably say that Consul will accept or reject the request depending on whether the token has appropriate authorization.

Authorization technically takes place when the operator creates the policy and applies it to the token. Policy enforcement, which is what this is referring to, is where access is granted or rejected based on permissions granted by the policy.

website/content/docs/discovery/dns.mdx
blake: This second phrasing feels more explicit and is also a more complete sentence.

website/content/docs/discovery/dns.mdx
trujillo-adam: I think I see 4 columns:

| Lookup           | Type    | Description                                                                                                    | Required ACLs          |
|------------------|---------|----------------------------------------------------------------------------------------------------------------|------------------------|
| *.node.consul    | Node    | Allows IP addresses to be resolved for the target node                                                         | node:read              |
| *.service.consul | Service | Allows IP addresses to be resolved for the target service and node on which the service instances are running. | service:read node:read |
| *.query.consul   | Service | Allows IP addresses to be resolved for the target service and node on which the service instances are running. | service:read node:read |

I also think that each lookup should get its own row, even if there is repetition within the table. 
I didn't understand what the information was trying to express until I saw @Blake 's suggestion

jkirschner-hashicorp: @trujillo-adam : thanks for the suggestion - I think that version is even more readable.

One part I'm not sure I agree with is "each lookup should get its own row, even if there is repetition within the table". When there's lots of repetition, it's hard to spot differences. If we have a row for `*.service.consul`, `*.connect.consul`, `*.ingress.consul`, then `*.query.consul`, it will require a careful read to see that the first three are identical, but the last one has subtle differences.

Unfortunately, with the table component, there's no way to make a cell span multiple rows or columns.... which would allow us to have one shared cell across the 3 identical rows. Given that this isn't an option, I think keeping service, connect, and ingress style service lookups in the same row makes sense.

I'll try out what I'm thinking and then post the image.

jkirschner-hashicorp: Here's the latest based on all your feedback

image

trujillo-adam: That is a good looking table!

website/content/docs/discovery/dns.mdx
```diff
@@ -325,3 +325,55 @@ to the `default` namespace:
...
+
+- The agent's [`default` token](/docs/agent/options#acl_tokens_default)
+  token, if configured.
+- Otherwise, the built-in [`anonymous` token](/docs/security/acl/acl-system#builtin-tokens).
+  Because the anonymous token is used when any request is made to Consul without
+  explicitly specifying a token, production deployments should not apply policies
+  needed for DNS to this token.
+
+Consul will either authorize or deny the request depending on the token's
+policy rules:
```
trujillo-adam, line 345:
```suggestion
policy rules. The following table describes the supported DNS lookups and required policies when ACLs are enabled:
```
We just need one more statement describing the contents of the table.

trujillo-adam (approved):
Add the final table with a brief description of what's in it and I think this particular section is good to go. We should create a task to add an example later.

website/content/docs/discovery/dns.mdx
jkirschner-hashicorp: Changing to something that I think is clearer based on a conversation we had on the network segments tutorial about order of precedence for joining.

```
Consul agents resolve DNS requests using a preconfigured tokens, listed in
order of precedence:

1. The agent's [`default` token](/docs/agent/options#acl_tokens_default).
2. The built-in [`anonymous` token](/docs/security/acl/acl-system#builtin-tokens).
   Because the anonymous token is used when any request is made to Consul without
   explicitly specifying a token, production deployments should not apply policies
   needed for DNS to this token.
```

jkirschner-hashicorp:
@trujillo-adam : I made that last change. Here's the present state of the section I added. Anything else that's needed for approval?

image

karl-cardenas-coding (approved):
lgtm

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/443316.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 63b2cebcb33f9fdd360e1b881cfc305c63d947aa onto `stable-website` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 63b2cebcb33f9fdd360e1b881cfc305c63d947aa onto `release/1.10.x` succeeded!

PR #10838, opened 2021-08-12
grpc: ensure that streaming gRPC requests work over mesh gateway based wan federation

### LINKED ISSUE #10796: Consul 1.10.1 does not support WAN-fed-via-mesh-gateways without disabling streaming
#### Overview of the Issue

In Consul 1.10.1, streaming is enabled by default. Streaming is a performance improvement that streams updates over gRPC. Streaming does not support WAN federation over mesh gateways. You will see these errors in client logs:

```
2021-08-04T23:18:59.670Z [ERROR] agent.rpcclient.health: subscribe call failed: err="rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = "transport: Error while dialing dial tcp 10.128.2.16:8300: connect: no route to host"" topic=ServiceHealthConnect key=static-server failure_count=10
2021-08-04T23:18:59.670Z [WARN]  agent.viewstore: handling error in Store.Notify: error="rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = "transport: Error while dialing dial tcp 10.128.2.16:8300: connect: no route to host"" request-type=agent.rpcclient.health.serviceRequest index=0
2021-08-04T23:19:55.530Z [ERROR] agent.rpcclient.health: subscribe call failed: err="rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = "transport: Error while dialing dial tcp 10.128.2.16:8300: connect: no route to host"" topic=ServiceHealthConnect key=static-server failure_count=11
2021-08-04T23:19:55.530Z [WARN]  agent.viewstore: handling error in Store.Notify: error="rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = "transport: Error while dialing dial tcp 10.128.2.16:8300: connect: no route to host"" request-type=agent.rpcclient.health.serviceRequest index=0
2021-08-04T23:21:45.530Z [ERROR] agent.rpcclient.health: subscribe call failed: err="rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = "transport: Error while dialing dial tcp 10.128.2.16:8300: connect: no route to host"" topic=ServiceHealthConnect key=static-server failure_count=12
```

And mesh communication from one DC to another will not work.

As a workaround, you can disable streaming. On Kubernetes, this can be disabled via `extraConfig`:

```yaml
client:
  extraConfig: |
    {"use_streaming_backend": false}
```

On VMs, the config is:

```hcl
use_streaming_backend = false
```

(see https://www.consul.io/docs/agent/options#use_streaming_backend)

rboyer:
Fixes #10796

### DIFF
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -371,6 +371,9 @@ func (f fakeGRPCConnPool) ClientConnLeader() (*grpc.ClientConn, error) {
 	return nil, nil
 }
 
+func (f fakeGRPCConnPool) SetGatewayResolver(_ func(string) string) {
+}
+
 func TestAgent_ReconnectConfigWanDisabled(t *testing.T) {
 	if testing.Short() {
 		t.Skip("too slow for testing.Short")
@@ -4524,6 +4527,9 @@ LOOP:
 }
 
 // This is a mirror of a similar test in agent/consul/server_test.go
+//
+// TODO(rb): implement something similar to this as a full containerized test suite with proper
+// isolation so requests can't "cheat" and bypass the mesh gateways
 func TestAgent_JoinWAN_viaMeshGateway(t *testing.T) {
 	if testing.Short() {
 		t.Skip("too slow for testing.Short")
@@ -4771,6 +4777,9 @@ func TestAgent_JoinWAN_viaMeshGateway(t *testing.T) {
 	})
 
 	// Ensure we can do some trivial RPC in all directions.
+	//
+	// NOTE: we explicitly make streaming and non-streaming assertions here to
+	// verify both rpc and grpc codepaths.
 	agents := map[string]*TestAgent{"dc1": a1, "dc2": a2, "dc3": a3}
 	names := map[string]string{"dc1": "bob", "dc2": "betty", "dc3": "bonnie"}
 	for _, srcDC := range []string{"dc1", "dc2", "dc3"} {
@@ -4780,20 +4789,39 @@ func TestAgent_JoinWAN_viaMeshGateway(t *testing.T) {
 				continue
 			}
 			t.Run(srcDC+" to "+dstDC, func(t *testing.T) {
-				req, err := http.NewRequest("GET", "/v1/catalog/nodes?dc="+dstDC, nil)
-				require.NoError(t, err)
+				t.Run("normal-rpc", func(t *testing.T) {
+					req, err := http.NewRequest("GET", "/v1/catalog/nodes?dc="+dstDC, nil)
+					require.NoError(t, err)
 
-				resp := httptest.NewRecorder()
-				obj, err := a.srv.CatalogNodes(resp, req)
-				require.NoError(t, err)
-				require.NotNil(t, obj)
+					resp := httptest.NewRecorder()
+					obj, err := a.srv.CatalogNodes(resp, req)
+					require.NoError(t, err)
+					require.NotNil(t, obj)
 
-				nodes, ok := obj.(structs.Nodes)
-				require.True(t, ok)
-				require.Len(t, nodes, 1)
-				node := nodes[0]
-				require.Equal(t, dstDC, node.Datacenter)
-				require.Equal(t, names[dstDC], node.Node)
+					nodes, ok := obj.(structs.Nodes)
+					require.True(t, ok)
+					require.Len(t, nodes, 1)
+					node := nodes[0]
+					require.Equal(t, dstDC, node.Datacenter)
+					require.Equal(t, names[dstDC], node.Node)
+				})
+				t.Run("streaming-grpc", func(t *testing.T) {
+					req, err := http.NewRequest("GET", "/v1/health/service/consul?cached&dc="+dstDC, nil)
+					require.NoError(t, err)
+
+					resp := httptest.NewRecorder()
+					obj, err := a.srv.HealthServiceNodes(resp, req)
+					require.NoError(t, err)
+					require.NotNil(t, obj)
+
+					csns, ok := obj.(structs.CheckServiceNodes)
+					require.True(t, ok)
+					require.Len(t, csns, 1)
+
+					csn := csns[0]
+					require.Equal(t, dstDC, csn.Node.Datacenter)
+					require.Equal(t, names[dstDC], csn.Node.Node)
+				})
 			})
 		}
 	}
--- a/agent/consul/client_test.go
+++ b/agent/consul/client_test.go
@@ -525,15 +525,13 @@ func newDefaultDeps(t *testing.T, c *Config) Deps {
 		Tokens:          new(token.Store),
 		Router:          r,
 		ConnPool:        connPool,
-		GRPCConnPool: grpc.NewClientConnPool(
-			builder,
-			nil,
-			grpc.TLSWrapper(tls.OutgoingRPCWrapper()),
-			nil,
-			tls.UseTLS,
-			true,
-			c.Datacenter,
-		),
+		GRPCConnPool: grpc.NewClientConnPool(grpc.ClientConnPoolConfig{
+			Servers:               builder,
+			TLSWrapper:            grpc.TLSWrapper(tls.OutgoingRPCWrapper()),
+			UseTLSForDC:           tls.UseTLS,
+			DialingFromServer:     true,
+			DialingFromDatacenter: c.Datacenter,
+		}),
 		LeaderForwarder: builder,
 		EnterpriseDeps:  newDefaultDepsEnterprise(t, logger, c),
 	}
--- a/agent/consul/client_test.go
+++ b/agent/consul/client_test.go
@@ -492,11 +492,11 @@ func newClient(t *testing.T, config *Config) *Client {
 	return client
 }
 
-func newTestResolverConfig(t *testing.T) resolver.Config {
+func newTestResolverConfig(t *testing.T, suffix string) resolver.Config {
 	n := t.Name()
 	s := strings.Replace(n, "/", "", -1)
 	s = strings.Replace(s, "_", "", -1)
-	return resolver.Config{Authority: strings.ToLower(s)}
+	return resolver.Config{Authority: strings.ToLower(s) + "-" + suffix}
 }
 
 // TODO(rb): add tests for the wanfed/alpn variations
@@ -512,7 +512,7 @@ func newDefaultDeps(t *testing.T, c *Config) Deps {
 	tls, err := tlsutil.NewConfigurator(c.TLSConfig, logger)
 	require.NoError(t, err, "failed to create tls configuration")
 
-	builder := resolver.NewServerResolverBuilder(newTestResolverConfig(t))
+	builder := resolver.NewServerResolverBuilder(newTestResolverConfig(t, c.NodeName+"-"+c.Datacenter))
 	r := router.NewRouter(logger, c.Datacenter, fmt.Sprintf("%s.%s", c.NodeName, c.Datacenter), builder)
 	resolver.Register(builder)
 
--- a/agent/consul/client_test.go
+++ b/agent/consul/client_test.go
@@ -9,6 +9,12 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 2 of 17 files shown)


agent/agent_test.go
```diff
@@ -4524,6 +4527,9 @@ LOOP:
 }
 
 // This is a mirror of a similar test in agent/consul/server_test.go
+//
+// TODO(rb): implement something similar to this as a full containerized test suite with proper
```
rboyer, line 4531: Calling out this TODO, since I could not find a way to do the "firewalling" of the servers from each other in-process.

agent/consul/options.go
```diff
@@ -24,8 +24,10 @@ type Deps struct {
 type GRPCClientConner interface {
 	ClientConn(datacenter string) (*grpc.ClientConn, error)
 	ClientConnLeader() (*grpc.ClientConn, error)
+	SetGatewayResolver(func(string) string)
```
rboyer, line 27: This grpc client pool is created in `agent/setup.go` universally for servers and clients, but only servers will eventually create a GatewayResolver for themselves in `agent/consul/server.go`.  For the regular connection pool for RPCs the gw resolver field is set after the struct was created (but before it's used). I did the same here to avoid having to heavily refactor the agent construction code.

agent/consul/rpc.go
```diff
@@ -293,7 +293,7 @@ func (s *Server) handleNativeTLS(conn net.Conn) {
 		s.handleSnapshotConn(tlsConn)
 
 	case pool.ALPN_RPCGRPC:
-		s.grpcHandler.Handle(conn)
+		s.grpcHandler.Handle(tlsConn)
```
rboyer, line 296: This was an existing bug likely due to a simple copypaste gone awry. All of the other branches take a `tlsConn` instead of a `conn` to be on the correct side of the multiplex envelope.

agent/consul/server.go
```diff
@@ -390,6 +394,7 @@ func NewServer(config *Config, flat Deps) (*Server, error) {
 			s.config.PrimaryDatacenter,
 		)
 		s.connPool.GatewayResolver = s.gatewayLocator.PickGateway
+		s.grpcConnPool.SetGatewayResolver(s.gatewayLocator.PickGateway)
```
rboyer, line 397: Here's the late-set mentioned earlier.

agent/consul/wanfed/wanfed.go
```diff
@@ -136,42 +131,24 @@ func (t *Transport) DialAddressTimeout(addr memberlist.Address, timeout time.Dur
...
-		_ = tcp.SetNoDelay(true)
-	}
-
-	tlsConn, err := wrapper(dc, nodeName, nextProto, rawConn)
-	if err != nil {
-		return nil, err
-	}
-
-	return tlsConn, nil
+func (t *Transport) dial(dc, nodeName, nextProto string) (net.Conn, error) {
```
rboyer, line 140: Instead of ending up with 3 similar copies of the same function structure I refactored stuff so that now there's only one implementation of this dialing approach shared by all 3 locations.

agent/grpc/resolver/resolver.go
```diff
@@ -34,12 +35,21 @@ type Config struct {
 	Authority string
 }
 
-func NewServerResolverBuilder(cfg Config) *ServerResolverBuilder {
+func NewServerResolverBuilder(cfg Config) (*ServerResolverBuilder, error) {
+	if cfg.Authority == "" {
```
rboyer, line 39: At some point of my adventure trying to get `agent.TestAgent_JoinWAN_viaMeshGateway` to pass (in an earlier draft where there were changes) I suspected grpc resolver global cross-talk and arrived at this solution which would prevent issues across any parallel test case execution (or multiple agents within a single test case) without forcing the user to do anything.

agent/pool/conn.go
```diff
@@ -28,19 +30,19 @@ func (t RPCType) ALPNString() string {
 const (
 	// keep numbers unique.
 	RPCConsul      RPCType = 0
-	RPCRaft                = 1
-	RPCMultiplex           = 2 // Old Muxado byte, no longer supported.
-	RPCTLS                 = 3
-	RPCMultiplexV2         = 4
-	RPCSnapshot            = 5
-	RPCGossip              = 6
+	RPCRaft        RPCType = 1
```
rboyer, line 33: Since an `iota` isn't used here, only `RPCConsul` had the type `RPCType`. Something I did in this PR surfaced this issue so I just fixed it in place.

agent/pool/pool.go
```diff
@@ -372,62 +377,54 @@ func (p *ConnPool) dial(
...
-
 	gwAddr := gatewayResolver(dc)
 	if gwAddr == "" {
 		return nil, nil, structs.ErrDCNotAvailable
 	}
 
-	dialer := &net.Dialer{LocalAddr: src, Timeout: defaultDialTimeout}
+	var dialTimeout time.Duration
+	if nextProto != ALPN_RPCGRPC && dialTimeout == 0 {
+		// TODO(rb): do we want this?
```
rboyer, line 408: Calling out this question for @dnephin

agent/pool/pool.go
```diff
@@ -372,62 +377,54 @@ func (p *ConnPool) dial(
...
 
 	rawConn, err := dialer.Dial("tcp", gwAddr)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	if tcp, ok := rawConn.(*net.TCPConn); ok {
-		_ = tcp.SetKeepAlive(true)
-		_ = tcp.SetNoDelay(true)
+	if nextProto != ALPN_RPCGRPC {
```
rboyer, line 418: I thought I saw a reference in the grpc setup code to keepalives being controlled elsewhere, so since this wasn't being done in the existing grpc pool code I opted to keep this section off for grpc.

agent/pool/pool.go
```diff
@@ -372,62 +377,54 @@ func (p *ConnPool) dial(
...
-	}
-
 	gwAddr := gatewayResolver(dc)
 	if gwAddr == "" {
 		return nil, nil, structs.ErrDCNotAvailable
 	}
 
-	dialer := &net.Dialer{LocalAddr: src, Timeout: defaultDialTimeout}
+	var dialTimeout time.Duration
+	if nextProto != ALPN_RPCGRPC && dialTimeout == 0 {
```
freddygv, line 407:
```suggestion
	if nextProto != ALPN_RPCGRPC {
```
Won't dialTimeout always be 0?

agent/grpc/client.go
```diff
@@ -102,17 +145,46 @@ func (c *ClientConnPool) dial(datacenter string, serverType string) (*grpc.Clien
...
+				srcAddr,
+				tlsutil.ALPNWrapper(alpnWrapper),
+				pool.ALPN_RPCGRPC,
+				dialingFromServer,
+				gwResolverDep.GatewayResolver,
+			)
+			return conn, err
+		}
+
+		// TODO(rb): should this get the standard rpc dial timeout of 10 seconds?
```
freddygv, line 184: Probably? it doesn't seem like we're using timeouts at all for gRPC since we're not passing `WithTimeout` into the `grpc.Dial` nor are we using `DialContext` and passing a context with a deadline:
https://pkg.go.dev/google.golang.org/grpc#WithTimeout

agent/pool/pool.go
freddygv: Same as above, maybe shouldn't be excluded for gRPC

agent/grpc/resolver/resolver.go
rboyer: It's doable, but it would require exposing a test-only setting on `config.RuntimeConfig` to be able to pass something down into `agent/setup.go` where `NewServerResolverBuilder` is actually ever called in the common case. 

I guess alternatively maybe i could make `agent/setup.go` generate an Authority from `<datacenter>.<nodeName>` which even in a multi-agent unit test would be sufficiently unique?

rboyer: I've adjusted this to not use UUIDs, but in the common case in `agent/setup.go` it just does `<datacenter>.<nodeID>` and uses that for the authority which is pretty cheap and doesn't require punching a test-only flag down through RuntimeCOnfig

agent/grpc/client.go
```diff
@@ -102,22 +157,39 @@ func (c *ClientConnPool) dial(datacenter string, serverType string) (*grpc.Clien
...
+				cfg.SrcAddr,
+				tlsutil.ALPNWrapper(cfg.ALPNWrapper),
+				pool.ALPN_RPCGRPC,
+				cfg.DialingFromServer,
+				gwResolverDep.GatewayResolver,
+			)
+			return conn, err
+		}
+
+		d := net.Dialer{LocalAddr: cfg.SrcAddr, Timeout: pool.DefaultDialTimeout}
```
rboyer, line 185: Made the non-gateway version of grpc also have a non-infinite dial timeout like all of the others.

agent/pool/pool.go
```diff
@@ -11,14 +12,15 @@ import (
...
+
 	"github.com/hashicorp/consul/agent/structs"
 	"github.com/hashicorp/consul/lib"
 	"github.com/hashicorp/consul/tlsutil"
-	msgpackrpc "github.com/hashicorp/net-rpc-msgpackrpc"
-	"github.com/hashicorp/yamux"
 )
 
-const defaultDialTimeout = 10 * time.Second
+const DefaultDialTimeout = 10 * time.Second
```
rboyer, line 23: Exported this so it could be used from the `agent/grpc` package as well.

agent/pool/pool.go
```diff
@@ -372,62 +377,49 @@ func (p *ConnPool) dial(
...
-		return nil, nil, fmt.Errorf("rpc type %d cannot be routed through a mesh gateway", actualRPCType)
-	}
-
 	gwAddr := gatewayResolver(dc)
 	if gwAddr == "" {
 		return nil, nil, structs.ErrDCNotAvailable
 	}
 
-	dialer := &net.Dialer{LocalAddr: src, Timeout: defaultDialTimeout}
+	dialer := &net.Dialer{LocalAddr: srcAddr, Timeout: DefaultDialTimeout}
```
rboyer, line 406: All rpc types get a 10s timeout now.

agent/consul/client_test.go
```diff
@@ -490,6 +492,14 @@ func newClient(t *testing.T, config *Config) *Client {
...
 }
 
+func newTestResolverConfig(t *testing.T, suffix string) resolver.Config {
+	n := t.Name()
+	s := strings.Replace(n, "/", "", -1)
+	s = strings.Replace(s, "_", "", -1)
+	return resolver.Config{Authority: strings.ToLower(s) + "-" + suffix}
+}
+
+// TODO(rb): add tests for the wanfed/alpn variations
```
freddygv, line 502: Did you want to do that in this PR?

freddygv (approved):
LGTM, thanks!

agent/consul/client_test.go
rboyer: Actually in this specific spot it doesn't make sense to test RPC via ALPN because as it all exists today that's only executed for Server-to-Server and this file is testing Client-to-Server which currently does not do ALPN.

rboyer:
@freddygv i added a test in `agent/grpc` that covers having the grpc dialing code correctly dial a server via a gateway using the ALPN approach

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/432909.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 5b6d96d27d80dbff40073e9775f2a2c187cc90c4 onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/432909)

PR #10813, opened 2021-08-09
deps: upgrade gogo-protobuf to v1.3.2

mikemorris:
Bump to the latest release of gogo/protobuf.

Longer term plan is still removing this dep and switching back to mainline google/protobuf as described in https://github.com/hashicorp/consul/issues/10471, but we should at least move to the latest release before doing that.

PR #10839, opened 2021-08-12
backport(1.10): deps: upgrade gogo-protobuf to v1.3.2

mikemorris:
Backports #10813, only conflict in a generated `.pb.go` file from adding `Unmarshal` function, regenerated with `make proto --always-make`


mikemorris:
Why would the changelog checker be failing here???

dnephin:
> Why would the changelog checker be failing here???

It's because the changelog checker is looking for a changelog with this PR number (10839), but the changelog file has a different filename.  I think it might make sense to only run the changelog check on PRs to the `main` branch. We expect the check to fail on all backports because the PR number won't match the filename.

dnephin (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/427035.

PR #10850, opened 2021-08-13
acl: Remove legacy bootstrap and legacy translate rules API endpoints

dnephin:
Branched from #10661

### DIFF
--- a/agent/acl_endpoint.go
+++ b/agent/acl_endpoint.go
@@ -105,55 +105,6 @@ func (s *HTTPHandlers) ACLRulesTranslate(resp http.ResponseWriter, req *http.Req
 	return nil, nil
 }
 
-func (s *HTTPHandlers) ACLRulesTranslateLegacyToken(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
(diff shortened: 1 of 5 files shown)


agent/acl_endpoint.go
```diff
@@ -40,40 +39,18 @@ func (s *HTTPHandlers) ACLBootstrap(resp http.ResponseWriter, req *http.Request)
 	args := structs.DCSpecificRequest{
 		Datacenter: s.agent.config.Datacenter,
 	}
-
-	legacy := false
-	legacyStr := req.URL.Query().Get("legacy")
```
rboyer, line 45: Should also update the website to indicate that it stopped working in 1.11.

rboyer (approved):
LGTM

agent/acl_endpoint.go
dnephin: Ya, good call. I looked at https://www.consul.io/api-docs/acl#bootstrap-acls, and didn't see this query parameter documented. Maybe it's hidden somewhere else?  I plan to do another full pass of all of these at the end.

PR #9081, opened 2020-11-02
ui: Update to not return metrics for ingress gateways

kaxcode:
This PR updates conditionals to not return metrics if service is an ingress-gateway.
Screen Shot 2020-11-02 at 11 48 42 AM

PR #10858, opened 2021-08-16
ui: Hide all metrics for ingress gateway services

### LINKED ISSUE #10745: Topology view shows no metrics for ingress gateways on initial load
#### Overview of the Issue

When initially loading the ingress gateway topology view it says "Unable to load metrics"

CleanShot 2021-07-31 at 22 49 06@2x

However if then you navigate to another service it loads the metrics for the ingress gateway and displays the metrics under its card:

CleanShot 2021-07-31 at 22 50 03@2x

And then when you navigate back to the ingress gateway then it shows the graph:

CleanShot 2021-07-31 at 22 51 17@2x

I don't see any of the `/query` API calls on initial load of the ingress gateway page like I do when I navigate back to it so hopefully that's a hint as to the underlying cause.

* Consul 1.10.0
* no ACLs
* prometheus metrics enabled for ingress gateway

### Log Fragments

No error logs in Consul server logs or JS console.

kaxcode:
### 🐛 Description:
Currently, our backend does not support metrics for Ingress Gateways. Our initial solution ([#9081](https://github.com/hashicorp/consul/pull/9081)) to this case was to hide all the metric stats in the Topology view for an ingress-gateway.

Bugs: 
- We did not hide the metrics graph for when an ingress-gateway is the main service. Users have been getting confused with the 'Unable to load metrics' messaging. This message makes it seem the metrics configurations are not set up properly. 
- Metric stats are appearing for downstream/upstream services that are ingress-gateway. Ingress gateway metrics should not be returning at all because they are not supported.

This PR fixes #10745.

[Demo Link]

### 🎉 Solution:
To avoid further confusion we are hiding the following:
- Metrics graph if the main service is an ingress-gateway
- Metrics stats if a downstream/upstream service is an ingress-gateway

### 📸 Screenshots:
Before:
Screen Shot 2021-08-16 at 5 32 20 PM

After:
- Hiding metrics graph if the main service is an ingress-gateway
Screen Shot 2021-08-16 at 5 28 22 PM

- Hiding metrics stats if a downstream/upstream service is an ingress-gateway
Screen Shot 2021-08-16 at 5 27 34 PM

### 🤡 Updates to mock-api:
- We updated the service-topology endpoint to include the `Kind` attribute in Upstream and Downstream services.

### 🧪 Testing:
Checks were added to test the following:
- All the metrics stats and the metrics graph are hidden when Ingress Gateway is the main service.
- Metrics stats are hidden if a downstream is an Ingress Gateway. Upstream service not covered but it would be the same.

### 📝 **PR Tasks:**

- [ ]  Backport
- [ ]  Changelog
- [ ]  Demo Link


lkysow:
What if it said `Viewing ingress gateway metrics is not currently supported` (or something)? That way folks wouldn't be trying to get it working.

kaxcode:
> What if it said `Viewing ingress gateway metrics is not currently supported` (or something)? That way folks wouldn't be trying to get it working.

@lkysow good point. We don't want to keep confusing users, but guiding them on what's happening would be useful 😄 I'll bring this up to the designer. Get their thoughts on the wording.

ui/packages/consul-ui/app/components/topology-metrics/status/index.hbs
```diff
@@ -1,12 +1,14 @@
-{{#if @noMetricsReason}}
+{{#if (eq @kind 'ingress-gateway')}}
+  <span class="topology-metrics-error">{{t "components.consul.topology-metrics.status.ingress-gateway"}}</span>
+{{else if @noMetricsReason}}
   <span class="topology-metrics-error">
-    Unable to load metrics
+    {{t "components.consul.topology-metrics.status.error"}}
     <span>
       <Tooltip>{{@noMetricsReason}}</Tooltip>
```
johncowen, line 7: Could we update this deprecated tooltip component to use the modifier one whilst we in this file? Ideally if you are editing a file and you see these would should be upgrading them as we go (happy for that to be here or in another PR).

ui/packages/consul-ui/translations/components/consul/en-us.yaml
```diff
@@ -156,9 +156,12 @@ topology-metrics:
       header: No traffic
       body: Add "{upstream}" as an explicit upstream of "{downstream}" or set the "{downstream}" proxy to "transparent" mode to enable traffic.
       action: Documentation
+  status:
+    error: Unable to load metrics
+    loading: Loading Metrics
+    ingress-gateway: Viewing metrics for ingress gateways is not currently supported in the UI.
```
johncowen, line 162: We use a mixture of title and capital case here:

```
Unable to load metrics
Loading Metrics
Viewing metrics
```

I'm never sure whether we have a rule to stick to with these, could you check in with design and see whether it should be `Metrics` or `metrics`, if it was me I'd stick to `metrics` unless at the start of a phrase/sentence (I think there might sometimes be an exception for where we capitalize 'Consul Nouns', which makes me wonder where we should use `Ingress Gateways` or `ingress gateways` also.

Ideally we'd have a written rule to follow so we can refer to it if we are unsure.

@trujillo-adam I wondered what your opinion on this was? If you can describe a pattern we should follow I can maybe add it to our Eng Docs somewhere so it's written down.

johncowen (commented):
Hey 👋 

I took a very quick scan of this, and left a couple of initial things, but I noticed this in the code previous to this PR/change and wondered what it did?

https://github.com/hashicorp/consul/blob/39b3c3e2bddc519ac1b79ca471967afd0c6d1641/ui/packages/consul-ui/app/components/topology-metrics/index.js#L102-L103

https://github.com/hashicorp/consul/blob/39b3c3e2bddc519ac1b79ca471967afd0c6d1641/ui/packages/consul-ui/app/components/topology-metrics/status/index.hbs#L1-L7

Both of those ^ are from before this PR 

If you could let me know and I can give this a proper look over 👍

kaxcode:
> Hey 👋
> 
> I took a very quick scan of this, and left a couple of initial things, but I noticed this in the code previous to this PR/change and wondered what it did?
> 
> https://github.com/hashicorp/consul/blob/39b3c3e2bddc519ac1b79ca471967afd0c6d1641/ui/packages/consul-ui/app/components/topology-metrics/index.js#L102-L103
> 
> https://github.com/hashicorp/consul/blob/39b3c3e2bddc519ac1b79ca471967afd0c6d1641/ui/packages/consul-ui/app/components/topology-metrics/status/index.hbs#L1-L7
> 
> Both of those ^ are from before this PR
> 
> If you could let me know and I can give this a proper look over 👍

@johncowen ah, totally forgot we mention something for ingress-gateways in tooltip. I'll verify with designer if we want to update the wording for tooltip or just show the reason up-front.

ui/packages/consul-ui/translations/components/consul/en-us.yaml
kaxcode: @johncowen I'll update to `metrics` and `Ingress Gateway`. 

If you want to have a full-fledged discussion about capitalization patterns please do so outside out this PR.

ui/packages/consul-ui/app/components/topology-metrics/status/index.hbs
kaxcode: I tried ```{{tooltip @noMetricsReason}}``` and it did not work. No changes made.

ui/packages/consul-ui/app/components/topology-metrics/index.js
```diff
@@ -99,8 +105,6 @@ export default class TopologyMetrics extends Component {
   calculate() {
     if (this.args.isRemoteDC) {
       this.noMetricsReason = 'Unable to fetch metrics for a remote datacenter';
-    } else if (this.args.service.Service.Kind === 'ingress-gateway') {
```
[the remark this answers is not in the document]
kaxcode: Updated this to be used for conditional messages instead of passing down `@kind`

ui/packages/consul-ui/app/components/topology-metrics/index.hbs
```diff
@@ -25,8 +25,9 @@
       @hasMetricsProvider={{@hasMetricsProvider}}
       @noMetricsReason={{this.noMetricsReason}}
     >
-    {{#if (and @hasMetricsProvider (not-eq @service.Service.Kind 'ingress-gateway'))}}
+    {{#if (and @hasMetricsProvider this.mainNotIngressService (not-eq item.Kind 'ingress-gateway'))}}
       <TopologyMetrics::Stats
+        data-test-downstream-metrics-stats
         @nspace={{or item.Namespace 'default'}}
```
johncowen, line 31: Could you add a comment here somewhere explaining that this is one of the only places in the UI where is ok to default to the `default` namespace? I'm pretty sure that's a correct statement?

ui/packages/consul-ui/app/components/topology-metrics/index.js
```diff
@@ -84,6 +84,12 @@ export default class TopologyMetrics extends Component {
     return items;
   }
 
+  get mainNotIngressService() {
+    const kind = get(this.args.service.Service, 'Kind') || '';
```
johncowen, line 88: Is `Kind` ever falsey here?

ui/packages/consul-ui/app/components/topology-metrics/status/index.hbs
```diff
@@ -1,12 +1,16 @@
...
+      <Tooltip>{{t "components.consul.topology-metrics.status.remote-dc"}}</Tooltip>
     </span>
   </span>
 {{else if @error}}
-  <span class="topology-metrics-error">Unable to load metrics</span>
+  <span class="topology-metrics-error">{{t "components.consul.topology-metrics.status.error"}}</span>
 {{else}}
-  <span class="topology-metrics-loader">Loading Metrics</span>
+  <span class="topology-metrics-loader">{{t "components.consul.topology-metrics.status.loading"}}</span>
 {{/if}}
```
johncowen, line 16: This area feels like the conditional could be formatted better somehow? The reason I noticed this is you have the `<span class="topology-metrics-error">` repeated multiple times.

ui/packages/consul-ui/tests/acceptance/dc/services/show/topology/stats.feature
```diff
@@ -33,4 +33,50 @@ Feature: dc / services / show / topology / stats
...
+        ID: ingress-gateway-with-id
+    ---
+    When I visit the service page for yaml
+    ---
+      dc: dc1
+      service: ingress-gateway
+    ---
+    And I see topology on the tabs
+    And I don't see the "[data-test-topology-metrics-stats]" element
+    And I see the "[data-test-sparkline]" element
```
johncowen, line 54:
```gherkin
Given metrics is enabled, all metrics are disabled for an ingress gateway Topology
Then I see the "[data-test-sparkline]" element
```
This reads like the metrics/sparkline is still visible? Whereas, from what I understand, it should be hidden here?

ui/packages/consul-ui/tests/acceptance/dc/services/show/topology/stats.feature
```diff
@@ -33,4 +33,50 @@ Feature: dc / services / show / topology / stats
...
+    - Service:
+        Name: service-0
+        ID: service-0-with-id
+    ---
+    And 1 topology model from yaml
+    ---
+      Upstreams: []
+      Downstreams:
+        - Name: db
+          Namespace: default
```
johncowen, line 70:
```suggestion
          Namespace: @namespace
```
That ^ should make it replace `@namespace` with whatever namespace is currently being tested (enterprise tests will run this test for an empty namespace, the `default` namespace, a `team-1` namespace and an `undefined` namespace.

johncowen (commented):
I took another look at this and left some questions/suggestions. I've not given it a test myself yet though, I'll try and do that today. I thought I'd leave this review incase I don't get around to that today.

ui/packages/consul-ui/translations/components/consul/en-us.yaml
johncowen: No problem!

https://github.com/hashicorp/consul/pull/10888

@trujillo-adam if you have anything to add here, we should probably use that ^ PR, thanks!

ui/packages/consul-ui/app/components/topology-metrics/index.hbs
kaxcode: I've added the comment.

ui/packages/consul-ui/app/components/topology-metrics/index.js
kaxcode: no, Kind doesn't return falsey. It simply doesn't return `Kind` making this line undefined. Should I remove the ```|| ''``` ?

ui/packages/consul-ui/app/components/topology-metrics/status/index.hbs
kaxcode: Refactored

ui/packages/consul-ui/tests/acceptance/dc/services/show/topology/stats.feature
kaxcode: I've updated the scenario.

ui/packages/consul-ui/tests/acceptance/dc/services/show/topology/stats.feature
kaxcode: I've updated this.

ui/packages/consul-ui/app/components/topology-metrics/status/index.hbs
```diff
@@ -1,12 +1,16 @@
...
-      <Tooltip>{{@noMetricsReason}}</Tooltip>
-    </span>
+    {{#if (eq @noMetricsReason 'ingress-gateway')}}
+      {{t "components.consul.topology-metrics.status.ingress-gateway"}}
+    {{else if @error}}
+      {{t "components.consul.topology-metrics.status.error"}}
+    {{else if (eq @noMetricsReason 'remote-dc')}}
+      {{t "components.consul.topology-metrics.status.error"}}
+      <span
+        {{tooltip (t "components.consul.topology-metrics.status.remote-dc")}}
```
kaxcode, line 10: I figured out what was wrong with this tooltip. It needed to be tucked inside the `span`.

ui/packages/consul-ui/mock-api/v1/internal/ui/service-topology/_
```diff
@@ -98,6 +99,7 @@ ${
         const allowed = hasPerms ? false : fake.random.boolean();
         return `
     {
+      "Kind": "${fake.helpers.randomize(['mesh-gateway', 'ingress-gateway', 'terminating-gateway', ''])}",
```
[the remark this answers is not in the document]
kaxcode: I've added `undefined`

ui/packages/consul-ui/tests/acceptance/dc/services/show/topology/stats.feature
```diff
@@ -33,4 +33,50 @@ Feature: dc / services / show / topology / stats
...
+    ---
+    And 1 topology model from yaml
+    ---
+      Upstreams: []
+      Downstreams:
+        - Name: db
+          Namespace: @namespace
+          Datacenter: dc1
+          Intention: {}
+    ---
```
johncowen, line 73: `Given metrics is enabled, metric stats are disabled for ingress gateway as downstream services`

I'm not sure if I can see here that we are setting up the test to test an ingress-gateway? I ran the test and I see this:

Screenshot 2021-08-25 at 11 40 17

Is `service-0` or `db` the ingress-gateway here?

I also noticed the "No metrics available" message here which is different to the screengrabs in your description? I'm not sure whether its supposed to show this message or not? I think that if db is the ingress, then judging from the screengrabs in the description I don't think this message is supposed to show? Not sure.

ui/packages/consul-ui/mock-api/v1/internal/ui/service-topology/_
```diff
@@ -98,6 +99,7 @@ ${
         const allowed = hasPerms ? false : fake.random.boolean();
         return `
     {
+      "Kind": "${fake.helpers.randomize(['mesh-gateway', 'ingress-gateway', 'terminating-gateway', '', undefined])}",
```
johncowen, line 102: Could we use this form instead here, it makes it more or a truer mock of the actual API, I'm pretty sure the API never returns actual `undefined` as a value.

```javascript
${fake.booleanOfSomeSort() ? `
  "Kind": "${fake.helpers.randomize(['mesh-gateway', 'ingress-gateway', 'terminating-gateway', ''])}",
` : ``}
```

(fake.booleanOfSomeSort() doesn't exist, but maybe look up what fits best here faker has a few ways to do this kind of thing, i.e. do you want it to be 50/50 of the time or 10/90 etc)

ui/packages/consul-ui/app/components/topology-metrics/status/index.hbs
```diff
@@ -1,12 +1,16 @@
...
+      />
+    {{else if @error}}
+      {{t "components.consul.topology-metrics.status.error"}}
+    {{/if}}
   </span>
-{{else if @error}}
-  <span class="topology-metrics-error">Unable to load metrics</span>
 {{else}}
-  <span class="topology-metrics-loader">Loading Metrics</span>
+  <span class="topology-metrics-status-loader">{{t "components.consul.topology-metrics.status.loading"}}</span>
```
johncowen, line 15: Thanks for changing this to follow our component-name/class-name pattern! 🎉

johncowen (commented):
I ran through this a little more and had a few more comments, but I think mainly I think this is a good change from a UX/design perspective, but I'm not sure if it fixes that issue reported by @lkysow in #10745.

From what I understand the the report, the sparkline metrics are initially hidden on first visit, and then become visible when you go elsewhere and come back. The only logic type thing I can see here is that we've added a check to make sure the `streams aren't ingress. Would this change affect the behaviour in the report? I dunno maybe I read this wrong. Did you manage to reproduce what Luke reported?

This is slightly related to another comment I had, and I think its where my initial confusion with the testing here comes in. Are we currently testing that the central 'main' sparkline graph doesn't show when the 'main' service is an ingress gateway?

ui/packages/consul-ui/tests/acceptance/dc/services/show/topology/stats.feature
kaxcode: `db` is the ingress-gateway here. I've added `Kind: ingress-gateway` to it. A typo didn't catch this. I've update the data-test to say `data-test-topology-metrics-downstream-stats`.

ui/packages/consul-ui/mock-api/v1/internal/ui/service-topology/_
kaxcode: Update to show `Kind` 70% of the time.

kaxcode:
> From what I understand the the report, the sparkline metrics are initially hidden on first visit, and then become visible when you go elsewhere and come back. The only logic type thing I can see here is that we've added a check to make sure the `streams aren't ingress. Would this change affect the behaviour in the report? I dunno maybe I read this wrong. Did you manage to reproduce what Luke reported?
> 

@johncowen almost missed this. yes, I've tested this. When I go back to the ingress gateway service the sparkline metrics are not shown, but the message is there within the sparkline wrapper.

> This is slightly related to another comment I had, and I think its where my initial confusion with the testing here comes in. Are we currently testing that the central 'main' sparkline graph doesn't show when the 'main' service is an ingress gateway?

No, I am currently testing that the sparkline wrapper shows up for a main service that is an ingress-gateway, because it carries the status message. I'll update the test to check the 'status' message shows up instead of just testing the sparkline-wrapper.

PR #10866, opened 2021-08-17
docs/nia: Add regex option to services condition

mkam:
- [Service condition configuration changes](https://consul-okp07abf1-hashicorp.vercel.app/docs/nia/configuration#services-condition)
- [Task Execution: Services Condition changes](https://consul-okp07abf1-hashicorp.vercel.app/docs/nia/tasks#services-condition)

### DIFF
--- a/website/content/docs/nia/configuration.mdx
+++ b/website/content/docs/nia/configuration.mdx
@@ -154,24 +154,23 @@ A `task` block can be optionally configured with a `condition` block to set the 
 
 #### Services Condition
 
-This is the default type of condition when no `condition` block is configured for a task. This condition will trigger the task on changes in services configured in [`task.services`](#services). The task can optionally be configured with a services condition block. However, this is not necessary since this is the default type of condition and there are currently no additional configuration options available for services condition.
+This is the default type of condition when no `condition` block is configured for a task. This condition will trigger the task on either changes in services configured in [`task.services`](#services) or changes in services that match the regular expression configured in `regexp`. The default is triggering on changes to [`task.services`](#services), and [`task.services`](#services) is required if `regexp` is not configured.
 
 See [Task Execution: Services Condition](/docs/nia/tasks#services-condition) for more details on how tasks are triggered with a services condition.
 
 ```hcl
 task {
   name = "services_condition_task"
-  description = "execute on changes to api, db, and web services"
-  services = ["api", "db", "web"]
+  description = "execute on changes to services whose name starts with web"
   providers = ["my-provider"]
   source = "path/to/services-condition-module"
 
-  // optional: declare empty services condition block
   condition "services" {
-    // no additional configuration options currently
+    regexp = "^web.*"
   }
 }
 ```
+- `regexp` - **(beta)** (string) Only services that have a name which matches the regular expression are used by the task. If `regexp` is configured, then [`task.services`](#services) must be omitted or empty. If both a list and a regex are needed, consider including the list as part of the regex or creating separate tasks.
 
 #### Catalog-Services Condition
 
--- a/website/content/docs/nia/tasks.mdx
+++ b/website/content/docs/nia/tasks.mdx
@@ -38,9 +38,9 @@ Below are details on the types of execution conditions that Consul-Terraform-Syn
 
 ### Services Condition
 
-The services condition is the default condition to execute a task. Tasks with the services condition monitor and execute on changes to a list of configured services.
+The services condition is the default condition to execute a task. Tasks with the services condition monitor and execute on either changes to a list of configured services or changes to any services that match a given regex.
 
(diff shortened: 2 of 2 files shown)


website/content/docs/nia/configuration.mdx
```diff
@@ -166,12 +166,12 @@ task {
   providers = ["my-provider"]
   source = "path/to/services-condition-module"
 
-  // optional: declare empty services condition block
   condition "services" {
-    // no additional configuration options currently
+    regexp = "^web.*"
```
findkim, line 170: The below detail on `regexp` states that "if regexp is configured, then task.services must be omitted or empty. But this example now has both `services` and `regexp` set.

Can you update this example code block to be one or the other but not both?

findkim: Another place where dedicated example blocks can be on the other page `/docs/nia/tasks#services-condition` similar to how Catalog-Services condition has an example block.

mkam: Oh whoops, will make the changes!

mkam: Also added an example block showing that the empty service condition block is the same as no condition block.

website/content/docs/nia/tasks.mdx
```diff
@@ -61,6 +61,45 @@ The services condition operates by monitoring the [Health List Nodes For Service
...
 
+The services condition is the default behavior if no `condition` block is configured for a task. The two tasks below have equivalent behavior, which is to execute when any of the listed services have changes to the previously stated attributes.
+
+```hcl
+task {
+  name = "services_condition_task_1"
+  description = "execute on changes to services whose name starts with web"
+  providers = ["my-provider"]
+  source = "path/to/services-condition-module"
+  description = "execute on changes to api, db, and web services"
```
lornasong, line 72: looks like there's two descriptions for this task. suggestion to replace the description at 69 with this one (line 72) so that this description can be at the top of the task config

website/content/docs/nia/tasks.mdx
```diff
@@ -61,6 +61,45 @@ The services condition operates by monitoring the [Health List Nodes For Service
...
+  services = ["api", "db", "web"]
+}
+
+task {
+  name = "services_condition_task_2"
+  description = "execute on changes to services whose name starts with web"
+  providers = ["my-provider"]
+  source = "path/to/services-condition-module"
+  description = "execute on changes to api, db, and web services"
```
lornasong, line 81: similar thing here with the two descriptions

website/content/docs/nia/configuration.mdx
```diff
@@ -154,24 +154,23 @@ A `task` block can be optionally configured with a `condition` block to set the
...
   }
 }
 ```
+- `regexp` - **(beta)** (string) Only services that have a name which matches the regular expression are used by the task. If `regexp` is configured, then [`task.services`](#services) must be omitted or empty. If both a list and a regex are needed, consider including the list as part of the regex or creating separate tasks.
```
lornasong, line 173: Wanted to double-check my understanding on the beta feature - I was thinking that the overall new usage of the services condition was in beta (e.g. there some additional configs we'd like to add) but it seems like I misunderstood and only the regexp feature itself is in beta?

lornasong (commented):
docs are looking good - left a minor suggestion and a question to make sure I understand things

website/content/docs/nia/configuration.mdx
findkim: It's clear! Thanks for the update

website/content/docs/nia/configuration.mdx
mkam: I probably didn't make this clear enough when we last discussed, sorry! I latched on to the idea of calling this a beta feature is because as I was starting to implement query parameters, I ran into more design questions which I've captured in the third use case of this [issue](https://github.com/hashicorp/consul-terraform-sync/issues/357). I think even if we did settle on a desired behavior, there's still the implementation challenge of using service blocks with the regex template function. Overall, I didn't think we'd be able to finish the query parameters in it's entirety in this release.

Without query parameters, the regex feature felt somewhat incomplete to me because it means that:

1. There's no way to include services not in the default datacenter or not in the default namespace when using a regex.
2. Service block configurations are not applied to a service discovered via regex even if the name matches.

But the basic functionality is still there, and users that don't need query parameters can still get value from the regex feature as is. That's why the term beta appealed to me. Happy to discuss more about this, though, if you think we should reevaluate!

lornasong: Gotcha! No, you were super clear. I also like the term beta as well. I should have added more context to my question, sorry!

To clarify, I was contemplating the placement of the word "beta" (e.g. next to `regexp` or "Services Condition" header) and wanted more insight into how you viewed the feature. I had misunderstood and thought that "Services Condition" block itself was in beta since it was missing configs but the regexp config itself worked. However, I see what you're saying that the addition of regex prevents the services condition from having its full functionality.

Thanks for clarifying!

lornasong (approved):
looks good to me for the things we chatted about!

PR #10877, opened 2021-08-19
ui: Remove routeName from tokens route, we do this app wide

johncowen:
I noticed this on my travels whilst working on something else.

At one point we defined a `routeName` on all out controllers so we had access to that property from within all of route templates. A while back we changed this so it was global instead (see our app wide consul specific abstract Route), so this is just dead code removal now. We must have missed this one when we did that PR from while back.


kaxcode (approved):
LGTM

johncowen:
Gonna close this, its been superseded by work in https://github.com/hashicorp/consul/pull/11017

Issue #10861, opened 2021-08-17
Provide clearer warning when Name is used when ID is needed instead


jkirschner-hashicorp:
#### Feature Description

Some entities within Consul have an ID and a Name (e.g., services). When the CLI or an API endpoint for such an entity specifically requires an ID, and will not work with a Name, Consul should provide an error message that makes this mistake clear to the user.

Example taken from #3122: let's say a user has a service with ID `web-service-id` and Name `web-service-name`. If the user executes `/v1/agent/service/deregister/web-service-name`, the following is output to the log:
```
2017/06/07 16:59:47 [WARN] agent: Failed to deregister service "web-service-name": Service does not exist
```

That error message is very misleading, as a service with that _name_ does exist... just not a service with that _id_. The user might reasonably conclude something is wrong with the Consul catalog / state, when the actual problem is that the wrong field was used in the API call. Perhaps if Consul fails to lookup a service with that _id_, it could lookup whether any services exist with that name and, if some do, output a clearer warning message... such as:

```
2017/06/07 16:59:47 [WARN] agent: Failed to deregister service "web-service-name": a service ID must be used to deregister a service, not a service name.
```

This description will be updated to include relevant cases in the list below.

API:
- `/v1/agent/service/deregister/web-service-name`

#### Use Case(s)

Ease troubleshooting of any CLI or API endpoints that accept an ID but not a name.

jkirschner-hashicorp:
@dnephin : I'm not actually sure this is accurate

> PUT `/catalog/deregister`: requires node ID, not name

The [docs for deregister](https://www.consul.io/api-docs/catalog#node-1) do say that the node argument must be "the ID of the node", but I think the code suggests that it's the node name that is used, not node ID. Node name is supposed to be [unique within the cluster](https://www.consul.io/docs/agent/options#_node), so that seems reasonable?

**Q:** What's your read on whether node name or node ID is used for `/catalog/deregister`? I can update the docs if needed.

Below is my attempt to trace through the code, though I feel like I went wrong somewhere... (because I ended up in a services table, not a node table)

------ 

- I think the `Node` arg from the request gets passed to catalog_endpoint `Deregister`: https://github.com/hashicorp/consul/blob/997547bd7fc7719bf4bc5ee608743408ad0ec9e5/agent/consul/catalog_endpoint.go#L334-L337
[...]

Issue #3122, opened 2017-06-08
service did not deregistered properly


waterdudu:
I have both consul cluster and nomad cluster running in same 3 nodes(node-0, node-1, node-2) and a service call `goddess-of-harvest` running with count of 2, deploying by nomad.

After releasing several versions of that service, I noticed a zombie service occured.

After checking the consul log file, I noticed the service was synced and deregistered infinitely. Below is the consul log.

```
2017/06/07 17:24:57 [INFO] agent: Synced service '_nomad-executor-4f37671b-4e02-88cd-e445-94fc626f1f8c-goddess-of-harvest-fe-goddess-of-harvest-fe-goddess-of-harvest-fe-http-service'
2017/06/07 17:25:01 [INFO] agent: Deregistered service '_nomad-executor-4f37671b-4e02-88cd-e445-94fc626f1f8c-goddess-of-harvest-fe-goddess-of-harvest-fe-goddess-of-harvest-fe-http-service'
2017/06/07 17:25:02 [INFO] agent: Synced service '_nomad-executor-4f37671b-4e02-88cd-e445-94fc626f1f8c-goddess-of-harvest-fe-goddess-of-harvest-fe-goddess-of-harvest-fe-http-service'
2017/06/07 17:25:06 [INFO] agent: Deregistered service '_nomad-executor-4f37671b-4e02-88cd-e445-94fc626f1f8c-goddess-of-harvest-fe-goddess-of-harvest-fe-goddess-of-harvest-fe-http-service'
2017/06/07 17:25:07 [INFO] agent: Synced service '_nomad-executor-4f37671b-4e02-88cd-e445-94fc626f1f8c-goddess-of-harvest-fe-goddess-of-harvest-fe-goddess-of-harvest-fe-http-service'
2017/06/07 17:25:11 [INFO] agent: Deregistered service '_nomad-executor-4f37671b-4e02-88cd-e445-94fc626f1f8c-goddess-of-harvest-fe-goddess-of-harvest-fe-goddess-of-harvest-fe-http-service'
2017/06/07 17:25:12 [INFO] agent: Synced service '_nomad-executor-4f37671b-4e02-88cd-e445-94fc626f1f8c-goddess-of-harvest-fe-goddess-of-harvest-fe-goddess-of-harvest-fe-http-service'
[...]

PR #10894, opened 2021-08-22
Clarify service and health check not found error messages

### LINKED ISSUE #9415: Unable to deregister service if multiple nodes available
We have consul with multiple nodes in our environment.
I'm using `spring-cloud-starter-consul-discovery` do register my app instances as a services. And I faced a problem when deregistering services. I'm getting an error:
```
Unknown service {"<service-name>" {}}
```
I tried to repeat the same request with a `curl` and got the same error. 
Command I execute:
```
curl -v -X PUT -H 'X-Consul-Token: ***' https://consul-host/v1/agent/service/deregister/<service-name>
```

What I noticed is that repeating the request multiple times results in success service deregistration.
I guess that what happens is some round-robin behavior, and each time the request is falling in a different node.
I don't know if this is an expected behavior, but maybe it's possible to find some workaround for this issue.
What I need is to deregister my service automatically when stopping/rebooting the application.

jkirschner-hashicorp:
Related to #10861.
Closes #9415.
Partial solution to #3122 (which also warrants a separate docs improvement).

Several practitioners had been confused by the `/agent/service/deregister/:service_id` endpoint (thinking it accepts name when it only accepts ID), and the resulting error message did not help them understand/fix the problem. This PR improves that error message. It also improves a related error message for health check endpoints.

Some examples of new error messages (as generated by the compiled `consul` dev binary):

### Get Service
```
$ curl --request GET http://127.0.0.1:8500/v1/agent/service/counting
Unknown service with ID "counting". Ensure that the service ID is passed, not the service name.
```

### Deregister Service
```
$ curl --request PUT http://127.0.0.1:8500/v1/agent/service/deregister/counting
Unknown service with ID "counting". Ensure that the service ID is passed, not the service name.
```

### Deregister Check
```
$ curl --request PUT http://127.0.0.1:8500/v1/agent/check/de
register/counting
Unknown check with ID "counting". Ensure that the check ID is passed, not the check name.
```

### DIFF
--- a/agent/local/state.go
+++ b/agent/local/state.go
@@ -272,7 +272,7 @@ func (l *State) addServiceLocked(service *structs.NodeService, token string) err
 	}
 
 	if l.agentEnterpriseMeta.PartitionOrDefault() != service.PartitionOrDefault() {
-		return fmt.Errorf("cannot add service %q to node in partition %q", service.CompoundServiceID(), l.config.Partition)
+		return fmt.Errorf("cannot add service ID %q to node in partition %q", service.CompoundServiceID(), l.config.Partition)
 	}
 
 	l.setServiceStateLocked(&ServiceState{
@@ -328,7 +328,14 @@ func (l *State) RemoveServiceWithChecks(serviceID structs.ServiceID, checkIDs []
 func (l *State) removeServiceLocked(id structs.ServiceID) error {
 	s := l.services[id]
 	if s == nil || s.Deleted {
-		return fmt.Errorf("Service %q does not exist", id)
+		// Take care if modifying this error message.
+		// deleteService assumes the Catalog.Deregister RPC call will include "Unknown service"
+		// in the error if deregistration fails due to a service with that ID not existing.
+
+		// When the service register endpoint is called, this error message is also typically
+		// shadowed by vetServiceUpdateWithAuthorizer, which checks for the existence of the
+		// service and, if none is found, returns an error before this function is ever called.
+		return fmt.Errorf("Unknown service ID %q. Ensure that the service ID is passed, not the service name.", id)
 	}
 
 	// To remove the service on the server we need the token.
@@ -539,7 +546,7 @@ func (l *State) addCheckLocked(check *structs.HealthCheck, token string) error {
 	// if there is a serviceID associated with the check, make sure it exists before adding it
 	// NOTE - This logic may be moved to be handled within the Agent's Addcheck method after a refactor
 	if _, ok := l.services[check.CompoundServiceID()]; check.ServiceID != "" && !ok {
-		return fmt.Errorf("Check %q refers to non-existent service %q", check.CheckID, check.ServiceID)
+		return fmt.Errorf("Check ID %q refers to non-existent service ID %q", check.CheckID, check.ServiceID)
 	}
 
 	l.setCheckStateLocked(&CheckState{
@@ -561,7 +568,7 @@ func (l *State) AddAliasCheck(checkID structs.CheckID, srcServiceID structs.Serv
 	defer l.Unlock()
 
 	if l.agentEnterpriseMeta.PartitionOrDefault() != checkID.PartitionOrDefault() {
-		return fmt.Errorf("cannot add alias check %q to node in partition %q", checkID.String(), l.config.Partition)
+		return fmt.Errorf("cannot add alias check ID %q to node in partition %q", checkID.String(), l.config.Partition)
 	}
 	if l.agentEnterpriseMeta.PartitionOrDefault() != srcServiceID.PartitionOrDefault() {
 		return fmt.Errorf("cannot add alias check for %q to node in partition %q", srcServiceID.String(), l.config.Partition)
@@ -612,7 +619,7 @@ func (l *State) RemoveCheck(id structs.CheckID) error {
 func (l *State) removeCheckLocked(id structs.CheckID) error {
 	c := l.checks[id]
 	if c == nil || c.Deleted {
-		return fmt.Errorf("Check %q does not exist", id)
+		return fmt.Errorf("Check ID %q does not exist", id)
 	}
 
 	// If this is a check for an aliased service, then notify the waiters.
--- a/.changelog/10894.txt
+++ b/.changelog/10894.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+api: Improve error message if service or health check not found by stating that the entity must be referred to by ID, not name
+```
--- a/agent/acl.go
+++ b/agent/acl.go
@@ -84,7 +84,13 @@ func (a *Agent) vetServiceUpdateWithAuthorizer(authz acl.Authorizer, serviceID s
 				structs.ServiceIDString(existing.Service, &existing.EnterpriseMeta))
 		}
 	} else {
-		return NotFoundError{Reason: fmt.Sprintf("Unknown service %q", serviceID)}
+		// Take care if modifying this error message.
+		// agent/local/state.go's deleteService assumes the Catalog.Deregister RPC call
+		// will include "Unknown service"in the error if deregistration fails due to a
+		// service with that ID not existing.
+		return NotFoundError{Reason: fmt.Sprintf(
+			"Unknown service ID %q. Ensure that the service ID is passed, not the service name.",
+			serviceID)}
 	}
 
 	return nil
@@ -143,7 +149,7 @@ func (a *Agent) vetCheckUpdateWithAuthorizer(authz acl.Authorizer, checkID struc
 			}
 		}
 	} else {
-		return fmt.Errorf("Unknown check %q", checkID.String())
+		return fmt.Errorf("Unknown check ID %q. Ensure that the check ID is passed, not the check name.", checkID.String())
 	}
 
 	return nil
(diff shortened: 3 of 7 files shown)


agent/local/state.go
```diff
@@ -325,10 +325,27 @@ func (l *State) RemoveServiceWithChecks(serviceID structs.ServiceID, checkIDs []
...
+		}
+	}
+
+	return false
+}
+
 func (l *State) removeServiceLocked(id structs.ServiceID) error {
 	s := l.services[id]
 	if s == nil || s.Deleted {
-		return fmt.Errorf("Service %q does not exist", id)
```
kisunji, line 331: Could we simply improve the error message to be something like:
```
"Service %q does not exist. Ensure that the service ID is passed, not the service name."
```

I'm a _little_ bit wary of the code doing extra `O(l.services)` work just to pass a better error message (in which case we might as well output the found service ID anyways).

jkirschner-hashicorp: @kisunji : my understanding (which could be wrong) is that service ID is required to be unique, but service name is not. I think that's why the endpoint requires a service ID to be passed rather than a name in the first place.

How often does the deregister endpoint get called/in what context? If it doesn't happen often, as long as `l.services` isn't that large, does the extra work have a meaningful impact?

If we are worried about that extra `O(l.services)` work, I like your suggestion to alter the error message in every case.

kisunji: The extra work is probably negligible, but I'd prefer a general error message for "not found" over splitting error states to "not found" and "not found but service with that name found". Just my opinion though! Could try to get more eyes from the team.

msiege2: How about 
```
"Service with ID %q does not exist. Ensure that the service ID is passed, not the service name."
```

My thought is that a customer who sees the new error message may think "well, if you know the service name, just deregister it!".  I understand your point about uniqueness for ID vs. name, but the customer probably doesn't realize that and may end up more frustrated.

jkirschner-hashicorp: I didn't originally go in that direction because I didn't want to mislead someone who was using the endpoint correctly, just with an ID for a service that does not exist anymore (or never did). Perhaps there's a way of altering the wording to satisfy both cases.

More generally, long term, it would be nice for error messages to include a unique ID that allows a user to easily look up troubleshooting information so that the nuance/steps can be conveyed there (while still keeping the essentials in the error message). Not relevant right now for this PR, though.

jkirschner-hashicorp: > The extra work is probably negligible, but I'd prefer a general error message for "not found" over splitting error states to "not found" and "not found but service with that name found".

I think it's worth getting other opinions so that we can have a consistent decision across future, similar situations. I prefer giving users as specific a message as we can for what they did wrong, though I'm open to being persuaded otherwise!

markan: I don't find the time complexity *too* objectionable, but I do have one question.

How does this interact with authorization? Do we check auth for the name? Are we potentially providing an oracle that could allow users to bypass auth and discover names?

jkirschner-hashicorp: @msiege2: Regarding `Service with ID %q does not exist.` ... I had been thinking the same thing. That does beg the question though, should all messages in that file be updated from `service %q` to `service with ID %q`?

e.g.,
`"cannot add service %q to node in partition %q"`
`"Check %q refers to non-existent service %q"`

jkirschner-hashicorp: @markan:

Those are good questions which raise yet another ...

From my read of the endpoint function `AgentDeregisterService`, the `RemoveService` call happens after `vetServiceUpdateWithAuthorizer`. Because `vetServiceUpdateWithAuthorizer` checks to see if a service with that ID exists, and returns an error if there is no service with that ID found, I'm not sure how the `RemoveService` call would ever even be reached from the endpoint. I would need to test this differently to ensure this error message is generated in response to incorrect use of the endpoint (which was the original goal).

https://github.com/hashicorp/consul/blob/d10dd5fbb196934560d72b4d48c723cc5481cb0c/agent/agent_endpoint.go#L1135-L1163

https://github.com/hashicorp/consul/blob/d10dd5fbb196934560d72b4d48c723cc5481cb0c/agent/acl.go#L73-L88

dhiaayachi: I agree with @kisunji. Adding logic to check if the user is mistakenly providing a service name instead of an ID to only change the error message is adding extra code complexity that is not justified in my opinion.

I think it's better if we just reword the generic error message to specify ID `Service with ID %q does not exist` for example.

agent/local/state.go
```diff
@@ -325,10 +325,27 @@ func (l *State) RemoveServiceWithChecks(serviceID structs.ServiceID, checkIDs []
 	return nil
 }
 
+// Checks to see whether the given service ID matches the name of an existing
+// service. Used to see whether a service lookup failed because the user
+// mistakenly provided a service name rather than a service ID.
+func (l *State) doesServiceExistWithNameOfID(lookupID structs.ServiceID) bool {
```
FFMMM, line 331: (still learning, feel free to disregard)

my .02 cents here is that **we could** do such a lookup but the philosophical question is whether we **we should** do it. I'm leaning towards "we shouldn't do this" as it introduces complexity/ surprises on our API path.

I think the right way here is to improve the err msg like Matt suggested:

https://github.com/hashicorp/consul/pull/10894/files#r704624702

jkirschner-hashicorp: I think everyone is agreed on intent: that we should give the user an indication of what they might have done wrong.

The difference of opinion is purely on embodiment. It's certainly simpler to give one error message in all cases. It also avoids the need to consider [Mark's good questions](https://github.com/hashicorp/consul/pull/10894/files#r704630005). I think I've been convinced to go with the "1 common error message" approach, and may need to extend that to 1-2 other places based on the other points raised (e.g., interaction with ACLs).

Thanks everyone!

agent/local/state.go
jkirschner-hashicorp: I ran a Consul dev instance with one service registered (name: `counting`, id: `counting-id`). When I called `/agent/service/deregister` with values `test`, `counting`, and `counting-id`, here's what appeared in the Consul logs:

```
2021-12-09T13:19:39.127-0800 [ERROR] agent.http: Request error: method=PUT url=/v1/agent/service/deregister/test from=127.0.0.1:45106 error="Unknown service "test""
2021-12-09T13:20:01.860-0800 [ERROR] agent.http: Request error: method=PUT url=/v1/agent/service/deregister/counting from=127.0.0.1:45108 error="Unknown service "counting""
2021-12-09T13:20:09.564-0800 [INFO]  agent: Deregistered service: service=counting-id
```

This suggests that, at least with Consul 1.10+, the error message I changed here isn't even exposed to the end user. The error message exposed to the user is definitely coming from `vetServiceUpdateWithAuthorizer`, which returns an error if a service doesn't exist before `RemoveService` is ever called.

https://github.com/hashicorp/consul/blob/bcce800f5514cd99b1d4173051ad9f8cfd8606fa/agent/acl.go#L86

kisunji: Ah yeah I've been bit with that as well. I don't see any reason why we can't change the error message to be a bit more descriptive in `vetServiceUpdateWithAuthorizer` (as well as many other places where we return `"Unknown <entity> %q"`

jkirschner-hashicorp: I've made the update in `vetServiceUpdateWithAuthorizer`. I'll see what else I can find of the form `"Unknown <entity> %q"` (e.g., health checks).

jkirschner-hashicorp: It looks like there's actually source code that depends on the string value of these error messages. I've found one so far. I'm worried there might be others.

https://github.com/hashicorp/consul/blob/bcce800f5514cd99b1d4173051ad9f8cfd8606fa/agent/local/state.go#L1259

Perhaps to reduce risk I should make the error message change to
```
"Unknown service with ID %q. Ensure that the service ID is passed, not the service name."
```

instead of:

```
"Service with ID %q does not exist. Ensure that the service ID is passed, not the service name."
```

?

jkirschner-hashicorp: The if-conditional above actually makes me wonder why the old error message (`"Service %q does not exist"`) never caused a problem, because it doesn't contain `"Unknown service"`... Maybe only because that error message was always (or in most cases) masked by `vetServiceUpdateWithAuthorizer`.

jkirschner-hashicorp:
@kisunji : ready for re-review. On some CI runs, a few tests fail that I don't _think_ are related to my changes, but I'm not sure.

agent/local/state.go
kisunji: Yeah if the test setup didn't involve ACLs it wouldn't have gotten that error

kisunji (approved):
Looks good! I think the issues around callers expecting specific error strings needs to be more elegantly handled (or maybe it's not possible due to RPC) but that's out of scope.

One minor nit/suggestion: do we need "with" in the phrase `"Check with ID"`? I think omitting it makes the error a bit shorter without sacrificing meaning.

jkirschner-hashicorp:
> I think the issues around callers expecting specific error strings needs to be more elegantly handled (or maybe it's not possible due to RPC) but that's out of scope.

Yes. It's bad practice and makes it easy to cause a regression. For now I'm handling that with a comment. But long-term we probably want to remove things like that from the codebase (here and anywhere else we find them).

> One minor nit/suggestion: do we need "with" in the phrase "Check with ID"? I think omitting it makes the error a bit shorter without sacrificing meaning.

That would apply to "Service with ID" also, right?

kisunji:
> That would apply to "Service with ID" also, right?
Yep!

jkirschner-hashicorp:
@kisunji : I addressed the small bit of feedback (removing "with") and rebased to get the tests passing. ready for final review.

kisunji (approved):
🎉

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/539939.

Issue #10715, opened 2021-07-28
Gracefully handle panics in all request handlers


dnephin:
[http.Server](https://pkg.go.dev/net/http#Server) handles panics in request handler goroutines using `recover`, but other servers used by Consul do not. By handling panics in request handlers we will prevent those panics from shutting down the entire process.

* [ ] HTTP API should already work correctly because `http.Server` handles panics
* [ ] XDS `grpc.Server` can use [grpc-middleware/recover](https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware/v2@v2.0.0-rc.2/interceptors/recovery) (or an interceptor with a similar implementation)
* [ ] Server RPC `grpc.Server` can use the same
* [ ] the Server port has many services behind it. `Server.handleConsulConn` can recover from most, but we should also look at `Server.handleSnapshotConn` and `Server.handleInsecureConn` and probably do the same there.
* [ ] DNS interface - unknown. needs investigation

bigmikes:
Concerning `grpc.Server`, what do we want to do with the panic? Just turn it into a `codes.Internal` error? `http.Server` also prints the stack trace to stderr. If stack trace is not needed, something like the following should suffice: 

https://github.com/bigmikes/consul/commit/2b14a9b59ae225d99a5308362cb32c72f0e2ea11

This is a prototype, if this approach is confirmed, I plan to add comments, tests and perhaps try to find a way to make the middleware registration common in a shared function.

PR #10895, opened 2021-08-22
grpc, xds: recovery middleware to return and log error in case of panic

bigmikes:
1. xds and grpc servers:
  1.1. to use recovery [middleware](https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware/v2@v2.0.0-rc.2/interceptors/recovery) with callback that prints stack trace to log
  1.2. callback turn the panic into a `core.Internal` error
  2. added unit test for grpc server

Partially fixes 2nd and 3rd tasks in list: #10715.


bigmikes:
@dnephin, I am not sure why `check-go-mod` check fails. I committed the new `go.mod` and `go.sum`. Is there any issue with that?

dhiaayachi:
Hi @bigmikes, 
Thank you for working on this PR.  We had a discussion with the team about this and the following came out of it:
- We agree on the principal of having a way to recover from a panic
- We are no sure that adding an extra dependency to consul [middleware](https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware/v2@v2.0.0-rc.2/interceptors/recovery) is justified for this.

I will spend some time to check what options are acceptable to solve this and add it in here. So you can adapt the PR to it, if you are still interested in working on it.  

Thank you again and sorry for the long delay.

bigmikes:
Hi @dhiaayachi, 

thanks for letting me know. No worries, let me know once you have the new instructions and I'll fix my PR.

agent/grpc/client_test.go
```diff
@@ -240,3 +243,41 @@ func registerWithGRPC(t *testing.T, b *resolver.ServerResolverBuilder) {
 		resolver.Deregister(b.Authority())
 	})
 }
+
+func TestRecoverMiddleware(t *testing.T) {
```
dnephin, line 247: This test looks great, but I think it should probably be moved to `handler_test.go` since it is mostly testing that the "handler" is being setup correctly.

I think this name would match our naming convention a bit better and make it easier to identify which behaviour is being tested:

```suggestion
func TestHandler_PanicRecoveryInterceptor(t *testing.T) {
```

agent/grpc/handler.go
```diff
@@ -32,6 +50,21 @@ func NewHandler(addr net.Addr, register func(server *grpc.Server)) *Handler {
...
+func newPanicHandler(logger Logger) recovery.RecoveryHandlerFuncContext {
+	return func(ctx context.Context, p interface{}) (err error) {
+		// Log the panic and the stack trace of the Goroutine that caused the panic.
+		stacktrace := hclog.Stacktrace()
+		logger.Error("panic serving grpc request",
+			"panic", p,
+			"stack", stacktrace,
+		)
+
+		return status.Errorf(codes.Internal, "grpc: panic serving request: %v", p)
```
dnephin, line 64: I think generally the panic message wouldn't be too helpful to users, and in the worst case could expose something we don't want exposed. It might be best to remove the panic message and leave just the generic part:

```suggestion
		return status.Errorf(codes.Internal, "grpc: panic serving request")
```

agent/xds/server.go
```diff
@@ -543,11 +545,38 @@ func tokenFromContext(ctx context.Context) string {
...
+		// Log the panic and the stack trace of the Goroutine that caused the panic.
+		stacktrace := hclog.Stacktrace()
+		logger.Error("panic serving grpc request",
+			"panic", p,
+			"stack", stacktrace,
+		)
+
+		return status.Errorf(codes.Internal, "grpc: panic serving request: %v", p)
+	}
+}
```
dnephin, line 561: What do you think about removing this and instead exporting `agent/grpc/NewPanicHandler` ?

`agent/grpc` was added after `agent/xds`, and I think over time it is likely that `agent/xds` will start using more things from `agent/grpc`.

agent/xds/server.go
```diff
@@ -543,11 +545,38 @@ func tokenFromContext(ctx context.Context) string {
...
 	opts := []grpc.ServerOption{
 		grpc.MaxConcurrentStreams(2048),
+		middleware.WithUnaryServerChain(
+			// Add middlware interceptors to recover in case of panics.
+			recovery.UnaryServerInterceptor(recoveryOpts...),
+		),
+		middleware.WithStreamServerChain(
+			// Add middlware interceptors to recover in case of panics.
+			recovery.StreamServerInterceptor(recoveryOpts...),
+		),
```
dnephin, line 579: I guess we could also expose a function from `agent/grpc` that returns these two options, and use that here?   The handler in `agent/grpc` could use the same `PanicHandlerMiddleware` function to add them to opts, that way its clear that both of these gRPC servers are using the same middleware.

```go
opts :=  []grpc.ServerOption{grpc.MaxConcurrentStreams(2048)}
opts = append(opts, agentgrpc.PanicHandlerMiddleware(s.Logger)...)
```

What do you think?

dnephin (commented):
Thank you for working on this! I think this is looking good. I left a couple suggestions below for the test and sharing implementation with the two servers.

We also noticed that `v2` of this middleware library is still a pre-release (rc2), so may not be production ready yet. I think we should either use `v1`, or write our own interceptor. I think we'd be fine with either.

bigmikes:
Hi @dhiaayachi, @dnephin. FYI, this PR is ready for review based on your latest comments.

dhiaayachi (approved):
Hey @bigmikes , sorry for the delay. This is good to go from my perspective. 
I will let @dnephin take another look before merging.

go.sum
```diff
@@ -214,6 +214,7 @@ github.com/gophercloud/gophercloud v0.1.0 h1:P/nh25+rzXouhytV2pUHBb65fnds26Ghl8/
 github.com/gophercloud/gophercloud v0.1.0/go.mod h1:vxM41WHh5uqHVBMZHzuwNOHh8XEoIEcSTewFxm1c5g8=
 github.com/gorilla/websocket v1.4.0/go.mod h1:E7qHFY5m1UJ88s3WnNqhKjPHQ0heANvMoAMk2YaljkQ=
 github.com/gregjones/httpcache v0.0.0-20180305231024-9cad4c3443a7/go.mod h1:FecbI9+v66THATjSRHfNgh1IVFe/9kFxbXtjV0ctIMA=
+github.com/grpc-ecosystem/go-grpc-middleware v1.0.0 h1:Iju5GlWwrvL6UBg4zJJt3btmonfrMlCDdsejg4CZE7c=
```
dnephin, line 217: I looking into using a newer 1.x version, but it requires updating gRPC , which I don't particularly want to do right now. I looked at the diff, and nothing important has changed since 1.0, so I think this is good.

dnephin (approved):
I fixed the merge conflict. If CI is still happy I will merge.

Thanks again!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/520564.

PR #10903, opened 2021-08-24
Add Support to for providing TLS certificates for Ingress listeners from an SDS source

banks:
This builds on #10613.

This PR adds a new low-level feature for Ingress Gateways. It allows operators to specify an external Envoy SDS (Secret Discovery Service) Service and allows loading TLS certificates for listeners.

The PR commits logically work through each package/area that needed updates with relevant tests so may be a cleaner way to view the diff.

Documentation still to-do but the intended usage is as shown in the integration test:

 1. Operator registers the Ingress service with one or more custom static cluster sin bootstrap using the `Proxy.Config.envoy_extra_static_clusters_json` configuration. This static cluster specifies how to connect to the SDS service(s) and takes care of authenticating the ingress to that SDS service however the SDS integration requires.
 2. The operator can then use new config fields in the `ingress-gateway` Config Entry to specify which SDS certificate resource names to load (from which clusters) for the whole gateway, each listener or per service.

## TODO Before 

This is ready for review as it is I think as the code is all here. There are several additional bits that I will either do in this PR or a follow up before this is released:

 - [x] Changelog entry
 - [x]  `api` package updates for Config Entry

## TODO In later PR

 - [ ] Document the new options in Ingress Gateway Config Entry
 - [ ] Document how to use SDS end-to-end (as an "Advanced Topic")
    - I _don't_ think we need a Learn guide here because this is low-level only really needed for folks that want to build a custom integration with an SDS service. We will later provide an opinionated way to do this that will hide all these details from end-users.


agent/xds/listeners.go
```diff
@@ -542,29 +599,136 @@ func (s *ResourceGenerator) makeIngressGatewayListeners(address string, cfgSnap
...
+	// If the upsteam service doesn't have any TLS overrides then it can just use
+	// the combined filterchain with all the merged routes.
+	if !ingressServiceHasSDSOverrides(s) {
+		return key.RouteName(), nil
+	}
+
+	// Return a specific route for this service as it needs a custom FilterChain
+	// to serve it's custom cert so we should attach it's routes to a separate
+	// Route too.
+	return fmt.Sprintf("%s_%s", key.RouteName(), s.ToServiceName().ToServiceID().String()), nil
```
banks, line 661: TODO: 
 - [ ] This causes conflicts in Ent tests because it means that the default ns service `web` has a route named `web` in OSS but `default/default/web` in Ent. causing all golden files to conflict. This probably needs to follow the same logic `structs.Upstream.Identifier` which omits default values to keep the identifiers consistent between OSS and Ent. Shame we can't easily re-use that method from here with the types available.

banks: Fixed in later commit.

banks:
Updated with fixes from Enterprise version.

kisunji (approved):

rboyer (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/455247.

PR #10905, opened 2021-08-24
Add address prefix path for using proxy address.

### LINKED ISSUE #10354: enhance consul/api to allow access of consul behind an HTTP API gateway (e.g. nginx)
#### Feature Description

I want to access a consul service that is behind an nginx API gateway (see
nginx.conf snippet below) i.e. instead of acessing consul directly like this:

`curl -v http://10.1.2.3:8500/v1/kv/x/y/z/?recurse=&wait=1800000ms`

I want to access consul behind an nginx API gateway, like this:

`curl -v http://10.1.2.3/foo/bar/consul/v1/kv/x/y/z/?recurse=&wait=1800000ms`

This works fine with curl creating the HTTP requests but the `api.Config`
 does not have any provision for adding a prefix path to
requests.  When I try the following

	c := &api.Config{
		Address: "http://10.1.2.3/foo/bar/consul",
	}

consul fails because it tries to resolve `10.1.2.3/foo%2Fbar%2Fconsul` as a hostname

error: `Get "http://10.1.2.3%2Faugtera%2Fconfdb/v1/kv/x/y/z/?recurse=&wait=1800000ms": dial tcp: lookup 10.1.2.3/foo/bar/consul: no such host`

#### Use Case(s)

using consul that is behind an API gateway

#### Sample diffs
I don't have permission to push a branch for a PR.  Here is a diff 

    $ git diff
    diff --git a/api/api.go b/api/api.go
    index 5cdd48680..0c74ac878 100644
    --- a/api/api.go
    +++ b/api/api.go
    @@ -285,6 +285,9 @@ type Config struct {
	    // Scheme is the URI scheme for the Consul server
	    Scheme string

    +       // Prefix for URIs for when consul is behind an API gateway
    +       PathPrefix string
    +
	    // Datacenter to use. If not provided, the default agent datacenter is used.
	    Datacenter string

    @@ -663,6 +666,13 @@ func NewClient(config *Config) (*Client, error) {
		    config.Address = parts[1]
	    }

    +       // extract API gateway PathPrefix, if present
    +       parts = strings.SplitN(config.Address, "/", 2)
    +       if len(parts) == 2 {
    +               config.Address = parts[0]
    +               config.PathPrefix = "/" + parts[1]
    +       }
    +
	    // If the TokenFile is set, always use that, even if a Token is configured.
	    // This is because when TokenFile is set it is read into the Token field.
	    // We want any derived clients to have to re-read the token file.
    @@ -896,7 +906,7 @@ func (c *Client) newRequest(method, path string) *request {
		    url: &url.URL{
			    Scheme: c.config.Scheme,
			    Host:   c.config.Address,
    -                       Path:   path,
    +                       Path:   c.config.PathPrefix + path,
		    },
		    params: make(map[string][]string),
		    header: c.Headers(),

#### nginx.conf snippet

    location /foo/bar/consul {
        rewrite ^/foo/bar/consul/(.*)  /$1 break;
        proxy_pass http://consul:8500;
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection 'upgrade';
        proxy_set_header Host $host;
        proxy_cache_bypass $http_upgrade;
        proxy_no_cache 1;
    }

ekhabarov:
Fixes #10354

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -291,11 +291,15 @@ type HttpBasicAuth struct {
 
 // Config is used to configure the creation of a client
 type Config struct {
-	// Address is the address of the Consul server
+	// Address is the address of the Consul server or reverse proxy server which
+	// redirects to Consul server.
 	Address string
 
 	// Scheme is the URI scheme for the Consul server
 	Scheme string
+
+	// Prefix for URIs for when consul is behind an API gateway.
+	PathPrefix string
 
 	// Datacenter to use. If not provided, the default agent datacenter is used.
 	Datacenter string
@@ -660,8 +664,19 @@ func NewClient(config *Config) (*Client, error) {
 		}
 	}
 
+	var addr, pref string
+
 	parts := strings.SplitN(config.Address, "://", 2)
 	if len(parts) == 2 {
+		switch parts[0] {
+		case "http", "https":
+			// for http and https addresses check if address contains prefix path.
+			parts := strings.SplitN(parts[1], "/", 2)
+			if len(parts) == 2 {
+				addr, pref = parts[0], "/"+parts[1]
+			}
+		}
+
 		switch parts[0] {
 		case "http":
 			// Never revert to http if TLS was explicitly requested.
@@ -680,7 +695,13 @@ func NewClient(config *Config) (*Client, error) {
 		default:
 			return nil, fmt.Errorf("Unknown protocol scheme: %s", parts[0])
 		}
+
 		config.Address = parts[1]
+
+		if addr != "" && pref != "" {
+			config.Address = addr
+			config.PathPrefix = pref
+		}
 	}
 
 	// If the TokenFile is set, always use that, even if a Token is configured.
@@ -922,7 +943,7 @@ func (c *Client) newRequest(method, path string) *request {
 		url: &url.URL{
 			Scheme: c.config.Scheme,
 			Host:   c.config.Address,
-			Path:   path,
+			Path:   c.config.PathPrefix + path,
 		},
 		params: make(map[string][]string),
 		header: c.Headers(),


api/api.go
```diff
@@ -660,8 +664,19 @@ func NewClient(config *Config) (*Client, error) {
...
 	}
 
+	var addr, pref string
+
 	parts := strings.SplitN(config.Address, "://", 2)
 	if len(parts) == 2 {
+		switch parts[0] {
+		case "http", "https":
+			// for http and https addresses check if address contains prefix path.
+			parts := strings.SplitN(parts[1], "/", 2)
```
eculver, line 674: I'm curious if we should be using `url.Parse` here instead? I realize we use `strings.SplitN` elsewhere in this package, but in this case, since we need to deal with both the address and the protocol prefix, I'm wondering if we should just use functionality from the standard library?

ekhabarov: Thanks for the response. It's a good point, haven't thought about that. I think using `url.Parse` makes code a bit clearer and it shouldn't increase the PR size overall.

eculver: Were you still planning to make this change?

eculver (commented):
I think all that is left is use of `url.Parse` and adding a test if possible. Sound good?

api/api.go
ekhabarov: @eculver working on it.

ekhabarov: Tried, the issue is `url.Parse` doesn't allow to parse schemeless address, which `parts[1]` is. It's possible to add schema here, but it's pointless, it's more clear and easier to leave this code "as is".

eculver: Got it. In this case, can we add a comment about the situation and then, if possible, just adding a test and we can get this merged.

eculver:
@funkiestj just checking in...my last comment is still valid if you're still up for taking this on.

funkiestj:
I am still interested but I won't get to it for a few weeks.

funkiestj:
I'd like to reopen this.  I'm ready to drive it to completion.

funkiestj:
I opened a[ new PR](https://github.com/hashicorp/consul/pull/12914) and linked that PR to this one

PR #10907, opened 2021-08-24
UI: Rollback attributes when saving/updating fails

DingoEatingFuzz:
This way the client-side state continues to match the server-side state.

This was observed when editing ACL roles when the active token only allowed read access. Attempting to save changes would result in a server error, but the client-side state would still show the change. This is ultimately harmless but still spooky.

Note that 1.10 introduced a permissions-based UI (#9835) that make this particular reported issue no longer possible. Since this change impacts the underlying repo service and blocking actions mixin, it makes sense to still commit this change.

### DIFF
--- a/ui/packages/consul-ui/app/mixins/with-blocking-actions.js
+++ b/ui/packages/consul-ui/app/mixins/with-blocking-actions.js
@@ -1,6 +1,7 @@
 import Mixin from '@ember/object/mixin';
 import { inject as service } from '@ember/service';
 import { set } from '@ember/object';
+import { TYPE_ERROR } from 'consul-ui/services/feedback';
 /** With Blocking Actions
  * This mixin contains common write actions (Create Update Delete) for routes.
  * It could also be an Route to extend but decoration seems to be more sense right now.
@@ -90,6 +91,9 @@ export default Mixin.create({
         },
         'update',
         (type, e) => {
+          if (type === TYPE_ERROR) {
+            this.repo.rollback(item);
+          }
           return this.errorUpdate(type, e);
         }
       );
--- a/ui/packages/consul-ui/app/services/feedback.js
+++ b/ui/packages/consul-ui/app/services/feedback.js
@@ -1,8 +1,8 @@
 import Service, { inject as service } from '@ember/service';
 import callableType from 'consul-ui/utils/callable-type';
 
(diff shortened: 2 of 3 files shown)


johncowen:
Hey @DingoEatingFuzz , thanks for digging into this!

Just a little process thing when working on Consul. Would you be able to change the branch name to `ui/something/something`. If a branch doesn't begin with `ui/` it means CI isn't kicked off for a PR. We have some brief "this is how we work" docs in our engineering-docs/consul-ui folder if you wanted to take a look in there, there might be the odd other thing. If not no worries.

I had a quick scan of the approach here and my main feedback (honestly didn't been the pun there 😅 😂 ) would be: I think this approach crosses over boundaries a little. I think my question would be why does this approach put ember-data housekeeping/model concerns in a feedback service 'feedback' being mainly for view concerns? Is there a better way to solve this? I think the `rollbackAttributes` is 👌 , it's just the where we should do it.

I think I have a suggestion for a different approach to solve this, lemme know.

Umm, also I think we have an `invalidate` method on Repository already, which invalidates all-the-things, how about we just reuse that? i.e. If it has a `item` argument we only invalidate/rollback a single item. If we haven't specified an item then just invalidate the whole lot as it does now?

johncowen:
Hey 👋 

I've spent some of this morning with my head in our Adapters for our admin partition work and I noticed we do throw an InvalidError but only when we get a 422 HTTP error code.

https://github.com/hashicorp/consul/blob/3254308fd3e018f67ed13abe0d652fdc8d8a359c/ui/packages/consul-ui/app/adapters/http.js#L119-L121

When we made our HTTPAdapter base adapter we took the ember-data 3.12 RESTAdapter and just reduced it down as much as we could but try to keep the same functionality as the RESTAdapter (which I think we inherited from prior to this), so I know we took that from here:

https://github.com/emberjs/data/blob/e0732481885abde4eb187d3a1573276374068a2f/packages/adapter/addon/rest.js#L901-L927

I don't think Consul ever sends us a 422, although I reverse proxy of some sort might? But overall I think we are good with the decision we came to here of `isError` being a good signal. From the info we compiled here, I don't think we need to be concern with what it said in the ember-data documentation:

> If true the adapter reported that it was unable to save local changes to the backend for any reason other than a server-side validation error.

PR #10835, opened 2021-08-12
ui: Create Routing Configurations route and page

kaxcode:
### ✨ Description:

This PR creates the Routing Configurations route and page.

Why: A routing configuration is not a real service. They are a routing configuration that route traffic to real services in Consul. For this reason we created a new route for routing-configs that does not follow the services `/dc/service/:name-of-service` pattern.

Changes:

- Created the new route `dc/routing-config/:name` and template for Routing Configurations. The Routing Config page will only show Discovery Chain. No other tabs/pages are expected to be added in the future.
- Create `Consul::Source` component. This is a component very similar to `Consul::Kind` . Consul uses the attribute Kind to describe Services, e.g. Terminating Gateway, Mesh Gateway, etc. Routing Configuration are not Services and so I create a new pill component to be used by any item by does not relate to Services.

[Demo Link](https://consul-ui-staging-4w5nzpsql-hashicorp.vercel.app/ui/dc1/routing-config/virtual-1#CONSUL_ACLS_ENABLE=1
)

### 📸 Screenshots:
Screen Shot 2021-08-12 at 9 42 08 AM

### ⚡ Backend Changes:

No backend changes required.

### 🤡 Updates to mock-api:

No updates to mock data.

### 🧪 Testing:

Mostly presentational tests. Tests cover the following...

- The URL redirects to the correct template by checking page title
- When redirected you don't get a 404 error because the URL exists
- Check that the new `Consul::Source` component is there. It should always be there.

### 📝  PR Tasks

- [x] Changelog
- [x] Demo Link

PR #10915, opened 2021-08-26
ui: Ensure we clean up javascript based DataSources

johncowen:
Continuation of https://github.com/hashicorp/consul/pull/10835

DataSources added via javascript require cleaning up by assigning it to
an EventSource component in the template, otherwise the blocking query
can stick around when its no longer needed.

We also took the opportunity here to make this consistent with the rest
of the application.

1. Don't use `@model`.
2. Use our `<Route />` component.
3. Only use extra divs/spans specifically for styling only as a last resort.

### DIFF
--- a/.changelog/10915.txt
+++ b/.changelog/10915.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Ensure routing-config page blocking queries are cleaned up correctly
+```
--- a/ui/packages/consul-ui/app/routes/dc/routing-config.js
+++ b/ui/packages/consul-ui/app/routes/dc/routing-config.js
@@ -13,7 +13,12 @@ export default class RoutingConfigRoute extends Route {
       dc: dc,
       nspace: nspace,
       slug: name,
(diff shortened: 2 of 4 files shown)


kaxcode (approved):
👍

johncowen:
This is now just for 1.10.x and is therefore currently waiting on https://github.com/hashicorp/consul/pull/11127

johncowen:
Ok so https://github.com/hashicorp/consul/pull/11095 and https://github.com/hashicorp/consul/pull/11127 are now in a rebased here and the tests are now passing as expected. We don't need this fix on `main` as we have a better approach there for this overall, but it is still needed for 1.10.x.

Problem now is, I added a changelog to explain the change here, but the GH action (changelog-checker) is telling me I haven't 😭 

Kinda wondering whether I should just bypass that and merge, will ask the team

johncowen:
Changelog is added here, we need to tweak the changelog-checker script slightly, which we'll do post merge.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/457052.

PR #10916, opened 2021-08-26
ui: Always show main navigation Key/Value link

johncowen:
Currently there is no way for us to use our HTTP authorization API
endpoint to tell us whether a user has access to any KVs (including the
case where a user may not have access to the root KV store, but do have
access to a sub item)

This is a little weird still as in the above case the user would click
on this link and still get a 403 for the root, and then have to manually
type in the URL for the KV they do have access to.

Despite this we think this change makes sense as at least something about KV is
visible in the main navigation.

Once we have the ability to know if any KVs are accessible, we can add
this guard back in.

We'd initially just removed the logic around the button, but then
noticed there may be further related KV issues due to the nested nature
of KVs so we finally decided on simply ignoring the responses from the
HTTP API, essentially reverting the KV area back to being a thin client.
This means when things are revisted in the backend we can undo this
change easily in one place.

- edit: fixes https://github.com/hashicorp/consul/issues/10944
- edit: fixes https://github.com/hashicorp/consul/issues/10997
- edit: fixes https://github.com/hashicorp/consul/issues/11023

### DIFF
--- a/.changelog/10916.txt
+++ b/.changelog/10916.txt
@@ -1,0 +1,4 @@
+```release-note:bug
+ui: Ignore reported permissions for KV area meaning the KV is always enabled
+for both read/write access if the HTTP API allows.
+```
--- a/ui/packages/consul-ui/app/abilities/kv.js
+++ b/ui/packages/consul-ui/app/abilities/kv.js
@@ -10,4 +10,15 @@ export default class KVAbility extends BaseAbility {
     }
     return resources;
   }
+  get canRead() {
+    return true;
+  }
+
+  get canList() {
+    return true;
+  }
+
+  get canWrite() {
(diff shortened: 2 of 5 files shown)


kaxcode (approved):
👍

kaxcode:
Test needs to be update to see kvs on navigation.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/453533.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit cf638ee551c2d6ea3a32a7f416c44ad435c49b26 onto `release/1.10.x` succeeded!

PR #10955, opened 2021-08-30
docs: Add TOC to top-level stanzas in K8s Helm docs

david-yu:
Draft, not merging since this will need to be included in the helm docs generation Golang code.


david-yu:
Closing as is addressed by https://github.com/hashicorp/consul-k8s/pull/742

PR #10956, opened 2021-08-30
Update contributing guide

dnephin:
[Preview](https://github.com/hashicorp/consul/blob/dnephin/update-contrib/.github/CONTRIBUTING.md) the changes.

* Replace the mailing list link with Discuss. The mailing list has not been used for a few years.
* `make format` no longer exists, so replace it with `gofmt`
* Remove the instructions about running all tests locally, and debugging flakes.
* Update note about vendoring of dependencies.
* Add link to developer docs.

I suspect the only potentially controversial change here is the guidance around running all tests and finding flaky tests. I believe this is a good change for a few reasons:
1. Tests passing locally is not really relevant. The important signal is that tests pass in a predictable environment (CI), so having contributors fight with our test suite that can fail for a number of reasons is unnecessary.
2. Finding and debugging flaky tests is challenging. The goal of this document is to make it easier for first time contributors. Pointing new contributors at this advanced workflow is more likely to be a distraction.
3. We rarely (maybe never) use this makefile target. It is likely broken, and it is not the method we are planning on using for detecting flakes in the future. So the sooner we remove it from the docs, the easier it will be for us to remove it from the Makefile later.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -26,7 +26,7 @@ are deployed from this repo.
 
 ### Reporting an Issue:
 >Note: Issues on GitHub for Consul are intended to be related to bugs or feature requests. 
->Questions should be directed to other community resources such as the: [Mailing List](https://groups.google.com/group/consul-tool/), [FAQ](https://www.consul.io/docs/faq.html), or [Guides](https://www.consul.io/docs/guides/index.html).
+>Questions should be directed to other community resources such as the: [Discuss Forum](https://discuss.hashicorp.com/c/consul/29), [FAQ](https://www.consul.io/docs/faq.html), or [Guides](https://www.consul.io/docs/guides/index.html).
 
 * Make sure you test against the latest released version. It is possible we 
 already fixed the bug you're experiencing. However, if you are on an older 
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -86,66 +86,32 @@ to work on the fork is to set it as a remote of the Consul project:
...
-
-* **FLAKE_CPUS** Amount of CPU resources for container
-
-* **FLAKE_N** Number of times to run tests
-
-Examples:
-`make test-flake FLAKE_PKG=connect/proxy`
-`make test-flake FLAKE_PKG=connect/proxy FLAKE_TEST=TestUpstreamListener`
-`make test-flake FLAKE_PKG=connect/proxy FLAKE_TEST=TestUpstreamListener FLAKE_CPUS=0.15 FLAKE_N=30`
+When a pull request is opened CI will run all tests and lint to verify the change.
```
kisunji, line 100: Is this true? I think I've seen some PRs require approval from a `member` to run some CI tests (maybe to prevent malicious test runs)

kisunji (approved):
Just one potential clarification; otherwise LGTM

.github/CONTRIBUTING.md
dnephin: Good question. Github has added a prompt for first time contributors that requires a member to click it. But I believe that only applies to github actions. I think I've seen CircleCI checks run even without that button being clicked. It's possible I don't fully understand how that new button works, but either way, one way or another we will use CI to verify the change.

There are some cases where CircleCI just doesn't seem to want to run anything (ex: #10701), but I think that is due to a bug. Normally CI should run on all PRs.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/437919.

PR #10958, opened 2021-08-30
Configurable max DNS requests over reused TCP conn

dpn:
Currently the DNS server implementation closes resused TCP connections after `maxTCPQueries` https://pkg.go.dev/github.com/miekg/dns#Server.MaxTCPQueries  have been processed. This change exposes this setting in the Consul agent config so that we can customize this value to our liking.

   `agent/config/config.go`.
   `agent/config/runtime.go`.
   translate.
      - [x] Add validation to Validate in `agent/config/builder.go`.
 - [ ] **If** your config should take effect on a reload/HUP.
        more of the following places:
        `TestAgent_reloadConfig*`.
 - [x] Add documentation to `website/content/docs/agent/options.mdx`.

Done! You can now use your new field in a client agent by accessing
`s.agent.Config.<FieldName>`.

If you need a CLI flag, access to the variable in a Server context, or touched
the Service Definition, make sure you continue on to follow the appropriate
checklists below.

### DIFF
--- a/agent/dns.go
+++ b/agent/dns.go
@@ -86,6 +86,7 @@ type dnsConfig struct {
 	UDPAnswerLimit   int
 	ARecordLimit     int
 	NodeMetaTXT      bool
+	MaxTCPQueries    int
 	SOAConfig        dnsSOAConfig
 	// TTLRadix sets service TTLs by prefix, eg: "database-*"
 	TTLRadix *radix.Tree
@@ -166,6 +167,7 @@ func GetDNSConfig(conf *config.RuntimeConfig) (*dnsConfig, error) {
 		DisableCompression: conf.DNSDisableCompression,
 		UseCache:           conf.DNSUseCache,
 		CacheMaxAge:        conf.DNSCacheMaxAge,
+		MaxTCPQueries:      conf.DNSMaxTCPQueries,
 		SOAConfig: dnsSOAConfig{
 			Expire:  conf.DNSSOA.Expire,
 			Minttl:  conf.DNSSOA.Minttl,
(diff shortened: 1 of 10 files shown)


markan:
This PR looks pretty solid, but I'm not totally sure what the problem you are solving is.

Can you give us a little background on the use case here? Is this a performance optimization, or is there a correctness or other issue?

agent/dns.go
```diff
@@ -243,6 +245,8 @@ func (d *DNSServer) ListenAndServe(network, addr string, notif func()) error {
 	if network == "udp" {
 		d.UDPSize = 65535
 	}
+
+	d.MaxTCPQueries = cfg.MaxTCPQueries
```
markan, line 249: This might need to be set on the dns.Server record, e.g. d.Server.MaxTCPQueries to have the effect you want.

dhiaayachi: I think this is correct because `Server` is embedded in `DNSServer` so `d.MaxTCPQueries` is equivalent to `d.Server.MaxTCPQueries`

dhiaayachi (commented):
Hey @dpn, Thank you for working on this. We need to have a test for this to merge it, are you willing to add it?

`TestDNS_Over_TCP` is a good starting point. we just need to send the request multiple times and expect the server to close the connection after the limit is reached.

Amier3:
Hey @dpn 

Checking in to see if you're still interested in working on this? 😄

dpn:
Hey @Amier3, we definitely are we've just had many other, larger fires (not Consul related :)) to put out so this has been de-prioritized on our end. Not sure if it makes sense to just close this to get if out of your queue until we're able to loop back around, but if it does feel free and I'll re-engage when we're able to.

dpn:
I mean, I guess I'll just go ahead and do that.  Will re-engage once we're able to support more from our end. Thanks everyone!

PR #10964, opened 2021-08-31
use intermediates when verifying

i0rek:
Until now Consul never used the intermediate certificates to verify TLS certs, but it should! Like the [stdlib](https://pkg.go.dev/crypto/tls#example-Config-VerifyConnection).

### DIFF
--- a/.changelog/10964.txt
+++ b/.changelog/10964.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+tls: use intermediate certificates when verifying TLS certificates.
+```
--- a/.changelog/10964.txt
+++ b/.changelog/10964.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-tls: use intermediate certificates when verifying TLS certificates.
+tls: consider presented intermediates during server connection tls handshake.
 ```
--- a/tlsutil/config.go
+++ b/tlsutil/config.go
@@ -935,7 +935,7 @@ func (c *Configurator) AuthorizeServerConn(dc string, conn *tls.Conn) error {
 		}
 		c.logger.Debug("AuthorizeServerConn failed certificate validation", "error", err)
 	}
-	return fmt.Errorf("a TLS certificate with a CommonName of %v is required for this operation", expected)
+	return fmt.Errorf("AuthorizeServerConn failed certificate validation with a CommonName of %v", expected)
 }
 
 // ParseCiphers parse ciphersuites from the comma-separated string into
--- a/tlsutil/config.go
+++ b/tlsutil/config.go
@@ -923,8 +923,8 @@ func (c *Configurator) AuthorizeServerConn(dc string, conn *tls.Conn) error {
 		opts := x509.VerifyOptions{
 			DNSName:       expected,
 			Intermediates: x509.NewCertPool(),
-			Roots: caPool,
-			KeyUsages: []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
+			Roots:         caPool,
+			KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
 		}
 		for _, cert := range cs.PeerCertificates[1:] {
 			opts.Intermediates.AddCert(cert)
--- a/tlsutil/config.go
+++ b/tlsutil/config.go
@@ -937,7 +937,7 @@ func (c *Configurator) AuthorizeServerConn(dc string, conn *tls.Conn) error {
 		}
 		multierror.Append(errs, err)
 	}
-	return fmt.Errorf("AuthorizeServerConn failed certificate validation for certificate with a CommonName of %v: %w", expected, errs)
+	return fmt.Errorf("AuthorizeServerConn failed certificate validation for certificate with a SAN.DNSName of %v: %w", expected, errs)
 
 }
 
--- a/tlsutil/config.go
+++ b/tlsutil/config.go
@@ -919,12 +919,16 @@ func (c *Configurator) AuthorizeServerConn(dc string, conn *tls.Conn) error {
 		if len(chain) == 0 {
 			continue
 		}
-		clientCert := chain[0]
-		_, err := clientCert.Verify(x509.VerifyOptions{
-			DNSName:   expected,
-			Roots:     caPool,
+		opts := x509.VerifyOptions{
+			DNSName:       expected,
+			Intermediates: x509.NewCertPool(),
+			Roots: caPool,
 			KeyUsages: []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
-		})
+		}
+		for _, cert := range chain.PeerCertificates[1:] {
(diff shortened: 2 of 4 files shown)


.changelog/10964.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+tls: use intermediate certificates when verifying TLS certificates.
```
crhino, line 2: I would perhaps fine tune this changelog a bit, since to be fair, we do use intermediate certificates when verifying TLS, by setting: https://github.com/hashicorp/consul/blob/ea451971d4a5c4eb88808eba5e95742e7778a1fa/tlsutil/config.go#L523-L525
That ends up [here](https://github.com/golang/go/blob/3920d6f2085559ae262d651b00bf8b29f953580a/src/crypto/tls/handshake_server.go#L813-L829) in the Go `tls` package.

It is just this particular verification that does not use the client's intermediate CAs.

i0rek: > // RequireAndVerifyClientCert indicates that a client certificate should be requested
	// during the handshake, and that at least one valid certificate is required
	// to be sent by the client.

crhino: During a TLS connection, the Go stdlib will end up [here](https://github.com/golang/go/blob/3920d6f2085559ae262d651b00bf8b29f953580a/src/crypto/tls/handshake_server.go#L813-L829) in `processCertsFromClient`:

Copying the code into the comment:
```
	if c.config.ClientAuth >= VerifyClientCertIfGiven && len(certs) > 0 {
		opts := x509.VerifyOptions{
			Roots:         c.config.ClientCAs,
			CurrentTime:   c.config.time(),
			Intermediates: x509.NewCertPool(),
			KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
		}

		for _, cert := range certs[1:] {
			opts.Intermediates.AddCert(cert)
		}

		chains, err := certs[0].Verify(opts)
		if err != nil {
			c.sendAlert(alertBadCertificate)
			return errors.New("tls: failed to verify client certificate: " + err.Error())
		}
```

So if `c.config.ClientAuth >= VerifyClientCertIfGiven && len(certs) > 0` is true, the `tls.Conn` will validate the client's certificate by adding the Intermediates like this PR does.

`RequireAndVerifyClientCert` is greater than `VerifyClientCertIfGiven`, so we should be falling into this statement.

i0rek: Thanks for clarifying!

i0rek: So that means that when the old code created a TLS connection as the server, it properly checked the intermediate certs. But when it was acting as the client, it wouldn't.

crhino: I only looked at the code around verifying client certificates, but I would not be surprised if the Go stdlib did the same thing when acting as a client validating a server's certificates.

tlsutil/config.go
```diff
@@ -915,22 +915,27 @@ func (c *Configurator) AuthorizeServerConn(dc string, conn *tls.Conn) error {
...
+			KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
+		}
+		for _, cert := range cs.PeerCertificates[1:] {
+			opts.Intermediates.AddCert(cert)
+		}
+		_, err := cs.PeerCertificates[0].Verify(opts)
 		if err == nil {
 			return nil
 		}
 		c.logger.Debug("AuthorizeServerConn failed certificate validation", "error", err)
```
mkeeler, line 936: It might be nice to also set this log to emit at the Warning level.

dnephin: I added this log message originally because I was worried about exposing too much detail in the error message, but I realize now that all we do with that error is log it at the WARN level (in the caller).

Since this debug log line is in a loop, I think it might not be ideal to make this a warning, because if the client sends multiple cert chains we'll get warnings here that are misleading. 

To better expose the errors, maybe we should build up a slice of errors, and append those to the error message so that the caller can log all the details. If we did that we could probably remove this debug line because we'd have all the details in the returned error.

tlsutil/config.go
```diff
@@ -915,22 +915,27 @@ func (c *Configurator) AuthorizeServerConn(dc string, conn *tls.Conn) error {
...
 	}
-	return fmt.Errorf("a TLS certificate with a CommonName of %v is required for this operation", expected)
+	return fmt.Errorf("AuthorizeServerConn failed certificate validation for certificate with a CommonName of %v", expected)
```
dnephin, line 938: I should not have left this as "CommonName". Earlier versions did use the common name, but I believe `Certificate.Verify` only checks the SAN DNSNames. So I guess something like this would be more correct:
```suggestion
	return fmt.Errorf("AuthorizeServerConn failed certificate validation for certificate with a SAN.DNSName of %v", expected)
```

dnephin (commented):
Thank you for fixing this! I think the fix looks correct.

I left a couple suggestions for improving the logging that came out of a retrospective.

I guess we should also have a test that shows this now works with intermediate certs.

i0rek:
@dnephin I changed it to use multi error and fixed the log line to use `SAN.DNSName`.

tlsutil/config.go
```diff
@@ -842,15 +843,11 @@ func (c *Configurator) wrapTLSClient(dc string, conn net.Conn) (net.Conn, error)
 		Intermediates: x509.NewCertPool(),
 	}
 
-	certs := tlsConn.ConnectionState().PeerCertificates
-	for i, cert := range certs {
-		if i == 0 {
-			continue
-		}
+	cs := tlsConn.ConnectionState()
+	for _, cert := range cs.PeerCertificates[1:] {
```
i0rek, line 847: I changed this to look similar to the other check. Just cosmetic changes though.

dnephin (approved):
LGTM!

With the caveat that I clearly don't have a deep understanding of TLS, so I can't be sure that this construction of root -> intermediate matches what we expect.  It seems reasonable from what I know. If the test failed without this change, and this construction matches the HCP TLS chain, then I think we're good.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/442235.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 13238dbab685e556815d923f3137ae050e3c58e9 onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/442235)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 13238dbab685e556815d923f3137ae050e3c58e9 onto `release/1.9.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/442235)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 13238dbab685e556815d923f3137ae050e3c58e9 onto `release/1.8.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/442235)

PR #10985, opened 2021-09-03
acl: remove legacy ACL replication

dnephin:
This is another step toward removing the legacy ACL system for 1.11.  Since replication uses the legacy RPC endpoints it was necessary to remove this first before converting the legacy RPC endpoints into errors.

PR #10986, opened 2021-09-03
acl: remove legacy ACL RPC - part 1

dnephin:
Branched from #10985
Likely best viewed by individual commit.

This PR starts the removal of RPC endpoints by replacing them with an implementation that returns an error. There are many tests that use `ACL.Apply`, so I've started to update the tests to use the new ACL system, but I'm going to split that over a few PRs.

This PR:
* removes some unused things
* makes `ACL.List` and `ACL.Get` an error
* updates 4 tests in `agent` package, and a bunch of tests in `agent/consul` to use the new RPC methods

One important thing to look for is the difference between ACL legacy rules and the new rules (documented here: https://www.consul.io/docs/security/acl/acl-legacy#rule-syntax-changes).

In some tests it was fine to leave the rule unmodified , and use the new strict "exact match" for the rules, but in other tests I had to change the rules to `_prefix` rules to get the tests to pass. If any of the `_prefix` changes look wrong, I can investigate more to see if it's masking a bug.


agent/consul/fsm/commands_oss.go
```diff
@@ -260,24 +260,7 @@ func (c *FSM) applyACLOperation(buf []byte, index uint64) interface{} {
 			return err
 		}
 		return enabled
-	case structs.ACLBootstrapNow:
```
rboyer, line 263: Should we leave this case in place so we can create a reasonable error message during raft replay?

dnephin: Ya, that sounds good. I'll do the same for `structs.ACLForceSet`, but I'll probably inline the values so we can still remove the old constants.

agent/consul/catalog_endpoint_test.go
```diff
@@ -2717,27 +2693,15 @@ func testACLFilterServer(t *testing.T) (dir, token string, srv *Server, codec rp
...
-	arg := structs.ACLRequest{
-		Datacenter: "dc1",
-		Op:         structs.ACLSet,
-		ACL: structs.ACL{
-			Name: "User token",
-			Type: structs.ACLTokenTypeClient,
-			Rules: `
-service "foo" {
+	rules := `
+service_prefix "foo" {
```
rboyer, line 2697: I suspect many of these `service "foo"` ones could be replaced by `service "foo" {} service "foo-sidecar-proxy" {}` instead of the foo prefix if you thought that was valuable.

Or you could switch these specific tests to use a "foo" ServiceIdentity (which would give it the right perms for doing a connect test).

agent/consul/health_endpoint_test.go
```diff
@@ -993,26 +993,15 @@ func TestHealth_ServiceNodes_ConnectProxy_ACL(t *testing.T) {
...
-	arg := structs.ACLRequest{
-		Datacenter: "dc1",
-		Op:         structs.ACLSet,
-		ACL: structs.ACL{
-			Name: "User token",
-			Type: structs.ACLTokenTypeClient,
-			Rules: `
-service "foo" {
+	rules := `
+service_prefix "foo" {
```
rboyer, line 997: This could also be the foo/foo-sidecar-proxy or ServiceIdentity if you prefer.

rboyer (approved):

agent/consul/catalog_endpoint_test.go
dnephin: Sounds good, I'll do a pass through some of these to see if I can make the rules a bit more strict.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/460212.

PR #9456, opened 2020-12-22
config: Use DeprecatedConfig struct for deprecated config fields

dnephin:
We have deprecated config fields in the past, but I believe this approach has a number of advantages.

1. All of the deprecation logic (copying the value to the new location, and creating a warning) is together in the same place. Previously it was scattered in 3 or more places, which resulting in a lot of inconsistency.
2. Deprecated fields no longer need to be part of the `Config`. The `Config` struct can be used exclusively as an intermediate for merging. The new `DeprecatedConfig` and `decodeTarget` structs allow us to decode into `DeprecatedConfig` and immediately copy to the new config location.
3. Building a `RuntimeConfig` from a `Config` is more straightforward because deprecated config is handled before that step.

PR #10988, opened 2021-09-03
acl: isolate deprecated config and warn when they are used

dnephin:
Branched from #9456

This PR moves deprecated ACL config fields to the `DeprecatedConfig` struct, and adds a warning when they are set. Since we have equivalent fields in the new ACL system, it did not seem necessary to error when one of these fields are set.

### DIFF
--- a/agent/config/runtime_test.go
+++ b/agent/config/runtime_test.go
@@ -5920,6 +5920,7 @@ func TestLoad_FullConfig(t *testing.T) {
 		deprecationWarning("acl_token", "acl.tokens.default"),
 		deprecationWarning("acl_master_token", "acl.tokens.master"),
 		deprecationWarning("acl_replication_token", "acl.tokens.replication"),
+		deprecationWarning("enable_acl_replication", "acl.enable_token_replication"),
 		`bootstrap_expect > 0: expecting 53 servers`,
 	}
 	expectedWarns = append(expectedWarns, enterpriseConfigKeyWarnings...)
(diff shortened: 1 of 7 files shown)


karl-cardenas-coding (approved):
Doc changes look fine

agent/config/runtime_test.go
```diff
@@ -1633,10 +1633,11 @@ func TestLoad_IntegrationWithFlags(t *testing.T) {
 		expectedWarnings: []string{`The 'acl_datacenter' field is deprecated. Use the 'primary_datacenter' field instead.`},
 	})
 	run(t, testCase{
-		desc: "acl_replication_token enables acl replication",
```
rboyer, line 1636: It might make sense to create a non-deprecated version of this test to verify that setting `acl.tokens.replication` does NOT enable token replication? Mostly to verify that the deprecated field logic doesn't apply to new configs, just old configs.

rboyer (approved):
LGTM with one suggestion

agent/config/runtime_test.go
dnephin: good call, I'll add that

mkeeler:
Its super nice to see this deprecated config handling. It's going to make some of my future work much easier.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/460068.

PR #10996, opened 2021-09-07
Update repo to use go:embed

kisunji:
### Summary
1. Replace bindata packages with stdlib go:embed
2. Modernize some `uiserver` code with newer interfaces introduced in go 1.16 (mainly working with fs.File instead of http.File) - functionality should stay the same
3. Remove steps that are no longer used from our build files
4. Add Github Action to detect diffs in `agent/uiserver/dist`

### Tip for reviewers
The `agent/uiserver/dist` files are compiled from `make ui` and can be minimized / treated as compiled assets


eculver:
> Should we prevent users from committing to `agent/uiserver/dist`?

I think so. We could add `dist` to `.gitignore`. This would also mean "Add check in case `agent/uiserver/dist` has a diff but not `ui`" is no longer necessary. Since building the binary will now depend on files in `dist`, we can require `make ui` to be a prerequisite of the build targets (`consul-docker` and `ui-docker`). 

What do you think of this approach?

johncowen:
FYI from what I remember, there was a distinct desire keep the ui artifact commited in the git repo so that folks didn't have to rebuild the UI every time they wanted to build the binary. I'm not sure if that still stands though.

I'm not entirely sure when reading the comments here what goes in `dist`, or the implications of most of this PR 😅 , so ignore me if I've missed whats going on here, or ping me offline if you want, just thought I'd mention that ^ so you knew.

kisunji:
> 
> 
> > Should we prevent users from committing to `agent/uiserver/dist`?
> 
> I think so. We could add `dist` to `.gitignore`. This would also mean "Add check in case `agent/uiserver/dist` has a diff but not `ui`" is no longer necessary. Since building the binary will now depend on files in `dist`, we can require `make ui` to be a prerequisite of the build targets (`consul-docker` and `ui-docker`).
> 
> What do you think of this approach?

I think adding `dist` to .gitignore would prevent our CI from also committing to it. Is there a way around that?
Otherwise I thought a pre-receive hook to diff that directory and reject commits might be useful (but unsure if used across hashicorp).

I'll apply the changes for the build targets!

kisunji:
> 
> 
> FYI from what I remember, there was a distinct desire keep the ui artifact commited in the git repo so that folks didn't have to rebuild the UI every time they wanted to build the binary. I'm not sure if that still stands though.
> 
> I'm not entirely sure when reading the comments here what goes in `dist`, or the implications of most of this PR 😅 , so ignore me if I've missed whats going on here, or ping me offline if you want, just thought I'd mention that ^ so you knew.

Yeah I had a mini discussion around that on slack as well. We will continue using CI to commit the `ui` artifacts but as the raw files under `agent/uiserver/dist` instead of being converted to bytes and dumped into a `bindata_assetfs.go` file.

kisunji:
> 
> Since building the binary will now depend on files in `dist`, we can require `make ui` to be a prerequisite of the build targets (`consul-docker` and `ui-docker`).
> 

I thought about this and realized the `dist` files will always be present after this PR is merged. I think the added runtime cost of `ui` as a prerequisite is not worth the check (`go build` will fail at compile time anyways if `dist` doesn't exist)

agent/uiserver/uiserver.go
```diff
@@ -92,20 +92,20 @@ func (h *Handler) ReloadConfig(newCfg *config.RuntimeConfig) error {
...
-		sub, err := fs.Sub(assets, "dist")
+		sub, err := fs.Sub(dist, "dist")
 		if err != nil {
 			return nil, err
 		}
-		fsys = http.FS(sub)
+		fsys = sub
 	} else {
-		fsys = http.Dir(cfg.UIConfig.Dir)
+		fsys = os.DirFS(cfg.UIConfig.Dir)
```
kisunji, line 104: Effectively the same behaviour, relying on `os.Open`:
[`http.Dir`](https://cs.opensource.google/go/go/+/refs/tags/go1.16.7:src/net/http/fs.go;l=71) vs [`os.DirFS`](https://cs.opensource.google/go/go/+/refs/tags/go1.16.7:src/os/file.go;l=636)

agent/uiserver/uiserver.go
```diff
@@ -92,20 +92,20 @@ func (h *Handler) ReloadConfig(newCfg *config.RuntimeConfig) error {
...
 	}
 
 	// Render a new index.html with the new config values ready to serve.
-	buf, info, err := h.renderIndex(cfg, fsys)
+	buf, err := h.renderIndexFile(cfg, fsys)
```
kisunji, line 108: we were passing around `[]bytes` and `FileInfo` to reconstruct a file later; refactored to do that work in this func and pass pointer to `fs.File` around

agent/uiserver/uiserver.go
```diff
@@ -196,33 +195,34 @@ func concatFile(buf *bytes.Buffer, file string) error {
...
-	f, err := fs.Open("/index.html")
+	f, err := fsys.Open("index.html")
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 	defer f.Close()
 
-	content, err := ioutil.ReadAll(f)
+	content, err := io.ReadAll(f)
```
kisunji, line 206: ioutil deprecated; same call under the hood

agent/uiserver/buffered_file.go
```diff
@@ -1,67 +1,33 @@
...
-	"bytes"
-	"errors"
-	"os"
-	"time"
+	"io"
+	"io/fs"
 )
 
-// bufferedFile implements http.File and allows us to modify a file from disk by
+// bufferedFile implements fs.File and allows us to modify a file from disk by
```
kisunji, line 8: `fs.File` is a smaller interface than `http.File`, making it more composable in the future

agent/uiserver/README.MD
```diff
@@ -0,0 +1,7 @@
+# uiserver
```
eculver, line 1: suggest renaming the file to `README.md` (lowercase extension) for consistency.

.circleci/config.yml
```diff
@@ -696,8 +679,8 @@ jobs:
               git config --local user.name "hc-github-team-consul-core"
 
               short_sha=$(git rev-parse --short HEAD)
-              git add agent/uiserver/bindata_assetfs.go
-              git commit -m "auto-updated agent/uiserver/bindata_assetfs.go from commit ${short_sha}"
+              git add agent/uiserver/dist/
```
dnephin, line 682: Now that we have multiple files and directories, not just a single file, I think we'll need to first `git rm -r agent/uiserver/dist` before we build anything to remove any stale files, right?

Otherwise we'll never be able to prune files. And I suspect since many of the files have fingerprints in the name we'd end up with a new copy of all the files every time.

Or is that remove done somewhere else that I missed?

In the removed job we have `mv pkg/packages/consul-ui/dist pkg/web_ui`. Do we need something similar to move these files over?

dnephin (commented):
Thanks for working on this. Lots of nice cleanup here! I left a couple questions inline. 

Not strictly necessary, but I wonder if it would be possible to omit the committed artifact from this PR so that only the automation commits it once this is merged into main. We could also test that flow by editing the `.circleci/config.yml` and having it run on some branch to preview what that will look like.

My other main question is about local workflow. I'm not at all familiar with building the UI locally. @johncowen and @kaxcode , would you like to try out this branch locally to make sure it still behaves the way you would expect for local dev?

build-support/docker/Build-UI.dockerfile
```diff
@@ -1,7 +1,7 @@
 ARG ALPINE_VERSION=3.13
 FROM alpine:${ALPINE_VERSION}
 
-ARG NODEJS_VERSION=14.16.1-r1
+ARG NODEJS_VERSION=14.17.6-r0
```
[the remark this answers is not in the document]
kisunji: I was having issues running `make ui` locally and the error messages hinted that I should bump the version. But now that I think about it I'm wondering how CI was building it fine 🤔

kisunji:
> Not strictly necessary, but I wonder if it would be possible to omit the committed artifact from this PR so that only the automation commits it once this is merged into main. We could also test that flow by editing the `.circleci/config.yml` and having it run on some branch to preview what that will look like.
> 
That makes good sense! Will cut the noise out of this PR too

build-support/docker/Build-UI.dockerfile
```diff
@@ -1,7 +1,7 @@
 ARG ALPINE_VERSION=3.13
```
david-yu, line 1: Sorry I couldn't help but lurk, it looks like we probably want to bump to 3.14 for all Dockerfiles since 3.13 has CVEs

.circleci/config.yml
kisunji: Ah good catch! I thought my changes in `build-support/functions/20-build.sh` would move the files but we don't run any `make` commands that end up calling that function

build-support/functions/30-release.sh
```diff
@@ -465,25 +465,7 @@ function build_release {
...
-      if is_set "${do_tag}"
-      then
-         git add "${sdir}/agent/uiserver/bindata_assetfs.go"
-         if test $? -ne 0
-         then
-            err "ERROR: Failed to git add the assetfs file" 
-            return 1
-         fi
-      fi
+      status "UI Built with Version: $(ui_version "${sdir}/agent/uiserver/dist/index.html")"
```
kisunji, line 468: todo: missing git add step here

.circleci/config.yml
kisunji: Now I'm a little confused as to how those build-support/ scripts relate to our CircleCI workflows... 😕

build-support/docker/Build-UI.dockerfile
kisunji: More eyes always appreciated 👍

build-support/docker/Build-UI.dockerfile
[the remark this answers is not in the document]
johncowen: Hey @kisunji 

First thing I noticed here was related to this change, I'm not sure if things have changed upstream since you made this change, but I had to move the `npm` installation to use a different name/version also in order to get `make ui` working. I was gonna PR this separately, but now not so sure. Maybe we should bump to alpine 3.14 as a completely different PR (are we gonna need that bump on 1.10.x/1.9.x branch etc?), kinda up to you.

build-support/functions/20-build.sh
```diff
@@ -139,59 +139,14 @@ function build_ui {
    # Copy UI over ready to be packaged into the binary
    if test ${ret} -eq 0
    then
-      rm -rf ${1}/pkg/web_ui
-      mkdir -p ${1}/pkg
-      cp -r ${1}/ui/dist ${1}/pkg/web_ui
+      rm -rf ${1}/agent/uiserver/dist
```
johncowen, line 142: Not 100% sure whats happening here, but I get a stray:

Screenshot 2021-09-27 at 11 09 37

After running `make ui` 🤔 not sure how thats happened? Oh wait I guess that REMOVEME.txt must have been commited somewhere?

johncowen (commented):
Hey @kisunji 

Thanks for tackling this, exciting stuff!

I spotted a few things going through the ropes with this, so I left some inline comments here. But also, when I came to run `make ui` then `make dev-docker` and then running the resultant image, there seems to be a problem with the content that needs serving up.

When I try to curl the `ui/` endpoint I get:

```
curl: (18) transfer closed with 111 bytes remaining to read
```

So there's something amiss still.

Kinda semi-related but I have a thing I did a while ago for e2e testing basic Consul UI (in real not mock Consul) so we could make sure 'this is working'. Would that be useful here? I could resurrect it.

build-support/functions/20-build.sh
kisunji: Yep! The `dist/` directory needs to exist for this PR to compile but I didn't check in the artifacts so it's a placeholder just so that git doesn't ignore `dist/`

build-support/docker/Build-UI.dockerfile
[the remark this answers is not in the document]
kisunji: @johncowen if you can make a separate PR for the version bumps (including the alphine) it would be appreciated! My PR isn't dependent on these changes at all; I just needed this version to build locally.

kisunji:
> When I try to curl the `ui/` endpoint I get:
> 
> ```
> curl: (18) transfer closed with 111 bytes remaining to read
> ```
>
> (with all the correct HTML content above that error)
> 
> So there's something amiss still.

Thanks for the bug report! I'll try to replicate it on my end

> 
> Kinda semi-related but I have a thing I did a while ago for e2e testing basic Consul UI (in real not mock Consul) so we could make sure 'this is working'. Would that be useful here? I could resurrect it.

Yeah that would be much appreciated! There are some backend tests to make sure that certain features of our `uiserver` works (redirects, custom files, etc) but I wasn't sure how comprehensive it was

.circleci/config.yml
```diff
@@ -672,22 +672,6 @@ jobs:
...
-      - attach_workspace:
-          at: ./pkg
-      - run: mv pkg/packages/consul-ui/dist pkg/web_ui # 'make static-assets' looks for the 'pkg/web_ui' path
-      - run: make static-assets
-      - persist_to_workspace:
-          root: .
-          paths:
-            - ./agent/uiserver/bindata_assetfs.go
-      - run: *notify-slack-failure
-
```
kisunji, line 690: This job was responsible for buiding bindata_assetfs.go and persisting it to the correct workspace path.

kisunji: Note that `pkg/packages/consul-ui/dist` is assumed to exist (it contains all the compiled ui files)

.circleci/config.yml
```diff
@@ -700,7 +684,12 @@ jobs:
       - attach_workspace:
           at: .
       - run:
-          name: commit agent/uiserver/bindata_assetfs.go if there are UI changes
+          name: move compiled ui files to agent/uiserver
+          command: |
+            rm -rf agent/uiserver/dist
+            mv ui/packages/consul-ui/dist agent/uiserver
```
kisunji, line 690: This added step is now responsible for moving the compiled ui files to the correct workspace path

.github/workflows/build.yml
```diff
@@ -101,12 +101,8 @@ jobs:
           echo "consul binary type is ${CONSUL_BINARY_TYPE}"
           echo "consul copyright year is ${CONSUL_COPYRIGHT_YEAR}"
           cd ui && make && cd ..
-          mkdir pkg
-          mv ui/packages/consul-ui/dist pkg/web_ui
-
-      - name: Build static-assets
-        run: make static-assets
-
```
kisunji, line 109: Previously we were making an intermediate well-known path `pkg/web_ui` which was used by `make static-assets` which dumped the compiled files into `agent/uiserver`.

Now that we are removing `make static-assets` we can move files from the ui directory straight into `agent/uiserver`

.changelog/10996.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+ui: removed external dependencies for serving UI assets in favor of Go 1.16's native embed capabilities
```
kisunji, line 2:
```suggestion
ui: removed external dependencies for serving UI assets in favor of Go's native embed capabilities
```

.circleci/config.yml
```diff
@@ -1101,20 +1090,12 @@ workflows:
...
             - frontend-cache
-      - build-static-assets:
-          requires:
-            - ember-build-prod
       - publish-static-assets:
-          filters:
-            branches:
-              only:
-                - main
-                - /release\/\d+\.\d+\.x$/
```
kisunji, line 1112: filters were never relevant because the `requires` chain resolves to `frontend-cache` which only ran on `main`

.github/workflows/embedded-asset-checker.yml
```diff
@@ -0,0 +1,50 @@
...
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0 # by default the checkout action doesn't checkout all branches
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 14
+      - name: Check for agent/uiserver/dist dir change in diff
+        run: |
+          dist_files=$(git --no-pager diff --name-only HEAD "$(git merge-base HEAD "origin/${{ github.event.pull_request.base.ref }}")" -- agent/uiserver/dist)
+          
+          if [ -n "$dist_files" ]; then
```
rboyer, line 33: You could avoid the indent with:
```
if [[ -z "${dist_files}" ]]; then
  exit 0
fi
<rest unindented>
```

.github/workflows/embedded-asset-checker.yml
```diff
@@ -0,0 +1,50 @@
...
+            cd ui
+            make
+            cd ..
+
+            # diff causes this step to exit code 1 if there is a diff so we append '|| true' to the end
+            diffs=$(diff -r --brief ui/packages/consul-ui/dist agent/uiserver/dist || true)
+
+            if [ -n "$diffs" ]; then
```
rboyer, line 41: You can also do something like:
```
if ! diff -r --brief <stuff> ; then
   # what to do when "diff" exits non-zero because of the ! earlier
fi
```
which avoids the intermediate var

kisunji: I think I tried a variant but Github Actions kept exiting with code 1 when evaluating `diff`. Will try your suggestion and see if it's different

rboyer (approved):
Some non-blocking suggestions but overall LGTM assuming you can get the new check to be green.

.github/workflows/embedded-asset-checker.yml
```diff
@@ -0,0 +1,50 @@
+# This workflow detects if there is a diff in the `agent/uiserver/dist` directory
+# which is used by Consul to serve its embedded UI. If there is a diff, this action
+# will build the ui again and diff `agent/uiserver/dist` with the newly compiled files
+# to ensure that the updates are valid.
+#
+# Note that since the latest commit SHA is embedded in index.html as CONSUL_VERSION
+# you must always compile the assets from the HEAD of your branch or this workflow fails.
+
+name: Embedded Asset Checker
```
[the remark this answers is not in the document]
kisunji: I'm intentionally failing it for the moment, but a label might be useful.

rboyer (approved):

PR #10618, opened 2021-07-14
docs: add deprecation version for ports.grpc settings

dnephin:
Follow up to #10588

Adds the version when these settings will be deprecated.

PR #10999, opened 2021-09-08
Revert config xds_port

dnephin:
Reverts #10588 and #10618

I learned an upcoming change will actually make use of this port for non-xds services, so the old grpc_port name will end up being correct.

This PR preserves the changes in 4ad80ccee331030c32b669fda9b14d1eacb1628f to use a stable API for the envoy command.


dnephin:
I think I should keep the fix for the envoy command, I'll restore that.

karl-cardenas-coding (approved):
👍🏼  from a docs perspective

agent/xds/server.go
```diff
@@ -548,15 +548,14 @@ func tokenFromContext(ctx context.Context) string {
 	return ""
 }
 
-// NewGRPCServer creates a grpc.Server, registers the Server, and then returns
-// the grpc.Server.
-func NewGRPCServer(s *Server, tlsConfigurator *tlsutil.Configurator) *grpc.Server {
```
freddygv, line 553: Keeping the old name of `NewGRPCServer` could be nice

freddygv (approved):
LGTM, one minor suggestion below

PR #11027, opened 2021-09-14
Remove udp_answer_limit (deprecated in 1.0.7)

### LINKED ISSUE #10694: Consul setting a_record_limit still requires deprecated setting udp_answer_limit to be set in some scenarios
After spending some time diagnosing issues where an external public facing service was registered in Consul (AWS Service) and us not being able to establish a TCP connection via the consul DNS name registered we noticed that CNAME wasn't always returning the resolved IP Address.   This ONLY occurred inside a container and NEVER on the host itself (dns resolution worked fine).   We found that setting this deprecated setting https://www.consul.io/docs/agent/options#udp_answer_limit to "99" resolved the issue BUT according to the Consul docs were should be using https://www.consul.io/docs/agent/options#a_record_limit which has no limit.   Looks like either the behavior isn't as expected or the setting scheduled for deprecation should be kept around.

Happy to provide detailed dumps if necessary.

Thanks
Ian

jkirschner-hashicorp:
Fix #10694 by removing [`udp_answer_limit`](https://www.consul.io/docs/agent/options#udp_answer_limit).

Consul's documentation states that the `udp_answer_limit` was deprecated and replaced by [`a_record_limit`](https://www.consul.io/docs/agent/options#a_record_limit) in Consul 1.0.7. However, these two settings were treated as independent options in the source which could both cause a message to be truncated.

`a_record_limit` affects the number of records contained in the answer section of an A, AAAA, or ANY DNS response (independent of network protocol). It won't affect DNS behavior unless the user opts in.

`udp_answer_limit` also limited the number of records contained in the answer section, but only applied to DNS requests without EDNS over UDP (not TCP). Because it defaulted to 3 (which isn't stated in our docs), it would affect DNS behavior even if the user had not opted in, despite being listed as a deprecated option. This can understandably confuse users, such as in #10694 in which a user found this deprecated setting was still necessary.

Therefore, to avoid future user confusion and unexpected DNS behavior, remove the `udp_answer_limit` option.

Additionally, the DNS code has been made slightly more DRY and commented.

Easiest to review by commit.

**Question to Reviewer**
- How do we properly remove a previously deprecated option? (e.g., how do we communicate this?)

### DIFF
--- a/.changelog/11027.txt
+++ b/.changelog/11027.txt
@@ -1,0 +1,3 @@
+```release-note:breaking-change
+dns: Remove dns_config.udp_answer_limit agent config option, which was had side effects (see GH-10694) when unspecified despite being deprecated and replaced by dns_config.a_record_limit in v1.0.7.
+```
--- a/agent/dns.go
+++ b/agent/dns.go
@@ -25,7 +25,6 @@ import (
 	"github.com/hashicorp/consul/agent/structs"
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/consul/ipaddr"
-	"github.com/hashicorp/consul/lib"
 	"github.com/hashicorp/consul/logging"
 )
 
@@ -48,11 +47,6 @@ var DNSSummaries = []prometheus.SummaryDefinition{
 }
 
 const (
-	// UDP can fit ~25 A records in a 512B response, and ~14 AAAA
-	// records. Limit further to prevent unintentional configuration
-	// abuse that would have a negative effect on application response
-	// times.
-	maxUDPAnswerLimit        = 8
 	maxRecurseRecords        = 5
 	maxRecursionLevelDefault = 3
 
@@ -83,7 +77,6 @@ type dnsConfig struct {
 	RecursorTimeout  time.Duration
 	Recursors        []string
 	SegmentName      string
-	UDPAnswerLimit   int
 	ARecordLimit     int
 	NodeMetaTXT      bool
 	SOAConfig        dnsSOAConfig
@@ -161,7 +154,6 @@ func GetDNSConfig(conf *config.RuntimeConfig) (*dnsConfig, error) {
 		RecursorStrategy:   conf.DNSRecursorStrategy,
 		RecursorTimeout:    conf.DNSRecursorTimeout,
 		SegmentName:        conf.SegmentName,
-		UDPAnswerLimit:     conf.DNSUDPAnswerLimit,
 		NodeMetaTXT:        conf.DNSNodeMetaTXT,
 		DisableCompression: conf.DNSDisableCompression,
 		UseCache:           conf.DNSUseCache,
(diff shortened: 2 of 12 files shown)


agent/dns.go
```diff
@@ -25,7 +25,7 @@ import (
 	"github.com/hashicorp/consul/agent/structs"
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/consul/ipaddr"
-	"github.com/hashicorp/consul/lib"
+	// "github.com/hashicorp/consul/lib"
```
jkirschner-hashicorp, line 28: Unused library - I need to remove this

agent/dns.go
```diff
@@ -1080,14 +1073,38 @@ func dnsBinaryTruncate(resp *dns.Msg, maxSize int, index map[string]dns.RR, hasE
 // trimTCPResponse limit the MaximumSize of messages to 64k as it is the limit
 // of DNS responses
 func trimTCPResponse(req, resp *dns.Msg) (trimmed bool) {
-	hasExtra := len(resp.Extra) > 0
-	// There is some overhead, 65535 does not work
+	// NOTE: 65523 is a magic number that has been changed several times in the
```
jkirschner-hashicorp, line 1076: Note to reviewer: I did not attempt to fix the magic numbers, but I did want to describe them so that someone with more DNS domain knowledge would have an easier time fixing these numbers in the future.

agent/dns.go
```diff
@@ -1099,50 +1116,47 @@ func trimTCPResponse(req, resp *dns.Msg) (trimmed bool) {
...
 	// Update to the maximum edns size
 	if edns := req.IsEdns0(); edns != nil {
-		if size := edns.UDPSize(); size > uint16(maxSize) {
-			maxSize = int(size)
+		if size := edns.UDPSize(); size > uint16(maxSizeTotal) {
+			maxSizeTotal = int(size)
 		}
 	}
 
+	// NOTE: 7 is a magic number added in commit 4f5d502 to stop an "overflow
```
jkirschner-hashicorp, line 1136: Note to reviewer: I did not attempt to fix the magic numbers, but I did want to describe them so that someone with more DNS domain knowledge would have an easier time fixing these numbers in the future.

.changelog/11027.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
```
jkirschner-hashicorp, line 1: Should this be `release-note:breaking-change` instead for a removal of a deprecated parameter?

.changelog/11027.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+dns: Remove dns_config.udp_answer_limit agent config option, which was had side effects (see GH-10694) when unspecified despite being deprecated and replaced by dns_config.a_record_limit in v1.0.7.
```
jkirschner-hashicorp, line 2: should this be `config:` instead of `dns:`?

.changelog/11027.txt
dnephin: Ya, that seems appropriate.

.changelog/11027.txt
dnephin: I think `dns` is good. It is related to config as well, but I think `dns` is more specific

jmurret:

PR #11031, opened 2021-09-14
WIP: Cross-Reference API Endpoints & CLI Commands; Include Command ACL Info

### LINKED ISSUE #3134: Add ACL information to CLI commands
Now that we have full ACL support by default, many of the CLI commands need a token with specific privileges to operate. We should add the same ACL tables that we have in the API docs to the CLI docs.

jkirschner-hashicorp:
Intended to close #3134. Meant to accomplish:
- Just like we give additional info (e.g., required ACLs) for using HTTP API endpoints, we should provide the same info for CLI commands
- CLI commands and HTTP API endpoints should cross-reference each other in the docs, in case the user (1) is looking for one but ends up on the other, such as through a Google search, or (2) wants to convert from one usage to another.

We should align on a consistent approach for one command/endpoint before applying the approach to all commands/endpoints.

Steps:
- [ ] Align on consistent approach for one endpoint
- [ ] Align on how to make this information easy to manage (e.g., include some of the info in partials?)
- [ ] Deploy approach across all commands/endpoints (based on a spreadsheet I've created mapping CLI commands, HTTP API endpoints, and required ACLs)

### DIFF
--- a/website/content/api-docs/acl/index.mdx
+++ b/website/content/api-docs/acl/index.mdx
@@ -24,9 +24,19 @@ and requires all Consul servers to be upgraded in order to operate.
 This provides a mechanism to bootstrap ACLs without having any secrets present in Consul's
 configuration files.
 
+**vv Cross ref option 1: text**
+
 | Method | Path             | Produces           |
 | ------ | ---------------- | ------------------ |
 | `PUT`  | `/acl/bootstrap` | `application/json` |
+
+The corresponding CLI command is [`consul acl bootstrap`](/commands/acl/bootstrap).
+
+**vv Cross ref option 2: table**
+
+| Method | Path             | Produces           | Corresponding CLI Command                         |
+| ------ | ---------------- | ------------------ | ------------------------------------------------- |
+| `PUT`  | `/acl/bootstrap` | `application/json` | [`consul acl bootstrap`](/commands/acl/bootstrap) |
 
 The table below shows this endpoint's support for
 [blocking queries](/api/features/blocking),
@@ -387,10 +397,7 @@ $ curl \
 
 ## OIDC Authorization URL Request
 
-<EnterpriseAlert>
-  {' '}
-  This is an enterprise only endpoint.
-</EnterpriseAlert>
+<EnterpriseAlert> This is an enterprise only endpoint.</EnterpriseAlert>
 
 This endpoint was added in Consul 1.8.0 and is used to obtain an authorization
 URL from Consul to start an [OIDC login flow](/docs/acl/auth-methods/oidc).
@@ -466,10 +473,7 @@ $ curl \
 
 ## OIDC Callback
 
-<EnterpriseAlert>
-  {' '}
-  This is an enterprise only endpoint.
-</EnterpriseAlert>
+<EnterpriseAlert> This is an enterprise only endpoint.</EnterpriseAlert>
 
 This endpoint was added in Consul 1.8.0 and is used to exchange an OIDC
 authorization code for an OIDC ID Token. The ID token will in turn be exchanged
--- a/website/content/commands/acl/bootstrap.mdx
+++ b/website/content/commands/acl/bootstrap.mdx
@@ -7,12 +7,83 @@ page_title: 'Commands: ACL Bootstrap'
 
 Command: `consul acl bootstrap`
 
+Corresponding HTTP API Endpoint:
+[`[GET] /v1/acl/bootstrap`](/api-docs/acl#bootstrap-acls)
+
+**^^ Cross ref option 1: inline with command**
+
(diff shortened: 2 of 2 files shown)


jkirschner-hashicorp:
@karl-cardenas-coding, @trujillo-adam, @blake : we should align on a consistent approach for one command/endpoint before I proceed any further. I've created some possible approaches for you to react to (that exist in this PR's diff and have been rendered into the images below).

In all images, the light blue box shows the content that has been added / differs between options. Let me know what your thoughts on which options are best or other options to try.

### Cross-referencing HTTP API endpoints from command docs

image

**TLDR: pick option 1?**

I think option 1 is much more scannable, though it won't be visually consistent with its counterpoint cross-ref on the HTTP API endpoint docs page (see options below). Still, I think option 1 is better.

### Cross-referencing CLI Commands from HTTP API endpoints docs

image

**TLDR: pick option 1? But if we pick option 2 instead, we might want to pick option 2 above as well.**

Option 1 is more scannable. Unfortunately, by adding a column, there's less space for some other columns (like the endpoint). For some endpoints, the content may wrap onto two lines... though I'm not sure that's a big deal. Maybe there's a more concise, but still clear, column header than "Corresponding CLI Command".

Option 2 can be made consistent with its counterpoint cross-ref on the command docs page (see option 2 above).

### Including ACL Info in CLI command docs

image

image

image

**TLDR: maybe the table from option 3, but decide whether/how we mention blocking queries, agent caching, and other consistency modes (`consistent`) not supported by commands but may be by the corresponding HTTP API endpoint?**

Options 1 and 2: I wanted to follow the same structure as on the HTTP API endpoint docs, but if we only have one column (because some of the other columns don't apply) ... it looks weird. I suppose we could just have a text line that says "ACL Required: `none`" instead of a table ... (EDIT: which is now option 5).

Option 3: technically, some commands support the `stale` consistency mode through the `-stale` flag. So this gives us a second column. Maybe this is the best option, since it's scannable and the most complete for what the command itself supports? (Options 1,2,5 don't provide a means to show consistency modes where supported.)

Option 4: gives the most complete information, but I find it very confusing to mix what's directly supported by the command versus what would be supported if the corresponding HTTP API endpoint was used instead. Included to have some variety to consider, but I don't think this is a good option.

Options 2-4 try different approaches to mention blocking queries, agent caching, and other consistency modes (`consistent`) that may be available via the HTTP API endpoint, but aren't via command. Option 1 chooses not to mention any of this. Which approach do you think makes the most sense for mentioning (or not) this specific info?

Option 5: simple, but not visually consistent with how the HTTP API endpoint docs present this info. Also doesn't include consistency mode.

karl-cardenas-coding:
@jkirschner-hashicorp 

I really like the options you have put together and thank you for making it easier for us all to review. Here are my thoughts.

Cross-referencing HTTP API endpoints from command docs -> Option 1

Cross-referencing CLI Commands from HTTP API endpoints docs -> Option 1

Including ACL Info in CLI command docs -> Option 3

I agree with all the points you made. The ACL info was a bit tougher to choose, but I think adding too much detail increases the complicity and could result in practitioners getting intimidated. Option 3 seems to be a good sweet spot imo.

jkirschner-hashicorp:
@trujillo-adam, @blake : any different opinions? It seems like @karl-cardenas-coding and I are both thinking:

Cross-referencing HTTP API endpoints from command docs -> Option 1

Cross-referencing CLI Commands from HTTP API endpoints docs -> Option 1

Including ACL Info in CLI command docs -> Option 3

trujillo-adam:
Regarding the *Cross-referencing HTTP API endpoints from command docs* change, I like the endpoint cross-ref in a paragraph under the description because my assumption is that the users' primary goal is get information about the command; the API endpoint is secondary. 

Regarding the *Cross-referencing CLI Commands from HTTP API endpoints docs* change, I generally like adding the command to the table, but I wonder if we should take this opportunity to reconsider which columns we're standardizing on.  would want to know what kind fo payloads a PUT or POST takes and what parameters are available, and which params are required. Does a table even make sense considering that there is only one row of data? I'me sure we've done it in the past, but it was probably driven by the context, i.e., done to maintain consistency within the same page of content. It's great that you are adding this info to the API refs, but it should be implemented as part of a larger improvement strategy. We should have that discussion outside of this PR.

Regarding the *Including ACL Info in CLI command docs* change, while I think we should be using tables more often, I am not sure that a single-row of information in table format is better than a bulleted list. That said, I am onboard with providing complete information a la option #4.

jkirschner-hashicorp:
Closing this unmerged. The work started here was carried forward in two other PRs since merged:
- https://github.com/hashicorp/consul/pull/11984
- https://github.com/hashicorp/consul/pull/12004

PR #11037, opened 2021-09-15
Document HTTP Header manipulation options added in #10613

banks:
Questions for reviewers:

 1. I choose to only document `HTTPHeaderModifiers` struct in one place and cross link to avoid duplication and getting out of sync. I'm happy with that, although picking one of the three places was pretty arbitrary. I went with `service-router` since that is perhaps the most natural fit for discovering HTTP manipulation config but I don't think it matters much since it's cross linked from other places anyway. Does this seem reasonable or would either duplicating or creating a more specific place for shared structures make more sense?
 2. I wanted to add at least a couple of examples of use. For ingress I was in two minds about whether I should make a whole new example. On the plus side a separate example would keep each example more specifically focussed on each feature and use-case, on the other hand I can see it getting harder to find as we grow features if every one has a separate example. I tried it combined for now in the more complex example block but open to changing that if we think it's better to split the examples.

Notes:
* We've not yet scheduled the work to update the Kube CRD. It may well get done before 1.11 so we can remove the `yaml: false` and "Not supported in Kube" messages here, but I thought I'd put it up like this reflecting reality now in case that work doesn't get done before 1.11 and we forget. I'm open to changing that if we decide it's simple to add them to the CRDs.

### DIFF
--- a/website/content/docs/connect/config-entries/service-router.mdx
+++ b/website/content/docs/connect/config-entries/service-router.mdx
@@ -574,6 +574,69 @@ spec:
       description:
         'A list of HTTP response status codes that are eligible for retry.',
     },
+    {
+      yaml: false,
+      name: 'RequestHeaders',
+      type: 'HTTPHeaderModifiers: <optional>',
+      description: `A set of [HTTP-specific header modification rules](/docs/connect/config-entries/service-router#httpheadermodifiers)
+      that will be applied to requests routed to this service.
+      This cannot be used with a \`tcp\` listener.`,
+    },
+    {
+      yaml: false,
+      name: 'ResponseHeaders',
+      type: 'HTTPHeaderModifiers: <optional>',
+      description: `A set of [HTTP-specific header modification rules](/docs/connect/config-entries/service-router#httpheadermodifiers)
+      that will be applied to responses from this service.
+      This cannot be used with a \`tcp\` listener.`,
+    },
+  ]}
+/>
+
+
+### `HTTPHeaderModifiers`
+
+<ConfigEntryReference
+  topLevel={false}
+  yaml={false}
+  keys={[
+    {
+      hcl: false,
+      name: 'Unsupported',
+      type: '',
+      description: `HTTP Header modification is not yet supported in our Kubernetes CRDs.`,
+    },
+    {
+      yaml: false,
+      name: 'Add',
+      type: 'map<string|string>: optional',
+      description: `The set of header value keyed by header name to
+      add. If a header with the same (case-insensitive) name already
+      exists, the value set here will be appended and both will be present.
+      If Envoy is used as the proxy, the value may contain
+      [variable placeholders](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#custom-request-response-headers) for example
+      \`%DOWNSTREAM_REMOTE_ADDRESS%\` to interpolate dynamic request
+      metadata into the value added.`,
+    },
+    {
+      yaml: false,
+      name: 'Set',
+      type: 'map<string|string>: optional',
+      description: `The set of header value keyed by header name to
+      add. If one or more header values with the same (case-insensitive) name already exist,
+      the value set here will replace them all.
+      If Envoy is used as the proxy, the value may contain
+      [variable placeholders](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#custom-request-response-headers) for example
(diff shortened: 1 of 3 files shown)


website/content/docs/connect/config-entries/service-router.mdx
```diff
@@ -574,6 +574,69 @@ spec:
       description:
         'A list of HTTP response status codes that are eligible for retry.',
     },
+    {
+      yaml: false,
+      name: 'RequestHeaders',
+      type: 'HTTPHeaderModifiers: <optional>',
+      description: `A set of [HTTP-specific header modification rules](/docs/connect/config-entries/service-router#httpheadermodifiers)
```
blake, line 581:
```suggestion
      description: `A set of [HTTP-specific header modification rules](#httpheadermodifiers)
```
minor suggestion: since this references an object on the same page, you can remove the URL path and just leave the fragment.

website/content/docs/connect/config-entries/service-router.mdx
```diff
@@ -574,6 +574,69 @@ spec:
...
+      name: 'Unsupported',
+      type: '',
+      description: `HTTP Header modification is not yet supported in our Kubernetes CRDs.`,
+    },
+    {
+      yaml: false,
+      name: 'Add',
+      type: 'map<string|string>: optional',
+      description: `The set of header value keyed by header name to
+      add. If a header with the same (case-insensitive) name already
```
blake, line 614:
```suggestion
      description: `The set of header values to add, keyed by header name.
      If a header with the same (case-insensitive) name already
```

website/content/docs/connect/config-entries/service-router.mdx
```diff
@@ -574,6 +574,69 @@ spec:
...
+      [variable placeholders](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#custom-request-response-headers) for example
+      \`%DOWNSTREAM_REMOTE_ADDRESS%\` to interpolate dynamic request
+      metadata into the value added.`,
+    },
+    {
+      yaml: false,
+      name: 'Remove',
+      type: 'array<string>: optional',
+      description: `The set of header names to remove. Only headers
+      whose names are an <i>case-insensitive</i> exact match will be removed`,
```
blake, line 638:
```suggestion
      whose names are a <i>case-insensitive</i> exact match will be removed`,
```

blake (approved):
I only had a few minor comments. Otherwise LGTM!

website/content/docs/connect/config-entries/service-router.mdx
banks: I might leave this on the basis that this block is copy pasted between three config entries right now. If we remove the URL on this page it makes it harder to copy paste between them for future updates?

website/content/docs/connect/config-entries/service-router.mdx
```diff
@@ -574,6 +574,69 @@ spec:
...
+      name: 'Unsupported',
+      type: '',
+      description: `HTTP Header modification is not yet supported in our Kubernetes CRDs.`,
+    },
+    {
+      yaml: false,
+      name: 'Add',
+      type: 'map<string|string>: optional',
+      description: `The set of header values to add, keyed by header name.
+      If a header with the same (case-insensitive) name already
```
trujillo-adam, line 614:
```suggestion
      description: `The set of key/value pairs that specify header values to add. Use header names as keys. Header names are case-sensitive.
      If a header with the same name already
```
Just thinking about non-native English speakers--not sure if 'keyed by header name' would make sense. Also, case-sensitivity seems important enough to call out as its own statement.

website/content/docs/connect/config-entries/service-router.mdx
```diff
@@ -574,6 +574,69 @@ spec:
...
+      If Envoy is used as the proxy, the value may contain
+      [variable placeholders](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#custom-request-response-headers) for example
+      \`%DOWNSTREAM_REMOTE_ADDRESS%\` to interpolate dynamic request
+      metadata into the value added.`,
+    },
+    {
+      yaml: false,
+      name: 'Set',
+      type: 'map<string|string>: optional',
+      description: `The set of header value keyed by header name to
```
trujillo-adam, line 625: Same comment as line 613.

trujillo-adam (approved):
Looks good. I added a few minor suggestions.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/466908.

Issue #10865, opened 2021-08-17
api: some api client library methods mask the underlying error when non-200 status codes are received


dnephin:
See #8816

#### Overview of the Issue

Copied from #8816

> Consul agent returns errors in plain text. Yet when this happens the api client still tries to decode them in JSON, failing, and hiding the real error message.

It seems like some methods in the `api` client library do properly check the error code with `requireOk`, where as others do not. The methods that do not properly check the error code will error with JSON decode errors when the HTTP response was not 200. This error masks the real underlying problem, making the api client library difficult to use.

FFMMM:
We merged https://github.com/hashicorp/consul/pull/11054 to allow the error status code to be queried programmatically. 

Keeping this issue open until I take a pass over all `doRequest()` func calls and make sure they're wrapped in `requireHttpCodes()` or return a `StatusError` type

FFMMM:
@dnephin with PR https://github.com/hashicorp/consul/pull/11158 I think we can close this issue. What do you think?

For the refactor discussed here https://github.com/hashicorp/consul/pull/11158#issuecomment-930436357 I can open an issue to track visibility.

dnephin:
Sounds good!

PR #8816, opened 2020-10-05
[API] Handle plain text errors in client

ShimmerGlass:
Consul agent returns errors in plain text. Yet when this happens the
api client still tries to decode them in JSON, failing, and hidding the
real error message.
In the case the Content-Type is text/plain, and the response code is an
error one (>= 400), we return the response body as an error instead.

PR #11054, opened 2021-09-16
add StatusError to api package

FFMMM:
### Overview

This PR adds a `StatusError` struct to the `api` package. Now devs can unwrap the errors returned thru `query()` or `write()` funcs and inspect the error code programatically. 

### Issue Related
https://github.com/hashicorp/consul/issues/10865

### Notes:
* added a `requireHttpCodes` method to mark acceptable codes (200, 429, etc)
  * adoption of that func to all/ most `doRequest()` calls can be added in a subsequent PR
* this PR mostly builds on feedback from https://github.com/hashicorp/consul/pull/8816

### PR Checklist
[x] `go fmt`
[x] `go mod`
[x] n/a pr/changelog -- no "public" api changes

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -972,7 +972,7 @@ func (c *Client) doRequest(r *request) (time.Duration, *http.Response, error) {
 func (c *Client) query(endpoint string, out interface{}, q *QueryOptions) (*QueryMeta, error) {
 	r := c.newRequest("GET", endpoint)
 	r.setQueryOptions(q)
-	rtt, resp, err := c.doRequest(r)
+	rtt, resp, err := requireOK(c.doRequest(r))
 	if err != nil {
 		return nil, err
 	}
@@ -1090,16 +1090,29 @@ func encodeBody(obj interface{}) (io.Reader, error) {
 
 // requireOK is used to wrap doRequest and check for a 200
 func requireOK(d time.Duration, resp *http.Response, e error) (time.Duration, *http.Response, error) {
+	return requireHttpCodes(d, resp, e, []int{200})
+}
+
+// requireHttpCodes checks for the "allowable" http codes for a response
+func requireHttpCodes(d time.Duration, resp *http.Response, e error, httpCodes []int) (time.Duration, *http.Response, error) {
+	// todo -- use HTTP codes for httpCodes and not ints
 	if e != nil {
 		if resp != nil {
 			closeResponseBody(resp)
 		}
 		return d, nil, e
 	}
-	if resp.StatusCode != 200 {
-		return d, nil, generateUnexpectedResponseCodeError(resp)
+
+	// if there is an http code that we require, return w no error
+	for _, httpCode := range httpCodes {
+		if resp.StatusCode == httpCode {
+			return d, resp, nil
+		}
 	}
-	return d, resp, nil
+
+	// if we reached here, then none of the http codes in resp matched any that we expected
+	// so err out
+	return d, nil, generateUnexpectedResponseCodeError(resp)
 }
 
 // closeResponseBody reads resp.Body until EOF, and then closes it. The read
--- a/api/api.go
+++ b/api/api.go
@@ -82,6 +82,15 @@ const (
 	HTTPPartitionEnvName = "CONSUL_PARTITION"
 )
 
+type StatusError struct {
+	Code int
+	Body string
+}
+
+func (e StatusError) Error() string {
+	return fmt.Sprintf("Unexpected response code: %d (%s)", e.Code, e.Body)
+}
+
 // QueryOptions are used to parameterize a query
 type QueryOptions struct {
 	// Namespace overrides the `default` namespace
@@ -1090,12 +1099,11 @@ func encodeBody(obj interface{}) (io.Reader, error) {
 
 // requireOK is used to wrap doRequest and check for a 200
 func requireOK(d time.Duration, resp *http.Response, e error) (time.Duration, *http.Response, error) {
-	return requireHttpCodes(d, resp, e, []int{200})
+	return requireHttpCodes(d, resp, e, 200)
 }
 
 // requireHttpCodes checks for the "allowable" http codes for a response
-func requireHttpCodes(d time.Duration, resp *http.Response, e error, httpCodes []int) (time.Duration, *http.Response, error) {
-	// todo -- use HTTP codes for httpCodes and not ints
+func requireHttpCodes(d time.Duration, resp *http.Response, e error, httpCodes ...int) (time.Duration, *http.Response, error) {
 	if e != nil {
 		if resp != nil {
 			closeResponseBody(resp)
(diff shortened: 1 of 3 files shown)


api/operator_autopilot_test.go
```diff
@@ -181,4 +181,9 @@ func TestAPI_OperatorAutopilotServerHealth_429(t *testing.T) {
 	out, err := client.Operator().AutopilotServerHealth(nil)
 	require.NoError(t, err)
 	require.Equal(t, &reply, out)
+
+	mapi.withReply("GET", "/v1/operator/autopilot/health", nil, 500, nil).Once()
+	out, err = client.Operator().AutopilotServerHealth(nil)
+
+	require.Equal(t, 500, err.(StatusError).Code)
```
FFMMM, line 188: not much of a fan of the UX here -- the original proposal on https://discuss.hashicorp.com/t/golang-consul-api-pulling-out-the-statuscode-from-deeply-nested-responses/18037 was a more straightforward approach than the type assertions that devs would need to do here.

what does everyone think about this?

dnephin: I would find it surprising to include the HTTP status code as part of the response type.

Here's how I would expect someone to use the `StatusError`:

```go
var status *statusError
if errors.As(err, status) {
    // do something with status.Code
}
```

The test should probably do something similar with `errors.As` or `errors.Is` to prevent a panic when the error is of a different type.

dnephin (commented):
Thank you for working on this!

I haven't had a chance to look at the diff yet, but left a response to your question below.

api/operator_autopilot_test.go
FFMMM: thanks for pointing that out !! 💯 

with that more idiomatic way of inspecting the err type, I think I'm less concerned about the UX here!

FFMMM:
@ShimmerGlass thanks for your work in https://github.com/hashicorp/consul/pull/8816 ; i followed up on @dnephin's feedback w this PR.

Feel free to comment / add any input here but no pressure !

api/api.go
```diff
@@ -1117,7 +1139,7 @@ func generateUnexpectedResponseCodeError(resp *http.Response) error {
 	var buf bytes.Buffer
 	io.Copy(&buf, resp.Body)
 	closeResponseBody(resp)
-	return fmt.Errorf("Unexpected response code: %d (%s)", resp.StatusCode, buf.Bytes())
+	return StatusError{Code: resp.StatusCode, err: errors.New(string(buf.Bytes()))}
```
kisunji, line 1142: I don't see too much value in keeping `err error` field since we convert from []byte -> string -> error -> string (when output as `%v` from `Error()`)
Would it make more sense as `body string`?

api/operator_autopilot.go
```diff
@@ -352,22 +352,17 @@ func (op *Operator) AutopilotServerHealth(q *QueryOptions) (*OperatorHealthReply
 	r := op.c.newRequest("GET", "/v1/operator/autopilot/health")
 	r.setQueryOptions(q)
 
-	// we cannot just use requireOK because this endpoint might use a 429 status to indicate
-	// that unhealthiness
-	_, resp, err := op.c.doRequest(r)
+	// we use 429 status to indicate unhealthiness
+	d, resp, err := op.c.doRequest(r)
+	_, resp, err = requireHttpCodes(d, resp, err, []int{200, 429})
```
kisunji, line 357: Just to keep in line with other usages
```suggestion
	_, resp, err := requireHttpCodes(op.c.doRequest(r), []int{200, 429})
```

kisunji (commented):
minor nit-picky comments but otherwise looks great!

api/operator_autopilot.go
kisunji: Nvm! Discussed offline that this doesn't work as I thought (details [here](https://golang.org/ref/spec#Calls))

api/api.go
FFMMM: Thanks for pointing that out -- agreed 💯 ✅

api/api.go
```diff
@@ -82,6 +82,15 @@ const (
 	HTTPPartitionEnvName = "CONSUL_PARTITION"
 )
 
+type StatusError struct {
+	Code int
+	Body string
+}
+
+func (e StatusError) Error() string {
+	return fmt.Sprintf("Unexpected response code: %d (%v)", e.Code, e.Body)
```
kisunji, line 91:
```suggestion
	return fmt.Sprintf("Unexpected response code: %d (%s)", e.Code, e.Body)
```

kisunji (approved):

api/api.go
```diff
@@ -1090,16 +1099,28 @@ func encodeBody(obj interface{}) (io.Reader, error) {
 
 // requireOK is used to wrap doRequest and check for a 200
 func requireOK(d time.Duration, resp *http.Response, e error) (time.Duration, *http.Response, error) {
+	return requireHttpCodes(d, resp, e, []int{200})
+}
+
+// requireHttpCodes checks for the "allowable" http codes for a response
+func requireHttpCodes(d time.Duration, resp *http.Response, e error, httpCodes []int) (time.Duration, *http.Response, error) {
```
dnephin, line 1106: very minor: Using a varargs for `httpCodes` might be nice:
```suggestion
func requireHttpCodes(d time.Duration, resp *http.Response, e error, httpCodes ...int) (time.Duration, *http.Response, error) {
```

That would allow the caller to do things like:

```go
requireHttpCodes(d, resp, e, 200)
requireHttpCodes(d, resp, err, 200, 429)
```

but maybe that's just personal preference, up to you

dnephin (approved):
Nice! LGTM

I left a couple comments about some possible stylistic changes, but nothing that should block a merge

api/api.go
```diff
@@ -82,6 +82,15 @@ const (
 	HTTPPartitionEnvName = "CONSUL_PARTITION"
 )
 
+type StatusError struct {
+	Code int
+	Body string
+}
+
+func (e StatusError) Error() string {
+	return fmt.Sprintf("Unexpected response code: %d (%s)", e.Code, e.Body)
```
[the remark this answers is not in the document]
FFMMM: I'm on board with this but I see a potential problem with breaking existing string manipulations.

For one, in our testing code we do a fair amount of `expected: Unexpected response code: ...`. So doing that style change now will result in touching a number of file. Happy to address this in a subsequent PR.

That said, I'm a bit concerned for the community at large. For example, folks may be checking for the capitalized `U` like here: https://discuss.hashicorp.com/t/golang-consul-api-pulling-out-the-statuscode-from-deeply-nested-responses/18037. Then by virtue of making this change, we may _subtly_ break their code checks/ error handling.

What are your thoughts on the second point here?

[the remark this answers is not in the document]
dnephin: Ah ya, good call! I missed that this was an existing error message. Since there was no other way for users to check for error code previous, I think you're right we need to keep it as-is.

api/api.go
FFMMM: That's fair -- I added this!

PR #11075, opened 2021-09-16
improve error messages for intentions

jkirschner-hashicorp:
Inspired by internal troubleshooting confusion when trying to figure out why using intentions in a specific way resulted in an error. The error message itself wasn't clear enough to the end user to make the problem obvious.

Error messages now consistently include:
- That it's an intention validation error
- Whether the problem is with the destination or a source
- Which field in the destination or source is the problem (if applicable)

Additionally, some error messages have been rewritten for better clarity.

Example UI and CLI errors (click to expand and get better text readability):

image

image

### DIFF
--- a/agent/structs/config_entry_intentions.go
+++ b/agent/structs/config_entry_intentions.go
@@ -1,6 +1,7 @@
 package structs
 
 import (
+	"errors"
 	"fmt"
 	"sort"
 	"strings"
@@ -8,6 +9,44 @@ import (
 
 	"github.com/hashicorp/consul/acl"
 )
+
+// NotFoundError should be returned by a handler when a resource specified does not exist
+type IntentionValidationError struct {
+	// Destination or Sources[index]
+	Position string
+
+	// Optional: field within Position (e.g., Destination.Name)
+	Field string
+
+	// Validation error at Destination or Destination.Field
+	Violation string
+}
+
+// Formats errors produced by validate. Examples:
+// - ('Destination', 'Name', 'my error') => 'intention validation error: Destination.Name: my error'
+// - ('Sources[0]', "", 'lorem ipsum')   => 'intention validation error: Sources[0]: lorem ipsum'
+func (e IntentionValidationError) Error() string {
+	// If there's a field string, insert into message after Position separated by '.'
+	invalidFieldString := ""
+	if e.Field != "" {
+		invalidFieldString = "." + e.Field
+	}
+
+	return fmt.Sprintf("intention validation error: %s%s: %v", e.Position, invalidFieldString, e.Violation)
+}
+
+// Used to add Position information to an error that may be of type IntentionValidationError
+// after initial error generation. This is needed in cases where the generator of the error
+// lacks the Position information.
+func AddPositionToIntentionValidationError(err error, position string) error {
+	var validationErr *IntentionValidationError
+	if errors.As(err, &validationErr) {
+		validationErr.Position = position
+		return validationErr
+	} else {
+		return err
+	}
+}
 
 type ServiceIntentionsConfigEntry struct {
 	Kind string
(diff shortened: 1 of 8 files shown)


agent/structs/config_entry_intentions.go
```diff
@@ -576,128 +600,148 @@ func (e *ServiceIntentionsConfigEntry) validate(legacyWrite bool) error {
...
+				return wrapValidationErrorString(position, "LegacyCreateTime", "must be omitted")
 			}
 			if src.LegacyUpdateTime != nil {
-				return fmt.Errorf("Sources[%d].LegacyUpdateTime must be omitted", i)
+				return wrapValidationErrorString(position, "LegacyUpdateTime", "must be omitted")
 			}
-		}
-
-		if legacyWrite {
-			if src.LegacyID == "" {
```
jkirschner-hashicorp, line 624: This check was moved to L641

agent/structs/config_entry_intentions.go
```diff
@@ -751,24 +811,37 @@ func (e *ServiceIntentionsConfigEntry) validate(legacyWrite bool) error {
 }
 
 // Wildcard usage verification
-func validateIntentionWildcards(name string, entMeta *EnterpriseMeta) error {
+// position argument is passed to wrapValidationErrorString for better error messages
+func validateIntentionWildcards(name string, entMeta *EnterpriseMeta, position string) error {
```
kisunji, line 815: We should avoid allocating/passing args that are only used for error handling; I think the positional information can be wrapped by the caller of this func instead of it being passed down just to format an error message.

agent/structs/config_entry_intentions.go
```diff
@@ -535,39 +535,63 @@ func (e *ServiceIntentionsConfigEntry) HasAnyPermissions() bool {
...
+		invalidFieldString = "." + field
+	}
+
+	return fmt.Errorf("intention validation error: %s%s: %v", position, invalidFieldString, err)
+}
+
+func wrapValidationErrorString(position string, field string, errString string) error {
+	return wrapValidationError(position, field, fmt.Errorf(errString))
+}
+
```
kisunji, line 557: I have some reservations about abstracting away error handling to helper funcs. IMO error handling should be kept simple (even at the cost of repetition) and close to the `if err != nil` check (as-is, a dev will have to step into 1-2 funcs to see what's being done with the args).

My gut feeling is to keep the verbose error-handling as-is (with added Destination/Source info) but it's probably best to get more opinions!

An alternative solution might be to define a custom error type so you can do something like
```
return &customErr{
    pos:  "Destination",
    field: "Name",
    msg: "service name is required and cannot be empty"
}
```
which is a bit more idiomatic and upfront about what the fields are.

jkirschner-hashicorp: > My gut feeling is to keep the verbose error-handling as-is (with added Destination/Source info) but it's probably best to get more opinions!

It certainly wouldn't be too difficult to rewrite this PR to directly use `fmt.Errorf`, it would just have a lot of copy-paste formatting and I was concerned about the maintainability of consistent error messages.

I think you've proposed a good alternative solution -- defining a custom error type. The `Error()` method would enforce consistent formatting. What are your thoughts on defining the `customErr` struct on one-line versus multiple? (I feel a bit weird about expanding all of the existing one-line messages into 5 lines, but I'm not super familiar with Golang conventions for this kind of thing.)

agent/structs/config_entry_intentions.go
```diff
@@ -536,168 +577,256 @@ func (e *ServiceIntentionsConfigEntry) HasAnyPermissions() bool {
...
+				Position: position,
+				Field: "Name",
+				Violation: "field is required and cannot be empty",
+			}
 		}
 
 		if err := validateIntentionWildcards(src.Name, &src.EnterpriseMeta); err != nil {
-			return fmt.Errorf("Sources[%d].%v", i, err)
+			// Add Position information (which can't be added by validateIntentionWildcards)
+			return AddPositionToIntentionValidationError(err, position)
```
jkirschner-hashicorp, line 635: @kisunji : this was my attempt to avoid passing position information to `validateIntentionWildcards`. I'm injecting it into the error after-the-fact.

agent/structs/config_entry_intentions.go
```diff
@@ -536,168 +577,256 @@ func (e *ServiceIntentionsConfigEntry) HasAnyPermissions() bool {
...
 		}
 
 		if err := validateSourceIntentionEnterpriseMeta(&src.EnterpriseMeta, &e.EnterpriseMeta); err != nil {
-			return fmt.Errorf("Sources[%d].%v", i, err)
+			return fmt.Errorf("%s.%v", position, err)
```
jkirschner-hashicorp, line 639: @kisunji : I didn't touch this error message (other than re-using `position` rather than recreating it from `Sources[%d]`) because I hadn't investigated what `validateSourceIntentionEnterpriseMeta` can return in enterprise (and how it would fit into this new framework for generating intention validation errors). Let me know if I need to look into that.

jkirschner-hashicorp:
@kisunji : I've made the requested changes. I think it's much better now with a specific `IntentionValidationError` type that centrally manages the error message formatting. Let me know what you think.

website/content/commands/intention/index.mdx
```diff
@@ -52,6 +52,7 @@ $ consul intention create web db
 Create an intention to deny "db" from initiating connections to _any_ service:
 
 ```shell-session
+# Wrap "*" in quotes to prevent wildcard expansion by the shell
 $ consul intention create -deny db '*'
```
kisunji, line 56: The command uses single quotes but should we be consistent with single vs double?

agent/structs/config_entry_intentions.go
```diff
@@ -9,6 +10,44 @@ import (
 	"github.com/hashicorp/consul/acl"
 )
 
+// NotFoundError should be returned by a handler when a resource specified does not exist
+type IntentionValidationError struct {
+	// Destination or Sources[index]
+	Position string
+
+	// Optional: field within Position (e.g., Destination.Name)
+	Field string
```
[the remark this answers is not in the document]
kisunji: I just realized I suggested the initial struct format 🤦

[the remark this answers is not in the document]
jkirschner-hashicorp: We're allowed to change our minds ;)

website/content/commands/intention/index.mdx
jkirschner-hashicorp: I think I'll stick to double-quotes. "*" shouldn't be interpreted by the shell AFAIK, so no need for single quote specifically

website/content/commands/intention/create.mdx
```diff
@@ -65,3 +65,11 @@ Create intentions from a directory using shell expansion:
 ```shell-session
 $ consul intention create -file intentions/*.json
 ```
+
+<EnterpriseAlert inline /> Create an intention from any service in any namespace
+to the assets-public service in the default namespace:
+
+```shell-session
+# Wrap "*/*" in quotes to prevent wildcard expansion by the shell
```
trujillo-adam, line 73:
```suggestion
# Wrap "*/*" in quotation marks to prevent wildcard expansion by the shell
```
Perhaps just a nit, but the _quote_ is the stuff between the _quotation marks_.

website/content/commands/intention/index.mdx
```diff
@@ -52,6 +52,7 @@ $ consul intention create web db
 Create an intention to deny "db" from initiating connections to _any_ service:
 
 ```shell-session
+# Wrap "*" in quotes to prevent wildcard expansion by the shell
```
trujillo-adam, line 55:
```suggestion
# Wrap "*" in quotation marks to prevent wildcard expansion by the shell
```

trujillo-adam (approved):
Reviewed the only thing I am qualified to review 🙂 and made a nit-level suggestion.

jmurret:

jkirschner-hashicorp:
This fell victim to "perfect is the enemy of good". If I find myself with spare time in the future, I may re-open, but I agree with closing for now.

PR #11078, opened 2021-09-17
Consul on Kubernetes - Features List Section

pbortnick:
🎟️ [Asana Task](https://app.asana.com/0/1200785913255669/1200992155224291/f)
🔍 [Preview Link](https://consul-git-pbfeatures-list-section-hashicorp.vercel.app/consul-on-kubernetes)
💅 [Design Reference](https://www.figma.com/file/0DT6uKPK214eFZtMG0pQUA/Consul-on-Kubernetes?node-id=102%3A6583)

This PR contains the Features List Section ("Why Consul on Kubernetes") of the new consul-on-kubernetes page


website/components/features-list/index.jsx
```diff
@@ -0,0 +1,22 @@
+import Feature from './feature'
+import s from './style.module.css'
+
+export default function FeaturesList({ title, features }) {
+  return (
+    <div
+      className={s.featureListContainer}
+      style={{
+        backgroundImage: `url(${require('./images/top-right-design.svg')}), url(${require('./images/bottom-left-design.svg')})`,
```
BrandonRomano, line 9: Nice love how you're passing this here, and including the images in this component.  Nice encapsulation 👌

website/components/features-list/index.jsx
```diff
@@ -0,0 +1,22 @@
...
+      }}
+    >
+      <div className="g-grid-container">
```
BrandonRomano, line 12: I try to avoid using global class names in a CSS module.  It slightly breaks encapsulation, and as a result of that I have a little trick to self-document this.

I'll DM ya a slack thread and context on this where we had a conversation as a team about this, but largely, I'll use a CSS module class name like `s.contentWrapper` here, and then define content wrapper as such:

```css
.contentWrapper {
  composes: g-grid-container from global;
}
```

This "from global" acts as a greppable artifact explicitly acknowledging that CSS module usage is being broken, where as the `g-grid-container` class usage relies on convention, which is not always followed (thus not absolutely greppable).

website/data/consul-on-kubernetes.js
```diff
@@ -0,0 +1,131 @@
...
+            <li>Work across Kubernetes & non-Kubernetes Environments</li>
+          </ul>
+        ),
+      },
+    ],
+    cta: {
+      text: 'Try It Now',
+      url: '#TODO',
+    },
+    image: '/img/consul-k8/features/multi-platform.png',
```
BrandonRomano, line 32: Let's switch all these PNGs to SVGs.  LMK if you need help exporting them!

website/components/features-list/style.module.css
```diff
@@ -0,0 +1,107 @@
...
+  background-repeat: no-repeat;
+  color: #fff;
+}
+
+.featureContainer {
+  display: flex;
+  align-items: center;
+  flex-wrap: wrap;
+
+  & .featureTextContainer {
```
BrandonRomano, line 14: Probably some margin-right on desktop here or something to ensure some breathing room at very specific breakpoints.

These but up against each-other.

CleanShot 2021-09-21 at 11 10 31@2x

website/components/features-list/style.module.css
```diff
@@ -0,0 +1,107 @@
...
+  row-gap: 120px;
+}
+
+.title {
+  composes: g-type-display-1 from global;
+  max-width: 488px;
+  margin: 0;
+}
+
+.listNumber {
```
BrandonRomano, line 56: Gets a little tight on mobile.  I'm willing to bet the intention is that these remain square if possible.

CleanShot 2021-09-21 at 11 13 04@2x

website/components/features-list/feature.jsx
```diff
@@ -0,0 +1,44 @@
...
+        <img src={image} alt={title} />
+      </div>
+      <div className={s.featureTextContainer}>
+        <span className={s.listNumber}>{number}</span>
+        <div className={s.featureText}>
+          <h3 className={s.featureTitle}>{title}</h3>
+          <p className={s.featureSubtitle}>{subtitle}</p>
+          <div className={s.infoSection}>
+            {infoSections.map(({ heading, content }) => (
+              <div key={`${title}-${heading}`}>
```
BrandonRomano, line 24: This title is unnecessary -- Keys in react are unique to the map -- so if there's another Feature that has a similar heading it won't cause an issue.

The `title` doesn't add any value here, as the title is the same value for each iteration of this map.

```suggestion
              <div key={heading}>
```

website/components/features-list/index.jsx
BrandonRomano: Ah, I saw you do this elsewhere, so you probably get it!

website/pages/consul-on-kubernetes/index.jsx
```diff
@@ -1,17 +1,21 @@
...
         <title key="title">Consul on Kubernetes</title>
       </Head>
+
       {/* hero */}
 
       {/* side by side section */}
 
-      {/* features list section */}
+      <section>
+        <FeaturesList title="Why Consul on Kubernetes" features={features} />
```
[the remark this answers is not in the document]
BrandonRomano: Ah, I see you are importing a `js` file here, and not a `json` file after further review, but my thought process still applies here; I'd prefer the content to be in this file here!

website/data/consul-on-kubernetes.js
BrandonRomano: Additionally, could we house these images rather than in the public directory, in the page's folder maybe in a subfolder called `image` or something.  Strengthens encapsulation of the page logic!

website/components/features-list/style.module.css
```diff
@@ -0,0 +1,107 @@
...
+      margin: 0;
+    }
+  }
+}
+
+.featuresContainer {
+  padding-top: 157px;
+  padding-bottom: 394px;
+  display: grid;
+  row-gap: 120px;
```
BrandonRomano, line 47: This is a really subtle, but excellent decision.

Often, I'll see the decision to put spacing on the *feature* item (as margin-top/margin-bottom), which is less preferable to than what you did here, as the spacing between each feature is a contextual decision to the fact they are in a list.  You approach sets us up for success if we ever need to show one of these feature cards in a different context.

Great stuff ❇️  definitely impressed.

BrandonRomano (changes requested):
I had a lot of really small comments, but this is an excellent PR.  Tons of great decisions throughout.  Specifically, your abstraction layers are excellent. 👨‍🍳 😗 👌 

Mostly being nitty-gritty here cause this is one of my first meaningful PRs I'm giving you, and you had to wait a while for my feedback here, so I wanted it to be good review ❇️

website/components/features-list/style.module.css
pbortnick: Yeah, this way we don't have to remove the margin on the last feature, etc. Thanks!

website/components/features-list/style.module.css
```diff
@@ -0,0 +1,112 @@
...
+}
+
+.featureContainer {
+  display: flex;
+  align-items: center;
+  flex-wrap: wrap;
+
+  & .featureTextContainer {
+    display: flex;
+    margin-right: 60px;
```
BrandonRomano, line 20: This is only needed when they are side-by-side -- mind wrapping this in the same breakpoint that you use to flip the flex-direction?

Gives a little more room for the content.

Screen Shot 2021-09-21 at 12 24 50 PM

BrandonRomano (changes requested):
One last teeny change, then this is good.

website/components/features-list/feature.tsx
```diff
@@ -0,0 +1,66 @@
...
+}
+
+export default function Feature({
+  number,
+  title,
+  subtitle,
+  infoSections,
+  cta,
+  image,
+}: FeatureProps) {
```
BrandonRomano, line 31: No feedback here on this TS conversion, looks exactly as I would have done it 👍

BrandonRomano (approved):
Excellent work 👏👏👏👏👏

PR #11017, opened 2021-09-13
ui: Partitions Application Layer

johncowen:
TBD

~Kind of still in draft~, but good to start getting 👀 here, so definitely open for review as it stands. There will be some changes ontop. I'll add a description shortly (best viewing commit by commit for now).

PR #11096, opened 2021-09-21
ui: Remove legacy ACLs

johncowen:
This PR removes all leftover old style "legacy ACLs" from the UI. It's 'kind of' a continuation of https://github.com/hashicorp/consul/pull/11017, in which we only removed things that were complicating adding support for Admin Partitions in the UI, ideally we would have done this PR before that one so we've decided it would be best to finish up removing legacy ACL support completely before moving on with finishing up Admin Partition support.

Just to be clear, this PR is not blocked on any backend support/removal of legacy ACLs.

Most changes are deletions, pretty much all that remains for some redirecting for where the old style ACLs used to live to take you straight to the ACLs Tokens page. There're probably still some CSS code that we no longer need here, but I'll do a larger CSS cleanup later as part of a separate iterative PR.

I'll try add some inline self-review a little later.


ui/packages/consul-ui/app/services/repository/token.js
```diff
@@ -53,21 +45,14 @@ export default class TokenService extends RepositoryService {
 
   @dataSource('/:partition/:ns/:dc/token/self/:secret')
   self(params) {
-    // TODO: Does this need ns passing through?
+    // This request does not need ns or partition passing through as its
+    // inferred from the token itself.
```
johncowen, line 49: I checked in with the team about this and confirmed that it does not need `ns` or `partition` passing, so I fixed this up while I was here and added some more information here to help future me when I next come here

johncowen (commented):
Added inline notes

kaxcode (approved):
👏

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/453704.

PR #11097, opened 2021-09-21
ui: Adds Flight icons to our icon set

johncowen:
This PR adds @hashicorp/flight-icons to our icon set so we can use them when we want to/need to.

Please note: This PR does not use the new icons at all, it just installs them ready for use. All unused icons are tree-shaken out of the final build so it has absolutely no impact on the current UI as it stands.


amyrlam (commented):
looks good, thanks for keeping us informed, please do on the other PR's if you're up for it! 

seems like makes sense to defer official approval to @kaxcode if that's cool?

johncowen:
Awesome, np thanks @amyrlam ! Will defo keep you folks in the loop with the next round of bits for this.

@kaxcode sounds like this is just gated on an approval from you now, lemme know if you have any questions 👍

amyrlam:
@johncowen @kaxcode for posterity, bumped you to a 1.0.0 here https://github.com/johncowen/iris/pull/3, figured might as well be on that version as move forward, even though no change to the consumer yet

cc @didoo @MelSumner @Dhaulagiri

johncowen:
Amazing, thanks for that @amyrlam 🙇 , makes perfect sense. All merged over there!

kaxcode (approved):
👍

johncowen:
Thanks all! 🌈 ✈️ 🎉

PR #11101, opened 2021-09-21
acl: remove legacy ACL.Apply RPC

dnephin:
Branched from #10986

This PR removes the remaining tests for `ACL.Apply` (and updates one for ACLResolver to not use ACL.Apply), and then replaces the implementation with an error.


agent/setup.go
```diff
@@ -296,7 +296,6 @@ func getPrometheusDefs(cfg lib.TelemetryConfig) ([]prometheus.GaugeDefinition, [
 		HTTPSummaries,
 		consul.ACLSummaries,
 		consul.ACLEndpointSummaries,
-		consul.ACLEndpointLegacySummaries,
```
rboyer, line 299: If you hadn't already, this should go on the short list of stuff that needs cleaned up on the website page for metrics.

rboyer (approved):
LGTM

agent/setup.go
dnephin: Oh, good call. I can probably add that to this PR

karl-cardenas-coding (approved):
Good from a doc perspective

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/460414.

PR #11095, opened 2021-09-21
ui: [BUGFIX] Re-enable namespace menus whilst editing intentions

johncowen:
This PR supersedes https://github.com/hashicorp/consul/pull/10706 and fixes https://github.com/hashicorp/consul/issues/10686 whilst making sure that saving intentions continues to work.

The original fix in https://github.com/hashicorp/consul/pull/10706 ignored the `change` action configured for the change event on the menus, meaning that the selected source/destination namespace could not be set by the user when editing/creating intentions. This, coupled with the fact that using the later intention `exact` endpoint for API requests endpoint means that you cannot use wildcard namespaces for saving intentions.

All in all this meant that intentions could no longer be saved using the UI (whilst using ENT):

intention-nspace-menus

This PR reverts https://github.com/hashicorp/consul/pull/10706 to fix the intention saving issue, and adds a fix for the original visual issue of nspaces doubling up in the menu once clicked. This meant repeating this functionality:

https://github.com/hashicorp/consul/blob/b2d17ac4487983b49b006928601d296d7466d2a2/ui/packages/consul-ui/app/components/consul/intention/form/index.js#L143

...for `nspaces` aswell as `services`. It did seem strange to me that the original issue was only apparent for the nspace menus and not the service menus which should all function exactly the same way.

There is potentially more to come here partly related to what the exact functionality should be, but I'm working with other folks to figure out what the best way forwards is longer term. In the meantime this brings us back to the original functionality with the visual issue fixed.

PR #11102, opened 2021-09-22
ui: [Backport] Backport of #11095 to 1.8.x

johncowen:
See #11095. Whilst this is a backport it only fixes the visual issue with the namespace menus (doubling up of selectable elements in the menu). The main issue that #11095 fixes is not an issue in 1.8.x as the original fix for this issue was never backported to 1.8.x.

I've included the tests I copy/pasted for 1.9.x, 1.10.x and 1.11.x here also, even though the bug introduced in those versions is not an issue in 1.8.x, it's still maybe useful to have those here also.


kaxcode (approved):
:shipit:

Issue #10897, opened 2021-08-23
Envoy listener escape hatch for prepared query upstreams ignored in 1.10


gzavitz:
#### Overview of the Issue

The `envoy_listener_json` field in the upstream configuration of a `prepared_query` destination is ignored in versions 1.10+. The cluster escape hatch does still work in 1.10.

#### Reproduction Steps

1. Run `consul agent -dev`
1. Create any prepared query
1. Register a service, then its sidecar configuration with Consul. Include an upstream to the previously created prepared query, including a listener escape hatch in `envoy_listener_json`.
1. Generate the bootstrap configuration for the sidecar and start Envoy with it
1. Observe both the agent trace xds logs and the config_dump admin interface to see that Consul is returning a default TCP proxy listener instead of the listener specified in the escape hatch.

### Consul info for both Client and Server

<details>
  <summary>Client info</summary>

```
agent:
	check_monitors = 0
	check_ttls = 0
	checks = 0
	services = 2
build:
	prerelease =
	revision = db839f18
	version = 1.10.1
consul:
	acl = disabled
	bootstrap = false
	known_datacenters = 1
	leader = true
	leader_addr = 127.0.0.1:8300
	server = true
raft:
	applied_index = 49
	commit_index = 49
	fsm_pending = 0
	last_contact = 0
	last_log_index = 49
	last_log_term = 2
	last_snapshot_index = 0
	last_snapshot_term = 0
	latest_configuration = [{Suffrage:Voter ID:0fcc7360-f8c3-72c7-48b8-63a401d289ce Address:127.0.0.1:8300}]
	latest_configuration_index = 0
	num_peers = 0
	protocol_version = 3
	protocol_version_max = 3
	protocol_version_min = 0
	snapshot_version_max = 1
	snapshot_version_min = 0
	state = Leader
	term = 2
runtime:
	arch = amd64
	cpu_count = 8
	goroutines = 125
	max_procs = 8
	os = linux
	version = go1.16.6
serf_lan:
	coordinate_resets = 0
	encrypted = false
	event_queue = 1
	event_time = 2
	failed = 0
	health_score = 0
	intent_queue = 0
	left = 0
	member_time = 1
	members = 1
	query_queue = 0
	query_time = 1
serf_wan:
	coordinate_resets = 0
	encrypted = false
	event_queue = 0
	event_time = 1
	failed = 0
	health_score = 0
	intent_queue = 0
	left = 0
	member_time = 1
	members = 1
[...]

PR #11109, opened 2021-09-22
connect: Allow upstream listener escape hatch for prepared queries

kisunji:
Fixes https://github.com/hashicorp/consul/issues/10897

We were missing a codepath for envoy listener escape hatches defined on Prepared Query upstream configs. The envoy tests were updated to cover this case.

### DIFF
--- a/agent/xds/listeners.go
+++ b/agent/xds/listeners.go
@@ -257,7 +257,7 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
 		if cfg.EnvoyListenerJSON != "" {
 			upstreamListener, err := makeListenerFromUserConfig(cfg.EnvoyListenerJSON)
 			if err != nil {
-				s.Logger.Warn("failed to parse envoy_listener_json",
+				s.Logger.Error("failed to parse envoy_listener_json",
 					"upstream", u.Identifier(),
 					"error", err)
 				continue
--- a/agent/xds/listeners.go
+++ b/agent/xds/listeners.go
@@ -257,7 +257,10 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
 		if cfg.EnvoyListenerJSON != "" {
 			upstreamListener, err := makeListenerFromUserConfig(cfg.EnvoyListenerJSON)
 			if err != nil {
-				return nil, err
+				s.Logger.Warn("failed to parse envoy_listener_json",
+					"upstream", u.Identifier(),
+					"error", err)
+				continue
 			}
 			resources = append(resources, upstreamListener)
 			continue
--- a/agent/xds/listeners.go
+++ b/agent/xds/listeners.go
@@ -252,6 +252,17 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
 			// default config if there is an error so it's safe to continue.
 			s.Logger.Warn("failed to parse", "upstream", u.Identifier(), "error", err)
 		}
+
+		// If escape hatch is present, create a listener from it and move on to the next
+		if cfg.EnvoyListenerJSON != "" {
+			upstreamListener, err := makeListenerFromUserConfig(cfg.EnvoyListenerJSON)
+			if err != nil {
+				return nil, err
+			}
+			resources = append(resources, upstreamListener)
+			continue
+		}
+
 		upstreamListener := makeListener(id, u, envoy_core_v3.TrafficDirection_OUTBOUND)
 
 		filterChain, err := s.makeUpstreamFilterChainForDiscoveryChain(
--- a/agent/xds/listeners_test.go
+++ b/agent/xds/listeners_test.go
@@ -159,20 +159,24 @@ func TestListenersFromSnapshot(t *testing.T) {
 			name:   "custom-upstream",
 			create: proxycfg.TestConfigSnapshot,
 			setup: func(snap *proxycfg.ConfigSnapshot) {
-				snap.Proxy.Upstreams[0].Config["envoy_listener_json"] =
-					customListenerJSON(t, customListenerJSONOptions{
-						Name: "custom-upstream",
-					})
+				for i := range snap.Proxy.Upstreams {
+					snap.Proxy.Upstreams[i].Config["envoy_listener_json"] =
+						customListenerJSON(t, customListenerJSONOptions{
+							Name: snap.Proxy.Upstreams[i].DestinationName + ":custom-upstream",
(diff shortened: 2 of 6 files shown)


agent/xds/listeners_test.go
```diff
@@ -159,20 +159,24 @@ func TestListenersFromSnapshot(t *testing.T) {
...
-				snap.Proxy.Upstreams[0].Config["envoy_listener_json"] =
-					customListenerJSON(t, customListenerJSONOptions{
-						Name: "custom-upstream",
-					})
+				for i := range snap.Proxy.Upstreams {
+					snap.Proxy.Upstreams[i].Config["envoy_listener_json"] =
+						customListenerJSON(t, customListenerJSONOptions{
+							Name: "custom-upstream",
+						})
+				}
```
kisunji, line 167: Test was only using escape hatch for the first upstream and was not asserting for prepared query upstream

agent/xds/testdata/listeners/custom-upstream.envoy-1-18-x.golden
```diff
@@ -41,14 +41,13 @@
               "name": "envoy.filters.network.tcp_proxy",
               "typedConfig": {
                 "@type": "type.googleapis.com/envoy.extensions.filters.network.tcp_proxy.v3.TcpProxy",
-                "statPrefix": "upstream.prepared_query_geo-cache",
-                "cluster": "geo-cache.default.dc1.query.11111111-2222-3333-4444-555555555555.consul"
+                "statPrefix": "foo-stats",
+                "cluster": "random-cluster"
```
freddygv, line 45: Where did this change come from? 🤔

agent/xds/testdata/listeners/custom-upstream.v2compat.envoy-1-16-x.golden
```diff
@@ -41,14 +41,13 @@
               "name": "envoy.filters.network.tcp_proxy",
               "typedConfig": {
                 "@type": "type.googleapis.com/envoy.config.filter.network.tcp_proxy.v2.TcpProxy",
-                "statPrefix": "upstream.prepared_query_geo-cache",
-                "cluster": "geo-cache.default.dc1.query.11111111-2222-3333-4444-555555555555.consul"
+                "statPrefix": "foo-stats",
+                "cluster": "random-cluster"
```
freddygv, line 45: Same as above

freddygv (commented):
Thanks for fixing this! Have a couple comments/questions below

agent/xds/testdata/listeners/custom-upstream.envoy-1-18-x.golden
kisunji: Previously the test setup was using escape hatch for only the first upstream:
```
			snap.Proxy.Upstreams[0].Config["envoy_listener_json"] =
				customListenerJSON(t, customListenerJSONOptions{
					Name: "custom-upstream",
				})
```
So the 2nd upstream (Prepared Query) was never being tested with the escape hatch.
I changed the test to apply the same escape hatch json to all the upstreams

agent/xds/listeners.go
```diff
@@ -252,6 +252,17 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
 			// default config if there is an error so it's safe to continue.
 			s.Logger.Warn("failed to parse", "upstream", u.Identifier(), "error", err)
 		}
+
+		// If escape hatch is present, create a listener from it and move on to the next
+		if cfg.EnvoyListenerJSON != "" {
+			upstreamListener, err := makeListenerFromUserConfig(cfg.EnvoyListenerJSON)
+			if err != nil {
+				return nil, err
```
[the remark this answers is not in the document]
kisunji: Sounds good!

agent/xds/testdata/listeners/custom-upstream.envoy-1-18-x.golden
freddygv: Ah, got it!

agent/xds/listeners.go
```diff
@@ -252,6 +252,20 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
 			// default config if there is an error so it's safe to continue.
 			s.Logger.Warn("failed to parse", "upstream", u.Identifier(), "error", err)
 		}
+
+		// If escape hatch is present, create a listener from it and move on to the next
+		if cfg.EnvoyListenerJSON != "" {
+			upstreamListener, err := makeListenerFromUserConfig(cfg.EnvoyListenerJSON)
+			if err != nil {
```
kisunji, line 259: Instead of returning on the error I decided to log a WARN here, similar to what's being done above

agent/xds/listeners.go
```diff
@@ -252,6 +252,20 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
...
+				s.Logger.Warn("failed to parse envoy_listener_json",
```
freddygv, line 260: Hmmm I'm wondering if this should be logged as an error instead, since someone is expecting a custom listener, but one isn't being generated.

What do you think?

kisunji: Should the Warn above this block be an error as well? It looks like the original intention for malformed configs was to warn and apply defaults.

In this scenario:
1. Config looks good
2. Upstream's escape hatch json is malformed

Should we output an error log and proceed with returning a default listener? OR
Simply return error here with context

freddygv: Looking at the other call to `makeListenerFromUserConfig` in `makeInboundListener`, there's the behavior of warning on parse errors, but returning an error if the custom listener can't be generated. I think that makes sense since it means that the single listener for inbound communication can't be generated as expected.

In this case, the listener from config is for one of potentially several upstreams, so we could log an error and skip the listener instead. That way at least some of the upstream listeners get configured.

I think that returning a default listener in the case of an escape hatch would be too unexpected. Particularly given that escape hatch listeners define the entire object.

freddygv: Regarding whether failing to parse the upstream config should yield an error, I'm leaning toward yes, since it has downstream implications on escape hatches

freddygv (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/454460.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit f972048ebce941e802dfde62cbc49d02dbdf1285 onto `release/1.10.x` succeeded!

PR #11110, opened 2021-09-22
acl: remove initializeLegacyACL and the rest of the legacy FSM commands

dnephin:
Branched from #11101

This PR removes:
*  `initializeLegaclACL` from the leader initialization
*  a state store function that is no longer used
* the rest of the legacy FSM command and `ACLOp`
* renames a few things with a deprecated prefix to indicate they are no longer used, but must remain


agent/consul/fsm/snapshot_oss.go
```diff
@@ -15,7 +15,7 @@ func init() {
 	registerRestorer(structs.KVSRequestType, restoreKV)
 	registerRestorer(structs.TombstoneRequestType, restoreTombstone)
 	registerRestorer(structs.SessionRequestType, restoreSession)
-	registerRestorer(structs.ACLRequestType, restoreACL)
+	registerRestorer(structs.DeprecatedACLRequestType, restoreACL)
```
rboyer, line 18: Did you want to also clean up `restoreACL` yet?

dnephin: Don't we still need `restoreACL` for snapshots that contain legacy ACL entries? I kept the test around to show we can still restore snapshots with this data.

rboyer: Ah, right. I wasn't sure how much we were just going to restore old ACL data directly into `/dev/null` or not.

rboyer (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/460777.

PR #11126, opened 2021-09-22
acl: remove ACL.GetPolicy RPC endpoint and ACLResolver.resolveTokenLegacy

dnephin:
The first commit also removes ACL serf tags from Client agents (they were never used), and removes ACL upgrading from client agents. The next PR will remove ACL upgrades from servers.


rboyer (approved):
LGTM

agent/consul/leader_federation_state_ae_test.go
```diff
@@ -383,7 +383,7 @@ func TestLeader_FederationStateAntiEntropyPruning_ACLDeny(t *testing.T) {
 	// Try to join.
 	joinWAN(t, s2, s1)
 	testrpc.WaitForLeader(t, s1.RPC, "dc1")
-	testrpc.WaitForLeader(t, s1.RPC, "dc2")
+	testrpc.WaitForLeader(t, s1.RPC, "dc2", testrpc.WithToken("root"))
```
dnephin, line 386: Calling out this test fix. I'm not sure why this is necessary now, but this fixed the test failure.

Since this is test setup, not the actual test logic, I'm thinking it's probably ok.

agent/consul/acl_endpoint.go
```diff
@@ -170,11 +170,6 @@ func (a *ACL) aclPreCheck() error {
 	if !a.srv.config.ACLsEnabled {
 		return acl.ErrDisabled
 	}
-
-	if a.srv.UseLegacyACLs() {
-		return fmt.Errorf("The ACL system is currently in legacy mode.")
-	}
```
dnephin, line 176: This was required to fix the test failures. The full removal of upgrades is happening in #11182

dnephin:
I had to make some changes to get the tests to pass, so re-requesting a review.

agent/consul/leader_federation_state_ae_test.go
freddygv: Hmm this is unusual, I'm running this test locally and it's failing on L379 due to ACL not found.

It doesn't pass when changing that line to include the root token either 🤔

freddygv: I think it's because the secondary DC is declaring a "root" token, but the global management token can only be initialized in the primary:
https://github.com/hashicorp/consul/blob/main/agent/consul/leader.go#L431-L468

So "root" isn't a known token in DC2 until they are WAN joined, which is why your fix is needed.

I think the WaitForLeader in L379 should potentially be removed because there's no token that can be used for those catalog queries.

freddygv: Ah it's a similar problem that requires your fix. The anonymous token is only created in the primary DC, so sending a request with an empty token to dc2 will lead to ACL not found.

freddygv: Not sure why this wasn't needed before this PR though.

agent/consul/acl_server.go
```diff
@@ -95,6 +95,7 @@ func (s *Server) updateSerfTags(key, value string) {
 	s.updateEnterpriseSerfTags(key, value)
 }
 
+// TODO:
```
freddygv, line 98: Missing comment?

agent/consul/leader_federation_state_ae_test.go
freddygv: Ah, in `ResolveTokenToIdentityAndAuthorizer`, `r.delegate.UseLegacyACLs()` was evaluating to true, and the legacy resolve method would handle missing tokens by returning `missingIdentity`, rather than "ACL not found".

agent/consul/acl_server.go
freddygv: This gets deleted in the followup PR

dnephin: Thanks for confirming! I thought it might , but didn't have a chance to look. I'll resolve this comment.

agent/consul/leader_federation_state_ae_test.go
dnephin: Ok, removing both this `testrpc.WithToken("root")` and the `WaitForLeader` on L379 does seem to work.

I think I understand why it worked previously (I had guessed the resolver legacy case was at play but didn't trace it back to the different error being related), and why adding `testrpc.WithToken("root")` fixed the failures (because it was actually waiting for replication, not simply the leader).

But I don't think I understand why not waiting for a leader in dc2 fixes the problem. Is this fix relying on some timing of the two servers which could cause the test to flake if the order changes?

I've pushed that change.

freddygv: > But I don't think I understand why not waiting for a leader in dc2 fixes the problem. Is this fix relying on some timing of the two servers which could cause the test to flake if the order changes?

Do you mean that you're not sure why removing this line helps? `testrpc.WaitForLeader(t, s2.RPC, "dc2")`

There's a couple things I'm seeing:
1. When s2 in dc2 starts it doesn't initialize any tokens in its state store, because it is not the primary datacenter. Only the primary DC will ensure the anonymous and global-management tokens exist. So specifying `c.ACLMasterToken = "root"` in s2 is a no-op, because a secondary DC can't persist the master token.
2. Since ACLs are in deny mode, `WaitForLeader` requires a token in order to query `Catalog.ListNodes`. 

However, due to point #1, there is no token that can be passed into `WaitForLeader` in s2. s2 can't create any tokens because it's not the authoritative ACL datacenter. It also doesn't have any tokens because it hasn't replicated any from the primary DC. 

s2 can't get out of this situation until it is WAN joined with dc1. 

I think the only potential source for flakiness here could be if: dc1 and dc2 are WAN joined, dc1 does not have a leader and therefore hasn't initialized ACLs, and then s2 attempts to use a token that needs to be replicated from s1 ("root"). 

Though I don't think we need to worry about this here because we wait for dc1 to have a leader before spinning up and joining s2, and then the dc2 WaitForLeader is wrapped in a retry as well. So that should provide enough time for the root token to be replicated to, and resolved in dc2.

freddygv: Forgot to add: The reason why a token isn't needed when calling `testrpc.WaitForLeader(t, s1.RPC, "dc1")` is that the anonymous token gets filled in here: https://github.com/hashicorp/consul/blob/19040586ce9102b083c82ea9bb3c5efd141b4712/agent/consul/acl.go#L1240. In the primary DC this token ID can be resolved to a policy.

The token doesn't have any permissions, but because it exists it won't lead to "ACL not found". `Catalog.ListNodes` does filtering, so the end result will be an empty list rather than a permission denied error.

freddygv (approved):

agent/consul/leader_federation_state_ae_test.go
dnephin: I think part of my confusion here is that `WaitForLeader` is implemented as an RPC call for some endpoint that requires ACLS, and ACL errors are assumed to mean something specific. If the implementation used something like [`/status/leader`](https://www.consul.io/api-docs/status#get-raft-leader) instead, which does not require ACLs, that might be a bit easier to reason about.

Either way, sounds like this test is working now, so we can look into those improvements some other time.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/463268.

PR #11129, opened 2021-09-23
ui: Add partition support to intentions

johncowen:
This PR is an iterative PR ontop of https://github.com/hashicorp/consul/pull/11017 to add further `partition` support for intentions.

Most of this is the same as our namespace support so it's pretty much adding `partition` everywhere with some copypasta.

Just to note, there are some little differences between namespaces and partitions surrounding differences between primaries and secondaries. I need to reach out to the team to refresh my memory on that, so there is nothing here that addresses any of those differences and they'll be added in a further iterative PR along with any other minor details.

### DIFF
--- a/.changelog/11129.txt
+++ b/.changelog/11129.txt
@@ -1,0 +1,4 @@
+```release-note:improvement
+ui: Add initial support for partitions to intentions
+```
+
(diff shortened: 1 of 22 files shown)


hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/456474.

PR #10902, opened 2021-08-24
ui: Disabling policy form fields from users with 'read' permissions

kaxcode:
### 🐛 Description:
User with `read` permission can `edit` policies. An error message denying their request appears. The change temporarily appears in policy list page. The changes do not persist. Refreshing the page resets the changes.

### 💻 Replication Steps:

Using a new Consul cluster with ACLs enabled with default deny.

- Create the a new policy with the following rules:
acl = "read" 
operator = "read" 
service_prefix "" { policy = "write" }
node_prefix "" { policy = "write" }
agent_prefix "" { policy = "write" }
- Create a new Role
- Update the Anonymous Token to use the new Role and Policy
- Log off and log back in with Anonymous Token
- Try to update a Policy name

### 🎉 Solution:

Fix up a typo in the `canWrite()` permission for policy.

Now, the user can't edit any policies because form fields are disabled. 

### 🧪 Testing:

Test coverage can be found [here](https://github.com/hashicorp/consul/blob/main/ui/packages/consul-ui/tests/acceptance/dc/acls/policies/update.feature#L19. ).

### 📝 PR Tasks:

- [x]  Backport
- [x]  Changelog

PR #11132, opened 2021-09-23
ui: Adds a set of basic unit tests for abilities

johncowen:
This PR adds a very small unit test to unit test all of our current and _future_ abilities and stems from an idea here:

https://github.com/hashicorp/consul/pull/10902#issuecomment-905331481

> If we wanted to go lower we could just test the abilities themselves, which would make it easier to just loop through them all and check that read, write, 'create', 'update' permissions report correctly for each model type we have (plus easily test any future models to prevent any typos like we have here). We could almost test the unit itself if it wasn't for the inheritance chain, so we'd be ever so slightly be blurring into integration, but in this case I think doing some sort of unit testing of the abilities themselves (in a future looking way) might be a good idea 🤔 . What do you think?

The PR does exactly that ^, but even when writing it I was very much thinking 'is this actually something that is worthwhile' as the logic we are testing here is extremely simple. I think the benefit here is more likely something that we will notice in the future _if_ it catches any typos etc. Right now I'm still unsure whether long term it will be worthwhile but happy to have it hanging around for a few months to see if its beneficial, we can just delete it later if it's not worthwhile.

I actually would like our adapters to follow a similar testing approach now that they are almost logic-less. Our adapters originally were full of all sorts of sprawly logic, for which the tests we have were super beneficial, especially when we moved to a more logic-less less sprawly approach. They gave us the confidence that everything continued to work correctly during and after the refactor. The thing is now there is not much to test in our adapters, the tests almost get in the way (I've noted this a few times on various admin partition related PRs). So at some point in the not too distant future we should probably decide whether to move all our adapter tests to a much easier to work with approach similar to this one.

### DIFF
--- a/ui/packages/consul-ui/tests/unit/abilities/-test.js
+++ b/ui/packages/consul-ui/tests/unit/abilities/-test.js
@@ -1,0 +1,65 @@
+/* globals requirejs */
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module('Unit | Ability | *', function(hooks) {
+  setupTest(hooks);
+
+  // Replace this with your real tests.
+  test('it exists', function(assert) {
+    const abilities = Object.keys(requirejs.entries)
+      .filter(key => key.indexOf('/abilities/') !== -1)
+      .map(key => key.split('/').pop())
+      .filter(item => item !== '-test');
+    abilities.forEach(item => {
+      const ability = this.owner.factoryFor(`ability:${item}`).create();
+      [true, false].forEach(bool => {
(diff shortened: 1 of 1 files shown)


kaxcode (approved):
👍

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/456963.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit ce21c5109b5234e59ab70df299fcec01b569a94d onto `release/1.10.x` succeeded!

PR #11147, opened 2021-09-25
addition of default field to /v1/discovery-chain/:service endpoint

### LINKED ISSUE #10533: add a field to the /v1/discovery-chain/:service endpoint to indicate it is the default
Today if the discovery chain endpoint was generated off of no `service-router`, `service-splitter`, or `service-resolver` config entry it looks like this:

```
{
  "Chain": {
    "ServiceName": "blah",
    "Namespace": "default",
    "Datacenter": "dc1",
    "Protocol": "tcp",
    "StartNode": "resolver:blah.default.dc1",
    "Nodes": {
      "resolver:blah.default.dc1": {
        "Type": "resolver",
        "Name": "blah.default.dc1",
        "Resolver": {
          "ConnectTimeout": "5s",
          "Default": true,
          "Target": "blah.default.dc1"
        }
      }
    },
    "Targets": {
      "blah.default.dc1": {
        "ID": "blah.default.dc1",
        "Service": "blah",
        "Namespace": "default",
        "Datacenter": "dc1",
        "MeshGateway": {},
        "Subset": {},
        "SNI": "blah.default.dc1.internal.a470f28e-0477-3b21-61de-3e5d5801cd24.consul",
        "Name": "blah.default.dc1.internal.a470f28e-0477-3b21-61de-3e5d5801cd24.consul"
      }
    }
  }
}
```

You can infer that no content went into this by checking:
1. is the top node a resolver node?
2. is the `Resolver.Default` field `true`?
3. is the `Resolver.Target` pointing to a target in the `Targets` map that points to the same name/namespace as the requested discovery chain (and with no subset), in the same datacenter.

It would be convenient to populate a top level field like `Default bool` server-side so that consumers who care about the distinction can just read a single field instead of doing some field introspection.

bisakhmondal:
This PR adds a new `Default:bool` field to the  `/v1/discovery-chain/:service` endpoint to indicate if the discovery chain is default or not.

The logic to assert default to a discovery chain is (already been taken care of by `IsDefault` method)
> You can infer that no content went into this by checking:
> 1. is the top node a resolver node?
> 2. is the Resolver.Default field true?
> 3. is the Resolver.Target pointing to a target in the Targets map that points to the same name/namespace as the requested discovery chain (and with no subset), in the same datacenter.

closes #10533

### DIFF
--- a/agent/discovery_chain_endpoint.go
+++ b/agent/discovery_chain_endpoint.go
@@ -85,6 +85,7 @@ func (s *HTTPHandlers) DiscoveryChainRead(resp http.ResponseWriter, req *http.Re
 		}
 	}
 	out.ConsistencyLevel = args.QueryOptions.ConsistencyLevel()
+	out.Chain.Default = out.Chain.IsDefault()
 
 	return discoveryChainReadResponse{Chain: out.Chain}, nil
 }
--- a/agent/discovery_chain_endpoint_test.go
+++ b/agent/discovery_chain_endpoint_test.go
@@ -79,6 +79,7 @@ func TestDiscoveryChainRead(t *testing.T) {
 				Partition:   "default",
 				Datacenter:  "dc1",
 				Protocol:    "tcp",
+				Default:     true,
 				StartNode:   "resolver:web.default.default.dc1",
 				Nodes: map[string]*structs.DiscoveryGraphNode{
 					"resolver:web.default.default.dc1": {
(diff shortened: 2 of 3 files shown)


bisakhmondal:
ci error not related to the PR
> cannot connect to the Docker daemon. Is 'docker daemon' running on this host?: dial tcp 34.74.79.241:2376: connect: connection timed out

Issue #11012, opened 2021-09-10
Confusing action limitation on intentions


3nprob:
**Old UI or New UI**
New UI

**Describe the problem you're having**
Intention created directly via Terraform provider. Actions -> "This intention is being managed through an Intention Custom Resource in your Kubernetes cluster. It is view only in the UI."

There has never been a Kubernetes cluster or configuration in this environment.

![Screenshot_2021-09-10 Intentions - Consul](https://user-images.githubusercontent.com/74199244/132902917-fcf4c971-2e9f-499f-a1f4-a58b909c593a.png)

**Describe the solution you'd like**
See available actions, e.g. "delete"

johncowen:
Hi @3nprob 

Thanks for the report! Would it be possible for you to post the response from:

`/v1/connect/intentions?dc=datacenter-name`

When you see this happening in the UI? You can get this information from opening the Web Inspector in your browser and clicking the Network panel. In there you should be able to copy out the response from that API request.

Thanks,

reskin89:
Also seeing this on consul 1.10.2.

The endpoint returns the intention and it looks normal:
```json
  {
    "SourceNS": "default",
    "SourceName": "REDACTED-SOURCE",
    "DestinationNS": "default",
    "DestinationName": "REDACTED-DESTINATION",
    "SourceType": "consul",
    "Action": "allow",
    "Precedence": 9,
    "CreateIndex": 4486909,
    "ModifyIndex": 6421921
  }
  ```

johncowen:
Hi @reskin89 👋 

Is that the response from single intention endpoint or the intentions listing endpoint?

reskin89:
It's the response from the intentions endpoint, I pasted one of the returns I'm seeing this message on. They all look similar aside from the time stamps and source/dest names.

It's all just an array of these objects no other info

johncowen:
Ok cool, thanks @reskin89 . Can I get some more information from you, I'm having a tough time reproducing this.

What browser/version are you using?

reskin89:
Thanks John!

We're seeing it in Firefox 92 and Chrome Version 93.0.4577.82

Oddly we're also noticing tokens with the write permissions on intentions aren't able to create intentions from the topology screen with the red X like they should.  It believes the token they have is read only and we've confirmed it is not.

I'm not sure if they're related

johncowen:
[...]

PR #11061, opened 2021-09-16
acl: fix intention:*:write checks

rboyer:
This is a partial revert of #10793

PR #11149, opened 2021-09-27
ui: Don't show the CRD menu for read-only intentions

johncowen:
Thanks to the amazing folks in https://github.com/hashicorp/consul/issues/11012 I realised that we had a small UI bug here that was further complicated by a problem with our `/v1/internal/acl/authorize` endpoint which was recently fixed here https://github.com/hashicorp/consul/pull/11061 . I initially thought this authorize endpoint error was in a 1.10.x release but looking a little harder, it seems like it isn't, it's only in our recent 1.11 alpha.

The UI bug here manifests itself only when a user/token is configured to have read-only access to intentions. Instead of only letting folks click to see a read only page of the intention, we would show an additional message saying that the intention was read-only due to it being 'Managed by [a kubernetes] CRD'. Whilst the intention was still read only, this extra message was still confusing for users.

This PR fixes up the conditional logic and further moves the logic to use `ember-can` - looking at the history of the files in question, this bug snuck itself in partly due to it being 'permission-y type stuff' previous to using `ember-can` and when something being editable or not was nothing to do with ACLs. Then we moved to start using `ember-can` without completely realising what `IsEditable` previously meant. So overall the code here is a tiny bit clearer/cleaner by adding a proper `can view CRD intention` instead of overloading the idea of 'editability'.

Backports-wise the bug only manifests itself in 1.10.x onwards as we only started using the authorize endpoint + ACLs permissions from 1.10.x onwards.

fixes https://github.com/hashicorp/consul/issues/11012

### DIFF
--- a/.changelog/11149.txt
+++ b/.changelog/11149.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Don't show a CRD warning for read-only intentions
+```
--- a/ui/packages/consul-ui/app/abilities/intention.js
+++ b/ui/packages/consul-ui/app/abilities/intention.js
@@ -4,6 +4,9 @@ export default class IntentionAbility extends BaseAbility {
   resource = 'intention';
 
   get canWrite() {
-    return super.canWrite && (typeof this.item === 'undefined' || this.item.IsEditable);
+    return super.canWrite && (typeof this.item === 'undefined' || !this.canViewCRD);
+  }
+  get canViewCRD() {
+    return (typeof this.item !== 'undefined' && this.item.IsManagedByCRD);
   }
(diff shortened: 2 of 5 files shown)


kaxcode (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/457098.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit bce724baf14768428a9f8f1b1b0a7d61b95316a3 onto `release/1.10.x` succeeded!

PR #11153, opened 2021-09-27
ui: Fixup route blueprint to reflect project conventions

johncowen:


kaxcode:
@johncowen does it break if you don't have the `ui/packages/consul-ui/blueprints/route-test/index.js` file?

kaxcode:
> Nah, I just tried it again, seems to work fine?

@johncowen why is the empty file there then?

PR #11158, opened 2021-09-27
wrap doRequest calls for error handling

FFMMM:
### Overview

This PR builds on https://github.com/hashicorp/consul/pull/11054 to wrap all remaining `doRequest()` calls around a `require*()` call.

### Issue Related
https://github.com/hashicorp/consul/issues/10865

### Notes:
* mostly refactoring, no logic changes

### PR Checklist
* [x] `go fmt`
* [x] `go mod`
* [x] n/a pr/changelog -- no "public" api changes

### DIFF
--- a/api/agent.go
+++ b/api/agent.go
@@ -2,8 +2,8 @@ package api
 
 import (
 	"bufio"
-	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -602,6 +602,8 @@ func (a *Agent) AgentHealthServiceByIDOpts(serviceID string, q *QueryOptions) (s
 	r.setQueryOptions(q)
 	r.params.Add("format", "json")
 	r.header.Set("Accept", "application/json")
+	// not a lot of value in wrapping the doRequest call in a requireHttpCodes call
+	// we manipulate the resp body  and the require calls "swallow" the content on err
 	_, resp, err := a.c.doRequest(r)
 	if err != nil {
 		return "", nil, err
@@ -641,6 +643,8 @@ func (a *Agent) AgentHealthServiceByNameOpts(service string, q *QueryOptions) (s
 	r.setQueryOptions(q)
 	r.params.Add("format", "json")
 	r.header.Set("Accept", "application/json")
+	// not a lot of value in wrapping the doRequest call in a requireHttpCodes call
+	// we manipulate the resp body  and the require calls "swallow" the content on err
 	_, resp, err := a.c.doRequest(r)
 	if err != nil {
 		return "", nil, err
@@ -1233,19 +1237,20 @@ func (a *Agent) updateTokenOnce(target, token string, q *WriteOptions) (*WriteMe
 	r.setWriteOptions(q)
 	r.obj = &AgentToken{Token: token}
 
-	rtt, resp, err := a.c.doRequest(r)
-	if err != nil {
-		return nil, 0, err
-	}
-	defer closeResponseBody(resp)
-
+	rtt, resp, err := requireOK(a.c.doRequest(r))
 	wm := &WriteMeta{RequestTime: rtt}
 
-	if resp.StatusCode != 200 {
-		var buf bytes.Buffer
-		io.Copy(&buf, resp.Body)
-		return wm, resp.StatusCode, fmt.Errorf("Unexpected response code: %d (%s)", resp.StatusCode, buf.Bytes())
+	if err != nil {
+		// if the error was bc of a non 200 response
+		// from requireOK
+		var statusE StatusError
+		if errors.As(err, &statusE) {
+			return wm, statusE.Code, statusE
+		}
+		// otherwise, the error came via doRequest
+		return nil, 500, err
 	}
+	defer closeResponseBody(resp)
 
 	return wm, resp.StatusCode, nil
 }
--- a/api/kv.go
+++ b/api/kv.go
@@ -129,6 +129,8 @@ func (k *KV) getInternal(key string, params map[string]string, q *QueryOptions) 
 		r.params.Set(param, val)
 	}
 	rtt, resp, err := k.c.doRequest(r)
+	rtt, resp, err = requireHttpCodes(rtt, resp, err, 200, 404)
+
 	if err != nil {
 		return nil, nil, err
 	}
--- a/api/kv.go
+++ b/api/kv.go
@@ -142,10 +142,8 @@ func (k *KV) getInternal(key string, params map[string]string, q *QueryOptions) 
 	if resp.StatusCode == 404 {
 		closeResponseBody(resp)
 		return nil, qm, nil
-	} else if resp.StatusCode != 200 {
-		closeResponseBody(resp)
-		return nil, nil, fmt.Errorf("Unexpected response code: %d", resp.StatusCode)
 	}
+
 	return resp, qm, nil
 }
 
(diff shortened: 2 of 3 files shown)


api/agent.go
```diff
@@ -641,6 +641,7 @@ func (a *Agent) AgentHealthServiceByNameOpts(service string, q *QueryOptions) (s
 	r.setQueryOptions(q)
 	r.params.Add("format", "json")
 	r.header.Set("Accept", "application/json")
+	// not a lot of value in wrapping the doRequest call in a requireHttpCodes call
```
freddygv, line 644: Why's that?

freddygv (commented):
Thanks for following up here! I left some questions below

api/agent.go
```diff
@@ -1237,15 +1238,15 @@ func (a *Agent) updateTokenOnce(target, token string, q *WriteOptions) (*WriteMe
...
-	defer closeResponseBody(resp)
 
 	wm := &WriteMeta{RequestTime: rtt}
 
-	if resp.StatusCode != 200 {
-		var buf bytes.Buffer
-		io.Copy(&buf, resp.Body)
-		return wm, resp.StatusCode, fmt.Errorf("Unexpected response code: %d (%s)", resp.StatusCode, buf.Bytes())
+	_, _, err = requireOK(rtt, resp, err)
+
```
kisunji, line 1245: minor: remove newline here

api/agent.go
```diff
@@ -1237,15 +1238,15 @@ func (a *Agent) updateTokenOnce(target, token string, q *WriteOptions) (*WriteMe
...
+	if err != nil {
+		return wm, resp.StatusCode, err
```
kisunji, line 1247: I'm a little confused about the order in which `requireOK` is called here - should this be handled close to L1238 right after `doRequest(r)`?

api/agent.go
```diff
@@ -602,6 +601,7 @@ func (a *Agent) AgentHealthServiceByIDOpts(serviceID string, q *QueryOptions) (s
 	r.setQueryOptions(q)
 	r.params.Add("format", "json")
 	r.header.Set("Accept", "application/json")
+	// not a lot of value in wrapping the doRequest call in a requireHttpCodes call
```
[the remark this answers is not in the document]
FFMMM: Good point! There's two ways I understand this comment:

1. Should we return a `StatusError` from the `require` calls and handle them inside the body of the func?

2. Should we return a `StatusError` our of this func?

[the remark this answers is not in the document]
FFMMM: Let's chat about 1) first.

From a cursory look, I was worried that we'd change the return format.

So I was thinking that some of callees of these two API paths (ie `/v1/agent/health/service/id/%v`) were error handling for the old err messages. 

In order to keep the same format, we'd have to wrap it the same so it'd end up looking something like

```
return HealthCritical, out, fmt.Errorf("Unexpected Error Code %v for %s", err.StatusCode, path)
```
--
That's why I also said that there isn't a lot of value to wrapping the call since there is sufficient "behavior" handling and a different error message returned.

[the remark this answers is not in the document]
FFMMM: For 2), I think we could get away with sending out an `*api.StatusError`. It's just a matter of whether it makes sense to expose that. I've not looked too hard into what changes would be required to do so if any at all.

What do you think ?

api/agent.go
FFMMM: tried to answer and consolidate the convo here --> https://github.com/hashicorp/consul/pull/11158/files#r717131243

api/agent.go
FFMMM: hm, I can certainly do that but I was hoping that any style nits were taken care of by `gofmt`

api/kv.go
```diff
@@ -129,6 +129,8 @@ func (k *KV) getInternal(key string, params map[string]string, q *QueryOptions)
 		r.params.Set(param, val)
 	}
 	rtt, resp, err := k.c.doRequest(r)
+	rtt, resp, err = requireHttpCodes(rtt, resp, err, 200, 404)
```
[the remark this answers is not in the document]
FFMMM: It does! I thought I had cleaned up the deadcode but I didn't it seems!

Let me rework that.

api/agent.go
FFMMM: Yes I think so -- thanks for spotting that!

PS: I think when I chose not to I forgot we actually check the err from `doRequest()` inside the require calls!

https://github.com/hashicorp/consul/blob/5c37819d0957f54242d46650e74756346e4e1b2c/api/api.go#L1107-L1111

FFMMM: Let me follow up here to explain the old odd/ smelly ordering!

So in the snippet above, if there is an error during the `doRequest()`, and we swallow the `resp` object and throw it back as `nil`. 

So there was not an easy way to combine the first err check, where we want to return 0 for some reason to the callee of `udpateTokenOnce()`, with the synthetic error checks.

As of now, I tried to re-write the snippet of code to make more sense -- hopefully it's clearer now but please let me know if it's not.

api/agent.go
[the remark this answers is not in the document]
kisunji: We should consistently return `api.StatusError`, especially in cases where error strings look similar (otherwise it might mislead clients to believe they're dealing with the custom type)

dnephin:
I haven't had a chance to catch up on this discussion, but I think that suggested refactor makes sense at a high level. Assuming the common case is `requireOk` we could even add a `doRequestRequireOk` that contains most of those lines. For the few places that don't call `requireOk` we can keep them as separate calls.

FFMMM:
hey @kisunji and @dnephin thank y'all for chiming in! let me agree with y'all below with some context and proposed  path:

> I think there's some refactoring that may need to be done for requireOK + requireHttpStatusCodes.

Yes that's actually something that @markan and I were chatting about when looking at the `api/agent.go` code. It's weird ergonomics that the `require()` funcs close the response body for `doRequest` errors and not for other errors.

I think I'd love to have a chance to address this refactor in a subsequent PR.

> Assuming the common case is requireOk we could even add a doRequestAndRequireOk that contains most of those lines. 

There's actually >130 usages of `requireOK()` calls when I looked at that yesterday. I'd argue that a refactor that does `doRequestAndRequireOk` would be a big diff and even one that touches the internals of `requireOK `to modify behavior would warrant its own separate PR. 

---

I'm hoping that we can merge this PR if it makes sense and I can take on the work to have a stab at refactoring the code in a subsequent PR. What do y'all think? 👍🏼 or 👎🏼  ?

kisunji (approved):
Sounds reasonable to defer the refactor to another PR!

low-prio todos from my perspective:
* revisit error-handling of `requireOK` to be more clear/explicit
* use `api.StatusError` in all `fmt.Errorf`'s where we embed the status code in the error message

api/agent.go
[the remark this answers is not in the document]
FFMMM: > I was mostly asking about 2), since the error string looks similar to the one from StatusError

> We should consistently return api.StatusError, especially in cases where error strings look similar (otherwise it might mislead clients to believe they're dealing with the custom type)

+1; should be something that I can address per our chat below in a separate PR

merging these changes for now.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/461161.

PR #11163, opened 2021-09-28
Add support for enabling connect-based ingress TLS per listener.

banks:
This adds support for a case I missed in #10903.

This is technically new functionality not directly related to SDS so I'll add a separate changelog entry but with the addition of nested TLS blocks in more places it makes sense for the fields there to actually function in a reasonable way.

This behaviour was proposed in the RFC. It adds the ability to use the built-in Connect TLS for only a subset of Ingress listeners which might be useful to some people but mostly is to avoid confusing disparity between built-in TLS being all or nothing while SDS is more fine-grained.


agent/xds/listeners_ingress.go
```diff
@@ -118,6 +127,49 @@ func (s *ResourceGenerator) makeIngressGatewayListeners(address string, cfgSnap
...
+	case mergedCfg.ClusterName == "" && mergedCfg.CertResource != "":
+		return nil, fmt.Errorf("missing SDS cluster name for listener on port %d", listenerKey.Port)
+
+	case mergedCfg.ClusterName != "" && mergedCfg.CertResource == "":
+		return nil, fmt.Errorf("missing SDS cert resource for listener on port %d", listenerKey.Port)
+	}
+
+	return &mergedCfg, nil
+}
+
```
banks, line 172: This is just a straight move from the ingress file - I missed this ingress-specific function when refactoring into a separate file in the last PR.

agent/xds/listeners_ingress.go
```diff
@@ -19,6 +19,15 @@ func (s *ResourceGenerator) makeIngressGatewayListeners(address string, cfgSnap
 	for listenerKey, upstreams := range cfgSnap.IngressGateway.Upstreams {
 		var tlsContext *envoy_tls_v3.DownstreamTlsContext
 
+		listenerCfg, ok := cfgSnap.IngressGateway.Listeners[listenerKey]
```
rboyer, line 22: Why fetch this here and then all over again in `resolveListenerSDSConfig` when you could just pass it down through?

banks: I did start that refactor but I ended up backing out as it ended up changing quite a few lines that reference `listenerKey` all just to save one map lookup... It doesn't allow us to remove the `error` return which was my initial motivation because there are other possible error sources too.

I can do that though if you think the code is significantly better that way?

banks:
@rboyer thanks for taking a look! I made the refactor you suggested let me know if there's anything else you can see blocking a merge!

PR #11047, opened 2021-09-15
[Assembly] Consul on Kubernetes Page

pbortnick:
🔍 [Preview Link](https://consul-git-pbassembly-consul-on-kubernetes-hashicorp.vercel.app/consul-on-kubernetes)
💅 [Design Reference](https://www.figma.com/file/0DT6uKPK214eFZtMG0pQUA/Consul-on-Kubernetes?node-id=102%3A6583)

This is a PR for the assembled Consul on Kubernetes page.

PR #11168, opened 2021-09-28
Consul On Kubernetes - Card List & Get Started Section 

pbortnick:
🎟️ [Asana Task](https://app.asana.com/0/1200785913255669/1200992155224295/f)
🔍 [Preview Link](https://consul-git-pbcard-list-hashicorp.vercel.app/consul-on-kubernetes)
💅 [Design Reference](https://www.figma.com/file/0DT6uKPK214eFZtMG0pQUA/Consul-on-Kubernetes?node-id=102%3A6583)

This PR contains the Card List and bottom section (**Ways to get started**) styles.

This is a feature PR targeting assembly branch: #11047

### DIFF
--- a/website/components/card-list/index.tsx
+++ b/website/components/card-list/index.tsx
@@ -26,12 +26,12 @@ export default function CardList({ title, cards }: CardListProps) {
       <h3 className={s.title}>{title}</h3>
       <div className={s.cardsWrapper}>
         {cards.map(({ image, description, cta, type }) => (
-          <div key={type} className={s.card}>
+          <a href={cta.url} key={type} className={s.card}>
             <div className={s.imageContainer}>
               <img src={image.src} alt={image.alt} />
             </div>
             <div className={s.cardContent}>
-              <div className={s.eyebrow}>
+              <div className={s.preHeader}>
                 <span>{type}</span>
                 <Button
                   title=""
@@ -44,11 +44,9 @@ export default function CardList({ title, cards }: CardListProps) {
                 />
               </div>
               <span className={s.description}>{description}</span>
-              <Link href={cta.url}>
-                <a>{cta.text}</a>
-              </Link>
+              <span className={s.fauxLink}>{cta.text}</span>
             </div>
-          </div>
+          </a>
         ))}
       </div>
     </>
--- a/website/components/card-list/index.tsx
+++ b/website/components/card-list/index.tsx
@@ -19,6 +19,7 @@ interface CardListProps {
   title: string
   cards: Card[]
 }
+
 export default function CardList({ title, cards }: CardListProps) {
   return (
     <>
--- a/website/components/card-list/index.tsx
+++ b/website/components/card-list/index.tsx
@@ -1,4 +1,5 @@
 import s from './style.module.css'
+import Button from '@hashicorp/react-button'
 
 interface Card {
   image: {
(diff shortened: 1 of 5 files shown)


website/components/card-list/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  image: {
+    src: string
+    alt: string
+  }
+  description: string
+  cta: {
+    url: string
+    text: string
+  }
+  type: string
```
BrandonRomano, line 13: _type_ slightly assumes the content here.  I'm actually not 100% sure what will go in here, but I am suspicious that it might be something that `type` will not read well with.

Maybe something generic to just describe the visual pattern, maybe like `eyebrow`?

website/components/card-list/style.module.css
```diff
@@ -0,0 +1,57 @@
+.cardsWrapper {
+  display: grid;
+  column-gap: 40px;
+  row-gap: 40px;
+  grid-template-columns: repeat(auto-fill, minmax(272px, 1fr));
+
+  & .card {
+    border: 1px solid #dcdde0;
```
BrandonRomano, line 8: Do we have a --gray variable for this value here?

website/pages/consul-on-kubernetes/index.tsx
```diff
@@ -230,39 +231,90 @@ export default function ConsulOnKubernetesPage() {
...
-                url: '#TODO',
-              },
-            },
-          ]}
-        />
+      <section className={s.getStartedWrapper}>
+        <h1 className={s.getStartedTitle}>Ways to get started</h1>
+        <div className={s.getStartedContent}>
+          <div className={s.getStartedCards}>
+            <CardList
```
BrandonRomano, line 238: The breakpoints between `1120 --> 1225` feel a little odd to have a single row of cards w/ documentation side-by-side here...  Not a super strong opinion.

CleanShot 2021-09-28 at 11 25 05@2x

website/components/card-list/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+}
+
+interface CardListProps {
+  title: string
+  cards: Card[]
+}
+
+export default function CardList({ title, cards }: CardListProps) {
+  return (
+    <>
```
BrandonRomano, line 23: Having a component returning multiple top-level objects will cause most of your consumers to have to wrap your component in a div (which is some complexity we could avoid).

I would probably prefer wrapping this component in a div itself, and potentially (if needed) accept a className to this component, and pass it to that div.

website/components/card-list/style.module.css
```diff
@@ -0,0 +1,57 @@
...
+    box-shadow: 0 2px 3px rgba(37, 41, 55, 0.08);
+    border-radius: 1px;
+    transition: box-shadow 0.25s, transform 0.25s;
+
+    &:hover {
+      box-shadow: 0 16px 28px rgba(37, 38, 45, 0.12);
+      transform: translateY(-4px);
```
BrandonRomano, line 15: Looks good!

BrandonRomano (approved):
Looks good!  Had some feedback, but none of it is mandatory / would hold up my approval.

If you feel so inspired to make changes after reading my comments, feel free to make the changes. Tap me again if you'd like me to review those.

website/components/card-list/style.module.css
pbortnick: I checked but didn't see one in [the global colors](https://github.com/hashicorp/mktg-global-styles/blob/main/custom-properties/color.css)

website/components/card-list/index.tsx
pbortnick: I agree and like the className idea idea

website/components/card-list/style.module.css
BrandonRomano: Often what I'll do here is just grab the most similar gray color; usually there's something visually close.

If not though, using a custom color is OK.

pbortnick: 👍 gray-5 looks very similar, we can go with that 
Screen Shot 2021-09-28 at 2 52 44 PM

website/pages/consul-on-kubernetes/index.tsx
pbortnick: Maybe we should `wrap` at a larger breakpoint?

pbortnick: Actually, we can decrease the min in `grid-template-columns` which makes it look smoother

website/components/docs-list/index.tsx
```diff
@@ -15,11 +15,16 @@ interface Doc {
...
 
-export default function DocsList({ title, docs }: DocsListProps) {
+export default function DocsList({
+  title,
+  docs,
+  className = null,
+}: DocsListProps) {
   return (
-    <>
+    <div className={className}>
```
BrandonRomano, line 27: Nice I like how you applied this here too ✨

BrandonRomano (approved):

PR #11175, opened 2021-09-29
Update docs: correct provider argument is tag_key

kevinwojo:


kevinwojo:
> 🤔 This PR has changes in the `website/` directory but does not have a `type/docs-cherrypick` label. If the changes are for the next version, this can be ignored. If they are updates to current docs, attach the label to auto cherrypick to the `stable-website` branch after merging.

This is for the current docs, I don't have the ability to label `stable-website`

karl-cardenas-coding (approved):

eculver:
Hey, thanks for the contribution. I looked into it and unless I'm missing something, I think the existing docs are correct. We get these keys from [`hashicorp/go-discover`](https://github.com/hashicorp/go-discover) and they appear to be correct by using `tag_name` and `tag_value`. See [the code in go-discover](https://github.com/hashicorp/go-discover/blob/master/provider/azure/azure_discover.go#L106) for reference. 

Can you provide a little context on where `tag_key` is coming from?

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/588986.

PR #11181, opened 2021-09-29
agent: Reject partitions in legacy intention endpoints

kisunji:
Legacy Create and SpecificUpdate endpoints should not handle Source- or Destination- partitions (in either OSS or ENT).

### DIFF
--- a/agent/intentions_endpoint.go
+++ b/agent/intentions_endpoint.go
@@ -67,7 +67,13 @@ func (s *HTTPHandlers) IntentionCreate(resp http.ResponseWriter, req *http.Reque
 		return nil, fmt.Errorf("Failed to decode request body: %s", err)
 	}
 
-	// TODO(partitions): reject non-empty/non-default partitions from the decoded body
+	if args.Intention.DestinationPartition != "" && args.Intention.DestinationPartition != "default" {
+		return nil, BadRequestError{Reason: "Cannot specify a destination partition with this endpoint"}
+	}
+	if args.Intention.SourcePartition != "" && args.Intention.SourcePartition != "default" {
+		return nil, BadRequestError{Reason: "Cannot specify a source partition with this endpoint"}
+	}
+
 	args.Intention.FillPartitionAndNamespace(&entMeta, false)
 
 	if err := s.validateEnterpriseIntention(args.Intention); err != nil {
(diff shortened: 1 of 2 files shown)


agent/intentions_endpoint.go
```diff
@@ -67,7 +67,13 @@ func (s *HTTPHandlers) IntentionCreate(resp http.ResponseWriter, req *http.Reque
 		return nil, fmt.Errorf("Failed to decode request body: %s", err)
 	}
 
-	// TODO(partitions): reject non-empty/non-default partitions from the decoded body
+	if args.Intention.DestinationPartition != "" && args.Intention.DestinationPartition != "default" {
+		return nil, BadRequestError{Reason: "Cannot specify a destination partition with this endpoint"}
+	}
+	if args.Intention.SourcePartition != "" && args.Intention.SourcePartition != "default" {
+		return nil, BadRequestError{Reason: "Cannot specify a source partition with this endpoint"}
+	}
```
kisunji, line 75: Can't think of an ENT/OSS-friendly clean way of doing this 😢

dhiaayachi: you can use `structs.PartitionOrDefault` for this

agent/intentions_endpoint_test.go
```diff
@@ -428,6 +428,25 @@ func TestIntentionCreate(t *testing.T) {
...
 	})
+
+	t.Run("partition rejected", func(t *testing.T) {
+		{
+			args := structs.TestIntention(t)
+			args.SourcePartition = "part1"
+			req, _ := http.NewRequest("POST", "/v1/connect/intentions", jsonReader(args))
+			resp := httptest.NewRecorder()
+			_, err := a.srv.IntentionCreate(resp, req)
+			require.Error(t, err)
```
dhiaayachi, line 439: It would be interesting to check for the error message

dhiaayachi (commented):
Thank you for working on this @kisunji, some minor comments

agent/intentions_endpoint.go
kisunji: I tried using it like below but in OSS it fails because `PartitionOrDefault` always returns `"default"` and therefore we can't capture the error case where a user has specified a non-empty, non-default partition.

```go
// oss: args.Intention.SourcePartition = "invalidpartition"

if structs.PartitionOrDefault(args.Intention.SourcePartition) != "default" {
    return nil, BadRequestError{Reason: "Cannot specify a source partition with this endpoint"}
}

// oss: passes through (bug)
```

freddygv (approved):

agent/intentions_endpoint.go
dhiaayachi: It's the only endpoint I know of that don't error on partition or ns passed in oss, but because this is a legacy endpoint I think it's not worth fixing

dhiaayachi (approved):
LGTM!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/462426.

PR #11182, opened 2021-09-29
acl: remove upgrade from legacy, start in non-legacy mode

dnephin:
Branched from #11126

This PR removes the rest of the "start in legacy" logic from `Server`. The upgrade token goroutine is kept around for cases where a user may restore from a snapshot with legacy tokens, or when a legacy token is created immediately before the server is shutdown for an upgrade.

### DIFF
--- a/agent/consul/acl.go
+++ b/agent/consul/acl.go
@@ -54,14 +54,6 @@ const (
 	// are not allowed to be displayed.
 	redactedToken = "<hidden>"
 
-	// aclUpgradeBatchSize controls how many tokens we look at during each round of upgrading. Individual raft logs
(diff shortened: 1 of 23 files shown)


freddygv (approved):

dnephin:
I added one more small commit to remove `updateEnterpriseSerfTags`, which I noticed is no longer needed because the only caller is a test helper, and the test passed in enterprise without `updateEnterpriseSerfTags`.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/463600.

PR #11183, opened 2021-09-29
acl: remove the remaining parts of structs/acl_legacy.go

dnephin:
Branched from #11182

Removes the rest of `structs/acl_legacy.go`. The last references to these structs were removed in the last few PRs. There is one that is still referenced from snapshot restore, which was moved next to the one place it is used.

There are two constants left in this file, which will be removed in the next PR.


agent/consul/fsm/snapshot_oss.go
```diff
@@ -574,9 +575,51 @@ func restoreACL(header *SnapshotHeader, restore *state.Restore, decoder *codec.D
...
+		Rules:             a.Rules,
+		Local:             false,
+		RaftIndex:         a.RaftIndex,
+	}
+
+	token.SetHash(true)
+	return token
+}
+
+// TODO(ACL-Legacy-Compat) - remove in phase 2
```
freddygv, line 612: Isn't ACL bootstrapping still done? why would the associated restore be removed later?

dnephin: Ah, yes it is. I thought we used a different request for the new system, but it appears we do still use this one. So I'll remove this TODO, and the other one on the request type.

dnephin: Actually, looking at this closer, I think this TODO is correct.

We use these request types in two ways: 
1. for routing FSM commands
2. for snapshot/restore

We still need   `ACLBootstrapRequestType` for the FSM, but this restore function is only for the legacy system. The new system writes to `tableIndex` on bootstrap, which is handled by 

```
registerRestorer(structs.IndexRequestType, restoreIndex)
```

So this TODO is correct, but the one above on `ACLBootstrapRequestType` is wrong. I'll remove that TODO.

agent/consul/fsm/snapshot_oss.go
```diff
@@ -15,8 +15,8 @@ func init() {
 	registerRestorer(structs.KVSRequestType, restoreKV)
 	registerRestorer(structs.TombstoneRequestType, restoreTombstone)
 	registerRestorer(structs.SessionRequestType, restoreSession)
-	registerRestorer(structs.DeprecatedACLRequestType, restoreACL)
-	registerRestorer(structs.ACLBootstrapRequestType, restoreACLBootstrap)
+	registerRestorer(structs.DeprecatedACLRequestType, restoreACL)         // TODO(ACL-Legacy-Compat) - remove in phase 2
+	registerRestorer(structs.ACLBootstrapRequestType, restoreACLBootstrap) // TODO(ACL-Legacy-Compat) - remove in phase 2
```
dnephin, line 19:
```suggestion
	registerRestorer(structs.ACLBootstrapRequestType, restoreACLBootstrap)
```

agent/consul/leader.go
```diff
@@ -344,6 +344,8 @@ func (s *Server) revokeLeadership() {
 
 	s.stopConfigReplication()
 
+	s.stopACLReplication()
```
dnephin, line 347: Calling attention to this. Seems like this has always been missing? I noticed there were no longer any calls to `stopACLReplication` after the last PR merged, and it seems like it should be called here.

PR #11188, opened 2021-09-30
ui: Add `is` and `test` helpers in a similar vein to `can`

johncowen:
First of a series of iterative PRs for adding CRUD for partitions.

TBD

### DIFF
--- a/ui/packages/consul-ui/app/helpers/is.js
+++ b/ui/packages/consul-ui/app/helpers/is.js
@@ -1,0 +1,24 @@
+import Helper from 'ember-can/helpers/can';
+import { get } from '@ember/object';
+
+import { camelize } from '@ember/string';
+export const is = (helper, [abilityString, model], properties) => {
+  let { abilityName, propertyName } = helper.can.parse(abilityString);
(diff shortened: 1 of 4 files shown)


kaxcode:
@johncowen Why is the `test` helper needed? You mention mostly using the `is` helper and the `can` helper.

johncowen:
> Why is the test helper needed? You mention mostly using the is helper and the can helper.

Looking forwards, I would imagine (and be in favour of) we remove the `is` and `can` helpers entirely (probably). But right now its very easy to bring in the 'more ergonomic' `test` helper and then slowly deprecate the others over time. Right now, its not important to do that and it would mean a large change over the entire codebase all in one go (we use `can` a lot). I'd rather support the 'old' way and the 'new' way together for a while and then deprecate the 'old' way.

Just a little offshoot here, using the `can` helper into the future has one 'pro' going for it, and that is it's relatively a 'pseudo-standard' amongst our codebases. I thought I'd mention that, whilst it wouldn't stop me from advocating against using it in favour of the more flexible/ergonomic `test` helper over it, it's worthwhile pointing that 'pro' out.

kaxcode (approved):
👍

PR #11200, opened 2021-10-01
ui: Replaces almost all remaining instances of SASS variables with CSS

johncowen:
This PR is 99.99% an automated change done by machine and should have zero effect on the user facing application itself. It's a modernisation PR that lays for ground for us for some time to come and prepares us for integrating with certain other workstreams coming down the line at some point. Additionally, as we are currently in alpha releases period, this feels like the perfect time to be PRing something like this (we have no automated CSS testing as yet).

This change follows on from:

- https://github.com/hashicorp/consul/pull/7249
- https://github.com/hashicorp/consul/pull/10298
- ...and possibly some others

...and basically converts almost all our old remaining SASS variable usage to use native CSS custom properties.

It's 'almost all' as we left some vars in:

- Our code editor syntax highlighting. This was originally just taken from vault-ui and has never been changed since then. It uses some other SASS features that we don't use anywhere else in the Consul UI and I noticed whilst doing this it has a bit of confusing double setting of variables, which paired with the use of SASS's `lighten` function made me decide to leave this file for a later PR as it needs a human to go through it and tweak some bits, at which point we may as well use our own colors to replicate (or change?) the color/theme for syntax highlighting there. I know there's an upcoming feature to work on our code editor, so that might be a better opportunity to visit this file.
- We emulate CSS custom-media rules using SASS vars. These are soon to be replaced by some iteration/version of container queries. It would be best to do that all at once. Alternatively if we chose to move away from SASS entirely sooner we could use PostCSS to bundle our files and use its custom-media query support instead of emulating it ourselves.

All in all, this leaves gives us the _option_ of moving entirely to use PostCSS for bundling CSS instead of SASS if we choose to do so (currently, until we move to using `:host` etc we still use the non-native `@extend` and `%placeholders` which are also supported by PostCSS).

I've also added @junyper as reviewer here as she has been instrumental in helping me push this modernisation forwards, and I'm grateful for the positive impact (and consequently the impact on the UI as a whole) and discussion we've had on this , this PR will probably be one of a few similar ones as a result of those discussions.

From an engineers perspective, whenever specifying colors from now on we should use the form:

```css
color: rgb(var(--tone-red-500));
```

Please note:

- Use `rgb`. This lets us do this like `rgb(var(--tone-red-500) / 10%)` so we can use a 10% opacity red-500 if we ever need to whilst still making use of our color tokens.
- Use `--tone-colorName-000` (so the prefix `tone`). Previously we could use a mix of `--gray-500: $gray-500` (note the left hand CSS prop and right hand SASS var) for the things we  need to theme currently. As we no longer use SASS we can't do `--gray-500: --gray-500`, so we now do `--tone-gray-500: --gray-500`.

Just for clarity after that, whenever specifying a color anywhere, use `rgb` and `--tone`. There is only one reason where you might not use `tone`, and that is if you never want a color to be affected by a theme (for example a background shadow probably always should use `--black`)

Lastly, please keep in mind this was almost all done via a machine/automated script, so its the same change over and over and over again.

@amyrlam little FYI here also

### DIFF
--- a/.changelog/11200.txt
+++ b/.changelog/11200.txt
@@ -1,0 +1,5 @@
+```release-note:improvement
+ui: Move the majority of our SASS variables to use native CSS custom
+properties
+```
+


amyrlam (commented):
thanks for the tag, sgtm but will leave the official review to @kaxcode  

@junyper it was nice working with you, hope to cross paths down the road!

PR #11151, opened 2021-09-27
nia/docs: Callouts for v0.4.0-beta

mkam:
TFC callouts:
https://consul-rf5ihwjmn-hashicorp.vercel.app/docs/nia/network-drivers
https://consul-rf5ihwjmn-hashicorp.vercel.app/docs/nia/network-drivers/terraform-cloud
https://consul-rf5ihwjmn-hashicorp.vercel.app/docs/nia/configuration#terraform-cloud-driver

Consul KV callouts:
https://consul-rf5ihwjmn-hashicorp.vercel.app/docs/nia/tasks#consul-kv-condition
https://consul-rf5ihwjmn-hashicorp.vercel.app/docs/nia/configuration#consul-kv-condition
https://consul-rf5ihwjmn-hashicorp.vercel.app/docs/nia/terraform-modules#consul-kv-variable

PR #11209, opened 2021-10-04
nia/docs: Remove references to v0.4.0-beta

mkam:
Removes beta callouts in preparation for GA release. Essentially reverts https://github.com/hashicorp/consul/pull/11151


findkim (approved):
Double checked and looks like all the callouts are caught 👍👍

PR #11211, opened 2021-10-04
ui: Exclude all debug-like files from the build

johncowen:
Luckily I have a thing I was playing with that made me realise we had ballooned our vendor javascript file from about 430kb gzipped to about 1MB gzipped / 3.3MB un-gzipped (around double our normal size). Pretty sure I'd managed to import the whole of `faker.js` 😬 .

This PR adds `**/*-debug.*` to our test/prod excluded files (realised I needed to add `test-support.js` also so added that here as its more or less the same thing). Conditionally juggling ES6 static imports (specifically debug ones) for this was also getting a little hairy, so I moved it all to use the same approach as our conditional routes. All in all it brings the vendor build back down to ~430kb gzipped.

I wasn't sure whether to pop this on the end of https://github.com/hashicorp/consul/pull/11188 but since thats already an umbrella issue, I decided to go with a separate PR (will add a new TODO over in 11188). Lastly, theres a little evolution here, and there's probably a tiny bit more cleanup and DRYing out to come, but it's fine to kick that can down the road for a little bit longer if that's preferable and put that in a separate PR.

### DIFF
--- a/ui/packages/consul-acls/vendor/consul-acls/routes.js
+++ b/ui/packages/consul-acls/vendor/consul-acls/routes.js
@@ -1,16 +1,18 @@
-(function(data) {
-  const appNameJS = data.appName.split('-')
-    .map((item, i) => i ? `${item.substr(0, 1).toUpperCase()}${item.substr(1)}` : item)
-    .join('');
-  data[`${appNameJS}Routes`] = JSON.stringify({
-    dc: {
-      acls: {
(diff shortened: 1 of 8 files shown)


evrowe:
@johncowen I'd love to block out some time soon to go over:

- The application's general architecture
- The routing solution being used
- History of decisions made around critical aspects of the architecture

PR #11215, opened 2021-10-05
connect: Remove envoy_version from bootstrap template

### LINKED ISSUE #11134: Remove envoy_version field from bootstrap node metadata
We really don't need to be embedding the requested envoy version into the bootstrap yaml anymore. This was originally added before envoy had a reliable way to detect its version at runtime and we don't actually use this piece of metadata anymore.  Now it just shows up in every `command/connect/envoy` testdata blob taking up space.

This is safe to remove: https://github.com/hashicorp/consul/blob/main/command/connect/envoy/bootstrap_tpl.go#L149

Note: we should _not_ remove the related `consul connect envoy` version setting, as that is still useful if at some point we need to change how the bootstrap yaml itself is generated for different envoy versions.

eculver:
Fixes #11134. This just removes the `envoy_version` property from the bootstrap template. See the issue for more details on the background. In effect, this reduces the PR noise when adding and removing support for versions of Envoy.

### DIFF
--- a/command/connect/envoy/bootstrap_tpl.go
+++ b/command/connect/envoy/bootstrap_tpl.go
@@ -145,8 +145,7 @@ const bootstrapTemplate = `{
     "id": "{{ .ProxyID }}",
     "metadata": {
       "namespace": "{{if ne .Namespace ""}}{{ .Namespace }}{{else}}default{{end}}",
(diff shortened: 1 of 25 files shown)


rboyer (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/469056.

PR #11216, opened 2021-10-05
ui: Topology - Fix up Default Allow and Permissive Intentions notices

kaxcode:
### ✨ Description:
This PR updates the mock date and updates the conditionals used to show the Default Allow and Permissive Intentions notices.

### 📸Screenshots:
No screenshots.

### ⚡ Backend Changes:
No BE changes but we did verify with BE what the response should look like. 

### 🤡 Updates to mock-api:
Removed `DefaultAllow` and `WildcardIntention` from root `/v1/internal/ui/service-topology` response.

### 🧪 Testing:
Removed the test that covered the Default Allow notice, because of limitations with mocking out API header response. Time-blocked the effort and deleted it in the end.

Updated the test that covered the Permissive Intentions notice.

### 📝 **PR Tasks:**

- [ ]  Backport
- [ ]  Changelog
- [ ]  Demo Link


ui/packages/consul-ui/app/components/collapsible-notices/index.hbs
```diff
@@ -1,10 +1,19 @@
-<div class="collapsible-notices {{if this.collapsed 'collapsed'}}">
+<div
```
kaxcode, line 1: Calling to count how many notices appear on page.

ui/packages/consul-ui/app/components/collapsible-notices/index.js
```diff
@@ -1,3 +1,16 @@
...
+export default class CollapsibleNotices extends Component {
+  // =attributes
+  @tracked collapsible = false;
+
+  @action
+  countNotices() {
+    const allNotices = [
+      ...document.querySelectorAll('div.collapsible-notices > div.notices > .notice'),
+    ];
+    this.collapsible = allNotices.length > 2;
```
kaxcode, line 14: In upcoming Topology changes we'll be adding new more notices. These code changes are a more efficient solution to showing or not the collapsible button.

ui/packages/consul-ui/app/models/topology.js
```diff
@@ -33,14 +31,19 @@ export default class Topology extends Model {
     return undefinedDownstream;
   }
 
-  @computed('FilteredByACL', 'DefaultAllow', 'WildcardIntention', 'notDefinedIntention')
```
kaxcode, line 36: I removed this code because it was not efficient to show or not the collapsible banner. Soon it will be impossible to figure out all the "notice combos" that will exist.

kaxcode: Also, we did find out with BE folks that default allow and wildcard intentions can exist at the same time.

ui/packages/consul-ui/app/templates/dc/services/show/topology.hbs
```diff
@@ -136,7 +98,7 @@ as |nspace dc items topology|}}
     {{#if config.data}}
       <TopologyMetrics
         @nspace={{nspace}}
-        @dc={{dc.Name}}
+        @dc={{dc}}
```
kaxcode, line 101: I had to pass the whole `dc` object since I need the `dc.DefaultACLPolicy` in topology-metrics/index.js

ui/packages/consul-ui/mock-api/v1/internal/ui/service-topology/_
```diff
@@ -53,17 +53,11 @@ ${
     fake.seed(index);
 
 
-    // Randomly make permissive intentions
-    const defaultAllow = fake.random.boolean();
```
kaxcode, line 57: I thought about keeping this, but we verified with BE that default allow and wildcard intentions can co-exist.

johncowen:
Super nice we can do this now with that new ACL header.

I have a few question based on your description before I give this a proper look:

>⚡ Backend Changes:
> No BE changes but we did verify with BE what the response should look like.

> 🤡 Updates to mock-api:
> Removed DefaultAllow and WildcardIntention from root /v1/internal/ui/service-topology response.

If there are no backend changes, should we be updating the mocks? The mocks should be as close to the backend as possible. Is there a plan to make the same changes to the backend that are made here in the mocks?

> Removed the test that covered the Default Allow notice, because of limitations with mocking out API header response. Time-blocked the effort and deleted it in the end.

Could you explain what you mean by the limitations here, off the top of my head I can't think of what that might be?

Lemme know and I can take a look over this

kaxcode:
@johncowen 
> If there are no backend changes, should we be updating the mocks? The mocks should be as close to the backend as possible. Is there a plan to make the same changes to the backend that are made here in the mocks?

The mocks were wrong and did not match BE response. No changes were needed in the BE.


> Could you explain what you mean by the limitations here, off the top of my head I can't think of what that might be?

I could not set `X-Consul-Default-Acl-Policy = 'allow'` as a header response from the `v1/catalog/datacenters` endpoint.

johncowen:
> The mocks were wrong and did not match BE response. No changes are needed in the BE.

Ah gotcha, thanks for the explanation!

> I could not set X-Consul-Default-Acl-Policy = 'allow' as a header response from the v1/catalog/datacenters endpoint.

Ah, so not a limitation as such, more you're not sure how it's done? How much time did you time block this for? I can either walk you through it, show you where we do this already or pop something on the end of this myself depending on whether you can spend some more time on this or not. Lemme know how big your time block is and we can sort either way.

ui/packages/consul-ui/app/components/collapsible-notices/index.js
johncowen: Curious as to why this could be more efficient? Mostly in modern frameworks the idea is to declaratively configure what you want the view to be based on the state and then the framework efficiently creates that view for you - using querySelector[All] is generally shied away from. I would say that generally if you find yourself using DOM methods like this it might be worth thinking of another way to do it (there's always an outlier but generally the reason for that outlier is clear if there is)

ui/packages/consul-ui/app/templates/dc/services/show/topology.hbs
johncowen: Cool yeah good move, thanks for the explanation.

ui/packages/consul-ui/app/models/topology.js
```diff
@@ -33,14 +31,19 @@ export default class Topology extends Model {
...
+    const downstreamWildcard =
+      this.Downstreams.filter(item => !item.Intention.HasExact && item.Intention.Allowed).length !==
+      0;
 
-    return false;
+    const upstreamWildcard =
+      this.Upstreams.filter(item => !item.Intention.HasExact && item.Intention.Allowed).length !==
+      0;
+
+    return downstreamWildcard || upstreamWildcard;
```
johncowen, line 47: Just a note here, not sure if it matters or not. Did you know we have all this intention information available to us in the API call we make for the Services > Intentions tab (we also use the same call to do the intention updating on this Topology Tab). I would guess reusing that would save a bunch of backend work here, and if the user was using TProxy I don't think (not entirely sure) we'd have to make the extra topology endpoint API call at all. Anyway, no action to be made on FEs part here, but if you are talking to BE about this it might free them up a little.

johncowen (changes requested):
I made a few inline comments also, but don't feel like you have to address them before we get a chance to talk about avoiding deleting the test here (re: mocking out the ACL header). I'll wait til you get back to me on the other (not-inline) comment I made and we can go from there.

edit: Is this the one we had a list of user stories to test for? The one we were going to make a spreadsheet for to make sure we covered them all?

kaxcode:
> Ah, so not a limitation as such, more you're not sure how it's done? How much time did you time block this for? I can either walk you through it, show you where we do this already or pop something on the end of this myself depending on whether you can spend some more time on this or not. Lemme know how big your time block is and we can sort either way.

@johncowen I time-blocked a whole afternoon for this one assertion. I mostly looked through the `tests/steps/doubles` files and the the `tests/steps.js` file. I did not see any examples of setting a API response header. Feel free to share any insight on how it could be done at a later time.

kaxcode:
@johncowen I've put the solution you suggested to the test in this branch `ui/chore/default-allow-notices-test`. I left a `console.log` in there so you can see the cookies make no impact on the mocked API headers response.

johncowen:
Did you try both approaches I linked to? The first one works fine for me.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit daec73e76ca30d9d3e7e3a295041d4e4c7500cda onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/469698)

PR #11221, opened 2021-10-05
acl: extract a backend type for the ACLResolverBackend

dnephin:
Related to #11337 

This is a small step to isolate the functionality that is used for the `ACLResolver` from the large Client and Server structs. This follows the pattern and naming we use for all new RPC endpoints.

### DIFF
--- a/agent/consul/acl.go
+++ b/agent/consul/acl.go
@@ -133,11 +133,12 @@ func tokenSecretCacheID(token string) string {
 	return "token-secret:" + token
 }
 
-type ACLResolverDelegate interface {
+type ACLResolverBackend interface {
 	ACLDatacenter() string
 	ResolveIdentityFromToken(token string) (bool, structs.ACLIdentity, error)
 	ResolvePolicyFromID(policyID string) (bool, *structs.ACLPolicy, error)
 	ResolveRoleFromID(roleID string) (bool, *structs.ACLRole, error)
+	// TODO: separate methods for each RPC call (there are 4)
 	RPC(method string, args interface{}, reply interface{}) error
 	EnterpriseACLResolverDelegate
 }
@@ -160,8 +161,9 @@ type ACLResolverConfig struct {
 	// CacheConfig is a pass through configuration for ACL cache limits
 	CacheConfig *structs.ACLCachesConfig
 
-	// Delegate that implements some helper functionality that is server/client specific
-	Delegate ACLResolverDelegate
+	// Backend is used to retrieve data from the state store, or perform RPCs
+	// to fetch data from other Datacenters.
+	Backend ACLResolverBackend
 
 	// DisableDuration is the length of time to leave ACLs disabled when an RPC
 	// request to a server indicates that the ACL system is disabled. If set to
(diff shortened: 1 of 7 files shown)


agent/consul/replication.go
```diff
@@ -110,7 +110,7 @@ func NewReplicator(config *ReplicatorConfig) (*Replicator, error) {
 		return nil, fmt.Errorf("Cannot create the Replicator without a config")
 	}
 	if config.Delegate == nil {
-		return nil, fmt.Errorf("Cannot create the Replicator without a Delegate set in the config")
```
kisunji, line 113: Is this intentional?

agent/consul/acl.go
```diff
@@ -303,7 +305,7 @@ func NewACLResolver(config *ACLResolverConfig) (*ACLResolver, error) {
 	if config == nil {
 		return nil, fmt.Errorf("ACL Resolver must be initialized with a config")
 	}
-	if config.Delegate == nil {
+	if config.Backend == nil {
 		return nil, fmt.Errorf("ACL Resolver must be initialized with a valid delegate")
```
kisunji, line 309:
```suggestion
		return nil, fmt.Errorf("ACL Resolver must be initialized with a valid backend")
```

agent/consul/replication.go
dnephin: oops, nope this should not have been renamed. Good catch!

agent/consul/replication.go
```diff
@@ -110,7 +110,7 @@ func NewReplicator(config *ReplicatorConfig) (*Replicator, error) {
...
+		return nil, fmt.Errorf("Cannot create the Replicator without a Backend set in the config")
```
dnephin, line 113:
```suggestion
		return nil, fmt.Errorf("Cannot create the Replicator without a Delegate set in the config")
```

agent/consul/acl_server.go
```diff
@@ -114,6 +119,7 @@ func (s *Server) ACLDatacenter() string {
 }
 
 // ResolveIdentityFromToken retrieves a token's full identity given its secretID.
+// TODO: why does some code call this directly instead of using ACLResolver.ResolveTokenToIdentity ?
```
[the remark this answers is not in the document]
dnephin: I remove two of the calls in #12166. I've left `forwardDC` for now, as that one may be fine as-is.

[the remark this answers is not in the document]
freddygv: Seems like this TODO should be removed

agent/consul/acl_client.go
```diff
@@ -23,23 +23,28 @@ var clientACLCacheConfig *structs.ACLCachesConfig = &structs.ACLCachesConfig{
 	Roles: 128,
 }
 
-func (c *Client) ACLDatacenter() string {
-	// For resolution running on clients, servers within the current datacenter
+type clientACLResolverBackend struct {
+	// TODO: un-embed
```
freddygv, line 27: Could the client and server be un-embedded in this PR?

freddygv (approved):
Looks good, I'm looking forward to all of these simplifications! Only left two minor TODO-related questions.

agent/consul/acl_client.go
dnephin: Good question! I had another look to see if that was possible. It looks like it's not because the `Client` and `Server` are still being used directly for the `RPC()` method, and also for `EnterpriseACLResolverDelegate`.

I've got another TODO about splitting up the RPC into more specific calls (ex: `ReadToken`, `PolicyResolve`). I think by doing that we could have the `Server` implementation call the endpoint directly, without going through the in-memory codec that uses reflection.

But that's definitly a larger change, and one we'll need to think about more, as it could impact which metrics are emitted for these calls.  So I think that's probably better left for a future PR.

dnephin:
A couple of the integration tests failed due to the CircleCI outage, but the others passed, and this behaviour doesn't change with the envoy version, so going to merge.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/567519.

agent/consul/acl_server.go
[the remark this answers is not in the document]
dnephin: Hmm ya, this one is no longer correctly worded. We still have the forwardDC call to work through, so I'll get this TODO update to be more specific to that one remaining call in a follow up PR.

Issue #11152, opened 2021-09-27
consul_autopilot_healthy is reporting zero on the leader


arielazem:
Overview of the Issue
With Consul v1.10.1 (Revision db839f18b), the metric consul_autopilot_healthy is zero for all the servers including the leader, even when autopilot reports the cluster is healthy

![Screen Shot 2021-09-27 at 11 36 42 AM](https://user-images.githubusercontent.com/33434607/134940067-5d469c9f-5743-43e6-b42d-5383f932fee2.png)

Zero on all the servers

![Screen Shot 2021-09-27 at 11 43 00 AM](https://user-images.githubusercontent.com/33434607/134941262-0bab0c66-4514-4c13-a794-f06e9b1e84fa.png)

The autopilot health endpoint reports the cluster is healthy.

![Screen Shot 2021-09-27 at 11 38 44 AM](https://user-images.githubusercontent.com/33434607/134940403-aa9f5e6f-d780-452a-852e-5ba26bf86f40.png)

Reproduction Steps
Steps to reproduce this issue, eg:

Run Consul v1.10.1
Go to /v1/agent/metrics?format=prometheus
Operating system and Environment details
Linux VM in GCP

arielazem:
@mikemorris , #10730 is about all the followers reporting 0, while this other ticket #11152 is about leader reporting 0 as well

FFMMM:
hey @arielazem thanks for opening this issue with us!


I actually do believe that the underlying cause is a change in one of our dependencies (go-metrics). The behavior gets manifested as metrics that were previously `NaN` (whether in a certain stage of leadership or not) showing up with a 0 value.

As the team works on a fix, we'll track progress on #10730  for now so I will close this issue as a duplicate. But please feel free to re-open as needed.

---

For anyone  that's coming after, I added some more descriptive steps on how to repro the problem with the metrics here:

https://github.com/hashicorp/consul/issues/10730#issuecomment-929844513

Issue #10730, opened 2021-07-29
0 returned instead of NaN for `consul.autopilot.healthy|consul_autpilot_healthy` metric in Consul 1.10


dschaaff:
#### Overview of the Issue

The documented behavior is that `consul.autopilot.healthy` should return `NaN` on follower servers https://www.consul.io/docs/agent/telemetry#autopilot. After upgrading to 1.10 server's are now returning `0`. This has caused issues with our alerting.

#### Reproduction Steps

View `consul_autopilot_healthy` metric on a follower server. The value will be `0` instead of `NaN`.

### Consul info for both Client and Server

<details>
  <summary>Server info</summary>

```
agent:
	check_monitors = 0
	check_ttls = 0
	checks = 2
	services = 2
build:
	prerelease =
	revision = db839f18
	version = 1.10.1
consul:
	acl = enabled
	bootstrap = false
	known_datacenters = 1
	leader = false
	leader_addr = 10.20.208.112:8300
	server = true
raft:
	applied_index = 373096103
	commit_index = 373096103
	fsm_pending = 0
	last_contact = 4.978615ms
	last_log_index = 373096103
	last_log_term = 388662
	last_snapshot_index = 373087177
	last_snapshot_term = 388662
	latest_configuration = [{Suffrage:Voter ID:6cd485d0-1ccc-5049-8105-79abb943105e Address:10.20.209.189:8300} {Suffrage:Voter ID:b254668f-b57e-f763-5f3e-3f5ca38ecbc5 Address:10.20.208.112:8300} {Suffrage:Voter ID:e0e2704e-96c9-589d-c4eb-02f0ecf11bd2 Address:10.20.202.159:8300}]
	latest_configuration_index = 0
	num_peers = 2
	protocol_version = 3
	protocol_version_max = 3
	protocol_version_min = 0
	snapshot_version_max = 1
	snapshot_version_min = 0
	state = Follower
	term = 388662
runtime:
	arch = arm64
	cpu_count = 2
	goroutines = 1478
	max_procs = 2
	os = linux
	version = go1.16.6
serf_lan:
	coordinate_resets = 0
	encrypted = true
	event_queue = 0
	event_time = 205547
	failed = 1
	health_score = 0
	intent_queue = 0
	left = 1
	member_time = 647725
	members = 47
	query_queue = 0
	query_time = 1
serf_wan:
	coordinate_resets = 0
	encrypted = true
	event_queue = 0
	event_time = 1
	failed = 0
[...]

PR #11231, opened 2021-10-05
fix consul_autopilot_healthy metric emission

FFMMM:
### Overview

This PR fixes the values for the `consul_autopilot_healthy` emitted on:
* server startup
* leadership loss

The desired behavior, as documented on https://www.consul.io/docs/agent/telemetry#autopilot, is for that metric to be `NaN` in the states above. 

Due to a change in `go-metrics` this behavior changed and now the value reported is `0`. This can be disruptive for folks, especially if they use the metric to monitor or alarm on `0`, which marks an "unhealthy" cluster.

See FAQ below.

### Issue Related
* https://github.com/hashicorp/consul/issues/10730
* https://github.com/hashicorp/consul/issues/11152

### Notes:
* this is more of a stop gap solution that only tackles the `consul_autopilot_healthy` metric and not any of the remainder of https://github.com/hashicorp/consul/issues/10730#issuecomment-929844513

### PR Checklist
* [x] `go fmt`
* [x] `go mod`
* [ ] pr/changelog added
* [x] unit tests
* [x] integration tests - n/a

### FAQ

* **Q**: Why a "stop gap" solution first? 
  * **A**: I think this behavior deviation is quite disruptive for practitioners and I'd like to fix this asap. While we can come up with more elegant solutions, this should unblock folks
* **Q**: Where is the testing for the leadership loss / raft state change?
  * **A**: At present, our "unit" testing harness does not allow for more than one metrics sink (ref). While we work on adding a test for that, here's how I tested that scenario:
  
  TODO
  
* **Q**: Why add a `metrics_test.go` file and not have the `consul_autopilot_healthy` metrics test as part of 
`autopilot_test.go`? 
  * **A**: I think we ought to add more testing around metrics emissions/ behavior in out of consul. I.e., I'm hoping we can add the `1/0` checks for `consul_autopilot_healthy` but also tests for any other metric here.


agent/consul/autopilot.go
```diff
@@ -52,6 +52,21 @@ func (d *AutopilotDelegate) NotifyState(state *autopilot.State) {
...
+	} else {
+		/* todo -- re-evaluate metrics emitting NaN
+		From go-metrics v0.3.6, we have stopped resetting to NaN metrics that have
+		no observation in a collection window. A list of such metrics can be found here:
+		https://github.com/hashicorp/consul/issues/10730#issuecomment-929844513
+		To "fix" the metrics and have all appear as NaN and not 0 in prometheus
+		on d.NotifyState() here we could take a pass over all of these metrics and NaN them out.
+		However, the only documented NaN metric is consul_autopilot_healthy so we are
+		only "fixing" this for now.
+		*/
```
dnephin, line 64: This is helpful context for the PR review, but might not age well in the code. Might be better as a PR comment.

agent/testagent.go
```diff
@@ -81,6 +82,9 @@ type TestAgent struct {
 	// Agent is the embedded consul agent.
 	// It is valid after Start().
 	*Agent
+
+	// NoWaitForStartup, if false, will start the anti entropy sync and waitForUp() (see *waitForUp()).
+	NoWaitForStartup bool
```
dnephin, line 87: I don't think I understand the need for this. Why does `StartSync` and `waitForUp` cause a problem for this test?

agent/testagent.go
```diff
@@ -185,7 +189,12 @@ func (a *TestAgent) Start(t *testing.T) error {
 		}
 		result, err := config.Load(opts)
 		if result.RuntimeConfig != nil {
-			result.RuntimeConfig.Telemetry.Disable = true
+			// If "telemetry" settings are present, then we probably want
+			// to actually emit consul specific metrics
+			// otherwise, disable telemetry
+			if !strings.Contains(a.HCL, "telemetry") {
+				result.RuntimeConfig.Telemetry.Disable = true
```
dnephin, line 196: It's unfortunate our test helper makes this so difficult. I wonder if we can test this without using `TestAgent`. In general I find things like `TestAgent` cause problems. Ideally we would make the real production interfaces nice enough so that we don't require complex test helpers to test them.

We have a couple tests that do not use `TestAgent`. I can help investigate that option.

agent/metrics_test.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+// TestMetrics_ConsulAutopilotHealthy_Prometheus adds testing around
+// the published autopilot metrics on https://www.consul.io/docs/agent/telemetry#autopilot
+func TestMetrics_ConsulAutopilotHealthy_Prometheus(t *testing.T) {
```
dnephin, line 12: I have found that it is generally easier to identify the unit under test when the test name includes the struct and method name. We haven't really documented this, but the pattern I follow is:

```
Test[<Struct>_]<Method>[_<Case>]
```


```suggestion
func TestHTTPHandlers_AgentMetrics_ConsulAutopilotHealthy_Prometheus(t *testing.T) {
```

dnephin (commented):
Nice! I think this is a good fix for Servers, but Clients will still emit a 0 for this metric, which I think will still be a problem.  To fix that, I think we need to make sure that `AutopilotGauges` are only added in `getPrometheusDefs` when in server mode. That will likely require some additional plumbing.


Left some comments about the tests. I think it would be good to test, but I'm concerned we're adding yet more complexity to what is already a very complex test helper. I'm hoping we can find a better way to test this.

agent/testagent.go
FFMMM: So the unit test that's here test the first scenario:

>* server startup

By the time `waitForUp` finishes:

> // waitForUp waits for leader election, or waits for the agent HTTP
> // endpoint to start responding, depending on the agent config.

that state will have passed already. So the metric will emit `1` since it's a healthy cluster.

---

I could probably leave `StartSync()` outside of the if block but I'm not sure it'd be useful for folks since I don't see a scenario for an agent in that state.

---

Does this make some sense ? 🤔

markblackman:
Can this get backported to 1.9 as well?

PR #11233, opened 2021-10-05
acl: move ACLCaches from structs to acl/aclauthz package

dnephin:
Related to #11337 

This `ACLCache` type is used exclusively by the `ACLResolver`, it does not belong in the structs package. This PR moves the caches and the cache config into a new `acl/aclauthz` package. In the future the full `ACLResolver` can be moved into this package, which may remove the need to export most of the `ACLCache` type and methods, and provide a better interface for resolving ACL tokens.

The first commit moves some functions around to prepare for extracting into a new package. The second commit moves the two files into a new package.


acl/aclauthz/acl_cache.go
```diff
@@ -264,3 +266,54 @@ func (c *ACLCaches) Purge() {
 		}
 	}
 }
+
+// TODO: unexport once this is moved into the same package as ACLResolver
+func NewPolicyAuthorizerWithCache(policies structs.ACLPolicies, cache *ACLCaches, entConf *acl.Config) (acl.Authorizer, error) {
```
dnephin, line 271: This was `ACLPolicies.Compile` , the only change here is to the function signature and the call to `resolveWithCache` since that function signature also changed.

acl/aclauthz/acl_cache.go
```diff
@@ -264,3 +266,54 @@ func (c *ACLCaches) Purge() {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to construct ACL Authorizer: %v", err)
+	}
+
+	// Update the cache
+	cache.PutAuthorizer(cacheKey, authorizer)
+	return authorizer, nil
+}
+
+func resolveWithCache(policies structs.ACLPolicies, cache *ACLCaches, entConf *acl.Config) ([]*acl.Policy, error) {
```
dnephin, line 296: This was previously `ACLPolicies.resolveWithCache`, the only change is the function signature

acl/aclauthz/acl_cache_test.go
```diff
@@ -0,0 +1,304 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/hashicorp/consul/acl"
+	"github.com/hashicorp/consul/agent/structs"
+)
+
+func TestStructs_ACLCaches(t *testing.T) {
```
dnephin: This test was moved, but should be otherwise unchanged.

acl/aclauthz/acl_cache_test.go
```diff
@@ -0,0 +1,302 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/hashicorp/consul/acl"
+	"github.com/hashicorp/consul/agent/structs"
+)
+
+func TestNewACLCaches(t *testing.T) {
```
dnephin, line 13: This was previously `TestStructs_ACLCaches`, only the test name has changed

acl/aclauthz/acl_cache_test.go
```diff
@@ -0,0 +1,302 @@
...
+		// we reset the Authorizer in the cache so now everything should be defaulted
+		require.Equal(t, acl.Deny, authz.NodeRead("foo", nil))
+		require.Equal(t, acl.Deny, authz.AgentRead("foo", nil))
+		require.Equal(t, acl.Deny, authz.KeyRead("foo", nil))
+		require.Equal(t, acl.Deny, authz.ServiceRead("foo", nil))
+		require.Equal(t, acl.Deny, authz.ACLRead(nil))
+	})
+}
+
+func TestResolveWithCache(t *testing.T) {
```
dnephin, line 215: Previously `TestStructs_ACLPolicies_resolveWithCache`, updated the test for the new function signature, but otherwise unchanged

PR #11235, opened 2021-10-06
fix social share image

pbortnick:
Sharing consul.io was still displaying the old share image. This PR should fix that and update it to the newer one.


website/pages/_app.js
```diff
@@ -29,7 +29,7 @@ export default function App({ Component, pageProps }) {
         title="Consul by HashiCorp"
         siteName="Consul by HashiCorp"
         description="Consul is a service networking solution to automate network configurations, discover services, and enable secure connectivity across any cloud or runtime."
-        image="https://www.consul.io/img/og-image.png"
+        image="/img/og-image.png"
```
mwickett, line 32: I believe `meta image` tags have to have an absolute URL to resolve. The image isn't showing for me when I check the preview link in https://metatags.io/.

PR #11237, opened 2021-10-06
ui: Fix up blocking reconciliation for multiple models

johncowen:
Follow up to https://github.com/hashicorp/consul/pull/11017

https://github.com/hashicorp/consul/pull/11017#discussion_r707463018

> In the future, this should all be moved to each individual repository now, which will mean we can finally get rid of this service.

This PR moves reconciliation to 'each individual repository'. I stopped short of getting rid of the service, but its so small now we pretty much don't need it. I'd rather wait until I look at the equivalent `DataSink` service and see if we can get rid of both equivalent services together (this also currently dependant on work soon to be merged)

Reconciliation of models (basically doing the extra work to clean up the ember-data store and bring our frontend 'truth' into line with the actual backend truth) when blocking/long-polling on different views/filters of data is slightly more complicated due to figuring out what should be cleaned up and what should be left in the store. This is especially apparent for KVs.

I built in a such a way to hopefully make sure it will all make sense for the future. I also checked that this all worked nicely with all our models, even KV which has never supported blocking queries. I left all that work in so that if we want to enable blocking queries/live updates for KV it now just involves deleting a couple of lines of code.

There is a tonne of old stuff that we can clean up here now (our 'fake headers' that we pass around) and I've added that to my list of thing for a 'Big Cleanup PR' that will remove lots of code that we no longer require.

### DIFF
--- a/.changelog/11237.txt
+++ b/.changelog/11237.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Ensure all types of data get reconciled with the backend data
+```


kaxcode (approved):
👍

PR #11240, opened 2021-10-06
ui: Default allow notices test

kaxcode:
Small PR to have a test for Default Allow notices.


kaxcode:
@johncowen At this time, I'm not making any further changes or research to make this test pass. If you don't feel it can be merged because of the mock-data then I will close it.
 
Please let me know. I would like to unblock #11216.

Issue #11207, opened 2021-10-02
AuthorizeServerConn failed certificate validation


Lord-Y:
Hi team,

I running a consul server cluster of 3 nodes VM.
My consul version was 1.9.8, and I deciced to upgrade my cluster to 1.10.3 version and in here is what I saw in the logs:

```
from=10.0.104.13:57854 operation="raft RPC"
2021-10-02T13:53:24.695Z [WARN]  agent.server.raft: Election timeout reached, restarting election
2021-10-02T13:53:24.696Z [INFO]  agent.server.raft: entering candidate state: node="Node at 10.0.104.12:8300 [Candidate]" term=168
2021-10-02T13:53:24.701Z [DEBUG] agent.server.raft: votes: needed=2
2021-10-02T13:53:24.701Z [DEBUG] agent.server.raft: vote granted: from=2afb25d5-8234-e9fc-dcef-babf75741651 term=168 tally=1
2021-10-02T13:53:24.707Z [ERROR] agent.server.raft: failed to make requestVote RPC: target="{Voter 24a6c35e-9bab-7bae-9a31-ae94400ac1ef 10.0.104.13:8300}" error=EOF
2021-10-02T13:53:24.708Z [ERROR] agent.server.raft: failed to make requestVote RPC: target="{Voter 3840bd0b-0980-8c1e-47c7-73bfef3bb5b9 10.0.104.10:8300}" error=EOF
2021-10-02T13:53:24.803Z [WARN]  agent.server.rpc: AuthorizeServerConn failed certificate validation for certificate with a SAN.DNSName of server.euw1.consul: %!w(<nil>): from=10.0.104.10:48992 operation="raft RPC"
2021-10-02T13:53:24.913Z [WARN]  agent.server.rpc: AuthorizeServerConn failed certificate validation for certificate with a SAN.DNSName of server.euw1.consul: %!w(<nil>): from=10.0.104.13:57858 operation="raft RPC"
2021-10-02T13:53:26.045Z [WARN]  agent.server.rpc: AuthorizeServerConn failed certificate validation for certificate with a SAN.DNSName of server.euw1.consul: %!w(<nil>): from=10.0.104.13:57862 operation="raft RPC"
2021-10-02T13:53:26.083Z [WARN]  agent.server.rpc: AuthorizeServerConn failed certificate validation for certificate with a SAN.DNSName of server.euw1.consul: %!w(<nil>): from=10.0.104.10:48998 operation="raft RPC"
```

[...]

PR #11257, opened 2021-10-08
config: default verify_incoming_rpc=true and deprecate verify_incoming

dnephin:
Related to #11207
Branched from #11255

The `verify_incoming` config setting has been a problem because it groups together two very different settings. `verify_incoming_https` can be false and it doesn't present any serious security risk to the cluster. `verify_incoming_rpc` however effectively disables any protection provided by enabling tls.  Grouping these two things together makes it too easy to introduce an insecure setup.

This PR makes two changes to address this problem:
1. Deprecate `verify_incoming`. For now it continues to exist as an alias for `verify_incoming_https` and prints a warning when it is used. We may remove it in the future.
2. When either `ca_file` or `ca_path` are set, `verify_incoming_rpc` now defaults to `true`. This is a breaking change that is called out in the release notes. This changes makes it easier for users to get a secure setup by removing the need to set another configuration field to true.

TODO:
* [ ] what to do about verify_outgoing=false (See comments below)
* [ ] audit our learn guides to make sure we no longer reference `verify_incoming`
* [ ] audit our consul.io docs to make sure we never reference `verify_incoming`

### DIFF
--- a/.changelog/11257.txt
+++ b/.changelog/11257.txt
@@ -1,0 +1,7 @@
+```release-note:breaking-change
+config: `verify_incoming_rpc` now defaults to true if either `ca_file` or `ca_path` are set. Previously it defaulted to false in all cases.
+```
+
+```release-note:deprecation
+config: `verify_incoming` is deprecated and may be removed in a future version. Use `verify_incoming_https` and `verify_incoming_rpc` instead.
+```
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -3915,7 +3915,7 @@ func TestAgent_ReloadConfigTLSConfigFailure(t *testing.T) {
 	require.Error(t, a.reloadConfigInternal(c))
 	tlsConf, err := tlsConf.GetConfigForClient(nil)
 	require.NoError(t, err)
-	require.Equal(t, tls.NoClientCert, tlsConf.ClientAuth)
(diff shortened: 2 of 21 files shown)


.changelog/11257.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:breaking-change
+config: `verify_incoming_rpc` now defaults to true if either `ca_file` or `ca_path` are set. Previously it defaulted to false in all cases.
+```
+
+```release-note:deprecation
+config: `verify_incoming` is deprecated and may be removed in a future version. Use `verify_incoming_https` and `verify_incoming_rpc` instead.
```
mkeeler, line 6: I think this would break the ability for non TLS'ed clusters to enable TLS without downtime. See [this learn guide](https://learn.hashicorp.com/tutorials/consul/tls-encryption-secure-existing-datacenter#configure-the-servers) on how the online migration is performed. Basically it involves setting up the certs with all verification off and then once client agents etc. are all initiating TLS connections, then the verification settings can be turned on.

Changing the default when ca material is provided seems like a good idea, but removal of the ability to disable that verification while still using certs is going to make securing an existing system require downtime so I think we should avoid it.

mkeeler (changes requested):

jmurret:

PR #11260, opened 2021-10-08
Confirmation dropdown on delete action

### LINKED ISSUE #11228: Confirmation dropdown for delete actions
### Overview of the issue

Clicking a delete button should open a confirmation dialog/dropdown. Some of our pages don't follow this interaction.

**Desired styling**

CleanShot 2021-10-05 at 17 04 18

Use the correct interaction on the corresponding listviews as guidance for wording.

**Change applies to:**
- K/V edit page
- Intention edit page
- Token edit page
- Policy edit page
- Role edit page

vinayak03:
Continuing from https://github.com/hashicorp/consul/pull/11252 since I had to change the branch.

Adds the confirmation dropdown for delete action
Updates the messages as per Confirmation dropdown for delete actions #11228

Closes #11228

Earlier we were using the `type-delete` class for css as well test actions, Incase you didn't want css effect however the test would break since both use the same thing. To separate the concerns I have introduced the data-test-id `data-test-confirm-delete` for reference in test.

Key/value delete
image

Intention delete
image

Token delete
image

Role delete
image

When I am deleting a policy, on clicking delete button just disappears. I see this issue on the main branch as well.

### DIFF
--- a/.changelog/11252.txt
+++ b/.changelog/11252.txt
@@ -1,0 +1,3 @@
+```release-note: improvement
+ui: Confirmation dropdown for delete actions in draft mode for kv, intention, token, role, policies
+```
--- a/ui/packages/consul-ui/app/components/confirmation-dialog/index.hbs
+++ b/ui/packages/consul-ui/app/components/confirmation-dialog/index.hbs
@@ -1,12 +1,13 @@
-<div class={{concat "with-confirmation" (if confirming " confirming" "") (if permanent " permanent" "")}} ...attributes>
+<div class={{concat "with-confirmation" (if confirming " confirming" "")}} ...attributes>
 {{yield}}
-<YieldSlot @name="action" @params={{block-params (action "confirm") (action "cancel")}}>
-{{#if (or permanent (not confirming))}}
+<YieldSlot @name="dialog" @params={{block-params (action "execute") (action "cancel") message actionName}}>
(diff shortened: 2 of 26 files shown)


johncowen:
> One thing I am noticing here is that once we click on the delete button confirmation dialog opens below so user has to scroll down below to take the action.

Yes definitely good call 👍 There's a class we add on the 'in listing' ones that might help with that (possibly `class="above"`) that makes the dialog popup instead of down. Again if you (and @jnwright ) wanted to do that here that that would be 💯 also, but no biggie if not.

I'd like to move all these popover type things to just use our PopperJS modifier eventually which automatically does all that for us (much bigger piece of work that tho, so probably not just yet)

> So may be for better user experience could this confirmation be a pop-up (model)?

Definitely a q for @jnwright that one, but good feedback!

vinayak03:
Hi @johncowen ,

I worked on most of the feedback items however I could not get it to show as a pop-up.

Role,
image

Token,
image

Key/Value,
image

Intention,
image

If the changes looks ok to you please merge,( I think the PR got a lot bigger to review at this stage)

johncowen:
Second everything @jkirschner-hashicorp said here @vinayak03 thanks for the work, it was nice working with you. Please feel free to pick something else up if that's something you'd be interested in.

Thanks both 🙏

Issue #11264, opened 2021-10-08
Admin UI: Cannot create intentions with service_prefix policy


GordonMcKinney:
#### Overview of the Issue

The following policy allows the CLI to create intentions but not via the admin UI. Version Consul 1.10.3

#### Reproduction Steps

I have a docker-compose rig to demonstrate this behavior. Please DM me for the details.

Using the following policy:

```
{
  "service_prefix": {
    "cl-pepr-": {
      "intentions": "write",
      "policy": "write"
    }
  },
  "node_prefix": {
    "": {
      "policy": "read"
    }
  },
  "service_prefix": {
    "": {
      "policy": "read"
    }
  }
}
```

Successful CLI:

```
$ consul -version
Consul v1.10.3
Revision c976ffd2d
$ export CONSUL_HTTP_TOKEN=<redacted>
$ consul intention create cl-apigw-web cl-pepr-payments
Created: cl-apigw-web => cl-pepr-payments (allow)
```

Observe the same actions via the Admin UI with the same token:

<img width="631" alt="Screen Shot 2021-10-08 at 8 56 12 AM" src="https://user-images.githubusercontent.com/4955046/136620978-7985bc47-f9bc-464e-a944-f8d48afcced9.png">

<img width="825" alt="Screen Shot 2021-10-08 at 8 56 52 AM" src="https://user-images.githubusercontent.com/4955046/136620986-3af3777f-7493-40c6-b041-cb3ba4722fd4.png">

### Consul info for both Client and Server

<details>
  <summary>Client info</summary>

```
agent:
	check_monitors = 0
	check_ttls = 0
	checks = 2
	services = 2
build:
	prerelease =
	revision = c976ffd2
	version = 1.10.3
consul:
	acl = enabled
	known_servers = 1
	server = false
runtime:
	arch = amd64
	cpu_count = 4
	goroutines = 77
	max_procs = 4
	os = linux
	version = go1.16.7
serf_lan:
	coordinate_resets = 0
	encrypted = false
	event_queue = 0
	event_time = 3
	failed = 0
	health_score = 0
	intent_queue = 0
	left = 0
	member_time = 2
	members = 3
	query_queue = 0
	query_time = 1
```

</details>

<details>
  <summary>Server info</summary>

```
agent:
	check_monitors = 0
	check_ttls = 0
	checks = 0
	services = 0
build:
	prerelease =
	revision = c976ffd2
	version = 1.10.3
consul:
	acl = enabled
	bootstrap = false
[...]

PR #11270, opened 2021-10-11
ui: Ensure we check intention service prefix permissions for per service

johncowen:
Previously, when showing some action buttons for 'per service intentions' we used a global 'can I do something with any intention' permission to decide whether to show a certain button or not. If a user has a token that does not have 'global' intention permissions, but does have intention permissions on one or more specific services (for example via service / service_prefix), this meant that we did not show them certain buttons required to create/edit the intentions for this specific service.

This PR adds that extra permissions check so we now check the intentions permissions per service instead of using the 'global' "can I edit intentions" question/request.

Fixes https://github.com/hashicorp/consul/issues/11264

One thing that maybe deserves a mention here is that mostly the UI is a thin client, so if Consul has permissions set to deny intention creation, there is no way to create that intention in Consul, whether there is a button in the UI or not. Previous to 1.10 the UI was 100% a thin client, so we would always show buttons for things no matter what, kinda like you can type any command you want into the CLI 'UI' to try and do things with Consul, and Consul will always deny a command if you are not allowed to do so. Here: CLI command = UI Button

Post 1.9 we started diverging from the idea of the UI as a thin client and began hiding buttons (commands) for things that you are not supposed to do. Whilst this can be a UX improvement, we will always be balancing a tightrope / risking not showing a button for something the user would want to do (or vice versa: showing a button for something the user does not want to do). Mostly these 'risks' are bugs, like this one, introduced that unfortunately block workflows (imagine not being able to actually type a command into the CLI because the UI for the CLI won't let you), but there are times when its not completely black and white and more open to opinion (@jkirschner-hashicorp this is very relevant to the convo we had at the end of last week, all food for thought)

P.S. I added a no-changelog label here again. It does have a changelog, just our changelog checker doesn't like it when you add a changelog on a non-main branch (still on my list to sort this out one way or another)

### DIFF
--- a/.changelog/11270.txt
+++ b/.changelog/11270.txt
@@ -1,0 +1,4 @@
+```release-note:bug
+ui: Ensure we check intention permissions for specific services when deciding
+whether to show action buttons for per service intention actions
+```


johncowen:
This is awesome @GordonMcKinney thankyou! It almost sounds like we might be left with a Safari issue here, which I can look into from my end. I'll give you a ping here once I've done that if you're ok/want to give it another run over.

Yesterday I caught up with one of our backend folks to chat through this, and I'll be catching up with larger team today and raising this. So I may have more info later on.

Thanks again!

johncowen:
Hey @GordonMcKinney 

I've had a little play with this myself and I don't think its just Safari I'm seeing rendering strangeness in both Chrome and Safari:

Screenshot 2021-10-13 at 15 16 43

Screenshot 2021-10-13 at 15 15 55

But I'm not entirely sure if these rendering issues are related to the permissions issues that were the original root of the issue we are working on.

I think the best thing to do is if I massage the tests into shape here, and then get this merged, which should give us/you an improvement. I can check in with the team separately and get some more guidance on the permissions things which I think are still an issue (see below) and then we can look at the rendering issue separately - it may or may not be perms related.

One more favour I'd like to ask of you if possible is, with the above setup that you have, when you go to the top level intentions page by clicking 'Intentions' in the left hand main menu (whilst using your intention:write token). I'm guessing you don't see a [Create] button on the top right (I think we actually should show it or _something_ at least here). FYI: When I try to reproduce this using your directions I _don't_ see this [Create] button, to I'm guessing you don't either?

Let me know what your thoughts are and possibly go ahead and work on getting this part merged.

johncowen:
Hey @GordonMcKinney 

Sorry for the little delay here, long story short: I 'think' we are there with this one 🎉 

> But I'm not entirely sure if these rendering issues are related to the permissions issues that were the original root of the issue we are working on.

Turns out it was related to the permissions stuff I'd added. Sorting the tests out sorted those issues (the issues weren't producing any errors in my web console which was why I was somewhat confused, I'll look into that separately as a DX thing)

So thats ^ the stuff thats most important, but I'd like to touch on some of our other discussion also if thats ok/you have the time.

> I hesitate to confirm this based on the language. Our security posture minimizes reconnoissance. The "global" intentions page (clicking the left bar "intentions") returns no results with our current token policy.

Yup, gotcha completely, makes perfect sense.

> That's fine for our use-case. Missing the "create" button in that circumstance is consequence we're happy with.

I touched on it a little in that PR description here that I've "unTBD'ed" to leave for other folks to read. But just to try reassure you slightly incase, our UI is mainly a thin client. Consul itself does the important stuff surrounding enforcing permissions. Even though now the UI is slightly less of a thin client, when it does go wrong, the only consequence would be understandable user frustration, not un-enforcement of permissions. A good way to think about this is, anyone could just work around any 'enforcement' in the UI by hitting the HTTP API instead using something like curl.

> You'd prefer folks to drill into the Service, then to the Intention tab for the Service that they have permission to add an Intention for and then create it from there?
> > Ideally the "X" on the topology page.

Do you ever have a need to deny a connection via the Topology page?

> I'm nervous about the UI bugs. My UX had the UI lock-up and require refresh after the topology page "Connection Refused" box appeared.

This sounds like a completely different issue. It might be one I've heard being discussed elsewhere, so I'm guessing it could be related. Do you have lots of services on your topology tab?

GordonMcKinney:
NOTE: I had to change the build to use: `ARG NODEJS_VERSION=14.17.6-r0` as Alpine wouldn't continue otherwise.

johncowen:
Hey @GordonMcKinney 

> NOTE: I had to change the build to use: ARG NODEJS_VERSION=14.17.6-r0 as Alpine wouldn't continue otherwise.

This shouldn't effect what we are doing here. But if you could keep the change locally so you can continue to help us debug that'd be great. Alternatively I can walk you through how to run the UI in dev mode against a real Consul cluster, happy to do that if its easier. (actually 🤔 it happens very rarely but we do have occasions where a difference between our dev environment and the real thing produces unexpected behaviour, which might explain why I can't repro - see below)

FYI: There is ongoing work elsewhere (in 10996) that touches this, but that work is not just about the UI and we are still discussing this internally. In fact I have another un-PRed branch from back in September related to that. 

> This branch, commit 5c7e4e2.

Is this the only branch you see the problem on?

My next step is to understand if the problem you are seeing is as a result of the changes here or not. If it is, I can't merge this. If it isn't a result of this change, I'm happy to merge.

If I know you can reproduce this on where this PR branches off from (or anywhere fairly recent before that point) then I've ruled out the changes here are doing something I haven't noticed.

Our other options are to do something to enable me to reproduce this myself. As of right now I can't. I've followed your config from https://github.com/hashicorp/consul/issues/11264 (same policy, and then in Topology - one LHS service with the central service). I clicked around and around on different tabs to try to make it hang. I also left it on its own for a bit without interacting and I couldn't make it hang. It sounds like you can make it break pretty consistently?

In relation to my "see below" comment above, until you are around I'm also going to try and repro this outside of our dev environment and see if that gives me any further clues.

If I can confirm that the problem only occurs on this branch (5c7e4e2) and no others then we can figure out how we can get a repro together.

Thanks again!

GordonMcKinney:
>  But if you could keep the change locally 

No problem. Making an OSS contribution requires effort within the framework set by my Company.

> Is this the only branch you see the problem on? If I know you can reproduce this on where this PR branches off from (or anywhere fairly recent before that point) then I've ruled out the changes here are doing something I haven't noticed.

It's the only branch I've tested. I'm happy to test another branch or commit, LMK what I should checkout.

I'm happy to host a screen share if you have availability.

GordonMcKinney:
Testing...

GordonMcKinney:
I believe I missed a `compose build` yesterday. Here are the new results (`/bin/consul` timestamp verified on each run):


## Branch: release/1.10.x
* Topology and intentions tabs are readonly with my service _admin_ token. UI topology exhibits no bugs. (Safari)


## Branch: ui/bugfix/per-service-intention-perms

Chrome:
* Topology screen works correctly with my service _admin_ token ✅
* Intentions screen delete works correctly ✅
* Intentions screen create works correctly ✅
* _testing halted, switched to safari_

Safari (New Private Window):
* Topology screen works correctly with my service _admin_ token ✅
* Intentions screen delete works correctly with my service _admin_ token ✅
* Intentions screen create works correctly with my service _admin_ token ✅
* Topology screen allows a different _admin_ token to click X but results in a 403 (screenshot) ✅
* Intentions screen create reports a 403 when using a different _admin_ token ✅
* Topology screen allows a different _service_ token (no `intention:write`) to click X but results in a 403 ⚠️
* Intentions screen create reports a 403 when using a different  _service_ token (no `intention:write`) ⚠️


I think the current behavior is great! You allow admins to approve their own intentions and report 403 when they're not allowed. I feel this is a better UX than silently disabling the "X" or "Create" buttons based upon context (leading to confusion).

I do feel that an ACL without `intention:write` shouldn't be allowed to interact with "X" and shouldn't see "Create"


----

Screen Shot 2021-10-19 at 9 04 03 AM

----

Screen Shot 2021-10-19 at 8 57 44 AM

----

## Service Token (no intention:write)

```
{
  "agent_prefix": {
    "cl-pepr-": {
      "policy": "write"
    }
  },
  "key_prefix": {
    "cl-pepr-": {
      "policy": "write"
    }
  },
  "node_prefix": {
    "": {
      "policy": "read"
    },
    "cl-pepr-": {
      "policy": "write"
    }
  },
  "service_prefix": {
    "": {
      "policy": "read"
    },
    "cl-pepr-": {
      "policy": "write"
    }
  },
  "session": {
    "cl-pepr-": {
      "policy": "write"
    }
  }
}
```

Issue #10224, opened 2021-05-11
Answers to SRV queries on alt_domain should also use the alt_domain


joel0:
#### Feature Description

When `alt_domain` is set in the config, I would expect SRV lookups on the alt domain to respond with the alt domain.  For example, if `alt_domain = "consul.example.com"` and I do `dig SRV myservice.service.sea.consul.example.com`, I'd expect
```
;; ANSWER SECTION:
myservice.service.sea.consul.example.com. 0 IN SRV 1 1 1337 server42.node.sea.consul.example.com.
```
but I get
```
;; ANSWER SECTION:
myservice.service.sea.consul.example.com. 0 IN SRV 1 1 1337 server42.node.sea.consul.
```

The `consul` TLD being unresolvable by the client is why I was using the alt domain.

#### Use Case

The use case we have is that we've put the alt domain into a global DNS resolver, but can't put the `consul` TLD in it.  Some of our services directly query Consul DNS with the `consul` TLD, while other remote services use the alt domain.  The remote services are unable to resolve the IP address because they try to query the `consul` TLD, despite the original query being for the alt domain.  An intermediate DNS resolver strips the additional section, so we can't configure the client to use the IP presented there.

In our specific case, it looks like we can simply set `domain = "consul.example.com"; alt_domain = "consul"` to resolve the problem, but I figured I'd submit a feature request anyway because it seems to make sense to not respond with `domain` to `alt_domain` queries.

jkirschner-hashicorp:
Hello Consul community members,

We would welcome a PR contributed by the community for this!

If you're interested, please comment here so anyone interested can stay informed.

The approach should ensure the following:
- The response to a SRV query should use the same domain as the request
  - If the request uses `alt_domain`, the response _should_ use `alt_domain`, but _currently_ uses `domain` instead
- Include unit tests to demonstrate that:
  - SRV queries using `domain` have a response using `domain`
  - SRV queries using `alt_domain` have a response using `alt_domain`
- Bonus points if you also check the behavior for supported DNS query types other than SRV and fix those if needed
  - Note: PR #10431 was recently submitted to fix this behavior for SOA records. That might be a useful reference for someone starting on this task.

JqckB:
For SRV record, we saw the same issue, but there is also an question about PTR : PTR is responding with .consul response, but we are using primarily our alt-domain as our domain.

For some service that use PTR record to validate an IP to a domain (as postgresql), it cause an issue.

[...]

PR #11274, opened 2021-10-11
Answers to SRV queries on alt_domain returned with alt_domain. 

kbabuadze:
Apart for the ` ";; ANSWER SECTION:" ` changes have also been made to the `;; ADDITIONAL SECTION:` where default domain was used. 
Fixing:  https://github.com/hashicorp/consul/issues/10224


kbabuadze:
@jkirschner-hashicorp Thanks for mentioning it. 
 I think I found one more place, where it might need a fix. 
 https://github.com/hashicorp/consul/blob/a5c292c000ace8b9bd6a0231f86039ba629753cd/agent/dns.go#L582-L585
 I'll check if there are more.

PR #11298, opened 2021-10-13
ui: Move the Role remove dialog to use InformedAction

johncowen:
Following on from https://github.com/hashicorp/consul/pull/11297 this fix is a lot better 😄 has main as the base branch, and will be backported to 1.9 (but not 1.8).

Thanks to @vinayak03's work in #11260 I spotted a little visual issue in the Token Form > Role Listing confirmation dialog.

Before:

Screenshot 2021-10-13 at 13 55 10

This dialog still used our super old confirmation dialog instead of our InformedAction component, so the fix here is just upgrading the actions here to use the InformedAction component.

After:

Screenshot 2021-10-13 at 13 54 30

Related https://github.com/hashicorp/consul/pull/11297

### DIFF
--- a/.changelog/11298.txt
+++ b/.changelog/11298.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fixed styling of Role remove dialog on the Token edit page
+```


hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/472458.

PR #11314, opened 2021-10-14
Ensure partition is handled by auto-encrypt

freddygv:
* The CA.Sign RPC endpoint now considers the partition of the requesting
agent when checking whether the request should be authorized.
* The CAManager now encodes the partition when signing certificates for
agents.
* Partitions will now be encoded in auto-config CSRs


hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/472681.

Issue #11330, opened 2021-10-15
Disabling Maintenance mode does not trigger an event


jakubgs:
#### Overview of the Issue

I was looking at disabling alerts in `consul-alerts` when host is in maintenance mode: https://github.com/AcalephStorage/consul-alerts/issues/213

And I found out that while `consul-alerts` does create an entry in the KV store for `_node_maintenance` service check when maintenance mode is disabled, it does NOT update that entry once maintenance mode is disabled:

![image](https://user-images.githubusercontent.com/2212681/137499477-b417c01c-f758-4ab6-b0c0-40f66304c155.png)

So I tried running `consul watch` myself same [as the `consul-alerts` service does](https://github.com/AcalephStorage/consul-alerts/blob/v0.6.0/watchers.go#L15-L25) and I found out that disabling maintenance mode does not trigger an event:
```
 > consul watch -type checks cat | jq '.[] | select(.CheckID == "_node_maintenance")'
{
  "Node": "node-01.example.org",
  "CheckID": "_node_maintenance",
  "Name": "Node Maintenance Mode",
  "Status": "critical",
  "Notes": "TEST",
  "Output": "",
  "ServiceID": "",
  "ServiceName": "",
  "ServiceTags": [],
  "Type": "maintenance",
  "Definition": {
    "Interval": "0s",
    "Timeout": "0s",
    "DeregisterCriticalServiceAfter": "0s",
    "HTTP": "",
    "Header": null,
    "Method": "",
    "Body": "",
    "TLSServerName": "",
    "TLSSkipVerify": false,
    "TCP": ""
  },
  "CreateIndex": 67303625,
  "ModifyIndex": 67303625
}
```
I receive one event with `Status` set to `critical`, but no recovery to `passing`, since the `_node_maintenance` check is removed.

For this reason `consul-alerts` never updates its stage of `_node_maintenance` check.

#### Reproduction Steps

1. Use version `1.10.1`
1. Run `consul watch -type checks cat`
1. See that no even is triggered for disabling of maintenance mode

### Operating system and Environment details

Ubuntu 20.04.3, `x86_64`

jakubgs:
Though now that I look at it adding a test to verify the status was changed to `passing` before removal will be a bit more tricky:
https://github.com/hashicorp/consul/blob/cc2abb79baa5989821ede371a19464fa691ec000/agent/agent_test.go#L3358-L3362
[...]

PR #11332, opened 2021-10-15
update maintenance check to passing before removing

jakubgs:
Without this update service using `consul watch` to monitor check changes will never see maintenance mode being disabled.

Resolves: https://github.com/hashicorp/consul/issues/11330

Question for reviewer:
- Open to suggestions on how I can check for status change in a check that was removed (so that tests can be updated)
- Should `SyncChanges` or `SyncFull` be used to force the health check update event to fire before the health check is removed? (And are there any negative repercussions of that?)

### DIFF
--- a/.changelog/11332.txt
+++ b/.changelog/11332.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-agent: update maintenance check to passing instead of removing the check
+agent: update maintenance check to passing before removing
 ```
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -4068,7 +4068,7 @@ func (a *Agent) EnableServiceMaintenance(serviceID structs.ServiceID, reason, to
 
 	// Check if maintenance mode is not already enabled
 	checkID := serviceMaintCheckID(serviceID)
-	if a.State.Check(checkID) != nil && a.State.Check(checkID).Status != api.HealthPassing {
+	if a.State.Check(checkID) != nil {
 		return nil
 	}
 
@@ -4077,7 +4077,7 @@ func (a *Agent) EnableServiceMaintenance(serviceID structs.ServiceID, reason, to
 		reason = defaultServiceMaintReason
 	}
 
-	// New Critical Health Check
+	// Create and register the critical health check
 	check := &structs.HealthCheck{
 		Node:           a.config.NodeName,
 		CheckID:        checkID.ID,
@@ -4089,17 +4089,7 @@ func (a *Agent) EnableServiceMaintenance(serviceID structs.ServiceID, reason, to
 		Type:           "maintenance",
 		EnterpriseMeta: checkID.EnterpriseMeta,
 	}
-
-	// If check exists, update status, else create a new check
-	if a.State.Check(checkID) != nil {
-		a.State.UpdateCheck(checkID, api.HealthCritical, "")
-	} else {
-		err := a.AddCheck(check, nil, true, token, ConfigSourceLocal)
-		if err != nil {
-			return nil
-		}
-	}
-
+	a.AddCheck(check, nil, true, token, ConfigSourceLocal)
 	a.logger.Info("Service entered maintenance mode", "service", serviceID.String())
 
 	return nil
@@ -4122,11 +4112,9 @@ func (a *Agent) DisableServiceMaintenance(serviceID structs.ServiceID) error {
 	// Update check to trigger an event for watchers
 	a.State.UpdateCheck(checkID, api.HealthPassing, "")
 	// Make sure state change is propagated
-	err := a.State.SyncFull()
-	if err != nil {
-		return err
-	}
-
+	a.State.SyncChanges()
+	// Deregister the maintenance check
+	a.RemoveCheck(checkID, true)
 	a.logger.Info("Service left maintenance mode", "service", serviceID.String())
 
 	return nil
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -3377,8 +3377,9 @@ func TestAgent_Service_MaintenanceMode(t *testing.T) {
 		t.Fatalf("err: %s", err)
 	}
 
-	// Ensure the check has updated status
-	if found := a.State.Check(checkID); found != nil && found.Status != api.HealthPassing {
+	// Ensure the check was deregistered
+
+	if found := a.State.Check(checkID); found != nil {
 		t.Fatalf("should have deregistered maintenance check")
 	}
 
@@ -3392,10 +3393,7 @@ func TestAgent_Service_MaintenanceMode(t *testing.T) {
 	if check == nil {
 		t.Fatalf("should have registered critical check")
 	}
-	if check.Notes != "broken" {
-		t.Fatalf("bad: %#v", check)
-	}
-	if check.Status != api.HealthCritical {
+	if check.Notes != defaultServiceMaintReason {
 		t.Fatalf("bad: %#v", check)
 	}
 }
@@ -3622,10 +3620,8 @@ func TestAgent_NodeMaintenanceMode(t *testing.T) {
 	// Leave maintenance mode
 	a.DisableNodeMaintenance()
 
-	// Ensure the check indicates passing status
-	if check.Status == api.HealthPassing {
-		t.Fatalf("bad: %#v", check)
-	}
+	// Ensure the check was deregistered
+	requireCheckMissing(t, a, structs.NodeMaint)
 
 	// Enter maintenance mode without passing a reason
 	a.EnableNodeMaintenance("", "")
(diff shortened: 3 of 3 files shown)


jkirschner-hashicorp:
Hi @jakubgs,

The engineering team is taking a look at your questions from the description (e.g., any guidance on how to add a test).

Additionally, they may consider whether this "update to passing before check removal" change should be done in a more central location. Doing so would affect other check types, which have benefits and/or unintended side effects.

I'll get back to you at the end of this week if you haven't heard from us already by then.

dhiaayachi:
Hi @jakubgs,
Thank you for the contribution and sorry for the delay on this. I went through the changes and it look good, the only missing piece would be a test.

I think you can build a test based on those tests `TestAgent_NodeMaintenance_Enable` and `TestAgent_NodeMaintenance_Disable` (or modify these tests) to check that the check update was correctly propagated. 

Let me know if you have questions or need more details on how to test this.

jakubgs:
Yes, that's exactly what I'm asking how to do. How can I possible check the state of a check that was removed? This is not trivial.

dhiaayachi:
@jakubgs Yes a good point, let me dig a bit to see if we can intercept the update somehow, or if we can test at a lower level by mocking the update.

dnephin:
If this fix is related to `consul watch`, maybe we should use that watch machinery and assert that we see the event we expect in the `watch` stream? I haven't written any test like that, so I'm not sure exactly what function we would use there, but something along those lines seems like what we want.

jakubgs:
Yes, considering the purpose is to have `consul watch` to notice the maintenance mode being turned off we could try to use that code to detect if the change has been registered. Although that makes the unit test less of a "unit" test and more of an "integration" test. But it's worth a try. No idea how that would work tho. I might not have time to look into that this week.

jakubgs:
I tried something like this:
```diff
diff --git a/agent/agent_test.go b/agent/agent_test.go
index 6638fa35c..4085a9c9a 100644
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -3355,9 +3355,20 @@ func TestAgent_NodeMaintenanceMode(t *testing.T) {
                t.Fatalf("bad: %#v", check)
        }
 
+       nodeMaintCheck := a.State.Check(structs.NodeMaintCheckID)
+
+       // Ensure the check state was updated
+       require.Equal(t, api.HealthCritical, nodeMaintCheck.Status)
+
+       fmt.Printf("%v\n", nodeMaintCheck)
+
        // Leave maintenance mode
        a.DisableNodeMaintenance()
 
+       // Ensure the check state was updated
+       fmt.Printf("%v\n", nodeMaintCheck)
+       require.Equal(t, api.HealthPassing, nodeMaintCheck.Status)
+
        // Ensure the check was deregistered
        requireCheckMissing(t, a, structs.NodeMaint)
```
But it fails, since the state is unchanged:
```
--- FAIL: TestAgent_NodeMaintenanceMode (0.44s)
    agent_test.go:3370: 
        	Error Trace:	agent_test.go:3370
        	Error:      	Not equal: 
        	            	expected: "passing"
        	            	actual  : "critical"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-passing
        	            	+critical
        	Test:       	TestAgent_NodeMaintenanceMode
FAIL
exit status 1
FAIL	github.com/hashicorp/consul/agent	0.469s
```
Which I guess means I'm either checking too soon, or the instance I'm checking is outdated.

jakubgs:
I also tried moving the second check after `requireCheckMissing()` but it fails the same way.

jakubgs:
It appears the `UpdateCheck()` function uses something called `TriggerSyncChanges()`:
https://github.com/hashicorp/consul/blob/cc2abb79baa5989821ede371a19464fa691ec000/agent/local/state.go#L704-L707
Which appears to be set here to `a.sync.SyncChanges.Trigger`:
https://github.com/hashicorp/consul/blob/3666401ae3ae924b1d1acd9261f5e187c610ac07/agent/agent.go#L558-L563
Which appears to be some kind of internal dark magic for service/client syncing.

jakubgs:
https://github.com/hashicorp/consul/blob/3666401ae3ae924b1d1acd9261f5e187c610ac07/agent/ae/ae.go#L76-L78
https://github.com/hashicorp/consul/blob/3666401ae3ae924b1d1acd9261f5e187c610ac07/agent/ae/trigger.go#L3-L19

Yeah, I really don't know what I'm looking at.

jakubgs:
One notable thing is that for the `TestAgent_NodeMaintenanceMode` we create a `NewTestAgent`:
https://github.com/hashicorp/consul/blob/3666401ae3ae924b1d1acd9261f5e187c610ac07/agent/agent_test.go#L3444-L3445
And while normal `Agent` defines its `TriggerSyncChanges`:
https://github.com/hashicorp/consul/blob/3666401ae3ae924b1d1acd9261f5e187c610ac07/agent/agent.go#L563
The `TestAgent` defined no such thing in its `Start` function:
https://github.com/hashicorp/consul/blob/3666401ae3ae924b1d1acd9261f5e187c610ac07/agent/testagent.go#L134-L136
So I'm not sure how this tracking of events is even supposed to work.

jakubgs:
I tried adding some prints in `State.UpdateCheck()` and found the memory address changes every time because of:
https://github.com/hashicorp/consul/blob/3666401ae3ae924b1d1acd9261f5e187c610ac07/agent/local/state.go#L660-L670
Which means we can't really monitor a copy previously fetched with `a.State.Check()`.

jakubgs:
I'm confused, where in the definition of `TestAgent` is there `State`?
https://github.com/hashicorp/consul/blob/3666401ae3ae924b1d1acd9261f5e187c610ac07/agent/testagent.go#L38-L44
I can't find it.

jakubgs:
Yeah. I don't get it, how exactly is `State` attached to `TestAgent`?

jakubgs:
@dhiaayachi can you explain to me how exactly `State` is part of `TestAgent`? And is there some existing way to watch events in the `State`?

dnephin:
Hi @jakubgs, thanks for working on these tests! 

The `TestAgent` has an embedded `*Agent` as the last field, so you can access `local.State` from `TestAgent.Agent.State`. But I'm not sure if that is our best option for emulating the behaviour of a watch.

I looked at the watch command, and I found the two API queries it is using here: https://github.com/hashicorp/consul/blob/v1.10.4/api/watch/funcs.go#L197-L199.  I'm not sure which one of those two we need to use , but they should both work about the same.

Since we are testing `DisableServiceMaintenance`, and that is called in only one place by an API handler, it seems like using the HTTP API to drive some of the test is pretty reasonable. 

To test the "watch" behaviour I think we'll probably need to use at least one goroutine. We can start a blocking query in a goroutine using `httpHandlers.AgentChecks`, and a query parameter of `index=` where the value is the last index of that check. You can call `AgentChecks` one time without `index=` to get the intial index.

That should block until the update is received. Then in the main test goroutine we can check to see if the event is received within some time window (probably 100ms is plenty of time to wait). I found an example of something like that here: 
https://github.com/hashicorp/consul/blob/v1.10.4/agent/service_checks_test.go#L70-L75

I hope this information helps. Please do continue to ask questions if anything is not clear. We'll do our best to get back to you promptly, but some times there may be a few days lag.

Amier3:
Hey @jakubgs,

Was the information given above able to help out? As Daniel said we're happy to answer any additional questions

jakubgs:
I just came back from a Christmas break, and this is not high priority for me, more like nice-to-have.

Thanks for the help. I'll get to it when I'll get to it.

siddarthkay:
I'm taking over this branch and will continue to write the test so that we can get this PR merged.

siddarthkay:
@dnephin @Amier3 : I tweaked what @jakubgs had done a bit and now instead of deregistering the check of maintenance mode I update its state to passing when maintenance mode is disabled.

This makes it easier to verify the changes and to write a test for it without making things too complicated in this PR.

I took this approach after trying various approaches to write a test for @jakubgs's changes and I was unable to get a deterministic test to work.

Would love to hear your thoughts on this approach.

Thank you.

siddarthkay:
Any feedback?
cc @dhiaayachi

siddarthkay:
Thanks for the input @jakubgs 
I've reverted my changes in the latest commit.
My intent was to get feedback but I haven't gotten any so far

For the reviewers : I still need feedback on the approach in the commit here -> https://github.com/hashicorp/consul/pull/11332/commits/02c65d66be49fddbbd94240c6cacb359928f112c

cc @blake @zalimeni @dhiaayachi @dnephin @Amier3 @rboyer 

next steps : I'll add few failed attempts at writing tests and maybe we can find a way to prove our change actually works.
OR
We deploy a version of consul on one of our nodes and prove that @jakubgs's work fixes the un-necessary alerts when we put our hosts in maintenance mode.

Thank you!

siddarthkay:
Can we please get some feedback on this PR ? 
cc @blake @zalimeni @dhiaayachi @dnephin @Amier3 @rboyer @dhiaayachi @dnephin @Amier3
if you're not the right person could you tag someone who can review this PR ?

Thank you

jakubgs:
It's not inactive, we are waiting for a review from the team.

siddarthkay:
I've just rebased this PR on latest main, This is not stale we are waiting on review form the team at hashicorp.

PR #11334, opened 2021-10-15
Get service identities from role by accessor

jkirschner-hashicorp:
Make consistent with how node identities are accessed from an `ACLRole`.

I noticed an inconsistency between how service identities and node identities are accessed from a role when merging into the service and node identity lists for an identity:
https://github.com/hashicorp/consul/blob/a5c292c000ace8b9bd6a0231f86039ba629753cd/agent/consul/acl.go#L639-L640

Node identities come from an accessor:
https://github.com/hashicorp/consul/blob/a5c292c000ace8b9bd6a0231f86039ba629753cd/agent/structs/acl_oss.go#L85-L95

Whereas service identities are accessed directly, despite previously in the same function by an accessor from `ACLIdentity`:
https://github.com/hashicorp/consul/blob/a5c292c000ace8b9bd6a0231f86039ba629753cd/agent/consul/acl.go#L614

This PR copies `ACLIdentity:ServiceIdentityList` to create `ACLRole:ServiceIdentityList`, then uses that new function to access that information from the role (rather than directly).

**Question for reviewer:**
- Is there some reason for this existing inconsistency? (I don't _know_ whether this change is needed, I only know that it's a seeming inconsistency in the code which isn't explained with a comment.)

**TODO:**
- Add changelog if reviewer(s) agree this change is needed

### DIFF
--- a/agent/consul/acl.go
+++ b/agent/consul/acl.go
@@ -636,7 +636,7 @@ func (r *ACLResolver) resolvePoliciesForIdentity(identity structs.ACLIdentity) (
 		for _, link := range role.Policies {
 			policyIDs = append(policyIDs, link.ID)
 		}
-		serviceIdentities = append(serviceIdentities, role.ServiceIdentities...)
+		serviceIdentities = append(serviceIdentities, role.ServiceIdentityList()...)
 		nodeIdentities = append(nodeIdentities, role.NodeIdentityList()...)
 	}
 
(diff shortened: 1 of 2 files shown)


jkirschner-hashicorp:
> Interesting! It seems like the purpose of this function is to make a copy of the identities. I'm not sure why we would need a copy. I don't think `ACLResolver` modifies these.

So I should go ahead and close this (as unneeded) then?

jkirschner-hashicorp:
@dnephin : I assume no one else has thoughts on this. If I don't hear anything, I'll close this in a few days.

mkeeler:
Closing as its unneeded.

Issue #11258, opened 2021-10-08
URL-decode resource names for HTTP API


kisunji:
### Overview

Consul CLI has many [commands](https://www.consul.io/commands) that make HTTP calls to the local agent's HTTP API. 

Since some resource names are not URL safe, the caller must URL-encode certain characters. However, our handlers naively parse the URL as-is without a decode step, making it impossible to interact with resources that may have URL unsafe characters in their name or ID.

We should URL-decode path arguments when HTTP handlers try to parse the resource name ([example](https://github.com/hashicorp/consul/blob/cc310224aa24b3d7ed88e700be7fc95154dbbd9f/agent/agent_endpoint.go#L343-L344); can find others by searching for `strings.TrimPrefix(req.URL.Path`)

Related issues:
https://github.com/hashicorp/consul/issues/11191

### Correction
We previously assumed we also had to URL-encode corresponding HTTP calls in `api` module. This has been proven wrong in https://github.com/hashicorp/consul/pull/12184 - we only need to add decoding on handlers.

### Notes to contributors
Since the number of modified files is expected to be large, you are welcome to start work on a subset of endpoints and ask for feedback on your proposed solution/pattern before applying it to more places. 

Feel free to ask any questions and a member of our team will respond.

### Files with endpoints to cover (40 total)
- [x] ./agent/acl_endpoint.go (#11957)
- [x] ./agent/agent_endpoint.go (#11335)
- [x] ./agent/catalog_endpoint.go (#11957)
- [x] ./agent/config_endpoint.go (#11957)
- [x] ./agent/connect_ca_endpoint.go (no affected endpoints)
- [x] ./agent/coordinate_endpoint.go (#11957)
- [x] ./agent/discovery_chain_endpoint.go (#11957)
- [x] ./agent/event_endpoint.go (#12103)
- [x] ./agent/federation_state_endpoint.go (#12103)
- [x] ./agent/health_endpoint.go (#12103)
  - 3 endpoints were missed ([example](https://github.com/dttung2905/consul/blob/39b0d783d95de678156a80b5eca425d394e34f7c/agent/health_endpoint.go#L83
)) (#12297)
- [x] ./agent/intentions_endpoint.go (#12103)
- [x] ./agent/kvs_endpoint.go (#12103)
- [x] ./agent/operator_endpoint.go (no affected endpoints)
- [x] ./agent/prepared_query_endpoint.go (#12103)
- [x] ./agent/session_endpoint.go (#12103)
[...]

PR #11335, opened 2021-10-15
URL-encode/decode resource names for HTTP API

littlestar642:
### Overview

Relates to: #11258

To encode/decode the path params in HTTP API, we can use golang's `url.PathEscape` and `url.PathUnescape` methods.

url.PathEscape - PathEscape escapes the string so it can be safely placed inside a URL path segment, replacing special characters (including /) with %XX sequences as needed.

### DIFF
--- a/agent/agent_endpoint.go
+++ b/agent/agent_endpoint.go
@@ -380,7 +380,10 @@ func (s *HTTPHandlers) AgentServices(resp http.ResponseWriter, req *http.Request
 // blocking watch using hash-based blocking.
 func (s *HTTPHandlers) AgentService(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
 	// Get the proxy ID. Note that this is the ID of a proxy's service instance.
-	id := strings.TrimPrefix(req.URL.Path, "/v1/agent/service/")
+	id, err := getPathSuffixUnescaped(req.URL.Path, "/v1/agent/service/")
+	if err != nil {
+		return nil, err
+	}
 
 	// Maybe block
 	var queryOpts structs.QueryOptions
@@ -400,7 +403,7 @@ func (s *HTTPHandlers) AgentService(resp http.ResponseWriter, req *http.Request)
 	}
 
 	// need to resolve to default the meta
-	_, err := s.agent.delegate.ResolveTokenAndDefaultMeta(token, &entMeta, nil)
+	_, err = s.agent.delegate.ResolveTokenAndDefaultMeta(token, &entMeta, nil)
 	if err != nil {
 		return nil, err
 	}
@@ -641,7 +644,10 @@ func (s *HTTPHandlers) AgentJoin(resp http.ResponseWriter, req *http.Request) (i
 	}
 
 	// Get the address
-	addr := strings.TrimPrefix(req.URL.Path, "/v1/agent/join/")
+	addr, err := getPathSuffixUnescaped(req.URL.Path, "/v1/agent/join/")
+	if err != nil {
+		return nil, err
+	}
 
 	if wan {
 		if s.agent.config.ConnectMeshGatewayWANFederationEnabled {
@@ -701,7 +707,10 @@ func (s *HTTPHandlers) AgentForceLeave(resp http.ResponseWriter, req *http.Reque
 	// Check if the WAN is being queried
 	_, wan := req.URL.Query()["wan"]
 
-	addr := strings.TrimPrefix(req.URL.Path, "/v1/agent/force-leave/")
+	addr, err := getPathSuffixUnescaped(req.URL.Path, "/v1/agent/force-leave/")
+	if err != nil {
+		return nil, err
+	}
 	if wan {
 		return nil, s.agent.ForceLeaveWAN(addr, prune, entMeta)
 	} else {
@@ -788,7 +797,11 @@ func (s *HTTPHandlers) AgentRegisterCheck(resp http.ResponseWriter, req *http.Re
 }
 
 func (s *HTTPHandlers) AgentDeregisterCheck(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
-	checkID := structs.NewCheckID(types.CheckID(strings.TrimPrefix(req.URL.Path, "/v1/agent/check/deregister/")), nil)
+	ID, err := getPathSuffixUnescaped(req.URL.Path, "/v1/agent/check/deregister/")
+	if err != nil {
+		return nil, err
+	}
+	checkID := structs.NewCheckID(types.CheckID(ID), nil)
 
 	// Get the provided token, if any, and vet against any ACL policies.
 	var token string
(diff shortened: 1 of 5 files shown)


kisunji (approved):
LGTM! Thanks for this

kisunji:
Will push a commit to add a changelog to this PR then commit

agent/agent_endpoint.go
```diff
@@ -359,7 +360,11 @@ func (s *HTTPHandlers) AgentServices(resp http.ResponseWriter, req *http.Request
 // blocking watch using hash-based blocking.
 func (s *HTTPHandlers) AgentService(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
 	// Get the proxy ID. Note that this is the ID of a proxy's service instance.
-	id := strings.TrimPrefix(req.URL.Path, "/v1/agent/service/")
+	id, err := url.PathUnescape(strings.TrimPrefix(req.URL.Path, "/v1/agent/service/"))
+	if err != nil {
+		s.agent.logger.Error(fmt.Sprintf("failure in unescaping path param: %s", err.Error()))
+		id = strings.TrimPrefix(req.URL.Path, "/v1/agent/service/")
+	}
```
jkirschner-hashicorp, line 367: I recommend that we simplify this to make it easier to maintain and apply consistently to all endpoints (remove redundant operations and boilerplate). I think this can be solved by abstracting into a function call like:
```
func GetPathSuffixUnescaped(path string, prefixToTrim string) (string, error) {
  // The suffix may be URL-encoded, so attempt to decode
  suffixRaw := strings.TrimPrefix(path, prefixToTrim)
  suffixUnescaped, err := url.PathUnescape(suffixRaw)
  
  if err != nil {
    return suffixRaw, fmt.Errorf("failure in unescaping path param %q: %v", suffixRaw, err)
  }
  
  return suffixUnescaped, nil
}
```

Then, each endpoint replacement would look something like this:
```suggestion
	id, err := GetPathSuffixUnescaped(req.URL.Path, "/v1/agent/service/")
	if err != nil {
		s.agent.logger.Error(err)
	}
```

---

@kisunji : I'm not sure the best place in the source code to put a shared function like this. Thoughts on that? And whether you agree with this code change suggestion?

@littlestar642: are you interested in making such an enhancement to the PR? If not (we realize some time has passed since you posted this), let us know!

jkirschner-hashicorp (commented):
> PR ready to merge, awaiting decision to backport single change or batch them in next release

@kisunji, @littlestar642: we decided we'll try to merge the changes just for this endpoint, then internally consider (manually) backporting to 1.9 and 1.10.

Since there's some work that goes into manual backporting, I think it makes sense to address the suggested change in my review first (as it should make rolling out to other endpoints easier). @littlestar642: let me know if you're interested in making that change (before we merge and then begin the backport process).

agent/agent_endpoint.go
kisunji: Great suggestion! I think `agent/http.go` is a reasonable place for a general function that can be reused by all endpoints.

jkirschner-hashicorp:
@kisunji : I rebased and made the change I proposed. This is ready for your final review.

agent/agent_endpoint.go
jkirschner-hashicorp: Thanks - change made! I also added a unit test for that function.

agent/agent_endpoint.go
```diff
@@ -380,7 +380,10 @@ func (s *HTTPHandlers) AgentServices(resp http.ResponseWriter, req *http.Request
 // blocking watch using hash-based blocking.
 func (s *HTTPHandlers) AgentService(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
 	// Get the proxy ID. Note that this is the ID of a proxy's service instance.
-	id := strings.TrimPrefix(req.URL.Path, "/v1/agent/service/")
+	id, err := getPathSuffixUnescaped(req.URL.Path, "/v1/agent/service/")
```
jkirschner-hashicorp, line 383: Ran this against a consul binary just to see for myself that it worked.

```
# Successfully decoded!
$ curl --request GET localhost:8500/v1/agent/service/counting%2Fid
unknown service ID: counting/idj
```

agent/agent_endpoint.go
```diff
@@ -380,7 +380,10 @@ func (s *HTTPHandlers) AgentServices(resp http.ResponseWriter, req *http.Request
...
+	if err != nil {
+		s.agent.logger.Error(err.Error())
```
jkirschner-hashicorp, line 385: When I ran against a consul binary with an invalid encoded value, this is what happened:
```
$ curl --request GET localhost:8500/v1/agent/service/counting%2id
400 Bad Request

# No error appears in consul log output
```

@kisunji : Is this what we'd expect/want to happen?

kisunji: Summarizing our discussion over slack: most middlewares reject invalid URLs early so the request never reaches the handler. It's almost impossible to externally test the error check via an HTTP client because you either: 1) face early rejection by go's http lib; or 2) encode into a valid url and never hit the error code path.

agent/http.go
```diff
@@ -1120,3 +1120,15 @@ func (s *HTTPHandlers) parseFilter(req *http.Request, filter *string) {
...
 	}
 }
+
+func getPathSuffixUnescaped(path string, prefixToTrim string) (string, error) {
+	// The suffix may be URL-encoded, so attempt to decode
+	suffixRaw := strings.TrimPrefix(path, prefixToTrim)
+	suffixUnescaped, err := url.PathUnescape(suffixRaw)
+
+	if err != nil {
+		return suffixRaw, fmt.Errorf("failure in unescaping path param %q: %v", suffixRaw, err)
```
kisunji, line 1130:
```suggestion
		return suffixRaw, fmt.Errorf("cannot unescape path param %q: %w", suffixRaw, err)
```

kisunji (changes requested):
Looks good! @jkirschner-hashicorp We discussed that we should return early on error from `getPathSuffixUnescaped`. Once that change is made we can merge 👍

agent/http.go
jkirschner-hashicorp: @kisunji : I agree with the wording change, but not sure I agree with the change from `%v` to `%w` for the error.

Per [this Golang page](https://go.dev/blog/go1.13-errors):
> ...wrapping an error makes that error part of your API. If you don’t want to commit to supporting that error as part of your API in the future, you shouldn’t wrap the error.

> It’s important to remember that whether you wrap or not, the error text will be the same. A person trying to understand the error will have the same information either way; the choice to wrap is about whether to give programs additional information so they can make more informed decisions, or to withhold that information to preserve an abstraction layer.

In this case, I don't think the caller needs to be able to able to unwrap the returned error to specifically look for the unescape error.

kisunji: That's totally fair! In this case it doesn't matter too much since our http handler ultimately consumes all the errors before returning plaintext or json to the client but certainly a good thing to keep in mind.

agent/agent_endpoint.go
```diff
@@ -1283,7 +1325,12 @@ func (s *HTTPHandlers) AgentDeregisterService(resp http.ResponseWriter, req *htt
 
 func (s *HTTPHandlers) AgentServiceMaintenance(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
 	// Ensure we have a service ID
-	sid := structs.NewServiceID(strings.TrimPrefix(req.URL.Path, "/v1/agent/service/maintenance/"), nil)
+	serviceID, err := getPathSuffixUnescaped(req.URL.Path, "/v1/agent/service/maintenance/")
+	if err != nil {
+		return nil, err
```
jkirschner-hashicorp, line 1330: Why do we sometime use `return nil, err` within these functions, and other times use:
```
		fmt.Fprint(resp, "some error message")
		return nil, nil
```

?

jkirschner-hashicorp: @kisunji, @rboyer, @kyhavlov : any ideas why this is? (I want to make sure my use of `return nil, err` is what I should be doing in these cases.)

kisunji: I raised https://github.com/hashicorp/consul/issues/11393 for this issue! We prefer `return nil, err` over writing to response.

I'm not sure about the historic context, but in go's stdlib the handlers don't return any errors and you write to `resp` directly. This pattern may have been carried over to consul's handlers.

kisunji: Linking to @Mathew-Estafanous 's https://github.com/hashicorp/consul/pull/11599 as well who started work on the refactoring

jkirschner-hashicorp: @kisunji : thanks for creating that issue! If I understand correctly, `return nil, err` is what we want to do here and elsewhere. And applying elsewhere is captured in #11393. Yes? If so, do you want to do a final review/approve?

kisunji (approved):
LGTM!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/539327.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit a36ddc31c7b9cac96eaf3c93ce9a724011fdbdc0 onto `release/1.11.x` succeeded!

Issue #11159, opened 2021-09-27
ca: Support separate Vault providers per DC


dnephin:
Currently the [Connect CA](https://www.consul.io/docs/connect/ca) requires that when the [Vault provider](https://www.consul.io/docs/connect/ca/vault) is used, all Consul DCs must either:
* point at the same Vault cluster
* use [Vault replication](https://www.vaultproject.io/docs/enterprise/replication) to mirror the certificate at `root_pki_path` to all clusters, or
* must manually copy the root cert to all clusters

This requirement comes with a number of disadvantages. The manual option is extra work that shouldn't be necessary, and both of the other options require setting separate paths otherwise things are pretty broken.

We can allow a separate Vault cluster per Consul DC by changing the two calls to `provider.ActiveRoot` in `SignCertificate` and `initializeSecondaryCA` to use the root CA that is fetched from the primary DC, instead of attempting to retrieve it directly from Vault. 

This would not only enable separate Vault instances, but would also allow someone to use different CA providers per datacenter.  While that may not be a desirable long term option, it does seem like a pretty important thing to support for doing an incremental switch over from one provider to another one DC at a time.

*Acceptance Criteria*

1. Write a test showing that a separate Vault instance can be used for the primary and secondary Consul DCs, and that the secondary is able to sign leaf certs.
2. Address the structural problems in the code that make it trivial to re-introduce the problem. Currently `ca.Provider` interface makes no distinction between the methods that are safe for use in the secondary DC. We should split up this interface into `ProviderPrimary` and `ProviderSecondary`. Also, any methods on `CAManager` should either be safe for both primary and secondary, or they should be explicitly named `primaryX` or `secondaryX` to make it clear where they are used. In the future we may be able to split those methods further onto separate types.

PR #11338, opened 2021-10-15
ca: clearly identify methods that are primary-only or secondary-only

dnephin:

The `CAMananger` has two distinct "modes". It can operator as a primary (in the primary DC), or a secondary (in a secondary DC). Some of the methods on `CAManager` are clearly identified as such, but not all.

This PR renames all methods that are primary-only or secondary-only to start with a `primary` or `secondary` prefix. This is one small step to help make these two "modes" more obvious, which will help prevent issues like the one described in #11159. In the future we may be able to make these two modes more concrete in code by having separate types for each.

### DIFF
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -1539,12 +1539,12 @@ func (c *CAManager) SignCertificate(csr *x509.CertificateRequest, spiffeID conne
 	return &reply, nil
 }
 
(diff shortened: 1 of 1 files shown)


hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/490362.

PR #11339, opened 2021-10-15
ca: reduce use of state in the secondary

dnephin:
Branched from #11338

The motivation is largely the same, make it easier to see the two distinct modes. Another barrier to creating types for these two modes is access to shared state. In this case I found that checking the state is unnecessary because the callers have already set the correct state.

I opened #11346 as potential next steps along these lines.

### DIFF
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -183,19 +183,15 @@ func (e *caStateError) Error() string {
 }
 
 // secondarySetPrimaryRoots updates the most recently seen roots from the primary.
-func (c *CAManager) secondarySetPrimaryRoots(newRoots structs.IndexedCARoots) error {
+func (c *CAManager) secondarySetPrimaryRoots(newRoots structs.IndexedCARoots) {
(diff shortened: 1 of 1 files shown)


agent/consul/leader_connect_ca.go
```diff
@@ -1311,29 +1305,17 @@ func (c *CAManager) secondaryInitializeProvider(provider ca.Provider, roots stru
...
 	return nil
 }
 
-// secondaryIsCAConfigured returns true if we have been initialized as a secondary datacenter's CA.
-func (c *CAManager) secondaryIsCAConfigured() bool {
-	c.stateLock.Lock()
-	defer c.stateLock.Unlock()
-	return c.actingSecondaryCA
+// secondaryHasProviderRoots returns true after providerRoot has been set. This
+// method is used to detect when the provider has been fully initialized a
```
kyhavlov, line 1312:
```suggestion
// method is used to detect when the provider has been fully initialized with a
```

kyhavlov (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/509891.

Issue #11336, opened 2021-10-15
Proposal: move Connect CA implementation to a new package


dnephin:
The [Consul Connect CA](https://www.consul.io/docs/connect/ca) is a sub-system that is responsible for managing a Certificate Authority for the service mesh and auto-config. Currently it is implemented across a number of packages and files:
* `agent/consul/connect_ca_endpoint.go`
* `agent/consul/leader_connect_ca.go`
* `agent/consul/server_connect.go`
* `agent/cache-types/connect_ca_leaf.go`
* `agent/consul/state/connect_ca.go`
* `agent/connect/ca`

This issue proposes we restructure this sub-system to make it easier to test, and easier to develop.

The proposed structure is:
* `agent/connect/ca` - the "core logic" of the system, currently in `agent/consul/leader_connect_ca.go`, and the state store functions from `agent/consul/state/connect_ca.go`
* `agent/connect/ca/carpc` - the RPC server, currently in `agent/consul/connect_ca_endpoint.go`
* `agent/connect/ca/caprovider` - the providers, currently in `agent/connect/ca`
* `agent/connect/ca/caleafs` - the client implementation for generating leaf certs, currently in `agent/cache-types/connect_ca_leaf.go`

PR #11350, opened 2021-10-18
ca: prepare CAMananger to be moved into a new package

dnephin:
Branched from #11339
Related to #11336

This PR makes  bunch of small changes (one per commit) to allow us to "refactor -> move" the `CAMananger` with a minimal diff. By splitting out the changes in this way, `git diff` will be able to show us just the few lines that changed as part of the file move, instead of making it look like a full delete, and full create of separate files. This can be seen in [this commit](https://github.com/hashicorp/consul/compare/dnephin/ca-manager-move-to-new-package...dnephin/ca-manager-move-to-new-package-2), where the file is moved to a new package, but the diff is only +/- 20 lines.

I can split some of these commits into separate PRs if anything is controversial, but I think most of these are valuable clean ups, even if we don't move `CAMananger` to a new package.

TODO: 
* [ ] [some flaky tests](https://app.circleci.com/pipelines/github/hashicorp/consul/22876/workflows/511035db-e7bf-45bc-9e7b-4824600859e6/jobs/476763) to look into. These were flaky before, but I'd like to make sure I haven't made it worse, and maybe look at making it less flaky if possible.
* [ ] re-add rate limiter for retry loop

### DIFF
--- a/agent/consul/connect_ca_backend.go
+++ b/agent/consul/connect_ca_backend.go
@@ -1,0 +1,65 @@
+package consul
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/consul/agent/consul/state"
+	"github.com/hashicorp/consul/agent/structs"
+)
+
+type caBackend struct {
+	*Server
+}
+
+func (c *caBackend) ServersSupportMultiDCConnectCA() (bool, bool) {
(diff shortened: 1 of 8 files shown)


jmurret:

PR #11356, opened 2021-10-20
Docs: Clarify distributed tracing vs Istio

bbros-dev:
Here are supporting references:
- [By design, Consul does not initiate traces on ingress](https://github.com/hashicorp/consul/blob/v1.8.0/agent/xds/listeners.go#L930-L933)
- [The Istio sidecar proxy (Envoy) generates the initial headers if they are not provided by the request](https://istio.io/latest/about/faq/#distributed-tracing)
- Istio Distributed Tracing Tasks documentation:

    - [Jaeger](https://istio.io/latest/docs/tasks/observability/distributed-tracing/jaeger/)
    - [Lightstep](https://istio.io/latest/docs/tasks/observability/distributed-tracing/lightstep/)
    - [Zipkin](https://istio.io/latest/docs/tasks/observability/distributed-tracing/zipkin/).

### DIFF
--- a/website/content/docs/intro/vs/istio.mdx
+++ b/website/content/docs/intro/vs/istio.mdx
@@ -42,10 +42,17 @@ whereas Consul is able to efficiently distribute updates and perform all
 work on the edge.
 
 Consul provides layer 7 features for path-based routing, traffic shifting,
-load balancing, and telemetry. Consul enforces authorization and identity to
+load balancing and some telemetry. Consul enforces authorization and identity to
 layer 4 only &mdash; either the TLS connection can be established or it can't.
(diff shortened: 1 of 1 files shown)


Amier3 (commented):
Hey @bbros-dev

Thank you for the effort put forward to clarify our docs! I think this section could use some improvements, but some of the assumptions made in these changes might add to the confusion about differences between Consul and Istio ( see comment below ). In addition to the accuracy of the changes, we're currently in the process assessing the value of the ' Consul vs ' section of our docs all together.  Depending on that assessment, the content of these pages could change significantly. These things in combination mean that now probably isn't the right time for this change.

Regardless, we appreciate the contribution and look forward to more PRs from you!

PR #11386, opened 2021-10-21
Update Consul ECS documentation with health sync changes

erichaberkorn:
This also switches the task startup image to an svg so it isn't pixelated
anymore


website/public/img/ecs-task-startup.svg
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
pglass, line 1: @erichaberkorn When I view this image in the preview, the `T0 T1 T2` headers have some empty space below them

Screen Shot 2021-11-03 at 2 21 44 PM
p

website/content/docs/ecs/architecture.mdx
```diff
@@ -57,3 +63,12 @@ token does not yet exist.
...
 tokens from AWS Secrets Manager when they start.
+
+### ECS Health Check Syncing
+
+Health checks are automatically synced from ECS into Consul.  This syncing
+automatically happens for all application containers that are marked as
+`essential`, have ECS `healthChecks`, and don't have native Consul health
+checks configured. The `mesh-init` container creates a TTL health check for
+every container that fits these criteria and the `health-sync` container ensures
+that the ECS and Consul health checks remain in sync.
```
trujillo-adam, line 74:
```suggestion
If the following conditions apply, ECS health checks automatically sync with Consul health checks for all application containers:

* marked as `essential` 
* have ECS `healthChecks` 
* are not configured with native Consul health checks 

The `mesh-init` container creates a TTL health check for
every container that fits these criteria and the `health-sync` container ensures
that the ECS and Consul health checks remain in sync.
```
Trying out a bulleted list for this information so that it's easier to scan - ignore if it doesn't seem like an improvement.

trujillo-adam (approved):
I made a few suggestions but nothing critical. 
I see some opportunities to improve this section outside of the scope of these changes, though.

trujillo-adam (commented):
LGTM

PR #11390, opened 2021-10-21
ui: Allow spaces in template interpolation

radiantly:
Allows spaces and newlines in template interpolation, for example `${ Datacenter }`. This is nearer to the js and mustache styles of interpolation that we're trying to replicate.

### DIFF
--- a/ui/packages/consul-ui/app/helpers/render-template.js
+++ b/ui/packages/consul-ui/app/helpers/render-template.js
@@ -5,11 +5,13 @@ import { inject as service } from '@ember/service';
 // what this regex does
 // (?:\$|\{)            - Match either $ or {
 // \{                   - Match {
+// [\s\n\r]*            - Skip spaces/newlines
 // ([a-z.0-9_-]+)       - Capturing group
+// [\s\n\r]*            - Skip spaces/newlines
 // (?:(?<=\$\{[^{]+)    - Use a positive lookbehind to assert that ${ was matched previously
 //   |\}            )   - or match a }
 // \}                   - Match }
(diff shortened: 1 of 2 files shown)


johncowen:
Hi @radiantly 

Unfortunately, if I'm not mistaken, it seems that a RegExp feature (positive lookbehind) that we've used here, and in the previous PR, is not supported any version of Safari, and the UI does need to support Safari.

The good news is, you commented that in so I noticed, hooray for comments! Bad news is, it seems like we will have to either fix this up to use a regexp/approach that does support Safari, or pull the previous PR back out.

Let me know if you are able to change this or have any other info/suggestions. If you could let me know whenever as soon as you get a second that'd be great.

Thanks,

radiantly:
Hmmm, I see.

Without a lookbehind, I can't think of a way we can use a single capturing group to detect both template types, so we'll have to switch to using two capturing groups in the regex. Will need to look into `createRegExpEncoder` to see if that is supported though.

johncowen:
Well let me know if I can help with anything there. I just took a quick look to refresh my mind, and I'm sure you'll see, it looks like it just ends up in a little function which looks to pretty much be a `string.replace` - I think the only other place we use it is via the `uri` helper, so if you think changing the signature of that `createRegExpEncoder ` function will help, then it'll have to happen in `uri` also.

Anyway, let me know if I can help with anything, or if you don't have time or whatever lemme know and I can try dig in instead.

(btw if you want to close this PR and start a fresh thats fine also)

johncowen:
I'm gonna close up this PR. I guess it's going to grow more and more stale. Feel free to open a fresh one with a different approach 🙇

PR #11415, opened 2021-10-25
acl: remove Policy.ID and Policy.Revision

dnephin:
From what I can tell these fields are not used. They are never accessed directly, and I can't see anywhere that they would be used as a map key or serialized/deserialized. This PR removes these 2 unused fields.

We do use the `structs.ACLPolicy.ID` field for the `ACLResolver` cache, but not the `acl.Policy.ID`.

### DIFF
--- a/acl/acl_test.go
+++ b/acl/acl_test.go
@@ -2549,7 +2549,7 @@ func TestACL_ReadAll(t *testing.T) {
 	body := func(t *testing.T, rules string, defaultPolicy Authorizer, check func(t *testing.T, authz Authorizer, prefix string, entCtx *AuthorizerContext)) {
 		t.Helper()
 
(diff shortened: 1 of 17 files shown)


FFMMM (approved):
This LGTM!  🏎️ 

I took some time to look over how we manipulate policies and it seems we don't store any part of that. So there's no need for "migrations" here which makes this change safe!

One more caveat before merging is that I'd appreciate a pair of 👀 from @mkeeler or @freddygv

freddygv (approved):
LGTM, couldn't find any usage of it worth keeping

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/495230.

Issue #11372, opened 2021-10-20
Config Entry CAS for delete requests


nickethier:
#### Feature Description

Consul Config Entries support supplying a modify index as the value for a `cas` parameter for update operations which if set only updates the entry if the current modify index matches the value. This enables isolated atomic updates of config entries. However these guarantees cannot be made with delete operations as they don't support a `cas` parameter.

#### Use Case(s)

The Consul API Gateway will update upstream service intentions to add and remove the gateway as a source when corresponding routes are created or deleted. In the deleted case, we need to remove the gateway from the intention's list of sources. If it is the only source then the intention should be deleted. However, we want to be sure the intention was not modified (ex. an additional source added) between the time it was read and deleted.

boxofrad:
Hey @nickethier 👋🏻 Just a heads up that this feature has just been merged 🎉

PR #11419, opened 2021-10-26
Support Check-And-Set deletion of config entries

boxofrad:
Adds support for Check-And-Set (CAS) deletion of config entries, similar to that [of KVs](https://www.consul.io/api-docs/kv#cas-1).

This is useful for cases where it is desirable to delete a config entry, but only if it has not been modified by another process since reading it (such as the scenario described in #11372).

Sidenote: this is my first PR to Consul 🎉 so any meta/style feedback is greatly appreciated!

### DIFF
--- a/.changelog/11419.txt
+++ b/.changelog/11419.txt
@@ -1,0 +1,6 @@
+```release-note:improvement
+config: Support Check-And-Set (CAS) deletion of config entries
+```
+```release-note:improvement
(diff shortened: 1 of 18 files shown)


hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/490134.

PR #11428, opened 2021-10-26
add root_cert_ttl option for consul connect, vault ca providers

FFMMM:
#### Overview

This change adds an option to the `CommonCA` config to allow a `root_cert_ttl` value which will control the TTL (time to live) for root certs issued by supported CA providers. It's mostly plumbing.

#### PR Notes/ Callouts
* At present the only providers supported are `consul connect `and `vault`. 
* Let's focus on the functionality (going QSF -- quality, security, feature), before delving into all the refactoring opportunities
* No corresponding "cli flag" option added as it's not a pattern for our current ca config

#### Config checklist
I used the config file checklist [here](https://github.com/hashicorp/consul/blob/main/docs/config/checklist-adding-config-fields.md). Click below to expand on a "filled in" version of it.

<details>
  <summary>Click to expand!</summary>
  
  #### Legend
  * `n/a` -- not applicable
  * `?` -- I believe I may have fully completed it
  
  #### Adding a Simple Config Field for Client Agents

  `agent/config/config.go`.
  `agent/config/runtime.go`.
  translate.
- [?] **If** your new config field needed some validation as it's only valid in
    - [?] Add validation to Validate in `agent/config/builder.go`.
    - [?] Add a test case to the table test `TestLoad_IntegrationWithFlags` in
- [?] **If** your config should take effect on a reload/HUP.
    - [x] Add necessary code to trigger a safe (locked or atomic) update to
      more of the following places:
    - [x] Add a test to `agent/agent_test.go` similar to others with prefix
      `TestAgent_reloadConfig*`.
- [?] Add documentation to `website/content/docs/agent/options.mdx`.

#### Adding a Simple Config Field for Servers

- [?] Do all of the steps in [Adding a Simple Config
- [?] Add code to set the values from the `RuntimeConfig` in the confusingly
  named `consulConfig` method in `agent/agent.go`
- [n/a] **If needed**, add a test to `agent_test.go` if there is some non-trivial
- [n/a] **If** your config should take effect on a reload/HUP
    - [n/a] Add necessary code to `ReloadConfig` in `agent/consul/server.go` this
      updated.
      - [n/a] Add a new test or a new assertion to `TestServer_ReloadConfig`

</details>

#### Issues related
* https://github.com/hashicorp/consul/issues/10762

#### TODO:
* [x] one off testing with https://github.com/dhiaayachi/consul-local-cluster

<details>
  <summary>Steps to expand!</summary>
  
  1. `make linux`
  2.  Use the binary with https://github.com/dhiaayachi/consul-local-cluster 
  3.  Add a `RootCertTTL` value to  the `vault-provider.json.template` file locally to something like:
  ```
  {
    "Provider": "vault",
    "Config": {
        "LeafCertTTL": "72h",
        "Address": "http://myvault:8200",
        "Token": {{VAULT_TOKEN}},
        "RootPKIPath": "connect-root",
        "IntermediatePKIPath": "connect-intermediate",
        "RootCertTTL": "8761h"
    },
    "ForceWithoutCrossSigning": false
}
  ```
  4. Assert root cert properties are as expected -- `consul connect ca get-config -token=<TOKEN>`
</details>

* [x] awspca --> https://github.com/hashicorp/consul/pull/11449
* [x] gofmt
* [x] go mod
* [x] docs
* [x] changelog
 

### DIFF
--- a/.changelog/11428.txt
+++ b/.changelog/11428.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+ca: Add a configurable ttl for root certs for connect, vault ca providers.
+```
--- a/.changelog/11428.txt
+++ b/.changelog/11428.txt
@@ -1,3 +1,3 @@
 ```release-note:feature
-ca: Add a configurable ttl for root certs for connect, vault ca providers.
+ca: Add a configurable TTL for Connect CA root certificates. The configuration is supported by the Vault and Consul providers.
 ```
--- a/agent/config/runtime_test.go
+++ b/agent/config/runtime_test.go
@@ -3144,7 +3144,7 @@ func TestLoad_IntegrationWithFlags(t *testing.T) {
 						"tls_skip_verify": true,
 						"token": "abc",
 						"root_pki_path": "consul-vault",
-						"root_cert_ttl": "87600h",
+						"root_cert_ttl": "96360h",
 						"intermediate_pki_path": "connect-intermediate"
 					}
 				}
@@ -3163,7 +3163,7 @@ func TestLoad_IntegrationWithFlags(t *testing.T) {
 						root_pki_path = "consul-vault"
 						token = "abc"
 						intermediate_pki_path = "connect-intermediate"
-						root_cert_ttl = "87600h"
+						root_cert_ttl = "96360h"
 					}
 				}
 			`},
@@ -3180,7 +3180,7 @@ func TestLoad_IntegrationWithFlags(t *testing.T) {
 				"TLSSkipVerify":       true,
 				"Token":               "abc",
 				"RootPKIPath":         "consul-vault",
-				"RootCertTTL":         "87600h",
+				"RootCertTTL":         "96360h",
 				"IntermediatePKIPath": "connect-intermediate",
 			}
 		},
@@ -5492,7 +5492,7 @@ func TestLoad_FullConfig(t *testing.T) {
 		ConnectCAConfig: map[string]interface{}{
 			"IntermediateCertTTL": "8760h",
 			"LeafCertTTL":         "1h",
-			"RootCertTTL":         "87600h",
+			"RootCertTTL":         "96360h",
 			"CSRMaxPerSecond":     float64(100),
 			"CSRMaxConcurrent":    float64(2),
 		},
@@ -6718,7 +6718,7 @@ func TestConnectCAConfiguration(t *testing.T) {
 				ConnectEnabled: true,
 				ConnectCAConfig: map[string]interface{}{
 					"foo":         "bar",
-					"RootCertTTL": "8761h", // 365 * 10 * 24h
+					"RootCertTTL": "8761h", // 365 * 24h + 1
 				},
 			},
 			expected: &structs.CAConfiguration{
@@ -6726,7 +6726,7 @@ func TestConnectCAConfiguration(t *testing.T) {
 				Config: map[string]interface{}{
 					"LeafCertTTL":         "72h",
 					"IntermediateCertTTL": "8760h", // 365 * 24h
-					"RootCertTTL":         "8761h", // 365 * 10 * 24h
+					"RootCertTTL":         "8761h", // 365 * 24h + 1
 					"foo":                 "bar",
 				},
 			},
--- a/agent/config/runtime_test.go
+++ b/agent/config/runtime_test.go
@@ -3127,6 +3127,65 @@ func TestLoad_IntegrationWithFlags(t *testing.T) {
 		},
 	})
 	run(t, testCase{
+		desc: "test connect vault provider configuration with root cert ttl",
+		args: []string{
+			`-data-dir=` + dataDir,
+		},
+		json: []string{`{
+				"connect": {
+					"enabled": true,
+					"ca_provider": "vault",
+					"ca_config": {
+						"ca_file": "/capath/ca.pem",
+						"ca_path": "/capath/",
+						"cert_file": "/certpath/cert.pem",
+						"key_file": "/certpath/key.pem",
+						"tls_server_name": "server.name",
+						"tls_skip_verify": true,
+						"token": "abc",
+						"root_pki_path": "consul-vault",
+						"root_cert_ttl": "87600h",
+						"intermediate_pki_path": "connect-intermediate"
+					}
+				}
+			}`},
+		hcl: []string{`
+			  connect {
+					enabled = true
+					ca_provider = "vault"
+					ca_config {
+						ca_file = "/capath/ca.pem"
+						ca_path = "/capath/"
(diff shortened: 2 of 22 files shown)


agent/structs/connect_ca.go
```diff
@@ -380,6 +382,17 @@ func (c CommonCAProviderConfig) Validate() error {
 		return nil
 	}
 
+	// if the root cert ttl is not set, set it to the default
+	if c.RootCertTTL == 0 {
+		c.RootCertTTL = 10 * 365 * 24 * time.Hour
```
kyhavlov, line 387: I think we can do `c.RootCertTTL = time.ParseDuration(DefaultRootCertTTL)` here and use the default that's already defined above

website/content/docs/agent/options.mdx
```diff
@@ -1267,6 +1267,11 @@ bind_addr = "{{ GetPrivateInterfaces | include \"network\" \"10.0.0.0/8\" | attr
       for more than twice the _current_ `leaf_cert_ttl`, it will be removed
       from the trusted list.
 
+    - `root_cert_ttl` ((#todo)) todo. Defaults to 10 years as `87600h`.
+      todo restrictions.
```
kyhavlov, line 1271: Still needs a description (leftover todo)

kyhavlov (commented):
Looking good, left a couple comments but nothing blocking - once the doc change is done this should be good to merge 👍

.changelog/11428.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+ca: Add a configurable ttl for root certs for connect, vault ca providers.
```
dnephin, line 2:
```suggestion
ca: Add a configurable TTL for Connect CA root certificates. The configuration is supported by the Vault and Consul providers.
```

agent/config/runtime_test.go
```diff
@@ -3126,6 +3126,65 @@ func TestLoad_IntegrationWithFlags(t *testing.T) {
...
+			rt.ConnectCAConfig = map[string]interface{}{
+				"CAFile":              "/capath/ca.pem",
+				"CAPath":              "/capath/",
+				"CertFile":            "/certpath/cert.pem",
+				"KeyFile":             "/certpath/key.pem",
+				"TLSServerName":       "server.name",
+				"TLSSkipVerify":       true,
+				"Token":               "abc",
+				"RootPKIPath":         "consul-vault",
+				"RootCertTTL":         "87600h",
```
dnephin, line 3183: I think we should use a non-default value in the tests, so that we notice when the field is not settable from the config. Right now the test could pass if all we accepted as the default.

agent/config/testdata/full-config.hcl
```diff
@@ -197,6 +197,7 @@ connect {
     ca_config {
         intermediate_cert_ttl = "8760h"
         leaf_cert_ttl = "1h"
+        root_cert_ttl = "87600h"
```
dnephin, line 200: same here, would be good to not use the default value. (also below in the json)

agent/connect/ca/provider_vault.go
```diff
@@ -170,7 +170,11 @@ func (v *VaultProvider) GenerateRoot() error {
 			Type:        "pki",
 			Description: "root CA backend for Consul Connect",
 			Config: vaultapi.MountConfigInput{
-				MaxLeaseTTL: "8760h",
+				// the max lease ttl denotes the maximum ttl that secrets are created from the engine
+				// the default lease ttl is the kind of ttl that will *reliably* set the ttl to v.config.RootCertTTL
+				// https://www.vaultproject.io/docs/secrets/pki#configure-a-ca-certificate
+				MaxLeaseTTL:     v.config.RootCertTTL.String(),
+				DefaultLeaseTTL: v.config.RootCertTTL.String(),
```
dnephin, line 177: :raised_hands:

agent/structs/connect_ca.go
dnephin: This function is `Validate`, but here we are normalizing (not validating).

I suspect this should this return an error instead?  In normal operations we expect `RuntimeConfig` to always have set a default, so users should only encounter this problem if they explicit set a 0 for TTL (I think).

website/content/docs/agent/options.mdx
dnephin: By restrictions I guess you mean "is only used when the CA system is first initialized, or when a configuration change causes a rotation or the root certificate" ? I think we should say something about that.

dnephin (commented):
Nice work! I see you found all the places where we plumb this config.

Mostly just small suggestions about testing values and docs.

agent/structs/connect_ca.go
FFMMM: > but here we are normalizing 
>  In normal operations we expect RuntimeConfig to always have set a default, 

Hey both, thanks for the engagement here -- I think per Daniel's comment, I'm inclined to take out this chunk of code.

If a callee actually sets the value to less than the intermediate cert ttl than the right error should come out.

I also see no such "normalization"/ defaulting for leaf and intermediate ttls

website/content/docs/agent/options.mdx
FFMMM: I was punting on the docs a bit to make sure that the stuff here was something we eventually wanted to check in.

Which looks like it is case. So I added the docs now :)

blake: @FFMMM Can you also make sure this field gets documented under https://github.com/hashicorp/consul/blob/main/website/content/partials/http_api_connect_ca_common_options.mdx?

The options in this file are displayed under the provider-specific CA config options. For example, https://www.consul.io/docs/connect/ca/vault#common-ca-config-options.

website/content/docs/agent/options.mdx
```diff
@@ -1267,6 +1267,18 @@ bind_addr = "{{ GetPrivateInterfaces | include \"network\" \"10.0.0.0/8\" | attr
...
+      Defaults to 10 years as `87600h`. This value, if provided, needs to be higher than the
+      intermediate certificate TTL.
+
+      This setting currently applies only to the consul connect and Vault CA providers. It is
+      ignored for the AWS acm pca provider. The value for root certificates issued by the AWS
+      CA provider is 5 years and not configurable at this time.
+
+      For the Vault provider, this value is only used if the backend is not initialized at first.
+
+      This value is also applied on ca set-config command.
```
kyhavlov, line 1280:
```suggestion
      This value is also applied on the `ca set-config` command.
```

kyhavlov (approved):
Looks good to merge, just had one tiny suggestion 👍

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/491685.

Issue #11396, opened 2021-10-21
[refactor] Test endpoints through ServeHTTP, not by raw handler


kisunji:
### Background
In many places in our tests, when we want to make API calls to handlers, we tend to use the raw handler (e.g. `a.srv.AgentToken(resp, req)`).

This is problematic because it skips the middleware that real HTTP requests go through, which includes (but is not limited to) multiplexing based on URL patterns, blocking not-allowed HTTP methods, and translating application-level errors to HTTP error codes and messages.

This means that developer errors like this may not get caught by the test suite:
```go
// http_register.go
// registerEndpoint("/v1/agent/join/", []string{"PUT"}, (*HTTPHandlers).AgentJoin)

req, _ := http.NewRequest("GET", "/v1/agent/node", nil) // wrong method and endpoint

resp := httptest.NewRecorder()
_, err := a.srv.AgentJoin(resp, req) // does not immediately reject the request
```

`AgentJoin` is not responsible for validating the request's URL or method and may fail in unpredictable ways, wasting developer time or misleading devs to make incorrect test assertions.

### Solution

The proposed solution is to start using `Agent.srv.h.ServeHTTP(resp, req)` to handle requests generically instead of specifying a typed handler.

See examples [here](https://github.com/hashicorp/consul/blob/main/agent/acl_endpoint_legacy_test.go#L26-L32).

### Related PRs

https://github.com/hashicorp/consul/pull/11445 (thanks @Mathew-Estafanous)

PR #11445, opened 2021-10-28
Convert (some) test endpoints to use ServeHTTP instead of direct calls to handlers.

Mathew-Estafanous:
Issue #11396

**Overview:**
Initiate the process of converting direct calls to HTTPHandler(s) to calls through ``ServeHTTP(resp, req)`` when testing. There are quite a lot of instances so as requested I just made a handful of changes to keep it small.

**What I Did:**
As mentioned, I converted a few of the tests in ``agent_endpoint_test`` to go through the proper channel. However, as discussed in the [issue comments](https://github.com/hashicorp/consul/issues/11396#issuecomment-953419523) there are many cases where the tests are checking for specific internal objects. As a result I did as recommended and converted the JSON responses back into the object representations to proceed with the tests.


Mathew-Estafanous:
How do I request a code review for this PR? To my knowledge, I don’t seem to have the ability to do it myself.

Mathew-Estafanous:
Ok @kisunji , I’m ready for a review.

agent/agent_endpoint_test.go
```diff
@@ -5397,13 +5417,9 @@ func TestAgent_Token(t *testing.T) {
...
 
-			_, err := a.srv.AgentToken(resp, req)
-			if tt.expectedErr != "" {
-				require.EqualError(t, err, tt.expectedErr)
-				return
-			}
-			require.NoError(t, err)
+			a.srv.h.ServeHTTP(resp, req)
 			require.Equal(t, tt.code, resp.Code)
+			require.Contains(t, resp.Body.String(), tt.expectedErr)
```
kisunji, line 5422: Could you wrap this in an `if tt.expectedErr != ""` block? We want to compare only if it's not blank and return early (so we don't compare the fields below).

kisunji (commented):
Looks good! Just one minor change

agent/agent_endpoint_test.go
Mathew-Estafanous: On it! I just pushed the change.

kisunji:
Thanks @Mathew-Estafanous! I'll merge it now. Feel free to submit pull requests for other tests (or look around at other issues that interest you)

kisunji (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/492140.

Mathew-Estafanous:
Thanks so much @kisunji ! I’m considering working on most of the test cases within ``agent_endpoint_test.go`` however I realize that it might turn out to be quite a large change. Would it be ok if I submit all those changes as one PR or would it be more appropriate to split them out into small PRs (similar to this size)?

kisunji:
> Thanks so much @kisunji ! I’m considering working on most of the test cases within `agent_endpoint_test.go` however I realize that it might turn out to be quite a large change. Would it be ok if I submit all those changes as one PR or would it be more appropriate to split them out into small PRs (similar to this size)?

Given your experience now I think one file = one PR is good!

PR #11450, opened 2021-10-29
Ensure calls to BestAddress consider partition

freddygv:
When checking which endpoint to use for an upstream we call BestAddress on the service instance. This function will optionally look for the tagged WAN address if it's a WAN request.

Originally we would only check whether the datacenters match, but now are also checking whether partitions match.

The last commit also fixes an issue with how the `GatewayKeys()` method wasn't deduplicating data.


rboyer (approved):
LGTM

rboyer (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/490953.

Issue #11448, opened 2021-10-28
Connect leaf certs issued from Vault CA don't contain full chain


krarey:
#### Overview of the Issue

When issuing a leaf certificate for a service identity or auto-encrypt agent from a Vault CA, Consul builds the following certificate chain:
```
leaf cert +
issuing CA
```
https://github.com/hashicorp/consul/blob/977be7749397d4fb81eb173f54608f1bd7caf633/agent/connect/ca/provider_vault.go#L441-L462

In environments with more than one subordinate CA (i.e. the issuing intermediate was, itself, issued from another intermediate), this breaks the certificate chain, unless the second-tier intermediate is explicitly trusted on the host.

In stead of using the returned `issuing_ca` key, it would be ideal for Consul to use the `ca_chain` key from Vault's response, which contains a list of every CA certificate leading back to the root.

#### Reproduction Steps

Steps to reproduce this issue, eg:

1. Set up a Vault environment with three PKI mounts:
1.1. Root CA
1.2 Intermediate CA A (Issuer = Root CA)
1.3. Intermediate CA B (Un-configured - used as the Consul Issuing CA)
2. Configure Consul to use Vault as its Connect PKI source.
```json
{
	"Provider": "vault",
	"Config": {
		"Address": "<Vault Addr>",
		"IntermediateCertTTL": "8760h",
		"IntermediatePKIPath": "<Intermediate CA B Mount>",
		"LeafCertTTL": "72h",
		"PrivateKeyBits": "2048",
		"PrivateKeyType": "rsa",
		"RootPKIPath": "<Intermediate CA A Mount>",
		"RotationPeriod": "2160h",
		"Token": "<...>"
	},
}
```
3. Check the cert chain returned for an issued mTLS or auto_encrypt agent certificate.

rdxmb:
I have the problem that the consul-server-cert does not contain the the Root CA, but only the intermediate CA. Should this be fixed? 

crosslink: https://github.com/hashicorp/consul/issues/12386#issuecomment-1792528206

PR #11451, opened 2021-10-29
Update Vault CA provider to apply full cert chain

krarey:
Modifies the Vault CA provider for Consul Connect to apply the complete CA chain to generated certificates. More information in issue #11448.

### DIFF
--- a/agent/connect/ca/provider_vault.go
+++ b/agent/connect/ca/provider_vault.go
@@ -450,7 +450,7 @@ func (v *VaultProvider) Sign(csr *x509.CertificateRequest) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("error issuing cert: %v", err)
 	}
-	if response == nil || response.Data["certificate"] == "" || response.Data["issuing_ca"] == "" {
+	if response == nil || response.Data["certificate"] == "" {
 		return "", fmt.Errorf("certificate info returned from Vault was blank")
 	}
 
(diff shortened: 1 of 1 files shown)


FFMMM:
Thanks for putting this out @krarey -- it overall LGTM! Would you mind adding some specific tests that tests for a cert chain presence?

---

Also, to err on the side of over caution, do we think such a change could break any practitioner workflows? As in, say folks are looking for just one cert and now receive more than one?

krarey:
Thanks @FFMMM!

>Would you mind adding some specific tests that tests for a cert chain presence?

Would you mind elaborating on what you have in mind? Are you thinking a test that actually parses the returned certificate chain and makes sure each issuer cert is provided in the expected order? If so, that may make more sense at a higher level, when the respective Sign() call is made against any of the Built-in/Vault/ACM engines.

>Also, to err on the side of over caution, do we think such a change could break any practitioner workflows? As in, say folks are looking for just one cert and now receive more than one?

I don't think this should have any impact – the existing code was already attaching the issuer CA, and in most cases (root + consul-managed intermediate), this patch will result in an identical chain. In this case we are just ensuring any additional subordinate CAs are also added, should they be available and properly configured on the issuing Vault intermediate (if the subordinate CAs haven't been attached to the intermediate ca_chain in Vault, the result would again be the same as before, `leaf cert + issuing CA`). That said, I haven't worked extensively on TLS code prior, so would absolutely invite another opinion!

dnephin:
I noticed CI has not run (`verify-ci` is "waiting for status"), which is usually a github permissions problem. We'll need to get that sorted out.

krarey:
Currently held while we work through changes to the secondary DC intermediate signing flow.

dnephin:
This ended up requiring a few bug fixes, and a few more changes to the vault provider. Existing releases of vault have this bug https://github.com/hashicorp/vault/issues/13489, which meant using `ca_chain` directly was not always possible. Those changes can be seen in #11910.

Thanks for getting us started and pointed in the right direction @krarey!  I'm going to close this PR since it is replaced by #11910.

PR #11452, opened 2021-10-29
Crt release/1.9.x

99:


.github/workflows/build.yml
```diff
@@ -0,0 +1,250 @@
+name: build
+
+on:
+  push:
+    # Sequence of patterns matched against refs/heads
+    branches:
+    # Provide the release branch
+      - crt_release/1.9.x
```
freddygv, line 8: Should this be `release/1.9.x` so that it matches our release branch name?

freddygv (approved):
LGTM, left one non-blocking question

.github/workflows/build.yml
```diff
@@ -0,0 +1,250 @@
...
+          name: metadata.json
+          path: ${{ steps.generate-metadata-file.outputs.filepath }}
+
+  build:
+    needs: get-product-version
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - {go: "1.17", goos: "linux", goarch: "386"}
```
freddygv, line 66: This should use go 1.15

.github/workflows/build.yml
```diff
@@ -0,0 +1,250 @@
...
+
+      - name: Install project dependencies
+        run: |
+          go install github.com/elazarl/go-bindata-assetfs/go-bindata-assetfs@38087fe
+          go install github.com/hashicorp/go-bindata/go-bindata@bf7910a
+
+      - name: Setup with node and yarn
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
```
freddygv, line 94: Previously this release series used node 12

freddygv (changes requested):

.github/workflows/build.yml
99: updated

.github/workflows/build.yml
99: updated

.github/workflows/build.yml
99: Yes updated to `release/1.9.x` (kept this to trigger the current branch releases to validate that all the updates are passing)

.github/workflows/build.yml
```diff
@@ -0,0 +1,251 @@
...
+
+      - name: Build
+        env:
+          GOOS: ${{ matrix.goos }}
+          GOARCH: ${{ matrix.goarch }}
+          CGO_ENABLED: 0
+          GOLDFLAGS: "${{needs.get-product-version.outputs.shared-ldflags}}"
+        run: |
+          mkdir dist out
+          go build -ldflags="$GOLDFLAGS" -o dist/ .
```
freddygv, line 121: Should `[ "${{ matrix.goarch }}" == "arm" ] && export GOARM=5` be added here?

99: thank you for catching my copypasta

freddygv (approved):

PR #11454, opened 2021-10-29
Crt release/1.10.x

99:


.github/workflows/build.yml
```diff
@@ -0,0 +1,250 @@
+name: build
+
+on:
+  push:
+    # Sequence of patterns matched against refs/heads
+    branches:
+    # Provide the release branch
+      - crt_release/1.10.x
```
freddygv, line 8: Should this be `release/1.10.x` so that it matches our release branch name?

freddygv (approved):
LGTM, left one non-blocking question

.github/workflows/build.yml
```diff
@@ -0,0 +1,250 @@
...
+          name: metadata.json
+          path: ${{ steps.generate-metadata-file.outputs.filepath }}
+
+  build:
+    needs: get-product-version
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - {go: "1.17", goos: "linux", goarch: "386"}
```
freddygv, line 66: This should use go 1.16

freddygv (changes requested):

.github/workflows/build.yml
99: updated

.github/workflows/build.yml
99: updated

freddygv (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/498127.

PR #11391, opened 2021-10-21
Add changelog creation to contributor docs

jkirschner-hashicorp:
Adds guidance on:
- submitting a pull request
- adding a changelog entry

Reorganizes the content in ways that I think improve clarity.

You can preview the file here: 
https://github.com/hashicorp/consul/blob/45acc91c26f9fe1b9cddb489a93bf5551c8637e0/.github/CONTRIBUTING.md

Issue #5371, opened 2019-02-22
Consul will accept an empty client_addr and not report a failure to create client services


karl-tpio:
#### Overview of the Issue

no warn/error emitted in logs when a `client_addr` directive is empty. consul continues to start up, but does not create a listening socket. Client services (dns/http) are unavailable.

#### Reproduction Steps

Currently, i have a 3 node cluster in a VPC. Almost bog-standard adherence to the [standard install guide](https://learn.hashicorp.com/consul/advanced/day-1-operations/deployment-guide#download-consul).

Each of the three nodes is identical, as they were all spun from the same AMI.

Here is how they are configured:

```
root@ip-172-25-26-150:/etc/consul.d# cat *.hcl
##
# Consul's defaults are pretty sane, a few things we must change:
#
# See: https: //www.consul.io/docs/agent/options.html
##

retry_join = [
  "provider=aws tag_key=consul.cluster tag_value=testing",
]

datacenter = "us-west-1"

encrypt = "<lol.nope>"

performance {
  raft_multiplier = 1
}

bind_addr = "{{ GetPrivateIP }}"

client_addr = "{{ GetInterfaceIP \"dummy0\" }}"

# TODO: this still needs to be tested, but i believe that systemd resolve conf requires we listen on 53
ports = {
  dns = 53
}

server = true

bootstrap_expect = 3

ui = true
```

I am expecting to see consul listening on port 53, on the IP address of the `dummy0` interface.

E.G.:

1. Consul is up, and talking with peers:

```
root@ip-172-25-26-150:/etc/consul.d# service consul status
● consul.service - "HashiCorp Consul - A service mesh solution"
   Loaded: loaded (/etc/systemd/system/consul.service; enabled; vendor preset: enabled)
   Active: active (running) since Fri 2019-02-22 02:04:35 UTC; 1min 38s ago
     Docs: https://www.consul.io/
 Main PID: 3294 (consul)
    Tasks: 11 (limit: 2320)
   CGroup: /system.slice/consul.service
           └─3294 /usr/local/bin/consul agent -config-dir=/etc/consul.d/

Feb 22 02:04:35 ip-172-25-26-150 consul[3294]: 2019/02/22 02:04:35 [DEBUG] raft-net: 172.25.26.150:8300 accepted connection fro
Feb 22 02:04:35 ip-172-25-26-150 consul[3294]: 2019/02/22 02:04:35 [DEBUG] raft-net: 172.25.26.150:8300 accepted connection fro
[...]

PR #11461, opened 2021-11-01
config: warn the user if client_addr is empty

deblasis:
if the provided value is an empty string then the client services
(DNS, HTTP, HTTPS, GRPC) are not listening and the user is not notified
in any way about what's happening.
Also, since a not provided client_addr defaults to 127.0.0.1, we make sure
we are not getting unwanted warnings

Signed-off-by: Alessandro De Blasis <alex@deblasis.net>

### DIFF
--- a/.changelog/11461.txt
+++ b/.changelog/11461.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+config: warn the user if client_addr is empty because client services won't be listening
+```
--- a/agent/config/builder.go
+++ b/agent/config/builder.go
@@ -552,6 +552,9 @@ func (b *builder) build() (rt RuntimeConfig, err error) {
 
 	// determine client addresses
 	clientAddrs := b.expandIPs("client_addr", c.ClientAddr)
+	if len(clientAddrs) == 0 {
+		b.warn("client_addr is empty, client services (DNS, HTTP, HTTPS, GRPC) will not be listening for connections")
+	}
 	dnsAddrs := b.makeAddrs(b.expandAddrs("addresses.dns", c.Addresses.DNS), clientAddrs, dnsPort)
 	httpAddrs := b.makeAddrs(b.expandAddrs("addresses.http", c.Addresses.HTTP), clientAddrs, httpPort)
 	httpsAddrs := b.makeAddrs(b.expandAddrs("addresses.https", c.Addresses.HTTPS), clientAddrs, httpsPort)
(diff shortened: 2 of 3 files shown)


jkirschner-hashicorp:
Linked to #5371. As a part of the review, we should determine whether this partially or fully addresses 5371.

.changelog/11446.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
```
jkirschner-hashicorp, line 1: Thanks for adding the changelog!

Something to address later (once you get a review back from engineering): this changelog file should match the PR number (`.changelog/11461`). This currently overwrites a different (existing) changelog entry.

deblasis: Ops! I wanted to be compliant with the style/process and I think I overdid it! 
OK, I am gonna remove it for now then and I'll wait for further instructions. 
Thanks @jkirschner-hashicorp!

jkirschner-hashicorp: You did the right thing! You just need to change the filename :) Should be `.changelog/11461.txt`, not `.changelog/11446.txt`.

jkirschner-hashicorp: I currently have a PR open to update the contributing docs to make the changelog entry creation process clearer: https://github.com/hashicorp/consul/pull/11391

I hope to get it merged soon!

deblasis: Mhh, not sure if `config` is the right code area 🤔, it's not a new config option even if the change affects config options. Thoughts?

FFMMM:
hey @deblasis thanks for putting this out!

Chiming in to point out that the recent test failures are part of an expired test cert. We actually "fixed" this in the upstream.

Would you mind rebasing on `main`?

.changelog/11446.txt
jkirschner-hashicorp: I think type `feature` and code area `config` would be used for a new config option. Type `improvement` area `config` feels appropriate here to me!

deblasis: Awesome, thanks @jkirschner-hashicorp

deblasis:
Hi @FFMMM! No worries, my pleasure. 

I use consul everywhere, the minimum I can do is contribute a little! 😃 
I rebased but in the meantime, I triggered a build too many... sorry about that! It should be ok now.

Anything else I can do please let me know

kyhavlov (approved):
Looks good to merge - thanks for the PR, @deblasis !

In the future we could consider erroring outright in this case in order to make this more visible to the user. The only method we officially document for disabling an interface is by setting -1 for the port - passing an empty address like this is less explicit and could potentially happen by accident with config management, so it might be better not to allow it and make sure the user is intentionally disabling the interface via `port = -1`.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/497846.

PR #11472, opened 2021-11-02
ui: Ensure the tokens default nspace is passed thru to the auth endpoint

johncowen:
Most HTTP API calls will use the default namespace of the calling token to additionally filter/select the data used for the response if one is not specified by the frontend.

The internal permissions/authorize endpoint does not do this (you can ask for permissions from different namespaces in on request).

Therefore this PR adds the tokens default namespace in the frontend only to our calls to the authorize endpoint. I tried to do it in a place that made it feel like it's getting added in the backend, i.e. in a place which was least likely to ever require changing or thinking about.

Note: 1.10.x base branch, a 1.11 PR to come that will add the same fix but for partitions also, I'm guessing it works better to fix on 1.10 and then add the 1.11 functionality, rather than fix on 1.11 and then remove the functionality not required on 1.10. Added a changelog and no-changelog label here to get around the changelog checker.

P.S. Quick note here. After speaking to @jkirschner-hashicorp and @dnephin we are probably going to change this internal endpoint to also inspect the tokens default namespace on the backend. At which point we can revert this commit/PR. Revert will need to happen on 1.10 and 1.11

### DIFF
--- a/.changelog/11472.txt
+++ b/.changelog/11472.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: **(Enterprise only)** When no namespace is selected, make sure to default to the tokens default namespace when requesting permissions
+```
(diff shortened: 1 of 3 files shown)


kaxcode (approved):
:shipit:

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/493648.

PR #11477, opened 2021-11-02
Support Vault Namespaces explicitly in CA config

clly:
If there is a Namespace entry included in the Vault CA configuration,
set it as the Vault Namespace on the Vault client

Currently the only way to support Vault namespaces in the Consul CA
config is by doing one of the following:
1) Set the VAULT_NAMESPACE environment variable which will be picked up
by the Vault API client
2) Prefix all Vault paths with the namespace

Neither of these are super pleasant. The first requires direct access
and modification to the Consul runtime environment. It's possible and
expected that the consul owner has that access but it's not super pleasant.

The second requires more indepth knowledge of Vault and how it uses
Namespaces and could be confusing for anyone without that context. It
also infers that it is not supported

### DIFF
--- a/.changelog/11477.txt
+++ b/.changelog/11477.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-connect: add Namespace configuration entry for Vault provider
+connect: add Namespace configuration setting for Vault CA provider
 ```
--- a/.changelog/11477.txt
+++ b/.changelog/11477.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+connect: add Namespace configuration entry for Vault provider
+```
--- a/agent/connect/ca/provider_vault.go
+++ b/agent/connect/ca/provider_vault.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"os"
 	"strings"
 	"time"
 
@@ -75,6 +76,12 @@ func (v *VaultProvider) Configure(cfg ProviderConfig) error {
 	}
 
 	client.SetToken(config.Token)
+
+	// We don't want to set the namespace if it's empty to prevent potential
+	// unknown behavior. It's also what the Vault client does.
+	if config.Namespace != "" {
+		client.SetNamespace(config.Namespace)
+	}
 	v.config = config
 	v.client = client
 	v.isPrimary = cfg.IsPrimary
@@ -582,6 +589,7 @@ func (v *VaultProvider) Stop() {
 func (v *VaultProvider) PrimaryUsesIntermediate() {}
 
 func ParseVaultCAConfig(raw map[string]interface{}) (*structs.VaultCAProviderConfig, error) {
+	fmt.Fprintf(os.Stderr, "%+v\n", raw)
 	config := structs.VaultCAProviderConfig{
 		CommonCAProviderConfig: defaultCommonConfig(),
 	}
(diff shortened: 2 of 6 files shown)


agent/connect/ca/provider_vault.go
```diff
@@ -75,6 +76,12 @@ func (v *VaultProvider) Configure(cfg ProviderConfig) error {
 	}
 
 	client.SetToken(config.Token)
+
+	// We don't want to set the namespace if it's empty to prevent potential
+	// unknown behavior. It's also what the Vault client does.
```
FFMMM, line 81: For my own curiosity? What kind of "unknown" behaviors?

Also, would you mind expanding on "It's also what the Vault client does." ?

agent/connect/ca/provider_vault.go
```diff
@@ -582,6 +589,7 @@ func (v *VaultProvider) Stop() {
 func (v *VaultProvider) PrimaryUsesIntermediate() {}
 
 func ParseVaultCAConfig(raw map[string]interface{}) (*structs.VaultCAProviderConfig, error) {
+	fmt.Fprintf(os.Stderr, "%+v\n", raw)
```
FFMMM, line 592: this may also be a silly question but why this addition and line 611?

agent/connect/ca/provider_vault_test.go
```diff
@@ -36,6 +38,35 @@ func TestVaultCAProvider_VaultTLSConfig(t *testing.T) {
 	require.Equal(config.TLSSkipVerify, tlsConfig.Insecure)
 }
 
+func TestVaultCAProvider_Configure(t *testing.T) {
+	testcases := []struct {
+		name          string
+		rawConfig     map[string]interface{}
+		expectedValue func(t *testing.T, v *VaultProvider)
+	}{
+		{
```
FFMMM, line 47: This may be already covered by previous tests, but how about having a test where the namespace is not set/ present?

agent/connect/ca/provider_vault.go
clly: Sorry, I was using this to test and forgot to remove it

agent/connect/ca/provider_vault.go
clly: I don't know what the behavior is in Vault if a namespace header is set but empty.

The behavior in the Vault API is to make sure the Namespace is empty before setting it: This part is the API Client getting the namespace from the environment: https://github.com/hashicorp/vault/blob/e24037fa8f713f884817e8bec30023e56af0d5cc/api/client.go#L525

This is the behavior in the SSH Agent: https://github.com/hashicorp/vault/blob/e24037fa8f713f884817e8bec30023e56af0d5cc/api/ssh_agent.go#L128

Considering it's validated as a non-empty string in the API code consuming it automatically I figured it'd be better safe than sorry.

agent/connect/ca/provider_vault_test.go
clly: Most of the other tests would hit the code but there's nothing explicitly checking it. I can add a test here.

agent/connect/ca/provider_vault.go
clly: I've changed the comment which I'm hoping will clarify this a bit

FFMMM (approved):
This LGTM! 🚀

.changelog/11477.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+connect: add Namespace configuration entry for Vault provider
```
crhino, line 2: smallest of nits: "configuration entry" is an overloaded term here, since Consul supports API objects named as such. Maybe "configuration setting" or just "configuration"

crhino (approved):
LGTM!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/494815.

PR #11479, opened 2021-11-02
ui: When certain menus are selected reset either nspace/partition

johncowen:
For our dc, nspace and partition 'bucket' menus, sometimes when selecting one 'bucket' we need to reset a different 'bucket' back to the one that your token has by default (or the default if not). For example when switching to a different partition whilst you are in a non-default namespace of another partition, we need to switch you to the token default namespace of the partition you are switching to.

### DIFF
--- a/.changelog/11479.txt
+++ b/.changelog/11479.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: When switching partitions reset the namespace back to the tokens default namespace or default
+```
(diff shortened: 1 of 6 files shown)


kaxcode (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/504876.

PR #11406, opened 2021-10-24
Use "Open dashboard" everywhere

lkysow:
Previously we had "Open metrics Dashboard" and "Configure metrics
dashboard" in the topology cards and then we had "Open Dashboard" in the
top nav when the dashboard was configured.

Now we use "Open dashboard" and "Configure dashboard".

This change was made for consistency in wording and casing. In addition,
the dashboard could be used for metrics but also other dashboards so
there's no need to scope it only to metrics. Also the config is:

```hcl
ui_config {
  dashboard_url_templates
}
```

Which does not mention metrics

Before:
image

PR #11482, opened 2021-11-03
Use "Open dashboard" everywhere (#11406)

lkysow:
Previously we had "Open metrics Dashboard" and "Configure metrics
dashboard" in the topology cards and then we had "Open Dashboard" in the
top nav when the dashboard was configured.

Now we use "Open dashboard" and "Configure dashboard".

This change was made for consistency in wording and casing. In addition,
the dashboard could be used for metrics but also other dashboards so
there's no need to scope it only to metrics. Also the config is:

```hcl
ui_config {
  dashboard_url_templates
}
```

Which does not mention metrics

### DIFF
--- a/ui/packages/consul-ui/app/components/topology-metrics/index.hbs
+++ b/ui/packages/consul-ui/app/components/topology-metrics/index.hbs
@@ -66,9 +66,9 @@
     {{/if}}
     <div class="link">
      {{#if @metricsHref}}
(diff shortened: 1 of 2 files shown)


johncowen (approved):
:shipit:

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/492435.

PR #11490, opened 2021-11-04
ui: [Port] Ensure the tokens default nspace (and partition) is passed thru to the auth endpoint

johncowen:
See https://github.com/hashicorp/consul/pull/11472

This port additionally adds support for the tokens default partition.

Note: Similar to https://github.com/hashicorp/consul/pull/11472 we are probably going to change this internal endpoint to also inspect the tokens default namespace on the backend. At which point we can revert this commit/PR. Revert will need to happen on 1.10 and 1.11

### DIFF
--- a/.changelog/11472.txt
+++ b/.changelog/11472.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: **(Enterprise only)** When no namespace is selected, make sure to default to the tokens default namespace when requesting permissions
+```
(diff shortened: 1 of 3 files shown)


kaxcode (approved):
👍

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/500537.

PR #11462, opened 2021-11-01
regenerate expired certs

dhiaayachi:
This is to fix CI, by generating a new set of certificates that replace the expired client certificates used for test

PR #11498, opened 2021-11-04
(1.8.x) regenerate expired certs

rboyer:
1.8.x backport of #11462


freddygv (approved):

markan (approved):
Any concerns about moving from RSA to EC keys?

rboyer:
> Any concerns about moving from RSA to EC keys?

@markan Not really on this branch. The keys are the same as in the other branches.

PR #10299, opened 2021-05-26
RPC Timeout/Retries account for blocking requests

dhiaayachi:
This PR is to rebase and merge https://github.com/hashicorp/consul/pull/8978

Closes #8978

PR #8978, opened 2020-10-17
RPC Timeout/Retries account for blocking requests

joel0:
I've ported the Nomad fixes from hashicorp/nomad#8921 and hashicorp/nomad#9064.

The only major differences:
* Consul has a configurable MaxQueryTime and DefaultQueryTime, which is a `const` in Nomad.
* agent/consul/rpc.go:ForwardRPC() didn't start the `firstCheck` timer until _after_ the first call.  I did not see any motivation for it to be different than agent/consul/client.go:RPC() and Nomad's client/rpc.go:RPC(), so I moved it to the top of the function.

Issue #6616, opened 2019-10-10
Release state lock before catalog operations in State.SyncChanges


freddygv:
Currently, [`State.SyncChanges`](https://github.com/hashicorp/consul/blob/master/agent/local/state.go#L966) holds the state lock for the entire duration of the function. This means that the time holding the lock is largely dependent on the time to sync or delete services/checks in the catalog. 

When there are network delays this function can wait up to 10s (default RPC dial timeout) for each catalog request.

As shown in [this issue](https://github.com/hashicorp/consul/issues/4460#issuecomment-534476416), that can lead to flapping TTL tests even if the service is healthy. For example:
1. A TTL check is updated to `passing` via the API
2. Network latency leads to long timeouts during `SyncChanges` while holding the state lock
3. The next TTL update fails because it can't acquire the lock and update the check before the TTL interval expires
4. The TTL check is marked as critical

This seems like it could be improved by only holding the state lock while constructing lists of services and checks that are out of sync or were deleted. The catalog operations like `l.syncService` and `l.deleteService` could then be done separately while iterating over those lists.

PumpkinSeed:
I can create a PR for this, but I'm not sure how I can test against. So let's say I do the changes and after if it's working I will figure out a test scenario where we can make sure it won't happen again.

dnephin:
Writing up some thoughts on this issue below based on my current understand.

The logic in `SyncChanges` is:

```
obtain state write lock
if not out of sync:
    sync Node
    mark Node as inSync
for each service and check:
    if item not inSync:
        update the item in the catalog
        mark item inSync
release state write lock
```

The problem is that we hold the write lock while performing requests over the network, which as #8504 shows, can last forever (we don't have read timeouts because of blocking queries and connection multiplexing, although arguably we should, even if they are very long timeouts).

One option for fixing this might be to split up the lock into multiple locks, so that we don't have to hold it while performing  requests over the network.

If we had a separate Sync-lock (to prevent multiple sync operations from running concurrently) we could change the logic to this:

```
obtain sync lock

obtain state read lock
requests = [request for each {node, service, check} that is not inSync]
release state read lock

perform all catalog sync requests

obtain state write lock
for each successful request:
    if the item from the request still matches the value stored in state:
        update the appropriate inSync value to true
release state write lock

release sync lock
```

[...]

PR #11500, opened 2021-11-05
Add timeout to Client RPC calls

### LINKED ISSUE #8504: Hanging RPC caused agent's state lock to be held for hours
#### Overview of the Issue

We had a Consul agent get stuck today such that it wouldn't respond to any HTTP API requests. Going by the [goroutine dump we got from the process](https://gist.github.com/a-robinson/ec5425d98cc8e0bcac9fc6298cf7a9a3), a goroutine was holding the `State` lock while [waiting on `syncCheck`](https://github.com/hashicorp/consul/blob/v1.6.3/agent/local/state.go#L1003), which was [waiting on an RPC to the server](https://github.com/hashicorp/consul/blob/v1.6.3/agent/local/state.go#L1186). This one goroutine holding the lock was causing more than 400 others to block attempting to grab the `RWMutex` in read mode. The stuck goroutine was here, and as you can see had been blocked for more than 7 hours:

```
goroutine 32 [select, 445 minutes]:
github.com/hashicorp/yamux.(*Stream).Read(0xc00071a780, 0xc000a78000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
        /cfsetup_build/consul16/tmp/build/pkg/mod/github.com/hashicorp/yamux@v0.0.0-20181012175058-2f1d1f20f75d/stream.go:133 +0x2a4
bufio.(*Reader).fill(0xc000216c60)
        /usr/local/go/src/bufio/bufio.go:100 +0x103
bufio.(*Reader).ReadByte(0xc000216c60, 0x457, 0xc001496000, 0xc00095fda0)
        /usr/local/go/src/bufio/bufio.go:252 +0x39
github.com/hashicorp/go-msgpack/codec.(*ioDecReader).readn1(0xc0007f00f0, 0x42ea61)
        /cfsetup_build/consul16/tmp/build/pkg/mod/github.com/hashicorp/go-msgpack@v0.5.5/codec/decode.go:90 +0x3d
github.com/hashicorp/go-msgpack/codec.(*msgpackDecDriver).initReadNext(0xc0012f40a0)
        /cfsetup_build/consul16/tmp/build/pkg/mod/github.com/hashicorp/go-msgpack@v0.5.5/codec/msgpack.go:540 +0x43
github.com/hashicorp/go-msgpack/codec.(*Decoder).decode(0xc000216d80, 0x2549960, 0xc0011694d0)
        /cfsetup_build/consul16/tmp/build/pkg/mod/github.com/hashicorp/go-msgpack@v0.5.5/codec/decode.go:635 +0x4c
github.com/hashicorp/go-msgpack/codec.(*Decoder).Decode(0xc000216d80, 0x2549960, 0xc0011694d0, 0x0, 0x0)
        /cfsetup_build/consul16/tmp/build/pkg/mod/github.com/hashicorp/go-msgpack@v0.5.5/codec/decode.go:630 +0x7b
github.com/hashicorp/net-rpc-msgpackrpc.(*MsgpackCodec).read(0xc0015f8100, 0x2549960, 0xc0011694d0, 0xc0015f8101, 0xc0011694d0)
        /cfsetup_build/consul16/tmp/build/pkg/mod/github.com/hashicorp/net-rpc-msgpackrpc@v0.0.0-20151116020338-a14192a58a69/codec.go:121 +0x55
github.com/hashicorp/net-rpc-msgpackrpc.(*MsgpackCodec).ReadResponseHeader(0xc0015f8100, 0xc0011694d0, 0x2987840, 0xc00028c000)
        /cfsetup_build/consul16/tmp/build/pkg/mod/github.com/hashicorp/net-rpc-msgpackrpc@v0.0.0-20151116020338-a14192a58a69/codec.go:74 +0x41
github.com/hashicorp/net-rpc-msgpackrpc.CallWithCodec(0x32687c0, 0xc0015f8100, 0x2d72640, 0x10, 0x2987840, 0xc00028c000, 0x254f720, 0x4af5458, 0xc0012f4100, 0x0)
        /cfsetup_build/consul16/tmp/build/pkg/mod/github.com/hashicorp/net-rpc-msgpackrpc@v0.0.0-20151116020338-a14192a58a69/client.go:29 +0xee
github.com/hashicorp/consul/agent/pool.(*ConnPool).rpc(0xc0002b1c70, 0xc000449c31, 0x5, 0x322f680, 0xc000768540, 0x2, 0x2d72640, 0x10, 0x1, 0x2987840, ...)
        /cfsetup_build/consul16/tmp/build/src/github.com/hashicorp/consul/agent/pool/pool.go:467 +0x1d0
github.com/hashicorp/consul/agent/pool.(*ConnPool).RPC(0xc0002b1c70, 0xc000449c31, 0x5, 0x322f680, 0xc000768540, 0x2, 0x2d72640, 0x10, 0x203001, 0x2987840, ...)
        /cfsetup_build/consul16/tmp/build/src/github.com/hashicorp/consul/agent/pool/pool.go:431 +0xfa
github.com/hashicorp/consul/agent/consul.(*Client).RPC(0xc0002b1ce0, 0x2d72640, 0x10, 0x2987840, 0xc00028c000, 0x254f720, 0x4af5458, 0x0, 0x0)
        /cfsetup_build/consul16/tmp/build/src/github.com/hashicorp/consul/agent/consul/client.go:307 +0x1f1
github.com/hashicorp/consul/agent/local.(*State).syncCheck(0xc0002f41c0, 0xc00049ede0, 0x15, 0xc00108fde8, 0x1)
        /cfsetup_build/consul16/tmp/build/src/github.com/hashicorp/consul/agent/local/state.go:1186 +0x269
github.com/hashicorp/consul/agent/local.(*State).SyncChanges(0xc0002f41c0, 0x0, 0x0)
        /cfsetup_build/consul16/tmp/build/src/github.com/hashicorp/consul/agent/local/state.go:1003 +0x33c
github.com/hashicorp/consul/agent/ae.(*StateSyncer).nextFSMState(0xc00023c980, 0x2d671c6, 0xb, 0x2d671c6, 0xb)
        /cfsetup_build/consul16/tmp/build/src/github.com/hashicorp/consul/agent/ae/ae.go:197 +0x2c8
github.com/hashicorp/consul/agent/ae.(*StateSyncer).runFSM(0xc00023c980, 0x2d60f2a, 0x8, 0xc00108ffb8)
        /cfsetup_build/consul16/tmp/build/src/github.com/hashicorp/consul/agent/ae/ae.go:153 +0x3a
github.com/hashicorp/consul/agent/ae.(*StateSyncer).Run(0xc00023c980)
        /cfsetup_build/consul16/tmp/build/src/github.com/hashicorp/consul/agent/ae/ae.go:147 +0x6e
created by github.com/hashicorp/consul/agent.(*Agent).StartSync
        /cfsetup_build/consul16/tmp/build/src/github.com/hashicorp/consul/agent/agent.go:1858 +0x43
```

As for why the RPC was hanging for so long, I don't know exactly. The consul servers were fine, and other agents in the cluster were behaving normally. The notable thing that happened on the affected agent was memory pressure on the machine that may have caused memory allocation failures. According to our collection of metrics from the agent, the agent had a GC pause of 13.8 seconds around this time. The memory pressure went away, but from this point on the agent was effectively stuck.

#### Reproduction Steps

I don't know how to reproduce the issue. I've only seen it happen once. It's plausible that you could reproduce it with enough memory pressure on a machine, but I don't know the odds of hitting it again.

Both client and server are running v1.6.3.

Debian Stretch

### Log Fragments

Full goroutine dump from sending the process SIGQUIT: https://gist.github.com/a-robinson/ec5425d98cc8e0bcac9fc6298cf7a9a3

This may be related to https://github.com/hashicorp/consul/issues/6616, but the big difference here is that the RPC attempt never timed out.

wjordan:
Fixes #8504. Adds a timeout (deadline) to client RPC calls, so that streams will no longer hang indefinitely in unstable network conditions.
See further discussion of this issue in https://github.com/hashicorp/consul/issues/6616#issuecomment-839265912:

> The problem is that we hold the write lock while performing requests over the network, which as #8504 shows, can last forever (we don't have read timeouts because of blocking queries and connection multiplexing, although arguably we should, even if they are very long timeouts).

This PR reuses the timeout values calculated by the retry-timeout logic in #10299 / #8978, adding an extra `Timeout` function to the `RPCInfo` interface that returns the timeout duration, to supplement the `HasTimedOut` function that returns a boolean.


dnephin:
Thank you for working on this problem! Unfortunately I think this lack of read timeouts is going to be a very challenging thing to solve. I looked to see if we have an issue that describes the problem, but I haven't found one.

I have not yet looked at this PR in much detail, but from a quick glance I think `rpcHoldTimeout` is the wrong value. It is only 7 seconds, but a blocking query needs to stay open for much longer, often 5 to 10 minutes.

There are a number of challenges in fixing this problem because we have many paths for client connections. I believe the current one is Yamux, so we'd need support to timeout a multiplexed connection. There are a few older [PRs open on the yamux repo](https://github.com/hashicorp/yamux/pulls?q=is%3Apr+is%3Aopen+timeout) that aim to add that support, but I'm not sure what state those are in.  Adding something like this will require a lot of testing.

I would still like to take the time to write up a better issue but that will take some more time to research.

wjordan:
@dnephin thanks for taking a look!

> I have not yet looked at this PR in much detail, but from a quick glance I think rpcHoldTimeout is the wrong value. It is only 7 seconds, but a blocking query needs to stay open for much longer, often 5 to 10 minutes.

Yes, the implementation refactors/reuses the logic previously added in #10299 / #8978 which already takes this into account. For blocking queries, the timeout is derived from the provided `MaxQueryTime`: https://github.com/hashicorp/consul/blob/d6e2cbbc73a817c6e7f77af6ab0e91885b7561d2/agent/structs/structs.go#L303-L317

> we'd need support [in Yamux] to timeout a multiplexed connection.

Yamux already supports global timeout of a multiplexed connection through its keepalive-ping feature (which immediately fails all in-flight RPCs going over the connection). More relevant here, Yamux also already supports timing out an individual stream through `setDeadline`: https://github.com/hashicorp/consul/blob/d6e2cbbc73a817c6e7f77af6ab0e91885b7561d2/agent/pool/pool.go#L593-L597

I didn't notice anything else that would be needed for this to work.

> I would still like to take the time to write up a better issue but that will take some more time to research.

If it's helpful, see `TestClient_RPC_Timeout` added by this PR (4f54651c4707b521e25deadda2185fde0ce0f8b8) which provides a concise implementation of the issue that this PR addresses- the test fails without this PR and succeeds with it.

wjordan:
After running this patch in production for a while, I observed two issues

1. RPCs for very large, non-blocking queries were timing out because their total transfer time was greater than `rpcHoldTimeout`. This makes sense- the timeout is only an expectation on the amount of time for an RPC call to complete, not on the amount of time the largest-possible response is expected to be transferred across the connection.

    To fix this, I wrapped the stream with configurable timeouts, where each read/write operation extends the stream deadline by a timeout value. This fix is in 2f2857e88e0a2f00991e0f763860fd613721ae01.

2. After applying the rpcHoldTimeout to all client RPC calls, I noticed one type of blocking query (`KVS.Get` / `KVS.List`) wasn't having its timeout set properly based on the query parameters. Removing an ampersand from the `args` parameter fixed this issue (https://github.com/hashicorp/consul/pull/11500/commits/92525c40ee6feefcd0b299af16095c34a2971adb). This probably also affects the existing RPC retry logic for this RPC method so could possibly be extracted as a separate fix PR.

dnephin:
@wjordan thanks for working on this! I'll need to dedicate some time to look into all of the links you've shared. It is quite possible that my understanding is stale and maybe we can improve the timeout support. I'm hoping to look into this in more detail in the next couple of weeks.

agent/pool/pool.go
```diff
@@ -55,6 +55,47 @@ type Conn struct {
 	clientLock sync.Mutex
 }
 
+// TimeoutConn wraps net.Conn with a default timeout.
+// readTimeout and writeTimeout apply to the subsequent operation.
```
kisunji, line 59: Is this comment accurate? `Read` and `Write` implementations look like `ReadTimeout` and `WriteTimeout` are applied first, zero'ed out, then `DefaultTimeout` is applied on subsequent operations.

agent/consul/server_test.go
```diff
@@ -177,7 +177,7 @@ func testServerConfig(t *testing.T) (string, *Config) {
 
 	// TODO (slackpad) - We should be able to run all tests w/o this, but it
 	// looks like several depend on it.
-	config.RPCHoldTimeout = 5 * time.Second
+	config.RPCHoldTimeout = 10 * time.Second
```
kisunji, line 180: Is this necessary for tests to pass?

agent/pool/pool.go
```diff
@@ -132,6 +175,9 @@ type ConnPool struct {
 	// TODO: consider refactoring to accept a full yamux.Config instead of a logger
 	Logger *log.Logger
 
+	// The default timeout for stream reads/writes
+	Timeout time.Duration
+
```
kisunji, line 180: Could you help me understand the difference between `ConnPool.Timeout` vs the timeout args that have been added to all the `rpc` methods?

I'm wondering if it's possible to achieve the same behaviour without adding an arg (which is defaulted to 0 in most cases) to an already long list of arguments to `rpc`.

agent/pool/pool.go
```diff
@@ -78,12 +119,14 @@ func (c *Conn) getClient() (*StreamClient, error) {
 		return nil, err
 	}
 
+	timeoutStream := &TimeoutConn{stream, c.pool.Timeout, 0, 0}
```
kisunji, line 122: It would be more readable to explicitly define struct fields here

kisunji: Do we ever set WriteTimeout anywhere else in this PR? Wondering if it's valuable to keep read/write timeouts separate (my naive assumption would be that the upper bound on size of reads and writes wouldn't vary enough to benefit from tuning them separately)

kisunji (commented):
Thank you for this PR and apologies for the delay in review. This PR touches on critical code paths so we'll try to get input from more members of our team.

In the mean time, I've left a couple of questions I'd like to see clarified

agent/pool/pool.go
```diff
@@ -55,6 +55,47 @@ type Conn struct {
...
+	}
+	deadline := time.Time{}
+	if timeout > 0 {
+		deadline = time.Now().Add(timeout)
+	}
+	if err := c.Conn.SetReadDeadline(deadline); err != nil {
+		return 0, err
+	}
+	return c.Conn.Read(b)
+}
```
kisunji, line 81: Probably out of scope for this PR but wondering if there's any room for optimization here and in `Write` by reducing allocations.

agent/pool/pool.go
wjordan: The comment is intentional but could perhaps be worded more clearly. What I meant was that setting a read/write timeout only affects the _very next_ operation, and then the timeout reverts to the default after that.

agent/consul/server_test.go
wjordan: Yes, I don't recall exactly (it's been a few months) but according to the commit message I left in cb4de4dc7e07b008f594ed5cc0b1701fbafcf990, the 5-second `RPCHoldTimeout` was causing some timeouts when running many tests concurrently.

agent/pool/pool.go
wjordan: `ConnPool.Timeout` ('The default timeout for stream reads/writes') is set from `RPCHoldTimeout`, and represents the default timeout for any stream read/write when an explicit timeout for a specific read/write call is not provided. `RPCHoldTimeout` represents an existing duration the server is tuned to expect round-trips to be shorter than, so it seemed like a reasonable timeout to use for most reads/writes going through this connection pool.

The `ConnPool.rpc` method sets a specific `timeout` for the first stream read after the query is sent. This is because the expected wait time for a (potentially blocking) query response can be much longer than the default round-trip timeout configuration.

> I'm wondering if it's possible to achieve the same behaviour without adding an arg (which is defaulted to 0 in most cases) to an already long list of arguments to rpc.

Thanks for the suggestion. Yes, Dropping the `timeout` argument by moving the RPC-timeout calculation from `Client.RPC` to `ConnPool.rpc` requires `MaxQueryTime` and `DefaultQueryTime` to be passed to `ConnPool`, but I think it does end up a bit cleaner that way.

agent/pool/pool.go
wjordan: `WriteTimeout` isn't used anywhere, so I can remove `WriteTimeout` and just make `TimeoutConn.Write` pass through to `c.Conn.Write`.

The read timeout needs to be kept separate from writes though, because the RPC timeout is read-specific, it applies to the first stream `Read` following the RPC request.

wjordan:
@kisunji thanks for the detailed review! Sorry for taking a while for my response. I've pushed a couple of commits that reduce the size of the PR diff based on your feedback and should hopefully address the helpful questions you raised.

kisunji:
Ah it looks like someone pushed some structural refactors to RPCs. Could you rebase?
Sorry for the delay in review+merge; I'm trying to find another reviewer for this PR

kisunji:
Hope you don't mind me pushing commits; trying not to let this go stale as this is a valuable fix.

kisunji (approved):

agent/structs/structs.go
```diff
@@ -315,18 +314,24 @@ func (q *QueryOptions) SetTokenSecret(s string) {
...
 		q.MaxQueryTime += lib.RandomStagger(q.MaxQueryTime / JitterFraction)
 
-		return time.Since(start) > (q.MaxQueryTime + rpcHoldTimeout), nil
+		return q.MaxQueryTime + rpcHoldTimeout
 	}
-	return time.Since(start) > rpcHoldTimeout, nil
+	return rpcHoldTimeout
+}
+
+func (q QueryOptions) HasTimedOut(start time.Time, rpcHoldTimeout, maxQueryTime, defaultQueryTime time.Duration) (bool, error) {
```
kyhavlov, line 333: Do we still need all these types to implement `HasTimedOut` if it's the same everywhere? It seems like we could have one version of this (not attached to a type) that gets a `structs.RPCInfo` passed in and we could call `Timeout` on that.

kisunji: Good point, I'll see if I can shrink the RPCInfo interface

kisunji: I spent some time refactoring locally but think it's too big of a change for this PR. Will raise an issue separately; the whole machinery with maxQueryTime and defaultQueryTime is confusing and only relevant for blocking queries

agent/kvs_endpoint.go
```diff
@@ -72,7 +72,7 @@ func (s *HTTPHandlers) KVSGet(resp http.ResponseWriter, req *http.Request, args
 
 	// Make the RPC
 	var out structs.IndexedDirEntries
-	if err := s.agent.RPC(method, &args, &out); err != nil {
+	if err := s.agent.RPC(method, args, &out); err != nil {
```
kisunji, line 75: > 2\. After applying the rpcHoldTimeout to all client RPC calls, I noticed one type of blocking query (`KVS.Get` / `KVS.List`) wasn't having its timeout set properly based on the query parameters. Removing an ampersand from the `args` parameter fixed this issue ([92525c4](https://github.com/hashicorp/consul/commit/92525c40ee6feefcd0b299af16095c34a2971adb)). This probably also affects the existing RPC retry logic for this RPC method so could possibly be extracted as a separate fix PR.

This seems like a really weird class of bug as the behaviour shouldn't depend on `args` being mutated during RPC. @wjordan was this encountered during your own testing?

kisunji: Just as precaution I'd like to revert this one-liner and file another issue regarding KVS in case we ever need to debug this PR in the future.

kyhavlov (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/648492.

agent/kvs_endpoint.go
wjordan: Yes, I discovered this issue in my own testing on a production cluster, without this change I was seeing timeouts on blocking queries to these methods. I can't say I quite understand why this change fixed the issue, only that it did seem to fix it for my particular setup.

Moving the fix to a separate PR sounds like a good precaution in any case.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/651287.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit c48120d005539b687391b6861f24c13d553f5b05 onto `release/1.12.x` succeeded!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/651305.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit c48120d005539b687391b6861f24c13d553f5b05 onto `release/1.12.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/651305)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit c48120d005539b687391b6861f24c13d553f5b05 onto `release/1.11.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/651305)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit c48120d005539b687391b6861f24c13d553f5b05 onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/651305)

PR #11507, opened 2021-11-05
Improved performance of the version.GetHumanVersion function by 50% on memory allocation.

haxandmat:
Improved performance and added benchmark for func version.GetHumanVersion.

Instructions:
```
$ go install golang.org/x/perf/cmd/benchstat@latest

$ go test -bench=GetHumanVersion -count 5 -benchmem ./version > old.txt
$ go test -bench=GetHumanVersion -count 5 -benchmem ./version > new.txt

$ benchstat old.txt new.txt
```

Results
```
name               old time/op    new time/op    delta
GetHumanVersion-8     177ns ± 2%      80ns ± 3%  -54.53%  (p=0.008 n=5+5)

name               old alloc/op   new alloc/op   delta
GetHumanVersion-8     32.0B ± 0%     16.0B ± 0%  -50.00%  (p=0.008 n=5+5)

name               old allocs/op  new allocs/op  delta
GetHumanVersion-8      3.00 ± 0%      2.00 ± 0%  -33.33%  (p=0.008 n=5+5)
```

### DIFF
--- a/version/version_test.go
+++ b/version/version_test.go
@@ -1,0 +1,11 @@
+package version
+
+import (
+	"testing"
+)
+
+func BenchmarkGetHumanVersion(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		GetHumanVersion()
+	}
+}
--- a/.changelog/11507.txt
+++ b/.changelog/11507.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+version: Improved performance of the version.GetHumanVersion function by 50% on memory allocation.
+```
--- a/version/version.go
+++ b/version/version.go
@@ -1,7 +1,6 @@
 package version
 
 import (
-	"fmt"
 	"strings"
 )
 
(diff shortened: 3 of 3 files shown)


eculver (approved):
LGTM!

version/version_test.go
```diff
@@ -0,0 +1,11 @@
+package version
+
+import (
+	"testing"
+)
+
+func BenchmarkGetHumanVersion(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		GetHumanVersion()
```
eculver, line 9: We're a little surprised that, because there's no assignment here, that the compiler doesn't do some optimizations but looks good.

eculver:
I'm not sure why this is still stuck but without the report back from CircleCI and without a new commit, it's just going to stay stuck. I tested this locally so I don't see any harm in just merging it so that's what I'll do :)

PR #11509, opened 2021-11-05
rename helper method to reflect the non-deprecated terminology

rboyer:
also add a state store helper function for unit tests

### DIFF
--- a/agent/consul/state/state_store.go
+++ b/agent/consul/state/state_store.go
@@ -203,6 +203,28 @@ func (s *Store) Snapshot() *Snapshot {
 	return &Snapshot{s, tx, idx}
 }
 
+// WalkAllTables basically lets you dump memdb generically and exists primarily
+// for very specific types of unit tests and should not be executed in
+// production code.
+func (s *Store) WalkAllTables(fn func(table string, item interface{}) bool) error {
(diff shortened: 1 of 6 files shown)


kisunji (approved):

markan (approved):

agent/consul/state/state_store.go
```diff
@@ -203,6 +203,28 @@ func (s *Store) Snapshot() *Snapshot {
 	return &Snapshot{s, tx, idx}
 }
 
+// WalkAllTables basically lets you dump memdb generically and exists primarily
+// for very specific types of unit tests and should not be executed in
+// production code.
+func (s *Store) WalkAllTables(fn func(table string, item interface{}) bool) error {
```
dnephin, line 209: If this is used for tests, maybe this could be in a test file instead of a production file?

dnephin (commented):

agent/consul/state/state_store.go
rboyer: It's used for tests in other packages so that wouldn't work.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/495054.

PR #11514, opened 2021-11-05
ca: properly handle the case where the secondary initializes after the primary

### LINKED ISSUE #11367: ca: do not generate a ClusterID in the secondary
This problem was noticed by the `consul-k8s` test suite. @freddygv and I found that:

1. The ["servers are upgraded" checks](https://github.com/hashicorp/consul/blob/v1.10.3/agent/consul/leader_connect_ca.go#L363-L369) should return an error to indicate that initialization was not complete, and should be retried
2. [InitializeCAConfig](https://github.com/hashicorp/consul/blob/v1.10.3/agent/consul/leader_connect_ca.go#L341-L345) should not be called in the secondary because it generates a ClusterID. We should initialize the config without generating a ClusterID.

The test failure may also be related to #10871 , because it appeared that a client agent may not have received the updated ClusterID after the secondary servers received the ClusterID from the primary.

dnephin:
Fixes #11367

Previously `secondaryInitialize` would return nil in this case, which prevented the
deferred initialize from happening, and left the CA in an uninitialized state until a config
update or root rotation.

To fix this I extracted the common parts into the delegate implementation. However looking at this
again, it seems like the handling in secondaryUpdateRoots is impossible, because that function
should never be called before the secondary is initialzied. I beleive we can remove some of that
logic in a follow up.

TODO:
* [ ] how can we test this?
* [ ] why is ClusterID even in the config?

### DIFF
--- a/agent/consul/leader_connect_test.go
+++ b/agent/consul/leader_connect_test.go
@@ -1094,7 +1094,6 @@ func TestLeader_SecondaryCA_UpgradeBeforePrimary(t *testing.T) {
 
 	// Wait for the secondary transition to happen and then verify the secondary DC
 	// has both roots present.
-	secondaryProvider, _ := getCAProviderWithLock(s2)
 	retry.Run(t, func(r *retry.R) {
 		state1 := s1.fsm.State()
 		_, roots1, err := state1.CARoots(nil)
@@ -1110,14 +1109,17 @@ func TestLeader_SecondaryCA_UpgradeBeforePrimary(t *testing.T) {
 		require.Equal(r, roots1[0].ID, roots2[0].ID)
 		require.Equal(r, roots1[0].RootCert, roots2[0].RootCert)
 
+		secondaryProvider, _ := getCAProviderWithLock(s2)
 		inter, err := secondaryProvider.ActiveIntermediate()
 		require.NoError(r, err)
 		require.NotEmpty(r, inter, "should have valid intermediate")
 	})
 
-	_, caRoot := getCAProviderWithLock(s1)
+	secondaryProvider, _ := getCAProviderWithLock(s2)
 	intermediatePEM, err := secondaryProvider.ActiveIntermediate()
 	require.NoError(t, err)
+
+	_, caRoot := getCAProviderWithLock(s1)
 
 	// Have dc2 sign a leaf cert and make sure the chain is correct.
 	spiffeService := &connect.SpiffeIDService{
--- a/agent/consul/server_connect.go
+++ b/agent/consul/server_connect.go
@@ -16,19 +16,23 @@ func (s *Server) getCARoots(ws memdb.WatchSet, state *state.Store) (*structs.Ind
 	if err != nil {
 		return nil, err
 	}
+	if config == nil {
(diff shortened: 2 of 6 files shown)


agent/consul/leader_connect_test.go
```diff
@@ -1110,15 +1109,18 @@ func TestLeader_SecondaryCA_UpgradeBeforePrimary(t *testing.T) {
 		require.Equal(r, roots1[0].ID, roots2[0].ID)
 		require.Equal(r, roots1[0].RootCert, roots2[0].RootCert)
 
+		secondaryProvider, _ := getCAProviderWithLock(s2)
```
dnephin, line 1112: This had to move into the retry loop, otherwise it was pointing at the old not initialized provider that existed before the secondary could be initialized.

agent/consul/state/connect_ca.go
```diff
@@ -180,8 +180,6 @@ func (s *Store) caSetConfigTxn(idx uint64, tx WriteTxn, config *structs.CAConfig
 	if prev != nil {
 		existing := prev.(*structs.CAConfiguration)
 		config.CreateIndex = existing.CreateIndex
-		// Allow the ClusterID to change if it's provided by an internal operation, such
-		// as a primary datacenter being switched to secondary mode.
```
dnephin, line 184: I deleted this comment since I came across it while tracing this code and it is misleading. We always allow changing the ClusterID. All these next lines do is default it when the new config has it unset.

kyhavlov (commented):
These changes look good - we could probably remove `ClusterID` from the config altogether in the future (we don't document it as a configurable field) and just persist it in the roots object in the state store. I think having it in two places is definitely a source of unnecessary confusion.

kyhavlov (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/496536.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 7d95d90fce00199dfdccbaa75eaaaba22c7be698 onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/496536)

agent/consul/server_connect.go
```diff
@@ -16,19 +16,23 @@ func (s *Server) getCARoots(ws memdb.WatchSet, state *state.Store) (*structs.Ind
...
+	signingID := connect.SpiffeIDSigningForCluster(config)
+	if signingID == nil {
+		// If CA is bootstrapped at all then this should never happen but be
+		// defensive.
+		return nil, fmt.Errorf("no cluster trust domain setup")
+	}
 
-		indexedRoots.TrustDomain = signingID.Host()
+	indexedRoots.TrustDomain = signingID.Host()
+	if indexedRoots.TrustDomain == "" {
```
dnephin, line 34: `Host()` returns `clusterID.Domain`, and Domain is hardcoded to `consul` in `SpiffeIDSigningForCluster`, so I don't think this will ever eventuate to true. Should this only check the `ClusterID` from config?

cc @freddygv

Issue #11291, opened 2021-10-12
improve ingress gateway xDS generation handling of proxycfg eventual consistency


rboyer:
In `ResourceGenerator.clustersFromSnapshotIngressGateway` [we loop over](https://github.com/hashicorp/consul/blob/main/agent/xds/clusters.go#L385) the `cfgSnap.IngressGateway.Upstreams` and lookup discovery chains from `cfgSnap.IngressGateway.DiscoveryChain` and fail if the chain is nil. This is problematic because these two things are populated from different blocking queries that respond at different times so in a way it's expected that during config-entry writes occasionally the discovery chain resolution will be guaranteed to be `nil`.

All of these upstream loops should likely be switched to use a helper method that only loops over the upstreams that are "ready" akin to a similar thing done for terminating gateways `cfgSnap.TerminatingGateway.ValidServices()`

PR #11516, opened 2021-11-06
xds, proxycfg: improve eventual consistency in ingress gateway generator

gmichelo:
* Implemented (*configSnapshotIngressGateway) ValidUpstreams to return the upstreams that are really ready to be used.
* Refactored (*ResourceGenerator) clustersFromSnapshotIngressGateway to use the new ValidUpstreams function.

Fixes: #11291

### DIFF
--- a/agent/proxycfg/snapshot.go
+++ b/agent/proxycfg/snapshot.go
@@ -379,6 +379,29 @@ func (c *configSnapshotIngressGateway) IsEmpty() bool {
 		len(c.WatchedUpstreamEndpoints) == 0
 }
 
+// ValidUpstreams returns the list of upstreams that have enough data to be used.
+func (c *configSnapshotIngressGateway) ValidUpstreams() structs.Upstreams {
+	out := make(structs.Upstreams, 0, len(c.Upstreams))
+	for _, upstreams := range c.Upstreams {
+		for _, u := range upstreams {
+			id := u.Identifier()
+
+			// Skip the Upstream if the discovery did not return data.
+			if chain := c.DiscoveryChain[id]; chain == nil {
+				continue
+			}
+
+			// Skip the Upstream if it does not have backing endpoints.
+			if _, ok := c.WatchedUpstreamEndpoints[id]; !ok {
+				continue
+			}
+
+			out = append(out, u)
+		}
+	}
+	return out
+}
+
 type IngressListenerKey struct {
 	Protocol string
 	Port     int
--- a/.changelog/11516.txt
+++ b/.changelog/11516.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+xds, proxycfg: improve eventual consistency in ingress gateway generator.
+```
--- a/agent/proxycfg/testing.go
+++ b/agent/proxycfg/testing.go
@@ -2104,6 +2104,66 @@ func TestConfigSnapshotIngress_MultipleListenersDuplicateService(t testing.T) *C
 	return snap
 }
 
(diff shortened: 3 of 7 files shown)


eculver:
Thanks for taking this @bigmikes. I'm going to give our integration tests a pass to see if we can express this case. I'm hoping we can get it merged in the next few days.

eculver:
Ok, I added a couple tests for this new functionality. I want to get @rboyer and @freddygv to sign off before we merge.

agent/proxycfg/snapshot.go
```diff
@@ -379,6 +379,29 @@ func (c *configSnapshotIngressGateway) IsEmpty() bool {
 		len(c.WatchedUpstreamEndpoints) == 0
 }
 
+// ValidUpstreams returns the list of upstreams that have enough data to be used.
+func (c *configSnapshotIngressGateway) ValidUpstreams() structs.Upstreams {
```
FFMMM, line 383: curious if this should be called something like ~ `ReadyUpstreams` ? Or `UpstreamsWithData`

kisunji (approved):
Looks good but tagging @rboyer for a final check before merge

agent/proxycfg/snapshot.go
bigmikes: I've used a consistent naming with [cfgSnap.TerminatingGateway.ValidServices()](https://github.com/hashicorp/consul/blob/a38be58eaa029f7383b78de9b23f9a5dbe2f7054/agent/xds/routes.go#L90). That is what @rboyer suggested in [his issue](https://github.com/hashicorp/consul/issues/11291).

I agree that `ReadyUpstreams` might be more appropriate.

Let me know if we prefer consistency or better naming.

kisunji: I cast my vote for `ReadyUpstreams` as it seems more descriptive here

rboyer:
There are a few key things to address about this PR before we can merge it.

1) https://github.com/hashicorp/consul/pull/12125 landed which changed many `proxycfg.ConfigSnapshot` map keys from `string -> UpstreamID`; so this definitely needs a rebase and slight tweak

2) When introducing conditional logic changes in one of the resource types (cluster/endpoint/listener/route) you have to make super sure that all 4 primary types' inner loops are going to follow identical conditional logic lest you end up in an xDS protocol deadlock-ish situation where for example a Listener loop tells envoy it should have a Cluster named X, but the Cluster loop filters out the X cluster before talking to envoy. The easy way around this is to ensure that all 4 xDS type loops for ingress gateways use the `ValidUpstreams` helper.

Doing (1) isn't too bad, and theoretically doing (2) shouldn't be too bad but it runs into an issue with how we currently construct the test comparison fixtures for `proxycfg.ConfigSnapshot` objects for use in the golden tests.

Due to the age of the PR I took a crack at (1) and (2) myself. I can either force push to your branch (is that even possible?) or open a fresh PR and carry over your commit attribution there? Let me know what you want.

i'll have to craft a longer message to explain it to the person but all 4 xDS types need to uniformly use the new method he created, not just the one. That avoids an xDS protocol 

when i tried to do (1)+(2) locally it does work just fine EXCEPT for it breaks like a huge swath of our existing unit tests in the agent/xds package because they fail to declare the very thing that the ValidUpstreams function is enforcing in the test fixture setups

bigmikes:
Sorry for the delayed response. Thanks for the review! 

>I can either force push to your branch (is that even possible?) or open a fresh PR and carry over your commit attribution there? Let me know what you want.

Feel free to open a fresh PR, I believe it's the cleaner approach.

Issue #11511, opened 2021-11-05
Force show all UI capabilities that are hidden for users with partial permissions


jkirschner-hashicorp:
#### Use Case(s)

Regain the ability to perform (in the GUI) all operations your token allows.

#### Feature Description

We conditionally show some GUI elements (e.g., create buttons, nav menu items) depending on the permissions of the token used to log into the GUI. As described in https://github.com/hashicorp/consul/issues/11098, the backend call to check these permissions...
- can answer "all" questions (does the token have this permission on all of this type of resource?), but
- can **not** answer "any" questions (does the user have this permission to at least one of this type of resource?)

For some resources, a token can have partial ("any") but not full ("all") access. For example: read access to a Key-Value sub-tree, but not all Key-Values. For tokens with partial (not full) access, the backend currently responds with a denial, causing some UI elements to be hidden _even though a subset of actions is allowed_. This results in a state where some operations possible via CLI or HTTP API are not possible through the GUI (because access is hidden).

Until #11098 is fixed, we should **temporarily** circumvent the hide logic for UI elements related to resources which can have "partial" access, which includes:
- key/values
- nodes
- services
- intentions
- sessions

This will allow GUI users to perform all operations they are allowed to perform with their token (just like from the HTTP API and CLI). This also means that, like with the HTTP API and CLI, the user will be able to attempt operations that will be denied (due to a specific resource not falling within their partial permissions).

Issue #11098, opened 2021-09-21
/internal/acl/authorize HTTP endpoint does not behave correctly


dnephin:
### Context

`/internal/acl/authorize` is used by the UI to answer the question "does this ACL token have permission to do this _kind_ of thing. However, the current implementation of this endpoint uses an `acl.Authorizer` which can only answer questions about specific resources (given the resource ID). It can't answer the more general question of if the acl token has permission to any instance of that resource.

This issue describes a medium term fix. The short-term fix will be https://github.com/hashicorp/consul/issues/11511.

### Necessary changes

The `/internal/acl/authorize` endpoint must be adjusted to answer:
> does this token have `level` (e.g., read, write) access to **"any"** of this kind of resource

... for the following resources ([all resource types that have "labels"](https://www.consul.io/docs/security/acl/acl-rules#resource-and-rule-reference) **and** [are used by the UI](https://github.com/hashicorp/consul/blob/main/ui/packages/consul-ui/app/services/repository/permission.js#L10)):
- [ ] key/value
- [ ] node
- [ ] service
  - [ ] intention
- [ ] session

Additionally, the following related changes should be made to the `/internal/acl/authorize` to ensure consistency with the behavior of other endpoints:

- [ ] Use the token's default namespace if no namespace is provided in the request
- [ ] Consider the token's default partition as well (now that partition support is needed)

#### Undo UI Workaround

- [ ] Undo the temporary workaround in https://github.com/hashicorp/consul/pull/11520 used to close https://github.com/hashicorp/consul/issues/11511.

### Related Problems
This problem has caused the following issues that we know of:
1. #10944 was a UI regression in 1.10.x
2. the bug fixed by #11061 was a regression in 1.10.2 only
3. https://github.com/hashicorp/consul/issues/11204 sounds like it may be related as well

johncowen:
Note to future self:

> What's a bit strange to me is that we're not using "enable_key_list_policy". We're using anonymous token and while via the CLI I could have read the keys without a problem, in the GUI it's requiring assigning an extra permission.

From: https://github.com/hashicorp/consul/issues/11204

dnephin:
I had a look to see how much work this would be fix.

One decision we need to make is how to represent this query in the API. I think we have two options:
1. We could continue to use `*` to mean "any"
2. We could add a new bool field to `structs.ACLAuthorizationRequest`  to represent `Any`

[...]

PR #11520, opened 2021-11-08
ui: Temporarily force all UI capabilities

johncowen:
Fixes https://github.com/hashicorp/consul/issues/11511

Related: https://github.com/hashicorp/consul/issues/11098

cc @jkirschner-hashicorp @dnephin

### DIFF
--- a/ui/packages/consul-ui/app/services/repository/permission.js
+++ b/ui/packages/consul-ui/app/services/repository/permission.js
@@ -152,6 +152,8 @@ export default class PermissionService extends RepositoryService {
     // Temporarily revert to pre-1.10 UI functionality by overwriting frontend
     // permissions. These are used to hide certain UI elements, but they are
     // still enforced on the backend.
+    // This temporary measure should be removed again once https://github.com/hashicorp/consul/issues/11098
+    // has been resolved
     this.permissions.forEach(item => {
       if(['key', 'node', 'service', 'intentions', 'session'].includes(item.Resource)) {
         item.Allow = true;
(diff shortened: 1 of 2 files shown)


ui/packages/consul-ui/app/services/repository/permission.js
```diff
@@ -148,6 +148,16 @@ export default class PermissionService extends RepositoryService {
   async findAll(params) {
     params.resources = REQUIRED_PERMISSIONS;
     this.permissions = await this.findByPermissions(params);
+    /**/
+    // Temporarily revert to pre-1.10 UI functionality by overwriting frontend
+    // permissions. These are used to hide certain UI elements, but they are
+    // still enforced on the backend.
```
evrowe, line 154: Appreciate the comment for context. I know that the first word in the comment is "Temporarily," but I wonder if we should be a little more explicit about the fact that we intend to return to the intended behavior once some back-end dependencies have been resolved? We don't need to commit to a specific time frame but I feel like making the intentions clear could be beneficial to external contributors.

evrowe (approved):
Looks good from here, left a non-blocking question .

ui/packages/consul-ui/app/services/repository/permission.js
jkirschner-hashicorp: We could mention GH-11098 specifically as the issue to follow for a longer-term fix.

johncowen: Added some more detail here inc. GH issue link

evrowe: :shipit:

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/498950.

Issue #10132, opened 2021-04-27
mesh-federated wan clusters do not reconnect on an secondary mesh-gateway outage


dekimsey:
#### Overview of the Issue

If the secondary datacenter's mesh-gateway service experience a full outage (including IP reassignment) the wan looses it's connectivity until the primary's mesh-gateway's consul service is restarted.

#### Reproduction Steps

Steps to reproduce this issue, eg:

1. Create two clusters, a primary and secondary with WAN federation working.
1a. In my environment our gateways are running on separate consul nodes, this may or may not be relevant.
2. Take down all secondary mesh-gateways and recreate with new IPs.
3. Observe wan is unable to re-establish it's connections even though the secondary servers are still able to directly reach the primary's gateway without issue.

In an attempt to fix:
1. Restart all secondary servers in attempt to re-kicktstart bootstrap
In theory, we should have a partially connected wan at this stage. (At least that's how the initial step 1 worked during the initial configuration). However observe the primary's consul gateway still has the old IPs for all the secondary servers. Essentially the initial registration, didn't clear/reset/poke the consul data on the mesh-gateway.
2. Restart the consul agent on the mesh-gateway's node.
Gateway now has all the new secondary gateway's IPs.

Questions:
1. What is the process for recovery when a secondary gateway experiences an outage?
2. What is the process for recovery when the primary gateways experiences an outage? 
3. Are there any operator commands that might be used on either the primary or secondary to force a re-bootstrap of the federated servers?
4. Is a stable IP for the gateways a requirement (aka load-balancer)? What happens when this changes ("stable" is just a word reality doesn't seem to care much for!)

### Consul info for both Client and Server

<details>
  <summary>Client info</summary>

```
agent:
        check_monitors = 0
        check_ttls = 0
        checks = 1
        services = 1
build:
        prerelease =
        revision = 3c1c2267
        version = 1.9.5
consul:
        acl = enabled
        known_servers = 3
        server = false
runtime:
        arch = amd64
        cpu_count = 2
        goroutines = 2753
        max_procs = 2
        os = linux
        version = go1.15.8
serf_lan:
        coordinate_resets = 0
        encrypted = true
        event_queue = 0
        event_time = 834
        failed = 0
        health_score = 0
        intent_queue = 0
        left = 0
        member_time = 39575
        members = 322
        query_queue = 0
        query_time = 1
```

</details>

<details>
  <summary>Server info</summary>

```
agent:
        check_monitors = 0
        check_ttls = 1
[...]

PR #11522, opened 2021-11-08
xds: prefer fed state gateway definitions if they're fresher

boxofrad:
Fixes an issue (described in #10132) where if two DCs are WAN federated over mesh gateways, and the gateway in the non-primary DC is terminated and receives a new IP address (as is commonly the case when running them
on ephemeral compute instances) the primary DC is unable to re-establish its connection until the agent running on its own gateway is restarted.

This was happening because we always preferred gateways discovered by the `Internal.ServiceDump` RPC (which would fail because there's no way to dial the remote DC) over those discovered in the federation state, which is replicated as long as the primary DC's gateway is reachable - see the comment in `endpointsFromSnapshotMeshGateway`.

### DIFF
--- a/agent/proxycfg/testing.go
+++ b/agent/proxycfg/testing.go
@@ -1532,6 +1532,31 @@ func TestConfigSnapshotMeshGatewayNewerInformationInFederationStates(t testing.T
 	return snap
 }
 
+func TestConfigSnapshotMeshGatewayOlderInformationInFederationStates(t testing.T) *ConfigSnapshot {
+	snap := TestConfigSnapshotMeshGateway(t)
+
+	// Create a duplicate entry in FedStateGateways, with a low ModifyIndex, to
+	// verify that stale data in the federation state is ignored in favor of the
+	// fresher data in GatewayGroups.
+	svc := structs.TestNodeServiceMeshGatewayWithAddrs(t,
+		"10.0.1.3", 8443,
+		structs.ServiceAddress{Address: "10.0.1.3", Port: 8443},
+		structs.ServiceAddress{Address: "198.18.1.3", Port: 443},
+	)
+	svc.RaftIndex.ModifyIndex = 0
+
+	snap.MeshGateway.FedStateGateways = map[string]structs.CheckServiceNodes{
+		"dc2": {
+			{
+				Node:    snap.MeshGateway.GatewayGroups["dc2"][0].Node,
+				Service: svc,
+			},
+		},
+	}
+
+	return snap
+}
+
 func TestConfigSnapshotMeshGatewayNoServices(t testing.T) *ConfigSnapshot {
 	return testConfigSnapshotMeshGateway(t, false, false)
 }
--- a/agent/proxycfg/testing.go
+++ b/agent/proxycfg/testing.go
@@ -1506,6 +1506,47 @@ func TestConfigSnapshotMeshGatewayUsingFederationStates(t testing.T) *ConfigSnap
 	return testConfigSnapshotMeshGateway(t, true, true)
 }
 
+func TestConfigSnapshotMeshGatewayNewerInformationInFederationStates(t testing.T) *ConfigSnapshot {
+	snap := TestConfigSnapshotMeshGateway(t)
+
+	csn1 := snap.MeshGateway.GatewayGroups["dc2"][0]
+	csn2 := snap.MeshGateway.GatewayGroups["dc2"][1]
+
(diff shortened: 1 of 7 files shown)


agent/xds/endpoints.go
```diff
@@ -123,15 +123,63 @@ func (s *ResourceGenerator) endpointsFromSnapshotMeshGateway(cfgSnap *proxycfg.C
...
+		//	  direction).
+		//
+		//	- At this point the Primary DC's mesh gateway should observe the new IP
+		//	  address and reconfigure its proxy, however as we always prioritised
+		//	  GatewayGroups this didn't happen and the connection remained severed.
+		fedStateEndpoints := cfgSnap.MeshGateway.FedStateGateways[key.String()]
+		for _, fse := range fedStateEndpoints {
+			idx := -1
+
+			for i, e := range endpoints {
```
boxofrad, line 162: I'm assuming here that `len(fedStateEndpoint) * len(endpoints)` is small. Maybe this should be a map lookup instead?

agent/xds/endpoints.go
```diff
@@ -123,15 +123,64 @@ func (s *ResourceGenerator) endpointsFromSnapshotMeshGateway(cfgSnap *proxycfg.C
...
+					idx = i
+					break
+				}
+			}
+
+			switch {
+			case idx == -1:
+				// Definition is only present in the federation state, add it wholesale.
+				endpoints = append(endpoints, fse)
+			case endpoints[idx].Service.RaftIndex.ModifyIndex < fse.Service.RaftIndex.ModifyIndex:
```
rboyer, line 173: For a moment I thought this was a raft index apples/oranges comparison, but as long as you stick to the Service part of a CheckServiceNode entity in `FedStateGateways` that is going to be the same as the original data fetched in `GatewayGroups`

agent/xds/endpoints.go
```diff
@@ -123,15 +123,64 @@ func (s *ResourceGenerator) endpointsFromSnapshotMeshGateway(cfgSnap *proxycfg.C
 			continue
 		}
 
-		endpoints, ok := cfgSnap.MeshGateway.GatewayGroups[key.String()]
-		if !ok {
-			endpoints, ok = cfgSnap.MeshGateway.FedStateGateways[key.String()]
-			if !ok { // not possible
-				s.Logger.Error("skipping mesh gateway endpoints because no definition found", "datacenter", key)
-				continue
+		endpoints := cfgSnap.MeshGateway.GatewayGroups[key.String()].ShallowClone()
```
rboyer, line 126: How would you feel about this alternative?

1. collect the max inner `Service.RaftIndex.ModifyIndex` value from all MGW instances in `GatewayGroups`
2. do the same for `FedStateGateways` instances
3. depending upon which of the two data sources has the larger value, wholesale use the entire slice of data that came from it

The argument here is that on their own, each of these data sets come from singular RPC calls awoken from single blocking queries. They are each independently self-consistent, whichever one has more recent data by necessity would be in aggregate more correct, so there's no need to do a deep merge like this. Doing the either/or merge instead would also be easier to grok.

boxofrad: That's much easier to reason about, thanks! 🙌🏻

agent/proxycfg/testing.go
```diff
@@ -1506,6 +1507,31 @@ func TestConfigSnapshotMeshGatewayUsingFederationStates(t testing.T) *ConfigSnap
 	return testConfigSnapshotMeshGateway(t, true, true)
 }
 
+func TestConfigSnapshotMeshGatewayNewerInformationInFederationStates(t testing.T) *ConfigSnapshot {
```
dhiaayachi, line 1510: Do we have a test that validate the other alternative?  If not it would be good to have a test that verify switching from a data source to the other work properly.

dhiaayachi (approved):
I have a question about a possible missing test, but other then that it LGTM!!

agent/proxycfg/testing.go
boxofrad: Great point! Have added ✅

rboyer (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/497440.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 50a1f20ff9b0257c5384a5100e8e63c86114fd76 onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/497440)

PR #8714, opened 2020-09-19
Make tracing parameter configurable on ingress GW

Gufran:
Closes: #8519 

This changeset allows tracing configuration to be specified as part of ingress-gateway connect config.

PR #11536, opened 2021-11-09
add a metrics test check GH action

FFMMM:
#### Overview

This PR wraps a small bash checker in a GH action. The checks are actually done on "diffs" and not necessarily the actual content. That said, I think this get us 80% there!

##### Pros/ Wins:

* quick eye check for reviewers on whether or not this PR has metric behavior changes in it 
* reminder to contributors to add testing for the metric behavior
* helps us build metric behavior testing in the long run

This is not meant to be a replacement for a good old fashioned, through review. It's meant to have the same call to action as a change log PR checker.

#### Testing

* I did a fair amount of testing in the enterprise repo (mostly manual); happy to reproduce some of it here on request.

#### More background

At present, there are some metric behavior checks like [here](https://github.com/hashicorp/consul/pull/11231/files#diff-472e516d540a393b43170efc09098871a521b8b8ed4788089412829311ff81c5). My hope is that we get to a place where we test the behavior of most of our exported metrics if not all.

### DIFF
--- a/.github/scripts/metrics_checker.sh
+++ b/.github/scripts/metrics_checker.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+set -o pipefail
+
+# search for any "new" or modified metric emissions
+metrics_modified=$(git --no-pager diff HEAD origin/main | grep -i "SetGauge\|EmitKey\|IncrCounter\|AddSample\|MeasureSince\|UpdateFilter")
+# search for PR body or title metric references
+metrics_in_pr_body=$(echo "$PR_BODY" | grep "metric")
+metrics_in_pr_title=$(echo "$PR_TITLE" | grep "metric")
+
+# if there have been code changes to any metric or mention of metrics in the pull request body
+if [ "$metrics_modified" ] || [ "$metrics_in_pr_body" ] || [ "$metrics_in_pr_title" ]; then
+  # need to check if there are modifications to metrics_test
+  modified_metrics_test_file=$(git --no-pager diff --name-only HEAD "$(git merge-base HEAD "origin/main")" | grep -i "metrics_test")
+  if [ "$modified_metrics_test_file" ]; then
+    # 1 happy path: metrics_test has been modified bc we modified metrics behavior
+    echo "PR seems to modify metrics behavior. It seems it has modified agent/metrics_test.go as well."
+    exit 0
+  else
+    echo "PR seems to modify metrics behavior. It seems no tests or test behavior has been modified in agent/metrics_test.go."
+    echo "Please update the PR with any relevant updated testing or add a pr/no-metrics-test label to skip this check."
+    exit 1
+  fi
+
+else
+  # no metrics modified in code, nothing to check
+  echo "No metric behavior seems to be modified."
+  exit 0
+fi
--- a/.github/scripts/metrics_checker.sh
+++ b/.github/scripts/metrics_checker.sh
@@ -1,19 +1,24 @@
 #!/usr/bin/env bash
 set -o pipefail
 
+### This script checks if any metric behavior has been modified.
+### The checks rely on the git diff against origin/main
+### It is still up to the reviewer to make sure that any tests added are needed and meaningful.
+
 # search for any "new" or modified metric emissions
 metrics_modified=$(git --no-pager diff HEAD origin/main | grep -i "SetGauge\|EmitKey\|IncrCounter\|AddSample\|MeasureSince\|UpdateFilter")
 # search for PR body or title metric references
-metrics_in_pr_body=$(echo "$PR_BODY" | grep "metric")
-metrics_in_pr_title=$(echo "$PR_TITLE" | grep "metric")
+metrics_in_pr_body=$(echo "$PR_BODY" | grep -i "metric")
+metrics_in_pr_title=$(echo "$PR_TITLE" | grep -i "metric")
 
 # if there have been code changes to any metric or mention of metrics in the pull request body
 if [ "$metrics_modified" ] || [ "$metrics_in_pr_body" ] || [ "$metrics_in_pr_title" ]; then
   # need to check if there are modifications to metrics_test
-  modified_metrics_test_file=$(git --no-pager diff --name-only HEAD "$(git merge-base HEAD "origin/main")" | grep -i "metrics_test")
+  metrics_test_file="agent/metrics_test.go"
+  modified_metrics_test_file=$(git --no-pager diff --name-only HEAD "$(git merge-base HEAD "origin/main")" | grep -i "$metrics_test_file")
   if [ "$modified_metrics_test_file" ]; then
     # 1 happy path: metrics_test has been modified bc we modified metrics behavior
-    echo "PR seems to modify metrics behavior. It seems it has modified agent/metrics_test.go as well."
+    echo "PR seems to modify metrics behavior. It seems it may have added tests to agent/metrics_test.go as well."
     exit 0
   else
     echo "PR seems to modify metrics behavior. It seems no tests or test behavior has been modified in agent/metrics_test.go."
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-metrics-test-checker.yml
```diff
@@ -0,0 +1,21 @@
+name: "Check for metrics tests"
+on:
+  pull_request:
+    types: [ opened, synchronize, labeled ]
+    # Runs on PRs to main and all release branches
```
FFMMM, line 5: comment needs updating

eculver:
Could you stage a change in this PR that triggers the check so we can see it in action?

.github/scripts/metrics_checker.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+set -o pipefail
```
eculver, line 2: Suggest using `set -euo pipefail` for all the reasons mentioned [here](https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/) (minus the `-x` which is good for debugging imho). In general, using these options will help us catch strange side-effects as things around this script potentially change.

.github/scripts/metrics_checker.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# search for any "new" or modified metric emissions
+metrics_modified=$(git --no-pager diff HEAD origin/main | grep -i "SetGauge\|EmitKey\|IncrCounter\|AddSample\|MeasureSince\|UpdateFilter")
+# search for PR body or title metric references
+metrics_in_pr_body=$(echo "$PR_BODY" | grep "metric")
```
eculver, line 7: should we use `grep -i` here to catch mixed case? eg. "Metrics for X updated to ...'"

.github/scripts/metrics_checker.sh
```diff
@@ -0,0 +1,28 @@
...
+metrics_in_pr_title=$(echo "$PR_TITLE" | grep "metric")
+
+# if there have been code changes to any metric or mention of metrics in the pull request body
+if [ "$metrics_modified" ] || [ "$metrics_in_pr_body" ] || [ "$metrics_in_pr_title" ]; then
+  # need to check if there are modifications to metrics_test
+  modified_metrics_test_file=$(git --no-pager diff --name-only HEAD "$(git merge-base HEAD "origin/main")" | grep -i "metrics_test")
+  if [ "$modified_metrics_test_file" ]; then
+    # 1 happy path: metrics_test has been modified bc we modified metrics behavior
+    echo "PR seems to modify metrics behavior. It seems it has modified agent/metrics_test.go as well."
```
eculver, line 16: Should this use `$modified_metrics_test_file` instead of a hard-coded `agent/metrics_test.go`? This makes me wonder if we should be using a different technique to target the test file. For instance, this would break if we ever moved or renamed this file 🤔

eculver (commented):
I have a few comments on the code and I'm also wondering if this could be applied more broadly? For example, wouldn't it make sense to ensure that _any_ path that changes has associated tests updated? It reminds me of checks that Codecov does to call out parts of the code in a PR that are missing test coverage. Basically, I'm wondering if there are other tools out there that do this already that we can leverage? My worry is that we do this for metrics and it's great, but what about all the other packages that could use similar treatment? 
Screen Shot 2021-11-17 at 9 44 49 AM
(from #8714 btw)

.github/scripts/metrics_checker.sh
FFMMM: Initially I had set the following but decided to go without:

`set -u`

for when we don't have variables hold a diff they are empty, which will fail the script. The alternative would be to assign placeholder values but I think that also can be a source of bugs!

.github/scripts/metrics_checker.sh
FFMMM: > Should this use `$modified_metrics_test_file` instead of a hard-code

oh yes actually this is a good point! I can add this in; the cmd will look at the diff and only `metrics_test` so this should be fine if the file moves.

FFMMM:
Hey @eculver thanks for looking at this and for your patience with my response here!

> Could you stage a change in this PR that triggers the check so we can see it in action (no pun intended)?

Yes, to reduce noise on this repo, there's staged PRs on the ent repo 🙈 

> I'm also wondering if this could be applied more broadly? 

Maybe, but my only concern here for now is for metrics eye checks. I think the intent here is to get contributors to test metric behaviors.

> Basically, I'm wondering if there are other tools out there that do this already that we can leverage? 

I did do some research on the Actions marketplace but couldn't find an off the shelf for this use case before I wrote this up.

 > My worry is that we do this for metrics and it's great, but what about all the other packages that could use similar treatment?
 
 At this time, I want us to make sure we don't break metrics behavior again. This has happened a few times in the past and it's not great. 
 
 Again, this check is more of an eye test/ nudge. It's still up to the reviewer to check that tests are correct.

eculver:
> At this time, I want us to make sure we don't break metrics behavior again. 

Got it. This is the point and I think it's the most important part of all of this. I overlooked the fact that the metrics behavior was broken at some point. With that in mind, I'm a +1.

.github/scripts/metrics_checker.sh
dnephin: I always use `set -u`. To deal with variables that might be empty you can do this: `${varname-}`. That tells bash, "it's ok for this value to be unset". That way you have to explicit allow something being unset, instead of it being implied everywhere.

dnephin: You can see that in the docs here: https://tldp.org/LDP/abs/html/parameter-substitution.html, the second example

FFMMM: thanks for the pointers -- i added that over the env var passed to the script 💯 !

.github/scripts/metrics_checker.sh
```diff
@@ -0,0 +1,33 @@
...
+set -uo pipefail
+
+### This script checks if any metric behavior has been modified.
+### The checks rely on the git diff against origin/main
+### It is still up to the reviewer to make sure that any tests added are needed and meaningful.
+
+# search for any "new" or modified metric emissions
+metrics_modified=$(git --no-pager diff HEAD origin/main | grep -i "SetGauge\|EmitKey\|IncrCounter\|AddSample\|MeasureSince\|UpdateFilter")
+# search for PR body or title metric references
+metrics_in_pr_body=$(echo "${PR_BODY-""}" | grep -i "metric")
```
eculver, line 11: I think this is missing a `:` after `PR_BODY` to get the right default behavior? So:
```
metrics_in_pr_body=$(echo "${PR_BODY:-""}" | grep -i "metric")
```
Same thing below for the `PR_TITLE`.

eculver:
I think we could at least try this out for a little while and, because the change includes an escape hatch (eg. pr/no-metrics-check label), we could work around it if something strange does come up that we haven’t thought about yet. What do you all think?

.github/scripts/metrics_checker.sh
FFMMM: daniel had shared these useful examples from https://tldp.org/LDP/abs/html/parameter-substitution.html and based off of those, I figured that `PR_BODY` and `PR_TITLE` should never be null.

They get set from the github action and empty bodies or titles should come out as the empty string. If they don't (if they are set to NULL), I'd like us to know about it because something is not right.

So that's why I left out the `:` -- does that make sense?

eculver: Yes, the logic makes sense. I think my comment wasn't clear. I was saying that without the colon it wasn't valid syntax but I was wrong. I guess the colon is optional and I've just always written it like that :) 
```
bash-3.2$ function foo { local bar; bar="${NOT_SET-default}"; echo $bar; }
bash-3.2$ foo
default
bash-3.2$ function foo { local bar; bar="${NOT_SET:-default}"; echo $bar; }
bash-3.2$ foo
default
```

eculver (approved):

dnephin (approved):
Cool! Let's try it out!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/533980.

PR #11539, opened 2021-11-09
1.10.x: properly handle the case where the secondary initializes after the primary

### LINKED ISSUE #11367: ca: do not generate a ClusterID in the secondary

freddygv:
Backport of #11514

---------

Fixes #11367

Previously `secondaryInitialize` would return nil in this case, which prevented the
deferred initialize from happening, and left the CA in an uninitialized state until a config
update or root rotation.

To fix this I extracted the common parts into the delegate implementation. However looking at this
again, it seems like the handling in secondaryUpdateRoots is impossible, because that function
should never be called before the secondary is initialzied. I beleive we can remove some of that
logic in a follow up.

### DIFF
--- a/.changelog/11514.txt
+++ b/.changelog/11514.txt
@@ -1,0 +1,6 @@
+```release-note:improvement
+connect/ca: Return an error when querying roots from uninitialized CA.
+```
+```release-note:bug
+connect/ca: Allow secondary initialization to resume after being deferred due to unreachable or incompatible primary DC servers.
+```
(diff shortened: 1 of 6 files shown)


kyhavlov (approved):

markan (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/498603.

PR #11540, opened 2021-11-09
Update filter chain creation for listeners

freddygv:
Backport of #11245

------------------------

The duo of `makeUpstreamFilterChainForDiscoveryChain` and `makeListenerForDiscoveryChain` were really hard to reason about, and led to concealing a bug in their branching logic. There were several issues here:

* They tried to accomplish too much: determining filter name, cluster name, and whether RDS should be used. 
* They embedded logic to handle significantly different kinds of upstream listeners (passthrough, prepared query, typical services, and catch-all)
* They needed to coalesce different data sources (Upstream and CompiledDiscoveryChain)

Rather than handling all of those tasks inside of these functions, this PR pulls out the RDS/clusterName/filterName logic.

This refactor also fixed a bug with the handling of [UpstreamDefaults](https://www.consul.io/docs/connect/config-entries/service-defaults#defaults). These defaults get stored as UpstreamConfig in the proxy snapshot with a DestinationName of "*", since they apply to all upstreams. However, this wildcard destination name must not be used when creating the name of the associated upstream cluster. The coalescing logic in the original functions here was in some situations creating clusters with a `*.` prefix, which is not a valid destination.

### DIFF
--- a/.changelog/11245.txt
+++ b/.changelog/11245.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+connect: fix issue with attempting to generate an invalid upstream cluster from UpstreamConfig.Defaults.
+```
--- a/agent/xds/listeners.go
+++ b/agent/xds/listeners.go
@@ -107,17 +107,31 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
 			continue
 		}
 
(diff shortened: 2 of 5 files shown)


rboyer (approved):
LGTM. Compared this patch vs the main patch and they were effectively the same minus the file relocations and partition removal.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/498678.

PR #11541, opened 2021-11-09
OSS Backport: Prevent writing partition-exports entries to secondary DCs

freddygv:
Best viewed by commit.

The first two commits add errors when a user attempts to target a secondary DC when writing a partition-exports config entry. The write would have typically been forwarded to the primary, regardless of the source DC and DC of the request. Since this config entry has a narrow scope we will reject certain writes to avoid silent failures.

The third commit prevents writing the config entry to a secondary DC via replication.

### DIFF
--- a/.changelog/11541.txt
+++ b/.changelog/11541.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+partitions: Prevent writing partition-exports entries to secondary DCs.
+```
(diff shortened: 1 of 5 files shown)


kyhavlov (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/499292.

PR #11545, opened 2021-11-10
Use ClusterID to check for readiness

freddygv:
Surfaced [here](https://github.com/hashicorp/consul/pull/11514/files/d9110136f239b6d89471df028d0ce1093f398ab9..cc5a7ed36c893622f3735977bdc3feb56335d70a#r746828464).

The TrustDomain is populated from the Host() method which includes the
hard-coded "consul" domain. This means that despite having an empty
cluster ID, the TrustDomain won't be empty.

### DIFF
--- a/agent/consul/server_connect.go
+++ b/agent/consul/server_connect.go
@@ -16,7 +16,7 @@ func (s *Server) getCARoots(ws memdb.WatchSet, state *state.Store) (*structs.Ind
 	if err != nil {
 		return nil, err
 	}
-	if config == nil {
+	if config == nil || config.ClusterID == "" {
 		return nil, fmt.Errorf("CA has not finished initializing")
 	}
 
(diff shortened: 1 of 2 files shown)


dnephin (commented):
Thanks for making this change! This should work for the primary DC, but I think doesn't work for the secondary because we no longer set the ClusterID on this field.

dnephin (approved):
Ah, nevermind! As you pointed you it is set in `persistNewRootAndConfig`

dnephin:
`TestEnvoy/case-wanfed-gw` failed every time,  it seems unhappy

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/499725.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 09ef65d053d4fa9a434b69ab806fd938d27182ff onto `release/1.10.x` succeeded!

PR #11565, opened 2021-11-12
changelog sync

rboyer:


.changelog/_1330.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+acl: **(Enterprise only)** fix namespace policy evaluation in the context of partitions
```
dnephin, line 2: Might be worth clarifying that it was specifically for namespace_prefix with the empty string (if I understood the bug correctly).

dnephin (approved):

.changelog/_1330.txt
rboyer: It was slightly more general in that if you had an exact and a prefix match that were factors in the same authz request, and only the prefix rule set `policy = <anything>` then it would be discarded

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/501971.

Issue #11276, opened 2021-10-12
Indicate when list in API endpoint response is filtered by ACLs


jkirschner-hashicorp:
#### Feature Description

Endpoints that return list of items will typically enforce ACLs by filtering the result set, such as:
- Agent endpoints for local checks/services
- RPC endpoints for queries to Consul servers

However, Consul doesn't give any indication that the results were filtered by ACLs.

For a user looking at the list of results, it is unclear whether they are viewing the full list, or whether they are seeing a partial (or empty) list due to incorrect permissions.

This information could be exposed through returning a new header.

For the service mesh use case, we could log in the proxycfg package when a query is filtered by ACLs. This would help discover and troubleshoot ACL permissions issues.

A separate issue can be created in the future to consider propagating this information to the Consul GUI.

#### Use Case(s)

General discovery and troubleshooting of ACL issues via the API endpoints. Potentially enabling surfacing this information on the Consul GUI in the future.

jkirschner-hashicorp:
Hi @OverDrone,

Can you tell me more about the difference for your use case between these two approaches?
- the approach in the issue description: return 200 OK with a Boolean to indicate whether some results were filtered due to ACLs
- the approach you suggested: opt-in to returning 403 if some results were filtered due to ACLs

Thanks for linking to your Discuss forum post for additional context!

OverDrone:
the main difference is in how client app treats filtered response. In my case if anything was filtered means exception. Also for kv there is no fine tune of permissions, you filter all or you filter none by acl. So when server tells "you can't view this whole kv tree due to lacking permissions" that's the semantics of 403 http error.
If there are cases when you still has some nonfiltered data left and you need to use it somehow then http 200 with flag is more appropriate.

PR #11569, opened 2021-11-12
Groundwork for exposing when queries are filtered by ACLs

boxofrad:
This PR lays the groundwork for exposing whether a query's results were filtered by enforcing ACLs. This allows us to indicate in the UI that the user may be unable to access certain resources because of their permissions (rather than them not existing) and should also ease debugging of ACL-related issues in mesh configuration (see https://github.com/hashicorp/consul/issues/11276).

We already expose this information in the UI's service topology endpoint, but this change aims to expose it more generally on the `QueryMeta` type and via the `X-Consul-Results-Filtered-By-ACLs` HTTP header.

In the spirit of splitting the work into smaller, more easily-reviewable chunks; support for this flag/header will be added to the endpoints themselves in subsequent PRs.

If you review the commit history, you'll see that I originally called the flag `FilteredByACLs`. Unfortunately, due to the way in which `QueryMeta` is embedded into query structs, this caused encoding collisions with a field of the same name on the `IndexedServiceTopology` struct — as this is an "internal" endpoint, we could consider renaming that field instead, but the shorter name doesn't seem a compelling enough reason to break backwards compatability.


agent/http_test.go
```diff
@@ -264,6 +264,22 @@ func TestSetKnownLeader(t *testing.T) {
 	}
 }
 
+func TestSetFilteredByACLs(t *testing.T) {
+	t.Parallel()
+	resp := httptest.NewRecorder()
+	setResultsFilteredByACLs(resp, true)
+	header := resp.Header().Get("X-Consul-Results-Filtered-By-ACLs")
```
FFMMM, line 271: nit picking but how about `X-Consul-Count-Filtered-By-ACLs` / `X-Consul-Count-Results-Filtered-By-ACLs` ? I know it's a bit on the long side

and should it be `ACL` ?

proto/pbcommon/common.proto
```diff
@@ -150,6 +150,14 @@ message QueryMeta  {
    // Having `discovery_max_stale` on the agent can affect whether
    // the request was served by a leader.
    string ConsistencyLevel = 4;
+
+   // Reserved for NotModified and Backend.
+   reserved 5, 6;
```
FFMMM, line 155: just out of curiosity -- how did you come across this?

FFMMM:
LGTM overall! 💯  -- thanks for working on this

will look over the endpoint PRs and come back here at the end for a final 👍🏼

proto/pbcommon/common.proto
boxofrad: I spotted that these fields were present [in `structs.go`](https://github.com/hashicorp/consul/blob/db5ee0e4d21c039a990889fcc7d59ec6bf7d4673/agent/structs/structs.go#L380-L387) but not here (when adding `ResultsFilteredByACLs`) so I decided to reserved their spaces so that we don't end up with inconsistent ordering if we add them to the protobuf message in the future.

agent/consul/rpc.go
```diff
@@ -968,6 +968,32 @@ RUN_QUERY:
...
+	//	  custom policies, or set namespace default policies. As these scenarios
+	//	  are less common and this flag is a best-effort UX improvement, we think
+	//	  the trade-off for reduced complexity is acceptable.
+	//
+	//	* It's safe to assume that the token has already been validated (and just
+	//	  check that it's not blank) at this point, because ResultsFilteredByACLs
+	//	  is only set to true when applying the already-resolved token's policies.
+	if queryOpts.GetToken() == "" {
+		queryMeta.SetResultsFilteredByACLs(false)
+	}
```
dnephin, line 995: This `blockingQuery` function does handle most "list" RPCs, however I think there are a few special cases that don't use `blockingQuery`, where we'll want to apply this same logic. Streaming is one case, `Txn.Read`, `PreparedQuery.Execute`, and `PreparedQuery.Explain` are a few more. I found some of these by looking at callers to `Server.setQueryMeta`.

Generally I also try to avoid large inline comments in functions, as it makes the code harder to read (although definitely the Consul code does not follow this convention well). Instead of a large inline comment I find a separate function with a godoc string on it works better as it makes the scope more explicit.

I think both of these would be addressed by making a new function for this operation, and moving the comment to be a godoc string for that function, and calling it here.

dnephin (commented):
Looks good! left one suggestion bellow

agent/consul/rpc.go
boxofrad: Good call! I've extracted this into a method that is called as part of `setQueryMeta`, so those special cases should be covered.

As for streaming, it looks like ACL-filtering is done as part of the `Subscribe` endpoint:
https://github.com/hashicorp/consul/blob/3666401ae3ae924b1d1acd9261f5e187c610ac07/agent/rpc/subscribe/subscribe.go#L78-L80

And the event payload doesn't seem to have an embedded `QueryMeta` struct, so I'm not 100% on how it should work, What do you think of deferring this for now?

dnephin: Ya, good call. It's probably less relevant for streaming.

agent/consul/rpc.go
```diff
@@ -1043,3 +1048,31 @@ func (s *Server) consistentRead() error {
...
+//	  account for the fact that operators can modify the anonymous token with
+//	  custom policies, or set namespace default policies. As these scenarios
+//	  are less common and this flag is a best-effort UX improvement, we think
+//	  the trade-off for reduced complexity is acceptable.
+//
+//	* This method assumes that the given token has already been validated (and
+//	  will only check whether it is blank or not). It's a safe assumption because
+//	  ResultsFilteredByACLs is only set to try when applying the already-resolved
+//	  token's policies.
+func (s *Server) maskResultsFilteredByACLs(token string, meta structs.QueryMetaCompat) {
```
kisunji, line 1074: Is there any particular reason for making this a method off `(s *Server)`?

agent/consul/txn_endpoint.go
```diff
@@ -204,5 +203,9 @@ func (t *Txn) Read(args *structs.TxnReadRequest, reply *structs.TxnReadResponse)
 	state := t.srv.fsm.State()
 	reply.Results, reply.Errors = state.TxnRO(args.Ops)
 	reply.Results = FilterTxnResults(authz, reply.Results)
+
+	// We have to do this ourselves since we are not doing a blocking RPC.
```
kisunji, line 207: Similar comment duplicate as previously mentioned

kisunji (approved):
Minor comments but looks good!

agent/consul/prepared_query_endpoint.go
```diff
@@ -383,6 +382,9 @@ func (p *PreparedQuery) Execute(args *structs.PreparedQueryExecuteRequest,
 	// might not be worth the code complexity and behavior differences,
 	// though, since this is essentially a misconfiguration.
 
+	// We have to do this ourselves since we are not doing a blocking RPC.
```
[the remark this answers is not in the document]
boxofrad: Yeah, I decided to copy this as _both_ `setQueryMeta` and `consistentRead` are usually done by `Server.blockingQuery` (but they had to be split up in this endpoint for ordering reasons).

agent/consul/rpc.go
```diff
@@ -938,8 +938,6 @@ func (s *Server) blockingQuery(queryOpts structs.QueryOptionsCompat, queryMeta s
 
 RUN_QUERY:
 	// Setup blocking loop
```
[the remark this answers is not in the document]
boxofrad: I _think_ it still applies 😄 decided not to remove to keep the diff small.

agent/consul/rpc.go
boxofrad: Nah, not really. Do you have a preference for making it a bare function instead?

kisunji: I'd personally prefer it be a bare function since methods add to noise if I ever try to autocomplete something on (*Server), but I think the signature also misleads you into thinking the server is somehow involved in the function.

boxofrad: Cool, makes sense to me. Have made it a function 🙇🏻

dnephin (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/515351.

PR #11573, opened 2021-11-15
Support auth methods for the vault connect CA provider

ishustava:
## Changes proposed
- Support vault auth methods for the Vault connect CA provider
- Rotate the token (re-authenticate to vault using auth method) when the token can no longer be renewed

## Notes
- The unit tests only test auth methods that don't require an external system (like aws or kubernetes)
- I've tested it manually on Kubernetes using the helm chart

### DIFF
--- a/.changelog/11573.txt
+++ b/.changelog/11573.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+connect: Support Vault auth methods for the Connect CA Vault provider.
+```
--- a/.changelog/11573.txt
+++ b/.changelog/11573.txt
@@ -1,3 +1,5 @@
 ```release-note:improvement
-connect: Support Vault auth methods for the Connect CA Vault provider.
+connect: Support Vault auth methods for the Connect CA Vault provider. Currently, we support any non-deprecated auth methods
+the latest version of Vault supports (v1.8.5), which include AppRole, AliCloud, AWS, Azure, Cloud Foundry, GitHub, Google Cloud,
+JWT/OIDC, Kerberos, Kubernetes, LDAP, Oracle Cloud Infrastructure, Okta, Radius, TLS Certificates, and Username & Password.
 ```
--- a/agent/connect/ca/provider_vault.go
+++ b/agent/connect/ca/provider_vault.go
@@ -8,9 +8,11 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"os"
 	"strings"
 	"time"
 
+	"github.com/hashicorp/consul/lib/decode"
 	"github.com/hashicorp/go-hclog"
 	vaultapi "github.com/hashicorp/vault/api"
 	"github.com/mitchellh/mapstructure"
@@ -19,7 +21,23 @@ import (
 	"github.com/hashicorp/consul/agent/structs"
 )
 
-const VaultCALeafCertRole = "leaf-cert"
+const (
+	VaultCALeafCertRole = "leaf-cert"
+
+	defaultK8SServiceAccountTokenPath = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+)
+
+const (
+	VaultAuthMethodTypeKubernetes = "kubernetes"
+
+	VaultAuthMethodTypeLDAP     = "ldap"
+	VaultAuthMethodTypeUserpass = "userpass"
+	VaultAuthMethodTypeOkta     = "okta"
+	VaultAuthMethodTypeRadius   = "radius"
+	VaultAuthMethodTypeOCI      = "oci"
+
+	VaultAuthMethodTypeToken = "token"
+)
 
 var ErrBackendNotMounted = fmt.Errorf("backend not mounted")
 var ErrBackendNotInitialized = fmt.Errorf("backend not initialized")
@@ -74,6 +92,13 @@ func (v *VaultProvider) Configure(cfg ProviderConfig) error {
 		return err
 	}
 
+	if config.AuthMethod != nil {
+		loginResp, err := vaultLogin(client, config.AuthMethod)
+		if err != nil {
+			return err
+		}
+		config.Token = loginResp.Auth.ClientToken
+	}
 	client.SetToken(config.Token)
 
 	// We don't want to set the namespace if it's empty to prevent potential
@@ -94,7 +119,7 @@ func (v *VaultProvider) Configure(cfg ProviderConfig) error {
 	if err != nil {
 		return err
 	} else if secret == nil {
-		return fmt.Errorf("Could not look up Vault provider token: not found")
+		return fmt.Errorf("could not look up Vault provider token: not found")
 	}
 	var token struct {
 		Renewable bool
@@ -105,7 +130,7 @@ func (v *VaultProvider) Configure(cfg ProviderConfig) error {
 	}
 
 	// Set up a renewer to renew the token automatically, if supported.
-	if token.Renewable {
+	if token.Renewable || config.AuthMethod != nil {
 		lifetimeWatcher, err := client.NewLifetimeWatcher(&vaultapi.LifetimeWatcherInput{
 			Secret: &vaultapi.Secret{
 				Auth: &vaultapi.SecretAuth{
@@ -118,10 +143,10 @@ func (v *VaultProvider) Configure(cfg ProviderConfig) error {
 			RenewBehavior: vaultapi.RenewBehaviorIgnoreErrors,
 		})
 		if err != nil {
-			return fmt.Errorf("Error beginning Vault provider token renewal: %v", err)
+			return fmt.Errorf("error beginning Vault provider token renewal: %v", err)
 		}
 
-		ctx, cancel := context.WithCancel(context.TODO())
+		ctx, cancel := context.WithCancel(context.Background())
 		v.shutdown = cancel
 		go v.renewToken(ctx, lifetimeWatcher)
 	}
@@ -129,7 +154,9 @@ func (v *VaultProvider) Configure(cfg ProviderConfig) error {
 	return nil
 }
 
-// renewToken uses a vaultapi.Renewer to repeatedly renew our token's lease.
+// renewToken uses a vaultapi.LifetimeWatcher to repeatedly renew our token's lease.
+// If the token can no longer be renewed and auth method is set,
+// it will re-authenticate to Vault using the auth method and restart the renewer with the new token.
 func (v *VaultProvider) renewToken(ctx context.Context, watcher *vaultapi.LifetimeWatcher) {
 	go watcher.Start()
 	defer watcher.Stop()
@@ -144,7 +171,35 @@ func (v *VaultProvider) renewToken(ctx context.Context, watcher *vaultapi.Lifeti
 				v.logger.Error("Error renewing token for Vault provider", "error", err)
 			}
 
-			// Watcher routine has finished, so start it again.
+			// If the watcher has exited and auth method is enabled,
+			// re-authenticate using the auth method and set up a new watcher.
+			if v.config.AuthMethod != nil {
+				// Login to Vault using the auth method.
+				loginResp, err := vaultLogin(v.client, v.config.AuthMethod)
+				if err != nil {
+					v.logger.Error("Error login in to Vault with %q auth method", v.config.AuthMethod.Type)
+					// Restart the watcher.
+					go watcher.Start()
+					continue
+				}
+
+				// Set the new token for the vault client.
+				v.client.SetToken(loginResp.Auth.ClientToken)
+				v.logger.Info("Successfully re-authenticated with Vault using auth method")
+
+				// Start the new watcher for the new token.
+				watcher, err = v.client.NewLifetimeWatcher(&vaultapi.LifetimeWatcherInput{
+					Secret:        loginResp,
+					RenewBehavior: vaultapi.RenewBehaviorIgnoreErrors,
+				})
+				if err != nil {
+					v.logger.Error("Error starting token renewal process")
+					go watcher.Start()
+					continue
+				}
+			}
+
+			// Restart the watcher.
 			go watcher.Start()
 
 		case <-watcher.RenewCh():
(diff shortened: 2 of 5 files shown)


website/content/docs/connect/ca/vault.mdx
```diff
@@ -88,12 +88,31 @@ The configuration options are listed below.
...
+
+   - `MountPath`/ `mount_path` (`string: <AuthMethod.Type>`) - The mount path of the auth method.
+      If not provided the auth method type will be used as the mount path.
+
+   - `Params`/`params` (`map: nil`) - The parameters to configure the auth method. Please see
+    [Vault Auth Methods](https://www.vaultproject.io/docs/auth) for information on how to configure the
+    auth method you wish to use. If using the Kuberentes auth method,
+    Consul will try to read the service account token from the
+    default mount path `/var/run/secrets/kubernetes.io/serviceaccount/token` if the `jwt` parameter
+    is not provided.
```
karl-cardenas-coding, line 114:
```suggestion
    Consul will read the service account token from the
    default mount path `/var/run/secrets/kubernetes.io/serviceaccount/token` if the `jwt` parameter
    is not provided.
```
Minor language cleanup

karl-cardenas-coding (changes requested):
I provided some docs feedback and suggestions.

agent/connect/ca/provider_vault.go
```diff
@@ -613,8 +667,12 @@ func ParseVaultCAConfig(raw map[string]interface{}) (*structs.VaultCAProviderCon
...
 	}
 
-	if config.Token == "" {
-		return nil, fmt.Errorf("must provide a Vault token")
+	if config.Token == "" && config.AuthMethod == nil {
+		return nil, fmt.Errorf("must provide a Vault token or configure a Vault auth method")
+	}
+
+	if config.Token != "" && config.AuthMethod != nil {
+		return nil, fmt.Errorf("only one of Vault token or auth method can be provided, but not both")
```
FFMMM, line 675: `... or Vault auth method ... `?

agent/connect/ca/provider_vault.go
```diff
@@ -19,7 +21,19 @@ import (
...
+const (
+	VaultCALeafCertRole = "leaf-cert"
+
+	VaultAuthMethodTypeKubernetes = "kubernetes"
+	VaultAuthMethodTypeLDAP       = "ldap"
+	VaultAuthMethodTypeUserpass   = "userpass"
+	VaultAuthMethodTypeOkta       = "okta"
+	VaultAuthMethodTypeRadius     = "radius"
+	VaultAuthMethodTypeOCI        = "oci"
+	VaultAuthMethodTypeToken      = "token"
```
FFMMM, line 33: It would have been nice if there was some reflection in the vault api or some sot of enum we could've latched onto here.

agent/connect/ca/provider_vault.go
```diff
@@ -94,7 +115,7 @@ func (v *VaultProvider) Configure(cfg ProviderConfig) error {
 	if err != nil {
 		return err
 	} else if secret == nil {
-		return fmt.Errorf("Could not look up Vault provider token: not found")
+		return fmt.Errorf("could not look up Vault provider token: not found")
```
FFMMM, line 118: here and on line 142:

Generally, I try not to modify err strings on the offchance that any callees actually do string processing. 

Here I think it's pretty safe to do it (as opposed to out of some API endpoint) but just wanted to call it out.

agent/connect/ca/provider_vault.go
```diff
@@ -144,7 +167,35 @@ func (v *VaultProvider) renewToken(ctx context.Context, watcher *vaultapi.Lifeti
...
+
+				// Start the new watcher for the new token.
+				watcher, err = v.client.NewLifetimeWatcher(&vaultapi.LifetimeWatcherInput{
+					Secret:        loginResp,
+					RenewBehavior: vaultapi.RenewBehaviorIgnoreErrors,
+				})
+				if err != nil {
+					v.logger.Error("Error starting token renewal process")
+					go watcher.Start()
+					continue
```
FFMMM, line 194: since we start a new watch here should we also try to find a place to call `.Stop()`? not sure if we can just add it w a `defer` keyword.

agent/connect/ca/provider_vault.go
```diff
@@ -637,3 +695,64 @@ func ParseVaultCAConfig(raw map[string]interface{}) (*structs.VaultCAProviderCon
...
+		return nil, err
+	}
+
+	resp, err := client.Logical().Write(loginPath, authMethod.Params)
+	if err != nil {
+		return nil, err
+	}
+	if resp == nil || resp.Auth == nil || resp.Auth.ClientToken == "" {
+		return nil, fmt.Errorf("login response did not return client token")
+	}
```
FFMMM, line 711: curious if we could surface any err codes from vault api if present

agent/connect/ca/provider_vault.go
```diff
@@ -637,3 +695,64 @@ func ParseVaultCAConfig(raw map[string]interface{}) (*structs.VaultCAProviderCon
...
+		} else {
+			return "", fmt.Errorf("failed to get 'role' from auth method params")
+		}
+	case VaultAuthMethodTypeToken:
+		return "", fmt.Errorf("'token' auth method is not supported via auth method configuration; " +
+			"please provide the token with the 'token' parameter in the CA configuration")
+	// The rest of the auth methods use auth/<auth method path> login API path.
+	default:
+		loginPath = fmt.Sprintf("auth/%s/login", authMethod.MountPath)
+	}
```
FFMMM, line 755: this may just me being confused here, but if the `authMethod.Type` was `ldap12345` (something not in our constants), we hit the default here, right?

In that case, we will err out down the code execution path when we call the vault api and that should still be caught in some err handling 🤔 

-- 

Should we, instead, define all the supported `VaultAuthMethod*`s in their own case and err out on `default` as an unrecognized option?

.changelog/11573.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+connect: Support Vault auth methods for the Connect CA Vault provider.
```
FFMMM, line 2: small nit -- can we call out which Vault auth methods are now supported?

IIUC, today, we would support all thru this PR. 

Thinking that as Vault may or may not add more methods, it'd be useful to look thru release notes to trace features additions, deprecations, etc

agent/connect/ca/provider_vault_test.go
```diff
@@ -507,6 +553,138 @@ func TestVaultProvider_Cleanup(t *testing.T) {
 	})
 }
 
+func TestVaultProvider_ConfigureWithAuthMethod(t *testing.T) {
+
+	SkipIfVaultNotPresent(t)
+
+	cases := []struct {
```
FFMMM, line 560: Curious if we could add some testing for the other vault auth methods as well?

website/content/docs/connect/ca/vault.mdx
```diff
@@ -88,12 +88,31 @@ The configuration options are listed below.
...
-  duration has expired.
+  duration has expired. You must either provide a token or configure an auth method below.
+
+- `AuthMethod` / `auth_method` (`map: nil`) - Vault auth method to use for logging in to Vault.
+   Please see [Vault Auth Methods](https://www.vaultproject.io/docs/auth) for more information
+   on how to configure individual auth methods. If auth method is provided, Consul will obtain a
+   a new token from Vault when the token can no longer be renewed.
+   You must either configure the auth method or provide a token above.
+
+   - `Type`/ `type` (`string: ""`) - The type of Vault auth method, e.g. `kubernetes` or `approle`.
```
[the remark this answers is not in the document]
ishustava: Nope these are just an example. i can remove this to avoid confusion.

agent/connect/ca/provider_vault.go
ishustava: Unfortunately, they don't have it in the API package.

agent/connect/ca/provider_vault.go
ishustava: There's a defer at the top of this function.

agent/connect/ca/provider_vault.go
ishustava: 🤔 In this case there aren't any and this is taken from the [code example](https://www.vaultproject.io/docs/auth/kubernetes#code-example) in the docs.

agent/connect/ca/provider_vault.go
ishustava: >his may just me being confused here, but if the authMethod.Type was ldap12345 (something not in our constants), we hit the default here, right?

right. Yep we will rely on vault there.

>Should we, instead, define all the supported VaultAuthMethod*s in their own case and err out on default as an unrecognized option?

I was on the fence on whether we should do that or if we should instead allow for some future auth method to be supported. This will only be the case if the auth method is using the `auth/<auth method name>/login` path though. In the end, I decided to keep it open for that future case to decrease the maintenance burden here a bit but definitely open to changing it like you suggest. WDYT?

agent/connect/ca/provider_vault_test.go
ishustava: Testing other auth methods is a bit tricky since we'd have to stub out the response from whatever the external system vault auth method integrates with, like AWS, GCP or Kubernetes. I'm not sure if there's value in doing that since at that point we're testing vault itself.

website/content/docs/connect/ca/vault.mdx
```diff
@@ -88,12 +88,31 @@ The configuration options are listed below.
...
   paths configured. In Consul 1.8.5 and later, if the token has the [renewable](https://www.vaultproject.io/api-docs/auth/token#renewable)
   flag set, Consul will attempt to renew its lease periodically after half the
-  duration has expired.
+  duration has expired. You must either provide a token or configure an auth method below.
+
+- `AuthMethod` / `auth_method` (`map: nil`) - Vault auth method to use for logging in to Vault.
+   Please see [Vault Auth Methods](https://www.vaultproject.io/docs/auth) for more information
+   on how to configure individual auth methods. If auth method is provided, Consul will obtain a
+   a new token from Vault when the token can no longer be renewed.
+   You must either configure the auth method or provide a token above.
```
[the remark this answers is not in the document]
ishustava: Added [preview](https://consul-1ukpzlbnz-hashicorp.vercel.app/docs/connect/ca/vault). Let me know what you think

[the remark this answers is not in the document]
karl-cardenas-coding: @ishustava Looks good but we probably just need one of them. I think we can remove the 2nd warning box that is lower on the page.

karl-cardenas-coding (approved):
LGTM from a document perspective. We just need to remove the 2nd warning box 👍🏼

agent/connect/ca/provider_vault.go
kschoche: This threw me off as well, is it possible to have a condition here where Stop() is called through the defer and then we do another Start()? Seems unlikely but wondering if there are some nuances I'm not aware of here.

agent/connect/ca/provider_vault_test.go
```diff
@@ -507,6 +553,138 @@ func TestVaultProvider_Cleanup(t *testing.T) {
...
+					},
+				}
+			},
+		},
+		{
+			authMethodType: "approle",
+			configureAuthMethodFunc: func(t *testing.T, vaultClient *vaultapi.Client) map[string]interface{} {
+				_, err := vaultClient.Logical().Write("auth/approle/role/my-role", nil)
+				require.NoError(t, err)
+				resp, err := vaultClient.Logical().Read("auth/approle/role/my-role/role-id")
```
kschoche, line 584: I've learned something new by reviewing this :) looks great!

kschoche (approved):
This looks great @ishustava !!

agent/connect/ca/provider_vault.go
FFMMM: Thanks for the context here! 

> WDYT?

Hm, I have a slight preference towards doing *some validation* somewhere. If that make sense.

So if we treated `default` as an error case, it'd also act a bit as validation in my mind (that the auth method provided is something we can deal with).

That said, this is not a stick point on my end right now! So if you wanted to add any validation here or a layer above, I can put some 👀 on it!

agent/connect/ca/provider_vault_test.go
FFMMM: Yeah I figured that this was the case. 

> that since at that point we're testing vault itself.

That's a fair point -- I think we gain the confidence that auth methods work from the vault testing 🤔

agent/connect/ca/provider_vault.go
FFMMM: > There's a defer at the top of this function.

Oh yes I think I got confused so I went and read up on using defers with pointer receivers. I think this is fine!

https://blog.learngoprogramming.com/gotchas-of-defer-in-go-1-8d070894cb01

agent/connect/ca/provider_vault.go
FFMMM: Oh I see, `resp` there is a `Secret*`.

Hmm that makes sense!

FFMMM (approved):
This LGTM overall! 🚀 

If you had time, I think it be worth having some validation of the auth method (ie this is supported, this is not) from the vault CA provider side but other than that no other concerns!

Thanks for being open to some of my feedback 💯 !!

agent/connect/ca/provider_vault.go
ishustava: > is it possible to have a condition here where Stop() is called through the defer and then we do another Start()?

I don't think that's possible since defer is called when the function returns, so I don't think it's possible to make another call to Start after that. Hope I understood your question correctly!

agent/connect/ca/provider_vault.go
ishustava: That makes sense! I've changed it to error out if the auth method is not supported and added a list of all currently supported methods. I've also added a test for this function to verify that it does the right thing. 

Let me know if that's what you had in mind @FFMMM!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/505952.

PR #11578, opened 2021-11-16
ui: Fix brand coloring for inline-code plus docs

johncowen:
I noticed the coloring of our `%inline-code` style wasn't picking up the Consul brand color. So this PR fixes that up by also moving `--brand` to use a `--tone-` prefix and require `rgb` like all our other colors, which fixes that up.

Before:

Screenshot 2021-11-16 at 15 29 52

After:

Screenshot 2021-11-16 at 15 29 25

I also took the opportunity to add some [documentation explaining all of the `tone`/`rgb` thing plus all of our colors swatches](https://consul-ui-staging-3zorctduu-hashicorp.vercel.app/ui/docs/styles/base/color). I made these easier to copypasta using our `with-copyable` modifier. So you can click on these swatches to get the right thing to paste.

While I was there I did the same for all our [icons](https://consul-ui-staging-3zorctduu-hashicorp.vercel.app/ui/docs/styles/base/icons) and [typography](https://consul-ui-staging-3zorctduu-hashicorp.vercel.app/ui/docs/styles/base/typography), which makes a good reference/document for when we need to tweak all this.


ui/packages/consul-ui/app/styles/base/color/README.mdx
```diff
@@ -0,0 +1,81 @@
...
+1. We get around to eventually adding semantic colors names (or adding something else) in which case these docs should be updated.
+2. You are absolutely sure the color you want should never be changed by a theme. These should be very very few and far between so even if we get that wrong it should be easy to update.
+
+`tones` should always using the following form:
+
+```css
+property: rgb(var(--tone-gray-500));
+```
+
+In other words always use `rgb` this gives us the flexibility to easily use thee variables with alpha if we ever need to.
```
evrowe, line 15: typo: `thee` should be `these` (i think)

ui/packages/consul-ui/app/styles/base/color/README.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+```css
+property: rgb(var(--tone-gray-500) / 50);
+```
+
+Currently, alphas are mainly used for shadows. Please consider whether you should be using an alpha before using an alpha'ed color.
```
evrowe, line 21: Good callout! Should we explicitly mention this is because it's important to avoid the use of alpha to simulate a tint or shade (especially because the effects can be unpredictable against backgrounds other than white)?

ui/packages/consul-ui/app/styles/base/color/README.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+Lastly, there is currently one non-color alias group which is `--tone-brand` which is a group of our current brand color.
```
evrowe, line 23: typo(?): `brand color` >> `brand colors`

ui/packages/consul-ui/app/styles/base/color/README.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+
+## %theme-light (default)
+
+```hbs preview-template
+<ul
+  {{css-props (set this 'tones')
```
evrowe, line 30: Ah ha, so this is what that modifier does. Cool 😎

evrowe (changes requested):
Just need to address the minor typos in docs and should be good to ship 👍🏻

ui/packages/consul-ui/app/styles/base/color/README.mdx
johncowen:
```suggestion
Currently, alphas are mainly used for shadows. Please consider whether you should be using an alpha before using an alpha'ed color, you probably shouldn't be using an alpha, please use our collection of `--tones` instead for tints/shades of colors. If you do use one - you probably shouldn't :) , also expect it to be picked up in code review, so consider highlighting it in a self comment as you PR.
```

---

I'd actually like to get a linter involved a lot more on these things. We should be able to lint for alpha's and use ignores for the "once in a blue moon" occasions. With that and code review we have a nice balance of flexibility on the very rare occasions we need it and making it hard to do the wrong thing.

ui/packages/consul-ui/app/styles/base/color/README.mdx
johncowen:
```suggestion
In other words always use `rgb` this gives us the flexibility to easily use these variables with alpha if we ever need to.
```

ui/packages/consul-ui/app/styles/base/color/README.mdx
johncowen:
```suggestion
Lastly, there is currently one non-color alias group which is `--tone-brand` which is a group of our current brand colors.
```

ui/packages/consul-ui/app/styles/base/color/README.mdx
johncowen: 😸 well kinda, tbh this one started out as a "only give me the variables for this element" - the important bit being "for this element". This one turned into getting things from the root element (not the modified element) so 'theoretically' shouldn't be a modifier - in the scheme of things (and as its only for documentation purposes) I just left it as is. I may move it to be a helper at some point but no biggie really (unless we use it in app code)

I always decide on a modifier if ever the "I just need a reference to the DOM element" slips into my head. The below `with-copyable` is probably a better example.

ui/packages/consul-ui/app/styles/base/color/README.mdx
evrowe: > I'd actually like to get a linter involved a lot more on these things.

💯 for tooling on stuff like this.

Regarding the suggested change, if we are going to make adjustments to this text I'd like to focus on the "why" somebody should avoid using alphas here rather than repeatedly saying "you shouldn't do this" with increasing emphasis.

If it were me, I'd probably go with something like:

"Currently, alphas are mainly used for shadows. Please avoid the use of alphas for adjusting hues as they can produce unpredictable results. We should be utilizing our pre-defined color palette or adding new colors to it with proper RGB values. If you do need to add a color with alpha, please include a code comment or a self-comment explaining the use case."

johncowen:
```suggestion
Currently, alphas are mainly used for shadows. Please avoid the use of alphas for adjusting hues as they can produce unpredictable results. We should be utilizing our pre-defined color palette or adding new colors to it with proper RGB values. If you do need to add a color with alpha, please include a code comment or a self-comment explaining the use case.
```

johncowen:
K those are in @evrowe gimme a yell/ 👍 when you are around and we can get this one in!

evrowe (approved):
🚀

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/507889.

Issue #11347, opened 2021-10-18
ca: improve the data model


dnephin:
The Connect CA system has three core structs that are used in many places as part of both its data model and RPC request/response.

* `IndexedCARoots` - the TrustDomain/ClusterID, ActiveRootID, and all the `CARoot` (active and not, which also include intermediates and other things, see below)
* `CARoot` - the root certificate, but also any intermediate certs, private and signing key details, root expiry time, `ExternalTrustDomain`.
* `CAConfiguration` - provider name, an opaque map of configuration for the provider, `ClusterID` (which can't be set by the user), `State` (which can only be initialized once 

While working on the CA system I noticed a few problems with this model which seem like they may be prone to cause bugs:

1. we manually reconstruct a `CARoot` to return it to the user, and as part of this we attempt to remove "sensitive" data: https://github.com/hashicorp/consul/blob/v1.10.3/agent/consul/server_connect.go#L102-L135. Since the fields still exist, it seems like it would be pretty easy to leak the wrong data. If we had a response struct with only the relevant fields that would be harder to get wrong.
2. The `ClusterID` is not actually settable by the user, but it appears in the `CAConfiguration` struct that is submitted by the user https://github.com/hashicorp/consul/blob/v1.10.3/agent/consul/leader_connect_ca.go#L756-L757.
3. In some places we call it `ClusterID`, and in others we call it `TrustDomain` (or `ExternalTrustDomain`), and yet these are all the same value.

(additional issues may be added to this list in the future)

I believe we could address many of these problems by:
1. use separate types for request/response and memdb storage. That way each type will only include the fields that are relevant for how the type is being used.
2. normalize the names of fields so they are consistent
3. we know the structure of the opaque config will include `CommonCAProviderConfig` and one of the provider config structs. We could parse this into a normalized structure earlier instead of dealing with opaque configuration in multiple places.

PR #11589, opened 2021-11-16
ca: small cleanup of SpiffeIDSigningForCluster args

dnephin:
Slightly related to #11347

This was some cleanup I found sitting on a branch. I did this while working on #11514, and I'm opening a PR now so I don't lose the changes.

The idea here is that we only use one field from this struct, and it seems very likely that we'll want to store this ClusterID somewhere else (not on this struct), so by changing this functional signature now it'll be easier to make the larger change later.

This also helps simplify the `ConsulProvider`, because the provider no longer needs to get all `State`, it only needs the provider specific state. I had another commit which changed the interface to make that explicit, but I dropped it for now as it wasn't necessary.

### DIFF
--- a/agent/connect/ca/provider_aws.go
+++ b/agent/connect/ca/provider_aws.go
@@ -603,7 +603,7 @@ func (a *AWSProvider) Sign(csr *x509.CertificateRequest) (string, error) {
 
 // SignIntermediate implements Provider
 func (a *AWSProvider) SignIntermediate(csr *x509.CertificateRequest) (string, error) {
-	err := validateSignIntermediate(csr, &connect.SpiffeIDSigning{ClusterID: a.clusterID, Domain: "consul"})
+	err := validateSignIntermediate(csr, connect.SpiffeIDSigningForCluster(a.clusterID))
 	if err != nil {
 		return "", err
 	}
--- a/agent/connect/ca/provider_consul.go
+++ b/agent/connect/ca/provider_consul.go
@@ -617,19 +617,13 @@ func (c *ConsulProvider) incrementAndGetNextSerialNumber() (uint64, error) {
 
 // generateCA makes a new root CA using the current private key
 func (c *ConsulProvider) generateCA(privateKey string, sn uint64, rootCertTTL time.Duration) (string, error) {
-	stateStore := c.Delegate.State()
-	_, config, err := stateStore.CAConfig(nil)
-	if err != nil {
-		return "", err
-	}
-
 	privKey, err := connect.ParseSigner(privateKey)
 	if err != nil {
 		return "", fmt.Errorf("error parsing private key %q: %s", privateKey, err)
 	}
 
 	// The URI (SPIFFE compatible) for the cert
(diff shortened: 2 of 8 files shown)


agent/connect/ca/provider_aws.go
```diff
@@ -603,7 +603,7 @@ func (a *AWSProvider) Sign(csr *x509.CertificateRequest) (string, error) {
 
 // SignIntermediate implements Provider
 func (a *AWSProvider) SignIntermediate(csr *x509.CertificateRequest) (string, error) {
-	err := validateSignIntermediate(csr, &connect.SpiffeIDSigning{ClusterID: a.clusterID, Domain: "consul"})
```
dnephin, line 606: The other two providers were using `connect.SpiffeIDSigningForCluster` instead of constructing a literal. I figured it would be good to have all 3 providers use this function to show that they do the same thing.

agent/connect/ca/provider_consul.go
```diff
@@ -617,19 +617,13 @@ func (c *ConsulProvider) incrementAndGetNextSerialNumber() (uint64, error) {
...
-		return "", err
-	}
-
 	privKey, err := connect.ParseSigner(privateKey)
 	if err != nil {
 		return "", fmt.Errorf("error parsing private key %q: %s", privateKey, err)
 	}
 
 	// The URI (SPIFFE compatible) for the cert
-	id := connect.SpiffeIDSigningForCluster(config)
```
dnephin, line 632: We already store the `clusterID` on `ConsulProvider` from the call to `Configure`, so we can use that directly instead of fetching the full config from the state store. This removes the only caller to `Delegate.State()` that was using something other than the provider config.

In the future we can reduce the interface here to a call that only returns the provider config.

agent/connect/uri_signing.go
```diff
@@ -64,6 +62,6 @@ func (id SpiffeIDSigning) CanSign(cu CertURI) bool {
 // break all certificate validation. That does mean that DNS prefix might not
 // match the identity URIs and so the trust domain might not actually resolve
 // which we would like but don't actually need.
-func SpiffeIDSigningForCluster(config *structs.CAConfiguration) *SpiffeIDSigning {
-	return &SpiffeIDSigning{ClusterID: config.ClusterID, Domain: "consul"}
+func SpiffeIDSigningForCluster(clusterID string) *SpiffeIDSigning {
```
dnephin, line 65: Generally I do prefer accepting a struct instead of a string arg, but in this case I think the string arg is better because:
1. this `structs.CAConfiguration` is part of a public interface, so its use here is being overloaded
2. the `ClusterID` is not actually part of the CAConfiguration. It's not settable by the user, and will likely be moved off of that struct.

If we ever need more args to `SpiffeIDSigningForCluster` we should create a separate struct for it.

rboyer (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/505041.

PR #11591, opened 2021-11-17
ui: Store the default partition when logging in

johncowen:
Make sure we store the default `Partition` for a users token.

PR #11592, opened 2021-11-17
ui: Upgrades token sourcing related components to Glimmer+docs

johncowen:
Whilst working on https://github.com/hashicorp/consul/pull/11591 I took the opportunity to upgrade both `TokenSource` and `JwtSource` components to use Glimmer, and added working docs for how to use (including changing to a real OIDC provider instead of the default mock one during development).

- [TokenSource Docs](https://consul-ui-staging-ivzjphmps-hashicorp.vercel.app/ui/docs/components/token-source)
- [JwtSource Docs](https://consul-ui-staging-ivzjphmps-hashicorp.vercel.app/ui/docs/components/jwt-source)


ui/packages/consul-ui/app/components/token-source/index.hbs
```diff
@@ -1,44 +1,61 @@
...
-  <Guard @name="isSecret" @cond={{action 'isSecret'}} />
+<StateChart
+  @src={{this.chart}}
+  @initial={{if (eq @type 'oidc') 'provider' 'secret'}}
+as |State Guard Action dispatch state|>
+  <Guard
+    @name="isSecret"
+    @cond={{this.isSecret}}
+  />
   {{!-- This `or` can be completely removed post 1.10 as 1.10 has optional params with default values --}}
```
evrowe, line 9: Looks like the `or` in question is gone, should the comment be removed?

ui/packages/consul-ui/app/components/token-source/index.js
```diff
@@ -1,36 +1,47 @@
...
+        },
       };
-      // TODO: We should probably put the component into idle state
-      this.onchange(e);
-    },
-  },
-});
+    };
+    // TODO: We should probably put the component into idle state
+    if(typeof this.args.onchange === 'function') {
```
evrowe, line 42: Do we ever use Ember Utils (i.e. https://api.emberjs.com/ember/3.28/functions/@ember%2Futils/typeOf) for things like this?

evrowe (approved):
Love seeing more components make the transition to Glimmer, and solid documentation as well. :shipit:

ui/packages/consul-ui/app/components/token-source/index.hbs
johncowen: Oh! Yes 😆 , will sort before I merge, good spot!

ui/packages/consul-ui/app/components/token-source/index.js
johncowen: Nah, we never use `typeOf`, partly due to this:

```javascript
console.log(typeOf(class {})); // function
console.log(typeOf(new (class {}))); // object
console.log(typeOf(EmberObject.extend())); // class
console.log(typeOf(EmberObject.create())); // instance
```

I think quite a few Ember folks don't use this anymore, I'm surprised they haven't deprecated it yet.

FYI we always use the forms:

```javascript
// checking undefined-ness as opposed to just null
typeof thing === 'undefined'
// checking whether something is a function (so callable but also not undefined)
typeof thing === 'function'
// checking for an array, I would guess this method probably didn't exist when ember's typeOf was originally written
Array.isArray(thing)
```

Generally it should be 'reasonably clear' for other primitive values what types they are, and the types should never change, and off the top of my head I don't think we ever do any other types of runtime type checking anywhere. In the past I've been very careful when writing things to make sure I'm pretty strict with this sort of stuff.

Bit of a tangent, but at one point I was running `tsc` on the project to ensure this kind of stuff (but importantly not using Typescript itself), actually theres an RFC about this somewhere and I wouldn't mind picking that up again and getting it committed to the project in some way. Another one to chat through at some point. Will link you the RFC offline.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/505861.

PR #11593, opened 2021-11-17
kv: support `ResultsFilteredByACLs` flag/header

boxofrad:
Adds support for the `ResultsFilteredByACLs` flag, and corresponding `X-Consul-Results-Filtered-By-ACLs` HTTP header (introduced in #11569) to the KV `List` and `ListKeys` endpoints.

### DIFF
--- a/agent/consul/kvs_endpoint.go
+++ b/agent/consul/kvs_endpoint.go
@@ -204,7 +204,10 @@ func (k *KVS) List(args *structs.KeyRequest, reply *structs.IndexedDirEntries) e
 			if err != nil {
 				return err
 			}
+
+			total := len(ent)
 			ent = FilterDirEnt(authz, ent)
+			reply.QueryMeta.ResultsFilteredByACLs = total != len(ent)
 
 			if len(ent) == 0 {
 				// Must provide non-zero index to prevent blocking
@@ -263,7 +266,9 @@ func (k *KVS) ListKeys(args *structs.KeyListRequest, reply *structs.IndexedKeyLi
 				reply.Index = index
 			}
 
+			total := len(entries)
 			entries = FilterDirEnt(authz, entries)
+			reply.QueryMeta.ResultsFilteredByACLs = total != len(entries)
 
 			// Collect the keys from the filtered entries
 			prefixLen := len(args.Prefix)
--- a/agent/consul/kvs_endpoint_test.go
+++ b/agent/consul/kvs_endpoint_test.go
@@ -287,6 +287,9 @@ func TestKVSEndpoint_List(t *testing.T) {
 			t.Fatalf("bad: %v", d)
 		}
 	}
+	if dirent.QueryMeta.ResultsFilteredByACLs {
+		t.Fatal("ResultsFilteredByACLs should not be true")
+	}
 
 	// Try listing a nonexistent prefix
 	getR.Key = "/nope"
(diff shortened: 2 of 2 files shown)


agent/consul/kvs_endpoint.go
```diff
@@ -204,7 +204,10 @@ func (k *KVS) List(args *structs.KeyRequest, reply *structs.IndexedDirEntries) e
 			if err != nil {
 				return err
 			}
+
+			total := len(ent)
 			ent = FilterDirEnt(authz, ent)
+			reply.QueryMeta.ResultsFilteredByACLs = total != len(ent)
```
FFMMM, line 210: [nit] hm, could be a style preference but curious if we could encapsulate the logic here in a func that can be used for `List` and ListKeys`?

agent/consul/kvs_endpoint_test.go
```diff
@@ -475,6 +478,9 @@ func TestKVSEndpoint_List_ACLDeny(t *testing.T) {
 			}
 		}
 	}
+	if !dirent.QueryMeta.ResultsFilteredByACLs {
+		t.Fatal("ResultsFilteredByACLs should be true")
+	}
 }
 
 func TestKVSEndpoint_List_ACLEnableKeyListPolicy(t *testing.T) {
```
FFMMM, line 486: could we add a test here (or somewhere else) to test that `ResultsFilterdByACLs` is set for the `List` endpoint as expected?

agent/consul/kvs_endpoint_test.go
```diff
@@ -287,6 +287,9 @@ func TestKVSEndpoint_List(t *testing.T) {
 			t.Fatalf("bad: %v", d)
 		}
 	}
+	if dirent.QueryMeta.ResultsFilteredByACLs {
+		t.Fatal("ResultsFilteredByACLs should not be true")
+	}
```
FFMMM, line 292: I think this suffices as a test but I was curious if we could actually test for the presence of the header somehow as well.

But this may actually be covered, generally, thru https://github.com/hashicorp/consul/pull/11569 ; need to double check

agent/consul/kvs_endpoint.go
boxofrad: Maybe, yeah! Were you thinking of something like `k.filterDirEnt(&reply)` that does the filtering and sets the flag?

agent/consul/kvs_endpoint_test.go
boxofrad: I think this is covered here:

https://github.com/hashicorp/consul/blob/b5f81dd0c0f8361e986dbb86fde96fe5644f50da/agent/consul/kvs_endpoint_test.go#L481-L483

agent/consul/kvs_endpoint_test.go
boxofrad: Ah yeah, it is covered more generally by the `setMeta` tests in that PR, but we certainly could have better coverage by testing it on the individual HTTP handlers themselves.

On balance, I decided that, aside from the HTTP-only endpoints (in #11610), testing at the RPC level is an acceptable trade-off of value vs maintainance burden.

What do you think? 🙇🏻‍♂️

FFMMM (approved):
🚀 this LGTM! Up to you if you want to tackle that nit!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/515836.

PR #11594, opened 2021-11-17
catalog: support `ResultsFilteredByACLs` flag/header

boxofrad:
Adds support for the `ResultsFilteredByACLs` flag, and corresponding `X-Consul-Results-Filtered-By-ACLs` HTTP header (introduced in #11569) to the catalog endpoints.

### DIFF
--- a/agent/consul/acl.go
+++ b/agent/consul/acl.go
@@ -1315,39 +1315,37 @@ func (f *aclFilter) filterNodeServices(services **structs.NodeServices) bool {
 }
 
 // filterNodeServices is used to filter services on a given node base on ACLs.
-func (f *aclFilter) filterNodeServiceList(services **structs.NodeServiceList) {
-	if services == nil || *services == nil {
-		return
+// Returns true if any elements were removed.
+func (f *aclFilter) filterNodeServiceList(services *structs.NodeServiceList) bool {
+	if services.Node == nil {
+		return false
 	}
 
 	var authzContext acl.AuthorizerContext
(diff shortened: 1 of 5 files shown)


agent/consul/acl.go
```diff
@@ -1265,77 +1272,83 @@ func (f *aclFilter) filterServiceNodes(nodes *structs.ServiceNodes) {
...
-	svcs := (*services).Services
-	modified := false
+	var removed bool
+	svcs := services.Services
 	for i := 0; i < len(svcs); i++ {
 		svc := svcs[i]
 		svc.FillAuthzContext(&authzContext)
 
-		if f.allowNode((*services).Node.Node, &authzContext) && f.allowService(svc.Service, &authzContext) {
+		if f.allowService(svc.Service, &authzContext) {
```
FFMMM, line 1333: wait, why are we not filtering on the Node first as well ?

boxofrad: I think it's safe to remove the check from this loop, because we're already checking node access on [L1322](https://github.com/hashicorp/consul/pull/11594/files/0714fc3260c9c96d69e194d69219d3feaa33507d#diff-e9bfbecd10c1931e8c8607866fcdc011f9701c9369fcbcefcc0187fc6638d0ddR1322) 🤞🏻

kisunji (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/516835.

PR #11610, opened 2021-11-19
agent: support `X-Consul-Results-Filtered-By-ACLs` header in agent-local endpoints

boxofrad:
Adds support for the `X-Consul-Results-Filtered-By-ACLs` HTTP header (introduced in #11569) to the agent-local endpoints.

### DIFF
--- a/agent/acl_oss.go
+++ b/agent/acl_oss.go
@@ -7,8 +7,13 @@ import (
 	"github.com/hashicorp/serf/serf"
 
 	"github.com/hashicorp/consul/acl"
+	"github.com/hashicorp/consul/api"
 )
 
 func serfMemberFillAuthzContext(m *serf.Member, ctx *acl.AuthorizerContext) {
+	// no-op
+}
+
+func agentServiceFillAuthzContext(s *api.AgentService, ctx *acl.AuthorizerContext) {
 	// no-op
 }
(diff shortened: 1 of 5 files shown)


agent/acl_oss.go
```diff
@@ -7,8 +7,13 @@ import (
...
+	"github.com/hashicorp/consul/api"
 )
 
 func serfMemberFillAuthzContext(m *serf.Member, ctx *acl.AuthorizerContext) {
 	// no-op
 }
+
+func agentServiceFillAuthzContext(s *api.AgentService, ctx *acl.AuthorizerContext) {
+	// no-op
+}
```
dnephin, line 19: I thought we generally used a method for this, but I see there is prior art above here. Any idea why we use a function sometimes and a method the rest of the time?

dnephin (approved):

agent/acl_oss.go
boxofrad: Here it's because the type is from the `api` package (above is from `serf`) so it wouldn't be appropriate to put it there.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/516381.

PR #11614, opened 2021-11-21
Fix target : pull_request_target instead of pull_request

mmorel-35:
See https://github.com/actions/labeler#create-workflow

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -1,6 +1,6 @@
 name: "Pull Request Labeler"
 on:
-  pull_request:
+  pull_request_target:
     types: [opened]
 
 jobs:


dnephin:
Thank you for the PR! Can you help me understand the problem this solves? Is it to match the example config?

I think [pull_request_target](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request_target) is a bit risky (see the red box in those docs). From my quick read of that action, right now it does not checkout the code in the repo, but it is nice to limit the access that we grant whenever possible.

My guess is that right now PRs from forks do not get these labels assigned, where as PRs from contributors that have write access to the repo will have the labels assigned. 

I'm not sure how well this labeler has been working for us. I think because of the structure of Consul it can be challenging to write a regex that finds the right labels. I think we'll want to review how effective this action has been for us, before granting additional access.

mmorel-35:
Hi @dnephin,
It's fixing this kind of errors https://github.com/hashicorp/consul/runs/4460352336?check_suite_focus=true

When a PR for a dependency update is created it's not being able to categorize it correctly.

dnephin:
@mmorel-35 Thanks again for the PR, and providing the additional information.

I think despite the examples, using `pull_request_target` here could be a concern. If I've understood this action correctly, it reads the config file from the PR branch, not from the main target branch.

This would mean that anyone could edit the config on their fork branch. They could edit the config to have the bot adds labels  to a PR. Some of our labels perform actions (like running additional test suites). So adding a label to a PR is not always safe for un-authorized users.

I asked a few team members and most of us haven't really seen much benefit from this PR labeling action. The Consul code is not really organized in a way that allows us to use regex to apply the right tags. Worse, I've often found this bot removing the labels I have manually added to a PR.

Given that we don't get much value from the action, and in some cases it is actually hurting us (forces us to repeat adding labels), I suspect we are more likely to remove this action than we are to allow it to run from fork PRs.

I'll leave this PR open for a bit longer for additional comments, but unless there are major objections from the team, I think we will likely close this PR without merging and look to remove the PR labeling action in the future.

markan:
Closing because there were no additional comments or objections.

PR #11616, opened 2021-11-22
event: support `X-Consul-Results-Filtered-By-ACLs` header in list endpoint

boxofrad:
Adds support for the `X-Consul-Results-Filtered-By-ACLs` HTTP header (introduced in #11569) to the event list endpoint.

### DIFF
--- a/agent/event_endpoint.go
+++ b/agent/event_endpoint.go
@@ -127,17 +127,6 @@ RUN_QUERY:
 	// Get the recent events
 	events := s.agent.UserEvents()
 
-	// Filter the events using the ACL, if present
-	for i := 0; i < len(events); i++ {
-		name := events[i].Name
-		if authz.EventRead(name, nil) == acl.Allow {
-			continue
-		}
-		s.agent.logger.Debug("dropping event from result due to ACLs", "event", name)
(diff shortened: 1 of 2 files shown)


agent/event_endpoint.go
```diff
@@ -127,25 +127,44 @@ RUN_QUERY:
...
 
+	// Filter the events if requested
+	if nameFilter != "" {
+		for i := 0; i < len(events); i++ {
+			if events[i].Name != nameFilter {
+				events = append(events[:i], events[i+1:]...)
+				i--
+			}
+		}
+	}
```
FFMMM, line 138: Just curious, why are we doing the name filter before the ACL filter?

boxofrad: See comment below (on L142) 😄 

```Go
// Note: we filter the results with ACLs *after* applying the user-supplied
// name filter, to ensure the filtered-by-acls header we set below does not
// include results that would be filtered out even if the user did have
// permission.
```

FFMMM (approved):
🚢

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/515999.

PR #11604, opened 2021-11-18
ui: Support for SSO with Admin Partitions

johncowen:
Initial MVP of SSO with admin partition support.

sso

If you try the [preview site](https://consul-ui-staging-m9x15lp2b-hashicorp.vercel.app/ui/dc1/services#CONSUL_PARTITIONS_ENABLE=1;CONSUL_NSPACES_ENABLE=1;CONSUL_SSO_ENABLE=1), be aware that our SSO callback URL has the wrong headers set in Vercel, so whilst you can see the entire flow through to our fake 3rd party provider, when you go to log in it tries to download the file instead of just rendering it. We can look into sorting Vercel headers that at some point.

Notes:

- There is some ember > glimmer component upgrading happening here, plus more documentation for the components I'm touching.
- Tabbed interface is coming soon/next. There'll probably be a bit of cleanup/moving around there also, but I wanted to get an MVP of this done and up. I'll PR ontop of this one.

PR #11619, opened 2021-11-22
ui: Tabbed Login with Token or SSO interface

johncowen:
Continuation of https://github.com/hashicorp/consul/pull/11604, now with tabbed interface.

sso

Notes:

- I upgraded our super old, almost the first ember component I wrote, to use glimmer/almost template only. This should use slots/contextual components somehow, but thats a bigger upgrade so I didn't go that far.
- I've been wanting to upgrade the shape of our `StateChart` component for a very long while now, here its very apparent that it would be much better to do this sooner rather than later. I left it as is for now, but there will be a PR coming soon with a slight reshaping of this component.
- Added a `did-upsert` modifier which is a mix of `did-insert/did-update`

Documentation added/amended for all the new things.

There is a better way to do the tabbing somewhere here, but its not quite come together in the PR, mostly as I wanted to do the StateChart re-shaping separately before making some sort of `TabGroup` component. So there are a couple of bits here that don't make complete sense from an isolation viewpoint.


evrowe (approved):
Gave everything a thorough look through, looks good to me.

PR #11620, opened 2021-11-22
query: support `ResultsFilteredByACLs` in query list endpoint

boxofrad:
Adds support for the `ResultsFilteredByACLs` flag, and corresponding `X-Consul-Results-Filtered-By-ACLs` HTTP header (introduced in #11569) to the prepared queries list and execute endpoints.

Note: we only set the `ResultsFilteredByACLs` flag if **named** queries were removed from the result set. This is because un-named queries are treated as ephemeral and can only be listed by a management token.


FFMMM (approved):
🚀

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/517336.

PR #11629, opened 2021-11-23
docs: `X-Consul-Results-Filtered-By-ACLs` header

boxofrad:
📚


website/content/api-docs/index.mdx
```diff
@@ -97,6 +97,15 @@ action if no intention matches.
 
 This is returned even if ACLs are disabled.
 
+## Results Filtered by ACLs
+
+As of Consul 1.11, most list endpoints support an `X-Consul-Results-Filtered-By-ACLs`
+HTTP response header. This indicates that the response contains a partial subset
+of results, because some have been filtered out by ACL policies.
```
FFMMM, line 104: I know it could be really tedious but it may be worth calling out exactly which endpoints support this today/ as of this change.

boxofrad: Yup! I appreciate that "most" is a bit of a [weasel word](https://en.wikipedia.org/wiki/Weasel_word) here 😄 do you know of anywhere else in the docs where we list endpoints like this?

I guess an alternative would be to call it out on the endpoint docs themselves?

kisunji (approved):
LGTM! Alex's comment is certainly a nice to have but we can iterate on docs anytime so I think this is good to go in for now

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/515783.

PR #11632, opened 2021-11-23
txn: support `ResultsFilteredByACLs` flag in `Read` endpoint

boxofrad:
Adds support for the `ResultsFilteredByACLs` flag, and corresponding `X-Consul-Results-Filtered-By-ACLs` HTTP header (introduced in #11569) to the txn `Read` endpoint.

Note: as the `Apply` endpoint's response type doesn't embed `QueryMeta` I didn't add support for the new header there.

### DIFF
--- a/agent/consul/txn_endpoint.go
+++ b/agent/consul/txn_endpoint.go
@@ -194,6 +194,14 @@ func (t *Txn) Read(args *structs.TxnReadRequest, reply *structs.TxnReadResponse)
 	if err != nil {
 		return err
 	}
+
+	// There are currently two different ways we handle permission issues.
+	//
+	// For simple reads such as KVGet and KVGetTree, the txn succeeds but the
+	// offending results are omitted. For more involved operations such as
+	// KVCheckIndex, the txn fails and permission denied errors are returned.
+	//
+	// TODO: Maybe we should unify these, or at least cover it in the docs?
 	reply.Errors = t.preCheck(authz, args.Ops)
 	if len(reply.Errors) > 0 {
 		return nil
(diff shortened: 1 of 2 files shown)


agent/consul/txn_endpoint.go
```diff
@@ -194,6 +194,14 @@ func (t *Txn) Read(args *structs.TxnReadRequest, reply *structs.TxnReadResponse)
...
 		return err
 	}
+
+	// There are currently two different ways we handle permission issues.
+	//
+	// For simple reads such as KVGet and KVGetTree, the txn succeeds but the
+	// offending results are omitted. For more involved operations such as
+	// KVCheckIndex, the txn fails and permission denied errors are returned.
+	//
+	// TODO: Maybe we should unify these, or at least cover it in the docs?
```
boxofrad, line 204: Please keep me honest here, is this right? Have I missed it in the docs? 😅

FFMMM: hm, I couldn't find something obvious on https://www.consul.io/api-docs/txn

Was there another place you had checked?

FFMMM (approved):
🚀 !

agent/consul/txn_endpoint.go
FFMMM: If it comes to it, let's throw an issue in our backlog regarding docs update/ refactor work.

boxofrad: Nah, that's where I looked too 😄 agree let's pop it in the backlog (didn't want to tackle now because it's a breaking change to that API's contract).

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/516494.

PR #11633, opened 2021-11-23
ui: Move nspace CRUD to use the same approach as partitions

johncowen:
This PR moves our Namespace CRUD to use the same approach as the partition CRUD.

This sounds a bit 'backwards' as the end goal here is to add an improved UX to partitions, not namespaces. The reason for doing it this way is that Namespaces already has a type of 'improved UX' CRUD in that it has one to many relationship in the form when saving your namespaces (the end goal for partitions). In moving Namespaces to use the same approach as partitions we:

1. Ensure the new approach works with one-to-many forms.
2. Test the new approach without writing a single test (we already have a bunch of tests for namespaces which are now testing the approach used by both namespaces _and_ partitions)

All in all this sets us up ready to do the improved UX for partition CRUD (and therefore was kind of like a engineering focussed milestone)

Additionally this PR:

- Fixes issue with missing default nspace in the nspace selector (~introduced a PR or 2 back 👀~ https://github.com/hashicorp/consul/pull/11479 )
- In doing when checking to see that things where consistent between the two,  I found a few little minor problems with the Admin Partition CRUD so fixed those up here also.
- Removed the old style Nspace notifications now that this was rebased onto the work from https://github.com/hashicorp/consul/pull/11577 and replaced it with the single non-default delete notifications.
- There are a couple of bits I'm not happy with here but I'm happy to punt for a little longer, mainly due to the fact that they are around testing and/or things that there is a plan to replace over the next few days or so. I'll add some inline comments around those.


ui/packages/consul-nspaces/app/components/consul/nspace/form/index.hbs
```diff
@@ -0,0 +1,169 @@
...
+
+as |readOnly item Name Description|}}
+<form
+  {{on 'submit' (fn writer.persist item)}}
+  {{disabled readOnly}}
+>
+
+<StateChart 
+  @src={{state-chart 'validate'}}
+as |State Guard ChartAction dispatch state|>
```
johncowen, line 58: I mentioned this in another PR, theres an upcoming refactor of StateChart exports/yields to use a single object rather than multiple exports. I want to do that in a separate PR on its own though instead of mixed up in feature work.

ui/packages/consul-nspaces/app/components/consul/nspace/form/index.hbs
```diff
@@ -0,0 +1,169 @@
...
+      @item={{item}}
+      @validations={{Description}}
+      @chart={{hash
+        state=state
+        dispatch=dispatch
+      }}
+    />
+  </fieldset>
+{{#if (can 'use acls')}}
+  <fieldset id="roles">
```
johncowen, line 86: We should be able to remove these `id` hangovers pretty soon. Just under-the0hood improvements tho.

ui/packages/consul-nspaces/app/components/consul/nspace/form/index.hbs
```diff
@@ -0,0 +1,169 @@
...
+    <h2>Roles</h2>
```
johncowen, line 87: Also would like to revisit `<legend />`s. I'm sure I've re-looked at legend multiple times during working on this, but I must have figured 'nope, still not usable' - would be good to revisit again and document why not (if its still a not) as I can't remember 😬

ui/packages/consul-nspaces/app/components/consul/nspace/form/index.hbs
```diff
@@ -0,0 +1,169 @@
...
+      @allowIdentity={{false}}
+      @items={{item.ACLs.PolicyDefaults}}
+    />
+  </fieldset>
+{{/if}}
+  <div>
+{{#if (and (is "new nspace" item=item) (can "create nspaces"))}}
+      <Action
+        @type="submit"
+        {{disabled (or (is "pristine nspace" item=item) (state-matches state "error"))}}
```
johncowen, line 126: I'm thinking we could change this to use `(not (is 'dirty nspace'))` just as the word `dirty` is more obvious what we are referring to rather than 'pristine', there might be a better word for this, not sure how common/widely `pristine` is used to describe this. Not important, but just thinking out loud in case there are suggestions.

ui/packages/consul-ui/app/services/repository/nspace.js
```diff
@@ -4,6 +4,8 @@ import RepositoryService, { softDelete } from 'consul-ui/services/repository';
 import { PRIMARY_KEY, SLUG_KEY } from 'consul-ui/models/nspace';
 import dataSource from 'consul-ui/decorators/data-source';
 
+import { defaultChangeset as changeset } from 'consul-ui/utils/form/builder';
+
```
johncowen, line 8: This is imported from our old 'nested form builder' which we are moving away from, this will eventually be imported from elsewhere as we don't need the majority of this file anymore.

ui/packages/consul-ui/app/templates/debug.hbs
```diff
@@ -1,6 +1,16 @@
...
+{{! Tell CSS what we have enabled }}
+{{#if (can "use acls")}}
+  {{document-attrs class="has-acls"}}
+{{/if}}
+{{#if (can "use nspaces")}}
+  {{document-attrs class="has-nspaces"}}
+{{/if}}
+{{#if (can "use partitions")}}
+  {{document-attrs class="has-partitions"}}
+{{/if}}
```
johncowen, line 13: These just add some root/html classes for our docs when certain things are enabled. It's just copied over from our prod `application.hbs` file where we've always done this, noticed we didn't do it on our docs pages but we should.

ui/packages/consul-ui/lib/startup/templates/body.html.js
```diff
@@ -57,13 +59,18 @@ ${
...
 `
     : `
 <script>
 (
   function(get, obj) {
     Object.entries(obj).forEach(([key, value]) => {
-      if(get(key)) {
+      if(get(key) || (key === 'CONSUL_NSPACES_ENABLE' && ${
+        env('CONSUL_NSPACES_ENABLED') === '1' ? `true` : `false`
+      })) {
```
johncowen, line 73: Ok so this is the bit I'm not happy with but ok to punt for a bit. We've always had this awkward `ENABLE/ENABLED` thing happening. It started out as just `ENABLE`, but "reasons" happened. So then I made both work inside the codebase instead of trying to keep things consistent. Thing is there are a couple of sharp edges, i.e. in CI to enabled nspace testing the CI script uses "ENABLED", which is the name of a shell env var. Ideally the quick `env` function I wrote here (see above) would smooth that out - but...

Additionally this is going to be needed for testing partitions separately (and potentially ACLs separately, and potentially other things). So we need to add something here that is more generic (not just hardcode in the `NSPACES`). This is enough for the moment so I'll come back and revisit in a week or 2. All this is to do with testing/making it possible to test external feature flags, so it's not affecting anything that we ship.

When I come to do that revisit, I might stop smoothing over the `ENABLE/ENABLED` thing and just move everything to require `ENABLE` which I prefer (tell don't ask). It's getting to the point that every time I see this I wince.

ui/packages/consul-ui/tests/acceptance/dc/nspaces/delete.feature
```diff
@@ -45,7 +46,7 @@ Feature: dc / nspaces / delete: Deleting items with confirmations, success and e
     Given the url "[URL]" responds with a 500 status
     And I click delete
     And I click confirmDelete
-    And "[data-notification]" has the "notification-delete" class
+    And "[data-notification]" has the "notification-update" class
     And "[data-notification]" has the "error" class
```
johncowen, line 50: New approach doesn't have a separate class for deletions. Having these separate classes are a bit of a hangover from previous things. The other changes here are just hardcoding in the nspace instead of using a table with a single row to do the test (if you look at the entire file you'll see what I mean)

johncowen (commented):
Self-review notes:

ui/packages/consul-nspaces/app/components/consul/nspace/notifications/README.mdx
```diff
@@ -0,0 +1,19 @@
+# Consul::Nspace::Notifications
+
+A Notification component specifically for namespaces. This is only a component as we currently use this in two places and if we need to add more types we can do so in one place.
+
+We currently one have one 'remove' type due to the fact that namespaces can't use the default 'delete' notification as they get 'marked for deletion' instead.
```
evrowe, line 5:
```suggestion
We currently only have one 'remove' type due to the fact that namespaces can't use the default 'delete' notification as they get 'marked for deletion' instead.
```

ui/packages/consul-nspaces/app/components/consul/nspace/selector/README.mdx
```diff
@@ -0,0 +1,40 @@
+# Consul::Nspace::Selector
+
+A conditional, autoloading, menu component specifically for making it easy to select namespaces.
+
+Please note: 
+
+- Currently at least, you must add this inside of a `<ul>` element.
```
evrowe, line 7:
```suggestion
- Currently at least, you must add this inside of a `<ul>` element, as the `<li>` elements output by this component are intended to be mixed with other sibling `<li>`s from other components or template code.
```

ui/packages/consul-ui/app/routing/route.js
```diff
@@ -22,11 +22,14 @@ export default class BaseRoute extends Route {
...
     const template = get(routes, `${routeName}._options.template`);
-    if(template) {
+    if (template) {
       this.templateName = template;
     }
     const queryParams = get(routes, `${routeName}._options.queryParams`);
-    if(queryParams && (this.routeName === 'dc.partitions.index' || this.routeName === 'oauth-provider-debug')) {
+    if (
+      queryParams &&
+      ['dc.partitions.index', 'dc.nspaces.index', 'oauth-provider-debug'].includes(this.routeName)
```
evrowe, line 31: Small QOL changes to the code like this really just warm my heart 😂

ui/packages/consul-ui/lib/startup/templates/body.html.js
evrowe: Yeah that's a really tough one to work with as the scope of the problem grows. We should definitely figure out how to push things toward consistency as this problem will only get worse as we go on.

Does this require coordination from the back-end to address properly/completely? I wonder if we could start on our end by using a function that can automatically `s/ENABLED/ENABLE` in env var names when they are first read into the startup script.

evrowe (approved):
Looks ready to ship, I left a few open, non-blocking questions and suggested two small copy changes

ui/packages/consul-ui/lib/startup/templates/body.html.js
johncowen: > Does this require coordination from the back-end to address properly/completely?

🤔 Not entirely sure, but maybe it's not worth changing things up there. We have a place where we can draw the line between backend/frontend and do what works best for us (and backend can do whats best for them). I think the line is more blurred around CI things which is where the problem arose here, and a change there is super straightforwards that we can probably do ourselves.

Related: We are slowly but surely moving everything to use `(can "use nspaces")` etc instead of the env var directly. Thinking more into it I think there's something here around "setting means `ENABLE` (i.e. telling) and getting `ENABLED` (i.e. asking), our `can` helper is all about asking" - I'm kind talking out loud again here but it will maybe/probably sort itself out as we move everything over to `can` and become: `Please use `*_ENABLE` to configure environment variables, in the codebase please use `can` to inspect feature flags etc.` or something like that.

When we come to do this we can come back here and figure out what we want to do 👍

.circleci/config.yml
```diff
@@ -22,7 +22,7 @@ references:
     test-results: &TEST_RESULTS_DIR /tmp/test-results
 
   cache:
-    yarn: &YARN_CACHE_KEY consul-ui-v5-{{ checksum "ui/yarn.lock" }}
+    yarn: &YARN_CACHE_KEY consul-ui-v6-{{ checksum "ui/yarn.lock" }}
```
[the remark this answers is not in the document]
johncowen: This is just a cache buster, really doesn't matter what it is as long as its different from the last one. I only ever do this if we get some sort of npm dependency caching problem in CI, and the quickest thing to try is to bump which will sometimes be the solution - other products have this problem sometimes also and its generally what we all do in this case. I half remember this time it was more to do with our workspaces in the end, so theoretically this could have been reverted - but its just a random hash more than anything so no biggie.

ui/packages/consul-nspaces/app/components/consul/nspace/list/README.mdx
```diff
@@ -1,29 +1,30 @@
----
-class: ember
----
-## Consul::Nspace::List
+# Consul::Nspace::List
+
+A presentational component for rendering Consul Partitions
```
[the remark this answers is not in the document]
johncowen:
```suggestion
A presentational component for rendering Consul Namespaces
```

😆 Ah! Good spot its for namespaces not Partitions. Damn you copypasta! 😆

ui/packages/consul-nspaces/app/components/consul/nspace/selector/README.mdx
```diff
@@ -0,0 +1,40 @@
...
+  <Consul::Nspace::Selector
+    @dc={{hash
+      Name='dc-1'
+    }}
+    @nspace='default'
+    @partition='default'
+    @nspaces={{or this.nspaces (array)}}
+    @onchange={{action (mut this.nspaces) value="data"}}
+  />
+</ul>
```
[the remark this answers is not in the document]
johncowen: Yeah I think you got it there from what you explained, ideally it would know whether it was in a ul or not and autoadd a li if so. Another approach would be to `display: none` empty `<li>`s - but thats a bit yuk - but we may or may not change to do that at somepoint, plus you could get caught up in whitespace issues where the `li` is empty but full of handlebars related whitespace (where `~` sometimes works and sometimes doesn't)

I think the suggestion you added below captures it perfectly so I'll add that in here 👍

ui/packages/consul-ui/app/routing/route.js
johncowen: 😂 

Yeah agree, as soon as things get past like 1 or 2 of these it's time to make it more readable.

FYI: This entire piece of the conditional shouldn't be here really, but is due to "reasons" - I'm hoping it will be gone by the time we hit 1.11 for real, if not it can disappear post 1.11.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/511907.

PR #11645, opened 2021-11-24
types: add types/tls.go for strongly-typed TLS versions and cipher suites

mikemorris:
Introduces types `TLSVersion` and `TLSCipherSuite` which will be used by both the agent TLS implemention (in #11647) as well as Envoy configuration (initially for ingress gateways).

The biggest outstanding question I have on this is whether it would be appropriate to include the lookup maps from these constant values to Go `crypto/tls` values in `types/tls.go`, or if it would be preferable to avoid adding that package import and keep those lookups over in `tlsutil/config.go` as they're currently implemented.

**TODO:**
- [x] initialize `TLSVerison` constants starting at zero
  - `iota - 1` was causing some undesirable behavior with structs when lacking an explicit initialization for a key taking a `TLSVersion` value would default to the zero-value of `TLSVersionAuto` instead of `TLSVersionInvalid`
- [x] ~~Determine whether this file should import `crypto/tls` and/or `github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3` or if those lookup maps should remain in implementation code in `goTLSVersions` and `goTLSCipherSuites` in https://github.com/hashicorp/consul/pull/11647 and `EnvoyTLSVersions` in https://github.com/hashicorp/consul/pull/11576 respectively~~
  - Not doing this initially, could happen as a followup if it's desirable.


types/tls.go
```diff
@@ -0,0 +1,173 @@
...
+	// deployment pattern of upgrading servers first. These should be updated to
+	// the newer config strings in a future release.
+	ConsulAutoConfigTLSVersionStrings = map[TLSVersion]string{
+		TLSVersionAuto: "",
+		TLSv1_0:        "tls10",
+		TLSv1_1:        "tls11",
+		TLSv1_2:        "tls12",
+		TLSv1_3:        "tls13",
+	}
+	EnvoyTLSVersionStrings = map[TLSVersion]string{
```
mikemorris, line 58: This map might be able to be removed, as the actual lookup needed is for `envoy_tls_v3.TlsParameters_TlsProtocol` typed constants.

types/tls_test.go
```diff
@@ -0,0 +1,24 @@
...
+	require.Greater(t, TLSv1_2, TLSv1_1)
+	require.Greater(t, TLSv1_1, TLSv1_0)
+
+	require.Less(t, TLSv1_2, TLSv1_3)
+	require.Less(t, TLSv1_1, TLSv1_2)
+	require.Less(t, TLSv1_0, TLSv1_1)
+}
+
+func TestTLSVersion_Invalid(t *testing.T) {
+	require.NotEqual(t, TLSVersionAuto, TLSVersionInvalid)
```
boxofrad, line 20: Is `TLSVersionAuto` the desired zero-value? If so, it might be worth making that explicit in a test:

```Go
var zeroValue TLSVersion
require.Equal(t, TLSVersionAuto, zeroValue)
```

types/tls.go
```diff
@@ -0,0 +1,173 @@
...
+		TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384:         "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384",
+
+		TLS_RSA_WITH_AES_128_GCM_SHA256: "TLS_RSA_WITH_AES_128_GCM_SHA256",
+		TLS_RSA_WITH_AES_128_CBC_SHA:    "TLS_RSA_WITH_AES_128_CBC_SHA",
+		TLS_RSA_WITH_AES_256_GCM_SHA384: "TLS_RSA_WITH_AES_256_GCM_SHA384",
+		TLS_RSA_WITH_AES_256_CBC_SHA:    "TLS_RSA_WITH_AES_256_CBC_SHA",
+
+		TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256: "TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256",
+		TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256:   "TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256",
+	}
```
boxofrad, line 156: Totally optional, but as this has all the same elements of `TLSCipherSuites`, what do you think of inverting it in code to avoid potential accidental typos, missed elements etc?

```suggestion
	HumanTLSCipherSuiteStrings = func() map[TLSCipherSuite]string {
		var inverted map[TLSCipherSuite]string
		for k, v := range TLSCipherSuites {
			inverted[v] = k
		}
		return inverted
	}()
```

boxofrad (approved):
Neat! LGTM 🚢 

Left a couple of small comments, please feel free to take or leave them.

types/tls_test.go
mikemorris: Initially I had thought it would be a more appropriate zero-value, but after some experimentation over in https://github.com/hashicorp/consul/pull/11576#discussion_r761496909 discovered `TLSVersionInvalid` would be the more appropriate zero-value, will add a test to cover that.

types/tls.go
```diff
@@ -0,0 +1,173 @@
...
+	HumanTLSVersionStrings = map[TLSVersion]string{
+		TLSVersionAuto: "Allow implementation to select TLS version",
+		TLSv1_0:        "TLS 1.0",
+		TLSv1_1:        "TLS 1.1",
+		TLSv1_2:        "TLS 1.2",
+		TLSv1_3:        "TLS 1.3",
+	}
+	// NOTE: these currently map to the deprecated config strings to support the
+	// deployment pattern of upgrading servers first. These should be updated to
+	// the newer config strings in a future release.
```
mikemorris, line 50:
```suggestion
	// be the inverse of the TLSVersions map with newer config strings in a future release.
```

types/tls.go
mikemorris: Ooh, I think I like that and does seem to make sense for this specific lookup map, thanks for the suggestion!

types/tls.go
kisunji: I'd like to agree with 
> those lookup maps should remain in implementation code
to keep this package clean from non stdlib imports

kisunji:
> I noticed in #11576 that these are members in a struct that is JSON-encoded — is the integer value okay here, or should this type implement a custom `MarshalJSON`? (maybe to output the string value instead).

Following up on the conversation on slack, if these do get serialized to JSON for some kind of config, I think there should be some custom (TLSVersion)MarshalJSON/UnmarshalJSON behavior so a user doesn't have to deal with figuring out what the ints mean

types/tls.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+)
+
+func (v TLSVersion) String() string {
+	return HumanTLSVersionStrings[v]
+}
+
+func (v TLSVersion) ConsulConfigString() string {
+	return ConsulConfigTLSVersionStrings[v]
+}
```
kisunji, line 81: I think `String()` should be the string representation of the enums, not the human-readable string (which I'm not sure we need since it's understandable as-is).

Instead of using `ConsulConfigTLSVersionStrings` I think it's simpler to add a switch-case for all the enum possibilities like:
```go
func (v TLSVersion) String() string {
	switch v {
	case TLSVersionAuto:
		return "TLS_AUTO"
	case TLSv1_0:
		return "TLSv1_0"      
	case TLSv1_1:
		return "TLSv1_1" 
        // ...so on
	default:
		return "INVALID"
	}
}
```

That way the type is actually immutable and we're not exporting unnecessary variables from `types`

types/tls.go
```diff
@@ -0,0 +1,188 @@
...
+
+func (v TLSVersion) MarshalJSON() ([]byte, error) {
+	return []byte(`"` + v.ConsulConfigString() + `"`), nil
```
kisunji, line 84: A common pattern for simple custom marshal/unmarshaling is to get the primitive you want then delegate the work to `json.Marshal` 
```go
func (v TLSVersion) MarshalJSON() ([]byte, error) {
	return json.Marshal(v.String())
}
```

kisunji: So you don't have to fiddle with parsing `"` and edge-case JSON stuff

kisunji (approved):
I'm fine approving and iterating moving forwards! It's hard to nail this down without having used it much.

mikemorris:
As a followup to this PR, we should consider adding public `ParseTLSVersion` and `ParseTLSCipherSuite` functions which can return errors and switch the lookup maps to be private.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/517777.

Issue #9983, opened 2021-04-07
Remove unnecessary `agent/config` struct copies


dnephin:
[agent/config.config.go](https://github.com/hashicorp/consul/blob/master/agent/config/config.go) defines many struct types. These structs generally have fields with pointer types. The structs are used for two purposes:
1. unmarshalling config from config files (using hcl or json to `map[string]interface{}`, and then mapstructure to decode into the types)
2. merging of multiple `Config` structs into a single one

These structs, in many cases, are mirrored in 3 other places (`api`, `agent/structs`, and `proto/*`). This means that adding a field to these structs requires adding it in 4 places!

I discovered recently that in some cases, the copy in `agent/config` is not necessary. If the field on `Config` is a slice or a pointer to a struct, then the merge logic always appends the values, it doesn't merge the fields. This makes the pointer unnecessary! In these cases we can use the types defined in the `api` package in the `Config` struct. This change would have a secondary benefit as well. In the past we've encountered a few bugs where the `api` and `struct` types decoded differently, resulting in some values being dropped. By using the same type we reduce the likelihood of those bugs. The config is effectively part of the public API of Consul, so using the API types for the config seems reasonable. The `agent/config` package already has a direct dependency on `api`, so this change doesn't impact our dependency tree.

The structs that are used in slices, that we should be able to remove, are:

* `CheckDefinition`
* `ServiceDefinition`
* `EnterpriseMeta` 
* `ServiceAddress`
* `ServiceWeights`
* `ServiceProxy`
* `Upstream`
* `MeshGatewayConfig`
* `ExposeConfig`
* `ExposePath`
* `ServiceConnect`

The first two `Definition` structs do technically exists as non-slice fields, however [as it says in this comment](https://github.com/hashicorp/consul/blob/v1.10.0-alpha/agent/config/config.go#L151) these definitions can not be spread across multiple sources. The same merging rules apply to pointers and slices, so we don't need pointer fields within the struct.

PR #11647, opened 2021-11-24
agent: use TLS types for config

mikemorris:
Although this deprecates the current config values for agent config `tls_min_version`, it _should_ be backwards compatible because those values will still be supported (with a warning) and the auto-config implementation will continue to use the _old, deprecated_ values for at least a full major release cycle, to support the pattern of upgrading servers before client agents. All internal implementation has been moved to use typed constants instead of passing the raw strings around.

The config values for `tls_cipher_suites` are unchanged, but internal implementation now uses typed constants instead of `uint16`.

_Aside_: changelog checker seemed to not enforce an entry on PR, I think because it touched `website`? That should probably be modified to check if _all_ files in the diff are in that directory rather than _any_.

TODO:
- [ ] Check if any duplication of structs can be avoided following the pattern in https://github.com/hashicorp/consul/issues/9983

### DIFF
--- a/tlsutil/config.go
+++ b/tlsutil/config.go
@@ -18,6 +18,7 @@ import (
 	"github.com/hashicorp/go-multierror"
 
 	"github.com/hashicorp/consul/logging"
+	"github.com/hashicorp/consul/types"
 )
 
 // ALPNWrapper is a function that is used to wrap a non-TLS connection and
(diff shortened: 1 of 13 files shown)


tlsutil/config.go
```diff
@@ -944,32 +974,55 @@ func (c *Configurator) AuthorizeServerConn(dc string, conn TLSConn) error {
...
 
-// ParseCiphers parse ciphersuites from the comma-separated string into
-// recognized slice
-func ParseCiphers(cipherStr string) ([]uint16, error) {
+// NOTE: any new cipher suites will also need to be added in types/tls.go
+// TODO: should this be moved into types/tls.go? Would importing Go's tls
+// package in there be acceptable?
+var goTLSCipherSuites = map[types.TLSCipherSuite]uint16{
+	// TODO: CHACHA20_POLY1305 cipher suites are not currently implemented for Consul agent TLS
+	// but are available in Go, add them?
```
mikemorris, line 982: Adding the CHACHA20_POLY1305 suites seems like a thing we should do.

website/content/docs/agent/options.mdx
```diff
@@ -2261,9 +2261,11 @@ signed by the CA can be used to gain full access to Consul.
...
-  the minimum supported version of TLS. Accepted values are "tls10", "tls11", "tls12",
-  or "tls13". This defaults to "tls12". WARNING: TLS 1.1 and lower are generally
-  considered less secure; avoid using these if possible.
+  the minimum supported version of TLS. Accepted values as of Consul 1.11.0 are "TLSv1_0",
+  "TLSv1_1", "TLSv1_2", or "TLSv1_3". This defaults to "TLSv1_2". WARNING: TLS 1.1 and
+  lower are generally considered less secure; avoid using these if possible.
+  Deprecated values of "tls10", "tls11", "tls12" and "tls13" are currently still accepted
+  but will emit a warning during configuration and will be removed in a future release.
 
 - `tls_cipher_suites` Added in Consul 0.8.2, this specifies the list of
```
mikemorris, line 2270: TODO: Update search link below.

website/content/docs/agent/options.mdx
```diff
@@ -2261,9 +2261,11 @@ signed by the CA can be used to gain full access to Consul.
...
 
 - `tls_min_version` Added in Consul 0.7.4, this specifies
-  the minimum supported version of TLS. Accepted values are "tls10", "tls11", "tls12",
-  or "tls13". This defaults to "tls12". WARNING: TLS 1.1 and lower are generally
-  considered less secure; avoid using these if possible.
+  the minimum supported version of TLS. Accepted values as of Consul 1.11.0 are "TLSv1_0",
+  "TLSv1_1", "TLSv1_2", or "TLSv1_3". This defaults to "TLSv1_2". WARNING: TLS 1.1 and
+  lower are generally considered less secure; avoid using these if possible.
+  Deprecated values of "tls10", "tls11", "tls12" and "tls13" are currently still accepted
+  but will emit a warning during configuration and will be removed in a future release.
```
trujillo-adam, line 2268:
```suggestion
  the minimum supported version of TLS. The following values are accepted:  
  * `TLSv1_0`
  * `TLSv1_1` 
  * `TLSv1_2` (default)
  * `TLSv1_3` 
WARNING: TLS 1.1 and older are generally considered less secure and should not be used if possible.
The following values are also valid, but they are deprecated and will be removed in a future release: 
* `tls10`
* `tls11` 
* `tls12` 
* `tls13`
A warning message will appear if a deprecated value is specified.
```
I'm not sure if we are able to use bullets here because this is already a list, but we should use code font for the possible values.

trujillo-adam (commented):
I don't know about the code, but I left a suggestion for the MDX content.

PR #11640, opened 2021-11-24
ui: Add `Service.Namespace` variable to dashboard URL templates

johncowen:
See https://www.consul.io/docs/connect/observability/ui-visualization#configuring-dashboard-urls

Allows only `Datacenter`, `Service.Name` (as before) this PR adds `Service.Namespace`.

I've added backport labels here but pretty sure 1.9 and 1.8 will have to be manual.

P.S. There needs to be a follow up after the backports are done to add `Partition` here for 1.11 also.

PR #11653, opened 2021-11-25
ui: [Backport/1.8.x] Add `Service.Namespace` variable to dashboard URL templates (#11640)

johncowen:
See #11640 Note: 1.8.x doesn't have a topology view, therefore we only need to do this in once place.

We currently allow only Datacenter, Service.Name, this PR adds Service.Namespace.


kaxcode (approved):
:shipit:

PR #11655, opened 2021-11-25
api: add keys watch

slql0011:


eculver:
This looks good. Thanks for contributing! Would you mind adding a description for your use-case? Also, it looks like there is some prior art in our tests to add some coverage for these watches. Would you mind taking a crack at adding coverage for this as well?

markan:
@slql0011 any updates on this?

Amier3:
Hey @slql0011 

 I saw you committed again recently, Would you be able to write tests for this change & let us know a little about the use case for this change?

PR #11660, opened 2021-11-25
ca: reduce consul provider backend interface a bit

dnephin:
This is a small change I made while working on other bug fixes. I did this as a way of using the ConsulProvider as an "external PKI", but I decided to take a different approach. While the original reason for this change no longer exists, I thought this was still a nice cleanup because it limits the access the Consul Provider has to the state store. This should prevent accidentally using something from the state store that other providers would not have access to.


boxofrad (approved):
Neat! LGTM 🚀

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/509786.

Issue #11598, opened 2021-11-17
ca: Support using an external CA as the Trusted CA


dnephin:
Related to #11448

Currently the [Connect CA](https://www.consul.io/docs/connect/ca/consul) uses the following certificates with the [Vault provider](https://www.consul.io/docs/connect/ca/vault):
* a Root CA, stored in Vault using the `root_pki_path`, used to sign intermediate CA certificates
* an intermediate CA signed by that root, stored in Vault using the `intermediate_pki_path`, used to sign leaf certificates

Using a non-root (not self signed) CA certificate as the `root_pki_path` almost works. Consul CA can sign intermediate and leaf certificates, but when an application attempts to verify one of the leaf certs that verification may fail with "unknown authority". The problem is that with this setup the [connect/ca API](https://www.consul.io/api-docs/connect/ca) and xDS API have no way of returning the real external root, because it is never stored in Consul.

Some applications may be able to distribute that external root CA and add it as a trusted CA, but that may not always be possible. It does not appear to be possible with Envoy which Consul uses as the connect proxy.

### Proposal

To support external root CAs as the trusted CA we need to store the public certificate of the real root CA in Consul, and return it as the `RootCert` in both the HTTP API and xDS API. The primary Connect CA will need to be added to the list of `IntermediateCerts` in that API response.

To make that happen, I believe we'll need to change the  [ca.Provider](https://pkg.go.dev/github.com/hashicorp/consul/agent/connect/ca#Provider) interface. The `GenerateRoot`, and `ActiveRoot` methods need to be changed to allow them to return the full chain (both the real root CA, and the intermediate used by Consul CA). That could be either as a single PEM encoded bundle, or it could be a slice of PEM encoded certificates.

To change that interface we need to audit all calls to those two methods. I believe that doing at least some of the work described in #11159 would benefit us here, as that would remove two of the calls to `ActiveRoot`, which removes the need to adjust them to this new interface. Any additional work required for #11159 can be left for a later time.

### Testing

Since this setup does not fail until the leaf certificates are verified, the testing setup to verify this change is pretty involved. Most of the setup is already documented in #11448, but I'll repeat it here.

**Environment:**
* create a root CA in Vault
* create an intermediate A, signed by the root CA, and use that path as `root_pki_path`
* setup a third pki path as `intermediate_pki_path`

We need to test the following three flows in both a primary and a secondary DC (because the secondary DC has some different code paths from the primary):

[...]

Issue #10532, opened 2021-06-30
connect/ca/roots endpoint do not return the intermediate cert when using vault provider


dhiaayachi:
When using a vault provider with consul and starting a new server. the call to `v1/connect/ca/roots` will not return an intermediate cert until half the root cert life time passed.

This happen because in `getIntermediateCAPrimary` we don't store the generated intermediate in raft

dnephin:
Thinking about this more, the issue might not simply be the API missing data. Envoy also won't receive the intermediate cert, which would mean the TLS chain can not be verified and it could lead to services not being able to connect to each other.

dnephin:
#10826 sounds like it might be the same issue

dnephin:
After looking at this more, I think Envoy is fine because it gets the intermediates from the `Sign` leaf API call, not from the response of `/connect/ca/roots`, but this is still a problem for anyone using the API.

PR #11661, opened 2021-11-25
ca: remove one call to Provider.ActiveRoot

dnephin:
This change is related to #11598, and #11159.  More detail in the commit messages, and I'll call out a few things below with inline comments.

The PR is not large, but it is still best viewed by individual commit, as there are a few independent changes, and the commit message has more context.

ActiveRoot should not be called from the secondary DC, because there should not be a requirement to run the same Vault instance in a secondary DC (#11159). `SignCertificate` is called in a secondary DC, so it should not call `Provider.ActiveRoot`. This PR removes that call.

We would also like to change the interface of `Provider.ActiveRoot` so that we can support using an intermediate cert as the primary CA in Consul (#11598). In preparation for making that change I am reducing the number of calls to `Provider.ActiveRoot`, so that there are fewer code paths to modify when the interface changes.

In order to make this change I had to modify the existing test a bit, and fix a couple problems with the mock. I also reduced the runtime of the test so that I wasted less time waiting for it to run.

### DIFF
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -1463,28 +1463,22 @@ func (c *CAManager) SignCertificate(csr *x509.CertificateRequest, spiffeID conne
 
 	connect.HackSANExtensionForCSR(csr)
 
-	root, err := provider.ActiveRoot()
-	if err != nil {
-		return nil, err
-	}
 	// Check if the root expired before using it to sign.
-	err = c.checkExpired(root)
+	// TODO: we store NotBefore and NotAfter on this struct, so we could avoid
+	// parsing the cert here.
+	err = c.checkExpired(caRoot.RootCert)
 	if err != nil {
 		return nil, fmt.Errorf("root expired: %w", err)
 	}
 
-	inter, err := provider.ActiveIntermediate()
-	if err != nil {
-		return nil, err
-	}
-	// Check if the intermediate expired before using it to sign.
-	err = c.checkExpired(inter)
-	if err != nil {
-		return nil, fmt.Errorf("intermediate expired: %w", err)
+	if c.isIntermediateUsedToSignLeaf() && len(caRoot.IntermediateCerts) > 0 {
+		inter := caRoot.IntermediateCerts[len(caRoot.IntermediateCerts)-1]
+		if err := c.checkExpired(inter); err != nil {
+			return nil, fmt.Errorf("intermediate expired: %w", err)
+		}
 	}
 
 	// All seems to be in order, actually sign it.
-
 	pem, err := provider.Sign(csr)
 	if err == ca.ErrRateLimited {
 		return nil, ErrRateLimited
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -1498,11 +1498,6 @@ func (c *CAManager) SignCertificate(csr *x509.CertificateRequest, spiffeID conne
 		pem = pem + ca.EnsureTrailingNewline(p)
 	}
 
-	// Append our local CA's intermediate if there is one.
-	if inter != root {
-		pem = pem + ca.EnsureTrailingNewline(inter)
-	}
-
 	modIdx, err := c.delegate.ApplyCALeafRequest()
 	if err != nil {
 		return nil, err
--- a/agent/consul/leader_connect_ca_test.go
+++ b/agent/consul/leader_connect_ca_test.go
@@ -422,6 +422,9 @@ func TestCAManager_SignCertificate_WithExpiredCert(t *testing.T) {
 		{"root in the future", time.Now().AddDate(0, 0, 1), time.Now().AddDate(0, 0, 2), time.Now().AddDate(0, 0, -1), time.Now().AddDate(0, 0, 2), false, ""},
 	}
 
+	caPrivKey, err := rsa.GenerateKey(rand.Reader, 4096)
+	require.NoError(t, err, "failed to generate key")
+
 	for _, arg := range args {
 		t.Run(arg.testName, func(t *testing.T) {
 			// No parallel execution because we change globals
@@ -439,11 +442,13 @@ func TestCAManager_SignCertificate_WithExpiredCert(t *testing.T) {
 			conf.ConnectEnabled = true
 			conf.PrimaryDatacenter = "dc1"
 			conf.Datacenter = "dc2"
+
+			rootPEM := generateCertPEM(t, caPrivKey, arg.notBeforeRoot, arg.notAfterRoot)
+			intermediatePEM := generateCertPEM(t, caPrivKey, arg.notBeforeIntermediate, arg.notAfterIntermediate)
+
 			delegate := NewMockCAServerDelegate(t, conf)
 			manager := NewCAManager(delegate, nil, testutil.Logger(t), conf)
 
-			rootPEM := generateCertPEM(t, arg.notBeforeRoot, arg.notAfterRoot)
-			intermediatePEM := generateCertPEM(t, arg.notBeforeIntermediate, arg.notAfterIntermediate)
 			manager.providerShim = &mockCAProvider{
 				callbackCh:      delegate.callbackCh,
 				rootPEM:         rootPEM,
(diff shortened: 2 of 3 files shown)


agent/structs/connect_ca.go
```diff
@@ -86,11 +86,20 @@ type CARoot struct {
...
-	// attach to any leaf certs signed by this CA.
+	// attach to any leaf certs signed by this CA. The list may include a
+	// certificate cross-signed by an old root CA, any subordinate CAs below the
+	// root CA, and the intermediate CA used to sign leaf certificates in the
+	// local Datacenter.
+	//
+	// If the provider which created this root uses an intermediate to sign
+	// leaf certificates (Vault provider), or this is a secondary Datacenter then
+	// the intermediate used to sign leaf certificates will be the last in the
+	// list.
```
dnephin, line 102: This convention of "leaf signing cert last" already existed for the new cert expiry metrics we added. I'm calling it out here more explicitly so that it is well understood. We are going to be depending on this convention even more as part of subsequent changes for #11598.

agent/consul/leader_connect_ca.go
```diff
@@ -1498,11 +1498,6 @@ func (c *CAManager) SignCertificate(csr *x509.CertificateRequest, spiffeID conne
 		pem = pem + ca.EnsureTrailingNewline(p)
 	}
 
-	// Append our local CA's intermediate if there is one.
-	if inter != root {
-		pem = pem + ca.EnsureTrailingNewline(inter)
-	}
```
dnephin, line 1506: Calling out this change loudly. I'm pretty sure this is safe, but I would appreciate some close attention to be sure I'm not missing something. Echo'ing the commit message here:

> Immediately above this line we are already appending the full list of
intermediates. The `provider.ActiveIntermediate` MUST be in this list of
intermediates because it must be available to all the other non-leader
Servers.  If it was not in this list of intermediates then any proxy
that received data from a non-leader would have the wrong certs.

> This is being removed now because we are planning on changing the
`Provider.ActiveIntermediate` interface, and removing these extra calls ahead of
time helps make that change easier.

From what I could tell in git history this logic was part of the initial implementation, so likely just an oversight that we haven't noticed yet.

agent/consul/leader_connect_ca_test.go
```diff
@@ -234,13 +236,14 @@ func initTestManager(t *testing.T, manager *CAManager, delegate *mockCAServerDel
 }
 
 func TestCAManager_Initialize(t *testing.T) {
-
 	conf := DefaultConfig()
 	conf.ConnectEnabled = true
 	conf.PrimaryDatacenter = "dc1"
 	conf.Datacenter = "dc2"
 	delegate := NewMockCAServerDelegate(t, conf)
+	delegate.secondaryIntermediate = delegate.primaryRoot.RootCert
```
dnephin, line 244: This is definitely weird, but it preserves the old behaviour of this test. I think we'll need to do a bit more work to make this mock a little more realistic.

agent/consul/leader_connect_ca_test.go
```diff
@@ -395,14 +400,13 @@ func TestCAManager_SignCertificate_WithExpiredCert(t *testing.T) {
 
 			// Simulate Wait half the TTL for the cert to need renewing.
 			manager.timeNow = func() time.Time {
-				return time.Now().Add(500 * time.Millisecond)
+				return time.Now().UTC().Add(500 * time.Millisecond)
```
dnephin, line 403: This was just to make debugging easier. It should not change any functionality.

agent/consul/leader_connect_ca.go
dnephin: Hmm, because of the bug described in #10532, this change is not safe. We'd have to fix that first for this change to be ok.

dnephin: #11671 fixes the problem that this is masking, so rebasing this PR on that change makes this safe once again.

agent/consul/leader_connect_ca_test.go
```diff
@@ -431,6 +436,7 @@ func generateCertPEM(t *testing.T, caPrivKey *rsa.PrivateKey, notBefore time.Tim
 		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
 		KeyUsage:              x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
 		BasicConstraintsValid: true,
+		URIs:                  []*url.URL{connect.SpiffeIDAgent{Host: "foo"}.URI()},
```
kisunji, line 439: Confirming this change is intentional?

kisunji (approved):
one minor question, but LGTM (I assume you can't merge until the linked issue is resolved)

agent/consul/leader_connect_ca.go
```diff
@@ -1548,3 +1546,16 @@ func (c *CAManager) checkExpired(pem string) error {
...
+
+func primaryUsesIntermediate(provider ca.Provider) bool {
+	_, ok := provider.(ca.PrimaryUsesIntermediate)
+	return ok
+}
+
+func (c *CAManager) isIntermediateUsedToSignLeaf() bool {
+	if c.serverConf.Datacenter != c.serverConf.PrimaryDatacenter {
+		return true
+	}
```
FFMMM, line 1547: so if we are not the primary DC we always sign with the intermediate?

agent/consul/leader_connect_ca_test.go
dnephin: Oh ya, great question! I should have given more of an explanation of this part of the change.

I was running into a panic in this test because of this line:
https://github.com/hashicorp/consul/blob/1a7b89c6cbb82ab497963846ce7e8fa39ca3b53a/agent/consul/leader_connect_ca.go#L1532

Basically the PEM encoded string coming out did not have any URIs. I tracked it back to how our fake was using the wrong certs for the local intermediate cert. Previously we would return the root cert as the "intermediate cert", and the root cert happened to have URIs on it, so the test was passing.

So once I fixed the fake to return the correct cert, I had to fix the generation of that cert to include the URIs.

The fake is still a bit strange. Normally the signed leaf cert should be the one with URIs on it, but the fake is returning an empty string for the actual leaf cert, so we have to rely on the URIs in one of the parents.

I think we need to do more to clean this up, but I didn't want to include too much test cleanup in this PR.

agent/consul/leader_connect_ca.go
dnephin: Yup, exactly! Every provider uses an intermediate certificate in a secondary DC to sign leaf certs.

agent/consul/leader_connect_ca.go
```diff
@@ -1465,28 +1463,22 @@ func (c *CAManager) SignCertificate(csr *x509.CertificateRequest, spiffeID conne
...
-	inter, err := provider.ActiveIntermediate()
-	if err != nil {
-		return nil, err
-	}
-	// Check if the intermediate expired before using it to sign.
-	err = c.checkExpired(inter)
-	if err != nil {
-		return nil, fmt.Errorf("intermediate expired: %w", err)
+	if c.isIntermediateUsedToSignLeaf() && len(caRoot.IntermediateCerts) > 0 {
+		inter := caRoot.IntermediateCerts[len(caRoot.IntermediateCerts)-1]
```
dnephin, line 1475: Leaving a TODO for myself. This can use `getLeafSigningCertFromRoot`  once #11671 is merged.

```suggestion
	if c.isIntermediateUsedToSignLeaf() && len(caRoot.IntermediateCerts) > 0 {
		inter := c.getLeafSigningCertFromRoot(caRoot)
```

kyhavlov (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/553249.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit f31e0b8b1a6bd5554c684a5781324202e9143d86 onto `release/1.11.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/568392)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit f31e0b8b1a6bd5554c684a5781324202e9143d86 onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/568392)

Issue #11662, opened 2021-11-25
ca: CARoot.SigningKeyID is wrong after config update when using the Vault CA provider


dnephin:
It looks like we mostly fixed this issue a couple years ago in #6513 (#6507). That PR fixed it by updating the signing key to the correct version during initialize, but I believe the same bug still exists in the `UpdateConfiguration` flow.

In `parseCARoot` we set the `SigningKeyID` to the encoded `rootCert.SubjectKeyId`, which is correct for the Consul and AWS providers in the Primary DC, but not for the Vault provider, because the Vault provider uses an intermediate (not the root) to sign leafs:
https://github.com/hashicorp/consul/blob/7bbad6fe560ab27f51dd6c5cd16f28418082424c/agent/consul/leader_connect_ca.go#L224

`parseCARoot` is called from two places: `primaryInitialize` (when a server is elected leader and initializes the provider) and `primaryUpdateRootCA` when a CA config change is made by a user.

In `primaryInitialize` this bug is hidden by the migration code that was added in #6513:
https://github.com/hashicorp/consul/blob/7bbad6fe560ab27f51dd6c5cd16f28418082424c/agent/consul/leader_connect_ca.go#L456-L461

Despite the comment, I believe this condition continues to be true today, because `parseCARoot` above still sets it to the wrong value.

In `primaryUpdateRootCA` there is no corresponding fix-up code, so we are left with the wrong signing key:
https://github.com/hashicorp/consul/blob/7bbad6fe560ab27f51dd6c5cd16f28418082424c/agent/consul/leader_connect_ca.go#L826

It is possible #7012 (#6886) mitigates this in the secondary DC.

I haven't yet written a test to reproduce the problem, but it seems like it should be possible to.

Issue #10634, opened 2021-07-16
ca: general improvements to the CAMananger state machine


dnephin:
The diagram below is a model of the [state machine](https://github.com/hashicorp/consul/blob/v1.10.1/agent/consul/leader_connect_ca.go#L23-L29) used by the `CAManager`.

![mermaid-diagram-20210716154911](https://user-images.githubusercontent.com/442180/126001289-3c9ac4f9-b9fe-4e60-99e3-6b2d2ea6ab0b.png)

While working on this diagram, I believe I've identified the following problems and areas for improvement:

**Bugs**
1. [setState](https://github.com/hashicorp/consul/blob/v1.10.1/agent/consul/leader_connect_ca.go#L107-L117) allows any caller to skip state transition validation, and if the current state is `caStateInitialized` any transition is allowed. Because of this there are two `INVALID` states that shouldn't be allowed, but are still supported by the state machine
2. There is a race condition that is not handled. If leadership is lost while `UpdateRoots`, `RenewIntermediate`,  `UpdateConfiguration`, or `InitializeCA` are running then the state will be changed to `Uninitialized` by `Stop`. But once the running operation ends, the `defer` will force the state back to `Initialized` (by calling `setState` with `validateState=false`). This could result in the server thinking it has an initializedCA. If that server were to regain leadership after some time, it would skip its `InitializeCA` thinking it was already initialized. I think this would leave the `CAManager` in a bad state with a `nil` `Provider`.

**Improvements**

The purpose of the state machine is to only allow valid state transitions, but because of `valdiateState=false` we ignore that validation in many cases. 

I also notice that for the most part, the states that seem to be relevant are: Uninitialized, Initiailizing, and Initialized. The other 2 states are actually more like a "lease", where we want to only allow a single reconfig or renew operation at a given time.

I think we could address these problems by moving most of this logic into its own type, and changing the interface to better model the code. Possibly something like this:

```go
type State interface {
    InitializationLease() (Complete, error)
    UpdateConfigLease() (Complete, error)
    RenewIntermediateLease() (Complete, error)
}

type Complete interface {
    Success()
    Failed()
}
```

(interfaces are used for example purposes, I would expect these types to actually be structs).

PR #11663, opened 2021-11-25
ca: remove second call to Provider.ActiveRoot

dnephin:
Branched from #11661, this change is related to #11598, and #11159.

See the previous PR and the linked issues for context. Full details about the change in the commit messages.

Short version: we should not be calling `Provider.ActiveRoot` in the secondary. Instead we should get the RootCert from the stored `primaryRoots`. This PR makes that change.

### DIFF
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -602,10 +602,12 @@ func (c *CAManager) getLeafSigningCertFromRoot(root *structs.CARoot) string {
 	return root.IntermediateCerts[len(root.IntermediateCerts)-1]
 }
 
-// secondaryInitializeIntermediateCA runs the routine for generating an intermediate CA CSR and getting
-// it signed by the primary DC if the root CA of the primary DC has changed since the last
-// intermediate. It should only be called while the state lock is held by setting the state
-// to non-ready.
+// secondaryInitializeIntermediateCA generates a Certificate Signing Request (CSR)
+// for the intermediate CA that is used to sign leaf certificates in the secondary.
+// The CSR is signed by the primary DC and then persisted in the state store.
+//
+// This method should only be called while the state lock is held by setting the
+// state to non-ready.
 func (c *CAManager) secondaryInitializeIntermediateCA(provider ca.Provider, config *structs.CAConfiguration) error {
 	activeIntermediate, err := provider.ActiveIntermediate()
 	if err != nil {
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -621,48 +621,22 @@ func (c *CAManager) secondaryInitializeIntermediateCA(provider ca.Provider, conf
 		return err
 	}
 
-	var (
-		storedRootID         string
-		expectedSigningKeyID string
-		currentSigningKeyID  string
-		activeSecondaryRoot  *structs.CARoot
-	)
+	_, activeRoot, err := c.delegate.State().CARootActive(nil)
+	if err != nil {
+		return err
+	}
+	var currentSigningKeyID string
+	if activeRoot != nil {
+		currentSigningKeyID = activeRoot.SigningKeyID
+	}
+
+	var expectedSigningKeyID string
 	if activeIntermediate != "" {
-		// In the event that we already have an intermediate, we must have
-		// already replicated some primary root information locally, so check
-		// to see if we're up to date by fetching the rootID and the
-		// signingKeyID used in the secondary.
-		//
-		// Note that for the same rootID the primary representation of the root
-		// will have a different SigningKeyID field than the secondary
-		// representation of the same root. This is because it's derived from
-		// the intermediate which is different in all datacenters.
-		storedRoot, err := provider.ActiveRoot()
-		if err != nil {
-			return err
-		}
-
-		storedRootID, err = connect.CalculateCertFingerprint(storedRoot)
-		if err != nil {
-			return fmt.Errorf("error parsing root fingerprint: %v, %#v", err, storedRoot)
-		}
-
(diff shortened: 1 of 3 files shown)


agent/structs/connect_ca.go
```diff
@@ -67,9 +67,14 @@ type CARoot struct {
...
-	// key used to sign leaf certificates. Is is the HexString format of the
-	// raw AuthorityKeyID bytes.
+	// SigningKeyID is the connect.HexString encoded id of the public key that
+	// corresponds to the private key used to sign leaf certificates in the
+	// local datacenter.
+	//
+	// The value comes from x509.Certificate.SubjectKeyId of the local leaf
+	// signing cert.
+	//
+	// See https://www.rfc-editor.org/rfc/rfc3280#section-4.2.1.1 for more detail.
```
dnephin, line 77: Calling out this docs change because it is significant. Previously we documented this as coming from the `AuthorityKeyId`. 

From my reading of the linked x509 RFC, it sounds like these two values (AuthorityKeyId and SubjectKeyId) are the same on a self-signed root cert, but otherwise are different.

On non-root certs the AuthorityKeyID is the `parentCert.SubjectKeyId`, and the SubjectKeyId is the encoded ID of the public key.

So I believe this field is the `SubjectKeyId` of the local signing cert (**not** the `AuthorityKeyID` of that cert). My reading of the code seems to match this understanding. We generally set this to `intermediateCert.SubjectKeyId`. The exception is in `parseCARoot`, which I believe is a bug, and I've documented that in this issue: #11662

agent/consul/leader_connect_ca.go
```diff
@@ -639,25 +628,15 @@ func (c *CAManager) secondaryInitializeIntermediateCA(provider ca.Provider, conf
...
+	newActiveRoot, err := c.secondaryGetActivePrimaryCARoot()
+	if err != nil {
+		return err
 	}
 
 	// Get a signed intermediate from the primary DC if the provider
 	// hasn't been initialized yet or if the primary's root has changed.
 	needsNewIntermediate := false
-	if activeIntermediate == "" || storedRootID != primaryRoots.ActiveRootID {
+	if activeIntermediate == "" {
```
dnephin, line 639: This is the core part of the change. By removing this comparison we no longer need to call `Provider.ActiveRoot`.

dnephin (commented):
Best viewed by individual commit, but I left some more detail inline as well.

agent/consul/leader_connect_ca.go
```diff
@@ -585,79 +592,45 @@ func (c *CAManager) primaryInitialize(provider ca.Provider, conf *structs.CAConf
...
-		}
-
-		storedRootID, err = connect.CalculateCertFingerprint(storedRoot)
-		if err != nil {
-			return fmt.Errorf("error parsing root fingerprint: %v, %#v", err, storedRoot)
-		}
+	_, activeRoot, err := c.delegate.State().CARootActive(nil)
+	if err != nil {
+		return err
+	}
```
dnephin, line 610: This is moved up from below, since it is needed to get the ID of the root stored in the secondary.

agent/consul/leader_connect_ca.go
```diff
@@ -585,79 +592,45 @@ func (c *CAManager) primaryInitialize(provider ca.Provider, conf *structs.CAConf
...
-// secondaryInitializeIntermediateCA runs the routine for generating an intermediate CA CSR and getting
-// it signed by the primary DC if the root CA of the primary DC has changed since the last
-// intermediate. It should only be called while the state lock is held by setting the state
-// to non-ready.
+// secondaryInitializeIntermediateCA generates a Certificate Signing Request (CSR)
+// for the intermediate CA that is used to sign leaf certificates in the secondary.
+// The CSR is signed by the primary DC and then persisted in the state store.
+//
+// This method should only be called while the state lock is held by setting the
+// state to non-ready.
```
kisunji, line 600: Do you know what this means? I looked at usages of this method and it seems like ` *CAManager.setState` is being called higher in the call stack but there are no references to `*CAManager.stateLock`

kisunji (approved):
LGTM!

agent/consul/leader_connect_ca.go
dnephin: Ya, you found the right thing. `setState` is kind of a state machine that "locks" access. I'm not sure why we use that terminology, it might be from a much older implementation I'm not aware of.

I've got a proposal for improving this state machine in #10634 that includes a diagram of the possible states. 

I guess "lock" refers to the fact that only one update can happen at a time, and that some operations can't happen until initialize is complete.

FFMMM (approved):
🚀 !

agent/consul/leader_connect_ca.go
```diff
@@ -602,79 +609,45 @@ func (c *CAManager) getLeafSigningCertFromRoot(root *structs.CARoot) string {
...
 func (c *CAManager) secondaryInitializeIntermediateCA(provider ca.Provider, config *structs.CAConfiguration) error {
```
kyhavlov, line 618: This whole function is a lot more concise and readable now 👍

kyhavlov (approved):

dnephin:
The earlier PR in this chain only auto-backported to 1.11, so adding the tag for that, and will manually backport for the other release branches.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/568431.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 9a59733b7d3362d71ce470bbb2da9d823cf00316 onto `release/1.11.x` succeeded!

PR #11665, opened 2021-11-26
Rename `master` and `agent_master` ACL tokens in the config file format

boxofrad:
This is a first step towards removing this non-inclusive language. In subsequent PRs we will update the naming [in the API](https://www.consul.io/api-docs/agent#update-acl-tokens), the `consul acl set-agent-token` command, the auto-config API, and internal references, struct field names etc.


agent/config/config.go
```diff
@@ -742,11 +742,16 @@ type ACL struct {
...
-	Replication *string `mapstructure:"replication"`
+	InitialManagement *string `mapstructure:"initial_management"`
+	Replication       *string `mapstructure:"replication"`
+	AgentRecovery     *string `mapstructure:"agent_recovery"`
+	Default           *string `mapstructure:"default"`
+	Agent             *string `mapstructure:"agent"`
+
+	// DEPRECATED (ACL) - renamed to "initial_management"
+	Master *string `mapstructure:"master"`
+	// DEPRECATED (ACL) - renamed to "agent_recovery"
```
dnephin, line 753: It doesn't work to put these into the `DeprecatedConfig` because they are nested structs, is that right?

We could do something like embed a `deprecated ACLTokens` type and move the fields to that type. Or maybe a bit easier, maybe we could rename these fields to `DeprecatedMaster` and `DeprecatedAgentMaster`? Since we have mapstructure names for them I think the field names don't matter.

agent/config/deprecated.go
```diff
@@ -34,11 +34,21 @@ func applyDeprecatedConfig(d *decodeTarget) (Config, []string) {
 	dep := d.DeprecatedConfig
 	var warns []string
 
+	// TODO(boxofrad): The DeprecatedConfig struct only holds fields that were once
+	// on the top-level Config struct (not nested fields e.g. ACL.Tokens) maybe we
+	// should rethink this a bit?
```
dnephin, line 39: What do you have in mind?

I think this pattern of normalizing in `applyDeprecatedConfig` is still good, because it means our merge config logic can ignore the deprecated fields. Even if the fields are still accessible to that logic.  

I left a suggestion above on how we might be able to make nested deprecated fields more obviously deprecated.

dnephin (approved):
Looks great, nice work!

One suggestion below for making the deprecated fields a little more explicit even though we can't use the DeprecatedConfig struct. Also happy to chat about other options if you think there is a better way.

I think any of that could be done as a follow up, so nothing blocking a merge I can see.

agent/config/config.go
boxofrad: Yep, I tried adding a nested `ACL` struct to `DeprecatedConfig` but it didn't look like mapstructure could correctly split the assignments between fields on this and the proper `ACL` struct on `Config`:

```Go
type DeprecatedConfig struct {
	ACL struct {
		Tokens struct {
			Master *string `mapstructure:"master"`
		} `mapstructure:"tokens"`
	} `mapstructure:"acl"`
}
```

```
* invalid config key acl.default_policy
* invalid config key acl.down_policy
* invalid config key acl.token_ttl
* invalid config key acl.policy_ttl
```

I think splitting the deprecated fields out into an embedded struct would make sense 👍🏻

agent/config/deprecated.go
boxofrad: Totally agree with keeping `applyDeprecatedConfig` 💯 

My main concern is that we've lost the clean separation of all the deprecated stuff living on `DeprecatedConfig` (which is embedded in `decodeTarget` rather than `Config` itself) because of the fields on nested structs.

It'd be cool if we could keep `ACL.Tokens`'s deprecated fields on `DeprecatedConfig` somehow. Maybe we could decode `Config` and `DeprecatedConfig` separately and merge them together — but I think it would complicate things like handling unrecognised field names. 🤷🏻

website/content/docs/agent/options.mdx
```diff
@@ -642,18 +642,33 @@ Valid time units are 'ns', 'us' (or 'µs'), 'ms', 's', 'm', 'h'."
...
-      this is done, restart the current leader to force a leader election. If
-      the `master` token is not supplied, then the servers do not create a
-      master token. When you provide a value, it should be a UUID. To maintain
-      backwards compatibility and an upgrade path this restriction is not
+      The `master` token is only installed when a server acquires cluster leadership.
+      If you would like to install or change it, set the new value for `master` in the
+      configuration for all servers. Once this is done, restart the current leader to
+      force a leader election. If the `master` token is not supplied, then the servers
+      do not create a master token. When you provide a value, it should be a UUID. To
+      maintain backwards compatibility and an upgrade path this restriction is not
```
jkirschner-hashicorp, line 671: Why repeat the text from `initial_management`? Can this just reference rather than copy? Similar to how [`-non-voting-server` is handled](https://www.consul.io/docs/agent/options#_non_voting_server)?

e.g.,
```suggestion
    - `master` ((#acl_tokens_master)) **Deprecated in Consul 1.11. See the
      [`acl.tokens.initial_management`](#acl_tokens_initial_management) field instead.** -
```

website/content/docs/agent/options.mdx
```diff
@@ -670,8 +685,16 @@ Valid time units are 'ns', 'us' (or 'µs'), 'ms', 's', 'm', 'h'."
...
       catalog such as metadata, or the node's tagged addresses.
 
-    - `agent_master` ((#acl_tokens_agent_master)) - Used to access [agent endpoints](/api/agent)
-      that require agent read or write privileges, or node read privileges, even if Consul servers aren't present to validate any tokens. This should only be used by operators during outages, regular ACL tokens should normally be used by applications.
+    - `agent_recovery` ((#acl_tokens_agent_recovery)) - Used to access [agent endpoints](/api/agent)
+      that require agent read or write privileges, or node read privileges, even if Consul servers
+      aren't present to validate any tokens. This should only be used by operators during outages,
+      regular ACL tokens should normally be used by applications.
+
+    - `agent_master` ((#acl_tokens_agent_master)) **Deprecated in Consul 1.11. Use the
```
jkirschner-hashicorp, line 693: Same comment as above (with master -> initial_management). Can we just reference rather than copy the docs text?

website/content/docs/agent/options.mdx
```diff
@@ -720,7 +743,7 @@ Valid time units are 'ns', 'us' (or 'µs'), 'ms', 's', 'm', 'h'."
   not impacted.
 
 - `acl_agent_master_token` ((#acl_agent_master_token_legacy)) - **Deprecated
-  in Consul 1.4.0. See the [`acl.tokens.agent_master`](#acl_tokens_agent_master)
+  in Consul 1.4.0. See the [`acl.tokens.agent_recovery`](#acl_tokens_agent_recovery)
```
jkirschner-hashicorp, line 746: If someone has this in their config file, they probably aren't upgrading directly from pre 1.4 to 1.11+. If they are upgrading to something <1.11, this flag doesn't exist, and the docs for it don't mention the older name of the flag. We need to fix that somehow... perhaps by referencing the old name from the new name?

website/content/docs/agent/options.mdx
```diff
@@ -642,18 +642,33 @@ Valid time units are 'ns', 'us' (or 'µs'), 'ms', 's', 'm', 'h'."
...
+      operators to bootstrap the ACL system with a token Secret ID that is well-known.
+
+      The `initial_management` token is only installed when a server acquires cluster
+      leadership. If you would like to install or change it, set the new value for
+      `initial_management` in the configuration for all servers. Once this is done,
+      restart the current leader to force a leader election. If the `initial_management`
+      token is not supplied, then the servers do not create an initial management token.
+      When you provide a value, it should be a UUID. To maintain backwards compatibility
+      and an upgrade path this restriction is not currently enforced but will be in a
+      future major Consul release.
```
jkirschner-hashicorp, line 657: We should mention which versions this is available in, and what to do if on an older version. Thoughts on whether this should be at the beginning or end of the option description? (We currently do both within this docs page.)

```suggestion
      future major Consul release.
      
      This is available in Consul 1.11+. In prior versions, use [`acl.tokens.master`](#acl_tokens_master).
```

jkirschner-hashicorp: Long-term, we need a better solution for handling versions (e.g., versioned docs pages, or a table for the options with a column for version compatibility), but not relevant for this PR.

website/content/docs/agent/options.mdx
```diff
@@ -670,8 +685,16 @@ Valid time units are 'ns', 'us' (or 'µs'), 'ms', 's', 'm', 'h'."
...
+    - `agent_recovery` ((#acl_tokens_agent_recovery)) - Used to access [agent endpoints](/api/agent)
+      that require agent read or write privileges, or node read privileges, even if Consul servers
+      aren't present to validate any tokens. This should only be used by operators during outages,
+      regular ACL tokens should normally be used by applications.
+
+    - `agent_master` ((#acl_tokens_agent_master)) **Deprecated in Consul 1.11. Use the
+      [`acl.tokens.agent_recovery`](#acl_tokens_agent_recovery) field instead.** - Used to access
+      [agent endpoints](/api/agent) that require agent read or write privileges, or node read
+      privileges, even if Consul servers aren't present to validate any tokens. This should only
+      be used by operators during outages, regular ACL tokens should normally be used by applications.
```
jkirschner-hashicorp, line 697: Same as above: we should mention which versions this is available in, and what to do if on an older version.

website/content/docs/security/acl/acl-system.mdx
```diff
@@ -202,16 +202,16 @@ the token itself.
...
 - **Anonymous Token** - The anonymous token is used when a request is made to Consul without specifying a bearer token.
   The anonymous token's description and policies may be updated but Consul will prevent this token's deletion. When created,
   it will be assigned `00000000-0000-0000-0000-000000000002` for its Accessor ID and `anonymous` for its Secret ID.
 
-- **Master Token** - When a master token is present within the Consul configuration, it is created and will be linked
-  With the builtin Global Management policy giving it unrestricted privileges. The master token is created with the Secret ID
-  set to the value of the configuration entry.
+- **Initial Management Token** - When an initial management token is present within the Consul configuration, it is created
+  and will be linked with the builtin Global Management policy giving it unrestricted privileges. The initial management
+  token is created with the Secret ID set to the value of the configuration entry.
```
jkirschner-hashicorp, line 214: I think we should mention that this was previously called the "Master" token (in case someone is confused why the "master" token doesn't appear in the docs because they are on Consul 1.4-1.10).

```suggestion
- **Initial Management Token** - When an initial management token is present within the Consul configuration, it is created
  and will be linked with the builtin Global Management policy giving it unrestricted privileges. The initial management
  token is created with the Secret ID set to the value of the configuration entry.
  In Consul 1.4 - 1.10, this was called the `master` token. It was renamed to `initial_management` token in Consul 1.11.
```

website/content/docs/security/acl/acl-system.mdx
```diff
@@ -291,18 +291,18 @@ as to whether they are set on servers, clients, or both.
...
+| [`acl.tokens.agent_recovery`](/docs/agent/options#acl_tokens_agent_recovery)         | `OPTIONAL` | `OPTIONAL` | Special token that can be used to access [Agent API](/api/agent) when remote bearer token resolution fails; used for setting up the cluster such as doing initial join operations, see the [ACL Agent Recovery Token](#acl-agent-recovery-token) section for more details |
+| [`acl.tokens.agent`](/docs/agent/options#acl_tokens_agent)                           | `OPTIONAL` | `OPTIONAL` | Special token that is used for an agent's internal operations, see the [ACL Agent Token](#acl-agent-token) section for more details                                                                                                                                       |
+| [`acl.tokens.initial_management`](/docs/agent/options#acl_tokens_initial_management) | `OPTIONAL` | `N/A`      | Special token used to bootstrap the ACL system, check the [Bootstrapping ACLs](https://learn.hashicorp.com/tutorials/consul/access-control-setup-production) tutorial for more details                                                                                    |
```
jkirschner-hashicorp, line 298: Should we also include the old name for these 2 tokens in this table?

website/content/docs/security/acl/acl-system.mdx
```diff
@@ -291,18 +291,18 @@ as to whether they are set on servers, clients, or both.
...
 
-All of these tokens except the `master` token can all be introduced or updated via the [/v1/agent/token API](/api/agent#update-acl-tokens).
+All of these tokens except the `initial_management` token can all be introduced or updated via the [/v1/agent/token API](/api/agent#update-acl-tokens).
 
-#### ACL Agent Master Token
+#### ACL Agent Recovery Token
 
-Since the [`acl.tokens.agent_master`](/docs/agent/options#acl_tokens_agent_master) is designed to be used when the Consul servers are not available, its policy is managed locally on the agent and does not need to have a token defined on the Consul servers via the ACL API. Once set, it implicitly has the following policy associated with it
+Since the [`acl.tokens.agent_recovery`](/docs/agent/options#acl_tokens_agent_recovery) is designed to be used when the Consul servers are not available, its policy is managed locally on the agent and does not need to have a token defined on the Consul servers via the ACL API. Once set, it implicitly has the following policy associated with it
```
jkirschner-hashicorp, line 305:
```suggestion
Since the [`acl.tokens.agent_recovery`](/docs/agent/options#acl_tokens_agent_recovery) is designed to be used when the Consul servers are not available, its policy is managed locally on the agent and does not need to have a token defined on the Consul servers via the ACL API. Once set, it implicitly has the following policy associated with it.
In Consul 1.4 - 1.10, this was called the `agent_master` token. It was renamed to `agent_recovery` token in Consul 1.11.
```

jkirschner-hashicorp (changes requested):
@boxofrad : thanks for taking this on! I proposed some changes that make the docs easier to understand for users on Consul 1.4 - 1.10. Let me know what you think.

website/content/docs/agent/options.mdx
```diff
@@ -642,18 +642,33 @@ Valid time units are 'ns', 'us' (or 'µs'), 'ms', 's', 'm', 'h'."
...
+      leadership. If you would like to install or change it, set the new value for
+      `initial_management` in the configuration for all servers. Once this is done,
+      restart the current leader to force a leader election. If the `initial_management`
+      token is not supplied, then the servers do not create an initial management token.
+      When you provide a value, it should be a UUID. To maintain backwards compatibility
+      and an upgrade path this restriction is not currently enforced but will be in a
+      future major Consul release.
+
+    - `master` ((#acl_tokens_master)) **Deprecated in Consul 1.11. Use the
+      [`acl.tokens.initial_management`](#acl_tokens_initial_management) field insead.** -
```
jkirschner-hashicorp, line 660: mispelling: instead

website/content/docs/agent/options.mdx
boxofrad: Ah, nice! Copied this approach from some of the other deprecated ACL fields (e.g. [`acl_datacenter`](https://www.consul.io/docs/agent/options#acl_datacenter)) but agree just referencing is much bettert.

website/content/docs/agent/options.mdx
boxofrad: Good point! Let's go with the old name.

website/content/docs/agent/options.mdx
boxofrad: Whoops! Thanks 😅

boxofrad:
Thanks @jkirschner-hashicorp, that's much better. Let me know what you think now?

website/content/api-docs/acl/index.mdx
```diff
@@ -16,7 +16,7 @@ the [ACL tutorial](https://learn.hashicorp.com/tutorials/consul/access-control-s
 ## Bootstrap ACLs
 
 This endpoint does a special one-time bootstrap of the ACL system, making the first
-management token if the [`acl.tokens.master`](/docs/agent/options#acl_tokens_master)
+management token if the [`acl.tokens.initial_management`](/docs/agent/options#acl_tokens_initial_management)
```
jkirschner-hashicorp, line 19: Is the intent that, if someone is on Consul 1.11, they will follow this link, see

> This is available in Consul 1.11 and later. In prior versions, use [`acl.tokens.master`](#acl_tokens_master)

and then find out the correct (previous) name that way?

I imagine we've had this problem before (and will have new instances of this problem in the future) when we rename things... do we mention every name everywhere? Or mention just the most recent name in most docs, but link to information about old names?

[Long term, this might be best solved by having versioned docs pages. But that's well outside the scope of this PR :) ]

website/content/docs/agent/options.mdx
```diff
@@ -642,19 +642,25 @@ Valid time units are 'ns', 'us' (or 'µs'), 'ms', 's', 'm', 'h'."
...
+      leadership. If you would like to install or change it, set the new value for
+      `initial_management` in the configuration for all servers. Once this is done,
+      restart the current leader to force a leader election. If the `initial_management`
+      token is not supplied, then the servers do not create an initial management token.
+      When you provide a value, it should be a UUID. To maintain backwards compatibility
+      and an upgrade path this restriction is not currently enforced but will be in a
+      future major Consul release.
+
+    - `master` ((#acl_tokens_master)) **Deprecated in Consul 1.11. Use the
+      [`acl.tokens.initial_management`](#acl_tokens_initial_management) field instead.**
```
jkirschner-hashicorp, line 663: For someone on Consul <1.11, I'm not sure this text makes it clear that the `initial_management` token is a direct replacement (a rename), so the docs for `initial_management` apply here, too.

Perhaps this? Let me know what you think.
```suggestion
    - `master` ((#acl_tokens_master)) **Renamed in Consul 1.11 to
      [`acl.tokens.initial_management`](#acl_tokens_initial_management).**
```

Or if we **must** include the word deprecated...
```suggestion
    - `master` ((#acl_tokens_master)) **Deprecated in Consul 1.11: renamed to
      [`acl.tokens.initial_management`](#acl_tokens_initial_management).**
```

website/content/docs/agent/options.mdx
```diff
@@ -670,8 +676,16 @@ Valid time units are 'ns', 'us' (or 'µs'), 'ms', 's', 'm', 'h'."
...
-      that require agent read or write privileges, or node read privileges, even if Consul servers aren't present to validate any tokens. This should only be used by operators during outages, regular ACL tokens should normally be used by applications.
+    - `agent_recovery` ((#acl_tokens_agent_recovery)) - This is available in Consul 1.11
+      and later. In prior versions, use [`acl.tokens.agent_master`](#acl_tokens_agent_master).
+
+      Used to access [agent endpoints](/api/agent) that require agent read or write privileges,
+      or node read privileges, even if Consul servers aren't present to validate any tokens.
+      This should only be used by operators during outages, regular ACL tokens should normally
+      be used by applications.
+
+    - `agent_master` ((#acl_tokens_agent_master)) **Deprecated in Consul 1.11. Use the
```
jkirschner-hashicorp, line 687: Same comment as above (mention that it was renamed)

jkirschner-hashicorp (approved):
Thanks for the changes, @boxofrad!

I added a few small comments to consider before merging, but have given my approval.

website/content/api-docs/acl/index.mdx
boxofrad: Yep! In this case, it's a bit more onerous because we've already renamed the token before (`acl_master_token` ⇒ `acl.tokens.master` ⇒ `acl.tokens.initial_management`) so users on very old versions have a bit of a trail to follow.

> do we mention every name everywhere? Or mention just the most recent name in most docs

That seems to be the case, yeah. For example, the current ACL System page [mentions](https://www.consul.io/docs/security/acl/acl-system#acl-agent-master-token) `acl.tokens.agent_master`, not `acl_agent_master_token`.

website/content/docs/agent/options.mdx
boxofrad: I agree, rename is clearer 👍🏻

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/512839.

PR #11666, opened 2021-11-26
ui: Refactor KV and Lock Sessions following partitions update

johncowen:
I had one more 'in application code' FIXME left to finish up from https://github.com/hashicorp/consul/pull/11017 (oh! I just found one more 😆 will sort in a sec), and this fixes that up.

This PR uses all our new ways of doing things to Lock Sessions and their interactions with KV and Nodes. This is mostly around are new under-the-hood things, but also I took the opportunity to upgrade some of the CSS to reuse some of our CSS utils that have been made over the past few months (`%csv-list` and `%horizontal-kv-list`).

Also added (and worked on existing) documentation for Lock Session related components.

I'll take a pass myself now and see if I can add some useful inlines for reviewing.


ui/packages/consul-ui/app/serializers/session.js
```diff
@@ -7,7 +7,20 @@ export default class SessionSerializer extends Serializer {
...
+            e.errors = [
+              {
+                status: '404',
+                title: 'Not found',
+              },
+            ];
+            throw e;
+          }
+          return cb(headers, body[0]);
+        }),
```
johncowen, line 23: When requesting a single session, I noticed that this is one of those Consul "it doesn't exist but I'll give you a 200 with an empty array" things. We now check for that and throw a fake 404. We used to cope with this differently but this makes more sense. Another little FYI is that we have a custom `HTTPError` shaped `Error` we could use for all of these at some point.

ui/packages/consul-ui/app/styles/layout.scss
```diff
@@ -53,21 +53,30 @@ html[data-route$='edit'] .app-view > header + div > *:first-child {
...
+%list-after-secondary-nav {
+  margin-top: 0 !important;
+}
+%list-after-secondary-nav ul {
+  border-top-width: 0 !important;
+}
+%list-after-filter-bar {
+  border-top-width: 0 !important;
+}
+
```
johncowen, line 66: Made it clearer what these styles are doing by naming them instead of just commenting. We can then use these as 'helper/utility' placeholders in individual components/routes until we can figure out a way not to need them at all.

ui/packages/consul-ui/app/templates/dc/kv/edit.hbs
```diff
@@ -121,17 +131,21 @@ as |parts|}}
...
+          Lock Session
+        </Action>
+      </h2>
+  {{#if session.ID}}
       <Consul::LockSession::Form
         @item={{session}}
-        @dc={{route.params.dc}}
-        @nspace={{route.params.nspace}}
-        @partition={{route.params.partition}}
-        @onsubmit={{action (noop) undefined}}
```
johncowen, line 131: We reference all of the 'buckets' from the session itself now so we don't need to pass these through individually anymore, just reference `@item.Partition` etc.

ui/packages/consul-ui/app/templates/dc/nodes/show/sessions.hbs
```diff
@@ -18,38 +18,97 @@ as |route|>
...
+            </collection.Empty>
+
+          </DataCollection>
+
         </BlockSlot>
-      </EmptyState>
-  {{/if}}
-  </div>
+      </DataWriter>
+    </div>
```
johncowen, line 111: More upgrading to new stuff, Actions, Octane, DataWriter (for deleting), DataCollection (we don't do search/sorting here but if we ever do then this is all there to be used now)

johncowen (commented):
Self-review inline notes:

ui/packages/consul-ui/app/serializers/session.js
evrowe: > I'll give you a 200 with an empty array" things

There's a perfectly good HTTP status code for this (I realize this is not the place to take my complaints haha) 😭 😭 😭

evrowe (approved):
All looks good, questions are non-blocking

ui/packages/consul-ui/app/serializers/session.js
johncowen: 🤣 yeah we have a few of these that date back quite a few years.

What is meant here is "The session you requested is not found" but Consul gives us a `[]` because here it always responds with arrays even for requests for single sessions. I think this is one of the cases where we don't change things with the benefit of hindsight for back-compat reasons, but no reason why we can't fake a 404 in the UI data layer.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/512082.

PR #11669, opened 2021-11-26
Rename `agent_master` ACL token in the API and CLI

boxofrad:
Following on from https://github.com/hashicorp/consul/pull/11665, this PR renames the agent master token in the API and CLI.


agent/agent_endpoint_test.go
```diff
@@ -5308,14 +5308,23 @@ func TestAgent_Token(t *testing.T) {
 			effective: tokens{master: "M"},
 		},
 		{
-			name:      "set master ",
+			name:      "set master",
```
dhiaayachi, line 5311: While we are at it we should change this too, I see a lot of other references that can be changed in the test file

agent/agent_endpoint.go
```diff
@@ -1444,7 +1444,7 @@ func (s *HTTPHandlers) AgentToken(resp http.ResponseWriter, req *http.Request) (
 				triggerAntiEntropySync = true
 			}
 
-		case "acl_agent_master_token", "agent_master":
+		case "acl_agent_master_token", "agent_master", "agent_recovery":
```
dhiaayachi, line 1447: Is the plan to deprecate the `acl_agent_master_token` and `agent_master` naming in favour of `agent_recovery`? If so I don't see any deprecation notice for it in the doc.
It also would be nice to have those defined as const and have a deprecation notice for the ones we want to remove in the future. 
What you think about that?

dhiaayachi (commented):
Thank you for working on this @boxofrad!!
I added few comments about `master` mention in tests and deprecation notice.

agent/agent_endpoint_test.go
dnephin: I believe the plan is to only focus on external references for now. We can cleanup all the internal stuff after the 1.11 release.

api/agent.go
```diff
@@ -1316,10 +1324,21 @@ func (a *Agent) updateToken(target, token string, q *WriteOptions) (*WriteMeta,
 	return meta, err
 }
 
-func (a *Agent) updateTokenFallback(target, fallback, token string, q *WriteOptions) (*WriteMeta, error) {
-	meta, status, err := a.updateTokenOnce(target, token, q)
-	if err != nil && status == 404 {
-		meta, _, err = a.updateTokenOnce(fallback, token, q)
+func (a *Agent) updateTokenFallback(token string, q *WriteOptions, targets ...string) (*WriteMeta, error) {
```
dnephin, line 1327: varargs, nice! :tada:

api/agent.go
```diff
@@ -1316,10 +1324,21 @@ func (a *Agent) updateToken(target, token string, q *WriteOptions) (*WriteMeta,
...
+
+	var (
+		meta *WriteMeta
+		err  error
+	)
+	for _, target := range targets {
+		var status int
+		meta, status, err = a.updateTokenOnce(target, token, q)
+		if err == nil || status != http.StatusNotFound {
+			break
```
dnephin, line 1340: I think `err` is always nil for status errors, so wouldn't this be:
```suggestion
		if err == nil && status != http.StatusNotFound {
			return meta, err
```

minor: I always prefer to `return` over `break` if possible, because it avoids needing to look at what might happen afterward. Seems like return would work here.

dnephin (commented):
Nice! 

One comment about the API client library logic maybe not handling errors correctly.

api/agent.go
dnephin: Oh, actually we do check this is `requireOK`, so simply checking `err` should be fine!
```suggestion
		if err == nil || status != http.StatusNotFound {
			return meta, err
```

agent/agent_endpoint_test.go
boxofrad: Yep, exactly that 👍🏻 

I have a WIP branch that renames the struct fields etc. but in this case, we'd leave the test name the same.

agent/agent_endpoint.go
boxofrad: Yeah. It's not super obvious from the diff, but the `target` we're switching on is a path suffix - so there's a deprecation notice in the API docs [here](https://github.com/hashicorp/consul/blob/44d8abd341ca33c5b927aa473562a61ee8bcfd01/website/content/api-docs/agent/index.mdx?plain=1#L742-L758) 🙂 

I'd be happy to move them into constants, but as they're only really used in this handler, I wonder if the indirection would be worth it, or make the code less clear?

api/agent.go
boxofrad: Oh, nice catch! I'll go with the `&&` variation, as that's what the original code did.

dnephin (approved):
LGTM!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/513279.

PR #11671, opened 2021-11-26
ca: fix storing the leaf signing cert with Vault provider

### LINKED ISSUE #10532: connect/ca/roots endpoint do not return the intermediate cert when using vault provider

### LINKED ISSUE #10826: Consul Connect: CA Root does not store the intermediate
#### Overview of the Issue

v1/agent/connect/ca/roots is returning the RootCert but IntermediateCerts is null, despite all leaf certs being signed by the intermediate.
Connect is configured to use the vault provider.
This happens for us after a consul leadership change. Looking at the logs and corresponding code I can see the SigningKeyID is changing (not what we'd expect...) but then the control flow of the code doesn't appear to set the intermediate on the rootCA.

https://github.com/hashicorp/consul/blob/v1.10.1/agent/consul/leader_connect_ca.go#L500

In the case that the SigningKeyID does NOT change, it's meant to set the intermediate:
https://github.com/hashicorp/consul/blob/v1.10.1/agent/consul/leader_connect_ca.go#L470

Surely it's meant to also set the intermediate on the rootCA on line 499? Or potentially before the raftApply, so line 486?

We're running 1.10.1, but I can see the same in main:

https://github.com/hashicorp/consul/blob/main/agent/consul/leader_connect_ca.go#L563

Consul 1.10.1 (enterprise)

### Log Fragments

```
[18148]: 2021-08-10T17:51:34.590+0100 [INFO]  agent.server.connect: Correcting stored SigningKeyID value: previous=b9:98:.... updated=19:5d:......
[18148]: 2021-08-10T17:51:34.602+0100 [INFO]  agent.server.connect: initialized primary datacenter CA with provider: provider=vault
[18148]: 2021-08-10T17:51:34.602+0100 [INFO]  agent.server.connect: Successfully initialized the Connect CA
[18148]: 2021-08-10T17:51:34.602+0100 [INFO]  agent.leader: started routine: routine="intermediate cert renew watch"
```

dnephin:
Fixes #10826 
Fixes #10532

We were not saving the leaf signing cert on the `CARoot` when the Vault provider was initialized. This was only a problem for API users (not envoy and xDS) because [this extra code](https://github.com/hashicorp/consul/pull/11661#discussion_r757059682) was masking the bug.

In #11661 I'd like to remove that masking code (because it is unnecessary, and because we need to change the Provider interface, so fewer callers makes that easier).

This PR updates an existing test to catch the bug, then fixes the bug. I added a few more assertions to both the primary and secondary test cases to make sure we did not have a similar bug in other places.

### DIFF
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -536,7 +536,7 @@ func (c *CAManager) primaryInitialize(provider ca.Provider, conf *structs.CAConf
 
 	// Add the local leaf signing cert to the rootCA struct. This handles both
 	// upgrades of existing state, and new rootCA.
-	if rootCA.LeafSigningCert() != interPEM {
+	if c.getLeafSigningCertFromRoot(rootCA) != interPEM {
 		rootCA.IntermediateCerts = append(rootCA.IntermediateCerts, interPEM)
 		rootUpdateRequired = true
 	}
@@ -591,6 +591,22 @@ func (c *CAManager) primaryInitialize(provider ca.Provider, conf *structs.CAConf
 	c.logger.Info("initialized primary datacenter CA with provider", "provider", conf.Provider)
 
 	return nil
+}
+
+// getLeafSigningCertFromRoot returns the PEM encoded certificate that should be used to
+// sign leaf certificates in the local datacenter. The SubjectKeyId of the
+// returned cert should always match the SigningKeyID of the CARoot.
+//
+// TODO: fix the data model so that we don't need this complicated lookup to
+// find the leaf signing cert. See github.com/hashicorp/consul/issues/11347.
+func (c *CAManager) getLeafSigningCertFromRoot(root *structs.CARoot) string {
+	if !c.isIntermediateUsedToSignLeaf() {
+		return root.RootCert
+	}
+	if len(root.IntermediateCerts) == 0 {
+		return ""
+	}
+	return root.IntermediateCerts[len(root.IntermediateCerts)-1]
 }
 
 // secondaryInitializeIntermediateCA runs the routine for generating an intermediate CA CSR and getting
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -520,12 +520,26 @@ func (c *CAManager) primaryInitialize(provider ca.Provider, conf *structs.CAConf
 		}
 	}
 
+	var rootUpdateRequired bool
+
 	// Versions prior to 1.9.3, 1.8.8, and 1.7.12 incorrectly used the primary
 	// rootCA's subjectKeyID here instead of the intermediate. For
 	// provider=consul this didn't matter since there are no intermediates in
 	// the primaryDC, but for vault it does matter.
 	expectedSigningKeyID := connect.EncodeSigningKeyID(intermediateCert.SubjectKeyId)
-	needsSigningKeyUpdate := (rootCA.SigningKeyID != expectedSigningKeyID)
+	if rootCA.SigningKeyID != expectedSigningKeyID {
+		c.logger.Info("Correcting stored CARoot values",
+			"previous-signing-key", rootCA.SigningKeyID, "updated-signing-key", expectedSigningKeyID)
+		rootCA.SigningKeyID = expectedSigningKeyID
+		rootUpdateRequired = true
+	}
+
+	// Add the local leaf signing cert to the rootCA struct. This handles both
+	// upgrades of existing state, and new rootCA.
+	if rootCA.LeafSigningCert() != interPEM {
+		rootCA.IntermediateCerts = append(rootCA.IntermediateCerts, interPEM)
+		rootUpdateRequired = true
+	}
 
 	// Check if the CA root is already initialized and exit if it is,
 	// adding on any existing intermediate certs since they aren't directly
@@ -537,24 +551,19 @@ func (c *CAManager) primaryInitialize(provider ca.Provider, conf *structs.CAConf
 	if err != nil {
 		return err
 	}
-	if activeRoot != nil && needsSigningKeyUpdate {
-		c.logger.Info("Correcting stored SigningKeyID value", "previous", rootCA.SigningKeyID, "updated", expectedSigningKeyID)
-
-	} else if activeRoot != nil && !needsSigningKeyUpdate {
+	if activeRoot != nil && !rootUpdateRequired {
 		// This state shouldn't be possible to get into because we update the root and
 		// CA config in the same FSM operation.
 		if activeRoot.ID != rootCA.ID {
 			return fmt.Errorf("stored CA root %q is not the active root (%s)", rootCA.ID, activeRoot.ID)
 		}
 
+		// TODO: why doesn't this c.setCAProvider(provider, activeRoot) ?
 		rootCA.IntermediateCerts = activeRoot.IntermediateCerts
 		c.setCAProvider(provider, rootCA)
 
+		c.logger.Info("initialized primary datacenter CA from existing CARoot with provider", "provider", conf.Provider)
 		return nil
-	}
-
-	if needsSigningKeyUpdate {
-		rootCA.SigningKeyID = expectedSigningKeyID
 	}
 
 	// Get the highest index
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -1149,10 +1149,8 @@ func (c *CAManager) RenewIntermediate(ctx context.Context, isPrimary bool) error
 
 	// If this is the primary, check if this is a provider that uses an intermediate cert. If
 	// it isn't, we don't need to check for a renewal.
-	if isPrimary {
-		if _, ok := provider.(ca.PrimaryUsesIntermediate); !ok {
-			return nil
-		}
+	if isPrimary && !primaryUsesIntermediate(provider) {
+		return nil
 	}
 
 	activeIntermediate, err := provider.ActiveIntermediate()
@@ -1535,4 +1533,17 @@ func (c *CAManager) checkExpired(pem string) error {
 		return fmt.Errorf("certificate expired, expiration date: %s ", cert.NotAfter.String())
 	}
 	return nil
+}
+
+func primaryUsesIntermediate(provider ca.Provider) bool {
+	_, ok := provider.(ca.PrimaryUsesIntermediate)
+	return ok
+}
+
+func (c *CAManager) isIntermediateUsedToSignLeaf() bool {
+	if c.serverConf.Datacenter != c.serverConf.PrimaryDatacenter {
+		return true
+	}
+	provider, _ := c.getCAProvider()
+	return primaryUsesIntermediate(provider)
 }
--- a/agent/consul/leader_connect_test.go
+++ b/agent/consul/leader_connect_test.go
@@ -403,7 +403,7 @@ func TestLeader_Vault_PrimaryCA_IntermediateRenew(t *testing.T) {
 	store := s1.caManager.delegate.State()
 	_, activeRoot, err := store.CARootActive(nil)
 	require.NoError(err)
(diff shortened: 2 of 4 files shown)


agent/consul/leader_connect_test.go
```diff
@@ -396,23 +396,35 @@ func TestLeader_Vault_PrimaryCA_IntermediateRenew(t *testing.T) {
...
 	require.NoError(err)
-	_, err = connect.ParseCert(intermediatePEM)
+	intermediateCert, err := connect.ParseCert(intermediatePEM)
 	require.NoError(err)
 
+	// Check that the state store has the correct intermediate
+	store := s1.caManager.delegate.State()
+	_, activeRoot, err := store.CARootActive(nil)
+	require.NoError(err)
+	require.Equal(intermediatePEM, activeRoot.LeafSigningCert())
```
dnephin, line 406: This is the line that fails before the bug fix. The other test changes are to be sure it did not exist in other places.

agent/consul/leader_connect_ca.go
```diff
@@ -550,13 +563,10 @@ func (c *CAManager) primaryInitialize(provider ca.Provider, conf *structs.CAConf
 		rootCA.IntermediateCerts = activeRoot.IntermediateCerts
 		c.setCAProvider(provider, rootCA)
 
+		c.logger.Info("initialized primary datacenter CA from existing CARoot with provider", "provider", conf.Provider)
 		return nil
 	}
 
-	if needsSigningKeyUpdate {
-		rootCA.SigningKeyID = expectedSigningKeyID
-	}
```
dnephin, line 558: This line was moved up to the place where we set `needsSigningKeyUpdate`, to keep the logic together.

agent/consul/leader_connect_ca.go
```diff
@@ -550,13 +563,10 @@ func (c *CAManager) primaryInitialize(provider ca.Provider, conf *structs.CAConf
 		rootCA.IntermediateCerts = activeRoot.IntermediateCerts
 		c.setCAProvider(provider, rootCA)
 
+		c.logger.Info("initialized primary datacenter CA from existing CARoot with provider", "provider", conf.Provider)
```
dnephin, line 566: This is unrelated, but something I noticed while working on this.

I'm pretty sure that previously we would not log anything when a server initializes itself using an existing stored `CARoot`. I guess this would happen if the raft leader election on any cluster that already had a root setup. 

I made the log message slightly different, so that it's clear this initialization was from an existing `CARoot`.

agent/structs/connect_ca.go
```diff
@@ -133,6 +133,22 @@ func (c *CARoot) Clone() *CARoot {
 	return &newCopy
 }
 
+// LeafSigningCert returns the PEM encoded certificate that should be used to
+// sign leaf certificates in the local datacenter. The SubjectKeyId of the
+// returned cert should always match the SigningKeyID of the CARoot.
+//
+// This method has no knowledge of the provider, so it makes assumptions about
+// which cert is used based on established convention.
```
dnephin, line 141: We are already using this convention in metrics. I think it's generally correct, but I think it could be wrong in the case where we have a cross-signed cert in the `InteremdiateCerts` list without any local signing cert.

I'm not sure how we fix this without changing the `CARoot` struct to use different fields for these certs. I think maybe we should do that now.

agent/consul/leader_connect_ca.go
```diff
@@ -537,10 +549,11 @@ func (c *CAManager) primaryInitialize(provider ca.Provider, conf *structs.CAConf
 	if err != nil {
 		return err
 	}
-	if activeRoot != nil && needsSigningKeyUpdate {
-		c.logger.Info("Correcting stored SigningKeyID value", "previous", rootCA.SigningKeyID, "updated", expectedSigningKeyID)
+	if activeRoot != nil && rootUpdateRequired {
+		c.logger.Info("Correcting stored CARoot values",
+			"previous-signing-key", rootCA.SigningKeyID, "updated-signing-key", expectedSigningKeyID)
```
FFMMM, line 554: might've gotten confused but the control flows here but wouldn't `rootCA.SigningKeyID` & `expectedSigningKeyID` be the same by the time we reach lines 553-54?

Since on line 531 we set:

```
rootCA.SigningKeyID = expectedSigningKeyID
```

did I miss something obvious 🤔 ?

agent/consul/leader_connect_test.go
```diff
@@ -396,23 +396,35 @@ func TestLeader_Vault_PrimaryCA_IntermediateRenew(t *testing.T) {
...
+	require.Equal(connect.HexString(intermediateCert.SubjectKeyId), activeRoot.SigningKeyID)
+	t.Log(activeRoot.SigningKeyID)
```
FFMMM, line 408: did we want to keep this log msg?

agent/consul/leader_connect_test.go
```diff
@@ -396,23 +396,35 @@ func TestLeader_Vault_PrimaryCA_IntermediateRenew(t *testing.T) {
...
 		provider, _ = getCAProviderWithLock(s1)
 		newIntermediatePEM, err := provider.ActiveIntermediate()
 		r.Check(err)
-		_, err = connect.ParseCert(intermediatePEM)
-		r.Check(err)
 		if newIntermediatePEM == intermediatePEM {
 			r.Fatal("not a renewed intermediate")
 		}
+		intermediateCert, err = connect.ParseCert(newIntermediatePEM)
+		r.Check(err)
```
FFMMM, line 420: out of curiosity, why do we rely on the `*retry.R` checks here and not a `require.NoError()` call? 📚

agent/consul/leader_connect_test.go
```diff
@@ -439,6 +451,8 @@ func TestLeader_Vault_PrimaryCA_IntermediateRenew(t *testing.T) {
 	// Check that the leaf signed by the new intermediate can be verified using the
 	// returned cert chain (signed intermediate + remote root).
 	intermediatePool := x509.NewCertPool()
+	// TODO: do not explicitly add the intermediatePEM, we should have it available
+	// from leafPEM. Use conenct.ParseLeafCerts to do the right thing.
```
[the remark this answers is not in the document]
FFMMM: & lines 604-05

agent/structs/connect_ca.go
FFMMM: > We are already using this convention in metrics

For my learning 🤓 , what did you mean by this?

FFMMM:
Thanks for working on this 🎉  -- left a couple of comments but otherwise LGTM!

agent/structs/connect_ca.go
dnephin: Ah yes, I was referring to this code: https://github.com/hashicorp/consul/blob/v1.11.0-beta3/agent/consul/leader_metrics.go#L83-L98

Dhia added this in 1.11, so we do already use this logic, but it is after we have checked the provider to see if the provider uses an intermediate, so it's a bit safer there.  I need to double check if I need the same here.

agent/consul/leader_connect_test.go
dnephin: Good question! 

For `retry.Run` to actually retry things the failure has to be reported to `r` and not `t`, so we have to be careful to use the right thing, but that may not be your question.

If you mean why are we using `r.Check(rrr)` instead of `require.NoError(t, err)`, I would say there's no real reason. I kept what existed mostly, but they should be equivalent.

I'll change this to use `require.NoError()` so that we can probably deprecate and remove `r.Check()`

dnephin: Oh, but because we've aliased `require := require.New(t)` this no longer works. I've always found `require.New(t)` to be more of a problem than a solution. It makes it way too easy to use the wrong `testing.T` when subtests are involved.

I don't want to do major cleanup to this test in this PR, so I'll keep `r.Check` for now, and we can try to clean some of that up later.

agent/consul/leader_connect_ca.go
dnephin: Ya, I found this `if / else if` to be not super obvious.

All of this logic (from about line 523) is about fixing existing data stored in raft and the state store. So line 531 is where we fix the local reference, but we still need to persist that fix back into raft with `ApplyCARequest`.

So here on line 553 it is correct in the local variable, but we are logging that we are about to (later on below the `if / else if`) persist it.

I would like to refactor this some more. #11687 makes some more changes in this area (for a different reason), and it moves the `activeRoot` call earlier in the function. I think once that is done we can probably refactor this to make it a bit more obvious that the logic is:

```
if nothingToFix {
   return
}

applyFixesToRaft
return
```

agent/structs/connect_ca.go
dnephin: Looking at this again I'm a little uncomfortable exposing this as a method on the struct, so I think Ill move this to be an unexported function near where it is used instead, and improve the godoc to say that the provider should be checked first.

dnephin: Ok, the latest commit cleans this up and makes it safe to use without relying on these assumptions.

agent/consul/leader_connect_test.go
FFMMM: Thanks for the context here!

agent/consul/leader_connect_ca.go
FFMMM: I think that makes sense and that I'm almost there!

So won't the logging actually log two identical values?

agent/consul/leader_connect_test.go
FFMMM: btw, I played around w `require` calls (with a shadowed `require := require.New(t)`) inside a `retry.Run()` call and the require checks worked as expected 🤷🏼

dnephin: Hmm, it shouldn't! If `t.Fatalf()` is called (instead of `r.Fatalf()`) it would stop the test run instead of retrying again. I'm not sure why it appeared to work, maybe it passed the first time instead of failing, so the retry never happened?

agent/consul/leader_connect_ca.go
dnephin: Ah, I see what you are saying now. Yes! I did mess up the logging. I've moved the log up to the if block where we check if it needs update, so it should log the correct thing now.

That allowed me to remove the `else if` and make it just an `if`, which is also nice.

agent/consul/leader_connect_ca.go
```diff
@@ -537,26 +551,21 @@ func (c *CAManager) primaryInitialize(provider ca.Provider, conf *structs.CAConf
...
-
-	} else if activeRoot != nil && !needsSigningKeyUpdate {
+	if activeRoot != nil && !rootUpdateRequired {
 		// This state shouldn't be possible to get into because we update the root and
 		// CA config in the same FSM operation.
 		if activeRoot.ID != rootCA.ID {
 			return fmt.Errorf("stored CA root %q is not the active root (%s)", rootCA.ID, activeRoot.ID)
 		}
 
+		// TODO: why doesn't this c.setCAProvider(provider, activeRoot) ?
```
FFMMM, line 561: +1

FFMMM (approved):
This LGTM 🚀  !

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/513905.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit ff4581092e2f65fab689c9694080788b0f428640 onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/513905)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit ff4581092e2f65fab689c9694080788b0f428640 onto `release/1.9.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/513905)

PR #11672, opened 2021-11-26
ca: set the correct SigningKeyID after config update with Vault provider

### LINKED ISSUE #11662: ca: CARoot.SigningKeyID is wrong after config update when using the Vault CA provider

dnephin:
Fixes #11662
Branched from #11671 

Two other instance of this same bug were fixed a year+ ago in #6513 and  #7012, but this last one remained.

The test failed before the fix, and I extracted a new function for this since the logic is the same in all places, and the method name helps indicate the significance of this intermediate certificate.

We already have logic to fix the `SigningKeyID` on stored `CARoot` when the `CAMananger` is initialized, so we don't need to add more logic for it.

### DIFF
--- a/agent/consul/leader_connect_ca_test.go
+++ b/agent/consul/leader_connect_ca_test.go
@@ -14,6 +14,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/hashicorp/consul/testrpc"
+
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
@@ -24,7 +26,6 @@ import (
 	"github.com/hashicorp/consul/agent/token"
 	"github.com/hashicorp/consul/sdk/testutil"
 	"github.com/hashicorp/consul/sdk/testutil/retry"
-	"github.com/hashicorp/consul/testrpc"
 )
 
 // TODO(kyhavlov): replace with t.Deadline()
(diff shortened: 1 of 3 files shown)


agent/consul/leader_connect_ca_test.go
```diff
@@ -462,3 +464,62 @@ func TestCADelegateWithState_GenerateCASignRequest(t *testing.T) {
...
+
+	_, newRoot, err := s1.fsm.State().CARootActive(nil)
+	require.NoError(t, err)
+	require.Len(t, newRoot.IntermediateCerts, 2,
+		"expected one cross-sign cert and one local leaf sign cert")
+	require.NotEqual(t, origRoot.ID, newRoot.ID)
+
+	cert, err = connect.ParseCert(newRoot.LeafSigningCert())
+	require.NoError(t, err)
+	require.Equal(t, connect.HexString(cert.SubjectKeyId), newRoot.SigningKeyID)
```
dnephin, line 524: This is the line that failed before this fix.

FFMMM (approved):
LGTM 📦 !

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/514121.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 0c7a2257ec7bccf4aee9f279c7589387fbb12026 onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/514121)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 0c7a2257ec7bccf4aee9f279c7589387fbb12026 onto `release/1.9.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/514121)

PR #11673, opened 2021-11-26
ca: clean up unnecessary raft.Apply response checking

dnephin:
In #9991 (d2ab767fef21244e9fe3b9887ea70fc177912381) `raftApply` was changed to handle this check in a single place, instead of requiring every caller check it. It looks like these few places were missed when I did that clean up.

This commit removes the remaining `resp.(error)` checks, since they are all no-ops now.

### DIFF
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -240,12 +240,9 @@ func (c *CAManager) initializeCAConfig() (*structs.CAConfiguration, error) {
 		Op:     structs.CAOpSetConfig,
 		Config: config,
 	}
-	if resp, err := c.delegate.ApplyCARequest(&req); err != nil {
+	if _, err := c.delegate.ApplyCARequest(&req); err != nil {
 		return nil, err
-	} else if respErr, ok := resp.(error); ok {
-		return nil, respErr
 	}
-
 	return config, nil
 }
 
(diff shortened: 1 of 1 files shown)


agent/consul/leader_connect_ca.go
```diff
@@ -564,17 +561,13 @@ func (c *CAManager) primaryInitialize(provider ca.Provider, conf *structs.CAConf
...
 		Index: idx,
 		Roots: []*structs.CARoot{rootCA},
 	})
 	if err != nil {
-		c.logger.Error("Raft apply failed", "error", err)
-		return err
-	}
-	if respErr, ok := resp.(error); ok {
-		return respErr
+		return fmt.Errorf("raft apply failed: %w", err)
```
dnephin, line 570: In general the pattern of "log the error, then return the error" is a problem. Errors should only be logged from where they are handled. The problem here is that we end up with separate log lines, that either makes it look like this event has happened twice, or otherwise repeats information at potentially different log levels. It makes it difficult to associate the "raft apply failed" message with the operation that failed (CA initialization).

I've fixed this in other places, so I'm doing here along with the other cleanup. The caller of this function will handle the error, and the full error with wrapped context is logged where it is handled.

kisunji (approved):

eculver (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/510781.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 33e8c105836778f188ae6816fe68a9a5c8ff1f52 onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/510781)

PR #11677, opened 2021-11-27
sdk: use t.Cleanup in freeport and remove unnecessary calls

dnephin:
This PR adds two new functions to `sdk/freeport` for getting free ports. They both use `t.Cleanup` to return the ports back to the pool. This allows us to remove the "return port" functions in a few places that used `freeport`.

The last commit removes a bunch of callers to `freeport`. As https://github.com/dnephin/freeport-test shows, it is actually safe to use port 0 (`127.0.0.1:0`) along with freeport as long as the listener holds the port the whole time, and doesn't try to close it and re-use it after close. This works because each system uses a different block of ports, freeport explicitly avoids the ephemeral port range.


kisunji (approved):

agent/consul/operator_raft_endpoint_test.go
```diff
@@ -140,13 +140,10 @@ func TestOperator_RaftRemovePeerByAddress(t *testing.T) {
...
 	testrpc.WaitForLeader(t, s1.RPC, "dc1")
 
-	ports := freeport.MustTake(1)
-	defer freeport.Return(ports)
-
 	// Try to remove a peer that's not there.
 	arg := structs.RaftRemovePeerRequest{
 		Datacenter: "dc1",
-		Address:    raft.ServerAddress(fmt.Sprintf("127.0.0.1:%d", ports[0])),
+		Address:    raft.ServerAddress(fmt.Sprintf("127.0.0.1:%d", freeport.Port(t))),
```
rboyer, line 146: I think this one is wrong. This test wants the same port on both new-line-146 and new-line-263 but this is doing two separate calls.

rboyer: never mind

sdk/freeport/freeport.go
```diff
@@ -381,11 +394,39 @@ func logf(severity string, format string, a ...interface{}) {
 	fmt.Fprintf(os.Stderr, "["+severity+"] freeport: "+format+"\n", a...)
 }
 
+type TestingT interface {
```
rboyer, line 397: `sdk/freeport` is a library, so did you want this to be exported and thus part of `sdk/freeport` forever?

dnephin: I think so. This follows the Go convention of "accept interface, return structs". 

Instead of accepting a large interface I think it's common to define a minimal interface so that if someone needs to use this library and doesn't have a `*testing.T` they can easily make a fake implementation.

I think I should probably document this interface better. It should say that new methods could be added to the interface, but generally those methods should always be implemented by `*testing.T`.

sdk/freeport/freeport.go
```diff
@@ -381,11 +394,44 @@ func logf(severity string, format string, a ...interface{}) {
...
+// should always be implemented by *testing.T
+type TestingT interface {
+	Helper()
+	Fatalf(format string, args ...interface{})
+	Cleanup(func())
+}
+
+// GetN returns n free ports from the reserved port block, and returns the
+// ports to the pool when the test ends. See Take for more details.
+func GetN(t TestingT, n int) []int {
```
dhiaayachi, line 410: The naming is becoming a bit confusing, what about `PortN` so we have all the higher level API (the one that take a `TestingT`) named the same?

dnephin: Ya, I'm happy to change the name of this.  I wonder if `PortN` could imply "reserve this TCP port for me". 

We could also change `Port` to `Get1(t *testing.T)` which would be more consistent.

What do you think? I'm happy to keep thinking about more options.

kisunji: I agree that `PortN` might be ambiguous.

I like `GetOne` and `GetN` since they play nicely with the package name:
```
freeport.GetOne
freeport.GetN
```

dnephin:
Sounds good! I've made the change to rename `Port` to `GetOne`. I'll leave this open for another day or so, let me know if there are any outstanding concerns. If I don't hear anything I'll merge it.

rboyer (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/513835.

PR #11679, opened 2021-11-29
ui: Adds basic support for the Routing tab viz with partitions

johncowen:
Partitions means that all of the discovery chain identifiers have an additional segment in them to convey the partition.

This PR adds the necessary code to pick that up so the viz works with a Consul 1.11 build.

Please note this PR does not add any extra user facing functionality, only support for new partition values in the HTTP response.

I also added a very bare docs file for the DiscoChain component while I was there.

### DIFF
--- a/.changelog/11679.txt
+++ b/.changelog/11679.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+ui: Adds support for partitions to the Routing visualization.
+```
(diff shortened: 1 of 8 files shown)


evrowe (approved):
🚀  once changelog is in

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/518027.

PR #11687, opened 2021-11-30
draft: use the CARoots as the source of truth for ClusterID

dnephin:
Relates to points 2 and 3 in #11347

Previously we used the `CAConfiguration` to pass around the "ClusterID". This resulted in:
1. a lot of code having to refer to the config when it already had access to the `CARoot`
2. having to update the stored config when a secondary received a new `CARoot` from the primary

I believe this shows that the "ClusterID" does not really belong as part of the `CAConfiguration`. Even if we wanted to make it settable for boostrapping, it should be clearly marked as a one-time setting , and likely set in a different config field.

Additionally "ClusterID" is quite misleading. It's actually a global ID for all federated clusters.

We were already storing this value as `ExternalTrustDomain` (which is also confusing because the `TrustDomain` field has a value of `<ClusterID>.consul`, but that is cleanup we'll have to leave for later).

This PR uses the existing `ExternalTrustDomain` field to communicate this value. Storing this value on the `CARoot` seems like a better fit. The life cycle of a `CARoot` is a bit closer to the life cycle of this value.

I'm sure there are still test to add, and failures to debug, but opening this PR now for visibility , and to see which tests fail.

### DIFF
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -12,16 +12,15 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-hclog"
-	uuid "github.com/hashicorp/go-uuid"
+	"github.com/hashicorp/go-uuid"
 	"golang.org/x/time/rate"
-
-	"github.com/hashicorp/consul/lib/semaphore"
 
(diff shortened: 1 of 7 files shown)


jmurret:

Issue #9862, opened 2021-03-10
Connect: Non-blocking query to Leaf Cert endpoint can return an expired certificate


JWT95:
#### Overview of the Issue
When using non-blocking queries on the /agent/connect/ca/leaf/:service endpoint, Consul can return an out of date certificate. This directly contradicts the docs at https://www.consul.io/api-docs/agent/connect#service-leaf-certificate
> The resulting certificate is cached and returned by this API until it is near expiry or the root certificates change.

#### Reproduction Steps

2 node cluster with 1 client node and 1 server node in Connect mode. Both agents running Consul 1.9.3.

Set the Connect leaf_cert_ttl to be 1h.

Get a cert for service "leaf-cert" on the client node

```
$ date && curl http://127.0.0.1:8500/v1/agent/connect/ca/leaf/leaf-cert | jq .
Wed Mar 10 12:31:30 UTC 2021
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1384  100  1384    0     0  38444      0 --:--:-- --:--:-- --:--:-- 38444
{
  "SerialNumber": "08",
  "CertPEM": "...",
  "PrivateKeyPEM": "...",
  "Service": "leaf-cert",
  "ServiceURI": "spiffe://a22a7789-e9fd-1720-9494-9994cad63968.consul/ns/default/dc/dc1/svc/leaf-cert",
  "ValidAfter": "2021-03-10T12:30:30Z",
  "ValidBefore": "2021-03-10T13:30:30Z",
  "CreateIndex": 21,
  "ModifyIndex": 21
}
```

Wait an hour, hit the same endpoint again. The same, now out of date, certificate is returned (look at the ValidBefore field)

```
$ date && curl http://127.0.0.1:8500/v1/agent/connect/ca/leaf/leaf-cert | jq .
Wed Mar 10 13:33:55 UTC 2021
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1384  100  1384    0     0  1351k      0 --:--:-- --:--:-- --:--:-- 1351k
{
  "SerialNumber": "08",
  "CertPEM": "...",
  "PrivateKeyPEM": "...",
  "Service": "leaf-cert",
[...]

PR #11693, opened 2021-11-30
add MustRevalidate flag to connect_ca_leaf cache type; always use on non-blocking queries

FFMMM:
### Overview

PR addresses https://github.com/hashicorp/consul/issues/10871 / https://github.com/hashicorp/consul/issues/9862 . 

Today, once we HIT/ load up a leaf cert in the agent cache, we can return expired certs or certs signed by a previously rotated CA.

These changes introduces a `MustRevalidate` flag to the `connect_ca_leaf` cache type that we always use on non-blocking queries.

### Notes

Always setting `MustRevalidate` on non blocking queries means we always "MISS" the cache on non blocking queries. In effect, the headers are now redundant. 

### PR checklist
* [x] `gofmt`
* [x] testing 
* [x] pr changelog
* ~[ ] docs update~ --> I want to pull this into a separate PR bc I want to change some of the docs wording

### DIFF
--- a/agent/agent_endpoint.go
+++ b/agent/agent_endpoint.go
@@ -1494,7 +1494,9 @@ func (s *HTTPHandlers) AgentConnectCARoots(resp http.ResponseWriter, req *http.R
 }
 
 // AgentConnectCALeafCert returns the certificate bundle for a service
-// instance. This supports blocking queries to update the returned bundle.
+// instance. This endpoint ignores all "Cache-Control" attributes.
+// This supports blocking queries to update the returned bundle.
+// Non-blocking queries will always verify that the cache entry is still valid.
 func (s *HTTPHandlers) AgentConnectCALeafCert(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
 	// Get the service name. Note that this is the name of the service,
 	// not the ID of the service instance.
@@ -1517,6 +1519,13 @@ func (s *HTTPHandlers) AgentConnectCALeafCert(resp http.ResponseWriter, req *htt
 	args.MinQueryIndex = qOpts.MinQueryIndex
 	args.MaxQueryTime = qOpts.MaxQueryTime
 	args.Token = qOpts.Token
+
+	// We don't want non-blocking queries to return expired leaf certs
+	// or leaf certs not valid under the current CA. So always revalidate
+	// the leaf cert on non-blocking queries (ie when MinQueryIndex == 0)
+	if args.MinQueryIndex == 0 {
+		args.MustRevalidate = true
+	}
 
 	if !s.validateRequestPartition(resp, &args.EnterpriseMeta) {
 		return nil, nil
--- a/agent/agent_endpoint.go
+++ b/agent/agent_endpoint.go
@@ -1520,6 +1520,7 @@ func (s *HTTPHandlers) AgentConnectCALeafCert(resp http.ResponseWriter, req *htt
 	args.MaxQueryTime = qOpts.MaxQueryTime
 	args.Token = qOpts.Token
 
+	// TODO(ffmmmm): maybe set MustRevalidate in ConnectCALeafRequest (as part of CacheInfo())
 	// We don't want non-blocking queries to return expired leaf certs
 	// or leaf certs not valid under the current CA. So always revalidate
 	// the leaf cert on non-blocking queries (ie when MinQueryIndex == 0)
--- a/agent/agent_endpoint_test.go
+++ b/agent/agent_endpoint_test.go
@@ -5837,11 +5837,8 @@ func TestAgentConnectCALeafCert_good(t *testing.T) {
 
 		// Test caching for the leaf cert
 		{
-			fetched := 0
-			for {
-				if fetched == 4 {
-					break
-				}
+			
+			for fetched := 0; fetched < 4; fetches++ {
 
 				// Fetch it again
 				resp := httptest.NewRecorder()
--- a/agent/agent_endpoint_test.go
+++ b/agent/agent_endpoint_test.go
@@ -5779,16 +5779,13 @@ func TestAgentConnectCALeafCert_good(t *testing.T) {
 		obj2, err := a.srv.AgentConnectCALeafCert(resp, req)
 		require.NoError(err)
 		require.Equal(obj, obj2)
-
-		// Should cache hit this time and not make request
-		require.Equal("HIT", resp.Header().Get("X-Cache"))
 	}
+
+	// Set a new CA
+	ca2 := connect.TestCAConfigSet(t, a, nil)
 
 	// Issue a blocking query to ensure that the cert gets updated appropriately
 	{
-		// Set a new CA
-		ca := connect.TestCAConfigSet(t, a, nil)
-
 		resp := httptest.NewRecorder()
 		req, _ := http.NewRequest("GET", "/v1/agent/connect/ca/leaf/test?index="+index, nil)
 		obj, err := a.srv.AgentConnectCALeafCert(resp, req)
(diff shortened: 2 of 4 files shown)


agent/agent_endpoint.go
```diff
@@ -1518,6 +1520,13 @@ func (s *HTTPHandlers) AgentConnectCALeafCert(resp http.ResponseWriter, req *htt
 	args.MaxQueryTime = qOpts.MaxQueryTime
 	args.Token = qOpts.Token
 
+	// We don't want non-blocking queries to return expired leaf certs
+	// or leaf certs not valid under the current CA. So always revalidate
+	// the leaf cert on non-blocking queries
+	if args.MinQueryIndex == 0 {
+		args.MustRevalidate = true
+	}
```
dnephin, line 1528: I think this logic might be better in `ConnectCALeafRequest.CacheInfo()`. I believe that method should have access to all the same data.

By moving it there we can be sure that all users of the cache do the right thing, instead of only this one HTTP API endpoint.

agent/cache-types/connect_ca_leaf.go
```diff
@@ -643,14 +659,15 @@ func (c *ConnectCALeaf) generateNewLeaf(req *ConnectCALeafRequest,
...
-	MaxQueryTime  time.Duration
+	Token          string
+	Datacenter     string
+	Service        string // Service name, not ID
+	Agent          string // Agent name, not ID
+	DNSSAN         []string
+	IPSAN          []net.IP
+	MinQueryIndex  uint64
+	MaxQueryTime   time.Duration
+	MustRevalidate bool
```
dnephin, line 670: I think if we move the logic into `CacheInfo` (suggested in the other comment) then we may be able to avoid adding this field.

agent/cache-types/connect_ca_leaf.go
```diff
@@ -401,10 +401,20 @@ func (c *ConnectCALeaf) Fetch(opts cache.FetchOptions, req cache.Request) (cache
...
 
+	// If we called Fetch() with MustRevalidate then this call came from a non-blocking query.
+	// To pickup any CA rotations that may have happened prior to this call, we will reduce the timeout
+	// to ~ 1 sec. Otherwise, we risk non-blocking queries taking ~ 10 minutes to wait on a CA rotation
+	// that could never happen.
+	// see https://github.com/hashicorp/consul/issues/10871, https://github.com/hashicorp/consul/issues/9862
+	// This is a hack, so, ideally, we can revisit the caching/ fetching logic in this case
+	if req.CacheInfo().MustRevalidate {
+		timeoutCh = time.After(time.Second)
+	}
```
dnephin, line 416: I think this works, but I also wonder if we could make the non-blocking case a bit more obvious by avoiding the select entirely.

What if we were to:
* move this `if` above where we create `rootUpdateCh` around line 382
* remove the second `if` below (we'll handle it in this same if block)
* instad of `timeoutCh = time.After(time.Second)` we do something like this:

```go
roots, err := c.rootsFromCache()
if err != nil {
    return lastResultWithNewState(), err
}
if activeRootHasKey(roots, state.authorityKeyID) {
        return lastResultWithNewState(), err
}
... // some shared logic here to calculate expiresAt
if state.forceExpireAfter.Before(expiresAt) {
    return c.generateNewLeaf(reqReal, lastResultWithNewState())
}
return lastResultWithNewState(), err
```

The `...` about calculating `expiresAt` could be extracted from the 461-470 into a new function and called from both places.

I haven't fully gone through that to be sure that works, but my impression is that something like that would work and allow us to avoid the `select` which makes it a bit more obvious that we are handling  a non-blocking request, without having to set a short timeout.

dnephin (commented):
Nice work! Some thoughts/suggestions

agent/cache-types/connect_ca_leaf.go
FFMMM: Hey thanks for bringing this up again -- I recall we chatted about this approach and dropped the ball as I was debugging the "vault issue".  💯  

I think it makes a lot of sense (readability, maintainability) to treat the `MustRevalidate`/ non-blocking case as "separate" from the select stmt. 

Immediately after we check : https://github.com/hashicorp/consul/blob/ed8a901be72a45d28b70ea357c44e0b009a83cb0/agent/cache-types/connect_ca_leaf.go#L378-L381

If we add something like ~:

```
	if req.CacheInfo().MustRevalidate {
		roots, err := c.rootsFromCache()
		if err != nil {
			return lastResultWithNewState(), err
		}
		if activeRootHasKey(roots, state.authorityKeyID) {
			return lastResultWithNewState(), nil
		}

		// if we reach here then the current leaf was not signed by the same CAs, just regen
		return c.generateNewLeaf(reqReal, lastResultWithNewState())
	}
```

That will catch most occurrences of https://github.com/hashicorp/consul/issues/10871. Sometimes, the CA may take a bit to be rotated but that's ok. "Certainty" / " strong consistency" is what the blocking queries are for.

Also note, that we don't need to "calculate" when to regen the cert. If for whatever reason we don't have a CA that signed the leaf, we need to regen the leaf cert.

Wdyt?

agent/agent_endpoint.go
FFMMM: I think I'm going to walk this back.

Here's what I ran into -- there's a non trivial amount of ConnectCALeaf testing that would need to change as a result of always passing in `MustRevalidate` assignment logic in the CacheInfo(). It kinda feels like this behavior is "breaking" with previous one. It seems, ergonomically, this func was designed more as a "getter".

agent/cache-types/connect_ca_leaf.go
FFMMM: addressed walking back this change here --> https://github.com/hashicorp/consul/pull/11693#discussion_r760408983

agent/agent_endpoint.go
dnephin: hmm, that is unfortunate. I think those tests must not be accurately testing the production behaviour very well, but that is cleanup we can leave for another time I think. Thanks for trying.

Maybe leave a TODO here instead? Not critical

agent/agent_endpoint_test.go
```diff
@@ -5798,12 +5795,64 @@ func TestAgentConnectCALeafCert_good(t *testing.T) {
...
+			require.Equal("MISS", resp.Header().Get("X-Cache"))
+		}
+
+		// Test caching for the leaf cert
+		{
+			fetched := 0
+			for {
+				if fetched == 4 {
+					break
+				}
```
dnephin, line 5844: I think this is equivalent ?

```suggestion
			
			for fetched := 0; fetched < 4; fetches++ {
```

And would allow you to remove the `fetched++` below as well?

dnephin (approved):
Nice, LGTM!

One small suggestion for the loop in the test. I think it might be worth to look into moving the `MustRevalidate` into the CacheInfo stuff, but that is not urgent, so would be fine as a TODO I think.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/513605.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 384d497f267aefbcd173e7b1fd7fca9f762a82fe onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/513605)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 384d497f267aefbcd173e7b1fd7fca9f762a82fe onto `release/1.9.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/513605)

PR #11702, opened 2021-12-01
ui: Adds basic support for partition exports to Service listings

johncowen:
This PR adds a `partition / namespace` icon/badge type thing to the rows of our Service lists.

Screenshot 2021-12-02 at 12 34 58

This element only appears if the Service's partition is different to your current partition (and has therefore been exported from another partition)

For testing purposes only you have to be in a non-default partition for it to appear (just to underline this is just a test thing in production this doesn't apply)

More info in the docs (also added here)


evrowe (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/518153.

PR #11713, opened 2021-12-02
ca: make test naming consistent

dnephin:
While working on the CA system it is important to be able to run all the tests related to the system, without having to wait for unrelated tests. There are many slow and unrelated tests in agent/consul, so we need some way to filter to only the relevant tests.

This PR renames all the CA system related tests to start with either:
*  `TestCAMananger` for tests of internal operations that don't have RPC endpoint, or
* `TestConnectCA` for tests of RPC endpoints. 

This allows us to run all the test with:

    go test -run 'TestCAMananger|TestConnectCA' ./agent/consul

The test naming follows an undocumented convention of naming tests as follows:

    Test[<struct name>_]<function name>[_<test case description>]

I tried to always keep Primary/Secondary at the end of the description, and `_Vault_` has to be in the middle because the regex used to run those tests as a separate CI job ([ref](https://github.com/hashicorp/consul/blob/main/GNUmakefile#L328-L333)).

You may notice some of the test names changed quite a bit. I did my best to identify the underlying method being tested, but I may have been slightly off in some cases.

I think this test naming convention is also very important so that we know which tests exist (and which ones are missing). Sometimes it is better to fix or add assertions to an existing test, than to write a new one. And when making a code changes, it's important to be able to quickly see which cases are already well tested, and which are not.

I'm planning on backporting this to reduce conflicts as we make other changes.

I also renamed `InitializeCA` to `Initialize` since `CA` is already in the struct type name.

### DIFF
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -311,7 +311,7 @@ func (c *CAManager) Start(ctx context.Context) {
 	// Attempt to initialize the Connect CA now. This will
 	// happen during leader establishment and it would be great
 	// if the CA was ready to go once that process was finished.
-	if err := c.InitializeCA(); err != nil {
+	if err := c.Initialize(); err != nil {
 		c.logger.Error("Failed to initialize Connect CA", "error", err)
 
 		// we failed to fully initialize the CA so we need to spawn a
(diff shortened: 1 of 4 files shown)


boxofrad (approved):
LOVE IT 😍

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/513997.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 96f95889db674531722df2c0325b7651b1c1dc03 onto `release/1.9.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/513997)

PR #11720, opened 2021-12-02
Use bbolt instead of the legacy bolt implementation

mkeeler:
In addition to just swapping out bbolt for bolt this also exposes a configuration to allow for disabling boltdb freelist syncing.
Finally it also enables metrics emission for the raft boltdb store to give us insights into boltdb performance.

### DIFF
--- a/.changelog/11720.txt
+++ b/.changelog/11720.txt
@@ -1,0 +1,11 @@
+```release-note:improvement
+raft: Use bbolt instead of the legacy boltdb implementation
+```
+
+```release-note:improvement
+raft: Emit boltdb related performance metrics
+```
+
+```release-note:improvement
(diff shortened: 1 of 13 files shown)


freddygv (approved):
LGTM

dnephin (approved):
I guess we'll want to document these new metrics, but not required for this PR

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/517055.

david-yu:
Btw I just noticed @mkeeler the latest version of bbolt is actually 1.3.6. It was discussed here: https://github.com/etcd-io/bbolt/issues/278, but it was released as a tag on the GitHub repo, but not actually released via GitHub releases.

Issue #1383, opened 2015-11-05
acl: how to deny deregister


anklos:
I use docker image `gliderlabs/docker-consul` to set up a single consul and have the consul.json config like below:

``` json
{
  "datacenter": "redbook1",
  "acl_master_token": "redbook",
  "acl_datacenter": "redbook1",
  "acl_default_policy": "deny",
  "data_dir": "/data",
  "ui_dir": "/ui",
  "client_addr": "0.0.0.0",
  "ports": {
    "dns": 53
  },
  "recursor": "8.8.8.8",
  "disable_update_check": true
}
```

So most things work as expected (i.e cannot register service without token, cannot update key/value without proper rules), except:
1. I can still deregister node without token on UI.
2. I can still see all the nodes and services without token on UI.

Is it possible to deny those requests from ACL? Thanks!

highlyunavailable:
As per the [ACL System](http://www.consul.io/docs/internals/acl.html) documentation, ACLs are currently only supported for the K/V store and Services.

So no, at this time, it's not possible to do what you want. I believe the Hashicorp guys have stated it's on the roadmap for the future but it's not available right now.

slackpad:
Service discovery ACLs were added in https://github.com/hashicorp/consul/pull/1024 and are part of the upcoming Consul 0.6. Deregister isn't currently covered so I'll mark this as an enhancement.

anklos:
Thanks!

cleung2010:
How come the UI on the demo (http://nyc3.demo.consul.io/ui/#/nyc3/nodes/nyc3-server-1) not have the "Deregister" button? Is it because it's running an older version of Consul/UI before #364 was implemented?

slackpad:
Hi Calvin - yes we are in the process of updating it. The demo has some
extra branding for Digital Ocean so we need to update a fork with those
changes and the latest ui code.

On Wed, Dec 2, 2015 at 8:37 AM, Calvin Leung Huang <notifications@github.com

zeldal:
Is there any1 working for this issue? Since 0.6 is out and it still doesnt control deregister request. we need this before production deployment, i will make fix for our deployments and pull here.

slackpad:
Hi @zeldal this didn't make it into 0.6.1 but we will try to get it into the next release.

zeldal:
We are gonna make some protections with different tools to avoid this for now. I will be eager to upgrade newer version when this is fixed. 
Best regards.

[...]

PR #11726, opened 2021-12-03
Cross port of ent #1383

markan:
Cross port of ent #1383 "Reject non-default datacenter when making partitioned ACLs"

On the OSS side this is a minor refactor to add some more checks that are no-ops on oss, but add function to the enterprise code.


hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/516184.

PR #11738, opened 2021-12-04
Transparent proxy support for cross-partition

freddygv:
There are a few changes here:
- Transparent proxy listener configuration needs to account for the new tagged address with virtual IPs generated by Consul. These should be considered both for cross-partition and partition-local traffic. The plain "virtual" tagged address should only be used for partition-local traffic, since the address is not guaranteed to be unique across partitions.
- The existing way of computing upstreams/downstreams from intentions was based on watching the "services" table and then evaluating a service's intentions against the unique service names. This won't work in 1.11 because the services table does not have a wildcard partition index, and cross-partition upstreams from intentions are supported. This watch was replaced with a watch of a table that stores service names by kind across all partitions.

### DIFF
--- a/agent/xds/listeners.go
+++ b/agent/xds/listeners.go
@@ -168,13 +168,22 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
 		// We do not match on all endpoints here since it would lead to load balancing across
 		// all instances when any instance address is dialed.
 		for _, e := range endpoints {
-			if vip := e.Service.TaggedAddresses[virtualIPTag]; vip.Address != "" {
+			if vip := e.Service.TaggedAddresses[structs.TaggedAddressVirtualIP]; vip.Address != "" {
 				uniqueAddrs[vip.Address] = struct{}{}
 			}
+
+			// The virtualIPTag is used by consul-k8s to store the ClusterIP for a service.
+			// We only match on this virtual IP if the upstream is in the proxy's partition.
+			// This is because the IP is not guaranteed to be unique across k8s clusters.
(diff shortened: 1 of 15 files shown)


agent/xds/listeners.go
```diff
@@ -168,13 +168,22 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
...
+			if structs.EqualPartitions(e.Node.PartitionOrDefault(), cfgSnap.ProxyID.PartitionOrDefault()) {
+				if vip := e.Service.TaggedAddresses[virtualIPTag]; vip.Address != "" {
+					uniqueAddrs[vip.Address] = struct{}{}
+				}
+			}
 		}
-		if len(uniqueAddrs) > 1 {
-			s.Logger.Warn("detected multiple virtual IPs for an upstream, all will be used to match traffic",
-				"upstream", id)
+		if len(uniqueAddrs) > 2 {
```
markan, line 184: Why is this now > 2 ? Is this because there's a virtual IP and a tagged virtual IP?

kisunji (approved):

markan (approved):
This looks pretty reasonable to me. One question, but shouldn't block merge.

agent/xds/listeners.go
freddygv: Yep one is populated by consul-k8s and one by Consul, so if we detect more than two it's unexpected

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/517827.

Issue #9422, opened 2020-12-17
Consul  Catalog Api support to query by tags Or pageing query


wangyushuai:
#### Feature Description

When using Catalog Api (**/v1/catalog/services**) to query the service list, I want to filter the list by Tag, because the total number of lists is very large, and I hope to support paging query；

#### Use Case(s)

curl /v1/catalog/services?filter='tag==[app=dashboard]'

nugend:
Would love to have this. Also, just being able to query _all_ service health statuses by tag without first specifying a service id would be really nice too.

dekimsey:
I'm upvoting this because we are running into issues with Prometheus pulling the whole service catalog and trying to set watches on each service so it can perform additional checks.

Right now, much of the data I need to make decisions on which services to monitor is in the `ServiceMeta`. However, the `/v1/catalog/services` API does not expose the filtering capabilities to par down the response. So my applications are having to enumerate each service individually in the catalog, which is expensive.

Since filter appears to have replace the NodeMeta/ServiceMeta parameters in other parts of the API, it'd be nice to have support here as well.

I spent some time digging into Consul's source to see how this might pan out. It's more involved than a change I'm comfortable with proposing at the moment. My first pass involved an update to the api, the rpc, and the store lookup because the current implementation takes the assumption (`{"service1: ["tags", ...] }` response) all the way down to the data-store lookup. I don't know if a change that deep would be accepted by the project, so I kinda stopped there. The end-result of the change would mean the RPC calls are somewhat larger since they'd have the full ServiceNode data response instead of the stripped down tags only response.

Thanks!

Note: There is some related discussion on making the catalog lookups more effective in prometheus, see prometheus/prometheus#7626. I believe adding filtering support on the catalog api would help in this effort.

dnephin:
@dekimsey Thank you for your interest in this issue! I was curious about why this one endpoint did not have filtering (most endpoints do support the `filter` parameter). I see what you are saying about this response data being in a different format. I guess that's why filtering was not added initially.

I did a quick pass to see what might be required to make this work. You can see those changes here: https://github.com/hashicorp/consul/compare/dnephin/catalog-service-list-filter. As it says in the commit message this is missing test changes and docs. I haven't testing it at all, but I would expect something like this to work. 

[...]

PR #11742, opened 2021-12-05
Add support for filtering the 'List Services' API

dekimsey:
1. Create a bexpr filter for performing the filtering
2. Change the state store functions to return the raw (not aggregated)
   list of ServiceNodes.
3. Move the aggregate service tags by name logic out of the state store
   functions into a new function called from the RPC endpoint
4. Perform the filtering in the endpoint before aggregation.

TODO:
* [x] update existing tests for the new function signatures
* [x] add new tests for filtering the results of this endpoint
* [x] update the api-docs

This change updates the `/v1/catalog/services` endpoint to now support the `filter` parameter. This change is desirable as it brings the API into alignment with other endpoints and more importantly makes it easier for a client to query for only part of the catalog based on more than just node metadata.

This is potentially of particular use to projects such as Prometheus, where scraping the entire catalog and then querying each service individually for their respective metadata is expensive. By making some of Consul's advanced filtering available earlier in the process, this can improve performance of target discovery and reduce resource utilization.

Please note the heavy lifting of updating the RPC/State was done by @dnephin! So, give props to him :)

I've added some of my own comments in-line. Thank you

This PR targets 1.9 because that's what I currently have deployed so that I can test it in my infrastructure. I will rebase accordingly onto main. But I wanted to get some early feedback first.

Addresses #9422


website/content/api-docs/catalog.mdx
```diff
@@ -408,6 +408,47 @@ The table below shows this endpoint's support for
   `X-Consul-Namespace` header. If not provided, the namespace will be inherited
   from the request's ACL token or will default to the `default` namespace. Added in Consul 1.7.0.
```
dekimsey, line 410: Should the `node_meta` argument be marked as deprecated?

agent/consul/state/catalog_test.go
```diff
@@ -1674,31 +1663,33 @@ func TestStateStore_ServicesByNodeMeta(t *testing.T) {
...
 
 	t.Run("Get an empty list for an invalid meta value", func(t *testing.T) {
 		_, res, err := s.ServicesByNodeMeta(ws, map[string]string{"invalid": "nope"}, nil)
 		if err != nil {
 			t.Fatalf("err: %s", err)
 		}
-		expected := structs.Services{}
+		// FIXME: Our changes now return nil instead of empty list, do we care about this change?
+		// (Only the test here caught the difference)
+		var expected []*structs.ServiceNode
```
dekimsey, line 1699: Just highlighting the FIXME here.

kisunji:
Hi @dekimsey, thanks for this PR!

I'd like to request that your PR targets `main` - we generally do not add features to older releases (unless they are bugfixes or security-related) but even in those cases it is easier to backport from `main` to the appropriate `release/` branches.

dekimsey:
Not a problem @kisunji, I updated the PR and rebased it accordingly.

kisunji:
Thanks @dekimsey! 

Our team is busy with a some releases this week but will aim to have this reviewed early next week!

dekimsey:
@kisunji thank you, no rush! I'm looking forward to the feedback.

agent/consul/state/catalog_test.go
```diff
@@ -1950,31 +1939,40 @@ func TestStateStore_ServicesByNodeMeta(t *testing.T) {
 		if err != nil {
 			t.Fatalf("err: %s", err)
 		}
-		expected := structs.Services{
-			"redis": []string{"primary", "prod"},
-		}
-		sort.Strings(res["redis"])
-		require.Equal(t, expected, res)
+		require.Len(t, res, 1)
+		require.True(t, res[0].ServiceID == "service1" && res[0].Node == "node0")
```
dnephin, line 1943: Something like this might be a bit more obvious, and give better error messages on failures. Does this work?

```suggestion
		require.Equal(t, res, []*structs.NodeServbice{ns1})
```

agent/consul/state/catalog_test.go
```diff
@@ -1950,31 +1939,40 @@ func TestStateStore_ServicesByNodeMeta(t *testing.T) {
...
+		for _, sn := range res {
+			require.Condition(t, func() bool {
+				for _, exp := range expected {
+					if exp.ServiceID == sn.ServiceID && exp.Node == sn.Node {
+						return true
+					}
+				}
+				return false
+			})
 		}
```
dnephin, line 1965: Similarly here, would this work?
```suggestion
		require.Equal(t, res, []*structs.NodeServbice{ns1. ns2})
```

agent/consul/state/catalog_test.go
dnephin: Good question! This should be ok. A `nil` slice is pretty common and can be passed to `append` , `len`, or `range` without trouble.  It's only a concern for when we serialize a response, and in this case we know we are going to process the slice first, so the RPC response will use a different value.

website/content/api-docs/catalog.mdx
dnephin: Ya, I think this would be a good time to deprecate `node_meta`.

dnephin (commented):
Thank you for working on this! It looks great. Mostly some minor suggestions for the tests, and responses to your questions.

agent/consul/state/catalog_test.go
dekimsey: Hrm, I'm not sure how to make this work. I like it, it's far cleaner. However, here we are receiving `[]*struct.ServiceNode`  but ns1, ns2 are defined as `structs.NodeService`.

(To be honest, the difference confuses me)

agent/consul/state/catalog_test.go
dekimsey: Same as before, this is the wrong struct type and I'm not sure how to handle that.

dekimsey (commented):
Thank you for the feedback @dnephin, in implementing it however I ran into an issue with the ServiceNode vs NodeService in one of the tests. Everything else went smoothly.

agent/consul/state/catalog_test.go
dnephin: Ah, oops, I missed they were different types. Ya I guess `ns1` and `ns2` won't work. Maybe something like this?

```go
 []*structs.NodeService{ns1.ToServiceNode("node0")}
```

The names do make it difficult to see the difference. One type is persisted and used for some endpoints, and the other type is only used for API/RPC endpoints.

dekimsey: Sorry about the delay in getting back to you on this. I tried this solution but reflect.Equals is tripping on the structs' RaftIndex values changing. I'm not sure if you would want to pull in something like go-cmp/gotestyourself or create a custom comparison function.
I've committed the recommended changes, I like the approach. Just not sure what the best way to tackle it in keeping with the project.

kisunji:
Would you be able to rebase from main? Looks like there's been several refactors since this PR was up. Let me know if anything gets complicated.

No worries about taking long to respond! We're thankful for your contributions regardless.

dekimsey:
@kisunji The tests were updated and cleaned up which I think addresses @dnephin's review. Please let me know what you think! Thanks

kisunji:
Thanks @dekimsey! I'm going to sanity check that this won't break any files in Enterprise then will try to get another set of eyes for review.

website/content/api-docs/catalog.mdx
```diff
@@ -410,13 +410,64 @@ The corresponding CLI command is [`consul catalog services`](/commands/catalog/s
...
 
 - `ns` `(string: "")` <EnterpriseAlert inline /> - Specifies the namespace of the services you lookup.
   You can also [specify the namespace through other methods](#methods-to-specify-namespace).
 
+- `filter` `(string: "")` - Specifies the expression used to filter the
+  queries results prior to returning the data.
+
+### Filtering
+
+The filter will be executed against each Service mapping within the catalog.
```
trujillo-adam, line 427:
```suggestion
The filter will be executed against each service mapping within the catalog.
```

trujillo-adam (commented):
In terms of docs, I made one small suggestion to address style.

trujillo-adam (approved):

dekimsey:
Hi all, I just wanted to check-in on the status of this. I've been trying to keep it up-to-date with main as it's moved a few times since I initially opened the MR. Our prometheus instances open a _ton_ of watchers because we don't currently have a good way to prune the target services on the List Services API call, this change would allow us to address that gap. Please let me know what I need to do or change to get this integrated.

Thank you!

kisunji:
Hi @dekimsey,

Sorry for the delay in getting back to you. There are some tests on the enterprise side that need updating so I will work on it and have this merged for our next minor release (1.14)

kisunji (approved):

PR #11747, opened 2021-12-06
ui: Amends to Routing visualization for partitions

johncowen:
There is one user facing fix here (something I missed in https://github.com/hashicorp/consul/pull/11679). Plus a bunch of DX/fixture amends for when we manually check the functionality of the discovery chain component.

I decided it would be best to just PR this altogether for ease, but will note the bugfix inline.

~P.S. Looking at the failing tests now 😅~

~Moving to Draft til I get more info here~

Turns out the reason why I thought it was a bug is because that bit wasn't quite finished up yet, that PR finishing that up is now in so all makes sense.

### DIFF
--- a/ui/packages/consul-ui/app/components/consul/discovery-chain/resolver-card/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/discovery-chain/resolver-card/index.hbs
@@ -38,9 +38,9 @@
     {{#if child.Redirect}}
           <dl class="redirect">
             <dt
-              {{tooltip "Redirect"}}
+              {{tooltip (concat child.Redirect ' redirect')}}
             >
-              Redirect
+              {{child.Redirect}} redirect
             </dt>
             <dd>
               {{child.Name}}
@@ -51,7 +51,7 @@
               <dt
                 {{tooltip (concat child.Failover.Type ' failover')}}
               >
-                {{concat child.Failover.Type ' failover'}}
+                {{child.Failover.Type}} failover
               </dt>
               <dd>
                 <ol>
--- a/ui/packages/consul-ui/app/components/consul/discovery-chain/utils.js
+++ b/ui/packages/consul-ui/app/components/consul/discovery-chain/utils.js
@@ -43,6 +43,7 @@ export const getSplitters = function(nodes) {
     const temp = item.Name.split('.');
     temp.reverse();
     temp.shift();
+    temp.shift();
     temp.reverse();
     return {
       ...item,
@@ -75,17 +76,17 @@ export const getResolvers = function(
     .forEach(function(item) {
       const parts = item.Name.split('.');
       let subset;
-      // this will leave behind the service.name.nspace.dc even if the service name contains a dot
+      // this will leave behind the service.name.nspace.partition.dc even if the service name contains a dot
       if (parts.length > 4) {
         subset = parts.shift();
       }
       parts.reverse();
-      // slice off from dc.nspace onwards leaving the potentially dot containing service name
+      // slice off from dc.partition.nspace onwards leaving the potentially dot containing service name
       // const nodeDc =
       parts.shift();
-      // const nodeNspace =
-      parts.shift();
       // const nodePartition =
+      parts.shift();
+      // const nodeNspace =
       parts.shift();
       // if it does contain a dot put it back to the correct order
       parts.reverse();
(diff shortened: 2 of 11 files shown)


ui/packages/consul-ui/app/components/consul/discovery-chain/resolver-card/index.hbs
```diff
@@ -51,7 +51,7 @@
               <dt
                 {{tooltip (concat child.Failover.Type ' failover')}}
               >
-                {{concat child.Failover.Type ' failover'}}
+                {{child.Failover.Type}} failover
```
johncowen, line 54: All of this stuff just makes things consistent, no reason why redirect was capitalized and failover wasn't - also now we have DC and/or Partition redirects I've added the type of redirect to the tooltip.

ui/packages/consul-ui/app/components/consul/discovery-chain/utils.js
```diff
@@ -43,6 +43,7 @@ export const getSplitters = function(nodes) {
     const temp = item.Name.split('.');
     temp.reverse();
     temp.shift();
+    temp.shift();
```
johncowen, line 46: This line was the "bugfix" previous to this splitters were showing `service-name.nspace` instead of just `service-name`. I think theoretically splitters can actually be in a different namespace, so longer term we should display some sort of namespace here, but desired feature at the minute is to just show the `service-name`.

ui/packages/consul-ui/app/components/consul/discovery-chain/utils.js
```diff
@@ -75,18 +76,18 @@ export const getResolvers = function(
...
-      // slice off from dc.nspace onwards leaving the potentially dot containing service name
+      // slice off from dc.partition.nspace onwards leaving the potentially dot containing service name
       // const nodeDc =
       parts.shift();
-      // const nodeNspace =
-      parts.shift();
       // const nodePartition =
       parts.shift();
+      // const nodeNspace =
+      parts.shift();
```
johncowen, line 90: Everything above here is just niceties, add partition into the comment etc The comments to explain/illustrate/document the shifting was actually out of order, but we don't use the resulting variables anyway so it didn't matter. Longer term we could probably use more destructuring here instead (this code was added when the codebase was in ES5 mode)

ui/packages/consul-ui/app/components/consul/discovery-chain/utils.js
```diff
@@ -124,7 +125,7 @@ export const getResolvers = function(
         const resolver = findResolver(resolvers, target.Service, nspace, partition, dc);
         // and add the redirect as a child, redirects are always children
         const child = {
-          Redirect: true,
+          Redirect: alternate.Type,
```
johncowen, line 128: Instead of just knowing if it's a "redirect" or not, we also need to know the type. Instead of adding further properties, we just rely on that fact that if the "redirect" type isn't set, then its not a "redirect". There's a larger question to be answered on what a redirect actually is in the UI, but thats another longer term thing.

ui/packages/consul-ui/mock-api/v1/discovery-chain/virtual-admin-6
```diff
@@ -0,0 +1,196 @@
...
+        "Namespace": "default",
+        "Datacenter": "dc-gcp",
+        "MeshGateway": {},
+        "Subset": {},
+        "SNI": "db.default.default.dc-gcp.internal.ede9daf4-b641-3954-5a52-15f4fd666f6e.consul",
+        "Name": "db.default.default.dc-gcp.internal.ede9daf4-b641-3954-5a52-15f4fd666f6e.consul"
+      }
+    }
+  }
+}
```
johncowen, line 196: These are all fixtures from a real Consul install/usecase, well filename numbers 1-5 are, number 6 I added with a partition usecase.

For filenames 1 - 5 I prettified and added the relevant new partition fields, for number 6 I copied number 1 and then added the extra partition usecase to try with.

johncowen (commented):
Notes:

johncowen:
Seems like this is relevant https://github.com/hashicorp/consul/pull/11757 here

ui/packages/consul-ui/app/components/consul/discovery-chain/utils.js
johncowen: So it turns out this wasn't a bugfix, it was an unfinished feature, and I added the feature to our mocks without realising it was not yet done in the backend. So until #11757 went in, this was essentially _introducing_ a bug.

kaxcode (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/522182.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit c410d295be7d39c655b8bd97d8143dea5336a619 onto `release/1.11.x` succeeded!

PR #11748, opened 2021-12-06
areas: make the gRPC server tracker network area aware

rboyer:
Fixes a bug whereby servers present in multiple network areas would be
properly segmented in the Router, but not in the gRPC mirror. This would
lead servers in the current datacenter leaving from a network area
(possibly during the network area's removal) from deleting their own
records that still exist in the standard WAN area.

The gRPC client stack uses the gRPC server tracker to execute all RPCs,
even those targeting members of the current datacenter (which is unlike
the net/rpc stack which has a bypass mechanism).

This would manifest as a gRPC method call never opening a socket because
it would block forever waiting for the current datacenter's pool of
servers to be non-empty.

NOTE: this should probably be backported to 1.10 but it'll have to be manual.

### DIFF
--- a/.changelog/11748.txt
+++ b/.changelog/11748.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+areas: **(Enterprise only)** make the gRPC server tracker network area aware
+```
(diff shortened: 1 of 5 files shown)


dhiaayachi (approved):
LGTM!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/518417.

PR #11756, opened 2021-12-06
ui: Make 'dangerous' buttons have white text even in dark theme

johncowen:
We recently added our 'auto theming' `--tone` colors, which are just reversed when using a dark theme. We also built in an escape hatch to be able to use colors that we didn't want ever to be themed i.e. we can then inject from The Outside using specific CSS selecting using a non `--tone`ed, non-reversing color just for those cases.

The TLDR:

Before:

Screenshot 2021-12-06 at 19 02 52

After:

Screenshot 2021-12-06 at 19 00 34

This one specifically was a good one to "go wrong" as it gave me a chance to have a little bit of a think how this approach would work with our ['composite color set' `%frame` idea from a while back](https://github.com/hashicorp/consul/pull/4623), as our 'dangerous button' styling here uses a `%frame`. First impressions of that are that we probably need a similar set of `%frame-tones` here as well as similar escape hatches. Another note here is that it _may not_ be worth doing that if these frames aren't shared across design. Whilst they are super useful to define combinations of colors to use from a DX perspective, if we can't reference them back to designs then they become less useful (half the reason why we don't use them as much as we could)

Small note: I've not added a 1.11 Milestone here as yet

### DIFF
--- a/ui/packages/consul-ui/app/styles/themes.scss
+++ b/ui/packages/consul-ui/app/styles/themes.scss
@@ -9,6 +9,11 @@
 %main-nav-horizontal {
   @extend %theme-dark;
 }
+%main-nav-horizontal .dangerous button:hover,
+%main-nav-horizontal .dangerous button:focus {
+  color: rgb(var(--white)) !important;
+}
+
 %main-nav-vertical .menu-panel a:hover,
 %main-nav-vertical .menu-panel a:focus {
   background-color: rgb(var(--tone-blue-500));


kaxcode (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/522036.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 340a0e03f5560a9474e22684bd09ec8c95994a6e onto `release/1.11.x` succeeded!

PR #11763, opened 2021-12-07
ui: Fixup notifications for token using and topology intention saving

johncowen:
I noticed that some of our notifications had lost their text content following our reshuffle of notifications in https://github.com/hashicorp/consul/pull/11577 (towards getting rid of the old style notifications). The changes here all kind of make sense:

- The intention notifications here are for the topology view which still have a small bit of old code that we are yet to move over - all other intention areas use the new approach still.
- "Using" a token is part of our completely untouched for ages (but soon to be) ACLs area and is still therefore using the old style

Please note when I say 'style' above I mean 'style of code' not 'visual style'

If we were keeping this code longer term, there would be questions about making this less copypasta i.e. there are better ways to do it - but as this is a means to an end, its a case of move what we already have to a place where its easy to slowly remove as we go.

(No changelog as https://github.com/hashicorp/consul/pull/11577 isn't in a release as yet)

### DIFF
--- a/ui/packages/consul-ui/app/components/consul/intention/notifications/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/intention/notifications/index.hbs
@@ -1,0 +1,24 @@
+{{#if (eq @type 'create')}}
+  {{#if (eq @status 'success') }}
+    Your intention has been added.
+  {{else}}
+    There was an error adding your intention.
+  {{/if}}
+{{else if (eq @type 'update') }}
+  {{#if (eq @status 'success') }}
+    Your intention has been saved.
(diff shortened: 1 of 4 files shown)


kaxcode (approved):
👍 LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/522103.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit c434fefda25ee480812306e1a6ee46af92684208 onto `release/1.11.x` succeeded!

PR #11765, opened 2021-12-07
ui: reuse BucketList for intention view pages

johncowen:
I noticed that our Intention view/readonly pages had an old design for showing our `partition / namespace` 'BucketList'. Additionally in this page we need to additionally show the service name, so I added the ability to include that in the 'BucketList' - for which I told myself that a `Service` is still a bucket as it contains `Service Instances` 😬 

Before:

Screenshot 2021-12-08 at 10 54 24

After:

Screenshot 2021-12-08 at 10 56 15

### DIFF
--- a/ui/packages/consul-ui/app/components/consul/bucket/list/README.mdx
+++ b/ui/packages/consul-ui/app/components/consul/bucket/list/README.mdx
@@ -11,22 +11,41 @@ the namespace will be displayed, whereas if the partition is different it will
 show both the partition and namespace (as a namespace called 'team-1' in
 `partition-1` is different to a namespace called 'team-1' in `partition-2`)
 
-If you don't need the nspace only support for the view you are building then
-omit the `@nspace` argument.
+Showing the service name is a tiny bit awkward (different boolean type,
+doesn't care about difference) and could be improved but we only use it for
+the read only view of intentions.
 
 At the time of writing, this is not currently used across the entire UI
 (specifically in intentions and maybe other areas) but eventually should be.
 
 
 ```hbs preview-template
-<DataSource
-  @src={{uri "/partition/default/dc-1/gateways/for-service/service-name"}} as |source|>
+  <figure>
+    <figcaption>Show everything</figcaption>
     <Consul::Bucket::List
-      @item={{object-at 0 source.data}}
+      @item={{hash
+        Namespace="different-nspace"
+        Partition="different-partition"
+        Service="service-name"
+      }}
+      @nspace={{'nspace'}}
+      @partition={{'partition'}}
+      @service={{true}}
+    />
+  </figure>
+
+  <figure>
+    <figcaption>Show just the top-level buckets</figcaption>
+    <Consul::Bucket::List
+      @item={{hash
+        Namespace="different-nspace"
+        Partition="different-partition"
+      }}
       @nspace={{'nspace'}}
       @partition={{'partition'}}
     />
-</DataSource>
+  </figure>
+
 ```
 
 ## Arguments
@@ -36,6 +55,7 @@ At the time of writing, this is not currently used across the entire UI
 | `item` | `array` |  | A Consul object that could have both a `Partition` and a `Namespace` property |
 | `nspace` | `string` | | The name of the current namespace |
 | `partition` | `string` | | The name of the current partition |
+| `service` | `boolean` | | Whether to show the service name on the end or not. Please note you must also pass a item.Service for it to show. We flag this incase an API request has a Service property but you don't want to show it |
 
 ## See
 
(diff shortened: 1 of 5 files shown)


ui/packages/consul-ui/app/components/consul/bucket/list/README.mdx
```diff
@@ -36,6 +55,7 @@ At the time of writing, this is not currently used across the entire UI
 | `item` | `array` |  | A Consul object that could have both a `Partition` and a `Namespace` property |
 | `nspace` | `string` | | The name of the current namespace |
 | `partition` | `string` | | The name of the current partition |
+| `service` | `boolean` | | Whether to show the service name on the end or not. Please note you must also pass a item.Service for it to show. We flag this incase an API request has a Service property but you don't want to show it |
```
jgwhite, line 58: Very minor take-it-or-leave-it suggestion just to prove I’ve read the docs:

Could we name this argument a little more explicitly? Maybe `showServiceName` rather than just `service`? My rationale it that its two sibling arguments (`namespace` and `partition`) both take strings and so my initial assumption was that `service` would take a service name string.

ui/packages/consul-ui/app/components/consul/bucket/list/index.hbs
```diff
@@ -19,11 +19,23 @@
       <dd>
         {{@item.Namespace}}
       </dd>
+
+{{#if (and @service @item.Service)}}
```
jgwhite, line 23: Is this indentation intentional?

ui/packages/consul-ui/app/components/consul/bucket/list/index.hbs
```diff
@@ -33,7 +45,16 @@
       <dd>
         {{@item.Namespace}}
       </dd>
+{{#if (and @service @item.Service)}}
```
jgwhite, line 48: Same question as above regarding indentation.

ui/packages/consul-ui/app/components/consul/intention/view/index.hbs
```diff
@@ -4,11 +4,29 @@
...
+          <Consul::Bucket::List
+            @item={{hash
+              Namespace=item.DestinationNS
+              Partition=item.DestinationPartition
+              Service=item.DestinationName
+            }}
+            @nspace="-"
+            @partition="-"
+            @service={{true}}
+          />
```
jgwhite, line 16: This reads super well now 👍

jgwhite (approved):
Looks great. Only very very minor comments from me.

ui/packages/consul-ui/app/components/consul/bucket/list/index.hbs
johncowen: I 'think' this is in reference to the two different indentation level thing I do quite often? If it is, then yeah its intentional. In Consul you'll often find files that have this idea of two levels of indentation. The HTML-y bits have their own indention which grows to the right the more nested it gets, as you would expect, I sometimes then put the logical handlebars-y stuff over to the left. In my eyes when scanning over a file quickly it's less dense and far easier to figure out whats going on. I know some folks find it strange 😅, but it works.

jgwhite: Cool, just checking it wasn’t an auto-formatter like Prettier making mischief 🙂

ui/packages/consul-ui/app/components/consul/bucket/list/README.mdx
johncowen: Yeah I get your thinking here, I wish all the arguments worked/were shaped the same also, I think its half the reason I spent a bit of extra effort on the docs and called it out in the PR, if I'm not happy with something myself, I'll generally call it out for ideas.

Opposing viewpoint here is: I'm always cautious of more than simple or commonly used argument names as they have a different problem of remembering them, or mistyping them.

I think the other alternative would be change it to accept the name of the service itself and if its set then use that, then we keep the super simple `service` lowercased argument names which is somewhat of a common thing in the codebase 🤔

Let me think a bit (or let me know if you'd prefer one over the other if we have two options). Typically this option has the downside of not working like the others which are kind of a "if its not this exact specified name then show the thing", whereas the service one would be just "show this thing"

I'd like to change it just not sure which way to go.

ui/packages/consul-ui/app/components/consul/bucket/list/index.hbs
johncowen: nah but as you can imagine it will make mischief once we come to use prettier on the templates 😂  I guess the advantage will be formatting just happens so it will be a case of out with the old in with the new 😄

msiege2 (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/525504.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit e4d33187b971874dbdd421593cc8e9897553e3a8 onto `release/1.11.x` succeeded!

PR #11767, opened 2021-12-07
ui: Ensure we show a special readonly page for intentions

johncowen:
I noticed we were no longer showing our specially designed read only page for intentions due to an undefined handlebars variable, so I fixed that up and added an acceptance test to prevent this happening here again.

It seems like this bug is only/specifically in 1.10.4 (and currently unreleased 1.11 branch)

### DIFF
--- a/ui/packages/consul-ui/app/components/consul/intention/form/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/intention/form/index.hbs
@@ -222,6 +222,7 @@ as |item readonly|}}
 
     <Consul::Intention::View
       @item={{item}}
+      data-test-readonly
     />
   {{/if}}
 {{/let}}
--- a/ui/packages/consul-ui/app/components/consul/intention/form/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/intention/form/index.hbs
@@ -56,7 +56,10 @@ as |api|>
     </BlockSlot>
 
     <BlockSlot @name="form">
-{{#let api.data as |item|}}
+{{#let
+  api.data
+  (not (can 'write intention' item=api.data))
+as |item readonly|}}
 {{#if (not readonly)}}
 
   {{#let (changeset-get item 'Action') as |newAction|}}
@@ -193,6 +196,7 @@ as |api|>
         </div>
       </form>
   {{else}}
+
     {{#if item.IsManagedByCRD}}
       <Notice
         class="crd"
@@ -215,6 +219,7 @@ as |api|>
         </notice.Footer>
       </Notice>
     {{/if}}
+
     <Consul::Intention::View
       @item={{item}}
     />
--- a/ui/packages/consul-ui/tests/acceptance/steps/dc/intentions/read-only-steps.js
+++ b/ui/packages/consul-ui/tests/acceptance/steps/dc/intentions/read-only-steps.js
@@ -1,0 +1,10 @@
+import steps from '../../steps';
+
+// step definitions that are shared between features should be moved to the
+// tests/acceptance/steps/steps.js file
+
+export default function(assert) {
+  return steps(assert).then('I should find a file', function() {
+    assert.ok(true, this.step);
+  });
+}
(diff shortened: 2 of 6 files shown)


ui/packages/consul-ui/app/components/consul/intention/form/index.hbs
```diff
@@ -56,7 +56,10 @@ as |api|>
     </BlockSlot>
 
     <BlockSlot @name="form">
-{{#let api.data as |item|}}
+{{#let
+  api.data
+  (not (can 'write intention' item=api.data))
```
johncowen, line 61: This line here was the bugfix, previously the `readonly` variable just below on line 63 was undefined always 😬 

Most of the other change here is to test this so it doesn't happen again

jgwhite: Maybe an opportunity to use [`cannot`](https://github.com/minutebase/ember-can#cannot)?

```suggestion
  (cannot 'write intention' item=api.data)
```

ui/packages/consul-ui/tests/acceptance/steps/dc/intentions/read-only-steps.js
```diff
@@ -0,0 +1,10 @@
+import steps from '../../steps';
+
+// step definitions that are shared between features should be moved to the
+// tests/acceptance/steps/steps.js file
+
+export default function(assert) {
+  return steps(assert).then('I should find a file', function() {
+    assert.ok(true, this.step);
+  });
+}
```
jgwhite, line 10: Did you end up using this step? Looks like it always passes?

jgwhite (changes requested):
Looks good. One little suggestion and one clarification regarding a bit of testing code.

ui/packages/consul-ui/app/components/consul/intention/form/index.hbs
johncowen: In the codebase we prefer using `not (can "thing")`, that way folks always use the same method to flip the conditional no matter if it's using `can` or some other flippable helper. So I'll probably leave this one.

ui/packages/consul-ui/tests/acceptance/steps/dc/intentions/read-only-steps.js
johncowen: Ah this is where it's just lack of context into the addon we use for our acceptance tests.

The addon _requires_ these files in order to run (the idea is you write a entire set of steps for every test, which seems like a lot of repetitive work). Instead we have about 25-30 reusable steps that we use for all our tests. Unfortunately the addon _requires_ these files to still exist and generates them exactly like this for each and every test. If you delete them it breaks. Therefore we have oodles of these files in a nested folder structure that are all exactly the same. It's annoying, but I just pretend they aren't there.

We had a PR a while ago to stop it from doing this which made it much nicer, but unfortunately it didn't go in and has been the root of confusion both previous to that PR and ever since 😹

ui/packages/consul-ui/app/components/consul/intention/form/index.hbs
jgwhite: Fair enough

ui/packages/consul-ui/tests/acceptance/steps/dc/intentions/read-only-steps.js
jgwhite: Yikes, that isn’t ideal. Still, explanation makes sense for the context of this PR.

jgwhite (approved):

msiege2 (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/525536.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit be23aab00101228b5fea06df8066f150a15548e9 onto `release/1.11.x` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit be23aab00101228b5fea06df8066f150a15548e9 onto `release/1.10.x` succeeded!

PR #11768, opened 2021-12-07
Bump go version to 1.17.4

kisunji:


kisunji:
Pending publish of `circleci/golang:1.17.4` image here: https://hub.docker.com/r/circleci/golang/tags?page=1&name=1.17.4

dnephin (commented):

I guess we don't need to update `.github/workflows/build.yml` that we use for releases because that uses a less specific version (1.17) ?

I wonder if we should use a more specific version for releases as well, that seems to be even more critical to pin than our CI version.

kisunji:
> Thanks for making this change!
> 
> I guess we don't need to update `.github/workflows/build.yml` that we use for releases because that uses a less specific version (1.17) ?
> 
> I wonder if we should use a more specific version for releases as well, that seems to be even more critical to pin than our CI version.

Yeah that's a great point. We need traceability of the version used to build releases in case of performance regressions.

PR #4672, opened 2018-09-13
connect/ca: add methods for generating and signing intermediate CSRs

kyhavlov:
This PR adds CA functionality for creating and signing intermediate CSRs to be used across providers in different datacenters.

PR #11783, opened 2021-12-08
ca: add a test that uses an intermediate CA as the primary CA

dnephin:
Related to #11598

This test found a bug in the secondary. We were appending the root cert to the PEM, but that cert was already appended. This was failing validation in Vault here:
https://github.com/hashicorp/vault/blob/sdk/v0.3.0/sdk/helper/certutil/types.go#L329

Previously this worked because self signed certs have the same SubjectKeyID and AuthorityKeyID. So having the same self-signed cert repeated doesn't fail that check.

However with an intermediate that is not self-signed, those values are different, and so we fail the check. A test I added in a previous commit (a4ba1f348ddbba770a876d32469fdbc981f24bfe) should show that this continues to work with self-signed root certs as well.

I traced this line back to the original PR that added support for secondaries (#4672), and I don't see any mention of why. So I think this may have simply been an oversight.

### DIFF
--- a/agent/connect/ca/provider_vault.go
+++ b/agent/connect/ca/provider_vault.go
@@ -16,10 +16,9 @@ import (
 	vaultapi "github.com/hashicorp/vault/api"
 	"github.com/mitchellh/mapstructure"
 
-	"github.com/hashicorp/consul/lib/decode"
-
 	"github.com/hashicorp/consul/agent/connect"
 	"github.com/hashicorp/consul/agent/structs"
+	"github.com/hashicorp/consul/lib/decode"
 )
 
 const (
@@ -400,17 +399,14 @@ func (v *VaultProvider) SetIntermediate(intermediatePEM, rootPEM string) error {
 		return fmt.Errorf("cannot set an intermediate using another root in the primary datacenter")
 	}
 
-	err := validateSetIntermediate(
-		intermediatePEM, rootPEM,
-		"", // we don't have access to the private key directly
-		v.spiffeID,
-	)
+	// the private key is in vault, so we can't use it in this validation
+	err := validateSetIntermediate(intermediatePEM, rootPEM, "", v.spiffeID)
 	if err != nil {
 		return err
 	}
 
 	_, err = v.client.Logical().Write(v.config.IntermediatePKIPath+"intermediate/set-signed", map[string]interface{}{
-		"certificate": fmt.Sprintf("%s\n%s", intermediatePEM, rootPEM),
+		"certificate": intermediatePEM,
 	})
 	if err != nil {
 		return err
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -654,7 +654,7 @@ func (c *CAManager) secondaryInitializeIntermediateCA(provider ca.Provider, conf
 	}
 
 	if needsNewIntermediate {
-		if err := c.secondaryRenewIntermediate(provider, newActiveRoot); err != nil {
+		if err := c.secondaryRequestNewSigningCert(provider, newActiveRoot); err != nil {
 			return err
 		}
 	} else {
@@ -1028,9 +1028,11 @@ func (c *CAManager) primaryRenewIntermediate(provider ca.Provider, newActiveRoot
 	return nil
 }
 
-// secondaryRenewIntermediate should only be called while the state lock is held by
-// setting the state to non-ready.
-func (c *CAManager) secondaryRenewIntermediate(provider ca.Provider, newActiveRoot *structs.CARoot) error {
+// secondaryRequestNewSigningCert creates a Certificate Signing Request, sends
+// the request to the primary, and stores the received certificate in the
+// provider.
+// Should only be called while the state lock is held by setting the state to non-ready.
+func (c *CAManager) secondaryRequestNewSigningCert(provider ca.Provider, newActiveRoot *structs.CARoot) error {
 	csr, err := provider.GenerateIntermediateCSR()
 	if err != nil {
 		return err
@@ -1145,7 +1147,7 @@ func (c *CAManager) RenewIntermediate(ctx context.Context, isPrimary bool) error
 	// Enough time has passed, go ahead with getting a new intermediate.
 	renewalFunc := c.primaryRenewIntermediate
 	if !isPrimary {
-		renewalFunc = c.secondaryRenewIntermediate
+		renewalFunc = c.secondaryRequestNewSigningCert
 	}
 	errCh := make(chan error, 1)
 	go func() {
--- a/agent/consul/leader_connect_ca_test.go
+++ b/agent/consul/leader_connect_ca_test.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	msgpackrpc "github.com/hashicorp/net-rpc-msgpackrpc"
+	vaultapi "github.com/hashicorp/vault/api"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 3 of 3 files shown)


agent/connect/ca/provider_vault.go
```diff
@@ -396,17 +396,14 @@ func (v *VaultProvider) SetIntermediate(intermediatePEM, rootPEM string) error {
...
-	)
+	// the private key is in vault, so we can't use it in this validation
+	err := validateSetIntermediate(intermediatePEM, rootPEM, "", v.spiffeID)
 	if err != nil {
 		return err
 	}
 
 	_, err = v.client.Logical().Write(v.config.IntermediatePKIPath+"intermediate/set-signed", map[string]interface{}{
-		"certificate": fmt.Sprintf("%s\n%s", intermediatePEM, rootPEM),
+		"certificate": intermediatePEM,
```
dnephin, line 406: The `intermediatePEM` already has the root cert appended. I can't see anywhere we do this in Consul, so it must come from the CSR response returned by Vault. The Vault API docs are not really clear about this: https://www.vaultproject.io/api/secret/pki

agent/consul/leader_connect_ca.go
```diff
@@ -1070,9 +1072,11 @@ func (c *CAManager) primaryRenewIntermediate(provider ca.Provider, newActiveRoot
...
 }
 
-// secondaryRenewIntermediate should only be called while the state lock is held by
-// setting the state to non-ready.
-func (c *CAManager) secondaryRenewIntermediate(provider ca.Provider, newActiveRoot *structs.CARoot) error {
+// secondaryRequestNewSigningCert creates a Certificate Signing Request, sends
+// the request to the primary, and stores the received certificate in the
+// provider.
+// Should only be called while the state lock is held by setting the state to non-ready.
+func (c *CAManager) secondaryRequestNewSigningCert(provider ca.Provider, newActiveRoot *structs.CARoot) error {
```
dnephin, line 1079: Renamed because this is not really renew, it's creating a new one, which is also done on initialize.

agent/consul/leader_connect_ca_test.go
```diff
@@ -607,6 +608,87 @@ func TestCAManager_UpdateConfiguration_Vault_Primary(t *testing.T) {
...
+		err := msgpackrpc.CallWithCodec(codec, "ConnectCA.Roots", &structs.DCSpecificRequest{}, &roots)
+		require.NoError(t, err)
+		require.Len(t, roots.Roots, 1)
+
+		leafCertPEM := getLeafCert(t, codec, roots.TrustDomain, "dc1")
+		verifyLeafCert(t, roots.Roots[0], leafCertPEM)
+	})
+
+	// TODO: renew primary leaf signing cert
+	// TODO: rotate root
```
dnephin, line 655: Planning on doing these TODOs in a follow up

agent/consul/leader_connect_ca_test.go
```diff
@@ -607,6 +608,87 @@ func TestCAManager_UpdateConfiguration_Vault_Primary(t *testing.T) {
...
+			c.PrimaryDatacenter = "dc1"
+			c.CAConfig = &structs.CAConfiguration{
+				Provider: "vault",
+				Config: map[string]interface{}{
+					"Address":             vault.Addr,
+					"Token":               vault.RootToken,
+					"RootPKIPath":         meshRootPath,
+					"IntermediatePKIPath": "pki-secondary/",
+					// TODO: there are failures to init the CA system if these are not set
+					// to the values of the already initialized CA.
```
dnephin, line 669: For this to work the user has to initialize the Vault mount themselves, which means they will need to set these to the values they used to initialize the mount.

I guess this will need to be documented, so I've left this TODO to remind myself, but I should probably find a better place for that reminder.

agent/connect/ca/provider_vault.go
dnephin: Some questions about this:
1. Do other providers also append the root cert to the PEM when responding to CSR requests? From what I can tell, the Consul provider does not.
2. Does it matter? Maybe having the root cert PEM here is only important for Vault.

dnephin: OK, so looking at other providers:

* The consul provider [stores it](https://github.com/hashicorp/consul/blob/main/agent/connect/ca/provider_consul.go#L268), but [never uses it](https://github.com/hashicorp/consul/blob/main/agent/connect/ca/provider_consul.go#L280-L290)  (because `SetIntermediate` is only called in secondary DCs). 
* The aws provider does basically the same thing as the Consul provider. Stores it in-process, but never uses it in the secondary.

So I think this change is only relevant to the vault provider.

agent/consul/leader_connect_ca_test.go
```diff
@@ -606,6 +607,88 @@ func TestCAManager_UpdateConfiguration_Vault_Primary(t *testing.T) {
...
+	_, s1 := testServerWithConfig(t, func(c *Config) {
+		c.CAConfig = &structs.CAConfiguration{
+			Provider: "vault",
+			Config: map[string]interface{}{
+				"Address":             vault.Addr,
+				"Token":               vault.RootToken,
+				"RootPKIPath":         meshRootPath,
+				"IntermediatePKIPath": "pki-intermediate/",
+				// TODO: there are failures to init the CA system if these are not set
+				// to the values of the already initialized CA.
```
kisunji, line 632:
```suggestion
				// to the values of the CA initialized by setupPrimaryCA.
```

agent/consul/leader_connect_ca_test.go
```diff
@@ -606,6 +607,88 @@ func TestCAManager_UpdateConfiguration_Vault_Primary(t *testing.T) {
...
+			c.PrimaryDatacenter = "dc1"
+			c.CAConfig = &structs.CAConfiguration{
+				Provider: "vault",
+				Config: map[string]interface{}{
+					"Address":             vault.Addr,
+					"Token":               vault.RootToken,
+					"RootPKIPath":         meshRootPath,
+					"IntermediatePKIPath": "pki-secondary/",
+					// TODO: there are failures to init the CA system if these are not set
+					// to the values of the already initialized CA.
```
kisunji, line 669:
```suggestion
					// to the values of the CA initialized by setupPrimaryCA.
```

kisunji (approved):
Non-blocking comment to clarify where `PrivateKeyType` and `PrivateKeyBits` came from but LGTM

agent/consul/leader_connect_ca_test.go
dnephin: I think these suggested edits are the opposite of what is happening. 

This test case initializes the CA in Vault itself, so we're setting the config values to those of the CA in Vault. `setupPrimaryCA` will ignore the values because Vault already has a CA setup.

I opened #12246 yesterday to track this TODO. Once that work is done I can come back to these tests and remove the TODO and the two lines below.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/573405.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit df2a7e2ae5340973f60cdd867e0c1347d5af889d onto `release/1.11.x` succeeded!

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit df2a7e2ae5340973f60cdd867e0c1347d5af889d onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/573405)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit df2a7e2ae5340973f60cdd867e0c1347d5af889d onto `release/1.9.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/573405)

agent/consul/leader_connect_ca_test.go
kisunji: Ah got it, thanks for clarifying!

PR #11787, opened 2021-12-09
Add test jobs for arm64 in CircleCI

odidev:
Added test jobs for arm64 in CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,6 +112,10 @@ commands:
           echo "export AWS_SESSION_TOKEN=$(echo $CREDENTIALS | jq -r '.SessionToken')" >> $BASH_ENV
 
   run-go-test-full:
+    parameters:
+      go_test_flags:
+        type: string
+        default: ""
     steps:
       - attach_workspace:
           at: /home/circleci/go/bin
@@ -127,6 +131,9 @@ commands:
             echo $PACKAGE_NAMES
             # some tests expect this umask, and arm images have a different default
             umask 0022
+
+            << parameters.go_test_flags >>
+
             gotestsum \
               --format=short-verbose \
               --jsonfile /tmp/jsonfile/go-test-${CIRCLE_NODE_INDEX}.log \
@@ -137,6 +144,7 @@ commands:
               --packages="$PACKAGE_NAMES" \
               --junitfile $TEST_RESULTS_DIR/gotestsum-report.xml -- \
               -tags="$GOTAGS" -p 2 \
+              ${GO_TEST_FLAGS-} \
               -cover -coverprofile=coverage.txt
 
       - store_test_results:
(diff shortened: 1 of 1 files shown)


dnephin (commented):
Thank you for the PR! I noticed CI status was not being updated on this PR, but I guess that's because you have CircleCI enabled on your fork, so the builds are running here: https://app.circleci.com/pipelines/github/odidev/consul?branch=arm64-testing

I pushed one commit to share more of the CI config with the two jobs, and remove the `dev-build` step. Installing Go takes some time, so I think by running `make dev` in the CI job we can avoid doing that work twice.

If CI is still happy with this change I think we're all set to merge.

dnephin:
Ok, tests here look good: https://app.circleci.com/pipelines/github/odidev/consul/9/workflows/ba72d55e-7313-4e3b-ad09-0cf2ec81ece9

dnephin (approved):

dnephin:
Hmm, but it looks like I can't merge without `verify-ci` passing. 

I think you may need to click "Stop Building" in circleCI for your fork. Then we can CI again, it should report the CI run status to this PR, so that we can merge it.

odidev:
> Hmm, but it looks like I can't merge without `verify-ci` passing.
> 
> I think you may need to click "Stop Building" in circleCI for your fork. Then we can CI again, it should report the CI run status to this PR, so that we can merge it.

I have stopped building in CircleCI. I hope you can merge this PR now.

dnephin:
Now that CI is running on this branch and reporting the time to github, we noticed that the runtime of this new Job is quite a bit longer than the rest. I think a couple minutes of this time is having to build a binary for arm64, but a lot of it is that the arm machines are just slower to run the tests(not sure why).

After some internal discussion we decided to try and run the tests with `-short` when the tests are being run on a PR. This will skip longer tests to reduce the runtime. On `main` and release branches we will continue to run the full test suite. 

To accomplish this we need to make some changes to the CI config. I can do that, but it may be week or more before I find the time. 

Letting you know since there has not been an update on this PR in a while.

dnephin:
Ok, I pushed some commits that use `-short` and skip building the binary when run on PR branches. On `main` and `release/` branches it will build a binary and run the full test suite.

odidev:
> Ok, I pushed some commits that use `-short` and skip building the binary when run on PR branches. On `main` and `release/` branches it will build a binary and run the full test suite.

I have also added and ran the test suite but only 60-70 tests were running for arm64 and the rest were getting skipped and the number of test cases running for each build is also not constant.

dhiaayachi (approved):
🚀

dnephin:
Ok, I'll merge this now and check to make sure it runs correctly on `main`. I did some manual testing, but it's hard to test exactly with the CI config ahead of time.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/577300.

PR #11796, opened 2021-12-10
testing: stop using an old version in testServer

dnephin:
Our tests should run with the current version by default.

Also deprecate some functions, see commit message for the rationale.

### DIFF
--- a/agent/consul/server_test.go
+++ b/agent/consul/server_test.go
@@ -66,21 +66,12 @@ func testTLSCertificates(serverName string) (cert string, key string, cacert str
 	return cert, privateKey, ca, nil
 }
 
-// testServerACLConfig wraps another arbitrary Config altering callback
-// to setup some common ACL configurations. A new callback func will
-// be returned that has the original callback invoked after setting
-// up all of the ACL configurations (so they can still be overridden)
-func testServerACLConfig(cb func(*Config)) func(*Config) {
-	return func(c *Config) {
-		c.PrimaryDatacenter = "dc1"
-		c.ACLsEnabled = true
-		c.ACLInitialManagementToken = TestDefaultMasterToken
-		c.ACLResolverSettings.ACLDefaultPolicy = "deny"
-
-		if cb != nil {
-			cb(c)
-		}
-	}
+// testServerACLConfig setup some common ACL configurations.
+func testServerACLConfig(c *Config) {
+	c.PrimaryDatacenter = "dc1"
+	c.ACLsEnabled = true
+	c.ACLInitialManagementToken = TestDefaultMasterToken
+	c.ACLResolverSettings.ACLDefaultPolicy = "deny"
 }
 
 func configureTLS(config *Config) {
@@ -185,14 +176,12 @@ func testServerConfig(t *testing.T) (string, *Config) {
 	return dir, config
 }
 
+// Deprecated: use testServerWithConfig instead. It does the same thing and more.
 func testServer(t *testing.T) (string, *Server) {
-	return testServerWithConfig(t, func(c *Config) {
-		c.Datacenter = "dc1"
-		c.PrimaryDatacenter = "dc1"
-		c.Bootstrap = true
-	})
+	return testServerWithConfig(t)
 }
 
+// Deprecated: use testServerWithConfig
 func testServerDC(t *testing.T, dc string) (string, *Server) {
 	return testServerWithConfig(t, func(c *Config) {
 		c.Datacenter = dc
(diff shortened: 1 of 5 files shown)


agent/consul/server_test.go
```diff
@@ -187,21 +185,20 @@ func testServerConfig(t *testing.T) (string, *Config) {
 	return dir, config
 }
 
+// Deprecated: use testServerWithConfig instead. It does the same thing and more.
 func testServer(t *testing.T) (string, *Server) {
-	return testServerWithConfig(t, func(c *Config) {
-		c.Datacenter = "dc1"
-		c.PrimaryDatacenter = "dc1"
-		c.Bootstrap = true
```
dnephin, line 194: These are already set in `testServerWithConfig`, so I don't think setting them here again does anything.

agent/consul/server_test.go
```diff
@@ -218,25 +216,16 @@ func testServerDCExpect(t *testing.T, dc string, expect int) (string, *Server) {
...
-	return testServerWithConfig(t, func(c *Config) {
-		c.Datacenter = dc
-		c.Bootstrap = false
-		c.BootstrapExpect = expect
-		c.ReadReplica = true
-	})
-}
-
-func testServerWithConfig(t *testing.T, cb func(*Config)) (string, *Server) {
+func testServerWithConfig(t *testing.T, configOpts ...func(*Config)) (string, *Server) {
```
dnephin, line 219: By making this a varargs, I think we'll be able to:
* rename this to `testServer` since calling it with no `configOpts` functions is fine
* instead of separate `testServerX` methods, we can create `configOpts` functions and pass them in (ex: one for enabling ACLs, one for TLS, one for secondary). That allows us to compose the setup functions without wrapping the main function.


I'd like to also remove `newServer` and rename `testServer` to `newTestServer`, but I thought I'd do all of those renames at once in a separate PR, since it will change a few hundred lines.

agent/consul/server_test.go
```diff
@@ -1284,7 +1273,11 @@ func TestServer_Expect_NonVoters(t *testing.T) {
 	}
 
 	t.Parallel()
-	dir1, s1 := testServerDCExpectNonVoter(t, "dc1", 2)
+	dir1, s1 := testServerWithConfig(t, func(c *Config) {
+		c.Bootstrap = false
+		c.BootstrapExpect = 2
+		c.ReadReplica = true
+	})
```
dnephin, line 1280: This was the only caller of `testServerDCExpectNonVoter` so I moved the settings here instead of deprecating the function.

The other ones have a few dozen callers, so I left them.

agent/consul/server_test.go
```diff
@@ -164,8 +155,6 @@ func testServerConfig(t *testing.T) (string, *Config) {
 	config.ServerHealthInterval = 50 * time.Millisecond
 	config.AutopilotInterval = 100 * time.Millisecond
 
-	config.Build = "1.7.2"
```
FFMMM, line 167: IIUC, there's one more instance of this in https://github.com/hashicorp/consul/blob/85fe875d0772a6da4b42c03795ab4b44e8f92182/agent/consul/leader_test.go#L1440 

these seem to be more like placeholders (gossiped around) than actual config knobs right?

https://github.com/hashicorp/consul/blob/eb21649f82bf78e27a75b012013f695d9beba6af/agent/consul/config.go#L176

FFMMM (approved):
📦 LGTM! Left a small nit/ ref if you want to tackle it!

agent/consul/server_test.go
dnephin: Oh ya, I think there are a few individual tests that do set this value. I think that's ok. We still want to be able to run some tests and pretend the agent is an older version. What I was fixing in this PR was defaulting to an old version. I think the default needs to be the current version.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/536234.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 1683da66b02f980834f1fa43c9744a48f4ba10a6 onto `release/1.11.x` succeeded!

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 1683da66b02f980834f1fa43c9744a48f4ba10a6 onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/536234)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 1683da66b02f980834f1fa43c9744a48f4ba10a6 onto `release/1.9.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/536234)

PR #11798, opened 2021-12-10
leader: move the virtual IP version check into a goroutine

kyhavlov:
This PR moves the server version check for enabling virtual IPs into a leader goroutine to make it more consistent with other logic of this type and limit the log warning rate a bit.

### DIFF
--- a/agent/consul/leader_connect.go
+++ b/agent/consul/leader_connect.go
@@ -125,9 +125,13 @@ func (s *Server) pruneCARoots() error {
 
 func (s *Server) runVirtualIPVersionCheck(ctx context.Context) error {
 	// Return early if the flag is already set.
-	_, err := s.setVirtualIPVersionFlag()
+	done, err := s.setVirtualIPVersionFlag()
 	if err != nil {
 		s.loggers.Named(logging.Connect).Warn("error enabling virtual IPs", "error", err)
+	}
+	if done {
+		s.leaderRoutineManager.Stop(virtualIPCheckRoutineName)
+		return nil
 	}
 
 	ticker := time.NewTicker(virtualIPVersionCheckInterval)
@@ -138,12 +142,12 @@ func (s *Server) runVirtualIPVersionCheck(ctx context.Context) error {
 		case <-ctx.Done():
 			return nil
 		case <-ticker.C:
-			ok, err := s.setVirtualIPVersionFlag()
+			done, err := s.setVirtualIPVersionFlag()
 			if err != nil {
 				s.loggers.Named(logging.Connect).Warn("error enabling virtual IPs", "error", err)
 				continue
 			}
-			if ok {
+			if done {
 				return nil
 			}
 		}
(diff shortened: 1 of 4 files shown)


agent/consul/leader_test.go
```diff
@@ -2126,6 +2126,8 @@ func TestLeader_EnableVirtualIPs(t *testing.T) {
 		t.Skip("too slow for testing.Short")
 	}
 
+	virtualIPVersionCheckInterval = 50 * time.Millisecond
```
dnephin, line 2129: Normally I guess we'd want to reset this back to the original value after the test ends, but I'm guessing in this case it won't matter too much. Might be good to do just in case the pattern is copied in the future.

```suggestion
        orig :=  virtualIPVersionCheckInterval
	virtualIPVersionCheckInterval = 50 * time.Millisecond
	t.Cleanup(func() { virtualIPVersionCheckInterval = orig })
```

agent/consul/leader_connect.go
```diff
@@ -111,6 +123,54 @@ func (s *Server) pruneCARoots() error {
 	return err
 }
 
+func (s *Server) runVirtualIPVersionCheck(ctx context.Context) error {
+	// Return early if the flag is already set.
+	_, err := s.setVirtualIPVersionFlag()
+	if err != nil {
+		s.loggers.Named(logging.Connect).Warn("error enabling virtual IPs", "error", err)
+	}
```
dnephin, line 131: I think we are missing the return here when the first return value is `true`.
```suggestion
	// Return early if the flag is already set.
	done, err := s.setVirtualIPVersionFlag()
	if err != nil {
		s.loggers.Named(logging.Connect).Warn("error enabling virtual IPs", "error", err)
	}
	if done {
		return nil
	}
```

Does that seem right?

dnephin (approved):
Nice! nothing blocking , but a couple suggestions for small improvements

agent/consul/leader_connect.go
kyhavlov: For some reason, doing this causes the goroutine to leak (and TestAgentLeaks_Server to fail) - I looked at this for a bit but i'm not sure why.

dnephin: Ohh, right! I ran into the same problem when I tried to do this, but I also was not sure why.

I think I found a workaround, which was to call `leaderRoutineManager.Stop()`, but now I can't find the link to where I did that. It might be possible to call `Stop` here before returning.

kyhavlov: That worked; calling `leaderRoutineManager.Stop()` before returning was enough.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/525070.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/529998.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 9dcaf0539cb4890753476e512d013154f1b835b2 onto `release/1.11.x` succeeded!

PR #11800, opened 2021-12-10
ui: Fix up if/let nesting in partition menu

johncowen:
I spotted a little bug which meant that when switching to a non-primary datacenter it would not display the 'default' word showing you the partition you are always in in a non-primary.

The fix essentially moves the `let` up by one indentation/conditional.

Whilst doing this I thought I may aswell just hardcode the 'default' word in. I left it in curlies as just as it feels like something that is hardcoded for reasons but is kind of not in a funny kind of way.


jgwhite (approved):

msiege2 (approved):

msiege2 (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/525729.

PR #11801, opened 2021-12-10
ui: Change the URL prefix of partitions from `-` to `_`

johncowen:
We currently use special non-DNS permitted prefixes to detect the URL segments for partitions and namespace in order to keep our URLs as backwards compatible as possible whilst avoiding the use of query parameters `?` for identifying completely different sets of data (as opposed to different views of the same data, which is what we mostly use query params for). For example our URLs can look like any of the following:

- `/ui/datacenter-name/services/service-name`
- `/ui/~namespace-name/datacenter-name/services/service-name`
- `/ui/-partition-name/~namespace-name/datacenter-name/services/service-name`

We decided to change the nicer looking `-` to a `_` for the less common partition segment in order to keep the nicer `-` for other potentially more common purposes.

TLDR; This PR changes the last URL in the above examples to:

- `/ui/_partition-name/~namespace-name/datacenter-name/services/service-name`

There are only a handful of places in the UI where need to change this as its almost all managed in the custom Ember Location we use for our "optional" URL segments, but in order to double check I did a search across the codebase for `~`, which whilst we aren't changing `~` brought up all the areas where we do anything to do with these special prefix characters.

### DIFF
--- a/.changelog/11801.txt
+++ b/.changelog/11801.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+ui: Change partition URL segment prefix from `-` to `_` 
+```
--- a/ui/packages/consul-ui/app/locations/fsm-with-optional.js
+++ b/ui/packages/consul-ui/app/locations/fsm-with-optional.js
@@ -1,7 +1,7 @@
 import { env } from 'consul-ui/env';
 const OPTIONAL = {};
 if (env('CONSUL_PARTITIONS_ENABLED')) {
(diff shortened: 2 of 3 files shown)


jgwhite (approved):
Looks good. General praise: this solution to URL nesting/namespacing is very inventive! I like it a lot.

msiege2 (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/525836.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 79b25901d750f70d7c9d66dd55c83b1b5924fac5 onto `release/1.11.x` succeeded!

PR #11802, opened 2021-12-10
changelog: fix PR links

boxofrad:
I noticed that the PR links in our `CHANGELOG.md` don't seem to work anymore 🤔 I wonder if GitHub no longer supports the `/issues/<id>` path for Pull Requests?


eculver (approved):
Nice catch. Bonus points for regenerating the entire changelog with the updated links :)

boxofrad:
@eculver I wasn't able to figure out how to regenerate the _entire_ changelog with `changelog-build` (particularly as the earlier entries predate the tool) so did a find-and-replace instead.

Interestingly, there are some older links to issues rather than PRs, but the `/pull/<id>` URL redirects to `/issues/<id>` correctly.

mikemorris:
Spot checked a few of the old `/issues/{PR_NUMBER}` links and they appear to be working now. Pull requests are just a "special" type of issue in GitHub's data model/API, could this have been a temporary problem? Suggest we close this unless some are still definitively broken.

PR #11805, opened 2021-12-10
ui: Adding Partition to topology card

kaxcode:
### ✨ Description:
This PR adds Admin Partition to the Topology Metrics Card component. 

Conditionals cover the following scenarios:
- if any service on the page is in a different partition, all the cards show both partition and namespace.
- if all services on the page are in the same partition but have different namespaces, show the namespace label on the ones outside of the current namespace

[DEMO](https://consul-ui-staging-4llmogh9c-hashicorp.vercel.app/ui/dc1/services/service-5/topology)

### 📸 Screenshots:
Screen Shot 2021-12-10 at 3 00 46 PM

### ⚡ Backend Changes:
No backend changes required.

### 🤡 Updates to mock-api:
Added `Partition` to the Upstream and Downstream services in the `service-topology` mock-api.

### 🧪 Testing:
No testing added.

### 📝 **PR Tasks:**

- [x]  Demo Link

### DIFF
--- a/ui/packages/consul-ui/app/components/topology-metrics/card/index.hbs
+++ b/ui/packages/consul-ui/app/components/topology-metrics/card/index.hbs
@@ -29,18 +29,35 @@
     <TopologyMetrics::SourceType @source='routing-config' />
   {{/if}}
   <div class="details">
-  {{#if (and (env 'CONSUL_NSPACES_ENABLED') (not-eq @item.Namespace @service.Namespace))}}
-    <dl class="nspace">
-      <dt>
-        <Tooltip>
-            Namespace
-        </Tooltip>
-      </dt>
-      <dd>
-        {{@item.Namespace}}
-      </dd>
-    </dl>
-  {{/if}}
+    {{#if (and (can 'use partitions') (can 'use nspaces'))}}
+    <div class={{if (not-eq @item.Partition @service.Partition) 'group'}}>
+      {{#if (not-eq @item.Partition @service.Partition)}}
+      <dl class="partition">
+        <dt>
+          <Tooltip>
+              Partition
+          </Tooltip>
+        </dt>
+        <dd>
+          {{@item.Partition}}
+        </dd>
+      </dl>
+      {{/if}}
+      <span></span>
+      {{#if (or (not-eq @item.Partition @service.Partition) (not-eq @item.Namespace @service.Namespace))}}
+      <dl class="nspace">
+        <dt>
+          <Tooltip>
+              Namespace
+          </Tooltip>
+        </dt>
+        <dd>
+          {{@item.Namespace}}
+        </dd>
+      </dl>
+      {{/if}}
+    </div>
+    {{/if}}
 
 {{#if (not-eq @item.Source 'routing-config')}}
 
--- a/ui/packages/consul-ui/app/components/topology-metrics/card/index.hbs
+++ b/ui/packages/consul-ui/app/components/topology-metrics/card/index.hbs
@@ -35,7 +35,7 @@
       <dl class="partition">
         <dt>
           <Tooltip>
-              Partition
+              Admin Partition
           </Tooltip>
         </dt>
         <dd>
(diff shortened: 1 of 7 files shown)


johncowen (changes requested):
Could you change the icons to use the icons we use elsewhere in the UI for partition and namespace please? Ideally if we are going to use the different icons we'd do it all at once so they all look the same. This means you can remove a little bit of change here also. Thankyou!

kaxcode:
> Could you change the icons to use the icons we use elsewhere in the UI for partition and namespace please? Ideally if we are going to use the different icons we'd do it all at once so they all look the same. This means you can remove a little bit of change here also. Thankyou!

@johncowen where else int he codebase do we use a partition icon?

johncowen:
Mostly in service listings and intentions. There's a `Consul::Bucket::List` component thing I made recently which uses it, from what I remember this is then reused in both service listings and intentions (and anywhere else we have this)

Screenshot 2022-01-05 at 09 42 53

ui/packages/consul-ui/app/components/topology-metrics/card/index.scss
```diff
@@ -74,5 +74,37 @@
   }
   .details {
     padding: 0 12px 12px 12px;
+    > *:not(:last-child) {
+      padding-bottom: 6px;
+    }
+    span {
+      display: none;
+    }
```
johncowen, line 82: Do you need this? Is it targeting the same span below where its set to `inline-block`?

ui/packages/consul-ui/app/components/topology-metrics/card/index.scss
```diff
@@ -74,5 +74,37 @@
...
+        display: inline-block;
+        grid-area: union;
+        padding-left: 7px;
+        margin-right: 0px;
+      }
+      span::before {
+        margin-right: 0px;
+        @extend %with-union-mask, %as-pseudo;
+        background-color: rgb(var(--tone-gray-500));
+      }
```
johncowen, line 100: I wondered, if this `span` is just for decorative purposes, could it use the pseudo element of the `dl` or `dd` here? Pretty sure thats what I ended up doing for the bucket list component. 

A bigger thing and not necessarily actionable here, but it would be nice if the same Bucket::List component could be used but just be styled differently depending on whether we want a slash separator or a `union` separator. The markup that Bucket::List uses is almost exactly the same as this. Anyway, not important this last bit (or the first bit for that matter), just half thinking out loud.

johncowen (approved):
LGTM, I left a couple of comments on the `span` usage, but not important in the scheme of things unless you specifically wanted to change up/agreed with any of that, otherwise :shipit:

ui/packages/consul-ui/app/components/topology-metrics/card/index.hbs
```diff
@@ -29,18 +29,35 @@
...
-      </dd>
-    </dl>
-  {{/if}}
+    {{#if (and (can 'use partitions') (can 'use nspaces'))}}
+    <div class={{if (not-eq @item.Partition @service.Partition) 'group'}}>
+      {{#if (not-eq @item.Partition @service.Partition)}}
+      <dl class="partition">
+        <dt>
+          <Tooltip>
+              Partition
```
johncowen, line 38: P.S. Just realised in Bucket::List the tooltip reads 'Admin Partition', tbh not sure which one we should use really, but they should be consistent, so if you could change this one up to Admin Partition while you are here that'd be great!

ui/packages/consul-ui/app/components/topology-metrics/card/index.scss
kaxcode: not needed. removed.

ui/packages/consul-ui/app/components/topology-metrics/card/index.hbs
kaxcode: Updated to say `Admin Partition`

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/547552.

PR #11817, opened 2021-12-13
ui: Only allow partition creation with a single datacenter setup

johncowen:
We have disabled partition creation on the backend if a user has more than one datacenter.

This PR reflects that in the UI by only showing the partition [Create] button if there is only one datacenter.

To share some thoughts/rationale here:

- The first API request we make is for a list of datacenters, before we even show you the UI.
- The list of datacenters is very unlikely to change without refreshing the UI, so I decided to use the EmberData cache seeing as we know we always have this list at the start (we have a thin 'repository' layer over Ember Data so we get this using EmberData's `peekAll`)
- There was a choice to pass the data into the ability via `can 'create partitions' dcs=(arrayOfDcs)` but I decided it would be better to inject the Datacenter Repository Service directly and access the count from there. I went this way thinking that this restriction may change in the future and it's better to have the fact that we need to know the amount of datacenters 'hidden' from the "template developer". If it does change in the future, it's just a matter of tweaking the ability class.

Note: No Changelog label here as I believe all our UI partition related changelogs are being squashed together in the next release.

### DIFF
--- a/ui/packages/consul-partitions/app/templates/dc/partitions/index.hbs
+++ b/ui/packages/consul-partitions/app/templates/dc/partitions/index.hbs
@@ -48,7 +48,15 @@ as |route|>
         </h1>
       </BlockSlot>
       <BlockSlot @name="actions">
-          <a data-test-create href="{{href-to 'dc.partitions.create'}}" class="type-create">Create</a>
+{{#if (can 'create partitions')}}
+        <a 
+          data-test-create 
+          class="type-create"
+          href="{{href-to 'dc.partitions.create'}}"
+        >
+          Create
+        </a>
+{{/if}}
       </BlockSlot>
       <BlockSlot @name="toolbar">
       {{#if (gt items.length 0)}}
--- a/ui/packages/consul-ui/app/abilities/partition.js
+++ b/ui/packages/consul-ui/app/abilities/partition.js
@@ -3,6 +3,7 @@ import { inject as service } from '@ember/service';
 
 export default class PartitionAbility extends BaseAbility {
   @service('env') env;
+  @service('repository/dc') dcs;
 
   resource = 'operator';
   segmented = false;
@@ -12,7 +13,16 @@ export default class PartitionAbility extends BaseAbility {
   }
 
   get canManage() {
-    return this.canCreate;
+    // management currently means "can I write", not necessarily just create
+    return this.canWrite;
+  }
+
+  get canCreate() {
+    // we can only currently create a partition if you have only one datacenter
+    if (this.dcs.peekAll().length > 1) {
+      return false;
+    }
+    return super.canCreate;
   }
 
   get canDelete() {
@@ -20,7 +30,7 @@ export default class PartitionAbility extends BaseAbility {
   }
 
   get canChoose() {
-    if(typeof this.dc === 'undefined') {
+    if (typeof this.dc === 'undefined') {
       return false;
     }
     return this.canUse && this.dc.Primary;


ui/packages/consul-partitions/app/templates/dc/partitions/index.hbs
```diff
@@ -48,7 +48,15 @@ as |route|>
         </h1>
       </BlockSlot>
       <BlockSlot @name="actions">
-          <a data-test-create href="{{href-to 'dc.partitions.create'}}" class="type-create">Create</a>
+{{#if (can 'create partitions')}}
+        <a
```
randallmorey, line 52: For context since I'm new, why is consul not using `LinkTo`?

ui/packages/consul-ui/app/abilities/partition.js
```diff
@@ -12,15 +13,24 @@ export default class PartitionAbility extends BaseAbility {
...
+    return super.canCreate;
   }
 
   get canDelete() {
     return this.item.Name !== 'default' && super.canDelete;
   }
 
   get canChoose() {
-    if(typeof this.dc === 'undefined') {
+    if (typeof this.dc === 'undefined') {
```
randallmorey, line 33: Can we do better?  [For example, Ember `isEmpty()`](https://api.emberjs.com/ember/release/functions/@ember%2Futils/isEmpty)?

ui/packages/consul-partitions/app/templates/dc/partitions/index.hbs
johncowen: It began as "because it's heavier than href-to" apparently technically but also "code-visually" (bad word but not sure how to explain that one) i.e. I preferred to just use an `<a>`. I also think maybe we've dodged some upgrade bullets due to that but not entirely sure.

More importantly, as we've had to add some of our custom routing things its be super helpful to have a better/customizable point to tweak things we needed to do (mainly around the optional routing/URL segments we have).

Longer term, we completely own `href-to` in our codebase now I believe and at some point soon I'd like to add relative routeName ability to it so we have 100% isolated routes that can be dropped in anywhere and "Just Work(tm)" (supposedly 😆 🤞 )

ui/packages/consul-ui/app/abilities/partition.js
johncowen: We don't use this Ember type things anywhere in the codebase at all, and prefer to use vanilla JS equivalents, and we use this specific form everywhere. I've found to be "caught out" by these Ember things not doing what you would expect sometimes. There's comment somewhere else recently similar to this, will find in a sec.

randallmorey: Totally get the desire to go vanilla.  Can it at least be wrapped in a custom utility function that you own?  `isUndefined()` for instance, to clean things up a little.  Just as a rule of thumb I try to avoid typeof and direct string matching in core code.  Some types are complicated to test, for example arrays and NaN, and really _need_ helpers.  All of this is potentially an argument for typed JS.  In the absence of that though, the logic could be wrapped.

johncowen: I'm strongly against changing this.

> Some types are complicated to test, for example arrays and NaN

We wouldn't use this form for Arrays or NaN as there are different native vanilla methods for type checking those which we'd use instead of a `typeof` (or instead of a custom helper function).

Folks quite often change the definition of "Undefined-ness" (is it empty, or undefined, or is it just null, or maybe {size: 0}) Ember functions being a case in point, which just leads to confusion. If we were to have `isUndefined` then we'd also need `isFunction` and then we get on another slippery slope (do we need an isClass and isMethod?), which could also be misinterpreted. In fact I find it quite relieving to find a codebase using these vanilla approaches, as I know I'm not going to be caught out by some re-interpretation of undefined or empty (I believe a lot of Ember things also use these under the hood instead of _their own_ helper functions)

If we did it here, we should change the entire codebase to use a helper function, lots of churn for not a lot of benefit.

Given the fact that any usage of a custom function would need a corresponding import, it's also easier to shorter to type the way it currently is.

I think given we are only talking about this due to prettier (a tool to avoid bikeshedding) adding a space here, its all a bit 🤷 😆 

Honestly, I can't think of many benefits to using a helper function.

randallmorey: It's an interesting conversation nonetheless.  Thanks.

randallmorey (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/540317.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 31dee2340f3acaa8298abbeeddff343f29c37cb9 onto `release/1.11.x` succeeded!

Issue #11471, opened 2021-11-02
Consul filter /catalog/nodes not working


Lord-Y:
Hello guys,

I'm running consul 1.10.3 with ACL enabled etc. So far everything is working perfectly.
The only problem that I have so far is  on the filtering management on /catalog/nodes.

Following the documentation here https://www.consul.io/api-docs/features/filtering and here https://www.consul.io/api-docs/catalog#list-nodes it doesn't works.

Exemple: `curl -H "X-Consul-Token: $CONSUL_HTTP_TOKEN" --request GET "${CONSUL_HTTP_ADDR}/v1/catalog/nodes" --data-urlencode 'filter=Node == "gke-xxxx"'`
This command return all list of nodes instead of the specific one.

Am I doing something wrong?

jkirschner-hashicorp:
Hi @Lord-Y,

Try replacing `--request GET` with `-G` (or `--get`) and let me know whether that fixes your problem.

Consul is expecting the filter to be included as a query parameter (which happens if you use `-G`), not in the body of the request (which happens if you use `--request GET`). This tripped me up also when I tried to reproduce what you were seeing.

Let me know if you have any suggestions for how to help avoid this confusion for others in the future.

For example, there may be an improvement we can make to the docs here. The docs you linked do show using `--get` (rather than `--request GET`), but I wonder whether we should call out that `--request` won't do what you need here.

Lord-Y:
@jkirschner-hashicorp my bad. Yes it's working. The documentation is right.
Maybe adding a warning saying to do not use `--request` will be helpful but I'm sure. 
Thx for the help, I'll close this issue.

jkirschner-hashicorp:
Reopening as a reminder to consider adding a relevant message on this to the docs.

jkirschner-hashicorp:
@Lord-Y: out of curiosity, what did you do to try to investigate what was going wrong here? (e.g., did you look at Consul log output at all?)

I'm wondering whether there's something Consul itself could have done here to help (separate from refining the documentation page).

PR #11821, opened 2021-12-13
http: error if GET request has non-empty body

jkirschner-hashicorp:
Give the user a hint that they might be doing something wrong if their GET request has a non-empty body, which can easily happen using curl's `--data-urlencode` if specifying request type via `--request GET` rather than `--get`.

Partial solution to #11471. Before closing said issue, we should consider whether a small docs change is warranted, too.

## PR History

For a GET request with a non-empty body, this PR initially returned an error. After internal discussions, we decided that ignoring the body of a GET request was conventional and were hesitant to return an error code other than `200 - OK`. As an alternative, we will log a WARN message instead.

## Questions for reviewers

- Do we have any `GET` requests for which having a non-empty body is valid?
  - Seems unlikely

#### curl output
```
curl --verbose --request GET localhost:8500/v1/agent/services --data-urlencode 'filter=Service == "counting"'
*   Trying 127.0.0.1:8500...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8500 (#0)
> GET /v1/agent/services HTTP/1.1
> Host: localhost:8500
> User-Agent: curl/7.68.0
> Accept: */*
> Content-Length: 40
> Content-Type: application/x-www-form-urlencoded
>
* upload completely sent off: 40 out of 40 bytes
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Content-Type: application/json
< Vary: Accept-Encoding
< X-Consul-Default-Acl-Policy: allow
< Date: Tue, 22 Feb 2022 20:25:55 GMT
< Content-Length: 3
<
{}
* Connection #0 to host localhost left intact
```
#### Log messages

```
...
2022-02-22T12:26:32.237-0800 [WARN]  agent.http: GET request has a non-empty body, check whether parameters were accidentally placed in the body rather than in the query string: url=/v1/agent/services from=127.0.0.1:48852
2022-02-22T12:26:32.238-0800 [DEBUG] agent.http: Request finished: method=GET url=/v1/agent/services from=127.0.0.1:48852 latency=68.706µs
...
```

### Valid GET request (filter in query params)

```
$ curl --verbose --get localhost:8500/v1/agent/services --data-urlencode 'filter=Service == "counting"'
...
> GET /v1/agent/services?filter=Service%20%3D%3D%20%22counting%22 HTTP/1.1
...
< HTTP/1.1 200 OK
...
{"counting-id":{"ID":"counting-id","Service":"counting","Tags":[],"Meta":{},"Port":9000,"Address":"","Weights":{"Passing":1,"Warning":1},"EnableTagOverride":false,"Datacenter":"dc1"}}
```

### DIFF
--- a/agent/http.go
+++ b/agent/http.go
@@ -543,6 +543,17 @@ func (s *HTTPHandlers) wrap(handler endpoint, methods []string) http.HandlerFunc
 		} else {
 			err = s.checkWriteAccess(req)
 
+			// Give the user a hint that they might be doing something wrong if they issue a GET request
+			// with a non-empty body (e.g., parameters placed in body rather than query string).
+			if req.Method == http.MethodGet {
+				if req.ContentLength > 0 {
+					httpLogger.Warn("GET request has a non-empty body that will be ignored; "+
+						"check whether parameters meant for the query string were accidentally placed in the body",
+						"url", logURL,
+						"from", req.RemoteAddr)
+				}
+			}
+
 			if err == nil {
 				// Invoke the handler
 				obj, err = handler(resp, req)
--- a/.changelog/11821.txt
+++ b/.changelog/11821.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+http: if a GET request has a non-empty body, log a warning that suggests a possible problem (parameters were meant for the query string, but accidentally placed in the body)
+```


jkirschner-hashicorp:
I still plan to merge _something_ to help here, but based on internal discussions, we might not want this to result in an error.

agent/http.go
```diff
@@ -543,6 +543,19 @@ func (s *HTTPHandlers) wrap(handler endpoint, methods []string) http.HandlerFunc
 		} else {
 			err = s.checkWriteAccess(req)
 
+			// Give the user a hint that they might be doing something wrong if their GET request
+			// has a non-empty body, which can easily happen using curl's --data-urlencode
+			// if specifying request type via "--request GET" rather than "--get".
+			// See https://github.com/hashicorp/consul/issues/11471.
```
kisunji, line 549: I think the comment makes the code look more severe/specific than it really is.
How's this generalized one?
```suggestion
			// Warn if the GET request has a non-empty body
```

agent/http.go
```diff
@@ -543,6 +543,19 @@ func (s *HTTPHandlers) wrap(handler endpoint, methods []string) http.HandlerFunc
...
+			if req.Method == http.MethodGet {
+				if req.ContentLength > 0 {
+					httpLogger.Warn("GET request has a non-empty body, "+
+						"check whether parameters were accidentally placed in the body rather than in the query string",
```
kisunji, line 553: I think the log is too long for most terminals; how's this for brevity?
```suggestion
					httpLogger.Warn("GET request has a non-empty body that will be ignored",
```

jkirschner-hashicorp: I like the brevity (and the addition of "that will be ignored").

However, I don't think it's obvious to a user that they probably meant to do something else (like put the parameters in the query string). And the message will already go beyond 80 or 120 characters. For example, this is 148 characters:
```
2022-02-22T12:26:32.237-0800 [WARN]  agent.http: GET request has a non-empty body that will be ignored: url=/v1/agent/services from=127.0.0.1:48852
```

Does the brevity actually help make a message display on a single terminal line? Or is there some other benefit (e.g., maybe there are other reasons why this might have happened other than accidentally putting stuff that belongs in the query string into the request body)?

kisunji: Maybe we can include "that will be ignored" and keep the prescriptive part of the log too

agent/http.go
jkirschner-hashicorp: Didn't do exactly what you suggested, but did modify the comment. Let me know if you agree.

agent/http.go
jkirschner-hashicorp: In that case, what do you think about the revised message?
```
GET request has a non-empty body that will be ignored; check whether parameters meant for the query string were accidentally placed in the body

jkirschner-hashicorp:
@kisunji : adjusted based on your comments. Let me know if this is good-to-go. Thanks!

kisunji (approved):
LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/607810.

PR #11836, opened 2021-12-14
Deprecate workspace_prefix for new workspaces.prefix

findkim:
image

Related https://github.com/hashicorp/consul-terraform-sync-enterprise/pull/243


lornasong (approved):
reads clearly to me

PR #11851, opened 2021-12-15
ui: Cope with case where we have 403'ed or zero'ed partitions

johncowen:
In certain Consul configurations our partitions and namespaces listing endpoints can return either a 403 or an `[]` when you don't have access to one or the other.

This can result in the side nav menus appearing with 'default' in them (which is arguably correct and arguably not correct depending on your opinion), but when you click on them they open but are empty and look like this until you login:

Screenshot 2021-12-15 at 13 48 56

This fix hides the entire menu (both of them) until you have actually logged in and get a non-403 or non-empty result back from the API.

P.S. There is a weird "only in test mode" flake here, and there's a chance I'll close this and try a different approach. I have an alternative idea forming. I'll let folks know once I'm clear myself, sorry for the noise!

### DIFF
--- a/ui/packages/consul-ui/app/adapters/nspace.js
+++ b/ui/packages/consul-ui/app/adapters/nspace.js
@@ -3,8 +3,8 @@ import { SLUG_KEY } from 'consul-ui/models/nspace';
 
 // namespaces aren't categorized by datacenter, therefore no dc
 export default class NspaceAdapter extends Adapter {
-  requestForQuery(request, { dc, partition, index, uri }) {
-    return request`
+  async requestForQuery(request, { dc, partition, index, uri, once }) {
+    const respond = await request`
       GET /v1/namespaces?${{ dc }}
       X-Request-ID: ${uri}
 
@@ -13,6 +13,10 @@ export default class NspaceAdapter extends Adapter {
         index,
       }}
     `;
+    if (once === 'once') {
+      await respond((headers, body) => delete headers['x-consul-index']);
+    }
+    return respond;
   }
 
   requestForQueryRecord(request, { dc, partition, index, id }) {
--- a/ui/packages/consul-ui/tests/integration/adapters/nspace-test.js
+++ b/ui/packages/consul-ui/tests/integration/adapters/nspace-test.js
@@ -6,14 +6,17 @@ module('Integration | Adapter | nspace', function(hooks) {
   setupTest(hooks);
   const id = 'slug';
   const dc = 'dc-1';
-  test('requestForQuery returns the correct url/method', function(assert) {
+  test('requestForQuery returns the correct url/method', async function(assert) {
     const adapter = this.owner.lookup('adapter:nspace');
     const client = this.owner.lookup('service:client/http');
-    const request = client.requestParams.bind(client);
+    const request = function() {
+      return () => client.requestParams.bind(client)(...arguments);
+    };
     const expected = `GET /v1/namespaces?dc=${dc}`;
-    const actual = adapter.requestForQuery(request, {
+    let actual = await adapter.requestForQuery(request, {
       dc: dc,
     });
+    actual = actual();
     assert.equal(`${actual.method} ${actual.url}`, expected);
   });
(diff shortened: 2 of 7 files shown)


ui/packages/consul-ui/tests/integration/adapters/nspace-test.js
```diff
@@ -6,14 +6,17 @@ module('Integration | Adapter | nspace', function(hooks) {
...
-    const request = client.requestParams.bind(client);
+    const request = function() {
+      return () => client.requestParams.bind(client)(...arguments);
+    };
     const expected = `GET /v1/namespaces?dc=${dc}`;
-    const actual = adapter.requestForQuery(request, {
+    let actual = await adapter.requestForQuery(request, {
       dc: dc,
     });
+    actual = actual();
```
johncowen, line 19: The work here was just because we had to change the shape of our test mocking as we started using `respond` in the adapters method that we added above in this git changeset. We used the exact same approach as we do in the equivalent partition test:

https://github.com/hashicorp/consul/blob/20cc028ac02af77ffff2bde19939c1c4fab37c1a/ui/packages/consul-ui/tests/integration/adapters/partition-test.js#L24-L32

Just to note, we don't test anything extra here in this PR, this was just to get the previous test passing again due to the shape change.

ui/packages/consul-ui/app/adapters/nspace.js
```diff
@@ -13,6 +13,10 @@ export default class NspaceAdapter extends Adapter {
         index,
       }}
     `;
+    if (once === 'once') {
+      await respond((headers, body) => delete headers['x-consul-index']);
+    }
+    return respond;
```
johncowen, line 19: If the HTTP headers don't have a `x-consul-index` on them then they don't block. That logic is done further into the codebase and has been around a long time and we use this approach elsewhere to disable blocking queries also. The `once === 'once'` is a little unfortunate here but it does the job with as little PR change as possible.

johncowen (commented):
Notes-da:

amyrlam (commented):
The code seems fine at first glance... As someone brand new to the codebase, could you provide some guidance on how I can see your screenshot locally? Couldn't repro your screenshot in local dev.

Also, the CircleCI `ember-test-ent` failure seemed legit?

Wasn't sure about `.github/workflows/changelog-checker.yml / changelog-check (pull_request)`

johncowen:
Hey 👋 

Thanks for taking a look! Yeah the code is fine but I'm not super keen on this to be honest.

From the description above 😄 :

> P.S. There is a weird "only in test mode" flake here, and there's a chance I'll close this and try a different approach. I have an alternative idea forming. I'll let folks know once I'm clear myself, sorry for the noise!

Theoretically it _should_ be fine and all makes sense code wise. But there is this weird pause once in while when running this in a test, and as we run the flaky test 4 times (for different variations of namespaces) it always fails as its more often than ~1 in 4.

Importantly, it only happens in test mode - so I was kind of in two minds as to whether I would have merged this even if I got an approval for 1.11 (there is something to be said for _very rarely and with extreme caution_ skipping a test in this circumstance) as its _seems_ to be a problem with our testing rather than our app, and just to say this is a very rare occurrence 😬 😄 and would need agreement with a teammate.

I had some more thoughts whilst this has been up though, so this is likely to be closed to be honest (I almost put it back in draft) - I'm just not entirely sure yet. A still vague alternative idea for a different approach means me working different in our app rather the sleuthing thorough Ember test code though so I'm thinking I'd rather try that alternative than figure out why this test is flakey.

For the changelog checker I sometimes leave until after review - how its been setup means it cancels our Vercel preview deployment if I do it too soon after the PR is made and I don't believe I can merge unless I then add one or add a skip check label.

Two important takeaways here:

1. This PR unfortunately is a bit of an outlier of how things usually go.
2. I believe we are gonna catch up offline about some of these ConsulRepo-isms so I can go through general process then Actually, there are some docs I could link to also 🤔  (offline)

amyrlam:
@johncowen Sounds good, seems like this one can wait til the Mon meeting?

PR #11882, opened 2021-12-17
Provide an 'escape-hatch' for website-checker workflow

eculver:
I'm proposing a change to the way the `website-checker` workflow behaves. It involves one change: if the PR has a `pr/no-docs-cherrypick`, skip the check. This would be similar in nature to our `pr/no-changelog` label for the `changelog-checker` workflow. 

This came up when I was doing manual oss->ent merges. Often, these changes will include docs changes that have already been cherry-picked. These manual merges are painful enough that it would be great to be able to exclude the `website-checker` workflow so that the PRs checks isn't marked as "failing" for something that should be resolved elsewhere.

PS - does the `website-checker` even make sense for Enterprise? I suppose we could just disable it entirely on that side if we're confident that we don't want this escape-hatch to exist. Food for thought.


.github/workflows/website-checker.yml
```diff
@@ -21,8 +21,10 @@ on:
 jobs:
   # checks that a 'type/docs-cherrypick' label is attached to PRs with website/ changes
   website-check:
-    # If there's a `type/docs-cherrypick` label we ignore this check
-    if: "!contains(github.event.pull_request.labels.*.name, 'type/docs-cherrypick')"
+    # If there's a `type/docs-cherrypick` label or an explicit `pr/no-docs-cherrypick` label, we ignore this check 
+    if: >-
```
eculver, line 25: In case anyone is wondering, I decided to use a '[Folded Block Scalar](https://yaml-multiline.info/)' here so that the block's newlines would be replaced with spaces rather than having to insert `\` as I'm not sure how GitHub expressions deal with those operators.

FFMMM (approved):
This makes sense to me! Let's get another set of eyes 👀 !

.github/workflows/website-checker.yml
dnephin: Yaml is awesome! I don't know why people hate on it

dnephin (approved):
I like it, thanks!

This check never "fails" right? So this will prevent the action from posting a comment on the PR if we add the label ahead of time, but otherwise it's a no-op ?

eculver:
> This check never "fails" right? So this will prevent the action from posting a comment on the PR if we add the label ahead of time, but otherwise it's a no-op ?

Take this with a grain of salt, but I had written a nice response for you last week thanking you for pointing out the fact that this check should really never fail, but it somehow never got submitted? I'm really not sure what happened, but I'll just try again:

You're totally correct: this check should never fail, so that begs the question: why bother creating an escape hatch for it? When I originally wrote this little PR I was dealing with an OSS->Ent merge and I had assumed that this check would fail when there were changes in the website directory, and I didn't think much of it when I  added the 'escape-hatch' and created this PR. However, when trying to respond to your comment @dnephin, I realized that you are totally right and that it should never fail. But, it definitely does fail ([example](https://github.com/hashicorp/consul-enterprise/runs/4780416056?check_suite_focus=true)). It fails because we are using `curl -f` which causes `curl` to exit non-zero whenever the server responds with an HTTP code that is treated as a "failure". I'm not sure why, but there was an issue with the token being used. I updated it last week when I was testing and it hasn't failed since. 

All of this is to say that an 'escape hatch' for a check that should never fail is somewhat silly. I'm going to go ahead and leave the escape hatch in place because I feel like all checks really _should_ have an escape hatch, but for this check, it will only prevent the reminder comment. 

And finally, I removed the `-f` from `curl` invocation since it's the true culprit of all this banter.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/561366.

PR #11893, opened 2021-12-20
Update exported-services.mdx

banks:
Correct a typo


banks:
Ha I realise this was not wrong and I was confused...

PR #11774, opened 2021-12-08
ui: Fixup names of Meta for instance search, also add Node

johncowen:
When searching through our Service Instance list, you can choose to limit the search to Meta data, but in Service Instance listing we have Service Meta _and_ Node Meta available to us. Previous to this PR both these options would read 'Meta'.

This PR changes these texts to read Service Meta and Node Meta.

We also took the opportunity to add Node Name here to the list of options you can limit the search to.

PR #11899, opened 2021-12-21
ui: [Backport 1.9.x] Fixup names of Meta for instance search, also add Node

johncowen:
Manual backport of https://github.com/hashicorp/consul/pull/11774

In 1.10.x we moved our translations file to use a split up files instead of one big one, hence the need for a manual intervention. This manual backport adds the necessary translation strings to the "one big file" instead of the split up ones.

No changelog label here as the changelog checker doesn't work if the base branch isn't main.


amyrlam (approved):
👍

PR #11900, opened 2021-12-21
ui: Configure routes in route config rather than classes

johncowen:
This PR moves the remaining `queryParams`, `templateName` and `redirects` from Routes to our routing configuration (we are still ignoring our ACLs section/routes for these kinds of upgrades for the moment).

There should be no user facing change here, this is just about putting configuration into central configuration files rather than dotted across lots of different javascript classes.

Note: Currently our largest 'configuration file' is the `router.js` file, but this will then be split out into separate sub packages similar to our `consul-partitions` and `consul-nspaces` sub packages to allow for runtime feature flagged route configuration.

There is a small amount of code change here (in the consul-ui/routing/route.js file) to allow us to pick up queryParams in a more generalized manner and to allow our redirect configuration to be relative paths instead of just a name of a subroute.

Lastly, no backport as we are targetting 1.12 here.


ui/packages/consul-ui/app/routing/route.js
```diff
@@ -16,36 +16,41 @@ export default class BaseRoute extends Route {
...
+            if (item === '..') {
+              prev.pop();
+            } else if (item !== '' || i === items.length - 1) {
+              prev.push(item);
+            }
+          }
+          return prev;
+        }, this.routeName.split('.'))
+        .join('.');
+      console.log(to);
```
amyrlam, line 48: remove console.log or is it intentional? 

could add https://eslint.org/docs/rules/no-console separately, just a suggestion

johncowen: :doh: Wow, thanks for spotting! Definitely not intentional, will fix up now.

Weirdly, there should be a eslint rule preventing this, will have a look at that separately.

johncowen: K thanks again @amyrlam this should be good to go now. I did a quick check on why console logs aren't being caught and I've figured why. There'll be a PR coming up with that soonish 🙇‍♂️

amyrlam (approved):
👍

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/540191.

Issue #11873, opened 2021-12-16
Sidecar proxy checks not showing up in UI until after a refresh


lkysow:
#### Overview of the Issue

Consul 1.11.1 (does not occur with 1.10.4)

I have a service with two instances, each instance has a sidecar proxy. When I view one instance I see the sidecar proxy checks as expected. When I then navigate to the other instance, the proxy checks aren't shown. Then when I refresh, the proxy checks appear. If I navigate back to the first instance, now its proxy checks aren't shown. See video:

https://user-images.githubusercontent.com/1034429/146455699-f0500a22-493a-4120-a705-9ee2a5ace63c.mp4

Service files for reproduction:

```hcl
service {
  name = "backend"
  port = 7000
  connect {
    sidecar_service {
      proxy {
      }
    }
}
```

```hcl
service {
  name = "backend"
  id = "backend-v2"
  port = 7001
  connect {
    sidecar_service {
      proxy {
      }
    }
  }
}
```

johncowen:
Thanks for spotting and reporting this @lkysow 🙏 theres a fix incoming for this https://github.com/hashicorp/consul/pull/11903

PR #11903, opened 2021-12-22
ui: Ensure service instance data does not get re-written on blocking refresh

johncowen:
Fixes https://github.com/hashicorp/consul/issues/11873

This PR is quite in-depth and there are a couple of things to understand here.

1. Routlet/Outlets/Routes - There is an RFC detailing this, ~I'll see if I can find the title here so folks can take a look if needed~ _CSL-164: Sharing data across Routes and Components in the Consul UI_. (Also _CSL-092 A componentized approach to data_, whilst not entirely necessary may also help here)
2. (Previous to this PR) Our approach to creating special munged together ProxyServiceInstances, which where a combination of the response from 2 endpoints - How we did this was the root of the issue here.

I've added quite a few inline comments and the link to the issue explains what the user facing problem/consequence was. But let me know if I can add any more color/help here. Also commit by commit review might help folks here.

This one is targeting our next 1.11.x point release, and needs review/approval/merge for then.

### DIFF
--- a/ui/packages/consul-ui/app/services/repository/service-instance.js
+++ b/ui/packages/consul-ui/app/services/repository/service-instance.js
@@ -1,12 +1,10 @@
 import RepositoryService from 'consul-ui/services/repository';
-import { inject as service } from '@ember/service';
 import { set } from '@ember/object';
 import { ACCESS_READ } from 'consul-ui/abilities/base';
 import dataSource from 'consul-ui/decorators/data-source';
 
 const modelName = 'service-instance';
 export default class ServiceInstanceService extends RepositoryService {
-  @service('repository/proxy') proxyRepo;
   getModelName() {
     return modelName;
   }
@@ -34,44 +32,6 @@ export default class ServiceInstanceService extends RepositoryService {
 
   @dataSource('/:partition/:ns/:dc/service-instance/:serviceId/:node/:id')
   async findBySlug(params, configuration = {}) {
-    if (typeof configuration.cursor !== 'undefined') {
-      params.index = configuration.cursor;
-      params.uri = configuration.uri;
-    }
-    return this.authorizeBySlug(
-      async () => this.store.queryRecord(this.getModelName(), params),
-      ACCESS_READ,
-      params
-    );
-  }
-
-  @dataSource('/:partition/:ns/:dc/proxy-service-instance/:serviceId/:node/:id')
-  async findProxyBySlug(params, configuration = {}) {
-    const instance = await this.findBySlug(...arguments);
-    let proxy = this.store.peekRecord('proxy', instance.uid);
-    // Currently, we call the proxy endpoint before this endpoint
-    // therefore proxy is never undefined. If we ever call this endpoint
-    // first we'll need to do something like the following
-    // if(typeof proxy === 'undefined') {
-    //   await proxyRepo.create({})
-    // }
-
-    // Copy over all the things to the ProxyServiceInstance
-    ['Service', 'Node', 'meta'].forEach(prop => {
-      set(proxy, prop, instance[prop]);
-    });
-    ['Checks'].forEach(prop => {
-      // completely wipe out any previous values so we don't accumulate things
-      // eternally
-      proxy.set(prop, []);
-      instance[prop].forEach(item => {
-        if (typeof item !== 'undefined') {
-          proxy[prop].addFragment(item.copy());
-        }
-      });
-    });
-    // delete the ServiceInstance record as we now have a ProxyServiceInstance
-    instance.unloadRecord();
-    return proxy;
+    return super.findBySlug(...arguments);
   }
 }
--- a/ui/packages/consul-ui/app/services/routlet.js
+++ b/ui/packages/consul-ui/app/services/routlet.js
@@ -70,21 +70,11 @@ export default class RoutletService extends Service {
     return key;
   }
 
-  addOutlet(name, outlet) {
-    outlets.set(name, outlet);
-  }
-
-  removeOutlet(name) {
-    outlets.delete(name);
-  }
-
-  // modelFor gets the model for Outlet specified by `name`, not the Route
-  modelFor(name) {
-    const outlet = outlets.get(name);
-    if (typeof outlet !== 'undefined') {
-      return outlet.model || {};
-    }
-    return {};
+  outletFor(routeName) {
+    const keys = [...outlets.keys()];
+    const pos = keys.indexOf(routeName);
+    const key = pos + 1;
+    return outlets.get(keys[key]);
   }
 
   paramsFor(name) {
@@ -128,20 +118,43 @@ export default class RoutletService extends Service {
     };
   }
 
-  addRoute(name, route) {
-    const keys = [...outlets.keys()];
-    const pos = keys.indexOf(name);
-    const key = pos + 1;
-    const outlet = outlets.get(keys[key]);
+
+  // modelFor gets the model for Outlet specified by `name`, not the Route
+  modelFor(name) {
+    const outlet = outlets.get(name);
     if (typeof outlet !== 'undefined') {
-      route._model = outlet.model;
+      return outlet.model;
+    }
+  }
+
+  addRoute(name, route) {
+    const outlet = this.outletFor(name);
+    if (typeof outlet !== 'undefined') {
       outlet.route = route;
       // TODO: Try to avoid the double computation bug
       schedule('afterRender', () => {
-        outlet.routeName = route.args.name;
+        outlet.routeName = name;
       });
     }
   }
 
-  removeRoute(name, route) {}
+  removeRoute(name, route) {
+    const outlet = this.outletFor(name);
+    route._model = undefined;
+    if (typeof outlet !== 'undefined') {
+      schedule('afterRender', () => {
+        outlet.route = undefined;
+      });
+    }
+  }
+
+  addOutlet(name, outlet) {
+    outlets.set(name, outlet);
+  }
+
+  removeOutlet(name) {
+    schedule('afterRender', () => {
+      outlets.delete(name);
+    });
+  }
 }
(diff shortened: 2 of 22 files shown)


ui/packages/consul-ui/app/services/repository/service-instance.js
```diff
@@ -34,44 +32,6 @@ export default class ServiceInstanceService extends RepositoryService {
...
-      proxy.set(prop, []);
-      instance[prop].forEach(item => {
-        if (typeof item !== 'undefined') {
-          proxy[prop].addFragment(item.copy());
-        }
-      });
-    });
-    // delete the ServiceInstance record as we now have a ProxyServiceInstance
-    instance.unloadRecord();
-    return proxy;
```
johncowen, line 75: This turned out to be the problematic code.

When we request details on a service we use a proxy endpoint to check if it has a proxy service, and if it does we use the previous same service endpoint to get the details for the proxy. But! The extra bit we did here was to munge together the details we got from the proxy endpoint and the details we get from the service endpoint together to make a special type of 'ProxyServiceInstance' - a special type of ServiceInstance.

Turns out, doing this was providing very little benefit to us. Moreover previously we could almost guarantee that the proxy request would always be called before the service request (almost in the fact that there was probably a very very edge case bug here in doing this anyway that has just never surfaced as its unlikely to happen).

Now its pretty common that the ServiceProxyInstance has been created and munged together, and then we check the details for the Proxy again and the `instance` in the above code gets 'refreshed/updated' again, but without then re-requesting and re-munging the extra details onto the instance, meaning a bunch of properties that we weere expecting to be set have been deleted/set back to undefined.

All in all its a little complex to explain well so would be a good one to chat through if you want the overall picture a little more clearly.

Alternatively - we just deleted the entire thing and used a different approach that doesn't have these issues, by using less EmberData and Fragments.

ui/packages/consul-ui/app/services/routlet.js
```diff
@@ -70,21 +70,11 @@ export default class RoutletService extends Service {
...
-    const outlet = outlets.get(name);
-    if (typeof outlet !== 'undefined') {
-      return outlet.model || {};
-    }
-    return {};
+  outletFor(routeName) {
+    const keys = [...outlets.keys()];
+    const pos = keys.indexOf(routeName);
+    const key = pos + 1;
+    return outlets.get(keys[key]);
```
johncowen, line 77: Here's the new `outletFor` code, which was extracted from the old `addRoute` code which you can see in the red side of the changeset. This approach is what we should be using (and we are now) for getting a specific outlet when you only have access to a routeName.

ui/packages/consul-ui/app/services/routlet.js
```diff
@@ -128,20 +118,43 @@ export default class RoutletService extends Service {
...
+    if (typeof outlet !== 'undefined') {
+      return outlet.model;
+    }
+  }
+
   addRoute(name, route) {
-    const keys = [...outlets.keys()];
-    const pos = keys.indexOf(name);
-    const key = pos + 1;
-    const outlet = outlets.get(keys[key]);
```
johncowen, line 135: Heres the same code for `outletFor` which we moved into its own method and reuse both here and in other places.

ui/packages/consul-ui/app/templates/dc/services/instance.hbs
```diff
@@ -101,8 +101,13 @@ as |item|}}
...
           {{! a HTTP request/long poll here }}
           {{#if meta.data.ServiceID}}
+          {{! if we have a proxy then get the additional instance information }}
+          {{! for the proxy itself so if the service is called `backend` }}
+          {{! its likely to have a proxy service called `backend-sidecar-proxy` }}
+          {{! and this second request get the info for that instance and saves }}
+          {{! it into the `proxy` variable }}
             <DataSource
-              @src={{uri '/${partition}/${nspace}/${dc}/proxy-service-instance/${id}/${node}/${name}'
+              @src={{uri '/${partition}/${nspace}/${dc}/service-instance/${id}/${node}/${name}'
```
johncowen, line 110: We removed `proxy-` here as now we don't need to do the ServiceInstance/ProxyInstance munging we can just as for the service information of the proxy using the same endpoint as retrieving the service information for the service itself.

ui/packages/consul-ui/app/templates/dc/services/instance.hbs
```diff
@@ -129,6 +135,8 @@ as |item|}}
             </h1>
             <Consul::ExternalSource @item={{item}} @withInfo={{true}} />
             <Consul::Kind @item={{item}} @withInfo={{true}} />
+            {{! TODO: Looks like we can get this straight from item.Proxy.Mode }}
+            {{! the less we need `proxy` the better }}
             {{#if (eq proxy.ServiceProxy.Mode 'transparent')}}
               <Consul::TransparentProxy />
             {{/if}}
```
johncowen, line 142: I noticed here that we also have access to `Mode` on the `item` (the service instance) as well as the `proxy`. The less we use proxy the better. Actually, glad I self reviewed this I need to look at this a little more 🤔  brb

ui/packages/consul-ui/app/templates/dc/services/instance.hbs
```diff
@@ -149,23 +157,17 @@ as |item|}}
...
+              compact
+                  (array
+                                          (hash label="Health Checks" href=(href-to "dc.services.instance.healthchecks") selected=(is-href "dc.services.instance.healthchecks"))
+(if (eq item.Service.Kind 'mesh-gateway') (hash label="Addresses" href=(href-to "dc.services.instance.addresses") selected=(is-href "dc.services.instance.addresses")))
+                                (if proxy (hash label="Upstreams" href=(href-to "dc.services.instance.upstreams") selected=(is-href "dc.services.instance.upstreams")))
+                                (if proxy (hash label="Exposed Paths" href=(href-to "dc.services.instance.exposedpaths") selected=(is-href "dc.services.instance.exposedpaths")))
+                                          (hash label="Tags & Meta" href=(href-to "dc.services.instance.metadata") selected=(is-href "dc.services.instance.metadata"))
+                  )
+              }}
+            />
```
johncowen, line 170: Just formatting

ui/packages/consul-ui/app/templates/dc/services/instance/healthchecks.hbs
```diff
@@ -27,7 +27,7 @@ as |route|>
       )
     )
 
-    route.model.item.MeshChecks
+    (merge-checks (array route.model.item.Checks route.model.proxy.Checks) route.model.proxy.ServiceProxy.Expose.Checks)
```
johncowen, line 30: So where we merge all the checks (those from the ServiceInstance and the Proxy) together before iterating through them. This uses the exact same code as previously, only we do it in the template instead of in the model which then avoids all manner of problems.

ui/packages/consul-ui/mock-api/v1/health/service/backend-sidecar-proxy
```diff
@@ -0,0 +1,204 @@
...
+        "Interval": "",
+        "Timeout": "",
+        "ExposedPort": 0,
+        "Definition": {},
+        "CreateIndex": 21,
+        "ModifyIndex": 21
+      }
+    ]
+  }
+]
```
johncowen, line 204: These are all real responses from a working Consul to give us a more realistic UI view, it also made it easier to reproduce and manually debug/test what was happening. We generally just use faker.js to give us test data, but we've done this sort of 'non-random fixtures' for the discovery-chain work to.

The consequences of this are that if you view the ServiceInstance page specifically for a service called `backend` only (`http://localhost:4200/ui/dc1/services/backend/instances`) - then you will get this exact data, not the randomized stuff.

ui/packages/consul-ui/tests/integration/services/routlet-test.js
```diff
@@ -0,0 +1,32 @@
...
+  assert.equal(actual.name, expected);
+
+  actual = routlet.outletFor('dc');
+  expected = 'application';
+  assert.equal(actual.name, expected);
+
+  actual = routlet.outletFor('application');
+  expected = undefined;
+  assert.equal(actual, expected);
+});
```
johncowen, line 32: Started out with some basic tests for `outletFor`, this is likely to be added to moving forwards, and potentially the interface may change a little (I saw that instead of passing `(name, route/outlet)` we could just pass `(route/outlet)` and get the name straight off the route/outlet `name` property. I'd rather do that in a refactor PR than a bugfix one though - so that may happen at some point in the future. For now this is just a bit of reassurance testing so I know it does what I need correctly.

johncowen (commented):
Notes:

ui/packages/consul-ui/app/templates/dc/services/instance.hbs
johncowen: After I looked more at this I realized that I needed to also swap our `proxy` for `meta` here and in a few other places. I've added another 4 commits on the end here to cover this (from `Make TProxy...` cc698d304fe2b9389f8264066084adaf3469bae1 to `Document CONSUL_...` 38ba3c0224a4c002256da128dec9a46c40ec5ba6)

ui/packages/consul-ui/app/templates/dc/services/instance/addresses.hbs
```diff
@@ -2,14 +2,14 @@
...
+  (entries route.model.item.Service.TaggedAddresses)
+as |items|}}
   <div class="tab-section">
-    {{#if item.Service.TaggedAddresses }}
+    {{#if (gt items.length 0)}}
       <TabularCollection
         data-test-addresses
         class="consul-tagged-addresses"
-        @items={{entries item.Service.TaggedAddresses}} as |taggedAddress index|
+        @items={{items}} as |taggedAddress index|
```
johncowen, line 12: I needed to check this file to see if we needed to move to use `meta` here, so I fixed up some formatting while I was here

natmegs (approved):
Pair reviewed

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/543358.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 6c240fbf2d1f11c5a280b29656d21303b94a0a1b onto `release/1.11.x` succeeded!

PR #11910, opened 2021-12-22
ca: add support for an external trusted CA

dnephin:
Related to #11598

See our [developer CA docs](https://github.com/hashicorp/consul/tree/main/docs/service-mesh/ca#ca-and-certificate-relationship) for some background about this part of the code.

This PR adds support for using an external root CA with the Vault CA provider. A user configures the `root_pki_path` in Vault with a PEM that has the intermediate cert used as the Primary CA as the first cert, and the external root CA as the last cert.

TODO:
* [x] more test cases for `newCARoot`
* [x] test intermediate renew
* [x] test with secondary DC
* [x] test root rotation from/to a provider with an external trusted CA
* [x] test with an enovy proxy and auto-encrypt cert for the xDS API
* [x]  document how to use an external trusted CA with the vault provider

Follow up items:
* [ ] rename `root_pki_path` to something more appropriate


agent/connect/ca/provider_vault.go
```diff
@@ -529,12 +529,7 @@ func (v *VaultProvider) Sign(csr *x509.CertificateRequest) (string, error) {
 	if !ok {
 		return "", fmt.Errorf("certificate was not a string")
 	}
-	ca, ok := response.Data["issuing_ca"].(string)
-	if !ok {
-		return "", fmt.Errorf("issuing_ca was not a string")
-	}
-
-	return EnsureTrailingNewline(cert) + EnsureTrailingNewline(ca), nil
+	return EnsureTrailingNewline(cert), nil
```
dnephin, line 563: There is more context in the commit message, but I believe this change is correct because it aligns the vault provider with the other providers, and because we append the full chain to this cert in the Sign RPC endpoint. So previous it would have been adding the same CA cert twice. 

That was fine before when it was just a single root cert, but it produces an unexpected chain when we have multiple certs in the chain. The tests added in subsequent commits should exercise this code extensively in both primary and secondary DC.

dnephin (commented):
There's a lot here, and I'm struggling to inline comment it properly. Please ask lots of questions and I'll fill in the missing details.

agent/connect/ca/provider_vault.go
```diff
@@ -288,12 +288,10 @@ func (v *VaultProvider) GenerateRoot() (RootResult, error) {
 		if err != nil {
 			return RootResult{}, err
 		}
-
-		// retrieve the newly generated cert so that we can return it
-		// TODO: is this already available from the Local().Write() above?
```
rboyer, line 293: I guess it was.

agent/connect/ca/provider_vault.go
```diff
@@ -577,6 +603,7 @@ func (v *VaultProvider) SignIntermediate(csr *x509.CertificateRequest) (string,
 // CrossSignCA takes a CA certificate and cross-signs it to form a trust chain
 // back to our active root.
 func (v *VaultProvider) CrossSignCA(cert *x509.Certificate) (string, error) {
+	// TODO: is this necessary? Doesn't vault check this for us?
```
rboyer, line 606: Can you elaborate?

agent/connect/parsing.go
```diff
@@ -90,21 +105,10 @@ func parseCerts(pemValue string) ([]*x509.Certificate, error) {
 	return out, nil
 }
 
-// CalculateCertFingerprint parses the x509 certificate from a PEM-encoded value
-// and calculates the SHA-1 fingerprint.
-func CalculateCertFingerprint(pemValue string) (string, error) {
```
rboyer, line 95: Possibly dumb question, but did you triple check that the right data is flowing into the newly refactored method for back compat?

agent/consul/leader_connect_ca.go
```diff
@@ -487,12 +483,12 @@ func (c *CAManager) primaryInitialize(provider ca.Provider, conf *structs.CAConf
...
 	}
 
-	rootCA, err := parseCARoot(root.PEM, conf.Provider, conf.ClusterID)
+	rootCA, err := newCARoot(root.PEM, conf.Provider, conf.ClusterID)
 	if err != nil {
 		return err
 	}
 
-	// Also create the intermediate CA, which is the one that actually signs leaf certs
+	// TODO: delete this
```
rboyer, line 491: In this PR?

agent/consul/leader_connect_ca.go
```diff
@@ -980,6 +976,7 @@ func (c *CAManager) primaryUpdateRootCA(newProvider ca.Provider, args *structs.C
 		}
 	}
 
+	// TODO: delete this
 	intermediate, err := newProvider.GenerateIntermediate()
```
rboyer, line 980: In this PR?

agent/connect/ca/provider.go
```diff
@@ -193,7 +194,12 @@ type SecondaryProvider interface {
 //
 // TODO: rename this struct
 type RootResult struct {
-	// PEM encoded certificate that will be used as the primary CA.
+	// PEM encoded bundle of CA certificates. The first certificate must be the
+	// primary CA used to sign intermediates for secondary datacenters, and the
+	// last certificate must be the trusted CA.
```
kyhavlov, line 198: If this can return multiple certificates in a chain now, could that cause any problems for downstream things that call the `/ca/roots` endpoint?

agent/consul/leader_connect_ca_test.go
```diff
@@ -704,10 +708,208 @@ func getLeafCert(t *testing.T, codec rpc.ClientCodec, trustDomain string, dc str
...
+		require.NoError(t, err)
+		require.Len(t, roots.Roots, 2)
+		active := roots.Active()
+		require.Len(t, active.IntermediateCerts, 1)
+
+		leafPEM := getLeafCert(t, codec, roots.TrustDomain, "dc1")
+		verifyLeafCert(t, roots.Active(), leafPEM)
+
+		// original certs from old root cert should still verify
+		verifyLeafCertWithRoots(t, roots, origLeaf)
```
kyhavlov, line 852: It might be worth also verifying that a leaf cert signed in the secondary still works after this root rotation (since that should also cause the secondary signing cert to be refreshed). We could also verify it below after the rotation to the external root.

website/content/docs/connect/ca/vault.mdx
```diff
@@ -116,16 +116,25 @@ The configuration options are listed below.
...
   When WAN Federation is enabled, each secondary datacenter must use the same Vault cluster and share the same `root_pki_path`
-  with the primary datacenter. 
+  with the primary datacenter.
+
+  To use an intermediate certificate as the primary CA in Consul initialize the
+  `RootPKIPath` in Vault with a PEM bundle. The first certificate in the bundle
+  must be the intermediate certificate that Consul will use as the primary CA.
+  The last certificate in the bundle must be a root certificate. The bundle
+  must contain a valid chain, where each certificate is followed by the certificate
+  that authorized it.
```
kyhavlov, line 137: It might be a good idea to be more specific in calling out which endpoint(s) to use to set up the pki backend like this. Probably better to be overly detailed when we expect the backend to be set up a certain way.

kyhavlov (commented):
This looks great - had a few questions/suggestions but nothing blocking.

agent/consul/leader_connect_ca_test.go
dnephin: Ya, good call! I think that should be possible as long as we use the roots from the secondary.

These verify methods attempt to verify using the CA chain from two difference places: one using the intermediates from the `/ca/roots` call, and the other using intermediates attached to the leaf cert.

Using the intermediates on the leaf always works, but using the intermediates on the `CARoot` only works if the `CARoot` came from the secondary, because the intermediate certs in the secondary are never sent to the primary.  

But I can extract a new verify helper that only does one of these checks to do that assertion.

website/content/docs/connect/ca/vault.mdx
dnephin: Ya, that's a good idea. It's a few steps, but I'll try and make it fit into this doc.  If it doesn't fit we might need to document it in a section below and link to it.

dnephin (commented):
Thanks for the review!

I'll push a commit with the improve TODO messages now, and work on the test and docs improvements.

agent/connect/ca/provider.go
rboyer: Along a similar line of reasoning, if someone doesn't switch to an external trusted CA immediately, would upgrading to a patched version of Consul ever _start_ returning multiple certs in a chain? If so that may break during the rolling upgrade of consul from a non-patched to a patched version.

dnephin: Right, it won't change in that case. The only way this starts returning multiple certs is when the user updates the config to do so. Upgrading Consul continues to return a single cert in this PEM.  

Existing versions of Consul also return multiple certs here if the `RootCert` is configured that way. I think that might already work with the built-in provider. That's why I think this is mostly just a docs change. It could have already been returning multiple certs, it just wasn't obvious before.

agent/consul/leader_connect_ca_test.go
dnephin: Added these extra checks in the most recent commit. I went with the approach of using the roots from the secondary. I ran the test ~100 times and haven't seen any flakes.

website/content/docs/connect/ca/vault.mdx
dnephin: I opened #12393 for these docs, so that it's easier to focus on the docs changes in isolation.

rboyer (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/591446.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 771df290d7efae6bae62f1f624206a71d122fd17 onto `release/1.11.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/591446)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 771df290d7efae6bae62f1f624206a71d122fd17 onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/591446)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 771df290d7efae6bae62f1f624206a71d122fd17 onto `release/1.9.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/591446)

Danmor33 (commented):
T

PR #9689, opened 2021-02-02
Avoid potential proxycfg/xDS deadlock using non-blocking send

freddygv:
Deadlock scenario:

    1. Due to scheduling, the state runner sends one snapshot into
    snapCh and then attempts to send a second. The first send succeeds
    because the channel is buffered, but the second blocks.

    2. Separately, Manager.Watch is called by the xDS server after
    getting a discovery request from Envoy. This function acquires the
    manager lock and then blocks on receiving the CurrentSnapshot from
    the state runner.

    3. Separately, there is a Manager goroutine that reads the snapshots
    from the channel in step 1. These reads are done to notify proxy
    watchers, but they require holding the manager lock. This goroutine
    goes to acquire that lock, but can't because it is held by step 2.

Now, the goroutine from step 3 is waiting on the one from step 2 to
release the lock. The goroutine from step 2 won't release the lock until
the goroutine in step 1 advances. But the goroutine in step 1 is waiting
for the one in step 3 to read from the channel it is blocking on. Deadlock.

By making the send in step 1 non-blocking then we can proceed. The coalesce
timer will be reset and a new snapshot (if valid) will be delivered after the
timer elapses or when one is requested by xDS.

**Impact:** This deadlock can block the delivery of proxy configuration to Envoy.

------

Also added some trace logging that would have helped while chasing this down.

PR #11924, opened 2021-12-29
reset `coalesceTimer` to nil as soon as the event is consumed

### LINKED ISSUE #11923: XDS snapshot updates are not correctly delivered after a first retry when certain conditions are met.
#### Overview of the Issue

This is sequel to #9689.

When an XDS snapshot is received and we set a timer to update after a `coalesceTimer` in [here](http://github.com/hashicorp/consul-enterprise/blob/cbb8cb741a8f299c3fbe68c84b9092cf7a17e0a8/agent/proxycfg/state.go#L721-L721), if the 1st condition from #9689 is met and we end up in the `default` case of the select in [here](http://github.com/hashicorp/consul-enterprise/blob/cbb8cb741a8f299c3fbe68c84b9092cf7a17e0a8/agent/proxycfg/state.go#L679-L679) the `coalesceTimer` is never reset  and snapshot updates end up never triggered again until the service is recreated because `coalesceTimer` will never be nil and would prevent `sendCh` from ever be signaled.

This issue would happen more frequently in overloaded systems, with a high number of services and a high number of XDS updates which will more likely trigger the 1st condition from #9689.

dhiaayachi:
This is fix #11923

### DIFF
--- a/agent/proxycfg/manager_test.go
+++ b/agent/proxycfg/manager_test.go
@@ -660,7 +660,7 @@ func TestManager_SyncState_No_Notify(t *testing.T) {
 	services := m.State.AllServices()
 	var notifyCH chan cache.UpdateEvent
 
-	for sid, _ := range services {
+	for sid := range services {
 		notifyCH = m.proxies[sid].ch
 	}
 
--- a/agent/proxycfg/manager_test.go
+++ b/agent/proxycfg/manager_test.go
@@ -598,7 +598,150 @@ func TestManager_SyncState_DefaultToken(t *testing.T) {
 
 	err = state.AddServiceWithChecks(srv, nil, "")
 	require.NoError(t, err)
-	m.syncState()
+	m.syncState(m.notifyBroadcast)
 
 	require.Equal(t, "default-token", m.proxies[srv.CompoundServiceID()].serviceInstance.token)
+}
+
+func TestManager_SyncState_No_Notify(t *testing.T) {
+	types := NewTestCacheTypes(t)
+	c := TestCacheWithTypes(t, types)
+	logger := testutil.Logger(t)
+	tokens := new(token.Store)
+	tokens.UpdateUserToken("default-token", token.TokenSourceConfig)
+
+	state := local.NewState(local.Config{}, logger, tokens)
+	state.TriggerSyncChanges = func() {}
+
+	m, err := NewManager(ManagerConfig{
+		Cache:  c,
+		Health: &health.Client{Cache: c, CacheName: cachetype.HealthServicesName},
+		State:  state,
+		Tokens: tokens,
+		Source: &structs.QuerySource{Datacenter: "dc1"},
+		Logger: logger,
+	})
+	require.NoError(t, err)
+	defer m.Close()
+
+	srv := &structs.NodeService{
+		Kind:    structs.ServiceKindConnectProxy,
+		ID:      "web-sidecar-proxy",
+		Service: "web-sidecar-proxy",
+		Port:    9999,
+		Meta:    map[string]string{},
+		Proxy: structs.ConnectProxyConfig{
+			DestinationServiceID:   "web",
+			DestinationServiceName: "web",
+			LocalServiceAddress:    "127.0.0.1",
+			LocalServicePort:       8080,
+			Config: map[string]interface{}{
+				"foo": "bar",
+			},
+		},
+	}
+
+	err = state.AddServiceWithChecks(srv, nil, "")
+	require.NoError(t, err)
+
+	readEvent := make(chan bool, 1)
+	snapSent := make(chan bool, 1)
+
+	m.syncState(func(ch <-chan ConfigSnapshot) {
+		for {
+			<-readEvent
+			snap := <-ch
+			m.notify(&snap)
+			snapSent <- true
+		}
+	})
+
+	// Get the relevant notification Channel, should only have 1
+	services := m.State.AllServices()
+	var notifyCH chan cache.UpdateEvent
+
+	for sid, _ := range services {
+		notifyCH = m.proxies[sid].ch
+	}
+
+	// update the leaf certs
+	roots, issuedCert := TestCerts(t)
+	notifyCH <- cache.UpdateEvent{
+		CorrelationID: leafWatchID,
+		Result:        issuedCert,
+		Err:           nil,
+	}
+	// at this point the snapshot should not be valid and not be sent
+	after := time.After(200 * time.Millisecond)
+	select {
+	case <-snapSent:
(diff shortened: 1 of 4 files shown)


agent/proxycfg/manager_test.go
```diff
@@ -598,7 +598,150 @@ func TestManager_SyncState_DefaultToken(t *testing.T) {
...
+		}
+	})
+
+	// Get the relevant notification Channel, should only have 1
+	services := m.State.AllServices()
+	var notifyCH chan cache.UpdateEvent
+
+	for sid := range services {
+		notifyCH = m.proxies[sid].ch
+	}
```
freddygv, line 665: I believe this can be simplified:
```suggestion
	notifyCH := m.proxies[srv.CompoundServiceID()].ch
```

agent/proxycfg/manager_test.go
```diff
@@ -598,7 +598,150 @@ func TestManager_SyncState_DefaultToken(t *testing.T) {
...
+	select {
+	case <-snapSent:
+		t.Fatal("snap should not be valid")
+	case <-after:
+
+	}
+
+	//prepare to read a snapshot update as the next update should make the snapshot valid
+	readEvent <- true
+	// update the intensions
```
freddygv, line 701: typo:
```suggestion

	// update the intentions
```

agent/proxycfg/manager_test.go
```diff
@@ -598,7 +598,150 @@ func TestManager_SyncState_DefaultToken(t *testing.T) {
...
+	// at this point the snapshot should not be valid and not be sent
+	after = time.After(200 * time.Millisecond)
+	select {
+	case <-snapSent:
+		t.Fatal("snap should not be valid")
+	case <-after:
+
+	}
+
+	//prepare to read a snapshot update as the next update should make the snapshot valid
```
freddygv, line 699:
```suggestion
	// prepare to read a snapshot update as the next update should make the snapshot valid
```

agent/proxycfg/manager_test.go
```diff
@@ -598,7 +598,150 @@ func TestManager_SyncState_DefaultToken(t *testing.T) {
...
+	select {
+	case <-snapSent:
+	case <-after:
+		t.Fatal("snap should be valid")
+
+	}
+
+	// send two snapshots back to back without reading them to overflow the snapshot channel and get to the default use case
+	for i := 0; i < 2; i++ {
+		time.Sleep(250 * time.Millisecond)
```
freddygv, line 719: What's this sleep used for? Could it be removed to reduce the running time for the test?

freddygv (approved):
Good find! I'm surprised this didn't come up during the scale tests where the issue from #9689 came up.

agent/proxycfg/state.go
```diff
@@ -294,6 +294,8 @@ func (s *state) run(ctx context.Context, snap *ConfigSnapshot) {
 			}
 
 		case <-sendCh:
+			// Allow the next change to trigger a send
+			coalesceTimer = nil
```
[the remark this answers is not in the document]
banks: I'm not sure I understand what this timer is doing any more. The original intent was to prevent updates to multiple blocking queries that return at the same time from causing multiple reconfigurations of Envoy in quick succession.

For example on startup, a proxy might start watching 100 upstream services and config entries, without coalescing here, we would end up delivering O(100) entire snapshot reconfigurations to Envoy in the space of a few hundred milliseconds as all the blocking queries return their initial results.

It's not clear to me if this is actually doing that any more or if immediately resetting it here is causing it to effectively be removed? I've not traced this code for long enough to fully remember how it works though.

The other unknown for me is whether this coalescing is even needed/useful anymore since we implemented Delta XDS? My understanding is that now, even if we deliver multiple snapshots, the delta part of delta XDS will ensure we only send the changes in each one down so from Envoy's perspective it might not make any difference. It still could save some CPU cycles on the agent that is doing the delivering though which as we move towards Servers doing that could be significant. It's not clear that they will even use this package as it is though so I perhaps wouldn't optimize too much for that.

I'd be interested to know at any rate whether this change has a material impact on the amount of reconfigurations sent to envoy when there are a lot of upstreams, and if not, whether we should just remove all the coallesce code entirely and simplify this loop?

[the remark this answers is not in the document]
banks: Nevermind. I get it now. I missed the main place the timer is reset outside of the Select in the for loop way below (not in the default branch). Originally before the last deadlock fix it was always reset in this branch anyway: https://github.com/hashicorp/consul/blame/cf9a14ab6ad25e8932eb2b07616c0c33ddc54b12/agent/proxycfg/state.go#L614-L631

I think Freddy might be right that the conditional in the default case on 322 is redundant as it will always be true again now in that case, but this doesn't break all of coalescing because there is still the conditional outside the select that is the main place we throttle updates being delivered to `sendCh`.

agent/proxycfg/manager_test.go
dhiaayachi: This sleep is to ensure that we saturate the snap channel and get to the default. because of the check in http://github.com/hashicorp/consul/blob/f68b1a4a77741ae4c293571efd32b943971ec9ab/agent/proxycfg/state.go#L357-L357 if we don't wait between updates we only send once which do not fill `snapCh` and get us to the `default` case.

agent/proxycfg/state.go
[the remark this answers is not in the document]
dhiaayachi: Yes that's right the check is not needed

The use case I found the  bug from have a lot of updates happening (specially for an ingress Kind) not sure why but saw a lot of updates happening per second for the same ingress service, I would guess that the `coalesceTimeout` is useful for this type of behaviours.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/540755.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/540783.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit e653f8191904e7d84d0e6403497d3897a6e18629 onto `release/1.11.x` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit e653f8191904e7d84d0e6403497d3897a6e18629 onto `release/1.10.x` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit e653f8191904e7d84d0e6403497d3897a6e18629 onto `release/1.9.x` succeeded!

Issue #11925, opened 2021-12-31
Edit kv in webui which key contains '@' cause 404


finchcn:
When filing a bug, please include the following headings if possible. Any example text in this template can be deleted.

#### Overview of the Issue

Edit kv in webui which key contains '@' cause 404

#### Reproduction Steps

Steps to reproduce this issue, eg:

1. Create a cluster with n client nodes n and n server nodes
1. Run `curl -X PUT "http://localhost:8500/v1/kv/somepath/a@b:c" -d "123"` 
2. Open WebUI and click the key
1. View error

### Consul info for both Client and Server

<details>
  <summary>Client info</summary>

```
output from client 'consul info' command here
```

</details>

<details>
  <summary>Server info</summary>

```
output from server 'consul info' command here
```

</details>

### Operating system and Environment details

OS, Architecture, and any other information you can provide about the environment.

Server OS: ubuntu
WebBrowser: Edge On Mac

### Log Fragments

Include appropriate Client or Server log fragments. If the log is longer than a few dozen lines, please include the URL to the [gist](https://gist.github.com/) of the log instead of posting it in the issue. Use `-log-level=TRACE` on the client and server to capture the maximum log detail.

I think the bug is url encode some char twice

Amier3:
Hey @finchcn 

Welcome to the hashicorp community! 

I tried to reproduce this using the steps you provided but wasn't able to. On a fresh cluster I was able to successfully add the key with an '@' sign without error.

<img width="1777" alt="Screen Shot 2022-01-03 at 1 10 29 PM" src="https://user-images.githubusercontent.com/92801461/147964732-f774ec31-aba2-477d-adf7-7a72f4121862.png">


Could you provide us with: 

- The version of consul you're running  
- A screenshot of the error you're seeing 
- The output from `consul info` on the cluster seeing the error

PR #11931, opened 2022-01-04
ui: Fix URL params decoding

johncowen:
In 1.11.0/1 we moved our routing to use our Routlets almost entirely, but we missed moving the extra URL decoding for Ember Wildcard routes over to our routlet code that deals with URL params 😿 (Ember does not urlencode/urldecode wildcard routes by default)

This PR moves the code from our base Route (which all of our routes inherit from) over to our Routlet Service via a `normalizeParamsFor` method, which we then call from both our base Route and the Routlet Service. Whilst we barely use Ember Routes anymore, I left the required decoding in our base Route just incase it's required in future.

I'm a little surprised we had no acceptance test for this, so this fixes an unfortunate regression in 1.11.0/1 plus adds an acceptance test which would have caught this (and will catch it if it ever happens again)

Fixes https://github.com/hashicorp/consul/issues/11925

Small note here: If we ever want this to be non-project specific we'll have to decide on how to inject the project specific routes into the routlet service, for the moment we simply `import`

### DIFF
--- a/.changelog/11931.txt
+++ b/.changelog/11931.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fixes a bug with URL decoding within KV area
+```
--- a/ui/packages/consul-ui/app/routing/route.js
+++ b/ui/packages/consul-ui/app/routing/route.js
@@ -2,17 +2,14 @@ import Route from '@ember/routing/route';
 import { get, setProperties, action } from '@ember/object';
 import { inject as service } from '@ember/service';
 
-// paramsFor
 import { routes } from 'consul-ui/router';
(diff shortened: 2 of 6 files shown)


randallmorey (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/538867.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit f8b336b9a02518b61b8ff212324c93b8778ce7d8 onto `release/1.11.x` succeeded!

PR #11937, opened 2022-01-05
ui: [BUGFIX] Fixes an issue when editing intentions from the service > intentions sub tab

johncowen:
We recently changed the intentions form to take a full model of a `dc` rather than just the string identifier (so `{Name: 'dc', Primary: true}` vs just `'dc'` in order to know whether the DC is the primary or not.

Unfortunately, we only did this on the global intentions page not the per service intentions page. This makes it impossible to save an intention from the per service intention page (whilst you can still save intentions from the global intention page as normal).

The fix here pretty much copy/pastes the approach taken in the global intention edit template over to the per service intention edit template.

Tests have been added for creation in the per service intention section, which again are pretty much just copied from the global one, unfortunately this didn't exist previously which would have helped prevent this.

### DIFF
--- a/ui/packages/consul-ui/tests/acceptance/steps/dc/services/show/intentions/index-steps.js
+++ b/ui/packages/consul-ui/tests/acceptance/steps/dc/services/show/intentions/index-steps.js
@@ -1,4 +1,4 @@
-import steps from '../../../steps';
+import steps from '../../../../steps';
 
 // step definitions that are shared between features should be moved to the
 // tests/acceptance/steps/steps.js file
--- a/ui/packages/consul-ui/tests/helpers/type-to-url.js
+++ b/ui/packages/consul-ui/tests/helpers/type-to-url.js
@@ -5,8 +5,10 @@ export default function(type) {
       requests = ['/v1/catalog/datacenters'];
       break;
     case 'service':
-    case 'instance':
       requests = ['/v1/internal/ui/services', '/v1/health/service/'];
+      break;
+    case 'instance':
+      requests = ['/v1/health/service/'];
       break;
     case 'proxy':
       requests = ['/v1/catalog/connect'];
--- a/ui/packages/consul-ui/tests/pages/dc/intentions/edit.js
+++ b/ui/packages/consul-ui/tests/pages/dc/intentions/edit.js
@@ -10,7 +10,12 @@ export default function(
 ) {
   return {
     scope: 'main',
-    visit: visitable(['/:dc/intentions/:intention', '/:dc/intentions/create']),
+    visit: visitable([
+      '/:dc/intentions/:intention',
+      '/:dc/services/:service/intentions/:intention',
+      '/:dc/services/:service/intentions/create',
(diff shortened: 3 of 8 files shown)


ui/packages/consul-ui/tests/acceptance/steps/dc/services/show/intentions/index-steps.js
```diff
@@ -0,0 +1,10 @@
+import steps from '../../../../steps';
+
+// step definitions that are shared between features should be moved to the
+// tests/acceptance/steps/steps.js file
+
+export default function(assert) {
+  return steps(assert).then('I should find a file', function() {
+    assert.ok(true, this.step);
+  });
+}
```
johncowen, line 10: Not sure who will be reviewing this, but if you aren't aware these files are unfortunately necessary, automatically generated, and never edited, but they need to exist.

ui/packages/consul-ui/tests/helpers/type-to-url.js
```diff
@@ -5,9 +5,11 @@ export default function(type) {
       requests = ['/v1/catalog/datacenters'];
       break;
     case 'service':
-    case 'instance':
       requests = ['/v1/internal/ui/services', '/v1/health/service/'];
       break;
+    case 'instance':
+      requests = ['/v1/health/service/'];
+      break;
```
johncowen, line 12: During acceptance tests when you write:

Given 2 _instance_ models from yaml:
\---
\- Service
...etc
\---

This means that it will only overwrite the mock data received from the health endpoint, not both the service listing _and_ the health endpoint. There are a few legacy/historical reasons why this happened this way, but this is more correct.

ui/packages/consul-ui/tests/pages/dc/intentions/edit.js
```diff
@@ -10,7 +10,12 @@ export default function(
 ) {
   return {
     scope: 'main',
-    visit: visitable(['/:dc/intentions/:intention', '/:dc/intentions/create']),
+    visit: visitable([
+      '/:dc/intentions/:intention',
+      '/:dc/services/:service/intentions/:intention',
+      '/:dc/services/:service/intentions/create',
+      '/:dc/intentions/create',
+    ]),
```
johncowen, line 18: In the acceptance tests when we write

When I visit the _intention_ page for yaml:

Depending on the data you pass to that step, it will try each of these paths/urls in order until it finds one where it has all the data for that path - so in this case if I don't pass a service slug it will fall through to the last entry here.

johncowen (commented):
Notes:

natmegs (approved):
Pair reviewed

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/543264.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit adb0fa38d2caf29403c5a7aa43fceff6d3d9c047 onto `release/1.11.x` succeeded!

PR #11938, opened 2022-01-05
ui: Add selective no-console eslint rule

johncowen:
Thankfully @amyrlam spotted this in a recent PR https://github.com/hashicorp/consul/pull/11900#discussion_r774261990 and this PR is off the back of that.

Back when we started in Ember 2.18 eslint tests would be run in the QUnit test runner along with the rest of our tests, initially this proved to be a little frustrating when trying to debug, create and fixup tests, as everytime you'd use a `console.log` statement somewhere a bunch more tests would fail confusing matters. Due to this we didn't add a no console rule to our eslint config, but instead kept a 'dev' eslint config that we could run separately containing a no-console rule. Nowadays we barely run this at all (and infact we commented it out quite a while back), but by chance we've been fairly good at keeping `console.log`s out of PRs and Draft PRs

During upgrades to Ember 3.something eslint tests stopped running in the QUnit test runner and we changed to use a specific make target which runs `eslint` directly using node. Therefore eslint no longer runs whilst you are debugging, creating or writing tests, only when you initially start up the tests. All in all this means there is no longer a need for a separate 'dev' eslint config and we can bring in these extra nice to have rules to our eslint setup.

Changes:

1. Deleted the old dev eslint setup things.
2. Moved over the no-console rule to our eslint config
3. Didn't bother with the snake case thing as it doesn't affect us anymore (it was originally setting that to a warn so it must have affected us at some point)
4. We had a few console.logs but these where for DX purposes and were therefore all wrapped in `runInDebug`. I changed all these to `console.info` which makes more sense for what their purpose is.
5. Used eslint config to allow console.error and console.info only as we use these for good reason (I'd rather do this using config rather than using eslint comments everywhere)

### DIFF
--- a/ui/packages/consul-ui/.dev.eslintrc.js
+++ b/ui/packages/consul-ui/.dev.eslintrc.js
@@ -1,8 +1,0 @@
-module.exports = {
-  extends: ['./.eslintrc.js'],
-  rules: {
-    'no-console': 'warn',
-    'no-unused-vars': ['error', { args: 'none' }],
-    'ember/routes-segments-snake-case': 'warn',
-  },
-};
--- a/ui/packages/consul-ui/.eslintrc.js
+++ b/ui/packages/consul-ui/.eslintrc.js
@@ -14,6 +14,7 @@ module.exports = {
     browser: true,
   },
   rules: {
+    'no-console': ['error', {allow: ['error', 'info']}],
     'no-unused-vars': ['error', { args: 'none' }],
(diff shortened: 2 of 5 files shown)


amyrlam (approved):
nice one 👍

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/540547.

PR #11942, opened 2022-01-05
Fix test for ENT

kisunji:


agent/agent_endpoint_test.go
```diff
@@ -1567,6 +1567,7 @@ func TestAgent_Metrics_ACLDeny(t *testing.T) {
 }
 
 func TestHTTPHandlers_AgentMetricsStream_ACLDeny(t *testing.T) {
+	t.Skip("this test panics without a license manager in enterprise")
```
eculver, line 1570: should we move this to an ent-only file then? If I'm reading this correctly, this will skip here in OSS and we want it to only skip in ENT?

kisunji: I'm not sure about the context behind this change; I missed merging these lines properly in a ENT->OSS sync so this is just to make them equal again.

kisunji: Actually I removed these lines from OSS. I think you are correct about skipping these only in ENT

kisunji: @dnephin Should these tests exist in an `agent_endpoint_oss_test.go`?

kisunji:
Messed up an amend commit; closing

PR #11955, opened 2022-01-06
ui: Allow Route templateName paths to be relative

johncowen:
Continuing on from our "Centralized Routing Configuration from The Outside" thing, this PR extracts the simplified `resolve` function (similar to NodeJS `path.resolve`), adds initial unit testing to that and then reuses that for non-default `templateName`/`template` configuration. Previously (and currently) we were using this for routing redirects also.

Further thoughts/notes:

1. I had a mini internal bikeshed on the `import` shape of `resolve` here, wasn't sure whether to copy the NodeJS-like form. In the end I decided to make it not like the NodeJS one incase anyone sees it and thinks all of NodeJS `path` is available.
2. Everything in the routing configuration can now be relative, paths (as is always the case in Ember), redirects and templates. Additionally redirects and templates can be absolute, but we probably won't ever need to do that.
3. Potentially confusingly path relative paths begin with `/` (or they can be just with no prefix) as these are from Ember for-real. We could make these require `./` at some point in the future so then everything is the same and makes sense. I've left that for future thinking for now though as its not important in the slightest.

Lastly, I split off a bit of work out of the PR here (that I still have here ready to PR) in order to keep this PR more specific. I'll finish that up at some point and get that PRed also. Its related to making all this config available in our Routes which means we can get rid of this:

https://github.com/hashicorp/consul/blob/065f6f89fbe9f44ae4d66d2096d258f99608b03f/ui/packages/consul-ui/app/routing/route.js#L89-L98

End game there is to reduce the amount of code in our custom `basic:route` (which is what that ^ code is from)

Vaguely unrelated: I spotted a file we hadn't deleted in the last PR around this (the service instance route file) so I deleted that here.


amyrlam (approved):
code LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/549704.

PR #11956, opened 2022-01-06
Enable Security Scan for CRT

claire-labry:
What this PR does is it enables security scanning (owned by the Security team) for this repository. This is a part of the Common Release Tooling (CRT) effort which `enables a consistent workflow for security scanning that can be used across products and services. It provides a single abstraction point to allow us integrate our own scanning capabilities, as well as leverage external tools and APIs from vendors, open source, or free-to-use projects.` Learn more about HashiCorp's security-scanner [here](https://github.com/hashicorp/security-scanner/).

While running the security scan tests, I ran into a [false positive](https://github.com/hashicorp/crt-workflows-common/runs/4550690803?check_suite_focus=true) with `go_modules` in the [binary](https://github.com/hashicorp/consul/blob/0b3dfcb195413298b9970267d37d72a201b7987c/.release/security-scan.hcl#L9) stanza in the security-scan.hcl file. This is apparently a known issue within your team-- this will need to be updated in order for the workflows to pass the `security-scan-binaries` event.

I was able to test out `security-scan-containers` by turning `go_modules` to false and the most recent tests are [passing](https://github.com/hashicorp/crt-workflows-common/actions/runs/1593008291).

You'll also see that I had to upgrade the Dockerfile to use alpine:3  as the older version caused the security-scanner to pick up some vulnerabilities, CVEs to be specific. See error outputs [here](https://github.com/hashicorp/crt-workflows-common/runs/4473388427?check_suite_focus=true) that arose with consul-terraform-sync. We worked with Security and they scanned different versions of the alpine base image and concluded that version 3 picks up the most stable + clean version of alpine and suggested to upgrade to that version to avoid any friction and potential vulnerabilities in the future and simply tidy up the container.

If anyone needs any further clarification or if nothing makes sense, feel free to comment or reach out to me! ❤️


sarahethompson (approved):

.github/workflows/build.yml
```diff
@@ -3,7 +3,9 @@ name: build
 on:
   push:
     # Sequence of patterns matched against refs/heads
-    branches: [ main ]
+    branches:
+      # Push events on the main branch
+      - main
```
eculver, line 9: This is where there is a conflict. It seems easy enough to fix to just remove this from the changeset.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # This Dockerfile creates a production release image for the project using crt release flow.
-FROM alpine:3.13 as  default
+FROM alpine:3 as  default
```
eculver, line 2: I'm curious if this was intentionally updated? I think there has been issues in the past upgrading point releases of Alpine, so we might want to stay pin to `3.13`.

.release/ci.hcl
```diff
@@ -42,8 +42,36 @@ event "upload-dev" {
...
+    config = "security-scan.hcl"
+  }
+
+  notification {
+    on = "fail"
+  }
+}
+
+event "security-scan-containers" {
+  depends = ["security-scan-binaries"]
```
eculver, line 60: Can these be run in parallel?

eculver: [I see](https://cs.github.com/?scopeName=All+repos&scope=&q=security-scan-containers+org%3Ahashicorp) that the other repositories that have been onboarded to CRT do it this way, so if you don't feel like bucking the trend or explaining, that's fine too. I am just curious.

.release/security-scan.hcl
```diff
@@ -0,0 +1,13 @@
...
+	secrets      = true
+}
+
+binary {
+	secrets      = true
+	go_modules   = true
+	osv          = true
+	oss_index    = true
+	nvd          = true
+}
```
eculver, line 13: I'm curious why these aren't the default settings?

.release/security-scan.hcl
```diff
@@ -0,0 +1,13 @@
+container {
+	dependencies = true
+	alpine_secdb = true
+	secrets      = true
```
eculver, line 4: Is this supposed to be a `block` type? I'm looking at the config docs [here](https://github.com/hashicorp/security-scanner/blob/main/CONFIG.md#block-attributes-1)

eculver (commented):
The change looks good, I just have a few small comments.

Dockerfile
claire-labry: Yes, this was intentionally updated to allow `alpine:3` to pick up the latest + cleanest version the alpine image. Security scanned `alpine:13` and found some known CVEs. If the goal is to be more prescriptive here, Security scanned `alpine:15` (which is latest + most secure) to be clean of any vulnerabilities. I can absolutely make that change.

.release/ci.hcl
claire-labry: I don't think these can be run in parallel as each event has a `depends` requirement. This is to allow workflows to be run consecutively. If the security-scanner picks up any vulnerabilities, it would stop any subsequent workflows which I believe is what we want. If there are any outputted vulnerabilities, it will need to be reported to #team-prodsec in Slack. cc: @sarahethompson to ensure I'm not leaving out any vital context.

.release/security-scan.hcl
claire-labry: This is a really good question that I don't think I quite have the answer to. Let me do some sleuthing and circle back to you on this.

.release/security-scan.hcl
claire-labry: Yes, this is supposed to be a `block` type. This is the configuration that has been set up for CRT (Common Release Tooling). I'm not sure if I quite answered your question here, but please add follow-up questions and I'd be happy to clarify any bits!

eculver: It not being a `block` type must not be causing any issues given that all the other projects using the scanner, but I'm curious if this means that it's the same as:
```
secrets {
    all = true
}
```

Ok, I went down the rabbit hole. Apparently, `secrets = true` is a deprecated configuration option, but [it is still supported](https://github.com/hashicorp/security-scanner/blob/feb4c96ac64d9d08fd837f0d11385c234d0b11b1/cmd/scan/detect_config.go#L67-L88). Using `secrets = true` is the same as `secrets { all = true }`.

I wonder if we should be using the updated `block` type here to future-proof it?

Dockerfile
eculver: Ok, makes sense. I did a little searching around Slack to see if there was any reason why we are pinned to `3.13` and I didn't find anything. However, we may want to do this as a separate change considering it could have ramifications that are outside the scope of security scanning. It's up to you though. 

Also, I see comments [like this](https://github.com/hashicorp/consul-terraform-sync/pull/545#discussion_r766927398) from other PRs and am realizing that this is not a new topic for you :) 

To be clear and reiterate, I think this is probably a safe change, but it's your call as to whether we include it here vs in a separate PR.

eculver (approved):
LGTM.

eculver:
PS - I added a changelog entry for it to satisfy the changelog-checker

Dockerfile
claire-labry: I think it'd be ok to keep this change in this PR and haven't seen any ramifications with this change among other products. Not to be over-generalizing, but to agree with you that this is a safe change :D

.release/security-scan.hcl
claire-labry: Update has been made!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/576550.

Issue #11773, opened 2021-12-08
consul agent logs not works properly on windows (empty file)


inorangestylee:
#### Overview of the Issue

after start consul agent (as windows service) it creates an empty log file and not appends it with log messages.
Directory ACL checked. While in work consul agent obtain file handler (cant delete or append anything in logfile).

#### Reproduction Steps

Config:
```
...
    "log_level": "INFO",
    "log_file": "c:\\consul\\logs\\",
    "log_rotate_duration": "24h",
    "log_rotate_max_files": 14,
...
```

Steps to reproduce this issue, eg:

1. Start consul agent as windows service `c:\consul\consul.exe agent -config-file c:\consul\config\config.json` or `c:\consul\consul.exe agent -config-file c:\consul\config\config.json -log-file c:\consul\logs\`
1. Start `.\consul.exe monitor -log-level=INFO`
1. View log and compare monitor stdout

### Consul info for both Client and Server

<details>
  <summary>Client info</summary>

```
PS C:\consul> .\consul.exe info
agent:
        check_monitors = 0
        check_ttls = 0
        checks = 1
        services = 2
build:
        prerelease =
        revision = a4f67ec2
        version = 1.8.15
consul:
        acl = enabled
        known_servers = 5
        server = false
runtime:
        arch = amd64
        cpu_count = 8
        goroutines = 56
        max_procs = 8
        os = windows
        version = go1.14.15
serf_lan:
        coordinate_resets = 0
        encrypted = false
        event_queue = 0
        event_time = 174
        failed = 0
        health_score = 0
        intent_queue = 0
        left = 0
        member_time = 222538
        members = 324
        query_queue = 0
        query_time = 194
```

</details>

### Operating system and Environment details

OS Name:                   Microsoft Windows Server 2016 Standard (x64)
OS Version:                10.0.14393 N/A Build 14393

kopetersen:
I have the same problem. I found a workaround though.
It´s more or less the same thing NSSM does when you use that.

If I install the service like below, I get a working log. So I guess the stream redirection is not working properly for Consul
`cmd.exe /C C:\Consul\consul.exe agent -config-dir=C:\Consul\config 2>&1`

On a sidenote, I have Nomad running on the same servers, and that logs just fine.

Consul info
```
agent:
        check_monitors = 0
        check_ttls = 0
        checks = 2
        services = 3
build:
        prerelease =
        revision = c976ffd2
        version = 1.10.3
consul:
        acl = disabled
[...]

PR #11960, opened 2022-01-06
Fix Windows logging to files

kisunji:
closes https://github.com/hashicorp/consul/issues/11773

Our logger uses [`io.MultiWriter`](https://pkg.go.dev/io#MultiWriter) which at minimum writes to `os.Stdout` and optionally to syslog or log file targets.

When Consul is run as a Windows Service, `os.Stdout` is not available and any writes by the logger fails silently, which inadvertently stopped writes to other targets because of `io.MultiWriter`'s behaviour:
```
If a listed writer returns an error, that overall write operation stops 
and returns the error; it does not continue down the list.
```

This PR uses a `noErrorWriter` wrapper which never returns an error so that even if writes to Stdout fail, writes to syslog and log file are unaffected.

### DIFF
--- a/.changelog/11960.txt
+++ b/.changelog/11960.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+windows: Fixes a bug with empty log files when Consul is run as a Windows Service
+```
--- a/command/agent/agent.go
+++ b/command/agent/agent.go
@@ -162,7 +162,10 @@ func (c *cmd) run(args []string) int {
 
 	// FIXME: logs should always go to stderr, but previously they were sent to
 	// stdout, so continue to use Stdout for now, and fix this in a future release.
-	logGate := &logging.GatedWriter{Writer: c.ui.Stdout()}
+	// UiWriter is being used here so that writes to Stdout do not error (e.g.
+	// Windows Services are non-interactive and do not have Stdout) and inadvertently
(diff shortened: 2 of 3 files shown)


dnephin (commented):
Nice fix!

`UIWriter` also strips trailing newlines, I wonder if that might cause other problems.

Would this also work?

```go
type noErrorWriter struct {
    w io.Writer
}

func (w noErrorWriter) Write(p []byte) (n int, err error) {
    n, _ := w.w.Write(p)
    return n, nil
}
```

dnephin (approved):
Nice! Thanks for debugging and fixing this!

kisunji:
Testing on windows: 
image

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/542579.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 08af4f7ffca9f96e1f55fec8d81e83208a706e5b onto `release/1.11.x` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 08af4f7ffca9f96e1f55fec8d81e83208a706e5b onto `release/1.10.x` succeeded!

Issue #11847, opened 2021-12-15
consul auto-encryption fails on IPv6-only cluster


markmartirosian:
#### Overview of the Issue

Client certificate distribution fails on a IPv6-only cluster using the auto-encryption method.  

### Consul info

<details>
  <summary>Server info</summary>

```
agent:
        check_monitors = 0
        check_ttls = 0
        checks = 3
        services = 3
build:
        prerelease =
        revision = 7bbad6fe
        version = 1.10.4
consul:
        acl = enabled
        bootstrap = false
        known_datacenters = 1
        leader = false
        leader_addr = [2a05:d014:d9e:c305:b0bc:d053:f019:ed77]:8300
        server = true
raft:
        applied_index = 9789
        commit_index = 9789
        fsm_pending = 0
        last_contact = 83.945604ms
        last_log_index = 9789
        last_log_term = 4
        last_snapshot_index = 0
        last_snapshot_term = 0
        latest_configuration = [{Suffrage:Voter ID:22da461e-212d-76e2-95e5-0fb99ae5097e Address:[2a05:d014:d9e:c303:e4d3:d281:a61d:8ebd]:8300} {Suffrage:Voter ID:6d370695-dac6-5205-6070-325b3cff86a8 Address:[2a05:d014:d9e:c305:b0bc:d053:f019:ed77]:8300} {Suffrage:Voter ID:c4ed1b5a-859f-ca1a-f6dd-ab2555f8e7cf Address:[2a05:d014:d9e:c304:fc9f:38cd:a734:5022]:8300}]
        latest_configuration_index = 0
        num_peers = 2
        protocol_version = 3
        protocol_version_max = 3
        protocol_version_min = 0
        snapshot_version_max = 1
        snapshot_version_min = 0
        state = Follower
        term = 4
runtime:
        arch = arm64
        cpu_count = 2
        goroutines = 110
        max_procs = 2
        os = linux
        version = go1.16.10
serf_lan:
        coordinate_resets = 0
        encrypted = true
        event_queue = 0
        event_time = 4
        failed = 0
        health_score = 0
        intent_queue = 0
        left = 0
        member_time = 7
        members = 3
        query_queue = 0
        query_time = 1
serf_wan:
        coordinate_resets = 0
        encrypted = true
        event_queue = 0
        event_time = 1
        failed = 0
        health_score = 0
        intent_queue = 0
        left = 0
        member_time = 1
        members = 1
        query_queue = 0
        query_time = 1
```

</details>

[...]

PR #11965, opened 2022-01-07
Ma/ipv6 robustness fixes

markan:
This is an effort towards being more robust in the face of ill formed address/port combinations, such as seen in https://github.com/hashicorp/consul/issues/11847


jmurret:

PR #11967, opened 2022-01-07
[DO NOT REVIEW] wip: envoy boostrap template fidling

FFMMM:
Not yet ready to review, likely not the right approach.

TODO:

* clarify how to actually apply to listeners
* validate envoy config (pretty sure template not valid atm)
* clarify global vs per listener config


FFMMM:
this has been on the back burner for a bit; will re open once it's actively worked on again

PR #11969, opened 2022-01-07
ui: Runtime Injectable Components

johncowen:
This PR does several things that are all slightly related:

1. Simplifies how we 'import' our configuration files a little in order to make them more grokable.
2. Starts to exclude files based on explicit configuration rather than convention.
3. Adds the first instance of us being able to select an implementation (of multiple) of a component at runtime.

## 1. Configuration files

When we first added these I tried made them uber non-polluting, so everything was using the `appName` to ensure we only used the configuration files from our app. If this was a bigger 'thing' then this would be important to do, but I felt in the scheme of things it was more important to make these files as simple and grokable as possible. So I took out all the `appName` things. We control the entire app and distribution so the non-pollution aspect isn't super important. If it ever becomes important we can always put this stuff back.

## 2. Explicit configuration rather than convention

These configuration files are now consumable both in a browser and in a CommonJS environment (e.g. NodeJS/ember-cli). This means we can easily reuse this configuration both in the browser environment during runtime and in the ember-cli/node environment during build time. This means we begin to know the exact filenames that we need to exclude from a production build instead of relying on wildcarding/globbing and naming conventions such as `*-debug`.

Note: I haven't removed the previous wildcarding/convention based approach as yet as we still need to make sure that things other than component and services can be excluded in this way (for example templates).

## 3. Injectable components at runtime

We have various features such as ACLs, Namespaces and Partitions which are optionally activated depending on what version of Consul you are running or what configuration you have (or potentially in the future, what other apps/services you have running). Currently features/components related to these features are added to the UI but enabled/disabled via various conditionals (using `ember-can`) dotted all over the code base.

Ideally, instead of having these components all over the app with conditionals wrapped around them, we would enable/disable them in one place in configuration at runtime. This PR is the first instance of us doing this.

Here for example, when running the app without partitions enabled, the `Consul::Partition::Selector` component is simply an instance of `@glimmer/component` i.e. an empty component that does absolutely nothing - kind of like a noop. Whereas when partitions are enabled, Consul itself (from the backend) therefore sources/loads the partition service configuration file which injects a different implementation of `Consul::Partition::Selector` (`consul-partitions/components/consul/partition/selector`) and the component then 'comes alive'. All in all this prevents us from having to wrap components in conditionals all over the codebase, we just add them where they are meant to be when they are 'brought to life'.

I've only done one component here mainly for example purposes, but there will be a follow up PR removing all the top-level conditionals from these app components (all those in `ui/packages/consul-{partitions,acls,nspaces}`.

All that is left to do now here towards this endgame (being able to load completely separate applications into our app at runtime) is separating off the source code for the external applications into separate JS files and sourcing those separately to the main `consul-ui.js` source file. Once we finish that up we can potentially ship features for these external applications on a completely different release schedule to the main Consul binary.


ui/packages/consul-partitions/vendor/consul-partitions/services.js
```diff
@@ -0,0 +1,9 @@
+(services => services({
+  "component:consul/partition/selector": {
+    "class": "consul-ui/components/consul/partition/selector"
```
johncowen, line 3: Use this class for the `Consul::Partition::Selector` component _only_ if this config file has been 'imported' (which is only when Consul has set the `PartitionsEnabled` feature flag to `true`)

ui/packages/consul-ui/app/instance-initializers/container.js
```diff
@@ -3,38 +3,33 @@ import require from 'require';
...
-  .map((item, i) => (i ? `${item.substr(0, 1).toUpperCase()}${item.substr(1)}` : item))
-  .join('');
 
 export const services = merge.all(
-  [].concat(
-    ...[...doc.querySelectorAll(`script[data-${appName}-services]`)].map($item =>
-      JSON.parse($item.dataset[`${appNameJS}Services`])
-    )
+  [...doc.querySelectorAll(`script[data-services]`)].map($item =>
+    JSON.parse($item.dataset[`services`])
```
johncowen, line 9: More simplified `<script src="the-config-file.js" />` slurping/'importing'.

ui/packages/consul-ui/app/instance-initializers/container.js
```diff
@@ -3,38 +3,33 @@ import require from 'require';
...
+          );
         } else {
           throw new Error(`Unable to locate '${value.class}'`);
         }
-      break;
+        break;
     }
   });
-}
+};
```
johncowen, line 29: In this PR: Yey `prettier`. Nothing to see here, please move on, do not pass go, do not collect $200 😄 

Previous to this PR: This is where we do:

```javascript
// container here is pretty much the same as the ember `application`
container.register('component:consul/partition/selector', 'consul-ui/components/consul/partition/selector');
```

We do an extra replacement to remove the word `torii` from our application config (we might switch that out at some point _not_ to use the `torii` addon)

ui/packages/consul-ui/ember-cli-build.js
```diff
@@ -66,6 +68,30 @@ module.exports = function(defaults, $ = process.env) {
...
+                prev
+              );
+            }
+            return prev;
+          },
+          prev
+        )
+      },
+      excludeFiles
+    );
```
johncowen, line 94: This is one of my favourite parts. I've eternally found it difficult in ember to be specific about the files I do/don't want complied into my final build. For example, I might want certain helpers only to be included during development, maybe I want to inject a mocked ember Service in place of the real ember Service only during development.

It's non-obvious how to keep these development only files out of a production build in Ember.

Previously (and still currently after this PR) we use wildcard/globbing to say "anything that ends in `-debug.js` please don't compile in for production" i.e. a convention (you'll see this if you explan the code above line 68).

In this PR we reuse the configuration that we already use to do the actual injection of the dev-time-only Services/Helpers to figure out exactly what files need excluding i.e. using the explicit configuration over the convention.

Currently we still use the `*-debug.js` convention also, just as I want to make sure we can use the same approach for all-the-things including templates etc. Once we get there we will eventually remove the `*-debug.js` convention.

ui/packages/consul-ui/ember-cli-build.js
```diff
@@ -88,18 +114,24 @@ module.exports = function(defaults, $ = process.env) {
...
+      });
+      trees.vendor = mergeTrees([
+        new Funnel('vendor'),
+      ].concat(
+        apps.map(item => new Funnel(`${item.path}/vendor`))
+      ));
+    }
+  // consul-ui will eventually be a separate app just like the others
+  // at which point we can remove this filter/extra scope
+  )(apps.filter(item => item.name !== 'consul-ui'));
```
johncowen, line 134: Lines 132-134 where the only lines added here. This piece of code is to copy over our feature-flagged applications that live outside of the main application, as we added `consul-ui` as an 'application' we don't need to copy the files over for that 'application' as its the 'host application'. Therefore we just wanted to filter `consul-ui` out of the application list just for this bit of code. This avoids renaming `apps` to something like `appsWithoutConsulUI = apps.filter()` by just making a new temporary scope with `apps` overwritten which I prefer.

ui/packages/consul-ui/ember-cli-build.js
```diff
@@ -147,20 +179,31 @@ module.exports = function(defaults, $ = process.env) {
...
+    if(devlike) {
+      build(`vendor/${item.name}/routes-debug.js`, {
+        root: item.path,
+        outputFile: `assets/${item.name}/routes-debug.js`,
       });
+      build(`vendor/${item.name}/services-debug.js`, {
+        root: item.path,
+        outputFile: `assets/${item.name}/services-debug.js`,
+      });
+    }
```
johncowen, line 206: Previously we would blindly run `app.import` without checking for file existance. The new `build` function just 'decorates' `app.import` with a file existence check.

ui/packages/consul-ui/lib/startup/templates/body.html.js
```diff
@@ -42,26 +42,28 @@ ${environment === 'production' ? `{{jsonEncode .}}` : JSON.stringify(config.oper
...
-  <script data-app-name="${appName}" data-${appName}-services src="${rootURL}assets/consul-ui/services.js"></script>
+  <script src="${rootURL}assets/consul-ui/services.js"></script>
+  <script src="${rootURL}assets/consul-ui/routes.js"></script>
 ${
   environment === 'development' || environment === 'staging'
     ? `
-  <script data-app-name="${appName}" data-${appName}-services src="${rootURL}assets/consul-ui/services-debug.js"></script>
-  <script data-app-name="${appName}" data-${appName}-routing src="${rootURL}assets/consul-ui/routes-debug.js"></script>
+  <script src="${rootURL}assets/consul-ui/services-debug.js"></script>
+  <script src="${rootURL}assets/consul-ui/routes-debug.js"></script>
```
johncowen, line 51: Removal of all the `data-*` attributes as explained at the top of these PR notes.

Also note, we only load in the `*-debug.js` configuration when we are in a development-like environment. This configuration never gets loaded in production.

ui/packages/consul-ui/lib/startup/templates/body.html.js
```diff
@@ -42,26 +42,28 @@ ${environment === 'production' ? `{{jsonEncode .}}` : JSON.stringify(config.oper
...
+  <script src="${rootURL}assets/consul-acls/routes.js"></script>
 {{end}}
 {{if .PartitionsEnabled}}
-  <script data-app-name="${appName}" data-${appName}-routing src="${rootURL}assets/consul-partitions/routes.js"></script>
+  <script src="${rootURL}assets/consul-partitions/services.js"></script>
+  <script src="${rootURL}assets/consul-partitions/routes.js"></script>
 {{end}}
 {{if .NamespacesEnabled}}
-  <script data-app-name="${appName}" data-${appName}-routing src="${rootURL}assets/consul-nspaces/routes.js"></script>
+  <script src="${rootURL}assets/consul-nspaces/routes.js"></script>
```
johncowen, line 66: This is where we 'import' or don't 'import' additional configuration depending on what Consul says, things like `.NamespacesEnabled` are set to true/false directly by the backend (our index.html file is run through go-template before it gets served up to the user). Note: currently we only have injectable service/component for partitions currently - once this PR is merged we will be using the same approach for Namespaces and ACLs and more.

ui/packages/consul-ui/vendor/consul-ui/routes-debug.js
```diff
@@ -1,19 +1,20 @@
...
-      .map((item, i) => i ? `${item.substr(0, 1).toUpperCase()}${item.substr(1)}` : item)
-      .join('');
-    data[`${appNameJS}Routes`] = JSON.stringify(json);
+    },
+  }))(
+  (
+    json,
+    data = typeof document !== 'undefined' ? document.currentScript.dataset : module.exports
+  ) => {
+    data[`routes`] = JSON.stringify(json);
```
johncowen, line 18: This route is only complied/enabled if you are in development mode (note the routes-debug.js name of the file). It's the page/route for our fake OIDC provider we use for manually testing SSO.

ui/packages/consul-ui/vendor/consul-ui/services-debug.js
```diff
@@ -1,15 +1,16 @@
...
+    },
+    'service:intl': {
+      class: 'consul-ui/services/i18n-debug',
+    },
+  }))(
+  (
+    json,
+    data = typeof document !== 'undefined' ? document.currentScript.dataset : module.exports
+  ) => {
+    data[`services`] = JSON.stringify(json);
```
johncowen, line 14: Development time only routes for:

1. application: This adds our docs menu/navigation if you go to anything under `/docs`
2. intl: various development utilities to help us convert the entire app to 100% support i18n (eventually! 😄 )

ui/packages/consul-ui/vendor/consul-ui/services.js
```diff
@@ -1,21 +1,25 @@
...
+    },
+    'service:state': {
+      class: 'consul-ui/services/state-with-charts',
+    },
+    'auth-provider:oidc-with-url': {
+      class: 'consul-ui/services/auth-providers/oauth2-code-with-url-provider',
+    },
+    'component:consul/partition/selector': {
+      class: '@glimmer/component',
+    },
```
johncowen, line 17: By default (if partitions aren't enabled) the `Consul::Partition::Selector` is just an empty component (an direct instance of `@glimmer/component`)

There are also some other injections we need here, but all of this file is _always_ loaded i.e. these are production injections.

johncowen (commented):
Notes:

amyrlam (approved):
Code 👍

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/557938.

PR #11913, opened 2021-12-23
ui: Upgrade AuthDialog

johncowen:
This PR has no user facing changes, under-the-hood upgrades include:

1. Move AuthDialog to use a Glimmer Component plus native named blocks/slots.
2. Unravel the Auth* contextual components, there wasn't a lot of point having them as contextual components and now the AuthDialog (non-view-specific state machine component) can be used entirely separately from the view-specific components (AuthForm and AuthProfile).
3. Move all the ACL related components that are in the main app chrome/navigation (our HashicorpConsul component) in our `consul-acls` sub package/module (which will eventually be loaded on demand only when ACLs are enabled)

Whilst the changelog looks like its a lot, it's mainly moving things around with not a lot of change.

This targets 1.12 so no backporting.

Theres more related 'moving around' to come here, but I stopped short here to prevent the PR getting too big, and it felt like a place I could stop and put up a meaningful/boxed off PR.

PR #11981, opened 2022-01-10
ui: Fix up wiring or empty state login button

johncowen:
During https://github.com/hashicorp/consul/pull/11913 I noticed that I'd re-wired thing up slightly wrong (we added a wrapper component in that PR and it was rewired slightly wrong). I also used a `did-insert` helper which would not fire on `did-update` so I changed this to use our older `Ref` component which fires on insert and update.

The consequence of this bug was that our login button that appears in our empty state (which also opens the login modal) would be missing as in the EmptyState component (where this button is added) the button will only be added if the `open` property is defined. As I rewired it wrong open would be undefined and therefore the button would not show.

This bug is not currently in a release, and the original PR where it was introduced was only merged to `main` hence no backport/changelog here.

### DIFF
--- a/ui/packages/consul-acls/app/components/consul/acl/selector/README.mdx
+++ b/ui/packages/consul-acls/app/components/consul/acl/selector/README.mdx
@@ -1,0 +1,35 @@
+# Consul::Acl::Selector
+
+A component to allow the user to 'select' the ACL area they wish to view.
+Whilst the 'Selector' naming here is a bit of a reach, it fits well with other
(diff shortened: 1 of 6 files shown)


johncowen (commented):
Notes:

amyrlam (approved):
Took a look in tandem with #11982, code LGTM

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/549884.

PR #11982, opened 2022-01-10
ui: Some ACL component documentation

johncowen:
Whilst working on https://github.com/hashicorp/consul/pull/11981 I added docs for the new TokenSelector and ACLSelector (which are mainly just wrapper components), I wasn't sure whether folks would prefer a separate PR so I decided to do this one separate.

There is one meaningful change here (which came to light due to writing the docs) in that previous we expected the user of the component to pass through a `@token` argument to AuthDialog (which then gets passed through to AuthProfile to show a truncated token.AccessorID.

As the AuthDialog component orchestrates the entire authorization dance (both for non-sso and sso flows) it reads and saves any user token changes and therefore knows about the token and 'exports'/yields it. Therefore I switched to use that reference to the `token` instead of the one we were passing in, which means one less `@argument` for the user of the TokenSelector component to worry about.

Actually now thinking about it I suppose the 'TokenSelector' component could export the token into `application.hbs`also 🤔 - will save that for a future PR so I can get this finished up, its not in the least bit important as of right now.

Lastly, as per the base branch of this PR, there is no backporting here as its only for a future yet to be released version of Consul (1.12)

### DIFF
--- a/ui/packages/consul-acls/app/components/consul/acl/selector/README.mdx
+++ b/ui/packages/consul-acls/app/components/consul/acl/selector/README.mdx
@@ -1,0 +1,35 @@
+# Consul::Acl::Selector
+
+A component to allow the user to 'select' the ACL area they wish to view.
+Whilst the 'Selector' naming here is a bit of a reach, it fits well with other
+similar 'Selector' components that we use within the main navigation area.
+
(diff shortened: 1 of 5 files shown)


amyrlam (approved):
Took a look in tandem with #11981, 👍

PR #11987, opened 2022-01-10
Improve CTS acronym usage (Top Half)

wilkermichael:
- Use Consul-Terraform-Sync at the first instance with CTS in brackets - Consul-Terraform-Sync (CTS) and then CTS for all following instances on a per-page basis.
- some exceptions: left usage of the term `Consul-Terraform-Sync` in config examples and where it made sense for hyperlinking


lornasong (approved):
nice! looks good to me

PR #11993, opened 2022-01-10
docs: improve ACL system introduction

dnephin:
This is a first step to improving the ACL system documentation. There are more changes to make here, but I am opening a PR for some early feedback.

So far the following changes are made:
1. Rewrite the introduction paragraph to describe the ACL system, describe the purpose of the document, and link to other related docs (learn tutorials and API/CLI reference). These links already exists in other places on the page, but this edit moves them to the top.
2. Replaces the overview of entities in the ACL system. Right now the replacement is just a diagram with a short description. I think we could possibly expand on the description.
3. Replace the ACL Token section. Instead of listing the properties of the entity describe how tokens work and link to the other related sections. I moved the ACL Token section from much lower on the page because tokens are the primary thing that people need to know about, so it seems like it should be first. 
4. Remove the "ACL" prefix from headers. These are all on the ACL page. 
5. Make all the important headers H2, so that they show up in the "table of contents" dropdown.

Next steps (probably in other PRs):
* [ ] redo the policies, roles, and {node, service} identifies sections to be more like the new tokens section
* [ ] remove the "Configuring ACLs" section. Move any relevant content to the [configuration reference](https://www.consul.io/docs/agent/options) and link from the ACL system docs to the ACL configuration section in the configuration reference
* [ ] remove the "ACL Rules and Scope" table, a new copy of that table now exists in the ACL rules reference, link to the new table instead of having two.
* [ ] remove the "Authorization" section, and make sure that's already available in the API reference. Link to it from the introduction paragraph.
* [ ] add a section about "ACLs with Admin Paritions and Namespace", and move the note about "restrictions on linking policies to tokens" to that section
* [ ] add a section about "ACLs with Multi-Datacenter", and move the note about "local tokens" to that section


website/content/docs/security/acl/acl-system.mdx
```diff
@@ -2,72 +2,52 @@
...
-  make requests to Consul.
+The diagram below shows the relationships between the core components of the ACL system:
 
-For many scenarios policies and tokens are sufficient, but more advanced setups
-may benefit from additional components in the ACL system:
-
-- **ACL Roles** - Roles allow for the grouping of a set of policies and service
-  identities into a reusable higher-level entity that can be applied to many
-  tokens. (Added in Consul 1.5.0)
+![ACL system component relationships](/img/acl-token-policy-rule-relationship.png)
```
dnephin, line 31: Maybe this diagram should be moved to the end of the "Tokens" section, what do you think?

That way it comes after the description of a token which talks about policies.

trujillo-adam:
Hey @dnephin - thanks for working on this. Good call on removing "ACL" from all of the headings. I don't know that I agree with all of the changes, though. I agree that moving tokens to the front might be a good because they are the primary interface, but you stripped a lot of information out of that section that I think is important to building understanding. For example, the token list command was removed, as well as the list of properties the user would see printed to the console when the agent starts. I also think all of the token-related sections (Special and Built-in Tokens) should be moved up front as subsections to the main token section. Also, we need to follow the style guide w/r/t hierarchy and headings, as opposed to arbitrarily using h2s for TOC purposes. If we're hacking around with the style guidance, then perhaps this page is too big and we need to make more concise topics in separate pages.

dnephin:
Thanks for taking a first pass! Happy to chat more about these changes synchronously, but some thoughts below.

> but you stripped a lot of information out of that section that I think is important to building understanding. 

I think a big problem with the docs today is that we overload the reader with way too much detail up front. Most readers are not going to care about every field on every entity. Mostly what they need from these docs is to understand the concepts. I believe by moving some of the detail in other sections we can accomplish that.

> the token list command was removed

Examples of running the token list command are probably better suited for the CLI reference docs, right? I do link directly to the API and CLI docs for creating tokens, but I think you're right that listing tokens may also be a key use case. I'll add a link tot he CLI command for listing tokens, and move the example output there.

> as well as the list of properties the user would see printed to the console when the agent starts

I'm not too sure what you are referring to here. Do you mean the properties of an ACL token? We don't print much about ACLs when the agent starts, and I don't see any of that output here.

We already have multiple references (api docs, and cli docs) which contain the full set of fields. I think we need to be linking to those, not duplicating them in a third place. I've linked to the API and cli docs from the Tokens section.

> I also think all of the token-related sections (Special and Built-in Tokens) should be moved up front as subsections to the main token section. 

My concern with this approach is that the reader most likely needs to learn more about other concepts (policies specifically) because they dig into specific use cases with tokens. I agree it is a bit awkward now, maybe we can find a better way or organizing this, but I think we also need to find a way to split up this work to keep it easy to review.

My first pass was to move the built-in tokens next to the other special tokens. All of these are related, so keeping them together seemed more appropriate.  Once we remove the Configuration,  Authorization, and "rules and Scope" table, the "built-in and special tokens" will be immediately after roles. This seems like a pretty good place for them because the page will flow nicely from "introduction of concepts" -> "more specific details about them". What do you think?

> we need to follow the style guide w/r/t hierarchy and headings

I believe my edits to the headers do follow the style guide. Do you see one that does not?

As far as I can tell, using H2 to make the page much more usable shouldn't conflict with the style guide.

website/content/docs/security/acl/acl-system.mdx
```diff
@@ -2,72 +2,57 @@
...
+Consul HTTP API requests accept a token from an [HTTP Header](/api-docs/index#authentication),
+and CLI requests accept a token from
+[`-token` command line flag](/commands#authentication) or
+[`CONSUL_HTTP_TOKEN_FILE` environment variable](/commands#consul_http_token_file).
 
-If the ACL system becomes inoperable, you can follow the
-[reset procedure](https://learn.hashicorp.com/tutorials/consul/access-control-troubleshoot#reset-the-acl-system) at any time.
 
-### ACL Policies
+## Policies
```
FFMMM, line 55: Why not `ACL Policies` like before?

dnephin: My thinking is:
1. since these are docs for the ACL system, repeating the system name as part of the entity names is not really helpful. I suspect this was done to mirror the name in the code. The names in the code are unfortunate as well. Generally they should be in an `acl` package, so they don't need ACL prefixes.
2. it reads strange, because ACL is supposed to be "Access Control List". It's not really an "Access Control List Policy". It's an "Access Control Policy", but it's not a policy about a list.  Using simply "Policy" seems like it helps to make this less confusing.

I suspect "Access Control system" would be a better name overall, but I'm not going to try and tackle a rename right now.

website/content/commands/index.mdx
```diff
@@ -91,6 +91,15 @@ Command Options
...
 ```
 
+## Authentication
+
+When the [ACL system is enabled](/docs/agent/options#acl_enabled) the Consul CLI will
+require an [ACL token](/docs/security/acl/acl-system#tokens) to perform API requests.
+
+The ACL token can be provided directly on the command line using the `-token` command line flag,
+from a file using the `-token-file` command line flag, or from the
+[`CONSUL_HTTP_TOKEN`](#consul_http_token_file) environment variable.
```
blake, line 101:
```suggestion
The ACL token can be provided using the `-token` command line flag or [`CONSUL_HTTP_TOKEN`](#consul_http_token) environment variable, or from a file using the `-token-file` command line flag
or [`CONSUL_HTTP_TOKEN_FILE`](#consul_http_token_file) environment variable.
```

website/content/docs/security/acl/acl-system.mdx
```diff
@@ -2,72 +2,57 @@
...
-  Connect](/docs/connect). At authorization time this acts like an
-  additional policy was attached, the contents of which are described further
-  below. These are directly attached to tokens and roles and are not
-  independently configured. (Added in Consul 1.5.0)
+Tokens are the core method of authentication in Consul. The `SecretID` of the token
+(often referred to as the ACL token) is an opaque string that is included in requests to
+identify the person or system making the request. The ACL system will look up the ACL
+token and grant or deny access based on the permissions associated with the token.
+Permissions are granted to a token by linking it to [policies](#policies), [roles](#roles),
+[service identities](#service-identities), and [node identities](#node-identities).
```
blake, line 40:
```suggestion
[service identities](#service-identities), or [node identities](#node-identities).
```

website/content/docs/security/acl/acl-system.mdx
```diff
@@ -2,72 +2,57 @@
...
+from a trusted external system.
 
-ACL tokens, policies, roles, auth methods, and binding rules are managed by
-Consul operators via Consul's [ACL API](/api/acl/acl),
-[ACL CLI](/commands/acl), or systems like
-[HashiCorp's Vault](https://www.vaultproject.io/docs/secrets/consul).
+Consul HTTP API requests accept a token from an [HTTP Header](/api-docs/index#authentication),
+and CLI requests accept a token from
+[`-token` command line flag](/commands#authentication) or
+[`CONSUL_HTTP_TOKEN_FILE` environment variable](/commands#consul_http_token_file).
```
blake, line 52:
```suggestion
[`CONSUL_HTTP_TOKEN` environment variable](/commands#consul_http_token).
```
It might be better to reference this env var since it is equivalent to `-token`.

website/content/docs/security/acl/acl-system.mdx
blake, line 55:
```suggestion
## Policies ((#acl-policies))
```
This should preserve the old anchor name so that existing links still point to this content, while also generating a new #policies anchor that can be used for new content going forward.

trujillo-adam:
@dnephin - I went way outside of the scope of this branch because it is difficult to understand the concepts without touching and rephrasing the rest of the content. I hope this doesn't mess us up too much. 
I am still struggling to understand the effects of configuring service/node identities in roles, so I was hoping you could fill in some of the gaps there. 
I didn't completely remove the index file from the ACL folder, but I updated the sidebar to reflect the new structure.
I have not implemented any redirects, yet, either. 
I'm sure that additional work has been done on this docs since the original branch was created, so I'm sure there will be merge conflicts. 
Based on my understanding, all ACL-related concepts seemed to stem from policies, tokens, and roles, so I organized the content to fit into that paradigm. 

https://consul-e04m4fqn5-hashicorp.vercel.app/docs/security/acl/acl-system

trujillo-adam:
I updated the service and node identities information in the latest commit.

website/content/docs/security/acl/acl-system.mdx
```diff
@@ -1,429 +1,90 @@
...
-| ------------------------------------------------------------------------------------ | ------------------------------------------------------------------------ |
-| [`acl.tokens.agent_recovery`](/docs/agent/options#acl_tokens_agent_recovery)         | [`acl.tokens.agent_master`](/docs/agent/options#acl_tokens_agent_master) |
-| [`acl.tokens.initial_management`](/docs/agent/options#acl_tokens_initial_management) | [`acl.tokens.master`](/docs/agent/options#acl_tokens_master)             |
-
-#### ACL Agent Recovery Token
-
-Since the [`acl.tokens.agent_recovery`](/docs/agent/options#acl_tokens_agent_recovery) is designed to be used when the Consul servers are not available, its policy is managed locally on the agent and does not need to have a token defined on the Consul servers via the ACL API. Once set, it implicitly has the following policy associated with it
-
-In Consul 1.4 - 1.10, this was called the `agent_master` token. It was renamed to `agent_recovery` token in Consul 1.11.
+## Service Identities
```
dnephin, line 70: I think it might be better to remove "Service Identities" and "Node identities" from this overview page.  These two things are convenience policy templates (not roles). They are much less central to the ACL system (when compared to tokens, policies, and roles), and by including them on this overview I think we might be distracting from the core conceptions.

I think introducing them on the policies page might be better.

website/content/docs/security/acl/acl-roles.mdx
```diff
@@ -0,0 +1,378 @@
...
+
+- `ID`: The `ID` is an auto-generated public identifier. You can specify the role `ID` when linking it to tokens.
+- `Name`: A unique meaningful name for the role. You can specify the role `Name` when linking it to tokens.
+- `Description`: (Optional) A human-readable description of the role.
+- `Policies`: Specifies a the list of policies that are applicable for the role. The object can reference the policy `ID` or `Name` attribute.
+- `ServiceIdentities`: Specifies a list of services that are applicable for the role. See [Service Identities](#service-identities) for details.
+- `NodeIdentities`: Specifies a list of nodes that are applicable for the role. See [Node Identities](#node-identities) for details.
+- `Namespace`: <EnterpriseAlert inline /> The namespace this policy resides within. Roles can only be linked to policies that are defined in the same namespace and admin partition. See [Namespaces](/docs/enterprise/namespaces) and [Admin Partitions](/docs/enterprise/admin-partitions) for additional information. Requires Consul Enterprise 1.7.0+.
+
+## Service Identities
```
dnephin, line 69: Related to the other suggestion about these "x identities", these are more about policies, not roles, so I think these sections (Service Identities and Node Identities) should be part on the policies page.

website/content/docs/security/acl/acl-roles.mdx
```diff
@@ -0,0 +1,378 @@
+---
+layout: docs
+page_title: Roles
+description: >-
+  This topic describes roles within the access control list (ACL) system. A role is a named set of policies and service identities.
+  They are intended to
+---
+
+# Roles
```
dnephin, line 9: If the Node/Service Identities are moved to the policies page, and the attributes are removed, I think there ends up being very little left here. I suspect Roles are not really a "core" concept either. They are a minor concept for convenience. Maybe they would be better off as a sub-section on the Policies or Tokens page, instead of their own page.

website/content/docs/security/acl/acl-policies.mdx
```diff
@@ -0,0 +1,435 @@
...
+Name:         my-app-policy
+Description:  Human-readable description of my policy
+Datacenters:  <specified with the -valid-datacenter option>
+Rules:
+<rules defined in the rules.hcl file>
+```
+
+You can can define several attributes that attach additional metadata and specify the scope of the policy. See [Policy Attributes](#policy-attributes) for details.
+
+### HTTP API Endpoint
```
dnephin, line 319: I still think it's better to link to the API docs, than duplicate the API docs here.   Including examples on the API docs page would be great.

website/content/docs/security/acl/acl-policies.mdx
```diff
@@ -0,0 +1,435 @@
...
+
+### Linking Policies to Tokens
+
+A policy that has been implemented must still be linked to a token before the policy has an effect. A service or agent presents the token when interacting with resources in the network. The ACL system processes evaluate the policies linked to the token to determine if the requester has access to the requested resource.
+
+The person responsible for administrating ACLs can use the command line or call the API endpoint to link policies to tokens. Tokens can also be generated dynamically from an external system using Consul's [auth methods](/docs/security/acl/auth-methods) functionality.
+
+Refer to the [tokens documentation](/docs/security/acl/acl-tokens), as well as the [ACL tutorial](https://learn.hashicorp.com/tutorials/consul/access-control-setup-production#create-the-agent-token), for details about creating and linking policies to tokens.
+
+## Policy Attributes
```
dnephin, line 368: I still think document the attributes of a policy is a mistake. These are internal implementation details that could change. We already have the public "attributes" documented in two places, once in the API docs reference, and again in the CLI reference. This third copy make the docs more difficult to maintain, and doesn't really help understanding the concepts.  My suggestion would be to delete this section, and move any important information to one of the existing references.

website/content/docs/security/acl/acl-roles.mdx
```diff
@@ -0,0 +1,378 @@
...
+
+Make a `PUT` call to the `acl/role` endpoint and specify the role configuration in the payload to create roles. You can save the role definition in a JSON file or use escaped JSON in the call. In the following example call, the payload is defined externally.
+
+```shell-session
+$ curl -X PUT --data @payload.json http://127.0.0.1:8500/v1/acl/role
+```
+
+Refer to the [API documentation](/api-docs/acl/roles) for details.
+
+## Role Attributes
```
dnephin, line 57: similar opinion about this not being the right place to document role attributes.

website/content/docs/security/acl/acl-tokens.mdx
```diff
@@ -0,0 +1,248 @@
...
+
+Specify the token in the HTTP `X-Consul-Token` header field to make an API request. Refer to the [HTTP API documentation](/api-docs/index#authentication) for details.
+
+The following example shows the header for a GET request to the `agent/members` endpoint.
+
+```shell-session
+$ curl --header "X-Consul-Token: <token>" "http://127.0.0.1:8500/v1/agent/members"
+```
+
+## Token Attributes
```
dnephin, line 136: same comment about token attributes

website/content/docs/security/acl/acl-tokens.mdx
```diff
@@ -0,0 +1,248 @@
...
+}
+```
+
+</CodeBlockConfig>
+
+</CodeTabs>
+
+Refer to the [service definitions documentation](/docs/discovery/services#service-definition) for additional information.
+
+### Agent Requests
```
dnephin, line 69: Is this a request or is this configuration?

My thinking is that we should link to the configuration docs for acls, instead of duplicating the config reference here.

dnephin (commented):
Nice! This is looking great. 

Mostly suggestions for things that I think we should delete since we already document them in other places. We can always link to the single source of truth instead of trying to maintain the docs in two (or more) places.

Clicking a link to find the information shouldn't be a big inconvenience, and by keeping things in a single place we increase the chances of our docs being consistent and correct. By duplicating them we increase the chance of them being inconsistent and causing additional confusion.

I know when I read docs, if I see two similar things on two different pages I end up wasting a bunch of time trying to figure out why these two things are duplicated and if there is any difference. I think we can save our readers some time here.

trujillo-adam:
Closing this PR and continuing the work in PR #12460

PR #11996, opened 2022-01-10
docs: remove acl config section from acl docs

dnephin:
Branched from #11993, so I'll be looking to merge that PR first.

These config settings are already documented in the config reference. Instead of duplicating them, we can link to the from the ACL docs. The table removed in this PR included columns for server vs client, but the one case that claimed it wasn't necessarily on clients was wrong. `default_policy` is also used the ACLResolver on clients. So I've left those details out of the config reference.

Also adds a bit more detail to the acl.enabled docs, and fixes the headers so that the configuration reference appears in the table of contents dropdown, and so that sub-headers  are always 1 below their parent (not H2 to H4).


blake (approved):

trujillo-adam:
Closing this as a dup of PR #12460

PR #9416, opened 2020-12-16
Document Kube CRDs alongside config entry docs

lkysow:
* Adds a new react component `ConfigEntryReference` that allows us to
    document config entries for both HCL and Kubernetes CRDs.
* Rewrites config entry docs to use new component
* Add CRD examples alongside HCL examples
* Removes examples from Kube CRD doc because the examples are now in the
    main CRD docs

Notes to reviewers:
* I performed a thorough visual inspection across all config entries to ensure the HCL docs have not changed since I had to copy and paste so much text
* I still need to document how namespaces work in Kubernetes, I will do this in a separate PR.

PR #12001, opened 2022-01-10
docs: fix exported-services to use ConfigEntryReference block

rboyer:


trujillo-adam (changes requested):
I think this is a step backward. We should be moving away from lists for this type of content because they are more difficult for readers to process.
What is the reason for this change?

rboyer:
It basically reverts the k8s-parity added by https://github.com/hashicorp/consul/pull/9416 which allowed a single config entry kind doc to work for the consul json/hcl format as well as the slightly different k8s yaml formats without a lot of duplication of efforts. Unfortunately this component only generates list forms, but the inputs are structured data so in theory the component could be updated to switch them all to a table + child specific sub-tables.

karl-cardenas-coding:
@trujillo-adam and @rboyer were you two able to come to an agreement on next steps for this PR?

jmurret:

PR #12011, opened 2022-01-10
docs: update config entry docs for proxy-defaults to follow new template

rboyer:


website/content/docs/connect/config-entries/proxy-defaults.mdx
```diff
@@ -12,51 +12,138 @@ description: >-
...
+
+## Requirements
+
+- A Consul binary
+
+## Usage
 
-Set the default protocol for all sidecar proxies:
+1. Verify that your datacenter meets the conditions specified in the [Requirements](#requirements).
+1. Apply the configuration using one of the following methods:
```
rboyer, line 37: I believe the same blob from `exported-services` needs to be adjusted like this to make sense. The usage is telling you the ways you can create it, but on the `exported-services` page it's presented like the agent config file bootstrap method is somehow step 2 instead of one of the 3 options for applying the config entry.

trujillo-adam (commented):
Thanks for working on this - it looks really good. I think adding these standard sections will provide useful signposts for readers. 
I made a few suggestions that I think we should implement across all CEs.
I have one more ask. Replace "config" with "configuration" throughout the topic (except for cases where the `config` is a keyword).

website/content/docs/connect/config-entries/proxy-defaults.mdx
rboyer: A configuration entry is a structured blob of data that has to eventually end up in the state store to do its job. It has (amusingly) nothing to do with the consul agent configuration file per-se other than the config entry bootstrap special agent configuration section (which folks should practically only be using for demos and toy environments).

The primary way to get a configuration entry into the state store is via an API call to the `v1/config` endpoint. This is the mechanism used by both `consul config write` and a k8s CRD under the covers. These all fall into the same bucket of usefulness (as would using `terraform` to do the same).

Since all of the primary and related ways to get configuration entries into the state store require a server to be up and a leader to be elected for bootstrapping small environments or demos it can be annoying to deterministically wait until things _can_ accept a write before doing your initial configuration entries writes.

For those scenarios we have the `config_entries { bootstrap = [ ... ] }` https://www.consul.io/docs/agent/options#config_entries_bootstrap **agent** configuration section that only applies to servers. This is applied every time _this_ agent becomes an elected leader and the data is re-updated into the database. For scenarios where you have more than one server (like production) it should not be used because if you have slightly different things in the configuration files for those different server agents simply switching which one is leader would cause the configuration entries to flip flop back and forth.

rboyer: To clarify from the exported-services page:
```
1. Verify that your datacenter meets the conditions specified in the Requirements.
2. Specify the exported-services configuration in the agent configuration file (see config_entries) as described in Configuration.
3. Apply the configuration using one of the following methods:
  - Kubernetes CRD: Refer to the Custom Resource Definitions documentation for details.
  - Issue the consul config write command: Refer to the Consul Config Write documentation for details.
```

(2) does not mean "come up with the struct you want to save" which would go quite well as a prereq to (3) "save the thing you just came up with"

(2) is "save the thing one way" and (3) is "save the thing two other ways"

website/content/docs/connect/config-entries/proxy-defaults.mdx
```diff
@@ -9,54 +9,141 @@ description: >-
...
+* Consul 1.5.0+ on other platforms. 
+
+## Usage
 
-Set the default protocol for all sidecar proxies:
+1. Verify that your datacenter meets the conditions specified in the [Requirements](#requirements).
+1. Apply the configuration using one of the following methods:
+   - Specify the `proxy-defaults` configuration in the agent configuration file (see [`config_entries`](/docs/agent/options#config_entries)) as described in [Configuration](#configuration).
+   - Kubernetes CRD: Refer to the [Custom Resource Definitions](/docs/k8s/crds) documentation for details.
+   - Issue the `consul config write` command: Refer to the [Consul Config Write](/commands/config/write) documentation for details.
```
trujillo-adam, line 40: I believe this is how we want to phrase this based on our external discussion:
```suggestion
1. Verify that your datacenter meets the conditions specified in the [Requirements](#requirements).
1. Determine the settings you want to implement (see [Configuration](#configuration)). You can create a file containing the configuration or pass them to the state store directly to apply the configuration.
1. Apply the configuration using one of the following methods:
   - Kubernetes CRD: Refer to the [Custom Resource Definitions](/docs/k8s/crds) documentation for details.
   - Issue the `consul config write` command: Refer to the [Consul Config Write](/commands/config/write) documentation for details.
```

trujillo-adam (approved):
Again, this looks really good. I made one final suggestion for the usage section that you can use or revise as you see fit.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/560988.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit b9e9f1106bc0316d02962205b4cac14d953616f8 onto `stable-website` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit b9e9f1106bc0316d02962205b4cac14d953616f8 onto `release/1.11.x` succeeded!

PR #12013, opened 2022-01-10
Update proxy-defaults.mdx

stallion01:
Adding a note to clarify that without proxy-defaults configured, service to service comms between DCs is not possible over MGW.
event/consul-docs-day

### DIFF
--- a/website/content/docs/connect/config-entries/proxy-defaults.mdx
+++ b/website/content/docs/connect/config-entries/proxy-defaults.mdx
@@ -14,7 +14,7 @@ description: >-
 
 The `proxy-defaults` config entry kind (`ProxyDefaults` on Kubernetes) allows for configuring global config
 defaults across all services for Connect proxy configuration. Currently, only
-one global entry is supported.
+one global entry is supported. This is a requirement for service to service communication between datacenters over mesh gateways.
 
 ## Sample Config Entries
 


website/content/docs/connect/config-entries/proxy-defaults.mdx
```diff
@@ -14,7 +14,7 @@ description: >-
 
 The `proxy-defaults` config entry kind (`ProxyDefaults` on Kubernetes) allows for configuring global config
 defaults across all services for Connect proxy configuration. Currently, only
-one global entry is supported.
+one global entry is supported. This is a requirement for service to service communication between datacenters over mesh gateways.
```
blake, line 17: <https://www.consul.io/docs/k8s/installation/multi-cluster/kubernetes#proxydefaults> states that users must create a proxy defaults entry with `meshGateway.mode=local` in order to use mesh gateways for service mesh traffic.

<https://www.consul.io/docs/connect/gateways/mesh-gateway/service-to-service-traffic-datacenters#connect-proxy-configuration> also states that the mesh gateway mode must be configured using one of the supported methods in order to use mesh gateways.

Given that proxy defaults can exist in a mesh when federation is not in use, I'm not sure it makes sense to add this sentence here.

If users often overlook the requirement to specify the mesh gateway mode, then perhaps a more explicit callout should be added to [Connect Proxy Configuration](https://www.consul.io/docs/connect/gateways/mesh-gateway/service-to-service-traffic-datacenters#connect-proxy-configuration) which draws the reader's attention, and emphasizes the need for this configuration.

PR #12015, opened 2022-01-10
Update schedule condition with table and add type column

mkam:
Converted the last condition configuration to the table format and add the type column to all tables.

I'm planning a follow up PR to standardize the tables across the different pages.

### DIFF
--- a/website/content/docs/nia/configuration.mdx
+++ b/website/content/docs/nia/configuration.mdx
@@ -337,7 +337,9 @@ task {
 }
 ```
 
-- `cron` - (string: required) The CRON expression that dictates the schedule to trigger the task. For more information on CRON expressions, see the [cronexpr parsing library](https://github.com/hashicorp/cronexpr).
+| Parameter | Required | Type | Description | Default |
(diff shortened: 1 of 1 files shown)


website/content/docs/nia/configuration.mdx
```diff
@@ -113,15 +113,15 @@ service {
...
+| Parameter | Required | Type | Description | Default |
+| --------- | -------- | ---- | ----------- | ------- |
+| `name` |  Required | string | Consul logical name of the service. | none |
+| `id` |  Optional | String | Service ID for CTS. This is used to explicitly identify the service config for a task to use. If no ID is provided, the service is identified by the service name within a [task definition](#task). | none |
```
lornasong, line 119: small nit - looks like type here is capitalized "String" whereas other types are lower cased

lornasong (approved):
Looks good to me. I like the new types column!

website/content/docs/nia/configuration.mdx
mkam: Good catch, thanks!

PR #12018, opened 2022-01-11
Ma/resource listing

markan:
This is an experiment on how to represent what capabilties are controlled by a given resource and access level.


website/content/partials/endpoints_acl_write.mdx
```diff
@@ -0,0 +1,19 @@
+| Endpoint Action      | API Endpoint           | Access Level   | Note   |
+| -------------------- | ---------------------- | -------------- | ------ |
+Create an Auth Method|[[PUT] /v1/acl/auth-method](https://www.consul.io/api-docs/acl/auth-methods#create-an-auth-method)|acl:write|
```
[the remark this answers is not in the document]
jkirschner-hashicorp: We could even try both approaches on the same page so it's easier to compare.

website/content/docs/security/acl/acl-system.mdx
```diff
@@ -323,13 +323,16 @@ rules:
...
 | [`event`](/docs/acl/acl-rules#event-rules)          | Listing and firing events in the [Event API](/api/event)                                                                                                                                                 |
 | [`key`](/docs/acl/acl-rules#key-value-rules)        | Key/value store operations in the [KV Store API](/api/kv)                                                                                                                                                |
-| [`keyring`](/docs/acl/acl-rules#keyring-rules)      | Keyring operations in the [Keyring API](/api/operator/keyring)                                                                                                                                           |
+| [`keyring`](/docs/acl/acl-rules#keyring-rules)      | Keyring operations in the [Keyring API](/api/operator/keyring)
+| [`mesh`](/docs/acl/acl-rules#mesh-rules)            | Access to ingress gateways, terminating gateways and mesh configuration entries [Config API](/api/config)
```
jkirschner-hashicorp, line 327: What do we have that uses `mesh` rules? I feel like I haven't seen that in the documentation. For example, if I go to the config HTTP API page, it doesn't show any `mesh` rules at all... are these just not documented yet?

(And `mesh` is obviously missing from this experiment so far because it's not in the source spreadsheet at all)

website/content/docs/security/acl/acl-endpoints-by-resource.mdx
```diff
@@ -0,0 +1,222 @@
...
+</Tab>
+
+</Tabs>
+
+## Key
+Write includes Read permissions
+
+<Tabs>
+
+<Tab heading="Read">
```
jkirschner-hashicorp, line 93: It looks like the special `list` policy disposition only exists for the `key` resource (see https://www.consul.io/docs/security/acl/acl-rules#policy-dispositions).

If I understand correctly, `list` is just a recursive version of `read`, so it would have the same privileges as `read. Maybe the table heading can be "Read or List"?

jkirschner-hashicorp (commented):
@markan, @luoxuan00733: left some ideas/comments/suggestions. Let me know what you think. Happy to pair with you on pushing this forward.

website/content/docs/security/acl/acl-endpoints-by-resource.mdx
```diff
@@ -0,0 +1,222 @@
...
+layout: docs
+page_title: Consul API Endpoint Access List
+description: |-
+  Consul API Access list based on the action and access level
+---
+
+# API Endpoint Access List
+
+Below are lists of API endpoints organized by the resource and access level that controls. Write access also provides read access.
+
```
jkirschner-hashicorp, line 11: @markan : we should add a section for which endpoints are accessible without need for any ACL permissions

website/content/partials/API_ENDPOINTS.tsv
```diff
@@ -0,0 +1,184 @@
+Name	API Endpoint	API Method	API Docs	Equiv Command	Command URL	ACL Required	Info Revealed with No Token	Additional Notes
```
jkirschner-hashicorp, line 1: @markan, @luoxuan00733 : we could potentially even link to this .tsv file if someone wanted access to a raw spreadsheet (rather than using our interface) as a sort of workaround to meet their own needs. Not sure this is useful, just wanted to write it down as an idea to consider before I forgot.

markan: After working with this for a bit, I honestly am not sure if it's valuable to commit this or not; it's a pain to edit and read. Wondering if some sort of more structured format (JSON anyone?) would be more useful.

jmurret:

PR #12019, opened 2022-01-11
docs: Describe token locality for auth methods

eculver:
## Summary

I took a stab at adding a "Token Locality" section to the Auth Method docs. I hope this makes sense, but the topic is rather nuanced and consequently, I found difficult to describe.

## Details
By default, auth methods create locally scoped tokens. Auth Methods can be configured to create globally scoped tokens, but there are a number of nuanced restrictions around the behavior. This PR attempts to document such things by adding a "Token Locality" section to the root Auth Method docs and then linking to the specifics in the docs about individual Auth Methods. 

_Note_: I think I got this right, but I am honestly not sure if I covered everything. If there are suggestions as to how to describe the "alternative" approach using the anonymous token, please share; this is one part of the change I was least sure of.

## TODO
- [ ] update index to introduce `-token-locality` flag as [Adam suggested](https://github.com/hashicorp/consul/pull/12019#discussion_r794790424)
- [ ] apply any/all suggestion about Kubernetes docs
- [ ] link to anonymous token docs
- [ ] add part about anonymous token that Luke mentioned [here](https://github.com/hashicorp/consul/pull/12019/files#r793071882)
- [ ] add [Luke’s suggestion](https://github.com/hashicorp/consul/pull/12019/files#r793073040) to k8s docs


dnephin (commented):
Thanks for writing these docs! I'm hoping to add a "Token Locality" section to the ACL Tokens docs as well. Once that section exists we can probably link to it from these docs.

Left one suggestion for clarification below.

website/content/docs/security/acl/auth-methods/index.mdx
```diff
@@ -136,3 +136,26 @@ bearer token for a Consul ACL token by using the login process:
...
+given Auth Method being used for the details related to token locality:
+
+- [`kubernetes`](/docs/acl/auth-methods/kubernetes#token-locality)
+- [`jwt`](/docs/acl/auth-methods/jwt#token-locality)
+- [`oidc`](/docs/acl/auth-methods/oidc#token-locality)
+
+For cross-datacenter requests that use a token provided by an Auth Method, one
+of the following must be true for requests to be successfully authenticated:
+- the Auth Method is configured with [`-token-locality=global`](/commands/acl/auth-method/create#token-locality)
+- the anonymous token policy in the remote datacenter grants access to read services
```
dnephin, line 156: I think using a local token for the target DC is also an option, but I'm not sure if that works for all workflows.

I know we use the anonymous token "workaround" in consul-k8s, but I'm not sure if that's something we want to document as a supported way of doing things. 

If we do want to keep the "anonymous token" docs here, are there other docs we can link to that explain how to update the policy for the anonymous token?

website/content/docs/security/acl/auth-methods/kubernetes.mdx
```diff
@@ -158,3 +158,13 @@ use in binding rule selectors and bind name interpolation.
...
 | `serviceaccount.name`      | Equal, Not Equal, In, Not In, Matches, Not Matches | yes                 |
 | `serviceaccount.uid`       | Equal, Not Equal, In, Not In, Matches, Not Matches | yes                 |
+
+## Token Locality
+
+By default, tokens created by the Kubernetes Auth Method will have local scope
+and will therefore only be valid for the datacenter in which they were created.
+For cases requiring global token scope, the primary Consul datacenter must have
+access to all Kubernetes clusters in order for tokens to be validated. This
+property is often prohibitive so using [`-token-locality=global`](/commands/acl/auth-method/create#token-locality)
```
dnephin, line 168: I don't think I follow why Consul would need access to all kube clusters. Couldn't it create global tokens from the one cluster used by the primary DC? I guess that assumes that the primary cluster contains all the users, but maybe that's a fine assumption sometimes?

website/content/docs/security/acl/auth-methods/jwt.mdx
```diff
@@ -172,3 +172,12 @@ If multiple methods are needed, another auth method of this type may be created
...
 @include 'jwt_claim_mapping_details.mdx'
+
+## Token Locality
+
+By default, tokens created by the JWT Auth Method will have local scope and will
+therefore only be valid for the datacenter in which they were created. When the
+Auth Method is created with [`-token-locality=global`](/commands/acl/auth-method/create#token-locality),
+all tokens will be replicated to all other datacenters. Cross-datacenter requests
+that are authenticated with tokens provided by this Auth Method must be configured
+to create tokens with global scope in order for requests to be validated.
```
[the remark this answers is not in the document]
eculver: > I think there's one small detail that is missing here (and the same applies to the other auth method pages) , which is that `global` is only supported in the primary DC ([ref](https://github.com/hashicorp/consul/blob/v1.11.1/agent/consul/acl_endpoint.go#L438-L440)).

Good call. I will call this out.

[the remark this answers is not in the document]
eculver: > Either a global token, or a local token in the target datacenter, is required to perform a cross-datacenter request.

This is also a good point. I didn't think about the case where the token was neither local to the source datacenter _nor_ a globally scoped token. I think this update covers this case.

website/content/docs/security/acl/auth-methods/jwt.mdx
```diff
@@ -172,3 +172,12 @@ If multiple methods are needed, another auth method of this type may be created
 with a different name.
 
 @include 'jwt_claim_mapping_details.mdx'
+
+## Token Locality
+
+By default, tokens created by the JWT Auth Method will have local scope and will
+therefore only be valid for the datacenter in which they were created.
+
+In the primary DC, the Auth Method can be created with [`-token-locality=global`](/commands/acl/auth-method/create#token-locality). With globaly token locality, the tokens will be replicated to all secondary datacenters.
```
eculver, line 181:
```suggestion
In the primary DC, the Auth Method can be created with [`-token-locality=global`](/commands/acl/auth-method/create#token-locality). All tokens with global locality will be replicated to all secondary datacenters.
```

eculver: @dnephin I made one tweak to your suggestion.

website/content/docs/security/acl/auth-methods/kubernetes.mdx
eculver: @mkeeler could you chime in here? I was basing this aspect on the comment [you made](https://hashicorp.slack.com/archives/C0253EQ5B40/p1629138964222900?thread_ts=1629133907.212800&cid=C0253EQ5B40): 
>  the caveat with global tokens from authmethods is that its not always possible to configure it that way. For example if you are using the k8s auth method and the primary dc doesn’t have access to every one of your k8s clusters. Then you cannot use global tokens there and may have to resort to the anonymous token approach.

mkeeler: The k8s auth method regardless of which DC its in and what locality you assign **must** have access to the k8s APIs of the cluster that is issuing the services JWT tokens that will be used for Consul's ACL login flow.

With that comment I was trying to say that you can only rely on just having k8s auth-methods that generate global tokens (in the primary dc) if all the k8s clusters you want to allow logins from can be accessed from the primary DC. 

If your primary dc doesn't have access to all the associated clusters then you must configure the anon token to allow those disconnected k8s cluster to login and get acl tokens.

> Couldn't it create global tokens from the one cluster used by the primary DC?

One k8s cluster cannot validate the JWT tokens generated for services in other k8s clusters. We need a separate auth-method for every cluster which means network/API access to each cluster from wherever the auth method is defined.

dnephin: Thank you for clarifying!

If I understand this correctly, there is nothing preventing someone from setting up a k8s auth method in the primary DC, and configuring that auth method to produce global tokens.

The only requirement is that the JWT they pass in comes from that one k8s cluster configured in the auth method. 

There's no requirement that the consul primary DC have access to all k8s clusters, if the JWTs are only ever coming from 1 cluster.

Does that sound right?

Would it be more correct to focus on the requirement of JWTs to Auth methods to network access?

My concern is that statements that claim "Consul must have access to all k8s clusters" is misleading. That's only true if you want a JWT to come from any cluster. If you have setup all service accounts in a single k8s cluster, then you only need an auth method for that one. But you do need to make sure the JWT all come from that one.

eculver: ok, I _think_ I have addressed this in some meaningful way with the latest revision. Please let me know what you think 🙏🏼

website/content/docs/security/acl/auth-methods/index.mdx
```diff
@@ -136,3 +136,30 @@ bearer token for a Consul ACL token by using the login process:
...
 For more details about specific auth methods and how to configure them, click
 on the name of the auth method type in the sidebar.
+
+# Token Locality
+
+By default, Auth Methods create ACL tokens with _local_ scope. ACL tokens with
+local scope are only valid in the local datacenter where they were created.
+Auth Methods created in the primary datacenter can be configured to create ACL
+tokens that have global scope; however, there are limitations depending on the
+type of the Auth Method being used. For all all methods, setting
```
lkysow, line 146:
```suggestion
type of the Auth Method being used. For all auth methods, setting
```

website/content/docs/security/acl/auth-methods/index.mdx
lkysow: I think more accurately you'd have to say that the anonymous token policy has the permissions needed for the request you're making (because in cross-dc calls, the local token is removed from the call and so when it arrives in the other dc it will be authenticated as an anonymous call)

website/content/docs/security/acl/auth-methods/index.mdx
```diff
@@ -136,3 +136,30 @@ bearer token for a Consul ACL token by using the login process:
...
+[`-token-locality=global`](/commands/acl/auth-method/create#token-locality) may
+not be possible. Alternatively, the operator can configure the Auth Method and
+associated Binding Rules manually in all datacenters as a workaround.
+
+For specific Auth Methods, please refer to the associated documentation for
+details related to token locality:
+
+- [`kubernetes`](/docs/acl/auth-methods/kubernetes#token-locality)
+- [`jwt`](/docs/acl/auth-methods/jwt#token-locality)
+- [`oidc`](/docs/acl/auth-methods/oidc#token-locality)
```
lkysow, line 165: maybe we could put a note here that the Kubernetes integration handles configuring all auth methods when https://www.consul.io/docs/k8s/helm#v-global-acls-managesystemacls is set to true

website/content/docs/security/acl/auth-methods/kubernetes.mdx
```diff
@@ -158,3 +158,20 @@ use in binding rule selectors and bind name interpolation.
 | `serviceaccount.namespace` | Equal, Not Equal, In, Not In, Matches, Not Matches | yes                 |
 | `serviceaccount.name`      | Equal, Not Equal, In, Not In, Matches, Not Matches | yes                 |
 | `serviceaccount.uid`       | Equal, Not Equal, In, Not In, Matches, Not Matches | yes                 |
+
+## Token Locality
```
lkysow, line 162: Should add a note here that when using the Kubernetes Helm chart's https://www.consul.io/docs/k8s/helm#v-global-acls-managesystemacls that the locality is not configurable.

lkysow (approved):
LGTM from kube perspective

website/content/docs/security/acl/auth-methods/index.mdx
dnephin: I think that's a great suggestion, but maybe that note should be on the kubernetes auth method page instead of this index page?

website/content/docs/security/acl/auth-methods/jwt.mdx
```diff
@@ -172,3 +172,12 @@ If multiple methods are needed, another auth method of this type may be created
...
 @include 'jwt_claim_mapping_details.mdx'
+
+## Token Locality
+
+By default, tokens created by the JWT Auth Method will have local scope and will
+therefore only be valid for the datacenter in which they were created.
+
+In the primary DC, the Auth Method can be created with [`-token-locality=global`](/commands/acl/auth-method/create#token-locality). All tokens with global locality will be replicated to all secondary datacenters.
+
+Either a global token, or a local token in the target datacenter, is required to perform a cross-datacenter request.
```
dnephin, line 183:
```suggestion
Either a global token, or a local token from the target datacenter, is required to perform a cross-datacenter request.
```

I'm not sure what others thing, but to me "from" reads a bit better here than "in".

website/content/docs/security/acl/auth-methods/kubernetes.mdx
```diff
@@ -158,3 +158,20 @@ use in binding rule selectors and bind name interpolation.
...
+
+By default, tokens created by the Kubernetes Auth Method will have local scope
+and will therefore only be valid for the datacenter in which they were created.
+On the other hand, by using [`-token-locality=global`](/commands/acl/auth-method/create#token-locality),
+tokens created by the Auth Method will be replicated to all other datacenters.
+What this means is that by using `global` locality, there are implications on
+the Consul servers' access to associated Kubernetes clusters. This may come as
+a limitation to the given Consul deployment so using
```
dnephin, line 170: Based on our discussion yesterday, I think we might need to adjust this a bit. Maybe something about along the lines of:
*  generally the kubernetes auth method is only intended to create tokens for services, not for humans
* each k8s auth method can only validate JWT tokens from a single k8s cluster
* since an auth method with token-locality=global can only be created in the primary DC, the primary DC would  need a k8s auth method for each k8s cluster in every dc , and Consul servers in  the primary DC would need direct network access to each of them
* this is often an undesirable setup, so configuring a policy on the anonymous token may be an easier way to allow cross-DC requests in kuberentes

Probably with a call out before all of this as Luke mentioned to say that anyone using consul-k8s with managedACLs does not need to care about this because it's handled for them.

Not sure if this helps, you probably already have better notes on this.

dnephin (commented):
Looks good! I'll +1 Luke's suggestions, and left some minor suggestions for edits

website/content/docs/security/acl/auth-methods/index.mdx
```diff
@@ -136,3 +136,30 @@ bearer token for a Consul ACL token by using the login process:
...
+
+# Token Locality
+
+By default, Auth Methods create ACL tokens with _local_ scope. ACL tokens with
+local scope are only valid in the local datacenter where they were created.
+Auth Methods created in the primary datacenter can be configured to create ACL
+tokens that have global scope; however, there are limitations depending on the
+type of the Auth Method being used. For all auth methods, setting
+[`-token-locality=global`](/commands/acl/auth-method/create#token-locality) for
+requests to secondary datacenters will fail.
```
trujillo-adam, line 148: Is the assumption that readers have encountered this concept? We seemed to jump from introduction to a specific implementation. It might be a good a idea to start a new paragraph that opens with something like: "You can specify authentication method scope using the `-token-locality` flag. You can set the value to either `global` or `local`. Setting to `global` configures tokens created in the datacenter to provide global access to the resource. Setting to `local` configures tokens created in the local datacenter to provide access to resources within the datacenter. For all authentication methods, setting `-token-locality` to `global` . . . ". 

I may not be stating the information accurately and I know that we'd be repeating a bit of information provided in the CLI section, but I think it would provide additional context to help readers at this point in the docs without having to jump around.    
```suggestion
type of the Auth Method being used. For all auth methods, setting
[`-token-locality=global`](/commands/acl/auth-method/create#token-locality) for
requests to secondary datacenters will fail. 
```

website/content/docs/security/acl/auth-methods/index.mdx
```diff
@@ -136,3 +136,30 @@ bearer token for a Consul ACL token by using the login process:
...
+
+For cross-datacenter requests that use a token provided by an Auth Method, one
+of the following must be true for requests to be successfully authenticated:
+- the Auth Method is configured with [`-token-locality=global`](/commands/acl/auth-method/create#token-locality)
+- the anonymous token policy in the remote datacenter grants access to read services
+
+Depending on the type of Auth Method being used, configuring it with
+[`-token-locality=global`](/commands/acl/auth-method/create#token-locality) may
+not be possible. Alternatively, the operator can configure the Auth Method and
+associated Binding Rules manually in all datacenters as a workaround.
```
trujillo-adam, line 158:
```suggestion
associated binding rules manually in all datacenters as a workaround.
```
As stated above - unless there's a compelling reason, I don't think we should treat auth(entication) methods as proper nouns, either.

website/content/docs/security/acl/auth-methods/jwt.mdx
```diff
@@ -172,3 +172,12 @@ If multiple methods are needed, another auth method of this type may be created
 with a different name.
 
 @include 'jwt_claim_mapping_details.mdx'
+
+## Token Locality
+
+By default, tokens created by the JWT Auth Method will have local scope and will
+therefore only be valid for the datacenter in which they were created.
```
trujillo-adam, line 179:
```suggestion
By default, tokens created by the JWT authentication method are scoped to the local datacenter in which they were created.
```
Simple present tense

website/content/docs/security/acl/auth-methods/oidc.mdx
```diff
@@ -212,3 +212,12 @@ be tricky to debug why things aren't working. Some tips for setting up OIDC:
   contain sensitive information, this option should not be used in production.
 
 @include 'jwt_claim_mapping_details.mdx'
+
+## Token Locality
+
+By default, tokens created by the OIDC Auth Method will have local scope and will
+therefore only be valid for the datacenter in which they were created. When the
```
trujillo-adam, line 219:
```suggestion
By default, tokens created by the OIDC Auth Method are scoped to the local datacenter in which they were created. When the
```
same as the suggestion in the JWT section

website/content/docs/security/acl/auth-methods/oidc.mdx
```diff
@@ -212,3 +212,12 @@ be tricky to debug why things aren't working. Some tips for setting up OIDC:
...
+Auth Method is created with [`-token-locality=global`](/commands/acl/auth-method/create#token-locality),
+all tokens will be replicated to all other datacenters. Cross-datacenter requests
```
trujillo-adam, line 221:
```suggestion
all tokens will be replicated to all other datacenters. Cross-datacenter requests
```
For clarity, future perfect tense is acceptable in this sentence because the action actually will take place later during the replication process.

trujillo-adam (commented):
Added some questions, comments, and feedback. It sounds like some of the details are still being hashed out in the Kubernetes section, so I didn't look too closely at that part.

website/content/docs/security/acl/auth-methods/index.mdx
eculver: yeah, see your point. I like these suggestions. I will update.

PR #12029, opened 2022-01-11
nia/docs: Standardize CTS config/parameter tables

mkam:
These tables should now have the same format across all our docs. The format was based off of what we have in the configuration tables.

- Standardize the Required field to have values of either Optional/Required rather than true/false
- Standardize the column order to `Name | Required | Type | Description | Default`
- Standardize string entries with no default values to read `none` rather than an empty string


lornasong (approved):
thanks for standardizing all of these! looks good to me

Issue #10531, opened 2021-06-30
Allow read/update/delete Prepared Query by name


engel75:
#### Feature Description

Currently the UUID is needed to read, update or delete a prepared query. This is quite unflexible as the name has to be unique as well. So it looks like there is no reason to use force the user to use the uuid.
It is even more strange as for execute7explain the uuid AND the name are valid parameters:

https://www.consul.io/api/query

Be allowing the query name for read/update/delete as well the API will become much more user friendly.

#### Use Case(s)

deploy, debugging, ...

jkirschner-hashicorp:
Hi @engel75,

Thank you for sharing with us this suggestion to enhance the API.

To help us think through this suggestion, can you tell us more about:

- What is the behavior you are hoping for if you provide a query name (instead of UUID) for `read/update/delete`?
  - e.g., affect/match a single query? or multiple?
  - e.g., exact match? or longest prefix match (like the `explain` and `execute` endpoints do currently)?
    - longest prefix match means that if you have a query template with name `abc`, then call the execute endpoint for name `abc-something`, the `abc` query template will be executed if queries have a longer prefix match (e.g., `abc-some`, `abc-something`)
- Can you tell me more about the scenario you encountered that inspired this suggestion? What are the steps you took to accomplish what you needed (because you couldn't read/update/delete by name)?

This information will help us better understand the context and consider approaches. Thank you for your contribution!

jkirschner-hashicorp:
Hi @engel75 - we'd appreciate your perspective on the questions above when you have a chance. We want to make sure we understand where you were coming from with this suggestion before we consider possible approaches. Thanks again!

engel75:
Hi @jkirschner-hashicorp,

the main problem, in my case, is, I do not know the UUID before the query was created. This is a problem if I try to manage my queries with, eg. ansible.

So how to manage those queries if the UUID is needed to read/update/delete them? As "name" is unique, why can't we use "name" to manage them?

jkirschner-hashicorp:
It sounds like you're looking to match to single queries on an exact name match for one read/update/delete call, rather than match multiple queries at once or perform a non-exact match.

I can see how it would be difficult to manage based on the UUID in your case. Have you figured out a workaround that does what you need for now, even though it's not ideal?

You are right that a non-empty name could be used to manage queries. That said, there are a few things we need to think through:
[...]

PR #12053, opened 2022-01-13
Allow reading/updating/deleting a PreparedQuery by exact name

jamestran201:
Related to: #10531

## Background context

When a PreparedQuery is created and managed by a tool such as Ansible, its ID is not known ahead of time. This can make it difficult for the user to manage this query [in their scripts](https://github.com/hashicorp/consul/issues/10531#issuecomment-882480893). For this use case, the user currently manages these queries [manually](https://github.com/hashicorp/consul/issues/10531#issuecomment-882550258).

The issue above proposes adding endpoints to read/update/delete a query using its exact name to help the user with this use case, see [this comment](https://github.com/hashicorp/consul/issues/10531#issuecomment-895598803) for more details.

## Changes in this PR

This PR adds the ability to get a query using its exact name. The implementation is adapted from the code that retrieves a query by its ID.

The current changes in this PR:
- Add a REST endpoint to retrieve the query using its exact name. The endpoint path is `/v1/query/exact-name/{query-name}`.
- Add a RPC endpoint to retrieve the query using its exact name. This RPC endpoint is called `PreparedQuery.GetByExactName`.
- Add a function to retrieve the query from the state store using its name.
- Add a function to the `api` package to call the new endpoint
- Add API doc for `GET` endpoint

## To do
- Implement endpoint to update a query given its name
- Implement endpoint to delete a query given its name

## Questions

- ~~The implementation for reading a query given its name has a lot of duplication with reading the query by its ID. Should the common code be extracted?~~ Answered by https://github.com/hashicorp/consul/pull/12053#discussion_r795790995
- ~~Should the code for updating and deleting a query be implemented in this PR or separate ones?~~ Answered by https://github.com/hashicorp/consul/pull/12053#discussion_r795790995


kisunji:
As for your questions, 
1: We're fine with duplicate code for now; we prefer not to prematurely refactor code until necessary (to avoid coupling code that may diverge later - we might see some of that with ACL auth here)
2: This PR is fine; update/delete can come after we've reviewed the ACL implications in this PR.

Thank you for your patience!

FFMMM:
Hey @tmnhat2001 thanks for your patience here.

Our inclination is that we probably want to have ACL permissions checking for prepared query read when getting a prepared query by name but we want to do more research into this.

agent/consul/prepared_query_endpoint.go
```diff
@@ -267,6 +267,43 @@ func (p *PreparedQuery) Get(args *structs.PreparedQuerySpecificRequest,
 		})
 }
 
+// GetByExactName returns a single prepared query by Name.
+func (p *PreparedQuery) GetByExactName(args *structs.PreparedQuerySpecificRequest,
+	reply *structs.IndexedPreparedQueries) error {
+	if done, err := p.srv.ForwardRPC("PreparedQuery.GetByExactName", args, reply); done {
+		return err
+	}
+
```
markan, line 276: So I think we want to check for PreparedQueryRead permission here.

The setup logic for the auth could look similar to this:
https://github.com/hashicorp/consul/blob/8a6e75ac815a5bdfd9fdbd3528c985d5fdde9c1b/agent/consul/acl_endpoint.go#L1411

jamestran201:
@markan @kisunji Sorry for the late reply 🙇‍♂️ . I recently changed jobs so haven't been able to work on this. Thank you for you guidance with ACL! I will start implementing it.

agent/consul/prepared_query_endpoint.go
jamestran201: Thank you for the example 👍 . I attempted to check the PreparedQueryRead permission in https://github.com/hashicorp/consul/pull/12053/commits/a08da226c1c4bb835821e5e3b49c812e9eb75d34. Please review it again when you have time.

PR #12059, opened 2022-01-13
remove api, sdk, and ui packages from dependabot

dnephin:
sdk and api are removed because those two are libraries, and updating libraries to the latest
dependency versions only serves to make the library harder to use. A
library which uses all the latest versions requires every application
which uses it to update all its dependencies in a single batch, which is
not great.
It is applications that need the latest version of dependencies.

Remove UI at the request as jc, since javascript libraries are updated so frequently it ends up
being too much noise.:wq


kisunji (approved):
Do we want to exclude `website` as well?

dnephin:
Good question, I'll leave that up to the team that manages the website. I'll ask in slack, but for now I'm thinking we could leave it.

Do we want to increase the number of open PRs from 5 to 8 or more?

dnephin:
PR only impacts dependabot config, and that check has passed, so going to merge.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/552334.

PR #12080, opened 2022-01-14
streaming: split event buffer by key

boxofrad:
## Summary

This is an improvement to Consul's streaming backend to address performance problems observed in cases of high parallelism - i.e. many subscribers, high service churn, and servers with a large number of CPU cores.

It reduces lock-contention in the kernel's futex subsystem (more on that below) by only waking subscribers when there is an update that is relevant to them.

## Background

Events are published to subscribers by appending them to a data structure called the `eventBuffer`. It is a single-writer, multiple-reader, linked list in which each element contains a channel that is used to notify subscribers when the subsequent element is available (similar to a [monitor/condition variable](https://en.wikipedia.org/wiki/Monitor_(synchronization))). Subscribers block on a receive from this channel until it is closed by the publisher — this pattern will be familiar if you've used the [context](https://pkg.go.dev/context) package (particularly the `Done` method).

Currently, there is a single `eventBuffer` for the entire topic. In other words, subscribers will be woken whenever _any_ health event occurs, regardless of whether or not it pertains to a service they're watching, and it's up to them to ignore events that they're not interested in.

On a busy cluster with many services and open streams, when an event is published, the **majority** of scheduled goroutines have no real work to do, and will immediately attempt to receive/block on the next element’s channel.

As well as the general overhead of scheduling these goroutines and spending a few CPU cycles on checking for relevant events, there's a much larger cost incurred by them all attempting to become receivers on the new channel at the same time.

Internally, channels maintain a [lock](https://github.com/golang/go/blob/0d0193409492b96881be6407ad50123e3557fdfb/src/runtime/chan.go#L45-L51) that is held at various points, including when adding a blocked receiver; therefore, the many subscriber goroutines will compete to acquire the lock, resulting in contention.

There is some additional nuance here too, though. It turns out that reducing the contention of this particular lock - e.g. by sharding the same large number of subscribers across many channels so they're not all competing for the same lock - doesn't actually solve the problem. 🤔

On Linux, the Go runtime's internal locks are backed by [futexes](https://en.wikipedia.org/wiki/Futex). The kernel stores futexes in a [hashmap](https://github.com/torvalds/linux/blob/fb3b0673b7d5b477ed104949450cd511337ba3c6/kernel/futex/core.c#L1126-L1130) in which each bucket is guarded by a [spinlock](https://en.wikipedia.org/wiki/Spinlock). The contention we're observing doesn't go away when we shard the channel because it isn't contention on the individual channel's lock, but rather the kernel's own internal (semi-global) lock! This explains why we have observed such a large amount of CPU time spent in `native_queued_spin_lock_slowpath`, and why the graph below shows such a high proportion of CPU time spent in kernel-space.

The downstream impact of this contention is that event publishing gets slower, and with it so does applying changes to the Raft state machine, which has a massive effect on the general performance of a server. As an example, sustained FSM Apply latency of just 20ms, when Raft's internal queues/buffers are full, can result in several **minutes** of RPC latency.

By splitting the `topicBuffer` by key (e.g. service name) we can limit the number of goroutines that are woken by an event to just those with "real" work to do — which likely also adds an element of jitter as the goroutine will need to wait on sending gRPC stream frames etc.

Naturally, this doesn't improve the pathological case where all subscribers are watching a single service. But in testing, I've found that this case doesn't exhibit the same issue, because the subscribers always have "real" work to do, and so won't immediately contend to re-acquire the channel lock.

Huge kudos to @banks, @mkeeler, @dnephin et al. for all of the incredible research work on this issue 👏🏻 ⭐

## Benchmark

To test the efficacy of this change, I reused the test harness from when we last compared the performance of the streaming backend to that of blocking queries.

I deployed a single server node on an `m6i.32xlarge` EC2 instance (128 CPU cores). I then ran the benchmark tool on 50 worker nodes, each creating 14 gRPC connections to the server and consuming 100 streams per connection (for a total of **70k concurrent subscriptions**). The tool then began generating write load by updating random tags on the watched services at a rate of 50Hz.

### Results

The results below show a significant reduction in "system" CPU time, lower FSM Apply time, and crucially less time spent waiting in Raft's internal queues/buffers (leading to pathologically bad RPC latency).

Before and after graphs


agent/consul/stream/event.go
```diff
@@ -101,6 +89,14 @@ func (p *PayloadEvents) HasReadPermission(authz acl.Authorizer) bool {
 	})
 }
 
+// TopicKey is required to satisfy the Payload interface but is not implemented
+// by PayloadEvents. PayloadEvents structs are constructed by Subscription.Next
+// *after* TopicKey has been used to dispatch the enclosed events to the correct
+// buffer.
```
banks, line 95: 🤔 I wouldn't block merging this because of it, but I wonder if this is a sign that our Interface abstraction isn't expressing quite the right thing any more.

I wonder if it would be cleaner to have two interfaces - one for any event that is deliverable `Payload` and another for not-yet-routed events that embeds `Payload` but adds the `TopicKey` method, maybe `RoutablePayload` (I know, names). That way we don't need strange No-op implementations and we also encode the intent of each type in the type system - these types are really only for use internally once events have been routed to a buffer etc..

@dnephin what do you think as you designed the original interface here?

agent/consul/stream/event_publisher.go
```diff
@@ -116,34 +124,74 @@ func (e *EventPublisher) Run(ctx context.Context) {
...
-		e.getTopicBuffer(topic).Append(events)
+	for topic, byKey := range eventsByTopic {
+		for key, events := range byKey {
+			// Note: bufferForPublishing returns nil if there are no subscribers for the
+			// given topic and key, in which case events will be dropped on the floor and
+			// future subscribers will catch up by consuming the snapshot.
+			if buf := e.bufferForPublishing(topic, key); buf != nil {
+				buf.Append(events)
+			}
+		}
```
banks, line 154: Note: IF it were possible to publish thousands of events in one call to Publish in the future, it might be worth avoiding the inner per-key loop here by checking there are even any subscribers at all for the topic first.

I think that would be a premature optimisation for now because:
 - We only ever publish a small number of events right now in practice.
 - Even with that check, in most clusters there is likely to be at least a few subscribers to every topic which means we have to loop by key anyway so we can't short circuit this inner loop.

tl;dr this is fine :)

agent/consul/stream/event_publisher.go
```diff
@@ -163,7 +211,28 @@ func (e *EventPublisher) Subscribe(req *SubscribeRequest) (*Subscription, error)
...
+		defer e.lock.Unlock()
+
+		topicBuf.refs--
+
+		if topicBuf.refs == 0 {
+			delete(e.topicBuffers[req.Topic], req.TopicKey())
+
+			// Evict cached snapshot too because the topic buffer will have been spliced
+			// onto it. If we don't do this, any new subscribers started before the cache
+			// TTL is reached will get "stuck" waiting on the old buffer.
```
banks, line 230: nice catch ;)

agent/consul/stream/event_publisher.go
```diff
@@ -197,11 +266,17 @@ func (e *EventPublisher) Subscribe(req *SubscribeRequest) (*Subscription, error)
...
 	result.buffer.AppendItem(snapFromCache.First)
-	return e.subscriptions.add(req, result.First), nil
+	return e.subscriptions.add(req, result.First, freeBuf), nil
 }
 
-func (s *subscriptions) add(req *SubscribeRequest, head *bufferItem) *Subscription {
-	sub := newSubscription(*req, head, s.unsubscribe(req))
+func (s *subscriptions) add(req *SubscribeRequest, head *bufferItem, freeBuf func()) *Subscription {
+	// We wrap freeBuf in a sync.Once as it's expected that Subscription.unsub is
+	// idempotent, but freeBuf decrements the reference counter on every call.
```
banks, line 274: Good thinking!

agent/consul/stream/event_publisher.go
```diff
@@ -233,19 +308,17 @@ func (s *subscriptions) closeSubscriptionsForTokens(tokenSecretIDs []string) {
 // This function is returned as a closure so that the caller doesn't need to keep
 // track of the SubscriptionRequest, and can not accidentally call unsubscribe with the
 // wrong pointer.
```
banks, line 310: This comment is incorrect now right?

agent/consul/stream/event_publisher.go
```diff
@@ -197,11 +266,17 @@ func (e *EventPublisher) Subscribe(req *SubscribeRequest) (*Subscription, error)
...
 
-func (s *subscriptions) add(req *SubscribeRequest, head *bufferItem) *Subscription {
-	sub := newSubscription(*req, head, s.unsubscribe(req))
+func (s *subscriptions) add(req *SubscribeRequest, head *bufferItem, freeBuf func()) *Subscription {
+	// We wrap freeBuf in a sync.Once as it's expected that Subscription.unsub is
+	// idempotent, but freeBuf decrements the reference counter on every call.
+	var once sync.Once
+	sub := newSubscription(*req, head, func() {
+		s.unsubscribe(req)
+		once.Do(freeBuf)
```
banks, line 278: I stopped to think for a minute about whether it matters that these two calls each take the lock separately. It means they are not atomic. i.e. there is a point in time where the reference count of subscriptions is higher that the sum of all the subscriptions in the subscription maps.

Assuming the code can't panic or exit in some way before the freeBuf call, I _think_ this is OK eventually, it just might mean we defer garbage collecting for a while after the subscription has been removed because we are waiting for the lock a second time.

 The sync.Once ensures it can't happen multiple times though so it should still always be correct eventually.

An alternative approach might be to document that `freeBuf` is called with the lock held, and then not take it inside the closure above, and then pass it in to `unsubscribe()` to be executed while the lock is held. That has a marginal performance gain under contention for not needing to take and release the lock twice in a single unsubscribe but i don't think that's important. I think this would have the other minor benefit of not needing the sync.Once. That is saving an allocation but it's not in a critical path so the main benefit would be simpler code.

What do you think?  I don't _think_ it matter from a correctness perspective though it's a little harder to reason about this way.

agent/consul/stream/event_publisher_test.go
```diff
@@ -56,6 +56,13 @@ func TestEventPublisher_SubscribeWithIndex0(t *testing.T) {
 		Payload: simplePayload{key: "sub-key", value: "the-published-event-payload"},
 	}
 	require.Equal(t, expected, next)
+
+	// Subscriber should not see events for other keys
+	publisher.Publish([]Event{{
+		Topic:   testTopic,
+		Payload: simplePayload{key: "other-key", value: "this-should-not-reach-the-subscriber"},
+	}})
+	assertNoResult(t, eventCh)
```
banks, line 65: The previous test that was removed had cases for all the permutations of namespace/name/partition etc. too. I don't think that necessary here provided we have an exhaustive test of the `TopicKey` matching method that is being used elsewhere. I'll assume we do but I haven't got to it yet!

agent/consul/stream/event_publisher_test.go
```diff
@@ -190,16 +194,18 @@ func TestEventPublisher_SubscribeWithIndex0_FromCache(t *testing.T) {
 
 	publisher := NewEventPublisher(newTestSnapshotHandlers(), time.Second)
 	go publisher.Run(ctx)
+
 	sub, err := publisher.Subscribe(req)
 	require.NoError(t, err)
-	sub.Unsubscribe()
+	defer sub.Unsubscribe()
```
banks, line 200: I'm curious about why this needed to change as it changes the behaviour of the test. Was it a bug before? How did it pass if so?

banks: Hmm I think I see. this subscription was just for the side-effect of priming the snapshot cache for the TopicKey. So it doesn't matter if it ends now or later because the real test here is that the second subscription hits the warm cache right?

But now it _must_ end later otherwise the immediate Unsubscribe will garbage collect the buffer and snapshot and so prevent the next subscription from hitting a cache as desired.

Cool!

agent/consul/stream/event_publisher_test.go
```diff
@@ -233,7 +239,20 @@ func TestEventPublisher_SubscribeWithIndexNotZero_CanResume(t *testing.T) {
...
+	require.NoError(t, err)
+	defer sub.Unsubscribe()
+
+	// Publish the testSnapshotEvent, to ensure that it is skipped over when
+	// splicing the topic buffer onto the snapshot.
+	//
+	// Note: we must evict the snapshot that was created by the above subscription
+	// first, as it has already had the topic buffer spliced onto it (so therefore
+	// would include the snapshot event again).
+	delete(publisher.snapCache[req.Topic], req.TopicKey())
```
banks, line 255: This delete is racey and would fail the race detector right? I think you'd need to hold the publisher's lock if you want to play with its internals here. Ideally all our tests should pass with race detector enabled, but especially this package with it's creative concurrency patterns!

agent/consul/stream/event_publisher_test.go
```diff
@@ -338,7 +357,20 @@ func TestEventPublisher_SubscribeWithIndexNotZero_NewSnapshotFromCache(t *testin
...
```
banks, line 373: ditto. won't point out the others :)

agent/consul/stream/event_publisher_test.go
```diff
@@ -495,3 +540,36 @@ func TestEventPublisher_Unsubscribe_ClosesSubscription(t *testing.T) {
...
+	require.NotNil(t, publisher.topicBuffers[req.Topic][req.TopicKey()])
+	require.NotNil(t, publisher.snapCache[req.Topic][req.TopicKey()])
+
+	// Close the other subscription and expect the buffer and snapshot to have
+	// been cleaned up.
+	sub2.Unsubscribe()
+
+	require.Nil(t, publisher.topicBuffers[req.Topic][req.TopicKey()])
+	require.Nil(t, publisher.snapCache[req.Topic][req.TopicKey()])
+}
```
banks, line 575: All of these reads from the internal maps should be done holding the lock to please the race detector right?

agent/consul/stream/topic_key.go
```diff
@@ -0,0 +1,36 @@
...
+	partition string
+}
+
+// NewTopicKey constructs a TopicKey for the given {key, namespace, partition}
+// combination. The given values will be normalized - converted to lower-case
+// with empty namespace/partition treated as the default namespace/partition -
+// to ensure equality when using the TopicKey as a map key.
+func NewTopicKey(key, namespace, partition string) TopicKey {
+	if namespace == "" {
+		namespace = "default"
```
banks, line 24: I forget where we landed on hard coding the string `default` vs it being a constant somewhere. I know at one point it was only a constant in some awkward places and I don't think it's worth taking a dependency on some huge irrelevant package for that, but perhaps we fixed that and have a clean import we could use rather than hard coding this in another place. @dnephin probably knows if there is a better alternative!

proto/pbsubscribe/subscribe.proto
```diff
@@ -51,9 +51,8 @@ message SubscribeRequest {
 
     // Key is a topic-specific identifier that restricts the scope of the
     // subscription to only events pertaining to that identifier. For example,
-    // to receive events for a single service, the service's name is
-    // specified as the key. An empty key indicates that all events in the topic
-    // are of interest.
+    // to receive events for a single service, the service's name is specified
+    // as the key.
```
banks, line 55: Nice update. Just a note to say that this is a change to the original design where we hoped we'd be able to do things like efficiently watch the entire catalog for changes by subscribing to the whole topic with no key.

In practice now they are split by key this is no longer possible but it's not a code path we actually expose anywhere today.

In the future if we require that, the likely path would be to allow `TopicKey` to hold explicit wildcard sentinel values in some or all parts of the key such that you could subscribe to the key `partition1/*/*` and get all update for a partition for example (which in OSS means the whole state store). That would require changing the matching logic in publish a little but otherwise seems like a reasonable approach as and when it's needed.

banks (changes requested):
Amazing work @boxofrad 👏 

I left a few inline comments, mostly notes to myself or random thoughts that don't need changes. There's one or two small things I think it would be nice to change though including fixing the racey map accesses in tests that presumably fail race detector?

The only other one that seems important to consider is whether we can simplify the code (and reasoning about correctness although I _think_ it is correct) and have the bookkeeping function happen atomically with the unsubscribe state being updated.

proto/pbsubscribe/subscribe.proto
kisunji: Does this mean `stream.SubscribeRequest` needs to be updated to enforce non-blank keys as well?

agent/consul/stream/topic_key.go
```diff
@@ -0,0 +1,36 @@
...
+// a map key when storing/retrieving event buffers and cached snapshots.
+//
+// Note: its fields are unexported as it's expected that users will call
+// NewTopicKey rather than constructing a TopicKey manually, as this method
+// normalizes the given values.
+type TopicKey struct {
+	key       string
+	namespace string
+	partition string
+}
```
dnephin, line 16: I think we might be better off with an opaque `TopicKey` similar to `Topic`. Nothing in the `stream` package should need to know about these individual fields. We could do something like:

```go
type TopicKey string
```

or similar to topic

```go
type TopicKey fmt.Stringer
```

agent/consul/stream/event.go
dnephin: I agree that having to implement methods on `Payload` types that will never have those methods called is unfortunate.  If I remember correctly that problem already existed when I extracted this interface, but extracting the Payload interface did make it more obvious.

I'd probably call it `KeyedPayload` (instead of `RoutablePayload`). If we were to do that I guess we have two options, either:
1. different types for the `[]Event` passed to `Publish` , and the event type we store internally and return to subscribers (ex: KeyedEvent vs Event), or
2. type-assert the `Payload` to `KeyedPayload` on Publish. 

I don't think we want to type-assert because we have no way of returning an error from `Publish`.

I think having a separate `Event` type to contain the different Payload types might be fine, but I'm not sure it's a big enough problem to expand the API the package in this way. If we were to do the suggestion I left below, and use a `type TopicKey string` then returning `""` from framing events doesn't seem too bad. A separate event type might be ok too.

Are there other options I've missed?

agent/consul/stream/event_publisher_test.go
dnephin: The `stream` package does run with the race detector enabled in `go-test-race`. So if this is a race we should be able to reproduce with a test in that package.

agent/consul/stream/topic_key.go
dnephin: I don't think the `stream` package should know anything about the default namespace. This defaulting should have already happened long before we get here, because it must happen either when we accept a subscription or before we publish an event.  I think `TopicKey` should be an opaque identifier.

proto/pbsubscribe/subscribe.proto
dnephin: I think f29769eb77473e056475e83878da52ddd2c2c0dc may do this.

agent/consul/stream/event_publisher.go
```diff
@@ -23,13 +23,13 @@ type EventPublisher struct {
...
-	// topicBuffers stores the head of the linked-list buffer to publish events to
+	// topicBuffers stores the head of the linked-list buffers to publish events to
 	// for a topic.
-	topicBuffers map[Topic]*eventBuffer
+	topicBuffers map[Topic]map[TopicKey]*eventBuffer
 
 	// snapCache if a cache of EventSnapshots indexed by topic and key.
 	// TODO(streaming): new snapshotCache struct for snapCache and snapCacheTTL
-	snapCache map[Topic]map[string]*eventSnapshot
+	snapCache map[Topic]map[TopicKey]*eventSnapshot
```
dnephin, line 32: Originally I was thinking that `TopicKey` would include both the Topic and the Key values, so we'd have `map[TopicKey]T`. 

I haven't thought about it too much. Do you think there are particular advantages to a two-level map over a single map? 

I thought maybe a single `map[TopicKey]T` might have been easier to manage, but maybe not.

agent/consul/stream/event.go
```diff
@@ -26,18 +26,14 @@ type Event struct {
...
-	// returned to every subscription.
-	MatchesKey(key, namespace, partition string) bool
-
 	// HasReadPermission uses the acl.Authorizer to determine if the items in the
 	// Payload are visible to the request. It returns true if the payload is
 	// authorized for Read, otherwise returns false.
 	HasReadPermission(authz acl.Authorizer) bool
+
+	// TopicKey is used to identify which subscribers should be notified of this
+	// event - e.g. those subscribing to health events for a particular service.
```
dnephin, line 35: minor: I think it would be helpful to include what kind of values we generally expect. Ex: for services this is the full service ID, for KV it would be the full key path, etc.  I think that might help readers understand this is simply an opque identifier for the data in the Payload, not some stream-specific tracking ID.

dnephin (commented):
Great work, and excellent detail in the description!

I've only done a quick review so far. Mostly responding to the questions and commenting on stylistic things like `TopicKey`. I'll hopefully have a chance to read over more of the functional changes in the last two commits next week.

agent/consul/stream/event_publisher.go
boxofrad: > these two calls each take the lock separately

I think they take separate locks? (`EventPublisher.lock` and `subscriptions.lock`). The point about atomicity still stands, though!

> I think this would have the other minor benefit of not needing the sync.Once.

To be honest, I've sort of taken the assertion that `unsub` needs to be idempotent (from its doc comment) as a given, and haven't taken the time to dig into why that's the case. Wouldn't we still need some way to prevent decrementing the reference counter twice?

> What do you think? I don't think it matter from a correctness perspective though it's a little harder to reason about this way.

Agreed! The dependencies here are a bit hard to grok. That said, without merging `subscriptions` into `EventPublisher`, I can't think of an obvious way to share the lock that isn't equally complex (e.g. passing `freeBuf` to `unsubscribe` as you mentioned).

agent/consul/stream/event_publisher.go
boxofrad: Good point! I think @dnephin's [idea](https://github.com/hashicorp/consul/pull/12080#discussion_r785216841) of including the `Topic` in `TopicKey` and making `topicBufs` a single-dimensional map means that we wouldn't need to think about this.

agent/consul/stream/event_publisher.go
boxofrad: Yes! Good catch 👏🏻

agent/consul/stream/event.go
boxofrad: Great point! 💡

agent/consul/stream/event_publisher_test.go
boxofrad: Bingo 🎰 😄

agent/consul/stream/event_publisher_test.go
boxofrad: Great call out! 🐎

agent/consul/stream/event_publisher_test.go
boxofrad: Yep! Good catch 😅

agent/consul/stream/topic_key.go
boxofrad: Makes sense! It probably makes more sense to do the normalization in `agent/rpc/subscribe` anyway 👍🏻

agent/consul/stream/event.go
boxofrad: +100. It definitely felt like the abstraction was beginning to creak a bit when adding those no-op methods. I'm going to play with this a bit and see if there are any other options 👷🏻

boxofrad: Ok, I've given this a bit more thought. I'm wondering if it feels like more of a problem right now because the streaming API currently only covers health/connect events, and so there are just a lot more framing events than "normal" events.

It seems to me that there are three possible approaches here:

1. Keep things as-is (every `Payload` implements `TopicKey`)
   - **Pro:** Type system makes it impossible to forget to provide a key
   - **Con:** Interface feels a bit peculiar (all these no-op implementations)
   - **Pro:** Reflects the intention that framing events are the special case and most events _will_ have a key
2. Move the `TopicKey` method to another type (e.g. `KeyedPayload`)
   - **Pro:** Interfaces feel a bit cleaner, no no-op implementations
   - **Con:** Weakens type safety. We'd need to do type-assertions and panic or return an error at run-time (maybe generics would help here? 🤷)
3. Make `TopicKey` a struct field on `Event`
   - **Pro:** Similar to option 2 it makes the interfaces clearer
   - **Con:** Also similar to option 2, we can't help implementers to remember the key at compile-time

Overall, I think it might me sense to leave things as-is for now, and revisit in the future if we find ourselves adding lots of other methods to `Payload` that we need to no-op for framing events.

agent/consul/stream/event_publisher.go
banks: > I think they take separate locks? (EventPublisher.lock and subscriptions.lock). The point about atomicity still stands, though!

Ah yes, my mistake. That makes it a bit more complicated 🤔 .

We currently rely on subscription lock being taken while publisher lock is held during `EventPublisher.Subscribe` so to make it atomic we'd need to ensure that `unsubscribe` first took `EventPublisher.Lock` before taking the `subscribers.lock` or it would deadlock. This has a slight potentially increase for contention over this implementation - unsubscribing would hold the `EventPublisher.lock` slightly longer (blocking Publish) since it would also include the subscription updates, but we already hold both for `Subscribe` which is probably more expensive anyway so I don't expect that to be a real issue.

So I think it would still be possible to do what I suggested, just with the addition of taking both locks in `unsubscribe` in the right order... but I'm also fine with leaving it as I can't think of a way it could actually stay in an invalid state right and we don't assume the ref count is perfectly consistent right now. Maybe we should document that the `freeBuf` call is not called under any locks and so its effects are not atomic/consistent with other subscription state?

> Wouldn't we still need some way to prevent decrementing the reference counter twice?

My assumption was that you'd get that for free thanks to the early return in `unsubscribe` if the sub state doesn't exist. If we removed all state atomically that would also mean the buffer state had been cleared too do nothing further needed. But that was under the assumption that it was the same lock being held in `unsubscribe` and `freeBuf` which was wrong.

So yeah I don't think there is anything substantially simpler here - the reason subscriptions have a separate lock anyway is to avoid blocking the publish mechanism for subscription activity.

boxofrad:
Thanks for the reviews folks! 🙇🏻‍♂️ 

@banks, @dnephin: I think I've addressed (or at least responded to) your comments, but do let me know if I've missed anything.

agent/consul/stream/event.go
banks: Fine with leaving it as-is!

agent/consul/state/catalog_events_test.go
```diff
@@ -11,11 +11,69 @@ import (
 	"github.com/hashicorp/consul/agent/consul/stream"
 	"github.com/hashicorp/consul/agent/structs"
 	"github.com/hashicorp/consul/api"
-	"github.com/hashicorp/consul/proto/pbcommon"
 	"github.com/hashicorp/consul/proto/pbsubscribe"
 	"github.com/hashicorp/consul/types"
 )
 
+func TestEventPayloadCheckServiceNode_SubjectMatchesRequests(t *testing.T) {
```
boxofrad, line 18: Reminder for self: add tests for different partition/namespace combinations in the enterprise equivalent of this PR 🏢

banks (approved):
I think all my comments were addressed here so LGTM! I think Daniel said he'd not reviewed in depth yet so I'd wait for an approval from him too!

agent/consul/stream/event_publisher_test.go
```diff
@@ -233,7 +237,22 @@ func TestEventPublisher_SubscribeWithIndexNotZero_CanResume(t *testing.T) {
...
+
+	// Publish the testSnapshotEvent, to ensure that it is skipped over when
+	// splicing the topic buffer onto the snapshot.
+	//
+	// Note: we must evict the snapshot that was created by the above subscription
+	// first, as it has already had the topic buffer spliced onto it (so therefore
+	// would include the snapshot event again).
+	publisher.lock.Lock()
+	delete(publisher.snapCache, req.topicSubject())
+	publisher.lock.Unlock()
```
dnephin, line 255: In normal operation the cache TTL would perform this operation, is that right? So the scenario this is setting up is when the cache has expired, but we still have an active subscribe?  

If that is correct, I wonder if this would be better as a helper function we can call from all 3 tests. The godoc comment on that function would be a great place to document that we are doing this to setup a specific scenario ("a cache has exceeded the TTL, but there are still active subscriptions"). We could use `t.Cleanup` to call the `sub.Unsubscribe` instead of the defer.

dnephin (approved):
Looks great! A couple minor questions and suggestions, but nothing blocking a merge.

agent/consul/stream/subscription_test.go
```diff
@@ -59,10 +81,6 @@ func TestSubscription(t *testing.T) {
 		"Event should have been delivered after short time, took %s", elapsed)
 	require.Equal(t, index, got.Index)
 
-	// Event with wrong key should not be delivered. Deliver a good message right
-	// so we don't have to block test thread forever or cancel func yet.
-	index++
-	publishTestEvent(index, eb, "nope")
```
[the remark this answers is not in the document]
boxofrad: Yeah, this test manually constructs the event buffer, passes in the subscription's head etc. The deleted section was from when we relied on the subscriber filtering out irrelevant events (which is handled by publishing them to a different buffer now).

agent/consul/stream/event_publisher_test.go
boxofrad: Great point! I've swapped this out for a test helper called `simulateExistingSubscriber` which I think captures the intent better (it actually took me a while to remember why I did this 😅).

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/569544.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit fdfe079674710684619c531e247567a7f45425ba onto `release/1.11.x` succeeded!

Issue #12073, opened 2022-01-13
In GUI, unable to create keys within a KV folder (Consul 1.11.x)


jkirschner-hashicorp:
#### Overview of the Issue

The Consul 1.11.x GUI doesn't allow creating keys properly from within a folder.

#### Reproduction Steps

Per @shantanugadgil (see [this comment](https://github.com/hashicorp/consul/issues/10906#issuecomment-1011138333)):

Use Consul version: 1.11.1

1. in the webui, go to Key/Value
2. create key named toplevel/ (with trailing slash)
3. click the new "folder" named "toplevel"
4. click the blue Create button.
5. I expect to see the combined boxes of Key or folder and Value, I see only the Value box. 😮 Add a value and click "Save".
6. Click the blue Create button again. Notice that the Value box contains the value you saved in step 5.
7. Return to Key/Value root, and create a key named toplevel/alpha with value aaa. This works as expected.

In step 3, Consul seems to be treating what I created in step 2 as a folder named "toplevel". However, in steps 5-6, it seems to be treating that same entity as a key named "toplevel/", which could explain why I can edit its value (like it's a key) but not create a new key within the folder (the "Key or folder" name input field is missing). Further evidence: In step 5, when I look at the network tab in Chrome dev tools, I see a put request to `http://localhost:8500/v1/kv/toplevel/?dc=dc1` with the test payload I entered (by clicking "Save") and receive a 200 OK response. 

### Consul info for both Client and Server

Reproduced in Consul 1.11.1.

Not reproduced in Consul 1.10.3+ent. We suspect this first appeared in 1.11.x.

johncowen:
Fix for this in https://github.com/hashicorp/consul/pull/12081

Whilst not great, I noticed that this bug seems to manifest if you create a folder and then go to create something in that folder in one go i.e. without refreshing the page (an ember-data related cache bug). If you refresh the page at anytime after creating the folder, or the folder is a folder that has been created before you came to use the UI, then the bug doesn't seem to manifest. 

i.e. refreshing the page after clicking the create button (when you don't see the input field) makes the input field appear again

Hopefully this info may help any folks how are blocked by this before this goes into a release.

Thanks again for the report!

PR #12081, opened 2022-01-14
ui: Fixup KV folder creation then further creation within that folder

johncowen:
Fixes https://github.com/hashicorp/consul/issues/12073

See ^ for details.

The KV edit/form flow is one of our oldest form flows and it is currently a little bit of a mash up of mostly one of our old ways of doing forms with a tiny bit of cross over into our newer way of doing forms (the idea is soon to most entirely to new style forms the same as our namespace and partitions forms).

The fix here is two fold:

1. We shouldn't be providing the DataSource (which loads the data) with an id when we are creating from within a folder (in the buggy code we are providing the parentKey of the new KV you are creating)
2. Being able to provide an empty id to the DataSource/KV repository and that repository responding with a newly created object is more towards the "new way of doing forms", therefore the corresponding code to return a newly created ember-data object. As we changed the actual bug in point 1 here, we need to make sure the repository responds with an empty object when the request id is empty.

### DIFF
--- a/ui/packages/consul-ui/app/services/repository/kv.js
+++ b/ui/packages/consul-ui/app/services/repository/kv.js
@@ -44,10 +44,19 @@ export default class KvService extends RepositoryService {
         });
       }
     } else {
-      item = await super.findBySlug(...arguments);
+      if (params.id === '') {
+        item = await this.create({
+          Datacenter: params.dc,
+          Namespace: params.ns,
+          Partition: params.partition,
+        });
+      } else {
+        item = await super.findBySlug(...arguments);
+      }
     }
-    // TODO: Whilst KV is using DataForm and DataForm does the model > changeset conversion
-    // a model > changeset conversion is not needed here
+    // TODO: Whilst KV is using DataForm and DataForm does the model >
+    // changeset conversion a model > changeset conversion is not needed here
+    // until we move KV to just use DataWriter like the other new stuff
     return item;
   }
 
--- a/ui/packages/consul-ui/app/templates/dc/kv/edit.hbs
+++ b/ui/packages/consul-ui/app/templates/dc/kv/edit.hbs
@@ -16,7 +16,7 @@ as |parentKey|}}
         partition=route.params.partition
         nspace=route.params.nspace
         dc=route.params.dc
-        key=(if (string-ends-with routeName 'create') parentKey route.params.key)
+        key=(if (string-ends-with routeName 'create') '' route.params.key)
       )
     }}
   as |loader|>
--- a/ui/packages/consul-ui/tests/acceptance/dc/kvs/create.feature
+++ b/ui/packages/consul-ui/tests/acceptance/dc/kvs/create.feature
@@ -49,3 +49,27 @@ Feature: dc / kvs / create
     And I click create
     And I see the text "New Key / Value" in "h1"
     And I see the text "key-value" in "[data-test-breadcrumbs] li:nth-child(2) a"
+    And I see the "[data-test-kv-key]" element
+  Scenario: Clicking create from within a just created folder
+    Given 1 datacenter model with the value "datacenter"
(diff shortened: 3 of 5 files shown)


ui/packages/consul-ui/app/templates/dc/kv/edit.hbs
```diff
@@ -16,7 +16,7 @@ as |parentKey|}}
         partition=route.params.partition
         nspace=route.params.nspace
         dc=route.params.dc
-        key=(if (string-ends-with routeName 'create') parentKey route.params.key)
+        key=(if (string-ends-with routeName 'create') '' route.params.key)
```
johncowen, line 19: This was the actual bug

ui/packages/consul-ui/app/services/repository/kv.js
```diff
@@ -44,10 +44,19 @@ export default class KvService extends RepositoryService {
...
-      item = await super.findBySlug(...arguments);
+      if (params.id === '') {
+        item = await this.create({
+          Datacenter: params.dc,
+          Namespace: params.ns,
+          Partition: params.partition,
+        });
+      } else {
+        item = await super.findBySlug(...arguments);
+      }
```
johncowen, line 55: This bit makes the "pass an empty id to get a new ember-data object" work

ui/packages/consul-ui/tests/acceptance/dc/kvs/create.feature
```diff
@@ -49,3 +49,27 @@ Feature: dc / kvs / create
...
+    ---
+    - key-value/
+    ---
+    And I submit
+    Then the url should be /datacenter/kv
+    And I click "[data-test-kv]"
+    And I click "[data-test-create]"
+    And I see the text "New Key / Value" in "h1"
+    And I see the text "key-value" in "[data-test-breadcrumbs] li:nth-child(2) a"
+    And I see the "[data-test-kv-key]" element
```
johncowen, line 75: Test was mainly copy/pasted from the above (not visible unless you view the entire file). We were already testing: "creating a folder" and "clicking create from within a folder" but not "creating a folder and then immediately clicking create form within that folder".

I also noticed that we weren't asserting for existence of the input field in the "clicking create from within a folder" test that already existed. Whilst this test wasn't failing even with the extra assertion, I thought it would be a good idea to add here to helper prevent any possible regression in the future (line 52)

johncowen (commented):
Notes:

jkirschner-hashicorp:
**This is just a QA test; someone else needs to conduct the front-end engineering review**

As far as I can tell, this fix does address the bug.

I ran this against a locally running Consul client agent (v1.11.x) via:
```
~/go/src/github.com/hashicorp/consul/ui/packages/consul-ui$ CONSUL_HTTP_ADDR=http://localhost:8500 make start-consul
```

With that setup, I can no longer reproduce the bug with the steps described above.

amyrlam (approved):
Code 👍

johncowen:
Thanks for checking that for me @jkirschner-hashicorp, also super glad you found it easy to test the reproduction.

```
~/go/src/github.com/hashicorp/consul/ui/packages/consul-ui$ CONSUL_HTTP_ADDR=http://localhost:8500 make start-consul
```

FYI (and anyone else who may read this) you can just do also:

```
~/go/src/github.com/hashicorp/consul/ui/packages/consul-ui$ make start-consul
```

As localhost:8500 is fairly common I made it so it will just use that by default if you don't set it.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/557877.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit cdb8a35501aee4a04f03f8bbaefb0c2efc8bfb3a onto `release/1.11.x` succeeded!

PR #12097, opened 2022-01-14
Fix the production dockerfile

mdeggies:
Make sure all dependencies are installed in the 'default' image that is built and released using CRT. Dependencies are listed here: https://github.com/hashicorp/docker-consul/blob/master/0.X/Dockerfile#L33. 

This PR also adds some explanatory text to the Dockerfile, and adds a new target called 'official' which includes the binaries downloaded from releases.hashicorp.com. In the future, if there's a dev docker image available, we can add the instructions to the root Dockerfile and build dev as a new target.


Dockerfile
```diff
@@ -1,17 +1,126 @@
-# This Dockerfile creates a production release image for the project using crt release flow.
-FROM alpine:3.13 as  default
+# This Dockerfile contains multiple targets.
+# Use 'docker build --target=<name> .' to build one.
+# e.g. `docker build --target=dev .`
+#
+# All non-dev targets have a VERSION argument that must be provided 
+# via --build-arg=VERSION=<version> when building. 
+# e.g. --build-arg=1.11.2
```
eculver, line 7:
```suggestion
# e.g. --build-arg VERSION=1.11.2
```

mdeggies: ++ thank you

Dockerfile
```diff
@@ -1,17 +1,126 @@
...
+# e.g. `docker build --target=dev .`
+#
+# All non-dev targets have a VERSION argument that must be provided 
+# via --build-arg=VERSION=<version> when building. 
+# e.g. --build-arg VERSION=1.11.2
+#
+# `default` is the production docker image which cannot be built locally. 
+# For local dev and testing purposes, please build and use the `dev` docker image.
+
+
```
eculver, line 12: Should we add a note here or maybe around the `default` target's `FROM` to include setting the `BIN_NAME` build arg as well?

Dockerfile
```diff
@@ -1,17 +1,126 @@
-# This Dockerfile creates a production release image for the project using crt release flow.
-FROM alpine:3.13 as  default
+# This Dockerfile contains multiple targets.
+# Use 'docker build --target=<name> .' to build one.
+# e.g. `docker build --target=dev .`
```
eculver, line 3: This image doesn't have a `dev` target afaict

eculver:
I'm curious if you could explain why the production (aka `default`) container can't be built locally? I was able to build it like this: 
```
docker build --build-arg VERSION=1.11.2 --build-arg BIN_NAME=consul .
```
I was also able to build an `official` image locally as well:
```
 
```

Also, which may be related, it looks like we might have an issue with the `get-product-version` step if we include changes from #12157. Specifically, this change causes the output of `build-support/scripts/version.sh -r` (what `get-product-version` uses to set the version) to output a future release eg. `1.12.0-dev`. We should rebase this from `main` and make sure it doesn't cause any issues. As it stands, if I try to build an `official` image with this version, it will fail:

```
 docker build --target official --build-arg VERSION=1.12.0-dev .
...
#7 3.629 wget: server returned error: HTTP/1.1 403 Forbidden
...
```

eculver:
I'm curious what magic sets the `TARGET` for the docker build? I see it always set to `default` in the build workflow step. Whatever sets it to `official` would need to run with assurances that the associated packages generated from its `VERSION` arg exist ahead of time. Perhaps this gets set as part of the release workflows?

Dockerfile
mdeggies: +1, updating this

mdeggies:
Evan- Thanks for your review! Comments down below:

> I'm curious if you could explain why the production (aka default) container can't be built locally? I was able to build it like this: ...

Building the `default` target is only supported in CI at the moment. The [docker build action](https://github.com/hashicorp/actions-docker-build) that we're using downloads the go binaries that have been built in CI and includes those in the docker image. This is different than what we've done in the past, which is to download the go binaries directly from releases.hashicorp.com after a release is live. (This is what the `official` target is doing and allows customers to build / test using consul releases that are live on releases.hashicorp.com.)

I think the build for target `default` may be working for you because you already have a local build of consul available at `dist/linux/amd64/consul`. It might be useful to create a `dev` target and a Makefile entry that builds a local version of consul, places it in the right location, and includes that in the docker image. We're doing that for boundary, for example, in this PR: https://github.com/hashicorp/boundary/pull/1786/files. I can take a stab at that in this PR if you'd welcome it! 

> Also, which may be related, it looks like we might have an issue with the get-product-version step if we include changes from #12157.

Trying to build the `official` target using a version of consul that is not live on releases.hashicorp.com will always fail with a 403 because the production release is not yet available. I purposely didn't set a default for the version argument for this target because of that -- end users will need to explitily set the version arg in order to build. LMK if the note I included above the target makes sense there. I wrote "Official docker image that includes binaries from releases.hashicorp.com. This downloads the release from releases.hashicorp.com and therefore requires that the release is published before building the Docker image."

> I'm curious what magic sets the TARGET for the docker build?

This is supported in docker: https://docs.docker.com/develop/develop-images/multistage-build/#name-your-build-stages. :) 
We look for existence of the `default` target specifically in the [docker build github action](https://github.com/hashicorp/actions-docker-build)

.github/workflows/build.yml
```diff
@@ -241,6 +241,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - name: Set up QEMU
```
mdeggies, line 244: Note: There's a PR open to update the docker-build action and include this step there.. once that is merged, I'll remove this step and bump the docker build action

eculver:
> I think the build for target default may be working for you because you already have a local build of consul available at `dist/linux/amd64/consul`

Yep, that's right. Thanks for clarifying.

> It might be useful to create a dev target and a Makefile entry that builds a local version of consul, places it in the right location, and includes that in the docker image. We're doing that for boundary ... I can take a stab at that in this PR if you'd welcome it!

The issue of building a local container also came up recently in the context of our integration tests too; we depend on the image in `./build-support/docker/Consul-Dev.dockerfile` to be built to run them. It would be awesome if we had a Makefile target for local testing. I would certainly welcome a PR and could help for this other case for the integration tests. 

>  I purposely didn't set a default for the version argument for this target because of that 

👍🏼  

Thanks for all the clarifications. I think we're good. I'm going to give it another pass and hopefully we can get it merged.

Dockerfile
```diff
@@ -1,17 +1,126 @@
...
+# e.g. --build-arg VERSION=1.11.2
+#
+# `default` is the production docker image which cannot be built locally. 
+# For local dev and testing purposes, please build and use the `dev` docker image.
+
+
+# Official docker image that includes binaries from releases.hashicorp.com. This
+# downloads the release from releases.hashicorp.com and therefore requires that
+# the release is published before building the Docker image.
+FROM alpine:3.13 as official
```
eculver, line 16: Should we be using our mirror for the base image? ie `docker.mirror.hashicorp.services/alpine:3.13.6`

eculver:
Sorry, meant to send this with my original feedback... 

I like the idea of having a `README` for the dockerfile [similar to how Boundary does it](https://github.com/hashicorp/boundary/blob/main/docker/README.md), but we can tackle that in a separate PR that also adds the docker Makefile targets. 

I think all that's left for this is adding the `dev` target? My other comment about the base image isn't blocking.

mdeggies:
Thanks, Evan! I started looking into the consul dev docker image pieces under `build-support/docker`, but they can't be easily ported into the main Dockerfile. I'm going to create a backlog task to look into this further when we have a little more time. Til then, I just pushed the last update to switch to the docker mirror (good catch-thx again!)

eculver (approved):
Thank you so much for getting this over the line. This is a huge assist!

PS- I'm looking into the metrics-checker failure. There appears to be a bug in the script. In the meantime, I'm going to add the `pr/no-metrics-test` label to circumvent the check.

mdeggies:
Just resolved a new merge conflict in the last commit- I'll plan to merge this in after the checks pass

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/576795.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/576830.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit ef46a66492cf3ab64cfbd68e8ed3742ea002720a onto `release/1.11.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/576830)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit ef46a66492cf3ab64cfbd68e8ed3742ea002720a onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/576830)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit ef46a66492cf3ab64cfbd68e8ed3742ea002720a onto `release/1.9.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/576830)

PR #12102, opened 2022-01-15
Perform the right login dance for aws auth

### LINKED ISSUE #11921: Vault AuthMethods for Connect CA
#### Overview of the Issue

We have Vault as our connect CA and are running 1.11.1.  I attempted to use the new AuthMethod settings to use the AWS instance profile to authenticate with vault.  I got an error that I hadn't provided the necessary parameters.  From looking at the code, it appears that the "extra" bits that vault cli does for an aws login are missing and so the only way to use this auth method is to manually construct the identity documents and so on.

Would a PR be welcome to add this?

I'd propose to extend the AWS leg of the current code in a similar way to the kubernetes leg (adding values to the Params map).  And it looks like the approach to determine them is already done here, so could be transplanted over:
https://github.com/hashicorp/vault/blob/a620c2dc8f82b91372521a149ce7e77b97587f9c/builtin/credential/aws/cli.go#L14

#### Reproduction Steps

PUT https://<consul-host>/v1/connect/ca/configuration

{
  "Provider": "vault",
  "Config": {
    "Address": "<vault>,
    "IntermediateCertTTL": "8760h0m0s",
    "IntermediatePKIPath": "pki/consul_connect",
    "RootPKIPath": "pki/root",
    "PrivateKeyType": "rsa",
    "PrivateKeyBits": 4096,
    "AuthMethod": { 
			"Type": "aws", 
			"MountPath": "global_aws", 
			"Params": { 
				"role": "role_consul_server",
				"region": "eu-west-1"
			} 
		}
  }
}

Code: 400. Errors:

* didn't supply required authentication values

mr-miles:
When the vault provider is initialised with the aws auth type, it needs to do the same steps as the vault cli client to achieve a successful login.  This PR enables it to do so.

Fixes #11921


agent/connect/ca/provider_vault.go
```diff
@@ -756,13 +757,54 @@ func configureVaultAuthMethod(authMethod *structs.VaultAuthMethod) (loginPath st
 		} else {
 			return "", fmt.Errorf("failed to get 'role' from auth method params")
 		}
+	// vault provides a lot of help for aws integration
+	// lifted from https://github.com/hashicorp/vault/builtin/credential/aws/cli.go
+	case VaultAuthMethodTypeAWS:
+
+        aws_access_key_id := authMethod.Params["aws_access_key_id"].(string)
+		aws_secret_access_key := authMethod.Params["aws_secret_access_key"].(string)
```
FFMMM, line 765: hm, i think linter will complain here

FFMMM:
hey @mr-miles , thanks for your PR here 💯 !

as a first pass, things LGTM!

What's your plan for testing the changes? Thinking if we could use our mock provider.

mr-miles:
I've made a shim which I think can be swapped out in the tests so the calls to awsutil can provide mock responses.  So next step will be to use it: I think the changes break the current tests for the loginpath anyhow.

I didn't find any prior art on tests/mocks around this part of the login dance, but if you're aware of any then do point me at them!

mr-miles:
Right I think this is finally functional.  Added some new tests and they pass.

FFMMM:
@mr-miles thanks for following up here and for your patience with our response.

It seems that the CircleCI is not running. We think that is because you have a build on your fork. Would you mind turning it off? Something like `Project Settings > Stop Build`

mr-miles:
can you try it again?  i found a web hook in the actions section of settings, which i just deleted.

Amier3:
Hey @mr-miles 

Apologies for the late reply @mr-miles , could you add a dummy commit so we can see if that worked? We forgot to note that the testing suite won't kick off again unless there's another commit ( which made it look like the build was still failing 😅 ) .

mr-miles:
@Amier3 - sorry, this has been v straightforward previously.  i have made some commits to restart things but i cant tell whether it is working: it just says "authorization required to deploy" and i can't find the test suite output.  i'm totally lost :-(

TomTucka:
@Amier3 Any news on getting this merged? We'd really like to use this functionality, Happy to help in anyway I can

ishustava:
Hi @mr-miles  

Thanks so much for the PR! I think this definitely is something we would like to improve so thank you for raising this PR as the first step.

From a deeper look, I see two main problems with merging what you're proposing as is.

First, from what I can tell this change would only support the IAM type for the AWS auth method. However, Vault [supports](https://github.com/hashicorp/vault/blob/3d01a88e614547a821062ca57957b07a1e6bc557/api/auth/aws/aws.go#L94) both EC2 and IAM and if we were to merge this change, we will no longer be able to support. Asking you to add support for EC2 login types wouldn't make sense since it would duplicate code that already exists in a newer vault API library. 

Second, I think we're not properly using the AWS SDK credentials tool chain. For example, we're ignoring creds provided as environment variables. Again, this is something that the newer Vault API library version will do for us: https://github.com/hashicorp/vault/blob/3d01a88e614547a821062ca57957b07a1e6bc557/api/auth/aws/aws.go#L148-L179

Based on that, I think the right approach would be to upgrade to the newer library version. However, because this upgrade will involve updating other dependencies that are not related to vault integration, it will probably be more safe for the Consul team to do this work. 

Since reviewing this PR we've realized that the AWS auth method experience could use some love on our side. So I'll close this PR for the time being but we'll prioritize upgrading the vault API library and clarifying the docs to make it clear what's supported, and how to work around the tricky authentication in the meantime.

Issue #9449, opened 2020-12-21
blocking k/v read on non-existent key returns once any key is written


phemmer:
#### Feature Description

Currently when watching a non-existent key in the key/value store, upon any modification to the store that blocking query returns. This is a request to make it so that watching a non-existent key doesn't return until it is created.

#### Use Case(s)

We have lots (hundreds) of keys which *could* exist, but do not. When they don't exist, the consuming application uses a default value. The consuming application then uses blocking queries to watch the keys, monitoring for changes on the ones that do exist, and creation on the ones that don't.

The problem is that upon any modification to the store, all the blocking queries on the non-existent keys immediately return. Because of this the client then has to fire off a new blocking query for each one. With hundreds of keys, and hundreds of servers, this floods consul with thousands of queries all at once.

&nbsp;

As I expect the first suggestion will be to use recursive watches, we are already are, otherwise it would be even worse. In specific we have a multi-tenant platform, and each tenant gets a prefix in the store. We then have consul-template writing various config files for each tenant, so consul-template is using a recursive watch for a specific "application" prefix under the tenant prefix (e.g. `kv/tenants/1234/app-foo/`). However unless the tenant changes settings, they're left at defaults and thus the keys are not created, resulting in this issue.

blake:
Hi @phemmer, which versions of Consul and consul-template were you using when you saw this behavior?

dnephin:
Thank you for reporting this issue! I believe you are correct that all blocking queries (not just those for KV) are not optimal when the query returns no results.

We will need to explore some options, but I think it may be possible to make blocking queries efficient in this case as well.

PR #12109, opened 2022-01-17
rpc: make blockingQuery easier to read

dnephin:
This is being done to make space for fixing #9449.  Best viewed by individual commit. There is more detail in the commit messages.

The current implementation of `blockingQueries` is difficult to read, and difficult to modify, for a few reasons.

1. the `goto`. While it is safe to use `goto` in Go, most readers will not be used to following logic that uses goto. A for loop is not only much easier to reason about, it removes the need to pre-define a bunch of variables at the top of the function. Sometimes a `goto` is the right solution for complicated logic, but I believe this refactor shows it is not necessary here, and that a for loop reads better.
2. there were many long inline block comments. These comments are almost always better as godoc comments on functions (even if that function is unexported). A large block comment like this breaks the flow of someone trying to read the code. If the flow is going to be broken anyway, the code that the comment is documented may as well be in a separate function. That allows the main function to read much better without losing the flow. I left a couple of the shorter multi-line comments, but I do think we would benefit from reducing or removing those as well.
3. the block and exit logic was difficult to reason about because it was not using guard clauses for early returns. It relied on falling through to the end of the function. By inverting these conditions we can see the flow much better.

The getter methods on `QueryOptionsCompat` and `QueryMetaCompat` also make the code a little harder to read. It'd be great to use a struct field directly, or rename the interface methods to have them read a bit more naturally. Those renames are left for a future commit.

### DIFF
--- a/agent/consul/rpc.go
+++ b/agent/consul/rpc.go
@@ -942,13 +942,10 @@ func (s *Server) blockingQuery(queryOpts structs.QueryOptionsCompat, queryMeta s
 	ctx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
-	// instrument blockingQueries
-	// atomic inc our server's count of in-flight blockingQueries and store the new value
-	queriesBlocking := atomic.AddUint64(&s.queriesBlocking, 1)
-	// atomic dec when we return from blockingQuery()
+	count := atomic.AddUint64(&s.queriesBlocking, 1)
+	metrics.SetGauge([]string{"rpc", "queries_blocking"}, float32(count))
+	// decrement the count when the function returns.
 	defer atomic.AddUint64(&s.queriesBlocking, ^uint64(0))
-	// set the gauge directly to the new value of s.blockingQueries
-	metrics.SetGauge([]string{"rpc", "queries_blocking"}, float32(queriesBlocking))
 
 	for {
 		if queryOpts.GetRequireConsistent() {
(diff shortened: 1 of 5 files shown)


agent/consul/rpc.go
```diff
@@ -1034,6 +1022,17 @@ func (s *Server) setQueryMeta(m structs.QueryMetaCompat, token string) {
...
+	// Always set a non-zero QueryMeta.Index. Generally we expect the
+	// QueryMeta.Index to be set to structs.RaftIndex.ModifyIndex. If the query
+	// returned no results we expect it to be set to the max index of the table,
+	// however we can't guarantee this always happens.
+	// To prevent a client from accidentally performing many non-blocking queries
+	// (which causes lots of unnecessary load), we always set a default value of 1.
+	// This is sufficient to prevent the unnecessary load in most cases.
+	if m.GetIndex() < 1 {
+		m.SetIndex(1)
+	}
```
dnephin, line 1014: Despite the implication by the old comment, this was done for all queries that went through `blockingQuery` , even those with index=0 that were not actually blocking queries.

With this change, we continue to set it for all queries that go through `blockingQuery`, but we will now also set the default response index in a few queries that do their own thing (ex: the Txn operation). That's why I had to modify the expected value of a couple tests below.

agent/consul/rpc.go
```diff
@@ -919,108 +919,74 @@ type queryFn func(memdb.WatchSet, *state.Store) error
...
+		}
+
+		if queryMeta.GetIndex() > minQueryIndex {
+			return nil
+		}
+
+		// block until something changes, or the timeout
+		if err := ws.WatchCtx(ctx); err != nil {
+			// exit if we've reached the timeout, or other cancellation
+			return nil
```
kisunji, line 980: I think there's a difference in codepath: previously,`blockingQuery` would return any non-nil `err` whereas here we suppress it and return `nil` instead.

If that's intentional, do we want a more robust error check here? It looks like we know what errors to expect but if `ws.WatchCtx` ever changes to return other kinds of errors this might lead to unexpected behavior

dnephin: I may be wrong, but I believe the old code would only ever return the `err` from `fn()`, which we still do here on line 970.

In the old code, the error from `WatchCtx`  would only be scoped to that one block, and nothing in that block returned.

That's a good point about future capability though. I was also thinking of improving the godoc for `WatchCtx` so that it doesn't change functionality. I'd like to make the godoc for that method be more explicit about when an error can be returned (only if the context is cancelled), and what errors can be returned (only the context error).

I think if we did that, any change to `WatchCtx` would be a breaking change. We could still guard against it here by checking the error.

kisunji: Ah good point, I missed the shadowed err from `WatchCtx`. I poked at the `WatchCtx` definition and what you said makes good sense! 👍

kisunji (approved):
Looks much more readable and easier to reason about! Thanks for this change

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/567779.

PR #12110, opened 2022-01-17
rpc: make blocking queries for non-existent items more efficient.

### LINKED ISSUE #9449: blocking k/v read on non-existent key returns once any key is written

dnephin:
Branched from #12109, fixes #9449

Full details about this change are in the commit messages. This is a problem that has been reported a few times. It is particularly problematic when connect is enabled, because agents will do many blocking queries for config-entries that often do not exist.

This PR improves the behaviour of blocking queries when the result of the query is not items found (no modify index to use as the `QueryMeta.Index`). This change makes `blockingQuery` more efficient, but there is still possibility for improvement. Currently we have to return on the first iteration if there was a write between the time the previous RPC call was returned and the new one was received (because the table max index would have changed).  We can fix that case by adding a new query option, which I will do in a follow up PR.

TODO:
* [ ] more test cases for state changes between existing->not-existing and not-existing->existing
* [ ] return `errNotFound` from all state store functions where appropriate, so that all queries benefit from this enhancement
* [ ] changelog entry


agent/consul/config_endpoint.go
```diff
@@ -142,7 +142,7 @@ func (c *ConfigEntry) Get(args *structs.ConfigEntryQuery, reply *structs.ConfigE
 
 			reply.Index = index
 			if entry == nil {
-				return nil
+				return errNotFound
```
dnephin, line 145: This is the change that needs to be applied to all the other RPC endpoints to take advantage of this enhancement.  Currently it is only implemented on this one `ConfigEntry.Get` endpoint.

banks:
Nice Daniel! This is a really smart way to save a lot of CPU time in this common case without invasive changes to how we store things!

If I understand correctly, this saves on the CPU overhead of building a response, serializing it, writing to the wire, completing the request, and then accepting a new request from the client, deserialising etc.

The one part that remains that _potentially_ costs something is the re-query of the state store for every change. But I expect that to be a very small portion of the CPU usage end to end here since in this case the query should be a single index lookup that returns no results and so a quick exit.

Did you see any data while debugging that indicated the proportion of CPU that could be saved here vs the internal query parts? No worries if not just interested.

For comparison, where this query work was an issue in the past it was for service health with a large number of instances because then there was O(N) work to do joining all the node, check and instance information to build the result, so when N got large, and many watchers were all doing this in a semi-hot loop (on every service update) it got really expensive.

So yeah this seems like a really simple change that should be a very significant CPU saving in this common case 👏

dnephin:
Yes, exactly! I have no measured the cost of the extra query, but I don't expect that to be significant. I think it's worth mentioning that this extra cost already exists in all of our blocking queries. I believe the `WatchSet` can fire any time a node is modified because of a delete or insert. The frequency of these fires is hard to predict, but it should be fair to say it happens often enough.

Any time the `WatchSet` fires due to a node update, we'd have the same behaviour as you describe. There would be an extra state store query, but the returned index would not have changed (because it should generally be the ModifyIndex), so we would never notice those queries in our logs or metrics.

I think given that this already happens today in normal blocking-queries, there's no real concern that it will continue to happen for blocking-queries for non-existent entries.

agent/consul/kvs_endpoint.go
```diff
@@ -160,18 +160,13 @@ func (k *KVS) Get(args *structs.KeyRequest, reply *structs.IndexedDirEntries) er
...
 
 			if ent == nil {
-				// Must provide non-zero index to prevent blocking
-				// Index 1 is impossible anyways (due to Raft internals)
-				if index == 0 {
-					reply.Index = 1
-				} else {
-					reply.Index = index
-				}
+				reply.Index = index
```
rboyer, line 163: Presumably this part of the change was intentional?

dnephin: Ya, this is now handled generally in `Server.blockingQuery` by calling `Server.setQueryMeta`, and setting the value to 1 if it is still 0.  I guess this logic in `kvs_endpoint` predated the general fix in `blockingQuery`.

agent/consul/rpc.go
```diff
@@ -966,7 +971,15 @@ func (s *Server) blockingQuery(queryOpts structs.QueryOptionsCompat, queryMeta s
 
 		err := fn(ws, state)
 		s.setQueryMeta(queryMeta, queryOpts.GetToken())
-		if err != nil {
+		switch {
+		case errors.Is(err, errNotFound):
+			if notFound {
```
rboyer, line 976: Why do we need `var notFound bool`? It seems like it moots the whole point, because the first `errNotFound` skips the conditional here, and proceeds exactly like the prior scenario, never executing line 978.

dnephin: Good question!

I believe what you are calling out is that on the first iteration of this loop we still rely on the index not having changed. If a write happens between the time the last blocking query returned and when the next blocking query request happens (which I think should generally only be a few milliseconds), then this optimization does not work. We continue to have the old behaviour of returning unnecessarily. Generally that should be ok because it only limits the optimization, should not impact correctness, and there is a very small window where a write must happen to cause this problem.

There is a way to work around this limitation (could be a follow up), but it introduces some risk, so I thought it would be better to avoid.  I haven't documented that anywhere yet (just some discussions in old slack threads), but I can do that here.

The work around for this limitation would be to add a new `X-Consul-Index-Not-Found` query parameter (as a sibling to `X-Consul-Index`). This parameter would be used to set the initial value of `var notFound bool`.  That way when a new query comes in, we'll already know if we're in a "not found" state, and only the creation of an item (or a timeout) will cause `blockingQuery` to return.  The risk here is that if  the user incorrectly sets this header, then they could miss a delete that happened, but I haven't thought enough about this case to know if it's actually a problem for correctness. Maybe we should do this as a follow up.

> Why do we need `var notFound bool`?

To answer the other part of this question, this is essential to the optimization.  I think this is explained briefly in 9abb49da51ad87312a5f34364d8e6a3c0e35c328, but I'm happy to add more detail since this is the main reason for the change.  As documented in #12343, there are two things to consider here: the "last modified index" set on `responseMeta.Index`, and the `WatchSet` blocking.

We can't fix the `WatchSet` blocking without turning every read operation into a write that modifies the memdb index (I'm pretty sure this is not an approach we want to investigation).  Maybe there are other ways of making `WatchSet` block correctly for not found items? I have not found any.

So that leaves the index comparison (`responseMeta.Index > query.MinIndex`). Generally we know this comparison will return true in  these "not found" cases, because `responseMeta.Index` will be the max index of the table, and it changes on every write. 

What we need is a comparison we can use instead of `responseMeta.Index > query.MinIndex`.

If we think about this comparison more abstractly, I think we'll see that the question this is asking is "has the result of the query changed from the last result sent to the client". There are plenty of ways that question could be answered, using the modify index is a very cost effective one, so that's generally what we use.  

However as we've noticed, using  that modify index doesn't work for cases where an item does not exist, because we don't have a stable modify index to compare against.  In this case we use  "not found" to indicate if the result has changed. 

Since we don't know the state from the client perspective (`X-Consul-Index-Not-Found` would give us that), we have to perform the query once. If the result of the query is the same "last modified index", then we know at that index the query returned no results. Until the query returns some result, the state has not changed, so we can continue to block until that time.

In the line below this, when we see "state has not changed, the result is still not found", we update the `minQueryIndex` to the latest `responseMeta.Index`, so that the comparison later keeps us in the for loop.

dnephin (commented):
I'll rebase this PR, it confliced with some renames from the docs PR.

I'll also update the godoc to document this special `errNotFound` that the query function can return.

agent/consul/rpc.go
dnephin: I opened #12363 for that potential future optimization.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/588613.

PR #12117, opened 2022-01-19
ui: Tweak some code related meta information

johncowen:
This PR _mostly_ tweaks some code related meta information.

### DIFF
--- a/ui/packages/consul-ui/app/utils/get-environment.js
+++ b/ui/packages/consul-ui/app/utils/get-environment.js
@@ -23,7 +23,7 @@ export default function(config = {}, win = window, doc = document) {
             .startsWith('CONSUL_')
         );
     };
-    win.Scenario = function(str = '') {
+    win['Scenario'] = function(str = '') {
       if (str.length > 0) {
         cookies(str).forEach(item => (doc.cookie = `${item};Path=/`));
         win.location.hash = '';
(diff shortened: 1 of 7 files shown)


ui/packages/consul-ui/app/utils/get-environment.js
```diff
@@ -23,7 +23,7 @@ export default function(config = {}, win = window, doc = document) {
             .startsWith('CONSUL_')
         );
     };
-    win.Scenario = function(str = '') {
+    win['Scenario'] = function(str = '') {
```
randallmorey, line 26: Why this change?

randallmorey (approved):

ui/packages/consul-ui/app/utils/get-environment.js
johncowen: `tsc` doesn't complain about it if I do this, and it's the best solution I could find given I want to write as little typescript as possible. It's not great I know but also importantly: it's wrapped in a runInDebug, so this is all compiled out of our production build, so I'm ok with it.

It's funny as I just realized that this is actually cheating a little as I wanted to see what was possible without changing any existing code, so I'm happy either way if this goes in or not.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/561712.

PR #12119, opened 2022-01-19
Fix testing.T scope in tests

kisunji:
Found this message in CircleCI

Screen Shot 2022-01-19 at 11 20 28 AM

This occurs when we initialize `require/assert` structs and use the same structs in subtests:
```go

require := require.New(t)

t.Run("this is a subtest", func(t *testing.T) {
    require.True(false) <- require variable is scoped with outer test's `*testing.T`
}
```

This PR fixes those subtests which use incorrectly scoped `require` or `assert` vars and removes any `require := require.New(t)` or `assert := assert.New(t)` calls (only in the affected tests).

Since we use subtests and retries extensively in our codebase, I prefer explicitly calling the exported functions with `testingT` explicitly as the first argument to prevent this kind of programmer error.

### DIFF
--- a/agent/agent_endpoint_test.go
+++ b/agent/agent_endpoint_test.go
@@ -5964,8 +5964,6 @@ func TestAgentConnectCALeafCert_good(t *testing.T) {
 
 	t.Parallel()
 
-	assert := assert.New(t)
-	require := require.New(t)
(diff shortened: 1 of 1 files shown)


kisunji:
Closing in favor of RB's more robust solution

PR #12131, opened 2022-01-20
update actions-generate-metadata to v1

99:


samsalisbury:
Looks stale so closing.

PR #12135, opened 2022-01-20
ui: Themeable icons

johncowen:
We are soon to be moving to use the new Flight icon set, and as discussed with the Design System Team, right now we want to continue to use our `*-500` scaling system for naming our icons.

We currently have 2 'types' of icon available to us for each icon, one that uses thicker lines than the other called `*-16` and `*-24` respectively.

We _may_ need to have our icons at more sizes than `16` and `24` in which case these names will be meaningless (or at the very least misleading).

We use a 3 digit numeral naming scheme for scaling the majority of our CSS related 'collections' so we decided to alias these pixel based numbers to the 3 digit number naming scheme. The idea is to use `%with-name-300` for the 16 size icon and `%with-name-700` for the 24 size icon. This means we should be able to add more and more sizes to the scale as required and resize, theme and color them using CSS as we do currently.

Additionally, in future we would like to use something like reduced motion user settings/queries to present different icons for the icons we currently animate. Using a slightly different naming system for our project here means we can do this kind of thing, globally with CSS, but still using meaningful names i.e. `%with-loading-300` might be our current spinner when we have default motion settings, but could be a motionless candybar looking icon when reduced motion is enabled. As we are doing this using globally using CSS at the top level of our project, the `%with-loading-300` naming doesn't have to change in potentially multiple places in the project, we can just add a single media query in the new `icons.scss` file here and job done.

At a project level, I'm imagining having say a `%with-nspace-300` icon also, then semantically within the project its clear that you need to use this placeholder when you want to add <whatever-icon-it-is-we-use-for-nspaces>. If we ever change the name of the icon we use for representing a namespace we can do it within the `%with-nspace-300` placeholder.

All in all this is probably a 'little-by-little' process and we might end up changing our approach to icons entirely further down the road, but in the meantime this approach works well for us.

Please note, this was mainly a small spike in order to check the feasibility of this approach, the initial full "move to flight icons" will be in a different PR, and that PR will use aliasing to move the entire app to use Flight all in one go (not a little-by-little refactor approach for a move to semantically named icon placeholders explained above) - I just wanted to get this straight first.

The work here includes a new artifact from `iris` (here our `base` folder) for the new vault coloring, and then proof of concept top level theming of an icon using this semantic/numerically scaled naming approach.

**Light:**

Screenshot 2022-01-20 at 11 44 53

**Dark:**

Screenshot 2022-01-20 at 11 45 35

As always please ask if there are any questions, ty!


ui/packages/consul-ui/app/styles/base/icons/base-variables.scss
```diff
@@ -3891,7 +3891,7 @@
...
+  --vault-16-svg: url('data:image/svg+xml;charset=UTF-8,<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="none" viewBox="0 0 16 16"><path fill="%23000" d="M0 0l7.971 15.516L16 0H0zm6.732 6.16h-1.27V4.89h1.27v1.27zm0-1.906h-1.27V2.985h1.27v1.269zm1.904 3.81h-1.27v-1.27h1.27v1.27zm0-1.905h-1.27V4.89h1.27v1.27zm0-1.905h-1.27V2.985h1.27v1.269zm1.894 1.905H9.26V4.89h1.27v1.27zM9.26 4.254V2.985h1.27v1.269H9.26z"/></svg>');
```
amyrlam, line 3894: question: Where are you getting the `url` from here?

amyrlam (approved):
Left one non-blocking question but the iterative approach sounds good to me!

Feel free to tag us on any icon / design system things also, even when Evan's back. I'll cc @didoo as seems RTYI

didoo:
@johncowen for context, we're releasing animated icons very soon (see https://github.com/hashicorp/flight/pull/383), and the animation values (duration) are very specific (see codepen in comment here https://github.com/hashicorp/flight/issues/282#issuecomment-1013500493). at the moment we've included the animation as SMIL (so embedded in the SVG itself) but we're discussing how to use CSS to allow `prefers-reduced-motion` to disable the animation. I suggest to follow that ticket so you'll be kept updated.

ui/packages/consul-ui/app/styles/base/icons/base-variables.scss
johncowen: Oh awesome, thanks for asking! I forgot to put this in the PR desc 🤦 

I grabbed it from here - Flight v2 https://github.com/hashicorp/flight/blob/main/flight-icons/svg/vault-16.svg 

There's a PR coming soon that does the full upgrade to flight v2, but I wanted to use this icon for my experiment here before doing the upgrade.

johncowen:
Awesome, thanks for the info @didoo 🙏 I'll take a look!

PR #12137, opened 2022-01-20
test: stop using require.New and assert.New from testify

rboyer:
Some individual commits have some of the bash used to do most of the conversion. It is expected that this thing is going to rot the moment i opened the PR, so I'll have to fix up the collisions after folks have a chance to review it.

I patched some irregular usages of testify and then just used this script:
```
#!/bin/bash

set -euo pipefail

unset CDPATH

cd "$(dirname "$0")"

for f in $(git grep '\brequire := require\.New(' | cut -d':' -f1 | sort -u); do
    echo "=== require: $f ==="
    sed -i '/require := require.New(t)/d' $f
    # require.XXX(blah) but not require.XXX(tblah) or require.XXX(rblah)
    sed -i 's/\brequire\.\([a-zA-Z0-9_]*\)(\([^tr]\)/require.\1(t,\2/g' $f
    # require.XXX(tblah) but not require.XXX(t, blah)
    sed -i 's/\brequire\.\([a-zA-Z0-9_]*\)(\(t[^,]\)/require.\1(t,\2/g' $f
    # require.XXX(rblah) but not require.XXX(r, blah)
    sed -i 's/\brequire\.\([a-zA-Z0-9_]*\)(\(r[^,]\)/require.\1(t,\2/g' $f
    gofmt -s -w $f
done

for f in $(git grep '\bassert := assert\.New(' | cut -d':' -f1 | sort -u); do
    echo "=== assert: $f ==="
    sed -i '/assert := assert.New(t)/d' $f
    # assert.XXX(blah) but not assert.XXX(tblah) or assert.XXX(rblah)
    sed -i 's/\bassert\.\([a-zA-Z0-9_]*\)(\([^tr]\)/assert.\1(t,\2/g' $f
    # assert.XXX(tblah) but not assert.XXX(t, blah)
    sed -i 's/\bassert\.\([a-zA-Z0-9_]*\)(\(t[^,]\)/assert.\1(t,\2/g' $f
    # assert.XXX(rblah) but not assert.XXX(r, blah)
    sed -i 's/\bassert\.\([a-zA-Z0-9_]*\)(\(r[^,]\)/assert.\1(t,\2/g' $f
    gofmt -s -w $f
done
```

Followup: lint to fail if the old methods are used in tests

PR #12139, opened 2022-01-20
lint: forbid require.New and assert.New

rboyer:
See #12137

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,6 +7,7 @@ linters:
     - staticcheck
     - ineffassign
     - unparam
+    - forbidigo
 
 issues:
   # Disable the default exclude list so that all excludes are explicitly
@@ -57,6 +58,14 @@ issues:
 linters-settings:
   gofmt:
     simplify: true
+  forbidigo:
+    # Forbid the following identifiers (list of regexp).
+    forbid:
+      - '\brequire\.New\b'
+      - '\bassert\.New\b'
+    # Exclude godoc examples from forbidigo checks.
+    # Default: true
+    exclude_godoc_examples: false
 
 run:
   timeout: 10m


rboyer:
An example of tripping it:
```
$ make lint
--> Running go golangci-lint
agent/xds/x_test.go:11:2: use of `require.New` forbidden by pattern `\brequire\.New\b` (forbidigo)
	require.New(t)
	^
make: *** [GNUmakefile:271: lint] Error 1
```

dnephin (approved):
Cool! I haven't used this linter before, but seems like a good fit for this

rboyer:
It's strictly string matching, which isn't great, but since we don't have obvious collisions it'll be fine for now. You can still get around it like this until we find a way to switch to a more semantic approach:
```
package foo

import (
  "testing"
  zrequire "github.com/stretchr/testify/require"
)

func TestTest(t *testing.T) {
   r := zrequire.New(t)
   ...
 }  
 ```

**Comment by hc-github-team-consul-core** [comment]

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/560717.

File: .golangci.yml [hunk 1]
```diff
@@ -57,6 +58,14 @@ issues:
 linters-settings:
   gofmt:
     simplify: true
+  forbidigo:
+    # Forbid the following identifiers (list of regexp).
+    forbid:
+      - '\brequire\.New\b'
+      - '\bassert\.New\b'
```
kisunji, line 65:
```suggestion
      - '\brequire\.New\b(# Use package-level functions instead)?'
      - '\bassert\.New\b(# Use package-level functions instead)?'
```

This will make the output like:
```
use of `ast.Inspect` forbidden because "use package-level functions instead" at blah.go:32:2
```

kisunji: Oh merge happened already 😆 not a big deal

PR #12144, opened 2022-01-20
Show URL decoding of query params is not needed

jkirschner-hashicorp:
On #11258, we stated:
> We intend to open a separate issue for applying the same encode/decode approach to URL query parameters (e.g., "/v1/example/endpoint?id=" + "12345"), which a community member can work on after the URL path parameters fix (this issue) is merged/closed!

However, it looks like that change isn't necessary.

The URL library's accessor function for query parameters (`req.URL.Query().Get("param")`) seems to already do that URL decoding for us. This is demonstrated in the added unit test.

This PR is not meant to be merged, only to get confirmation from others that no change is necessary to URL decode query parameters.

### DIFF
--- a/agent/http.go
+++ b/agent/http.go
@@ -1156,3 +1156,21 @@ func getPathSuffixUnescaped(path string, prefixToTrim string) (string, error) {
 
 	return suffixUnescaped, nil
 }
+
+func getQueryParamUnescaped(queryParamRaw string) (string) {
+	// If unescaping causes an error, the original input will be returned.
+	// This function does not return an error for convenience of use:
+	// not needing to define error handling everywhere a query parameter is
+	// read. The caller does not need to adapt its behavior based on whether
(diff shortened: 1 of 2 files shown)


agent/http.go
```diff
@@ -1156,3 +1156,21 @@ func getPathSuffixUnescaped(path string, prefixToTrim string) (string, error) {
 
 	return suffixUnescaped, nil
 }
+
+func getQueryParamUnescaped(queryParamRaw string) (string) {
```
jkirschner-hashicorp, line 1160: Ignore this function. I wrote it intending to use it lots of places, then realized I should check to see if this is handled already for us by `Query().Get()`.

kisunji:
Oh neat! This makes sense to me - we shouldn't need those followup PRs

jkirschner-hashicorp:
Thanks @kisunji. Always nice when something is already done for you under the hood :) Closing this PR.

PR #12147, opened 2022-01-21
packaging: fix issues in pre/postremove scripts

elfchief:
Fixes several issues with the pre/postremove scripts for both rpm and deb packages. Specifically:

For postremove:

- the postremove script now functions correctly (i.e. restarts consul after a package upgrade) on rpm-based systems (where `$1` is numeric rather than `purge` or `upgrade`)
- `systemctl daemon-reload` is called on package removal (rather than only on upgrade)
- calls `systemctl try-restart` instead of `systemctl restart`, which will only (re)start consul if it was already running when the upgrade happened.

For preremove:
- if the package is being completely uninstalled (rather than upgraded), stop consul before removing the package

Tested on CentOS 7.9 and Ubuntu 20.04, but should be applicable to all rpm and deb based distributions.

### DIFF
--- a/.release/linux/preremove
+++ b/.release/linux/preremove
@@ -1,0 +1,11 @@
+#!/bin/bash
+case "$1" in
+    remove | 0)
+        if [ -d "/run/systemd/system" ]; then
+            systemctl --no-reload disable consul.service > /dev/null || :
+            systemctl stop consul.service > /dev/null || :
+        fi
+        ;;
+esac
+
+exit 0
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,7 +3,7 @@ name: build
 on:
   push:
     # Sequence of patterns matched against refs/heads
-    branches: [ 
+    branches: [
       "main"
     ]
 
@@ -145,6 +145,7 @@ jobs:
           config_dir: ".release/linux/package"
           preinstall: ".release/linux/preinstall"
           postinstall: ".release/linux/postinstall"
+          preremove: ".release/linux/preremove"
           postremove: ".release/linux/postremove"
 
       - name: Set Package Names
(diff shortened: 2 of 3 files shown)


elfchief:
Not sure why the envoy integration tests are failing, didn't touch anything that could remotely affect any actual code tests, so will assume just a flakey test unless told otherwise. (I can't see a way to just re-run the test, either, as a random user)

.release/linux/preremove
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+case "$1" in
+    remove | 0)
```
eculver, line 3: do we care about the `upgrade` case here too?

eculver:
This all _looks_ correct but it's a little over my head with regards to the systemd and nfpm bits. I would have to defer to others who know more about these tools to safely approve, maybe @kpenfound? Otherwise, this looks good to me! The part about calling `try-restart` makes a lot of sense.

.release/linux/preremove
alinsavix: In this case, no, because we don't want to stop/disable anything if we're doing an upgrade -- the postremove handles restarting things, so we just let the upgrade case fall through. Normal package upgrade behavior, barring specific reasons to do things a different way, is "leave things running, swap out for the new version, then conditionally restart".

kpenfound (approved):
Yeah looks great @elfchief , thanks for the contribution! Definitely better handling of upgrade and remove than before

mdeggies:
Want to do the honors of merging, @eculver ? :)

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/572746.

PR #12149, opened 2022-01-21
Make LockRetryTime configurable in api locks

akzi:
The lock time is not configurable and by default, it is set to 5 seconds which may not be suitable to consumers.

This preserves the default and allows the user to set it with a new config field called `LockRetryTime` which is similar to the existing `LockWaitTime` option.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "makefile.extensionOutputFolder": "./.vscode"
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "makefile.extensionOutputFolder": "./.vscode"
-}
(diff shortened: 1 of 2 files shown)


FFMMM (changes requested):
hey @akzi thanks loads for opening this!

I took a first pass here and things look good.

Would you mind adding more context on this PR? Specifically, what's the problem you are trying to solve and what use cases are you encountering it in?

Also, as reviewers, we always appreciate a good PR description.

Lastly, and more importantly, what kind of testing can we add to this change?

akzi:
many clients try to lock and unlock the same locker. sometime it will sleep for 5 second to retry when 
lock failed. this take too much time.

FFMMM:
hey @akzi thanks for following up here and for your patience with our response.

Things look good overall.

What kind of testing could we add here?

eculver:
> What kind of testing could we add here?

Just to chime in with a suggestion, if you look in the [`api/lock_test.go`](https://github.com/hashicorp/consul/blob/main/api/lock_test.go#L533-L536) file, you can see similar test consideration for the `LockWaitTime` option that you can follow for this new option. It's just a suggestion; I thought it might help get you started.

Amier3:
Hey @akzi 

Are you still interested in working on this?

PR #12165, opened 2022-01-22
acl: remove some of the duplicate resolve token methods

dnephin:
This is the first step in #11690, best viewed by individual commit

* Embed `ACLResolver` in `Client` and `Server`
* remove the methods off of `Client` and `Server`, and move one to `ACLResolver`

Issue #11690, opened 2021-11-30
acl: improve the ResolveToken interface


dnephin:
The interface for resolving a token today is a bit fragmented. The following methods exist:
* `Server.ResolveTokenAndDefaultMeta`
* `Client.ResolveTokenAndDefaultMeta`  - duplicate of `Server.ResolveTokenAndDefaultMeta`
* `Server.ResolveTokenToIdentity` - mostly the same underlying logic as above, but different args and return value
* `Server.Resolvetoken` - basically the same with fewer args and return values, legacy call that has need been cleaned up yet
* `Client.ResolveTokenToIdentity` - duplicate of `Server.ResolveTokenToIdentity`
* `ACLResolver.ResolveTokenToIdentityAndAuthorizer` - called by the above
* `ACLResolver.ResolveTokenToIdentity` similar to the above
* `Server.ResolveIdentityFromToken` - used internally by `ACLResolver`, but occasionally not, see #11221

The large number of similar methods makes it difficult to understand which one to use, and increase the potential for bugs when the wrong one is called.

The "main" call is supposed to be `{Server, Client}.ResolveTokenAndDefaultMeta`. That method has a few problems:
1. the `entMeta` and `authzContext` args are things to "fill". There are other ways to fill these two things, and it's not really obvious from the function signature how this "fill" works
2. there is no way to access the `acl.Identity` from the return value, which is why we have the additional `ResolveTokenToIdentity` calls. In practice all we really need from the Identity is an accessor ID. The rest of the `acl.Identity` methods are internal to the ACL system.
3. the returned `acl.Authorizer` is a nice interface, but all of its methods return an `EnforcementDecision` instead of an error. This makes it difficult to solve problems like #8428 and #9933. We'd like to be able to return more information to authenticated, but not authorized requests, but we can't do that from a central place right now because of this interface.

### Proposal

To address these problems I propose we replace all of these methods with a single new method on `ACLResolver`. 

The `agent.delegate` interface can replace its two methods with a single `ACLResolver() ACLResolver` for getting access to the resolve. That along removes the duplication between Client and Server.

The remaining problems can be addressed by changing the interface to something like this:

```go
func (a *ACLResolver) ResolveToken(req Request) (Authorizer, error) {
...
}

type Request interface {
    TokenSecret() string
    EnterpriseIdentity() structs.EnterpriseMeta
}
```

[...]

PR #12166, opened 2022-01-22
acl: remove ResolveTokenToIdentity

dnephin:
Branched from #12165, continues the work described in #11690


agent/consul/acl_server.go
```diff
@@ -159,13 +159,12 @@ func (s *Server) ResolveRoleFromID(roleID string) (bool, *structs.ACLRole, error
 	return s.InPrimaryDatacenter() || index > 0, role, acl.ErrNotFound
 }
 
+// TODO: remove
 func (s *Server) ResolveToken(token string) (acl.Authorizer, error) {
```
dnephin, line 163: This is done in #12167

agent/local/state.go
```diff
@@ -1538,16 +1539,13 @@ func (l *State) notifyIfAliased(serviceID structs.ServiceID) {
 // critical purposes, such as logging. Therefore we interpret all errors as empty-string
 // so we can safely log it without handling non-critical errors at the usage site.
 func (l *State) aclAccessorID(secretID string) string {
-	ident, err := l.Delegate.ResolveTokenToIdentity(secretID)
```
freddygv, line 1541: Isn't this doing significantly more work now? Maybe not for this PR, but could we avoid building the authorizer here? This call is fairly expensive from what I've seen in some CPU profiles

agent/acl.go
```diff
@@ -15,18 +15,15 @@ import (
 // critical purposes, such as logging. Therefore we interpret all errors as empty-string
 // so we can safely log it without handling non-critical errors at the usage site.
 func (a *Agent) aclAccessorID(secretID string) string {
-	ident, err := a.delegate.ResolveTokenToIdentity(secretID)
```
freddygv, line 18: Same concern here

agent/consul/intention_endpoint.go
```diff
@@ -480,13 +481,11 @@ func (s *Intention) Get(args *structs.IntentionQueryRequest, reply *structs.Inde
...
 			// Filter
-			if err := s.srv.filterACL(args.Token, reply); err != nil {
-				return err
-			}
+			s.srv.filterACLWithAuthorizer(authz, reply)
 
 			// If ACLs prevented any responses, error
 			if len(reply.Intentions) == 0 {
-				accessorID := s.aclAccessorID(args.Token)
+				accessorID := authz.AccessorID()
```
freddygv, line 488: This is great, love that this avoids re-resolving the token

freddygv (commented):
This generally looks good, my only concern is that now some of the calls to get the accessor ID are more expensive. Left some comments below, let me know what you think.

agent/local/state.go
dnephin: Yes, good point! 

We only call this method when an RPC call returns either:
* `acl.ErrNotFound`
* `acl.ErrPermissionDenied` 

So it should be pretty rare, but we do have some options to make it less expensive.

I was going to suggest we could cache the AccessorID, but I noticed `ACLResolver` already caches the `Authorizer`. It seems like we cache up to 256 Authorizers (1024 on servers), so I wonder how likely it would be that we miss the cache.

We could restore the method to retrieve only the `Identity` from a token, but I'm not sure if these rare cases are enough to warrant that extra code.

We could also decide to add a new cache of `secretID` -> `AccessorID` that is local to only the `agent/local.State`.

I wonder if we could actually remove this call entirely soon. We are planning on adding the AccessorID to "permission denied" error. Presumably if the error is "acl not found", then there is no accessor ID to return (or if it's a policy that is not found, we could include the accessor ID in that error as well).

Given that we are planning on improve those errors in the next release, I wonder if we could leave this as-is, and remove this `aclAccessorID` method once we've enriched the errors on the server side. Does that sound like a good approach? I think that gives us better errors without having to do any kind of lookup for AccessorID.

agent/acl.go
dnephin: I see this method being called in two places: `Agent.filterMembers`, and `Agent.sendCoordinate`.

`Agent.sendCoordinate` is similar to the other case in `agent/local`. It's only done when we receive a permission denied error, and we will be included the accessorID in that error soon, so we should be able to remove that call entirely.

The second call `Agent.filterMembers` already has an `acl.Authorizer`. I missed this on the first pass! I'll push a commit which removes this call, and uses `authz.AccessorID()` to get the  value.

dnephin (commented):
Ya, good call! Most of the changes here reduce work, but there are 3 cases in `agent` (or `agent/local`) where we do more work.  I'll comment inline on each of them. I suspect we'll be able to remove those calls soon.

agent/local/state.go
markan: That seems pretty reasonable; can we capture this work somewhere to make sure it isn't missed?

markan (approved):

agent/local/state.go
dnephin: Good call! I opened #12233 to track this cleanup.

agent/consul/acl.go
```diff
@@ -1115,31 +1115,17 @@ func (r *ACLResolver) ResolveTokenToIdentityAndAuthorizer(token string) (structs
...
+}
 
-	if ident, _, ok := r.resolveLocallyManagedToken(token); ok {
-		return ident, nil
+func (a ACLResolveResult) AccessorID() string {
+	if a.ACLIdentity == nil {
+		return ""
 	}
-
-	defer metrics.MeasureSince([]string{"acl", "ResolveTokenToIdentity"}, time.Now())
```
dnephin, line 1140: I just noticed I am removing this metric. I'm going to push one more commit that documents that this metric is being removed.  I'll add a changelog as well.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/571786.

PR #12167, opened 2022-01-23
acl: rename ResolveTokenToIdentityAndAuthorizer to ResolveToken

dnephin:
Branched from #12166,  continues the work described in #11690

This change allows us to remove one of the last remaining duplicate resolve token methods (`Server.ResolveToken`).

With this change we are down to only 2, where the second one also handles setting the default EnterpriseMeta from the token.

PR #12168, opened 2022-01-23
acl: demonstrate ResolveTokenForRequest

dnephin:
Branched from #12167, this PR demonstrates the proposed (in #11690) replacement for `ResolveTokenAndDefaultMeta`.

This new method `ResolveTokenForRequest` accepts a single arg and handles setting the default enterprise meta from the token.

It does not set those values in the `AuthorizerContext`. Very few calls to `ResolveTokenAndDefaultMeta` actually make use of this `AuthorizerContext` arg. As part of the migration we can populate the `AuthorizerContext` from the caller, but in the future we should be able to remove the need for `AuthorizerContext` by instead accepting an interface to the methods on `Authorizer`. An example of that can be seen in #12169.

TODO:
* [ ] replace all calls to `ResolveTokenAndDefaultMeta` with `ResolveTokenForRequest`

### DIFF
--- a/agent/consul/acl.go
+++ b/agent/consul/acl.go
@@ -1055,13 +1055,16 @@ func (r *ACLResolver) resolveLocallyManagedToken(token string) (structs.ACLIdent
 	return r.resolveLocallyManagedEnterpriseToken(token)
 }
 
-func (r *ACLResolver) ResolveTokenToIdentityAndAuthorizer(token string) (structs.ACLIdentity, acl.Authorizer, error) {
+// ResolveToken to an acl.Authorizer and structs.ACLIdentity. The acl.Authorizer
+// can be used to check permissions granted to the token, and the ACLIdentity
+// describes the token and any defaults applied to it.
+func (r *ACLResolver) ResolveToken(token string) (ACLResolveResult, error) {
 	if !r.ACLsEnabled() {
-		return nil, acl.ManageAll(), nil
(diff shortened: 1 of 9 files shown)


agent/consul/acl.go
```diff
@@ -1169,7 +1170,30 @@ func (r *ACLResolver) ResolveTokenAndDefaultMeta(token string, entMeta *structs.
...
+}
+
+type ACLRequest interface {
+	TokenSecret() string
+
+	// Merge sets the fields in this request to the values from source, if the
+	// request does not already have a value.
+	// TODO: rename
+	Merge(source *structs.EnterpriseMeta)
+}
```
dnephin, line 1183: Almost every one of our existing request structs already implements this interface (as described in #11690). Some may not have a `Merge` method if they don't have EnterpriseMeta, but we can easily implement a no-op version to embed in those structs.

agent/consul/acl.go
```diff
@@ -1169,7 +1170,30 @@ func (r *ACLResolver) ResolveTokenAndDefaultMeta(token string, entMeta *structs.
...
 
-	return result, err
+	return result, nil
+}
+
+type ACLRequest interface {
+	TokenSecret() string
+
+	// Merge sets the fields in this request to the values from source, if the
+	// request does not already have a value.
```
dnephin, line 1180:
```suggestion
	// Merge sets the fields in this request to the values from source, if the
	// request does not already have a value for those fields.
```

PR #12169, opened 2022-01-23
acl: demonstrate new authz interface

dnephin:
Branched from #12167, demonstrates the authorization interface proposed in #11690.

This PR also demonstrates how we can used a typed error, built in a centralized place, to provide better error messages when permission is denied.

TODO:
* remove `Cause` field from `PermissionDeniedError`
* move `ACLResolveResult` and the types used in its method signatures to a new package under `acl/` maybe `acl/aclauthz` or something like that.

### DIFF
--- a/agent/consul/acl.go
+++ b/agent/consul/acl.go
@@ -1117,6 +1117,7 @@ func (r *ACLResolver) ResolveToken(token string) (ACLResolveResult, error) {
 	return ACLResolveResult{Authorizer: acl.NewChainedAuthorizer(chain), ACLIdentity: identity}, nil
 }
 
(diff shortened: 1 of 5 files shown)


agent/consul/acl.go
```diff
@@ -1132,6 +1133,27 @@ func (a ACLResolveResult) AccessorID() string {
 	return a.ACLIdentity.ID()
 }
 
+type ACLKVResourceID interface {
+	KVResourceID() string
```
dnephin, line 1137: I'm not sure if we want a separate interface for each resource type, or a small number of interfaces (one for resources with identifiers+entMeta, one for just identifiers, and one for just entMeta).

PR #12160, opened 2022-01-21
fix unknown cluster issue

fredwangwang:
when the envoy sidecar get disconnected from the xds stream, the reconnecting request will
contain `InitialResourceVersions` and `ResourceNamesSubscribe`. This is OK for endpoint and route type,
as these two are driven by (child types of) cluster and listener type respectively.

However, register `cluster` type as subscription instead of wildcard would cause envoy
not able to get any new cluster updates for the rest of this life. Same goes for `listener`.

This pr is to always set cluster and listener type to wildcard, to ensure the envoy sidecar will get
those updates after disconnecting from xds stream for whatever reason (network blip/consul restart/etc).

Issue #11833, opened 2021-12-14
Race condition adding/updating listener


lkysow:
#### Overview of the Issue

I can't reproduce this but creating a ticket to see if other people have run into the same issue.

One day I applied a proxy defaults config and my proxy went into this super tight loop where it was spamming errors and Consul was spamming these errors. Thousands of these messages every couple of seconds:

```
[ERROR] agent.envoy.xds: got error response from envoy proxy: service_id=frontend-sidecar-proxy \
typeUrl=type.googleapis.com/envoy.config.listener.v3.Listener xdsVersion=v3 nonce=000510fe \
error="rpc error: code = Internal desc = Error adding/updating listener(s) backend:127.0.0.1:6001: \
route: unknown cluster 'backend.default.dc1.internal.48b8d386-6855-4145-3bf2-90dd6fb29e96.consul
```

(newlines added for readability)

It looks like a race condition occurred where we were trying to XDS add a listener and expected a cluster to be there when it wasn't.

To fix it I had to restart the proxy.

### Consul info for both Client and Server

Consul 1.10.4

fredwangwang:
Hi @lkysow 

we came across the same issue with Consul 1.10.3

We use nomad consul integration,  after one particular job deployment (replace one service with another service), the ingress gateway started to spamming the error message:

`
{"@level":"error","@message":"got error response from envoy proxy","@module":"agent.envoy.xds.ingress_gateway","@timestamp":"2022-01-12T05:45:05.408164Z","error":"rpc error: code = Internal desc = route: unknown cluster '<new-service-name>.default.<dc-name>.internal.fe0699d0-86e3-6d4f-91d4-cefd95609e5e.consul'","nonce":"02a94654","service_id":"_nomad-task-08fc9334-9d1a-33db-784f-c53ee867bdd2-group-ingress-group-ingress-inbound","typeUrl":"type.googleapis.com/envoy.config.route.v3.RouteConfiguration","xdsVersion":"v3"}
`

Had to restart all ingress allocations after the issue happened.

fredwangwang:
@lkysow  I was able to reliably reproduce:

1. deploy ingress gateway (envoy sends initial cluster xds req, consul registers cluster handler as [**wildcard mode**](https://github.com/hashicorp/consul/blob/fb11931d5e1d559f3c6ad4f86683e78a4dab0f4e/agent/xds/delta.go#L453) )
2. restart consul agent
[...]

PR #12174, opened 2022-01-24
Fix xds delta reconnect wildcard

rboyer:
When a wildcard xDS type (LDS/CDS/SRDS) reconnects from a delta xDS stream, prior to envoy `1.19.0` it would populate the `ResourceNamesSubscribe` field with the full list of currently subscribed items, instead of simply omitting it to infer that it wanted everything (which is what wildcard mode means).

This upstream issue was filed in envoyproxy/envoy#16063 and fixed in envoyproxy/envoy#16153 which went out in Envoy `1.19.0` and is fixed in later versions (later refactored in envoyproxy/envoy#16855).

This PR conditionally forces LDS/CDS to be wildcard-only even when the connected Envoy requests a non-wildcard subscription, but only does so on versions prior to `1.19.0`, as we should not need to do this on later versions.

This replaces PR https://github.com/hashicorp/consul/pull/12160 by @fredwangwang that solved the failure case as described here: #11833 (comment)

### DIFF
--- a/agent/xds/delta.go
+++ b/agent/xds/delta.go
@@ -106,18 +106,18 @@ func (s *Server) processDelta(stream ADSDeltaStream, reqCh <-chan *envoy_discove
 	handlers := map[string]*xDSDeltaType{
 		ListenerType: newDeltaType(generator, stream, ListenerType, func(kind structs.ServiceKind) bool {
 			return cfgSnap.Kind == structs.ServiceKindIngressGateway
-		}),
+		}, true),
 		RouteType: newDeltaType(generator, stream, RouteType, func(kind structs.ServiceKind) bool {
 			return cfgSnap.Kind == structs.ServiceKindIngressGateway
-		}),
+		}, false),
 		ClusterType: newDeltaType(generator, stream, ClusterType, func(kind structs.ServiceKind) bool {
 			// Mesh, Ingress, and Terminating gateways are allowed to inform CDS of
 			// no clusters.
 			return cfgSnap.Kind == structs.ServiceKindMeshGateway ||
 				cfgSnap.Kind == structs.ServiceKindTerminatingGateway ||
 				cfgSnap.Kind == structs.ServiceKindIngressGateway
-		}),
-		EndpointType: newDeltaType(generator, stream, EndpointType, nil),
+		}, true),
+		EndpointType: newDeltaType(generator, stream, EndpointType, nil, false),
 	}
 
 	// Endpoints are stored within a Cluster (and Routes
@@ -379,6 +379,10 @@ type xDSDeltaType struct {
 	// specific resource names. subscribe/unsubscribe are ignored.
 	wildcard bool
 
+	// alwaysWildCard indicates that this type is always wildcard regardless of
+	// specific resource names in the initial request.
+	alwaysWildCard bool
+
 	// sentToEnvoyOnce is true after we've sent one response to envoy.
 	sentToEnvoyOnce bool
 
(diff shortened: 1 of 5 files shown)


agent/xds/delta.go
```diff
@@ -180,6 +174,12 @@ func (s *Server) processDelta(stream ADSDeltaStream, reqCh <-chan *envoy_discove
 				}
 			}
 
+			if handler, ok := handlers[req.TypeUrl]; ok {
```
rboyer, line 177: Need to move this to _after_ the version-sniff so we can use the detected version to change behavior.

agent/xds/envoy_versioning.go
```diff
@@ -13,8 +13,7 @@ var (
 	// the zero'th point release of the last element of proxysupport.EnvoyVersions.
 	minSupportedVersion = version.Must(version.NewVersion("1.17.0"))
 
-	// add min version constraints for associated feature flags when necessary, for example:
-	// minVersionAllowingEmptyGatewayClustersWithIncrementalXDS = version.Must(version.NewVersion("1.16.0"))
+	minVersionWithLDSAndCDSProperlyUsingWildcardsWithIncrementalXDS = version.Must(version.NewVersion("1.19.0"))
```
kisunji, line 16: nit: could we somehow make this var shorter and add more detail in the comment?

agent/xds/envoy_versioning.go
```diff
@@ -26,16 +25,17 @@ type unsupportedVersion struct {
 }
 
 type supportedProxyFeatures struct {
-	// add version dependent feature flags here
+	// Older versions of Envoy incorrectly exploded a wildcard subscription for
+	// LDS and CDS into specific line items on incremental xDS reconnect. They
+	// would populate both InitialResourceVersions and ResourceNamesSubscribe.
+	// In this scenario they SHOULD have populated left ResourceNamesSubscribe
+	// empty (or used an explicit "*" in later versions) to imply wildcard
+	// mode. On reconnect this causes newly created listeners and clusters to
```
kisunji, line 33:
```suggestion
	// LDS and CDS into specific line items on incremental xDS reconnect. They
	// would populate both InitialResourceVersions and ResourceNamesSubscribe
	// when they SHOULD have populated left ResourceNamesSubscribe empty
	// (or used an explicit "*" in later versions) to imply wildcard mode.
	// On reconnect this causes newly created listeners and clusters to
```

kisunji: Also,
> On reconnect this causes newly created listeners and clusters to be ignored

^ does `this` refer to "setting the flag true" or "the scenario above"?

kisunji (commented):
Some Qs about clarity in comments

agent/xds/envoy_versioning.go
eculver: Also,

> In this scenario they SHOULD have ~populated~ left ResourceNamesSubscribe

agent/xds/envoy_versioning.go
rboyer: Historically I'd gone with horribly awful names deliberately because this is code meant to be erased in 1-2 consul releases when the relevant envoy versions roll away out of the support matrix.

agent/xds/envoy_versioning.go
rboyer: I reworded this.

agent/xds/envoy_versioning.go
rboyer: I aligned the naming here with the feature struct name.

kisunji (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/564788.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit d2c0945f524ff0dc6cd9c13de363e1e1b7867d52 onto `release/1.11.x` succeeded!

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit d2c0945f524ff0dc6cd9c13de363e1e1b7867d52 onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/564788)

PR #12182, opened 2022-01-25
Leadership transfer cmd

dhiaayachi:
This is an attempt to implement a command that would trigger a raft leadership transfer. 
The idea about this trace back to a customer issue where we needed to trigger leader transfer to be able to upgrade a leader node. 

This dependent to a change in raft that fix some of the gaps in leadership transfer, those gaps would make the transfer fail. https://github.com/hashicorp/raft/pull/487

I made an attempt in naming the sub command but open to any suggestions.


dhiaayachi:
Closed in favour of https://github.com/hashicorp/consul/pull/14132

PR #12195, opened 2022-01-26
Prevent xDS tight loop on cfg errors

freddygv:
Previously the Delta xDS protocol could get into a tight loop where:
1. Consul sends xDS resources it believes are valid
2. Envoy rejects them with a NACK
3. Consul re-generates and sends the resources
4. Loop back to step 2

This tight loop leads to massive amounts of error logs in both Envoy and
Consul logs.

This commit updates the delta xDS loop to wait for new requests or
snapshots after a NACK is received from Envoy. By skipping to the top of
the for loop we avoid re-sending the resources that Envoy has already
rejected.

### DIFF
--- a/agent/xds/delta.go
+++ b/agent/xds/delta.go
@@ -182,6 +182,8 @@ func (s *Server) processDelta(stream ADSDeltaStream, reqCh <-chan *envoy_discove
 				}
 				if nack {
 					generator.Logger.Trace("got nack response for type", "typeUrl", req.TypeUrl)
+					// There is no reason to believe that generating new xDS resources from the same snapshot
+					// would lead to an ACK from Envoy. So instead we skip to waiting for a new request or snapshot.
 					goto WAIT
 				}
 			}
--- a/agent/xds/delta.go
+++ b/agent/xds/delta.go
@@ -142,6 +142,7 @@ func (s *Server) processDelta(stream ADSDeltaStream, reqCh <-chan *envoy_discove
 	}
 
 	for {
+	WAIT:
 		select {
 		case <-authTimer:
 			// It's been too long since a Discovery{Request,Response} so recheck ACLs.
@@ -175,8 +176,13 @@ func (s *Server) processDelta(stream ADSDeltaStream, reqCh <-chan *envoy_discove
 			}
 
 			if handler, ok := handlers[req.TypeUrl]; ok {
-				if handler.Recv(req, generator.ProxyFeatures) {
+				newSub, nack := handler.Recv(req, generator.ProxyFeatures)
+				if newSub {
 					generator.Logger.Trace("subscribing to type", "typeUrl", req.TypeUrl)
+				}
+				if nack {
+					generator.Logger.Trace("got nack response for type", "typeUrl", req.TypeUrl)
+					goto WAIT
 				}
 			}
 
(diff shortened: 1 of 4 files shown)


agent/xds/delta.go
```diff
@@ -434,9 +442,9 @@ func newDeltaType(
 // Recv handles new discovery requests from envoy.
 //
 // Returns true the first time a type receives a request.
-func (t *xDSDeltaType) Recv(req *envoy_discovery_v3.DeltaDiscoveryRequest, sf supportedProxyFeatures) bool {
+func (t *xDSDeltaType) Recv(req *envoy_discovery_v3.DeltaDiscoveryRequest, sf supportedProxyFeatures) (newSubscription, nack bool) {
```
kisunji, line 445: also nit: it would be cleaner for the second bool to be `ack`, so that `true, true` represents positive meanings "is new subscription, is ack-ed"

This would complicate the `if t == nil` check but I think that block can be removed (I don't see a codepath there this could be `nil` and it would be programming error anyways)

freddygv:
Thanks for the early feedback! Need to see what's going on with these tests as well.

agent/xds/delta.go
```diff
@@ -175,9 +176,16 @@ func (s *Server) processDelta(stream ADSDeltaStream, reqCh <-chan *envoy_discove
...
-				if handler.Recv(req, generator.ProxyFeatures) {
+				newSub, nack := handler.Recv(req, generator.ProxyFeatures)
+				if newSub {
 					generator.Logger.Trace("subscribing to type", "typeUrl", req.TypeUrl)
 				}
+				if nack {
+					generator.Logger.Trace("got nack response for type", "typeUrl", req.TypeUrl)
+					// There is no reason to believe that generating new xDS resources from the same snapshot
+					// would lead to an ACK from Envoy. So instead we skip to waiting for a new request or snapshot.
+					goto WAIT
```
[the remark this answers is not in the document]
freddygv: Ah didn't know that was possible, I definitely prefer that.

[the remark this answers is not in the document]
rboyer: You can also `continue LABEL` to be picky about which scope to return to.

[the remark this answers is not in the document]
kisunji: Yeah it's kind of confusing that `break` works on `switch`/`select` but not `continue` so I wouldn't be opposed to a label for clarity

agent/xds/delta.go
rboyer: An alternate idea would be making a new enum type:
```
type DeltaRecvResponse int

const (
	DeltaRecvResponseNACK DeltaRecvResponse = iota // zero value is nack which seems appropriate
	DeltaRecvResponseACK
	DeltaRecvResponseNewSubscription
)
```
Then you only have to return a single thing, instead of two bools.

freddygv: Ended up going with the enum. I added `DeltaRecvUnknownType` since we can end up with `t == nil` if Envoy decides to issue a request for a type we don't have a handler for.

agent/xds/delta.go
```diff
@@ -26,6 +26,15 @@ import (
...
 )
 
+type DeltaRecvResponse int
+
+const (
+	DeltaRecvResponseNack DeltaRecvResponse = iota
+	DeltaRecvResponseAck
+	DeltaRecvNewSubscription
+	DeltaRecvUnknownType
+)
```
kisunji, line 36: Are these meant to be exported? Would prefer to keep them imported until necessary outside `xds`

kisunji (approved):
Comment about exported consts but LGTM!

agent/xds/delta.go
```diff
@@ -175,8 +184,18 @@ func (s *Server) processDelta(stream ADSDeltaStream, reqCh <-chan *envoy_discove
 			}
 
 			if handler, ok := handlers[req.TypeUrl]; ok {
-				if handler.Recv(req, generator.ProxyFeatures) {
+				resp := handler.Recv(req, generator.ProxyFeatures)
+				switch resp {
```
rboyer, line 188:
```suggestion
				switch handler.Recv(req, generator.ProxyFeatures) {
```

rboyer (approved):
LGTM, but I too share @kisunji 's comment about exported vs un-exported symbol names.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 378a7258e3f0c793735dce892c01fd6d54727460 onto `release/1.11.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/580062)

PR #12202, opened 2022-01-27
ui: class-map helper

johncowen:
`{{class-map}}` is used to easily add a list of classes, conditionally, and
have them all formatted nicely ready to be printed in a DOM `class` attribute.

For ease, as well as using entries, you can also just provide a simple string
without the boolean and that class will always be added.

@didoo this might help you folks out also. If you ever think it will, lemme know and we can extract and publish elsewhere.

[Rendered Docs](https://consul-ui-staging-em57io68m-hashicorp.vercel.app/ui/docs/helpers/class-map)

### DIFF
--- a/ui/packages/consul-ui/app/helpers/class-map.js
+++ b/ui/packages/consul-ui/app/helpers/class-map.js
@@ -1,0 +1,17 @@
+import { helper } from '@ember/component/helper';
+
+/**
+ * Conditionally maps classInfos (classes) to a string ready for typical DOM
+ * usage (i.e. space delimited)
+ *
+ * @typedef {([string, boolean] | [string])} classInfo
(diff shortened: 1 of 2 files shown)


jgwhite (approved):
Looks good to me.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/567998.

didoo:
> @didoo this might help you folks out also. If you ever think it will, lemme know and we can extract and publish elsewhere.

Thanks @johncowen for thinking of us :)
At the moment we're using/testing this approach:
https://github.com/hashicorp/design-system-components/blob/ffe8a8d31b7c1455f356781ff7726b0ea5af5124/addon/components/hds/button/index.js#L169-L188
that works quite well for us, but I'll keep your alternative in mind.

PR #12203, opened 2022-01-27
ui: style-map helper

johncowen:
`{{style-map}}` is used to easily add a list of styles, conditionally, and
have them all formatted nicely to be printed in a DOM `style` attribute.

As well as an entry-like array you can also pass an additional `unit` property
as the 3rd item in the array. This is to make it easier to do mathematical
calculations for units without having to use `(concat)`.

If any property has a value of `null` or `undefined`, that style property will
not be included in the resulting string.

Additionally for times when `style=""` CSP headers are important, an additional transpiler can be included to turn this into a modifier in the background. It's used as a helper here to match [`class-map`](https://github.com/hashicorp/consul/pull/12202), and also we aren't yet in a position where we need to do a transpilation.

[Rendered Docs](https://consul-ui-staging-l3jhhonkv-hashicorp.vercel.app/ui/docs/helpers/style-map)

### DIFF
--- a/ui/packages/consul-ui/app/helpers/style-map.js
+++ b/ui/packages/consul-ui/app/helpers/style-map.js
@@ -1,0 +1,22 @@
+import { helper } from '@ember/component/helper';
+
+/**
+ * Conditionally maps styles to a string ready for typical DOM
+ * usage (i.e. semi-colon delimited)
+ *
+ * @typedef {([string, (string | undefined), string] | [string, (string | undefined)])} styleInfo
+ * @param {styleInfo[]} entries - An array of `styleInfo`s to map
+ * @param {boolean} transform=true - whether to perform the build-time 'helper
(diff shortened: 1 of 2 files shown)


randallmorey (commented):
Why are inline styles even necessary?  What's a real-world use case?

johncowen:
Sometimes you need to set a CSS value via javascript, granted not often but sometimes.

This helper will not be used much, but it's still useful sometimes. Code review will prevent overuse when it would be better to use classes instead.

johncowen:
👍 Yeah for sure I completely agree with what he's saying, you'll see in soon to come PRs where it's necessary (and used very minimally)

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/572529.

PR #12204, opened 2022-01-27
Add QueryRaw method to api

fxposter:
Sometimes there is no need to parse a payload of a request (especially if it's large enough) or other
parsing solutions are preferable (fastjson/easyjson/custom structs/etc). It may be convenient to allow doing such things
on top of a default consul client


FFMMM:
@fxposter thanks for opening this 💯 

> Sometimes there is no need to parse a payload of a request

Could you kindly provide more details about the problem you are trying to solve on your end?

Have you considered writing a custom client (api lib) for your use cases?  For us, the api lib is meant as a high level support for Consul and it seems like it may not fit your use case here.

---

What other solutions or approaches have you considered?

---

If you did decide to pursue this approach, what are some test plans to use here?

PR #12206, opened 2022-01-27
ui: on-outside modifier

johncowen:
`{{on-outside 'click' (fn @callback}}` works similarly to `{{on }}` but allows
you to attach handlers that is specifically not the currently modified
element.

Note: I'm aware there is an `on-click-outside` addon, but it is not in the shape I wanted (the same shape as `on`), plus we already have all the code needed for this already in our `dom` service.

[Rendered Docs](https://consul-ui-staging-m1vvynuio-hashicorp.vercel.app/ui/docs/modifiers/on-outside)

~Just realised the docs page for this one depends on `style-map` so the docs don't render without [that PR](https://github.com/hashicorp/consul/pull/12203) going in first 🤦 , will rebase and post a rendered docs link once that goes down. Or quite happy to merge this as it, the code itself doesn't depend on anything not already in main, the docs will then work again once `style-map` is down. I'm easy either way.~

### DIFF
--- a/ui/packages/consul-ui/app/modifiers/on-outside.js
+++ b/ui/packages/consul-ui/app/modifiers/on-outside.js
@@ -1,0 +1,43 @@
+import Modifier from 'ember-modifier';
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+
+export default class OnOutsideModifier extends Modifier {
+  @service('dom') dom;
+
+  constructor() {
+    super(...arguments);
+    this.doc = this.dom.document();
+  }
+  async connect(params, options) {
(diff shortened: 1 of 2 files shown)


johncowen:
I thought I may aswell switch the `style-map` usage out here for just normal `style=""` which means the docs render correctly without having to wait for https://github.com/hashicorp/consul/pull/12203 to go down and be rebased. I added the Render Docs preview link in the description also.

randallmorey (approved):
FYI:  https://github.com/hashicorp/boundary-ui/blob/main/addons/rose/addon/modifiers/on-click-outside.js

PR #12223, opened 2022-01-28
Prevent duplicate passthrough addresses

freddygv:
Transparent proxies can set up filter chains that allow direct
connections to upstream service instances. Services that can be dialed
directly are stored in the PassthroughUpstreams map of the proxycfg
snapshot.

Previously these addresses were not being cleaned up based on new
service health data. The list of addresses associated with an upstream
service would only ever grow.

As services scale up and down, eventually they will have instances
assigned to an IP that was previously assigned to a different service.
When IP addresses are duplicated across filter chain match rules the
listener config will be rejected by Envoy.

This PR updates the proxycfg snapshot management to prevent ending
up in this situation.

Additional context in each commit.

TODO: 
- [x] See if additional proxycfg test cases are needed
- [x] Test against known repro of issue on k8s


agent/proxycfg/upstreams.go
```diff
@@ -92,56 +90,25 @@ func (s *handlerUpstreams) handleUpdateUpstreams(ctx context.Context, u cache.Up
...
-						Addrs: make(map[string]struct{}),
-					}
-				}
-
-				// Make sure to use an external address when crossing partitions.
-				isRemote := !structs.EqualPartitions(svc.PartitionOrDefault(), s.proxyID.PartitionOrDefault())
-				addr, _ := node.BestAddress(isRemote)
-
-				upstreamsSnapshot.PassthroughUpstreams[NewUpstreamIDFromServiceName(svc)].Addrs[addr] = struct{}{}
+			if snap.Proxy.Mode != structs.ProxyModeTransparent || !node.Service.Proxy.TransparentProxy.DialedDirectly {
```
rboyer: The first part of this conditional (`if snap.Proxy.Mode != structs.ProxyModeTransparent`) is constant, so you could move it out of the loop entirely. I think you skip ALL of the nodes when that is true.

agent/xds/clusters.go
```diff
@@ -171,36 +171,51 @@ func makePassthroughClusters(cfgSnap *proxycfg.ConfigSnapshot) ([]proto.Message,
 		})
 	}
 
-	for _, passthrough := range cfgSnap.ConnectProxy.PassthroughUpstreams {
-		// Prefixed with passthrough to distinguish from non-passthrough clusters for the same upstream.
-		name := "passthrough~" + passthrough.SNI
+	for _, target := range cfgSnap.ConnectProxy.PassthroughUpstreams {
```
rboyer, line 174: Would this be better if it were named `targets` since it's multiple?

agent/xds/listeners.go
```diff
@@ -218,26 +218,33 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
...
-			filterChain, err := s.makeUpstreamFilterChain(filterChainOpts{
-				clusterName: "passthrough~" + passthrough.SNI,
-				filterName:  filterName,
-				protocol:    "tcp",
-			})
-			if err != nil {
-				return nil, err
-			}
-			filterChain.FilterChainMatch = makeFilterChainMatchFromAddrs(passthrough.Addrs)
+				u := structs.Upstream{
```
rboyer, line 228: I don't think you even need to make this object, since you only use it to acces the same fields on line 234 that you used to construct it.

agent/proxycfg/upstreams.go
freddygv: Yep that was changed in a later commit

reneeckstein:
Hi @freddygv 
that's a really great feature, we currently evaluate Consul Connect Transparent Proxy on an EKS cluster with about 30 microservices and now facing this issue. It took us some time to figure out this as the root cause. We were also surprised how often EKS reuses pod IP addresses, even in large /20 VPC subnets.

This leads to many connection failures after some time on a development k8s cluster with many CI/CD deployments running.
Another negative side effect is the k8s log flooding, up to 32k/minute log records on affected pods in the `envoy-sidecar` container, with these two repeating messages

```
[2022-02-07 14:13:30.300][1][warning][config] [source/common/config/delta_subscription_state.cc:159] delta config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) outbound_listener:127.0.0.1:15001: error adding listener '127.0.0.1:15001': multiple filter chains with overlapping matching rules are defined

[2022-02-07 14:13:30.300][1][warning][config] [source/common/config/grpc_subscription_impl.cc:126] gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) outbound_listener:127.0.0.1:15001: error adding listener '127.0.0.1:15001': multiple filter chains with overlapping matching rules are defined
```

so we really looking forward for this PR to get merged soon 🥇

agent/xds/testdata/clusters/transparent-proxy-dial-instances-directly.envoy-1-20-x.golden
```diff
@@ -206,7 +206,7 @@
     },
     {
       "@type": "type.googleapis.com/envoy.config.cluster.v3.Cluster",
-      "name": "passthrough~kafka.default.dc1.internal.e5b08d03-bfc3-c870-1833-baddb116e648.consul",
+      "name": "passthrough~kafka.default.dc1.internal.11111111-2222-3333-4444-555555555555.consul",
```
freddygv, line 209: These changed because now we're not storing the SNI nor SPIFFE ID in the config snapshot. The trust domain is plugged in from the roots when these clusters are generated.

agent/proxycfg/naming.go
```diff
@@ -13,6 +13,13 @@ type UpstreamID struct {
 	structs.EnterpriseMeta
 }
 
+func (u UpstreamID) Matches(other UpstreamID) bool {
```
rboyer, line 16: If you normalize the entmeta, you should be able to do just `==` on the `UpstreamID` instead of using a whole other method. If you needed this it probably means something else is subtly wrong, since UpstreamID has to be used as a map key in various places which requires it being normalized early.

agent/proxycfg/snapshot.go
```diff
@@ -52,15 +51,31 @@ type ConfigSnapshotUpstreams struct {
...
 	UpstreamConfig map[UpstreamID]*structs.Upstream
 
-	// PassthroughEndpoints is a map of: UpstreamID -> ServicePassthroughAddrs.
-	PassthroughUpstreams map[UpstreamID]ServicePassthroughAddrs
+	// PassthroughEndpoints is a map of: UpstreamID -> (map of TargetID ->
+	// (set of IP addresses)). It contains the upstream endpoints that
+	// can be dialed directly by a transparent proxy.
+	PassthroughUpstreams map[UpstreamID]map[string]map[string]struct{}
+
+	// PassthroughIndices is a map of: string -> indexedTarget.
```
rboyer, line 59: Can you explicitly mention what the string means here?

agent/proxycfg/connect_proxy.go
```diff
@@ -326,6 +327,18 @@ func (s *handlerConnectProxy) handleUpdate(ctx context.Context, u cache.UpdateEv
 				delete(snap.ConnectProxy.WatchedDiscoveryChains, uid)
 			}
 		}
+		for uid := range snap.ConnectProxy.PassthroughUpstreams {
+			if _, ok := seenUpstreams[uid]; !ok {
+				delete(snap.ConnectProxy.PassthroughUpstreams, uid)
+
+				for addr, indexed := range snap.ConnectProxy.PassthroughIndices {
```
rboyer, line 334: Should you separately do a range over the PassthroughIndices map to do a belt-and-suspenders cleanup of stray references too?

agent/proxycfg/upstreams.go
```diff
@@ -92,55 +90,60 @@ func (s *handlerUpstreams) handleUpdateUpstreams(ctx context.Context, u cache.Up
...
-		var passthroughAddrs map[string]ServicePassthroughAddrs
+		if s.kind != structs.ServiceKindConnectProxy || s.proxyCfg.Mode != structs.ProxyModeTransparent {
+			return nil
+		}
 
-		for _, node := range resp.Nodes {
-			if snap.Proxy.Mode == structs.ProxyModeTransparent && node.Service.Proxy.TransparentProxy.DialedDirectly {
-				if passthroughAddrs == nil {
-					passthroughAddrs = make(map[string]ServicePassthroughAddrs)
+		if _, ok := upstreamsSnapshot.PassthroughUpstreams[uid]; ok {
```
rboyer, line 97: I think this block needs a comment, because I am reading this and it seems possibly wrong, or I haven't grokked what is supposed to happen here:
> for all node addresses associated with this  target, if there's an index entry for the address that matches our target and upstream (so it's our own) then _delete it (?)_

rboyer: Is the intent to first "clear out" this upstreamid+targetid from `PassthroughIndices` and `PassthroughUpstreams` and then repopulate them?

agent/proxycfg/upstreams.go
```diff
@@ -92,55 +90,60 @@ func (s *handlerUpstreams) handleUpdateUpstreams(ctx context.Context, u cache.Up
...
+		passthroughs := make(map[string]struct{})
+
+		for _, node := range resp.Nodes {
+			if !node.Service.Proxy.TransparentProxy.DialedDirectly {
+				continue
+			}
 
-				sni := connect.ServiceSNI(svc.Name, "", svc.NamespaceOrDefault(), svc.PartitionOrDefault(), snap.Datacenter, snap.Roots.TrustDomain)
+			// Make sure to use an external address when crossing partitions.
+			// Datacenter is not considered because transparent proxies cannot dial other datacenters.
```
rboyer, line 114: Should you somehow explicitly skip this if there's a datacenter mismatch instead of just letting it do something weird?

agent/proxycfg/upstreams.go
```diff
@@ -92,55 +90,60 @@ func (s *handlerUpstreams) handleUpdateUpstreams(ctx context.Context, u cache.Up
...
-				svcUID := NewUpstreamIDFromServiceName(svc)
-				if _, ok := upstreamsSnapshot.PassthroughUpstreams[svcUID]; !ok {
-					upstreamsSnapshot.PassthroughUpstreams[svcUID] = ServicePassthroughAddrs{
-						SNI:      sni,
-						SpiffeID: spiffeID,
+			// The modify index is chosen based on the logic from BestAddress
+			csnIdx := node.Service.ModifyIndex
+			if node.Service.Address == "" {
+				csnIdx = node.Node.ModifyIndex
+			}
```
rboyer, line 124: You could do two things to make this easier to track:

1. move this into a sibling method `BestAddressModifyIndex`
2. change the return value of `BestAddress` to be `(uint64 /*raft index*/, string /*addr*/, int /*port*/)`

agent/proxycfg/naming.go
freddygv: Ah nope didn't need that. Removed.

agent/proxycfg/connect_proxy.go
freddygv: Yeah, breaking this off

agent/proxycfg/upstreams.go
freddygv: Yeah it was to clean up data from service instances that aren't in the latest response. I added a comment to that effect.

agent/proxycfg/upstreams.go
freddygv: Hmm looking at this again it is possible for a tproxy to have an upstream target in another DC. A transparent proxy's upstreams are inferred from intentions, but one of those upstreams from intentions could configure a redirect to a different DC. 

So I think the right thing here is to not skip, but rather to incorporate a DC check. Made that change.

agent/proxycfg/naming_test.go
```diff
@@ -8,6 +8,43 @@ import (
 	"github.com/hashicorp/consul/agent/structs"
 )
 
+// TODO(freddy): Needs enterprise test
```
rboyer, line 11: reminder

rboyer (approved):

agent/proxycfg/naming_test.go
freddygv: Yep, I have this queued up in an ent commit, will cherry-pick it onto the oss->ent merge fix

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/580330.

Issue #11393, opened 2021-10-21
[refactor] Use HTTP error-handling consistently across handlers


kisunji:
In our handlers/endpoints we have many places where we explicitly write a status code header and a body message ([example](https://github.com/hashicorp/consul/blob/8f754aba141e6d914e1a36731f1e9c67002d2d99/agent/agent_endpoint.go#L613-L615)).

We also have custom error types like [`BadRequestError`](https://github.com/hashicorp/consul/blob/8f754aba141e6d914e1a36731f1e9c67002d2d99/agent/http.go#L55) which are propagated up from the handler and centrally handled in this [function](https://github.com/hashicorp/consul/blob/8f754aba141e6d914e1a36731f1e9c67002d2d99/agent/http.go#L435-L477).

Having two ways of doing things is a source of confusion and tech debt, especially in Go where error-handling becomes a pain point very quickly when you stray from `if err != nil { return err }`

Since our handler signatures are `func(resp, req) (interface{}, error)`, my vote goes towards idiomatic error handling (`return nil, err`) with defined HTTP error types. Developers working on handler logic should not be concerned about writing to response headers or body (unless there's some lower-level network handling required like streaming) and defer the handling to the HTTPServer middleware.

Further down, we could leverage Go's error-wrapping features for type/behaviour assertions so that we can define custom error types as required, fulfill an interface declared at our `consul/agent/http` layer like `HasStatusCode`, and confidently append error details in a call stack without worrying about breaking a type assertion or having to add a custom type to the top-level error multiplexing.

dnephin:
Thanks for writing up this issue! Definitely agree it would be valuable to be consistent here.

I'd like to mention that this is also a problem for the HTTP response body (not just the status code). All of those handlers could write the response directly to `resp`, or they can return something as the first return value which will be JSON encoded.

It would definitely be a lot safer if the handlers had only one option, but there may be a few special cases  (like the monitor and metrics endpoints) where we need to stream a response.

To handle special cases, I think we might want a layered approach. Most endpoints implement an interface that doesn't give them direct access to the response, and we use an adapter to convert those to an interface  that does write things directly to a response. So each individual endpoint only has one option, but we still have the flexibility to do different things for endpoints that need to stream.

Mathew-Estafanous:
[...]

PR #12225, opened 2022-01-29
Change error-handling across handlers. 

Mathew-Estafanous:
issue #11393 

This is a continuation of a previous PR (#11599). Transition all the remaining endpoints to returning an error as opposed to writing directly to the response body.

### DIFF
--- a/agent/txn_endpoint.go
+++ b/agent/txn_endpoint.go
@@ -101,7 +101,7 @@ func (s *HTTPHandlers) convertOps(resp http.ResponseWriter, req *http.Request) (
 			// if the Content-Length header was not set by the client.
 			return nil, 0, EntityTooLargeError{
 				Reason: fmt.Sprintf("Request body too large, max size: %d bytes. See %s.",
-				maxTxnLen, "https://www.consul.io/docs/agent/options.html#txn_max_req_len"),
+					maxTxnLen, "https://www.consul.io/docs/agent/options.html#txn_max_req_len"),
 			}
 		} else {
 			// Note the body is in API format, and not the RPC format. If we can't
--- a/agent/txn_endpoint.go
+++ b/agent/txn_endpoint.go
@@ -88,8 +88,8 @@ func (s *HTTPHandlers) convertOps(resp http.ResponseWriter, req *http.Request) (
 	// Check Content-Length first before decoding to return early
 	if req.ContentLength > maxTxnLen {
 		return nil, 0, EntityTooLargeError{
-			Reason: fmt.Sprintf("Request body too large, max size: %d bytes. See %s.",
-				maxTxnLen, "https://www.consul.io/docs/agent/options.html#txn_max_req_len"),
+			Reason: fmt.Sprintf("Request body(%d bytes) too large, max size: %d bytes. See %s.",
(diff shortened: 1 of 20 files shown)


agent/txn_endpoint.go
```diff
@@ -87,13 +87,10 @@ func (s *HTTPHandlers) convertOps(resp http.ResponseWriter, req *http.Request) (
...
 	if req.ContentLength > maxTxnLen {
-		resp.WriteHeader(http.StatusRequestEntityTooLarge)
-		fmt.Fprintf(resp,
-			"Request body(%d bytes) too large, max size: %d bytes. See %s.",
-			req.ContentLength, maxTxnLen,
-			"https://www.consul.io/docs/agent/options.html#txn_max_req_len",
-		)
-		return nil, 0, false
+		return nil, 0, EntityTooLargeError{
+			Reason: fmt.Sprintf("Request body too large, max size: %d bytes. See %s.",
```
kisunji, line 91: This is missing `(%d bytes)` from the original

kisunji (changes requested):
Thanks again for your contribution! One minor comment about error string and could you remove `agent/uiserver/bindata_assetfs.go` from the PR (we try to generate it by CI)?

Mathew-Estafanous:
> could you remove `agent/uiserver/bindata_assetfs.go` from the PR (we try to generate it by CI)?

I’m not exactly sure how I could undo that change since I already pushed it. I believe the change happened when I used ``gofmt -s -w``.

Is there a way to go back to the commit history and unstage that specific file from the commit history? I look up how to do that, but I’m worried about messing up the git history.
Would using this work? ``git filter-branch --force --index-filter 'git rm --cached --ignore-unmatch ./agent/uiserver/bindata_assetfs.go' --prune-empty --tag-name-filter cat -- --all``

kisunji:
> > could you remove `agent/uiserver/bindata_assetfs.go` from the PR (we try to generate it by CI)?
> 
> I’m not exactly sure how I could undo that change since I already pushed it. I believe the change happened when I used `gofmt -s -w`.
> 
> Is there a way to go back to the commit history and remove that specific file from the commit history? I looked up how to do that, but I’m worried about messing up the git history. Would using this work? `git filter-branch --force --index-filter 'git rm --cached --ignore-unmatch ./agent/uiserver/bindata_assetfs.go' --prune-empty --tag-name-filter cat -- --all`

Since this is your own branch, you can simply `git checkout main -- agent/uiserver/bindata_assetfs.go` and push a new commit. I'll be squashing your commits into one so there isn't really a need to keep your history clean

kisunji:
Out of scope for this PR but might be interesting for you to think about:

We have a bunch of custom error types now for common HTTP Status Codes - most of them have a `Reason string` field and a custom type assertion function in `http.go`. One downside of our approach is that any new status code we want to return requires the creation of a new error type and assertion func (as you've done with `EntityTooLargeError` in this PR)

Do you think we could refactor these into a more general HTTP error type?

kisunji (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/570828.

PR #12242, opened 2022-02-01
ui: Change approach to loading debug.css

johncowen:
We need a way to load certain CSS based on the environment you are viewing, i.e. we have debug CSS that we use for our Eng Documentation and various other DX utilities that shouldn't be compiled into our production or test builds.

Previously we would compile two entirely different CSS files (`app` and `debug`) and the load one or the other depending on which environment you were in.

This approach just empties out the debug.css file in certain environments (prod/test) which means we can just import that file from app. When in staging/development this imports the contents of debug.css (quite a bit of CSS) whereas when building for production/test this debug.css is emptied out during the build process.

There is a slight little hack in order to have this work, we import `_debug.scss` which imports the `debug.scss` file. I couldn't for the life of me figure out how to have broccoli empty out a file during the build process, so instead we essentially copy over debug.scss during dev and create an empty file during prod to `_debug.scss`.

When using `make build` to build an artifact for production CSS remains at ~58kb (during dev its a lot bigger than this)

### DIFF
--- a/ui/packages/consul-ui/GNUmakefile
+++ b/ui/packages/consul-ui/GNUmakefile
@@ -5,8 +5,8 @@ all: build
 
 deps: ../../node_modules clean
 
+# incase we ever need to clean anything again
 clean:
-	rm -rf ./tmp
 
 build-staging: deps
 	yarn run build:staging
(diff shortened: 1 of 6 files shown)


randallmorey (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/574218.

PR #12250, opened 2022-02-02
acl: un-embed ACLIdentity

dnephin:
This is safer than embedding two interface because there are a number of places where we check the concrete type. If we check the concrete type on the top-level interface it will fail. So instead expose the ACLIdentity from a method.

The top-level type (`ACLResultResult`) is a struct, but in other packages we may consume it as an interface, which introduces the problem.

### DIFF
--- a/agent/acl_test.go
+++ b/agent/acl_test.go
@@ -39,6 +39,12 @@ type TestACLAgent struct {
 func NewTestACLAgent(t *testing.T, name string, hcl string, resolveAuthz authzResolver, resolveIdent identResolver) *TestACLAgent {
 	t.Helper()
 
+	if resolveIdent == nil {
+		resolveIdent = func(s string) (structs.ACLIdentity, error) {
+			return nil, nil
+		}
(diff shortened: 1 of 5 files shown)


agent/acl_test.go
```diff
@@ -39,6 +39,12 @@ type TestACLAgent struct {
 func NewTestACLAgent(t *testing.T, name string, hcl string, resolveAuthz authzResolver, resolveIdent identResolver) *TestACLAgent {
 	t.Helper()
 
+	if resolveIdent == nil {
+		resolveIdent = func(s string) (structs.ACLIdentity, error) {
+			return nil, nil
+		}
+	}
```
dnephin, line 46: backport from enterprise, apparently in OSS we don't call this fake with a `nil`.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/573145.

PR #12259, opened 2022-02-03
ui: ShadowTemplate component

johncowen:
See [Rendered Docs](https://consul-ui-staging-bivp363bv-hashicorp.vercel.app/ui/docs/components/shadow-template)

### DIFF
--- a/ui/packages/consul-ui/app/components/shadow-template/index.hbs
+++ b/ui/packages/consul-ui/app/components/shadow-template/index.hbs
@@ -1,0 +1,11 @@
+{{#if @shadowRoot}}
+  {{#in-element @shadowRoot}}
+    {{#if @stylesheets}}
+      {{adopt-styles
+        @shadowRoot
+        @stylesheets
+      }}
+    {{/if}}
+  {{yield}}
+  {{/in-element}}
+{{/if}}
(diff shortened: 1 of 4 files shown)


ui/packages/consul-ui/app/components/shadow-template/index.hbs
```diff
@@ -0,0 +1,11 @@
+{{#if @shadowRoot}}
```
jgwhite, line 1: So without `@shadowRoot` this component is a no-op?

jgwhite (approved):
The code looks fine. I don’t pretend to understand the full implications of this component but I’m keen to read any relevant RFCs.

ui/packages/consul-ui/app/components/shadow-template/index.hbs
jgwhite: Ah, gotcha

johncowen:
No entirely relevant RFC on this one as yet unfortunately (but it is extremely related and an extension of _CSL-165: Re-organizing Consul UIs component file structure_). I would definitely write up my up to date findings if you'd be interested in helping there tho? I think overall tho its mainly "Use The Platform", i.e. ShadowDOM is a thing and I would like to make use of it in a way that fits well with Ember. Lastly note we aren't using this anywhere as yet, but its ready for when we do want to use it, or indeed iterate on it before use.

PR #12263, opened 2022-02-03
ui: Disclosure Component

johncowen:
See [Rendered Docs](https://consul-ui-staging-l8deir0r3-hashicorp.vercel.app/ui/docs/components/disclosure)

I realised after building the docs that you might want to place the trigger underneath the panel (DOM-wise), so a contextual pattern would actually be better here. I'm happy to leave that to a future (soonish) PR to amend. The work that this is working towards only uses this in one place, so a quick refactor once the feature is complete won't be a huge issue. I'd rather keep going with the feature than slightly rethink things.

### DIFF
--- a/ui/packages/consul-ui/app/components/disclosure/README.mdx
+++ b/ui/packages/consul-ui/app/components/disclosure/README.mdx
@@ -7,7 +7,7 @@ you want to contruct your own trigger, disclosure has all the properties to
 enable you to do so.
 
 You should make use of the `disclosure.panel` property in order to 'tag' the
-discloure panel you are using.
+disclosure panel you are using.
 
 Clicking outside will not close the disclosure by default, if you require this
 functionality please combine with our `{{on-outside 'click'}}` modifier (see example).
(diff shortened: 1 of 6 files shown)


randallmorey (changes requested):
Disclosure has a binary state, right?  Does this warrant the complexity of a state machine?

johncowen:
I was guessing that might come up 😂 

Once we add 'closing' and 'opening' for animations it will have 4 states, don't need that just yet though but I'd rather start as I mean to go on.

Plus, overall I like that we are telling the UI what state we want it to be in rather than asking by looking at computed property. Using StateMachines gives us that characteristic. From the original RFC which touches on this idea:

> Using state charts inverts what is generally done in ember to work with different component states. Generally, multiple pieces of data (the extended state) are queried or observed to derive the state based mostly on conditionals (a bottom up approach), which can lead to some quite complicated conditional statements. State charts invert this by defining the states themselves (a top down approach)

randallmorey:
Can animation be achieved with pure CSS?  Just wondering if this can be simplified.  I don't object to state machines when necessary but this component seems like it is has a very simple binary state.  Not going to hold up merge though.

randallmorey (approved):

johncowen:
> Can animation be achieved with pure CSS?

Yeah it probably will, but it will have an onanimationend event which sends it to an open/closed state, and you can't go from opening to closed, you need to go from opening to closing to closed etc.

> Just wondering if this can be simplified. I don't object to state machines when necessary but this component seems like it is has a very simple binary state. Not going to hold up merge though.

Yeah for sure, honestly glad you noticed and asked 👍

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/578150.

PR #12265, opened 2022-02-03
sdk: add TestLogLevel for setting log level in tests

### LINKED ISSUE #12143: testing: reduce verbosity of test failures
Many of the tests in the `agent` and `agent/consul` are integration style tests that setup a full `Agent`, or full `Server`. These tests setup logging with the `Debug` level.

When one of these tests fail, the entire debug log is printed to stdout. I have found this makes tests difficult to debug for 2 reasons:
1. generally the most important thing we want to see is which assertion failed and where. This output is hidden in the hundreds of lines of debug logging output
2. if there are log lines relevant to the failure, they can be difficult to find. Generally I would expect those log lines to be at WARN or ERROR level. The hundreds of INFO and DEBUG lines add significant time to debug any test failure.

I propose we make the following change:
1. default the log level to `WARN`
2. add support for a `TEST_LOG_LEVEL` env var, which can be used to change the log level to DEBUG, INFO, or TRACE.

I believe that change would need to be made in at least these two places, but maybe there are more I'm not familiar with.

https://github.com/hashicorp/consul/blob/6cecd35e6e44de6cb94782613546c98fc2811f2a/agent/testagent.go#L162

https://github.com/hashicorp/consul/blob/6cecd35e6e44de6cb94782613546c98fc2811f2a/agent/consul/client_test.go#L510

The expected workflow for debugging a test would be:
1. read the test output, which should contain only the assertion failure text, and a few WARN and ERROR lines
2. if those lines do not provide enough information, re-run the test with `TEST_LOG_LEVEL=debug` to get more information. If this happens, it's a good signal that we should change the test to make it easier to debug without the full debug log. Generally by adding more information to assertions, or adding a few `t.Log` for information information that may be needed to debug a failure.

I believe this change will have the following positive benefits:
1. test failures should take less time to debug, because we'll spend less time reading hundreds of unrelated log lines
2. it will encourage writing better tests that are more explicit about why they failed
3. it should give us important signals about how useful our logs are. If we have WARN or ERROR logs that are at the wrong level, or if we're missing important warnings in our logs, we'll have a better opportunity to notice these problems.

This change should be quite small, most likely under 20 lines of test code to change.

dnephin:
Closes #12143

And default log level to WARN.

### DIFF
--- a/agent/ae/ae_test.go
+++ b/agent/ae/ae_test.go
@@ -8,10 +8,11 @@ import (
 	"testing"
 	"time"
 
-	"github.com/hashicorp/consul/lib"
-	"github.com/hashicorp/consul/sdk/testutil"
 	"github.com/hashicorp/go-hclog"
 	"github.com/stretchr/testify/assert"
+
+	"github.com/hashicorp/consul/lib"
+	"github.com/hashicorp/consul/sdk/testutil"
 )
 
 func TestAE_scaleFactor(t *testing.T) {
@@ -401,7 +402,6 @@ func (m *mock) SyncChanges() error {
 
 func testSyncer(t *testing.T) *StateSyncer {
 	logger := hclog.New(&hclog.LoggerOptions{
-		Level:  0,
 		Output: testutil.NewLogBuffer(t),
 	})
 
--- a/agent/acl_test.go
+++ b/agent/acl_test.go
@@ -67,7 +67,7 @@ func NewTestACLAgent(t *testing.T, name string, hcl string, resolveAuthz authzRe
 
 	bd.Logger = hclog.NewInterceptLogger(&hclog.LoggerOptions{
 		Name:       name,
-		Level:      hclog.Debug,
+		Level:      testutil.TestLogLevel,
 		Output:     logBuffer,
 		TimeFormat: "04:05.000",
 	})
--- a/agent/consul/client_test.go
+++ b/agent/consul/client_test.go
@@ -507,7 +507,7 @@ func newDefaultDeps(t *testing.T, c *Config) Deps {
 
 	logger := hclog.NewInterceptLogger(&hclog.LoggerOptions{
 		Name:   c.NodeName,
-		Level:  hclog.Debug,
+		Level:  testutil.TestLogLevel,
 		Output: testutil.NewLogBuffer(t),
 	})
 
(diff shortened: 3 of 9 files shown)


agent/ae/ae_test.go
```diff
@@ -401,7 +402,6 @@ func (m *mock) SyncChanges() error {
 
 func testSyncer(t *testing.T) *StateSyncer {
 	logger := hclog.New(&hclog.LoggerOptions{
-		Level:  0,
```
dnephin, line 404: `0` is the zero value for this field, so no need to explicit set it to 0 here. I removed a couple instances of the zero in this PR while I was looking over all the places we set level in tests.

eculver (approved):
This all looks good. If you wanted to get _extra_ fancy, you could even add a test for the new `TestLogLevel` test code :trollface:

eculver:
By putting the `pr/no-changelog` label on this, you're basically saying it doesn't need a changelog entry. I'm curious what the rationale is here? Is it because it's only in test code? I'm wondering if it would be good to publicize this new method via the CHANGELOG?

FFMMM:
>  I'm wondering if it would be good to publicize this new method via the CHANGELOG?

Yeah I'm curious about this too; Theoretically, this could break folks that write their own tests on top of ours, right? 

> And default log level to WARN.

Could we keep the CI on to DEBUG? What do y'all think? Alternatively, could we parametrize thru the CI log level? (assuming it's not already)

dnephin:
Looks like I broke some tests that needed the more verbose log level. I'll look at fixing those.

I forgot to mention that I tested this manually with different levels and with the env var unset.

> changelog

The changelog is for users of Consul. This change does not impact users. Even someone using the SDK will not be impacted, because the log level is not set by the SDK.

> Could we keep the CI on to DEBUG?

We could if we find that to be valuable. Personally I would prefer that we stop relying on the logs and we force ourselves to write tests that can be debugged based on the assertions made by the test, or explicit `t.Log` messages. I think that is a better long term approach. Having to read through long debug logs to maybe find some relevant logs is a bad developer experience.

dnephin:
Ok, CI is happy with the last commit I pushed to fix the test failures. The commit is pretty minor. Changes a couple tests to use `hclog.Debug` and adds a `LogLevel` field to `TestAgent` to allow tests to set the level explicitly when they need to.

I'll wait a bit to allow for review/comments on the last commit.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/577262.

PR #12267, opened 2022-02-03
ca: change the PrivateKey type/bits validation

### LINKED ISSUE #12246: ca: relax (or remove) validation of PrivateKeyType/PrivateKeyBits
This was added in #10331 as part of #9572.

The current validation works well if Consul is responsible for generating the root CA, but there's no requirement to have Consul generate the root CA. For both the built-in and Vault providers the user can setup the root CA ahead of time, and Consul will use the already configured CA. This manual setup will be required for #11910 (#11598).

For the manually configured root CA scenario, this validation causes problems. An example of the problem can be seen in this test case: https://github.com/hashicorp/consul/blob/2881f232dc2d9f01871768c896c1ffc2dc121b60/agent/consul/leader_connect_ca_test.go#L649-L652

The user is required to specify these values in the Consul config exclusively to get past the validation. The config values will never be used by Consul for the root CA.

Another example is this [comment](https://github.com/hashicorp/consul/issues/9572#issuecomment-995260937) on the issue. The validation is preventing the user from changing the key bits used to generate intermediate certificates.

dnephin:
Fixes #12246

This commit makes two changes to the validation.

Previously we would call this validation in GenerateRoot, which happens
both on initialization (when a follower becomes leader), and when a
configuration is updated. We only want to do this validation during
config update so the logic was moved to the `UpdateConfiguration`
method.

Previously we would compare the config values against the actual cert.
This caused problems when the cert was created manually in Vault (not
created by Consul).  Now we compare the new config against the previous
config. Using an already created CA cert should never error now.

Adding the key bit and types to the config should only error when
the previous values were not the defaults.

This does mean that changing these values away from the defaults without
changing the root_pki_path will continue to silently do nothing, but I think
that is a better trade-off than return an error incorrectly at times.

In the future we should look to better structure the CA config to indicate which
values are supposed to rotate the root cert, and which ones are not.

I believe the original change only went into 1.11.x, so I am only backporting to one
release branch.

### DIFF
--- a/agent/connect/ca/provider_vault.go
+++ b/agent/connect/ca/provider_vault.go
@@ -160,6 +160,34 @@ func (v *VaultProvider) Configure(cfg ProviderConfig) error {
 	return nil
 }
 
+func (v *VaultProvider) ValidateConfigUpdate(prevRaw, nextRaw map[string]interface{}) error {
+	prev, err := ParseVaultCAConfig(prevRaw)
+	if err != nil {
+		return err
+	}
+	next, err := ParseVaultCAConfig(nextRaw)
+	if err != nil {
+		return err
+	}
+
+	if prev.RootPKIPath != next.RootPKIPath {
+		return nil
+	}
+
+	if prev.PrivateKeyType != "" && prev.PrivateKeyType != connect.DefaultPrivateKeyType {
+		if prev.PrivateKeyType != next.PrivateKeyType {
+			return fmt.Errorf("cannot update the PrivateKeyType field without changing RootPKIPath")
+		}
+	}
+
+	if prev.PrivateKeyBits != 0 && prev.PrivateKeyBits != connect.DefaultPrivateKeyBits {
+		if prev.PrivateKeyBits != next.PrivateKeyBits {
+			return fmt.Errorf("cannot update the PrivateKeyBits field without changing RootPKIPath")
+		}
+	}
+	return nil
+}
+
 // renewToken uses a vaultapi.LifetimeWatcher to repeatedly renew our token's lease.
 // If the token can no longer be renewed and auth method is set,
 // it will re-authenticate to Vault using the auth method and restart the renewer with the new token.
@@ -271,31 +299,6 @@ func (v *VaultProvider) GenerateRoot() (RootResult, error) {
 	default:
 		if err != nil {
 			return RootResult{}, err
-		}
-
-		if rootPEM != "" {
-			rootCert, err := connect.ParseCert(rootPEM)
-			if err != nil {
-				return RootResult{}, err
-			}
-
-			// Vault PKI doesn't allow in-place cert/key regeneration. That
-			// means if you need to change either the key type or key bits then
-			// you also need to provide new mount points.
-			// https://www.vaultproject.io/api-docs/secret/pki#generate-root
-			//
-			// A separate bug in vault likely also requires that you use the
-			// ForceWithoutCrossSigning option when changing key types.
-			foundKeyType, foundKeyBits, err := connect.KeyInfoFromCert(rootCert)
-			if err != nil {
-				return RootResult{}, err
-			}
-			if v.config.PrivateKeyType != foundKeyType {
-				return RootResult{}, fmt.Errorf("cannot update the PrivateKeyType field without choosing a new PKI mount for the root CA")
-			}
-			if v.config.PrivateKeyBits != foundKeyBits {
-				return RootResult{}, fmt.Errorf("cannot update the PrivateKeyBits field without choosing a new PKI mount for the root CA")
-			}
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


agent/connect/ca/provider_vault.go
```diff
@@ -160,6 +160,34 @@ func (v *VaultProvider) Configure(cfg ProviderConfig) error {
...
 
+func (v *VaultProvider) ValidateConfigUpdate(prevRaw, nextRaw map[string]interface{}) error {
+	prev, err := ParseVaultCAConfig(prevRaw)
+	if err != nil {
+		return err
+	}
+	next, err := ParseVaultCAConfig(nextRaw)
+	if err != nil {
+		return err
+	}
```
kisunji, line 171: nit: Can we wrap the errors from these two calls to better distinguish where it may have come from?

agent/consul/connect_ca_endpoint_test.go
```diff
@@ -573,28 +570,56 @@ func TestConnectCAConfig_Vault_TriggerRotation_Fails(t *testing.T) {
...
 
-	// Capture the current root.
-	{
-		rootList, _, err := getTestRoots(s1, "dc1")
-		require.NoError(t, err)
-		require.Len(t, rootList.Roots, 1)
-	}
-
-	cases := []struct {
+	// note: unlikely many table tests, the ordering of these cases does matter
```
kisunji, line 575:
```suggestion
	// note: unlike many table tests, the ordering of these cases does matter
```

agent/consul/connect_ca_endpoint_test.go
```diff
@@ -573,28 +570,56 @@ func TestConnectCAConfig_Vault_TriggerRotation_Fails(t *testing.T) {
...
+	testSteps := []struct {
 		name      string
-		configFn  func() (*structs.CAConfiguration, error)
+		configFn  func() *structs.CAConfiguration
 		expectErr string
 	}{
 		{
-			name: "cannot edit key bits",
-			configFn: func() (*structs.CAConfiguration, error) {
+			name: "allow modifying key type and bits from default",
```
kisunji, line 584: In this test and `"allow update that does not change key type or bits"` below, it's hard for me to parse what fields were changed (i.e. what the default values were at the start)

agent/consul/leader_connect_ca.go
```diff
@@ -858,6 +867,19 @@ func (c *CAManager) UpdateConfiguration(args *structs.CARequest) (reterr error)
 	return nil
 }
 
+// ValidateConfigUpdater is an optional interface that may be implemented
+// by a ca.Provider. If the provider implements this interface, the
+// ValidateConfigurationUpdate will be called when a user attempts to change the
+// CA configuration, and the provider type has not changed from the previous
+// configuration.
+type ValidateConfigUpdater interface {
```
kisunji, line 875: aside: I think `ConfigUpdateValidator` is less ambiguous but `ConfigUpdateValidate()` OTOH is less clear

agent/consul/connect_ca_endpoint_test.go
dnephin: I can add the default values to the initial config. Do you think that is enough to address this?

agent/consul/leader_connect_ca.go
dnephin: Ya, this interface could probably use a better name. It is hard to name things in Consul because we have way too many things in these packages.

agent/consul/connect_ca_endpoint_test.go
dnephin: Oh, I can also use the vars to specify the defaults, to make it more clear.

dnephin: I tried something else. I used an inline function to hide the common parts, and accept just the two values that are changing. Do you think this helps?

kisunji: Yeah, it's more clear this way. I'm still unclear what the difference between this test and `allow update that does not change key type or bits` is - the inputs are the same if I'm not mistaken?

kisunji: Ah I think I get it - first test case is `(default bits, default type -> rsa, 4096)`, whereas the last test has no diff but should still be a successful RPC call

kisunji (approved):

agent/consul/connect_ca_endpoint_test.go
dnephin: Yup, exactly!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/576482.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 74ee46e6f5bbce27393c21dca0e304e2d5cf0efb onto `release/1.11.x` succeeded!

PR #12269, opened 2022-02-04
Remove S3 deployment reference for our ent binaries

allthingsclowd:
My understanding is that we no longer support the S3 deployment mechanism and it has been turned off. 
Ideally we should remove the whole block referring to this mechanism? It will confuse customers that have previously used this approach.


allthingsclowd:
There's a good example of the language that the Vault team used to relay this information [here](https://www.vaultproject.io/docs/enterprise/license/autoloading)

PR #12277, opened 2022-02-04
catalog: initialize the refs map to prevent a nil panic

### LINKED ISSUE #12258: Consul cluster enteres infinite rebooting (panic) loop
#### Overview of the Issue

We encountered case in which a nomad job triggered consul to restart in an infinite loop. After which it was hard to revert to "workable" state. Reverting the job change was not helpful, stopping / restarting the whole cluster either. In general we were forced to reconstruct the cluster from scratch (drop consul data) to recover.  

The job that triggered the issue, was changing nginx based load balancer to ingress gateway. Below there is minimized version of it:
```hcl
  group "load-balancer" {
    constraint {
      attribute = "${attr.unique.hostname}"
      value = "example.com"
    }

    service {
      name = "service-name"
      tags = ["core", "http"]
      port = "http"

      connect {
        gateway {
          proxy {}
          ingress {
            listener {
              port     = 5080
              protocol = "tcp"
              service {
                name = "service-name-api"
              }
            }
          }
        }
      }
    }

    network {
      mode ="bridge"
      port "http" {
        static = 5080
        host_network = "localhost"
      }
    }
  }
```
`localhost` host_network is just a `lo` interface.

Essentially the issue reduces to the panic below:

```
Feb 03 13:18:02 consul[1888392]: panic: assignment to entry in nil map
Feb 03 13:18:02 consul[1888392]: goroutine 85 [running]:
Feb 03 13:18:02 consul[1888392]: github.com/hashicorp/consul/agent/consul/state.updateMeshTopology({0x30bcf00, 0xc0015edd10}, 0x29c, {0xc002d80d60, 0xf}, 0xc001740480, {0x2adfec0, 0xc001c78c40})
Feb 03 13:18:02 consul[1888392]:         /home/runner/work/consul/consul/agent/consul/state/catalog.go:3917 +0x7cd
Feb 03 13:18:02 consul[1888392]: github.com/hashicorp/consul/agent/consul/state.ensureServiceTxn({0x30bcf00, 0xc0015edd10}, 0x29c, {0xc002d80d60, 0xf}, 0x0, 0xc001740480)
Feb 03 13:18:02 consul[1888392]:         /home/runner/work/consul/consul/agent/consul/state/catalog.go:762 +0x43c
Feb 03 13:18:02 consul[1888392]: github.com/hashicorp/consul/agent/consul/state.(*Store).ensureRegistrationTxn(0xc001870e80, {0x30bcf00, 0xc0015edd10}, 0xc00297a3c0, 0x0, 0xc00297a3c0, 0x0)
Feb 03 13:18:02 consul[1888392]:         /home/runner/work/consul/consul/agent/consul/state/catalog.go:214 +0x690
Feb 03 13:18:02 consul[1888392]: github.com/hashicorp/consul/agent/consul/state.(*Store).EnsureRegistration(0xc00186be00, 0x6e4, 0x6e4)
Feb 03 13:18:02 consul[1888392]:         /home/runner/work/consul/consul/agent/consul/state/catalog.go:124 +0xaf
Feb 03 13:18:02 consul[1888392]: github.com/hashicorp/consul/agent/consul/fsm.(*FSM).applyRegister(0xc000dda120, {0xc00027d101, 0x6e4, 0x6e4}, 0xc0005dbb86)
Feb 03 13:18:02 consul[1888392]:         /home/runner/work/consul/consul/agent/consul/fsm/commands_oss.go:144 +0x1a5
Feb 03 13:18:02 consul[1888392]: github.com/hashicorp/consul/agent/consul/fsm.NewFromDeps.func1({0xc00027d101, 0xc00182fa40, 0x203000}, 0x203000)
Feb 03 13:18:02 consul[1888392]:         /home/runner/work/consul/consul/agent/consul/fsm/fsm.go:99 +0x36
Feb 03 13:18:02 consul[1888392]: github.com/hashicorp/consul/agent/consul/fsm.(*FSM).Apply(0xc000dda120, 0xc0019abc20)
Feb 03 13:18:02 consul[1888392]:         /home/runner/work/consul/consul/agent/consul/fsm/fsm.go:133 +0x16c
Feb 03 13:18:02 consul[1888392]: github.com/hashicorp/go-raftchunking.(*ChunkingFSM).Apply(0xc00164f8c0, 0x0)
Feb 03 13:18:02 consul[1888392]:         /home/runner/go/pkg/mod/github.com/hashicorp/go-raftchunking@v0.6.1/fsm.go:66 +0x254
Feb 03 13:18:02 consul[1888392]: github.com/hashicorp/raft.(*Raft).runFSM.func1(0xc0017f8110)
Feb 03 13:18:02 consul[1888392]:         /home/runner/go/pkg/mod/github.com/hashicorp/raft@v1.3.3/fsm.go:90 +0x235
Feb 03 13:18:02 consul[1888392]: github.com/hashicorp/raft.(*Raft).runFSM.func2({0xc00168f200, 0x40, 0xc00186bf50})
Feb 03 13:18:02 consul[1888392]:         /home/runner/go/pkg/mod/github.com/hashicorp/raft@v1.3.3/fsm.go:113 +0x48f
Feb 03 13:18:02 consul[1888392]: github.com/hashicorp/raft.(*Raft).runFSM(0xc00183f180)
Feb 03 13:18:02 consul[1888392]:         /home/runner/go/pkg/mod/github.com/hashicorp/raft@v1.3.3/fsm.go:216 +0x35a
Feb 03 13:18:02 consul[1888392]: github.com/hashicorp/raft.(*raftState).goFunc.func1()
Feb 03 13:18:02 consul[1888392]:         /home/runner/go/pkg/mod/github.com/hashicorp/raft@v1.3.3/state.go:146 +0x62
Feb 03 13:18:02 consul[1888392]: created by github.com/hashicorp/raft.(*raftState).goFunc
Feb 03 13:18:02 consul[1888392]:         /home/runner/go/pkg/mod/github.com/hashicorp/raft@v1.3.3/state.go:144 +0x92
Feb 03 13:18:02 systemd[1]: consul.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
Feb 03 13:18:02 systemd[1]: consul.service: Failed with result 'exit-code'.
Feb 03 13:18:02 systemd[1]: consul.service: Scheduled restart job, restart counter is at 18.
Feb 03 13:18:02 systemd[1]: Stopped "HashiCorp Consul - A service mesh solution".
Feb 03 13:18:02 systemd[1]: Started "HashiCorp Consul - A service mesh solution".
Feb 03 13:18:03 consul[1888481]: ==> Starting Consul agent...
Feb 03 13:18:03 consul[1888481]:            Version: '1.11.2'
```

The failure was reproducible by re-applying the job again.

Running on Ubuntu, software installed from deb packages.
```
build:
        prerelease = 
        revision = 37c7d06b
        version = 1.11.2
runtime:
        arch = amd64
        cpu_count = 16
        goroutines = 838
        max_procs = 16
        os = linux
        version = go1.17.5
```

dnephin:
Fixes #12258

The test was able to reproduce the panic, and the second commit fixes the test. There are other ways we could fix this, but this one seemed like the most obvious. Open to other suggestions.

TODO:
* [x] changelog

### DIFF
--- a/agent/consul/state/catalog.go
+++ b/agent/consul/state/catalog.go
@@ -4016,6 +4016,7 @@ func insertGatewayServiceTopologyMapping(tx WriteTxn, idx uint64, gs *structs.Ga
 	mapping := upstreamDownstream{
 		Upstream:   gs.Service,
 		Downstream: gs.Gateway,
+		Refs:       make(map[string]struct{}),
 		RaftIndex:  gs.RaftIndex,
 	}
 	if err := tx.Insert(tableMeshTopology, &mapping); err != nil {
--- a/.changelog/12277.txt
+++ b/.changelog/12277.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+config-entry: fix a panic when creating an ingress gateway config-entry and a proxy service instance, where both provided the same upstream and downstream mapping.
+```
--- a/agent/consul/state/catalog_test.go
+++ b/agent/consul/state/catalog_test.go
@@ -5977,6 +5977,40 @@ func setupIngressState(t *testing.T, s *Store) memdb.WatchSet {
 	return ws
 }
 
+func TestStore_EnsureService_DoesNotPanicOnIngressGateway(t *testing.T) {
+	store := NewStateStore(nil)
+
+	err := store.EnsureConfigEntry(1, &structs.IngressGatewayConfigEntry{
+		Kind: structs.IngressGateway,
+		Name: "the-ingress",
+		Listeners: []structs.IngressListener{
+			{
+				Port:     12345,
+				Protocol: "tcp",
+				Services: []structs.IngressService{{Name: "the-service"}},
+			},
+		},
+	})
+	require.NoError(t, err)
+
+	err = store.EnsureRegistration(2, &structs.RegisterRequest{
+		Node: "the-node",
+		Service: &structs.NodeService{
+			Kind:    structs.ServiceKindConnectProxy,
+			Service: "the-proxy",
+			Proxy: structs.ConnectProxyConfig{
+				DestinationServiceName: "the-ingress",
+				Upstreams: []structs.Upstream{
+					{
+						DestinationName: "the-service",
+					},
+				},
+			},
+		},
+	})
+	require.NoError(t, err)
+}
+
 func TestStateStore_DumpGatewayServices(t *testing.T) {
 	s := testStateStore(t)
 


agent/consul/state/catalog.go
```diff
@@ -4016,6 +4016,7 @@ func insertGatewayServiceTopologyMapping(tx WriteTxn, idx uint64, gs *structs.Ga
 	mapping := upstreamDownstream{
 		Upstream:   gs.Service,
 		Downstream: gs.Gateway,
+		Refs:       make(map[string]struct{}),
```
dnephin, line 4019: We only create this `upstreamDownstream` type in two places. The other place already properly initializes the `Refs`, so this seemed like the most direct fix.

boxofrad (approved):

freddygv (changes requested):
Thanks for submitting this!

A limitation I see here is that by initializing the Refs map as empty, the lifecycle of the upstreamDownstream mapping then becomes tied to the (rogue?) proxy service registration. If the proxy registration that sets the ingress as a downstream is deleted, then the topology mapping is deleted even though the ingress gateway config entry still exists.

Another bit of weirdness is that the ingress gateway config code expects to be the one managing the lifecycle of these mappings for gateways. When the config entry is updated we truncate and re-add these upstream/downstream mappings where the gateway is a downstream. So adding/removing Refs when proxies are registered, where the panic happened, is incompatible with that.

We should rethink some of these assumptions here.

dnephin:
> A limitation I see here is that by initializing the Refs map as empty, the lifecycle of the upstreamDownstream mapping then becomes tied to the (rogue?) proxy service registration

I don't think I totally follow. We should not be writing code that considers a `nil` map different from an empty map. So I think this change should be safe.

Are you saying there are other bugs here that need to be addressed? Do we need to do this as part of fixing this panic, or would those be enhancements we could make separately?

luckymike:
Is there an ETA on this bugfix being released? We reported this via Enterprise Support (ticket 65689) as we encountered the same issue attempting roughly the same configuration.

freddygv:
@luckymike We expect to have a series of patch releases before the end of the month, but the timing is still under discussion

dnephin:
Got it, I think that makes sense. I'll have another read of the code to make sure I understand the implications of this. 

Do you think I need any additional tests for this change?

freddygv:
One option is to expand the test in this PR to de-register the connect-proxy registration and then query the topology of the gateway. Adding this to your test shows the issue I'm referring to:

```go
    ca := &structs.CAConfiguration{
		Provider: "consul",
	}
	err = store.CASetConfig(0, ca)
	require.NoError(t, err)

	err = store.DeleteService(3, "the-node", "the-proxy", nil)

	_, topology, err := store.ServiceTopology(nil, "dc1", "the-ingress", structs.ServiceKindIngressGateway, acl.Allow, nil)
	require.NoError(t, err)

    # Should not be 0 because the config entry still defines "the-service" as an upstream of the gateway
	require.Len(t, topology.Upstreams, 1)
```

Note that the CA config was set because the topology endpoint traverses discovery chains and compiling chains requires CA config to know the trust domain.

freddygv (approved):

dnephin:
For posterity, Freddy and I chatted about this offline. It looks like fixing the reference count is more involved, so we're going to defer that fix so that we can get this patch released ASAP.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/578850.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 01784470f3b3d7d3795097b2263b1ef9773ec611 onto `release/1.11.x` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 01784470f3b3d7d3795097b2263b1ef9773ec611 onto `release/1.10.x` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 01784470f3b3d7d3795097b2263b1ef9773ec611 onto `release/1.9.x` succeeded!

msiege2:
@luckymike This fix was released in our patch releases on Feb 11, 2022.

luckymike:
🙏🏻

PR #12283, opened 2022-02-07
ci: fix arm64 build

dnephin:
'make dev' is not enough, we need to move it into PATH as well.


rboyer (approved):
LGTM

dnephin (commented):
Hmm, apparently `~/go/bin` is not in the path either for this arm image on the machine executor. I change this to use `~/bin` which does appear to be in the default `PATH`.

PR #12294, opened 2022-02-09
systemd unit: validate config on reload

pznamensky:
Hey,
After creating a new service file under /etc/consul.d/ we perform `systemctl reload consul` either by hands or by a configuration management system.
Unfortunately, if we did a mistake or typo, we can't know it immediately because `systemctl reload consul` just sends a signal to Consul and this command always return success.
On the one hand, we can run `consul validate` each time we want to add a new config file from the other hand, we'd like not to do it.
Instead of it would be great to get feedback from `systemctl` right after `systemctl reload ...`.
Fortunately,  systemd provides a straightforward way to do it: we're able to add multiple reload commands and in case of failure of one of them, users/configuration management system will get an error.
I'm sure that this improvement will make it easier to interact with Consul.
At least for us 🙂

### DIFF
--- a/.release/linux/package/usr/lib/systemd/system/consul.service
+++ b/.release/linux/package/usr/lib/systemd/system/consul.service
@@ -10,6 +10,7 @@ EnvironmentFile=-/etc/consul.d/consul.env
 User=consul
 Group=consul
 ExecStart=/usr/bin/consul agent -config-dir=/etc/consul.d/
+ExecReload=/usr/bin/consul validate /etc/consul.d/
 ExecReload=/bin/kill --signal HUP $MAINPID
 KillMode=process
 KillSignal=SIGTERM


Amier3:
Hey @pznamensky 

Sorry for the delayed response. Thanks for raising this PR.  We feel your pain on this, running `consul validate` before each `systemctl reload consul` isn't an ideal workflow. 

Changing systemd files for everyone could be tricky, considering that some users may have built out workflows that aren't expecting config validations on reload. So this'll require a little bit of testing on our end to make sure this change wouldn't break anything. 

I assume you've been using this on your setup since making this PR, how's it been working for you?

pznamensky:
Hey @Amier3 
Thanks for the reply.
> So this'll require a little bit of testing on our end to make sure this change wouldn't break anything.

Yes, sure!
This is not a huge change, but stability is quite important. 
>I assume you've been using this on your setup since making this PR, how's it been working for you?

In our environment it works fine.

pznamensky:
@Amier3 thanks for the feedback.
Didn't think about partial configs.
As you noticed in systemd world it's quite easy to overwrite a unit configurations.
So it would't be a big problem to add an overwrite file.
Thanks.

PR #12297, opened 2022-02-09
URL-encode/decode resource names for HTTP API part 5

dttung2905:
the last fix for #11258


eculver (approved):
LGTM. I realize it's a little late to the game with this being the last PR in the set, but it would be nice if the tests expressed this change in some way. Bonus points if you feel like adding this to the tests!

kisunji (approved):

Issue #11811, opened 2021-12-11
ca: Vault provider creates a new intermediate on every leader rotation


dnephin:
Currently we call [Provider.GenerateIntermediate](https://github.com/hashicorp/consul/blob/v1.10.4/agent/consul/leader_connect_ca.go#L429) in the primary DC every time a new leader is elected (either on startup or leader election). This is not a problem for the Consul or AWS PCA providers because those don't use intermediate certs in the primary.

The Vault provider however, [creates a new intermediate certificate every time](https://github.com/hashicorp/consul/blob/v1.10.4/agent/connect/ca/provider_vault.go#L379-L408).  In a scenario where there are more frequent leader elections, this could lead to a very large list of `IntermediateCerts` being stored and sent to clients unnecessarily.

In the secondary we check `ActiveIntermediate` which prevents the creation of these extra intermediates.

We should do something similar in the primary DC. Instead of generating a new one each time, we should see if one already exists and use it.

I have not written a test to reproduce this behaviour.

dnephin:
Seems worth mentioning that this issue will cause the "secondary CA roots watch" routine to run in all secondary DCs on every leader rotation in the primary DC.  The logic seems correct right now in a secondary and it shouldn't also generate a new intermediate in the secondary as a result. But it's possible this was wrong in some versions and would have also triggered a new intermediate in secondary DCs.

lstoll-sfdc:
I don't think this is completely accurate - while a new intermediate is generated on leader election, the list does not grow. Immediately [before the intermediate is generated](https://github.com/hashicorp/consul/blob/7cb2af17c6757b4084465518773c0c5c8762ef6f/agent/consul/leader_connect_ca.go#L487-L496) the `newCARoot` function is called, which [constructs a fresh copy of the stored state](https://github.com/hashicorp/consul/blob/7cb2af17c6757b4084465518773c0c5c8762ef6f/agent/consul/leader_connect_ca.go#L267-L279). The new intermediate is then [appended on to](https://github.com/hashicorp/consul/blob/7cb2af17c6757b4084465518773c0c5c8762ef6f/agent/consul/leader_connect_ca.go#L534-L539) this empty, freshly initialised root and persisted.

This results in only the active intermediate being passed around. While it could probably be generated less, in practice I don't think it has much impact.

lstoll:
[...]

PR #12298, opened 2022-02-09
Avoid raft change when no config is provided on persistNewRootAndConfig

jorgemarey:
I saw in the consul logs that every 10 minutes:

```
Feb  8 20:07:08 ctrl-5c33edfc-2b {"@level":"info","@message":"updated root certificates from primary datacenter","@module":"agent.server.connect","@timestamp":"2022-02-08T20:07:08.217605Z"}
Feb  8 20:17:14 ctrl-5c33edfc-2b {"@level":"info","@message":"updated root certificates from primary datacenter","@module":"agent.server.connect","@timestamp":"2022-02-08T20:17:14.111028Z"}
Feb  8 20:27:34 ctrl-5c33edfc-2b {"@level":"info","@message":"updated root certificates from primary datacenter","@module":"agent.server.connect","@timestamp":"2022-02-08T20:27:34.153630Z"}
Feb  8 20:37:37 ctrl-5c33edfc-2b {"@level":"info","@message":"updated root certificates from primary datacenter","@module":"agent.server.connect","@timestamp":"2022-02-08T20:37:37.123134Z"}
Feb  8 20:46:14 ctrl-5c33edfc-2b {"@level":"info","@message":"updated root certificates from primary datacenter","@module":"agent.server.connect","@timestamp":"2022-02-08T20:46:14.057424Z"}
Feb  8 20:56:18 ctrl-5c33edfc-2b {"@level":"info","@message":"updated root certificates from primary datacenter","@module":"agent.server.connect","@timestamp":"2022-02-08T20:56:18.133358Z"}
```
I checked that actually no change was made to the roots or configuration so I started to look at the code and I think I found the problem.

```
// secondaryCARootWatch maintains a blocking query to the primary datacenter's
// ConnectCA.Roots endpoint to monitor when it needs to request a new signed
// intermediate certificate.
func (c *CAManager) secondaryCARootWatch(ctx context.Context) error {
....
		// Attempt to update the roots using the returned data.
		if err := c.secondaryUpdateRoots(roots); err != nil {
			return err
		}
....

	return nil
}

// secondaryUpdateRoots updates the cached roots from the primary and regenerates the intermediate
// certificate if necessary.
func (c *CAManager) secondaryUpdateRoots(roots structs.IndexedCARoots) error {
	// Update the state first to claim the 'lock'.
	if _, err := c.setState(caStateReconfig, true); err != nil {
		return err
	}
	defer c.setState(caStateInitialized, false)

	// Update the cached primary roots now that the lock is held.
	c.secondarySetPrimaryRoots(roots)

	provider, _ := c.getCAProvider()
	if provider == nil {
		// this happens when leadership is being revoked and this go routine will be stopped
		return nil
	}

	// Run the secondary CA init routine to see if we need to request a new
	// intermediate.
	if c.secondaryHasProviderRoots() {
		if err := c.secondaryInitializeIntermediateCA(provider, nil); err != nil {
			return fmt.Errorf("Failed to initialize the secondary CA: %v", err)
		}
		return nil
	}
...
}


func (c *CAManager) secondaryInitializeIntermediateCA(provider ca.Provider, config *structs.CAConfiguration) error {
....
	// Determine whether a root update is needed, and persist the roots/config accordingly.
	var newRoot *structs.CARoot
	if activeRoot == nil || needsNewIntermediate {
		newRoot = newActiveRoot
	}
	if err := c.persistNewRootAndConfig(provider, newRoot, config); err != nil {
		return err
	}

	c.setCAProvider(provider, newActiveRoot)
	return nil
}
```

persistNewRootAndConfig is called throught the replication process with nil config and roots when no change needs to be performed.

With this PR,  a change to the raft store is avoided when no roots or config are provided to persistNewRootAndConfig


dnephin (commented):
Thank you for the PR! I think you are correct, we can avoid this unnecessary raft apply, nice find!

This comparison is already pretty challenging to read. I think we'd benefit from extracting this to a function. That allows us to break up the logic a little bit like this:

(writing this quickly, so the naming and logic may not be right)
```go
func shouldPersistRootAndConfig(root *structs.CARoot, prevConfig, nextConfig *structs.CAConfiguration) bool {
    if root != nil {
        return true
    }

    if nextConfig == nil {
        return false
    }

    return nextConfig.Provider != prevConfig.Provider || !reflect.DeepEqual(nextConfig.Config, prevConfig.Config)
}
```

It would also be great to have a test case that shows this change fixes the problem. I guess that updating the config in the primary in such a way that changes the `CARoot` struct, but does not create a new root cert would accomplish that. This https://github.com/hashicorp/consul/issues/11811#issuecomment-998218691 mentions one case I noticed where this can happen.

jorgemarey:
Hi @dnephin I made the suggested changes and added a test case for this. I also added a chagelog file.

agent/consul/leader_connect_ca.go
```diff
@@ -762,6 +762,17 @@ func (c *CAManager) persistNewRootAndConfig(provider ca.Provider, newActiveRoot
...
 }
 
+func shouldPersistNewRootAndConfig(newActiveRoot *structs.CARoot, oldConfig, newConfig *structs.CAConfiguration) bool {
+	if newActiveRoot != nil {
+		return true
+	}
+
+	if newConfig == nil {
+		return false
+	}
```
dnephin, line 772: Just calling this out for anyone else reviewing or reading over this PR.

This condition is the new one that fixes the problem. Previously we were checking if `config != nil` to prevent reading a field from a nil. Here we instead say that if the config is nil then we know there is nothing to update, otherwise use the existing logic to see if the config has changed.

agent/consul/leader_connect_ca_test.go
```diff
@@ -689,6 +689,86 @@ func TestCAManager_Initialize_Vault_WithIntermediateAsPrimaryCA(t *testing.T) {
...
+		t.Skip("too slow for testing.Short")
+	}
+	ca.SkipIfVaultNotPresent(t)
+
+	vault := ca.NewTestVaultServer(t)
+	vclient := vault.Client()
+	generateExternalRootCA(t, vclient)
+
+	meshRootPath := "pki-root"
+	primaryCert := setupPrimaryCA(t, vclient, meshRootPath)
```
dnephin, line 703: This setup is only for a specific scenario (external root CA). Since the bug you are fixing applies to all setups with a secondary DC, I think we can remove this to focus the test on the setup that is required to reproduce the issue.

```suggestion
```

agent/consul/leader_connect_ca_test.go
```diff
@@ -689,6 +689,86 @@ func TestCAManager_Initialize_Vault_WithIntermediateAsPrimaryCA(t *testing.T) {
...
+			Provider: "vault",
+			Config: map[string]interface{}{
+				"Address":             vault.Addr,
+				"Token":               vault.RootToken,
+				"RootPKIPath":         meshRootPath,
+				"IntermediatePKIPath": "pki-intermediate/",
+				// TODO: there are failures to init the CA system if these are not set
+				// to the values of the already initialized CA.
+				"PrivateKeyType": "ec",
+				"PrivateKeyBits": 256,
```
dnephin, line 716: I fixed the TODO, and removed these lines from the test case you copied. And if we're not using an external CA, I think we can reduce this to:

```suggestion
				"RootPKIPath":         "pki-root/",
				"IntermediatePKIPath": "pki-intermediate/",
```

agent/consul/leader_connect_ca_test.go
```diff
@@ -689,6 +689,86 @@ func TestCAManager_Initialize_Vault_WithIntermediateAsPrimaryCA(t *testing.T) {
...
+				Provider: "vault",
+				Config: map[string]interface{}{
+					"Address":             vault.Addr,
+					"Token":               vault.RootToken,
+					"RootPKIPath":         meshRootPath,
+					"IntermediatePKIPath": "pki-secondary/",
+					// TODO: there are failures to init the CA system if these are not set
+					// to the values of the already initialized CA.
+					"PrivateKeyType": "ec",
+					"PrivateKeyBits": 256,
```
dnephin, line 750:
```suggestion
					"RootPKIPath":         "pki-root/",
					"IntermediatePKIPath": "pki-secondary/",
```

agent/consul/leader_connect_ca_test.go
```diff
@@ -689,6 +689,86 @@ func TestCAManager_Initialize_Vault_WithIntermediateAsPrimaryCA(t *testing.T) {
...
+		defer sDC2.Shutdown()
+		joinWAN(t, sDC2, s1)
+		testrpc.WaitForActiveCARoot(t, sDC2.RPC, "dc2", nil)
+
+		codec := rpcClient(t, sDC2)
+		var configBefore structs.CAConfiguration
+		err := msgpackrpc.CallWithCodec(codec, "ConnectCA.ConfigurationGet", &structs.DCSpecificRequest{}, &configBefore)
+		require.NoError(t, err)
+
+		sDC2.caManager.secondaryUpdateRoots(roots)
```
dnephin, line 763: To reproduce the problem you are fixing, I think we have to update the roots in the primary (where as this is updating the CA config in the secondary).  

One way to do that would be to call `renewLeafSigningCert` (it's used in one other test in this file) to renew the leaf signing cert in the primary. That should update the `CARoot`, which should trigger the `secondaryUpdateRoots` in the secondary, which we want to be a no-op.

I believe that will correctly mimic the event that causes the logs you were seeing.

agent/consul/leader_connect_ca_test.go
```diff
@@ -689,6 +689,86 @@ func TestCAManager_Initialize_Vault_WithIntermediateAsPrimaryCA(t *testing.T) {
...
+
+		var configAfter structs.CAConfiguration
+		err = msgpackrpc.CallWithCodec(codec, "ConnectCA.ConfigurationGet", &structs.DCSpecificRequest{}, &configAfter)
+		require.NoError(t, err)
+		require.EqualValues(t, configBefore.ModifyIndex, configAfter.ModifyIndex)
```
dnephin, line 768: Nice! this looks like the right thing to check that the fix works.

dnephin (commented):
Thanks for adding the test! Sorry it has taken me a few weeks to respond. 

This is looking great, but I think the test case is close, but doesn't quite exercise the behaviour we want yet. I left a few suggestions for how to remove some things we no longer need, and to better test the behaviour that was changed.

agent/consul/leader_connect_ca_test.go
```diff
@@ -689,6 +689,86 @@ func TestCAManager_Initialize_Vault_WithIntermediateAsPrimaryCA(t *testing.T) {
 	})
 }
 
+func TestCAManager_Verify_NoChangeToSecondaryConfig(t *testing.T) {
```
dnephin, line 692: Oh, I almost missed this!  It's unfortunate, but because of how our CI is setup right now, this test is not going to be run in CI.

To have it run properly we need the name to include the string `_Vault_`.  Otherwise the main suite skips this test (because the vault binary is not installed), and the "ca provider" test suite doesn't run the test because the name doesn't match.

It's really unfortunate that it's so easy to accidentally write a test that will never run. I've created #12466 to document this problem, and propose some options for fixing it.

For now, if you add `_Vault_` to the name, the test should run correctly.

jorgemarey:
Hi @dnephin I pushed a new commit with the suggested changes. I hope I have understood everything correctly.

agent/consul/leader_connect_ca_test.go
```diff
@@ -693,6 +693,59 @@ func TestCAManager_Initialize_Vault_WithIntermediateAsPrimaryCA(t *testing.T) {
...
+	defer sDC2.Shutdown()
+	joinWAN(t, sDC2, sDC1)
+	testrpc.WaitForActiveCARoot(t, sDC2.RPC, "dc2", nil)
+
+	codec := rpcClient(t, sDC2)
+	var configBefore structs.CAConfiguration
+	err := msgpackrpc.CallWithCodec(codec, "ConnectCA.ConfigurationGet", &structs.DCSpecificRequest{}, &configBefore)
+	require.NoError(t, err)
+
+	renewLeafSigningCert(t, sDC1.caManager, sDC1.caManager.primaryRenewIntermediate)
```
dnephin, line 740: I tried this test out on `main` (without the fix) and I saw it passing. So I added this small sleep to get it to fail. I generally try to avoid any kind of sleeping in a test, but in this case I don't see any way to avoid it. We need to give the secondary a few milliseconds to notice the update before we query it again.

The test fails reliably for me with this 100ms sleep, both on main and on this branch. The test could flake, but a flake would be when the test passes instead of fails. That should be less of a concern because it reliably fails in most cases, and won't ever incorrectly fail.

```suggestion
	renewLeafSigningCert(t, sDC1.caManager, sDC1.caManager.primaryRenewIntermediate)
	
	// Give the secondary some time to notice the update
	time.Sleep(100 * time.Millisecond)	
```

agent/consul/leader_connect_ca_test.go
```diff
@@ -693,6 +693,59 @@ func TestCAManager_Initialize_Vault_WithIntermediateAsPrimaryCA(t *testing.T) {
...
+	_, sDC2 := testServerWithConfig(t, func(c *Config) {
+		c.Datacenter = "dc2"
+		c.PrimaryDatacenter = "dc1"
+		c.CAConfig = &structs.CAConfiguration{
+			Provider: "vault",
+			Config: map[string]interface{}{
+				"Address":             vault.Addr,
+				"Token":               vault.RootToken,
+				"RootPKIPath":         "pki-root/",
+				"IntermediatePKIPath": "pki-intermediate/",
```
dnephin, line 727:
```suggestion
				"IntermediatePKIPath": "pki-intermediate-2/",
```

This needs to be a different path from the primary. The test was failing on this branch (after I added the sleep), because both the primary and secondary were using the same path.

dnephin (commented):
Nice! This is exactly what I was thinking. I think with these two small changes this should be ready to merge. I'll commit these suggestions and see if CI is happy.

dnephin (approved):
Looks like CI is happy with those changes. Thanks for making this improvement!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/601036.

PR #12301, opened 2022-02-10
file watcher to be used for configuration auto-reload feature

dhiaayachi:
This PR add a file watcher that combine the use of https://github.com/fsnotify/fsnotify and a reconcile loop that id files by inode ID in linux and Creation Time in Windows to detect if file has been recreated and call a handler with the event accordingly.

For the time being this is not yet used in Consul but the the intent is to use it to automatically reload some of the configuration options when the file get recreated.

The 2 events that could trigger the handlers are:

- Create: if a file watched is created or a file under a directory which is watched is created.
- Remove: this will trigger an event only if the removed file is recreated (a file move use case)
- In addition, a reconcile loop will run every 200 Milliseconds to go through all the watched files and check if any have a changed ID (Inode or creation time) and will trigger the event in this case.

Symlinks are not supported as fsnotify/fsnotify and the go filesystem packages are not consistent on how to treat symlinks and could get different behaviours (follow or not follow) depending on OS. If a symlink is added an error will be returned.


agent/config/file_watcher.go
```diff
@@ -0,0 +1,225 @@
...
+		configFile.iNode = id
+		return w.handleFunc(&WatcherEvent{Filename: event.Name})
+	}
+	if isCreate(event) {
+		return w.handleFunc(&WatcherEvent{Filename: event.Name})
+	}
+	return nil
+}
+
+func (w *FileWatcher) isWatched(filename string) (*watchedFile, string, bool) {
```
thisisnotashwin, line 180: This method will terminate right? 🤔 at some point we will hit a `dir` that has to be watched

dhiaayachi: Oh I fixed something in here but I don't see it, I think I did it in the wrong place!! Let me bring it here.

dhiaayachi: Should be better now

agent/config/file_watcher.go
```diff
@@ -0,0 +1,230 @@
...
+	timeout := time.After(timeoutDuration)
+	select {
+	case w.toBeRemoved <- filename:
+		return nil
+	case <-timeout:
+		return fmt.Errorf("file remove timedout %s", filename)
+	}
+}
+
+func (w *FileWatcher) Add(filename string) error {
```
dhiaayachi, line 79: I thought about making the `Add` return a `chan error` and pass the channel to the main loop to return the error if any, I refrained because I don't think we need it in the implementation. I will add it if there is a need.

agent/config/file_watcher.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+func (w *FileWatcher) remove(filename string) error {
+	if err := w.watcher.Remove(filename); err != nil {
+		return err
+	}
+	delete(w.configFiles, filename)
+	return nil
+}
+func (w *FileWatcher) Remove(filename string) error {
```
dhiaayachi, line 69: Same for `Remove`

dhiaayachi: `Remove` and `Add` are Async to avoid locking the loop when adding new files.

agent/config/file_watcher_windows.go
```diff
@@ -0,0 +1,25 @@
...
+
+package config
+
+import (
+	"fmt"
+	"os"
+	"syscall"
+)
+
+func (w *FileWatcher) getFileId(filename string) (uint64, error) {
```
dhiaayachi, line 12: Windows don't have an Inode so we have a special implementation that use `CreationTime`

agent/config/file_watcher_test.go
```diff
@@ -0,0 +1,375 @@
...
+	const charset = "abcdefghijklmnopqrstuvwxyz" +
+		"ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+	var seededRand *rand.Rand = rand.New(
+		rand.NewSource(time.Now().UnixNano()))
+	b := make([]byte, length)
+	for i := range b {
+		b[i] = charset[seededRand.Intn(len(charset))]
+	}
+	return string(b)
+}
```
dhiaayachi, line 375: Added this helper to generate random strings, but if we already have something similar I could reuse.

agent/config/file_watcher.go
```diff
@@ -0,0 +1,230 @@
...
+type watchedFile struct {
+	iNode   uint64
+	watched bool
+}
+
+type WatcherEvent struct {
+	Filename string
+}
+
+func New(handleFunc func(event *WatcherEvent) error) (*FileWatcher, error) {
```
rboyer, line 37: This should accept a logger, rather than create one.

agent/config/file_watcher.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	delete(w.configFiles, filename)
+	return nil
+}
+func (w *FileWatcher) Remove(filename string) error {
+	timeout := time.After(timeoutDuration)
+	select {
+	case w.toBeRemoved <- filename:
+		return nil
+	case <-timeout:
```
rboyer, line 74: Inline `timeout`

agent/config/file_watcher.go
```diff
@@ -0,0 +1,230 @@
...
+		return fmt.Errorf("file remove timedout %s", filename)
+	}
+}
+
+func (w *FileWatcher) Add(filename string) error {
+	timeout := time.After(timeoutDuration)
```
rboyer, line 80: Inline `timeout`

agent/config/file_watcher.go
```diff
@@ -0,0 +1,230 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/fsnotify/fsnotify"
+	"github.com/hashicorp/go-hclog"
+)
+
+const timeoutDuration = 200 * time.Millisecond
```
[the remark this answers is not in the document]
rboyer: Note: you likely still need this for reconcile, just not the rest.

agent/config/file_watcher.go
```diff
@@ -0,0 +1,221 @@
...
+
+func (w *FileWatcher) Start(ctx context.Context) {
+	cancelCtx, cancel := context.WithCancel(ctx)
+	w.cancel = cancel
+	go w.watch(cancelCtx)
+}
+
+func (w *FileWatcher) add(filename string) error {
+	if isSymLink(filename) {
+		return fmt.Errorf("symbolic link are not supported %s", filename)
```
rboyer, line 66:
```suggestion
		return fmt.Errorf("symbolic links are not supported %s", filename)
```

agent/config/file_watcher.go
```diff
@@ -0,0 +1,221 @@
...
+			}
+		case _, ok := <-w.watcher.Errors:
+			if !ok {
+				w.logger.Error("watcher error channel is closed")
+				return
+			}
+		case <-ticker.C:
+			w.reconcile()
+		case <-ctx.Done():
+			close(w.done)
```
rboyer, line 120: Should this actually be in a defer on line 99 instead? (top of the function). You have several ways to exit this method, but only one of them closes the done channel.

agent/config/file_watcher.go
```diff
@@ -0,0 +1,221 @@
...
+			return
+		}
+	}
+
```
rboyer, line 124: stray line
```suggestion
```

agent/config/file_watcher.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+func (w *FileWatcher) handleEvent(event fsnotify.Event) error {
+	w.logger.Debug("event received ", "filename", event.Name, "OP", event.Op)
+	// we only want Create and Remove events to avoid triggering a relaod on file modification
```
rboyer, line 129:
```suggestion
	// we only want Create and Remove events to avoid triggering a reload on file modification
```

agent/config/file_watcher.go
```diff
@@ -0,0 +1,221 @@
...
+	return event.Op&fsnotify.Remove == fsnotify.Remove
+}
+
+func isWrite(event fsnotify.Event) bool {
+	return event.Op&fsnotify.Write == fsnotify.Write
+}
+
+func isRename(event fsnotify.Event) bool {
+	return event.Op&fsnotify.Rename == fsnotify.Rename
+}
```
rboyer, line 212: These functions live in a shared package `agent/config` (not a package just for this machinery) so they should probably have names that are a little more specific, like `isCreateEvent` etc.

agent/config/file_watcher.go
```diff
@@ -0,0 +1,221 @@
...
+		if isRemove(event) {
+			// If the file was removed, try to reconcile and see if anything changed.
+			w.logger.Debug("attempt a reconcile ", "filename", event.Name, "OP", event.Op)
+			configFile.id = 0
+			w.reconcile()
+		}
+	}
+	if isCreate(event) || isWrite(event) || isRename(event) {
+		w.logger.Debug("call the handler", "filename", event.Name, "OP", event.Op)
+		go w.handleFunc(&WatcherEvent{Filename: filename})
```
rboyer, line 150: Why spawn a new goroutine for every one of these events that trigger a reload? You may have multiple racing reloads in play if you do that.

It might be better to have a buffered handoff channel that you write to here, and a single goroutine that serially reloads parts of the config as they change.

agent/config/file_watcher.go
```diff
@@ -0,0 +1,221 @@
...
+	stat, err := os.Stat(filename)
+	if err != nil {
+		return nil, path, false
+	}
+	if !stat.IsDir() {
+		w.logger.Trace("not a dir")
+		// try to see if the watched path is the parent dir
+		NewPath := filepath.Dir(path)
+		w.logger.Trace("get dir", "dir", NewPath)
+		configFile, ok = w.configFiles[NewPath]
```
rboyer, line 171:
```suggestion
		newPath := filepath.Dir(path)
		w.logger.Trace("get dir", "dir", newPath)
		configFile, ok = w.configFiles[newPath]
```

super weird to have a local var start with a capital letter

agent/config/file_watcher.go
```diff
@@ -0,0 +1,221 @@
...
+	return event.Op&fsnotify.Rename == fsnotify.Rename
+}
+
+func (w *FileWatcher) getFileId(filename string) (uint64, error) {
+	fileInfo, err := os.Stat(filename)
+	if err != nil {
+		return 0, err
+	}
+
+	return uint64(fileInfo.ModTime().Nanosecond()), err
```
rboyer, line 220: `(time.Time).Nanosecond` only returns the nanoseconds _within the current second_, not the total elapsed nanoseconds since 1970, that is `(time.Time).UnixNano`.

But a better question is why even convert this to a number at all? This function (were it renamed) could just return `(time.Time, error)` and then do comparisons using `(time.Time).{Equal, After, Before}` as necessary instead.

agent/config/file_watcher.go
dhiaayachi: yes, that's a good point. I think I will change the API to have `NewFileWatcher` return a channel and the user can read the events from that channel.

agent/config/file_watcher.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+func (w *FileWatcher) isWatched(filename string) (*watchedFile, string, bool) {
+	path := filename
+	configFile, ok := w.configFiles[path]
+	if ok {
+		return configFile, path, true
+	}
+
+	stat, err := os.Stat(filename)
```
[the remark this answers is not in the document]
dhiaayachi: agreed, I will also add a test for that specific use case

agent/config/file_watcher.go
```diff
@@ -0,0 +1,221 @@
...
+
+		err = w.watcher.Add(filename)
+		if err != nil {
+			w.logger.Error("failed to add file to watcher", "file", filename, "err", err)
+			continue
+		}
+		if configFile.id != newId {
+			w.logger.Debug("call the handler", "filename", filename, "old id", configFile.id, "new id", newId)
+			w.configFiles[filename].id = newId
+			go w.handleFunc(&WatcherEvent{Filename: filename})
```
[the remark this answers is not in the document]
dhiaayachi: see my comment on changing the API

agent/config/file_watcher.go
```diff
@@ -0,0 +1,222 @@
...
+		logger:           logger.Named("file-watcher"),
+		configFiles:      cfgFiles,
+		EventsCh:         make(chan *WatcherEvent),
+		reconcileTimeout: timeoutDuration,
+		done:             make(chan interface{}),
+	}
+	for _, f := range configFiles {
+		err = w.add(f)
+		if err != nil {
+			return nil, err
```
rboyer, line 52:
```suggestion
			return nil, fmt.Errorf("error adding file %q: %w", f, err)
```

agent/config/file_watcher.go
```diff
@@ -0,0 +1,222 @@
...
+	"github.com/fsnotify/fsnotify"
+	"github.com/hashicorp/go-hclog"
+)
+
+const timeoutDuration = 200 * time.Millisecond
+
+type FileWatcher struct {
+	watcher          *fsnotify.Watcher
+	configFiles      map[string]*watchedFile
+	EventsCh         chan *WatcherEvent
```
rboyer, line 19: Should document this public field. It can be useful to keep the exported and un-exported struct members separated out.

Alternatively you could leave this one unexported, and provide an accessor method that forces it to be a receive-only channel:

```
...
    eventsCh chan *WatcherEvent
...
func (w *FileWatcher) EventsCh() <-chan *WatcherEvent { return w.eventsCh }
```

agent/config/file_watcher.go
```diff
@@ -0,0 +1,222 @@
...
+	if err != nil {
+		return nil, err
+	}
+	cfgFiles := make(map[string]*watchedFile)
+
+	w := &FileWatcher{
+		watcher:          ws,
+		logger:           logger.Named("file-watcher"),
+		configFiles:      cfgFiles,
+		EventsCh:         make(chan *WatcherEvent),
```
rboyer, line 45: You probably want this to a the very least be a 1-buffered channel to give it just a bit of concurrency instead of forcing a hard sync between the event channel here and the thing on the other side.

Though really this is going to be gated on how often folks add/remove config files in the config directory, which is going to be on the scale of human-initiated quantities so you could make an argument for just picking a small number like 5-10 and calling that good for now.

agent/config/file_watcher.go
rboyer: Should also (like Start/Stop) describe how to correctly use this channel to avoid blocking this detection loop, something like "be prompt in removing items from the channel" or something to that effect.

agent/config/file_watcher.go
```diff
@@ -0,0 +1,222 @@
...
+	ws, err := fsnotify.NewWatcher()
+	if err != nil {
+		return nil, err
+	}
+	cfgFiles := make(map[string]*watchedFile)
+
+	w := &FileWatcher{
+		watcher:          ws,
+		logger:           logger.Named("file-watcher"),
+		configFiles:      cfgFiles,
```
rboyer, line 44: trivial

```suggestion
	w := &FileWatcher{
		watcher:          ws,
		logger:           logger.Named("file-watcher"),
		configFiles:      make(map[string]*watchedFile),
```

agent/config/file_watcher.go
```diff
@@ -0,0 +1,222 @@
...
+	reconcileTimeout time.Duration
+	cancel           context.CancelFunc
+	done             chan interface{}
+}
+
+type watchedFile struct {
+	modTime time.Time
+}
+
+type WatcherEvent struct {
```
rboyer, line 30: Maybe better to name this `FileWatcherEvent` as it is related to a `FileWatcher` and is in the `agent/config` package instead of its own?

agent/config/file_watcher_test.go
```diff
@@ -0,0 +1,341 @@
...
+		require.NoError(t, err)
+	}()
+	_, err := file.WriteString("test config")
+	require.NoError(t, err)
+	err = file.Sync()
+	require.NoError(t, err)
+	return file.Name()
+}
+
+func createTempConfigDir(t *testing.T, dirname string) string {
```
rboyer, line 328: Why have this helper when the helper does very little?

agent/config/file_watcher.go
rboyer: OK. Usually if there's some sort of handoff between two goroutines I tend to prefer a 1-buffered chan to an un-buffered channel as it breaks the hard interlock. If you've already thought about the hard interlock scenario then it's fine.

agent/config/file_watcher.go
```diff
@@ -0,0 +1,249 @@
...
+func (w *FileWatcher) Start(ctx context.Context) {
+	if w.cancel == nil {
+		cancelCtx, cancel := context.WithCancel(ctx)
+		w.cancel = cancel
+		go w.watch(cancelCtx)
+	}
+}
+
+// Stop the file watcher
+// calling Stop multiple times is a noop
```
rboyer, line 78: Also might be useful to say "Stop must only be called after Start" since the sequence of `Stop, Start, Stop` would leave the goroutine running.

agent/config/file_watcher_test.go
```diff
@@ -0,0 +1,337 @@
...
+	file, err := os.Create(name)
+	require.NoError(t, err)
+	err = file.Close()
+	require.NoError(t, err)
+	w, err := NewFileWatcher([]string{filepath}, hclog.New(&hclog.LoggerOptions{}))
+	require.NoError(t, err)
+	w.Start(context.Background())
+	defer func() {
+		_ = w.Stop()
+	}()
```
rboyer, line 282: trivial; i think this works:

```suggestion
	t.Cleanup(w.Stop)
```

agent/config/file_watcher_test.go
```diff
@@ -0,0 +1,337 @@
...
+
+	_, err = file.WriteString("test config 2")
+	require.NoError(t, err)
+	err = file.Sync()
+	require.NoError(t, err)
+	require.NoError(t, assertEvent(file.Name(), w.EventsCh, defaultTimeout))
+}
+
+func TestEventWatcherRead(t *testing.T) {
```
rboyer, line 75: Is this testing that reading from a file (which may touch `atime`) doesn't notify you?

agent/config/file_watcher_test.go
```diff
@@ -0,0 +1,337 @@
...
+	require.NoError(t, err)
+
+	w, err := NewFileWatcher([]string{file.Name()}, hclog.New(&hclog.LoggerOptions{}))
+	require.NoError(t, err)
+	w.Start(context.Background())
+	defer func() {
+		_ = w.Stop()
+	}()
+
+	file.Chmod(0777)
```
rboyer, line 108: This may error.

agent/config/file_watcher_test.go
```diff
@@ -0,0 +1,337 @@
...
+	defer func() {
+		_ = w.Stop()
+	}()
+
+	require.NoError(t, err)
+	err = os.Remove(filepath)
+	require.NoError(t, err)
+	recreated, err := os.Create(filepath)
+	require.NoError(t, err)
+	_, err = recreated.WriteString("config 2")
```
rboyer, line 128: It's interesting you include this test setup since the order of operations here is actually the "unsafe" one you discuss elsewhere. In this case the filesystem sees:

1. file name=X is erased
2. file name=X is created and is empty
3. file name=X is updated to not be empty

as 3 separate events, i guess here you're only going to see one of these. You could i guess end the test with an assert that times out to prove that you don't wake up more times than you expect.

agent/config/file_watcher_test.go
```diff
@@ -0,0 +1,337 @@
...
+		err = os.Rename(filepath2, filepath)
+		require.NoError(t, err)
+		require.NoError(t, assertEvent(filepath, w.EventsCh, defaultTimeout))
+	}
+}
+
+func TestEventReconcileMove(t *testing.T) {
+	filepath := createTempConfigFile(t, "temp_config1")
+	//make sure mod time is different between the 2 files.
+	time.Sleep(10 * time.Millisecond)
```
rboyer, line 158: Instead of sleeping to do this, you could use `os.Chtimes` to backdate one of the files.

rboyer (approved):
LGTM with some non-blocking remaining comments.

agent/config/file_watcher_test.go
dhiaayachi: yes, this should not trigger any event but just in case another filesystem do something special.

agent/config/file_watcher_test.go
dhiaayachi: TIL!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/590786.

PR #12303, opened 2022-02-10
docs/nia - Get Task API docs & Task Status API deprecations

lornasong:
We are deprecating some fields in Task Status API and replacing them with a new Get Task API

Changes (details in commit msg):
 - new docs for Get Task API (also prepares for Create Task API)
 - deprecate fields in Task Status API
 - update APIs to align on `:task_name` in request parameter (open to feedback)
 - misc fixes

[get status api preview link](https://consul-7qhowv9sm-hashicorp.vercel.app/docs/nia/api/tasks#get-task)

### DIFF
--- a/website/content/docs/nia/api/tasks.mdx
+++ b/website/content/docs/nia/api/tasks.mdx
@@ -211,19 +211,19 @@ $ curl --request DELETE \
 
 The task object is used by the Task APIs as part of a request or response. It represents the task's [configuration information](/docs/nia/configuration#task).
 
-| Name	| Type | Description |
-| ----- | ---- | ----------- |
-| `buffer_period` | object |  The buffer period for triggering task execution. |
-| `buffer_period.enabled` | bool | Whether the buffer period is enabled or disabled. |
-| `buffer_period.min` | string | The minimum period of time to wait after changes are detected before triggering the task. |
-| `buffer_period.max` | string | The maximum period of time to wait after changes are detected before triggering the task. |
-| `condition` | object | The [condition](/docs/nia/configuration#task-condition) on which to trigger the task to execute. <br/><br/> If the task has the deprecated `services` field configured as a module input, it is represented here as `condition.services`. |
-| `description` | string | The human readable text to describe the task. |
-| `enabled` | bool | Whether the task is enabled or disabled from executing. |
-| `module` | string | The location of the Terraform module. |
-| `module_input` | object | The additional [module input(s)](/docs/nia/configuration#task-source-input) that the tasks provides to the Terraform module on execution. <br/><br/> If the task has the deprecated `services` field configured as a module input, it is represented here as `module_input.services`. |
-| `name` | string | The unique name of the task. |
-| `providers` | list[string] | The list of provider names that the task's module uses. |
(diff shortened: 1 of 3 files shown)


findkim (approved):

website/content/docs/nia/api/tasks.mdx
```diff
@@ -122,3 +206,24 @@ $ curl --request DELETE \
   "request_id": "9b23eea7-a435-2797-c71e-10c15766cd73"
 }
 ```
+
+## Task Object
+
+The task object is used by the Task APIs as part of a request or response. It represents the task's [configuration information](/docs/nia/configuration#task).
```
wilkermichael, line 212: Do you think it makes sense to add something about how these fields share the same required fields as what is presented in the `configuration information` ?

Something like:

```suggestion
The task object is used by the Task APIs as part of a request or response. It represents the task's [configuration information](/docs/nia/configuration#task) and shares the same required fields.
```

lornasong: Oh... if I understand you, I think what you're encountering is that if the Create Task API shares this table, it's missing information on which fields are required (and maybe also default values?). Let me know if I misunderstand. What are your thoughts on:

> The task object is used by the Task APIs as part of a request or response. It represents the task's configuration information.
> 
> When configuring this task object for the Create Task API, please refer to the [configuration docs](/docs/nia/configuration#task) for required fields and default values.

Thinking outloud... it might be nice to have the required/default value info on the same page so that users won't have to toggle.

I'm also open to other options, e.g.
A) splitting up and each API has its own table (thought it would be nice to share a table, but we don't have to if it doesn't make for a good ux)
B) add new columns for required and default value to the shared table

What are your thoughts?

wilkermichael: I think it makes sense for them to share an object (in this case a table). They do after all share an object in the API spec including the required fields. That is to say that whoever is consuming the response can also expect, at a minimum, that the fields required would be present.

With that in mind, I don't think it would be confusing to add a required column? Defaults might be nice too, since then when a user receives a response with set values they know they didn't set (i.e. buffer period values) then they understand why the values are present and set to the defaults.

wilkermichael: Oh, and yep! you are correct that my comment was in regards to this being a shared object between Create/Get

lornasong: Thanks! For posterity, we chatted offline:
- I will add a required and default value columns
- I won't edit the blurb above the table
- We don't think the new columns will cause confusion for users reading the Get Task API

lornasong:
@wilkermichael, I updated the Task Object table ([preview here](https://consul-7qhowv9sm-hashicorp.vercel.app/docs/nia/api/tasks#task-object)) with the two new columns to make this more shareable with Create Task API. Would you mind reviewing it?

wilkermichael:
> @wilkermichael, I updated the Task Object table ([preview here](https://consul-7qhowv9sm-hashicorp.vercel.app/docs/nia/api/tasks#task-object)) with the two new columns to make this more shareable with Create Task API. Would you mind reviewing it?

Looks great to me!

PR #12305, opened 2022-02-10
ui: Replace CollapsibleNotices with more a11y focussed Disclosure component

johncowen:
Following on from https://github.com/hashicorp/consul/pull/12304

This replaces our older single use CollapsibleNotices component and helper with our new a11y focussed multi-use Disclosure component.

This mainly started out as one of the proof of concepts for uses of the Disclosure component, having this particular one really helped with the final shape of that component, and its probably good for folks wh have been reviewing this piece of work to see one of the components actually in use rather than just docs.

Note: I've noticed that this isn't very easy to test on a preview site so I've left some handy `or`s in if folks what to checkout the branch and just insert some 'true's in those to test. I'll point it out inline. We can either leave these in here, or just remove after approval (this is only a temporary thing until I get a chance to make this testable/previewable)

---

One last thing is the `route.t` thing that you might notice here. This is a tiny change but I decided to try this out here to see how it felt. The main idea with this is that I want all Routes to be completely self contained so you can move them around however you like (even maybe between applications) without having to think of too much stuff that goes with it. Refering to things like `dc.services.show.topology.even.deeper.thing` is not only long-winded to type, but it's one of the things that makes it difficult to do move routes around, and we use these keys for both routeNames and translation keys. Ideally I just want to write `event.deeper.thing`, the route should already know all the bits before that, and if I move the route somewhere I don't want to change all those keys. The TLDR is it's relative keying. I thought I'd put this up here to see what folks thought. I'm not super happy with it as in order to export something helperlike from a component you need to wrap it in a compute call (for the moment at least, I'm aware of being able to just use helpers in later versions of ember I think?)


ui/packages/consul-ui/app/templates/dc/services/show/topology.hbs
```diff
@@ -27,66 +27,107 @@ as |route|>
...
+        {{! The `false` here is so we can easily open all the notices up by changing to `true` }}
+        {{#let
+          (from-entries (array
+            (array 'filtered-by-acls' (or false topology.FilteredByACLs))
+            (array 'default-allow' (or false (eq dc.DefaultACLPolicy 'allow')))
+            (array 'wildcard-intention' (or false topology.wildcardIntention))
+            (array 'not-defined-intention' (or false topology.notDefinedIntention))
+            (array 'no-dependencies' (or false (and topology.noDependencies (can 'use acls'))))
+            (array 'acls-disabled' (or false (and topology.noDependencies (not (can 'use acls')))))
+          ))
```
johncowen, line 48: If you change all of the `false`s here in the `or`s to `true` it'll show you all panels and therefore the disclosure-ness kicks in (if you can see less than 3 panels you don't get disclosure-ness)

ui/packages/consul-ui/app/templates/dc/services/show/topology.hbs
```diff
@@ -27,66 +27,107 @@ as |route|>
...
+            {{! If more than 2 notices are enabled let the user close them }}
+            {{#if (gt noticesEnabled.length 2)}}
+              <disclosure.Action
+                {{on 'click' disclosure.toggle}}
+              >
+              {{compute (fn route.t 'notices.${expanded}'
+                (hash
+                  expanded=(if disclosure.expanded 'close' 'open')
+                )
+              )}}
```
johncowen, line 120: Heres an example of the `route.t` thing, its a little involved here as I'm using template literal style replacement to avoid using `concat` all the time. But usually it would be just :

```
{{compute (fn route.t 'notices.header')}}
```

I think in later versions of Ember I'll be able to get rid of the `compute` and the `fn`. Anyway thoughts as always appreciated but do keep the "this is to do relative keying" usecase in mind, ty!

johncowen (commented):
Notes:

johncowen:
Rebasing kerfuffle seems to have made tests have borken here, lemme take a proper look see what borked in the rebase.

johncowen:
Unborked, sorry for the noise!

natmegs (approved):
Pulled and tested locally by modifying the conditions. LGTM 👍

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/589191.

Issue #12239, opened 2022-02-01
Make PermissionDeniedError struct more detailed


markan:
The PermissionDeniedError struct is contains only simple text string created at the time of error generation. We'd like to make that data structured to allow better user messages.

Currently we have
```
type PermissionDeniedError struct {
	Cause string // which might be the full error message rendered at the point of creation...
}
```

In many cases we have detailed information about the reason for rejection and would like to capture that.
We should extend with this info. Not all PermissionDenied errors fit that template, so we should create a new structure to capture this. 

```
type PermissionDeniedByACLError struct {
	Accessor string // "token guid"
        Permission string // e.g. "service:read" Perhaps split into resource and level 
        Type string // e.g. service
        Name string // e.g. "sidecar-proxy-1"
        EnterpriseMeta 
}
```

The acl.PermissionDenied helper function usage should also be examined to see where we need to convert to the new form.

A  string conversion function similar to what PermissionDeniedError has should be added.

https://github.com/hashicorp/consul-enterprise/pull/1533

PR #12308, opened 2022-02-10
Refactor to make ACL errors more structured.

markan:
Port of the structured error work in consul-enterprise.

This is the OSS side of #12239

### DIFF
--- a/acl/errors.go
+++ b/acl/errors.go
@@ -82,16 +82,17 @@ type PermissionDeniedError struct {
 // 1) Named entities without a context
 // 2) Unnamed entities with a context
 // 3) Completely context free checks (global permissions)
-// 4) Errors that only have a cause and bad
+// 4) Errors that only have a cause (for example bad token)
 func (e PermissionDeniedError) Error() string {
 	var message strings.Builder
 	message.WriteString(errPermissionDenied)
 
-	// This is used where we
+	// Type 4)
 	if e.Cause != "" {
 		fmt.Fprintf(&message, ": %s", e.Cause)
 		return message.String()
 	}
+	// Should only be empty when default struct is used.
 	if e.Resource == "" {
 		return message.String()
 	}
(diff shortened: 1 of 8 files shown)


acl/errors.go
```diff
@@ -61,18 +61,67 @@ func IsErrPermissionDenied(err error) bool {
 	return err != nil && strings.Contains(err.Error(), errPermissionDenied)
 }
 
+// Arguably this should be some sort of union type.
+// The usage of Cause and the rest of the fields is entirely disjoint.
```
dnephin, line 65: I'm pretty sure we'll be able to make this not entirely disjoint in the future! There are at least 2 places in enterprise where we'll want to use `Cause` along with the other fields. The current implementation doesn't support that yet, but I think that's ok. We can add support when it comes up.

acl/errors.go
```diff
@@ -61,18 +61,67 @@ func IsErrPermissionDenied(err error) bool {
...
+// There are generally three classes of errors
+// 1) Named entities without a context
+// 2) Unnamed entities with a context
+// 3) Completely context free checks (global permissions)
+// 4) Errors that only have a cause and bad
 func (e PermissionDeniedError) Error() string {
+	var message strings.Builder
+	message.WriteString(errPermissionDenied)
+
+	// This is used where we
```
dnephin, line 90: Missing end of comment?

acl/errors.go
```diff
@@ -61,18 +61,67 @@ func IsErrPermissionDenied(err error) bool {
...
+	// e.g. "sidecar-proxy-1"
+	ResourceID ResourceDescriptor
 }
 
+// Initially we may not have attribution information; that will become more complete as we work this change through
+// There are generally three classes of errors
+// 1) Named entities without a context
+// 2) Unnamed entities with a context
+// 3) Completely context free checks (global permissions)
+// 4) Errors that only have a cause and bad
```
dnephin, line 85: maybe missing a word on this last line? "and bad?"

dnephin (commented):
Looks good! Just noticing a few minor things in comments

dnephin (approved):
LGTM!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/581580.

PR #12318, opened 2022-02-11
ensure make dev target puts the newly constructed binary onto the PATH

rboyer:
This restores the prior behavior of make dev and ensures that tests
using the sdk package (like the api package) will correctly locate the
consul binary under test.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -145,7 +145,8 @@ dev: dev-build
 
 dev-build:
 	mkdir -p bin
-	CGO_ENABLED=0 go build -o ./bin -ldflags "$(GOLDFLAGS)" -tags "$(GOTAGS)"
+	CGO_ENABLED=0 go install -ldflags "$(GOLDFLAGS)" -tags "$(GOTAGS)"
+	cp -f ${MAIN_GOPATH}/bin/consul ./bin/consul
 
 dev-docker: linux
 	@echo "Pulling consul container image - $(CONSUL_IMAGE_VERSION)"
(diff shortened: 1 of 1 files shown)


dnephin:
>  ensures that tests using the sdk package (like the api package) will correctly locate the consul binary under test.

I don't believe this goal is accomplished by this PR. It depends on what PATH the user has. If they have consul installed in one of the earlier directories in $PATH, then `make test` has always run the wrong binary.

GNUmakefile
```diff
@@ -145,7 +147,8 @@ dev: dev-build
 
 dev-build:
 	mkdir -p bin
-	CGO_ENABLED=0 go build -o ./bin -ldflags "$(GOLDFLAGS)" -tags "$(GOTAGS)"
+	CGO_ENABLED=0 go install -ldflags "$(GOLDFLAGS)" -tags "$(GOTAGS)"
+	cp -f ${MAIN_GOPATH}/bin/consul ./bin/consul
```
dnephin, line 151: I am still weary about installing things to `PATH` from a target called `build`, but this was the previous behaviour, and it seems you feel strongly about preserving it, so I guess I will have to live with it.

dnephin (approved):
LGTM, thanks for fixing this.

I think just the `export PATH` would be enough to fix the problem, but I guess it doesn't hurt to make the other change as well.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/581037.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 49431c7532958bbb852681e46209598adac38a8a onto `release/1.11.x` succeeded!

PR #12329, opened 2022-02-14
auto-reload configuration when config files change

dhiaayachi:
This add the ability to auto-reload config based on config file change. use `FileWatcher` merged in #12301.

Missing tests, but ready for a first round of review.


agent/agent.go
```diff
@@ -684,6 +692,26 @@ func (a *Agent) Start(ctx context.Context) error {
...
 
+	// start a go routine to reload config based on file watcher events
+	if a.baseDeps.RuntimeConfig.AutoReloadConfig {
+		if a.WatchedFiles != nil {
+			w, err := config.NewFileWatcher(a.WatchedFiles, a.baseDeps.Logger)
+			if err != nil {
+				a.baseDeps.Logger.Error("error loading config", "error", err)
+			}
+			a.FileWatcher = w
+			go func() {
```
rboyer, line 703: What's going to stop this goroutine?

dhiaayachi: The for loop on line 704 would end if the `EventsCh` is closed, which happen when we call `Stop` on the `FileWatcher`.

That said, I'm still missing some code in line 3724 related to this (stop the watcher and create a new routine).

agent/agent.go
```diff
@@ -359,6 +359,10 @@ type Agent struct {
 	// run by the Agent
 	routineManager *routine.Manager
 
+	//FileWatcher is the watcher responsible to report events when a config file
```
rboyer, line 362:
```suggestion
	// FileWatcher is the watcher responsible to report events when a config file
```

agent/agent.go
```diff
@@ -684,6 +695,27 @@ func (a *Agent) Start(ctx context.Context) error {
...
+				for event := range a.FileWatcher.EventsCh {
+					a.baseDeps.Logger.Debug("auto-reload config triggered", "event-file", event.Filename)
+					err := a.ReloadConfig(true)
+					if err != nil {
+						a.baseDeps.Logger.Error("error loading config", "error", err)
+					}
+				}
+			}()
+			a.baseDeps.Logger.Debug("starting file watcher")
+			a.FileWatcher.Start(context.Background())
```
rboyer, line 716: Seems a little strange to start _after_ running the goroutine to consume the events that aren't being generated yet.

agent/agent.go
```diff
@@ -1284,11 +1316,11 @@ func segmentConfig(config *config.RuntimeConfig) ([]consul.NetworkSegment, error
 		if config.ReconnectTimeoutLAN != 0 {
 			serfConf.ReconnectTimeout = config.ReconnectTimeoutLAN
 		}
-		if config.EncryptVerifyIncoming {
-			serfConf.MemberlistConfig.GossipVerifyIncoming = config.EncryptVerifyIncoming
+		if config.NotAutoReloadableRuntimeConfig.EncryptVerifyIncoming {
```
rboyer, line 1319: :bikeshed:
`s/NotAutoReloadableRuntimeConfig/StaticRuntimeConfig/g`

agent/agent.go
```diff
@@ -3700,6 +3737,24 @@ func (a *Agent) ReloadConfig() error {
 	// breaking some existing behavior.
 	newCfg.NodeID = a.config.NodeID
 
+	//if auto reload is enabled, make sure we have the right certs file watched.
+	if autoReload {
+		for _, f := range []struct {
+			oldCfg string
+			newCfg string
+		}{{a.config.KeyFile, newCfg.KeyFile}, {a.config.KeyFile, newCfg.KeyFile}} {
```
rboyer, line 3745: I am very confused because both of these slice tuples are identical.

agent/config/file_watcher.go
```diff
@@ -53,6 +54,7 @@ func NewFileWatcher(configFiles []string, logger hclog.Logger) (*FileWatcher, er
 		reconcileTimeout: timeoutDuration,
 		done:             make(chan interface{}),
 		stopOnce:         sync.Once{},
+		configFilesLock:  sync.RWMutex{},
```
rboyer, line 57:
```suggestion
```

The zero value is sufficient, you don't need to initialize them.

agent/config/file_watcher.go
```diff
@@ -81,12 +83,25 @@ func (w *FileWatcher) Stop() error {
...
 		<-w.done
-		close(w.EventsCh)
 		err = w.watcher.Close()
 	})
 	return err
 }
 
+// Add a file to the file watcher
+// Add will lock the file watcher during the add
+func (w *FileWatcher) Add(filename string) error {
```
rboyer, line 93: I thought in the other PR we decided to not need add/remove here, in favor of passing them all in at construction time?

dhiaayachi: I needed to revise that decision because I did not consider the case where a user change the path to the cert files in the config. In that case we need to remove the old cert files from the watcher and add the new ones.

agent/agent.go
dhiaayachi: yes your right, copy pasta mistake and I need to add a test case!!

agent/config/runtime.go
```diff
@@ -29,6 +29,77 @@ type RuntimeSOAConfig struct {
 	Minttl  uint32 // 0,
 }
 
+// StaticRuntimeConfig specifies the subset of configuration the consul agent actually
+// uses and that are not reloadable by configuration auto reload.
+type StaticRuntimeConfig struct {
```
boxofrad, line 34: Just a heads up that we're shortly going to be splitting the TLS settings here (e.g. `CertFile`) out by listener (e.g. `GRPC.CertFile`, `HTTPS.CertFile`).

Your PR will probably get merged first, so no immediate action required, but I'll add you as a reviewer on mine to make sure I'm doing it right!

dhiaayachi: @boxofrad do you think the approach I took would make sense with the splitting? Are you planning to split those config in different sub struct in the RuntimeConfig or just split the fields?

boxofrad: I need to take a proper look to say for sure, let me get back to you! Plan is for `RuntimeConfig` to have a `tlsutil.Config` struct, with members for each listener `InternalRPC`, `GRPC` and `HTTPS`.

boxofrad: I think it should be fine 🤞🏻 (we'll just put the `tlsutil.Config` field on `StaticRuntimeConfig` instead) but there will be more cert files to watch 😄

.changelog/12329.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+config: add `auto-reload-config` to automatically reload config when a config file change.
```
rboyer, line 2:
```suggestion
config: add `auto-reload-config` to automatically reload config when a config file changes.
```

agent/agent.go
```diff
@@ -3700,6 +3738,27 @@ func (a *Agent) ReloadConfig() error {
...
 
+	//if auto reload is enabled, make sure we have the right certs file watched.
+	if autoReload {
+		for _, f := range []struct {
+			oldCfg string
+			newCfg string
+		}{{a.config.KeyFile, newCfg.KeyFile}, {a.config.CertFile, newCfg.CertFile}} {
+			if f.oldCfg != f.newCfg {
+				a.FileWatcher.Remove(f.oldCfg)
+				err = a.FileWatcher.Add(f.newCfg)
```
rboyer, line 3749: trivial:
I wonder if there's utility in adding a `FileWatcher.Replace(old, new)` command to cut down on the weirdness of this 2-element 2-tuple loop setup.

agent/config/builder.go
```diff
@@ -106,14 +106,17 @@ func Load(opts LoadOpts) (LoadResult, error) {
 	if err := b.validate(cfg); err != nil {
 		return r, err
 	}
-	return LoadResult{RuntimeConfig: &cfg, Warnings: b.Warnings}, nil
+	watcherFiles := make([]string, len(opts.ConfigFiles))
+	copy(watcherFiles, opts.ConfigFiles)
```
rboyer, line 110: trivial:
you could move `structs.CloneStringSlice` to `lib/stringslice` as a new `Clone` method and then use that instead.

agent/config/builder.go
```diff
@@ -1004,16 +1006,26 @@ func (b *builder) build() (rt RuntimeConfig, err error) {
...
+		StaticRuntimeConfig: StaticRuntimeConfig{
+			EncryptVerifyIncoming: boolVal(c.EncryptVerifyIncoming),
+			VerifyIncoming:        boolVal(c.VerifyIncoming),
+			VerifyIncomingHTTPS:   boolVal(c.VerifyIncomingHTTPS),
+			VerifyIncomingRPC:     boolVal(c.VerifyIncomingRPC),
+			VerifyOutgoing:        verifyOutgoing,
+			VerifyServerHostname:  verifyServerName,
+			EncryptVerifyOutgoing: boolVal(c.EncryptVerifyOutgoing),
+			CAFile:                stringVal(c.CAFile),
+			CAPath:                stringVal(c.CAPath),
```
rboyer, line 1018: Can you sort these please? I think generally the RuntimeConfig has sorted fields since there's soooooo many it helps in locating what you want to change.

agent/config/builder.go
```diff
@@ -1072,6 +1084,7 @@ func (b *builder) build() (rt RuntimeConfig, err error) {
 		Services:                    services,
 		SessionTTLMin:               b.durationVal("session_ttl_min", c.SessionTTLMin),
 		SkipLeaveOnInt:              skipLeaveOnInt,
+		AutoReloadConfig:            boolVal(c.AutoReloadConfig),
```
rboyer, line 1087: Similarly to above, can you sort this up with the A-fields?

agent/agent.go
dhiaayachi: I was thinking about this but it make the `FileWatcher` API weird with a replace and no add nor a remove

agent/agent.go
```diff
@@ -3700,6 +3738,27 @@ func (a *Agent) ReloadConfig() error {
...
+				if err != nil {
+					return err
+				}
+			}
+		}
+		if !reflect.DeepEqual(newCfg.StaticRuntimeConfig, a.config.StaticRuntimeConfig) {
+			a.logger.Warn("Static Runtime config has changed and need a manual config reload to be applied", "StaticRuntimeConfig", a.config.StaticRuntimeConfig, "StaticRuntimeConfig From file", newCfg.StaticRuntimeConfig)
```
FFMMM, line 3756: Is this clearer?

```suggestion
			a.logger.Warn("Static Runtime config has changed. Reloading config.", "StaticRuntimeConfig", a.config.StaticRuntimeConfig, "StaticRuntimeConfig From file", newCfg.StaticRuntimeConfig)
```

agent/agent_endpoint.go
```diff
@@ -244,7 +244,7 @@ func (s *HTTPHandlers) AgentReload(resp http.ResponseWriter, req *http.Request)
 		return nil, acl.ErrPermissionDenied
 	}
 
-	return nil, s.agent.ReloadConfig()
```
FFMMM, line 247: Trying to wrap my head around this -- why do we say `false` here ?

agent/config/flags.go
```diff
@@ -76,6 +76,7 @@ func AddFlags(fs *flag.FlagSet, f *LoadOpts) {
 	add(&f.FlagValues.DNSRecursors, "recursor", "Address of an upstream DNS server. Can be specified multiple times.")
 	add(&f.FlagValues.PrimaryGateways, "primary-gateway", "Address of a mesh gateway in the primary datacenter to use to bootstrap WAN federation at start time with retries enabled. Can be specified multiple times.")
 	add(&f.FlagValues.RejoinAfterLeave, "rejoin", "Ignores a previous leave and attempts to rejoin the cluster.")
+	add(&f.FlagValues.AutoReloadConfig, "auto-reload-config", "Auto reload Config on config files change.")
```
FFMMM, line 79:
```suggestion
	add(&f.FlagValues.AutoReloadConfig, "auto-reload-config", "Watches config files for changes and auto reloads the files when modified.")
```

agent/config/builder.go
```diff
@@ -18,6 +18,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/hashicorp/consul/lib/stringslice"
```
FFMMM, line 21: @kisunji was this in line with how we wanted to organize imports?

agent/agent.go
dhiaayachi: Actually this is not accurate, when a `StaticRuntimeConfig` is updated we never reload it. This is by design to avoid deactivating important security features.

agent/agent_endpoint.go
dhiaayachi: The main reason for the flag is to say if we are calling the `ReloadConfig` manually or automatically through auto-config-reload. I can maybe create a wrapper that would underneath call the function with the right flag value to make it more readable.

agent/agent_test.go
```diff
@@ -5328,9 +5328,421 @@ func uniqueAddrs(srvs []apiServer) map[string]struct{} {
...
+	a := NewTestAgent(t, ``)
+	defer a.Shutdown()
+
+	require.False(t, a.Agent.config.AutoReloadConfig)
+}
+
+func TestAgent_AutoReloadDoReload_WhenCertAndKeyUpdated(t *testing.T) {
+	if testing.Short() {
+		t.Skip("too slow for testing.Short")
+	}
```
boxofrad, line 5360: I know that a _lot_ of tests in this package get skipped locally because their side-effects (e.g. writing to the filesystem) make them slow. But I wonder if there's an opportunity to decrease some coupling and improve testability here, for example by making the `FileWatcher` an interface and using an in-memory implementation of [`io/fs`](https://pkg.go.dev/io/fs).

Maybe a bit out of the scope of this PR, so don't feel like you need to address it. I just feel a bit uncomfortable about how deeply we rely on CI and actual sleeping.

boxofrad (commented):
This is looking great. Nice job 👏🏻  Let's pair on rebasing this onto the recent TLS changes.

agent/config/builder.go
kisunji: No it should be in the bottom import group!

The catch with goimports and most other import formatting tools is that it preserves import groups by default. So this line will never be automatically moved to the bottom group because the format tools assume you whitespaced around it for a reason

agent/agent.go
```diff
@@ -3708,6 +3755,39 @@ func (a *Agent) ReloadConfig() error {
...
+				if err != nil {
+					return err
+				}
+			}
+			if revertStaticConfig(f.oldCfg, f.newCfg) {
+				a.logger.Warn("Changes to your configuration were detected that for security reasons cannot be automatically applied by 'auto_reload_config'. Manually reload your configuration (e.g. with 'consul reload') to apply these changes.", "StaticRuntimeConfig", f.oldCfg, "StaticRuntimeConfig From file", f.newCfg)
+			}
+		}
+		if !reflect.DeepEqual(newCfg.StaticRuntimeConfig, a.config.StaticRuntimeConfig) {
+			a.logger.Warn("Changes to your configuration were detected that for security reasons cannot be automatically applied by 'auto_reload_config'. Manually reload your configuration (e.g. with 'consul reload') to apply these changes.", a.config.StaticRuntimeConfig, "StaticRuntimeConfig From file", newCfg.StaticRuntimeConfig)
```
rboyer, line 3785:
```suggestion
			a.logger.Warn("Changes to your configuration were detected that for security reasons cannot be automatically applied by 'auto_reload_config'. Manually reload your configuration (e.g. with 'consul reload') to apply these changes.", "StaticRuntimeConfig", a.config.StaticRuntimeConfig, "StaticRuntimeConfig From file", newCfg.StaticRuntimeConfig)
```

agent/agent_endpoint.go
rboyer: `AutoReloadConfig` and `ReloadConfig` could both call `reloadConfig(bool)`

agent/config/file_watcher.go
```diff
@@ -81,16 +81,38 @@ func (w *FileWatcher) Stop() error {
...
+
+// Replace a file in the file watcher
+// Replace will lock the file watcher during the replace
+func (w *FileWatcher) Replace(oldFile, newFile string) error {
+	if oldFile == newFile {
+		return nil
 	}
+	w.configFilesLock.Lock()
+	defer w.configFilesLock.Unlock()
+	delete(w.configFiles, oldFile)
```
rboyer, line 111: `Replace` calls `add()` at the bottom, which also locks, then this would deadlock itself.

Either make `Replace` non-atomic  (lock, delete, unlock, lock, add, unlock) or refactor Replace and add together so that you only lock once.

agent/config/file_watcher.go
```diff
@@ -179,6 +204,8 @@ func (w *FileWatcher) handleEvent(ctx context.Context, event fsnotify.Event) err
 
 func (w *FileWatcher) isWatched(filename string) (*watchedFile, string, bool) {
 	path := filename
+	w.configFilesLock.RLock()
+	defer w.configFilesLock.RUnlock()
 	configFile, ok := w.configFiles[path]
```
rboyer, line 209:
```suggestion
	w.configFilesLock.RLock()
	configFile, ok := w.configFiles[path]
	w.configFilesLock.RUnlock()
```

Avoid having a lingering `defer unlock` for longer than necessary so you don't deadlock with line 222.

agent/config/file_watcher.go
```diff
@@ -192,14 +219,17 @@ func (w *FileWatcher) isWatched(filename string) (*watchedFile, string, bool) {
 		// try to see if the watched path is the parent dir
 		newPath := filepath.Dir(path)
 		w.logger.Trace("get dir", "dir", newPath)
+		w.configFilesLock.RLock()
+		defer w.configFilesLock.RUnlock()
 		configFile, ok = w.configFiles[newPath]
```
rboyer, line 224:
```suggestion
		w.configFilesLock.RLock()
		configFile, ok = w.configFiles[newPath]
		w.configFilesLock.RUnlock()
```

This also doesn't need to be deferred.

Consider a helper method:

```
func (w *FileWatcher) getConfigFile(path string) (*watchedFile, bool) {
	w.configFilesLock.RLock()
	defer w.configFilesLock.RUnlock()
	configFile, ok := w.configFiles[newPath]
	return configFile, ok
}
```

rboyer (changes requested):
Looking pretty good. I had a few issues I noticed.

agent/agent_test.go
dhiaayachi: I converted the Watcher into an interface for the purpose of implementing a rate limited watcher for the time being. I will revisit those tests later to remove some of them in favour of tests with a mock watcher (I will probably keep 1 or 2 to make sure the whole thing work though)

agent/agent.go
```diff
@@ -692,6 +712,28 @@ func (a *Agent) Start(ctx context.Context) error {
 		{Name: "pre_release", Value: a.config.VersionPrerelease},
 	})
 
+	// start a go routine to reload config based on file watcher events
+	if a.baseDeps.RuntimeConfig.AutoReloadConfig {
+		if len(a.baseDeps.WatchedFiles) > 0 {
```
rboyer, line 717: trivial:
These could be merged into a single conditional and an entire level of indentation removed.

agent/config/file_watcher.go
```diff
@@ -100,25 +107,51 @@ func (w *FileWatcher) add(filename string) error {
...
+	}
+	modTime, err := w.getFileModifiedTime(newFile)
+	if err != nil {
+		return err
 	}
-	return false
+	w.configFilesLock.Lock()
+	defer w.configFilesLock.Unlock()
+	delete(w.configFiles, oldFile)
+	w.configFiles[newFile] = &watchedFile{modTime: modTime}
```
rboyer, line 142: _(purely optional; codegolf)_
You could move just the interaction of the lock and the map into some helpers to keep them tighter, like:
```
func (w *fileWatcher) replaceFile(oldFile, newFile string, modTime time.Time) {
	w.configFilesLock.Lock()
	defer w.configFilesLock.Unlock()
	delete(w.configFiles, oldFile)
	w.configFiles[newFile] = &watchedFile{modTime: modTime}
}

func (w *fileWatcher) addFile(newFile string, modTime time.Time) {
	w.configFilesLock.Lock()
	defer w.configFilesLock.Unlock()
	w.configFiles[filename] = &watchedFile{modTime: modTime}
}

func (w *fileWatcher) removeFile(oldFile string) {
	w.configFilesLock.Lock()
	defer w.configFilesLock.Unlock()
	delete(w.configFiles, oldFile)
}
```

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/619946.

PR #12281, opened 2022-02-07
Update security scanner logic

mdeggies:
Temporarily turn off alpine_secdb container scanning and secret scanning until underlying issues with the CLI are resolved. 

Related thread: https://hashicorp.slack.com/archives/C010VJT0FRP/p1644018582517429?thread_ts=1644016638.974929&cid=C010VJT0FRP

PR #12276, opened 2022-02-04
Update docker image base to alpine:3.15

mdeggies:
This is required, as alpine:3.13 has a number of CVE's (which are all fixed in 3.15)!

PR #12275, opened 2022-02-04
Fix Security Scan

claire-labry:
:facepalm: Ok so I should’ve totally tested out the new `{ secrets = true }` block before merging b/c it completely halted y’alls workflow events lol in crt-workflows-common — I chatted with Kent from Security and he mentioned that the scanner can’t quite read the new block just yet. 

More info about the secrets block in security-scanner [here](https://github.com/hashicorp/security-scanner/pull/154)

This PR reverts the changes in `security-scan.hcl` file and allows the workflows to assume again. 

Workflows are passing: 
[security-scan-binaries:](https://github.com/hashicorp/crt-workflows-common/actions/runs/1796795542) ✅ 
[security-scan-container:](https://github.com/hashicorp/crt-workflows-common/actions/runs/1796801189) ✅ 

NOTE: you’ll see that in the binary stanza, go_modules has been turned to false due to the scanner picking up jwt-go vuln. It's a false positive — Security has a fix in the backlog to fix this but in the meantime, to allow the events in CRT workflows-common to continue, the immediate fix is disabling the scanner for go modules. Once there's a fix in place for this, simply turning `go_modules` back to true will enable go module scanning again. 

Here’s an example of what i’m talking about with consul-ent with [jwt-go issue](https://github.com/hashicorp/crt-workflows-common/runs/5071702944?check_suite_focus=true)

PR #12347, opened 2022-02-15
[1.9.x] Backport CRT-related fixes

eculver:
This is a backport of a number of either failed or unapplied changes relating to CRT and Docker. In particular, these are the PRs that contain the original changes that have already been merged to `main`:

- #12097 - The bulk of changes to fix the production Dockerfile and become compatible with CRT
- #12276 - Update Dockerfile to use `alpine:3.15`
- #11956 - Enable security scanner for Consul binaries and containers
- #12275 - Update security scanner config to work with not-yet-deprecated keys
- #12281 - Temporarily turn off `alpine_secdb` container scanning and secret scanning
- #12147 -  Fix issues in Linux `preremove`/`postremove` package scripts

See example runs of the build workflow [here](https://github.com/hashicorp/consul/actions/workflows/build.yml?query=branch%3Aeculver%2F1.9%2Fdockerfile-fixes)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,6 @@ on:
     branches:
     # Provide the release branch
       - release/1.9.x
-      - eculver/1.9/dockerfile-fixes
 
 env:
   PKG_NAME: consul
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,7 @@ on:
     branches:
     # Provide the release branch
       - release/1.9.x
+      - eculver/1.9/dockerfile-fixes
 
 env:
   PKG_NAME: consul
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -174,7 +175,7 @@ jobs:
     strategy:
       matrix:
         goos: [ darwin ]
-        goarch: [ "amd64" ]
+        goarch: [ "amd64", "arm64" ]
```
kisunji, line 178: darwin/arm64 was only introduced in 1.11.x

eculver: Ah, ok. I was wondering why it was left out of these older release series. Is there a technical reason why it is 1.11.0 and beyond?

kisunji: Nothing technical; the decision was simply to support darwin/arm64 moving forwards, since it doesn't make sense to do a patch release introducing a new architecture as a backport

mdeggies: 100%!

mdeggies (approved):

kisunji (approved):

PR #12350, opened 2022-02-15
ui: Start using mermaid state diagrams in our docs

johncowen:
GitHub recently added mermaid diagram rendering to their markdown previews.

The majority of Consul UI uses XState flavoured state machines to manage state.

This PR tests out using GHs rendering of Mermaid flavoured state machine syntax to use in or internal UI documentation.

Longer term we'll need an XState to Mermaid conversion along with a `doctoc` style git hook embed so that we don't have to manually sync up our XState state machine config with the documentation diagrams. Until that is done, this _could_ be just a one off PoC. We may or may not add more of these as we go.

You can see the GH rendered diagrams at the bottom of [here](https://github.com/hashicorp/consul/blob/a744ea3f216d52a9a5703c9b0ec7e710793521e6/ui/packages/consul-ui/app/components/disclosure/README.mdx). Funnily enough, this GH rendering is currently much better than our own JS spruced up docs where we only show the bare mermaid syntax. Again longer term, we'll want to render these in our own non-GH docs also, but ideally the diagram here would be interactive a la XState's visualizer.

Sometimes a picture is worth a thousand words 🎉

### DIFF
--- a/ui/packages/consul-ui/package.json
+++ b/ui/packages/consul-ui/package.json
@@ -64,7 +64,7 @@
     "@glimmer/tracking": "^1.0.0",
     "@hashicorp/ember-cli-api-double": "^3.1.0",
     "@lit/reactive-element": "^1.2.1",
-    "@mapbox/rehype-prism": "^0.6.0",
+    "@mapbox/rehype-prism": "^0.8.0",
     "@xstate/fsm": "^1.4.0",
     "a11y-dialog": "^6.0.1",
     "babel-eslint": "^10.0.3",
@@ -163,7 +163,7 @@
     "pretty-ms": "^7.0.1",
     "qunit-dom": "^1.0.0",
     "react-is": "^17.0.1",
-    "refractor": "^3.3.1",
+    "refractor": "^3.5.0",
     "remark-autolink-headings": "^6.0.1",
     "remark-hbs": "^0.4.0",
     "sass": "^1.28.0",
@@ -176,6 +176,7 @@
   "resolutions": {
     "ember-basic-dropdown": "^3.0.10",
     "ember-changeset": "3.10.1",
+    "refractor": "^3.5.0",
     "validated-changeset": "0.10.0"
   },
   "engines": {
--- a/ui/packages/consul-ui/.docfy-config.js
+++ b/ui/packages/consul-ui/.docfy-config.js
@@ -1,9 +1,11 @@
 const path = require('path');
 
 const autolinkHeadings = require('remark-autolink-headings');
+const prism = require('@mapbox/rehype-prism');
 const refractor = require('refractor');
 const gherkin = require('refractor/lang/gherkin');
-const prism = require('@mapbox/rehype-prism');
+const mermaid = require('refractor/lang/mermaid');
+const handlebars = require('refractor/lang/handlebars');
 
 const fs = require('fs');
 const read = fs.readFileSync;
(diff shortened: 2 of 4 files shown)


ui/packages/consul-ui/package.json
```diff
@@ -176,6 +176,7 @@
   "resolutions": {
     "ember-basic-dropdown": "^3.0.10",
     "ember-changeset": "3.10.1",
+    "refractor": "^3.5.0",
```
jgwhite, line 179: Why did you need to specify a resolution for refractor? Yarn not figuring out a common version?

johncowen: Oh let me double check that. I had trouble getting a version of `refractor` that had the mermaid lang that came with it for free (they added it in 3.5.0). I _think_ I had to do this to get docfy to internally use the same version and hence pick up the syntax alias additions.

I also couldn't just upgrade docfy (which I think also has an upgraded refractor) as I got a bunch of embroider errors 😬 

I might be completely misremembering tho (and I only did it this morning!) so I'll triple check. Thanks for the spot!

johncowen: Ah right, so in the end it was the `@mapbox/rehype-prism` upgrade that made things work in the end (I did misremember 😆 ). The resolution was a previous attempt. Thanks for the nudge! I just removed it.

(the problem here was, once I'd upgraded refractor, when running docfy it wasn't picking up my refactor syntax aliases. I tried upgrading docfy but I ended up in embroider errors so backed out of that. Messed around with resolutions incase it was somehow using two instances of refractor or something, but then when I figured rehype-prism was in the mix I figured an upgrade might make things magically work there - and it did - but then forgot to remove the resolution)

jgwhite (approved):
This seems reasonable to me. The diff in `yarn.lock` is larger than I would have expected for two minor version bumps but not necessarily an issue. Will leave it to your judgement.

johncowen:
Ugh yeah, its between @mapbox/rehype-prism 0.7 and 0.8 they started to depend on `mrm`:

https://github.com/sapegin/mrm/blob/master/packages/mrm/package.json

mmmmmrmm indeed 😆 

🤔 really don't want a lot of that, feels like `mrm` should be a `devDependency` of the mapbox package not a direct dependency. Doesn't massively surprise me considering... Anyway I'm gonna leave this sit for a little. I only used this package as it was the one used in the examples for tweaking things within the docfy docs. I can see there's a different non-mapbox package so I might take that one for a spin instead.

Thanks again @jgwhite !

P.S.

Oh look looks like from the commit message here that they could have intended to add it as a devDep https://github.com/mapbox/rehype-prism/commit/ddbd86f5848abee9774ab5071204ab5b0bc50e61

Ouch, looking at what the package is I might even be better just rolling my own 🤔

johncowen:
Hey @jgwhite 

So I rolled my own in the end (based off of the mapbox one), its only a few lines of code and definitely not worth depending on for all the fun that comes with that. It's actually quite handy as I had a chance to get my hands dirty with a bit of the transpiling stuff, which will be good knowledge for when I want to add some more features here.

I also noticed that putting the resolutions up to the top-level of our workspace seemed to pin things down better.

Anyway I'm now much happier with this, so I'll leave hanging for a little while incase you wanted another peek and then I'll merge.

Thanks again for the review, top notch! 🙇

Issue #10426, opened 2021-06-18
Services not appearing in UI with ACL turned on if not logged in, despite write policy allowing it


zupalex:
#### Overview of the Issue

Hello, I'm facing an issue which looks kind of similar to [this one](https://github.com/hashicorp/consul/issues/701) which is quite old, but it's at the same time not the same scenario.
We had a set of 3 Consul server running without ACL for a while and it was working fine. In the UI we could see the services that got registered. However after recreating the cluster, turning on ACL, and edit the anonymous token to allow write for services resources with an empty prefix, we cannot see anymore the services in the UI when not logged in. Contacting any server `/catalog/services` endpoint via curl without providing any token returns properly the list of services though, so the edit of the anonymous token policies did work.

#### Reproduction Steps

- Create a 3 servers cluster
- Turn on ACL
- Try to query `catalog/services` anonymously and see it is refused as expected
- Edit the anonymous token with the following policy:
```
Posted to Consul as 'service-readwrite'
service_prefix "" {
  policy = "write"
}
-----------------
curl -X PUT --header "X-Consul-Token: $consul_token" "http://${consul_host}:${consul_port}/v1/acl/token/00000000-0000-0000-0000-000000000002" --data '{"Policies": [ { "Name": "service-readwrite" }]}'
```
- Register some services
- Query the `catalog/services` endpoint anonymously and see that we get the expected list of services
- Check the web UI services page, it returns nothing
- Log in with the management token, services show up

There was no errors in the web browser console.

### Consul info for both Client and Server

Consul servers running version 1.8.4

### Operating system and Environment details

Using the official Consul Docker image from Dockerhub

johncowen:
Hi @zupalex 👋 

Thanks for submitting the report!

I'm guessing that using querying the `/v1/internal/ui/services?dc=your-dc` endpoint via curl also gives you an empty 200 response (`[]`)? Is that right?

Thanks!

zupalex:
Hello. I just tried and indeed it does respond with an empty array and status code 200.

zupalex:
It turns out that the anonymous token needs to also have read access to "node" and/or "agent". Should have been obvious that it needs it since it displays info about the nodes where the service is registered 🙄 
I think though that the UI should display the list of services with their tags. It should basically display what a query to `/catalog/services` shows.

johncowen:
Hi @zupalex , thanks for the additional information.

[...]

PR #12354, opened 2022-02-15
ui: Add more explanatory texts for empty states

johncowen:
Related: https://github.com/hashicorp/consul/issues/10426#issuecomment-868432159

This PR does two things:

1. From a human perspective it updates all our 'empty state' texts to include information on the [ACL policy dispositions](https://www.consul.io/docs/security/acl/acl-rules#policy-dispositions) required to view data on that specific view.
2. Moves all of this related text to our translation files.

Small note here:

I noticed whilst doing this that we seem to be very close to full i18n'ed text in all our Routes now. In fact we are probably at a stage where we should make a move to specifically go ahead and make sure all our Routes are 100% i18n'ed seeing as we must be about 95% there, we may aswell go ahead and get the last 5% done soon-ish.

### DIFF
--- a/ui/packages/consul-ui/app/templates/dc/services/instance/exposedpaths.hbs
+++ b/ui/packages/consul-ui/app/templates/dc/services/instance/exposedpaths.hbs
@@ -7,9 +7,7 @@ as |route|>
 as |item proxy|}}
   <div class="tab-section">
   {{#if (gt proxy.ServiceProxy.Expose.Paths.length 0)}}
-    <p>
-      The following list shows individual HTTP paths exposed through Envoy for external services like Prometheus. Read more about this in our <Action @href={{concat (env 'CONSUL_DOCS_URL') '/connect/registration/service-registration#expose-paths-configuration-reference'}} @external={{true}}>documentation</Action>.
-    </p>
+    {{t 'routes.dc.services.instance.exposedpaths.intro' htmlSafe=true}}
     <Consul::ExposedPath::List
       @items={{proxy.ServiceProxy.Expose.Paths}}
       @address={{or item.Service.Address item.Node.Address}}
@@ -17,9 +15,7 @@ as |item proxy|}}
   {{else}}
     <EmptyState>
       <BlockSlot @name="body">
-        <p>
-          There are no individual HTTP paths exposed through Envoy for external services like Prometheus. Read more about this in our <Action @href={{concat (env 'CONSUL_DOCS_URL') '/connect/registration/service-registration#expose-paths-configuration-reference'}} @external={{true}}>documentation</Action>.
-        </p>
+        {{t 'routes.dc.services.instance.exposedpaths.empty.body' htmlSafe=true}}
       </BlockSlot>
     </EmptyState>
   {{/if}}
--- a/ui/packages/consul-ui/app/templates/dc/services/instance/upstreams.hbs
+++ b/ui/packages/consul-ui/app/templates/dc/services/instance/upstreams.hbs
@@ -53,19 +53,10 @@ as |route|>
             </h3>
           </notice.Header>
           <notice.Body>
-            <p>
-              {{t "routes.dc.services.instance.upstreams.tproxy-mode.body"}}
-            </p>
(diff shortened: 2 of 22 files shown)


ui/packages/consul-ui/app/templates/dc/services/instance/upstreams.hbs
```diff
@@ -53,19 +53,10 @@ as |route|>
...
           <notice.Footer>
-            <p>
-              <Action
-                @href={{concat (env 'CONSUL_DOCS_URL') '/connect/transparent-proxy'}}
-                @external={{true}}
-              >
-                {{t "routes.dc.services.instance.upstreams.tproxy-mode.footer"}}
-              </Action>
-            </p>
+            {{t "routes.dc.services.instance.upstreams.tproxy-mode.footer"}}
```
natmegs, line 59: I think these two might need `htmlSafe=true`?

ui/packages/consul-ui/app/templates/dc/services/instance/exposedpaths.hbs
```diff
@@ -7,19 +7,15 @@ as |route|>
 as |item proxy|}}
   <div class="tab-section">
   {{#if (gt proxy.ServiceProxy.Expose.Paths.length 0)}}
-    <p>
-      The following list shows individual HTTP paths exposed through Envoy for external services like Prometheus. Read more about this in our <Action @href={{concat (env 'CONSUL_DOCS_URL') '/connect/registration/service-registration#expose-paths-configuration-reference'}} @external={{true}}>documentation</Action>.
-    </p>
+    {{t 'routes.dc.services.instance.exposedpaths.intro' htmlSafe=true}}
```
natmegs, line 10: Do we need to pass the `CONSUL_DOCS_URL` variable to the translations?

johncowen: Ah good question! Nah, they are all auto injected here:

https://github.com/hashicorp/consul/blob/9b22300c59c2751e08fd339fc77e1af2f2d7dcf4/ui/packages/consul-ui/app/services/i18n.js#L15-L23

johncowen:
This has a little conflict now, likely to be done after the weekend will ping properly once its ready again

johncowen:
😬 YAML conflicts are far more hairy than they should be. I have a few improvements in the works that I've not been able to PR as yet for various reasons that would help here. Pretty sure this one is correct now, looking at the diff here there was a bit of moving around of yaml-ness.

@natmegs I guess it's ready for review now, but I'll be holding off on the merge for a little while whichever.

evrowe (approved):

johncowen:
🎉 Thanks for the approval!

Just a quick note here, there are a couple of questions outstanding on the textual content here so just gonna leave this hanging a little more until we get clarity on that. We may just merge and follow up with an additional PR or just pop more on that end here. Following up offline.

natmegs (approved):

johncowen:
Just to follow up here, I noticed the test we have added here might not be correct, if I've understood correctly the text we are using here only specifies that you need `service:read` for the service listing page, whereas according to the user below you need a few more policy dispositions in order to see that view.

https://github.com/hashicorp/consul/issues/10426#issuecomment-868414378

> It turns out that the anonymous token needs to also have read access to "node" and/or "agent". Should have been obvious that it needs it since it displays info about the nodes where the service is registered 🙄

So I've checked in with the team to see whether to merge this or wait until we get corrected text (if it does need correcting)

johncowen:
I tried this out and `node:read` is required for viewing the service listing page, so I added that piece of text. Another rebase was required which needed a little bit of manual yaml'ing also, so that all went in the same commit.

Let me know if anything else needs to happen here, otherwise I'll leave it hanging for a little while longer then merge.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/632517.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit e62275642b3dd9a7a64f465fdb117a2b0a1c763f onto `release/1.12.x` succeeded!

PR #12356, opened 2022-02-15
Add GOARM=5 when building on arm

smacfarlane:
This corrects  `build.yml` on `main` to include `GOARM=5` when building on `GOARCH=arm`.  This change was added to the 1.10.x and 1.9.x release branches as part of onboarding to CRT but was missed on main and the 1.11.x branch.  

Signed-off-by: Scott Macfarlane <smacfarlane@hashicorp.com>


eculver:
Sorry, this slipped through the cracks. I remember wondering about this a while back and assuming that there must be something about `main` that didn't require it. BUT, I guess I was wrong :)

eculver (approved):

PR #12364, opened 2022-02-15
structs: remove compat structs

dnephin:
Follow up from #12343

Replace them with un-exported interfaces in the http package, where they are used. Many of the methods were removed from the interface because they were not used here. This helps make code easier to follow by using a smaller interface with only the used methods.

### DIFF
--- a/agent/http.go
+++ b/agent/http.go
@@ -769,13 +769,24 @@ func setLastContact(resp http.ResponseWriter, last time.Duration) {
 }
 
 // setMeta is used to set the query response meta data
-func setMeta(resp http.ResponseWriter, m structs.QueryMetaCompat) {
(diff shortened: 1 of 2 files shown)


dnephin:
There are a bunch of godoc comments that reference the old interfaces. Not sure what to do about those.

kisunji (approved):
LGTM

kisunji:
> There are a bunch of godoc comments that reference the old interfaces. Not sure what to do about those.

Could you give an example?

dnephin:
This might require some enterprise changes, so I wont merge it myself as I won't be able to fix those conflicts.

kisunji:
Okay, thanks! I've been wrangling with these interfaces trying to track down test failures for https://github.com/hashicorp/consul/pull/12507 so I'll try to get this merged in enterprise proper

PR #12366, opened 2022-02-16
Document API ability to URL-decode resource names

jkirschner-hashicorp:
To-Do:

- [ ] Add changelog entry
- [ ] Apply pattern shown in Agent HTTP API docs to all other endpoints once reviewers align on approach

This documentation applies to functionality that will be released with Consul 1.12, so it should not have the `type/docs-cherrypick` label.

### DIFF
--- a/website/content/api-docs/agent/index.mdx
+++ b/website/content/api-docs/agent/index.mdx
@@ -640,7 +640,8 @@ The corresponding CLI command is [`consul join`](/commands/join).
 ### Parameters
 
 - `address` `(string: <required>)` - Specifies the address of the other agent to
-  join. This is specified as part of the URL.
+  join. This is specified as part of the URL as a path segment and
+  [can be URL-encoded](/api-docs#parses-url-encoded-resource-names).
 
 - `wan` `(bool: false)` - Specifies to try and join over the WAN pool. This is
   only optional for agents running in server mode. This is specified as part of
@@ -720,7 +721,9 @@ The corresponding CLI command is [`consul force-leave`](/commands/force-leave).
 
 ### Parameters
 
-- `node` `(string: <required>)` - Specifies the name of the node to be forced into `left` state. This is specified as part of the URL.
+- `node` `(string: <required>)` - Specifies the name of the node to be forced into `left` state.
+  This is specified as part of the URL as a path segment and
+  [can be URL-encoded](/api-docs#parses-url-encoded-resource-names).
 
 - `prune` `(bool: false)` - Specifies whether to forcibly remove the node from the list of members.
   Pruning a node in the `left` or `failed` state removes it from the list altogether. 
--- a/website/content/api-docs/index.mdx
+++ b/website/content/api-docs/index.mdx
@@ -75,6 +75,20 @@ $ curl \
     http://127.0.0.1:8500/v1/kv/foo
 ```
 
+## Parses URL-Encoded Resource Names
+
+Some Consul HTTP API endpoints require resource names in the URL. To pass a resource
+name containing characters that are invalid in a URL path segment, such as `/` or ` `,
+the resource name must be URL-encoded into the HTTP API call URL.
+
+Though any resource name *can* be used with the HTTP API if URL-encoded,
+we typically recommend using resource names that don't require URL-encoding anyways.
+Depending on the validation that Consul applies to a resource name, Consul may still
+reject a request if it considers the resource name invalid for that endpoint. And
+even if Consul considers the resource name valid, it may degrade other functionality,
+such as failed [DNS lookups](/docs/discovery/dns)
+for nodes or services with names containing invalid DNS characters.
+
 ## Translated Addresses
 
 Consul 0.7 added the ability to translate addresses in HTTP response based on


website/content/api-docs/agent/index.mdx
```diff
@@ -720,7 +721,9 @@ The corresponding CLI command is [`consul force-leave`](/commands/force-leave).
 
 ### Parameters
 
-- `node` `(string: <required>)` - Specifies the name of the node to be forced into `left` state. This is specified as part of the URL.
+- `node` `(string: <required>)` - Specifies the name of the node to be forced into `left` state.
+  This is specified as part of the URL as a path segment and
```
jkirschner-hashicorp, line 725: I tried defining this sentence in a partial so it could be defined in one place and referenced everywhere else, but that doesn't seem to work. You can't include a partial inline, so the sentence couldn't be injected as a part of the parameter description itself (it would show up as a separate sentence underneath, unindented).

website/content/api-docs/index.mdx
```diff
@@ -75,6 +75,20 @@ $ curl \
     http://127.0.0.1:8500/v1/kv/foo
 ```
 
+## Parses URL-Encoded Resource Names
```
jkirschner-hashicorp, line 78: I chose the word "Parses" rather than "Accepts" because Consul may still reject the request if the resource name is not valid for that endpoint.

website/content/api-docs/index.mdx
```diff
@@ -75,6 +75,29 @@ $ curl \
     http://127.0.0.1:8500/v1/kv/foo
 ```
 
+## Accepts URL-Encoded Resource Names
```
jkirschner-hashicorp, line 78: Does this section seem to be in the right place on this page? Or should it be higher/lower?

website/content/commands/index.mdx
```diff
@@ -254,3 +254,19 @@ products that support Enterprise namespaces.
 ```
 CONSUL_NAMESPACE=default
 ```
+
+## Accepts URL-Invalid Characters
```
jkirschner-hashicorp, line 258: Does this section seem to be in the right place on this page? Or should it be higher/lower?

website/content/commands/force-leave.mdx
```diff
@@ -44,6 +44,9 @@ are not supported from commands, but may be from the corresponding HTTP endpoint
 
 Usage: `consul force-leave [options] node`
 
+`node` can [contain URL-invalid characters](/commands#accepts-url-invalid-characters),
```
jkirschner-hashicorp, line 47: As far as I can tell, we don't have a place in most of our command docs pages for a CLI parameter (like `node`), only for the `[options]`. I'm open to suggestions on a better way to include information about this capability in the CLI docs.

website/content/api-docs/index.mdx
```diff
@@ -75,6 +75,29 @@ $ curl \
...
+for nodes or services with names containing invalid DNS characters.
+
+This HTTP API capability also allows the
+[CLI to accept arguments with URL-invalid characters](/commands#accepts-url-invalid-characters).
+
+### Exceptions: Characters that are Always Invalid
+
+The inclusion of certain characters in a URL resource name will cause a request
+to be rejected *even if the character is URL-encoded*, including:
+- Linefeed (`%0a`)
```
jkirschner-hashicorp, line 99: @kisunji : could use your eyes for technical accuracy (since you're familiar with this code).

website/content/commands/index.mdx
```diff
@@ -254,3 +254,19 @@ products that support Enterprise namespaces.
...
+and then the [HTTP API endpoint will URL-decode those parameters](/api-docs#accepts-url-encoded-resource-names).
+
+Though any HTTP API resource name *can* be used with the CLI
+([with a few exceptions](/api-docs#exceptions-characters-that-are-always-invalid)),
+we typically recommend using resource names that don't require URL-encoding anyways.
+Depending on the validation that Consul applies to a resource name, Consul may still
+reject a request if it considers the resource name invalid for that endpoint. And
+even if Consul considers the resource name valid, it may degrade other functionality,
+such as failed [DNS lookups](/docs/discovery/dns)
+for nodes or services with names containing invalid DNS characters.
```
jkirschner-hashicorp, line 272: @kisunji : could use your eyes for technical accuracy (since you're familiar with this code).

website/content/api-docs/index.mdx
```diff
@@ -75,6 +75,29 @@ $ curl \
...
 ```
 
+## Accepts URL-Encoded Resource Names
+
+Some Consul HTTP API endpoints require resource names in the URL. To pass a resource
+name containing characters that are invalid in a URL path segment, such as `/` or ` `,
+the resource name must be URL-encoded into the HTTP API call URL.
+
+Though any resource name *can* be used with the HTTP API if URL-encoded,
+we typically recommend using resource names that don't require URL-encoding anyways.
```
kisunji, line 85: I think the first paragraph isn't really necessary - it was a bug that Consul did not decode correctly but _generally_ people would expect most servers to respect URL-encoded paths

```suggestion
Consul HTTP API endpoints can parse URL-encoded resource names but we typically 
recommend using names that will not require URL-encoding.
```

website/content/commands/index.mdx
kisunji, line 272: I think this is redundant actually; people can follow the hyperlink and read the details.

The core piece of information is: you don't need to encode resource names - the CLI will do it for you

karl-cardenas-coding: I agree, we can simplify this a bit as it's a bit verbose.

website/content/commands/index.mdx
karl-cardenas-coding: I'm all for us putting it before the environment variables.  That way we transition from CLI behaviors to environment variables vs jumping back into CLI behavior context after reveiwing env vars.

karl-cardenas-coding (approved):
Left some minor suggestions but overall 👍🏼

kisunji (approved):

website/content/api-docs/index.mdx
jkirschner-hashicorp: Though being concise is important, I don't want to be concise at the expense of assuming what users will know or expect. If our assumptions are wrong for just 25% of users, that's still a lot of users who may be confused.

website/content/api-docs/index.mdx
jkirschner-hashicorp: Per Karl, I only gave this its own section for improved scannability

website/content/commands/index.mdx
jkirschner-hashicorp: I'll submit a more concise version for consideration

website/content/api-docs/index.mdx
trujillo-adam, line 78:
```suggestion
## URL-Encoded Resource Names
```
Maybe it's just me, but the verb at the beginning of this heading was jarring, especially considering that all (or most) of the other headings in this section have a noun-phrase. Ignore if I'm the odd man out on this one.

website/content/api-docs/index.mdx
trujillo-adam: I'm with @kisunji on this one -- especially because there appears to really only be one exception. Maybe change the section heading to something like "Invalid Characters" or something else so that the information would be scannable but topical enough to merit its own section?

website/content/commands/force-leave.mdx
trujillo-adam: Just to back it up a second, does `node` in this section literally mean that you must include the word `node`? Or do we mean `<node_name>`?

website/content/commands/index.mdx
```diff
@@ -254,3 +254,19 @@ products that support Enterprise namespaces.
...
+## Accepts URL-Invalid Characters
+
+Many CLI commands use Consul HTTP API endpoints that require resource names in the request URL.
+To enable the CLI to interact with resources containing characters that are invalid in a URL
+path segment, the CLI will URL-encode parameters to be included in the HTTP API path,
+and then the [HTTP API endpoint will URL-decode those parameters](/api-docs#accepts-url-encoded-resource-names).
+
+Though any HTTP API resource name *can* be used with the CLI
+([with a few exceptions](/api-docs#exceptions-characters-that-are-always-invalid)),
+we typically recommend using resource names that don't require URL-encoding anyways.
```
trujillo-adam, line 267:
```suggestion
we typically recommend using resource names that don't require URL-encoding, anyway.
```

trujillo-adam (commented):
Didn't ask for it, but you got my feedback

jkirschner-hashicorp:
@trujillo-adam : I always want your feedback :) just don't want to bombard with 10 PRs at a time!

website/content/api-docs/agent/index.mdx
jkirschner-hashicorp: Not that I'm aware of. I think I only discovered this feature by stumbling across its usage in a few pages and then looking into how it worked. I don't think there's any guidance in the `website/README.md` file.

website/content/api-docs/index.mdx
jkirschner-hashicorp: I agree. I felt weird about this when I wrote it, too.

website/content/api-docs/agent/index.mdx
trujillo-adam: I can see us backing ourselves into an undesirable position without developing some ground rules. At the very least, we should include a comment describing known sections where the partial has been used. 

Ideally, we would leverage this ability to create fully modular docs that could be assembled into different outputs. If we had conditional processing feature, then we could tag some blocks as "enterprise" or "Kubernetes", for example,  and generate custom sites for different experiences. We would probably need a CMS layer . . . but I digress.

website/content/commands/index.mdx
jkirschner-hashicorp: @kisunji : changing to 

---

## Arguments with URL-Invalid Characters

The CLI automatically URL-encodes arguments, which are then [URL-decoded by the underlying HTTP API endpoints](/api-docs#accepts-url-encoded-resource-names). To avoid double-encoding arguments, do not URL-encode arguments passed to the CLI.

website/content/commands/force-leave.mdx
jkirschner-hashicorp: Changed to ...

---

## Usage

Usage: `consul force-leave [options] <node_name>`

#### Positional Arguments

- `node_name` - The name of the node to be forced into the `left` state. If the name contains URL-invalid characters, [provide the raw name without URL-encoding](/commands#arguments-with-url-invalid-characters).

#### API Options

...

website/content/api-docs/index.mdx
jkirschner-hashicorp: Changed to:

---

### Invalid Characters

The linefeed character (`%0a`) will cause a request to be rejected even if it is URL-encoded.

.changelog/12366.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
+api: URL-encode/decode resource names for all HTTP API endpoints (added in Consul 1.12.0 via #11335, #11957, #12103, #12190, #12297)
```
jkirschner-hashicorp, line 2: @kisunji : how should we handle this changelog entry? The functionality was released in 1.12.0 in the aforementioned PR, but using the normal merge process, this would actually show up in the 1.12.1 section of the changelog (AFAIK).

We're adding an entry for something that already exists (but isn't documented yet), rather than adding/releasing new functionality with this PR.

jkirschner-hashicorp:
@trujillo-adam , @kisunji : if you agree, I'm going to keep this PR limited to changing the CLI and HTTP API main docs pages. I'll remove the example changes to the force leave CLI and HTTP API docs that reference this new capability, since that requires a follow-up conversation with @trujillo-adam.

@kisunji : I need your advice on how to handle the changelog entry.

@trujillo-adam : I believe I've addressed all feedback on the main CLI and HTTP API main docs pages. I just need to remove the changes to the force-leave docs from this PR.

.changelog/12366.txt
jkirschner-hashicorp: We agreed I'd just remove the changelog entry from the PR

website/content/api-docs/agent/index.mdx
```diff
@@ -720,17 +721,19 @@ The table below shows this endpoint's support for
 
 The corresponding CLI command is [`consul force-leave`](/commands/force-leave).
 
-### Parameters
+### Path Parameters
```
jkirschner-hashicorp, line 724: @trujillo-adam and I agreed to split the HTTP API parameters into path parameters and query parameters sections, but I'll move that to a separate PR

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/669000.

PR #12369, opened 2022-02-16
Upload new partner badges and update Consul Integration page.

bear359:
Uploading new partner badges for new partner tiering program in Alliances and updating Consul Integration program page.


karl-cardenas-coding:
@bear359 can you request access to the GitHub `consul` team (Helpdesk) and create a PR that is not from a fork please. We can't determine how this will look without the Vercel preview and I want to ensure we don't release anything that might break the UI.

PR #12370, opened 2022-02-16
Parse datacenter from request

thisisnotashwin:
Parse the value of the datacenter from the create/delete requests for AuthMethods and BindingRules so that they can be created in and deleted from the datacenters specified in the request.

### DIFF
--- a/.changelog/12370.txt
+++ b/.changelog/12370.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-agent: Parse datacenter from Create/Delete requests for AuthMethods and RoleBindings.
+agent: Parse datacenter from Create/Delete requests for AuthMethods and BindingRules.
 ```
--- a/.changelog/12370.txt
+++ b/.changelog/12370.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+agent: Parse datacenter from Create/Delete requests for AuthMethods and RoleBindings.
+```
--- a/agent/acl_endpoint_test.go
+++ b/agent/acl_endpoint_test.go
@@ -1174,7 +1174,7 @@ func TestACL_LoginProcedure_HTTP(t *testing.T) {
 				},
 			}
 
-			req, _ := http.NewRequest("PUT", "/v1/acl/auth-method?token=root", jsonBody(methodInput))
+			req, _ := http.NewRequest("PUT", "/v1/acl/auth-method?token=root&dc=dc1", jsonBody(methodInput))
 			resp := httptest.NewRecorder()
 			obj, err := a.srv.ACLAuthMethodCreate(resp, req)
 			require.NoError(t, err)
@@ -1222,6 +1222,26 @@ func TestACL_LoginProcedure_HTTP(t *testing.T) {
 			methodMap[method.Name] = method
 		})
 
+		t.Run("Create in invalid datacenter", func(t *testing.T) {
+			methodInput := &structs.ACLAuthMethod{
+				Name:        "other",
+				Type:        "testing",
+				Description: "test",
+				Config: map[string]interface{}{
+					"SessionID": testSessionID,
+				},
+				TokenLocality: "global",
+				MaxTokenTTL:   500_000_000_000,
+			}
+
+			req, _ := http.NewRequest("PUT", "/v1/acl/auth-method?token=root&dc=invalid", jsonBody(methodInput))
+			resp := httptest.NewRecorder()
+			_, err := a.srv.ACLAuthMethodCRUD(resp, req)
+			require.Error(t, err)
+			_, ok := err.(BadRequestError)
+			require.True(t, ok)
+		})
+
 		t.Run("Update Name URL Mismatch", func(t *testing.T) {
 			methodInput := &structs.ACLAuthMethod{
 				Name:        "test",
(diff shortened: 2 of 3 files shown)


agent/acl_endpoint_test.go
```diff
@@ -1341,7 +1361,7 @@ func TestACL_LoginProcedure_HTTP(t *testing.T) {
 				BindName:    "web",
 			}
 
-			req, _ := http.NewRequest("PUT", "/v1/acl/binding-rule?token=root", jsonBody(ruleInput))
```
FFMMM, line 1344: hm, are those breaking changes?

as in, if I'm a practitioner on 1.10.x or 1.11.x and don't URL encode the datacenter, will my existing code still work?

It doesn't seem like it right?

rboyer: `parseDC` should also default the datacenter to the current one:
```
// parseDC is used to parse the ?dc query param
func (s *HTTPHandlers) parseDC(req *http.Request, dc *string) {
	if other := req.URL.Query().Get("dc"); other != "" {
		*dc = other
	} else if *dc == "" {
		*dc = s.agent.config.Datacenter
	}
}
```
I don't think these test assertions should need `&dc=dc1`

.changelog/12370.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+agent: Parse datacenter from Create/Delete requests for AuthMethods and RoleBindings.
```
ishustava, line 2:
```suggestion
agent: Parse datacenter from Create/Delete requests for AuthMethods and BindingRules.
```

agent/acl_endpoint_test.go
thisisnotashwin: It is not a breaking change! If no datacenter is provided, it will continue to behave as it previously did. In case someone was passing a query param that referenced a datacenter that was incorrect, that will fail, but that would be desired behavior.

This actually ensures consistent behavior between the read/list requests and the write and delete requests!

thisisnotashwin: Updated the tests @rboyer

agent/acl_endpoint_test.go
```diff
@@ -1222,6 +1222,26 @@ func TestACL_LoginProcedure_HTTP(t *testing.T) {
...
+				Config: map[string]interface{}{
+					"SessionID": testSessionID,
+				},
+				TokenLocality: "global",
+				MaxTokenTTL:   500_000_000_000,
+			}
+
+			req, _ := http.NewRequest("PUT", "/v1/acl/auth-method?token=root&dc=invalid", jsonBody(methodInput))
+			resp := httptest.NewRecorder()
+			_, err := a.srv.ACLAuthMethodCRUD(resp, req)
```
dnephin, line 1239: Not a blocker, just an FYI that we've started to change our tests to do this: #11396

Right now the url path on line 1237 is ignored (the query params are used though), and the test doesn't exercise any of the wrappers around endpoints.

Instead of calling `ACLAuthMethodCRUD` we are moving to call `a.srv.h.ServeHTTP`, which makes the test closer to a real API call.

agent/acl_endpoint_test.go
```diff
@@ -1222,6 +1222,26 @@ func TestACL_LoginProcedure_HTTP(t *testing.T) {
 			methodMap[method.Name] = method
 		})
 
+		t.Run("Create in invalid datacenter", func(t *testing.T) {
```
dnephin, line 1225: Not a blocker, but I believe the purpose of this test is to show we accept a `dc` query parameter, right? Since we don't have a backend interface we can fake, and we don't want to have to setup a whole second server in a different DC the test happens to fail with invalid DC. But I believe the underlying thing we are testing is that `parseDC` is called and sets the DC value correct. The current name suggests it is more about testing validation of the `dc` query parameter, but I think that might be more incidental than the primary purpose of the case.

I would suggest changing the name to something like this:

```suggestion
		t.Run("Create in remote DC", func(t *testing.T) {
```

dnephin (approved):
LGTM, thanks for making this change! 

A couple non-blocking suggestion for the tests. I think it applies to both of the new cases, but I only commented on one.

agent/acl_endpoint_test.go
thisisnotashwin: That is a good catch. I was testing incidental behavior purely for the reasons you have described. Ill rename these tests to reflect that behavior they do test.

agent/acl_endpoint_test.go
thisisnotashwin: 🙏 thanks for the heads up! I don't mind making this change across the file.

agent/acl_endpoint_test.go
thisisnotashwin: i might have spoken too soon! the refactor seems large, and ill punt on it for now. Thanks so much for the heads up!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/588266.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 6e6cd928a22d03d670089fc1c7a2f66af235d318 onto `release/1.10.x` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 6e6cd928a22d03d670089fc1c7a2f66af235d318 onto `release/1.9.x` succeeded!

PR #12372, opened 2022-02-16
Document intermediate_cert_ttl

lkysow:


website/content/docs/agent/options.mdx
```diff
@@ -1396,13 +1396,17 @@ bind_addr = "{{ GetPrivateInterfaces | include \"network\" \"10.0.0.0/8\" | attr
...
-      This value cannot be lower than 1 hour or higher than 1 year.
+      This value cannot be lower than 1 hour or higher than 1 year. Must use hour
+      units or smaller, for example, `1d` is not valid but `24h` is.
 
       This value is also used when rotating out old root certificates from
       the cluster. When a root certificate has been inactive (rotated out)
       for more than twice the _current_ `leaf_cert_ttl`, it will be removed
       from the trusted list.
 
+    - `intermediate_cert_ttl` ((#ca_intermediate_cert_ttl)) The time to live (TTL) for the
```
lkysow, line 1407: @kyhavlov I noticed this wasn't documented. I added this here but there's probably more we should say?

website/content/docs/agent/options.mdx
```diff
@@ -1396,13 +1396,17 @@ bind_addr = "{{ GetPrivateInterfaces | include \"network\" \"10.0.0.0/8\" | attr
       certificate will be requested by a proxy before this limit is reached. This
       is also the effective limit on how long a server outage can last (with no leader)
       before network connections will start being rejected. Defaults to `72h`.
-      This value cannot be lower than 1 hour or higher than 1 year.
+      This value cannot be lower than 1 hour or higher than 1 year. Must use hour
+      units or smaller, for example, `1d` is not valid but `24h` is.
```
trujillo-adam, line 1400:
```suggestion
      This value cannot be lower than 1 hour or higher than 1 year. Must use hour
      units or smaller, for example, `1d` is not valid but `24h` is.
```
I'm confused -- does this param only accept values measured in `h`? If so can we just say it that way and that the minimum is `1h` and the maximum is `8760h`?

trujillo-adam (commented):
Had a few questions

website/content/docs/agent/options.mdx
blake: The supported time format values are specified here <https://www.consul.io/docs/agent/options#config_key_reference>. `h`[our] is the largest unit supported by Go's `time.ParseDuration` function.

lkysow: Given that ☝️ how would you like this worded @trujillo-adam ?

website/content/docs/agent/options.mdx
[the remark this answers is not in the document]
lkysow: Should that be applied to every key here that starts with "The"? I see a lot that say "foo - The something something" that could be "foo - Specifies something something".

As long as we're consistent.

trujillo-adam (approved):

lkysow:
Needed to create https://github.com/hashicorp/consul/pull/12604 to get CLA bot to pass.

PR #12376, opened 2022-02-17
ci: combine 'enhancement' entry type with 'improvement'

eculver:
Prior to this change, we excluded entries with the `release-note:enhancement` heading despite it being a valid entry type. This adds it per the suggestion in [the template that ships with go-changelog](https://github.com/hashicorp/go-changelog/blob/main/cmd/changelog-build/changelog.tmpl#L30-L36).

We'll still need to go backport the 12 entries that have been excluded in their associated release. Future releases will do the right thing by including those that are part of the release.

### DIFF
--- a/.changelog/12376.txt
+++ b/.changelog/12376.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+ci: include 'enhancement' entry type in IMPROVEMENTS section of changelog.
+```
--- a/.changelog/changelog.tmpl
+++ b/.changelog/changelog.tmpl
@@ -22,10 +22,11 @@ FEATURES:
 {{ end -}}
 {{- end -}}
 
-{{- if .NotesByType.improvement }}
+{{- $improvements := combineTypes .NotesByType.improvement .NotesByType.enhancement -}}
+{{- if $improvements }}
 IMPROVEMENTS:
 
-{{range .NotesByType.improvement -}}
+{{range $improvements | sort -}}
 * {{ template "note" . }}
 {{ end -}}
 {{- end -}}


blake (approved):

mikemorris:
There's a potential UX improvement to the `changelog-checker` CI task associated with this, to validate not just that a changelog file exists, but that each entry (files can include multiple) actually generates a release note. I think https://github.com/hashicorp/go-changelog/tree/main/cmd/changelog-pr-body-check should be able to _mostly_ catch this, with the caveat that the list of types is currently hardcoded rather than being configurable to match the changelog template.

eculver:
Ah, I think I see what you're saying. We would have caught this had we validated that entries with this type would have resulted in a generated output for it.

On a related note, I was thinking some changes to our process to manage the `CHANGELOG.md`. Instead of generating it at release time, it would be nice if the file was updated for every PR. We could in effect, achieve what you were suggesting by taking the expected output, ensuring that it exists and then automatically updating the `CHANGELOG.md` to include it. The checker could just check that there was a diff to `CHANGELOG.md`. The bonus is that when we go to release, we don't have to generate a new batch of entries, they will already exist in the `UNRELEASED` heading, and so the release would just update the heading and a new `UNRELEASED` section would be born.

Sorry, that is a bit long-winded, but it's how I would imagine all of this coming together. Maybe @mdeggies has some thoughts as well? 

Either way, I think these updates are outside of the scope of this PR? I'm happy to explore later on as part of the effort I alluded to above.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/588797.

PR #12383, opened 2022-02-17
docs: Hyptothetical release note structure for work streams

david-yu:
Testing out release notes per workstream


trujillo-adam:
I'm a little confused about what's happening. The nav shows version numbers like "0.41.0", but the page that's presented is Consul 1.11.0. Is this intended or is the content for placement only?

david-yu:
@trujillo-adam I was testing this out to see what others think. The content of the Consul K8s releases notes are just placeholder pages copied over from the Core release notes. 

Do you think this outline would make sense? cc @karl-cardenas-coding. The PM team was discussing moving towards this model.

karl-cardenas-coding:
I like this idea too, it's helpful 👍🏼 . But I see what Adam means by it being odd with the overall versioning. That being said, the sub-components iterate faster than core, so we would have this problem without versioning.  @trujillo-adam once the web design shares the design for the breakout of sub-components we can revisit the changelog.

david-yu:
@karl-cardenas-coding I don't know if Consul K8s and Consul ECS are necessarily subcomponents (i.e. they should be listed under Consul Core). I think more thought is needed here. Consul K8s is tied to Core for main releases but API GW and ECS are not at the moment. 

> once the web design shares the design for the breakout of sub-components we can revisit the changelog.

trujillo-adam:
Hey @david-yu , can we close this PR since we've adopted this approach?

PR #12399, opened 2022-02-19
Fix fine-grained blocking queries for Catalog.NodeServices

### LINKED ISSUE #12398: Spurious blocking-query returns on catalog/node queries
#### Overview of the Issue

Our Consul deployment has a small number of Nodes with a large number of Services, and tracks changes to a Node and its contained Services through blocking queries to [`/catalog/node/:node`](https://www.consul.io/api-docs/catalog#retrieve-map-of-services-for-a-node). However, the blocking-query behavior has been unreliable, resulting in a large amount of non-blocking responses (along with a large amount of unnecessary RPC traffic) which has increased significantly as the number of nodes+services in our cluster has grown.

After testing and reading the code, it looks like a blocking query will return immediately if _any_ node or service has been updated in the time since the provided index commit has happened. It appears that the `index` table entry for the `nodes` table does not have any fine-grained entries for individual node IDs, so the query simply takes the max index value across both the `nodes` and `services` values for its query response.

This is a similar issue to the one discussed in #3890 (and which was fixed in #3899) but for a different query pattern.

#### Reproduction Steps

1. Create a dev cluster (`consul agent -dev`)
2. Register client node a:
    ```
    curl -X PUT localhost:8500/v1/catalog/register --data '{"Node": "a", "Address": "http://127.0.0.1"}'
    ```
3. Record node a's current index:
    ```
    index=$(curl -si localhost:8500/v1/catalog/node/a | grep 'X-Consul-Index' | grep -Po \\d+)
    echo $index # => 16
    ```
3. Start a blocking query on node a with its current index, e.g.:
    ```
    curl -s "localhost:8500/v1/catalog/node/a?index=$index"
    ```
4. (in another process) Register client node b:
    ```
    curl -X PUT localhost:8500/v1/catalog/register --data '{"Node": "b", "Address": "http://127.0.0.1"}'
    ```
    Note that the blocking query doesn't return.
5. Attempt another blocking query on node a with the previously-recorded index:
    ```
    curl -si "localhost:8500/v1/catalog/node/a?index=$index"
    ```
    This second blocking query returns immediately, with different values for `ModifyIndex` (16) and `X-Consul-Index` (17).

Expected: second query should block, and the `X-Consul-Index` should be equal to the node's `ModifyIndex`.

wjordan:
Fixes #12398 by adding fine-grained `node.[node]` entries to the `index` table, allowing blocking queries to return fine-grained indexes that prevent them from returning immediately when unrelated nodes/services are updated.

### DIFF
--- a/agent/consul/state/catalog.go
+++ b/agent/consul/state/catalog.go
@@ -17,11 +17,15 @@ import (
 	"github.com/hashicorp/consul/types"
 )
 
-// indexServiceExtinction keeps track of the last raft index when the last instance
-// of any service was unregistered. This is used by blocking queries on missing services.
-const indexServiceExtinction = "service_last_extinction"
+const (
+	// indexServiceExtinction keeps track of the last raft index when the last instance
+	// of any service was unregistered. This is used by blocking queries on missing services.
+	indexServiceExtinction = "service_last_extinction"
 
-const indexNodeExtinction = "node_last_extinction"
+	// indexNodeExtinction keeps track of the last raft index when the last instance
+	// of any node was unregistered. This is used by blocking queries on missing nodes.
+	indexNodeExtinction = "node_last_extinction"
+)
 
 const (
 	// minUUIDLookupLen is used as a minimum length of a node name required before
@@ -36,6 +40,8 @@ var (
 	startingVirtualIP = net.IP{240, 0, 0, 0}
 
 	virtualIPMaxOffset = net.IP{15, 255, 255, 254}
+
+	ErrNodeNotFound = errors.New("node not found")
 )
 
 func resizeNodeLookupKey(s string) string {
@@ -59,7 +65,7 @@ func (s *Snapshot) Nodes() (memdb.ResultIterator, error) {
 
 // Services is used to pull the full list of services for a given node for use
 // during snapshots.
-func (s *Snapshot) Services(node string, entMeta *structs.EnterpriseMeta) (memdb.ResultIterator, error) {
+func (s *Snapshot) Services(node string, entMeta *acl.EnterpriseMeta, peerName string) (memdb.ResultIterator, error) {
 	// TODO: accept non-pointer value
 	if entMeta == nil {
 		entMeta = structs.NodeEnterpriseMetaInDefaultPartition()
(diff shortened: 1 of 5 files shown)


wjordan:
This patch provides a significant impact to overall load in our Consul cluster, which performs a heavy number of blocking queries to [`/catalog/node/:node`](https://www.consul.io/api-docs/catalog#retrieve-map-of-services-for-a-node):

image

image

agent/consul/state/catalog_test.go
```diff
@@ -1455,7 +1455,8 @@ func TestStateStore_EnsureService(t *testing.T) {
 	if err != nil {
 		t.Fatalf("err: %s", err)
 	}
-	if idx != 30 {
```
lukluk, line 1458: this changes is not necessary

wjordan: The test fails without this change, which follows the intended change in behavior. Could you elaborate why you think it's not necessary?

kisunji:
Thanks for this PR! Trying to get another set of eyes on it from the team. There will be some work involved on our side to merge to our enterprise side cleanly but hoping to get back to you by end of next week.

agent/consul/state/catalog.go
```diff
@@ -676,6 +676,9 @@ func (s *Store) deleteNodeTxn(tx WriteTxn, idx uint64, nodeName string, entMeta
 	if err := catalogUpdateNodesIndexes(tx, idx, entMeta); err != nil {
 		return fmt.Errorf("failed updating index: %s", err)
 	}
+	if err := catalogUpdateNodeIndexes(tx, nodeName, idx, entMeta); err != nil {
+		return fmt.Errorf("failed updating node index: %s", err)
+	}
```
dnephin, line 681: I think for deletes we probably need to do something like `catalogUpdateServiceExtinctionIndex`. If we track an index for every deleted node, that will grow forever. We'll keep an index for every deleted node name, which is unbounded.

I think by using an extinction index for deleted items we only ever have 1 entry for every deleted node. It's slightly less optimal in terms of reducing blockingQuery churn, but it is much safer in terms of space used by the index tracking.

dnephin (commented):
Nice, thank you for the PR!  I think at a high level this is a good approach. I haven't looked to see if `catalogUpdateNodeIndexes` is called in all necessary places, but generally it seems right.

Left one comment about dealing with deleted nodes.

agent/consul/state/catalog.go
wjordan: Yes, I noticed that extra 'extinction index' used by the `services` query implementation for this purpose, and punted on it here since it didn't affect my use-case (the nodes in our cluster are pretty stable), I agree it's necessary in general though. I can work on adding it sometime soon.

wjordan: OK, my first attempt at this is in 440db29cb1e7326e1abbbb6239c6cbb64718d7bb, but it's largely a copy-paste of `catalogUpdateServiceExtinctionIndex` and related methods (this code is all unfamiliar territory to me). I haven't tested beyond the existing unit tests, and could use some guidance on any extra testing/refactoring that might be helpful for this extra part.

kisunji:
@wjordan are you okay with me force-pushing to your upstream? I've rebased from main and fixed a bunch of conflicts to prep for merge

PR #12404, opened 2022-02-21
ui: PagedCollection component

johncowen:
Since day 1 of the Consul UI rebuild we've used an ember addon to give us virtual scrolling through lists of 1000s of 'items'. This addon has unfortunately been the root of many blockers/bugs for us in doing certain things in the UI due to the amount of absolute positioning it requires to function (and it uses ~2018 ember).

The component in this PR is a much much simpler approach to virtual-scrolling, and the component itself is built to allow multiple approaches to virtual-scrolling, and hopefully allow us to eventually use a non-absolutely positioned approach to virtual scrolling across the UI and therefore remove the ember addon.

The approach used here and easily used by default is to set a `--paged-top` CSS custom property on an element that contains the scrollable items, that element will then have a 'buffer' `::before` pseudo element to push the rest of the content down to its correct position in the scrolled container.

[Rendered Docs](https://consul-ui-staging-bosvgbbrp-hashicorp.vercel.app/ui/docs/components/paged-collection#CONSUL_NODE_COUNT=100;)

Notes:
- You could completely ignore this pseudo element buffer approach by using the `startHeight` and `totalHeight` properties to construct a different approach (I've tried many many many alternatives, this felt like the best)
- Re: https://github.com/hashicorp/consul/pull/12203#pullrequestreview-864997308 this is one real-world use case. Setting a CSS custom property on an element using a value calculated via Javascript, that property then gets picked up by CSS and used to set something else, in this case the `height` of the `::before`. Also note this also takes advantage of native CSS properties that cascade rather than SASS properties that don't - yey modern web platform.
- The PagedCollection component currently includes an `index` mode/type (a.k.a. traditional pagination) that was easy enough to include here. Right now we don't need this, but it's there to be polished a little and used if we do.
- There is a bit more work I'd like to do here to make the component similar to our Search components which follow a bridge like pattern of being able to provide search approaches across different data models. In this component the consumer of the component would be able to load up/inject completely custom paging strategies if they wanted to. As we don't need to do this right now, I've left it for a fun PR for the future.


amyrlam (approved):
neat, functionality is cool

johncowen:
Awesome, gonna merge the implementation down here first then merge-aroo from that (the other one has approval from @randallmorey already). Thanks all. This run of work (disclosures, menus, scrollers) is done for the minute, onto the next 🎉

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/596581.

PR #12396, opened 2022-02-19
Backport 1.10.x: tproxy http upstream fix

kyhavlov:
Backporting some test changes to fix the 1.10.x branch. Slight differences from #12395 because of partitions being added in 1.11.

PR #12407, opened 2022-02-22
fix the fix for the backport

rboyer:


rboyer:
Fixing using https://github.com/hashicorp/consul/pull/12396 instead

PR #12413, opened 2022-02-22
[1.11.x] xds: fix some test assertions for enterprise

rboyer:
This will need to be backported slightly differently to 1.10.x

This forces the oss/ent generated xds to be identical for things in the default partition and default namespace wherever a proxycfg map key is addressed as a `ServiceName` flavored string instead of the upstream-id ones used elsewhere.

Properly fixing that would require backporting https://github.com/hashicorp/consul/pull/12125

### DIFF
--- a/agent/xds/listeners_test.go
+++ b/agent/xds/listeners_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"path/filepath"
 	"sort"
+	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/592220.

PR #12415, opened 2022-02-22
[1.10.x] xds: fix some test assertions for enterprise

rboyer:
Backport of #12413 to `release/1.10.x`

The fix is slightly different to account for there not being partitions in 1.10.


kisunji (approved):

rboyer:
Still trying to grok this, but it might not be necessary on 1.10.x...

rboyer:
Actually not necessary.

PR #12307, opened 2022-02-10
server: partly fix config entry replication issue that prevents replication in some circumstances

rboyer:
There are some cross-config-entry relationships that are enforced during
"graph validation" at persistence time that are required to be
maintained. This means that config entries may form a digraph at times.

Config entry replication proceeds in a particular sorted order by kind
and name.

Occasionally there are some fixups to these digraphs that end up
replicating in the wrong order and replicating the leaves
(ingress-gateway) before the roots (service-defaults) leading to
replication halting due to a graph validation error related to things
like mismatched service protocol requirements.

This PR changes replication to give each computed change (upsert/delete)
a fair shot at being applied before deciding to terminate that round of
replication in error. In the case where we've simply tried to do the
operations in the wrong order at least ONE of the outstanding requests
will complete in the right order, leading the subsequent round to have
fewer operations to do, with a smaller likelihood of graph validation
errors.

This does not address all scenarios, but for scenarios where the edits
are being applied in the wrong order this should avoid replication
halting.

Fixes #9319

The scenario that is NOT ADDRESSED by this PR is as follows:

1. create: service-defaults: name=new-web, protocol=http
2. create: service-defaults: name=old-web, protocol=http
3. create: service-resolver: name=old-web, redirect-to=new-web
4. delete: service-resolver: name=old-web
5. update: service-defaults: name=old-web, protocol=grpc
6. update: service-defaults: name=new-web, protocol=grpc
7. create: service-resolver: name=old-web, redirect-to=new-web

If you shutdown dc2 just before (4) and turn it back on after (7)
replication is impossible as there is no single edit you can make to
make forward progress.

PR #12437, opened 2022-02-24
[1.9.x] server: partly fix config entry replication issue that prevents replication in some circumstances

rboyer:
Backport of #12307 to 1.9.x

### DIFF
--- a/agent/consul/config_replication.go
+++ b/agent/consul/config_replication.go
@@ -7,8 +7,10 @@ import (
 	"time"
 
 	"github.com/armon/go-metrics"
-	"github.com/hashicorp/consul/agent/structs"
 	"github.com/hashicorp/go-hclog"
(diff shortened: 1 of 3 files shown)


agent/consul/config_replication.go
```diff
@@ -98,12 +101,13 @@ func (s *Server) reconcileLocalConfig(ctx context.Context, configs []structs.Con
 		}
 
 		resp, err := s.raftApply(structs.ConfigEntryRequestType, &req)
-		if err != nil {
-			return false, fmt.Errorf("Failed to apply config %s: %v", op, err)
+		if err == nil {
```
rboyer, line 104: This section had to be different since there are TWO error branches on 1.9 instead of just the one.

kisunji (approved):

PR #12444, opened 2022-02-24
catalog: compare node names case insensitively in more places

rboyer:
Many places in consul already treated node names case insensitively.
The state store indexes already do it, but there are a few places that
did a direct byte comparison which have now been corrected.

One place of particular consideration is ensureCheckIfNodeMatches
which is executed during snapshot restore (among other places). If a
node check used a slightly different casing than the casing of the node
during register then the snapshot restore here would deterministically
fail. This has been fixed.

Primary approach:

    git grep -i "node.*[!=]=.*node" -- ':!*_test.go' ':!docs'
    git grep -i '\[[^]]*member[^]]*\]
    git grep -i '\[[^]]*\(member\|name\|node\)[^]]*\]' -- ':!*_test.go' ':!website' ':!ui' ':!agent/proxycfg/testing.go:' ':!*.md'

### DIFF
--- a/.changelog/12444.txt
+++ b/.changelog/12444.txt
@@ -1,0 +1,7 @@
+```release-note:bug
+compare node names case insensitively in more places
+```
+
+```release-note:bug
+state: fix bug blocking snapshot restore when a node check and node differed in casing of the node string
+```
--- a/.changelog/12444.txt
+++ b/.changelog/12444.txt
@@ -1,5 +1,5 @@
 ```release-note:bug
-compare node names case insensitively in more places
+catalog: compare node names case insensitively in more places
 ```
 
 ```release-note:bug
--- a/agent/structs/structs.go
+++ b/agent/structs/structs.go
@@ -452,7 +452,7 @@ func (r *RegisterRequest) ChangesNode(node *Node) bool {
 
 	// Check if any of the node-level fields are being changed.
 	if r.ID != node.ID ||
-		r.Node != node.Node ||
+		!strings.EqualFold(r.Node, node.Node) ||
 		r.PartitionOrDefault() != node.PartitionOrDefault() ||
 		r.Address != node.Address ||
 		r.Datacenter != node.Datacenter ||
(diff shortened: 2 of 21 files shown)


.changelog/12444.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:bug
+compare node names case insensitively in more places
```
mkeeler, line 2:
```suggestion
catalog: compare node names case insensitively in more places
```

Maybe this is an appropriate component. It does apply more generally but I think this could be close enough.

agent/structs/structs.go
```diff
@@ -452,7 +452,7 @@ func (r *RegisterRequest) ChangesNode(node *Node) bool {
 
 	// Check if any of the node-level fields are being changed.
 	if r.ID != node.ID ||
-		r.Node != node.Node ||
+		!strings.EqualFold(r.Node, node.Node) ||
```
mkeeler, line 455: Going down the testing rabbit hole a bit but we should probably have tests for this behavior modifications in this file as well.

command/rtt/rtt.go
```diff
@@ -103,11 +103,11 @@ func (c *cmd) Run(args []string) int {
 		var area1, area2 string
 		for _, dc := range dcs {
 			for _, entry := range dc.Coordinates {
-				if dc.Datacenter == dc1 && entry.Node == node1 {
+				if dc.Datacenter == dc1 && strings.EqualFold(entry.Node, node1) {
```
mkeeler, line 106: Ditto about some tests for this changes.

mkeeler (commented):
I think I found a couple more places that could use tests.

The isSerfMember in agent/consul/util.go could potentially as well but we don't test that method at all right now.

agent/structs/structs.go
rboyer: done

mkeeler (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/595905.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 957146401ef0f55d4ddf82ffffb9eb9a0119824a onto `release/1.11.x` succeeded!

Issue #11947, opened 2022-01-05
rpc: improve RPC logging


dnephin:
Building on #10250 (using the interceptor proposed in #10574) we can make Consul easier to operate by adding logging for every RPC request.  Initially these could be DEBUG logs, but we might also consider making them INFO level.

On each request we should log the RPC call, along with all of the labels used on the metric for #10250.

In addition to those labels, we should add identifiers for the request. Each request type should implement an interface that returns the full identifier with namespace and partition. For example:
* `RegisterRequest` would return an identifier with `ID`, and if applicable `CheckID`, and `ServiceID`.
* `CheckServiceNode` would return a similar identifier
* `KVSRequest` would return an identifier with `DirEntry.Key`
* `ConfigEntryRequest` would return an identifier with `Entry.Kind` and `Entry.Name`
* etc

We may also want to include other identifiers, such as KV operation.

To gather all these identifiers for the log entry, we could implement the following interface on all request types:

```go
type RequestIdentifier interface {
    // RequestID returns an identifier for the entity being written (for writes) or requested (for reads)
    RequestID() string
    // RequestMetadata may return pairs of request metadata used to log the request.
    RequestMetadata() []string
}
```

freddygv:
An additional consideration here are raft apply operations that bypass the RPC interface. 

For example, the leader server [reconciles members](https://github.com/hashicorp/consul/blob/4cd2542a3e4bc0a204eb2a950d873a2123bfbe0b/agent/consul/leader.go#L963) based on events from Serf. These are applied through raft directly and can unblock Health RPC queries. 

Another example are resources persisted via replication, such as config entries and ACLs.

PR #12445, opened 2022-02-24
wip: start instrumenting calls with new rpc metrics

dnephin:
There are 3 places we need to instrument:
* `net/rpc.Server.ServeRequest` - uses the new `net/rpc` interceptor
* gRPC interceptor - no OSS code uses this yet, so will likely be done in enterprise first, nothing done in this PR
* any important `raft.Apply` calls that come from a leader goroutine (not from an RPC endpoint) - this PR enumerates all of those locations with a TODO, and implements the TODO by calling observe in most of them.

The `raft.Apply` calls from leader routines mostly fall into the following categories:
* gossip `reconcileMembers`
* replication in secondary DCs
* some form of garbage collection (ACL token TTL, session TTL, expired CA roots, KV tombstones, etc)
* bootstrapping (config entries, ACL tokens, etc)

### DIFF
--- a/agent/consul/acl_replication.go
+++ b/agent/consul/acl_replication.go
@@ -354,17 +354,17 @@ func (s *Server) fetchACLTokens(lastRemoteIndex uint64) (*structs.ACLTokenListRe
 }
 
 func (s *Server) replicateACLTokens(ctx context.Context, logger hclog.Logger, lastRemoteIndex uint64) (uint64, bool, error) {
-	tr := &aclTokenReplicator{}
+	tr := &aclTokenReplicator{rpcObserver: s.rpcObserver}
 	return s.replicateACLType(ctx, logger, tr, lastRemoteIndex)
 }
 
 func (s *Server) replicateACLPolicies(ctx context.Context, logger hclog.Logger, lastRemoteIndex uint64) (uint64, bool, error) {
-	tr := &aclPolicyReplicator{}
+	tr := &aclPolicyReplicator{rpcObserver: s.rpcObserver}
 	return s.replicateACLType(ctx, logger, tr, lastRemoteIndex)
 }
 
 func (s *Server) replicateACLRoles(ctx context.Context, logger hclog.Logger, lastRemoteIndex uint64) (uint64, bool, error) {
-	tr := &aclRoleReplicator{}
+	tr := &aclRoleReplicator{rpcObserver: s.rpcObserver}
 	return s.replicateACLType(ctx, logger, tr, lastRemoteIndex)
 }
 
--- a/agent/consul/acl_replication_types.go
+++ b/agent/consul/acl_replication_types.go
@@ -3,6 +3,9 @@ package consul
 import (
 	"context"
 	"fmt"
+	"time"
+
+	"github.com/hashicorp/consul/agent/rpc"
 
 	"github.com/hashicorp/consul/agent/structs"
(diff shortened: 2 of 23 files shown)


dhiaayachi:
@dnephin, @FFMMM, General question about this. 
If the intent here is to observe the `raftApply` calls, why not embed the call of the observe inside the apply. The overhead of the request  creation and the call seems negligible and it would avoid polluting all the endpoints code with the observer calls code. Am I missing the point here?

dnephin:
> If the intent here is to observe the raftApply calls, why not embed the call of the observe inside the apply. Am I missing the point here?

Ya, good question! There's some context in [this comment](https://github.com/hashicorp/consul/issues/11947#issuecomment-1006101517). The godoc for `rpcTypeLeader` in this PR also tries to explain it a bit.

To summarize it:

At a high level, `raft.Apply` can be called from two places:
   1. an RPC endpoint (could be net/rpc, or gRPC)
   2. some goroutine that runs in the leader. The description of this PR calls out the 4 high level types of goroutines, but as you can see, there are actually many call sites.

Capturing the metrics/logs for "RPC endpoints" is easy. We can do that from a single interceptor for net/rpc, and another one for GRPC.

Capturing the metrics/logs for "goroutines that runs in the leader" is not as easy. All of the calls you see in this PR are for that second case.  

Every call to `raft.Apply` should not be followed by `Observe`, only calls from `"goroutines that run on the leader" need this extra call, because those are not already wrapped by an RPC interceptor. 

> why not embed the call of the observe inside the apply? The overhead of the request creation and the call seems negligible and it would avoid polluting all the endpoints code with the observer calls code. 

We can't simply put the call to `Observe` into `Server.raftApply` because then we would double count every RPC write request. Any call to `raftApply` from an RPC endpoint is already wrapped by the interceptor. We want to include the RPC request time when it's relevant.

Ya, building a struct is negligible time, it's not important to capture that. You'll see in some cases though (ex: the calls from `reconcileMembers`) that there is a little more work being done that would be good to capture.

But to your point, yes, there is a second approach we could use for the "leader goroutine calls to `raft.Apply`". That approach would be to create another wrapper around `Server.raftApply` called something like `Server.raftApplyWithObserve`, which would do both the `raft.apply` and the call to observe. 

That approach could work, but I don't like it for 2 reasons:
1. as we add more entities that use protobuf we will need to start using `raftApplyProtobuf` and `raftApplyMsgpack` instead of just `raftApply`, this means we'd end up with 4 wrappers around `raft.Apply`. This kind of API is really hard to use properly.
2. calling `raftApplyWithObserve` from an RPC endpoint would result in double counting the request in metrics. Trying to document how to use `raftApplyWithObserve` correctly is a challenge and seems like it would be very error prone.

Considering the only downside to the approach in this PR is a few extra calls calls to `Observe` (which one might argue is more explicit), I think the downsides of the alternative are worse, and that this approach is the better option.

dhiaayachi:
Thank you for the explanation @dnephin !!

My main concern about this approach is how it would evolve in the future when we add new calls. ( We need to remember to add the  observers to the right calls)

I think this can be alleviated by documenting how we chose the calls which need to be observed and not.

FFMMM:
Hey @dnephin thanks loads for trail blazing here. At this stage, a lot of this project and work done here has been completed 🥳 . 

I'll close this PR now.

PR #12462, opened 2022-02-28
catalog: compare node names case insensitively in ACL authz evaluation

rboyer:
Extension of #12444 but with the slightly backwards incompatible change
of having node and agent resources be case insensitive.

### DIFF
--- a/acl/authorizer.go
+++ b/acl/authorizer.go
@@ -21,6 +21,10 @@ const (
 	Default
 )
 
+func (d EnforcementDecision) GoString() string {
+	return d.String()
+}
+
 func (d EnforcementDecision) String() string {
 	switch d {
 	case Allow:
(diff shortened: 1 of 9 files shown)


rboyer:
Comment from @mkeeler from another thread:

> Would the ACL changes here be a potentially breaking change?
> 
> I think the only way it could cause trouble would be for the user to previously have had two rules for a node name using different cases.
>     
>     node "foo" {
>     policy = "write"
>     }
>     
>     node "Foo" {
>        policy = "deny"
>     }
>     
> Before writing the node “foo” would only consider the `write` rule but now I think the overall permission would be `deny` due to `deny`  taking precedence over all other permissions for a given leaf in the radix tree.

acl/authorizer.go
```diff
@@ -21,6 +21,10 @@ const (
 	Default
 )
 
+func (d EnforcementDecision) GoString() string {
```
rboyer, line 24: This helps render the actual string value in testify comparison errors instead of their integer values.

markan (approved):
Generally looks good

PR #12463, opened 2022-02-28
Improve consistency mode docs

jkirschner-hashicorp:
Changes include:
- Add diagrams of the operation of different consistency modes
- Note that only stale reads benefit from horizontal scaling
- Increase scannability with headings

[Preview - consistency modes page](https://consul-ivhvcy9uc-hashicorp.vercel.app/api-docs/features/consistency)

### To-Do (after review)

- [x] Correct legend (should say "Consul Server: Follower" not "Consul Client: Follower")
- [x] Minor visual cleanup of images (e.g., ensure layering of components is correct, add white background instead of transparent to be readable in dark mode)
- [x] Add SVG files corresponding to the exported PNG files
- [x] Add mention in the multi-DC section that the diagram shows direct connection from servers, and that mesh gateways are another option for control plane traffic
- [x] Cleanup commits (squash/reword for clean history)


website/content/api-docs/features/consistency.mdx
```diff
@@ -37,15 +37,64 @@ The three read modes are:
...
+When making a request across federated Consul datacenters, requests are forwarded from
+a local server to any remote server. Once in the remote datecenter, the request path
+is the same as a [local request with the same consistency mode](#intra-datacenter-request-behavior).
+
+[![consistency mode behavior across Consul datacenters](/img/consistency-modes/consistency-mode-cross-datacenter.png)](/img/consistency-modes/consistency-mode-cross-datacenter.png)
+
+## Usage
+
+All read requests use the `default` consistency mode unless overridden.
+To specify a different consistency mode, include the desired consistency mode
```
jkirschner-hashicorp, line 85: There's an extra space before the em dash in the rendered version for some reason

jkirschner-hashicorp:
@hannahhearth, @jnwright : FYI - adding some more diagrams to the docs site (this time for consistency modes).

website/content/api-docs/features/consistency.mdx
```diff
@@ -37,15 +37,101 @@ The three read modes are:
...
+When making a request across federated Consul datacenters, requests are forwarded from
+a local server to any remote server. Once in the remote datecenter, the request path
+is the same as a [local request with the same consistency mode](#intra-datacenter-request-behavior).
+
+[![consistency mode behavior across Consul datacenters](/img/consistency-modes/consistency-mode-cross-datacenter.png)](/img/consistency-modes/consistency-mode-cross-datacenter.png)
+
+## Usage
+
+### Consistency Mode Defaults and Overriding
+
```
jkirschner-hashicorp, line 85: I'll probably add a brief summary sentence under this `###` section heading after a reviewer checks the `####` subsections below for technical accuracy.

website/content/api-docs/features/consistency.mdx
```diff
@@ -37,15 +37,101 @@ The three read modes are:
...
+to a value greater than zero in the configuration of the Consul agent receiving requests,
+the `stale` consistency mode will be used by default unless the data is sufficiently stale.
+Setting a positive value for [`discovery_max_stale`](/docs/agent/options#discovery_max_stale)
+is analogous to changing the default consistency mode for
+[Consul DNS queries](#consul-dns-queries) by setting [`dns_config.max_stale`](/docs/agent/options#max_stale).
+
+The HTTP API documentation for each endpoint lists which consistency modes can be used
+to override the default on a per-request basis. To use a supported consistency mode
+other than `default` for a request,
+include the desired consistency mode—`stale` or `consistent`—as a
```
jkirschner-hashicorp, line 122: If `discovery_max_stale` is >0, is it accurate to say that the consistency mode is `stale` (assuming the Raft entry log index isn't too far behind the leader)?

And if so, if `discovery_max_stale` is in use, does that mean there's no way to force the `default` consistency mode on a request? If I understand correctly, there's no query param that sets `default` consistency mode.

website/content/api-docs/features/consistency.mdx
```diff
@@ -37,15 +37,101 @@ The three read modes are:
...
+set [`dns_config.allow_stale`](/docs/agent/options#allow_stale) to `false`
+to make Consul use the `default` consistency mode when interfacing with the
+underlying Consul service discovery HTTP APIs.
+
+The consistency mode underlying Consul DNS queries cannot be overridden on a
+per-request basis.
+
+#### Consul HTTP API Queries
+
+All HTTP API read requests use the `default` consistency mode by default
```
jkirschner-hashicorp, line 106: I should probably clarify that no write requests support defining a consistency mode, and that all writes behave using `consistent` consistency mode.

website/content/api-docs/features/consistency.mdx
jkirschner-hashicorp: @banks : given your summary, it sounds like the start of the `*max_stale` should be something like **not recommended for most use cases** along with a brief description of why (a condensed version of your first bullet in the summary).

> Caveats: there is a reasonable chance I've misremembered some of these details and the above is technically inaccurate in some way.

Does this caveat apply to the summary? Or the technical details backing the summary? We probably don't need those full details on the docs site, but if the **consequence** of setting those params is accurate, we should definitely document that clearly.

jkirschner-hashicorp:
> I'll do a more comprehensive review a little later after the content is firmed out a bit, but I have some feedback about the introduction. Would it be helpful to introduce the concept of consistency? Is it a familiar enough concept to the target audience?

> Also, I think we should lead off with a statement about what a consistency mode actually is. The explanation doesn't come until much later, and even then it's a little stilted: " the stale or consistent query parameters should be provided on requests."

@trujillo-adam : I appreciate the fresh perspective on this. I can make a pass at that.

jkirschner-hashicorp:
@trujillo-adam : to set mutual expectations around timeline for this PR, my goal is to get this merged before Consul 1.12 release.

website/content/api-docs/features/consistency.mdx
banks: > Does this caveat apply to the summary?

Caveat applies to the detailed analysis here since it's be at least a year or more since I actually debugged this issue so could have mis-remembered. I'm reasonably sure the outcome is the same: we should warn against setting these and the biggest reason is the load-amplification one. If we write up the exact behaviour in the examples in some longer form answer though, that probably warrants a little more fact checking of how Consul actually behaves. Even if I'm broadly right, there could be other complications I'm unaware of, especially around timeouts and retries.

website/content/api-docs/features/consistency.mdx
```diff
@@ -37,15 +37,101 @@ The three read modes are:
   values. Since this mode allows reads without a leader, a cluster that is
   unavailable will still be able to respond to queries.
 
-To switch these modes, either the `stale` or `consistent` query parameters
-should be provided on requests. It is an error to provide both.
+  ~> **Scaling read requests**: horizontally scaling Consul servers under
+  heavy request load will only help if many of the requests are `stale` reads.
```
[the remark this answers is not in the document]
jkirschner-hashicorp: I think a load balancer visual would make it much easier to understand the latter part of your suggestion.

The part about making more reads `stale` being the most effective approach is important. I'll see if I can work that in succinctly. I also want to add a dedicated section about scalability to the docs (see [in-progress draft](https://docs.google.com/document/d/1BfVezVVnLcXA6IQT0HI7aYIfgFPmmnONtOsaScOz1LM/edit#)).

website/content/api-docs/features/consistency.mdx
jkirschner-hashicorp: Here's the caveat I added to `dns_config.allow_stale`. I added a similar caveat for `discovery_max_stale`. If this looks good, I can also apply a caveat on the agent configuration page entries for these params.

image

website/content/api-docs/features/consistency.mdx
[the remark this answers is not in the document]
jkirschner-hashicorp: @trujillo-adam : thoughts on making `Scaling Read Requests` its own subheading rather than a `~>` notice under its current heading?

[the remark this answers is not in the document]
jkirschner-hashicorp: @banks : I added the following sentence to this notice based on your feedback:
> The most effective way to increase read scalability is to convert non-`stale` reads to `stale` reads.

jkirschner-hashicorp:
@trujillo-adam. @banks : I've made edits based on your feedback. I think this is now ready for a full review.

website/content/api-docs/features/consistency.mdx
```diff
@@ -2,20 +2,47 @@
...
-distributed system.
+Every Consul HTTP API read request uses one of several "consistency modes" that offer
+different levels of response consistency and performance. By default, Consul uses a
+consistency mode that is ideal for most use cases. Users can override consistency modes
+to fine-tune these trade-offs for their own use case at two levels:
+- Per interface (HTTP API, DNS)
+- Per request (HTTP API requests only)
 
-The three read modes are:
+## What is Consistency?
```
jkirschner-hashicorp, line 20: @trujillo-adam : I've added this section to introduce the concept of consistency

website/content/api-docs/features/consistency.mdx
```diff
@@ -37,15 +64,140 @@ The three read modes are:
...
+to override the default on a per-request basis. To use a supported consistency mode
+other than the default, including the desired consistency mode as a query parameter in the query:
+- `stale`: use the `stale` query parameter
+- `consistent`: use the `consistent` query parameter
+- `default`: use the `leader` query parameter;
+   only relevant [if the default consistency mode is changed to `stale`](#changing-the-default-consistency-mode-advanced-usage)
+
+#### Changing the Default Consistency Mode (Advanced Usage)
+
+~> **Not recommended**: we do not recommend using
```
jkirschner-hashicorp, line 167: @banks : can you check the accuracy of this subsection?

website/content/api-docs/features/consistency.mdx
```diff
@@ -37,15 +64,140 @@ The three read modes are:
...
+#### Changing the Default Consistency Mode
+
+If stronger consistency is required at the expense of increased latency,
+set [`dns_config.allow_stale`] to `false` to make Consul use the
+`default` consistency mode instead of `stale`
+when interfacing with the underlying Consul service discovery HTTP APIs.
+
+#### Limiting Staleness (Advanced Usage)
+
+~> **Not recommended**: we do not recommend using [`dns_config.max_stale`]
```
jkirschner-hashicorp, line 132: @banks : can you check the accuracy of this subsection?

website/content/api-docs/features/consistency.mdx
```diff
@@ -37,15 +37,53 @@ The three read modes are:
...
+  [redundancy zones](/docs/enterprise/redundancy).
 
-Note that some endpoints support a `cached` parameter which has some of the same
-semantics as `stale` but different trade offs. This behavior is described in
-[agent caching feature documentation](/api/features/caching).
+The following diagrams show how the request path varies per consistency mode and
+the relative trade-offs between level of consistency and performance.
+
+<Tabs>
+<Tab heading="Stale reads">
```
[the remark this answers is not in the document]
banks: I don't think the inconsistency is strange. I like the idea of presenting `stale` first.

website/content/api-docs/features/consistency.mdx
jkirschner-hashicorp: I agree that the current framing is "here's a feature that sounds great but we tell you not to use". I think that is actually what we want to communicate clearly (that people probably shouldn't use it). I could see someone then asking the reasonable question "well, why was this ever added in the first place?", but I don't think the history matters to much to them.

Maybe the simplest solution is to change the bolded part to:
```suggestion
~> **Use with caution**: we do not recommend using [`dns_config.max_stale`]
```

I considered this initially, but thought it might not be strong enough to discourage usage at-a-glance. What do you think?

website/content/api-docs/features/consistency.mdx
```diff
@@ -37,15 +64,140 @@ The three read modes are:
...
+[`discovery_max_stale`](/docs/agent/options#discovery_max_stale)
+to a value greater than zero in the configuration of the Consul agent receiving requests,
+the `stale` consistency mode will be used by default unless the data is sufficiently stale.
+Setting a positive value for [`discovery_max_stale`](/docs/agent/options#discovery_max_stale)
+is analogous to [limiting the staleness of Consul DNS queries](#limiting-staleness-advanced-usage)
+by setting [`dns_config.max_stale`].
+
+The most likely cause for response staleness to exceed [`dns_config.max_stale`] is that
+Consul servers are overloaded by their request load. Under these circumstances, [`dns_config.max_stale`]
+will exacerbate the problem by converting cheap `stale` reads into more expensive `default` reads.
```
trujillo-adam, line 184:
```suggestion
The most likely cause for response staleness to exceed [`dns_config.max_stale`] is that
Consul servers are overloaded by their request load. Under these circumstances, [`dns_config.max_stale`]
will exacerbate the problem by converting cheap `stale` reads into more expensive `default` reads.
```
If you agree with my previous suggestion, then you can use a similar wording for this information.

trujillo-adam (commented):
A few comments and suggestions

website/content/api-docs/features/consistency.mdx
```diff
@@ -37,15 +64,140 @@ The three read modes are:
...
+
+Only the service discovery HTTP API endpoints
+([catalog](/api-docs/catalog), [health](/api-docs/health), and [query](/api-docs/query))
+allow changing the consistency mode used by default. By setting
+[`discovery_max_stale`](/docs/agent/options#discovery_max_stale)
+to a value greater than zero in the configuration of the Consul agent receiving requests,
+the `stale` consistency mode will be used by default unless the data is sufficiently stale.
+Setting a positive value for [`discovery_max_stale`](/docs/agent/options#discovery_max_stale)
+is analogous to [limiting the staleness of Consul DNS queries](#limiting-staleness-advanced-usage)
+by setting [`dns_config.max_stale`].
```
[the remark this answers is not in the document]
banks: > Only the following service discovery HTTP API endpoints support changing the default consistency mode:

That makes it sound like you can only use the `?stale` query param on these endpoints which is not true.

What this paragraph is trying to say is that, for these particular "service discovery" endpoints, it's possible to configure an agent to use `stale` consistency mode _even if not requested by the client_. Which is useful for spreading read load without changing lots of clients to add `?stale` to their requests.

I prefer the bullet points in your suggestion in general but I think overall this section is not as clear as it could be that we are changing the mode the agent serves _without_ any explicit mode in the query param?

website/content/api-docs/features/consistency.mdx
banks: I'm not sure how to word that @trujillo-adam. It's not so much that setting that causes overload, it's more that in general the most likely times it will have any effect are times where it will tend to do more harm than good. It's one of those things that sounds like a good idea and you can contrive situations where it is desirable behaviour to be able to limit staleness of responses, but in practice it's much more likely to cause problems than actually have the benefit you might assume.

I tried to capture that nuance but I can see it's perhaps a bit wordy/narrative focussed for reference docs.

@jkirschner-hashicorp do you think a reasonable compromise between not spelling out the history and explaining the recommendation would be to use words like: `Warning: setting this has caused undesirable side effects such as causing high server load to turn into a prolonged outage in some cases. We currently don't recommend setting it.`?

trujillo-adam: Ah, OK. I think a `Caution` label is an appropriate label for the callout because while this is something you can do, you should be wary of doing it. `Warning` might sound like failure is imminent. Maybe we shouldn't allow people to do this?

Jared, sure, something like what you have is probably clear. Here's my attempt:

```
`Caution`: Configuring the `dns_config.max_stale` to limit staleness may result 
in a prolonged outage if your server becomes overloaded. We do not recommend 
configuring this setting if the potential for overloading your servers exists.
```
I like "prolonged outage" because it's something concrete, but I guess we can say something more general like the original language you provided " . . . may exacerbate issues associated with high server loads . . ."

website/content/api-docs/features/consistency.mdx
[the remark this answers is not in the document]
trujillo-adam: Ah, ok. Then maybe:

You can configure agents to respond in `stale` consistency mode, even if the client makes requests in a different mode by adding the `?stale` query parameter to one of the following endpoints, depending on the data you want query: 

Is that accurate?

website/content/api-docs/features/consistency.mdx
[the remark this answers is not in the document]
banks: works for me!

website/content/api-docs/features/consistency.mdx
banks: > Maybe we shouldn't allow people to do this?

Honestly I'd prefer to just remove the feature but I think that's a longer discussion as some will be using it and not have had issues etc.

I like your suggested wording @trujillo-adam, although I think the second sentence more or less sums up the problem here - it's practically impossible for an operator to guarantee that their servers won't get overloaded so this warning actually amounts to "never use this" 😄! I prefer your first sentence though. How about if we shift the focus of the second sentence to when you _maybe_ should:

> ```
> `Caution`: Configuring the `dns_config.max_stale` to limit staleness may result in a prolonged outage if your server becomes overloaded. We do not recommend configuring this setting unless your application's correctness would be compromised if results are too stale and you'd rather be unavailable than incorrect.

Or something, I'm sure you can make that last sentence more eloquent but I'm trying to get at the real tradeoff here: is it better to be down or a bit more stale than you would like.

website/content/api-docs/features/consistency.mdx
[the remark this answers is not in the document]
banks: Almost:

> You can configure agents to respond in stale consistency mode, even if the client makes requests with no explicit consistency mode in the query string to one of the following endpoints:

It also probably makes sense to explain _why_ you'd want this somewhere though I don't know if it's better to do it after the sentence above and bullet list. But something like:

> This allows Consul operators to force the majority of service discovery read load to be `stale` and so scale across multiple servers without needing to reconfigure or update every application that makes discovery requests.

jkirschner-hashicorp:
I'll pick this back up within the next 2 weeks.

website/content/api-docs/features/consistency.mdx
```diff
@@ -37,15 +64,140 @@ The three read modes are:
...
+
+When making a request across federated Consul datacenters, requests are forwarded from
+a local server to any remote server. Once in the remote datecenter, the request path
+is the same as a [local request with the same consistency mode](#intra-datacenter-request-behavior).
+
+[![consistency mode behavior across Consul datacenters](/img/consistency-modes/consistency-mode-cross-datacenter.png)](/img/consistency-modes/consistency-mode-cross-datacenter.png)
+
+## Usage
+
+### Consul DNS Queries: Defaults and Overriding
```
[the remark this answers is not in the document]
jkirschner-hashicorp: I'm describing both the default and how to override it, so I don't think it can't be "Overriding Consul DNS Query Defaults". Maybe "Consul DNS Query Mode Default and Overriding"? Or just "Consul DNS Queries" since it's under the "Usage" heading?

website/content/api-docs/features/consistency.mdx
```diff
@@ -37,15 +64,140 @@ The three read modes are:
...
+
+When DNS queries are issued to [Consul's DNS interface](/docs/discovery/dns),
+Consul uses the `stale` consistency mode by default when interfacing with its
+underlying Consul service discovery HTTP APIs
+([catalog](/api-docs/catalog), [health](/api-docs/health), and [query](/api-docs/query)).
+Though this default can be changed, doing so is not recommended.
```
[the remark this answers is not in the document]
jkirschner-hashicorp: I'm changing this to a link that points to the section below on "Limiting Staleness (Advanced Usage)"

website/content/api-docs/features/consistency.mdx
jkirschner-hashicorp: Here's what I'm thinking

---

**Caution**: Configuring [`dns_config.max_stale`] may result in a prolonged outage if your Consul servers become overloaded. We do not recommend using this setting to limit staleness unless it would be preferable for your services to be unavailable than to receive results that are too stale. If bounded result consistency is required by a service, consider modifying the service to use `consistent` service discovery HTTP API queries instead of DNS lookups.

website/content/api-docs/features/consistency.mdx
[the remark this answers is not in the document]
jkirschner-hashicorp: Good point. The only downside is "increased latency" which we already talk about in the next subsection... which is OK. What we don't recommend is using dns_config.max_stale to limit staleness.

I'm thinking the best thing to do is just to remove this sentence.

website/content/api-docs/features/consistency.mdx
```diff
@@ -37,15 +64,140 @@ The three read modes are:
...
+
+The most likely cause for response staleness to exceed [`dns_config.max_stale`] is that
+Consul servers are overloaded by their request load. Under these circumstances, [`dns_config.max_stale`]
+will exacerbate the problem by converting cheap `stale` reads into more expensive `default` reads.
+
+### Consul HTTP API Queries: Defaults and Overriding
+
+All HTTP API read requests use the `default` consistency mode by default
+unless overridden on a per-request basis. Though this default can be changed,
+doing so is not recommended.
```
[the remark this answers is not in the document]
jkirschner-hashicorp: Changing to

---

All HTTP API read requests use the `default` consistency mode by default unless [overridden on a per-request basis](#overriding-a-requests-consistency-mode). We do [not recommend changing the consistency mode used by default](#changing-the-default-consistency-mode-advanced-usage)

website/content/api-docs/features/consistency.mdx
```diff
@@ -2,50 +2,277 @@
...
+
+### Visibility into Response Staleness
 
 To support bounding the acceptable staleness of data, responses provide the
 `X-Consul-LastContact` header containing the time in milliseconds that a server
 was last contacted by the leader node. The `X-Consul-KnownLeader` header also
 indicates if there is a known leader. These can be used by clients to gauge the
 staleness of a result and take appropriate action.
+
+### Visibility into Consistency Mode Used
```
jkirschner-hashicorp, line 237: @banks, @trujillo-adam : this response header wasn't previously documented, so I decided to add it here (along with an example).

jkirschner-hashicorp:
@banks , @trujillo-adam : I believe I've address all of the good feedback you both gave me. I think I'm ready for a final review. Here's the preview link: https://consul-ivhvcy9uc-hashicorp.vercel.app/api-docs/features/consistency

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/668286.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit f4e1ade46ad31afe4333470cda875553c42b1b40 onto `stable-website` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit f4e1ade46ad31afe4333470cda875553c42b1b40 onto `release/1.12.x` succeeded!

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit f4e1ade46ad31afe4333470cda875553c42b1b40 onto `release/1.11.x` succeeded!

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit f4e1ade46ad31afe4333470cda875553c42b1b40 onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/668286)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit f4e1ade46ad31afe4333470cda875553c42b1b40 onto `release/1.9.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/668286)

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit f4e1ade46ad31afe4333470cda875553c42b1b40 onto `release/1.8.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/668286)

PR #12467, opened 2022-02-28
ca: require that tests that use Vault are named correctly

### LINKED ISSUE #12466: ci: make it more difficult to accidentally skip tests that use Vault
Currently the way our CI is setup:
1. The main test suite will skip tests that require vault with `ca.SkipIfVaultNotPresent`
2. The `test-connect-ca-providers` will run some tests that have `_Vault_` in their name (see [GNUmakefile](https://github.com/hashicorp/consul/blob/v1.11.3/GNUmakefile#L328-L342))

This setup makes it very easy to accidentally write a test that doesn't use the name `_Vault_`, which will never run in CI! A test that never runs in CI is a big risk.

There are a few options for fixing this:
1. Install Vault as part of the main `go-test` job, so that these tests can run normally. If we do this, we should change `ca.SkipIfVaultNotPresent` to check if the `CI` env var is set, and if it is, never allow the test to be skipped.
2. change `ca.SkipIfVaultNotPresent` to call `t.Name()`, and fail the test if the name doesn't include `_Vault_`. That way trying to write a test that uses `ca.SkipIfVaultNotPresent` will never pass unless the test is named correctly.

I almost missed this problem in #12298, which is what prompted me to create this issue.

dnephin:
Fixes #12466 with option 2 described in that issue.

Previously we were using two different criteria to decide where to run a
test.  The main `go-test` job would skip Vault tests based on the
presence of the `vault` binary, but the `test-connect-ca-providers` job
would run tests based on the name.

This led to a scenario where a test may never run in CI.

To fix this problem I added a name check to the function we use to skip
the test. This should ensure that any test that requires vault is named
correctly to be run as part of the `test-connect-ca-providers` job.

At the same time I relaxed the regex we use. I verified this runs the
same tests using `go test --list Vault`.  I made this change because a
bunch of tests in `agent/connect/ca` used `Vault` in the name, without
the underscores. Instead of changing a bunch of test names, this seemed
easier.

With this approach, the worst case is that we run a few extra tests in
the `test-connect-ca-providers` job, which doesn't seem like a problem.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -330,15 +330,15 @@ ifeq ("$(CIRCLECI)","true")
 # Run in CI
 	gotestsum --format=short-verbose --junitfile "$(TEST_RESULTS_DIR)/gotestsum-report.xml" -- -cover -coverprofile=coverage.txt ./agent/connect/ca
 # Run leader tests that require Vault
-	gotestsum --format=short-verbose --junitfile "$(TEST_RESULTS_DIR)/gotestsum-report-leader.xml" -- -cover -coverprofile=coverage-leader.txt -run '.*_Vault_' ./agent/consul
+	gotestsum --format=short-verbose --junitfile "$(TEST_RESULTS_DIR)/gotestsum-report-leader.xml" -- -cover -coverprofile=coverage-leader.txt -run Vault ./agent/consul
 # Run agent tests that require Vault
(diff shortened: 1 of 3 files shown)


kisunji (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/599404.

Issue #12241, opened 2022-02-01
Cleanup config_entry CanRead/CanWrite code to give detailed error messages


markan:
Currently the config_entry.go subsystem delegates authorization decisions via the ConfigEntry interface CanRead and CanWrite code. Unfortunately this returns a true/false value and loses the details of the source. 

This is not helpful, especially since it the config subsystem can be more complex to understand, since it covers so many domains.

We should refactor CanRead/CanWrite to return a structured error message (PermissionDenied or the like) with more details about the reason for denial.

Issue #12240, opened 2022-02-01
Go through permission checks and rework to generated detailed information.


markan:
This builds on #12239

We currently don't give very detailed information when acl checks fail. We should fix this by going through our code and wherever possible capturing the full information available at the point of error construction. This intended to be relatively low impact work and should not include cases where the enforcement decision is not near the construction of the error.

Note: For review purposes this work will probably be broken down into smaller PRs on a subsystem basis.

Issue #11337, opened 2021-10-15
Proposal: move (more) of the ACL system to the acl package and subpackages


dnephin:
The [ACL](https://www.consul.io/docs/security/acl/acl-system) sub-system is responsible for authenticating and authorizing request handled by Consul HTTP, RPC, xDS, and gRPC interfaces. Currently some of the ACL system is implemented in the `acl` package, however large parts of it are in:
* `agent/consul/acl.go`
* `agent/consul/acl_client.go`
* `agent/consul/acl_server.go`
* `agent/consul/acl_endpoint.go`
* `agent/consul/leader.go`
* `agent/structs/acl.go`
* `agent/structs/acl_cache.go`
* `agent/consul/state/acl.go`

This issue proposes we move more of the logic into packages under `acl` to make it easier to test, and easier to develop.

The proposed structure is:
* `acl` - remains the package for the rules and policies system
* `acl/aclauthz` - the `ACLResolver` and `ACLCaches`, currently in `agent/consul/acl.go` and `agent/structs/acl_cache.go`. Started in #11233 and #11221.
* `acl/aclrpc` - the RPC server, currently in `agent/consul/acl_endpoint.go`
* TBD - a package for the initialization and goroutine management that is currently in the `leader.go`

PR #12470, opened 2022-03-01
Bulk acl message fixup oss

markan:
This does a bulk conversion for improved Permission Denied errors. 

Part of issue https://github.com/hashicorp/consul/issues/12240

The intent is to incrementally move us towards higher resolution errors. While this is a relatively large PR. This can be read by individual commit, which should hopefully be relatively digestible. I've tried to split commits so semantic changes and mechanical changes are done separately in the hope that it will be easier to review.

There are a few areas we will need to pick up in later issues
* Capturing identity information [#12481](https://github.com/hashicorp/consul/issues/12481)
* Reworking ResolveResult and AllowAuthorizer into one structure, and working the final product all the way through the system. This probably involves moving more stuff into acl to simplify things. This may be one as part of the capture of identity information [#11337](https://github.com/hashicorp/consul/issues/11337)
* reworking the catalog information to return errors [#12241](https://github.com/hashicorp/consul/issues/12241)
* thinking about how to give better hints on filters.
* Cleaning up some of places where we directly create PermissionDenied structs to include detailed information

### DIFF
--- a/.changelog/12470.txt
+++ b/.changelog/12470.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+Provide fuller detail in the error messsage when an ACL denies access.
+```
--- a/.changelog/12470.txt
+++ b/.changelog/12470.txt
@@ -1,3 +1,3 @@
 ```release-note:enhancement
-Provide fuller detail in the error messsage when an ACL denies access.
+acl: Provide fuller detail in the error messsage when an ACL denies access.
 ```
--- a/acl/authorizer.go
+++ b/acl/authorizer.go
@@ -42,6 +42,7 @@ const (
 	ResourceEvent     Resource = "event"
 	ResourceIntention Resource = "intention"
 	ResourceKey       Resource = "key"
+	ResourceKeyPrefix Resource = "key prefix"
 	ResourceKeyring   Resource = "keyring"
 	ResourceNode      Resource = "node"
 	ResourceOperator  Resource = "operator"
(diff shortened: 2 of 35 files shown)


markan:
The enterprise side of this is https://github.com/hashicorp/consul-enterprise/pull/1533

acl/authorizer.go
```diff
@@ -161,6 +162,280 @@ type Authorizer interface {
 
 	// Embedded Interface for Consul Enterprise specific ACL enforcement
 	enterpriseAuthorizer
+
+	// ToAllowAuthorizer is needed until we can use ResolveResult in all the places this interface is used.
+	ToAllowAuthorizer() AllowAuthorizer
+}
```
markan, line 168: Eventually this will be unified with the ACLResolveResult structure, but want to take that on as a separate line item. https://github.com/hashicorp/consul/issues/12481

.changelog/12470.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
+Provide fuller detail in the error messsage when an ACL denies access.
```
rboyer, line 2:
```suggestion
acl: Provide fuller detail in the error messsage when an ACL denies access.
```

acl/authorizer.go
```diff
@@ -42,6 +42,7 @@ const (
 	ResourceEvent     Resource = "event"
 	ResourceIntention Resource = "intention"
 	ResourceKey       Resource = "key"
+	ResourceKeyPrefix Resource = "key prefix"
```
rboyer, line 45: This appears to be dead code.

acl/authorizer.go
```diff
@@ -161,6 +162,280 @@ type Authorizer interface {
...
+	}
+	return nil
+}
+
+// IntentionDefaultAllowAllowed determines the default authorized behavior
+// when no intentions match a Connect request.
+func (a AllowAuthorizer) IntentionDefaultAllowAllowed(ctx *AuthorizerContext) error {
+	if a.Authorizer.IntentionDefaultAllow(ctx) != Allow {
+		// This is a bit nuanced, in that this isn't set by a rule, but inherited globally
+		// TODO revisit when we have full accessor info
```
rboyer, line 232: Can you scope this TODO with an area like `TODO(acl-enhance)` or something topical?

acl/authorizer.go
```diff
@@ -161,6 +162,280 @@ type Authorizer interface {
...
+	}
+	return nil
+}
+
+// KeyWritePrefixAllowed checks for permission to write to an
+// entire key prefix. This means there must be no sub-policies
+// that deny a write.
+func (a AllowAuthorizer) KeyWritePrefixAllowed(name string, ctx *AuthorizerContext) error {
+	if a.Authorizer.KeyWritePrefix(name, ctx) != Allow {
+		// TODO revisit this message; we may need to do some extra plumbing inside of KeyWritePrefix to
```
rboyer, line 284: Please scope this todo.

acl/authorizer.go
```diff
@@ -161,6 +162,280 @@ type Authorizer interface {
...
+	if a.Authorizer.NodeRead(name, ctx) != Allow {
+		return PermissionDeniedByACL(a, ctx, ResourceNode, AccessRead, name)
+	}
+	return nil
+}
+
+// NodeReadAllAllowed checks for permission to read (discover) all nodes.
+func (a AllowAuthorizer) NodeReadAllAllowed(ctx *AuthorizerContext) error {
+	if a.Authorizer.NodeReadAll(ctx) != Allow {
+		// TODO Revisit this to see if we can return a more detailed answer
```
rboyer, line 337: Please scope this todo.

acl/authorizer.go
```diff
@@ -161,6 +162,280 @@ type Authorizer interface {
...
+	if a.Authorizer.ServiceRead(name, ctx) != Allow {
+		return PermissionDeniedByACL(a, ctx, ResourceService, AccessRead, name)
+	}
+	return nil
+}
+
+// ServiceReadAllAllowed checks for permission to read all services
+func (a AllowAuthorizer) ServiceReadAllAllowed(ctx *AuthorizerContext) error {
+	if a.Authorizer.ServiceReadAll(ctx) != Allow {
+		return PermissionDeniedByACLUnnamed(a, ctx, ResourceService, AccessRead) // read
```
rboyer, line 399: Did you want this error message to maybe be slightly different since this is a read-all inquiry?

acl/authorizer.go
```diff
@@ -161,6 +162,280 @@ type Authorizer interface {
...
+	if a.Authorizer.SessionWrite(name, ctx) != Allow {
+		return PermissionDeniedByACL(a, ctx, ResourceSession, AccessWrite, name)
+	}
+	return nil
+}
+
+// SnapshotAllowed checks for permission to take and restore snapshots.
+func (a AllowAuthorizer) SnapshotAllowed(ctx *AuthorizerContext) error {
+	if a.Authorizer.Snapshot(ctx) != Allow {
+		// Implimetation of this currently just checks acl write
```
rboyer, line 435:
```suggestion
		// Implementation of this currently just checks acl write
```

acl/authorizer.go
```diff
@@ -161,6 +162,280 @@ type Authorizer interface {
 
 	// Embedded Interface for Consul Enterprise specific ACL enforcement
 	enterpriseAuthorizer
+
+	// ToAllowAuthorizer is needed until we can use ResolveResult in all the places this interface is used.
+	ToAllowAuthorizer() AllowAuthorizer
```
rboyer, line 167: Does this need to be a method? Do any of the implementation of this interface do more than `&AllowAuthorizer{a}`?

If they are all the same there may be a benefit towards just having `func acl.NewAllowAuthorizer(Authorizer)` instead

acl/errors.go
```diff
@@ -126,3 +128,24 @@ func PermissionDeniedByACLUnnamed(_ Authorizer, context *AuthorizerContext, reso
...
+	t.Helper()
+	if err == nil {
+		t.Fatal("An error is expected but got nil.")
+	}
+	if v, ok := err.(PermissionDeniedError); ok {
+		require.Equal(t, v.Resource, resource)
+		require.Equal(t, v.AccessLevel, accessLevel)
+		require.Equal(t, v.ResourceID.Name, resourceID)
+	} else {
+		t.Fatalf("Expected a permission denied error got %v", err)
```
rboyer, line 142: Sometimes if there's a type assertion involved, I like to get `%T` into the error messages too, so you know what the type of the value was as well.

acl/errors.go
```diff
@@ -126,3 +128,24 @@ func PermissionDeniedByACLUnnamed(_ Authorizer, context *AuthorizerContext, reso
...
+	}
+}
+
+func RequirePermissionDeniedMessage(t testing.TB, msg string, _ Authorizer, _ *AuthorizerContext, resource Resource, accessLevel AccessLevel, resourceID string) {
+	require.Contains(t, msg, errPermissionDenied)
+	require.Contains(t, msg, resource)
+	require.Contains(t, msg, accessLevel.String())
+	require.Contains(t, msg, resourceID)
```
rboyer, line 150: Should all of these parts appear in the same order? If so perhaps it would be stricter to assemble a quick regex with word boundaries between the 4 components instead.

acl/errors.go
```diff
@@ -126,3 +128,24 @@ func PermissionDeniedByACLUnnamed(_ Authorizer, context *AuthorizerContext, reso
...
+	if v, ok := err.(PermissionDeniedError); ok {
+		require.Equal(t, v.Resource, resource)
+		require.Equal(t, v.AccessLevel, accessLevel)
+		require.Equal(t, v.ResourceID.Name, resourceID)
+	} else {
+		t.Fatalf("Expected a permission denied error got %v", err)
+	}
+}
+
+func RequirePermissionDeniedMessage(t testing.TB, msg string, _ Authorizer, _ *AuthorizerContext, resource Resource, accessLevel AccessLevel, resourceID string) {
```
rboyer, line 146: Perhaps also putting a `require.NotEmpty(t, msg, "expected non-empty error message")` to make a class of errors more obvious

acl/authorizer.go
```diff
@@ -161,6 +162,280 @@ type Authorizer interface {
...
+
+// AllowAuthorizer is a wrapper to expose the *Allowed methods.
+// This and the ToAllowAuthorizer function exist to tide us over until the ResolveResult struct
+// is moved into acl.
+type AllowAuthorizer struct {
+	Authorizer
+}
+
+// ACLReadAllowed checks for permission to list all the ACLs
+func (a AllowAuthorizer) ACLReadAllowed(ctx *AuthorizerContext) error {
```
rboyer, line 178: Followup: do these even need to be methods, or could they be a suite of package functions with a leading argument `Authorizer` so your in-endpoint checks would be:
```
if acl.ACLReadAllowed(authz, ctx) {
  ...
}
```
which is terser than what shows up down below.

agent/consul/connect_ca_endpoint.go
```diff
@@ -65,8 +65,8 @@ func (s *ConnectCA) ConfigurationGet(
 	if err != nil {
 		return err
 	}
-	if authz.OperatorWrite(nil) != acl.Allow {
-		return acl.ErrPermissionDenied
+	if err := authz.ToAllowAuthorizer().ToAllowAuthorizer().OperatorWriteAllowed(nil); err != nil {
```
rboyer, line 68:
```suggestion
	if err := authz.ToAllowAuthorizer().OperatorWriteAllowed(nil); err != nil {
```

agent/consul/health_endpoint.go
```diff
@@ -220,6 +220,8 @@ func (h *Health) ServiceNodes(args *structs.ServiceSpecificRequest, reply *struc
 	// If we're doing a connect or ingress query, we need read access to the service
 	// we're trying to find proxies for, so check that.
 	if args.Connect || args.Ingress {
+		// TODO: ACL Error improvements; can this be improved? What happens if we returned an error here?
+		// Is this similar to filters where we might want to return a hint?
 		if authz.ServiceRead(args.ServiceName, &authzContext) != acl.Allow {
```
rboyer, line 225: The `catalog` and `health` endpoints (and parts of `agent`) expect the filter style of ACL enforcement (200s instead of 403s).

agent/consul/internal_endpoint.go
```diff
@@ -464,16 +464,16 @@ func (m *Internal) KeyringOperation(
 	}
 	switch args.Operation {
 	case structs.KeyringList:
-		if authz.KeyringRead(nil) != acl.Allow {
-			return fmt.Errorf("Reading keyring denied by ACLs")
```
rboyer, line 468: For the record this error message is now changing.

agent/consul/internal_endpoint.go
```diff
@@ -464,16 +464,16 @@ func (m *Internal) KeyringOperation(
...
+		if err := authz.ToAllowAuthorizer().KeyringReadAllowed(nil); err != nil {
+			return err
 		}
 	case structs.KeyringInstall:
 		fallthrough
 	case structs.KeyringUse:
 		fallthrough
 	case structs.KeyringRemove:
-		if authz.KeyringWrite(nil) != acl.Allow {
-			return fmt.Errorf("Modifying keyring denied due to ACLs")
```
rboyer, line 476: For the record this error message is now changing.

agent/consul/txn_endpoint_test.go
```diff
@@ -554,57 +554,64 @@ func TestTxn_Apply_ACLDeny(t *testing.T) {
 	}
 
 	// Verify the transaction's return value.
-	var expected structs.TxnResponse
+	var outPos int
```
rboyer, line 557: Can you explain the intent here please?

agent/xds/delta_test.go
```diff
@@ -1156,7 +1156,7 @@ func TestServer_DeltaAggregatedResources_v3_ACLEnforcement(t *testing.T) {
 			case err := <-errCh:
 				if tt.wantDenied {
 					require.Error(t, err)
-					require.Contains(t, err.Error(), "permission denied")
+					require.Contains(t, err.Error(), "Permission denied")
```
rboyer, line 1159: Does this casing change here mean that older consul clients and servers that do `string.Contains` checks will fail to handle denies when you are straddling versions?

rboyer (changes requested):
I left a few comments.

acl/errors.go
```diff
@@ -3,7 +3,9 @@ package acl
 import (
 	"errors"
 	"fmt"
+	"github.com/stretchr/testify/require"
```
kisunji, line 6: nit: should be in its own grouping away from stdlib packages

acl/errors.go
```diff
@@ -126,3 +128,24 @@ func PermissionDeniedByACLUnnamed(_ Authorizer, context *AuthorizerContext, reso
 	desc := NewResourceDescriptor("", context)
 	return PermissionDeniedError{Accessor: "", Resource: resource, AccessLevel: accessLevel, ResourceID: desc}
 }
+
+func RequirePermissionDeniedError(t testing.TB, err error, _ Authorizer, _ *AuthorizerContext, resource Resource, accessLevel AccessLevel, resourceID string) {
```
kisunji, line 132: I don't think our production code should be importing `testing` and `testify` here. Could these be moved to test-specific files?

agent/acl.go
```diff
@@ -2,13 +2,11 @@ package agent
...
 import (
 	"fmt"
-
-	"github.com/hashicorp/serf/serf"
-
 	"github.com/hashicorp/consul/acl"
 	"github.com/hashicorp/consul/agent/structs"
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/consul/types"
+	"github.com/hashicorp/serf/serf"
```
kisunji, line 9: nit: We conventionally group packages with `go imports -local "github.com/hashicorp/consul"`.

acl/authorizer.go
kisunji: I'm curious about this too; it feels like added indirection and I think RB's suggested functions read better

acl/authorizer.go
```diff
@@ -161,6 +162,280 @@ type Authorizer interface {
...
 	enterpriseAuthorizer
+
+	// ToAllowAuthorizer is needed until we can use ResolveResult in all the places this interface is used.
+	ToAllowAuthorizer() AllowAuthorizer
+}
+
+// AllowAuthorizer is a wrapper to expose the *Allowed methods.
+// This and the ToAllowAuthorizer function exist to tide us over until the ResolveResult struct
+// is moved into acl.
+type AllowAuthorizer struct {
```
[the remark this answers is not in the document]
markan: Yes. That is by far the most common pattern we see.

acl/authorizer.go
markan: This is a transitional method; in upcoming PRs more functionality will be loaded onto it (in particular capturing authorization source information), but eventually I'd like to move a bunch of stuff into the ACL directory and hopefully remove this entirely.

acl/errors.go
markan: Good idea.

acl/authorizer.go
markan: The intent is to merge AllowAuthorizer and ACLResolveResult from consul/acl.go in a future PR. These would be methods on the joined class

agent/consul/txn_endpoint_test.go
markan: The broad purpose here is to perform a slightly stricter test on what we return; in particular make sure we have the right resource and permission information. 

The original code attempted to build an identical result to the expected return value from the API call and directly compare them, but that seemed impractical given the new structure. So I altered this into an element by element comparison inline..

agent/xds/delta_test.go
markan: Hmm. That's not something I'd considered. Wondering if I should back the change out for the xds code for now.

markan: Do we have any known cases where the message string is tested for "permission denied" rather than using the status code codes.PermissionDenied? I spent some time code spelunking but didn't really find much either

rboyer: Sorry in my initial skim I neglected to really notice this is not the _consul_ perm denied used all over acls, this is likely fine then.

rboyer (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/605613.

PR #12488, opened 2022-03-02
ui: Remove Settings Ember Route

johncowen:
This PR replaces our super old settings code with a Route-less approach and a fun use of our renderless `Disclosure` component to 'disclose' data loading or data saving - this was one of the usecases I had in mind whilst writing that component. In Consul UI all our data loading/saving is completely declarative. You drop a HTML (_ehem Glimmer_) component in the page and it loads or saves for you. It's that straightforwards.

This lets you use these declarative components slightly more imperatively/javascript-y as `Disclosure` also has a slightly more imperative/javascript-y usage via its exposed methods.

Our data loading components also work like this when just using CSS (you can use `display: none` to stop them loading), so just a little self-reminder that our data saving should also be implemented in the same way.

There was one thing that we still wanted to be in a Route for the moment, and that is aborting any in flight blocking queries if you turn blocking queries off. This kind of made sense to move to the top-level of the application (we may end up putting an additional toggle in the user dropdown or something 🤷 ), so I moved it up there for the moment at least.

🎼 Bm, bm, bm, another `Route` bites the dust. 🎶 

Lastly, our settings page was always located at `/setting` so I moved this to `/settings` and added a redirect from `/setting` plus a note to remove this redirect once we get far enough away from this change. This change targets 1.12, our next major.

Longer term we actually want to make this an operator level/Consul config value so an operator can configure this globally for all running UIs for all users, but a small note here, even for when we do get round to doing that we will probably keep this user level setting:

> Auto-updating
For any auto-updating information that (1) starts automatically and (2) is presented in parallel with other content, there is a mechanism for the user to pause, stop, or hide it or to control the frequency of the update unless the auto-updating is part of an activity where it is essential.

https://www.w3.org/WAI/WCAG21/Understanding/pause-stop-hide.html

Our current user level "turn live-updates on/off" setting means we pass this accessibility requirement.


evrowe (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/604712.

PR #12490, opened 2022-03-02
add a rate limiter to config auto-reload

dhiaayachi:
This add a coalesce timer to auto config reload to reduce the risk that we reload incomplete configuration or DDOS Consul by modifying the config file.
The coalesce timer is implemented as a Watcher interface that wrap the original file watcher.


agent/agent.go
```diff
@@ -366,6 +366,8 @@ type Agent struct {
 
 	// enterpriseAgent embeds fields that we only access in consul-enterprise builds
 	enterpriseAgent
+
+	coalesceTimerShim func(inputCh chan *config.FileWatcherEvent, coalesceDuration time.Duration) chan []config.FileWatcherEvent
```
FFMMM, line 370: I like this! Would this be better placed in `FileWatcher.go` ?

agent/agent.go
```diff
@@ -702,24 +705,60 @@ func (a *Agent) Start(ctx context.Context) error {
...
-				}
-			}()
+				}()
+			}
 		}
 	}
 	return nil
 }
 
+func coalesceTimer(inputCh chan *config.FileWatcherEvent, coalesceDuration time.Duration) chan []config.FileWatcherEvent {
```
boxofrad, line 727: I mentioned in [a comment](https://github.com/hashicorp/consul/pull/12329#discussion_r831028102) on your other PR that I'd consider making the `FileWatcher` an interface for improved testability. Another benefit of this approach would be that we could contain this coalescing logic in its own type wrapping the underlying file watcher.

```Go
package config

type FileWatcher interface {
	Events() <-chan []FileWatcherEvent
}

func NewRateLimitedFileWatcher(w FileWatcher, coalesceInterval time.Duration) FileWatcher {
	rlw := &rateLimitedFileWatcher{
		watcher:  w,
		interval: coalesceInterval,
		eventsCh: make(chan []FileWatcherEvent),
	}
	go rlw.run() // Note: this goroutine will exit when the w.Events() channel is closed.
	return rlw
}

func (rlw *rateLimitedFileWatcher) Events() <-chan []FileWatcherEvent { return rlw.eventsCh }

func (rlw *rateLimitedFileWatcher) run() {
	for {
		// Read from rlw.w.Events() and dispatch on eventsCh.
	}
}
```

This would mean that in tests we could provide a non-wrapped or even mock watcher, rather than the test shim approach we're using at the moment.

What do you think?

rboyer: I like this.

rboyer (changes requested):
Seconding the readability suggestion from @boxofrad

agent/config/builder.go
```diff
@@ -1004,64 +1004,65 @@ func (b *builder) build() (rt RuntimeConfig, err error) {
...
+		StartJoinAddrsWAN:                b.expandAllOptionalAddrs("start_join_wan", c.StartJoinAddrsWAN),
+		TaggedAddresses:                  c.TaggedAddresses,
+		TranslateWANAddrs:                boolVal(c.TranslateWANAddrs),
+		TxnMaxReqLen:                     uint64Val(c.Limits.TxnMaxReqLen),
+		UIConfig:                         b.uiConfigVal(c.UIConfig),
+		UnixSocketGroup:                  stringVal(c.UnixSocket.Group),
+		UnixSocketMode:                   stringVal(c.UnixSocket.Mode),
+		UnixSocketUser:                   stringVal(c.UnixSocket.User),
+		Watches:                          c.Watches,
+		AutoReloadConfigCoalesceInterval: 1 * time.Second,
```
dhiaayachi, line 1065: This is not an exposed config yet but I thought we can expose it. in the future if we have requests for it.

agent/config/file_watcher_test.go
```diff
@@ -64,6 +64,23 @@ func TestWatcherAddRemove(t *testing.T) {
 
 }
 
+func TestWatcherReplace(t *testing.T) {
```
dhiaayachi, line 67: This is not related to rate limiting but it was missing a test

agent/agent_test.go
```diff
@@ -5720,3 +5721,111 @@ func TestAgent_AutoReloadDoReload_WhenKeyThenCertUpdated(t *testing.T) {
 		require.NotEqual(r, cert2.PrivateKey, srv.tlsConfigurator.Cert().PrivateKey)
 	})
 }
+
+func Test_coalesceTimerTwoPeriods(t *testing.T) {
```
dhiaayachi, line 5725: This test to make sure that the rate limiting logic work correctly with the agent

agent/config/file_watcher.go
```diff
@@ -44,7 +44,7 @@ type watchedFile struct {
 }
 
 type FileWatcherEvent struct {
-	Filename string
+	Filenames []string
```
dhiaayachi, line 47: I had to change this to make sure both implementations respect the interface

agent/config/ratelimited_file_watcher_test.go
```diff
@@ -0,0 +1,91 @@
...
+	}()
+
+	_, err = file.WriteString("test config 2")
+	require.NoError(t, err)
+	err = file.Sync()
+	require.NoError(t, err)
+	require.NoError(t, assertEvent(file.Name(), w.EventsCh(), defaultTimeout))
+}
+
+func TestEventRateLimitedWatcherMove(t *testing.T) {
```
dhiaayachi, line 72: This test the rate limiting logic. Send 10 events and make sure we get one

agent/agent.go
```diff
@@ -462,6 +462,13 @@ func New(bd BaseDeps) (*Agent, error) {
 			a.baseDeps.WatchedFiles = append(a.baseDeps.WatchedFiles, f.Cfg.CertFile)
 		}
 	}
+	if a.baseDeps.RuntimeConfig.AutoReloadConfig && len(a.baseDeps.WatchedFiles) > 0 {
+		w, err := config.NewRateLimitedFileWatcher(a.baseDeps.WatchedFiles, a.baseDeps.Logger, a.baseDeps.RuntimeConfig.AutoReloadConfigCoalesceInterval)
+		if err != nil {
+			a.baseDeps.Logger.Error("error loading config", "error", err)
```
rboyer, line 468: Shouldn't this actually `return nil, err` like other errors in this method?

agent/config/ratelimited_file_watcher.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/hashicorp/go-hclog"
+)
+
+type rateLimitedFileWatcher struct {
+	watcher          Watcher
+	eventCh          chan *FileWatcherEvent
+	coalesceInterval time.Duration
+}
+
+func (r rateLimitedFileWatcher) Start(ctx context.Context) {
```
rboyer, line 16: I think all of these receivers should be pointers. In almost every circumstance that's what you want, the non-pointer receiver should be an exception.

```suggestion
func (r *rateLimitedFileWatcher) Start(ctx context.Context) {
```

agent/config/ratelimited_file_watcher.go
```diff
@@ -0,0 +1,85 @@
...
+		watcher:          watcher,
+		coalesceInterval: coalesceInterval,
+		eventCh:          make(chan *FileWatcherEvent),
+	}, nil
+}
+
+func (r rateLimitedFileWatcher) coalesceTimer(inputCh chan *FileWatcherEvent, coalesceDuration time.Duration) {
+	var coalesceTimer *time.Timer = nil
+	sendCh := make(chan struct{})
+	FileWatcherEvents := make([]string, 0)
```
rboyer, line 57:
```suggestion
	var fileWatcherEvents []string
```
Combo suggestion:

1. using a capitalized variable name in a method body is strange because normally that is how you indicate exported fields
2. Usually if you don't have a good sense for how big a slice should be, we usually start them out as the `nil` slice and let them grow via `append`, rather than explicitly allocating a zero-length slice which immediately will have to be reallocated on first append anyway.

rboyer, line 57: (personal preference)

If i have a few variables being defined at once I like the look of a `var` block:
```
var (
	coalesceTimer     *time.Timer
	sendCh            = make(chan struct{})
	fileWatcherEvents []string
)
```

agent/config/ratelimited_file_watcher.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	return rateLimitedFileWatcher{
+		watcher:          watcher,
+		coalesceInterval: coalesceInterval,
+		eventCh:          make(chan *FileWatcherEvent),
+	}, nil
+}
+
+func (r rateLimitedFileWatcher) coalesceTimer(inputCh chan *FileWatcherEvent, coalesceDuration time.Duration) {
+	var coalesceTimer *time.Timer = nil
```
rboyer, line 55: I don't see where you are using `coalesceTimer.C`

rboyer: OH never mind i just hadn't used `time.AfterFunc` in like a million years.

agent/config/ratelimited_file_watcher.go
```diff
@@ -0,0 +1,85 @@
...
+	sendCh := make(chan struct{})
+	FileWatcherEvents := make([]string, 0)
+	go func() {
+		for {
+			select {
+			case event, ok := <-inputCh:
```
rboyer, line 61: Do you need an explicit `ctx.Done()` select case here, or should the closing of `inputCh` be enough to shut things down?

agent/config/ratelimited_file_watcher.go
dhiaayachi: Nice catch, it was a typo

agent/config/ratelimited_file_watcher.go
dhiaayachi: The channel close should be enough. we for range on the channel and if it's closed we leave the loop.

kisunji (approved):
LGTM! Small nits

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/622677.

PR #12493, opened 2022-03-02
ui: Overview UI Routing

johncowen:
This PR adds routes and an initial landing page for the Cluster Overview page:

Screenshot 2022-03-02 at 15 32 06

UI 'index page' now looks like that ^

You'll see the odd commented out bit in here, which is awaiting decisions on backend support etc. Other than that I've spotted some other opportunities for modernizing a little and making some of the code here consistent, so I wanted to do that in a separate PR, I'll probably do these on top of here.

I also spotted a very weird ember-can error right at the very end or the work, and whilst the code related to `ember-can` in the changeset here gets things working without a big read unfathomable console error - it doesn't read well: `is license` should be `can license` and use ember-can directly. I've briefly looked into this and converting the Helpers here to native classes helped, but I still get the error when using `can`. I quickly tried an upgrade but that unfortunately breaks things differently - all in all I wanted to keep this PR as clean as possibl.  I'll continue to look at this ontop in a separate PR as its already growing in size due to `this ember-can` problem.

But in saying all of that this PR would be fine to go in as is if need be, at least a potential review/approval shouldn't wait on those further PRs.

### DIFF
--- a/ui/packages/consul-ui/app/abilities/hcp.js
+++ b/ui/packages/consul-ui/app/abilities/hcp.js
@@ -1,0 +1,11 @@
+import BaseAbility from './base';
+import { inject as service } from '@ember/service';
+
+export default class HcpAbility extends BaseAbility {
+  @service('env') env;
+
+  get is() {
+    return false;
+    // return this.env.var('CONSUL_HCP');
+  }
+}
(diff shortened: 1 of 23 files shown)


johncowen:
FYI: This one is waiting on https://github.com/hashicorp/consul/pull/12505 before merging

PR #12506, opened 2022-03-03
ui: Add license endpoint/datasource

johncowen:
[Rendered Docs](https://consul-ui-staging-i4g3fq5xq-hashicorp.vercel.app/ui/docs/repositories/license)

Just to be clear that ^ is just to show the `uri` to use and what the response is, and I suppose show that it refreshes automatically every... (whooops :typing-cat: _commits_ ) 30 seconds 😅 . Longer term probably would be good to make that more visual somehow also 🤔 

This PR does several things:

- Adds data via Consul's HTTP backend for the `operator/license` endpoint
- Make a first step on the long road towards removing `ember-data` entirely from our application 🎉 
- Makes a second step at creating a pattern to document all our data and how to access it.

## Removing Ember Data

For various reasons we've decided to make a move away from ember-data. It's never been something I've been that on board with seeing as our source of truth should be the Consul backend itself, not a secondary cached source of truth on the frontend. I think even in ember-data docs they warned against using it for an application that was mostly read-only and to me that makes perfect sense.

`ember-data` also encourages you to do lots of data reshaping for the sole reason of making it easier for Ember's `Route.model` based flow to work. This flow is generally a 'once' based flow that is completely finished by the time you render your template. This means that `ember-data` has all sorts of Ember Proxies and Ember PromiseProxies etc (now kind of partly deprecated) to continue to keep the data live once the `model` flow has completed via the use of mutable data structures. Additionally, all of `ember-data`s relationships are modelled way on up in the Data Layer, and whilst this initially might seem to make sense, it does make it difficult to achieve patterns like conditional relationships, nested relationships etc etc.

`ember-data` encourages you to split you code into adapters and serializers. Whilst at first glance this might feel like a reasonable place for separation, but this separation is largely not required. How you call the HTTP APIs and how they are shaped are generally things that are very highly coupled, and probably should be, especially in a piece of software like Consul. It's unlikely that they need to be separate, and in fact separating them introduces additional overhead and makes it harder to figure out what is happening under the hood.

Perhaps lastly we recently found an issue with the addon itself with the way that ember-data encourages you to do extra data reshaping which could cause issues on larger Consul installs.

All of the above has almost completely been replaced by our `DataSource` components and related code, which lets you model data relationships in Glimmer/Handlebars components, which means that modelling conditional or nested relationships is a case of putting a component within another, possibly with a conditional wrapped around it.

`DataSources` are based on an EventTarget/Event Listener model, i.e. not 'onces' and therefore continue to listen for data well after the `model` hook has ended.

We don't want to do this removal all at once, and as this endpoint is a straightforwards endpoint it was a good opportunity to make a start.

Notes:

1. We'd always planned doing this removal eventually, and most of the HTTP interaction was built in such a way that whilst I didn't know at the time exactly how it would be pieced together, it was built in pieces so it could be pieced together differently. So the related code here is simply threading that different together slightly differently - importantly whilst continuing to support the old way of requesting data via `ember-data`. This probably isn't the final shape of things once you get past the repository, but it is closer to how the repository code.
2. Just to reiterate, we _have_ to support both approaches at once, as a friend/colleague often puts it "change the wheels on the car whilst its moving along". This also means that if we missed something that we need, its pretty easy to drop back to our old approach whilst we re-assess, hopefully that won't be necessary but the choice to do that is still there if we need to.
3. We aren't currently removing the ember-data store or the usage of it. Partly because we probably will always need some sort of centralized cache/store (possibly not ember-data's), and partly because again we don't want to do too much all at once.
4. I doubt we will change existing endpoints for a little while yet, but for new endpoints, where there is the opportunity to, I'll use this new approach.

## Repository Docs

I kind of wanted a way to isolate our data interaction and make them visible, both for building them out and for documentation purposes. It also helps for making PRs that include a'component' but don't yet use it and providing some way for reviews to play with the new component easily - luckily all our data is built as components.

There's probably more to come here, but I didn't want to go too much further as I mainly wanted an easy way for reviews to see what this PR was adding, but I'm hoping this will be super useful for other engineers in the future, it was really helpful to build out the data for this having this page to use.


evrowe (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/604838.

PR #12507, opened 2022-03-03
Add a smaller interface for protobuf-generated message types to implement

kisunji:
Since we are phasing out gogo-protobuf moving forwards, we lose the convenience of generated files already implementing `RPCInfo`. This interface will reduce the number of methods that generated types will manually have to implement.

### DIFF
--- a/agent/structs/structs.go
+++ b/agent/structs/structs.go
@@ -202,6 +202,13 @@ func ValidStatus(s string) bool {
 	return s == api.HealthPassing || s == api.HealthWarning || s == api.HealthCritical
 }
 
+// RPCRequest is a shim that allows structs to implement a single method
+// for a given RPC request rather than all of RPCInfo at once.
+// This is relevant as we phase away from gogo-protobuf's embedding features.
+type RPCRequest interface {
+	RPCInfo() RPCInfo
+}
+
 // RPCInfo is used to describe common information about query
 type RPCInfo interface {
 	RequestDatacenter() string
--- a/agent/structs/structs.go
+++ b/agent/structs/structs.go
@@ -207,11 +207,11 @@ func ValidStatus(s string) bool {
 // This is relevant as we phase away from gogo-protobuf's embedding features.
 type RPCRequest interface {
 	RPCInfo() RPCInfo
+	RequestDatacenter() string
 }
 
 // RPCInfo is used to describe common information about query
 type RPCInfo interface {
-	RequestDatacenter() string
 	IsRead() bool
 	AllowStaleRead() bool
 	TokenSecret() string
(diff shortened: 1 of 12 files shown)


kisunji:
Trying to think of a way to avoid the plumbing work

agent/structs/structs.go
```diff
@@ -202,9 +202,18 @@ func ValidStatus(s string) bool {
...
 
+// RPCRequest is a shim that allows structs to implement a smaller interface
+// without embedding types like QueryOptions, WriteRequest, ReadRequest, etc.
+// This is relevant for protobuf types as we phase away from gogo-protobuf's
+// embedding features. Protobuf-generated types should have fields that implement
+// RPCInfo instead of implementing it themselves.
+type RPCRequest interface {
+	RPCInfo() RPCInfo
+	RequestDatacenter() string
+}
```
FFMMM, line 213: I may be lacking context here but I don't readily understand how this helps minimize the interface? Won't `RPCRequest`s need to implement `RPCInfo`s? Which still require the same funcs to be implemented.

FFMMM:
Thanks for putting this out Chris !! 

I tried re-running the tests but they still fail 🤔 ; I think it may be related to the changes but not sure if you wanted to rebase just to be sure!

agent/structs/structs.go
kisunji: Basically when we make new protobuf types, we need to implement `RPCInfo`:
```
type RPCInfo interface {
	RequestDatacenter() string
	IsRead() bool
	AllowStaleRead() bool
	TokenSecret() string
	SetTokenSecret(string)
	HasTimedOut(since time.Time, rpcHoldTimeout, maxQueryTime, defaultQueryTime time.Duration) bool
}
```

This was easily done by defining a type (e.g. [WriteRequest](https://github.com/hashicorp/consul/blob/main/proto/pbcommon/common.proto#L33)) that already [implemented it](https://github.com/hashicorp/consul/blob/main/proto/pbcommon/common.go#L253-L280) then _embedding_ that struct with gogo-protobuf ([example from enterprise](https://github.com/hashicorp/consul-enterprise/blob/main/proto/pbpartition/partition.proto#L65)).

Since we want to move away from gogo-protobuf features, we can no longer embed structs like `WriteRequest` and therefore every gRPC message type needs to implement the 6 methods above [by hand](https://github.com/hashicorp/consul-enterprise/blob/long-lived/peering/proto/pbpeering/peering.go).

What this interface allows us to do is have `WriteRequest` (which already implements `RPCInfo`) as a field (not embedded) and _return that field_ as `RPCInfo()`.

So instead of implementing 6 methods per protobuf message, we can now do 2 per message (Datacenter is treated separately for more complicated reasons).

Does that answer your question?

kisunji:
> Thanks for putting this out Chris !!
> 
> I tried re-running the tests but they still fail 🤔 ; I think it may be related to the changes but not sure if you wanted to rebase just to be sure!

Yeah I'm convinced that this PR broke something (even though the tests say flaky there's no way it could flake 100% of the times I've tried it.. right??)

PR #12511, opened 2022-03-03
server: ensure that service-defaults meta is incorporated into the discovery chain response

rboyer:
Also add a new "Default" field to the discovery chain response to clients (cc @johncowen )

TODO:

- [x] update this https://www.consul.io/docs/connect/l7-traffic/discovery-chain

### DIFF
--- a/website/content/docs/connect/l7-traffic/discovery-chain.mdx
+++ b/website/content/docs/connect/l7-traffic/discovery-chain.mdx
@@ -121,8 +121,14 @@ resolved by name using the [`Targets`](#targets) field.
   balancer data plane objects to avoid sharing customized and non-customized
   versions.
 
+- `Default` `(bool: <optional>)` -  Indicates if this discovery chain is based on no
+   `service-resolver`, `service-splitter`, or `service-router` config entries.
+
 - `Protocol` `(string)` - The overall protocol shared by everything in the
   chain.
+
+- `ServiceMeta` `(map<string|string>)` - The metadata from the underlying `service-defaults` config
+   entry for the service named `ServiceName`.
 
 - `StartNode` `(string)` - The first key into the `Nodes` map that should be
   followed when traversing the discovery chain.


agent/consul/discovery_chain_endpoint.go
```diff
@@ -93,7 +93,7 @@ func (c *DiscoveryChain) Get(args *structs.DiscoveryChainRequest, reply *structs
 			reply.Index = index
 			reply.Chain = chain
 
-			if chain.IsDefault() {
```
rboyer, line 96: :thinking: Actually I think I have to pull this change out and backport it to 1.11 and 1.10

erichaberkorn (approved):
The service meta portion of this looks good to me 👍 I reviewed the other part too, but don't have enough context to do a proper review.

api/discovery_chain.go
```diff
@@ -109,9 +109,17 @@ type CompiledDiscoveryChain struct {
 	// non-customized versions.
 	CustomizationHash string
 
+	// Default indicates if this discovery chain is based on no
```
mkeeler, line 112: Should the API docs be updated to account for these changes.

mkeeler (commented):
Code looks great. Just the two minor things.

agent/consul/discoverychain/compile.go
```diff
@@ -327,14 +333,51 @@ func (c *compiler) compile() (*structs.CompiledDiscoveryChain, error) {
...
+//	resolver action being applied is redirection to another resolver that is
+//	default, so we double check the resolver matches the requested resolver.
+//
+// NOTE: "default chain" mostly means that this is compatible with how things
+// worked (roughly) in consul 1.5 pre-discovery chain, not that there are zero
+// config entries in play (like service-defaults).
+func (c *compiler) determineIfDefaultChain() bool {
+	if false {
+		return c.entries.IsChainEmpty()
+	}
```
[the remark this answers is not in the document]
rboyer: Reverted refactor mostly that didn't go anywhere that worked.

api/discovery_chain.go
rboyer: IIRC we don't currently document the full disco chain response at all.

mkeeler (approved):

PR #12514, opened 2022-03-03
ui: Add auto-pilot/state endpoint usage (merged into DC models)

johncowen:
The `auto-pilot/state` endpoint gives us a tonne of extra information into 'Voting Server Health' such as gossip and raft health and information (as well as other things). Whilst these things come from the auto-pilot endpoint, these things are 'per datacenter', and therefore they align well with our existing `Dc` model.

You can see all of the new data in the [Rendered Docs](https://consul-ui-staging-fz3eifjza-hashicorp.vercel.app/ui/docs/repositories/dc)

This PR adds a new dataSource uri/method for requesting the autopilot endpoint and then merging it with the existing Datacenter information we have from the `/catalog/datacenters` endpoint, again by moving the entire DatacenterRepository to our new approach.

As there was an old endpoint to upgrade to the new approach here, there is a little more work than just adding a new endpoint, but its was largely just copying over code from other areas of the UI.

Note: We skipped a test here (one specifically for the DatacenterRepository which we are keeping, so we'd like to come back to unskip this at a later date) plus we deleted adapter and serializer tests as those no longer exist.


amyrlam (approved):
Code +1

Thanks for doing the atomic commits, makes it a lot easier to follow sans as much product context! 💯

johncowen:
Hey thanks @amyrlam ! Just so you know I'm holding this off from merge until the base branch is approved, otherwise it will change the changeset of the base PR.

johncowen:
Merged approved https://github.com/hashicorp/consul/pull/12520 on to here, gonna merge this down one down also, 🙌 @amyrlam ty!

PR #12516, opened 2022-03-03
Added Consul enterprise version image

vanphan24:
Not sure if there's a better way but I added this to ensure Consul is really looking for the Ent License when it comes up. Basically, I just set the image to deploy an Ent version of Consul so that it will then retrieve from Vault. Otherwise, user can't really tell if it's working or not. Maybe there's a better way to confirm?

### DIFF
--- a/website/content/docs/k8s/installation/vault/enterprise-license.mdx
+++ b/website/content/docs/k8s/installation/vault/enterprise-license.mdx
@@ -80,6 +80,7 @@ use the enterprise license key in Vault:
 
 ```yaml
 global:
+  image: hashicorp/consul-enterprise:1.11.0-ent
   secretsBackend:
     vault:
       enabled: true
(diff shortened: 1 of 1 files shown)


website/content/docs/k8s/installation/vault/enterprise-license.mdx
```diff
@@ -80,6 +80,7 @@ use the enterprise license key in Vault:
 
 ```yaml
 global:
+  image: hashicorp/consul-enterprise:1.11.0-ent
```
david-yu, line 83:
```suggestion
  image: hashicorp/consul-enterprise:1.11.4-ent
```

david-yu: We should use the latest image if possible.

vanphan24: yeah agreed, I shoulda checked the latest.

david-yu (approved):

david-yu:
Thanks will close, somehow can't merge because of merge conflicts. Will restart here: https://github.com/hashicorp/consul/pull/13047

PR #12518, opened 2022-03-03
reduce flakiness/raciness of errNotFound and errNotChanged blocking query tests

rboyer:
Improves tests from #12362 

These tests try to setup the following concurrent scenario:

1. (goroutine 1) execute read RPC with index=0
2. (goroutine 1) get response from (1) @ index=10
3. (goroutine 1) execute read RPC with index=10 **and block**
4. (goroutine 2) WHILE (3) is blocking, start slamming the system with stray writes that will cause the WatchSet to wakeup
5. (goroutine 2) after doing all writes, shut down the reader above
6. (goroutine 1) stops reading, double checks that it only ever woke up once (from 1)

### DIFF
--- a/agent/consul/rpc_test.go
+++ b/agent/consul/rpc_test.go
@@ -1780,5 +1780,5 @@ func rpcBlockingQueryTestHarness(
 		}
 	}
 
-	require.Equal(t, 1, count)
+	require.Equal(t, 1, count, "if this fails, then the timer likely needs to be increased above")
 }
--- a/agent/consul/rpc_test.go
+++ b/agent/consul/rpc_test.go
@@ -21,7 +21,6 @@ import (
 
 	"github.com/hashicorp/consul-net-rpc/go-msgpack/codec"
 	msgpackrpc "github.com/hashicorp/consul-net-rpc/net-rpc-msgpackrpc"
-	"github.com/hashicorp/consul-net-rpc/net/rpc"
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-memdb"
 	"github.com/hashicorp/raft"
@@ -1686,7 +1685,7 @@ func getFirstSubscribeEventOrError(conn *grpc.ClientConn, req *pbsubscribe.Subsc
 // channelCallRPC lets you execute an RPC async. Helpful in some
 // tests.
 func channelCallRPC(
-	codec rpc.ClientCodec,
+	srv *Server,
 	method string,
 	args interface{},
 	resp interface{},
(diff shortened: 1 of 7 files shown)


agent/consul/rpc_test.go
```diff
@@ -1682,23 +1683,98 @@ func getFirstSubscribeEventOrError(conn *grpc.ClientConn, req *pbsubscribe.Subsc
...
-// errNotChanged.
-//
-// Those tests are a bit racy because of the timing of the two goroutines, so
-// we relax the check for the count to be within a small range.
-//
-// The blocking query is going to wake up every interval, so use the elapsed test
-// time with that known timing value to gauge how many legit wakeups should
-// happen and then pad it out a smidge.
-func assertBlockingQueryWakeupCount(t testing.TB, interval time.Duration, start time.Time, gotCount int) {
+// channelCallRPC lets you execute an RPC async. Helpful in some
```
rboyer, line 1686: Start here

rboyer: This is a blocking query that cannot be canceled (thanks `net/rpc`). By wrapping it in a 1-buffered channel we can effectively orphan it randomly in the test and it will eventually be GCed (since it can write into the channel and then die) and if it does randomly conclude it isn't able to interact with the original caller's state. So it's a waste of a bit of ram during the test, but it's just dark matter in the system now.

agent/consul/rpc_test.go
```diff
@@ -1682,23 +1683,96 @@ func getFirstSubscribeEventOrError(conn *grpc.ClientConn, req *pbsubscribe.Subsc
...
+
+	for ctx.Err() == nil {
+		// The first iteration is an orientation iteration, as we don't pass an
+		// index value so there is no actual blocking that will happen.
+		//
+		// Since the data is not changing, we don't expect the second iteration
+		// to return soon, so we wait a bit after kicking it off before
+		// launching the write-storm.
+		var timerCh <-chan time.Time
+		if count == 1 {
```
rboyer, line 1750: `count=1` is the loop through where we expect it to block forever, so we schedule our writer storm to start 50ms after this read kicks off which is the only remaining racy aspect of this test.

kisunji (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/602837.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 8307e40f2bb54ea7060bc0765834f4b5f43992a4 onto `release/1.11.x` succeeded!

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 8307e40f2bb54ea7060bc0765834f4b5f43992a4 onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/602837)

PR #7963, opened 2020-05-27
Replace lib.TranslateKeys with a mapstructure decode hook

dnephin:
Replaces #7707 as the first of 3-4 PRs. In order to safely release the changes in #7935 we will need to first introduce forward compatibility. This PR is a pre-requisite to introducing that forward compatibility because it must happen between the normalization of HCL blocks and the decode.

This hook replaces lib.TranslateKeys and has a number of advantages:

1. Primarily, aliases for fields are defined on the field itself, making
   the aliases much easier to maintain, and more obvious to the reader.
2. TranslateKeys translation rules are not aware of structure. It could
   very easily incorrectly translate a key on one struct that was intended
   to be a translation rule for a completely different struct, leading
   to very hard to debug errors. The hook removes the need for the
   unexpected "translation rule is an empty string to indicate stop
   traversal" special case.
3. TranslateKeys attempts to duplicate a bunch of tree traversal logic
   that already exists in mapstructure. Using mapstructure for traversal
   removes the need to traverse the entire structure multiple times, and
   makes the behaviour more obvious to the reader.

This change is being made to enable a future change of replacing
PatchSliceOfMaps. TranslateKeys sits in between PatchSliceOfMaps and
mapstructure.Decode, so it must be converted to a hook first, before
PatchSliceOfMaps can be replaced by a decode hook.

PR #8646, opened 2020-09-10
Move IntermediateCertTTL to common CA config

kyhavlov:
This PR moves the IntermediateCertTTL field into the common CA config so that it can be used in the Vault provider. The documentation sort of implied this was already the case, but it was actually a field specific to the Consul CA.

PR #7213, opened 2020-02-04
add validations to ensure values are ok

i0rek:
Fixes #7201, by enforcing TTLs that workaround that problem. Comments in the code has more details about what happens.

PR #12519, opened 2022-03-03
config: finish removing lib.TranslateKeys

dnephin:
I started this work almost 2 years ago in #7963. Most uses of `lib.TranslateKeys` were replaced by `decode.HookTranslateKeys`.

The last remaining usage was the CA configuration. I tried a couple approaches, and I think this one worked without having to make major changes.

The second commit does a bit of cleanup so to the last remaining piece of code that was still accessing the opaque config directly. The correct way to access this config is from one of the `ca.Parse<ProviderName>CAConfig` functions, which is now documented.


agent/consul/leader_connect_ca.go
```diff
@@ -220,27 +220,18 @@ func (c *CAManager) initializeCAConfig() (*structs.CAConfiguration, error) {
...
-		for k, v := range dup.Config {
-			copied[k] = v
+	config = c.serverConf.CAConfig
+
+	if c.serverConf.Datacenter == c.serverConf.PrimaryDatacenter && config.ClusterID == "" {
+		id, err := uuid.GenerateUUID()
+		if err != nil {
+			return nil, err
 		}
-		copied["IntermediateCertTTL"] = connect.DefaultIntermediateCertTTL.String()
```
rboyer, line 239: Where is this default-on-read behavior happening now?

dnephin: Good question! The test should show that it does work (even when the previous config came from a version that did not set a default), but I didn't explain why it worked.  

The defaulting happens in [defaultCommonConfig](https://github.com/hashicorp/consul/blob/v1.11.4/agent/connect/ca/provider_consul_config.go#L49-L57). Every provider is setup using `Provider.Configure`, and each of them calls the appropriate `Parse<Provider>CAConfig` function. And all the parse functions use `defaultCommonConfig` to populate the target of the decode. So those values are used as defaults and remain after the decode unless something overwrites them.

It looks like when this was added in #7213 we did not default it that way. The changes in #8646 removed the need for this line.

Since the default is set before in-process while constructing the provider it doesn't need to be persisted to raft. Even if the config came from an earlier version, the defaults will be set correctly.

PR #12520, opened 2022-03-04
ui: Catalog Health Overview DataSource

johncowen:
This PR adds a new currently non-existing endpoint in Consul.

The endpoint shows the health of the catalog as a whole (as opposed to the health of the Cluster itself). Again this is _per datacenter_, so whilst we've called it `catalog/health` is lives in the Dc Repository for the moment. The backend will give us a grouping of 'totals' per partition/namespace to we do an additional aggregation on the frontend to give us totals over:

1. All partitions
2. All namespaces
3. The entire datacenter/catalog

I set the poll to 30 seconds again.

Small note: We are getting to the point that these dataSource methods could very well just be functions rather than a method on a class - i.e. Repositories are becoming less tied to EmberData models and/or Ember Services.

[Rendered Docs](https://consul-ui-staging-azaji1z0i-hashicorp.vercel.app/ui/docs/repositories/dc)

This new one is the top one in the docs, still thinking of how I can improve these new repository docs, probably need some deeplinking happening or something 🤔 . One for the future.

### DIFF
--- a/ui/packages/consul-ui/app/services/repository/dc.js
+++ b/ui/packages/consul-ui/app/services/repository/dc.js
@@ -10,6 +10,52 @@ import {
 const SECONDS = 1000;
 const MODEL_NAME = 'dc';
 
+const zero = {
+  Total: 0,
+  Passing: 0,
+  Warning: 0,
+  Critical: 0
+};
+const aggregate = (prev, body, type) => {
+
+  return body[type].reduce((prev, item) => {
+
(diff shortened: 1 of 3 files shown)


johncowen:
Thanks @amyrlam ! For this one aswell, just so you know I'm holding this off from merge until the base branches are approved, otherwise it will change the changeset of the base PRs.

johncowen:
Merging, thanks again @amyrlam !

PR #12522, opened 2022-03-04
agent: convert listener config to TLS types

mikemorris:
Replaces #11647 after the listener TLS refactoring work in #12504.

I think I've pushed the use of types versus as close to the surface I can  -`agent/config/builder.go` appears to be a boundary where generic string values capturing user config input that could be invalid can be parsed into types, and `pbconfig.TLS` appears to be another boundary for autoconfig to avoid breaking backwards compatibility.

**Reviewer Notes:**
- `ParseTLSVersion` and `ParseCiphers` both implement some logic specific to agent config (the fallback logic to support `tls10` etc values and the comma-delimited cipher string) so I don't think these would belong in `types/tls.go`
  - Per suggestion from @boxofrad, these have been subsumed into agent config builder parsing functions directly.
- It appears `CipherString` is only used for generating `pbconfig.TLS` autoconfig, should it be renamed for clarity?

**TODO:**
- [x] Update website docs per comment in https://github.com/hashicorp/consul/pull/11647#pullrequestreview-857459511
- [x] Update `tls_cipher_suites` search link in website docs
- [x] Add validation to error or warn when configuring cipher suites with TLS min verison 1.3
- [x] Close out remaining FIXMEs and TODOs
- [x] Add changelog entries


agent/config/runtime_test.go
```diff
@@ -6330,8 +6330,8 @@ func TestLoad_FullConfig(t *testing.T) {
 				CAPath:         "nu4PlHzn",
 				CertFile:       "1yrhPlMk",
 				KeyFile:        "1bHapOkL",
-				TLSMinVersion:  "mK14iOpz",
-				CipherSuites:   []uint16{tls.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA, tls.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256},
+				TLSMinVersion:  types.TLSv1_3,
```
mikemorris, line 6333: This should fail, TLS 1.3 cipher suites are not configurable.

tlsutil/config.go
```diff
@@ -36,13 +36,42 @@ type DCWrapper func(dc string, conn net.Conn) (net.Conn, error)
...
-// tlsLookup maps the tls_min_version configuration to the internal value
-var tlsLookup = map[string]uint16{
-	"":      tls.VersionTLS10, // default in golang
-	"tls10": tls.VersionTLS10,
-	"tls11": tls.VersionTLS11,
-	"tls12": tls.VersionTLS12,
-	"tls13": tls.VersionTLS13,
+// goTLSVersions maps types.TLSVersion to the Go internal value
+var goTLSVersions = map[types.TLSVersion]uint16{
+	types.TLSVersionAuto: tls.VersionTLS10, // default in golang
```
mikemorris, line 41: Should this be updated to set TLSVersionAuto to TLS 1.2 instead?

tlsutil/config.go
```diff
@@ -736,7 +753,7 @@ func (c *Configurator) AutoConfigTLSSettings() (*pbconfig.TLS, error) {
 	return &pbconfig.TLS{
 		VerifyOutgoing:       cfg.VerifyOutgoing,
 		VerifyServerHostname: cfg.VerifyServerHostname || c.autoTLS.verifyServerHostname,
-		MinVersion:           cfg.TLSMinVersion,
+		MinVersion:           types.ConsulAutoConfigTLSVersionStrings[cfg.TLSMinVersion],
```
mikemorris, line 756: Why does this not seem to be working as expected?

mikemorris: Oops, think I incorrectly updated these in the test fixtures.

mikemorris:
The remaining test failures actually look good! Invalid config that was previously being passed through is now being rejected, will update test expectations next, but this looks almost ready!

agent/config/builder.go
```diff
@@ -1969,15 +1969,30 @@ func (b *builder) cidrsVal(name string, v []string) (nets []*net.IPNet) {
 	return
 }
 
-func (b *builder) tlsCipherSuites(name string, v *string) []uint16 {
+func (b *builder) tlsVersion(name string, v *string) types.TLSVersion {
+	// TODO: should the nil check be moved into ParseTLSVersion and the arguments
+	// switched to take a pointer?
```
boxofrad, line 1974: I think swapping `nil` for something more meaningful at the builder level is the right thing to do (no changes nessasary IMO) 👍🏻

agent/config/builder.go
```diff
@@ -1969,15 +1969,30 @@ func (b *builder) cidrsVal(name string, v []string) (nets []*net.IPNet) {
...
+	if v == nil {
+		return types.TLSVersionAuto
+	}
+
+	a, err := tlsutil.ParseTLSVersion(*v)
+	if err != nil {
+		// TODO: should a warning for deprecated config values be surfaced here somehow?
```
boxofrad, line 1981: This should probably be done in `applyDeprecatedTLSConfig`:

https://github.com/hashicorp/consul/blob/6db469699dbb25cd400acc6415ad0da4352ec2ab/agent/config/deprecated.go#L178-L267

agent/config/runtime_test.go
```diff
@@ -5456,21 +5455,22 @@ func TestLoad_IntegrationWithFlags(t *testing.T) {
...
-			rt.TLS.InternalRPC.CipherSuites = []uint16{tls.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256}
+			rt.TLS.InternalRPC.CipherSuites = []types.TLSCipherSuite{types.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256}
 			rt.TLS.InternalRPC.VerifyIncoming = true
 
 			rt.TLS.HTTPS.CAFile = "default_ca_file"
 			rt.TLS.HTTPS.CAPath = "default_ca_path"
 			rt.TLS.HTTPS.CertFile = "https_cert_file"
 			rt.TLS.HTTPS.TLSMinVersion = "tls13"
-			rt.TLS.HTTPS.CipherSuites = []uint16{tls.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256}
+			// FIXME: this should fail, TLS 1.3 cipher suites are not configurable
```
boxofrad, line 5465: Good catch!

agent/consul/auto_config_endpoint.go
```diff
@@ -283,6 +283,9 @@ func (ac *AutoConfig) updateTLSSettingsInConfig(_ AutoConfigOptions, resp *pbaut
 	}
 
 	var err error
+
+	// FIXME: is the response from this not actually validated in
+	// auto_config_endpoint_test.go?
```
boxofrad, line 288: I think it is and I forgot to move the tests during refactoring #12504 🙈

tlsutil/config.go
boxofrad: Maybe, yeah! I guess it'd be a breaking-change, but I think this is all going out in a major release anyway, so probably ok?

tlsutil/config.go
```diff
@@ -36,13 +36,42 @@ type DCWrapper func(dc string, conn net.Conn) (net.Conn, error)
...
+		// removed
+		if v, ok := types.DeprecatedConsulAgentTLSVersions[tlsVersionString]; ok {
+			// TODO: log warning about deprecated config
+			return v, nil
+		}
+
+		// Only suggest non-deprecated values if configured value is invalid
+		return types.TLSVersionInvalid, tlsVersionErr
+	}
+	return v, nil
```
boxofrad, line 74: What do you think about moving this to the config package, and handling the deprecated values in `applyDeprecatedTLSConfig`? I think it's best to try to keep as much of the complexity of handling previous configuration formats in one place so that they don't leak into the rest of the codebase.

tlsutil/config.go
```diff
@@ -71,27 +100,16 @@ type ListenerConfig struct {
...
-	//
-	// todo(fs): IMHO, we should also support the raw 0xNNNN values from
-	// todo(fs): https://golang.org/pkg/crypto/tls/#pkg-constants
-	// todo(fs): since they are standardized by IANA.
-	CipherSuites []uint16
+	// We don't support the raw 0xNNNN values from
+	// https://golang.org/pkg/crypto/tls/#pkg-constants
+	// even though they are standardized by IANA because it would increase
+	// the likelihood of an operator inadvertently setting an insecure configuration
+	CipherSuites []types.TLSCipherSuite
```
boxofrad, line 112: I think this is a good call 👏🏻

tlsutil/config.go
```diff
@@ -572,7 +580,14 @@ func (c *Configurator) commonTLSConfig(state listenerConfig, cfg ListenerConfig,
 
 	// Set the cipher suites
 	if len(cfg.CipherSuites) != 0 {
-		tlsConfig.CipherSuites = cfg.CipherSuites
+		// TODO: is it safe to ignore the error case here?
+		// Should be checked on input, same as tlsConfig.MinVersion
+
+		// FIXME: move cipherSuiteLookup to be called externally, maybe
+		// in agent/config/runtime parsing before the tlsutil.Config struct is
+		// created?
```
boxofrad, line 588: Yep agree, this should be done up-front rather than when we're trying to use the cipher suites.

tlsutil/config.go
```diff
@@ -1115,29 +1132,38 @@ func ParseCiphers(cipherStr string) ([]uint16, error) {
...
+		suites[i] = types.TLSCipherSuite(cipher)
+	}
+
+	err := types.ValidateConsulAgentCipherSuites(suites)
+	if err != nil {
+		return []types.TLSCipherSuite{}, err
+	}
+
+	return suites, nil
+}
```
boxofrad, line 1155: Again, I wonder if this should live in the `config` package instead? Because the comma-separated string format is very much an "external" detail 🤷🏻

boxofrad (commented):
This is awesome! Thanks so much for taking the time to re-work your changes on top of #12504, @mikemorris 👏🏻 🎉 

I've left a few comments for your consideration. Let me know if you'd like to talk any of them through synchronously.

tlsutil/config.go
mikemorris: Yea, I think that makes sense - this function isn't called anywhere else, and the shared logic now lives in https://github.com/hashicorp/consul/blob/a1110845df0e4b1dfb588c4577484029b41bb01f/types/tls.go#L95-L101

agent/consul/auto_config_endpoint.go
mikemorris: Looks like it's tested now after most recent rebase!

tlsutil/config.go
mikemorris: Looking at this again, this actually feels like an appropriate place to handle the conversion from `types.TLSCipherSuite` to `crypto/tls` values for creating the `tls.Config` object.

agent/config/runtime_test.go
mikemorris: Config combinations like this are being handled as expected now (config builder error when specifying cipher suites alongside or with an inherited TLS min version 1.3, cipher suites omitted from config when they would be inherited by a listener config only specifying a TLS min version of 1.3).

mikemorris:
@boxofrad I can't quite figure out why these failing `TestAutoConfig_Integration` tests seem to be skipping the deprecated config parsing. It does appear to be appropriately setting `tls12` in `pbconfig.TLS` for backwards compatibility, but instead of the agent just logging a warning and updating the config to `InternalRPC.TLS.TLSMinVersion: TLSv1_2` internally, it just hits the validation error and fails to start the agent.

```
--- FAIL: TestAutoConfig_Integration (1.76s)
    agent_test.go:5065: testagent.go:107: b3a9ac58-755b-c31f-f007-24cd304f7085 test-client Error starting agent: failed to fully resolve configuration: 1 error occurred:
                * tls.defaults.tls_min_version: invalid TLS version: no matching TLS version found for tls12, please specify one of [TLS_AUTO, TLSv1_0, TLSv1_1, TLSv1_2, TLSv1_3]
```

agent/consul/auto_config_endpoint_test.go
```diff
@@ -173,7 +174,7 @@ func TestAutoConfigInitialConfiguration(t *testing.T) {
 		c.TLSConfig.InternalRPC.VerifyOutgoing = true
 		c.TLSConfig.InternalRPC.VerifyIncoming = true
 		c.TLSConfig.InternalRPC.VerifyServerHostname = true
-		c.TLSConfig.InternalRPC.TLSMinVersion = "tls12"
+		c.TLSConfig.InternalRPC.TLSMinVersion = types.TLSv1_2
```
mkeeler, line 177: Assuming this test passes, it would indicate a backwards compatibility issue with older client agents running v1.11.x or lower.

mkeeler: I see now. This is just the server config and is not setting up a mock response.

Nothing to see here.

agent/config/builder.go
```diff
@@ -1971,15 +1971,63 @@ func (b *builder) cidrsVal(name string, v []string) (nets []*net.IPNet) {
...
+}
+
+// validateTLSVersionCipherSuitesCompat checks that the specified TLS version supports
+// specifying cipher suites
+func validateTLSVersionCipherSuitesCompat(tlsMinVersion types.TLSVersion) error {
+	if tlsMinVersion == types.TLSv1_3 {
+		return fmt.Errorf("TLS 1.3 cipher suites are not configurable")
+	}
+	return nil
+}
```
boxofrad, line 2002: Thanks for adding this validation, it's a great UX improvement 🙌🏻

agent/config/deprecated.go
```diff
@@ -219,7 +221,16 @@ func applyDeprecatedTLSConfig(dep DeprecatedConfig, cfg *Config) []string {
 
 	if v := dep.TLSMinVersion; v != nil {
 		if defaults.TLSMinVersion == nil {
-			defaults.TLSMinVersion = v
+			// NOTE: This inner check for deprecated values should eventually be
+			// removed
+			if version, ok := types.DeprecatedConsulAgentTLSVersions[*v]; ok {
+				// Log warning about deprecated config values
+				warns = append(warns, fmt.Sprintf("The value for 'tls_min_version' field is deprecated, please specify one of %v", types.TLSVersions()))
```
boxofrad, line 228: nit: as we know which version the user intended to use, maybe we could make this error message even more specific?

```suggestion
				warns = append(warns, fmt.Sprintf("'tls_min_version' value '%s' is deprecated, please specify '%s' instead", v, version))
```

boxofrad (approved):
Great job 👏🏻  I've left one last nit, but please feel free to take or leave it.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/613317.

PR #12533, opened 2022-03-07
test: switch many tests to avoid using the disk

rboyer:
This drastically speeds up some tests on macos by 2x, 4x, or in isolated examples 10x.


boxofrad:
Awesome! It seems like the majority of tests don't care about writing stuff to disk, should this be the default?

On a related note: I've been wondering about how possible it'd be to make Consul use an abstract filesystem interface (e.g. [`io/fs`](https://pkg.go.dev/io/fs)) such that we could sub it with an in-memory one in tests, and maybe even inject a little failure to check our crash resilience.

PR #12534, opened 2022-03-08
poc: turn on prometheus style metrics sink by default

FFMMM:
### Overview

This PR shows a poc, minimal diff that turns on prometheus style metrics sink by default.

Per our own documentation ( [ref](https://www.consul.io/docs/agent/options#telemetry-prometheus_retention_time) ), there may be a performance impact to this. Not sure if we ever benchmarked it:

```
...
The duration [...] will aggregates all counters for the duration specified (it might have an impact on Consul's memory usage). 
...
```

### next steps

* [ ] pulse -- is this something that would be useful for folks?
* [ ] polish code 
* [ ] docs

### testing

* manually did some `curl`s over `/metrics`, `/metrics?format=prometheus` and `/metrics/prometheus` to show that they endpoints work as expected 💯

### DIFF
--- a/agent/agent_endpoint.go
+++ b/agent/agent_endpoint.go
@@ -175,6 +175,41 @@ func (s *HTTPHandlers) AgentMetrics(resp http.ResponseWriter, req *http.Request)
 	return s.agent.baseDeps.MetricsHandler.DisplayMetrics(resp, req)
 }
 
+func (s *HTTPHandlers) AgentMetricsPrometheus(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
+	// Fetch the ACL token, if any, and enforce agent policy.
+	var token string
+	s.parseToken(req, &token)
+	authz, err := s.agent.delegate.ResolveTokenAndDefaultMeta(token, nil, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// Authorize using the agent's own enterprise meta, not the token.
+	var authzContext acl.AuthorizerContext
+	s.agent.AgentEnterpriseMeta().FillAuthzContext(&authzContext)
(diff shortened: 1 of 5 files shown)


FFMMM:
IIUC, the failing tests come out of how we actually ini `TestAgent`s. Fixing them may require some test harness changes!

agent/agent_endpoint.go
```diff
@@ -175,6 +175,41 @@ func (s *HTTPHandlers) AgentMetrics(resp http.ResponseWriter, req *http.Request)
 	return s.agent.baseDeps.MetricsHandler.DisplayMetrics(resp, req)
 }
 
+func (s *HTTPHandlers) AgentMetricsPrometheus(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
+	// Fetch the ACL token, if any, and enforce agent policy.
```
FFMMM, line 179: copy pasta from `AgentMetrics`

agent/config/builder.go
```diff
@@ -942,7 +942,7 @@ func (b *builder) build() (rt RuntimeConfig, err error) {
 			StatsdAddr:                         stringVal(c.Telemetry.StatsdAddr),
 			StatsiteAddr:                       stringVal(c.Telemetry.StatsiteAddr),
 			PrometheusOpts: prometheus.PrometheusOpts{
-				Expiration: b.durationVal("prometheus_retention_time", c.Telemetry.PrometheusRetentionTime),
+				Expiration: b.durationValWithDefault("prometheus_retention_time", c.Telemetry.PrometheusRetentionTime, 1 * time.Hour),
```
FFMMM, line 945: We should be more intentional with the default value here. Went with `1hr` per some offline chats.

FFMMM:
This did not gain a lot of traction and it has been shown internally. Closing for now.

PR #12536, opened 2022-03-08
Add the `connect.enable_serverless_plugin` configuration option

erichaberkorn:
This configuration option will be used from the xDS code to selectively enable the serverless plugin.

Checklist:
   `agent/config/config.go`.
   `agent/config/runtime.go`.
   translate.
   some cases or with some values (often true). [Eric: Not required]
 - [x] **If** your new config field needs a non-zero-value default. [Eric: Not required]
      - [ ] Add that to `DefaultSource` in `agent/config/defaults.go`.
 - [x] **If** your config should take effect on a reload/HUP. [Eric: Not required]
        more of the following places:
        `TestAgent_reloadConfig*`.
 - [x] Add documentation to `website/content/docs/agent/options.mdx`. [Eric: Not required yet]

### DIFF
--- a/agent/config/builder.go
+++ b/agent/config/builder.go
@@ -669,6 +669,7 @@ func (b *builder) build() (rt RuntimeConfig, err error) {
 	connectEnabled := boolVal(c.Connect.Enabled)
 	connectCAProvider := stringVal(c.Connect.CAProvider)
 	connectCAConfig := c.Connect.CAConfig
+	serverlessPluginEnabled := boolVal(c.Connect.EnableServerlessPlugin)
 
 	// autoEncrypt and autoConfig implicitly turns on connect which is why
 	// they need to be above other settings that rely on connect.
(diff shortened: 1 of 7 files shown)


rboyer:
I didn't double check, but did you see this official checklist for adding a new config item? https://github.com/hashicorp/consul/blob/main/docs/config/checklist-adding-config-fields.md

erichaberkorn:
> I didn't double check, but did you see this official checklist for adding a new config item? https://github.com/hashicorp/consul/blob/main/docs/config/checklist-adding-config-fields.md

Thanks @rboyer. I

.changelog/12536.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+server: Add the `connect.enable_serverless_plugin` configuration option.
```
rboyer, line 2: So sometimes for new features that require a lot of changes to make function we ignore line item changes like this that all contribute to "add the big new splashy thing", so that there's just one big "make new thing" line item in the changelog at release time.

rboyer (approved):
LGTM with one comment about removing the changelog entry.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/605855.

PR #12538, opened 2022-03-08
ui: Update faker to a maintained fork

ST-DDT:
The original Faker library is no longer maintained (as you may have heard).
That's why we created a community fork that takes on the further maintenance.
Read more about it here: https://fakerjs.dev/update.html

We have completed the typescript rewrite and now try to update a few dependent projects to ensure we didn't break backwards compatibility.

I don't have a working make setup, so I will use the CI to ensure that this works.

Ref: https://github.com/faker-js/faker/issues/542

### DIFF
--- a/ui/packages/consul-ui/app/services/i18n-debug.js
+++ b/ui/packages/consul-ui/app/services/i18n-debug.js
@@ -1,7 +1,7 @@
 import I18nService, { formatOptionsSymbol } from 'consul-ui/services/i18n';
 import ucfirst from 'consul-ui/utils/ucfirst';
 
-import faker from 'faker';
+import { faker } from '@faker-js/faker';
 
 // we currently use HTML in translations, so anything 'word-like' with these
 // chars won't get translated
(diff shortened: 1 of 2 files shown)


Amier3:
Hey @ST-DDT 

Thanks for bringing this to our attention. @johncowen  could you take a look at this?

johncowen:
Hey @ST-DDT 👋 

So sorry for the delay here (thanks @Amier3 for the ping!)

I took a quick look and this looks great. Let me check it out and take a proper look though.

FYI: We only use faker in this repository/project for developer tooling, so unless i18n testing is enabled with our developer tooling, it isn't used (and also nothing will be picked up by CI). There is probably another repo I would love to get this updated in also, which would be _very_ exposed to CI, but let me check this one out first and we can go from there.

Also wanted to say, super grateful for what you folks are doing over there in fakerjs-land. Truly amazing. Super flattered you came over here contribute this here 🙇 

I'll come back by the end of today hopefully

PR #12539, opened 2022-03-08
Make the xdscommon package

erichaberkorn:
That will be shared between Consul and the serverless plugin. For now this is just a prefactoring and shouldn't change the behavior of anything.

### DIFF
--- a/agent/xds/server.go
+++ b/agent/xds/server.go
@@ -23,6 +23,7 @@ import (
 	agentgrpc "github.com/hashicorp/consul/agent/grpc"
 	"github.com/hashicorp/consul/agent/proxycfg"
 	"github.com/hashicorp/consul/agent/structs"
+	"github.com/hashicorp/consul/agent/xds/xdscommon"
 	"github.com/hashicorp/consul/tlsutil"
 )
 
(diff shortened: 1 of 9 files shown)


agent/xds/server.go
```diff
@@ -37,23 +38,6 @@ var StatsGauges = []prometheus.GaugeDefinition{
...
-	ClusterType = apiTypePrefix + "envoy.config.cluster.v3.Cluster"
-
-	// RouteType is the TypeURL for Route discovery responses.
-	RouteType = apiTypePrefix + "envoy.config.route.v3.RouteConfiguration"
-
-	// ListenerType is the TypeURL for Listener discovery responses.
-	ListenerType = apiTypePrefix + "envoy.config.listener.v3.Listener"
-
 	// PublicListenerName is the name we give the public listener in Envoy config.
 	PublicListenerName = "public_listener"
```
erichaberkorn, line 42: For now, I only copy over the minimal code I need based on the POC but moving other constants like this probably makes sense too.

rboyer: IIRC many of these constants exist in the go-control-plane library too. Things changed during the xds v2-to-v3 conversion and some of the library constants were the older deprecated names and some were the new ones and it was unreliable so i switched a lot of things to be local constants instead I could control. Maybe after a future upgrade we could go back to library constants for these things.

erichaberkorn: That makes sense to me. From the perspective of the serverless patcher, it needs to match whatever Consul uses so I think it's still worth putting these here for now.

rboyer (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/605810.

PR #12452, opened 2022-02-25
ui: DistributionMeter Component

johncowen:
[Rendered Docs](https://consul-ui-staging-paqi22d2o-hashicorp.vercel.app/ui/docs/components/distribution-meter#CONSUL_LATENCY=3000)

I turned up the BlockingQuery wakeup rate on that link ^ to 3 seconds so it's easy to see what it looks like when data is being updated. Note you'll randomly get a set of data with different states, sometimes those states will be 100% in critical for example, just wait for different data to come in if so)

Honestly pretty happy that all this is just 5-6 lines of JS.

Two PRs that will need to be added ontop of here that I'll PR after review are:

1. Compile out the documentation metadata from the Glimmer templates (mostly the descriptions)
2. Tool for extracting the metadata from the Glimmer templates and inserting it into the README.mdx files on git stage.

I'll do those 2 once these are approved, for the moment just pretend they exist already!

PR #12451, opened 2022-02-25
ui: CustomElement component

johncowen:
Builds on `attach-shadow`, `adopt-styles` and `ShadowTemplate`, this PR adds `ShadowHost` and finally (😅)`CustomElement`.

[Rendered Docs](https://consul-ui-staging-3tqpfqutu-hashicorp.vercel.app/ui/docs/components/custom-element)

(also [ShadowHost Docs](https://consul-ui-staging-3tqpfqutu-hashicorp.vercel.app/ui/docs/components/shadow-host))

CustomElement is a renderless component to help with the creation of native HTML Custom Elements along with runtime type checking and self-documentation for attributes, slots, cssprops and cssparts. As you will probably see there is a little more work to come here. But in the same breath, everything would be fine to go in as is.

I've left a bunch of utility functions in the top of the component for the moment, as I'm still unsure whether now is the time to PR all of this as a separate addon (`ember-chocolate`). If it turns out now is the time to do that, it will be much easier to move all of this into and external addon rather than files dotted all over the place in `utils` etc. There are a few more PRs to add (and feature work to do) before I'll know whether its the right time or not.

I plan on branching off of here and PRing a small component using this to give folks an idea of real-world usage, if that plan changes, I'll ping again here. Theres also a PR to come ontop somewhere to do the documentation related things.

PR #12540, opened 2022-03-09
ui: Add CustomElement Transform

johncowen:
https://github.com/hashicorp/consul/pull/12451 finished up the beginnings of an initial approach to using ShadowDOM and CustomElements in Ember.

ShadowDOM and CustomElements gives us almost complete isolation of application components, the clearest benefit of this right now is almost complete isolation of CSS. This has the huge benefit of having more control over the cascading nature of CSS. LightDOM CSS does not punch through the ShadowDOM boundary unless you want it to via the use of CSS parts and CSS custom properties.

The benefits here are huge, and longer term as we add more and more of these components the CSS is less likely to drift slightly as the project scales.

Here we are fixing/preventing the issue at source rather than curing after the fact.

This also means less tooling/approaches are required around things like CSS regression testing as the components are far less likely to break if you are changing an unrelated component/global CSS rule. So the problem _partly_ just goes away and therefore requires less testing.

No alarms and no surprises.

With that said, CSS regression testing and individual component testing still has its place, but once a component is built, unless you are actually changing the code, it is no where near as necessary. @evrowe I know we spoke only yesterday about how I'd been suggesting we should be using some sort of CSS regression testing for quite some time now. Just to be clear, I am still on the same page as you and most definitely pro to using it. But I realized off the back of our convo that moving forwards with this approach means its far less necessary and this is something I only realised after we spoke yesterday, so thanks for bringing it up. It's good to talk.

We can also use further Web Platform features such as native slots for conditional slotting, multiple slots of the same name and supporting a11y by providing an explicit bridge between the LightDOM and the ShadowDOM.

Back to this PR 😅 

We also added almost self-documentation features and fairly rudimentary type checking for Glimmerized web components while we were here. Both features can make for a higher quality codebase that is easier to onboard onto for other folks. I've been gathering insights and experience on this from various members of my team both past and present over the past few weeks to understand what benefit others see and how successful current approaches are in documenting and type checking web components. This was a good opportunity to use that to hopefully further increase the ease of onboarding for our already highly documented and reasonably 'type aware' application.

@jgwhite @randallmorey and others no longer here 😿 @ZahraTee and @junyper, thanks all for the help 🙇  

The one problem with the documentation of course is that we don't want to be bundling unnecessarily code/copy to our users if we can at all avoid it, and in writing the documentation in with the code for the component means bundling all this copy in where it's not needed, ouch!

This PR adds an Ember-Glimmer/Handlebars transform to drop the documentation copy out of the build, therefore it is not bundled into the code either at development time or production build time. 

Just one last quick note, this is part of the `ember-chocolate` addon that folks internally are aware of. As mentioned a in another PR I don't think this is quite the right time treat that as a separate addon, so I just PRed this bit here as an ember 'in repo addon' at some point in the future it will probably all be put back together as a proper addon. Oh there's also a similar piece that extracts the same attributes and commits it to the documentation README.mdx files so we can view those in GH or any other documentation system like docfy. That's not urgent though so it can wait.

P.S.

There's also a reference to this work here:

https://github.com/hashicorp/consul/pull/12452

> Two PRs that will need to be added ontop of here that I'll PR after review are:
> Compile out the documentation metadata from the Glimmer templates (mostly the descriptions)

> I'll do those 2 once these are approved

This PR is just one of those ^

### DIFF
--- a/ui/packages/consul-ui/lib/custom-element/index.js
+++ b/ui/packages/consul-ui/lib/custom-element/index.js
@@ -1,0 +1,63 @@
+'use strict';
+
+module.exports = {
+  name: require('./package').name,
+  getTransform: function() {
+    return {
+      name: 'custom-element',
+      plugin: class {
+        transform(ast) {
+          this.syntax.traverse(ast, {
+            ElementNode: (node) => {
+              if(node.tag === 'CustomElement') {
+                node.attributes = node.attributes
+                  // completely remove these ones, they are not used runtime
+                  // element is potentially only temporarily being removed
+                  .filter(item => !['element', 'description', 'slots', 'cssparts'].includes(`${item.name.substr(1)}`))
+                  .map(item => {
+                    switch(true) {
+                      // these ones are ones where we need to remove the documentation only
+                      // the attributes themselves are required at runtime
+                      case ['attrs', 'cssprops'].includes(`${item.name.substr(1)}`):
+                        item.value.params = item.value.params.map(item => {
+                          // we can't use arr.length here as we don't know
+                          // whether someone has used the documentation entry
+                          // in the array or not We use the hardcoded `3` for
+                          // the moment if that position needs to change per
+                          // property we can just add more cases for the
+                          // moment
+                          item.params = item.params.filter((item, i, arr) => i < 3);
+                          return item;
+                        });
+                        break;
+
+                    }
+                    return item;
+                  });
+              }
+            },
+          });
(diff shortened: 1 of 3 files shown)


amyrlam (approved):
code 👍

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/605684.

Issue #12048, opened 2022-01-12
ca/ leaf cert: non blocking queries issued after blocking queries may block


FFMMM:
#### Overview of the Issue

[This commit](https://github.com/hashicorp/consul/commit/384d497f267aefbcd173e7b1fd7fca9f762a82fe) introduces a new bug.

The intent of the original commit was to fix situations where expired certs or invalid certs (after a root rotation) were returned.

Now, however, a non blocking query issued after a blocking query can an end up blocking on the former blocking query's completion or timeout.
 
#### Reproduction Steps

```
1. set up a CA provider
2. curl {CONSUL_HTTP_ADDR}/v1/agent/connect/ca/leaf/something\?index\={MODIFY_INDEX} -k
3. cancel the query
4. curl {CONSUL_HTTP_ADDR}/v1/agent/connect/ca/leaf/something -k
```
The last query will be blocking but it shouldn't! 

#### Explanation

The blocking query sets up a Waiter. If that query is cancelled by the client, then the following non-blocking query needs to wait on that existing Waiter for the cache key because of `MustRevalidate` being set for all non-blocking queries. But that Waiter channel won't close until the previous blocking query's Fetch returns, and that won't happen until a new cert needs to be generated.

#### Workaround

A possible workaround was suggested to use `wait=N` where n is << small.

Another way was to modify the QueryOptions prior to kicking off the watch go routines::

```
	leafCert, _, err := c.consul.Agent().ConnectCALeaf(c.service, &api.QueryOptions{
		WaitTime:  1 * time.Second,
	})
	if err != nil {
		return err
	}
	c.handleLeafWatch(nil, leafCert)
```

#### Possible solutions

1. A solution that Daniel mentioned before was to rethink this cache type.

```
We'd need a LeafCertStore that is basically a map[key]LeafCert . We'd pass the LeafCertStore to the cache-type, so that it can store the leaf cert after creating it. Then we'd need some way to expire the entries in the LeafCertStore , either by using a separate TTL, or we could have the cache entry clear out the LeafCertStore.

The HTTPHandler would only use the cache for blocking queries. Non-blocking queries would go directly to the LeafCertStore, so would avoid the problem entirely.
```

2. Freddy suggested we think about supporting cancellation in the cache

```
1. Get a context and cancel func on calls to fetch from the cache
2. Plumb the context through the calls to Fetch the cache types
3. Store the cancel func in the cacheEntry
[...]

PR #12554, opened 2022-03-14
Test fixing cache revalidation behavior

kisunji:
Possibly fixes https://github.com/hashicorp/consul/issues/12048

### DIFF
--- a/agent/cache/cache.go
+++ b/agent/cache/cache.go
@@ -557,6 +557,7 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
 	// If we already have an entry, it is actively fetching, and we don't need
 	// to revalidate the cache, then return the currently active waiter.
 	if ok && entry.Fetching {
+		// If we need to revalidate, return immediately
 		if r.Info.MustRevalidate {
 			ch := make(chan struct{})
 			close(ch)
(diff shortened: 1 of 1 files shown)


agent/cache/cache.go
```diff
@@ -554,9 +554,14 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
...
-	// If we already have an entry and it is actively fetching, then return
-	// the currently active waiter.
+	// If we already have an entry, it is actively fetching, and we don't need
+	// to revalidate the cache, then return the currently active waiter.
 	if ok && entry.Fetching {
+		if r.Info.MustRevalidate {
+			ch := make(chan struct{})
+			close(ch)
+			return ch
+		}
```
kisunji, line 564: Not too familiar with this part of the code but this seems to solve the problems and not fail any tests.

Basically, scenario is:
```
1. GET http://localhost:8500/v1/agent/connect/ca/leaf/something?index=35
2. Cancel blocking query
```
I believe this would leave entry in a `Fetching` state until CA was rotated.

Any non-blocking queries would fetch this entry then be forced to block anyways since the `entry.Waiter` would not send until the CA rotation.

This patch will force any non-BQ requests with `MustRevalidate` where the preceding request left `entry` in a `Fetching` state to immediately `goto RETRY_GET` and return the cached entry instead.

I'd appreciate some 👀  to check if this is a sane solution

rboyer:
Patch carried over to https://github.com/hashicorp/consul/pull/12820

Issue #12481, opened 2022-03-01
Add authorizer information into PermissionDeniedByACL errors. 


markan:
Connect up authorizer information with the PermissionDenied error generation code. Currently we don't have context information around the source of the authority (token, etc) at the time the error is generated. 

This should implemented along the lines suggested in https://github.com/hashicorp/consul/issues/11337

This has a couple of subtasks:

1) ACLResolveResult and AllowAuthorizer should be unified. (Probably as AllowAuthorizer) Currently the ACLResolver (and ACLResolveResult) are in the agent/consul package, but for general utility it should be moved to the acl package.

For now we may keep ACLResolver in agent/consul, depending on implementation difficulty.

2) acl.Authorizer should be replaced with the new AllowAuthorizer authorizer and threaded through the call stack. 

Once that is done, the error generation code should have appropriate context to identify the token used.

PR #12567, opened 2022-03-16
Add source of authority annotations to the PermissionDeniedError output.

markan:
This extends the acl.AllowAuthorizer with source of authority information.

The next step is to unify the AllowAuthorizer and ACLResolveResult structures; that will be done in a separate PR.

Part of https://github.com/hashicorp/consul/issues/12481

Signed-off-by: Mark Anderson [manderson@hashicorp.com](mailto:manderson@hashicorp.com)

### DIFF
--- a/acl/errors.go
+++ b/acl/errors.go
@@ -127,11 +127,13 @@ func PermissionDeniedByACLUnnamed(authz Authorizer, context *AuthorizerContext, 
 	return PermissionDeniedError{Accessor: ExtractAccessorInformation(authz), Resource: resource, AccessLevel: accessLevel, ResourceID: desc}
 }
 
-func ExtractAccessorInformation(authz Authorizer) string {
+func ExtractAccessorInformation(authz interface{}) string {
 	var accessor string
 	switch v := authz.(type) {
 	case AllowAuthorizer:
 		accessor = v.AccessorID
+	case string:
+		accessor = v
 	}
 	return accessor
 }
--- a/acl/errors.go
+++ b/acl/errors.go
@@ -117,12 +117,21 @@ func PermissionDenied(msg string, args ...interface{}) PermissionDeniedError {
 }
 
 // TODO Extract information from Authorizer
-func PermissionDeniedByACL(_ Authorizer, context *AuthorizerContext, resource Resource, accessLevel AccessLevel, resourceID string) PermissionDeniedError {
+func PermissionDeniedByACL(authz Authorizer, context *AuthorizerContext, resource Resource, accessLevel AccessLevel, resourceID string) PermissionDeniedError {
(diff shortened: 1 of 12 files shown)


acl/testing.go
```diff
@@ -6,7 +6,7 @@ import (
 	"testing"
 )
 
-func RequirePermissionDeniedError(t testing.TB, err error, _ Authorizer, _ *AuthorizerContext, resource Resource, accessLevel AccessLevel, resourceID string) {
+func RequirePermissionDeniedError(t testing.TB, err error, authz *Authorizer, _ *AuthorizerContext, resource Resource, accessLevel AccessLevel, resourceID string) {
```
eculver, line 9: Is `Authorizer` an interface here? I don't think it should be a pointer to the interface if that's the case.

acl/errors.go
```diff
@@ -117,12 +117,23 @@ func PermissionDenied(msg string, args ...interface{}) PermissionDeniedError {
...
+}
+
+func ExtractAccessorID(authz interface{}) string {
+	var accessor string
+	switch v := authz.(type) {
+	case AllowAuthorizer:
+		accessor = v.AccessorID
+	case string:
+		accessor = v
+	}
```
eculver, line 137: Are there other authorizers that can exist here (eg. `ChainedAuthorizer`)? If so, that would result in an empty return value. Is this ok?

acl/errors.go
```diff
@@ -117,12 +117,23 @@ func PermissionDenied(msg string, args ...interface{}) PermissionDeniedError {
...
 }
 
-func PermissionDeniedByACLUnnamed(_ Authorizer, context *AuthorizerContext, resource Resource, accessLevel AccessLevel) PermissionDeniedError {
+func PermissionDeniedByACLUnnamed(authz Authorizer, context *AuthorizerContext, resource Resource, accessLevel AccessLevel) PermissionDeniedError {
 	desc := NewResourceDescriptor("", context)
-	return PermissionDeniedError{Accessor: "", Resource: resource, AccessLevel: accessLevel, ResourceID: desc}
+	return PermissionDeniedError{Accessor: ExtractAccessorID(authz), Resource: resource, AccessLevel: accessLevel, ResourceID: desc}
+}
+
+func ExtractAccessorID(authz interface{}) string {
```
eculver, line 130: Does this need to be exported? It seems to only be called internally by the `acl` package.

acl/errors.go
markan: Empty value is ok; the string conversion routine handles that gracefully.

kisunji (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/608725.

PR #12582, opened 2022-03-21
pre register new rpc metric, rename metric

FFMMM:
### Overview

We want to pre register all prometheus metrics so we don't expire them by default. The first commit does that for our new rpc metric.

The second commit renames the metric bc there already exists a `rpc.request` and we want to avoid confusion with that.

### DIFF
--- a/agent/rpc/middleware/interceptors.go
+++ b/agent/rpc/middleware/interceptors.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/armon/go-metrics"
+	"github.com/armon/go-metrics/prometheus"
 	"github.com/hashicorp/consul-net-rpc/net/rpc"
 	"github.com/hashicorp/go-hclog"
 )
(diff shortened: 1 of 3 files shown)


agent/rpc/middleware/interceptors.go
```diff
@@ -19,9 +20,16 @@ import (
 // used for this historically is "RPC", so we continue to use that here.
 const RPCTypeInternal = "internal"
 
-var metricRPCRequest = []string{"rpc", "server", "request"}
+var metricRPCRequest = []string{"rpc", "server", "call"}
 var requestLogName = "rpc.server.request"
 
+var NewRPCCounters = []prometheus.CounterDefinition{
+	{
+		Name: metricRPCRequest,
```
kisunji, line 28: any reason to not in-line here? Do we use `metricRPCRequest` anywhere else?

kisunji (approved):
minor nit but LGTM

agent/rpc/middleware/interceptors.go
eculver: looks like the others in `agent/setup.go` inline as a `[]string` too, so would be consistent with that at the very least 🤷🏼

FFMMM: I was using this a bunch for some testing here: https://github.com/hashicorp/consul/pull/12585 and in the `Record()` call below

🙈 Unsure what the most gopher way to do this is but I am willing to punt and take another pass over this as needed!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/609957.

PR #12583, opened 2022-03-21
Add IAM Auth Method

pglass:
This adds an IAM auth method to Consul.

This takes a good bit of code from the Vault IAM auth method, which is added to an internal package `internal/iamauth` package.

### Usage
Configuring the auth method (example):
```
consul acl auth-method create -type aws-iam \
  -name <name> \
  -config='{
    "BoundIAMPrincipalARNs": [],
    "EnableIAMEntityDetails": false,
    "IAMEntityTags": []
    "MaxRetries": -1,
    "IAMEndpoint": "",
    "STSEndpoint": "",
    "STSRegion": "",
    "ServerIDHeaderValue": "",
    "AllowedSTSHeaderValues": [],
}'
```

Login
```
consul login -type aws-iam -method <name> -aws-auto-bearer-token -token-sink-file token.txt
```

### DIFF
--- a/agent/consul/authmethod/awsauth/aws.go
+++ b/agent/consul/authmethod/awsauth/aws.go
@@ -1,0 +1,191 @@
+package awsauth
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/consul/agent/consul/authmethod"
+	"github.com/hashicorp/consul/agent/structs"
+	"github.com/hashicorp/consul/internal/iamauth"
+	"github.com/hashicorp/go-hclog"
+)
+
+const (
+	authMethodType string = "aws-iam"
+
+	IAMServerIDHeaderName  string = "X-Consul-IAM-ServerID"
+	GetEntityMethodHeader  string = "X-Consul-IAM-GetEntity-Method"
+	GetEntityURLHeader     string = "X-Consul-IAM-GetEntity-URL"
+	GetEntityHeadersHeader string = "X-Consul-IAM-GetEntity-Headers"
+	GetEntityBodyHeader    string = "X-Consul-IAM-GetEntity-Body"
+)
+
+func init() {
+	// register this as an available auth method type
+	authmethod.Register(authMethodType, func(logger hclog.Logger, method *structs.ACLAuthMethod) (authmethod.Validator, error) {
+		v, err := NewValidator(logger, method)
+		if err != nil {
+			return nil, err
+		}
+		return v, nil
+	})
+}
+
+type Config struct {
+	// BoundIAMPrincipalARNs are the trusted AWS IAM principal ARNs that are permitted
+	// to login to the auth method. These can be the exact ARNs or wildcards. Wildcards
+	// are only supported if EnableIAMEntityDetails is true.
+	BoundIAMPrincipalARNs []string `json:",omitempty"`
+
+	// EnableIAMEntityDetails will fetch the IAM User or IAM Role details to include
+	// in binding rules. Required if wildcard principal ARNs are used.
+	EnableIAMEntityDetails bool `json:",omitempty"`
+
+	// IAMEntityTags are the specific IAM User or IAM Role tags to include as selectable
+	// fields in the binding rule attributes. Requires EnableIAMEntityDetails = true.
+	IAMEntityTags []string `json:",omitempty"`
+
+	// ServerIDHeaderValue adds a X-Consul-IAM-ServerID header to each AWS API request.
+	// This helps protect against replay attacks.
+	ServerIDHeaderValue string `json:",omitempty"`
+
+	// MaxRetries is the maximum number of retries on AWS API requests for recoverable errors.
+	MaxRetries int `json:",omitempty"`
+	// IAMEndpoint is the AWS IAM endpoint where iam:GetRole or iam:GetUser requests will be sent.
+	// Note that the Host header in a signed request cannot be changed.
+	IAMEndpoint string `json:",omitempty"`
+	// STSEndpoint is the AWS STS endpoint where sts:GetCallerIdentity requests will be sent.
+	// Note that the Host header in a signed request cannot be changed.
+	STSEndpoint string `json:",omitempty"`
+	// STSRegion is the region for the AWS STS service. This should only be set if STSEndpoint
+	// is set, and must match the region of the STSEndpoint.
+	STSRegion string `json:",omitempty"`
+
+	// AllowedSTSHeaderValues is a list of additional allowed headers on the sts:GetCallerIdentity
+	// request in the bearer token. A default list of necessary headers is allowed in any case.
+	AllowedSTSHeaderValues []string `json:",omitempty"`
+
+	enterpriseConfig `mapstructure:",squash"`
+}
+
+func (c *Config) convertForLibrary() *iamauth.Config {
+	return &iamauth.Config{
+		BoundIAMPrincipalARNs:  c.BoundIAMPrincipalARNs,
+		EnableIAMEntityDetails: c.EnableIAMEntityDetails,
+		IAMEntityTags:          c.IAMEntityTags,
+		ServerIDHeaderValue:    c.ServerIDHeaderValue,
+		MaxRetries:             c.MaxRetries,
+		IAMEndpoint:            c.IAMEndpoint,
+		STSEndpoint:            c.STSEndpoint,
+		STSRegion:              c.STSRegion,
+		AllowedSTSHeaderValues: c.AllowedSTSHeaderValues,
+
+		ServerIDHeaderName:     IAMServerIDHeaderName,
+		GetEntityMethodHeader:  GetEntityMethodHeader,
+		GetEntityURLHeader:     GetEntityURLHeader,
+		GetEntityHeadersHeader: GetEntityHeadersHeader,
+		GetEntityBodyHeader:    GetEntityBodyHeader,
+	}
+}
+
+type Validator struct {
+	name   string
+	config *iamauth.Config
+	logger hclog.Logger
+
+	auth *iamauth.Authenticator
+}
+
+func NewValidator(logger hclog.Logger, method *structs.ACLAuthMethod) (*Validator, error) {
+	if method.Type != authMethodType {
+		return nil, fmt.Errorf("%q is not an AWS IAM auth method", method.Name)
+	}
+
+	var config Config
+	if err := authmethod.ParseConfig(method.Config, &config); err != nil {
+		return nil, err
+	}
+	iamConfig := config.convertForLibrary()
+
+	auth, err := iamauth.NewAuthenticator(iamConfig, logger)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Validator{
+		name:   method.Name,
+		config: iamConfig,
+		logger: logger,
+		auth:   auth,
+	}, nil
+}
+
+// Name implements authmethod.Validator.
+func (v *Validator) Name() string { return v.name }
+
+// Stop implements authmethod.Validator.
+func (v *Validator) Stop() {}
+
+// ValidateLogin implements authmethod.Validator.
+func (v *Validator) ValidateLogin(ctx context.Context, loginToken string) (*authmethod.Identity, error) {
+	details, err := v.auth.ValidateLogin(ctx, loginToken)
+	if err != nil {
+		return nil, err
+	}
+
+	vars := map[string]string{
+		"entity_name": details.EntityName,
+		"entity_id":   details.EntityId,
+		"account_id":  details.AccountId,
+	}
+	fields := &awsSelectableFields{
+		EntityName: details.EntityName,
+		EntityId:   details.EntityId,
+		AccountId:  details.AccountId,
+	}
+
+	if v.config.EnableIAMEntityDetails {
+		vars["entity_path"] = details.EntityPath
+		fields.EntityPath = details.EntityPath
+		fields.EntityTags = map[string]string{}
+		for _, tag := range v.config.IAMEntityTags {
+			vars["entity_tags."+tag] = details.EntityTags[tag]
+			fields.EntityTags[tag] = details.EntityTags[tag]
+		}
+	}
+
+	result := &authmethod.Identity{
+		SelectableFields: fields,
+		ProjectedVars:    vars,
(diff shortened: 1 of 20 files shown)


go.sum
```diff
@@ -265,6 +265,8 @@ github.com/hashicorp/go-retryablehttp v0.6.7 h1:8/CAEZt/+F7kR7GevNHulKkUjLht3CPm
 github.com/hashicorp/go-retryablehttp v0.6.7/go.mod h1:vAew36LZh98gCBJNLH42IQ1ER/9wtLZZ8meHqQvEYWY=
 github.com/hashicorp/go-rootcerts v1.0.2 h1:jzhAVGtqPKbwpyCPELlgNWhE1znq+qwJtW5Oi2viEzc=
 github.com/hashicorp/go-rootcerts v1.0.2/go.mod h1:pqUvnprVnM5bf7AOirdbb01K4ccR319Vf4pU3K5EGc8=
+github.com/hashicorp/go-secure-stdlib/strutil v0.1.2 h1:kes8mmyCpxJsI7FTwtzRqEy9CdjCtrXrXGuOpxEA7Ts=
+github.com/hashicorp/go-secure-stdlib/strutil v0.1.2/go.mod h1:Gou2R9+il93BqX25LAKCLuM+y9U2T4hlwvT1yprcna4=
```
pglass, line 269: This is used for only 2-3 functions. There's also one strutil function here copied over: https://github.com/hashicorp/consul/blob/main/internal/go-sso/oidcauth/internal/strutil/util.go.

Thoughts on adding this dependency vs copying the functions over?

rboyer: There's `lib/stringslice` in consul that has `stringslice.Contains` which is the same thing as `strutil.StrListContains`

rboyer:
Probably good to in the PR somewhere and likely also in a comment in one of the packages indicating which version/sha of various vault libraries the various AWS bits were lifted from. That way there's a paper trail to figure out which possible future patches on the vault side need to be re-replicated here.

internal/iamauth/auth.go
```diff
@@ -0,0 +1,336 @@
...
+
+// https://github.com/hashicorp/vault/blob/ba533d006f2244103648785ebfe8a9a9763d2b6e/builtin/credential/aws/path_login.go#L1321-L1361
+func (a *Authenticator) validateIdentity(clientArn string) error {
+	if strutil.StrListContains(a.config.BoundIAMPrincipalARNs, clientArn) {
+		// Matches one of BoundIAMPrincipalARNs, so it is trusted
+		return nil
+	}
+	if a.config.EnableIAMEntityDetails {
+		for _, principalArn := range a.config.BoundIAMPrincipalARNs {
+			if strings.HasSuffix(principalArn, "*") && strutil.GlobbedStringsMatch(principalArn, clientArn) {
```
rboyer, line 135: Should be pretty easy to lift `GlobbedStringsMatch` over into `github.com/hashicorp/consul/lib` to avoid needing a little dependency for just two functions (one of which already exists in `lib/stringslice`)

command/login/login.go
```diff
@@ -36,6 +36,8 @@ type cmd struct {
 	tokenSinkFile   string
 	meta            map[string]string
 
+	aws *AWSLogin
```
rboyer, line 39: Why make this a pointer if it's always initialized? You could just skip line 64 and make this `aws AWSLogin`

internal/iamauth/responses/arn.go
```diff
@@ -0,0 +1,94 @@
...
+	AccountNumber string
+	Type          string
+	Path          string
+	FriendlyName  string
+	SessionInfo   string
+}
+
+// https://github.com/hashicorp/vault/blob/ba533d006f2244103648785ebfe8a9a9763d2b6e/builtin/credential/aws/path_login.go#L1482-L1530
+// However, instance profiles are not support in Consul.
+func ParseArn(iamArn string) (*ParsedArn, error) {
```
pglass, line 20: Compared to the Vault implementation of this, I added a few checks here for empty strings on parts, and it doesn't support ARNs for EC2 instance profiles, since those are only supported by Vault for the EC2 auth method.

(This could be simplified a bit by using [arn.ParseArn](https://docs.aws.amazon.com/sdk-for-go/api/aws/arn/#Parse))

internal/iamauth/responses/responses.go
```diff
@@ -0,0 +1,92 @@
...
+type IAMEntity interface {
+	EntityPath() string
+	EntityArn() string
+	EntityName() string
+	EntityId() string
+	EntityTags() map[string]string
+}
+
+var _ IAMEntity = (*Role)(nil)
+var _ IAMEntity = (*User)(nil)
```
pglass, line 31: Not sure if we like these sorts of "implements interface" assertion hacks in Consul? It was useful to catch myself during dev, but unit tests should take care of this going forward.

internal/iamauth/responses/arn.go
```diff
@@ -0,0 +1,94 @@
+package responses
```
pglass, line 1: I split off this package to avoid an import cycle in the tests.

* I had `iamauth > iamauthtest > iamauth`
* Now, `iamauth > responses` and `iamauth > responsestest` and `iamauth > iamauthtest`

Works, but seems kind of overly nested. I followed what `internal/go-sso` did for test utils, so curious if I'm using `<package>test` the right way, or if I could organize this differently.

internal/iamauth/token.go
```diff
@@ -0,0 +1,335 @@
...
+		}
+		token.iamEntityType = entityType
+
+		// TODO:
+		// if err := t.validateServerIDHeader(); err != nil {
+		// 	return err
+		// }
+		// if err := t.validateAllowedSTSHeaderValues(); err != nil {
+		// 	return err
+		// }
```
pglass, line 121: TODO: this todo is done!

internal/iamauth/token.go
```diff
@@ -0,0 +1,335 @@
...
+		// the semantics
+		default:
+			// Not expecting any other values
+			return fmt.Errorf("iam_request_body contains unexpected values")
+		}
+	}
+	return nil
+}
+
+func (t *BearerToken) validateIAMEntityBody() (IAMEntityType, error) {
```
pglass, line 164: TODO: refactor into common response body parsing/validating function(s) (and avoid the loop label - ew)

internal/iamauth/util.go
```diff
@@ -0,0 +1,167 @@
...
+	GetEntityBodyHeader    string
+}
+
+// GenerateLoginData populates the necessary data to send for the bearer token.
+// https://github.com/hashicorp/go-secure-stdlib/blob/main/awsutil/generate_credentials.go#L232-L301
+func GenerateLoginData(in *LoginInput) (map[string]interface{}, error) {
+
+	// Use the credentials we've found to construct an STS session
+	// TODO: Do we need this? Or can just assume the region from the credentials?
+	// region, err := GetRegion(configuredRegion)
```
pglass, line 43: TODO: Need to dive into this. Relates to my other comment about a custom credential chain vs using the AWS SDK's standard credential chain.

internal/iamauth/util.go
```diff
@@ -0,0 +1,167 @@
...
+		"iam_request_headers":     base64.StdEncoding.EncodeToString(headersJson),
+		"iam_request_body":        base64.StdEncoding.EncodeToString(requestBody),
+	}, nil
+}
+
+// STS is a really weird service that used to only have global endpoints but now has regional endpoints as well.
+// For backwards compatibility, even if you request a region other than us-east-1, it'll still sign for us-east-1.
+// See, e.g., https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html#id_credentials_temp_enable-regions_writing_code
+// So we have to shim in this EndpointResolver to force it to sign for the right region
+func stsSigningResolver(service, region string, optFns ...func(*endpoints.Options)) (endpoints.ResolvedEndpoint, error) {
```
pglass, line 117: TODO: Figure out if the [STSRegionalEndpoint](https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/#STSRegionalEndpoint) config option takes care of this for us.

internal/iamauth/token.go
```diff
@@ -0,0 +1,335 @@
...
+
+type IAMEntityType string
+
+const (
+	IAMEntityTypeNone IAMEntityType = ""
+	IAMEntityTypeRole               = "role"
+	IAMEntityTypeUser               = "user"
+)
+
+func NewBearerToken(loginToken string, config *Config) (*BearerToken, error) {
```
pglass, line 62: Some of the Vault parsing and validation of the login string is reframed here as this BearerToken type.

internal/iamauth/responses/responses.go
rboyer: I love them.

go.sum
pglass: That's perfect. I switched lib/stringslice and added the string glob/wildcard function to lib/glob.go (+ unit test)

command/login/aws.go
```diff
@@ -0,0 +1,138 @@
...
+// It will discover AWS credentials which are used to sign AWS API requests.
+// Alternatively, static credentials can be passed as flags.
+//
+// The bearer token contains a signed sts:GetCallerIdentity request.
+// If aws-include-entity is specified, a signed iam:GetRole or iam:GetUser request is
+// also included. The AWS credentials are used to retrieve the current user's role
+// or user name for the iam:GetRole or iam:GetUser request.
+func (a *AWSLogin) createAWSBearerToken() (string, error) {
+	cfg := aws.Config{
+		Endpoint: aws.String(a.stsEndpoint),
```
rboyer, line 89: Are `stsEndpoint` and `region` required? I don't see verification implying they are .

command/login/aws.go
```diff
@@ -0,0 +1,137 @@
...
+
+	var creds *credentials.Credentials
+	if a.accessKeyId != "" {
+		// Use creds from flags.
+		creds = credentials.NewStaticCredentials(
+			a.accessKeyId, a.secretAccessKey, a.sessionToken,
+		)
+	} else {
+		// Session loads creds from standard sources (env vars, file, EC2 metadata, ...)
+		sess, err := session.NewSessionWithOptions(session.Options{Config: cfg})
```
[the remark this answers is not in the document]
rboyer: It might be worth figuring out what value `GenerateCredentialChain` is bringing and erring on the side of using it over not.

command/login/login_test.go
```diff
@@ -470,3 +516,196 @@ func TestLoginCommand_jwt(t *testing.T) {
...
+		canonicalRoleArn    = "arn:aws:iam::1234567890:role/my-role"
+		roleArnWithPath     = "arn:aws:iam::1234567890:role/some/path/my-role"
+		roleArnWildcard     = "arn:aws:iam::1234567890:role/some/path/*"
+		roleName            = "my-role"
+		userArn             = "arn:aws:sts::1234567890:user/my-user"
+		userArnWildcard     = "arn:aws:sts::1234567890:user/*"
+		userName            = "my-user"
+		entityId            = "AAAsameuniqueid"
+		entityIdWithSession = entityId + ":some-session"
+		accountId           = "1234567890"
```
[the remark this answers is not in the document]
pglass: 👍 A test fixture is a great idea, especially since I need this in a third place (to test the `awsauth` package)

command/login/aws.go
pglass: Agree, needs clarification. They are not required CLI flags.

The AWS SDK needs a region, but there are various standard ways to pass the region (and other config) to the SDK aside from the flags we support here ([env vars, file, and others](https://docs.aws.amazon.com/sdk-for-go/api/aws/session/)).

command/login/aws.go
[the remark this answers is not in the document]
pglass: I've dug through the credential loading a good bit in the past couple days. I time boxed this, but there's nothing I've found that Vault supports in its custom credential chain that the SDK doesn't also support out of the box.

Users have also noticed differences between Vault's credential chain and the SDK, and Vault has generally moved toward matching the SDK behavior where possible (such as: https://github.com/hashicorp/vault/pull/7738 and https://github.com/hashicorp/vault/issues/13807).

So I think we should rely on the SDK's credential discovery in favor of being more standardized.

command/login/aws.go
pglass: I clarified this a bit in the flag descriptions. I also fixed the logic here so that this always checks that the region and credentials were either passed or found by the SDK's credential chain.

(fwiw, the login attempt will fail if creds or region are missing, but the error message is more obtuse in that case)

agent/consul/authmethod/awsauth/aws.go
```diff
@@ -0,0 +1,191 @@
...
+	// in binding rules. Required if wildcard principal ARNs are used.
+	EnableIAMEntityDetails bool `json:",omitempty"`
+
+	// IAMEntityTags are the specific IAM User or IAM Role tags to include as selectable
+	// fields in the binding rule attributes. Requires EnableIAMEntityDetails = true.
+	IAMEntityTags []string `json:",omitempty"`
+
+	// ServerIDHeaderValue adds a X-Consul-IAM-ServerID header to each AWS API request.
+	// This helps protect against replay attacks.
+	ServerIDHeaderValue string `json:",omitempty"`
```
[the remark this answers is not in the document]
pglass: Yeah, I'll detail this in the docs on consul.io.

pglass:
Created the following couple of issues to follow up on before 1.12 GA:

* https://github.com/hashicorp/consul/issues/12661
* https://github.com/hashicorp/consul/issues/12662

internal/iamauth/auth.go
```diff
@@ -0,0 +1,311 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	callerIdentity, err := a.submitCallerIdentityRequest(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+	a.logger.Info("iamauth login attempt", "arn", callerIdentity.Arn)
```
rboyer, line 75: This should probably be `Debug`

rboyer (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/619124.

PR #12585, opened 2022-03-21
factor out recording func, add unit tests

FFMMM:
### Overview

PR adds package level testing for the `RequestRecorder`.  Credit to @dhiaayachi for some of these ideas here!

### DIFF
--- a/agent/rpc/middleware/interceptors.go
+++ b/agent/rpc/middleware/interceptors.go
@@ -18,16 +18,18 @@ import (
 // Really what we are measuring here is a "cluster operation". The term we have
 // used for this historically is "RPC", so we continue to use that here.
 const RPCTypeInternal = "internal"
+const RPCTypeNetRPC = "net/rpc"
 
 var metricRPCRequest = []string{"rpc", "server", "request"}
 var requestLogName = "rpc.server.request"
 
 type RequestRecorder struct {
-	Logger hclog.Logger
+	Logger       hclog.Logger
+	recorderFunc func(key []string, start time.Time, labels []metrics.Label)
 }
 
 func NewRequestRecorder(logger hclog.Logger) *RequestRecorder {
-	return &RequestRecorder{Logger: logger}
+	return &RequestRecorder{Logger: logger, recorderFunc: metrics.MeasureSinceWithLabels}
 }
 
 func (r *RequestRecorder) Record(requestName string, rpcType string, start time.Time, request interface{}, respErrored bool) {
@@ -42,9 +44,8 @@ func (r *RequestRecorder) Record(requestName string, rpcType string, start time.
 		{Name: "rpc_type", Value: rpcType},
 	}
 
-	// TODO(rpc-metrics-improv): consider using Telemetry API call here
-	// It'd be neat if we could actually pass the elapsed observed above
-	metrics.MeasureSinceWithLabels(metricRPCRequest, start, labels)
+	// TODO(FFMMM): it'd be neat if we could actually pass the elapsed observed above
+	r.recorderFunc(metricRPCRequest, start, labels)
 
 	r.Logger.Debug(requestLogName,
 		"method", requestName,
@@ -67,11 +68,6 @@ func GetNetRPCInterceptor(recorder *RequestRecorder) rpc.ServerServiceCallInterc
 
 		err := handler()
 
-		responseErr := false
-		if err != nil {
-			responseErr = true
-		}
-
-		recorder.Record(reqServiceMethod, "net/rpc", reqStart, argv.Interface(), responseErr)
+		recorder.Record(reqServiceMethod, RPCTypeNetRPC, reqStart, argv.Interface(), err != nil)
 	}
 }
--- a/agent/rpc/middleware/interceptors_test.go
+++ b/agent/rpc/middleware/interceptors_test.go
@@ -1,0 +1,126 @@
+package middleware
+
+import (
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/armon/go-metrics"
+	"github.com/hashicorp/go-hclog"
+	"github.com/stretchr/testify/require"
+)
+
+// obs holds all the things we want to assert on that we recorded correctly in our tests.
+type obs struct {
+	key    []string
+	start  time.Time
+	labels []metrics.Label
+}
+
+// recorderStore acts as an in-mem mock storage for all the RequestRecorder.Record() recorderFunc calls.
+type recorderStore struct {
+	lock  sync.Mutex
+	store map[string]obs
+}
+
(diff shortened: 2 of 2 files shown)


agent/rpc/middleware/interceptors.go
```diff
@@ -42,9 +44,8 @@ func (r *RequestRecorder) Record(requestName string, rpcType string, start time.
 		{Name: "rpc_type", Value: rpcType},
 	}
 
-	// TODO(rpc-metrics-improv): consider using Telemetry API call here
-	// It'd be neat if we could actually pass the elapsed observed above
-	metrics.MeasureSinceWithLabels(metricRPCRequest, start, labels)
+	// TODO(FFMMM): it'd be neat if we could actually pass the elapsed observed above
```
FFMMM, line 47: This would require a change in `go-metrics` to support an "elapsed" time concept when recording.

Since this is not particularly tied to the effort, I reclassified the todo

agent/rpc/middleware/interceptors_test.go
```diff
@@ -0,0 +1,126 @@
...
+func (rr readRequest) IsRead() bool {
+	return true
+}
+
+func (wr writeRequest) IsRead() bool {
+	return false
+}
+
+// TestRequestRecorder_SimpleOK tests that the RequestRecorder can record a simple request.
+func TestRequestRecorder_SimpleOK(t *testing.T) {
```
FFMMM, line 62: i can be persuaded to add some of these tests in a for loop but for now i thought it was easier to read like so.

agent/rpc/middleware/interceptors_test.go
```diff
@@ -0,0 +1,126 @@
...
+type recorderStore struct {
+	lock  sync.Mutex
+	store map[string]obs
+}
+
+func (rs *recorderStore) put(key []string, o obs) {
+	rs.lock.Lock()
+	defer rs.lock.Unlock()
+
+	actualKey := strings.Join(append(key, o.labels[0].Value), "")
```
FFMMM, line 31: the `key` here is always going to be the metric name, i.e. `rpc_server_call`. We need something else to use the map properly.

So we're making `rpc["" ]server[""]call[""]A.B" or whatever the method name is.

Curious if others have a better idea about this one.

dhiaayachi: what about making it a stack instead of a map so you can also check the expected order if you have a use case with multiple calls?

agent/rpc/middleware/interceptors_test.go
dhiaayachi: I agree those could be transformed to a table test, but I have no preference.

agent/rpc/middleware/interceptors.go
```diff
@@ -18,16 +18,18 @@ import (
...
 
 type RequestRecorder struct {
-	Logger hclog.Logger
+	Logger       hclog.Logger
+	recorderFunc func(key []string, start time.Time, labels []metrics.Label)
 }
 
 func NewRequestRecorder(logger hclog.Logger) *RequestRecorder {
-	return &RequestRecorder{Logger: logger}
+	return &RequestRecorder{Logger: logger, recorderFunc: metrics.MeasureSinceWithLabels}
```
dhiaayachi, line 32: Wondering if we should not inject this to the ctor to be able to use the `recorderStore` at a higher level (Server) and use the same technique of mocking to test at higher level too.

dhiaayachi (approved):
Thank you for working on this @FFMMM. I can't wait to see how it will transfer when testing at the server level.
I have some open questions but they are non blocking, LGTM!

agent/rpc/middleware/interceptors_test.go
FFMMM: Hm, open to doing that. But expected order should not matter 🤔 

I don't think ordering is guaranteed by the underlying implementation of `go-metrics`. I think it's just a side effect.

agent/rpc/middleware/interceptors.go
FFMMM: Oh I see -- we could pass it a "mock"/ fake recordFunc in some `agent_test` or `server_test` use cases 🤔 .

I think that way, couple with the basic testing in `metrics_test` should be enough to gain reasonable confidence and make sure there's no regressions.

agent/rpc/middleware/interceptors_test.go
dhiaayachi: I agree in this use case it's not useful. As long as the key is something unique enough but can be guessed it should be fine.

agent/rpc/middleware/interceptors_test.go
FFMMM: I'd be inclined to keep the tests like so for now. Thinking that if we add more logic in `RequestRecorder` in terms of parsing the request, then we can refactor and use table tests for brevity.

kisunji: > I'd be inclined to keep the tests like so for now. Thinking that if we add more logic in `RequestRecorder` in terms of parsing the request, then we can refactor and use table tests for brevity.

I like this approach! No need for table tests if there isn't a large combination of inputs

agent/rpc/middleware/interceptors_test.go
```diff
@@ -0,0 +1,126 @@
...
+	actualKey := strings.Join(key, "")
+	return rs.store[actualKey]
+}
+
+var store = recorderStore{store: make(map[string]obs)}
+var simpleRecorderFunc = func(key []string, start time.Time, labels []metrics.Label) {
+	o := obs{key: key, start: start, labels: labels}
+
+	store.put(key, o)
+}
```
kisunji, line 48: Can these be scoped to a test? I don't see a need for them to be global

agent/rpc/middleware/interceptors.go
kisunji: Doesn't `metrics.MeasureSinceWithLabels` measure elapsed? I thought that's what `start` was for

agent/rpc/middleware/interceptors_test.go
```diff
@@ -0,0 +1,126 @@
...
+	actualKey := strings.Join(append(key, o.labels[0].Value), "")
+	rs.store[actualKey] = o
+}
+
+func (rs *recorderStore) get(key []string) obs {
+	rs.lock.Lock()
+	defer rs.lock.Unlock()
+
+	actualKey := strings.Join(key, "")
+	return rs.store[actualKey]
```
kisunji, line 40: I think it looks weird that the `get` key is different from the `put` key.
I see you synthesizing the `get` key manually in tests but wonder if there's a cleaner symmetric solution.

kisunji (commented):
I just realized you merged as I was mid-review! None of my comments are blocking but worth thinking about as follow-ups!

PR #12589, opened 2022-03-22
xds: connect sidecar proxies can control min/max TLS versions and cipher suites from service-defaults now

### LINKED ISSUE #11966: allow configuring tls version and cipher for envoy connect
#### Feature Description

Envoy by default still allows TLS 1.0 for listeners, which has long been deprecated and is a potential security risk.
The minimum tls version, along with the supported ciphers, are configurable through the [tls_params](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#extensions-transport-sockets-tls-v3-tlsparameters) option, however those configurations are not exposed through consul.

Requesting to expose those configurations in consul so we can tighten up the security for envoy connect tls communication.

#### Use Case(s)

Tighten up the security for connect tls.

cc @pavanrangain

rboyer:
Fixes #11966

Per-service you can edit the `tls{}` block of `service-defaults` to control these things. You can also update the same in `proxy-defaults` to apply to all services.

TODO:

- [ ] website update for `service-defaults`
- [ ] website update for `proxy-defaults`
- [ ] test for service vs proxy defaults overriding
- [ ] consider changing the default for `TLSMinVersion` to 1.2 and adding a backwards compatibility node in the upgrade guide

### DIFF
--- a/agent/consul/config_endpoint.go
+++ b/agent/consul/config_endpoint.go
@@ -16,6 +16,7 @@ import (
 	"github.com/hashicorp/consul/agent/configentry"
 	"github.com/hashicorp/consul/agent/consul/state"
 	"github.com/hashicorp/consul/agent/structs"
+	"github.com/hashicorp/consul/types"
 )
 
 var ConfigSummaries = []prometheus.SummaryDefinition{
@@ -445,6 +446,15 @@ func (c *ConfigEntry) ResolveServiceConfig(args *structs.ServiceConfigRequest, r
 		return err
 	}
 
+	if args.NoUpstreams {
+		if len(args.UpstreamIDs) > 0 {
+			return fmt.Errorf("cannot specify both NoUpstreams and UpstreamIDs at once")
+		}
+		if len(args.Upstreams) > 0 {
+			return fmt.Errorf("cannot specify both NoUpstreams and Upstreams at once")
+		}
+	}
+
 	var (
 		priorHash uint64
 		ranOnce   bool
@@ -481,6 +491,7 @@ func (c *ConfigEntry) ResolveServiceConfig(args *structs.ServiceConfigRequest, r
 				&args.EnterpriseMeta,
 				upstreamIDs,
 				args.Mode,
+				args.NoUpstreams,
 			)
 			if err != nil {
 				return err
@@ -509,6 +520,7 @@ func (c *ConfigEntry) ResolveServiceConfig(args *structs.ServiceConfigRequest, r
 				args,
 				upstreamIDs,
 				legacyUpstreams,
+				args.NoUpstreams,
 				entries,
 			)
 			if err != nil {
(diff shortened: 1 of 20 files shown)


agent/structs/config_entry.go
```diff
@@ -613,6 +624,9 @@ type ServiceConfigRequest struct {
 	// Mode indicates how the requesting proxy's listeners are dialed
 	Mode ProxyMode
 
+	// NoUpstreams forces just data about the specified service to be returned and no related data.
+	NoUpstreams bool
```
rboyer, line 628: This bypasses all upstream related data collection, even in tproxy mode to save on resources.

agent/structs/config_entry.go
```diff
@@ -1125,6 +1142,60 @@ func validateConfigEntryMeta(meta map[string]string) error {
...
 }
 
+func validateProxyTLSConfig(cfg *ProxyTLSConfig) error {
+	if cfg == nil {
+		return nil
+	}
+	return validateTLSConfig(cfg.TLSMinVersion, cfg.TLSMaxVersion, cfg.CipherSuites)
+}
+
+func validateTLSConfig(
```
rboyer, line 1152: This is just pushed around from the existing code used by `ingress-gateway`.

agent/xds/listeners.go
```diff
@@ -1662,3 +1667,66 @@ func makeCommonTLSContextFromFiles(caFile, certFile, keyFile string) *envoy_tls_
 
 	return &ctx
 }
+
+func validateListenerTLSConfig(tlsMinVersion types.TLSVersion, cipherSuites []types.TLSCipherSuite) error {
```
rboyer, line 1671: Much of the rest of this file is just pushed around from the existing code used by `ingress-gateway`.

agent/proxycfg/connect_proxy.go
```diff
@@ -70,6 +70,18 @@ func (s *handlerConnectProxy) initialize(ctx context.Context) (ConfigSnapshot, e
 		return snap, err
 	}
 
+	// Get additional information about the service itself.
```
rboyer, line 73: All `connect-proxy` instances now fetch this resolved service config data.

agent/consul/config_endpoint.go
```diff
@@ -594,6 +615,21 @@ func (c *ConfigEntry) computeResolvedServiceConfig(
 			thisReply.Mode = serviceConf.Mode
 		}
 
+		if serviceConf.TLS != nil {
+			if thisReply.TLS == nil {
```
rboyer, line 619: `service-defaults` beats `proxy-defaults`, but note there's no way to easily get back to the passive (i.e. blank) default values for a specific service if these values are set for all proxies via `proxy-defaults` since `[]` or `""` in `service-defaults` will be treated as an omission and will continue to use the common default instead.

agent/proxycfg/snapshot.go
```diff
@@ -451,7 +455,8 @@ func (s *ConfigSnapshot) Valid() bool {
 		}
 		return s.Roots != nil &&
 			s.ConnectProxy.Leaf != nil &&
-			s.ConnectProxy.IntentionsSet
+			s.ConnectProxy.IntentionsSet &&
+			s.ConnectProxy.TLSConfigSet
```
mkeeler, line 459: Is the TLS config a required field before sending back resources so that we don't reconfigure the transport socket of a listener unnecessarily?

rboyer: Yes, so that the first time you try to do TLS you follow the user's intent and don't introduce a racy downgrade window.

rboyer:
Closing in favor of #12601

PR #12594, opened 2022-03-22
Unify various status errors into one HTTP error type.

Mathew-Estafanous:
issue #11393

At the moment, every http code has its own error type, which makes adding a new http code response more difficult than it should be.

This PR aims at unifying the various errors types (``BadRequestError``, ``ForbiddenError``) into one error type ``HTTPError``. This new type allows for much fewer error structs as well as the flexibility to returning new status codes.

**PR is not finished**


kisunji:
Hmm that's weird. I think on Github the `go-tests` is stuck on 'In progress' but CircleCI reports all tests as passing. 

Are `TestAgent_*` tests still failing locally?

Unfortunately it's a huge package so we hardly run tests on them locally; it could be OS or resource-specific errors that we never caught

Mathew-Estafanous:
Looks like ``TestAgent_Host`` is the only test that is consistently failing locally.
image

The other test cases passed when I ran them individually, which is a little weird.

kisunji:
You're all green on the CircleCI side! Mark me as reviewer when you want a review (or ping me if you want a sanity check that CI is passing 🤣 )

Mathew-Estafanous:
I see. Hopefully GitHub figure is out soon.

One thing before a review. I’m thinking of making a constructor for ``HTTPError`` that the handlers would use instead of initializing with a struct literal. My reasoning is that the wrapper makes the assumption that the status code and reason are filled out properly and the constructor would  ‘enforce’ this by requiring a status code and reason.

Constructor like this.
```go
func NewHTTPError(statusCode int, reason string) HTTPError {
	return HTTPError{statusCode, reason}
}
```

Then the handler would return the error like this.
```go
func (s *HTTPHandlers) SomeHandler(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
     // some code here...
    return nil, NewHTTPError(http.StatusBadRequest, “You didn’t format it right")
}
```

This might be overkill, but thought I’d suggest in case. 🤷🏽‍♂️ 

If what I suggested above shouldn’t be implemented then I’m ready for review!

Mathew-Estafanous:
Also, I don’t have permission to add reviewers so I’ll probably have to stick with pinging you.

Mathew-Estafanous:
@kisunji I pushed the change. Looks like everything passed, so I'm ready for a review.

agent/connect_auth.go
```diff
@@ -23,7 +24,7 @@ import (
 // The ACL token and the auth request are provided and the auth decision (true
 // means authorized) and reason string are returned.
 //
-// If the request input is invalid the error returned will be a BadRequestError,
+// If the request input is invalid the error returned will be ann BadRequest HTTPError,
```
kisunji, line 27:
```suggestion
// If the request input is invalid the error returned will be a BadRequest HTTPError,
```

agent/txn_endpoint_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/base64"
 	"fmt"
+	"github.com/stretchr/testify/require"
```
kisunji, line 7: nit-picky but can we move this import to the 2nd import group?

agent/txn_endpoint_test.go
```diff
@@ -63,11 +61,19 @@ func TestTxnEndpoint_Bad_Size_Item(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf("err: %v", err)
+			}
+		} else {
+			if err, ok := err.(HTTPError); ok {
+				if err.StatusCode != 413 {
+					t.Fatalf("expected 413 but got %d", err.StatusCode)
+				}
+			} else {
+				t.Fatalf("excected HTTP error but go %v", err)
```
kisunji, line 75:
```suggestion
				t.Fatalf("excected HTTP error but got %v", err)
```

agent/txn_endpoint_test.go
```diff
@@ -138,11 +144,19 @@ func TestTxnEndpoint_Bad_Size_Net(t *testing.T) {
...
```
kisunji, line 158:
```suggestion
				t.Fatalf("excected HTTP error but got %v", err)
```

kisunji (approved):
This is awesome! A few typo fixes then we can merge

kisunji:
Thanks for your contribution! It might take a bit of time to merge because we have a separate enterprise repo that I'll have to manually update

kisunji:
@Mathew-Estafanous Are you able to view the running jobs on your fork from CircleCI? It looks like this PR is stuck on a CI job but I can't tell where.

Mathew-Estafanous:
Hmm, no I don’t see it either. I only see that ``verify-ci`` has been blocked, but there is no details linking to more information about it.

Should I push an empty commit to try restarting things?

kisunji:
Yeah try it to see if it will unblock! Otherwise I can open another PR and cherrypick all your commits. Sometimes our CI behaves strangely with PRs from forks

kisunji:
Hmm I think CircleCI is being smart and refusing to rerun the test since there's no diff. I re-ran it manually but in any case I think this is fine merging as-is (since I'm confident it won't impact integration tests)

Mathew-Estafanous:
Sounds good! Also, for future issues should I ping you or should I just go about making the PR, etc and wait for others to review?

kisunji:
You are welcome to continue pinging me but I may take longer to respond depending on my workload. Our team meets weekly to review new or updated issues and PRs so you'll ideally get a response from someone within a week when you raise a PR or make a comment on an issue

kisunji:
Just updating to say I'm working on the enterprise side of things! I can fix the intentions_endpoint.go here as well

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/656224.

Issue #10471, opened 2021-06-23
remove gogo/protobuf direct dependency


rboyer:
The `github.com/gogo/protobuf` library is [now unmaintained](https://github.com/gogo/protobuf/issues/691) and no longer tracks with the latest major release of `github.com/golang/protobuf` (the 1.4.x series). The two will _sorta_ interoperate up to a point, and that point manifests as marshal/unmarshal `panic` which are very undesirable.

All of the protobufs and generated code under the `./proto` directory are generated with the extended options offered by gogo, with features like optionally embedding structs, using `time.Time` or `time.Duration` in generated code, and adding struct tags to fields.

Unfortunately as gogo/protobuf is effectively dead, we have to cease using all of the bells and whistles it provided before we can laterally go back to direct use of `golang/protobuf` (1.3.x) and after that migration can bump to the 1.4.x series (or jump to the new repo where v2 is maintained).

Some of these steps can happen in incremental commits, but unfortunately due to the cross-cutting nature of these changes many of these will have to be monolithic singular commits.

1. Merge improvements (https://github.com/hashicorp/consul/pull/10229) to the `mog` tool and move it to its own repo to make the tool easier to use, rather than having it hide in an unmerged branch of this repo.
2. Selectively go through various `./proto` files and remove gogo extensions, regenerate the code, and fix anything that is broken by those removals.
    1. Likely any place we use the generated protobuf structs and expect to marshal to/from json into those will need to stop, since we rely upon stuff like embedded structs (like the raft fields) and struct tags (for `bexpr` filtering or `hashstructure` stuff) to get the "shape" of the json to be what our API package expects. For this we can just use `mog` to transmogrify the protobuf structs into either `api` package structs or new hand-rolled structs in `agent/structs` as is appropriate.
    2. Fix how `time.Time` and `time.Duration` work. Noting that the protobuf time structs have less precision than the stdlib structs so various `auto-config` or `agent/connect` CA tests will fail due to this and need precision truncation to correct.
3. Once we aren't using gogo extensions for any protobufs, laterally moving over to the version of `golang/protobuf` should be straightforward.
    1. We **likely** do not want to upgrade the pinned version of `protoc` yet in this procedure until we're ready for 1.4.x.
    2. ...unless gRPC has compilation errors somewhere due to some back compat issues they had in which case try bumping `protoc` by very tiny  semver version increments until it isn't broken.
4. After there's no more gogo you can take a break and start tackling direct official protobuf version upgrades in another issue.

[...]

PR #12607, opened 2022-03-23
Fix leaked Vault LifetimeRenewers

clly:
When the Vault CA Provider is reconfigured we do not stop the
LifetimeRenewers which can cause them to leak until the Consul processes
recycles. On Configure execute stopWatcher if it exists and is not nil
before starting a new renewal

Also add jitter before restarting the LifetimeWatcher to prevent overwhelming Vault
when it's configured with rate limiting.

If we fail to login to Vault or our token is no longer valid we can
overwhelm a Vault instance with many requests very quickly by restarting
the LifetimeWatcher. Before restarting the LifetimeWatcher provide a
backoff time of 1 second or less.

### DIFF
--- a/agent/connect/ca/provider_vault.go
+++ b/agent/connect/ca/provider_vault.go
@@ -227,7 +227,7 @@ func (v *VaultProvider) renewToken(ctx context.Context, watcher *vaultapi.Lifeti
 				v.logger.Error("Error renewing token for Vault provider", "error", err)
 			}
 
-			// wait after getting an error or the lifetime watcher returns
+			// wait at least 1 second after returning from the lifetime watcher
 			retrier.Wait(ctx)
 
 			// If the watcher has exited and auth method is enabled,
--- a/agent/connect/ca/provider_vault.go
+++ b/agent/connect/ca/provider_vault.go
@@ -12,8 +12,8 @@ import (
 	"strings"
 	"time"
 
-	"github.com/hashicorp/consul/lib"
 	"github.com/hashicorp/consul/lib/decode"
+	"github.com/hashicorp/consul/lib/retry"
 	"github.com/hashicorp/go-hclog"
 	vaultapi "github.com/hashicorp/vault/api"
 	"github.com/mitchellh/mapstructure"
@@ -45,7 +45,9 @@ const (
 
 	defaultK8SServiceAccountTokenPath = "/var/run/secrets/kubernetes.io/serviceaccount/token"
 
-	restartJitter = 1 * time.Second
+	retryMin    = 1 * time.Second
+	retryMax    = 5 * time.Second
+	retryJitter = 20
 )
 
 var ErrBackendNotMounted = fmt.Errorf("backend not mounted")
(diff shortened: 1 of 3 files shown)


agent/connect/ca/provider_vault.go
```diff
@@ -173,6 +179,11 @@ func (v *VaultProvider) renewToken(ctx context.Context, watcher *vaultapi.Lifeti
 			return
 
 		case err := <-watcher.DoneCh():
+			// In the event we fail to login to Vault or our token is no longer valid we can overwhelm a Vault
+			// instance with rate limit configured. We would make these requests to Vault as fast as we possibly
+			// could and start causing all client's to receive 429 response codes. To mitigate that we're sleeping 1
+			// second or less before restarting the LifetimeWatcher. Once we can upgrade to
+			// github.com/hashicorp/vault/api@v1.1.1 or later the LifetimeWatcher _should_ perform that backoff for us.
```
banks, line 186: What was the change on Vault side? I see backoff in their code from over a year ago.

If we upgrade to 1.1.1, do we still need to sleep here? If so we could maybe use an exponential backoff too - it's still a bit gross to be making on average 2 failing auth calls per second forever if we have a bad token. I'd probably use our `lib.Retry` for it in `/lib`.

Finally, is there something stopping us from upgrading to api@1.1.1 now?

banks (commented):
Thanks @clly this looks pretty good and may be OK to merge as-is. I had a few questions inline though.

Mostly trying to work out if the hot loop was caused by Vault's API code (so the real fix is to upgrade) or if Vault's API code is already doing backoff internally and the hot loop we saw was actually caused by this code because the lifetime watcher is failing fast but we are immediately restarting it?

If it's the latter then I think we should maybe be more aggressive with backoff here as it's not just an edgecase but a real error that can happen easily!

agent/connect/ca/provider_vault.go
clly: The change of the Vault side _was_ the addition of the exponential backoff from over a year ago. If we upgrade to 1.1.1 we do not need to sleep here because the watcher should get an error, retry with the exponential backoff, then exit. The Vault code is currently returning attempting to renew, returning quickly and we are immediately restarting it causing another renew attempt.

Upgrading to api@1.1.1 upgrades some other dependencies including protobuf, the aws-sdk, and genproto. I'm not super familiar with what those would have affect or what our normal process is for upgrading our dependencies so I put the sleep here instead. The exponential retry in the Vault code would cap out at 1 second so we'd still have 1 failing call per second forever (or until consul is reconfigured).

I can switch to lib.Retry fairly easily or upgrade to the newer vault/api version.

banks: Hmm yeah the protobuf thing probably means we can't until we are done moving off gogo but I'll defer to the Core team to work out if that's true!

I think this is OK for a temporary solution but maybe we could:
 - make it `lib.RandomJitter(4 *time.Second) + time.Second` so it's 1-5 seconds rather than 0-1. (or lib.Retry which is not too hard?) either seems fine
 - Add `TODO` comments here and maybe link this PR in the issue we are tracking protobuf upgrades so we don't forget to come back and upgrade?

I think it's worth trying to merge and ship this fix as soon as possible as it could affect production Vault stability.

@hashicorp/consul-core, what do you think?

clly: I've switched from RandomJitter to lib.Retry. I've also created a separate issue to upgrade to a newer version of the vault/api and linked it here and to https://github.com/hashicorp/consul/issues/10471

agent/connect/ca/provider_vault.go
```diff
@@ -232,11 +257,12 @@ func (v *VaultProvider) renewToken(ctx context.Context, watcher *vaultapi.Lifeti
 					continue
 				}
 			}
-
 			// Restart the watcher.
+
 			go watcher.Start()
 
 		case <-watcher.RenewCh():
+			retrier.Reset()
```
banks, line 265: Does this get called after every successful request to vault from the watcher? For example does it fire after the initial token fetch/validation? If not, we might need to also reset above when we re-start the watcher?

agent/connect/ca/provider_vault_test.go
```diff
@@ -212,6 +212,52 @@ func TestVaultCAProvider_RenewToken(t *testing.T) {
...
+	require.NoError(t, err)
+	providerToken := secret.Auth.ClientToken
+
+	provider, err := createVaultProvider(t, true, testVault.Addr, providerToken, nil)
+	require.NoError(t, err)
+
+	var gotStopped = false
+	provider.stopWatcher = func() {
+		gotStopped = true
+	}
```
banks, line 238: Is this callback will called from another goroutine? I think it is, which probably means this would fail the race detector (even though bool assign is atomic on some CPUs anyway).

Could replace it with:
```suggestion
	var gotStopped = uint32(0)
	provider.stopWatcher = func() {
		atomic.StoreUint32(&gotStopped, 1)
	}
```

then later `require.Equal(t, uint32(1),  atomic.LoadUint32(&gotStopped))`

banks (approved):
A couple of minor things but LGTM. Would be great if someone from Core could review too!

agent/connect/ca/provider_vault.go
clly: This gets called on every successful renewal which should happen immediately after start and then the LifetimeWatcher sleeps.

agent/connect/ca/provider_vault_test.go
clly: It shouldn't be called as part of a goroutine (at least in this test) but is a call back from `provider.Configure.` Using atomic should make it more robust if that changes in the future.

kyhavlov (approved):
LGTM 👍

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/615051.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 922619dfc33572c4f6076fae7e7dce0ec5fe450c onto `release/1.11.x` succeeded!

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 922619dfc33572c4f6076fae7e7dce0ec5fe450c onto `release/1.10.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/615051)

PR #12586, opened 2022-03-22
[OSS] Restructure gRPC server setup

boxofrad:
OSS sync of enterprise changes at 0b44395e

PR #12611, opened 2022-03-24
fix bad oss sync, use gauges not counters

FFMMM:
We made those changes to not expire prom metrics in here: https://github.com/hashicorp/consul/pull/12582.

There was a small mix up with this OSS sync: https://github.com/hashicorp/consul/pull/12586/files#diff-92974f7ecae0d7201c3bd5ad4a8e8b0e49a66e40ee72d1661644fbc3cee199c1L28-L268 that removed the prom gauge pre declaration.

Adding it back and actually using a gauge!

### DIFF
--- a/agent/metrics_test.go
+++ b/agent/metrics_test.go
@@ -94,30 +94,6 @@ func TestAgent_NewRPCMetrics(t *testing.T) {
 
 		assertMetricExists(t, respRec, metricsPrefix+"_rpc_server_call")
 	})
-
-	t.Run("Check that new rpc metrics can be filtered out", func(t *testing.T) {
-		metricsPrefix := "new_rpc_metrics_2"
-		hcl := fmt.Sprintf(`
-		telemetry = {
-			prometheus_retention_time = "5s"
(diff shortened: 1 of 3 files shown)


kisunji (approved):

agent/metrics_test.go
```diff
@@ -94,30 +94,6 @@ func TestAgent_NewRPCMetrics(t *testing.T) {
 
 		assertMetricExists(t, respRec, metricsPrefix+"_rpc_server_call")
 	})
-
-	t.Run("Check that new rpc metrics can be filtered out", func(t *testing.T) {
```
FFMMM, line 98: removing this since we are changing the default behavior for this new feature to be opt in by default (not opt out). I have a separate diff / patch that addresses the change but want a clean working tree in `main` in the meantime 💯

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/613552.

PR #12615, opened 2022-03-25
xds: Add SAN validation for terminating gateways

kyhavlov:
Fix for https://github.com/hashicorp/security-vulnerabilities/issues/207 - adds SAN validation to terminating gateway upstream definitions.

### DIFF
--- a/.changelog/12583.txt
+++ b/.changelog/12583.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+acl: Added an AWS IAM auth method that allows authenticating to Consul using AWS IAM identities
+```
--- a/agent/consul/acl_authmethod.go
+++ b/agent/consul/acl_authmethod.go
@@ -8,6 +8,7 @@ import (
 	"github.com/hashicorp/go-bexpr"
 
 	// register these as a builtin auth method
+	_ "github.com/hashicorp/consul/agent/consul/authmethod/awsauth"
 	_ "github.com/hashicorp/consul/agent/consul/authmethod/kubeauth"
 	_ "github.com/hashicorp/consul/agent/consul/authmethod/ssoauth"
 )
--- a/agent/consul/authmethod/awsauth/aws.go
+++ b/agent/consul/authmethod/awsauth/aws.go
@@ -1,0 +1,193 @@
+package awsauth
+
+import (
+	"context"
+	"fmt"
+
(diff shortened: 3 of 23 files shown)


agent/xds/clusters.go
```diff
@@ -385,6 +385,16 @@ func (s *ResourceGenerator) injectGatewayServiceAddons(cfgSnap *proxycfg.ConfigS
 			if mapping.SNI != "" {
 				tlsContext.Sni = mapping.SNI
 			}
+			err := injectSANMatcher(tlsContext.CommonTlsContext, connect.SpiffeIDService{
```
freddygv, line 388: I think that this should instead be using the SNI when available. The issue with using a consul SPIFFE ID is that terminating gateways route to external services, to which we don't provision certificates. 

Since specifying the SNI in a terminating gateway config entry is optional, there are a couple more things we could do:
* Add a strong recommendation to the config entry docs that the SNI should be set when using TLS.
* Log a warning on config write, when any of: CAFile, CertFile, or KeyFile are specified in the config entry, but not SNI.

kyhavlov: That makes sense - I guess it would be strange to expect people to fetch connect certs for their external service. So all we can really do here is update the docs and maybe add that log message when creating/updating a terminating gateway config entry.

msiege2: @picatz Are you on board with just a doc change for this item?

freddygv: It wouldn't just be a docs change, we would also start validating SANs when an SNI is provided in the terminating-gateway config entry.

The docs change is about recommending that people provide an SNI, so that we can do that upstream validation

hc-github-team-consul-core:
:repeat: gotestsum re-ran some tests in https://circleci.com/gh/hashicorp/consul/619697

```
github.com/hashicorp/consul/agent/consul.TestPreparedQuery_Wrapper: 2 runs, 1 failures
github.com/hashicorp/consul/agent/consul.TestServer_Enterprise_JoinLAN_then_Leave_PartitionsAndSegments: 2 runs, 1 failures
github.com/hashicorp/consul/agent/consul.TestServer_Enterprise_JoinLAN_then_Leave_PartitionsAndSegments/make_sure_the_serf_joins_have_finished: 2 runs, 1 failures
```

PR #12617, opened 2022-03-25
Enable running autopilot state updates on all servers

mkeeler:
Previously we started autopilot when a server gained leadership and stopped it when a server lost leadership. For some upcoming features we need autopilot on all servers to continually track the state of all servers. This PR pulls in a raft-autopilot update and enables that functionality.

When a Consul server is started, autopilot will be started but with reconciliation disabled to prevent it from attempting raft config modifications. Upon gaining leadership we will tell the running autopilot to enable reconciliation and then to disable reconciliation once the server loses leadership.
 
There are a few related changes in this PR as well.

1. The `autopilot.healthy` and `autopilot.failure_tolerance` metrics will now be emitted regularly on both followers and the leader. There is now a "leader" label added to the metrics to make it simpler to pick out the leaders view for the purposes of alerting.

2. The `Operator.AutopilotState` and `Operator.ServerHealth` RPCs no longer forcefully forward to the leader. Our HTTP server and CLI will default to non-stale queries so the overall behavior should be unchanged for existing uses. The exception is if the `-stale` CLI parameter or corresponding query parameter are being used then it will cause those RPCs to not be forward to the leader and instead return the follower servers view of the state.

Lastly the second commit in this PR is unrelated. Its just fixing a linter warning that popped up when I was working on this. 

TODOS:
- [x] Pull in official version of raft-autopilot once the upstream PR is merged and a release is made: https://github.com/hashicorp/raft-autopilot/pull/16


boxofrad (approved):
LGTM 💯

agent/consul/rpc_test.go
```diff
@@ -817,7 +817,8 @@ func TestRPC_RPCMaxConnsPerClient(t *testing.T) {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			dir1, s1 := testServerWithConfig(t, func(c *Config) {
-				c.RPCMaxConnsPerClient = 2
+				// we have to set this to 3 because autopilot is going to keep a connection open
+				c.RPCMaxConnsPerClient = 3
```
mkeeler, line 821: Previously this test executing successfully was relying on having all the connections getting opened prior to autopilot being started. Once leadership was established and autopilot is started, a conn will be kept open taking up one of the available slots for connections.

The changes here up the limit to 3 and also wait for establishing leadership further down. Just introducing the leader establishment wait in the previous code is enough to cause the test to reliably fail.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/629100.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit a553982506f27d4e4d4c66510f285c9c62332d61 onto `release/1.12.x` succeeded!

PR #12624, opened 2022-03-25
polish rpc.service.call metric behavior

FFMMM:
### overview

This diff polishes the behavior for the new rpc metric:

* it turns off the metric by default per conversations with product; we'll advertise this under beta as we roll out 1.12
  * refactors some builder code to prove that the `prefix_filter` serves all of the use cases
* i had to use a `SetGauge...` func bc the `MeasureSince...` does not actually work as a gauge 😢 ; prolly need to flag this into `go-metrics`
* renaming "new rpc" to "one twelve" rpc per feedback form matt s

### reviewer note

* best seen by commits! check out some of my inline comments 

### todo:

* [ ] fix full config and flag tests ⌨️ 🔨

### DIFF
--- a/agent/config/builder.go
+++ b/agent/config/builder.go
@@ -2581,7 +2581,7 @@ func (b *builder) parsePrefixFilter(telemetry *Telemetry) ([]string, []string) {
 
 	// TODO(FFMMM): Once one twelve style RPC metrics get out of Beta, don't remove them by default.
 	operatorPassedOneTwelveRPCMetric := false
-	oneTwelveRPCMetric := *telemetry.MetricsPrefix + "." + strings.Join(middleware.NewRPCGauges[0].Name, ".")
+	oneTwelveRPCMetric := *telemetry.MetricsPrefix + "." + strings.Join(middleware.OneTwelveRPCGauges[0].Name, ".")
 
 	for _, rule := range telemetry.PrefixFilter {
 		if rule == "" {
--- a/agent/config/builder.go
+++ b/agent/config/builder.go
@@ -641,32 +641,7 @@ func (b *builder) build() (rt RuntimeConfig, err error) {
 	}
 
 	// Parse the metric filters
-	var telemetryAllowedPrefixes, telemetryBlockedPrefixes []string
-
-	operatorEnableOneTwelveRPCMetric := false
-	oneTwelveRPCMetric := *c.Telemetry.MetricsPrefix + "." + strings.Join(middleware.NewRPCGauges[0].Name, ".")
-
-	for _, rule := range c.Telemetry.PrefixFilter {
-		if rule == "" {
-			b.warn("Cannot have empty filter rule in prefix_filter")
-			continue
-		}
-		switch rule[0] {
-		case '+':
-			if rule[1:] == oneTwelveRPCMetric {
-				operatorEnableOneTwelveRPCMetric = true
-			}
-			telemetryAllowedPrefixes = append(telemetryAllowedPrefixes, rule[1:])
-		case '-':
-			telemetryBlockedPrefixes = append(telemetryBlockedPrefixes, rule[1:])
-		default:
-			b.warn("Filter rule must begin with either '+' or '-': %q", rule)
-		}
-	}
-
-	if !operatorEnableOneTwelveRPCMetric {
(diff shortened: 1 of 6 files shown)


agent/config/builder.go
```diff
@@ -2588,3 +2575,38 @@ func (b *builder) buildTLSConfig(rt RuntimeConfig, t TLS) (tlsutil.Config, error
 
 	return c, nil
 }
+
+func (b *builder) parsePrefixFilter(telemetry *Telemetry) ([]string, []string) {
+	var telemetryAllowedPrefixes, telemetryBlockedPrefixes []string
+
+	// TODO(FFMMM): Once one twelve style RPC metrics get out of Beta, don't remove them by default.
+	operatorPassedOneTwelveRPCMetric := false
+	oneTwelveRPCMetric := *telemetry.MetricsPrefix + "." + strings.Join(middleware.OneTwelveRPCGauges[0].Name, ".")
```
mkeeler, line 2584: Should this end with a `.` suffix?

FFMMM: hey matt, I don't think I understand this. 

`oneTwelveRPCMetric` will look something like

```
oneTwelveRPCMetric = "consul" + "." + "rpc.server.call" = "consul.rpc.server.call"
```

> It looks as though there is one metric that can have many labels.

Yeah, for 1.12, there will be only `consul.rpc.server.call` with labels {`method`, `errored`, `request_type`, `rpc_type` }

agent/rpc/middleware/interceptors.go
```diff
@@ -52,9 +52,8 @@ func (r *RequestRecorder) Record(requestName string, rpcType string, start time.
 		{Name: "rpc_type", Value: rpcType},
 	}
 
-	// TODO(FFMMM): it'd be neat if we could actually pass the elapsed observed above
-	r.recorderFunc(metricRPCRequest, start, labels)
-
+	// math.MaxInt64 < math.MaxFloat32 is true so we should be good!
+	r.recorderFunc(metricRPCRequest, float32(elapsed), labels)
```
FFMMM, line 56: please keep me honest here

agent/rpc/middleware/interceptors.go
```diff
@@ -22,27 +23,26 @@ const RPCTypeInternal = "internal"
...
 
 type RequestRecorder struct {
 	Logger       hclog.Logger
-	recorderFunc func(key []string, start time.Time, labels []metrics.Label)
+	recorderFunc func(key []string, val float32, labels []metrics.Label)
 }
 
 func NewRequestRecorder(logger hclog.Logger) *RequestRecorder {
-	return &RequestRecorder{Logger: logger, recorderFunc: metrics.MeasureSinceWithLabels}
+	return &RequestRecorder{Logger: logger, recorderFunc: metrics.AddSampleWithLabels}
```
FFMMM, line 41: `MeasureSinceWithLabels` does cycle work which can take additional time. We have the `elaspsed` time in hand can pass it in as a value instead to avoid any additional error margins.

agent/rpc/middleware/interceptors.go
```diff
@@ -22,27 +23,26 @@ const RPCTypeInternal = "internal"
 const RPCTypeNetRPC = "net/rpc"
 
 var metricRPCRequest = []string{"rpc", "server", "call"}
-var requestLogName = "rpc.server.request"
+var requestLogName = strings.Join(metricRPCRequest, "_")
```
FFMMM, line 26: `logging/named.go` seems to encourage `_` in our loggers names.

mkeeler: 👍

agent/rpc/middleware/interceptors.go
mkeeler, line 56:
```suggestion
	r.recorderFunc(metricRPCRequest, float32(elapsed.Milliseconds()), labels)
```

You are right about the maxes. However most of our timing related metrics are in MS so we should convert the nanosecond duration to milliseconds here. go-metrics timings force ms precision.

mkeeler (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/619434.

PR #12626, opened 2022-03-25
Emit Ready Server Events

mkeeler:
As autopilot updates its state (which happens every 10s), it notifies Consul of the updated state.

This PR modifies the behavior to also generate a new `Event` from the `autopilot.State` which includes a slice of the servers which are considered "ready" for use. Only healthy servers, or previous voters who are still alive according to Serf are output in the event. As this event generation can happen every 10s, its likely that the set of ready servers has remained stable. If that is the case then we will prevent publishing an event and waking other servers.


agent/consul/autopilot.go
```diff
@@ -62,8 +97,117 @@ func (d *AutopilotDelegate) RemoveFailedServer(srv *autopilot.Server) {
...
+// newReadyServersEvent will create a stream.Event with the provided ready server info.
+func newReadyServersEvent(servers []ReadyServerInfo) stream.Event {
+	return stream.Event{
+		Topic:   EventTopicReadyServers,
+		Index:   uint64(time.Now().UnixMicro()),
+		Payload: EventPayloadReadyServers{Servers: servers},
+	}
+}
+
+// readyServersSnapshotHandler
```
boxofrad, line 196: nit: Remember to either fill this comment out or delete it 😉

boxofrad (approved):
This is awesome, great work! 👏🏻

boxofrad (approved):

agent/consul/autopilotevents/ready_servers_events.go
```diff
@@ -0,0 +1,305 @@
...
+		// if the event publisher were to not be able to keep up with procesing events
+		// then its possible this blocks. It could cause autopilot to not update its
+		// state as often as it should. However if this blocks for over 10s then
+		// not updating the autopilot state as quickly is likely the least of our
+		// concerns. If we need to make this async then we probably need to single
+		// flight these to ensure proper event ordering.
+		r.snapshotLock.Lock()
+		r.snapshot = events
+		r.snapshotLock.Unlock()
+		r.Publisher.Publish(events)
```
boxofrad, line 142: very tiny nit: it may help readability a bit to put this comment on the `Publish` call rather than the lock-assign-unlock block.

```suggestion
		r.snapshotLock.Lock()
		r.snapshot = events
		r.snapshotLock.Unlock()

		// if the event publisher were to not be able to keep up with procesing events
		// then its possible this blocks. It could cause autopilot to not update its
		// state as often as it should. However if this blocks for over 10s then
		// not updating the autopilot state as quickly is likely the least of our
		// concerns. If we need to make this async then we probably need to single
		// flight these to ensure proper event ordering.
		r.Publisher.Publish(events)
```

agent/consul/autopilotevents/ready_servers_events.go
```diff
@@ -0,0 +1,305 @@
...
+	// servers to see if there are differences. If the number of servers has changed
+	// we know that an event should be generated and sent.
+	if len(r.previous) == len(servers) {
+		diff := false
+		// We are relying on the fact that both of the slices will be sorted and that
+		// we don't care what the actual differences are but instead just that they
+		// have differences. We can then just iterate through all elements of both
+		// slices at the same time and check the elements for equality. Also,
+		// this is taking advantage of Go being able to do equality checking on
+		// structs without having to check all the fields separately. If we
```
boxofrad, line 169: Is this comment still correct? It looks like we're comparing the fields separately now.

agent/consul/autopilotevents/ready_servers_events.go
```diff
@@ -0,0 +1,305 @@
...
+	// we will attempt to lookup the node for the autopilot server. We use this to get the
+	// tagged addresses so that consumers of these events will be able to distinguish LAN
+	// vs WAN addresses as well as IP protocol differentiation. At first I thought we may
+	// need to hook into catalog events so that if the tagged addresses change then
+	// we can synthesize new events. That would be pretty complex so this code does not
+	// deal with that. The reasoning why that is probably okay is that autopilot will
+	// send us the state at least once every 30s. That means that we will grab the nodes
+	// from the catalog at that often and publish the events. So while its not quite
+	// as responsive as actually watching for the Catalog changes, its MUCH simpler to
+	// code and reason about and having those addresses be updated within 30s is good enough.
```
boxofrad, line 248: Good call 👏🏻

boxofrad (approved):
Awesome work! I've left a couple of small (mostly readability) comments. Please feel free to take or leave!

agent/consul/autopilotevents/ready_servers_events.go
mkeeler: It is no longer accurate. I have removed that part of the comment.

agent/consul/autopilotevents/ready_servers_events.go
mkeeler: I moved things around as suggested.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/643728.

PR #12627, opened 2022-03-25
Fix logic for website checker

lkysow:
Workflow should run when no docs/cherry-pick label && no pr/docs-label

### DIFF
--- a/.github/workflows/website-checker.yml
+++ b/.github/workflows/website-checker.yml
@@ -23,7 +23,7 @@ jobs:
   website-check:
     # If there's already a `type/docs-cherrypick` label or an explicit `pr/no-docs` label, we ignore this check 
     if: >-
-      !contains(github.event.pull_request.labels.*.name, 'type/docs-cherrypick') ||
+      !contains(github.event.pull_request.labels.*.name, 'type/docs-cherrypick') &&
       !contains(github.event.pull_request.labels.*.name, 'pr/no-docs')
     runs-on: ubuntu-latest
 


.github/workflows/website-checker.yml
```diff
@@ -23,7 +23,7 @@ jobs:
   website-check:
     # If there's already a `type/docs-cherrypick` label or an explicit `pr/no-docs` label, we ignore this check
```
FFMMM, line 24:
```suggestion
    # If there's already a `type/docs-cherrypick` and a `pr/no-docs` label, we ignore this check 
```

lkysow: I think that's not accurate though?

There are three cases of labels:
1. no matching labels => should run workflow
2. type/docs-cherry-pick => shouldn't run workflow
3. pr/no-docs => shouldn't run workflow


The conditional will evaluate to:

1. !false && !false => true && true => true
2. !true && !false => false && true => false
3. !false && !true => true && false => false

So it's accurate to say if there's a docs-cherry-pick *or* a pr-no-docs we ignore the check

eculver: The intention was to disable the checker in the presence of _either_ label so I think we should keep the comment as is. Otherwise, it would suggest that _both_ labels are required to disable the checker.

eculver (approved):
The dyslexia definitely played a part in this one. This makes my brain hurt. Thanks for fixing this @lkysow.

.github/workflows/website-checker.yml
lkysow: 👍 I'll merge it but if I'm totally wrong I'm happy to edit that comment.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/614610.

Protek5150iworld (commented):
Connect Lequariusjackson0920@protek.pr WIKI R

PR #12633, opened 2022-03-28
proto-gen-rpc-glue: fix behavior of renamed fields

rboyer:


internal/tools/proto-gen-rpc-glue/e2e/source.pb.go
```diff
@@ -0,0 +1,32 @@
...
+type ExampleReadRequest struct {
+	Value            string
+	ReadRequest      *pbcommon.ReadRequest
+	TargetDatacenter *pbcommon.TargetDatacenter
+}
+
+// @consul-rpc-glue: WriteRequest=AltWriteRequest
+type AltExampleWriteRequest struct {
+	Value           int
+	AltWriteRequest *pbcommon.WriteRequest
```
rboyer, line 25: I added this sort of behavior but hadn't used it yet (naming the field something other than the type name).

PR #12639, opened 2022-03-28
Remove the Rest of Gogo

erichaberkorn:


.circleci/config.yml
```diff
@@ -235,11 +235,9 @@ jobs:
...
             rm protoc-*.zip
       - run:
-          name: Install gogo/protobuf
+          name: Install protobuf
           command: |
-            gogo_version=$(go list -m github.com/gogo/protobuf | awk '{print $2}')
             go install -v github.com/hashicorp/protoc-gen-go-binary@master
-            go install -v github.com/gogo/protobuf/protoc-gen-gofast@${gogo_version}
             go install -v github.com/favadi/protoc-go-inject-tag@v1.3.0
             go install -v github.com/golang/protobuf/protoc-gen-go@v1.3.5
```
rboyer, line 242: I think actually you should switch this to a grep of `go.mod` for `golang/protobuf` so we install the correct lockstep generator

GNUmakefile
```diff
@@ -2,13 +2,11 @@
 # https://www.consul.io/docs/install#compiling-from-source
 
 SHELL = bash
-GOGOVERSION?=$(shell grep github.com/gogo/protobuf go.mod | awk '{print $$2}')
 GOTOOLS = \
 	github.com/elazarl/go-bindata-assetfs/go-bindata-assetfs@master \
```
rboyer, line 6: This section is missing the `github.com/golang/protobuf/protoc-gen-go@v1.3.5` install (and also should get version intepolation as above).

build-support/scripts/proto-gen.sh
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
```
rboyer, line 1: did you mean to pick one of either proto-gen-entry or proto-gen and retain a single file?

erichaberkorn: Oops yeah. I meant to remove proto-gen-entry 👍

PR #12651, opened 2022-03-29
build: auto install correct version of protoc locally

rboyer:


GNUmakefile
```diff
@@ -353,22 +358,25 @@ else
...
+.PHONY: protoc-install
+protoc-install:
+	$(info locally installing protocol buffer compiler version if needed (expect: $(PROTOC_VERSION)))
+	@if [[ ! -x $(PROTOC_ROOT)/bin/protoc ]]; then \
+		mkdir -p .protobuf/tmp ; \
+		if [[ ! -f .protobuf/tmp/$(PROTOC_ZIP) ]]; then \
+			( cd .protobuf/tmp && wget $(PROTOC_URL) ) ; \
+		fi ; \
+		mkdir -p $(PROTOC_ROOT) ; \
+		unzip -d $(PROTOC_ROOT) .protobuf/tmp/$(PROTOC_ZIP) ; \
```
kyhavlov, line 370: Won't this fail in the case where we have to download `protoc` because it'll have `cd`'d into `.protobuf/tmp` above on 367?

```suggestion
		unzip -d $(PROTOC_ROOT) $(PROTOC_ZIP) ; \
```

rboyer: Nope. The `cd` happens in a subshell (the parentheses) so after the closing parenthesis the working directory is back to the original.

erichaberkorn (approved):
I just tried this out and it works as expected 👍

PR #12665, opened 2022-03-30
build: run mog separately after the protobufs are generated

rboyer:
Also ensure that we run mog serially on each package in dependency order.

Depends on https://github.com/hashicorp/mog/pull/9 landing first.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -41,6 +41,7 @@ GOLDFLAGS=-X $(GIT_IMPORT).GitCommit=$(GIT_COMMIT)$(GIT_DIRTY)
 PROTOFILES?=$(shell find . -name '*.proto' | grep -v 'vendor/' | grep -v '.protobuf' )
 PROTOGOFILES=$(PROTOFILES:.proto=.pb.go)
 PROTOGOBINFILES=$(PROTOFILES:.proto=.pb.binary.go)
+PROTO_MOG_ORDER=$(shell go list -tags '$(GOTAGS)' -deps ./proto/pb... | grep "consul/proto")
 
 ifeq ($(FORCE_REBUILD),1)
 NOCACHE=--no-cache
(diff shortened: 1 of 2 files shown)


GNUmakefile
```diff
@@ -372,7 +373,21 @@ protoc-install:
 		chmod +x $(PROTOC_ROOT)/bin/protoc ; \
 	fi
 
-proto: protoc-install $(PROTOGOFILES) $(PROTOGOBINFILES)
+.PHONY: proto
+proto: -protoc-files mog
+
+.PHONY: mog
+mog:
```
rboyer, line 380: The `proto` target operates on files, but the `mog` target operates on packages which meant that the `pbservice` build was running mog extra times.

GNUmakefile
```diff
@@ -372,7 +373,21 @@ protoc-install:
...
+	@for FULL_PKG in $(PROTO_MOG_ORDER); do \
+		PKG="$${FULL_PKG/#github.com\/hashicorp\/consul\//.\/}" ; \
+		find "$$PKG" -name '*.gen.go' -delete ; \
+		echo "mog -tags '$(GOTAGS)' -source \"$${PKG}/*.pb.go\"" ; \
+		mog -tags '$(GOTAGS)' -source "$${PKG}/*.pb.go" ; \
```
rboyer, line 385: `-ignore-package-load-errors` is not required now with the change in https://github.com/hashicorp/mog/pull/9. It uses this glob syntax to selectively load files in the package that only match the glob, and avoids loading the bonus code that we have living in those packages.

Also ensuring that we execute mog in dependency graph order cuts down on this further.

GNUmakefile
```diff
@@ -41,6 +41,7 @@ GOLDFLAGS=-X $(GIT_IMPORT).GitCommit=$(GIT_COMMIT)$(GIT_DIRTY)
 PROTOFILES?=$(shell find . -name '*.proto' | grep -v 'vendor/' | grep -v '.protobuf' )
 PROTOGOFILES=$(PROTOFILES:.proto=.pb.go)
 PROTOGOBINFILES=$(PROTOFILES:.proto=.pb.binary.go)
+PROTO_MOG_ORDER=$(shell go list -tags '$(GOTAGS)' -deps ./proto/pb... | grep "consul/proto")
```
dhiaayachi, line 44: This is good!! The reason I did it with proto the first time was because I did not know how to get the right order.

GNUmakefile
dhiaayachi: can we make mog dependent on proto here, so when we run mog target it proto first or does that mean that proto will regenerate the files every time?

dhiaayachi (approved):
I had a question about dependencies but it's not blocking. LGTM!!

GNUmakefile
```diff
@@ -22,7 +22,7 @@ GOPROTOTOOLS = \
 	github.com/golang/protobuf/protoc-gen-go@$(GOPROTOVERSION) \
 	github.com/hashicorp/protoc-gen-go-binary@master \
 	github.com/favadi/protoc-go-inject-tag@v1.3.0 \
-	github.com/hashicorp/mog@v0.1.2
+	github.com/hashicorp/mog@glob-match
```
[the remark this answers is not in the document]
rboyer: v0.2.0 now

PR #12673, opened 2022-03-31
Move to using a shared EventPublisher

mkeeler:
Previously we had 1 EventPublisher per state.Store. When a state store was closed/abandoned such as during a consul snapshot restore, this had the behavior of force closing subscriptions for that topic and evicting event snapshots from the cache.

The intention of this commit is to keep all that behavior. To that end, the shared EventPublisher now supports the ability to refresh a topic. That will perform the force close + eviction. The FSM upon abandoning the previous state.Store will call RefreshTopic for all the topics with events generated by the state.Store.


agent/consul/server.go
```diff
@@ -378,6 +382,9 @@ func NewServer(config *Config, flat Deps, publicGRPCServer *grpc.Server) (*Serve
 	loggers := newLoggerStore(serverLogger)
 
 	recorder := middleware.NewRequestRecorder(serverLogger)
+
+	eventPublisher := stream.NewEventPublisher(10 * time.Second)
```
mkeeler, line 386: If we need an event publisher outside of the Server in the future then we could move this further out into the Deps struct.

agent/consul/server.go
```diff
@@ -341,6 +342,9 @@ type Server struct {
 	// Manager to handle starting/stopping go routines when establishing/revoking raft leadership
 	leaderRoutineManager *routine.Manager
 
+	// publisher is the precreated publisher
+	publisher *stream.EventPublisher
```
boxofrad, line 346: nit: This comment feels like it could use a little more context. What do you think of mentioning the kinds of events that are published, and which components publish or subscribe to it?

agent/consul/server.go
```diff
@@ -651,18 +661,21 @@ func NewServer(config *Config, flat Deps, publicGRPCServer *grpc.Server) (*Serve
...
+func newFSMFromConfig(logger hclog.Logger, gc *state.TombstoneGC, config *Config, publisher *stream.EventPublisher) *fsm.FSM {
+	deps := fsm.Deps{
+		Logger: logger,
+		NewStateStore: func() *state.Store {
+			return state.NewStateStore(gc)
+		},
+	}
+
 	if config.RPCConfig.EnableStreaming {
+		deps.Publisher = publisher
```
boxofrad, line 673: I think it's no longer the case that we only want to set the publisher if streaming is enabled, right? Will `WatchRoots` and `WatchServers` depend on this config flag?

agent/consul/stream/event_publisher.go
```diff
@@ -100,12 +100,41 @@ func NewEventPublisher(handlers SnapshotHandlers, snapCacheTTL time.Duration) *E
...
+	if _, found := e.snapshotHandlers[topic]; !found {
+		return fmt.Errorf("topic %s is not registered", topic)
+	}
+
+	e.forceEvictByTopic(topic)
+	e.subscriptions.closeAllByTopic(topic)
+
+	return nil
+}
+
```
boxofrad, line 137: Do we need to hold `e.lock` in these methods (particularly `RefreshTopic`). Or are they guarded by a higher-level lock?

boxofrad (commented):
Looking good! Left a couple of questions 🙇🏻‍♂️

agent/consul/server.go
mkeeler: This is a good point. I think it should be safe to always pass in an event publisher and thus remove this EnableStreaming check.

The main difference will be that we are actually tracking events whereas before if streaming was disabled we would configure the state store with a NoOpEventPublisher.

I think its sufficient to allow publishing events but have the config just disable the gRPC APIs around streaming.

agent/consul/stream/event_publisher.go
mkeeler: Not here but I did add one in `forceEvictByTopic`.

Access to the snapshotHandlers map isn't locked because the expectation is that all handlers get registered prior to executing Run (documented that in the RegisterHandler doc comment)

agent/consul/server.go
mkeeler: Done

agent/consul/stream/event_publisher.go
boxofrad: Nice, thanks for clarifying!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/634853.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 8bad5105b74e2629fc1a1ba41bcbe0cd781e501b onto `release/1.12.x` succeeded!

PR #12692, opened 2022-04-04
[rpc/middleware][consul] plumb intercept off, add server level happy test

FFMMM:
### overview

This patch does 2 main things:

* plumbs down a config `GetNetRPCInterceptorFunc` option (not exposed thru to the agent)
* uses that func to add server level happy path testing AND prove that we can turn the `net/rpc` interceptor off if we ever need to

Additionally, it adds two small changes to the `rpc/middleware` package:

* exports the recorder func
* make `interceptor_test.go` run in parallel if needed
 

### reviewer notes

* best viewed by commits


agent/consul/server_test.go
```diff
@@ -1129,6 +1133,126 @@ func TestServer_RPC(t *testing.T) {
...
+
+	s, err := newServer(t, conf)
+	if err != nil {
+		t.Fatalf("err: %v", err)
+	}
+	defer s.Shutdown()
+	testrpc.WaitForTestAgent(t, s.RPC, "dc1")
+
+	// asserts
+	t.Run("test happy path for metrics interceptor", func(t *testing.T) {
```
FFMMM, line 1235: * see `interceptors_test.gp` for "data" level and unit level testing
* see `metrics_test.go` for e2e/ bdd testing

all in main already

agent/consul/server_test.go
```diff
@@ -1129,6 +1133,126 @@ func TestServer_RPC(t *testing.T) {
...
+		t.Fatalf("err: %v", err)
+	}
+	defer s.Shutdown()
+	testrpc.WaitForTestAgent(t, s.RPC, "dc1")
+
+	// "hijack" the rpc recorder for asserts;
+	// note that there will be "internal" net/rpc calls made
+	// that will still show up; those don't go thru the net/rpc interceptor;
+	// see consul.agent.rpc.middleware.RPCTypeInternal for context
+	s.rpcRecorder = &middleware.RequestRecorder{
```
mkeeler, line 1166: Is there a data race here? I assume the test agent behind the scenes will be performing RPCs and swapping in an RPC recorder seems like it could cause the test to panic in the worst case.

agent/consul/config.go
```diff
@@ -116,6 +118,11 @@ type Config struct {
 	// configured at this point.
 	NotifyListen func()
 
+	// GetNetRPCInterceptorFunc, if not nil, sets the net/rpc rpc.ServerServiceCallInterceptor on
+	// the server side to record metrics around the RPC requests. If nil, no interceptor is added to
+	// the rpc server.
+	GetNetRPCInterceptorFunc func(recorder *middleware.RequestRecorder) rpc.ServerServiceCallInterceptor
```
mkeeler, line 124: I don't think I have an opinion either way but did you consider putting this in the Deps instead of the Config?

*Most* of whats in Config are simple values and not types/funcs (one exception being the `NotifyListen` func above here).

Then again, I don't know how applicable it is with what else is in Deps. What do you think?

agent/consul/server.go
```diff
@@ -380,21 +380,20 @@ func NewServer(config *Config, flat Deps, publicGRPCServer *grpc.Server) (*Serve
 	recorder := middleware.NewRequestRecorder(serverLogger)
```
dhiaayachi, line 380: Maybe inject the recorder to the server by adding it as a param to NewServer. This way we can avoid the race that @mkeeler is referring to.

agent/consul/server.go
```diff
@@ -380,21 +380,20 @@ func NewServer(config *Config, flat Deps, publicGRPCServer *grpc.Server) (*Serve
...
+		grpcConnPool:            flat.GRPCConnPool,
+		eventChLAN:              make(chan serf.Event, serfEventChSize),
+		eventChWAN:              make(chan serf.Event, serfEventChSize),
+		logger:                  serverLogger,
+		loggers:                 loggers,
+		leaveCh:                 make(chan struct{}),
+		reconcileCh:             make(chan serf.Member, reconcileChSize),
+		router:                  flat.Router,
+		rpcRecorder:             recorder,
+		rpcServer:               rpc.NewServer(),
```
dhiaayachi, line 395: What are the implication of creating a new rpc server in here and then creating a new one later, do the NewServer() call use any system resources (ex socket)?
I think another way in here is to make `GetNetRPCInterceptorFunc` return a nil (instead of being nil) in the case we don't want to intercept the calls and make sure that the nil interceptor is a no op.

agent/consul/server_test.go
```diff
@@ -1129,6 +1133,126 @@ func TestServer_RPC(t *testing.T) {
...
+		t.Skip("too slow for testing.Short")
+	}
+
+	storage := make(map[string]float32)
+	keyMakingFunc := func(key []string) string {
+		return strings.Join(key, "+")
+	}
+
+	simpleRecorderFunc := func(key []string, val float32, labels []metrics.Label) {
+		storage[keyMakingFunc(key)] = val
```
dhiaayachi, line 1147: maybe just fail the test directly in here, we don't expect any recording to happen in here right?

dhiaayachi (changes requested):
Thank you for working on this @FFMMM. The changes in here make the tests for rpc metrics definitely easier to write and reason about.

I had a few comments around the recorder creation in `NewServer` and the rpc server creation code.

agent/consul/config.go
FFMMM: That's a fair point. Chatting with Dhia more, I want to move this to `Deps` and I think it should work better ✍🏼

agent/consul/server_test.go
FFMMM: that's a good point -- I recall that folks were not inclined to do this when I was adding interceptor;

i'm slightly biased to pretend that this function does some recording and then to look at the `map` to see that the right values are in there.

agent/consul/server_test.go
FFMMM: ‼️ that is quite possible; I'm moving the `recorder` to be a part of deps too so this shouldn't be a problem anymore.

FFMMM:
@dhiaayachi, I addressed the changes requested. It seems that `verify-ci` is currently stuck so I am going to dry to dismiss the review to see if it kicks the tires on the workflow.

FFMMM:
or maybe it's the outage 🙈  https://status.circleci.com/

dhiaayachi (approved):
LGTM with a small nit!

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/627963.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/628205.

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 5245251bbf73b20943eec4d4f8dd5038916cd3f8 onto `release/1.12.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/628205)

PR #12702, opened 2022-04-05
Add versions compatibility tests between Consul 

dhiaayachi:
### Problem
As we add new features to Consul a new possible family of bugs arise that is related to compatibility between different Consul version. So questions like `Am I able to run a Consul cluster with Servers in version X and Clients in version Y?` or `Am I able to upgrade my Consul Servers from version X to version X+1 without service interruption?` become more and more relevant.
Until now, performing this kind of test automatically was not possible in Consul, because most of the tests that Consul rely on fall in one of following categories:
- Unit tests that test a small feature contained by it self
- Open box tests that run a Consul node or rarely multiple nodes inside the test
- API tests that run Consul binary to test the API package (very hard to run multiple binaries in the same environment)
- Integration tests that are specifics to test some integrations (Consul-Nomad, Consul-Envoy)
The only category in here where versions compatibility test can fall is the Integration tests category but the way those tests are constructed is not flexible enough to allow this without a major modification. Moreover, a lot of `test glue` is written in bash for those tests which make them hard to work with.

### Proposed solution
Based on the fact that Consul code is written in go and all the developers working in Consul are proficient with go, the most efficient way of writing integrations tests for Consul is to do it in go. Also, Consul is not designed to have multiples instances of it running in the same environment, so using Containers would allow to isolate multiple Consul instances running in parallel and reduce environment dependencies.

In this PR a new test module is introduced, that test module have 2 packages:
- consul-container: This is an implementation of instantiating Consul within a Container with a given config, it's based on [test-containers](https://github.com/testcontainers/testcontainers-go) library 
- consul-cluster is an abstraction layer to instantiate a Consul cluster with a given set of Consul configurations.
This PR also introduce a test package `upgrade` that introduce a set of simple tests to check that multiple scenarios of different Consul versions work properly together. The test is parametrized using command line flags and need two parameters:
- uut-image: docker image to be used as UUT (unit under test)
- latest-image: docker image to be used as latest
those parameters can be used to run different versions compatibility tests in CI.
This PR also provide a docker file for local environment that would help a user to create a docker image based on a local branch and use it to run the tests on a local environment given that Docker is available.


test/integration/consul-container/upgrade/basic_test.go
```diff
@@ -0,0 +1,253 @@
...
+		close(ch)
+	}()
+	err = client.Agent().ServiceRegister(&api.AgentServiceRegistration{Name: "api", Port: 9998})
+	timer := time.NewTimer(1 * time.Second)
+	select {
+	case <-ch:
+	case <-timer.C:
+		t.Fatalf("test timeout")
+	}
+}
```
dhiaayachi, line 145: @rboyer is this the kind of test we discussed for streaming?

rboyer: For extra bonus points see if you can get the right `X-header` that will tell you it definitely used streaming to serve you the response. You _may_ need to drop to a straight `net/http` instead for that metadata though.

test/integration/consul-container/upgrade/basic_test.go
```diff
@@ -0,0 +1,253 @@
...
+	require.Equal(t, "api", service[0].ServiceName)
+	require.Equal(t, 9999, service[0].ServicePort)
+
+	ch := make(chan struct{})
+	go func() {
+		service, _, err := client.Catalog().Service("api", "", &api.QueryOptions{WaitIndex: meta.LastIndex})
+		require.NoError(t, err)
+		require.Len(t, service, 1)
+		require.Equal(t, "api", service[0].ServiceName)
+		require.Equal(t, 9998, service[0].ServicePort)
```
rboyer, line 135: `t.Fatalf` (or things that wrap it like require) should not be used from goroutines. If you need to do an assertion then change things to be more like:
```
var (
  errCh = make(chan error)
  resultCh = make(chan RESULTTYPE)
)
go func() {
    res, err := STUFF
    if err != nil {
       errCh <- err
    } else {
      resultCh <- res
    }
}()

select {
case err := <-errCh:
     require.NoError(t, err)
case res := <-resultCh:
    < all your require assertions>
case <-timer.C:
    die
    }
    ```
}

File: build-support/docker/Consul-Dev-Build.dockerfile [hunk 3]
```diff
@@ -0,0 +1,9 @@
+ARG CONSUL_IMAGE_VERSION=latest
+FROM golang:1.17-alpine As builder
+RUN apk update && apk add iptables make bash git go
+RUN rm -rf ./bin/consul
+WORKDIR /consul
+COPY . .
+RUN make dev
+FROM consul:${CONSUL_IMAGE_VERSION}
+COPY --from=builder /consul/bin/consul  /bin/consul
```
FFMMM, line 9:
```suggestion
FROM golang:1.17-alpine AS builder
RUN apk update && apk add iptables make bash git go
RUN rm -rf ./bin/consul
WORKDIR /consul
COPY . .
RUN make dev
FROM consul:${CONSUL_IMAGE_VERSION}
COPY --from=builder /consul/bin/consul  /bin/consul

```

test/integration/consul-container/go.mod
```diff
@@ -0,0 +1,30 @@
+module github.com/hashicorp/consul/integration/consul-container
+
+go 1.16
```
FFMMM, line 3: could we go w 1.18 for this since it's a new module?

test/integration/consul-container/upgrade/basic_test.go
```diff
@@ -0,0 +1,324 @@
...
+	case service := <-ch:
+		require.Len(t, service, 1)
+		require.Equal(t, serviceName, service[0].Service.Service)
+		require.Equal(t, 9998, service[0].Service.Port)
+	case <-timer.C:
+		t.Fatalf("test timeout")
+	}
+}
+
+func clientsCreate(numClients int) ([]consulNode.Node, error) {
```
FFMMM, line 270: if we plan on putting test cases largely grouped into files like `basic_test.go`, maybe we could factor these out in some `testutils` package?

test/integration/consul-container/upgrade/basic_test.go
```diff
@@ -0,0 +1,324 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+var curImage = flag.String("uut-version", "local", "docker image to be used as UUT (unit under test)")
+var latestImage = flag.String("latest-version", "latest", "docker image to be used as latest")
+
+const retryTimeout = 10 * time.Second
+const retryFrequency = 500 * time.Millisecond
+
+func TestLatestGAServersWithCurrentClients(t *testing.T) {
```
FFMMM, line 27: Could we add doc strings for the intention of each test? I find it helpful for future proofing and it helps me understand the goal of test better 🤓

FFMMM (commented):
This is awesome !! 💯 So glad you put this out.

It helps us increase our test coverage and helps with compatibility confidence. I'm super excited for what more we can do with this as folks build off of it and add their own test use cases.

I'm also curious to see what others think of this. For now, I left some questions and small nits.

test/integration/consul-container/go.mod
dhiaayachi: This is not the version used, this the lowest version that can be compatible with the module. We actually use the same version as the one we use to compile Consul to run this, which is for now 1.17

test/integration/consul-container/libs/consul-cluster/consul-cluster.go
```diff
@@ -0,0 +1,91 @@
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func (c *Cluster) AddNodes(nodes []consulnode.Node) error {
+	n0 := c.Nodes[0]
```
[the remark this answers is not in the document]
dhiaayachi: good catch, I will add a check for this.

test/integration/consul-container/libs/consul-node/consul-container.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+type Node interface {
+	Terminate() error
+	GetClient() *api.Client
+	GetAddr() (string, int)
+}
+
+const bootLogLine = "Consul agent running"
+const disableRYUKEnv = "TESTCONTAINERS_RYUK_DISABLED"
```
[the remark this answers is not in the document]
dhiaayachi: I think we should not extract this prematurely, if we have more of those we can still extract them.

test/integration/consul-container/libs/consul-cluster/consul-cluster.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	return nil
+}
+
+func (c *Cluster) Terminate() error {
+	for _, n := range c.Nodes {
+		err := n.Terminate()
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
dhiaayachi: I went with that decision randomly to make the code simpler.
test-container use [RYUK](https://hub.docker.com/r/testcontainers/ryuk) to monitor the created containers and do clean up when the test done. So terminating the containers properly is not that important.

test/integration/consul-container/upgrade/basic_test.go
dhiaayachi: It will very likely be the case in the future but I would wait for another test suite to be created before doing that. I don't want to extract this prematurely without having a use case for extracting it.

test/integration/consul-container/upgrade/basic_test.go
```diff
@@ -0,0 +1,324 @@
...
+					log_level="TRACE"`,
+				Cmd:     []string{"agent", "-client=0.0.0.0"},
+				Version: *curImage,
+			})
+	}
+	return Clients, err
+}
+
+func serviceCreate(t *testing.T, client *api.Client, serviceName string) (error, uint64) {
+	err := client.Agent().ServiceRegister(&api.AgentServiceRegistration{Name: serviceName, Port: 9999})
```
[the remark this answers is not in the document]
dhiaayachi: I think that using the api should be the default (and hopefully the only way) for the following reasons:
- we have an api package for this exact use case (interact with Consul through go code)
- it will help us improve our api package, by using it (like the change I did in the PR)
- it help keep the implementation details encapsulated (we can have another way of instantiating Consul that satisfy the node interface and still run the same tests)

test/integration/consul-container/libs/consul-cluster/consul-cluster.go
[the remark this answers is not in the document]
FFMMM: That makes sense! I would suggest adding some inline comments or docstrings about this then 📚 as not everyone will be familiar w `test-containers`.

test/integration/consul-container/upgrade/basic_test.go
```diff
@@ -0,0 +1,324 @@
...
+		if err != nil {
+			errCh <- err
+		} else {
+			ch <- service
+		}
+	}()
+	err = client.Agent().ServiceRegister(&api.AgentServiceRegistration{Name: serviceName, Port: 9998})
+	timer := time.NewTimer(1 * time.Second)
+	select {
+	case err := <-errCh:
```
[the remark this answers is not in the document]
dhiaayachi: I think I can pull this as a table test but it will impact test readability (This is why I'm always weary about introducing table test, unless it's just data variation of the same test)

.circleci/config.yml
```diff
@@ -802,7 +800,55 @@ jobs:
           working_directory: ui/packages/consul-ui
           command: make test-coverage-ci
       - run: *notify-slack-failure
-
+      
+  upgrade-integration-test-latest:
```
FFMMM, line 804: could we call this something else? Like:

```suggestion
  testcontainer-integration-test:
```

Today we just test the `compatibility` use case but I can see this easily growing to encompass more use cases!

FFMMM (approved):
🚀 this LGTM 💯 

thanks for taking these set of changes on; i think there's one non blocking nit left

.circleci/config.yml
```diff
@@ -257,8 +255,8 @@ jobs:
       - run:
          command: |
             sudo rm -rf /usr/local/go
-            wget https://dl.google.com/go/go1.18.1.linux-arm64.tar.gz
-            sudo tar -C /usr/local -xzvf go1.18.1.linux-arm64.tar.gz
+            wget https://golang.org/dl/go${GO_VERSION}.linux-arm64.tar.gz
```
eculver, line 258: Small, but this was updated to use `dl.google.com` back in #12818 since it is the canonical location for the tarball and there are two redirects between it and `golang.org/dl`.

test/integration/consul-container/libs/consul-node/consul-container.go
```diff
@@ -0,0 +1,130 @@
...
+	err = os.Mkdir(tmpDir+"/config", 0777)
+	if err != nil {
+		return nil, err
+	}
+	configFile := tmpDir + "/config/config.hcl"
+	err = os.WriteFile(configFile, []byte(config.HCL), 0644)
+	if err != nil {
+		return nil, err
+	}
+	skipReaper := isRYUKDisabled()
```
mkeeler, line 56: Why does RYUK mean? I see that it controls whether the lib setups a reaper for this container but feel like I am missing some meaning here.

test/integration/consul-container/libs/consul-node/consul-container.go
```diff
@@ -0,0 +1,130 @@
...
+func (c *consulContainerNode) Terminate() error {
+	return c.container.Terminate(c.ctx)
+}
+
+func isRYUKDisabled() bool {
+	skipReaperStr := os.Getenv(disableRYUKEnv)
+	skipReaper := true
+	if skipReaperStr == "1" || strings.ToLower(skipReaperStr) == "true" {
+		skipReaper = true
+	}
```
mkeeler, line 128: nit: You could use `strconv.ParseBool` here instead of checking for the two specific strings. I think the main differences is that it also supports single character `t` or `f` case insensitive. Then again you would have to handle the error return so maybe its not worth it.

test/integration/consul-container/upgrade/healthcheck_test.go
```diff
@@ -0,0 +1,328 @@
...
+)
+
+var curImage = flag.String("uut-version", "local", "docker image to be used as UUT (unit under test)")
+var latestImage = flag.String("latest-version", "latest", "docker image to be used as latest")
+
+const retryTimeout = 10 * time.Second
+const retryFrequency = 500 * time.Millisecond
+
+// Test health check GRPC call using Current Clients and Latest GA Servers
+func TestLatestGAServersWithCurrentClients(t *testing.T) {
```
mkeeler, line 28: Technically speaking this scenario isn't supported is it.

Server versions must always be greater than the client versions. Maybe I am reading this wrong but it sounds like this would be used to test maybe a 1.11.5 servers with 1.11.6 clients (assuming 1.11.6 hasn't been released yet)

test/integration/consul-container/upgrade/healthcheck_test.go
```diff
@@ -0,0 +1,328 @@
...
+
+// Test health check GRPC call using Mixed (majority current) Servers and Latest GA Clients
+func TestMixedServersMajorityCurrentGAClient(t *testing.T) {
+	t.Parallel()
+	var configs []consulNode.Config
+	configs = append(configs,
+		consulNode.Config{
+			HCL: `node_name="` + utils.RandName("consul-server") + `"
+					log_level="TRACE"
+					bootstrap_expect=3
```
mkeeler, line 209:
```suggestion
```

I would remove this to prevent this node from gaining leadership accidentally. It appears like this scenario is trying to mimic the one where new servers are being added to the cluster to perform an upgrade but there are not enough of the newer ones there yet. In that case leaving out the bootstrap expect ensures this node doesn't accidentally gain first leadership when the cluster is bootstrapped and therefore will more accurately mimic the desired state.

test/integration/consul-container/libs/consul-cluster/consul-cluster.go
```diff
@@ -0,0 +1,105 @@
+package consul_cluster
```
boxofrad, line 1: Effective Go [recommends](https://go.dev/doc/effective_go#package-names) that package names are kept short and to the point:

>  package name should be good: short, concise, evocative. By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps.

What do you think of naming this `cluster` and the other package `node`?

test/integration/consul-container/libs/consul-cluster/consul-cluster.go
```diff
@@ -0,0 +1,105 @@
...
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	consulnode "github.com/hashicorp/consul/integration/consul-container/libs/consul-node"
```
boxofrad, line 8: Similarly, if this package was called `node` we wouldn't need the import alias here 🙇🏻‍♂️

test/integration/consul-container/libs/consul-cluster/consul-cluster.go
```diff
@@ -0,0 +1,105 @@
...
+
+// AddNodes add a number of nodes to the current cluster and join them to the cluster
+func (c *Cluster) AddNodes(nodes []consulnode.ConsulNode) error {
+	if len(c.Nodes) < 1 {
+		return fmt.Errorf("cannot add a node to an empty cluster")
+	}
+	n0 := c.Nodes[0]
+	for _, node := range nodes {
+		addr, _ := n0.GetAddr()
+		err := node.GetClient().Agent().Join(addr, false)
```
boxofrad, line 45: nit: as we're only using `n0` to capture its address, I wonder if grabbing the address outside of the loop (and giving it a more explicit name) would help readaibility. What do you think?

```suggestion
	joinAddr, _ := c.Nodes[0].GetAddr()
	for _, node := range nodes {
		err := node.GetClient().Agent().Join(joinAddr, false)
```

test/integration/consul-container/libs/consul-cluster/consul-cluster.go
```diff
@@ -0,0 +1,105 @@
...
+	cluster := Cluster{}
+
+	for _, c := range configs {
+		n, err := consulnode.NewConsulContainer(context.Background(), c)
+		if err != nil {
+			return nil, err
+		}
+		cluster.Nodes = append(cluster.Nodes, n)
+	}
+	err := cluster.join()
```
boxofrad, line 30: It looks like the logic in `join` already exists in`AddNode`. Could we just call that method from here instead?

```suggestion
	nodes := make([]consulnode.ConsulNode, len(configs))
	for idx, c := range configs {
		n, err := consulnode.NewConsulContainer(context.Background(), c)
		if err != nil {
			return nil, err
		}
		nodes[idx] = n
	}
	if err := cluster.AddNodes(nodes); err != nil {
		return nil, err
	}
```

test/integration/consul-container/libs/consul-node/consul-container.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/testcontainers/testcontainers-go/wait"
+
+	"github.com/docker/docker/pkg/ioutils"
+
+	"github.com/hashicorp/consul/integration/consul-container/libs/utils"
+
+	"github.com/testcontainers/testcontainers-go"
+
+	"github.com/hashicorp/consul/api"
+)
```
boxofrad, line 19: nit: we generally organize our imports into three groups: stdlib, 3rd party, this module ([`goimports -local`](https://pkg.go.dev/golang.org/x/tools/cmd/goimports) can do this for you automatically).

```suggestion
import (
	"context"
	"fmt"
	"os"
	"strings"
	"time"

	"github.com/docker/docker/pkg/ioutils"
	"github.com/testcontainers/testcontainers-go"
	"github.com/testcontainers/testcontainers-go/wait"
	"github.com/hashicorp/consul/api"

	"github.com/hashicorp/consul/integration/consul-container/libs/utils"
)
```

test/integration/consul-container/libs/consul-node/consul-container.go
boxofrad: Same! A quick Google brought up [this character](https://en.wikipedia.org/wiki/Ryuk_(Death_Note))? 💀😅 Could we just call it `skipReaper` instead?

test/integration/consul-container/libs/consul-node/consul-node.go
```diff
@@ -0,0 +1,17 @@
+package consul_node
+
+import "github.com/hashicorp/consul/api"
+
+// ConsulNode represent a Consul node abstraction
+type ConsulNode interface {
```
boxofrad, line 6: Could we go with `Node`? We're already in the Consul repo 😉 

```suggestion
type Node interface {
```

test/integration/consul-container/libs/consul-node/consul-container.go
```diff
@@ -0,0 +1,130 @@
...
+	container testcontainers.Container
+	ip        string
+	port      int
+}
+
+// NewConsulContainer create a ConsulNode implemented as a consulContainerNode
+func NewConsulContainer(ctx context.Context, config Config) (ConsulNode, error) {
+
+	name := utils.RandName("consul-")
+	ctx = context.WithValue(ctx, "name", name)
```
boxofrad, line 38: I couldn't find where this context value gets read? Is it inside testcontainers?

test/integration/consul-container/upgrade/healthcheck_test.go
```diff
@@ -0,0 +1,328 @@
...
+	consulCluster "github.com/hashicorp/consul/integration/consul-container/libs/consul-cluster"
+	consulNode "github.com/hashicorp/consul/integration/consul-container/libs/consul-node"
+
+	"github.com/hashicorp/consul/integration/consul-container/libs/utils"
+	"github.com/hashicorp/consul/sdk/testutil/retry"
+
+	"github.com/stretchr/testify/require"
+)
+
+var curImage = flag.String("uut-version", "local", "docker image to be used as UUT (unit under test)")
```
boxofrad, line 21: Could we be more explicit in this flag name? UUT wasn't immediately obvious to me without the explanatory text, but `target` or `current` probably would be?

```suggestion
var curImage = flag.String("target-version", "local", "docker image to be used as test target")
```

test/integration/consul-container/upgrade/healthcheck_test.go
```diff
@@ -0,0 +1,328 @@
...
+var latestImage = flag.String("latest-version", "latest", "docker image to be used as latest")
+
+const retryTimeout = 10 * time.Second
+const retryFrequency = 500 * time.Millisecond
+
+// Test health check GRPC call using Current Clients and Latest GA Servers
+func TestLatestGAServersWithCurrentClients(t *testing.T) {
+	t.Parallel()
+	numServers := 3
+	Cluster, err := serversCluster(t, numServers, *latestImage)
```
boxofrad, line 31: nit: only package-exported variables/constants/types should have names beginning with a capital letter.

boxofrad (commented):
This is really awesome, impactful work! I love the approach, and I'm really excited to have this type of test coverage 👏🏻 ❤️ 🎉 

I've left a bunch of mostly stylistic and readability comments. Please shout if you'd like to talk about them synchronously.

build-support/docker/Consul-Dev-Build.dockerfile
```diff
@@ -0,0 +1,9 @@
+ARG CONSUL_IMAGE_VERSION=latest
```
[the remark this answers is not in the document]
dhiaayachi: This is not for the CI, I added this to simplify the local setup/run. That said, I forgot to document that, I will add some documentation on how to run it locally.

.circleci/config.yml
dhiaayachi: Thank you, I did a manual merge and did not see the update. I will change it using the GO_VERSION env var to ensure we don't have to change a lot of things when we upgrade go version.

test/integration/consul-container/libs/consul-node/consul-container.go
```diff
@@ -0,0 +1,130 @@
...
+		return nil, err
+	}
+	skipReaper := isRYUKDisabled()
+	req := testcontainers.ContainerRequest{
+		Image:        "consul:" + config.Version,
+		ExposedPorts: []string{"8500/tcp"},
+		WaitingFor:   wait.ForLog(bootLogLine).WithStartupTimeout(10 * time.Second),
+		AutoRemove:   false,
+		Name:         name,
+		BindMounts:   map[string]string{"/consul/config/config.hcl": configFile},
```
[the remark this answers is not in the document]
dhiaayachi: I will check that, I think they merged VolumeMounts and BindMounts. I will try to upgrade and fix that.

test/integration/consul-container/libs/consul-node/consul-container.go
dhiaayachi: https://github.com/testcontainers/moby-ryuk

test/integration/consul-container/upgrade/healthcheck_test.go
dhiaayachi: I did not know that the requirements are to always have servers upgraded first. Good to know I will remove this scenario.

test/integration/consul-container/libs/consul-node/consul-container.go
dhiaayachi: it was a left over of an experiment I was doing. I will remove it

.circleci/config.yml
dhiaayachi: I don't want to tie this to testcontainer necessarily. I think we can also create other ways of instantiating Consul and run the exact same tests in different environments. I renamed it to `upgrade-integration-test` We can rename it again when we know what other use cases will fit.

test/integration/consul-container/libs/consul-node/consul-container.go
boxofrad: Gotcha, thanks! Mind putting that link in a doc comment on the method, for future readers?

build-support/docker/Consul-Dev-Build.dockerfile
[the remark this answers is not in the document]
mkeeler: Does this container need to be different than the one built by the `dev-docker` make target. That make target will build the linux binary and then shove it into the Consul container. It looks to have an end result which is the same as this dockerfile. The main advantage of building the binary outside the container is that the go module cache can be used and can greatly speed up the build.

test/integration/consul-container/libs/node/consul-container.go
```diff
@@ -0,0 +1,129 @@
...
+		return nil, err
+	}
+	skipReaper := isRYUKDisabled()
+	req := testcontainers.ContainerRequest{
+		Image:        "consul:" + config.Version,
+		ExposedPorts: []string{"8500/tcp"},
+		WaitingFor:   wait.ForLog(bootLogLine).WithStartupTimeout(10 * time.Second),
+		AutoRemove:   false,
+		Name:         name,
+		BindMounts:   map[string]string{"/consul/config/config.hcl": configFile},
```
mkeeler: I think the original comment got outdated due to other changes. I think you are right though. It does look like the field got replaced with a generic `Mounts` field of type [`ContainerMounts`](https://pkg.go.dev/github.com/testcontainers/testcontainers-go#ContainerMounts).

Still, with new code I think it might be worth pulling in the latest version of the lib and making the updates so we are not immediately using deprecated/removed functionality.

test/integration/consul-container/libs/cluster/cluster.go
```diff
@@ -0,0 +1,90 @@
...
+		if err != nil {
+			return err
+		}
+		c.Nodes = append(c.Nodes, node)
+	}
+	return nil
+}
+
+// Terminate will attempt to terminate all the nodes in the cluster
+// if a node termination fail, Terminate will abort and return and error
```
boxofrad, line 57:
```suggestion
// Terminate will attempt to terminate all nodes in the cluster. If any node
// termination fails, Terminate will abort and return an error.
```

test/integration/consul-container/libs/cluster/cluster.go
```diff
@@ -0,0 +1,90 @@
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Leader return the cluster leader node
+// if no leader is available or the leader is not part of the cluster
+// an error will be returned
```
boxofrad, line 70:
```suggestion
// Leader returns the cluster leader node, or an error if no leader is
// available.
```

test/integration/consul-container/libs/node/consul-container.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/testcontainers/testcontainers-go/wait"
+
+	"github.com/hashicorp/consul/integration/consul-container/libs/utils"
+)
+
+const bootLogLine = "Consul agent running"
+const disableRYUKEnv = "TESTCONTAINERS_RYUK_DISABLED"
+
+// consulContainerNode implement a Node
+// it instantiate Consul as a container
```
boxofrad, line 22:
```suggestion
// consulContainerNode implements the Node interface by running a Consul node
// in a container.
```

test/integration/consul-container/libs/node/consul-container.go
```diff
@@ -0,0 +1,128 @@
...
+type consulContainerNode struct {
+	ctx       context.Context
+	client    *api.Client
+	container testcontainers.Container
+	ip        string
+	port      int
+}
+
+// NewConsulContainer create a Node implemented as a consulContainerNode
```
boxofrad, line 31:
```suggestion
// NewConsulContainer starts a Consul node in a container with the given config.
```

test/integration/consul-container/libs/node/consul-container.go
```diff
@@ -0,0 +1,128 @@
...
+	c.client, err = api.NewClient(apiConfig)
+	c.ctx = ctx
+
+	if err != nil {
+		return nil, err
+	}
+	return c, nil
+}
+
+// GetClient return the client associated with the Node
```
boxofrad, line 102:
```suggestion
// GetClient returns an API client that can be used to communicate with the Node.
```

test/integration/consul-container/libs/node/consul-container.go
```diff
@@ -0,0 +1,128 @@
...
+func (c *consulContainerNode) GetClient() *api.Client {
+	return c.client
+}
+
+// GetAddr return the network address associated with the Node
```
boxofrad, line 107:
```suggestion
// GetAddr return the network address associated with the Node.
```

test/integration/consul-container/libs/node/consul-container.go
```diff
@@ -0,0 +1,128 @@
...
+func (c *consulContainerNode) GetAddr() (string, int) {
+	return c.ip, c.port
+}
+
+// Terminate will attempt to terminate a consulContainerNode
+// if this fail the container will be killed by RYUK if enabled
```
boxofrad, line 113:
```suggestion
// Terminate attempts to terminate the container. On failure, an error will be
// returned and the reaper process (RYUK) will handle cleanup.
```

test/integration/consul-container/libs/node/consul-container.go
```diff
@@ -0,0 +1,128 @@
...
+func (c *consulContainerNode) Terminate() error {
+	return c.container.Terminate(c.ctx)
+}
+
+// isRYUKDisabled check if we need to disable
+// the use a reaper container.
+// https://github.com/testcontainers/moby-ryuk
```
boxofrad, line 120:
```suggestion
// isRYUKDisabled returns whether the reaper process (RYUK) has been disabled
// by an environment variable.
//
// https://github.com/testcontainers/moby-ryuk
```

test/integration/consul-container/upgrade/healthcheck_test.go
```diff
@@ -0,0 +1,274 @@
...
+	t.Parallel()
+	numServers := 3
+	cluster, err := serversCluster(t, numServers, *targetImage)
+	require.NoError(t, err)
+	defer Terminate(t, cluster)
+	numClients := 1
+
+	Clients, err := clientsCreate(numClients)
+	client := cluster.Nodes[0].GetClient()
+	err = cluster.AddNodes(Clients)
```
boxofrad, line 38:
```suggestion
	clients, err := clientsCreate(numClients)
	client := cluster.Nodes[0].GetClient()
	err = cluster.AddNodes(clients)
```

test/integration/consul-container/upgrade/healthcheck_test.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	cluster, err := cluster.New(configs)
+	require.NoError(t, err)
+	defer Terminate(t, cluster)
+
+	numClients := 1
+	Clients, err := clientsCreate(numClients)
+	client := Clients[0].GetClient()
+	err = cluster.AddNodes(Clients)
```
boxofrad, line 110:
```suggestion
	clients, err := clientsCreate(numClients)
	client := Clients[0].GetClient()
	err = cluster.AddNodes(clients)
```

test/integration/consul-container/upgrade/healthcheck_test.go
```diff
@@ -0,0 +1,274 @@
...
+	for i := 0; i < numClients; i++ {
+		Clients[i], err = node.NewConsulContainer(context.Background(),
+			node.Config{
+				HCL: `node_name="` + utils.RandName("consul-client") + `"
+					log_level="TRACE"`,
+				Cmd:     []string{"agent", "-client=0.0.0.0"},
+				Version: *targetImage,
+			})
+	}
+	return Clients, err
```
boxofrad, line 232:
```suggestion
	clients := make([]node.Node, numClients)
	var err error
	for i := 0; i < numClients; i++ {
		clients[i], err = node.NewConsulContainer(context.Background(),
			node.Config{
				HCL: `node_name="` + utils.RandName("consul-client") + `"
					log_level="TRACE"`,
				Cmd:     []string{"agent", "-client=0.0.0.0"},
				Version: *targetImage,
			})
	}
	return clients, err
```

boxofrad (approved):
This is looking great! 👏🏻 I hope you don't mind but I've made a few suggestions for formatting/grammatical changes to some of the doc comments — please forgive my pedantry, and feel free to take or leave them.

build-support/docker/Consul-Dev-Build.dockerfile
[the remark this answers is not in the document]
dhiaayachi: removed this Dockerfile in favor of using `make dev-docker`

mkeeler (approved):

hc-github-team-consul-core:
:cherries::x: Cherry pick of commit 4488b6c3392f2be423d9ff8702fc8f5eb3b5c772 onto `release/1.12.x` failed! [Build Log](https://circleci.com/gh/hashicorp/consul/650785)

PR #12719, opened 2022-04-07
options.mdx: `addresses` affects client ifaces

asymmetric:


asymmetric:
@Amier3 yes, it was confusing while setting up a cluster as there are multiple options that deal with binding, and hhopefully this change helps clarify what the `addresses` option is for.

asymmetric:
@Amier3 i've signed the CLA.

Issue #12221, opened 2022-01-28
UDP check for service stanza


gfonseca-tc:
Please search the existing issues for relevant feature requests, and use the reaction feature (https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to add upvotes to pre-existing requests.

#### Feature Description

Consul support many different check types but not UDP. It could be very useful for all sorts of applications that listen to UDP ports.

#### Use Case(s)

One example: the StatsD receiver, in the OpenTelemetry collector, listens to UDP. Currently there is no way of creating a Consul check against it.

Amier3:
Hey @gfonseca-tc 

Thanks for making this request! We've gotten similar requests in the past for UDP check support, so I appreciate you bringing this back to the forefront 👍

fulviodenza:
I'd love to work on this with someone else from hashicorp if possible @Amier3

dhiaayachi:
Hey @fulviodenza,
Sorry for the late response I was caught on other stuff. Here is some notes that we have about  [the possible entry points](https://github.com/hashicorp/consul/blob/main/docs/service-discovery/health-checks.md) for a check in Consul. 

I hope that can get you started, let us know if you have any questions?

fulviodenza:
Hi @dhiaayachi thank you for the response! I've read the agent/checks/check.go file and It was pretty clear the way to implement UDP check, but I think I'll need help to test it. I think I'll start the development of this feature next week.
Before starting implementation, I'll analyze the changing to do and write down all files and struct that I think will be affected, if this list will be complete, I'll start the implementation. 
I'll write in this thread next few days with this list. Let me know if something doesn't sound good.

fulviodenza:
OK @dhiaayachi, here I am, I just located these files that need to be changed, if there are other files to change tell me:

- [ ] [agent/agent.go](https://github.com/hashicorp/consul/blob/main/agent/agent.go)
- [ ] [agent/check.go](https://github.com/hashicorp/consul/blob/main/agent/check.go)
- [ ] [agent/structs/structs.go](https://github.com/hashicorp/consul/blob/main/agent/structs/structs.go)
- [ ] [agent/structs/check_type.go](https://github.com/hashicorp/consul/blob/main/agent/structs/check_type.go)
- [ ] [agent/config/config.go](https://github.com/hashicorp/consul/blob/main/agent/config/config.go)
[...]

PR #12722, opened 2022-04-07
UDP check for service stanza #12221

fulviodenza:
Hi! This is the PR for the #12221 issue. UDP check for service stanza.
I tested timeout condition, critical condition, and passing condition using a mocked server in check_test.
Consul, as is, does not allow UDP check, after this, with a configuration as in the issue, it should allow UDP health check.
I decided to use the Dial function instead of the DialUDP function since gives an opportunity to easily include, in the future, udp4 and udp6 also.
Let me know if there is something to adjust.


fulviodenza:
Hi @Amier3 , It was a pleasure to work on this, I think some tests fail because the port is already bound. This is a problem I encountered on my laptop. I solved simply changing the port. When we start the mocked UDP server, I don't close it, I'd like to know how to do it, in this way we could also solve the already bound port.

agent/checks/check_test.go
```diff
@@ -1101,6 +1102,111 @@ func TestCheckTCPPassing(t *testing.T) {
...
+		_, remoteaddr, err := ser.ReadFromUDP(b)
+		log.Printf("Read a message from %v %s \n", remoteaddr, b)
+
+		if err != nil {
+			log.Fatalf("Error reading from UDP")
+		}
+		go sendResponse(ser, remoteaddr)
+	}
+
+}
```
dhiaayachi, line 1137: I think this should fix the mock server not stopping 
```suggestion
func mockUDPServer(ctx context.Context, network string, port int) {

	b := make([]byte, 1024)

	addr := fmt.Sprintf(`127.0.0.1:%d`, port)
	udpAddr, err := net.ResolveUDPAddr(network, addr)
	if err != nil {
		log.Fatal("Error resolving UDP address: ", err)
	}

	ser, err := net.ListenUDP("udp", udpAddr)
	if err != nil {
		log.Fatal("Error listening UDP: ", err)
	}
	defer ser.Close()
	chClose := make(chan interface{})
	wg := sync.WaitGroup{}

	wg.Add(1)
	go func() {
		for {
			log.Print("Waiting for UDP message")
			_, remoteaddr, err := ser.ReadFromUDP(b)
			log.Printf("Read a message from %v %s \n", remoteaddr, b)

			if err != nil {
				log.Fatalf("Error reading from UDP %s", err.Error())
			}
			sendResponse(ser, remoteaddr)
			select {
			case <-chClose:
				fmt.Println("cancelled")
				wg.Done()
				return
			default:
			}
		}
	}()
	select {
	case <-ctx.Done():
		fmt.Println("cancelled")
		close(chClose)
	}
	wg.Wait()
	return
}
```

agent/checks/check_test.go
```diff
@@ -1101,6 +1102,111 @@ func TestCheckTCPPassing(t *testing.T) {
...
+		if got, want := notif.State(cid), status; got != want {
+			r.Fatalf("got state %q want %q", got, want)
+		}
+	})
+}
+
+func TestCheckUDPTimeoutPassing(t *testing.T) {
+	t.Parallel()
+
+	go mockUDPServer(`udp`, `4243`)
```
dhiaayachi, line 1193: what about using `freeport.GetOne(t)` to avoid getting ports that are used by the environment.

agent/checks/check_test.go
```diff
@@ -1101,6 +1102,111 @@ func TestCheckTCPPassing(t *testing.T) {
...
+	expectUDPTimeout(t, `127.0.0.1:4243`, api.HealthPassing) // Should pass since timeout is handled as success, from specification
+}
+func TestCheckUDPCritical(t *testing.T) {
+	t.Parallel()
+
+	go mockUDPServer(`udp`, `4244`)
```
dhiaayachi, line 1199: what about using `freeport.GetOne(t)` to avoid getting ports that are used by the environment.

agent/checks/check_test.go
```diff
@@ -1101,6 +1102,111 @@ func TestCheckTCPPassing(t *testing.T) {
...
+	expectUDPStatus(t, `127.0.0.1:4241`, api.HealthCritical) // Should be unhealthy since we never connect to mocked udp server.
+}
+
+func TestCheckUDPPassing(t *testing.T) {
+	t.Parallel()
+
+	go mockUDPServer(`udp`, "4242")
```
dhiaayachi, line 1206: what about using `freeport.GetOne(t)` to avoid getting ports that are used by the environment.

dhiaayachi (changes requested):
Hi @fulviodenza,
thank you for working on this, I made a first pass of review and I have a few comments and suggestions.
About this specific failing test `TestRuntimeConfig_Sanitize`, you need to run that test with the `-update` flag so it update the golden file used to validate in the test.

fulviodenza:
Hey @dhiaayachi,
Thank you really much for the review! It was pretty clear and really formative! Hope now is good, if not, let me know, i also changed the comment for the UDP check to make it more informative.

fulviodenza:
Hi @dhiaayachi !
Any news about this PR? Let me know if I need to make further changes

agent/checks/check_test.go
```diff
@@ -1101,6 +1106,142 @@ func TestCheckTCPPassing(t *testing.T) {
...
+		}
+		if got, want := notif.State(cid), status; got != want {
+			r.Fatalf("got state %q want %q", got, want)
+		}
+	})
+}
+
+func TestCheckUDPTimeoutPassing(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
```
dhiaayachi, line 1217: I would recommend a Cancel request here so you can stop the mockUDPServer server when test is done

dhiaayachi, line 1217: I would recommend a Cancel context here so you can stop the mockUDPServer server when test is done

agent/checks/check_test.go
```diff
@@ -1101,6 +1106,142 @@ func TestCheckTCPPassing(t *testing.T) {
...
+	port := freeport.GetOne(t)
+	serverUrl := "127.0.0.1:" + strconv.Itoa(port)
+
+	go mockUDPServer(ctx, `udp`, port)
+	expectUDPTimeout(t, serverUrl, api.HealthPassing) // Should pass since timeout is handled as success, from specification
+}
+func TestCheckUDPCritical(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
```
dhiaayachi, line 1226: Same here about the Cancel context

dhiaayachi, line 1226: Same here about the Cancel context

agent/checks/check_test.go
```diff
@@ -1101,6 +1106,142 @@ func TestCheckTCPPassing(t *testing.T) {
...
+	notExistentPort := freeport.GetOne(t)
+	serverUrl := "127.0.0.1:" + strconv.Itoa(notExistentPort)
+
+	go mockUDPServer(ctx, `udp`, port)
+	expectUDPStatus(t, serverUrl, api.HealthCritical) // Should be unhealthy since we never connect to mocked udp server.
+}
+
+func TestCheckUDPPassing(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
```
dhiaayachi, line 1237: Same here about the Cancel context

dhiaayachi, line 1237: Same here about the Cancel context

agent/checks/check.go
```diff
@@ -698,6 +699,133 @@ func (c *CheckTCP) check() {
...
+				"check", c.CheckID.String(),
+				"error", err,
+			)
+			c.StatusHandler.updateCheck(c.CheckID, api.HealthCritical, err.Error())
+			return
+		}
+	} else if err == nil {
+		c.StatusHandler.updateCheck(c.CheckID, api.HealthPassing, fmt.Sprintf("UDP connect %s: Success", c.UDP))
+	}
+	conn.Close()
```
dhiaayachi, line 826: I think this conn.Close should go in defer in line 769, otherwise we are leaking a connection in case of an early return.

dhiaayachi, line 826: I think this conn.Close should go in defer in line 769, otherwise we are leaking a connection in case of an early return.

dhiaayachi (changes requested):
Hey @fulviodenza,
sorry for the delay, I was caught with other stuff. I did another pass and this is very close to be in a mergeable state, I left a few comments. 
I also noticed a test failure, I will take some time Friday to debug it.

Thank you again for this great contribution, ping me if you have any question or ready for another round of review.

fulviodenza:
Hey @dhiaayachi,

Thank you for the second pass, I solved these little fixes also, they were pretty easy, I noticed you wrote 2 times comments, some of them with little changes, like the first two, what is the difference between a cancel request and a cancel context? 

For what concerns the `defer conn.Close()` I think its ideal position is after the error check since if I get a timeout error, the defer will fail since the conn is not a thing and there is no leak if any other error exists because if it exists then the conn doesn't

Let me know about this observation, I'll also wait for the response about that failing test.

Thanks!

dhiaayachi:
Hey @fulviodenza 
I think this is ready now from code perspective. I added a missing piece (related to protobuf definition) and a changelog.
That said, I think we should add a test at the agent level to make sure UDP checks are threaded all the way from an API perspective.
you can look to tests that exercise TCP checks in `agent_endpoint_test.go`. I should be able to merge this after those tests are added. Let me know if you have any question about those.

Thank you again for the great work!

fulviodenza:
Hey @dhiaayachi !
I changed agent_endpoint_test.go but I literally copied the test with the TCP with UDP, I'm not sure this approach was correct. Anyway, all tests seem to pass now. There is only that conflict to resolve.

fulviodenza:
Hi @dhiaayachi, 
sorry I tried to resolve conflict in the pb but I think to have done something that make failed test. I tried to revert and still tests failed

dhiaayachi:
@fulviodenza I think I made it worse by doing a rebase 😆 . Don't worry I will fix that later today

fulviodenza:
@dhiaayachi resolving conflicts in proto files is never a good idea 😂

dhiaayachi:
> @dhiaayachi resolving conflicts in proto files is never a good idea 😂

yeah, I usually just accept others in auto generated files and then rerun the proto generation again (you can do that in consul running `make proto`)

fulviodenza:
Hey @dhiaayachi , I've seen the status of your review is "requested changes", is any other things I can do or we are waiting for the consul documentation to document this feature?

dhiaayachi:
Hey @fulviodenza,
I think that the code is complete but we are missing some tests. Checks can be create from multiple endpoints and here we are only testing the agent endpoint. If you look into [this](https://github.com/hashicorp/consul/blob/main/docs/service-discovery/health-checks.md) list you will see few other location from which we can create a check. I think we need to write a test for each of those entry points.

fulviodenza:
Ehi @dhiaayachi , here I am with tests on 

- catalog_endpoint
- agent_endpoint
- transaction api

and some changes to other structs to support UDP health check.
I'm having just an issue in txn_test.go:

That test fails because this:

```
Diff:
--- Expected
+++ Actual
@@ -185,3 +185,3 @@
TCP: (string) "",
-    UDP: (string) (len=7) "1.1.1.1",
+    UDP: (string) "",
GRPC: (string) "",
@@ -224,3 +224,3 @@
TCP: (string) "",
-    UDP: (string) (len=7) "2.2.2.2",
+    UDP: (string) "",
GRPC: (string) "",
Test:       	TestAPI_ClientTxn

```

In short it expects the udp string address but if you check at line 63 and 74 I put the udp string address in the HealthCheckDefinition.

fulviodenza:
Thanks for fixing the test! @dhiaayachi

fulviodenza:
Hey @dhiaayachi , any news about this PR? Is there something to change?

dhiaayachi:
Hey @fulviodenza, 
Sorry for not getting back to you sooner. I think this is complete from code perspective. We need to add some documentation around it tough, would you be able to add it? 

Mostly following what we have as documentation for TCP checks, adapted to UDP  in the following doc files:
https://github.com/hashicorp/consul/blob/15a6d150f96c0b6504384eef4dc8216859d1753b/website/content/docs/discovery/checks.mdx
https://github.com/hashicorp/consul/blob/15a6d150f96c0b6504384eef4dc8216859d1753b/website/content/api-docs/agent/check.mdx
https://github.com/hashicorp/consul/blob/15a6d150f96c0b6504384eef4dc8216859d1753b/website/content/docs/ecs/configuration-reference.mdx

Thank you again for the great work!!

fulviodenza:
Hey @dhiaayachi I enjoyed so much working on this! working on clean code as this is really enjoyable. 
I added some documentation but I think it needs a review from your perspective. I will continue following actively this PR waiting for other comments on docs files. (Cannot understand the go-test failing, I only changed docs files)

gfonseca-tc:
Great to see that you got it working folks! Looking forward to see this feature merged!

website/content/api-docs/agent/check.mdx
```diff
@@ -134,7 +134,7 @@ The table below shows this endpoint's support for
   to the `default` namespace. Added in Consul 1.7.0.
 
 - `Interval` `(string: "")` - Specifies the frequency at which to run this
-  check. This is required for HTTP and TCP checks.
+  check. This is required for HTTP, TCP and UDP checks.
```
trujillo-adam, line 137:
```suggestion
  check. This is required for HTTP, TCP, and UDP checks.
```
Oxford comma

website/content/api-docs/agent/check.mdx
```diff
@@ -238,6 +238,11 @@ The table below shows this endpoint's support for
   made to both addresses, and the first successful connection attempt will
   result in a successful check.
 
+- `UDP` `(string: "")` - Specifies a `UDP` to send datagrams against the value of `UDP`
+  (expected to be an IP or hostname plus port combination) every `Interval`. If
+  the datagram sending is successful or gets a timeout, the check is `passing`. 
+  Any other failing result is unsuccessful, the check is `critical`.
```
trujillo-adam, line 244:
```suggestion
- `UDP` `(string: "")` - Specifies a `UDP` IP address, hostname, port, or combination thereof. The check sends datagrams to the value specified at the interval specified in the `Interval` configuration. If the datagram is sent successfully or a timeout is returned, the check is set to the `passing` state.  The check is logged as `critical' if the datagram is sent unsuccessfully.
```
Just trying to be clearer about the actual value and how to understand the results.

website/content/docs/discovery/checks.mdx
```diff
@@ -85,6 +85,16 @@ There are several different kinds of checks:
...
 
+- `UDP + Interval` - These checks make the client send periodically UDP datagrams
+  to the specified IP/hostname and port, waiting `interval` amount of time 
+  between attempts (e.g. 30 seconds), if any response from the UDP server is 
+  received, the server is marked as healthy, 
+  any other case sets the status as `critical`.
+  By default, UDP checks will be configured with a request timeout of 10 seconds.
+  It is possible to configure a custom UDP check timeout value by specifying the
+  `timeout` field in the check definition. If any timeout on read exists, this will
+  still be considered healthy.
```
trujillo-adam, line 96:
```suggestion
- `UDP + Interval` - These checks direct the client to periodically send UDP datagrams
  to the specified IP/hostname and port. The duration specified in the `interval` field sets the amount of time 
  between attempts, such as `30s` to indicate 30 seconds. The check is logged as healthy if any response from the UDP server is received. Any other result sets the status to `critical`.
  The default interval for, UDP checks is `10s`, but you can configure a custom UDP check timeout value by specifying the
  `timeout` field in the check definition. If any timeout on read exists, the check is still considered healthy.
```

website/content/docs/ecs/configuration-reference.mdx
```diff
@@ -82,14 +82,15 @@ Defines the Consul checks for the service. Each check may contain these fields.
 | `h2pingUseTls` | `boolean` | optional | Specifies whether TLS is used for an h2ping check.  |
 | `header` | `object` | optional | Specifies a set of headers that should be set for HTTP checks. Each header can have multiple values.  |
 | `http` | `string` | optional | Specifies this is an HTTP check. Must be a URL against which request is performed every `interval`.  |
-| `interval` | `string` | optional | Specifies the frequency at which to run this check. Required for HTTP and TCP checks.  |
+| `interval` | `string` | optional | Specifies the frequency at which to run this check. Required for HTTP, TCP and UDP checks.  |
```
trujillo-adam, line 85:
```suggestion
| `interval` | `string` | optional | Specifies the frequency at which to run this check. Required for HTTP, TCP, and UDP checks.  |
```

website/content/docs/ecs/configuration-reference.mdx
```diff
@@ -82,14 +82,15 @@ Defines the Consul checks for the service. Each check may contain these fields.
...
 | `name` | `string` | optional | The name of the check.  |
 | `notes` | `string` | optional | Specifies arbitrary information for humans. This is not used by Consul internally.  |
 | `status` | `string` | optional | Specifies the initial status the health check. Must be one of `passing`, `warning`, `critical`, `maintenance`, or`null`. |
 | `successBeforePassing` | `integer` | optional | Specifies the number of consecutive successful results required before check status transitions to passing.  |
 | `tcp` | `string` | optional | Specifies this is a TCP check. Must be an IP/hostname plus port to which a TCP connection is made every `interval`.  |
-| `timeout` | `string` | optional | Specifies a timeout for outgoing connections in the case of a Script, HTTP, TCP, or gRPC check. Must be a duration string, such as `10s` or `5m`.  |
+| `udp` | `string` | optional | Specifies this is a UDP check. Must be an IP/hostname plus port to which UDP datagrams are sent every `interval`.  |
+| `timeout` | `string` | optional | Specifies a timeout for outgoing connections in the case of a Script, HTTP, TCP, UDP, or gRPC check. Must be a duration string, such as `10s` or `5m`.  |
```
trujillo-adam, line 93:
```suggestion
| `timeout` | `string` | optional | Specifies a timeout for outgoing connections. Applies to script, HTTP, TCP, UDP, and gRPC checks. Must be a duration string, such as `10s` or `5m`.  |
```

trujillo-adam (commented):
A few suggestions in the docs

fulviodenza:
> A few suggestions in the docs

Thanks for correction!

build-support/scripts/devtools.sh
```diff
@@ -165,10 +165,10 @@ function install_unversioned_tool {
     local install="$2"
```
dhiaayachi, line 165: The changes in here are related to a weird behaviour when running in CI. Circle-CI seems to fail setting the bash colonization options even if the env variable for colors print is set to 1. see [this](https://app.circleci.com/pipelines/github/hashicorp/consul/32228/workflows/cdfacfa6-5309-416f-a341-1363f3daec6f/jobs/698057)

website/content/api-docs/agent/check.mdx
```diff
@@ -237,6 +237,8 @@ The table below shows this endpoint's support for
   made to both addresses, and the first successful connection attempt will
   result in a successful check.
 
+- `UDP` `(string: "")` - Specifies a `UDP` IP address, hostname, port, or combination thereof. The check sends datagrams to the value specified at the interval specified in the `Interval` configuration. If the datagram is sent successfully or a timeout is returned, the check is set to the `passing` state.  The check is logged as `critical' if the datagram is sent unsuccessfully.
```
jkirschner-hashicorp, line 240: > Specifies a `UDP` IP address, hostname, port, or combination thereof.

@fulviodenza, @dhiaayachi, @trujillo-adam: Is this wording accurate? For example, can this string _just_ be a port? To me, this wording seems like any component can be used separately, and any combination of them is possible.

I think it would be clearer to use the same wording as used below: https://github.com/hashicorp/consul/pull/12722/files#diff-fde54c45fa092a8b19350f419572bdfa28505a986952361ca683d127ef7a9663R88-R89

There's also a small mistake: critical has ` on the left, but ' on the right.

I propose the following:
```suggestion
- `UDP` `(string: "")` - Specifies a `UDP` IP address/hostname and port.
The check sends datagrams to the value specified at the interval specified in the `Interval` configuration.
If the datagram is sent successfully or a timeout is returned, the check is set to the `passing` state.
The check is logged as `critical` if the datagram is sent unsuccessfully.
```

Thoughts?

jkirschner-hashicorp (changes requested):
Providing a "request changes" review until we close out the docs comments. I left one additional comment we should agree on.

build-support/scripts/devtools.sh
jkirschner-hashicorp: @dhiaayachi : it looks like this branch now has merge conflicts with this file?

dhiaayachi: @jkirschner-hashicorp it should be fixed now

jkirschner-hashicorp (approved):

trujillo-adam (approved):

PR #12727, opened 2022-04-08
add more labels to RequestRecorder

FFMMM:
### overview

This PR adds more labels to the RequestRecorder. The following are mandatory labels:

* `leader` = `{true | false | unreported}`

The optional labels, given duck typing, are:

* `blocking`  = `{true | false}`
* `target_datacenter` = `{string}`
* `locality` = `{forwarded | local`} 

These labels allow further observability for the new rpc metric(s).

#### todos left:

* [x] docs update (separate PR) --> https://github.com/hashicorp/consul/pull/12757
* [x] changelog
* [ ] ent shadow PR/ make sure it merges


agent/consul/server.go
```diff
@@ -425,6 +404,29 @@ func NewServer(config *Config, flat Deps, publicGRPCServer *grpc.Server) (*Serve
...
+		rpcServer = rpc.NewServer()
+		insecureRPCServer = rpc.NewServer()
+	} else {
+		rpcServer = rpc.NewServerWithOpts(rpc.WithServerServiceCallInterceptor(flat.GetNetRPCInterceptorFunc(recorder)))
+		insecureRPCServer = rpc.NewServerWithOpts(rpc.WithServerServiceCallInterceptor(flat.GetNetRPCInterceptorFunc(recorder)))
+	}
+
+	s.rpcRecorder = recorder
+	s.rpcServer = rpcServer
+	s.insecureRPCServer = insecureRPCServer
```
FFMMM, line 428: keep me honest here please, but this should be 100% ok and no racing should occur given that we are still just setting up the server struct here.

agent/rpc/middleware/interceptors.go
```diff
@@ -33,33 +33,79 @@ var OneTwelveRPCSummary = []prometheus.SummaryDefinition{
...
 	r.RecorderFunc(metricRPCRequest, float32(elapsed), labels)
-	r.Logger.Trace(requestLogName,
-		"method", requestName,
-		"errored", respErrored,
-		"request_type", reqType,
-		"rpc_type", rpcType,
-		"elapsed", elapsed)
+
+	labelsArr := flattenLabels(labels)
+	r.Logger.Trace(requestLogName, labelsArr...)
```
FFMMM, line 70: i did my best to not add another helper func here -- let me know if there's a easier method to enumerate keys, values as `[]interface{}` in go 🙈

agent/rpc/middleware/interceptors_test.go
```diff
@@ -57,75 +59,195 @@ func (wr writeRequest) IsRead() bool {
...
 
-	r.Record("B.A", RPCTypeNetRPC, start, readRequest{}, true)
+var testCases = []testCase{
+	{
+		name:        "simple ok",
+		description: "This is a simple happy path test case. We check for pass through and normal request processing",
+		requestName: "A.B",
+		requestI:    struct{}{},
+		rpcType:     "test",
+		errored:     false,
```
eculver, line 108: If you wanted to make this a little more terse, you could omit this `errored` field. Also, I don't see this field ever flipped to `true` and asserted that it makes its way through the recorder. Given that `false` is the zero value for `bool`, it's probably worth adding another case for it.

agent/consul/server.go
```diff
@@ -425,6 +404,29 @@ func NewServer(config *Config, flat Deps, publicGRPCServer *grpc.Server) (*Serve
...
 	}
 
+	var recorder *middleware.RequestRecorder
+	if flat.NewRequestRecorderFunc != nil {
+		recorder = flat.NewRequestRecorderFunc(serverLogger, s.IsLeader, s.config.Datacenter)
+	} else {
+		return nil, fmt.Errorf("cannot initialize server without an RPC request recorder provider")
+	}
+	if recorder == nil {
+		return nil, fmt.Errorf("cannot initialize server without a non nil RPC request recorder")
```
eculver, line 414: I think it may be a little easier to read if there is not a double-negative in the message. What do you think?
```suggestion
		return nil, fmt.Errorf("cannot initialize server with a nil RPC request recorder")
```

agent/consul/server.go
eculver: Seems fine to me. You could even remove the variable declaration and just assign directly to `s` too:
```
if flat.GetNetRPCInterceptorFunc == nil {
		s.rpcServer = rpc.NewServer()
		s.insecureRPCServer = rpc.NewServer()
} else {
		...
}
```

eculver (approved):
LGTM, just a few non-blocking suggestions.

agent/consul/server_test.go
```diff
@@ -1344,11 +1347,13 @@ func TestServer_RPC_MetricsIntercept(t *testing.T) {
 			{Name: "errored", Value: "false"},
 			{Name: "request_type", Value: "read"},
 			{Name: "rpc_type", Value: "test"},
+			{Name: "server_role", Value: "unreported"},
```
eculver, line 1350: Shouldn't this get wired up with the test's `(*Server).IsLeader` func and thus not be `"unreported"`?

agent/consul/server.go
```diff
@@ -425,6 +404,29 @@ func NewServer(config *Config, flat Deps, publicGRPCServer *grpc.Server) (*Serve
 		fsm:                     newFSMFromConfig(flat.Logger, gc, config),
 	}
 
+	var recorder *middleware.RequestRecorder
+	if flat.NewRequestRecorderFunc != nil {
+		recorder = flat.NewRequestRecorderFunc(serverLogger, s.IsLeader, s.config.Datacenter)
```
dhiaayachi, line 409: I'm not sure what are the implication of passing `IsLeader` to the request recorder in here. IsLeader go as deep as atomically reading the raft state to be able to tell and that could have sides effects that are hard to reason about (specially performance effects)
I would recommend to have a separate state for the recorder that we can update when we acquire or loose leadership.

agent/consul/server.go
dhiaayachi: Yes this should still be safe, but if we remove the dependency to isLeader we can revert it as it was.

agent/rpc/middleware/interceptors.go
```diff
@@ -77,6 +123,30 @@ func requestType(req interface{}) string {
...
+		} else {
+			return "follower"
+		}
+	}
+
+	// This logical branch should not happen. If it happens
+	// it means that we have not plumbed down a way to verify
+	// whether the server handling the request was a leader or not
+	return "unreported"
+}
```
dhiaayachi, line 139: Actually the node can have multiple states. `IsLeader` will report true if the node is a leader. If it report false it mean the node have one of the following states:
- Follower
- Candidate
- Shutdown

agent/consul/server_test.go
```diff
@@ -1344,11 +1347,13 @@ func TestServer_RPC_MetricsIntercept(t *testing.T) {
...
 		}
 
 		key := keyMakingFunc(middleware.OneTwelveRPCSummary[0].Name, expectedLabels)
 
 		if _, ok := storage[key]; !ok {
+			// the compound key will look like: "rpc+server+call+Status.Ping+false+read+test+unreported"
```
dhiaayachi, line 1356: maybe we can simplify the content of the key and use assertion to verify the values in here. I think in the test you can most of the time infer the right entry without having all of this in the key.

dhiaayachi (commented):
Thank you for working on this @FFMMM. I have few comments related to the use of IsLeader mostly.

agent/consul/server.go
FFMMM: thanks for this suggestion 💯 

i'm going to apply locally to resolve the tests and all other feedback in one go

agent/consul/server.go
FFMMM: > I would recommend to have a separate state for the recorder that we can update when we acquire or loose leadership.

Would you mind chatting thru this some more with me? When I thought about this, I started to think that it can get too complex keeping tabs on the leadership state for a server but I may be wrong here 🤔 

---

On another note, I see this "pattern" (of passing the func/ a smaller interface) being loosely used in `gateway` code here:

https://github.com/hashicorp/consul/blob/09d61e643f4faacb6eb19e2252f8641549d9e7c3/agent/consul/gateway_locator.go#L272-L277

agent/rpc/middleware/interceptors.go
FFMMM: Hm, wondering if we could change the label around `IsLeader` then instead:

```
{
"is_leader": {true | false}
}
```

agent/consul/server_test.go
FFMMM: I think we definitely needs the name of the rpc call (otherwise the key gets overwritten in the map).

The labels help distinguish the happy path for us so  we don't "fail silently" on the calls and mark the test as passed.

agent/rpc/middleware/interceptors.go
```diff
@@ -33,33 +33,79 @@ var OneTwelveRPCSummary = []prometheus.SummaryDefinition{
...
 	elapsed := time.Since(start).Milliseconds()
 	reqType := requestType(request)
+	serverRole := r.getServerRole()
 
 	labels := []metrics.Label{
 		{Name: "method", Value: requestName},
 		{Name: "errored", Value: strconv.FormatBool(respErrored)},
 		{Name: "request_type", Value: reqType},
 		{Name: "rpc_type", Value: rpcType},
+		{Name: "server_role", Value: serverRole},
```
mkeeler, line 61: nit/bikeshed:

```suggestion
		{Name: "leader", Value: strconv.FormatBool(isLeader)},
```

RPC metrics will only be emitted by servers so there is no need to prefix the metric name with `server`. Secondly since there are only two possible values of leader and follower I think the metric might be more intuitive for someone exploring their grafana or prometheus to have the name be `leader` and the value be a boolean.

FFMMM: this makes a lot of sense!  and that's what I am changing for a new commit ⌨️ 

I think it would be nice to keep `unreported`, and subsequently keep this as a tri state 

```
leader: {`true`, `false`, `unreported`}
```

for a defense in depth kind of approach. If the function is not passed to the `RequestRecorder`, I wouldn't think it useful or accurate to report `leader: false`

agent/consul/server.go
FFMMM: Per offline chats, we are going to keep `IsLeader` func arg as is for now

agent/consul/server.go
FFMMM: We were chatting offline and believe that the underlying `raft.getState()` per RPC call would not adversely impact performance.

We are still interested in coming up with a different way of the server telling the request recorder about its leadership state. I will backlog an issue for this.

agent/rpc/middleware/interceptors.go
FFMMM: reworked this func for a tri-state

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/635231.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit a46bbe892d56e42a3d58251785cbc3055389d2a9 onto `release/1.12.x` succeeded!

Issue #11740, opened 2021-12-04
Replace function registry pattern


dnephin:
There are a number of packages (see below for the list) where we use a registry to store functions in a map. In 05d5ec480497a0b0952eba2de5305e611cfb6d73 we replaced one of those instances with a different pattern. The new pattern was simply a function that accepted a list of other functions.

The new pattern of simply calling functions directly has a few advantages over the "function registry" pattern:
* it is much easier to trace the code. The function registry pattern requires many hops (init  -> register -> map definition -> map usage), with a direct function call it is simply one hop (function -> caller).
* it is easier to write a reasonable test when the code being tested accepts its dependencies as inputs, instead of them being declared statically.
* in some cases this pattern forces us to keep all the functions in the same package, or even as methods on the same struct (because the registry if of unbound methods instead of functions). This leads to very large packages and sub-optimal code layout which cause significant developer productivity issues.

This issue proposes we addresses these problems by replacing each of these instances of the "function registry" pattern with direct function calls, similar to 05d5ec480497a0b0952eba2de5305e611cfb6d73.

We can see this pattern in the following places:

* CLI commands - `command/commands_oss.go`
* HTTP handlers - `agent/http_register.go`
* FSM operations - `agent/consul/fsm/commands_oss.go`
* FSM snapshot persist functions - `agent/consul/fsm/snapshot_oss.go`
* FSM snapshot restore functions - `agent/consul/fsm/snapshot_oss.go`
* ACL auth methods  - `agent/consul/authmethod`
* RPC handlers - `agent/consul/server_register.go`
* Watch functions - `api/watch/funcs.go`

PR #12729, opened 2022-04-08
Replace CLI command registry with a new pattern.

Mathew-Estafanous:
issue #11740

Replaces the current CLI command registry pattern with the newer approach recommended in #11740. 

This PR is my understanding of the intended new pattern. I wasn’t too sure what to name a few types or functions (ie. ``Entry``) so if there are better alternatives then I’m more than wiling to hear them.


Mathew-Estafanous:
I’m not sure what the failing ``check-generated-protobuf`` is caused by and whether or not it’s related to my change.

command/registry.go
```diff
@@ -14,35 +121,141 @@ import (
 // Factory is a function that returns a new instance of a CLI-sub command.
 type Factory func(cli.Ui) (cli.Command, error)
 
-// Register adds a new CLI sub-command to the registry.
-func Register(name string, fn Factory) {
-	if registry == nil {
-		registry = make(map[string]Factory)
-	}
+// Entry is a function that returns a command's name and a Factory for that command.
+type Entry func() (string, Factory)
```
kisunji, line 125: Can you see if `Factory` and `Entry` can be unexported? Not sure if they're used outside this package

kisunji: Also, I think if we make `Entry` a struct instead we can simply pass around structs than funcs

kisunji (commented):
Looks good! Just a few clean up comments

command/registry.go
Mathew-Estafanous: oh ya, a struct makes a lot more sense. I don’t know what I was thinking 😅

Mathew-Estafanous:
I’ve made the changes, but I’m honestly not too sure what my change did to make the tests fail. I ran a few of the failing unit tests on my end and they pass.

kisunji:
> I’ve made the changes, but I’m honestly not too sure what my change did to make the tests fail. I ran a few of the failing unit tests on my end and they pass.

Oh no problem, we've been having issues with flaky tests that fail on CircleCI. I'll try to rerun them to get them passing

kisunji (approved):
LGTM! As with the other PR, I'm going to need to make some enterprise-side changes so it might take some time to merge

kisunji:
@Mathew-Estafanous I have a request for you when you have time;

Right now `CommandsFromRegistry` sets up all the commands and returns the map which is used as a registry. It works nicely in OSS, but in Enterprise we have a few more commands that we need to hook up. Instead of directly returning the map, could you refactor so we can add more stuff to the map? Something like this [commit](https://github.com/hashicorp/consul/commit/05d5ec480497a0b0952eba2de5305e611cfb6d73#diff-726beb4fa62534d1bbe133827978023ae286777ad9370c567a6f6c3d36b7f924R41).
You can add a no-op function for enterprise in a file like `registry_oss.go`.

Mathew-Estafanous:
Hey @kisunji 

Just pushed the changes. Sorry that it took some time for me to get this. Let me know if I misunderstood your request.

kisunji:
This is great, thanks! Helps on the enterprise side a lot

PR #12740, opened 2022-04-11
ui: Don't automatically move rz read-replicas out of the rz

johncowen:
There is a case where a read-replica could be configured to be in a redundancy zone. We originally decided to use the frontend to display this configuration with the read-replica as not being in the redundancy zone. Since then we've decided that we should show the users configuration rather than display that different to what it actually is rather than fix it up on the frontend.

This PR removes the little hack we made to fix up the config if this case happens.


evrowe (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/636412.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 4c8c163a48ea572f31bee650308a1343709ade6c onto `release/1.12.x` succeeded!

PR #12743, opened 2022-04-11
Improve Vault Connect CA docs

jkirschner-hashicorp:
Includes:
- Provide explicit guidance on use with Vault Namespaces
- Include Consul 1.12 support for using different Vault Namespaces for the
  root and intermediate PKI secrets engines
- Improve scannability by adding new subheadings and call-out text

[Doc preview on Vercel](https://consul-g7qso2imx-hashicorp.vercel.app/docs/connect/ca/vault)

## To-Do

- [ ] Confirm with @markan whether read permission on `/sys/mounts` is still required

### DIFF
--- a/website/content/docs/connect/ca/vault.mdx
+++ b/website/content/docs/connect/ca/vault.mdx
@@ -19,7 +19,7 @@ Please read the [certificate management overview](/docs/connect/ca)
 page first to understand how Consul manages certificates with configurable
 CA providers.
 
--> **NOTE**: A Learn [tutorial](https://learn.hashicorp.com/tutorials/consul/vault-pki-consul-connect-ca?in=consul/vault-secure) is available to help you configure Vault as the Consul Connect service mesh Certification Authority.
+-> **Tutorial**: A Learn [tutorial](https://learn.hashicorp.com/tutorials/consul/vault-pki-consul-connect-ca?in=consul/vault-secure) is available to help you configure Vault as the Consul Connect service mesh Certification Authority.
 
 ## Requirements
 
@@ -52,7 +52,7 @@ connect {
   ca_provider = "vault"
   ca_config {
     address = "http://localhost:8200"
-    token = "..."
+    token = "<Vault token>"
     root_pki_path = "connect-root"
     intermediate_pki_path = "connect-intermediate"
   }
@@ -68,7 +68,7 @@ connect {
   "Provider": "vault",
   "Config": {
     "Address": "http://localhost:8200",
-    "Token": "<token>",
+    "Token": "<Vault token>",
     "RootPKIPath": "connect-root",
     "IntermediatePKIPath": "connect-intermediate"
   }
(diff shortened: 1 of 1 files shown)


website/content/docs/connect/ca/vault.mdx
```diff
@@ -114,36 +185,7 @@ The configuration options are listed below.
     default mount path `/var/run/secrets/kubernetes.io/serviceaccount/token` if the `jwt` parameter
     is not provided.
 
-
```
jkirschner-hashicorp, line 117: This content was all moved up into "Vault Resource Locations" subsection

website/content/docs/connect/ca/vault.mdx
```diff
@@ -168,10 +210,6 @@ The configuration options are listed below.
 - `TLSSkipVerify` / `tls_skip_verify` (`bool: false`) - Specifies if SSL peer
   validation should be enforced.
 
-- `Namespace` / `namespace` (`string: <optional>`) - The Vault Namespace that
```
jkirschner-hashicorp, line 171: This content was moved up into "Vault Resource Locations" subsection

website/content/docs/connect/ca/vault.mdx
```diff
@@ -279,6 +317,137 @@ path "/connect_inter/*" {
...
+
+The [`Namespace`](#namespace) option is used as the `X-Vault-Namespace` header
+for all API requests made by this CA provider to Vault.
+All Vault namespaces included in paths are relative to the value of this header.
+For more details on how paths to namespaced resources can be specified,
+refer to the [Vault Namespace usage documentation](https://www.vaultproject.io/docs/enterprise/namespaces#usage).
+
+The PKI path options ([`RootPKIPath`](#rootpkipath) and [`IntermediatePKIPath`](#intermediatepkipath)
+both support including a Vault Namespace in the path.
+However, the Vault authentication options ([`Token`](#token) and [`AuthMethod`](#authmethod))
```
jkirschner-hashicorp, line 337: @markan : my conclusion from reading the [auth method code](https://github.com/hashicorp/consul/pull/11573/files#diff-7bf30611a760296e2e6ffcd7fe955b0eb8317fe0a43c470b8b52d311af4a2515R744) is that namespaces cannot be provided in the path for that configuration. `AuthMethod.MountPath` seems to be injected to API calls with:
```
loginPath = fmt.Sprintf("auth/%s/login", authMethod.MountPath)
```

... when the Vault Namespace would have to come before `auth/`. So, for both Vault tokens and auth methods, the namespace must be provided in the `Namespace` option rather than in the path. I tried to document that clearly in this section.

website/content/docs/connect/ca/vault.mdx
```diff
@@ -279,6 +317,137 @@ path "/connect_inter/*" {
...
+}
+```
+
+</CodeBlockConfig>
+
+</CodeTabs>
+
+-> **Why the Namespace option isn't used**:
+  In this case, the [`Namespace`](#namespace) option cannot be set to `ns-consul/` because
+  the Vault `Token` is in the root Vault Namespace.
```
trujillo-adam, line 397:
```suggestion
The [`Namespace`](#namespace) option is not included in this case because the Vault `Token` is in the root Vault namespace. As a result, the Consul `Namespace` cannot be set to `ns-consul/`. 
```
I think this is directly related and probably doesn't need a callout box

website/content/docs/connect/ca/vault.mdx
```diff
@@ -279,6 +317,137 @@ path "/connect_inter/*" {
...
+
+### Example 2: Token in Vault Namespace, PKI Paths in Different Namespaces
+
+Consider the following setup of Vault resources and their namespaces:
```
trujillo-adam, line 401: See previous comment about streamlining these examples.

website/content/docs/connect/ca/vault.mdx
```diff
@@ -279,6 +317,137 @@ path "/connect_inter/*" {
...
+do not allow specifying a Vault Namespace in their configuration.
+
+**If using an authentication mechanism that exists within a non-root Vault Namespace**:
+- set the [`Namespace`](#namespace) option to the Vault authentication mechanism's Namespace, and
+- `RootPKIPath` and `IntermediatePath` must be in that same Namespace or a child Namespace;
+  the relative path to a child Namespace must be included in the path
+
+### Example 1: PKI Paths in Same Namespace
+
+Consider the following setup of Vault resources and their namespaces:
```
trujillo-adam, line 347:
```suggestion
The following example demonstrates how to configure Consul service mesh to use Vault as the CA when the Vault PKI paths are in the same namespace.

#### Vault Setup
```

website/content/docs/connect/ca/vault.mdx
```diff
@@ -279,6 +317,137 @@ path "/connect_inter/*" {
...
+
+| Vault Resource                       | Belongs to Vault Namespace | Resource Name              |
+| ------------------------------------ | -------------------------- | -------------------------- |
+| `Token`                              | root                       | N/A                        |
+| `RootPKIPath` (Primary CA)           | `ns-consul`                | `connect-root`             |
+| `IntermediatePKIPath` (Secondary CA) | `ns-consul`                | `connect-intermediate-dc1` |
+
+The correct Connect CA configuration for this setup is:
```
trujillo-adam, line 355:
```suggestion
#### Consul Service Mesh Configuration

The following `connect` configuration configures Consul service mesh to use Vault as the CA: 
```
I had started adding suggestions to change the language a bit, but I think we can streamline these examples by adding a more descriptive introduction to the examples and using headings to indicate what is happening in each subsection.

trujillo-adam (commented):
It might be easier to go over these suggestions in person.

website/content/docs/connect/ca/vault.mdx
markan: You are correct. Also we should note that the Namespace field doesn't affect this in 1.11.0-1.11.5 because of a bug. This is has been corrected in 1.12.0 and backported to the 1.11.x branch (https://github.com/hashicorp/consul/pull/12793). No change is needed, noting this mostly for posterity.

jkirschner-hashicorp: I'll probably include this caveat in the 1.11 version of the docs (backport to that).

I need to rewrite these docs anyways given what we're learned / the config UX enhancements we're in the process of making in 1.12.x for Vault namespaces.

markan (approved):

PR #12771, opened 2022-04-13
ui: Disable licensing tab if the UI is running via HCP

johncowen:
See title ^

Currently this is guarded by a `go-template` hardcoded `false` which should made dynamic based on whether Consul is running in HCP or not. I'll note where the hardcoded boolean is in self-review notes


ui/packages/consul-ui/lib/startup/templates/body.html.js
```diff
@@ -66,6 +66,9 @@ ${
 {{if .NamespacesEnabled}}
   <script src="${rootURL}assets/consul-nspaces/routes.js"></script>
 {{end}}
+{{if true}}
```
johncowen, line 69: ^ This `true` here needs to be replaced with something like `.HcpEnabled`, or anything that can be flipped between true/false via the backend depending on whether we are in HCP or not

ui/packages/consul-ui/lib/startup/templates/body.html.js
```diff
@@ -66,6 +66,9 @@ ${
 {{if .NamespacesEnabled}}
   <script src="${rootURL}assets/consul-nspaces/routes.js"></script>
 {{end}}
+{{if false}}
```
johncowen, line 69: ^ This `false` here needs to be replaced with something like .HcpEnabled, or anything that can be flipped between true/false via the backend depending on whether we are in HCP or not

evrowe: Updated to use the env var supplied by the backend

meirish (approved):
I don't know the details of env var -> go template value, but the code changes look good to me assuming that part plumbs through properly.

evrowe:
Converting this PR back to a draft until we can perform end to end validation on these changes, which we are in the process of doing now.

johncowen:
See https://github.com/hashicorp/consul/pull/13165, for a full e2e manual test that PR needs to be in ENT for the licensing tab to show up. I've tested all of this manually/short term by building consul with https://github.com/hashicorp/consul/pull/13165 and this PR applied then running

```sh
CONSUL_HCP_ENABLED=true consul agent -dev
```

And confirming that the HCP-only routing configuration appears (this configuration when loaded disables the licensing page)

Screenshot 2022-05-20 at 11 35 54

I also confirmed that just running
```sh
consul agent -dev
```

Removes the above script tag.

Once the backend PR is merged here we can take this one out of draft and revisit.

randallmorey (approved):

ui/packages/consul-ui/app/templates/dc/show.hbs
```diff
@@ -13,9 +13,9 @@ as |route|>
 
 {{#let
     (from-entries (array
-      (array 'serverstatus' true)
+      (array 'serverstatus' (compute (fn route.exists 'serverstatus')))
       (array 'cataloghealth' false)
-      (array 'license' (can 'read license'))
+      (array 'license' (compute (fn route.exists 'license')))
```
evrowe, line 18: How will this interact with changes from #12795?

Can we ensure that the changes from that PR (to hide the license tab in OSS) are compatible with these so both use cases are accounted for?

johncowen: Oh good call! Yeah I think we spotted that bug and fixed it after we PRed this, so something definitely needs adding in here. I think just `and`ing the `can 'read license'` onto here will do for the moment. But ideally we should make another feature flagged, side loaded 'app' `consul-licensing` so licensing things only load if you have that enabled. Definitely would rather get the backend changes merged in and up onto here before finally merging this one.

johncowen: Oh sorry, did have my head screwed on right 😬 

`route.exists` also checks `route.allowed`

https://github.com/hashicorp/consul/pull/12771/files#diff-2dd4d29b2f108582aa031c6b6c2cd758cc5bddcc34f56c0f230d1d2d812686a7R67-R83

which checks these abilities in our router:

https://github.com/hashicorp/consul/blob/fd7a403e1190ef7f8cfc35a13bf388945d94e37e/ui/packages/consul-ui/vendor/consul-ui/routes.js#L30-L34

So for `route.exists` to be true the route has to exist in the configuration (not the case in HCP) _and_ all of the abilities in the router configuration need to pass (`read license` equates to `operator:read`)

The backend changes have been merged now so I'm just re-verifying all of this.

johncowen:
Verified with the backend merged changes, all looks good to me. I'd be happy to merge this, if @evrowe is.

ui/packages/consul-ui/app/templates/dc/show.hbs
evrowe: Got it, and just to make sure I understand correctly, this logic also ensures the license will not display for OSS because `operator:read` is only true for ENT?

Or is it that having `'read license'` listed within the `abilities` in the router just reads from `app/abilities/license.js`, which covers the OSS/ENT concern for us?

johncowen: Neither of those quite cover all of it if I understand those correctly.


> because operator:read is only true for ENT

^ This bit isn't quite right, `operator:read` can be true in OSS or ENT, but aswell as checking `operator:read` in the ability we also check for ENT (which is what jake added)

Maybe this helps a little more, there are two levels of "enabling" the license route here:

1. The routing config is only sourced/loaded if `.HCPEnabled` is `false`. If its `true` (ie we are in HCP) the routing config for licensing isn't even loaded, it's almost as if the code was never written, i.e. this feature does not exist at all. This uses the JSON data format routing config we've been talking about lately.
2. If `.HCPEnabled` is `false` then the routing config for licensing _is_ loaded, but you can only see it if you have the the `read license` ability. So this bit that you said "having 'read license' listed within the abilities in the router just reads from app/abilities/license.js" <- thats correct, but this covers the authorization (am I allowed to see this) concern, _and_ the feature (ENT vs OSS) concern, because in that `License.canRead` ability we use the CONSUL_NSPACES_ENABLED feature flag/env var as well as checking `operator:read`

(Small note, ideally we would use also only load the JSON data format routing config for ENT also, similar to the HCP feature flag, but order of things made it not be like this, little improvement there if we ever want to do that)

evrowe: Got it, thanks for the explanation!

evrowe (approved):
Looks good, let's merge it in

PR #12775, opened 2022-04-13
set vault namespaces on vault client prior to logging in with the vault auth method

jmurret:
The namespace needs tobe set before trying to call the auth method.  Otherwise it will hit the general auth handler for most routes and look for a client token.  It returns this error:

```
2022-04-13T16:59:50.366Z [ERROR] connect.ca: Failed to initialize Connect CA: error="error configuring provider: Error making API request.

URL: PUT https://vault-cluster.private.vault.a718f64a-c2d8-4d80-a330-54bce161cf25.aws.hashicorp.cloud:8200/v1/auth/kubernetes/login
Code: 400. Errors:

* missing client token"
```

### DIFF
--- a/agent/connect/ca/provider_vault.go
+++ b/agent/connect/ca/provider_vault.go
@@ -104,6 +104,15 @@ func (v *VaultProvider) Configure(cfg ProviderConfig) error {
 		return err
 	}
 
+	// We don't want to set the namespace if it's empty to prevent potential
+	// unknown behavior (what does Vault do with an empty namespace). The Vault
+	// client also makes sure the inputs are not empty strings so let's do the
+	// same.
+	if config.Namespace != "" {
+		v.namespace = config.Namespace
+		client.SetNamespace(config.Namespace)
+	}
+
 	if config.AuthMethod != nil {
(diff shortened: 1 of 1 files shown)


jkirschner-hashicorp:
We should probably backport this to 1.12 (in which Vault namespace and auth method support was introduced).

The reviewer should also consider any potential impacts on backwards-compatibility and how to deal with them. For example, if the current behavior allows use of a Vault token and/or auth method in the Vault root namespace together with root/intermediate PKI engines in the specified `Namespace`... what happens if we merge this PR? Would that no longer work (the Vault token or auth method would need to be in the specified `Namespace` rather than the Vault root namespace)?

jmurret:
@jkirschner-hashicorp that's a great question

For added context, at the moment, this is just a PR to @markan 's branch and is really more for his consideration as he is doing existing work, in a "if you also see these issues, this worked for me." 

I am still also trying to set up an acceptance test in kubernetes for a straightforward namespace scenario where there are no nested namespaces.  Just someone operating in one namespace. Similar to what occurs with admin namespace with HCP Vault.  If I can get an acceptance test for that working, I will circle back and see what other scenarios we might add as acceptance tests.

PR #12807, opened 2022-04-15
acl: Clarify node/service identities must be lowercase

blake:
Modify ACL error message for invalid node/service identities names to clearly state only lowercase alphanumeric characters are supported.

### DIFF
--- a/agent/consul/acl_endpoint.go
+++ b/agent/consul/acl_endpoint.go
@@ -770,7 +770,7 @@ func (a *ACL) tokenSetInternal(args *structs.ACLTokenSetRequest, reply *structs.
 			return fmt.Errorf("Service identity %q cannot specify a list of datacenters on a local token", svcid.ServiceName)
 		}
 		if !isValidServiceIdentityName(svcid.ServiceName) {
-			return fmt.Errorf("Service identity %q has an invalid name. Only alphanumeric characters, '-' and '_' are allowed", svcid.ServiceName)
+			return fmt.Errorf("Service identity %q has an invalid name. Only lowercase alphanumeric characters, '-' and '_' are allowed", svcid.ServiceName)
 		}
 	}
 	token.ServiceIdentities = dedupeServiceIdentities(token.ServiceIdentities)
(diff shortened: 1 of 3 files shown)


agent/consul/acl_endpoint.go
```diff
@@ -770,7 +770,7 @@ func (a *ACL) tokenSetInternal(args *structs.ACLTokenSetRequest, reply *structs.
 			return fmt.Errorf("Service identity %q cannot specify a list of datacenters on a local token", svcid.ServiceName)
 		}
 		if !isValidServiceIdentityName(svcid.ServiceName) {
-			return fmt.Errorf("Service identity %q has an invalid name. Only alphanumeric characters, '-' and '_' are allowed", svcid.ServiceName)
+			return fmt.Errorf("Service identity %q has an invalid name. Only lowercase alphanumeric characters, '-' and '_' are allowed", svcid.ServiceName)
```
FFMMM, line 773: In the past, engineering talked about how some practitioners may do some string processing on these errors.

This is more of a nit, but if we have a strong indication that this may break some folks's checks, we may want to call it out in the upgrade path.

What do you think?

FFMMM (approved):
👍🏼 just a small nit

agent/consul/acl_endpoint.go
Jeff-Apple: It is standard practice for log monitoring systems (like Splunk) to match known strings to identify specific error messages and parse the details out (like service name).

At a previous company, I saw this cause some major problems for customers when we changed some of our log messages. Alerts didn't get triggered when things happened, which resulted in outages not getting responded to when they should have been.

boxofrad: I totally agree in principle, but also (perhaps naively) wonder whether the risk is low enough with this particular error message for it to be outweighed by the UX benefits of making things clearer?

Ideally, customers wouldn't have to match on error messages, but that's obviously a much bigger topic (our own client agents even have to because of `net/rpc'`s lack of support for richer error information).

kisunji: Agreed with @boxofrad here; these instances are more of a one-off error message in response to an input vs a diagnostic error message that someone might script around so I'm not too concerned about breaking things.

kisunji (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/647629.

PR #5597, opened 2019-04-03
TLS cert provisioning

i0rek:
## Intro

This PR introduces automatic certificate provisioning for consul clients to communicate securely via RPC. The goal is to not provision client certificate and key, but get that from the servers instead. Depending on the client configuration, the setup is as secure as manually provisioning certificates or less secure.

* **secure**: ACL enabled, CA provided, `verify_server_hostname` enabled
* **less secure**: ACL disabled **or** CA and `verify_server_hostname` not provided
* **insecure**: No ACL enabled, no CA provided

Even with an insecure configuration, the setup ends up being more secure because TLS is enabled nevertheless.

### New RPCType

When calling `AutoEncrypt.Sign`, the client has a CA at best, maybe not even that. But this endpoint has to have TLS enabled, no matter what. Since this is different than every other existing RPC a new RPCType was necessary: `RPCTLSInsecure`. On the server side every TLS connection is accepted and no client cert is required. The client will establish the TLS connection even though it might not have the server CA and certainly doesn't have a client cert.

### New RPCEndpoint

There is a new Endpoint `AutoEncrypt.Sign` which returns the combined result of `ConnectCA.Roots` and `ConnectCA.Sign`. This is the only endpoint so far that always requires TLS and that accepts TLS connections without a client certificate.

### Client Startup

If `auto_encrypt.tls` is set on the client, it will retry indefinitely to get certificates from the server. If it fails, it will never continue startup. Because the operator set the configuration, we assume TLS is wanted and the client won't proceed without it. The client does not rely on discovering servers with serf, but will use every address in `-join` and `-retry-join`. That means one of these addresses needs to be a server in order to allow the client to fetch certificates.

Also the Consul startup banner now announces that the agent is about to start instead of the successful start. I had to change that in order to be able to display the attempts to reach the `AutoEncrypt` endpoint.

### Prerequisites

TLS needs to be turned on for RPC on the server. (has now its own warning, so that this configuration is invalid)

### New options

This PR introduces a new option for servers: 

```json
{
	"auto_encrypt": {
		"allow_tls": true
	}
}
```

[...]

PR #12808, opened 2022-04-15
Update go version to 1.18.1

DanStough:
Updates to facilitate Go 1.18. Highlights on Changes:
* MOG updates which were released to deal with 1.18 changes
* Linter Update to Support 1.18 changes
* Misc Linter Fixes
* CI Runner updates
* Regenerated Test Certificates w/o using SHA1
* Deprecation Fixes

Deprecations:
* Verifying Certs w/ SHA1 signatures
* `x509.CertPool.Subjects()`
* `Strings.Title()`


connect/tls.go
```diff
@@ -111,6 +112,17 @@ func devTLSConfigFromFiles(caFile, certFile,
...
+func rawSubjectsFromCertPool(cp *x509.CertPool) []string {
+	subjectsIter := reflect.ValueOf(cp).Elem().FieldByName("byName").MapRange()
+	subjects := []string{}
+	for subjectsIter.Next() {
+		k := subjectsIter.Key()
+		subjects = append(subjects, k.String())
+	}
+	return subjects
+}
+
```
DanStough, line 125: Sadly, this is the first time I've used `reflect`. Something about this seems like it needs a safety check but I didn't dig too deep to figure out where.

agent/consul/leader_connect_ca.go
```diff
@@ -264,9 +266,10 @@ func newCARoot(pemValue, provider, clusterID string) (*structs.CARoot, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error extracting root key info: %v", err)
 	}
+	caser := cases.Title(language.English)
 	return &structs.CARoot{
 		ID:                  connect.CalculateCertFingerprint(primaryCert.Raw),
-		Name:                fmt.Sprintf("%s CA Primary Cert", strings.Title(provider)),
+		Name:                fmt.Sprintf("%s CA Primary Cert", caser.String(provider)),
```
rboyer, line 272: It's funny that the only place we use this is to pretty print these strings:
```
const (
	ConsulCAProvider = "consul"  // -> "Consul"
	VaultCAProvider  = "vault"   // -> "Vault"
	AWSCAProvider    = "aws-pca" // -> "Aws-Pca" :-(
)
```

api/api_test.go
```diff
@@ -1098,3 +1101,39 @@ func TestAPI_GenerateEnvHTTPS(t *testing.T) {
 
 	require.Equal(t, expected, c.GenerateEnv())
 }
+
+func getExpectedCaPoolByDir(t *testing.T) *x509.CertPool {
+	pool := x509.NewCertPool()
+	entries, err := os.ReadDir("../test/ca_path")
+	if err != nil {
+		t.Fatal("could not open test dir ../test/ca_path for reading")
+	}
```
rboyer, line 1110:
```suggestion
	require.NoError(t, err)
```

api/api_test.go
```diff
@@ -1098,3 +1101,39 @@ func TestAPI_GenerateEnvHTTPS(t *testing.T) {
...
+
+	for _, entry := range entries {
+		filename := path.Join("../test/ca_path", entry.Name())
+
+		data, err := ioutil.ReadFile(filename)
+		if err != nil {
+			t.Fatalf("could not open test file %s for reading", filename)
+		}
```
rboyer, line 1118:
```suggestion
		require.NoError(t, err)
```

agent/agent_test.go
```diff
@@ -5845,3 +5856,51 @@ func Test_coalesceTimerTwoPeriods(t *testing.T) {
 	})
 
 }
+
+func getExpectedCaPoolByFile(t *testing.T) *x509.CertPool {
+	pool := x509.NewCertPool()
+	data, err := ioutil.ReadFile("../test/ca/root.cer")
+	if err != nil {
+		t.Fatal("could not open test file ../test/ca/root.cer for reading")
+	}
```
rboyer, line 5865:
```suggestion
	require.NoError(t, err)
```

agent/agent_test.go
```diff
@@ -5845,3 +5856,51 @@ func Test_coalesceTimerTwoPeriods(t *testing.T) {
...
+	return pool
+}
+
+func getExpectedCaPoolByDir(t *testing.T) *x509.CertPool {
+	pool := x509.NewCertPool()
+	entries, err := os.ReadDir("../test/ca_path")
+	if err != nil {
+		t.Fatal("could not open test dir ../test/ca_path for reading")
+	}
```
rboyer, line 5877:
```suggestion
	require.NoError(t, err)
```

agent/agent_test.go
```diff
@@ -5845,3 +5856,51 @@ func Test_coalesceTimerTwoPeriods(t *testing.T) {
...
+
+	for _, entry := range entries {
+		filename := path.Join("../test/ca_path", entry.Name())
+
+		data, err := ioutil.ReadFile(filename)
+		if err != nil {
+			t.Fatalf("could not open test file %s for reading", filename)
+		}
```
rboyer, line 5885:
```suggestion
		require.NoError(t, err)
```

tlsutil/config.go
```diff
@@ -455,12 +456,23 @@ func newX509CertPool(groups ...[]string) (*x509.CertPool, error) {
...
 		}
 	}
-	if len(pool.Subjects()) == 0 {
+	if !haveSubjects(pool) {
 		return nil, nil
 	}
 	return pool, nil
 }
 
+// getSubjects uses the relect package to test for rawSubject fields.
```
rboyer, line 465:
```suggestion
// getSubjects uses the reflect package to test for rawSubject fields.
```

tlsutil/config.go
```diff
@@ -455,12 +456,23 @@ func newX509CertPool(groups ...[]string) (*x509.CertPool, error) {
 			}
 		}
 	}
-	if len(pool.Subjects()) == 0 {
+	if !haveSubjects(pool) {
```
rboyer, line 459: I think I would actually prefer if we drop this conditional entirely so we don't need this reflect trickery in non-test code.

rboyer: cc: @i0rek  who might have more context on what this was for originally in #5597

.changelog/12808.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:note
+chore: Upgrade to use Go 1.18.1
```
FFMMM, line 2:
```suggestion
dependency: Upgrade to use Go 1.18.1
```

tlsutil/config.go
DanStough: If we're only worried that a cert was added and not that it had a valid name, this might work:
```golang
func newX509CertPool(groups ...[]string) (*x509.CertPool, error) {
	var haveCerts bool
	pool := x509.NewCertPool()
	for _, group := range groups {
		for _, pem := range group {
			if !pool.AppendCertsFromPEM([]byte(pem)) {
				return nil, fmt.Errorf("failed to parse PEM %s", pem)
			}
			if len(pem) > 0 {
				haveCerts = true
			}
		}
	}
	if !haveCerts {
		return nil, nil
	}
	return pool, nil
}
```

agent/consul/leader_connect_ca.go
```diff
@@ -13,6 +13,8 @@ import (
 
 	"github.com/hashicorp/go-hclog"
 	uuid "github.com/hashicorp/go-uuid"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
```
FFMMM, line 17: I'm not really sure if it's worth pulling this in 🤔 

For what we use it later it seems that it's redundant. I can see the full value here if Consul could support internationalization of strings but it doesn't yet. Unless you or the team feels strongly for this, I'd suggest we punt on this change 🙈 

Hopefully I didn't miss a critical use case for this 🤦🏼 🤞🏼

api/api_test.go
```diff
@@ -1098,3 +1101,35 @@ func TestAPI_GenerateEnvHTTPS(t *testing.T) {
...
+		if !pool.AppendCertsFromPEM(data) {
+			t.Fatalf("could not add test ca %s to pool", filename)
+		}
+	}
+
+	return pool
+}
+
+// lazyCerts has a func field which can't be compared.
+var cmpCertPool = cmp.Options{
```
FFMMM, line 1125: this is 👨🏼‍🍳 👌🏼 , 💯 !

.github/workflows/build.yml
```diff
@@ -173,7 +173,7 @@ jobs:
       matrix:
         goos: [ darwin ]
         goarch: [ "amd64", "arm64" ]
-        go: [ "1.17.5" ]
+        go: [ "1.18.1" ]
```
FFMMM, line 176: nothing actionable now but pointing out that this may cause merge conflicts in ENT 📚

agent/connect_ca_endpoint_test.go
```diff
@@ -290,6 +290,4 @@ func TestConnectCARoots_PEMEncoding(t *testing.T) {
 	require.NoError(t, err)
 	pool := x509.NewCertPool()
 	require.True(t, pool.AppendCertsFromPEM(data))
-	// expecting the root cert from dc1 and an intermediate in dc2
-	require.Len(t, pool.Subjects(), 2)
```
FFMMM, line 294: do we not care/ want for this check anymore 🤔 ?

sdk/testutil/testlog.go
```diff
@@ -7,7 +7,7 @@ import (
 	"sync"
 	"testing"
 
-	"github.com/hashicorp/go-hclog"
+	hclog "github.com/hashicorp/go-hclog"
```
FFMMM, line 10: why ? 🙈

tlsutil/config.go
```diff
@@ -455,12 +456,23 @@ func newX509CertPool(groups ...[]string) (*x509.CertPool, error) {
...
 		return nil, nil
 	}
 	return pool, nil
 }
 
+// getSubjects uses the reflect package to test for rawSubject fields.
+func haveSubjects(cp *x509.CertPool) bool {
+	subjectsIter := reflect.ValueOf(cp).Elem().FieldByName("byName").MapRange()
```
FFMMM, line 467: this feels like it could panic if there is no field `byName` right?

FFMMM (commented):
This is awesome 💯 ! Thanks for working on this 🥇 ; and congrats on your first PR 🎉 !!

I left some questions and nits; It looks like you're working with RB so I'll defer to him for final approval.

Things LGTM overall!!

agent/consul/leader_connect_ca.go
DanStough: Yeah, I think I was just trying to leave things as-is as much as possible. Per the Slack thread with @rboyer, I think we can just remove the use case.

api/api_test.go
DanStough: Totally ~stolen~ borrowed from https://github.com/hashicorp/consul/commit/3e20bd25bdbec9b1b50fb5b1e0591314ec0c5fe2 😅

.github/workflows/build.yml
DanStough: Is it because ENT is on a different GO version? Is that intentional?

sdk/testutil/testlog.go
DanStough: I thought this was causing a linting error since the package is used as `hclog` but looks like I was wrong. Great catch!

tlsutil/config.go
DanStough: Yeah, that's a good call. I had a previous comment about how I needed to check something but wasn't familiar enough with `reflect`. I think I know how to handle it now.

agent/connect_ca_endpoint_test.go
DanStough: Originally this seemed like a pretty low value assertion like we were just testing the certPool add function, but I see now we're trying to validate the endpoint is returning the right number of cert. I found another way to do this.

tlsutil/config.go
DanStough: Actually this is the same code block RB pointed out above with my proposed soln. Will mark this resolved and we can move the discussion there.

connect/service_test.go
```diff
@@ -288,3 +290,15 @@ func TestService_HasDefaultHTTPResolverFromAddr(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, expected, got)
 }
+
+func getSubjects(cp *x509.CertPool) string {
```
rboyer, line 294: Maybe drop a TODO here for the future to reevaluate the kind of comparison we _should_ do here.

.github/workflows/build.yml
rboyer: `.circleci`, `.github`, `.release`, and `GNUMakefile` usually have collisions in general because of required divergences

rboyer (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/642045.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 95250e7915d07ba17217a7374cf96447a731142e onto `release/1.12.x` succeeded!

PR #12819, opened 2022-04-19
Implement the ServerDiscovery.WatchServers gRPC endpoint

mkeeler:
This PR adds a ServerDiscovery.WatchServers gRPC endpoint.

Additionally I did a bit of refactoring to of gRPC test functionality to make sharing more of it possible.


agent/consul/grpc_integration_test.go
```diff
@@ -73,12 +52,66 @@ func TestGRPCIntegration_ConnectCA_Sign(t *testing.T) {
 		Service:    "foo",
 	})
 
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	t.Cleanup(cancel)
+
+	ctx = public.ContextWithToken(ctx, TestDefaultInitialManagementToken)
```
mkeeler, line 58: Adding a token to the context because the new testGRPCIntegrationServer turns ACLs on.

agent/grpc/public/services/connectca/sign.go
```diff
@@ -48,6 +48,7 @@ func (s *Server) Sign(ctx context.Context, req *pbconnectca.SignRequest) (*pbcon
 	var rsp *pbconnectca.SignResponse
 	handled, err := s.ForwardRPC(&rpcInfo, func(conn *grpc.ClientConn) error {
 		logger.Trace("forwarding RPC")
+		ctx := public.ForwardMetadataContext(ctx)
```
mkeeler, line 51: This is critical to have the ACL token (and any other metadata from the incoming request) forwarded. This is due to the gRPC library using different keys to store the incoming and outgoing requests metadata.

agent/consul/server_test.go
```diff
@@ -257,6 +257,26 @@ func testACLServerWithConfig(t *testing.T, cb func(*Config), initReplicationToke
 	return dir, srv, codec
 }
 
+func testGRPCIntegrationServer(t *testing.T, cb func(*Config)) (*Server, *grpc.ClientConn) {
```
mkeeler, line 260: Kind of self explanatory but the purpose here is to create a test server with a listening gRPC server configured for it.

agent/grpc/public/services/connectca/server_test.go
```diff
@@ -2,86 +2,42 @@ package connectca
...
+	structs "github.com/hashicorp/consul/agent/structs"
 	"github.com/hashicorp/consul/proto-public/pbconnectca"
 )
 
 func noopForwardRPC(structs.RPCInfo, func(*grpc.ClientConn) error) (bool, error) {
 	return false, nil
 }
 
-func testStateStore(t *testing.T, publisher state.EventPublisher) *state.Store {
+func setupFSMAndPublisher(t *testing.T) (*testutils.FakeFSM, state.EventPublisher) {
```
mkeeler, line 21: The bulk of what used to be here was moved to a shared location to reduce some of the boilerplate.

agent/grpc/public/services/serverdiscovery/watch_servers.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/hashicorp/consul/proto-public/pbserverdiscovery"
+	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-uuid"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// We tag logs with a unique identifier to ease debugging. In the future this
+// should probably be an Open Telemetry trace ID.
+func streamID() string {
```
riddhi89, line 21: We can re-use this helper I added under utils: https://github.com/hashicorp/consul/blob/main/agent/grpc/public/utils.go#L7

agent/grpc/public/services/serverdiscovery/watch_servers.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	return id
+}
+
+// WatchServers provides a stream on which you can receive the list of servers
+// that are ready to receive incoming requests including stale queries. The
+// current set of ready servers are sent immediately at the start of the
+// stream and new updates will be sent whenver the set of ready servers changes.
+func (s *Server) WatchServers(req *pbserverdiscovery.WatchServersRequest, serverStream pbserverdiscovery.ServerDiscoveryService_WatchServersServer) error {
+	logger := s.Logger.Named("watch-servers").With("trace_id", streamID())
```
riddhi89, line 34: Looks like we use a different identifier for each the public grpc request (`trace_id`, `request_id`, `stream_id`). I think when we add the middleware/interceptor, we could make it consistent.

agent/consul/grpc_integration_test.go
```diff
@@ -73,12 +52,66 @@ func TestGRPCIntegration_ConnectCA_Sign(t *testing.T) {
...
+	}
+
+	// The gRPC endpoint itself well-tested with mocks. This test checks we're
+	// correctly wiring everything up in the server by:
+	//
+	//	* Starting a server
+	// * Initiating the gRPC stream
+	// * Validating the snapshot
+	// * Adding another server
+	// * Validating another message is sent.
```
boxofrad, line 82:
```suggestion
	//	* Starting a server
	// 	* Initiating the gRPC stream
	// 	* Validating the snapshot
	// 	* Adding another server
	// 	* Validating another message is sent.
```

agent/consul/server_test.go
boxofrad: Nice! Thanks for pulling this out 🙇🏻‍♂️

agent/grpc/public/services/connectca/sign.go
boxofrad: Wow, good catch! I wonder if there's a way to make this happen with an interceptor, that wouldn't risk leaking tokens to any other gRPC APIs we integrate with in the future 🤔

proto-public/pbserverdiscovery/serverdiscovery.proto
```diff
@@ -0,0 +1,33 @@
...
+
+syntax = "proto3";
+
+package serverdiscovery;
+
+option go_package = "github.com/hashicorp/consul/proto-public/pbserverdiscovery";
+
+
+service ServerDiscoveryService {
+   rpc WatchServers(WatchServersRequest) returns (stream WatchServersResponse) {};
```
boxofrad, line 12: As this is a public API, would it be possible to add an explanatory doc comment for this method? The one from the Go implementation is great!

agent/grpc/public/services/serverdiscovery/watch_servers.go
```diff
@@ -0,0 +1,157 @@
...
+	logger.Debug("starting stream")
+	defer logger.Trace("stream closed")
+
+	token := public.TokenFromContext(serverStream.Context())
+
+	// Serve the ready servers from an EventPublisher subscription. If the subscription is
+	// closed due to an ACL change, we'll attempt to re-authorize and resume it to
+	// prevent unnecessarily terminating the stream.
+	var idx uint64
+	for {
```
boxofrad, line 45: Not for now (as I think it'd be premature) but if we ended up writing this event-consumer-loop for a third time then it'd probably be worth extracting 😄

boxofrad (approved):
Looks great, nice job!

agent/grpc/public/services/serverdiscovery/watch_servers.go
mkeeler: Absolutely we should. For now though I can at least unify them to request_id.

mkeeler: All of the endpoints should now be using `request_id`

agent/grpc/public/services/serverdiscovery/watch_servers.go
mkeeler: 👍

proto-public/pbserverdiscovery/serverdiscovery.proto
mkeeler: Done

mkeeler:
Just waiting for green tests again (had to regen the protobuf files after adding comments to make CI happy)

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/647772.

PR #12822, opened 2022-04-19
Expose functions to set defaults and setting client transport with TLC config so you can easily pass in your http client to api.NewClient

jmurret:
### Background
Consul k8s has a kubernetes webhook controller called endpoints controller that listens for when new k8s services are created, deleted, and updated.  It will take the appropriate action to register, update, or deregister the service with Consul.  When it does this it will use the conul client agent that is available on the same host as the service.  We have had a couple of customers report that there are times when they are using cluster auto scale down or spot instances, where the node gets terminated and this controller process hangs for sometimes up to twenty minutes with the consul client agent not being available.  For the bug in consul-k8s where endpoints controller starts queueing requests in these scenarios, a simple timeout out would cause the current service record to fail and be reprocessed, at which point a new list of active consul agents would refresh and not include the offline consul agent...in effect, adding a timeout would cause a self-correcting requeue.

### Overview
We found that `api.NewClient()` uses the default go http client and does not enforce any client timeout.  `api.NewClient()` does allow passing in your own httpclient as part of the config.   So, it is set up to allow the endpoints controller to create its own httpClient with a timeout and pass that into the configuration.  We did find though that though that the default behavior of setting TLS properties on `client.Transport()` where private configured inside `api.NewHttpClient()` and there was not a way to access setting this default behavior.  

This PR does not change existing behavior within Consul and only exposes two functions that allow gaining access to setting default configuration and client.Transport settings that were only accessible if calling `api.NewClient()` without passing in an httpClient.

This [PR in consul-k8s](https://github.com/hashicorp/consul-k8s/pull/1178/files#diff-e26f3f069a83cd05579a03865c772ef61f179d3da5cf9636e6029c430d44f25bR15-R27) shows how we will use these functions to set up.

### Out of scope
- Setting any default timeouts within Consul itself (although we should consider researching and setting this in future work)
- Changing behavior in Consul

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -680,6 +680,11 @@ func SetClientConfig(config *Config) error {
 		if err != nil {
 			return err
 		}
+	} else {
+		err := setClientTransportWithTLSConfig(config.HttpClient, config.Transport, config.TLSConfig)
+		if err != nil {
+			return err
+		}
 	}
 
 	if config.Namespace == "" {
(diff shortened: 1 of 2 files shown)


dhiaayachi (commented):
Hey @jmurret, thank you for working on this. I left few comments about some of the behaviour. As a general comment about the API, the new methods (`SetClientConfig` and `SetClientTransportWithTLSConfig`) seems to need to be called before starting using the client otherwise it could lead to non expected behaviour (from user perspective). I'm wondering if we could, instead of mutating the transport and the config after the Client creation, introduce new ways of creating the client with the right transport/config? 

That way once the client is created it's fully functional.

jmurret:
Closing, did this in consul-k8s

PR #12831, opened 2022-04-20
[1.12.x] docs: Backport agent configuration hierarchy reorganization

blake:
Backport #12562 to 1.12.x branch.

PR #12848, opened 2022-04-23
Ma/stable website merge v2

markan:
Do not merge

This is just for review before pushing to stable-website.


trujillo-adam:
I reviewed the changes from a content perspective and they _seem_ OK. I had very little involvement this cycle, so I can't say for sure that everything is accounted for. For what it's worth, I did not see changes related to [#12831], which was the update I knew about and was looking for. But maybe it's not part of this PR because it was already merged into release/1.12.x? I can't really speak to anything else.

markan:
Ended up abandoning this approach, and pushed 1.12.x directly to stable-website with later fixups.

PR #12878, opened 2022-04-28
Support x-forwarded-client-cert 

### LINKED ISSUE #12852: Support envoy x-forwarded-client-cert header in consul
#### Feature Description

Envoy provides support forwarding and annotating the x-forwarded-client-cert header via the forward_client_cert_details set_current_client_cert_details filter fields. It would be helpful for consul to support this directly in its config. The escape hatches are a bit cumbersome for this purpose. 

#### Use Case(s)
This information is useful for various application side checks and validation. 

#### Implementation proposal

Extend the service registration proxy parameters [structure](https://www.consul.io/docs/connect/registration/service-registration#mesh-gateway-configuration-reference) to include a new structure
```
filter_options = {
     forward_client_cert_details = SANITIZE|FORWARD_ONLY|APPEND_FORWARD|ALWAYS_FORWARD_ONLY
     set_current_client_cert_details = {
         subject = <BOOL>
         cert = <BOOL>
         chain = <BOOL>
         dns = <BOOL>
         uri = <BOOL>
    }
}
```

This is a direct representation of the envoy config fields [extensions.filters.network.http_connection_manager.v3.HttpConnectionManager.ForwardClientCertDetails](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#enum-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-forwardclientcertdetails) and 
[extensions.filters.network.http_connection_manager.v3.HttpConnectionManager.SetCurrentClientCertDetails](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-setcurrentclientcertdetails)

##### Alternatives
We could flatten those fields out and not introduce the filter_options structure, but that seems unnecessarily verbose and clutters the config structure. We may also wish to add other fields at a later date. 

We could also use the ```config``` blob to insert something similar without extending the structure. This structure is described as 'opaque' in the main description but in practice it seems we use that for other proxy specific [overrides](https://www.consul.io/docs/connect/proxies/envoy#escape-hatch-overrides) and would avoid further expansion of the proxy structure. The primary reason to not follow this pattern is that it more complex to do early parameter validation.

markan:
### Description

Add x-fowarded-client-cert information where appropriate

Envoy provides support forwarding and annotating the x-forwarded-client-cert header via the forward_client_cert_details set_current_client_cert_details filter fields. It would be helpful for consul to support this directly in its config. The escape hatches are a bit cumbersome for this purpose.

### Links
Closes #12852

Asana [consul-x-forwarded-client-cert](https://go.hashi.co/consul-x-fowarded-client-cert)

### PR Checklist

* [x] Unit tests
* [x] End to end tests verifying headers are correct
* [ ] not a security concern

### DIFF
--- a/.changelog/12878.txt
+++ b/.changelog/12878.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+agent: Envoy now inserts x-forwarded-client-cert for incoming proxy connections
+```
--- a/.changelog/12878.txt
+++ b/.changelog/12878.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-agent: Envoy now inserts x-forwarded-client-cert for incoming proxy connections
+xds: Envoy now inserts x-forwarded-client-cert for incoming proxy connections
 ```
--- a/agent/xds/listeners.go
+++ b/agent/xds/listeners.go
@@ -859,6 +859,10 @@ func (s *ResourceGenerator) makeInboundListener(cfgSnap *proxycfg.ConfigSnapshot
 		if err != nil {
 			return nil, err
 		}
+		if meshConfig := cfgSnap.MeshConfig(); meshConfig == nil || meshConfig.HTTP == nil || !meshConfig.HTTP.SanitizeXForwardedClientCert {
+			filterOpts.forwardClientDetails = true
+			filterOpts.forwardClientPolicy = envoy_http_v3.HttpConnectionManager_APPEND_FORWARD
+		}
 	}
 	filter, err := makeListenerFilter(filterOpts)
 	if err != nil {
@@ -1146,6 +1150,13 @@ func (s *ResourceGenerator) makeFilterChainTerminatingGateway(
 
 		opts.cluster = ""
 		opts.useRDS = true
+
+		if meshConfig := cfgSnap.MeshConfig(); meshConfig == nil || meshConfig.HTTP == nil || !meshConfig.HTTP.SanitizeXForwardedClientCert {
+			opts.forwardClientDetails = true
+			// Note: filter Connection may not be mTLS, so then ALWAYS_FORWARD_ONLY. For mTLS connections we might want APPEND_FORWARD.
+			// Open question; how do I determine if this is mTLS or not?
+			opts.forwardClientPolicy = envoy_http_v3.HttpConnectionManager_ALWAYS_FORWARD_ONLY
+		}
 	}
 
 	filter, err := makeListenerFilter(opts)
@@ -1366,16 +1377,18 @@ func (s *ResourceGenerator) getAndModifyUpstreamConfigForListener(
 }
 
 type listenerFilterOpts struct {
-	useRDS           bool
-	protocol         string
-	filterName       string
-	routeName        string
-	cluster          string
-	statPrefix       string
-	routePath        string
-	requestTimeoutMs *int
-	ingressGateway   bool
-	httpAuthzFilter  *envoy_http_v3.HttpFilter
+	useRDS               bool
+	protocol             string
+	filterName           string
+	routeName            string
+	cluster              string
+	statPrefix           string
+	routePath            string
+	requestTimeoutMs     *int
+	ingressGateway       bool
+	httpAuthzFilter      *envoy_http_v3.HttpFilter
+	forwardClientDetails bool
+	forwardClientPolicy  envoy_http_v3.HttpConnectionManager_ForwardClientCertDetails
 }
 
 func makeListenerFilter(opts listenerFilterOpts) (*envoy_listener_v3.Filter, error) {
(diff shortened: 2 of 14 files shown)


.changelog/12878.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+agent: Envoy now inserts x-forwarded-client-cert for incoming proxy connections
```
blake, line 2:
```suggestion
xds: Envoy now inserts x-forwarded-client-cert for incoming proxy connections
```

website/content/docs/connect/config-entries/mesh.mdx
```diff
@@ -359,6 +359,20 @@ Note that the Kubernetes example does not include a `partition` field. Configura
...
+      name: 'HTTP',
+      type: 'HTTPConfig: <optional>',
+      description: 'HTTP configuration for the service mesh.',
+      children: [
+        {
+          name: 'SanitizeXForwardedClientCert',
+          yaml: false,
+          type: 'bool: <optional>',
+          description: `Set the envoy forwardClientCertDetails to SANITIZE everywhere. Ordinarily Consul will configure Envoy to
+                        insert x-forwarded-client-cert headers where appropriate. This returns Consul to the pre 1.12.1 behavior`,
```
blake, line 372:
```suggestion
          description: `Set the envoy \`forward_client_cert_details\` option to \`SANITIZE\` for all proxies. This configures Envoy to not send the \`x-forwarded-client-cert\` header to the next hop. If unspecified or \`false\`, the XFCC header is propagated to upstream applications.`,
```

agent/xds/listeners.go
```diff
@@ -1146,6 +1150,13 @@ func (s *ResourceGenerator) makeFilterChainTerminatingGateway(
 
 		opts.cluster = ""
 		opts.useRDS = true
+
+		if meshConfig := cfgSnap.MeshConfig(); meshConfig == nil || meshConfig.HTTP == nil || !meshConfig.HTTP.SanitizeXForwardedClientCert {
+			opts.forwardClientDetails = true
+			// Note: filter Connection may not be mTLS, so then ALWAYS_FORWARD_ONLY. For mTLS connections we might want APPEND_FORWARD.
+			// Open question; how do I determine if this is mTLS or not?
```
freddygv, line 1157: These filter chains use mTLS, the TLS context set above requires client certificates:
https://github.com/hashicorp/consul/blob/3b3f001580ea919057478657902cd5f4aae378aa/agent/xds/listeners.go#L1097

I don't think that `ALWAYS_FORWARD_ONLY` would work for terminating gateways because the header would never be set.  

`APPEND_FORWARD` is currently only set on the _inbound_ listener, and terminating gateways would only be dialed by a connect proxy's _upstream_ listener. 

This might need to be `APPEND_FORWARD`.

agent/xds/listeners.go
```diff
@@ -859,6 +859,10 @@ func (s *ResourceGenerator) makeInboundListener(cfgSnap *proxycfg.ConfigSnapshot
 		if err != nil {
 			return nil, err
 		}
+		if meshConfig := cfgSnap.MeshConfig(); meshConfig == nil || meshConfig.HTTP == nil || !meshConfig.HTTP.SanitizeXForwardedClientCert {
+			filterOpts.forwardClientDetails = true
+			filterOpts.forwardClientPolicy = envoy_http_v3.HttpConnectionManager_APPEND_FORWARD
```
freddygv, line 864: FYI @rboyer this is making it seem like with peering L7 applications could end up with two forwarded certificates. The one for the mesh gateway and the one for the source service. 

I don't think that's necessarily bad, but should probably be documented somewhere. Before peering there will only ever be one certificate present because gateways don't terminate mTLS.

freddygv (commented):
This PR probably has the right shape, but I don't fully understand the intended behavior.

Who is expected to set the `x-forwarded-client-cert`? The source application? or do we only want to forward the certificate from the source proxy (which was generated by Consul)?

From reading through the background I think we only want to forward the source mesh proxy certificate. But if that's not the case, then we also need to modify outbound listeners for connect proxies. (Upstream listeners for explicit upstreams, and outbound_listener for transparent proxies)

agent/xds/listeners.go
markan: I agree, changed.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/661256.

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit 2fcac5224eff32da7624d579601112562c37b666 onto `release/1.12.x` succeeded!

PR #12881, opened 2022-04-28
Validate port on mesh service registration

riddhi89:
### Description
Without a port in the registration, the mesh is not functional. For Agentless, there will not be any client agent for auto port assignment. Agentless phase 1 is for k8s workloads which are already able to specify a port on mesh service registration. 
This PR change is to require a port to be specified on registration of a mesh service (connect proxy, all gateway types (except ingress), sidecar definitions, connect native services) - this is mostly the case already except the validation was missing for connect native services. 
Centralized port allocation by servers would be done in the future to support other workloads (example: VM based) that depend on port assignment.

### DIFF
--- a/agent/consul/catalog_endpoint.go
+++ b/agent/consul/catalog_endpoint.go
@@ -204,6 +204,10 @@ func servicePreApply(service *structs.NodeService, authz ACLResolveResult, authz
 		}
 	}
 
+	if service.Connect.SidecarService.Port == 0 {
+		return fmt.Errorf("Side car service must have non-zero port")
+	}
+
 	return nil
 }
 
--- a/agent/consul/catalog_endpoint.go
+++ b/agent/consul/catalog_endpoint.go
@@ -204,7 +204,7 @@ func servicePreApply(service *structs.NodeService, authz ACLResolveResult, authz
 		}
 	}
 
-	if service.Connect.SidecarService.Port == 0 {
+	if service.Connect.SidecarService != nil && service.Connect.SidecarService.Port == 0 {
 		return fmt.Errorf("Side car service must have non-zero port")
 	}
 
--- a/agent/consul/catalog_endpoint.go
+++ b/agent/consul/catalog_endpoint.go
@@ -204,10 +204,6 @@ func servicePreApply(service *structs.NodeService, authz ACLResolveResult, authz
 		}
 	}
 
-	if service.Connect.SidecarService != nil && service.Connect.SidecarService.Port == 0 {
-		return fmt.Errorf("Side car service must have non-zero port")
-	}
-
 	return nil
 }
 
--- a/agent/structs/structs.go
+++ b/agent/structs/structs.go
@@ -1334,8 +1334,8 @@ func (s *NodeService) Validate() error {
 					"services"))
 		}
 
-		if s.Port == 0 {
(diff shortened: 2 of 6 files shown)


agent/structs/structs.go
```diff
@@ -1464,6 +1463,16 @@ func (s *NodeService) Validate() error {
 					"A SidecarService cannot have a nested SidecarService"))
 			}
 		}
+
+		if s.Connect.SidecarService.Port == 0 {
```
riddhi89, line 1467: We will need to update the docs for this, yes? https://www.consul.io/docs/connect/registration/sidecar-service#minimal-example

riddhi89: Actually, I moved the check for the side car service only for the catalog endpoint. The agent endpoint could still continue to use it. So might not need to update the docs?

agent/structs/structs.go
```diff
@@ -1466,6 +1465,11 @@ func (s *NodeService) Validate() error {
 		}
 	}
 
+	if s.Connect.Native && s.Port == 0 {
+		result = multierror.Append(result, fmt.Errorf(
+			"Connect native service must have non-zero port"))
+	}
+
 	return result
 }
```
riddhi89, line 1474: To note is this Validate() helper is used by both the [catalog](https://github.com/hashicorp/consul/blob/main/agent/consul/catalog_endpoint.go#L166) and [agent endpoints](https://github.com/hashicorp/consul/blob/main/agent/agent_endpoint.go#L1156).

agent/structs/structs.go
```diff
@@ -1333,9 +1333,8 @@ func (s *NodeService) Validate() error {
 					"services"))
 		}
 
-		if s.Port == 0 && s.SocketPath == "" {
```
mkeeler, line 1336: This makes me think that we should require one of port or socket path everywhere.

agent/consul/catalog_endpoint.go
```diff
@@ -204,6 +204,10 @@ func servicePreApply(service *structs.NodeService, authz ACLResolveResult, authz
 		}
 	}
 
+	if service.Connect.SidecarService != nil && service.Connect.SidecarService.Port == 0 {
```
mkeeler, line 207: This will generally never be hit because SidecarService will always be nil.

https://github.com/hashicorp/consul/blob/a668c369306a1441377bbdaa1970d61fd5121b6f/agent/agent_endpoint_test.go#L3649-L3652

So I think the better validation would be to check the top level Port (or I guess SocketPath) to ensure one of those is filled in for services with kind connect-proxy.

mkeeler (changes requested):

agent/structs/structs.go
riddhi89: Added. Let me know if it is not the case for gateway types - in which case I will remove the sockpath validation below.

agent/consul/catalog_endpoint.go
riddhi89: Removed this check.

mkeeler (approved):
LGTM, validating that a port is specified for connect native services seems like the right thing to do here.

agent/structs/structs.go
riddhi89: Removed for gateway types per our discussion.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/662637.

PR #12885, opened 2022-04-28
Store and return RPC error in ACL cache entries

kyhavlov:
This PR adds an `Error` field to the `IdentityCacheEntry` to preserve the error returned when attempting to resolve a token. Currently we return `acl.ErrNotFound` in this case which bubbles up to the client agent that made the request and the down policy ends up not being respected if it's set to `allow`.

### DIFF
--- a/agent/consul/acl.go
+++ b/agent/consul/acl.go
@@ -360,20 +360,20 @@ func (r *ACLResolver) fetchAndCacheIdentityFromToken(token string, cached *struc
 	err := r.backend.RPC("ACL.TokenRead", &req, &resp)
 	if err == nil {
 		if resp.Token == nil {
-			r.cache.PutIdentity(cacheID, nil)
+			r.cache.PutIdentity(cacheID, nil, nil)
 			return nil, acl.ErrNotFound
 		} else if resp.Token.Local && r.config.Datacenter != resp.SourceDatacenter {
-			r.cache.PutIdentity(cacheID, nil)
+			r.cache.PutIdentity(cacheID, nil, nil)
 			return nil, acl.PermissionDeniedError{Cause: fmt.Sprintf("This is a local token in datacenter %q", resp.SourceDatacenter)}
 		} else {
-			r.cache.PutIdentity(cacheID, resp.Token)
+			r.cache.PutIdentity(cacheID, resp.Token, nil)
 			return resp.Token, nil
 		}
 	}
 
 	if acl.IsErrNotFound(err) {
 		// Make sure to remove from the cache if it was deleted
-		r.cache.PutIdentity(cacheID, nil)
+		r.cache.PutIdentity(cacheID, nil, err)
 		return nil, acl.ErrNotFound
 
 	}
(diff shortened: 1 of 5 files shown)


agent/consul/acl.go
```diff
@@ -416,7 +416,7 @@ func (r *ACLResolver) resolveIdentityFromToken(token string) (structs.ACLIdentit
 	waitForResult := cacheEntry == nil || r.config.ACLDownPolicy != "async-cache"
 	if !waitForResult {
 		// waitForResult being false requires the cacheEntry to not be nil
-		return cacheEntry.Identity, nil
+		return cacheEntry.Identity, cacheEntry.Error
```
kisunji, line 419: Since the caller has special handling for `ACLRemoteError`, we need to make sure we wrap the error like we do in L430, otherwise we'll miss the `ACLRemoteError` codepath

```suggestion
		if cacheEntry.Error != nil && !acl.IsErrNotFound(cacheEntry.Error) {
			return cacheEntry.Identity, ACLRemoteError{Err: cacheEntry.Error}
		}
		return cacheEntry.Identity, cacheEntry.Error
```

agent/consul/acl_test.go
```diff
@@ -1264,6 +1264,35 @@ func TestACLResolver_DownPolicy(t *testing.T) {
...
+		})
+
+		// Attempt to resolve the token - this should set up the cache entry with a nil
+		// identity and permission denied error.
+		authz, err := r.ResolveToken(secretID)
+		require.NoError(t, err)
+		require.NotNil(t, authz)
+		entry := r.cache.GetIdentity(tokenSecretCacheID(secretID))
+		require.Nil(t, entry.Identity)
+		require.Equal(t, acl.ErrPermissionDenied, entry.Error)
```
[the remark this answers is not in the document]
kisunji: After applying the proposed change in `acl.go` the second `ResolveToken` call passes for me

agent/consul/acl_test.go
```diff
@@ -1264,6 +1264,35 @@ func TestACLResolver_DownPolicy(t *testing.T) {
...
+		secretID := foundToken.SecretID
+
+		delegate := &ACLResolverTestDelegate{
+			enabled:       true,
+			datacenter:    "dc1",
+			legacy:        false,
+			localTokens:   false,
+			localPolicies: false,
+			tokenReadFn: func(_ *structs.ACLTokenGetRequest, reply *structs.ACLTokenResponse) error {
+				return acl.ErrPermissionDenied
```
kisunji, line 1279: slight nit but can we return some `testErr := fmt.Errorf("network error")`?

The type of error returned here isn't too important but I think using an error from the `acl` package (and the fact that we have special handling for a bunch of ACL error types) makes it a bit confusing.

agent/consul/acl_test.go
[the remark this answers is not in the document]
kisunji: If I wasn't clear:
```go
		authz, err := r.ResolveToken(secretID)
		require.NoError(t, err)
		require.NotNil(t, authz)
		entry := r.cache.GetIdentity(tokenSecretCacheID(secretID))
		require.Nil(t, entry.Identity)
		require.Equal(t, acl.ErrPermissionDenied, entry.Error)

		authz, err = r.ResolveToken(secretID)
		require.NoError(t, err) <- your implementation fails here but it shouldn't
		require.NotNil(t, authz)
		entry = r.cache.GetIdentity(tokenSecretCacheID(secretID))
		require.Nil(t, entry.Identity)
		require.Equal(t, acl.ErrPermissionDenied, entry.Error)
```

kisunji (changes requested):
Need to add special handling to the newly returned errors and should update the test a bit to replicate the bug this PR aims to address

kisunji (approved):
LGTM! Thanks for working on this. Hopefully this cuts down on all the flakes that the bug subtly caused

agent/consul/acl.go
kyhavlov: Thanks, I totally missed the part about needing the return the err wrapped in an `ACLRemoteError` - I thought the bug was only that we weren't caching the error at all and returning an ErrNotFound when we shouldn't be

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/659162.

PR #12887, opened 2022-04-28
chore(ci): add initial support for backport assistant

DanStough:
### Description
Teams at Hashicorp are adopting [backport-assistant](https://github.com/hashicorp/backport-assistant) for managing commit backports to release branches. 

Right now, this work is performed by bespoke shell scripts in Consul. Furthermore, there have been issues with backporting changes that tests would have been caught by testing if commits hadn't been merged directly.

The purpose of the PR is two-fold, to pay down tech debt with CircleCI and address deficiencies in the current process of testing backport changes.

### Execution Plan

I'm proposing activating this feature in parallel with the current cherry-pick.sh action. As advertised, this action will open PRs for backport changes. Even if the cherry-pick script merges directly to the target release branches, my assumption is that it will run tests against the non-existent changes with the base branch. 

This will create the added burden of closing empty PRs in the near term, but after a couple of days, we should be able to have the confidence to turn off the cherry-pick script and use this method. In addition, we can add another job to detect backport changes to the website.

### Security Implications

IAW the `backport-assistant` guide, I created a GH PAT for this action. Since GHA mask tokens and this action only runs against protected branches, I think there is minimal risk in leaking the token through an action or running the action against a fork.

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/.github/workflows/backport-assistant.yml
+++ b/.github/workflows/backport-assistant.yml
@@ -1,0 +1,33 @@
+# This action creates downstream PRs for PRs with backport labels defined.
+# See docs here: https://github.com/hashicorp/backport-assistant
+
+name: Backport Assistant Runner
+
+on:
+  pull_request_target:
+    types:
+      - closed
+      - labeled
+
+jobs:
+  backport:
+    if: github.event.pull_request.merged
+    runs-on: ubuntu-latest
+    container: hashicorpdev/backport-assistant:0.2.3
+    steps:
+      - name: Run Backport Assistant for stable-website
+        # Update this to auto-merge when we have confidence in the process working and kill Circle
+        run: |
+          backport-assistant backport -merge-method=rebase
+        env:
+          BACKPORT_LABEL_REGEXP: "type/docs-cherrypick"
+          BACKPORT_TARGET_TEMPLATE: "stable-website"
+          GITHUB_TOKEN: ${{ secrets.ELEVATED_GITHUB_TOKEN }}
+      - name: Run Backport Assistant for release branches
+        # Update this to auto-merge when we have confidence in the process working and kill Circle
+        run: |
+          backport-assistant backport -merge-method=rebase
+        env:
+          BACKPORT_LABEL_REGEXP: "backport\/(?P<target>\\d+\\.\\d+)"
+          BACKPORT_TARGET_TEMPLATE: "release/{{.target}}.x"
+          GITHUB_TOKEN: ${{ secrets.ELEVATED_GITHUB_TOKEN }}


kyhavlov (approved):
This looks great - I think having the PRs opened alongside the existing cherry pick automation will be a good way to test it out.

.github/workflows/backport-assistant.yml
```diff
@@ -0,0 +1,28 @@
+# This action creates downstream PRs for PRs with backport labels defined.
+# See docs here: https://github.com/hashicorp/backport-assistant
+
+name: Backport Assistant Runner
+
+on:
+  # TODO: Figure out if this should be `pull_request_target` if we have forks.
+  pull_request:
```
eculver, line 8: I think we do want `pull_request_target` here. From the docs for `pull_request_target`:
> This event runs in the context of the base of the pull request, rather than in the context of the merge commit. This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow.

Also, looking at the other users of `backport-assistant` ([query](https://cs.github.com/?scopeName=All+repos&scope=&q=+hashicorpdev%2Fbackport-assistant+path%3A.github)), the majority use `pull_request_target`.

.github/workflows/backport-assistant.yml
```diff
@@ -0,0 +1,28 @@
...
+    types:
+      - closed
+      - labeled
```
eculver, line 11: Am I correct in assuming that the `labeled` event would trigger this whenever someone adds a label to the PR? If that is true, it seems that it would run prior to being merged. Is this what we want? If that's true, it seems to be at odds with the `closed` event above it. Can you confirm the intended behavior here? 

I did find this in [the docs for the `label` event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#label):
>  If you want to run your workflow when a label is added to or removed from an issue, pull request, or discussion, use the labeled or unlabeled activity types

This seems to indicate that this workflow would run when a label is added which could be pre-merge.

.github/workflows/backport-assistant.yml
DanStough: Oh, that's a much clearer description that the backport-assistant docs. I don't think it's too much of an issue since this only runs on merge but `pull_request_target` is the better choice.

.github/workflows/backport-assistant.yml
DanStough: Good question. My understanding is that this is intended, with the usage being that you could add the backport labels post-merge and they would still get applied. Line 15 below (`if: github.event.pull_request.merged`) would guard against the workflow running pre-merge.

DanStough:
I added the stable-website stuff to this PR too if anyone has any objection.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/657310.

Issue #10274, opened 2021-05-21
Retry policy in service-router is not working as expected


arushi315:
Hi Team, 
  
I am using ingress gateway for my service in consul service mesh with ring_hash load balancing policy and I am facing issue with the retry policy. Here is how my configuration for service-resolver and service-router looks like: 

``` json
{
            "Kind": "service-resolver",
            "Name": "myservice",
            "LoadBalancer": {
                "Policy": "ring_hash",
                "HashPolicies": [
                    {
                        "Field": "query_parameter",
                        "FieldValue": "myserviceid",
                        "Terminal": true
                    },
                    {
                        "Field": "header",
                        "FieldValue": "myserviceid", 
                        "Terminal": true
                    }
                ]
            }
        } 
```

``` json
{
    "Kind": "service-router",
    "Name": "ingress-http-backend",
    "Routes": [
        {
            "Match": {
                "HTTP": {
                    "PathPrefix": "/myservice"
                }
            },
            "Destination": {
                "RequestTimeout": "30m0s",
                "Service": "myservice",
                "NumRetries": 3,
                "RetryOnConnectFailure": true,
                "RetryOnStatusCodes": [
                    503,
                    504
                ]
            }
        }
    ],
    "CreateIndex": 160,
    "ModifyIndex": 251
} 
```

snippet of envoy config: 
```
"dynamic_route_configs": [
    {
     "version_info": "00000009",
     "route_config": {
      "@type": "type.googleapis.com/envoy.config.route.v3.RouteConfiguration",
      "name": "8080",
      "virtual_hosts": [
       {
        "name": "ingress-http-backend",
        "domains": [
         "*",
         "*:8080"
        ],
        "routes": [
         {
          "match": {
           "prefix": "/myservice"
          },
          "route": {
           "cluster": "myservice.default.dev-setup.internal.1c0cc154-47ba-2984-0f53-4dc905625cff.consul",
           "timeout": "1800s",
           "retry_policy": {
            "retry_on": "connect-failure,retriable-status-codes",
            "num_retries": 3,
            "retriable_status_codes": [
             503,
             504
            ]
           },
           "hash_policy": [
            {
             "terminal": true,
             "query_parameter": {
              "name": "myserviceid"
             }
            },
            {
             "header": {
              "header_name": "myserviceid"
             },
             "terminal": true
            }
           ]
          }
         },
         {
          "match": {
           "prefix": "/"
          },
          "route": {
[...]

PR #12890, opened 2022-04-29
Add the ability to retry on reset connection to service-routers

aoskotsky-amplify:
### Description
This adds a `RetryOnReset` option to `service-router` configs. 

My apps are getting an error intermittently when using consul connect: `upstream connect error or disconnect/reset before headers. reset reason: connection termination`. Envoy has an option to automatically retry in case of this kind of error. Consul does support does not support this option at the moment so I'm implementing it here.

Create a `service-router` with the `RetryOnReset` option like below:

```json
{
    "Kind": "service-router",
    "Name": "test",
    "Routes": [
        {
            "Match": {
                "HTTP": {
                    "PathPrefix": "/"
                }
            },
            "Destination": {
                "NumRetries": 3,
                "RetryOnReset": true,
                "Service": "test"
            }
        }
    ]
}

```

### Links
Issue #10274 

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/structs/config_entry_discoverychain.go
+++ b/agent/structs/config_entry_discoverychain.go
@@ -409,6 +409,11 @@ type ServiceRouteDestination struct {
 	// 4 failure bubbling up to layer 7.
 	RetryOnConnectFailure bool `json:",omitempty" alias:"retry_on_connect_failure"`
 
+	// RetryOnReset allows for disconnects and connection resets to trigger a
+	// retry. This should be expressible in other proxies as it's just a layer
+	// 4 failure bubbling up to layer 7.
+	RetryOnReset bool `json:",omitempty" alias:"retry_on_reset"`
+
 	// RetryOnStatusCodes is a flat list of http response status codes that are
 	// eligible for retry. This again should be feasible in any reasonable proxy.
 	RetryOnStatusCodes []uint32 `json:",omitempty" alias:"retry_on_status_codes"`
@@ -455,7 +460,7 @@ func (e *ServiceRouteDestination) UnmarshalJSON(data []byte) error {
 }
 
 func (d *ServiceRouteDestination) HasRetryFeatures() bool {
-	return d.NumRetries > 0 || d.RetryOnConnectFailure || len(d.RetryOnStatusCodes) > 0
+	return d.NumRetries > 0 || d.RetryOnConnectFailure || d.RetryOnReset || len(d.RetryOnStatusCodes) > 0
 }
 
 // ServiceSplitterConfigEntry defines how incoming requests are split across
--- a/agent/proxycfg/testing_upstreams.go
+++ b/agent/proxycfg/testing_upstreams.go
@@ -632,6 +632,16 @@ func setupTestVariationDiscoveryChain(
 					},
 					{
 						Match: httpMatch(&structs.ServiceRouteHTTPMatch{
+							PathPrefix: "/retry-reset",
+						}),
+						Destination: &structs.ServiceRouteDestination{
+							Service:      "retry-reset",
+							NumRetries:   15,
+							RetryOnReset: true,
+						},
+					},
+					{
+						Match: httpMatch(&structs.ServiceRouteHTTPMatch{
 							PathPrefix: "/retry-codes",
 						}),
 						Destination: &structs.ServiceRouteDestination{
@@ -642,11 +652,12 @@ func setupTestVariationDiscoveryChain(
 					},
 					{
 						Match: httpMatch(&structs.ServiceRouteHTTPMatch{
-							PathPrefix: "/retry-both",
+							PathPrefix: "/retry-all",
 						}),
 						Destination: &structs.ServiceRouteDestination{
-							Service:               "retry-both",
+							Service:               "retry-all",
 							RetryOnConnectFailure: true,
+							RetryOnReset:          true,
 							RetryOnStatusCodes:    []uint32{401, 409, 451},
 						},
 					},
--- a/agent/xds/routes.go
+++ b/agent/xds/routes.go
@@ -397,17 +397,21 @@ func makeUpstreamRouteForDiscoveryChain(
 					}
 
 					// The RetryOn magic values come from: https://www.envoyproxy.io/docs/envoy/v1.10.0/configuration/http_filters/router_filter#config-http-filters-router-x-envoy-retry-on
+					var retryStrings []string
 					if destination.RetryOnConnectFailure {
-						retryPolicy.RetryOn = "connect-failure"
+						retryStrings = append(retryStrings, "connect-failure")
 					}
+
+					if destination.RetryOnReset {
+						retryStrings = append(retryStrings, "reset")
+					}
+
 					if len(destination.RetryOnStatusCodes) > 0 {
-						if retryPolicy.RetryOn != "" {
-							retryPolicy.RetryOn = retryPolicy.RetryOn + ",retriable-status-codes"
-						} else {
-							retryPolicy.RetryOn = "retriable-status-codes"
-						}
+						retryStrings = append(retryStrings, "retriable-status-codes")
 						retryPolicy.RetriableStatusCodes = destination.RetryOnStatusCodes
 					}
+
+					retryPolicy.RetryOn = strings.Join(retryStrings, ",")
 
 					routeAction.Route.RetryPolicy = retryPolicy
 				}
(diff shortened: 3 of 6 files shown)


Amier3:
Hey @aoskotsky-amplify 

Thanks for the first-time contribution ( hope to see many more 👀  )! Generally we expose envoy-config on a per-case basis, but on first glance this use-case seems pretty solid. Since this wasn't made off an associated github issue, i'll talk this over with the team and let you know tommorow  if we'd like to expose this config to everyone.  

Then hopefully we'll get this reviewed and merged shortly after  👍

Amier3:
@aoskotsky-amplify 

Blake just let me know that this change came from a conversation on gitter, so you're all good to go 👍 . We'll get this reviewed shortly and thank you again for the contribution!

aoskotsky-amplify:
> @aoskotsky-amplify
> 
> Blake just let me know that this change came from a conversation on gitter, so you're all good to go 👍 . We'll get this reviewed shortly and thank you again for the contribution!

Awesome. Thanks!

blake:
Hi @aoskotsky-amplify,

Thank you for this contribution. I had an opportunity to discuss this PR with the team. Here's the feedback from our review.

Envoy supports a number of retry conditions. At the time of writing, there are ten (10) [`x-envoy-retry-on`](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#config-http-filters-router-x-envoy-retry-on) conditions and five (5) [`x-envoy-retry-grpc-on`](https://www.envoyproxy.io/docs/envoy/v1.22.0/configuration/http/http_filters/router_filter#x-envoy-retry-grpc-on) conditions. Instead of exposing condition through their own variables–as is the current approach used in Consul and the one followed by this PR–we think a better solution is to expose `retry_on` as a top-level array in the service router configuration where users can specify a list of retry conditions.

Would you be willing to modify your PR to accommodate this alternative UX? If so, we can share a few UX options we are considering and get your feedback on them prior to you making any changes to the PR.

aoskotsky-amplify:
Sure, I can modify the PR. What UX options are you considering?

blake:
Hi @aoskotsky-amplify, here's the UX we were considering.

It would probably be beneficial to have input validation on the array to limit the retry conditions that can be specified. After looking at Envoy's docs, I believe we'd only want to allow the following conditions to be set:

* `5xx`
* `gateway-error`
* `reset`
* `connect-failure`
* `envoy-ratelimited`
* `retriable-4xx`
* `refused-stream`

**gRPC retry conditions**
* `cancelled`
* `deadline-exceeded`
* `internal`
* `resource-exhausted`
* `unavailable`

## Special handling for certain conditions

### `connect-failure`

The `connect-failure` condition would need some special handling since there is currently a dedicated variable for enabling that retry condition, [RetryOnConnectFailure](https://www.consul.io/docs/connect/config-entries/service-router#retryonconnectfailure).

If `RetryOnConnectFailure` is set to true, and `connect-failure` is not already present in `retryPolicy.RetryOn`, the `connect-failure` value should be appended to the list of retry conditions.

### `retriable-status-codes`

I don't believe it makes sense to allow `retriable-status-codes` to be set directly in the the array of retry conditions because it is dependent on `retryPolicy.RetriableStatusCodes` also being configured. Since Consul does not configure Envoy to allow the [`x-envoy-retriable-status-codes`](https://www.envoyproxy.io/docs/envoy/v1.22.0/configuration/http/http_filters/router_filter#x-envoy-retriable-status-codes) header to be used to provide this list of codes (this is an alternative method for specifying the retry codes), the only option users have is to specify the list with [RetryOnStatusCodes](https://www.consul.io/docs/connect/config-entries/service-router#retryonstatuscodes) in the service router configuration.

In this case, I think it makes sense for that condition to only be enabled if one or more status codes are specified in `RetryOnStatusCodes`.

### `retriable-headers`

The `retriable-headers` is dependent on a list of retriable headers also being configured. Similar to the retriable status codes, the list of headers cannot be specified using the [x-envoy-retriable-header-names](https://www.envoyproxy.io/docs/envoy/v1.22.0/configuration/http/http_filters/router_filter#x-envoy-retriable-header-names) HTTP header.

Being that Consul does not have a configuration option that allows configuring [`retryPolicy.RetriableHeaders`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-retrypolicy-retriable-headers), I do not think it makes sense to allow this retry condition to be specified in the list.

What are your thoughts on this proposed UX?

aoskotsky-amplify:
Hi @blake, that UX makes sense to me. I'll start working on it.

I have a couple of questions though: 

1. Does Consul support service-router configs with gRPC services? I see the [service-router docs](https://www.consul.io/docs/connect/config-entries/service-router#requirements) say services need to be HTTP. If if it does, do you think the retry conditions should be a separate config option? Would all the options be available under a `RetryOn` option or would there separately be a `GrpcRetryOn`?

2. Do you think in the future there should be a `RetryOnHeaders` option in a separate PR?

blake:
Hi @aoskotsky-amplify, 

Service routers do support routing to gRPC services. The beginning of the doc implies under requirements that service routers can only be used with services using the `http` protocol, but this is not correct. Service routers can also route based on gRPC methods. See <https://www.consul.io/docs/connect/config-entries/service-router#grpc-routing> for an example.

> Do you think in the future there should be a RetryOnHeaders option in a separate PR?

Yes, we would like to support this feature as well. I think it makes sense to add it in a separate PR.

aoskotsky-amplify:
Oh I see. Do you think there should be separate options like `RetryOn` and `GrpcRetryOn`?

aoskotsky-amplify:
@blake I updated this PR with your suggestions. Can you give this another review? Thanks

aoskotsky-amplify:
@blake @Amier3 Can you please give this PR another review?

agent/structs/config_entry_discoverychain.go
```diff
@@ -251,6 +257,26 @@ func isValidHTTPMethod(method string) bool {
 	}
 }
 
+func isValidRetryCondition(retryOn string) bool {
```
mkeeler, line 260: It would be good to add a test for this validation function in agent/structs/config_entry_discoverychain_test.go to enumerate all the valid retry conditions. It is obvious that it works today but if we refactor things in the future it would be great to know if we break something.

Additionally it would be great if you could add two more cases to `TestServiceRouterConfigEntry` in that same file. One to ensure that a router with this field configured with correct entries is accepted and one to test the validation error.

mkeeler (changes requested):
The code looks good to me. In addition to the one request I added a comment for I would also ask that the `ServiceRouteDestination` type in api/config_entry_discoverychain.go also be updated to reflect the `RetryOn` field (and a test case added to `TestAPI_ConfigEntry_DiscoveryChain` in the corresponding test file).

As for your question about whether we should have a `RetryOnGRPC` separate from the main `RetryOn` field. My opinion is that since Envoy makes no distinction in the routing policy and can do the right thing, then its best to keep it as 1 field in Consul.

website/content/docs/connect/config-entries/service-router.mdx
```diff
@@ -703,6 +708,13 @@ spec:
       type: 'bool: false',
       description: 'Allows for connection failure errors to trigger a retry.',
     },
+    {
+      name: 'RetryOn',
+      type: 'array<string>',
+      description: `Allows retrying requests for a list of conditions. One of:
```
trujillo-adam, line 714:
```suggestion
      description: `Allows Consul to retry requests when the requests meet a specific set of conditions. You can specify one of the following lists:
```
Is that what this means?

trujillo-adam (commented):
Just a clarification to the docs

website/content/docs/connect/config-entries/service-router.mdx
aoskotsky-amplify: Yes. How about this? `Allows Consul to retry requests when they meet one of a set of conditions. The available conditions are:`

trujillo-adam: That's pretty good. We can probably even trim it down a bit more and improve consistency with the style guide:

`Allows Consul to retry requests when they meet one of the following sets of conditions:`

aoskotsky-amplify: Updated

agent/structs/config_entry_discoverychain.go
aoskotsky-amplify: Updated. Let me know if the tests look fine now

PR #12804, opened 2022-04-15
ui: Revert to Services route as default

evrowe:
Reverting the default route for Consul UI to be the Services page for the initial 1.12 release

PR #12894, opened 2022-04-29
ui: [Backport/1.12.x] Revert to Services route as default (#12804)

evrowe:
Backport of #12804

Reverts the default route back to Services list rather than Overview

no changelog due to base not being `main`


evrowe:
Ignore this PR, accidentally set the wrong base branch 😬

PR #12758, opened 2022-04-12
ui: Show read-replica health status

johncowen:
Previously we omitted the health status for read-replicas. We've decided that these should be shown just like all other servers.

PR #12896, opened 2022-04-29
ui: [Backport/1.12.x] Show read-replica health status (#12758)

evrowe:
Backport of #12804

Previously we omitted the health status for read-replicas. We've decided that these should be shown just like all other servers.

no changelog due to base not being `main`


johncowen (approved):
🚢

PR #12903, opened 2022-04-29
Docs: Show CLI command-specific options before general options

jkirschner-hashicorp:
## Context

A number of our CLI docs pages show the command-specific options below the list of general options that apply across many commands. (Example screenshot below from [keyring command](https://www.consul.io/commands/keyring#usage))

I see several advantages to moving command-specific options above the general ones:
- The unique behavior and use of the command is easier to understand. The explanation of what the command does is adjacent to its flags, rather than being separated by general stuff.
- In many cases, the general flags will by familiar (from using other commands) are already handled (such as by setting environment variables for the Consul address and token).

image

## Status

[Preview](https://consul-6qvyke6un-hashicorp.vercel.app/commands/acl/auth-method/create)

This PR currently shows the proposed change applied to a single file to get alignment on approach. If we agree that this is the right approach, I can work on extending it to the many other command pages. Aligning on a single command is important to avoid rework in many places later.

image

### DIFF
--- a/website/content/commands/acl/auth-method/create.mdx
+++ b/website/content/commands/acl/auth-method/create.mdx
@@ -23,12 +23,6 @@ are not supported from commands, but may be from the corresponding HTTP endpoint
 
 Usage: `consul acl auth-method create [options] [args]`
 
-#### API Options
-
-@include 'http_api_options_client.mdx'
-
-@include 'http_api_options_server.mdx'
-
 #### Command Options
 
 - `-description=<string>` - A description of the auth method.
@@ -71,6 +65,8 @@ Usage: `consul acl auth-method create [options] [args]`
 
 #### Enterprise Options
 
+@include 'http_api_partition_options.mdx'
+
 @include 'http_api_namespace_options.mdx'
 
 - `-namespace-rule-bind-namespace=<value>` - Namespace to bind on match. Can
@@ -80,7 +76,11 @@ Usage: `consul acl auth-method create [options] [args]`
   verified identity attributes returned from the auth method during login to
   determine if the namespace rule applies. Added in Consul 1.8.0.
 
-@include 'http_api_partition_options.mdx'
+#### API Options
+
+@include 'http_api_options_client.mdx'
+
+@include 'http_api_options_server.mdx'
 
 ## Examples
 


website/content/commands/acl/auth-method/create.mdx
```diff
@@ -71,6 +65,8 @@ Usage: `consul acl auth-method create [options] [args]`
 
 #### Enterprise Options
 
+@include 'http_api_partition_options.mdx'
```
jkirschner-hashicorp, line 68: Since partitions sit above namespaces in the Consul organizational hierarchy, I think it makes sense to show partition above namespace in the list. Thoughts?

website/content/commands/acl/auth-method/create.mdx
```diff
@@ -80,7 +76,11 @@ Usage: `consul acl auth-method create [options] [args]`
   verified identity attributes returned from the auth method during login to
   determine if the namespace rule applies. Added in Consul 1.8.0.
 
-@include 'http_api_partition_options.mdx'
+#### API Options
```
jkirschner-hashicorp, line 79: Because not all commands have enterprise options, I decided on the following sequencing:
- Command Options (which are specific to this command)
- Enterprise Options (which apply to a number of commands, but not all ... and have some variation)
- API Options (which apply to every command)

There are other ways we could break this up (e.g., include merge enterprise options into the command options section and use the inline enterprise alert), but I wasn't convinced that would be better.

trujillo-adam (approved):
I think this is fine. It would be nice to put the options into tables, though.

blake:
> I think this is fine. It would be nice to put the options into tables, though.

IMO the table format is problematic on pages with a large number of options because it does not provide a way to link to a specific option. Take for example, <https://www.consul.io/docs/discovery/services#service>. This format does not allow linking directly to content lower in the page, like the `connect` field/row. This requires the user to manually scroll and visually scan a significant amount of content to find the desired option.

Compare this to a link such as <https://www.consul.io/commands/acl/token/update#upgrade-legacy>. This element is lower on the page. The anchor link brings the reader directly to the desired option, eliminating the need to scan the table content to find the desired docs for the config option.

If we want to transition to a table format for options/flags across the site, I strongly recommend we retain the anchor links. Otherwise, it feels like a UX regression.

jkirschner-hashicorp:
@blake : I 100% agree that the ability to link to a specific parameter description is important and should be preserved (tables or not)

PR #12904, opened 2022-04-30
Support vault namespaces in connect CA

markan:
Follow on to some missed items from #12655

### Description

From an internal ticket "Support standard "Vault namespace in the
path" semantics for Connect Vault CA Provider"

Vault allows the namespace to be specified as a prefix in the path of
a PKI definition, but our usage of the Vault API includes calls that
don't support a namespaced key. In particular the sys.* family of
calls simply appends the key, instead of prefixing the namespace in
front of the path.

Unfortunately it is difficult to reliably parse a path with a
namespace; only vault knows what namespaces are present, and the '/'
separator can be inside a key name, as well as separating path
elements. This is in use in the wild; for example
'dc1/intermediate-key' is a relatively common naming schema.

Instead we add two new fields: RootPKINamespace and
IntermediatePKINamespace, which are the absolute namespace paths
'prefixed' in front of the respective PKI Paths.

[ ] manual testing

### Links
[Asana task](https://go.hashi.co/ui/edit/consul-vault-namespace)

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/config/builder.go
+++ b/agent/config/builder.go
@@ -697,16 +697,18 @@ func (b *builder) build() (rt RuntimeConfig, err error) {
 			"intermediate_cert_ttl": "IntermediateCertTTL",
 
 			// Vault CA config
-			"address":               "Address",
-			"token":                 "Token",
-			"root_pki_path":         "RootPKIPath",
-			"intermediate_pki_path": "IntermediatePKIPath",
-			"ca_file":               "CAFile",
-			"ca_path":               "CAPath",
-			"cert_file":             "CertFile",
-			"key_file":              "KeyFile",
-			"tls_server_name":       "TLSServerName",
-			"tls_skip_verify":       "TLSSkipVerify",
+			"address":                    "Address",
+			"token":                      "Token",
+			"root_pki_path":              "RootPKIPath",
+			"root_pki_namespace":         "RootPKINamespace",
+			"intermediate_pki_path":      "IntermediatePKIPath",
+			"intermediate_pki_namespace": "IntermediatePKINamespace",
+			"ca_file":                    "CAFile",
+			"ca_path":                    "CAPath",
+			"cert_file":                  "CertFile",
+			"key_file":                   "KeyFile",
+			"tls_server_name":            "TLSServerName",
+			"tls_skip_verify":            "TLSSkipVerify",
 
 			// AWS CA config
 			"existing_arn":   "ExistingARN",
--- a/agent/connect/ca/provider_vault.go
+++ b/agent/connect/ca/provider_vault.go
@@ -54,8 +54,9 @@ var ErrBackendNotMounted = fmt.Errorf("backend not mounted")
 var ErrBackendNotInitialized = fmt.Errorf("backend not initialized")
 
 type VaultProvider struct {
-	config *structs.VaultCAProviderConfig
-	client *vaultapi.Client
+	config        *structs.VaultCAProviderConfig
+	client        *vaultapi.Client
+	baseNamespace string
 
 	stopWatcher func()
 
@@ -109,6 +110,7 @@ func (v *VaultProvider) Configure(cfg ProviderConfig) error {
 	// same.
 	if config.Namespace != "" {
 		client.SetNamespace(config.Namespace)
+		v.baseNamespace = config.Namespace
 	}
 
 	if config.AuthMethod != nil {
@@ -282,7 +284,7 @@ func (v *VaultProvider) GenerateRoot() (RootResult, error) {
 	}
 
 	// Set up the root PKI backend if necessary.
-	rootPEM, err := v.getCA(v.config.RootPKIPath)
+	rootPEM, err := v.getCA(v.config.RootPKINamespace, v.config.RootPKIPath)
 	switch err {
 	case ErrBackendNotMounted:
 
@@ -307,7 +309,7 @@ func (v *VaultProvider) GenerateRoot() (RootResult, error) {
 		if err != nil {
 			return RootResult{}, err
 		}
-		resp, err := v.client.Logical().Write(v.config.RootPKIPath+"root/generate/internal", map[string]interface{}{
+		resp, err := v.writeNamespaced(v.config.RootPKINamespace, v.config.RootPKIPath+"root/generate/internal", map[string]interface{}{
 			"common_name": connect.CACN("vault", uid, v.clusterID, v.isPrimary),
 			"uri_sans":    v.spiffeID.URI().String(),
 			"key_type":    v.config.PrivateKeyType,
@@ -328,7 +330,7 @@ func (v *VaultProvider) GenerateRoot() (RootResult, error) {
 		}
 	}
 
-	rootChain, err := v.getCAChain(v.config.RootPKIPath)
+	rootChain, err := v.getCAChain(v.config.RootPKINamespace, v.config.RootPKIPath)
 	if err != nil {
 		return RootResult{}, err
 	}
@@ -359,7 +361,7 @@ func (v *VaultProvider) setupIntermediatePKIPath() error {
 		return nil
 	}
 
-	_, err := v.getCA(v.config.IntermediatePKIPath)
+	_, err := v.getCA(v.config.IntermediatePKINamespace, v.config.IntermediatePKIPath)
 	if err != nil {
 		if err == ErrBackendNotMounted {
 			err := v.mountNamespaced(v.config.IntermediatePKINamespace, v.config.IntermediatePKIPath, &vaultapi.MountInput{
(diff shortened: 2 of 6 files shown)


agent/config/builder.go
```diff
@@ -697,16 +697,18 @@ func (b *builder) build() (rt RuntimeConfig, err error) {
...
-			"token":                 "Token",
-			"root_pki_path":         "RootPKIPath",
-			"intermediate_pki_path": "IntermediatePKIPath",
-			"ca_file":               "CAFile",
-			"ca_path":               "CAPath",
-			"cert_file":             "CertFile",
-			"key_file":              "KeyFile",
-			"tls_server_name":       "TLSServerName",
-			"tls_skip_verify":       "TLSSkipVerify",
+			"address":                    "Address",
```
jkirschner-hashicorp, line 700: I noticed some config params (`AuthMethod`, `Namespace`) aren't here. Do they need to be?

agent/connect/ca/provider_vault.go
```diff
@@ -285,7 +285,8 @@ func (v *VaultProvider) GenerateRoot() (RootResult, error) {
 	rootPEM, err := v.getCA(v.config.RootPKIPath)
 	switch err {
 	case ErrBackendNotMounted:
-		err := v.client.Sys().Mount(v.config.RootPKIPath, &vaultapi.MountInput{
+
+		err := v.mountNamespaced(v.config.RootPKINamespace, v.config.RootPKIPath, &vaultapi.MountInput{
```
jkirschner-hashicorp, line 289: We may want to add a comment in here somewhere (maybe in `mountNamespaced` or `unmountNamespaced`?) about how `RootPKINamespace` and `IntermediatePKINamespace` interact with `Namespace`.

I'm a bit confused myself.

If `Namespace` was set to `foo`, and `RootPKINamespace` was set to `foo`, wouldn't the API call actually go to something in the non-existent namespace `foo/foo`? My understanding is that `Namespace` is setting the `X-Vault-Namespace` header, and then any namespace "in the path" ([like set in makePathHelper](https://github.com/hashicorp/consul/pull/12904/files#diff-7bf30611a760296e2e6ffcd7fe955b0eb8317fe0a43c470b8b52d311af4a2515R743)) will be treated as relative to the namespace set in `X-Vault-Namespace`.

If we aren't sure, we probably want to test this case and make sure we understand the behavior... I'm not sure if `RootPKINamespace` and `IntermediatePKINamespace` currently "overrode" `Namespace` as we intended.

jkirschner-hashicorp (commented):
@markan : a few questions we can discuss Monday!

website/content/docs/connect/ca/vault.mdx
```diff
@@ -136,6 +136,9 @@ The configuration options are listed below.
   must contain a valid chain, where each certificate is followed by the certificate
   that authorized it.
 
+- `RootPKINamespace` / `root_pki_namespace` (`string: <optional>`) - The absolute namespace
```
jkirschner-hashicorp, line 139: We should mention the version that this exists in (Consul 1.12.1 and later).

website/content/docs/connect/ca/vault.mdx
```diff
@@ -145,6 +148,9 @@ The configuration options are listed below.
   When WAN Federation is enabled, every secondary
   datacenter must specify a unique `intermediate_pki_path`.
 
+- `IntermediatePKINamespace` / `intermedial_pki_namespace` (`string: <optional>`) - The absolute namespace
```
jkirschner-hashicorp, line 151: We should mention the version that this exists in (Consul 1.12.1 and later).

agent/connect/ca/provider_vault.go
markan: Good catch as it turns out. This was marked as a draft because I hadn't tested that case, and there were a few fixes needed to get the behavior sorted out. Tested and seems ok now.

agent/connect/ca/provider_vault.go
```diff
@@ -709,6 +722,66 @@ func (v *VaultProvider) Stop() {
...
 
+// We use raw path here
+func (v *VaultProvider) mountNamespaced(namespace, path string, mountInfo *vaultapi.MountInput) error {
+	defer v.client.SetNamespace(v.baseNamespace)
+	if namespace != "" {
+		v.client.SetNamespace(namespace)
+	}
+
+	r := v.client.NewRequest("POST", fmt.Sprintf("/v1/sys/mounts/%s", path))
+	if err := r.SetJSONBody(mountInfo); err != nil {
```
boxofrad, line 733: I see that we use raw `client.RawRequest` and sort-of manually interrogate Vault's HTTP API in lots of places in this provider. I'm guessing this is because we can no longer use `client.Sys` because of the need to inject the namespace into the path? Could we use `client.Logical` instead, though? (sorry in advance if I've missed what you were trying to say in your PR description)

agent/connect/ca/provider_vault.go
```diff
@@ -709,6 +722,66 @@ func (v *VaultProvider) Stop() {
 
 func (v *VaultProvider) PrimaryUsesIntermediate() {}
 
+// We use raw path here
+func (v *VaultProvider) mountNamespaced(namespace, path string, mountInfo *vaultapi.MountInput) error {
+	defer v.client.SetNamespace(v.baseNamespace)
+	if namespace != "" {
+		v.client.SetNamespace(namespace)
+	}
```
boxofrad, line 730: Do we need to guard the provider with a mutex to prevent a race condition here?

markan: I'd convinced myself that this couldn't happen. But seems safer to have the mutex, so added it.

agent/connect/ca/provider_vault.go
boxofrad: Caught up about this in-person (well, Zoom 😅) — the logical API client isn't suitable for this.

boxofrad:
@markan as discussed (again, I'm totally fine with copy-pasta 🍝) here are a couple of options for extracting the concurrency-safe namespace swapping pattern into a helper method:

```Go
func (v *VaultProvider) setNamespace(namespace string) func() {
	v.mu.Lock()
	v.client.SetNamespace(namespace)
	return func() {
		v.client.SetNamsepace(v.baseNamspace)
		v.mu.Unlock()
	}
}

func (v *VaultProvider) someMethod(namespace string, arg2 any) {
	restoreNamespace := v.setNamespace(namespace)
	defer restoreNamespace()

	// or..

	defer v.setNamsepace(namespace)()
}
```

```Go
func (v *VaultProvider) withNamespace(namespace string, fn func()) func() {
	v.mu.Lock()
	v.client.SetNamespace(namespace)
	fn()
	v.client.SetNamsepace(v.baseNamspace)
	v.mu.Unlock()
}

func (v *VaultProvider) someMethod(namespace string, arg2 any) someType {
	var result someType
	v.withNamespace(namespace, func() {
		result = v.client.SomeMethod()
	})
	return result
}

**Review by boxofrad — APPROVED** [review]
Looks good to me, nice work! 🎉

File: website/content/docs/connect/ca/vault.mdx [hunk 7]
```diff
@@ -136,6 +136,9 @@ The configuration options are listed below.
   must contain a valid chain, where each certificate is followed by the certificate
   that authorized it.
 
+- `RootPKINamespace` / `root_pki_namespace` (`string: <optional>`) - The absolute namespace
+  that the RootPKIPath is in. Setting this overrides the Namespace option for the RootPKIPath. Introduced in 1.12.1
```
trujillo-adam, line 140:
```suggestion
  that the `RootPKIPath` is in. Setting this overrides the `Namespace` option for the `RootPKIPath`. Introduced in 1.12.1
```
We should use the code font if we are referring to actual configuration options or files, as opposed to the general word for the entity (i.e., _a_ namespace vs. _the_ `namespace` configuration).

website/content/docs/connect/ca/vault.mdx
```diff
@@ -145,6 +148,9 @@ The configuration options are listed below.
   When WAN Federation is enabled, every secondary
   datacenter must specify a unique `intermediate_pki_path`.
 
+- `IntermediatePKINamespace` / `intermedial_pki_namespace` (`string: <optional>`) - The absolute namespace
+  that the IntermediatePKIPath is in. Setting this overrides the Namespace option for the IntermediatePKIPath. Introduced in 1.12.1
```
trujillo-adam, line 152:
```suggestion
  that the `IntermediatePKIPath` is in. Setting this overrides the `Namespace` option for the IntermediatePKIPath. Introduced in 1.12.1
```

trujillo-adam (commented):
One bit of feedback about formatting.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/662352.

Issue #12813, opened 2022-04-18
[middleware] `RequestRecorder` needs to report < 1ms calls as well


acpana:
### overview

Today, the `RequestRecorder` records service RPC calls. If those take less than `< 0ms`, the recorder marks them as `0ms`.

### proposal

Make the `RequestRecorder` report the `elapsed` time for service calls that take less than `< 0ms` as decimal values, instead of 0.

#### testing

If you take this on, I'd kindly advise to start in a TDD fashion. Write a test in `metrics_test` to look at, say, `Status.Ping` RPC calls. These will almost always come at < 0ms so they will be reported as `0 ms`. Let the test fail, then do the necessary logic changes in `RequestRecorder.Record`.

#### docs

Today, the docs call this out. If this is worked on, please make the necessary changes in the docs as well.

Gabrielopesantos:
Hello, I have never used consul but this seems an interesting issue. Can I take it?

Gabrielopesantos:
Hello, I have opened up a draft pull request with a proposal for the issue but also looking for feedback on an issue that I am having. Let me know if I should make it ready to be reviewed. Thank you.

PR #12905, opened 2022-04-30
Middlware: `RequestRecorder` reports calls bellow 1ms as decimal value

Gabrielopesantos:
### Description
`RequestRecorder` records service RPC calls as a decimal number when when are `<1ms`.
#12813 

* Test created

A test has been created, however, I am not sure about the test name and arrange/preparation steps.
Also, with the changes the value is supposed to have 3 decimal digits, however it is returned as the following example: 
new_rpc_metrics_2_rpc_server_call{errored="false",leader="true",method="Status.Ping",request_type="read",rpc_type="net/rpc",quantile="0.5"} 0.08299999684095383
Do you have any idea why this is happening?

### PR Checklist

* [X] updated test coverage
* [ ] not a security concern

### DIFF
--- a/agent/metrics_test.go
+++ b/agent/metrics_test.go
@@ -133,7 +133,7 @@ func assertMetricExistsWithValue(t *testing.T, respRec *httptest.ResponseRecorde
 	}
 }
 
-func assertMetricsWithLabelHasValueDifferentFrom(t *testing.T, respRec *httptest.ResponseRecorder, label, labelValue string, value string) {
+func assertMetricsWithLabelHaveValueDifferentFrom(t *testing.T, respRec *httptest.ResponseRecorder, label, labelValue string, value string) {
 	if respRec.Body.String() == "" {
 		t.Fatalf("Response body is empty.")
 	}
@@ -149,11 +149,10 @@ func assertMetricsWithLabelHasValueDifferentFrom(t *testing.T, respRec *httptest
 		if strings.Contains(line, labelWithValueTarget) {
 			s := strings.SplitN(line, " ", 2)
 			if s[1] == value {
-				t.Fatal("Should have not been zero")
+				t.Fatalf("Metric with label provided \"%s:%s\" has the same outcome as the provided value `%s`", label, labelValue, value)
 			}
 		}
 	}
-
 }
 
 func assertMetricNotExists(t *testing.T, respRec *httptest.ResponseRecorder, metric string) {
@@ -444,7 +443,7 @@ func TestHTTPHandlers_AgentMetrics_CACertExpiry_Prometheus(t *testing.T) {
 
 }
 
-func TestNotSureAboutTheName(t *testing.T) {
+func TestMetricsOutput(t *testing.T) {
 	skipIfShortTesting(t)
 
 	t.Run("RPC calls with elapsed time below 1ms are reported as decimal", func(t *testing.T) {
@@ -468,6 +467,6 @@ func TestNotSureAboutTheName(t *testing.T) {
 		respRec := httptest.NewRecorder()
 		recordPromMetrics(t, a, respRec)
 
-		assertMetricsWithLabelHasValueDifferentFrom(t, respRec, "method", "Status.Ping", "0")
+		assertMetricsWithLabelHaveValueDifferentFrom(t, respRec, "method", "Status.Ping", "0")
 	})
 }
--- a/agent/metrics_test.go
+++ b/agent/metrics_test.go
@@ -148,8 +148,8 @@ func assertMetricsWithLabelHasValueDifferentFrom(t *testing.T, respRec *httptest
 
 		if strings.Contains(line, labelWithValueTarget) {
 			s := strings.SplitN(line, " ", 2)
-			if s[1] != value {
-				t.Fatal("Should have not been a zero")
+			if s[1] == value {
+				t.Fatal("Should have not been zero")
 			}
 		}
 	}
--- a/agent/metrics_test.go
+++ b/agent/metrics_test.go
@@ -44,7 +44,7 @@ func assertMetricExists(t *testing.T, respRec *httptest.ResponseRecorder, metric
 	}
 }
 
-// assertMetricExistsWithLabels looks in the prometheus metrics reponse for the metric name and all the labels. eg:
+// assertMetricExistsWithLabels looks in the prometheus metrics response for the metric name and all the labels. eg:
 // new_rpc_metrics_rpc_server_call{errored="false",method="Status.Ping",request_type="unknown",rpc_type="net/rpc"}
 func assertMetricExistsWithLabels(t *testing.T, respRec *httptest.ResponseRecorder, metric string, labelNames []string) {
 	if respRec.Body.String() == "" {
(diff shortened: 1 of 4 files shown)


agent/metrics_test.go
```diff
@@ -420,3 +442,31 @@ func TestHTTPHandlers_AgentMetrics_CACertExpiry_Prometheus(t *testing.T) {
...
+		a := StartTestAgent(t, TestAgent{HCL: hcl})
+		defer a.Shutdown()
+
+		var out struct{}
+		_ = a.RPC("Status.Ping", struct{}{}, &out)
+
+		respRec := httptest.NewRecorder()
+		recordPromMetrics(t, a, respRec)
+
+		assertMetricsWithLabelHaveValueDifferentFrom(t, respRec, "method", "Status.Ping", "0")
```
acpana, line 470: Thanks for following the testing framework we outlined!

Let's please move this check `assertMetricsWithLabel...` in [`TestAgentOneTwelve...`](https://github.com/hashicorp/consul/pull/12905/files#diff-472e516d540a393b43170efc09098871a521b8b8ed4788089412829311ff81c5L148) since this would be an improvement

agent/metrics_test.go
```diff
@@ -133,6 +133,28 @@ func assertMetricExistsWithValue(t *testing.T, respRec *httptest.ResponseRecorde
 	}
 }
 
+func assertMetricsWithLabelHaveValueDifferentFrom(t *testing.T, respRec *httptest.ResponseRecorder, label, labelValue string, value string) {
```
acpana, line 136: [nit]:

```suggestion
func assertMetricsWithLabelIsNonZero(t *testing.T, respRec *httptest.ResponseRecorder, label, labelValue string) {
```

and change the check on line 151 to look at `"0"` specifically.

agent/rpc/middleware/interceptors.go
```diff
@@ -49,7 +50,11 @@ func NewRequestRecorder(logger hclog.Logger, isLeader func() bool, localDC strin
 }
 
 func (r *RequestRecorder) Record(requestName string, rpcType string, start time.Time, request interface{}, respErrored bool) {
-	elapsed := time.Since(start).Milliseconds()
+	elapsedNs := time.Since(start)
+	elapsed := float64(elapsedNs.Milliseconds())
+	if elapsed == 0 {
+		elapsed = math.Round(float64(elapsedNs.Microseconds())) / math.Pow(10, 3)
```
acpana, line 56: IIUC, you were asking about the behavior of `.Round` not actually rounding in the PR description.

I think for Round to work like we want to it would have to look like:

```
math.Round(float64(elapsedNs.Microseconds()) / math.Pow(10, 3))
```

but if we did that then the answer is always 0 per the spec:

https://pkg.go.dev/math#Round

acpana (commented):
Hey @Gabrielopesantos thanks loads for taking this on 💯 🥇 and thank you for your patience with our response.

The PR is almost there and I left some guidance. Let me know if my comments and unclear please.

agent/rpc/middleware/interceptors.go
```diff
@@ -49,7 +50,11 @@ func NewRequestRecorder(logger hclog.Logger, isLeader func() bool, localDC strin
 }
 
 func (r *RequestRecorder) Record(requestName string, rpcType string, start time.Time, request interface{}, respErrored bool) {
-	elapsed := time.Since(start).Milliseconds()
+	elapsedNs := time.Since(start)
+	elapsed := float64(elapsedNs.Milliseconds())
+	if elapsed == 0 {
```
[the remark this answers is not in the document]
Gabrielopesantos: Interesting idea but, unfortunately, does not seem to work as we have to convert the value to float32 when passing to the `RecordFunc`.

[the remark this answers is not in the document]
Gabrielopesantos: Interesting idea, I have tested this approach but it does not seem to work. The string values are only represented as two decimal cases when printed, but they have to be converted to `float32` in order to be passed to `RecordFunc`.

I am not sure what would be the best approach, the keep the changes minimal I think the best would be to only convert to decimal values value 0 (as done the PR).

agent/metrics_test.go
```diff
@@ -203,6 +225,32 @@ func TestAgent_OneTwelveRPCMetrics(t *testing.T) {
...
+		defer a.Shutdown()
+
+		var out struct{}
+		err := a.RPC("Status.Ping", struct{}{}, &out)
+		require.NoError(t, err)
+
+		respRec := httptest.NewRecorder()
+		recordPromMetrics(t, a, respRec)
+
+		assertMetricsWithLabelIsNonZero(t, respRec, "method", "Status.Ping")
```
acpana, line 251: Awesome, thanks for adding this in!

We don't actually need a full `t.Run` to use this assertion. Let's move this line (251) right after the 226 line and remove this test since this one and the above are essentially testing the same use case.

agent/rpc/middleware/interceptors.go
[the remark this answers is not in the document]
acpana: let me pull this locally and have a look -- I want to make sure we get the right values 🙈 👀

acpana (commented):
We're almost there! It seems the linter is complaining so please `gofmt -s -w` the modified `.go` files 🙏🏼

agent/rpc/middleware/interceptors.go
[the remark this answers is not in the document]
Gabrielopesantos: Alright, let me know if it works as expected or if we will need to try some other approach.

Pushed a commit with the suggested changes and run the `gofmt` file but didn't point point any issue.

acpana (approved):
I think this looks good to me!

After looking more into why the `.Round` doesn't seem to do what we want, I think it has to do with how `float64` is implemented. I came to this realization by going off of this [stackoverflow link](https://stackoverflow.com/questions/39544571/golang-round-to-nearest-0-05).

I'll leave it to someone else to merge this in 🥇 ; thanks for working on this and your patience!!

agent/rpc/middleware/interceptors.go
```diff
@@ -49,7 +50,11 @@ func NewRequestRecorder(logger hclog.Logger, isLeader func() bool, localDC strin
 }
 
 func (r *RequestRecorder) Record(requestName string, rpcType string, start time.Time, request interface{}, respErrored bool) {
-	elapsed := time.Since(start).Milliseconds()
+	elapsedNs := time.Since(start)
```
kisunji, line 53: I think it would be cleaner to start with `time.Since(start).Microseconds()` and convert it to a float64 milisecond rather than "if milliseconds rounds to 0, then use microseconds".

Gabrielopesantos: Thanks for the feedback. Your suggestions is easier to understand and requires less code (If I actually understood it and implemented it correctly).

agent/rpc/middleware/interceptors.go
```diff
@@ -49,7 +50,8 @@ func NewRequestRecorder(logger hclog.Logger, isLeader func() bool, localDC strin
 }
 
 func (r *RequestRecorder) Record(requestName string, rpcType string, start time.Time, request interface{}, respErrored bool) {
-	elapsed := time.Since(start).Milliseconds()
+	elapsed := time.Since(start).Microseconds()
+	elapsedMs := float64(elapsed) / math.Pow(10, 3)
```
kisunji, line 54: I think `float64` was originally used because of `math.Pow` but isn't necessary if we simply use 1000.

The following code should keep its precision
```suggestion
	elapsedMs := float32(elapsed) / 1000
```

kisunji (approved):
One minor comment but looks good! Thanks for the contribution

agent/rpc/middleware/interceptors.go
Gabrielopesantos: You are right. Thanks

dhiaayachi (approved):
LGTM!

PR #12908, opened 2022-05-02
build: speed up linting by 1.5x

rboyer:
### Description

Calling the `make lint` target with an empty `golangci-lint` cache takes about 1m. I noticed that in CI we had been upping the concurrency from the default to 2, so I doubled that, and made it the general default so it applied to the `make lint` target which now takes about 35s for me.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -191,7 +191,7 @@ jobs:
       - run:
           name: lint
           command: &lintcmd |
-            golangci-lint run --build-tags="$GOTAGS" -v --concurrency 2
+            golangci-lint run --build-tags="$GOTAGS" -v
       - run:
           name: lint api
           working_directory: api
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -175,7 +175,7 @@ jobs:
         default: ""
     docker:
       - image: *GOLANG_IMAGE
-    resource_class: large
+    resource_class: xlarge
```
rboyer, line 178: Bumping this takes the key lint run part (for the main consul module) down from:

    INFO Execution took 2m12.841127973s 

to

    INFO Execution took 1m41.403607416s

eculver (approved):
LGTM! I kicked the integration test failures too btw.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/663421.

PR #12911, opened 2022-05-02
[sync oss] api: add peering api module

FFMMM:
OSS sync with ENT over { be63e99c3 f1aca2618 1c549d814 }

### DIFF
--- a/agent/peering_endpoint_test.go
+++ b/agent/peering_endpoint_test.go
@@ -195,6 +195,41 @@ func TestHTTP_Peering_Initiate(t *testing.T) {
 	})
 }
 
+func TestHTTP_Peering_MethodNotAllowed(t *testing.T) {
+	if testing.Short() {
+		t.Skip("too slow for testing.Short")
+	}
+
+	t.Parallel()
+	a := NewTestAgent(t, "")
+
+	testrpc.WaitForTestAgent(t, a.RPC, "dc1")
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	// Insert peerings directly to state store.
+	// Note that the state store holds reference to the underlying
+	// variables; do not modify them after writing.
+	foo := &pbpeering.PeeringWriteRequest{
+		Peering: &pbpeering.Peering{
+			Name:                "foo",
+			State:               pbpeering.PeeringState_INITIAL,
+			PeerCAPems:          nil,
+			PeerServerName:      "fooservername",
+			PeerServerAddresses: []string{"addr1"},
+		},
+	}
+	_, err := a.rpcClientPeering.PeeringWrite(ctx, foo)
+	require.NoError(t, err)
+
+	req, err := http.NewRequest("PUT", "/v1/peering/foo", nil)
+	require.NoError(t, err)
+	resp := httptest.NewRecorder()
+	a.srv.h.ServeHTTP(resp, req)
+	require.Equal(t, http.StatusMethodNotAllowed, resp.Code)
+}
+
 func TestHTTP_Peering_Read(t *testing.T) {
 	if testing.Short() {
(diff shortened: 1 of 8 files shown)


agent/peering_endpoint_test.go
```diff
@@ -257,6 +292,76 @@ func TestHTTP_Peering_Read(t *testing.T) {
...
+	a := NewTestAgent(t, "")
+
+	testrpc.WaitForTestAgent(t, a.RPC, "dc1")
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	// Insert peerings directly to state store.
+	// Note that the state store holds reference to the underlying
+	// variables; do not modify them after writing.
```
freddygv, line 310: This might be a copy-paste error, the write is going through the gRPC client and not directly into the state store 

I'm also not sure that the next two lines apply, given that it's not a direct state store write

agent/peering_endpoint_test.go
```diff
@@ -257,6 +292,76 @@ func TestHTTP_Peering_Read(t *testing.T) {
...
+		require.Equal(t, http.StatusNotFound, resp.Code)
+	})
+
+	t.Run("delete a token that does not exist", func(t *testing.T) {
+		req, err := http.NewRequest("DELETE", "/v1/peering/baz", nil)
+		require.NoError(t, err)
+		resp := httptest.NewRecorder()
+		a.srv.h.ServeHTTP(resp, req)
+
+		// TODO(peering): it may be a security concern, but do we want to say 404 here?
```
freddygv, line 360: No it shouldn't return 404 here, we return 200 for non-existent deletions in our APIs

api/peering.go
```diff
@@ -0,0 +1,213 @@
...
+
+const (
+	// Undefined represents an unset value for PeeringState during
+	// writes.
+	UNDEFINED PeeringState = 0
+	// INITIAL Initial means a Peering has been initialized and is awaiting
+	// acknowledgement from a remote peer.
+	INITIAL PeeringState = 1
+	// Active means that the peering connection is active and healthy.
+	// ACTIVE PeeringState = 2
```
freddygv, line 19: Should this line be commented out?

api/peering_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// basic checks
+	require.NoError(t, err3)
+	require.NotEmpty(t, wm3)
+
+	// at first the token will be undefined
+	require.Equal(t, UNDEFINED, PeeringState(respi.Status))
+
+	// wait for the peering backend to finish the peering connection
+	time.Sleep(2 * time.Second)
```
freddygv, line 79: We should prefer using `retry.Run` over Sleep when waiting for some state. If the peering is ready sooner than 2s then the test will run faster. Also if it takes longer than 2s there's some flexibility there before the test fails because the default retry is for 7s.

freddygv (approved):
This looks good, thanks for syncing this over. Now it'll be available for importing and building on top of 🚀 

I left a couple requests, but will approve in advance

FFMMM:
interestingly, the integ tests pass locally with:


```
ok  	github.com/hashicorp/consul/test/integration/connect/envoy	819.145s
```

🤔 seeing if circle is having issues

FFMMM:
hey @freddygv, thanks for reviewing these 💯 

I took note of all of your comments above and will follow up on them in a polish up PR.

For now, I'd like to get this in for parity and to prevent any mergdle headaches if that's ok.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/657755.

api/peering.go
FFMMM: I was going to keep this in since there is a commented out line to use that state for better testing

PR #12914, opened 2022-05-03
add the ability to specify a path prefix for when consul is behind a …

funkiestj:
…reverse proxy / API gateway

### Description
I am using consul behind an nginx reverse proxy (API gateway).  different services are identified by different path prefixes.

nginx strips off the path prefix and forwards the request to consul.

unit test verifies that api.Config parsing is working as expected.

I am running the code from my forked consul in my docker lab.

### Links
fixes issue https://github.com/hashicorp/consul/issues/10354

This PR was previously submitted by ekhabarov but he did not follow through on the review feedback to add some comments.

https://github.com/hashicorp/consul/pull/10905

### PR Checklist

* [x ] updated test coverage
* [x ] external facing docs updated
* [ x] not a security concern
* [ x] checklist [folder](./../docs/config) consulted


Amier3:
Hey @funkiestj 

Thanks a ton for pushing #10905  over the hump! We'll get this reviewed shortly 👍

api/api_test.go
```diff
@@ -1119,6 +1119,65 @@ func TestAPI_GenerateEnvHTTPS(t *testing.T) {
...
 }
 
+// TestAPI_PrefixPath() validates that Config.Address is split into
+// Config.Address and Config.PathPrefix as expected.  If we want to add end to
+// end testing in the future this will require configuring and running an
+// API gateway / reverse proxy (e.g. nginx)
+func TestAPI_PrefixPath(t *testing.T) {
+	t.Parallel()
+
+	// reverse proxy PathPrefix should be separated out
```
eculver, line 1129: I'm curious if this might be simpler to express with a table of inputs and expected outputs? This is basically what this is already doing except for the examples without a prefix are separated below. If we defined a `testCase` type that looked like:
```
type testCase struct {
    name   string
    addr   string
    expect string
}
```

You could define all your cases like:
```
cases := []testCase{
    {
        name: "with http and prefix",
        addr: ""http://reverse.proxy.com/consul/path/prefix",
        expectAddr: "reverse.proxy.com",
        expectPrefix: "/consul/path/prefix",
    },
    ...
}
```
and then loop over cases like:
```
for _, tc := range cases {
    t.Run(t, func(t *testing.T) {
        c := &Config{Address: tc.addr}
        client, err := NewClient(c)
        require.NoError(t, err)
        require.Equal(t, tc.expectAddr, client.config.Address) # using the require package here which is used elsewhere
        require.Equal(t, tc.expectPrefix, client.config.PathPrefix)
    })
}
```

Something like that? We follow a pattern like this in a lot of our tests (example [here](https://github.com/hashicorp/consul/blob/main/api/agent_test.go#L1794-L1828) and [here](https://github.com/hashicorp/consul/blob/main/api/health_test.go#L60-L191)) when it's easy to express input/output cases. At the very least, it makes the test code little easier to read all at once, and a little more concise, but not necessarily "better" than what you have already. 

If that makes sense to you cool, if not, it's just as suggestion. I just spat that out and didn't try to run it or anything but should illustrate what I'm talking about. Take from it what you will.

eculver (commented):
Thanks for following up here! This is looking good. I'd like to hear your thoughts on the `url.Parse` thing and then the comment on the tests is just a non-blocking suggestion.

api/api.go
```diff
@@ -712,6 +717,18 @@ func NewClient(config *Config) (*Client, error) {
 			return nil, fmt.Errorf("Unknown protocol scheme: %s", parts[0])
 		}
 		config.Address = parts[1]
+
+		// separate out a reverse proxy prefix, if it is present.
+		// NOTE: Rewriting this code to use url.Parse() instead of
+		// strings.SplitN() breaks existing test cases.
+		switch parts[0] {
+		case "http", "https":
+			parts := strings.SplitN(parts[1], "/", 2)
```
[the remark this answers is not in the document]
funkiestj: I spent quite a while trying to refactor and use url.Parse() but when I did, my code worked but I broke various unit tests.

I've come to the conclusion that the authors chose strings.SplitN() in the original code for a reason -- because they wanted to support some inputs that give url.Parse() indigestion.

api/api_test.go
funkiestj: sure, I'll do that and push an update.

funkiestj: done

funkiestj:
ping

funkiestj:
I think I've addressed all review issues.  Let me know if I've missed something.

assuming all review comments have been addressed, what is the expected timeline for getting this merged?

api/api.go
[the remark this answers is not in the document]
eculver: Fair enough. Thanks for adding the note!

eculver (approved):
Nice! Thanks for following up here! We can merge it right away assuming we get a green from all our checks. I can help with that.

PR #12923, opened 2022-05-03
Add a default max number of retries when joining

freddygv:
### Description
Preivously Consul agents would retry joining a LAN pool indefinitely.
This is problematic when certain errors are encountered while joining.
For example, if the target node is reachable but the push/pull process
fails then there could continously be expensive push/pulls that lead to
errors.

Given that agents typically join Consul servers, when this class of
error is encountered then there can be multiple client agents stuck in
an erroring loop needlessly consuming server resources.

With the default 30s retry-join interval, 10 attempts leads to roughly a
5 minute timeout before the agent shuts down.

* Ran `consul agent -dev -retry-join 8.8.8.8 -retry-interval 1s`
* Agent shut down after 10 "Failed to join" errors.

### PR Checklist

* [ ] updated test coverage
  * No unit tests were updated, should there be one?
* [x] not a security concern

### DIFF
--- a/.changelog/12923.txt
+++ b/.changelog/12923.txt
@@ -1,0 +1,4 @@
+```release-note:breaking-change
+agent: Cap the number of attempts when joining the LAN gossip pool at 10. Agents will no longer attempt to retry indefinitely.
+The previous default can be restored by setting the retry-max agent option.
+```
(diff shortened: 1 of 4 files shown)


mkeeler (commented):
LGTM, can you also update the upgrade specific docs to note the change and how it could affect someone who was maybe taking more than 5m to join a cluster.

PR #12928, opened 2022-05-03
chore(ci): exempt backport PRs from changelog-checker

DanStough:
### Description
Without further changes to `backport-assistant`, backport PRs will never contain the original `pr/no-changelog` label. This will cause GitHub checks on the PR to fail ([example](https://github.com/hashicorp/consul/pull/12926)), which will eventually also block auto-merging pull requests. Adding the `pr/no-changelog` label should be something that is handled on the front-end of the originating PR before merging. This PR exempts the bot PRs from the check. We can decide if upstream changes to `backport-assistant` are needed at a later point.

N/A

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/.github/workflows/changelog-checker.yml
+++ b/.github/workflows/changelog-checker.yml
@@ -12,8 +12,8 @@ on:
 jobs:
   # checks that a .changelog entry is present for a PR
   changelog-check:
-    # If there  a `pr/no-changelog` label we ignore this check
-    if: "!contains(github.event.pull_request.labels.*.name, 'pr/no-changelog')"
+    # If there  a `pr/no-changelog` label we ignore this check. Also, we ignore PRs created by the bot assigned to `backport-assistant`
+    if: "! ( contains(github.event.pull_request.labels.*.name, 'pr/no-changelog') || github.event.pull_request.user.login == 'hc-github-team-consul-core' )" 
     runs-on: ubuntu-latest
 
     steps:


.github/workflows/changelog-checker.yml
```diff
@@ -12,8 +12,8 @@ on:
 jobs:
   # checks that a .changelog entry is present for a PR
   changelog-check:
-    # If there  a `pr/no-changelog` label we ignore this check
-    if: "!contains(github.event.pull_request.labels.*.name, 'pr/no-changelog')"
+    # If there  a `pr/no-changelog` label we ignore this check. Also, we ignore PRs created by the bot assigned to `backport-assistant`
+    if: "! ( contains(github.event.pull_request.labels.*.name, 'pr/no-changelog') || github.event.pull_request.user.login == 'hc-github-team-consul-core' )"
```
kyhavlov, line 16: Could this check by the backport labels explicitly instead? It's probably fine to bypass the check on bot-created PRs altogether but that might be a more specific way to do this

DanStough: Hmm, I'm having trouble picturing how that would work. AFAIK we do want the check to run on the initial PR that has the `backport/X` labels, and right now there are no labels attached to the automated PRs generated. Maybe to your point, I think a better solution would be adding a label like `automated-backport` to all the generated PRs, but right now that feature doesn't exist in `backport-assistant` and it doesn't export the PRs numbers it creates.

kyhavlov: Ok, that makes sense. I think it's fine to avoid to the changelog check on bot PRs for now - I was thinking of a case like the Dependabot PRs where automatically opened ones might still want a changelog depending on the contents.

kyhavlov (approved):

.github/workflows/changelog-checker.yml
DanStough: Oh, didn't know we were using Dependabot. In my experience that is usually it's own user though, so hopefully this wouldn't affect it. If it does we can go the route of adding labels to automated PRs from `backport-assistant`.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/665045.

PR #12934, opened 2022-05-04
api: add PeeeringList, polish

FFMMM:
### Description
The main reason for this PR is to add some polish (more tests, punted feedback) for the peering changes in the `api` module. While at it, I also added the `PeeringList` endpoint.

* updated unit tests

### TODO:
* [ ] ENT companion PR link


api/peering.go
```diff
@@ -9,14 +9,12 @@ import (
 type PeeringState int32
 
 const (
-	// Undefined represents an unset value for PeeringState during
+	// UNDEFINED represents an unset value for PeeringState during
```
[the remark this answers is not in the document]
kisunji: I know I asked to remove the prefix in your previous PR 😅 
But let's make it idiomatic Go and avoid ALLCAPS

api/peering.go
```diff
@@ -9,14 +9,12 @@ import (
...
 	// writes.
 	UNDEFINED PeeringState = 0
 	// INITIAL Initial means a Peering has been initialized and is awaiting
 	// acknowledgement from a remote peer.
 	INITIAL PeeringState = 1
-	// Active means that the peering connection is active and healthy.
```
kisunji, line 18: Should fix above comment too (INITIAL Initial)

api/peering.go
```diff
@@ -96,14 +99,21 @@ func (c *Client) Peerings() *Peerings {
...
 		return nil, nil, fmt.Errorf("peering name cannot be empty")
 	}
 
-	req := p.c.newRequest("GET", fmt.Sprintf("/v1/peering/%s", name))
+	var url string
+	if peeringReq.Partition != "" {
+		url = fmt.Sprintf("/v1/peering/%s?partition=%s", peeringReq.Name, peeringReq.Partition)
+	} else {
+		url = fmt.Sprintf("/v1/peering/%s", peeringReq.Name)
+	}
```
kisunji, line 112: I think the convention is to pass partition through `QueryOptions`. We shouldn't need to use a PeeringRequest

api/peering.go
```diff
@@ -211,3 +221,31 @@ func (p *Peerings) Initiate(ctx context.Context, i PeeringInitiateRequest, wq *W
 
 	return &out, wm, nil
 }
+
+func (p *Peerings) List(ctx context.Context, plr PeeringListRequest, q *QueryOptions) ([]*Peering, *QueryMeta, error) {
```
kisunji, line 225: I don't think we need `PeeringListRequest` here at all; the handler only expects EnterpriseMeta from the request which is set by `*QueryOptions`

api/peering_test.go
```diff
@@ -2,14 +2,155 @@ package api
...
+	t.Run("call Read with no name", func(t *testing.T) {
+		resp, qm, err := peerings.Read(ctx, PeeringRequest{}, nil)
+
+		// basic checks
+		require.EqualError(t, err, "peering name cannot be empty")
+		require.Empty(t, qm)
+		require.Empty(t, resp)
+	})
+
+	t.Run("read token that does not exist on server", func(t *testing.T) {
```
kisunji, line 52:
```suggestion
	t.Run("read peer that does not exist on server", func(t *testing.T) {
```

api/peering_test.go
```diff
@@ -2,14 +2,155 @@ package api
...
+		resp, qm, err := peerings.Read(ctx, PeeringRequest{Name: "peer1"}, nil)
+
+		// basic checks
+		require.NotNil(t, err) // 404
+		require.Empty(t, qm)
+		require.Empty(t, resp)
+	})
```
kisunji, line 59: I think we should add a test for a successful Read

api/peering_test.go
```diff
@@ -2,14 +2,155 @@ package api
...
+
+}
+
+// TestAPI_Peering_List
+func TestAPI_Peering_List(t *testing.T) {
```
kisunji, line 64: This test has subtests that mutate a shared state; we should either call `c, s := makeClientWithCA(t)` once in every test or stop using `t.Run`

api/peering.go
kisunji: Maybe we could leave the request as-is in case we ever need to add to it in the future but get rid of the fields so it's clear they're not used

api/peering_test.go
```diff
@@ -2,14 +2,155 @@ package api
...
+		require.NotEmpty(t, qm)
+		require.NotEmpty(t, peering)
+
+		peeringsList, qm, err := peerings.List(ctx, PeeringListRequest{}, nil)
+
+		// basic checks
+		require.NoError(t, err)
+		require.NotEmpty(t, qm)
+
+		require.True(t, peerExistsInPeerListings(peering, peeringsList), "expected to find peering in list response")
```
kisunji, line 103: We should simply do a `len(peeringsList)` check and compare names; `peerExistsInPeerListings` is too brittle and unnecessary IMO (we might forget to update it when we add fields to Peering)

api/peering_test.go
```diff
@@ -2,14 +2,155 @@ package api
...
+
+		// modify peering to some non existent peering
+		peering.Name = "not_peer1"
+
+		require.False(t, peerExistsInPeerListings(peering, peeringsList), "did not expect to find peering in list response")
```
kisunji, line 108: I don't see much value in these lines as you're testing `peerExistsInPeerListings` and not `List` - can they be removed?

api/peering_test.go
```diff
@@ -2,14 +2,155 @@ package api
...
+	})
+
+	t.Run("call List with dc1", func(t *testing.T) {
```
kisunji, line 111: This test can be removed; the user never actually interacts with the datacenter; it gets overwritten with the agent's dc in the handler

api/peering_test.go
```diff
@@ -78,17 +216,19 @@ func TestAPI_Peering_GenerateToken_Read_Initiate_Delete(t *testing.T) {
...
-	require.NoError(t, err4)
-	require.NotEmpty(t, qm2)
+		// basic ok checking
+		require.NoError(t, err4)
+		require.NotEmpty(t, qm2)
 
-	// require that the peering state is not undefined
-	require.Equal(t, INITIAL, respr.State)
+		// require that the peering state is not undefined
+		require.Equal(t, INITIAL, respr.State)
```
kisunji, line 227: In the retry closure we should use `r` not `t` (or else go test will be very angry at you on failure)

kisunji (changes requested):
I had a second thought about reusing `PeeringRequest` - we should rename it to `PeeringDeleteRequest` and avoid sharing Request types across methods.

Since this is a public-facing API we want to future-proof the signatures as much as possible to avoid breaking changes.

If multiple methods use the same request type and we need to add a new field for only one method, it would end up being shared with all methods relying on that request type regardless of whether they use the new field or not.

api/peering_test.go
```diff
@@ -2,14 +2,155 @@ package api
...
+
+	return false
+}
+
+func TestAPI_Peering_Read(t *testing.T) {
+	t.Parallel()
+	c, s := makeClientWithCA(t)
+	defer s.Stop()
+	s.WaitForSerfCheck(t)
+	ctx := context.Background()
```
kisunji, line 40: Just noticed this (and in other tests) should have a timeout otherwise a bad network connection might cause this test to hang for a long time

api/peering.go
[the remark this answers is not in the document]
rboyer: I second @kisunji 's comment about renaming the constants.

api/peering_test.go
FFMMM: ah that's a great catch, thank you!

api/peering_test.go
eculver: Yeah, I agree with @FFMMM. While it is a little annoying to have to fix the tests when adding a field, but it's kind of expected and at least it's in one spot. Plus your point about removing fields is valid too.

eculver (approved):
Looks great! Thanks for making all those changes that weren't necessarily linked to your original PR(s).

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/663698.

PR #12936, opened 2022-05-04
Update CHANGELOG.md

kisunji:
Update with missed changelog entry


jkirschner-hashicorp (commented):
The changelog entry itself looks great. I leave it to someone on the engineering side to decide which branch(es) this needs to be merged to.

eculver:
Thanks for adding to the CHANGELOG. Do you think we should also add the entry so that it generates correctly?

eculver:
> Thanks for adding to the CHANGELOG. Do you think we should also add the entry so that it generates correctly?

Based on our offline discussion, this is a NO. That would result in the entry showing up in 1.12.1. and require us to manually remove the entry when we do the next release which is a pain.

Issue #10603, opened 2021-07-12
New WARN in 1.10.0 caused by shuffling the servers in the gRPC ClientConn pool


shellfu:
**Note from @lkysow:** I'm moving this to hashicorp/consul because the discuss post shows a user on EC2 also saw this error.

### Overview of the Issue
New 1.10.0 on New K8s Cluster results in `[WARN]  agent: grpc: addrConn.createTransport failed to connect to {10.200.65.16:8300 0 consul-server-2.primary <nil>}. Err :connection error: desc = "transport: Error while dialing dial tcp 10.200.65.16:8300: operation was canceled". Reconnecting...`

These WARNS appear in both the server and clients.

### Reproduction Steps
1. When running helm install with the following `values.yml`:
```
client:
  enabled: true
  grpc: true
connectInject:
  aclBindingRuleSelector: serviceaccount.name!=default
  default: false
  enabled: true
  metrics:
    defaultEnableMerging: true
    defaultEnabled: true
    defaultMergedMetricsPort: 20100
    defaultPrometheusScrapePath: /metrics
    defaultPrometheusScrapePort: 20200
  transparentProxy:
    defaultEnabled: true
    defaultOverwriteProbes: true
controller:
  enabled: true
dns:
  enabled: true
global:
  acls:
    createReplicationToken: true
    manageSystemACLs: true
  datacenter: primary
  enabled: true
  federation:
    createFederationSecret: true
    enabled: true
  gossipEncryption:
    secretKey: key
    secretName: consul-gossip-encryption-key
  image: hashicorp/consul:1.10.0
  imageEnvoy: envoyproxy/envoy-alpine:v1.18.3
  imageK8S: hashicorp/consul-k8s:0.26.0
  logJSON: true
  metrics:
    agentMetricsRetentionTime: 1m
    enableAgentMetrics: false
    enableGatewayMetrics: true
    enabled: true
  name: consul
  tls:
    enableAutoEncrypt: true
    enabled: true
    httpsOnly: true
    serverAdditionalDNSSANs:
    - '*.consul'
    - '*.svc.cluster.local'
    - '*.my.customdomain.com'
    verify: false
meshGateway:
  enabled: true
  service:
    enabled: true
    port: 443
    type: LoadBalancer
  wanAddress:
    port: 443
    source: Service
server:
  bootstrapExpect: 5
  connect: true
  disruptionBudget:
    enabled: true
    maxUnavailable: 2
  enabled: true
[...]

PR #12940, opened 2022-05-04
Use native grpc round robin balancer

kisunji:
Possibly fixes https://github.com/hashicorp/consul/issues/10603

### Description
Users are reporting frequently logged WARNs which are being output by gRPC.
They are caused by our gRPC resolver attempting to shuffle server addresses by:
1. Randomly shuffle server addresses every T interval
2. Pass the first address to update the gRPC client connection which kills any existing connections (unless the first address happens to be equal to the existing)
3. Pass all the server addresses to update the connection to provide fallback addresses.

The WARN log seems to be caused by step 3 occurring as the connection is in a transitionary state from applying step 2.

Testing locally, it seems that removing the extra call to `UpdateState` also eliminates the WARN logs.

This PR attempts to replace our custom client-side shuffling of server addresses with native gRPC-go's round robin load balancer.

### Impact
There are 2 changed behaviors from this PR:

1. Clients will no longer have sticky connections to Servers
   **Before**: a client making a gRPC call always routes to the same server until the shuffle happens
   **Now**: a client making a gRPC call will always route to a new server
2. Existing connections (streams) will no longer be terminated 
   **Before**: a server address shuffle usually leads to existing connections being terminated (see step 2 above)
   **Now**: there is no server address shuffle

### DIFF
--- a/agent/consul/subscribe_backend_test.go
+++ b/agent/consul/subscribe_backend_test.go
@@ -20,6 +20,7 @@ import (
 	"github.com/hashicorp/consul/agent/structs"
 	"github.com/hashicorp/consul/proto/pbservice"
 	"github.com/hashicorp/consul/proto/pbsubscribe"
+	"github.com/hashicorp/consul/sdk/testutil/retry"
 	"github.com/hashicorp/consul/testrpc"
 )
 
@@ -204,11 +205,10 @@ func TestSubscribeBackend_IntegrationWithServer_TLSReload(t *testing.T) {
 	server.tlsConfigurator.Update(newConf)
 
 	// Try the subscribe call again
-	retryFailedConn(t, conn)
-
-	streamClient = pbsubscribe.NewStateChangeSubscriptionClient(conn)
-	_, err = streamClient.Subscribe(ctx, req)
-	require.NoError(t, err)
+	retry.Run(t, func(r *retry.R) {
+		_, err = streamClient.Subscribe(ctx, req)
+		require.NoError(r, err)
+	})
 }
 
 func clientConfigVerifyOutgoing(config *Config) {
--- a/agent/grpc/private/client.go
+++ b/agent/grpc/private/client.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/balancer/roundrobin"
 	"google.golang.org/grpc/keepalive"
 
 	"github.com/hashicorp/consul/agent/metadata"
@@ -130,8 +131,8 @@ func (c *ClientConnPool) dial(datacenter string, serverType string) (*grpc.Clien
 		grpc.WithContextDialer(c.dialer),
 		grpc.WithDisableRetry(),
 		grpc.WithStatsHandler(newStatsHandler(defaultMetrics())),
-		// nolint:staticcheck // there is no other supported alternative to WithBalancerName
-		grpc.WithBalancerName("pick_first"),
+		// nolint:staticcheck
+		grpc.WithBalancerName(roundrobin.Name),
 		// Keep alive parameters are based on the same default ones we used for
 		// Yamux. These are somewhat arbitrary but we did observe in scale testing
 		// that the gRPC defaults (servers send keepalives only every 2 hours,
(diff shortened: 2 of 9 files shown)


agent/grpc/private/client.go
```diff
@@ -130,8 +131,8 @@ func (c *ClientConnPool) dial(datacenter string, serverType string) (*grpc.Clien
 		grpc.WithContextDialer(c.dialer),
 		grpc.WithDisableRetry(),
 		grpc.WithStatsHandler(newStatsHandler(defaultMetrics())),
-		// nolint:staticcheck // there is no other supported alternative to WithBalancerName
-		grpc.WithBalancerName("pick_first"),
```
kisunji, line 134: This is no longer true https://pkg.go.dev/google.golang.org/grpc/balancer#section-directories

agent/grpc/private/client_test.go
```diff
@@ -373,21 +373,11 @@ func TestClientConnPool_IntegrationWithGRPCResolver_Rebalance(t *testing.T) {
 	first, err := client.Something(ctx, &testservice.Req{})
 	require.NoError(t, err)
 
-	t.Run("rebalance a different DC, does nothing", func(t *testing.T) {
```
kisunji, line 376: No longer a relevant test since calls are round-robin

agent/grpc/private/resolver/resolver.go
```diff
@@ -270,19 +243,6 @@ func (r *serverResolver) updateAddrs(addrs []resolver.Address) {
...
-	// first address as long as it is still in the list, making it impossible to
-	// rebalance until that address is removed.
-	var firstAddr []resolver.Address
-	if len(addrs) > 0 {
-		firstAddr = []resolver.Address{addrs[0]}
-	}
-	r.clientConn.UpdateState(resolver.State{Addresses: firstAddr})
-
-	// Call UpdateState again with the entire list of addrs in case we need them
-	// for failover.
```
kisunji, line 285: If we do want to preserve the old behavior of killing streams (almost) every time server addresses are shuffled, we could add some logic here.

kisunji:
Looks like this broke some tests. Will look into them when I'm back from vacation

kyhavlov (commented):
Does this mean that every client will maintain an open connection to every server (and it's only the target server being shuffled)? Or that when a new request happens we're always opening a new connection to a server? In the former case that might cause servers to hit open connection/file descriptor limits more often (5x the connections with 5 servers).

agent/consul/subscribe_backend_test.go
```diff
@@ -204,11 +205,10 @@ func TestSubscribeBackend_IntegrationWithServer_TLSReload(t *testing.T) {
...
 	// Try the subscribe call again
-	retryFailedConn(t, conn)
-
-	streamClient = pbsubscribe.NewStateChangeSubscriptionClient(conn)
-	_, err = streamClient.Subscribe(ctx, req)
-	require.NoError(t, err)
+	retry.Run(t, func(r *retry.R) {
+		_, err = streamClient.Subscribe(ctx, req)
+		require.NoError(r, err)
+	})
```
kisunji, line 211: There's a slight timing issue (<1s) that flakes the second try. I'm unsure why the previous code did not run into the same issue but I didn't consider this to be a significant regression.

mkeeler:
@kisunji I am not so sure this is desirable:

> Existing connections (streams) will no longer be terminated
Before: a server address shuffle usually leads to existing connections being terminated (see step 2 above)
Now: there is no server address shuffle

We use the shuffling as a way of ensuring load gets properly distributed across all servers. For example if we have 3 servers and each one has 10k streams open. The load is too high so more servers (possibly read replicas) are added. If we don't shuffle then those existing 30k streams will remain pinned to the original 3 servers.

If we do get rid of the shuffle in the grpc resolver mechanisms we probably need to replace it with some more graceful mechanism for stream reestablishment to happen (or some other means of distributed load balancing)

mkeeler:
I haven't delved too deeply into the grpc client so its possible that the behavior of keeping a connection open to all addresses is already happening. 

The most important thing is likely not having a means to force stream reestablishment to another server.

kisunji:
The `pick_first` policy stops traversing the list on a successful connection so this PR would definitely cause more overhead.

mkeeler (commented):
Leaving this comment as a means of marking myself as having reviewed the PR. 

Overall I have a few concerns but feel free to correct me if my knowledge of how the current gRPC resolver / connection pooling works today isn't accurate.

PR #12947, opened 2022-05-05
Add Linked Endpoints to Terminating Gateway Config Entry

DanStough:
### Description
As part of the work to facilitate transparent proxying through Terminating Gateways, this PR adds "Endpoints" to the Terminating Gateway Config Entry, which explicitly enables external endpoints for the mesh without adding services to the Consul catalog.

Test cases included. You can always `consul config write` a gateway with the new `endpoints` field.

### Links
PRD
RFC

### PR Checklist

* [x] updated test coverage
* [ ] !external facing docs updated!
* [x] not a security concern

### DIFF
--- a/agent/structs/config_entry_gateways.go
+++ b/agent/structs/config_entry_gateways.go
@@ -2,6 +2,7 @@ package structs
 
 import (
 	"fmt"
+	"net"
 	"sort"
 	"strings"
 
@@ -435,9 +436,10 @@ func (s *IngressService) ToServiceName() ServiceName {
 // TerminatingGatewayConfigEntry manages the configuration for a terminating service
 // with the given name.
 type TerminatingGatewayConfigEntry struct {
-	Kind     string
-	Name     string
-	Services []LinkedService
+	Kind      string
+	Name      string
+	Services  []LinkedService
+	Endpoints []LinkedEndpoint
 
 	Meta               map[string]string `json:",omitempty"`
 	acl.EnterpriseMeta `hcl:",squash" mapstructure:",squash"`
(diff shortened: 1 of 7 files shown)


agent/structs/config_entry_gateways.go
```diff
@@ -526,10 +561,10 @@ func (e *TerminatingGatewayConfigEntry) Validate() error {
 		}
 
 		// Check for duplicates within the entry
-		if ok := seen[cid]; ok {
+		if ok := serviceSet[cid]; ok {
```
dhiaayachi, line 564: I think we should do the same validation for duplicate endpoints, because we are doing a separate set of intentions for endpoints I don't think that a service and an endpoint with the same name is an issue though

agent/structs/config_entry_gateways.go
```diff
@@ -539,6 +574,56 @@ func (e *TerminatingGatewayConfigEntry) Validate() error {
...
+	var valid bool
+
+	ip := net.ParseIP(address)
+	valid = ip != nil
+
+	_, _, err := net.ParseCIDR(address)
+	valid = valid || err == nil
+
+	// Hostname
+	re := regexp.MustCompile(DNSName)
```
dhiaayachi, line 621: Maybe for consistency we should use something like [validateHost](https://github.com/hashicorp/consul/blob/c6ff4ba7d861a9ea1ec875141e2200ad0faa488e/agent/structs/config_entry_gateways.go#L343-L343)

dhiaayachi (changes requested):
Thank you for picking this up @DanStough. 
I think the failing test is related to the regex, see my comment about reusing an already existing validation.
I also think we are missing a test that would verify the state store write/read for this.

agent/structs/config_entry_gateways.go
dhiaayachi: yes, envoy support wildcard as a subdomain only. so *.domain.com would work but api.*main.com won't, see [here](https://github.com/envoyproxy/envoy/blob/ad667a2c71c2081236368dad450abb879afc4fb8/api/envoy/config/listener/v3/listener_components.proto#L156)

agent/structs/config_entry_gateways.go
DanStough: I agree with your point but I create a new map for endpoints to check for duplicates ([line 578](https://github.com/hashicorp/consul/pull/12947/files#diff-26c007c7c0d5e30fabd316255cc13ac3ffcfac5295ae8a66bcca29c197054664R578)). Could you be more specific?

agent/structs/config_entry_gateways.go
DanStough: I changed it to `validateHost`, but had to get rid of a lot of test cases for invalid IPs and CIDRs because it lets pretty much anything slide.

DanStough:
We're going in a different implementation direction on this feature. Stay tuned.

Issue #1779, opened 2016-03-01
Feature Request: Organize services in K/V-like tree


ChrisAubuchon:
Some examples:
- `dev/webapp`
- `qa/webapp`
- `prod/webapp`
### DNS interface

The Current DNS interface in Consul resolves services like this:
- `tag.serviceName.service.domain`
- `_serviceName._tag.service.domain`

Using a K/V like tree structure, DNS resolution would look like this (using `group1/group2/.../groupN` as the service):
- `tag.service.groupN...group2.group1.service.domain`
- `_service._tag.groupN...group2.group1.service.domain`

Using `dev/eu/webapp` with tags `[ "http", "https" ]` as an example:

```
http.webapp.eu.dev.service.consul
https.webapp.eu.dev.service.consul
_webapp._http.eu.dev.service.consul
_webapp._https.eu.dev.service.consul
```
### Issues

Organizing services in a tree structure can introduce ambiguity in DNS resolution:

```
service eu/dev/webapp
service eu/dev with tag "webapp"
```

Both services are valid when trying to resolve `webapp.dev.eu.service.consul`. The solution for this would be to treat `webapp.dev.eu` as `eu/dev/webapp` first and `eu/dev with tag webapp` second. Essentially trying to match the longest service name.

Allowing '.'s in a tag could require more searches since, in the case of `webapp.dev.eu.service.consul` there could be no tag, `webapp`, and `webapp.dev` as a tag. Tags with '.'s don't really work in the RFC 2782 case so maybe tags with '.'s just don't resolve.

stevendborrelli:
The idea behind this is to support kubernetes, which currently uses skydns as an add-on.

BrianHicks:
Any update about this @slackpad? What are you thinking about here?

slackpad:
We are thinking through a more general design for multiple tenants like dev, qa, prod so that may go deeper than just extra tag support. Also, trying to see how these types of lookups would work given our indexing in the state store so it can be efficient. Haven't had a chance yet to look at this in detail.

slackpad:
Hi all - one thing I wanted to understand better is how this maps into Kubernetes concepts. Doing a quick look over there (I'm not experienced with k8s) there's the concept of a namespace and labels (which are flat). How does this map into the hierarchy proposed here? I can see one level of hierarchy where services are in a namespace, but with labels it seems like we are trying to do an equality selector with the additional groups. Where does k8s support hierarchy other than the namespace?

[...]

PR #12957, opened 2022-05-05
peering: add store.PeeringsForService implementation

eculver:
This is the OSS side of [Enterprise #1779](https://github.com/hashicorp/consul-enterprise/pull/1779).

### Description

In order to configure connect proxies for peered services, we need a way to list all peers for a given service so that we can then fetch all the trust bundles for those associated peers. This is all the state store changes to support the integration.

In a subsequent PR will be the RPC endpoint that consumes this store.PeeringsForService method and pull it all together. Once these two are done, we should then be able to update the leader routine to watch and configure connect proxies for mesh fed v2.

I added coverage for the `store.PeeringsForService` method as well as package tests for some of the helper code.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


rboyer (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/664594.

PR #12961, opened 2022-05-06
Ma/fix enterprise version string oss

markan:
### Description
Changes to how the version string was handled created a small regression that was noticed after the release of Consul 1.12.0 Enterprise. This is the OSS companion of https://github.com/hashicorp/consul-enterprise/pull/1800 because most of the code paths are shared; however because we don't have the ent metadata flag this PR should not change any user visible output.

Many tools use the Config:Version field reported by the agent/self resource to determine whether Consul is an enterprise or OSS instance, expect something like 1.12.0+ent for enterprise and simply 1.12.0 for OSS. This was accidentally broken during the runup to 1.12.x

In previous versions of Consul Enterprise we would expect the agent/self endpoint to return something like:
```
 curl 127.0.0.1:8500/v1/agent/self | jq '.Config.Version'
"1.11.1+ent"
```

In the released version of 1.12.0 we got.
```
 curl 127.0.0.1:8500/v1/agent/self | jq '.Config.Version'
"1.12.0"
````

Updated the metrics endpoint as well, but had trouble getting a stable metrics test working.

Expect:
```
curl 127.0.0.1:8500/v1/agent/metrics  | jq '.Gauges[] | select(.Name == "consul.version")' 

{
  "Name": "consul.version",
  "Value": 1,
  "Labels": {
    "pre_release": "dev",
    "version": "1.12.0"
  }
}
```
Added test for the ent flag in version in in the self endpoint for both OSS and enterprise.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/.changelog/12961.txt
+++ b/.changelog/12961.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+api: agent/self now returns version with +ent suffix for Enterprise Consul
+```
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -720,7 +720,7 @@ func (a *Agent) Start(ctx context.Context) error {
 
 	// consul version metric with labels
 	metrics.SetGaugeWithLabels([]string{"version"}, 1, []metrics.Label{
-		{Name: "version", Value: a.config.Version},
+		{Name: "version", Value: a.config.VersionWithMetadata()},
 		{Name: "pre_release", Value: a.config.VersionPrerelease},
 	})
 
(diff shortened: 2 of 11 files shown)


boxofrad (approved):

hc-github-team-consul-core:
:cherries::white_check_mark: Cherry pick of commit c479cdce01e4c5178674564d9be42b3c0f17501b onto `release/1.12.x` succeeded!

PR #12964, opened 2022-05-06
expose meta tags for peering

FFMMM:
### Description
Meta tags are useful for the API usage for CRD controllers.

* updated unit tests

### Links
Asana task ( [ref](https://app.asana.com/0/1200427797338423/1202198915682709/f) )


api/peering.go
```diff
@@ -68,6 +70,8 @@ type PeeringGenerateTokenRequest struct {
 	Partition  string `json:"Partition,omitempty"`
 	Datacenter string
 	Token      string
+	// Meta is a mapping of some string value to any other string value
+	Meta map[string]string
```
rboyer, line 74:
```suggestion
	Meta map[string]string `json:",omitempty"`
```

api/peering.go
```diff
@@ -24,6 +24,8 @@ type Peering struct {
 	Name string
 	// Partition is the local partition connecting to the peer.
 	Partition string `json:"Partition,omitempty"`
+	// Meta is a mapping of some string value to any other string value
+	Meta map[string]string
```
rboyer, line 28:
```suggestion
	Meta map[string]string `json:",omitempty"`
```

api/peering.go
```diff
@@ -83,6 +87,8 @@ type PeeringInitiateRequest struct {
 	PeeringToken string
 	Datacenter   string
 	Token        string
+	// Meta is a mapping of some string value to any other string value
+	Meta map[string]string
```
rboyer, line 91:
```suggestion
	Meta map[string]string `json:",omitempty"`
```

agent/rpc/peering/service.go
```diff
@@ -158,9 +158,9 @@ func (s *Service) GenerateToken(
 	writeReq := pbpeering.PeeringWriteRequest{
 		Peering: &pbpeering.Peering{
 			Name: req.PeerName,
-
 			// TODO(peering): Normalize from ACL token once this endpoint is guarded by ACLs.
 			Partition: req.PartitionOrDefault(),
+			Meta:      req.Meta,
```
rboyer, line 163: Before accepting this metadata here and in the other endpoint we should apply a similar pre-flight validation like we do for config entry meta. You could slightly rename this method, export it and then repurpose it here:
https://github.com/hashicorp/consul/blob/main/agent/structs/config_entry.go#L1118

api/peering.go
FFMMM: Hm, could I ask why add the `omitempty`? would this be ent specific? I think in a previous PR, I understood from Chris that we only add `\`json:",omitempty"\`` for ENT things like namespaces or partitions

agent/rpc/peering/service.go
FFMMM: Thanks for the call out; let me work that in ⌨️

api/peering.go
rboyer: You add it for anything that may be empty that you don't want to send via json to the other side as empty.

agent/rpc/peering/service_test.go
```diff
@@ -167,11 +168,13 @@ func TestPeeringService_Initiate(t *testing.T) {
 			req: &pbpeering.InitiateRequest{
 				PeerName:     "peer1-usw1",
 				PeeringToken: validTokenB64,
+				Meta:         map[string]string{"foo": "bar"},
 			},
 			expectResp: &pbpeering.InitiateResponse{},
 			expectPeering: peering.TestPeering(
 				"peer1-usw1",
 				pbpeering.PeeringState_INITIAL,
+				map[string]string{"foo": "bar"},
```
eculver, line 177: I know we have a TODO for the `PeeringWrite` endpoint to potentially be removed: https://github.com/hashicorp/consul/blob/main/proto/pbpeering/peering.proto#L21-L23 But I was wondering if we should add a similar map there to write and then read them back?

api/peering_test.go
```diff
@@ -205,6 +208,7 @@ func TestAPI_Peering_GenerateToken_Read_Initiate_Delete(t *testing.T) {
 
 		// require that the peering state is not undefined
 		require.Equal(r, PeeringStateInitial, respr.State)
+		require.Equal(r, map[string]string{"foo": "bar"}, resp2.Meta)
```
eculver, line 211: I think this should be looking at `respr.Meta` and not `resp2.Meta`.

```suggestion
		require.Equal(r, map[string]string{"foo": "bar"}, respr.Meta)
```

eculver (commented):
What do you think about also adding the meta fields to the HTTP endpoint tests in `agent/peering_endpoint_test.go`?

agent/rpc/peering/service_test.go
FFMMM: The way I actually parsed this feedback is that "we should add tests for Peering_Read in the `peering_endpoint` to make sure we are actually storing these in the state store". Please let me know if I got that wrong 🙈 

I added a test for that in the commit [TODO].

Also, we make sure we get the meta data in the API layer as well. So between these two places, I think we can prove that meta data tags get written as expected.

api/peering_test.go
FFMMM: ah yes great catch!

agent/peering_endpoint.go
```diff
@@ -128,6 +133,10 @@ func (s *HTTPHandlers) PeeringInitiate(resp http.ResponseWriter, req *http.Reque
 		return nil, HTTPError{StatusCode: http.StatusBadRequest, Reason: "PeeringToken is required in the payload when initiating a peering."}
 	}
 
+	if err := structs.ValidateMetaTags(args.Meta); err != nil {
```
rboyer, line 136: This should actually happen in `agent/rpc/peering/service.go`, as that runs on the (trusted) servers instead of the (less trusted) clients (where the json api happens). Generally you want correctness validation to happen on the servers so you prevent bad data from being recorded.

FFMMM: that's a great call out, thank you!

I recall a school of thought that encouraged both:

* do client side validation
* do the same validation on the server side too

what do you think about doing that here as well? Would that make sense form a defense in a depth perspective or just redundant in this case?

FFMMM: Chatted offline; we probably don't want to have this validation on the client side too as it would make upgrades tricky. Moved to `agent/rpc/peering/service.go`

agent/rpc/peering/service.go
```diff
@@ -132,6 +132,10 @@ func (s *Service) GenerateToken(
 		return nil, fmt.Errorf("%s is not a valid peer name: %w", req.PeerName, err)
 	}
 
+	if err := structs.ValidateMetaTags(req.Meta); err != nil {
+		return nil, fmt.Errorf("meta tags failed validation: %v", err)
```
rboyer, line 136:
```suggestion
		return nil, fmt.Errorf("meta tags failed validation: %w", err)
```

agent/rpc/peering/service.go
```diff
@@ -214,6 +218,10 @@ func (s *Service) Initiate(
 		return nil, err
 	}
 
+	if err := structs.ValidateMetaTags(req.Meta); err != nil {
+		return nil, fmt.Errorf("meta tags failed validation: %v", err)
```
rboyer, line 222:
```suggestion
		return nil, fmt.Errorf("meta tags failed validation: %w", err)
```

rboyer (approved):
LGTM with a suggestion to use `%w` instead of `%v` for errors passed to `fmt.Errorf`

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/667735.

PR #12986, opened 2022-05-09
add err msg on PeeringRead not found

FFMMM:
### Description
Came across this as a convenience when working with nitya; would be nice to have!

* updated tests

### DIFF
--- a/agent/peering_endpoint.go
+++ b/agent/peering_endpoint.go
@@ -48,7 +48,7 @@ func (s *HTTPHandlers) peeringRead(resp http.ResponseWriter, req *http.Request, 
 		return nil, err
 	}
 	if result.Peering == nil {
-		return nil, HTTPError{StatusCode: http.StatusNotFound, Reason: fmt.Sprintf("peering with name \"%s\" does not exist", name)}
+		return nil, HTTPError{StatusCode: http.StatusNotFound, Reason: fmt.Sprintf("Peering not found for %q", name)}
 	}
 
 	// TODO(peering): replace with API types
--- a/agent/peering_endpoint.go
+++ b/agent/peering_endpoint.go
@@ -48,7 +48,7 @@ func (s *HTTPHandlers) peeringRead(resp http.ResponseWriter, req *http.Request, 
 		return nil, err
 	}
 	if result.Peering == nil {
-		return nil, HTTPError{StatusCode: http.StatusNotFound}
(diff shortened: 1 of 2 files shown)


DanStough (approved):
lgtm!

eculver (approved):
Yeah, this does seem handy. I do always wonder about information leakage when it comes to 404s so we might want to get another set of eyes on it to make sure we aren't sharing more info than we should be in these cases. I wonder if @rboyer @freddygv @banks @mkeeler would have thoughts?

agent/peering_endpoint.go
```diff
@@ -48,7 +48,7 @@ func (s *HTTPHandlers) peeringRead(resp http.ResponseWriter, req *http.Request,
 		return nil, err
 	}
 	if result.Peering == nil {
-		return nil, HTTPError{StatusCode: http.StatusNotFound}
+		return nil, HTTPError{StatusCode: http.StatusNotFound, Reason: fmt.Sprintf("peering with name \"%s\" does not exist", name)}
```
freddygv, line 51: I would prefer if we used the same error convention that we do for partitions/namespaces:
```suggestion
		return nil, HTTPError{StatusCode: http.StatusNotFound, Reason: fmt.Sprintf("Peering not found for %q", name)}
```

mkeeler:
FWIW, this PR does not leak any new information. The name output in the error reason was user supplied input and they could have inferred as much from just a 404. As for 404s, once we have ACLs for looking up peering information we should:

1. Resolve the token.
2. Authorize the request and return a 403 if needed
3. Lookup the peer
4. Return info or return a 404.

That way if you wouldn't be authorized to view the peer we 403 early and don't let you know whether it exists or not.

mkeeler (approved):

agent/peering_endpoint_test.go
```diff
@@ -289,6 +289,7 @@ func TestHTTP_Peering_Read(t *testing.T) {
 		resp := httptest.NewRecorder()
 		a.srv.h.ServeHTTP(resp, req)
 		require.Equal(t, http.StatusNotFound, resp.Code)
+		require.Equal(t, "peering with name \"baz\" does not exist", resp.Body.String())
```
mkeeler, line 292: I think you need to update this to match the change to the message.

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/668069.

PR #13001, opened 2022-05-10
Add support for merge-central-config query param

riddhi89:
### Description
Adds a new query param `merge-central-config` for use with the below endpoints:
1. /catalog/service/:service
2. /catalog/connect/:service
3. /health/service/:service
4. /health/connect/:service

If set on the request, the response will include a fully resolved service definition which is merged with the proxy-defaults/global and service-defaults/:service config entries (on-demand style). This is useful to view the full service definition for a mesh service (connect-proxy kind or gateway kind) which might not be merged before being written into the catalog (example: in case of services in the agentless model). 

**Sample demo**: https://gist.github.com/riddhi89/8a7f2e2ef4956decaf72bfff8a49acfe


agent/consul/config_endpoint.go
```diff
@@ -513,11 +510,12 @@ func (c *ConfigEntry) ResolveServiceConfig(args *structs.ServiceConfigRequest, r
 				ranOnce = true
 			}
```
riddhi89, line 512: @mkeeler I did not move this entire piece of code to be re-usable and instead picked out relevant calls and moved it into `merge_service_config.go/mergeNodeServiceWithCentralConfig`.
The main reason being there is the optimization here for spurious wakeup suppression. 

Do we want to add this optimization in some manner on looping through all service resolutions?

boxofrad (commented):
This is looking great, nice work 👏🏻 (happy to approve once you mark it as ready to review)

agent/consul/catalog_endpoint_test.go
```diff
@@ -1522,6 +1522,7 @@ func TestCatalog_ListServices_NodeMetaFilter(t *testing.T) {
 	}
 }
 
+// RIDDHI example of test blocking call
```
mkeeler, line 1525:
```suggestion
```

Just a reminder to back this out before the final merge.

agent/consul/config_endpoint.go
mkeeler: Would it be terribly difficult to bring over the spurious wakeup suppression. While this should just be the UI performing these queries we have the potential to be constantly sending back this response if we don't have the suppression.

agent/rpcclient/health/health.go
```diff
@@ -38,7 +38,8 @@ func (c *Client) ServiceNodes(
 	ctx context.Context,
 	req structs.ServiceSpecificRequest,
 ) (structs.IndexedCheckServiceNodes, cache.ResultMeta, error) {
-	if c.useStreaming(req) && (req.QueryOptions.UseCache || req.QueryOptions.MinQueryIndex > 0) {
+	// Note: if MergeCentralConfig is requested, default to using the RPC backend for now.
```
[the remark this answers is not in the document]
mkeeler: That is the reasoning. +1 to noting that reason here.

mkeeler (commented):
Looks great. There are 2 changes I think we should make.

1. Move the MergeCentralConfig from the QueryOptions to ServiceSpecificRequest (this likely means a lot of the query compat code may be able to be removed.

2. We should carry over the spurious wakeup suppression.

agent/consul/merge_service_config.go
```diff
@@ -0,0 +1,408 @@
...
+	"github.com/hashicorp/consul/agent/structs"
+	"github.com/hashicorp/go-hclog"
+	memdb "github.com/hashicorp/go-memdb"
+	"github.com/imdario/mergo"
+	"github.com/mitchellh/copystructure"
+)
+
+// mergeNodeServiceWithCentralConfig merges a service instance (NodeService) with the
+// proxy-defaults/global and service-defaults/:service config entries.
+// This a common helper is used by the blocking query function of differnet RPC endpoints
```
rboyer, line 17:
```suggestion
// This a common helper is used by the blocking query function of different RPC endpoints
```

agent/structs/structs.go
```diff
@@ -311,6 +311,12 @@ type QueryOptions struct {
 	// QueryMeta.Index, the response can be left empty and QueryMeta.NotModified
 	// will be set to true to indicate the result of the query has not changed.
 	AllowNotModifiedResponse bool
+
+	// MergeCentralConfig when set to true returns a service definition merged with
+	// the proxy-defaults/global and service-defaults/:service config entries.
+	// This can be used to ensure a full service definition is returned in the response
+	// especially when the service might not be written into the catalog that way.
+	MergeCentralConfig bool
```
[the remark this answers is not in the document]
riddhi89: Done.

agent/rpcclient/health/health.go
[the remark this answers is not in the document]
riddhi89: Done

agent/consul/catalog_endpoint.go
```diff
@@ -672,10 +672,34 @@ func (c *Catalog) ServiceNodes(args *structs.ServiceSpecificRequest, reply *stru
...
 
-			reply.Index, reply.ServiceNodes = index, services
+			mergedServices := services
+
+			if args.MergeCentralConfig {
+				var mergedServiceNodes structs.ServiceNodes
+				for _, sn := range services {
+					mergedsn := sn
+					ns := sn.ToNodeService()
+					if ns.IsSidecarProxy() || ns.IsGateway() {
```
riddhi89, line 682: Calling out that the merging would happen only for mesh services to mirror when the agent registration performs merging [here](https://github.com/hashicorp/consul/blob/main/agent/agent.go#L2072-L2073).

mkeeler (approved):
Everything looks great to me.

PR #13012, opened 2022-05-10
Watch the singular service resolver instead of the list + filtering to 1

mkeeler:
### Description
I noticed that we were using the listing version of the cache type here, relying on the RPC to filter to a singular config entry. The consuming side of the cache watch also was explicitly validating that exactly 1 entry was returned.

While functionally correct, this would incur extra overhead of iterating and filtering out all other service resolvers. This PR switches that watch to a singular config entry watch.

* If envoy integration tests pass then things are fine. This should not have any functional difference.

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/.changelog/13012.txt
+++ b/.changelog/13012.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+proxycfg: Fixed a minor bug that would cause configuring a terminating gateway to watch too many service resolvers and waste resources doing filtering.
+```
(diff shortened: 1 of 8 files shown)


freddygv (approved):
Thanks! Would you be able to also rename the `ConfigEntriesName` to be something like `ConfigEntriesList`? That plural in the middle is not obvious at all

mkeeler:
I renamed things from `ConfigEntries` to `ConfigEntryList`. The registered name is still `ConfigEntryListName` (as all of them are suffixed with `Name`)

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/676909.

PR #12098, opened 2022-01-14
Support for connect native services in topology view.

apollo13:
This is a draft implementation for #12096. Misses tests and probably cleanups of the JS code :)

PR #13023, opened 2022-05-11
Fix topology view when displaying mixed connect-native/normal services.

apollo13:
### Description

If a service is registered twice, once with connect-native and once
without, the topology views would prune the existing intentions. This
change brings the code more in line with the transparent proxy behavior.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/.changelog/13023.txt
+++ b/.changelog/13023.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: the topology view now properly displays services with mixed connect and non-connect instances.
+```
--- a/agent/consul/state/catalog.go
+++ b/agent/consul/state/catalog.go
@@ -4535,13 +4535,17 @@ func (s *Store) ServiceTopology(
 		maxIdx = idx
 	}
 
-	// Store downstreams with at least one instance in transparent proxy mode.
+	// Store downstreams with at least one instance in transparent proxy or connect native mode.
 	// This is to avoid returning downstreams from intentions when none of the downstreams are transparent proxies.
-	tproxyMap := make(map[structs.ServiceName]struct{})
+	proxyMap := make(map[structs.ServiceName]struct{})
 	for _, downstream := range unfilteredDownstreams {
 		if downstream.Service.Proxy.Mode == structs.ProxyModeTransparent {
 			sn := structs.NewServiceName(downstream.Service.Proxy.DestinationServiceName, &downstream.Service.EnterpriseMeta)
-			tproxyMap[sn] = struct{}{}
+			proxyMap[sn] = struct{}{}
+		}
+		if downstream.Service.Connect.Native {
+			sn := downstream.Service.CompoundServiceName()
+			proxyMap[sn] = struct{}{}
 		}
 	}
 
@@ -4551,7 +4555,7 @@ func (s *Store) ServiceTopology(
 		if downstream.Service.Kind == structs.ServiceKindConnectProxy {
 			sn = structs.NewServiceName(downstream.Service.Proxy.DestinationServiceName, &downstream.Service.EnterpriseMeta)
 		}
-		if _, ok := tproxyMap[sn]; !ok && !downstream.Service.Connect.Native && downstreamSources[sn.String()] != structs.TopologySourceRegistration {
+		if _, ok := proxyMap[sn]; !ok && downstreamSources[sn.String()] != structs.TopologySourceRegistration {
 			// If downstream is not a transparent proxy or connect native, remove references
 			delete(downstreamSources, sn.String())
 			delete(downstreamDecisions, sn.String())
@@ -4580,6 +4584,7 @@ func (s *Store) combinedServiceNodesTxn(tx ReadTxn, ws memdb.WatchSet, names []s
 		maxIdx uint64
 		resp   structs.CheckServiceNodes
 	)
+	dedupMap := make(map[string]structs.CheckServiceNode)
 	for _, u := range names {
 		// Collect typical then connect instances
 		idx, csn, err := checkServiceNodesTxn(tx, ws, u.Name, false, &u.EnterpriseMeta, peerName)
@@ -4589,7 +4594,9 @@ func (s *Store) combinedServiceNodesTxn(tx ReadTxn, ws memdb.WatchSet, names []s
 		if idx > maxIdx {
 			maxIdx = idx
 		}
-		resp = append(resp, csn...)
+		for _, item := range csn {
+			dedupMap[item.Node.Node+"/"+item.Service.ID] = item
+		}
 
 		idx, csn, err = checkServiceNodesTxn(tx, ws, u.Name, true, &u.EnterpriseMeta, peerName)
 		if err != nil {
@@ -4598,7 +4605,12 @@ func (s *Store) combinedServiceNodesTxn(tx ReadTxn, ws memdb.WatchSet, names []s
 		if idx > maxIdx {
 			maxIdx = idx
 		}
-		resp = append(resp, csn...)
+		for _, item := range csn {
+			dedupMap[item.Node.Node+"/"+item.Service.ID] = item
+		}
+	}
+	for _, item := range dedupMap {
+		resp = append(resp, item)
 	}
 	return maxIdx, resp, nil
 }
--- a/agent/ui_endpoint_test.go
+++ b/agent/ui_endpoint_test.go
@@ -1687,19 +1687,43 @@ func TestUIServiceTopology(t *testing.T) {
 				SkipNodeUpdate: true,
 				Service: &structs.NodeService{
 					Kind:    structs.ServiceKindTypical,
-					ID:      "cproxy",
+					ID:      "cproxy-https",
 					Service: "cproxy",
 					Port:    1111,
 					Address: "198.18.1.70",
+					Tags:    []string{"https"},
 					Connect: structs.ServiceConnect{Native: true},
 				},
 				Checks: structs.HealthChecks{
 					&structs.HealthCheck{
 						Node:        "cnative",
-						CheckID:     "cnative:cproxy",
+						CheckID:     "cnative:cproxy-https",
 						Name:        "cproxy-liveness",
 						Status:      api.HealthPassing,
-						ServiceID:   "cproxy",
+						ServiceID:   "cproxy-https",
+						ServiceName: "cproxy",
+					},
+				},
+			},
+			"Service cproxy/http on cnative": {
+				Datacenter:     "dc1",
+				Node:           "cnative",
+				SkipNodeUpdate: true,
+				Service: &structs.NodeService{
+					Kind:    structs.ServiceKindTypical,
+					ID:      "cproxy-http",
+					Service: "cproxy",
+					Port:    1112,
+					Address: "198.18.1.70",
+					Tags:    []string{"http"},
+				},
+				Checks: structs.HealthChecks{
+					&structs.HealthCheck{
+						Node:        "cnative",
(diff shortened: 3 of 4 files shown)


agent/ui_endpoint_test.go
```diff
@@ -1838,6 +1862,42 @@ func TestUIServiceTopology(t *testing.T) {
...
+				TransparentProxy: false,
+				Upstreams:        []*ServiceTopologySummary{},
+				Downstreams: []*ServiceTopologySummary{
+					{
+						ServiceSummary: ServiceSummary{
+							Name:           "cproxy",
+							Datacenter:     "dc1",
+							Tags:           []string{"http", "https"},
+							Nodes:          []string{"cnative", "cnative", "cnative"},
+							InstanceCount:  3,
```
apollo13, line 1882: Note to reviewers: I have no idea why we have three instances here and I am not sure this is correct (but not a result of my PR anyways). I think I managed to track it down to https://github.com/hashicorp/consul/blob/1bac1544d094ef356787974cd0440783142e53e1/agent/consul/state/catalog.go#L4124 but I am not exactly sure.

apollo13:
@kisunji This is a follow up to #12098 -- would be great if you could give that a review

agent/consul/state/catalog.go
```diff
@@ -4081,13 +4081,17 @@ func (s *Store) ServiceTopology(
...
-	tproxyMap := make(map[structs.ServiceName]struct{})
+	proxyMap := make(map[structs.ServiceName]struct{})
 	for _, downstream := range unfilteredDownstreams {
 		if downstream.Service.Proxy.Mode == structs.ProxyModeTransparent {
 			sn := structs.NewServiceName(downstream.Service.Proxy.DestinationServiceName, &downstream.Service.EnterpriseMeta)
-			tproxyMap[sn] = struct{}{}
+			proxyMap[sn] = struct{}{}
+		}
+		if downstream.Service.Connect.Native {
+			sn := structs.NewServiceName(downstream.Service.Service, &downstream.Service.EnterpriseMeta)
```
apollo13, line 4093: I am also not sure if `Service.Service` is 100% correct here, but it doesn't seem to fail tests? :D

Amier3:
Hey @apollo13 

Hope you're doing well! Apologies for the delayed review on this, looks like it slipped through the cracks. We'll get this triaged shortly! 😄

apollo13:
@Amier3 Friendly ping :) any chance of a review?

apollo13:
Hi @Amier3, @kisunji. Any chance of a review? I realize that this is a small fix in the grand scheme of things, but nevertheless would improve traefik integration quite a bit.

kisunji:
Hi @apollo13, I don't have the bandwidth to review at this time. If you could describe what the bug is that you are trying to fix (maybe replication steps if it is simple enough) it might help future reviewers understand what behavioral change you are making here.

apollo13:
@Amier3 Any chance to get someone to review this?

ventaubain:
I'm interested by seeing this commit validated to fix #14795

apollo13:
@kisunji I have rebased the code. Any chance you or someone else could spare the time to review this?

agent/consul/state/catalog.go
kisunji: I'd recommend `downstream.Service.CompoundServiceName()` (which effectively does the same thing)

agent/ui_endpoint_test.go
apollo13: So it is correct for `combinedServiceNodesTxn` to return what it returns or should the filtering already happen in it? I guess I am a bit at a loss when trying to understand why `cproxy-https` shows up in both cases -- can you shed some light on it?

agent/consul/state/catalog.go
```diff
@@ -4535,31 +4535,46 @@ func (s *Store) ServiceTopology(
...
+			sn := downstream.Service.CompoundServiceName()
+			proxyMap[sn] = struct{}{}
 		}
 	}
 
 	var downstreams structs.CheckServiceNodes
+	refsToRemove := make(map[structs.ServiceName]struct{})
+	servicesSeen := make(map[string]struct{})
 	for _, downstream := range unfilteredDownstreams {
+		serviceNodeID := fmt.Sprintf("%s/%s", downstream.Node.Node, downstream.Service.ID)
```
apollo13, line 4556: @kisunji Is there a better way to dedupe services than by `Node.Node` & `Service.ID`. Interestingly enough `Node.ID` is empty, is that due to the tests or is that expected to be empty?

agent/ui_endpoint_test.go
kisunji: I'm not sure if that is the correct behavior but before your earlier PR #12098 I think all these services would have been filtered out anyways since they were not TProxy. 

It's probably safest to dedupe outside of `combinedServiceNodesTxn` in the `ServiceTopology` method though.

kisunji: Actually, staring at the code a bit more, I think it makes sense to de-dupe _inside_ `combinedServiceNodesTxn`. `checkServiceNodesTxn` with `connect: false` will always return the a service that overlaps with `checkServiceNodesTxn` with `connect: true` due to the indexing.

This bug was hidden because previously we filtered all non-TProxy services.

agent/consul/state/catalog.go
kisunji: `Node.ID` is optional so checking `Node.Node` & `Service.ID` should be sufficient for uniqueness

kisunji: I responded to a separate comment but could this deduping happen within `combinedServiceNodesTxn`?

The implementation should be simpler in that function:
* Instead of appending `csn` to `resp` directly, add to a `map[{node+serviceid}]structs.CheckServiceNode`
* After finishing the loop over `names`, loop through the map and add to `resp`

agent/consul/state/catalog.go
```diff
@@ -4551,7 +4555,7 @@ func (s *Store) ServiceTopology(
 		if downstream.Service.Kind == structs.ServiceKindConnectProxy {
 			sn = structs.NewServiceName(downstream.Service.Proxy.DestinationServiceName, &downstream.Service.EnterpriseMeta)
 		}
-		if _, ok := tproxyMap[sn]; !ok && !downstream.Service.Connect.Native && downstreamSources[sn.String()] != structs.TopologySourceRegistration {
+		if _, ok := proxyMap[sn]; !ok && downstreamSources[sn.String()] != structs.TopologySourceRegistration {
 			// If downstream is not a transparent proxy or connect native, remove references
 			delete(downstreamSources, sn.String())
```
apollo13, line 4560: Thanks for the review @kisunji, I have pushed the requests changes. One thing I am not 100% confident is this if here. Once we remove from `downStreamSources` the if in the line above probably cannot find `structs.TopologySourceRegistration` anymore (even if it was in there before). So I wonder if we should delay the deletion or can this occurrence not happen?

kisunji: Sorry I don't quite understand your concern here. I don't think anything about this if conditional is being changed:

Before: `If downstream is NOT TProxy && NOT Connect && NOT structs.TopologySourceRegistration`

After: `If downstream is NOT (TProxy OR Connect) && NOT structs.TopologySourceRegistration`

Am I misunderstanding your question?

apollo13: I was mainly questioning if it is already wrong before my changes :) Assuming we have duplicated a `sn` (which afaik is still possible, since the dedup happens on the node & service combo) then after the first `if` this `sn` would have been removed from `downstreamSources` and so it will never trigger the `if` (and therefore the `continue` inside) again. My code does not make this any/much worse (I guess?), but I don't understand the code completely either.

kisunji: Ah, the `sn` (ServiceName) is the logical name for the service (e.g. `cproxy`).

Our problem was duplicated CheckServiceNodes (service instances).

`cproxy` ServiceName had `['cproxy-https', 'cproxy-http', 'cproxy-https']` CheckServiceNodes.

So no worries about duplicated service names here!

.changelog/13023.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+ui: the topology view now properly checks for intentions with mixed connect native / normal services.
```
kisunji, line 2:
```suggestion
ui: the topology view now properly displays services with mixed connect and non-connect instances.
```

kisunji (approved):
Looks good 👍 Please consider my feedback on the changelog and perhaps test locally to ensure things still look as expected.

apollo13:
Thank you for your help! I have squashed the commits to reduce the noise (the later commits where just small fixes anyways) and will test if it works correctly (aside from the tests) over the weekend.

agent/ui_endpoint_test.go
```diff
@@ -2125,6 +2149,42 @@ func TestUIServiceTopology(t *testing.T) {
...
+			want: &ServiceTopology{
+				Protocol:         "http",
+				TransparentProxy: false,
+				Upstreams:        []*ServiceTopologySummary{},
+				Downstreams: []*ServiceTopologySummary{
+					{
+						ServiceSummary: ServiceSummary{
+							Name:           "cproxy",
+							Datacenter:     "dc1",
+							Tags:           []string{"http", "https"},
```
apollo13, line 2167: @kisunji  I found a possible bug here. `http` and it's service `cproxy-http` is not connect enabled. Should this tag show up here then or not? Only `cproxy-https` with the tag `https` is connect enabled.

apollo13: And I guess `Nodes` below could get deduped as well like it is done for the tags:
https://github.com/hashicorp/consul/blob/449e0507417d01a4b44f57b64890c4d8dfccfc4a/agent/ui_endpoint.go#L597 ?

apollo13: I pushed a commit to dedupe the nodes in this branch as well: https://github.com/hashicorp/consul/pull/13023/commits/bf3865f819156c1bd3cd14d8073b9a63f7b925c9

apollo13: Regarding my first comment in this thread: For transparent proxies it is checked if all instances are in the same "state" ( https://github.com/hashicorp/consul/blob/449e0507417d01a4b44f57b64890c4d8dfccfc4a/agent/ui_endpoint.go#L586-L588 ). I guess it would make sense to apply a similar code connect-native services -- I will add code towards that end in another commit. If we think this should not be the case we revert / remove it easily again.

apollo13: Actually I am leaning towards a service to be considered connect native if any of it's instances are. Connect native is not exactly the same as transparent proxies and allows for even more flexibility (like taking to only a subset of instances of a service which are connect enabled etc…). Will push a commit towards that goal.

apollo13:
I have addressed the issues I ran over and successfully tested the changes in our cluster, jay :)

agent/ui_endpoint_test.go
kisunji: Yep, I agree with your decision. Thanks for being thorough with the changes!

hc-github-team-consul-core:
Backport failed @kisunji. Run: https://github.com/hashicorp/consul/actions/runs/5714489659

hc-github-team-consul-core:
Backport failed @kisunji. Run: https://github.com/hashicorp/consul/actions/runs/5714489659

apollo13:
Thank you for reviewing and merging, much appreciated 👍

PR #13049, opened 2022-05-12
Add expanding on environment for stats tags

jorgemarey:
This PR ands enviroment expansion for envoy stats tags. This, along with https://github.com/hashicorp/nomad/pull/12959 will allow us to do things like in the proxy configuration:

```
kind = "proxy-defaults"
name = "global"

Config {
      envoy_dogstatsd_url = "udp://172.26.64.1:9125"
      envoy_stats_tags = [
        "alloc_id=${NOMAD_ALLOC_ID}",
        "nomad_group=${NOMAD_GROUP_NAME}",
        "nomad_job=${NOMAD_JOB_NAME}",
        "nomad_namespace=${NOMAD_NAMESPACE}",
      ]
}
```

### DIFF
--- a/.changelog/13049.txt
+++ b/.changelog/13049.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+connect: Expand variables provided on envoy bootstrap stats tags
+```
--- a/command/connect/envoy/bootstrap_config.go
+++ b/command/connect/envoy/bootstrap_config.go
@@ -505,6 +505,7 @@ func generateStatsTags(args *BootstrapTplArgs, initialTags []string, omitDepreca
 	)
 
 	for _, tag := range initialTags {
+		tag := os.ExpandEnv(tag)
 		parts := strings.SplitN(tag, "=", 2)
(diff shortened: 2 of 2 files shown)


Amier3:
Hey @jorgemarey 

Thanks for making this PR! The team talked about this today and we really like the change, but we're not sure if this should be set on the proxy-defaults layer ( mainly because adding new variables would require a envoy restart, which might lead to a bad UX ). There's also some security considerations we have to make.

We'll sync up on this again shortly and hopefully get you some feedback back on how we'd like to approach this.  Thank you again!

jorgemarey:
Hi @Amier3 any news on this?

jorgemarey:
I think that if you don't consider this to be OK. We can close this in favor of https://github.com/hashicorp/nomad/pull/14311

david-yu:
Hi @jorgemarey I'll go ahead and close this in favor of the Nomad PR you referenced. We believe that is probably a better path forward t get those stats tags for Nomad in.

jorgemarey:
Hi @david-yu, Yeah, you're right, it looks like a better path.

PR #13056, opened 2022-05-12
fix small typo in multi-cluster k8s install docs

sofixa:
Added missing "primary cluster" in the multi-cluster k8s install docs.


im2nguyen (approved):
thanks

PR #13066, opened 2022-05-12
Actually block when syncing subscriptions

freddygv:
By changing to use `WatchCtx` we will actually block for changes to the peering list. `WatchCh` creates a goroutine to collect errors from `WatchCtx` and returns immediately.

The existing behavior wouldn't result in a tight loop because of the rate limiting in the surrounding function, but it would still lead to more work than is necessary.


eculver (approved):

hc-github-team-consul-core:

To retroactively trigger a backport after merging, add backport labels and re-run https://circleci.com/gh/hashicorp/consul/677637.

PR #13069, opened 2022-05-13
peering: fix panic in test

eculver:
### Description
I was futzing around with the API tests and noticed that when the happy path isn't so happy, the tests could panic. It's a one-line change to move a reference to after the assertions are made. 

If you build an older version that doesn't have support for peering and then run the `api` tests, this peering test case will panic. 

### PR Checklist

* [x] updated test coverage

### DIFF
--- a/api/peering_test.go
+++ b/api/peering_test.go
@@ -153,11 +153,12 @@ func TestAPI_Peering_GenerateToken_Read_Initiate_Delete(t *testing.T) {
 	var token1 string
 	// Generate a token happy path
 	resp, wm, err := peerings.GenerateToken(ctx, p1, options)
-	token1 = resp.PeeringToken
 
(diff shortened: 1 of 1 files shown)


api/peering_test.go
```diff
@@ -153,12 +153,13 @@ func TestAPI_Peering_GenerateToken_Read_Initiate_Delete(t *testing.T) {
 	var token1 string
 	// Generate a token happy path
 	resp, wm, err := peerings.GenerateToken(ctx, p1, options)
-	token1 = resp.PeeringToken
```
eculver, line 156: Technically, `resp` can be `nil` here when the endpoint doesn't return in the happy path (aka a test failure)

acpana (approved):
🙇🏼‍♂️ 🚀 🥇

PR #13071, opened 2022-05-13
When a host header is defined override `req.Host` in the metrics ui

### LINKED ISSUE #10164: Metrics Proxy config cannot be set to correct Host header
#### Overview of the Issue

I am running a Nomad cluster with Consul Connect enabled. Running in the cluster, I have a Prometheus server prospecting sidecar proxies to push the data to a Victoriametrics job which is also running in the cluster. Since Consul is running at the host-level it would easiest for me to configure the UI metrics proxy to point to an A record which is backed by an AWS NLB and Traefik which routes to Victoriametrics.

Setting the `base_url` did not work as intended since the original `Host` header was being passed through to Traefik, leading to a `404 page not found` response. Additionally, I attempted to explicitly set `add_headers` to my configuration, however, due to nuances described in https://github.com/golang/go/issues/7682 and https://github.com/golang/go/issues/28168, the `Host` header is not set since the original `req.Host` is never modified.

Briefly researching a potential fix, it is fairly simple to set up an assertion on the `req.Host` in the existing test `TestUIEndpoint_MetricsProxy` to reproduce this behavior. The fix is fairly easy to implement at that point by setting `req.Host` and an `X-Forwarded-For` header within the handler, however, I'm unsure of the best way to present this in the configuration.

#### Reproduction Steps

1. Create a cluster of any size
1. Configure the `metrics_proxy` block with a backend requiring proper `Host` header set
1. Set `add_headers` such that the `Host` is configured to the desired host of the metrics API
1. Consul UI will return 404s as a result of the `Host` header mismatch

<details>
  <summary>Client info</summary>

```
N/A
```

</details>

<details>
  <summary>Server info</summary>

```
agent:
	check_monitors = 0
	check_ttls = 0
	checks = 3
	services = 3
build:
	prerelease =
	revision = f55da930
	version = 1.9.3
consul:
	acl = enabled
	bootstrap = false
	known_datacenters = 1
	leader = false
	leader_addr = 10.4.94.126:8300
	server = true
raft:
	applied_index = 120167925
	commit_index = 120167925
	fsm_pending = 0
	last_contact = 33.212251ms
	last_log_index = 120167925
	last_log_term = 86
	last_snapshot_index = 120159892
	last_snapshot_term = 86
	latest_configuration = [{Suffrage:Voter ID:ef8c875a-703d-038e-7096-cd0205933cf1 Address:10.4.94.126:8300} {Suffrage:Voter ID:ce521f08-fcbc-b2cb-f37e-821f1ffeeb9d Address:10.4.15.246:8300} {Suffrage:Voter ID:f95851fd-f8a3-3b8b-2e47-52110642e146 Address:10.4.86.47:8300}]
	latest_configuration_index = 0
	num_peers = 2
	protocol_version = 3
	protocol_version_max = 3
	protocol_version_min = 0
	snapshot_version_max = 1
	snapshot_version_min = 0
	state = Follower
	term = 86
runtime:
	arch = amd64
	cpu_count = 2
	goroutines = 489
	max_procs = 2
	os = linux
	version = go1.15.6
serf_lan:
	coordinate_resets = 0
	encrypted = true
	event_queue = 0
	event_time = 44
	failed = 0
	health_score = 0
	intent_queue = 0
	left = 0
	member_time = 188156
	members = 16
	query_queue = 0
	query_time = 1
serf_wan:
	coordinate_resets = 0
	encrypted = true
	event_queue = 0
	event_time = 1
	failed = 0
	health_score = 0
	intent_queue = 0
	left = 0
	member_time = 30704
	members = 3
	query_queue = 0
	query_time = 1
```

</details>

- Amazon Linux 2
- Consul 1.7.4

### Log Fragments

The related debug logs are irrelevant since only the result of the `URL.String()` function is logged.

dhiaayachi:
### Description
Fix #10164
[The behaviour of `req.Header.Set` in the go `http` package ignore the Header when it's named Host](https://github.com/golang/go/issues/7682). To se the Host header this need to be set in `req.Host`. 
This PR add the ability to do so for the ui internal endpoint, so users can add an `add_header` directive in Consul config to overwrite the Host header so a proxy can be configured.

Note:
Header names are case insensitive as defined in [here](http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2)

### DIFF
--- a/.changelog/13071.txt
+++ b/.changelog/13071.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+Fix a bug when configuring an `add_headers` directive named `Host` the header is not set for `v1/internal/ui/metrics-proxy/` endpoint.
+```
--- a/agent/ui_endpoint.go
+++ b/agent/ui_endpoint.go
@@ -740,7 +740,11 @@ func (s *HTTPHandlers) UIMetricsProxy(resp http.ResponseWriter, req *http.Reques
 
 	// Add any configured headers
 	for _, h := range cfg.AddHeaders {
-		req.Header.Set(h.Name, h.Value)
+		if strings.ToLower(h.Name) == "host" {
+			req.Host = h.Value
+		} else {
+			req.Header.Set(h.Name, h.Value)
+		}
 	}
 
 	log.Debug("proxying request", "to", u.String())


markan (approved):
TIL: the host header is handled separately from the rest of the headers.

PR #13084, opened 2022-05-13
[CI-only] Build and publish dev dockerhub images 

mdeggies:
### Description

This introduces build/publishing of dev dockerhub images to the hashicorppreview dockerhub org. This will be kicked off when a PR is merged to any of the release branches (or main) listed in the `release_branches` array in `.release/ci.hcl`. 

I tested out a publish from this test branch in commit: https://github.com/hashicorp/consul/commit/c29e03af47aaaa451e17be375b96aebffa54d2fa. You can see the images available here: https://hub.docker.com/repository/docker/hashicorppreview/consul/tags?page=1&ordering=last_updated (which I will delete after this PR is merged). 

### Links
Link to full documentation describing how dev images are created and published, their naming conventions, etc: https://go.hashi.co/dev-docker-images

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,6 @@ on:
     branches:
       # Push events on the main branch
       - main
-      - dev-image-publishing
 
 env:
   PKG_NAME: consul
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -251,3 +251,6 @@ jobs:
           tags: |
             docker.io/hashicorp/${{env.repo}}:${{env.version}}
             public.ecr.aws/hashicorp/${{env.repo}}:${{env.version}}
+          dev_tags: |
+            hashicorppreview/${{ env.repo }}:${{ env.version }}-dev
+            docker.io/hashicorppreview/${{ env.repo }}:${{ env.version }}-${{ github.sha }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,7 @@ on:
     branches:
       # Push events on the main branch
       - main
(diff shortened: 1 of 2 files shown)


alvin-huang (approved):

.github/workflows/build.yml
```diff
@@ -251,3 +251,6 @@ jobs:
           tags: |
             docker.io/hashicorp/${{env.repo}}:${{env.version}}
             public.ecr.aws/hashicorp/${{env.repo}}:${{env.version}}
+          dev_tags: |
+            hashicorppreview/${{ env.repo }}:${{ env.version }}-dev
+            docker.io/hashicorppreview/${{ env.repo }}:${{ env.version }}-${{ github.sha }}
```
eculver, line 256: nit: it looks like the convention in the rest of this file is to not use spaces between the brackets.
```suggestion
            hashicorppreview/${{env.repo}}:${{env.version}}-dev
            docker.io/hashicorppreview/${{env.repo}}:${{env.version}}-${{github.sha}}
```

eculver: Should these tags be different? It looks like one is:
```
v1.12.0-dev
```
And the other is
```
v1.12.0-aabbcc
```

alvin-huang: (the docker.io part is optional)

but the difference between the two tags is one updates `<version>-dev` to be the latest commit on that release branch whereas the other one will be pushed per commit

eculver: Got it. I guess it's small but it would be nice if the two URIs are consistent (either both include `docker.io` domain or both omit it), but not that big of a deal :/

eculver (approved):

Issue #3419, opened 2017-08-25
Consul fails to start if Datadog is down


slackpad:
This will likely be fixed in armon/go-metrics, but capturing here to get the details. This follows up from https://twitter.com/eriktweets/status/901140699164422144.

erkolson:
The startup log is very short.  All three servers look like this:
```
==> WARNING: LAN keyring exists but -encrypt given, using keyring
==> WARNING: WAN keyring exists but -encrypt given, using keyring
==> WARNING: Expect Mode enabled, expecting 3 servers
==> lookup dd-consul-dogstatsd.kube-system.svc.cluster.local on 100.64.0.10:53: no such host
```
The containers get restarted immediately so I assume it exits.

server.json looks like this:
```
{
  "telemetry": {
    "dogstatsd_addr": "dd-consul-dogstatsd.kube-system.svc.cluster.local:8125"
  },
  "raft_protocol": 3,
  "autopilot": {
    "cleanup_dead_servers": true,
    "last_contact_threshold": "200ms",
    "max_trailing_logs": 250,
    "server_stabilization_time": "10s"
  },
  "leave_on_terminate": true
}
```

slackpad:
This one is tricky since the name resolution is failing early on when it's trying to make the statsd client. It's a similar issue to https://github.com/armon/go-metrics/issues/39. Need to think on this a bit but we log a warning and save the address off and periodically try to connect to it. This would also let us reconnect for https://github.com/armon/go-metrics/issues/39 if the connection closes.

PR #13068, opened 2022-05-13
Enable manual triggering of load test

huikang:
### Description
Describe why you're making this change, in plain English.

Allow developers to manually trigger dispatch from Actions

### Testing & Reproduction steps
* In the case of bugs, describe how to replicate
* If any manual tests were done, document the steps and the conditions to replicate
* Call out any important/ relevant unit tests, e2e tests or integration tests you have added or are adding

### Links
Include any links here that might be helpful for people reviewing your PR (Tickets, GH issues, API docs, external benchmarks, tools docs, etc). If there are none, feel free to delete this section.

Please be mindful not to leak any customer or confidential information. HashiCorp employees may want to use our internal URL shortener to obfuscate links.

### PR Checklist

* [ ] updated test coverage
* [ ] external facing docs updated
* [x] not a security concern
* [ ] checklist [folder](./../docs/config) consulted

PR #13091, opened 2022-05-16
Retry on bad dogstatsd connection

huikang:
### Description

- Introduce a new telemetry configurable parameter `dogstatsd_exit_bad_connection`. User can set the value to false to let consul agent continue its start process on failed connection to datadog server. The default behavior is true, which exits the agent.

- An error message is emitted in case `dogstatsd_exit_bad_connection=false` and the dns name of the dogstatds can't be resolved:
```
2022-05-16T09:18:14.638-0400 [WARN]  agent: bootstrap = true: do not enable unless necessary
2022-05-16T09:18:14.698-0400 [ERROR] agent: failed connection to datadog sink: error="lookup dogstatss on 192.168.1.1:53: no such host"
```

**Why not retrying?** When connecting to datadog server in udp, it seems the only place to verify the connection is during consul's initialization phase at https://github.com/DataDog/datadog-go/blob/8bfdc335936a79b55b3e2c1080930bc5a3eb57f2/statsd/udp.go#L23
Since datadog agent doesn't send any ack packet, it is hard to detect if the connection is lost. Therefore, retrying may not be a good solution or help in mitigating the situation.

**Follow up** To overcome the issue of lost connection and inform user of the status, shall we add `dogstatds_connection` to consul's health check?

Start the consult agent with the following telemetry configuration:

```json
  "telemetry": {
    "dogstatsd_exit_bad_connection": false,
    "dogstatsd_addr": "bad.dogstatsd.name:8125"
  }

```

### Links

close https://github.com/hashicorp/consul/issues/3419

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/.changelog/13091.txt
+++ b/.changelog/13091.txt
@@ -1,0 +1,5 @@
+```release-note:improvement
+config: introduce `telemetry.retry_failed_connection` in agent configuration to
+retry on failed connection to any telemetry backend. This prevents agent from being
+stopped on unresolable dogstats dns name; 
+```
--- a/.changelog/13091.txt
+++ b/.changelog/13091.txt
@@ -1,5 +1,5 @@
 ```release-note:improvement
 config: introduce `telemetry.retry_failed_connection` in agent configuration to
-retry on failed connection to any telemetry backend. This prevents agent from being
-stopped on unresolable dogstats dns name; 
+retry on failed connection to any telemetry backend. This prevents the agent from
+exiting if the given DogStatsD DNS name is unresolvable, for example.
 ```
--- a/.github/workflows/load-test.yml
+++ b/.github/workflows/load-test.yml
@@ -3,6 +3,7 @@ on:
     branches:
       - main
     types: [labeled]
+  workflow_dispatch: {}
 
 jobs:
   trigger-load-test:
--- a/.github/workflows/load-test.yml
+++ b/.github/workflows/load-test.yml
@@ -3,7 +3,6 @@ on:
     branches:
       - main
     types: [labeled]
-  workflow_dispatch: {}
 
 jobs:
   trigger-load-test:
--- a/agent/testagent.go
+++ b/agent/testagent.go
@@ -25,6 +25,7 @@ import (
 	"github.com/hashicorp/consul/agent/consul"
 	"github.com/hashicorp/consul/agent/structs"
 	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/consul/lib"
 	"github.com/hashicorp/consul/sdk/freeport"
 	"github.com/hashicorp/consul/sdk/testutil"
 	"github.com/hashicorp/consul/sdk/testutil/retry"
@@ -216,7 +217,9 @@ func (a *TestAgent) Start(t *testing.T) error {
 	bd.Logger = logger
 	// if we are not testing telemetry things, let's use a "mock" sink for metrics
 	if bd.RuntimeConfig.Telemetry.Disable {
-		bd.MetricsHandler = metrics.NewInmemSink(1*time.Second, time.Minute)
+		bd.MetricsConfig = &lib.MetricsConfig{
+			Handler: metrics.NewInmemSink(1*time.Second, time.Minute),
+		}
 	}
 
 	if a.Config != nil && bd.RuntimeConfig.AutoReloadConfigCoalesceInterval == 0 {
--- a/lib/telemetry.go
+++ b/lib/telemetry.go
@@ -1,6 +1,7 @@
 package lib
 
 import (
+	"context"
 	"errors"
 	"net"
 	"time"
@@ -9,6 +10,7 @@ import (
 	"github.com/armon/go-metrics/circonus"
 	"github.com/armon/go-metrics/datadog"
 	"github.com/armon/go-metrics/prometheus"
+	"github.com/hashicorp/consul/lib/retry"
 	"github.com/hashicorp/go-hclog"
 )
 
@@ -156,10 +158,10 @@ type TelemetryConfig struct {
 	// hcl: telemetry { dogstatsd_tags = []string }
 	DogstatsdTags []string `json:"dogstatsd_tags,omitempty" mapstructure:"dogstatsd_tags"`
 
-	// DogstatsdExitBadConnection verify connection to dogstatsd server
+	// DogstatsdRetryBadConnection retries connection to dogstatsd server on initial bad connection
 	//
-	// hcl: telemetry { dogstatsd_exit_bad_connection = (true|false)
-	DogstatsdExitBadConnection bool `json:"dogstatsd_exit_bad_connection,omitempty" mapstructure:"dogstatsd_exit_bad_connection"`
+	// hcl: telemetry { dogstatsd_retry_bad_connection = (true|false)
+	DogstatsdRetryBadConnection bool `json:"dogstatsd_retry_bad_connection,omitempty" mapstructure:"dogstatsd_retry_bad_connection"`
 
 	// FilterDefault is the default for whether to allow a metric that's not
 	// covered by the filter.
@@ -320,6 +322,35 @@ func InitTelemetry(cfg TelemetryConfig, logger hclog.Logger) (*metrics.InmemSink
 		return nil
 	}
 
+	retryWithBackoff := func() {
+		w := &retry.Waiter{}
+		ctx := context.Background()
+
+	retryLoop:
+		for {
+			logger.Info("retry connecting to datadog sink")
+			err := addSink(dogstatdSink)
+
+			switch {
+			case err == nil:
+				logger.Info("connected to datadog sink")
+				break retryLoop
+			// TODO: exit the retry routine on agent exit
+			// case <- done
+			default:
+				logger.Error("failed connection to datadog sink", "error", err)
+			}
+
+			if err := w.Wait(ctx); err != nil {
+				logger.Error("waiting for retry", "error", err)
+				return
+			}
+		}
+
+		sinks = append(sinks, memSink)
+		metrics.NewGlobal(metricsConf, sinks)
+	}
+
 	if err := addSink(statsiteSink); err != nil {
 		return nil, err
 	}
@@ -328,8 +359,9 @@ func InitTelemetry(cfg TelemetryConfig, logger hclog.Logger) (*metrics.InmemSink
 	}
 	if err := addSink(dogstatdSink); err != nil {
 		var dnsError *net.DNSError
-		if errors.As(err, &dnsError) && dnsError.IsNotFound && !cfg.DogstatsdExitBadConnection {
+		if errors.As(err, &dnsError) && dnsError.IsNotFound && cfg.DogstatsdRetryBadConnection {
 			logger.Error("failed connection to datadog sink", "error", err)
+			go retryWithBackoff()
 		} else {
 			return nil, err
 		}
--- a/lib/telemetry.go
+++ b/lib/telemetry.go
@@ -158,10 +158,10 @@ type TelemetryConfig struct {
 	// hcl: telemetry { dogstatsd_tags = []string }
 	DogstatsdTags []string `json:"dogstatsd_tags,omitempty" mapstructure:"dogstatsd_tags"`
 
-	// DogstatsdRetryBadConnection retries connection to dogstatsd server on initial bad connection
+	// RetryFailedConfiguration retries connection to dogstatsd server on initial bad connection
 	//
-	// hcl: telemetry { dogstatsd_retry_bad_connection = (true|false)
-	DogstatsdRetryBadConnection bool `json:"dogstatsd_retry_bad_connection,omitempty" mapstructure:"dogstatsd_retry_bad_connection"`
+	// hcl: telemetry { retry_failed_connection = (true|false)
+	RetryFailedConfiguration bool `json:"retry_failed_connection,omitempty" mapstructure:"retry_failed_connection"`
 
 	// FilterDefault is the default for whether to allow a metric that's not
 	// covered by the filter.
@@ -293,17 +293,7 @@ func circonusSink(cfg TelemetryConfig, hostname string) (metrics.MetricSink, err
 	return sink, nil
 }
 
-// InitTelemetry configures go-metrics based on map of telemetry config
-// values as returned by Runtimecfg.Config().
-func InitTelemetry(cfg TelemetryConfig, logger hclog.Logger) (*metrics.InmemSink, error) {
-	if cfg.Disable {
-		return nil, nil
-	}
-	// Setup telemetry
-	// Aggregate on 10 second intervals for 1 minute. Expose the
-	// metrics over stderr when there is a SIGUSR1 received.
-	memSink := metrics.NewInmemSink(10*time.Second, time.Minute)
-	metrics.DefaultInmemSignal(memSink)
+func configureSinks(cfg TelemetryConfig, hostName string, memSink metrics.MetricSink) error {
 	metricsConf := metrics.DefaultConfig(cfg.MetricsPrefix)
 	metricsConf.EnableHostname = !cfg.DisableHostname
 	metricsConf.FilterDefault = cfg.FilterDefault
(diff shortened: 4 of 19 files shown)


.github/workflows/load-test.yml
```diff
@@ -3,6 +3,7 @@ on:
     branches:
       - main
     types: [labeled]
+  workflow_dispatch: {}
```
eculver, line 6: I think this was inadvertently included from [another PR](https://github.com/hashicorp/consul/pull/13068)?

lib/telemetry.go
```diff
@@ -153,6 +155,11 @@ type TelemetryConfig struct {
 	// hcl: telemetry { dogstatsd_tags = []string }
 	DogstatsdTags []string `json:"dogstatsd_tags,omitempty" mapstructure:"dogstatsd_tags"`
 
+	// DogstatsdExitBadConnection verify connection to dogstatsd server
+	//
+	// hcl: telemetry { dogstatsd_exit_bad_connection = (true|false)
```
eculver, line 160: This is not a blocker, but I am curious what you think about calling this `dogstatsd_verify_connection` if it's not too much trouble to change?  I saw mention of a `splunk_verify_connection` from the external ticket and thought it was simple and descriptive and wanted to get your thoughts.

boxofrad:
Hey @huikang 👋🏻 

Thanks for taking this on!

Although (because this is UDP) we can't be sure that the connection remains good, the problem we're trying to solve here is specifically what happens when the given hostname cannot be resolved but may be resolved in the future.

To that end, I think we _should_ aim to retry with backoff. Otherwise, a transient DNS problem could leave an agent in a state of _never_ emitting metrics.

We could target this specific failure-mode like so:

```Go
var dnsError *net.DNSError
if errors.As(err, &dnsError) && dnsError.NotFound {
	go retryWithBackoff()
}
```

What do you think?

lib/telemetry_test.go
```diff
@@ -0,0 +1,9 @@
+package lib
+
+import (
+	"testing"
+)
+
+func TestInitTelemetry(t *testing.T) {
+	// TODO: add test cases for init telemetry sink
```
eculver, line 8: It might make sense to just cover the case for the datadog config.

eculver (commented):
I'm most curious about @boxofrad' question too. The feedback is pretty small.

.github/workflows/load-test.yml
huikang: thanks for catching this. Will remove this line in the update commit.

lib/telemetry_test.go
huikang: Sounds good. Will add the test case.

boxofrad:
Good question @huikang, and sorry that I'd only briefly looked at the code before commenting 😅 

On reflection, I'm wondering if it'd be easier to understand if we wrapped the entire sink-building process in a retry loop, such that if we fail to configure _any_ of the sinks we'll use whichever we were able to configure and keep trying the others.

I'm imagining a method (called `configureSinks` in the snippet below) that always builds the `sinks` slice from scratch, that way you wouldn't need to worry about guarding it with a mutex in order to mutate it in a retry goroutine.

```Go
func InitTelemetry(cfg TelemetryConfig, logger hclog.Logger) *metrics.InmemSink {
	// ...
	errCh := make(chan error, 1)
	go func() {
		waiter := &retry.Waiter{}

		for {
			sinks, err := configureSinks(cfg, metricsConf.HostName, memSink)
			metrics.NewGlobal(metricsConf, sinks)

			select {
			case errCh <- err:
			default:
			}

			if err == nil || !cfg.RetryFailedConfiguration {
				return
			}

			logger.Warn("failed to configure metric sinks", "retries", waiter.Failures())
			waiter.Wait(context.Background())
		}
	}()

	var err error
	if !cfg.RetryFailedConfiguration {
		err = <-errCh
	}
	return memSink, err
}

func configureSinks(cfg TelemetryConfig, hostName string, memSink metrics.MetricSink) (metrics.MetricSink, error) {
	var (
		sinks  metrics.FanoutSink
		errors error
	)
	addSink := func(fn func(TelemetryConfig, string) (metrics.MetricSink, error)) {
		s, err := fn(cfg, hostName)
		if err != nil {
			errors = multierror.Append(errors, err)
		}
		if s != nil {
			sinks = append(sinks, s)
		}
	}

	addSink(statsiteSink)
	addSink(statsdSink)
	addSink(dogstatdSink)
	addSink(circonusSink)
	addSink(prometheusSink)

	if len(sinks) == 0 {
		return memSink, errors
	}

	return append(sinks, memSink), errors
}
```

We may still want to target specific (i.e. transient) errors to retry, rather than configuration-level errors. Ideally these would be caught during config validation, but I'm not sure if they are or not.

lib/telemetry.go
```diff
@@ -338,5 +338,65 @@ func InitTelemetry(cfg TelemetryConfig) (*metrics.InmemSink, error) {
...
+	metrics.DefaultInmemSignal(memSink)
+	metricsConf := metrics.DefaultConfig(cfg.MetricsPrefix)
+
+	retryWithBackoff := func() {
+		waiter := &retry.Waiter{
+			MaxWait: 5 * time.Minute,
+		}
+
+		for {
+			logger.Warn("failed to configure metric sinks", "retries", waiter.Failures())
```
eculver, line 363: It looks like this "failed to configure ..." message will get printed twice for every failure (here and on line 374). Should this first one say something like "retrying.." or something that follows the failure message below?

lib/telemetry.go
```diff
@@ -338,5 +328,67 @@ func InitTelemetry(cfg TelemetryConfig) (*metrics.InmemSink, error) {
...
+		return nil, nil
+	}
+
+	memSink := metrics.NewInmemSink(10*time.Second, time.Minute)
+	metrics.DefaultInmemSignal(memSink)
+	metricsConf := metrics.DefaultConfig(cfg.MetricsPrefix)
+
+	retryWithBackoff := func() {
+		waiter := &retry.Waiter{
+			MaxWait: 5 * time.Minute,
```
eculver, line 349: We talked about potentially making this configurable, at least on the `TelemetryConfig` and not necessarily externally configurable by operators. This could allow us to set it to a short wait in tests and assert that it reached the timeout, maybe? I just wanted to note it here. It may not actually work out without larger changes.

huikang: Agreed. 

The PR still misses handling the case where the agent process exits and the retry routine is still running. My understanding is that to catch the agent process shutdown signal, we have to extract the [InitTelemetry from the BaseOps method](https://github.com/hashicorp/consul/blob/9acce31237fd9f23363711f4ef25fd1de10fe726/agent/setup.go#L93) and move it after [agent.New](https://github.com/hashicorp/consul/blob/9acce31237fd9f23363711f4ef25fd1de10fe726/command/agent/agent.go#L176). With that, we can pass the agent.ShutdownCh to the retry routine.

huikang:
@boxofrad @boxofrad , I implemented two approaches that stops the retry routine on agent's exit:
1. Pass the agent's shutdownCh to `InitTelemetry` https://github.com/hashicorp/consul/pull/13091/commits/5aa172e6953ec21270922937f6f072d8be1cbb08
2. Add a `Cancel function` to the metricsHanlder struct: https://github.com/hashicorp/consul/pull/13091/files#diff-78adaa1dd41819d1ec7f10c4a7ca9497675cf1ef56515cfb49988283c526a412R220-R224
and call the cancel method in agent shutdown: https://github.com/hashicorp/consul/pull/13091/files#diff-62b6ad581fe3a3059ae8c85ef0f31dde4092bfdecfa7d6857c470bcacaa8cc8b

The first one seems less complex, but move the InitTelemetry out of the Setup approach. The second follows how autoconfig is stopped https://github.com/hashicorp/consul/blob/15b6494d0bdd3316bfd3b4a001de72e39a8cec77/agent/agent.go#L1429-L1431

Please let me know which one is preferred or there is another solution. Thanks.

.changelog/13091.txt
```diff
@@ -0,0 +1,5 @@
+```release-note:improvement
+config: introduce `telemetry.retry_failed_connection` in agent configuration to
+retry on failed connection to any telemetry backend. This prevents agent from being
+stopped on unresolable dogstats dns name;
```
boxofrad, line 4:
```suggestion
retry on failed connection to any telemetry backend. This prevents the agent from
exiting if the given DogStatsD DNS name is unresolvable, for example.
```

lib/telemetry.go
```diff
@@ -338,5 +341,78 @@ func InitTelemetry(cfg TelemetryConfig) (*metrics.InmemSink, error) {
...
+			logger.Error("failed configure sinks", "error", multierror.Flatten(err))
+
+			if err := waiter.Wait(ctx); err != nil {
+				if errors.Is(err, context.Canceled) {
+					logger.Info("stop retrying configure metrics sinks")
+				} else {
+					logger.Error("waiting for retry", "error", err)
+				}
+				return
+			}
```
boxofrad, line 390: I think `context.Canceled` is the only error that could be returned from `Wait` here. That said, I'm a bit confused about the log message we're emitting in the `else` branch (as it will exit/not retry).

lib/telemetry.go
```diff
@@ -338,5 +341,78 @@ func InitTelemetry(cfg TelemetryConfig) (*metrics.InmemSink, error) {
...
+	return metricsConfig, nil
+}
+
+func isRetriableError(errs *multierror.Error) bool {
+	for _, err := range errs.WrappedErrors() {
+		var dnsError *net.DNSError
+		if errors.As(err, &dnsError) && dnsError.IsNotFound {
+			return true
+		}
+	}
```
boxofrad, line 416: I think you can call `erorrs.As` with the multierror directly, rather than iterating over the wrapped errors like this.

From the [docs](https://pkg.go.dev/github.com/hashicorp/go-multierror#Error.Unwrap):

> The resulting error supports errors.As/Is/Unwrap so you can continue to use the stdlib errors package to introspect further.

lib/telemetry.go
```diff
@@ -283,53 +307,32 @@ func circonusSink(cfg TelemetryConfig, hostname string) (metrics.MetricSink, err
...
-func InitTelemetry(cfg TelemetryConfig) (*metrics.InmemSink, error) {
-	if cfg.Disable {
-		return nil, nil
-	}
-	// Setup telemetry
-	// Aggregate on 10 second intervals for 1 minute. Expose the
-	// metrics over stderr when there is a SIGUSR1 received.
-	memSink := metrics.NewInmemSink(10*time.Second, time.Minute)
-	metrics.DefaultInmemSignal(memSink)
+func configureSinks(cfg TelemetryConfig, hostName string, memSink metrics.MetricSink) (metrics.FanoutSink, *multierror.Error) {
```
boxofrad, line 310: I think part of the beauty of the `go-multierror` package is that you can pass them around as standard errors. If we change `isRetryableError` to not manually iterate over the errors, then we can also simplify the type signature here.

```suggestion
func configureSinks(cfg TelemetryConfig, hostName string, memSink metrics.MetricSink) (metrics.FanoutSink, error) {
```

boxofrad (commented):
This is looking great, nice work @huikang 👏🏻 🎉 

I've a handful of comments inline. I think the only real blocker is the `IsRetrying` thread-safety, though.

In answer to your question, I much prefer the pattern of having the agent responsible for shutting the goroutine down (by calling `Cancel`).

lib/telemetry.go
huikang: @eculver , when I revisit your question, I realized that Waiter doesn't have a timeout field https://github.com/hashicorp/consul/blob/5f96b63ae7d663b50b68a03c7b6855194b5119f4/lib/retry/retry.go#L34-L50

In the latest version, a cancel function is added to stop the retry routine on agent exit. Please take a look. Thanks.

huikang: @eculver , when I revisit your question, I realized that Waiter doesn't have a timeout field https://github.com/hashicorp/consul/blob/5f96b63ae7d663b50b68a03c7b6855194b5119f4/lib/retry/retry.go#L34-L50

In the latest version, a cancel function is added to stop the retry routine on agent exit. Please take a look. Thanks.

lib/telemetry.go
```diff
@@ -338,5 +351,67 @@ func InitTelemetry(cfg TelemetryConfig) (*metrics.InmemSink, error) {
...
+			logger.Warn("retrying configure metric sinks", "retries", waiter.Failures())
+			_, err := configureSinks(cfg, metricsConf.HostName, memSink)
+			if err == nil {
+				logger.Info("successfully configured metrics sinks")
+				return
+			}
+			logger.Error("failed configure sinks", "error", multierror.Flatten(err))
+
+			if err := waiter.Wait(ctx); err != nil {
+				logger.Info("stop retrying configure metrics sinks")
```
boxofrad, line 390: nit: info may be a bit noisy for this.

```suggestion
				logger.Trace("stop retrying configure metrics sinks")
```

boxofrad (approved):
Awesome work! Thanks again for taking this on 🙇🏻

lib/telemetry.go
eculver: Thanks for revisiting. I am wondering if rather than morphing it into a "timeout" we would just expose the configuration on the `TelemetryConfig` (eg. `MinFailures`, `MaxWait`, `MinWait`, `Jitter`) so that the callers could configure the behavior if necessary. It is certainly not a blocking concern. I will leave it up to you.

lib/telemetry_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func newCfg() TelemetryConfig {
+	cfg := TelemetryConfig{
+		StatsdAddr:    "statsd.host:1234",
+		StatsiteAddr:  "statsite.host:1234",
+		DogstatsdAddr: "mydog.host:8125",
+	}
+	return cfg
```
eculver, line 20: nit: you can just inline this.
```suggestion
	return TelemetryConfig{
		StatsdAddr:    "statsd.host:1234",
		StatsiteAddr:  "statsite.host:1234",
		DogstatsdAddr: "mydog.host:8125",
	}
```

lib/telemetry.go
```diff
@@ -199,6 +212,27 @@ type TelemetryConfig struct {
...
+	cancelFn context.CancelFunc
+}
+
+func (cfg *MetricsConfig) Cancel() {
+	cfg.mu.Lock()
+	defer cfg.mu.Unlock()
+
+	if fn := cfg.cancelFn; fn != nil {
+		fn()
+	}
```
eculver, line 233: nit: can't we write this without the assignment?
```suggestion
	if cfg.cancelFn != nil {
		cfg.cancelFn()
	}
```

agent/testagent.go
```diff
@@ -216,7 +217,9 @@ func (a *TestAgent) Start(t *testing.T) error {
 	bd.Logger = logger
 	// if we are not testing telemetry things, let's use a "mock" sink for metrics
 	if bd.RuntimeConfig.Telemetry.Disable {
-		bd.MetricsHandler = metrics.NewInmemSink(1*time.Second, time.Minute)
```
eculver, line 219: This is interesting because I would think that we should be able to configure multiple handlers, but this would lead me to believe that technically, we only allow a single metrics handler? It may be rare that users set more than one, but I'm wondering if this is just a deficiency in [our docs](https://www.consul.io/docs/agent/config/config-files#telemetry). If I'm reading them right, I would assume that any configured backend would be used including if I configure more than one.

eculver (commented):
I really like how this has shaped up! Good stuff. I just have a few questions, most are small.

lib/telemetry.go
```diff
@@ -338,5 +351,67 @@ func InitTelemetry(cfg TelemetryConfig) (*metrics.InmemSink, error) {
...
+
+			if err := waiter.Wait(ctx); err != nil {
+				logger.Trace("stop retrying configure metrics sinks")
+			}
+		}
+	}
+
+	if _, errs := configureSinks(cfg, metricsConf.HostName, memSink); errs != nil {
+		if isRetriableError(errs) && cfg.RetryFailedConfiguration {
+			logger.Error("failed configure sinks", "error", multierror.Flatten(errs))
```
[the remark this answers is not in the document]
huikang: Fixed

eculver (approved):

PR #13098, opened 2022-05-16
fix insecure cidr_blocks in load test

huikang:
### Description

### Links

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/test/load/terraform/consul.tf
+++ b/test/load/terraform/consul.tf
@@ -37,9 +37,9 @@ module "consul_servers" {
 
   vpc_id                  = module.vpc.vpc_id
   subnet_ids              = module.vpc.public_subnets
-  allowed_ssh_cidr_blocks = ["10.0.0.0/16"]
+  allowed_ssh_cidr_blocks = [var.vpc_cidr]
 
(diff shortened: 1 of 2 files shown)


test/load/terraform/consul.tf
```diff
@@ -37,9 +37,9 @@ module "consul_servers" {
 
   vpc_id                  = module.vpc.vpc_id
   subnet_ids              = module.vpc.public_subnets
-  allowed_ssh_cidr_blocks = ["0.0.0.0/0"]
+  allowed_ssh_cidr_blocks = ["10.0.0.0/16"]
```
mkeeler, line 40: Should all of these CIDRS come from `var.vpc_cidr`? That way if the VPC's network addressing is not 10.0.0.0/16 we can still get to them from within the VPC?

mkeeler (approved):
LGTM

Tracking this all the way back. CI seems to always be using 10.0.0.0/16 as the VPC cidr which eventually gets passed as the var to this tf module.

https://github.com/hashicorp/consul/blob/78412ae069e6b056ed460ca2c619dc5f0b72a829/.circleci/terraform/load-test/main.tf#L17

So moving to using the var seems perfect.

PR #13099, opened 2022-05-16
ci: update the version of the underlying vm for these two test types

rboyer:
### Description

I noticed a few of the vm CircleCI jobs were using slightly different ubuntu VM images so this PR unifies them to use the same running.

Maybe this fixes the flakes on the `go-test-arm64` job, but i'm not 100% on that.


kisunji (approved):
if this fixes the arm64 flakes :chefs-kiss:

eculver (approved):

Issue #11250, opened 2021-10-08
Allow terminating gateways to pick up system certs


freddygv:
#### Feature Description

Currently for terminating gateways the certificate file needs to be specified explicitly in the terminating gateway's config entry:
https://www.consul.io/docs/connect/config-entries/terminating-gateway#cafile

We should improve this UX by looking in known directories for different operating systems:
https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/crypto/x509/root_linux.go
https://github.com/golang/go/blob/ce9a3b79d5bb783f5f31c9d41665a488fe63f546/src/crypto/x509/root_bsd.go
https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/crypto/x509/root_unix.go
https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/crypto/x509/root_solaris.go
https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/crypto/x509/root_aix.go
https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/crypto/x509/root_js.go

#### Use Case(s)

This would simplify the use of terminating gateways when they route to HTTPS services with certificates signed by well-known authorities.

Debugging TLS errors due to this issue has been very time consuming in the times I've encountered it.

#### Implementation notes

1. To avoid enabling this behavior by default, we could configure it with the terminating gateway itself at the service level. (Or maybe gateway level)

2. That setting would then need to be made available for proxy configuration in the proxycfg package. That could potentially be done via the GatewayService type.

3. Then these certificates would need to be used when generating Envoy's TLS config [here](https://github.com/hashicorp/consul/blob/1c9b58a8af49f448e7dad2868eba52347b202337/agent/xds/clusters.go#L379).

PR #13109, opened 2022-05-17
feat(tgateway): allow finding system CAs on service level

DanStough:
# !Draft!

### Description
First pass implementation of checking for SSL certificates that requires enabling it on a service-to-service level. See  https://github.com/hashicorp/consul/issues/11250. 

#### TODO:
* [ ] Figure out a way to set this as at proxydefaults or gateway level configuration.
* [ ] Figure out if the protocol of the external service can be inspected for https and System CAs can be enabled automatically.

### Assumptions
1. It doesn't make sense to specify TLS options in transparent proxy mode (end-destination TLS is the client's responsibility, envoy should manage mTLS tunnel).
1. CAFile takes precedent over System specification

### Links

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/agent/catalog_endpoint_test.go
+++ b/agent/catalog_endpoint_test.go
@@ -1436,6 +1436,11 @@ func TestCatalog_GatewayServices_Terminating(t *testing.T) {
 					KeyFile:  "client.key",
 					SNI:      "my-alt-domain",
 				},
+				{
+					Name:        "example",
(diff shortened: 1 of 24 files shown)


dhiaayachi (commented):
I think for the time being we are tabling this. We will see if we still need it after the work on the egress tproxy.

PR #13126, opened 2022-05-18
test: cleanup and unflake parts of the upgrade compat tests

rboyer:
### Description

The compatibility tests seemed particularly flaky so I took a pass to attempt to unflake them. I also did some light cleanup and refactoring and added in some missing error handling.

I also added `make test-compat-integ` to aid in local debugging.

### DIFF
--- a/test/integration/consul-container/libs/node/consul-container.go
+++ b/test/integration/consul-container/libs/node/consul-container.go
@@ -26,6 +26,24 @@ type consulContainerNode struct {
 	container testcontainers.Container
 	ip        string
 	port      int
+	config    Config
+	req       testcontainers.ContainerRequest
+	dataDir   string
+}
+
+func (c *consulContainerNode) GetConfig() Config {
+	return c.config
+}
+
+func newConsulContainerWithReq(ctx context.Context, req testcontainers.ContainerRequest) (testcontainers.Container, error) {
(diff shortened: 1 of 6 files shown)


test/integration/consul-container/upgrade/healthcheck_test.go
```diff
@@ -72,6 +70,8 @@ func TestTargetServersWithLatestGAClients(t *testing.T) {
 
 // Test health check GRPC call using Mixed (majority latest) Servers and Latest GA Clients
 func TestMixedServersMajorityLatestGAClient(t *testing.T) {
+	t.Parallel()
```
dhiaayachi, line 73: I removed Parallel run at some point because it added to the flakiness. If we are confident it's fixed now we can put it back everywhere otherwise let's keep it with no parallel run for now.

test/integration/consul-container/libs/node/consul-container.go
```diff
@@ -26,6 +26,24 @@ type consulContainerNode struct {
 	container testcontainers.Container
```
rboyer, line 26: These changes are picking up some other changes that had not been backported from enterprise.

dhiaayachi (approved):
A small comment about `t.Parallel` otherwise LGTM!

Issue #12628, opened 2022-03-26
Can't register multiple services when using HTTP2/gRPC listeners for ingress gateway


Dgotlieb:
### Consul version
Consul v1.10.1
Revision db839f18b

### Operating system and Environment details
Distributor ID:	Ubuntu
Description: Ubuntu 18.04.6 LTS
Release: 18.04
Codename: bionic

### Background
According to the [Ingress gateway documentation](https://www.consul.io/docs/connect/gateways/ingress-gateway#ingress-gateways):
> For listeners with a [protocol](https://www.consul.io/docs/connect/config-entries/ingress-gateway#protocol) other than tcp, multiple services can be specified for a single listener.

### Issue
When trying to register multiple services for `HTTP2` **OR** `gRPC` listener an error is thrown:
`Error submitting job: Unexpected response code: 500 (rpc error: Unexpected response code: 500 (Multiple services per listener are only supported for protocol = 'http' (listener on port 8011)))`

### Expected Result
Allow registering multiple services for `HTTP2` and `gRPC` listeners

### Note
I'm using Nomad to register my services, but given the fact that this is Consul error, I assume it doesn't matter..

### Job file
``` hcl
job "test" {
  type = "service"
  datacenters = ["dc1"]

  group "test" {
    count = 1

    network {
      mode = "bridge"

      port "test" {
        to  = 8011
      }
    }

    service {
      name = "test"

      connect {
        gateway {
          proxy {}
          ingress {
             listener {
               port     = 8011
               protocol = "http2" # or gRPC
               service {
                 name  = "service-a"
                 hosts = ["foo"]
               }
               service {
                 name  = "service-b"
                 hosts = ["bar"]
               }
             }
          }
        }
      }
    }
  }
}
```

### P.S.
If this is by design and not a bug, please fix the documentation to:  
> For listeners with a `http` [protocol](https://www.consul.io/docs/connect/config-entries/ingress-gateway#protocol) ~other than tcp,~ multiple services can be specified for a single listener.  

Thanks!

PR #13127, opened 2022-05-18
fix: multiple grpc/http2 services for ingress listeners

DanStough:
### Description
https://github.com/hashicorp/consul/issues/12628 reported that there was a mismatch between the documentation and the consul CLI when it comes to using multiple grpc services per ingress listener. The error received is a validation on the config entry. Removing the validation for grpc/https allows traffic to be routed as expected, which is inline with the expectations in the docs.

* I set up two grpc services (fortio and moul/grpcbin) IAW with the [tutorial](https://learn.hashicorp.com/tutorials/consul/service-mesh-ingress-gateways), using `service-defaults` to set their protocols to `grpc`. Using the proposed code, I was able to route traffic to each of the services using the `authority` header.

`fortio-service.hcl`
```hcl
service {
 name = "fortio"
 id = "fortio-1"
 port = 8079

 connect {
   sidecar_service {}
 }

 check {
  id = "fortio-check"
  name = "Fortio Health Check"
  grpc = "127.0.0.1:8079"
  interval = "10s"
 }

}
```

`grpcbin-service.hcl`
```hcl
service {
 name = "grpcbin"
 id = "grpcbin-1"
 port = 9000

 connect {
   sidecar_service {}
 }
}
```

`ingress-gateway.hcl`
```hcl
Kind = "ingress-gateway"
Name = "ingress-service"

Listeners = [
 {
   Port = 10000
   Protocol = "grpc"
   Services = [
     {
       Name = "fortio"
       Hosts = ["fortio"]
     },
     {
       Name = "grpcbin"
       Hosts = ["grpcbin"]
     }
   ]
 }
]
```
Commands to the listener
`grpcurl -plaintext -authority fortio 192.168.205.20:10000 fgrpc.PingServer/Ping`
`grpcurl -plaintext -authority grpcbin 192.168.205.20:10000 hello.HelloService/SayHello`

### PR Checklist

* [X] updated test coverage
* [X] not a security concern

### DIFF
--- a/agent/structs/config_entry_gateways_test.go
+++ b/agent/structs/config_entry_gateways_test.go
@@ -141,6 +141,26 @@ func TestIngressGatewayConfigEntry(t *testing.T) {
 				Listeners: []IngressListener{
 					{
 						Port:     1111,
+						Protocol: "http",
+						Services: []IngressService{
+							{
+								Name: "db1",
+							},
+							{
+								Name: "db2",
+							},
+						},
+					},
+				},
+			},
+		},
+		"http features: multiple services on tcp listener": {
+			entry: &IngressGatewayConfigEntry{
+				Kind: "ingress-gateway",
+				Name: "ingress-web",
+				Listeners: []IngressListener{
+					{
+						Port:     1111,
 						Protocol: "tcp",
 						Services: []IngressService{
 							{
@@ -153,7 +173,7 @@ func TestIngressGatewayConfigEntry(t *testing.T) {
 					},
 				},
 			},
-			validateErr: "Multiple services per listener are only supported for protocol",
+			validateErr: "Multiple services per listener are only supported for L7",
 		},
 		// ==========================
 		"tcp listener requires a defined service": {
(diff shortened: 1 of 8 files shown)


DanStough:
This feels like something that might need an E2E test, but not sure where to write the test or if someone can point me in the right direction.

kyhavlov (commented):
I think this looks good - you could also add a golden file test similar to this listener test for an ingress gateway with multiple http services: https://github.com/hashicorp/consul/blob/main/agent/xds/listeners_test.go#L672-L675

That would show that multiple http2/grpc services get synthesized to the right Envoy config.

DanStough:
Added tests per @kyhavlov's suggestion.

agent/structs/config_entry_gateways_test.go
```diff
@@ -135,6 +135,26 @@ func TestIngressGatewayConfigEntry(t *testing.T) {
...
+							},
+							{
+								Name: "db2",
+							},
+						},
+					},
+				},
+			},
+		},
+		"http features: multiple services on tpc listener": {
```
kyhavlov, line 157: Typo (should be `tcp listener`)

kyhavlov (approved):
Nice! The tests look good, just one non-blocking typo 👍

PR #13143, opened 2022-05-19
Add connection limit setting to service defaults

kyhavlov:
This PR adds a setting, `max_inbound_connections` to the `service-defaults`, for limiting the number of inbound connections to each service instance at a time. This is done by setting both a `maxConnections` circuit breaker value on the `local_app` cluster as well as an `envoy.filters.network.connection_limit` filter on the public listener.


boxofrad (approved):
Looks good to me! But may be worth a 👍🏻 from somebody with a little more Envoy-fu. Nice job!

go.mod
```diff
@@ -18,9 +18,9 @@ require (
 	github.com/coreos/go-oidc v2.1.0+incompatible
 	github.com/docker/go-connections v0.3.0
 	github.com/elazarl/go-bindata-assetfs v0.0.0-20160803192304-e1a2a7ec64b0
-	github.com/envoyproxy/go-control-plane v0.9.9
+	github.com/envoyproxy/go-control-plane v0.10.1
 	github.com/fsnotify/fsnotify v1.5.1
-	github.com/golang/protobuf v1.4.3
+	github.com/golang/protobuf v1.5.0
```
kisunji, line 23: Are these version bumps necessary? I feel like the gogo folks intentionally left the protobuf modules at a certain version but if nothing's broken maybe it's all good

rboyer: We only stopped at a set of `protoc`, `golang/protobuf`, `grpc`, `go-control-plane` libraries that mostly worked together. If this new set of versions also work together than that's fine.

paging @mkeeler just in case this at all matters for `buf` things.

PR #13145, opened 2022-05-19
test: fix more flakes in the compatibility test

rboyer:
### Description

I discovered a few more things that would reduce flakes (or the threat of flakes) in the compat tests.

- waited until the _client_ could detect a leader sometimes
- ensures each test cluster used a uniquely random serf encryption key to prevent stray old carryover containers from automatically joining new clusters by accident
- fixed some errant references to `t` instead of `r` in `retry.RunWith` (the actual bug most likely)


dhiaayachi (approved):
🚢

PR #13159, opened 2022-05-19
feat: add endpoint struct to ServiceConfigEntry

DanStough:
### Description
First PR of many covering new terminating gateway functionality allowing external traffic w/o node registration and transparent proxying enabled.

This PR adds an `endpoint` field to `service-defaults` which represents a service external to Consul. It can be a hostname (w/ or w/o a wildcard), ipv4/ipv6 address or ipv4/ipv6 CIDR. Service defaults will represent the "virtual service", which will be consumed by the existing `Services` definition in `TerminatingGatewayConfigEntry`.

Unit tests added. Higher-level testing will happen when we start integrating the watch with the terminating gateway xDS generation.

### PR Checklist

* [X] updated test coverage
* [X] not a security concern

### DIFF
--- a/agent/structs/config_entry.go
+++ b/agent/structs/config_entry.go
@@ -207,13 +207,6 @@ func (e *ServiceConfigEntry) Validate() error {
 		if e.Endpoint.Port < 1 || e.Endpoint.Port > 65535 {
 			validationErr = multierror.Append(validationErr, fmt.Errorf("Invalid Port number %d", e.Endpoint.Port))
 		}
-
-		// If either client cert config file was specified then the CA file, client cert, and key file must be specified
-		// Specifying only a CAFile is allowed for one-way TLS
-		if (e.Endpoint.CertFile != "" || e.Endpoint.KeyFile != "") &&
-			!(e.Endpoint.CAFile != "" && e.Endpoint.CertFile != "" && e.Endpoint.KeyFile != "") {
-			validationErr = multierror.Append(validationErr, errors.New("Endpoint must have a CertFile, CAFile, and KeyFile specified for TLS origination"))
-		}
 	}
 
 	return validationErr
@@ -236,22 +229,6 @@ func validateEndpointAddress(address string) error {
 		return fmt.Errorf("Could not validate address %s as an IP, CIDR block or Hostname", address)
 	}
 	return nil
-}
-
-func (e *ServiceConfigEntry) Warnings() []string {
-	if e == nil {
-		return nil
-	}
-	warnings := make([]string, 0)
-
-	if e.Endpoint != nil {
-		if (e.Endpoint.CAFile != "" || e.Endpoint.CertFile != "" || e.Endpoint.KeyFile != "") && e.Endpoint.SNI == "" {
-			warning := fmt.Sprintf("TLS is configured but SNI is not set for the endpoint. Enabling SNI is strongly recommended when using TLS.")
-			warnings = append(warnings, warning)
-		}
-	}
-
-	return warnings
 }
 
 func (e *ServiceConfigEntry) CanRead(authz acl.Authorizer) error {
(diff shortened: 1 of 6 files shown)


agent/structs/config_entry.go
```diff
@@ -253,6 +320,30 @@ func (c *UpstreamConfiguration) Clone() *UpstreamConfiguration {
 	return &c2
 }
 
+// EndpointConfig represents a virtual service, i.e. one that is external to Consul
+type EndpointConfig struct {
```
DanStough, line 324: I left out protocol since it's already part of `service-defaults` and I think it's confusing to have it in two places.

I left in the TLS stuff because I think the spirit of this is that the tls authorization should be at the service level and not the gateway level.

agent/structs/config_entry.go
```diff
@@ -190,9 +199,67 @@ func (e *ServiceConfigEntry) Validate() error {
 		}
 	}
 
+	if e.Endpoint != nil {
+		gatewayTLSEnabled := e.Endpoint.CAFile != ""
+
+		// TODO (dans): need to find a way to validate this against the proxy defaults.
+		if gatewayTLSEnabled && e.Mode == ProxyModeTransparent {
+			validationErr = multierror.Append(validationErr, errors.New("Certificates can not be defined in an endpoint when proxy is in transparent mode; these are the responsibility of the service dialer"))
```
DanStough, line 207: Based on what we talked about earlier in the week.

DanStough:
I can't really explain why, but bumping the version of `bats` solved the `integration-test` [problems](https://app.circleci.com/pipelines/github/hashicorp/consul/31652/workflows/f550327a-535d-4b3f-a910-2bea75a1920a/jobs/687002) 🤷‍♂️ .

agent/structs/config_entry_test.go
```diff
@@ -427,6 +427,48 @@ func TestDecodeConfigEntry(t *testing.T) {
 				},
 			},
 		},
+		{
```
rboyer, line 430: Kitchen sink config entry tests exist in 3 places. You're only missing `command/config/write/config_write_test.go`

test/integration/connect/envoy/Dockerfile-bats
```diff
@@ -1,6 +1,6 @@
 FROM docker.mirror.hashicorp.services/fortio/fortio AS fortio
 
-FROM docker.mirror.hashicorp.services/bats/bats:1.6.0
```
rboyer, line 3: Why did you need to bump this?

agent/structs/config_entry.go
rboyer: I think the TLS stuff should probably still continue to live in the gateway config entry.

These reference files that must already exist on the machine. `service-defaults` sorta represents the service in the abstract and may be used in a variety of consuming services on many machines whereas the terminating gateway config entry largely exists 1:1 with some actual real instance of the gateway running on a machine or small set of machines where you can arrange for those files to exist.

test/integration/connect/envoy/Dockerfile-bats
DanStough: See above https://github.com/hashicorp/consul/pull/13159#issuecomment-1133395868

rboyer (approved):
LGTM with non blocking comment about a dead function argument.

agent/structs/config_entry.go
```diff
@@ -190,9 +199,38 @@ func (e *ServiceConfigEntry) Validate() error {
...
+
+		if e.Endpoint.Port < 1 || e.Endpoint.Port > 65535 {
+			validationErr = multierror.Append(validationErr, fmt.Errorf("Invalid Port number %d", e.Endpoint.Port))
+		}
+	}
+
 	return validationErr
 }
 
+func validateEndpointAddress(tlsEnabled bool, address string) error {
```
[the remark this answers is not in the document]
DanStough: Nope, good call. Will remove

rboyer (approved):

PR #13162, opened 2022-05-20
ci: name unnamed GHA workflows

eculver:
### Description
This updates our GitHub Action workflows to all have `name` specified so that the filename isn't used instead, which is a little nicer on the eyes and easier to see which is which.

See the checks for this PR. This illustrates the state before this change:
Screen Shot 2022-05-19 at 6 14 39 PM


acpana (approved):
🚢 💯 !

PR #13164, opened 2022-05-20
peering: expose IsLeader, hung up on dialer if follower

acpana:

### Description
This patch adds `IsLeader` to the peering backend to use when handling `StreamResource()` as a receiver. If the service runs on a consul that is not a leader, then we hang up on the dialer. Otherwise, calls to raft apply as a non leader would fail.

In the future (future PR), we may want to send a message to the dialer with the address of the leader but for now, we just hung up.

* `stream_test.go` updated

### Links
Asana tasks [1](https://app.asana.com/0/1201832349181086/1202294390627574) & [2](https://app.asana.com/0/1201832349181086/1202045399358213)

### PR Checklist

* [x] updated test coverage

### DIFF
--- a/agent/consul/peering_backend.go
+++ b/agent/consul/peering_backend.go
@@ -15,6 +15,7 @@ import (
 )
 
 type peeringBackend struct {
+	// TODO(peering): accept a smaller interface; maybe just funcs from the sever that we actually need: DC, IsLeader, etc
 	srv      *Server
 	connPool GRPCClientConner
 	apply    *peeringApply
@@ -31,6 +32,7 @@ func NewPeeringBackend(srv *Server, connPool GRPCClientConner) peering.Backend {
 	}
 }
 
+// Forward should not be used to initiate forwarding over bidirectional streams
 func (b *peeringBackend) Forward(info structs.RPCInfo, f func(*grpc.ClientConn) error) (handled bool, err error) {
 	// Only forward the request if the dc in the request matches the server's datacenter.
 	if info.RequestDatacenter() != "" && info.RequestDatacenter() != b.srv.config.Datacenter {
@@ -101,6 +103,10 @@ func (b *peeringBackend) Apply() peering.Apply {
 
 func (b *peeringBackend) EnterpriseCheckPartitions(partition string) error {
 	return b.enterpriseCheckPartitions(partition)
+}
+
+func (b *peeringBackend) IsLeader() bool {
+	return b.srv.IsLeader()
 }
 
 type peeringApply struct {
--- a/agent/consul/peering_backend.go
+++ b/agent/consul/peering_backend.go
@@ -15,7 +15,7 @@ import (
 )
 
 type peeringBackend struct {
-	// TODO(peering): accept a smaller interface; maybe just funcs from the sever that we actually need: DC, IsLeader, etc
+	// TODO(peering): accept a smaller interface; maybe just funcs from the server that we actually need: DC, IsLeader, etc
 	srv      *Server
 	connPool GRPCClientConner
 	apply    *peeringApply
--- a/agent/rpc/peering/service.go
+++ b/agent/rpc/peering/service.go
@@ -31,7 +31,6 @@ var (
 	errPeeringTokenEmptyServerAddresses = errors.New("peering token server addresses value is empty")
 	errPeeringTokenEmptyServerName      = errors.New("peering token server name value is empty")
 	errPeeringTokenEmptyPeerID          = errors.New("peering token peer ID value is empty")
-	errPeeringBackendNil                = errors.New("peering backend was not initialized")
 )
 
 // errPeeringInvalidServerAddress is returned when an initiate request contains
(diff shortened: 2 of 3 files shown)


agent/consul/peering_backend.go
```diff
@@ -15,6 +15,7 @@ import (
 )
 
 type peeringBackend struct {
+	// TODO(peering): accept a smaller interface; maybe just funcs from the sever that we actually need: DC, IsLeader, etc
```
rboyer, line 18:
```suggestion
	// TODO(peering): accept a smaller interface; maybe just funcs from the server that we actually need: DC, IsLeader, etc
```

agent/rpc/peering/service.go
```diff
@@ -98,6 +99,9 @@ type Backend interface {
 
 	Subscribe(req *stream.SubscribeRequest) (*stream.Subscription, error)
 
+	// IsLeader indicates whether the consul server is a in a leader state or not
```
rboyer, line 102:
```suggestion
	// IsLeader indicates whether the consul server is in a leader state or not
```

agent/rpc/peering/service.go
```diff
@@ -427,6 +431,19 @@ type BidirectionalStream interface {
 
 // StreamResources handles incoming streaming connections.
 func (s *Service) StreamResources(stream pbpeering.PeeringService_StreamResourcesServer) error {
+	if s.Backend == nil {
+		s.logger.Error("cannot establish stream without a backend", "error", errPeeringBackendNil)
+		return errPeeringBackendNil
+	}
```
rboyer, line 437: I don't think this is necessary as no other endpoints check this. You will get a nil pointer dereference elsewhere anyway, and a failure to have configured a backend is a programmer error, not something that will happen at runtime.

agent/rpc/peering/service.go
```diff
@@ -427,6 +431,19 @@ type BidirectionalStream interface {
...
+
+	if !s.Backend.IsLeader() {
+		// we are not the leader so we will hang up on the dialer
+
+		// TODO(peering): in the future we want to indicate the address of the leader server as a message to the dialer (best effort, non blocking)
+		s.logger.Error("cannot establish a peering stream on a follower node;")
+		return io.EOF
```
rboyer, line 444: I think you also need to do something similar down around line 581 in the `for...select` loop in HandleResources.

agent/rpc/peering/service.go
```diff
@@ -38,6 +38,7 @@ var (
 	errPeeringTokenEmptyServerAddresses = errors.New("peering token server addresses value is empty")
 	errPeeringTokenEmptyServerName      = errors.New("peering token server name value is empty")
 	errPeeringTokenEmptyPeerID          = errors.New("peering token peer ID value is empty")
+	errPeeringBackendNil                = errors.New("peering backend was not initialized")
```
kisunji, line 41: This sounds like it should be a panic or an error defined at the validation level. I don't think it belongs with the other sentinel errors here since it's not a class of errors where a caller can fix it

agent/rpc/peering/service.go
kisunji, line 444: Can we return something like `grpcstatus.Error` instead of a raw `io.EOF`? I find the EOF misleading since the stream hasn't even been established yet and it sort of pretends that it has been terminated.

agent/rpc/peering/stream_test.go
```diff
@@ -24,6 +24,50 @@ import (
...
+			Response: &pbpeering.ReplicationMessage_Response{
+				ResourceURL: pbpeering.TypeURLService,
+				ResourceID:  "api-service",
+				Nonce:       "2",
+			},
+		},
+	}
+
+	err := client.Send(input)
+	require.NoError(t, err)
```
kisunji, line 63: I'm not sure if these lines are valuable; you're testing a `MockClient.Send` which will never return errors. We don't need to send a message for `StreamResources` to return an error.

kisunji (approved):

agent/rpc/peering/service.go
kisunji: Ah RB also mentioned it below

agent/rpc/peering/stream_test.go
rboyer: You are correct about the current "start of stream" leader check. If he were to add the "during a stream" leader check too, then it would need to be done using a Send.

agent/rpc/peering/service.go
acpana: taken out

agent/rpc/peering/service.go
acpana: thanks for all the guidance last week; added a similar check in `HandleStream`

acpana: I was trying to find a "spec" like thing and at some point I saw `EOF` as a good way to indicate that the stream is closed.  I added a `grpcstatus.Error` w a failed precondition code. Wdyt?

rboyer (approved):

PR #13168, opened 2022-05-20
agent: allow for service discovery queries involving peer name to use streaming

rboyer:
### Description

Requesting service discovery data for peered services that have been imported should make use of streaming if enabled.

### Links

[Asana](https://app.asana.com/0/1201832349181086/1201834451438964/f)

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/health_endpoint_test.go
+++ b/agent/health_endpoint_test.go
@@ -770,6 +770,16 @@ func TestHealthServiceNodes(t *testing.T) {
 }
 
 func TestHealthServiceNodes_Blocking(t *testing.T) {
+	t.Run("local data", func(t *testing.T) {
+		testHealthServiceNodes_Blocking(t, structs.DefaultPeerKeyword)
+	})
+
+	t.Run("peered data", func(t *testing.T) {
+		testHealthServiceNodes_Blocking(t, "my-peer")
+	})
+}
+
+func testHealthServiceNodes_Blocking(t *testing.T, peerName string) {
 	cases := []struct {
 		name         string
 		hcl          string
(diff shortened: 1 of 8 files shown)


agent/health_endpoint_test.go
```diff
@@ -959,16 +1005,26 @@ use_streaming_backend = true
 		},
 	}
 
+	suffix := func(peerName string) string {
+		if peerName == "" {
+			return ""
+		}
+		return "&peer=" + peerName
+	}
```
dhiaayachi, line 1013: This is defined in a couple of places, maybe extract it as helper?

proto/pbservice/ids.go
```diff
@@ -22,12 +22,14 @@ func (m *CheckServiceNode) UniqueID() string {
 	switch {
 	case m.Node != nil:
 		builder.WriteString(m.Node.Partition + "/")
+		builder.WriteString(m.Node.PeerName + "/")
```
dhiaayachi, line 25: Are we cool with a PeerName being empty here and have a `//` in the UniqueID?

dhiaayachi (approved):
A question and a non blocking comment but other than that LGTM!

proto/pbservice/ids.go
rboyer: We already do this for Partition and Namespace here, as the `ids.go` file is not split.

There's already a test for:
```
		{
			name: "without partition or namespace or peer name",
			csn: &CheckServiceNode{
				Node: &Node{
					Node: "the-node-name",
				},
				Service: &NodeService{
					ID: "the-service-id",
				},
			},
			expected: "//the-node-name//the-service-id",
		},
```

PR #12897, opened 2022-04-29
ui: [Backport/1.12.x] Fixup icons not displaying in Safari (#12738)

evrowe:
Backport of #12738

Icons that were using currentColor were not displaying their color in Safari (leading to invisible icons). See original PR for more details.

no changelog due to base not being `main`

PR #13183, opened 2022-05-23
ui: Icon related fixups

johncowen:
Also see:
- https://github.com/hashicorp/consul/pull/12738
- https://github.com/hashicorp/consul/pull/12897

### Description
I went through and fixed up anything icon related that needed it

Small fixes:

1. Fix invisible "meatball/popup menu" icons that we use within all our list views.
2. Fix up some invisible icons in the Routing Tab
3. Fix up the badges in the Intentions Permissions filter

Larger Fix.

Previous to releasing 1.12.0 we saw that our icons had turned invisible in Safari. It turns out that the approach we are using to support easily defining icons of different resolutions isn't going to work due to this.

There's quite a lot of background to take into account here, so please ask if more information is required (theres another PR that I closed once I'd got waaaay into the weeds with this: https://github.com/hashicorp/consul/pull/13166).

The TLDR is, when defining a currentColor within more than one keyframe for an animation, Safari completely ignores the setting (within one keyframe its fine, its just when you add it to more than one)

```css
/* Doesn't work in Safari: */
@keyframes icon-name {
  50% {
      mask-image: url(thin-icon);
      background-color: currentColor;
   }
  100% {
      mask-image: url(thick-icon);
      background-color: currentColor;
   }
}
/* Works: */
@keyframes icon-name {
  100% {
      mask-image: url(thick-icon);
      background-color: currentColor;
   }
}
```

A previous fix meant defaulting all pseudo elements to have a background-color of currentColor (https://github.com/hashicorp/consul/pull/12738), which whilst worked, felt weird, so I wasn't happy with that. Now I've figured out exactly what the problem is, I'd rather not do this.

Instead, as we only use a different variation of icon in one place, and its so minor I've reverted that to the same resolution as all our other icons. Longer term, we can use a different approach for the odd time we need to use the different resolution icons.

Small note: The last commit here is just a machine re-generation of all our icon files using just a single keyframe.

### Testing
Manual/visual QA in Chrome, Safari and FF

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


evrowe (approved):
lgtm

PR #13184, opened 2022-05-23
build: wire up remaining 5 helper tools into the same auto-install logic used for protobuf tools

rboyer:
### Description

When doing the recent updates to the protobuf toolchain some auto-install logic was added to the build to ensure that supporting go tools matched the desired versions at use time.

This PR uses the same machinery for the bindata tools, mock generation, and linting tools.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -2,34 +2,16 @@
 # https://www.consul.io/docs/install#compiling-from-source
 
 SHELL = bash
-GOTOOLS = \
-	github.com/elazarl/go-bindata-assetfs/go-bindata-assetfs@master \
-	github.com/hashicorp/go-bindata/go-bindata@master \
-	github.com/vektra/mockery/v2@latest \
-	github.com/golangci/golangci-lint/cmd/golangci-lint@v1.46.2 \
-	github.com/hashicorp/lint-consul-retry@master
 
 ###
-# BUF_VERSION can be either a valid string for "go install <module>@<version>"
+# These version variables can either be a valid string for "go install <module>@<version>"
 # or the string @DEV to imply use what is currently installed locally.
 ###
+GOLANGCI_LINT_VERSION='v1.46.2'
+MOCKERY_VERSION='v2.12.2'
 BUF_VERSION='v1.4.0'
-
-###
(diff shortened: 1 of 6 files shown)


GNUmakefile
```diff
@@ -283,29 +265,34 @@ other-consul:
...
 
-lint:
-	@echo "--> Running go golangci-lint"
+lint: lint-tools
+	@echo "--> Running golangci-lint"
 	@golangci-lint run --build-tags '$(GOTAGS)' && \
 		(cd api && golangci-lint run --build-tags '$(GOTAGS)') && \
 		(cd sdk && golangci-lint run --build-tags '$(GOTAGS)')
+	@echo "--> Running lint-consul-retry"
+	@lint-consul-retry
```
rboyer, line 274: Now we can all enjoy this tool locally.

build-support/functions/25-tools.sh
```diff
@@ -0,0 +1,59 @@
+function install_unversioned_tool {
```
rboyer, line 1: Straight lift out of their former home.

build-support/scripts/devtools.sh
```diff
@@ -0,0 +1,102 @@
...
+        'golangci-lint' \
+        'github.com/golangci/golangci-lint' \
+        "${golangci_lint_version}" \
+        'github.com/golangci/golangci-lint/cmd/golangci-lint'
+}
+
+function bindata_install {
+    install_unversioned_tool \
+        'go-bindata' \
+        'github.com/hashicorp/go-bindata/go-bindata@bf7910a'
```
rboyer, line 77: These hashes were the ones pinned in the GH actions file. This stuff can all go away with https://github.com/hashicorp/consul/pull/10996 eventually.

build-support/scripts/devtools.sh
```diff
@@ -0,0 +1,102 @@
...
+        'mockery' \
+        'github.com/vektra/mockery/v2' \
+        "${mockery_version}" \
+        'github.com/vektra/mockery/v2'
+
+    lint_install
+    bindata_install
+
+    return 0
+}
```
mkeeler, line 99: Would it be worthwhile to have a proto_install target here and move the installation bits from protobuf.sh into that?

That way all our dependency/tools installs are setup in this code?

PR #13192, opened 2022-05-24
Update main to reflect 1.13.0-dev

eculver:
### Description
Currently, the version reported is `1.12.0-dev`. Since `1.12.0` is now GA, the version reported should be `1.13.0-dev`. Going forward, we should do this bump after every major release. For the other release series (eg. `release/1.12.x`), the reported version will still be `1.12.X-dev` where `X` should match the latest in the series. 

`make version`

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -14,7 +14,7 @@ var (
 	//
 	// Version must conform to the format expected by github.com/hashicorp/go-version
 	// for tests to work.
-	Version = "1.12.0"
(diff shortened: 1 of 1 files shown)


rboyer (approved):
Note I think you'll have to do something similar in enterprise for a ent-only file.

eculver:
> Note I think you'll have to do something similar in enterprise for a ent-only file.

Yep. I'm on it.

PR #13196, opened 2022-05-24
feat: Listener Generation for Terminating Gateway Destinations

DanStough:
### Description
This PR adds xDS listener generation to Destinations (fka "Endpoints") for terminating gateways. To recap, destinations are external services that are not part of the Consul cluster. They are nodeless and also support IP CIDRs. Destinations are referenced the same as regular `services` in the Terminating Gateway Config Entry but are defined in `service-defaults` (#13159).

### Testing
Tests with Golden files are included for the three types of Destinations.

### Links

Related PRs:
1. #13217 Updates the gateway-service table to qualify different kinds ("default" and "destinations").
2. https://github.com/hashicorp/consul/pull/13341 to update the intentions to work with destinations.

### PR Checklist

* [X] updated test coverage
* [X] not a security concern

### DIFF
--- a/agent/proxycfg/terminating_gateway.go
+++ b/agent/proxycfg/terminating_gateway.go
@@ -63,6 +63,7 @@ func (s *handlerTerminatingGateway) initialize(ctx context.Context) (ConfigSnaps
 	snap.TerminatingGateway.ServiceResolversSet = make(map[structs.ServiceName]bool)
 	snap.TerminatingGateway.ServiceGroups = make(map[structs.ServiceName]structs.CheckServiceNodes)
 	snap.TerminatingGateway.GatewayServices = make(map[structs.ServiceName]structs.GatewayService)
+	snap.TerminatingGateway.DestinationServices = make(map[structs.ServiceName]structs.GatewayService)
 	snap.TerminatingGateway.HostnameServices = make(map[structs.ServiceName]structs.CheckServiceNodes)
 	return snap, nil
 }
@@ -111,10 +112,15 @@ func (s *handlerTerminatingGateway) handleUpdate(ctx context.Context, u UpdateEv
 			svcMap[svc.Service] = struct{}{}
 
 			// Store the gateway <-> service mapping for TLS origination
-			snap.TerminatingGateway.GatewayServices[svc.Service] = *svc
+			if svc.ServiceKind == structs.GatewayServiceKindDestination {
+				snap.TerminatingGateway.DestinationServices[svc.Service] = *svc
+			} else {
+				snap.TerminatingGateway.GatewayServices[svc.Service] = *svc
+			}
 
 			// Watch the health endpoint to discover endpoints for the service
-			if _, ok := snap.TerminatingGateway.WatchedServices[svc.Service]; !ok {
+			if _, ok := snap.TerminatingGateway.WatchedServices[svc.Service]; !ok && isGatewayServiceKindService(svc) {
+
 				ctx, cancel := context.WithCancel(ctx)
 				err := s.dataSources.Health.Notify(ctx, &structs.ServiceSpecificRequest{
 					Datacenter:     s.source.Datacenter,
@@ -213,7 +219,8 @@ func (s *handlerTerminatingGateway) handleUpdate(ctx context.Context, u UpdateEv
 
 			// Watch service resolvers for the service
 			// These are used to create clusters and endpoints for the service subsets
-			if _, ok := snap.TerminatingGateway.WatchedResolvers[svc.Service]; !ok {
+			if _, ok := snap.TerminatingGateway.WatchedResolvers[svc.Service]; !ok && isGatewayServiceKindService(svc) {
+
 				ctx, cancel := context.WithCancel(ctx)
 				err := s.dataSources.ConfigEntry.Notify(ctx, &structs.ConfigEntryQuery{
 					Datacenter:     s.source.Datacenter,
@@ -239,6 +246,13 @@ func (s *handlerTerminatingGateway) handleUpdate(ctx context.Context, u UpdateEv
 		for sn := range snap.TerminatingGateway.GatewayServices {
 			if _, ok := svcMap[sn]; !ok {
 				delete(snap.TerminatingGateway.GatewayServices, sn)
+			}
+		}
+
+		// Delete endpoint service mapping for services that were not in the update
+		for sn := range snap.TerminatingGateway.DestinationServices {
+			if _, ok := svcMap[sn]; !ok {
+				delete(snap.TerminatingGateway.DestinationServices, sn)
 			}
 		}
 
@@ -371,4 +385,8 @@ func (s *handlerTerminatingGateway) handleUpdate(ctx context.Context, u UpdateEv
 	}
 
 	return nil
+}
+
+func isGatewayServiceKindService(svc *structs.GatewayService) bool {
+	return svc.ServiceKind == structs.GatewayServiceKindService || svc.ServiceKind == structs.GatewayServiceKindUnknown
 }
--- a/agent/proxycfg/testing_terminating_gateway.go
+++ b/agent/proxycfg/testing_terminating_gateway.go
@@ -6,12 +6,7 @@ import (
 	"github.com/hashicorp/consul/agent/structs"
 )
 
-func TestConfigSnapshotTerminatingGateway(
-	t testing.T,
-	populateServices bool,
-	nsFn func(ns *structs.NodeService),
-	extraUpdates []UpdateEvent,
-) *ConfigSnapshot {
+func TestConfigSnapshotTerminatingGateway(t testing.T, populateServices bool, nsFn func(ns *structs.NodeService), extraUpdates []UpdateEvent) *ConfigSnapshot {
 	roots, _ := TestCerts(t)
 
 	var (
@@ -34,6 +29,7 @@ func TestConfigSnapshotTerminatingGateway(
 		},
 	}
 
+	tgtwyServices := []*structs.GatewayService{}
 	if populateServices {
 		webNodes := TestUpstreamNodes(t, web.Name)
 		webNodes[0].Service.Meta = map[string]string{"version": "1"}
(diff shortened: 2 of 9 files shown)


agent/xds/listeners.go
```diff
@@ -1068,14 +1063,48 @@ func (s *ResourceGenerator) makeTerminatingGatewayListener(
...
 	// listeners at idle.
 	sort.Slice(l.FilterChains, func(i, j int) bool {
-		return l.FilterChains[i].FilterChainMatch.ServerNames[0] < l.FilterChains[j].FilterChainMatch.ServerNames[0]
+		outI, _ := proto.Marshal(l.FilterChains[i].FilterChainMatch)
+		crcI := crc32.Checksum(outI, crc32q)
+
+		outJ, _ := proto.Marshal(l.FilterChains[j].FilterChainMatch)
+		crcJ := crc32.Checksum(outJ, crc32q)
+
+		return crcI < crcJ
```
DanStough, line 1107: I would definitely be open to a lighter-weight solution here. Even if we went with this scheme, the new order scheme seems like it should be something that only rolls on a major update.

agent/xds/testdata/listeners/terminating-gateway-custom-and-tagged-addresses.latest.golden
```diff
@@ -14,7 +14,7 @@
         {
           "filterChainMatch": {
             "serverNames": [
-              "api.default.dc1.internal.11111111-2222-3333-4444-555555555555.consul"
+              "cache.default.dc1.internal.11111111-2222-3333-4444-555555555555.consul"
             ]
```
DanStough, line 18: The order of these changed because now I am taking the checksum of the filter.

agent/cache-types/resolved_service_config.go
```diff
@@ -17,6 +17,7 @@ type ResolvedServiceConfig struct {
 	RPC RPC
 }
 
+// TODO (dans): This is what queries for services
```
DanStough, line 20: Junk

agent/xds/listeners.go
```diff
@@ -1068,14 +1056,48 @@ func (s *ResourceGenerator) makeTerminatingGatewayListener(
...
 	// listeners at idle.
 	sort.Slice(l.FilterChains, func(i, j int) bool {
-		return l.FilterChains[i].FilterChainMatch.ServerNames[0] < l.FilterChains[j].FilterChainMatch.ServerNames[0]
+		if len(l.FilterChains[i].FilterChainMatch.PrefixRanges) > 0 && len(l.FilterChains[j].FilterChainMatch.PrefixRanges) > 0 {
+			return l.FilterChains[i].FilterChainMatch.PrefixRanges[0].AddressPrefix < l.FilterChains[j].FilterChainMatch.PrefixRanges[0].AddressPrefix
+		}
+		if len(l.FilterChains[i].FilterChainMatch.ServerNames) > 0 && len(l.FilterChains[j].FilterChainMatch.ServerNames) > 0 {
+			return l.FilterChains[i].FilterChainMatch.ServerNames[0] < l.FilterChains[j].FilterChainMatch.ServerNames[0]
+		}
+		return false
```
dhiaayachi, line 1100: I changed this to use string compare. It's less generic but preserve the order. Also, I think we had a bug in here. If serverNames were empty it would panic.

dhiaayachi:
@DanStough I made few changes in here to fix the order in the sort algorithm and regenerate the files. I also opened https://github.com/hashicorp/consul/pull/13300 to add the leaf cert generation and some testing around having valid endpoints. 

I think we need to do some renaming in this PR before being able to open it for review. So it match https://github.com/hashicorp/consul/pull/13217 naming convention.

agent/proxycfg/terminating_gateway.go
```diff
@@ -111,10 +112,14 @@ func (s *handlerTerminatingGateway) handleUpdate(ctx context.Context, u UpdateEv
 			svcMap[svc.Service] = struct{}{}
 
 			// Store the gateway <-> service mapping for TLS origination
-			snap.TerminatingGateway.GatewayServices[svc.Service] = *svc
+			if svc.IsEndpoint {
```
DanStough, line 115: This will change to check the `kind` field when rebased against #13217 
https://github.com/hashicorp/consul/pull/13217/files#diff-26c007c7c0d5e30fabd316255cc13ac3ffcfac5295ae8a66bcca29c197054664R610

agent/structs/config_entry_gateways.go
```diff
@@ -599,6 +599,7 @@ type GatewayService struct {
 	KeyFile      string   `json:",omitempty"`
 	SNI          string   `json:",omitempty"`
 	FromWildcard bool     `json:",omitempty"`
+	IsEndpoint   bool     `json:",omitempty"`
```
DanStough, line 602: See above comment about this changing in #13217

DanStough:
Asking for a preliminary review. I still need Dhia's PR to land and rename all the "endpoint" stuff to "desintations" and also adopt the `GatewayServiceKind`.

DanStough:
On second thought, I think this PR needs some more work w.r.t. transparent proxy mode in the tests.

DanStough:
Updates were made based on the new design. Ready to review.

agent/proxycfg/terminating_gateway.go
```diff
@@ -213,7 +219,8 @@ func (s *handlerTerminatingGateway) handleUpdate(ctx context.Context, u UpdateEv
 
 			// Watch service resolvers for the service
 			// These are used to create clusters and endpoints for the service subsets
-			if _, ok := snap.TerminatingGateway.WatchedResolvers[svc.Service]; !ok {
+			if _, ok := snap.TerminatingGateway.WatchedResolvers[svc.Service]; !ok && !(svc.ServiceKind == structs.GatewayServiceKindDestination) {
```
dhiaayachi, line 222: Same comment as above

agent/proxycfg/terminating_gateway.go
```diff
@@ -242,6 +249,13 @@ func (s *handlerTerminatingGateway) handleUpdate(ctx context.Context, u UpdateEv
 			}
 		}
 
+		// Delete endpoint service mapping for services that were not in the update
+		for sn := range snap.TerminatingGateway.DestinationServices {
+			if _, ok := svcMap[sn]; !ok {
+				delete(snap.TerminatingGateway.DestinationServices, sn)
+			}
+		}
```
dhiaayachi, line 257: What happen when the service is deleted in a future iteration with no update to the destination with the same name?

dhiaayachi (changes requested):
Thank you for working on this @DanStough. I had a few questions/comments.

agent/proxycfg/terminating_gateway.go
```diff
@@ -111,10 +112,15 @@ func (s *handlerTerminatingGateway) handleUpdate(ctx context.Context, u UpdateEv
...
-			snap.TerminatingGateway.GatewayServices[svc.Service] = *svc
+			if svc.ServiceKind == structs.GatewayServiceKindDestination {
+				snap.TerminatingGateway.DestinationServices[svc.Service] = *svc
+			} else {
+				snap.TerminatingGateway.GatewayServices[svc.Service] = *svc
+			}
 
 			// Watch the health endpoint to discover endpoints for the service
-			if _, ok := snap.TerminatingGateway.WatchedServices[svc.Service]; !ok {
+			if _, ok := snap.TerminatingGateway.WatchedServices[svc.Service]; !ok && !(svc.ServiceKind == structs.GatewayServiceKindDestination) {
```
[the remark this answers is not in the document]
DanStough: Agree!

As an aside, I think we should rename `GatewayServiceKindService` to `GatewayServiceKindTypical` to match mirror regular service kinds.

agent/proxycfg/testing_terminating_gateway.go
```diff
@@ -342,6 +340,123 @@ func TestConfigSnapshotTerminatingGateway(
 	}, nsFn, nil, testSpliceEvents(baseEvents, extraUpdates))
 }
 
+func TestConfigSnapshotTerminatingGatewayDestinations(t testing.T, populateDestinations bool, extraUpdates []UpdateEvent) *ConfigSnapshot {
```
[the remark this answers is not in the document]
DanStough: Yes, you are correct, but I would clarify my understanding:

- a root cert: [Provided Here](https://github.com/hashicorp/consul/blob/b312e4126ad378277b4fd107d0e984190b9fb7a2/agent/proxycfg/testing_terminating_gateway.go#L352-L355)
- a leaf cert: [Provided Here](https://github.com/hashicorp/consul/blob/b312e4126ad378277b4fd107d0e984190b9fb7a2/agent/proxycfg/testing_terminating_gateway.go#L403-L409)
- an intention with that destination as target: An empty intention match is [provided here](https://github.com/hashicorp/consul/blob/b312e4126ad378277b4fd107d0e984190b9fb7a2/agent/proxycfg/testing_terminating_gateway.go#L386-L393). I didn't dig into why the existing logic only looks that the list of matches is greater than zero, but I'm assuming that means at least one intention allows this interaction. This is the same as the existing logic for GatewayServicesKindService.
- the destination service-defaults config-entry: this is implied by the existence in the GatewayServices table and your other PR, https://github.com/hashicorp/consul/pull/13217.
- a tgtwy that reference the destination: same as above, this is implied by the GatewayServices entry that maps the Service <-> Gateway

> In the test I don't see the intention configured and I think we are missing the leaf cert which is added in https://github.com/hashicorp/consul/pull/13300

I'm not sure why the [existing watch for leaf-certs](https://github.com/hashicorp/consul/blob/b312e4126ad378277b4fd107d0e984190b9fb7a2/agent/proxycfg/terminating_gateway.go#L178-L196) wouldn't apply here. We don't qualify it by the `GatewayServiceKind`. I think that's what the test case should demonstrate unless you disagree.

agent/xds/listeners.go
```diff
@@ -1068,14 +1056,64 @@ func (s *ResourceGenerator) makeTerminatingGatewayListener(
 		}
 	}
 
+	for _, svc := range cfgSnap.TerminatingGateway.ValidDestinations() {
+		clusterName := connect.ServiceSNI(svc.Name, "", svc.NamespaceOrDefault(), svc.PartitionOrDefault(), cfgSnap.Datacenter, cfgSnap.Roots.TrustDomain)
+
+		intentions := cfgSnap.TerminatingGateway.Intentions[svc]
+		svcConfig := cfgSnap.TerminatingGateway.ServiceConfigs[svc]
+
+		cfg, err := ParseProxyConfig(svcConfig.ProxyConfig)
```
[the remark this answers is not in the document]
DanStough: Hmm, I think it's a little confusing to add it to the `destination` struct since it's already in the `service-defaults` and `proxy-defaults` structs. We could make it private, but then I think it's more work to make sure it's inherited correctly.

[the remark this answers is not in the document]
DanStough: Per our discussion, we'll keep it as-is now and we can add this separate field later if needed.

DanStough:
@markan I'm closing this one in favor of #13391, which has these changes and a few more to cover all of the xDS on the T-gtwy side.

PR #13199, opened 2022-05-24
[CI-only] Update tagging for dev_tags

mdeggies:
### Description

Small update to remove the hardcoded `-dev` suffix from dev_tags, which is causing tags to be in the format `1.12.0-dev-dev` instead of just `1.12.0-dev`. I'll clean up the old tags before making the dockerhub repo public, which will be available https://hub.docker.com/r/hashicorppreview/consul. 

dev_tags were first introduced in https://github.com/hashicorp/consul/pull/13084


eculver (approved):
Aha! I noticed this recently but hadn't gotten around to fixing it. Thanks!

mdeggies:
Thanks! I noticed this right after merging the last one, but took awhile to get the PR up 🙈 this should fix it

PR #13204, opened 2022-05-24
api: ensure peering API endpoints do not use protobufs

rboyer:
### Description

I noticed that the JSON api endpoints for peerings json encodes protobufs directly, rather than converting them into their `api` package equivalents before marshal/unmarshaling them.

I updated this and used `mog` to do the annoying part in the middle. 

Other changes:
- the status enum was converted into the friendlier string form of the enum for readability with tools like `curl`
- some of the `api` library functions were slightly modified to match other similar endpoints in UX (cc: @ndhanushkodi )
- peeringRead returns `nil` if not found
- partitions are NOT inferred from the agent's partition (matching 1.11-style logic)


proto/pbpeering/peering.go
```diff
@@ -83,3 +87,88 @@ func (p *Peering) ShouldDial() bool {
 func (x ReplicationMessage_Response_Operation) GoString() string {
 	return x.String()
 }
+
+// TODO(rb): figure how to use to use mog to maintain this mapping when states are added
```
kisunji, line 91: I've used https://github.com/reillywatson/enumcover to lint for complete enum coverage

ndhanushkodi (approved):
The signature changes lgtm! I'll watch for this being merged.

api/peering.go
```diff
@@ -23,53 +35,38 @@ type Peering struct {
...
 	// PeerServerAddresses contains all the connection addresses for the remote peer.
-	PeerServerAddresses []string
+	PeerServerAddresses []string `json:",omitempty"`
 	// CreateIndex is the Raft index at which the Peering was created.
 	CreateIndex uint64
 	// ModifyIndex is the latest Raft index at which the Peering. was modified.
 	ModifyIndex uint64
 }
 
-type PeeringReadRequest struct {
```
acpana, line 47: hm, I recall the rationale for adding these in specifically was to future proof the API. Do we see no reason to keep these around at all?

agent/peering_endpoint.go
```diff
@@ -18,30 +20,29 @@ func (s *HTTPHandlers) PeeringEndpoint(resp http.ResponseWriter, req *http.Reque
...
 // peeringRead fetches a peering that matches the name and partition.
 // This assumes that the name and partition parameters are valid
-func (s *HTTPHandlers) peeringRead(resp http.ResponseWriter, req *http.Request, name, partition string) (interface{}, error) {
+func (s *HTTPHandlers) peeringRead(resp http.ResponseWriter, req *http.Request, name string) (interface{}, error) {
 	args := pbpeering.PeeringReadRequest{
 		Name:       name,
 		Datacenter: s.agent.config.Datacenter,
-		Partition:  partition, // should be "" in OSS
 	}
+	var entMeta acl.EnterpriseMeta
```
acpana, line 41: just out of curiosity, what were the trade offs for deciding to move this check into `peeringRead` and `peeringWrite`?  As opposed to keeping it above,  in the general `PeeringEndpoint` func?

api/peering.go
```diff
@@ -23,53 +35,38 @@ type Peering struct {
...
 	Meta map[string]string `json:",omitempty"`
 	// State is one of the valid PeeringState values to represent the status of
 	// peering relationship.
 	State PeeringState
-	// PeerID is the ID that our peer assigned to this peering.
-	// This ID is to be used when dialing the peer, so that it can know who dialed it.
-	PeerID string
+	// PeerID is the ID that our peer assigned to this peering. This ID is to
+	// be used when dialing the peer, so that it can know who dialed it.
+	PeerID string `json:",omitempty"`
```
acpana, line 46: 👀 I recall @kisunji and I chatting about whether these need to be here. I thought only ent specific things like `Partition` or `Namespace` had to have these.

acpana (commented):
hey @rboyer thanks for taking this on 💯  -- it looks great 🥇 ; I had a couple questions, some for my learning, some bc I am confused. 

I didn't get a chance to review this yesterday so I'm leaving this in as post commit commentary.

agent/peering_endpoint.go
rboyer: Consistency mostly. Almost all other endpoints follow this sort of structure for extracting partition/namespace data from requests to populate rpc fields.

agent/peering_endpoint.go
```diff
@@ -51,76 +52,77 @@ func (s *HTTPHandlers) peeringRead(resp http.ResponseWriter, req *http.Request,
...
 }
 
 // PeeringList fetches all peerings in the datacenter in OSS or in a given partition in Consul Enterprise.
 func (s *HTTPHandlers) PeeringList(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
-	entMeta := s.agent.AgentEnterpriseMeta()
-	if err := s.parseEntMetaPartition(req, entMeta); err != nil {
-		return nil, err
-	}
-
 	args := pbpeering.PeeringListRequest{
```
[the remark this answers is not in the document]
rboyer: Yes. Missed this.

api/peering.go
rboyer: When sending data to the consul agent ideally you would only be sending fields with actual data. If you don't include this struct tag then every field is sent do the agent in something like a write request. In the past we've had subtle version incompatibilities caused when the api library became aware of a new field and then that new field was sent to an older consul version, since we hard-fail requests at the consul (json api) agents if they see a field they don't understand.

api/peering.go
rboyer: Peering read and Peering list were both http `GET` operations, which do not commonly take payloads, and any request body IIRC behaves slightly differently than something like a POST.

Peering delete is an http `DELETE` operation, which also does not expect a request body.

Having these methods accept a large request struct may give the illusion that you could send the struct over the wire, which you could not. Everything has to be sent via path elements, headers, and query parameters.

In existing endpoints in situations where we wanted extensibility we've done something like adding `PeeringReadOpt{s,ions}` to optionally follow the primary REST identifier (peer name), where each of those gets parked onto the request in a header or parameter. We  could add those if extensibility was a concern.

PR #13217, opened 2022-05-25
update gateway-services table with endpoints

dhiaayachi:
### Description
This part of the work needed to add destinations (formerly endpoints) to the terminating gateway. Destinations are defined using a service-default config entry and represent an external entity (a set of addresses and a port) which can be reachable through a given terminating gateway.

In this PR we add the association between a given destination and a terminating gateway. This include the use case of a terminating gateway that define a wildcard in his LinkedServices.

### PR Checklist

* [X] updated test coverage
* [X] not a security concern

### DIFF
--- a/agent/consul/state/catalog.go
+++ b/agent/consul/state/catalog.go
@@ -846,9 +846,12 @@ func ensureServiceTxn(tx WriteTxn, idx uint64, node string, preserveIndexes bool
 	}
 
 	if svc.PeerName == "" {
-		// Check if this service is covered by a gateway's wildcard specifier
-		if err = checkGatewayWildcardsAndUpdate(tx, idx, svc); err != nil {
-			return fmt.Errorf("failed updating gateway mapping: %s", err)
+		// Do not associate non-typical services with gateways or consul services
+		if svc.Kind == structs.ServiceKindTypical && svc.Service != "consul" {
+			// Check if this service is covered by a gateway's wildcard specifier
+			if err = checkGatewayWildcardsAndUpdate(tx, idx, &structs.ServiceName{Name: svc.Service, EnterpriseMeta: svc.EnterpriseMeta}, false); err != nil {
+				return fmt.Errorf("failed updating gateway mapping: %s", err)
+			}
 		}
 	}
 	if err := upsertKindServiceName(tx, idx, svc.Kind, svc.CompoundServiceName()); err != nil {
@@ -3469,6 +3472,7 @@ func terminatingConfigGatewayServices(
 		return false, nil, fmt.Errorf("unexpected config entry type: %T", conf)
 	}
 
+	// TODO(egress-tproxy): Check if it's an endpoint by quering the service and the service defaut config entry.
 	// Check if service list matches the last known list for the config entry, if it does, skip the update
 	_, c, err := configEntryTxn(tx, nil, conf.GetKind(), conf.GetName(), entMeta)
 	if err != nil {
@@ -3483,6 +3487,10 @@ func terminatingConfigGatewayServices(
 
 	var gatewayServices structs.GatewayServices
 	for _, svc := range entry.Services {
+		isEndpoint, err := isEndpoint(tx, svc.Name, &svc.EnterpriseMeta)
+		if err != nil {
+			return false, nil, fmt.Errorf("failed to query endpoints: %v", err)
+		}
 		mapping := &structs.GatewayService{
 			Gateway:     gateway,
 			Service:     structs.NewServiceName(svc.Name, &svc.EnterpriseMeta),
@@ -3491,11 +3499,34 @@ func terminatingConfigGatewayServices(
 			CertFile:    svc.CertFile,
 			CAFile:      svc.CAFile,
 			SNI:         svc.SNI,
+			IsEndpoint:  isEndpoint,
 		}
 
 		gatewayServices = append(gatewayServices, mapping)
 	}
 	return false, gatewayServices, nil
+}
+
+func isEndpoint(tx ReadTxn, name string, entMeta *acl.EnterpriseMeta) (bool, error) {
+	serviceIter, err := tx.Get(tableServices, indexService, Query{
+		Value:          name,
+		EnterpriseMeta: *entMeta,
+	})
+	if err != nil {
+		return false, err
+	}
+	for service := serviceIter.Next(); service != nil; service = serviceIter.Next() {
+		return false, nil
+	}
+
+	_, entry, err := configEntryTxn(tx, nil, structs.ServiceDefaults, name, entMeta)
+	if err != nil {
+		return false, err
+	}
+	if entry != nil {
+		return true, nil
+	}
+	return false, nil
 }
 
 // updateGatewayNamespace is used to target all services within a namespace
@@ -3531,6 +3562,41 @@ func updateGatewayNamespace(tx WriteTxn, idx uint64, service *structs.GatewaySer
 		mapping := service.Clone()
 
 		mapping.Service = structs.NewServiceName(sn.ServiceName, &service.Service.EnterpriseMeta)
+		mapping.FromWildcard = true
+
+		err = updateGatewayService(tx, idx, mapping)
+		if err != nil {
+			return err
+		}
+	}
+	endpoints, err := tx.Get(tableConfigEntries, indexID+"_prefix", ConfigEntryKindQuery{Kind: structs.ServiceDefaults})
+	if err != nil {
+		return fmt.Errorf("failed querying endpoints: %s", err)
+	}
+	for endpoint := endpoints.Next(); endpoint != nil; endpoint = endpoints.Next() {
+		e := endpoint.(*structs.ServiceConfigEntry)
+		if e.Endpoint == nil {
+			continue
+		}
+
+		sn := structs.ServiceName{
+			Name:           e.Name,
+			EnterpriseMeta: e.EnterpriseMeta,
+		}
+		existing, err := tx.First(tableGatewayServices, indexID, service.Gateway, sn, service.Port)
+		if err != nil {
+			return fmt.Errorf("gateway service lookup failed: %s", err)
+		}
+		if existing != nil {
+			// If there's an existing service associated with this gateway then we skip it.
+			// This means the service was specified on its own, and the service entry overrides the wildcard entry.
+			continue
+		}
+
+		mapping := service.Clone()
+
+		mapping.Service = structs.NewServiceName(e.Name, &service.Service.EnterpriseMeta)
+		mapping.IsEndpoint = true
 		mapping.FromWildcard = true
 
 		err = updateGatewayService(tx, idx, mapping)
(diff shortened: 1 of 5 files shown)


agent/consul/state/catalog.go
```diff
@@ -3538,6 +3569,41 @@ func updateGatewayNamespace(tx WriteTxn, idx uint64, service *structs.GatewaySer
 			return err
 		}
 	}
+	endpoints, err := tx.Get(tableConfigEntries, indexID+"_prefix", ConfigEntryKindQuery{Kind: structs.ServiceDefaults})
```
rboyer, line 3572: Isn't this `ConfigEntryKindQuery` missing an `EnterpriseMeta` field?

agent/consul/state/config_entry.go
```diff
@@ -376,12 +376,19 @@ func insertConfigEntryWithTxn(tx WriteTxn, idx uint64, conf structs.ConfigEntry)
...
 	// that associates gateways <-> services.
+
 	if conf.GetKind() == structs.TerminatingGateway || conf.GetKind() == structs.IngressGateway {
 		err := updateGatewayServices(tx, idx, conf, conf.GetEnterpriseMeta())
 		if err != nil {
 			return fmt.Errorf("failed to associate services to gateway: %v", err)
 		}
 	}
+	isEndpoint := conf.GetKind() == structs.ServiceDefaults && conf.(*structs.ServiceConfigEntry).Endpoint != nil
+	if isEndpoint {
```
rboyer, line 387: This blob might read better as either a `switch conf.GetKind()` or a bare `switch {` instead of two conditionals with two clauses each.

rboyer (changes requested):
I think there's a missing part of a memdb lookup.

api/config_entry_test.go
```diff
@@ -106,16 +106,16 @@ func TestAPI_ConfigEntries(t *testing.T) {
 			},
 		}
 
-		endpoint := &EndpointConfig{
+		endpoint := &DestinationConfig{
```
DanStough, line 109:
```suggestion
		dest := &DestinationConfig{
```

api/config_entry_test.go
```diff
@@ -106,16 +106,16 @@ func TestAPI_ConfigEntries(t *testing.T) {
...
 
 		service2 := &ServiceConfigEntry{
-			Kind:     ServiceDefaults,
-			Name:     "bar",
-			Protocol: "tcp",
-			Endpoint: endpoint,
+			Kind:        ServiceDefaults,
+			Name:        "bar",
+			Protocol:    "tcp",
+			Destination: endpoint,
```
DanStough, line 118:
```suggestion
			Destination: dest,
```

agent/consul/state/catalog.go
```diff
@@ -3491,13 +3503,36 @@ func terminatingConfigGatewayServices(
...
+	serviceIter, err := tx.Get(tableServices, indexService, Query{
+		Value:          name,
+		EnterpriseMeta: *entMeta,
+	})
+	if err != nil {
+		return structs.GatewayservicekindService, err
+	}
+	for service := serviceIter.Next(); service != nil; service = serviceIter.Next() {
+		return structs.GatewayservicekindUnknown, nil
+	}
```
DanStough, line 3524:
```suggestion
	if err != nil {
		return structs.GatewayservicekindUnknown, err
	}
	for service := serviceIter.Next(); service != nil; service = serviceIter.Next() {
		return structs.GatewayservicekindService, nil
	}
```

agent/consul/state/catalog.go
```diff
@@ -3538,6 +3573,41 @@ func updateGatewayNamespace(tx WriteTxn, idx uint64, service *structs.GatewaySer
 			return err
 		}
 	}
+	endpoints, err := tx.Get(tableConfigEntries, indexID+"_prefix", ConfigEntryKindQuery{Kind: structs.ServiceDefaults, EnterpriseMeta: *entMeta})
+	if err != nil {
+		return fmt.Errorf("failed querying endpoints: %s", err)
+	}
+	for endpoint := endpoints.Next(); endpoint != nil; endpoint = endpoints.Next() {
+		e := endpoint.(*structs.ServiceConfigEntry)
```
DanStough, line 3581:
```suggestion
	entries, err := tx.Get(tableConfigEntries, indexID+"_prefix", ConfigEntryKindQuery{Kind: structs.ServiceDefaults, EnterpriseMeta: *entMeta})
	if err != nil {
		return fmt.Errorf("failed querying entries: %s", err)
	}
	for entry := entries.Next(); entry != nil; entry = entries.Next() {
		e := entry.(*structs.ServiceConfigEntry)
```

agent/consul/state/catalog.go
```diff
@@ -3483,6 +3491,10 @@ func terminatingConfigGatewayServices(
 
 	var gatewayServices structs.GatewayServices
 	for _, svc := range entry.Services {
+		kind, err := GatewayServiceKind(tx, svc.Name, &svc.EnterpriseMeta)
+		if err != nil {
+			return false, nil, fmt.Errorf("failed to query endpoints: %v", err)
```
DanStough, line 3496:
```suggestion
			return false, nil, fmt.Errorf("failed to get gateway service kind for service %s: %v", svc.Name, err)
```

agent/consul/state/config_entry.go
```diff
@@ -344,6 +344,27 @@ func deleteConfigEntryTxn(tx WriteTxn, idx uint64, kind, name string, entMeta *a
...
+	switch c.GetKind() {
+	case structs.ServiceDefaults:
+		if c.(*structs.ServiceConfigEntry).Destination != nil {
+			gsKind, err := GatewayServiceKind(tx, sn.Name, &sn.EnterpriseMeta)
+			if err != nil {
+				return fmt.Errorf("failed updating gateway mapping: %s", err)
+			}
+			if gsKind == structs.GatewayservicekindDestination {
+				gsKind = structs.GatewayservicekindUnknown
+			}
```
DanStough, line 358: Feel free to ignore this one but I didn't understand this check. Maybe there is a case where `service-default` has a `destination` but is mislabeled? Then would it be:
```suggestion
			if gsKind == structs.GatewayservicekindUnknown {
				gsKind = structs.GatewayservicekindDestination
			}
```

DanStough (changes requested):

agent/consul/state/config_entry.go
dhiaayachi: So this is because the config entry is not yet deleted, so when we look it up we find it as a Destination type (potentially, if a service with the same name do not exist) so in that case we set it to unknown kind because we are deleting the config entry just after.

agent/structs/config_entry_gateways.go
```diff
@@ -635,6 +644,7 @@ func (g *GatewayService) IsSame(o *GatewayService) bool {
 		g.CertFile == o.CertFile &&
 		g.KeyFile == o.KeyFile &&
 		g.SNI == o.SNI &&
+		g.Kind == o.Kind &&
```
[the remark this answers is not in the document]
dhiaayachi: Yes right!

agent/consul/state/config_entry_test.go
```diff
@@ -310,6 +409,440 @@ func TestStore_ConfigEntries(t *testing.T) {
 		Protocol: "tcp",
 	}))
 	require.True(t, watchFired(ws))
+
+}
+
+func TestStore_ServiceDefaults_Kind_Destination(t *testing.T) {
```
DanStough, line 415: I should probably do an actual diff, but just looking at `TestStore_ServiceDefaults_isDestination` above and this function, I didn't see the difference.

agent/consul/state/config_entry_test.go
```diff
@@ -310,6 +409,440 @@ func TestStore_ConfigEntries(t *testing.T) {
...
+	require.True(t, watchFired(ws))
+
+	_, gatewayServices, err = s.GatewayServices(ws, "Gtwy1", nil)
+	require.NoError(t, err)
+	require.Len(t, gatewayServices, 1)
+	require.Equal(t, gatewayServices[0].Kind, structs.GatewayservicekindUnknown)
+
+}
+
+func TestStore_ServiceDefaults_Kind_Destination_Wildcard(t *testing.T) {
```
DanStough, line 671: Same comment here about redundancy with `TestStore_ServiceDefaults_isDestination_Wildcard`

DanStough (approved):
A couple non-blocking things

agent/consul/state/config_entry_test.go
dhiaayachi: right all the `isDestination` tests are duplicate. I will remove them

agent/structs/config_entry_gateways.go
```diff
@@ -586,19 +586,28 @@ func (e *TerminatingGatewayConfigEntry) Warnings() []string {
 	return warnings
 }
 
+type GatewayServiceKind string
+
+const (
+	GatewayservicekindUnknown     GatewayServiceKind = ""
+	GatewayservicekindDestination GatewayServiceKind = "gateway-service-destination"
+	GatewayservicekindService     GatewayServiceKind = "gateway-service-service"
```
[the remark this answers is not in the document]
rboyer: Also, why do these strings have the `gateway-service-` prefix on them if they are already a new unique kind? why not just `"service"` and `"destination"`?

agent/consul/state/catalog.go
```diff
@@ -3491,13 +3502,36 @@ func terminatingConfigGatewayServices(
...
+	if serviceIter != nil {
+		return structs.GatewayservicekindService, err
+	}
+
+	_, entry, err := configEntryTxn(tx, nil, structs.ServiceDefaults, name, entMeta)
+	if err != nil {
+		return structs.GatewayservicekindUnknown, err
+	}
+	if entry != nil {
+		return structs.GatewayservicekindDestination, nil
```
rboyer, line 3530: Shouldn't this deep-inspect the config entry to see if it's got destination-y bits set? It's valid to have no registered instances of a service and a `service-defaults` for something that _will_ be in the catalog.

dhiaayachi: yes right, good catch! Also that point me that I'm missing a test related a normal service-defaults.

agent/structs/config_entry_gateways.go
```diff
@@ -586,19 +586,28 @@ func (e *TerminatingGatewayConfigEntry) Warnings() []string {
...
-	FromWildcard bool     `json:",omitempty"`
+	Port         int                `json:",omitempty"`
+	Protocol     string             `json:",omitempty"`
+	Hosts        []string           `json:",omitempty"`
+	CAFile       string             `json:",omitempty"`
+	CertFile     string             `json:",omitempty"`
+	KeyFile      string             `json:",omitempty"`
+	SNI          string             `json:",omitempty"`
+	FromWildcard bool               `json:",omitempty"`
+	Kind         GatewayServiceKind `json:",omitempty"`
```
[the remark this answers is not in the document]
rboyer: I think it would be enough to call it anything else here, possibly even `ServiceKind GatewayServiceKind`

agent/consul/state/catalog.go
```diff
@@ -3491,13 +3502,42 @@ func terminatingConfigGatewayServices(
...
+		return structs.GatewayServiceKindUnknown, err
+	}
+	if entry != nil {
+		sd, ok := entry.(*structs.ServiceConfigEntry)
+		if !ok {
+			return structs.GatewayServiceKindUnknown, fmt.Errorf("invalid config entry type %T", entry)
+		}
+		if sd.Destination != nil {
+			return structs.GatewayServiceKindDestination, nil
+		}
```
rboyer, line 3536:
```suggestion
		}
		return structs.GatewayServiceKindService, nil
```
Would this be appropriate? Or are you elsewhere going to interpret "unknown" as "service"?

agent/consul/state/config_entry.go
```diff
@@ -344,6 +344,27 @@ func deleteConfigEntryTxn(tx WriteTxn, idx uint64, kind, name string, entMeta *a
 			return fmt.Errorf("failed updating gateway-services index: %v", err)
 		}
 	}
+
+	c := existing.(structs.ConfigEntry)
+	switch c.GetKind() {
```
rboyer, line 349: (trivial) could do this instead:
```
switch x := c.(type) {
case *structs.ServiceConfigEntry:
  if x.Destination != nil {
  <etc>
```

agent/consul/state/catalog.go
dhiaayachi: It will be dealt with in the watch receiving part https://github.com/hashicorp/consul/pull/13196
Once this PR got merged we will update https://github.com/hashicorp/consul/pull/13196 to have the right logic.

dhiaayachi: The reason I added the unknown kind is to differentiate between existing services/destinations and references in the tgtw to non existing, it will help for debugging purpose.

PR #13220, opened 2022-05-25
Backport of Update CHANGELOG to mention removal of Envoy 1.17.4 and 1.18.6 into release/1.12.x

hc-github-team-consul-core:
## Backport

This PR is auto-generated from #13207 to be assessed for backporting due to the inclusion of the label backport/1.12.

**WARNING** automatic cherry-pick of commits failed. Commits will require human attention.

---

### Description
For Consul 1.12.0, we had removed support for Envoy 1.17.4 and 1.18.6. Historically, we have considered these as "Improvements" but since we are removing support for some versions, it can also be seen as a "Breaking Change". 

Going forward, we want to call this out in our release notes as a "Breaking Change" so that users see the version compatibility change more prominently.


eculver:
This change got included in #13225.

PR #13221, opened 2022-05-25
CI: Verify built binaries in build job

huikang:
### Description

Smoke test the linux built binaries: 386, amd64, arm, and arm64

### limitation
Since qemu can't be installed in github action macOS runner, we can't smoke test the arm and arm64 binaries for darwin

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,6 @@ on:
     branches:
       # Push events on the main branch
       - main
-      - CTIA-8-verify-builds
 
 env:
   PKG_NAME: consul
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,7 @@ on:
     branches:
       # Push events on the main branch
       - main
+      - CTIA-8-verify-builds
 
 env:
   PKG_NAME: consul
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -6,6 +6,7 @@ on:
     branches:
       # Push events on the main branch
       - main
+      - CTIA-8-verify-builds
```
eculver, line 9: Assume we should remove this before merging.

.github/workflows/build.yml
```diff
@@ -232,6 +233,15 @@ jobs:
...
       - uses: actions/checkout@v2
+      - name: Download artifact for arm verification
+        uses: actions/download-artifact@v2
+        if: ${{ matrix.arch == 'arm64' || matrix.arch == 'arm' }}
+        with:
+          name: ${{ env.PKG_NAME }}_${{env.version}}_linux_${{ matrix.arch }}.zip
+      - name: Unzip for ${{ matrix.arch }} linux zip
+        if: ${{ matrix.arch == 'arm64' || matrix.arch == 'arm' }}
+        run: |
+          unzip ${{ env.PKG_NAME }}_${{env.version}}_linux_${{ matrix.arch }}.zip
```
eculver, line 244: small nit: I've been putting a space between the brackets. 
```suggestion
          unzip ${{ env.PKG_NAME }}_${{ env.version }}_linux_${{ matrix.arch }}.zip
```

.github/workflows/build.yml
```diff
@@ -244,3 +254,101 @@ jobs:
...
+      fail-fast: true
+    env:
+      version: ${{needs.get-product-version.outputs.product-version}}
+    name: Verify ${{ matrix.arch }} linux zip
+    steps:
+      - uses: actions/checkout@v2
+      - name: Download artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: ${{ env.PKG_NAME }}_${{env.version}}_linux_${{ matrix.arch }}.zip
```
eculver, line 285: same thing here.
```suggestion
          name: ${{ env.PKG_NAME }}_${{ env.version }}_linux_${{ matrix.arch }}.zip
```

eculver (approved):
Looks good, just a few small style things.

eculver:
One small bit we can do in a follow-up is compare the actual output of `consul version` with the output of the `get-product-version` action to make sure they match. I'm doing this in the docker smoke test PR [here](https://github.com/hashicorp/consul/pull/13200/files#diff-ae9082d0c5a60c4e5d7310e01594f4780a4e85cbee9aba1c26d49c95101675d0R5-R21).

.github/workflows/build.yml
huikang: Removed. this is the execution before the line is removed: https://github.com/hashicorp/consul/runs/6629909144?check_suite_focus=true

eculver:
We discussed this offline and we decided this is probably not a blocking change but adding here for posterity:

It would be great if the zip verification was done in a separate job rather than extending the docker build job. I actually made this change [in my branch](https://github.com/hashicorp/consul/commit/d8a4c0e24b21b1fad6f0a218314d3a68206d992f) so it's easier to just keep it there.

PR #13226, opened 2022-05-25
rename "endpoint" to "destination"

dhiaayachi:
### Description
This based on the choice to go with "destination" as name for the new external endpoints managed in the terminating gateway


rboyer (approved):
:building_construction: LGTM

PR #13227, opened 2022-05-25
remove a source of test panics

rboyer:
### Description

This test was updating the `delegate` value after the test agent had started. Firstly this is racy, and there's no good way to create an Agent first, update the delegate, and then start it because starting it is what initially populates the delegate.

Secondly (and relevant to this PR) this delegate didn't populate it's embedded `delegate` field so sometimes the test agent during the test would call a method on it and nil panic.

For now since the original test didn't feel terribly important at what it was testing, I just removed it completely.

### DIFF
--- a/agent/agent_endpoint_test.go
+++ b/agent/agent_endpoint_test.go
@@ -1703,6 +1703,7 @@ func TestAgent_Reload(t *testing.T) {
 	})
 
 	shim := &delegateConfigReloadShim{delegate: a.delegate}
+	// NOTE: this may require refactoring to remove a potential test panic
 	a.delegate = shim
 	if err := a.reloadConfigInternal(cfg2); err != nil {
 		t.Fatalf("got error %v want nil", err)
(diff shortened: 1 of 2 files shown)


agent/agent_endpoint_test.go
```diff
@@ -1703,6 +1703,7 @@ func TestAgent_Reload(t *testing.T) {
 	})
 
 	shim := &delegateConfigReloadShim{delegate: a.delegate}
+	// NOTE: this may require refactoring to remove a potential test panic
```
rboyer, line 1706:
```suggestion
	// NOTE: this may require refactoring to remove a potential test race
```

kisunji:
We'll have to backport the removals too from 1.12-1.10

kisunji:
cc @huikang I didn't catch this in my review but letting you know we're removing the test

kisunji (approved):

kisunji (approved):

PR #13244, opened 2022-05-26
Enable servers to configure arbitrary proxies from the catalog

boxofrad:
### Description
OSS port of enterprise PR 1822.

Includes the necessary changes to the `proxycfg` and `xds` packages to enable Consul servers to configure arbitrary proxies using catalog data.

Broadly, `proxycfg.Manager` now has public methods for registering, deregistering, and listing registered proxies — the existing local agent state-sync behavior has been moved into a separate component that makes use of these methods.

When an xDS session is started for a proxy service in the catalog, a goroutine will be spawned to watch the service in the server's state store and re-register it with the `proxycfg.Manager` whenever it is updated (and clean it up when the client goes away).

See [this gist](https://gist.github.com/boxofrad/43fe0e9554e431dabc8e6d90a3885ffb) 🤓


.changelog/13244.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+xds: Consul servers can now handle xDS streams for arbitrary proxy services in the catalog
```
boxofrad, line 2: @mkeeler I wasn't sure whether to hold off on the changelog entry until the feature is production-ready. What do you think?

mkeeler: I would either hold off on the changelog or switch this to be an `improvement` instead of `feature`. That way for people viewing the changelog don't see "new feature" and think its ready for general use.

mkeeler (approved):

.changelog/13244.txt
boxofrad: Yeah, let's hold off to avoid confusion 🙇🏻

Issue #13236, opened 2022-05-25
Consul 1.12.0 fails when putting to the kv store with a key containing a percent sign (%) 


DavidOhsieAtTurbo:
#### Overview of the Issue

On consul v1.20.0, an error is returned when trying to put to the key-value store to a key with a '%' (percent) character in it.  The server reports an error: "failure in unescaping path param".  It appears to attempting to percent decode twice which causes the failure.  This works fine in v1.11.5.   It is trivial to reproduce the problem as described below.  One merely needs to use the CLI to do a put with a key containing a '%' character.

#### Reproduction Steps

% ./consul agent -dev -bind 127.0.0.1 &
% ./consul -v
Consul v1.12.0
Revision 09a8cdb4
Protocol 2 spoken by default, understands 2 to 3 (agent will automatically use protocol >2 when speaking to compatible agents)

% ./consul kv put 'foo%foo' bar      
2022-05-25T17:25:56.398-0400 [ERROR] agent.http: Request error: method=PUT url=/v1/kv/foo%25foo from=127.0.0.1:51115 error="failure in unescaping path param "foo%foo": invalid URL escape "%fo""
2022-05-25T17:25:56.398-0400 [DEBUG] agent.http: Request finished: method=PUT url=/v1/kv/foo%25foo from=127.0.0.1:51115 latency=141.32µs
Error! Failed writing data: Unexpected response code: 500 (failure in unescaping path param "foo%foo": invalid URL escape "%fo")

This works fine in v1.11.5.

% ./consul agent -dev -bind 127.0.0.1 &
% ./consul-3 -v                  
Consul v1.11.5
Revision d8983fc9
Protocol 2 spoken by default, understands 2 to 3 (agent will automatically use protocol >2 when speaking to compatible agents)

% ./consul-3 kv put 'foo%foo' bar
2022-05-25T17:20:32.723-0400 [DEBUG] agent.http: Request finished: method=PUT url=/v1/kv/foo%25foo from=127.0.0.1:50501 latency=171.483µs
Success! Data written to: foo%foo

### Consul info for both Client and Server

<details>
  <summary>Client info</summary>

```
./consul info
2022-05-25T17:32:47.048-0400 [DEBUG] agent.http: Request finished: method=GET url=/v1/agent/self from=127.0.0.1:51939 latency=590.641µs
agent:
	check_monitors = 0
	check_ttls = 0
	checks = 0
	services = 0
build:
	prerelease = 
	revision = 09a8cdb4
[...]

PR #13256, opened 2022-05-26
Revert getPathSuffixUnescaped

kisunji:
Fixes https://github.com/hashicorp/consul/issues/13236
1.11.x: https://github.com/hashicorp/consul/pull/13265
1.10.x: https://github.com/hashicorp/consul/pull/13264

### Description
This PR reverts all of the changes made in https://github.com/hashicorp/consul/issues/11258 as it turns out that URLs are natively escaped and we had not tested the HTTP requests 🤦 

```
❯ consul agent -dev
...
❯ consul services register -name test/%
Registered service: test/%
```

Before PR
```
❯ curl --request GET localhost:8500/v1/agent/service/test%2F%25
failure in unescaping path param "test/%": invalid URL escape "%"%
```

With PR
```
❯ curl --request GET localhost:8500/v1/agent/service/test%2F%25
{
    "ID": "test/%",
    "Service": "test/%",
    "Tags": [],
    "Meta": {},
    "Port": 0,
    "Address": "",
    "Weights": {
        "Passing": 1,
        "Warning": 1
    },
    "EnableTagOverride": false,
    "ContentHash": "d31fe3644baf62d8",
    "Namespace": "default",
    "Partition": "default",
    "Datacenter": "dc1"
}
```


agent/catalog_endpoint.go
```diff
@@ -356,16 +357,8 @@ func (s *HTTPHandlers) catalogServiceNodes(resp http.ResponseWriter, req *http.R
 		args.TagFilter = true
 	}
 
-	if _, ok := params["merge-central-config"]; ok {
-		args.MergeCentralConfig = true
-	}
```
kyhavlov, line 361: Should this be getting reverted too? It doesn't seem related to the url escape changes

kisunji: oh good catch! Probably a victim of my deletes

agent/health_endpoint.go
```diff
@@ -203,10 +192,6 @@ func (s *HTTPHandlers) healthServiceNodes(resp http.ResponseWriter, req *http.Re
 		args.TagFilter = true
 	}
 
-	if _, ok := params["merge-central-config"]; ok {
```
jkirschner-hashicorp, line 206: @kisunji : same question as https://github.com/hashicorp/consul/pull/13256/files#r882931918 - is this delete supposed to be here?

jkirschner-hashicorp (commented):
@kisunji : found another place we seem to be removing the code that @kyhavlov spotted, and maybe shouldn't be removed.

kisunji:
Thanks for the thorough review @kyhavlov and @jkirschner-hashicorp! Turns out the merge-central-config changes didn't land in enterprise when I started this PR from the enterprise side and I overwrote the files in OSS

jkirschner-hashicorp:
@kyhavlov : looks good to me, but I defer to you (as an actual member of the eng team) to give the official approval.

DavidOhsieAtTurbo:
Just a general comment.  Would it be easier to simply redefine getPathSuffixUnescaped to not do the double escape rather than finding all the places where that is called and removing them?

kisunji:
> Just a general comment. Would it be easier to simply redefine getPathSuffixUnescaped to not do the double in escape rather than finding all the places where that is called and removing them?

`getPathSuffixUnescaped` was returning errors from the extra escaping, otherwise it was simply calling `strings.TrimPrefix` under the hood. I think it's cleaner to remove the abstraction (original code looked like this in the first place).

askmike1:
Is there an ETA for this getting released (I'm assuming as part of 1.12.3)?

jkirschner-hashicorp:
@askmike1 : yes, the current plan is for this to ship in 1.12.3. I'll keep this comment up-to-date: https://github.com/hashicorp/consul/issues/13236#issuecomment-1138893291

PR #13257, opened 2022-05-26
monitor leadership in peering service

acpana:

### Description
This diff adds a way for the peering service to monitor leadership changes on the consul server by looking at raft observations and recording the `leader address`. This will help us tell dialers what the address of the leader actually is.

* add server level testing in `server_test.go`

### Links
* [asana task](https://app.asana.com/0/1201832349181086/1202045399358214)

#### Other options considered

* poll approach: using [`monitorLeadership()`](https://github.com/hashicorp/consul/blob/f507f62f3cebd0ff9a4259d44487e75b2596492a/agent/consul/leader.go#L64) / adding a channel in peering service

-we would have to duplicate monitoring logic and have to add another go routine in the peering service, which can be more complex that we actually need

### DIFF
--- a/agent/consul/peering_backend.go
+++ b/agent/consul/peering_backend.go
@@ -129,8 +129,8 @@ func (m *leadershipMonitor) UpdateLeaderAddr(addr string) {
 }
 
 func (m *leadershipMonitor) GetLeaderAddr() string {
-	m.lock.Lock()
-	defer m.lock.Unlock()
+	m.lock.RLock()
+	defer m.lock.RUnlock()
 
 	return m.leaderAddr
 }
--- a/agent/consul/peering_backend.go
+++ b/agent/consul/peering_backend.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strconv"
+	"sync"
 
 	"google.golang.org/grpc"
 
@@ -19,6 +20,7 @@ type peeringBackend struct {
 	srv      *Server
 	connPool GRPCClientConner
 	apply    *peeringApply
+	monitor  *leadershipMonitor
 }
 
 var _ peering.Backend = (*peeringBackend)(nil)
@@ -29,6 +31,7 @@ func NewPeeringBackend(srv *Server, connPool GRPCClientConner) peering.Backend {
 		srv:      srv,
 		connPool: connPool,
 		apply:    &peeringApply{srv: srv},
+		monitor:  &leadershipMonitor{},
 	}
 }
 
@@ -101,12 +104,35 @@ func (b *peeringBackend) Apply() peering.Apply {
 	return b.apply
 }
 
+func (b *peeringBackend) LeadershipMonitor() peering.LeadershipMonitor {
+	return b.monitor
+}
+
 func (b *peeringBackend) EnterpriseCheckPartitions(partition string) error {
 	return b.enterpriseCheckPartitions(partition)
 }
 
 func (b *peeringBackend) IsLeader() bool {
 	return b.srv.IsLeader()
+}
+
+type leadershipMonitor struct {
+	lock       sync.RWMutex
+	leaderAddr string
+}
+
+func (m *leadershipMonitor) UpdateLeaderAddr(addr string) {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	m.leaderAddr = addr
+}
+
+func (m *leadershipMonitor) GetLeaderAddr() string {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	return m.leaderAddr
 }
 
 type peeringApply struct {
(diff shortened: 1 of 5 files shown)


agent/rpc/peering/stream_test.go
```diff
@@ -795,9 +795,28 @@ func TestStreamResources_Server_ServiceUpdates(t *testing.T) {
...
-	leader func() bool
+	pub               state.EventPublisher
+	store             *state.Store
+	leader            func() bool
+	leadershipMonitor *leadershipMonitor
+}
+
+var _ LeadershipMonitor = (*leadershipMonitor)(nil)
+
+type leadershipMonitor struct {
```
acpana, line 806: we could actually stub test the behavior of `trackLeadership` but then we would just test that a accessor/ mutator pair works under lock -- so there's not a lot of value there imo; that's why I wanted to test this in actuality at the consul server level

agent/rpc/peering/service.go
```diff
@@ -96,6 +96,20 @@ type Backend interface {
...
+// the server's leadership status
+// server addresses should look like: ip:port
+type LeadershipMonitor interface {
+	// UpdateLeaderAddr is called on a raft.LeaderObservation in a go routine in the consul server;
+	// see trackLeaderChanges()
+	UpdateLeaderAddr(leaderAddr string)
+
+	// GetLeaderAddr provides the best hint for the current address of the leader.
+	// There is no guarantee that this is the actual address of the leader.
+	GetLeaderAddr() string
```
acpana, line 112: We will call this method where we call `IsLeader()` and we are not the leader; we will give the best hint as to what the leader's address is.

agent/rpc/peering/service.go
```diff
@@ -96,6 +96,20 @@ type Backend interface {
 
 	Store() Store
 	Apply() Apply
+	LeadershipMonitor() LeadershipMonitor
+}
+
+// LeadershipMonitor provides a way for the consul server to update the peering service about
+// the server's leadership status
+// server addresses should look like: ip:port
```
freddygv, line 104:
```suggestion
// the server's leadership status.
// Server addresses should look like: ip:port
```

agent/consul/server.go
```diff
@@ -1655,6 +1655,8 @@ func (s *Server) trackLeaderChanges() {
 			}
 
 			s.grpcLeaderForwarder.UpdateLeaderAddr(s.config.Datacenter, string(leaderObs.LeaderAddr))
+			// TODO(peering): might need to send DC or use DC on the peering backend in the future
```
freddygv, line 1658: What do you mean by this? DC names are largely not relevant in the context of peering

freddygv (approved):
Looks good, nice progress

agent/consul/server_test.go
```diff
@@ -19,12 +19,12 @@ import (
...
 
-	"github.com/hashicorp/consul/agent/rpc/middleware"
-	"github.com/hashicorp/consul/ipaddr"
-
 	"github.com/hashicorp/go-uuid"
 	"golang.org/x/time/rate"
 
+	"github.com/hashicorp/consul/agent/rpc/middleware"
+	"github.com/hashicorp/consul/ipaddr"
+
```
kisunji, line 27: Can we put these in the right block below?

agent/consul/server.go
acpana: I was chatting with RB and maybe mesh gateways may need to know this so that's why I dropped this comment; I'm also open to removing it if it's causing confusion

agent/consul/peering_backend.go
```diff
@@ -109,6 +116,25 @@ func (b *peeringBackend) IsLeader() bool {
...
+func (m *leadershipMonitor) UpdateLeaderAddr(addr string) {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	m.leaderAddr = addr
+}
+
+func (m *leadershipMonitor) GetLeaderAddr() string {
+	m.lock.Lock()
+	defer m.lock.Unlock()
```
kisunji, line 133:
```suggestion
	m.lock.RLock()
	defer m.lock.RUnlock()
```

agent/consul/server.go
freddygv: I lean toward removing it, if it's needed we likely don't need this TODO as a reminder

kisunji (commented):
LeadershipMonitor feels misleading as a name; it doesn't "monitor" anything and is simply a field with locks. Perhaps "LeaderAddr" is more accurate with `LeaderAddr.Get()` `LeaderAddr.Set(addr)` methods?

agent/consul/server_test.go
acpana: not sure what happened here 🤔 , thanks for the call out!

acpana:
hey @kisunji thanks for your feedback ❤️ 

regarding naming, let me punt on that for now if it's okay. Since I will work on using the `LeaderAddr` from the receiving side next, I can fast follow up on the naming.

PR #13258, opened 2022-05-26
Add tag to resources provisioned during load test

huikang:
### Description
The PR addresses the following 3 issues in the current load test
1.  The current circleCI load-test job always checkouts the main branch, although the CI passes the commit of a PR:

https://github.com/hashicorp/consul/blob/fd7a403e1190ef7f8cfc35a13bf388945d94e37e/.circleci/config.yml#L980-L987
This makes it difficult to run load test against any PR or any change made to the `./test/load/terraform`.

2. The tag key `consul-servers` is too general and applied to both client and server agent, so the PR change the key name to `consul-ci-load-test`:
Screen Shot 2022-05-26 at 10 19 48 PM

3. `terraform destroy` may fail transiently; add a retry to make sure resources are always destroyed. 

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/test/load/terraform/variables.tf
+++ b/test/load/terraform/variables.tf
@@ -106,6 +106,5 @@ variable "consul_download_url" {
 variable "consul_version" {
   type        = string
   description = "Version of the Consul binary to install"
-  default     = "1.9.0"
+  default     = "1.12.0"
 }
-
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -983,6 +983,7 @@ jobs:
               LOCAL_COMMIT_SHA="<< pipeline.parameters.commit >>"
             fi
             echo "export LOCAL_COMMIT_SHA=${LOCAL_COMMIT_SHA}" >> $BASH_ENV
+            git checkout ${LOCAL_COMMIT_SHA}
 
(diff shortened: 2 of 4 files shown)


test/load/terraform/variables.tf
```diff
@@ -106,6 +106,5 @@ variable "consul_download_url" {
 variable "consul_version" {
   type        = string
   description = "Version of the Consul binary to install"
-  default     = "1.9.0"
+  default     = "1.12.0"
```
mkeeler, line 109: I don't actually think this gets used for CI triggered jobs.

Once the EC2 instances are running they execute the user-data scripts which will download a version of consul. 

https://github.com/hashicorp/consul/blob/2427e3883914f14777eb0077c0230e403a54037f/test/load/terraform/user-data-server.sh#L14-L18

If a download URL is provided  then we ignore the consul_version and just pull whatever URL we were given. Our circle ci paramters cause this to be the case:

https://github.com/hashicorp/consul/blob/2427e3883914f14777eb0077c0230e403a54037f/.circleci/config.yml#L991

mkeeler (approved):

test/load/terraform/variables.tf
huikang: Thanks for clarifying this!

PR #13267, opened 2022-05-26
chore(ci): add no-op jobs for branch protection rules

DanStough:
### Description
Right now branch protection rules for either UI or Go testing to pass cannot be applied Consul because CircleCI only selectively runs these checks depending on the branch name. This PR added `noop` jobs to each of the workflows so that each important check always appears in the PR, even if it does no meaningful work. The downside is that there will be a bit more noise in the CircleCi consul for each branch.

This PR and enabling the branch protections will make sure that tests pass on all of our automated PRs before they get merged in.

N/A

### PR Checklist

* [ ] ~updated test coverage~
* [X] not a security concern


DanStough:
> Does this need backporting too?

💯 . Thanks!

PR #13280, opened 2022-05-27
Refactor some functions for better enterprise use

kisunji:
Followup to https://github.com/hashicorp/consul/pull/12729

### DIFF
--- a/command/registry.go
+++ b/command/registry.go
@@ -1,6 +1,11 @@
 package command
 
 import (
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+
 	"github.com/hashicorp/consul/command/acl"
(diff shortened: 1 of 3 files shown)


command/registry.go
```diff
@@ -109,40 +114,16 @@ import (
...
-		m[ent.name] = func() (mcli.Command, error) {
-			return thisFn(ui)
-		}
-	}
-	return m
-}
-
-// CommandsFromRegistry returns a realized mapping of available CLI commands in a format that
-// the CLI class can consume. This should be called after all registration is
-// complete.
```
kisunji, line 143: This is no longer relevant since we don't register in `init()` any more; we declaratively register commands here

PR #13294, opened 2022-05-27
ci: Add package verifications to build workflow

eculver:
### Description
This is the last PR in the series to artifacts as part of the build workflow. The primary target for this last PR is for validating rpm and deb packages. There were TODOs for these last two jobs, which I'm adding. However, I'm also going back and making a few changes to the previous two PRs (#13200 and #13221):
- Verifications now all check the contents of the version string (as opposed to just running `consul version`)
- Renaming scripts to all match `verify_*.sh` pattern
- Moving zip/binary verification into its own job

The output can be seen in this run: 

### Links
- [CTIA-14](https://hashicorp.atlassian.net/browse/CTIA-14)
- [CTIA-17](https://hashicorp.atlassian.net/browse/CTIA-17)


.github/workflows/build.yml
```diff
@@ -283,90 +265,193 @@ jobs:
...
-        if: ${{ matrix.arch == 'amd64' }}
-        run: |
-          unzip ${{ env.PKG_NAME }}_${{env.version}}_darwin_${{ matrix.arch }}.zip
-          ./consul version
+          name: ${{ env.zip_name }}
+
+      - name: Unzip ${{ matrix.arch }} linux zip
+        run: unzip ${{ env.zip_name }}
+
+      - name: Run verification for ${{ matrix.arch }} linux binary
```
huikang, line 341: nit: darwin binary

.github/workflows/build.yml
```diff
@@ -283,90 +265,193 @@ jobs:
...
       fail-fast: true
+    env:
+      version: ${{ needs.get-product-version.outputs.product-version }}
+      deb_name: consul_${{ needs.get-product-version.outputs.product-version }}-*-${{ matrix.arch }}.deb
 
     name: Verify ${{ matrix.arch }} debian package
     steps:
       - uses: actions/checkout@v2
-
+
```
huikang, line 361: nit: remove the extra tab

.github/workflows/build.yml
```diff
@@ -283,90 +265,193 @@ jobs:
...
+      - name: Verify ${{ matrix.arch }} debian package (using Docker linux/arm/v7)
+        if: ${{ matrix.arch == 'arm' }}
         run: |
-          echo "TODO: run verification for ${{ matrix.arch }} debian package"
+          docker run \
+            --platform=linux/arm/v7 \
+            -v $(pwd):/workdir \
+            -w /workdir \
+            arm64v8/debian \
```
huikang, line 385: Should this be an `arm32v7/debian` image?

huikang (commented):
Some minor comments

.github/workflows/build.yml
```diff
@@ -6,6 +6,7 @@ on:
     branches:
       # Push events on the main branch
       - main
+      - CTIA-17-verify-linux-packages
```
huikang, line 9: Let's make sure this line is removed before merging

.github/workflows/build.yml
```diff
@@ -276,90 +258,230 @@ jobs:
...
         with:
-          name: ${{ env.PKG_NAME }}_${{env.version}}_darwin_${{ matrix.arch }}.zip
-      - name: Run verification for ${{ matrix.arch }} linux zip
-        if: ${{ matrix.arch == 'amd64' }}
-        run: |
-          unzip ${{ env.PKG_NAME }}_${{env.version}}_darwin_${{ matrix.arch }}.zip
-          ./consul version
+          name: ${{ env.zip_name }}
+
+      - name: Unzip ${{ matrix.arch }} darwin zip
```
huikang, line 331: Since verify-darwin workflow doesn't have the arch matrix, not sure if it will cause any issue in these steps.

huikang (approved):

.github/workflows/build.yml
eculver: Good call. I'll update it to use a static string.

PR #13304, opened 2022-05-31
Add isLeader metric to track if a server is a leader

### LINKED ISSUE #13169: Prometheus metric to show whether a server is a leader or follower
#### Feature Description

There should be a consul server prometheus metric showing whether a server considers itself to be a leader, follower (or neither, in the case of a candidate). This should be possible for the current time (i.e. as of the last prometheus scrape) or for an arbitrary point in the past. Currently there is no straightforward way to determine this.

Furthermore, there should be a metric dedicated to this purpose and stable in future versions of consul (rather than being a metric for some other purpose that indicates the leader as a side effect, and so is liable to change).

#### Use Case(s)

For normal operation of a consul server cluster, there should be exactly one leader server, and all other servers should be followers. It should be possible to monitor that these conditions are satisfied, and alert if not, by means of simple prometheus query expressions.

#### Non-solutions

At first glance, it looks like the `consul_raft_state_*` metrics offer this. But those are counters that increment upon entry to the relevant state. So their values at a point in time do not show the leader and followers. For example, if a server reports a non-zero value of `consul_raft_state_leader` that means it became leader at some point, but it does not tell you that it is the leader _now_.  (These counters do not even reliably tell the outcome of an election, as multiple elections may occur within a single prometheus scrape interval.)

In the past, there were gauge metrics that suggested the leader by their presence, for instance `consul_raft_apply` and `consul_autopilot_healthy`. But because those were only updated on the leader, when a server ceased to be leader they would contain stale values for a time controlled by the `telemetry.prometheus_retention_period` config setting. Furthermore, subsequent commits mean that those metrics no longer indicate the leader (https://github.com/hashicorp/consul/pull/9198 exposed `consul_raft_apply` on every node; https://github.com/hashicorp/consul/pull/12617 exposed `consul_autopilot_healthy` on every server).

While there are counter metrics that only increase on the leader, using them to reliably determine the leader requires a very cumbersome prometheus query expression (especially if the case of a standalone consul server is handled).

huikang:
### Description
Add "isLeader" to server metrics to track if a server is the leader of the cluster

### Links
fix #13169

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


acpana:
Also, let's make sure to remember to add a changelog entry for these changes

agent/consul/session_ttl.go
```diff
@@ -153,6 +157,11 @@ func (s *Server) updateMetrics() {
 
 			metrics.SetGauge([]string{"raft", "applied_index"}, float32(s.raft.AppliedIndex()))
 			metrics.SetGauge([]string{"raft", "last_index"}, float32(s.raft.LastIndex()))
+			if s.IsLeader() {
+				metrics.SetGauge([]string{"raft", "is_leader"}, float32(1))
+			} else {
+				metrics.SetGauge([]string{"raft", "is_leader"}, float32(0))
+			}
```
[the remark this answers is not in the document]
huikang: Sounds good. I'll move the metrics the leadership state change. Thanks.

agent/metrics_test.go
```diff
@@ -205,6 +205,29 @@ func TestAgent_OneTwelveRPCMetrics(t *testing.T) {
...
+		}
+		`
+
+		a := StartTestAgent(t, TestAgent{HCL: hcl})
+		defer a.Shutdown()
+
+		respRec := httptest.NewRecorder()
+		recordPromMetrics(t, a, respRec)
+
+		assertMetricExists(t, respRec, "is_leader")
```
acpana, line 227:
```suggestion
		assertMetricExistsWithValue(t, respRec, "is_leader", "1")
```

agent/metrics_test.go
```diff
@@ -205,6 +205,29 @@ func TestAgent_OneTwelveRPCMetrics(t *testing.T) {
...
+	})
```
acpana, line 228: could we also try to start a follower agent and assert that :

```
assertMetricExistsWithValue(t, respRec, "is_leader", "0")
```

agent/setup.go
```diff
@@ -200,6 +200,10 @@ func getPrometheusDefs(cfg lib.TelemetryConfig, isServer bool) ([]prometheus.Gau
 			Name: []string{"raft", "leader", "oldestLogAge"},
 			Help: "This measures how old the oldest log in the leader's log store is.",
 		},
+		{
+			Name: []string{"raft", "is_leader"},
```
acpana, line 204: ah, I had made a comment above about using a different gauge var for this maybe something like 

```
consulServerGauges := ...
```

test/integration/consul-container/metrics/leader_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	libcluster "github.com/hashicorp/consul/integration/consul-container/libs/cluster"
+	"github.com/hashicorp/consul/integration/consul-container/libs/node"
+	"github.com/hashicorp/consul/integration/consul-container/libs/utils"
+	"github.com/hashicorp/consul/lib/retry"
+)
+
+func TestLeaderShipMetrics(t *testing.T) {
```
acpana, line 21: nit

```suggestion
func TestLeadershipMetrics(t *testing.T) {
```

test/integration/consul-container/metrics/leader_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	libcluster "github.com/hashicorp/consul/integration/consul-container/libs/cluster"
+	"github.com/hashicorp/consul/integration/consul-container/libs/node"
+	"github.com/hashicorp/consul/integration/consul-container/libs/utils"
+	"github.com/hashicorp/consul/lib/retry"
+)
+
```
acpana, line 20: Could you kindly add a small comment about what this test is meant to do? You could use the `given, when, then` framework if you'd like.

I think, especially for integ tests, this will help maintainability down the road.

test/integration/consul-container/upgrade/version_oss.go
```diff
@@ -1,9 +0,0 @@
-//go:build !consulent
```
acpana, line 1: Oh I see you are moving these under a `utils` package -- I think that's a great idea!

For this kind of refactor, you will need to make an "extra" commit to fix ENT specific changes since it will break when it carries over to our ENT repo (you can browse the repo to see that there's ent specific functions there).

So, I think this is fine to do, just something to keep in mind as you get to merge this in that  we will need to fix in ENT.

Alternatively, you could factor out these changes (the refactor into `utils`) and follow up in a subsequent PR from the ENT side. Refactoring is usually easier from that side.

test/integration/consul-container/metrics/leader_test.go
huikang: Done

agent/metrics_test.go
huikang: @acpana , thanks for the suggestions. To get the correct measurement, I think we need to wait until the metrics is updated. So could you kindly point me to some place where I can update the `MetricsReportingInterval` of the TestAgent? Thanks.

acpana: it seems like you plumbed it thru 💯

.changelog/13304.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+telemetry: Added a `is_leader` metric to track if a server is a leader or not.
```
acpana, line 2: let's use the full name:

```suggestion
telemetry: Added a `consul.server.is_leader` metric to track if a server is a leader or not.
```

acpana: Also, I'm sorry I didn't think about it before, but we probably don't want to use `_` in the name as `_` can be a special character (for instance, for prometheus metrics). We have some metrics today that have `_` in their name but I think they are in the minority.

https://www.consul.io/docs/agent/telemetry

Let's do Carmel case: `isLeader` -- what do you think?

website/content/docs/agent/telemetry.mdx
```diff
@@ -85,6 +85,7 @@ These are some metrics emitted that can help you understand the health of your c
 | `consul.raft.leader.lastContact` | Measures the time since the leader was last able to contact the follower nodes when checking its leader lease. | ms        | timer   |
 | `consul.raft.state.candidate`    | Increments whenever a Consul server starts an election.                                                   | elections | counter |
 | `consul.raft.state.leader`       | Increments whenever a Consul server becomes a leader.                                                     | leaders   | counter |
+| `consul.server.is_leader`        | Track if a server is a leader.                                                                            | 1 or 0    | gauge   |
```
acpana, line 88: thanks for adding docs!

This metrics seems pretty important and convenient. 

I wonder if there is value in having a section dedicated to it even before:

* https://www.consul.io/docs/agent/telemetry#transaction-timing

You could also do the docs in a follow up PR

acpana (commented):
This is great! Let's change `is_leader` to something like `isLeader` please but otherwise LGTM!

.changelog/13304.txt
huikang: Sure, I can make the change along with the other changes.

acpana (approved):
🚀

PR #13314, opened 2022-05-31
Fix broken CI step

kisunji:
Follow up to https://github.com/hashicorp/consul/pull/10996

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -687,7 +687,7 @@ jobs:
           name: move compiled ui files to agent/uiserver
           command: |
             rm -rf agent/uiserver/dist
-            mv ui/packages/consul-ui/dist agent/uiserver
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -687,7 +687,7 @@ jobs:
           name: move compiled ui files to agent/uiserver
           command: |
             rm -rf agent/uiserver/dist
-            mv ui/packages/consul-ui/dist agent/uiserver
+            mv packages/consul-ui/dist agent/uiserver
```
kisunji, line 690: The glob `packages/consul-ui/dist` is persisted to workspace root, not `ui/`. So once we use `attach_workspace.at: .` the directory will actually be at the root of the workspace

rboyer (approved):

PR #13321, opened 2022-06-01
Configure upstream TLS context with peer root certs

freddygv:
### Description

For mTLS to work between two proxies in peered clusters with different root CAs,
proxies need to configure their outbound listener to use different root certificates
for validation. 

Up until peering was introduced proxies would only ever use one set of root certificates
to validate all mesh traffic, both inbound and outbound. Now an upstream proxy
may have a leaf certificate signed by a CA that's different from the dialing proxy's.

This PR makes changes to proxycfg and xds so that the upstream TLS validation
uses different root certificates depending on which cluster is being dialed.

#### proxycfg
When an explicit upstream is in a peered cluster, we watch the trust bundle
for that cluster. These trust bundles then get stored in a map keyed on peer name. 

#### xds
When generating upstream clusters for discovery chains we use the RootPEMs
from the appropriate peer trust bundle if the upstream is in another cluster.

These changes have unit tests, but have not been validated against running Envoy proxies yet. That will happen after this PR and the public listener changes have been merged.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/proxycfg/connect_proxy.go
```diff
@@ -193,6 +195,20 @@ func (s *handlerConnectProxy) initialize(ctx context.Context) (ConfigSnapshot, e
...
+				// Check whether a watch for this peer exists to avoid duplicates.
+				if _, ok := snap.ConnectProxy.WatchedUpstreamTrustBundles[uid.Peer]; !ok {
+					if err := s.cache.Notify(ctx, cachetype.TrustBundleReadName, &pbpeering.TrustBundleReadRequest{
+						Name:      uid.Peer,
+						Partition: uid.PartitionOrDefault(),
+					}, peerTrustBundleWatchID+uid.Peer, s.ch); err != nil {
+						return snap, fmt.Errorf("error while watching trust bundle for peer %q: %w", uid.Peer, err)
+					}
+
+					// Assigning a nil entry lets us track that the watch was initiated, even if the trust bundle hasn't
```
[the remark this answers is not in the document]
rboyer: If this is something that needs to be canceled selectively in the future, wouldn't you need a per-peer cancelFunc? If so, then you could just use the map for that as your "did i make a watch yet?" check.

agent/proxycfg/state.go
```diff
@@ -38,6 +38,7 @@ const (
 	coalesceTimeout                    = 200 * time.Millisecond
 	rootsWatchID                       = "roots"
 	leafWatchID                        = "leaf"
+	peerTrustBundleWatchID             = "peer-trust-bundle:"
```
rboyer, line 41: Since this is a prefix the other prefix style names would suggest including "prefix" in the var name somewhere like `peerTrustBundleIDPrefix`

agent/xds/clusters.go
```diff
@@ -94,6 +93,11 @@ func (s *ResourceGenerator) clustersFromSnapshotConnectProxy(cfgSnap *proxycfg.C
 			return nil, fmt.Errorf("no endpoint map for upstream %q", uid)
 		}
 
+		if _, ok := cfgSnap.ConnectProxy.WatchedUpstreamTrustBundles[uid.Peer]; uid.Peer != "" && !ok {
+			// The trust bundle for this upstream is not available yet, skip for now.
+			continue
+		}
```
rboyer, line 99: This conditional skip needs to go into the EDS loop as well to keep clusters and endpoints in sync.

agent/cache-types/trust_bundle_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+// NewMockTrustBundleReader creates a new instance of MockTrustBundleReader. It also registers the testing.TB interface on the mock and a cleanup function to assert the mocks expectations.
+func NewMockTrustBundleReader(t testing.TB) *MockTrustBundleReader {
+	mock := &MockTrustBundleReader{}
+	mock.Mock.Test(t)
+
+	t.Cleanup(func() { mock.AssertExpectations(t) })
+
+	return mock
+}
```
kisunji, line 150: Would be nice to keep auto-generated types/methods in a different file with the commands to regenerate if necessary (I like `//go:generate mockery blah...` at the top of the file cause Goland can run it)

agent/proxycfg/connect_proxy.go
```diff
@@ -231,6 +247,17 @@ func (s *handlerConnectProxy) handleUpdate(ctx context.Context, u UpdateEvent, s
 			return fmt.Errorf("invalid type for response: %T", u.Result)
 		}
 		snap.Roots = roots
+
+	case strings.HasPrefix(u.CorrelationID, peerTrustBundleWatchID):
```
kisunji, line 251: Just a general comment: would be nice if we could change "prefix-y" names to something like `peerTrustBundleWatchIDPrefix` since it would hint to a developer that you need to add something to it (otherwise if I was mucking around with watches I might not know at a glance that `peerTrustBundleWatchID` isn't a standalone correlationID).

agent/proxycfg/snapshot.go
```diff
@@ -532,6 +539,15 @@ func (s *ConfigSnapshot) Leaf() *structs.IssuedCert {
 	}
 }
 
+// RootPEMs returns all PEM-encoded public certificates for the root CA.
+func (s *ConfigSnapshot) RootPEMs() string {
+	var rootPEMs string
+	for _, root := range s.Roots.Roots {
+		rootPEMs += ca.EnsureTrailingNewline(root.RootCert)
+	}
+	return rootPEMs
```
kisunji, line 548: nit: could use `strings.Builder` here to save memory in case there's a non-trivial number of roots

agent/proxycfg/testing_peering.go
```diff
@@ -1,6 +1,7 @@
 package proxycfg
 
 import (
+	"github.com/hashicorp/consul/proto/pbpeering"
```
kisunji, line 4: nit: wrong import group

agent/proxycfg/connect_proxy.go
```diff
@@ -231,6 +247,17 @@ func (s *handlerConnectProxy) handleUpdate(ctx context.Context, u UpdateEvent, s
...
+		resp, ok := u.Result.(*pbpeering.TrustBundleReadResponse)
+		if !ok {
+			return fmt.Errorf("invalid type for response: %T", u.Result)
+		}
+		peer := strings.TrimPrefix(u.CorrelationID, peerTrustBundleWatchID)
+		if resp.Bundle != nil {
+			snap.ConnectProxy.WatchedUpstreamTrustBundles[peer] = resp.Bundle
+		}
```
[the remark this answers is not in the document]
freddygv: Right, because the output isn't usable

agent/proxycfg/connect_proxy.go
```diff
@@ -193,6 +196,20 @@ func (s *handlerConnectProxy) initialize(ctx context.Context) (ConfigSnapshot, e
 				if err := (*handlerUpstreams)(s).resetWatchesFromChain(ctx, uid, chain, &snap.ConnectProxy.ConfigSnapshotUpstreams); err != nil {
 					return snap, fmt.Errorf("error while resetting watches from chain: %w", err)
 				}
+
+				// Check whether a watch for this peer exists to avoid duplicates.
+				if _, ok := snap.ConnectProxy.WatchedPeerTrustBundles[uid.Peer]; !ok {
```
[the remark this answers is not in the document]
freddygv: That would be a tproxy concern. Will add it to Asana.

agent/proxycfg/snapshot.go
```diff
@@ -532,6 +544,15 @@ func (s *ConfigSnapshot) Leaf() *structs.IssuedCert {
...
+}
+
```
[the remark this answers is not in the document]
rboyer: Disregard. The CDS/EDS skip logic handles this already.

proto/pbpeering/peering.go
```diff
@@ -88,6 +93,54 @@ func (x ReplicationMessage_Response_Operation) GoString() string {
...
+	if b == nil {
+		return ""
+	}
+
+	var rootPEMs string
+	for _, pem := range b.RootPEMs {
+		rootPEMs += pem
+
+		// We do not use the "ca.EnsureTrailingNewline" helper here because importing "connect/ca"
+		// would lead to an import cycle.
```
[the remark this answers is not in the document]
freddygv: Moved the function to `lib`. 

The cycle was:

`pbpeering -> connect/ca -> agent/consul/state`

rboyer (approved):

PR #13337, opened 2022-06-01
[OSS] Pull split ns/partition var out of testing file

freddygv:
### Description
The api module previously had defaultPartition and defaultNamespace vars
for when we need default/empty split usage between ent/oss respectively.

This commit moves those two variables out of test code so that they can
be used for the service exports config entry's `GetNamespace()` method.

Previously `GetNamespace()` would return "default" in both OSS and enterprise,
which can trip up automation that passes the result of this method as the
namespace to write a config entry.

The split vars are kept private to prevent external usage, and prefixed with
`split` for more clarity about their behavior.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


dhiaayachi (approved):
🚢

PR #13341, opened 2022-06-02
Egress gtw/connect destination intentions

dhiaayachi:
### Description 
- Add the ability to filter the intention topology by an `IntentionDestinationType`
- Add a new ServiceNameOfKind Kind which will represent Destinations

This will be later used to add a new Watch which will be getting the Intention topology for Destinations which return a Destination List, each destination in this list will be watched in the connect proxy proxy config to generate the right envoy config for each destination on the connect proxy side.


agent/consul/state/intention.go
```diff
@@ -1017,6 +1008,16 @@ func (s *Store) intentionTopologyTxn(tx ReadTxn, ws memdb.WatchSet,
 			maxIdx = index
 		}
 		services = append(services, ingress...)
+	} else {
+		// destinations can only ever be upstream, since they are only allowed as intention destination.
+		index, destinations, err := serviceNamesOfKindTxn(tx, ws, structs.ServiceKindDestination, *wildcardMeta)
+		if err != nil {
+			return index, nil, fmt.Errorf("failed to list ingress service names: %v", err)
```
DanStough, line 1015:
```suggestion
			return index, nil, fmt.Errorf("failed to list destination service names: %v", err)
```

agent/consul/state/config_entry_intention.go
```diff
@@ -294,8 +289,27 @@ func readDestinationIntentionsFromConfigEntriesTxn(tx ReadTxn, ws memdb.WatchSet
...
+			if entry != nil {
+				switch destinationType {
+				case structs.IntentionDestinationService:
+					if kind == structs.GatewayServiceKindService || kind == structs.GatewayServiceKindUnknown {
+						results = append(results, entry.ToIntentions()...)
+					}
+				case structs.IntentionDestinationDestination:
+					if kind == structs.GatewayServiceKindDestination {
+						results = append(results, entry.ToIntentions()...)
+					}
```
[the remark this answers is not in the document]
dhiaayachi: I'm not sure we can collapse both in the same `case` as it will make the if statement a bit complex.

agent/consul/state/config_entry.go
```diff
@@ -356,12 +356,16 @@ func deleteConfigEntryTxn(tx WriteTxn, idx uint64, kind, name string, entMeta *a
 			if gsKind == structs.GatewayServiceKindDestination {
 				gsKind = structs.GatewayServiceKindUnknown
 			}
-			if err := checkGatewayWildcardsAndUpdate(tx, idx, &structs.ServiceName{Name: c.GetName(), EnterpriseMeta: *c.GetEnterpriseMeta()}, gsKind); err != nil {
+			serviceName := structs.NewServiceName(c.GetName(), c.GetEnterpriseMeta())
+			if err := checkGatewayWildcardsAndUpdate(tx, idx, &serviceName, gsKind); err != nil {
```
acpana, line 360: hm, is this just still preference?

agent/consul/state/config_entry.go
```diff
@@ -356,12 +356,16 @@ func deleteConfigEntryTxn(tx WriteTxn, idx uint64, kind, name string, entMeta *a
...
 				return fmt.Errorf("failed updating gateway mapping: %s", err)
 			}
-			if err := checkGatewayAndUpdate(tx, idx, &structs.ServiceName{Name: c.GetName(), EnterpriseMeta: *c.GetEnterpriseMeta()}, gsKind); err != nil {
+			if err := checkGatewayAndUpdate(tx, idx, &serviceName, gsKind); err != nil {
 				return fmt.Errorf("failed updating gateway mapping: %s", err)
 			}
+			if err := cleanupKindServiceName(tx, idx, serviceName, structs.ServiceKindDestination); err != nil {
+				return fmt.Errorf("failed to persist service name: %v", err)
```
acpana, line 367: I know the other errors don't have it, but could we say:
```suggestion
				return fmt.Errorf("failed to persist service name: \"%s\"; err: %v", serviceName, err)
```

agent/consul/state/config_entry_intention.go
```diff
@@ -292,13 +302,15 @@ func readDestinationIntentionsFromConfigEntriesTxn(tx ReadTxn, ws memdb.WatchSet
 	names := getIntentionPrecedenceMatchServiceNames(serviceName, entMeta)
 	for _, sn := range names {
 		_, entry, err := getServiceIntentionsConfigEntryTxn(tx, ws, sn.Name, nil, &sn.EnterpriseMeta)
+		if err != nil {
+			return 0, nil, err
+		}
 		if err != nil {
 			return 0, nil, err
 		} else if entry != nil {
```
acpana, line 310: Is this an accidental dupe?

```suggestion
		if err != nil {
			return 0, nil, err
		} else if entry != nil {
```

agent/consul/state/intention.go
```diff
@@ -838,23 +839,14 @@ func (s *Store) legacyIntentionMatchTxn(tx ReadTxn, ws memdb.WatchSet, args *str
 //
 // The returned intentions are sorted based on the intention precedence rules.
 // i.e. result[0] is the highest precedent rule to match
-func (s *Store) IntentionMatchOne(
-	ws memdb.WatchSet,
-	entry structs.IntentionMatchEntry,
-	matchType structs.IntentionMatchType,
-) (uint64, structs.Intentions, error) {
+func (s *Store) IntentionMatchOne(ws memdb.WatchSet, entry structs.IntentionMatchEntry, matchType structs.IntentionMatchType, destinationType structs.IntentionTargetType) (uint64, structs.Intentions, error) {
```
acpana, line 842: I'm a bit partial to the other style but that's just personal preference 🤷🏼 

do we have a team stance here?

acpana (commented):
Thanks for giving me a chance to review this folks!

I could only do a surface level/ level 1 check now. Otherwise, things LGTM.

Only curiosity is if there are any backwards compat concerns -- but I see the `update` integ tests work so this should be ok 💯

agent/consul/state/config_entry.go
```diff
@@ -356,12 +356,16 @@ func deleteConfigEntryTxn(tx WriteTxn, idx uint64, kind, name string, entMeta *a
 			if gsKind == structs.GatewayServiceKindDestination {
 				gsKind = structs.GatewayServiceKindUnknown
 			}
-			if err := checkGatewayWildcardsAndUpdate(tx, idx, &structs.ServiceName{Name: c.GetName(), EnterpriseMeta: *c.GetEnterpriseMeta()}, gsKind); err != nil {
+			serviceName := structs.ServiceName{Name: c.GetName(), EnterpriseMeta: *c.GetEnterpriseMeta()}
```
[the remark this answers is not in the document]
dhiaayachi: No real reason, fixed it

agent/consul/state/intention.go
dhiaayachi: I'm not sure I understand what you mean by "the other style"?

agent/consul/state/config_entry.go
dhiaayachi: I'm not sure I understand the question?

acpana (approved):
Thanks for following up on the feedback 🥇 

This LGTM but I will defer to Dan for more domain specific feedback.

DanStough (approved):
A few non-blocking things. 🚀

PR #13342, opened 2022-06-02
Ffmmm/peering send leader addr

acpana:
#### patch contents

* as a new `ReplicationMessage` type of `LeaderAddress` 
* follows up on feedback from past PR to rename `LeadershipMonitor` to `LeaderAddress`. [ref](https://github.com/hashicorp/consul/pull/13257#pullrequestreview-986664903)

### Description
If as the receiver of a bidistream we are on a follower consul node, It would be more *efficient* if we told the dialer who we think the leader is by sending a message with the leader address.

* `stream_test.go` updated

### Links
Asana task [ref](https://app.asana.com/0/1201832349181086/1202294390627575)


agent/rpc/peering/service.go
```diff
@@ -490,9 +491,14 @@ func (s *Service) StreamResources(stream pbpeering.PeeringService_StreamResource
 	if !s.Backend.IsLeader() {
 		// we are not the leader so we will hang up on the dialer
 
-		// TODO(peering): in the future we want to indicate the address of the leader server as a message to the dialer (best effort, non blocking)
 		s.logger.Error("cannot establish a peering stream on a follower node")
-		return grpcstatus.Error(codes.FailedPrecondition, "cannot establish a peering stream on a follower node")
+		var merr *multierror.Error
```
acpana, line 495: not really sure if we want to use `merr` here.

we could just best effort log `sendLeaderAddrMsg` and drop the error on the floor; that kinda feels wrong.

what I don't like is that now the return error is not a `grpcstatus.Error` 🤔 

wdyt? remove?

agent/rpc/peering/stream_test.go
```diff
@@ -63,10 +63,25 @@ func TestStreamResources_Server_Follower(t *testing.T) {
...
-	require.EqualError(t, err, "rpc error: code = FailedPrecondition desc = cannot establish a peering stream on a follower node")
+	require.NoError(t, err)
+	require.NotEmpty(t, msg)
+	prototest.AssertDeepEqual(t, expectedMsg, msg)
+
+	// expect error
+	msg2, err2 := client.Recv()
+	require.Nil(t, msg2)
+	require.Error(t, err2)
+	require.Contains(t, err2.Error(), "rpc error: code = FailedPrecondition desc = cannot establish a peering stream on a follower node")
```
acpana, line 84: This was a little un-intuitive to me.

As a dialer, we receive the `LeaderAddress` message then the error as a subsequent "message". The more I started at the code (the mock `Recv()` code that is) the more I convinced myself that this is emulating real world scenario and is in fact what we want but please keep me honest if I got it wrong here.

agent/rpc/peering/service.go
```diff
@@ -677,9 +683,14 @@ func (s *Service) HandleStream(req HandleStreamRequest) error {
...
-				// TODO(peering): in the future we want to indicate the address of the leader server as a message to the dialer (best effort, non blocking)
 				logger.Error("node is not a leader anymore; cannot continue streaming")
-				return grpcstatus.Error(codes.FailedPrecondition, "node is not a leader anymore; cannot continue streaming")
+				var merr *multierror.Error
+				if err := s.sendLeaderAddrMsg(req.Stream, status, logger); err != nil {
+					merr = multierror.Append(merr, err)
+				}
+
+				merr = multierror.Append(grpcstatus.Error(codes.FailedPrecondition, "node is not a leader anymore; cannot continue streaming"), merr)
+				return merr.ErrorOrNil()
```
kisunji, line 693: I think this implementation right now can be streamlined by either treating "is not leader" state as:
1. A regular replication message (similar to `Terminated`) and don't return errors but assume client will handle the LeaderAddress message gracefully; or
2. An error state where we send a `grpcstatus.Error` with `grpcstatus.Status` [holding Details about the error](https://cloud.google.com/apis/design/errors#error_model)

Doing both feels redundant and prone to more bugs since the client needs to handle 2 messages instead of one

agent/consul/peering_backend.go
```diff
@@ -104,7 +104,7 @@ func (b *peeringBackend) Apply() peering.Apply {
 	return b.apply
 }
 
-func (b *peeringBackend) LeadershipMonitor() peering.LeadershipMonitor {
+func (b *peeringBackend) LeaderAddress() peering.LeaderAddress {
 	return b.monitor
```
kisunji, line 108: I think you missed this line? `peeringBackend` still has references to `leadershipMonitor`

acpana: ah yes, thanks for that call out 💯

proto/pbpeering/peering.proto
```diff
@@ -362,4 +363,11 @@ message ReplicationMessage {
   // Terminated is sent when a peering is deleted locally.
   // This message signals to the peer that they should clean up their local state about the peering.
   message Terminated {}
+
+  // LeaderAddress is sent when the peering service runs on a consul node
+  // that is not a leader. The node either was a leader before and not it is not, or it never was.
```
kisunji, line 368:
```suggestion
  // that is not a leader. The node either lost leadership, or never was a leader.
```

agent/rpc/peering/service.go
acpana: 👍🏼 I'm with you there. Doing both kinda seems wrong.

I've been leaning towards `2` -- packing things into the error itself.  To me, sending an error to the dialer indicates that something is wrong and it needs to handle that.

do you (or others) have a preference?

kisunji: I don't think we've really set a precedence for `2` yet but it might be worth ironing things out. The downside is we'll have to add intricate error-handling logic on the receiving side. Any thoughts @freddygv @rboyer

agent/rpc/peering/service.go
rboyer: I think it might be better to only log the `sendLeaderAddrMsg` error and not try to reply with it with multierror. This is purely an opportunistic kind of attempt to send some data, so if it doesn't work it's not like the other peer server can do anything with this error since it's going to get the OTHER error at the same time and do the same thing regardless.

agent/rpc/peering/service.go
rboyer: I like either way. Once we get server-list replication going, we might end up refactoring this anyway. At that point it would be good to replicate the list of server addresses with an annotation that says which one is currently leader.

If we had that, then what this "last send before error" would likely be better off doing is just re-issuing one of those normal replication messages instead, and then maybe issue a gRPC plain error here and let the receiving side use the replicated data to update its own choice of servers to contact.

acpana: sounds like we are going with the `2`/ `error` approach -- I will update the PR with that 🤝

acpana: We should have the leader address baked into the error details now.

agent/rpc/peering/stream_test.go
acpana: changed approach, irrelevant now

agent/rpc/peering/service.go
acpana: we changed approach so now there's no more `sendLeaderAddrMsg`

agent/consul/peering_backend.go
```diff
@@ -20,7 +20,7 @@ type peeringBackend struct {
 	srv      *Server
 	connPool GRPCClientConner
 	apply    *peeringApply
-	monitor  *leadershipMonitor
+	monitor  *leaderAddr
```
kisunji, line 23: this is still called `monitor` 😆

agent/rpc/peering/service.go
```diff
@@ -677,9 +693,22 @@ func (s *Service) HandleStream(req HandleStreamRequest) error {
...
+					Address: s.Backend.LeaderAddress().Get(),
+				})
+				if err != nil {
+					s.logger.Error(fmt.Sprintf("failed to marshal the leader address in response; err: %v", err))
+				}
+
+				// if there was an error above while marshaling details, then details is nil
+				// clients need to check types casts either way so it should be safe to send {nil}
+				return grpcstatus.FromProto(&spb.Status{
+					Code:    9,
```
acpana, line 708: tried avoiding magic number

agent/rpc/peering/stream_test.go
```diff
@@ -63,10 +66,32 @@ func TestStreamResources_Server_Follower(t *testing.T) {
...
 	require.EqualError(t, err, "rpc error: code = FailedPrecondition desc = cannot establish a peering stream on a follower node")
+
+	// expect a status error
+	st, ok := status.FromError(err)
+	require.True(t, ok, "need to get back a grpc status error")
+	deets := st.Details()
+	// expect a ReplicationMessage_LeaderAddress
+	{
+		found := false
+		for _, deet := range deets {
```
acpana, line 82: this is pretty good approximation of how the dialer will look for a leader address if we don't tag the server addresses

agent/rpc/peering/service.go
```diff
@@ -492,13 +494,21 @@ func (s *Service) StreamResources(stream pbpeering.PeeringService_StreamResource
...
 
-		merr = multierror.Append(grpcstatus.Error(codes.FailedPrecondition, "cannot establish a peering stream on a follower node"), merr)
-		return merr.ErrorOrNil()
+		// if there was an error above while marshaling details, then details is nil
+		// clients need to check types casts either way so it should be safe to send {nil}
+		return grpcstatus.FromProto(&spb.Status{
+			Code:    9,
+			Message: "cannot establish a peering stream on a follower node",
+			Details: []*apb.Any{details},
+		}).Err()
```
kisunji, line 494: Just realized there's a cleaner way of doing this that wasn't obvious from godocs because of this [type alias](https://pkg.go.dev/google.golang.org/grpc@v1.47.0/status#Status). The useful functions were hidden in the godocs of the internal/status.Status page.

```go
status.New(codes.FailedPrecondition, "cannot establish a peering stream on a follower node").WithDetails(&pbpeering.LeaderAddress{
	Address: s.Backend.LeaderAddress().Get(),
})
```

agent/rpc/peering/stream_test.go
```diff
@@ -63,10 +66,32 @@ func TestStreamResources_Server_Follower(t *testing.T) {
...
+			if !ok {
+				continue
+			} else {
+				found = true
+				require.Equal(t, "expected:address", la.Address)
+			}
+		}
+
+		require.True(t, found, "expected to find a LeaderAddress message in error details")
+	}
```
kisunji, line 94: Can we simplify this with `require.ElementsMatch`? Unsure if it works but this is quite verbose for what should be a simple slice assertion

agent/rpc/peering/stream_test.go
```diff
@@ -147,10 +175,32 @@ func TestStreamResources_Server_LeaderBecomesFollower(t *testing.T) {
...
 }
```
kisunji, line 204: Same as above - would be nice to give an expected slice of elements and compare rather than this verbose loop

agent/rpc/peering/service.go
```diff
@@ -11,12 +11,15 @@ import (
 	"github.com/armon/go-metrics"
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
+	apb "github.com/golang/protobuf/ptypes/any"
```
rboyer, line 14: Why does this use a custom import name?

agent/rpc/peering/service.go
```diff
@@ -11,12 +11,15 @@ import (
...
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-memdb"
 	"google.golang.org/genproto/googleapis/rpc/code"
+	spb "google.golang.org/genproto/googleapis/rpc/status"
```
rboyer, line 18: Why does this use a custom import name?

agent/rpc/peering/service.go
rboyer: `google.golang.org/grpc/codes` `codes.FailedPrecondition`
https://pkg.go.dev/google.golang.org/grpc@v1.47.0/codes#Code

agent/rpc/peering/service.go
acpana: did away with this

agent/rpc/peering/service.go
acpana: did away with this

agent/rpc/peering/service.go
acpana: thanks for the suggestion and offline clarifications!! made the change

agent/rpc/peering/stream_test.go
acpana: Chatting offline, I don't think it's possible to use `require.ElementsMatch` or `require.Contains` since `deets` contains `interface{}` types that need to be typecast. If we went to typecast, then we might as well look for the element we want to find like in the closure above.

kisunji: It's actually not about the typecast; you can do something like:
```go
	type Blah struct {
		Name string
	}
	expect := []Blah{{Name: "hi"}}
	got := []interface{}{Blah{Name: "hi"}}
	require.ElementsMatch(t, expect, got)
```

The reason it doesn't work in our case is because proto structs [have internal state](https://developers.google.com/protocol-buffers/docs/reference/go/faq#deepequal).

But there are solutions!
```go
	deets := st.Details()

	exp := []interface{}{
		&pbpeering.LeaderAddress{
			Address: "expected:address",
		},
	}
	prototest.AssertDeepEqual(t, exp, deets)
```

kisunji (approved):
LGTM! Thanks for being the guinea pig for using error with details :)

agent/rpc/peering/service.go
```diff
@@ -478,20 +475,15 @@ func (s *Service) StreamResources(stream pbpeering.PeeringService_StreamResource
...
 
-		details, err := anypb.New(&pbpeering.LeaderAddress{
-			Address: s.Backend.LeaderAddress().Get(),
-		})
+		st, err := grpcstatus.New(codes.FailedPrecondition,
+			"cannot establish a peering stream on a follower node").WithDetails(
+			&pbpeering.LeaderAddress{Address: s.Backend.LeaderAddress().Get()})
 		if err != nil {
 			s.logger.Error(fmt.Sprintf("failed to marshal the leader address in response; err: %v", err))
+			return grpcstatus.Error(codes.FailedPrecondition, "cannot establish a peering stream on a follower node")
```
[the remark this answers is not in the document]
acpana: spoke offline to bikeshed this as we might take a refactoring pass soon;

feedback point is still super valid!

Issue #13303, opened 2022-05-31
`i/o deadline reached` using blocking queries on consul 1.12.1


kevinschoonover:
#### Overview of the Issue

Hey Folks! I recently updated my cluster to consul 1.12.1 and started notice a lot of i/o deadline reached errors from vault-agent / nomad / consul-template. It seems like consul template is performing the following request on a consul client add getting a `rpc error making call: i/o deadline reached` after ~7 seconds

```console
> curl "http://localhost:8500/v1/kv/infrastructure/aws/vault_unseal_key_id?index=4594355&stale=&wait=60000ms"
rpc error making call: i/o deadline reached
```

when i downgrade to 1.12.0 this error goes away

```
# curl "http://localhost:8500/v1/kv/infrastructure/aws/vault_unseal_key_id?index=4594355&stale=&wait=60000ms"
[{"LockIndex":0,"Key":"infrastructure/aws/vault_unseal_key_id","Flags":0,"Value":"<redacted>","CreateIndex":4594355,"ModifyIndex":4594355}]
```

If you perform this same request on the server, you also don't get the i/o deadline reached error.

Not sure if this is something I did wrong in my configuration, but i didnt see any related timeout settings.

#### Reproduction Steps
1. Create consul server with 1.12.1 and consul client with 1.12.1
2. Create a key
3. Attempt to query that key with `curl "http://localhost:8500/v1/kv/<key>?index=<latest index>&stale=&wait=60000ms` where localhost:8500 is the **clients** HTTP ADDR. The server does not time out the http request

### Consul info for both Client and Server

<details>
  <summary>Client info</summary>

```
root@prod-do-sfo3-s-2vcpu-4gb-amd-vault-server-c970:/etc/consul.d# consul info
agent:
        check_monitors = 0
        check_ttls = 1
        checks = 5
        services = 5
build:
        prerelease =
        revision = e57f76ae
        version = 1.12.1
        version_metadata =
consul:
        acl = enabled
        known_servers = 3
        server = false
runtime:
        arch = amd64
        cpu_count = 2
        goroutines = 83
        max_procs = 2
        os = linux
        version = go1.18.1
serf_lan:
        coordinate_resets = 0
        encrypted = true
        event_queue = 0
        event_time = 491
        failed = 0
        health_score = 0
        intent_queue = 0
        left = 0
        member_time = 33109
        members = 14
        query_queue = 0
        query_time = 31
```

</details>

<details>
  <summary>Server info</summary>

```
[...]

PR #13344, opened 2022-06-02
Fix KVSGet method to handle QueryOptions properly

kisunji:
Fixes https://github.com/hashicorp/consul/issues/13303

### Description
Originally reported by @wjordan [here](https://github.com/hashicorp/consul/pull/11500#issuecomment-990110027), I requested the bugfix be removed from https://github.com/hashicorp/consul/pull/11500 assuming it was a separate issue. Unfortunately it turns out that broke KVS.Get RPC operations in 1.12.1.

### Cause
The KV.Get endpoint was being passed `**structs.KeyRequest`, which when handled by a client's RPC method, was not fulfilling this assertion: https://github.com/hashicorp/consul/blob/c48120d005539b687391b6861f24c13d553f5b05/agent/pool/pool.go#L632-L635

causing the ReadTimeout to be set to the default value of `rpc_hold_timeout` ([source](https://github.com/hashicorp/consul/blob/f507f62f3cebd0ff9a4259d44487e75b2596492a/agent/consul/client_test.go#L532)).

Set up a cluster with server and client using [vagrant](https://learn.hashicorp.com/tutorials/consul/get-started-create-datacenter).

```sh
# 1.12.1
$ consul join 192.168.56.10
Successfully joined cluster by contacting 1 nodes.

$ consul members
Node       Address             Status  Type    Build   Protocol  DC   Partition  Segment
agent-one  192.168.56.10:8301  alive   server  1.12.1  2         dc1  default    <all>
agent-two  192.168.56.11:8301  alive   client  1.12.1  2         dc1  default    <default>

$ consul kv put test foo
Success! Data written to: test

$ time curl "http://localhost:8500/v1/kv/test"
[{"LockIndex":0,"Key":"test","Flags":0,"Value":"Zm9v","CreateIndex":11,"ModifyIndex":11}]
real	0m0.013s
user	0m0.000s
sys	0m0.004s

$ time curl "http://localhost:8500/v1/kv/test?index=99&stale=&wait=60000ms"
rpc error making call: i/o deadline reached
real	0m7.013s <- deadline of rpc_hold_timeout
user	0m0.004s
sys	0m0.000s
```

```sh
# 1.13.0dev patch (version comes from main; ignore)
$ consul join 192.168.56.10
Successfully joined cluster by contacting 1 nodes.

$ consul members
Node       Address             Status  Type    Build      Protocol  DC   Partition  Segment
agent-one  192.168.56.10:8301  alive   server  1.12.1     2         dc1  default    <all>
agent-two  192.168.56.11:8301  alive   client  1.13.0dev  2         dc1  default    <default>

$ consul kv put test foo
Success! Data written to: test

$ time curl "http://localhost:8500/v1/kv/test"
[{"LockIndex":0,"Key":"test","Flags":0,"Value":"Zm9v","CreateIndex":12,"ModifyIndex":12}]
real	0m0.044s
user	0m0.012s
sys	0m0.004s

$ time curl "http://localhost:8500/v1/kv/test?index=99&stale=&wait=60000ms"
[{"LockIndex":0,"Key":"test","Flags":0,"Value":"Zm9v","CreateIndex":12,"ModifyIndex":12}]
real	1m2.708s <- wait=60000ms correctly applied
user	0m0.008s
sys	0m0.000s
```

### DIFF
--- a/.changelog/13344.txt
+++ b/.changelog/13344.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+kvs: Fixed a bug where query options were not being applied to KVS.Get RPC operations.
+```
--- a/agent/kvs_endpoint.go
+++ b/agent/kvs_endpoint.go
@@ -69,7 +69,7 @@ func (s *HTTPHandlers) KVSGet(resp http.ResponseWriter, req *http.Request, args 
 
 	// Make the RPC
 	var out structs.IndexedDirEntries
-	if err := s.agent.RPC(method, &args, &out); err != nil {
+	if err := s.agent.RPC(method, args, &out); err != nil {
 		return nil, err
 	}
 	setMeta(resp, &out.QueryMeta)


dhiaayachi (approved):
🚀 
It's likely not for the scope of this fix, but I'm wondering why we don't transition to an RPCInfo arg type for the RPC interface?

kisunji:
> 🚀 It's likely not for the scope of this fix, but I'm wondering why we don't transition to an RPCInfo arg type for the RPC interface?

Do you mean instead of `interface{}`? `RPCInfo` is embedded by a lot of RPC request type structs to hold RPC metadata but won't hold the data itself (unless we give it a payload method)

PR #13354, opened 2022-06-03
Egress gtw/intention rpc endpoint

dhiaayachi:
### Description
This add a new RPC endpoint to query upstream intentions for destinations. This will be used in the proxycfg package to have a watch that trigger on it. That watch will be used to trigger a watch on the destination config-entry which will in its turn be used to generate the right xDS configuration in the downstream connect-proxy sidecar.

* Test that the filtering is effective and that the right intentions are returned
* Test for the cache fetch


agent/consul/internal_endpoint.go
```diff
@@ -224,6 +224,27 @@ func (m *Internal) IntentionUpstreams(args *structs.ServiceSpecificRequest, repl
 	if done, err := m.srv.ForwardRPC("Internal.IntentionUpstreams", args, reply); done {
 		return err
 	}
+	return m.internalUpstreams(args, reply, structs.IntentionTargetService)
+}
+
+// IntentionUpstreamsDestination returns the upstreams of a service. Upstreams are inferred from intentions.
+// If intentions allow a connection from the target to some candidate service, the candidate service is considered
+// an upstream of the target.
```
kisunji, line 232: Can we edit the docs here and above to clarify the difference in behavior? (e.g. what's the difference between target service and destination?)

kisunji, line 232: Can we edit the docs here and above to clarify the difference in behavior? (e.g. what's the difference between target service and destination?)

agent/consul/state/intention.go
```diff
@@ -990,7 +990,13 @@ func (s *Store) intentionTopologyTxn(tx ReadTxn, ws memdb.WatchSet,
 	//				 Ideally those should be excluded as well, since they can't be upstreams/downstreams without a proxy.
 	//				 Maybe narrow serviceNamesOfKindTxn to services represented by proxies? (ingress, sidecar-
 	wildcardMeta := structs.WildcardEnterpriseMetaInPartition(structs.WildcardSpecifier)
-	index, services, err := serviceNamesOfKindTxn(tx, ws, structs.ServiceKindTypical, *wildcardMeta)
+	var services []*KindServiceName
+	if intentionTarget == structs.IntentionTargetService {
+		index, services, err = serviceNamesOfKindTxn(tx, ws, structs.ServiceKindTypical, *wildcardMeta)
+	} else {
+		// destinations can only ever be upstream, since they are only allowed as intention destination.
```
kisunji, line 997: Why is it one or the other now?

agent/consul/state/intention.go
```diff
@@ -990,7 +990,13 @@ func (s *Store) intentionTopologyTxn(tx ReadTxn, ws memdb.WatchSet,
...
+		index, services, err = serviceNamesOfKindTxn(tx, ws, structs.ServiceKindDestination, *wildcardMeta)
+	}
```
kisunji, line 999: General Q: Why if-else now? Compared to before where they were appended

kisunji, line 999: Why if-else now? Compared to before where they were appended

agent/consul/state/intention_test.go
```diff
@@ -2185,7 +2185,166 @@ func TestStore_IntentionTopology(t *testing.T) {
...
+					{
+						Name:           "kafka.store.org",
+						EnterpriseMeta: *structs.DefaultEnterpriseMetaInDefaultPartition(),
+					},
+				},
+			},
+		},
+		{
+			name:            "(upstream) acl allow all check only destinations show, service target",
+			defaultDecision: acl.Deny,
```
kisunji, line 2305: Either name or defaultDecision is inaccurate

kisunji, line 2305: Either name or defaultDecision is inaccurate

kisunji (approved):

agent/consul/state/intention.go
dhiaayachi: In the other PR (that @acpana is reviewing) I was intending to merge both in the same RPC endpoint and return a list of destinations and services, but I found out that this would be complex as it return a list of serviceNames which will make it hard to differentiate between the two later in the ProxyCfg.
So I chose to go with a separate RPC endpoint to avoid that and also avoid the retro compatibility dilemma.

agent/consul/internal_endpoint.go
dhiaayachi: Tried to clarify the doc and call out the deference with the other endpoint

PR #13361, opened 2022-06-03
[OSS] consul connect envoy command changes for agentless

riddhi89:
Changes the sourcing of the envoy bootstrap configuration to not use agent APIs and instead use the catalog(server) API.
This is done by passing a node-name flag to the command, (which can only be used with proxy-id).

Also fixes a bug where the golden envoy bootstrap config files used for tests did not use the expected destination service name in certain places for connect proxy kind.

Manual testing/demo: https://gist.github.com/riddhi89/b5823d922f6b76477b2e685325a6c8be


mkeeler (approved):

ishustava (approved):
:shipit:

PR #13367, opened 2022-06-06
Clean up imported nodes/services/checks as needed

freddygv:
### Description
Previously, imported data would never be deleted. As
nodes/services/checks were registered and deregistered, resources
deleted from the exporting cluster would accumulate in the imported
cluster.

This commit makes updates to replication so that whenever an update is
received for a service name we reconcile what was present in the catalog
against what was received.

This handleUpdateService method can handle both updates and deletions.

* Added unit tests for various cases. Ran them in both OSS and ENT.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/rpc/peering/replication.go
+++ b/agent/rpc/peering/replication.go
@@ -247,7 +247,7 @@ func (s *Service) handleUpdateService(
 
 	structsNodes, err := pbNodes.CheckServiceNodesToStruct()
 	if err != nil {
-		return fmt.Errorf("failed to convert protobuf instances to structs")
+		return fmt.Errorf("failed to convert protobuf instances to structs: %w", err)
 	}
 
 	// Normalize the data into a convenient form for operation.
@@ -300,7 +300,7 @@ func (s *Service) handleUpdateService(
 		// Missing nodes are not assumed to be deleted because there may be other service names
 		// registered on them.
 		// Inside we also track a map of node checks associated with the node.
-		unusedNodes = make(map[types.NodeID]struct{})
+		unusedNodes = make(map[string]struct{})
 
 		// deletedNodeChecks tracks node checks that were not present in the latest response.
 		// A single node check will be attached to all service instances of a node, so this
@@ -309,7 +309,7 @@ func (s *Service) handleUpdateService(
 	)
 	for _, csn := range storedInstances {
 		if _, ok := snap.Nodes[csn.Node.ID]; !ok {
-			unusedNodes[csn.Node.ID] = struct{}{}
+			unusedNodes[string(csn.Node.ID)] = struct{}{}
 
 			// Since the node is not in the snapshot we can know the associated service
 			// instance is not in the snapshot either, since a service instance can't
@@ -322,7 +322,7 @@ func (s *Service) handleUpdateService(
 				PeerName:       peerName,
 			})
 			if err != nil {
-				return fmt.Errorf("failed to deregister service: %w", err)
+				return fmt.Errorf("failed to deregister service %q: %w", csn.Service.CompoundServiceID(), err)
 			}
 
 			// We can't know if a node check was deleted from the exporting cluster
@@ -342,7 +342,8 @@ func (s *Service) handleUpdateService(
 				PeerName:       peerName,
 			})
 			if err != nil {
-				return fmt.Errorf("failed to deregister service: %w", err)
+				ident := fmt.Sprintf("partition:%s/peer:%s/node:%s/service_id:%s", csn.Service.PartitionOrDefault(), peerName, csn.Node.Node, csn.Service.ID)
+				return fmt.Errorf("failed to deregister service %q: %w", ident, err)
 			}
 
 			// When a service is deleted all associated checks also get deleted as a side effect.
@@ -374,7 +375,8 @@ func (s *Service) handleUpdateService(
 					PeerName:       peerName,
 				})
 				if err != nil {
-					return fmt.Errorf("failed to deregister check: %w", err)
+					ident := fmt.Sprintf("partition:%s/peer:%s/node:%s/check_id:%s", chk.PartitionOrDefault(), peerName, chk.Node, chk.CheckID)
+					return fmt.Errorf("failed to deregister check %q: %w", ident, err)
 				}
 			}
 		}
@@ -382,20 +384,23 @@ func (s *Service) handleUpdateService(
 
 	// Delete all deduplicated node checks.
 	for chk := range deletedNodeChecks {
+		nodeMeta := structs.NodeEnterpriseMetaInPartition(sn.PartitionOrDefault())
 		err := s.Backend.Apply().CatalogDeregister(&structs.DeregisterRequest{
 			Node:           chk.node,
 			CheckID:        chk.checkID,
-			EnterpriseMeta: *structs.NodeEnterpriseMetaInPartition(sn.PartitionOrDefault()),
+			EnterpriseMeta: *nodeMeta,
 			PeerName:       peerName,
 		})
 		if err != nil {
-			return fmt.Errorf("failed to deregister check: %w", err)
+			ident := fmt.Sprintf("partition:%s/peer:%s/node:%s/check_id:%s", nodeMeta.PartitionOrDefault(), peerName, chk.node, chk.checkID)
+			return fmt.Errorf("failed to deregister node check %q: %w", ident, err)
 		}
 	}
 
 	// Delete any nodes that do not have any other services registered on them.
 	for node := range unusedNodes {
-		_, ns, err := s.Backend.Store().NodeServices(nil, string(node), &sn.EnterpriseMeta, peerName)
+		nodeMeta := structs.NodeEnterpriseMetaInPartition(sn.PartitionOrDefault())
+		_, ns, err := s.Backend.Store().NodeServices(nil, node, nodeMeta, peerName)
 		if err != nil {
 			return fmt.Errorf("failed to query services on node: %w", err)
 		}
@@ -406,11 +411,13 @@ func (s *Service) handleUpdateService(
 
 		// All services on the node were deleted, so the node is also cleaned up.
 		err = s.Backend.Apply().CatalogDeregister(&structs.DeregisterRequest{
-			Node:     string(node),
-			PeerName: peerName,
+			Node:           node,
+			PeerName:       peerName,
+			EnterpriseMeta: *nodeMeta,
 		})
 		if err != nil {
-			return fmt.Errorf("failed to deregister node: %w", err)
+			ident := fmt.Sprintf("partition:%s/peer:%s/node:%s", nodeMeta.PartitionOrDefault(), peerName, node)
+			return fmt.Errorf("failed to deregister node %q: %w", ident, err)
 		}
 	}
 	return nil
--- a/agent/rpc/peering/replication.go
+++ b/agent/rpc/peering/replication.go
@@ -342,7 +342,8 @@ func (s *Service) handleUpdateService(
 				PeerName:       peerName,
 			})
 			if err != nil {
-				ident := fmt.Sprintf("partition:%s/peer:%s/node:%s/service_id:%s", csn.Service.PartitionOrDefault(), peerName, csn.Node.Node, csn.Service.ID)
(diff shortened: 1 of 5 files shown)


agent/rpc/peering/replication.go
```diff
@@ -282,8 +288,136 @@ func (s *Service) handleUpsertService(
...
+	}
+
+	// Delete any nodes that do not have any other services registered on them.
+	for node := range unusedNodes {
+		_, ns, err := s.Backend.Store().NodeServices(nil, string(node), &sn.EnterpriseMeta, peerName)
+		if err != nil {
+			return fmt.Errorf("failed to query services on node: %w", err)
+		}
+		if ns != nil && len(ns.Services) >= 1 {
+			// At least one service is still registered on this node, so we keep it.
```
kisunji, line 408: general Q: is it not valid to have unusedNodes with node checks?

agent/rpc/peering/replication.go
```diff
@@ -229,24 +230,29 @@ func (s *Service) handleUpsert(
...
-	structsNodes := make([]structs.CheckServiceNode, 0, len(csn.Nodes))
-	for _, pb := range csn.Nodes {
-		instance, err := pbservice.CheckServiceNodeToStructs(pb)
-		if err != nil {
-			return fmt.Errorf("failed to convert instance: %w", err)
-		}
-		structsNodes = append(structsNodes, *instance)
+	structsNodes, err := pbNodes.CheckServiceNodesToStruct()
+	if err != nil {
+		return fmt.Errorf("failed to convert protobuf instances to structs")
```
kisunji, line 255: are we suppressing original `err` intentionally here?

agent/rpc/peering/replication.go
```diff
@@ -282,8 +288,136 @@ func (s *Service) handleUpsertService(
...
+			// exist without a node.
+			// This will also delete all service checks.
+			err := s.Backend.Apply().CatalogDeregister(&structs.DeregisterRequest{
+				Node:           csn.Node.Node,
+				ServiceID:      csn.Service.ID,
+				EnterpriseMeta: csn.Service.EnterpriseMeta,
+				PeerName:       peerName,
+			})
+			if err != nil {
+				return fmt.Errorf("failed to deregister service: %w", err)
```
kisunji, line 330: Do we want to fail fast here vs using `merr` and grabbing all the raft apply errors? Wondering if it is meaningful to output node/service/check IDs in the error

agent/rpc/peering/stream_test.go
```diff
@@ -1221,3 +1268,855 @@ func expectReplEvents(t *testing.T, client *MockClient, checkFns ...func(t *test
...
+							},
+						},
+					},
+					{
+						Node: &structs.Node{
+							ID:        "af913374-68ea-41e5-82e8-6ffd3dffc461",
+							Node:      "node-foo",
+							Partition: defaultMeta.PartitionOrEmpty(),
+							PeerName:  peerName,
+							Address:   "10.0.0.1",
```
kisunji, line 1431: I don't think Address is asserted in this test and can be removed

agent/rpc/peering/replication.go
```diff
@@ -282,8 +288,136 @@ func (s *Service) handleUpsertService(
...
+		// unusedNodes tracks node names that were not present in the latest response.
+		// Missing nodes are not assumed to be deleted because there may be other service names
+		// registered on them.
+		// Inside we also track a map of node checks associated with the node.
+		unusedNodes = make(map[types.NodeID]struct{})
+
+		// deletedNodeChecks tracks node checks that were not present in the latest response.
+		// A single node check will be attached to all service instances of a node, so this
+		// deduplication prevents issuing multiple deregistrations for a single check.
+		deletedNodeChecks = make(map[nodeCheckTuple]struct{})
```
rboyer, line 313: Conveniently after i re-activate mesh-gateway-only-mode there will be no node checks replicated to worry about since the checks on a service will be flattened into a single summary check per service instance.

kisunji:
The comments really helped me review and understand what's going on in this PR! 👍

agent/rpc/peering/replication.go
```diff
@@ -282,8 +288,136 @@ func (s *Service) handleUpsertService(
...
+				})
+				if err != nil {
+					return fmt.Errorf("failed to deregister check: %w", err)
+				}
+			}
+		}
+	}
 
+	// Delete all deduplicated node checks.
+	for chk := range deletedNodeChecks {
```
rboyer, line 389: I don't think you can do this, because they'd be shared by other services right? Or are you assuming that if it vanished from THIS service on the node that implies it vanished from the node and is thus safe for erasure?

agent/rpc/peering/replication.go
```diff
@@ -282,8 +288,136 @@ func (s *Service) handleUpsertService(
...
+			PeerName:       peerName,
+		})
+		if err != nil {
+			return fmt.Errorf("failed to deregister check: %w", err)
+		}
+	}
+
+	// Delete any nodes that do not have any other services registered on them.
+	for node := range unusedNodes {
+		_, ns, err := s.Backend.Store().NodeServices(nil, string(node), &sn.EnterpriseMeta, peerName)
```
rboyer, line 403: This enterprise meta is going to do a single namespace and partition. I believe in this scenario you want the modifier that wildcards the namespace so you can check for services in any namespace on that node.

agent/rpc/peering/replication.go
```diff
@@ -282,8 +288,136 @@ func (s *Service) handleUpsertService(
...
+			return fmt.Errorf("failed to query services on node: %w", err)
+		}
+		if ns != nil && len(ns.Services) >= 1 {
+			// At least one service is still registered on this node, so we keep it.
+			continue
+		}
+
+		// All services on the node were deleted, so the node is also cleaned up.
+		err = s.Backend.Apply().CatalogDeregister(&structs.DeregisterRequest{
+			Node:     string(node),
```
rboyer, line 414: You need to specify the partition for the node using a empty-namespace, populated-partition entmeta here.

agent/rpc/peering/replication.go
freddygv: > Or are you assuming that if it vanished from THIS service on the node that implies it vanished from the node and is thus safe for erasure?

Yep that's right. Is there a case where that assumption wouldn't hold?

rboyer: I'm going to rescind this comment after I read through the rest of this.

agent/rpc/peering/replication.go
rboyer: Not when we're replicating things by service only. Nodes (and node checks) get created as side effects of service instance replication. If all of the services on a node are gone then the local node and checks that was replicated should be deleted as well.

agent/rpc/peering/replication.go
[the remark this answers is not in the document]
freddygv: From looking at catalog code I don't think the full name would show up, will add that here

freddygv: Hmmm I'm a bit hesitant to use `merr` for raft apply errors because the error could get really unwieldy. For example, if leadership is lost and all raft applies fail because of it there could be thousands of errors

Thoughts @rboyer ?

agent/rpc/peering/replication.go
freddygv: I don't think that's the case, the `indexNode` of the services table doesn't consider namespaces:
https://github.com/hashicorp/consul-enterprise/blob/main/agent/consul/state/catalog_ent.go#L87

Nodes table ID index also doesn't consider namespace:
https://github.com/hashicorp/consul-enterprise/blob/main/agent/consul/state/catalog_ent.go#L62

I'll swap this out for a node meta to make it more explicit though.

agent/rpc/peering/replication.go
freddygv: It's valid. Deregistering a node will deregister everything on that node: the node itself, services, checks.
 
So if a node is being deregistered we don't need to account for the checks separately.

agent/rpc/peering/replication.go
freddygv: Ah yep good catch

freddygv: I'll add some ent tests in a separate PR

agent/rpc/peering/replication.go
rboyer: Right, unless we wanted to squash those checks as well.

agent/rpc/peering/replication.go
rboyer: Almost all times a raft apply will work. If it doesn't work it almost certainly is going to be due to leadership loss so we'd want fast failure at the first raft error, not collection of errors and display at the end.

agent/rpc/peering/replication.go
```diff
@@ -282,8 +288,136 @@ func (s *Service) handleUpsertService(
...
+			EnterpriseMeta: *structs.NodeEnterpriseMetaInPartition(sn.PartitionOrDefault()),
+			PeerName:       peerName,
+		})
+		if err != nil {
+			return fmt.Errorf("failed to deregister check: %w", err)
+		}
+	}
+
+	// Delete any nodes that do not have any other services registered on them.
+	for node := range unusedNodes {
```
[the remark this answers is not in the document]
freddygv: Added a post-GA Asana task

agent/rpc/peering/replication.go
freddygv: I don't see a way for an error to come out of that function, but I just made a change to propagate it anyways

agent/rpc/peering/replication.go
```diff
@@ -282,8 +288,143 @@ func (s *Service) handleUpsertService(
...
+		sid := csn.Service.CompoundServiceID()
+		if _, ok := snap.Nodes[csn.Node.ID].Services[sid]; !ok {
+			err := s.Backend.Apply().CatalogDeregister(&structs.DeregisterRequest{
+				Node:           csn.Node.Node,
+				ServiceID:      csn.Service.ID,
+				EnterpriseMeta: csn.Service.EnterpriseMeta,
+				PeerName:       peerName,
+			})
+			if err != nil {
+				ident := fmt.Sprintf("partition:%s/peer:%s/node:%s/service_id:%s", csn.Service.PartitionOrDefault(), peerName, csn.Node.Node, csn.Service.ID)
```
rboyer, line 350: You are missing the namespace here.

agent/rpc/peering/replication.go
```diff
@@ -282,8 +288,143 @@ func (s *Service) handleUpsertService(
...
+				// Service checks that were not present can be deleted immediately because
+				// checks for a given service ID will only be attached to a single CheckServiceNode.
+				err := s.Backend.Apply().CatalogDeregister(&structs.DeregisterRequest{
+					Node:           chk.Node,
+					CheckID:        chk.CheckID,
+					EnterpriseMeta: chk.EnterpriseMeta,
+					PeerName:       peerName,
+				})
+				if err != nil {
+					ident := fmt.Sprintf("partition:%s/peer:%s/node:%s/check_id:%s", chk.PartitionOrDefault(), peerName, chk.Node, chk.CheckID)
```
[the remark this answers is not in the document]
rboyer: If you need to add ServiceID above, then it should also go here.

agent/rpc/peering/replication.go
```diff
@@ -282,8 +288,143 @@ func (s *Service) handleUpsertService(
...
+					deletedNodeChecks[tuple] = struct{}{}
+					continue
+				}
+
+				// If the check isn't a node check then it's a service check.
+				// Service checks that were not present can be deleted immediately because
+				// checks for a given service ID will only be attached to a single CheckServiceNode.
+				err := s.Backend.Apply().CatalogDeregister(&structs.DeregisterRequest{
+					Node:           chk.Node,
+					CheckID:        chk.CheckID,
```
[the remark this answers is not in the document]
rboyer: You do not, and actually that would deregister the whole service instance: https://github.com/hashicorp/consul/blob/main/agent/consul/fsm/commands_oss.go#L171

rboyer (approved):
LGTM with non blocking comment about some missing namespaces in error output. Also you need a rebase I believe.

PR #13369, opened 2022-06-06
test: retry the compat tests as often as other tests

rboyer:
### Description

The upgrade compatibility tests need frequent retrying to pass because they are slightly flaky. To alleviate manual work borrow similar `gotestsum` logic from the main go tests CI job to rerun them automatically here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -830,7 +830,15 @@ jobs:
             docker run consul:local consul version
             cd ./test/integration/consul-container
             echo "$subtests_formatted"
-            gotestsum -- -timeout=30m ./$subtests_formatted --target-version local --latest-version latest
+            gotestsum \
+              --format=short-verbose \
+              --debug \
+              --rerun-fails=3 \
+              -- \
+              -timeout=30m \
+              ./$subtests_formatted \
+              --target-version local \
+              --latest-version latest
             ls -lrt
           environment:
             # this is needed because of incompatibility between RYUK container and circleci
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -822,15 +822,19 @@ jobs:
...
             mkdir -p /tmp/test-results/
-            docker run consul:local consul version
             cd ./test/integration/consul-container
-            echo "$subtests_formatted"
-            gotestsum -- -timeout=30m ./$subtests_formatted --target-version local --latest-version latest
+            docker run --rm consul:local consul version
+            gotestsum \
+              --format=short-verbose \
+              --debug \
+              --rerun-fails=3 \
```
rboyer, line 831: In order to rerun failures and also pass args to `go test` (like timeout, target-version, etc) the packages need to be specified with `--packages`. For the custom args like target-version to work the need to come after the package placement, so we just avoid doing test splits here and specify `./...` twice to make it work at all.

.circleci/config.yml
```diff
@@ -822,15 +822,19 @@ jobs:
...
+              --packages="./..." \
+              -- \
+              -timeout=30m \
```
dhiaayachi, line 834: isn't 30m timeout extreme?

dhiaayachi (approved):
🚀 just a question about the timeout.

.circleci/config.yml
rboyer: It already was 30m prior to this PR?

PR #13245, opened 2022-05-26
Initial AWS Lambda documentation

erichaberkorn:
### Description
This is the initial version of the Lambda documentation. I still need to:
- Document the `enable_serverless_plugin` configuration option.
- Explain when Lambdas need to be registered, reconfigured, and deregistered in Consul.
- Link to these docs from other relevant pages

PR #13372, opened 2022-06-06
Document the `enable_serverless_plugin` Agent Configuration Option

erichaberkorn:


erichaberkorn:
> Overall LGTM.
> 
> I made a couple of suggestions and explained them. But they're definitely not blockers.
> 
> Flagging that https://consul.io/docs/lambda does not exist. Please double-check the link target. If the link is correct and the documentation does not exist yet, please wait until the Lambda documentation has been published before merging so that you can avoid the broken link.
> 
> Please let me know if you have any questions! Thanks!

Thanks for the review. I'm merging this into `lambda-beta-docs` rather than `main` and by the time everything is merged [/docs/lambda will exist](https://github.com/hashicorp/consul/pull/13245)

PR #13373, opened 2022-06-06
test: break dep on main consul module

rboyer:
### Description

The main consul module is not a great library and complicates some oss/ent module issues.

This undoes https://github.com/hashicorp/consul/pull/13371


dhiaayachi (approved):
🚀

dhiaayachi:
@rboyer test is failing because of an error with Jitter != nil. I think it's related Jitter not being a pointer.

PR #13376, opened 2022-06-06
Backport of ci: Add package verifications to build workflow onto release/1.11.x

eculver:
## Backport

This is a manually created backport for #13294. It also includes the resolved contents of #13200 that haven't been backported yet.

### Description
This is the last PR in the series to artifacts as part of the build workflow. The primary target for this last PR is for validating rpm and deb packages. There were TODOs for these last two jobs, which I'm adding. However, I'm also going back and making a few changes to the previous two PRs (#13200 and #13221):
- Verifications now all check the contents of the version string (as opposed to just running `consul version`)
- Renaming scripts to all match `verify_*.sh` pattern
- Moving zip/binary verification into its own job

The output can be seen in this run: 

### Links
- [CTIA-14](https://hashicorp.atlassian.net/browse/CTIA-14)
- [CTIA-17](https://hashicorp.atlassian.net/browse/CTIA-17)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,7 @@ on:
     # Sequence of patterns matched against refs/heads
     branches:
       - release/1.11.x
+      - backport/CTIA-17-verify-linux-packages/1.11
 
 env:
   PKG_NAME: consul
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -278,4 +280,137 @@ jobs:
...
+      - name: Run verification for ${{ matrix.arch }} binary
+        run: .github/scripts/verify_artifact.sh ${{ env.zip_name }} v${{ env.version }}
+
+  verify-darwin:
+    needs:
+      - get-product-version
+      - build-darwin
+    runs-on: macos-latest
+    strategy:
+      fail-fast: true
```
huikang, line 324: Do we need line 324 given strategy has empty matrix?

eculver: Good point. We could do without this. I will make a follow-up PR that gets backported so that everything goes in `main`.

PR #13382, opened 2022-06-07
Load test, upgrade packer version, fix k6s installation

huikang:
## Description:
Changes to load test
- upgrade packer version and the template file (json to hcl2 format)
- fix k6s installation
- fix the tag issue in aws provider
- add name tag to the load generator server, `Name = "consul-load-generator-server-${local.random_name}"`

## Links
[CTIA-27](https://hashicorp.atlassian.net/browse/CTIA-27)

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


.circleci/terraform/load-test/main.tf
```diff
@@ -7,12 +7,6 @@ provider "aws" {
   assume_role {
     role_arn = var.role_arn
   }
-
-  default_tags {
```
eculver, line 11: It would be good to keep these tags so that if the TF destroy fails, we can go back and clean them up manually. It's also helpful for cost accounting if we ever wanted to look at what this costs to run.

test/load/terraform/test-servers.tf
```diff
@@ -54,6 +54,9 @@ resource "aws_instance" "test-server" {
   vpc_security_group_ids      = toset([aws_security_group.test-servers.id])
   associate_public_ip_address = var.test_public_ip
   subnet_id                   = (module.vpc.public_subnets)[0]
+  tags = {
```
eculver, line 57: Should we add similar tags to the other resources? And to my other comment, I'm wondering if we should add the same `Environment = "consul-load-test"` tag to all the resources created so that we can track them all down if we need to.

test/load/terraform/test-servers.tf
```diff
@@ -37,13 +37,13 @@ resource "aws_security_group" "test-servers" {
...
     protocol    = "6"
-    cidr_blocks = [var.vpc_cidr]
+    cidr_blocks = ["0.0.0.0/0"]
   }
   egress {
     from_port   = 0
     to_port     = 0
     protocol    = "-1"
-    cidr_blocks = [var.vpc_cidr]
+    cidr_blocks = ["0.0.0.0/0"]
```
eculver, line 46: I'm wondering why this was set to `var.vpc_cidr` in the first place? Shouldn't this be more restrictive than allowing "anywhere"?

eculver (commented):
This looks good, just a couple questions. I also want to run through it all myself to see about any other docs changes.

.circleci/terraform/load-test/main.tf
huikang: I agreed, but the `default_tags` seems not working with this old version of aws provider (see the [error log](https://app.circleci.com/pipelines/gh/hashicorp/consul/32504/workflows/736b3714-1777-4121-b433-37e5bba6149b/jobs/704717?invite=true#step-108-1478)). After removing these lines, the terraform apply works.

There is a `always` step in our circleCI job to destroy the resources:
https://github.com/hashicorp/consul/blob/915f1089f03f25b3ca0bf6edd7c719d0c9b00f4e/.circleci/config.yml#L1012-L1017

Maybe we can re-enable the global tag after we upgrade the aws provider and module. WDYT?

test/load/terraform/test-servers.tf
huikang: It was set to vpc_cidr (="10.0.0.1/16") because of the security alert. However, we have to allow ingress on port 22 in order to allow the terraform provisioner to run the workload command https://github.com/hashicorp/consul/blob/915f1089f03f25b3ca0bf6edd7c719d0c9b00f4e/test/load/terraform/test-servers.tf#L57-L69

The egress is to allow the test server to run some install script: https://github.com/hashicorp/consul/blob/main/test/load/packer/loadtest-ami/scripts/install-k6.sh 

I think the consul cluster instances allows this too for downloading consul: https://github.com/hashicorp/terraform-aws-consul/blob/81b6173e8f0869ce96cbc93545dce4b8f68a16e0/modules/consul-cluster/main.tf#L173-L178

test/load/terraform/test-servers.tf
huikang: Some of the resources are already tagged like consul EC2 instances; I think this one is a leftover. Please see my earlier comments about the idea of global tag.

huikang (commented):
@eculver , thanks for your comments and suggestions. I think some of them needs to be updated to the deprecated aws-consul module (https://github.com/hashicorp/consul/tree/main/test/load). I added some of them to the TODO list in this [doc](https://docs.google.com/document/d/1EOtBjJH57m2y8GROnlXr3YC-0ONv4pY3mDgBhYMq5Jo/edit#bookmark=id.g4x33jwpl1af).

test/load/packer/loadtest-ami/loadtest.pkr.hcl
```diff
@@ -0,0 +1,51 @@
+packer {
```
eculver, line 1: We should update the [`./test/load/packer/README.md`](https://github.com/hashicorp/consul/tree/7f3c3b093221f62779281012eb9ac0e26774c7da/test/load/packer#load-test-ami) to point to this new file.

eculver:
We should add a note in [`./test/load/terraform/README.md`](https://github.com/hashicorp/consul/blob/7f3c3b093221f62779281012eb9ac0e26774c7da/test/load/terraform/README.md#terraform-consul-load-testing) to run `terraform init` first if you have never ran it before.

.circleci/terraform/load-test/main.tf
eculver: Sounds good to me.

test/load/terraform/test-servers.tf
eculver: Sorry, I probably should have put this comment on the line for the ingress block. I was hoping we could restrict the ingress to something other than "anywhere". For now, if we kept it as a variable instead of hard-coding it as "0.0.0.0" we could let the user set it in the `var.tfvars` file. What do you think about this option?

huikang: @eculver , agreed. So I added a new variable named `vpc_allwed_ssh_cidr` and set it to `10.0.0.0/16` by default and add some note in `test/load/terraform/README.md` (please see the last commit for the change)

Issue #10637, opened 2021-07-19
windows service health check


rismoney:
Agents should have a native health check to determine if a service is running.  Many services dont have network listeners and handle either outbound, or local service capability.  Avoiding process creation/kill of calling a cmd sc/pwsh cmdlet to check the existence of a service would be great to just handle in native go.

I cant imagine this to be a big ask, as other health checks are seemingly simplistic functions.

jkirschner-hashicorp:
Hi @rismoney,

It sounds like there's a service with no network listeners that you want to have health checks on. Rather than using a `Script + Interval` check, you'd prefer a native check to see if the service/process is running.

If I've understood correctly (let me know), I have a few follow-up questions:

- Is a `Script + Interval` check what you're using instead today? If so, what are you hoping would improve/change for your use case as a result of switching to a native check?
- What configuration do you imagine passing in? (e.g., service name, interval, anything else?)
- What information do you imagine the health status (passing, warning, critical) being based on? (e.g., based on [ServiceController.Status](https://docs.microsoft.com/en-us/dotnet/api/system.serviceprocess.servicecontroller.status?view=dotnet-plat-ext-5.0#System_ServiceProcess_ServiceController_Status), or something else)

rismoney:
> It sounds like there's a service with no network listeners that you want to have health checks on. Rather than using a Script Interval check, you'd prefer a native check to see if the service/process is running.

Yes.

> Is a Script + Interval check what you're using instead today? If so, what are you hoping would improve/change for your use case as a result of switching to a native check?

This is what we _would_ use.  I am running very latency sensitive financial transaction applications.   Spawning to cmd/powershell would get expensive as the service count were to rise invoking a get-service/sc or similar operation.  We are trying to do service inventory on 500 nodes with 25-50 custom services per node, and would not want the health check to be shell/script based with another process invocation.  Also my understanding is any approach that uses wmi calls isn't going to scale well (the prometheus exporter kind of taught that lesson)


> What configuration do you imagine passing in? (e.g., service name, interval, anything else?)

Yes, unfortunately the only thing we will know is it's name and that it's running so those two attributes are nearly all we have.  That itself would be the health check.  I know it's kind of primitive,   The applications with listeners are also fairly sensitive, and we don't want to interrogate them by port.  They have their own built in health capabilities, that essentially will down themselves if  unhealthy.

> What information do you imagine the health status
[...]

PR #10194, opened 2021-05-06
ui: Serf Health Check warning notice

johncowen:
When the Consul serf health check is failing, this means that the health checks registered with the agent may no longer be correct. Therefore we show a notice to the user when we detect that the serf health check is failing both for the health check listing for nodes and for service instances:

Screenshot 2021-05-06 at 11 18 21

Acceptance tests added for both Node and Service Instance pages.

There were a few little things we fixed up whilst we were here:

1. We use our `@replace` decorator to replace an empty `Type` with `serf` in the model.
2. We noticed that `ServiceTags` can be `null`, so we replace that with an empty array.
3. We added docs for both our `Notice` component and the `Consul::HealthCheck::List` component. `Notice` now defaults to `@type=info`.

PR #13388, opened 2022-06-07
Feature: Health checks windows service

deblasis:
### Description
To provide windows service health check functionality addressing the issue #10637

The PR is Draft because I would like you guys to point me to your preferred way of mocking syscalls and I will update the feature branch with the relevant tests.

### Links

- Issue: #10637
- PRD (kinda 🙂): https://gist.github.com/deblasis/b6dd645a8bd5e2b3c9b95bb30bbb238c

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


Amier3:
Hey @deblasis 

Super excited to see this PR come through, thank you! After looking ( and asking ) around it doesn't look like we have any established pattern for mocking syscalls. You're free to add the tests in whenever you get a chance 👍 😄

trujillo-adam (commented):
I reviewed the documentation updates and made a few suggestions to improve clarity.

jkirschner-hashicorp:
@deblasis : for context, our education/docs team has performed a review (above). The engineering review still needs to be conducted! I'd recommend waiting for an engineering review before making changes so you can address all the feedback in one go.

agent/checks/os_service_windows.go
```diff
@@ -0,0 +1,114 @@
...
+
+	m := win.getWindowsSvcMgr(client.scHandle)
+	defer func() {
+		errDisconnect := m.Disconnect()
+		if isHealthy || errDisconnect == nil || err != nil {
+			return
+		}
+		//unreachable at the moment but we might want to log this error. leaving here for code-review
+		err = errDisconnect
+	}()
```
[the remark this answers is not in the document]
kisunji: From a quick glance I think the fail conditions for Disconnect and Close are unlikely and can be ignored. 

In the worst case scenario the process holding the ServiceManager open doesn't get cleaned up but if the OS is degraded to the point where it can't close there may be worse problems 😅 

I'd vote for a simple defer but it's not a strong opinion by any means.

agent/checks/os_service_windows.go
```diff
@@ -0,0 +1,114 @@
...
+	win windowsSystem = windowsOS{}
+)
+
+type OSServiceClient struct {
+	scHandle windows.Handle
+}
+
+func NewOSServiceClient() (*OSServiceClient, error) {
+	var s *uint16
+	scHandle, err := win.OpenSCManager(s, nil, windows.SC_MANAGER_CONNECT)
```
kisunji, line 25: My guess is that this should work as well
```suggestion
	scHandle, err := win.OpenSCManager(nil, nil, windows.SC_MANAGER_CONNECT)
```

agent/checks/os_service_windows.go
```diff
@@ -0,0 +1,114 @@
...
+
+	if err != nil {
+		return nil, fmt.Errorf("error connecting to service manager: %w", err)
+	}
```
kisunji, line 29:
```suggestion
	if err != nil {
		return nil, fmt.Errorf("error connecting to service manager: %w", err)
	}
```

kisunji: Removing newline

agent/checks/check.go
```diff
@@ -1047,6 +1048,99 @@ func (c *CheckGRPC) Stop() {
 	}
 }
 
+type CheckOSService struct {
+	CheckID       structs.CheckID
+	ServiceID     structs.ServiceID
+	OSService     string
+	Interval      time.Duration
+	Timeout       time.Duration
```
kisunji, line 1056: I don't see `Timeout` being used anywhere. Is it worth implementing a timeout error? Unlike gRPC or script checks I would assume OS service polling shouldn't take too long and it's not worth enforcing.

kisunji (approved):
Thank you for this large chunk of design and implementation! I'm not sure if we'll be able to implement Windows machines for CI to test this but I'll try it out locally on my desktop and ask the team how we want to handle this windows-specific feature.

I left a few minor comments that are non-blocking.

agent/checks/os_service_windows.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return fmt.Errorf("error querying service status: %w", err)
+	}
+
+	switch status.State {
+	case svc.Running:
+		err = nil
+		isHealthy = true
+	case svc.Paused, svc.Stopped:
+		err = ErrOSServiceStatusCritical
```
dhiaayachi, line 72: maybe we can wrap the service state inside `ErrOSServiceStatusCritical` to help user have what state is failing the check.

agent/checks/os_service_windows.go
```diff
@@ -0,0 +1,114 @@
...
+		if isHealthy || errClose == nil || err != nil {
+			return
+		}
+		//unreachable at the moment but we might want to log this error. leaving here for code-review
+		err = errClose
+	}()
+	status, err := service.Query()
+	if err != nil {
+		return fmt.Errorf("error querying service status: %w", err)
+	}
```
dhiaayachi, line 65: This is a bit confusing for me. I don't know how in details the windows library work, but looking to this, `getWindowsSvc` will only allocate the struct and don't do any access, right? in that case what will happen if Query fail for some reason but we still call `service.Close` in the defer? will `service.Close` error?

agent/checks/os_service_windows.go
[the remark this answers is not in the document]
dhiaayachi: Same comment here as the one about `getWindowsSvc`. What would happen if `OpenService`  fail and we execute Disconnect anyway?

dhiaayachi (commented):
Thank you for working on this @deblasis! 
I left few questions and an improvement suggestion but nothing substantial to add to @kisunji review.

agent/checks/check.go
```diff
@@ -1047,6 +1048,99 @@ func (c *CheckGRPC) Stop() {
...
+		}
+	}
+}
+
+func (c *CheckOSService) doCheck() (string, error) {
+
+	target := c.OSService
+	if c.OSService != "" {
+		target = c.OSService
+	}
```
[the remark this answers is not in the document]
jkirschner-hashicorp: Said another way: can't L1115-L1120 be replaced with just the below?
```
err := c.Client.Check(c.OSService)
```

It seems like the `if` conditional has no effect.

deblasis:
Hi guys, sorry for the delay. I have updated the PR taking into account your feedback and I made a small improvement in the reported service state that's now in string format.

Please also check if the updated website/content/docs/discovery/checks.mdx  from `main` is OK since you guys updated the format and I had to make a couple of adjustments.

kisunji (approved):
LGTM! Have you tested this yourself as a Windows Service (beyond the mocked unit tests in this PR)?

I'll try to set aside some time to test locally then merge if things look good and there are no other concerns from the team.

agent/agent.go
```diff
@@ -2900,6 +2906,45 @@ func (a *Agent) addCheck(check *structs.HealthCheck, chkType *structs.CheckType,
...
+				OSService:     chkType.OSService,
+				Timeout:       chkType.Timeout,
+				Interval:      chkType.Interval,
+				Logger:        a.logger,
+				Client:        a.osServiceClient,
+				StatusHandler: statusHandler,
+			}
+			if prev := a.checkOSServices[cid]; prev != nil {
+				prev.Stop()
+			}
```
jkirschner-hashicorp, line 2944: @deblasis, @kisunji : what's the difference between this condition and the one on lines L2910-L2913?
```
			if existing, ok := a.checkOSServices[cid]; ok {
				existing.Stop()
				delete(a.checkOSServices, cid)
			}
```

It seems like both cases involve checking to see if a previous check exists, and if it does, stopping it. But the first invocation also has a delete operation.

Am I misreading something?

website/content/docs/discovery/checks.mdx
```diff
@@ -273,6 +273,43 @@ check = {
 
 </CodeTabs>
 
+### OSService check ((#osservice-interval))
```
jkirschner-hashicorp, line 276: The `((#osservice-interval))` part should be removed. That's only used to create backwards compatibility with an old anchor tag (if a section is renamed). This section never previously existed, so there's no need for an alias!

```suggestion
### OSService check
```

website/content/docs/discovery/checks.mdx
```diff
@@ -273,6 +273,43 @@ check = {
...
+
+OSService checks periodically direct the Consul agent to monitor the health of a service running on
+the host operating system as either a Windows service (Windows) or a SystemD service (Unix).
+The check is logged as `healthy` if the service is running.
+If it is stopped or not running, the status is `critical`. All other result sets
```
jkirschner-hashicorp, line 281:
```suggestion
If it is stopped or not running, the status is `critical`. All other results set
```

jkirschner-hashicorp (commented):
A few small suggestions and one question. Nice work, @deblasis :)

agent/agent.go
kisunji: Good eye! I think this may be a copy-paste from the docker check case above. This block looks unnecessary:
```go
if prev := a.checkOSServices[cid]; prev != nil {
	prev.Stop()
}
```

deblasis: You are right guys, it's a copy-paste from the Docker check above. 

I guess we should log that we have to remove the redundant lines in the Docker check as well.
Unless you want to do it in this PR

kisunji: I wouldn't mind if you just removed the line from the docker case as well

kisunji:
@deblasis would you be able to rebase or merge from `main`? It should have fixes for the failing integration tests

kisunji:
Thanks! Tests are all passing. Will make some time this week to test on Windows

kisunji:
Apologies for the delay; I didn't have time this week but will make sure to test next week. I will also post a comment on the original github issue to see if there is anyone who would like to test out the feature.

kisunji:
I discovered a few issues while testing:
1. We need to initialize `checkOSServices` otherwise there is a panic due to assigning to a nil map
2. (not a blocker) UI does not recognize OS as a type of check so it appears as a "Serf Check" in the UI. 
3. I am seeing the following in the logs:
```
2022-09-20T17:29:34.057-0400 [DEBUG] agent: Check failed: check=service-check error="error accessing service: The handle is invalid."
```

kisunji:
Reproduction steps:
```hcl
// config/nodecheck.hcl
check = {
  name = "node-check"
  os_service = "AdobeARMService"
  interval = "10s"
}

// config/svccheck.hcl
check = {
  name = "service-check"
  os_service = "AdobeARMService"
  service_id = "hello"
  interval = "10s"
}
```

```powershell
# powershell as admin
.\consul.exe agent -dev -config-dir=".\config"
```

Confirming that AdodeARMService is running
``` powershell
PS C:\Users\Chris> Get-Service -Name "AdobeARMService"

Status   Name               DisplayName
------   ----               -----------
Running  AdobeARMservice    Adobe Acrobat Update Service
```

deblasis:
Hi @kisunji! Sorry about that. But I have good news 👇

> * We need to initialize `checkOSServices` otherwise there is a panic due to assigning to a nil map

Addressed in https://github.com/hashicorp/consul/pull/13388/commits/fc0dd92dcf54e6a06abce5c24b47464263fca94f

> * (not a blocker) UI does not recognize OS as a type of check so it appears as a "Serf Check" in the UI.

I have noticed that I missed some properties and the corresponding generated mappings (fixed in https://github.com/hashicorp/consul/pull/13388/commits/461b42ed480cd39e8a41eb9207470958420ff2ce) but however I believe there's a bug elsewhere, I have added another check to... check 🙂 (repro config below) and I am seeing `serf` in there as well:
image

Since I was looking at the UI, I noticed the lack of message so I have added a "Healthy" message with https://github.com/hashicorp/consul/pull/13388/commits/5719fd65607b8af263ed06455cd6affe75c0d0c4

> * I am seeing the following in the logs:
> ```
> 2022-09-20T17:29:34.057-0400 [DEBUG] agent: Check failed: check=service-check error="error accessing service: The handle is invalid."
> ```

Addressed in https://github.com/hashicorp/consul/pull/13388/commits/f440966a386ad8a9140ac44d72bdb1f4fa0eaf7d

I have tested the above with:
```hcl
// config/nodecheck.hcl
check = {
  name = "node-check"
  os_service = "Dhcp"
  interval = "10s"
}

// config/h2ping.hcl
check = {
  id = "h2ping-check"
  name = "h2ping"
  h2ping = "localhost:22222"
  interval = "10s"
  h2ping_use_tls = false
}
```

kisunji:
Ah it must be a UI bug then. Thanks for the quick updates, will try to test again soon

jkirschner-hashicorp:
@johncowen: any ideas on why a newer health check type (`os_service`, `h2ping`) would show up as a "Serf" health check on the UI?

johncowen:
If the `Type` field in the HTTP response that the UI uses is empty then we assume that is a serf check. Possibly or possibly not useful UI PR https://github.com/hashicorp/consul/pull/10194

I would guess that the new healthchecks are coming out with empty `Type`s in the HTTP responses.

Umm separate thing but if we have new healthcheck types then the UI code might need to be aware of those for menus etc. not totally sure if we inspect those or if they are hardcoded.

At the very least UI should add these possible values to our mock API I think

deblasis:
Hi @johncowen !  Nice seeing you again around here (you reviewed a PR of mine back in the day) :)

> If the `Type` field in the HTTP response that the UI uses is empty then we assume that is a serf check. Possibly or possibly not useful UI PR #10194
> 
> I would guess that the new healthchecks are coming out with empty `Type`s in the HTTP responses.

I can confirm that this is the case, the `Type` is empty in the HTTP responses even for "older" checks like **h2ping**.

website/content/api-docs/agent/check.mdx
```diff
@@ -245,6 +245,8 @@ The check sends datagrams to the value specified at the interval specified in th
 If the datagram is sent successfully or a timeout is returned, the check is set to the `passing` state.
 The check is logged as `critical` if the datagram is sent unsuccessfully.
 
+- `OSService` `(string: "")` - Specifies the name of a service to check that is running as an OS-level service; either Windows Services on Windows or SystemD services on Unix.
```
trujillo-adam, line 248:
```suggestion
- `OSService` `(string: "")` - Specifies an OS-level service to check. You can specify either `Windows Services` on Windows or `SystemD` services on Unix.
```
The current wording reads as if we are checking the actual Windows Service or SystemD application, but stated in a convoluted way. I think we might mean, though, that this option tells Consul that the service it is configured to check runs as a Windows Service or SystemD service. Is that correct?

website/content/docs/ecs/configuration-reference.mdx
```diff
@@ -186,6 +186,7 @@ Defines the Consul checks for the service. Each `check` object may contain the f
 | `method` | `string` | optional | Specifies the HTTP method to be used for an HTTP check. When no value is specified, `GET` is used.  |
 | `name` | `string` | optional | The name of the check.  |
 | `notes` | `string` | optional | Specifies arbitrary information for humans. This is not used by Consul internally.  |
+| `os_service` | `string` | optional | Specifies an `OS Service` check. Must be the name of the service which status will be checked every `interval`.  |
```
trujillo-adam, line 189:
```suggestion
| `os_service` | `string` | optional | Specifies the name of a service on which to perform an [OS service check](/docs/discovery/checks#osservice-check). The check runs according the frequency specified in the `interval` parameter.  |
```

website/content/api-docs/agent/check.mdx
deblasis: Correct. Currently, this feature works only on windows but the main pieces for making it work with SystemD are in place. The design was intended to offer the same look-n-feel to the user independently of the implementation details/OS.

deblasis:
@trujillo-adam: 

> Had a few suggestions and questions w/r/t the docs. You did a really good job on the main description of the OS service check, BTW.

Suggestions accepted! Thanks but @jkirschner-hashicorp helped a lot already giving it the first pass a few weeks ago. 
My initial version was not that clear. Now it looks pretty good. 👍

website/content/api-docs/agent/check.mdx
```diff
@@ -245,6 +245,8 @@ The check sends datagrams to the value specified at the interval specified in th
 If the datagram is sent successfully or a timeout is returned, the check is set to the `passing` state.
 The check is logged as `critical` if the datagram is sent unsuccessfully.
 
+- `OSService` `(string: "")` - Specifies an OS-level service to check. You can specify either `Windows Services` on Windows or `SystemD` services on Unix.
```
trujillo-adam, line 248:
```suggestion
- `OSService` `(string: "")` - Specifies the type of OS-level service to check. You can specify either `Windows Services` on Windows or `SystemD` services on Unix.
```
Apologies, I think my last suggestion was still too ambiguous. I think this makes the purpose of the parameter clearer.

trujillo-adam (commented):
One more tweak

website/content/api-docs/agent/check.mdx
deblasis: It's not the type that we have to specify but its name

website/content/api-docs/agent/check.mdx
deblasis: I just realized that I accepted this by mistake and I feel like it should be reverted and/or rephrased to indicate that it's the `name` of the OS-specific service that goes into that.

website/content/api-docs/agent/check.mdx
deblasis: Also, this is related: https://github.com/hashicorp/consul/pull/13388#discussion_r977810915

How about: 
```suggestion
- `OSService` `(string: "")` - Specifies the name/identifier of an OS-level service to check. You can specify either `Windows Services` on Windows or `SystemD` services on Unix.
```

trujillo-adam, line 248:
```suggestion
- `OSService` `(string: "")` - Specifies the name of an OS-level service. Configure this option if the service you intend to check runs as an OS-level service. You can specify either `Windows Services` on Windows or `SystemD` services on Unix. 
```
I think trying to combine the action to specify the name and the action of performing the check (if that makes sense) is what's causing confusion for me. We should state them separately. First, what the param is for. Second, what the param means in the context of the overall API call. What do you think?

deblasis: Hey @trujillo-adam ! Any thoughts on this?

deblasis:
👍

website/content/api-docs/agent/check.mdx
deblasis: @trujillo-adam: sounds good!

```suggestion
- `OSService` `(string: "")` - Specifies the identifier of an OS-level service to check. You can specify either `Windows Services` on Windows or `SystemD` services on Unix.
```

deblasis:
I think it should be OK now.  Shall we🚢 it ? 😄

kisunji:
@deblasis I've tested this successfully on Windows so this is good to 🚢 ! I'll try to fix the rebases

PR #13391, opened 2022-06-07
feat: xDS Generation for Terminating Gateway Destinations

DanStough:
### Description
We are currently adding support for Destinations (nodeless egress services) in Consul Terminating Gateways. This PR adds the xDS generation at the Terminating gateway side of the connection, including listeners and clusters.

Since traffic may be encrypted from the client, all traffic appears as TCP. For hostname traffic, which may be part of a wildcard definition, we route traffic based on SNI to a dynamic forward proxy. For IP traffic, we use a placeholder SNI on the listener to identify the destination and route to a static envoy cluster.

The forward proxy configuration was tested with a local envoy instance based on [this setup](https://github.com/DanStough/envoy-examples/tree/main/tcp-dynamic-forward-proxy).

### PR Checklist

* [X] updated test coverage
* [X] not a security concern

### DIFF
--- a/agent/xds/clusters.go
+++ b/agent/xds/clusters.go
@@ -9,6 +9,8 @@ import (
 	envoy_cluster_v3 "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
 	envoy_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	envoy_endpoint_v3 "github.com/envoyproxy/go-control-plane/envoy/config/endpoint/v3"
+	envoy_cluster_dynamic_forward_proxy_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/clusters/dynamic_forward_proxy/v3"
+	envoy_common_dynamic_forward_proxy_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/common/dynamic_forward_proxy/v3"
 	envoy_tls_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3"
 	envoy_upstreams_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/upstreams/http/v3"
 	envoy_matcher_v3 "github.com/envoyproxy/go-control-plane/envoy/type/matcher/v3"
@@ -27,6 +29,12 @@ import (
 	"github.com/hashicorp/consul/agent/structs"
 )
 
+const (
+	dynamicForwardProxyClusterName         = "dynamic_forward_proxy_cluster"
+	dynamicForwardProxyClusterTypeName     = "envoy.clusters.dynamic_forward_proxy"
+	dynamicForwardProxyClusterDNSCacheName = "dynamic_forward_proxy_cache_config"
+)
+
 // clustersFromSnapshot returns the xDS API representation of the "clusters" in the snapshot.
 func (s *ResourceGenerator) clustersFromSnapshot(cfgSnap *proxycfg.ConfigSnapshot) ([]proto.Message, error) {
 	if cfgSnap == nil {
@@ -37,7 +45,7 @@ func (s *ResourceGenerator) clustersFromSnapshot(cfgSnap *proxycfg.ConfigSnapsho
 	case structs.ServiceKindConnectProxy:
 		return s.clustersFromSnapshotConnectProxy(cfgSnap)
 	case structs.ServiceKindTerminatingGateway:
-		res, err := s.makeGatewayServiceClusters(cfgSnap, cfgSnap.TerminatingGateway.ServiceGroups, cfgSnap.TerminatingGateway.ServiceResolvers)
+		res, err := s.clustersFromSnapshotTerminatingGateway(cfgSnap)
 		if err != nil {
 			return nil, err
 		}
@@ -276,7 +284,7 @@ func (s *ResourceGenerator) clustersFromSnapshotMeshGateway(cfgSnap *proxycfg.Co
 			continue // skip local
 		}
 
-		opts := gatewayClusterOpts{
+		opts := clusterOpts{
 			name:              connect.GatewaySNI(key.Datacenter, key.Partition, cfgSnap.Roots.TrustDomain),
 			hostnameEndpoints: cfgSnap.MeshGateway.HostnameDatacenters[key.String()],
 			isRemote:          true,
@@ -298,7 +306,7 @@ func (s *ResourceGenerator) clustersFromSnapshotMeshGateway(cfgSnap *proxycfg.Co
 			if key.Datacenter == cfgSnap.Datacenter {
 				hostnameEndpoints = nil
 			}
-			opts := gatewayClusterOpts{
+			opts := clusterOpts{
 				name:              cfgSnap.ServerSNIFn(key.Datacenter, ""),
 				hostnameEndpoints: hostnameEndpoints,
 				isRemote:          !key.Matches(cfgSnap.Datacenter, cfgSnap.ProxyID.PartitionOrDefault()),
@@ -309,7 +317,7 @@ func (s *ResourceGenerator) clustersFromSnapshotMeshGateway(cfgSnap *proxycfg.Co
 
 		// And for the current datacenter, send all flavors appropriately.
 		for _, srv := range cfgSnap.MeshGateway.ConsulServers {
-			opts := gatewayClusterOpts{
+			opts := clusterOpts{
 				name: cfgSnap.ServerSNIFn(cfgSnap.Datacenter, srv.Node.Node),
 			}
 			cluster := s.makeGatewayCluster(cfgSnap, opts)
@@ -325,6 +333,25 @@ func (s *ResourceGenerator) clustersFromSnapshotMeshGateway(cfgSnap *proxycfg.Co
 	clusters = append(clusters, c...)
 
 	return clusters, nil
+}
+
+// clustersFromSnapshotTerminatingGateway returns the xDS API representation of the "clusters"
+// for a terminating gateway. This will include 1 cluster per Destination associated with this terminating gateway.
+func (s *ResourceGenerator) clustersFromSnapshotTerminatingGateway(cfgSnap *proxycfg.ConfigSnapshot) ([]proto.Message, error) {
+	res := []proto.Message{}
+	gwClusters, err := s.makeGatewayServiceClusters(cfgSnap, cfgSnap.TerminatingGateway.ServiceGroups, cfgSnap.TerminatingGateway.ServiceResolvers)
+	if err != nil {
+		return nil, err
+	}
+	res = append(res, gwClusters...)
+
+	destClusters, err := s.makeDestinationClusters(cfgSnap)
+	if err != nil {
+		return nil, err
+	}
+	res = append(res, destClusters...)
+
+	return res, nil
 }
 
 func (s *ResourceGenerator) makeGatewayServiceClusters(
@@ -367,7 +394,7 @@ func (s *ResourceGenerator) makeGatewayServiceClusters(
 			isRemote = !cfgSnap.Locality.Matches(services[svc][0].Node.Datacenter, services[svc][0].Node.PartitionOrDefault())
 		}
 
-		opts := gatewayClusterOpts{
+		opts := clusterOpts{
 			name:              clusterName,
 			hostnameEndpoints: hostnameEndpoints,
 			connectTimeout:    resolver.ConnectTimeout,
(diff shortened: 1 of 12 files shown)


agent/xds/clusters.go
```diff
@@ -907,10 +978,13 @@ type gatewayClusterOpts struct {
 
 	// hostnameEndpoints is a list of endpoints with a hostname as their address
 	hostnameEndpoints structs.CheckServiceNodes
+
+	// ipEndpoint is a singular ip/port endpoint
+	ipEndpoint structs.DestinationConfig
```
dhiaayachi, line 983: Not sure about the naming here, maybe `AddressEndpoint`?

agent/xds/clusters.go
```diff
@@ -1013,6 +1087,90 @@ func (s *ResourceGenerator) makeGatewayCluster(snap *proxycfg.ConfigSnapshot, op
...
+	if opts.connectTimeout <= 0 {
+		opts.connectTimeout = time.Duration(cfg.ConnectTimeoutMs) * time.Millisecond
+	}
+
+	cluster := &envoy_cluster_v3.Cluster{
+		Name:           opts.name,
+		ConnectTimeout: durationpb.New(opts.connectTimeout),
+
+		// Having an empty config enables outlier detection with default config.
+		OutlierDetection: &envoy_cluster_v3.OutlierDetection{},
```
dhiaayachi, line 1145: What's the `OutlierDetection` is needed for here?

agent/xds/clusters.go
```diff
@@ -1013,6 +1087,90 @@ func (s *ResourceGenerator) makeGatewayCluster(snap *proxycfg.ConfigSnapshot, op
...
+		},
+	}
+
+	return cluster
+}
+
+func getCommonDNSCacheConfiguration() *envoy_common_dynamic_forward_proxy_v3.DnsCacheConfig {
+	return &envoy_common_dynamic_forward_proxy_v3.DnsCacheConfig{
+		Name:            dynamicForwardProxyClusterDNSCacheName,
+		DnsLookupFamily: envoy_cluster_v3.Cluster_V4_ONLY,
```
dhiaayachi, line 1170: does that mean that it will only do IPv4 DNS lookup? We should support IPv6 as well

agent/xds/listeners.go
```diff
@@ -1121,10 +1154,19 @@ func (s *ResourceGenerator) makeFilterChainTerminatingGateway(
...
-	filterChain := &envoy_listener_v3.FilterChain{
-		FilterChainMatch: makeSNIFilterChainMatch(cluster),
-		Filters:          make([]*envoy_listener_v3.Filter, 0, 3),
-		TransportSocket:  transportSocket,
+	var filterChain *envoy_listener_v3.FilterChain
+	if dest != nil {
+		filterChain = &envoy_listener_v3.FilterChain{
+			FilterChainMatch: makeEndpointFilterChainMatch(cluster, dest),
+			Filters:          make([]*envoy_listener_v3.Filter, 0, 3),
+			TransportSocket:  transportSocket,
```
dhiaayachi, line 1162: I'm a bit confused about this, what would happen if we have let say a hostname as an Address. 
I think we will end up with duplicate filters, am I mistaken?

agent/xds/listeners.go
```diff
@@ -1436,6 +1508,13 @@ func makeSNIClusterFilter() (*envoy_listener_v3.Filter, error) {
 	return makeFilter("envoy.filters.network.sni_cluster", &envoy_sni_cluster_v3.SniCluster{})
 }
 
+func makeSNIDynamicForwardProxyFilter() (*envoy_listener_v3.Filter, error) {
+	return makeFilter("envoy.filters.network.sni_dynamic_forward_proxy", &envoy_sni_dynamic_forward_proxy_v3.FilterConfig{
+		DnsCacheConfig: getCommonDNSCacheConfiguration(),
+		PortSpecifier:  &envoy_sni_dynamic_forward_proxy_v3.FilterConfig_PortValue{PortValue: 443},
```
dhiaayachi, line 1514: Why the port is hardcoded to 443 here?

dhiaayachi (commented):
I added few comments/questions. Feel free to disregard those until we talk about it with @markan

agent/xds/clusters.go
```diff
@@ -385,6 +413,49 @@ func (s *ResourceGenerator) makeGatewayServiceClusters(
...
 }
 
+func (s *ResourceGenerator) makeDestinationClusters(cfgSnap *proxycfg.ConfigSnapshot) ([]proto.Message, error) {
+	var createDynamicForwardProxy bool
+	serviceConfigs := cfgSnap.TerminatingGateway.ServiceConfigs
+
+	clusters := make([]proto.Message, 0, len(cfgSnap.TerminatingGateway.DestinationServices))
+
+	for _, svcName := range cfgSnap.TerminatingGateway.ValidDestinations() {
+		svcConfig, ok := serviceConfigs[svcName]
```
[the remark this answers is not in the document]
DanStough: Yes. I kept the check here in case there were any weird inconsistency issues with the serviceConfigs updating in the background, but if you don't think that's an issue, I'm happy to remove.

agent/xds/clusters.go
DanStough: This was part of the existing `makeGatewayCluster`, so I thought it would be relevant to Destination clusters. Reading the [envoy docs](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/outlier.html?highlight=outlier%20detection) though, upstream ejection doesn't seem relevant in our usecase. I will remove.

agent/xds/listeners.go
DanStough: Do you mean duplicates like the "tcp_proxy" and "sni_dynamic_forward_proxy" filters? [All of the docs ](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/sni_dynamic_forward_proxy_filter)use both filters in tandem and I kept the RBAC one for TLS. Or are you talking about de-duplicating similar hostnames?

agent/xds/listeners.go
DanStough: :shame: I just missed it when copying over the configuration. I'll make sure this is the Destination Port.

agent/xds/listeners.go
```diff
@@ -1121,10 +1154,19 @@ func (s *ResourceGenerator) makeFilterChainTerminatingGateway(
...
 	}
 
-	filterChain := &envoy_listener_v3.FilterChain{
-		FilterChainMatch: makeSNIFilterChainMatch(cluster),
-		Filters:          make([]*envoy_listener_v3.Filter, 0, 3),
-		TransportSocket:  transportSocket,
+	var filterChain *envoy_listener_v3.FilterChain
+	if dest != nil {
+		filterChain = &envoy_listener_v3.FilterChain{
+			FilterChainMatch: makeEndpointFilterChainMatch(cluster, dest),
```
DanStough, line 1160:
```suggestion
			FilterChainMatch: makeDestinationFilterChainMatch(cluster, dest),
```

agent/xds/clusters.go
```diff
@@ -1013,6 +1087,87 @@ func (s *ResourceGenerator) makeGatewayCluster(snap *proxycfg.ConfigSnapshot, op
...
+		},
+	}
+
+	return cluster
+}
+
+func getCommonDNSCacheConfiguration() *envoy_common_dynamic_forward_proxy_v3.DnsCacheConfig {
+	return &envoy_common_dynamic_forward_proxy_v3.DnsCacheConfig{
+		Name:            dynamicForwardProxyClusterDNSCacheName,
+		DnsLookupFamily: envoy_cluster_v3.Cluster_V4_PREFERRED,
```
DanStough, line 1167: We should file an issue about making this configurable.

agent/xds/clusters.go
```diff
@@ -385,6 +413,49 @@ func (s *ResourceGenerator) makeGatewayServiceClusters(
...
+
+	for _, svcName := range cfgSnap.TerminatingGateway.ValidDestinations() {
+		svcConfig, ok := serviceConfigs[svcName]
+		if !ok {
+			continue
+		}
+		dest := svcConfig.Destination
+
+		// If IP, create a cluster with the fake name.
+		if ip := net.ParseIP(dest.Address); ip != nil {
```
DanStough, line 430: Use convenience method here.

agent/xds/listeners.go
```diff
@@ -1068,14 +1057,64 @@ func (s *ResourceGenerator) makeTerminatingGatewayListener(
...
+		clusterName := connect.ServiceSNI(svc.Name, "", svc.NamespaceOrDefault(), svc.PartitionOrDefault(), cfgSnap.Datacenter, cfgSnap.Roots.TrustDomain)
+
+		intentions := cfgSnap.TerminatingGateway.Intentions[svc]
+		svcConfig := cfgSnap.TerminatingGateway.ServiceConfigs[svc]
+
+		cfg, err := ParseProxyConfig(svcConfig.ProxyConfig)
+		if err != nil {
+			// Don't hard fail on a config typo, just warn. The parse func returns
+			// default config if there is an error so it's safe to continue.
+			s.Logger.Warn(
```
markan, line 1070: Do we warn anywhere earlier in the parse process?

agent/xds/listeners.go
```diff
@@ -1068,14 +1057,64 @@ func (s *ResourceGenerator) makeTerminatingGatewayListener(
...
 	// effect on how they operate, but it does mean that we won't churn
 	// listeners at idle.
 	sort.Slice(l.FilterChains, func(i, j int) bool {
-		return l.FilterChains[i].FilterChainMatch.ServerNames[0] < l.FilterChains[j].FilterChainMatch.ServerNames[0]
+		si := ""
+		sj := ""
+		if len(l.FilterChains[i].FilterChainMatch.PrefixRanges) > 0 {
+			si += l.FilterChains[i].FilterChainMatch.PrefixRanges[0].AddressPrefix +
+				"/" + l.FilterChains[i].FilterChainMatch.PrefixRanges[0].PrefixLen.String() +
+				":" + l.FilterChains[i].FilterChainMatch.DestinationPort.String()
```
markan, line 1102: I don't think this pattern works if AddressPrefix is ipv6. Need to do something like [::1]:8500 to be correct.

agent/xds/listeners.go
```diff
@@ -1068,14 +1057,64 @@ func (s *ResourceGenerator) makeTerminatingGatewayListener(
...
+		}
+		if len(l.FilterChains[i].FilterChainMatch.ServerNames) > 0 {
+			si += l.FilterChains[i].FilterChainMatch.ServerNames[0]
+		}
+
+		if len(l.FilterChains[j].FilterChainMatch.PrefixRanges) > 0 {
+			sj += l.FilterChains[j].FilterChainMatch.PrefixRanges[0].AddressPrefix +
+				"/" + l.FilterChains[j].FilterChainMatch.PrefixRanges[0].PrefixLen.String() +
```
markan, line 1110: Same ivp6 issue as above.

markan (commented):
A few ivp6 issues, but generally looks good.

agent/xds/clusters.go
DanStough: Per our discussion with Mark, going to make this AUTO to start with IPV6.

agent/xds/listeners.go
DanStough: There should be validation on the front end when the proxy config entry gets created: https://github.com/hashicorp/consul/blob/9c4798695f89cc48b8674fbd01555f6745708f74/agent/structs/config_entry.go#L361-L362

agent/xds/listeners.go
DanStough: This is an interesting callout because we added this work thinking we would support CIDR blocks, but since you can't pass this information from the connect proxy to the gateway, it wasn't going to work. Since even IP addresses have a fake SNI as `server_name`, turns out we don't even need any changes to this method.

agent/xds/listeners.go
DanStough: Same comment as above about no longer needing these changes.

DanStough:
Thank you @markan @dhiaayachi 🙏 !. Ready for re-review.

agent/xds/clusters.go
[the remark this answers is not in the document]
dhiaayachi: All of this run in the same thread so I don't think this check could ever happen. I think it's safe to remove it.

agent/xds/clusters.go
markan: Auto sounds reasonable to me.

markan (approved):

PR #13392, opened 2022-06-07
Update IntentionSourceIndex to handle Peers, add IntentionMatch tests for Peers

kisunji:
This PR is a follow up to https://github.com/hashicorp/consul/pull/13390

### Description
To confirm IntentionMatch works as intended, I added new tests to insert config entries referencing `SourceIntention.Peer` and then called IntentionMatch to see them synthesized and returned as `structs.Intention`.

The index for `indexSource` was to be updated to handle the new Peer field.

* Tests added as described above

### TODO
* after https://github.com/hashicorp/consul/pull/13390 is merged, rebase and point to main


agent/consul/state/intention_test.go
```diff
@@ -1199,6 +1199,131 @@ func TestStore_IntentionsList(t *testing.T) {
 	})
 }
 
+func TestStore_IntentionMatch_ConfigEntries(t *testing.T) {
```
kisunji, line 1202: At first I tried to re-jig `TestStore_IntentionMatch_table` to handle peers but found it a bit too rigid to work with, since it depends on reversing inputs to test source/destination match queries.

Hopefully this test can be the new matrix test after we remove legacy intentions.

agent/consul/state/config_entry_intention.go
```diff
@@ -101,8 +101,17 @@ func (s *ServiceIntentionSourceIndex) FromObject(obj interface{}) (bool, [][]byt
 
 	vals := make([][]byte, 0, len(ixnEntry.Sources))
 	for _, src := range ixnEntry.Sources {
-		sn := src.SourceServiceName()
-		vals = append(vals, []byte(sn.String()+"\x00"))
+		peer := src.Peer
+		if peer == "" {
+			peer = structs.LocalPeerKeyword
+		}
```
kisunji, line 107: we could simply choose not to write this index if there is a peer, but I thought we might eventually need to support it

freddygv (approved):

PR #13394, opened 2022-06-08
ui: upgrade ember composable helper

LevelbossMike:
### Description
Upgrade [ember-composable-helpers ](https://github.com/DockYard/ember-composable-helpers) to version 5.x. This version contains the `pick`-helper which makes composition in the template layer easier with Octane.

```hbs
{{!-- this is usually hard to do with Octane --}}
<input {{on "input" (pick "target.value" this.updateText)}} .../>
```

Version 5.x also fixes a regression with `sort-by` that according to @johncowen was the reason why the version was pinned to `4.0.0` at the moment.

Version 5 of `ember-composable-helpers` removes the `contains`-helper in favor of `includes` which I changed all occurences for.

Tests are passing locally.

### Links
[Fix for slow sort-by - PR](https://github.com/DockYard/ember-composable-helpers/pull/419)

### DIFF
--- a/.changelog/13394.txt
+++ b/.changelog/13394.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: upgrade ember-composable-helpers to v5.x
+```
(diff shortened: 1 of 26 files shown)


johncowen (commented):
Oh amazing, so excited for this to go in. There are a few helpers I've been wanting to use that are only "in later than 4" versions 🎉 

I spotted a couple more references to `contains` in the `packages/consul-*` folders (nspaces, lock-sessions, partitions). If we can fix those up also and we can get this in!

Oh, did you manage to check the performance of sorting on 10s of thousands of rows to make sure its as quick as the 4.0.0 sort-by?

johncowen (approved):
:shipit: thanks again for doing this

PR #13397, opened 2022-06-08
Fix intentions wildcard dest

dhiaayachi:
### Description
This fix an issue introduced in https://github.com/hashicorp/consul/pull/13354 which fail to retrieve intentions when using enterprise namespace and partitions wildcard.

### DIFF
--- a/agent/consul/state/config_entry_intention.go
+++ b/agent/consul/state/config_entry_intention.go
@@ -283,12 +283,11 @@ func readSourceIntentionsFromConfigEntriesForServiceTxn(
 		return nil, fmt.Errorf("failed config entry lookup: %s", err)
 	}
 	ws.Add(iter.WatchCh())
-
+	kind := structs.GatewayServiceKindService
 	for v := iter.Next(); v != nil; v = iter.Next() {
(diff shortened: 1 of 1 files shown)


agent/consul/state/config_entry_intention.go
```diff
@@ -288,10 +288,13 @@ func readSourceIntentionsFromConfigEntriesForServiceTxn(
...
 				entMeta := entry.DestinationServiceName().EnterpriseMeta
-				kind, err := GatewayServiceKind(tx, entry.DestinationServiceName().Name, &entMeta)
-				if err != nil {
-					return nil, err
+				if entMeta.NamespaceOrDefault() != acl.WildcardName && entMeta.PartitionOrDefault() != acl.WildcardName {
+					kind, err = GatewayServiceKind(tx, entry.DestinationServiceName().Name, &entMeta)
+					if err != nil {
+						return nil, err
+					}
 				}
```
kisunji, line 298: Can these be assigned outside of the for-loop? So we don't need to query per source

kisunji (changes requested):
minor comment

kisunji (approved):

PR #13401, opened 2022-06-08
Updated the default service name to lower case

noejbrown:
Changes to doc :
Changed default service_name to consul-terraform-sync. Service name is used in service registration it was previously proper case, it is now lower case.

Preview Link :
https://consul-puh2px6pp-hashicorp.vercel.app/docs/nia/configuration#service-registration


trujillo-adam (approved):
👍🏽

PR #13404, opened 2022-06-08
Update RBAC to handle imported services

kisunji:
### Description
When converting from Consul intentions to xds RBAC rules, services imported from other peers must encode additional data like partition (from the remote cluster) and trust domain.

This PR updates the PeeringTrustBundle to hold the sending side's local partition as `ExportedPartition`. It also updates RBAC code to encode SpiffeIDs of imported services with the `ExportedPartition` and `TrustDomain`.

* Will add more unit tests

### PR Checklist

* [x] updated test coverage
  * [ ] (stretch) add a peered L4 intention envoy integ test similar to `case-cross-peers`
* [x] not a security concern
* [ ] update enterprise side tests


agent/xds/rbac.go
```diff
@@ -518,18 +568,22 @@ func ixnSourceMatches(tester, against structs.ServiceName) bool {
 		return false
 	}
 
-	matchesAP := tester.PartitionOrDefault() == against.PartitionOrDefault() || against.PartitionOrDefault() == structs.WildcardSpecifier
```
kisunji, line 521: I removed `|| against.PartitionOrDefault() == structs.WildcardSpecifier` which isn't applicable since we validate against it

mkeeler (approved):
This all looks great. I have one comment about a useful optimization but it could also wait for a future PR so I am going ahead and approving.

agent/xds/rbac.go
```diff
@@ -395,8 +438,12 @@ func makeRBACRules(intentions structs.Intentions, intentionDefaultAllow bool, is
 
 	// TODO(banks,rb): Implement revocation list checking?
 
+	trustBundlesByPeer := make(map[string]*pbpeering.PeeringTrustBundle, len(peeringTrustBundles))
+	for _, ptb := range peeringTrustBundles {
+		trustBundlesByPeer[ptb.PeerName] = ptb
+	}
```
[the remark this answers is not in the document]
kisunji: That's a great point. I was wondering how often this codepath gets invoked; I'll have to check how @rboyer initially used the slice but assume it won't be too much of an effort

agent/rpc/peering/subscription_manager.go
```diff
@@ -451,8 +460,9 @@ func (m *subscriptionManager) subscribeCARoots(ctx context.Context, idx uint64,
 		updateCh <- cache.UpdateEvent{
 			CorrelationID: subCARoot,
 			Result: &pbpeering.PeeringTrustBundle{
-				TrustDomain: m.trustDomain,
-				RootPEMs:    rootPems,
+				TrustDomain:       m.trustDomain,
+				RootPEMs:          rootPems,
+				ExportedPartition: partition,
```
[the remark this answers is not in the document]
rboyer: Maybe leave a TODO here to revisit before GA for futureproofing ourselves.

agent/xds/rbac.go
```diff
@@ -188,11 +204,22 @@ func removePermissionPrecedence(perms []*rbacPermission, intentionDefaultAction
...
-		Source:     ixn.SourceServiceName(),
+		Source: rbacService{
+			ServiceName: ixn.SourceServiceName(),
+			peer:        ixn.SourcePeer,
+		},
 		Precedence: ixn.Precedence,
 	}
+
+	// imported services will have addition metadata used to override SpiffeID creation
+	// TODO(peering): is this true? Can there be a race condition where the bundle is not yet replicated before xds runs?
```
rboyer, line 217: You can require that the proxycfg have received the peer bundle here:
```
// Valid returns whether or not the snapshot has all required fields filled yet.
func (s *ConfigSnapshot) Valid() bool {
	switch s.Kind {
	case structs.ServiceKindConnectProxy:
		if s.Proxy.Mode == structs.ProxyModeTransparent && !s.ConnectProxy.MeshConfigSet {
			return false
		}
		return s.Roots != nil &&
			s.ConnectProxy.Leaf != nil &&
			s.ConnectProxy.IntentionsSet &&
			s.ConnectProxy.MeshConfigSet // <---- add something here
```
in agent/proxycfg/snapshot.go

agent/xds/rbac.go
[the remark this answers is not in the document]
rboyer: @freddygv only was using the slice by walking it, building a map of derived data keyed by trust domain, and then sorting the resulting slice after using the map by trust domain again (for SAN validation).

You could switch the proxycfg to store a map without issue. Usually the one thing when switching is verifying xds:
- does not need the relative ordering of elements
- will not suddenly generate different output every time b/c of random map iteration order

acpana (commented):
comments, learnings 📚

agent/xds/rbac.go
```diff
@@ -580,37 +634,52 @@ func idPrincipal(src structs.ServiceName) *envoy_rbac_v3.Principal {
...
-	if sourceNS == structs.WildcardSpecifier && sourceName != structs.WildcardSpecifier {
-		panic(fmt.Sprintf("not possible to have a wildcarded namespace %q but an exact service %q", sourceNS, sourceName))
+const anyPath = `[^/]+`
+
+func makeSpiffePattern(src rbacService) string {
+	var (
+		host = anyPath
+		ap   = src.PartitionOrDefault()
+		ns   = src.NamespaceOrDefault()
+		svc  = src.Name
```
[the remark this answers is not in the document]
kisunji: I don't really see a reason to - the prefix doesn't distinguish anything here and rather makes variables look like they have extra meaning

agent/xds/rbac.go
```diff
@@ -580,37 +634,52 @@ func idPrincipal(src structs.ServiceName) *envoy_rbac_v3.Principal {
 	}
 }
 
-func makeSpiffePattern(sourceAP, sourceNS, sourceName string) string {
-	if sourceNS == structs.WildcardSpecifier && sourceName != structs.WildcardSpecifier {
-		panic(fmt.Sprintf("not possible to have a wildcarded namespace %q but an exact service %q", sourceNS, sourceName))
+const anyPath = `[^/]+`
```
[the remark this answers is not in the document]
kisunji: At one point in my life I preferred grouping them at the top of a file but dnephin convinced me they are better left closer to their usages. No strong preference but since this isn't exported I think there's little value moving it to the top

agent/xds/rbac.go
kisunji: Just to clarify, do I need to call `Valid` in the codepaths? Or is the snapshot validated before it is sent to xds (and I just need to add an extra clause)

kisunji: Nvm I see it being called before the send

rboyer: Yeah that function is the gating function to even send a ConfigSnapshot _to_ the xds code.

agent/xds/rbac.go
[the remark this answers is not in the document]
rboyer: Another way to think of it is that "business logic" constants might make more sense at the top of a file, for visibility into arbitrary knobs we might want to tweak.

Plumbing constants like this one don't need visibility and thus can live closer to where they are used.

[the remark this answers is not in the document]
kisunji: Yeah if we're writing libraries with consts / enums / sentinel errors that users should know and/or easily discover, those are best grouped at the top of files (alternatively I've seen them defined in consts.go)

agent/xds/rbac.go
kisunji: Hm just realized adding it to `Valid` stops config snapshot from sending in non-peered contexts. I may have to add validation to error out if peer != nil but there's no corresponding trust bundle in the map

agent/xds/rbac.go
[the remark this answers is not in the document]
kisunji: Just realized we have a map after all in `ConfigSnapshotUpstreams` ([ref](https://github.com/hashicorp/consul/blob/70c93bda53d1b2f2b474a4037461ac81a5fdacb8/agent/proxycfg/snapshot.go#L53)).

agent/xds/rbac.go
```diff
@@ -237,9 +263,20 @@ func intentionActionFromString(s structs.IntentionAction) intentionAction {
...
 
+type rbacService struct {
+	structs.ServiceName
+
+	// peer, exportedPartition, and trustDomain are
+	// only applicable to imported services and are
+	// used to override SPIFFEID fields.
+	peer              string
+	exportedPartition string
+	trustDomain       string
```
rboyer, line 274:
```suggestion
	Peer              string
	ExportedPartition string
	TrustDomain       string
```

Given that you embedded `structs.ServiceName` this means that one of your 4 fields is exported and the others are not leading to some odd feeling struct declarations with mixed case.

This `rbacService` is an entirely internal construction which itself is not exported so it wouldn't hurt to just "export" the fields for consistency.

agent/xds/rbac.go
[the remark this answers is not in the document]
rboyer: The SPIFFE patterns only match on the source side anyway, so it isn't necessary. Also turns out when I wrote this function the first time I was aliasing the `ns := sourceNS` and `svc := sourceName` anyway for brevity so Chris just finished the job.

agent/connect/uri_service.go
```diff
@@ -35,3 +32,20 @@ func (id SpiffeIDService) URI() *url.URL {
...
+
+	// Although OSS has no support for partitions, it still needs to be able to
+	// handle exportedPartition from peered Consul Enterprise clusters in order
+	// to generate the correct SpiffeID.
+	// We intentionally avoid using pbpartition.DefaultName here to be OSS friendly.
+	if ap := id.PartitionOrDefault(); ap != "" && ap != "default" {
+		return "/ap/" + ap + path
+	}
+	return path
+}
```
kisunji, line 51: Sanity check: is there a reason we don't want /ap/ in OSS, side from "partitions are enterprise-only"? Wondering if there are side effects

rboyer: Backwards compatibility. We defined the `/ap` prefix as "only present if not the default"

agent/xds/rbac.go
```diff
@@ -237,9 +269,20 @@ func intentionActionFromString(s structs.IntentionAction) intentionAction {
 	return intentionActionDeny
 }
 
+type rbacService struct {
+	structs.ServiceName
+
+	// peer, exportedPartition, and trustDomain are
```
rboyer, line 275:
```suggestion
	// Peer, ExportedPartition, and TrustDomain are
```

agent/xds/rbac.go
```diff
@@ -580,37 +637,52 @@ func idPrincipal(src structs.ServiceName) *envoy_rbac_v3.Principal {
...
 }
 
-func makeSpiffePattern(sourceAP, sourceNS, sourceName string) string {
-	if sourceNS == structs.WildcardSpecifier && sourceName != structs.WildcardSpecifier {
-		panic(fmt.Sprintf("not possible to have a wildcarded namespace %q but an exact service %q", sourceNS, sourceName))
+const anyPath = `[^/]+`
+
+func makeSpiffePattern(src rbacService) string {
+	var (
+		host = anyPath
```
rboyer, line 644:
```suggestion
		host = anyPath // TrustDomain is not verified by RBAC, so we match on any value.
```
or something up here for now

proto/pbpeering/peering.proto
```diff
@@ -108,12 +108,16 @@ message PeeringTrustBundle {
...
-  // Partition isolates the bundle from other trust bundles in separate partitions.
+  // Partition isolates the bundle from other trust bundles in separate local partitions.
   string Partition = 3;
 
   // RootPEMs holds ASN.1 DER encoded X.509 certificate data for the trust bundle.
   repeated string RootPEMs = 4;
 
+  // ExportedPartition references the remote partition of the peer
+  // which sent this trust bundle. Used for generating SpiffeIDs.
+  string ExportedPartition = 7;
```
rboyer, line 119: Feel free to renumber these to make the fields have the right order with respect to the numbers. We didn't ship this to GA yet.

agent/xds/rbac_test.go
```diff
@@ -340,17 +362,17 @@ func TestRemoveIntentionPrecedence(t *testing.T) {
 				},
 			},
 		},
-		"default-deny-allow-all-and-path-deny": {
```
rboyer, line 343: What happened to the prior case? Can you restore it?

kisunji: oh good catch

kisunji: This was just a rename blunder

agent/xds/rbac.go
kisunji: Added a todo to use local trust domain

rboyer (approved):
LGTM pending passing tests.

Issue #12686, opened 2022-04-04
Configure custom metadata on gRPC healthchecks


julb:
#### Feature Description

When configuring gRPC healthchecks, it might be useful to add metadata to the gRPC request, similarly to what we can do for HTTP healthchecks with HTTP headers.

Considering the example retrieved from your docs with added fields:

```
check = {
  id = "mem-util"
  name = "Service health status"
  grpc = "127.0.0.1:12345/my_service"
  grpc_use_tls = true
+  grpc_metadata = {
+    user-agent =  "xxxx"
+  }
  interval = "10s"
}
```

#### Use Case(s)

It can be used to:
* Set specific metadata keys to identify that the request is coming from Consul
* Set specific tracing information

julb:
@Amier3 hey, I just saw this "help-wanted" tag.
Are you open to PR for this change ?

fulviodenza:
I follow this topic because I followed some health-check topics and could help if needed

blake:
Hi @julb, yes we are open to a community contribution for this change.

fulviodenza:
Hey @julb , are you working on the pr? If not I can work on it.

julb:
Hey @fulviodenza, go ahead, I don't have time to work on it now.
Thanks a lot for your help.

fulviodenza:
Hi! Sorry for the delay, I'll start working on this in the next few days.

PR #13408, opened 2022-06-08
Configure custom metadata on gRPC healthchecks

fulviodenza:
### Description
This PR implements the feature requested in #12686 issue, the custom metadata on gRPC healthchecks

All grpc healthcheck tests were featured with a new field called header, it consists of a `map[string][]string` (it's the field shared with HTTP healthcheck).
 
### Links
[Issue referred](https://github.com/hashicorp/consul/issues/12686)

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/checks/grpc.go
```diff
@@ -49,17 +50,26 @@ func NewGrpcHealthProbe(target string, timeout time.Duration, tlsConfig *tls.Con
...
 	ctx, cancel := context.WithTimeout(context.Background(), probe.timeout)
 	defer cancel()
 
+	var headerMD map[string]string = make(map[string]string)
+	for k, v := range header {
+		headerMD[k] = strings.Join(v, ", ")
+	}
+
+	md := metadata.New(headerMD)
+	ctx = metadata.NewOutgoingContext(context.Background(), md)
```
Amier3: Hey @fulviodenza 

Thank you for taking up the `help-wanted` issue, I think a lot of people will appreciate this change! We're fairly close to a release so i'm not sure when we'll be able to get a full review going -- but I did see one of the lint tests failed on this line for ineffectual assignment.

Merging the context timeout line (56) with this one should fix this: 

`ctx, cancel := context.WithTimeout(metadata.NewOutgoingContext(context.Background(), md), probe.timeout)`

fulviodenza:
Hi @Amier3! Any news about this PR?

agent/checks/grpc.go
```diff
@@ -49,17 +50,25 @@ func NewGrpcHealthProbe(target string, timeout time.Duration, tlsConfig *tls.Con
...
+	md := metadata.New(headerMD)
+
+	ctx, cancel := context.WithTimeout(metadata.NewOutgoingContext(context.Background(), md), probe.timeout)
 	defer cancel()
 
 	connection, err := grpc.DialContext(ctx, serverWithScheme, probe.dialOptions...)
 	if err != nil {
 		return err
 	}
+
```
huikang, line 71: nit: remove the empty line

agent/checks/grpc.go
```diff
@@ -49,17 +50,25 @@ func NewGrpcHealthProbe(target string, timeout time.Duration, tlsConfig *tls.Con
...
 
-	ctx, cancel := context.WithTimeout(context.Background(), probe.timeout)
+	var headerMD map[string]string = make(map[string]string)
+	for k, v := range header {
+		headerMD[k] = strings.Join(v, ", ")
+	}
+
+	md := metadata.New(headerMD)
+
+	ctx, cancel := context.WithTimeout(metadata.NewOutgoingContext(context.Background(), md), probe.timeout)
```
huikang, line 64: Can we conditionally run line 57 to line 64?

```go
if len(header) > 0 {
    // ...
    ctx, cancel := context.WithTimeout(metadata.NewOutgoingContext(context.Background(), md), probe.timeout)
}
```

agent/checks/grpc.go
```diff
@@ -49,17 +50,25 @@ func NewGrpcHealthProbe(target string, timeout time.Duration, tlsConfig *tls.Con
 
 // Check if the target of this GrpcHealthProbe is healthy
 // If nil is returned, target is healthy, otherwise target is not healthy
-func (probe *GrpcHealthProbe) Check(target string) error {
+func (probe *GrpcHealthProbe) Check(target string, header map[string][]string) error {
 	serverAndService := strings.SplitN(target, "/", 2)
 	serverWithScheme := fmt.Sprintf("%s:///%s", resolver.GetDefaultScheme(), serverAndService[0])
 
-	ctx, cancel := context.WithTimeout(context.Background(), probe.timeout)
+	var headerMD map[string]string = make(map[string]string)
```
huikang, line 57:
```suggestion
	headerMD := map[string]string{}
```

huikang (changes requested):
Nice job @fulviodenza ! Could you add the following to the PR to make it complete

1. add a changelog to describe briefly the new feature
2. update the doc (https://github.com/hashicorp/consul/blob/d457d8b6ce5cf825224ea318db5e3415c7083877/website/content/docs/discovery/checks.mdx) with some example

agent/checks/grpc.go
fulviodenza: Hi, huikang! Thank you for your review! I think this should be changed because `ctx` is declared in that line and we need it in the next lines, for the `cancel` variable, instead, we need to call it in the `defer cancel() `, so I think this condition should go from line 57 to line 65 and declaring the `ctx` outside the scope of the `if len(header) > 0 ` block

fulviodenza:
I'm not sure about this:
In the documentation at this page https://grpc.io/docs/what-is-grpc/core-concepts/ is written that 

> Keys are case insensitive and consists of ASCII letters, digits, and special characters -, _, . and must not be start with grpc- (which are reserved for gRPC itself).

There should be a validation on consul side when metadata are inserted?

agent/checks/grpc.go
huikang: I think we can remove this extraneous line.

agent/checks/grpc.go
huikang: Anything wrong with this suggestion?

agent/checks/grpc.go
huikang: Thanks. By looking at the `grpc/metadata` package I think we can do the following

```go
ctx, cancel := context.WithTimeout(context.Background(), probe.timeout)

if len(header) > 0 {
    // line 57 - line 62
    // headerMD := map[string]string{}
    // we might need to change MD to kv...

    ctx = metadata.AppendToOutgoingContext(ctx, kv...) context.Context](https://pkg.go.dev/google.golang.org/grpc/metadata#AppendToOutgoingContext)
}
```
This way we can both conditionally add header to metadata, and reuse `ctx, and cancel`. WDTY

huikang:
> 

> There should be a validation on consul side when metadata are inserted?

I don't think there is a need to validate this; we can rely on grpc's returned results and surface the error to user.
(Be better if grpc-metada provides a method to validate the header; however I couldn't find any.)

agent/checks/grpc.go
```diff
@@ -49,11 +50,23 @@ func NewGrpcHealthProbe(target string, timeout time.Duration, tlsConfig *tls.Con
...
 
 	ctx, cancel := context.WithTimeout(context.Background(), probe.timeout)
+	if len(header) > 0 {
+		var headerMD map[string]string = map[string]string{}
+		var kv []string
+
+		for k, v := range header {
+			headerMD[k] = strings.Join(v, ", ")
+			kv = append(kv, k)
+			kv = append(kv, headerMD[k])
```
huikang, line 65: nit: I don't think we need `headerMD`. How about change it to `kv = append(kv, k, strings.Join(v, ", "))`

huikang (commented):
Could you add a change log file, named `.changelog/13408.txt`; there are many examples under `.changelog`. Thanks.

fulviodenza:
> Hi, @fulviodenza, could you resolve the conflict in the PR? Thanks.

Hi @huikang , sorry for the devastating delay but I was on holiday in that period and github notifications were too much when I came back. Anyway I fixed the conflicts, now it should work. Let me know if there's anything other I can do

PR #13409, opened 2022-06-09
ui: Fix intl keys in order to render correct messages for empty states

johncowen:
### Description

Some of our i18n intl keys were wrong for our empty states. As missing intl keys default to show the last part of the key capitalised (`dc.nodes.sessions.empty.header` => `Header`). This resulted in some empty states showing `Header` and `Body` instead of the actual text.

we initially added the "capitalise the last part of the missing key as a default/missing piece of text" thing to help machine > human word replacement for lots of small pieces of text that were always going to be the last part of the key but capitalized, which seemed like a good approach at the time 😬 . This unfortunately means we don't notice when we don't type the intl key correctly.

So additionally here, I've added a dev time only log message (using `runInDebug` so it doesn't compile in) so we can go through and explicitly make sure every key has a corresponding piece of text. After which, longer term we can remove this "defaulting" and hard error instead, so this can never happen again.

Further replacement of the keys that we currently expect to "default" to have explicit corresponding text, will be done in a further PR.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-ui/app/utils/intl/missing-message.js
+++ b/ui/packages/consul-ui/app/utils/intl/missing-message.js
@@ -1,3 +1,4 @@
+/* eslint no-console: ["error", { allow: ["debug"] }] */
 import { runInDebug } from '@ember/debug';
 
 // if we can't find the message, take the last part of the identifier and
--- a/ui/packages/consul-ui/app/utils/intl/missing-message.js
+++ b/ui/packages/consul-ui/app/utils/intl/missing-message.js
@@ -1,6 +1,11 @@
+import { runInDebug } from '@ember/debug';
+
 // if we can't find the message, take the last part of the identifier and
 // ucfirst it so it looks human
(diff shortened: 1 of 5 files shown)


ui/packages/consul-ui/app/utils/intl/missing-message.js
```diff
@@ -1,6 +1,12 @@
+/* eslint no-console: ["error", { allow: ["debug"] }] */
+import { runInDebug } from '@ember/debug';
+
 // if we can't find the message, take the last part of the identifier and
 // ucfirst it so it looks human
 export default function missingMessage(key, locales) {
+  runInDebug(
+    _ => console.debug(`Translation key not found: ${key}`)
```
evrowe, line 8: 🙌🏻

evrowe (approved):
Looks good! Solid explanation of the problem in the PR description as well 👍

PR #9639, opened 2021-01-26
Add UpgradeConfig to HTTP Connection Manager

Legogris:
This should solve #8283 

Credit goes to @blake , I simply rebased his commit on master with small changes to get CI passing to get this PR considered for merging.

Issue #8283, opened 2020-07-10
Support HTTP connection protocol upgrades thru Envoy


blake:
#### Feature Description

HTTP supports a [protocol upgrade mechanism](https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism) which allows upgrading existing connections to a different protocol such as WebSocket. Envoy supports [WebSocket and HTTP upgrades](https://www.envoyproxy.io/docs/envoy/v1.14.3/intro/arch_overview/http/websocket), however this is not enabled by default. It requires providing an [upgrade_config](https://www.envoyproxy.io/docs/envoy/v1.14.3/api-v2/config/filter/network/http_connection_manager/v2/http_connection_manager.proto#envoy-api-field-config-filter-network-http-connection-manager-v2-httpconnectionmanager-upgrade-configs) to the HTTP Connection Manger (see https://github.com/hashicorp/consul/commit/fd993ccebb37ebdf5c08450da44a5ce8e22c6b30 for example).

This setting is not configurable in Consul as of 1.8.0, and because of this it is not possible to establish WebSocket connections through either ingress gateways or sidecar proxies when the service's configured [protocol](https://www.consul.io/docs/agent/config-entries/service-defaults.html#protocol) is `http`.

```shell
curl 'http://dashboard.ingress.consul/socket.io/?EIO=3&transport=websocket' \
  --include \
  --header 'Origin: http://dashboard.ingress.consul' \
  --header 'Connection: keep-alive, Upgrade' \
  --header 'Upgrade: websocket' \
  --header 'Sec-WebSocket-Version: 13' \
  --header 'Sec-WebSocket-Key: O1b2cWbnz4bk8LyYKDiE3Q=='

HTTP/1.1 403 Forbidden
date: Wed, 08 Jul 2020 06:29:45 GMT
server: envoy
connection: close
content-length: 0
```
It would be useful if Consul either enables WebSocket connection upgrades by default, or provides a way for operators to enable this via configuration.

#### Use Case(s)

Enables any application utilizing WebSockets to work through Ingress gateways or sidecar proxies.

Below is a successful connection upgrade request after deploying a custom version of Consul which configures Envoy with the appropriate upgrade config.

```shell
curl 'http://dashboard.ingress.consul/socket.io/?EIO=3&transport=websocket' \
  --include \
  --header 'Origin: http://dashboard.ingress.consul' \
  --header 'Connection: keep-alive, Upgrade' \
[...]

PR #13418, opened 2022-06-10
[DRAFT] [NEED FEEDBACK] Support websocket through envoy

davide-baldo:
This PR aim to implement websocket support, referenced here https://github.com/hashicorp/consul/issues/8283 .

Rather than always enable the websocket support for everything (like in this old PR https://github.com/hashicorp/consul/pull/9639), I've created a new `websocket` parameter which can be used wherever `protocol` can be used, usually when you specify the service protocol of the service or when you override the path.
In this patch, websocket it's disabled by default.

Basic unit testing was added but I couldn't validate them since i couldn't perform a clean test run with intellij, make test or make test-docker.
I've manually tested in a 2 node server installation and envoy configuration gets created correctly and it's routing correctly.

Since this is the first time I approach consul codebase i ask, if possible, for guidance before committing more time, especially if the overall approach is correct.

### DIFF
--- a/agent/config/builder.go
+++ b/agent/config/builder.go
@@ -1794,6 +1794,7 @@ func (b *builder) pathsVal(v []ExposePath) []structs.ExposePath {
 			Path:          stringVal(p.Path),
 			LocalPathPort: intVal(p.LocalPathPort),
 			Protocol:      stringVal(p.Protocol),
+			Websocket:     boolVal(p.Websocket),
 		}
 	}
 	return paths
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -581,6 +581,9 @@ type ExposePath struct {
 	// Protocol describes the upstream's service protocol.
 	Protocol *string `mapstructure:"protocol"`
 
+	// Websocket whether websocket upgrade is supported.
+	Websocket *bool `mapstructure:"websocket"`
+
 	// LocalPathPort is the port that the service is listening on for the given path.
 	LocalPathPort *int `mapstructure:"local_path_port"`
 }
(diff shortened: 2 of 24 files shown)


davide-baldo:
Hi @Amier3,

Unfortunately the [config guide](https://github.com/hashicorp/consul/blob/main/docs/config/checklist-adding-config-fields.md) doesn't cover the `service-defaults` and apply only partially (I've added `websocket` attribute in the exposed paths in proxy configuration as well).

CLI and struct changes and envoy generation should be complete AFAIK.

 - [ ] Tests
 - [ ] PR Changelog
 - [ ] Docs
      - [ ] Update docs in `website/source/docs/agent/services.html.md`
      - [ ] Consider if it's worth adding examples to feature docs or API docs
        that show the new field's usage.

davide-baldo:
just rebased against lastest main branch.
@Amier3 Could you give me some feedback on whether this is the right approach?
thanks,

simonctrlz:
@davide-baldo how do you deal with this requirement internally? do you maintain a fork with your patches?
Embarrassing show on Hashicorp's part on that front, really.

davide-baldo:
@simonctrlz 
If you don't need any routing/filter you can proxy websocket as TCP, if you need more control you need to patch consul (and rebuild it, luckily in go it's really easy to compile/cross-compile) in order to generate the right envoy configuration.

the smallest patch you can apply is this: https://github.com/hashicorp/consul/pull/9639/files#diff-3f5a22a74e450c254308e27dccb905bdce627f0a29ec602ca66df792d1dbf5ce 

```go
		UpgradeConfigs: []*envoyhttp.HttpConnectionManager_UpgradeConfig{
			{
				UpgradeType: "websocket",
			},
		},
```

This PR is mainly about enabling/disabling websocket capabilities through service configuration.
You **might** also find a way to modify envoy configuration directly from the consul proxy configuration to avoid any consul patch, but it's undocumented I don't know how to do it, or even if it's possible.

PR #7999, opened 2020-06-02
Enable gateways to resolve hostnames to IPv4 addresses

freddygv:
This PR enables Terminating Gateways and Mesh Gateways to resolve hostnames.

The DNS resolution will be handled by Envoy and defaults to [LOGICAL_DNS](https://www.envoyproxy.io/docs/envoy/v1.14.1/intro/arch_overview/upstream/service_discovery#logical-dns). This discovery type can be overridden on a per-gateway basis with the `envoy_dns_discovery_type` [Gateway Option](https://github.com/hashicorp/consul/blob/21ff48eca797d0926f9a980e3fb95c8f95e2f7ef/website/pages/docs/connect/proxies/envoy.mdx#gateway-options).

If a service contains an instance with a hostname as an address we set the Envoy cluster to use DNS as the discovery type rather than EDS. Since both mesh gateways and terminating gateways route to clusters using SNI, whenever there is a mix of hostnames and IP addresses associated with a service we use the hostname + CDS rather than the IPs + EDS.

Note that we detect hostnames by attempting to parse the service instance's address as an IP. If it is not a valid IP we assume it is a hostname.

PR #13422, opened 2022-06-10
xds: allow for peered upstreams to use tagged addresses that are hostnames

rboyer:
### Description

Mesh gateways [can use hostnames in their tagged addresses](https://github.com/hashicorp/consul/pull/7999/files). This is useful if you were to expose a mesh gateway using a cloud networking load balancer appliance that gives you a DNS name but no reliable static IPs.

Envoy cannot accept hostnames via EDS and those must be configured using CDS. There was already logic when configuring gateways in other locations in the code, but given the illusions in play for peering the downstream of a peered service wasn't aware that it should be doing that.

Also:
- ensuring that we always try to use wan-like addresses to cross peer boundaries.

### Links

- [Asana - wan hostnames](https://app.asana.com/0/1201832349181086/1202428211194184/f)
- [Asana - wan addr](https://app.asana.com/0/0/1202428211194183/f)


agent/proxycfg/testing_peering.go
```diff
@@ -51,6 +51,16 @@ func TestConfigSnapshotPeering(t testing.T) *ConfigSnapshot {
 							Service: "payments-sidecar-proxy",
```
rboyer, line 51: There are two cross-peer upstreams here and I've only adjusted one of them so we can test both scenarios at once.

agent/xds/endpoints.go
```diff
@@ -96,14 +96,20 @@ func (s *ResourceGenerator) endpointsFromSnapshotConnectProxy(cfgSnap *proxycfg.
...
+
 		endpoints, ok := cfgSnap.ConnectProxy.PeerUpstreamEndpoints[uid]
 		if ok {
 			la := makeLoadAssignment(
 				clusterName,
 				[]loadAssignmentEndpointGroup{
 					{Endpoints: endpoints},
 				},
-				cfgSnap.Locality,
+				GatewayKey{ /*empty so it never matches*/ },
```
rboyer, line 112: For the EDS cases, this ensures we use wan-like addresses.

agent/xds/clusters.go
```diff
@@ -602,6 +596,35 @@ func (s *ResourceGenerator) makeUpstreamClusterForPeerService(
...
+
+		useEDS := true
+		if _, ok := cfgSnap.ConnectProxy.PeerUpstreamEndpointsUseHostnames[uid]; ok {
+			useEDS = false
+		}
+
+		// If none of the service instances are addressed by a hostname we
+		// provide the endpoint IP addresses via EDS
+		if useEDS {
+			c.ClusterDiscoveryType = &envoy_cluster_v3.Cluster_Type{Type: envoy_cluster_v3.Cluster_EDS}
```
kisunji, line 608: nit: this is already set in L583

kisunji (approved):

agent/xds/clusters.go
```diff
@@ -1054,17 +1077,39 @@ func (s *ResourceGenerator) makeGatewayCluster(snap *proxycfg.ConfigSnapshot, op
...
+	return cluster
+}
+
+func configureClusterWithHostnames(
+	logger hclog.Logger,
+	cluster *envoy_cluster_v3.Cluster,
+	dnsDiscoveryType string,
+	// hostnameEndpoints is a list of endpoints with a hostname as their address
+	hostnameEndpoints structs.CheckServiceNodes,
+	// isRemote determines whether the cluster is in a remote DC and we should prefer a WAN address
```
freddygv, line 1100:
```suggestion
	// isRemote determines whether the cluster is in a remote DC or partition and we should prefer a WAN address
```

freddygv (approved):
LGTM, one minor suggestion

PR #13432, opened 2022-06-13
[docs] update terminating gateway docs for trust store path

kschoche:
### Description

With Consul Helm chart `0.44.0` the default envoy proxy image is no longer based on the alpine base image, the trust store location is changed in this base image and needs to be updated in our docs when specifying the config-entry.

This should help users avoid situations like: https://github.com/hashicorp/consul-k8s/issues/1262

### Links

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/website/content/docs/k8s/connect/terminating-gateways.mdx
+++ b/website/content/docs/k8s/connect/terminating-gateways.mdx
@@ -205,13 +205,15 @@ metadata:
 spec:
   services:
     - name: example-https
-      caFile: /etc/ssl/cert.pem
+      caFile: /etc/ssl/certs/ca-certificates.crt
 ```
 
 </CodeBlockConfig>
 
-~> If TLS is enabled a `caFile` must be provided, it must point to the system trust store of the terminating gateway
-container (`/etc/ssl/cert.pem`).
+~> If TLS is enabled a `caFile` must be provided, it must point to the system trust store of the terminating gateway container (`/etc/ssl/certs/ca-certificates.crt`).
+
+~> If TLS is enabled and you are using Consul Helm chart 0.43 or older, or an Envoy image with an alpine base
+image, use the following location for the ca certificate file `caFile: /etc/ssl/cert.pem`.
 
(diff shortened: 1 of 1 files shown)


website/content/docs/k8s/connect/terminating-gateways.mdx
```diff
@@ -205,13 +205,15 @@ metadata:
...
 ```
 
 </CodeBlockConfig>
 
-~> If TLS is enabled a `caFile` must be provided, it must point to the system trust store of the terminating gateway
-container (`/etc/ssl/cert.pem`).
+~> If TLS is enabled a `caFile` must be provided, it must point to the system trust store of the terminating gateway container (`/etc/ssl/certs/ca-certificates.crt`).
+
+~> If TLS is enabled and you are using Consul Helm chart 0.43 or older, or an Envoy image with an alpine base
+image, use the following location for the ca certificate file `caFile: /etc/ssl/cert.pem`.
```
trujillo-adam, line 216:
```suggestion
If TLS is enabled, you must include the `caFile` parameter  that points to the system trust store of the terminating gateway container. By default, the trust store is located in the `/etc/ssl/certs/ca-certificates.crt` directory.

Configure the `caFile` parameter to point to the `/etc/ssl/cert.pem` directory if TLS is enabled and you are using one of the following components:
 * Consul Helm chart 0.43 or older 
 * Or an Envoy image with an alpine base image
```
I don't think these need to be in callout boxes. They seem to be directly related to the topic being described. 
I also thought the second paragraph might be easier to parse as a list, but feel free to ignore that suggestion.

trujillo-adam (commented):
A few suggestions

website/content/docs/k8s/connect/terminating-gateways.mdx
david-yu: I agree would be good to call this out in a regular paragraph instead of a callout box.

david-yu (approved):

PR #13435, opened 2022-06-14
peering: intentions list test

acpana:
### Description
This diff adds a test to prove that the http api can list intentions for peered services. 

#### links

* [asana](https://app.asana.com/0/1201631904755166/1202439043013562/f)

### DIFF
--- a/agent/intentions_endpoint_test.go
+++ b/agent/intentions_endpoint_test.go
@@ -62,9 +62,10 @@ func TestIntentionList(t *testing.T) {
 			ids = append(ids, reply)
 		}
 
+		// set up an intention for a peered service
 		// TODO(peering): when we handle Upserts, we can use the for loop above. But it may be that we
 		// rip out legacy intentions before supporting that use case so run a config entry request instead here.
-		testutil.RunStep(t, "set up peered intention", func(t *testing.T) {
+		{
 			configEntryIntention := structs.ServiceIntentionsConfigEntry{
 				Kind: structs.ServiceIntentions,
 				Name: "bar",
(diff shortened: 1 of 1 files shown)


agent/intentions_endpoint_test.go
```diff
@@ -62,6 +62,35 @@ func TestIntentionList(t *testing.T) {
 			ids = append(ids, reply)
 		}
 
+		// TODO(peering): when we handle Upserts, we can use the for loop above. But it may be that we
+		// rip out legacy intentions before supporting that use case so run a config entry request instead here.
+		testutil.RunStep(t, "set up peered intention", func(t *testing.T) {
```
kisunji, line 67: this shouldn't be in a subtest; it's setup code that belongs in the same level as the writes above. You could alternatively wrap it in braces to distinguish the setup from the actual assertion tests
```
// SETUP
{
...
}
```

agent/intentions_endpoint_test.go
```diff
@@ -71,21 +100,23 @@ func TestIntentionList(t *testing.T) {
 		require.NoError(t, err)
 
 		value := obj.(structs.Intentions)
-		require.Len(t, value, 4)
+		require.Len(t, value, 5)
 
-		require.Equal(t, []string{"bar->db", "foo->db", "zim->gir", "*->db"},
+		require.Equal(t, []string{"bar->db", "foo->db", "zim->gir", "peered->bar", "*->db"},
```
kisunji, line 105: We probably want an assertion for the `Peer` field as well

kisunji (commented):
Some minor comments but looks good!

agent/intentions_endpoint_test.go
acpana: ✅

kisunji (approved):

PR #13438, opened 2022-06-14
Add server local blocking queries and watches

mkeeler:
### Description
For some upcoming work we have a need to watch the local memdb in a manner similar to how the agent Cache is used today.

This PR implements that mechanism.

* Unit Tests

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/consul/watch/server_local.go
```diff
@@ -0,0 +1,277 @@
...
+//
+// Similar to the older method:
+//
+// 1. Errors returned from the query will be propagated back to the caller.
+//
+// The query function must follow these rules:
+//
+//   1. To access data it must use the passed in StoreType (which will be a state.Store when
+//      everything gets stiched together outside of unit tests).
+//   2. It must return an index greather than the minIndex if the results returned by the query
```
boxofrad, line 68:
```suggestion
//   2. It must return an index greater than the minIndex if the results returned by the query
```

agent/consul/watch/server_local.go
```diff
@@ -0,0 +1,277 @@
...
+		if err := ws.WatchCtx(ctx); err != nil {
+			// exit if the context was cancelled
+			return index, result, nil
+		}
+
+		select {
+		case <-state.AbandonCh():
+			return index, result, nil
+		default:
+		}
```
boxofrad, line 168: The first time I read this, I thought that abandoning the state store would cause the above `WatchCtx` to block forever (so we'd never reach this `select`). I now realise that it's fine because we add the `AbandonCh` to the watchset on L106. Would you mind adding a comment to point this out for future readers? 😄

agent/consul/watch/server_local_test.go
```diff
@@ -0,0 +1,424 @@
...
+		})
+
+	err := ServerLocalNotify(ctx, "test", provider.getStore, provider.query, provider.notify)
+	require.NoError(t, err)
+
+	// wait for the context cancellation which will happen when the "query" func is run the third time
+	<-ctx.Done()
+	// wait another 100ms so that when the mock expectations are checked we would have had time to pick
+	// up on any extra invocations if the notification wasn't cancelled successfully.
+	time.Sleep(100 * time.Millisecond)
```
boxofrad, line 367: Totally not a comment on your approach here, but I really wish there were a way to achieve this without sleeping, that didn't introduce extra complexity (like threading in a done channel) 🤔

boxofrad (approved):
This is great, nice job! 👏🏻 🎉

agent/consul/watch/server_local.go
mkeeler: Sure. I will add a comment both where we call WatchCtx and where the abandon chan is added to the watch set

agent/consul/watch/server_local_test.go
mkeeler: I threaded through a done function (for package internal use only) and then modified the tests to pass that in and wait for the cancellation to finish.

PR #13444, opened 2022-06-14
connect: Use Envoy 1.22.2 instead of 1.22.1

eculver:
### Description
When I originally updated the latest versions in #13431, the latest version was 1.22.1. Since then, Envoy has released 1.22.2. It was really confusing for a bit. This updates to use the latest latest. 

Note: I originally wrote 1.22.2 in the changelog entry by mistake, but it turns out that's what we want 🤷🏼 .

CI will tell us.

### Links
[CTIA-31](https://hashicorp.atlassian.net/browse/CTIA-31).


agent/xds/envoy_versioning_test.go
```diff
@@ -136,8 +136,8 @@ func TestDetermineSupportedProxyFeaturesFromString(t *testing.T) {
 	for _, v := range []string{
 		"1.19.0", "1.19.1", "1.19.2", "1.19.3", "1.19.4", "1.19.5",
 		"1.20.0", "1.20.1", "1.20.2", "1.20.3", "1.20.3",
-		"1.21.1", "1.21.2", "1.21.3",
-		"1.22.0", "1.22.1",
+		"1.21.0", "1.21.2", "1.21.3", "1.21.4",
```
rboyer, line 139: where did `1.21.1` go in the test?

rboyer (approved):

agent/xds/envoy_versioning_test.go
eculver: oh whoops.

rboyer (approved):

PR #13447, opened 2022-06-14
feat: check mesh write acl for destinations

DanStough:
### Description
We are adding support for Destinations (nodeless egress services) in Consul Terminating Gateways. This PR adds checks ACL enforcement for `mesh:write` on `service-defaults` with a `destination`. This includes modifying an existing destination.

Test cases included for `service-default` authz. I didn't include RPC testing because I think that serves as more of a plumbing check.

### PR Checklist

* [X] updated test coverage
* [X] not a security concern
  * the goal of the PR is to add a higher level of enforcement to certain config entry kinds. The relatively safe assumption is that `mesh:write` policies, an operator-level permission with no resource scoping, is also allowed to create `service-defaults`
  * We discussed privileged escalation by `service:write` tokens creating a catalog service with the same name as the destination. In this case the service takes precedence over the destination in terms of routing, so there is no way the unprivileged user could abuse the mesh destination.


agent/structs/config_entry.go
```diff
@@ -232,15 +232,36 @@ func validateEndpointAddress(address string) error {
 	return nil
 }
 
+// CanRead allows authorization for ServiceConfigEntries on either Service-Read, Mesh Read,
+// or Operator Read since Destinations are an operator-level construct
```
DanStough, line 236: I was messing with this method, no longer needed.

kyhavlov (approved):
Everything here LGTM 👍

DanStough:
After talking internally about this, we're going to stick with the `service` ACL permissions for `service-defaults` that contain a destination.

PR #13451, opened 2022-06-15
peering: block Intention.Apply ops

acpana:

### Description

I know `Intentions.*` is on its way out but today someone can still hit these endpoints with a SourcePeer like `Intention.Apply` hoping to create a peer service intention. This diff blocks that (and any other `.Apply` op) which is good api hygiene at least.

### DIFF
--- a/agent/consul/intention_endpoint.go
+++ b/agent/consul/intention_endpoint.go
@@ -77,7 +77,7 @@ func (s *Intention) Apply(args *structs.IntentionRequest, reply *string) error {
 		return ErrConnectNotEnabled
 	}
 
-	if args.Intention.SourcePeer != "" {
+	if args.Intention != nil && args.Intention.SourcePeer != "" {
 		return fmt.Errorf("SourcePeer field is not supported on this endpoint. Use config entries instead")
 	}
 
--- a/agent/consul/intention_endpoint.go
+++ b/agent/consul/intention_endpoint.go
@@ -77,6 +77,10 @@ func (s *Intention) Apply(args *structs.IntentionRequest, reply *string) error {
 		return ErrConnectNotEnabled
 	}
 
+	if args.Intention.SourcePeer != "" {
+		return fmt.Errorf("SourcePeer field is not supported on this endpoint. Use config entries instead")
+	}
+
 	// Ensure that all service-intentions config entry writes go to the primary
 	// datacenter. These will then be replicated to all the other datacenters.
 	args.Datacenter = s.srv.config.PrimaryDatacenter
--- a/agent/consul/intention_endpoint_test.go
+++ b/agent/consul/intention_endpoint_test.go
@@ -281,8 +281,7 @@ func TestIntentionApply_NoSourcePeer(t *testing.T) {
 
 	t.Parallel()
 
-	dir1, s1 := testServer(t)
-	defer os.RemoveAll(dir1)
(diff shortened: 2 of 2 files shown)


agent/consul/intention_endpoint_test.go
```diff
@@ -273,6 +273,44 @@ func TestIntentionApply_updateGood(t *testing.T) {
...
+func TestIntentionApply_NoSourcePeer(t *testing.T) {
+	if testing.Short() {
+		t.Skip("too slow for testing.Short")
+	}
+
+	t.Parallel()
+
+	dir1, s1 := testServer(t)
+	defer os.RemoveAll(dir1)
+	defer s1.Shutdown()
```
kisunji, line 286: these defers shouldn't be necessary (unsure about the codec)

freddygv: codec `Close()` can also be removed since it's in a `t.Cleanup()` of `rpcClient`

agent/consul/intention_endpoint.go
```diff
@@ -77,6 +77,10 @@ func (s *Intention) Apply(args *structs.IntentionRequest, reply *string) error {
 		return ErrConnectNotEnabled
 	}
 
+	if args.Intention.SourcePeer != "" {
+		return fmt.Errorf("SourcePeer field is not supported on this endpoint. Use config entries instead")
+	}
```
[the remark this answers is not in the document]
acpana: ah yeah, great call out!

@rboyer , why should it live after the `ForwardRPC` call? Wouldn't it be best to fail fast rather than waste another hop?

```
if args.Intention != nil && ...has SourcePeer... {
   // err
}
```

agent/consul/intention_endpoint_test.go
acpana: 🤔 the set up was gentle copy pasta 🍝 from other tests; let me have a look and happy to break the mold here.

acpana: So `os.RemoveAll(dir)` is not actually called in the cleanup; perhaps we can do a big refactor later per this but that can come in a separate PR

agent/consul/intention_endpoint.go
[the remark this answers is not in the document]
acpana: Let me know what y'all think here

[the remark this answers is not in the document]
kyhavlov: In almost all the other RPC endpoints we forward the request before doing any kind of logic, including validation, so it would mostly be for consistency there. I don't think it's wrong to put it first though - it is quicker/less load on the leader to fail before forwarding the request here.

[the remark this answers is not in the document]
kisunji: I vaguely remember asking this same question around last year 😆 One side effect of validating before the forward is that the leader will validate again, which is probably negligible unless the validation is costly

agent/consul/intention_endpoint_test.go
acpana: > So os.RemoveAll(dir) is not actually called in the cleanup; perhaps we can do a big refactor later per this but that can come in a separate PR

🤦🏼 ah, I didn't actually look at `testutil.TempDir` :( -- thanks for calling that out Chris !! 🥇

kisunji (approved):

PR #13467, opened 2022-06-16
Move ACLResolveResult into acl/resolver package

boxofrad:
### Description
This is the OSS portion of enterprise PR 2055.

Having this type live in the `agent/consul` package makes it difficult to put anything that relies on token resolution (e.g. the new gRPC services) in separate packages without introducing import cycles.

For example, if package `foo` imports `agent/consul` for the `ACLResolveResult` type it means that `agent/consul` cannot import `foo` to register its service.

We've previously worked around this by wrapping the `ACLResolver` to "downgrade" its return type to an `acl.Authorizer` - aside from the added complexity, this also loses the resolved identity information.

In the future, we may want to move the whole `ACLResolver` into the `acl/resolver` package. For now, putting the result type there at least, fixes the immediate import cycle issues.


markan (approved):
+1 as long as this is sync'd to any changes on the enterprise side.

PR #13477, opened 2022-06-16
Add type info to options

lkysow:


trujillo-adam (approved):
This is fine -- I think we need a long-term solution for documenting these params.  I would like to use tables, but I understand the arguments against it. We proposed a strategy for documenting API gateway parameters that we can explore offline.

PR #13481, opened 2022-06-17
Add TLS support in Envoy Prometheus endpoint

kyhavlov:
This PR adds support for enabling TLS on the Envoy Prometheus endpoint (enabled when `envoy_prometheus_bind_addr` is set) via new CLI flags for the CA/cert/key. The intention is to provide these in kubernetes via a mounted volume and have Envoy watch the files to dynamically reload the certs when the volume is updated.

### PR Checklist
* [ ] ~~external facing docs updated~~ (separate PR)


dhiaayachi (approved):
LGTM!

david-yu:
@kyhavlov Will this be back ported to 1.12.x?

PR #13499, opened 2022-06-19
Delete definition of metric `consul.acl.blocked.node.deregistration`

maxb:
Although the metric is defined, there is no code which ever sets its
value - the code in question is genuinely asymmetric - there are 3 types
of object for which registration can be tracked, but only 2 for which
deregistration can be tracked.

This is already reflected in the documentation at:
https://www.consul.io/docs/agent/telemetry#metrics-reference
It's just this part of the code which is out of sync.

### DIFF
--- a/agent/local/state.go
+++ b/agent/local/state.go
@@ -46,10 +46,6 @@ var StateCounters = []prometheus.CounterDefinition{
 		Name: []string{"acl", "blocked", "node", "registration"},
 		Help: "Increments whenever a registration fails for a node (blocked by an ACL)",
 	},
-	{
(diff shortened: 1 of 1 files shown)


Amier3:
Good catch @maxb 

Good catch and thanks for the PR!. For some reasons this is failing some tests (?), but we'll look into that and get back to you soon.

maxb:
> For some reasons this is failing some tests (?), but we'll look into that and get back to you soon.

It looks rather like a spurious failure. Could you trigger a retry?

freddygv (approved):
Thank you!

PR #13507, opened 2022-06-20
docs: Use "error" to use standard log level value

david-yu:
### Description

"err" was used for backwards compatibility, but we should use "error" as the standard for now per [log_levels.go](https://github.com/hashicorp/consul/blob/401221de5852432d742f2bdb1c4f17fb0d0e7362/logging/log_levels.go#L10) 

### Links

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/website/content/commands/monitor.mdx
+++ b/website/content/commands/monitor.mdx
@@ -32,6 +32,6 @@ Usage: `consul monitor [options]`
 - `-log-level` - The log level of the messages to show. By default this
   is "info". This log level can be more verbose than what the agent is
   configured to run at. Available log levels are "trace", "debug", "info",
-  "warn", and "err".
+  "warn", and "error".
 - `-log-json` - Toggles whether the messages are streamed in JSON format.
   By default this is false.
--- a/website/content/commands/monitor.mdx
+++ b/website/content/commands/monitor.mdx
@@ -31,7 +31,7 @@ Usage: `consul monitor [options]`
 
 - `-log-level` - The log level of the messages to show. By default this
   is "info". This log level can be more verbose than what the agent is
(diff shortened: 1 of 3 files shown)


kisunji (approved):

david-yu:
Retrying after CI failures due to Github Actions outage.

trujillo-adam (approved):
Ideally, these would all be in monospace font.

david-yu:
@trujillo-adam Do you mean like `trace`, `debug`, `warn`, etc?

trujillo-adam:
> @trujillo-adam Do you mean like `trace`, `debug`, `warn`, etc?

Yeah. The rationale is that we should format text written by a user and text printed to an interface (console or GUI) as code font. According to our style guide, we don't use quotation marks for anything other than direct quotes. It's something we are currently debating. The argument could be made that we are quoting the interface, and therefore the printed text should be in quotation marks. FWIW, I probably should have kept that comment to myself since the debate is still unsettled.

david-yu:
@trujillo-adam I see. I believe I would like to as well but on the Consul Core side it looks like config is heavily tied to HCL formatting as you can see here in this example from https://www.consul.io/docs/agent/config/config-files. Without quotes, this would render a broken HCL config file. I'll leave the formatting as is, since I think we may need to revisit holistically. 

```
datacenter = "east-aws"
data_dir = "/opt/consul"
log_level = "INFO"
node_name = "foobar"
server = true
watches = [
  {
    type = "checks"
    handler = "/usr/bin/health-check-handler.sh"
  }
]

telemetry {
  statsite_address = "127.0.0.1:2180"
}
```

PR #13508, opened 2022-06-20
Pass trust domain to RBAC to validate and fix use of wrong peer trust bundles

kisunji:
### Description
In xds RBAC, we currently do not validate on hosts, assuming that clusters in a federated mesh would share the same primary trust domain. Peering breaks that model since clusters may now have independent trust domains.

This PR adds validation to hosts so that RBAC can correctly distinguish and apply different sets of rules between cluster-local requests and peered-cluster requests.

This PR also fixes the wrong use of upstream peering trust bundles with the correct inbound peering trust bundles. It renames the fields to be more explicit to avoid future confusion.


freddygv (approved):
:shipit:

PR #13558, opened 2022-06-22
Make memdb indexers generic

kisunji:
### Description
We have many indexer functions in Consul which take `interface{}` and type assert before building the index.

Example:
```go
p, ok := raw.(*structs.ACLToken)
if !ok {
	return nil, fmt.Errorf("unexpected type %T for structs.ACLToken index", raw)
}
```

We can use generics to get rid of the initial plumbing and pass around functions with better defined signatures.

This has two benefits:
1. Less verbosity (see lines removed from this PR)
2. Developers can parse the argument types to memdb schemas without having to introspect the function for the type assertion

* This PR adds no new functionality; tests should pass as-is

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/consul/state/acl_schema.go
+++ b/agent/consul/state/acl_schema.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/hashicorp/go-memdb"
 
-	"github.com/hashicorp/consul/acl"
 	"github.com/hashicorp/consul/agent/structs"
 )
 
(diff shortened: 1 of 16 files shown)


agent/consul/state/acl_schema.go
```diff
@@ -36,77 +35,77 @@ func tokensTableSchema() *memdb.TableSchema {
 				// DEPRECATED (ACL-Legacy-Compat) - we should not AllowMissing here once legacy compat is removed
 				AllowMissing: true,
 				Unique:       true,
-				Indexer: indexerSingle{
-					readIndex:  readIndex(indexFromUUIDString),
-					writeIndex: writeIndex(indexAccessorIDFromACLToken),
+				Indexer: indexerSingle[string, *structs.ACLToken]{
```
boxofrad, line 38: It's a shame that the compiler cannot infer the type arguments here from `readIndex` and `writeIndex`'s types. Maybe in a future release? 🤞🏻

agent/consul/state/catalog_schema.go
```diff
@@ -284,33 +246,28 @@ func connectNameFromServiceNode(sn *structs.ServiceNode) (string, bool) {
...
-	return func(raw interface{}) ([]byte, error) {
+func indexWithPeerName[T any](
+	fn func(T) ([]byte, error),
+) func(T) ([]byte, error) {
+	return func(raw T) ([]byte, error) {
+		pi, ok := any(raw).(peerIndexable)
+		if !ok {
+			return nil, fmt.Errorf("wrapped generic type must be peerIndexable: %T", raw)
+		}
+
```
boxofrad, line 264: Could we remove this type-assertion by constraining `T` to the `peerIndexable` interface like so?

```suggestion
func indexWithPeerName[T peerIndexable](
	fn func(T) ([]byte, error),
) func(T) ([]byte, error) {
	return func(pi T) ([]byte, error) {
```

boxofrad (approved):
Probably needs a 👍🏻 from somebody more familiar with the state store code, but I'm hugely in favor of this! Nice job.

I've left one suggestion, but it's not a blocker.

agent/consul/state/catalog_schema.go
kisunji: Yep! Matt made a similar comment on the enterprise side. The latest commit should address this

boxofrad (approved):

PR #13577, opened 2022-06-23
peering, internal: support UIServices, UINodes, UINodeInfo 

acpana:
### Description
We need to add peering support for the following UI endpoints: `UIServices`, `UINodes`, `UINodeInfo` in order to add data about peered services and nodes thru the UI.

Out of the 3, `UINodeInfo` is the simplest because it either takes a `PeerName` or `""`. `""` is denoted to mean local fetch, aka, local look ups, aka services or nodes that are not peered.

> Note: If `UIServices` or `UINodes` RPC endpoints receive a request for a specific peer (i.e. `args.PeerName != ""), we error out since we don't support that case right now bc it's not needed.

* updated `_test` files

### Links
 * https://app.asana.com/0/1202437710094276/1202439043013559
 * https://app.asana.com/0/1202437710094276/1202454391493007

### DIFF
--- a/agent/consul/internal_endpoint.go
+++ b/agent/consul/internal_endpoint.go
@@ -43,14 +43,7 @@ func (m *Internal) NodeInfo(args *structs.NodeSpecificRequest,
 				return err
 			}
 
-			// old UIs may not handle .Dump so we added all peered nodes in .ImportedDump
-			if args.PeerName != "" {
-				reply.ImportedDump = dump
-			} else {
-				reply.Dump = dump
-			}
-
-			reply.Index = index
+			reply.Index, reply.Dump = index, dump
 			return m.srv.filterACL(args.Token, reply)
 		})
 }
@@ -76,51 +69,44 @@ func (m *Internal) NodeDump(args *structs.DCSpecificRequest,
 		&args.QueryOptions,
 		&reply.QueryMeta,
 		func(ws memdb.WatchSet, state *state.Store) error {
-			peerings := make([]string, 0)
-			peerings = append(peerings, "") // this represents a local fetch
-
-			// if a PeerName is passed, then it's we only get the NodeDump for that peer
-			// otherwise, we get all the local nodes and all the nodes for every peering
+			// we don't support calling this endpoint for a specific peer
 			if args.PeerName != "" {
-				peerings = append(peerings, args.PeerName)
-			} else {
-				// get a list of all peerings
-				_, listedPeerings, err := state.PeeringList(ws, args.EnterpriseMeta)
-				if err != nil {
-					m.logger.Error("could not list peers for node dump", err)
-				} else {
-					for _, lp := range listedPeerings {
-						peerings = append(peerings, lp.Name)
-					}
-				}
+				return fmt.Errorf("this endpoint does not support specifying a peer: %q", args.PeerName)
 			}
 
-			// maxIndex will be set to the highest index of the NodeDump below
+			// this maxIndex will be the max of the NodeDump calls and the PeeringList call
 			var maxIndex uint64
-			for _, p := range peerings {
-				if p != "" {
-					index, importedDump, err := state.NodeDump(ws, &args.EnterpriseMeta, p)
-					if err != nil {
-						return err
-					}
-					reply.ImportedDump = append(reply.ImportedDump, importedDump...)
+			// Get data for local nodes
+			index, dump, err := state.NodeDump(ws, &args.EnterpriseMeta, "")
+			if err != nil {
+				return err
+			}
 
-					if index > maxIndex {
-						maxIndex = index
-					}
-				} else {
-					index, dump, err := state.NodeDump(ws, &args.EnterpriseMeta, "")
-					if err != nil {
-						return err
-					}
+			if index > maxIndex {
+				maxIndex = index
+			}
+			reply.Dump = dump
 
-					// TODO(peering): we should not ever be here twice but maybe it would be important
-					// to log if we are as that could be a programming error
-					reply.Dump = dump
+			// get a list of all peerings
+			index, listedPeerings, err := state.PeeringList(ws, args.EnterpriseMeta)
+			if err != nil {
+				return fmt.Errorf("could not list peers for node dump %w", err)
+			}
 
-					if index > maxIndex {
-						maxIndex = index
-					}
+			if index > maxIndex {
+				maxIndex = index
+			}
+
+			// get node dumps for all peerings
+			for _, p := range listedPeerings {
+				index, importedDump, err := state.NodeDump(ws, &args.EnterpriseMeta, p.Name)
+				if err != nil {
+					return err
+				}
+				reply.ImportedDump = append(reply.ImportedDump, importedDump...)
+
+				if index > maxIndex {
+					maxIndex = index
 				}
 			}
 			reply.Index = maxIndex
@@ -167,53 +153,45 @@ func (m *Internal) ServiceDump(args *structs.ServiceDumpRequest, reply *structs.
 		&args.QueryOptions,
 		&reply.QueryMeta,
 		func(ws memdb.WatchSet, state *state.Store) error {
-			// Get, store, and filter nodes
-
-			peerings := make([]string, 0)
-			peerings = append(peerings, "") // this represents a local fetch
-
-			// if a PeerName is passed, then it's we only get the NodeDump for that peer
-			// otherwise, we get all the local nodes and all the nodes for every peering
+			// we don't support calling this endpoint for a specific peer
 			if args.PeerName != "" {
-				peerings = append(peerings, args.PeerName)
-			} else {
-				// get a list of all peerings
-				_, listedPeerings, err := state.PeeringList(ws, args.EnterpriseMeta)
-				if err != nil {
-					m.logger.Error("could not list peers for service dump", err)
-				} else {
-					for _, lp := range listedPeerings {
-						peerings = append(peerings, lp.Name)
-					}
-				}
+				return fmt.Errorf("this endpoint does not support specifying a peer: %q", args.PeerName)
 			}
-			var maxIdx uint64
-			for _, p := range peerings {
-				var outterIdx uint64
-				if p != "" {
-					innerIdx, importedNodes, err := state.ServiceDump(ws, args.ServiceKind, args.UseServiceKind, &args.EnterpriseMeta, p)
-					if err != nil {
-						return err
-					}
 
-					reply.ImportedNodes = append(reply.ImportedNodes, importedNodes...)
-					outterIdx = innerIdx
-				} else {
-					innerIdx, nodes, err := state.ServiceDump(ws, args.ServiceKind, args.UseServiceKind, &args.EnterpriseMeta, "")
-					if err != nil {
-						return err
-					}
+			// this maxIndex will be the max of the ServiceDump calls and the PeeringList call
+			var maxIndex uint64
 
-					// TODO(peering): we should not ever be here twice but maybe it would be important
-					// to log if we are as that could be a programming error
-					reply.Nodes = nodes
-					outterIdx = innerIdx
+			// get a local dump for services
+			index, nodes, err := state.ServiceDump(ws, args.ServiceKind, args.UseServiceKind, &args.EnterpriseMeta, "")
+			if err != nil {
+				return err
+			}
+
+			if index > maxIndex {
+				maxIndex = index
+			}
+			reply.Nodes = nodes
+
+			// get a list of all peerings
+			index, listedPeerings, err := state.PeeringList(ws, args.EnterpriseMeta)
+			if err != nil {
+				m.logger.Error("could not list peers for service dump", err)
+			}
+
+			if index > maxIndex {
+				maxIndex = index
+			}
+
+			for _, p := range listedPeerings {
+				index, importedNodes, err := state.ServiceDump(ws, args.ServiceKind, args.UseServiceKind, &args.EnterpriseMeta, p.Name)
+				if err != nil {
+					return err
 				}
 
-				if outterIdx > maxIdx {
-					maxIdx = outterIdx
+				if index > maxIndex {
+					maxIndex = index
 				}
-				reply.Index = maxIdx
+				reply.ImportedNodes = append(reply.ImportedNodes, importedNodes...)
 			}
 
 			// Get, store, and filter gateway services
@@ -223,10 +201,10 @@ func (m *Internal) ServiceDump(args *structs.ServiceDumpRequest, reply *structs.
 			}
 			reply.Gateways = gatewayServices
 
-			if idx > maxIdx {
-				maxIdx = idx
+			if idx > maxIndex {
+				maxIndex = idx
 			}
-			reply.Index = maxIdx
+			reply.Index = maxIndex
 
 			raw, err := filter.Execute(reply.Nodes)
 			if err != nil {
--- a/agent/consul/internal_endpoint.go
+++ b/agent/consul/internal_endpoint.go
@@ -77,9 +77,9 @@ func (m *Internal) NodeDump(args *structs.DCSpecificRequest,
 			// this maxIndex will be the max of the NodeDump calls and the PeeringList call
 			var maxIndex uint64
 			// Get data for local nodes
-			index, dump, err := state.NodeDump(ws, &args.EnterpriseMeta, "")
+			index, dump, err := state.NodeDump(ws, &args.EnterpriseMeta, structs.DefaultPeerKeyword)
 			if err != nil {
-				return err
+				return fmt.Errorf("could not get a node dump for local nodes: %w", err)
 			}
 
 			if index > maxIndex {
@@ -101,7 +101,7 @@ func (m *Internal) NodeDump(args *structs.DCSpecificRequest,
 			for _, p := range listedPeerings {
 				index, importedDump, err := state.NodeDump(ws, &args.EnterpriseMeta, p.Name)
 				if err != nil {
-					return err
+					return fmt.Errorf("could not get a node dump for peer's %q nodes: %w", p.Name, err)
 				}
 				reply.ImportedDump = append(reply.ImportedDump, importedDump...)
 
@@ -162,9 +162,9 @@ func (m *Internal) ServiceDump(args *structs.ServiceDumpRequest, reply *structs.
 			var maxIndex uint64
 
 			// get a local dump for services
-			index, nodes, err := state.ServiceDump(ws, args.ServiceKind, args.UseServiceKind, &args.EnterpriseMeta, "")
+			index, nodes, err := state.ServiceDump(ws, args.ServiceKind, args.UseServiceKind, &args.EnterpriseMeta, structs.DefaultPeerKeyword)
 			if err != nil {
-				return err
+				return fmt.Errorf("could not get a service dump for local nodes: %w", err)
 			}
 
 			if index > maxIndex {
@@ -185,7 +185,7 @@ func (m *Internal) ServiceDump(args *structs.ServiceDumpRequest, reply *structs.
 			for _, p := range listedPeerings {
 				index, importedNodes, err := state.ServiceDump(ws, args.ServiceKind, args.UseServiceKind, &args.EnterpriseMeta, p.Name)
 				if err != nil {
-					return err
+					return fmt.Errorf("could not get a service dump for peer's %q nodes: %w", p.Name, err)
 				}
 
 				if index > maxIndex {
--- a/agent/consul/internal_endpoint.go
+++ b/agent/consul/internal_endpoint.go
@@ -43,7 +43,14 @@ func (m *Internal) NodeInfo(args *structs.NodeSpecificRequest,
 				return err
 			}
 
-			reply.Index, reply.Dump = index, dump
+			// old UIs may not handle .Dump so we added all peered nodes in .ImportedDump
+			if args.PeerName != "" {
+				reply.ImportedDump = dump
+			} else {
+				reply.Dump = dump
+			}
+
+			reply.Index = index
 			return m.srv.filterACL(args.Token, reply)
 		})
 }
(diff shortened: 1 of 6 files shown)


agent/consul/internal_endpoint.go
```diff
@@ -43,7 +43,14 @@ func (m *Internal) NodeInfo(args *structs.NodeSpecificRequest,
 				return err
 			}
 
-			reply.Index, reply.Dump = index, dump
+			// old UIs may not handle .Dump so we added all peered nodes in .ImportedDump
```
freddygv, line 46: We can continue to assign all results to `reply.Dump` since an old UI would not populate `args.PeerName`

agent/consul/internal_endpoint.go
```diff
@@ -69,18 +76,67 @@ func (m *Internal) NodeDump(args *structs.DCSpecificRequest,
...
 		func(ws memdb.WatchSet, state *state.Store) error {
-			index, dump, err := state.NodeDump(ws, &args.EnterpriseMeta, args.PeerName)
-			if err != nil {
-				return err
+			peerings := make([]string, 0)
+			peerings = append(peerings, "") // this represents a local fetch
+
+			// if a PeerName is passed, then it's we only get the NodeDump for that peer
+			// otherwise, we get all the local nodes and all the nodes for every peering
+			if args.PeerName != "" {
```
freddygv, line 84: We could avoid supporting this use-case of dumping nodes from a peer for now, since we don't have a need for it. Could return an error if a PeerName is specified before resolving the ACL token above.

agent/consul/internal_endpoint.go
```diff
@@ -69,18 +76,67 @@ func (m *Internal) NodeDump(args *structs.DCSpecificRequest,
 		&args.QueryOptions,
 		&reply.QueryMeta,
 		func(ws memdb.WatchSet, state *state.Store) error {
-			index, dump, err := state.NodeDump(ws, &args.EnterpriseMeta, args.PeerName)
-			if err != nil {
-				return err
+			peerings := make([]string, 0)
+			peerings = append(peerings, "") // this represents a local fetch
```
freddygv, line 80:
```suggestion
			peerings := []string{structs.DefaultPeerKeyword} // this represents a local fetch
```

agent/consul/internal_endpoint.go
```diff
@@ -69,18 +76,67 @@ func (m *Internal) NodeDump(args *structs.DCSpecificRequest,
...
+			}
+			reply.Index = maxIndex
 
 			raw, err := filter.Execute(reply.Dump)
 			if err != nil {
 				return err
 			}
 			reply.Dump = raw.(structs.NodeDump)
 
+			importedRaw, err := filter.Execute(reply.ImportedDump)
```
freddygv, line 134: Can we test specifying a filter selector that points at `ImportedDump`?

agent/consul/internal_endpoint.go
```diff
@@ -69,18 +76,67 @@ func (m *Internal) NodeDump(args *structs.DCSpecificRequest,
...
+						peerings = append(peerings, lp.Name)
+					}
+				}
 			}
-			reply.Index, reply.Dump = index, dump
+
+			// maxIndex will be set to the highest index of the NodeDump below
+			var maxIndex uint64
+			for _, p := range peerings {
+				if p != "" {
```
freddygv, line 101: Since we're populating different fields `Dump` vs `ImportedDump` we can keep the state store queries separate

agent/consul/internal_endpoint.go
```diff
@@ -69,18 +76,67 @@ func (m *Internal) NodeDump(args *structs.DCSpecificRequest,
...
+
+			// if a PeerName is passed, then it's we only get the NodeDump for that peer
+			// otherwise, we get all the local nodes and all the nodes for every peering
+			if args.PeerName != "" {
+				peerings = append(peerings, args.PeerName)
+			} else {
+				// get a list of all peerings
+				_, listedPeerings, err := state.PeeringList(ws, args.EnterpriseMeta)
+				if err != nil {
+					m.logger.Error("could not list peers for node dump", err)
```
freddygv, line 90: I think we should return the error here, rather than proceeding with peering data. Seeing that imported nodes are missing from the UI while local ones are present could cause confusion.

agent/consul/internal_endpoint.go
```diff
@@ -69,18 +76,67 @@ func (m *Internal) NodeDump(args *structs.DCSpecificRequest,
...
+					if index > maxIndex {
+						maxIndex = index
+					}
+				} else {
+					index, dump, err := state.NodeDump(ws, &args.EnterpriseMeta, "")
+					if err != nil {
+						return err
+					}
+
+					// TODO(peering): we should not ever be here twice but maybe it would be important
```
freddygv, line 117: If the local query is split off from the peering queries this TODO is no longer relevant

agent/consul/internal_endpoint.go
```diff
@@ -69,18 +76,67 @@ func (m *Internal) NodeDump(args *structs.DCSpecificRequest,
...
+			peerings := make([]string, 0)
+			peerings = append(peerings, "") // this represents a local fetch
+
+			// if a PeerName is passed, then it's we only get the NodeDump for that peer
+			// otherwise, we get all the local nodes and all the nodes for every peering
+			if args.PeerName != "" {
+				peerings = append(peerings, args.PeerName)
+			} else {
+				// get a list of all peerings
+				_, listedPeerings, err := state.PeeringList(ws, args.EnterpriseMeta)
```
freddygv, line 88: This index should be considered for `maxIdx` or the blocking query might not fire when the list of peerings changes and affects the result set

agent/consul/internal_endpoint.go
```diff
@@ -112,11 +168,53 @@ func (m *Internal) ServiceDump(args *structs.ServiceDumpRequest, reply *structs.
 		&reply.QueryMeta,
 		func(ws memdb.WatchSet, state *state.Store) error {
 			// Get, store, and filter nodes
-			maxIdx, nodes, err := state.ServiceDump(ws, args.ServiceKind, args.UseServiceKind, &args.EnterpriseMeta, args.PeerName)
-			if err != nil {
-				return err
+
+			peerings := make([]string, 0)
```
freddygv, line 172: All the comments from above apply here as well

agent/ui_endpoint_test.go
```diff
@@ -318,13 +359,61 @@ func TestUiServices(t *testing.T) {
...
+				"os":  "linux",
+			},
+			PeerName: "peer1",
+		},
+		// register peer service
+		{
+			Datacenter:     "dc1",
+			Node:           "foo",
+			SkipNodeUpdate: true,
+			Service: &structs.NodeService{
```
freddygv, line 383: This service definition can be combined into the first registration request with the node info, that way you only have one `RegisterRequest`.

agent/consul/state/catalog.go
```diff
@@ -290,8 +290,18 @@ func validateRegisterRequestPeerNamesTxn(_ ReadTxn, args *structs.RegisterReques
 
 	{
 		// TODO(peering): validate the node's peering exists (skip check on restore)
-		peerName := args.PeerName
-		peerNames[peerName] = struct{}{}
+		if len(peerNames) == 1 {
```
freddygv, line 293: This should be reverted, the peer name for all objects in the request must match

agent/ui_endpoint.go
```diff
@@ -395,7 +417,7 @@ func summarizeServices(dump structs.ServiceDump, cfg *config.RuntimeConfig, dc s
 		hasProxy = make(map[structs.ServiceName]bool)
```
freddygv, line 417: `hasProxy` and `summary` should be keyed on `structs.PeeredServiceName`, otherwise the peering information would be lost

agent/ui_endpoint.go
```diff
@@ -413,7 +436,7 @@ func summarizeServices(dump structs.ServiceDump, cfg *config.RuntimeConfig, dc s
 	for _, csn := range dump {
 		if cfg != nil && csn.GatewayService != nil {
 			gwsvc := csn.GatewayService
-			sum := getService(gwsvc.Service)
+			sum := getService(gwsvc.Service, "")
```
freddygv, line 439:
```suggestion
			sum := getService(gwsvc.Service, structs.DefaultPeerKeyword)
```

freddygv (changes requested):
Nice work making progress on these changes this week, they're quite close to being done

agent/consul/internal_endpoint.go
acpana, line 46: also, from chatting with Freddy, we are changing this to use `reply.Dump` in either case.

It makes sense bc old UIs, for instance, won't ask for peer nodes

agent/consul/internal_endpoint_test.go
```diff
@@ -1665,6 +1709,93 @@ func TestInternal_GatewayServiceDump_Ingress_ACL(t *testing.T) {
...
+				"lan": "127.0.0.2",
+				"wan": "198.18.0.2",
+			},
+			NodeMeta: map[string]string{
+				"env": "production",
+				"os":  "linux",
+			},
+			PeerName: "peer1",
+		},
+		"peered service": {
```
acpana, line 1781: from offline chats, we prefer the service registration to happen at the same time as the node

agent/consul/internal_endpoint.go
acpana: I coded in a short circuit for when `args.PeerName` is passed in at all per our chat.

agent/consul/internal_endpoint.go
acpana: Completely did away with this when we split the local vs peer fetches

agent/consul/internal_endpoint.go
acpana: erroring out now

agent/ui_endpoint_test.go
acpana: deduped registration requests

agent/consul/state/catalog.go
acpana: I was the problem here hehe

agent/consul/internal_endpoint.go
```diff
@@ -111,31 +153,71 @@ func (m *Internal) ServiceDump(args *structs.ServiceDumpRequest, reply *structs.
...
 
 			raw, err := filter.Execute(reply.Nodes)
 			if err != nil {
 				return err
 			}
 			reply.Nodes = raw.(structs.CheckServiceNodes)
 
+			importedRaw, err := filter.Execute(reply.ImportedNodes)
+			if err != nil {
+				return err
```
kisunji, line 217: Whenever we have multiple calls to the same function with different arguments, we should wrap the errors with context so we can debug which call threw the error.

agent/consul/internal_endpoint_test.go
```diff
@@ -175,6 +213,21 @@ func TestInternal_NodeDump(t *testing.T) {
 	if !foundFoo || !foundBar {
 		t.Fatalf("missing foo or bar")
 	}
+
+	require.Equal(t, 1, len(out2.ImportedDump))
```
kisunji, line 217:
```suggestion
	require.Len(t, out2.ImportedDump, 1)
```

agent/consul/internal_endpoint_test.go
```diff
@@ -175,6 +213,21 @@ func TestInternal_NodeDump(t *testing.T) {
...
+	require.Equal(t, "peer1", out2.ImportedDump[0].PeerName)
+	require.Equal(t, "foo-peer", out2.ImportedDump[0].Node)
+
+	// test filter on nodes
+	var out3 structs.IndexedNodeDump
+	req2 := structs.DCSpecificRequest{
+		Datacenter:   "dc1",
+		QueryOptions: structs.QueryOptions{Filter: "friend in PeerName"},
```
kisunji, line 225: Not too familiar with filters but it is possible to filter for not empty peerNames to only show peered nodes?

agent/consul/state/catalog.go
```diff
@@ -289,9 +289,13 @@ func validateRegisterRequestPeerNamesTxn(_ ReadTxn, args *structs.RegisterReques
...
 
 	{
+		// todo alex -- is this correct?
 		// TODO(peering): validate the node's peering exists (skip check on restore)
-		peerName := args.PeerName
-		peerNames[peerName] = struct{}{}
+		if len(peerNames) == 0 {
+			peerName := args.PeerName
+			peerNames[peerName] = struct{}{}
+		}
```
kisunji, line 297: I think the len check isn't necessary here (spending O(1) to save O(1) if my memory serves me right)

agent/ui_endpoint.go
```diff
@@ -117,7 +119,22 @@ RPC:
...
+			info.Services = make([]*structs.NodeService, 0)
+		}
+		if info.Checks == nil {
+			info.Checks = make([]*structs.HealthCheck, 0)
+		}
+	}
+
+	if out.ImportedDump == nil {
+		out.ImportedDump = make(structs.NodeDump, 0)
+	}
```
kisunji, line 135: nit: Not necessary to allocate since we `append` to a non-nil `out.Dump` anyways (ImportedDump is optional)

agent/ui_endpoint_test.go
```diff
@@ -103,20 +134,30 @@ func TestUiNodes(t *testing.T) {
...
+	require.Equal(t, a.Config.NodeName, nodes[0].Node)
+	require.NotNil(t, nodes[0].Services)
+	require.Equal(t, 1, len(nodes[0].Services))
+	require.NotNil(t, nodes[0].Checks)
+	require.Equal(t, 1, len(nodes[0].Checks))
+	require.Equal(t, "test", nodes[1].Node)
+	require.NotNil(t, nodes[1].Services)
+	require.Equal(t, 0, len(nodes[1].Services))
+	require.NotNil(t, nodes[1].Checks)
+	require.Equal(t, 0, len(nodes[1].Checks))
```
kisunji, line 149: This block is hard to read; can we make the `len` checks `require.Len` and maybe add some comments to break up the chunks of assertions?

kisunji (approved):
Minor comments, but LGTM!

agent/consul/internal_endpoint.go
```diff
@@ -111,31 +153,71 @@ func (m *Internal) ServiceDump(args *structs.ServiceDumpRequest, reply *structs.
...
+			// we don't support calling this endpoint for a specific peer
+			if args.PeerName != "" {
+				return fmt.Errorf("this endpoint does not support specifying a peer: %q", args.PeerName)
+			}
+
+			// this maxIndex will be the max of the ServiceDump calls and the PeeringList call
+			var maxIndex uint64
+
+			// get a local dump for services
+			index, nodes, err := state.ServiceDump(ws, args.ServiceKind, args.UseServiceKind, &args.EnterpriseMeta, "")
```
freddygv, line 165: Same as above regarding `structs.DefaultPeerKeyword`

agent/ui_endpoint.go
freddygv: I believe that for the UI they prefer that nothing be nil. Presumably it's javascript related.

freddygv: Should check with John before changing this

kisunji: This endpoint joins `out.Dump` and `out.ImportedDump` into one though so it shouldn't matter, right?

agent/ui_endpoint.go
```diff
@@ -389,40 +413,53 @@ RPC:
...
 	}
 
 	for _, csn := range dump {
+		var peerName string
+		// all entities will have the same peer name so it is safe to use the node's peer name
+		if csn.Node != nil {
+			peerName = csn.Node.PeerName
+		} else {
+			peerName = structs.DefaultPeerKeyword
+		}
```
freddygv, line 445: This will never be nil, and if it was there's a bunch of other places below that would panic
```suggestion
		peerName = csn.Node.PeerName
```

freddygv: When building a CheckServiceNode we start from a service instance (`structs.ServiceNode`), and then fetch its node (`structs.Node`). An invariant in the catalog is that services can't exist without a node, so the node for a service that exists won't be nil.

agent/consul/state/catalog.go
freddygv: This change should be reverted because it loosens the validation.

Say that `args.PeerName == "foo"` and `args.Service.PeerName == "bar"`. 

This input would not return an error because `peerNames` would only contain `bar`, despite there being conflicting peer names for a service on a node.

agent/consul/internal_endpoint_test.go
```diff
@@ -29,56 +31,80 @@ func TestInternal_NodeInfo(t *testing.T) {
...
+	t.Run("get peered node", func(t *testing.T) {
+		var out structs.IndexedNodeDump
+		req := structs.NodeSpecificRequest{
+			Datacenter: "dc1",
+			Node:       "foo",
+			PeerName:   "peer1",
+		}
+		require.NoError(t, msgpackrpc.CallWithCodec(codec, "Internal.NodeInfo", &req, &out))
+
+		// here we use ImportedNodes as well
```
freddygv, line 102: What's meant by this comment?

agent/consul/internal_endpoint.go
acpana: By the way, we added tests for these!

agent/consul/internal_endpoint_test.go
freddygv: @kisunji I'm not sure that would be useful here since there's already tests for filtering the `.Dump` field. 

The question we had was about whether the new `ImportedDump` could be filtered on.

agent/consul/internal_endpoint_test.go
```diff
@@ -175,6 +213,21 @@ func TestInternal_NodeDump(t *testing.T) {
 	if !foundFoo || !foundBar {
 		t.Fatalf("missing foo or bar")
 	}
+
+	require.Equal(t, 1, len(out2.ImportedDump))
+	require.Equal(t, "peer1", out2.ImportedDump[0].PeerName)
+	require.Equal(t, "foo-peer", out2.ImportedDump[0].Node)
+
+	// test filter on nodes
```
freddygv, line 221: There's a test for filtering below: `TestInternal_NodeDump_Filter`, this should be moved over thtere

agent/ui_endpoint.go
freddygv: Ah yep good catch

agent/consul/internal_endpoint_test.go
acpana: we can -- we can basically construct any boolean expression over the structs

https://www.consul.io/api-docs/features/filtering#creating-expressions

I can add a test for that right now

agent/consul/state/catalog.go
acpana: This should have been reverted ! 🤦🏼

agent/ui_endpoint.go
acpana: I can appreciate that; I was thinking more of the lines of defense in depth and such but I can take this out

acpana:
working 🔧 on the test

agent/ui_endpoint.go
acpana: actually it seems that UIGateway tests call this summarize func with nil  `cfn.Node`; not sure if feature or bug

acpana: https://app.circleci.com/pipelines/github/hashicorp/consul/34387/workflows/03191cf7-b29c-45d0-ad6d-a5d6845e46e6/jobs/747784

acpana: For instance, looking at some test setups --> 

```
&{Node:<nil> Service:<nil> Checks:[] GatewayService:0xc000e24620} 
```

freddygv: Ah ok. Yep I'm seeing the `continue` in L457 now.

agent/consul/internal_endpoint.go
```diff
@@ -101,7 +101,7 @@ func (m *Internal) NodeDump(args *structs.DCSpecificRequest,
 			for _, p := range listedPeerings {
 				index, importedDump, err := state.NodeDump(ws, &args.EnterpriseMeta, p.Name)
 				if err != nil {
-					return err
+					return fmt.Errorf("could not get a node dump for peer's %q nodes: %w", p.Name, err)
```
freddygv, line 104: Minor grammar point, the `'s` should be after the peer name:
`"could not get a node dump for peer %s's nodes: %w"`

But that aside we can shorten this:
`"could not get a node dump for peer %q: %w"`

agent/consul/internal_endpoint.go
```diff
@@ -185,7 +185,7 @@ func (m *Internal) ServiceDump(args *structs.ServiceDumpRequest, reply *structs.
 			for _, p := range listedPeerings {
 				index, importedNodes, err := state.ServiceDump(ws, args.ServiceKind, args.UseServiceKind, &args.EnterpriseMeta, p.Name)
 				if err != nil {
-					return err
+					return fmt.Errorf("could not get a service dump for peer's %q nodes: %w", p.Name, err)
```
freddygv, line 188: Same as above, could be shortened:
`"could not get a service dump for peer %q: %w"`

freddygv (approved):
LGTM, thanks for going through these edits!

agent/consul/internal_endpoint.go
```diff
@@ -69,18 +69,60 @@ func (m *Internal) NodeDump(args *structs.DCSpecificRequest,
...
+			if index > maxIndex {
+				maxIndex = index
 			}
-			reply.Index, reply.Dump = index, dump
+
+			// get node dumps for all peerings
+			for _, p := range listedPeerings {
+				index, importedDump, err := state.NodeDump(ws, &args.EnterpriseMeta, p.Name)
+				if err != nil {
+					return fmt.Errorf("could not get a node dump for peer's %q nodes: %w", p.Name, err)
```
acpana, line 104:
```suggestion
					return fmt.Errorf("could not get a node dump for peer %q: %w", p.Name, err)
```

agent/consul/internal_endpoint.go
```diff
@@ -111,31 +153,71 @@ func (m *Internal) ServiceDump(args *structs.ServiceDumpRequest, reply *structs.
...
+			}
+
+			if index > maxIndex {
+				maxIndex = index
+			}
+
+			for _, p := range listedPeerings {
+				index, importedNodes, err := state.ServiceDump(ws, args.ServiceKind, args.UseServiceKind, &args.EnterpriseMeta, p.Name)
+				if err != nil {
+					return fmt.Errorf("could not get a service dump for peer's %q nodes: %w", p.Name, err)
```
acpana, line 188:
```suggestion
					return fmt.Errorf("could not get a service dump for peer %q: %w", p.Name, err)
```

PR #13582, opened 2022-06-23
Add new index for PeeredServiceName and ServiceVirtualIP

kisunji:
For TProxy we will be leveraging the VirtualIP table, which need to become peer-aware

* Updated tests to reflect new shape of indexes
* Added more test cases to show proper indexing of peered services

### DIFF
--- a/agent/consul/state/catalog.go
+++ b/agent/consul/state/catalog.go
@@ -7,7 +7,7 @@ import (
 	"reflect"
 	"strings"
 
-	memdb "github.com/hashicorp/go-memdb"
+	"github.com/hashicorp/go-memdb"
 	"github.com/mitchellh/copystructure"
 
 	"github.com/hashicorp/consul/acl"
@@ -872,9 +872,10 @@ func ensureServiceTxn(tx WriteTxn, idx uint64, node string, preserveIndexes bool
 				return fmt.Errorf("failed updating gateway mapping: %s", err)
 			}
 		}
-	}
-	if err := upsertKindServiceName(tx, idx, svc.Kind, svc.CompoundServiceName()); err != nil {
-		return fmt.Errorf("failed to persist service name: %v", err)
+		// Only upsert KindServiceName if service is local
+		if err := upsertKindServiceName(tx, idx, svc.Kind, svc.CompoundServiceName()); err != nil {
+			return fmt.Errorf("failed to persist service name: %v", err)
+		}
 	}
 
 	// Update upstream/downstream mappings if it's a connect service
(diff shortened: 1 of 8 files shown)


agent/consul/state/catalog.go
```diff
@@ -3336,7 +3344,9 @@ func getTermGatewayVirtualIPs(tx WriteTxn, services []structs.LinkedService, ent
 	addrs := make(map[string]structs.ServiceAddress, len(services))
 	for _, s := range services {
 		sn := structs.ServiceName{Name: s.Name, EnterpriseMeta: *entMeta}
-		vip, err := assignServiceVirtualIP(tx, sn)
+		// TODO(peering): should this take peername?
```
freddygv, line 3347: No it shouldn't take a peer name. Can you pass in structs.DefaultPeerKeyword here?

Terminating gateway virtual IPs are for transparent proxies to dial services represented by a terminating gateway. But terminating gateways can't route to services in peered clusters.

agent/consul/state/catalog.go
```diff
@@ -1981,7 +1982,13 @@ func (s *Store) deleteServiceTxn(tx WriteTxn, idx uint64, nodeName, serviceID st
 
 // freeServiceVirtualIP is used to free a virtual IP for a service after the last instance
 // is removed.
-func freeServiceVirtualIP(tx WriteTxn, svc string, excludeGateway *structs.ServiceName, entMeta *acl.EnterpriseMeta) error {
+func freeServiceVirtualIP(
+	tx WriteTxn,
+	svc string,
+	excludeGateway *structs.ServiceName,
+	entMeta *acl.EnterpriseMeta,
+	peerName string,
```
freddygv, line 1990: Could we pass in structs.PeeredServiceName rather than svc, entMeta, and peerName here?

agent/consul/state/catalog_test.go
```diff
@@ -1799,7 +1799,7 @@ func TestStateStore_EnsureService_VirtualIPAssign(t *testing.T) {
 	require.NoError(t, s.EnsureService(10, "node1", ns1))
 
 	// Make sure there's a virtual IP for the foo service.
-	vip, err := s.VirtualIPForService(structs.ServiceName{Name: "foo"})
+	vip, err := s.VirtualIPForService(structs.PeeredServiceName{ServiceName: structs.ServiceName{Name: "foo"}})
```
freddygv, line 1802: Can you add a basic test to ensure the peering separation is enforced? Similar to what we often test for partitions/namespaces.

Could register "foo" as a local service and an imported service, then verify that yields two distinct virtual IPs.

freddygv (approved):
Copied the same suggestions from ENT

PR #13596, opened 2022-06-24
egress gateway connect proxy

dhiaayachi:
Description
When a connect proxy associated service is the source of an intention that have a destination as target, configure the connect proxy to route the requests to that destination to the associated terminating gateway.

Testing
Test Watches, test xDS generation, test added RPC endpoints


DanStough:
Closing and re-opening as #13613 with bugfixes/edits.

PR #13597, opened 2022-06-24
[1.12] Remove trigger-oss-merge CI job

eculver:
### Description
We recently moved our OSS->Ent automation to a GitHub action. It is already set up to handle the release branches as well, so we need to remove these triggers to stop the CircleCI robots from doing it.

The effect is that jobs will no longer be triggered for merges to the release branch, so I'll be testing the absence of an event once this merges.


dhiaayachi (approved):
🚀

PR #13604, opened 2022-06-27
ui: Gradual deprecation of old StateChart interface

johncowen:
### Description

TLDR

This is a a new `StateMachine` component inheriting from `StateChart` and has the exact same functionality for interface changing purposes, changes are:

1. exported `|State Guard Action dispatch state|` properties have changed to a object containing those properties instead instead > `|fsm|` (i.e. `fsm.State`, `fsm.dispatch` etc)
2. `Guard` and `Action` were removed, once we need them this will be managed via entry-like `@guards` and `@actions` args.

--- 

Our `<StateChart />` and related components could do with an upgrade, mainly due to the less than ergonomic interface which is proving difficult to work with. This is reasonably often, making it feel like a unicorn dies every time we use it 😅 . We are looking to use this quite a bit in our upcoming work.

We also use this component already in various places so we want to support the old interface whilst beginning to introduce the more ergonomic new interface. Therefore we chose to make a new `StateMachine` component inheriting from the old component and add the new interface to that (the StateMachine name is better anyway so I'm kinda lucky there), so its the same functionality just a different interface. Please note: We don't need Guard and Action for the current work they are doing, but the new interface will just provide these as attributes instead of using components for this, hence these are currently not supported by this new iteration. Once we need that functionality (if we need it in the near future) we can add that then.

I also noticed to upgrade `<State />` to Octane was relatively straightforwards and a quick win whilst we were here.

Lastly, I replaced our require helper with something more dynamic now I'm pretty sure this is fine moving forwards with embroider. We will be using this in future instead of our state-chart helper to provide the chart to the machine. Making it dynamic just means less hassle to do this.

P.S. Tests here are currently failing for unrelated reasons, that is being worked on separately and we can rebase here once thats ready (see https://github.com/hashicorp/consul/pull/13633).

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-ui/app/components/state-chart/README.mdx
+++ b/ui/packages/consul-ui/app/components/state-chart/README.mdx
@@ -3,6 +3,8 @@ class: ember
 ---
 # StateChart
 
+**This component is deprecated and is currently undergoing an interface change in <StateMachine />**
+
 ```hbs
 <StateChart
   @chart={{xstateStateChartObject}}
--- a/ui/packages/consul-ui/app/components/state-machine/README.mdx
+++ b/ui/packages/consul-ui/app/components/state-machine/README.mdx
@@ -1,0 +1,61 @@
+---
+class: ember
+---
+# StateMachine
+
+```hbs
+<StateMachine
+  @chart={{xstateStateChartObject}}
+  as |fsm|>
+</StateMachine>
+```
+
+`<StateMachine />` is a renderless component that eases rendering of different states
+from within templates using XState State Machine and Statechart objects.
+
+**Please note**: This component is currently a gradual replacement for
+StateChart, which has a less ergonmic interface. Guard and Action component are
+currently omitted due to a preference to use entries to define those instead of
+components (which is WIP)
+
+### Arguments
+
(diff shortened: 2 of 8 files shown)


LevelbossMike:
Not a huge fan of this approach of doing this in the template layer but this looks like an interface improvement to what the current `StateChart`-component is doing. :+1:

[ember-statecharts](https://ember-statecharts.com) provides the same functionality but let's you use the statecharts as [resources](https://github.com/wycats/polaris-sketchwork/blob/main/src/reactivity.md#resources) from javascript. This is probably an approach that goes more into the direction that the framework is taking and gives us better ergonomics because it is easier to react to args changes etc.

evrowe:
@johncowen Would you mind adding a section to the PR description that explicitly and concisely defines what the interface changes are? I realize the code itself is a reference to this but it would be good to add the high level context here.

evrowe:
@johncowen Nice thank you!

johncowen:
Thanks @evrowe , we could definitely add some tests around here if we wanted to in the future.

PR #13607, opened 2022-06-27
Fix Payload Passthrough in the Serverless Plugin

erichaberkorn:
### Description
Renamed the payload passthrough key from "passhthrough" to "passthrough".

### DIFF
--- a/.changelog/13607.txt
+++ b/.changelog/13607.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+xds: Fix a bug that resulted in Lambda services not using the payload-passthrough option as expected.
+```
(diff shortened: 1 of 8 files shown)


.changelog/13607.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
```
lkysow, line 1: this should be release-note:bug right?

erichaberkorn: Nice  catch

lkysow (commented):
Should our tests run through both passthrough false and passthrough true?

lkysow (approved):
With comments resolved from ent PR

PR #13610, opened 2022-06-27
test: run Envoy integration tests against both servers and clients

boxofrad:
### Description
This is the OSS portion of enterprise PR 2170.

With agentless, we source proxy configuration data differently on servers and clients, so it's necessary to run the integration tests against both to catch regressions.

This PR introduces a new environment variable `XDS_TARGET` that controls whether Envoy will be configured to a Consul server or client agent for its xDS sessions.

Setting `XDS_TARGET=server` maintains the existing behavior, where we run a single Consul process in `-dev` (and therefore server) mode. Whereas setting `XDS_TARGET=client` also runs a client agent and disables the server's gRPC port.

There's a fairly lengthy comment in `start_consul` that expands on some of the details of how all this works (e.g. hostnames, configuration).

* Run the integration tests (or watch the run in CI) 😎

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,6 +23,10 @@ references:
     BASH_ENV: .circleci/bash_env.sh
     VAULT_BINARY_VERSION: 1.9.4
     GO_VERSION: 1.18.1
+  envoy-versions: &supported_envoy_versions
+    - &default_envoy_version "1.19.5"
+    - "1.20.4"
+    - "1.21.3"
   images:
     # When updating the Go version, remember to also update the versions in the
     # workflows section for go-test-lib jobs.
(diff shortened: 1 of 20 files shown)


.circleci/config.yml
```diff
@@ -891,21 +905,6 @@ jobs:
...
-
-  envoy-integration-test-1_21_3:
-    <<: *ENVOY_TESTS
-    environment:
-      ENVOY_VERSION: "1.21.3"
-
-  envoy-integration-test-1_22_2:
-    <<: *ENVOY_TESTS
-    environment:
-      ENVOY_VERSION: "1.22.2"
```
eculver, line 907: Looks like we accidentally removed 1.22 from the support matrix here :/

eculver: I'm re-applying in https://github.com/hashicorp/consul/pull/13807

PR #13619, opened 2022-06-28
Acpana/peering hint state

acpana:

### Description
We need to show that a peering is active for UI usage. This is a hack to look at the stream tracker and see if it's connected. If so, we call that a `PeeringState.Active`. 

In the future, we want to monitor the health of a peering and do the reconciliation there.

*completed the api test: it now encompasses a full peering 💯

### DIFF
--- a/agent/rpc/peering/service.go
+++ b/agent/rpc/peering/service.go
@@ -370,7 +370,25 @@ func (s *Service) PeeringList(ctx context.Context, req *pbpeering.PeeringListReq
 	if err != nil {
 		return nil, err
 	}
+
+	for _, p := range peerings {
+		p.State = s.reconciledStreamStateHint(p.ID, p.State)
+	}
 	return &pbpeering.PeeringListResponse{Peerings: peerings}, nil
+}
+
+// TODO(peering): Maybe get rid of this when actually monitoring the stream health
+// reconciledStreamStateHint peaks into the streamTracker and determines whether a peering should be marked
+// as PeeringState.Active or not
+func (s *Service) reconciledStreamStateHint(pID string, pState pbpeering.PeeringState) pbpeering.PeeringState {
+	streamState, found := s.streams.streamStatus(pID)
+
+	if found && streamState.Connected {
+		return pbpeering.PeeringState_ACTIVE
+	}
+
+	// default, no reconciliation
+	return pState
 }
 
 // TODO(peering): As of writing, this method is only used in tests to set up Peerings in the state store.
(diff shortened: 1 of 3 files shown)


agent/rpc/peering/service.go
```diff
@@ -370,9 +370,27 @@ func (s *Service) PeeringList(ctx context.Context, req *pbpeering.PeeringListReq
...
 }
 
+// TODO(peering): Maybe get rid of this when actually monitoring the stream health
+// reconciledStreamStateHint peaks into the streamTracker and determines whether a peering should be marked
+// as PeeringState.Active or not
+func (s *Service) reconciledStreamStateHint(pID string, pState pbpeering.PeeringState) pbpeering.PeeringState {
+	streamState, found := s.streams.streamStatus(pID)
+
+	if found && streamState.Connected {
+		return pbpeering.PeeringState_ACTIVE
```
acpana, line 387: Maybe this is not enough to just say that the streamState is connected.

Maybe we need to say something like "if no errors in the last 10 seconds received or sent" then it is active and not failing." In that case, the corollary would give the failing state hint : "if one or more errors were received or sent, then it is active but failing. " and we'd mark the state as PeeringState_Failing.

mkeeler: For the time being I think this is fine. Once we have the stream heartbeating then we can revisit the connected state implying Active with some better backing data.

agent/rpc/peering/service_test.go
```diff
@@ -912,7 +912,7 @@ func newTestServer(t *testing.T, cb func(conf *consul.Config)) testingServer {
 	testrpc.WaitForLeader(t, server.RPC, conf.Datacenter)
 
 	backend := consul.NewPeeringBackend(server, deps.GRPCConnPool)
-	handler := &peering.Service{Backend: backend}
```
mkeeler, line 915: I think we also need a test in this file to ensure the active peering state shows up. I know you implemented one in the api module but I would also love to see one here in the same package where we actually compute the state.

mkeeler (changes requested):

mkeeler (approved):

api/peering_test.go
```diff
@@ -140,11 +140,21 @@ func TestAPI_Peering_GenerateToken(t *testing.T) {
...
-	ctx, cancel := context.WithTimeout(context.Background(), DefaultCtxDuration)
+	// make a "client" server in second DC for peering
+	c2, s2 := makeClientWithConfig(t, nil, func(conf *testutil.TestServerConfig) {
+		conf.Datacenter = "dc2"
+	})
+	defer s2.Stop()
+
+	testutil.RunStep(t, "register services to get synced dc2", func(t *testing.T) {
+		testNodeServiceCheckRegistrations(t, c2, "dc2")
+	})
```
kisunji, line 155: This shouldn't need a RunStep (it's a helper func)

acpana:
ah I raced with Chris's feedback 🤦🏼 💻 ; Thanks for offering feedback Chris 💯 🥇 

I will follow up on it in my next PR

PR #13620, opened 2022-06-28
UI/fix/peering route param

LevelbossMike:
### Description
This PR moves from using a route-param for indicating the peer of a peered service/node to using the service- / node-name portion of the `dc.services.show` / `dc.nodes.show`-route instead. This mirrors the mechanism we use for identifying a peer of an `Intention` and fixes the failures we are seeing in the `ent`-test-suite atm.

#### Why?
The Ember router expects the url nesting to reflect:

1. the outlet nesting of UI
2. nested data-loading

This means when we are dealing with the following routing structure we break the way the Ember router expects the url to be structured when we prepend a peering - name to the url like we do currently (`/:peer-name/dc-name/services/service-name`):

```javascript
this.route('dc', { path: '/:dc-name'}, function() {
  this.route('services', function() {
     this.route('show', { path: '/:name'} 
  });
} 

/* 
^ would usually create a URL that looked like this: /<dc-name>/services/<service-name>
we are creating the following URL for a peered service though: /:<peer-name>/<dc-name>/services/<service-name>
*/
```

The current implementation of indicating a `peering` in the URL breaks the assumptions that the Ember router makes around the URL-structure and this mismatch leaks across the entire application and currently breaks the `ent`-test-suite.

I.e. to "reset" a routing param when we transition from a peered service in the `dc.services.show` route into the `dc.services`  route (
`/:<peer-name>/<dc-name>/services/<service-name>` -> `/<dc-name>/services/<service-name>`) we explicitly need to reset the `peer`-route param via using the `href-to`-helper by explicitly passing a route-param of `peer=undefined`. Explicitly unsetting the peer breaks other assumptions that the application is making regarding `partitions` and `namespaces` that both also use a prepended route-param to scope the application by `partition` and `namespace`.

To fix this  issue we move the `peer`-indicator into the portion of the url where it belongs from an ember router nesting perspective and we get rid of the need to explicitly reset the peer param across the entire application. This also better reflects the fact that a peered service/node is uniquely identified not by its name but by the __combination__ of the `PeerName` and `Name`.

#### ~~Usage of `Peering::Provider` looks weird - why?~~ Updates in `routlet`-service - why?

The way the application is currently structured is working around the usual way of loading data in `Route`'s `model`-hooks and uses a declarative way of loading data via the `DataSource`/ `DataLoader`-components instead. This means we can't parse the `params` we receive via the URL in a Route's `model`-hook like we would do regularly but we need to do this in the template layer because that's the only place where we have access to the `route.params` we are interested in. An alternative would be to pass the `route`-params we receive in the `model`-hook into the controller layer and make them available to the template that way but this

a) would not use the same mechanism the rest of the application is using to access routing-params
b) need to be done in multiple places

Because of the reasons outlined I decided ~~to create a provider-component~~ to add a method to the `routlet`-service that encapsulates the logic of parsing a "peered"-`service`/`node`-name-combination to receive the two parts of the URL we are interested in `PeerName` and `Name`.

#### What's up with the `SlugWithPeer` computed in `Service`, `Node` and `ServiceInstance` ?
`SlugWithPeer` is a new tracked getter that I added to the `Service`, `ServiceInstance` and `Node`-models because we need to create links to the respective detail routes of these resources with their `peer` included in the URL.

Instead of creating links like `/<dc-name>/services/<service-name>` we will create `/<dc-name>/services/peer:<peer-name>:<service-name>` instead. This encodes the peer-name in the URL and allows us to send the correct query-data to the API to receive back a peered service, node or service-instance.

#### Why aren't we using a query-param instead?

We could achieve the same we are doing here with adding a query-parameter to the `dc.services.show` and`dc.nodes.show` routes that indicate the peering. We decided against this approach because a `PeerName` can be considered part of the unique identifier of a service as only the combination of `PeerName` and `Name` of a service or node will allow the API to query a peered `Service` or `Node`.

* Running the test suite now doesn't show test failures anymore - this only applied to the `ent`-test-suite for the UI

### PR Checklist

* [x] updated test coverage
* ~~[ ] external facing docs updated~~
* [x] not a security concern

cc @johncowen

### DIFF
--- a/ui/packages/consul-peerings/app/components/consul/peer/selector/index.hbs
+++ b/ui/packages/consul-peerings/app/components/consul/peer/selector/index.hbs
@@ -10,9 +10,7 @@
   class={{if (is-href 'dc.peers' @dc.Name) 'is-active'}}
 >
     <a
-      href={{href-to 'dc.peers' @dc.Name
-        params=(hash peer=undefined)
-      }}
+      href={{href-to 'dc.peers' @dc.Name}}
     >
       Peers
     </a>
--- a/ui/packages/consul-ui/app/components/consul/node/list/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/node/list/index.hbs
@@ -21,7 +21,7 @@ as |item index|>
         </Tooltip>
       </dd>
     </dl>
-    <a data-test-node href={{href-to "dc.nodes.show" item.Node params=(hash peer=item.PeerName)}}>
+    <a data-test-node href={{href-to "dc.nodes.show" item.SlugWithPeer}}>
       {{item.Node}}
     </a>
   </BlockSlot>
--- a/ui/packages/consul-ui/app/components/consul/service-instance/list/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/service-instance/list/index.hbs
@@ -10,11 +10,11 @@ as |proxies|}}
 as |item index|>
   <BlockSlot @name="header">
     {{#if (eq @routeName "dc.services.show")}}
-      <a data-test-service-name href={{href-to @routeName item.Service.Service}}>
+      <a data-test-service-name href={{href-to @routeName item.ServiceSlugWithPeer}}>
         {{item.Service.ID}}
       </a>
     {{else}}
-      <a data-test-service-name href={{href-to @routeName item.Service.Service item.Node.Node (or item.Service.ID item.Service.Service)}}>
+      <a data-test-service-name href={{href-to @routeName item.ServiceSlugWithPeer item.Node.Node (or item.Service.ID item.Service.Service)}}>
         {{item.Service.ID}}
       </a>
     {{/if}}
--- a/ui/packages/consul-ui/app/components/consul/service/list/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/service/list/index.hbs
@@ -27,15 +27,13 @@
 {{#if (gt item.InstanceCount 0)}}
   <a
     data-test-service-name
-    href={{href-to "dc.services.show.index" item.Name
+    href={{href-to "dc.services.show.index" item.SlugWithPeer
       params=(if (not-eq item.Partition @partition)
         (hash
           partition=item.Partition
           nspace=item.Namespace
-          peer=item.PeerName
         )
         (hash
-          peer=item.PeerName
         )
       )
     }}
--- a/ui/packages/consul-ui/app/components/hashicorp-consul/index.hbs
+++ b/ui/packages/consul-ui/app/components/hashicorp-consul/index.hbs
@@ -81,7 +81,7 @@
 
   <:home-nav>
     <a
-      href={{href-to 'index' params=(hash peer=undefined)}}
+      href={{href-to 'index'}}
     ><Consul::Logo /></a>
   </:home-nav>
 
@@ -115,7 +115,7 @@
           }}
         >
           <Action
-            @href={{href-to 'dc.show' @dc.Name params=(hash peer=undefined)}}
+            @href={{href-to 'dc.show' @dc.Name}}
           >
             Overview
           </Action>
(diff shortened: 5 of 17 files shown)


LevelbossMike:
Yes pulling this into the `Route`-component came to mind yesterday. With this approach we could consider the `Route`-component to be a provider for the `RouterService` that makes route params we are interested in available to the template layer.

The nice thing about this approach is that we could use this to encapsulate the parsing logic for the `:name`-param for services et al. inside the `Route`-Provider-Component. Nothing prevents using the params that we can access via the `RouterService` and transforming them into something that makes sense for us in the context of a peered service, node, and service-instance routes.

For a peered service this probably means to have `route.params.peer` available as well as `route.params.name` - if we are handling a non-peered resource we'd only have a populated value for `route.params.name`. I can update the PR accordingly which will get rid of the granted pretty weird looking `PeeringProvider`-component.

johncowen:
I think your suggestions in your last comment sound great! I really like the idea of encapsulating parsing logic the behind the `Route` component also.

LevelbossMike:
@johncowen I updated the PR and pulled in the peer-identifier parsing into the routlet service so we can access the peer name via the `Route`-component.

johncowen:
Wanted to leave a note here on where we left this conversation from offline

The main concern here is that service names can contain any characters, especially in a service discovery usecase (specifically here the `:` is valid in a service name). So we want to avoid any clashes between valid services names and service names that are peered.

By default the `:` does not get encoded to `%3A` when being added to the URL, otherwise we would have avoided clashes by using having URLs like `peer-name:dev%3Aservice-name` which would have been impossible to replicate as an end user, trying to produce that as an end user would result in `peer-name%3Adev%3Aservice-name`.

Knowing the lengths we have gone to in the past to avoid these sorts of clashes and after consulting other members of the core engineering team, the opinion is, ideally if we can avoid these clashes then we should try to.

As it stands the existing implementation works and avoids this clash and I've not seen or thought of any issues with using that approach as yet, so unless there is a change of opinion from the rest of engineering team I am inclined to go with that for the moment at least.

Lastly just to note, there is plenty of time to revisit this if we decide and I think the last set of amends you've made here make this much better in my opinion.

Overall thanks for the work to date here! I'm glad you got to work a little in the guts of the Route component. 🙌

LevelbossMike:
@johncowen  the way that the peering identifier is structured is the following:

`peer:<peer-name (dns-label)>:<service-name>`

Examples:

`peer:billing:servica-a`⁣ – will parse billing as the peer identifier and service-a as the service name
`billing:service-a⁣` – will consider the entire thing to be the service-name
`service-a` - will work as expected

The only case where this might be an issue is literally when people would prefix the service name with `peer:<something>:`. To me, this feels very much like an edge case.

The current implementation has a _severe_ issue outlined in the PR description. The way the peering-parameter is set up currently makes it easy to break the application if we forget to ever unset the peer-parameter by specifying `params=(hash peer=undefined)` in every link that will transition higher up out of a peered resources route-hierarchy. You can see the concept breaking down in the main navigation already, as this not only applies to new links that we add to the application but also old links that you can access from templates higher up the outlet-hierarchy. This is an issue for every link in the application that will link away from a potentially peered resource, and at least to me feels much more problematic in the long-term as the peer-identifier edge case outlined above.

evrowe:
Agreed, there are trade-offs involved in this decision; both have their advantages and disadvantages.

My thoughts on this are that either way, I see us having to continue writing additional code to work around an increasing number of edge cases; the more things we have to parse out of the route parameters, the more complex this code is going to get. I think we will need to re-examine how we are approaching this problem from top to bottom and work toward something that is more resilient and can scale more safely.

That said, that effort sounds larger in scope than what we would want to address in this PR, and in this case it would probably be best if we continue to use the solution we have in place already, accept that there is some risk involved with that, and commit to prioritizing work to improve the entire implementation of route parsing in the coming quarter.

LevelbossMike:
@evrowe sounds good to me - closing.

PR #13636, opened 2022-06-29
ui: peering chores

LevelbossMike:
### Description
Two small changes to move the peering UI forward:

* Update empty downstream topology message to take peers into account
Screenshot 2022-06-29 at 17 41 13
* Add a filter on the service listing that allows for filtering services that don't have any ExternalSources set (this will include peered services)
Screenshot 2022-06-29 at 17 59 35

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

cc @johncowen


johncowen (approved):
This all looks super good to me! Like that you did the tests here due to the slightly more complex logic than usual, thanks for that.

Approving as of now, but I'm just giving it a quick test run locally, and following up on a question I had offline (that is non-blocking) re: external sources and peered services.

Overall good to merge 👍

PR #13646, opened 2022-06-29
Allow the /v1/internal/acl/authorize endpoint to authorize the “peering" resource

mkeeler:
### Description
The UI uses the endpoint to determine if a user has permissions to do certain operations. This resource will be checked to see if a user has the ability to read/manage peerings.

Currently the permissions checked are just `operator:{read,write}`. Once we implement ACLs for peering this will need to be changed. In the mean time though, have this implemented like so will unblock frontend folks from working on integration.

* Unit tests
* Run a dev agent with ACLs enabled. Bootstrap ACLs and then `curl -X POST -H "X-Consul-Token: $CONSUL_HTTP_TOKEN" localhost:8500/v1/internal/acl/authorize -d '[{"resource": "peering", "access": "read"}, {"resource": "peering", "access": "write"}]'`

### PR Checklist

* [x] updated test coverage
* [ ] not a security concern

### DIFF
--- a/acl/authorizer.go
+++ b/acl/authorizer.go
@@ -49,6 +49,7 @@ const (
 	ResourceQuery     Resource = "query"
 	ResourceService   Resource = "service"
 	ResourceSession   Resource = "session"
+	ResourcePeering   Resource = "peering"
 )
 
 // Authorizer is the interface for policy enforcement.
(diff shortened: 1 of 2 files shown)


acpana (approved):
LGTM! 🚢

PR #13648, opened 2022-06-30
ui: Peer listing with dc/ns/partition/name based unique IDs and polling deletion

johncowen:
### Description

All previously discussed offline:

This PR alters the way we load in the peer listing mainly because:

1. We need to create an ID consisting of datacenter, namespace, partition and name as there are places where we do not know the peer ID and we need to generate a page (and consequently a URL) and a ED record based on a unique ID
2. Previous to this PR deleting a peer would not be reflected in the UI
3. We need to pass the `partition`, probably conditionally depending on whether we are in Enterprise or not to the `peerings` endpoint.
4. Also fixed up a `route` > `routeName` typo whilst I was here

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

/cc @LevelbossMike


LevelbossMike:
:+1: lgtm

johncowen:
Merged https://github.com/hashicorp/consul/pull/13651 down onto here post approval, merging this one down also too!

PR #13649, opened 2022-06-30
proxycfg: check Ingress Gateway watches during discovery chain

adrien-f:
### Description

Greetings, I'm not sure if you still accept fixes for the v1.10 branch but I thought you might find this fix useful.

We've been suffering from a memory leak on our Consul Clients hosting the Ingress Gateway service and only them. After investigation, we found out that the number of goroutines was ever-growing:

image

We went to take some profiling dumps and found out a lot of goroutines were parked:

image

And the heap was taken by `watchDiscoveryChain` and `notifyBlockingQuery`:

image

This leads us to look into `agent/proxycfg/state.go`:

https://github.com/hashicorp/consul/blob/de95b6f546f7f1b891cfd596dd4e3293a3a3abcd/agent/proxycfg/state.go#L1819-L1851

I immediately see at the beginning a key existence check for `snap.ConnectProxy.WatchedDiscoveryChains`, and then, when creating the watch request, an assignment in either `snap.IngressGateway.WatchedDiscoveryChains` or
`snap.ConnectProxy.WatchedDiscoveryChains`.

This made us add the extra check for existence in `snap.IngressGateway.WatchedDiscoveryChains`. Since then, we've observed a pretty stable goroutines amount and no more OOMKill:

image 

We are not in the right configuration to upgrade yet from v1.10 which is why we hope this could still be useful for the community if anyone else is also experiencing this issue.

### Links

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/proxycfg/state.go
+++ b/agent/proxycfg/state.go
@@ -1821,6 +1821,10 @@ func (s *state) watchDiscoveryChain(snap *ConfigSnapshot, opts discoveryChainWat
 		return nil
 	}
 
+	if _, ok := snap.IngressGateway.WatchedDiscoveryChains[opts.id]; ok {
+		return nil
+	}
+
 	ctx, cancel := context.WithCancel(s.ctx)
 	err := s.cache.Notify(ctx, cachetype.CompiledDiscoveryChainName, &structs.DiscoveryChainRequest{
 		Datacenter:             s.source.Datacenter,


Amier3:
Hey @adrien-f 

Thank you for this PR and the detailed explanation ( that always helps! ). I'll have to get an answer back to you if we support changes for this branch. I'm pretty sure we do, but i'll have to confirm. While I figure that out for you could you sign the CLA agreement so we can add this to the queue 😄 ?

adrien-f:
👋 Hey @Amier3 , sorry it took some time to get the CLA approval with the summer holidays!

PR #13653, opened 2022-07-01
peering: count imported services

acpana:

### Description
For now, we are choosing to track the imported services count in the stream tracker shim. In the future, the tracking should make its way through raft.

This information will eventually be used by the UI.

* stream tracker testing updated

#### pr todo

* [ ] reset counter on streamTracker `.delete()`
* [ ] fix 32bit test

### DIFF
--- a/agent/rpc/peering/replication.go
+++ b/agent/rpc/peering/replication.go
@@ -129,6 +129,7 @@ func marshalToProtoAny[T proto.Message](in any) (*anypb.Any, T, error) {
 }
 
 func (s *Service) processResponse(
+	peerLocalID string,
 	peerName string,
 	partition string,
 	resp *pbpeering.ReplicationMessage_Response,
@@ -155,7 +156,7 @@ func (s *Service) processResponse(
 			), err
 		}
 
-		if err := s.handleUpsert(peerName, partition, resp.ResourceURL, resp.ResourceID, resp.Resource); err != nil {
+		if err := s.handleUpsert(peerLocalID, peerName, partition, resp.ResourceURL, resp.ResourceID, resp.Resource); err != nil {
 			return makeReply(
 				resp.ResourceURL,
 				resp.Nonce,
@@ -167,7 +168,7 @@ func (s *Service) processResponse(
 		return makeReply(resp.ResourceURL, resp.Nonce, code.Code_OK, ""), nil
 
 	case pbpeering.ReplicationMessage_Response_DELETE:
-		if err := s.handleDelete(peerName, partition, resp.ResourceURL, resp.ResourceID); err != nil {
+		if err := s.handleDelete(peerLocalID, peerName, partition, resp.ResourceURL, resp.ResourceID); err != nil {
 			return makeReply(
 				resp.ResourceURL,
 				resp.Nonce,
@@ -194,6 +195,7 @@ func (s *Service) processResponse(
 }
 
 func (s *Service) handleUpsert(
+	peerLocalID string,
 	peerName string,
 	partition string,
 	resourceURL string,
@@ -210,7 +212,7 @@ func (s *Service) handleUpsert(
 			return fmt.Errorf("failed to unmarshal resource: %w", err)
 		}
 
-		return s.handleUpdateService(peerName, partition, sn, csn)
+		return s.handleUpdateService(peerLocalID, peerName, partition, sn, csn)
 
 	case pbpeering.TypeURLRoots:
 		roots := &pbpeering.PeeringTrustBundle{}
@@ -234,11 +236,13 @@ func (s *Service) handleUpsert(
 //		- A reconciliation against nil or empty input pbNodes leads to deleting all stored catalog resources
 //		  associated with the service name.
 func (s *Service) handleUpdateService(
+	peerLocalID string,
 	peerName string,
 	partition string,
 	sn structs.ServiceName,
 	pbNodes *pbservice.IndexedCheckServiceNodes,
 ) error {
+	logger := s.logger.With("peerName", peerName, "peerID", peerLocalID)
 	// Capture instances in the state store for reconciliation later.
 	_, storedInstances, err := s.Backend.Store().CheckServiceNodes(nil, sn.Name, &sn.EnterpriseMeta, peerName)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


agent/rpc/peering/replication.go
```diff
@@ -265,6 +269,16 @@ func (s *Service) handleUpdateService(
 			req.Service = svcSnap.Service
 			if err := s.Backend.Apply().CatalogRegister(&req); err != nil {
 				return fmt.Errorf("failed to register service: %w", err)
+			} else {
+				if !servicePresentInStored(req.Service, storedInstances) {
```
mkeeler, line 273: Doing this in the loop will overcount the services. We would do the counting here if we wanted to measure the number of service *instances*.

The same thing goes for the delete.

I think something like the following at the very end of this method would work:

```
if len(storedInstances) > 0 && len(pbNodes.GetNodes()) < 1 {
   // decrement the service count as it was removed
} else if len(storedInstances) < 1 && len(pbNodes.GetNodes()) > 0 {
   // increment the service count as no instances previously existed but they do now.
}
```

agent/rpc/peering/service.go
```diff
@@ -830,7 +830,7 @@ func (s *Service) HandleStream(req HandleStreamRequest) error {
 
 			if resp := msg.GetResponse(); resp != nil {
 				// TODO(peering): Ensure there's a nonce
-				reply, err := s.processResponse(req.PeerName, req.Partition, resp)
```
mkeeler, line 833: The PeeringList and PeeringReady endpoints in this file also need updating so that the count is actually emitted in the API responses.

agent/rpc/peering/replication.go
```diff
@@ -450,13 +473,22 @@ func (s *Service) handleDelete(
...
+		return s.handleUpdateService(peerLocalID, peerName, partition, sn, nil)
 
 	default:
 		return fmt.Errorf("unexpected resourceURL: %s", resourceURL)
 	}
 }
 
+func servicePresentInStored(svc *structs.NodeService, stored structs.CheckServiceNodes) bool {
+	for _, csn := range stored {
+		if svc.CompoundServiceName() == csn.Service.CompoundServiceName() {
```
kisunji, line 485: Super-nit: calling `CompoundServiceName` makes allocations per iteration. Would be more efficient to capture `svc.CompoundServiceName()` in a variable to compare

kisunji (commented):
Some minor comments; I would _really_ prefer not to pass peerID if not necessary.

The 32-bit test failures are a fun one! Basically all the 64-bit operations you added require the field to be 64-bit aligned on 32-bit environments ([source](https://pkg.go.dev/sync/atomic#pkg-note-BUG)). IIRC, the easiest solution is to make `ImportedServicesCount` the first field in `StreamStatus`

freddygv (changes requested):
Let's chat about this when you get back, I'm not sure we should be hooking into the stream handler for imported service counts when they could be queried through the catalog instead

agent/rpc/peering/service.go
acpana: yes, that was going to come after getting this round of feedback (even potentially as a next PR)

agent/rpc/peering/stream_tracker.go
```diff
@@ -85,6 +86,17 @@ func (t *streamTracker) deleteStatus(id string) {
 	delete(t.streams, id)
 }
 
+func (t *streamTracker) getLockableStatus(id string) (*lockableStreamStatus, error) {
```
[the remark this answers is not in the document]
acpana: this is meant to get the `LockableStatus` rather than the `Status`; after the refactor, this struct is called `MutableStatus`

agent/rpc/peering/stream_tracker.go
```diff
@@ -210,3 +225,15 @@ func (s *lockableStreamStatus) status() StreamStatus {
...
+	return atomic.LoadUint64(&s.ImportedServicesCount)
+}
+
+func (s *lockableStreamStatus) incrementImportedServicesCount() {
+	atomic.AddUint64(&s.ImportedServicesCount, 1)
+}
+
+func (s *lockableStreamStatus) decrementImportedServicesCount() {
+	atomic.AddUint64(&s.ImportedServicesCount, ^uint64(0))
+}
```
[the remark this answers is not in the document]
acpana: that's a good point; I am going with a new approach we are going with now, we probably want to have these in the `MutableStatus`

agent/rpc/peering/replication.go
acpana: that's a good point! I'm doing away with this since we don't want to count at the service instances.

agent/rpc/peering/replication.go
```diff
@@ -323,6 +337,14 @@ func (s *Service) handleUpdateService(
 			})
 			if err != nil {
 				return fmt.Errorf("failed to deregister service %q: %w", csn.Service.CompoundServiceID(), err)
+			} else {
+				logger.Trace("decrementing imported services count", "serviceName", csn.Service.CompoundServiceName(), "node", csn.Node.Node)
+				st, err := s.streams.getLockableStatus(peerLocalID)
+				if err != nil {
+					logger.Warn("failed to decrement imported services count", "error", err)
```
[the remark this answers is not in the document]
acpana: I think the message here should be that we abandon the `increment/ decrement` operation for this operation

agent/rpc/peering/replication.go
```diff
@@ -234,11 +236,13 @@ func (s *Service) handleUpsert(
 //		- A reconciliation against nil or empty input pbNodes leads to deleting all stored catalog resources
 //		  associated with the service name.
 func (s *Service) handleUpdateService(
+	peerLocalID string,
```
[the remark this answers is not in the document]
acpana: I can use the state store!

acpana:
hey all, thanks for the feedback 🎉 🥇 ; with the refactor that just happened, I decided to follow up on the feedback here in a separate PR:

https://github.com/hashicorp/consul/pull/13718

I will still keep this one around while the above is in draft.

agent/rpc/peering/service.go
acpana: I added a reconciliation here: https://github.com/hashicorp/consul/pull/13718

acpana:
alright folks, I think I have addressed all the feedback offered in this new PR: https://github.com/hashicorp/consul/pull/13718 

I'll be closing this PR since it has too many conflicts.

PR #13654, opened 2022-07-01
docs: improve large version change upgrade path

jkirschner-hashicorp:
The existing large version change upgrade docs present the following challenges:
- They mention upgrading to 1.10.0 and then the current version, even though users must [first upgrade to 1.10.7+](https://www.consul.io/docs/upgrading/upgrade-specific#1-10-compatibility) before upgrading to 1.11+
- (less importantly) They mention upgrading to release series which are not the latest version in their series

This PR is intended to make the ideal upgrade path clearer.

### DIFF
--- a/website/content/docs/upgrading/instructions/index.mdx
+++ b/website/content/docs/upgrading/instructions/index.mdx
@@ -9,30 +9,39 @@ description: >-
 # Upgrade Instructions
 
 This document is intended to help users who find themselves many versions behind to upgrade safely.
-Our recommended upgrade path is moving from version 0.8.5 to 1.2.4 to 1.6.9 to 1.8.13 to the current version. To get
-started, you will want to choose the version you are currently on below and then follow the instructions
-until you are on the latest version. The upgrade guides will mention notable changes and link to relevant
-changelogs – we recommend reviewing the changelog for versions between the one you are on and the
-one you are upgrading to at each step to familiarize yourself with changes.
+
+## Upgrade Path
+
+Our recommended upgrade path is to move through the following sequence of versions:
+- 0.8.5 (final 0.8.x)
+- 1.2.4 (final 1.2.x)
+- 1.6.10 (final 1.6.x)
+- 1.8.19 (final 1.8.x)
+- Latest 1.10.x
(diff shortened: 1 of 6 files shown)


website/content/docs/upgrading/instructions/index.mdx
```diff
@@ -9,30 +9,39 @@ description: >-
...
 ## Getting Started
 
-To get instructions for your upgrade, please choose the release series you are _currently using_:
+To get instructions for your upgrade, follow the instructions given below for
+your _currently installed_ release series until you are on the latest current version.
+The upgrade guides will mention notable changes and link to relevant changelogs –
+we recommend reviewing the changelog for versions between the one you are on and the
+one you are upgrading to at each step to familiarize yourself with changes.
 
-- [0.8.x](/docs/upgrading/instructions/upgrade-to-1-2-x)
```
jkirschner-hashicorp, line 22: I reversed the order from ascending to descending versions. Users are more likely to be on higher versions (e.g., 1.6.x - 1.8.x) than lower versions (e.g., 0.8.x - 1.2.x), so this makes it faster to find the version relevant to you. This change becomes increasingly important as the version list grows further.

mkeeler: Should 1.10.x and 1.11.x docs be added to tell you how to migrate to 1.12? Given that we are updating the Upgrade Path section to mention 1.11 and 1.12 it seems like not having it here isn't great. That is obviously quite a bit more work so if we wanted to have that be some follow on work I would be fine with that too.

mkeeler (approved):

website/content/docs/upgrading/instructions/index.mdx
jkirschner-hashicorp: I totally agree that's needed, but I'd rather get this current PR out ASAP and have someone address that separately.

We probably want to complete that follow-on work before the release of 1.13 (unless upgrading straight from latest 1.10.x straight to 1.13.x is recommended).

im2nguyen (approved):
looks good to me!

PR #13655, opened 2022-07-01
docs: add Envoy upgrade step to std upgrade docs

jkirschner-hashicorp:
Upgrading Envoy instance versions was not previously included in the upgrade docs.

Someone from engineering should review this in addition to a member of the docs team.

### DIFF
--- a/website/content/docs/upgrading/index.mdx
+++ b/website/content/docs/upgrading/index.mdx
@@ -34,14 +34,22 @@ Consul is A, and version B is released.
    there are no compatibility issues that will affect your workload. If there
    are plan accordingly before continuing.
 
-2. On each server, install version B of Consul.
+2. On each Consul server agent, install version B of Consul.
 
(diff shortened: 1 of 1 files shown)


im2nguyen (approved):
this looks good to me -- i like including a link to the compatible envoy version

kyhavlov (approved):
This looks right - updating the envoy version first should be safe, as long as it's a common supported version between the two consul versions used (like envoy 1.19.x with consul 1.11 and 1.12).

jkirschner-hashicorp:
>  as long as it's a common supported version between the two consul versions used (like envoy 1.19.x with consul 1.11 and 1.12)

@kyhavlov : the text doesn't currently call that out, and that seems like an important detail. I'll adjust.

jkirschner-hashicorp:
@im2nguyen : FYI, I updated the text for step 4 (around upgrading Envoy proxies) after speaking with @kyhavlov. It now looks like:

image

kyhavlov (approved):

PR #13659, opened 2022-07-01
Replace deprecated ioutil pkg with os & io

abhinavnair:
### Description
As of Go 1.16, the same functionality is now provided by package io or
package os, and those implementations should be preferred in new code.

So replacing all usage of `ioutil` pkg with `io` & `os`.

N/A

### Links
N/A

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net"
 	"net/http"
 	"os"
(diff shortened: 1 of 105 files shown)


Amier3:
Hey @abhinavnair 

Thanks for this PR, the help is much appreciated. We're working on a few rather large PRs at the moment, so this may take a bit of time for us to get to. Please let us know when this is out of draft so we can add it to the queue 😄 .

Also linking the deprecation [docs](https://go.dev/doc/go1.16#ioutil ) for the future

PR #13662, opened 2022-07-04
ui: Fixup peering imported/exported service counts

johncowen:
### Description
1. Use `format-number` helper for humanized counts
2. Remove truthy check for displaying the counts, we want to make sure we report zero counts also.
3. Fixup `name` > `Name` typo for displaying the name of the peer in the tooltips

I tested this on a real cluster, and bumped into an issue where INITIAL was still being reported as a state. A backend PR for that has subsequently been merged, so theres an extra commit on the end here to remove what I had originally done as a little fix up. You can probably ignore this commit and this fact.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

/cc @LevelbossMike


LevelbossMike:
Thanks for putting this together. Lgtm 👍

natmegs (approved):

PR #13665, opened 2022-07-05
ui: Peer Deletion

johncowen:
### Description
Enable deletion of peers from the 'actions menu' in each row.

Note: we are using our newer 'use less Ember Data' approach to CRUD here, and this is the first time we need to do write-like things with this approach.

This PR includes:

1. Inject our `client.request` into data-sink so we can then inject that into the methods themselves via an argument. This is exactly the same as how our data-sources work, we just hadn't had reason to do this corresponding data-sink plumbing as yet. Just a bit of FYI, this exact method signature is probably temporary until we get everything moved over, or we end up moving to something else entirely first.
2. Move the peer listing into a component. Although the peer list is currently "single use", moving it to a component means we can easily add a documentation page and work on the component in isolation. It also make is much easier to find and work with as the functionality grows. All DataLoading and DataSaving happens outside the component using a data-down/actions-up approach, thus making the list itself just a presentational component who's data and functionality could be controlled by our Data* components or anything else.
3. Add the "Delete" action to each peer row, which is just copied and pasted from elsewhere.
4. Use our DataWriter component to actually delete and deal with the notification for that.

Please note, the majority of this code is just copy and pasted from elsewhere.

The "soft delete" functionality is also mostly copied from elsewhere but there is one additional change with the soft delete, as we don't have our blocking query support for peers. Therefore we 'assume' that the backend has successfully marked the peer for deletion when we receive a successful request and therefore pre-emptively update the peers' `State` on the frontend.

Manually tested both in dev/mocks mode and with a real binary/peers.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

/cc @LevelbossMike

### DIFF
--- a/ui/packages/consul-ui/app/components/consul/peer/list/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/peer/list/index.hbs
@@ -62,7 +62,7 @@ as |item index|>
       </Action>
       <Action
         data-test-delete-action
-        @onclick={{action @ondelete item}}
+        @onclick={{fn @ondelete item}}
         class="dangerous"
       >
         <BlockSlot @name="label">
--- a/ui/packages/consul-ui/app/components/consul/peer/list/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/peer/list/index.hbs
@@ -1,0 +1,93 @@
+<ListCollection
+  class="consul-peer-list"
+  ...attributes
+  @items={{@items}}
+  @linkable="linkable peer"
+as |item index|>
+  <BlockSlot @name="header">
+{{#if (can 'delete peer' item=item)}}
+      <a
+        data-test-peer={{item.Name}}
+        href={{href-to 'dc.peers.edit' item.Name}}
+      >
+        {{item.Name}}
+      </a>
+{{else}}
+      <p>
+        {{item.Name}}
+      </p>
+{{/if}}
+  </BlockSlot>
+  <BlockSlot @name="details">
+    <div class="peers__list__peer-detail">
+      <Peerings::Badge @peering={{item}} />
+
+      <div
+        {{tooltip
+          (t 'routes.dc.peers.index.detail.imported.tooltip'
+            name=item.Name
+          )
+        }}
+      >
+        {{t 'routes.dc.peers.index.detail.imported.count'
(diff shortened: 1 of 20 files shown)


ui/packages/consul-ui/app/components/consul/peer/list/index.hbs
```diff
@@ -0,0 +1,93 @@
...
+        data-test-edit-action
+        @href={{href-to 'dc.peers.edit' item.Name}}
+      >
+        <BlockSlot @name="label">
+          View
+        </BlockSlot>
+      </Action>
+      <Action
+        data-test-delete-action
+        @onclick={{action @ondelete item}}
```
LevelbossMike, line 65: `{{fn @onDelete item}}` is more idiomatic in Octane

ui/packages/consul-ui/app/templates/dc/peers/index.hbs
```diff
@@ -58,111 +58,86 @@ as |sort filters items|}}
...
-                </a>
-              </li>
-            </BlockSlot>
-          </EmptyState>
-        </collection.Empty>
-      </DataCollection>
+        @label="Peer"
+        @ondelete={{refresh-route}}
+      as |writer|>
+        <BlockSlot @name="removed" as |after|>
```
LevelbossMike, line 73: who's defining this block-slot?

ui/packages/consul-ui/app/components/consul/peer/list/index.hbs
johncowen: Do you know if there is a codemod to do this? I'd like to get rid of all our `action` things now if possible, now that we have `pick` we don't need `value=""` ability that `action` has.

In the meantime will switch this one up, it was just copied from elsewhere.

LevelbossMike:
With the idea of landing this asap I'm 👍 - we can get back to looking at the solution in depth when I'm back.

ui/packages/consul-ui/app/components/consul/peer/list/index.hbs
johncowen: Switched up this one `action` > `fn`

johncowen:
@natmegs will ping you offline also but I think this one is good to go now.

PR #13666, opened 2022-07-05
Preserve PeeringState on upsert

kisunji:
### Description
Fixes a bug where if the generate token is called twice, the second call upserts the zero-value (undefined) of PeeringState.

* Added a test suite for calling generate token at various points in a peer's lifecycle

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/consul/state/peering.go
+++ b/agent/consul/state/peering.go
@@ -214,7 +214,7 @@ func (s *Store) PeeringWrite(idx uint64, p *pbpeering.Peering) error {
 		}
 
 		if p.State == 0 {
-			p.State = pbpeering.PeeringState_PENDING
+			return fmt.Errorf("missing peering state when updating existing peering")
 		}
 		// TODO(peering): Confirm behavior when /peering/token is called more than once.
 		// We may need to avoid clobbering existing values.
--- a/agent/consul/state/peering.go
+++ b/agent/consul/state/peering.go
@@ -213,6 +213,13 @@ func (s *Store) PeeringWrite(idx uint64, p *pbpeering.Peering) error {
 			return fmt.Errorf("cannot write to peering that is marked for deletion")
 		}
 
+		if p.State == 0 {
+			p.State = pbpeering.PeeringState_PENDING
+		}
+		// TODO(peering): Confirm behavior when /peering/token is called more than once.
+		// We may need to avoid clobbering existing values.
+		p.ImportedServiceCount = existing.ImportedServiceCount
+		p.ExportedServiceCount = existing.ExportedServiceCount
 		p.CreateIndex = existing.CreateIndex
 		p.ModifyIndex = idx
 	} else {
--- a/agent/consul/state/peering.go
+++ b/agent/consul/state/peering.go
@@ -213,7 +213,7 @@ func (s *Store) PeeringWrite(idx uint64, p *pbpeering.Peering) error {
 			return fmt.Errorf("cannot write to peering that is marked for deletion")
 		}
 
-		if p.State == 0 {
+		if p.State == pbpeering.PeeringState_UNDEFINED {
 			p.State = existing.State
 		}
 		// TODO(peering): Confirm behavior when /peering/token is called more than once.
--- a/agent/consul/state/peering.go
+++ b/agent/consul/state/peering.go
@@ -214,7 +214,7 @@ func (s *Store) PeeringWrite(idx uint64, p *pbpeering.Peering) error {
 		}
 
 		if p.State == 0 {
-			return fmt.Errorf("missing peering state when updating existing peering")
+			p.State = existing.State
 		}
 		// TODO(peering): Confirm behavior when /peering/token is called more than once.
 		// We may need to avoid clobbering existing values.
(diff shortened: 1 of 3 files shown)


agent/consul/state/peering.go
```diff
@@ -212,6 +212,13 @@ func (s *Store) PeeringWrite(idx uint64, p *pbpeering.Peering) error {
 			return fmt.Errorf("cannot write to peering that is marked for deletion")
 		}
 
+		if p.State == 0 {
+			p.State = pbpeering.PeeringState_PENDING
```
rboyer, line 216: Shouldn't the caller be setting the state to a nonzero value on the re-peer?

kisunji: Hmm the caller naively does a write but I suppose it should introspect the current state a bit more

agent/rpc/peering/service.go
```diff
@@ -178,6 +178,52 @@ func (s *Server) GenerateToken(
 		return resp, err
 	}
 
+	var peering *pbpeering.Peering
```
kisunji, line 181: this hunk replaces the original RETRY_ONCE block below with some changes:
1: Uses a for-loop instead of `goto`
2: Renames `peeringOrNil` to `peering` and puts the nil-check explicitly upfront

agent/rpc/peering/service.go
```diff
@@ -194,56 +240,6 @@ func (s *Server) GenerateToken(
...
-		Value:          strings.ToLower(req.PeerName),
-		EnterpriseMeta: *structs.NodeEnterpriseMetaInPartition(req.Partition),
-	}
-	_, peering, err := s.Backend.Store().PeeringRead(nil, q)
-	if err != nil {
-		return nil, err
-	}
-	if peering == nil {
-		return nil, fmt.Errorf("peering was deleted while token generation request was in flight")
-	}
```
kisunji, line 245: I removed this check for the following reasons:
* Very unlikely scenario
* Peering may still be deleted after token generation

We can never eliminate this class of bug (it will eventually be caught by the dialer) so it's better to remove the false sense of security IMO. Let me know if there's reasons to keep it.

agent/rpc/peering/service.go
```diff
@@ -679,9 +668,9 @@ func (s *Server) getExistingPeering(peerName, partition string) (*pbpeering.Peer
 //
 // We define a DIALER as a peering that has server addresses (or a peering that is created via the Establish endpoint)
 // Conversely, we define an ACCEPTOR as a peering that is created via the GenerateToken endpoint
-func validatePeer(peering *pbpeering.Peering, allowedToDial bool) error {
-	if peering != nil && peering.ShouldDial() != allowedToDial {
-		if allowedToDial {
+func validatePeer(peering *pbpeering.Peering, shouldDial bool) error {
+	if peering != nil && peering.ShouldDial() != shouldDial {
+		if shouldDial {
```
kisunji, line 673: By aligning the var name with `peering.ShouldDial` it's easier to understand at a glance

agent/consul/state/peering.go
```diff
@@ -213,6 +213,13 @@ func (s *Store) PeeringWrite(idx uint64, p *pbpeering.Peering) error {
 			return fmt.Errorf("cannot write to peering that is marked for deletion")
 		}
 
+		if p.State == 0 {
+			p.State = existing.State
+		}
+		// TODO(peering): Confirm behavior when /peering/token is called more than once.
+		// We may need to avoid clobbering existing values.
+		p.ImportedServiceCount = existing.ImportedServiceCount
+		p.ExportedServiceCount = existing.ExportedServiceCount
```
kisunji, line 222: This is the actual bugfix; the rest are refactors

agent/rpc/peering/service.go
```diff
@@ -227,57 +274,6 @@ func (s *Server) GenerateToken(
...
-		Value:          strings.ToLower(req.PeerName),
-		EnterpriseMeta: *entMeta,
-	}
-	_, peering, err := s.Backend.Store().PeeringRead(nil, q)
-	if err != nil {
-		return nil, err
-	}
-	if peering == nil {
-		return nil, fmt.Errorf("peering was deleted while token generation request was in flight")
-	}
```
kisunji, line 279: I removed this check for the following reasons:
* Very unlikely scenario
* Peering may still be deleted after token generation

We can never eliminate this class of bug (it will eventually be caught by the dialer) so it's better to remove the false sense of security IMO. Let me know if there's reasons to keep it.

agent/consul/state/peering.go
```diff
@@ -213,6 +213,13 @@ func (s *Store) PeeringWrite(idx uint64, p *pbpeering.Peering) error {
 			return fmt.Errorf("cannot write to peering that is marked for deletion")
 		}
 
+		if p.State == 0 {
```
mkeeler, line 216: nit

```suggestion
		if p.State == pbpeering.PeeringState_UNDEFINED {
```

agent/rpc/peering/service.go
mkeeler: I LOVE that you got rid of the goto.

mkeeler (approved):
1 minor nit but otherwise LGTM.

PR #13683, opened 2022-07-07
Update index.mdx

ml4:
Typo, no links presented.  Removed verbiage.


trujillo-adam (approved):
This is fine for now, but I think in the future we should get out of the habit of referencing design/navigational elements on the page (i.e., sidebar). Thanks for making this change.

PR #13685, opened 2022-07-07
ui: Add filtering peer listing by State

johncowen:
### Description
Adds filtering peer listing by State

Screenshot 2022-07-07 at 12 45 39

Please note there are a few things that may be tweaked here moving forwards, i.e. specific icons etc. Right now we are just matching the previously used icons.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

/cc @LevelbossMike

### DIFF
--- a/ui/packages/consul-ui/app/components/consul/peer/search-bar/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/peer/search-bar/index.hbs
@@ -64,6 +64,38 @@ as |key value|}}
         </search.Select>
       </search.Search>
     </:search>
+    <:filter as |search|>
+      <search.Select
+        class="type-state"
+        @position="left"
+        @onchange={{action @filter.state.change}}
+        @multiple={{true}}
+      as |components|>
+        <BlockSlot @name="selected">
+          <span>
+            {{t "components.consul.peer.search-bar.state.name"}}
+          </span>
+        </BlockSlot>
+        <BlockSlot @name="options">
+  {{#let components.Optgroup components.Option as |Optgroup Option|}}
+    {{#each (array
+      'pending'
+      'establishing'
+      'active'
+      'failing'
+      'terminated'
+      'deleting'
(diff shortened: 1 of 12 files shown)


ui/packages/consul-ui/app/components/consul/peer/search-bar/index.hbs
```diff
@@ -64,6 +64,38 @@ as |key value|}}
...
+        </BlockSlot>
+        <BlockSlot @name="options">
+  {{#let components.Optgroup components.Option as |Optgroup Option|}}
+    {{#each (array
+      'pending'
+      'establishing'
+      'active'
+      'failing'
+      'terminated'
+      'deleting'
```
natmegs, line 87: Def not blocking but I wonder if an enum could help here as a single source of truth for these states (to ensure they are in sync between here and predicates/peer for example)?

natmegs (approved):
I"m probably doing something wrong but I'm seeing some overlap in the peers table?
Screen Shot 2022-07-07 at 8 27 03 AM

johncowen:
Hmm, that definitely doesn't look right 😆 thanks for spotting, let me take a look at this again brb

ui/packages/consul-ui/app/components/consul/peer/search-bar/index.hbs
johncowen: Yeah agree completely, I tried to mostly centralise these in the model (I added an exported `schema`) in there, but as I'm mostly implementing (i.e. copy/pasting how its done elsewhere) I didn't put to much thought into doing anything extra. Agree its not a blocker for this, but let me have a think about this also 🤔

johncowen: I reused our `schema.State.allowedValues` array here (imported from our model files), just waiting for tests to pass again, once thats done I'll merge away

PR #13687, opened 2022-07-07
server: broadcast the public grpc port using lan serf and update the consul service in the catalog with the same data

rboyer:
### Description

Currently servers exchange information about their WAN serf port and RPC port with serf tags, so that they all learn of each other's addressing information. We intend to make larger use of the new public-facing gRPC port exposed on all of the servers, so this PR addresses that by passing around the gRPC port via serf tags and then ensuring the generated `consul` service in the catalog has metadata about that new port as well for ease of non-serf-based lookup.

### DIFF
--- a/agent/metadata/server_test.go
+++ b/agent/metadata/server_test.go
@@ -4,6 +4,7 @@ import (
 	"net"
 	"testing"
 
+	"github.com/hashicorp/go-version"
 	"github.com/hashicorp/serf/serf"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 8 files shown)


agent/metadata/server_test.go
```diff
@@ -53,173 +54,136 @@ func TestServer_Key_params(t *testing.T) {
 }
 
 func TestIsConsulServer(t *testing.T) {
```
rboyer, line 56: I rewrote this test to be easier to add cases to and folded the following optional one in.

mkeeler (approved):

PR #13699, opened 2022-07-08
Respect http2 protocol for upstreams of terminating gateways

kyhavlov:
This PR fixes a bug with terminating gateway upstreams that have `http2` set as the protocol. Currently, the terminating gateway won't communicate to the upstream correctly with the http2 protocol options set - this PR adds that logic in `agent/xds/clusters.go`


mkeeler (approved):

eculver (approved):
LGTM - tests too. I see that the `type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions` is getting set now which is important.

PR #13701, opened 2022-07-11
ui: Remove horizontal scrollbar from peering list rows

johncowen:
### Description
In https://github.com/hashicorp/consul/pull/13685#pullrequestreview-1031744420 we noticed that a horizontal scroll had been added to the peering list rows in https://github.com/hashicorp/consul/pull/13425. This had the effect of pushing rendering/stacking out of place if you have scrollbars showing permanently.

This PR removes the CSS that was causing this.

Additionally, I noticed that I could move the rest of the CSS for the peer listing alongside the component that we now have for the listing so it's easier to find, so I moved the other related CSS other there.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

/cc @LevelbossMike

### DIFF
--- a/ui/packages/consul-ui/app/components/consul/peer/index.scss
+++ b/ui/packages/consul-ui/app/components/consul/peer/index.scss
@@ -1,4 +1,5 @@
 @import './components';
 
+@import './list';
 @import './search-bar';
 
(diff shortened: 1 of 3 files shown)


LevelbossMike (approved):
certainly a weird issue that ember-collection doesn't seem to be able to handle scrollbars properly. 👍 for fixing the issue - we probably want to get back to this at some point thought and fix the vertical-scroll handling for long horizontal content.

evrowe (approved):

PR #13705, opened 2022-07-11
ui: Make our old TabNav component easily usable with a state machine

johncowen:
### Description
Update our super old TabNav component to easily work with a state machine.

This approach was chosen to be the easiest to use right now, whilst keeping the old interface of the component, which is used in many other places in the UI. At some point we can upgrade the component completely, and then replace all the usages of it in our app in one go. The time to do that is not right now.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

/cc @LevelbossMike

### DIFF
--- a/ui/packages/consul-ui/app/components/tab-nav/index.hbs
+++ b/ui/packages/consul-ui/app/components/tab-nav/index.hbs
@@ -32,7 +32,7 @@ as |select name|}}
     <Action
       {{on 'click'
         (if (not-eq @onclick undefined)
-          (fn @onclick (uppercase item.label))
+          (fn @onclick (or item.state (uppercase item.label)))
           (noop)
         )
       }}
--- a/ui/packages/consul-ui/app/components/tab-nav/README.mdx
+++ b/ui/packages/consul-ui/app/components/tab-nav/README.mdx
@@ -11,7 +11,9 @@ Each item in the list should be a hash of `label`, `href` and `selected`.
 - `label`: The text to show
 - `href`: a href, probably generated via `href-to`
 - `selected`: whether the item is in the selected state or not, probably
-  generated via `is-href`
+- `state`: when used with an @onclick using `StateMachine.dispatch`, the name
+  of the state to transition to.
+
 
 **Please note:** This component should probably be rebuilt using contextual
 components, alternatively this could be hand built with native HTML using the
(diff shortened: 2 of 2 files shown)


ui/packages/consul-ui/app/components/tab-nav/index.hbs
```diff
@@ -32,7 +32,7 @@ as |select name|}}
     <Action
       {{on 'click'
```
LevelbossMike, line 33: how about we pass an additional `{{on "click"}}` here and call the action differently, passing the entire item instead - just adding this change to be able to dispatch an event to a statechart feels less future proof to me. what if we wanted to do something else based on the item?

```hbs

<figure>
  <figcaption>A TabNav with using a StateMachine.dispatch</figcaption>
  <TabNav @items={{
    compact
      (array
        (hash
          label="1"
          selected=(state-matches fsm.state 'one')
          state="ONE"
        )
        (hash
          label="2"
          selected=(state-matches fsm.state 'two')
          state="TWO"
        )
      )
    }}
    @onTabClicked={{pick 'state' fsm.dispatch}}
  />
</figure>
```

johncowen: I would very much like to upgrade this `TabNav` entirely, and not just continue bolting things onto it. The current interface, including the passing of a string through to `@onclick` is being used elsewhere. So if we make you suggested change here now, we'd also have to go through and update everywhere we use this in the app already (if I follow correctly).

All in all, if we are going to go through the app and change how `TabNav `works (or replace it) I would rather make `TabNav` work exactly how we want first, and then change the app. Your suggestion here would probably be closer to how we want it to work, but I'm trying to avoid changing lots of unrelated things throughout the application.

Let me know if that make sense or not.

LevelbossMike: I was talking about adding an additional click handler not changing the existing one. you can register two click listeners that trigger when a tab-nav item is clicked. because both functions are optional you'd be able to use the new interface and the old one side by side.

What I'm suggesting is similar to this:

```hbs
    <Action
      {{on 'click'
        (if (not-eq @onclick undefined)
          (fn @onclick (uppercase item.label))
          (noop)
        )
      }}
      {{on 'click'
        (if @onTabClicked
          (fn @onTabClicked item)
          (noop)
        )
      }}
      @href={{item.href}}
    >
      {{item.label}}
    </Action>
```

johncowen: So have two arguments one called `@onclick` and one called `@onTabClick` and depending on which one you specify when the tab is clicked they emit different shaped things?

LevelbossMike: yes - add a new arg that gets called with the item when the click handler is called. that way the old interface stays the same and new invocations can use the new interface if need be.

johncowen: Gotcha, I do prefer this approach, and longer term I would like to move this to use real events now we have `pick` (gotta love `pick` 😄 ) I've made some changes here, let me know what you think.

LevelbossMike (approved):
:+1:

evrowe (approved):

PR #13709, opened 2022-07-11
proto: ensure buf formatter has been applied to protobufs

rboyer:
### Description

Ensure that the committed protobuf files always align with the current `buf` formatting rules, and also fix up two files that had drifted.


.circleci/config.yml
```diff
@@ -241,15 +241,17 @@ jobs:
       - run:
           name: Install protobuf
           command: make proto-tools
-
+      - run:
+          name: "Protobuf Format"
+          command: make proto-format
```
rboyer, line 246: By doing the format first, we can rely upon the `git diff` below to catch formatting drift free of charge.

kisunji (approved):

boxofrad (approved):

PR #13714, opened 2022-07-11
Return error if ServerAddresses is empty

kisunji:
### Description
Encountered an issue while tracking a regression in K8s acceptance tests where a dialer was passed a token with no ServerAddresses.

### DIFF
--- a/agent/consul/peering_backend.go
+++ b/agent/consul/peering_backend.go
@@ -71,7 +71,7 @@ func (b *PeeringBackend) GetServerAddresses() ([]string, error) {
 		addrs = append(addrs, node.Address+":"+grpcPortStr)
 	}
 	if len(addrs) == 0 {
-		return nil, fmt.Errorf("grpc port is not configured on any servers")
+		return nil, fmt.Errorf("a grpc bind port must be specified in the configuration for all servers")
 	}
 	return addrs, nil
 }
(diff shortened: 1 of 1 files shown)


freddygv (commented):
How did this come up? Was the gRPC port not being specified in the test [server config](https://www.consul.io/docs/agent/config/config-files#grpc_port)?

kisunji:
> How did this come up? Was the gRPC port not being specified in the test [server config](https://www.consul.io/docs/agent/config/config-files#grpc_port)?

I think K8s acceptance test setup didn't specify a grpc port on servers (only clients)

agent/consul/peering_backend.go
```diff
@@ -70,6 +70,9 @@ func (b *PeeringBackend) GetServerAddresses() ([]string, error) {
 		}
 		addrs = append(addrs, node.Address+":"+grpcPortStr)
 	}
+	if len(addrs) == 0 {
+		return nil, fmt.Errorf("grpc port is not configured on any servers")
```
freddygv, line 74: How about something a bit more strongly worded? We don't want users to end up in a situation where some servers have a gRPC port specified but not all

```suggestion
		return nil, fmt.Errorf("a grpc bind port must be specified in the configuration for all servers")
```

freddygv (approved):

PR #13718, opened 2022-07-12
peering: track imported services

acpana:
### Description

For now, we are choosing to track the imported services count in the stream tracker shim. In the future, the tracking should make its way through raft.

* stream tracker tests updated
* NOTE: `leader_peering_test` contains a more e2e test approach to peering sync; let's keep this test in the leader for now since the `PeeringList` call needs to always get served by the leader

### DIFF
--- a/agent/consul/leader_peering_test.go
+++ b/agent/consul/leader_peering_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/hashicorp/consul/proto/pbpeering"
 	"github.com/hashicorp/consul/sdk/testutil/retry"
 	"github.com/hashicorp/consul/testrpc"
+	"github.com/hashicorp/consul/types"
 )
 
 func TestLeader_PeeringSync_Lifecycle_ClientDeletion(t *testing.T) {
@@ -309,11 +310,6 @@ func insertTestPeeringData(t *testing.T, store *state.Store, peer string, lastId
 				Node:        "aaa",
 				PeerName:    peer,
 			},
-			{
-				CheckID:  structs.SerfCheckID,
-				Node:     "aaa",
-				PeerName: peer,
-			},
 		},
 	}))
 
@@ -335,11 +331,6 @@ func insertTestPeeringData(t *testing.T, store *state.Store, peer string, lastId
 				ServiceID:   "b-service-1",
 				Node:        "bbb",
 				PeerName:    peer,
-			},
-			{
-				CheckID:  structs.SerfCheckID,
-				Node:     "bbb",
-				PeerName: peer,
 			},
 		},
 	}))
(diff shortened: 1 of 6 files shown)


agent/grpc/public/services/peerstream/replication.go
```diff
@@ -240,12 +241,48 @@ func (s *Server) handleUpdateService(
...
+	logger := s.Logger.With("peerName", peerName, "peerID", peering.ID)
+
 	// Capture instances in the state store for reconciliation later.
 	_, storedInstances, err := s.GetStore().CheckServiceNodes(nil, sn.Name, &sn.EnterpriseMeta, peerName)
 	if err != nil {
 		return fmt.Errorf("failed to read imported services: %w", err)
 	}
 
+	// account for imported services
+	if len(storedInstances) > 0 && len(pbNodes.GetNodes()) < 1 {
```
freddygv, line 262: `len(pbNodes.GetNodes()) == 0` could be a bit clearer here

The [Go language spec](https://go.dev/ref/spec#Array_types) dictates the length of an array type won't be negative

agent/grpc/public/services/peerstream/replication.go
```diff
@@ -240,12 +241,48 @@ func (s *Server) handleUpdateService(
...
+		logger.Trace("decrementing imported services count", "serviceName", sn.String())
+
+		id := peering.GetID()
+		st, err := s.Tracker.MutableStreamStatus(id)
+		if err != nil {
+			logger.Trace("failed to find stream status, abandoning decrement for imported services count", "serviceName", sn.String())
+		} else {
+			st.RemoveImportedService(sn)
+		}
+	} else if len(storedInstances) < 1 && len(pbNodes.GetNodes()) > 0 {
```
freddygv, line 273: Same as above: `len(storedInstances) == 0`

agent/grpc/public/services/peerstream/stream_tracker.go
```diff
@@ -73,6 +75,18 @@ func (t *Tracker) StreamStatus(id string) (resp Status, found bool) {
...
 }
 
+// MutableStreamStatus gives a handle of the MutableStatus for a given peer ID
+func (t *Tracker) MutableStreamStatus(id string) (status *MutableStatus, err error) {
+	t.mu.RLock()
+	defer t.mu.RUnlock()
+
+	s, ok := t.streams[id]
+	if !ok {
+		return nil, fmt.Errorf("did not find a stream for id: %q", id)
```
freddygv, line 85: Can this return a boolean instead of an error for consistency with `StreamStatus`?

agent/grpc/public/services/peerstream/replication.go
```diff
@@ -240,12 +241,48 @@ func (s *Server) handleUpdateService(
...
 
+	// account for imported services
+	if len(storedInstances) > 0 && len(pbNodes.GetNodes()) < 1 {
+		// decrement the service count as the service was removed/ is being removed
+		logger.Trace("decrementing imported services count", "serviceName", sn.String())
+
+		id := peering.GetID()
+		st, err := s.Tracker.MutableStreamStatus(id)
+		if err != nil {
+			logger.Trace("failed to find stream status, abandoning decrement for imported services count", "serviceName", sn.String())
```
freddygv, line 269: I don't think that logging an error and moving forward is appropriate here

Take a look at what leads to creating/deleting entries in the stream tracker

freddygv (commented):
Thanks for updating this! Left mostly minor comments, with the most significant one being about the error handling

agent/grpc/public/services/peerstream/replication.go
acpana: that's fair; erroring out now

agent/grpc/public/services/peerstream/replication.go
```diff
@@ -240,12 +241,48 @@ func (s *Server) handleUpdateService(
...
 	pbNodes *pbservice.IndexedCheckServiceNodes,
 ) error {
+	_, peering, err := s.GetStore().PeeringRead(nil, state.Query{Value: peerName})
+	if err != nil {
+		return fmt.Errorf("failed to read peering: %w", err)
+	}
+	// an invariant here should be that a peering exists for PeeringRead
+	if peering == nil {
+		return fmt.Errorf("did not find peering %q", peerName)
+	}
```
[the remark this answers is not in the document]
acpana: Moved to state store fetch per suggestion here: https://github.com/hashicorp/consul/pull/13653#discussion_r913264741

agent/grpc/public/services/peerstream/stream_tracker.go
```diff
@@ -142,6 +156,10 @@ type Status struct {
 	// - The error message when we failed to store a resource replicated FROM the peer.
 	// - The last error message when receiving from the stream.
 	LastReceiveErrorMessage string
+
+	// TODO(peering): consider keeping track of imported service counts thru raft
+	// ImportedServices is set that keeps track of which service names are imported for the peer
+	ImportedServices map[string]struct{}
```
mkeeler, line 162: Why does this need to be a map instead of a simple int?

With an int, you could get rid of the `GetIMportedServicesCount` method. I would also renamed the other two functions to `IncrementImportedServiceCount` and `DecrementImportedServiceCount`

agent/grpc/public/services/peerstream/replication.go
```diff
@@ -240,12 +241,49 @@ func (s *Server) handleUpdateService(
 	sn structs.ServiceName,
 	pbNodes *pbservice.IndexedCheckServiceNodes,
 ) error {
+	_, peering, err := s.GetStore().PeeringRead(nil, state.Query{Value: peerName})
```
mkeeler, line 244: I think it would improve things to thread the `MutableStreamStatus` all the way through as a parameter to this function. You wouldn't have to access the stream tracker or state store in order to increment/decrement the service count.

One reason is that other bits of the various replication methods will also need access to the stream tracker so threading it through to the fn call here can serve a dual purpose.

agent/grpc/public/services/peerstream/replication.go
```diff
@@ -240,12 +241,49 @@ func (s *Server) handleUpdateService(
...
+	if err != nil {
+		return fmt.Errorf("failed to read peering: %w", err)
+	}
+	// an invariant here should be that a peering exists for PeeringRead
+	if peering == nil {
+		return fmt.Errorf("did not find peering %q", peerName)
+	}
+
+	logger := s.Logger.With("peerName", peerName, "peerID", peering.ID)
```
mkeeler, line 253: This should probably have a `partition` field as well if the parameter is not the empty string.

mkeeler (changes requested):

agent/grpc/public/services/peerstream/stream_tracker.go
acpana: IIUC, we were going to use a map for `ExportedServiceCount` and we wanted to use a map for consistency here, right? @freddygv (maybe I got this wrong)

per matt's comment above, I guess we could keep the imported services count an actual counter 🤔

agent/rpc/peering/service.go
```diff
@@ -337,7 +337,17 @@ func (s *Server) PeeringRead(ctx context.Context, req *pbpeering.PeeringReadRequ
 	if peering == nil {
 		return &pbpeering.PeeringReadResponse{Peering: nil}, nil
 	}
+
 	cp := copyPeeringWithNewState(peering, s.reconciledStreamStateHint(peering.ID, peering.State))
```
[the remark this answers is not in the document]
acpana: Could we revisit this when I come back around with `exported services` ?

agent/grpc/public/services/peerstream/replication.go
acpana: Added in two params:

* mutable stream status from handle upstream
* logger from handle upstream

agent/consul/leader_peering_test.go
```diff
@@ -363,13 +354,269 @@ func insertTestPeeringData(t *testing.T, store *state.Store, peer string, lastId
 				Node:        "ccc",
 				PeerName:    peer,
 			},
+		},
+	}))
+
+	return lastIdx
```
acpana, line 360: the diff tool is getting confused here fyi

agent/consul/leader_peering_test.go
```diff
@@ -336,11 +332,6 @@ func insertTestPeeringData(t *testing.T, store *state.Store, peer string, lastId
 				Node:        "bbb",
 				PeerName:    peer,
 			},
-			{
-				CheckID:  structs.SerfCheckID,
```
acpana, line 340: these checks will actually cause reaping of the nodes -- let's not use them at all

agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -301,8 +301,15 @@ func (s *Server) HandleStream(streamReq HandleStreamRequest) error {
 			}
 
 			if resp := msg.GetResponse(); resp != nil {
+
+				st, found := s.Tracker.MutableStreamStatus(streamReq.LocalID)
```
mkeeler, line 305: We already have the status from back on line 168 so there is no need to look it up again.

agent/grpc-external/services/peerstream/replication.go
```diff
@@ -219,14 +223,29 @@ func (s *Server) handleUpdateService(
 	peerName string,
 	partition string,
 	sn structs.ServiceName,
+	mutableStatus *MutableStatus,
 	pbNodes *pbservice.IndexedCheckServiceNodes,
+	logger hclog.Logger,
```
mkeeler, line 228: 🤔 I wonder if the logger for this stream should be stored in the stream tracker. This is fine for now, but I will think on this some more and maybe write up some future tasks to expand the role of the MutableStatus.

agent/grpc-external/services/peerstream/stream_test.go
```diff
@@ -1219,36 +1226,52 @@ func expectReplEvents(t *testing.T, client *MockClient, checkFns ...func(t *test
...
+		Name: peerName},
+	))
+
+	// connect the stream
+	_, err := srv.Tracker.Connected(peerID)
+	require.NoError(t, err)
+
 	run := func(t *testing.T, tc testCase) {
+
+		mst, found := srv.Tracker.MutableStreamStatus(peerID)
```
mkeeler, line 1265: This is fine but you wouldn't have needed for this to be in the stream tracker. For the purposes of checking for imported service counts it would have been fine to just create a new MutableStatus outside of the tracker and pass it in.

agent/grpc-external/services/peerstream/stream_test.go
```diff
@@ -1668,6 +1699,7 @@ func TestHandleUpdateService(t *testing.T) {
 					},
 				},
 			},
+			expectedImportedServicesCount: 0, // IRL, this count would be 1, but we pre-seed the redis service
```
mkeeler, line 1702: In the future, one way you could have worked around this would be to seed the imported service count too.

agent/rpc/peering/service.go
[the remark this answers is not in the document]
mkeeler: Sure.

agent/rpc/peering/service.go
```diff
@@ -586,17 +605,19 @@ func (s *Server) getExistingOrCreateNewPeerID(peerName, partition string) (strin
 
 func copyPeeringWithNewState(p *pbpeering.Peering, state pbpeering.PeeringState) *pbpeering.Peering {
 	return &pbpeering.Peering{
```
mkeeler, line 607: I just realized this whole function could be:

```
copy := *p
copy.State = state
return &copy
```

That way if we add fields we dont keep needing to  add more manual copying of fields here

agent/grpc/public/services/peerstream/stream_tracker.go
mkeeler: I cannot think of what else we would need to track for imported services.

For exported services, we likely need a map because we don't have the ability to check against memdb to see what services have been exported.

I guess a map is okay, but if we stick with it for imports then there is no reason to make the modifications to `handleUpdateService`. Instead you could add a `TrackImportedService` call to `handleUpsert` within the block handling service updates and a similar call to `RemoveTrackedService` within `handleDelete`. Both of those have the service name and with the map we wont double count services by tracking them before we know whether they actually are for a net-new service.

mkeeler (changes requested):
I left a number of comments but half were more just to provide information on alternative ways you could have implemented something or responses to comments of yours in the code. All of these should be marked with `future` or `future/learning` at the beginning of the comment. Read them and then resolve them as no action is needed for this PR.

There are a few that I would consider blockers. Those do not have any markers within the comment.

agent/grpc-external/services/peerstream/stream_resources.go
acpana: oh, I missed that 🤦🏼

agent/grpc-external/services/peerstream/stream_test.go
acpana: I had a version of this where there was a seeding of imported services count too so I'm brining that back!

agent/rpc/peering/service.go
acpana: ah actually this seems to make `go vet` understandably unhappy. Gotta love go vet ❤️ 

https://app.circleci.com/pipelines/github/hashicorp/consul/35170/workflows/565b2994-d7ad-4da8-8825-baee49215c59/jobs/767931

PR #13721, opened 2022-07-12
grpc: rename public/private directories to external/internal

boxofrad:
### Description
This is the OSS portion of enterprise PR 2263.

Previously, public referred to gRPC services that are both exposed on the dedicated gRPC port and have their definitions in the `proto-public` directory (so were considered usable by 3rd parties). Whereas private referred to services on the multiplexed server port that are only usable by agents and other servers (and had their definitions in `proto`).

Now, we're splitting these definitions, such that external/internal refers to the port and public/private refers to whether they can be used by 3rd parties.

This is necessary because the peering replication API needs to be exposed on the dedicated/external port, but is not (yet) suitable for use by 3rd parties.

Also updated the docs (see: [docs/rpc/README.md](https://github.com/hashicorp/consul/blob/boxofrad/move-grpc-directories/docs/rpc/README.md)).


rboyer (approved):
LGTM pending a rebase.

PR #13724, opened 2022-07-12
ui: peer permission handling

LevelbossMike:
### Description
This PR adds peering permission handling for the peer resources.

* it adds the peering permission to the requested permissions in the `dc`-route when the peering feature is enabled
* it updates the peer - ability to use the `peering`-resource to determine permissions not `operator`
* it updates the `canDelete`-permission for a peer to also take the peering:write permission into account not only the peering's State.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

cc @johncowen


johncowen (approved):
LGTM! ty!

PR #13732, opened 2022-07-12
peerstream: some cosmetic refactors to make this easier to follow

rboyer:
### Description
- Use some protobuf construction helper methods for brevity.
- Rename a local variable to avoid later shadowing.
- Rename the Nonce field to be more like xDS's naming.
- Be more explicit about which PeerID fields are empty.


agent/grpc/public/services/peerstream/stream_resources.go
```diff
@@ -143,18 +157,21 @@ func (s *Server) DrainStream(req HandleStreamRequest) {
 
 // The localID provided is the locally-generated identifier for the peering.
 // The remoteID is an identifier that the remote peer recognizes for the peering.
-func (s *Server) HandleStream(req HandleStreamRequest) error {
+func (s *Server) HandleStream(streamReq HandleStreamRequest) error {
```
rboyer, line 160: This variable is shadowed below when we fetch `GetRequest()`

PR #13733, opened 2022-07-12
server: ensure peer replication can successfully use TLS over external gRPC

rboyer:
### Description

Ensure that the peer stream replication rpc can successfully be used with TLS activated.

Also:
- If key material is configured for the gRPC port but HTTPS is not enabled now TLS will still be activated for the gRPC port.
- peerstream replication stream opened by the establishing-side will now ignore `grpc.WithBlock` so that TLS errors will bubble up instead of being awkwardly delayed or suppressed

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -761,15 +761,7 @@ func (a *Agent) Failed() <-chan struct{} {
 }
 
 func (a *Agent) buildExternalGRPCServer() {
-	// TLS is only enabled on the gRPC server if there's an HTTPS port configured.
-	var tls *tlsutil.Configurator
-	if a.config.HTTPSPort > 0 {
-		tls = a.tlsConfigurator
-	}
-	if a.tlsConfigurator != nil && a.tlsConfigurator.GRPCTLSConfigured() {
-		tls = a.tlsConfigurator
-	}
-	a.externalGRPCServer = external.NewServer(a.logger.Named("grpc.external"), tls)
+	a.externalGRPCServer = external.NewServer(a.logger.Named("grpc.external"), a.tlsConfigurator)
 }
 
 func (a *Agent) listenAndServeGRPC() error {
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -766,6 +766,9 @@ func (a *Agent) buildExternalGRPCServer() {
 	if a.config.HTTPSPort > 0 {
 		tls = a.tlsConfigurator
 	}
+	if a.tlsConfigurator != nil && a.tlsConfigurator.GRPCTLSConfigured() {
+		tls = a.tlsConfigurator
+	}
 	a.externalGRPCServer = external.NewServer(a.logger.Named("grpc.external"), tls)
 }
 
--- a/agent/consul/leader_peering.go
+++ b/agent/consul/leader_peering.go
@@ -6,6 +6,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"time"
 
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-memdb"
@@ -14,6 +15,7 @@ import (
 	"golang.org/x/time/rate"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/keepalive"
 
 	"github.com/hashicorp/consul/acl"
 	"github.com/hashicorp/consul/agent/consul/state"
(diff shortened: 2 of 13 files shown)


tlsutil/config.go
```diff
@@ -88,7 +88,7 @@ type ProtocolConfig struct {
 	// certificate authority. This is used to verify authenticity of server
 	// nodes.
 	//
-	// Note: this setting doesn't apply to the gRPC configuration, as Consul
+	// Note: this setting doesn't apply to the external gRPC configuration, as Consul
 	// makes no outgoing connections using this protocol.
```
boxofrad, line 92: This assertion isn't true anymore, right? Now that Consul is making outgoing gRPC connections for peering replication?

agent/agent.go
```diff
@@ -766,6 +766,9 @@ func (a *Agent) buildExternalGRPCServer() {
 	if a.config.HTTPSPort > 0 {
 		tls = a.tlsConfigurator
 	}
+	if a.tlsConfigurator != nil && a.tlsConfigurator.GRPCTLSConfigured() {
+		tls = a.tlsConfigurator
+	}
```
boxofrad, line 771: ⚠️ This could make upgrading client agents difficult because in many cases it'll (unexpectedly?) enable TLS on the gRPC port that Envoy is using for xDS.

This is because `GRPCTLSConfigured` is true in _any_ of the following cases:

- `tls.grpc.{cert_file,key_file}` are set
- `tls.defaults.{cert_file,key_file}` are set
- Auto-config or auto-encrypt are in use

Which is why we kept the https port/grpc coupling when introducing the new `tls` stanza, even though it doesn't make much sense now that TLS is configured separately.

If we're sure we want to go ahead with this change, we should probably just remove the above check (`if a.config.HTTPSPort > 0`) and always pass `tls`, because we already check `GRPCTLSConfigured` in `external.NewServer`.

boxofrad (changes requested):

agent/agent.go
mkeeler: FWIW, I would be open to removing configuring TLS being dependent on HTTPs enablement.

In 1.12 we had to do differently to allow for an upgrade path to work at all.

For 1.13 you can just go through 1.12, fix up your settings and then move along to 1.13.

So for me, I saw we rip out the HTTPSPort check, always pass the tls configurator and add to the upgrade_specific.mdx with the details about the breaking change.

tlsutil/config.go
mkeeler: We should probably get rid of this check: 

https://github.com/hashicorp/consul/blob/68e79b8180ca89e8cfca291b40a30d943039bd49/agent/config/builder.go#L2522-L2524

And the corresponding tests.

mkeeler (commented):
I don't have much to add beyond what Dan pointed out.

At a high level I think we want to 1:

1. Have a breaking change for decoupling https and grpc tls enablement
2. Enable the `tls.grpc.verify_outgoing` setting to be set (and to default its value from `tls.defaults.verify_outgoing`

agent/agent.go
boxofrad: That's a great point about there being an upgrade path through 1.12 now. I'm in favor of dropping it too!

tlsutil/config.go
rboyer: This is because the peer conveys dialing instructions via the peering token, rather than us getting any of them from the config file.

agent/consul/leader_peering_test.go
```diff
@@ -20,15 +21,28 @@ import (
...
 	_, s1 := testServerWithConfig(t, func(c *Config) {
-		c.NodeName = "s1.dc1"
+		c.NodeName = "bob"
 		c.Datacenter = "dc1"
 		c.TLSConfig.Domain = "consul"
+		if enableTLS {
+			c.TLSConfig.GRPC.CAFile = "../../test/hostname/CertAuth.crt"
+			c.TLSConfig.GRPC.CertFile = "../../test/hostname/Bob.crt"
+			c.TLSConfig.GRPC.KeyFile = "../../test/hostname/Bob.key"
+			c.TLSConfig.GRPC.VerifyOutgoing = true
```
mkeeler, line 44: The user is currently unable to set this to true.

https://github.com/hashicorp/consul/blob/adb5ffa1a6dd025822d264b75c203a4e3674fcf0/agent/config/builder.go#L2522-L2524

I think we probably ought to remove that check (and update tests accordingly)

mkeeler (approved):

agent/consul/leader_peering.go
```diff
@@ -238,8 +245,15 @@ func (s *Server) establishStream(ctx context.Context, logger hclog.Logger, peer
...
-			grpc.WithBlock(),
+			// TODO(peering): use a grpc.WithStatsHandler here?)
 			tlsOption,
+			// For keep alive parameters there is a larger comment in ClientConnPool.dial about that.
+			grpc.WithKeepaliveParams(keepalive.ClientParameters{
+				Time:    30 * time.Second,
+				Timeout: 10 * time.Second,
+				// send keepalive pings even if there is no active streams
+				PermitWithoutStream: true,
+			}),
```
lkysow, line 256: @rboyer do you need to set the enforcement policy on the external gRPC server (https://github.com/hashicorp/consul/blob/main/agent/grpc-external/server.go#L15) similar to how it is set on the internal server (https://github.com/hashicorp/consul/blob/main/agent/grpc-internal/handler.go#L37-L39)?

lkysow: I've confirmed that we'll need to set this. Otherwise we get an error after a while and it closes the connection:

```
INFO[0108] [transport] transport: Got too many pings from the client, closing the connection.  system=system
```

lkysow: Fixed by https://github.com/hashicorp/consul/pull/13796

PR #13745, opened 2022-07-13
peerstream: dialer should reconnect when stream closes 

lkysow:
### Description
If the stream is closed unexpectedly (i.e. when we haven't received
a terminated message), the dialer should attempt to re-establish the
stream.

Previously, the `HandleStream` would return `nil` when the stream
was closed. The caller then assumed the stream was terminated on purpose
and so didn't reconnect when instead it was stopped unexpectedly and
the dialer should have attempted to reconnect.

Start two servers, peer them, stop the accepting peer server and then restart it, see that the dialing server never reestablishes its connection.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/consul/leader_peering_test.go
+++ b/agent/consul/leader_peering_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/json"
+	"github.com/hashicorp/consul/sdk/freeport"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


agent/consul/leader_peering_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/json"
+	"github.com/hashicorp/consul/sdk/freeport"
```
acpana, line 7: nit: import ordering

agent/consul/leader_peering_test.go
```diff
@@ -282,6 +283,120 @@ func TestLeader_Peering_DeferredDeletion(t *testing.T) {
...
+	// Re-insert the peering state.
+	require.NoError(t, acceptingServerRestart.fsm.State().PeeringWrite(2000, &pbpeering.Peering{
+		ID:    dialingServerPeerID,
+		Name:  "my-peer-dialing-server",
+		State: pbpeering.PeeringState_PENDING,
+	}))
+
+	// The dialing peer should eventually reconnect.
+	retry.Run(t, func(r *retry.R) {
+		connStreams := acceptingServerRestart.peeringServer.Tracker.ConnectedStreams()
```
acpana, line 395: nit:

```suggestion
		connStreams := acceptingServerRestart.peeringServer.ConnectedStreams()
```

agent/consul/leader_peering_test.go
```diff
@@ -282,6 +283,120 @@ func TestLeader_Peering_DeferredDeletion(t *testing.T) {
...
+		require.Len(r, connStreams, 1)
```
acpana, line 396: nit: this is fine but could we actually look by the peer id of `dialingServerPeerID` and assert that it's in the map.

acpana (commented):
this LGTM! 🚢 

Just a couple of non blocking nits

kisunji (approved):
LGTM!

agent/consul/leader_peering_test.go
lkysow: Sounds good. Should we add a linter for this?

agent/consul/leader_peering_test.go
lkysow: I'm not seeing that method on `peeringServer`

rboyer: `peerStreamServer` has a `ConnectedStreams` method that is a pass through.

lkysow: Okay thanks. And just for my knowledge why is that preferable vs reaching through `peeringServer`?

PR #13750, opened 2022-07-14
Latest submodule versions

eculver:
### Description
We cut `api/v1.13.1` and `sdk/v0.10.0` for today's release. This updates our `go.mod` with those so that it stays mostly up-to-date for when we cut the next release branch.

CI's `build-distros` check shouldn't complain


DanStough (commented):
At some point, these packages need to make it into `go.sum`. Do we do that now (and CI won't pass, but consumers can pull in the root @`ref/main`), or do we make this part of our release branch process?

DanStough:
I guess I'm confused then. I thought we always kept an entry in `go.sum` for release branches so consumers of the root module didn't need their own `replace` directive.  It makes sense from a CI perspective to keep it out, but it's not obvious when we first introduce it.

We [added it here today on 1.12.3](https://github.com/hashicorp/consul/blob/release/1.12.3/go.sum#L302-L303), but I don't see it in any previous release branch. 

I also it was intentionally removed in [this commit],(https://github.com/hashicorp/consul/commit/a10fc50aa936d858a78cc87c9edffc59bd6815b8)

Maybe @rboyer has some perspective on how this was managed before CRT.

rboyer (approved):

rboyer (changes requested):
api/go.mod's sdk reference is stale

DanStough:
>My understanding of it is that we require users to use the replace directives, as there's hardly a use-case for building with tagged (and potentially stale) submodules.

If that's the case then the release docs need to be updated so we don't need any of the magic with commenting in/out the `replace` directives.

eculver:
> 

Ah, good call. I missed that one.

kisunji (approved):

kisunji:
Looks like we need another go mod tidy?

PR #13753, opened 2022-07-14
ui: wan federation message dc-dropdown

LevelbossMike:
### Description
This PR adds a change to how the datacenter dropdown works:

1. when only one datacenter is available we won't show a dropdown - we will just show the name of the datacenter
Screenshot 2022-07-14 at 09 51 01

2. when multiple datacenters are available we will show a dropdown + a message indicating why multiple datacenters are displayed.
Screenshot 2022-07-14 at 09 48 38

cc @johncowen - implemented as discussed 

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


ui/packages/consul-ui/app/components/consul/datacenter/selector/index.hbs
```diff
@@ -17,6 +17,9 @@
         @onchange={{action (mut @dcs) value="data"}}
       />
         <panel.Menu as |menu|>
+          <p class="dcs-message">
+            Datacenters shown in this dropdown are available through WAN Federation.
+          </p>
```
johncowen, line 22: I think this adds a `p` tag into the `ul`. If you move it one level up out of the menu though, its parent will be a `div` instead, which is allowed.

ui/packages/consul-ui/app/components/hashicorp-consul/index.hbs
```diff
@@ -83,16 +83,23 @@
     <a
       href={{href-to 'index' params=(hash peer=undefined)}}
     ><Consul::Logo /></a>
+    {{#if (eq @dcs.length 1)}}
+      <div class="dc-name" data-test-datacenter>
+        {{@dc.Name}}
+      </div>
+    {{/if}}
```
johncowen, line 90: What would you think of moving this logic in the `Datacenter::Selector` component? I know we aren't giving the user the ability to 'select' a datacenter as they only have one, but the logic essentially 'selects' that single datacenter for the user as they only have one, which makes sense to me.

Also if we ever feature flag out datacenters entirely (not sure if this will ever happen, but it could), then this entire `Datacenter::Selector` component can use out wiring/injection approach to feature flagging, as everything to do with showing/selecting your selected datacenter is all self-contained in the one component we can just inject in/out different implementations.

This is also pretty much how the Partition::Selector works:

https://github.com/hashicorp/consul/blob/610038ce670213055da323ecfc4f9cf06f4f430b/ui/packages/consul-partitions/app/components/consul/partition/selector/index.hbs#L71-L77

ui/packages/consul-ui/app/components/hashicorp-consul/index.scss
```diff
@@ -2,6 +2,12 @@
   nav .dcs {
     @extend %main-nav-vertical-hoisted;
     left: 100px;
+
+    .dcs-message {
+      padding: 8px 12px;
+      border-bottom: 1px solid rgb(var(--tone-gray-400));
+      max-width: fit-content;
+    }
```
johncowen, line 10: Self-note: We used to do this elsewhere, so we might have something that already exists for this, I'm gonna check in case. This one is not a blocker at all, just wanted to check in a bit.

ui/packages/consul-ui/tests/integration/components/hashicorp-consul-test.js
```diff
@@ -21,4 +22,70 @@ module('Integration | Component | hashicorp consul', function(hooks) {
...
+
+      await render(hbs`<HashicorpConsul @dcs={{this.dcs}} @dc={{this.dc}} />`);
+
+      assert
+        .dom('[data-test-datacenter]')
+        .doesNotExist('we are displaying more than just the name of the first dc');
+
+      assert.dom('[data-test-datacenter-menu]').exists('datacenter dropdown is displayed');
+    });
+  });
```
johncowen, line 90: Am I right in thinking these tests are testing `if arr.length > 1`?

Wondered if there was anything else we are testing here?

I think if we want to test this, we should move these tests to the datacenter::selector (if we implement the suggestions from above)

johncowen (commented):
I think this looks great from a user perspective 👏 .

There is a HTML 'kinda error' here though, and I had a suggestion for implementing this slightly differently following how we do this elsewhere. Let me know what you think!

ui/packages/consul-ui/tests/integration/components/hashicorp-consul-test.js
johncowen: Ah ok, so my main point here is what is the logic we are testing, it looks like we are testing `arr.length > 1`?

FYI: Our testing approach closely aligns with BoundaryUIs:

> We strive for "the right amount of testing". Use test coverage as a guide to help you identify untested high-value code.

This isn't a huge concern, but just wanted to double check the logic we are testing here and I'm wondering whether this fits with the "right amount of testing" idea, what are your thoughts on that?

I'm actually more interested in why we need to patch the router in a component test? Is it because of href-to?

LevelbossMike: This is a test that makes sure that when we have one datacenter we display the name only when we have multiple we display the menu - which is the change that I'm making in this PR. I'm not adding a test for the message because that's just the template layer of Ember that I would be testing.

The patch was necessary because the helpers and how routing is setup breaks when testing the `HashicorpConsul`-component in isolation. With the latest addition I moved to a component test of the selector component itself because now everything is contained in the selection component. No more router patching necessary.

johncowen: > I'm not adding a test for the message because that's just the template layer of Ember that I would be testing.

Awesome, yes that sounds great!

> The patch was necessary because the helpers

Gotcha, thats what I was wondering, if we needed to patch here, I would have preferred to patch the helper. 

Glad now we've moved this around a little we no longer need to patch the router, thanks for doing that!

Yet to look at this altogether now, but this sounds much better 👏

johncowen (approved):
Ok I think this is good to merge as is. Thanks for shifting this around a little more.

I did notice something that would be good to fix up either here or in a follow up PR if you wanted to do that (otherwise I can also tackle it at some point)

I noticed the passing on the top of the DATACENTERS title is rather large, inspecting that shows me this:

Screenshot 2022-07-14 at 16 06 50

And then clicking the layout.scss file shows me this:

Screenshot 2022-07-14 at 16 07 01

I think that rule on line 39 could do with being `%main-nav-vertical > ul > [role='separator']` to only target the immediate separators (the ones that aren't inside dropdown menus). There's possibly more that we could do there from a code org perspective, but completely up to you how you want to play that, just this little amend here would be fine for the moment I think.

johncowen:
Oh 🤦 forgot I'm merging here also 😬 @LevelbossMike did you want to do that little padding/selector tweak mentioned above first? Let me know otherwise I can just go ahead and merge

johncowen:
Thanks for adding that amend! Looks good, gonna merge!

PR #13758, opened 2022-07-14
Make envoy resources for inferred peered upstreams

kisunji:
### Description
Peered upstreams has a separate loop in xds from discovery chain upstreams. This PR adds similar but slightly modified code to add filters for peered upstream listeners, clusters, and endpoints in the case of transparent proxy.

* Added golden test

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/xds/clusters.go
+++ b/agent/xds/clusters.go
@@ -122,7 +122,7 @@ func (s *ResourceGenerator) clustersFromSnapshotConnectProxy(cfgSnap *proxycfg.C
 	// NOTE: Any time we skip an upstream below we MUST also skip that same
 	// upstream in endpoints.go so that the sets of endpoints generated matches
 	// the sets of clusters.
-	for _, uid := range cfgSnap.ConnectProxy.PeeredUpstreamIDs() {
+	for _, uid := range cfgSnap.ConnectProxy.PeeredUpstreamIDsFromConfig() {
 		upstreamCfg := cfgSnap.ConnectProxy.UpstreamConfig[uid]
 
 		explicit := upstreamCfg.HasLocalPortOrSocket()
--- a/agent/xds/clusters.go
+++ b/agent/xds/clusters.go
@@ -122,7 +122,7 @@ func (s *ResourceGenerator) clustersFromSnapshotConnectProxy(cfgSnap *proxycfg.C
 	// NOTE: Any time we skip an upstream below we MUST also skip that same
 	// upstream in endpoints.go so that the sets of endpoints generated matches
 	// the sets of clusters.
-	for _, uid := range cfgSnap.ConnectProxy.PeeredUpstreamIDsFromConfig() {
+	for _, uid := range cfgSnap.ConnectProxy.PeeredUpstreamIDs() {
 		upstreamCfg := cfgSnap.ConnectProxy.UpstreamConfig[uid]
 
 		explicit := upstreamCfg.HasLocalPortOrSocket()
@@ -134,7 +134,7 @@ func (s *ResourceGenerator) clustersFromSnapshotConnectProxy(cfgSnap *proxycfg.C
 
 		peerMeta := cfgSnap.ConnectProxy.UpstreamPeerMeta(uid)
 
-		upstreamCluster, err := s.makeUpstreamClusterForPeerService(upstreamCfg, peerMeta, cfgSnap)
+		upstreamCluster, err := s.makeUpstreamClusterForPeerService(uid, upstreamCfg, peerMeta, cfgSnap)
 		if err != nil {
 			return nil, err
 		}
@@ -693,6 +693,7 @@ func (s *ResourceGenerator) makeAppCluster(cfgSnap *proxycfg.ConfigSnapshot, nam
 }
 
 func (s *ResourceGenerator) makeUpstreamClusterForPeerService(
+	uid proxycfg.UpstreamID,
 	upstream *structs.Upstream,
 	peerMeta structs.PeeringServiceMeta,
 	cfgSnap *proxycfg.ConfigSnapshot,
@@ -701,8 +702,6 @@ func (s *ResourceGenerator) makeUpstreamClusterForPeerService(
 		c   *envoy_cluster_v3.Cluster
 		err error
 	)
-
-	uid := proxycfg.NewUpstreamID(upstream)
 
 	cfg := s.getAndModifyUpstreamConfigForPeeredListener(uid, upstream, peerMeta)
 	if cfg.EnvoyClusterJSON != "" {
--- a/agent/xds/listeners.go
+++ b/agent/xds/listeners.go
@@ -264,8 +264,9 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
 		}
 		outboundListener.ListenerFilters = append(outboundListener.ListenerFilters, tlsInspector)
 	}
-	// Looping over explicit upstreams is only needed for cross-peer because
-	// they do not have discovery chains.
+
+	// Looping over explicit and implicit upstreams is only needed for cross-peer
+	// because they do not have discovery chains.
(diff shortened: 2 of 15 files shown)


agent/xds/clusters.go
```diff
@@ -130,7 +130,7 @@ func (s *ResourceGenerator) clustersFromSnapshotConnectProxy(cfgSnap *proxycfg.C
 	// NOTE: Any time we skip an upstream below we MUST also skip that same
 	// upstream in endpoints.go so that the sets of endpoints generated matches
 	// the sets of clusters.
-	for _, uid := range cfgSnap.ConnectProxy.PeeredUpstreamIDs() {
+	for _, uid := range cfgSnap.ConnectProxy.PeeredUpstreamIDsFromConfig() {
```
kisunji, line 133: Wasn't sure whether I should: 1) split this function (current decision); or 2) make `PeeredUpstreamIDs` return implicit+explicit upstreamIDs and do a `continue` in the loop if we detect that `upstreamCfg == nil`

agent/xds/listeners.go
```diff
@@ -289,7 +289,50 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
 		// Below we create a filter chain per upstream, rather than a listener per upstream
 		// as we do for explicit upstreams above.
 
-		// TODO(peering): tproxy
+		filterChain, err := s.makeUpstreamFilterChain(filterChainOpts{
+			routeName:   uid.EnvoyID(),
+			clusterName: clusterName,
+			filterName:  uid.EnvoyID(),
```
kisunji, line 295: this diverges from disco chain filter names but is the same name format we give explicit peered upstreams. Thoughts?
```go
filterName := fmt.Sprintf("%s.%s.%s.%s", chain.ServiceName, chain.Namespace, chain.Partition, chain.Datacenter)
```

mkeeler: Seems fine to me and I think having the query param style syntax that EnvoyID uses will make it easier to understand/debug if you are looking at a config dump.

mkeeler: Should we update the disco chain filter names (in the future) to use this syntax?

agent/xds/testdata/clusters/transparent-proxy-with-peered-upstreams.latest.golden
```diff
@@ -0,0 +1,97 @@
+{
```
mkeeler, line 1: It looks like something may be missing here.

The listeners/transparent-proxy-with-peered-upstreams golden file reference a cluster with name `db.default.default.cloud.external.1c053652-8512-4373-90cf-5a7f6263a994.consul`. I would expect that same cluster to show up in these resources but I don't believe its in here.

agent/xds/testdata/endpoints/transparent-proxy-with-peered-upstreams.latest.golden
```diff
@@ -0,0 +1,29 @@
+{
```
mkeeler, line 1: Same as the previous comment. I would expect the db upstreams endpoints to be in here.

mkeeler (commented):
Correct me if I am totally wrong but I left a few comments in the golden files for clusters and endpoints about what seems like missing configuration regarding the implicit db upstream.

agent/xds/listeners.go
```diff
@@ -264,8 +264,9 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
 		}
 		outboundListener.ListenerFilters = append(outboundListener.ListenerFilters, tlsInspector)
 	}
-	// Looping over explicit upstreams is only needed for cross-peer because
-	// they do not have discovery chains.
+
+	// Looping over explicit and implicit upstreams is only needed for cross-peer
+	// because they do not have discovery chains.
 	for _, uid := range cfgSnap.ConnectProxy.PeeredUpstreamIDs() {
```
rboyer, line 270: Why is this not using `PeeredUpstreamIDsFromConfig` like the clusters and endpoints loops?

kisunji: `UpstreamConfig` doesn't contain the implicit peered upstreams. So `PeeredUpstreamIDs` is explicit + implicit upstreams (which mirrors disco chains in that they contain both implicit and explicit upstreams)

agent/xds/testdata/clusters/transparent-proxy-with-peered-upstreams.latest.golden
rboyer: Any Listener or Route that points to a cluster must ultimately have that cluster emitted by CDS. Similarly if a cluster is configured and expects EDS, then the EDS endpoints should be emitted for that cluster else xDS effectively deadlocks waiting for a condition that will never resolve.

agent/xds/clusters.go
```diff
@@ -702,8 +703,6 @@ func (s *ResourceGenerator) makeUpstreamClusterForPeerService(
 		err error
 	)
 
-	uid := proxycfg.NewUpstreamID(upstream)
-
```
kisunji, line 706: upstream (from `UpstreamConfig`) could be nil here

agent/xds/listeners.go
kisunji: All resources use `PeeredUpstreamIDs` and the `FromConfig` variant has been removed

agent/xds/listeners.go
```diff
@@ -326,7 +327,50 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
...
+		// all instances when any instance address is dialed.
+		for _, e := range endpoints {
+			if vip := e.Service.TaggedAddresses[structs.TaggedAddressVirtualIP]; vip.Address != "" {
+				uniqueAddrs[vip.Address] = struct{}{}
+			}
+
+			// The virtualIPTag is used by consul-k8s to store the ClusterIP for a service.
+			// For services imported from a peer,the partition will be equal in all cases.
+			if acl.EqualPartitions(e.Node.PartitionOrDefault(), cfgSnap.ProxyID.PartitionOrDefault()) {
+				if vip := e.Service.TaggedAddresses[virtualIPTag]; vip.Address != "" {
```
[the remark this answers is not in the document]
kisunji: Maybe @freddygv will have some context? I poked around at the history and it looks like the `virtualIPTag` was for initial K8s support of TProxy before Consul was later updated to use `structs.TaggedAddressVirtualIP` and assign VIPs directly. Some tests assign different IPs to the two keys but I'm not sure if that's valid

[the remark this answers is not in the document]
freddygv: * `virtual` is for the Kubernetes-assigned ClusterIP for services. As Chris mentioned this is the address that's expected to be used for cluster-local communication, since it holds the address returned by kubedns queries.
* `consul-virtual` is for the one assigned by Consul itself, it was introduced for cross-partition communication in 1.11, and is also used for addressing peered services. This requires configuring/using Consul DNS.
* `consul-virtual:<service-name>` is a variant attached to terminating gateways. Since a gateway endpoint can route to multiple services, there is a virtual IP attached to every gateway instance for each of those linked service names.

agent/xds/testdata/clusters/transparent-proxy-with-peered-upstreams.latest.golden
mkeeler: Should we be emitting the no-endpoints listener, cluster and EDS endpoint info (with an empty set of endpoints). 

Per RBs comment above I _think_ that we should suppress emission of that somehow.

mkeeler: If it works fine for tproxy already then I guess its fine. Maybe its only a deadlock if we never send EDS resources for the cluster which is subtly different from sending EDS resource with 0 endpoints defined.

mkeeler (approved):
LGTM

PR #13760, opened 2022-07-14
Extract AWS auth implementation out of Consul

pglass:
### Description
This extracts the AWS IAM auth method implementation from its internal package into a separate repository, [hashicorp/consul-awsauth](https://github.com/hashicorp/consul-awsauth), so that it can be used by other Hashicorp tools.

In particular, this exposes the [`GenerateLoginData`](https://github.com/hashicorp/consul-awsauth/blob/83146e702996bb3a304eb5b9a3368b96c1802516/util.go#L39) function in the separate repo to generate the bearer token for Consul's Login API request, so that we can login without needing to use the `consul login` cli.

The existing tests are enough.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


erichaberkorn:
Since this is only a refactoring, it should be safe to fix the "Changelog Checker" check with the 'pr/no-changlelog' tag.

PR #13768, opened 2022-07-14
peering: add warning about AllowStaleRead

acpana:
### Description

Freddy was mentioning that we need to add a warning about the constraint of having the leader field read request for peering calls.

This adds a TODO to add that context.

### DIFF
--- a/internal/tools/proto-gen-rpc-glue/main.go
+++ b/internal/tools/proto-gen-rpc-glue/main.go
@@ -108,6 +108,9 @@ func processFile(path string) error {
 			if ann.ReadTODO != "" {
 				log.Printf("    ReadTODO from %s", ann.ReadTODO)
 			}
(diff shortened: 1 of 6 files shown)


DanStough (commented):
🚀

kisunji (approved):

internal/tools/proto-gen-rpc-glue/main.go
```diff
@@ -108,6 +108,9 @@ func processFile(path string) error {
 			if ann.ReadTODO != "" {
 				log.Printf("    ReadTODO from %s", ann.ReadTODO)
 			}
+			if ann.LeaderReadTODO != "" {
+				log.Printf("    LeaderReadTODO from %s", ann.ReadTODO)
```
rboyer, line 112: This line should have used `LeaderReadTODO`

PR #13770, opened 2022-07-14
Add Consul Lambda integration tests

cthain:
### Description
This PR adds tests for the serverless plugin integration with AWS Lambda. The serverless plugin has been exercised to an extent with manual testing and through the Lambda registrator acceptance tests. These approaches don't scale and won't be effective in catching potential regressions in the envoy configs.

* This PR adds the tests for the service mesh -> Lambda use case. The tests are in the `test/integration/connect/envoy/case-mesh-to-lambda` directory.
* These tests require AWS credentials with permission to invoke Lambda functions. The creds are configured via env vars. If the expected variables do not exist in the environment then the tests will be skipped so as not to break this for everyone else.

### PR Checklist

* [ ] updated test coverage
* [x] ~external facing docs updated~ N/a
* [x] not a security concern


test/integration/connect/envoy/case-mesh-to-lambda/config_entries.hcl
```diff
@@ -0,0 +1,34 @@
...
+      }
+    ]
+  }
+  bootstrap {
+    kind = "service-defaults",
+    name = "l1",
+    protocol = "http"
+    meta {
+      "serverless.consul.hashicorp.com/v1alpha1/lambda/enabled" = "true"
+      "serverless.consul.hashicorp.com/v1alpha1/lambda/arn" = "arn:aws:lambda:us-west-2:977604411308:function:consul-envoy-integration-test"
```
cthain, line 18: We have some options on how we deploy a Lambda function for the integration tests to interact with:

1. Create a long-lived, well-known Lambda function outside of CI. Then reference this function in the tests. This approach is simple and will work as long as no one deletes the Lambda function.
1. Dynamically create the function when the tests run as part of the setup. This could be done via the `aws` CLI or Terraform. This is potentially more flexible and would keep the Lambda function code colocated with the test; however, it introduces more dependencies and complexity within the test setup.

For now I've implemented option 1. I plan to stick with this approach unless there are objections.

test/integration/connect/envoy/case-mesh-to-lambda/service_s1.hcl
```diff
@@ -0,0 +1,20 @@
...
+          },
+          {
+            destination_name = "l2"
+            local_bind_port = 5678
+          }
+        ]
+      }
+    }
+  }
+}
```
cthain, line 20:
```suggestion
}

```

test/integration/connect/envoy/case-mesh-to-lambda/setup.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -eEuo pipefail
+
+# Copy lambda config files into the register dir
+# TODO: not sure if we want to make this step more generic and add it to the init_workdir func?
+find ${CASE_DIR} -maxdepth 1 -name 'lambda_*.json' -type f -exec cp -f {} workdir/${CLUSTER}/register \;
```
cthain, line 7: I put this in the setup for the `mesh-to-lambda` case because using JSON and the Consul HTTP API is currently specific to the registration of the lambda functions. The registration of the lambdas requires the use of the Consul HTTP API because the CLI doesn't support registering the lambdas "services" to the catalog.

I didn't think through this very far but we could probably massage this to be more generic if we think it will be useful for other cases in the future.

test/integration/connect/envoy/case-mesh-to-lambda/verify.bats
```diff
@@ -0,0 +1,39 @@
...
+@test "s1 can call l2 through the terminating gateway" {
+  run retry_default curl -s -f -H "Content-type: application/json" -d '"hello"' 'localhost:5678'
+  [ "$status" -eq 0 ]
+  [ "$output" == '"hello"' ]
+
+   # Omitting the Content-type in the request will cause envoy to base64 encode the request.
+  run curl -s -f -d '{"message":"hello"}' 'localhost:5678'
+  [ "$status" -eq 0 ]
+  [ "$output" == '{"message":"hello"}' ]
+}
```
cthain, line 39: Need to add tests for more configurations:
- Asynchronous invocations
- Namespaces
- Admin partitions

test/integration/connect/envoy/helpers.bash
```diff
@@ -966,3 +966,57 @@ function create_peering {
...
+  local REGION=$3
+  local PORT=$4
+
+  local BODY=$(get_envoy_dynamic_cluster_once $HOSTPORT $NAME_PREFIX)
+  [ -n "$BODY" ]
+
+  [ "$(echo $BODY | jq -r '.cluster.type')" == "LOGICAL_DNS" ]
+  [ "$(echo $BODY | jq -r '.cluster.transport_socket.typed_config.sni')" == '*.amazonaws.com' ]
+  [ "$(echo $BODY | jq -r '.cluster.dns_lookup_family')" == "V4_ONLY" ]
+  [ "$(echo $BODY | jq -r '.cluster.metadata.filter_metadata|.["com.amazonaws.lambda"].egress_gateway')" == "true" ]
```
cthain, line 1000: I added more validation for the lambda filters than there seems to be for the other envoy config.. The config fields are important but if this is too verbose I can thin this out.

test/integration/connect/envoy/run-tests.sh
```diff
@@ -43,6 +43,16 @@ function network_snippet {
     echo "--net container:envoy_consul-${DC}_1"
 }
 
+function aws_creds_snippet {
+  local snippet=""
+  if [[ -n "$AWS_ACCESS_KEY_ID" && -n "$AWS_SECRET_ACCESS_KEY" ]]; then
+    snippet="-e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY"
+    # This allows the Lambda integration tests to run when AWS credentials are obtained through STS
+    [[ -n "$AWS_SESSION_TOKEN" ]] && snippet="${snippet} -e AWS_SESSION_TOKEN=$AWS_SESSION_TOKEN"
```
cthain, line 51: The session token is required for our local testing because our AWS creds are provisioned via STS. I guess I could clarify the comment...

test/integration/connect/envoy/case-mesh-to-lambda/config_entries.hcl
erichaberkorn: IMO 1 is better because it will be significantly faster and less flaky.

erichaberkorn: I think it's worth adding a comment about this to the code

test/integration/connect/envoy/helpers.bash
erichaberkorn: I think we can safely remove most of these cases. At this level of testing, the most important thing is that Lambda is configured at all. We have a bunch of golden tests that capture the details ([snapshot](https://github.com/hashicorp/consul/blob/main/agent/xds/testdata/serverless_plugin/listeners/lambda-terminating-gateway-with-service-resolvers.latest.golden#L153-L160), [test](https://github.com/hashicorp/consul/blob/main/agent/xds/serverless_plugin_oss_test.go)).

test/integration/connect/envoy/case-mesh-to-lambda/config_entries.hcl
cthain: Added to the CircleCI config file.

test/integration/connect/envoy/case-mesh-to-lambda/verify.bats
cthain: I've created a task to add these in a future PR.

test/integration/connect/envoy/helpers.bash
cthain: Done. I kept the check to make sure the function ARN is correct, but removed the rest.

test/integration/connect/envoy/run-tests.sh
cthain: Actually the session token is also required in CI.

test/integration/connect/envoy/case-mesh-to-lambda/service_defaults_l1.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Kind": "service-defaults",
+  "Name": "l1",
+  "Protocol": "http",
+  "Meta": {
+    "serverless.consul.hashicorp.com/v1alpha1/lambda/enabled": "true",
+    "serverless.consul.hashicorp.com/v1alpha1/lambda/region": "${AWS_LAMBDA_REGION}",
+    "serverless.consul.hashicorp.com/v1alpha1/lambda/arn": "${AWS_LAMBDA_ARN}",
```
cthain, line 8: Lambda configuration is injected as evironment variables and a variable expansion is performed here so that we don't expose information about our AWS account in our CI config.

cthain:
@erichaberkorn @pglass I've got the tests running successfully in CI now. This is ready for review. I'll request a review from core once I have addressed any comments that you have.

erichaberkorn (approved):
This is awesome. Great work

test/integration/connect/envoy/run-tests.sh
```diff
@@ -43,6 +43,17 @@ function network_snippet {
     echo "--net container:envoy_consul-${DC}_1"
 }
 
+function aws_snippet {
+  local snippet=""
+  [ -n "$(set | grep '^AWS_ACCESS_KEY_ID=')" ] && snippet="${snippet} -e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID"
+  [ -n "$(set | grep '^AWS_SECRET_ACCESS_KEY=')" ] && snippet="${snippet} -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY"
+  [ -n "$(set | grep '^AWS_SESSION_TOKEN=')" ] && snippet="${snippet} -e AWS_SESSION_TOKEN=$AWS_SESSION_TOKEN"
+  [ -n "$(set | grep '^AWS_LAMBDA_REGION=')" ] && snippet="${snippet} -e AWS_LAMBDA_REGION=$AWS_LAMBDA_REGION"
+  [ -n "$(set | grep '^AWS_LAMBDA_ARN=')" ] && snippet="${snippet} -e AWS_LAMBDA_ARN=$AWS_LAMBDA_ARN"
```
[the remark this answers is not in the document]
cthain: Yes, good call :+1:

freddygv (approved):
LGTM, thanks for adding this!

PR #13771, opened 2022-07-15
ui: make nodes searchable by peer-name

LevelbossMike:
### Description
Makes nodes searchable by peer-name.

This PR also fixes the issue that services would show the peer search filter although the peering feature was turned off.

Thea addition of the `PeeredResourceController` is necessary because the custom router dsl and search predicates abstraction doesn't seem to be customizable based on features being on or off.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

cc @johncowen

### DIFF
--- a/ui/packages/consul-ui/app/controllers/_peered-resource.js
+++ b/ui/packages/consul-ui/app/controllers/_peered-resource.js
@@ -1,0 +1,16 @@
+import Controller from '@ember/controller';
+import { inject as service } from '@ember/service';
+
+export default class PeeredResourceController extends Controller {
+  @service abilities;
+
+  get _searchProperties() {
+    const { searchProperties } = this;
+
+    if (!this.abilities.can('use peers')) {
+      return searchProperties.filter(propertyName => propertyName !== 'PeerName');
+    } else {
+      return searchProperties;
+    }
+  }
+}
--- a/ui/packages/consul-ui/app/controllers/dc/nodes/index.js
+++ b/ui/packages/consul-ui/app/controllers/dc/nodes/index.js
@@ -1,0 +1,3 @@
+import PeeredResourceController from 'consul-ui/controllers/_peered-resource';
+
+export default class DcServicesController extends PeeredResourceController {}
--- a/ui/packages/consul-ui/app/controllers/dc/services/index.js
+++ b/ui/packages/consul-ui/app/controllers/dc/services/index.js
@@ -1,0 +1,3 @@
+import PeeredResourceController from 'consul-ui/controllers/_peered-resource';
+
+export default class DcServicesController extends PeeredResourceController {}
--- a/ui/packages/consul-ui/app/search/predicates/node.js
+++ b/ui/packages/consul-ui/app/search/predicates/node.js
@@ -1,5 +1,6 @@
 export default {
   Node: item => item.Node,
   Address: item => item.Address,
+  PeerName: item => item.PeerName,
   Meta: item => Object.entries(item.Meta || {}).reduce((prev, entry) => prev.concat(entry), []),
 };
--- a/ui/packages/consul-ui/app/templates/dc/nodes/index.hbs
+++ b/ui/packages/consul-ui/app/templates/dc/nodes/index.hbs
@@ -44,10 +44,10 @@ as |route|>
       searchproperty=(hash
         value=(if (not-eq searchproperty undefined)
           (split searchproperty ',')
-          searchProperties
+          this._searchProperties
         )
         change=(action (mut searchproperty) value="target.selectedItems")
-        default=searchProperties
+        default=this._searchProperties
       )
     )
 
--- a/ui/packages/consul-ui/app/templates/dc/services/index.hbs
+++ b/ui/packages/consul-ui/app/templates/dc/services/index.hbs
@@ -43,10 +43,10 @@ as |route|>
     searchproperty=(hash
       value=(if (not-eq searchproperty undefined)
         (split searchproperty ',')
-        searchProperties
+        this._searchProperties
       )
       change=(action (mut searchproperty) value="target.selectedItems")
-      default=searchProperties
+      default=this._searchProperties
     )
 
   )
--- a/ui/packages/consul-ui/vendor/consul-ui/routes.js
+++ b/ui/packages/consul-ui/vendor/consul-ui/routes.js
@@ -215,7 +215,7 @@
               status: 'status',
               searchproperty: {
                 as: 'searchproperty',
-                empty: [['Node', 'Address', 'Meta']],
+                empty: [['Node', 'Address', 'Meta', 'PeerName']],
               },
               search: {
                 as: 'filter',


johncowen (commented):
Wondered if you tried amending the `consul-peering` JSON configuration to add `PeerName` to the search properties? That way, if we don't load the `consul-peering` sub-application/plugin/configuration then the search properties will not include `PeerName` meaning you then can't search by that property.

I 'think' there is a slight sharp edge here that I vaguely remember from a while back, but let me know if you tried that ^ and we can discuss further. Need to remember the exact detail myself.

LevelbossMike:
@johncowen adding this to the `consul-peerings`-sub-application's route configuration doesn't work. The resulting routes-configuration looks like this:

```
this.route('index', {"path":"/","queryParams":{"sortBy":"sort","status":"status","source":"source","kind":"kind","searchproperty":{"as":"searchproperty","empty":[["Partition","Name","Tags"],["Partition","Name","Tags","PeerName"]]},"search":{"as":"filter","replace":true}}});
```

i.e. the empty-property for the `searchProperty`-property won't be merged it will be two item array with different search-property arrays.

To me quite frankly this also feels super weird that a sub-app can change non-sup-app routes.

LevelbossMike:
I added some commits to a fork that demonstrate the changes that needed to be made if we wanted to do this via the route config. 

https://github.com/LevelbossMike/consul/pull/1

I suggest to merge this as is and come back to this when we find the time. This would be a huge change for just making the `PeerName` optional that worked around the framework in a very weird way.

The ability to do this via some kind of injection mechanism certainly would be nice but we probably want to think of a proper api how to do this instead of just relying on what [deepMerge.all](https://github.com/LevelbossMike/consul/blob/701a5fcc09d42f367bb3bdcdba3682aad42ea355/ui/packages/consul-ui/app/router.js#L12) does internally.

johncowen:
Hey @LevelbossMike 

I would very much like to avoid controllers. I just finished up a comment on your PR over there https://github.com/LevelbossMike/consul/pull/1#issuecomment-1188786511.

Please let me know your thoughts on that

LevelbossMike:
@johncowen my thought on this as already outlined in the draft-PR:

1. Controllers are part of the framework and there is no fleshed out plan to get rid of them yet - I have seen multiple applications work around them since the old "controllers are going away"-meme started > 8 years ago and never ever has just getting rid of them resulted in easier to understand code.
2. Even the official guides for 4.5 go into detail about the usage of controllers and declare them [a fine feature to use](https://guides.emberjs.com/release/routing/controllers/#toc_should-we-use-controllers-in-my-application-ive-heard-theyre-going-away) for 🥁 query-params _and_ when needing to do some computations based on the `model`-hook of the route. Which is exactly what we are doing here.
3. You are __not__ not using controllers - you are just hiding them if you don't create them explicitly. To support query-params you need to use them and even if you aren't all the existing workarounds like creating a top-level component per route are verbose and unnecessary. And you still need to know about how controllers and routes work together. This is how the framework works today.
4. The alternative of putting this logic into the template layer also doesn't feel right to me. To achieve what we are trying to do we would need to rely on nesting multiple helpers which is fine in some cases but unnecessary in this case. We are already working with a pretty verbose template. Why would we want to add additional logic to it? All search-bars pretty much use a copy-pasted template. Why would we handle that in that particular template and not where the search-property is actually assigned, i.e. at the controller level? - I linked to [a blog post](https://simplabs.com/blog/2017/03/21/on-computed-properties-vs-helpers/) that goes into detail about this.
5. The suggested approach shares the logic of not showing the `PeerName` filter between the `dc.services.index` and `dc.nodes.index`-controller by creating a base-controller. This way we don't need to repeat ourselves. We would need to repeat ourselves if we did that in the template layer. An alternative would be to create a helper that does this but why? It would also be tightly coupled to the way the `Search-Bar`-component currently works.

johncowen:
Hey @LevelbossMike 

Thanks for summarising you viewpoint on this, let me read over what you've said properly and I'll come back to you.

evrowe:
@johncowen @LevelbossMike 

> Yes, agree. They are part of the framework currently, but it seems very clear (to me at least) that a lot of folks would like them not to be part of the framework.

This may be true, however the Ember Core Team have stated on multiple occasions over the past several years that controllers are not going away and encourage their use where appropriate. The fact that not everybody who uses the framework doesn't like them isn't a good justification to completely avoid them. (For the record, I am one of those folks who would prefer they were not in the framework).

I generally prefer to avoid relying on Controllers for most things, _except_ for the specific things they are good at and which are too burdensome to re-implement in other ways. In those cases, writing a few lines of controller code is usually the best option. I don't think there is anything wrong with having the occasional defined Controller in this or any other Ember application, and this seems like the kind of use case where I would typically implement one.

I agree that it would be preferable if: Ember's `Route` class could handle the special/unique cases that we need to rely on the `Controller` class for, and/or Ember's Router supported Routable Components, but unfortunately neither of these things are the case and it's important that we are willing and able to accept that sometimes, we may have to work with things we'd otherwise prefer to avoid.

I'd like to use this as an opportunity to re-evaluate our rationale for not using Controllers, and request that we consider adjusting our approach to "use Controllers sparingly, but when appropriate"; this is the approach that I have found to be the best compromise over the years in terms of not not utilizing them in the majority of scenarios but being able to take advantage of what they do offer when it's helpful.

> As part of the modernization of (not necessarily future proofing) our application we planned to remove mixins and controllers plus some other things, we would like to add the [no-controllers lint rule](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-controllers.md) that is specifically for this (we already have a lint rule for avoiding mixins)

Controllers are, for better or worse, not going away, they are still included in the context of what is considered to be a modern Ember application, and I don't feel that it's hurting us or the application code to include them on occasions where doing so is reasonable.

A final question to consider: _If we were to move forward with the Controller approach in this PR, what would the concrete negative impacts to the Cluster Peering feature and the Consul UI application be?_ Would the feature cease to be usable? Would the application cease to function? Does the use of a Controller present immediate detrimental impact to the usability of the application by the end user? I understand that there is a clear difference in stylistic preferences in the discussion in this PR, but I want to ensure we are thinking critically about whether this approach meaningfully compromises the application in some way.

johncowen:
Hi @evrowe  👋 

Thanks for dropping by! I think you make some great points here, but!..

I think before asking a bit more about your points, there is one super important thing I wanted to point out, and I know we have discussed this offline but want to make this super clear here also:

The main reason for me opening the discussion here was not around "Controllers or not Controllers", more around the fact that the approach in this PR is in contradiction to the various RFCs that have been discussed and planned as a team (design, UI and core). That is around the 'train' towards a 'pluggable' UI (a.k.a sub-applications, a phrase I don't really like to describe this).

Personally my ideal solution here uses the "centralized wiring" feature-flagging approach that we have also spoken about offline. Using this approach is preferred, and is detailed as my initial reason for opening the discussion in the PR over here https://github.com/LevelbossMike/consul/pull/1#issuecomment-1188786511. The Controller issue was a secondary reason and I think we unfortunately gone off on the Controller tangent as the principal reasoning is in a PR elsewhere, sorry about that.

That being said, to dig into and investigate your a points/comments a little more here:

> This may be true, however the Ember Core Team have stated on multiple occasions over the past several years that controllers are not going away and encourage their use where appropriate. The fact that not everybody who uses the framework doesn't like them isn't a good justification to completely avoid them. (For the record, I am one of those folks who would prefer they were not in the framework).

I agree that it isn't a good justification if it was this alone, but its also good to hear similarities to my opinion here. I'd like to reference the setRouteComponent ember-core RFC (26 March 2021) by Chris Garret (core team member who you probably already know had a large impact on the overall shape of Octane) which says:

> **Alternatives to controllers**: As mentioned above, this API will allow users to explore building Ember apps without using controllers. Removing controllers has been an agenda item on the roadmap for some time, they can be confusing to the modern Octane mental model and are a stumbling block for new users.

https://github.com/emberjs/rfcs/blob/set-route-component/text/0731-set-route-component.md#exploration-directions-unlocked-by-this-api

Whilst I agree with you that they are not currently deprecated I would have a differing viewpoint on you opinion that they are not going away, as per Chris's problem statement, _removing controllers has been an agenda item for some time, they can be confusing to the modern Octane mental model and are a stumbling block for new users_. All good reasons to avoid using Controllers (i.e. avoid creating Controller files ourselves)

There are also various places in the core Ember RFC repository that I saw that echo this sentiment.

You mentioned that the core team has stated on multiple occasions in the past, do you know how long ago you read this, could things have changed since then with the advent of a more modern Octane model?

> I generally prefer to avoid relying on Controllers for most things, except for the specific things they are good at and which are too burdensome to re-implement in other ways. In those cases, writing a few lines of controller code is usually the best option. I don't think there is anything wrong with having the occasional defined Controller in this or any other Ember application, and this seems like the kind of use case where I would typically implement one.

Agree with the preference to avoid relying on Controllers for most things, but I would also like to understand in more detail why you view the multiple alternatives I've outlined here are too burdensome?

My current preferred approaches in order are:

1. Use the [centralized feature-flag/wiring approach](https://github.com/hashicorp/consul/compare/ui/scratch/empty) we've been using since ~1.11 (see above for reasons why I prefer this). I would like to note that the change here is very small (not as big as the changeset in @LevelbossMike 's draft PR), I also split up the commits to help review.
2. Use [`ember-truth-helpers` `or`](https://github.com/jmurphyau/ember-truth-helpers#usage) within the template. This is a simple template logic conditional and in my mind is easier to grok very simple template logic conditionals within the template, agree that differences in opinion here are at arguably style preference.
3. Use CSS, we have an 'engineering feature' in the app so when nspaces are enabled, the root html node has a `has-nspaces` class, we use this very occasionally to visually hide things based on this class. i.e. `html:not(.has-peers) .consul-service-search-bar .peername { display: none; }`
4. Use a getter in the backing class for the search bar and filter there instead of in a controller, this centralizes the functionality in the component meaning its 100% reusable.

Perhaps relatedly, I usually shy away from inheritance for implementing functionality like this using inheritance, it can eventually lead to using Mixins. I much prefer composition over inheritance, but I agree this is arguably a question of personal architectural style.

> I agree that it would be preferable if: Ember's Route class could handle the special/unique cases that we need to rely on the Controller class for, and/or Ember's Router supported Routable Components, but unfortunately neither of these things are the case and it's important that we are willing and able to accept that sometimes, we may have to work with things we'd otherwise prefer to avoid.

If the alternatives are too burdensome, I would agree with you here. So, I'd like to hear more how the alternatives suggested (or otherwise) are too burdensome in case I've missed something there.

> I'd like to use this as an opportunity to re-evaluate our rationale for not using Controllers, and request that we consider adjusting our approach to "use Controllers sparingly, but when appropriate"; this is the approach that I have found to be the best compromise over the years in terms of not not utilising them in the majority of scenarios but being able to take advantage of what they do offer when it's helpful.

I partially agree with you're statement here (considering Chris Garret's statement above), if we add the `no-controller` eslint rule it can be overridden on a case by case basis, but also 'nudge' the engineer to think if a Controller really is the most appropriate tool to use. I think it would make a great addition, but I think we can decide this outside of this PR.

>> As part of the modernization of (not necessarily future proofing) our application we planned to remove mixins and controllers plus some other things, we would like to add the no-controllers lint rule that is specifically for this (we already have a lint rule for avoiding mixins)

> Controllers are, for better or worse, not going away, 

> they are still included in the context of what is considered to be a modern Ember application, and I don't feel that it's hurting us or the application code to include them on occasions where doing so is reasonable.

Also agree, a `no-controller` lint rule would prompt the engineer to think about whether this is the best approach.

> A final question to consider: If we were to move forward with the Controller approach in this PR, what would the concrete negative impacts to the Cluster Peering feature and the Consul UI application be? Would the feature cease to be usable? Would the application cease to function? Does the use of a Controller present immediate detrimental impact to the usability of the application by the end user? I understand that there is a clear difference in stylistic preferences in the discussion in this PR, but I want to ensure we are thinking critically about whether this approach meaningfully compromises the application in some way.

Thats a really good question, and applicable to lots of situations. I would like to ask those questions of all the alternatives/options given here at once if possible as I think the answer is almost the same.

For all 5 or so options and alternatives, the answer is I personally would say the answer is no, apart from in one case I can think of, which I partly detailed at the very top of this response.

To try to give a simplified example of what I mean in that we can ask this question of all options and get the same answer. Given a glass of milk, water, fizzy drink, or beer - which one will quench your thirst? All of them? I'd say they are all fine to be used to quench your thirst. Which one is more likely to damage your health longer term? Well thats different, I'd dare say the beer 😬 😆 

Separately, to be completely honest the Controller thing isn't my main concern here (but still a concern), its more that we are not sticking to what the team has been discussing, writing about and working towards for 'I think' over a year.

Lastly I thought of another alternative that I'd like to propose temporarily to ship this feature soon and then come back and revisit at a later date.

Using [Routes](https://github.com/hashicorp/consul/compare/main...ui/scratch/empty-w-route), which is where you would usually specify the model for the template using Ember, which is pretty much what we are doing here right now.

I'm assuming that ^ would be acceptable to all of us if staying on the "centralized wiring/feature-flagging" train is not an option in the immediate term, and I think is a good compromise in order to merge this PR.

As always, please let me know if you have any further questions or concerns off the back of this, I really believe we can come to some sort of compromise here and I thank both of your for the work and discussion so far 🙏

evrowe:
It's clear that we are not going to make forward progress on this discussion here and we will need to take it offline to come to a resolution. @johncowen @LevelbossMike I will be following up with both of you.

evrowe:
I would like for this Pull Request to move forward using the Controller implementation that it is currently relying on.

If we find after merge that this is causing issues which cannot be fixed in any way other than changing the implementation approach to do something else, we can re-evaluate the proposed alternatives and select something else.

@LevelbossMike Would you mind re-pushing this branch to kick tests back off and make sure CI passes?

evrowe (approved):
Provided there are no other changes required to address issues encountered after re-running CI, this is good to merge

LevelbossMike:
Thanks @evrowe - Closing in favor of https://github.com/hashicorp/consul/pull/13875 - couldn't figure out how to execute tests properly when PRing from my fork.

PR #13780, opened 2022-07-15
peerstream: fix test assertions

rboyer:
This ensures the test cases pass in both OSS and Enterprise. Example of how the assertion fails in enterprise:

```
   	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -36,3 +36,3 @@
        	            	  ImportedServices: (map[string]struct {}) (len=1) {
        	            	-  (string) (len=3) "api": (struct {}) {
        	            	+  (string) (len=19) "default/default/api": (struct {}) {
        	            	   }
```


acpana (approved):
🤦🏼 ; thank you for taking this on!

mkeeler (approved):

PR #13792, opened 2022-07-18
ui: Peer token use form

johncowen:
### Description
Add a 'token use' form, very similar to the last 2 PRs on peer establishment flow.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

/cc @LevelbossMike

### DIFF
--- a/ui/packages/consul-peerings/app/components/consul/peer/form/generate/index.hbs
+++ b/ui/packages/consul-peerings/app/components/consul/peer/form/generate/index.hbs
@@ -52,7 +52,7 @@
             item=@item
             token=fsm.state.context.PeeringToken
             regenerate=@regenerate
-            onclick=(queue (set @item 'Name' '') (fn fsm.dispatch "RESET"))
+            onclick=(queue (set @item 'Name' ''))
           )
           Actions=(component "consul/peer/form/token/actions"
             token=fsm.state.context.PeeringToken
--- a/ui/packages/consul-peerings/app/components/consul/peer/form/generate/index.hbs
+++ b/ui/packages/consul-peerings/app/components/consul/peer/form/generate/index.hbs
@@ -47,6 +47,19 @@
       </fsm.State>
 
       <fsm.State @matches={{'success'}}>
+        {{yield (hash
+          Fieldsets=(component "consul/peer/form/token/fieldsets"
+            item=@item
+            token=fsm.state.context.PeeringToken
+            regenerate=@regenerate
+            onclick=(queue (set @item 'Name' '') (fn fsm.dispatch "RESET"))
+          )
+          Actions=(component "consul/peer/form/token/actions"
+            token=fsm.state.context.PeeringToken
+            item=@item
+            id=id
+          )
+        )}}
       </fsm.State>
 
     </form>
--- a/ui/packages/consul-peerings/app/components/consul/peer/form/token/actions/index.hbs
+++ b/ui/packages/consul-peerings/app/components/consul/peer/form/token/actions/index.hbs
@@ -1,0 +1,18 @@
+<div
+  class={{class-map
+    'consul-peer-form-token-actions'
+  }}
+  ...attributes
+>
+  <Action
+    ...attributes
+    {{with-copyable @token}}
+  >
+    Copy token
+  </Action>
+  <Action
+    {{on 'click' @onclose}}
+  >
+    Close
+  </Action>
+</div>
--- a/ui/packages/consul-peerings/app/components/consul/peer/form/index.scss
+++ b/ui/packages/consul-peerings/app/components/consul/peer/form/index.scss
@@ -1,0 +1,62 @@
+
+.consul-peer-form-token-actions {
+  button:first-of-type {
+    @extend %primary-button;
+  }
+  button:last-of-type {
+    @extend %secondary-button;
+  }
+}
+
+.consul-peer-form-generate {
+  & {
+    width: 416px;
+    min-height: 200px;
+  }
+  ol {
+    list-style-position: outside;
+    list-style-type: none;
+    counter-reset: hexagonal-counter;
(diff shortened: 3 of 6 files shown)


ui/packages/consul-peerings/app/components/consul/peer/form/generate/index.hbs
```diff
@@ -47,6 +47,19 @@
       </fsm.State>
 
       <fsm.State @matches={{'success'}}>
+        {{yield (hash
+          Fieldsets=(component "consul/peer/form/token/fieldsets"
+            item=@item
+            token=fsm.state.context.PeeringToken
+            regenerate=@regenerate
+            onclick=(queue (set @item 'Name' '') (fn fsm.dispatch "RESET"))
```
LevelbossMike, line 55: I think we got rid of the RESET for the generate form, correct? Do we still need it though? Why do we dispatch `RESET` here? Will we exit the success state without handling `RESET` when we click?

ui/packages/consul-peerings/app/components/consul/peer/form/token/actions/index.hbs
```diff
@@ -0,0 +1,18 @@
...
+  ...attributes
+>
+  <Action
+    ...attributes
+    {{with-copyable @token}}
+  >
+    Copy token
+  </Action>
+  <Action
+    {{on 'click' @onclose}}
```
LevelbossMike, line 14: I'm not seeing the @onclose action being passed when yielding this component. do we want to handle the case for when users don't provide @onclose?

ui/packages/consul-peerings/app/components/consul/peer/form/generate/index.hbs
```diff
@@ -47,6 +47,19 @@
       </fsm.State>
 
       <fsm.State @matches={{'success'}}>
+        {{yield (hash
+          Fieldsets=(component "consul/peer/form/token/fieldsets"
+            item=@item
+            token=fsm.state.context.PeeringToken
```
LevelbossMike, line 53: who is assigning to fsm's state context? I don't see this being handled in the machine config for the generate form and no update to the configuration in this PR?

ui/packages/consul-peerings/app/components/consul/peer/form/generate/index.hbs
```diff
@@ -47,6 +47,19 @@
...
+            regenerate=@regenerate
```
LevelbossMike, line 54: I know this is already being used in this template but where are we passing this?

LevelbossMike (commented):
some comments

ui/packages/consul-peerings/app/components/consul/peer/form/generate/index.hbs
johncowen: Oh good spot, yeah I think we can get rid of this seeing as we got rid of it in another PR, sorry! Gimme 2 tics

ui/packages/consul-peerings/app/components/consul/peer/form/token/actions/index.hbs
johncowen: 👍 I can add add an `optional`, 2 tics

ui/packages/consul-peerings/app/components/consul/peer/form/generate/index.hbs
johncowen: Oh let me look I may have missed a file here 👀 , oh or maybe thats coming in the "wire all this together" PR, let me look and I'll come back to you

ui/packages/consul-peerings/app/components/consul/peer/form/generate/index.hbs
johncowen: Sometimes you want to "generate" a token, therefore you need to see a form to allow you to type the peers name, before seeing a token. Whereas sometimes you already have the peer, and you just need to 'regenerate' the token, therefore not needing to enter the name. This `@regenerate` boolean is just to say whether to immediately load a token, or show a form to ask for a name first.

ui/packages/consul-peerings/app/components/consul/peer/form/token/actions/index.hbs
johncowen: Hmm honestly, I would rather this errored if you don't add an `@onclose` otherwise the button doesn't do anything. This is also a single usecase component, mainly a component for better organisation, so highly unlikely that this will never be set, and again if that is the case then I would rather it wasn't silent. Are you ok me leaving this as is?

ui/packages/consul-peerings/app/components/consul/peer/form/generate/index.hbs
johncowen: This is set here I believe, as a result of the DataSource call:

https://github.com/hashicorp/consul/blob/main/ui/packages/consul-peerings/app/components/consul/peer/form/generate/index.hbs#L44

(just incase, I didn't miss a file here this is all good still)

johncowen: Does this mean you are ok to approve this PR?

I agree that this line would be better to use something else, probably event passing longer term, but its hidden under here so it's no big deal when we want to change that.

The interface we are using in https://github.com/hashicorp/consul/blob/main/ui/packages/consul-peerings/app/components/consul/peer/form/generate/index.hbs#L44 would probably stay the same.

Little note here, we are using `xstate/fsm` not `xstate`, so whilst its similar its not the same thing.

LevelbossMike: Yes I'm ok with merging 

On a quick note - xstate/fsm pretty much works the same without nesting etc. assigning is the same for example

PR #13784, opened 2022-07-15
peering: track exported services

acpana:
### Description

This patch exposes the `ExportedServicesCount` thru our List/Read endpoints for peerings.

This PR is quite similar to https://github.com/hashicorp/consul/pull/13718 .

### Testing & Reproduction steps
* updated the `stream_tracker`
* updated the `leader_peering_test`

PR #13795, opened 2022-07-18
peering: refactor reconcile, cleanup

acpana:
### Description

This is a small refactor that addresses:

* refactoring the way we copy a peer on `PeeringRead` and `PeeringList`
* punted/ missed feedback from https://github.com/hashicorp/consul/pull/13619 & https://github.com/hashicorp/consul/pull/13784

### DIFF
--- a/agent/rpc/peering/service.go
+++ b/agent/rpc/peering/service.go
@@ -13,6 +13,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	grpcstatus "google.golang.org/grpc/status"
+	"google.golang.org/protobuf/proto"
 
 	"github.com/hashicorp/consul/acl"
 	"github.com/hashicorp/consul/agent/consul/state"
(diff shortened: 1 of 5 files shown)


agent/rpc/peering/service.go
```diff
@@ -379,24 +369,27 @@ func (s *Server) PeeringList(ctx context.Context, req *pbpeering.PeeringListRequ
...
 	return &pbpeering.PeeringListResponse{Peerings: cPeerings}, nil
 }
 
+// TODO(peering): Get rid of this func when we stop using the stream tracker for imported/ exported services and the peering state
+// reconcilePeering enriches the peering with the following information:
+// -- PeeringState.Active if the peering is active
+// -- ImportedServicesCount and ExportedServicesCount
+// NOTE: we return a new peering with this additional data
+func (s *Server) reconcilePeering(peering *pbpeering.Peering) *pbpeering.Peering {
+	newPeeringState := s.reconciledStreamStateHint(peering.ID, peering.State)
```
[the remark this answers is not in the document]
acpana: Thanks for the call outs; I think I addressed all during the last couple of commits.

agent/rpc/peering/service.go
```diff
@@ -379,34 +370,49 @@ func (s *Server) PeeringList(ctx context.Context, req *pbpeering.PeeringListRequ
...
+// reconcilePeering enriches the peering with the following information:
+// -- PeeringState.Active if the peering is active
+// -- ImportedServicesCount and ExportedServicesCount
+// NOTE: we return a new peering with this additional data
+func (s *Server) reconcilePeering(peering *pbpeering.Peering) *pbpeering.Peering {
+	streamState, found := s.Tracker.StreamStatus(peering.ID)
+	if !found {
+		s.Logger.Trace("did not find peer in stream tracker; cannot populate imported and"+
+			" exported services count or reconcile peering state", "peerID", peering.ID)
+		return peering
```
mkeeler, line 390: Nice! We even have copy-on-write in the case where we aren't tracking the stream.

kisunji: Should this be a WARN? It looks like a bad state we'd want to detect

PR #13796, opened 2022-07-18
peerstream: set keepalive enforcement to 15s

lkysow:
The client is set to send keepalive pings every 30s. The server
keepalive enforcement must be set to a number less than that,
otherwise it will disconnect clients for sending pings too often.
MinTime governs the minimum amount of time between pings.

Need to set this to match the client policy set in https://github.com/hashicorp/consul/pull/13733/files. Otherwise it defaults to 5m which caused the server to disconnect in my testing. This setting matches the internal gRPC server (https://github.com/hashicorp/consul/blob/main/agent/grpc-internal/handler.go#L37-L39).

Without this I'm seeing connection resets in my testing:

```
INFO[0108] [transport] transport: Got too many pings from the client, closing the connection.  system=system
```

See also https://www.evanjones.ca/grpc-is-tricky.html.

### DIFF
--- a/agent/grpc-external/server.go
+++ b/agent/grpc-external/server.go
@@ -5,6 +5,8 @@ import (
 	recovery "github.com/grpc-ecosystem/go-grpc-middleware/recovery"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/keepalive"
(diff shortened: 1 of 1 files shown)


freddygv (approved):
Good find, thanks for fixing this

PR #13797, opened 2022-07-18
makefile: give better error for tool installed by wrong package

lkysow:
I had protoc-gen-go installed through `google.golang.org/protobuf` instead of
`github.com/golang/protobuf` and `make proto` was failing silently.
This change will ensure you get an error:

```
protoc-gen-go is already installed by module "google.golang.org/protobuf" but
should be installed by module "github.com/golang/protobuf".
Delete it and re-run to re-install.
```

### DIFF
--- a/build-support/scripts/devtools.sh
+++ b/build-support/scripts/devtools.sh
@@ -188,8 +188,14 @@ function install_versioned_tool {
     fi
 
     if command -v "${command}" &>/dev/null ; then
-        got="$(go version -m $(which "${command}") | grep '\bmod\b' | grep "${module}" |
-            awk '{print $2 "@" $3}')"
+        mod_line="$(go version -m "$(which "${command}")" | grep '\smod\s')"
+        act_mod=$(echo "${mod_line}" | awk '{print $2}')
+        if [[ "$module" != "$act_mod" ]]; then
+            err "${command} is already installed by module \"${act_mod}\" but should be installed by module \"${module}\". Delete it and re-run to re-install."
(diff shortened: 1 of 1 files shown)


build-support/scripts/devtools.sh
```diff
@@ -188,8 +188,14 @@ function install_versioned_tool {
     fi
 
     if command -v "${command}" &>/dev/null ; then
```
lkysow, line 190: I'm splitting up the `got` command into three parts. First getting the line, then getting the module name.

I've also changed the grep from `\b` to `\s` because it was incorrectly matching on `mod` for:

```
> go version -m /Users/lkysow/go/bin/mog | grep '\bmod\b'                                                                                                                                                                                                                                                                                                  ⎈ eks-1  
        mod     github.com/hashicorp/mog        v0.3.0  h1:wQkGfQAreeVrzyBiqX6nS8ZD9mtAY/MCbkclSCJJMXI=
        dep     golang.org/x/mod        v0.6.0-dev.0.20220106191415-9b9b3d81d5e3        h1:kQgndtyPBW/JIYERgdxfwMYh3AVStj88WQTlNDi2a+o=

```

rboyer (approved):
LGTM. I temporarily erased my `$GOPATH/bin` and reran `make tools` off of your branch twice and it did the right thing even on linux.

kisunji (approved):
Tested on mac and LGTM

lkysow:
❤️ thanks for the testing!!!

PR #13799, opened 2022-07-19
ui: chore - fix CI test-suite

LevelbossMike:
### Description

Fixes current failures in CI that were missed because people with non-contributor rights, like yours truly, won't have the CI test-suite executed for them.

Here's the link to the actual failure in the OSS test-suite: https://github.com/hashicorp/consul/runs/7402516620?check_suite_focus=true#step:6:1401

Link to one of the failures in ENT test-suite: https://github.com/hashicorp/consul/runs/7402525270?check_suite_focus=true#step:6:5038 ← you can see that all 4 test-failures are failing because of the same reason as OSS version is failing.

The fix here is to make the datacenter-item selector in the datacenter-selector-page-object be more specific. I introduced a new `data-test-`-attribute for that that only gets added to the actual datacenter items in the listing but not the `Separator`-component I introduced in https://github.com/hashicorp/consul/pull/13753

I also fixed a linting issue that I missed because, 🥁  non-contributor PRs won't execute the CI pipeline lint task properly.

cc @johncowen 

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-ui/app/components/hashicorp-consul/pageobject.js
+++ b/ui/packages/consul-ui/app/components/hashicorp-consul/pageobject.js
@@ -50,7 +50,7 @@ export default (collection, clickable, attribute, is, authForm, emptyState) => s
     ':checked',
     '[data-test-nspace-menu] > input[type="checkbox"]'
   );
-  page.navigation.dcs = collection('[data-test-datacenter-menu] li', {
+  page.navigation.dcs = collection('[data-test-datacenter-menu] [data-test-dc-item]', {
     name: clickable('a'),
   });
   return page;
--- a/ui/packages/consul-ui/app/components/consul/datacenter/selector/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/datacenter/selector/index.hbs
@@ -27,6 +27,7 @@
             </menu.Separator>
             {{#each menu.items as |item|}}
               <menu.Item
+                data-test-dc-item
                 aria-current={{if (eq @dc.Name item.Name) 'true'}}
                 class={{class-map
                   (array 'is-local' item.Local)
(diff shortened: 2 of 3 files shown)


ui/packages/consul-ui/app/components/hashicorp-consul/pageobject.js
```diff
@@ -50,7 +50,7 @@ export default (collection, clickable, attribute, is, authForm, emptyState) => s
     ':checked',
     '[data-test-nspace-menu] > input[type="checkbox"]'
   );
-  page.navigation.dcs = collection('[data-test-datacenter-menu] li', {
+  page.navigation.dcs = collection('[data-test-datacenter-menu] [data-test-dc-item]', {
```
johncowen, line 53: I was wondering why the `li` didn't work here and spotted we've added some erroneous HTML:

https://github.com/hashicorp/consul/blob/bea497b97ab1c1a1060ac6eaae487a40d2125716/ui/packages/consul-ui/app/components/consul/datacenter/selector/index.hbs#L59

^ That means we have:

```html
<ul>
  <li>
    <li>dcname</li>
  </li>
</ul>
```
...so an li directly inside an li, which I don't think is allowed.

As this is all related to https://github.com/hashicorp/consul/pull/13753 could this also be fixed up here? (sorry I missed this in the original review)

(P.S. I don't think this spot answers the question of the `li` not working btw, but I was just curious about that more than anything)

johncowen (commented):
🥁 Hey sorry about the CI things, hopefully we can get that resolved for you (and me!) as soon as possible.

I left an inline comment that I think we should fix up here also.

ui/packages/consul-ui/app/components/hashicorp-consul/pageobject.js
LevelbossMike: The separator also will display as a `li` in the list. that's the reason for the change.

johncowen: oh! Of course 😆 thankyou!

johncowen (approved):
Thanks for fixing up that HTML here also!

PR #13805, opened 2022-07-19
peering: Add send mutex to protect against concurrent sends

lkysow:
### Description
As part of the heartbeating work, I'm going to use a separate goroutine to send heartbeats. This simplifies the code vs adding the heartbeat sending to the main work loop and also follows how waypoint sends heartbeats, but it necessitates that  we add a mutex to sending because the `Send()` function isn't goroutine safe.

### DIFF
--- a/agent/grpc-external/services/peerstream/stream_resources.go
+++ b/agent/grpc-external/services/peerstream/stream_resources.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"sync"
 
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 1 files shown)


agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -215,7 +219,10 @@ func (s *Server) HandleStream(streamReq HandleStreamRequest) error {
 		})
 		logTraceSend(logger, sub)
 
-		if err := streamReq.Stream.Send(sub); err != nil {
+		sendMutex.Lock()
```
mkeeler, line 222: Do you think it would be worthwhile create a func like:

```
sendLocked := func(msg *pbpeerstream.ReplicationMessage) error {
   sendMutex.Lock()
   defer sendMutext.Unlock()
   return streamReq.Stream.Send(msg)
}
```

And then use this everywhere as opposed to copying the mutex handling a bunch of places.

lkysow: Yep, I like that.

agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -204,6 +205,25 @@ func (s *Server) HandleStream(streamReq HandleStreamRequest) error {
 	)
 	subCh := mgr.subscribe(streamReq.Stream.Context(), streamReq.LocalID, streamReq.PeerName, streamReq.Partition)
 
+	// We need a mutex to protect against simultaneous sends to the client.
+	var sendMutex sync.Mutex
+
+	// streamSend is a helper function that sends msg over the stream
```
lkysow, line 211: This helper wraps the mutex and also extracts `logTraceSend` and `status.TrackSendError`. You'll see those calls removed from where we use the helper now.

agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -213,16 +233,12 @@ func (s *Server) HandleStream(streamReq HandleStreamRequest) error {
 			ResourceURL: resourceURL,
 			PeerID:      streamReq.RemoteID,
 		})
-		logTraceSend(logger, sub)
-
-		if err := streamReq.Stream.Send(sub); err != nil {
+		if err := streamSend(sub); err != nil {
 			if err == io.EOF {
 				logger.Info("stream ended by peer")
-				status.TrackReceiveError(err.Error())
```
lkysow, line 221: I think this was a mistake to use `status.TrackReceiveError` here instead of `status.trackSendError`

agent/grpc-external/services/peerstream/stream_resources.go
lkysow: done

PR #13806, opened 2022-07-19
peering: Add heartbeating to peering streams

lkysow:
### Description
Add a heartbeating mechanism to peering streams.

Follows closely waypoint's approach: https://github.com/hashicorp/waypoint/pull/98

Senders and receivers each send heartbeats at a defined interval (15s). If a heartbeat isn't received for 2 minutes, the connection is closed.

To dos in future PRs:
* Track last receipt of any message (including heartbeats) in the peering status
* Add jitter to heartbeat sending

### DIFF
--- a/agent/grpc-external/services/peerstream/stream_resources.go
+++ b/agent/grpc-external/services/peerstream/stream_resources.go
@@ -292,12 +292,14 @@ func (s *Server) HandleStream(streamReq HandleStreamRequest) error {
 	}()
 
 	// incomingHeartbeatCtx will complete if incoming heartbeats time out.
-	incomingHeartbeatCtx, cancel := context.WithCancel(context.Background())
-	// incomingHeartbeatTimer cancels incomingHeartbeatCtx if it isn't reset in time.
-	// It gets reset by receiving heartbeats.
-	incomingHeartbeatTimer := time.AfterFunc(s.incomingHeartbeatTimeout, func() {
-		cancel()
-	})
+	incomingHeartbeatCtx, incomingHeartbeatCtxCancel :=
+		context.WithTimeout(context.Background(), s.incomingHeartbeatTimeout)
+	// NOTE: It's important that we wrap the call to cancel in a wrapper func because during the loop we're
+	// re-assigning the value of incomingHeartbeatCtxCancel and we want the defer to run on the last assigned
+	// value, not the current value.
+	defer func() {
+		incomingHeartbeatCtxCancel()
+	}()
 
 	for {
 		select {
@@ -470,7 +472,16 @@ func (s *Server) HandleStream(streamReq HandleStreamRequest) error {
 			}
 
 			if msg.GetHeartbeat() != nil {
-				incomingHeartbeatTimer.Reset(s.incomingHeartbeatTimeout)
+				// Reset the heartbeat timeout by creating a new context.
+				// We first must cancel the old context so there's no leaks. This is safe to do because we're only
+				// reading that context within this for{} loop, and so we won't accidentally trigger the heartbeat
+				// timeout.
+				incomingHeartbeatCtxCancel()
+				// NOTE: IDEs and govet think that the reassigned cancel below never gets
+				// called, but it does by the defer when the heartbeat ctx is first created.
+				// They just can't trace the execution properly for some reason (possibly golang/go#29587).
+				incomingHeartbeatCtx, incomingHeartbeatCtxCancel =
+					context.WithTimeout(context.Background(), s.incomingHeartbeatTimeout)
 			}
 
 		case update := <-subCh:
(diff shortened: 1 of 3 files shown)


agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -25,6 +26,11 @@ type BidirectionalStream interface {
 	Context() context.Context
 }
 
+var (
+	outgoingHeartbeatInterval = 15 * time.Second
+	incomingHeartbeatTimeout  = 2 * time.Minute
```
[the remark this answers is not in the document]
mkeeler: I think the usual way would be to wire them through server config as unexported fields. `NewServer` would then default them to these values (which could be consts) when unset/zero.

agent/grpc-external/services/peerstream/stream_test.go
```diff
@@ -869,6 +868,179 @@ func TestStreamResources_Server_CARootUpdates(t *testing.T) {
...
+
+	// Set up a goroutine to send the heartbeat every 1/2 of the timeout.
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	go func() {
+		// This is just a do while loop. We want to send the heartbeat right away to start
+		// because the test setup above takes some time and we might be close to the heartbeat
+		// timeout already.
+		for {
+			require.NoError(t, client.Send(heartbeatMsg))
```
mkeeler, line 1025: Using the `t` within a go routine that isn't the test function executed by the go test framework make `go test` angry. It outputs warnings about this being bad and potentially could cause issues with things modify test state after the test has completed.

I think the general approach for getting around this is to create a `chan error` and have the go routine send to that chan any error it may encounter or nil. Then the main test func would recv from that chan (you would also need to manually call the cancel func earlier) and check the value for a non-nil error.

agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -433,6 +481,10 @@ func (s *Server) HandleStream(streamReq HandleStreamRequest) error {
 				return nil
 			}
 
+			if msg.GetHeartbeat() != nil {
+				incomingHeartbeatTimer.Reset(incomingHeartbeatTimeout)
```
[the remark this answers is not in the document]
lkysow: I'm not against changing this but digging into it a bit more. I agree that the docs aren't super clear but they do say:

> For a Timer created with AfterFunc(d, f), Reset either reschedules when f will run, in which case Reset returns true, or schedules f to run again, in which case it returns false. When Reset returns false, Reset neither waits for the prior f to complete before returning nor does it guarantee that the subsequent goroutine running f does not run concurrently with the prior one. If the caller needs to know whether the prior execution of f is completed, it must coordinate with f explicitly.


In the case that we're resetting the timer right after it's been triggered, we'll just end up calling `cancel()` twice from:

```go
	incomingHeartbeatTimer := time.AfterFunc(s.incomingHeartbeatTimeout, func() {
		cancel()
	})
```

Which isn't an issue.

Also this is the same pattern used by waypoint which means this has been in the wild for the last year (https://github.com/hashicorp/waypoint/blob/main/internal/server/boltdbstate/job.go#L1015).

So give that I'm slightly inclined to leave it but happy to change it too if you or other reviewers think it's still a bit dangerous.

agent/grpc-external/services/peerstream/stream_test.go
lkysow: Awesome catch thanks for that tip! I added the errCh and I think I'm handling it as suggested.

agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -286,6 +319,11 @@ func (s *Server) HandleStream(streamReq HandleStreamRequest) error {
 
 			return nil
 
+		// We haven't received a heartbeat within the expected interval. Kill the stream.
+		case <-incomingHeartbeatCtx.Done():
+			logger.Error("ending stream due to heartbeat timeout")
+			return fmt.Errorf("heartbeat timeout")
```
kisunji, line 325: The caller of `HandleStream` will retry on error; I'm assuming heartbeat timeout is something we consider retryable?

agent/grpc-external/services/peerstream/stream_resources.go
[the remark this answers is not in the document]
lkysow: I chatted with Matt a bit more and we're going to go with the context approach.

agent/grpc-external/services/peerstream/stream_resources.go
lkysow: Sounds good, thanks for thinking this through!

mkeeler (approved):

PR #13811, opened 2022-07-20
peering: emit exported services count metric

acpana:

### Description
This patch adds a `consul_peering_exported_services` metric with labels for the `peer name` and `partition` (for ENT). This, in addition to more metrics, will help drive a better understanding about the overall state of a peering.

To be clear, we can probably add more metrics using this pattern but we are starting with just the one for now.

* `leader_peering_test` with mock sink and metrics instance.

#### pr todo:

* [x] add `peer_id` label to metrics
* ~[ ] add external facing docs about the metric~ --> adding this as a separate PR

### DIFF
--- a/agent/consul/leader_peering.go
+++ b/agent/consul/leader_peering.go
@@ -91,7 +91,8 @@ func (s *Server) emitPeeringMetrics(ctx context.Context, logger hclog.Logger, ti
 				esc := status.GetExportedServicesCount()
 				part := peer.Partition
 				labels := []metrics.Label{
-					{Name: "peering", Value: peer.Name},
+					{Name: "peer_name", Value: peer.Name},
+					{Name: "peer_id", Value: peer.ID},
 				}
 				if part != "" {
 					labels = append(labels, metrics.Label{Name: "partition", Value: part})
--- a/agent/consul/leader_peering.go
+++ b/agent/consul/leader_peering.go
@@ -51,20 +51,6 @@ func (s *Server) runPeeringMetrics(ctx context.Context) error {
 	logger := s.logger.Named(logging.PeeringMetrics)
 	defaultMetrics := metrics.Default
 
-	retryLoopBackoff(ctx, func() error {
-		if err := s.emitPeeringMetrics(ctx, logger, ticker, defaultMetrics()); err != nil {
-			return err
-		}
-		return nil
-
-	}, func(err error) {
-		s.logger.Error("error emitting peering stream metrics", "error", err)
-	})
-
-	return nil
-}
-
-func (s *Server) emitPeeringMetrics(ctx context.Context, logger hclog.Logger, ticker *time.Ticker, metricsImpl *metrics.Metrics) error {
 	for {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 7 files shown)


agent/consul/leader_peering.go
```diff
@@ -27,8 +29,83 @@ import (
...
+				if !found {
+					logger.Trace("did not find status for", "peer", peer.Name)
+				}
+
+				esc := status.GetExportedServicesCount()
+				part := peer.Partition
+				labels := []metrics.Label{
+					{Name: "peering", Value: peer.Name},
+				}
+				if part != "" {
```
acpana, line 101: not that we grab the partition from the state store value; I think this is more "accurate" for the peering. We can prepend the `partition` label of the metric with `peer_partition` to make it clear.

agent/consul/leader_peering_test.go
```diff
@@ -615,7 +616,7 @@ func insertTestPeeringData(t *testing.T, store *state.Store, peer string, lastId
 	return lastIdx
 }
 
-// TODO(peering): once we move away from leader only request for PeeringList, move this test to consul/server_test maybe
+// TODO(peering): once we move away from keeping state in stream tracker only on leaders, move this test to consul/server_test maybe
```
acpana, line 619: the wording was bugging me here; ( i wrote it hehe 😅 )

It's not the PeeringList endpoint's problem, but rather the fact that we rely on the stream tracker information that requires us to make these `leader` only for now

agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -474,8 +474,8 @@ func getTrustDomain(store StateStore, logger hclog.Logger) (string, error) {
 	return connect.SpiffeIDSigningForCluster(cfg.ClusterID).Host(), nil
 }
 
-func (s *Server) StreamStatus(peer string) (resp Status, found bool) {
-	return s.Tracker.StreamStatus(peer)
+func (s *Server) StreamStatus(peerID string) (resp Status, found bool) {
```
acpana, line 477: Even just as a passthrough, this confused me if it wanted the name or ID for the peering so I renamed it to make it clearer to myself

mkeeler (commented):
Regarding the implementation here, did you see an advantage to do this in the leader code vs just having the stream tracker emit the metric as new services are added/removed?

The other approach would be significantly less code to maintain.

agent/consul/leader_peering.go
```diff
@@ -27,8 +29,83 @@ import (
...
+
+			// "Zero-out" the metric on exit so that when prometheus scrapes this
+			// metric from a non-leader, it does not get a stale value.
+			metrics.SetGauge(leaderExportedServicesCountKey, float32(0))
+			return nil
+		case <-ticker.C:
+			stateS := s.fsm.State()
+			ws := memdb.NewWatchSet()
+			ws.Add(stateS.AbandonCh())
+			ws.Add(ctx.Done())
```
kisunji, line 81: We're not blocking anything with this watchset; should it be removed?

agent/consul/leader_peering.go
```diff
@@ -27,8 +29,83 @@ import (
...
+}
+
+func (s *Server) runPeeringMetrics(ctx context.Context) error {
+	ticker := time.NewTicker(s.config.MetricsReportingInterval)
+	defer ticker.Stop()
+
+	logger := s.logger.Named(logging.PeeringMetrics)
+	defaultMetrics := metrics.Default
+
+	retryLoopBackoff(ctx, func() error {
```
kisunji, line 54: This retry loop shouldn't be necessary if we're ticking at a set interval and errors are suppressed, right?

agent/consul/leader_peering.go
acpana: ah yes 🤦🏼 , thanks for spotting that!

agent/consul/leader_peering.go
acpana: actually, we may error out on the `PeeringList` call for instance

acpana:
> Regarding the implementation here, ...

@mkeeler and I chatted offline about this; we have to go with this polling approach here because of `go-metrics`'s metric expiry behavior. If we don't keep remitting the metrics, they will get expired.

So we are continuing with the approach on this PR

agent/consul/leader_peering.go
```diff
@@ -27,8 +29,79 @@ import (
...
+			if err != nil {
+				return err
+			}
+
+			for _, peer := range peers {
+				logger.Info("emitting metrics", "peer", peer.Name)
+
+				status, found := s.peerStreamServer.StreamStatus(peer.ID)
+				if !found {
+					logger.Trace("did not find status for", "peer", peer.Name)
```
kisunji, line 88: this needs a `continue`

agent/consul/leader_peering.go
kisunji: The error eventually gets captured by `retryLoopBackoff` and logged anyways so let's remove the retry loop and log it directly.

It's very unlikely for `PeeringList` to fail and if it does, other parts of Consul are probably also failing.

The issue with the retry loop is that you're no longer emitting metrics at a regular interval; if hypothetically something was wrong, we would want the error logs to happen regularly instead of backing off exponentially.

agent/consul/leader_peering_test.go
```diff
@@ -904,3 +906,140 @@ func TestLeader_Peering_ImportedExportedServicesCount(t *testing.T) {
...
+		require.NoError(t, errM)
+	}()
+
+	retry.Run(t, func(r *retry.R) {
+		intervals := sink.Data()
+		require.Len(r, intervals, 1)
+		intv := intervals[0]
+
+		// the keys for a Gauge value look like: {serviceName}.{prefix}.{key_name};{label=value};...
+		keyMetric1 := fmt.Sprintf("consul.peering.test.consul.peering.exported_services;peer_name=my-peer-s1;peer_id=%s", s2PeerID1)
```
kisunji, line 1033: why does `consul.peering` get repeated here?

agent/consul/leader_peering.go
acpana: hm the cost of not having metrics is high for me but I don't like not having the metric on a regular cadence as you are pointing out;

let me try something; I'll remove the `retryLoopBackoff` and isolate the metric emission to a `emitOnce` func

agent/consul/leader_peering.go
acpana: oh yes, great catch ‼️

agent/consul/leader_peering_test.go
acpana: yeah this is a little confusing and we can change it;

The explanation basically folllows the comment above

On line 1016 we define the servicename as 

```
cfg := metrics.DefaultConfig("consul.peering.test")
```

The prefix for our metrics is `consul` by default. 

`peering.exported_services` is the metric key.

Put together, we have the first blurb: `consul.peering.test.consul.peering.exported_services` . The rest are the label key values.

kisunji: Is this a test-only label? Can we adjust it so make it look like how it would in prod or a realistic config?

agent/consul/leader_peering.go
```diff
@@ -27,8 +29,74 @@ import (
...
+}
+
+func (s *Server) emitPeeringMetricsOnce(logger hclog.Logger, metricsImpl *metrics.Metrics) error {
+	_, peers, err := s.fsm.State().PeeringList(nil, *structs.NodeEnterpriseMetaInPartition(structs.WildcardSpecifier))
+	if err != nil {
+		return err
+	}
+
+	for _, peer := range peers {
+		logger.Info("emitting metrics for", "peer_name", peer.Name)
```
kisunji, line 78: Log should be removed

kisunji (approved):
LGTM. One comment about test metrics labels and another on Info logs that can be removed

agent/consul/leader_peering.go
```diff
@@ -27,8 +29,74 @@ import (
...
+	defaultMetrics := metrics.Default
+
+	for {
+		select {
+		case <-ctx.Done():
+			logger.Info("stopping peering metrics")
+
+			// "Zero-out" the metric on exit so that when prometheus scrapes this
+			// metric from a non-leader, it does not get a stale value.
+			metrics.SetGauge(leaderExportedServicesCountKey, float32(0))
```
mkeeler, line 61: FYI this is unnecessary due to the bug I mentioned in the other comment. Its totally fine to leave this as-is.

mkeeler (changes requested):
I left one request regarding symmetric start/stop methods for the metrics routine and 2 informational comments about how other bugs could effect the metric emission.

mkeeler (approved):

PR #13812, opened 2022-07-20
ui: no partition and peer in bucket-list at the same time

LevelbossMike:
### Description
This PR updates the bucket-list component to never display `partition` and `peer` at the same time. Whenever we have a peer the bucket-list component will only display a peer and the corresponding namespace.

When partitions and namespaces are not supported (OSS) we won't surface the namespace anymore.

This PR:

* Updates the bucket-list component as described above
* Adds tests for the bucket-list component
* Updates the intentions table to use the bucket-list to surface Source information

#### Notes
* this PR does not update the intentions table destination to use the bucket-list component as it won't surface peering information and we don't want to change this behavior

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

cc @johncowen

### DIFF
--- a/ui/packages/consul-ui/app/components/consul/bucket/list/README.mdx
+++ b/ui/packages/consul-ui/app/components/consul/bucket/list/README.mdx
@@ -11,6 +11,9 @@ the namespace will be displayed, whereas if the partition is different it will
 show both the partition and namespace (as a namespace called 'team-1' in
 `partition-1` is different to a namespace called 'team-1' in `partition-2`)
 
+When the passed item contains a `PeerName`, this will be displayed in place of
+the `Partition`.
+
 Showing the service name is a tiny bit awkward (different boolean type,
 doesn't care about difference) and could be improved but we only use it for
 the read only view of intentions.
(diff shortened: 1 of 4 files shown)


johncowen (commented):
Hey 👋 

Thanks for getting onto this so quickly!

I'm going to give this a spin in but I had a few comments that we could do while I'm doing that:

- Did we add documentation for the new attribute? I just took a very quick look and it looks like its in the examples, but not in the args docs, also wondering if you could replace the text used in the docs to explain the functionality, I'm not totally sure if the docs that are there are correct now. It would be super handy if you could update that following your understanding of how this works.
- I think there are other places in the UI where we should reuse this component (basically anywhere we show a 'bucket' e.g. a peer, nspace or partition). Is this something you are planning on doing?
- This code is looking very difficult to grok https://github.com/hashicorp/consul/blob/b403ef8db04c0c9982b473ff7f1038d773efea7e/ui/packages/consul-ui/app/components/consul/bucket/list/index.js#L13-L34 and I'm not totally sure if it makes sense just reading through it (if the item.Namespace doesn't equal the nspace then additionally add the Service?). I need to stare at it a bit more, but I was wondering if it would be an idea to add some comments in here, not sure?

Let me know, non of these are blockers as such, but in the meantime I'm gonna give this a spin!

johncowen (approved):
Looks good!

We spoke briefly offline about replacing `Consul::Node::PeerList` with `Consul::Bucket::List` also to avoid duplicating components that do the same thing. Whilst we'd like to do this, it's not super straightforwards, so we can leave that to a future PR.

PR #13813, opened 2022-07-20
ui: Add peering establishment to the peer listing page

johncowen:
### Description
Adds first draft of Peering Establishment flow to the peer listing page

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

/cc @LevelbossMike

### DIFF
--- a/ui/packages/consul-peerings/app/templates/dc/peers/index.hbs
+++ b/ui/packages/consul-peerings/app/templates/dc/peers/index.hbs
@@ -1,4 +1,6 @@
-<Route @name={{routeName}} as |route|>
+<Route
+  @name={{routeName}}
+as |route|>
   <DataLoader
     @src={{
       uri '/${partition}/${nspace}/${dc}/peers'
@@ -45,10 +47,13 @@ as |sort filters items|}}
   <AppView>
     <BlockSlot @name="header">
       <h1>
-        <route.Title @title="Peers" />
+        <route.Title
+          @title="Peers"
+        />
       </h1>
     </BlockSlot>
     <BlockSlot @name="toolbar">
+
       {{#if (gt items.length 0)}}
         <Consul::Peer::SearchBar
           @search={{search}}
@@ -61,7 +66,55 @@ as |sort filters items|}}
       {{/if}}
 
     </BlockSlot>
+    <BlockSlot @name="actions">
+
+
+      <ModalDialog
+        @aria={{hash
+          label="Add peer connection"
+        }}
+        @onclose={{optional this.form.reset}}
+      as |modal|>
+        <BlockSlot @name="header">
+          {{did-insert (set this 'create' modal)}}
+          <h2>
+            Add peer connection
+          </h2>
+        </BlockSlot>
+        <BlockSlot @name="body">
+
+          {{#if modal.opened}}
+            <Consul::Peer::Form
+              @params={{route.params}}
+            as |form|>
+              <form.Form
+                @onchange={{loader.invalidate}}
+                @onsubmit={{modal.close}}
+              as |form|>
+                {{did-insert (set this 'form' form)}}
+                <form.Fieldsets />
+              </form.Form>
+            </Consul::Peer::Form>
+          {{/if}}
+
+        </BlockSlot>
+        <BlockSlot @name="actions">
+          <this.form.Actions
+            @onclose={{this.create.close}}
+          />
+        </BlockSlot>
+      </ModalDialog>
+      <Action
+        data-test-create
+        class="type-create"
+        {{on "click" (optional this.create.open)}}
+      >
+        Add peer connection
+      </Action>
+
+    </BlockSlot>
     <BlockSlot @name="content">
+
       <DataWriter
         @sink={{uri '/${partition}/${dc}/${nspace}/peer/'
           (hash
@@ -72,8 +125,8 @@ as |sort filters items|}}
         }}
         @type="peer"
         @label="Peer"
-        @ondelete={{refresh-route}}
       as |writer|>
+
         <BlockSlot @name="removed" as |after|>
           <Consul::Peer::Notifications
             {{notification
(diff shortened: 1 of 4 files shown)


ui/packages/consul-peerings/app/templates/dc/peers/index.hbs
```diff
@@ -60,8 +65,56 @@ as |sort filters items|}}
...
+
+
+      <ModalDialog
+        @aria={{hash
+          label="Add peer connection"
+        }}
+        @onclose={{optional this.form.reset}}
+      as |modal|>
+        <BlockSlot @name="header">
+          {{did-insert (set this 'create' modal)}}
```
LevelbossMike, line 79: is it possible to add the create button in the scope of the modal block instead? Or is this the usual pattern of registering the component instance on the template context to be able to trigger a modal to open/close?

ui/packages/consul-peerings/app/templates/dc/peers/index.hbs
```diff
@@ -60,8 +65,56 @@ as |sort filters items|}}
...
 
+    </BlockSlot>
+    <BlockSlot @name="actions">
+
+
+      <ModalDialog
+        @aria={{hash
+          label="Add peer connection"
+        }}
+        @onclose={{optional this.form.reset}}
```
LevelbossMike, line 76: won't the form state be reset when the `modal.opened` conditional is falsy?

ui/packages/consul-peerings/app/templates/dc/peers/index.hbs
```diff
@@ -83,62 +136,97 @@ as |sort filters items|}}
...
-            <collection.Collection>
+
+            <ModalDialog
+              @aria={{hash
+                label="Regenerate token"
+              }}
+              @onclose={{set this 'item' undefined}}
+            as |modal|>
+              <BlockSlot @name="header">
+                {{did-insert (set this 'regenerate' modal)}}
```
LevelbossMike, line 147: this is a pattern we really should look into - ideally this would just be yielded by the modal-dialogue itself to be able to do that

can we just yield out the action to open/close the modal [here](https://github.com/hashicorp/consul/blob/main/ui/packages/consul-ui/app/components/modal-dialog/index.hbs#L5) and then not use this register pattern?

LevelbossMike (changes requested):
Left some comments regarding the usage of the modal - not sure if this can be changed in reasonable time.

ui/packages/consul-peerings/app/templates/dc/peers/index.hbs
johncowen: Again, not totally sure if I understood correctly. The create button is being added in the scope of the modal block. But yes this is a usual pattern.

ui/packages/consul-peerings/app/templates/dc/peers/index.hbs
johncowen: Not totally sure I understand, the action to open/close the modal is being yielded by the modal-dialog itself. In the above code the `|modal|` is the yielded API:

```hbs
<ModalDialog ... as |modal|>
```

It gets yielded from inside the component [here](https://github.com/hashicorp/consul/blob/8d4baafd84532cb950a810fe494e16c4d9a2e559/ui/packages/consul-ui/app/components/modal-dialog/index.hbs#L40-L45)

In case it helps this is roughly the equivalent to the following javascript:

```javascript
class Template {
  constructor() {
    modal({aria: {label: 'Regenerate token'}}, (modal) => this.regenerate = modal);
    // use the modals API
    this.regenerate.close()
  }
}
```

Ideally it would be more like:

```javascript
class Template {
  constructor() {
    this.regenerate = modal({aria: {label: 'Regenerate token'}});
    // use the modals API
    this.regenerate.close()
  }
}
```

but that's a consequence of embers component structure.

Could you help me to understand your query further as I maybe don't follow?

johncowen (commented):
Added some clarifying info plus some questions, ty!

johncowen:
Hey @LevelbossMike

Could you clarify what changes are requested here for me, its not super clear from your comments, thanks!

ui/packages/consul-peerings/app/templates/dc/peers/index.hbs
evrowe: I'm not certain, but I think what @LevelbossMike is asking is whether the form will reset automatically when the `modal.opened` conditional changes, and therefore whether this is necessary to specify?

ui/packages/consul-peerings/app/templates/dc/peers/index.hbs
LevelbossMike: My question points to the following:

You are using a a render helper here to set the `regenerate` property on the template-context (the controller) here to the object that you are yielding out in the header block-slot. As far as I can tell this is basically what is happening here:

```js
this.set('regenerate', modalApiYieldedToHeaderBlockSlot)
```

You seem to need to do that because you want to interact with the modal from [other places of the template](https://github.com/hashicorp/consul/blob/8e27be16003215cd41a9f64fa6b7ca79852aafbd/ui/packages/consul-peerings/app/templates/dc/peers/index.hbs#L110). This is pretty unusual. Ember provides the ability to `yield` for being able to use the yielded values directly:

```hbs
<Modal as |modal|>
  <button type="button" {{on "click" modal.open}}>
    Open modal
  </button>
    <button type="button" {{on "click" modal.close}}>
    Close modal
  </button>
</Modal>
```

Instead of accessing the yielded api directly we are registering the yielded object to the template context. The way we are using the block slots seems to get into the way here - can we yield the modal api (maybe via different slot) so the above would be possible? Or is this a bigger refactor? Would this work?

```hbs
{{!-- ModalDialog.hbs --}}
<YieldSlot @name="api" @params={{block-params (hash
  open=(action "open")
  close=(action "close")
  opened=this.isOpen
  aria=aria
)}}>
  {{yield}}
</YieldSlot>

{{!-- ... Rest of modal template --}}

{{!-- and then use it like this --}}

<ModalDialog>
  {{!-- Rest of template ... --}}
   <BlockSlot @name="api" as |modal|>
    <Action
      data-test-create
      class="type-create"
      {{on "click" modal.open}}
     >
       Add peer connection
     </Action>
  </BlockSlot>
</ModalDialog>
```
The way the modal-dialogue uses the yield-slots doesn't seem to be the way they work. you have to pass the `params` instead of yielding to its content block apparently.

Using it like this would mean we won't need to register an object on the template context to be able to interact with the modal-component with the additional benefit that everything gets cleaned up properly when the template unrenders and we don't set properties that we then don't unset on the template context anymore.

evrowe: @LevelbossMike @johncowen This seems like a worthwhile discussion to have/revisit in the near future as we are able to dedicate more time to examining this implementation in detail, and determining potential paths for a refactor.

For now with regards to this modal implementation, given that this pattern is what we're using for other modals in the app, I would advise that we stick with what we have and focus on whether there are any major issues with this particular instance of the implementation. Let's make sure to log this as an item to come back around and do a technical review on in a few weeks post 1.13.

LevelbossMike:
@johncowen The only thing that is left to investigate is the resetting of the form - it looks like the form will unrender and should thus reset its state anyway. So we probably don't need to reset it explicity?

Regarding the modal dialoge I agree with @evrowe that this is not something that needs to be fixed now but we should investigate in the future of how we might be able to get around the registering of the modal dialogue on the template context to be able to interact with it - yielding out the api and using it directly should probably be good enough. But this is not something to change in this PR.

ui/packages/consul-peerings/app/templates/dc/peers/index.hbs
johncowen: The modal component is one of the components I've been wanting to upgrade for a very long time, but not been given the opportunity to do so. For historical background, a while ago I slowly but surely converted the majority of our components that use the Ember 2.18 supported block slots to Ember 3 supported slots (or contextual components) but I had to stop part way through - our Modal component is one of the ones that is still left over. So yes, lets upgrade it, but just not right now.

One thing I've noticed here which might help with some understanding:

```javascript
<ModalDialog>
  {{!-- Rest of template ... --}}
   <BlockSlot @name="api" as |modal|>
    <Action
      data-test-create
      class="type-create"
      {{on "click" modal.open}}
     >
       Add peer connection
     </Action>
  </BlockSlot>
</ModalDialog>
```

If the button to open the modal is inside the modal panel itself, how then do you click that button when the modal is closed? Just to be clear the component here is the modal panel itself, like the overlay and 'window' for the modal.

johncowen:
Ok sounds like the 2 things that were important here were:

1. Our modal component
2. The optionally call `this.form.reset` method.

I think everyone here agrees that we shouldn't be rebuilding our modal component right now, so I think I can safely leave that as is. Would love to have the opportunity to upgrade in future though.

In regards to the `reset` method, this is essentially a `noop`. We removed all the `RESET`s from the code in PRs leading up to this, so `this.form.reset` is undefined but guarded with a `optional` which means that will do absolutely nothing if it is unset, which is the case here.

I've added a commit removing that line. If for whatever reason the form does need to be `reset` in future, I'm happy to add this back in then.

This PR needs a rebase it seems, so I'll take a look at that now, but other than that hopefully 🤞 this is ready to go. Let me know if you had any other questions.

LevelbossMike (approved):
:clap: thanks for looking into the reset functionality.

PR #13818, opened 2022-07-20
Update envoy metrics label extraction for peered clusters and listeners

kisunji:
### Description
Now that peered upstreams can generate envoy resources (https://github.com/hashicorp/consul/pull/13758), we need a way to disambiguate local from peered resources in our metrics. The key difference is that `datacenter` and `partition` will be replaced with `peer`, since in the context of peered resources `partition` is ambiguous (could refer to the partition in a remote cluster or one that exists locally). The partition and datacenter of the proxy will always be that of the source service.

Regexes were updated to make emitting `datacenter` and `partition` labels mutually exclusive with `peer` labels.

Listener filter names were updated to better match the existing regex.

Cluster names assigned to peered upstreams were updated to be synthesized from local peer name (it previously used the externally provided primary SNI, which contained the peer name from the other side of the peering). Integration tests were updated to assert for the new peer labels.

* Added test cases
* Updated integration tests

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/xds/listeners.go
+++ b/agent/xds/listeners.go
@@ -304,10 +304,9 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
 		if upstreamCfg != nil && upstreamCfg.HasLocalPortOrSocket() {
 			filterChain, err := s.makeUpstreamFilterChain(filterChainOpts{
 				clusterName: clusterName,
-				filterName: fmt.Sprintf("%s.%s.%s.%s",
+				filterName: fmt.Sprintf("%s.%s.%s",
 					upstreamCfg.DestinationName,
 					upstreamCfg.DestinationNamespace,
-					upstreamCfg.DestinationPartition,
 					upstreamCfg.DestinationPeer),
 				routeName:  uid.EnvoyID(),
 				protocol:   cfg.Protocol,
@@ -335,10 +334,9 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
 		filterChain, err := s.makeUpstreamFilterChain(filterChainOpts{
 			routeName:   uid.EnvoyID(),
 			clusterName: clusterName,
-			filterName: fmt.Sprintf("%s.%s.%s.%s",
+			filterName: fmt.Sprintf("%s.%s.%s",
 				uid.Name,
 				uid.NamespaceOrDefault(),
-				uid.PartitionOrDefault(),
 				uid.Peer),
 			protocol:   cfg.Protocol,
 			useRDS:     false,
--- a/agent/xds/listeners.go
+++ b/agent/xds/listeners.go
@@ -304,10 +304,15 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
 		if upstreamCfg != nil && upstreamCfg.HasLocalPortOrSocket() {
 			filterChain, err := s.makeUpstreamFilterChain(filterChainOpts{
 				clusterName: clusterName,
-				filterName:  uid.EnvoyID(),
-				routeName:   uid.EnvoyID(),
-				protocol:    cfg.Protocol,
-				useRDS:      false,
+				filterName: fmt.Sprintf("%s.%s.%s.%s",
+					upstreamCfg.DestinationName,
+					upstreamCfg.DestinationNamespace,
+					upstreamCfg.DestinationPartition,
+					upstreamCfg.DestinationPeer),
+				routeName:  uid.EnvoyID(),
+				protocol:   cfg.Protocol,
+				useRDS:     false,
+				statPrefix: "upstream_peered",
 			})
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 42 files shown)


agent/xds/listeners.go
```diff
@@ -304,10 +304,15 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
...
 				clusterName: clusterName,
-				filterName:  uid.EnvoyID(),
-				routeName:   uid.EnvoyID(),
-				protocol:    cfg.Protocol,
-				useRDS:      false,
+				filterName: fmt.Sprintf("%s.%s.%s.%s",
+					upstreamCfg.DestinationName,
+					upstreamCfg.DestinationNamespace,
+					upstreamCfg.DestinationPartition,
+					upstreamCfg.DestinationPeer),
```
kisunji, line 311: This is more inline with how filterNames were generated for discovery chain targets

agent/xds/listeners.go
```diff
@@ -304,10 +304,15 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
...
+				routeName:  uid.EnvoyID(),
+				protocol:   cfg.Protocol,
+				useRDS:     false,
+				statPrefix: "upstream_peered",
```
kisunji, line 315: Considered this the easiest way to handle the regex split between `datacenter` vs `peer`

command/connect/envoy/bootstrap_config.go
```diff
@@ -446,8 +457,8 @@ func resourceTagSpecifiers(omitDeprecatedTags bool) ([]string, error) {
 					reSegment, reSegment, reSegment, reSegment, reSegment, reSegment, reSegment, reSegment)},
 
 			{"consul.datacenter",
-				fmt.Sprintf(`^cluster\.((?:%s~)?(?:%s\.)?%s\.%s\.(?:%s\.)?(%s)\.%s\.%s\.consul\.)`,
-					reSegment, reSegment, reSegment, reSegment, reSegment, reSegment, reSegment, reSegment)},
+				fmt.Sprintf(`^cluster\.((?:%s~)?(?:%s\.)?%s\.%s\.(?:%s\.)?(%s)\.internal[^.]*\.%s\.consul\.)`,
```
kisunji, line 460: Change:
`[^.]+` -> `internal[^.]*`

command/connect/envoy/bootstrap_config.go
```diff
@@ -403,25 +409,30 @@ func resourceTagSpecifiers(omitDeprecatedTags bool) ([]string, error) {
 		// Upstream listener metrics are prefixed by consul.upstream
 		//
 		// Listener metric name format:
-		// <tcp|http>.upstream.<service>.<namespace>.<partition>.<datacenter>
+		// <tcp|http>.upstream<_peered>.<service>.<namespace>.<partition>.<datacenter>
```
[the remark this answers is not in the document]
kisunji: Oh actually, IIUC the outer capture group gets removed from the metric name so it doesn't matter

agent/xds/testdata/listeners/connect-proxy-with-peered-upstreams.latest.golden
```diff
@@ -17,7 +17,7 @@
               "name": "envoy.filters.network.tcp_proxy",
               "typedConfig": {
                 "@type": "type.googleapis.com/envoy.extensions.filters.network.tcp_proxy.v3.TcpProxy",
-                "statPrefix": "upstream.payments?peer=cloud",
+                "statPrefix": "upstream_peered.payments.default.default.cloud",
```
kisunji, line 20: the `default.default` is aligned with existing tproxy golden files; they appear in OSS as well

test/integration/connect/envoy/case-cross-peers-http-router/primary/verify.bats
```diff
@@ -51,7 +51,7 @@ load helpers
 }
 
 @test "s1 upstream should have healthy endpoints for s2" {
-  assert_upstream_has_endpoints_in_status 127.0.0.1:19000 s2.default.default.alpha-to-primary.external HEALTHY 1
+  assert_upstream_has_endpoints_in_status 127.0.0.1:19000 s2.default.primary-to-alpha.external HEALTHY 1
```
kisunji, line 54: Note that the new clusterName flips alpha-to-primary -> primary-to-alpha. This is correct since s1 is in the primary cluster

agent/xds/testdata/endpoints/transparent-proxy-with-peered-upstreams.latest.golden
```diff
@@ -3,7 +3,7 @@
   "resources": [
     {
       "@type": "type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment",
-      "clusterName": "api-a.default.default.cloud.external.1c053652-8512-4373-90cf-5a7f6263a994.consul",
+      "clusterName": "api-a.default.peer-a.external.1c053652-8512-4373-90cf-5a7f6263a994.consul",
```
kisunji, line 6: tproxy tests copy-pasted the PeerMeta from the connect proxy tests, so the clusters contained a reference to `cloud`. In the actual setup, the peer name is `peer-a` so this is more correct.

agent/xds/testdata/endpoints/transparent-proxy-with-peered-upstreams.latest.golden
```diff
@@ -47,7 +47,7 @@
     },
     {
       "@type": "type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment",
-      "clusterName": "no-endpoints?peer=peer-a",
+      "clusterName": "no-endpoints.default.peer-a.external.1c053652-8512-4373-90cf-5a7f6263a994.consul",
```
kisunji, line 50: Endpoints with no PrimarySNI defaulted to envoyID. Now there is only one static format for cluster names.

PR #13823, opened 2022-07-20
peering: use ShouldDial to validate peer role 

acpana:

### Description
We introduce a `PeerRole` enum for a `Peering` to start distinguishing between the side that creates the token (`Acceptor`) and the side that will dial the peer (`Dialer`).

This is useful when validating that peer names don't collide as an `Acceptor` and `Dialer`. The validation is added in this patch too

* `peering/service_test.go`, `api/peering_test.go`


agent/rpc/peering/service.go
```diff
@@ -171,6 +171,11 @@ func (s *Server) GenerateToken(
 		return resp, err
 	}
 
+	// validate that this peer name is not being used as a PeerRole_DIALER
+	if err = s.validatePeerRole(req.PeerName, req.Partition, pbpeering.PeerRole_ACCEPTOR); err != nil {
```
kisunji, line 175: This makes the exact same state store call as `getExistingOrCreateNewPeerID` function below. Maybe there's a chance to fetch once and pass a `Peering` around to both functions as required?

agent/rpc/peering/service.go
```diff
@@ -171,6 +171,11 @@ func (s *Server) GenerateToken(
...
+		return nil, err
+	}
+
```
kisunji, line 178: The `GetAgentCACertificates` and `GetServerAddresses` calls below can be moved a bit closer to where they're actually used. That way if these validations fail we don't make unnecessary backend calls.

agent/rpc/peering/service.go
acpana: I made a small func to get an existing peering

agent/rpc/peering/service.go
acpana: Moved those on top of this call

agent/rpc/peering/service.go
```diff
@@ -600,6 +627,32 @@ func (s *Server) getExistingOrCreateNewPeerID(peerName, partition string) (strin
...
+	return peering, nil
+}
+
+// validatePeerRole enforces the following rule for an existing peering:
+// - if a peering already exists, it can only be used as an acceptor or dialer
+//
+// We define a DIALER as a peering that has server addresses (or a peering that is created via the Establish endpoint)
+// Conversely, we define an ACCEPTOR as a peering that is created via the GenerateToken endpoint
+func (s *Server) validatePeerRole(peering *pbpeering.Peering, allowedToDial bool) error {
```
kisunji, line 648: This doesn't need to be a method off *Server

agent/rpc/peering/service.go
```diff
@@ -600,6 +627,32 @@ func (s *Server) getExistingOrCreateNewPeerID(peerName, partition string) (strin
...
+	if peering != nil && peering.ShouldDial() != allowedToDial {
+		return fmt.Errorf("cannot create peering with name: %q; there is already a peering with that name and different role", peering.Name)
```
kisunji, line 650: Can we reword the error message and function name? "role" is no longer a used term here

agent/rpc/peering/service.go
```diff
@@ -181,13 +181,29 @@ func (s *Server) GenerateToken(
...
+		id = peeringOrNil.ID
+	} else {
+		id, err = lib.GenerateUUID(s.Backend.CheckPeeringUUID)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	canRetry := true
+RETRY_ONCE:
```
kisunji, line 205: The fetching and ID validation must be done under the RETRY_ONCE block

agent/rpc/peering/service.go
```diff
@@ -581,16 +612,12 @@ func (s *Server) TrustBundleListByService(ctx context.Context, req *pbpeering.Tr
 }
 
 func (s *Server) getExistingOrCreateNewPeerID(peerName, partition string) (string, error) {
```
kisunji, line 614: Is this function used any more?

agent/rpc/peering/service.go
acpana: Why? 🤔 

I think I see that if we don't do that then we could use the wrong ID right?

Just wanted to double check.

agent/rpc/peering/service.go
acpana: yeah this wraps around the `getPeeringOrNil` for the `PeeringWrite` endpoint IIRC

agent/rpc/peering/service.go
acpana: what would be a better phrase here?

agent/rpc/peering/service.go
kisunji: If you read the comments below it intends for a retry to re-read from the state store and get the existing peer ID.

Otherwise you're retrying a failed write request once again with no changes in input

kisunji: Actually, maybe this is reason enough to keep the ID generation and validation seprate with their own calls to the state store 🤔 Wouldn't want to couple the validation with the retry logic

agent/rpc/peering/service.go
kisunji: Assuming you're going to end up adding more validations here the function could simply be `validatePeer`.

For error message maybe a branch to say "there is an existing peering expecting to be dialed" vs "there is already an established peering"?

agent/rpc/peering/service.go
acpana: Yeah we can use the old func as is then 🤔

agent/rpc/peering/service.go
acpana: done

agent/rpc/peering/service.go
acpana: thanks for the suggestion! added it

kisunji (approved):
🎉

PR #13834, opened 2022-07-21
ui: Update peerings empty state copy

LevelbossMike:
### Description
Updates the empty state copy of the peerings page to the latest copy. It will now also display differently based on the fact if the user is using OSS or not.

* [ ] updated test coverage
* [x] not a security concern

cc @johncowen


johncowen (approved):
Good stuff! Thanks for adding the p tags here also! 🙈 🎉

PR #13835, opened 2022-07-21
Add Cluster Peering Failover Support to Prepared Queries

erichaberkorn:
### Description
Add cluster peering failover support to prepared queries.

* [Unit test validations](https://github.com/hashicorp/consul/compare/prepared-query-peering-failover?expand=1#diff-9865c52cf7cb8ef7edb4f8980519eef139049c970208a1581eb38bb42f708818R91-R105)
* [Unit test peering failover](https://github.com/hashicorp/consul/compare/prepared-query-peering-failover?expand=1#diff-9865c52cf7cb8ef7edb4f8980519eef139049c970208a1581eb38bb42f708818R3301)
* [Integration test peering failover](https://github.com/hashicorp/consul/compare/prepared-query-peering-failover?expand=1#diff-9865c52cf7cb8ef7edb4f8980519eef139049c970208a1581eb38bb42f708818R1468)

### PR Checklist

* [X] updated test coverage
* [ ] external facing docs updated - I'll do this in a follow-up PR
* [X] not a security concern

### DIFF
--- a/agent/consul/prepared_query_endpoint.go
+++ b/agent/consul/prepared_query_endpoint.go
@@ -573,7 +573,7 @@ func (p *PreparedQuery) execute(query *structs.PreparedQuery,
 	// Stamp the result with its this datacenter or peer.
 	if peerName := query.Service.PeerName; peerName != "" {
 		reply.PeerName = peerName
-		reply.Datacenter = peerName
+		reply.Datacenter = ""
 	} else {
 		reply.Datacenter = p.srv.config.Datacenter
 	}
--- a/agent/consul/prepared_query_endpoint.go
+++ b/agent/consul/prepared_query_endpoint.go
@@ -187,9 +187,14 @@ func parseService(svc *structs.ServiceQuery) error {
 		return fmt.Errorf("Must provide a Service name to query")
 	}
 
+	failover := svc.Failover
 	// NearestN can be 0 which means "don't fail over by RTT".
-	if svc.Failover.NearestN < 0 {
+	if failover.NearestN < 0 {
 		return fmt.Errorf("Bad NearestN '%d', must be >= 0", svc.Failover.NearestN)
+	}
+
+	if (failover.NearestN != 0 || len(failover.Datacenters) != 0) && len(failover.Targets) != 0 {
+		return fmt.Errorf("Targets cannot be populated with NearestN or Datacenters")
 	}
 
 	// Make sure the metadata filters are valid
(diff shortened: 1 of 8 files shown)


agent/consul/prepared_query_endpoint.go
```diff
@@ -565,8 +570,13 @@ func (p *PreparedQuery) execute(query *structs.PreparedQuery,
 	reply.Nodes = nodes
 	reply.DNS = query.DNS
 
-	// Stamp the result for this datacenter.
-	reply.Datacenter = p.srv.config.Datacenter
+	// Stamp the result with its this datacenter or peer.
+	if peerName := query.Service.PeerName; peerName != "" {
+		reply.PeerName = peerName
+		reply.Datacenter = peerName
```
erichaberkorn, line 576: I copied this naming convention from the results of `state.CheckServiceNodes` but maybe we want something different?

agent/consul/prepared_query_endpoint_test.go
```diff
@@ -1456,6 +1489,70 @@ func TestPreparedQuery_Execute(t *testing.T) {
 	// check for RPC forwarding
 	testrpc.WaitForLeader(t, s1.RPC, "dc1", testrpc.WithToken("root"))
 	testrpc.WaitForLeader(t, s1.RPC, "dc2", testrpc.WithToken("root"))
+	testrpc.WaitForLeader(t, s3.RPC, "dc3")
+
+	acceptingPeerName := "my-peer-accepting-server"
+	dialingPeerName := "my-peer-dialing-server"
+
+	// Set up peering between dc1 (dailing) and dc3 (accepting) and export the foo service
```
erichaberkorn, line 1497: This test is massive. I added on because I would have needed to copy over a bunch of boilerplate code 🙈

agent/consul/prepared_query_endpoint.go
freddygv, line 576: I lean toward making Datacenter empty here when the results are from a peer, would be misleading otherwise since a peer doesn't necessarily represent a datacenter

agent/consul/prepared_query_endpoint_test.go
freddygv: That's OK, ideally this will be the last change to prepared queries :P

agent/consul/prepared_query_endpoint_test.go
```diff
@@ -1576,6 +1673,17 @@ func TestPreparedQuery_Execute(t *testing.T) {
 		assert.True(t, reply.QueryMeta.KnownLeader)
 	}
 
+	expectFailoverPeerNodes := func(t *testing.T, query *structs.PreparedQueryRequest, reply *structs.PreparedQueryExecuteResponse, n int) {
+		t.Helper()
+		assert.Len(t, reply.Nodes, n)
+		assert.Equal(t, acceptingPeerName, reply.Datacenter)
```
freddygv, line 1679: This would need to be removed if reply.Datacenter is unset

freddygv (commented):
This looks great, awesome that you were able to get it turned around quickly. 

I looked through the whole thing and left some small requests below.

freddygv (approved):

PR #13836, opened 2022-07-21
ui: Remove peering detail page

johncowen:
### Description
This PR removes all the peering detail views, peer rows are now not clickable, but peers are still deletable.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

/cc @LevelbossMike


LevelbossMike (approved):
:+1:

PR #13840, opened 2022-07-21
ui: peered services only show instance- and tags-tabs 

LevelbossMike:
### Description
Peered service-instances will now only surface the `Instances` and `Tags` Tab on the `services.show`-view.

Screenshot 2022-07-21 at 16 51 05

#### Notes
This change made the `show-with-slashes`-feature test fail. That is because the first service that we load from the mock-api by default, when the peering feature is on, will be peered. To "fix" the failing test, I updated the URL that we are checking against in the test. A peered service will not show the `topology`-tab anymore and show the `instances`-tab instead. Because the URL isn't relevant for the test, this feels fine and was easier than to change the mocked service in the test.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

cc @johncowen

### DIFF
--- a/ui/packages/consul-ui/app/templates/dc/services/show.hbs
+++ b/ui/packages/consul-ui/app/templates/dc/services/show.hbs
@@ -117,12 +117,12 @@ as |items item dc|}}
       {{/if}}
   {{#let
   (hash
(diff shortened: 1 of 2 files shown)


johncowen (approved):
Looks good to me, thanks for explaining about the test, changes there make sense to me also 🎉

PR #13844, opened 2022-07-21
Update generate token endpoint to take external addresses

ndhanushkodi:
Update generate token endpoint (rpc, http, and api module)

If ServerExternalAddresses are set, it will override any addresses gotten from the "consul" service, and be used in the token instead, and dialed by the dialer. This allows for setting up a load balancer for example, in front of the consul servers.

(Addresses comments from https://github.com/hashicorp/consul-enterprise/pull/2363, and 2363 will get closed.)

### PR Checklist

* [x] updated test coverage
* [ ] external facing docs updated -future PR
* [x] not a security concern


freddygv (approved):
🚀

PR #13847, opened 2022-07-21
Fix goroutine leaks in proxycfg when using ingress gateway

kyhavlov:
This PR fixes 2 separate goroutine leaks happening in `agent/proxycfg` that occur when registering an ingress gateway:
- Whenever the list of gateway services was updated, a new goroutine was spawned via `cache.Notify()` for every service (even ones that were already being watched) because we were wrongly checking the `WatchedDiscoveryChains` in `snap.ConnectProxy` instead of `snap.IngressGateway`.
- When cleaning up services removed from the gateway services list, we were only canceling the watches for discovery chains and not their associated upstreams as well.

### DIFF
--- a/.changelog/13847.txt
+++ b/.changelog/13847.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+connect: Fixed a goroutine/memory leak that would occur when using the ingress gateway.
+```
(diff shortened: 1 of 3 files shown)


agent/submatview/store.go
```diff
@@ -221,7 +221,7 @@ func (s *Store) readEntry(req Request) (string, Materializer, error) {
 		return "", nil, err
 	}
 
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(ctx)
```
kyhavlov, line 224: For some reason, this change is breaking `TestServiceMultipleTagsWatch` in `api/watch`. Still investigating why.

boxofrad (approved):
Great catches! 👏🏻

kisunji (approved):
👏 probably needs a bugfix changelog entry?

PR #13849, opened 2022-07-21
peering: read endpoints can now return failing status

lkysow:
When the stream exits due to an error, e.g. the heartbeat failed or the connection quit, track this in the tracker. When the Read or List RPCs are called, use the tracker to set the status as failing.

Changes:
1. tracker: Differentiate a connection ending gracefully from a connection ending due to error by adding a field `DisconnectErrorMessage` that will be set to an error string if the connection ended due to an error.
2. tracker: rename `Disconnected()` => `DisconnectedGracefully()` and add `DisconnectedDueToErr()`
3. stream_resources: wrap `HandleStream()` to correctly set the tracker status based on why the connection was ended
4. add `TrackReceiveHeartbeat()`. This isn't used anywhere yet but can be used for better status info later.
5. rename `TrackReceiveSuccess() => TrackReceiveResourceSuccess()` to differentiate between receiving a replicated resource and a heartbeat

### DIFF
--- a/agent/grpc-external/services/peerstream/stream_test.go
+++ b/agent/grpc-external/services/peerstream/stream_test.go
@@ -423,7 +423,7 @@ func TestStreamResources_Server_StreamTracker(t *testing.T) {
 		})
 	})
 
-	var lastRecvSuccess time.Time
+	var lastRecvResourceSuccess time.Time
 
 	testutil.RunStep(t, "response applied locally", func(t *testing.T) {
 		resp := &pbpeerstream.ReplicationMessage{
@@ -437,7 +437,7 @@ func TestStreamResources_Server_StreamTracker(t *testing.T) {
 				},
 			},
 		}
-		lastRecvSuccess = it.FutureNow(1)
+		lastRecvResourceSuccess = it.FutureNow(1)
 		err := client.Send(resp)
 		require.NoError(t, err)
 
@@ -475,11 +475,11 @@ func TestStreamResources_Server_StreamTracker(t *testing.T) {
 		api := structs.NewServiceName("api", nil)
 
 		expect := Status{
-			Connected:          true,
-			LastAck:            lastSendSuccess,
-			LastNack:           lastNack,
-			LastNackMessage:    lastNackMsg,
-			LastReceiveSuccess: lastRecvSuccess,
+			Connected:                  true,
+			LastAck:                    lastSendSuccess,
+			LastNack:                   lastNack,
+			LastNackMessage:            lastNackMsg,
+			LastReceiveResourceSuccess: lastRecvResourceSuccess,
 			ImportedServices: map[string]struct{}{
 				api.String(): {},
 			},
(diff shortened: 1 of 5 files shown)


agent/rpc/peering/service.go
```diff
@@ -394,6 +394,8 @@ func (s *Server) reconcilePeering(peering *pbpeering.Peering) *pbpeering.Peering
 		// reconcile pbpeering.PeeringState_Active
 		if streamState.Connected {
 			cp.State = pbpeering.PeeringState_ACTIVE
```
lkysow, line 396: here is where this results in the peering endpoints showing the state as failing

agent/grpc-external/services/peerstream/stream_test.go
```diff
@@ -868,8 +869,8 @@ func TestStreamResources_Server_CARootUpdates(t *testing.T) {
 	})
 }
 
-// Test that when the client doesn't send a heartbeat in time, the stream is terminated.
-func TestStreamResources_Server_TerminatesOnHeartbeatTimeout(t *testing.T) {
+// Test that when the client doesn't send a heartbeat in time, the stream is disconnected.
+func TestStreamResources_Server_DisconnectsOnHeartbeatTimeout(t *testing.T) {
```
lkysow, line 873: renamed this to disambiguate between Terminated messages and disconnects on errors

agent/grpc-external/services/peerstream/stream_tracker.go
```diff
@@ -235,10 +248,19 @@ func (s *MutableStatus) TrackConnected() {
 	s.mu.Unlock()
 }
 
-func (s *MutableStatus) TrackDisconnected() {
+func (s *MutableStatus) TrackDisconnectedGracefully() {
+	s.mu.Lock()
+	s.Connected = false
+	s.DisconnectTime = s.timeNow().UTC()
+	s.DisconnectErrorMessage = ""
```
kisunji, line 255: Should `TrackConnected` also blank out `DisconnectErrorMessage`?

lkysow: Yes! Great catch.

agent/grpc-external/services/peerstream/stream_test.go
```diff
@@ -423,7 +423,7 @@ func TestStreamResources_Server_StreamTracker(t *testing.T) {
 		})
```
lkysow, line 423: Changes to these tests:
1. rename some `lastRecvSuccess` to `lastRecvResourceSuccess` to match new method name
2. also test the heartbeat receive tracking

agent/grpc-external/services/peerstream/stream_tracker.go
```diff
@@ -75,13 +75,23 @@ func (t *Tracker) connectedLocked(id string) (*MutableStatus, error) {
...
 	t.mu.Lock()
 	defer t.mu.Unlock()
 
 	if status, ok := t.streams[id]; ok {
-		status.TrackDisconnected()
+		status.TrackDisconnectedGracefully()
+	}
+}
+
+// todo
```
kisunji, line 88: Missed todo?

agent/grpc-external/services/peerstream/stream_tracker.go
```diff
@@ -153,8 +167,11 @@ type Status struct {
 	// LastSendErrorMessage tracks the last error message when sending into the stream.
 	LastSendErrorMessage string
 
-	// LastReceiveSuccess tracks the time we last successfully stored a resource replicated FROM the peer.
-	LastReceiveSuccess time.Time
+	// LastReceiveHeartbeat tracks when we last received a heartbeat from our peer.
+	LastReceiveHeartbeat time.Time
+
+	// LastReceiveResourceSuccess tracks the time we last successfully stored a resource replicated FROM the peer.
+	LastReceiveResourceSuccess time.Time
```
kisunji, line 174: Were these renamed to include `Resource` due to separating Heartbeat vs Resource?

We may want to align `LastReceiveError` and `LastReceiveErrorMessage` to be qualified with `Resource` as well. And I personally wouldn't mind `Receive` shortened to `Recv` since the names are getting long 😆

agent/grpc-external/services/peerstream/stream_test.go
```diff
@@ -905,10 +940,13 @@ func TestStreamResources_Server_TerminatesOnHeartbeatTimeout(t *testing.T) {
 	})
 
 	testutil.RunStep(t, "stream is disconnected due to heartbeat timeout", func(t *testing.T) {
+		disconnectTime := it.FutureNow(1)
 		retry.Run(t, func(r *retry.R) {
 			status, ok := srv.StreamStatus(peerID)
 			require.True(r, ok)
-			require.False(r, status.Connected)
+			require.Equal(r, false, status.Connected)
```
kisunji, line 947: nit: Why `require.False -> require.Equal` here?

kisunji (approved):
minor comments but LGTM

agent/grpc-external/services/peerstream/stream_tracker.go
lkysow: LastReceiveError and LastReceiveErrorMessage are still generic for both resources and heartbeats as well as other errors such as the stream being disconnected.

I can do the shortening for sure.

agent/grpc-external/services/peerstream/stream_test.go
lkysow: Can change for sure. I used require.Equal because I was doing a bunch of field comparisons so just had them all grouped as require.Equals.

agent/grpc-external/services/peerstream/stream_tracker.go
lkysow: I'll merge this as is, got a PR for the shortening and then I can come back to separating out the lastreceiveerror etc into LastReceiveError (for generic receive errors) and then LastReceiveResourceError (for replicated resource receive errors). Could also wait on that change until we figure out how we're going to surface these errors.

PR #13851, opened 2022-07-22
peering: prevent peering in same partition

acpana:
### Description
We want to prevent a `peering` from being created in the same partition (ENT) or cluster (OSS).

We do this by looking for an intersection (any) between the server addresses encoded in the token and those of the current cluster's.


freddygv:
Hey Alex, server addresses are not specific to a partition, they're specific to a datacenter. Checking whether server addresses match would prevent users from peering two different partitions in the same Consul datacenter.

Another wrinkle is that addresses embedded in a token may not be the same ones that come from `GetServerAddresses()` now that this PR is merged: https://github.com/hashicorp/consul/pull/13844

Could the peer ID help instead?

acpana:
thanks for the guidance @freddygv ! how is this?

mkeeler (approved):
LGTM.

_Food for thought_

Should we allow peering with any partition within the same cluster. We can already export services across partitions and it seems less than desirable to duplicate those within memdb.

kisunji (changes requested):
I think you accidentally deleted some lines with the rebase

agent/rpc/peering/service.go
```diff
@@ -399,6 +402,21 @@ func (s *Server) Establish(
 	return resp, nil
 }
 
+// validatePeeringInPartition makes sure that we don't create a peering in the same partition. We do so by checking
+// to see if the current server's server addresses slice has any intersection with the slice from the token.
```
kisunji, line 406: comment should be updated

agent/rpc/peering/service.go
```diff
@@ -399,6 +402,21 @@ func (s *Server) Establish(
...
+func (s *Server) validatePeeringInPartition(remotePeerID, partition string) error {
+	_, peering, err := s.Backend.Store().PeeringReadByID(nil, remotePeerID)
+	if err != nil {
+		return fmt.Errorf("cannot validate peering in partition: %w", err)
```
kisunji, line 410: should just return err instead of wrapping (misleading error message)

acpana: I think it's worth saying the "why" the `PeeringReadByID` is made in the first place.

As a user here, the UX of the "why" (eg `cannot validate peering in partition: error_msg`) let's me know that there is a validation issue with my  request.

kisunji: In that case, a general rule of thumb I like is:
Wrapping error messages should contextualize where the error came from (and avoid self-referencing their containing functions).

In this case, the failure is `PeeringReadByID`, so it's misleading to prematurely wrap the error.
```suggestion
		return fmt.Errorf("cannot read peering by ID: %w", err)
```

The actual context should be wrapped by the caller
```go
	if err := s.validatePeeringInPartition(tok.PeerID, req.Partition); err != nil {
		return nil, fmt.Errorf("cannot validate peering in partition: %w", err)
	}
```

agent/rpc/peering/service.go
```diff
@@ -399,6 +402,22 @@ func (s *Server) Establish(
...
+// the remotePeerID from the PeeringToken and looking up for a peering in the partition. If there is one and the
+// request partition is the same, then we are attempting to peer within the partition, which we shouldn't.
+func (s *Server) validatePeeringInPartition(remotePeerID, partition string) error {
+	_, peering, err := s.Backend.Store().PeeringReadByID(nil, remotePeerID)
+	if err != nil {
+		return fmt.Errorf("cannot validate peering in partition: %w", err)
+	}
+
+	if peering != nil && peering.Partition == partition {
+		return fmt.Errorf("cannot create a peering within the same partition (ENT) or cluster (OSS)")
```
kisunji, line 415: Can we better describe what the error is? This applies to the PR body as well. It's unclear what the error condition is

kisunji: nvm I re-read a bunch of times and it makes sense as-is!

kisunji (approved):
minor comments but LGTM

agent/rpc/peering/service.go
```diff
@@ -354,6 +354,9 @@ func (s *Server) Establish(
 	if err = validatePeer(peeringOrNil, true); err != nil {
 		return nil, err
 	}
+	if err := s.validatePeeringInPartition(tok.PeerID, req.Partition); err != nil {
```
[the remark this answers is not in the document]
acpana: added a comment right above 🥇

PR #13854, opened 2022-07-22
ui: Improve peered service empty downstreams message

LevelbossMike:
### Description
Improves the empty downstreams message in service-topology to only mention "imported services" when the peering feature is enabled.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


evrowe (approved):
Change looks good! 👍

PR #13855, opened 2022-07-22
Add peering ACL docs

freddygv:
This PR adds documentation for the ACL rule that controls access to the peering endpoints.

There were also some minor updates to the `Mesh Rules` section:
* Fixed the header sizing
* Clarified the scope, since it cannot be specified inside of a namespace rule


website/content/docs/security/acl/acl-rules.mdx
```diff
@@ -600,6 +611,30 @@ operator = "read"
 
 </CodeTabs>
 
+## Peering Rules
+The `peering` resource controls access to cluster peerings in the [Cluster Peering API](/api-docs/peering).
```
jkirschner-hashicorp, line 615: @freddygv : do any built-in policies have `peering` resource permissions?

For example, I know `namespace-management` wouldn't. Is there an equivalent built-in for partitions that would?

freddygv: It's only embedded in the global management policy

There's no built-in partition management policy at the moment

PR #13867, opened 2022-07-22
peering: add config to enable/disable peering

lkysow:
Add config:

```
peering {
  enabled = true
}
```

Defaults to true. When disabled:
1. All peering RPC endpoints will return an error
2. Leader won't start its peering establishment goroutines
3. Leader won't start its peering deletion goroutines

   `agent/config/config.go`.
   `agent/config/runtime.go`.
   translate.
   `agent/config/testdata/full-config.*`, which should cause the `TestLoad_FullConfig` test to fail.
 - [ ] **If** your new config field needed some validation as it's only valid in
      - [x] Add that to `DefaultSource` in `agent/config/default.go`.
 - [ ] **If** your config should take effect on a reload/HUP.
      - [ ] Add necessary code to trigger a safe (locked or atomic) update to
        more of the following places:
        `TestAgent_reloadConfig*`.

## Adding a Simple Config Field for Servers
 - [x] Do all of the steps in [Adding a Simple Config
 - [x] Add code to set the values from the `RuntimeConfig` in the confusingly
   named `newConsulConfig` method in `agent/agent.go`
 - [ ] **If needed**, add a test to `agent_test.go` if there is some non-trivial
 - [ ] **If** your config should take effect on a reload/HUP
      - [ ] Add necessary code to `ReloadConfig` in `agent/consul/server.go` this
        updated.
       - [ ] Add a new test or a new assertion to `TestServer_ReloadConfig`

### DIFF
--- a/agent/consul/leader_peering_test.go
+++ b/agent/consul/leader_peering_test.go
@@ -1036,3 +1036,89 @@ func TestLeader_PeeringMetrics_emitPeeringMetrics(t *testing.T) {
 		require.Equal(r, float32(2), metric2.Value) // for d, e services
 	})
 }
+
+// Test that the leader doesn't start its peering deletion routing when
+// peering is disabled.
+func TestLeader_Peering_NoDeletionWhenPeeringDisabled(t *testing.T) {
+	if testing.Short() {
+		t.Skip("too slow for testing.Short")
+	}
+
+	_, s1 := testServerWithConfig(t, func(c *Config) {
+		c.NodeName = "s1.dc1"
+		c.Datacenter = "dc1"
+		c.TLSConfig.Domain = "consul"
+		c.PeeringEnabled = false
+	})
+	testrpc.WaitForLeader(t, s1.RPC, "dc1")
+
+	var (
+		peerID   = "cc56f0b8-3885-4e78-8d7b-614a0c45712d"
+		peerName = "my-peer-s2"
+		lastIdx  = uint64(0)
+	)
+
+	// Simulate a peering initiation event by writing a peering to the state store.
+	lastIdx++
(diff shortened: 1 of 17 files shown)


agent/consul/leader_peering_test.go
```diff
@@ -1036,3 +1036,89 @@ func TestLeader_PeeringMetrics_emitPeeringMetrics(t *testing.T) {
 		require.Equal(r, float32(2), metric2.Value) // for d, e services
 	})
 }
+
```
lkysow, line 1039: These tests take 7s each because they wait for nothing to happen and if peering was enabled it takes about 7s for something to happen. It is important that the behaviour is tested though so probably worth the time.

agent/consul/leader_peering_test.go
```diff
@@ -1036,3 +1036,89 @@ func TestLeader_PeeringMetrics_emitPeeringMetrics(t *testing.T) {
...
+	// Mark the peering for deletion to trigger the termination sequence.
+	lastIdx++
+	require.NoError(t, s1.fsm.State().PeeringWrite(lastIdx, &pbpeering.Peering{
+		ID:        peerID,
+		Name:      peerName,
+		DeletedAt: structs.TimeToProto(time.Now()),
+	}))
+
+	// The leader routine shouldn't be running so the peering should never get deleted.
+	require.Never(t, func() bool {
```
freddygv, line 1077: TIL about `require.Never`

freddygv (approved):
🚀

agent/consul/leader_peering_test.go
lkysow: I found it in waypoint :)

PR #13874, opened 2022-07-25
Allow hostname based addresses in peering token

ndhanushkodi:
When supplying an external address to put into the peering token, the external address could be a hostname of a load balancer, which for example is the default in EKS. We'd want to allow hostname based addresses in the peering token in this case. 

I was able to test a hostname based load balancer with this change and it worked as the dialer does not require an IP address.

### PR Checklist

* [x] updated test coverage
* [ ] external facing docs updated - docs change for all of the expose servers feature to come in future
* [x] not a security concern

### DIFF
--- a/agent/rpc/peering/testing.go
+++ b/agent/rpc/peering/testing.go
@@ -31,6 +31,7 @@ not valid
 `
 
 var validAddress = "1.2.3.4:80"
+var validHostnameAddress = "foo.bar.baz:80"
 
 var validServerName = "server.consul"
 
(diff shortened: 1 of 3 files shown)


kisunji:
Confirming that the grpc call will default to DNS: https://github.com/grpc/grpc/blob/master/doc/naming.md#name-syntax=

Called here (`ServerAddresses` written to `PeerServerAddresses` written to ring buffer): https://github.com/hashicorp/consul/blob/927cee692b5743e4a309292e11e5238876f6bb76/agent/consul/leader_peering.go#L314=

kisunji (approved):

PR #13887, opened 2022-07-25
block PeerName register requests

acpana:

### description
 
We probably want to block the `Catalog.Register` endpoint from taking in requests with any `PeerName` (on nodes, services, checks).

### proposal

The approach taken here is to add a config value that is non user configurable. But our test agent/servers can turn it on.


agent/testagent.go
```diff
@@ -182,6 +182,12 @@ func (a *TestAgent) Start(t *testing.T) error {
 
 	portsConfig := randomPortsSource(t, a.UseTLS)
 
+	//a.Overrides = `
```
acpana, line 185: We could turn this on for all agents but I'd recommend turning it off by default and just enabling it for the tests that need it.

mkeeler (commented):
I understand the need to for the new config to bypass the restriction in the case of tests. However, it makes me think we have gotten the interfaces and boundaries between components wrong if we need to do this.

Do you know how many/which tests get broken if we just unconditionally restrict this? I think there are a few different ways tests should inject peered services.

1. Within the `agent/consul` package those tests can insert directly into the state store or they can call `raftApply` on the server directly.
2. Outside of `agent/consul` unit tests should use mocks that don't require real data injection  (or an actual server)
3. Outside of `agent/consul` integration tests should set up multiple agents and let the peering replication stream get the data injected instead of doing it directly.

With that approach, we shouldn't need any configurability of allowing/disallowing setting the peer name through the RPC endpoint.

I realize that we may have too many tests to make that feasible for this PR which is why I would be curious which/how many tests it breaks to just always enforce the restriction.

agent/catalog_endpoint_test.go
```diff
@@ -20,6 +20,25 @@ import (
 	"github.com/hashicorp/consul/testrpc"
 )
 
+func TestCatalogRegister_DenyPeeringRegistration(t *testing.T) {
```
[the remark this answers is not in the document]
acpana: So I wanted to test turning off the setting at the agent level but also add more cases at the `consul` level; what do you think of the current approach?

agent/consul/catalog_endpoint_test.go
```diff
@@ -2765,6 +2765,109 @@ node_prefix "" {
...
+		},
+	}
+
+	type testcase struct {
+		name      string
+		reqCopyFn func(arg *structs.RegisterRequest) structs.RegisterRequest
+	}
+
+	/*
+		NOTE that there is enforcement for one and only PeerName to be present in the state store.
```
kisunji, line 2807:
```
NOTE that there is enforcement for one and only PeerName to be present in the state store.
```
Could you clarify what this means? Unsure what the implications are

agent/config/default.go
```diff
@@ -181,6 +181,11 @@ func DevSource() Source {
 		ports = {
 			grpc = 8502
 		}
+		
+		peering {
+			# We use peer registration for various testing
+			test_allow_peer_registrations = false
+		}
```
kisunji, line 188: We should be declaring this at `NonUserSource` only, right? Otherwise you have duplicate entries eventually merged to the final config

mkeeler, line 188:
```suggestion
```

agent/consul/catalog_endpoint_test.go
mkeeler: I _think_ this is mostly applicable to the "peer names everywhere" test case that sets multiple different peer names at each point in the register request. That register request would be deemed invalid and any raftApply for it would fail.

mkeeler (approved):

agent/config/default.go
mkeeler: To @kisunji point in his other comment, we shouldn't need to have this be disabled here in the DevSource as the NonUserSource will be applied anyways.

mkeeler (changes requested):
I took a quick peak at the CI failures and they look legit. Mostly due to not being able to register services with peer names.

agent/testagent.go
mkeeler: I think not having this is the cause of all the existing CI failures. However if you uncomment this then your new tests regarding checking the enforcement behavior of not allowing the PeerName field to be set would then fail.

I generally agree that the best course of action would be to update tests that do the nasty things to opt into the behavior.

acpana:
We discussed internally and we DO NOT WANT to include this change in `1.13`; this change will go out in `1.13.1` or later. No backport labels should be applied.

agent/config/default.go
acpana: Oh yes, great catch!

agent/consul/catalog_endpoint_test.go
acpana: So yeah matt's spot on above; it refers to the validations we do at the store level ... but that knowledge comes form an open box thinking; I removed the comment since it's not that applicable to what we are testing. We want to block any mention of `PeerName`, regardless of the values of the field.

agent/testagent.go
acpana: I think as of now, all tests that need this config option on have it so I should remove this commented block

mkeeler (approved):

PR #13898, opened 2022-07-25
fix: ipv4 destination dns resolution

DanStough:
### Description
Currently, there are issues with Destination hostnames that resolve to IPV6 addresses.  The terminating gateway can resolve the names into addresses, but when dialing the IPV6 address, it reports the connection unreachable. 

We still need to consider if hardcoding this to IPV4 is a good long-term decision.

* try contacting `api.myip.com` using Destinations.

### PR Checklist

* [X] updated test coverage
* [ ] not a security concern


dhiaayachi (approved):
🚀

Issue #8254, opened 2020-07-07
[connect] race-condition registering multiple proxies, duplicate public_listener ports are assigned


dekimsey:
#### Overview of the Issue

In our Consul Connect rollout we found that a subset of our proxies would look like they started (running, reporting healthy) but would not have any `public_listeners` listed (detected via envoy admin api, `http//...:19000/listeners`). We found the envoy proxies were failing to bind to their ports, `Address already in use`. Tracing it we found another legitimate proxy registered to the same agent was running on that port.

Additionally, due to the basic TCP check, Consul was unawares the wrong proxy was responding on the port.

#### Reproduction Steps

Walking through the agent sidecar registration method, [AgentRegisterService](https://github.com/hashicorp/consul/blob/master/agent/agent_endpoint.go#L847), we see roughly the following logic at play:

```
// AgentRegisterService()
// ...
// See if we have a sidecar to register too (`sidecarServiceFromNodeService`)
  -> // Allocate port if needed (min and max inclusive).  <-- duplicate port is chosen here.
// Add the service (`AddService` or `AddServiceAndReplaceChecks`)
  -> // Lock
  -> // addServiceLocked
  -> // Unlock
// ...
```

The port allocation occurs in ascending order (used to be random, making this more likely to happen now) and outside of the locking used to commit the service to the state. This makes it easy for two API calls to obtain the same assigned public_listening port.

So a couple of options here:
  validateService() could test for proxy addr:port uniqueness and reject, perhaps requiring sidecarServiceFromNodeService() to reselect a port.
  Port selection/assignment probably should be delegated to AddService so that it happens during the locked code.

I'm thinking the best solution would be to have AddService (perhaps via a callback to do so) to uniquely assign the connect proxy's port. Unfortunately right now, AddService isn't connect aware, but I don't see how to fix this without giving it some hint that this registration needs to happen during the stateLock.

Steps to reproduce this issue, e.g.:

1. Register many Connect-enabled services to Consul API (`/v1/agent/service/register`) in parallel using default connect sidecar settings:
```
{"name": "foo", "connect": {"sidecar_service": {}}}
```

Local agents will show duplicate Port assignments for the connect proxies:
```
$ curl -s http://localhost:8500/v1/agent/services | python -m json.tool | grep '"Port": 21' -A5
        "Port": 21002,
        "Proxy": {
            "DestinationServiceID": "customer-merge-service",
            "DestinationServiceName": "customer-merge-service",
            "Expose": {},
[...]

PR #13906, opened 2022-07-26
Separate port and socket path validation for local agent

skpratt:
### Description
[PR #12881](https://github.com/hashicorp/consul/pull/12881) Introduced a requirement for either port or socket path to be provided in all requests. Due to [a race in automatic port assignment](https://github.com/hashicorp/consul/issues/8254) within the agent, automatic assignment needs [to be moved from the request-building stage to the locked registration stage post-validation](https://github.com/hashicorp/consul/pull/13928). One solution (this PR) is to exempt the port requirement from request validation when the request is coming from an agent that is trying to do auto-assignment.

* Manually registering a service from the agent still works with and without a port specified.
* Existing unit tests seem to cover many cases, and pass without need for changes.

### Links
[PR #12881](https://github.com/hashicorp/consul/pull/12881)
[Port assignment issue](https://github.com/hashicorp/consul/issues/8254)
[Follow-on PR that will need this](https://github.com/hashicorp/consul/pull/13928)

### PR Checklist

* [x] updated test coverage
* [N/A] external facing docs updated

### DIFF
--- a/agent/structs/structs.go
+++ b/agent/structs/structs.go
@@ -1411,6 +1411,27 @@ func (s *NodeService) IsGateway() bool {
 func (s *NodeService) Validate() error {
 	var result error
 
+	if s.Kind == ServiceKindConnectProxy {
+		if s.Port == 0 && s.SocketPath == "" {
+			result = multierror.Append(result, fmt.Errorf("Port or SocketPath must be set for a %s", s.Kind))
+		}
+	}
+
+	commonValidation := s.ValidateForAgent()
+	if commonValidation != nil {
+		result = multierror.Append(result, commonValidation)
+	}
+
+	return result
+}
+
+// ValidateForAgent does a subset validation, with the assumption that a local agent can assist with missing values.
+//
+// I.e. in the catalog case, a local agent cannot be assumed to facilitate auto-assignment of port or socket path,
+// so additional checks are needed.
+func (s *NodeService) ValidateForAgent() error {
+	var result error
+
 	// TODO(partitions): remember to double check that this doesn't cross partition boundaries
 
 	// ConnectProxy validation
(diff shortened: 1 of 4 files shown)


agent/structs/structs_test.go
```diff
@@ -1157,6 +1157,16 @@ func TestStructs_NodeService_ValidateConnectProxy(t *testing.T) {
 	}
 }
 
+func TestStructs_NodeService_ValidateConnectProxyWithAgentAutoAssign(t *testing.T) {
+	t.Run("connect-proxy: no port set", func(t *testing.T) {
+		ns := TestNodeServiceProxy(t)
+		ns.Port = 0
+
+		err := ns.ValidateForAgent()
+		assert.True(t, err == nil)
```
huikang, line 1166: nit: assert.NoError(t, err)

agent/structs/structs.go
```diff
@@ -1411,6 +1411,27 @@ func (s *NodeService) IsGateway() bool {
...
+	}
+
+	return result
+}
+
+// ValidateForAgent does a subset validation, with the assumption that a local agent can assist with missing values.
+//
+// I.e. in the catalog case, a local agent cannot be assumed to facilitate auto-assignment of port or socket path,
+// so additional checks are needed.
+func (s *NodeService) ValidateForAgent() error {
```
huikang, line 1432: Maybe a less disruptive change will be add a parameter to `Validate()` method, so it will look like

```
func (s *NodeService) Validate(atAgent bool) error {
    // skip if validate at agent endpoint
    if (!atAgent) && s.Port == 0 && s.SocketPath == "" {
	    result = multierror.Append(result, fmt.Errorf("Port or SocketPath must be set for a %s", s.Kind))
    }
}
```
What do you think?

agent/structs/structs_test.go
skpratt: Done

agent/structs/structs.go
skpratt: I contemplated going this way. However there are several call sites outside the agent where it felt messy to add this implementation information, especially with the upcoming shifts for Agentless. Keeping awareness of agent-specific capabilities with the agent code as much as possible felt like the better approach to me.
What do you think?

huikang: Sounds good. Thanks for the explanation, Sarah!

riddhi89 (approved):
LGTM

jkirschner-hashicorp:
@skpratt : does this close #8254?

PR #13907, opened 2022-07-26
chore: clarify the error message: service.service must not be empty

huikang:
### Description
Normally, registering a service, we use "name". However, when register service using catalog endpoint, the key of service
name actually should be "service". 

Adding this information to the error message will help user quickly fix in the request.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/consul/catalog_endpoint.go
+++ b/agent/consul/catalog_endpoint.go
@@ -176,7 +176,7 @@ func servicePreApply(service *structs.NodeService, authz resolver.Result, authzC
 
 	// Verify ServiceName provided if ID.
 	if service.ID != "" && service.Service == "" {
-		return fmt.Errorf("Must provide service name with ID")
+		return fmt.Errorf("Must provide service name (Service.Service) when service ID is provided")
 	}
 
 	// Check the service address here and in the agent endpoint


agent/consul/catalog_endpoint.go
```diff
@@ -176,7 +176,7 @@ func servicePreApply(service *structs.NodeService, authz resolver.Result, authzC
 
 	// Verify ServiceName provided if ID.
 	if service.ID != "" && service.Service == "" {
-		return fmt.Errorf("Must provide service name with ID")
+		return fmt.Errorf("Must provide service name with ID (service.service can't be empty)")
```
eculver, line 179: Shall we capitalize to match [the example payloads in the docs](https://www.consul.io/api-docs/catalog#sample-payload)?
```suggestion
		return fmt.Errorf("Must provide service name with ID (Service.Service can't be empty)")
```

eculver: I'm also wondering if saying "Must provide service name when service ID is empty" helps further clarify?

eculver: Yeah, you're right, it's `Service` not `Name`. My last comment was confusing/wrong; it's when service ID _is_ provided.

I like your suggestion. If I had to make one last suggestion, it would be a combination of them all: 

> "Must provide service name (Service.Service) when service ID is provided"

That's the last I'll say. Sorry to belabor the point. I'll leave it up to you as to what you go with.

eculver (approved):

agent/consul/catalog_endpoint.go
huikang: The latest version version is the clearest so far, so I will use that one. Thanks, @eculver .

PR #13902, opened 2022-07-26
Add CLI Command Characteristics and HTTP API Cross-References

jkirschner-hashicorp:
Follow-up to https://github.com/hashicorp/consul/pull/12045

Applies that pattern to all CLI commands

PR #13917, opened 2022-07-27
docs: show CLI cmd-specific opts before general opts

jkirschner-hashicorp:
Applies the pattern agreed upon in #12903 to all other CLI commands

### DIFF
--- a/website/content/commands/acl/set-agent-token.mdx
+++ b/website/content/commands/acl/set-agent-token.mdx
@@ -26,9 +26,9 @@ are not supported from commands, but may be from the corresponding HTTP endpoint
 
 ## Usage
 
-Usage: `consul acl set-agent-token [options] TYPE TOKEN`
+Usage: `consul acl set-agent-token [options] token_type token_secret_id`
 
-### Token Types
+The token types are:
 
 - `default` - The default token is the token that the agent will use for
   both internal agent operations and operations initiated by the HTTP
--- a/website/content/commands/acl/translate-rules.mdx
+++ b/website/content/commands/acl/translate-rules.mdx
@@ -26,12 +26,6 @@ are not supported from commands, but may be from the corresponding HTTP endpoint
 
 Usage: `consul acl translate-rules [options] TRANSLATE`
 
-#### API Options
-
-@include 'http_api_options_client.mdx'
-
-@include 'http_api_options_server.mdx'
-
 #### Command Options
 
 - `TRANSLATE` - The rules to translate. If `-` is used, then
@@ -46,6 +40,12 @@ Usage: `consul acl translate-rules [options] TRANSLATE`
 - `-token-accessor` - Specifies that what the `TRANSLATE` argument
   holds is not a rule set but rather the token accessor ID of a
   legacy ACL token that holds the rule set.
+
+#### API Options
+
+@include 'http_api_options_client.mdx'
+
+@include 'http_api_options_server.mdx'
 
 ### Examples
 
--- a/website/content/commands/connect/ca.mdx
+++ b/website/content/commands/connect/ca.mdx
@@ -89,11 +89,13 @@ Usage: `consul connect ca set-config [options]`
 
 Corresponding HTTP API Endpoint: [\[PUT\] /v1/connect/ca/configuration](/api-docs/connect/ca#update-ca-configuration)
 
-#### API Options
+The output looks like this:
 
-@include 'http_api_options_client.mdx'
+```
+Configuration updated!
+```
 
-@include 'http_api_options_server.mdx'
+The return code will indicate success or failure.
 
 #### Command Options
 
@@ -108,10 +110,8 @@ Corresponding HTTP API Endpoint: [\[PUT\] /v1/connect/ca/configuration](/api-doc
   Cross-Signing](/docs/connect/ca#forced-rotation-without-cross-signing)
   for more detail.
 
-The output looks like this:
+#### API Options
 
-```
-Configuration updated!
-```
+@include 'http_api_options_client.mdx'
 
-The return code will indicate success or failure.
+@include 'http_api_options_server.mdx'
--- a/website/content/commands/operator/area.mdx
+++ b/website/content/commands/operator/area.mdx
@@ -61,11 +61,13 @@ are not supported from commands, but may be from the corresponding HTTP endpoint
 
 Usage: `consul operator area create [options]`
 
-#### API Options
+The following example output shows the ID of the newly-created network area:
 
-@include 'http_api_options_client.mdx'
+```text
+Created area "d2872ec5-68ea-b862-b75d-0bee99aca100" with peer datacenter "other"!
+```
 
-@include 'http_api_options_server.mdx'
+The return code indicates success or failure.
 
 #### Command Options
 
@@ -79,13 +81,11 @@ Usage: `consul operator area create [options]`
 - `-use-tls=<value>` Specifies whether gossip over this area should be encrypted with
   TLS if possible. Must be either `true` or `false`.
 
-The output looks like this, displaying the ID of the newly-created network area:
+#### API Options
 
-```text
-Created area "d2872ec5-68ea-b862-b75d-0bee99aca100" with peer datacenter "other"!
-```
+@include 'http_api_options_client.mdx'
 
-The return code will indicate success or failure.
+@include 'http_api_options_server.mdx'
 
 ## delete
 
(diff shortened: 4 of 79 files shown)


website/content/commands/acl/set-agent-token.mdx
```diff
@@ -26,7 +26,7 @@ are not supported from commands, but may be from the corresponding HTTP endpoint
 
 ## Usage
 
-Usage: `consul acl set-agent-token [options] TYPE TOKEN`
+Usage: `consul acl set-agent-token [options] [TYPE](#token-types) TOKEN`
```
jkirschner-hashicorp, line 29: Note to self: check whether this anchor link works as expected

website/content/commands/acl/translate-rules.mdx
```diff
@@ -24,13 +24,7 @@ are not supported from commands, but may be from the corresponding HTTP endpoint
...
 ### Usage
 
-Usage: `consul acl translate-rules [options] TRANSLATE`
-
-#### API Options
-
-@include 'http_api_options_client.mdx'
-
-@include 'http_api_options_server.mdx'
+Usage: `consul acl translate-rules [options] [TRANSLATE](#translate)`
```
jkirschner-hashicorp, line 27: Note to self: check whether this anchor link works as expected

website/content/commands/connect/ca.mdx
```diff
@@ -89,11 +89,13 @@ Usage: `consul connect ca set-config [options]`
 
 Corresponding HTTP API Endpoint: [\[PUT\] /v1/connect/ca/configuration](/api-docs/connect/ca#update-ca-configuration)
 
-#### API Options
+The output looks like this:
```
jkirschner-hashicorp, line 92: Moved the output to just below the usage (rather than after the long list of general API options).

website/content/commands/operator/area.mdx
```diff
@@ -170,6 +156,20 @@ to join the given address.
 
 The return code will indicate success or failure.
 
+#### Command Options
+
+- `-id=<value>` - Looks up the area to operate on by its ID. This can be given
+  instead of a peer datacenter.
+
+- `-peer-datacenter=<value>` - Looks up the area to operate on by its peer
+  datacenter. This can be given instead of an ID.
```
trujillo-adam, line 165:
```suggestion
- `-peer-datacenter=<value>` - Specifies the peer datacenter of a network area to operate on. Use this option as an alternative to `-id`.
```

website/content/commands/operator/area.mdx
```diff
@@ -273,6 +259,20 @@ to the given server, in a human-readable format. This is computed using
 
 The return code will indicate success or failure.
 
+#### Command Options
+
+- `-id=<value>` - Looks up the area to operate on by its ID. This can be given
+  instead of a peer datacenter.
```
trujillo-adam, line 265:
```suggestion
- `-id=<value>` - Specifies the ID of a network area to operate on. Use this option as an alternative to `-peer-datacenter`.
```

website/content/commands/operator/area.mdx
```diff
@@ -273,6 +259,20 @@ to the given server, in a human-readable format. This is computed using
...
+
+- `-peer-datacenter=<value>` - Looks up the area to operate on by its peer
+  datacenter. This can be given instead of an ID.
```
trujillo-adam, line 268:
```suggestion
- `-peer-datacenter=<value>` - Specifies the peer datacenter of a network area to operate on. Use this option as an alternative to `-id`.
```

website/content/commands/operator/area.mdx
```diff
@@ -289,11 +289,13 @@ are not supported from commands, but may be from the corresponding HTTP endpoint
...
-#### API Options
+The output looks like this:
 
-@include 'http_api_options_client.mdx'
+```text
+Updated area "d2872ec5-68ea-b862-b75d-0bee99aca100"
+```
 
-@include 'http_api_options_server.mdx'
+The return code will indicate success or failure.
```
trujillo-adam, line 298:
```suggestion
The return code indicates success or failure.
```

website/content/commands/operator/autopilot.mdx
```diff
@@ -77,11 +77,14 @@ are not supported from commands, but may be from the corresponding HTTP endpoint
...
+#### Sample Output
 
-@include 'http_api_options_client.mdx'
+```sh
+$ consul operator autopilot set-config -min-quorum=3
+Configuration updated!
+```
 
-@include 'http_api_options_server.mdx'
+The return code will indicate success or failure.
```
trujillo-adam, line 87:
```suggestion
The return code indicates success or failure.
```

website/content/commands/operator/raft.mdx
```diff
@@ -70,6 +66,12 @@ Raft configuration.
 `Voter` is "true" or "false", indicating if the server has a vote in the Raft
 configuration.
 
+#### Command Options
+
+- `-stale` - Optional and defaults to "false" which means the leader provides
+  the result. If the cluster is in an outage state without a leader, you may need
+  to set this to "true" to get the configuration from a non-leader server.
```
trujillo-adam, line 73:
```suggestion
- `-stale` - Enables non-leader servers to provide cluster state information. We recommend using this option if the cluster is in an outage state without a leader. Default is `false`. 
```

website/content/commands/services/deregister.mdx
```diff
@@ -38,23 +38,23 @@ in HCL or JSON format.
 This flexibility makes it easy to pair the command with the
 `services register` command since the argument syntax is the same.
 
-#### API Options
+#### Command Options
 
-@include 'http_api_options_client.mdx'
+The flags below should only be set if _no arguments_ are given. If no
+arguments are given, the flags below can be used to deregister a single
+service.
```
trujillo-adam, line 45:
```suggestion
```

website/content/commands/services/deregister.mdx
```diff
@@ -38,23 +38,23 @@ in HCL or JSON format.
...
 
-#### Enterprise Options
+- `-id` - The ID of the service.
```
trujillo-adam, line 47:
```suggestion
- `-id` - Specifies the ID of a single service to deregister. Do not use this flag when providing a file containing multiple services that you want to deregister. 
```
The docs for this command are confusing, but I think this is what we mean. It was a lot of words/space for a single flag with one usage caveat.

website/content/commands/snapshot/inspect.mdx
```diff
@@ -40,6 +40,13 @@ The following fields are displayed when inspecting a snapshot:
 
 Usage: `consul snapshot inspect [options] FILE`
 
+#### Command Options
+
+- `-kvdetails` - Optional, provides a space usage breakdown for any KV data stored in Consul.
+- `-kvdepth` - Can only be used with `-kvdetails`. Used to adjust the grouping level of keys. Defaults to 2.
+- `-kvfilter` - Can only be used with `-kvdetails`. Used to specify a key prefix that excludes keys that don't match.
+- `-format` - Optional, allows from changing the output to JSON. Parameters accepted are "pretty" and "JSON".
+
```
trujillo-adam, line 49:
```suggestion
- `-kvdetails` - Provides information about space usage for KV data stored in Consul.
- `-kvdepth` - Adjusts the grouping levels of keys. Can only be used with `-kvdetails`. Defaults is `2`.
- `-kvfilter` - Specifies a key prefix to exclude from the response. Keys that match the specified value are excluded. Can only be used with `-kvdetails`.
- `-format` - Specifies an output format for the response. Specify either `pretty` to format the response with line breaks or `JSON` to format the response as JSON.
```

trujillo-adam (approved):
I left a bunch of out-of-scope suggestions to clean up some content while we were at it.

website/content/commands/acl/translate-rules.mdx
jkirschner-hashicorp: Darn. I think links can only work if placed _outside_ the tickmarks. I'll have to take a different approach to this. (I think you may have suggested an alternative - need to finish looking at the review comments!)

website/content/commands/acl/set-agent-token.mdx
jkirschner-hashicorp: Yes, but it looks like this won't work. Was trying to make an explicit connection to the list of types below (rather than an implicit one).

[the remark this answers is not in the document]
jkirschner-hashicorp: It used to say `TYPE`, which felt ambiguous ... type of _what_?

I don't think I can surround by `[]`, as the typically suggests that the argument is optional. Required, positional arguments usually usually just list a name (e.g., `token_type`).

Example:
image

This is something we should align on for the CLI command docs generally and then apply everywhere, but we'll leave it outside the scope of this PR (which is already large enough).

website/content/commands/acl/translate-rules.mdx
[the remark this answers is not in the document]
jkirschner-hashicorp: I think listing two separate usages is the easiest thing to do.

That said, this particular command is deprecated and will likely be removed in an upcoming release, so I won't spend much time on figuring it out for just _this_ command. (If it applies to other, non-deprecated commands, I'll mock something up.)

website/content/commands/operator/area.mdx
jkirschner-hashicorp: Thanks for this comment, I just realized that the term "peer datacenter" here needs to be disambiguated from "peer datacenter" in the context of cluster peering (which is related, but not the same / incompatible with network areas). I'll see that we address that in a subsequent PR.

website/content/commands/acl/translate-rules.mdx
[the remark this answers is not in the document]
jkirschner-hashicorp: Just going to leave this the way it originally was (since this command is deprecated)

jkirschner-hashicorp:
Crossing my fingers that this won't create too many merge conflicts I need to resolve with https://github.com/hashicorp/consul/pull/13902 ...

PR #13923, opened 2022-07-27
Fix some flaky tests

kisunji:
#### TestAgentConfigWatcherSidecarProxy
The watch is established in a background goroutine and the first assertion proves that the watcher is active so there is no reason for the update to happen in a racy goroutine.

Note that this does not completely remove the race condition as the first call to testGetConfigValTimeout could time out before a config is returned.

#### TestServerExportedPeeredServices
The ServiceName slice was being sourced from a non-deterministic `SliceOfKeys` function.
Updated code to sort the names deterministically so tests don't flake on the slice order.

#### TestInternal_PeeredUpstreams
Shortened the polling period for VIP metadata check and added retries (aligned with existing VIP tests)

#### TestConnectCA_ConfigurationSet_ChangeKeyConfig_Primary
This one is one of the top flaked tests but ONLY in arm64 environments. I haven't found the root cause but assuming the environment is low on resources, the 16 parallel tests spinning up servers and RPC polling probably doesn't help. Removed some redundant tests and switched from RPC polling to checking a server's own raft leadership status.

* No added functionality; tests should pass as-is (at a higher success rate)

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/consul/helper_test.go
+++ b/agent/consul/helper_test.go
@@ -1213,10 +1213,12 @@ func registerTestRoutingConfigTopologyEntries(t *testing.T, codec rpc.ClientCode
 func registerLocalAndRemoteServicesVIPEnabled(t *testing.T, state *state.Store) {
 	t.Helper()
 
-	_, entry, err := state.SystemMetadataGet(nil, structs.SystemMetadataVirtualIPsEnabled)
-	require.NoError(t, err)
-	require.NotNil(t, entry)
-	require.Equal(t, "true", entry.Value)
+	retry.Run(t, func(r *retry.R) {
+		_, entry, err := state.SystemMetadataGet(nil, structs.SystemMetadataVirtualIPsEnabled)
+		require.NoError(r, err)
+		require.NotNil(r, entry)
+		require.Equal(r, "true", entry.Value)
+	})
 
 	// Register a local connect-native service
 	require.NoError(t, state.EnsureRegistration(10, &structs.RegisterRequest{
--- a/agent/consul/leader_connect_test.go
+++ b/agent/consul/leader_connect_test.go
@@ -36,7 +36,7 @@ func TestConnectCA_ConfigurationSet_ChangeKeyConfig_Primary(t *testing.T) {
 		keyBits int
 	}{
 		{connect.DefaultPrivateKeyType, connect.DefaultPrivateKeyBits},
-		{"ec", 256},
+		// {"ec", 256}, skip since values are same as Defaults
 		{"ec", 384},
 		{"rsa", 2048},
 		{"rsa", 4096},
@@ -55,7 +55,7 @@ func TestConnectCA_ConfigurationSet_ChangeKeyConfig_Primary(t *testing.T) {
 				providerState := map[string]string{"foo": "dc1-value"}
 
 				// Initialize primary as the primary DC
-				dir1, srv := testServerWithConfig(t, func(c *Config) {
+				_, srv := testServerWithConfig(t, func(c *Config) {
 					c.Datacenter = "dc1"
 					c.PrimaryDatacenter = "dc1"
 					c.Build = "1.6.0"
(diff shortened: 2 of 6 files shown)


connect/proxy/config_test.go
```diff
@@ -143,23 +142,17 @@ func TestAgentConfigWatcherSidecarProxy(t *testing.T) {
 			},
 		},
 	}
-
 	require.Equal(t, expectCfg, cfg)
 
 	// Now keep watching and update the config.
-	go func() {
```
kisunji, line 150: Correct me if I'm wrong on this! This test isn't _too_ flaky but I have seen it flake recently so wanted to fix it.

The watch is established in a background goroutine and the first assertion proves that the watcher is active so there is no reason for the update to happen in a racy goroutine.

Note that this does not completely remove the race condition as the first call to testGetConfigValTimeout could time out before a config is returned.

agent/proxycfg-glue/exported_peered_services_test.go
```diff
@@ -69,7 +70,7 @@ func TestServerExportedPeeredServices(t *testing.T) {
 		)
 	})
 
-	t.Run("update exported services", func(t *testing.T) {
+	testutil.RunStep(t, "update exported services", func(t *testing.T) {
```
kisunji, line 73: This step can't pass without the first passing (consumes the queued event)

agent/consul/state/peering.go
```diff
@@ -353,7 +353,9 @@ func (s *Store) ExportedServicesForAllPeersByName(ws memdb.WatchSet, entMeta acl
 		}
 		m := list.ListAllDiscoveryChains()
 		if len(m) > 0 {
-			out[peering.Name] = maps.SliceOfKeys(m)
+			sns := maps.SliceOfKeys[structs.ServiceName, structs.ExportedDiscoveryChainInfo](m)
```
kisunji, line 356: Added explicit types to prove the type conversion works
```
structs.ServiceList = []structs.ServiceName
```

agent/consul/helper_test.go
```diff
@@ -1213,10 +1213,12 @@ func registerTestRoutingConfigTopologyEntries(t *testing.T, codec rpc.ClientCode
 func registerLocalAndRemoteServicesVIPEnabled(t *testing.T, state *state.Store) {
 	t.Helper()
 
-	_, entry, err := state.SystemMetadataGet(nil, structs.SystemMetadataVirtualIPsEnabled)
-	require.NoError(t, err)
-	require.NotNil(t, entry)
-	require.Equal(t, "true", entry.Value)
+	retry.Run(t, func(r *retry.R) {
```
kisunji, line 1216: This is aligned with an existing test for Virtual IPs: https://github.com/hashicorp/consul/blob/2e4cb6f77d2be36b02e9be0b289b24e5b0afb794/agent/consul/leader_test.go#L2201

connect/proxy/config_test.go
kisunji: The original issue I wanted to fix was that `require.NoError` within the goroutine could fire and make go test angry (causing completely unrelated tests to abort). While I was looking for ways to fix it, I figured the registration didn't need to happen in a goroutine.

agent/consul/leader_connect_test.go
```diff
@@ -55,20 +55,17 @@ func TestConnectCA_ConfigurationSet_ChangeKeyConfig_Primary(t *testing.T) {
...
 					c.CAConfig.Config["PrivateKeyBits"] = src.keyBits
 					c.CAConfig.Config["test_state"] = providerState
 				})
-				defer os.RemoveAll(dir1)
-				defer srv.Shutdown()
 				codec := rpcClient(t, srv)
-				defer codec.Close()
 
-				testrpc.WaitForLeader(t, srv.RPC, "dc1")
+				waitForLeaderEstablishment(t, srv)
```
kisunji, line 68: This is a nice RPC-free way to wait for leader establishment. It makes no RPCs and reads from the server's own raft state.

kisunji: https://github.com/hashicorp/consul/blob/bb4d4040fb0c7fd6df56ebd0798876289927adb7/agent/consul/server_test.go#L95-L107

agent/consul/leader_connect_test.go
```diff
@@ -36,7 +36,7 @@ func TestConnectCA_ConfigurationSet_ChangeKeyConfig_Primary(t *testing.T) {
 		keyBits int
 	}{
 		{connect.DefaultPrivateKeyType, connect.DefaultPrivateKeyBits},
-		{"ec", 256},
+		// {"ec", 256}, skip since values are same as Defaults
```
lkysow, line 39: Should we just delete this line then?

lkysow (approved):
nice work! 🔍

agent/consul/leader_connect_test.go
kisunji: I wanted to leave it in case the Defaults ever changed in the future (in which case we might want to un-comment)

lkysow: 👍

PR #13924, opened 2022-07-27
peering: don't track imported services/nodes in usage

lkysow:
### Description
Services/nodes that are imported from other peers are stored in
state. We don't want to count those as part of our own cluster's usage.

A follow-up PR will be added to track these imported resources separately.

### DIFF
--- a/agent/consul/state/usage.go
+++ b/agent/consul/state/usage.go
@@ -99,11 +99,20 @@ func updateUsage(tx WriteTxn, changes Changes) error {
 
 		switch change.Table {
 		case tableNodes:
+			node := changeObject(change).(*structs.Node)
+			if node.PeerName != "" {
+				// TODO(peering) track peered nodes separately. For now not tracking to avoid double billing.
+				continue
+			}
 			usageDeltas[change.Table] += delta
 			addEnterpriseNodeUsage(usageDeltas, change)
 
 		case tableServices:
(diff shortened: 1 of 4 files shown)


agent/grpc-external/services/peerstream/health_snapshot.go
```diff
@@ -29,8 +29,6 @@ func newHealthSnapshot(all []structs.CheckServiceNode, partition, peerName strin
 	// For all nodes, services, and checks we override the peer name and
 	// partition to be the local partition and local name for the peer.
 	for _, instance := range all {
-		// For all nodes, services, and checks we override the peer name and partition to be
-		// the local partition and local name for the peer.
 		instance.Node.PeerName = peerName
 		instance.Node.OverridePartition(partition)
```
lkysow, line 34: Cleaning up duplicate comment

agent/consul/state/usage.go
```diff
@@ -99,11 +99,20 @@ func updateUsage(tx WriteTxn, changes Changes) error {
 
 		switch change.Table {
 		case tableNodes:
```
lkysow, line 101: Reviewers: is checking for `PeerName` field all that's needed to catch if these objects are imported?

kisunji: Yep, should be enough.

kisunji (approved):
LGTM

PR #13925, opened 2022-07-27
Ensure connections are closed before WaitGroup marked as done

kisunji:
The previous ordering of defers meant the listener's connWG could fire and wake up other goroutines before the connection closed. Unsure if this caused any real bugs but this commit should make the code more correct.

* No behavioral change 

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/connect/proxy/listener.go
+++ b/connect/proxy/listener.go
@@ -166,9 +166,11 @@ func (l *Listener) Serve() error {
 
 // handleConn is the internal connection handler goroutine.
 func (l *Listener) handleConn(src net.Conn) {
-	defer src.Close()
-	// Make sure Listener.Close waits for this conn to be cleaned up.
-	defer l.connWG.Done()
+	defer func() {
+		// Make sure Listener.Close waits for this conn to be cleaned up.
+		src.Close()
+		l.connWG.Done()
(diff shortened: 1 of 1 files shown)


connect/proxy/listener.go
```diff
@@ -166,9 +166,11 @@ func (l *Listener) Serve() error {
 
 // handleConn is the internal connection handler goroutine.
 func (l *Listener) handleConn(src net.Conn) {
-	defer src.Close()
-	// Make sure Listener.Close waits for this conn to be cleaned up.
-	defer l.connWG.Done()
```
kisunji, line 171: Defers are stack-based so `src.Close()` will be the last to be called in this function

connect/proxy/listener.go
```diff
@@ -166,9 +166,11 @@ func (l *Listener) Serve() error {
...
+	defer func() {
+		// Make sure Listener.Close waits for this conn to be cleaned up.
+		src.Close()
+		l.connWG.Done()
```
kisunji, line 172: Now we serially close the connection first then mark waitgroup as done

boxofrad (approved):

PR #13929, opened 2022-07-27
[sync] fix empty partitions matching

acpana:
This was introduced in `ef7edd79c4515789f767c28c36b443da737a0837` in ENT.

Essentially, `peering` objects from the state store come up with an empty string partition `""`. This overrides that to be called `default`. Maybe there's a better struct / way to do this so please comment if so!

### DIFF
--- a/agent/rpc/peering/service.go
+++ b/agent/rpc/peering/service.go
@@ -346,7 +346,7 @@ func (s *Server) Establish(
 		return nil, err
 	}
 
-	if err := s.validatePeeringInPartition(tok.PeerID, entMeta.PartitionOrDefault()); err != nil {
+	if err := s.validatePeeringInPartition(tok.PeerID, entMeta.PartitionOrEmpty()); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


agent/rpc/peering/service.go
```diff
@@ -409,8 +408,15 @@ func (s *Server) validatePeeringInPartition(remotePeerID, partition string) erro
...
-		return fmt.Errorf("cannot create a peering within the same partition (ENT) or cluster (OSS)")
+	if peering != nil {
+		stateStorePart := peering.Partition
+		if stateStorePart == "" {
+			stateStorePart = "default"
+		}
+
+		if stateStorePart == partition {
+			return fmt.Errorf("cannot create a peering within the same partition (ENT) or cluster (OSS)")
+		}
```
eculver, line 419: Turns out `peering.PartitionOrDefault()` is a thing.
```suggestion
	if peering != nil {
		if peering.PartitionOrDefault() == partition {
			return fmt.Errorf("cannot create a peering within the same partition (ENT) or cluster (OSS)")
		}
```

kisunji: We should use `acl.EqualPartitions` which treats `"" == "default"`

kisunji (approved):

PR #13938, opened 2022-07-28
peering: retry establishing connection more quickly on certain errors

lkysow:
When running with an LB in front of the Consul servers, dialers need to retry their requests until they land on a leader server. The default retry backoff was too long because it started at 2s and went exponentially from there: 4s, 8s, etc. This change adds new retry logic for specific errors to retry quickly.

I opted to treat all `FailedPrecondition` errors as quick-retry errors because they should all be resolved quickly.

Changes:
1. replace `retryLoopBackoff` with `retryLoopBackoffPeering` that implements the new retry mechanic
2. change logging everywhere we receive a `FailedPrecondition` error to not log as an error because they aren't indicative of an actual problem
3. refactor the main `HandleStream()` loop to add a new channel for errors. Previously, any error received from `Recv()` would be swallowed and then we'd close the `recvCh` and use that closing to trigger the loop to exit. Now we're sending that error through the new channel and handling it within the main `for{}` loop.

### DIFF
--- a/agent/consul/leader_peering.go
+++ b/agent/consul/leader_peering.go
@@ -361,10 +361,7 @@ func (s *Server) establishStream(ctx context.Context, logger hclog.Logger, peer 
 		if err == nil {
 			stream.CloseSend()
 			s.peerStreamServer.DrainStream(streamReq)
-
-			// This will cancel the retry-er context, letting us break out of this loop when we want to shut down the stream.
 			cancel()
-
 			logger.Info("closed outbound stream")
 		}
 		return err
@@ -544,9 +541,8 @@ func (s *Server) deleteTrustBundleFromPeer(ctx context.Context, limiter *rate.Li
 // This function is modelled off of retryLoopBackoffHandleSuccess but is specific to peering
 // because peering needs to use different retry times depending on which error is returned.
 // This function doesn't use a rate limiter, unlike retryLoopBackoffHandleSuccess, because
-// the rate limiter is only needed when retrying when loopFn returns a nil error. In the streaming
-// case, when loopFn returns a nil error, the ctx has already been cancelled, and so we won't re-loop
-// but instead exit.
+// the rate limiter is only needed in the success case when loopFn returns nil and we want to
+// loop again. In the peering case, we exit on a successful loop so we don't need the limter.
 func retryLoopBackoffPeering(ctx context.Context, logger hclog.Logger, loopFn func() error, errFn func(error),
 	retryTimeFn func(failedAttempts uint, loopErr error) time.Duration) {
 	var failedAttempts uint
@@ -571,16 +567,7 @@ func retryLoopBackoffPeering(ctx context.Context, logger hclog.Logger, loopFn fu
 			}
 			continue
 		}
-
-		failedAttempts = 0
-		select {
-		// NOTE: this is important to check here before re-entering the loop because our
-		// caller will cancel the context when the stream exits gracefully, and so we don't want to
-		// reconnect.
-		case <-ctx.Done():
-			return
-		default:
-		}
+		return
 	}
 }
 
--- a/agent/consul/leader_peering.go
+++ b/agent/consul/leader_peering.go
@@ -16,8 +16,10 @@ import (
 	"github.com/hashicorp/go-uuid"
 	"golang.org/x/time/rate"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/keepalive"
+	grpcstatus "google.golang.org/grpc/status"
 
 	"github.com/hashicorp/consul/acl"
 	"github.com/hashicorp/consul/agent/consul/state"
@@ -300,7 +302,7 @@ func (s *Server) establishStream(ctx context.Context, logger hclog.Logger, peer 
 	}
 
 	// Establish a stream-specific retry so that retrying stream/conn errors isn't dependent on state store changes.
-	go retryLoopBackoff(retryCtx, func() error {
+	go retryLoopBackoffPeering(retryCtx, logger, func() error {
 		// Try a new address on each iteration by advancing the ring buffer on errors.
 		defer func() {
 			buffer = buffer.Next()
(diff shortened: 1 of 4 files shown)


agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -331,13 +332,34 @@ func (s *Server) realHandleStream(streamReq HandleStreamRequest) error {
 
 			return nil
 
+		case err := <-errCh:
+			if err == io.EOF {
+				logger.Info("stream ended by peer")
```
kisunji, line 337: Missing a `status.TrackRecvError(err.Error())` here?

agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -331,13 +332,34 @@ func (s *Server) realHandleStream(streamReq HandleStreamRequest) error {
...
+				return nil
+			}
+			grpcErr, ok := grpcstatus.FromError(err)
+			if ok {
+				// todo: return a sentinel error, don't string match.
```
kisunji, line 342: I think in gRPC cases since errors are transferred over the wire it might be hard to serialize sentinel go errors in a straightforward way. We could make the error strings sentinel consts so changes are reflected on both ends of the wire.

agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -331,13 +332,34 @@ func (s *Server) realHandleStream(streamReq HandleStreamRequest) error {
...
+				if grpcErr.Code() == codes.FailedPrecondition && grpcErr.Message() == "cannot establish a peering stream on a follower node" {
+					logger.Debug("connected with a follower", "error", err)
+					return ConnectedWithFollowerErr
```
kisunji, line 345: do we want to `status.TrackRecvError` here as well?

agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -251,26 +257,21 @@ func (s *Server) realHandleStream(streamReq HandleStreamRequest) error {
...
+				return
 			}
-
-			if err == io.EOF {
-				logger.Info("stream ended by peer")
-				status.TrackRecvError(err.Error())
+			logTraceRecv(logger, msg)
+			select {
+			case recvCh <- msg:
+			case <-ctx.Done():
```
kisunji, line 272: Should the heartbeat goroutine below also use this `ctx`? Currently it uses `streamReq.Stream.Context()` but this `ctx` will kill the loop faster.

agent/grpc-external/services/peerstream/stream_resources.go
kisunji: Design Q: isn't it redundant to error check and return a separate sentinel error? Could this checking happen by the caller and we just return the error as is?

agent/consul/leader_connect.go
```diff
@@ -259,6 +261,48 @@ func retryLoopBackoffHandleSuccess(ctx context.Context, loopFn func() error, err
...
+		}
+		retryTime := (1 << failedAttempts) * time.Second
+
+		if err = loopFn(); err != nil {
+			errFn(err)
+
+			if err == peerstream.ConnectedWithFollowerErr {
+				fmt.Printf("setting retry to 10ms\n")
+				retryTime = 10 * time.Millisecond
+			}
```
kisunji, line 287: I made an earlier comment below about moving some of the error-checking here so we don't pass around an extra sentinel error. I think in an ideal case, any gRPC errors with `FailedPrecondition` should be retried quickly but I haven't combed through all the possible errors to see if that's true or not.

kisunji (commented):
Overall looks good but I left a few questions

agent/grpc-external/services/peerstream/stream_resources.go
lkysow: I found a way that works pretty well by type matching on `Details()` that  i got from https://jbrandhorst.com/post/grpc-errors/

agent/grpc-external/services/peerstream/stream_resources.go
lkysow: a) it's needed to stop the error log and instead use a debug log b) totally we could just return the error but then I thought we're leaking our protobufs too far outside of this method

agent/consul/leader_peering.go
```diff
@@ -358,13 +360,47 @@ func (s *Server) establishStream(ctx context.Context, logger hclog.Logger, peer
 		return err
 
 	}, func(err error) {
```
lkysow, line 362: Log errors differently so as to not spam logs with not actual errors.

agent/consul/leader_peering.go
```diff
@@ -358,13 +360,47 @@ func (s *Server) establishStream(ctx context.Context, logger hclog.Logger, peer
...
+				"error", err)
+			return
+		}
+		logger.Error("error managing peering stream", "error", err)
+	}, peeringRetryTimeout)
 
 	return nil
 }
 
+// peeringRetryTimeout returns the time that should be waited between re-establishing a peering
```
lkysow, line 376: Just made up this backoff algo. Open to feedback on the params cc @ndhanushkodi

agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -242,58 +245,53 @@ func (s *Server) realHandleStream(streamReq HandleStreamRequest) error {
 			PeerID:      streamReq.RemoteID,
 		})
 		if err := streamSend(sub); err != nil {
```
lkysow, line 247: No longer swallowing this EOF error because if the stream is disconnected at this early stage then we definitely want to retry connecting. We should only ever return nil if the stream is disconnected gracefully and we don't want to reconnect.

agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -242,58 +245,53 @@ func (s *Server) realHandleStream(streamReq HandleStreamRequest) error {
...
 		}
 	}
 
-	// TODO(peering): Should this be buffered?
-	recvChan := make(chan *pbpeerstream.ReplicationMessage)
+	// recvCh sends messages from the gRPC stream.
+	recvCh := make(chan *pbpeerstream.ReplicationMessage)
+	// recvErrCh sends errors received from the gRPC stream.
+	recvErrCh := make(chan error)
+
```
lkysow, line 257: This function is refactored to send errors through `errCh`. Since we're sending errors through, we don't need to close `recvChan` as a signal that this goroutine has exited. Instead in our main for{} we can receive the error and then know that this goroutine has exited.

agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -331,28 +330,30 @@ func (s *Server) realHandleStream(streamReq HandleStreamRequest) error {
...
-				// The only time we expect the stream to end is when we've received a "Terminated" message.
-				// We handle the case of receiving the Terminated message below and then this function exits.
-				// So if the channel is closed while this function is still running then we haven't received a Terminated
-				// message which means we want to try and reestablish the stream.
-				// It's the responsibility of the caller of this function to reestablish the stream on error and so that's
-				// why we return an error here.
-				return fmt.Errorf("stream ended unexpectedly")
-			}
-
+		case msg := <-recvCh:
```
lkysow, line 350: We don't close `recvChan` anymore so no point checking if it's open

agent/grpc-external/services/peerstream/stream_resources.go
lkysow: I ended up returning the error 😄

agent/grpc-external/services/peerstream/stream_resources.go
lkysow: Yep, added!

agent/consul/leader_connect.go
lkysow: Implemented that!

agent/consul/leader_peering.go
```diff
@@ -517,3 +535,94 @@ func (s *Server) deleteTrustBundleFromPeer(ctx context.Context, limiter *rate.Li
...
+// This function is modelled off of retryLoopBackoffHandleSuccess but is specific to peering
+// because peering needs to use different retry times depending on which error is returned.
+// This function doesn't use a rate limiter, unlike retryLoopBackoffHandleSuccess, because
+// the rate limiter is only needed when retrying when loopFn returns a nil error. In the streaming
+// case, when loopFn returns a nil error, the ctx has already been cancelled, and so we won't re-loop
+// but instead exit.
+func retryLoopBackoffPeering(ctx context.Context, logger hclog.Logger, loopFn func() error, errFn func(error),
+	retryTimeFn func(failedAttempts uint, loopErr error) time.Duration) {
+	var failedAttempts uint
+	var err error
```
lkysow, line 553: err doesn't need to be here

kisunji (approved):
Looks reasonable! Approving and hopefully we can tweak as needed during beta

PR #13953, opened 2022-07-28
Tgtwy egress HTTP support

dhiaayachi:
### Description
This PR add the ability to perform http proxy to destination. 

Compared to TCP proxy where we use server_name to match on the destination and the port for non TLS traffic the server_name field is not populated by the http_inspector filter in envoy so we need a different way of matching traffic.

To do so we are matching on ports only at the TCP level and we are using the http_manager to match on the domain name for each destination and proxy traffic to the right cluster.

- FilterMatch will match on the common Port (for all destinations) 
- each port matching will have multiple virtual hosts which will match on the domain name and proxy to the right cluster.
- use RDS to generate the right routes in the virtual hosts.

This adds a limitation that the same port cannot be used for TCP and http like protocols at the same time, but unless we patch the http_inspector filter in envoy to populate the server_name field from the HOST header this is the only possible solution.

Added specific test for http with the associated golden file

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/proxycfg/testing_tproxy.go
+++ b/agent/proxycfg/testing_tproxy.go
@@ -555,6 +555,33 @@ func TestConfigSnapshotTransparentProxyDestination(t testing.T) *ConfigSnapshot 
 		}
 	)
 
+	serviceNodes := structs.CheckServiceNodes{
+		{
+			Node: &structs.Node{
+				Node:       "node1",
+				Address:    "172.168.0.1",
+				Datacenter: "dc1",
+			},
+			Service: &structs.NodeService{
+				ID:      "tgtw1",
+				Address: "172.168.0.1",
+				Port:    8443,
+				Kind:    structs.ServiceKindTerminatingGateway,
+				TaggedAddresses: map[string]structs.ServiceAddress{
+					structs.TaggedAddressLANIPv4:   {Address: "172.168.0.1", Port: 8443},
+					structs.TaggedAddressVirtualIP: {Address: "240.0.0.1"},
+				},
+			},
+			Checks: []*structs.HealthCheck{
+				{
+					Node:        "node1",
+					ServiceName: "tgtw",
+					Name:        "force",
+					Status:      api.HealthPassing,
+				},
+			},
+		},
+	}
 	return TestConfigSnapshot(t, func(ns *structs.NodeService) {
 		ns.Proxy.Mode = structs.ProxyModeTransparent
 	}, []UpdateEvent{
@@ -592,65 +619,13 @@ func TestConfigSnapshotTransparentProxyDestination(t testing.T) *ConfigSnapshot 
 		{
 			CorrelationID: DestinationGatewayID + googleUID.String(),
 			Result: &structs.IndexedCheckServiceNodes{
-				Nodes: structs.CheckServiceNodes{
-					{
-						Node: &structs.Node{
-							Node:       "node1",
-							Address:    "172.168.0.1",
-							Datacenter: "dc1",
-						},
-						Service: &structs.NodeService{
-							ID:      "tgtw1",
-							Address: "172.168.0.1",
-							Port:    8443,
-							Kind:    structs.ServiceKindTerminatingGateway,
-							TaggedAddresses: map[string]structs.ServiceAddress{
-								structs.TaggedAddressLANIPv4:   {Address: "172.168.0.1", Port: 8443},
-								structs.TaggedAddressVirtualIP: {Address: "240.0.0.1"},
-							},
-						},
-						Checks: []*structs.HealthCheck{
-							{
-								Node:        "node1",
-								ServiceName: "tgtw",
-								Name:        "force",
-								Status:      api.HealthPassing,
-							},
-						},
-					},
-				},
+				Nodes: serviceNodes,
 			},
 		},
 		{
 			CorrelationID: DestinationGatewayID + kafkaUID.String(),
 			Result: &structs.IndexedCheckServiceNodes{
-				Nodes: structs.CheckServiceNodes{
-					{
-						Node: &structs.Node{
-							Node:       "node1",
-							Address:    "172.168.0.1",
-							Datacenter: "dc1",
-						},
-						Service: &structs.NodeService{
-							ID:      "tgtw1",
-							Address: "172.168.0.1",
-							Port:    8443,
-							Kind:    structs.ServiceKindTerminatingGateway,
-							TaggedAddresses: map[string]structs.ServiceAddress{
-								structs.TaggedAddressLANIPv4:   {Address: "172.168.0.1", Port: 8443},
-								structs.TaggedAddressVirtualIP: {Address: "240.0.0.1"},
-							},
-						},
-						Checks: []*structs.HealthCheck{
-							{
-								Node:        "node1",
-								ServiceName: "tgtw",
-								Name:        "force",
-								Status:      api.HealthPassing,
-							},
-						},
-					},
-				},
+				Nodes: serviceNodes,
 			},
 		},
 	})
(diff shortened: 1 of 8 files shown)


agent/xds/listeners.go
```diff
@@ -227,30 +228,50 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
...
+				return nil
+			}
+			clusterName := clusterNameForDestination(cfgSnap, uid.Name, fmt.Sprintf("%d", svcConfig.Destination.Port), uid.NamespaceOrDefault(), uid.PartitionOrDefault())
 			filterChain, err := s.makeUpstreamFilterChain(filterChainOpts{
-				routeName:   uid.EnvoyID(),
-				clusterName: clusterName,
-				filterName:  clusterName,
-				protocol:    svcConfig.Protocol,
-				useRDS:      structs.IsProtocolHTTPLike(svcConfig.Protocol),
+				routeName:  clusterName,
```
DanStough, line 245: optional: clusterName is formatted `destination.<address>.<service>.....` Should we change the syntax for the routes since they can handle multiple addresses?

agent/xds/listeners.go
```diff
@@ -227,30 +228,50 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
...
+				useRDS:     true,
 			})
 			if err != nil {
 				return err
 			}
-
-			filterChain.FilterChainMatch = makeFilterChainMatchFromAddressWithPort(address, svcConfig.Destination.Port)
+			filterChain.FilterChainMatch = makeFilterChainMatchFromAddressWithPort("", svcConfig.Destination.Port)
 			outboundListener.FilterChains = append(outboundListener.FilterChains, filterChain)
+			requiresHTTPInspector = len(filterChain.FilterChainMatch.ServerNames) != 0 || requiresHTTPInspector
```
DanStough, line 255: If you got this far, this should always be true, correct?
```suggestion
			requiresHTTPInspector = true
```

agent/xds/listeners.go
```diff
@@ -227,30 +228,50 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
...
 
 	err = cfgSnap.ConnectProxy.DestinationsUpstream.ForEachKeyE(func(uid proxycfg.UpstreamID) error {
 		svcConfig, ok := cfgSnap.ConnectProxy.DestinationsUpstream.Get(uid)
 		if !ok || svcConfig == nil {
 			return nil
 		}
-
-		for _, address := range svcConfig.Destination.Addresses {
-			clusterName := clusterNameForDestination(cfgSnap, uid.Name, address, uid.NamespaceOrDefault(), uid.PartitionOrDefault())
+		configuredPorts := make(map[int]interface{})
```
DanStough, line 238: This is created on each key, so I think it's always going to be empty.

Also, do we ever write the ports we have "seen"?

agent/xds/routes.go
```diff
@@ -167,6 +204,21 @@ func (s *ResourceGenerator) makeRoutes(
...
+	route, err := makeNamedAddressesRouteWithLB(name, addresses, lb, autoHostRewrite)
+	if err != nil {
+		s.Logger.Error("failed to make route", "cluster", "error", err)
+		return nil, err
+	}
+	resources = append(resources, route)
+
+	return resources, nil
+}
+
```
DanStough, line 221: Any reason for the indirection with this function to call `makeNamedAddressesRouteWithLB`? Do we need any of the LoadBalancer logic?

DanStough (changes requested):

agent/xds/listeners.go
dhiaayachi: I'm not sure about this. My understanding is that the way routes relate to filterchains is by route_name.
In that case we have a filterchain per port so a route per port and for each address for a given port we will have a mapping in the relevant route to a cluster.

routes and filterchaines are named `destination.<port>.<service>....`
clusters and routes mapping are named `destination.<address>.<service>....`

am I missing something?

agent/xds/listeners.go
dhiaayachi: yeah I missed some of the logic

agent/xds/listeners.go
DanStough: Ah, I see you are replacing the address with the port number in the name now. Cool!

dhiaayachi: yeah but you are still right about not to use svcConfig.Name as it will be specific to the current destination

agent/proxycfg/testing_tproxy.go
```diff
@@ -655,3 +655,117 @@ func TestConfigSnapshotTransparentProxyDestination(t testing.T) *ConfigSnapshot
...
+							},
+						},
+					},
+				},
+			},
+		},
+		{
+			CorrelationID: DestinationGatewayID + kafkaUID.String(),
+			Result: &structs.IndexedCheckServiceNodes{
+				Nodes: structs.CheckServiceNodes{
```
kyhavlov, line 741: This is the same as the CheckServiceNodes above for `DestinationGatewayID + googleUID.String()`, could set it up above and reference it in both events.

agent/proxycfg/testing_tproxy.go
```diff
@@ -655,3 +655,117 @@ func TestConfigSnapshotTransparentProxyDestination(t testing.T) *ConfigSnapshot
 		},
 	})
 }
+
+func TestConfigSnapshotTransparentProxyDestinationHTTP(t testing.T) *ConfigSnapshot {
```
kyhavlov, line 659: This helper function is mostly identical to `TestConfigSnapshotTransparentProxyDestination` above aside from the protocol in the config entries - that function could probably be reused and the protocol/config entry modification parameterized to avoid copying the whole thing.

agent/proxycfg/testing_tproxy.go
```diff
@@ -655,3 +655,117 @@ func TestConfigSnapshotTransparentProxyDestination(t testing.T) *ConfigSnapshot
...
+		google    = structs.NewServiceName("google", nil)
+		googleUID = NewUpstreamIDFromServiceName(google)
+		googleCE  = structs.ServiceConfigEntry{Name: "google", Destination: &structs.DestinationConfig{Addresses: []string{"www.google.com"}, Port: 443}, Protocol: "http"}
+
+		kafka    = structs.NewServiceName("kafka", nil)
+		kafkaUID = NewUpstreamIDFromServiceName(kafka)
+		kafkaCE  = structs.ServiceConfigEntry{Name: "kafka", Destination: &structs.DestinationConfig{Addresses: []string{"192.168.2.1"}, Port: 9093}, Protocol: "http"}
+	)
+
+	return TestConfigSnapshot(t, func(ns *structs.NodeService) {
```
kyhavlov, line 672: `TestConfigSnapshot` would really benefit from a `populateServices` bool like `TestConfigSnapshotTerminatingGateway` and `TestConfigSnapshotIngressGateway` have to avoid adding base services that don't have anything to do with the test (just add clutter to the golden files generated). This is just a thought though, definitely not a blocker to merging.

kyhavlov (commented):
I had a few test comments but nothing blocking beyond what Dan has already mentioned. Looking good so far.

agent/xds/listeners.go
dhiaayachi: I changed the logic to fix that, please take another look

agent/proxycfg/testing_tproxy.go
dhiaayachi: This now have different data set. There is still something that could be done by extracting the entries and passing them in the call site but if that's ok I prefer to postpone that to another PR after the release.

agent/proxycfg/testing_tproxy.go
dhiaayachi: I will make sure to add that in the test clean up PR after the release.

agent/xds/listeners.go
```diff
@@ -227,30 +228,53 @@ func (s *ResourceGenerator) listenersFromSnapshotConnectProxy(cfgSnap *proxycfg.
...
 		}
 
-		for _, address := range svcConfig.Destination.Addresses {
-			clusterName := clusterNameForDestination(cfgSnap, uid.Name, address, uid.NamespaceOrDefault(), uid.PartitionOrDefault())
+		if structs.IsProtocolHTTPLike(svcConfig.Protocol) {
+			if _, ok := configuredPorts[svcConfig.Destination.Port]; ok {
+				return nil
+			}
+			configuredPorts[svcConfig.Destination.Port] = struct{}{}
+			const name = "~http" //name used for the shared route name
```
DanStough, line 245: I removed the destination prefix `~` based on Matt K's feedback about using this in the SNI. No strong opinions; I think it's only important that it's consistent.

agent/xds/routes.go
```diff
@@ -71,11 +72,62 @@ func (s *ResourceGenerator) routesForConnectProxy(cfgSnap *proxycfg.ConfigSnapsh
...
+			return nil
+		}
+		if !structs.IsProtocolHTTPLike(svcConfig.Protocol) {
+			// Routes can only be defined for HTTP services
+			return nil
+		}
+
+		for _, address := range svcConfig.Destination.Addresses {
+
+			routeName := clusterNameForDestination(cfgSnap, "~http", fmt.Sprintf("%d", svcConfig.Destination.Port), svcConfig.NamespaceOrDefault(), svcConfig.PartitionOrDefault())
```
DanStough, line 88: same comment here about consistency

DanStough (approved):

agent/xds/listeners.go
dhiaayachi: the nuance is that we don't use it in the sni for this case only the route name

DanStough: Fair enough!

dhiaayachi: and it avoid to have a special destination name that can't be used

kyhavlov (approved):

PR #13955, opened 2022-07-28
ci: Use pattern for CRT release branch config

eculver:
This adds a `release/**` pattern to match all of our release branches.

### Description
It was [recently announced](https://hashicorp.atlassian.net/wiki/spaces/ENSV/pages/2383447278/July+15+2022+-+Pattern+matching+now+supported+for+branch+names+in+ci.hcl) that pattern matching is now supported in CRT configuration for branch names. What this means is that with a pattern for our release branches we will no longer have to update `.release/ci.hcl` whenever we need to trigger CRT for a new release branch. 

This is one of the patterns listed in their examples, so I would imagine it works. We will know once we push a change to the new `release/1.13.x` branch.

Also, I triggered the workflows for my dev branch (technically `eculver/**` to express the wildcard case) which can be seen here: https://github.com/hashicorp/consul/pull/13955/checks.

### Links
- [CRT Newletter: July 15, 2022 - Pattern matching now supported for branch names in ci.hcl!](https://hashicorp.atlassian.net/wiki/spaces/ENSV/pages/2383447278/July+15+2022+-+Pattern+matching+now+supported+for+branch+names+in+ci.hcl)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,7 @@ on:
     branches:
       # Push events on the main branch
       - main
+      - release/**
 
 env:
   PKG_NAME: consul
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,7 @@ on:
       # Push events on the main branch
       - main
       - release/**
+      - eculver/**
 
 env:
   PKG_NAME: consul
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,6 @@ on:
       # Push events on the main branch
       - main
       - release/**
-      - eculver/**
 
 env:
   PKG_NAME: consul
--- a/.release/ci.hcl
+++ b/.release/ci.hcl
@@ -11,10 +11,7 @@ project "consul" {
     repository = "consul"
     release_branches = [
       "main",
-      "release/1.9.x",
-      "release/1.10.x",
-      "release/1.11.x",
-      "release/1.12.x",
+      "release/**",
     ]
   }
 }
--- a/.release/ci.hcl
+++ b/.release/ci.hcl
@@ -12,7 +12,6 @@ project "consul" {
     release_branches = [
       "main",
       "release/**",
-      "eculver/**",
     ]
   }
 }
--- a/.release/ci.hcl
+++ b/.release/ci.hcl
@@ -12,6 +12,7 @@ project "consul" {
     release_branches = [
       "main",
       "release/**",
+      "eculver/**",
     ]
   }
 }


eculver:
I decided to add the same pattern to our build workflow too for parity.

.github/workflows/build.yml
```diff
@@ -6,6 +6,8 @@ on:
     branches:
       # Push events on the main branch
       - main
+      - release/**
+      - eculver/**
```
dhiaayachi, line 10: maybe we can change this to something more generic ?
```suggestion
      - test-release/**
```

.release/ci.hcl
```diff
@@ -11,10 +11,8 @@ project "consul" {
     repository = "consul"
     release_branches = [
       "main",
-      "release/1.9.x",
-      "release/1.10.x",
-      "release/1.11.x",
-      "release/1.12.x",
+      "release/**",
+      "eculver/**",
```
dhiaayachi, line 15: same here I guess
```suggestion
      "test-release/**",
```

dhiaayachi (approved):
Just a small comment about naming, otherwise all good!

.github/workflows/build.yml
```diff
@@ -6,6 +6,7 @@ on:
     branches:
       # Push events on the main branch
       - main
+      - release/**
```
acpana, line 9: > we will no longer have to update .release/ci.hcl whenever we need to trigger CRT for a new release branch.

This sounds awesome 🎉 🥳 

I'm thinking about all the "older" release branches. Quick questions. Not just for this `build` wf but for the `ci` wfs, do we want to block releases older than `1.10` or `1.9` ?

acpana (approved):
Just a UX questions; non blocking; LGTM!

.github/workflows/build.yml
eculver: Oh, that was just to show the crowd (and for my own satisfaction) that the pattern matching was doing what it was supposed to. I removed mine from the list.

.github/workflows/build.yml
eculver: Yeah, you bring up a good question that I thought about a bit. Basically, I thought about writing a more strict pattern to match against that involved only matching our active release series, but this option has a couple downsides:
1. it's less readable
2. it still means we need to update whenever we have a new release series which defeats some of the value of having a wildcard at all
3. it doesn't really buy us much because these old release branches aren't being updated

So, with those bits in mind, I just went for the simpler, always up-to-date, less restrictive option that you're seeing here. It also means that if someone wants to create a hotfix branch like `release/1.11.3-myfixes` they can and get build artifacts right away. 

I'm open to discussing more, but those are my thoughts so far. Thanks for bringing it up.

Issue #13759, opened 2022-07-14
connect: Proxy defaults adds non connect-enabled services to ingress gateway


jmicceri:
#### Overview of the Issue

We're beginning to migrate all of our Consul services (1000+) into the service mesh. We use Nomad for our deployment tooling and have deployed an ingress gateway service in order to route to connect enabled services from outside of the mesh (Local users hitting service endpoints)

The ingress gateway listener is configured for all (`*`)services. As I understand it, with the wildcard listener that we have set an ingress gateway configured with the `http` protocol will automatically add all `http` services as upstreams to the gateway. Below is an excerpt from our deployed Nomad job definition where we configure the gateway and listener:

```
"Gateway": {
              "Proxy": {
                "ConnectTimeout": 5000000000,
                "EnvoyGatewayBindTaggedAddresses": false,
                "EnvoyGatewayBindAddresses": {
                  "*": {
                    "Address": "0.0.0.0",
                    "Port": 8443
                  }
                },
                "EnvoyGatewayNoDefaultBind": true,
                "EnvoyDNSDiscoveryType": "",
                "Config": {
                  "envoy_stats_bind_addr": "0.0.0.0:8404"
                }
              },
              "Ingress": {
                "TLS": {
                  "Enabled": true
                },
                "Listeners": [
                  {
                    "Port": 8443,
                    "Protocol": "http",
                    "Services": [
                      {
                        "Name": "*",
                        "Hosts": null
                      }
                    ]
                  }
                ]
              },
              "Terminating": null,
              "Mesh": null
            }
          },
          "
```

I know that by default Consul services are registered as `tcp` and as such they didn't get picked up as upstreams for the ingress gateway. To get around this, we applied the following proxy defaults configuration in our cluster so we didn't have to add a service defaults configuration for each service:
```
Kind      = "proxy-defaults"
Name      = "global"
Namespace = "default" # Can only be set to "default".
Config {
  protocol = "http"
  envoy_stats_bind_addr = "0.0.0.0:8404"
}
```

[...]

PR #13958, opened 2022-07-29
Fix wildcard picking up services it shouldn't for ingress/terminating gateways

kyhavlov:
This PR fixes an issue (https://github.com/hashicorp/consul/issues/13759) where a wildcard service definition in a gateway service config entry was causing the wrong type of services to be included for ingress and terminating gateways. Ingress gateways shouldn't pick up non-Connect services this way, and terminating gateways shouldn't pick up Connect services.

There's a couple changes to the service registration/gatewayServices table logic here:
- When a service is registered, it'll only be added to ingress wildcard mappings if it's the first connect instance, instead of any instance of the service. The reverse is true for terminating gateways - only non-connect instances will cause the mapping to get created.
- When a service is deregistered, it'll only remove ingress wildcard mappings if it's the last connect instance (and the reverse for terminating gateways).

### DIFF
--- a/agent/consul/state/catalog.go
+++ b/agent/consul/state/catalog.go
@@ -871,7 +871,7 @@ func ensureServiceTxn(tx WriteTxn, idx uint64, node string, preserveIndexes bool
 		if svc.Kind == structs.ServiceKindTypical && svc.Service != "consul" {
 			// Check if this service is covered by a gateway's wildcard specifier, we force the service kind to a gateway-service here as that take precedence
 			sn := structs.NewServiceName(svc.Service, &svc.EnterpriseMeta)
-			if err = checkGatewayWildcardsAndUpdate(tx, idx, &sn, structs.GatewayServiceKindService); err != nil {
+			if err = checkGatewayWildcardsAndUpdate(tx, idx, &sn, svc, structs.GatewayServiceKindService); err != nil {
 				return fmt.Errorf("failed updating gateway mapping: %s", err)
 			}
 			if err = checkGatewayAndUpdate(tx, idx, &sn, structs.GatewayServiceKindService); err != nil {
@@ -1984,11 +1984,6 @@ func (s *Store) deleteServiceTxn(tx WriteTxn, idx uint64, nodeName, serviceID st
 			if err := catalogUpdateServiceExtinctionIndex(tx, idx, entMeta, svc.PeerName); err != nil {
 				return err
 			}
-			if svc.PeerName == "" {
-				if err := cleanupGatewayWildcards(tx, idx, svc); err != nil {
-					return fmt.Errorf("failed to clean up gateway-service associations for %q: %v", name.String(), err)
-				}
-			}
 			psn := structs.PeeredServiceName{Peer: svc.PeerName, ServiceName: name}
 			if err := freeServiceVirtualIP(tx, idx, psn, nil); err != nil {
 				return fmt.Errorf("failed to clean up virtual IP for %q: %v", name.String(), err)
@@ -1999,6 +1994,12 @@ func (s *Store) deleteServiceTxn(tx WriteTxn, idx uint64, nodeName, serviceID st
 		}
 	} else {
 		return fmt.Errorf("Could not find any service %s: %s", svc.ServiceName, err)
+	}
+
+	if svc.PeerName == "" {
+		if err := cleanupGatewayWildcards(tx, idx, svc); err != nil {
+			return fmt.Errorf("failed to clean up gateway-service associations for %q: %v", name.String(), err)
+		}
 	}
 
 	return nil
@@ -3652,6 +3653,18 @@ func updateGatewayNamespace(tx WriteTxn, idx uint64, service *structs.GatewaySer
 			continue
 		}
 
+		supportsIngress, supportsTerminating, err := serviceConnectInstances(tx, sn.ServiceName, entMeta)
+		if err != nil {
+			return err
+		}
+
+		if service.GatewayKind == structs.ServiceKindIngressGateway && !supportsIngress {
+			continue
+		}
+		if service.GatewayKind == structs.ServiceKindTerminatingGateway && !supportsTerminating {
+			continue
+		}
+
 		existing, err := tx.First(tableGatewayServices, indexID, service.Gateway, sn.CompoundServiceName(), service.Port)
 		if err != nil {
 			return fmt.Errorf("gateway service lookup failed: %s", err)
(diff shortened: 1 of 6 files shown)


agent/consul/state/catalog.go
```diff
@@ -3717,6 +3730,42 @@ func updateGatewayNamespace(tx WriteTxn, idx uint64, service *structs.GatewaySer
...
+	}
+	if svc != nil {
+		hasConnectInstance = true
+	}
+
+	hasNonConnectInstance := false
+	nonConnectQuery := Query{
+		Value:          serviceName,
+		EnterpriseMeta: *entMeta,
+	}
```
huikang, line 3753: Looks like the `nonConnectQuery` is identical to connectQuery (line 3737). Any reason to have 2 variables?

agent/consul/state/catalog.go
```diff
@@ -3717,6 +3730,42 @@ func updateGatewayNamespace(tx WriteTxn, idx uint64, service *structs.GatewaySer
 	return nil
 }
 
+// serviceConnectInstances returns whether the service has at least one connect instance,
+// and at least one non-connect instance.
+func serviceConnectInstances(tx WriteTxn, serviceName string, entMeta *acl.EnterpriseMeta) (bool, bool, error) {
```
huikang, line 3735: It is a little difficult to connect the function name to what it does. So how about renaming it like `serviceHasConnectInstances`?

agent/consul/state/catalog.go
```diff
@@ -3754,14 +3803,31 @@ func updateGatewayService(tx WriteTxn, idx uint64, mapping *structs.GatewayServi
...
 // gateway-services table.
-func checkGatewayWildcardsAndUpdate(tx WriteTxn, idx uint64, svc *structs.ServiceName, kind structs.GatewayServiceKind) error {
+func checkGatewayWildcardsAndUpdate(tx WriteTxn, idx uint64, svc *structs.ServiceName, ns *structs.NodeService, kind structs.GatewayServiceKind) error {
 	sn := structs.ServiceName{Name: structs.WildcardSpecifier, EnterpriseMeta: svc.EnterpriseMeta}
 	svcGateways, err := tx.Get(tableGatewayServices, indexService, sn)
 	if err != nil {
 		return fmt.Errorf("failed gateway lookup for %q: %s", svc.Name, err)
 	}
+
+	supportsIngress, supportsTerminating, err := serviceConnectInstances(tx, svc.Name, &svc.EnterpriseMeta)
```
huikang, line 3813: Shall we name these two variables `hasConnectInstance, hasNonConnectInstance` as in line 3891

agent/consul/state/catalog.go
```diff
@@ -3717,6 +3730,42 @@ func updateGatewayNamespace(tx WriteTxn, idx uint64, service *structs.GatewaySer
...
+		Value:          serviceName,
+		EnterpriseMeta: *entMeta,
+	}
+	iter, err := tx.Get(tableServices, indexService, nonConnectQuery)
+	if err != nil {
+		return false, false, fmt.Errorf("failed service lookup: %s", err)
+	}
+	for service := iter.Next(); service != nil; service = iter.Next() {
+		sn := service.(*structs.ServiceNode)
+		if !sn.ServiceConnect.Native {
```
huikang, line 3760: Just out of curiosity, is envoy-based proxy service a type of native connect(make `sn.ServiceConnect.Native=true`)? (Asking because I remember there is something called Connect-Native app)

kyhavlov: Yes, a service is 'connect-enabled' if it either has a sidecar connect proxy, or if it has `ServiceConnect.Native = true` indicating it can handle connect requests/auth itself.

agent/consul/state/catalog.go
```diff
@@ -3652,6 +3653,18 @@ func updateGatewayNamespace(tx WriteTxn, idx uint64, service *structs.GatewaySer
...
 
+		supportsIngress, supportsTerminating, err := serviceConnectInstances(tx, sn.ServiceName, entMeta)
+		if err != nil {
+			return err
+		}
+
+		if service.GatewayKind == structs.ServiceKindIngressGateway && !supportsIngress {
+			continue
+		}
+		if service.GatewayKind == structs.ServiceKindTerminatingGateway && !supportsTerminating {
```
dhiaayachi, line 3664: I'm not quiet sure I understand this. I'm wondering how that would work with destinations, which are service-defaults without service registration and that could be used with terminating-gateways?

agent/consul/state/catalog.go
```diff
@@ -3754,14 +3803,31 @@ func updateGatewayService(tx WriteTxn, idx uint64, mapping *structs.GatewayServi
...
 	}
+
+	supportsIngress, supportsTerminating, err := serviceConnectInstances(tx, svc.Name, &svc.EnterpriseMeta)
+	if err != nil {
+		return err
+	}
+	if ns != nil && ns.Connect.Native {
+		supportsIngress = true
+	} else {
+		supportsTerminating = true
```
dhiaayachi, line 3820: if ns == nil we will have  supportsTerminating = true, is that the wanted behaviour here?

agent/consul/state/catalog.go
```diff
@@ -3818,12 +3884,28 @@ func cleanupGatewayWildcards(tx WriteTxn, idx uint64, svc *structs.ServiceNode)
...
 	// Do the updates in a separate loop so we don't trash the iterator.
 	for _, m := range mappings {
 		// Only delete if association was created by a wildcard specifier.
 		// Otherwise the service was specified in the config entry, and the association should be maintained
 		// for when the service is re-registered
 		if m.FromWildcard {
+			if m.GatewayKind == structs.ServiceKindIngressGateway && hasConnectInstance {
+				continue
+			}
+			if m.GatewayKind == structs.ServiceKindTerminatingGateway && hasNonConnectInstance {
```
dhiaayachi, line 3905: same comment here about destinations

dhiaayachi (commented):
Thank you for working on this @kyhavlov.
I had few questions comments mostly about the impact of this on destinations.

agent/consul/state/catalog.go
kyhavlov: Do destinations get picked up by a wildcard entry in a gateway currently? If so, we'd need to make an exception somehow for services with a `service-defaults` config entry set.

agent/consul/state/catalog.go
kyhavlov: No, thanks for catching that, it should be `else if ns != nil`.

agent/consul/state/catalog.go
kyhavlov: I added an exception for terminating gateways here when the service kind passed in is `GatewayServiceKindDestination` - that should mean the behavior of destinations won't be affected by this change.

dhiaayachi: Shouldn't we add the check for `GatewayServiceKindDestination` in all the places we currently test on `structs.ServiceKindTerminatingGateway && !supportsTerminating`?

agent/consul/state/catalog.go
dhiaayachi: I think that's another spot we need to exclude destination even if `hasNonConnectInstance` is false

dhiaayachi (changes requested):
I think there is at least to other spots that we need to add the check for `GatewayServiceKindDestination` 
Also, If possible to add a test variation for destination but I'm ok with adding this in a subsequent PR.

agent/consul/state/catalog.go
kyhavlov: 👍 Updated this spot and added a couple tests for deleting a destination that also involve having an instance of the service registered.

dhiaayachi (approved):
LGTM!

agent/consul/state/catalog.go
```diff
@@ -3818,12 +3888,44 @@ func cleanupGatewayWildcards(tx WriteTxn, idx uint64, svc *structs.ServiceNode)
...
 	// Do the updates in a separate loop so we don't trash the iterator.
 	for _, m := range mappings {
 		// Only delete if association was created by a wildcard specifier.
 		// Otherwise the service was specified in the config entry, and the association should be maintained
 		// for when the service is re-registered
 		if m.FromWildcard {
+			if m.GatewayKind == structs.ServiceKindIngressGateway && hasConnectInstance {
+				continue
+			}
+			if m.GatewayKind == structs.ServiceKindTerminatingGateway && (hasNonConnectInstance || hasDestination) {
```
kirooshu, line 3925: Hello, I think this logic here needs to be 

```
if m.GatewayKind == structs.ServiceKindTerminatingGateway && (hasNonConnectInstance || hasDestination) && !hasConnectInstance {
```

The reason being is that if you create services from Nomad, they are already registered as both `service` and `connect` service. This logic here just works really for external services (`hasDestination`) but does not help at all for services that are created from Nomad.

Example:


```
  service {
      name = "api-admin"
      port = "80"

      connect {
        sidecar_service {}
      }

      check {
        expose         = true
        initial_status = "critical"
        name           = "http"
        type           = "http"
        interval       = "10s"
        timeout        = "2s"
        path           = "/_status"
      }
    }
```

kyhavlov: @kirooshu Are you saying a connect instance like that gets picked up by a terminating gateway with a wildcard entry, but shouldn't be? Or the other way around?

kirooshu: The former.

PR #13967, opened 2022-07-29
docs: link pq docs to relevant DNS lookup section

jkirschner-hashicorp:
Minor improvement: link directly to the relevant DNS lookup section for prepared queries, rather than the long DNS overview page.

### DIFF
--- a/website/content/api-docs/query.mdx
+++ b/website/content/api-docs/query.mdx
@@ -11,7 +11,7 @@ The `/query` endpoints create, update, destroy, and execute prepared queries.
 Prepared queries allow you to register a complex service query and then execute
 it later via its ID or name to get a set of healthy nodes that provide a given
 service. This is particularly useful in combination with Consul's
-[DNS Interface](/docs/discovery/dns) as it allows for much richer queries than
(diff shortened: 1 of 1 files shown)


website/content/api-docs/query.mdx
```diff
@@ -11,7 +11,7 @@ The `/query` endpoints create, update, destroy, and execute prepared queries.
 Prepared queries allow you to register a complex service query and then execute
 it later via its ID or name to get a set of healthy nodes that provide a given
 service. This is particularly useful in combination with Consul's
-[DNS Interface](/docs/discovery/dns) as it allows for much richer queries than
+[DNS Interface](/docs/discovery/dns#prepared-query-lookups) as it allows for much richer queries than
```
trujillo-adam, line 14:
```suggestion
[DNS Interface](/docs/discovery/dns#prepared-query-lookups) because it enables more specific queries than
```
We might as well update the description with more concrete language while we're at it.

trujillo-adam (approved):
I suggested an additional change, but approving as is.

PR #13979, opened 2022-08-01
Various peering fixes

freddygv:
### Description
* Avoid logging StreamSecretID
* Wrap additional errors in stream handler
* Fix flakiness in leader test and rename servers for clarity. There was
  a race condition where the peering was being deleted in the test
  after the initial handling in `StreamResources` but before the stream was active. Now the test waits for the stream to be
  connected on both sides before deleting the associated peering.

* Spotted the secret in peering leader unit test logs

### Links
`StreamSecretID` was introduced in #13977

### PR Checklist

* [x] updated test coverage
* [ ] not a security concern

### DIFF
--- a/agent/consul/leader_peering.go
+++ b/agent/consul/leader_peering.go
@@ -237,7 +237,7 @@ func (s *Server) syncPeeringsAndBlock(ctx context.Context, logger hclog.Logger, 
 		}
 	}
 
-	logger.Trace("checking connected streams", "streams", s.peerStreamServer.ConnectedStreams(), "sequence_id", seq)
+	logger.Trace("checking connected streams", "streams", connectedStreams, "sequence_id", seq)
 
 	// Clean up active streams of peerings that were deleted from the state store.
 	// TODO(peering): This is going to trigger shutting down peerings we generated a token for. Is that OK?
--- a/agent/consul/leader_peering_test.go
+++ b/agent/consul/leader_peering_test.go
@@ -41,8 +41,8 @@ func testLeader_PeeringSync_Lifecycle_ClientDeletion(t *testing.T, enableTLS boo
 		t.Skip("too slow for testing.Short")
 	}
 
-	_, s1 := testServerWithConfig(t, func(c *Config) {
-		c.NodeName = "bob"
+	_, acceptor := testServerWithConfig(t, func(c *Config) {
+		c.NodeName = "acceptor"
 		c.Datacenter = "dc1"
 		c.TLSConfig.Domain = "consul"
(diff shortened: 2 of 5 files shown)


agent/consul/leader_peering.go
```diff
@@ -237,7 +237,7 @@ func (s *Server) syncPeeringsAndBlock(ctx context.Context, logger hclog.Logger,
 		}
 	}
 
-	logger.Trace("checking connected streams", "streams", s.peerStreamServer.ConnectedStreams(), "sequence_id", seq)
+	logger.Trace("checking connected streams", "streams", connectedStreams, "sequence_id", seq)
```
freddygv, line 240: This was logging against a new fetch of `ConnectedStreams()` rather than the `connectedStreams` var that we actually reconcile against below

agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -684,10 +686,20 @@ func logTraceProto(logger hclog.Logger, pb proto.Message, received bool) {
 		dir = "received"
 	}
 
+	// Redact the long-lived stream secret to avoid leaking it in trace logs.
+	pbToLog := pb
```
freddygv, line 690: Will add a test for this actually

freddygv:
Looking into the test failure

agent/consul/leader_peering_test.go
```diff
@@ -94,55 +94,61 @@ func testLeader_PeeringSync_Lifecycle_ClientDeletion(t *testing.T, enableTLS boo
...
 
 	retry.Run(t, func(r *retry.R) {
-		status, found := s2.peerStreamServer.StreamStatus(p.Peering.ID)
+		status, found := dialer.peerStreamServer.StreamStatus(p.Peering.ID)
+		require.True(r, found)
+		require.True(r, status.Connected)
+	})
+
+	retry.Run(t, func(r *retry.R) {
+		status, found := acceptor.peerStreamServer.StreamStatus(p.Peering.PeerID)
```
freddygv, line 129: This was the change to wait for the stream to be connected before deleting

freddygv: Though there's still a race that can happen outside of tests. The issue is this sequence:
1. `StreamResources` handler is invoked on acceptor, and peering exists, so the connection is allowed
2. Peering is deleted on acceptor
3. HandleStream begins on acceptor

Now the stream is tracked but won't be torn down until the leader routine wakes up due to a change in the list of peerings

mkeeler (approved):

PR #13993, opened 2022-08-01
Separate tag from protocol in RFC 2782 DNS lookup

### LINKED ISSUE #11046: Inject tags into SRV records
#### Feature Description

Currently, SRV records for a service can only be separated out by changing the service name. This is not possible in some cases (for instance, Puppet uses _x-puppet._tcp.[srv_domain] for service discovery, and there is no way to change anything other than the [srv_domain] portion). Consul already allows for getting A records in the form of <servicename>.<.tag>.service<.domain>, doing the same with automatic SRV records would bring this into parity with the A records and allow for services to separate clusters.

#### Use Case(s)

This can be used any time multiple separate clusters need provide the same service but must be kept separate during discovery. In the example above of using Puppet, when upgrading from one major Puppet version to another, the ability to tag the version number that a node is providing would make it so clients can be moved between the older and newer revision in a controlled manner.

jkirschner-hashicorp:
### Description

Closes #11046 (though ideally want to confirm with the originally poster of that enhancement request)

Previously, to perform tag-based filtering with an RFC 2782 style DNS lookup,
the tag needed to be specified in the RFC 2782 protocol field.
Some external systems require the protocol field to specify the protocol
(e.g., TCP) and do not allow using an arbitrary value (e.g., service tag).

This commit allows the tag to be specified separately from the protocol field:

	_<service>._<protocol>.<tag>.service[.<datacenter>].<domain>

### TODO

- [ ] Add changelog entry
- [ ] Attempt to confirm the utility of this change with original issue poster

See new unit tests

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/dns.go
```diff
@@ -764,26 +764,43 @@ func (d *DNSServer) dispatch(remoteAddr net.Addr, req, resp *dns.Msg, maxRecursi
...
 
-			// Grab the tag since we make nuke it if it's tcp
-			tag := queryParts[1][1:]
+		tag := ""
 
-			// Treat _name._tcp.service.consul as a default, no need to filter on that tag
-			if tag == "tcp" {
-				tag = ""
+		// If the first two query parts begin with _, this is an RFC 2782 style SRV lookup.
+		if strings.HasPrefix(queryParts[0], "_") && strings.HasPrefix(queryParts[1], "_") && n >= 2 {
```
jkirschner-hashicorp, line 771: Note: I changed the order of these if-conditional checks. I'm not sure why we'd check whether the 1st query part starts with `_` before checking the 0th.

I also changed `n == 2` to `n >= 2`. The `n == 2` codepath is unchanged. But the `n >= 3` codepath is new.

website/content/docs/discovery/dns.mdx
```diff
@@ -52,7 +52,7 @@ There are fundamentally two types of queries: node lookups and service lookups.
 A node lookup, a simple query for the address of a named node, looks like this:
 
 ```text
-<node>.node[.datacenter].<domain>
+<node>.node[.<datacenter>].<domain>
```
jkirschner-hashicorp, line 55: Most of these small DNS docs improvements (e.g., clarifying which values are variables by wrapping with `<>`) will be moved into a separate PR.

agent/dns_test.go
```diff
@@ -3556,40 +3556,48 @@ func TestDNS_ServiceLookup_TagPeriod(t *testing.T) {
...
 	}
 
-	m := new(dns.Msg)
-	m.SetQuestion("v1.primary.db.service.consul.", dns.TypeSRV)
-
-	c := new(dns.Client)
-	in, _, err = c.Exchange(m, a.DNSAddr())
-	if err != nil {
-		t.Fatalf("err: %v", err)
+	validQuestions := []string{
```
jkirschner-hashicorp, line 3559: Modified this test to use an array of questions, including the previously-existing standard lookup and the new RFC 2782 lookup format for tag-based filtering

website/content/docs/discovery/dns.mdx
```diff
@@ -168,15 +168,24 @@ foobar.node.dc1.consul.	0	IN	A	10.1.10.12
 
 ### RFC 2782 Lookup
 
-The format for RFC 2782 SRV lookups is:
+Valid formats for RFC 2782 SRV lookups are:
 
-    _<service>._<protocol>[.service][.datacenter][.domain]
+```shell-session
+# 1: No filtering on tag
```
jkirschner-hashicorp, line 174: I need to make these lines show up as comments. Right now, they look like regular content/text/code.

jkirschner-hashicorp: Changed format

website/content/docs/discovery/dns.mdx
jkirschner-hashicorp: Refer to https://github.com/hashicorp/consul/pull/13999

jkirschner-hashicorp:
@boruszak : thanks for the feedback! I merged the other DNS-related PR so this docs diff is now much smaller. I've incorporated your feedback.

jkirschner-hashicorp:
@boruszak : don't merge this. The eng team needs to review the code changes first.

jkirschner-hashicorp:
Still need an engineering review

agent/dns.go
```diff
@@ -764,26 +764,42 @@ func (d *DNSServer) dispatch(remoteAddr net.Addr, req, resp *dns.Msg, maxRecursi
...
-		if n == 2 && strings.HasPrefix(queryParts[1], "_") && strings.HasPrefix(queryParts[0], "_") {
 
-			// Grab the tag since we make nuke it if it's tcp
-			tag := queryParts[1][1:]
+		tag := ""
+
+		// If the first two query parts begin with _, this is an RFC 2782 style SRV lookup.
+		if strings.HasPrefix(queryParts[0], "_") && strings.HasPrefix(queryParts[1], "_") && n >= 2 {
+			// Support RFC 2782 style syntax in one of the following forms:
+			// 1) _<service-name>._<tag or tcp>[.service][.<datacenter>].<domain>    (n == 2)
```
jkirschner-hashicorp, line 773: Is [.service] actually optional? The docs say yes ... but I want to double-check that in the code.

agent/dns.go
```diff
@@ -764,26 +764,42 @@ func (d *DNSServer) dispatch(remoteAddr net.Addr, req, resp *dns.Msg, maxRecursi
...
 		}
-		// Support RFC 2782 style syntax
-		if n == 2 && strings.HasPrefix(queryParts[1], "_") && strings.HasPrefix(queryParts[0], "_") {
 
-			// Grab the tag since we make nuke it if it's tcp
-			tag := queryParts[1][1:]
+		tag := ""
+
+		// If the first two query parts begin with _, this is an RFC 2782 style SRV lookup.
+		if strings.HasPrefix(queryParts[0], "_") && strings.HasPrefix(queryParts[1], "_") && n >= 2 {
```
jkirschner-hashicorp, line 771: Need to confirm queryParts has at least 2 elements before indexing into it

```suggestion
		if n >= 2 && strings.HasPrefix(queryParts[0], "_") && strings.HasPrefix(queryParts[1], "_")  {
```

agent/dns.go
jkirschner-hashicorp: Confirmed that [.service] is optional for this first RFC 2782 style lookup format.

PR #13997, opened 2022-08-01
cleanup agents and servers within the tests

acpana:
This is a bit of a different PR, meant to address flaky-ness due to setup issues in our tests.

#### Hypothesis

A number of flakes occur for “integration” or end-to-end tests that need a free port to stand up agents error out when there is no free port. The retry-able error will look something like:

```
pending ports are still in use; something probably didn't wait around for the port to be closed!
```

Eventually, the test fails since it can't get a free port.

#### Cause

The cause for this is an ordering/ data flow issue. Every test T is a child of the testing T instance ( [ref](https://cs.opensource.google/go/go/+/master:src/testing/testing.go;l=1256-1272;drc=f32519e5fbcf1b12f9654a6175e5e72b09ae8f3a) ). Parallel tests run to completion before calling `t.cleanup` ( [ref](https://cs.opensource.google/go/go/+/master:src/testing/testing.go;l=1404-1415;drc=f32519e5fbcf1b12f9654a6175e5e72b09ae8f3a) ). It would be best to `defer a.Shutdown()` and other clean up tasks as part of the test itself so these resources, like ports in use, get freed up faster.

### DIFF
--- a/agent/consul/server_test.go
+++ b/agent/consul/server_test.go
@@ -1211,7 +1211,7 @@ func TestServer_RPC_MetricsIntercept_Off(t *testing.T) {
 		if err != nil {
 			t.Fatalf("err: %v", err)
 		}
-		t.Cleanup(func() { s1.Shutdown() })
+		defer s1.Shutdown()
 
 		var out struct{}
 		if err := s1.RPC("Status.Ping", struct{}{}, &out); err != nil {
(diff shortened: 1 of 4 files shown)


mkeeler:
I agree with the premise but differ in opinion on what the solution should be.

Instead of avoiding t.Cleanup we should be avoiding parallel tests as they cause all sorts of issues.

boxofrad:
I'm sort of torn on this. On the one hand, it makes sense for test cases to release ports as soon as they're done with them, but I understand Matt's point about the inevitable resource contention running integration-style tests in parallel creates.

Given we have quite a lot (too many?) of this kind of test, I'm guessing that removing `t.Parallel` would significantly slow our suite down, but I wonder if that'd be paid for by the increased stability meaning we don't need to re-run the suite as often?

acpana:
@kisunji I'm going to close this for now -- feel free to reopen as needed.

I still think that it makes sense to cleanup resources before the test "completes" but will leave that up to the team.

PR #13998, opened 2022-08-02
Add new envoy tracing configuration

jorgemarey:
### Description

The current tracing configuration provided to envoy is [deprecated](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/bootstrap/v3/bootstrap.proto#envoy-v3-api-msg-config-bootstrap-v3-bootstrap). Should be replaced by the [HttpConnectionManager.Tracing](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-msg-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-tracing) configuration. That also allows more custom config as allowing us to add custom tags. This is an example of a valid config JSON.

```
{
    "@type" : "type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager.Tracing",
    "provider" : {
        "name" : "envoy.tracers.zipkin",
        "typed_config" : {
            "@type" : "type.googleapis.com/envoy.config.trace.v3.ZipkinConfig",
            "collector_cluster" : "customcolector",
            "collector_endpoint" : "/api/v2/spans",
            "collector_endpoint_version" : "HTTP_JSON",
            "shared_span_context" : false
        }
    },
    "custom_tags" : [
        {
            "tag" : "custom_tag",
            "request_header" : {
                "name" : "x-tag",
                "default_value" : ""
            }
        },
        {
            "tag" : "alloc_id",
            "environment" : {
                "name" : "NOMAD_ALLOC_ID"
            }
        }
    ]
}
```

This PR adds a new ListenerTracingJSON that's configured in all listeners (minus in the exposed checks). The old tracing option is not removed for compatiblity.

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/.changelog/13998.txt
+++ b/.changelog/13998.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+connect: change deprecated tracing configuration on envoy
+```
--- a/.changelog/13998.txt
+++ b/.changelog/13998.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-connect: change deprecated tracing configuration on envoy
+connect: expose new tracing configuration on envoy
 ```
--- a/agent/xds/config.go
+++ b/agent/xds/config.go
@@ -27,6 +27,12 @@ type ProxyConfig struct {
 	// Note: This escape hatch is compatible with the discovery chain.
 	PublicListenerJSON string `mapstructure:"envoy_public_listener_json"`
 
+	// LstenerTracingJSON is a complete override ("escape hatch") for the
+	// listeners tracing configuration.
+	//
+	// Note: This escape hatch is compatible with the discovery chain.
+	LstenerTracingJSON string `mapstructure:"envoy_listener_tracing_json"`
+
 	// LocalClusterJSON is a complete override ("escape hatch") for the
 	// local application cluster.
 	//
--- a/agent/xds/config.go
+++ b/agent/xds/config.go
@@ -27,11 +27,11 @@ type ProxyConfig struct {
 	// Note: This escape hatch is compatible with the discovery chain.
 	PublicListenerJSON string `mapstructure:"envoy_public_listener_json"`
 
-	// LstenerTracingJSON is a complete override ("escape hatch") for the
+	// ListenerTracingJSON is a complete override ("escape hatch") for the
 	// listeners tracing configuration.
 	//
 	// Note: This escape hatch is compatible with the discovery chain.
-	LstenerTracingJSON string `mapstructure:"envoy_listener_tracing_json"`
+	ListenerTracingJSON string `mapstructure:"envoy_listener_tracing_json"`
 
 	// LocalClusterJSON is a complete override ("escape hatch") for the
 	// local application cluster.
--- a/website/content/docs/connect/proxies/envoy.mdx
+++ b/website/content/docs/connect/proxies/envoy.mdx
@@ -759,6 +759,45 @@ definition](/docs/connect/registration/service-registration) or
   </CodeTabs>
 
 
+- `envoy_listener_tracing_json` - Specifies a [tracing 
+  configuration](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-msg-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-tracing)
+  to be inserter in the public and upstreams listeners of the proxy.
+
+  <CodeBlockConfig heading="Example envoy_listener_tracing_json">
+
+  ```json
+  {
+    "@type" : "type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager.Tracing",
+    "provider" : {
+      "name" : "envoy.tracers.zipkin",
+      "typed_config" : {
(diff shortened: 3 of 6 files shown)


agent/xds/config.go
```diff
@@ -27,6 +27,12 @@ type ProxyConfig struct {
 	// Note: This escape hatch is compatible with the discovery chain.
 	PublicListenerJSON string `mapstructure:"envoy_public_listener_json"`
 
+	// LstenerTracingJSON is a complete override ("escape hatch") for the
+	// listeners tracing configuration.
```
kisunji, line 31:
```suggestion
	// upstream listener's tracing configuration.
```

agent/xds/config.go
```diff
@@ -27,6 +27,12 @@ type ProxyConfig struct {
 	// Note: This escape hatch is compatible with the discovery chain.
 	PublicListenerJSON string `mapstructure:"envoy_public_listener_json"`
```
kisunji, line 28: Couldn't we define the custom tracing json as part of the larger override?

What's the use-case for targeting the tracing configuration only?

kisunji (commented):
The `envoy_public_listener_json` should be able to define an override for tracing already; what's the reason for escape hatching the tracing configuration only?

Currently it's hard to maintain a good encapsulation for envoy configurations and my concern would be that this sets a precedent for other "escape hatch for X field" PRs.

jorgemarey:
Hi @kisunji, thanks for taking the time to look into this. I'll try to explain the reasoning behind this. 

> The `envoy_public_listener_json` should be able to define an override for tracing already; what's the reason for escape hatching the tracing configuration only?

When we were implementing tracing in our consul setup, we needed a way to add some nomad tags to get information related to deployment and also being able to send a custom header to the collector due to how our tracing platform works. In the specification it's used a custom header to inform the traceID, so all applications deployed are using that header instead of the common ones. 

When looking how we could do this into consul, we saw the `envoy_tracing_json` configuration option, and thought that maybe we could add there the [envoy tracing configuration](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-msg-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-tracing), but digging into it, we found out that the needed tracing specification is old (and [deprecated](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/bootstrap/v3/bootstrap.proto#envoy-v3-api-msg-config-bootstrap-v3-bootstrap)) and doesn't allow setting custom_tags. So I thought about doing a PR and updating the envoy tracing escape-hatch to be the new tracing provided by envoy instead of the old one, that's going to be removed at some point (in the PR the old remains for compatibility until it's removed in a future release)

Regarding the `envoy_public_listener_json` option. I guess it could work with that, but having to set this over a lot for services having each one several instances (we're deploying using nomad), I don't know how to work around the `port_value` for each one. Also that we would need to specify the same for all the configured upstreams with a `envoy_listener_json` (I don't know if that can be specified several times) so that outgoing connections are also traced. This makes the configuration a lot more complicated than this. With this we can just set this configuration in a `proxy-defaults` for all the services (or in a specific service but avoiding the public and upstreams listener configuration).

> Currently it's hard to maintain a good encapsulation for envoy configurations and my concern would be that this sets a precedent for other "escape hatch for X field" PRs.

I get this, in most cases it should be enough with the already existent escape-hatchs, but this is created as a substitution of a deprecated feature providing more functionality.

agent/xds/config.go
jorgemarey: This won't only affect the upstreams listeners but also the application listener. Should I make this change?

kisunji: Ah, I missed that. Disregard this then

.changelog/13998.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+connect: change deprecated tracing configuration on envoy
```
kisunji, line 2:
```suggestion
connect: expose new tracing configuration on envoy 
```

kisunji (commented):
Would you mind rebasing from `main`? There was an issue with integration tests on forked PRs which should have been resolved in main now.

Once someone from our docs team approves, I can get this merged.

website/content/docs/connect/proxies/envoy.mdx
```diff
@@ -759,6 +759,45 @@ definition](/docs/connect/registration/service-registration) or
   </CodeTabs>
 
 
+- `envoy_listener_tracing_json` - Specifies a [tracing 
+  configuration](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-msg-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-tracing)
+  to be inserter in the public and upstreams listeners of the proxy.
```
im2nguyen, line 764:
```suggestion
  to be inserted in the proxy's public and upstreams listeners.
```

im2nguyen (approved):
small nit for clarity, but overall looks good!

kisunji:
This should go out in our next patch release of 1.13!

PR #14007, opened 2022-08-02
ui: peering UI fixes - api contract change / wrong link in peerings list

LevelbossMike:
### Description

This PR fixes some things that broke the peering UI

* Api contract was changed - we can't send the `Datacenter` attribute anymore when creating a token / establishing a peer
* We were still surfacing a links to `peers.edit` which was breaking the list and surfacing an error whenever we established a peer.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-peerings/app/components/consul/peer/list/index.hbs
+++ b/ui/packages/consul-peerings/app/components/consul/peer/list/index.hbs
@@ -42,14 +42,6 @@ as |item index|>
 {{#if (can 'delete peer' item=item)}}
 
     <Actions as |Action|>
-      <Action
-        data-test-edit-action
-        @href={{href-to 'dc.peers.edit' item.Name}}
-      >
-        <BlockSlot @name="label">
-          View
-        </BlockSlot>
-      </Action>
 {{#if (can "write peer" item=item)}}
         <Action
           data-test-edit-action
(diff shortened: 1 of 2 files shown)


mkeeler (approved):

johncowen (approved):
Few things to note here:

1. My `prettier` setup has obvs gone squewiff 😬 , I'll take a look a that shortly.
2. Looks like one change is the removal of the `Datacenter` property from the POST body, that looks fine to me!
3. The [View] action was already removed here https://github.com/hashicorp/consul/pull/13836/files#diff-3905c75810dc20a3d84bb9e0c7f73f4531fedc61c652c28932e2a3eb449ff2cdL55 so whilst what is done here is fine. I'm a little confused at how it ended up back in here. It could be just from a bad rebase, but I'll take a look at that separately now.

---

I just took a quick look into point 3 ^ here. Looks like those PRs where opened in this order:

1. https://github.com/hashicorp/consul/pull/13813
2. https://github.com/hashicorp/consul/pull/13836

but approved and merged in this (the reverse) order, guessing with a rebase in there somewhere:

1. https://github.com/hashicorp/consul/pull/13836
2. https://github.com/hashicorp/consul/pull/13813

So the [View] button was added back in as the first PR took longer to go in than the second 🤦 , a bit of toe stepping basically.

---

All in all changes here look good to me

johncowen:
Just to note, I added a no changelog here also. I think that's correct, but feel free to remove if not.

PR #13950, opened 2022-07-28
add config parameters doc for destination in service-defaults

dhiaayachi:
### Description
This adds documentation around destination feature.

PR #14008, opened 2022-08-02
docs: destination docs for k8s

DanStough:
### Description
Adding Kubernetes documentation for the new transparent proxy Destination feature.

N/A

### Links
Additional Doc PR #13950 

### PR Checklist

* [ ] ~updated test coverage~
* [X] not a security concern

### DIFF
--- a/website/content/docs/k8s/connect/terminating-gateways.mdx
+++ b/website/content/docs/k8s/connect/terminating-gateways.mdx
@@ -88,6 +88,58 @@ Registering the external services with Consul is a multi-step process:
 
 ### Register external services with Consul
 
+There are two ways to register an external service with Consul:
+1. If [`TransparentProxy`](/docs/k8s/helm#v-connectinject-transparentproxy) is enabled, you can declare external endpoints in the [`Destination`](/docs/connect/config-entries/service-defaults#terminating-gateway-destination) field of `service-defaults`.
+1. You can add the service as a node in the Consul catalog.
+
+#### Register an external service as a Destination
+
+`Destination` fields allow clients to dial the external service directly and are valid only in [`TransparentProxy`](/docs/k8s/helm#v-connectinject-transparentproxy) mode.
+The following table describes traffic behaviors when using `Destination`s to route traffic through a terminating gateway:
+
+| External Services Layer | Client dials | Client uses TLS | Allowed | Notes |
+|---|---|---|---|---|
+| L4 | Hostname | Yes | Allowed | `CAFiles` are not allowed because traffic is already end-to-end encrypted by the client. |
+| L4 | IP | Yes | Allowed | `CAFiles` are not allowed because traffic is already end-to-end encrypted by the client. |
+| L4 | Hostname | No | Not allowed | The sidecar is not protocol aware and can not identify traffic going to the external service. |
+| L4 | IP | No | Allowed | There are no limitations on dialing IPs without TLS. |
+| L7 | Hostname | Yes | Not allowed | Because traffic is already encrypted before the sidecar, it cannot route as L7 traffic. |
+| L7 | IP | Yes | Not allowed | Because traffic is already encrypted before the sidecar, it cannot route as L7 traffic. |
+| L7 | Hostname | No | Allowed | A `Host` or `:authority` header is required. |
+| L7 | IP | No | Allowed | There are no limitations on dialing IPs without TLS.  |
+
+You can provide a `caFile` to secure traffic between unencrypted clients that connect to external services through the terminating gateway.
+Refer to [Create the configuration entry for the terminating gateway](/docs/k8s/connect/terminating-gateways#create-the-configuration-entry-for-the-terminating-gateway) for details.
+
+Create a `service-defaults` custom resource for the external service:
+
+<CodeBlockConfig filename="service-defaults.yaml">
+
+```yaml
+  apiVersion: consul.hashicorp.com/v1alpha1
+  kind: ServiceDefaults
+  metadata:
+    name: example-https
+  spec:
+    protocol: tcp
+    destination:
+      addresses:
+        - "example.com"
+      port: 443
+```
+
+</CodeBlockConfig>
+
+Apply the `ServiceDefaults` resource with `kubectl apply`:
+
+```shell-session
+$ kubectl apply --filename service-defaults.yaml
+```
+
+All other terminating gateway operations can use the name of the `service-defaults` in place of a typical Consul service name.
+
+#### Register an external service as a Catalog Node
+
 -> **Note:** Normal Consul services are registered with the Consul client on the node that
 they're running on. Since this is an external service, there is no Consul node
 to register it onto. Instead, we will make up a node name and register the
@@ -205,15 +257,14 @@ metadata:
 spec:
   services:
     - name: example-https
-      caFile: /etc/ssl/certs/ca-certificates.crt
 ```
 
 </CodeBlockConfig>
 
-If TLS is enabled, you must include the `caFile` parameter  that points to the system trust store of the terminating gateway container. By default, the trust store is located in the `/etc/ssl/certs/ca-certificates.crt` directory.
-
+-> **NOTE**: If TLS is enabled for external services registered through the Consul catalog, you must include the `caFile` parameter that points to the system trust store of the terminating gateway container.
+By default, the trust store is located in the `/etc/ssl/certs/ca-certificates.crt` directory.
 Configure the `caFile` parameter to point to the `/etc/ssl/cert.pem` directory if TLS is enabled and you are using one of the following components:
- * Consul Helm chart 0.43 or older 
+ * Consul Helm chart 0.43 or older
  * Or an Envoy image with an alpine base image
 
 Apply the `TerminatingGateway` resource with `kubectl apply`:
@@ -313,6 +364,18 @@ deployment "static-client" successfully rolled out
 
 You can verify connectivity of the static-client and terminating gateway via a curl command:
 
+<CodeBlockConfig heading="External services registered with the Consul catalog">
+
 ```shell-session
 $ kubectl exec deploy/static-client -- curl -vvvs --header "Host: example-https.com" http://localhost:1234/
 ```
+
+</CodeBlockConfig>
+
+<CodeBlockConfig heading="External services registered with `service-defaults` destinations">
+
+```shell-session
+$ kubectl exec deploy/static-client -- curl -vvvs https://example.com/
+```
+
+</CodeBlockConfig>


website/content/docs/k8s/connect/terminating-gateways.mdx
```diff
@@ -88,6 +88,54 @@ Registering the external services with Consul is a multi-step process:
...
+1. You can add the service as a Node in the Consul catalog.
+
+#### Register an external service as a Destination
+
+Destinations are valid only for [`TransparentProxy`](/docs/k8s/helm#v-connectinject-transparentproxy) mode and allow clients to dial the external service directly.
+Also note these additional limitations when using Destinations to route traffic through a terminating gateway:
+
+| Layer 4 External Services (TCP) | Client dials hostname | Client dials IP |
+|---|---|---|
+| **with TLS** | ✅ Clients can dial successfully. `CAFiles` are not allowed because traffic is already end-to-end encrypted by the client. | ✅ Clients can dial successfully. CAFiles are not allowed because traffic is already end-to-end encrypted by the client. |
```
dhiaayachi, line 102: We should at least include SNI with CAFiles.

website/content/docs/k8s/connect/terminating-gateways.mdx
```diff
@@ -88,6 +88,54 @@ Registering the external services with Consul is a multi-step process:
...
+| **no TLS** | ❌ Because the Sidecar is not protocol aware, it will not be able to identify traffic going to the external service. | ✅ Clients can dial successfully. |
```
dhiaayachi, line 103: For IP with no TLS we should say that you can encrypt traffic at the tgtw using SNI and certs.

website/content/docs/k8s/connect/terminating-gateways.mdx
```diff
@@ -88,6 +88,54 @@ Registering the external services with Consul is a multi-step process:
...
+
+| Layer 7 External Services (HTTP, HTTP2 or gRPC) | Client dials hostname | Client dials IP |
+|---|---|---|
+| **with TLS** | ❌ Because traffic is already encrypted before the sidecar, it cannot route as L7 Traffic. | ❌ Because traffic is already encrypted before the sidecar, it cannot route as L7 Traffic. |
+| **no TLS** | ✅ Clients can dial successfully. A `Host` or `:authority` header is required. | ✅ Clients can dial successfully. |
```
dhiaayachi, line 108: Same here about encrypting traffic at the tgtwy

dhiaayachi (approved):
Few comments about SNI and certs.

website/content/docs/k8s/connect/terminating-gateways.mdx
DanStough: Per our discussion, let's add this to the config entry field descriptions.

dhiaayachi (approved):
🚢

website/content/docs/k8s/connect/terminating-gateways.mdx
```diff
@@ -88,6 +88,57 @@ Registering the external services with Consul is a multi-step process:
 
 ### Register external services with Consul
 
+There are two ways to register an external service with Consul:
+1. If you have [`TransparentProxy`](/docs/k8s/helm#v-connectinject-transparentproxy) enabled, you can declare external endpoints with the [Destination](/docs/connect/config-entries/service-defaults#terminating-gateway-destination) field of `service-defaults`.
```
trujillo-adam, line 92:
```suggestion
1. If [`TransparentProxy`](/docs/k8s/helm#v-connectinject-transparentproxy) is enabled, you can declare external endpoints in the [`Destination`](/docs/connect/config-entries/service-defaults#terminating-gateway-destination) field of `service-defaults`.
```
The docs for the `destination` field are still in production, correct?

website/content/docs/k8s/connect/terminating-gateways.mdx
```diff
@@ -88,6 +88,57 @@ Registering the external services with Consul is a multi-step process:
...
+1. You can add the service as a Node in the Consul catalog.
```
trujillo-adam, line 93:
```suggestion
1. You can add the service as a node in the Consul catalog.
```
If we mean Kubernetes Node, then we should capitalize to distinguish between nodes in Consul and Nodes in Kubernetes.

website/content/docs/k8s/connect/terminating-gateways.mdx
```diff
@@ -88,6 +88,57 @@ Registering the external services with Consul is a multi-step process:
...
+
+#### Register an external service as a Destination
+
+Destinations are valid only for [`TransparentProxy`](/docs/k8s/helm#v-connectinject-transparentproxy) mode and allow clients to dial the external service directly.
+Also note these additional limitations when using Destinations to route traffic through a terminating gateway:
+
+| Layer 4 External Services (TCP) | Client dials hostname | Client dials IP |
+|---|---|---|
+| **with TLS** | ✅ Clients can dial successfully. `CAFiles` are not allowed because traffic is already end-to-end encrypted by the client. | ✅ Clients can dial successfully. CAFiles are not allowed because traffic is already end-to-end encrypted by the client. |
```
trujillo-adam, line 102: A few things:
1. As far as I can see, we don't use the checkmark or cross-out icons anywhere else. For consistency, I think we should remove them.
2. It took me a while to figure out that these table represents a matrix with conditions along the x and y axises, as opposed to tabular reference information where the column headings represent a features associated with the items in the left-most column.
3. We introduce the matrix as describing limitations, but most of the reference points show successful calls. I would expect the matrix that describes limitations to only show instances when something is unsuccessful.

I think I would put this information into a more linear format.

website/content/docs/k8s/connect/terminating-gateways.mdx
```diff
@@ -88,6 +88,57 @@ Registering the external services with Consul is a multi-step process:
...
+| **no TLS** | ❌ Because the Sidecar is not protocol aware, it will not be able to identify traffic going to the external service. | ✅ Clients can dial successfully. |
+
+| Layer 7 External Services (HTTP, HTTP2 or gRPC) | Client dials hostname | Client dials IP |
+|---|---|---|
+| **with TLS** | ❌ Because traffic is already encrypted before the sidecar, it cannot route as L7 Traffic. | ❌ Because traffic is already encrypted before the sidecar, it cannot route as L7 Traffic. |
+| **no TLS** | ✅ Clients can dial successfully. A `Host` or `:authority` header is required. | ✅ Clients can dial successfully. |
+
+-> **NOTE**: Traffic from unencrypted clients can still be secured from the terminating gateway to the external service by providing a `caFile`.
+See the discussion below under [Create the configuration entry for the terminating gateway](/docs/k8s/connect/terminating-gateways#create-the-configuration-entry-for-the-terminating-gateway).
```
trujillo-adam, line 111:
```suggestion
You provide a `caFile` to secure traffic between unencrypted clients that connect to external services through the terminating gateway.
Refer to [Create the configuration entry for the terminating gateway](/docs/k8s/connect/terminating-gateways#create-the-configuration-entry-for-the-terminating-gateway) for details.
```
This didn't seem like orthogonal information that should be labeled in a side note. 
Also restated in active voice.

website/content/docs/k8s/connect/terminating-gateways.mdx
```diff
@@ -313,6 +363,14 @@ deployment "static-client" successfully rolled out
 
 You can verify connectivity of the static-client and terminating gateway via a curl command:
 
+**For external services registered with the Consul catalog**
```
trujillo-adam, line 366:
```suggestion
**For external services registered with the Consul catalog**
```
I get that we're bolding this sentence fragment so that it reads like a label for the following command, but I wonder if it wouldn't be better to wrap a CodeBlockConfig element around it and use the heading=""" attribute to make it look nicer.

trujillo-adam (commented):
A couple nits and more-than-nits comments.

website/content/docs/k8s/connect/terminating-gateways.mdx
DanStough: Those docs are not merged but already approved here: https://github.com/hashicorp/consul/pull/13950 
I'll wait until those get merged before merging this one.

website/content/docs/k8s/connect/terminating-gateways.mdx
DanStough: Ah, TIL. 🙏

website/content/docs/k8s/connect/terminating-gateways.mdx
DanStough: I used the note more to emphasize "pay attention" rather than "something related", but happy to adopt the convention here.

website/content/docs/k8s/connect/terminating-gateways.mdx
DanStough: OK, I'll try that out.

website/content/docs/k8s/connect/terminating-gateways.mdx
DanStough: I was using the table [here](https://learn.hashicorp.com/tutorials/consul/terminating-gateways-connect-external-services#prerequisites) as prior art, although now I see they use the 🚫 symbols instead. I'm not married to this but I would like your take on the reference provided.

When you say linear format, would you be OK with something like the following that doesn't use the Y axis? I like some sort of table organization because I personally found it hard to grok all these possible scenarios without having them laid out together.

| Service Layer | TLS | Client dials | Works out of the Box | Notes |
|---|---|---|---|---|
| L4 | No | Hostname | 🚫 | Because the Sidecar is not protocol-aware, it will not be able to identify traffic going to the external service. |

website/content/docs/k8s/connect/terminating-gateways.mdx
DanStough: I agree it looks nicer!

website/content/docs/k8s/connect/terminating-gateways.mdx
DanStough: See changes for the actual layout I picked.

website/content/docs/k8s/connect/terminating-gateways.mdx
```diff
@@ -88,6 +88,58 @@ Registering the external services with Consul is a multi-step process:
 
 ### Register external services with Consul
 
+There are two ways to register an external service with Consul:
+1. If [`TransparentProxy`](/docs/k8s/helm#v-connectinject-transparentproxy) is enabled, you can declare external endpoints in the [`Destination`](/docs/connect/config-entries/service-defaults#terminating-gateway-destination) field of `service-defaults`.
+1. You can add the service as a node in the Consul catalog.
+
+#### Register an external service as a Destination
```
lkysow, line 95: non-blocking: It would be amazing if you folks could provide some examples. e.g. for `https://www.twitter.com` use this `ServiceDefaults` (I'm guessing with protocol: tcp), and you can then call Twitter from your service with `curl https://www.twitter.com`

DanStough: I think this doc reads a little like a tutorial that ends with a similar "try this" command. I would favor replacing `example.com` with something like `httpbin.org` with a simple response. WDYT @trujillo-adam @lkysow?

lkysow: Ahh I see, sorry I didn't look at the full context. I still think that reading the table of caveats is hard to understand when mapping to what my use-case is. I don't have strong feelings about replacing httpbin but I do think that having examples for when I should use L4 vs L7 and what the servicedefaults and eventual `curl` calls would look like would make it much easier to understand.

DanStough: I agree the table is challenging, but I struggled to find a better representation. The best solution would be to have some permissive or auto setting to make figuring this out more accessible, but that's forward-looking. Open to any suggestions here.

trujillo-adam: W/r/t examples, I'm in favor of using as realistic configurations as possible so that readers can form a mental model of what we are describing. Regarding the table of behaviors, I actually think this version of the table is pretty straightforward. There are a few blank cells under the Notes column that we should complete, but as an outside reader I feel that I would be able to align my use case to the information being presented. Maybe Notes should be changed to Description.

website/content/docs/k8s/connect/terminating-gateways.mdx
```diff
@@ -88,6 +88,58 @@ Registering the external services with Consul is a multi-step process:
...
+| External Services Layer | Client dials | Client uses TLS | Allowed | Notes |
+|---|---|---|---|---|
+| L4 | Hostname | Yes | Allowed | `CAFiles` are not allowed because traffic is already end-to-end encrypted by the client. |
+| L4 | IP | Yes | Allowed | `CAFiles` are not allowed because traffic is already end-to-end encrypted by the client. |
+| L4 | Hostname | No | Not allowed | Because the Sidecar is not protocol aware, it will not be able to identify traffic going to the external service. |
+| L4 | IP | No | Allowed | |
+| L7 | Hostname | Yes | Not allowed | Because traffic is already encrypted before the sidecar, it cannot route as L7 Traffic. |
+| L7 | IP | Yes | Not allowed | Because traffic is already encrypted before the sidecar, it cannot route as L7 Traffic. |
+| L7 | Hostname | No | Allowed | A `Host` or `:authority` header is required. |
+| L7 | IP | No | Allowed | |
```
trujillo-adam, line 109: A couple of cells are empty in the Notes column. For consistency, we should describe the reason for the behavior for each row.

website/content/docs/k8s/connect/terminating-gateways.mdx
DanStough: OK, I think we'll keep it as-is for now and work on more examples as this feature develops. We have some additional features planned where we can add to this.

website/content/docs/k8s/connect/terminating-gateways.mdx
DanStough: Added some content. I also rebased.

website/content/docs/k8s/connect/terminating-gateways.mdx
```diff
@@ -88,6 +88,58 @@ Registering the external services with Consul is a multi-step process:
...
+| External Services Layer | Client dials | Client uses TLS | Allowed | Notes |
+|---|---|---|---|---|
+| L4 | Hostname | Yes | Allowed | `CAFiles` are not allowed because traffic is already end-to-end encrypted by the client. |
+| L4 | IP | Yes | Allowed | `CAFiles` are not allowed because traffic is already end-to-end encrypted by the client. |
+| L4 | Hostname | No | Not allowed | Because the Sidecar is not protocol aware, it will not be able to identify traffic going to the external service. |
+| L4 | IP | No | Allowed | There are no limitations on dialing IPs without TLS |
+| L7 | Hostname | Yes | Not allowed | Because traffic is already encrypted before the sidecar, it cannot route as L7 Traffic. |
+| L7 | IP | Yes | Not allowed | Because traffic is already encrypted before the sidecar, it cannot route as L7 Traffic. |
+| L7 | Hostname | No | Allowed | A `Host` or `:authority` header is required. |
+| L7 | IP | No | Allowed | There are no limitations on dialing IPs without TLS  |
```
trujillo-adam, line 109:
```suggestion
| L4 | Hostname | No | Not allowed | The sidecar is not protocol aware and can not identify traffic going to the external service. |
| L4 | IP | No | Allowed | There are no limitations on dialing IPs without TLS. |
| L7 | Hostname | Yes | Not allowed | Because traffic is already encrypted before the sidecar, it cannot route as L7 traffic. |
| L7 | IP | Yes | Not allowed | Because traffic is already encrypted before the sidecar, it cannot route as L7 traffic. |
| L7 | Hostname | No | Allowed | A `Host` or `:authority` header is required. |
| L7 | IP | No | Allowed | There are no limitations on dialing IPs without TLS.  |
```
Just a few super minor style things: use simple present tense and use lower case unless it's a proper name.

trujillo-adam (approved):
There are a few style things that we should fix, but approving so I don't slow you down.

PR #14016, opened 2022-08-03
ui: Ensure CopyableCode copy icon/button is fully clickable

johncowen:
### Description
Ensure CopyableCode copy icon/button is fully clickable

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-ui/app/components/copyable-code/index.scss
+++ b/ui/packages/consul-ui/app/components/copyable-code/index.scss
@@ -27,7 +27,14 @@
   .copy-button {
     position: absolute;
     top: 7px;
-    right: 12px;
+    right: 0;
+    z-index: 1;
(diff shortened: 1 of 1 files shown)


LevelbossMike:
Not sure if I'm doing something wrong but I still can't seem to press the entire (gray) button to copy the token - it looks like only the blue portion of the button is clickable.

Screenshot 2022-08-03 at 14 58 25

johncowen:
We avoid a full 100% height here as the gray area can extend to be really tall and I don't think its is meant to all be clickable. I suppose I can make this square though, 1 tic I can add that

johncowen:
K this now has the same 40px height also

LevelbossMike (approved):
:clap:

PR #14017, opened 2022-08-03
dataplane: update envoy bootstrap params for consul-dataplane

boxofrad:
### Description
This PR contains 2 changes to the `GetEnvoyBootstrapParams` response to support consul-dataplane.

#### Exposing `node_name` and `node_id`
consul-dataplane will support providing either the `node_id` or `node_name` in its configuration.

Unfortunately, supporting both in the xDS meta adds a fair amount of complexity (partly because most tables are currently indexed on `node_name`) so for now we're going to return them both from the bootstrap params endpoint, allowing consul-dataplane to exchange a `node_id` for a `node_name` (which it will supply in the xDS meta).

#### Properly setting `service` for gateways
To avoid the need to special case gateways in consul-dataplane, `service` will now either be the destination service name for connect proxies, or the gateway service name. This means it can be used as-is in Envoy configuration (i.e. as a cluster name or in metric tags).


riddhi89 (commented):
LGTM. 
Holding off on approving in case we decide to include more fields in the response based on your open PR in consul-dataplane.

riddhi89 (approved):
LGTM. Thanks for adding the changes 🙌

PR #14018, opened 2022-08-03
ui: Add additional generation/initiation data-sources

johncowen:
### Description
I noticed that occasionally you can get a JS/Ember error when switching tabs in the peering generation/establishment modal. Looking/thinking about it this is probably related to how we cache/reuse our data-source responses, which in this case here are brand new ember data records. 

Usually, datasource responses are cached _and_ shared, so if you use the same data-source uri in two completely different routes, they use the same request and therefore result. From briefly looking at what is happening here, as we 'request' a new ember data record in two different states it seems that sometimes one datasource is removed after another is added when you switch tabs (this would possibly go deep into how Ember/glimmer works when components are removed from the DOM).

To help understanding, where this surfaces is here, only sometimes when we switch between the two states/tabs.

https://github.com/hashicorp/consul/blob/main/ui/packages/consul-peerings/app/components/consul/peer/form/index.hbs#L27-L59

I tried simply making two separate data-source URIs so there is no chance that these 'requests' get shared, and that consistently resolved the bug.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-ui/app/services/repository/peer.js
+++ b/ui/packages/consul-ui/app/services/repository/peer.js
@@ -50,14 +50,18 @@ export default class PeerService extends RepositoryService {
       });
   }
 
+  @dataSource('/:partition/:ns/:dc/peer-generate/')
+  @dataSource('/:partition/:ns/:dc/peer-initiate/')
   @dataSource('/:partition/:ns/:dc/peer/:name')
   async fetchOne({partition, ns, dc, name}, { uri }, request) {
     if (name === '') {
-      return this.create({
+      const item = this.create({
         Datacenter: dc,
         Namespace: '',
         Partition: partition,
       });
+      item.meta = {cacheControl: 'no-store'};
+      return item;
     }
     return (await request`
       GET /v1/peering/${name}
--- a/ui/packages/consul-peerings/app/components/consul/peer/form/index.hbs
+++ b/ui/packages/consul-peerings/app/components/consul/peer/form/index.hbs
@@ -27,7 +27,7 @@
     <fsm.State @matches={{array 'generate'}}>
 
       <DataSource
-        @src={{uri '/${partition}/${nspace}/${dc}/peer/'
+        @src={{uri '/${partition}/${nspace}/${dc}/peer-generate/'
           @params
         }}
       as |source|>
@@ -44,7 +44,7 @@
     <fsm.State @matches="initiate">
 
       <DataSource
-        @src={{uri '/${partition}/${nspace}/${dc}/peer/'
+        @src={{uri '/${partition}/${nspace}/${dc}/peer-initiate/'
           @params
         }}
       as |source|>


ui/packages/consul-ui/app/services/repository/peer.js
```diff
@@ -50,14 +50,18 @@ export default class PeerService extends RepositoryService {
       });
   }
 
+  @dataSource('/:partition/:ns/:dc/peer-generate/')
+  @dataSource('/:partition/:ns/:dc/peer-initiate/')
```
LevelbossMike, line 54: It looks like we were relying on this data-source for the creation and the edit case. I.e. when `name` is an empty string, we will return a newly created `Peer`-model if there's already a name we will fetch from the API.

Two questions:

1. If we wanted to stick to the data-source to generate a new ember-data model instance, can we get rid of the `name === ''` check and create a new method that only creates a record and annotate this with these new annotations? This would clean up the `fetchOne` method to only request from the backend and probably make it easier to understand what's going on. One method fetches data and the other one creates a new model instance.
2. Wouldn't it be easier to not rely on the data-source here and put the logic of creating a new peer model and cleaning it up if need be into the form itself?

LevelbossMike (changes requested):
I can't reproduce an error being thrown when switching tabs. What is the error that you are seeing?

I see the `DataSource`-component trying to `rollbackAttributes()` when it gets disconnected which will throw an error if we continued working with the then destroyed record.

Is that the error you are seeing?

`Error: Attempted to set 'Name' to 'something' on the deleted record <peer:null>`

 Surprising that that would happen as disconnecting is tied to `will-destroy` of the `DataSource`-component but if it returned a `cached` record that might really be the issue 🤔

johncowen:
> Is that the error you are experiencing?
> `Error: Attempted to set 'Name' to 'something' on the deleted record <peer:null>`

Yes it was something like that as far as I remember.

I had a little more of a look into it, and my original thought was correct:

> From briefly looking at what is happening here, as we 'request' a new ember data record in two different states it seems that sometimes one datasource is removed after another is added when you switch tabs (this would possibly go deep into how Ember/glimmer works when components are removed from the DOM).

So during glimmer component insertion/destruction, the "switched to tab" is inserted first, before the "switched from tab" is destroyed, meaning there is a period of time when both tabs are in existence. This means the data represented by the two exact same data-sources is shared, which is expected behaviour for data-sources. 

The simplest solution here is to just make the two data-sources distinct (which is what I've done here), one creates an object specifically for generation, and one creates and object specifically for initialisation. To me this actually make a bit more sense also.

ui/packages/consul-ui/app/services/repository/peer.js
johncowen: > 1. If we wanted to stick to the data-source to generate a new ember-data model instance, can we get rid of the name === '' check and create a new method that only creates a record and annotate this with these new annotations? This would clean up the fetchOne method to only request from the backend and probably make it easier to understand what's going on. One method fetches data and the other one creates a new model instance.

We could do, but personally I don't feel like a single conditional here makes the method harder to understand, not only that but whether you are creating a 'new' record or not, you are still being returned a 'filled out' record, so them both being the same thing makes sense to me, they both 'fetch' new data from the repository. The fact that one makes a HTTP request and one doesn't is hidden by the repository layer.

> 2. Wouldn't it be easier to not rely on the data-source here and put the logic of creating a new peer model and cleaning it up if need be into the form itself?

I would say no it wouldn't be easier. This is how the rest of the app works, we pass data into the components from the outside. We avoid loading any sort of data deep down inside component trees. All our components are largely presentational.

Our abstract/base repository class has this very same behaviour and its used by all our models.

https://github.com/hashicorp/consul/blob/155c4bc2af77f71ca9cd5067f3c18d1c9ff67ecf/ui/packages/consul-ui/app/services/repository.js#L186-L193

LevelbossMike (approved):
approving as this needs to go in for peering-ui

PR #14021, opened 2022-08-03
ui: Ensure we dispatch validation state

johncowen:
### Description
Ensure we dispatch validation extended state when we have it, instead of just appending to the extended state, otherwise validation errors aren't reported properly.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-ui/app/modifiers/validate.js
+++ b/ui/packages/consul-ui/app/modifiers/validate.js
@@ -34,10 +34,10 @@ export default class ValidateModifier extends Modifier {
     }
     if(Object.keys(errors).length > 0) {
       state.context.errors = errors;
-      this.hash.chart.dispatch("ERROR");
+      this.hash.chart.dispatch("ERROR", state.context);
     } else {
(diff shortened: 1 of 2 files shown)


ui/packages/consul-ui/app/modifiers/validate.js
```diff
@@ -34,10 +34,10 @@ export default class ValidateModifier extends Modifier {
     }
     if(Object.keys(errors).length > 0) {
       state.context.errors = errors;
-      this.hash.chart.dispatch("ERROR");
+      this.hash.chart.dispatch("ERROR", state.context);
```
LevelbossMike, line 37: does it make sense to add a test for this here - or is it better to do this as a follow-up?

LevelbossMike (approved):
changes look fine - errors should now be present on the context of the statemachine. Ideally we'd be using [assign](https://xstate.js.org/docs/packages/xstate-fsm/#example) from xstate directly but that's obviously a bigger refactor and for a follow-up.

PR #14024, opened 2022-08-03
Peering secret UUID management fixes

freddygv:
### Description
This PR addresses some bugs related to the peering secret management. 

There are two commits, each with additional context in the commit message:
1. Dialers should not touch the peering secret UUIDs table when writing new secrets nor when deleting them. Dialing peers are not responsible for managing the uniqueness of secrets for peers that they dial. That is the responsibility of the peer that generated the peering token.
2. Tracking the UUIDs to add/delete from the associated table is now done by explicitly stating the operation that is triggering the secrets write. Previously these side-effects were inferred solely by reconciling against the existing secrets for a peer. Now it is done by passing (and requiring) the operation.

* Consul unit tests
* consul-k8s unit tests
* consul-k8s acceptance tests

### Links
Follow-up to [ENT-2422](https://github.com/hashicorp/consul-enterprise/pull/2422)

### PR Checklist

* [x] updated test coverage
* [ ] not a security concern


freddygv:
Looking into some lingering issues

agent/grpc-external/services/peerstream/server_test.go
```diff
@@ -47,8 +50,5 @@ func TestServer_ExchangeSecret(t *testing.T) {
 		require.NoError(t, err)
 
 		require.Equal(t, secret, s.GetStream().GetPendingSecretID())
-
-		// Active stream secret persists until pending secret is promoted during peering establishment.
```
freddygv, line 51: This is tested in the state store instead

sdk/testutil/testlog.go
```diff
@@ -21,7 +21,7 @@ func testLogLevel() hclog.Level {
 	if level != hclog.NoLevel {
 		return level
 	}
-	return hclog.Warn
+	return hclog.Trace
```
mkeeler, line 24: Did you intend to leave this in?

The only downside I see is if someone were to run tests with the `-v` flag then this would dump a ton of logs. Something like INFO seems more appropriate and then you can override the level with the TEST_LOG_LEVEL env var.

agent/consul/state/peering.go
```diff
@@ -251,45 +251,42 @@ func (s *Store) peeringSecretsWriteTxn(tx WriteTxn, secret *pbpeering.PeeringSec
 
 	var toDelete []string
 	if existing != nil {
```
mkeeler, line 253: Should we add validation that existing != nil unless the operation is generatetoken?

agent/consul/state/peering.go
```diff
@@ -251,45 +251,42 @@ func (s *Store) peeringSecretsWriteTxn(tx WriteTxn, secret *pbpeering.PeeringSec
...
-		existingActive := existing.GetStream().GetActiveSecretID()
-		if existingActive != "" && existingActive != secret.GetStream().GetActiveSecretID() {
-			toDelete = append(toDelete, existingActive)
-		}
+		case pbpeering.PeeringSecretsWriteRequest_OPERATION_EXCHANGESECRET:
+			// Avoid invalidating existing active secrets when exchanging establishment secret for pending.
+			secret.Stream.ActiveSecretID = existing.GetStream().GetActiveSecretID()
+
+			// When exchanging an establishment secret we invalidate the existing establishment secret.
+			if existingEstablishment := existing.GetEstablishment().GetSecretID(); existingEstablishment != "" {
```
mkeeler, line 272: Should we error out if the establishment secret isn't set?

I am thinking that ExchangeSecret is called in quick succession and that we end up sending two raft logs with different pending secrets through. The first has already removed the secret and the next potentially is going to overwrite the pending one. Its super racey but I wonder if moving more validation logic into the state store like this would be helpful to mitigate those races.

mkeeler:
FWIW I do think this fixes the bug and if we wouldn't have time to refactor further then I don't have strong reservations about merging this as is.

proto/pbpeering/peering.proto
```diff
@@ -195,10 +213,10 @@ message PeeringWriteRequest {
   // Peering is the peering to write with the request.
   Peering Peering = 1;
 
-  // PeeringSecrets contains the optional peering secrets to persist
+  // Secret contains the optional peering secrets to persist
```
acpana, line 216: SecretsRequest
```suggestion
  // SecretsRequest contains the optional peering secrets to persist
```

acpana (commented):
just leaving some non blocking first pass comments

agent/grpc-external/services/peerstream/stream_test.go
```diff
@@ -194,7 +198,13 @@ func TestStreamResources_Server_ActiveSecretValidation(t *testing.T) {
 		srv, store := newTestServer(t, nil)
 
 		// Write a seed peering.
-		require.NoError(t, store.PeeringWrite(1, tc.seed))
+		if tc.seed != nil {
+			require.NoError(t, store.PeeringWrite(1, &pbpeering.PeeringWriteRequest{Peering: tc.seed.peering}))
+
+			for _, s := range tc.seed.secrets {
```
freddygv, line 204: This is a slice because of there not being a method to insert a batch of peering secrets as-is outside of the restore function. Each step in the slice goes through secrets writes as they would in production: generate token, exchange secret, then promote pending.

sdk/testutil/testlog.go
freddygv: Nope, missed reverting that before committing. It's reverted now.

agent/consul/state/peering.go
freddygv: Yep put that in

proto/pbpeering/peering.proto
```diff
@@ -53,6 +53,24 @@ enum PeeringState {
...
 
+message PeeringSecretsWriteRequest {
+  enum Operation {
+    OPERATION_UNSPECIFIED = 0;
+
+    OPERATION_GENERATETOKEN = 1;
+
+    OPERATION_EXCHANGESECRET = 2;
+
+    OPERATION_PROMOTEPENDING = 3;
```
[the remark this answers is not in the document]
freddygv: Resolving since this was moved from enum to distinct messages, and added comments there

agent/consul/state/peering.go
freddygv: I put in a check to make sure it's set, but I think I'm actually going to pass in the establishment secret that was validated against to guard against the case where it was non-empty but different

mkeeler (approved):

mkeeler (approved):

PR #14025, opened 2022-08-04
ui: Make DataWriter also deal with more standard shaped errors

johncowen:
### Description
We started out with our application using `ember-data`, but for many reasons we are slowly moving away from using that. As well as a few other benefits it also means our errors (API ones and any others) are all standard shaped errors - all errors are the same standard recognisable javascript errors.

This PR alters our `DataWriter` to work with standard shaped errors as well as `ember-data` shaped ones (as it did previously). This means we can keep deprecating our `ember-data` usage as we go. Eventually, if we get to the point we we aren't using `ember-data` shaped errors anymore, we can remove the majority of this code and just support standard javascript errors.

**Note**: We purposefully used an old style `action` here to keep consistent with the style/APIs already being used in the component (the component is currently an old school ember one). Now is not the time to do a full upgrade of this component and I'd rather the component was consistent using one single style rather than both old school ember and new school Octane both at the same time. Upgrading the component entirely to Octane can be done at some point in the future.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-ui/app/components/data-writer/index.js
+++ b/ui/packages/consul-ui/app/components/data-writer/index.js
@@ -21,5 +21,17 @@ export default Component.extend(Slotted, {
       set(this, 'data', data);
       this.dispatch('PERSIST');
     },
+    error: function(data, e) {
+      if (e && typeof e.preventDefault === 'function') {
+        e.preventDefault();
+      }
+      set(
+        this,
+        'error',
+        typeof data.error.errors !== 'undefined' ?
+          data.error.errors.firstObject : data.error
+      );
+      this.dispatch('ERROR');
+    },
   },
 });
--- a/ui/packages/consul-peerings/app/components/consul/peer/form/generate/index.hbs
+++ b/ui/packages/consul-peerings/app/components/consul/peer/form/generate/index.hbs
@@ -17,7 +17,20 @@
       id={{id}}
     >
 
-      <fsm.State @matches={{array 'idle'}}>
+      <fsm.State @matches={{array 'idle' 'error'}}>
+        <fsm.State @matches={{'error'}}>
+          <Notice
+            @type="error"
+            role="alert"
+          as |notice|>
+            <notice.Body>
+              <p>
+                <strong>Error</strong><br />
+                {{fsm.state.context.error.message}}
+              </p>
+            </notice.Body>
+          </Notice>
+        </fsm.State>
         {{yield (hash
           Fieldsets=(component "consul/peer/form/generate/fieldsets"
             item=@item
@@ -42,6 +55,9 @@
           @onchange={{queue
             @onchange
             (pick 'data' (fn fsm.dispatch 'SUCCESS'))
+          }}
+          @onerror={{queue
+            (fn fsm.dispatch 'ERROR')
           }}
         />
       </fsm.State>
--- a/ui/packages/consul-peerings/app/components/consul/peer/form/initiate/index.hbs
+++ b/ui/packages/consul-peerings/app/components/consul/peer/form/initiate/index.hbs
@@ -17,6 +17,19 @@
     @label={{'peer'}}
     @onchange={{fn (optional @onsubmit) @item}}
     as |writer|>
+      <BlockSlot @name="error" as |after error|>
+        <Notice
+          @type="error"
+          role="alert"
+        as |notice|>
+          <notice.Body>
+            <p>
+              <strong>Error</strong><br />
+              {{error.message}}
+            </p>
+          </notice.Body>
+        </Notice>
+      </BlockSlot>
       <BlockSlot @name="content">
 {{#let
   (unique-id)
@@ -27,15 +40,15 @@ as |id|}}
         >
           {{yield (hash
             Fieldsets=(component "consul/peer/form/initiate/fieldsets"
-             item=@item
+              item=@item
             )
             Actions=(component "consul/peer/form/initiate/actions"
-             item=@item
-             id=id
+              item=@item
+              id=id
             )
           )}}
         </form>
 {{/let}}
-      </BlockSlot>
+    </BlockSlot>
   </DataWriter>
 </div>
--- a/ui/packages/consul-ui/app/components/data-writer/index.hbs
+++ b/ui/packages/consul-ui/app/components/data-writer/index.hbs
@@ -19,7 +19,7 @@
       @item={{data}}
       @data={{null}}
       @onchange={{action dispatch "SUCCESS"}}
-      @onerror={{queue (action (mut error) value="error.errors.firstObject") (action dispatch "ERROR")}}
+      @onerror={{action "error"}}
     />
   </State>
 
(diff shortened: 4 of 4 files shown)


ui/packages/consul-ui/app/components/data-writer/index.js
```diff
@@ -21,5 +21,17 @@ export default Component.extend(Slotted, {
       set(this, 'data', data);
       this.dispatch('PERSIST');
     },
+    error: function(data, e) {
+      if (e && typeof e.preventDefault === 'function') {
```
LevelbossMike, line 25: why are we `preventingDefault` here? can we add a test-case for this change?

ui/packages/consul-ui/app/components/data-writer/index.js
```diff
@@ -21,5 +21,17 @@ export default Component.extend(Slotted, {
...
+        e.preventDefault();
+      }
+      set(
+        this,
+        'error',
+        typeof data.error.errors !== 'undefined' ?
+          data.error.errors.firstObject : data.error
```
LevelbossMike, line 32: We are hard-coding against another "special" shape here, we expect an object with an `error`-property. The template of `DataWriter` also still expects `error.status` and `error.detail` to be available - will this need to be changed in this PR?

May I suggest to allow users to specify a transformation function instead, that will get the error payload passed? That would let the caller specify how the error-payload gets transformed before we set it as the `error`-property on the writer, or have the caller deal with the error entirely:

```hbs
{{!-- allow caller handle or transform the error --}}
<DataWriter @onError={{this.handleError}} />
```
This would allow us to keep the old way as the default behaviour but have callers handle special cases via `@onError`.

LevelbossMike (changes requested):
Added some comments. Would be great to add some tests for this behaviour as well.

ui/packages/consul-ui/app/components/data-writer/index.js
johncowen: > We are hard-coding against another "special" shape here

The shape is no longer "special", it's a Javascript event based on an [ErrorEvent](https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent) which has an error property (containing a standard shaped javascript error)

> May I suggest to allow users to specify a transformation function instead, that will get the error payload passed?

Ideally the component should always stick as closely as possible to existing Javascript objects, it only used/uses non-standard ones because `ember-data`.

Additionally, all of this is dealt with inside the component via interactions between DataSink (the equivalent of DataSource) and DataWriter itself (the equivalent of DataLoader). So as a user you just need to fill out the `error` slot.

> This would allow us to keep the old way as the default behaviour but have callers handle special cases via @onError.

The new way is is now the "default behaviour" the old way is deprecated whilst we switch over to using less `ember-data`, once we finish the switch over there will be no other "cases" apart from the default one. I would prefer this sharp edge whilst we are switching over to be hidden from the user, so inside the component itself in my eyes is the best place to do it - now all instances of DataWriter support standard JS erroring.

---

> still expects error.status and error.detail to be available

Are you referring to this line here?

https://github.com/hashicorp/consul/blob/155c4bc2af77f71ca9cd5067f3c18d1c9ff67ecf/ui/packages/consul-ui/app/components/data-writer/index.hbs#L109-L111

Where we conditionally check if `error.status` and `error.detail` exists?

If so, I guess I can add a check for `error.message` here also, it doesn't effect this PR or the follow up though.

ui/packages/consul-ui/app/components/data-writer/index.js
johncowen: If the event that is fired has some sort of default we want to prevent it.

This is currently a `noop`, as I'm pretty sure events passed here will never have a default. So happy to leave in or take out, it mainly for consistency with the above `persist`. We might eventually have a case where the event does have a default that needs preventing, and this way both `persist` and `error` work the same way. 

All in all I'm easy either way.

> can we add a test-case for this change?

Which bit would you like me to test here? The

```javascript
e && typeof e.preventDefault === 'function'
```
?

LevelbossMike: a test case testing the newly added behavior would be great

ui/packages/consul-ui/app/components/data-writer/index.js
LevelbossMike: aha now I see the actual error that is used here seems to originate from here:

https://github.com/hashicorp/consul/blob/155c4bc2af77f71ca9cd5067f3c18d1c9ff67ecf/ui/packages/consul-ui/app/utils/dom/event-source/index.js#L141-L144

And we can always expect an `ErrorEvent`-like data-structure.

Still think the callers should probably be able to handle this error somehow for flexibility sake but for now I guess it's fine 🤷

LevelbossMike (approved):
Marking as approval for sake of landing for the peering release - we can add tests for this behavior in a follow up.

ui/packages/consul-ui/app/components/data-writer/index.js
johncowen: Can I just make sure you understand what was changed here?

We didn't add any new behaviour as such, we just added a fork for a different shape of data. So to me right now, it sounds like you want me to test either (or both?)

```javascript
e && typeof e.preventDefault === 'function'
```

and/or

```javascript
typeof data.error.errors !== 'undefined'
```

Just to note, acceptance tests for the whole peering feature will be added once everything is in.

ui/packages/consul-ui/app/components/data-writer/index.js
johncowen: Here are some docs for that for reference https://developer.mozilla.org/en-US/docs/Web/API/EventSource/error_event

ui/packages/consul-ui/app/components/data-writer/index.js
LevelbossMike: Is there a test for this behavior anywhere? You are changing code but you are not adding/updating a test-case for the change. We are also adding behavior of `preventDefault`ing the regular event. Why is this needed? Do we want to add a test-case that showcases why this change is necessary?

We can add that as a follow-up as the change is probably fine but don't we want to add a test harness for this behavior?

ui/packages/consul-ui/app/components/data-writer/index.js
johncowen: Forgot to mention, the value of the `error` property on the `ErrorEvent` does get 'exposed' and you can then use that as flexibly as you like. The error gets yielded out basically and you can use that however you like from thereon.

For example here in policy notifications. We know `Policy` uses ember-data shaped errors so we use that shape here

https://github.com/hashicorp/consul/blob/155c4bc2af77f71ca9cd5067f3c18d1c9ff67ecf/ui/packages/consul-ui/app/components/consul/policy/notifications/index.hbs#L20-L24

I think you mentioned offline that we may want to parse a specific error, so you could do that quite easily here/like this if that was necessary.

PR #14026, opened 2022-08-04
ui: Add error states/messages for peering establishment

johncowen:
### Description
Adds backend error messaging to the peering establishment flows.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-peerings/app/components/consul/peer/form/generate/index.hbs
+++ b/ui/packages/consul-peerings/app/components/consul/peer/form/generate/index.hbs
@@ -17,7 +17,20 @@
       id={{id}}
     >
 
-      <fsm.State @matches={{array 'idle'}}>
+      <fsm.State @matches={{array 'idle' 'error'}}>
+        {{#if fsm.state.context.error}}
+          <Notice
+            @type="error"
+            role="alert"
+          as |notice|>
+            <notice.Body>
+              <p>
+                <strong>Error</strong><br />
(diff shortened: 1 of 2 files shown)


ui/packages/consul-peerings/app/components/consul/peer/form/generate/index.hbs
```diff
@@ -17,7 +17,20 @@
       id={{id}}
     >
 
-      <fsm.State @matches={{array 'idle'}}>
+      <fsm.State @matches={{array 'idle' 'error'}}>
+        {{#if fsm.state.context.error}}
```
LevelbossMike, line 21: May I suggest to, instead of relying on the context here, to add another `<fsm.State @matches={{array 'error'}}>`-block? That way, we will display the notice based on the state of the state-machine, not based on a property set on `context`.

LevelbossMike (changes requested):
I added a small comment - as this PR depends on https://github.com/hashicorp/consul/pull/14025 I'm marking as requesting for change.

ui/packages/consul-peerings/app/components/consul/peer/form/generate/index.hbs
johncowen: Oooh yes! I like that much better, I added a commit on the end here

LevelbossMike (approved):

PR #14043, opened 2022-08-08
Add retries and debugging to flaky test

kisunji:
### Description
`TestAgentConfigWatcherSidecarProxy` keeps flaking because the second config fired by the watch sometimes does not reflect the updated config. I can't reproduce this locally so I added some retries and additional assertions to see what field changes may have fired the watch.

### DIFF
--- a/connect/proxy/config_test.go
+++ b/connect/proxy/config_test.go
@@ -4,12 +4,14 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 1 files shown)


freddygv (approved):
This seems reasonable, curious to see what comes up from the new debug statements

The watches are based on the `ContentHash` of the `AgentService`, and after reading through both the [blocking](https://github.com/hashicorp/consul/blob/main/api/watch/plan.go#L59) and the [hash creation](https://github.com/hashicorp/consul/blob/main/agent/agent_endpoint.go#L449) nothing jumped out

PR #14055, opened 2022-08-08
Add debugging logs to freeport

kisunji:
### Description
Many of our flaky / failing tests seem to be caused by ports being held by parallel tests. @acpana tried to address this in https://github.com/hashicorp/consul/pull/13997.

```
[WARN] freeport: 1 out of 5 pending ports are still in use; something probably didn't wait around for the port to be closed!
[WARN] freeport: 1 out of 1 pending ports are still in use; something probably didn't wait around for the port to be closed!
[WARN] freeport: 1 out of 1 pending ports are still in use; something probably didn't wait around for the port to be closed!
```

This PR does not attempt a solution but adds debug lines to our tests to better trace which tests are still holding onto ports.

### DIFF
--- a/sdk/testutil/types.go
+++ b/sdk/testutil/types.go
@@ -9,4 +9,5 @@ type TestingTB interface {
 	Logf(format string, args ...interface{})
 	Name() string
 	Fatalf(fmt string, args ...interface{})
+	Helper()
 }
--- a/sdk/freeport/freeport.go
+++ b/sdk/freeport/freeport.go
@@ -195,6 +195,8 @@ func checkFreedPortsOnce() {
 		if used := isPortInUse(port); !used {
 			freePorts.PushBack(port)
 			remove = append(remove, elem)
+		} else {
+			logf("WARN", "port %d still being used", port)
 		}
 	}
 
@@ -314,7 +316,6 @@ func Take(n int) (ports []int, err error) {
 		ports = append(ports, port)
 	}
 
-	// logf("DEBUG", "free ports: %v", ports)
 	return ports, nil
 }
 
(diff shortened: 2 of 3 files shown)


sdk/testutil/types.go
```diff
@@ -9,4 +9,5 @@ type TestingTB interface {
 	Logf(format string, args ...interface{})
 	Name() string
 	Fatalf(fmt string, args ...interface{})
+	Helper()
```
dhiaayachi, line 12: honest question, why we needed to add `Helper` to this interface?

dhiaayachi (approved):
General question about this, I remember at some point we were discussing that we should not use `Take` anymore and use `GetOne` and `GetN` as `Take` would not automatically clean after. I'm wondering why we do not deprecate/remove `Take` in favour of the other two, even in the multithreaded scenario we should be able to do port allocation up front in the main thread and pass them after.

sdk/testutil/types.go
kisunji: I switched `defaultServerConfig` to use `GetN` instead of `Take`, which requires this interface to have `Helper()`

kisunji:
> General question about this, I remember at some point we were discussing that we should not use `Take` anymore and use `GetOne` and `GetN` as `Take` would not automatically clean after. I'm wondering why we do not deprecate/remove `Take` in favour of the other two, even in the multithreaded scenario we should be able to do port allocation up front in the main thread and pass them after.

This PR actually removes the last usage of `Take` in our tests so technically we could make it package private (but I wasn't sure if there were other dependencies on the package). As far as Consul is concerned we use `GetOne` or `GetN` everywhere now.

PR #14056, opened 2022-08-08
Refactor sidecar_service method to separate port assignment

skpratt:
### Description
This separates the port assignment logic out into a helper. That way it can later be moved to a different step in the registration process.

* Refactor, no functionality changes intended.
* Unit test also refactored into separate tests.

### Links
[Follow-on PR](https://github.com/hashicorp/consul/pull/14057)

### PR Checklist

* [x] updated test coverage
* [N/A] external facing docs updated


agent/sidecar_service.go
```diff
@@ -114,9 +114,32 @@ func (a *Agent) sidecarServiceFromNodeService(ns *structs.NodeService, token str
...
+	if len(checks) < 1 {
+		checks = sidecarDefaultChecks(ns.ID, sidecar.Proxy.LocalServiceAddress, sidecar.Port)
+	}
+
+	return sidecar, checks, token, nil
+}
+
+// sidecarPortFromServiceID is used to allocate a unique port for a sidecar proxy.
+// This is called immediately before registration to avoid value collisions. This function assumes the state lock is already held.
+func (a *Agent) sidecarPortFromServiceIDLocked(sidecarPort int, sidecarCompoundServiceID structs.ServiceID) (int, error) {
```
kisunji, line 138: IMO the `-Locked` suffix is most helpful when the locks exist near the functions being called. In this refactor I had to search the call stack to check that a state lock was being held somewhere so I think it might trip up other devs who look at this function and don't see a lock in scope.

Not sure if there's a clean solution (since it is worth calling out that we need a lock) but just calling it out.

kisunji: Signature-wise it's a little confusing to take `sidecarPort` as input and also return a port. Could we clarify in the docs what the input is for or perhaps mutate the sidecar directly?

agent/sidecar_service_test.go
```diff
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"fmt"
+	"github.com/hashicorp/consul/acl"
```
kisunji, line 5: Our team uses `goimports -local github.com/hashicorp/consul` to group local packages together. If you use Goland here's the setting: 
image

kisunji (approved):
👍 I left some open-ended comments but no blockers.

Will there be a follow-up PR to address the original Github issue?

dhiaayachi (approved):
LGTM! I second @kisunji non blocking suggestion though

agent/sidecar_service_test.go
skpratt: done

agent/sidecar_service.go
skpratt: done

PR #14070, opened 2022-08-08
docs: Update references for transparent proxy with peers

kisunji:
### Description
Did another pass through of docs re: tproxy.

One thing I noticed was that it's very hard to navigate from Transparent Proxy to any of the relevant information (service-registration and dns)

### DIFF
--- a/website/content/docs/connect/transparent-proxy.mdx
+++ b/website/content/docs/connect/transparent-proxy.mdx
@@ -29,8 +29,8 @@ Without transparent proxy, application owners need to:
 
 With transparent proxy:
 
-1. Upstreams are inferred from service intentions, so no explicit configuration
-   is needed.
+1. Upstreams are inferred from service intentions and/or imported services,
+   so no explicit configuration is needed.
 1. Outbound connections pointing to a KubeDNS name "just work" — network rules
    redirect them through the proxy.
 1. Inbound traffic is forced to go through the proxy to prevent unauthorized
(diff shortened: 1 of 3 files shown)


website/content/docs/discovery/dns.mdx
```diff
@@ -488,18 +488,17 @@ fields must be present:
...
-For node lookups, only the partition and datacenter need to be specified as nodes cannot be
-namespaced.
+For imported lookups, only the namespace and peer need to be specified as the partition that is peered can be inferred:
 
 ```text
-[tag.]<service>.service.<partition>.ap.<datacenter>.dc.<domain>
+<service>.virtual[.namespace][.peer].<domain>
 ```
 
-For node lookups, only the partition and datacenter need to be specified (nodes cannot be
```
kisunji, line 498: The diff is misleading; there were similarly worded sections for node (which had the wrong copy-pasted example)

website/content/docs/connect/transparent-proxy.mdx
```diff
@@ -29,8 +29,8 @@ Without transparent proxy, application owners need to:
 
 With transparent proxy:
 
-1. Upstreams are inferred from service intentions, so no explicit configuration
-   is needed.
+1. Upstreams are inferred from service intentions and/or imported services,
```
mkeeler, line 32: Should we clarify that the intentions we infer upstreams from are the ones from the local cluster and not ones set in remote peers?

kisunji: Yes, that's a good point.

trujillo-adam (approved):
lgtm

PR #14072, opened 2022-08-09
fix:  missing MaxInboundConnections field in service-defaults config entry

huikang:
### Description
Setting `max_inbound_connections` [has been implemented](https://github.com/hashicorp/consul/blob/main/.changelog/13143.txt), but its description is missing in the doc.

This PR adds the description to the related doc. Since consul-k8s doesn't have this field in its service-defaults CRD ([link](https://github.com/hashicorp/consul-k8s/issues/1400) to issue), this PR doesn't add the description to the k8s page (`yaml: false`).

**Update**:
- It turns out the `MaxInboundConnections` field is not included in the serviceDefaults API structs, so it can't be passed to the envoy configuration. This PR fixes the issue by adding `MaxInboundConnections` and updating the mergeconfig to allow the value to be passed to envoy configuration.

### Reproduce

```
Kind      = "service-defaults"
Name      = "frontend"
Protocol  = "http"

UpstreamConfig = {
  Defaults = {
    MeshGateway = {
      Mode = "local"
    }
    passive_health_check {
      max_failures = 20
    }
  }
}
MaxInboundConnections = 100
```

```
consul config write  service-defaults.hcl
Failed to decode config entry input: 1 error occurred:
	* invalid config key "MaxInboundConnections"
```

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


huikang:
> One minor nitpick: What if we moved this new config block up to be next to `Protocol` and `ConnectTimeoutMs`? Since it's similarly shaped configuration

@freddygv , thanks for your feedback. I think `Protocol` and `ConnectTimeoutMs` are under the `UpstreamConfiguration`, while the `MaxInboundConnections` controls the downstream connection. So I think they should be separate. Please correct me if I miss anything.

freddygv:
@huikang Good catch, I misread the indentation on the markdown. Can leave it as is.

huikang:
@freddygv , it seems that there is something missing when I tested setting `MaxInboundConnections` in service-defaults, so I expand the PR to fix it. Could you help take another look? Thanks.

kisunji (approved):
LGTM

website/content/docs/connect/config-entries/service-defaults.mdx
```diff
@@ -687,6 +687,12 @@ represents a location outside the Consul cluster. They can be dialed directly wh
         },
       ]
     },
+    {
+      name: 'MaxInboundConnections',
+      description: 'The maximum number of concurrent inbound connections to each service instance.',
+      type: 'int: 0',
+      yaml: false,
```
DanStough: Just FYI, you'll want to come back and remove this when you add to the CRD.

api/config_entry.go
```diff
@@ -218,21 +218,22 @@ type UpstreamLimits struct {
...
+	Namespace             string                  `json:",omitempty"`
+	Protocol              string                  `json:",omitempty"`
+	Mode                  ProxyMode               `json:",omitempty"`
+	TransparentProxy      *TransparentProxyConfig `json:",omitempty" alias:"transparent_proxy"`
+	MeshGateway           MeshGatewayConfig       `json:",omitempty" alias:"mesh_gateway"`
+	Expose                ExposeConfig            `json:",omitempty"`
+	ExternalSNI           string                  `json:",omitempty" alias:"external_sni"`
+	UpstreamConfig        *UpstreamConfiguration  `json:",omitempty" alias:"upstream_config"`
+	Destination           *DestinationConfig      `json:",omitempty"`
+	MaxInboundConnections int
```
DanStough: Should this be an `*int` w/ `omitempty` to be backward compatible? Like this one: https://github.com/hashicorp/consul/blob/0e2eed12a2ef9e8c2b9b671a5112730f15531d59/agent/structs/config_entry.go#L991-L992

lkysow, line 233: Should this use `json:",omitempty" alias:"max_inbound_connections"`

DanStough: Although I see the `structs` package doesn't have this 🤔

DanStough (approved):
I see that the check in `listeners.go` for MaxInboundConnections>0, but I think it's a little weird to write it out all the time when deserializing the object. Since it's not in scope for this PR and you got a bunch of other approvals, I won't block it.

api/config_entry.go
```diff
@@ -218,21 +218,22 @@ type UpstreamLimits struct {
...
+	Namespace             string                  `json:",omitempty"`
+	Protocol              string                  `json:",omitempty"`
+	Mode                  ProxyMode               `json:",omitempty"`
+	TransparentProxy      *TransparentProxyConfig `json:",omitempty" alias:"transparent_proxy"`
+	MeshGateway           MeshGatewayConfig       `json:",omitempty" alias:"mesh_gateway"`
+	Expose                ExposeConfig            `json:",omitempty"`
+	ExternalSNI           string                  `json:",omitempty" alias:"external_sni"`
+	UpstreamConfig        *UpstreamConfiguration  `json:",omitempty" alias:"upstream_config"`
+	Destination           *DestinationConfig      `json:",omitempty"`
+	MaxInboundConnections int                     `json:",omitempty" alias:"max_inbound_connections"`
```
lkysow, line 233: It's not clear to me how we decide on using a pointer or not for these fields. In the upstream configs for example we sometimes use a pointer.

api/config_entry.go
kisunji: The `alias` always confuses me because AFAIK Go can marshal to snake_case fine (there are passing unit tests for snake case). So I wonder if the aliases are actually perpetuating a misunderstanding of Go's marshalling magic

huikang: I think we can keep the annotation here to make it consistent with the other fields. If we decide to remove them, we can do it altogether.

PR #14078, opened 2022-08-09
ui: don't surface partitions in services search source dropdown anymore

LevelbossMike:
### Description
Removes the partitions-selection in the sources-dropdown in the services search. Users can search Partitions by name in the search-input. After this change, the sources-dropdown in the services-searchbar will look like in the following screenshot:

Screenshot 2022-08-09 at 16 04 09

__This PR:__
* ~~changes the `Optgroup`-component to only surface the `@label` when we actually pass one - we can use that to not surface a label of an opt-group if we don't want to~~
* removes the partitions-selector in the sources-dropdown in the services search-bar

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-ui/app/components/consul/service/search-bar/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/service/search-bar/index.hbs
@@ -139,27 +139,8 @@ as |key value|}}
           </span>
         </BlockSlot>
         <BlockSlot @name="options">
-  {{#let components.Optgroup components.Option as |Optgroup Option|}}
-{{#let
-  (reject-by 'Partition' @partition @partitions)
-as |nonDefaultPartitions|}}
-{{#if (gt nonDefaultPartitions.length 0)}}
-    <Optgroup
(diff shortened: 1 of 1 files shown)


ui/packages/consul-ui/app/components/consul/service/search-bar/index.hbs
```diff
@@ -140,26 +140,8 @@ as |key value|}}
...
-    {{/each}}
-    </Optgroup>
-{{/if}}
-{{/let}}
-
 {{#if (gt @sources.length 0)}}
-    <Optgroup
-      @label={{t 'common.search.integrations'}}
-    >
+    <Optgroup>
```
johncowen, line 144: Pretty sure you can just remove the optgroup here, which will mean the amend made underneath in this PR in `ui/packages/consul-ui/app/components/popover-select/optgroup/index.hbs` isn't required. Optgroups are only really used to provide a label to the group, otherwise they are unnecessary.

evrowe (approved):
Looks good, :shipit:

PR #14079, opened 2022-08-09
Add retry in TestAgentConnectCALeafCert_good

kisunji:
### Description
There is a race between the CA being updated and the server responding to the leaf cert request. Added retries to make the test more consistent.

Similar retry loop exists in `TestAgentConnectCALeafCert_goodNotLocal`
https://github.com/hashicorp/consul/blob/61f86790115339972c4f6ad7328dd1ea3ccb73e8/agent/agent_endpoint_test.go#L6956-L6979

### DIFF
--- a/agent/agent_endpoint_test.go
+++ b/agent/agent_endpoint_test.go
@@ -6799,7 +6799,7 @@ func TestAgentConnectCALeafCert_good(t *testing.T) {
 	ca2 := connect.TestCAConfigSet(t, a, nil)
 
 	// Issue a blocking query to ensure that the cert gets updated appropriately
-	{
+	t.Run("test blocking queries update leaf cert", func(t *testing.T) {
 		resp := httptest.NewRecorder()
 		req, _ := http.NewRequest("GET", "/v1/agent/connect/ca/leaf/test?index="+index, nil)
 		a.srv.h.ServeHTTP(resp, req)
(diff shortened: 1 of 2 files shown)


agent/agent_endpoint_test.go
```diff
@@ -6834,33 +6834,26 @@ func TestAgentConnectCALeafCert_good(t *testing.T) {
...
-			requireLeafValidUnderCA(t, issued2, ca3)
+			retry.Run(t, func(r *retry.R) {
+				resp := httptest.NewRecorder()
+				a.srv.h.ServeHTTP(resp, req)
 
-			// Should not be a cache hit!
-			require.Equal(t, "MISS", resp.Header().Get("X-Cache"))
-		}
+				// Should not be a cache hit!
+				require.Equal(r, "MISS", resp.Header().Get("X-Cache"))
```
dhiaayachi, line 6845: I guess it's expected that we have few cache hits before updating, shouldn't we make this in its own retry loop to increase the change of the test passing?

dhiaayachi (approved):
Thank you for fighting flakes @kisunji !!
LGTM! I left a comment around cache hit/miss, if my understanding is right it would make this pass even more often.

agent/agent_endpoint_test.go
kisunji: I think something about the design of leaf cache or the test makes it impossible to get a cache HIT. I tried making a subtest for hitting cache and never got it to pass.

kisunji: There's a test below called `TestAgentConnectCALeafCert_nonBlockingQuery_after_blockingQuery_shouldNotBlock` which does assert for a cache hit; I didn't want to spend too much time looking at this 😆

dhiaayachi: No worries, it was just an idea. Let's see if it's still failing before trying to over optimize 😄

PR #14081, opened 2022-08-09
agent: Close existing active listeners when agent fails on an address

kisunji:
### Description
If `startListeners` successfully created listeners for some of its input addresses but eventually failed, the function would return an error and existing listeners would not be cleaned up.

* Added testcase which fails without this PR

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


boxofrad (approved):
This is awesome! 👏🏻

mkeeler (approved):

PR #14111, opened 2022-08-10
ui: Improve display peer info in service list

LevelbossMike:
### Description
This PR changes the following:

* Show namespace when peer info is surfaced - Whenever we display a peer and we are not on OSS we will surface the namespace as well. The rest of the ui logic of the bucket list has not changed.
* Display peer-info after instance-count

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-ui/app/components/consul/bucket/list/index.js
+++ b/ui/packages/consul-ui/app/components/consul/bucket/list/index.js
@@ -58,7 +58,7 @@ export default class ConsulBucketList extends Component {
 
   get namespacePart() {
     const { item, nspace } = this.args;
-    const { abilities, partitionPart } = this;
+    const { abilities, partitionPart, peerPart } = this;
 
     const nspaceItem = {
       type: 'nspace',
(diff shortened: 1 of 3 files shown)


ui/packages/consul-ui/app/components/consul/bucket/list/index.js
```diff
@@ -71,15 +71,13 @@ export default class ConsulBucketList extends Component {
...
     if (nspace && abilities.can('use nspaces')) {
       if (item.Namespace !== nspace) {
-        return [
-          {
-            type: 'nspace',
-            label: 'Namespace',
-            item: item.Namespace,
-          },
-        ];
+        return [nspaceItem];
```
evrowe, line 80: Would it be correct to assume that the structure of `nspaceItem` the same as the object that was being returned previously (or that it at least has the same keys)?

evrowe (approved):
Left a non-blocking question, looks good

ui/packages/consul-ui/app/components/consul/bucket/list/index.js
LevelbossMike: yes - we could have returned `nspaceItem` before. I realized that when looking at the code again today.

PR #14119, opened 2022-08-10
Fix peering establishment bugs on followers

kisunji:
### Description
1. `isFailedPreconditionErr` did not properly handle wrapped status errors
2. PeerStreamService was not registered with the internal gRPC server, making gRPC forwarding to leaders throw `rpc error: code = Unimplemented desc = unknown service hashicorp.consul.internal.peerstream.PeerStreamService`

* Added testcase that will break if grpc library ever changes its implementation

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/consul/leader_peering_test.go
+++ b/agent/consul/leader_peering_test.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/armon/go-metrics"
 	"github.com/hashicorp/go-hclog"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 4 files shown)


agent/consul/server.go
```diff
@@ -816,6 +816,7 @@ func newGRPCHandlerFromConfig(deps Deps, config *Config, s *Server) connHandler
 
 		// Note: these external gRPC services are also exposed on the internal server to
 		// enable RPC forwarding.
+		s.peerStreamServer.Register(srv)
```
freddygv, line 819: Can you add a test to ensure that forwarding is working? Here's an example:
https://github.com/hashicorp/consul-enterprise/blob/main/agent/consul/partition_backend_test.go#L84

agent/consul/leader_peering_test.go
```diff
@@ -1332,3 +1333,13 @@ func TestLeader_Peering_retryLoopBackoffPeering_cancelContext(t *testing.T) {
 		fmt.Errorf("error 1"),
 	}, allErrors)
 }
+
+func Test_isFailedPreconditionErr(t *testing.T) {
```
ndhanushkodi, line 1337: ❤️

agent/consul/server.go
kisunji: Added in [5078d6e](https://github.com/hashicorp/consul/pull/14119/commits/5078d6e0a25ee8bd3ffc8f792b10aac49752a913)

kisunji: Confirmed `unknown service` comes up without this change

freddygv (approved):
Thanks for the quick fix! :shipit:

PR #14132, opened 2022-08-10
Leadership transfer cmd

dhiaayachi:
### Description
This is an attempt to implement a command that would trigger a raft leadership transfer.
The idea about this trace back to a customer issue where we needed to trigger leader transfer to be able to upgrade a leader node.

This dependent to a change in raft that fix some of the gaps in leadership transfer, those gaps would make the transfer fail. https://github.com/hashicorp/raft/pull/487

I made an attempt in naming the sub command but open to any suggestions.

Added tests:
- RPC call without and with ACL
- API call negative test (with one agent)
- CMD call negative test (with one agent)


dhiaayachi:
@mkeeler I undusted this PR, if you happen to have some spare time. I had to create a new one and close https://github.com/hashicorp/consul/pull/12182 because of some git weirdness but I went though all your comments 

>Could we add one more test for OSS to verify the ACL enforcement and one enterprise test to verify that validateEnterpriseToken enforcement

Added an OSS test I will add one for enterprise after it get merged.

>Instead of structs.DCSpecificRequest could we use a new request type which also includes an optional parameter representing the raft.ServerID of a target server to transfer leadership to.

Added the ID to the request and threaded through to the API and the CLI.

>We could call LeadershipTransfer on Raft itself. This other function is no longer necessary as every 1.9+ cluster will unconditionally use leadership transfers for autopilot purposes so we can rely on it being available here.

I changed `attemptLeadershipTransfer` to be able to call the raft API with or without an ID. I left the version check, I can remove it if you think it's not useful anymore

jkirschner-hashicorp:
@dhiaayachi : what are your thoughts on aligning the naming of this command more with the existing Vault name for similar functionality: `step-down` rather than `leader-transfer`?

Vault provides an operator-level [CLI command](https://www.vaultproject.io/docs/commands/operator/step-down) and [HTTP API](https://www.vaultproject.io/api-docs/system/step-down) endpoint for forcing the current leader to gracefully step down. Someone familiar with this capability in Vault might look for something with the same name in Consul (internally, at least one person has asked if Consul had a similar "step down" capability).

That said, I can see the advantages of `consul operator raft leader-transfer` instead of `consul operator raft step-down`. Semantically, it's not _raft_ that's stepping down... the leader is stepping down. Perhaps we could keep the name `leader-transfer` but, in the docs, mention that this is analogous to `vault operator step-down` in Vault?

Just wanted to share these thoughts with you to consider.

---

Side question: when running this command, can the leader immediately be re-elected? Or is a different server agent always elected (if available)?

.changelog/14132.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
+raft: add an operator api endpoint and a command to initiate raft leadership transfer.
```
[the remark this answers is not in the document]
dhiaayachi: yes definitely, this is not part of our planned work, so I'm doing at my pace.

agent/consul/operator_raft_endpoint.go
```diff
@@ -2,15 +2,43 @@ package consul
...
+	}
+	// This action requires operator read access.
+	authz, err := op.srv.ResolveToken(args.Token)
+	if err != nil {
+		return err
+	}
+	if err := op.srv.validateEnterpriseToken(authz.Identity()); err != nil {
+		return err
+	}
+	if authz.OperatorRead(nil) != acl.Allow {
```
jkirschner-hashicorp, line 30: Should this be `OperatorWrite`? Or `OperatorRead`?

dhiaayachi: yes definitely!

agent/consul/operator_raft_endpoint_test.go
```diff
@@ -15,9 +15,136 @@ import (
...
+	retry.Run(t, func(r *retry.R) {
+		beforeLeader := s1.raft.Leader()
+		require.NotEmpty(r, beforeLeader)
+	})
+	beforeLeader := s1.raft.Leader()
+	require.NotEmpty(t, beforeLeader)
+
+	var reply structs.LeadershipTransferResponse
+	arg := structs.DCSpecificRequest{
+		Datacenter: "dc1",
```
mkeeler, line 104: It would be nice to create a token with `operator:read` privileges and ensure that it fails with that and not just fails due to the anon token having no privileges.

agent/consul/operator_raft_endpoint.go
dhiaayachi: I will also make sure that the ACL test use a token with the right access rights to catch this

agent/consul/server.go
```diff
@@ -1170,20 +1170,32 @@ func (s *Server) Shutdown() error {
...
 
-func (s *Server) attemptLeadershipTransfer() (success bool) {
+func (s *Server) attemptLeadershipTransfer(id raft.ServerID) (err error) {
 	leadershipTransferVersion := version.Must(version.NewVersion(LeaderTransferMinVersion))
 
 	ok, _ := ServersInDCMeetMinimumVersion(s, s.config.Datacenter, leadershipTransferVersion)
 	if !ok {
-		return false
+		return fmt.Errorf("minimum version not supported, minimu required version %s", LeaderTransferMinVersion)
 	}
```
mkeeler, line 1180: This whole block can be deleted. Everything at least as far back as 1.9 will support this and we will not support <1.9 to 1.14 upgrades directly without passing through 1.9, 1.10, 1.11, 1.12, or 1.13

agent/structs/structs.go
```diff
@@ -2871,6 +2872,18 @@ func (r *KeyringRequest) RequestDatacenter() string {
...
+	Datacenter         string
+	ID                 raft.ServerID
+	acl.EnterpriseMeta `hcl:",squash" mapstructure:",squash"`
+	QueryOptions
+}
+
+func (l LeaderTransferRequest) RequestDatacenter() string {
+	return l.Datacenter
+}
+
```
mkeeler, line 2886: Should this live in agent/structs/operator.go along with the request type.

mkeeler (commented):
The code generally looks good. 

The biggest concern is that we would be adding additional net-new RPC endpoints that are not gRPC. I would love to see protobuf definitions for the request/response types and a gRPC service. The HTTP/cli bits as well as the internal `attemptLeaderTransfer` code shouldn't need alteration.

agent/consul/operator_raft_endpoint_test.go
dhiaayachi: Added a test for that in the ACL test

command/registry.go
```diff
@@ -214,6 +215,7 @@ func RegisteredCommands(ui cli.Ui) map[string]mcli.CommandFactory {
 		entry{"operator raft", func(cli.Ui) (cli.Command, error) { return operraft.New(), nil }},
 		entry{"operator raft list-peers", func(ui cli.Ui) (cli.Command, error) { return operraftlist.New(ui), nil }},
 		entry{"operator raft remove-peer", func(ui cli.Ui) (cli.Command, error) { return operraftremove.New(ui), nil }},
+		entry{"operator raft leader-transfer", func(ui cli.Ui) (cli.Command, error) { return leadertransfer.New(ui), nil }},
```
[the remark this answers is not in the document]
dhiaayachi: did the rename to transfer-leader, let me know if that make sense

dhiaayachi:
@mkeeler @jkirschner-hashicorp I did some changes to this based on your comments. Please take a look when you have time.

@mkeeler I implemented this as a GRPC endpoint. I tried to make it in a way that we can move all the operators endpoints to it but it's my first attempt implementing such an endpoint in Consul, let me know if I missed some spots.

mkeeler (commented):
The grpc service looks great. One last request I would make would be to add some unit tests within the agent/rpc/operator package utilizing a mock backend.

The tests we would want to see are that acl authorization works as expected, rpc forwarding works and that it makes the appropriate Transfer leader request. The public connectca.Sign gRPC endpoint tests provide some good examples: https://github.com/hashicorp/consul/blob/main/agent/grpc-external/services/connectca/sign_test.go

agent/rpc/operator/service_test.go
```diff
@@ -0,0 +1,278 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+type MockBackend struct {
+	mock.Mock
+	authorizer acl.Authorizer
+}
+
+type mockAuthorizer struct {
```
mkeeler, line 23: What would you think about moving this to an acl/acltest package and allowing the MockAuthorizer to be reused from many places?

Also the acl.Authorizer interface has extra methods in enterprise so this one wont actually meet that interface when this gets merged.

mkeeler (approved):
I think there will be some issues with an enterprise merge that I mentioned in a comment. Otherwise this LGTM

agent/rpc/operator/service_test.go
dhiaayachi: I was thinking of reusing the same mock but did not find a good place to move it to be shareable. I think acl/acltest make sense. I will do so.

.changelog/14132.txt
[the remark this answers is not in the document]
dhiaayachi: @jkirschner-hashicorp Doc changes are in! if you would like to give it a read :)

website/content/api-docs/operator/raft.mdx
```diff
@@ -145,3 +145,41 @@ $ curl \
...
+[blocking queries](/api-docs/features/blocking),
+[consistency modes](/api-docs/features/consistency),
+[agent caching](/api-docs/features/caching), and
+[required ACLs](/api-docs#authentication).
+
+| Blocking Queries | Consistency Modes | Agent Caching | ACL Required     |
+| ---------------- | ----------------- | ------------- | ---------------- |
+| `NO`             | `none`            | `none`        | `operator:write` |
+
+The corresponding CLI command is [`consul operator raft tranfer-leader`](/commands/operator/raft#tranfer-leader).
```
jkirschner-hashicorp, line 172:
```suggestion
The corresponding CLI command is [`consul operator raft transfer-leader`](/commands/operator/raft#transfer-leader).
```

website/content/api-docs/operator/raft.mdx
```diff
@@ -145,3 +145,41 @@ $ curl \
...
+
+### Query Parameters
+
+- `id` `(string: "")` - Specifies the ID of the raft peer to transfer leadership to, if empty the leadership
+will be transferred to a random server.
```
jkirschner-hashicorp, line 177: Is the "ID" here a "node ID"? Or some either kind of ID? We should make that explicit.

```suggestion
- `id` `(string: "")` - Specifies the node ID of the raft peer to transfer leadership to.
If empty, leadership will be transferred to a random server agent.
```

website/content/commands/operator/raft.mdx
```diff
@@ -104,3 +104,29 @@ Usage: `consul operator raft remove-peer -address="IP:port"`
 - `-id` - ID of the server to remove.
 
 The return code will indicate success or failure.
+
+## transfer-leader
+
+Corresponding HTTP API Endpoint: [\[POST\] /v1/operator/raft/transfer-leader](/api-docs/operator/raft#transfer-leader)
```
jkirschner-hashicorp, line 110: I think the anchor tag needs to be corrected to work.

```suggestion
Corresponding HTTP API Endpoint: [\[POST\] /v1/operator/raft/transfer-leader](/api-docs/operator/raft#transfer-raft-leadership)
```

website/content/commands/operator/raft.mdx
```diff
@@ -104,3 +104,29 @@ Usage: `consul operator raft remove-peer -address="IP:port"`
...
+
+This command transfer the Consul leader to another server. If an `id` is provided, raft leadership will be transferred
```
jkirschner-hashicorp, line 112:
```suggestion
This command transfer the Consul leader to another server agent. If an `id` is provided, raft leadership will be transferred
```

website/content/commands/operator/raft.mdx
```diff
@@ -104,3 +104,29 @@ Usage: `consul operator raft remove-peer -address="IP:port"`
...
+to the server with that id.
+
+[`consul members`](/commands/members) command, it is preferable to
+clean up by simply running
+[`consul force-leave`](/commands/force-leave)
+instead of this command.
+
```
jkirschner-hashicorp, line 119: Is this supposed to be here?

```suggestion
[`consul members`](/commands/members) command, it is preferable to
clean up by simply running
[`consul force-leave`](/commands/force-leave)
instead of this command.

```

website/content/commands/operator/raft.mdx
```diff
@@ -104,3 +104,29 @@ Usage: `consul operator raft remove-peer -address="IP:port"`
...
+[blocking queries](/api-docs/features/blocking) and [agent caching](/api-docs/features/caching)
+are not supported from commands, but may be from the corresponding HTTP endpoint.
+
+| ACL Required     |
+| ---------------- |
+| `operator:write` |
+
+Usage: `consul operator raft transfer-leader -id="server id"`
+
+- `-id` - ID of the server to transfer leadership to.
```
jkirschner-hashicorp, line 130:
```suggestion
- `-id` - Specifies the node ID of the raft peer to transfer leadership to.
If empty, leadership will be transferred to a random server agent.
```

website/content/commands/operator/raft.mdx
```diff
@@ -104,3 +104,29 @@ Usage: `consul operator raft remove-peer -address="IP:port"`
...
+
+The return code will indicate success or failure.
```
jkirschner-hashicorp, line 132: Do we mention this for other commands? Or is this implied (not stated) because it's true for all commands?

website/content/commands/operator/raft.mdx
jkirschner-hashicorp, line 112: Is it worthwhile to mention _why_ someone might want to use this command?

For example:
"This command provides a way to transition leadership without restarting the leader node, maintaining quorum and workload capacity."

website/content/api-docs/operator/raft.mdx
```diff
@@ -145,3 +145,41 @@ $ curl \
...
 ```
+
+## Transfer Raft Leadership
+
+This endpoint transfer the raft leadership from the current raft leader to a new node or
+the node with the provided id.
+
+
+If ACLs are enabled, the client will need to supply an ACL Token with `operator`
+write privileges.
```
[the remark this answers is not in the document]
dhiaayachi: probably not, you're right!

website/content/commands/operator/raft.mdx
dhiaayachi: copy-pasta!

website/content/commands/operator/raft.mdx
dhiaayachi: I copy pasted this from another command just above

website/content/commands/operator/raft.mdx
```diff
@@ -104,3 +104,30 @@ Usage: `consul operator raft remove-peer -address="IP:port"`
...
+
+Corresponding HTTP API Endpoint: [\[POST\] /v1/operator/raft/transfer-leader](/api-docs/operator/raft#transfer-raft-leadership)
+
+This command transfer the Consul leader to another server agent. If an `id` is provided, raft leadership will be transferred
+to the server with that id.
+
+[`consul members`](/commands/members) command, it is preferable to
+clean up by simply running
+[`consul force-leave`](/commands/force-leave)
+instead of this command.
```
dhiaayachi, line 118:
```suggestion
```

jkirschner-hashicorp:
The leadership transfer command will be available starting in Consul 1.15.0!

PR #14145, opened 2022-08-11
Replace FSM’s old registry pattern with newly suggested pattern

Mathew-Estafanous:
issue #11740

As suggested by the linked issue, I transitioned to the new pattern for the FSM operations and snapshot registries. I wasn’t sure if any of the registries needed enterprise mapping as well, so I went ahead and created a noop function for each registry where enterprise mappings can be added.

This PR is a continuation of a previously merged PR, #12729


Mathew-Estafanous:
I'm not sure what the cause of the failed integration tests is. It seems unrelated to my commit.

Could someone rerun the tests to see if this was a one-off error?

Mathew-Estafanous:
I mistakenly dropped my commits and pushed it, resulting in the PR getting automatically merged since there were no changes.

#14266 includes the intended commits

Issue #14107, opened 2022-08-10
Failed to start up the cluster when upgrading to 1.13.0


tuxillo:
#### Overview of the Issue

Upgrading from 1.12.3 to 1.13.0 leads to a failure to start the cluster.

#### Reproduction Steps

Steps to reproduce this issue, eg:

1. We have a 3-node cluster 
1. Upgrade nodes to 1.13.0
1. The nodes won't start up due to an error (see log fragments below)

### Consul info for both Client and Server

<details>
  <summary>Client info</summary>

```
$ consul info
Error querying agent: Get "https://localhost:8501/v1/agent/self": dial tcp [::1]:8501: connect: connection refused
```

</details>

<details>
  <summary>Server info</summary>

```
$ consul info
Error querying agent: Get "https://localhost:8501/v1/agent/self": dial tcp [::1]:8501: connect: connection refused
```

</details>

### Operating system and Environment details

Centos 7 LXC running on Proxmox.

Linux consul-01.mysite 5.11.22-5-pve #1 SMP PVE 5.11.22-10 (Tue, 28 Sep 2021 08:15:41 +0200) x86_64 x86_64 x86_64 GNU/Linux

### Log Fragments

```
Aug 10 07:40:59 consul-01.mysite consul[1683882]: ==> Starting Consul agent...
Aug 10 07:40:59 consul-01.mysite consul[1683882]:            Version: '1.13.0'
Aug 10 07:40:59 consul-01.mysite consul[1683882]:         Build Date: '2022-08-09 18:28:11 +0000 UTC'
Aug 10 07:40:59 consul-01.mysite consul[1683882]:            Node ID: '1aaaf50c-d91e-8d64-ec70-ed9d68341ff3'
Aug 10 07:40:59 consul-01.mysite consul[1683882]:          Node name: 'consul-01'
Aug 10 07:40:59 consul-01.mysite consul[1683882]:         Datacenter: 'mysite' (Segment: '<all>')
Aug 10 07:40:59 consul-01.mysite consul[1683882]:             Server: true (Bootstrap: false)
Aug 10 07:40:59 consul-01.mysite consul[1683882]:        Client Addr: [0.0.0.0] (HTTP: 8500, HTTPS: 8501, gRPC: 8502, DNS: 8600)
Aug 10 07:40:59 consul-01.mysite consul[1683882]:       Cluster Addr: 192.168.10.11 (LAN: 8301, WAN: 8302)
[...]

PR #14149, opened 2022-08-11
Handle breaking change for ServiceVirtualIP restore

kisunji:
fixes https://github.com/hashicorp/consul/issues/14107

### Description
Consul 1.13.0 changed ServiceVirtualIP to use PeeredServiceName instead of ServiceName which was a breaking change for those using service mesh and wanted to restore their snapshot after upgrading to 1.13.0.

This commit handles existing data with older ServiceName and converts it during restore so that there are no incompatibility issues when restoring from older snapshots.

* Manually tested in OSS and ENT by making a snapshot using Consul 1.12.3, then attempting to restore with the changes in this branch.
* Added unit tests restoring both versions of the struct

### DIFF
--- a/agent/structs/structs.go
+++ b/agent/structs/structs.go
@@ -2211,8 +2211,8 @@ type PeeredServiceName struct {
 }
 
 type ServiceName struct {
-	Name string
-	acl.EnterpriseMeta
+	Name               string
+	acl.EnterpriseMeta `mapstructure:",squash"`
 }
 
 func NewServiceName(name string, entMeta *acl.EnterpriseMeta) ServiceName {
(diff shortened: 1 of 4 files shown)


agent/structs/structs.go
```diff
@@ -2211,8 +2211,8 @@ type PeeredServiceName struct {
 }
 
 type ServiceName struct {
-	Name string
-	acl.EnterpriseMeta
+	Name               string
+	acl.EnterpriseMeta `mapstructure:",squash"`
```
kisunji, line 2215: Noticed in testing that `mapstructure` does not handle embedded structs without this tag

mkeeler (commented):
The code looks good. Would it be possible to add a test?

kisunji:
> The code looks good. Would it be possible to add a test?

Yep I'll try to write a test for this

mkeeler (approved):

PR #14191, opened 2022-08-12
peering: update docs with exposing servers

ndhanushkodi:
And update a helm value


david-yu:
Could you rebased, and also change references of 0.47.0 to 0.47.1? Thanks!

david-yu (approved):

Issue #13494, opened 2022-06-19
Some Consul metrics are reported with a doubled prefix: consul.consul....


maxb:
Consul includes a `metrics_prefix` setting, which is added as a prefix to every metric name, through the built-in functionality of the underlying https://github.com/armon/go-metrics library.

The default value is set here: https://github.com/hashicorp/consul/blob/edbf19f4e88e81cda0fa52bbbb463cf1d686fe12/agent/config/default.go#L129

As you can see, the default value is `consul`. Most of the metrics defined throughout Consul therefore are defined expecting this will be added - e.g.: https://github.com/hashicorp/consul/blob/9e986c62e4d6500e10422f941d159a2327adbeca/agent/consul/session_ttl.go#L152-L155

However some of them are defined ignoring that this prefixing behaviour exists - e.g.: https://github.com/hashicorp/consul/blob/1e02460bd1a5655093928ea2927901a71a37060a/agent/consul/usagemetrics/usagemetrics_oss.go#L73

with the result that when the operator of the Consul cluster gets to see the above metric, it is actually called `consul.consul.state.kv_entries` - or, given the name mapping applied to metrics when using Prometheus, `consul_consul_state_kv_entries`.

This is confusing, and the discrepancy is not mentioned in the docs - <https://www.consul.io/docs/agent/telemetry> - which document metrics without the unintentional doubling - just `consul.state.kv_entries`.

This mistake should be corrected - but there are compatibility concerns. I am willing to create a PR - this issue is for the purpose of discussing the compatibility mechanism prior to implementation.

I am proposing to duplicate the approach already in use for the `consul.http.` -> `consul.api.http.` transition - write both metrics for an overlap period, and allow transitional control via the config file option `telemetry { disable_compat_1.9 = true }`.

A question for HashiCorp release management people: I see Consul 1.13.0-alpha1 has been released - can the initial steps, i.e.:

* Adding new correctly named versions of the metrics
* Adding new config option `telemetry { disable_compat_1.13 }` with initial default value `false`
* Declaring old `consul.consul.*` metrics deprecated and to be removed in a future version

still target 1.13 at this point, or do they need to target 1.14?

maxb:
The above-mentioned PR #13497 is not intended to be merged, as it contains no compatibility handling, but serves to display the metrics code needing changes.

Amier3:
Hey @maxb 

[...]

Issue #13495, opened 2022-06-19
Consul emits extra dummy metrics to Prometheus which always have zero values


maxb:
# Example

```
# HELP consul_consul_state_services Measures the current number of unique services registered with Consul, based on service name. It is only emitted by Consul servers. Added in v1.9.0.
# TYPE consul_consul_state_services gauge
consul_consul_state_services 0
consul_consul_state_services{datacenter="dc1"} 1
```

# Explanation

The real values for `consul_consul_state_services` are emitted with a `datacenter` label.

However there is an extra metric with an empty label set, which always has value zero, and is never updated.

It would be nice to remove this metric, however the underlying go-metrics library in use, does not allow this, without also sacrificing the help message. I have opened https://github.com/armon/go-metrics/issues/136 about this.

The primary purpose of opening this Consul issue in the short term, is to:
* Record that this quirk exists and IMO would be worth documenting in the "Telemetry" documentation
* Provide an URL that can be referenced in comments in the source code, to provide a central explanation why various metrics are defined but never used.

jakubgs:
Just noticed this issue when I tried to enable metrics for my cluster. The result is that I see zeroes in the queries:

![image](https://user-images.githubusercontent.com/2212681/184848945-e6fea71a-09d1-4aa2-b630-df2399619db1.png)

PR #14193, opened 2022-08-12
Standardize on non-double-prefixed metrics

skpratt:
### Description
We have dueling metrics patterns - sometimes we hard-code the "consul" prefix, but then we also automatically prepend it again to all metrics. See [Issue 13494](https://github.com/hashicorp/consul/issues/13494) for more info.

This PR takes the first step toward reconciling these patterns by establishing reliance on auto-prepending prefixes as the preferred pattern, and declaring existing metrics that hard-code prefixes as deprecated in help strings.

It applies the duplication tactic in examples such as https://github.com/hashicorp/consul/blob/c04515a844cf9348198fb16c637466604a94fc09/agent/consul/fsm/commands_oss.go#L351 to all legacy metrics.

* This does not eliminate double-prefixed metrics, but it adds a single-prefixed metric for every double-prefixed metrics, so that they can be eventually deleted.
* Tests updated to check for new metrics.

### Links
[Issue 13494](https://github.com/hashicorp/consul/issues/13494)

### PR Checklist

* [x] updated test coverage
* [N/A] external facing docs updated (this change makes metrics consistent with docs)

### DIFF
--- a/agent/cache/cache.go
+++ b/agent/cache/cache.go
@@ -37,6 +37,10 @@ import (
 var Gauges = []prometheus.GaugeDefinition{
 	{
 		Name: []string{"consul", "cache", "entries_count"},
+		Help: "Deprecated - please use cache_entries_count instead.",
+	},
+	{
+		Name: []string{"cache", "entries_count"},
 		Help: "Represents the number of entries in this cache.",
 	},
 }
@@ -45,18 +49,34 @@ var Gauges = []prometheus.GaugeDefinition{
 var Counters = []prometheus.CounterDefinition{
 	{
 		Name: []string{"consul", "cache", "bypass"},
+		Help: "Deprecated - please use cache_bypass instead.",
+	},
+	{
+		Name: []string{"cache", "bypass"},
 		Help: "Counts how many times a request bypassed the cache because no cache-key was provided.",
 	},
 	{
 		Name: []string{"consul", "cache", "fetch_success"},
+		Help: "Deprecated - please use cache_fetch_success instead.",
+	},
+	{
+		Name: []string{"cache", "fetch_success"},
 		Help: "Counts the number of successful fetches by the cache.",
 	},
 	{
 		Name: []string{"consul", "cache", "fetch_error"},
+		Help: "Deprecated - please use cache_fetch_error instead.",
+	},
+	{
+		Name: []string{"cache", "fetch_error"},
 		Help: "Counts the number of failed fetches by the cache.",
 	},
 	{
 		Name: []string{"consul", "cache", "evict_expired"},
+		Help: "Deprecated - please use cache_evict_expired instead.",
+	},
+	{
+		Name: []string{"cache", "evict_expired"},
 		Help: "Counts the number of expired entries that are evicted.",
 	},
 }
@@ -397,6 +417,7 @@ func entryExceedsMaxAge(maxAge time.Duration, entry cacheEntry) bool {
 func (c *Cache) getWithIndex(ctx context.Context, r getOptions) (interface{}, ResultMeta, error) {
 	if r.Info.Key == "" {
 		metrics.IncrCounter([]string{"consul", "cache", "bypass"}, 1)
+		metrics.IncrCounter([]string{"cache", "bypass"}, 1)
 
 		// If no key is specified, then we do not cache this request.
 		// Pass directly through to the backend.
@@ -443,6 +464,7 @@ RETRY_GET:
 		meta := ResultMeta{Index: entry.Index}
 		if first {
 			metrics.IncrCounter([]string{"consul", "cache", r.TypeEntry.Name, "hit"}, 1)
+			metrics.IncrCounter([]string{"cache", r.TypeEntry.Name, "hit"}, 1)
 			meta.Hit = true
 		}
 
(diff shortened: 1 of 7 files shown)


agent/cache/cache.go
```diff
@@ -37,6 +37,10 @@ import (
 var Gauges = []prometheus.GaugeDefinition{
 	{
 		Name: []string{"consul", "cache", "entries_count"},
+		Help: "Deprecated - please use cache_entries_count instead.",
```
acpana, line 40: Nice! I think ideally, this is all that is needed.

I think, in practice, IIRC, these messages only get "emitted" on the dummy metrics like https://github.com/hashicorp/consul/issues/13495 . Apologies if you are already well aware of this but I think for real world metrics (especially label based), the help message is not populate properly. But that's not something to be fixed in `consul`, rather something to get fixed in `go-metrics` I think.

agent/consul/leader_peering.go
```diff
@@ -30,8 +31,13 @@ import (
 	"github.com/hashicorp/consul/proto/pbpeerstream"
 )
 
+var leaderExportedServicesCountKeyDeprecated = []string{"consul", "peering", "exported_services"}
 var leaderExportedServicesCountKey = []string{"consul", "peering", "exported_services"}
```
acpana, line 35: should line 33 be without the `consul` prefix then ?

Also 🤦🏼 since I added that code line and forgot about the double prefixing issue.

Again, not sure if this is widely known so please bear with me as I brain dump something about prefixing these w the service name, ie `consul`.

This is mostly a Promethues thing, so this func runs over all metrics (gauges, counters, summaries) and prepends the service name: https://github.com/hashicorp/consul/blob/main/agent/setup.go#L192-L347 

But that only works if they are predeclared somewhere like `LeaderPeeringMetrics` below.

So it may be possible that some of these metrics are never predeclared (not the peering metrics here but some of the older ones). In that case, emitting the `consul` prefix might not be ideal. I've not spent time looking at which metrics don't actually get predeclared so I am just offering this as a warning -- apologies in advance if it's a wild goose chase instead.

agent/consul/usagemetrics/usagemetrics_oss.go
```diff
@@ -25,6 +30,11 @@ func (u *UsageMetricsReporter) emitPeeringUsage(peeringUsage state.PeeringUsage)
 		float32(peeringUsage.Peerings),
 		u.metricLabels,
 	)
+	metrics.SetGaugeWithLabels(
```
acpana, line 33: nice! since these are all additive changes, they merge cleanly OSS-->ENT but it would be nice if we had ENT tests mirroring this. If that adds any value of course.

acpana (commented):
This looks great !! 🥇 thanks for taking it on sarah 💯 

---

I left a couple of brain dumpy, non actionable comments.

Also, for this PR, did you want to create a changelog? I think it could be useful to call out these changes (deprecating the metrics, emitting 2x for deprecated metrics). Totally fine to add it as a follow up PR too but I think it'd be cleaner to have it in here.

Finally, I did a search and found a few more metrics prefixed with `"consul", "fsm"` in `commands_oss.go`. Not sure if those qualify for this patch or if we wanted to tackle those as well.

agent/cache/cache.go
skpratt: Yep. That will have to be fixed separately.

agent/consul/usagemetrics/usagemetrics_oss.go
skpratt: Will add in ENT project

agent/consul/leader_peering.go
skpratt: thanks for that catch! it looks like all of the metrics updated in this PR are predeclared.  But enforcing that should be a follow up to this.

skpratt:
> 

Great points! Thank you for adding them here, it will greatly help with guiding follow up work.

agent/cache/cache.go
eculver: Should we get this fixed in `go-metrics` first so that we can deliver the help message?

agent/consul/usagemetrics/usagemetrics_oss.go
eculver: Alternatively, if these are additive and the ENT tests aren't any different, can we just add them here?

agent/cache/cache.go
skpratt: there is a workaround for getting the help strings - as acpana pointed to in #13495 - and I think we should consider that a separate issue with bigger scope

agent/consul/usagemetrics/usagemetrics_oss.go
skpratt: It looks like the ENT tests will need updates, similar to the test additions here.

eculver (approved):
After some discussion with @skpratt offline, I think this is good to go. We discussed updating the code to add `TODO`s for the next steps but it seems like creating issues for these follow-ups is a better place to capture them. We also decided that since this whole process of getting to a world without double-prefixed metrics is going to be a long one, we might as well not make any false promises in the changelog and/or docs as to when this would actually come to fruition.

TL;DR - looks great!

PR #14161, opened 2022-08-11
feat(telemetry): add labels to serf and memberlist metrics

huikang:
### Description

To help user measure the gossip stability in the cluster, we need to separate the measurements between segment or partitions (for Enterprise). This PR adds the label to the serf metrics.

For OSS, we only label lan and wan network

- Example 1: server with 2 segments and wan gossip

```
consul_memberlist_gossip_count{job="server-1", network="lan", partition="default"} | 442
consul_memberlist_gossip_count{job="server-1", network="lan", partition="default", segment="alpha"} | 442
consul_memberlist_gossip_count{job="server-1", network="lan", partition="default", segment="beta"} | 442
consul_memberlist_gossip_count{job="server-1", network="wan", partition="default"} | 177
```

- Example 2: server with 2 partitions and one client in team 1 partition

consul_memberlist_gossip_count{job="client-1", network="lan", partition="team1"} | 72
consul_memberlist_gossip_count{job="server-1", network="lan", partition="default"} | 217

### Links
[Asana ticket](https://app.asana.com/0/0/1202764564348923/f)

### PR Checklist

* [ ] updated test coverage
* [x] external facing docs updated
* [x] not a security concern

PR #14194, opened 2022-08-12
fix: missing segment and partition

huikang:
### Description
This is the follow-up PR to https://github.com/hashicorp/consul/pull/14161 to fix the missing segment and partition in client and server's serf setup for ent

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/agent/consul/server_serf.go
+++ b/agent/consul/server_serf.go
@@ -232,7 +232,7 @@ func (s *Server) setupSerfConfig(opts setupSerfOptions) (*serf.Config, error) {
 
 	conf.ReconnectTimeoutOverride = libserf.NewReconnectOverride(s.logger)
 
-	addSerfMetricsLabels(conf, opts.WAN, "", "", "")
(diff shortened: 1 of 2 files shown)


agent/consul/server_serf.go
```diff
@@ -232,7 +232,7 @@ func (s *Server) setupSerfConfig(opts setupSerfOptions) (*serf.Config, error) {
 
 	conf.ReconnectTimeoutOverride = libserf.NewReconnectOverride(s.logger)
 
-	addSerfMetricsLabels(conf, opts.WAN, "", "", "")
+	addSerfMetricsLabels(conf, opts.WAN, opts.Segment, s.config.AgentEnterpriseMeta().PartitionOrDefault(), "")
```
rboyer, line 235: This is not necessary since we omit segment/partition for WAN.

huikang: opts.WAN could be false; in that case we need the segment and partition in ENT, right?

rboyer: Oh right, I briefly interpreted that line as an enum-y thing.

rboyer (approved):

PR #14197, opened 2022-08-12
Proxy register port race 2 test

skpratt:
### Description

### Links

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/test/integration/connect/envoy/Dockerfile-consul-envoy
+++ b/test/integration/connect/envoy/Dockerfile-consul-envoy
@@ -1,7 +1,7 @@
 # Note this arg has to be before the first FROM
 ARG ENVOY_VERSION
 
-FROM consul-dev as consul
+FROM --platform=linux/amd64 consul-dev as consul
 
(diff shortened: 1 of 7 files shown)


test/integration/connect/envoy/Dockerfile-consul-envoy
```diff
@@ -1,7 +1,7 @@
 # Note this arg has to be before the first FROM
 ARG ENVOY_VERSION
 
-FROM consul-dev as consul
+FROM --platform=linux/amd64 consul-dev as consul
```
DanStough, line 4: I see that the default version of envoy is [1.23.0](https://github.com/hashicorp/consul/blob/d281c89ad08fc0966f115134f621eea79e96b8c9/test/integration/connect/envoy/run-tests.sh#L15-L16), but that version of the arm container is [broken](https://github.com/envoyproxy/envoy/issues/22261). I wonder if you used 1.22.4 if you'd need this change.

PR #14213, opened 2022-08-16
docs: Fix misleading phrase about X-Consul-Index header

kisunji:
### Description
For KVS GET requests which are non-recurse, the `ModifyIndex` does not correspond to the `X-Consul-Index` header, which is the max Raft index of the entire KVS table: https://github.com/hashicorp/consul/blob/main/agent/consul/state/kvs.go#L193

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/website/content/api-docs/kv.mdx
+++ b/website/content/api-docs/kv.mdx
@@ -107,8 +107,9 @@ $ curl \
 - `CreateIndex` is the internal index value that represents when the entry was
   created.
 
-- `ModifyIndex` is the last index that modified this key. It can be used to establish
-  blocking queries by setting the `?index` query parameter.
+- `ModifyIndex` is the last index that modified this key which can be used for
+  Check-and-Set operations on the key. For blocking queries, the `X-Consul-Index`
+  header value should be used to set the `?index` query parameter.
   You can even perform blocking queries against entire subtrees of the KV store:
   if `?recurse` is provided, the returned `X-Consul-Index` corresponds to the
   latest `ModifyIndex` within the prefix, and a blocking query using that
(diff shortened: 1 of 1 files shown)


mkeeler (changes requested):
hmm, while I agree the docs may be a little misleading, not advocating for populating the `index` query param with the value of X-Consul-Index could lead to a hot loop and degraded server performance / overloaded server.

The scenario I am thinking of is:

KV data exists at foo (index 3) and bar (index 5). The kv table index and value returned by X-Consul-Index is therefore 5.

1. User makes query for initial key: GET /kvs/foo - the ModifyIndex returned is 3
2. User takes a blocking query for the key: `GET /kvs/foo?index=3`.
3. Because the query index of 5 is greater than the index of 3 the data will be returned immediately without blocking.
4. Steps 2 and 3 repeat infinitely - hopefully there is some rate limiting involved or else this will consume tons of cpu time on the server handling the request.

I think a better alteration to the docs here would be to say that the ModifyIndex field may not be the appropriate blocking index and to instead rely on the value of the X-Consul-Index index.

mkeeler:
Also, arguably we should have a separate state store method for retrieving an individual key which returns a different blocking query index. That would prevent the hot loop behavior as well.

kisunji:
> Also, arguably we should have a separate state store method for retrieving an individual key which returns a different blocking query index. That would prevent the hot loop behavior as well.

Yeah I haven't looked too deeply but it should be possible, given KVS.List returns finer-grained indexes.

mkeeler (approved):

kisunji:
False alarm, the index is set correctly for a given key here: https://github.com/hashicorp/consul/blob/e00e3a0bc3e55a5468bdf3415973db910e6e50b4/agent/consul/kvs_endpoint.go#L169

Turns out the original issue (raised internally) applies to config entries, not KV

PR #14222, opened 2022-08-16
chore: add multi-arch docker build for testing

DanStough:
### Description
I wanted an easier way to build and push images that could be consumed anywhere, especially on M1 machines. I tried to add the commands in-place but there is a lot of intersection of CI and other `make` commands with this Dockerfile, so I duplicated the existing one to try and minimize the blast radius. 

Run: 
* `make dev-docker` for a local image
* ` REMOTE_DEV_IMAGE=ttl.sh/dans/consul-dev-666:24h make remote-docker` to push a remote multi-arch image

### Links
Approach is borrowed from [consul-k8s](https://github.com/hashicorp/consul-k8s/blob/43b4995e374e3197488624cb210e57e3334b6239/Makefile#L35-L45)

### PR Checklist

* [ ] ~updated test coverage~
* [X] not a security concern

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -134,6 +134,8 @@ ifdef SKIP_DOCKER_BUILD
 ENVOY_INTEG_DEPS=noop
 endif
 
+SYSTEM_ARCH?=$(shell uname -m)
+
 all: dev-build
 
 # used to make integration dependencies conditional
@@ -152,7 +154,26 @@ dev-docker: linux
 	@docker pull consul:$(CONSUL_IMAGE_VERSION) >/dev/null
 	@echo "Building Consul Development container - $(CONSUL_DEV_IMAGE)"
 	#  'consul:local' tag is needed to run the integration tests
-	@DOCKER_DEFAULT_PLATFORM=linux/amd64 docker build $(NOCACHE) $(QUIET) -t '$(CONSUL_DEV_IMAGE)' -t 'consul:local' --build-arg CONSUL_IMAGE_VERSION=$(CONSUL_IMAGE_VERSION) $(CURDIR)/pkg/bin/linux_amd64 -f $(CURDIR)/build-support/docker/Consul-Dev.dockerfile
+	docker buildx create --use && docker buildx build -t 'consul:local' \
+       --platform linux/$(SYSTEM_ARCH) \
+	   --build-arg CONSUL_IMAGE_VERSION=$(CONSUL_IMAGE_VERSION) \
+       --load \
+       -f $(CURDIR)/build-support/docker/Consul-Dev-Multiarch.dockerfile $(CURDIR)/pkg/bin/
+
+check-remote-dev-image-env:
+ifndef REMOTE_DEV_IMAGE
+	$(error REMOTE_DEV_IMAGE is undefined: set this image to <your_docker_repo>/<your_docker_image>:<image_tag>, e.g. hashicorp/consul-k8s-dev:latest)
+endif
+
+remote-docker: check-remote-dev-image-env linux-amd64 linux-arm64
+	@echo "Pulling consul container image - $(CONSUL_IMAGE_VERSION)"
+	@docker pull consul:$(CONSUL_IMAGE_VERSION) >/dev/null
+	@echo "Building and Pushing Consul Development container - $(REMOTE_DEV_IMAGE)"
+	docker buildx create --use && docker buildx build -t '$(REMOTE_DEV_IMAGE)' \
+       --platform linux/amd64,linux/arm64 \
+	   --build-arg CONSUL_IMAGE_VERSION=$(CONSUL_IMAGE_VERSION) \
+       --push \
+       -f $(CURDIR)/build-support/docker/Consul-Dev-Multiarch.dockerfile $(CURDIR)/pkg/bin/
 
 # In CircleCI, the linux binary will be attached from a previous step at bin/. This make target
 # should only run in CI and not locally.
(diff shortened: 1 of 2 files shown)


skpratt (approved):

GNUmakefile
```diff
@@ -152,7 +154,26 @@ dev-docker: linux
...
+
+remote-docker: check-remote-dev-image-env linux-amd64 linux-arm64
+	@echo "Pulling consul container image - $(CONSUL_IMAGE_VERSION)"
+	@docker pull consul:$(CONSUL_IMAGE_VERSION) >/dev/null
+	@echo "Building and Pushing Consul Development container - $(REMOTE_DEV_IMAGE)"
+	@docker buildx create --use && docker buildx build -t '$(REMOTE_DEV_IMAGE)' \
+       --platform linux/amd64,linux/arm64 \
+	   --build-arg CONSUL_IMAGE_VERSION=$(CONSUL_IMAGE_VERSION) \
+       --push \
+       -f $(CURDIR)/build-support/docker/Consul-Dev-Multiarch.dockerfile $(CURDIR)/pkg/bin/
```
dhiaayachi, line 176: I'm wary of having docker push as part of the same make command. I prefer push to be always intentional. What you think?

build-support/docker/Consul-Dev-Multiarch.dockerfile
```diff
@@ -0,0 +1,5 @@
+ARG CONSUL_IMAGE_VERSION=latest
+FROM consul:${CONSUL_IMAGE_VERSION}
+RUN apk update && apk add iptables
+ARG TARGETARCH
+COPY linux_${TARGETARCH}/consul /bin/consul
```
dhiaayachi, line 5: I'm a bit confused about why we need this new dockerfile. Why not use the same dockerfile with a default to the current arch used (amd64 I believe)?

GNUmakefile
```diff
@@ -174,11 +195,19 @@ ifeq ($(CIRCLE_BRANCH), main)
...
+	CGO_ENABLED=0 GOOS=linux GOARCH=$(GOARCH) go build -o ./pkg/bin/linux_$(GOARCH) -ldflags "$(GOLDFLAGS)" -tags "$(GOTAGS)"
+
+linux-amd64:
 	@mkdir -p ./pkg/bin/linux_amd64
 	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o ./pkg/bin/linux_amd64 -ldflags "$(GOLDFLAGS)" -tags "$(GOTAGS)"
 
+linux-arm64:
+	@mkdir -p ./pkg/bin/linux_arm64
+	CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -o ./pkg/bin/linux_arm64 -ldflags "$(GOLDFLAGS)" -tags "$(GOTAGS)"
+
```
dhiaayachi, line 210: why not combine those 3 and have the same build command wit different `GOARCH` values?

dhiaayachi (commented):
I have few questions about the approach, specifically around the need of new dockerfile and the fact that push and build are combined

GNUmakefile
```diff
@@ -152,7 +154,26 @@ dev-docker: linux
 	@docker pull consul:$(CONSUL_IMAGE_VERSION) >/dev/null
 	@echo "Building Consul Development container - $(CONSUL_DEV_IMAGE)"
 	#  'consul:local' tag is needed to run the integration tests
-	@DOCKER_DEFAULT_PLATFORM=linux/amd64 docker build $(NOCACHE) $(QUIET) -t '$(CONSUL_DEV_IMAGE)' -t 'consul:local' --build-arg CONSUL_IMAGE_VERSION=$(CONSUL_IMAGE_VERSION) $(CURDIR)/pkg/bin/linux_amd64 -f $(CURDIR)/build-support/docker/Consul-Dev.dockerfile
+	@docker buildx create --use && docker buildx build -t 'consul:local' \
```
lkysow, line 157: This will actually cause the integration tests to not work because the image cache for buildx images is not the regular system cache and the integration tests don't use this buildx cache right now. You can test this out by running:

```
docker rmi consul-dev
docker rmi consul:local
```

And then running `make make test-envoy-integ`. It should fail with "image not found in registry" errors.

That's probably okay, but we'll need a new target for the integration tests to depend on (right now they depend on dev-docker) that builds using regular docker build.

DanStough: @lkysow I'm doing some experiments on the side w/ skaffold and I believe the following setup might work better for dev environments. It would fix the scoping issue you mentioned and also increase performance with caching 
```suggestion
	@docker buildx use default && docker buildx build -t 'consul:local' \
```

build-support/docker/Consul-Dev-Multiarch.dockerfile
DanStough: I believe I went this route because I didn't know what the default value should be depending on the architecture of the machine. Thinking about it more, maybe we could just change the existing one to always pass `TARGETARCH` when we're not doing a multi-arch build.

GNUmakefile
DanStough: I think I envisioned this workflow only being used for build+push, while we have other targets for making a local copy. I also think it could eliminate some of errors with pushing the wrong image - I've done this a couple of times now where I push an image that's not the right architecture for CI. IMHO it's not easy to tell which images are multi-arch without digging into the `buildx` commands.

Originally this was essential because the new `buildx` context can't load docker images into the local cache, but the change I proposed to Luke below should resolve that concern.

GNUmakefile
DanStough: The honest answer is I wasn't sure how to programmatically exec make targets with different values of `GOARCH`, and it wasn't something I could find at the top of StackOverflow. Maybe you know the secret sauce? :D

DanStough: Figured it out. Will update 🙇

build-support/docker/Consul-Dev-Multiarch.dockerfile
DanStough: Reviewing now, the existing file is used in CI, with the consul binary ending up in a different location to copy over. I think there was a lot less risk/effort in creating the new file than trying to unwind what's going in CI.

dhiaayachi (approved):
LGTM!

Issue #11422, opened 2021-10-26
Add additional Envoy consecutive_5xx outlier ejection configuration


agrahamlincoln:
#### Feature Description

**Background:**
Consul provides upstream configuration [`passive_health_check`](https://www.consul.io/docs/connect/proxies/envoy#passive_health_check) to allow a user to configure Envoy's outlier ejection `interval` and `consecutive_5xx` settings. [[Envoy Outlier Detection Documentation]](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/outlier_detection.proto#envoy-v3-api-field-config-cluster-v3-outlierdetection-consecutive-5xx)

**The Feature:** Add additional configuration parameters to `passive_health_check` for other Envoy outlier ejection settings. Specifically `enforcing_consecutive_5xx` to allow users to disable/slow-down outlier ejection for consecutive_5xx 

#### Use Case(s)

In our use-case, we are connecting nginx (running on VMs) to proxy components in K8S. These "proxy" components are expected to respond to the user with 5xx if that upstream proxy component receives 5xx responses from it's upstream. ejecting the proxy component from our nginx sidecar can cause service outages for all users if one user somehow creates a traffic pattern that results in 5xx responses.

`[VM: nginx --> consul-sidecar-nginx ] ---envoy tunnel---> [K8s: consul-sidecar-proxy --> http-proxy]`

In this use-case, we want to disable outlier ejection so that our proxies are not ejected from the nginx envoy sidecar when it responds to the user with 5xx responses.

more specifically, the proxy we're using is part of a [jupyterhub](https://zero-to-jupyterhub.readthedocs.io/en/latest/) service. Users are able to generate 5xx responses in the upstream proxy by writing "bad" jupyter notebooks that cause 5xx errors to be sent back to the user in response. We want these users to get their 5xx responses, instead of receiving only a few and then causing the proxy to be ejected for all other users of jupyterhub.

lkysow:
Hi Graham, I know this isn't exactly what you're asking for (to configure `enforcing_consecutive_5xx`) but just wondering if you want to simply disable outlier detection whether you can do something like:

```hcl
Kind = "service-defaults"
Name = "nginx"
Protocol = "http"
UpstreamConfig {
  Defaults {
    PassiveHealthCheck {
      MaxFailures = 4294967295 # max 32 bit unsigned int
    }
  }
}
```

PR #14238, opened 2022-08-17
Add additional parameters to envoy passive health check config

malizz:
### Description
This PR exposes an additional parameter (`EnforcingConsecutive5xx` in envoy `passive_health_check`) in Consul proxy configurations.

Resolves: https://github.com/hashicorp/consul/issues/11422

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/consul/config_endpoint_test.go
+++ b/agent/consul/config_endpoint_test.go
@@ -1399,8 +1399,9 @@ func TestConfigEntry_ResolveServiceConfig_Upstreams(t *testing.T) {
 							Protocol:    "http",
 							MeshGateway: structs.MeshGatewayConfig{Mode: structs.MeshGatewayModeRemote},
 							PassiveHealthCheck: &structs.PassiveHealthCheck{
-								Interval:    10,
-								MaxFailures: 2,
+								Interval:                10,
+								MaxFailures:             2,
+								EnforcingConsecutive5xx: 60,
 							},
 						},
 						Overrides: []*structs.UpstreamConfig{
@@ -1432,8 +1433,9 @@ func TestConfigEntry_ResolveServiceConfig_Upstreams(t *testing.T) {
 						Upstream: wildcard,
 						Config: map[string]interface{}{
 							"passive_health_check": map[string]interface{}{
-								"Interval":    int64(10),
-								"MaxFailures": int64(2),
+								"Interval":                int64(10),
+								"MaxFailures":             int64(2),
+								"EnforcingConsecutive5xx": int64(60),
 							},
 							"mesh_gateway": map[string]interface{}{
 								"Mode": "remote",
--- a/agent/consul/config_endpoint_test.go
+++ b/agent/consul/config_endpoint_test.go
@@ -1447,8 +1447,9 @@ func TestConfigEntry_ResolveServiceConfig_Upstreams(t *testing.T) {
 						Upstream: mysql,
 						Config: map[string]interface{}{
 							"passive_health_check": map[string]interface{}{
-								"Interval":    int64(10),
-								"MaxFailures": int64(2),
+								"Interval":                int64(10),
+								"MaxFailures":             int64(2),
+								"EnforcingConsecutive5xx": int64(60),
 							},
 							"mesh_gateway": map[string]interface{}{
 								"Mode": "local",
--- a/agent/consul/config_endpoint_test.go
+++ b/agent/consul/config_endpoint_test.go
@@ -1401,7 +1401,7 @@ func TestConfigEntry_ResolveServiceConfig_Upstreams(t *testing.T) {
 							PassiveHealthCheck: &structs.PassiveHealthCheck{
 								Interval:                10,
 								MaxFailures:             2,
-								EnforcingConsecutive5xx: 60,
+								EnforcingConsecutive5xx: uintPointer(60),
 							},
 						},
 						Overrides: []*structs.UpstreamConfig{
@@ -2509,4 +2509,8 @@ func Test_gateWriteToSecondary_AllowedKinds(t *testing.T) {
 			require.NoError(t, gateWriteToSecondary(tcase.targetDC, tcase.localDC, tcase.primaryDC, tcase.kind))
 		})
 	}
+}
+
+func uintPointer(v uint32) *uint32 {
+	return &v
 }
--- a/agent/consul/config_endpoint_test.go
+++ b/agent/consul/config_endpoint_test.go
@@ -1402,7 +1402,6 @@ func TestConfigEntry_ResolveServiceConfig_Upstreams(t *testing.T) {
 								Interval:                10,
 								MaxFailures:             2,
 								EnforcingConsecutive5xx: 60,
-								MaxEjectionPercent:      10,
 							},
 						},
 						Overrides: []*structs.UpstreamConfig{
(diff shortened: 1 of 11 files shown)


agent/consul/config_endpoint_test.go
```diff
@@ -1399,8 +1399,9 @@ func TestConfigEntry_ResolveServiceConfig_Upstreams(t *testing.T) {
 							Protocol:    "http",
 							MeshGateway: structs.MeshGatewayConfig{Mode: structs.MeshGatewayModeRemote},
 							PassiveHealthCheck: &structs.PassiveHealthCheck{
-								Interval:    10,
-								MaxFailures: 2,
+								Interval:                10,
+								MaxFailures:             2,
+								EnforcingConsecutive5xx: 60,
```
dhiaayachi, line 1404: Maybe add MaxEjectionPercent in here as well

website/content/docs/connect/config-entries/service-defaults.mdx
```diff
@@ -636,6 +648,18 @@ represents a location outside the Consul cluster. They can be dialed directly wh
                   description: `The number of consecutive failures which cause a host to be
                     removed from the load balancer.`,
                 },
+                {
+                  name: 'EnforcingConsecutive5xx',
+                  type: 'int: 0',
```
dhiaayachi, line 653: default values are not quiet right here I think

website/content/docs/connect/config-entries/service-defaults.mdx
```diff
@@ -636,6 +648,18 @@ represents a location outside the Consul cluster. They can be dialed directly wh
...
+                  description: `The % chance that a host will be actually ejected 
+                      when an outlier status is detected through consecutive 5xx.`,
+                },
+                {
+                  name: 'MaxEjectionPercent',
+                  type: 'int: 0',
```
dhiaayachi, line 659: Same about default value

api/config_entry.go
```diff
@@ -196,6 +196,16 @@ type PassiveHealthCheck struct {
 	// MaxFailures is the count of consecutive failures that results in a host
 	// being removed from the pool.
 	MaxFailures uint32 `alias:"max_failures"`
+
+	// EnforcingConsecutive5xx is the % chance that a host will be actually ejected
+	// when an outlier status is detected through consecutive 5xx.
+	// This setting can be used to disable ejection or to ramp it up slowly. Defaults to 100.
```
dhiaayachi, line 202: I'm not sure we are setting the default value, so this will result to not doing any ejection, right? 
We should either change this doc or change the default value.

api/config_entry.go
```diff
@@ -196,6 +196,16 @@ type PassiveHealthCheck struct {
...
+	EnforcingConsecutive5xx uint32 `json:",omitempty" alias:"enforcing_consecutive_5xx"`
+
+	// MaxEjectionPercent is the maximum % of an upstream cluster that can be
+	// ejected due to outlier detection.
+	// Defaults to 10% but will eject at least one host regardless of the value.
```
dhiaayachi, line 207: Same comment here about the default value

agent/xds/config.go
```diff
@@ -168,5 +168,11 @@ func ToOutlierDetection(p *structs.PassiveHealthCheck) *envoy_cluster_v3.Outlier
 	if p.MaxFailures != 0 {
 		od.Consecutive_5Xx = &wrappers.UInt32Value{Value: p.MaxFailures}
 	}
+	if p.EnforcingConsecutive5xx != 0 {
+		od.EnforcingConsecutive_5Xx = &wrappers.UInt32Value{Value: p.EnforcingConsecutive5xx}
+	}
+	if p.MaxEjectionPercent != 0 {
+		od.MaxEjectionPercent = &wrappers.UInt32Value{Value: p.MaxEjectionPercent}
+	}
```
dhiaayachi, line 176: I think we need to add a test for this part, using an xds golden file.

dhiaayachi (changes requested):
Thank you for working on this @malizz! I have few minor comments but the approach is good
I think we are missing few tests:
- at the config-entry level, we are missing testing 1 of the parameters
-  at the xds generation level we are missing a test to make sure that the values are mapped correctly
And default values are not properly set.

website/content/docs/connect/config-entries/service-defaults.mdx
malizz: Should the default values be the values mentioned in envoy documentations?

agent/xds/config.go
```diff
@@ -168,5 +168,12 @@ func ToOutlierDetection(p *structs.PassiveHealthCheck) *envoy_cluster_v3.Outlier
 	if p.MaxFailures != 0 {
 		od.Consecutive_5Xx = &wrappers.UInt32Value{Value: p.MaxFailures}
 	}
+
+	if p.EnforcingConsecutive5xx != nil {
+		od.EnforcingConsecutive_5Xx = &wrappers.UInt32Value{Value: *p.EnforcingConsecutive5xx}
+	} else {
+		od.EnforcingConsecutive_5Xx = &wrappers.UInt32Value{Value: uint32(100)}
```
dhiaayachi, line 175: This mean that we are setting default to 100, right? I thought that we wanted to leave default behaviour same which is not set the parameter.

dhiaayachi (commented):
I added a question about default behaviour, also I think it would be good to add some XDS tests which trigger this.

dhiaayachi (changes requested):
Hey @malizz,
This is getting close to a merge state. I have few comments mostly about doc.
I also think we missed one test spot in [here](https://github.com/hashicorp/consul/blob/87f2f0da6263a28dfc88f8a70a8247c76ad095bb/agent/structs/config_entry_test.go#L2695-L2695)
We should be able to ass a case to test the new value.

malizz:
@hashicorp/consul-docs The default value comes from [envoy docs](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/outlier_detection.proto#envoy-v3-api-field-config-cluster-v3-outlierdetection-consecutive-5xx)

website/content/docs/connect/config-entries/service-defaults.mdx
```diff
@@ -636,6 +642,12 @@ represents a location outside the Consul cluster. They can be dialed directly wh
                   description: `The number of consecutive failures which cause a host to be
                     removed from the load balancer.`,
                 },
+                {
+                  name: 'EnforcingConsecutive5xx',
+                  type: 'int: 0',
+                  description: `The % chance that a host will be actually ejected
```
david-yu, line 648: This should show up both for hcl and yaml in my opinion.

website/content/docs/connect/config-entries/service-defaults.mdx
```diff
@@ -504,6 +504,12 @@ represents a location outside the Consul cluster. They can be dialed directly wh
                   description: `The number of consecutive failures which cause a host to be
                       removed from the load balancer.`,
                 },
+                {
+                  name: 'EnforcingConsecutive5xx',
+                  type: 'int: 100',
+                  description: `The % chance that a host will be actually ejected
```
david-yu, line 510: This should show up for HCL and YAML as well

david-yu:
Hi @malizz thanks for doing this, would you also be open to updating the CRDs for Consul K8s as that is needed in order to use the config entries on Kubernetes. I believe @DanStough recently went through this for a feature as well related to Envoy. See https://github.com/hashicorp/consul-k8s/blob/main/CONTRIBUTING.md#generating-yaml

david-yu (approved):
Approving on the docs side. Thank you!

dhiaayachi (approved):
Thank you for working on this @malizz, LGTM!

PR #14242, opened 2022-08-17
set PRODUCT_VERSION for docker build

mdeggies:
### Description

In `actions-docker-build` we [pass](https://github.com/hashicorp/actions-docker-build/blob/05c370a26e61b06be46c5095d6e914c9f0ea4f3d/scripts/docker_build#L49) `PRODUCT_VERSION` to the docker build command. Since this was not set, the label did not populate properly which is used in a comparison to determine the `minor-latest` and `latest` docker image tags. A change merged in today (https://github.com/hashicorp/actions-docker-build/pull/25) checks for this value being set in the docker build action, and errors out if it does not exist. This is causing docker build failures in this repo, for example: https://github.com/hashicorp/consul/actions/runs/2876677829

How I've tested this PR:
 - build the image up to the point of label creation and pass in `--build-arg PRODUCT_VERSION=1.2.3`
 - inspect the image for the label with the above command

How I expect reviewers to test this PR:
- same as above

### Links

Related [internal-only] post about this: https://hashicorp.atlassian.net/wiki/spaces/RELENG/pages/2416934922/August+17+2022-+Docker+Build+Failures

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -110,13 +110,13 @@ CMD ["agent", "-dev", "-client", "0.0.0.0"]
 # Remember, this image cannot be built locally.
 FROM docker.mirror.hashicorp.services/alpine:3.15 as default
 
-ARG VERSION
+ARG PRODUCT_VERSION
 ARG BIN_NAME
 
 # PRODUCT_NAME and PRODUCT_VERSION are the name of the software on releases.hashicorp.com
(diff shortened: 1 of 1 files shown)


eculver (approved):
Interesting. I'm surprised we hadn't caught this yet. Thanks for fixing!

PR #14266, opened 2022-08-18
Replace FSM’s old registry pattern with newly suggested pattern

Mathew-Estafanous:
### Description
As suggested by the linked issue, I transitioned to the new pattern for the FSM operations and snapshot registries. I wasn’t sure if any of the registries needed enterprise mapping as well, so I went ahead and created a noop function for each registry where enterprise mappings can be added.

### Links
Working on issue #11740

This PR is a continuation of a previously merged PR, https://github.com/hashicorp/consul/pull/12729


Mathew-Estafanous:
@kisunji
Sorry for the ping, but do you have any ideas as to what might be the cause for the failed integration tests? I don’t really understand how my changes are causing the failed tests.

kisunji:
> @kisunji Sorry for the ping, but do you have any ideas as to what might be the cause for the failed integration tests? I don’t really understand how my changes are causing the failed tests.

Hey! If you rebase from `main` I think the tests should stop failing. It was an authorization issue with forked repos so your logic wouldn't have caused anything to fail

Mathew-Estafanous:
Ahh ok, that makes a lot more sense. Thanks!

I also just realized that I opened this PR with my dev branch instead of the proper branch `replace-fsm-registry-pattern`. 😅 Is it ok to continue with a code review with this PR or should I open anther one with the proper branch name?

kisunji:
> Ahh ok, that makes a lot more sense. Thanks!
> 
> I also just realized that I opened this PR with my dev branch instead of the proper branch `replace-fsm-registry-pattern`. 😅 Is it ok to continue with a code review with this PR or should I open anther one with the proper branch name?

That's totally fine. Is this ready for review? It will need some enterprise-side changes on our end but looks simple enough

Mathew-Estafanous:
yup it should be ready for review!

agent/consul/fsm/commands_oss.go
```diff
@@ -107,41 +107,66 @@ var CommandsSummaries = []prometheus.SummaryDefinition{
...
+		entry{structs.PeeringTerminateByIDType, (*FSM).applyPeeringTerminate},
+		entry{structs.PeeringTrustBundleWriteType, (*FSM).applyPeeringTrustBundleWrite},
+		entry{structs.PeeringTrustBundleDeleteType, (*FSM).applyPeeringTrustBundleDelete},
+		entry{structs.PeeringSecretsWriteType, (*FSM).applyPeeringSecretsWrite},
+	)
+	registerEnterpriseCommands(commands, fsm)
+	return commands
+}
+
+func registerEnterpriseCommands(registry map[structs.MessageType]command, fsm *FSM) {}
```
kisunji, line 158: Could you move most of this file to a new file called `commands.go` and leave the empty `registerEnterpriseCommands` in `commands_oss.go` with this directive at the top:
```
//go:build !consulent
```
On the enterprise side we have another file called `commands_ent.go` which will have the enterprise implementation.

agent/consul/fsm/snapshot_oss.go
```diff
@@ -13,98 +13,70 @@ import (
...
-	}
-	if err := s.persistPeeringTrustBundles(sink, encoder); err != nil {
-		return err
-	}
-	if err := s.persistPeeringSecrets(sink, encoder); err != nil {
-		return err
-	}
-	return nil
-}
+func RegisteredRestorer() map[structs.MessageType]restorer {
```
kisunji, line 16: same comment as `commands_oss.go`. Could the common code go to `snapshot.go`, leaving only `registerEnterpriseRestorers` in the oss-version?

Mathew-Estafanous: do you need ``registerEnterprisePersisters`` or is it only for ``Restorers``.

Mathew-Estafanous: Also should I add 
```
//go:build !consulent
```
to the top of ``snapshot_oss.go`` as well or is it just for ``commands_oss.go``?

kisunji: Yeah please add it! We would want `registerEnterprisePersisters` as well.

agent/consul/fsm/snapshot.go
```diff
@@ -104,3 +83,934 @@ func (s *snapshot) Persist(sink raft.SnapshotSink) error {
 func (s *snapshot) Release() {
 	s.state.Close()
 }
+
+func RegisteredRestorer() map[structs.MessageType]restorer {
+	registry := map[structs.MessageType]restorer{
```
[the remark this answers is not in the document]
Mathew-Estafanous: Just pushed those changes.

I also thought that it might be better to make a small(ish) change to ``command.go`` by removing the named struct (`entry`) and just making it an anonymous struct. I thought that it would prevent misuse or confusion.

agent/consul/fsm/snapshot.go
```diff
@@ -104,3 +83,952 @@ func (s *snapshot) Persist(sink raft.SnapshotSink) error {
...
+		s.persistConnectCAConfig,
+		s.persistConfigEntries,
+		s.persistFederationStates,
+		s.persistSystemMetadata,
+		s.persistIndex,
+		s.persistPeerings,
+		s.persistPeeringTrustBundles,
+		s.persistPeeringSecrets,
+	}
+	registerEnterprisePersisters(registry, s)
```
kisunji, line 162: Found another issue while testing this. Unlike maps, slices are passed as copies so `registerEnterprisePersisters` cannot actually add persisters to the `registry`.

Can we change the signature so it simply returns a slice of only the enterprise persisters?

Mathew-Estafanous:
@kisunji any updates on this PR?

Issue #14253, opened 2022-08-18
Consul 1.13 cant disable TLS for gRPC Server when AutoEncrypt is in use (backwards compatibility & envoy usage broken)


pruiz:
#### Overview

After upgrading to consul 1.13 our mesh services broke, due to envoy not being able to fetch configuration using xDS from consul agent. It looks like when AutoEncrypt is enabled (auto_encrypt.tls = true), even if ports.https is not defined, GRPC will be using TLS. This is similar to #14211, but the workaround proposed wont work in this case, as auto-encrypt forces usage of TLS on gRPC server no matter what.

Having gRPC server listen only on TLS breaks consul connect (envoy) support, as consul is unable to connect using TLS, nor present the required certificates with the boostrap configuration  generated by consul. This fails when using 'consul connect envoy' directly, and also when using nomad, especially with bridge networking, as in such cases nomad uses a socket to present a unix socket (proxied by nomad to the gRPC TCP listener from consul).

From what I've seen this issue has it's origin in two complementary changes applied to v1.13:

* https://github.com/hashicorp/consul/pull/12504 -> This PR introduced independent (per-listener) TLS settings, so we could set each interface using it's own TLS-related settings.
* https://github.com/hashicorp/consul/pull/13733 -> This PR removed the dependency between ports.https & enabling of grpc server's TLS, however this PR (IMHO) missed the additional effect of AutoEncrypt, effectively negating a way to use gRPC listener w/o TLS when AutoEncrypt is in use.

Additionally, even setting addresses.grpc to 'unix:///some.sock', so it listens over a unix socket instead of TCP, does not disable TLS on gRPC listener. Not sure this might be considered a bug or a feature.. ;)

PD: This other issue is probably related too: https://github.com/hashicorp/consul/issues/13124

#### Reproduction Steps

Use the following config on consul agents:

```
{
  "auto_encrypt": {
     "tls": true
  },
  "server": false,
  "encrypt": ...,
  "ports": {
     "http": 8500,
     "grpc": 8502
  },
  "tls": ... can be omitted or have any values ...
}
```

### Consul info for both Client and Server

<details>
  <summary>Client info</summary>

```
agent:
	check_monitors = 0
	check_ttls = 0
	checks = 2
	services = 3
build:
	prerelease =
	revision = c6d0f9ec
	version = 1.13.1
	version_metadata =
consul:
	acl = enabled
	known_servers = 3
	server = false
runtime:
	arch = amd64
	cpu_count = 32
[...]

PR #14269, opened 2022-08-18
Added new auto_encrypt.grpc_server_tls config option to control AutoTLS enabling of GRPC Server's TLS usage (for main branch)

pruiz:
### Description
Added new auto_encrypt.grpc_server_tls config option to control AutoTLS enabling of GRPC Server's TLS usage.

See https://github.com/hashicorp/consul/issues/14253

### PR Checklist

* [X] updated test coverage
* [X] not a security concern

### DIFF
--- a/tlsutil/config.go
+++ b/tlsutil/config.go
@@ -198,7 +198,7 @@ type Configurator struct {
 		connectCAPems        []string
 		cert                 *tls.Certificate
 		verifyServerHostname bool
-		grpcIncoming         bool
+		grpcServerUseTLS     bool
 	}
 
 	// logger is not protected by a lock. It must never be changed after
@@ -273,17 +273,12 @@ func (c *Configurator) Update(config Config) error {
 	return nil
 }
 
-func (c *Configurator) SetAutoTLSGRPCIncoming(value bool) error {
+// SetAutoTLSGRPCServer sets autoTLS' grpcServerUseTLS value.
+func (c *Configurator) SetAutoTLSGRPCServer(value bool) error {
 	c.lock.Lock()
 	defer c.lock.Unlock()
-	c.autoTLS.grpcIncoming = value
+	c.autoTLS.grpcServerUseTLS = value
 	return nil
-}
-
-func (c *Configurator) AutoTLSGRPCIncoming() bool {
-	c.lock.RLock()
-	defer c.lock.RUnlock()
-	return c.autoTLS.grpcIncoming
 }
 
 // loadProtocolConfig loads the certificates etc. for a given ProtocolConfig
@@ -642,7 +637,7 @@ func (c *Configurator) Cert() *tls.Certificate {
 func (c *Configurator) GRPCServerUseTLS() bool {
 	c.lock.RLock()
 	defer c.lock.RUnlock()
-	return c.grpc.cert != nil || (c.autoTLS.grpcIncoming && c.autoTLS.cert != nil)
+	return c.grpc.cert != nil || (c.autoTLS.grpcServerUseTLS && c.autoTLS.cert != nil)
 }
 
 // VerifyIncomingRPC returns true if we should verify incoming connnections to
--- a/tlsutil/config.go
+++ b/tlsutil/config.go
@@ -171,6 +171,10 @@ type protocolConfig struct {
 	// combinedCAPool is a pool containing both manualCAPEMs and the certificates
 	// received from auto-config/auto-encrypt.
 	combinedCAPool *x509.CertPool
+
+	// useAutoCert indicates wether we should use auto-encrypt/config data
+	// for TLS server/listener. NOTE: Only applies to external GRPC Server.
+	useAutoCert bool
 }
 
 // Configurator provides tls.Config and net.Dial wrappers to enable TLS for
@@ -202,7 +206,6 @@ type Configurator struct {
 		connectCAPems        []string
 		cert                 *tls.Certificate
 		verifyServerHostname bool
-		grpcServerUseTLS     bool
 	}
 
 	// logger is not protected by a lock. It must never be changed after
@@ -271,7 +274,6 @@ func (c *Configurator) Update(config Config) error {
 	c.grpc = *grpc
 	c.https = *https
 	c.internalRPC = *internalRPC
-	c.autoTLS.grpcServerUseTLS = config.GRPC.UseAutoCert
 
 	atomic.AddUint64(&c.version, 1)
 	c.log("Update")
@@ -329,6 +331,7 @@ func (c *Configurator) loadProtocolConfig(base Config, pc ProtocolConfig) (*prot
 		manualCAPEMs:   pems,
 		manualCAPool:   manualPool,
 		combinedCAPool: combinedPool,
+		useAutoCert:    pc.UseAutoCert,
 	}, nil
 }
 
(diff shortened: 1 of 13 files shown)


mkeeler (changes requested):
First off, great work on all of this. Navigating consuls configuration code and tests for it are not the most straightforward. However, I feel I must apologize as I have inadvertently sent you down a slightly incorrect path. Originally I thought this issue would only affect auto-encrypt and if that were still the case this PR would be perfect and I would merge it immediately. The problem goes a bit deeper as it also affects auto-config (which is really just a super-set of auto-encrypt functionality).

In light of that there are two possible options I see. First would be to add more configuration to the `auto_config` stanza identical to what you have done for `auto_encrypt` or to change course a bit and add the necessary configuration to the `tls.grpc` stanza like so:

```
tls {
   grpc {
      # this should default to false and require users to explicitly enable
      use_auto_cert = true
   }
}
```

Having the same configuration in two places seems like a less desirable configuration UX so I  think the best course of action would be to modify the TLS gRPC settings instead.

If you have the time and inclination to continue with the fix and make those changes we would appreciate your contribution.

pruiz:
Hi @mkeeler 

Just one question regarding 'use_auto_cert' default value. You state that it should default to false and thus require the user to explicitly enable this feature. 

But doesnt that conflict with the original behaviour (by which just enabling auto_encrypt.. enabled it to apply to grpc too?)

I mean, a user with auto_encrypt.tls=true, but w/o tls.grpc.use_auto_cert=true, will end up with grpc in clear text mode. Is that the intended (new) behaviour? (just checking to ensure I understood it right, not that I see any major issues)

Regards

mkeeler:
@pruiz It is a divergence in behavior but I think it leaves the default behavior that most users would desire in a better place.

TLDR: All client agent incoming gRPC is for Envoy and this should usually be over a loopback interface so there is no need for TLS. That plus the fact that correctly managing roots distribution if the auto TLS cert was used mean that its likely that the majority of users would prefer to not use the auto TLS cert and justify having the default of `false`

Using auto TLS certs properly for gRPC would be quite difficult. Once the client agent has the certificate, you would have to manually retrieve the CA roots from Consul's HTTP API and then distribute to all Envoy instances (or wherever `consul connect envoy` is run) throughout your environment. Assuming you get past initial distribution you need to be aware that background intermediate CA certs as well as root certs can be rotated at any time. Intermediate cert rotation shouldn't be an issue but roots rotation is potentially problematic. Consul does cross-sign root certs unless the user overrides that or is using a CA provider that doesn't support cross signing. If cross-signing isn't supported then the first time there is a root rotation all Envoy instances would become no longer able to connect to Consul. Even if cross signing is supported, if a proxy doesn't update its CA root certs after the first rotation and another one happens then it will still get disconnected from Consul (because the third root is only signed by the second and there isn't a link back to the original roots).

pruiz:
I've just pushed a new commit refactoring the new setting as tls.grpc.use_auto_cert. I will be doing a bit more of testing later today.

Regards
Pablo

tlsutil/config.go
```diff
@@ -266,6 +271,7 @@ func (c *Configurator) Update(config Config) error {
 	c.grpc = *grpc
 	c.https = *https
 	c.internalRPC = *internalRPC
+	c.autoTLS.grpcServerUseTLS = config.GRPC.UseAutoCert
```
mkeeler, line 274: Do we need the `grpcServerUseTLS` field any longer? Couldn't the `GRPCServerUseTLS` method just check `c.grpc.UseAutoCert`?

pruiz: I will need to add a new property to protocolConfig in this case, I think. But I can do that, no prob.

pruiz: Pushed a new commit removing Configurator's autoTLS.grpcServerUseTLS, using a protocolConfig.useAutoCert instead.

pruiz:
@mkeeler I have an issue with this current constraint (tls.grpc was provided but TLS will NOT be enabled on the gRPC listener without an HTTPS listener configured (e.g. via ports.https)), by which I am supposed not to define tls.grpc when ports.https is disabled.

I am trying to think the best way around this.. On one way I could remove this limitation, so tls.grpc does not depend on ports.https.. but I have the feeling that this will break another backwards compatibility stuff somewhere else.

May you have some guidance here I can benefit from? ;)

mkeeler:
@pruiz for 1.13 we intended to break the dependency on gRPC TLS with having an https listener.

https://www.consul.io/docs/upgrading/upgrade-specific#changes-to-grpc-and-http-interface-configuration

This commit removed the HTTPs port check when the Agent is creating the gRPC server instance:
https://github.com/hashicorp/consul/commit/bb4d4040fb0c7fd6df56ebd0798876289927adb7

It would appear that we forgot to remove the warning from the config builder code so if you could just remove the check all together that would be good as the warning is no inaccurate anyways.

pruiz:
@mkeeler I've removed the aforementioned check, along with an existing unit-test related to such check. Now all tests pass.. if the changes look good, I can backport them to the branch intended for 1.13.x.

Regards
Pablo

mkeeler (approved):
Excellent work! Everything looks good to me. Before spending tons of effort backporting we can wait and see if our backport assistant bot can handle it. I will add the appropriate backport labels and the bot should chime in shortly after merging if it was unable to backport the changes.

I am checking with some folks to determine if there is a way around the CI failures for not having AWS credentials. I thought the test suite used to gracefully just not run the AWS specific bits requiring credentials for PRs from forked repos but apparently something has changed.

I would love to see those checks pass but if not, I may merge anyways and let those CI checks run on the main branch (and fix them if necessary)

mkeeler:
@pruiz I believe if you rebase your branch against main it will pull in some fixes that landed Friday to work out the CI credentials issues. It was in #14257 and the corresponding commits do not appear to be in your branch.

pruiz:
Rebased, but now it looks like I have a new unrelated test failing (TestConfigEntry_ResolveServiceConfig_Upstreams_Blocking/blocking_query_for_foo_wakes_on_bar_entry_delete)... I wait for tomorrow to see if a fix for this lands in master, so I can rebase again.

website/content/docs/agent/config/config-files.mdx
```diff
@@ -2019,6 +2019,9 @@ specially crafted certificate signed by the CA can be used to gain full access t
 
     - `verify_incoming` - ((#tls_grpc_verify_incoming)) Overrides [`tls.defaults.verify_incoming`](#tls_defaults_verify_incoming).
 
+    - `use_auto_cert` - (Defaults to `false`) When true, auto_encrypt TLS settings will be applied to gRPC listener.
+      This is mainly intended for disabling TLS on gRPC server while using auto_encrypt for other TLS usages (like HTTPS).
```
trujillo-adam, line 2023:
```suggestion
    - `use_auto_cert` - (Defaults to `false`) Enables or disables TLS on gRPC servers. Set to `true` to allow `auto_encrypt` TLS settings to apply to gRPC listeners. We recommend disabling TLS on gRPC servers if you are using `auto_encrypt` for other TLS purposes, such as enabling HTTPS.
```
A few grammar fixes and switching to active voice per the style guide.

trujillo-adam (commented):
Just a suggestion for the documentation.

trujillo-adam (approved):
LGTM

mkeeler:
@pruiz FWIW I pulled down your branch, did a `git rebase origin/main` and `go test ./agent/consul -run TestConfigEntry_ResolveServiceConfig_Upstreams_Blocking` and it passed.

I suspect this is nothing more than a flakey test (this one unfortunately happens to have a few timing constraints in it that could cause spurious failures). I would go ahead and push up what you have and it will probably pass the CI checks.

pruiz:
@mkeeler perfect, just let me know if you need something from me.

mkeeler:
@pruiz Thanks for all your work on this. It is greatly appreciated.

PR #14281, opened 2022-08-19
doc: tproxy destination fixes

DanStough:
This PR makes some small tweaks based on internal feedback on the new docs for Transparent Proxy Destinations.

### PR Checklist

* [ ] ~updated test coverage~
* [X] not a security concern

### DIFF
--- a/website/content/docs/k8s/connect/terminating-gateways.mdx
+++ b/website/content/docs/k8s/connect/terminating-gateways.mdx
@@ -89,13 +89,13 @@ Registering the external services with Consul is a multi-step process:
 ### Register external services with Consul
 
 There are two ways to register an external service with Consul:
-1. If [`TransparentProxy`](/docs/k8s/helm#v-connectinject-transparentproxy) is enabled, you can declare external endpoints in the [`Destination`](/docs/connect/config-entries/service-defaults#terminating-gateway-destination) field of `service-defaults`.
+1. If [`TransparentProxy`](/docs/connect/transparent-proxy) is enabled, the preferred method is to declare external endpoints in the [`destination`](/docs/connect/config-entries/service-defaults#terminating-gateway-destination) field of `ServiceDefaults`.
 1. You can add the service as a node in the Consul catalog.
 
-#### Register an external service as a Destination
+#### Register an external service as a destination
 
-`Destination` fields allow clients to dial the external service directly and are valid only in [`TransparentProxy`](/docs/k8s/helm#v-connectinject-transparentproxy) mode.
-The following table describes traffic behaviors when using `Destination`s to route traffic through a terminating gateway:
+The [`destination`](/docs/connect/config-entries/service-defaults#terminating-gateway-destination) field of the `ServiceDefaults` Custom Resource Definition (CRD) allows clients to dial the external service directly. It is valid only in [`TransparentProxy`](/docs/connect/transparent-proxy)) mode.
+The following table describes traffic behaviors when using `destination`s to route traffic through a terminating gateway:
 
 | External Services Layer | Client dials | Client uses TLS | Allowed | Notes |
 |---|---|---|---|---|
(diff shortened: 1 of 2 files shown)


website/content/docs/k8s/connect/terminating-gateways.mdx
```diff
@@ -261,19 +263,20 @@ spec:
 
 </CodeBlockConfig>
 
--> **NOTE**: If TLS is enabled for external services registered through the Consul catalog, you must include the `caFile` parameter that points to the system trust store of the terminating gateway container.
+-> **NOTE**: If TLS is enabled for external services registered through the Consul catalog and you are not using [transparent proxy `Destinations`](#register-an-external-service-as-a-destination), you must include the `caFile` parameter that points to the system trust store of the terminating gateway container.
 By default, the trust store is located in the `/etc/ssl/certs/ca-certificates.crt` directory.
 Configure the `caFile` parameter to point to the `/etc/ssl/cert.pem` directory if TLS is enabled and you are using one of the following components:
  * Consul Helm chart 0.43 or older
  * Or an Envoy image with an alpine base image
+For Transparent Proxy Destinations, see the [guidance in the table above](#register-an-external-service-as-a-destination).
```
lkysow, line 271: CleanShot 2022-08-22 at 12 03 57

The rendering is messed up here. Part of the note has text it shouldn't and the last bullet has text it shouldn't.

website/content/docs/k8s/connect/terminating-gateways.mdx
```diff
@@ -261,19 +263,20 @@ spec:
 
 </CodeBlockConfig>
 
--> **NOTE**: If TLS is enabled for external services registered through the Consul catalog, you must include the `caFile` parameter that points to the system trust store of the terminating gateway container.
+-> **NOTE**: If TLS is enabled for external services registered through the Consul catalog and you are not using [transparent proxy `Destinations`](#register-an-external-service-as-a-destination), you must include the `caFile` parameter that points to the system trust store of the terminating gateway container.
```
lkysow, line 266: What `caFile` parameter is this talking about? I know that this was here before your change but maybe we can link to the param? It's confusing reading this what it's talking about.

website/content/docs/k8s/connect/terminating-gateways.mdx
```diff
@@ -291,6 +294,7 @@ spec:
 ```
 
 </CodeBlockConfig>
+-> **NOTE**: L7 Intentions are not currently supported for transparent proxy Destinations.
```
lkysow, line 297: What are L7 Intentions? There's not a good page to link to unfortunately but we could at least link to https://consul-17na2ubrq-hashicorp.vercel.app/docs/connect/config-entries/service-intentions#permissions

lkysow (commented):
I know these seem nitpicky but I'm worried that right now users won't figure out how to actually use this amazing feature.

* [Here](https://github.com/hashicorp/consul/pull/14281/files#diff-410d7d4b9c3c94aa3b4a160a0032b52d42b34f1237fc16e4615942e014ffe7e0R92) can we say that this is the recommended way if transparent proxy is enabled (the default)? It's not clear for the user which way to choose/which way is best. 
* We use `Destination` everywhere but the actual field in k8s is lowercase: `destination`.
* When saying "if tproxy is enabled" we should link to the tproxy docs page: https://consul-17na2ubrq-hashicorp.vercel.app/docs/connect/transparent-proxy not the config setting because the config setting doesn't tell you enough.
* We shouldn't use `service-defaults` for the k8s docs, the config entry kind is `ServiceDefaults` in k8s.
* Is there a feature name for this? Is it "Destinations", is it "service-defaults destinations", is it "transparent proxy Destinations"? We should stick to one.
*  CleanShot 2022-08-22 at 11 58 37 isn't rendering properly.

DanStough:
@lkysow, I appreciate the thorough review and agree on all points. Let's see if the proposed revisions get us closer to the desired consistency.

website/content/docs/k8s/connect/terminating-gateways.mdx
```diff
@@ -133,10 +135,10 @@ Create a `service-defaults` custom resource for the external service:
 Apply the `ServiceDefaults` resource with `kubectl apply`:
 
 ```shell-session
-$ kubectl apply --filename service-defaults.yaml
+$ kubectl apply --filename serviceDefaults.yaml
```
lkysow, line 138: not a problem but the filename is fine to be `-` separated and is pretty normal for k8s yaml files

lkysow (approved):

PR #14283, opened 2022-08-20
ci: Add PR number and title to oss-merge payload

eculver:
### Description
We want to use the titles from PRs in OSS when we create their accompanying Enterprise merges. With the title and the number in the trigger payload, we can consume it on the Enterprise side when we create the PR. This updates the trigger action to provide those values.

The biggest hurdle in testing here was dealing with the potential for there to be quotes or other special characters in the PR title that wouldn't play nicely when provided in a `POST` request body to `curl`. I ended up using [the same method that we use for triggering a load test](https://github.com/hashicorp/consul/blob/ca22ac9b298f3b66b57c751e38041335a8c0fa1e/.github/workflows/load-test.yml#L21-L30). This uses `jq --arg` functionality to escape the values in the payload. One thing to note with this is that I had to remove hyphens from the arg names for it to work.

### DIFF
--- a/.github/workflows/oss-merge-trigger.yml
+++ b/.github/workflows/oss-merge-trigger.yml
@@ -15,12 +15,20 @@ jobs:
     steps:
       - name: Trigger Merge
         env:
-          GIT_REF: ${{ github.ref_name }}
-          GIT_SHA: ${{ github.sha }}
           GH_PAT: ${{ secrets.ELEVATED_GITHUB_TOKEN }}
-          GIT_ACTOR: ${{ github.actor }}
         run: |
+          # Build JSON payload to trigger oss-merge GitHub Action in Enterprise
+          # the primary reason for doing this is that jq should handle escaping
+          # JSON strings for us. This only runs the load test pipeline on the 'main' branch.
+          jsonData=$(jq --null-input -r \
+            --arg gitref ${{ github.ref_name }} \
+            --arg gitsha ${{ github.sha }} \
+            --arg gitactor ${{ github.actor }} \
+            --arg prnumber ${{ github.event.pull_request.number }} \
+            --arg prtitle ${{ github.event.pull_request.title }} \
+            '{ "event_type": "oss-merge", "client_payload": { "git-ref": $gitref, "git-sha": $gitsha, "git-actor": $gitactor, "pr-number": $prnumber, "pr-title": $prtitle } }' )
+          echo "Passing JSON payload to GitHub API: $jsonData"
           curl -H "Authorization: token $GH_PAT" \
             -H 'Accept: application/json' \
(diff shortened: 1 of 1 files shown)


kisunji:
I'm not sure if this would work in the case where this triggers for PR OSS_0, OSS_1 but on the enterprise side only OSS_1 gets merged with the changes of `{ OSS_0, OSS_1 }`. 

One thing I found while trying solutions for myself was that the real commits _do_ get merged over but are accompanied by 2 merge commits: 1 into the oss-merge branch and then another into `main` when the PR is approved. So the screenshots of the 8 consecutive merge commits are actually empty merges (e.g. from someone merging OSS_0 when OSS_1 already had been merged)

eculver:
> I'm not sure if this would work in the case where this triggers for PR `OSS_0`, `OSS_1` but on the enterprise side only `OSS_1` gets merged with the changes of `{ OSS_0, OSS_1 }`.

This is a good point. The way I have been treating the resulting PRs is that, since there _should_ be a 1:1 now for each OSS PR that gets merged, each PR should get merged despite there being a backlog of other commits included in one when it gets created. Even if we unblock the head of the line and merge multiple changes in one, we can still have a reference for each OSS change (by virtue of its title and OSS SHA) even if it become a no-ops. 

Your point is valid though. Unless we have a way to apply the changes out-of-order these PR titles may be misleading when they contain more than one change.

eculver:
> I'm not sure if this would work in the case where this triggers for PR `OSS_0`, `OSS_1` but on the enterprise side only `OSS_1` gets merged with the changes of `{ OSS_0, OSS_1 }`.

One more clarification about this... I think it should technically still "work" as the title is not pulled in on the Enterprise side from the `git log` but rather it's provided by the OSS PR closed event. I think @kisunji points out that the titles will be misleading in the case he mentions because `OSS_1` will be titled for a single PR, but contain the changes for `{ OSS_0, OSS_1 }`.

One way we could handle this wrinkle is by somehow indicating a PR is a "rollup" of many OSS PRs, but that will require some much more work on the Enterprise side to handle and I don't think it would be pretty. 

Another we we could handle this is to ensure that each Enterprise PR only contains a single OSS PR either by "social" contract (aka manually merging them in order) or by complex software we write on the Enterprise side (arm wavy something something maybe it's possible?)

kisunji:
> This is a good point. The way I have been treating the resulting PRs is that, since there should be a 1:1 now for each OSS PR that gets merged, each PR should get merged despite there being a backlog of other commits included in one when it gets created. Even if we unblock the head of the line and merge multiple changes in one, we can still have a reference for each OSS change (by virtue of its title and OSS SHA) even if it becomes a no-op.

Not sure if I agree here or if I'm misunderstanding;

We need 1:1 commits to exist, but if one mega-backlogged PR gets merged containing 100 commits from OSS, technically all the other previous PRs don't need to be merged (and will be empty merge commits which I'd want to avoid)

dhiaayachi (approved):

PR #14285, opened 2022-08-22
peering: Subscribe to server address changes and push updates to peers

kisunji:
### Description
Server addresses are currently bundled into the original peering token but those can become stale as server hosts/pods are rolled. 

Autopilot has a topic for server address changes we can subscribe to then send updates over the replication stream.

Best viewed by commit.

* Added unit tests streaming address updates
* Updated several broken tests to handle new replication topic

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/consul/autopilotevents/ready_servers_events.go
```diff
@@ -276,6 +281,38 @@ func (r *ReadyServersEventPublisher) getTaggedAddresses(srv *autopilot.ServerSta
...
+		nil,
+		n.Node,
+		structs.ConsulServiceID,
+		structs.NodeEnterpriseMetaInDefaultPartition(),
+		structs.DefaultPeerKeyword,
+	)
+	if err != nil || ns == nil || ns.Meta == nil {
+		return 0
+	}
+	if str, ok := ns.Meta["grpc_port"]; ok {
```
kisunji, line 306: Reference: https://github.com/hashicorp/consul/commit/ea58f235f5da416224ba615405269661ba1f4d8d

agent/grpc-external/services/peerstream/stream_test.go
```diff
@@ -1326,7 +1330,7 @@ func makeClient(t *testing.T, srv pbpeerstream.PeerStreamServiceServer, peerID s
 		// Pass errors from server handler into ErrCh so that they can be seen by the client on Recv().
 		// This matches gRPC's behavior when an error is returned by a server.
 		if err := srv.StreamResources(client.ReplicationStream); err != nil {
-			errCh <- srv.StreamResources(client.ReplicationStream)
+			errCh <- err
```
kisunji, line 1333: this was a bug

agent/consul/autopilotevents/ready_servers_events_test.go
```diff
@@ -184,12 +185,24 @@ func TestAutopilotStateToReadyServersWithTaggedAddresses(t *testing.T) {
...
-	).Once().Return(
+	).Times(2).Return(
 		uint64(0),
 		&structs.Node{TaggedAddresses: map[string]string{"wan": "9.8.7.6"}},
 		nil,
 	)
 
+	store.On("NodeService",
+		memdb.WatchSet(nil),
+		"",
```
mkeeler, line 196: Shouldn't this be a non-empty node name?

mkeeler (commented):
The code seems good. There looks to be some lingering things in the ready servers event tests. 

When all that is sorted just re-request my review.

agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -544,14 +557,11 @@ func (s *Server) realHandleStream(streamReq HandleStreamRequest) error {
...
-				case req.ResponseNonce == "" && req.Error != nil:
-					return grpcstatus.Error(codes.InvalidArgument, "initial subscription request for a resource type must not contain an error")
-
-				case req.ResponseNonce != "" && req.Error == nil: // ACK
+				case req.Error == nil: // ACK
 					// TODO(peering): handle ACK fully
 					status.TrackAck()
 
-				case req.ResponseNonce != "" && req.Error != nil: // NACK
+				case req.Error != nil: // NACK
```
kisunji, line 564: Calling out the removal of this check. Since the initial stream tracking happens in the block above, I don't think checking the ResponseNonce is necessary (correct me if I'm wrong).

The first case in particular was causing errors in tests where the initial replication request for server addresses failed (peering was deleted). The server was stuck in a loop sending replication messages.

agent/consul/autopilotevents/ready_servers_events_test.go
kisunji: The node names all seem to be empty (has NodeIDs instead)

agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -161,8 +161,20 @@ func (s *Server) StreamResources(stream pbpeerstream.PeerStreamService_StreamRes
 	if p == nil {
 		return grpcstatus.Error(codes.InvalidArgument, "initial subscription for unknown PeerID: "+req.PeerID)
 	}
+	if !p.IsActive() {
```
kisunji, line 164: Addressed this TODO

agent/consul/autopilotevents/ready_servers_events.go
kisunji, line 306: Reference: https://github.com/hashicorp/consul/commit/ea58f235f5da416224ba615405269661ba1f4d8d

agent/grpc-external/services/peerstream/subscription_manager_test.go
```diff
@@ -634,12 +634,23 @@ func TestSubscriptionManager_ServerAddrs(t *testing.T) {
...
+	payload := autopilotevents.EventPayloadReadyServers{
+		autopilotevents.ReadyServerInfo{
+			ID:          "9aeb73f6-e83e-43c1-bdc9-ca5e43efe3e4",
+			Address:     "198.18.0.1",
+			Version:     "1.13.1",
+			ExtGRPCPort: 8502,
+		},
+	}
+	// mock handler only gets called once during the initial subscription
+	backend.handler.expect("", 0, 1, payload)
```
kisunji, line 652: Moved the expect call to before the subscription otherwise a data race could occur

agent/consul/autopilotevents/ready_servers_events_test.go
mkeeler: Ah, its empty because the mocked Nodes a few lines up don't set a `Name`

mkeeler: Maybe it would be best to have those return a name and then expect the NodeService calls to pass those names.

kisunji: Yep, I can make that change

agent/grpc-external/services/peerstream/subscription_manager.go
```diff
@@ -721,3 +737,109 @@ const syntheticProxyNameSuffix = "-sidecar-proxy"
...
+
+		var serverAddrs = make([]string, 0, len(payload))
+
+		for _, srv := range payload {
+			if srv.ExtGRPCPort == 0 {
+				continue
+			}
+			grpcAddr := srv.Address + ":" + strconv.Itoa(srv.ExtGRPCPort)
+			serverAddrs = append(serverAddrs, grpcAddr)
+		}
```
kisunji, line 834: Should we do a length check on `serverAddrs` here and prevent sending if len == 0? If we catch it on the server, client will continue to dial the old addresses until they fail to listen.

mkeeler: Ignore events with 0 servers seems like a good idea to me

agent/grpc-external/services/peerstream/stream_resources.go
mkeeler: This seems fine to me.

mkeeler (commented):
I think that 0 length check you proposed is a good idea. Otherwise I have no concerns to merging this.

kisunji:
PR is failing due to changes from https://github.com/hashicorp/consul/pull/14004.
Will try to address with https://github.com/hashicorp/consul/pull/14371

agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -161,8 +161,20 @@ func (s *Server) StreamResources(stream pbpeerstream.PeerStreamService_StreamRes
...
+			},
+		}
+		logTraceSend(logger, term)
 
-	// TODO(peering): If the peering is marked as deleted, send a Terminated message and return
+		err := stream.Send(term)
+		if err != nil {
+			return grpcstatus.Error(codes.FailedPrecondition, "peering is marked as deleted: "+req.PeerID)
+		}
+	}
```
kisunji, line 177: note: add return in non-err case

agent/grpc-external/services/peerstream/stream_test.go
```diff
@@ -1343,11 +1347,19 @@ func makeClient(t *testing.T, srv pbpeerstream.PeerStreamServiceServer, peerID s
...
+	receivedSub3, err := client.Recv()
+	require.NoError(t, err)
+
+	// This is required when the client subscribes to server address replication messages.
+	// We assert for the handler to be called at least once but the data doesn't matter.
+	srv.mockSnapshotHandler.expect("", 0, 0, nil)
 
-	// Issue a services and roots subscription pair to server
+	// Issue services, roots, and server address subscription to server.
+	// Note that server address may not come as an initial message
```
kisunji, line 1358: This message is misleading; if we mock data in the `expect` above we should see an initial server addresses

kisunji: todo

agent/consul/autopilotevents/ready_servers_events_test.go
```diff
@@ -139,6 +140,7 @@ func TestAutopilotStateToReadyServers(t *testing.T) {
 func TestAutopilotStateToReadyServersWithTaggedAddresses(t *testing.T) {
```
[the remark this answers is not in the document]
kisunji: Yeah that makes sense

agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -161,8 +161,19 @@ func (s *Server) StreamResources(stream pbpeerstream.PeerStreamService_StreamRes
...
+			Payload: &pbpeerstream.ReplicationMessage_Terminated_{
+				Terminated: &pbpeerstream.ReplicationMessage_Terminated{},
+			},
+		}
+		logTraceSend(logger, term)
 
-	// TODO(peering): If the peering is marked as deleted, send a Terminated message and return
+		// we don't care if send fails; stream will be killed by termination message or grpc error
+		_ = stream.Send(term)
+		return grpcstatus.Error(codes.FailedPrecondition, "peering is marked as deleted: "+req.PeerID)
```
freddygv, line 175: Hmm I just noticed that this is another use-case for `codes.FailedPrecondition`. This code is one that can trigger a tight loop by the dialer on the assumption that the precondition is that we dialed a follower. But in this case that's not true.

I'll write up an issue about that.

freddygv (commented):
🙌🏽 

I don't have any major concerns but left some questions/comments

agent/grpc-external/services/peerstream/subscription_manager.go
```diff
@@ -721,3 +737,112 @@ const syntheticProxyNameSuffix = "-sidecar-proxy"
...
+func (m *subscriptionManager) subscribeServerAddrs(
+	ctx context.Context,
+	idx uint64,
+	updateCh chan<- cache.UpdateEvent,
+) (uint64, error) {
+	// following code adapted from serverdiscovery/watch_servers.go
+	sub, err := m.backend.Subscribe(&stream.SubscribeRequest{
+		Topic:   autopilotevents.EventTopicReadyServers,
+		Subject: stream.SubjectNone,
+		Token:   "", // using anonymous token for now
```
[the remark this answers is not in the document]
kisunji: My understanding is that subscriptions are cancelled if any ACL policy changes impact this `Token`. It isn't actually used for ACL enforcement

agent/grpc-external/services/peerstream/subscription_manager.go
```diff
@@ -721,3 +737,112 @@ const syntheticProxyNameSuffix = "-sidecar-proxy"
...
+) {
+	// Wait until this is subscribed-to.
+	select {
+	case <-m.serverAddrsSubReady:
+	case <-ctx.Done():
+		return
+	}
+
+	var idx uint64
+	// TODO(peering): retry logic; fail past a threshold
```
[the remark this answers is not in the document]
kisunji: Yeah this is a copy paste from other loops but my original intention was to prevent tight loops from spamming the logs

agent/consul/autopilotevents/ready_servers_events.go
```diff
@@ -276,6 +281,38 @@ func (r *ReadyServersEventPublisher) getTaggedAddresses(srv *autopilot.ServerSta
...
 }
 
+// getGRPCPort will get the external gRPC port for a Consul server.
+// Returns 0 if there is none assigned or if an error is encountered.
+func (r *ReadyServersEventPublisher) getGRPCPort(srv *autopilot.ServerState) int {
+	if r.GetStore == nil {
+		return 0
+	}
+
+	_, n, err := r.GetStore().GetNodeID(types.NodeID(srv.Server.ID), structs.NodeEnterpriseMetaInDefaultPartition(), structs.DefaultPeerKeyword)
```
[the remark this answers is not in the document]
kisunji: Yeah it really should be `GetNodeByID`

agent/grpc-external/services/peerstream/stream_resources.go
kisunji: Yeah I think we need a pass through our codes and try to make them as standardized as possible. I could make this another code to prevent that tight loop

agent/grpc-external/services/peerstream/subscription_manager.go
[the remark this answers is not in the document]
freddygv: I see. 

Not for this PR but I wonder if we could use the root/bootstrap token here to avoid that needless churn. As a server we can just read it, and users can't delete it.

freddygv (approved):

PR #14289, opened 2022-08-22
Fix: add missing ent meta for test

huikang:
### Description
- This is to add the ent meta so that it can pass the ent test
- This fix the test issue caused by #14072  in ent

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


eculver (approved):
Validated already offline with @huikang. LGTM!

Issue #12684, opened 2022-04-01
Deleting a proxy-defaults config entry by any name will delete the global entry


dekimsey:
#### Overview of the Issue

An operator unwittingly deleted the global proxy-defaults with the following command:
```
consul config delete -kind proxy-defaults -name foobar
```

While there is only one valid name for proxy-defaults, "global", deletion of the record should be explicit. Manipulation should be explicit and/or produce an error.
 
#### Reproduction Steps

Steps to reproduce this issue, eg:

1. Define a connect config entry with  kind `proxy-defaults`, name `global`.
2. Delete a `proxy-defaults` entry by any name.

```
$ consul config read -name global -kind proxy-defaults
Error reading config entry proxy-defaults/global: Unexpected response code: 404 (Config entry not found for "proxy-defaults" / "global")
$ consul config read -name foobar -kind proxy-defaults
Error reading config entry proxy-defaults/foobar: Unexpected response code: 404 (Config entry not found for "proxy-defaults" / "foobar")
$ cat proxy-defaults.hcl
Kind      = "proxy-defaults"
Name      = "global"
MeshGateway {
  Mode = "local"
}
$ consul config write proxy-defaults.hcl
Config entry written: proxy-defaults/global
$ consul config read -name foobar -kind proxy-defaults
Error reading config entry proxy-defaults/foobar: Unexpected response code: 404 (Config entry not found for "proxy-defaults" / "foobar")
$ consul config read -name global -kind proxy-defaults
{
    "Kind": "proxy-defaults",
    "Name": "global",
    "TransparentProxy": {},
    "MeshGateway": {
        "Mode": "local"
    },
    "Expose": {},
    "CreateIndex": 16268829,
    "ModifyIndex": 16268829
}
$ consul config delete -name foobar -kind proxy-defaults
Config entry deleted: proxy-defaults/foobar
$ consul config read -name global -kind proxy-defaults
Error reading config entry proxy-defaults/global: Unexpected response code: 404 (Config entry not found for "proxy-defaults" / "global")
```

### Consul info for both Client and Server

<details>
  <summary>Client info</summary>

```
Same as server
```

</details>

<details>
  <summary>Server info</summary>

```
agent:
	check_monitors = 0
	check_ttls = 0
	checks = 0
	services = 0
build:
	prerelease =
	revision = 2c56447e
	version = 1.11.1
consul:
	acl = enabled
	bootstrap = false
	known_datacenters = 2
	leader = true
[...]

PR #14290, opened 2022-08-22
validate args before deleting proxy defaults

malizz:
### Description
This PR fixes a bug where proxy defaults could be deleted by any name. Previously the code did not verify the name before proxy config deletion.

Fixes https://github.com/hashicorp/consul/issues/12684

* Replications steps are described [here](https://github.com/hashicorp/consul/issues/12684)

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/structs/config_entry.go
+++ b/agent/structs/config_entry.go
@@ -364,6 +364,9 @@ func (e *ProxyConfigEntry) Normalize() error {
 
 	e.Kind = ProxyDefaults
 
+	// proxy default config only accepts global configs
+	// this check is replicated in normalize() and validate(),
+	// since validate is not called by all the endpoints (e.g., delete)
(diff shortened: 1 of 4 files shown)


agent/structs/config_entry.go
```diff
@@ -362,6 +363,10 @@ func (e *ProxyConfigEntry) Normalize() error {
 	}
 
 	e.Kind = ProxyDefaults
+
+	if e.Name != "" && e.Name != ProxyConfigGlobal {
+		return fmt.Errorf("invalid name (%q), only %q is supported", e.Name, ProxyConfigGlobal)
+	}
```
DanStough, line 369: I would add a comment here to describe why this needed to be here instead of `Validate()`.

DanStough (approved):
LGTM 🥳 🍾 ! I would just add a note since this wasn't intuitive.

PR #14294, opened 2022-08-22
Add separate grpc tls port.

hashi-derek:
### Description
The mixture of TLS and plain-text communication over the singular gRPC port has been a source of bugs and complication to users in the past. This PR attempts to split the port into two distinct plain-text and TLS configuration ports. This should result in simpler logic for users to understand whether traffic is encrypted or not, while also reducing long-term complexity in the codebase.

For backwards compatibility, the original gRPC port is still allowed to create TLS listeners, but this will be removed in a future release. The following logic is now performed:
      `ports.grpc > 0 && ports.grpc_tls > 0` spawns both plain-text and tls ports.
      `ports.grpc > 0 && tls.grpc == undefined` spawns a single plain-text port.
      `ports.grpc > 0 && ports.grpc_tls <= 0 && tls.grpc == defined` spawns a single tls port (backwards compat mode).

Manual testing was performed by:

```bash
consul agent -dev  # With tls.grpc = undefined, ports.grpc = 8502, ports.grpc_tls = -1
    consul connect envoy  # Works with no grpc-addr specified.
    lsof -M | grep "consul"  # Shows proper port listeners
    curl -vvv http://127.0.0.1:8502/  # Recieves HTTP/0.9 (plain-text gRPC handshake).
    curl http://127.0.0.1:8500/v1/agent/self  # XDS.Port == 8502 && XDS.Ports.TLS == -1

consul agent -dev  # With tls.grpc = defined, ports.grpc = 8502, ports.grpc_tls = -1;
    consul connect envoy  # FAILS with no grpc-addr specified. Works with grpc-addr specified.
    lsof -M | grep "consul"  # Shows proper port listeners
    openssl s_client -connect 127.0.0.1:8503 | openssl x509 -noout -text  # Uses correct certificate.
    curl --cacert tls/consul-agent-ca.pem -vvv https://127.0.0.1:8503/  # TLS handshake succeeded.
    curl http://127.0.0.1:8500/v1/agent/self  # XDS.Port == 8502 && XDS.Ports.TLS == -1

consul agent -dev  # With tls.grpc = defined, ports.grpc = -1, ports.grpc_tls = 8503
    consul connect envoy  # Works with no grpc-addr specified.
    lsof -M | grep "consul"   # Shows proper port listeners
    openssl s_client -connect 127.0.0.1:8503 | openssl x509 -noout -text   # Uses correct certificate.
    curl --cacert tls/consul-agent-ca.pem -vvv https://127.0.0.1:8503/  # TLS handshake succeeded.
    curl http://127.0.0.1:8500/v1/agent/self  # XDS.Port == XDS.Ports.TLS == 8503

consul agent -dev  # With tls.grpc = defined, ports.grpc = 8502, ports.grpc_tls = 8503
    consul connect envoy  # Works with no grpc-addr specified.
    lsof -M | grep "consul"  # Shows proper port listeners
    openssl s_client -connect 127.0.0.1:8502 | openssl x509 -noout -text  # Fails to establish TLS.
    openssl s_client -connect 127.0.0.1:8503 | openssl x509 -noout -text  # Uses correct certificate.
    curl --cacert tls/consul-agent-ca.pem -vvv http://127.0.0.1:8502/  # Recieves HTTP/0.9 (plain-text gRPC handshake).
    curl --cacert tls/consul-agent-ca.pem -vvv https://127.0.0.1:8503/  # TLS handshake succeeded.
    curl http://127.0.0.1:8500/v1/agent/self  # XDS.Port == XDS.Ports.TLS == 8503
```

The only thing of note here was the second test, which requires the user to specify `-grpc-addr` when calling `consul connect envoy` since the command has no understanding on whether the TLS is enabled on the original port. This behavior existed prior to these commits.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/config/runtime.go
```diff
@@ -692,6 +697,21 @@ type RuntimeConfig struct {
...
 	GRPCAddrs []net.Addr
 
+	// GRPCTLSAddrs contains the list of TCP addresses and UNIX sockets the gRPC
+	// server will bind to. If the gRPC endpoint is disabled (ports.grpc <= 0)
+	// the list is empty.
+	//
+	// The addresses are taken from 'addresses.grpc_tls' which should contain a
+	// space separated list of ip addresses, UNIX socket paths and/or
+	// go-sockaddr templates. UNIX socket paths must be written as
+	// 'unix://<full path>', e.g. 'unix:///var/run/consul-grpc.sock'.
```
hashi-derek, line 707: This comment is copied from `GRPCPort`. Should we even allow unix socket definitions for this TLS-only port?

agent/agent_endpoint.go
```diff
@@ -45,7 +45,11 @@ type Self struct {
 
 type XDSSelf struct {
 	SupportedProxies map[string][]string
-	Port             int
+	// Port could be used for either TLS or plain-text communication
+	// up through version 1.14. In order to maintain backwards-compatibility,
+	// Port will now default to TLS and fallback to the standard port value.
+	Port    int
```
hashi-derek, line 51: Note that the semantics of this `Port` field are changing somewhat. Rather than being a direct representation of the `ports.grpc` field, it is instead referring to the "preferred" port so that backwards compatibility is not broken.

If you look at the first commit in this PR, you will see that an additional `PortPlainText` field was added prior to making backwards compatibility tweaks. This should likely be re-introduced after the deprecated TLS compatibility mode is eventually removed.

hashi-derek:
It looks like there may be a few more changes required if we want to include the new tls port in the serf gossip (the original port is included for some reason). I'll look into it to see how it works and will probably have to add some extra code.

agent/agent_endpoint.go
hashi-derek: This documentation is why I was hesitant to change it that way.

https://www.consul.io/api-docs/agent#read-configuration

"its format will not change in a backwards incompatible way between releases."

agent/agent.go
```diff
@@ -213,10 +213,14 @@ type Agent struct {
 	// depending on the configuration
 	delegate delegate
 
-	// externalGRPCServer is the gRPC server exposed on the dedicated gRPC port (as
+	// externalGRPCServer is the gRPC server exposed on a dedicated gRPC port (as
 	// opposed to the multiplexed "server" port).
 	externalGRPCServer *grpc.Server
 
+	// externalGRPCTLSServer is the gRPC server exposed on a dedicated gRPC-TLS port (as
```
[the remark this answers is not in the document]
hashi-derek: Okay, but will that be an issue in the future when we want to restrict peering to `ports.grpc_tls` only? I figured keeping them completely separate would be simpler long-term if each port wanted to have different endpoints registered to them.

agent/agent_endpoint.go
hashi-derek: Actually, I just realized that I accidentally thought the `XDS` field was nested underneath the `Config` element, but it's not. So I suppose it's safe to change (I doubt anyone would reference XDS but us).

agent/agent.go
```diff
@@ -3934,6 +3970,9 @@ func (a *Agent) reloadConfigInternal(newCfg *config.RuntimeConfig) error {
 		return fmt.Errorf("Failed reloading tls configuration: %s", err)
 	}
 
+	// Setup the external GRPC server.
+	a.externalGRPCServer = external.NewServer(a.logger.Named("grpc.external"))
```
mkeeler, line 3974: Why do we need to rebuild the grpc server here?

mkeeler (approved):

PR #14313, opened 2022-08-23
no-op: refactor usagemetrics tests for clarity and DRY cases

skpratt:
### Description
These tests copy a lot of repeated data, making it hard to discern what elements change between different cases, and any changes to the metrics themselves would require contributors to copy and paste them many times. This change makes the metrics tests more compliant with DRY principle, so it is easier to see what things change between test cases and add new metrics.

No-op, refactor

### PR Checklist

* [x] updated test coverage


boxofrad (approved):
Looks great, nice job! 👍🏻 

That said, I'm not _super_ familiar with this part of Consul, so if you did want a more thorough review @crhino might be your person.

PR #14314, opened 2022-08-23
Replace ring buffer with async version

kisunji:
### Description
Follow-up to https://github.com/hashicorp/consul/pull/14285

Currently we have a static list of server addresses from the peering token that may get stale over time. We need to watch for changes to peerings and update the server addresses which get served by the ring buffer.

* Added a testcase which updates an established peering with a bad server address, then confirms that we try to dial it at least once.
* Did not test end-to-end from a cluster generating raft autopilot events (mock events were tested in https://github.com/hashicorp/consul/pull/14285)

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/consul/leader_peering.go
+++ b/agent/consul/leader_peering.go
@@ -277,13 +277,6 @@ func (s *Server) establishStream(ctx context.Context, logger hclog.Logger, ws me
 		return fmt.Errorf("failed to build TLS dial option from peering: %w", err)
 	}
 
-	// Create a ring buffer to cycle through peer addresses in the retry loop below.
-	buffer := ring.New(len(peer.PeerServerAddresses))
-	for _, addr := range peer.PeerServerAddresses {
-		buffer.Value = addr
-		buffer = buffer.Next()
-	}
-
 	secret, err := s.fsm.State().PeeringSecretsRead(ws, peer.ID)
 	if err != nil {
 		return fmt.Errorf("failed to read secret for peering: %w", err)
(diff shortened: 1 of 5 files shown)


agent/consul/state/peering.go
```diff
@@ -7,12 +7,13 @@ import (
 	"strings"
 
 	"github.com/golang/protobuf/proto"
+	"github.com/hashicorp/go-memdb"
+
```
DanStough, line 11: weird

DanStough (commented):
Just stopped in to absorb the other work on peering. LGTM but I'll let others chime in.

agent/consul/leader_peering.go
```diff
@@ -294,27 +287,96 @@ func (s *Server) establishStream(ctx context.Context, logger hclog.Logger, ws me
...
+				// (not the one fetched just above) because the caller
+				// has pre-checked `peer.ShouldDial`, guaranteeing
+				// at least one server address. Further reads from the
+				// state store may not have this guarantee.
+				ringbuf = ring.New(len(peer.PeerServerAddresses))
+				for _, addr := range peer.PeerServerAddresses {
+					ringbuf.Value = addr
+					ringbuf = ringbuf.Next()
+				}
+				return nil
```
erichaberkorn, line 329: I might be missing something, but it looks like we can eliminate `init`,  move this out of `fetchAddrs` and replace the call to `fetchAddrs` prior to the `for` loop with this.

kisunji: We rely on the first `fetchAddrs` call to set up watches the `innerWs` as well; I could keep it separate if it makes it easier to parse. No strong opinions here

erichaberkorn: Yeah, same. I feel like it's slightly simpler to make `fetchAddrs` have one mode, but it's essentially equivalent. I'm just leaving this comment in passing while I read peering PRs to get up to speed. Feel free to ignore this.

I think this is equivalent: 

```diff
diff --git a/agent/consul/leader_peering.go b/agent/consul/leader_peering.go
index d2b3f071e..afc6b0302 100644
--- a/agent/consul/leader_peering.go
+++ b/agent/consul/leader_peering.go
@@ -304,30 +304,15 @@ func (s *Server) establishStream(ctx context.Context, logger hclog.Logger, ws me
 	go func() {
 		defer close(nextServerAddr)
 
-		var (
-			init    bool
-			ringbuf *ring.Ring
-			innerWs memdb.WatchSet
-		)
+		innerWs := memdb.NewWatchSet()
+		ringbuf := ring.New(len(peer.PeerServerAddresses))
+		for _, addr := range peer.PeerServerAddresses {
+			ringbuf.Value = addr
+			ringbuf = ringbuf.Next()
+		}
 
 		fetchAddrs := func() error {
-			innerWs = memdb.NewWatchSet()
 			_, peering, err := s.fsm.State().PeeringReadByID(innerWs, peer.ID)
-			if !init {
-				init = true
-				// On the first call to `fetchAddrs` we initialize the
-				// ring buffer with the peer passed to `establishStream`
-				// (not the one fetched just above) because the caller
-				// has pre-checked `peer.ShouldDial`, guaranteeing
-				// at least one server address. Further reads from the
-				// state store may not have this guarantee.
-				ringbuf = ring.New(len(peer.PeerServerAddresses))
-				for _, addr := range peer.PeerServerAddresses {
-					ringbuf.Value = addr
-					ringbuf = ringbuf.Next()
-				}
-				return nil
-			}
 			if err != nil {
 				return fmt.Errorf("failed to fetch peer %q: %w", peer.ID, err)
 			}
@@ -345,11 +330,6 @@ func (s *Server) establishStream(ctx context.Context, logger hclog.Logger, ws me
 			}
 			return nil
 		}
-		if err := fetchAddrs(); err != nil {
-			s.logger.Warn("watchset for peer was fired but failed to update server addresses",
-				"peer_id", peer.ID,
-				"error", err)
-		}
 
 		for {
 			select {

```

kisunji: Your diff is missing the first call to `s.fsmState().PeeringReadByID` and its error handling (otherwise the watchset never wakes up) but I think it does read cleaner. Will push a change

agent/consul/state/peering.go
```diff
@@ -981,7 +982,7 @@ func peeringsForServiceTxn(tx ReadTxn, ws memdb.WatchSet, serviceName string, en
 		if idx > maxIdx {
 			maxIdx = idx
 		}
-		if peering == nil || !peering.IsActive() {
+		if !peering.IsActive() {
```
erichaberkorn, line 985: Nice catch

agent/consul/state/peering.go
kisunji: It got moved because it's not `hashicorp/consul` 😆

agent/consul/leader_peering_test.go
```diff
@@ -1343,3 +1345,137 @@ func Test_isFailedPreconditionErr(t *testing.T) {
...
+		// we will update the token with bad addresses to assert it doesn't clobber existing ones
+		token.ServerAddresses = []string{"1.2.3.4:1234"}
+
+		badToken, err := acceptor.peeringBackend.EncodeToken(token)
+		require.NoError(t, err)
+
+		ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+		t.Cleanup(cancel)
+
+		// try establishing
```
freddygv, line 1445: Took me a minute to see what was going on here. Does this seem accurate to you?
```suggestion
		// Try establishing.
		// This call will only succeed if the bad address was not used in the calls to exchange the peering secret.
```

agent/consul/leader_peering_test.go
```diff
@@ -1343,3 +1345,137 @@ func Test_isFailedPreconditionErr(t *testing.T) {
...
+		}
+		_, err = dialerClient.Establish(ctx, &establishReq)
+		require.NoError(t, err)
+
+		p, err := dialerClient.PeeringRead(ctx, &pbpeering.PeeringReadRequest{Name: "my-peer-acceptor"})
+		require.NoError(t, err)
+		require.NotContains(t, p.Peering.PeerServerAddresses, "1.2.3.4:1234")
+	})
+
+	testutil.RunStep(t, "updated server addresses are picked up by the leader", func(t *testing.T) {
```
freddygv, line 1458: One option to test this in a more realistic scenario could be:
* LAN join a new acceptor server to dc1
* Wait for dialer to receive update by reading the peering
* Shut down the old acceptor
* Verify that the stream recovers

freddygv: Actually I just noticed this isn't branched off of your other PR

freddygv (approved):
Looks great! Left some non-blocking comments

agent/consul/leader_peering.go
```diff
@@ -294,27 +287,89 @@ func (s *Server) establishStream(ctx context.Context, logger hclog.Logger, ws me
...
+	streamCtx, cancel := context.WithCancel(ctx)
+	cancelFns[peer.ID] = cancel
+
+	// The following goroutine sends an up-to-date peer server address to nextServerAddr.
+	// It loads the server addresses into a ring buffer and cycles through them until:
+	//   1) streamCtx is cancelled (peer is deleted)
+	//   2) the peer is modified and the watchset fires.
+	// In case (2) we refetch the peering and rebuild the ring buffer.
+	nextServerAddr := make(chan string)
+	go func() {
```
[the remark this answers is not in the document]
kisunji: yep, will make the change

agent/consul/leader_peering_test.go
kisunji: Yeah once both PRs are in I can replace this test with a proper LAN-joined cluster one

PR #12434, opened 2022-02-24
[1.10.x] Add support for Envoy 1.20.2

eculver:
This follows #12433 to add Envoy 1.20 to the integration test suite. The same notes from #12433 apply for dealing with the version.

PR #14334, opened 2022-08-24
[1.10] docs: Correct Envoy support matrix

eculver:
### Description
Long ago, when I was first learning how our Envoy support matrix worked, I inadvertently backported support for Envoy 1.20 to Consul 1.10: https://github.com/hashicorp/consul/pull/12434. Since this patch was related to a CVE, my guess is that we wanted to be able to give users an upgrade path to a patched Envoy. However, for Consul 1.10, they could do this with Envoy 1.18.6 as opposed to adding support for Envoy 1.20. Alas, this is the history and I cannot go back and change it.

With that said, we don't want to confuse users about our Envoy support matrix, so this resets our docs for 1.10 to what the expected support should be for Consul 1.10. These values are what should also be documented in the Consul 1.12 and 1.11 docs. There will be accompanying PRs for Consul 1.12 and 1.11 linked below.

In the end, all the docs linked below should have consistent Envoy versions for each release series.

### Links
- [Envoy integration docs (1.13)](https://www.consul.io/docs/v1.13.x/connect/proxies/envoy)
- [Envoy integration docs (1.12)](https://www.consul.io/docs/v1.12.x/connect/proxies/envoy) 
- [Envoy integration docs (1.11)](https://www.consul.io/docs/v1.11.x/connect/proxies/envoy)
- [Envoy integration docs (1.10)](https://www.consul.io/docs/v1.10.x/connect/proxies/envoy)
- https://github.com/hashicorp/consul/pull/14336
- https://github.com/hashicorp/consul/pull/14335


eculver:
Ignore the integration test failures. It is a known issue for them to fail since they are testing against `nomad@main` which now depends on Go 1.19.

david-yu (approved):

PR #14335, opened 2022-08-24
[1.11] docs: Update latest Envoy to 1.20.6, make docs consistent

eculver:
### Description
There are currently discrepancies between the docs in each release series, specifically for 1.10. This updates them all to agree. There will be an accompanying PR referenced below for 1.10.

### Links
- [Envoy integration docs (1.13)](https://www.consul.io/docs/v1.13.x/connect/proxies/envoy)
- [Envoy integration docs (1.12)](https://www.consul.io/docs/v1.12.x/connect/proxies/envoy) 
- [Envoy integration docs (1.11)](https://www.consul.io/docs/v1.11.x/connect/proxies/envoy)
- [Envoy integration docs (1.10)](https://www.consul.io/docs/v1.10.x/connect/proxies/envoy)
- https://github.com/hashicorp/consul/pull/14334

### DIFF
--- a/website/content/docs/connect/proxies/envoy.mdx
+++ b/website/content/docs/connect/proxies/envoy.mdx
@@ -36,8 +36,8 @@ Consul supports **four major Envoy releases** at the beginning of each major Con
 
 | Consul Version      | Compatible Envoy Versions                                                          |
 | ------------------- | -----------------------------------------------------------------------------------|
-| 1.11.x              | 1.20.4, 1.19.5, 1.18.6, 1.17.4<sup>1</sup>                                         |
-| 1.10.x              | 1.18.6, 1.17.4<sup>1</sup>, 1.16.5<sup>1</sup> , 1.15.5<sup>1</sup>                |
+| 1.11.x              | 1.20.6, 1.19.5, 1.18.6, 1.17.4<sup>1</sup>                                         |
+| 1.10.x              | 1.18.6, 1.17.4<sup>1</sup>, 1.16.5<sup>1</sup>, 1.15.5<sup>1</sup>                 |
 | 1.9.x               | 1.16.5<sup>1</sup>, 1.15.5<sup>1</sup>, 1.14.7<sup>1,2</sup>, 1.13.7<sup>1,2</sup> |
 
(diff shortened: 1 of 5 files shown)


website/content/docs/connect/proxies/envoy.mdx
```diff
@@ -36,8 +36,8 @@ Consul supports **four major Envoy releases** at the beginning of each major Con
 
 | Consul Version      | Compatible Envoy Versions                                                          |
 | ------------------- | -----------------------------------------------------------------------------------|
-| 1.11.x              | 1.20.4, 1.19.5, 1.18.6, 1.17.4<sup>1</sup>                                         |
-| 1.10.x              | 1.18.6, 1.17.4<sup>1</sup>, 1.16.5<sup>1</sup> , 1.15.5<sup>1</sup>                |
+| 1.11.x              | 1.20.6, 1.19.5, 1.18.6, 1.17.4<sup>1</sup>                                         |
```
jkirschner-hashicorp, line 39: @eculver : let's say a user is currently running Envoy 1.20.4 with Consul 1.11.x. Is Envoy 1.20.4 _incompatible_ with Consul? Or is it just not _recommended_ (such as because of bugs/CVEs in Envoy that have since been fixed in 1.20.6)?

david-yu (approved):

website/content/docs/connect/proxies/envoy.mdx
david-yu: It's not recommended due to CVEs but is compatible to with Consul 1.11.x

jkirschner-hashicorp: I wonder if we should make a distinction between compatible vs recommended. This table suggestions that only 1.20.6 is compatible, which is a distinction that may matter for existing users (even if it doesn't for new users).

david-yu: I would suggest we state what we actually test, so that would depend on whether we actually do test against all the patch versions of Envoy. If not I think it's ok to add an asterisk here to say that although all patch releases are compatible we do not test against every patch release version.

eculver: Consul 1.11.x and Envoy 1.20.4 are compatible. It's a `<=` semantic for the Envoy minor versions. So, in the case of 1.11.x and Envoy 1.20, the meaning of this could be read as "Consul 1.11.x is compatible with Envoy versions _up to_ 1.20.6".

Issue #14295, opened 2022-08-22
Setup consul UI on Kubernetes + nginx ingress controller on path "/consul" instead of "/" 


neo3matrix:
Hi all,
This may sound like a repeat question and I checked some similar answers but not sure what is going wrong with my setup, so asking this question here. (referred [issue#758](https://github.com/hashicorp/consul-helm/issues/758) and tried solution but not working for me).
I am sure I am doing something wrong - someone please help me.

**What I have so far:**
- Have a k8s cluster with metallb load balancer and nginx ingress controller already setup.

**What I expect:**
- With helm installation of consul, I need to setup consul ingress in such a way that the UI path should be:  `myweb.mydommain.com/consul`

**What's happening:**
- After setup on consul, if I visit browser and type: myweb.mydomain.com/consul - I get following error: 
`Invalid URL path: if attempting to use the HTTP API, ensure the path starts with '/v1/'
`
- But, If I change the "path" from "/consul" to "/" of consul ingress values in values.yaml, then `myweb.mydomain.com` shows me consul UI properly.

**My setup:** 
K8s version: 

> $ kubectl version
> Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.10", GitCommit:"a7a32748b5c60445c4c7ee904caf01b91f2dbb71", GitTreeState:"clean", BuildDate:"2022-02-16T11:24:04Z", GoVersion:"go1.16.14", Compiler:"gc", Platform:"linux/amd64"}
> Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.10", GitCommit:"a7a32748b5c60445c4c7ee904caf01b91f2dbb71", GitTreeState:"clean", BuildDate:"2022-02-16T11:18:16Z", GoVersion:"go1.16.14", Compiler:"gc", Platform:"linux/amd64"}

nginx and consul helm charts:

> $ helm list -A (Trimmed output)
> NAME                  	NAMESPACE             	REVISION	UPDATED                                	STATUS  	CHART                                 	APP VERSION
> consul                	consul                	1       	2022-08-22 19:42:43.496703951 +0000 UTC	deployed	consul-0.46.1                         	1.12.3
> nginx-stable          	nginx-stable          	1       	2022-08-03 22:33:45.242779695 +0000 UTC	deployed	nginx-ingress-0.14.0                  	2.3.0
> 

consul values.yaml:
```
global:
  name: consul
ui:
  ingress:
    enabled: true
[...]

Issue #11627, opened 2021-11-22
Sharing Hashicorp Vault and Consul UI under one roof behind NGINX proxy_pass


radioipcloud:
### Overview of the Issue

It might come in handy sometimes to share both Vault and Consul UI under the same roof behind nginx.

### Describe the solution you'd like

A single configuration location file for nginx to hide on the same host 
consul and vault in <hostname>/ui/<consul/vault>/

### Consul Version

Consul version 1.10.3
Vault version 1.8.5

### Solution

location /ui {
	if ($http_referer ~ (/ui/vault)) { 
            proxy_pass http://127.0.0.1:8200;
	}
	if ($http_referer ~ (/ui/consul)) {
	    proxy_pass http://127.0.0.1:8500;
	}
}

location /v1 {
        if ($http_referer ~ (/ui/vault)) {
            proxy_pass http://127.0.0.1:8200;
        }
        if ($http_referer ~ (/ui/consul)) {
            proxy_pass http://127.0.0.1:8500;
        }
}

location /ui/vault/ {
	proxy_pass http://127.0.0.1:8200/ui/;
}

location /ui/consul {
        proxy_pass http://127.0.0.1:8500;
}

### Additionnal Notes

- Make sure to set -ui-content-path=/ui/consul/ when starting consul.

Amier3:
Hey @radioipcloud 

I see this is your first issue so welcome to the hashicorp community! This is a really cool idea, and i'm interested to know what problems you were/are running into that led you to this solution?

radioipcloud:
Thanks @Amier3 Well mainly we were using nginx to hide various web services and tooling i.e. pgadmin, vault ui and consul ui and wanted to see if we could have them coexist under the same umbrella so to speak. This is how we managed to do it. We had seen  various solutions online but none seem to work. Decided in the end to share the experience so that others could do the same if need be.

JeevanAleti:
Hi,

Is there anyway to make the nginx proxy_pass working without using the `location /v1 {` block, because in my nginx file the /v1 location is being used by another service.

Issue #10374, opened 2021-06-10
Web UI .js files code ignores -ui-content-path flag


ptecza:
#### Overview of the Issue

I noticed that Web UI of Consul v1.9.3 generates .js files which ignore `-ui-content-path` flag. I checked the latest Consul code and it seems that it has a hardcoded `GET /v1/...` HTTP requests as well. I have to use `sub_filter` rules of my Nginx reverse proxy to fix the invalid requests.

#### Reproduction Steps

Steps to reproduce this issue:

1. Create a cluster with 3 nodes using `-ui` and `-ui-content-path /foo/ui/` flags for them
2. Configure Nginx reverse proxy for handling the cluster, e.g.:
```
upstream foo_consul_backend {
        server consul1.foo.example.com:8500;
        server consul2.foo.example.com:8500;
        server consul3.foo.example.com:8500;
}

server {
        listen 443;
        server_name consul.example.com;

        ssl on;
        ssl_certificate     /etc/letsencrypt/live/consul.example.com/fullchain.pem;
        ssl_certificate_key /etc/letsencrypt/live/consul.example.com/privkey.pem;

        access_log /var/log/nginx/consul-access-ssl.log;
        error_log  /var/log/nginx/consul-error-ssl.log;

        location / {
                root /var/www/html/consul/;
                index index.html;
        }

        location /foo/ui/ {
                proxy_set_header Accept-Encoding "";
                proxy_set_header X-Real-IP $remote_addr;
                proxy_set_header Host $host;
                proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
                proxy_pass http://foo_consul_backend/foo/ui/;
        }
        location /foo/v1/ {
                proxy_pass http://foo_consul_backend/v1/;
        }
```
3. Go to https://consul.example.com/foo/ui/ using your favourite web browser
4. See an empty page

### Consul info for Server

<details>
  <summary>Server info</summary>

```
agent:
        check_monitors = 1
        check_ttls = 0
        checks = 1
        services = 0
build:
        prerelease =
        revision = f55da930
        version = 1.9.3
consul:
        acl = disabled
        bootstrap = false
        known_datacenters = 1
        leader = false
        leader_addr = 10.10.10.15:8300
        server = true
raft:
        applied_index = 774
        commit_index = 774
        fsm_pending = 0
        last_contact = 84.006927ms
        last_log_index = 774
        last_log_term = 2
        last_snapshot_index = 0
[...]

Issue #9829, opened 2021-02-25
Allow add prefix for API url in the UI


Rwing:
In the reverse proxy case, the Consul mapped to http://example.com/consul, and the correct api url is /consul/v1/aa/bb not /v1/aa/bb, so the url prefix should be allowed to be modified in the UI's javascript, like this
https://github.com/hashicorp/consul/blob/be89557fb43a4b752fbd087d9c7f8f96e6566d4c/ui/packages/consul-ui/app/adapters/dc.js#L6

johncowen:
Hi @Rwing 

Thanks for the feature request! We have discussed this within the team and there is already a little thought elsewhere in the UI to allow us to let folks do this.

We also recently added a dedicated config block (`ui_config {}`) for allowing the operator to configure UI related things as well as just straightforward enable UI/disable UI, so we have most things we would need in place now to be able to do this.

I'll bring it up again with the rest of the team and I'll try and can keep you posted here on any further progress.

thesobercoder:
@johncowen Any update on this. This is still not working. :(

johncowen:
Hi @thesobercoder !

This is something we are actively talking about, and it's something we intend to address. I'd say maybe keep your eyes peeled here and I'll try and keep you updated. Just to be clear, this isn't a feature that exists as yet.

Hopefully I can give some more info soonish.

Thanks!

jkirschner-hashicorp:
Related to #7928

danparisi:
Was it fixed here? 
https://github.com/hashicorp/consul/pull/14342

If yes, can anyone explain how does it work? I cannot find any docs

PR #14342, opened 2022-08-25
ui: Add support for prefixing the API path

johncowen:
### Description
Adds support for prefixing the path of API call made by the UI. Along with [`ui-config.content_path`](https://www.consul.io/docs/agent/config/config-files#ui_config_content_path), this will enable the UI to be usable at different paths behind reverse proxies etc, which is one of the UIs most requested features.

Please note: We have always spoken about the possibility of our `ui_config` being provided by an API requests rather than being supplied via our "env vars". This variable is slightly strange in that it can never be provided by an API call, as it provides us with the information as to where we need to make API calls, therefore this variable should always be thought of as an "env var"

The variable defaults to an empty string, and can be mocked out during development via either our "cookie env vars" or via a real env var at boot time (when running `make start`)

### Links

This (along with some yet to be completed backend changes) will eventually help the following issues, amongst others.

- https://github.com/hashicorp/consul/issues/14295
- https://github.com/hashicorp/consul/issues/11627
- https://github.com/hashicorp/consul/issues/10374
- https://github.com/hashicorp/consul/issues/9829
- https://github.com/hashicorp/consul/issues/8240
- https://github.com/hashicorp/consul/issues/7928

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

/cc @clly we'll possibly chat about this later

### DIFF
--- a/ui/packages/consul-ui/config/environment.js
+++ b/ui/packages/consul-ui/config/environment.js
@@ -120,6 +120,7 @@ module.exports = function(environment, $ = process.env) {
           enabled: true,
           endpoints: {
             '/v1': '/mock-api/v1',
+            '/prefixed-api': '/mock-api/prefixed-api',
           },
         },
         APP: Object.assign({}, ENV.APP, {
--- a/ui/packages/consul-ui/config/environment.js
+++ b/ui/packages/consul-ui/config/environment.js
@@ -86,6 +86,7 @@ module.exports = function(environment, $ = process.env) {
       PartitionsEnabled: false,
       LocalDatacenter: env('CONSUL_DATACENTER_LOCAL', 'dc1'),
       PrimaryDatacenter: env('CONSUL_DATACENTER_PRIMARY', 'dc1'),
+      APIPrefix: env('CONSUL_API_PREFIX', '')
     },
 
     // Static variables used in multiple places throughout the UI
@@ -111,6 +112,7 @@ module.exports = function(environment, $ = process.env) {
           PartitionsEnabled: env('CONSUL_PARTITIONS_ENABLED', false),
           LocalDatacenter: env('CONSUL_DATACENTER_LOCAL', 'dc1'),
           PrimaryDatacenter: env('CONSUL_DATACENTER_PRIMARY', 'dc1'),
+          APIPrefix: env('CONSUL_API_PREFIX', '')
         },
 
         '@hashicorp/ember-cli-api-double': {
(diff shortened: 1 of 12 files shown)


ui/packages/consul-ui/config/environment.js
```diff
@@ -176,7 +179,9 @@ module.exports = function(environment, $ = process.env) {
       ENV = Object.assign({}, ENV, {
         // in production operatorConfig is populated at consul runtime from
         // operator configuration
-        operatorConfig: {},
+        operatorConfig: {
+          APIPrefix: ''
+        },
```
johncowen, line 184: @clly would it be possible to default this to an empty string from your side? i.e. if operator hasn't set the config then it comes through as `''` to us?

LevelbossMike (changes requested):
Looks good to me. Can we add a test for this behavior?

ui/packages/consul-ui/tests/acceptance/steps/api-prefix-steps.js
```diff
@@ -0,0 +1,11 @@
+import steps from './steps';
+
+// step definitions that are shared between features should be moved to the
+// tests/acceptance/steps/steps.js file
+
+export default function(assert) {
```
LevelbossMike, line 6: Do we need this file?

ui/packages/consul-ui/tests/steps.js
```diff
@@ -87,6 +87,7 @@ export default function({
     api.server.respondWith(url.split('?')[0], data);
   };
   const setCookie = function(key, value) {
+    document.cookie = `${key}=${value}`;
```
LevelbossMike, line 90: Will we share this cookie state between tests? do we need to reset this cookie after the test is complete again?

LevelbossMike (approved):
Thanks for adding the test – I have two more questions, but should be good to go based on your judgement.

ui/packages/consul-ui/tests/acceptance/steps/api-prefix-steps.js
johncowen: Yeah, every feature file MUST have a corresponding one of these due to the ember addon we use, it's annoying I know, a while back I used something different to get around this but it never made it in.

ui/packages/consul-ui/tests/steps.js
johncowen: Nope no sharing between tests. All cookies are reset in a beforeEach hook

feigaogit:
Adds support for prefixing the path of API call made by the UI. Along with [ui-config.content_path](https://www.consul.io/docs/agent/config/config-files#ui_config_content_path), this will enable the UI to be usable at different paths behind reverse proxies etc, which is one of the UIs most requested features.


How to use this parameter is not explained in the document，May I ask how to use it?

OOOllie:
To add to this, I've added an environment variable inside the docker container CONSUL_API_PREFIX but it does not change the requests or where they go to. Is there any fixes/tests for this?

danparisi:
Same for me, I cannot let it work

qasim-sajid-snkeos:
Is this feature complete and works already?
If yes, is it documented somewhere how to do it?

Issue #12106, opened 2022-01-17
how to configure sidecar proxy to advertise "h2" as ALPN


rgf2004:
#### The question:
How to configure the sidecar proxy to advertise `h2` as the ALPN?

#### The Scenario:
I have a gRPC service with a sidecar proxy. To it, other services should communicate natively. 

#### Steps 
- I followed the steps mentioned [here](https://learn.hashicorp.com/tutorials/consul/service-mesh-with-envoy-proxy?in=consul/getting-started) to register a service, and its sidecar proxy.
- I followed the steps mentioned [here](https://www.consul.io/docs/connect/native) to connect natively. 
- When other services attempt to establish a secure connection with this service's sidecar proxy, it fails with `Failed ALPN negotiation: Unable to find compatible protocol`. this is because the sidecar proxy doesn't expose the ALPN protocol, I assume so.  
- I see that envoy has a configuration for `alpn_protocols` [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/tls.proto) but I don't know if consul supports such configuration. 

Service definition, `grpc-service.hcl`
```
service {
  name = "grpc-service"
  id = "grpc-service-1"
  port = 9000

  connect = {
    sidecar_service = {}
  }

  check {
    id       = "grpc-service-check"
    grpc =  "127.0.0.1:9000"
    grpc_use_tls = false
    interval =  "10s"
  }
}
``` 
and this one, `additional-config.hcl`
```
Kind = "service-defaults"
Name = "grpc-service"
Protocol = "grpc"
MeshGateway {
	Mode = "none"
}
Meta { 
}
```

These are the used commands:
```
consul agent -dev -node machine
```
```
consul services register grpc-service.hcl 
```
```
consul config write additional-config.hcl
```
```
consul connect envoy -sidecar-for grpc-service-1 
```

Amier3:
This is related to #11907 and should be fixed when the PR  #12170 is merged

jcantwell-JC:
We are sad to see that PR https://github.com/hashicorp/consul/pull/12170 was closed. We have an Ingress Gateway configured through Nomad that needs to support `grpc`. However, without the envoy gateway broadcasting the appropriate `alpn` header, `grpc` clients will not connect to it (i.e. `python` client in this case).

```
Cannot check peer: missing selected ALPN property.
```

[...]

Issue #11907, opened 2021-12-22
connect native to a sidecar proxy fails with "Failed ALPN negotiation: Unable to find compatible protocol"


rgf2004:
The sidecar proxy returns null/empty `applicationProtocol` during the TLS negotiation. 

- I'm trying to establish a native integration with a gRPC java service running in nomad and has a sidecar proxy (envoy proxy). 
- The issue happens during the TLS negotiation that the proxy replies with null/empty `applicationProtocol`.
- I tried to update the envoy configuration `DownstreamTlsContext.CommonTlsContext.alpn_protocols` for the public listener to add `h2` as the application protocol using this consul [configuration parameter](https://www.consul.io/docs/connect/proxies/envoy#envoy_public_listener_json) however this didn't work for the following reasons:
   
Reason - 1: My configuration is not reflected in the envoy configuration. I assume this because consul overwrites it by the Connect TLS certificates as per this documentation
```
Every FilterChain added to the listener will have its TlsContext overridden by the Connect TLS certificates 
and validation context. This means there is no way to override Connect's mutual TLS for the public listener.
```

   Reason - 2: Even if I override the public listener configuration, which I don't know how yet:), I can't find a way how to get the dynamic port value that proxy gets in `listener.address.socket_address.port_value` when it is created without overriding. 

Java stack trace
```
io.grpc.StatusRuntimeException: UNAVAILABLE: Connection closed while performing protocol negotiation for [SslHandler#0, ProtocolNegotiators$ClientTlsHandler#0, WriteBufferingAndExceptionHandler#0, DefaultChannelPipeline$TailContext#0]
	at io.grpc.Status.asRuntimeException(Status.java:526)
	at io.grpc.netty.WriteBufferingAndExceptionHandler.channelInactive(WriteBufferingAndExceptionHandler.java:81)
	...
17:04:24.182 [grpc-nio-worker-ELG-1-4] DEBUG io.netty.handler.ssl.SslHandler.setHandshakeSuccess(1829) - [id: 0x46522a81, L:/127.0.0.1:57819 - R:/127.0.0.1:28430] HANDSHAKEN: protocol:TLSv1.3 cipher suite:TLS_AES_128_GCM_SHA256
17:04:24.182 [grpc-nio-worker-ELG-1-4] DEBUG io.grpc.netty.ProtocolNegotiators.logSslEngineDetails(902) - TLS negotiation failed.
SSLEngine Details: [
[...]

PR #14356, opened 2022-08-26
Configure Envoy alpn_protocols based on service protocol

oulman:
### Description

This PR is largely based on @blake's work in https://github.com/hashicorp/consul/pull/12170 but updated to use the TLSContext functions introduced in https://github.com/hashicorp/consul/pull/13321. I'm not sure what the best path forward for configuring this from the Consul teams perspective but I wanted to raise a PR to start a conversation.

This fixes https://github.com/hashicorp/consul/issues/11907, resolves https://github.com/hashicorp/consul/issues/12106, and [this internal feature request](https://go.hashi.co/alpn_protocols_fr).

In my testing I verified that gRPC and HTTP2 services can negotiate h2 and http/1.1 and that services configured as HTTP will only negotiate http/1.1. 

### Links

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


agent/xds/listeners.go
```diff
@@ -1022,6 +1023,19 @@ func (s *ResourceGenerator) injectConnectTLSForPublicListener(cfgSnap *proxycfg.
 	return nil
 }
 
+func getAlpnProtocols(protocol string) []string {
+	var alpnProtocols []string
+
+	switch protocol {
+	case "grpc", "http2":
```
malizz, line 1030: I believe `grpc` and `http2` cases behave differently. `http2` should only return `h2`.

malizz:
Hi @oulman! Thanks for this PR, it looks good, but we can work on adding more tests and verifying the intended behavior.

agent/xds/listeners.go
oulman: Thanks! I've updated the function to return `h2` only for `http2`.

oulman:
Hi @malizz - Thanks! I've pushed a couple of changes.

- The linter was failing because the error handling on ParseProxyConfig was empty. The original PR had a commented out call to a logger but I didn't see one available in the function. In the interim I went ahead and excluded that [here](https://github.com/hashicorp/consul/pull/14356/commits/b4cb2537014cf20ab52b9d037f6e7de3ba180bae)
- I created a [test for getAlpnProtocol()](https://github.com/hashicorp/consul/pull/14356/commits/e2e342bd9b10f2bdb8f0ffa9b1c0754ac5491ab4)
- I added [tests](https://github.com/hashicorp/consul/pull/14356/commits/bd6ec7f1ce510462853bed071a477d9a2fb8ed42) for http2 and grpc public listeners to validate that the Envoy `alpn_protocols` setting was getting configured correctly.

malizz:
@oulman Question: you made the changes to listeners tlsContext only, but I’m wondering if we need to update the TLS contexts for other listeners (ingress listener for example) or pass through clusters too.

oulman:
Hi @malizz! I updated the context for the Ingress Gateway listener and added tests.

Looking at pass-through clusters I'm a little confused on what actually needs to be done to support this. I found some issues  ([#1](https://github.com/envoyproxy/envoy/issues/11909#issuecomment-654903211), [#2](https://github.com/envoyproxy/envoy/issues/14378#issuecomment-743713707)) that make it sound like for Envoy <-> Envoy traffic ALPN isn't used in protocol negotiation. If mesh-to-mesh is OK as-is, do we need to enable ALPN negotiation for external upstream HTTP clusters?[1]

Thanks!

1. I also found [this Envoy PR](https://github.com/envoyproxy/envoy/pull/13922) which enables ALPN negotiation via [http_protocol_options](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/upstreams/http/v3/http_protocol_options.proto#envoy-v3-api-msg-extensions-upstreams-http-v3-httpprotocoloptions-autohttpconfig).

agent/proxycfg/testing_ingress_gateway.go
```diff
@@ -1485,6 +1485,164 @@ func TestConfigSnapshotIngressGateway_SingleTLSListener(t testing.T) *ConfigSnap
 		})
 }
```
malizz, line 1487: I have a question, is it possible to have an Ingress gateway listener that listens to port x for grpc and port y for http2?

malizz (changes requested):
The PR looks mostly great. I added a question and requested a few changes. Thanks.

agent/xds/listeners_test.go
```diff
@@ -1115,3 +1141,34 @@ func TestResolveListenerSDSConfig(t *testing.T) {
...
+		},
+		"grpc": {
+			protocol: "grpc",
+			want:     []string{"h2", "http/1.1"},
+		},
+		"empty": {
+			protocol: "",
+			want:     nil,
+		},
+	}
```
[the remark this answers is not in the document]
oulman: 👍 I added an explicit case for `tcp`.

agent/xds/listeners.go
```diff
@@ -1030,13 +1046,22 @@ func createDownstreamTransportSocketForConnectTLS(cfgSnap *proxycfg.ConfigSnapsh
 		return nil, fmt.Errorf("cannot inject peering trust bundles for kind %q", cfgSnap.Kind)
 	}
 
+	// Determine listener protocol type from configured service protocol. Don't hard fail on a config typo,
+	//The parse func returns default config if there is an error, so it's safe to continue.
+	cfg, _ := ParseProxyConfig(cfgSnap.Proxy.Config)
+
```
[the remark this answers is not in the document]
oulman: 🤔 Right now calling getAlpnProtocols with `tcp` is returning an implicit nil because it's not matched. To confirm, we want to add an case statement for tcp but it should still return the empty `alpnProtocols` slice.

agent/proxycfg/testing_ingress_gateway.go
blake: A listener only binds to a single port. It is possible to configure an ingress gateway with one listener on port X with the protocol as `grpc`, and different listener on port Y with the protocol configured as `http2`.

malizz: Thanks @blake. @oulman could you add a test function for this use case?

oulman: Sure! I just added a test for `TestConfigSnapshotIngressGateway_MultiTLSListener_MixedHTTP2gRPC` which validates two listeners (grpc and http2) with per-listener TLSConfig's.

The other scenario this made me think about is configuring TLS at the gateway [level](https://www.consul.io/docs/connect/config-entries/ingress-gateway#tls). From my read it looks like the Gateway level TLS is still merged down to the listener level and setting ALPN in `makeDownstreamTLSContextFromSnapshotListenerConfig()` still applies. I added another test `TestConfigSnapshotIngressGateway_GWTLSListener_MixedHTTP2gRPC` to validate that case. 

Thanks!

malizz (approved):
LGTM, thanks!

agent/xds/listeners.go
```diff
@@ -1022,6 +1023,21 @@ func (s *ResourceGenerator) injectConnectTLSForPublicListener(cfgSnap *proxycfg.
 	return nil
 }
 
+func getAlpnProtocols(protocol string) []string {
+	var alpnProtocols []string
+
+	switch protocol {
+	case "grpc":
+		alpnProtocols = append(alpnProtocols, "h2", "http/1.1")
+	case "http2":
```
malizz, line 1032: I think the correct list of protocols is "h2" and "http/1.1" here. Sorry for the mixup.

oulman: OK. Thanks! I pushed a new commit that switches this.

PR #14364, opened 2022-08-26
Allow terminated peerings to be deleted

freddygv:
### Description
Peerings are terminated when a peer decides to delete the peering from
their end. Deleting a peering sends a termination message to the peer
and triggers them to mark the peering as terminated but does NOT delete
the peering itself. This is to prevent peerings from disappearing from
both sides just because one side deleted them.

Previously the Delete endpoint was skipping the deletion if the peering
was not marked as active. However, terminated peerings are also
inactive.

This PR makes some updates so that peerings marked as terminated can be
deleted by users.

* Unit tests at state store and gRPC service layers

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/consul/state/peering_test.go
+++ b/agent/consul/state/peering_test.go
@@ -1112,16 +1112,22 @@ func TestStore_PeeringWrite(t *testing.T) {
 	// Each case depends on the previous.
 	s := NewStateStore(nil)
 
+	testTime := time.Now()
+
+	type expectations struct {
+		peering *pbpeering.Peering
+		secrets *pbpeering.PeeringSecrets
+		err     string
+	}
 	type testcase struct {
-		name          string
-		input         *pbpeering.PeeringWriteRequest
-		expectSecrets *pbpeering.PeeringSecrets
-		expectErr     string
+		name   string
+		input  *pbpeering.PeeringWriteRequest
+		expect expectations
 	}
 	run := func(t *testing.T, tc testcase) {
 		err := s.PeeringWrite(10, tc.input)
-		if tc.expectErr != "" {
-			testutil.RequireErrorContains(t, err, tc.expectErr)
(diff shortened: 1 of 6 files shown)


kisunji:
> Not sure why the CI result isn't being reported back to GH but it's green in circle: https://app.circleci.com/pipelines/github/hashicorp/consul/39187/workflows/cfba6351-2991-4330-8be9-727cb8465044

Usually happens when Github webhooks go down. Pushing an empty commit should kick it off again

agent/consul/state/peering_test.go
```diff
@@ -1133,95 +1142,281 @@ func TestStore_PeeringWrite(t *testing.T) {
...
+					Partition:           structs.NodeEnterpriseMetaInDefaultPartition().PartitionOrEmpty(),
 				},
 				SecretsRequest: &pbpeering.SecretsWriteRequest{
 					PeerID: testBazPeerID,
-					Request: &pbpeering.SecretsWriteRequest_GenerateToken{
-						GenerateToken: &pbpeering.SecretsWriteRequest_GenerateTokenRequest{
-							EstablishmentSecret: testBazSecretID,
+					Request: &pbpeering.SecretsWriteRequest_Establish{
+						Establish: &pbpeering.SecretsWriteRequest_EstablishRequest{
+							ActiveStreamSecret: testBazSecretID,
```
kisunji, line 1171: Should we have a separate create testcase for the acceptor side which generates tokens?

kisunji (approved):

agent/consul/state/peering_test.go
freddygv: I'm not sure that we gain much there since the logic for writing secrets is tested in `TestStore_PeeringSecretsWrite`

PR #14365, opened 2022-08-26
Run integration tests locally using amd64

lkysow:
Locally, always run integration tests using amd64, even if running
on an arm mac. This ensures the architecture locally always matches
the CI/CD environment.

In addition:
* Use consul:local for envoy integration and upgrade tests. Previously,
  consul:local was used for upgrade tests and consul-dev for integration
  tests. I didn't see a reason to use separate images as it's more
  confusing.
* By default, disable the requirement that aws credentials are set.
  These are only needed for the lambda tests and make it so you
  can't run any tests locally, even if you're not running the
  lambda tests. Now they'll only run if the LAMBDA_TESTS_ENABLED
  env var is set.
* Split out the building of the Docker image for integration
  tests into its own target from `dev-docker`. This allows us to always
  use an amd64 image without messing up the `dev-docker` target.
* Add support for passing GO_TEST_FLAGs to `test-envoy-integ` target.
* Add a wait_for_leader function because tests were failing locally
  without it.

To test on your own mac:

```
make test-envoy-integ GO_TEST_FLAGS="-run TestEnvoy/case-basic"
```

### DIFF
--- a/test/integration/connect/envoy/helpers.bash
+++ b/test/integration/connect/envoy/helpers.bash
@@ -562,14 +562,14 @@ function assert_intention_denied {
 function docker_consul {
   local DC=$1
   shift 1
-  docker run -i --rm --network container:envoy_consul-${DC}_1 consul-dev "$@"
+  docker run -i --rm --network container:envoy_consul-${DC}_1 consul:local "$@"
 }
 
 function docker_consul_for_proxy_bootstrap {
   local DC=$1
   shift 1
 
-  docker run -i --rm --network container:envoy_consul-${DC}_1 consul-dev "$@"
+  docker run -i --rm --network container:envoy_consul-${DC}_1 consul:local "$@" 2> /dev/null
 }
 
 function docker_wget {
@@ -581,7 +581,7 @@ function docker_wget {
 function docker_curl {
   local DC=$1
   shift 1
-  docker run --rm --network container:envoy_consul-${DC}_1 --entrypoint curl consul-dev "$@"
+  docker run --rm --network container:envoy_consul-${DC}_1 --entrypoint curl consul:local "$@"
 }
 
 function docker_exec {
@@ -806,7 +806,14 @@ function delete_config_entry {
 
 function register_services {
   local DC=${1:-primary}
+  wait_for_leader "$DC"
   docker_consul_exec ${DC} sh -c "consul services register /workdir/${DC}/register/service_*.hcl"
+}
+
+# wait_for_leader waits until a leader is elected.
+# Its first argument must be the datacenter name.
+function wait_for_leader {
+  retry_default docker_consul_exec "$1" sh -c '[[ $(curl --fail -sS http://127.0.0.1:8500/v1/status/leader) ]]'
 }
 
 function setup_upsert_l4_intention {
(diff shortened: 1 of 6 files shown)


test/integration/connect/envoy/helpers.bash
```diff
@@ -569,7 +569,7 @@ function docker_consul_for_proxy_bootstrap {
   local DC=$1
   shift 1
 
-  docker run -i --rm --network container:envoy_consul-${DC}_1 consul-dev "$@"
+  docker run -i --rm --network container:envoy_consul-${DC}_1 consul-dev "$@" 2> /dev/null
```
lkysow, line 572: This isn't needed specifically for this change but when I was trying to use arm images I noticed that any docker run errors/warnings are captured along with the `-bootstrap` output which malforms the json.

test/integration/connect/envoy/helpers.bash
```diff
@@ -806,9 +806,16 @@ function delete_config_entry {
 
 function register_services {
```
lkysow, line 807: This kept failing when running locally without the wait_for_leader

test/integration/connect/envoy/run-tests.sh
```diff
@@ -44,17 +46,19 @@ function network_snippet {
...
-  # The Lambda integration cases assume that a Lambda function exists in $AWS_REGION with an ARN of $AWS_LAMBDA_ARN.
-  # The AWS credentials must have permission to invoke the Lambda function.
-  [ -n "$(set | grep '^AWS_ACCESS_KEY_ID=')" ] && snippet="${snippet} -e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID"
-  [ -n "$(set | grep '^AWS_SECRET_ACCESS_KEY=')" ] && snippet="${snippet} -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY"
-  [ -n "$(set | grep '^AWS_SESSION_TOKEN=')" ] && snippet="${snippet} -e AWS_SESSION_TOKEN=$AWS_SESSION_TOKEN"
-  [ -n "$(set | grep '^AWS_LAMBDA_REGION=')" ] && snippet="${snippet} -e AWS_LAMBDA_REGION=$AWS_LAMBDA_REGION"
-  [ -n "$(set | grep '^AWS_LAMBDA_ARN=')" ] && snippet="${snippet} -e AWS_LAMBDA_ARN=$AWS_LAMBDA_ARN"
-
-  echo "$snippet"
+  if [ "$LAMBDA_TESTS_ENABLED" = true ]; then
```
lkysow, line 49: @cthain I added this to be able to run the tests locally. The env var gets set in CI to trigger the lambda tests but by default we don't check these env vars.

test/integration/connect/envoy/README.me
```diff
@@ -0,0 +1,14 @@
...
+
+These tests validate Envoy configuration. They set up various scenarios using Docker containers and then run
+[Bats](https://github.com/sstephenson/bats) tests (a Bash test framework) to validate the expected results.
+
+## Running Tests
+
+To run the tests locally, `cd` into the root of the repo and run:
+
+```console
+make
```
DanStough, line 13:
```suggestion
make test-envoy-integ GO_TEST_FLAGS="-run TestEnvoy/<test-case>"
```

DanStough (approved):
LGTM! Just one comment on the README. Thank you for fixing these! 🙏 👏

test/integration/connect/envoy/README.me
lkysow: thanks sorry I meant to include this file in a separate PR.

PR #14369, opened 2022-08-26
docs: Cluster peering with Transparent Proxy updates 

david-yu:
### Description

Numerous changes to document flow for using cluster peering with TProxy. This update now provides a working example for using Cluster Peering with Transparent Proxy. 

### Links

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/website/content/docs/connect/cluster-peering/k8s.mdx
+++ b/website/content/docs/connect/cluster-peering/k8s.mdx
@@ -27,7 +27,7 @@ You must implement the following requirements to create and use cluster peering 
 
 ### Pre-setup 
 
-1. After provisioning a Kubernetes cluster and setting up your kubeconfig file to manage access to multiple Kubernetes clusters, export the Kubernetes context names for future with `kubectl`. 
+1. After provisioning a Kubernetes cluster and setting up your kubeconfig file to manage access to multiple Kubernetes clusters, export the Kubernetes context names for future use with `kubectl`. 
 
 You can use the following methods to get the context names for your clusters:
 * Issue the `kubectl config current-context` command to get the context for the cluster you are currently in.
--- a/website/content/docs/connect/cluster-peering/k8s.mdx
+++ b/website/content/docs/connect/cluster-peering/k8s.mdx
@@ -30,8 +30,9 @@ You must implement the following requirements to create and use cluster peering 
 1. After provisioning a Kubernetes cluster and setting up your kubeconfig file to manage access to multiple Kubernetes clusters, export the Kubernetes context names for future use with `kubectl`. 
 
   You can use the following methods to get the context names for your clusters:
-  * Issue the `kubectl config current-context` command to get the context for the cluster you are currently in.
-  * Issue the `kubectl config get-contexts` command to get all configured contexts in your kubeconfig file.  
+  
+    * Issue the `kubectl config current-context` command to get the context for the cluster you are currently in.
+    * Issue the `kubectl config get-contexts` command to get all configured contexts in your kubeconfig file.  
 
   ```shell-session
   $ export CLUSTER1_CONTEXT=<CONTEXT for first Kubernetes cluster>
--- a/website/content/docs/connect/cluster-peering/k8s.mdx
+++ b/website/content/docs/connect/cluster-peering/k8s.mdx
@@ -276,7 +276,7 @@ To peer Kubernetes clusters running Consul, you need to create a peering token a
     $ kubectl --context $CLUSTER2_CONTEXT apply --filename intention.yml
     ```
 
-1. For the services in `cluster-01` that you want to access the "backend-service," add the following annotations to the service file. In order dial the upstream service from an application, ensure that the requests are sent to the correct DNS name as specified in [Service Virtual IP Lookups](https://www.consul.io/docs/discovery/dns#service-virtual-ip-lookups). 
+1. For the services in `cluster-01` that you want to access the "backend-service," add the following annotations to the service file. To dial the upstream service from an application, ensure that the requests are sent to the correct DNS name as specified in [Service Virtual IP Lookups](https://www.consul.io/docs/discovery/dns#service-virtual-ip-lookups). 
 
     <CodeBlockConfig filename="frontend-service.yaml">
 
--- a/website/content/docs/connect/cluster-peering/k8s.mdx
+++ b/website/content/docs/connect/cluster-peering/k8s.mdx
@@ -25,7 +25,7 @@ You must implement the following requirements to create and use cluster peering 
 - At least two Kubernetes clusters
 - The installation must be running on Consul on Kubernetes version 0.47.1 or later
 
-###  Preparation for install
+###  Prepare for install
 
 1. After provisioning a Kubernetes cluster and setting up your kubeconfig file to manage access to multiple Kubernetes clusters, export the Kubernetes context names for future use with `kubectl`. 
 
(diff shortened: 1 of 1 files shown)


website/content/docs/connect/cluster-peering/k8s.mdx
```diff
@@ -25,9 +25,26 @@ You must implement the following requirements to create and use cluster peering
 - At least two Kubernetes clusters
 - The installation must be running on Consul on Kubernetes version 0.47.1 or later
 
+### Pre-setup 
+
+1. After provisioning a Kubernetes cluster and setting up your kubeconfig file to manage access to multiple Kubernetes clusters, export the Kubernetes context names for future with `kubectl`. 
+
```
trujillo-adam, line 31:
```suggestion
### Pre-setup 

1. After provisioning a Kubernetes cluster and setting up your kubeconfig file to manage access to multiple Kubernetes clusters, export the Kubernetes context names for future with `kubectl`. 

```
First, is this a prerequisite or an additional step to the overall workflow? Calling it a pre-step (IMO) is little confusing. To keep it simple, this is either a prerequisite or the first step in the procedure.

Also, "After provisioning a Kubernetes cluster" tracks with the second bullet in the prerequisites section ("At least two Kubernetes clusters"), but "[After]. . . setting up your kubeconfig file to manage access to multiple Kubernetes clusters" is new. I think we should make this step it's own action either as part of the prerequisites or as the first (second?) step in the overall procedure.

Does that make sense?

website/content/docs/connect/cluster-peering/k8s.mdx
```diff
@@ -25,9 +25,26 @@ You must implement the following requirements to create and use cluster peering
...
+  You can discover the context names using either `kubectl config current-context` for the current cluster you are on, or by issuing `kubectl config get-contexts` to get all configured contexts for your kubeconfig file.
```
trujillo-adam, line 32:
```suggestion
You can use the following methods to get the context names for your clusters:
* Issue the `kubectl config current-context` command to get the context for the cluster you are currently in.
* Issue the `kubectl config get-contexts` command to get all configured contexts in your kubeconfig file.  
```
Is that any cleaner?

website/content/docs/connect/cluster-peering/k8s.mdx
```diff
@@ -25,9 +25,26 @@ You must implement the following requirements to create and use cluster peering
...
+  ```
+
 ### Helm chart configuration
 
-To establish cluster peering through Kubernetes, deploy clusters with the following Helm values.
+1. To establish cluster peering through Kubernetes, create a values.yaml file with the following Helm values. 
+   
+  With these values,
+  the servers in each cluster will be exposed over a Kubernetes Load balancer service. This service can be customized
+  using [`server.exposeService`](/docs/k8s/helm#v-server-exposeservice).
```
trujillo-adam, line 45:
```suggestion
  These values expose the servers in each cluster will over a Kubernetes Load balancer service. You can configure the [`server.exposeService`](/docs/k8s/helm#v-server-exposeservice) value to customize the service.
```
The more I think about, the more I think that we should position the content under "Pre-setup" and "Helm chart configuration" as the first step in the procedure with a heading like "Prepare your Kubernetes environment". We would just need to indicate that the practitioner only needs to perform this part of the procedure once (right?).

website/content/docs/connect/cluster-peering/k8s.mdx
```diff
@@ -25,9 +25,26 @@ You must implement the following requirements to create and use cluster peering
...
 ### Helm chart configuration
 
-To establish cluster peering through Kubernetes, deploy clusters with the following Helm values.
+1. To establish cluster peering through Kubernetes, create a values.yaml file with the following Helm values. 
+   
+  With these values,
+  the servers in each cluster will be exposed over a Kubernetes Load balancer service. This service can be customized
+  using [`server.exposeService`](/docs/k8s/helm#v-server-exposeservice). 
+
+  When generating a peering token from one of the clusters, the address(es) of the load balancer will be used in the peering token, so the peering stream will go through the load balancer in front of the servers. For customizing the addresses used in the peering token, see [`global.peering.tokenGeneration`](/docs/k8s/helm#v-global-peering-tokengeneration).
```
trujillo-adam, line 47:
```suggestion
  When generating a peering token from one of the clusters, Consul uses the address(es) of the load balancer in the peering token so that the peering stream goes through the load balancer in front of the servers. For customizing the addresses used in the peering token, refer to [`global.peering.tokenGeneration`](/docs/k8s/helm#v-global-peering-tokengeneration).
```
Switched to present tense and active voice per the style guide. Also, do we mean "While generating a peering token . . ." in the first sentence? IF the operation can happen more than once, then we should say "While . . .", but if it is a one-time operation, then "When . . ." is OK.

website/content/docs/connect/cluster-peering/k8s.mdx
```diff
@@ -195,37 +271,117 @@ To peer Kubernetes clusters running Consul, you need to create a peering token a
 1. Apply the intentions to the second cluster.
 
     ```shell-session
-    $ kubectl apply --filename intention.yml
+    $ kubectl --context $CLUSTER2_CONTEXT apply --filename intention.yml
     ```
 
-1. For the services in `cluster-01` that you want to access the "backend-service," add the following annotations to the service file.
+1. For the services in `cluster-01` that you want to access the "backend-service," add the following annotations to the service file. In order dial the upstream service from an application, ensure that the requests are sent to the correct DNS name as specified in [Service Virtual IP Lookups](https://www.consul.io/docs/discovery/dns#service-virtual-ip-lookups).
```
trujillo-adam, line 277:
```suggestion
1. For the services in `cluster-01` that you want to access the "backend-service," add the following annotations to the service file. To dial the upstream service from an application, ensure that the requests are sent to the correct DNS name as specified in [Service Virtual IP Lookups](https://www.consul.io/docs/discovery/dns#service-virtual-ip-lookups). 
```

trujillo-adam (commented):
This is a great addition  -- just have a few questions and suggestions.

website/content/docs/connect/cluster-peering/k8s.mdx
david-yu: It is not an explicit pre-req but makes the overall process of issuing commands a lot easier if you are working with multiple clusters.

website/content/docs/connect/cluster-peering/k8s.mdx
david-yu: I've added preparation for install as I believe that is the better way to describe the retrieval of kube contexts and creation of the values file under one heading.

website/content/docs/connect/cluster-peering/k8s.mdx
```diff
@@ -25,9 +25,27 @@ You must implement the following requirements to create and use cluster peering
 - At least two Kubernetes clusters
 - The installation must be running on Consul on Kubernetes version 0.47.1 or later
 
-### Helm chart configuration
+###  Preparation for install
 
-To establish cluster peering through Kubernetes, deploy clusters with the following Helm values.
+1. After provisioning a Kubernetes cluster and setting up your kubeconfig file to manage access to multiple Kubernetes clusters, export the Kubernetes context names for future use with `kubectl`.
```
trujillo-adam, line 30:
```suggestion
1. After provisioning a Kubernetes cluster and setting up your kubeconfig file to manage access to multiple Kubernetes clusters, export the Kubernetes context names for future use with `kubectl`. 
```
Should "Your kubeconfig file should be set up to manage access to multiple Kubernetes clusters" be one of the bullets under the main Prerequisites section? It is a new step that we are assuming the user already did by this point.

trujillo-adam (approved):
Just a few minor suggestions. I'll go ahead and approve so that you're unblocked.

website/content/docs/connect/cluster-peering/k8s.mdx
david-yu: I think its ok to leave this to the user but I'll add a link to the k8s docs to make this more explicit and reduce confusion.

PR #14371, opened 2022-08-26
Adjust metrics reporting for peering tracker

kisunji:
Follow-up to https://github.com/hashicorp/consul/pull/14004

### Description
https://github.com/hashicorp/consul/pull/14004 introduced a way of computing the health of a peering but I think the implementation could be simplified:

#### 1. Deriving health from heartbeats is redundant (see edit notes)
If heartbeats time out, the stream is disconnected:
https://github.com/hashicorp/consul/blob/64d6b46f75d6c37379a3cfd10b791d643c371f85/agent/grpc-external/services/peerstream/stream_resources.go#L457-L458

Then the error gets tracked:
https://github.com/hashicorp/consul/blob/64d6b46f75d6c37379a3cfd10b791d643c371f85/agent/grpc-external/services/peerstream/stream_resources.go#L277-L280

And eventually the leader retries the connection.

So the heartbeat health (among other classes of transient errors) can simply be inferred from the DisconnectTime. If the leader manages to connect to the peer again, the disconnect time is cleared and the status will be marked healthy.

*EDIT*: I chatted with Matt and we agreed that there should be some grace period for peers to recover before we mark peerings as unhealthy. I am preserving https://github.com/hashicorp/consul/pull/14004 usage of heartbeat timeouts as the deadline for peers to recover.

#### 2. ACK and NACK times should be used to detect "sending health"
https://github.com/hashicorp/consul/pull/14004 introduced `LastSendSuccess` as a data point and compared it with `LastSendError` to measure health. Unfortunately, our current handling of `LastSendError` is not always correct. This means non-send errors can be incorrectly tracked as send errors and make our health reporting inaccurate.

Example:
https://github.com/hashicorp/consul/blob/6ddcc046136d4e5a6a5f05fb59afdac0e070f7d3/agent/consul/leader_peering.go#L379-L386

We should instead use `LastAck` and `LastNack`, which indicate whether the sent messages were processed correctly or not.

If a server were to consistently fail at sending, the stream would be in a disconnected-but-retrying state captured by `DisconnectTime`

#### 3. We should consider "receiving health"
If a server fails to successfully process a resource replicated from the peer, it should also mark the peering as unhealthy.

* Updated unit tests

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/consul/server.go
+++ b/agent/consul/server.go
@@ -742,7 +742,6 @@ func NewServer(config *Config, flat Deps, externalGRPCServer *grpc.Server) (*Ser
 			return s.ForwardGRPC(s.grpcConnPool, info, fn)
 		},
 	})
-	s.peerStreamTracker.SetHeartbeatTimeout(s.peerStreamServer.Config.IncomingHeartbeatTimeout)
 	s.peerStreamServer.Register(s.externalGRPCServer)
 
 	// Initialize internal gRPC server.
@@ -1575,12 +1574,12 @@ func (s *Server) Stats() map[string]map[string]string {
 // GetLANCoordinate returns the coordinate of the node in the LAN gossip
 // pool.
 //
-// - Clients return a single coordinate for the single gossip pool they are
-//   in (default, segment, or partition).
+//   - Clients return a single coordinate for the single gossip pool they are
+//     in (default, segment, or partition).
 //
-// - Servers return one coordinate for their canonical gossip pool (i.e.
-//   default partition/segment) and one per segment they are also ancillary
-//   members of.
+//   - Servers return one coordinate for their canonical gossip pool (i.e.
+//     default partition/segment) and one per segment they are also ancillary
+//     members of.
 //
 // NOTE: servers do not emit coordinates for partitioned gossip pools they
 // are ancillary members of.
--- a/agent/consul/server.go
+++ b/agent/consul/server.go
@@ -370,9 +370,9 @@ type Server struct {
 
 	// peerStreamServer is a server used to handle peering streams from external clusters.
 	peerStreamServer *peerstream.Server
+
 	// peeringServer handles peering RPC requests internal to this cluster, like generating peering tokens.
-	peeringServer     *peering.Server
-	peerStreamTracker *peerstream.Tracker
+	peeringServer *peering.Server
 
 	// embedded struct to hold all the enterprise specific data
 	EnterpriseServer
@@ -724,11 +724,9 @@ func NewServer(config *Config, flat Deps, externalGRPCServer *grpc.Server) (*Ser
 		Logger:      logger.Named("grpc-api.server-discovery"),
 	}).Register(s.externalGRPCServer)
 
-	s.peerStreamTracker = peerstream.NewTracker()
 	s.peeringBackend = NewPeeringBackend(s)
 	s.peerStreamServer = peerstream.NewServer(peerstream.Config{
 		Backend:        s.peeringBackend,
-		Tracker:        s.peerStreamTracker,
 		GetStore:       func() peerstream.StateStore { return s.FSM().State() },
 		Logger:         logger.Named("grpc-api.peerstream"),
 		ACLResolver:    s.ACLResolver,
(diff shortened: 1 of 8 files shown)


agent/grpc-external/services/peerstream/server.go
```diff
@@ -42,8 +42,8 @@ type Config struct {
 	// outgoingHeartbeatInterval is how often we send a heartbeat.
 	outgoingHeartbeatInterval time.Duration
 
-	// IncomingHeartbeatTimeout is how long we'll wait between receiving heartbeats before we close the connection.
-	IncomingHeartbeatTimeout time.Duration
+	// incomingHeartbeatTimeout is how long we'll wait between receiving heartbeats before we close the connection.
+	incomingHeartbeatTimeout time.Duration
```
kisunji, line 46: reverting the export of this field since we no longer need it outside the package

agent/grpc-external/services/peerstream/stream_resources.go
```diff
@@ -642,7 +643,6 @@ func (s *Server) realHandleStream(streamReq HandleStreamRequest) error {
 			if err := streamSend(replResp); err != nil {
 				return fmt.Errorf("failed to push data for %q: %w", update.CorrelationID, err)
 			}
-			status.TrackSendSuccess()
```
kisunji, line 645: No need to track send success; the peer will respond with Ack or Nack.

agent/grpc-external/services/peerstream/stream_tracker.go
```diff
@@ -216,38 +203,40 @@ func (s *Status) GetExportedServicesCount() uint64 {
 
 // IsHealthy is a convenience func that returns true/ false for a peering status.
 // We define a peering as unhealthy if its status satisfies one of the following:
-// -  If heartbeat hasn't been received within the IncomingHeartbeatTimeout
-// -  If the last sent error is newer than last sent success
+// -  If it is disconnected
```
mkeeler, line 206: If we go based on just disconnection then during every peer server upgrade or leadership transition there will be a period where we would report the peering as unhealthy.

This seems a little alaramist to me as these are normal operations. As long as we don't remain disconnected for too long (have received a heartbeat recently enough) we can assume the relationship is healthy.

agent/grpc-external/services/peerstream/stream_resources.go
kisunji: Send errors aside from heartbeats will cause the stream to disconnect and try the next address in the retry loop, so it will be caught in the "disconnected" case if the sends repeatedly fail

agent/grpc-external/services/peerstream/stream_tracker.go
kisunji: Both options seem reasonable. I'll go with the "received a heartbeat recently enough" for now unless there are other opinions.

agent/consul/server.go
```diff
@@ -370,9 +370,9 @@ type Server struct {
 
 	// peerStreamServer is a server used to handle peering streams from external clusters.
 	peerStreamServer *peerstream.Server
+
 	// peeringServer handles peering RPC requests internal to this cluster, like generating peering tokens.
-	peeringServer     *peering.Server
-	peerStreamTracker *peerstream.Tracker
```
kisunji, line 375: `peerStreamTracker` existed in a weird circular reference with `peerStreamServer`. I removed it from `Server` and it is now contained within `peerStreamServer` which is a narrower and more appropriate context for it to live in.

agent/grpc-external/services/peerstream/server.go
```diff
@@ -26,11 +26,12 @@ const (
 
 type Server struct {
 	Config
+
+	Tracker *Tracker
```
kisunji, line 30: Our `Config` is overloaded with what should be called dependencies; I moved Tracker out of it.

agent/grpc-external/services/peerstream/stream_tracker.go
```diff
@@ -136,6 +137,39 @@ func (t *Tracker) DeleteStatus(id string) {
...
 
+// IsHealthy is a calculates the health of a peering status.
+// We define a peering as unhealthy if its status has been in the following
+// states for longer than the configured incomingHeartbeatTimeout.
+//   - If it is disconnected
+//   - If the last received Nack is newer than last received Ack
+//   - If the last received error is newer than last received success
+//
+// If none of these conditions apply, we call the peering healthy.
+func (t *Tracker) IsHealthy(s Status) bool {
```
kisunji, line 148: I moved `IsHealthy` to be a method under `Tracker` because some of the calculations require dependencies which didn't make sense to live in a data object. If we consider `Status` as a bundle of data points, it should be the `Tracker`'s responsibility to interpret it as healthy or not.

mkeeler (approved):

PR #14373, opened 2022-08-29
Update max_ejection_percent on outlier detection for peered clusters to 100%

erichaberkorn:
### Description
We can't trust health checks when service resolvers, splitters, and routers are used for peered clusters.

### PR Checklist
* [X] updated test coverage
* [X] not a security concern

### DIFF
--- a/agent/xds/clusters.go
+++ b/agent/xds/clusters.go
@@ -772,6 +772,13 @@ func (s *ResourceGenerator) makeUpstreamClusterForPeerService(
 
 	clusterName := generatePeeredClusterName(uid, tbs)
 
+	outlierDetection := ToOutlierDetection(cfg.PassiveHealthCheck)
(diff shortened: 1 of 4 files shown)


kisunji (approved):

agent/xds/clusters.go
```diff
@@ -772,6 +772,9 @@ func (s *ResourceGenerator) makeUpstreamClusterForPeerService(
 
 	clusterName := generatePeeredClusterName(uid, tbs)
 
+	outlierDetection := ToOutlierDetection(cfg.PassiveHealthCheck)
+	outlierDetection.MaxEjectionPercent = &wrappers.UInt32Value{Value: 100}
```
kisunji, line 776: Could we add a developer note explaining why?

If I were to trace this commit it would end up on this PR which doesn't have much context either

kisunji (approved):

PR #14387, opened 2022-08-29
Backport of Allow terminated peerings to be deleted into release/1.13.x

hc-github-team-consul-core:
## Backport

This PR is auto-generated from #14364 to be assessed for backporting due to the inclusion of the label backport/1.13.

**WARNING** automatic cherry-pick of commits failed. Commits will require human attention.

---

### Description
Peerings are terminated when a peer decides to delete the peering from
their end. Deleting a peering sends a termination message to the peer
and triggers them to mark the peering as terminated but does NOT delete
the peering itself. This is to prevent peerings from disappearing from
both sides just because one side deleted them.

Previously the Delete endpoint was skipping the deletion if the peering
was not marked as active. However, terminated peerings are also
inactive.

This PR makes some updates so that peerings marked as terminated can be
deleted by users.

* Unit tests at state store and gRPC service layers

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


freddygv:
Replaced by #14390

PR #14394, opened 2022-08-30
ui: Setup Hashicorp Design System for usage in consul-ui

LevelbossMike:
### Description
This PR adds a setup that makes it possible to use [HashiCorp Design System (HDS)](https://github.com/hashicorp/design-system) in consul-ui.

This PR adds several packages that in combination allows the application to use the entirety of HDS:

* [@hashicorp/design-system-tokens](https://github.com/hashicorp/design-system/tree/main/packages/tokens) - a package that makes design tokens available as custom CSS properties
* [@hashicorp/design-system-components](https://github.com/hashicorp/design-system/tree/main/packages/components) - an ember add-on that bundles ready to use components that
* [@hashicorp/ember-flight-icons](https://github.com/hashicorp/design-system/tree/main/packages/ember-flight-icons) - an ember add-on that bundles flight icons

## Wait what? Why tailwindcss?
The PR in addition to HDS sets up [tailwindcss](https://tailwindcss.com/). We set up tailwindcss because tailwind is built with customization in mind and allows us to use the design system tokens without verbose manual usage of CSS properties or synchronizing CSS files manually. Instead, we use [tailwind's configuration](https://tailwindcss.com/docs/configuration) feature to auto-generate utility-classes for us, that make it easy to work with colors from HDS. All color utilities are prefixed with `hds-` to make it easy to see when scanning markup what utilities refer to HDS colors. The mechanism could be extended for other design tokens like elevation as well, but HDS already bundles utility classes for things like typography and elevation, so there's really not much for us to add.

### Tailwindcss - utility classes for everything — colors and layout mainly
Tailwind comes with [plenty of utility classes](https://tailwindcss.com/docs/utility-first) out of the box that make it easy to build up UIs quickly. It's easy to add hover, focus, and other state handling with it, supports animations and comes with responsive utilities. This utility approach mixes well with HDS and allows for quick iteration without writing any custom CSS code. The alternative to using tailwindcss would be to come up with all these spacing, flexbox etc. utility classes yourself or write custom CSS. This is obviously much more time-consuming than using what tailwind gives you out of the box for free and can be, as demonstrated, customized based on the design system easily.  I added an example to the engineering docs that showcases how HDS and tailwindcss can be used in combination to build up a navigation header:

```hbs
<nav
  class="bg-hds-foreground-strong hds-font-weight-medium flex h-16 w-full items-center justify-between px-4"
>
  <ul class="flex items-center">
    <li>
      {{! should probably be a context-switcher }}
      <FlightIcon
        @name="menu"
        class="fill-current text-hds-neutral-200 cursor-pointer"
      />
    </li>
    <li>
      <FlightIcon
        @name="consul"
        class="fill-current text-hds-consul-brand ml-4 h-8 w-8"
      />
    </li>
  </ul>
  <ul class="flex items-center">
    <li>
      <Hds::Dropdown as |dd|>
        <dd.ToggleButton
          @text="Help"
          @color="secondary"
          class="text-hds-neutral-300"
        />
        <dd.Title @text="Consul HDS" />
        <dd.Interactive @text="Documentation" @icon="docs-link" />
        <dd.Interactive @text="HashiCorp Learn" @icon="learn-link" />
        <dd.Separator />
        <dd.Interactive @text="Provide Feedback" @icon="github" />
      </Hds::Dropdown>
    </li>
    <li>
      <Hds::Button
        @text="Settings"
        @color="tertiary"
        @icon="settings"
        @iconPosition="trailing"
        class="text-hds-neutral-300"
      />
    </li>
  </ul>
</nav>
```

The above code produces the following:

Screenshot 2022-08-30 at 13 53 25

### This will bloat the build, right?
No. tailwindcss comes with a JIT compiler that will only bundle the classes you are actually using. We really get all of this functionality for free.

## Important changes

To achieve the described functionality, this PR adds several changes to the build:

1. We switch from using [ember-cli-sass](https://github.com/adopted-ember-addons/ember-cli-sass) to using [ember-cli-postcss](https://github.com/jeffjewiss/ember-cli-postcss) instead. We don't lose anything by switching over to postcss, our current SCSS codebase still works as it did before. Building up a postcss pipeline allows us to bundle tailwind and will also enable other optimizations in the future, like for example minifying CSS via [css-nano](https://github.com/cssnano/cssnano) or similar.
2. I had to update the build to bundle up style files from sub-app-packages. This is because sub-apps aren't implemented as ember add-ons, and will thus not bundle their style-files into the `styles`-tree in the application build. This worked with `ember-cli-sass` because it apparently operates on the entire application tree, but won't work with `ember-cli-postcss` as it is scoped to the styles tree.
3. We are switching to `ember-auto-import`-v2 in this PR. This is necessary as @hashicorp/design-system-components is a v2 add-on and depends on v2.
4. We are upgrading [ember-cli-yadda](https://github.com/kaliber5/ember-cli-yadda) because only its latest version is compatible with `ember-auto-import` v2.

### Links
I added [a section about HDS in the engineering docs](http://localhost:4200/ui/docs/hds) that goes into detail of how to use HDS in the application. You have to boot consul-ui to see the docs.

### Things to do before merging

- [ ] Check CSS size after broccoli changes — I don't think I did anything that would increase the build-size significantly by changing the way we handle the `styles`-tree for sub-apps, but want to make sure before this gets merged.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


LevelbossMike:
The build is failing due to [ember-cli-api-double](https://github.com/hashicorp/ember-cli-api-double) pulling in babel 6. 

I'm still investigating how to fix this. I upgraded ember-cli-api-double to the latest ember-cli in a fork of mine, which seems to work fine, but then it still breaks on the testing setup because of including Qunit multiple. Probably related to the upgrade of ember-auto-import. I will upgrade to a more up-to-date version of an ember-cli update to try to fix this.

PR #14397, opened 2022-08-30
xDS Load Balancing

boxofrad:
### Description
Prior to #13244, connect proxies and gateways could only be configured by an xDS session served by the local client agent. 

In an upcoming release, it will be possible to deploy a Consul service mesh without client agents. In this model, xDS sessions will be handled by the servers themselves, which necessitates load-balancing to prevent a single server from receiving a disproportionate amount of load and becoming overwhelmed.

This PR introduces a simple form of load-balancing where Consul will attempt to achieve an even spread of load (xDS sessions) between all healthy servers. It does so by implementing a concurrent session limiter (`limiter.SessionLimiter`) and adjusting the limit according to autopilot state and proxy service registrations in the catalog.

If a server is already over capacity (i.e. the session limit is lowered), Consul will begin draining sessions to rebalance the load. This will result in the client receiving a `RESOURCE_EXHAUSTED` status code. It is the client's responsibility to observe this response and reconnect to a different server. Users of the gRPC client connection brokered by the [consul-server-connection-manager](https://github.com/hashicorp/consul-server-connection-manager) library will get this for free.

The rate at which Consul will drain sessions to rebalance load is scaled dynamically based on the number of proxies in the catalog:

| Number of proxies | Drain rate |
| - | - |
|0-512|1 session per second|
|513-2815|linearly scaled by 1/second for every additional 256 proxies|
|2816+|10 sessions per second|

### Testing Steps
- Start a Consul dev server (`consul agent -dev`)
- Register 10 proxy services using the HTTP API
- Start Envoy proxies for them (tip: folks on the Consul team can use the `consul-dataplane` binary for this!)
- Join another server to the cluster (`consul agent -dev -retry-join "127.0.0.1:8301" -server-port 7300 -serf-wan-port -1 -serf-lan-port 7301 -dns-port -1 -http-port -1 -grpc-port -1 -node "other-server"`)
- You should see ~4 proxies receive `RESOURCE_EXHAUSTED` errors as the server tries to rebalance load.


website/content/docs/agent/telemetry.mdx
```diff
@@ -542,6 +542,8 @@ These metrics are used to monitor the health of the Consul servers.
...
+| `consul.xds.server.sessionDrained`                  | Counts the number of xDS sessions that are drained/terminated to rebalance load between servers.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   | sessions                          | counter |
```
boxofrad, line 546: I went with the `xds.server` prefix here for consistency with the existing metric rather than the RFC. We also currently refer to xDS sessions as streams here, so I wondered whether we should adopt this in the new metrics for consistency too?

website/content/docs/agent/config/config-files.mdx
```diff
@@ -2198,3 +2198,16 @@ tls {
...
+- `xds` This is a nested object that configures how Consul serves the [xDS
+  protocol](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol)
+  to Envoy and consul-dataplane.
+  - `rebalance_drain_rate` Configures the rate at which xDS sessions will be
+    drained to rebalance load between servers, in terminations per second.
+    Defaults to 10 per second.
+  - `rebalance_drain_burst` The size of the token bucket used to recharge the
+    xDS session drain rate limiter. Defaults to 50 tokens.
+    See https://en.wikipedia.org/wiki/Token_bucket for more details about how
+    token bucket rate limiters operate.
```
boxofrad, line 2213: I'm not particularly happy with these docs. I think we could do a better job of explaining how to choose the right bucket size vs rate.

agent/config/runtime.go
```diff
@@ -1414,6 +1414,16 @@ type RuntimeConfig struct {
...
 
+	// XDSRebalanceDrainRateLimit and XDSRebalanceDrainMaxBurst control the rate
+	// at which xDS sessions will be terminated to rebalance load between servers,
+	// using a token bucket rate limiter.
+	//
+	// See https://en.wikipedia.org/wiki/Token_bucket for more about token buckets.
+	//
+	// hcl: xds { rebalance_drain_rate = (float64|MaxFloat64) rebalance_drain_burst = int }
+	XDSRebalanceDrainRateLimit rate.Limit
+	XDSRebalanceDrainMaxBurst  int
```
boxofrad, line 1425: Having talked this through a bit on Slack, we're going to remove these user-facing options and scale the rate limit dynamically based on number of proxies.

website/content/docs/agent/telemetry.mdx
```diff
@@ -542,6 +542,8 @@ These metrics are used to monitor the health of the Consul servers.
...
+| `consul.xds.server.idealSessionsMax`                | The ideal maximum number of xDS sessions per server.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               | sessions                          | gauge   |
```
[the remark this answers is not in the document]
boxofrad: Great point! It's the server's opinion of the number of xDS sessions it should handle to achieve an even spread of load across the cluster.

It is used as a hard limit, but might take a while to be enforced because the server will slowly drain excess sessions, to avoid causing major disruption by draining them all at once.

I think the reason we chose to include "ideal" in the metric name is to prevent folks from getting confused by seeing that a server's "maximum" number of sessions is lower than its actual number of sessions.

Do you think we should drop "ideal" from the name too? Or maybe have a clearer name?

website/content/docs/agent/telemetry.mdx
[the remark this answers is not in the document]
boxofrad: 1. Yeah sorry, that's a bad/unfortunate habit of mine 😉 let's commit to the word "drain" (and drop "terminated") because it's more common to speak of "connection draining" when talking about load balancers.
2. That's a really good point. I meant to say that "sessions are drained in order to rebalance load between servers" but I think "when rebalancing" is clearer 👍🏻

boxofrad:
Thanks so much for taking the time to review this @boruszak 🙌🏻 

We decided not to expose the rate-limiter configuration to users, which means that we can drop that whole section of the docs - and I think the feature is better for it.

agent/grpc-external/limiter/limiter.go
```diff
@@ -0,0 +1,245 @@
...
+//
+//	1. Call End on the session when finished.
+//	2. Receive on the session's Terminated channel and exit (e.g. close the gRPC
+//	   stream) when it is closed.
+//
+// The maximum number of concurrent sessions is controlled with SetMaxSessions.
+// If there are more than the given maximum sessions already in-flight,
+// SessionLimiter will drain randomly-selected sessions at a rate controlled
+// by SetDrainRateLimit.
+type SessionLimiter struct {
```
boxofrad, line 38: Note: we refer to xDS sessions as "streams" in most places for consistency with existing metrics etc. but this component is called the "session limiter" because it is generic and can be used to limit any concurrent operations.

agent/grpc-external/limiter/limiter.go
```diff
@@ -0,0 +1,245 @@
...
+	s.l.mu.Lock()
+	defer s.l.mu.Unlock()
+
+	idx := -1
+	for i, id := range s.l.sessionIDs {
+		if id == s.id {
+			idx = i
+			break
+		}
+	}
```
boxofrad, line 230: This linear search of `sessionIDs` isn't great, but because we grow and shrink the slice as sessions are added and removed we can't track the index in the `Session` struct.

Given `n` is bounded by the number of open xDS streams, I don't think it could become large enough to become a problem worth solving with a more complicated data structure, but I'm open to feedback!

boxofrad: Actually thinking about it, because session IDs increase monotonically, we could fairly easily store them sorted and do a binary search instead `O(log n)`.

agent/consul/xdscapacity/capacity.go
```diff
@@ -0,0 +1,205 @@
...
+	for _, csn := range csns {
+		if csn.Service.Kind.IsProxy() {
+			count++
+		}
+	}
+	return ws, count, nil
+}
+
+type Store interface {
+	watch.StateStore
```
boxofrad, line 201: Note to self: we don't actually use the `watch` package here, so should just add the `AbandonCh` method to this interface directly.

agent/grpc-external/limiter/limiter.go
```diff
@@ -0,0 +1,254 @@
...
+	// max and inFlight are read/written using atomic operations.
+	max, inFlight uint32
+
+	// wakeCh is used to trigger the Run loop to start draining excess sessions.
+	wakeCh chan struct{}
+
+	// Everything below here is guarded by mu.
+	mu           sync.Mutex
+	maxSessionID uint64
+	sessionIDs   []uint64 // sessionIDs must be sorted so we binary search it.
```
boxofrad, line 52:
```suggestion
	sessionIDs   []uint64 // sessionIDs must be sorted so can we binary search it.
```

agent/consul/xdscapacity/capacity.go
```diff
@@ -0,0 +1,203 @@
...
+			count++
+		}
+	}
+	return ws, count, nil
+}
+
+type Store interface {
+	AbandonCh() <-chan struct{}
+	ServiceDump(ws memdb.WatchSet, kind structs.ServiceKind, useKind bool, entMeta *acl.EnterpriseMeta, peerName string) (uint64, structs.CheckServiceNodes, error)
+	PeeringList(ws memdb.WatchSet, entMeta acl.EnterpriseMeta) (uint64, []*pbpeering.Peering, error)
```
riddhi89, line 202: I do not see this being used anywhere. Does the interface need it?

riddhi89 (approved):
Beautifully implemented 😍 Great job!

Would be ideal if someone else from the core team also gave this a look, but LGTM 👏

agent/consul/xdscapacity/capacity.go
boxofrad: Good catch! This is a leftover from when I thought we needed to count proxy services in peered datacenters (we don't 😄)

agent/consul/xdscapacity/capacity.go
```diff
@@ -0,0 +1,203 @@
...
+			a.updateMaxSessions(numServers, numProxies)
+		case <-ws.WatchCh(ctx):
+			var count uint32
+			ws, count, err = a.countProxies()
+			if err == nil {
+				numProxies = count
+				a.updateDrainRateLimit(numProxies)
+				a.updateMaxSessions(numServers, numProxies)
+			} else {
+				a.cfg.Logger.Error("failed to count proxy services", "error", err)
```
mkeeler, line 101: ❗ (blocking)

Is there any feasible way to recover here? I think if we hit this case we will never again update our proxy count. Could we do something with exponential backoff where we infinitely retry until counting proxies returns a non-error result?

agent/consul/xdscapacity/capacity.go
```diff
@@ -0,0 +1,203 @@
...
+	)
+	if err != nil {
+		return ws, 0, err
+	}
+	for _, csn := range csns {
+		if csn.Service.Kind.IsProxy() {
+			count++
+		}
+	}
+	return ws, count, nil
```
mkeeler, line 196: 🔔  (non-blocking suggestion)

This should perform quite a bit better when there are many service instances and takes advantage of the fact that we are already tracking counts of services by their kind.

```suggestion
	var count int
	_, usage, err := store.ServiceUsage()
	if err != nil {
		return ws, 0, err
	}

	for kind, kindCount := range usage.ConnectServiceInstance {
		if structs.ServiceKind(kind).IsProxy() {
			count += kindCount
		}
	}
	return ws, count, nil
```

agent/consul/xdscapacity/capacity_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/hashicorp/consul/agent/structs"
+	"github.com/hashicorp/consul/sdk/testutil"
+)
+
+func TestController(t *testing.T) {
+	const index = 123
+
+	store := state.NewStateStore(nil)
+
+	for _, kind := range []structs.ServiceKind{
```
mkeeler, line 24: It might be useful to add a short comment prior to this loop noting the desired outcome of the nested loops (100 proxy services + 25 non-proxy services).

agent/grpc-external/limiter/limiter.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+func (l *SessionLimiter) deleteSessionLocked(idx int) {
+	delete(l.sessions, l.sessionIDs[idx])
+
+	// Note: it's important that we preserve the order here (which most allocation
+	// free deletion tricks don't) because we binary search the slice.
+	l.sessionIDs = append(l.sessionIDs[:idx], l.sessionIDs[idx+1:]...)
+
+	atomic.AddUint32(&l.inFlight, ^uint32(0))
```
mkeeler, line 195: 🎨 (code style / non-blocking)

Is there any reason we couldn't use a signed integer for these counts? Then this could be `atomic.AddInt32(&l.inFlight, -1)` which might be more immediately obvious to future readers what is happening.

agent/grpc-external/limiter/limiter.go
```diff
@@ -0,0 +1,254 @@
...
+
+	idx := sort.Search(len(s.l.sessionIDs), func(i int) bool {
+		return s.l.sessionIDs[i] >= s.id
+	})
+
+	if idx == len(s.l.sessionIDs) || s.l.sessionIDs[idx] != s.id {
+		panic(fmt.Sprintf("could not find session id: %d", s.id))
+	}
+
+	s.l.deleteSessionLocked(idx)
```
mkeeler, line 245: ❗ (blocking)

This feels like it should live in a separate method on the SessionLimiter. What do you think? Both are in the same package/file so it would be an incredibly minor difference. It feels a little off for this types methods to be performing the binary search on the internal sessions of the limiter.

Also there seems to be a race condition around looking up the session here? The scenario I am thinking of is:

* The xDS session is ending and it starts to call the `End()` method on the session.
* At the same time the session limiter chooses that session to be terminated. 
   * First it closes the sessions chan
   * Next it gains the lock and deletes the session from tracking
* Now the `End()` method continues execution. The search fails to find the session and panics.

I think the fix would be to just not panic. If the limiter isn't tracking the session its because its already been delete and thus everything is as it should be.

mkeeler (changes requested):
Great work! I especially thought the xDS session limiter code was very well written and documented.

I have attempted to mark each comment with regards to its severity/blocking status. Please let me know if thats not clear though. I am trying out a few things from here:

https://blog.carbonfive.com/how-to-use-actionable-emojis-in-your-pull-request-reviews/

agent/consul/xdscapacity/capacity.go
boxofrad: Oh nice! I had no idea about this 🙌🏻

agent/grpc-external/limiter/limiter.go
boxofrad: I went with unsigned because they shouldn't ever be negative (so it felt best make it impossible) but I'm happy to be pragmatic about this for readability!

agent/grpc-external/limiter/limiter.go
boxofrad: Yeah I agree, it does feel a bit funky for `session` to be reaching into `SessionLimiter`'s state - will move it to a method there.

Good catch with the race condition. It took me a little while to grok! I think if we were to remove the `panic` we could also remove the whole `CAS(s.done)` thing too.

mkeeler (approved):

PR #14408, opened 2022-08-30
ci: update backport-assistant to pick merge commit

mikemorris:
Refs hashicorp/vault#15571, https://github.com/hashicorp/backport-assistant/issues/37, https://github.com/hashicorp/backport-assistant/pull/40, https://github.com/hashicorp/backport-assistant/issues/26

> Cherry-pick the merge commit instead of the individual commits in the PR. This will be less confusing, as it would eliminate the partial-success case, and may result in fewer failed cherry-picks in the first place.

This actually brings us back [closer to the implementation we had before switching](https://github.com/hashicorp/backport-assistant/issues/25#issuecomment-999895503) over to backport-assistant.

### DIFF
--- a/.github/workflows/backport-assistant.yml
+++ b/.github/workflows/backport-assistant.yml
@@ -16,7 +16,7 @@ jobs:
   backport:
     if: github.event.pull_request.merged
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


DanStough (approved):
LGTM 🚀 . I added the backport labels so this works on older branches as well.

PR #14340, opened 2022-08-25
Add support for configuring Envoys route idle_timeout  

oulman:
### Description

This PR adds support for configuring the Envoy [route level idle_timeout](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-idle-timeout) for service-routers and local_app. For our teams scenario, this will let us extend the timeout for long-running gRPC services so that Envoy doesn't terminate the connection prematurely. 

For service-routers a `IdleTimeout` parameter has been added, and for local_app a proxy config `local_idle_timeout_ms` parameter was created. I followed the [RequestTimeout](https://www.consul.io/docs/connect/config-entries/service-router#requesttimeout) and [local_request_timeout_ms](https://www.consul.io/docs/connect/proxies/envoy#local_request_timeout_ms) definitions closely.

##### service-router IdleTimeout
Created a service-router and configured the `IdleTimeout` to 60s

```
Kind = "service-router"
Name = "waitserver"
Routes = [
  {
    Destination {
      Service = ""
      RequestTimeout = "60m"
      IdleTimeout = "60s"
    }
  }
]
```

Dumped the Envoy config and verified that the `idle_timeout` configuration parameter was set

```
<snip>
        "name": "waitserver",
        "domains": [
         "*"
        ],
        "routes": [
         {
          "match": {
           "prefix": "/"
          },
          "route": {
           "cluster": "waitserver.default.dc1.internal.ba4be18f-a3c7-7e2f-eff4-a25a8a1fce57.consul",
           "timeout": "3600s",
           "idle_timeout": "60s"
          }
         },
<snip>
```

##### local_idle_timeout_ms

Configure `local_idle_timeout_ms` on a service proxy.config

```
service {
  name = "waitserver"
  id = "waitserver"
  address = "10.5.0.20"
  port = 9222

  connect {
    sidecar_service {
      port = 10201

      proxy {
        config {
          local_request_timeout_ms = 0
          local_idle_timeout_ms = 0
        }
      }
      check {
        name = "Connect Envoy Sidecar"
        tcp = "10.5.0.20:10201"
        interval ="10s"
      }
    }
  }
}
```

Dumped the Envoy config and verified that the `idle_timeout` configuration parameter was set on the public_listener route for `local_app`

```
{
     "route_config": {
      "@type": "type.googleapis.com/envoy.config.route.v3.RouteConfiguration",
      "name": "public_listener",
[...]

PR #14418, opened 2022-08-31
Adds idle_timeout envoy proxy local app config knob

johnalotoski:
### Description
We require override of envoy's default `stream_idle_timeout` which we are unable to do via Nomad and escape hatches at this time AFAICT.  This has also been asked for in the community.  See refs below.

### Links
* https://github.com/hashicorp/nomad/issues/14403
* https://discuss.hashicorp.com/t/override-stream-idle-timeout-5-min-default-on-envoy/30175

### PR Checklist

* [ ] updated test coverage
* [X] not a security concern

### DIFF
--- a/.changelog/14403.txt
+++ b/.changelog/14403.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+connect: Add local_idle_timeout_ms to allow configuring the Envoy idle timeout on local_app
+```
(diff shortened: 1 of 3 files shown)


aoskotsky-amplify:
Could someone review this? I also need this. I have some services that are blocked by this because they need more than the default 5 minute timeout in envoy

rddimon:
@johnalotoski @aoskotsky-amplify This is already released in https://github.com/hashicorp/consul/pull/14340

david-yu:
Thanks @rddimon . @johnalotoski This was indeed released in 1.14.2. Will close this PR. Thank you.

PR #14429, opened 2022-08-31
Prune old expired intermediate certs when appending a new one

### LINKED ISSUE #14159: Prune expired intermediate CA certs from stored roots
Consul currently doesn't remove expired intermediates from the stored root CA struct - it only ever appends new ones:
- https://github.com/hashicorp/consul/blob/201d1458c307e9b04fe3d6f22a2ad9d6d1358bdd/agent/consul/leader_connect_ca.go#L538
- https://github.com/hashicorp/consul/blob/201d1458c307e9b04fe3d6f22a2ad9d6d1358bdd/agent/consul/leader_connect_ca.go#L1101

This is currently only handled when a root rotation happens, because the old list of intermediates will get left behind, but when the intermediate cert TTL is significantly lower than the root's, it's possible to have many intermediate certs build up in the list. This can cause long certificate chains to be sent to Envoy proxies, which in turn can cause the TLS handshake to fail due to too many certs in the chain. We should add some logic to prune old expired intermediate certs from the list when a new one is appended, to prevent them from building up indefinitely.

kyhavlov:
This PR adds some CA logic to prune old expired intermediate certs from the active root when a new intermediate is fetched. Since we renew the intermediate 1/3-1/2 of the way through its lifetime, this means there should never be more than a few old intermediates left around the root CA regardless of the configured root cert TTL.

Fixes #14159

### DIFF
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -1100,6 +1100,28 @@ func setLeafSigningCert(caRoot *structs.CARoot, pem string) error {
 
 	caRoot.IntermediateCerts = append(caRoot.IntermediateCerts, pem)
 	caRoot.SigningKeyID = connect.EncodeSigningKeyID(cert.SubjectKeyId)
+	return pruneExpiredIntermediates(caRoot)
+}
+
+// pruneExpiredIntermediates removes expired intermediate certificates
+// from the given CARoot.
+func pruneExpiredIntermediates(caRoot *structs.CARoot) error {
+	var newIntermediates []string
+	now := time.Now()
+	for i, intermediatePEM := range caRoot.IntermediateCerts {
+		cert, err := connect.ParseCert(intermediatePEM)
+		if err != nil {
+			return fmt.Errorf("error parsing leaf signing cert: %w", err)
+		}
+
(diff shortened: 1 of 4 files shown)


agent/consul/leader_connect_ca.go
```diff
@@ -1100,6 +1100,28 @@ func setLeafSigningCert(caRoot *structs.CARoot, pem string) error {
...
+	now := time.Now()
+	for i, intermediatePEM := range caRoot.IntermediateCerts {
+		cert, err := connect.ParseCert(intermediatePEM)
+		if err != nil {
+			return fmt.Errorf("error parsing leaf signing cert: %w", err)
+		}
+
+		// Only keep the intermediate cert if it's still valid, or if it's the most
+		// recently added (and thus the active signing cert).
+		if cert.NotAfter.After(now) || i == len(caRoot.IntermediateCerts) {
```
erichaberkorn, line 1119: To me, it feels like it might be safer to prune the expired certificates prior to appending the new one. This would allow us to get rid of the `i == len(caRoot.IntermediateCerts)` check and eliminates the possibility of a bug causing all certs to be pruned.

erichaberkorn (approved):

PR #14430, opened 2022-08-31
ui: chore upgrade to ember-qunit v5

LevelbossMike:
### Description
Upgrade to ember-qunit v5. This is preparation for updating to 3.24 and https://github.com/hashicorp/consul/pull/14394

### DIFF
--- a/ui/packages/consul-ui/tests/acceptance/dc/services/show-routing.feature
+++ b/ui/packages/consul-ui/tests/acceptance/dc/services/show-routing.feature
@@ -52,6 +52,12 @@ Feature: dc / services / show-routing: Show Routing for Service
       service: service-1
     ---
     And I see routing on the tabs
+    # something weird is going on with this test
+    # without waiting we issue a url reload that
+    # will make the test timeout.
+    # waiting will "fix" this - we should look into
+    # the underlying reason for this soon. This is
+    # only a quick-fix to land ember-qunit v5.
     And pause for 1000
     And I visit the service page for yaml
     ---
--- a/ui/packages/consul-ui/tests/acceptance/dc/services/show-routing.feature
+++ b/ui/packages/consul-ui/tests/acceptance/dc/services/show-routing.feature
@@ -52,7 +52,7 @@ Feature: dc / services / show-routing: Show Routing for Service
       service: service-1
     ---
     And I see routing on the tabs
-    And I wait a tick
+    And pause for 1
     And I visit the service page for yaml
     ---
       dc: dc1
--- a/ui/packages/consul-ui/tests/acceptance/dc/services/show-routing.feature
+++ b/ui/packages/consul-ui/tests/acceptance/dc/services/show-routing.feature
@@ -52,6 +52,7 @@ Feature: dc / services / show-routing: Show Routing for Service
       service: service-1
     ---
     And I see routing on the tabs
+    And I wait a tick
     And I visit the service page for yaml
     ---
       dc: dc1
--- a/ui/packages/consul-ui/tests/acceptance/dc/services/show-routing.feature
+++ b/ui/packages/consul-ui/tests/acceptance/dc/services/show-routing.feature
@@ -52,7 +52,7 @@ Feature: dc / services / show-routing: Show Routing for Service
       service: service-1
     ---
     And I see routing on the tabs
-    And pause for 1
+    And pause for 1000
     And I visit the service page for yaml
     ---
       dc: dc1
--- a/ui/packages/consul-ui/tests/steps/assertions/page.js
+++ b/ui/packages/consul-ui/tests/steps/assertions/page.js
@@ -41,9 +41,6 @@ export default function(scenario, assert, find, currentPage, $) {
       assert[negative ? 'notOk' : 'ok'](res, this.step);
       setTimeout(() => next());
     })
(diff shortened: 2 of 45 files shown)


LevelbossMike:
I can confirm from running tests locally that this PR + https://github.com/hashicorp/ember-cli-api-double/pull/24 makes the tests pass in https://github.com/hashicorp/consul/pull/14394 :tada:

ui/packages/consul-ui/tests/acceptance/dc/services/show-routing.feature
```diff
@@ -52,6 +52,7 @@ Feature: dc / services / show-routing: Show Routing for Service
       service: service-1
     ---
     And I see routing on the tabs
+    And I wait a tick
```
johncowen, line 55: Really these kind of implementation details shouldn't be in the tests themselves

Apart from that FYI: We have a `And pause for $milliseconds` > `And pause for 0`:

https://github.com/hashicorp/consul/blob/095764a482afb9bfa90f619a2937c7d93112dcd4/ui/packages/consul-ui/tests/steps/debug/index.js#L12-L16

Which is in the `debug` folder, as it should only really be used for debugging (unless we specifically want to model behaviour where we wait for so many milliseconds), so this is just FYI, overall we shouldn't have to tell the tests they need to wait for a tick here.

ui/packages/consul-ui/tests/steps/assertions/page.js
```diff
@@ -41,6 +41,9 @@ export default function(scenario, assert, find, currentPage, $) {
       assert[negative ? 'notOk' : 'ok'](res, this.step);
       setTimeout(() => next());
     })
+    .then('I wait a tick', function() {
+      return new Promise(resolve => setTimeout(resolve, 0));
+    })
```
johncowen, line 46: See above comment on not needing this. But also FYI for any future step additions, this isn't asserting anything, and isn't concerned with page objects (the `page` name of the file), so ideally it wouldn't be in this folder/file.

We have organised the steps folder into `assertions`, `interactions`, `debug` and `doubles` (mocking/setting up), so steps are easier to find based on those categories (and filenames within those would also help), and I think we at least mostly have kept to that. Overall, nothing to do here (apart from remove this) so this is just for knowledge sharing for future.

ui/packages/consul-ui/tests/steps/interactions/form.js
```diff
@@ -57,21 +57,23 @@ export default function(scenario, find, fillIn, triggerKeyEvent, currentPage) {
...
                 throw e;
               }
             }
           } else {
-            return fillInElement(prev, name, data[item]);
+            return await fillInElement(obj, name, data[item]);
           }
-        }, obj);
+        }, Promise.resolve());
         await new Promise(resolve => setTimeout(resolve, 0));
```
johncowen, line 77: Guessing we can remove the 'nextTickPromise' on line 77  here now we've upgraded ember-test-helpers? Also, do we need to `return await` instead of just `return` on line 74?

ui/packages/consul-ui/tests/steps/interactions/form.js
```diff
@@ -11,7 +11,7 @@ export default function(scenario, find, fillIn, triggerKeyEvent, currentPage) {
       return page;
     } else {
       const $el = document.querySelector(`[name="${name}"]`);
-      fillIn($el, value);
+      await fillIn($el, value);
       return page;
```
johncowen, line 15: Thanks for adding the async/await here 👏 . I took a little more of a look into this, and it looks like the ember-cli-page-object we were previously using for this was not an async API.

I much prefer not using ember-cli-page-object to do this, but I _am_ curious as to why filling a text field is has an async API 🤔 , but I am glad we had failing tests due to the upgrade change 🎉

johncowen (commented):
Looks good, thanks for working on this! Can't wait to be able to upgrade to ember 4.4 and beyond 🎉 

I left a little thing about the `wait a tick` step which I think we should remove and another async related question, lmk!

ui/packages/consul-ui/tests/acceptance/dc/services/show-routing.feature
LevelbossMike: The tests break without this tick. I'm not sure what's going on and there is nothing wrong with the behavior itself so I added this in.

ui/packages/consul-ui/tests/steps/assertions/page.js
LevelbossMike: thanks for the background info

ui/packages/consul-ui/tests/acceptance/dc/services/show-routing.feature
LevelbossMike: Yeah, we could probably look into this test again in the future and rewrite it without ember-cli-page-object. For now, I think we should be fine for getting ember-qunit v5 landed and then tackling ember-cli-page-object as a follow-up.

ui/packages/consul-ui/tests/steps/interactions/form.js
johncowen: We spoke about this offline and @LevelbossMike decided the extra await here was necessary, so skipping this

johncowen (approved):
Good to go! :shipit:

PR #14433, opened 2022-08-31
Use proxy address for default check

lkysow:
When a sidecar proxy is registered, a check is automatically added.
Previously, the address this check used was the underlying service's
address instead of the proxy's address, even though the check is testing
if the proxy is up.

This worked in most cases because the proxy ran on the same IP as the
underlying service but it's not guaranteed and so the proper default
address should be the proxy's address.

The docs say:

> By default we add a TCP check on the local address and port for the proxy, and a [service alias check](https://www.consul.io/docs/discovery/checks#alias) for the parent service. If either check or checks fields are set, only the provided checks are registered.

Which is now correct but previously was incorrect.

* Tested via unit tests

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/sidecar_service.go
```diff
@@ -202,14 +202,16 @@ func (a *Agent) sidecarPortFromServiceID(sidecarCompoundServiceID structs.Servic
 	return sidecarPort, nil
 }
 
-func sidecarDefaultChecks(serviceID string, localServiceAddress string, port int) []*structs.CheckType {
+func sidecarDefaultChecks(serviceID string, proxyAddress string, port int) []*structs.CheckType {
+	if proxyAddress == "" {
```
lkysow, line 206: `sidecar.Address` is passed in here. If empty, it inherits from the parent service's address. If the parent service's address is empty, then it will default to `127.0.0.1`. This mimics the previous logic, where if `sidecar.Proxy.LocalServiceAddress` was empty then it defaulted to `127.0.0.1` (https://github.com/hashicorp/consul/blob/main/agent/sidecar_service.go#L108-L110).

agent/sidecar_service.go
```diff
@@ -202,14 +213,11 @@ func (a *Agent) sidecarPortFromServiceID(sidecarCompoundServiceID structs.Servic
 	return sidecarPort, nil
 }
 
-func sidecarDefaultChecks(serviceID string, localServiceAddress string, port int) []*structs.CheckType {
```
lkysow, line 205: I removed the comments here because the first one:

```// Setup default check if none given```

Is already commented where this function is called and the second one is now commented where we pass in the address.

skpratt (approved):

PR #14434, opened 2022-08-31
[OSS] fix(api): OSS<->ENT exported service incompatibility

DanStough:
### Description
The nomad team identified an issue using the OSS API with the Consul ENT flag. The OSS and ENT API modules should be compatible and not dependent on build flags. This change removes the build dependency of default Partitions/Namespaces from the code path. 

### PR Checklist

* [X] updated test coverage
* [X] not a security concern


api/oss_test.go
```diff
@@ -6,5 +6,5 @@ package api
 // The following defaults return "default" in enterprise and "" in OSS.
 // This constant is useful when a default value is needed for an
 // operation that will reject non-empty values in OSS.
-const splitDefaultNamespace = ""
-const splitDefaultPartition = ""
+const defaultNamespace = ""
+const defaultPartition = ""
```
DanStough, line 10: We don't have to change the names but the term "split" left me more confused

hashi-derek (approved):

Issue #9577, opened 2021-01-15
Unable to deregister orphan checks and service instances after accessorid token is lost and set to anonymous


idrennanvmware:
### Consul info for both Client and Server
Consul 1.9.1

After experiencing an unexpected cluster scenario (we had some nodes that had VMs crash) we have a number of orphaned services and checks in the Consul Service Discovery of a cluster (the clusters are in a healthy state besides these checks).....  However we are completely unable to deregister them.   

Here's what we see in the logs for Consul

```
 check=service:_nomad-task-9523fe99-72eb-5742-a85a-c17f759c04c8-group-device-log-consumer-group-device-log-consumer-device_log_consumer_http-s>
 check=service:_nomad-task-a8ca2701-18f0-074b-cfd1-fb7a315f8ecc-group-device-log-api-group-device-log-api-api_http-sidecar-proxy:1 accessorID=>
 check=_nomad-check-583c1b1a627be50caeb61c1bb2de8ba2ecad9d1e accessorID=00000000-0000-0000-0000-000000000002
 check=_nomad-check-91cb4d33d9a27e0cea69eaea8c00cf17702ccde4 accessorID=00000000-0000-0000-0000-000000000002
 check=fabio-uem-services-qe-genesis-csl2-29998-ttl accessorID=00000000-0000-0000-0000-000000000002
 check=_nomad-check-d6a48d1f861f0e27a18b9f99608110f2fd1ff4c0 accessorID=00000000-0000-0000-0000-000000000002
 check=service:_nomad-task-a8ca2701-18f0-074b-cfd1-fb7a315f8ecc-group-device-log-api-group-device-log-api-api_http-sidecar-proxy:2 accessorID=>
 check=_nomad-check-d7f9b9aa185bbc9dfdee2bafa655816d37f34ef0 accessorID=00000000-0000-0000-0000-000000000002
 check=_nomad-check-3d4c0b7928a4fd3306f6e88a0af2bf931ca11adc accessorID=00000000-0000-0000-0000-000000000002
 check=_nomad-check-485e79b78ae284830003473c8300f23c3ef9d22f accessorID=00000000-0000-0000-0000-000000000002
 ```
 
 Note the AccessorID - these services were not registered with that, but interestingly enough using the CLI and HTTP we are unable to clean these up.
 
 ```
 ./consul services deregister -id _nomad-task-a8ca2701-18f0-074b-cfd1-fb7a315f8ecc-group-device-log-api-group-device-log-api-api_http
[...]

Issue #4372, opened 2018-07-10
check status inconsistency 


danaps:
Hi folks,
we run Consul 1.2.0 and encountered the following issue:
in the consul UI, when looking at some node Health check we see it is failing. 

`curl http://localhost:8500/v1/health/node/ip-10-1-7-170

  {
    "Node": "ip-10-1-7-170",
    "CheckID": "c49af2eedf67fdbf7135045051aad72c5f1d8a4c",
    "Name": "Nomad Client HTTP Check",
    "Status": "critical",
    "Notes": "",
    "Output": "",
    "ServiceID": "_nomad-client-vsvkabaz4izafp6d5ejwyhrurtl2pdeq",
    "ServiceName": "nomad-client",
    "ServiceTags": [
      "http"
    ],
    "Definition": {},
    "CreateIndex": 170787117,
    "ModifyIndex": 170787117
  },

`
after checking underlying API call /v1/agent/checks on the node the check is actually passing.

` curl http://localhost:8500/v1/agent/checks 

"c49af2eedf67fdbf7135045051aad72c5f1d8a4c": {
    "Node": "ip-10-1-7-170",
    "CheckID": "c49af2eedf67fdbf7135045051aad72c5f1d8a4c",
    "Name": "Nomad Client HTTP Check",
    "Status": "passing",
    "Notes": "",
    "Output": "HTTP GET http://0.0.0.0:4646/v1/agent/health?type=client: 200 OK Output: {\"client\":{\"message\":\"ok\",\"ok\":true}}",
    "ServiceID": "_nomad-client-vsvkabaz4izafp6d5ejwyhrurtl2pdeq",
    "ServiceName": "nomad-client",
    "ServiceTags": [
      "http"
    ],
    "Definition": {},
    "CreateIndex": 0,
    "ModifyIndex": 0
  },
`

burdandrei:
Let me rephrase this:
After running 1.2.0 in staging for a week, a week ago we updated our production cluster to 1.2.0.
For monitoring, we use Telegraf that send the data to InfluxDB. After some time we noticed that in grafana we can see less healthy hosts, that are running. After all the debugging we found that some checks that we see in UI (Both official and Hashi-UI) as critical are shown as passing of we're checking it on the agent locally `curl http://localhost:8500/v1/agent/checks`.
It looks like the UI is using "consensus" endpoint to  show the status checks: `curl http://localhost:8500/v1/health/node/:node`
[...]

PR #14436, opened 2022-09-01
agent: Fall back to using the Consul agents token when faced with "ACL not found" during deregistrations

weeezes:
### Description

Currently we're seeing an issue when generating Consul ACL tokens for a service with the [Vault Consul Secret Engine](https://www.vaultproject.io/api-docs/secret/consul), where services/checks become "zombies" in the Consul catalog when the Vault token gets revoked or the Consul token gets removed before the service using those tokens has a chance to deregister itself and its checks. The "zombie" services remain in a "healthy" state, even though the TTL checks aren't updated anymore and thus should fall in to failing state.

In case of an "ACL not found" error remove the token associated with the service and check in the state, so that further attemts fall back to using the agent token.

This makes the agent clear the services token from its local state (making the assumption that the token is actually gone and won't ever come back again), and the next re-try falls back to trying out the agent token.

Our test setup:

* Consul deployed with Helm
* Vault deployed with Helm
* Consul Secret Backend configured to Vault
* A service that registers itself to the local Consul Agent with [Agent.ServiceRegister](https://pkg.go.dev/github.com/hashicorp/consul/api#Agent.ServiceRegister), using a token generated through Vault using the Consul Secret Backend 

When for any reason the Vault token or the generated Consul token gets revoked/removed before the service has a chance to deregister itself there should be services/checks left behind in the servers Catalog, and the agent will log lines like this:

```
2022-09-01T09:41:51.976Z [ERROR] agent.client: RPC failed to server: method=Catalog.Register server=10.42.0.66:8300 error="rpc error making call: ACL not found"
2022-09-01T09:41:51.976Z [WARN]  agent: Check registration blocked by ACLs: check=service:YSbpY7U3HghRGcEijnBEB3 accessorID=
```

When using `DeregisterCriticalServiceAfter` with a TTL check the service and checks do get removed from the agent properly, which is handy because the service doesn't have a working ACL token anymore at that point so that it could update the TTL. But that state doesn't get propagated to the catalog. 

### Links

There are a few issues that I believe are also related to this:

https://github.com/hashicorp/consul/issues/9577
https://github.com/hashicorp/nomad/issues/13537

The response [here](https://github.com/hashicorp/consul/issues/9577#issuecomment-771203024) gave me the idea that this might be a way to fix the problem.

Also this issue seems relevant: https://github.com/hashicorp/consul/issues/4372

### PR Checklist

I might need a bit of help in figuring out what tests to update, and if there's any relevant documentation that should be updated regarding this 😊 .

* [ ] updated test coverage
* [x] not a security concern


weeezes:
Hi! Any tips on how to move this forward :)? Is the approach wrong?

weeezes:
Apparently some potentially flaky tests that are causing failures, am I interpreting that right?

Is there something I can do to help this move forward :)?

pglass:
Hi @weeezes, thanks for this change!

If you could do the following, please:

* Update this PR with main so that the tests are re-run
* Add a [changelog entry](https://github.com/hashicorp/go-changelog#change-file-formatting). For this PR, you can add the file `.changelog/14436.txt` (where 14436 is number of this PR) with contents like the following:

  ``````
  $ cat .changelog/14436.txt
  ```release-note:improvement
  agent: client agents fallback to using the agent token for 
  ```
  ``````

Thanks!

agent/local/state.go
```diff
@@ -1295,7 +1295,10 @@ func (l *State) deleteService(key structs.ServiceID) error {
 		l.logger.Info("Deregistered service", "service", key.ID)
 		return nil
 
-	case acl.IsErrPermissionDenied(err), acl.IsErrNotFound(err):
+	case acl.IsErrNotFound(err):
+		l.services[key].Token = ""
+		fallthrough
+	case acl.IsErrPermissionDenied(err):
```
pglass, line 1301: This works!

What I'd suggest actually is not using the service token for the deregistration. That will simplify the logic and behavior slightly.

Like on [line 1273](https://github.com/hashicorp/consul/pull/14436/files#diff-3f2d3e802e18646676be327dbda91428644785290c9d37d5ca37c2124c4170e7R1273), we could do the following (and same for the deleteCheck function)

```diff
- st := l.aclTokenForServiceSync(key, l.tokens.AgentToken)
+ st := l.tokens.AgentToken
```

What do you think abut this? cc @jkirschner-hashicorp @rboyer

weeezes:
Thanks for the comments! I'm on parental leave until early next month, I'll check in more detail when I'm back :).

pglass:
Hi, thanks for this PR! 

After some internal discussion on the approach and in order to include a fix in time for the next Consul release, I've opened a new PR which takes the approach of always using the agent token: https://github.com/hashicorp/consul/pull/16097. I've included your commit from this PR for credit 🙂, and I'll close this PR for now.

weeezes:
awesome, thank you ❤️ !

PR #14437, opened 2022-09-01
Allow Operator Generated bootstrap token

apollo13:
### Description
Added support to provide an initial token via the bootstrap HTTP API, similar to https://github.com/hashicorp/nomad/pull/12520

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


jkirschner-hashicorp:
Hello @apollo13,

Thank you for submitting this PR for our consideration! To help with that consideration, can you provide more context on:
- The need that this PR is intended to address (since there is no linked Github issue)?
- The ways in which this PR addresses said need in a way that is more suitable for your use case than the existing Consul server agent configuration option [`acl.tokens.initial_management`](https://www.consul.io/docs/agent/config/config-files#acl_tokens_initial_management)?
  - As an aside, if you weren't previously aware of the `acl.tokens.initial_management` option, that suggests an opportunity for us to improve the ACL bootstrapping docs

apollo13:
Hi @jkirschner-hashicorp,

> Thank you for submitting this PR for our consideration! To help with that consideration, can you provide more context on:

Sure, thank you for answering.

> * The need that this PR is intended to address (since there is no linked Github issue)?

I want to be able to programmatically bootstrap a Consul cluster. If nothing else, the hashicorp products should preferably behave the same way and this makes it nicely match nomad. 

> * The ways in which this PR addresses said need in a way that is more suitable for your use case than the existing Consul server agent configuration option [`acl.tokens.initial_management`](https://www.consul.io/docs/agent/config/config-files#acl_tokens_initial_management)?

I am aware of the initial management token but I rather not write (even if it is just an initial token) secrets like that onto the harddisk of a server if not needed. I understand that the raft/storage-layer will write the token anyways, but that storage layer is ecxpected to hold sensitive data (connect key material among others and in the worst case kv data is sensitive as well).

To make matters worse I am trying to precreate usable containers/vm images. The less key material I have to roll out the better. Initial server bootstrap works nicely via mdns cloud-join (got that working for nomad & consul). I understand that I will probably have to put other tokens into the config as well, but the fallout of those other tokens would be far less than the root key…

I hope this helps; I realize I am probably not making the best arguments for this feature -- it is not an absolute must-have (I understand that), but nevertheless it would make things easier for a few people out here.

jkirschner-hashicorp:
Thank you for the context! I'll share this with the team.

apollo13:
//Unstable

command/acl/bootstrap/bootstrap.go
```diff
@@ -43,13 +46,46 @@ func (c *cmd) Run(args []string) int {
...
+	if len(args) == 1 {
+		switch args[0] {
+		case "":
+			terminalToken = []byte{}
+		case "-":
+			terminalToken, err = ioutil.ReadAll(os.Stdin)
+		default:
+			file := args[0]
+			terminalToken, err = ioutil.ReadFile(file)
+		}
```
rboyer, line 67: There is already a [CLI helper](https://github.com/hashicorp/consul/blob/main/command/helpers/helpers.go#L57) to assist with this sequence of operations that we should use:

```suggestion
		terminalToken, err = helpers.LoadDataSourceNoRaw(args[0], os.Stdin)
```
Note that this is going to return a `string` already, so the typing for `terminalToken` needs fixing too. You could probably parse it directly into `var initialManagementToken string`.

command/acl/bootstrap/bootstrap.go
```diff
@@ -43,13 +46,46 @@ func (c *cmd) Run(args []string) int {
...
+		if err != nil {
+			c.UI.Error(fmt.Sprintf("Error reading provided token: %v", err))
+			return 1
+		}
+	}
+
+	// Remove newline from the token if it was passed by stdin
+	boottoken := strings.TrimSuffix(string(terminalToken), "\n")
```
rboyer, line 75: Since we know further down in the processing pipeline that we'll only accept a valid UUID for this field, it might be simpler to avoid any whitespace issues here by using `strings.TrimSpace` instead.

rboyer (changes requested):
Great job on this! I have a variety of small comments and suggestions I'm adding below but none of them should be surprising or tricky to clean up before merging.

One of the longer comments is not actionable and is pointing out a `msgpack` non-issue to a future PR archaeologist.

agent/acl_endpoint.go
```diff
@@ -34,9 +34,20 @@ func (s *HTTPHandlers) ACLBootstrap(resp http.ResponseWriter, req *http.Request)
 		return nil, aclDisabled
 	}
 
-	args := structs.DCSpecificRequest{
+	args := structs.ACLInitialTokenBootstrapRequest{
```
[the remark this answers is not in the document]
jkirschner-hashicorp: Is this a line of code that a future dev might look at and go "wait, this looks like it might be wrong..."? If so, we might want an in-source comment. I defer to @rboyer

command/acl/bootstrap/bootstrap_test.go
```diff
@@ -87,3 +89,50 @@ func TestBootstrapCommand_JSON(t *testing.T) {
...
+	}`)
+
+	defer a.Shutdown()
+	testrpc.WaitForLeader(t, a.RPC, "dc1")
+
+	ui := cli.NewMockUi()
+	cmd := New(ui)
+
+	// Create temp file
+	f, err := ioutil.TempFile("", "consul-token.token")
```
kisunji, line 113: Could we avoid using `ioutil`? Our linters will probably flag this as deprecated.
```suggestion
	f, err := os.CreateTemp("", "consul-token.token")
```

command/acl/bootstrap/bootstrap_test.go
```diff
@@ -87,3 +89,50 @@ func TestBootstrapCommand_JSON(t *testing.T) {
...
+	assert.Nil(t, err)
+	defer os.Remove(f.Name())
+
+	// Write the token to the file
+	err = ioutil.WriteFile(f.Name(), []byte("2b778dd9-f5f1-6f29-b4b4-9a5fa948757a"), 0700)
```
kisunji, line 118:
```suggestion
	err = os.WriteFile(f.Name(), []byte("2b778dd9-f5f1-6f29-b4b4-9a5fa948757a"), 0700)
```

command/acl/bootstrap/bootstrap_test.go
```diff
@@ -2,6 +2,8 @@ package bootstrap
 
 import (
 	"encoding/json"
+	"io/ioutil"
```
kisunji, line 5:
```suggestion
```

apollo13:
Thank you for the reviews. I will need a few days to get this updated.

agent/consul/acl_endpoint.go
```diff
@@ -209,9 +209,24 @@ func (a *ACL) BootstrapTokens(args *structs.DCSpecificRequest, reply *structs.AC
...
+		secret, err = lib.GenerateUUID(a.srv.checkTokenUUID)
+		if err != nil {
+			return err
+		}
+	} else {
+		_, err = uuid.ParseUUID(secret)
+		if err != nil {
+			return err
+		}
+		ok, err := a.srv.checkTokenUUID(secret)
```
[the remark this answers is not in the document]
apollo13: I `checkTokenUUID` also checks that the token does not start with `ACLReservedPrefix`, I'd like to at least keep that check. Should I switch to using `ACLIDReserved(secret)`?

agent/structs/acl.go
```diff
@@ -1350,10 +1350,20 @@ type ACLTokenBatchDeleteRequest struct {
 	TokenIDs []string // Tokens to delete
 }
 
+type ACLInitialTokenBootstrapRequest struct {
+	BootstrapSecret string
```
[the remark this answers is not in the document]
apollo13: I don't have strong feelings; I mirrored nomad: https://github.com/hashicorp/nomad/blob/fbe9f590489e291b3abcfb0fb1548a9f435a5e2d/nomad/structs/structs.go#L12169 -- I'd be fine with changing it if you think another name is better.

apollo13:
@kisunji Any chance of another review?

rboyer (approved):

PR #14445, opened 2022-09-01
Implement Cluster Peering Redirects

erichaberkorn:
### Description
This finishes implement service resolver redirects to services running on cluster peers.

* xDS golden tests
* Updating an integration test to test this case.

### PR Checklist
* [X] updated test coverage
* [ ] external facing docs updated - I'll update them next
* [X] not a security concern

### DIFF
--- a/agent/xds/clusters.go
+++ b/agent/xds/clusters.go
@@ -980,7 +980,7 @@ func (s *ResourceGenerator) makeUpstreamClustersForDiscoveryChain(
 	// Mesh gateways are exempt because upstreamsSnapshot is only used for
 	// cluster peering targets and transative failover/redirects are unsupported.
 	if err != nil && !forMeshGateway {
-		return nil, fmt.Errorf("No upstream snapshot for gateway mode %q", cfgSnap.Kind)
+		return nil, err
 	}
 
 	rawUpstreamConfig, err := structs.ParseUpstreamConfigNoDefaults(upstreamConfigMap)
--- a/agent/xds/clusters.go
+++ b/agent/xds/clusters.go
@@ -1038,11 +1038,11 @@ func (s *ResourceGenerator) makeUpstreamClustersForDiscoveryChain(
 		// These variables are prefixed with primary to avoid shaddowing bugs.
 		primaryTargetID := node.Resolver.Target
 		primaryTarget := chain.Targets[primaryTargetID]
-		primaryClusterName := CustomizeClusterName(primaryTarget.Name, chain)
-		upstreamConfig := finalizeUpstreamConfig(rawUpstreamConfig, node.Resolver.ConnectTimeout)
-		if forMeshGateway {
-			primaryClusterName = meshGatewayExportedClusterNamePrefix + primaryClusterName
+		primaryTargetClusterData, ok := s.getTargetClusterData(upstreamsSnapshot, chain, primaryTargetID, forMeshGateway, false)
+		if !ok {
+			continue
 		}
+		upstreamConfig := finalizeUpstreamConfig(rawUpstreamConfig, node.Resolver.ConnectTimeout)
 
 		if forMeshGateway && !cfgSnap.Locality.Matches(primaryTarget.Datacenter, primaryTarget.Partition) {
 			s.Logger.Warn("ignoring discovery chain target that crosses a datacenter or partition boundary in a mesh gateway",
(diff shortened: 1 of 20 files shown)


test/integration/connect/envoy/case-cfg-resolver-cluster-peering-failover/primary/verify.bats
```diff
@@ -85,3 +91,27 @@ load helpers
...
+  reset_envoy_metrics 127.0.0.1:19000
+}
+
+@test "s1 upstream should have healthy endpoints for virtual-s2" {
+  assert_upstream_has_endpoints_in_status 127.0.0.1:19000 s2.default.primary-to-alpha.external HEALTHY 1
+}
+
+@test "s1 upstream should have healthy endpoints for s2 in the cluster peer" {
+  assert_upstream_has_endpoints_in_status 127.0.0.1:19000 s2.default.primary-to-alpha.external HEALTHY 1
+}
```
DanStough, line 107: These are the same steps. I think you wanted to test the virtual service in the first one, correct?

agent/xds/clusters_test.go
```diff
@@ -275,6 +275,12 @@ func TestClustersFromSnapshot(t *testing.T) {
 				return proxycfg.TestConfigSnapshotDiscoveryChain(t, "failover-to-cluster-peer", nil, nil)
 			},
 		},
+		{
+			name: "connect-proxy-with-chain-and-redirect-to-cluster-peer",
+			create: func(t testinf.T) *proxycfg.ConfigSnapshot {
+				return proxycfg.TestConfigSnapshotDiscoveryChain(t, "redirect-to-cluster-peer", nil, nil)
+			},
+		},
```
DanStough, line 283: RB showed me that you can accomplish the same thing as adding this test case in 4 different places by just adding the test to [resources_test.go](https://github.com/hashicorp/consul/blob/main/agent/xds/resources_test.go#L197-L218).

DanStough (commented):
Just was looking over this for my own learning and had some comments.

agent/xds/routes.go
```diff
@@ -500,19 +501,26 @@ func generateUpstreamIngressDomains(listenerKey proxycfg.IngressListenerKey, u s
...
 	var routes []*envoy_route_v3.Route
 
 	startNode := chain.Nodes[chain.StartNode]
 	if startNode == nil {
 		return nil, fmt.Errorf("missing first node in compiled discovery chain for: %s", chain.ServiceName)
 	}
 
+	upstreamsSnapshot, err := cfgSnap.ToConfigSnapshotUpstreams()
+	if err != nil && !forMeshGateway {
+		return nil, fmt.Errorf("No upstream snapshot for gateway mode %q", cfgSnap.Kind)
```
freddygv, line 521: How come this isn't returning the error from `ToConfigSnapshotUpstreams`?

agent/xds/routes.go
```diff
@@ -599,44 +611,43 @@ func makeUpstreamRouteForDiscoveryChain(
 		}
 
 	case structs.DiscoveryGraphNodeTypeSplitter:
-		routeAction, err := makeRouteActionForSplitter(startNode.Splits, chain, clusterNamePrefix)
-		if err != nil {
-			return nil, err
-		}
+		routeAction, err := s.makeRouteActionForSplitter(upstreamsSnapshot, startNode.Splits, chain, forMeshGateway)
```
freddygv, line 614: Why not return if there's an error here? Also the error isn't being checked anymore

agent/xds/routes.go
```diff
@@ -599,44 +611,43 @@ func makeUpstreamRouteForDiscoveryChain(
...
-			Action: routeAction,
+			routes = []*envoy_route_v3.Route{defaultRoute}
 		}
 
-		routes = []*envoy_route_v3.Route{defaultRoute}
-
 	case structs.DiscoveryGraphNodeTypeResolver:
-		routeAction := makeRouteActionForChainCluster(startNode.Resolver.Target, chain, clusterNamePrefix)
+		routeAction, ok := s.makeRouteActionForChainCluster(upstreamsSnapshot, startNode.Resolver.Target, chain, forMeshGateway)
+		if ok {
```
freddygv, line 634: Could reduce the amount of indentation here by breaking if `!ok`
```
if !ok {
	break
}
```

erichaberkorn: Ah, nice. I tried to use `continue` and it didn't work. This is awesome.

test/integration/connect/envoy/case-cfg-resolver-cluster-peering-failover/primary/verify.bats
erichaberkorn: I remember how this works now. The cluster name is the name of the destination cluster and doesn't take the local service name into account. For example, in our tests we [redirect db to db in cluster peer cluster-01](https://github.com/hashicorp/consul/pull/14445/files#diff-55a80e75e8a5a8ed1d8360d217c766ebbe0949abf67ce3d51195ac6e2d8fe9c2R311-R320). The resulting cluster is [named `db.default.cluster-01.external.peer1.domain` rather than the local db cluster name](https://github.com/hashicorp/consul/pull/14445/files#diff-e61d0ee1542c3bcd0de50f6fb9eff56f35bc209cfa013e0c162c3ef18e2aff58R6).

I removed the duplicate assertion here.

agent/xds/clusters.go
```diff
@@ -1052,54 +1052,28 @@ func (s *ResourceGenerator) makeUpstreamClustersForDiscoveryChain(
...
-					if !ok {
-						s.Logger.Debug("peer trust bundle not ready for discovery chain target",
-							"peer", targetUID.Peer,
-							"target", tid,
-						)
-						continue
-					}
-
-					clusterName = generatePeeredClusterName(targetUID, tbs)
+				if td, ok := s.getTargetClusterData(upstreamsSnapshot, chain, tid, forMeshGateway, true); ok {
```
DanStough, line 1061: `forMeshGateway` is always true here, correct?

DanStough (approved):
Just one nit. Otherwsie 👍

PR #14474, opened 2022-09-02
Add kv txn get-not-exists operation.

hashi-derek:
### Description
Adds a `get-or-empty` operation to the kv txn api. There is currently no operation for fetching the value of a key that may not yet exist. The current `get` command will cause the entire transaction to roll-back if the key doesn't exist. The new op will return a `nil` value for a key if-not-found.

https://www.consul.io/api-docs/txn#kv-operations

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/consul/state/txn.go
+++ b/agent/consul/state/txn.go
@@ -60,6 +60,13 @@ func (s *Store) txnKVS(tx WriteTxn, idx uint64, op *structs.TxnKVOp) (structs.Tx
 			err = fmt.Errorf("key %q doesn't exist", op.DirEnt.Key)
 		}
 
+	case api.KVGetOrEmpty:
+		_, entry, err = kvsGetTxn(tx, nil, op.DirEnt.Key, op.DirEnt.EnterpriseMeta)
+		if entry == nil && err == nil {
+			entry = &op.DirEnt
+			entry.Value = nil
+		}
+
 	case api.KVGetTree:
 		var entries structs.DirEntries
 		_, entries, err = s.kvsListTxn(tx, nil, op.DirEnt.Key, op.DirEnt.EnterpriseMeta)
@@ -95,7 +102,7 @@ func (s *Store) txnKVS(tx WriteTxn, idx uint64, op *structs.TxnKVOp) (structs.Tx
 	// value (we have to clone so we don't modify the entry being used by
 	// the state store).
 	if entry != nil {
-		if op.Verb == api.KVGet {
+		if op.Verb == api.KVGet || op.Verb == api.KVGetOrEmpty {
 			result := structs.TxnResult{KV: entry}
 			return structs.TxnResults{&result}, nil
 		}
--- a/agent/consul/state/txn_test.go
+++ b/agent/consul/state/txn_test.go
@@ -579,6 +579,22 @@ func TestStateStore_Txn_KVS(t *testing.T) {
 		},
 		&structs.TxnOp{
 			KV: &structs.TxnKVOp{
+				Verb: api.KVGetOrEmpty,
+				DirEnt: structs.DirEntry{
+					Key: "foo/update",
+				},
+			},
+		},
+		&structs.TxnOp{
+			KV: &structs.TxnKVOp{
+				Verb: api.KVGetOrEmpty,
+				DirEnt: structs.DirEntry{
+					Key: "foo/not-exists",
+				},
+			},
+		},
+		&structs.TxnOp{
+			KV: &structs.TxnKVOp{
 				Verb: api.KVCheckIndex,
 				DirEnt: structs.DirEntry{
 					Key: "foo/update",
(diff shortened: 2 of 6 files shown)


agent/consul/state/txn.go
```diff
@@ -60,6 +60,12 @@ func (s *Store) txnKVS(tx WriteTxn, idx uint64, op *structs.TxnKVOp) (structs.Tx
 			err = fmt.Errorf("key %q doesn't exist", op.DirEnt.Key)
 		}
 
+	case api.KVGetNotExists:
+		_, entry, err = kvsGetTxn(tx, nil, op.DirEnt.Key, op.DirEnt.EnterpriseMeta)
+		if entry == nil && err == nil {
+			entry = &op.DirEnt
```
hashi-derek, line 66: The `api.KVSet` does a similar operation by returning the provided `op.DirEnt` to represent an empty key.

banks: 🤔 I'm not sure I understand the comment here. In the case of `KVSet` `entry = &op.DirEntry` is just setting the entry to return to be the same one passed in but it should have a non-empty value. If the value is actually empty in the `op.DirEntry` then that means that after the op the key _will_ exist in the KV just with an explicitly empty value which is different to not existing. 

That said, the actual behaviour seems reasonable here, but it's not super obvious and perhaps a comment in the code to the effect of `return the passed `DirEntry` since it has the key set and a nil value as the response value`?

agent/consul/state/txn_test.go
```diff
@@ -577,6 +577,22 @@ func TestStateStore_Txn_KVS(t *testing.T) {
...
+				DirEnt: structs.DirEntry{
+					Key: "foo/update",
+				},
+			},
+		},
+		&structs.TxnOp{
+			KV: &structs.TxnKVOp{
+				Verb: api.KVGetNotExists,
+				DirEnt: structs.DirEntry{
+					Key: "foo/not-exists",
```
banks, line 592: nit: do you think it would be clearer if we comment that we expect the Value to be nil here explicitly rather than just omitting it?

website/content/api-docs/txn.mdx
```diff
@@ -266,20 +266,21 @@ look like this:
...
-| `delete-tree`      | Delete all keys with a prefix           | `x` |       |       |       |         |
-| `delete-cas`       | Delete, but with CAS semantics          | `x` |       |       |  `x`  |         |
+| Verb               | Operation                                | Key | Value | Flags | Index | Session |
+| ------------------ | ---------------------------------------- | :-: | :---: | :---: | :---: | :-----: |
+| `set`              | Sets the `Key` to the given `Value`      | `x` |  `x`  |  `o`  |       |         |
+| `cas`              | Sets, but with CAS semantics             | `x` |  `x`  |  `o`  |  `x`  |         |
+| `lock`             | Lock with the given `Session`            | `x` |  `x`  |  `o`  |       |   `x`   |
+| `unlock`           | Unlock with the given `Session`          | `x` |  `x`  |  `o`  |       |   `x`   |
+| `get`              | Get the key, fails if it does not exist  | `x` |       |       |       |         |
+| `get-not-exists`   | Get the key, or nil if it does not exist | `x` |       |       |       |         |
```
banks, line 276: nit: the example in the API docs are referring to JSON response values where we mention in the docs above this that the `Value` field is "a base64 encoded string or null". Should we be consistent and say `null` rather than `nil` here?

banks (commented):
Amazing job adding this small enhancement @hashi-derek 👏 

Overall this is exactly what I imagined too when thinking about this problem.

I have a couple of minor nits inline which I don't think block merging this.

My biggest bit of feedback would be on the name chosen. I think `get-if-exists` is slightly more clear/descriptive  of the behaviour vs. `get-not-exists`. I see how there is a consistency argument with `check-not-exists` but I think that makes more sense in context "Check that key X does not exist" is naturally expressed by `check-not-exists` whereas "Get the value of key X if it exists or empty otherwise" requires a double-negative to map to `get-not-exists` and maps more naturally to `get-if-exists` in my mind anyway 😄.

If you disagree though it'd not a mountain I feel the need to die on! Should be a quick enough update if you do!

agent/consul/state/txn.go
hashi-derek: Yeah, my comment was a little confusing -- sorry about that. Further down in the function, there's this little bit of code that clears out the value from the `KVSet` ops.

```
clone := entry.Clone()
clone.Value = nil
```

I was mainly pointing out that the `KVSet` still puts an entry in the results by re-using the `op.DirEnt` and nulling the value. That being said, I explicitly put in a `entry.Value = nil` into the code now so that it's guaranteed to always be nil, rather than returning back what the user specified, which is more clear.

hashi-derek:
@kisunji Yeah, I wasn't a fan of any of the names I had come up with. `get-or-empty` seems like a reasonable choice. Since the action will always return a result, it's probably more clear than `get-if-exists` as well.

kisunji (approved):
👍

PR #14485, opened 2022-09-02
Add SpiffeID for Consul server agents

freddygv:
### Description
By adding a SpiffeID for server agents, servers can now request a leaf
certificate from the Connect CA.

This new Spiffe ID has a key property: servers are identified by their
datacenter name and trust domain. All servers that share these
attributes will share a ServerURI.

The aim is to use these certificates to verify the server name of ANY
server in a Consul datacenter.

* Unit tests for Spiffe ID validation and parsing
* Test with test server to validate certificate signing flow

### Links
https://go.hashi.co/csl-204-server-cert

### PR Checklist

* [x] updated test coverage
* [ ] not a security concern


agent/connect/uri_signing_test.go
```diff
@@ -119,6 +119,30 @@ func TestSpiffeIDSigning_CanSign(t *testing.T) {
...
 			want:  false,
 		},
+		{
+			name:  "server - good",
+			id:    testSigning,
+			input: &SpiffeIDServer{Host: TestClusterID + ".consul", Datacenter: "dc1"},
+			want:  true,
+		},
+		{
+			name:  "server - good midex case",
```
erichaberkorn, line 129:
```suggestion
			name:  "server - good mixed case",
```

erichaberkorn (approved):

agent/structs/connect_ca.go
```diff
@@ -224,6 +224,10 @@ type IssuedCert struct {
 	// AgentURI is the cert URI value.
 	AgentURI string `json:",omitempty"`
 
+	// ServerURI is the URI value of a cert issued for a server agent.
+	// The same URI is shared by all servers in a Consul datacenter.
+	ServerURI string `json:",omitempty"`
+
```
DanStough, line 230: Just curious why `IssuedCert` isn't exported from the `API` package? Looks like it's returned from at least one endpoint `/v1/agent/connect/ca/leaf/test?token=<token>`.

freddygv: @DanStough It is in the `api` package, but for some reason it has a different name there. I assume it was for clarity? https://github.com/hashicorp/consul/commit/9638466b88aab7c415fba1bf035bb37e8d4efd66

Cert type: https://github.com/hashicorp/consul/blob/8c46e48e0dafc9d474c8333a86b4d652a43a5f5a/api/connect_ca.go#L108
Endpoint client: https://github.com/hashicorp/consul/blob/8c46e48e0dafc9d474c8333a86b4d652a43a5f5a/api/agent.go#L1112

We don't need to expose `ServerURI` via that type because users can't request leaf certs for Consul servers.

PR #14417, opened 2022-08-31
ui: Adds a HCP home link when in HCP

johncowen:
### Description
Adds a Back to HCP link only when HCP is enabled, the link uses a yet to exist `CONSUL_HCP_URL` "env var"

### PR Checklist

* [ ] updated test coverage
* [ ] external facing docs updated
* [x] not a security concern

PR #14489, opened 2022-09-05
ui: Use HCPURL / .HCPURL

johncowen:
### Description
Second part of https://github.com/hashicorp/consul/pull/14417, adds picking up the `.HCPURL` "env var"

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-hcp/app/components/consul/hcp/home/index.hbs
+++ b/ui/packages/consul-hcp/app/components/consul/hcp/home/index.hbs
@@ -2,7 +2,7 @@
   class="consul-hcp-home"
   ...attributes
 >
-  <a href={{env 'CONSUL_HCP_URL'}}>
+  <a href={{env 'CONSUL_HCP_URL'}} data-native-href="true">
     Back to HCP
   </a>
 </div>
--- a/ui/packages/consul-hcp/app/components/consul/hcp/home/index.test.js
+++ b/ui/packages/consul-hcp/app/components/consul/hcp/home/index.test.js
@@ -1,0 +1,38 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import hbs from 'htmlbars-inline-precompile';
+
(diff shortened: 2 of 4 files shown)


ui/packages/consul-hcp/app/components/consul/hcp/home/index.hbs
```diff
@@ -2,7 +2,7 @@
   class="consul-hcp-home"
   ...attributes
 >
-  <a href={{env 'CONSUL_HCP_URL'}}>
+  <a href={{env 'CONSUL_HCP_URL'}} data-native-href="true">
```
LevelbossMike, line 5: Why are we adding this data-attribute here? Is this changing the behavior of the link or its styling? I don't see it being used anywhere else in the application. Is this a left-over from development and can be removed?

LevelbossMike (approved):
I added one question. Also, do we want to add a test for this behavior?

ui/packages/consul-hcp/app/components/consul/hcp/home/index.hbs
johncowen: It disables the helper from the ember href-to addon from preventingDefault and dealing with the event itself, so no, we do need this right now

johncowen:
> Also, do we want to add a test for this behavior?

Me personally I would say no not really, but lets ask @evrowe

ui/packages/consul-hcp/app/components/consul/hcp/home/index.hbs
LevelbossMike: I might be missing something here, but aren't we implementing the `href-to` ourselves? I don't see where we are including/extending the ember-href-to-addon?!

evrowe:
If the behavior invoked by adding `data-native-href="true"` to the code is covered by the tests for our implementation of `href-to`, then I would be fine with not adding tests for this particular instance.

That being said, I do think it would be worthwhile to add a/some test(s) validating that when `CONSUL_HCP_URL` is populated, the rendered HTML contains the value of that ENV variable as expected.

PR #14495, opened 2022-09-06
ui: Detect token in a cookie and passthrough

johncowen:
### Description
Detects existence of a `TokenSecretID` cookie. When set, this gets passthrough to localStorage and therefore "automatically logs you in". Only happens when ACLs are enabled and you don't already have a token.

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/.changelog/14495.txt
+++ b/.changelog/14495.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+ui: Detect a TokenSecretID cookie and passthrough to localStorage
+```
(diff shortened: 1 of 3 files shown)


evrowe (approved):

LevelbossMike (changes requested):
Change looks fine to me. This is changing the way the application boots up. Can we add an acceptance test for this new behavior?

PR #14499, opened 2022-09-06
Make `proto-public` a Go module

boxofrad:
### Description
Our original intention was for projects to consume and generate their own Go code for our public protobuf packages using [Buf](https://buf.build/).

While this is still the best route for many projects, it causes some headaches when using a library (e.g. [consul-server-connection-manager](https://github.com/hashicorp/consul-server-connection-manager)) that pulls in the same protobuf package as your project, as Go's protobuf implementation [only allows for a package/namespace to be registered once](https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict).

In such cases, projects can depend on this Go module instead, as a single place where these protobuf packages are registered.

Closes https://github.com/hashicorp/consul/pull/14493


proto-public/go.mod
```diff
@@ -0,0 +1,16 @@
+module github.com/hashicorp/consul/proto-public
+
+go 1.19
+
+require (
+	github.com/golang/protobuf v1.5.0
```
riddhi89, line 6: Do we need this?

boxofrad: I think so because it go pulled in automatically with `go mod tidy` 🤷🏻

riddhi89 (approved):

PR #14510, opened 2022-09-07
ui: fix - proper await page object step

LevelbossMike:
### Description
This PR resolves an issue in the way we were working with ember-cli-page-object in our steps. The change introduced here makes sure we handle the fact that calling a page-object method is asynchronous. We need to return the promise instead of just `Promise.resolve`, so we properly await whatever happens based on the interaction with the page object.

This allows us to remove a (bad) sleep workaround introduced in this PR: https://github.com/hashicorp/consul/pull/14430

In addition to that, properly awaiting page-object promises in the `I see`-step creates the need to change the test order in another tab-nav related test. Because `see`ing a tab-item will actually click the item, we have to change the order of what things are being testing in that test.

This PR also updates ember-cli-page-object to the latest 1.17.x release, which was released recently.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-ui/package.json
+++ b/ui/packages/consul-ui/package.json
@@ -108,7 +108,7 @@
     "ember-cli-flash": "^2.1.1",
     "ember-cli-htmlbars": "^5.2.0",
     "ember-cli-inject-live-reload": "^2.0.2",
-    "ember-cli-page-object": "^1.16.2",
+    "ember-cli-page-object": "^1.17.10",
     "ember-cli-sass": "^10.0.1",
     "ember-cli-sri": "^2.1.1",
     "ember-cli-string-helpers": "^5.0.0",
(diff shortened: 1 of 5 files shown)


johncowen (approved):
Thanks for spotting and sorting this!

I took a look at it myself also, just to try and follow whats happening here, and it seems like ember-cli-page-objects methods don't return promises, they return 'thenables' i.e. an object with a `then` method (but no `catch`) and a bunch of other stuff, rather like Ember Promise Proxies. If you call `.then()` on the result of the ember-cli-page-object, _that then gives you the promise._

I think apart from that from our discussion offline we both aren't happy with how we are calling "I see", and what it is actually doing is a "I see and click" whenever you are asserting seeing anything on the "tab-nav":

https://github.com/hashicorp/consul/blob/0abb923d661350f0bea344887f16a1d58c1836df/ui/packages/consul-ui/app/components/tab-nav/pageobject.js#L21

 but we want to get this fix in now and then do a quick follow up with how to tackle that afterwards. Ideally it would involve us moving off of this older step and onto the newer replacement one for this - but all depends on how much effort that is.

PR #14516, opened 2022-09-08
Fix inconsistent TTL behavior in CA providers

kyhavlov:
This PR has a couple changes related to CA provider configuration:
- Update the docs to clarify that `IntermediateCertTTL` is only valid in the primary datacenter, as that's where the intermediates are signed. It isn't used at all in the secondary DCs.
- Update the Vault provider to update the intermediate PKI mount and role when `Configure()` is called. Previously these values were only set on creation of the mount/role and then any updates to the CA config were not reflected in Vault.

### DIFF
--- a/agent/connect/ca/provider_vault.go
+++ b/agent/connect/ca/provider_vault.go
@@ -66,11 +66,10 @@ type VaultProvider struct {
 
 	stopWatcher func()
 
-	isPrimary                    bool
-	clusterID                    string
-	spiffeID                     *connect.SpiffeIDSigning
-	setupIntermediatePKIPathDone bool
-	logger                       hclog.Logger
+	isPrimary bool
+	clusterID string
+	spiffeID  *connect.SpiffeIDSigning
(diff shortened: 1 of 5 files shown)


agent/connect/ca/provider_vault.go
```diff
@@ -737,6 +724,19 @@ func (v *VaultProvider) mountNamespaced(namespace, path string, mountInfo *vault
 	return err
 }
 
+func (v *VaultProvider) tuneMountNamespaced(namespace, path string, mountConfig *vaultapi.MountConfigInput) error {
```
kisunji, line 727: Could you explain what this does in place of the readNamespaced codepath it replaces?

kyhavlov: it's not a replacement for `readNamespaced` - it's basically the same as `mountNamespaced` above but for reconfiguring the pki mount for the intermediate CA cert. In the case where that mount already exists we have to call `/tune` instead.

PR #14517, opened 2022-09-08
ui: update tab nav steps to use `isVisible` vs `clickable`

LevelbossMike:
### Description
Change tests that check visibility of tabs to not click, but only check for visibility via `isVisible`.

The tests changed in this PR accidentally clicked the tab because the page-object method being used was `clickable`, although we were only interested in checking for their visibility.

This should get rid of some recent occurred flakiness after upgrading ember-qunit et al.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


johncowen (approved):
👏 looks good to me! :shipit:

PR #14519, opened 2022-09-08
ui: Pass through default environment settings when when booting dev/test

johncowen:
### Description
Previous to this PR we had a way during dev or testing to enable/disable all sorts of feature flags via cookies, this PR gives those feature flags defaults based on what is in the corresponding Ember environment variables for the given environment. Note: This is only happens in non-production environments.

There are some tiny coincidental changes here due to HCPEnabled not having a default boolean value when not in production environments (`undefined` being falsey), in prod envs that boolean comes in from the backend.

This PR is required to move forwards with some additional peering testing, that will come in some follow up PRs

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


LevelbossMike (approved):
We should probably investigate how to handle this in a less complicated way. It feels like the environment variable setup we have currently touches many layers and is complicated to work with. This apparently comes from the custom startup we have in place. In a common ember app setup, this would just work™ based on `config/environment.js` that would either exist at build time or would get injected at runtime before application boot from the backend. 

Approving to unblock the acceptance test, but in my mind we should really investigate how we can improve this for the future.

PR #14518, opened 2022-09-08
ui: chore - upgrade ember and friends 

LevelbossMike:
### Description
WIP—will update description as I move along.

### PR Checklist

* [ ] updated test coverage
* [ ] external facing docs updated
* [x] not a security concern

PR #14520, opened 2022-09-08
ui: When peerings are enabled add a HTML.class to tell the UI CSS

johncowen:
### Description
Whilst working on some upgrades with @LevelbossMike we noticed that a HTML class was missing for when peerings was enabled. This adds that class so it is consistent with other larger feature flagged features and available if/when needed.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-ui/app/templates/application.hbs
+++ b/ui/packages/consul-ui/app/templates/application.hbs
@@ -16,6 +16,9 @@ as |route|>
 {{#if (can "use partitions")}}
   {{document-attrs class="has-partitions"}}
 {{/if}}
+{{#if (can "use peers")}}
+  {{document-attrs class="has-peerings"}}
+{{/if}}
 
 {{! Listen out for blocking query/client setting changes }}
 <DataSource


LevelbossMike (changes requested):
I don't see this class being used anywhere in the codebase. Why is this change necessary? I understand the need for consistency when we want to implement something similar, but when we don't need this class in the codebase currently, why would we want to add this?

johncowen:
> Why is this change necessary? I understand the need for consistency when we want to implement something similar

Well I suppose mostly the need for consistency 😄 , to be honest with you I thought we'd done this, but looked when we were talking about this helper offline and realized we hadn't.

Apart from the need for consistency, I can imagine future-me trying to do something, say for example hiding all peering related badges when peering is disabled with a CSS 1 liner using `html:not(.has-peerings)` and wondering why it doesn't work. I suppose now I've realised that we've not added this future-me would remember that (maybe?) and re-add this line from this PR when it was needed, so I'm happy to close this if you feel we shouldn't add it right now. Just let me know.

Actually, partly related to the comment I just made on this PR (https://github.com/hashicorp/consul/pull/14519#issuecomment-1243496501) - it might be an idea to add this capability/requirement to the more generic feature-flagging functionality so that we don't forget to add this class in the future - that way it would just happen without us even realizing.

Anyway, let me know if you want me to close this for now or not.

johncowen:
Closing this for the moment, we can circle back at a later date after https://github.com/hashicorp/consul/pull/14518 is in and the Big Bang lint is done

PR #14521, opened 2022-09-08
Reuse http.DefaultTransport in UIMetricsProxy

kisunji:
http.Transport keeps a pool of connections and should be reused when possible. We instantiate a new http.DefaultTransport for every metrics request, making large numbers of concurrent requests inefficiently spin up new connections instead of reusing open ones.

* No logic change
* No existing tests should fail

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/ui_endpoint.go
+++ b/agent/ui_endpoint.go
@@ -771,6 +771,7 @@ func (s *HTTPHandlers) UIMetricsProxy(resp http.ResponseWriter, req *http.Reques
 		Director: func(r *http.Request) {
 			r.URL = u
 		},
+		Transport: s.proxyTransport,
 		ErrorLog: log.StandardLogger(&hclog.StandardLoggerOptions{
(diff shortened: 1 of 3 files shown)


agent/ui_endpoint.go
```diff
@@ -771,6 +771,7 @@ func (s *HTTPHandlers) UIMetricsProxy(resp http.ResponseWriter, req *http.Reques
 		Director: func(r *http.Request) {
 			r.URL = u
 		},
+		Transport: s.proxyTransport,
```
kisunji, line 774: If nil, Transport defaults to `http.DefaultTransport`: 
https://github.com/golang/go/blob/go1.19.1/src/net/http/httputil/reverseproxy.go#L52-L54

DanStough (approved):
TIL 🚀 . Thanks for the great background links.

PR #14524, opened 2022-09-08
ui: Fix up tippy console warning...

johncowen:
...enabling/disabling now depends on whether the string is non-empty

### Description
Whilst building out work for peering testing I noticed a new console warning due to tippy warning us of an extra unsupported parameter. I decided it would be nice to disable the tooltip based on whether the string given to the tooltip is non-empty or not. I updated the documentation to include this fact also.

Note: There looks to be a lot of change in the tooltip modifier, but its just a removal of a conditional, plus the indenting change that came with that. The only change in this file is the return early type/emptiness check at the start.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-ui/app/components/peerings/badge/index.hbs
+++ b/ui/packages/consul-ui/app/components/peerings/badge/index.hbs
@@ -1,5 +1,5 @@
 {{#if @peering.State}}
-  <div class="peerings-badge {{lowercase @peering.State}}" {{tooltip this.tooltip options=(hash hideTooltip=(not this.tooltip))}}>
+  <div class="peerings-badge {{lowercase @peering.State}}" {{tooltip this.tooltip}}>
     <Peerings::Badge::Icon @state="{{@peering.State}}" />
     <span class="peerings-badge__text">{{capitalize (lowercase @peering.State)}}</span>
   </div>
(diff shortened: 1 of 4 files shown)


LevelbossMike (approved):

johncowen:
Thanks for the approval @LevelbossMike , I've actually just realised something and had a suggestion to make.

I'm currently looking to add test coverage to some of the beta peering work now we've validated most of the UX/API interaction for it, and I noticed (partly nudged by this PR https://github.com/hashicorp/consul/pull/14527). That the JS class here: 

https://github.com/hashicorp/consul/blob/a43fd09c8d5c67c45027fda0dcb4dbfe467a42c6/ui/packages/consul-ui/app/components/peerings/badge/index.js

..could instead be replaced with the ember-intl `t` helper and we move the copy into the translation files along with all the other copy for the application.

Benefits in doing that are:

1. We would no longer need to write tests for this JS class (we probably don't want to test that the `t` helper works)
2. We are using a recognised ember addon/pattern that we are already using in the codebase to achieve the same thing (string based key/value lookup)

Anyway, let me know whether you like that suggestion or not (change to use the t helper), or whether you'd prefer that we left that for a later date and I add some tests to this during this round of work.

LevelbossMike:
@johncowen sounds like a reasonable suggestion to me. Before the lookup table also bundled icon names. Now where it only encapsulates the tooltip text, I think moving to `t` would be even better.

PR #14527, opened 2022-09-08
ui: Amends to existing topology notice/banner texts

johncowen:
### Description
We've decided to make some text and logic amends to the topology notices/banners. This PR addresses the text changes only, partly because the logic changes are likely to affect line styling (dotted vs non-dotted) and other things in the topology visualisation itself.

Unchanged Texts:

- No Dependencies
- Limited Access

Additional texts ready for usage with to come logic.

- No intentions

All other texts apart from the above have been changed, please let me know if the unchanged texts do need changing.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-ui/translations/routes/en-us.yaml
+++ b/ui/packages/consul-ui/translations/routes/en-us.yaml
@@ -185,25 +185,25 @@ dc:
             header: Limited Access
             body: This service may have dependencies you won’t see because you don’t have access to them.
           default-allow:
-            header: Intentions are set to default allow
-            body: Your Intention settings are currently set to default allow. This means that this view will show connections to every service in your cluster. We recommend changing your Intention settings to default deny and creating specific Intentions for upstream and downstream services for this view to be useful.
+            header: Restrict which services can connect
+            body: Your current ACL settings allow all services to connect to each other. Either create a deny intention between all services, or set your default ACL policy to deny to improve your security posture and make this topology view reflect the actual upstreams and downstreams of this service.
             footer: |
               <p>
-                <a href="{route_intentions}">Edit Intentions</a>
+                <a href="{route_intentions}">Create a wildcard deny Intention</a>
               </p>
           wildcard-intention:
-            header: Permissive Intention
-            body: One or more of your Intentions are set to allow traffic to and/or from all other services in a namespace. This Topology view will show all of those connections if that remains unchanged. We recommend setting more specific Intentions for upstream and downstream services to make this visualization more useful.
+            header: Restrict which services can connect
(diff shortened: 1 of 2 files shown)


LevelbossMike:
Taking myself off review as I'm not aware what the exact texts should be.

ui/packages/consul-ui/translations/routes/en-us.yaml
```diff
@@ -213,12 +213,19 @@ dc:
                 <a href="{CONSUL_DOCS_URL}/connect/registration/service-registration#upstream-configuration-reference" target="_blank" rel="noopener noreferrer">Read the documentation</a>
               </p>
           acls-disabled:
-            header: Enable ACLs
-            body: This connect-native service may have dependencies, but Consul isn't aware of them when ACLs are disabled. Enable ACLs to make this view more useful.
+            header: All service-to-service connections are allowed
```
jkirschner-hashicorp, line 216: @jnwright , @johncowen : Should this be consistent with the other "Restrict which services can connect" headers?

```suggestion
            header: Restrict which services can connect
```

jnwright: Hm, I'm not sure. The title you and I chose during our sessions is the one John wrote. I remember us wanting to differentiate it for a reason, but do you think it should be consistent?

jkirschner-hashicorp: I vaguely recall such a conversation, but not what the outcome/decision was!

The two messages are basically the same, just with a different focus. "Restrict which services can connect" focuses on what the user should do. "All service-to-service connections are allowed" focuses on communicating current behavior. I'm not sure why we'd frame the same message in two different ways depending on whether ACLs are disabled vs enabled with default allow. ... Maybe we did have a reason when we last spoke about it, but I can't remember if it we did!

jnwright: I'm fine with changing it to be the same title ¯\_(ツ)_/¯

johncowen: What was the outcome here? Shall I go ahead and commit the suggestion or leave as is?

jnwright: yeah, I say we go ahead and change it! And that'll wrap this up 😎

johncowen: Perfect, k thats done, just needs a 👍 /approval on this one then and we can merge

WenInCode (approved):

jnwright (approved):

PR #14540, opened 2022-09-08
Add endpoint to query peering health

kisunji:
### Description
The UI needs an endpoint to query for a given peering stream's health.

Originally I thought it should be internal but wondering if it's valuable to expose publicly since we have no API for that.

Request:
```json
{
  "PeerName": "foo",
  "Partition": "ap1"
}
```

Response:
```json
{
  "State": <currently string form of our enums>
  "LastHeartBeat": <timestamp>
  "LastReceived": <timestamp>
  "LastSent": <timestamp>
}
```

@johncowen I wasn't sure if it was easier for frontend to parse our int-based enums or use strings. Currently we output strings based on https://github.com/hashicorp/consul/blob/09c0fe22f22c274e2005659288c0698b1536532e/proto/pbpeering/peering.pb.go#L53-L59

* Added test cases

### PR Checklist

* [x] updated test coverage
* [ ] external facing docs updated <-TODO
* [x] not a security concern

### DIFF
--- a/agent/consul/internal_endpoint.go
+++ b/agent/consul/internal_endpoint.go
@@ -2,8 +2,9 @@ package consul
 
 import (
 	"fmt"
+	"time"
 
-	bexpr "github.com/hashicorp/go-bexpr"
+	"github.com/hashicorp/go-bexpr"
 	"github.com/hashicorp/go-hclog"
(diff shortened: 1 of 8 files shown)


kisunji:
Closing in favor of incorporating these fields to the Peering object itself

PR #14057, opened 2022-08-08
Move sidecar port assignment logic to locked stage of add flow

skpratt:
Closes #8254

### Description
This relocates automatic sidecar port assignment logic to a locked method immediately before registration, thereby preventing collisions when handling simultaneous requests.

### Testing & Reproduction steps
Manual verification:
* Submit many simultaneous requests for registration with sidecar proxy and no port specified:
`for name in $(shuf -i 1-10000); do curl -d '{"name": "'$name'", "connect": {"sidecar_service": {}}}' -X PUT localhost:8500/v1/agent/service/register &; echo registered $name; done`
* See how many duplicate ports have been registered:
`curl localhost:8500/v1/agent/services | jq -r ' . | to_entries | map(select(.key | match("sidecar";"i"))) | map(.value) | .[] | .Port' | sort | uniq -d`
* Before this change: there should be several duplicate ports
* After this change: no duplicate ports

### Links
[GH Issue](https://github.com/hashicorp/consul/issues/8254)
[Pre-req PR 1](https://github.com/hashicorp/consul/pull/13906)
[Pre-req PR 2](https://github.com/hashicorp/consul/pull/14056)

### PR Checklist

* [ ] updated test coverage
* [N/A] external facing docs updated
* [N/A] not a security concern

PR #14541, opened 2022-09-09
PR #14057 follow up fix: service id parsing from sidecar id

skpratt:
Follow up to PR #14057

### Description
PR #14057 introduced a bug where a function to extract the serviceID from the sidecarServiceID did not precisely undo the append operation that was used to generate the sidecarServiceID. Submitting a sidecar registration for a service with a hyphen in the ID string creates an incorrect default alias check. This PR changes the serviceIDFromSidecarID function to exactly undo the function as intended.

* Register a service with a hyphenated name, accompanied by the unexpanded default sidecar syntax. The default sidecar alias check should correctly identify the service and pass (previously would fail).

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/sidecar_service.go
+++ b/agent/sidecar_service.go
@@ -10,31 +10,15 @@ import (
 	"github.com/hashicorp/consul/agent/structs"
 )
 
-const sidecarIDSuffix string = "-sidecar-proxy"
+const sidecarIDSuffix = "-sidecar-proxy"
 
 func sidecarIDFromServiceID(serviceID string) string {
 	return serviceID + sidecarIDSuffix
 }
 
 // reverses the sidecarIDFromServiceID operation
-func serviceIDFromSidecarID(sidecarServiceID string) string {
-	// we only want to un-append "-sidecar-proxy" from the very end
-	// so we reverse the string, remove only the first suffix match
-	// and then reverse the string again to get the result
-	revID := ""
-	for _, v := range sidecarServiceID {
-		revID = string(v) + revID
-	}
-	revSuffix := ""
-	for _, v := range sidecarIDSuffix {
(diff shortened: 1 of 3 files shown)


agent/sidecar_service.go
```diff
@@ -10,12 +10,31 @@ import (
 	"github.com/hashicorp/consul/agent/structs"
 )
 
-func sidecarServiceID(serviceID string) string {
-	return serviceID + "-sidecar-proxy"
+const sidecarIDSuffix string = "-sidecar-proxy"
```
kisunji, line 13:
```suggestion
const sidecarIDSuffix = "-sidecar-proxy"
```

agent/sidecar_service.go
```diff
@@ -10,12 +10,31 @@ import (
...
+
+func sidecarIDFromServiceID(serviceID string) string {
+	return serviceID + sidecarIDSuffix
 }
 
+// reverses the sidecarIDFromServiceID operation
 func serviceIDFromSidecarID(sidecarServiceID string) string {
-	return strings.Split(sidecarServiceID, "-")[0]
+	// we only want to un-append "-sidecar-proxy" from the very end
```
kisunji, line 21: I think we can simply call `strings.TrimSuffix` here, right? https://pkg.go.dev/strings#TrimSuffix

agent/sidecar_service.go
kisunji: Since go infers type here we conventionally leave out `string`

agent/sidecar_service.go
skpratt: great find! thank you! that makes it much cleaner

kisunji (approved):
LGTM 👍

PR #14554, opened 2022-09-09
[docs] use values.yaml instead of config.yaml everywhere for helm configurations

kschoche:
### Description
We have been using `values.yaml` and `config.yaml` interchangeably in our docs to describe the custom helm configurations.
This PR proposes that we use `values.yaml` to describe the helm chart configurations and use `config.yaml` in existing places where it is meant to describe other objects (like CRDs, etc).

This aligns our docs with whats in the consul-k8s repo and also what is described here: https://helm.sh/docs/chart_template_guide/values_files/

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


david-yu (approved):

im2nguyen (approved):

david-yu:
Not sure why this is hanging, perhaps try pushing an empty commit?

PR #14556, opened 2022-09-09
Add handling in agent cache for server leaf certs

freddygv:
### Description
- Adds a new internally generated and locally managed ACL token for servers to use for management actions. This new ACL token has a corresponding ACL identity and unlimited permissions given that it is generated for server agents.

- Adds a new server certificate manager for requesting a certificate for servers and updating the TLS configurator as needed.

* Manual testing
* Unit testing with multiple test agents

### Links
https://go.hashi.co/csl-204-server-cert

### PR Checklist

* [x] updated test coverage
* [ ] not a security concern


freddygv:
Hmm will take a look at these failures on Monday

agent/structs/acl.go
```diff
@@ -1838,3 +1839,51 @@ func (id *AgentRecoveryTokenIdentity) IsLocal() bool {
 func (id *AgentRecoveryTokenIdentity) EnterpriseMetadata() *acl.EnterpriseMeta {
 	return nil
 }
+
+const ServerManagementToken = "server-management-token"
```
boxofrad, line 1843: This is the Accessor ID, not the Secret ID right? For clarity, I wonder if we should call this something like `ServerManagementTokenAccessorID`?

tlsutil/config.go
```diff
@@ -199,13 +199,15 @@ type Configurator struct {
...
-	// auto-config features.
+	// autoTLS stores configuration that is received from:
+	// - The auto-encrypt or auto-config features for client agents
+	// - The servercert.CertManager for server agents.
 	autoTLS struct {
 		extraCAPems          []string
 		connectCAPems        []string
 		cert                 *tls.Certificate
 		verifyServerHostname bool
+		peeringServerName    string
```
boxofrad, line 210: nit, please feel free to ignore:

What do you think of calling this `serverName`? (and the method below `UpdateAutoTLSServerName`).

From the TLS configurator's point of view, there's nothing peering-specific about this. It's just that sometimes the automatically provisioned certificate has an associated server name, which (when it matches the client's request) causes it to be preferred over the manually provisioned one.

boxofrad (changes requested):
This looks great, nice job! 👏🏻  I really enjoyed getting this review request because it nudged me to read the RFC and learn about this feature.

I've left a bunch of mostly non-blocking comments. The only one I think needs to be addressed is how we're hanging onto stale state stores.

agent/cache-types/connect_ca_leaf.go
```diff
@@ -679,6 +693,7 @@ type ConnectCALeafRequest struct {
 	Service        string              // Service name, not ID
 	Agent          string              // Agent name, not ID
 	Kind           structs.ServiceKind // only mesh-gateway for now
+	Server         bool
```
kisunji, line 696: Can we leave a short doc on what fields are mutually exclusive? It's not immediately obvious here without looking at the switch-case

agent/consul/servercert/manager.go
```diff
@@ -0,0 +1,261 @@
...
+			return nil
+		}
+		if token.Value == "" {
+			// This should never happen. If the leader stored a token with this key it will not be empty.
+			return fmt.Errorf("empty token")
+		}
+		m.logger.Debug("server management token watch fired - resetting leaf cert watch")
+
+		// Cancel existing the leaf cert watch and spin up new one any time the server token changes.
+		// We need a valid token that the leader server recognizes, and when the
```
kisunji, line 143: unfinished comment

agent/consul/servercert/manager.go
```diff
@@ -0,0 +1,261 @@
...
+func (m *CertManager) watchServerToken(ctx context.Context) {
+	// We keep the last iteration's cancel function to reset watches.
+	var (
+		notifyCtx context.Context
+		cancel    context.CancelFunc = func() {}
+	)
+	retryLoopBackoff(ctx, func() error {
+		ws := memdb.NewWatchSet()
+		ws.Add(m.store.AbandonCh())
+		ws.Add(ctx.Done())
```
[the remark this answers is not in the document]
kisunji: Adding it here lets `SystemMetadataGet` wake on a Close if I'm reading correctly

agent/consul/servercert/manager.go
```diff
@@ -0,0 +1,261 @@
...
+		if token.Value == "" {
+			// This should never happen. If the leader stored a token with this key it will not be empty.
+			return fmt.Errorf("empty token")
+		}
+		m.logger.Debug("server management token watch fired - resetting leaf cert watch")
+
+		// Cancel existing the leaf cert watch and spin up new one any time the server token changes.
+		// We need a valid token that the leader server recognizes, and when the
+		cancel()
+		notifyCtx, cancel = context.WithCancel(ctx)
```
kisunji, line 145: Do we want to be using `WithCancel` here? If parent `ctx` is cancelled the `cache.Notify` will be cancelled as well; IIUC we want `notifyCtx` to be completely independent of the parent `ctx` in this retry loop.

agent/agent.go
```diff
@@ -583,6 +587,24 @@ func (a *Agent) Start(ctx context.Context) error {
...
+
+		if a.config.PeeringEnabled && a.config.ConnectEnabled {
+			d := servercert.Deps{
+				Logger: a.logger.Named("server.cert-manager"),
+				Config: servercert.Config{
+					Datacenter:  a.config.Datacenter,
+					ACLsEnabled: a.config.ACLsEnabled,
+				},
+				Cache:           a.cache,
+				Store:           server.FSM().State(),
```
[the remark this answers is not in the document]
freddygv: Oh good to know! Fixed.

tlsutil/config.go
freddygv: I'm OK with switching it to `serverName`.

My thinking was that there may be additional server names to handle this way in the future (maybe for agentless?)

agent/consul/leader.go
```diff
@@ -527,6 +528,20 @@ func (s *Server) initializeACLs(ctx context.Context) error {
 			}
 			s.logger.Info("Created ACL anonymous token from configuration")
 		}
+
+		// Generate or rotate the server management token on leadership transitions.
+		// This token is used by Consul servers for authn/authz when making
+		// requests to themselves through public APIs such as the agent cache.
+		// It is stored as system metadata because it is internally
+		// managed and users are not meant to see it or interact with it.
```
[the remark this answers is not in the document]
freddygv: I think something like that would work if this was for purely local operations, but an issue is that the `Sign` requests are sent to the leader over the network

agent/consul/servercert/manager.go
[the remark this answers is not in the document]
freddygv: Ah yep we don't need this. Removed.

agent/cache-types/connect_ca_leaf.go
freddygv: Yep, done

agent/consul/servercert/manager.go
freddygv: Thanks!

agent/consul/servercert/manager.go
```diff
@@ -0,0 +1,261 @@
...
+
+	for {
+		if err := waiter.Wait(ctx); err != nil {
+			// The error will only be non-nil if the context is canceled.
+			return
+		}
+
+		if err := loopFn(); err != nil {
+			errorFn(err)
+		}
```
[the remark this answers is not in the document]
freddygv: Nice catch! Updated the docs as well.

freddygv:
Err the TLS configurator side needs more thinking. I've been trying to fix the test failures but there is a legitimate issue here:

When an autoTLS cert is present we ALWAYS use it as the outbound cert. However, for WANFed via mesh gateways the secondary DCs can't validate the autoTLS server certificates from servers in the primary because the secondaries don't have the Connect CA roots until federation succeeds. This wasn't an issue before because servers never had an autoTLS cert configured.

For server agents we shouldn't use the autoTLS certs for any outbound connections other than peering. But the configurator currently isn't aware of whether it's attached to a server or client agent. 

Defaulting to using the manually configured certs for outbound connections would work, but it's a breaking change. To be honest I don't know why we default to autoTLS when a manual cert exists. Seems kind of odd. Also the auto_config docs say `These configurations get merged in as defaults with any user-supplied configuration on the client agent able to override them.` 

Thoughts @boxofrad ?

agent/consul/leader.go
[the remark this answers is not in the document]
boxofrad: Ah yes, I'd missed that we were forwarding `Sign` requests to the leader. Never mind!

tlsutil/config.go
boxofrad: Oh, maybe! For agentless, right now we're imagining folks will use "real" TLS certs, signed by a public CA, so they don't need to configure roots on the "client".

boxofrad:
Ah yes, I remember staring at this for a long while when we switched to a per-listener TLS configuration. 😅 

I'm not 100%, but maybe @i0rek remembers? why we always prefer the auto-encrypt/config certificate in outgoing connections. If I had to hazard a guess, I'd say it's because if you're using one of those features, only their automatically provisioned client certificates (signed by the Connect CA) are accepted by the server, not certificates created by `consul tls cert create`.

I think it'd be ok to break backwards compatibility here _just_ for the gRPC port, as long as we leave the server/multiplexed port alone. Only peering makes outgoing connections on that port, and we've[ told folks using auto-{encrypt,config} not to upgrade to 1.13 yet](https://www.consul.io/docs/upgrading/upgrade-specific#service-mesh-deployments-using-auto-encrypt-or-auto-config), so I don't think it'd impact anyone.

boxofrad (approved):

PR #14563, opened 2022-09-09
[OSS] Validate Peering token Locality

DanStough:
### Description
Adding some additional validation when establishing a peering connection as part of enhancements around mesh gateways for peering control plane traffic. The proposed changes look for a matching server name in the token, but an unknown PeerID, which indicates this token is from another cluster with the same datacenter name. This is unsupported.

* Unit tests + integration tests

### PR Checklist

* [X] updated test coverage
* [X] not a security concern

### DIFF
--- a/agent/rpc/peering/service.go
+++ b/agent/rpc/peering/service.go
@@ -194,8 +194,6 @@ func (s *Server) GenerateToken(
 		return nil, fmt.Errorf("meta tags failed validation: %w", err)
 	}
 
-	defer metrics.MeasureSince([]string{"peering", "generate_token"}, time.Now())
-
 	resp := &pbpeering.GenerateTokenResponse{}
 	handled, err := s.ForwardRPC(&writeRequest, func(conn *grpc.ClientConn) error {
 		ctx := external.ForwardMetadataContext(ctx)
@@ -206,6 +204,8 @@ func (s *Server) GenerateToken(
 	if handled || err != nil {
 		return resp, err
 	}
+
+	defer metrics.MeasureSince([]string{"peering", "generate_token"}, time.Now())
 
 	var authzCtx acl.AuthorizerContext
 	entMeta := structs.DefaultEnterpriseMetaInPartition(req.Partition)
--- a/agent/rpc/peering/service.go
+++ b/agent/rpc/peering/service.go
@@ -374,7 +374,7 @@ func (s *Server) Establish(
 		return nil, err
 	}
 
-	if err := s.validatePeeringInPartition(tok.PeerID, entMeta.PartitionOrEmpty()); err != nil {
+	if err := s.validatePeeringLocality(tok, entMeta.PartitionOrEmpty()); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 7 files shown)


agent/rpc/peering/service.go
```diff
@@ -463,15 +463,23 @@ func (s *Server) Establish(
 	return resp, nil
 }
 
-// validatePeeringInPartition makes sure that we don't create a peering in the same partition. We validate by looking at
-// the remotePeerID from the PeeringToken and looking up for a peering in the partition. If there is one and the
-// request partition is the same, then we are attempting to peer within the partition, which we shouldn't.
-func (s *Server) validatePeeringInPartition(remotePeerID, partition string) error {
-	_, peering, err := s.Backend.Store().PeeringReadByID(nil, remotePeerID)
+// validatePeeringLocality makes sure that we don't create a peering in the cluster/partition it was generated.
+// We validate by looking at the remotePeerID from the PeeringToken and looking up for a peering in the partition.
```
kisunji, line 467:
```suggestion
// We validate by looking at the remote PeerID from the PeeringToken and looking up that peering in the partition.
```

agent/rpc/peering/service.go
```diff
@@ -463,15 +463,23 @@ func (s *Server) Establish(
...
+// If there is one and the request partition is the same, then we are attempting to peer within the partition, which we shouldn't.
+// We also perform a check to verify if the ServerName of the PeeringToken overlaps with our own, we do not process it
+// unless we've been able to find the peering in the store, i.e. this peering is between two local partitions.
+func (s *Server) validatePeeringLocality(token *structs.PeeringToken, partition string) error {
+	_, peering, err := s.Backend.Store().PeeringReadByID(nil, token.PeerID)
 	if err != nil {
 		return fmt.Errorf("cannot read peering by ID: %w", err)
 	}
 
+	// If the token has the same server name as thing cluster, but we can't find the peering
```
kisunji, line 477:
```suggestion
	// If the token has the same server name as this cluster, but we can't find the peering
```

agent/rpc/peering/service.go
```diff
@@ -463,15 +463,23 @@ func (s *Server) Establish(
...
+	// in our store, it indicates a naming conflict.
+	if s.Backend.GetServerName() == token.ServerName && peering == nil {
+		return fmt.Errorf("conflict - peering token \"ServerName\" matches the current cluster but was not generated here")
+	}
```
kisunji, line 481: We could also hit this scenario if an operator deletes a peering after generating a token, but later tries to establish within the same cluster with that token.

kisunji: We could also expose the server name with `%q` for easier debugging in the log message

agent/rpc/peering/service.go
kisunji: I know this wasn't your text but just minor improvements

kisunji (commented):
Minor comment about the error message if server names are equal but peering does not exist. Maybe a more general message is required? Or something like "peering may have been deleted or the token was generated in a cluster with the same server name"

agent/rpc/peering/service.go
DanStough: Hmm, I see your point. I don't really like the either-or message because we should be able to point operators to the real problem, but after staring at this for a while I don't know a better way. I also don't want to optimize too much for the same-cluster case (which seems like the degenerate case). How about tweaking it just a bit:
`conflict - peering token \"ServerName\" matches the current cluster \"ServerName\", %q, but there is no record in the database`

kisunji (approved):

agent/rpc/peering/service.go
kisunji: Yeah it sounds more general 👍 

Styling-wise I prefer keeping server name as a normal noun (e.g. peering token's server name does not match the current cluster's server name) but not a blocker

DanStough: Normal-case works for me. 🚀

PR #14574, opened 2022-09-13
ui: Remove ember-cli-uglify config

johncowen:
### Description
In https://github.com/hashicorp/consul/pull/14518 we have upgrade ember to 3.24, this included switching from uglify to terser for JS minification etc, so we need to remove this config to prevent the following warning now that the config doesn't do anything:

Screenshot 2022-09-13 at 10 49 37

I had a quick look to see if there is an equivalent config for terser and there doesn't seem to be.

Note the base branch for this is for https://github.com/hashicorp/consul/pull/14518, I know @LevelbossMike is AFK so I thought it would be best to branch off and make this tiny amend here.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-ui/ember-cli-build.js
+++ b/ui/packages/consul-ui/ember-cli-build.js
@@ -161,13 +161,6 @@ module.exports = function (defaults, $ = process.env) {
           'mode/loadmode.js',
         ],
       },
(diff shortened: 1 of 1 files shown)


evrowe (approved):
:shipit:

PR #14585, opened 2022-09-13
Add CSR check for number of URIs.

hashi-derek:
This is a backport of: https://github.com/hashicorp/consul/pull/14579

It needed to be stripped down significantly due to version differences.

### DIFF
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -1397,6 +1397,15 @@ func (c *CAManager) SignCertificate(csr *x509.CertificateRequest, spiffeID conne
 		return nil, fmt.Errorf("CA is uninitialized and unable to sign certificates yet: no root certificate")
 	}
 
+	// Note that only one spiffe id is allowed currently. If more than one is desired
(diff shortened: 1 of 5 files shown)


agent/consul/leader_connect_ca.go
```diff
@@ -1397,6 +1397,15 @@ func (c *CAManager) SignCertificate(csr *x509.CertificateRequest, spiffeID conne
 		return nil, fmt.Errorf("CA is uninitialized and unable to sign certificates yet: no root certificate")
 	}
 
+	// Note that only one spiffe id is allowed currently. If more than one is desired
+	// in future implmentations, then each ID should have authorization checks.
+	if len(csr.URIs) != 1 {
```
kyhavlov, line 1402: Is it worth porting the same change to `auto_config_endpoint.go` like in the original PR? That code still exists in 1.11 and looks like it has the same issue: https://github.com/hashicorp/consul/blob/v1.11.8/agent/consul/auto_config_endpoint.go#L388

kyhavlov (approved):

PR #14598, opened 2022-09-13
connect/ca: Don't discard old roots on primaryInitialize

kyhavlov:
There are a couple cases in `primaryInitialize` in the `CAManager` that can cause the root to need updating when the leader is starting up. Currently when this happens, the old roots are discarded and only the new root is preserved, which could cause problems if the root was just rotated out and we still need to keep the old root around for interoperability during the switch to the new root. 

This PR changes that behavior to keep the old roots around in this case in `primaryInitialize` - they'll still be pruned through the normal leader routine once enough time has passed after they've been rotated out.

### DIFF
--- a/.changelog/14598.txt
+++ b/.changelog/14598.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+connect: Fixed a bug where old root CAs would be removed from the primary datacenter after switching providers and restarting the cluster.
+```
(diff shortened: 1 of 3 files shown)


dhiaayachi (approved):
LGTM!
Is this fix meant to be backported?

Jason-123-cyber (commented):
JJ

Jason-123-cyber:
M

Jason-123-cyber (commented):
j

Jason-123-cyber:
b

Jason-123-cyber:
jl

Issue #12553, opened 2022-03-14
Support Centralized Proxy Configuration in `service-defaults`


mkeeler:
#### Feature Description

Today you can configure [Envoy proxy configuration](https://www.consul.io/docs/connect/proxies/envoy#proxy-config-options) both within a service instance and in the `proxy-defaults/global` configuration entry. The `protocol` configuration can also be set within the `service-defaults` configuration entry. Other configurations like `local_request_timeout_ms` and `local_connect_timeout_ms` cannot be set here.

The feature request is to support setting some of these configurations within the `service-defaults` config entry. However some others such as the `bind_address` probably don't make sense to set centrally so its not quite so simple as just merge an extra config source.

#### Use Case(s)

I encountered a gRPC API where the local app could take several minutes to return data for a specific API. I had wanted to set `local_request_timeout_ms` for all instances of the service but it wasn't possible. I had to resort to configuring by modifying all instances of the service.

huikang:
Hi, @mkeeler ,  I am interested in implementing this feature and would like to give it a shot.
Could you help clarify the following questions regarding the new feature?

1. In the definition of service-defaults, there is a similar instance-level field, named [MaxInboundConnections](https://github.com/hashicorp/consul/blob/main/agent/structs/config_entry.go#L110). It seems to me that the `local_request_timeout_ms` and `local_connect_timeout_ms` should be in the same level as [MaxInboundConnections](https://github.com/hashicorp/consul/blob/main/agent/structs/config_entry.go#L110). So the modified struct will look like

```go
type ServiceConfigEntry struct {
    // omitted fields
    MaxInboundConnections int
+    LocalConnectTimeOut int
+   LocalRequestTimeOut int
}
```
What do you think?

2. It seems that most of the [dynamic configurations](https://www.consul.io/docs/connect/proxies/envoy#dynamic-configuration) has already been added to the service-default. Is there any other fields we need to consider with these 2?

Thanks in advance.

PR #14639, opened 2022-09-15
feat: propogate local_request_timeout_ms to downstream

huikang:
### Description
**Situation**: "_If an application is configured with a local_request_timeout_ms value, would it be a good idea to auto-propagate this timeout to consumers/downstreams of that service. That way we don't have to configure the timeout in 2 different places. Maybe downstreams would want to timeout sooner. In that case they could specify the timeout and this would just be a way to modify the default timeout value._"

- The implementation of this PR is to loop over the upstream proxy services to find the smallest `local_request_timeout_ms`. If the value is smaller than the default, will propagate to the downstream proxy service.
    - call `makeInboundListener` after looping all the upstream

- doc is updated
- unit test is added: to make the upstream service to have its own `local_request_timeout_ms`, a new parameter is used to configure the proxy.Config of the upstream proxy service.

### Links
Address https://github.com/hashicorp/consul/issues/12553

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/.changelog/14639.txt
+++ b/.changelog/14639.txt
@@ -1,0 +1,4 @@
+```release-note:feature
+proxycfg: progagate `local_request_timeout_ms` downstream, if its value of the upstream is smaller than
+that of the downstream
+```
--- a/agent/proxycfg/manager.go
+++ b/agent/proxycfg/manager.go
@@ -148,7 +148,7 @@ func (m *Manager) Register(id ProxyID, ns *structs.NodeService, source ProxySour
 		return err
 	}
 
-	if _, err = state.Watch(); err != nil {
+	if err = state.Watch(); err != nil {
 		return err
 	}
 	m.proxies[id] = state
--- a/agent/proxycfg/manager_test.go
+++ b/agent/proxycfg/manager_test.go
@@ -192,7 +192,7 @@ func TestManager_BasicLifecycle(t *testing.T) {
 			setup: func(t *testing.T, dataSources *TestDataSources) {
 				// Note that we deliberately leave the 'geo-cache' prepared query to time out
 				dataSources.Health.Set(dbHealthReq, &structs.IndexedCheckServiceNodes{
-					Nodes: TestUpstreamNodes(t, db.Name),
+					Nodes: TestUpstreamNodes(t, db.Name, structs.ConnectProxyConfig{}),
 				})
 				dataSources.CompiledDiscoveryChain.Set(dbChainReq, &structs.DiscoveryChainResponse{
 					Chain: dbDefaultChain(),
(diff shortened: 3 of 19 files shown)


mkeeler (commented):
I think it would be good for someone else to take a second look here so I have tagged the consul service mesh team.

My hesitation is that this seems like it could introduce issues with requests to upstreams that have higher timeouts than others.

Take the following scenario:

```mermaid
flowchart TD
   %% Basic States
   otherServices(Other Services)
   subgraph foo
      fooService(foo)
      subgraph fooSidecar [foo sidecar proxy]
         direction TB
         fooInbound(Inbound Listener)
         fooBarUpstream(Bar Upstream)
         fooBazUpstream(Baz Upstream)
      end
   end
   
   subgraph bar
      direction LR
      barService(bar)
      subgraph barSidecar [bar sidecar proxy]
         barInbound(Inbound Listener)
      end
   end
   
   subgraph baz
      direction LR
      bazService(baz)
      subgraph bazSidecar [baz sidecar proxy]
         direction LR         
         bazInbound(Inbound Listener)
      end
   end
   %% classDef default fill:#f9decd,stroke:#e06e2d,stroke-width:4px,color:black;

 %% State Transitions
   otherServices-- Timeout: 5s  -->fooInbound;
   fooInbound-->fooService;
   fooService-->fooBarUpstream;
   fooService-->fooBazUpstream;
   fooBarUpstream-->barInbound;
   fooBazUpstream-->bazInbound;
   barInbound-- Timeout: 60s -->barService
   bazInbound-- Timeout: 5s -->bazService
```

Here there are 2 upstreams and the min request timeout for them to their local app is 5s. I think this means that for the flow where some service makes a request to foo which needs to make a request to bar, the timeout may be too short and kill the connections too early.

Ideally we would love the diagram to look more like the following:

```mermaid
flowchart TD
   %% Basic States
   otherServices(Other Services)
   subgraph foo
      fooService(foo)
      subgraph fooSidecar [foo sidecar proxy]
         direction TB
         fooInbound(Inbound Listener)
         fooBarUpstream(Bar Upstream)
         fooBazUpstream(Baz Upstream)
      end
   end
   
   subgraph bar
      direction LR
      barService(bar)
      subgraph barSidecar [bar sidecar proxy]
         barInbound(Inbound Listener)
      end
   end
   
   subgraph baz
      direction LR
      bazService(baz)
      subgraph bazSidecar [baz sidecar proxy]
         direction LR         
         bazInbound(Inbound Listener)
      end
   end

 %% State Transitions
   otherServices-- Timeout: none -->fooInbound;
   fooInbound-->fooService;
   fooService-- Timeout: 60s -->fooBarUpstream;
   fooService-- Timeout: 5s -->fooBazUpstream;
   fooBarUpstream-- Timeout: 60s -->barInbound;
   fooBazUpstream-- Timeout: 5s -->bazInbound;
   barInbound-- Timeout: 60s -->barService
   bazInbound-- Timeout: 5s -->bazService
```

Here the timeouts are configured per-upstream so requests from the `foo` app of the `bar` service will use the `bar` services local app timeout.

PR #14644, opened 2022-09-15
Allow setting TLS for gRPC with deprecated options

freddygv:
### Description
Currently TLS for gRPC can only be enabled using the options nested in the tls.grpc configuration stanza.

That leads to a breaking change where the TLS options deprecated in 1.12 cannot be used to enable TLS for gRPC.

This commit updates the logic for determining whether TLS should be used on the public gRPC port: If the 1.12 tls stanza is not specified we default to the original behavior, which is to enable TLS for gRPC if the HTTPS port is set.

The change allows for consul-k8s to continue to use TLS flags compatible with 1.11 until 1.14 is released.

* Unit tests
* Manual tests:
  * Given multiple config files where some use the TLS stanza and others don't --> used post 1.13 logic to set TLS since at least one specified the new TLS stanza.
  * Given multiple config files where TLS certs were specified with deprecated flags and the HTTPS port was set --> used old logic and enabled TLS

### Links
Related to:
- #14253
- #13733

### PR Checklist

* [x] updated test coverage
* [ ] not a security concern

### DIFF
--- a/agent/config/builder.go
+++ b/agent/config/builder.go
@@ -189,9 +189,10 @@ func newBuilder(opts LoadOpts) (*builder, error) {
 	b.Tail = append(b.Tail, LiteralSource{Name: "flags.values", Config: values})
 	for i, s := range opts.HCL {
 		b.Tail = append(b.Tail, FileSource{
-			Name:   fmt.Sprintf("flags-%d.hcl", i),
-			Format: "hcl",
-			Data:   s,
+			Name:     fmt.Sprintf("flags-%d.hcl", i),
+			Format:   "hcl",
+			Data:     s,
+			FromUser: true,
 		})
 	}
 	b.Tail = append(b.Tail, NonUserSource(), DefaultConsulSource(), OverrideEnterpriseSource(), defaultVersionSource())
@@ -282,7 +283,7 @@ func newSourceFromFile(path string, format string) (Source, error) {
 	if format == "" {
 		format = formatFromFileExtension(path)
 	}
-	return FileSource{Name: path, Data: string(data), Format: format}, nil
+	return FileSource{Name: path, Data: string(data), Format: format, FromUser: true}, nil
 }
 
 // shouldParse file determines whether the file to be read is of a supported extension
@@ -2521,6 +2522,12 @@ func validateAbsoluteURLPath(p string) error {
 
 func (b *builder) buildTLSConfig(rt RuntimeConfig, t TLS) (tlsutil.Config, error) {
 	var c tlsutil.Config
+
+	// This flag indicates whether any of the per-listener configuration was set.
+	// The flag was populated earlier when applying deprecated options.
+	if t.SpecifiedTLSStanza != nil {
+		c.SpecifiedTLSStanza = *t.SpecifiedTLSStanza
+	}
 
 	// Consul makes no outgoing connections to the public gRPC port (internal gRPC
 	// traffic goes through the multiplexed internal RPC port) so return an error
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -3,6 +3,7 @@ package config
 import (
 	"encoding/json"
 	"fmt"
+	"reflect"
 	"time"
 
 	"github.com/hashicorp/consul/agent/consul"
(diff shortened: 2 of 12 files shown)


agent/config/builder.go
```diff
@@ -189,9 +189,10 @@ func newBuilder(opts LoadOpts) (*builder, error) {
 	b.Tail = append(b.Tail, LiteralSource{Name: "flags.values", Config: values})
 	for i, s := range opts.HCL {
 		b.Tail = append(b.Tail, FileSource{
-			Name:   fmt.Sprintf("flags-%d.hcl", i),
-			Format: "hcl",
-			Data:   s,
+			Name:     fmt.Sprintf("flags-%d.hcl", i),
+			Format:   "hcl",
+			Data:     s,
+			FromUser: true,
```
freddygv, line 195: This is for when users specify `-hcl` via the CLI.

agent/config/builder.go
```diff
@@ -282,7 +283,7 @@ func newSourceFromFile(path string, format string) (Source, error) {
 	if format == "" {
 		format = formatFromFileExtension(path)
 	}
-	return FileSource{Name: path, Data: string(data), Format: format}, nil
+	return FileSource{Name: path, Data: string(data), Format: format, FromUser: true}, nil
```
freddygv, line 286: This is for any provided config files/directories.

agent/config/builder.go
```diff
@@ -2522,6 +2523,12 @@ func validateAbsoluteURLPath(p string) error {
 func (b *builder) buildTLSConfig(rt RuntimeConfig, t TLS) (tlsutil.Config, error) {
 	var c tlsutil.Config
 
+	// This flag indicates whether any of the per-listener configuration was set.
+	// The flag was populated earlier when applying deprecated options.
```
kisunji, line 2527: Could we clarify this comment with whether this is for backwards compatibility and when we can expect to remove it?

agent/config/config.go
```diff
@@ -872,12 +877,28 @@ type TLSProtocolConfig struct {
 	UseAutoCert          *bool   `mapstructure:"use_auto_cert"`
 }
 
+func (c TLSProtocolConfig) IsZero() bool {
+	v := reflect.ValueOf(c)
+
+	hasValues := false
+	for i := 0; i < v.NumField(); i++ {
+		if !v.Field(i).IsNil() {
+			hasValues = true
```
kisunji, line 886: we can simply `return false` on the first field that is not nil

agent/config/deprecated.go
```diff
@@ -172,15 +172,26 @@ func applyDeprecatedConfig(d *decodeTarget) (Config, []string) {
...
 	var warns []string
 
 	tls := &cfg.TLS
+
+	// If the TLS stanza was specified by the user then we set a flag to indicate that.
+	// This check MUST happen before applying the deprecated options below, or else
+	// the tls struct will never be empty.
+	//
+	// This check was added for compatibility with Consul 1.11 style TLS configuration.
+	// Associated code can be removed once 1.14 is out and 1.11 is no longer supported.
```
kisunji, line 190: In previous jobs I found it helpful to add a Jira/GH issue number to each code location that can be removed so anyone on the team can grep for that number and remove the code easily.

agent/config/config.go
```diff
@@ -872,12 +877,28 @@ type TLSProtocolConfig struct {
...
+
 type TLS struct {
 	Defaults    TLSProtocolConfig `mapstructure:"defaults"`
 	InternalRPC TLSProtocolConfig `mapstructure:"internal_rpc"`
 	HTTPS       TLSProtocolConfig `mapstructure:"https"`
 	GRPC        TLSProtocolConfig `mapstructure:"grpc"`
 
+	// SpecifiedTLSStanza indicates whether the per-protocol tls stanza from configuration was used.
+	// If unspecified, and TLS is configured, that implies that the deprecated flags were used.
+	SpecifiedTLSStanza *bool
```
kisunji, line 900: do we need `mapstructure:"-"` here? I'm assuming this is not meant to be user-defined

kisunji (approved):
Just minor questions

agent/config/deprecated.go
freddygv: I actually just realized that this PR can be instead pointed at `release/1.13.x` exclusively, so that it doesn't end up in 1.14 at all 😎

freddygv: Going to open a new PR

freddygv:
Replaced by #14668 since these changes were going to be undone for 1.14 anyways.

PR #14670, opened 2022-09-16
fix: peering two partitions in the same dc

huikang:
### Description
Peering 2 partitions in the same datacenter is disallowed by the validation. If this is intended, maybe we can return a better error message.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


DanStough:
@huikang yes, it was not the intent to disable it. With your fix though, it seems like there could still be some problems if you are trying to `establish` from the default partition. Do you have the commands you used to `generate` and `establish` the token handy?

DanStough:
We are disallowing this for the time being, and there is a PR up at #14923. The recommended approach would be to export services between partitions directly, which avoids the overhead of the peer stream and replication.

PR #14677, opened 2022-09-19
doc(peering): missing network requirement for peering and the partiti…

huikang:
### Description
- The peering doc doesn't mention about the gRPC port, which is by defautl only bound to localhost.
- the list peering API allows `partition` query parameter

- Without setting client_addr to non-localhost address, peering establishment fails
- partition queyr parameter works, but missing in the doc

```
curl http://127.0.0.1:8500/v1/peerings\?partition\=foo 
[
  {
    "ID": "7c25adf4-73cd-6d1d-284c-5f96e36ace33",
    "Name": "dc3",
    "Partition": "foo",
    "State": "ACTIVE",
    "ImportedServiceCount": 0,
    "ExportedServiceCount": 0,
    "CreateIndex": 23,
    "ModifyIndex": 23
  }
]
```

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/website/content/api-docs/peering.mdx
+++ b/website/content/api-docs/peering.mdx
@@ -270,6 +270,11 @@ The table below shows this endpoint's support for
 | ---------------- | ----------------- | ------------- | -------------- |
 | `NO`             | `consistent`      | `none`        | `peering:read` |
 
+### Query Parameters
+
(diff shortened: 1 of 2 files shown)


kisunji:
Thanks! While you're here, could you update the Delete part of the API docs to move `partition` from PATH parameter to QUERY?

kisunji (approved):

PR #14475, opened 2022-09-02
[OSS] feat: add PeerThroughMeshGateways to mesh config

DanStough:
### Description
In anticipation of adding mesh gateway support for peering control-plane traffic, this PR adds a new field to the `mesh` config entry that will be used to enable the feature.

### Testing & Reproduction steps
* Unit tests included

### Links
Kube CRD PR [TBD]

### PR Checklist

* [X] updated test coverage
* [X] external facing docs updated
* [X] not a security concern

PR #14694, opened 2022-09-21
Return mesh gateway addrs if peering through mgw

freddygv:
### Description
Exposing servers through mesh gateways will involve sharing mesh gateway addresses with peers instead of true server addresses.

There are two flows where these addresses are shared:

* When generating a peering token.
* When pushing server address updates in the peering stream.

Each flow was addressed in a separate commit.

* Unit tests

### Links
* NET-644
* PR with config entry changes: #14475

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/consul/peering_backend_test.go
+++ b/agent/consul/peering_backend_test.go
@@ -2,10 +2,13 @@ package consul
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"testing"
 	"time"
 
+	"github.com/hashicorp/consul/agent/structs"
+	"github.com/hashicorp/consul/types"
 	"github.com/stretchr/testify/require"
 	gogrpc "google.golang.org/grpc"
 
@@ -17,7 +20,9 @@ import (
 )
 
 func TestPeeringBackend_ForwardToLeader(t *testing.T) {
-	t.Parallel()
+	if testing.Short() {
+		t.Skip("too slow for testing.Short")
+	}
 
 	_, conf1 := testServerConfig(t)
 	server1, err := newServer(t, conf1)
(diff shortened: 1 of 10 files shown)


agent/grpc-external/services/peerstream/subscription_manager.go
```diff
@@ -239,16 +243,10 @@ func (m *subscriptionManager) handleEvent(ctx context.Context, state *subscripti
 
 		pending := &pendingPayload{}
 
-		// Directly replicate information about our mesh gateways to the consuming side.
-		// TODO(peering): should we scrub anything before replicating this?
-		if err := pending.Add(meshGatewayPayloadID, u.CorrelationID, csn); err != nil {
```
freddygv, line 244: We were never actually sent these to the peer, the handler for these is a no-op in `stream_resources.go`. That case statement has been removed now as well.

agent/consul/peering_backend_test.go
```diff
@@ -60,6 +65,83 @@ func TestPeeringBackend_ForwardToLeader(t *testing.T) {
 	})
 }
 
+func TestPeeringBackend_GetServerAddresses(t *testing.T) {
+	if testing.Short() {
+		t.Skip("too slow for testing.Short")
+	}
+
+	_, cfg := testServerConfig(t)
+	cfg.GRPCTLSPort = 8505
```
kisunji, line 74: Will it be an issue if this port is in use by other tests? Could use `freeport.GetOne`

agent/grpc-external/services/peerstream/subscription_manager.go
```diff
@@ -744,32 +742,125 @@ func (m *subscriptionManager) notifyServerAddrUpdates(
...
-	// TODO(peering): retry logic; fail past a threshold
 	for {
 		var err error
-		// Typically, this function will block inside `m.subscribeServerAddrs` and only return on error.
-		// Errors are logged and the watch is retried.
+
 		idx, err = m.subscribeServerAddrs(ctx, idx, updateCh)
-		if errors.Is(err, stream.ErrSubForceClosed) {
-			m.logger.Trace("subscription force-closed due to an ACL change or snapshot restore, will attempt resume")
+		if errors.Is(err, context.Canceled) && ctx.Err() == nil {
```
kisunji, line 845: I don't think this is possible; is the `&& ctx.Err() == nil` necessary?

agent/grpc-external/services/peerstream/subscription_manager.go
```diff
@@ -744,32 +742,125 @@ func (m *subscriptionManager) notifyServerAddrUpdates(
...
+			logger.Trace("subscription force-closed due to an ACL change or snapshot restore, will attempt resume")
+
 		} else if !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
-			m.logger.Warn("failed to subscribe to server addresses, will attempt resume", "error", err.Error())
-		} else {
-			m.logger.Trace(err.Error())
+			logger.Warn("failed to subscribe to server addresses, will attempt resume", "error", err.Error())
+
+		} else if err != nil {
+			logger.Trace(err.Error())
```
kisunji, line 855: I think given the else-if 🚂 only `context.Canceled` and `context.DeadlineExceeded` will reach here, in which case we may want to return early

agent/grpc-external/services/peerstream/subscription_manager.go
```diff
@@ -744,32 +742,125 @@ func (m *subscriptionManager) notifyServerAddrUpdates(
...
 
-		select {
-		case <-ctx.Done():
+		err = waiter.Wait(ctx)
+		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
 			return
-		default:
+
+		} else if err != nil {
+			logger.Error("failed to wait before re-trying sync", "error", err)
```
kisunji, line 863: `waiter.Wait` only returns context errors; we could remove this branch since I don't think there's any reason for a waiter function to return anything else

kisunji (approved):
minor comments but LGTM!

agent/grpc-external/services/peerstream/subscription_manager.go
freddygv: Good call

PR #14710, opened 2022-09-22
docs: Consul Dataplane Version Compatibility

pglass:
### Description

This updates the version compatibility matrix to include Consul Dataplane / Envoy versions for the upcoming beta.

### DIFF
--- a/website/content/docs/connect/proxies/envoy.mdx
+++ b/website/content/docs/connect/proxies/envoy.mdx
@@ -48,8 +48,7 @@ Envoy 1.16.x and older releases are no longer supported (see [HCSEC-2022-07](htt
 
 ### Envoy and Consul Dataplane
 
-Consul Dataplane was introduced with Consul 1.14. It requires both Consul and Envoy to be at the
-following versions. Each version of Consul Dataplane supports one specific version of Envoy.
+Consul Dataplane is a feature introduced in Consul v1.14. Because each version of Consul Dataplane supports one specific version of Envoy, you must use the following versions of Consul, Consul Dataplane, and Envoy together.
 
 | Consul Version      | Consul Dataplane Version | Bundled Envoy Version |
 | ------------------- | ------------------------ | ---------------------- |
--- a/website/content/docs/connect/proxies/envoy.mdx
+++ b/website/content/docs/connect/proxies/envoy.mdx
@@ -51,9 +51,9 @@ Envoy 1.16.x and older releases are no longer supported (see [HCSEC-2022-07](htt
 Consul Dataplane was introduced with Consul 1.14. It requires both Consul and Envoy to be at the
 following versions. Each version of Consul Dataplane supports one specific version of Envoy.
 
-| Consul Version      | Consul Dataplane Version | Required Envoy Version |
+| Consul Version      | Consul Dataplane Version | Bundled Envoy Version |
 | ------------------- | ------------------------ | ---------------------- |
-| 1.14.x              | 1.0                      | 1.23.1                 |
+| 1.14.x              | 1.0.x                      | 1.23.x                |
 
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


website/content/docs/connect/proxies/envoy.mdx
```diff
@@ -44,6 +46,15 @@ Consul supports **four major Envoy releases** at the beginning of each major Con
...
 Envoy 1.16.x and older releases are no longer supported (see [HCSEC-2022-07](https://discuss.hashicorp.com/t/hcsec-2022-07-consul-s-connect-service-mesh-affected-by-recent-envoy-security-releases/36332)). Consul 1.9.x clusters should be upgraded to 1.10.x and Envoy upgraded to the latest supported Envoy version for that release, 1.18.6.
 
+### Envoy and Consul Dataplane
+
+Consul Dataplane was introduced with Consul 1.14. It requires both Consul and Envoy to be at the
+following versions. Each version of Consul Dataplane supports one specific version of Envoy.
+
+| Consul Version      | Consul Dataplane Version | Required Envoy Version |
+| ------------------- | ------------------------ | ---------------------- |
+| 1.14.x              | 1.0                      | 1.23.1                 |
```
david-yu, line 56:
```suggestion
| Consul Version      | Consul Dataplane Version | Bundled Envoy Version |
| ------------------- | ------------------------ | ---------------------- |
| 1.14.x              | 1.0.x                      | 1.23.x                |
```

david-yu: Changed from Bundled, since its not possible to swap out Envoy versions. Also made the .x more apparent so we don't have to continue changing this for all versions of Envoy and Dataplane.

david-yu (changes requested):

website/content/docs/connect/proxies/envoy.mdx
pglass: Thanks. I'm personally fine with this! 

nit: My only thought about using something like "required" or "supported" (vs packaged/bundled) was for people that decide to build their own image for Consul Dataplane thinking they can use a different version of Envoy. Using "required" indicates they shouldn't change the Envoy version. Whereas something like "packaged" or "bundled" maybe leaves it open to other versions.

website/content/docs/connect/proxies/envoy.mdx
```diff
@@ -44,6 +46,16 @@ Consul supports **four major Envoy releases** at the beginning of each major Con
 1. Envoy 1.20.1 and earlier are vulnerable to [CVE-2022-21654](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-21654) and [CVE-2022-21655](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-21655). Both CVEs were patched in Envoy versions 1.18.6, 1.19.3, and 1.20.2.
 Envoy 1.16.x and older releases are no longer supported (see [HCSEC-2022-07](https://discuss.hashicorp.com/t/hcsec-2022-07-consul-s-connect-service-mesh-affected-by-recent-envoy-security-releases/36332)). Consul 1.9.x clusters should be upgraded to 1.10.x and Envoy upgraded to the latest supported Envoy version for that release, 1.18.6.
 
+### Envoy and Consul Dataplane
+
+~> **Note:** Consul Dataplane is currently in beta.
```
pglass, line 51: I added a "beta" disclaimer here (cc @boruszak for wording review)

website/content/docs/connect/proxies/envoy.mdx
david-yu: I see what you are saying, my thinking was the opposite that bundled would imply you can't really change or mess with the Envoy version. Right now though we will supply the container image that contains both envoy and the data plane binary so I find that building this yourself will likely not occur very often.

david-yu (approved):

PR #14722, opened 2022-09-22
Add new internal endpoint to list exported services to a peer

kisunji:
### Description
In Consul UI we want to display a condensed list of service names that are being exported to a peer. The state store functions already exist so this PR simply exposes a new endpoint.

* Added new testcases at UI and internal level

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


dhiaayachi (approved):
LGTM

PR #14724, opened 2022-09-22
Support Stale Queries for Trust Bundle Lookups

malizz:
### Description
Support Stale Queries for Trust Bundle Lookups

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/grpc-external/options_test.go
+++ b/agent/grpc-external/options_test.go
@@ -3,6 +3,7 @@ package external
 import (
 	"context"
 	"testing"
+	"time"
 
 	"github.com/hashicorp/consul/agent/structs"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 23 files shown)


agent/grpc-external/options_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/hashicorp/consul/agent/structs"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/metadata"
+)
+
+func TestQueryOptionsFromContextRoundTrip(t *testing.T) {
+
+	expected := structs.QueryOptions{
+		Token: "123",
```
erichaberkorn, line 15: Could you test all the fields on `QueryOptions`. I think `QueryOptionsFromContext` might result in errors when non-string fields are set.

malizz: Great catch, thanks! I updated the decode logic.

erichaberkorn (approved):

PR #14751, opened 2022-09-26
Fix tproxy listeners with discovery chains.

hashi-derek:
### Description

Transparent proxy does not spawn outbound listeners for explicit upstreams to service-resolvers. When upstreams are explicitly defined, a corresponding listener should be spawned.

Create a service with the following fields configured:

```
consul.hashicorp.com/connect-inject: "true"
consul.hashicorp.com/connect-service-upstreams: "upstream:9001"
consul.hashicorp.com/transparent-proxy: "true"
```

Create a service resolver that will be used as the upstream.
```
apiVersion: consul.hashicorp.com/v1alpha1
kind: ServiceResolver
metadata:
  name: upstream
spec:
  redirect:
    service: other-svc
```

Check the Envoy admin config dump and observe that no listeners are spawned.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/proxycfg/connect_proxy.go
+++ b/agent/proxycfg/connect_proxy.go
@@ -300,7 +300,6 @@ func (s *handlerConnectProxy) handleUpdate(ctx context.Context, u UpdateEvent, s
 		}
 
 		seenUpstreams := make(map[UpstreamID]struct{})
-
 		for _, psn := range resp.Services {
 			uid := NewUpstreamIDFromPeeredServiceName(psn)
 
@@ -403,16 +402,11 @@ func (s *handlerConnectProxy) handleUpdate(ctx context.Context, u UpdateEvent, s
 			return fmt.Errorf("invalid type for response %T", u.Result)
 		}
 
-		// Contains the list of all implicit / explicit upstreams found.
 		seenUpstreams := make(map[UpstreamID]struct{})
-		// Contains only the implicit (intention-based) upstreams.
-		intentionUpstreams := make(map[UpstreamID]struct{})
-
 		for _, svc := range resp.Services {
 			uid := NewUpstreamIDFromServiceName(svc)
 
 			seenUpstreams[uid] = struct{}{}
-			intentionUpstreams[uid] = struct{}{}
 
 			cfgMap := make(map[string]interface{})
 			u, ok := snap.ConnectProxy.UpstreamConfig[uid]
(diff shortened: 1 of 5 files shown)


agent/proxycfg/connect_proxy.go
```diff
@@ -155,11 +155,6 @@ func (s *handlerConnectProxy) initialize(ctx context.Context) (ConfigSnapshot, e
 		if u.Datacenter != "" {
 			dc = u.Datacenter
 		}
-		if s.proxyCfg.Mode == structs.ProxyModeTransparent && (dc == "" || dc == s.source.Datacenter) {
```
hashi-derek, line 158: Listeners in `xds/listeners.go` are spawned by the discovery chain watch that is initialized a few lines below this. I'm not entirely sure if introducing this discovery chain will cause other side-effects elsewhere in the application. It doesn't appear as if it will cause any issues.

freddygv: An issue with this approach is that this discovery chain for this service will eventually get deleted.

When the list of upstreams from intentions comes in we clean up any data for service names that weren't present in that response:
https://github.com/hashicorp/consul/blob/main/agent/proxycfg/connect_proxy.go#L466

However, virtual services from resolvers won't be returned by that watch. Intention upstreams are computed by starting from the list of service names in the catalog, and then filtering out based on whether intentions allow/deny access to that name. 

For this explicit upstream to persist you'll likely need to either:
* Update the reconciling logic in proxycfg to not delete data from explicit upstreams when we receive the list of intention upstreams
* Return service resolvers as valid intention upstreams

This issue doesn't show up in the new unit test because the `db-redir` discovery chain event happens after the `intentionUpstreamsID` update, but if the order was flipped, the output would change because of the reconciliation.

freddygv (commented):
I think this is a good start! Though there is a gotcha here because there are assumptions in the code that discovery chain watches for tproxy only get managed in one place. 

I left a comment below explaining.

agent/proxycfg/connect_proxy.go
```diff
@@ -461,7 +462,14 @@ func (s *handlerConnectProxy) handleUpdate(ctx context.Context, u UpdateEvent, s
...
+
+		for uid, upstream := range snap.ConnectProxy.UpstreamConfig {
+			// Prevent cleanup of discovery chains for explicit upstream definitions.
+			if !upstream.CentrallyConfigured {
+				seenUpstreams[uid] = struct{}{}
+			}
+		}
 
 		// Clean up data from services that were not in the update
 		for uid, targets := range snap.ConnectProxy.WatchedUpstreams {
```
freddygv, line 475: What if instead we modify the check for cross-datacenter upstreams to instead check `!CentrallyConfigured`? That way we avoid allocating a new map. We're already checking UpstreamConfig in the reconciliation. 

```go
if upstream, ok := snap.ConnectProxy.UpstreamConfig[uid]; ok && upstream.Datacenter != "" && upstream.Datacenter != s.source.Datacenter {
	continue
}
```

Could become:
```go
if upstream, ok := snap.ConnectProxy.UpstreamConfig[uid]; ok && !upstream.CentrallyConfigured {
	continue
}
```

This works because a cross-DC watch can only be done explicitly, and not via `service-defaults`/`CentrallyConfigured`.

freddygv: (Applies through L549))

PR #14796, opened 2022-09-29
Use internal server certificate for peering TLS

freddygv:
### Description
#14485  and #14556 introduced an internally-managed server certificate
to use for peering-related purposes.

Now the peering token has been updated to match that behavior:
- The server name matches the structure of the server cert
- The CA PEMs correspond to the Connect CA

* Unit and integration tests

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/rpc/peering/service.go
```diff
@@ -476,8 +473,13 @@ func (s *Server) validatePeeringLocality(token *structs.PeeringToken, partition
...
 	// If the token has the same server name as this cluster, but we can't find the peering
 	// in our store, it indicates a naming conflict.
-	if s.Backend.GetServerName() == token.ServerName && peering == nil {
-		return fmt.Errorf("conflict - peering token's server name matches the current cluster's server name, %q, but there is no record in the database", s.Backend.GetServerName())
+	serverName, _, err := s.Backend.GetTLSMaterials()
+	if err != nil {
+		return fmt.Errorf("failed to fetch TLS materials: %w", err)
+	}
+
+	if serverName != "" && token.ServerName != "" && serverName == token.ServerName && peering == nil {
```
freddygv, line 481: Actually this needs some more thinking regarding peering a server with Connect to one without or a server with TLS to one without.

agent/consul/peering_backend.go
```diff
@@ -51,10 +53,39 @@ func (b *PeeringBackend) GetLeaderAddress() string {
...
-// GetAgentCACertificates gets the server's raw CA data from its TLS Configurator.
-func (b *PeeringBackend) GetAgentCACertificates() ([]string, error) {
-	// TODO(peering): handle empty CA pems
-	return b.srv.tlsConfigurator.GRPCManualCAPems(), nil
+// GetTLSMaterials returns the TLS materials for the dialer to dial the acceptor using TLS.
+// It returns the server name to validate, and the CA certificate to validate with.
+func (b *PeeringBackend) GetTLSMaterials() (string, []string, error) {
+	// Do not send TLS materials to the dialer if we to not have TLS configured for gRPC.
+	if b.srv.config.GRPCTLSPort <= 0 && !b.srv.tlsConfigurator.GRPCServerUseTLS() {
+		return "", nil, nil
```
freddygv, line 61: This may actually need to change to always return a server name, need to think about it a bit more

agent/rpc/peering/service.go
freddygv: TLS is required when generating a token now, so the token.ServerName will always be populated.

agent/consul/peering_backend.go
freddygv: Updated to require TLS + Connect. No longer considering the manual cert in favor of auto-TLS.

freddygv:
Hmm will look into the integration test failures tomorrow

agent/consul/peering_backend.go
```diff
@@ -51,10 +53,34 @@ func (b *PeeringBackend) GetLeaderAddress() string {
 	return b.leaderAddr
 }
 
-// GetAgentCACertificates gets the server's raw CA data from its TLS Configurator.
-func (b *PeeringBackend) GetAgentCACertificates() ([]string, error) {
-	// TODO(peering): handle empty CA pems
-	return b.srv.tlsConfigurator.GRPCManualCAPems(), nil
+// GetTLSMaterials returns the TLS materials for the dialer to dial the acceptor using TLS.
+// It returns the server name to validate, and the CA certificate to validate with.
+func (b *PeeringBackend) GetTLSMaterials(generatingToken bool) (string, []string, error) {
```
erichaberkorn, line 58: I'm probably missing something, but am not seeing the code for falling back to the manually configured
certs. Where does that happen?

freddygv: Good catch! We decided to not have that fallback anymore. Since Connect is now required when generating tokens we're just always going to use the internally managed cert.

If someone wants to adjust details about how the CA is managed they can do that by updating the Connect CA config.

I updated the PR description to remove the fallback message.

erichaberkorn: Awesome, thank you

erichaberkorn (approved):

PR #14800, opened 2022-09-29
Add TCP keepalive settings to proxy config for mesh gateways

kyhavlov:
This PR adds TCP keepalive settings to the proxy config to be applied to connections between remote mesh gateways. This makes it easier to handle the case of a proxy in between with its own idle timeout settings (like an AWS NLB, which has a non-configurable 350s idle connection timeout).

TODO:
- [x] Update docs


boxofrad (approved):
Looks good to me 👏🏻 For completeness, should we also expose the `# of probes` parameter? Or do you think it's best to leave this as OS default?

kyhavlov:
I can add the probes parameter - I was debating whether it was useful enough to add to avoid clutter, but it's probably worth it for completeness

boxofrad (approved):

PR #14811, opened 2022-09-30
Add DNS gRPC proxying support

chapmanc:
### Description
Adding  support to provide a gRPC endpoint that will proxy take in dns requests with a Protobuf message that takes a Msg and a Protocol type. 

This PR is best viewed by commit by commit. 
1. a5f9306 adds the proto file the dns  proxy 
2. c4a73a2 generate the new protobufs
3. 2c6ef41 adds in automatic generation of protobufs
4. 29fe2bd generate the mocks with `make proto`
5. 7bc9cad add in gRPC handler for proxying dns requests
6. Bind a DNSServer to the gRPC hadler and fix maximum size requirements for UDP

### Links

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


mkeeler (approved):

agent/dns.go
```diff
@@ -1286,8 +1295,9 @@ func trimUDPResponse(req, resp *dns.Msg, udpAnswerLimit int) (trimmed bool) {
 	// will allow our responses to be compliant even if some downstream server
 	// uncompresses them.
 	// Even when size is too big for one single record, try to send it anyway
-	// (useful for 512 bytes messages)
-	for len(resp.Answer) > 1 && resp.Len() > maxSize-7 {
+	// (useful for 512 bytes messages). 8 is removed from maxSize to ensure that we account
+	// for the udp header (8 bytes).
+	for len(resp.Answer) > 1 && resp.Len() > maxSize-8 {
```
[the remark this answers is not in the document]
chapmanc: Chatted offline about this but it should be fine. Not all maxSize will be set to the maxUDPDatagramSize so we account for that remaining 8 bytes of header here. Additionally the binary search trimming also introduces fuzziness so this shouldn't be something we ever run into. Correcting 7 bytes to 8 bytes since that's the actual size of the header.

agent/agent.go
```diff
@@ -920,6 +921,15 @@ func (a *Agent) listenAndServeDNS() error {
 			}
 		}(addr)
 	}
+	s, _ := NewDNSServer(a)
+
+	grpcDNS.NewServer(grpcDNS.Config{
+		Logger:       a.logger.Named("grpc-api.dns"),
+		DNSServeMux:  s.mux,
+		LocalAddress: grpcDNS.Local{IP: net.IPv4(127, 0, 0, 1), Port: a.config.GRPCPort},
```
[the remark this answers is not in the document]
chapmanc: Noted. Since local address really doesn't determine much here I could also pass that port through but it seems unnecessary considering gRPC will handle the connection regardless. We just need something in local address and I don't _think_ the port will affect the underlying dns message?

agent/grpc-external/services/dns/server.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/miekg/dns"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/peer"
+	"google.golang.org/grpc/status"
+
+	"github.com/hashicorp/consul/proto-public/pbdns"
+)
+
+type Local struct {
```
loshz, line 18: NIT: This should be `LocalAddr`

agent/grpc-external/services/dns/server.go
```diff
@@ -0,0 +1,136 @@
...
+		b.Logger.Error("error packing message", "err", err)
+		return err
+	}
+	b.ResponseBuffer = msgBytes
+	return nil
+}
+
+// Write writes a raw buffer back to the client.
+func (b *BufferResponseWriter) Write(m []byte) (int, error) {
+	b.Logger.Info("Write was called")
```
loshz, line 73: Should this be debug?

agent/grpc-external/services/dns/server.go
```diff
@@ -0,0 +1,136 @@
...
+	copy(b.ResponseBuffer, m)
+	return len(b.ResponseBuffer), nil
```
loshz, line 75: `copy` returns the number of elements copied, so you don't have to count the buffer twice:
```suggestion
	return copy(b.ResponseBuffer, m), nil
```

agent/grpc-external/services/dns/server.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// TsigTimersOnly sets the tsig timers only boolean.
+func (b *BufferResponseWriter) TsigTimersOnly(bool) {}
+
+// Hijack lets the caller take over the connection.
+// After a call to Hijack(), the DNS package will not do anything with the connection. {
+func (b *BufferResponseWriter) Hijack() {}
+
+func (s *Server) Query(ctx context.Context, req *pbdns.QueryRequest) (*pbdns.QueryResponse, error) {
```
loshz, line 97: I think it's worth adding a comment to this method.

proto-public/pbdns/dns.proto
```diff
@@ -0,0 +1,28 @@
...
+
+service DNSService {
+  // Query sends a DNS request over to Consul server and returns a DNS reply message.
+  rpc Query(QueryRequest) returns (QueryResponse) {}
+}
+
+enum Protocol {
+  PROTOCOL_UNSET_UNSPECIFIED = 0;
+  PROTOCOL_TCP = 1;
+  PROTOCOL_UDP = 2;
```
loshz, line 15: Why do we need the `PROTOCOL_` prefix?

agent/grpc-external/services/dns/server.go
```diff
@@ -0,0 +1,136 @@
...
+
+type Config struct {
+	Logger       hclog.Logger
+	DNSServeMux  *dns.ServeMux
+	LocalAddress Local
+}
+
+type Server struct {
+	Config
+}
```
loshz, line 31: Is it worth us combining these structs? Not sure what benefit we get from a separate `Config`?

agent/grpc-external/services/dns/server.go
```diff
@@ -0,0 +1,136 @@
...
+	return &Server{cfg}
+}
+
+func (s *Server) Register(grpcServer *grpc.Server) {
+	pbdns.RegisterDNSServiceServer(grpcServer, s)
+}
+
+// BufferResponseWriter writes a DNS response to a byte buffer.
+type BufferResponseWriter struct {
+	ResponseBuffer []byte
```
loshz, line 43: Does the buf need to be exported? Do we intend for it to be read/wrote externally?

loshz (commented):
I've left a couple of non-blocking comments, but overall great job!

proto-public/pbdns/dns.proto
chapmanc: Linter :(. proto-lint insisted and I was not happy about changing it.

agent/grpc-external/services/dns/server.go
chapmanc: I think it's simply to allow the `NewServer` method to not have to have individual arguments.

proto-public/pbdns/dns.proto
mkeeler: We use `buf lint` for all of this. 

IIRC the rationale is is many languages (Go included) cannot properly namespace enum variants and therefore you need to encoded the prefix into the name to help prevent naming collisions. This is especially true for the UNSET/UNSPECIFIED enum variant as you would likely name it identically in all enums.

PR #14812, opened 2022-09-30
ui: fix link back to hcp not showing up

LevelbossMike:
### Description
Make sure to use the "sub-app"-approach of injecting the hcp/back-link properly so that it actually gets bundled in production environments.

#### Notable
This issue probably occurred because testing against a real backend is hard at the moment. We most likely want to create an easy-to-use setup in the near future that allows us to run the UI against a proper consul-agent.

* run `make ui-docker` and `make docker-dev` after that
* setup a `docker-compose`-file that starts an agent and pass the following flag: `CONSUL_HCP_ENABLED=true`, alternatively you can run proper HCP and boot consul inside of it.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-ui/lib/startup/templates/body.html.js
+++ b/ui/packages/consul-ui/lib/startup/templates/body.html.js
@@ -71,6 +71,7 @@ ${
   <script src="${rootURL}assets/consul-nspaces/routes.js"></script>
 {{end}}
 {{if .HCPEnabled}}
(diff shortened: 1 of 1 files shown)


WenInCode (approved):

WenInCode:
I used your peering setup and passed in the HCP env var. Thanks for putting that repo together 👍

PR #14817, opened 2022-09-30
Make the mesh gateway changes to allow `local` mode for cluster peering data plane traffic

erichaberkorn:
### Description
This PR creates Envoy clusters and envoys on mesh gateways that forward requests to a mesh gateway in another admin partition. These Envoy resources are used to enable mesh gateway local mode with cluster peering.

* Manually set up cluster peering and verified that this works
* Unit tests
* Integration tests

### PR Checklist

* [X] updated test coverage
* [X] not a security concern

### DIFF
--- a/agent/structs/testing_catalog.go
+++ b/agent/structs/testing_catalog.go
@@ -55,8 +55,8 @@ func TestNodeServiceWithName(t testing.T, name string) *NodeService {
 
 const peerTrustDomain = "1c053652-8512-4373-90cf-5a7f6263a994.consul"
 
-func TestNodeServiceWithNameInPeer(t testing.T, name string, peer string) *NodeService {
-	return &NodeService{
+func TestCheckNodeServiceWithNameInPeer(t testing.T, name, peer, ip string, useHostname bool) CheckServiceNode {
+	service := &NodeService{
 		Kind:     ServiceKindTypical,
 		Service:  name,
 		Port:     8080,
@@ -72,6 +72,29 @@ func TestNodeServiceWithNameInPeer(t testing.T, name string, peer string) *NodeS
 				Protocol: "tcp",
 			},
 		},
+	}
+
+	if useHostname {
+		service.TaggedAddresses = map[string]ServiceAddress{
+			TaggedAddressLAN: {
+				Address: ip,
+				Port:    443,
+			},
+			TaggedAddressWAN: {
+				Address: name + ".us-east-1.elb.notaws.com",
+				Port:    8443,
+			},
+		}
+	}
+
+	return CheckServiceNode{
+		Node: &Node{
+			ID:         "test1",
+			Node:       "test1",
+			Address:    ip,
+			Datacenter: "cloud-dc",
+		},
+		Service: service,
 	}
 }
 
--- a/agent/structs/testing_catalog.go
+++ b/agent/structs/testing_catalog.go
@@ -56,18 +56,18 @@ func TestNodeServiceWithName(t testing.T, name string) *NodeService {
 const peerTrustDomain = "1c053652-8512-4373-90cf-5a7f6263a994.consul"
 
 func TestNodeServiceWithNameInPeer(t testing.T, name string, peer string) *NodeService {
-	service := "payments"
 	return &NodeService{
-		Kind:    ServiceKindTypical,
-		Service: name,
-		Port:    8080,
+		Kind:     ServiceKindTypical,
+		Service:  name,
+		Port:     8080,
+		PeerName: peer,
 		Connect: ServiceConnect{
 			PeerMeta: &PeeringServiceMeta{
 				SNI: []string{
-					service + ".default.default." + peer + ".external." + peerTrustDomain,
+					name + ".default.default." + peer + ".external." + peerTrustDomain,
 				},
 				SpiffeID: []string{
-					"spiffe://" + peerTrustDomain + "/ns/default/dc/" + peer + "-dc/svc/" + service,
+					"spiffe://" + peerTrustDomain + "/ns/default/dc/" + peer + "-dc/svc/" + name,
 				},
 				Protocol: "tcp",
 			},
--- a/agent/xds/clusters.go
+++ b/agent/xds/clusters.go
@@ -575,22 +575,45 @@ func (s *ResourceGenerator) makeGatewayOutgoingClusterPeeringServiceClusters(cfg
 
 	for _, serviceGroups := range cfgSnap.MeshGateway.PeeringServices {
 		for sn, serviceGroup := range serviceGroups {
-			var clusterName string
-			var isRemote bool
-			if len(serviceGroup) > 0 {
-				node := serviceGroup[0]
-				if node.Service == nil {
-					return nil, fmt.Errorf("couldn't get SNI for peered service %s", sn.String())
-				}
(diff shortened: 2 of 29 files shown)


agent/xds/clusters.go
```diff
@@ -533,6 +540,39 @@ func (s *ResourceGenerator) makeGatewayServiceClusters(
...
+		for sn, serviceGroup := range serviceGroups {
+			var clusterName string
+			var isRemote bool
+			if len(serviceGroup) > 0 {
+				node := serviceGroup[0]
+				if node.Service == nil {
+					return nil, fmt.Errorf("couldn't get SNI for peered service %s", sn.String())
+				}
+				clusterName = node.Service.Connect.PeerMeta.PrimarySNI()
+				isRemote = !cfgSnap.Locality.Matches(node.Node.Datacenter, node.Node.PartitionOrDefault())
```
erichaberkorn, line 560: Is this right? I copied the outline of this function from `makeGatewayServiceClusters` and am not 100% sure what to do here.

agent/xds/testdata/clusters/ingress-with-defaults-service-max-connections.latest.golden
```diff
@@ -16,7 +16,7 @@
       },
       "connectTimeout": "33s",
       "circuitBreakers": {
-        "thresholds":[
+        "thresholds": [
```
erichaberkorn, line 19: I'm not sure why this happened 🤔

test/integration/connect/envoy/case-cross-peers/primary/service_s1.hcl
```diff
@@ -9,6 +9,9 @@ services {
             destination_name = "s2"
             destination_peer = "primary-to-alpha"
             local_bind_port  = 5000
+            mesh_gateway {
```
erichaberkorn, line 12: Using local mode on the existing case tests both code paths because the local mesh gateway just forwards requests to the remote one. Does it make sense to test both cases separately? I left just one case because it feels more consistent with our existing integration tests.

agent/structs/testing_catalog.go
```diff
@@ -56,18 +56,18 @@ func TestNodeServiceWithName(t testing.T, name string) *NodeService {
 const peerTrustDomain = "1c053652-8512-4373-90cf-5a7f6263a994.consul"
 
 func TestNodeServiceWithNameInPeer(t testing.T, name string, peer string) *NodeService {
-	service := "payments"
```
erichaberkorn, line 59: This change fixed some incorrect failover golden snapshots.

test/integration/connect/envoy/case-cross-peers/primary/service_s1.hcl
freddygv: Seems OK to me.

agent/xds/testdata/clusters/ingress-with-defaults-service-max-connections.latest.golden
freddygv: 👻

agent/xds/endpoints.go
```diff
@@ -353,6 +360,37 @@ func (s *ResourceGenerator) endpointsFromServicesAndResolvers(
...
+func (s *ResourceGenerator) makeEndpointsForOutgoingPeeredServices(
+	cfgSnap *proxycfg.ConfigSnapshot,
+) ([]proto.Message, error) {
+	var resources []proto.Message
+
+	// generate the endpoints for the linked service groups
+	for _, serviceGroups := range cfgSnap.MeshGateway.PeeringServices {
+		for sn, serviceGroup := range serviceGroups {
+			var clusterName string
+			if len(serviceGroup) > 0 {
```
freddygv, line 372: Doesn't seem like we should proceed if `len(serviceGroup) == 0`, since there won't be any endpoints and the clusterName won't be populated

agent/xds/clusters.go
```diff
@@ -533,6 +540,39 @@ func (s *ResourceGenerator) makeGatewayServiceClusters(
...
+		return nil, fmt.Errorf("unsupported gateway kind %q", cfgSnap.Kind)
+	}
+
+	var clusters []proto.Message
+
+	for _, serviceGroups := range cfgSnap.MeshGateway.PeeringServices {
+		for sn, serviceGroup := range serviceGroups {
+			var clusterName string
+			var isRemote bool
+			if len(serviceGroup) > 0 {
```
freddygv, line 554: Same as with endpoints, if we can't get a clusterName we probably shouldn't attempt to make a cluster

Could do:
```go
if len(serviceGroup) == 0 { 
	continue
}
```

agent/xds/clusters.go
freddygv: `isRemote` could just be set to true on L565 since we already know we're dialing a different cluster

Though an issue here and in `endpoints.go` is that the hostname handling is missing. We need to check whether the addresses attached to these services are hostnames.

Take a look at `makeUpstreamLoadAssignmentForPeerService` and `makeUpstreamClusterForPeerService`. 

Since the remote addresses are mesh gateway addresses they're likely to have a hostname address (cloud provider LB) and we can't send hostnames via EDS. We instead attach them to a cluster with DNS as the discovery type.

test/integration/connect/envoy/case-cross-peers/primary/verify.bats
```diff
@@ -55,3 +55,7 @@ load helpers
 @test "s1 upstream made 1 connection to s2" {
   assert_envoy_metric_at_least 127.0.0.1:19000 "cluster.s2.default.primary-to-alpha.external.*cx_total" 1
 }
+
+@test "s1 upstream made 1 connection to s2 through the primary mesh gateway" {
+  assert_envoy_metric_at_least 127.0.0.1:19001 "cluster.s2.default.default.alpha-to-primary.external.*cx_total" 1
```
freddygv, line 60: This seems unusual, why is `alpha-to-primary` here? That's the name that `alpha` uses to refer to `primary`, so why would the local primary gateway be aware of that?

freddygv (changes requested):
Nice progress! I left some comments below. 

My main request is that the cluster/endpoint generation needs some adjustment to account for hostname endpoints, since they can't be sent over EDS. Let me know if you have any questions!

test/integration/connect/envoy/case-cross-peers/primary/verify.bats
erichaberkorn: Yeah, I was confused by this but was hoping there was some logic behind it. This matches [the SNI created by the sidecar proxy](https://github.com/hashicorp/consul/blob/main/agent/xds/clusters.go#L919) 🤔

freddygv: Ah I see what's going on. That's how the remote side can know that this traffic is for an exported service as opposed to the usual mesh gateway route for `s2`. See this comment: https://github.com/hashicorp/consul/blob/main/agent/xds/clusters.go#L1284-L1298

I'm a bit worried about whether there could be collisions on the dialing side though.

It encodes: `(service, namespace, partition, peerName, trustDomain)` so I think there could be a collision if there is a separate peering with all the same data but different DC.

freddygv (approved):
:shipit:

Issue #10809, opened 2021-08-09
If you write a proxy-defaults without name set the success message is confusing


lkysow:
#### Feature Description

Consul 1.10? introduced the ability to not set `Name = "global"` for proxy defaults configs, however when you write them with `consul config write` the success message looks weird:

```hcl
Kind = "proxy-defaults"
Config {
  protocol = "http"
}
```

```console
$ consul config write proxy-defaults.hcl
Config entry written: proxy-defaults/
```

Perhaps this could read

```
Config entry written: proxy-defaults
```

Or 

```
Config entry written: proxy-defaults/global
```

Which might help with knowing what to use when running `consul config read -kind proxy-defaults` since that command still requires the `-name global` flag.

dnephin:
I wanted to make the name "global" implied for `proxy-defaults`, but that PR did not merge: #10067 . We did something similar with the new `mesh` config-entry, but that shouldn't have impacted the `proxy-defaults`. I suspect this has always worked this way.

The value being printed here comes directly from the parsed config (no round-trip through the API handler), so we can handle this entirely in the `GetName()` method for `api.ProxyConfigEntry` and `api.MeshConfigEntry`.

PR #14833, opened 2022-10-02
config entry: hardcode proxy-defaults name as global

lkysow:
proxy-defaults can only have the name global. Because of this, we support not even setting the name in the config file:

```
kind = "proxy-defaults"
```

Previously, writing this would result in the output:

```
Config entry written: proxy-defaults/
```

Now it will output:

```
Config entry written: proxy-defaults/global
```

This change follows what was done for the new Mesh config entry.

Fixes https://github.com/hashicorp/consul/issues/10809

### DIFF
--- a/.changelog/14833.txt
+++ b/.changelog/14833.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+cli: always use name "global" for proxy-defaults config entries
+```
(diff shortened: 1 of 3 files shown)


ishustava (approved):
❤️

PR #14835, opened 2022-10-03
fix(api): missing peer name in query option

huikang:
### Description
When using API to query a service imported by a peering, we need to set the `peer` in query option, which is missing in the current implementation. This PR fixes this issue.

```go
cli.Health().Service("foo", "", false, &api.QueryOptions{
    Peer: "dc-abc",
})
```

* [x] updated test coverage
* [ ] not a security concern

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -111,6 +111,9 @@ type QueryOptions struct {
 	// by the Config
 	Datacenter string
 
+	// Providing a peer name in the query option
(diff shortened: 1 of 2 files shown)


DanStough (approved):
LGTM. Seems like magic we don't need to read from this anywhere in the actual endpoint, but that seems like how it works from looking at the other fields 🤷

PR #14836, opened 2022-10-03
Add peering integration tests

hashi-derek:
Adds tests for:

- Connect proxy -> split -> redirect -> peer
- Ingress gateway -> split -> redirect -> peer
- Ingress gateway -> redirect -> failover peer

Probably best viewed commit-by-commit.

### DIFF
--- a/test/integration/connect/envoy/case-cfg-splitter-cluster-peering/primary/verify.bats
+++ b/test/integration/connect/envoy/case-cfg-splitter-cluster-peering/primary/verify.bats
@@ -1,0 +1,74 @@
+#!/usr/bin/env bats
+
+load helpers
+
+@test "s1 proxy is running correct version" {
+  assert_envoy_version 19000
+}
+
+@test "s1 proxy admin is up" {
+  retry_default curl -f -s localhost:19000/stats -o /dev/null
+}
+
+@test "s2 proxy admin is up" {
+  retry_default curl -f -s localhost:19001/stats -o /dev/null
+}
+
+@test "s1 proxy listener should be up and have right cert" {
+  assert_proxy_presents_cert_uri localhost:21000 s1
+}
+
+@test "s2 proxies should be healthy in primary" {
+  assert_service_has_healthy_instances s2 1 primary
+}
+
+@test "s2 proxies should be healthy in alpha" {
+  assert_service_has_healthy_instances s2 1 alpha
+}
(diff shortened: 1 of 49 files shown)


test/integration/connect/envoy/case-cfg-splitter-cluster-peering/primary/verify.bats
```diff
@@ -0,0 +1,74 @@
...
+@test "s1 upstream should have healthy endpoints for s2 primary and alpha" {
+  assert_upstream_has_endpoints_in_status 127.0.0.1:19000 s2.default.primary.internal HEALTHY 1
+  assert_upstream_has_endpoints_in_status 127.0.0.1:19000 s2.default.primary-to-alpha.external HEALTHY 1
+}
+
+@test "s1 upstream should be split between peer and local dc" {
+  retry_long assert_url_header "http://127.0.0.1:5000/" "x-test-split" "primary"
+  [ "$status" -eq 0 ]
+  retry_long assert_url_header "http://127.0.0.1:5000/" "x-test-split" "alpha"
+  [ "$status" -eq 0 ]
```
erichaberkorn, line 50: My understanding is that `retry_long` retries once per second for 30 seconds. This means that given a 50-50 splitter the chances of each case flaking is (1 / 2) ^ 30 or approximately 1 out of a billion.

test/integration/connect/envoy/case-cfg-splitter-peering-ingress-gateways/primary/verify.bats
```diff
@@ -0,0 +1,70 @@
...
+}
+
+@test "requests through ingress should proxy to primary s1" {
+  assert_expected_fortio_name s1 s1.ingress.consul 10001
+  assert_envoy_metric 127.0.0.1:19000 "http.public_listener.rq_total" 1
+}
+
+@test "requests through ingress to splitter should go to alpha" {
+  retry_default assert_expected_fortio_name s1-alpha split.ingress.consul 10002
+  retry_default assert_expected_fortio_name s2-alpha split.ingress.consul 10002
```
erichaberkorn, line 69: If I'm reading this right, these will flake  1 / (2 ^5) = 1 / 32 times. What do you think about using `retry_long`?

erichaberkorn (approved):

PR #14853, opened 2022-10-04
ui: Allow managed-runtime badge to be dynamic

LevelbossMike:
### Description
Allow the `managed-runtime`-badge to display dynamic values. Before this change, we hard-coded the text to always display `Self-managed`. This change will make the badge dynamic by using the value that is set as the `HCPEnabledRuntime`-option as the text for the badge.

I did not add a test for this behavior, as it is nearly impossible to add a test for configuration options in only one test with the current setup of environment-variables.

Screenshot 2022-10-04 at 10 46 40

* add `HCPManagedRuntime` to the `operatorConfig` in `/config/environment.js` and set its value to the text you want the badge to display. In production, this will be taken care of by the server, for development this is the easiest way to change this option.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-ui/app/components/consul/datacenter/selector/index.hbs
+++ b/ui/packages/consul-ui/app/components/consul/datacenter/selector/index.hbs
@@ -1,67 +1,64 @@
-<li
-  class="dcs"
-  data-test-datacenter-menu
->
+<li class='dcs' data-test-datacenter-menu>
   {{#if (gt @dcs.length 1)}}
     <DisclosureMenu
-      aria-label="Datacenter"
+      aria-label='Datacenter'
       @items={{sort-by 'Primary:desc' 'Local:desc' 'Name:asc' @dcs}}
       data-test-datacenter-disclosure-menu
-    as |disclosure|>
-      <disclosure.Action
-        {{on 'click' disclosure.toggle}}
-      >
+      as |disclosure|
+    >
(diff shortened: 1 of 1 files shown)


ui/packages/consul-ui/app/components/consul/datacenter/selector/index.hbs
```diff
@@ -1,67 +1,64 @@
...
       </disclosure.Menu>
     </DisclosureMenu>
   {{else}}
-    <div class="dc-name" data-test-datacenter-single>
+    <div class='dc-name' data-test-datacenter-single>
       {{@dcs.firstObject.Name}}
-      {{#if (env 'CONSUL_HCP_MANAGED_RUNTIME')}}
-        <span>Self-managed</span>
-      {{/if}}
+      {{#let (env 'CONSUL_HCP_MANAGED_RUNTIME') as |managedRuntime|}}
```
LevelbossMike, line 57: This is the relevant change of the PR. The rest is an automatic update based on the template-lint setup.

WenInCode (approved):
Is there any documentation anywhere on what values might get set on that env var? Just curious!

LevelbossMike:
> Is there any documentation anywhere on what values might get set on that env var? Just curious!

I'm not sure 😕  For this task, I looked into the existing implementation and figured out the environment variables used that way.

PR #14869, opened 2022-10-04
Merge central config for GetEnvoyBootstrapParams

pglass:
### Description

This fixes GetEnvoyBootstrapParams to merge in proxy-defaults and service-defaults.

* Steps to reproduce: [Gist link](https://gist.github.com/pglass/d768358a85390abf90b98c6b33c10d54)
* A script to test: Run `consul agent -dev` and then run [`check.sh` here](https://gist.github.com/pglass/e68ecbcadfaa8b08d4d2da7c65b3c436#file-check-sh) and you should see [this output](https://gist.github.com/pglass/e68ecbcadfaa8b08d4d2da7c65b3c436#file-output-txt)
* Added unit tests

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/consul/server.go
+++ b/agent/consul/server.go
@@ -732,7 +732,6 @@ func NewServer(config *Config, flat Deps, externalGRPCServer *grpc.Server) (*Ser
 		Logger:                            logger.Named("grpc-api.dataplane"),
 		ACLResolver:                       s.ACLResolver,
 		Datacenter:                        s.config.Datacenter,
-		MergeNodeServiceWithCentralConfig: mergeNodeServiceWithCentralConfig,
 	}).Register(s.externalGRPCServer)
 
 	serverdiscovery.NewServer(serverdiscovery.Config{
--- a/agent/consul/server.go
+++ b/agent/consul/server.go
@@ -728,10 +728,10 @@ func NewServer(config *Config, flat Deps, externalGRPCServer *grpc.Server) (*Ser
 	s.externalConnectCAServer.Register(s.externalGRPCServer)
 
 	dataplane.NewServer(dataplane.Config{
-		GetStore:                          func() dataplane.StateStore { return s.FSM().State() },
-		Logger:                            logger.Named("grpc-api.dataplane"),
-		ACLResolver:                       s.ACLResolver,
-		Datacenter:                        s.config.Datacenter,
+		GetStore:    func() dataplane.StateStore { return s.FSM().State() },
+		Logger:      logger.Named("grpc-api.dataplane"),
+		ACLResolver: s.ACLResolver,
+		Datacenter:  s.config.Datacenter,
 	}).Register(s.externalGRPCServer)
 
 	serverdiscovery.NewServer(serverdiscovery.Config{
--- a/agent/consul/server.go
+++ b/agent/consul/server.go
@@ -728,10 +728,11 @@ func NewServer(config *Config, flat Deps, externalGRPCServer *grpc.Server) (*Ser
 	s.externalConnectCAServer.Register(s.externalGRPCServer)
 
 	dataplane.NewServer(dataplane.Config{
-		GetStore:    func() dataplane.StateStore { return s.FSM().State() },
-		Logger:      logger.Named("grpc-api.dataplane"),
-		ACLResolver: s.ACLResolver,
-		Datacenter:  s.config.Datacenter,
+		GetStore:                          func() dataplane.StateStore { return s.FSM().State() },
+		Logger:                            logger.Named("grpc-api.dataplane"),
+		ACLResolver:                       s.ACLResolver,
+		Datacenter:                        s.config.Datacenter,
+		MergeNodeServiceWithCentralConfig: mergeNodeServiceWithCentralConfig,
 	}).Register(s.externalGRPCServer)
 
 	serverdiscovery.NewServer(serverdiscovery.Config{
--- a/agent/grpc-external/services/dataplane/get_envoy_bootstrap_params.go
+++ b/agent/grpc-external/services/dataplane/get_envoy_bootstrap_params.go
@@ -89,7 +89,7 @@ func (s *Server) GetEnvoyBootstrapParams(ctx context.Context, req *pbdataplane.G
 	)
 	if err != nil {
 		logger.Error("Error merging with central config", "error", err)
-		return nil, status.Error(codes.Unknown, fmt.Sprintf("Error merging central config: %v", err))
+		return nil, status.Errorf(codes.Unknown, "Error merging central config: %v", err)
 	}
 
 	bootstrapConfig, err := structpb.NewStruct(ns.Proxy.Config)
--- a/agent/grpc-external/services/dataplane/get_envoy_bootstrap_params.go
+++ b/agent/grpc-external/services/dataplane/get_envoy_bootstrap_params.go
@@ -3,16 +3,17 @@ package dataplane
 import (
 	"context"
 	"errors"
+	"fmt"
 	"strings"
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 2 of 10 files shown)


agent/grpc-external/services/dataplane/server.go
```diff
@@ -20,7 +22,8 @@ type Config struct {
 	Logger      hclog.Logger
 	ACLResolver ACLResolver
 	// Datacenter of the Consul server this gRPC server is hosted on
-	Datacenter string
+	Datacenter                        string
+	MergeNodeServiceWithCentralConfig func(memdb.WatchSet, *state.Store, *structs.ServiceSpecificRequest, *structs.NodeService, hclog.Logger) (uint64, *structs.NodeService, error)
```
pglass, line 26: This function is passed in to avoid an import cycle (Open to other ideas here)

agent/grpc-external/services/dataplane/get_envoy_bootstrap_params_test.go
```diff
@@ -280,3 +364,185 @@ func TestGetEnvoyBootstrapParams_PermissionDenied(t *testing.T) {
 	require.Equal(t, codes.PermissionDenied.String(), status.Code(err).String())
 	require.Nil(t, resp)
 }
+
+// Copied implementation from agent/consul/merge_service_config.go to avoid import cycle.
+func testMergeNodeServiceWithCentralConfig(
```
pglass, line 369: I copied the function over for the unit tests because I didn't know a better way to avoid the import cycle.

agent/grpc-external/services/dataplane/get_envoy_bootstrap_params.go
```diff
@@ -73,7 +74,24 @@ func (s *Server) GetEnvoyBootstrapParams(ctx context.Context, req *pbdataplane.G
...
+		&structs.ServiceSpecificRequest{
+			Datacenter:   s.Datacenter,
+			QueryOptions: options,
+		},
+		svc.ToNodeService(),
+		logger,
+	)
+	if err != nil {
+		logger.Error("Error merging with central config", "error", err)
+		return nil, status.Error(codes.Unknown, fmt.Sprintf("Error merging central config: %v", err))
```
pglass, line 91: I could try to discern a better error code here?

agent/grpc-external/services/dataplane/get_envoy_bootstrap_params.go
```diff
@@ -73,7 +74,24 @@ func (s *Server) GetEnvoyBootstrapParams(ctx context.Context, req *pbdataplane.G
 		NodeId:      string(svc.ID),
 	}
 
-	bootstrapConfig, err := structpb.NewStruct(svc.ServiceProxy.Config)
+	// This is awkward because it's designed for different requests, but
+	// this fakes the ServiceSpecificRequest so that we can reuse code.
+	_, ns, err := s.MergeNodeServiceWithCentralConfig(
+		nil,
```
pglass, line 80: I pass a nil here for the `memdb.WatchSet`. Is that fine? 

We also do make multiple reads from the store: one for the `store.ServiceNode` and then at least one here to read the config entries, so I was wondering if those should maybe be batched into a transaction or such?

agent/grpc-external/services/dataplane/get_envoy_bootstrap_params.go
riddhi89: Seems reasonable to me (since we emit the same code if there is an issue parsing this config below).

riddhi89 (approved):
LGTM !
Although it is a bit quirky to see the helper passed around, I think it is a good first pass for now.

Would be good if Dan/another core engineer also gave this look over.

agent/grpc-external/services/dataplane/get_envoy_bootstrap_params.go
boxofrad: Yep, calling `Add` on a nil watch set is a no-op, so it's fine.

As for doing all the queries in a single transaction, I guess the main benefit would be [isolation](https://en.wikipedia.org/wiki/Isolation_(database_systems)) - that is, you'd never get a mix of config entries from different versions of the database (before and after a write).

Which is an important enhancement, but I agree with @riddhi89 isn't critical for right now.

agent/grpc-external/services/dataplane/get_envoy_bootstrap_params.go
```diff
@@ -73,7 +74,24 @@ func (s *Server) GetEnvoyBootstrapParams(ctx context.Context, req *pbdataplane.G
...
+		store.(*state.Store),
```
boxofrad, line 81: Would it be possible to have the method accept an interface scoped to the queries it needs to make? I'm guessing not because of the import-cycle problem?

agent/grpc-external/services/dataplane/get_envoy_bootstrap_params.go
boxofrad, line 91:
```suggestion
		return nil, status.Errorf(codes.Unknown, "Error merging central config: %v", err)
```

agent/consul/server.go
```diff
@@ -728,10 +728,11 @@ func NewServer(config *Config, flat Deps, externalGRPCServer *grpc.Server) (*Ser
...
 	dataplane.NewServer(dataplane.Config{
-		GetStore:    func() dataplane.StateStore { return s.FSM().State() },
-		Logger:      logger.Named("grpc-api.dataplane"),
-		ACLResolver: s.ACLResolver,
-		Datacenter:  s.config.Datacenter,
+		GetStore:                          func() dataplane.StateStore { return s.FSM().State() },
+		Logger:                            logger.Named("grpc-api.dataplane"),
+		ACLResolver:                       s.ACLResolver,
+		Datacenter:                        s.config.Datacenter,
+		MergeNodeServiceWithCentralConfig: mergeNodeServiceWithCentralConfig,
```
boxofrad, line 735: I'm wondering how much work it'd be to move `mergeNodeServiceWithCentralConfig` into `agent/configentry`.

Aside from the `state.Store` parameter, I can't immediately see anything that'd create an import cycle (and we can swap that parameter for an interface).

We'll also need this method to merge config for xDS, so I think moving it is worthwhile.

boxofrad (changes requested):
Looking good! 🙌🏻  I think we should try to move that method though, shout if I've missed something that makes it tricky.

agent/consul/server.go
pglass: You are right! Moving this was easier than I thought: https://github.com/hashicorp/consul/pull/14869/commits/9935ac249523d40e5499a23a5b89cb457ae6b0b8

It's now at `configentry.MergeNodeServiceWithCentralConfig` and it accepts an interface for the state store like you suggested. A lot cleaner 🙏

boxofrad (approved):
Great job 🎉 👏🏻

agent/grpc-external/services/dataplane/get_envoy_bootstrap_params.go
```diff
@@ -73,7 +74,24 @@ func (s *Server) GetEnvoyBootstrapParams(ctx context.Context, req *pbdataplane.G
 		NodeId:      string(svc.ID),
 	}
 
-	bootstrapConfig, err := structpb.NewStruct(svc.ServiceProxy.Config)
+	// This is awkward because it's designed for different requests, but
+	// this fakes the ServiceSpecificRequest so that we can reuse code.
+	_, ns, err := configentry.MergeNodeServiceWithCentralConfig(
+		nil,
+		store.(configentry.StateStore),
```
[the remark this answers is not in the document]
pglass: Ah, right. I had the cast because I was missing the `ReadResolvedServiceConfigEntries` from the `dataplane.StateStore` interface.

Fixed that, so no more cast

PR #14888, opened 2022-10-05
Fix connect secondary CA initialization

dclaisse:
### Description
During a new datacenter opening using consul 1.12.5, we faced multiple issues with connect as CA wouldn't initialize.
This datacenter is a secondary one, initializing its CA from another one, using internal consul provider.

Investigations showed that our primary datacenter was exposing an empty `ExternalTrustDomain` when calling `/v1/connect/ca/roots endpoint`, which is used in turn to set cluster ID of secondary datacenter, and would make CA appear uninitialized.

Observations also showed that `/v1/connect/ca/configuration` was not updating secondary CA configuration due to a logic inversion when moving condition into a function.

Finally, leaf certificates generated by connect were not valid, because intermediate certificate of secondary CA had no Subject Common Name, and in turn leaf certificates had no Issuer Common Name, which violates RFC 5280.

* Root CA state realignment issue can be observed by adding a new field in structs.CARoot structure, and filling it in newCARoot function, without this patch the `/v1/connect/ca/roots` endpoint will show this new field as empty. I also added a test to ensure no unnecessary update is done.
* For intermediate certificate Subject Common Name, I believe commit is self-explanatory and I added a test to prevent future regressions.
* Secondary CA configuration logic issue also looks self-explanatory at reading original change in https://github.com/hashicorp/consul/pull/12298

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/connect/ca/provider_consul.go
+++ b/agent/connect/ca/provider_consul.go
@@ -222,7 +222,13 @@ func (c *ConsulProvider) GenerateIntermediateCSR() (string, error) {
 		return "", err
 	}
 
-	csr, err := connect.CreateCACSR(c.spiffeID, signer)
+	uid, err := connect.CompactUID()
+	if err != nil {
+		return "", err
+	}
+	cn := connect.CACN("consul", uid, c.clusterID, c.isPrimary)
+
+	csr, err := connect.CreateCACSR(c.spiffeID, cn, signer)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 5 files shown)


dclaisse:
Hi @rboyer, can I do anything to help with this PR code review? FYI, we've successfully opened 2 new secondary CA datacenters since we deployed it internally.

manu-ns:
This is still valid

PR #14908, opened 2022-10-06
cache: prevent goroutine leak in agent cache

rboyer:
### Description

There is a bug in the error handling code for the Agent cache subsystem discovered by @boxofrad :

1. `NotifyCallback` calls `notifyBlockingQuery` which calls `getWithIndex` in a loop (which backs off on-error up to 1 minute)
2. `getWithIndex` calls `fetch` if there’s no valid entry in the cache
3. `fetch` starts a goroutine which calls `Fetch` on the cache-type, waits for a while (again with backoff up to 1 minute for errors) and then calls `fetch` to trigger a refresh

The end result being that every 1 minute `notifyBlockingQuery` spawns an ancestry of goroutines that essentially lives forever.

This PR ensures that the goroutine started by `fetch` cancels any prior goroutine spawned by the same line for the same key.

In isolated testing where a cache type was tweaked to indefinitely error, this patch prevented goroutine counts from skyrocketing.

### DIFF
--- a/agent/cache/cache.go
+++ b/agent/cache/cache.go
@@ -139,6 +139,10 @@ type Cache struct {
 	entries           map[string]cacheEntry
 	entriesExpiryHeap *ttlcache.ExpiryHeap
 
+	fetchLock    sync.Mutex
+	lastFetchID  uint64
+	fetchHandles map[string]fetchHandle
+
 	// stopped is used as an atomic flag to signal that the Cache has been
 	// discarded so background fetches and expiry processing should stop.
 	stopped uint32
@@ -148,6 +152,11 @@ type Cache struct {
 	options          Options
 	rateLimitContext context.Context
 	rateLimitCancel  context.CancelFunc
+}
+
+type fetchHandle struct {
+	id     uint64
+	stopCh chan struct{}
 }
 
 // typeEntry is a single type that is registered with a Cache.
@@ -225,6 +234,7 @@ func New(options Options) *Cache {
 		types:             make(map[string]typeEntry),
 		entries:           make(map[string]cacheEntry),
 		entriesExpiryHeap: ttlcache.NewExpiryHeap(),
+		fetchHandles:      make(map[string]fetchHandle),
 		stopCh:            make(chan struct{}),
 		options:           options,
 		rateLimitContext:  ctx,
@@ -614,8 +624,18 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
 	metrics.SetGauge([]string{"cache", "entries_count"}, float32(len(c.entries)))
 
 	tEntry := r.TypeEntry
-	// The actual Fetch must be performed in a goroutine.
-	go func() {
+
+	// The actual Fetch must be performed in a goroutine. Ensure that we only
+	// have one in-flight at a time, but don't use a deferred
+	// context.WithCancel style termination so that these things outlive their
+	// requester.
+	//
+	// By the time we get here the system WANTS to make a replacement fetcher, so
+	// we terminate the prior one and replace it.
+	handle := c.getOrReplaceFetchHandle(key)
+	go func(handle fetchHandle) {
+		defer c.deleteFetchHandle(key, handle.id)
+
 		// If we have background refresh and currently are in "disconnected" state,
 		// waiting for a response might mean we mark our results as stale for up to
 		// 10 minutes (max blocking timeout) after connection is restored. To reduce
@@ -664,6 +684,14 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
 		result, err := r.Fetch(fOpts)
 		if connectedTimer != nil {
 			connectedTimer.Stop()
+		}
+
+		// If we were stopped while waiting on a blocking query now would be a
+		// good time to detect that.
+		select {
+		case <-handle.stopCh:
+			return
+		default:
 		}
 
 		// Copy the existing entry to start.
@@ -820,13 +848,15 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
 			}
 
 			// If we're over the attempt minimum, start an exponential backoff.
-			if wait := backOffWait(attempt); wait > 0 {
-				time.Sleep(wait)
-			}
+			wait := backOffWait(attempt)
 
 			// If we have a timer, wait for it
-			if tEntry.Opts.RefreshTimer > 0 {
-				time.Sleep(tEntry.Opts.RefreshTimer)
+			wait += tEntry.Opts.RefreshTimer
+
+			select {
+			case <-time.After(wait):
+			case <-handle.stopCh:
+				return
 			}
 
 			// Trigger. The "allowNew" field is false because in the time we were
@@ -836,9 +866,44 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
 			r.Info.MinIndex = 0
 			c.fetch(key, r, false, attempt, true)
 		}
-	}()
+	}(handle)
 
 	return entry.Waiter
+}
+
+func (c *Cache) getOrReplaceFetchHandle(key string) fetchHandle {
+	c.fetchLock.Lock()
+	defer c.fetchLock.Unlock()
+
+	if prevHandle, ok := c.fetchHandles[key]; ok {
+		close(prevHandle.stopCh)
+	}
+
+	c.lastFetchID++
+
+	handle := fetchHandle{
+		id:     c.lastFetchID,
+		stopCh: make(chan struct{}),
+	}
+
+	c.fetchHandles[key] = handle
+
+	return handle
+}
+
+func (c *Cache) deleteFetchHandle(key string, fetchID uint64) {
+	c.fetchLock.Lock()
+	defer c.fetchLock.Unlock()
+
+	// Only remove a fetchHandle if it's YOUR fetchHandle.
+	handle, ok := c.fetchHandles[key]
+	if !ok {
+		return
+	}
+
+	if handle.id == fetchID {
+		delete(c.fetchHandles, key)
+	}
 }
 
 func backOffWait(failures uint) time.Duration {
--- a/agent/cache/cache_test.go
+++ b/agent/cache/cache_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/hashicorp/consul/acl"
 	"github.com/hashicorp/consul/lib/ttlcache"
 	"github.com/hashicorp/consul/sdk/testutil"
+	"github.com/hashicorp/consul/sdk/testutil/retry"
 )
 
 // Test a basic Get with no indexes (and therefore no blocking queries).
@@ -1750,12 +1751,22 @@ func TestCache_RefreshLifeCycle(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, true, result)
 
+	waitUntilFetching := func(expectValue bool) {
+		retry.Run(t, func(t *retry.R) {
+			c.entriesLock.Lock()
+			defer c.entriesLock.Unlock()
+			entry, ok := c.entries[key]
+			require.True(t, ok)
+			if expectValue {
+				require.True(t, entry.Fetching)
+			} else {
+				require.False(t, entry.Fetching)
+			}
+		})
+	}
+
 	// ensure that the entry is fetching again
-	c.entriesLock.Lock()
-	entry, ok := c.entries[key]
-	require.True(t, ok)
-	require.True(t, entry.Fetching)
-	c.entriesLock.Unlock()
+	waitUntilFetching(true)
 
 	requestChan := make(chan error)
 
@@ -1789,11 +1800,7 @@ func TestCache_RefreshLifeCycle(t *testing.T) {
 	}
 
 	// ensure that the entry is fetching again
-	c.entriesLock.Lock()
-	entry, ok = c.entries[key]
-	require.True(t, ok)
-	require.True(t, entry.Fetching)
-	c.entriesLock.Unlock()
+	waitUntilFetching(true)
 
 	// background a call that will wait for a newer version - will result in an acl not found error
 	go getError(5)
@@ -1814,11 +1821,7 @@ func TestCache_RefreshLifeCycle(t *testing.T) {
 
 	// ensure that the ACL not found error killed off the background refresh
 	// but didn't remove it from the cache
-	c.entriesLock.Lock()
-	entry, ok = c.entries[key]
-	require.True(t, ok)
-	require.False(t, entry.Fetching)
-	c.entriesLock.Unlock()
+	waitUntilFetching(false)
 }
 
 type fakeType struct {
--- a/.changelog/14908.txt
+++ b/.changelog/14908.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cache: prevent goroutine leak in agent cache
+```


dhiaayachi (commented):
Thank you both for finding and fixing this @rboyer and @boxofrad 
I'm wondering if we can fix this in a simpler way by using [singleflight](https://pkg.go.dev/golang.org/x/sync/singleflight) on the go routine spawn by fetch 🤔. We already have a key passed to fetch that we can leverage for this.

rboyer:
I think you'd need something like "preemptible singleflight" to match what this cache is currently doing. Basically if you try to do a fetch for `X` and there is already an `X`, cancel the old one and make a new one. `singleflight` instead would say to avoid making a new and wait for the old.

dhiaayachi:
> 

Mostly for my own knowledge but why do we need to cancel the old one and start a new one. Wouldn't that result in the same. Which is the element pointed by that key stored/refreshed in cache?

agent/cache/watch.go
```diff
@@ -106,148 +106,192 @@ func (c *Cache) notifyBlockingQuery(ctx context.Context, r getOptions, correlati
...
+	index = currIndex
+	failures = currFailures
 
-		// Check context hasn't been canceled
-		if ctx.Err() != nil {
-			return
-		}
+	// Ensure that each invocation terminates instead of stacking.
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
dhiaayachi, line 132: Wouldn't this cancel the context as soon as it's out of scope at the end of `notifyBlockingQueryOnce`?
If so that would terminate the go routine inside fetch for the current iteration not the former one. Am I missing something here?

agent/cache/watch.go
```diff
@@ -106,148 +106,192 @@ func (c *Cache) notifyBlockingQuery(ctx context.Context, r getOptions, correlati
...
 
 func (c *Cache) notifyPollingQuery(ctx context.Context, r getOptions, correlationID string, cb Callback) {
-	index := uint64(0)
-	failures := uint(0)
-
-	var lastValue interface{} = nil
+	var (
+		index         = uint64(0)
+		failures      = uint(0)
+		lastValue any = nil
```
dhiaayachi, line 187:
```suggestion
		index    uint64
		failures  uint
		lastValue any
```

agent/cache/watch.go
```diff
@@ -106,148 +106,192 @@ func (c *Cache) notifyBlockingQuery(ctx context.Context, r getOptions, correlati
...
+	failures = currFailures
+	lastValue = currValue
 
-		// Check context hasn't been canceled
-		if ctx.Err() != nil {
-			return
-		}
+	// Ensure that each invocation terminates instead of stacking.
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
dhiaayachi, line 217: I think this would work here because the wait is done at this level so by the time cancel is called we really want to cancel the context because we waited enough and it was not done.

agent/cache/watch.go
```diff
@@ -106,148 +106,192 @@ func (c *Cache) notifyBlockingQuery(ctx context.Context, r getOptions, correlati
...
+		// again where the age of the data is strictly equal to the MaxAge
+		// and then immediately have to re-fetch again. That wouldn't
+		// be terrible but it would expend a bunch more cpu cycles when
+		// we can definitely avoid it.
+		wait += lib.RandomStagger(r.Info.MaxAge / 16)
 	}
+
+	select {
+	case <-time.After(wait):
+	case <-ctx.Done():
```
dhiaayachi, line 293: What happen if we call a cancel on a done context does it add an error to the context or have no effect?

dhiaayachi (commented):
@rboyer I had a few questions about this. I'm not sure I understand the logic for `notifyBlockingQueryOnce`.
If you think it's easier we can do a synchronous review of this early next week.

agent/cache/cache.go
```diff
@@ -819,6 +820,12 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
 				return
 			}
 
+			select {
+			case <-ctx.Done():
+				return
+			default:
+			}
```
boxofrad, line 827: I notice you've used `Err()` and this non-blocking select pattern interchangeably to check whether the context has been canceled. Is there any reason to prefer the more verbose version? If not, shall we go with `Err()` throughout?

agent/cache/watch.go
```diff
@@ -106,148 +106,192 @@ func (c *Cache) notifyBlockingQuery(ctx context.Context, r getOptions, correlati
...
+	}
+}
+func (c *Cache) notifyBlockingQueryOnce(
+	ctx context.Context,
+	r getOptions,
+	correlationID string,
+	cb Callback,
+	currIndex uint64,
+	currFailures uint,
+) (index uint64, failures uint) {
```
boxofrad, line 125: nit: Could we avoid the named return values? I think it makes it harder to follow 🙏🏻

boxofrad (approved):
Thanks for picking this up @rboyer 🙌🏻 

This approach makes sense to me, but I'll reiterate that I don't fully understand everything that's going on here (so take my approval with a pinch of salt!).

Is there somebody else who has more context? Maybe @banks or @mkeeler?

agent/cache/watch.go
rboyer: So I moved the body of the cancelable portion into `notifyBlockingQueryOnce` so that I could `defer` the cancel and avoid trying to do that safely without defer in a loop.

If you look at `notifyBlockingQuery` it doesn't even sleep so it immediately re-enters `notifyBlockingQueryOnce` and creates a new one.

In a prior iteration of this I tried to do the "new one cancels the prior one" but 
a) it had some gross conditional cancellation
b) the linter had red flags go off at suspected escaped cancels because of (a)

agent/cache/watch.go
rboyer: From the stdlib docs:
> A CancelFunc tells an operation to abandon its work. A CancelFunc does not wait for the work to stop. A CancelFunc may be called by multiple goroutines simultaneously. After the first call, subsequent calls to a CancelFunc do nothing.

rboyer:
OK I changed this 100%.

Now it solely deals with the goroutine that `fetch` spawns. Rather than handle the cancelling at the level of `getWithIndex` instead the fetch-spawned goroutine (which includes the child goroutine chain it spawns) is assigned both a unique identity and a stop channel which are stored in a map.

If another fetch for the same key gets to the current line of code which fires off the goroutine it cancels the former one before spawning the new one.

To ensure the tracking map is cleaned up at the right time, the unique identity is used in a defer in the goroutine so that the item in the map for this key is only deleted if the identity matches.

agent/cache/cache.go
```diff
@@ -819,14 +866,28 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
...
 
+			select {
+			case <-handle.stopCh:
+				return
+			default:
+			}
+
 			// If we're over the attempt minimum, start an exponential backoff.
 			if wait := backOffWait(attempt); wait > 0 {
-				time.Sleep(wait)
```
rboyer, line 824: After the initial refactor it turned out that the goroutine leak was **back!** 

Then I changed `time.Sleep` into cancellable ones via the `select` pattern and the leak went away.

agent/cache/cache_test.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/hashicorp/consul/acl"
 	"github.com/hashicorp/consul/lib/ttlcache"
 	"github.com/hashicorp/consul/sdk/testutil"
+	"github.com/hashicorp/consul/sdk/testutil/retry"
```
rboyer, line 21: The changes in this file remove a flake from the `-race` version of these tests.

banks:
Thanks all for finding this and working on it.

At a first glance I'm not super clear on where the original lead happened - the cache was _supposed_ to single flight requests so we didn't support cancellation because there should only ever be at most one `Fetch` being called at anyone time for a given type and key, those goroutines might continue for a long time (the refresh loop) even if the item is no longer being watched. But don't "leak" in the sense that when the cache item's TTL expires (72 hours in some cases) then the entry is removed and the background refresh loop stopped next time it unblocks (max 10 mins later).

The PR description mentions error handling but I'm not quite sure what are the actual steps to cause the leak? Could you explain how the leak happens as far as you understand RB?

The PR here looks reasonable, but I'm not really clear on how the leak shows up so it's a little hard to judge. I'm also interested in the details being super clearly documented here because I'm super wary of how complex agent cache is - we've often end up introducing other bugs when we try to fix bugs in here! I'm assuming it is a correct fix, but would be awesome to have really good info about exactly _why_ we think it's the right fix for next time we have to come back here and understand!

agent/cache/cache.go
```diff
@@ -836,11 +872,46 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
 			r.Info.MinIndex = 0
 			c.fetch(key, r, false, attempt, true)
```
rboyer, line 873: This is where `fetch` calls itself, but it does so in a goroutine and then returns, which exits the original goroutine. The net at the end of this call is that you have the same number of goroutines when one exits.

banks:
@rboyer and I looked at this together yesterday. I want to record my understanding before it slips through my fingers again!

The _intended_ behaviour (simplified from actual code for clarity) here for refreshing cache types is:

 1. Client fetch (either via `Get` or `Notify` both internally call `getWithIndex`
 2. `getWithIndex` calls `fetch` which returns the waiter for the cache item (creating it if it didn't already exist, but just returning the waiter if it did **). The client goroutine (G1) is now blocked on that `Waiter`.
 3. meanwhile `fetch`, if it found _no existing, "fetching" cache entry_  it created one with `Fetching == true`, spawned a new anonymous goroutine (G2) that I'll refer to as `fetch.internal`.
 4. `fetch.internal` does the actual blocking RPC to servers (possibly blocking for a while). When it is done either with a new result, the same result (blocking timeout) or error, it updates the cache entry and sets `Fetching == false`. Once it's released the lock it signals the _old_ `Waiter` which unblocks any waiting goroutines.
 5. G1 (client goroutine) unblocks, finds the update or error and returns (usually)
 6. Meanwhile... the `fetch.internal` prepares to continue. It waits if there was an error and some backoff is needed, it also waits if there is a `RefreshTimer` configured for that type. **All the time it's waiting, the cache entry exists but with `Fetching == false`**
 7. When it's done waiting, the `fetch.internal` goroutine recursively calls `fetch` in the same goroutine (G2)
 8. G2 now makes it through the `fetch` method back to the point where it decides if it should continue - since `Fetching == false` (probably, we'll get to that) it effectively goes back to 3 above - sets entry to `Fetching == true`, spawns a new goroutine (G3) to run a new `fetch.internal`

The intent of all this is that the cache has "single flighted" the request so that many concurrent requests will just see the existing entry has a waiter and will wait on that. The background refresh "loop" (which is really tail call recursion) continues until the cache entry is evicted which breaks the tail call) which won't happen until TTL has passed without an active request for that entry.

## The Bug

So the _intention_ is that only a single fetching routine is active at any one time for a given cache key... but there is an major flaw in the design that makes that not true in actually quite a lot of cases!

For the whole time that the `fetch.internal` call is waiting before recusing back into `fetch` for a refresh, a new request for the same key will see `Fetching == false` and so will spawn a new separate `fetch.internal`.

So it's fatally flawed! But because of the complexities of the recursive calls, it's _really_ hard to fix cleanly.

 * If we just don't set `Fetching == false` once the RPC is done then when the same goroutine re-enters `fetch` it will see `Fetching == false` and so it will exit, breaking the background refresh loop.
 * If we add another field that explicitly marks the entry as "refreshing" and use that to decide whether to spawn a new goroutine or not too then you hit the same problem as above.
 * Singleflight doesn't work because that also breaks the recursive refresh loop in a similar way - it sees the request as already "in progress" when it re-enters and so stops, breaking the refresh loop for everyone.

@rboyer and I concluded that this whole thing could probably be a lot simpler to reason about if we refactored it so that the background refresh was a single method with a for loop instead of recursive calls. I don't know if RB has looked at that since we spoke but our guess is that it's a big enough change that we'd be very nervous about rushing it in with the current history of breaking agent cache whenever we touch it and the desire to fix this issue soon.

So the approach @rboyer took here fixes the bug (we _think_ without causing any new ones) specifically because:

 1. There is a separate data structure tracking which keys have active fetches. (The main cache _could_ do this but it's complicated because we replace the cache entry each time it updates)
 2. The separate data structure tracks not just boolean state, but also identifies each entrance into `fetch.internal` with a unique ID.
 3. Each entrance into `fetch.internal` closes the chan of the previous handle - which is checked after blocking in the RPC AND waiting for any backoff before re-entering `fetch`. Which breaks the old goroutine chain.
 4. Each `fetch.internal` _defers_ a cleanup of its own handle, but leaves the state alone if another routine has taken over since it has a known ID.

Writing this up, I think I've spotted _two_ bugs with this approach which I'll write up inline. I _think_, if valid, we can fix those though.

The tl;dr - this is very much a band-aid fix for a bug that is a violation of the core design goal of the cache. But we think it's the least likely thing to cause additional bugs and so be safest to merge quickly (this week) so be in the latest release.

agent/cache/cache.go
```diff
@@ -819,14 +847,22 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
...
 			// If we have a timer, wait for it
 			if tEntry.Opts.RefreshTimer > 0 {
-				time.Sleep(tEntry.Opts.RefreshTimer)
+				totalWait += tEntry.Opts.RefreshTimer
+			}
+
+			select {
+			case <-time.After(totalWait):
+			case <-handle.stopCh:
+				return
```
banks, line 865: What happens if another `fetch` happens in another goroutine right after this point but before our own `fetch` call?

 * other `fetch` will see our handle, close our chan (no effect) and replace it with their own with a greater ID (say 10)
 * other `fetch` will then spawn `fetch.internal` and start an RPC 
 * we continue into our own `fetch` see the handle with ID 10 and close it's chan, make our new one with ID 11, and then re-enter `fetch.internal`
 * other `fetch.internal` returns from RPC, see's it's chan closed and exits (wasted RPC)
 * we complete our RPC and assuming no new one has started, update the cache entry.

**Conclusion:** seems safe although it does mean we still end up doing some redundant RPCs. I think that's OK as a quick fix.

Hmmm so one thing this could possible cause would be effectively a DOS of servers on error. Normally we'd have the retry backoff protecting servers if repeated errors happen. It would look like this:

 1. `fetch.internal 1` gets an error, updates entry, starts backoff wait
 2. client loop has a smaller backoff (or none) so retries which results in a new `fetch 2`
 3. `fetch 2` will see `Fetching == false` and so spawn a `fetch.internal 2` while closing the stop chan for `fetch.internal 1`. it will _immediately_ make another request to servers even though we are supposed to be waiting.
 4. eventually `fetch.internal 1` finishes it's backoff wait and see's it's been deposed and so just returns
 5. `fetch.internal 2` also hits error, starts backoff
 6. client retries sooner again etc.

We _might_ be OK because our internal cache clients generally use `Notify` which _probably_ does the right thing in that it does its own backoff loop IIRC. **But this one seems at least a bit risky that this fix potentially means that we'll bypass error backoffs?**

agent/cache/cache.go
```diff
@@ -819,14 +847,22 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
...
 			}
 
 			// If we have a timer, wait for it
 			if tEntry.Opts.RefreshTimer > 0 {
-				time.Sleep(tEntry.Opts.RefreshTimer)
+				totalWait += tEntry.Opts.RefreshTimer
+			}
+
+			select {
+			case <-time.After(totalWait):
```
banks, line 863: I find this a little subtle to read - it takes me a while to convince myself that the first case is not falling through to the `return` as well...

It's a nit as most go programmers are probably OK with this but I'd consider adding a comment to explicitly call out that the first case is just continuing execution. Or re-order the cases?

agent/cache/cache.go
```diff
@@ -836,11 +872,46 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
...
 
+func (c *Cache) getOrReplaceFetchHandle(key string) fetchHandle {
+	c.fetchLock.Lock()
+	defer c.fetchLock.Unlock()
+
+	if prevHandle, ok := c.fetchHandles[key]; ok {
+		close(prevHandle.stopCh)
+	}
+
+	c.lastFetchID++
```
dhiaayachi, line 888: Why do we need to manage an ID at the cache level? I think it will be sufficient to use a unique id per key. Which mean that we can use the last `fetchHandles` id and increment it.
If we do so we could possibly use a sync Map for this.

dhiaayachi (changes requested):
This look good! I think this approach would work. I have few questions which could lead to simplifications.

agent/cache/cache.go
banks: > I think it will be sufficient to use a unique id per key

This is to identity specific goroutines for the _same_ key which is the exact bug here. Does my explanation above help explain why it's needed?

Or maybe I'm misundestanding.

> we could possibly use a sync Map for this

In general `sync.Map` is almost never the right choice - they are specifically designed for cases like reloadable config where reads are extremely common and updates are so rare that it's effectively never. This case will be writing the value every minutes and only marginally less than it reads I'd guess which means a mutex is almost certainly the better approach unless profiling tells us otherwise!

From the docs:

> The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.

rboyer:
Here is the more complete attempt at fixing this problem by removing the weird recursion and then introducing a new tracking variable to the cacheEntry type: https://github.com/hashicorp/consul/pull/14956

agent/cache/cache.go
```diff
@@ -819,14 +847,22 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
...
 			if wait := backOffWait(attempt); wait > 0 {
-				time.Sleep(wait)
+				totalWait += wait
 			}
 
 			// If we have a timer, wait for it
 			if tEntry.Opts.RefreshTimer > 0 {
-				time.Sleep(tEntry.Opts.RefreshTimer)
+				totalWait += tEntry.Opts.RefreshTimer
+			}
```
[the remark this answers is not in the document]
rboyer: :+1:

agent/cache/cache.go
```diff
@@ -819,14 +847,22 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
 				return
 			}
 
+			var totalWait time.Duration
+
 			// If we're over the attempt minimum, start an exponential backoff.
 			if wait := backOffWait(attempt); wait > 0 {
-				time.Sleep(wait)
+				totalWait += wait
```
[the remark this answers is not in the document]
rboyer: :+1:

agent/cache/cache.go
rboyer: > Hmmm so one thing this could possible cause would be effectively a DOS of servers on error. 
That's actually already happening on error with the existing code, since getting an error is `Fetching=false`.

rboyer:
I've rephrased the follow-on [PR](https://github.com/hashicorp/consul/pull/14956) to better state that it fixes the RPC issue since this PR fixes the goroutine leak. That should make it easier to rationalize landing both changes on `main` for `1.15.0`

agent/cache/cache.go
dhiaayachi: Thank you both for explaining this. Til that sync.Map is almost never a good choice 😓

agent/cache/cache.go
```diff
@@ -614,8 +624,18 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
...
+
+	// The actual Fetch must be performed in a goroutine. Ensure that we only
+	// have one in-flight at a time, but don't use a deferred
+	// context.WithCancel style termination so that these things outlive their
+	// requester.
+	//
+	// By the time we get here the system WANTS to make a replacement fetcher, so
+	// we terminate the prior one and replace it.
+	handle := c.getOrReplaceFetchHandle(key)
+	go func(handle fetchHandle) {
```
dhiaayachi, line 636: Maybe we can consider extracting this giant anonymous function into a function

rboyer: That's a side effect of the other PR (https://github.com/hashicorp/consul/pull/14956) already.

Actually doing it has some tracking state that is currently part of the closure, but would need to be adjusted as input/output parameters and it gets fiddly.

I did it on the road to #14956 and I wouldn't want to do that here on the band-aid for fear of introducing an inadvertent bug in the data flow.

PR #14916, opened 2022-10-07
bug: fix goroutine leaks caused by incorrect usage of `WatchCh`

boxofrad:
### Description
[memdb's `WatchCh` method](https://pkg.go.dev/github.com/hashicorp/go-memdb#WatchSet.WatchCh) creates a goroutine that will publish to the returned channel when the watchset is triggered or the given context is canceled. Although this is called out in its godoc comment, it's not obvious that this method creates a goroutine who's lifecycle you need to manage.

In the xDS capacity controller, we were calling `WatchCh` on each iteration of the control loop, meaning the number of goroutines would grow on each autopilot event until there was catalog churn.

In the catalog config source, we were calling `WatchCh` with the background context, meaning that the goroutine would keep running after the sync loop had terminated.

### DIFF
--- a/.changelog/14916.txt
+++ b/.changelog/14916.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+server: fix goroutine/memory leaks in the xDS subsystem (these were present regardless of whether or not xDS was in-use)
+```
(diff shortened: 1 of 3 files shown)


riddhi89 (approved):
Good catch! LGTM

pglass (approved):

PR #14921, opened 2022-10-07
ui: hide node name and meta on agentless instances

WenInCode:
### Description
- Hides node name and node checks on service instances that are using agentless
- Hides the external source badges on the individual instances if they are all matching (shows at the top).

- To test the badges, I set the service instances in the mock-api to be all have one kind of source, and then set the node count to be something smaller like 3. After a few refreshes I got an page with all instances with the same source type.
- To test out the node meta removal, I set the external-source to always be kubernetes in the mock-api, which currently also sets the `synthetic-node: true` meta attribute for agentless. 

### Links
- [NET-950](https://hashicorp.atlassian.net/browse/NET-950)
- [Design](https://www.figma.com/file/S20gHKN9SoKfkabNjWKVuE/Agentless?node-id=2%3A75)
- [Discussion about changing logic of showing the badges](https://hashicorp.slack.com/archives/C9X0VGJNL/p1665157455446329)

### Screenshots
Screen Shot 2022-10-07 at 10 21 13 AM
Screen Shot 2022-10-07 at 4 34 58 AM

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


ui/packages/consul-ui/app/components/consul/service-instance/list/index.hbs
```diff
@@ -45,18 +45,24 @@ as |proxy|}}
...
-    <Consul::InstanceChecks
-      @type="node"
-      @items={{reject-by 'ServiceID' '' checks}}
-    />
+
+    {{#unless item.Node.Meta.synthetic-node}}
+      <Consul::InstanceChecks
+        @type="node"
+        @items={{reject-by 'ServiceID' '' checks}}
+        data-test-node-health-checks
```
evrowe, line 63: Good seeing the addition of more `data-test-*` attributes here

ui/packages/consul-ui/app/components/consul/service-instance/list/index.js
```diff
@@ -0,0 +1,12 @@
+import Component from '@glimmer/component';
+
+export default class ServiceInstanceList extends Component {
+  get areAllExternalSourcesMatching() {
+    const firstSource = this.args.items[0]?.Service?.Meta?.['external-source'];
+
+    const matching = this.args.items.every(
+      (instance) => instance.Service?.Meta?.['external-source'] === firstSource
+    );
```
evrowe, line 9: 👨‍🍳 🤌🏻

evrowe (approved):
Looks solid, :shipit: !

WenInCode:
I realized I might have missed something from the [designs](https://www.figma.com/file/S20gHKN9SoKfkabNjWKVuE/Agentless?node-id=101%3A12100). They mention that the external source should read `Synced by [source]`. Currently, in agent world it reads `Registered via [source]`. I've made the assumption that this should be the case whenever an instance is agentless and has an external source.

**A technical note:** the `<Consul::ExternalSource />` component currently takes a Service as the item. I am getting the `synthetic-node` meta key from the `Node` to determine if it's agentless. Instead of changing the external source to take a different item, I've added an `agentless` argument to check and determine the prefix. Pros to this is that it doesn't affect any other usages, downsides are that you have to manually pass in that it's agentless.

Screen Shot 2022-10-07 at 4 29 40 PM
Screen Shot 2022-10-07 at 4 29 20 PM

It would be great to have @jnwright confirm that my understanding of this change is correct, and then it would also be good to get another set of eyes on the last commit by @evrowe or @LevelbossMike on Monday.

Appreciate it folks!

jnwright:
@tylert 🤦 no, that was my mistake. I pulled it from the master design file, and it seems like it's incorrect in there for some reason. I'm sorry! It should say "registered via.." for everything. I usually try to annotate new changes on the side, so if it's not annotated then I made a mistake 😞

WenInCode:
@jnwright no worries at all, easy to revert that change!

PR #14922, opened 2022-10-07
gRPC server metrics

pglass:
### Description
This updates the internal and external gRPC servers with the following labeled metrics, where the `server_type` label is either `internal` or `external`.

Server metrics:
* `grpc.server.connection.count{server_type=internal|external}`
* `grpc.server.connections{server_type=internal|external}`
* `grpc.server.request.count{server_type=internal|external}`
* `grpc.server.stream.count{server_type=internal|external}`
* `grpc.server.streams{server_type=internal|external}`

Client metrics (a client only exists for the internal gRPC server)
* `grpc.client.connection.count{server_type=internal}`
* `grpc.client.connections{server_type=internal}`
* `grpc.client.request.count{server_type=internal}`

These are the same metrics that currently exist for the internal gRPC server, except those are now emitted with a `server_type = internal` label.

The stats handler is moved to the `agent/grpc-middleware` package, and some shared unit test stuff is moved to the `agent/grpc-middleware/testutil` package.

* Unit tests

### PR Checklist
* [x] updated test coverage
* [x] not a security concern


agent/grpc-external/stats_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"google.golang.org/grpc"
+
+	"github.com/hashicorp/go-hclog"
+
+	"github.com/hashicorp/consul/agent/grpc-middleware/testutil"
+	"github.com/hashicorp/consul/agent/grpc-middleware/testutil/testservice"
+	"github.com/hashicorp/consul/proto/prototest"
+)
+
+func TestServer_EmitsStats(t *testing.T) {
```
pglass, line 23: This copies the [`TestHandler_EmitStats`](https://github.com/hashicorp/consul/blob/4abad02abdd892a05001280748fa496624d7758f/agent/grpc-internal/stats_test.go#L25) test. I did some deduplication of the helpers by moving things over to `grpc-middleware/testutil`. (I could do more but kind of wanted a sanity check before I go too far.)

agent/grpc-external/stats_test.go
```diff
@@ -0,0 +1,130 @@
...
+	sink, reset := patchGlobalMetrics(t)
```
boxofrad, line 24: What do you think of passing the metrics object as a parameter to `NewServer` so we don't have to change global state like this?

```Go
func NewServer(logger agentmiddleware.Logger, metrics *metrics.Metrics) *grpc.Server {
	if metrics == nil {
		metrics = agentmiddleware.DefaultMetrics
	}
```

boxofrad: I also appreciate that these tests were copied from `grpc-internal` so please feel free to leave as-is! 🙇🏻‍♂️

boxofrad (approved):
Nice work! 🚢

PR #14923, opened 2022-10-07
Disallow peering to the same cluster.

hashi-derek:


agent/rpc/peering/service.go
```diff
@@ -471,32 +471,16 @@ func (s *Server) Establish(
...
+func (s *Server) validatePeeringLocality(token *structs.PeeringToken) error {
 	serverName, _, err := s.Backend.GetTLSMaterials(false)
 	if err != nil {
 		return fmt.Errorf("failed to fetch TLS materials: %w", err)
 	}
-
-	if serverName == token.ServerName && peering == nil {
-		return fmt.Errorf("conflict - peering token's server name matches the current cluster's server name, %q, but there is no record in the database", serverName)
+	if serverName == token.ServerName {
+		return fmt.Errorf("cannot create a peering within the same cluster %q", serverName)
```
DanStough, line 482: I wonder if we could still check if `!acl.EqualPartitions(peering.GetPartition(), partition)` here and throw a different message telling operators to use service-exports for partitions instead of peering.

DanStough (approved):
Good to me as-is, but adding a suggestion for error handling.

agent/rpc/peering/service.go
hashi-derek: That could be useful, but it adds extra complexity to the codebase with the extra query, parameter, and branching statement. I'm fine with either way. Do you think this provides enough value to users to warrant it?

Maybe I could change it to say the following always:

`cannot create a peering within the same cluster %q. Refer to the 'exported-services' config documentation if you want to export between partitions without peering.`

DanStough: I'm fine with that message, but I'm not sure how others would feel about referencing an enterprise feature in OSS logs. IMHO anything we can do now to point people to the most effective patterns without filing an issue is worth it.

PR #14924, opened 2022-10-07
Service http checks data source for agentless proxies

riddhi89:
### Description
Adds another datasource for `proxycfg.HTTPChecks`, for use on server agents. Typically these checks are performed by local client agents and there is no equivalent of this in agentless (where servers configure consul-dataplane proxies).
Hence, the data source is mostly a no-op on servers but in the case where the service is present within the local state, it delegates to the cache data source.

#### Background
Using the cache data source for configuring agentless proxies when the service is not in the server local state causes an increase in the http checks fetch error rate which in turn causes a spike in memory usage (eventually leading to server OOM kill). 
Although this issue was only the symptom and the current PR solves this symptom ; the root cause fix of the memory leak is being handled here: https://github.com/hashicorp/consul/pull/14908. The current PR also removes the red-herring that was creating noise in the server logs. 

Ran the consul agent in server mode locally. Registered a service and established an xDS session with the server via consul-dataplane. The below warnings go away with the fix in the PR
```2022-10-07T13:17:03.060-0700 [WARN]  agent.cache: handling error in Cache.Notify: cache-type=service-http-checks error="Internal cache failure: service 'db-1' not in agent state" index=0```

### DIFF
--- a/agent/proxycfg-glue/glue.go
+++ b/agent/proxycfg-glue/glue.go
@@ -55,6 +55,8 @@ type Store interface {
 //
 // Note: there isn't a server-local equivalent of this data source because
 // "agentless" proxies obtain certificates via SDS served by consul-dataplane.
+// If SDS is not supported on consul-dataplane, data is sourced from the server agent cache
+// even for "agentless" proxies.
 func CacheCARoots(c *cache.Cache) proxycfg.CARoots {
 	return &cacheProxyDataSource[*structs.DCSpecificRequest]{c, cachetype.ConnectCARootName}
 }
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: agent/proxycfg-glue/glue.go [hunk 1]
```diff
@@ -55,6 +55,8 @@ type Store interface {
 //
 // Note: there isn't a server-local equivalent of this data source because
 // "agentless" proxies obtain certificates via SDS served by consul-dataplane.
+// If SDS is not supported on consul-dataplane, data is sourced from the server agent cache
+// even for "agentless" proxies.
```
boxofrad, line 59: Thanks for clarifying these overly optimistic comments! 😅

agent/proxycfg-glue/service_http_checks.go
```diff
@@ -0,0 +1,39 @@
...
+	cacheSource proxycfg.HTTPChecks
+	localState  *local.State
+}
+
+func (c serverHTTPChecks) Notify(ctx context.Context, req *cachetype.ServiceHTTPChecksRequest, correlationID string, ch chan<- proxycfg.UpdateEvent) error {
+	// if the service exists in the server agent local state, delegate to the cache data source
+	if c.localState.ServiceExists(structs.ServiceID{ID: req.ServiceID, EnterpriseMeta: req.EnterpriseMeta}) {
+		return c.cacheSource.Notify(ctx, req, correlationID, ch)
+	}
+	c.deps.Logger.Debug("service-http-checks: no-op")
```
boxofrad, line 37: Maybe worth tagging the log with the service ID?

agent/proxycfg-glue/service_http_checks.go
```diff
@@ -0,0 +1,39 @@
...
+	deps        ServerDataSourceDeps
+	cacheSource proxycfg.HTTPChecks
+	localState  *local.State
+}
+
+func (c serverHTTPChecks) Notify(ctx context.Context, req *cachetype.ServiceHTTPChecksRequest, correlationID string, ch chan<- proxycfg.UpdateEvent) error {
+	// if the service exists in the server agent local state, delegate to the cache data source
+	if c.localState.ServiceExists(structs.ServiceID{ID: req.ServiceID, EnterpriseMeta: req.EnterpriseMeta}) {
+		return c.cacheSource.Notify(ctx, req, correlationID, ch)
+	}
```
boxofrad, line 36: Service IDs are technically only unique within a node. So it's possible (albeit unlikely) that the xDS stream is for a service with the same ID registered to a different node, and returning this node's checks would be incorrect.

I think for completeness we should probably extend the `ServiceHTTPChecksRequest` to include the node name and compare it to the local node name.

Sorry! I should've mentioned this before.

boxofrad (commented):
This is looking great, nice job! 🙌🏻 

Just one thing I think we need to address. Totally my fault for not mentioning it earlier, sorry! 🙈

agent/proxycfg-glue/service_http_checks.go
riddhi89: Makes sense!  Updated!

boxofrad (approved):

PR #14925, opened 2022-10-07
tests: Fix flaky upgrade/compatibility test

eculver:
### Description
In #13658 (1.13+), we accidentally introduced a backwards incompatibility in our subscription/streaming API. Specifically, we 
[removed a field from the `SubscribeRequest` ](https://github.com/hashicorp/consul/pull/13658/commits/c7f491d174bb18c469f9fce4ee8cfad31bd78de5#diff-b23e1b89cf62b2990d0dcfcb707064c0e50fb8d1ba3e8d996807f0d05630276bL28) that gets sent for health requests. This is fine when all clients and servers are on the same version, but in the rare case where clients are on newer versions than the servers, it means that the requests they generate won't be compatible with the servers they are talking to. 

Ideally, we would remediate this issue in 1.13 and restore the compatibility, but that doesn't help us right now for the 1.12 tests which are really flaky. What makes them so flaky is the fact the test cases specifically target these mixed modes, in both directions, and the clients that are being used are always using the latest version; if the server that the client used in the test happens to be one of the older servers, the test will fail. Depending on the test case, this can be as much at 2/3rds of the time. 

My plan is to fix this here so that 1.12 builds are much less flaky and then also patch 1.13 to restore the compatibility and then remove the code again in our 1.14 since users should no longer hit this issue at that point since they shouldn't be doing 1.12->1.14 upgrades.

- Run a test with `-count=3` and you will most likely see a failure. With this fix, you never see the failure.

### Links
- https://github.com/hashicorp/consul/pull/13658

### DIFF
--- a/test/integration/consul-container/upgrade/healthcheck_test.go
+++ b/test/integration/consul-container/upgrade/healthcheck_test.go
@@ -41,7 +41,7 @@ func TestTargetServersWithLatestGAClients(t *testing.T) {
 
 	go func() {
 		service, q, err := client.Health().Service(serviceName, "", false, &api.QueryOptions{WaitIndex: index})
-		if err == nil && q.QueryBackend != api.QueryBackendBlockingQuery {
+		if err == nil && q.QueryBackend != api.QueryBackendStreaming {
 			err = fmt.Errorf("invalid backend for this test %s", q.QueryBackend)
 		}
 		if err != nil {
--- a/test/integration/consul-container/upgrade/healthcheck_test.go
+++ b/test/integration/consul-container/upgrade/healthcheck_test.go
@@ -153,7 +153,8 @@ func TestMixedServersMajorityTargetGAClient(t *testing.T) {
 				HCL: `node_name="` + utils.RandName("consul-server") + `"
 					log_level="TRACE"
 					bootstrap_expect=3
-					server=true`,
+					server=true
+					rpc { enable_streaming=false }`,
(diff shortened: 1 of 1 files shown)


boxofrad:
> Ideally, we would remediate this issue in 1.13 and restore the compatibility

While I'm definitely in favor of maintaining compatibility where it's possible, I think our recommended upgrade strategy is to roll the servers before the client agents (as client agents will inevitably rely on APIs that don't exist in prior versions).

That said, I've seen cases where K8s users have run into this by upgrading the Helm chart without realizing that it upgrades the client version. So I think it's worth the relatively low effort to restore compatibility here.

boxofrad (approved):
Thanks for fixing this! Sorry for the disruptive change 😅

eculver:
> > Ideally, we would remediate this issue in 1.13 and restore the compatibility
> 
> While I'm definitely in favor of maintaining compatibility where it's possible, I think our recommended upgrade strategy is to roll the servers before the client agents (as client agents will inevitably rely on APIs that don't exist in prior versions).
> 
> That said, I've seen cases where K8s users have run into this by upgrading the Helm chart without realizing that it upgrades the client version. So I think it's worth the relatively low effort to restore compatibility here.

Yeah, totally agree. I had similar thoughts about whether it would make sense to address given it being outside our guidance but there are certainly sharp edges as you mentioned that I think make it worthwhile to try to fix.

PR #14927, opened 2022-10-10
Prevent exporting consul service to peers.

hashi-derek:
Prior to this change, it was possible for the "consul" service to be exported. This change prevents it from being exported via wildcards, explicit exports, and discovery chains (resolvers, splitters, routers).

### DIFF
--- a/agent/consul/state/peering.go
+++ b/agent/consul/state/peering.go
@@ -784,6 +784,10 @@ func exportedServicesForPeerTxn(
 	// - use connect native mode
 
 	for _, svc := range conf.Services {
+		// Prevent exporting the "consul" service.
+		if svc.Name == structs.ConsulServiceName {
+			continue
+		}
 		svcMeta := acl.NewEnterpriseMetaWithPartition(entMeta.PartitionOrDefault(), svc.Namespace)
 
 		sawPeer := false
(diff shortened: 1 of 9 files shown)


agent/consul/state/peering.go
```diff
@@ -852,18 +865,24 @@ func exportedServicesForPeerTxn(
...
-			if err != nil {
-				return fmt.Errorf("failed to get discovery chain targets for service %q: %w", svc, err)
-			}
+		idx, targets, err := discoveryChainOriginalTargetsTxn(tx, ws, dc, svc.Name, &svc.EnterpriseMeta)
+		if err != nil {
+			return fmt.Errorf("failed to get discovery chain targets for service %q: %w", svc, err)
+		}
+		if idx > maxIdx {
+			maxIdx = idx
+		}
```
DanStough, line 874: Does moving this out of the check for L4 protocols change the semantics of when we update the maxIdx? Seems like we might want to wait to increment until we decide if the targets are valid.

DanStough (commented):
I just have one question.

agent/consul/state/peering.go
hashi-derek: Ah, I see what you mean. So maybe move the `if idx > maxIdx{` until right before the return for the function?

DanStough (approved):
LGTM 🚀

PR #14932, opened 2022-10-10
Fix incorrect backoff-wait logic.

hashi-derek:
This change fixes a situation where every call to `syncViaBlockingQuery` gets exponentially slower, until reaching a cap. The change now modifies the logic to exponentially backoff and only sleep in the event of a failure. The wait time also is reset back to normal whenever a successful call is made.

### DIFF
--- a/agent/grpc-external/services/peerstream/subscription_blocking.go
+++ b/agent/grpc-external/services/peerstream/subscription_blocking.go
@@ -93,30 +93,27 @@ func (m *subscriptionManager) syncViaBlockingQuery(
 
 	store := m.getStore()
 
-	for {
+	for ctx.Err() == nil {
 		ws := memdb.NewWatchSet()
 		ws.Add(store.AbandonCh())
 		ws.Add(ctx.Done())
 
-		if result, err := queryFn(ctx, store, ws); err != nil && ctx.Err() == nil {
+		if result, err := queryFn(ctx, store, ws); err != nil {
+			// Return immediately if the context was cancelled.
+			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+				return
+			}
 			logger.Error("failed to sync from query", "error", err)
-
+			waiter.Wait(ctx)
 		} else {
 			select {
 			case <-ctx.Done():
 				return
 			case updateCh <- cache.UpdateEvent{CorrelationID: correlationID, Result: result}:
+				waiter.Reset()
 			}
-
 			// Block for any changes to the state store.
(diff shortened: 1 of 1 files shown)


hashi-derek:
Not sure if I should be adding a changelog entry for beta features that aren't technically fully-released yet.

agent/grpc-external/services/peerstream/subscription_blocking.go
```diff
@@ -93,30 +93,27 @@ func (m *subscriptionManager) syncViaBlockingQuery(
...
-
 			// Block for any changes to the state store.
 			ws.WatchCtx(ctx)
 		}
-
-		err := waiter.Wait(ctx)
-		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
-			return
-		} else if err != nil {
-			logger.Error("failed to wait before re-trying sync", "error", err)
```
huikang, line 119: Are we missing the case to emit this logger error?

agent/grpc-external/services/peerstream/subscription_blocking.go
```diff
@@ -93,30 +93,27 @@ func (m *subscriptionManager) syncViaBlockingQuery(
 
 	store := m.getStore()
 
-	for {
+	for ctx.Err() == nil {
```
huikang, line 96: Could you elaborate why `ctx.Err() == nil` is in the for condition? In the original version, the error msg will be caught by 
```go
} else if err != nil {
			logger.Error("failed to wait before re-trying sync", "error", err)
```
Now we seem to miss handling the error.

agent/grpc-external/services/peerstream/subscription_blocking.go
hashi-derek: The `waiter.Wait` command can only error whenever the context is cancelled, according to the documentation on the method, so the old `else-if err != nil` should be encountered ever. The new implementation will simply stop and break the loop quietly, just like the old one.

freddygv: Good catch

freddygv (approved):
This looks good to me, but maybe wait for @huikang to confirm that it resolves the issue before merging

huikang (approved):
The PR fixes the performance issue, reducing the exporting latency to ~200ms. 
However, at certain point, exporting to certain peering cluster will be blocked. But I think this is a different story and I will document elsewhere. That being said, this PR can be merged. Thanks for quick fix, @hashi-derek

PR #14933, opened 2022-10-10
Add -node-name flag to redirect-traffic command

ishustava:
### Description
Add -node-name flag to redirect-traffic command to support running it in environments without client agents.

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/command/connect/redirecttraffic/redirect_traffic.go
+++ b/command/connect/redirecttraffic/redirect_traffic.go
@@ -171,6 +171,9 @@ func (c *cmd) generateConfigFromFlags() (iptables.Config, error) {
 			if len(svcList.Services) < 1 {
 				return iptables.Config{}, fmt.Errorf("proxy service with ID %q not found", c.proxyID)
 			}
+			if len(svcList.Services) > 1 {
+				return iptables.Config{}, fmt.Errorf("expected to find only one proxy service with ID %q, but more were found", c.proxyID)
+			}
 			svc = svcList.Services[0]
 		}
 
--- a/command/connect/redirecttraffic/redirect_traffic.go
+++ b/command/connect/redirecttraffic/redirect_traffic.go
@@ -36,6 +36,7 @@ type cmd struct {
 	client *api.Client
 
 	// Flags.
+	nodeName             string
 	consulDNSIP          string
 	proxyUID             string
 	proxyID              string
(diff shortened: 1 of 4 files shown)


command/connect/redirecttraffic/redirect_traffic.go
```diff
@@ -150,9 +153,25 @@ func (c *cmd) generateConfigFromFlags() (iptables.Config, error) {
 			}
 		}
 
-		svc, _, err := c.client.Agent().Service(c.proxyID, nil)
-		if err != nil {
-			return iptables.Config{}, fmt.Errorf("failed to fetch proxy service from Consul Agent: %s", err)
+		var svc *api.AgentService
+		if c.nodeName == "" {
+			var err error
```
riddhi89, line 158: This might not be needed?

command/connect/redirecttraffic/redirect_traffic.go
```diff
@@ -150,9 +153,25 @@ func (c *cmd) generateConfigFromFlags() (iptables.Config, error) {
...
+				Filter:             fmt.Sprintf("ID == %q", c.proxyID),
+				MergeCentralConfig: true,
+			})
+			if err != nil {
+				return iptables.Config{}, fmt.Errorf("failed to fetch proxy service from Consul: %s", err)
+			}
+			if len(svcList.Services) < 1 {
+				return iptables.Config{}, fmt.Errorf("proxy service with ID %q not found", c.proxyID)
+			}
+			svc = svcList.Services[0]
```
riddhi89, line 174: Nit: We also could check if there is more than 1 service returned (should not happen because of the filter but just in case the filter is broken) - similar to https://github.com/hashicorp/consul/blob/main/command/connect/envoy/envoy.go#L572

riddhi89 (commented):
LGTM. Just some minor stuff.

command/connect/redirecttraffic/redirect_traffic.go
ishustava: I originally had that code there, but then removed because I couldn't write a test for it. There should always be one instance with a given service ID on a node. Let me know though if you think it's worth having it just in case.

command/connect/redirecttraffic/redirect_traffic.go
ishustava: This is needed for the call below.

riddhi89: 🤔 interesting, i thght it would just use https://github.com/hashicorp/consul/blob/main/command/connect/redirecttraffic/redirect_traffic.go#L145, but maybe I am missing something.

command/connect/redirecttraffic/redirect_traffic.go
ishustava: done!

riddhi89: Thanks!

riddhi89 (approved):

command/connect/redirecttraffic/redirect_traffic.go
ishustava: 🤦‍♀️ you're totally right! will remove.

boxofrad (approved):
LGTM! Nice job 👏🏻

PR #14934, opened 2022-10-10
perf: remove expensive reflection from xDS hot path

boxofrad:
### Description
Replaces the reflection-based implementation of proxycfg's `ConfigSnapshot.Clone` with code generated by [deep-copy](https://github.com/globusdigital/deep-copy).

While load testing server-based xDS (for consul-dataplane) we discovered this method is **extremely** expensive.

Screenshot 2022-10-06 at 12 16 42

The `ConfigSnapshot` struct, directly or indirectly, contains a copy of many of the structs in the `agent/structs` package, which creates a large graph for `copystructure.Copy` to traverse at runtime, on every proxy reconfiguration. This is particularly a problem when updating "global" resources such as wildcard intentions or `proxy-defaults` which triggers a thundering herd of updates (and in our tests caused Raft leadership instability).

We test the completeness of `Clone` by using [gofuzz](https://github.com/google/gofuzz) to construct a "full" snapshot, cloning it, and then comparing the clone to the original.

The Envoy integration tests should also catch any regressions.


agent/proxycfg/deep-copy.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+# Uses: https://github.com/globusdigital/deep-copy
+deep-copy -pointer-receiver \
```
rboyer, line 4: Does this correctly nil out the cancelfunc maps?

agent/proxycfg/snapshot_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	snapshot := new(ConfigSnapshot)
+	f.Fuzz(snapshot)
+
+	clone := snapshot.Clone()
+
+	diff := cmp.Diff(snapshot, clone,
+		cmpopts.IgnoreUnexported(indexedTarget{}),
+		cmpopts.IgnoreUnexported(pbpeering.PeeringTrustBundle{}),
+		cmpopts.IgnoreTypes(context.CancelFunc(nil)),
```
rboyer, line 46: The copied snapshots should _not_ have the cancelfunc maps populated IIRC.

proto/pbpeering/peering.deepcopy.go
```diff
@@ -0,0 +1,17 @@
+// generated by deep-copy -pointer-receiver -type PeeringTrustBundle -o ./peering.deepcopy.go ./; DO NOT EDIT.
+
+package pbpeering
+
+// DeepCopy generates a deep copy of *PeeringTrustBundle
```
rboyer, line 5: Shouldn't protobufs use `proto.Clone` instead?

proto/pbpeering/peering.go
```diff
@@ -14,6 +15,7 @@ import (
 	"github.com/hashicorp/consul/agent/structs"
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/consul/lib"
+	"github.com/hashicorp/consul/lib/stringslice"
```
rboyer, line 18: Stray import?

boxofrad: Yep! From my previous attempt at hand-rolling it, will remove.

GNUmakefile
```diff
@@ -321,6 +322,10 @@ lint-tools:
 proto-tools:
 	@$(SHELL) $(CURDIR)/build-support/scripts/devtools.sh -protobuf
 
+.PHONY: codegen-tools
```
rboyer, line 325: I think we need a `make` target to correctly re-execute all of the codegen. Ideally also do something in CI that fails if a fresh run of the deep-copy generators produces a diff (like the `go.mod/go.sum` divergence checks).

boxofrad: Sounds good. Would something like this do the trick?

```
generate-deep-copy:
	go generate ./agent/structs ./proto/pbpeering ./agent/proxycfg
```

That'd technically run any other `go:generate` directives in those packages, so would it be better to run the `deep-copy.sh` scripts directly?

boxofrad: Added a make target and CI check ✅ (went with the latter approach)

rboyer (approved):

PR #14936, opened 2022-10-11
adding configuration option cloud.scada_address

i0rek:
### Description

When adding the `cloud` configuration stanza in https://github.com/hashicorp/consul/pull/14723, we forget to add `scada_address`. This is required when targeting other scada scada servers than the hcp sdk default ones. 

This addition doesn't have documentation because the other cloud configuration don't have that either.

I tested this by building consul and using the configuration option.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

   `agent/config/config.go`.
   `agent/config/runtime.go`.
   translate.
      - [ ] Add that to `DefaultSource` in `agent/config/defaults.go`.
      - [ ] If the config needs to be defaulted for the test server used in unit tests,
            also add it to `DefaultConfig()` in `agent/consul/defaults.go`.
 - [x] **If** your config should take effect on a reload/HUP.
        more of the following places:
        `TestAgent_reloadConfig*`.

### DIFF
--- a/.changelog/14936.txt
+++ b/.changelog/14936.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+agent: Added configuration option cloud.scada_address.
+```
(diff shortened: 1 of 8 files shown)


boxofrad (approved):
LGTM 👍🏻

nickethier (approved):

jjtimmons (approved):

PR #14947, opened 2022-10-11
ui: peer detail view

LevelbossMike:
### Description
This PR adds the peer-detail view for peerings. I'm adding this as a draft because I still have to add tests for this behavior.

##### Imported Services tab
Screenshot 2022-10-11 at 17 08 58

##### Exported Services tab
Screenshot 2022-10-11 at 17 09 05

##### StreamStatus - Bento-Box with smart date handling
Screenshot 2022-10-11 at 17 09 20

This PR depends on https://github.com/hashicorp/consul/pull/14938

* like https://github.com/hashicorp/consul/pull/14938

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


ui/packages/consul-peerings/app/controllers/dc/peers/edit/exported.js
```diff
@@ -0,0 +1,17 @@
+import Controller from "@ember/controller";
+import { tracked } from "@glimmer/tracking";
+import { action } from "@ember/object";
+
+export default class PeersEditExportedController extends Controller {
+  queryParams = {
```
WenInCode, line 6: I remember John saying that the app didn't use query params as prescribed by Ember (declaring them in the controller like this) I don't really remember the reasoning. Happy to see this approach, but figured it would be prudent to ask if there could be any issues with how it's working elsewhere. I imagine not an issue.

ui/packages/consul-peerings/app/templates/dc/peers/edit/addresses.hbs
```diff
@@ -1,7 +1,40 @@
...
+        </li>
+        <li class="learn-link">
+          <a
+            href="{{env
+              'CONSUL_DOCS_URL'
+            }}/connect/cluster-peering/create-manage-peering"
+            rel="noopener noreferrer"
+            target="_blank"
+          >
+            Take the tutorial
```
WenInCode, line 34: It would be nice to use the translations file for this copy

WenInCode (commented):
Looking good so far, liking the use of providers.

ui/packages/consul-peerings/app/controllers/dc/peers/edit/exported.js
LevelbossMike: It's not an issue to declare the query parameter in the controller like you would do in a regular app. Actually, I had to declare it like this because without the query parameter setup on the controller, the URL wouldn't update when changing `search`. So, why does this work for `peers.edit.imported` and not `peers.edit.exported`?  I dug a little bit to find out why query parameter work with the custom DSL but weren't working here when declaring the controller file:

It turns out that declaring the query parameters on the route is good enough when you use the `mut`-helper to create a two-way data-binding for the search property. We do in the regular `SearchBar`-setup, [that is used across the rest of the application](https://github.com/hashicorp/consul/pull/14947/files#diff-f19b6467ddb33d4df2c09ab9007d19a886adcd842e1e24d6811a908b2c009fb7R62), but we aren't doing that in the exported-template as I wanted to stick to data-down-actions-up as much as possible for new code.

__TLDR;__: We actually need to declare the query-param in the controller, when we want to use a tracked property that should update a query-parameter in the URL.

ui/packages/consul-peerings/vendor/consul-peerings/routes.js
```diff
@@ -1,30 +1,76 @@
...
+                as: "filter",
+                replace: true,
+              },
             },
-            search: {
-              as: 'filter',
-              replace: true,
+          },
+        },
+        edit: {
```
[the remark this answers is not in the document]
LevelbossMike: I pulled back in the setup that John created originally, that's why this is called `edit`. I'll rename to show.

ui/packages/consul-ui/app/components/providers/dimension/index.js
```diff
@@ -0,0 +1,53 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+import { ref } from 'ember-ref-bucket';
+import { htmlSafe } from '@ember/template';
+
+export default class DimensionsProvider extends Component {
```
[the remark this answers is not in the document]
LevelbossMike: Done

ui/packages/consul-peerings/app/templates/dc/peers/edit/addresses.hbs
LevelbossMike: Done

jnwright:
Copy on exported tab tooltip is incorrect. Should read "Services exported **to**.." RIght now it says from.

PR #14956, opened 2022-10-11
cache: refactor agent cache fetching to prevent unnecessary fetches on error

rboyer:
### Description

This continues the work done in https://github.com/hashicorp/consul/pull/14908 where a crude solution to prevent a goroutine leak was implemented. The former code would launch a perpetual goroutine family every iteration (`+1 +1`) and the fixed code simply caused a new goroutine family to first cancel the prior one to prevent the leak (`-1 +1 == 0`).

This PR refactors this code completely to:
- make it more understandable
- remove the recursion-via-goroutine strangeness
- prevent unnecessary RPC fetches when the prior one has errored.

The core issue arose from a conflation of the `entry.Fetching` field to mean:
1. there is an RPC (blocking query) in flight _right now_
2. there is a goroutine running to manage the RPC fetch retry loop

The problem is that the goroutine-leak-avoidance check would treat `Fetching` like (2), but within the body of a goroutine it would flip that boolean back to `false` before the retry sleep. This would cause a new chain of goroutines to launch which #14908 would correct crudely.

The refactored code uses a plain for-loop and changes the semantics to track state for "is there a goroutine associated with this cache entry" instead of the former.

We use a `uint64` unique identity per goroutine instead of a boolean so that any orphaned goroutines can tell when they've been replaced when the expiry loop deletes a cache entry while the goroutine is still running and is later replaced.


rboyer:
For this I slowly duplicated and pushed around the existing code starting from the two places `fetch` was called until I could get the recursive goroutine launch to be invoking the identical method that was calling it. At that point the goroutine launch could be easily swapped for a `for` loop and then I cleaned up the rest.

I left one specific section in an extra anonymous block to help during the PR review since the bulk of it did not change.

agent/cache/cache.go
```diff
@@ -555,46 +555,41 @@ func makeEntryKey(t, dc, peerName, token, key string) string {
...
-// if it doesn't exist. If this is false, then fetch will do nothing
-// if the entry doesn't exist. This latter case is to support refreshing.
-func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ignoreExisting bool) <-chan struct{} {
-	// We acquire a write lock because we may have to set Fetching to true.
+// fetch triggers a new background fetch for the given Request. If a background
+// fetch is already running or a goroutine to manage that still exists for a
+// matching Request, the waiter channel for that request is returned. The
+// effect of this is that there is only ever one blocking query and goroutine
+// for any matching requests.
+func (c *Cache) fetch(key string, r getOptions) <-chan struct{} {
```
rboyer, line 563: This now only has a single entry point (`getWithIndex`) and is not called pseudo-recursively.

agent/cache/cache.go
```diff
@@ -555,46 +555,41 @@ func makeEntryKey(t, dc, peerName, token, key string) string {
...
-	}
+	case ok && entry.Fetching:
+		// If we already have an entry and it is actively fetching, then return
+		// the currently active waiter.
+		return entry.Waiter
 
-	// If we already have an entry and it is actively fetching, then return
-	// the currently active waiter.
-	if ok && entry.Fetching {
+	case ok && entry.HasGoroutine:
```
rboyer, line 584: New conditional branch to handle the "not fetching, but there's already a goroutine" scenario underlying the bug.

agent/cache/cache.go
```diff
@@ -609,13 +604,93 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
...
 	entry.Fetching = true
+	entry.HasGoroutine = true
 	c.entries[key] = entry
 	metrics.SetGauge([]string{"consul", "cache", "entries_count"}, float32(len(c.entries)))
 	metrics.SetGauge([]string{"cache", "entries_count"}, float32(len(c.entries)))
 
-	tEntry := r.TypeEntry
 	// The actual Fetch must be performed in a goroutine.
-	go func() {
+	go c.launchBackgroundFetcher(key, r, entry)
```
rboyer, line 613: We mark that the entry `HasGoroutine=true` and then launch the looping goroutine. It has a defer to flip that back to `false` so it should allow for a subsequent call to re-launch the goroutine should it become needed again.

agent/cache/cache.go
```diff
@@ -609,13 +604,91 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
...
+		// waiting to refresh we may have expired and got evicted. If that
+		// happened, we don't want to create a new entry.
+		r.Info.MustRevalidate = false
+		r.Info.MinIndex = 0
+
+		// We acquire a write lock because we may have to set Fetching to true.
+		c.entriesLock.Lock()
+
+		var ok bool
+		ok, _, entry = c.getEntryLocked(r.TypeEntry, key, r.Info)
```
rboyer, line 663: Here we have the latest copy of the `entry` assigned (not _shadowed_) so that the next loop will use the updated value.

agent/cache/cache.go
```diff
@@ -609,13 +604,91 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
...
+		if !ok || entry.Fetching {
```
rboyer, line 664: Locally I just thought of changing this line to `if !ok || !entry.HasGoroutine {` since the latter half of this conditional post-refactor seemed like dead code.

That seems to have triggered at least one test to occasionally double-close the waiter channel and panic.

agent/cache/cache.go
```diff
@@ -609,13 +604,91 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
...
+			// If we don't have an existing entry, return immediately.
+			//
+			// Also if we already have an entry and it is actively fetching, then
+			// return immediately.
+			c.entriesLock.Unlock()
```
dhiaayachi, line 669: I think it will be more readable and safer if we did the inverse logic on the if (`if ok && !entry.Fetching`) and defer the unlock.

agent/cache/cache.go
```diff
@@ -609,13 +604,91 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
...
+			return
+		}
+
+		// Set that we're fetching to true, which makes it so that future
+		// identical calls to fetch will return the same waiter rather than
+		// perform multiple fetches.
+		entry.Fetching = true
```
dhiaayachi, line 676: I don't quite understand why we set Fetching to true here while we just set it to true before calling `launchBackgroundFetcher`?

agent/cache/cache.go
```diff
@@ -609,13 +604,91 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
...
+
+		// If we're over the attempt minimum, start an exponential backoff.
+		if wait := backOffWait(attempt); wait > 0 {
+			totalWait += wait
+		}
+
+		// If we have a timer, wait for it
+		if r.TypeEntry.Opts.RefreshTimer > 0 {
+			totalWait += r.TypeEntry.Opts.RefreshTimer
+		}
```
dhiaayachi, line 645: By my comment on the former PR, those 2 could be simplified by removing the if statements.
```suggestion

		// If we're over the attempt minimum, start an exponential backoff.
		totalWait := backOffWait(attempt)

		// If we have a timer, wait for it
		totalWait += r.TypeEntry.Opts.RefreshTimer
		
```

dhiaayachi (commented):
I had few questions and some suggestions to simplify the code. This is getting close, thank you @rboyer for going through this complex code and make it simpler 👏

agent/cache/cache.go
```diff
@@ -609,13 +604,91 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
...
+			entry.Fetching = false
+		}
+	})
+
+	var (
+		attempt uint
+		entry   = initialEntry
+	)
+	for {
+		if !c.runBackgroundFetcherOnce(key, r, entry, &attempt) {
```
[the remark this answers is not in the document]
rboyer: I like that. I mostly was trying to avoid this feeling like the syntax of `append()` where you pass the prior and return the updated and the caller has to do the actual update.

agent/cache/cache.go
rboyer: Can't defer the unlock because we acquired the lock inside the body of the for loop.

agent/cache/entry.go
```diff
@@ -26,9 +26,10 @@ type cacheEntry struct {
 	Index uint64
 
 	// Metadata that is used for internal accounting
-	Valid    bool          // True if the Value is set
-	Fetching bool          // True if a fetch is already active
-	Waiter   chan struct{} // Closed when this entry is invalidated
+	Valid        bool          // True if the Value is set
+	Fetching     bool          // True if a fetch RPC is already active
+	HasGoroutine bool          // True if a fetch goroutine is running.
```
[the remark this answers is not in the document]
rboyer: That's actually the bug being fixed here.

The code pre-PR slightly conflated two scenarios:

1. `fetching` means there is a blocking query that has not returned yet
2. `fetching` means that there is a goroutine that _may_ be executing an RPC right now

Within the body of that goroutine, when the RPC returns it updated the fetching value to false which _breaks_ code that assumes (2). But we need to do that so other code can use the existence of a blocked RPC to imply:
```
			// There is an active blocking query for this data, which has not
			// returned. We can logically deduce that the contents of the cache
			// are actually current, and we can simply return this while
			// leaving the blocking query alone.
			return true, true, entry
```

So we sorta need to know (1) for the above cache freshness inference and (2) to avoid the goroutine leak.

rboyer:
> Only reason not to is to think about release and testing plan. I think it would be better to not try an squeeze this so soon before 1.14 GA. If we already have release branches cut for 1.14 I'd be OK merging to main now, but if not we might want to hold off and merge after the release and think a bit about any additional testing we might want to do before the next release?
@banks I was definitely going to ensure this only landed AFTER we cut the real 1.14.x branch.

agent/cache/entry.go
```diff
@@ -26,9 +26,10 @@ type cacheEntry struct {
 	Index uint64
 
 	// Metadata that is used for internal accounting
-	Valid    bool          // True if the Value is set
-	Fetching bool          // True if a fetch is already active
-	Waiter   chan struct{} // Closed when this entry is invalidated
+	Valid       bool          // True if the Value is set
+	Fetching    bool          // True if a fetch RPC is currently active
+	GoroutineID uint64        // Nonzero if a fetch goroutine is running.
```
rboyer, line 31: Switched this from a `bool` to a `uint64` so that goroutines can fingerprint their own entry.

This is important in this scenario:

1. bg fetch launches for KEY
2. time passes
3. expiry loop clears entries map for KEY
4. new request re-inserts entry for KEY
5. **new** bg fetch launches for KEY
6. original bg fetch doesn't bail because KEY exists!

Then you end up with two goroutines fetching the same KEY.

This occasionally failed `TestCacheGet_expireBackgroudRefresh`

agent/cache/cache.go
```diff
@@ -613,9 +613,97 @@ func (c *Cache) fetch(key string, r getOptions, allowNew bool, attempt uint, ign
...
+		entry.Fetching = true
+		c.entries[key] = entry
+		metrics.SetGauge([]string{"consul", "cache", "entries_count"}, float32(len(c.entries)))
+		metrics.SetGauge([]string{"cache", "entries_count"}, float32(len(c.entries)))
+		c.entriesLock.Unlock()
+	}
+}
+
+func (c *Cache) runBackgroundFetcherOnce(goroutineID uint64, key string, r getOptions) (shouldStop, shouldBackoff bool) {
+	// Freshly re-read this, rather than relying upon the caller to fetch it
```
rboyer, line 691: I also took the opportunity to clean this up and just fetch the data every time, instead of fetching it before calling this function, which had some error prone logic.

agent/cache/entry.go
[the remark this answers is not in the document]
dhiaayachi: Thank you for explaining this.

I see it now. The nuance is between: I have a blocking query for it (which mean what I have is up to date) and I'm trying to update the cache but nothing is in there for the moment or I have some error.

agent/cache/cache.go
dhiaayachi: 🤦 I didn't see the loop

dhiaayachi (approved):
Great work @rboyer !!

rboyer:
After some discussion with @dhiaayachi we realized that the `entry.Fetching` boolean was unnecessary now and removed that complicating factor.

rboyer:
It's worth pointing out that there is now a subtle change in behavior (for the better?) in the fixed code as it pertains to refreshing cache types:

In the current AND fixed code, successful RPCs are cached and each request to the cache does not cause an immediate RPC. This is **intended and desirable**.

In the current code only, an erroring RPC:
  * will return quickly and return the error to the caller.
  * will set `entry.Fetching = false`, which is the condition that keeps another goroutine from running.
  * and it will sleep for the backoff interval.
 
This means that a subsequent request through the agent cache for the same key DURING the sleep interval will cause a new goroutine to immediately request a new RPC which may also error. An endpoint that is erroring a lot will force all calls to basically bypass the cache and send load straight to the servers.

Contrast that with the fixes in the PR: an error RPC
* will return quickly and return the error to the caller.
* will prevent another goroutine from launching.
* and it will sleep for the backoff interval.

This means that for the fixed version of the code, **errors behave more like they are cached** (which I believe was also the original intent) in that they don't cause a 1:1 relationship between requests to the cache and requests to the server.

One unfortunate thing is that this breaks/flakes a variety of tests that I have yet to correct.

banks (approved):
Looks great. Thanks RB for chasing this down and explaining the workings for us all!

agent/cache-types/connect_ca_leaf.go
```diff
@@ -165,6 +165,7 @@ func (c *ConnectCALeaf) fetchDone(rootUpdateCh chan struct{}) {
 	if len(c.rootWatchSubscribers) == 0 && c.rootWatchCancel != nil {
 		// This was the last request. Stop the root watcher.
 		c.rootWatchCancel()
+		c.rootWatchCancel = nil
```
rboyer, line 168: NOTE: I'll move this to its own bug since I only fixed it here because fixing this PR uncovered this "2 bugs cancel out" situation.

agent/setup.go
```diff
@@ -53,7 +53,7 @@ type BaseDeps struct {
 
 type ConfigLoader func(source config.Source) (config.LoadResult, error)
 
-func NewBaseDeps(configLoader ConfigLoader, logOut io.Writer) (BaseDeps, error) {
+func NewBaseDeps(configLoader ConfigLoader, logOut io.Writer, providedLogger hclog.InterceptLogger) (BaseDeps, error) {
```
rboyer, line 56: TODO: move this to a different PR to help with test logger threading.

rboyer:
That latest round of patches (which will be done in separate precursor PRs) have fixed the remaining test failures.

rboyer:
Trivial test related PR extraction: https://github.com/hashicorp/consul/pull/14996

rboyer:
Resolving conflict on main by reverting #14908 and rebasing. In effect canceling out the "band-aid" in favor of the refactor.
Also rebased on a cherry pick from #14996 so the best way to view this in isolation is to ignore the first two commits.

agent/setup.go
rboyer: https://github.com/hashicorp/consul/pull/14996

boxofrad (approved):
This is so much easier to reason about. Great job 👏🏻

PR #14958, opened 2022-10-12
Add basic nonce management

freddygv:
### Description
This commit adds a monotonically increasing nonce to include in peering replication response messages. Every ack/nack from the peer handling a response will include this nonce, allowing to correlate the ack/nack with a specific resource.

At the moment nothing is done with the nonce when it is received. In the future we may want to add functionality such as retries on NACKs, depending on the class of error.

* Various unit tests from both the request/response sides

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/grpc-external/services/peerstream/stream_test.go
+++ b/agent/grpc-external/services/peerstream/stream_test.go
@@ -1887,6 +1887,7 @@ func expectReplEvents(t *testing.T, client *MockClient, checkFns ...func(t *test
 		// Ensure every nonce was unique.
 		if resp := out[i].GetResponse(); resp != nil {
 			require.NotContains(t, nonces, resp.Nonce)
+			nonces[resp.Nonce] = struct{}{}
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


kisunji (commented):
Thinking out loud: a change of leader will reset the nonce and the client (in the future) will also need to reset the nonce it's tracking, right? In other words, the nonce management is limited to the peer stream connection, not the higher level peering relationship

agent/grpc-external/services/peerstream/stream_test.go
```diff
@@ -1809,8 +1880,14 @@ func expectReplEvents(t *testing.T, client *MockClient, checkFns ...func(t *test
 		}
 	})
 
+	nonces := make(map[string]struct{})
 	for i := 0; i < num; i++ {
 		checkFns[i](t, out[i])
+
+		// Ensure every nonce was unique.
+		if resp := out[i].GetResponse(); resp != nil {
+			require.NotContains(t, nonces, resp.Nonce)
```
hashi-derek, line 1889: Are we ever adding to the nonces map?

kisunji, line 1889: I think you're missing a line to populate `nonces`

freddygv: Fixed

freddygv: 😳

hashi-derek (approved):

PR #14960, opened 2022-10-12
proxycfg: rate-limit delivery of config snapshots

boxofrad:
### Description
Adds a user-configurable rate limiter to `proxycfg` snapshot delivery, with a default limit of 250 updates per second.

This addresses a problem observed in our load testing of Consul Dataplane where updating a "global" resource such as a wildcard intention or the `proxy-defaults` config entry could starve the Raft or memberlist goroutines of CPU time, causing general cluster instability.

Tested using the Consul Dataplane load test harness. Observed that creating a wildcard intention with 5k connected streams no longer triggers missed memberlist pings (and suspicions from other peers).

### DIFF
--- a/agent/proxycfg/state.go
+++ b/agent/proxycfg/state.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-hclog"
+	"golang.org/x/time/rate"
 
 	cachetype "github.com/hashicorp/consul/agent/cache-types"
 	"github.com/hashicorp/consul/agent/structs"
@@ -79,6 +80,8 @@ type state struct {
 	ch     chan UpdateEvent
 	snapCh chan ConfigSnapshot
 	reqCh  chan chan *ConfigSnapshot
+
+	rateLimiter *rate.Limiter
 }
 
 // failed returns whether run exited because a data source is in an
@@ -148,7 +151,7 @@ func copyProxyConfig(ns *structs.NodeService) (structs.ConnectProxyConfig, error
 //
 // The returned state needs its required dependencies to be set before Watch
 // can be called.
-func newState(id ProxyID, ns *structs.NodeService, source ProxySource, token string, config stateConfig) (*state, error) {
+func newState(id ProxyID, ns *structs.NodeService, source ProxySource, token string, config stateConfig, rateLimiter *rate.Limiter) (*state, error) {
 	// 10 is fairly arbitrary here but allow for the 3 mandatory and a
 	// reasonable number of upstream watches to all deliver their initial
 	// messages in parallel without blocking the cache.Notify loops. It's not a
@@ -176,6 +179,7 @@ func newState(id ProxyID, ns *structs.NodeService, source ProxySource, token str
 		ch:              ch,
 		snapCh:          make(chan ConfigSnapshot, 1),
 		reqCh:           make(chan chan *ConfigSnapshot, 1),
+		rateLimiter:     rateLimiter,
 	}, nil
 }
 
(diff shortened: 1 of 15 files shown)


website/content/docs/agent/config/config-files.mdx
```diff
@@ -2206,3 +2206,23 @@ tls {
...
+    resources at the expense of other work, such as Raft and Gossip, which
+    could cause general cluster instability.
+
+    It defaults to 250 updates per second, which is conservative and was chosen
+    based on a load-test of 5k streams connected to a single server with 2 CPU
+    cores.
+
+    This can be reloaded using `consul reload` or sending the server a `SIGHUP`
+    to allow lowering or increasing the limit without a rolling restart in
+    emergencies.
```
trujillo-adam, line 2228:
```suggestion
- `xds`: This object allows you to configure the behavior of Consul's
[xDS protocol](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol)
server. 

  - `update_max_per_second`: Specifies the number of proxy configuration updates across all connected xDS streams that are allowed per second. This configuration prevents updates to global resources, such as wildcard intentions, from consuming system resources at the expense of other processes, such as Raft and Gossip, which could cause general cluster instability.

    The default value is `250`. It is based on a load test of 5,000 streams connected to a single server with two CPU cores. 

    If necessary, you can lower or increase the limit without a rolling restart by using the `consul reload` command or by sending the server a `SIGHUP`. 
```
Just a few suggestions for consistency with our style guide. I removed the subjective statement about the default being conservative. If we really want to communicate that there is room for increasing the value, we can say something about how the default is below the threshold at which system instability began to occur during our tests. Also, I removed the reference to the version number because this update should ONLY appear in 1.14 documentation and later, which would make the sentence unnecessary.

trujillo-adam (commented):
Looks good. I made a few suggestions related to style guide consistency.

agent/proxycfg/state.go
```diff
@@ -311,6 +315,17 @@ func (s *state) run(ctx context.Context, snap *ConfigSnapshot) {
...
 	var coalesceTimer *time.Timer
 
+	scheduleUpdate := func() {
+		r := s.rateLimiter.Reserve()
+
+		coalesceTimer = time.AfterFunc(coalesceTimeout+r.Delay(), func() {
+			// This runs in another goroutine so we can't just do the send
+			// directly here as access to snap is racy. Instead, signal the main
+			// loop above.
+			sendCh <- struct{}{}
```
boxofrad, line 325: ⚠️ There's a (pre-existing) goroutine leak here, where if `ctx` is canceled when there's a pending update this goroutine will never halt.

The fix is simple, but for posterity, I'll add it in a follow-up PR once this one is merged.

website/content/docs/agent/config/config-files.mdx
boxofrad: Awesome, thanks 🙌🏻

agent/proxycfg/state.go
```diff
@@ -303,6 +307,17 @@ func (s *state) run(ctx context.Context, snap *ConfigSnapshot) {
 	sendCh := make(chan struct{})
 	var coalesceTimer *time.Timer
 
+	scheduleUpdate := func() {
+		r := s.rateLimiter.Reserve()
+
+		coalesceTimer = time.AfterFunc(coalesceTimeout+r.Delay(), func() {
```
mkeeler, line 313: Should this instead wait for `max(coalesceTimeout, r.Delay())`?

coalesce timeout will ensure this states snapshot doesn't get emitted too frequently
r.Delay ensures that we don't emit snapshots too frequently globally.

This might be sort of a premature optimization but figured I would note it anyways.

mkeeler (approved):

agent/proxycfg/state.go
boxofrad: Good point! I think either is fine honestly, but I guess using the `max` could save 200ms of unnecessary waiting - will change it 🙇🏻‍♂️

PR #14962, opened 2022-10-12
xds: properly merge central config for "agentless" services

boxofrad:
### Description
Fixes an omission in the 1.14 beta, where central service configuration (i.e. the `proxy-defaults` and `service-defaults` config entries) was not applied to the generated xDS resources.

1. Start a dev server
1. Register a node, service, and sidecar proxy via the catalog API
1. Run `consul-dataplane` with the proxy service's ID
1. Create a `proxy-defaults` config entry which sets the default protocol to HTTP
1. Observe in the Envoy admin server's `config_dump` page that the public listener has an `HttpConnectionManager` filter
1. Create a `service-defaults` entry which overrides the protcol to TCP
1. Observe that the `HttpConnectionManager` filter is gone

### DIFF
--- a/agent/proxycfg-sources/catalog/config_source.go
+++ b/agent/proxycfg-sources/catalog/config_source.go
@@ -9,6 +9,7 @@ import (
 	"github.com/hashicorp/go-memdb"
 
 	"github.com/hashicorp/consul/acl"
+	"github.com/hashicorp/consul/agent/configentry"
 	"github.com/hashicorp/consul/agent/local"
 	"github.com/hashicorp/consul/agent/proxycfg"
 	"github.com/hashicorp/consul/agent/structs"
@@ -117,7 +118,6 @@ func (m *ConfigSource) startSync(closeCh <-chan chan struct{}, proxyID proxycfg.
 		ws.Add(store.AbandonCh())
 
 		_, ns, err := store.NodeService(ws, proxyID.NodeName, proxyID.ID, &proxyID.EnterpriseMeta, structs.DefaultPeerKeyword)
-
 		switch {
 		case err != nil:
 			logger.Error("failed to read service from state store", "error", err.Error())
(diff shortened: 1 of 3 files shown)


agent/proxycfg-sources/catalog/config_source.go
```diff
@@ -130,14 +130,29 @@ func (m *ConfigSource) startSync(closeCh <-chan chan struct{}, proxyID proxycfg.
...
-			}
-			return ws, nil
 		}
+
+		_, ns, err = configentry.MergeNodeServiceWithCentralConfig(
+			ws,
+			store,
+			// TODO(agentless): it doesn't seem like we actually *need* any of the
+			// values on this request struct - we should try to remove the parameter
+			// in case that changes in the future as this call-site isn't passing them.
```
boxofrad, line 140: @pglass Have I missed anything here? When I dug down into the methods it doesn't look like we use any of the fields on this struct (the `Datacenter` and `Source` aren't used for anything).

agent/proxycfg-sources/catalog/config_source.go
```diff
@@ -130,14 +130,29 @@ func (m *ConfigSource) startSync(closeCh <-chan chan struct{}, proxyID proxycfg.
...
-		default:
-			err := m.Manager.Register(proxyID, ns, source, proxyID.Token, false)
-			if err != nil {
-				logger.Error("failed to register service", "error", err.Error())
-				return nil, err
-			}
-			return ws, nil
 		}
+
+		_, ns, err = configentry.MergeNodeServiceWithCentralConfig(
```
boxofrad, line 135: I'll re-run the load tests against this, because I'm not sure what impact re-registering every service when updating `proxy-defaults` will have (particularly as it'll restart watches etc.) ⌛

pglass (approved):

Issue #14732, opened 2022-09-23
rpc_hold_timeout has been reused for client RPC call timeout


dclaisse:
#### Overview of the Issue

Since versions 1.11.9, 1.12.1 and 1.13.0, there is now a timeout for client RPC calls (PR #11500).
This PR reused [rpc_hold_timeout](https://www.consul.io/docs/agent/config/config-files#rpc_hold_timeout) that defaults to 7s.

This timeout can be too low in some cases, such as fetching list of nodes from a big (~10k nodes) remote datacenter with high latency (350 ms in my case).

Issue is that while I'd like to increase client RPC timeout, I shouldn't have to mess with `rpc_hold_timeout` which has not the same role at all as per documentation (timeout for internal RPC calls during leader (re)election).

#### Reproduction Steps

1. Create a "faraway" cluster with 10k client nodes and high latency to reach it, and federate it to a local cluster
1. Run on a local consul agent `curl http://localhost:8500/v1/catalog/nodes?dc=faraway`
1. If request takes more than 7s, you'll receive an error: `"rpc error making call: i/o deadline reached"`

### Consul info for both Client and Server

Consul version 1.12.4 on both sides.

### Operating system and Environment details

Not relevant for this issue.

### Log Fragments

On local agent:
```
[ERROR] agent.client: RPC failed to server: method=Catalog.ListNodes server=x.x.x.x:8300 error="rpc error making call: i/o deadline reached"
[ERROR] agent.http: Request error: method=GET url=/v1/catalog/nodes?dc=faraway from=y.y.y.y:35670 error="rpc error making call: i/o deadline reached"
```

dnlopes:
I just hit this issue. My consul-replicate daemon was hitting the timeout and I had to increase the hold-timeout as a workaround. It's OK as a workaround, but I wonder if this will have other impacts that I can't foresee.

kisunji:
The fix for this issue has been merged to main and will be available in the next releases for versions:
```
1.14.0
1.13.3
1.12.6
1.11.11
```

it exposes a new config called `limits.rpc_client_timeout` which defaults to 60s and can be tuned separately from `rpc_hold_timeout`.

PR #14965, opened 2022-10-12
Refactor client RPC timeouts

kisunji:
### Description
This PR has two objectives:

1. Fix an issue identified in https://github.com/hashicorp/consul/issues/14732 where `rpc_hold_timeout` was being used as the timeout for non-blocking queries. Users should be able to tune read timeouts without fiddling with `rpc_hold_timeout`. A new configuration `rpc_read_timeout` was created.
2. Refactor some implementation from the original PR https://github.com/hashicorp/consul/pull/11500 to remove the misleading linkage between `RPCInfo`'s timeout (used to retry in case of certain modes of failures) and the client RPC timeouts.

* Added test cases
* Manually tested with a client-server setup where I ran the following queries against the client concurrently and confirmed deadlines were respected:
  * Long non-blocking query (KVS.Get modified to sleep on a certain key)
  * Short non-blocking query (KVS.Get)
  * Blocking query (KVS.Get)

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/pool/pool.go
```diff
@@ -364,7 +364,7 @@ func (p *ConnPool) dial(
 	tlsRPCType RPCType,
 ) (net.Conn, HalfCloser, error) {
 	// Try to dial the conn
-	d := &net.Dialer{LocalAddr: p.SrcAddr, Timeout: p.Timeout}
+	d := &net.Dialer{LocalAddr: p.SrcAddr, Timeout: DefaultDialTimeout}
```
kisunji, line 367: This was changed in https://github.com/hashicorp/consul/pull/11500 but I reverted it back to `DefaultDialTimeout` to be consistent with other usages.

agent/pool/pool.go
```diff
@@ -56,36 +57,6 @@ type Conn struct {
...
-	}
-	var deadline time.Time
-	if timeout > 0 {
-		deadline = time.Now().Add(timeout)
-	}
-	if err := c.Conn.SetReadDeadline(deadline); err != nil {
-		return 0, err
-	}
-	return c.Conn.Read(b)
-}
```
kisunji, line 83: This type was introduced in https://github.com/hashicorp/consul/pull/11500 but while debugging timeout issues I found this difficult to understand.

Now, instead of a wrapped `net.Conn` which sets a deadline per `Read` call, we simply set the deadline during the `rpc` call

agent/pool/pool.go
```diff
@@ -620,6 +590,17 @@ func (p *ConnPool) rpcInsecure(dc string, addr net.Addr, method string, args int
 	return nil
 }
 
+// BlockableQuery represents a read query which can be blocking or non-blocking.
+// This interface is used to set an appropriate read timeout for rpc connections.
+type BlockableQuery interface {
+	// BlockingTimeout returns duration > 0 if the query is blocking.
+	// Otherwise returns 0 for non-blocking queries.
+	BlockingTimeout(maxQueryTime, defaultQueryTime time.Duration) time.Duration
```
kisunji, line 598: I found that narrowly defining this interface was better than giving `RPCInfo` a `Timeout` method and forcing every type to return a timeout (in most cases is rpc_hold_timeout)

agent/consul/client_test.go
```diff
@@ -896,33 +895,53 @@ func TestClient_RPC_Timeout(t *testing.T) {
...
+	})
+
+	t.Run("check that deadline does not persist across calls", func(t *testing.T) {
+		var out struct{}
+		err := c1.RPC("Long.Wait", &structs.NodeSpecificRequest{}, &out)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "rpc error making call: i/o deadline reached")
+		// We use structs.KVSRequest, which does not implement pool.BlockableQuery
+		// and should have no timeouts defined.
+		require.NoError(t, c1.RPC("Long.Wait", &structs.KVSRequest{}, &out))
```
kisunji, line 922: I wanted to sanity check that the setting the deadline on the stream connection did not persist when the client was cached in the connection pool.

huikang (commented):
Do we need to add the new field to api?

kisunji:
> Do we need to add the new field to api?

No, the new fields are config and don't need to be in api

agent/pool/pool.go
```diff
@@ -31,7 +32,7 @@ type muxSession interface {
 
 // streamClient is used to wrap a stream with an RPC client
 type StreamClient struct {
-	stream *TimeoutConn
+	stream net.Conn
```
kisunji, line 35: This is simply reverting a change from https://github.com/hashicorp/consul/pull/11500

agent/config/default.go
```diff
@@ -98,6 +98,7 @@ func DefaultSource() Source {
 			http_max_conns_per_client = 200
 			https_handshake_timeout = "5s"
 			rpc_handshake_timeout = "5s"
+			rpc_read_timeout = "10m"
```
hashi-derek, line 101: 10 minutes is really long. Is there a reason why we chose this?

agent/structs/structs.go
```diff
@@ -355,13 +355,13 @@ func (q QueryOptions) Timeout(rpcHoldTimeout, maxQueryTime, defaultQueryTime tim
...
-		return q.MaxQueryTime + rpcHoldTimeout
+		return q.MaxQueryTime
 	}
-	return rpcHoldTimeout
+	return 0
 }
 
 func (q QueryOptions) HasTimedOut(start time.Time, rpcHoldTimeout, maxQueryTime, defaultQueryTime time.Duration) (bool, error) {
-	return time.Since(start) > q.Timeout(rpcHoldTimeout, maxQueryTime, defaultQueryTime), nil
+	return time.Since(start) > rpcHoldTimeout+q.BlockingTimeout(maxQueryTime, defaultQueryTime), nil
```
hashi-derek, line 364: Perhaps a comment explaining the reason for adding them together will help us in the future?

agent/pool/pool.go
```diff
@@ -629,9 +610,20 @@ func (p *ConnPool) rpc(dc string, nodeName string, addr net.Addr, method string,
...
+		timeout := info.BlockingTimeout(p.MaxQueryTime, p.DefaultQueryTime)
+		if timeout <= 0 {
+			// must be non-blocking
+			timeout = p.ReadTimeout
+		}
+		if timeout > 0 {
+			deadline = time.Now().Add(timeout)
+		}
+	}
+	if err := sc.stream.SetReadDeadline(deadline); err != nil {
```
hashi-derek, line 625: It looks like `deadline` isn't initialized if we cannot cast to a BlockableQuery. Will that cause issues for this `SetReadDeadline(deadline)` that will receive a timestamp from the past?

kisunji: A zero value for `SetReadDeadline` will set no deadline, which is intended for writes and I think an acceptable fallback for any other types of queries that don't implement `BlockableQuery`

agent/config/default.go
kisunji: I do agree it is long but wanted to set a high upper bound so this config largely remains out of people's concerns. The only data point I have is from a customer who reported `25s` query times. Maybe `1m` is more pragmatic?

hashi-derek: Yeah, I'm not sure. If a customer was reaching a 25s query, then I suppose we should be careful and try to keep it at a larger number. So 10m is probably appropriately long to not affect anyone.

Everything is being served from in-memory, so I'm not sure if it really should ever be taking that long though. It would be nice if we had good stats around this -- we could set what we believe to be a sane default, rather than the upper bound.

hashi-derek (approved):

agent/config/default.go
banks: Good thoughts here folks. I think 60s seems like a conservative enough default - if your RPCs are taking more than that, the server is in a pretty awful shape and clients should probably be timing out already.

We could consider documenting in our production guide that this should be reduced ideally to just a few seconds.

It would be ideal if this config is reloadable too so that it's easy if it is set too short and causing amplification due to bad retry behaviour or something that the operator can get out of jail without rolling all servers! You probably did that already @kisunji -- i've not looked yet but will leave the note here to come back to!

agent/consul/client_test.go
```diff
@@ -896,33 +895,53 @@ func TestClient_RPC_Timeout(t *testing.T) {
...
+	})
+
+	t.Run("blocking query succeeds", func(t *testing.T) {
+		// Blocking requests have a longer timeout (100ms) so this should pass since we
+		// add the maximum jitter which should be 16ms
```
banks, line 927: If these tests depend on timing differences of 16ms then they'll probably end up being pretty flakey in CI right? Or did I misunderstand the comment?

trujillo-adam (approved):

agent/pool/pool.go
banks: I'm having a hard time working out whether this is the same behaviour as before - did `RPCInfo` only get implemented for Reads? If not then it seems like this new code is now never setting a timeout on write requests where it was before.

Separate issue: assuming it's the same, is there a reason we allow writes to take arbitrarily long? I could understand having a separate timeout for them because they might have to wait for Raft etc. but it seems like it would be ideal if we didn't have zero timeout on writes...

That said, (assuming the behaviour is same as before) it's better to do that in another PR.

banks (approved):
I think this looks good assuming the behavior for writes is equivalent to before.

I wonder if we should have a separate timeout for writes, but I think we should investigate that separately as the considerations (Raft etc.) are different.

agent/config/default.go
kisunji: I haven't made it reloadable but it would definitely be a nice UX for operators! Will add a commit

agent/pool/pool.go
kisunji: Hmm I took a look at the original PR and it seems to apply `rpc_hold_timeout` to all RPCs (I conflated `net.Conn.Read` with "read" requests). 

I think the correct solution would be to rename `rpc_read_timeout` to `rpc_client_timeout` and have it apply the same read timeout to all requests (queries, writes, basically anything except blocking-queries).

agent/consul/client_test.go
kisunji: Blocking query default timeout (in the test setup): 100ms
`Long.Wait` returns after: 101ms
The max jitter is added when calculating blocking query timeouts: 100ms + 100/16 = 116ms

So theoretically if max jitter is ever applied by the server and it returns in 117ms then the test will flake. I will make the `Long.Wait` exactly 100ms

jkirschner-hashicorp: Side comment to consider in "the future":

Do we have a lot of timing dependent (possibly flaky) tests? If so, it might be worthwhile at some point to develop a mockable interface passed in via dependency injection so we can use real timing capabilities in the real code, but mocked versions (that can't flake) in the tests.

agent/pool/pool.go
kisunji: @banks I understand the concern about writes. But given the fact that we had 7s timeouts on writes in the last few releases and the only issues raised by the community and customers were regarding non-blocking queries, I think it's safe to apply a larger (maybe 300s) timeout across the board. What do you think?

agent/consul/client_test.go
kisunji: Agreed! We actually have an established mock in a few places like https://github.com/hashicorp/consul/blob/80e51ff907b2b68ac42e0d29773e480d07e59bd1/agent/grpc-external/services/peerstream/testing.go#L136

agent/pool/pool.go
kisunji: We can solve the issue of not having timeouts on writes first and if anything comes up where we find that write RPCs need to be tuned separately, maybe another config may be more appropriate.

agent/consul/client_test.go
banks: In the past timing-dependent tests have needed error margins in the order of 500ms or more to not be very flaky in CI. 

What happens if CI CPU is starved and so the goroutine handling the RPC isn't scheduled for 200ms and so doesn't "notice" the timeout, that would flake right?

Usually I'd treat tests like this that use actual non-deterministic goroutines and wall clock time as a very very rough smoke test that the timeout does anything at all rather than trying to assert anything very specific about exactly how long the timeout was. i.e. make it so that if you comment out the timeout code it always fails but maybe after like several _seconds_ then you can use much larger margins.

All that said, I've been much further from out test suite than you for a long time @kisunji so if you have good reason to think this will be robust I defer to your greater knowledge! I just know that any sleep-style test with timing tolerances lower than multiple hundreds of milliseconds used to be a non-starter in CI 😆 !

agent/pool/pool.go
banks: Oh nice. Yeah I couldn't recall if this was everything. If the old behaviour was to timeout writes too then I'd def not want to _remove_ that!

> I think the correct solution would be to rename rpc_read_timeout to rpc_client_timeout and have it apply the same read timeout to all requests (queries, writes, basically anything except blocking-queries).

Yeah this sounds ideal to me for now. We can add more config if we really need to.

banks (changes requested):
I think we do need to apply this to writes as well rather than "regress" and remove a timeout that was present for them in previous versions. Hopefully not a huge effort though.

kisunji:
> I think we do need to apply this to writes as well rather than "regress" and remove a timeout that was present for them in previous versions. Hopefully not a huge effort though.

Addressed in https://github.com/hashicorp/consul/pull/14965/commits/135a7ff8a17e289639ae8788c46bbdc867574c42

agent/pool/pool.go
```diff
@@ -387,6 +389,26 @@ func (p *ConnPool) dial(
 	return conn, hc, nil
 }
 
+func (p *ConnPool) RPCClientTimeout() time.Duration {
+	// Since clientTimeout represents uint32 milliseconds,
+	// convert to int64 first then multiply by 1000 to
+	// match time.Duration's nanoseconds.
+	return time.Duration(int64(atomic.LoadUint32(p.clientTimeout) * 1000))
```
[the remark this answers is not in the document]
banks: Also, I'd suggest `time.Millisecond` instead of the magic 1000. So:

```go
return time.Duration(atomic.LoadUint32(&p.clientTimeout)) * time.Millisecond
```

agent/pool/pool.go
```diff
@@ -127,16 +127,18 @@ func (c *Conn) markForUse() {
 // streams allowed. If TLS settings are provided outgoing connections
 // use TLS.
 type ConnPool struct {
+	// The default timeout for client RPC requests in milliseconds.
+	// Stored as an atomic uint32 value to allow for reloading.
+	// TODO: once we move to go1.19, change to atomic.Uint32.
+	clientTimeout *uint32
```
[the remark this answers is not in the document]
banks: personal preference thing but I'd be tempted to call it `clientTimeoutMs` or something to make it more obvious without reading the comment specifically what the number represents, but totally optional!

agent/pool/pool.go
```diff
@@ -387,6 +389,26 @@ func (p *ConnPool) dial(
...
 
+func (p *ConnPool) RPCClientTimeout() time.Duration {
+	// Since clientTimeout represents uint32 milliseconds,
+	// convert to int64 first then multiply by 1000 to
+	// match time.Duration's nanoseconds.
+	return time.Duration(int64(atomic.LoadUint32(p.clientTimeout) * 1000))
+}
+
+func (p *ConnPool) SetRPCClientTimeout(timeout time.Duration) {
+	if timeout/1000 > time.Duration(^uint32(0)) {
```
banks, line 404: I don't think this is right - `time.Duration` is expressed in nanoseconds and there are 1 million in a millisecond not 1000 so this seems off by three orders of magnitude.

In general you should never need to use hard-coded numbers like `1000` when working with durations in Go - the built in `time.Millisecond` etc. are more readable.

I think overall rather than trying to work out the absolute largest timeout we can accept which is ludicrously large I'd be tempted to make this whole method way simpler with a reasonable upper bound, something like:

```go

func (p *ConnPool) SetRPCClientTimeout(timeout time.Duration) {
  if timeout > time.Hour {
    // Prevent unreasonably large timeouts that might overflow a uint32
    timeout = time.Hour
  }
  
  atomic.StoreUint32(&p.clientTimeout, uint32(timeout.Milliseconds()))
}
```

banks (changes requested):
Overall this looks good.

Couple of small fixes noted inline but should be quick!

agent/pool/pool.go
kisunji: That sounds more reasonable and readable!

agent/pool/pool.go
```diff
@@ -158,18 +127,21 @@ func (c *Conn) markForUse() {
 // streams allowed. If TLS settings are provided outgoing connections
 // use TLS.
 type ConnPool struct {
+	// The default timeout for client RPC requests in milliseconds.
+	// Stored as an atomic uint32 value to allow for reloading.
+	// TODO: once we move to go1.19, change to atomic.Uint32.
```
banks, line 132:
```suggestion
	// clientTimeoutMs is the default timeout for client RPC requests 
	// in milliseconds. Stored as an atomic uint32 value to allow for 
	// reloading.
	// TODO: once we move to go1.19, change to atomic.Uint32.
```

banks: super minor nit but go idiom is to start comments with the name of the thing you're commenting though that's not even consistent in this file so totally optional!

agent/consul/server.go
```diff
@@ -1646,6 +1645,7 @@ func (s *Server) ReloadConfig(config ReloadableConfig) error {
 	s.rpcConnLimiter.SetConfig(connlimit.Config{
 		MaxConnsPerClientIP: config.RPCMaxConnsPerClient,
 	})
+	s.connPool.SetRPCClientTimeout(config.RPCClientTimeout)
```
banks, line 1648: Happy to leave it for now if there is time pressure to merge, but I didn't see a reload test covering this like you added in client_test.go. There might even be an existing Reload test you can just add a couple of lines to to cover this. Fine to do it as a follow up though!

banks (approved):
Amazing job @kisunji 🥳 

Thanks for patiently going through those extra bits to add reloading etc. This seems solid now!

agent/consul/server.go
kisunji: Added!

freddygv:
@kisunji can #14732 be closed now?

kisunji:
> @kisunji can #14732 be closed now?

Yeah I'm handling the backporting. Will close when everything is completely merged

PR #14974, opened 2022-10-13
ui: enable synthetic nodes in mock data

WenInCode:
### Description
Enable synthetic nodes in mock data by setting a cookie. I had to step through the `get-environment` utility to figure out what was going on, so I added a few comments. Michael and I already discussed some ways we could improve this, but that won't be included in this PR.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-ui/app/utils/get-environment.js
+++ b/ui/packages/consul-ui/app/utils/get-environment.js
@@ -18,6 +18,9 @@ export default function (config = {}, win = window, doc = document) {
         .filter((item) => item !== '')
         .filter((item) => item.split('=').shift().startsWith('CONSUL_'));
     };
+
+    // Define the function that reads in "Scenarios", parse and set cookies and set theme if specified.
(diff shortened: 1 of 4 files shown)


ui/packages/consul-ui/app/utils/get-environment.js
```diff
@@ -18,6 +18,9 @@ export default function (config = {}, win = window, doc = document) {
         .filter((item) => item !== '')
         .filter((item) => item.split('=').shift().startsWith('CONSUL_'));
     };
+
+    // Define the function that reads in "Scenarios", parse and set cookies and set theme if specified.
+    // See https://github.com/hashicorp/consul/blob/main/ui/packages/consul-ui/docs/bookmarklets.mdx
```
evrowe, line 23: Comments/hints like this are extremely useful, thanks for adding these!

evrowe (approved):
Looks good, make it so

PR #14977, opened 2022-10-13
ui: Bento-Box show scrollbars only when necessary

LevelbossMike:
### Description
Only show scrollbars on the peering bento-box when necessary.

Before (on Windows and with scrollbars set to `always` on macOS):
Screenshot 2022-10-13 at 20 29 44

After:
Screenshot 2022-10-13 at 20 30 09

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


jnwright (approved):
LGTM

WenInCode (approved):

PR #14986, opened 2022-10-14
UI: filter node healthchecks on agentless service instances

WenInCode:
### Description
Removes the "Node" type in "Search Across" dropdown on the service/instance/healthchecks page if the instance is using agentless. Also filters any healthchecks to not show any node checks, in case there were any.

- Kubernetes instances will be agentless on the vercel app so you can go to one of those and check it out.

### Screenshots
Screen Shot 2022-10-13 at 6 48 55 PM
Screen Shot 2022-10-13 at 6 48 08 PM

### Links
- [NET-1126](https://hashicorp.atlassian.net/browse/NET-1126)

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/.changelog/14986.txt
+++ b/.changelog/14986.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+ui: Filter out node health checks on agentless service instances
+```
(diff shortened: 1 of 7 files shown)


LevelbossMike (changes requested):
:clap:

PR #14988, opened 2022-10-14
proxycfg: fix goroutine leak when service is re-registered

boxofrad:
### Description
Fixes a bug where we'd leak a goroutine in `state.run` when the given context was canceled while there was a pending update.


boxofrad:
@mkeeler I decided not to swap `AfterFunc` (for something that doesn't hide the goroutine) in this PR to keep the change small. But we should probably revisit that!

mkeeler (approved):

riddhi89 (approved):
:shipit:

PR #14994, opened 2022-10-14
peering: skip registering duplicate node and check from the peer

huikang:
### Description
When some resource such as a check is imported we currently always apply the newly received resource. This happens even if the resource was not modified.

For each imported resource we should compute a diff relative to the data stored locally. This will cost some CPU time but can save trips through Raft when data has not changed.

- Node: skip CatalogRegister duplicate node
- Service: skip CatalogRegister duplicate service instance
- Check: skip CatalogRegister duplicate check

1. Register a service in the acceptor cluster with some check
2. Export the service
3. Change the meta or tag of the service, which changes the service. Then register the service again.
4. Issue a request to `/v1/health/service/` of the peering cluster.  As a result, the Index of the service will be changed, but that of check keeps the same.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/grpc-external/services/peerstream/replication.go
+++ b/agent/grpc-external/services/peerstream/replication.go
@@ -334,14 +334,14 @@ func (s *Server) handleUpdateService(
 
 	// Normalize the data into a convenient form for operation.
 	snap := newHealthSnapshot(structsNodes, partition, peerName)
+	storedNodesMap, storedChecksMap := buildStoredMap(storedInstances)
 
 	for _, nodeSnap := range snap.Nodes {
 		// First register the node - skip the existent ones
 		exist := false
-		for _, csn := range storedInstances {
-			if csn.Node.IsSame(nodeSnap.Node) {
+		if storedNode, ok := storedNodesMap[nodeSnap.Node.ID]; ok {
+			if storedNode.IsSame(nodeSnap.Node) {
 				exist = true
-				break
 			}
 		}
 
@@ -351,7 +351,6 @@ func (s *Server) handleUpdateService(
 				return fmt.Errorf("failed to register node: %w", err)
 			}
 		}
-
 		// Then register all services on that node
 		for _, svcSnap := range nodeSnap.Services {
 			req.Service = svcSnap.Service
@@ -366,19 +365,9 @@ func (s *Server) handleUpdateService(
 		for _, svcSnap := range nodeSnap.Services {
 			for _, c := range svcSnap.Checks {
 				exist = false
-				for _, stored := range storedInstances {
-					if stored.Service.Service != svcSnap.Service.Service {
-						continue
-					}
-
-					for _, chk := range stored.Checks {
-						if chk.IsSame(c) {
-							exist = true
-							break
-						}
-					}
-					if exist {
-						break
+				if chk, ok := storedChecksMap[makeNodeCheckID(nodeSnap.Node.ID, c.CheckID)]; ok {
+					if chk.IsSame(c) {
+						exist = true
 					}
 				}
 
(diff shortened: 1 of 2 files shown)


agent/grpc-external/services/peerstream/replication.go
```diff
@@ -351,17 +361,38 @@ func (s *Server) handleUpdateService(
...
 		for _, svcSnap := range nodeSnap.Services {
 			for _, c := range svcSnap.Checks {
-				chks = append(chks, c)
+				exist = false
+				for _, stored := range storedInstances {
+					if stored.Service.Service != svcSnap.Service.Service {
+						continue
+					}
+
+					for _, chk := range stored.Checks {
```
hashi-derek, line 374: Is there another, more performant way we could do this? I worry about 4 nested-loops being a performance issue in some situations.

huikang: Agreed; one option is to preload a map using the key of combined server and check ID and reuse it for every check.

agent/grpc-external/services/peerstream/replication.go
```diff
@@ -589,3 +609,23 @@ func makeReplicationResponse(resp *pbpeerstream.ReplicationMessage_Response) *pb
...
+	nodesMap := map[types.NodeID]*structs.Node{}
+
+	// key of checksmap will be nodeid + checkID
+	checksMap := map[string]*structs.HealthCheck{}
+	for _, csn := range storedInstances {
+		nodesMap[csn.Node.ID] = csn.Node
+
+		for _, chk := range csn.Checks {
+			// checksMap[string(csn.Node.ID)+"-"+string(chk.CheckID)] = chk
+			checksMap[makeNodeCheckID(csn.Node.ID, chk.CheckID)] = chk
```
huikang, line 627: Is this a right way to form a unique check id across the cluster? My understand is that the `c.CheckID` is unique per node; compounded with node.ID, it will generate a unique string for the check.

hashi-derek: I'm not completely sure. It is possible to use a struct as a key in Go, so maybe that could help to avoid any confusion relating to concatenating ids.

hashi-derek: When I look at the `newHealthSnapshot` function, it looks like it relies on `nodeId, serviceId, checkId` for uniqueness, but I don't know if all of that is necessary for identifying a check.

rboyer: Yes. Checks are tied to services, and services are tied to nodes. Service IDs are only unique on a node, so `(nodeID + serviceID)` is needed to identify a service. Putting it together then you can only uniquely identify a check by `(nodeID + serviceID + checkID)`

agent/grpc-external/services/peerstream/replication.go
```diff
@@ -589,3 +609,32 @@ func makeReplicationResponse(resp *pbpeerstream.ReplicationMessage_Response) *pb
 		},
 	}
 }
+
+// nodeCheckIdentity uniquely identifies a check imported from a peering cluster
+type nodeCheckIdentity struct {
+	peer    string
```
huikang, line 615: Adding peer name in order to avoid an odd case where two nodes from two peering clusters have the same ID with the same check ID.

hashi-derek: Is it possible to see that in this function, since the query includes the `peerName` to filter with?

`s.GetStore().CheckServiceNodes(nil, sn.Name, &sn.EnterpriseMeta, peerName)`

huikang: fixed. Thanks.

freddygv (approved):
Thanks!

Issue #11719, opened 2021-12-02
Consul logs error when ingress gateway is registered


lkysow:
#### Overview of the Issue

When an ingress gateway is registered through a config file Consul logs an error:

```
2021-12-02T13:44:13.139-0800 [ERROR] agent.proxycfg: Failed to handle update from watch: service_id=ingress-gateway id=gateway-config error="invalid type for config entry: <nil>"
```

#### Reproduction Steps

```hcl
# service.hcl
services {
  name   = "ingress-gateway"
  kind   = "ingress-gateway"
  port   = 20000
  checks = [
    {
      name     = "ingress-gateway listening"
      tcp      = "localhost:20000"
      interval = "30s"
    }
  ]
}
```

```
consul agent -dev -config-file=service.hcl
```

You'll then see the log.

### Consul info for both Client and Server

Consul 1.10.3

### Full logs

```
consul agent -dev -config-file consul.hcl
==> Starting Consul agent...
           Version: '1.10.3'
           Node ID: '28acf0bb-ac55-5917-f1a8-9eaecabcdce2'
         Node name: 'lkysow-C02WW13ZJG5L'
        Datacenter: 'dc1' (Segment: '<all>')
            Server: true (Bootstrap: false)
       Client Addr: [127.0.0.1] (HTTP: 8500, HTTPS: -1, gRPC: 8502, DNS: 8600)
      Cluster Addr: 127.0.0.1 (LAN: 8301, WAN: 8302)
           Encrypt: Gossip: false, TLS-Outgoing: false, TLS-Incoming: false, Auto-Encrypt-TLS: false

==> Log data will now stream in as it occurs:

2021-12-02T13:44:13.041-0800 [INFO]  agent.server.raft: initial configuration: index=1 servers="[{Suffrage:Voter ID:28acf0bb-ac55-5917-f1a8-9eaecabcdce2 Address:127.0.0.1:8300}]"
2021-12-02T13:44:13.042-0800 [INFO]  agent.server.raft: entering follower state: follower="Node at 127.0.0.1:8300 [Follower]" leader=
2021-12-02T13:44:13.043-0800 [INFO]  agent.server.serf.wan: serf: EventMemberJoin: lkysow-C02WW13ZJG5L.dc1 127.0.0.1
2021-12-02T13:44:13.044-0800 [INFO]  agent.server.serf.lan: serf: EventMemberJoin: lkysow-C02WW13ZJG5L 127.0.0.1
[...]

PR #15001, opened 2022-10-16
ingress-gateways: don't log error when registering gateway

lkysow:
Previously, when an ingress gateway was registered without a corresponding ingress gateway config entry, an error was logged because the watch on the config entry returned a nil result. This is expected since the first call to watch will return a nil entry so don't log an error.

Fixes https://github.com/hashicorp/consul/issues/11719

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/proxycfg/ingress_gateway.go
```diff
@@ -91,6 +91,9 @@ func (s *handlerIngressGateway) handleUpdate(ctx context.Context, u UpdateEvent,
 		if !ok {
 			return fmt.Errorf("invalid type for response: %T", u.Result)
 		}
```
lkysow, line 93: The fix is to guard on `resp.Entry` being nil. One thing this doesn't handle is if the config entry is deleted. If it's deleted, I think the expectation would be that everything is reset back to the baseline. This can be done in a different PR though and likely isn't worth it if the api gateway work is completed.

ishustava (approved):

PR #15005, opened 2022-10-17
ca: fix a masked bug in leaf cert generation that would not be notified of root cert rotation after the first one

rboyer:
### Description

In practice this was masked by #14956 and was only uncovered fixing the other bug.

    go test ./agent -run TestAgentConnectCALeafCert_goodNotLocal

would fail when only #14956 was fixed.

### DIFF
--- a/agent/cache-types/connect_ca_leaf.go
+++ b/agent/cache-types/connect_ca_leaf.go
@@ -165,6 +165,7 @@ func (c *ConnectCALeaf) fetchDone(rootUpdateCh chan struct{}) {
 	if len(c.rootWatchSubscribers) == 0 && c.rootWatchCancel != nil {
 		// This was the last request. Stop the root watcher.
 		c.rootWatchCancel()
+		c.rootWatchCancel = nil
 	}
 }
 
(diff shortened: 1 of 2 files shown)


agent/cache-types/connect_ca_leaf.go
```diff
@@ -165,6 +165,7 @@ func (c *ConnectCALeaf) fetchDone(rootUpdateCh chan struct{}) {
 	if len(c.rootWatchSubscribers) == 0 && c.rootWatchCancel != nil {
 		// This was the last request. Stop the root watcher.
 		c.rootWatchCancel()
+		c.rootWatchCancel = nil
```
rboyer, line 168: `fetchStart` uses the nil-ness of `rootWatchCancel` to decide if it needs to launch the goroutine again, so when we invoke the cancel and do not nil it out it means that the goroutine will never run again.

The fix for #14956 mostly caused the cache requests for things like ca roots to act more like the responses were cached or at least singleflighted better, and apparently that rugged "just try again" spirit kept this bug from manifesting in practice in leaf cert generation.

boxofrad (approved):
Yikes! Great catch 👏🏻

PR #15021, opened 2022-10-17
test: possibly fix flake in TestIntentionGetExact

rboyer:
### Description

Restructure test setup to be similar to `TestAgent_ServerCertificate` and see if that's enough to avoid flaking after join.


dhiaayachi (approved):
🚀

PR #15023, opened 2022-10-18
ui: Fix token in cookie passthrough

LevelbossMike:
### Description
Fixes the logic that was added in https://github.com/hashicorp/consul/pull/14495 to always fetch a token based on `TokenSecretID` when that is different from what we persisted to settings before.

This makes sure we can override the `TokenSecretID` and not only set it up once.

In addition to fixing the login logic when the application boots up, I added tests and created a mechanism that we can use from now on when we want to stub out configuration on the `env`-service.

* setup `TokenSecretID` in a cookie and see that it will call the token endpoint with the provided value when the application boots up.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-ui/tests/acceptance/hcp-login-test.js
+++ b/ui/packages/consul-ui/tests/acceptance/hcp-login-test.js
@@ -1,0 +1,173 @@
+import { module, test } from 'qunit';
+import { visit } from '@ember/test-helpers';
+import { setupApplicationTest } from 'ember-qunit';
+import { setupTestEnv } from 'consul-ui/services/env';
+import TokenRepo from 'consul-ui/services/repository/token';
+import SettingsService from 'consul-ui/services/settings';
+
+const TOKEN_SET_BY_HCP = 'token-set-by-hcp';
+
+module('Acceptance | hcp login', function (hooks) {
+  setupApplicationTest(hooks);
+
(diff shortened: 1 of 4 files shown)


ui/packages/consul-ui/tests/acceptance/hcp-login-test.js
```diff
@@ -0,0 +1,173 @@
...
+  module('with `CONSUL_HTTP_TOKEN` not set', function () {
+    hooks.beforeEach(function () {
+      setupTestEnv(this.owner, {
+        CONSUL_ACLS_ENABLED: true,
+      });
+    });
+
+    test('we do not call the token endpoint', async function (assert) {
+      this.owner.register(
+        'service:pository/token',
```
WenInCode, line 22:
```suggestion
        'service:repository/token',
```

WenInCode (changes requested):
Minor change (typo). Env stub looks great 👍

ui/packages/consul-ui/tests/acceptance/hcp-login-test.js
LevelbossMike: Thanks for spotting this!

WenInCode (approved):

PR #15025, opened 2022-10-18
proxycfg: watch service-defaults config entries

ishustava:
### Description
To support `Destinations` on the service-defaults (for tproxy with terminating gateway), we need to now also make servers watch service-defaults config entries. 

* Tested on Kubernetes with this test: https://github.com/hashicorp/consul-k8s/blob/0e18cb4a0375aeda87f43f11210b118fdef460cb/acceptance/tests/terminating-gateway/terminating_gateway_destinations_test.go and a consul image built with these changes.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/structs/config_entry.go
```diff
@@ -990,10 +990,10 @@ type PassiveHealthCheck struct {
 	// being removed from the pool.
 	MaxFailures uint32 `json:",omitempty" alias:"max_failures"`
 
-	// EnforcingConsecutive5xx is the % chance that a host will be actually ejected
+	// EnforcingConsecutive5XX is the % chance that a host will be actually ejected
 	// when an outlier status is detected through consecutive 5xx.
 	// This setting can be used to disable ejection or to ramp it up slowly. Defaults to 100.
-	EnforcingConsecutive5xx *uint32 `json:",omitempty" alias:"enforcing_consecutive_5xx"`
+	EnforcingConsecutive5XX *uint32 `json:",omitempty" alias:"enforcing_consecutive_5xx"`
```
ishustava, line 996: I renamed this but I'm not sure if it's safe. The reason for name is because when protoc runs, it generates a go struct with the `EnforcingConsecutive5xx` field according to naming conventions which then results in a field named `EnforcingConsecutive5Xx` (notice the capitalized `X` in the `5Xx`). This then results in `mog` failing to convert because the generated go struct doesn’t have field `EnforcingConsecutive5xx`.

@boxofrad let me know what you think.

boxofrad: Great call out! To tell you the truth, I'm not entirely sure whether it's safe or not.

I think it probably isn't because it'd change the field name in the messagepack representation we send in net/rpc calls and store in Raft snapshots. You could probably override `MarshalBinary` to get around this, but I'm wondering whether it's worth the risk of regressions.

An alternative workaround would be to manually implement the conversion function for the `PassiveHealthCheck` type, which is a shame for sure, but probably safer.

boxofrad (commented):
Looking great, nice job! 👏🏻 (see comment about field renaming)

agent/structs/config_entry.go
ishustava: Oh thanks for that information! Yeah, I had an uneasy feeling when I was doing a rename 😄 

I'll do it manually, thanks!

ishustava: Done! Please take a another look when you get a chance!

boxofrad (approved):
🥳

PR #15028, opened 2022-10-18
Update upgrade docs for 1.13.2.

freddygv:
### Description
In 1.13.2 we added a new flag called use_auto_cert to address issues previously documented in the upgrade guide. Originally there was no way to disable TLS for gRPC when auto-encrypt was in use, because TLS was enabled for gRPC due to the presence of auto-encrypt certs.

As of 1.13.2, using auto-encrypt certs as the signal to enable TLS for gRPC is opt-in only. Meaning that if anyone who had upgraded to 1.13 relied on that side-effect, they now need to explicitly configure it.

### Links
PR that included `use_auto_cert`: https://github.com/hashicorp/consul/pull/14269

### PR Checklist

* [ ] ~updated test coverage~
* [x] not a security concern

### DIFF
--- a/website/content/docs/upgrading/upgrade-specific.mdx
+++ b/website/content/docs/upgrading/upgrade-specific.mdx
@@ -27,9 +27,9 @@ upgrade flow.
 
 Prior to Consul 1.14, it was possible to encrypt communication between Consul and Envoy over `ports.grpc` using these settings.
 
-Consul 1.14 introduces [`ports.grpc_tls`](/docs/agent/config/config-files#grpc_tls_port), a new configuration 
-for encrypting communication over gRPC. The existing [`ports.grpc`](/docs/agent/config/config- 
-files#grpc_port) configuration **will stop supporting encryption in a future release**. As of version 1.14, 
+Consul 1.14 introduces [`ports.grpc_tls`](/docs/agent/config/config-files#grpc_tls_port), a new configuration
+for encrypting communication over gRPC. The existing [`ports.grpc`](/docs/agent/config/config-
+files#grpc_port) configuration **will stop supporting encryption in a future release**. As of version 1.14,
 `ports.grpc_tls` is the recommended configuration to encrypt gRPC traffic.
 
 For most environments, the Envoy communication to Consul is loop-back only and does not benefit from encryption.
@@ -51,7 +51,7 @@ review the following guidances relevant to your deployment:
 
 Upgrade to **Consul version 1.13.1 or later**.
 
-Consul 1.13.0 contains a bug that prevents Consul server agents from restoring 
+Consul 1.13.0 contains a bug that prevents Consul server agents from restoring
 saved state on startup if the state
 
 1. was generated before Consul 1.13 (such as during an upgrade), and
@@ -61,25 +61,22 @@ This bug is fixed in Consul versions 1.13.1 and newer.
 
 #### Service mesh deployments using auto-encrypt or auto-config
 
-**Do not upgrade to Consul 1.13 yet** if using
-[auto-encrypt](/docs/agent/config/config-files#auto_encrypt) or
-[auto-config](/docs/agent/config/config-files#auto_config).
+Upgrade to **Consul version 1.13.2 or later**.
 
-In Consul 1.13, auto-encrypt and auto-config both cause Consul
+In Consul 1.13.1, auto-encrypt and auto-config both cause Consul
 to require TLS for gRPC communication with Envoy proxies.
 In environments where Envoy proxies are not already configured
-to use TLS for gRPC, upgrading Consul 1.13 will cause
+to use TLS for gRPC, upgrading Consul 1.13.1 will cause
 Envoy proxies to disconnect from the control plane (Consul agents).
 
-The underlying cause is the same as discussed in
-[deployments without the HTTPS port enabled on Consul agents](#service-mesh-deployments-without-the-https-port-enabled-on-consul-agents).
-However, when using auto-encrypt or auto-config,
-the problem **cannot** currently be avoided by 
-[modifying the agent's TLS configuration](#modify-the-consul-agent-s-tls-configuration)
-because auto-encrypt and auto-config automatically set
-interface-generic TLS configuration in a manner similar to
-[`tls.defaults`](/docs/agent/config/config-files#tls_defaults).
-We are working to address this problem in an upcoming 1.13 patch release.
+If upgrading to version 1.13.2 from 1.13.1 you must enable
+[tls.grpc.use_auto_cert](https://developer.hashicorp.com/consul/docs/agent/config/config-files#use_auto_cert)
+if you currently use the Connect CA and auto-encrypt certs for TLS
+on the gRPC port, and do not have certificates configured via
+[tls.grpc](https://developer.hashicorp.com/consul/docs/agent/config/config-files#tls_grpc)
+or [tls.defaults](https://developer.hashicorp.com/consul/docs/agent/config/config-files#tls_defaults).
+The new `use_auto_cert` flag enables TLS for gRPC based on the presence
+of auto-encrypt certs.
 
 #### Service mesh deployments without the HTTPS port enabled on Consul agents ((#grpc-tls))
 


website/content/docs/upgrading/upgrade-specific.mdx
```diff
@@ -61,25 +61,22 @@ This bug is fixed in Consul versions 1.13.1 and newer.
 
 #### Service mesh deployments using auto-encrypt or auto-config
 
-**Do not upgrade to Consul 1.13 yet** if using
-[auto-encrypt](/docs/agent/config/config-files#auto_encrypt) or
-[auto-config](/docs/agent/config/config-files#auto_config).
+Upgrade to **Consul version 1.13.2 or later**.
```
jkirschner-hashicorp, line 64: I think we should still qualify the conditions in which this is critical (auto-encrypt/config).
```suggestion
Upgrade to **Consul version 1.13.2 or later** if using
[auto-encrypt](/docs/agent/config/config-files#auto_encrypt) or
[auto-config](/docs/agent/config/config-files#auto_config).
```

website/content/docs/upgrading/upgrade-specific.mdx
```diff
@@ -61,25 +61,22 @@ This bug is fixed in Consul versions 1.13.1 and newer.
...
 
-In Consul 1.13, auto-encrypt and auto-config both cause Consul
+In Consul 1.13.1, auto-encrypt and auto-config both cause Consul
```
jkirschner-hashicorp, line 66:
```suggestion
In Consul 1.13.0 - 1.13.1, auto-encrypt and auto-config both cause Consul
```

website/content/docs/upgrading/upgrade-specific.mdx
```diff
@@ -61,25 +61,22 @@ This bug is fixed in Consul versions 1.13.1 and newer.
...
 to require TLS for gRPC communication with Envoy proxies.
 In environments where Envoy proxies are not already configured
-to use TLS for gRPC, upgrading Consul 1.13 will cause
+to use TLS for gRPC, upgrading Consul 1.13.1 will cause
```
jkirschner-hashicorp, line 69:
```suggestion
to use TLS for gRPC, upgrading Consul Consul 1.13.0 - 1.13.1 will cause
```

website/content/docs/upgrading/upgrade-specific.mdx
```diff
@@ -61,25 +61,22 @@ This bug is fixed in Consul versions 1.13.1 and newer.
...
-The underlying cause is the same as discussed in
-[deployments without the HTTPS port enabled on Consul agents](#service-mesh-deployments-without-the-https-port-enabled-on-consul-agents).
-However, when using auto-encrypt or auto-config,
-the problem **cannot** currently be avoided by 
-[modifying the agent's TLS configuration](#modify-the-consul-agent-s-tls-configuration)
-because auto-encrypt and auto-config automatically set
-interface-generic TLS configuration in a manner similar to
-[`tls.defaults`](/docs/agent/config/config-files#tls_defaults).
-We are working to address this problem in an upcoming 1.13 patch release.
+If upgrading to version 1.13.2 from 1.13.1 you must enable
```
jkirschner-hashicorp, line 72: @freddygv :
- Does someone upgrading from <1.13 to 1.13.2+ need to touch `tls.grpc.use_auto_cert` at all? Or in some cases?
- Do we need to provide guidance on upgrading from 1.13.0/1 to 1.13.2? We had previously explicitly said "don't upgrade to 1.13 if using auto-config/encrypt", so I worry this paragraph is unnecessary detail that makes the guidance seem more complex.

jkirschner-hashicorp (commented):
@freddygv : Thanks for picking this up. A few questions/comments.

Also, should we update the [1.13.0 changelog entry](https://github.com/hashicorp/consul/blob/main/CHANGELOG.md#1130-august-9-2022) to say "this TLS for gRPC requirement is removed in 1.13.2" instead of "will be removed in a future 1.13 patch release"? The changelog modification may need to be a separate PR ... I've found that changelog modification backporting can cause merge conflicts (which would affect the docs backport merge as well). Easiest to just update the changelog on `main` only.

I also added the docs-cherrypick label so the change will apply immediately.

boxofrad (approved):
Looks good to me! ⭐ 

Well, I'm a little sad about the upgrade complexity and wish I'd caught it earlier, but the docs and final outcome look good 😄

website/content/docs/upgrading/upgrade-specific.mdx
freddygv: The motivation for adding this was a user incident where they upgraded from 1.13.1 to 1.13.2, and not using this new flag broke their setup. This set of circumstances is likely quite rare, but I erred on the side of documenting it because of someone having run into it.

Though this is actually relevant to someone upgrading from <= 1.13.1. The restriction about upgrading from 1.13.1 can be relaxed.

website/content/docs/upgrading/upgrade-specific.mdx
```diff
@@ -61,25 +61,24 @@ This bug is fixed in Consul versions 1.13.1 and newer.
...
+Upgrade to **Consul version 1.13.2 or later** if using
 [auto-encrypt](/docs/agent/config/config-files#auto_encrypt) or
 [auto-config](/docs/agent/config/config-files#auto_config).
 
-In Consul 1.13, auto-encrypt and auto-config both cause Consul
+In Consul 1.13.0 - 1.13.1, auto-encrypt and auto-config both cause Consul
 to require TLS for gRPC communication with Envoy proxies.
 In environments where Envoy proxies are not already configured
-to use TLS for gRPC, upgrading Consul 1.13 will cause
+to use TLS for gRPC, upgrading Consul Consul 1.13.0 - 1.13.1 will cause
```
jkirschner-hashicorp, line 71:
```suggestion
to use TLS for gRPC, upgrading to Consul 1.13.0 - 1.13.1 will cause
```

website/content/docs/upgrading/upgrade-specific.mdx
```diff
@@ -61,25 +61,24 @@ This bug is fixed in Consul versions 1.13.1 and newer.
...
-The underlying cause is the same as discussed in
-[deployments without the HTTPS port enabled on Consul agents](#service-mesh-deployments-without-the-https-port-enabled-on-consul-agents).
-However, when using auto-encrypt or auto-config,
-the problem **cannot** currently be avoided by 
-[modifying the agent's TLS configuration](#modify-the-consul-agent-s-tls-configuration)
-because auto-encrypt and auto-config automatically set
-interface-generic TLS configuration in a manner similar to
-[`tls.defaults`](/docs/agent/config/config-files#tls_defaults).
-We are working to address this problem in an upcoming 1.13 patch release.
+If upgrading to version 1.13.2 you must also enable
```
jkirschner-hashicorp, line 74:
```suggestion
If upgrading to version 1.13.2 or later, you must enable
```

website/content/docs/upgrading/upgrade-specific.mdx
```diff
@@ -61,25 +61,24 @@ This bug is fixed in Consul versions 1.13.1 and newer.
...
+[tls.grpc.use_auto_cert](/docs/agent/config/config-files#use_auto_cert)
+if you currently use the Connect CA and auto-encrypt certs for TLS
```
jkirschner-hashicorp, line 76: Doesn't the "if you currently use the Connect CA" part apply to anyone using Consul service mesh? (Why is that part of the qualifier needed?)

jkirschner-hashicorp (commented):
A few follow up questions / comments!

website/content/docs/upgrading/upgrade-specific.mdx
freddygv: The idea is that non-mutual TLS requires: servers to provide a server certificate, and clients to trust the CA that signed the server certificate. 

So "Connect CA" is referring to the CA portion, and "auto-encrypt" certs to the server cert portion.

Though given that auto-encrypt requires the Connect CA, this could be rephrased as:
```suggestion
if you currently use the auto-encrypt certs as the server certs for TLS
```

What do you think?

website/content/docs/upgrading/upgrade-specific.mdx
```diff
@@ -61,25 +61,24 @@ This bug is fixed in Consul versions 1.13.1 and newer.
...
-because auto-encrypt and auto-config automatically set
-interface-generic TLS configuration in a manner similar to
-[`tls.defaults`](/docs/agent/config/config-files#tls_defaults).
-We are working to address this problem in an upcoming 1.13 patch release.
+If upgrading to version 1.13.2 or later, you must enable
+[tls.grpc.use_auto_cert](/docs/agent/config/config-files#use_auto_cert)
+if you currently use the Connect CA and auto-encrypt certs for TLS
+on the gRPC port, and do not have certificates configured via
+[tls.grpc](/docs/agent/config/config-files#tls_grpc)
+or [tls.defaults](/docs/agent/config/config-files#tls_defaults).
```
freddygv, line 79:
```suggestion
if you currently use the auto-encrypt or auto-config certs as the 
server certs for TLS on the gRPC port.
```

jkirschner-hashicorp: Regarding "as the server certs for TLS on the gRPC port", what do we mean by "server" in this case? I don't think we mean "server agent certs", since I didn't think Consul server agents received auto-encrypt/config certs.

jkirschner-hashicorp: And is the tls.grpc or tls.defaults stuff still relevant? I noticed that was removed.

freddygv:
```suggestion
if you currently rely on Consul agents presenting the auto-encrypt or
auto-config certs as the TLS server certs for gRPC port.
```

freddygv: What's meant by "server" is the server side of the TLS connection, updated

jkirschner-hashicorp (approved):
LGTM

PR #15032, opened 2022-10-18
Add docs about upgrading primary mesh gateways

freddygv:
### Description
Care must be taken when replacing mesh gateways in the primary datacenter. If the old gateway addresses become unreachable before the secondary datacenters receive the new addresses, then the primary datacenter overall will become unreachable.

This commit adds docs related to this class of upgrades.

### DIFF
--- a/website/content/docs/connect/gateways/mesh-gateway/wan-federation-via-mesh-gateways.mdx
+++ b/website/content/docs/connect/gateways/mesh-gateway/wan-federation-via-mesh-gateways.mdx
@@ -187,10 +187,12 @@ expected result:
 
 ### Upgrading the primary gateways
 
-Once federation has been established, updates for the addresses of new primary
-gateways are propagated through the gateways in the primary themselves. If the primary
-gateways are upgraded, and their previous instances are decommissioned before
-the updates are propagated, then the primary datacenter will become unreachable.
+Once federation is established, secondary datacenters will continuously request
+updated mesh gateway addresses from the primary datacenter. These requests
+themselves flow through the mesh gateways of the primary datacenter, since
+secondary datacenters cannot dial the primary datacenter's Consul servers directly.
+If the primary gateways are upgraded, and their previous instances are decommissioned
+before the updates are propagated, then the primary datacenter will become unreachable.
 
 To safely upgrade primary gateways, we recommend that you apply one of the following policies:
 - Avoid decommissioning primary gateway IP addresses. This is because the [primary_gateways](/docs/agent/config/config-files#primary_gateways) addresses configured on the secondary servers act as a fallback mechanism for re-establishing connectivity to the primary.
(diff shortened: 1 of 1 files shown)


rboyer (approved):

trujillo-adam (commented):
Made a few suggestions that I think will improve readability and will make the content more consistent with our styles

website/content/docs/connect/gateways/mesh-gateway/wan-federation-via-mesh-gateways.mdx
```diff
@@ -184,3 +184,17 @@ expected result:
 - Ensure any API request that activates datacenter request forwarding. such as
   [`/v1/catalog/services?dc=<OTHER_DATACENTER_NAME>`](/api-docs/catalog#dc-1)
   succeeds.
+
+### Upgrading the primary gateways
+
+Once federation has been established, updates for the addresses of new primary
+gateways are propagated through the gateways in the primary themselves. If the primary
```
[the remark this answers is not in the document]
freddygv: No it's not, the updates that need to be propagated are address changes for mesh gateways in the primary datacenter. But your question helped me think of a different way to clarify this.

I pushed an update to make explicit that this propagation is actually initiated by the secondary datacenters. 

Here's an example that helped me think through this:
1. Primary datacenter has three mesh gateways: [1,2,3]
2. Secondary datacenter is continuously dialing the primary datacenter via gateways [1,2,3] to receive updated primary gateway addresses.
3. Primary datacenter quickly swaps gateways [1,2,3] for [4,5,6], taking down [1,2,3].

If step 3 happens fast enough, then the queries from the secondary datacenter in step 2 do not return addresses of [4,5,6] before [1,2,3] become unavailable. Step 2 will continue to dial [1,2,3] because those are the only addresses it knows about, but because they're gone that means the primary datacenter is unreachable.

website/content/docs/connect/gateways/mesh-gateway/wan-federation-via-mesh-gateways.mdx
```diff
@@ -184,3 +184,18 @@ expected result:
 - Ensure any API request that activates datacenter request forwarding. such as
   [`/v1/catalog/services?dc=<OTHER_DATACENTER_NAME>`](/api-docs/catalog#dc-1)
   succeeds.
+
+### Upgrading the primary gateways
+
+Once federation is established, secondary datacenters will continuously request
+updated mesh gateway addresses from the primary datacenter. These requests
+themselves flow through the mesh gateways of the primary datacenter, since
+secondary datacenters cannot dial the primary datacenter's Consul servers directly.
```
trujillo-adam, line 193:
```suggestion
updated mesh gateway addresses from the primary datacenter. Consul route the requests
 through the primary datacenter's mesh gateways. This is because
secondary datacenters cannot directly dial the primary datacenter's Consul servers.
```
Trying to make the language more precise by specifically attributing the action to Consul ("Consul routes requests . . .") and putting the adverb directly in front of the verb it's modifies ("directly dial"). Ignore if you don't feel these are improvements.

trujillo-adam (approved):
This looks good and makes sense. I made a few very minor suggestions, but feel free to reject them. Approving so you're not blocked.

website/content/docs/connect/gateways/mesh-gateway/wan-federation-via-mesh-gateways.mdx
freddygv: Looks good, will apply it.

PR #15035, opened 2022-10-18
Warn instead of returning error when missing intermediate mount tune permissions

kyhavlov:
This PR changes the behavior added in https://github.com/hashicorp/consul/pull/14516 - specifically, the fix required using the tune endpoint to update the `MaxLeaseTTL` in the intermediate PKI mount, which was a new permission in Vault we weren't explicitly documenting. This changes the failure case to emitting a warning log message instead of an error, which would've caused attempts to update the CA configuration to fail if this permission was missing.

### DIFF
--- a/agent/connect/ca/provider_vault.go
+++ b/agent/connect/ca/provider_vault.go
@@ -388,7 +388,7 @@ func (v *VaultProvider) setupIntermediatePKIPath() error {
 	} else {
 		err := v.tuneMountNamespaced(v.config.IntermediatePKINamespace, v.config.IntermediatePKIPath, &mountConfig)
 		if err != nil {
-			return err
+			v.logger.Warn("Could not update intermediate PKI mount settings", "path", v.config.IntermediatePKIPath, "error", err)
 		}
 	}
 
--- a/agent/connect/ca/provider_vault_test.go
+++ b/agent/connect/ca/provider_vault_test.go
@@ -20,13 +20,29 @@ import (
 )
 
 const pkiTestPolicy = `
-path "sys/mounts/*"
+path "sys/mounts"
 {
-  capabilities = ["create", "read", "update", "delete", "list", "sudo"]
+	capabilities = ["read"]
+}
+path "sys/mounts/pki-root"
+{
+	capabilities = ["create", "read", "update", "delete", "list"]
+}
+path "sys/mounts/pki-intermediate"
+{
+	capabilities = ["create", "read", "update", "delete", "list"]
+}
+path "sys/mounts/pki-intermediate/tune"
+{
+	capabilities = ["update"]
+}
+path "pki-root/*"
+{
+	capabilities = ["create", "read", "update", "delete", "list"]
 }
 path "pki-intermediate/*"
 {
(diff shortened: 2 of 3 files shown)


agent/connect/ca/provider_vault.go
```diff
@@ -388,7 +388,7 @@ func (v *VaultProvider) setupIntermediatePKIPath() error {
 	} else {
 		err := v.tuneMountNamespaced(v.config.IntermediatePKINamespace, v.config.IntermediatePKIPath, &mountConfig)
 		if err != nil {
-			return err
+			v.logger.Warn("Could not update intermediate PKI mount settings", "path", v.config.IntermediatePKIPath, "error", err)
```
jkirschner-hashicorp, line 391: @kyhavlov : Does `err` here contain any messaging explaining that the problem is that the read capability in Vault is missing on the path?

agent/connect/ca/provider_vault_test.go
```diff
@@ -794,6 +794,98 @@ func TestVaultProvider_RotateAuthMethodToken(t *testing.T) {
 	}, 10*time.Second, 100*time.Millisecond)
 }
 
+func TestVaultProvider_ReconfigureIntermediateTTL(t *testing.T) {
+	SkipIfVaultNotPresent(t)
+
+	policy := `
```
jkirschner-hashicorp, line 800: @kyhavlov : Should we modify the other tests in here to use the minimal, recommended permissions (demonstrated in this test) rather than the [maximal (not recommended) permissions](https://github.com/hashicorp/consul/blob/da8d6db30df512746c76acf0ca3773c8088ccfef/agent/connect/ca/provider_vault_test.go#L22-L30) currently used for most of the tests?

That would help us catch similar issues in the future, if the Vault CA provider code is modified to require access to a new endpoint (that needs new permissions). Without that change to the test permissions, it's easy for an issue like this one to recur.

agent/connect/ca/provider_vault.go
kyhavlov: Yes - if it's due to permission denied, it'll have that in the error message as well as the endpoint it was trying to access (`../tune` in this case)

kyhavlov: The log message looks like this: 
```
2022-10-18T11:38:23.609-0700 [WARN]  Could not update intermediate PKI mount settings: path=pki-intermediate/
  error=
  | Error making API request.
  | 
  | URL: POST http://127.0.0.1:16001/v1/sys/mounts/pki-intermediate/tune
  | Code: 403. Errors:
  | 
  | * 1 error occurred:
  | 	* permission denied
  | ```

agent/connect/ca/provider_vault_test.go
jkirschner-hashicorp: @kyhavlov : Should the read permissions on the `tune` path be included in the standard Vault policies for all tests? (I realize you'd need to temporarily remove that for the test case where you are seeing what happens if that isn't present, but it's now a part of the "standard" permissions we recommended, right?)

kyhavlov: Good call - none of the other tests need it yet, but it could be relevant in future tests.

PR #15050, opened 2022-10-19
Allow redirection to a different consul dns port

ishustava:
### Description

### Links

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


trujillo-adam (approved):
Docs look good

sdk/iptables/iptables.go
```diff
@@ -107,14 +111,27 @@ func Setup(cfg Config) error {
...
 			cfg.IptablesProvider.AddRule("iptables", "-t", "nat", "-A", "OUTPUT", "-p", "tcp", "--dport", "53", "-j", DNSChain)
+		} else if cfg.ConsulDNSPort != 0 {
+			consulDNSIP := "127.0.0.1"
+			if cfg.ConsulDNSIP != "" {
+				consulDNSIP = cfg.ConsulDNSIP
+			}
+			consulDNSHostPort := fmt.Sprintf("%s:%d", consulDNSIP, cfg.ConsulDNSPort)
+			// Traffic in the DNSChain is directed to the Consul DNS Service IP.
+			cfg.IptablesProvider.AddRule("iptables", "-t", "nat", "-A", DNSChain, "-p", "udp", "-d", consulDNSIP, "--dport", "53", "-j", "DNAT", "--to-destination", consulDNSHostPort)
+			cfg.IptablesProvider.AddRule("iptables", "-t", "nat", "-A", DNSChain, "-p", "tcp", "-d", consulDNSIP, "--dport", "53", "-j", "DNAT", "--to-destination", consulDNSHostPort)
```
boxofrad, line 130: Disclaimer: I know very little about `iptables` so this might be a silly question! 😅 

It looks to me like setting both `ConsulDNSIP` and `ConsulDNSPort` causes us to pass the given IP as the `-d` (destination) flag here, which we don't do if only `ConsulDNSIP` is given.

Is that correct? Wouldn't that only redirect packets already destined for the given `ConsulDNSIP`, when presumably we want to redirect all packets on port 53?

ishustava: Yep that's correct. Prior to this release, we allowed to pass consul dns IP, but now we also want to pass a port. So I kept previous behavior for backward compatibility.

riddhi89 (approved):
LGTM

PR #15065, opened 2022-10-19
fix: wan address isn't used by peering token

### LINKED ISSUE #15051: Cluster peering disregards `advertise_addr_wan` and `advertise_addr_wan_ipv4`
I have been unable to have Consul respect `advertise_addr_wan` and `advertise_addr_wan_ipv4` in the tokens generated via `consul peering generate-token`.

The base64 encoded json token contents always contain the `advertise_addr` / advertise_addr_ipv4`, which is unreachable from the remote cluster:
```
ServerAddresses: [ '__internal_lan_ip__:8502', '__internal_lan_ip__:8502', '__internal_lan_ip__:8502' ]
```

On the cluster initiating the connection (with the generated token), the following is logged:
```
error: desc = \"transport: Error while dialing dial tcp __remote_internal_lan_ip__:8502: i/o timeout
error managing peering stream
rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing dial tcp __remote_internal_lan_ip__:8502: i/o timeout
```

The proper address can be forced by editing the token:
```
let x = JSON.parse(atob('...'))
x[0] = '__external_wan_ip__:8502';
...
x[N] = '__external_wan_ip__:8502';
console.log(btoa()JSON.stringify(x)))
```
however, after the initial peering is successful, the `__internal_lan_ip__` from servers on either side are synced and replace the forced `__external_wan_ip__`; immediately afterwards, the peering connection fails

huikang:
### Description
Use wan address in the peering token.

With the following agent config,
```
{
   bind_addr =  “172.16.1.10”
   advertise_addr_wan = “192.168.1.10"
}
ports {
  grpc_tls = 8502
  serf_wan = 8302
},
```
the generated token always has the private bind address in "ServerAddresses":["172.16.1.10:8502"]," , which prevents the peering cluster from reaching the acceptor cluster.

### Links
fix #15051

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/agent/consul/peering_backend.go
+++ b/agent/consul/peering_backend.go
@@ -207,14 +207,9 @@ func meshGatewayAdresses(state *state.Store, ws memdb.WatchSet, wan bool) ([]str
 
 func parseNodeAddr(node *structs.ServiceNode) string {
 	// Prefer the wan address
-	if v, ok := node.TaggedAddresses[structs.TaggedAddressWANIPv4]; ok {
-		return v
-	}
-
 	if v, ok := node.TaggedAddresses[structs.TaggedAddressWAN]; ok {
 		return v
 	}
-
 	return node.Address
 }
 
--- a/agent/consul/peering_backend.go
+++ b/agent/consul/peering_backend.go
@@ -205,6 +205,19 @@ func meshGatewayAdresses(state *state.Store, ws memdb.WatchSet, wan bool) ([]str
 	return addrs, nil
 }
 
+func parseNodeAddr(node *structs.ServiceNode) string {
+	// Prefer the wan address
+	if v, ok := node.TaggedAddresses[structs.TaggedAddressWANIPv4]; ok {
+		return v
+	}
+
+	if v, ok := node.TaggedAddresses[structs.TaggedAddressWAN]; ok {
+		return v
+	}
+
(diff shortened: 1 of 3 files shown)


agent/consul/peering_backend.go
```diff
@@ -205,23 +205,38 @@ func meshGatewayAdresses(state *state.Store, ws memdb.WatchSet, wan bool) ([]str
 	return addrs, nil
 }
 
+func parseNodeAddr(node *structs.ServiceNode) string {
+	// Prefer the wan address
+	if v, ok := node.TaggedAddresses[structs.TaggedAddressWANIPv4]; ok {
```
quinndiggitypolymath, line 210: Just a note - I don't make use of `advertise_addr_wan_ipv6` (`ipv6.disable=1`), but others may; what should the preference/priority be between `advertise_addr_wan`, `advertise_addr_wan_ipv4`, and `advertise_addr_wan_ipv6`?

freddygv: That's a good question, will see about how that preference could be communicated in a follow-up PR.

For now will defer to the parsed `TaggedAddressWAN`, which could be IPv6.

freddygv (approved):

PR #15066, opened 2022-10-19
Committing for - 15055

### LINKED ISSUE #15055: Need a better error log message when a logEntry in the snapshot corresponding to Consul Ent is being restored by Consul OSS
Please search the existing issues for relevant feature requests, and use the reaction feature (https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to add upvotes to pre-existing requests.

#### Feature Description

a. When accidentally switching from Consul Ent to Consul OSS, users may see an error message similar to `'Unrecognized msg type %d'`
b. This message does not hint at the actual error.
c. The proposal here is to modify the message a bit to let the practitioner know that they might be operating using a flavor different than intended.
d. Here is the current error message - `'Unrecognized msg type %d'` that seems incomplete to the end-user. 
e. An error message like the one below does provide an additional hint on the possible mismatch of the flavors of the Consul binary used for restoring.
Error - `'msg type <%d> is a Consul Enterprise log entry. Consul OSS does not recognize or load it'`. 
  This could clarify or hint at the actual issue.

#### Use Case(s)

Modifying the error message could serve as a quick hint to the practitioner on a (possibly unintentional) mismatch in the flavor of the Consul binary.

Please note: I have the change & its corresponding unit test case ready. I will submit a PR soon.

vyanamandra:
Closes #15055

### Description
The new error message will indicate that the log entry is a Consul Enterprise one and will not be recognized or loaded by Consul OSS.

Repro:
* Create a snapshot using a Consul Enterprise build.
* Change the binary to Consul OSS and try to restore the snapshot taken from the above step.
* You will notice the message - `'Unrecognized msg type %d'` and the startup would end.

Testing the changes:
* Repeating the above steps with the new binary will lead to the message - `'msg type <%d> is a Consul Enterprise log entry. Consul OSS does not recognize or load it'`
* Unit test case has been added to - `agent/consul/fsm/snapshot_oss_test.go`
* You could also test the entire `fsm` piece using - `go test github.com/hashicorp/consul/agent/consul/fsm`

### Links
This is for -> https://github.com/hashicorp/consul/issues/15055

### PR Checklist

* [Y] updated test coverage
* [N] external facing docs updated
* [Y] not a security concern

### DIFF
--- a/agent/consul/fsm/fsm.go
+++ b/agent/consul/fsm/fsm.go
@@ -197,7 +197,11 @@ func (c *FSM) Restore(old io.ReadCloser) error {
 				return err
 			}
 		default:
-			return fmt.Errorf("Unrecognized msg type %d", msg)
+			if msg > 63 {
+				return fmt.Errorf("msg type <%d> is a Consul Enterprise log entry. Consul OSS does not recognize or load it", msg)
+			} else {
+				return fmt.Errorf("Unrecognized msg type %d", msg)
+			}
 		}
 		return nil
 	}
--- a/agent/consul/fsm/fsm.go
+++ b/agent/consul/fsm/fsm.go
@@ -197,7 +197,7 @@ func (c *FSM) Restore(old io.ReadCloser) error {
 				return err
 			}
 		default:
-			if msg > 63 {
+			if msg >= 64 {
 				return fmt.Errorf("msg type <%d> is a Consul Enterprise log entry. Consul OSS cannot restore it", msg)
 			} else {
 				return fmt.Errorf("Unrecognized msg type %d", msg)
--- a/agent/consul/fsm/fsm.go
+++ b/agent/consul/fsm/fsm.go
@@ -198,7 +198,7 @@ func (c *FSM) Restore(old io.ReadCloser) error {
 			}
 		default:
 			if msg > 63 {
-				return fmt.Errorf("msg type <%d> is a Consul Enterprise log entry. Consul OSS does not recognize or load it", msg)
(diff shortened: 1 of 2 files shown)


agent/consul/fsm/fsm.go
```diff
@@ -197,7 +197,11 @@ func (c *FSM) Restore(old io.ReadCloser) error {
 				return err
 			}
 		default:
-			return fmt.Errorf("Unrecognized msg type %d", msg)
+			if msg > 63 {
```
jkirschner-hashicorp, line 200: Optional comment: I find `msg >= 64` more readable than `msg > 63`.

Will ask engineering for review!

kisunji: Agreed with favoring >=64

agent/consul/fsm/fsm.go
```diff
@@ -197,7 +197,11 @@ func (c *FSM) Restore(old io.ReadCloser) error {
...
+				return fmt.Errorf("msg type <%d> is a Consul Enterprise log entry. Consul OSS does not recognize or load it", msg)
```
kisunji, line 201: Would this be a little clearer?
```suggestion
				return fmt.Errorf("msg type <%d> is a Consul Enterprise log entry. Consul OSS cannot restore it", msg)
```

kisunji (approved):
minor comments but LGTM

agent/consul/fsm/fsm.go
vyanamandra: Thank you for the review. I have modified the check to make it readable as suggested.

PR #15069, opened 2022-10-20
integration test: specify image name for testing dev container

huikang:
### Description
Sometimes we need to run integration for a dev image such as `hashicorppreview/consul:1.14-dev`. The current implementation hardcoded the image name to be `consul`.

This PR adds options to specify the image name to be tested.
- Some note is added to `README`

NA

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/test/integration/consul-container/libs/node/consul-container.go
+++ b/test/integration/consul-container/libs/node/consul-container.go
@@ -144,7 +144,7 @@ func newContainerRequest(config Config, name, configFile, dataDir, license strin
 
 	app := testcontainers.ContainerRequest{
 		NetworkMode: dockercontainer.NetworkMode("container:" + name + "-pod"),
-		Image:       consulImage + ":" + config.Version,
+		Image:       config.Image + ":" + config.Version,
 		WaitingFor:  wait.ForLog(bootLogLine).WithStartupTimeout(10 * time.Second),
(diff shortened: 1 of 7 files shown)


dhiaayachi:
I'm not against this change but I want to give context on why the version was initially part of the config and not the image. The config is supposed to abstract a consul agent/server without exposing any information about how it was deployed. 

At the time when this was implemented I was not able to find the right abstraction that would work for different environments (ex: test-containers, AWS VMs, K8S...) so I decided that it would be hardcoded at the the test-container layer.

I think if we would like to expose it we need to find a right abstraction for it otherwise that mean the config is now specific to test-container.

dhiaayachi:
I think that the best course is to separate the test-container dependencies in a separate package other than `node` for example `environment` and do some composition to compose the environment and the node and create a node out of it, that way we can create the environment separately and inject its own config in it which in this case would have the image.
That said, I think that is a bigger change that we should get to if we decide to support anything else other than test-container. For the time being, I think it's ok to expose the image for now as a config and maybe add a TODO in the code to remind about this.

dhiaayachi (approved):
LGTM!

huikang:
@dhiaayachi , sounds a plan; a TODO is added to the code. Thanks for the review.

PR #15083, opened 2022-10-20
Avoid blocking child type updates on parent ack

freddygv:
### Description
When providing updates to Envoy it is safest to follow the update order of: Clusters, Endpoints, Listeners, Routes. This order is encoded in our delta xDS response generation code.

Another noteworthy attribute of these resources is that there are parent->child relationships between: Cluster->Endpoints and Listener->Routes. Envoy couples the storage of a parent and a child type such that whenever a control plane sends an update for a parent type, it MUST also send fresh data for the child type, even if Envoy previously had the latest data for the child type.

A wrinkle that isn't accounted for currently is that Envoy does not actually ACK parent types until after attempting to receive data for the child types. This behavior also actually differs between Clusters and Listeners:
- When a cluster is sent, Envoy will wait up to a 15s timeout for endpoints to arrive before sending the cluster ACK.
- When a listener is sent, and it specifies RDS routes, Envoy will wait until those routes arrive before ACKing the listener. Though it's not clear to me what the timeout is for this, if any, since it exceeds a minute from my tests.

However, a behavior encoded in our xDS update order is that we avoid sending ANY data if we are waiting on ACKs for ANY resource. Meaning that when we first send a cluster to Envoy and it requests endpoints for that cluster, the endpoints do not actually get sent for at least 15s. This is because the endpoint update is paused by Consul until the cluster ACK, which Envoy pauses until it gets endpoints or times out.

This commit changes the xDS order gating such that we only block listener/route updates if there are cluster/endpoin updates pending. This is to avoid routing to an invalid destination.

* Ran manual tests for various scenarios against a live Envoy to validate assumptions (Envoy v1.21 and v1.23)
* Updated the unit tests to more accurately reflect Envoy's behavior
* Anecdotally the integration tests seem to be running ~1min faster than usual with this patch

### Links
Likely related to: https://github.com/envoyproxy/envoy/pull/5887

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/xds/delta.go
+++ b/agent/xds/delta.go
@@ -444,10 +444,6 @@ type xDSDeltaChild struct {
 	childrenNames map[string][]string
 }
 
-func (c *xDSDeltaChild) forceResubscribe(logger *hclog.Logger, parent string, child string) {
-
-}
-
 type xDSDeltaType struct {
 	generator    *ResourceGenerator
 	stream       ADSDeltaStream
@@ -730,7 +726,9 @@ func (t *xDSDeltaType) SendIfNew(
 	// we MUST send new data for all its children. Envoy will NOT re-subscribe to the child data upon
 	// receiving updates for the parent, so we need to handle this ourselves.
 	//
-	// Note that we do not check whether the deltaChild.childType is registered here, since we
+	// Note that we do not check whether the deltaChild.childType is registered here, since we send
+	// parent types before child types, meaning that it's expected on first send of a parent that
+	// there are no subscriptions for the child type.
 	if t.deltaChild != nil {
 		for name := range updates {
 			if children, ok := resourceMap.ChildIndex[t.typeURL][name]; ok {
--- a/agent/xds/delta.go
+++ b/agent/xds/delta.go
@@ -362,11 +362,17 @@ func (s *Server) processDelta(stream ADSDeltaStream, reqCh <-chan *envoy_discove
 					if clusterHandler := handlers[xdscommon.ClusterType]; clusterHandler.registered && len(clusterHandler.pendingUpdates) > 0 {
 						generator.Logger.Trace("Skipping delta computation for resource because there are dependent updates pending",
 							"typeUrl", op.TypeUrl, "dependent", xdscommon.ClusterType)
+
+						// Receiving an ACK from Envoy will unblock the select statement above,
+						// and re-trigger an attempt to send these skipped updates.
 						break
 					}
 					if endpointHandler := handlers[xdscommon.EndpointType]; endpointHandler.registered && len(endpointHandler.pendingUpdates) > 0 {
 						generator.Logger.Trace("Skipping delta computation for resource because there are dependent updates pending",
 							"typeUrl", op.TypeUrl, "dependent", xdscommon.EndpointType)
+
+						// Receiving an ACK from Envoy will unblock the select statement above,
+						// and re-trigger an attempt to send these skipped updates.
 						break
 					}
 				}
--- a/agent/xds/delta.go
+++ b/agent/xds/delta.go
@@ -4,7 +4,6 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
-	"sort"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -158,8 +157,14 @@ func (s *Server) processDelta(stream ADSDeltaStream, reqCh <-chan *envoy_discove
 	// representation of envoy state to force an update.
 	//
 	// see: https://github.com/envoyproxy/envoy/issues/13009
-	handlers[xdscommon.ListenerType].childType = handlers[xdscommon.RouteType]
-	handlers[xdscommon.ClusterType].childType = handlers[xdscommon.EndpointType]
+	handlers[xdscommon.ListenerType].deltaChild = &xDSDeltaChild{
+		childType:     handlers[xdscommon.RouteType],
+		childrenNames: make(map[string][]string),
+	}
+	handlers[xdscommon.ClusterType].deltaChild = &xDSDeltaChild{
+		childType:     handlers[xdscommon.EndpointType],
+		childrenNames: make(map[string][]string),
+	}
 
 	var authTimer <-chan time.Time
 	extendAuthTimer := func() {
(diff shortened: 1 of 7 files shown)


freddygv:
@erichaberkorn Something that would be good to look into is having a couple Delta xDS protocol tests like in `delta_test.go` but against live Envoys, to catch these sorts of expectations mismatches

agent/xds/delta.go
```diff
@@ -346,30 +351,26 @@ func (s *Server) processDelta(stream ADSDeltaStream, reqCh <-chan *envoy_discove
...
 
 			for _, op := range xDSUpdateOrder {
-				err, sent := handlers[op.TypeUrl].SendIfNew(
+				if op.TypeUrl == xdscommon.ListenerType || op.TypeUrl == xdscommon.RouteType {
+					if clusterHandler := handlers[xdscommon.ClusterType]; clusterHandler.registered && len(clusterHandler.pendingUpdates) > 0 {
+						generator.Logger.Trace("Skipping delta computation for resource because there are dependent updates pending",
+							"typeUrl", op.TypeUrl, "dependent", xdscommon.ClusterType)
+						break
+					}
+					if endpointHandler := handlers[xdscommon.EndpointType]; endpointHandler.registered && len(endpointHandler.pendingUpdates) > 0 {
```
freddygv, line 367: This endpoints check might be redundant given that Envoy waits for endpoints before ACKing clusters

agent/xds/endpoints.go
```diff
@@ -650,10 +650,6 @@ func (s *ResourceGenerator) endpointsFromDiscoveryChain(
 		primaryTargetID := node.Resolver.Target
 		failover := node.Resolver.Failover
 
-		type targetLoadAssignmentOption struct {
```
freddygv, line 653: Dead code

agent/xds/delta.go
```diff
@@ -721,14 +726,21 @@ func (t *xDSDeltaType) SendIfNew(
...
-	if t.childType != nil {
-		// Capture the relevant child resource names on this pending update so
-		// we can properly clean up the linked children when this change is
-		// ACKed.
-		for name, obj := range updates {
+	// Certain xDS types are children of other types, meaning that if an update is pushed for a parent,
+	// we MUST send new data for all its children. Envoy will NOT re-subscribe to the child data upon
+	// receiving updates for the parent, so we need to handle this ourselves.
+	//
+	// Note that we do not check whether the deltaChild.childType is registered here, since we
```
erichaberkorn, line 733: Could you complete this sentence?

agent/xds/delta.go
```diff
@@ -346,30 +351,26 @@ func (s *Server) processDelta(stream ADSDeltaStream, reqCh <-chan *envoy_discove
...
+						generator.Logger.Trace("Skipping delta computation for resource because there are dependent updates pending",
+							"typeUrl", op.TypeUrl, "dependent", xdscommon.ClusterType)
+						break
+					}
+					if endpointHandler := handlers[xdscommon.EndpointType]; endpointHandler.registered && len(endpointHandler.pendingUpdates) > 0 {
+						generator.Logger.Trace("Skipping delta computation for resource because there are dependent updates pending",
+							"typeUrl", op.TypeUrl, "dependent", xdscommon.EndpointType)
+						break
+					}
+				}
```
erichaberkorn, line 372: Is this to enforce the Clusters, Endpoints, Listeners, Routes order?

erichaberkorn: How are the xDS updates being retriggered when we skip them here?

agent/xds/delta.go
```diff
@@ -435,16 +433,30 @@ func (op *xDSUpdateOperation) errorLogNameReplyPrefix() string {
...
+	// Upserts of THIS type should potentially trigger dependent named
+	// resources within the child to be re-configured.
+	childType *xDSDeltaType
+
+	// childrenNames is map of parent resource names to a list of associated child resource
+	// names.
+	childrenNames map[string][]string
+}
+
+func (c *xDSDeltaChild) forceResubscribe(logger *hclog.Logger, parent string, child string) {
```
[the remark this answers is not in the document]
freddygv: Ah nope, this became `ensureChildResend`. Removed!

agent/xds/delta.go
freddygv: Fixed

agent/xds/delta.go
freddygv: > Is this to enforce the Clusters, Endpoints, Listeners, Routes order?
Yes.

> How are the xDS updates being retriggered when we skip them here?

The updates get unblocked by the select statement at the beginning of this big loop. Pushing xDS updates is generally triggered by Envoy sending us a message (receiving on `reqCh`) or a new proxycfg snapshot due to internal updates (receiving on `stateCh`). The message from Envoy could be subscribe/unsubscribe/ack/nack. 

So if we block updating listeners because of pending updates for endpoints, once those endpoints are ACKd. In the NACK case we actually re-block for changes because we assume that we got bad data from proxycfg and want to avoid a hot loop of errors (has happened before).

Re-blocking on NACK should be OK since if clusters/endpoints are NACKed we still don't want to send listeners.

erichaberkorn: Thank you

erichaberkorn (approved):

agent/xds/delta.go
```diff
@@ -346,30 +351,26 @@ func (s *Server) processDelta(stream ADSDeltaStream, reqCh <-chan *envoy_discove
...
 
 			streamStartOnce.Do(func() {
 				metrics.MeasureSince([]string{"xds", "server", "streamStart"}, streamStartTime)
 			})
 
 			for _, op := range xDSUpdateOrder {
-				err, sent := handlers[op.TypeUrl].SendIfNew(
+				if op.TypeUrl == xdscommon.ListenerType || op.TypeUrl == xdscommon.RouteType {
+					if clusterHandler := handlers[xdscommon.ClusterType]; clusterHandler.registered && len(clusterHandler.pendingUpdates) > 0 {
+						generator.Logger.Trace("Skipping delta computation for resource because there are dependent updates pending",
```
rboyer, line 363:
```suggestion
						generator.Logger.Trace("Skipping remaining delta computation for resource because there are dependent updates pending",
```
Small clarification to the log.

agent/xds/delta.go
```diff
@@ -346,30 +351,26 @@ func (s *Server) processDelta(stream ADSDeltaStream, reqCh <-chan *envoy_discove
...
+							"typeUrl", op.TypeUrl, "dependent", xdscommon.ClusterType)
+						break
+					}
+					if endpointHandler := handlers[xdscommon.EndpointType]; endpointHandler.registered && len(endpointHandler.pendingUpdates) > 0 {
+						generator.Logger.Trace("Skipping delta computation for resource because there are dependent updates pending",
```
rboyer, line 368:
```suggestion
						generator.Logger.Trace("Skipping remaining delta computation for resource because there are dependent updates pending",
```
Small clarification to the log.

rboyer (approved):

PR #15085, opened 2022-10-20
ui: NET-889 pre-populate partition SSO login

WenInCode:
### Description
- Default the "admin partition" input on SSO to the partition that is selected or "default" if one is not selected.

### Testing
- I added a test for the default case, but failed to add a test for a custom partition. I managed to get it to visit an `/example-partition/dc-1/services` url but I don't think it was properly creating the partition 🤷. This didn't take long but I've been trying for a while to get this test to work. Open to suggestions.

### Links
- [NET-889](https://hashicorp.atlassian.net/browse/NET-889)
- [Design](https://www.figma.com/file/E7shRv4L2rhgdYef18zSOD/SSO-updates?node-id=2%3A75)

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-acls/app/components/consul/token/selector/index.hbs
+++ b/ui/packages/consul-acls/app/components/consul/token/selector/index.hbs
@@ -51,7 +51,6 @@
               />
               {{#if (can "use SSO")}}
                 <authForm.Method @matches="sso">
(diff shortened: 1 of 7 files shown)


evrowe (approved):
Solid 😎

Issue #3897, opened 2018-02-19
`consul validate` fails with "Config validation failed: Multiple private IPv4 addresses found."


edganiukov:
### Description of the Issue (and unexpected/desired result)
When host has two or more interfaces, `consul validate` fails with:
```
Config validation failed: Multiple private IPv4 addresses found. Please configure one with 'bind' and/or 'advertise'.
```
In the same time, consul agent works, because it has `bind` address configured, so validation passed. The error comes from:
https://github.com/hashicorp/consul/blob/ecee15b4668e33762d36850b74e4d462bb39b917/command/validate/validate.go#L48

In documentation:
> The `consul validate` command performs a thorough sanity test on Consul configuration files. For each file or directory given, the command will attempt to parse the contents just as the `consul agent` command would, and catch any errors.

But with `consul validate` I am not able to provide, for example, `-bind` CLI flag (as I do for `consul agent`). 

### `consul version` for both Client and Server
Client: `1.0.6`
Server: `1.0.6`

gmreburn:
I also see this error. I am setting the `bind_addr` in the json config file but it fails to validate for this reason.

devopscloudnexus:
I am getting this error as well. Any fixes at the moment?

edganiukov:
@devopscloudnexus so far I see only one solution - use here
https://github.com/hashicorp/consul/blob/ecee15b4668e33762d36850b74e4d462bb39b917/command/validate/validate.go#L48
`b.Validate()` instead of `b.BuildAndValidate()`

But this is definitely a bug in design.

miaoxiaoy:
+1 on this error.  `bind_addr` in config.json not working. 
Consul v1.2.2

miaoxiaoy:
But set both `bind_addr` and `advertise_addr` is working ..

```
{
    "acl_agent_token": "xxxx",
    "acl_datacenter": "aliyun",
    "acl_default_policy": "deny",
    "acl_master_token": "xxxx",
    "bootstrap_expect": 2,
    "client_addr": "0.0.0.0",
    "bind_addr": "0.0.0.0",
    "advertise_addr": "192.168.1.200",
    "data_dir": "/opt/consul",
    "datacenter": "aliyun",
    "dns_config": {
        "allow_stale": true,
        "enable_truncate": true,
        "node_ttl": "60s",
        "service_ttl": {
            "*": "5s"
        }
    },
    "enable_script_checks": true,
    "encrypt": "xxxx",
[...]

PR #15107, opened 2022-10-21
fix: return error when config file with unknown extension is passed

nfi-hashicorp:
### Description

This should clear up a little confusion over error messages:

```bash
$ consul -version
Consul v1.13.3
Build Date 1970-01-01T00:00:01Z
Protocol 2 spoken by default, understands 2 to 3 (agent will automatically use protocol >2 when speaking to compatible agents)

$ cat > config.js # generate basic config
{"data_dir": "/tmp"}
```

#### Before

```bash
$ consul validate config.js # confusing error message: "I thought I set that?"
Config validation failed: data_dir cannot be empty
```

#### After

```bash
$ go run . validate config.js # after this PR: "ah, I need to set config format"
Config validation failed: file config.js has unknown extension; must be .hcl or .json, or config format must be set
exit status 1
$ go run . validate -config-format=json config.js
Configuration is valid!
```

Unlike in the config directory situation, where there might be unrelated files in the directory (e.g., README) that we should just skip, the user is directly specifying the file to load. If it is unloadable, that should be a fatal error.

Discovered in https://github.com/hashicorp/consul/issues/3897#issuecomment-1287368405

See above

### PR Checklist

* [x] ~~updated test coverage~~: not sure we can test this in any nice way besides checking strings
* [x] ~~external facing docs updated~~: N/A
* [x] not a security concern

### DIFF
--- a/agent/config/builder.go
+++ b/agent/config/builder.go
@@ -56,6 +56,10 @@ type LoadOpts struct {
 
 	// ConfigFiles is a slice of paths to config files and directories that will
 	// be loaded.
+	//
+	// It is an error for any config files to have an extension other than `hcl`
+	// or `json`, unless ConfigFormat is also set. However, non-HCL/JSON files in
+	// a config directory are merely skipped, with a warning.
 	ConfigFiles []string
 
 	// ConfigFormat forces all config files to be interpreted as this format
@@ -228,8 +232,7 @@ func (b *builder) sourcesFromPath(path string, format string) ([]Source, error) 
 
 	if !fi.IsDir() {
 		if !shouldParseFile(path, format) {
-			b.warn("skipping file %v, extension must be .hcl or .json, or config format must be set", path)
-			return nil, nil
+			return nil, fmt.Errorf("file %v has unknown extension; must be .hcl or .json, or config format must be set", path)
 		}
 
 		src, err := newSourceFromFile(path, format)
--- a/.changelog/15107.txt
+++ b/.changelog/15107.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cli: fatal error if config file does not have HCL or JSON extension, instead of warn and skip
+```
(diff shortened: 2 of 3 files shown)


agent/config/builder.go
```diff
@@ -223,8 +227,7 @@ func (b *builder) sourcesFromPath(path string, format string) ([]Source, error)
 
 	if !fi.IsDir() {
 		if !shouldParseFile(path, format) {
-			b.warn("skipping file %v, extension must be .hcl or .json, or config format must be set", path)
-			return nil, nil
+			return nil, fmt.Errorf("file %v has unknown extension; must be .hcl or .json, or config format must be set", path)
```
huikang, line 230: Will this change break the current behavior? Say, if user runs `consul agent -config-dir=/some-config-dir`. Under this `/some-config-dir` directory, there are some non-hcl/json file. Before this change, consul agent can proceed with the warning message, but this PR may stop process from being started.

nfi-hashicorp: This section is only for files, not directories. The directory logic starts at https://github.com/hashicorp/consul/blob/068cb2263ab52bb1947dce5e3d23d249e09317a6/agent/config/builder.go#L240, and remains the same (skip and warn).

But technically yes, it does break current behavior. I'm arguing current behavior is incorrect though: if I run `consul agent -config-file=config.badextension`, that should fail, not skip.

nfi-hashicorp:
Back in draft so I can improve the tests :)

agent/config/builder.go
huikang: Got it. It makes sense. Thanks.

nfi-hashicorp:
I cleaned up the tests a bit and codified the new behavior. Probably easiest to understand commit-by-commit.

huikang (commented):
It looks good to me. Could we have another consul engineer who stays on the project longer to take a look? Thanks.

nfi-hashicorp:
> It looks good to me. Could we have another consul engineer who stays on the project longer to take a look? Thanks.

Yep, I'll see if I can find someone. Thanks!

eculver (approved):
Looks great! Nice improvement.

PR #15114, opened 2022-10-24
Support consul connect envoy command on Windows

joselo85:
### Description
This PR introduces changes to enable the `consul connect envoy` command on Windows. To do this, we had to introduce two main changes:

First, we set a valid default value in the admin-access-log-path parameter of the Envoy bootstrap file for Windows. Now, the OS Platform where is Consul being executed is taking in account to set the default value for that parameter in that OS. This is during the Envoy bootstrap file creation of the consul connect envoy -bootstrap command.

Windows default value:  
`admin-access-log-path: "nul"`  
Linux/Mac (Unix OS) default value:  
`admin-access-log-path: "/dev/null" (current one without changes)`  
Also, the case: default-windows was added in TestGenerateConfig test to check this behavior

And last, we created the exec_windows.go file that execute Envoy using the `consul connect envoy command`


command/connect/envoy/envoy.go
```diff
@@ -22,13 +23,14 @@ import (
...
+func New(ui cli.Ui, osPlatform string) *cmd {
 	c := &cmd{UI: ui}
-	c.init()
+	c.init(osPlatform)
 	return c
 }
 
-const DefaultAdminAccessLogPath = "/dev/null"
+const DefaultUnixAdminAccessLogPath = "/dev/null"
+const DefaultWindowsAdminAccessLogPath = "nul"
```
dhiaayachi, line 33: The general approach that the Consul team took is to have OS specific values defined in OS specific files. I think if we create a windows and a non windows files with each defining different value of `DefaultAdminAccessLogPath` it would reduce the number of branches in the code below.

command/connect/envoy/exec_windows.go
```diff
@@ -0,0 +1,121 @@
+//go:build windows
```
dhiaayachi, line 1: I see some duplicate code between exec_windows.go and exec_unix.go, wondering if we can add an exec_supported file which will have the shared code between windows and unix?

dhiaayachi (changes requested):
Thank you for working on this @joselo85!
I think the implementation is good and have all the bits needed to make it work. I have some comments about the approach, specifically the use of OS specific values in the shared file and some code duplication.

command/connect/envoy/exec_windows.go
```diff
@@ -0,0 +1,121 @@
...
+	f.Sync()
+
+	// We can't wait for the process since we need to exec into Envoy before it
+	// will be able to complete so it will be remain as a zombie until Envoy is
+	// killed then will be reaped by the init process (pid 0). This is all a bit
+	// gross but the cleanest workaround I can think of for Envoy 1.10 not
+	// supporting /dev/fd/<fd> config paths any more. So we are done and leaving
+	// the child to run it's course without reaping it.
+	return tempFile, nil
+}
```
dhiaayachi, line 72: This could be a security concerns. The temp file in here will potentially have the agent token written into it. Is there a way in Windows to emulate an in-memory file? maybe something like https://learn.microsoft.com/en-us/dotnet/api/system.io.memorymappedfiles.memorymappedfile.createnew?view=net-7.0 for golang?

dhiaayachi (changes requested):
👋 @joselo85 
I added a comment about a possible security concern that we should address before this gets merged. I will try to find some time to spin an environment to test this this week.

david-yu:
Ping @ishustava looks like this needs some review for Windows support on VMs.

command/connect/envoy/exec_windows.go
absolutelightning: addressed in https://github.com/hashicorp/consul/pull/17694

david-yu:
Closing as the changes are now incorporated in https://github.com/hashicorp/consul/pull/17694 with commits from @joselo85

PR #15126, opened 2022-10-24
Move oss-only test to its own file

kisunji:
https://github.com/hashicorp/consul/pull/15066 added a test for OSS in a file misleadingly called `snapshot_oss_test.go` which actually runs on both oss and ent. 

This PR moves `snapshot_oss_test.go -> snapshot_test.go` and extracts the OSS-only test to `snapshot_oss_test.go`


mkeeler (approved):

DanStough (approved):
Awesome 🚀

PR #15148, opened 2022-10-25
UI: update hds usage to hds best practices

LevelbossMike:
### Description
This PR updates our usage of [HashiCorp Design System](https://github.com/hashicorp/design-system) to match the best practices that the HDS team was telling us about in a recent call.

The changes update the code according to the following best-practices:

1. Don't generate custom tailwind utility classes for colors
2. Use HDS color utilities instead of creating our own variants for colors when displaying text base on tailwind color utilities
3. Update flight icons usage to make use of `@color` instead of setting a text-color via Tailwind and using `fill-current`

This PR also updates the section of our engineering docs about HDS to better reflect the HDS best practices discussed.

* Visit peer-detail page and see that everything looks the same as it did before
* Look at engineering docs to see updated documentation around HDS

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/ui/packages/consul-ui/docs/hds.mdx
+++ b/ui/packages/consul-ui/docs/hds.mdx
@@ -4,50 +4,50 @@ title: Hashicorp Design System
 
 # Hashicorp Design System
 
-The application integrates setup that make it possible to use the [Hashicorp Design System (HDS)](https://github.com/hashicorp/design-system) in the application.
+The application integrates setup that make it possible to use the [Hashicorp Design System (HDS)](https://github.com/hashicorp/design-system) in the application. The application also
+bundles [TailwindCSS](https://tailwindcss.com/) to follow a utility-based CSS approach that mixes well with HDS utility classes. Tailwind is expected to be used
+as a supplement to HDS's utility classes, and should mainly be used for layouting.
 
 ## Design Tokens
 
-HDS ships a set of [design tokens](https://design-system-components-hashicorp.vercel.app/foundations/tokens), which are implemented as CSS custom properties.
-To make it easy to work with these design tokens without having to work with verbose CSS properties all over the place, we have setup a [TailwindCSS](https://tailwindcss.com/)-configuration that makes the color tokens
-available via functional [utility classes](https://tailwindcss.com/docs/customizing-colors).
-
 ### Colors
 
-You can work with any color from the HDS by prefixing colors with the `hds-`-prefix. The `hds-`-prefix makes it easy to see what classes are auto-generated from HDS. When we wanted to color a header tag with the `consul-brand`-color we could do it like this for example:
+To make sure we follow HDS-guidelines, we disabled Tailwind's color utilities completely. Instead of relying on Tailwind for colors, you should be working with the colors from HashiCorp-Design-System and follow the [guidance in the docs](https://design-system-components-hashicorp.vercel.app/foundations/colors) on how to use them:
 
 ```hbs preview-template
-<h2 class="text-hds-consul-brand">
+<h2 class='hds-foreground-strong'>
   HDS is awesome
 </h2>
 ```
 
-Regular tailwindCss colors aren't available. You can only use colors defined in HDS with this pattern.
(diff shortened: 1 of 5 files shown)


ui/packages/consul-ui/docs/hds.mdx
```diff
@@ -57,44 +57,40 @@ Because we are working with a customized tailwind configuration it is easy to co
 All components from Hashicorp Design System are available for you to use. Here's an example that shows how to implement a navigation bar with HDS and Tailwind in combination.
 
 ```hbs preview-template
-<nav class="h-16 w-full bg-hds-foreground-strong flex items-center justify-between px-4 hds-font-weight-medium">
-  <ul class="flex items-center">
+<nav
+  class='h-16 w-full bg-[color:var(--token-color-palette-neutral-700)] flex items-center justify-between px-4 hds-font-weight-medium'
```
LevelbossMike, line 61: We probably want to switch to a class name for the background here. I wonder if we should be using a custom utility class here, and how to name said utility class, as no dark surface color with a `hds-`-prefix exists yet: 

We could leave out the prefix altogether
```css
.surface-nav {
  background: var(--token-color-palette-neutral-700);
}
```
or create our own prefix to indicate that we tried using HDS but couldn't as a utility that does what we want didn't exist yet:

```css
.consul-surface-nav {
  background: var(--token-color-palette-neutral-700);
}
```

ui/packages/consul-ui/docs/hds.mdx
```diff
@@ -4,50 +4,50 @@ title: Hashicorp Design System
...
 
 ```hbs preview-template
-<h2 class="text-red-400">
+<h2 class='text-red-400'>
   TailwindCSS colors won't work
 </h2>
 ```
 
-### Other tokens
+### HDS tokens
```
didoo, line 31: I am worried this may lead to confusion. [In HDS we call `tokens` the key/value pairs in code](https://design-system-components-hashicorp.vercel.app/foundations/tokens) (generally, CSS custom props). What you're describing here are probably more similar to the the "foundation" (or "core") styles. They include colors, typography, elevation, and focus ring, and they all can be expressed via CSS helper/utility classes.

ui/packages/consul-ui/docs/hds.mdx
didoo: @LevelbossMike I haven't looked at the Consul UI codebase, but I would expect that there is only one instance of this UI element (the page header) in it, and it's expressed via a component (eg. `Header`) or directly in the page template. This means (under my assumption) that probably defining a utility class only for that single instance of background color is a bit of an overkill. Probably it would be (in principle) more "correct" to define a local class (eg. `.consul-header`), assign the background color as suggested above (`background: var(--token-color-palette-neutral-700);`) and then apply that class to the `<nav>` element (the one suggested above).

My suggestion is (since this is not actual code, but a README) to change the example, or use it but assign the "local" class name and explain how and why you have used it this way (consider it an example of what _not_ to do :) ). Unless the approach is to use only tailwind + hds classes in the handlebars templates, and never create specific local/custom class names for the Consul UI components/elements, in which case... it's going to be challenging and probably you have to rely on tailwind helpers like `bg-[color:var(--token-color-palette-neutral-700)]`.

didoo (commented):
I have added a couple of comments (non blocker per se, but give them some thought).
Happy to jump on a call, if you want to discuss them more in depth.

ui/packages/consul-ui/docs/hds.mdx
LevelbossMike: Thanks for the input! I discussed this with @WenInCode and for consistency’s sake, we'll take the custom utilty approach here. With Tailwind, this reads nicer, and we will only have one place where this custom surface class would be defined. 

Example:

```hbs
<nav class="flex items-center space-x-2 consul-surface-brand">
```
vs.

```hbs
<nav class="flex items-center space-x-2 consul-header" />
```

WenInCode (approved):

PR #15153, opened 2022-10-25
ui: update icons to use flight icons component

WenInCode:
### Description
**In Progress**

Started out looking into replacing the aws icon, but decided to take a crack at replacing all the external source icons with `<FlightIcon>` (from HDS) where possible. 

Screen Shot 2022-10-25 at 2 15 36 PM
Screen Shot 2022-10-25 at 2 15 11 PM
Screen Shot 2022-10-25 at 2 14 46 PM

### Links
[NET-898](https://hashicorp.atlassian.net/browse/NET-898)

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


ui/packages/consul-ui/app/components/custom-element/README.mdx
```diff
@@ -1,87 +0,0 @@
-# CustomElement
```
WenInCode, line 1: I saw the shadow template component was using an icon I removed in the documentation, then I checked where it and it's related components (shadow-host/custom-element/disclosure-card/etc.) were being used and saw that they weren't being used at all. I decided to remove them all. If we really want them back we can reintroduce them through git when they need to be used.

snap

LevelbossMike (approved):
Screenshot 2022-10-27 at 19 35 45

👏 

This is wonderful! Very excited to see this land! There's a "consul logo" component that we might also want to switch over to the flight-icon version?! This is the logo that is displayed in the header.

PR #15155, opened 2022-10-25
Allow consul debug on non-ACL consul servers

kisunji:
### Description
https://github.com/hashicorp/consul/pull/10273 had an unintended side-effect of crashing `consul debug` CLI command if ACLs were disabled and `debug_enabled` was false.

The correct behavior would be to disable the HTTP endpoints for profiling information but still allowing other debug profile targets like `logs` and `metrics` to run without issues.

* Modified tests to assert for previous behavior

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/http.go
+++ b/agent/http.go
@@ -209,13 +209,6 @@ func (s *HTTPHandlers) handler(enableDebug bool) http.Handler {
 			var token string
 			s.parseToken(req, &token)
 
-			// If enableDebug is not set, and ACLs are disabled, write
-			// an unauthorized response
-			if !enableDebug && s.checkACLDisabled() {
-				resp.WriteHeader(http.StatusUnauthorized)
-				return
-			}
-
 			authz, err := s.agent.delegate.ResolveTokenAndDefaultMeta(token, nil, nil)
 			if err != nil {
 				resp.WriteHeader(http.StatusForbidden)
@@ -247,12 +240,14 @@ func (s *HTTPHandlers) handler(enableDebug bool) http.Handler {
 		handleFuncMetrics(pattern, s.wrap(bound, methods))
 	}
 
-	// Register wrapped pprof handlers
-	handlePProf("/debug/pprof/", pprof.Index)
-	handlePProf("/debug/pprof/cmdline", pprof.Cmdline)
-	handlePProf("/debug/pprof/profile", pprof.Profile)
-	handlePProf("/debug/pprof/symbol", pprof.Symbol)
-	handlePProf("/debug/pprof/trace", pprof.Trace)
+	// If enableDebug or ACL enabled, register wrapped pprof handlers
+	if enableDebug || !s.checkACLDisabled() {
+		handlePProf("/debug/pprof/", pprof.Index)
+		handlePProf("/debug/pprof/cmdline", pprof.Cmdline)
+		handlePProf("/debug/pprof/profile", pprof.Profile)
+		handlePProf("/debug/pprof/symbol", pprof.Symbol)
+		handlePProf("/debug/pprof/trace", pprof.Trace)
+	}
 
 	if s.IsUIEnabled() {
 		// Note that we _don't_ support reloading ui_config.{enabled, content_dir,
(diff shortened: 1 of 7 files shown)


agent/http.go
```diff
@@ -247,12 +240,14 @@ func (s *HTTPHandlers) handler(enableDebug bool) http.Handler {
...
-	handlePProf("/debug/pprof/symbol", pprof.Symbol)
-	handlePProf("/debug/pprof/trace", pprof.Trace)
+	// If enableDebug or ACL enabled, register wrapped pprof handlers
+	if enableDebug || !s.checkACLDisabled() {
+		handlePProf("/debug/pprof/", pprof.Index)
+		handlePProf("/debug/pprof/cmdline", pprof.Cmdline)
+		handlePProf("/debug/pprof/profile", pprof.Profile)
+		handlePProf("/debug/pprof/symbol", pprof.Symbol)
+		handlePProf("/debug/pprof/trace", pprof.Trace)
+	}
```
kisunji, line 250: I thought it would make more sense for the handler registration to be guarded by the if condition here rather than allowing the endpoints to be exposed but returning a 401 every time.

Would this cause any side effects?

agent/http.go
```diff
@@ -209,13 +209,6 @@ func (s *HTTPHandlers) handler(enableDebug bool) http.Handler {
 			var token string
 			s.parseToken(req, &token)
 
-			// If enableDebug is not set, and ACLs are disabled, write
-			// an unauthorized response
-			if !enableDebug && s.checkACLDisabled() {
```
kisunji, line 214: Instead of checking these conditions per request, I've moved the check below to conditionally register pprof handlers instead.

command/debug/debug.go
```diff
@@ -374,11 +389,12 @@ func (c *cmd) captureInterval(ctx context.Context) error {
 func captureShortLived(c *cmd) error {
 	g := new(errgroup.Group)
 
-	dir, err := makeIntervalDir(c.output, c.timeNow())
-	if err != nil {
-		return err
-	}
 	if c.captureTarget(targetProfiles) {
+		dir, err := makeIntervalDir(c.output, c.timeNow())
```
kisunji, line 393: avoids making empty directories if this branch is never entered

command/debug/debug.go
```diff
@@ -274,6 +273,22 @@ func (c *cmd) prepare() (version string, err error) {
 		c.capture = defaultTargets
 	}
 
+	// If EnableDebug is not true, skip collecting pprof
+	enableDebug, ok := self["DebugConfig"]["EnableDebug"].(bool)
```
kisunji, line 277: Reintroduces lines removed in https://github.com/hashicorp/consul/pull/10273/files#diff-e4c25813d9467f20b9374e5dd16394ccf77aaf1db8fee3d1c7bba52b716b3cdeL257-L278

agent/connect/ca/testing.go
```diff
@@ -194,13 +194,13 @@ func (v *TestVaultServer) WaitUntilReady(t testing.T) {
...
-			r.Fatalf("err: %v", err)
+			r.Errorf("err: %v", err)
 		}
 		if !resp.Initialized {
-			r.Fatalf("vault server is not initialized")
+			r.Error("vault server is not initialized")
 		}
 		if resp.Sealed {
-			r.Fatalf("vault server is sealed")
+			r.Error("vault server is sealed")
```
kisunji, line 203: I noticed this test flaking often and believe the retry is never run because `Fatalf` panics and kills the loop

mkeeler: > believe the retry is never run because Fatalf panics and kills the loop

If we were calling `t.Fatalf` then that would be the case. `retry.Run` however will recover from the panic induced by the `r.Fatalf` call and continue retrying until it thinks it should stop (after 7s in this case)

kisunji: Ah I see! Dropped this commit

dhiaayachi (approved):
LGTM!

PR #15173, opened 2022-10-27
Fix peered service protocols using proxy-defaults.

hashi-derek:
This change prevents `service-defaults` and `proxy-defaults` from overwriting the protocol on services imported through peering. Without this change, the protocol could mismatch what was expected on the remote mesh gateway.


erichaberkorn (commented):
@hashi-derek Could you add a test for this? Nice job figuring this out.

erichaberkorn (approved):

PR #15185, opened 2022-10-28
Update hcp-scada-provider to fix diamond dependency problem with go-msgpack

kisunji:


go.mod
```diff
@@ -19,11 +19,11 @@ require (
 	github.com/coredns/coredns v1.6.6
 	github.com/coreos/go-oidc v2.1.0+incompatible
 	github.com/docker/go-connections v0.3.0
-	github.com/envoyproxy/go-control-plane v0.10.1
+	github.com/envoyproxy/go-control-plane v0.10.2-0.20220325020618-49ff273808a1
 	github.com/fsnotify/fsnotify v1.5.1
 	github.com/go-openapi/runtime v0.24.1
 	github.com/go-openapi/strfmt v0.21.3
-	github.com/golang/protobuf v1.5.0
+	github.com/golang/protobuf v1.5.2
```
rboyer, line 26: This feels unexpected, too.

go.mod
```diff
@@ -19,11 +19,11 @@ require (
 	github.com/coredns/coredns v1.6.6
 	github.com/coreos/go-oidc v2.1.0+incompatible
 	github.com/docker/go-connections v0.3.0
-	github.com/envoyproxy/go-control-plane v0.10.1
+	github.com/envoyproxy/go-control-plane v0.10.2-0.20220325020618-49ff273808a1
```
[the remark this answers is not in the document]
kisunji: @erichaberkorn do you know if this will be okay? A bunch of these are side effects from updating scada provider

agent/grpc-internal/client.go
```diff
@@ -132,8 +133,6 @@ func (c *ClientConnPool) dial(datacenter string, serverType string) (*grpc.Clien
 		grpc.WithContextDialer(c.dialer),
 		grpc.WithDisableRetry(),
 		grpc.WithStatsHandler(agentmiddleware.NewStatsHandler(metrics.Default(), metricsLabels)),
-		// nolint:staticcheck // there is no other supported alternative to WithBalancerName
-		grpc.WithBalancerName("pick_first"),
```
kisunji, line 136: This method is gone from the API but "pick_first" was the default balancer anyways so this should have no impact.

rboyer, line 136: How are you going to turn on the "pick the first server every time" behavior now?

kisunji: I believe "pick_first" is the default balancer.

If I'm wrong, then this should be the new way to define balancer config: https://github.com/weaveworks/common/issues/239#issuecomment-1115920318

rboyer:
Does this need a 1.14 backport?

go.mod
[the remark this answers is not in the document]
erichaberkorn: It theoretically should be fine 😂 I'm more concerned about the protobuf upgrade.

rboyer (approved):

PR #15186, opened 2022-10-28
Fix mesh gateway configuration with proxy-defaults

hashi-derek:
Prior to this PR, proxy-default configurations for mesh gateways were not properly inherited on upstreams.


agent/configentry/resolve.go
```diff
@@ -161,9 +162,12 @@ func ComputeResolvedServiceConfig(
 			// upstreams that are inferred from intentions and do not have explicit upstream configuration.
 			cfgMap := make(map[string]interface{})
 			upstreamDefaults.MergeInto(cfgMap)
+			if !args.MeshGateway.IsZero() {
+				cfgMap["mesh_gateway"] = args.MeshGateway
```
hashi-derek, line 166: This is weird. If we leave the old logic in-place, then `NodeService.Proxy.MeshGateway` will NOT override the value from `ServiceDefaults.UpstreamConfig.Defaults` for the wildcard upstream.

This means that a user of tproxy could encounter a situation where they configure `NodeService.Proxy.MeshGateway` with a value that is ignored for all implicit upstreams, but honored for explicit ones.

agent/configentry/resolve.go
```diff
@@ -190,49 +193,52 @@ func ComputeResolvedServiceConfig(
...
-		// 	0. Value from centralized upstream_defaults
-		// 	1. Value from local proxy registration
-		// 	2. Value from centralized upstream_config
-		// 	3. Value from local upstream definition. This last step is done in the client's service manager.
+		// The goal is to flatten the mesh gateway mode in this order (larger number wins):
+		//  1. Value from the proxy-defaults
+		//  2. Value from centralized upstream defaults (ServiceDefaults.UpstreamConfig.Defaults)
+		//  3. Value from local proxy registration (NodeService.Proxy.MeshGateway)
+		//  4. Value from centralized upstream override (ServiceDefaults.UpstreamConfig.Overrides)
+		//  5. Value from local upstream definition (NodeProxy.Proxy.Upstreams[].MeshGateway)
```
hashi-derek, line 204:
```suggestion
		//  5. Value from local upstream definition (NodeService.Proxy.Upstreams[].MeshGateway)
```

erichaberkorn (approved):

PR #15188, opened 2022-10-28
build: ensure 'make dev-docker' also produces a matching CLI binary at the same time

rboyer:
### Description

In many local setups you may run `make dev-docker` with the intent to run that consul image on a local docker install. But frequently you will also want to interact with that running install using the CLI running on your host machine.

This ensures that `make dev-docker` will also produce a dev binary (`make dev`) at the same time so that they can match.


DanStough (approved):
🚀 🚀 🚀

PR #15195, opened 2022-10-28
test: fix flaky TestSubscribeBackend_IntegrationWithServer_DeliversAllMessages test

rboyer:
### Description

Allow for some message duplication in subscription events during assertions.

I'm pretty sure the subscriptions machinery allows for messages to occasionally be duplicated instead of dropping them, as a once-and-only-once queue is a pipe dream and you have to pick one of the other two options.

### DIFF
--- a/agent/consul/subscribe_backend_test.go
+++ b/agent/consul/subscribe_backend_test.go
@@ -444,13 +444,18 @@ func verifyMonotonicStreamUpdates(ctx context.Context, logger testLogger, client
 			if err != nil {
 				return err
 			}
-			if expectPort != svc.Port {
+			switch svc.Port {
+			case expectPort:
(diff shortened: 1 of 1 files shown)


kisunji (approved):
Is the repeated port guaranteed to lag by 1? Or is this more of a "allow this much wiggle room"?

rboyer:
> Is the repeated port guaranteed to lag by 1? Or is this more of a "allow this much wiggle room"?
In the test we generate a series of service registrations that will monotonically increase the port number every time. My guess is that there are a few scenarios where there's like a radix tree rebalance or some other no-op that happens in the middle of the test that makes the subscription event emitter re-emit the most recent event, which I why I made the test be ok with getting some duplicate serial events.

PR #15214, opened 2022-10-31
Prevent peering acceptor from subscribing to addr updates.

hashi-derek:
When peering two clusters, the acceptor will always display the following warning:

```
[WARN]  agent.grpc-api.peerstream.stream: client peer was unable to apply resource: dialed=false peer_id=57f283a6-7c25-5125-fe1b-bdb11150f2ce peer_name=cluster1 code=13 error="upsert error, ResourceURL: \"type.googleapis.com/hashicorp.consul.internal.peering.PeeringServerAddresses\", ResourceID: \"server-addrs\": Cannot switch peering dialing mode from false to true"
```

This occurs because the acceptor is attempting to save a new list of server addrs, which is not possible (a server addr list length greater than zero indicates that the peer is a dialer and not an acceptor).

### DIFF
--- a/agent/grpc-external/services/peerstream/stream_resources.go
+++ b/agent/grpc-external/services/peerstream/stream_resources.go
@@ -277,7 +277,7 @@ type HandleStreamRequest struct {
 	Stream BidirectionalStream
 }
 
-func (r HandleStreamRequest) WasDialed() bool {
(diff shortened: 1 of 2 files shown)


DanStough (approved):
LGTM

PR #15218, opened 2022-11-01
Regenerate certificates.

hashi-derek:
Create new certificates for the CICD pipeline tests that were failing.


DanStough (approved):

DanStough:
This would be backported to 1.13 and 1.12 as well, correct?

hashi-derek:
> This would be backported to 1.13 and 1.12 as well, correct?

I'm not sure about 1.12, but I think it would probably be fine for 1.13. If the backport didn't automatically merge, then I would have added the labels. I can cherry pick this and see if it passes for the 1.12

PR #15225, opened 2022-11-01
fix: persist peering CA updates to dialing clusters

DanStough:
### Description
During peering stream updates, accepting clusters send updates to dialers about TrustBundle changes, but these changes are not filtered back into the peering object. If the connection is interrupted after a CA rotation, the dialer will not have the correct CA info to dial the acceptor. 

This PR also updates the Peering records when a trust bundle is written to the state store.

* Unit Tests
* Run the integration test `rotate_server_and_ca_then_fail_test.go`

### PR Checklist

* [X] updated test coverage


DanStough:
Superceded by #15243

PR #15233, opened 2022-11-02
integ test: reduce flakiness due to compound output from retry

huikang:
### Description

The output of retry in bats test contains the failed message (see below example), a condition that will cause the step to fail because the `status=0` while output contains extraneous content.

```
not ok 14 s1 upstream should be able to connect to s2
# (in test file /workdir/primary/bats/verify.bats, line 62)
#   `[ "$output" = "hello" ]' failed
# status = 0
# output = Command failed. Attempt 1/5:
# hello
```

This PR fixes the flakiness by using `string contain` test condition.

```
 - [ "$output" = "hello" ]
 + [[ "$output" == *"hello"* ]]
```

- auto-format done by the editor

### Links

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/.changelog/15233.txt
+++ b/.changelog/15233.txt
@@ -1,0 +1,3 @@
+```release-note: improvement
+integ test: fix flakiness due to test condition from retry app endoint
+```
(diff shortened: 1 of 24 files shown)


eculver (approved):
Looks straightforward to me. Thanks for the thorough explanation in the description.

eculver:
I think if you rebase on latest main, the tests should get straightened out.

huikang:
> I think if you rebase on latest main, the tests should get straightened out.

@eculver , thanks for the feedback; PR is rebased.

PR #15239, opened 2022-11-02
Added check for empty peeringsni in restrictPeeringEndpoints

sarahalsmiller:
### Description
restrictPeeringEndpoints introduced some failures in consul-api-gateway. This check will prevent the middleware from erroring when no peeringsni is set

- consul-api-gateway tests pass when using an image built from this branch

### Links
* https://github.com/hashicorp/consul-api-gateway/actions/workflows/conformance.yml Pipeline failures in consul-api-gateway

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/agent/grpc-middleware/auth_interceptor.go
+++ b/agent/grpc-middleware/auth_interceptor.go
@@ -66,7 +66,7 @@ func (a *AuthInterceptor) InterceptStream(
 // present a mutual TLS certificate, and is allowed to bypass the `tls.grpc.verify_incoming`
 // check as a special case. See the `tlsutil.Configurator` for this bypass.
 func restrictPeeringEndpoints(authInfo credentials.AuthInfo, peeringSNI string, endpoint string) error {
-	//No peering connection has been configured
+	// No peering connection has been configured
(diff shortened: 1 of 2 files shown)


agent/grpc-middleware/auth_interceptor.go
```diff
@@ -75,6 +79,7 @@ func restrictPeeringEndpoints(authInfo credentials.AuthInfo, peeringSNI string,
 	if !ok {
 		return status.Error(codes.Unauthenticated, "invalid transport credentials")
 	}
+
 	if tlsAuth.State.ServerName == peeringSNI {
```
hashi-derek, line 83: I think we should probably move the check to down here, but otherwise this seems okay.

```
if peeringSNI != "" && peeringSNI == tlsAuth.State.ServerName
```

hashi-derek: Or just move it as-is further down.

```
if peeringSNI == "" {
  return nil
}
if peeringSNI == tlsAuth.State.ServerName {
```

PR #15253, opened 2022-11-03
Detect Vault 1.11+ import, update default issuer

cipherboy:
`Signed-off-by: Alexander Scheel <alex.scheel@hashicorp.com>`

### Description
Background: https://support.hashicorp.com/hc/en-us/articles/11308460105491

Consul used to rely on implicit issuer selection when calling Vault endpoints to issue new CSRs. Vault 1.11+ changed that behavior, which caused Consul to check the wrong (previous) issuer when renewing its Intermediate CA. This patch allows Consul to explicitly set a default issuer when it detects that the response from Vault is 1.11+.

* Tested manually in K8s setup and local unit tests using different versions of Vault

### Links
https://support.hashicorp.com/hc/en-us/articles/11308460105491

Vault's PR to add multiple issuer support in PKI: https://github.com/hashicorp/vault/pull/15277

Vault's PR to add flag to opt-in to previous behavior: https://github.com/hashicorp/vault/pull/17824

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/connect/ca/provider_vault.go
```diff
@@ -551,13 +551,39 @@ func (v *VaultProvider) GenerateIntermediate() (string, error) {
...
 	// Set the intermediate backend to use the new certificate.
-	_, err = v.writeNamespaced(v.config.IntermediatePKINamespace, v.config.IntermediatePKIPath+"intermediate/set-signed", map[string]interface{}{
+	importResp, err := v.writeNamespaced(v.config.IntermediatePKINamespace, v.config.IntermediatePKIPath+"intermediate/set-signed", map[string]interface{}{
 		"certificate": intermediate.Data["certificate"],
 	})
 	if err != nil {
 		return "", err
 	}
 
+	if importResp != nil {
```
kisunji, line 561: Can Vault <1.11 also return a non-nil response? Wondering if we need to guard against POST-ing to `config/issuers` with empty string.

agent/connect/ca/provider_vault.go
```diff
@@ -551,13 +551,39 @@ func (v *VaultProvider) GenerateIntermediate() (string, error) {
...
+		mapping := importResp.Data["mapping"].(map[string]string)
+		var intermediateId string
+		for issuer, key := range mapping {
+			if key != "" {
+				intermediateId = issuer
+
+				// Could be safe and check:
+				// importedIssuers := importResp.Data["imported_issuers"].([]string)
+				// and make sure this newly imported issuer is in the set of new unique
+				// issuers.
```
kisunji, line 573: I think this is unnecessary since Consul will always generate a new UUID for intermediate CSR. Correct me if I'm wrong

agent/connect/ca/provider_vault.go
cipherboy: Only [I believe](https://github.com/hashicorp/vault/blob/release/1.10.x/builtin/logical/pki/path_intermediate.go#L244) in [error cases](https://github.com/hashicorp/vault/blob/release/1.10.x/builtin/logical/pki/path_config_ca.go#L91). I'm not sure what `v.writeNamespaced` is, but in Vault's API client, we always return `(*vault.api.Secret, error)` -- the former cannot contain an error response. If you're directly using a lower-level API (one giving a raw response akin to `(vault.api.Response, error)`), you might end up with an error in the response object (e.g., if `logical.ErrorResponse` [is used on the server](https://github.com/hashicorp/vault/blob/release/1.10.x/builtin/logical/pki/path_intermediate.go#L204)), so you'd have to translate that to an `err` first.


_Assuming_ that has been done, and _assuming_ this is a Vault `*api.Secret` object, then the answer is no, and we strictly need to deal with non-nil responses here from `set-signed` on Vault 1.11+.

agent/connect/ca/provider_vault.go
```diff
@@ -551,13 +551,39 @@ func (v *VaultProvider) GenerateIntermediate() (string, error) {
...
+	importResp, err := v.writeNamespaced(v.config.IntermediatePKINamespace, v.config.IntermediatePKIPath+"intermediate/set-signed", map[string]interface{}{
 		"certificate": intermediate.Data["certificate"],
 	})
 	if err != nil {
 		return "", err
 	}
 
+	if importResp != nil {
+		// Assume we're on Vault 1.11+. We'll assume we only have one issuer
+		// with a key.
```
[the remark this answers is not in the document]
cipherboy: This was mostly added during the initial session with @jkirschner-hashicorp to make sure we codified the changes I was suggesting to this endpoint to make them more clear than meeting minutes.

I'm more than happy for you to take this branch and run with it as you see fit. :-) 

---

As a nit though, note that this is not strictly correct:

> no longer overwrites the default issuer when it imports certificates

_Technically_ on Vault <1.11, the overwrite occurred when _creating_ the new key material earlier (`POST /intermediate/generate/internal`).

At this point in time, while the old intermediate CA was still returned (from `GET /cert/ca`), you'd lack the old key material and _all signing requests would fail_.

This is why the multi-issuer functionality (originally called safer CA rotation) was so necessary: any new root or intermediate generation operation on the mount silently destroyed existing key material, making using the mount not viable until the new issuer was generated.

If this process ever failed for Consul (e.g., root expired) -- you'd break all intermediate issuance at your 50% mark, before the intermediate itself expired.

agent/connect/ca/provider_vault.go
cipherboy: Hmm, no this is a Vault generated/returned UUID.

Vault's new import APIs (in 1.11, they've all been unified -- so `/config/ca` and `/intermediate/set-signed` all function the same) return the following data structure:

```
imported_issuers: []string
imported_keys: []string
mapping: dict(issuer->key)
```

So if you post say:

```
Cert A
Key for Cert A
Cert B
Key for Cert B
```

and nothing exists, you'd end up with:

```
imported_issuers: A, B
imported_keys: A, B
mapping:
  A->A
  B->B
```

But if you import:

```
Cert A
Cert B
Key B
```

and the key for A already exists in the mount, you'd get:

```
imported_issuers: A, B
imported_keys: B
mapping:
  A->A
  B->B
```

i.e, the key wasn't present on import, but was present on the mount. 

---

^^ This is what we want to guard against, with the above comment. It shouldn't happen, but it is good for data safety.

In particular, if Consul posts `{Root Cert, Intermediate Cert}` and someone posted the Root Cert's key into this mount, your mapping would look like the above (minus the `imported_keys` since you didn't import a key -- it was already internal on generation).

So, we need to ensure that only the new issuer gets selected as default, and not the root cert. However, if the root cert was net new (someone previously imported its key, but didn't import the root itself), this method would fail. It seems unlikely to me, tbh.


Alternatively, on the `POST /intermediate/generate/internal` call, you could save the `key_id` response and ensure that this newly imported/selected issuer is the one matching that existing `key_id`. This is safer and what I'd prefer if you don't like the above.

---

Similarly, if you do:

```
Key A
Cert B
Key for B
```

you'd get:

```
imported_issuers: B
imported_keys: A, B
mapping:
  B->B
```

because, while the key was imported and new, you wouldn't see the cert in the mapping.

What we want to guard against.

cipherboy:
@kisunji Hope that helps, happy to answer questions if you've got some more. :-)

agent/connect/ca/provider_vault.go
```diff
@@ -551,16 +563,71 @@ func (v *VaultProvider) GenerateIntermediate() (string, error) {
...
+// [/intermediate/set-signed]: https://developer.hashicorp.com/vault/api-docs/secret/pki#import-ca-certificates-and-keys
+// [/intermediate/generate/internal]: https://developer.hashicorp.com/vault/api-docs/secret/pki#generate-intermediate-csr
+func (v *VaultProvider) setDefaultIntermediateIssuer(vaultResp *vaultapi.Secret, keyId string) error {
+	if vaultResp.Data["mapping"] == nil {
+		return fmt.Errorf("expected Vault response data to have a 'mapping' key")
+	}
+	if keyId == "" {
+		return fmt.Errorf("expected non-empty keyId")
+	}
+	mapping, ok := vaultResp.Data["mapping"].(map[string]string)
```
cipherboy, line 602: [Since you're calling](https://github.com/hashicorp/consul/blob/359a908bea16ef6e10f70295045bb90269a305b7/agent/connect/ca/provider_vault.go#L766-L770) `client.Logical().Write(...)` here, this will panic: the API will respond with a `map[string]interface{}` due to JSON parsing I believe.

As you loop, you'll probably have to type assert `key` to `.(string)`.

agent/connect/ca/provider_vault.go
```diff
@@ -551,16 +563,71 @@ func (v *VaultProvider) GenerateIntermediate() (string, error) {
...
+			break
+		}
+	}
+	if intermediateId == "" {
+		return fmt.Errorf("could not find key_id %q in response from vault", keyId)
+	}
+
+	// Post it as the default issuer.
+	_, err := v.writeNamespaced(v.config.IntermediatePKINamespace, v.config.IntermediatePKIPath+"config/issuers", map[string]interface{}{
+		"default": intermediateId,
```
cipherboy, line 622: So the one concern here is that this is a POST, overwriting all current values. I think we'd have to read the value of the existing config, and set this back with `default` and `default_follows_latest_issuer` (from its existing value, if any).

Unknown parameters are warned but not err'd on, so you are safe to send it (defaulting to `false`) if you don't get it on the read (i.e., earlier Vault version without the new feature). 

I want to make this a PATCH API eventually as well, but note that obviously it wouldn't exist on old versions, so you're probably stuck doing the GET+POST anyways.

agent/connect/ca/provider_vault.go
```diff
@@ -551,16 +563,71 @@ func (v *VaultProvider) GenerateIntermediate() (string, error) {
...
+// multiple issuers, no longer overwrites the default issuer when
+// generateIntermediateCSR (intermediate/generate/internal) is called.
+//
+// The response we get from calling [/intermediate/set-signed]
+// should contain a "mapping" data field we can use to cross-reference
+// with the keyId returned when calling [/intermediate/generate/internal].
+//
+// [/intermediate/set-signed]: https://developer.hashicorp.com/vault/api-docs/secret/pki#import-ca-certificates-and-keys
+// [/intermediate/generate/internal]: https://developer.hashicorp.com/vault/api-docs/secret/pki#generate-intermediate-csr
+func (v *VaultProvider) setDefaultIntermediateIssuer(vaultResp *vaultapi.Secret, keyId string) error {
```
cipherboy, line 595: Looking good, like this refactor!

agent/connect/ca/provider_vault.go
kisunji: Good catch, Kyle encountered this during testing as well

agent/connect/ca/provider_vault.go
kisunji: Ah so the concern here is clobbering `default_follows_latest_issuer` if it was true, right? Will add a GET+POST. Thanks for the heads up

cipherboy: Yep!

agent/connect/ca/provider_vault.go
```diff
@@ -551,16 +563,81 @@ func (v *VaultProvider) GenerateIntermediate() (string, error) {
...
+
+	mapping, ok := vaultResp.Data["mapping"].(map[string]any)
+	if !ok {
+		return fmt.Errorf("unexpected type for 'mapping' value in Vault response")
+	}
+
+	var intermediateId string
+	// The value in this KV pair is called "key"
+	for issuer, key := range mapping {
+		if key == keyId {
```
cipherboy, line 611: I think we'd need the type assertion here to `key.(string)`?

cipherboy: Ah interesting, I guess not, never mind!

cipherboy (commented):
I can't really ACK it since its my PR, but the changes look good from a Vault perspective, thank you!

agent/connect/ca/provider_vault_test.go
```diff
@@ -902,6 +902,29 @@ func TestVaultProvider_ReconfigureIntermediateTTL(t *testing.T) {
 	require.Equal(t, 333*3600, mountConfig.MaxLeaseTTL)
 }
 
+func TestVaultCAProvider_GenerateIntermediate(t *testing.T) {
```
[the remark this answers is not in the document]
jkirschner-hashicorp: Thank you for including this detail :)

agent/connect/ca/provider_vault_test.go
```diff
@@ -902,6 +902,28 @@ func TestVaultProvider_ReconfigureIntermediateTTL(t *testing.T) {
...
+
+	SkipIfVaultNotPresent(t)
+
+	provider, _ := testVaultProviderWithConfig(t, true, nil)
+
+	orig, err := provider.ActiveIntermediate()
+	require.NoError(t, err)
+
```
kschoche, line 913: Wow you weren't joking this did turn out to be easier than we thought! Nice work!

kisunji (approved):

kschoche (approved):

kisunji:
Backports:
1.12.x: https://github.com/hashicorp/consul/pull/15432
1.13.x: https://github.com/hashicorp/consul/pull/15436
1.14.x: https://github.com/hashicorp/consul/pull/15437

PR #15254, opened 2022-11-03
Backport various fixes from ENT.

hashi-derek:


eculver (approved):
Looks good. I am primarily looking at the original PRs from which these are derived.

PR #15260, opened 2022-11-04
Backport tests from ent.

hashi-derek:


agent/grpc-external/services/peerstream/stream_test.go
```diff
@@ -1204,6 +1204,9 @@ func TestStreamResources_Server_AckNackNonce(t *testing.T) {
 		require.NoError(t, err)
 		require.Equal(t, "5678", msg.GetRequest().ResponseNonce)
 	})
+	// Add in a sleep to prevent the test from flaking.
+	// The mock client expects certain calls to be made.
+	time.Sleep(50 * time.Millisecond)
```
hashi-derek, line 1209: Adding in this 50ms sleep allowed the test to run with `-count=1000` without failing at all. Without the sleep, it would fail every ~30 calls.

erichaberkorn (approved):

PR #15267, opened 2022-11-04
Bring back parameter ServerExternalAddresses in GenerateToken endpoint

hashi-derek:
This PR finishes up work started in: https://github.com/hashicorp/consul/pull/15215

Revert "Remove ability to specify external addresses in GenerateToken endpoint (https://github.com/hashicorp/consul/pull/14930)"
This reverts commit https://github.com/hashicorp/consul/commit/5e156772f6a7fba5324eb6804ae4e93c091229a6.

Originally ServerExternalAddresses were added to support peering across networks in k8s before peering over mesh gateways for the control plane was supported. Support for that parameter was reverted since it was no longer needed for k8s (where mesh gateways are now the required approach for control plane traffic for peering), but bringing it back because it's useful for managed deployments where deploying mesh gateways is currently out of scope.


api/go.mod
```diff
@@ -5,10 +5,15 @@ go 1.12
 replace github.com/hashicorp/consul/sdk => ../sdk
```
hashi-derek, line 6: Not sure why this file changed. It was updated in the original PR.

erichaberkorn (approved):

erichaberkorn (commented):
I wonder if `ServerAddresses` and `ManualServerAddresses` should be separate fields that both contain server addresses rather than `ManualServerAddresses` being a bool. Then we wouldn’t even need to worry about this squashing issue. What do you think?

agent/proxycfg/mesh_gateway.go
```diff
@@ -621,7 +621,13 @@ func (s *handlerMeshGateway) handleUpdate(ctx context.Context, u UpdateEvent, sn
 				continue
 			}
 
-			hostnames, ips := peerHostnamesAndIPs(meshLogger, peering.Name, peering.PeerServerAddresses)
+			// Use the manual server addresses, if defined. Otherwise, use the peer addresses.
+			serverAddresses := peering.ManualServerAddresses
+			if len(serverAddresses) == 0 {
+				serverAddresses = peering.PeerServerAddresses
+			}
```
erichaberkorn, line 628: Should we put this logic into a helper function? It looks like this is repeated a few times.

agent/rpc/peering/validate.go
```diff
@@ -36,6 +36,17 @@ func validatePeeringToken(tok *structs.PeeringToken) error {
...
+	}
+	for _, addr := range tok.ManualServerAddresses {
+		if err := validAddr(addr); err != nil {
+			return err
+		}
+	}
+	for _, addr := range tok.ServerAddresses {
+		if err := validAddr(addr); err != nil {
+			return err
+		}
```
erichaberkorn, line 49: Should it be an error if both of them are set?

agent/proxycfg/mesh_gateway.go
hashi-derek: I can probably put that into a method that's called `GetAddressesToDial()` or something like that.

agent/rpc/peering/validate.go
hashi-derek: Well, right now it sends both of them in the token still. I suppose I could change the logic to not populate those addresses if the manual is specified. I don't know if the tokens are automatically rotated and would have issues, but I can check.

erichaberkorn (approved):

PR #15272, opened 2022-11-07
fix(mesh-gateway): remove deregistered service from mesh gateway

huikang:
### Description
- Deregistered service persists in the mesh-gateway's clusters. This occurs when the mesh-gateway `handleUpdate` un-watches a service without removing it from the ServiceGroups.

- This PR also fixes another issue of exporting unregistered services (details below). An exported service is still being available to importing service, while it is deregistered and removed from the catalog of exporting cluster.

* Establish peering between 2 clusters
1. Export a service from cluster A to cluster B ( A ---> B)
2. Verify the service in B can access service in A
3. Deregister the service in A
4. `Export~service` still exists in mesh gateway of A. Further, the service in B can still access the exported service, although it has been deregistered in cluster A.

* Single DC
1. Launch a consul server agent
2. Create a mesh gateway and make sure mesh gateway is healthy in consul UI
3. Register a service and wait till the service is added to the gateway clusters
4. Deregister the service and the service still exists in mesh gateway clusters

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


agent/proxycfg/mesh_gateway.go
```diff
@@ -258,6 +258,10 @@ func (s *handlerMeshGateway) handleUpdate(ctx context.Context, u UpdateEvent, sn
 				//                 Do those endpoints get cleaned up some other way?
 				delete(snap.MeshGateway.WatchedServices, sid)
 				cancelFn()
+
+				if len(snap.MeshGateway.ServiceGroups) > 0 {
```
hashi-derek, line 262: Is there a reason why this is wrapped with a length check?

huikang: it was used for debugging. Removed. Thanks for reviewing.

agent/proxycfg/snapshot.go
```diff
@@ -482,6 +482,10 @@ func (c *ConfigSnapshot) MeshGatewayValidExportedServices() []structs.ServiceNam
 			continue // not possible
 		}
 
+		if _, ok := c.MeshGateway.ServiceGroups[svc]; !ok {
```
hashi-derek, line 485: I'm a little unfamiliar with the ServiceGroups. So it looks like the `MeshGatewayValidExportedServices` function is used in a few places. Is this going to cause issues elsewhere in the code, or have you already verified all of that logic?

huikang: I verified changes for the exported services in peering and and wan federated clusters.

Based on my understanding, `MeshGatewayValidExportedServices` is used to return a list of services that are exported. However, since the exported service could be deregistered, these services should be removed from the mesh gateway, i.e., removing from the `ServiceGroups`. The current code doesn't remove them because deregister service doesn't trigger the event at https://github.com/hashicorp/consul/blob/985a4ee1b1644f9ab74de90d39dfc0a4233d8d60/agent/proxycfg/mesh_gateway.go#L695  ;
Actually, the notify is unwatched, so it will never triggered
https://github.com/hashicorp/consul/blob/985a4ee1b1644f9ab74de90d39dfc0a4233d8d60/agent/proxycfg/mesh_gateway.go#L232-L262

This is noted https://github.com/hashicorp/consul/blob/985a4ee1b1644f9ab74de90d39dfc0a4233d8d60/agent/proxycfg/mesh_gateway.go#L256-L260

hashi-derek (approved):

agent/proxycfg/snapshot.go
hashi-derek: Thanks for taking the time to explain all of that.

PR #15293, opened 2022-11-08
connect/ca: clarify error message when tuning Vault mount fails

boxofrad:
### Description
We recently helped a customer troubleshoot a failed upgrade, where their Connect CA failed to initialize because it encountered a [bug](https://github.com/hashicorp/vault/pull/16246) in Vault's role validation.

The real problem was obscured by the log message that this PR changes (it wasn't clear what the implications of `tune` failing were) which delayed us finding the bug.

### DIFF
--- a/agent/connect/ca/provider_vault.go
+++ b/agent/connect/ca/provider_vault.go
@@ -52,8 +52,11 @@ const (
 	retryJitter = 20
 )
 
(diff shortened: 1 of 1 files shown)


dhiaayachi (approved):
🚢

rboyer:
Should this be backported?

boxofrad:
Looks like this is superseded by #15525

PR #15297, opened 2022-11-08
[wip] removes ioutil usage everywhere which was deprecated in go1.16

kschoche:
### Description
This is a cleanup PR that I'm testing, it would be sweet if it passed all the unit tests :-D

```
Deprecated: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.
https://pkg.go.dev/io/ioutil 

```

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/api/acl.go
+++ b/api/acl.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/url"
 	"time"
 
@@ -1047,7 +1046,7 @@ func (a *ACL) RulesTranslate(rules io.Reader) (string, error) {
 	parseQueryMeta(resp, qm)
 	qm.RequestTime = rtt
 
-	ruleBytes, err := ioutil.ReadAll(resp.Body)
+	ruleBytes, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return "", fmt.Errorf("Failed to read translated rule body: %v", err)
 	}
@@ -1074,7 +1073,7 @@ func (a *ACL) RulesTranslateToken(tokenID string) (string, error) {
 	parseQueryMeta(resp, qm)
 	qm.RequestTime = rtt
 
-	ruleBytes, err := ioutil.ReadAll(resp.Body)
+	ruleBytes, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return "", fmt.Errorf("Failed to read translated rule body: %v", err)
 	}
--- a/command/logout/logout_test.go
+++ b/command/logout/logout_test.go
@@ -226,7 +226,7 @@ func TestLogoutCommand_k8s(t *testing.T) {
 	})
 
 	// go to the trouble of creating a login token
-	// require.NoError(t, ioutil.WriteFile(bearerTokenFile, []byte(acl.TestKubernetesJWT_B), 0600))
+	// require.NoError(t, os.WriteFile(bearerTokenFile, []byte(acl.TestKubernetesJWT_B), 0600))
 
 	// spin up a fake api server
 	testSrv := kubeauth.StartTestAPIServer(t)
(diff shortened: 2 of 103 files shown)


kschoche:
@dhiaayachi I'm not sure if this one's been proposed before, but thought I'd see if there was any interest in this PR. Thoughts?

command/logout/logout_test.go
```diff
@@ -226,7 +226,7 @@ func TestLogoutCommand_k8s(t *testing.T) {
 	})
 
 	// go to the trouble of creating a login token
-	// require.NoError(t, ioutil.WriteFile(bearerTokenFile, []byte(acl.TestKubernetesJWT_B), 0600))
+	// require.NoError(t, os.WriteFile(bearerTokenFile, []byte(acl.TestKubernetesJWT_B), 0600))
```
dhiaayachi, line 229: Maybe delete this commented code

dhiaayachi (approved):
Thank you for working on this @kschoche!
Seems like your experiment is working 🚀 
I added a comment about a commented code that I think we should remove. 

I also have a suggestion to avoid regression for this (as long as ioutils is existing in golang). Maybe we can use something like this https://github.com/le0tk0k/noioutil to vet the code in the CI but feel free to merge this as is and have this in a subsequent PR.

rboyer:
You might be able to use the `forbidigo` golanglint-ci configuration to prohibit calls: https://github.com/hashicorp/consul/blob/main/.golangci.yml#L85

kschoche:
> You might be able to use the `forbidigo` golanglint-ci configuration to prohibit calls: https://github.com/hashicorp/consul/blob/main/.golangci.yml#L85

Ahh this is a great find, thanks!
```
--> Running golangci-lint
command/logout/logout_test.go:34:12: use of `ioutil.ReadFile` forbidden because "Use io and os packages instead of ioutil" (forbidigo)
        _, err := ioutil.ReadFile("abcd")
                  ^
make: *** [lint] Error 1
```

.changelog/15297.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+all: Remove the use of the deprecated ioutil package throughout the codebase in favour of io and os packages which are still maintained and also introduce a lint rule which forbids the use of ioutil in commits.
```
rboyer, line 2: Many times this sort of fixup doesn't necessitate a changelog entry. Your call if you think we should have one.

kschoche: Makes sense! I was on the fence about it, so I'll remove it then, I think the commit message should be good enough!

api/acl.go
```diff
@@ -1047,7 +1046,7 @@ func (a *ACL) RulesTranslate(rules io.Reader) (string, error) {
 	parseQueryMeta(resp, qm)
 	qm.RequestTime = rtt
 
-	ruleBytes, err := ioutil.ReadAll(resp.Body)
+	ruleBytes, err := io.ReadAll(resp.Body)
```
rboyer, line 1049: Usually we handle go version compatibility differently in the `api` package so that folks using the library aren't forced to upgrade to a future version before their ready. I think the policy seems to be assuming folks are using a go version that  is still receiving security updates (which 1.16 is definitely NOT anymore).

So in this case it's fine to do the swap. I wonder if we should correlate this with bumping the `go` directive in `api/go.mod` from `1.12` to at least `1.16` or maybe all the way to `1.18` (which is the oldest version we test against in CI).

rboyer: The same logic would apply to `sdk/go.mod` too.

kschoche: 👍🏽 Can do, I think `1.18` makes sense here then.
There is a *ton* of module tidying, save that for another PR?

rboyer (approved):

PR #15302, opened 2022-11-08
update config defaults, add docs

malizz:
### Description
Change 1.14 Config Defaults
- Peering is enabled by default
- Set the gRPC TLS port to 8503 by default (on Consul servers)
- Connect is enabled by default

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


website/content/docs/upgrading/upgrade-specific.mdx
```diff
@@ -30,7 +32,8 @@ Prior to Consul 1.14, it was possible to encrypt communication between Consul an
 Consul 1.14 introduces [`ports.grpc_tls`](/docs/agent/config/config-files#grpc_tls_port), a new configuration
 for encrypting communication over gRPC. The existing [`ports.grpc`](/docs/agent/config/config-
 files#grpc_port) configuration **will stop supporting encryption in a future release**. As of version 1.14,
-`ports.grpc_tls` is the recommended configuration to encrypt gRPC traffic.
+`ports.grpc_tls` is the recommended configuration to encrypt gRPC traffic. The default value for gRPC TLS port is 8503
+in Consul servers.
```
hashi-derek, line 36:
```suggestion
for Consul servers.
```

website/content/docs/agent/config/config-files.mdx
```diff
@@ -610,8 +610,8 @@ Valid time units are 'ns', 'us' (or 'µs'), 'ms', 's', 'm', 'h'."
     in `-dev` mode. The `grpc` port currently supports either plaintext or TLS traffic for
     backwards-compatibility, but TLS support is deprecated and will be removed in a future
     release. Refer to `grpc_tls` for more information on configuring a TLS-enabled port.
-  - `grpc_tls` ((#grpc_tls_port)) - The gRPC API with TLS connections, -1 to disable. Default -1 (disabled).
-    **We recommend using `8502` for `grpc_tls`** as your conventional gRPC port number, as it allows some
+  - `grpc_tls` ((#grpc_tls_port)) - The gRPC API with TLS connections, -1 to disable. gRPC_TLS is enable by default.
```
hashi-derek, line 613:
```suggestion
  - `grpc_tls` ((#grpc_tls_port)) - The gRPC API with TLS connections, -1 to disable. gRPC_TLS is enabled by default on port 8503 for Consul servers.
```

agent/config/builder.go
```diff
@@ -436,6 +436,9 @@ func (b *builder) build() (rt RuntimeConfig, err error) {
 	serverPort := b.portVal("ports.server", c.Ports.Server)
 	grpcPort := b.portVal("ports.grpc", c.Ports.GRPC)
 	grpcTlsPort := b.portVal("ports.grpc_tls", c.Ports.GRPCTLS)
+	if grpcTlsPort == -1 && boolVal(c.ServerMode) { // default gRPC TLS port for servers is 8503
```
hashi-derek, line 439: If we check after doing the `b.portVal()` call, then users will have no way to disable the port manually.

We should only set the default if `c.Ports.GRPCTLS == nil`, because that will mean that the field was never populated with a value.

agent/config/builder.go
```diff
@@ -962,7 +965,7 @@ func (b *builder) build() (rt RuntimeConfig, err error) {
 		AutoEncryptAllowTLS:                    autoEncryptAllowTLS,
 		AutoConfig:                             autoConfig,
 		Cloud:                                  b.cloudConfigVal(c.Cloud),
-		ConnectEnabled:                         connectEnabled,
+		ConnectEnabled:                         boolValWithDefault(&connectEnabled, true),
```
hashi-derek, line 968: I think this should probably be in the `agent/config/default.go` file now that I think about it, because the connect enable flag is something that is set on both agents and servers.

agent/config/builder.go
```diff
@@ -1021,7 +1024,7 @@ func (b *builder) build() (rt RuntimeConfig, err error) {
 		NodeMeta:                          c.NodeMeta,
 		NodeName:                          b.nodeName(c.NodeName),
 		ReadReplica:                       boolVal(c.ReadReplica),
-		PeeringEnabled:                    boolVal(c.Peering.Enabled),
+		PeeringEnabled:                    boolValWithDefault(c.Peering.Enabled, true),
```
hashi-derek, line 1027: This one should probably also have the same `agent/config/default.go` treatment (even though the field only appears to be used by servers currently). We could also enable it for dev mode by default.

david-yu:
Hi @malizz Thanks for doing this, could we also call out the connect and cluster peering defaults as breaking changes in our changelog. In our upgrade specific guideline, we should also note that during upgrades those settings must be set to false prior to the upgrade if you do not want to enable such functionality.

agent/config/builder.go
malizz: Do we want it to be enable even when `autoEncryptAllowTLS || autoConfig.Enabled` is false?

hashi-derek: Connect can function without auto encrypt or auto config, so I think it should be okay.

.changelog/15302.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:improvement
+config: update 1.14 config defaults: Enable `peering` and `connect` by default.
```
david-yu, line 2: Is there a reason why peering and connect are not breaking changes? I would think it would be breaking as it may cause issues for folks upgrading who previously did not enable those features.

malizz: @hashi-derek may have more context here. 
@david-yu Are there any specific cases that enabling peering or connect can break things?

hashi-derek: I think he means to set the tag to `release-note:breaking-change`

hashi-derek: Sorry, I only saw the conversation and two lines of code in the conversation overview, and I didn't see the whole file here. I think it wouldn't hurt to put it as a breaking change if there are users that explicitly wanted to disable connect / peering for some reason.

I don't think peering will have any adverse side-effects if nobody attempts to use the feature set, but it could be a security concern for operators, I suppose. Connect is complicated, so it could potentially have some undesired effects.

david-yu: @hashi-derek For large service discovery use cases I'm wondering if it does cause issues. I would recommend keeping it at breaking change for connect perhaps unless we have some internal testing that provides evidence that upgrades on large scale SD system on 1.13 looks ok.

hashi-derek: Yeah, let's just list them both as breaking so that operators are more aware that these are enabled.

erichaberkorn (approved):

Issue #11092, opened 2021-09-20
dnsNames in csr for ingress should not include port


fredwangwang:
#### Overview of the Issue

We are using consul connect with Nomad, and we have an ingress job looks like follows (with some other hosts omitted):
![image](https://user-images.githubusercontent.com/21690857/134071604-5d30c062-8566-48e1-a526-635acdd542ac.png)

Before we are using the builtin Connect CA provider, as we plan to migrate to using Vault as the Connect CA provider, we found an issue when consul is trying to obtain the cert for ingress:
```
[ERROR] agent.proxycfg: Failed to handle update from watch: service_id=_nomad-task-e4b4a015-8d2e-b6f3-3453-70043288bbd1-group-ingress-group-ingress-inbound id=leaf error="error filling agent cache: error issuing cert: Error making API request.
 
URL: PUT http://vault.service.consul:8200/v1/pki_consul_connect_int/sign/leaf-cert
Code: 400. Errors:
 
* subject alternate name ingress.service.consul:8080 not allowed by this role"
```

As the error suggest consul simply takes the value from `hosts` and put them in the csr. However the port should not be there in the dns names for csr if i understand correctly.

#### Reproduction Steps

1. Using a host with port in `hosts` when configuring ingress
2. using vault as Connect CA provider

Looks like the related part is here: https://github.com/hashicorp/consul/blob/959299077322953ce82ce0756ae9a8626ce907db/agent/proxycfg/ingress_gateway.go#L229

blake:
Hi @fredwangwang, Consul 1.8.0 required the ingress port to be specified in the hosts value when using a non-standard port. This was fixed in Consul 1.8.1 by PR #8190, but the callout to include the port was not removed from either the Consul or Nomad documentation until last month (PR #10827 and hashicorp/nomad#11042).

If you are using a version of Consul newer than 1.8.0, you can safely remove the port from the `hosts` field in your job specification.

fredwangwang:
Hi @blake  we are actually using consul 1.10.1 (sorry didn't include the version in the post)

And tried to remove the port it does not work. Here the detailed explain:

we have client side lb listen on port 29999, and we have ingress listen on port 8080. For our non mesh service to talk to mesh service, we have them going through clb (29999) instead of ingress directly for backwards compatibility. All clb does is forwarding req to ingress.

So our network flow:
srvA --> localhost:29999/srvB --> localhost:8080/srvB --> srvB in mesh.

[...]

PR #15320, opened 2022-11-09
connect: strip port from DNS SANs for ingress gateway leaf cert

kyhavlov:
Closes https://github.com/hashicorp/consul/issues/11092 - currently we're passing along whatever's in the `Hosts` field verbatim to construct the list of SANs. This PR changes that to strip the `:` and everything after it from each `Host` string so that it's in the format Vault expects when signing the leaf cert.

### DIFF
--- a/agent/connect/csr.go
+++ b/agent/connect/csr.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"strings"
 )
 
 // SigAlgoForKey returns the preferred x509.SignatureAlgorithm for a given key
@@ -47,11 +48,28 @@ func SigAlgoForKeyType(keyType string) x509.SignatureAlgorithm {
 // along with the PEM-encoded private key for this certificate.
 func CreateCSR(uri CertURI, privateKey crypto.Signer,
 	dnsNames []string, ipAddresses []net.IP, extensions ...pkix.Extension) (string, error) {
+
+	// Drop everything after the ':' from the name when constructing the DNS SANs.
+	uniqueNames := make(map[string]struct{})
+	formattedDNSNames := make([]string, 0)
+	for _, host := range dnsNames {
+		hostSegments := strings.Split(host, ":")
+		formattedHost := hostSegments[0]
+		if _, ok := uniqueNames[formattedHost]; ok {
+			continue
+		}
+
+		if len(hostSegments) >= 1 {
+			formattedDNSNames = append(formattedDNSNames, formattedHost)
+			uniqueNames[formattedHost] = struct{}{}
+		}
+	}
+
 	template := &x509.CertificateRequest{
 		URIs:               []*url.URL{uri.URI()},
 		SignatureAlgorithm: SigAlgoForKey(privateKey),
 		ExtraExtensions:    extensions,
-		DNSNames:           dnsNames,
+		DNSNames:           formattedDNSNames,
 		IPAddresses:        ipAddresses,
 	}
 	HackSANExtensionForCSR(template)
--- a/agent/connect/csr.go
+++ b/agent/connect/csr.go
@@ -54,12 +54,12 @@ func CreateCSR(uri CertURI, privateKey crypto.Signer,
 	formattedDNSNames := make([]string, 0)
 	for _, host := range dnsNames {
 		hostSegments := strings.Split(host, ":")
-		formattedHost := hostSegments[0]
-		if _, ok := uniqueNames[formattedHost]; ok {
+		if len(hostSegments) == 0 || hostSegments[0] == "" {
 			continue
 		}
 
-		if len(hostSegments) >= 1 {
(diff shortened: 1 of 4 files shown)


agent/proxycfg/ingress_gateway.go
```diff
@@ -288,7 +289,19 @@ func (s *handlerIngressGateway) generateIngressDNSSANs(snap *ConfigSnapshot) []s
...
 
-	dnsNames = append(dnsNames, snap.IngressGateway.Hosts...)
+	addedHosts := make(map[string]struct{})
+	for _, host := range snap.IngressGateway.Hosts {
+		if _, ok := addedHosts[host]; ok {
+			continue
+		}
+
+		// Drop everything after the ':' from the host when constructing the DNS SANs.
+		hostSegments := strings.Split(host, ":")
```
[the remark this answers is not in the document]
kyhavlov: The list of hosts also ends up in the `domains` field of the `VirtualHosts` we send to envoy, which needs to include the port in some cases: https://www.envoyproxy.io/docs/envoy/latest/faq/debugging/why_is_my_route_not_found 

So including the port is needed for the route, but causes an error when signing the cert if it's included in the DNS SANs

hashi-derek (approved):

jkirschner-hashicorp:
@kyhavlov : Thanks! A few follow-up questions:
- Which versions do we intend to attempt to backport this to?
- Should we intervene where you did (at the ingress gateway point)? Or is there any reason to strip host out from dnsnames at a more general point ([like in the CSR code](https://github.com/hashicorp/consul/blob/325c1c0dd7af97601edd72601b4ab4e190b25b00/agent/connect/csr.go#L54))?
- Do we need any automated tests for this? (Since there are no unit tests, I assume we manually tested instead?)

agent/connect/csr.go
```diff
@@ -47,11 +48,28 @@ func SigAlgoForKeyType(keyType string) x509.SignatureAlgorithm {
 // along with the PEM-encoded private key for this certificate.
 func CreateCSR(uri CertURI, privateKey crypto.Signer,
 	dnsNames []string, ipAddresses []net.IP, extensions ...pkix.Extension) (string, error) {
+
+	// Drop everything after the ':' from the name when constructing the DNS SANs.
+	uniqueNames := make(map[string]struct{})
+	formattedDNSNames := make([]string, 0)
+	for _, host := range dnsNames {
+		hostSegments := strings.Split(host, ":")
+		formattedHost := hostSegments[0]
```
jkirschner-hashicorp, line 57: Should we confirm that `len(hostSegments) > 0` here before accessing the 0th element, to ensure we don't have an out-of-bounds access?

agent/connect/csr.go
```diff
@@ -47,11 +48,28 @@ func SigAlgoForKeyType(keyType string) x509.SignatureAlgorithm {
...
+		if _, ok := uniqueNames[formattedHost]; ok {
+			continue
+		}
+
+		if len(hostSegments) >= 1 {
```
jkirschner-hashicorp, line 62: This check seems to be ineffective in its current position, because we're already assuming that `hostSegments` has at least 1 element by accessing its 0th element without a guard.

jkirschner-hashicorp (changes requested):
@kyhavlov: I _think_ I spotted something that can quickly be fixed.

agent/connect/csr.go
jkirschner-hashicorp: To test this edge case, I guess you could add a host `""` to the array in your test

pglass: This should be okay. [`strings.Split`](https://pkg.go.dev/strings#Split) only returns an empty list when both arguments are empty.

`formattedHost` could be empty here, but that would only occur if the one of `dnsNames` had no hostname, like `:8080`.

agent/connect/csr_test.go
```diff
@@ -0,0 +1,32 @@
...
+	require.NoError(t, err)
+	spiffeID := &SpiffeIDService{
+		Host:       "7528f42f-92e5-4db4-b84c-3405c3ca91e6",
+		Service:    "srv1",
+		Datacenter: "dc1",
+	}
+	csr, err := CreateCSR(spiffeID, pk, []string{
+		"foo.example.com",
+		"foo.example.com:8080",
+		"bar.example.com",
```
pglass, line 20: Maybe throw in a wildcard hostname in this test, like `*.example.com` because [that is allowed](https://developer.hashicorp.com/consul/docs/connect/config-entries/ingress-gateway#hosts)

pglass (approved):

agent/connect/csr.go
jkirschner-hashicorp: We wouldn't want a bad configuration (like "`:8080`") to cause a panic though, right?

jkirschner-hashicorp (approved):
LGTM!

PR #3431, opened 2017-08-30
Network segments

kyhavlov:
Adds the OSS components for network segments. This feature allows clients to be split up into named network segments, in order to separate their gossip layer from clients in different segments. The client's segment is stored in the `consul-network-segment` key in its node metadata.

Servers are always in all segments, and each segment can be configured with its own bind/advertise address and port (along with an option for an rpc listener if the segment's bind address is different from the default rpc addr).

PR #15360, opened 2022-11-14
Ensure imported services are filtered in IndexedNodesWithGateways

freddygv:
1.13 backport of #3431

Previously the ACL filter was not operating on ImportedNodes of
IndexedNodesWithGateways, which means that the UI Nodes views
will always return data imported from a peer without filtering.
    
Because of this, imported nodes/services/checks could be viewed by
users lacking the appropriate ACL permissions.


freddygv:
Closing, will re-do

PR #15370, opened 2022-11-14
auto-config: relax node name validation for JWT authorization

kyhavlov:
This changes the JWT authorization logic to allow all non-whitespace, non-quote characters when validating node names. Consul had previously allowed these characters in node names, until this validation was added to fix a security vulnerability with whitespace/quotes being passed to the `bexpr` library. This unintentionally broke node names with characters like `.` which aren't related to this vulnerability. This PR also adds a warning on agent startup if an invalid character is present in the node name (at some point we can consider changing this to a hard error).

### DIFF
--- a/website/content/docs/agent/config/cli-flags.mdx
+++ b/website/content/docs/agent/config/cli-flags.mdx
@@ -482,8 +482,7 @@ information.
 
 - `-node` ((#\_node)) - The name of this node in the cluster. This must
   be unique within the cluster. By default this is the hostname of the machine.
-  This cannot contain whitespace or quotes, and in order for the node to be
-  discoverable via DNS it must only contain alphanumerics and dashes (`-`).
+  The node name cannot contain whitespace or quotation marks. To query the node from DNS, the name must only contain alphanumeric characters and hyphens (`-`).
 
 - `-node-id` ((#\_node_id)) - Available in Consul 0.7.3 and later, this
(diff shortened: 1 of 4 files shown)


website/content/docs/agent/config/cli-flags.mdx
```diff
@@ -482,6 +482,8 @@ information.
 
 - `-node` ((#\_node)) - The name of this node in the cluster. This must
   be unique within the cluster. By default this is the hostname of the machine.
+  This cannot contain whitespace or quotes, and in order for the node to be
+  discoverable via DNS it must only contain alphanumerics and dashes (`-`).
```
trujillo-adam, line 486:
```suggestion
  The node name cannot contain whitespace or quotation marks. To query the node from DNS, the name must only contain alphanumeric characters and hyphens (`-`).
```
I know these are kind of nitpick-y, but _quotes_ are the words inside of the _quotation marks_. We tend to misuse and overuse _via_, which means going from point A to point C by passing through point B. Finally, a single `-` is a _hyphen_. It's improbable that someone would misinterpret _dash_ to mean em dash, underscore, back-tick, etc., but it's probably best to use the correct word.

trujillo-adam (commented):
A few nits for the docs.

trujillo-adam (approved):
Docs look good!

hashi-derek (approved):

PR #15373, opened 2022-11-15
doc: minor clarifiation to mesh gateway

huikang:
### Description
Minor update to the mesh gateway doc

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/website/data/docs-nav-data.json
+++ b/website/data/docs-nav-data.json
@@ -471,7 +471,7 @@
                 "path": "connect/gateways/mesh-gateway"
               },
               {
-                "title": "WAN Federation",
+                "title": "Enabling WAN Federation Control Plane Traffic",
                 "path": "connect/gateways/mesh-gateway/wan-federation-via-mesh-gateways"
               },
               {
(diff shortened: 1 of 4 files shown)


website/data/docs-nav-data.json
```diff
@@ -471,7 +471,7 @@
                 "path": "connect/gateways/mesh-gateway"
               },
               {
-                "title": "WAN Federation",
+                "title": "Enabling WAN Federation Control Plane Traffic",
```
DanStough, line 474: I like this new name a lot. Optional: do you think we should change the page H1 title as well? I think yes.

DanStough (approved):
Great edits @huikang! I just have one comment you can take or leave.

DanStough:
Note: since there is already a release branch now, I think we should use the `type/docs-cherry` label and waiting until the release is finished to merge this branch. It will get added to `stable-website` one the 1.14 documentation is added.

website/data/docs-nav-data.json
huikang: Updated the title in a subsequent commit. Hope it makes sense to you.

PR #15374, opened 2022-11-15
docs: support for peer label in DNS lookups

jkirschner-hashicorp:
Ready for education team review at [this preview link](https://consul-e0js5mavl-hashicorp.vercel.app/consul/docs/discovery/dns)

### DIFF
--- a/website/content/docs/discovery/dns.mdx
+++ b/website/content/docs/discovery/dns.mdx
@@ -149,16 +149,24 @@ The following are valid formats for standard service lookups:
   [<tag>.]<service>.service.<peer>.peer.<domain>
   ```
 
+  The `<peer>` label identifies the cluster peer in which
+  to lookup the specified `<service>`.
+
 - Lookup a service in a WAN federated datacenter:
 
   ```text
   [<tag>.]<service>.service.<datacenter>[.dc].<domain>
   ```
 
-  Use of the optional `.dc` label is recommended for readability.
+  We recommend including the optional `.dc` label to improve readability.
+  The `<datacenter>` label identifies the WAN federated datacenter in which
+  to lookup the specified `<service>`.
 
-In all formats, `tag` is optional.
-If no tag is provided, no filtering is done on tag.
+You can filter the results to only include instances of a service with a given tag
+by specifying `<tag>` at the start of the lookup.
+If you do not specify a tag, the lookup returns all instances of the service.
+Refer to the [service configuration documentation](/consul/docs/discovery/services#service)
+for information about adding tags to your service.
 
 To lookup services directly from a mesh service's application code
 with transparent proxy enabled, refer to 
@@ -210,31 +218,37 @@ foobar.node.dc1.consul. 0 IN  A 10.1.10.12
 
 ### RFC 2782 Lookup
 
-RFC 2782 style lookups have the same behavior as
-[standard service lookups](#standard-lookup),
-but differ in format.
+You can use [RFC 2782](https://tools.ietf.org/html/rfc2782) lookups to query for services
+as an alternative to [standard lookups](#standard-lookup).
 The RFC 2782 lookup format supports the same suffixes as standard lookups
 for optionally specifying a cluster peer or WAN federated datacenter,
 but differs in how service name and optional tag are specified.
 The valid RFC 2782 lookup formats below assume a local service lookup:
 
-- No filtering on service tag:
+- Lookup a local service:
 
    ```text
   _<service>._tcp[.service].<domain>
   ```
 
-- Filtering on service tag specified in the RFC 2782 protocol field:
+- Lookup a local service with filtering on service tag
+  specified in the RFC 2782 protocol field:
 
    ```text
   _<service>._<tag>[.service].<domain>
   ```
 
+We recommend including the optional `.service` label to improve readability.
+
 Per [RFC 2782](https://tools.ietf.org/html/rfc2782), SRV queries must
-prepend an underscore (`_`) to the `service` and `protocol` values in a query to
-prevent DNS collisions.
-To perform no tag-based filtering, specify `tcp` in the RFC 2782 protocol field.
-To filter results on a service tag, specify the tag in the RFC 2782 protocol field.
+prepend an underscore (`_`) to the RFC 2782 `service` and `protocol` values
+in a query to prevent DNS collisions.
+
+You can filter the results to only include instances of a service with a given tag
+by specifying `<tag>` at the start of the lookup.
+If you do not specify a tag, the lookup returns all instances of the service.
+Refer to the [service configuration documentation](/consul/docs/discovery/services#service)
+for information about adding tags to your service.
 
 The following table shows example lookups for the `api` service
 depending on its location and whether you want to filter for instances
@@ -417,17 +431,20 @@ In each format, all fields in square brackets are optional.
   [<tag>.]<service>.service[.<namespace>.ns][.<peer>.peer][.<partition>.ap].<domain>
   ```
 
-  This format supports `.service` lookup types in Consul 1.14.2 or later
-  and supports `.virtual` lookup types in Consul 1.14.0 or later.
+  This format supports `.service` and `.virtual` service lookup types.
 
-  The `peer` is a cluster peer of the `partition`.
-  If no `peer` is specified, the service lookup applies to the specified `partition`
-  rather than one of its cluster peers.
+  ~> **Compatibility warning:** Consul 1.14.2+ is required for `.service` lookup types.
 
-  For example, assume that an agent in the `default` partition is acting as the
-  DNS server, the `k8s-app-1` partition is in the same datacenter, and
-  the `k8s-app-1` partition has a cluster peer named `billing`.
-  To perform a lookup of non-mesh service `api` in namespace `foo`
+  If `.<peer>.peer` is included, the `<peer>` label identifies a cluster peer
+  of the query's partition in which to lookup the specified `<service>`.
+  If the query does not specify a partition,
+  the query uses the partition of the Consul agent that received the DNS query.
+
+  In the following example,
+  a Consul agent in the `default` partition acts as the DNS server,
+  the `k8s-app-1` partition is in the same datacenter,
+  and the `k8s-app-1` partition has a cluster peer named `billing`.
+  To perform a lookup of service `api` in namespace `foo`
   of that cluster peer, use the query
   `api.service.foo.ns.billing.peer.k8s-app-1.partition.consul`.
 
@@ -437,12 +454,13 @@ In each format, all fields in square brackets are optional.
   [<tag>.]<service>.service[.<namespace>.ns][.<partition>.ap][.<datacenter>.dc].<domain>
   ```
 
-  This format supports `.service`, `.connect`, `.virtual`, and `.ingress` service lookup types
-  in Consul 1.14.0 or later.
+  This format supports `.service`, `.connect`, `.virtual`, and `.ingress` service lookup types.
 
-  Cluster peering must be used to connect datacenters containing admin partitions,
-  not WAN federation. Therefore, this format is not applicable to lookup a partition
-  in a different datacenter.
+  This format does not support specifying both a partition and another datacenter.
+  Datacenters containing admin partitions can only be connected using cluster peering,
+  not WAN federation. To lookup a service in a partition of another datacenter,
+  you must establish a cluster peering relationship with that partition and
+  use the cluster peer lookup format.
 
 The following table shows example `.service` lookups for the `api` service
 depending on its location and whether you want to filter for instances
--- a/website/content/docs/discovery/dns.mdx
+++ b/website/content/docs/discovery/dns.mdx
@@ -508,7 +508,9 @@ use one of the following formats:
 - Use the [canonical format](#canonical-format) to optionally specify
   namespace, peer, and/or partition:
 
+  ```text
   <service>.virtual[.<namespace>.ns][.<peer>.peer][.<partition>.ap].<domain>
+  ```
 
 To lookup services outside the mesh or not imported from a cluster peer,
 refer to [service lookups for Consul Enterprise](#service-lookups-for-consul-enterprise) instead.
--- a/website/content/docs/discovery/dns.mdx
+++ b/website/content/docs/discovery/dns.mdx
@@ -133,25 +133,46 @@ it is recommended to use the HTTP API to retrieve the list of nodes.
 
 ### Standard Lookup
 
-The format of a standard service lookup is:
+The following are valid formats for standard service lookups:
 
-```text
-[<tag>.]<service>.service[.<datacenter-or-peer>].<domain>
-```
+- Lookup a local service:
 
-The `tag` is optional. If no tag is provided, no filtering is done on tag.
+  ```text
+  [<tag>.]<service>.service.<domain>
+  ```
 
-The `datacenter-or-peer` is also optional.
-If `datacenter-or-peer` is not provided, the datacenter of the Consul agent is assumed.
-A provided value must be the name a WAN federated datacenter or a cluster peer
-of the Consul agent acting as the DNS server.
-To lookup a mesh-participating service that's imported from a cluster peer,
-refer to [service virtual IP lookup](#service-virtual-ip-lookups) instead.
+  The lookup uses the datacenter of the Consul agent acting as the DNS server.
 
-If we want to find any redis service providers in our local datacenter,
-we could query `redis.service.consul.` If we want to find the PostgreSQL
-primary in a WAN federated datacenter or cluster peer named `dc2`,
-we could query `primary.postgresql.service.dc2.consul.`
+- Lookup a service in a cluster peer:
+
+  ```text
+  [<tag>.]<service>.service.<peer>.peer.<domain>
+  ```
+
+- Lookup a service in a WAN federated datacenter:
+
+  ```text
+  [<tag>.]<service>.service.<datacenter>[.dc].<domain>
+  ```
+
+  Use of the optional `.dc` label is recommended for readability.
+
+In all formats, `tag` is optional.
+If no tag is provided, no filtering is done on tag.
+
+To lookup services directly from a mesh service's application code
+with transparent proxy enabled, refer to 
+[service virtual IP lookups for Consul Enterprise](#service-virtual-ip-lookups) instead.
+
+The following table shows example lookups for the `api` service
+depending on its location and whether you want to filter for instances
+of the service by tag. All queries assume the `domain` is `consul`.
+| Tag | Location | Query |
+| --- | -------- | ----- |
+| None | Local   | `api.service.consul` |
+| `v2` | Local   | `v2.api.service.consul` |
+| None | Peer `us-east`   | `api.service.us-east.peer.consul` |
+| None | WAN federated datacenter `dc3` | `api.service.dc3.dc.consul` |
 
 The DNS query system makes use of health check information to prevent routing
 to unhealthy nodes. When a service query is made, any services failing their health
@@ -188,19 +209,24 @@ foobar.node.dc1.consul. 0 IN  A 10.1.10.12
 
 ### RFC 2782 Lookup
 
-Valid formats for RFC 2782 SRV lookups depend on
-whether you want to filter results based on a service tag:
+RFC 2782 style lookups have the same behavior as
+[standard service lookups](#standard-lookup),
+but differ in format.
+The RFC 2782 lookup format supports the same suffixes as standard lookups
+for optionally specifying a cluster peer or WAN federated datacenter,
+but differs in how service name and optional tag are specified.
+The valid RFC 2782 lookup formats below assume a local service lookup:
 
 - No filtering on service tag:
 
    ```text
-  _<service>._tcp[.service][.<datacenter-or-peer>].<domain>
+  _<service>._tcp[.service].<domain>
   ```
 
 - Filtering on service tag specified in the RFC 2782 protocol field:
 
    ```text
-  _<service>._<tag>[.service][.<datacenter-or-peer>].<domain>
+  _<service>._<tag>[.service].<domain>
   ```
 
 Per [RFC 2782](https://tools.ietf.org/html/rfc2782), SRV queries must
@@ -209,8 +235,15 @@ prevent DNS collisions.
 To perform no tag-based filtering, specify `tcp` in the RFC 2782 protocol field.
 To filter results on a service tag, specify the tag in the RFC 2782 protocol field.
 
-Other than the query format and default `tcp` protocol/tag value, the behavior
-of the RFC style lookup is the same as the standard style of lookup.
+The following table shows example lookups for the `api` service
+depending on its location and whether you want to filter for instances
+of the service by tag. All queries assume the `domain` is `consul`.
+| Tag | Location | Query |
+| --- | -------- | ----- |
+| None | Local   | `_api._tcp.service.consul` |
+| `v2` | Local   | `_api._v2.service.consul` |
+| None | Peer `us-east`   | `_api._tcp.service.us-east.peer.consul` |
+| None | WAN federated datacenter `dc3` | `_api._tcp.service.dc3.dc.consul` |
 
 If you registered the service `rabbitmq` on port 5672 and tagged it with `amqp`,
 you could make an RFC 2782 query for its SRV record as `_rabbitmq._amqp.service.consul`:
@@ -366,13 +399,14 @@ If DNS queries are addressed to Consul server agents,
 service lookups to non-`default` partitions must explicitly specify
 the partition of the target service.
 
-To lookup a mesh-participating service that's imported from a cluster peer,
-refer to [service virtual IP lookup for Consul Enterprise](#service-virtual-ip-lookups-for-consul-enterprise) instead.
+To lookup services directly from a mesh service's application code
+with transparent proxy enabled, refer to 
+[service virtual IP lookups for Consul Enterprise](#service-virtual-ip-lookups-for-consul-enterprise) instead.
 
 #### Canonical format
 
-Use the following query formats to specify namespace, partition, datacenter, or peer
-for `.service`, `.connect`, `.virtual`, and `.ingress` service lookup types.
+Use the following query formats to specify namespace, partition,
+WAN federated datacenter, or cluster peer in service lookups.
 In each format, all fields in square brackets are optional.
 
 - Lookup a service, optionally in a cluster peer:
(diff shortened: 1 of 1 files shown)


website/content/docs/discovery/dns.mdx
```diff
@@ -444,33 +473,44 @@ If you need more complex behavior, please use the
...
+- Specify peer and optional namespace in the lookup:
+
+  ```text
+  <service>.virtual[.<namespace>].<peer>.<domain>
+  ```
+
+- Use the [canonical format](#canonical-format) to optionally specify
+  namespace, peer, and/or partition:
+
+  <service>.virtual[.<namespace>.ns][.<peer>.peer][.<partition>.ap].<domain>
```
jkirschner-hashicorp, line 511:
````suggestion
  ```text
  <service>.virtual[.<namespace>.ns][.<peer>.peer][.<partition>.ap].<domain>
  ```
````

File: website/content/docs/discovery/dns.mdx [hunk 2]
```diff
@@ -353,25 +358,49 @@ $ echo -n "20010db800010002cafe000000001337" | perl -ne 'printf join(":", unpack
...
+  The `peer` is a cluster peer of the `partition`.
+  If no `peer` is specified, the service lookup applies to the specified `partition`
+  rather than one of its cluster peers.
+
+  For example, assume that an agent in the `default` partition is acting as the
+  DNS server, the `k8s-app-1` partition is in the same datacenter, and
+  the `k8s-app-1` partition has a cluster peer named `billing`.
+  To perform a lookup of non-mesh service `api` in namespace `foo`
+  of that cluster peer, use the query
+  `api.foo.ns.billing.peer.k8s-app-1.partition.consul`.
```
jkirschner-hashicorp, line 393:
```suggestion
  `api.service.foo.ns.billing.peer.k8s-app-1.partition.consul`.
```

website/content/docs/discovery/dns.mdx
```diff
@@ -353,25 +358,49 @@ $ echo -n "20010db800010002cafe000000001337" | perl -ne 'printf join(":", unpack
...
-All three fields (`namespace`, `partition`, `datacenter`) are optional.
-```text
-[<tag>.]<service>.service[.<namespace>.ns][.<partition>.ap][.<datacenter>.dc]<domain>
-```
+In each format, all fields in square brackets are optional.
+
+- Lookup a service, optionally in a cluster peer:
+
+  ```text
+  [<tag>.]<service>.service[.<namespace>.ns][.<peer>.peer][.<partition>.ap].<domain>
```
hashi-derek, line 381: We don't currently support this format with `.service.p1.peer` in 1.14.0, so we'll need to remove anything referring to it. I believe we'll only be able to use the new `.peer` syntax with `virtual` entries for the time-being.

jkirschner-hashicorp: Sounds good - I'll move that stuff out into another PR ...

... once the accuracy of some of the other content is reviewed (so I can batch all my changes).

I'm particularly not sure about the changes I made to the `### Service Virtual IP Lookups` section.

hashi-derek (approved):

website/content/docs/discovery/dns.mdx
```diff
@@ -136,17 +136,22 @@ it is recommended to use the HTTP API to retrieve the list of nodes.
 The format of a standard service lookup is:
 
 ```text
-[<tag>.]<service>.service[.<datacenter>].<domain>
+[<tag>.]<service>.service[.<datacenter-or-peer>].<domain>
```
ndhanushkodi, line 139: This is not supported, because it's never passed in as a peer here: https://github.com/hashicorp/consul/blob/main/agent/dns.go#L765
Its only passed in as a datacenter.

ndhanushkodi: This would be something we intend to support in the future.

website/content/docs/discovery/dns.mdx
```diff
@@ -136,17 +136,22 @@ it is recommended to use the HTTP API to retrieve the list of nodes.
...
 ```
 
-The `tag` is optional, and, as with node lookups, the `datacenter` is as
-well. If no tag is provided, no filtering is done on tag. If no
-datacenter is provided, the datacenter of this Consul agent is assumed.
+The `tag` is optional. If no tag is provided, no filtering is done on tag.
+
+The `datacenter-or-peer` is also optional.
```
ndhanushkodi, line 144: Intended to support in the future, but no support now since it's always parsed as dc

website/content/docs/discovery/dns.mdx
```diff
@@ -136,17 +136,22 @@ it is recommended to use the HTTP API to retrieve the list of nodes.
...
+of the Consul agent acting as the DNS server.
+To lookup a mesh-participating service that's imported from a cluster peer,
+refer to [service virtual IP lookup](#service-virtual-ip-lookups) instead.
 
 If we want to find any redis service providers in our local datacenter,
 we could query `redis.service.consul.` If we want to find the PostgreSQL
-primary in a particular datacenter, we could query
-`primary.postgresql.service.dc2.consul.`
+primary in a WAN federated datacenter or cluster peer named `dc2`,
+we could query `primary.postgresql.service.dc2.consul.`
```
ndhanushkodi, line 154: Intended to support in the future, but no support now since it's always parsed as dc

website/content/docs/discovery/dns.mdx
```diff
@@ -189,13 +194,13 @@ whether you want to filter results based on a service tag:
 - No filtering on service tag:
 
    ```text
-  _<service>._tcp[.service][.<datacenter>].<domain>
+  _<service>._tcp[.service][.<datacenter-or-peer>].<domain>
```
ndhanushkodi, line 197: Intended to support in the future, but no support now since it's always parsed as dc

website/content/docs/discovery/dns.mdx
```diff
@@ -189,13 +194,13 @@ whether you want to filter results based on a service tag:
...
   ```
 
 - Filtering on service tag specified in the RFC 2782 protocol field:
 
    ```text
-  _<service>._<tag>[.service][.<datacenter>].<domain>
+  _<service>._<tag>[.service][.<datacenter-or-peer>].<domain>
```
ndhanushkodi, line 203: Intended to support in the future, but no support now since it's always parsed as dc

website/content/docs/discovery/dns.mdx
```diff
@@ -353,25 +358,49 @@ $ echo -n "20010db800010002cafe000000001337" | perl -ne 'printf join(":", unpack
 
 By default, all service lookups use the `default` namespace
 within the partition and datacenter of the Consul agent that received the DNS query.
-To lookup services in another namespace, partition, and/or datacenter,
+To lookup services in another namespace, partition, datacenter, or cluster peer,
```
ndhanushkodi, line 361: Intended to support in the future, but no support now since it's always parsed as dc

website/content/docs/discovery/dns.mdx
```diff
@@ -353,25 +358,49 @@ $ echo -n "20010db800010002cafe000000001337" | perl -ne 'printf join(":", unpack
...
+
+- Lookup a service, optionally in a WAN federated datacenter:
+
+  ```text
+  [<tag>.]<service>.service[.<namespace>.ns][.<partition>.ap][.<datacenter>.dc].<domain>
+  ```
+
+  Cluster peering must be used to connect datacenters containing admin partitions,
+  not WAN federation. Therefore, this format is not applicable to lookup a partition
+  in a different datacenter.
```
ndhanushkodi, line 403: I think everything under Canonical format except for L395-403 can be saved for a future pr once .service queries are supported. We can keep L395-403 though

website/content/docs/discovery/dns.mdx
```diff
@@ -381,7 +410,7 @@ you can use the following query formats specify namespace but not partition:
 - Specify both namespace and datacenter:
 
   ```text
-  [<tag>.]<service>.service.<namespace>.<datacenter>.<domain>
+  [<tag>.]<service>.service.<namespace>.<datacenter-or-peer>.<domain>
```
ndhanushkodi, line 413: Intended to support in the future, but no support now since it's always parsed as dc

website/content/docs/discovery/dns.mdx
```diff
@@ -444,33 +473,46 @@ If you need more complex behavior, please use the
...
 
-This will return the unique virtual IP for any [Connect-capable](/docs/connect)
-service. Each Connect service has a virtual IP assigned to it by Consul - this is used
-by sidecar proxies for the [Transparent Proxy](/docs/connect/transparent-proxy) feature.
-The peer name is an optional part of the FQDN, and it is used to query for the virtual IP
-of a service imported from that peer.
+Consul assigns a virtual IP to each service in the service mesh.
+A downstream service in the mesh must address an upstream service
+using its virtual IP in the following scenarios:
+- The upstream service is imported from a cluster peer
```
ndhanushkodi, line 485: You can still address an upstream service imported from a cluster peer using explicit upstreams. In that case you wouldn't need to use the virtual IP. So the only time we must address an upstream using its virtual ip is when the downstream has tproxy enabled. 

"The peer name is an optional part of the FQDN, and it is used to query for the virtual IP
of a service imported from that peer." <- i think this sentence is still accurate

ndhanushkodi (commented):
Left some comments on what to keep as part of this and what's for the future!

website/content/docs/discovery/dns.mdx
jkirschner-hashicorp: If I understand correctly, this format is not supported, but it is now possible to pass a peer as of https://github.com/hashicorp/consul/pull/15596

If I understand correctly, the formats are now:
- `[<tag>.]<service>.service[.<datacenter>].<domain>`
- `[<tag>.]<service>.service[.<datacenter>].dc.<domain>`
- `[<tag>.]<service>.service[.<peer>].peer.<domain>`

jkirschner-hashicorp: @hashi-derek : Is it accurate to say that `[<tag>.]<service>.service[.<datacenter-or-peer>].<domain>` isn't supported, because in that form, the `<datacenter-or-peer>` label is always parsed as a datacenter?

website/content/docs/discovery/dns.mdx
jkirschner-hashicorp: Note to self: `.peer` support varies by service lookup type and version.

`.virtual` has supported as of 1.14.0.
`.service` and `.node`  have support [as of 1.14.2](https://github.com/hashicorp/consul/blob/main/CHANGELOG.md#1142-november-30-2022).

`.ingress` and `.connect` lack support.

website/content/docs/discovery/dns.mdx
jkirschner-hashicorp: If I understand correctly, this still isn't supported as of 1.14.2 (CC @hashi-derek). That seems fine - this isn't the format we recommend for specifying namespace going forward. I just want to make sure my update is accurate :)

website/content/docs/discovery/dns.mdx
jkirschner-hashicorp: Gotcha. So `datacenter-or-peer` should not appear **anywhere** in these docs.

hashi-derek: Correct, unless there's an edge case that I'm not accounting for. To my knowledge, I avoided any potential for that to occur.

website/content/docs/discovery/dns.mdx
jkirschner-hashicorp: @ndhanushkodi : So if using explicit upstreams, you can use either a `.virtual` or a `.service` lookup? I guess I'm still confused about when you'd need to use a `.virtual` lookup versus just using a `.service` lookup all the time (as of 1.14.2).

jkirschner-hashicorp: Is the only time we need `.virtual` "when the downstream has tproxy enabled"? And in all other cases, use `.service`?

blake: @jkirschner-hashicorp If a downstream application is using an explicit upstream, it would be communicating to `localhost:<port>`, and would not be using the IP address returned from a Consul DNS lookup.

Accessing a virtual address _requires_ the use of tproxy. The VIPs are allocated from an IP range that is not likely to be present on a user's internal network, and thus are not routable by the upstream network. The application's sidecar needs to have tproxy enabled so that Envoy is correctly configured to map the destination [V]IP to the correct upstream service.

jkirschner-hashicorp: > If a downstream application is using an explicit upstream, it would be communicating to localhost:<port>, and would not be using the IP address returned from a Consul DNS lookup.

@blake : I don't think I had ever considered how namespace, partition, and peer are specified in an explicit upstream. Somehow I had thought that format would use a DNS-style lookup, but it doesn't. It uses `proxy.upstreams[].destination_*` fields.

I think I understand now:
- Implicit upstreams with transparent proxy must use the `.virtual` format (if using Consul DNS format rather than Kube DNS format)
- DNS formats doesn't come into the picture at all for explicit upstreams
- For service discovery DNS lookup, use `.service`.

---

Side note-to-self: We don't document `destination_peer` (if there is such a field) in the upstream config reference: https://developer.hashicorp.com/consul/docs/connect/registration/service-registration#upstream-configuration-reference

It also seems like there's duplicate content between https://developer.hashicorp.com/consul/docs/connect/registration/sidecar-service and the link above that can lead to inconsistencies, like the omission of destination namespace and partition in one place.

jkirschner-hashicorp: Though, use of this `.virtual` format requires that [`dns.enableRedirection` is set to `true`](https://developer.hashicorp.com/consul/docs/k8s/helm#v-dns-enableredirection), right?

blake: > We don't document destination_peer (if there is such a field) in the upstream config reference: https://developer.hashicorp.com/consul/docs/connect/registration/service-registration#upstream-configuration-reference

This field *is* documented on that page. At the time of writing this comment, it is documented in the third row in the table of supported parameters under [Upstream configuration reference](https://developer.hashicorp.com/consul/docs/connect/registration/service-registration#upstream-configuration-reference).

> It also seems like there's duplicate content between https://developer.hashicorp.com/consul/docs/connect/registration/sidecar-service…

This page shows the [default config options](https://developer.hashicorp.com/consul/docs/connect/registration/sidecar-service#sidecar-service-defaults) used by Consul when a service registration has Connect enabled using the minimal `"connect": { "sidecar_service": {} }` config stanza. It also shows an example of how these parameters can be overridden with minimal additional configuration.

The [/docs/connect/registration/service-registration](https://developer.hashicorp.com/consul/docs/connect/registration/service-registration) contains an exhaustive list of all parameters supported for a proxy service registration.

jkirschner-hashicorp: > This field is documented on that page. At the time of writing this comment, it is documented in the third row in the table of supported parameters under [Upstream configuration reference](https://developer.hashicorp.com/consul/docs/connect/registration/service-registration#upstream-configuration-reference).

Ugh, didn't see that. Thanks!

website/content/docs/discovery/dns.mdx
```diff
@@ -444,33 +473,46 @@ If you need more complex behavior, please use the
 
 ### Service Virtual IP Lookups
 
-To find the unique virtual IP allocated for a service:
+To find the unique virtual IP allocated for a [Connect-capable](/docs/connect) service:
```
jkirschner-hashicorp, line 476: Note to self before resubmitting PR: add cross-references to when to use virtual IP lookups back in

website/content/docs/discovery/dns.mdx
jkirschner-hashicorp: Modifying the text to explicit call out what's 1.14.0 vs 1.14.2

jkirschner-hashicorp:
@im2nguyen, @trujillo-adam: This is finally ready for education team review!

website/content/docs/discovery/dns.mdx
```diff
@@ -353,32 +393,77 @@ $ echo -n "20010db800010002cafe000000001337" | perl -ne 'printf join(":", unpack
...
+  ```text
+  [<tag>.]<service>.service[.<namespace>.ns][.<partition>.ap][.<datacenter>.dc].<domain>
+  ```
+
+  This format supports `.service`, `.connect`, `.virtual`, and `.ingress` service lookup types
+  in Consul 1.14.0 or later.
+
+  Cluster peering must be used to connect datacenters containing admin partitions,
+  not WAN federation. Therefore, this format is not applicable to lookup a partition
+  in a different datacenter.
```
trujillo-adam, line 445:
```suggestion
  You must establish a peer connection between your Consul clusters to connect datacenters containing admin partitions. You cannot query an admin partition in another datacenter over a WAN federated network. 
```

trujillo-adam (commented):
Left some comments and feedback. Let me know when you want me to have another look.

website/content/docs/discovery/dns.mdx
```diff
@@ -133,20 +133,47 @@ it is recommended to use the HTTP API to retrieve the list of nodes.
...
+  [<tag>.]<service>.service.<peer>.peer.<domain>
+  ```
+
+- Lookup a service in a WAN federated datacenter:
+
+  ```text
+  [<tag>.]<service>.service.<datacenter>[.dc].<domain>
+  ```
+
+  Use of the optional `.dc` label is recommended for readability.
```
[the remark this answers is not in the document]
jkirschner-hashicorp: It's a bit confusing. "Label" is a DNS query specific term referring to anything between two periods.

`<datacenter>` is the label that identifies the datacenter.

The `dc` part is optional and just makes it clearer to the reader that the preceding label identifies the datacenter to use in the query.

In other words...

`my-service.service.us-west1.consul` is the same as `my-service.service.us-west1.dc.consul`. The `.dc` is optional, but helps if your dc name isn't obviously the name of a dc (`us-west1` vs `dc1`).

I'm not sure how best to deal with that. Maybe just leaving as:
```suggestion
  We recommend adding the `.dc` label to improve readability.
```

website/content/docs/discovery/dns.mdx
```diff
@@ -133,20 +133,47 @@ it is recommended to use the HTTP API to retrieve the list of nodes.
...
-```
+- Lookup a local service:
+
+  ```text
+  [<tag>.]<service>.service.<domain>
+  ```
+
+  The lookup uses the datacenter of the Consul agent acting as the DNS server.
+
+- Lookup a service in a cluster peer:
```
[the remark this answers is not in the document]
jkirschner-hashicorp: Mostly. They need to specify two labels. If the peer name is "foo", they'd have to add `foo.peer`.

At a higher level, I'm wondering about the benefit of changing this text from "context in which the format is useful" to "actions you need to take to use the format below". It feels strange to me to have one of the "valid formats for standard service lookups" be titled "lookup a local service" and the next be in a different form: "add the `peer` segment to lookup a service in a cluster peer".

website/content/docs/discovery/dns.mdx
```diff
@@ -353,32 +393,77 @@ $ echo -n "20010db800010002cafe000000001337" | perl -ne 'printf join(":", unpack
...
-for `.service`, `.connect`, `.virtual`, and `.ingress` service lookup types.
-All three fields (`namespace`, `partition`, `datacenter`) are optional.
-```text
-[<tag>.]<service>.service[.<namespace>.ns][.<partition>.ap][.<datacenter>.dc]<domain>
-```
+Use the following query formats to specify namespace, partition,
+WAN federated datacenter, or cluster peer in service lookups.
+In each format, all fields in square brackets are optional.
+
+- Lookup a service, optionally in a cluster peer:
```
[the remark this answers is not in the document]
jkirschner-hashicorp: Everything in square brackets is optional, as stated in Line 412:
> In each format, all fields in square brackets are optional.

If you don't include `.peer-name.peer`, then it does a local lookup. Partition and namespace are also optional.

Therefore, I'm inclined to decline this change.

website/content/docs/discovery/dns.mdx
jkirschner-hashicorp: The limitation is not on connection / querying. WAN federation and admin partitions are mutually exclusive. If two datacenters are WAN federated, they can't have (non-default) admin partitions.

I'm not sure if that nuance is important here. I'll have to think about whether there's a clearer way than the original text...

website/content/docs/discovery/dns.mdx
```diff
@@ -183,19 +210,24 @@ foobar.node.dc1.consul. 0 IN  A 10.1.10.12
...
 
-Valid formats for RFC 2782 SRV lookups depend on
-whether you want to filter results based on a service tag:
+RFC 2782 style lookups have the same behavior as
+[standard service lookups](#standard-lookup),
+but differ in format.
+The RFC 2782 lookup format supports the same suffixes as standard lookups
+for optionally specifying a cluster peer or WAN federated datacenter,
+but differs in how service name and optional tag are specified.
+The valid RFC 2782 lookup formats below assume a local service lookup:
```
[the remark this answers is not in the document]
jkirschner-hashicorp: The challenge here is that standard and RFC 2782 lookups support all the same stuff after `.service`, but RFC 2782 lookups provide multiple ways to specify a tag. (There's an outstanding PR that will add a format not currently listed here.)

What's I'm trying to communicate is: The only difference is how `<service>` and `<tag>` can be specified. Everything else (`<datacenter>` and `<peer>`) is the same as a standard lookup.

website/content/docs/discovery/dns.mdx
```diff
@@ -353,32 +393,77 @@ $ echo -n "20010db800010002cafe000000001337" | perl -ne 'printf join(":", unpack
...
+
+- Lookup a service, optionally in a cluster peer:
+
+  ```text
+  [<tag>.]<service>.service[.<namespace>.ns][.<peer>.peer][.<partition>.ap].<domain>
+  ```
+
+  This format supports `.service` lookup types in Consul 1.14.2 or later
+  and supports `.virtual` lookup types in Consul 1.14.0 or later.
+
```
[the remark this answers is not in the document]
jkirschner-hashicorp: We still need to mention somewhere which lookup types are supported by each format. This suggested change removes mention that `.virtual` lookup types are supported.

website/content/docs/discovery/dns.mdx
```diff
@@ -353,32 +393,77 @@ $ echo -n "20010db800010002cafe000000001337" | perl -ne 'printf join(":", unpack
...
+  rather than one of its cluster peers.
+
+  For example, assume that an agent in the `default` partition is acting as the
+  DNS server, the `k8s-app-1` partition is in the same datacenter, and
+  the `k8s-app-1` partition has a cluster peer named `billing`.
+  To perform a lookup of non-mesh service `api` in namespace `foo`
+  of that cluster peer, use the query
+  `api.service.foo.ns.billing.peer.k8s-app-1.partition.consul`.
+
+- Lookup a service, optionally in a WAN federated datacenter:
```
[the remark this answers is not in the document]
jkirschner-hashicorp: I took your 2nd paragraph, but changed the first paragraph to:
```
If `.<peer>.peer` is included, the `<peer>` label identifies a cluster peer
of the query's partition in which to lookup the specified `<service>`.
If the query does not specify a partition,
the query uses the partition of the Consul agent that received the DNS query.
```

jkirschner-hashicorp:
@trujillo-adam : I considered all of the feedback. In a few cases, I decided not to make the suggested changes (and left a comment with my thinking). Let me know your follow-up thoughts - it's ready for review again. Thanks!

website/content/docs/discovery/dns.mdx
```diff
@@ -353,32 +407,81 @@ $ echo -n "20010db800010002cafe000000001337" | perl -ne 'printf join(":", unpack
...
+
+  ```text
+  [<tag>.]<service>.service[.<namespace>.ns][.<partition>.ap][.<datacenter>.dc].<domain>
+  ```
+
+  This format supports `.service`, `.connect`, `.virtual`, and `.ingress` service lookup types.
+
+  This format does not support specifying both a partition and another datacenter.
+  Datacenters containing admin partitions can only be connected using cluster peering,
+  not WAN federation. To lookup a service in a partition of another datacenter,
```
trujillo-adam, line 461:
```suggestion
   You can only use peer connections to join datacenters that contain admin partitions, not WAN federation. To lookup a service in a partition of another datacenter,
```
Active voice

website/content/docs/discovery/dns.mdx
```diff
@@ -394,6 +497,9 @@ you can use the following query formats specify namespace but not partition:
   [<tag>.]<service>.service.<namespace>.<domain>
   ```
 
+To lookup a namespaced service in a cluster peer,
+the [canonical format](#canonical-format) must be used instead.
```
trujillo-adam, line 501:
```suggestion
Use the [canonical format](#canonical-format) to lookup services bound to a namespace in a cluster peer.
```
It's a bit of a gray area, but I think "namespace" is a noun. I recall seeing it in the K8s docs as verb, but it read as poor grammar to me. "Bound" might not be the right verb, either. I won't die on this hill if you disagree.

website/content/docs/discovery/dns.mdx
```diff
@@ -444,34 +550,53 @@ If you need more complex behavior, please use the
...
 By default, a service virtual IP lookup uses the `default` namespace
 within the partition and datacenter of the Consul agent that received the DNS query.
 
-To lookup services imported from a cluster peered partition or open-source datacenter,
-specify the namespace and peer name in the lookup:
-```text
-<service>.virtual[.<namespace>].<peer>.<domain>
-```
+Use the following query formats to perform a service virtual IP lookup
+on namespaced or partitioned services:
```
trujillo-adam, line 582:
```suggestion
Use the following query formats to perform a service virtual IP lookup
on services within a namespace or partition:
```
Same comment as above about parts of speech w/r/t namespace (and partition).

website/content/docs/discovery/dns.mdx
```diff
@@ -444,34 +550,53 @@ If you need more complex behavior, please use the
...
+
+- Use the [canonical format](#canonical-format) to optionally specify
+  namespace, peer, and/or partition:
+
+  ```text
+  <service>.virtual[.<namespace>.ns][.<peer>.peer][.<partition>.ap].<domain>
+  ```
+
+- Specify cluster peer and optional namespace in the lookup:
```
trujillo-adam, line 591:
```suggestion
- Use the [canonical format](#canonical-format) to specify
  namespace, peer, and/or partition:

  ```text
  <service>.virtual[.<namespace>.ns][.<peer>.peer][.<partition>.ap].<domain>
  ```

- Specify cluster peer and namespace in the lookup:
```
Do we need to keep prefacing as these parts as optional?

Issue #14212, opened 2022-08-16
Consul Lock does not work as expected with the datacenter option


Deeksha282528:
Using -datacenter option in the consul lock to specify the datacenter throws an error "Lock acquisition failed: failed to create session: Unexpected response code: 500 (rpc error making call: rpc error making call: apply failed: Missing node registration)"

Reproduction Steps

Steps to reproduce this issue, eg:
1. consul lock test1 echo test - works totally fine. 
Output - test
2. consul lock -datacenter=xyz test2 echo test - Does not work 
Output - "Lock acquisition failed: failed to create session: Unexpected response code: 500 (rpc error making call: rpc error making call: apply failed: Missing node registration)"

huikang:
@Deeksha282528 , I tested with the latest consul version, and it works fine

```
consul lock -datacenter dc2 test2 echo test 
test

consul version
Consul v1.14.0-dev
Revision 92bcdb9a9+CHANGES
```

What version of consul did you test with?

Deeksha282528:
@huikang I have tested with 
`Consul v1.12.3`

Deeksha282528:
We have tested the  `consul lock -datacenter xyz test2 echo test ` command with the version 
`Consul v1.14.0-dev
Revision 9675faeab`

The similar issue for your reference [https://github.com/hashicorp/consul/issues/4886](url)

Please note that the command works fine when the datacenter specified is same as the datacenter that you are testing in. 
If the testing datacenter and the specified datacenter is different then we get an error 

`Lock acquisition failed: failed to create session: Unexpected response code: 500 (rpc error making call: rpc error making call: apply failed: Missing node registration)`

Deeksha282528:
Hi, Is there any update on this?

huikang:
I don't have any issue in running the same command across datacneters

two datacenters are federated

```
consul members  -wan                          
Node          Address              Status  Type    Build      Protocol  DC   Partition  Segment
server-1.dc1  192.168.1.165:8302   alive   server  1.14.0dev  2         dc1  default    <all>
server-1.dc2  192.168.1.166:8302   alive   server  1.14.0dev  2         dc2  default    <all>
```

```
 consul lock -datacenter dc2 test2 echo test 
test
```

One thing you may want to try is to create some kv pair in one dc and then check if you can read the kv pair from the other dc.

Deeksha282528:
I have tried with KV pair and it works. But the lock is not.

huikang:
[...]

PR #15394, opened 2022-11-16
fix: clarifying error message when acquiring a lock in remote dc

huikang:
### Description
When user uses `consul lock` to acquire a lock in a remote dc, if the command is run at a client agent, the returned error message is 
`Lock acquisition failed: failed to create session: Unexpected response code: 500 (rpc error making call: rpc error making call: apply failed: Missing node registration)`
which is unclear about the actual root cause. This PR updates the error message to 

`Lock acquisition failed: failed to create session: Unexpected response code: 500 (cross datacenter lock must be created at server agent)`

Please refer to the issue https://github.com/hashicorp/consul/issues/14212#issuecomment-1316215464

### Links
fix https://github.com/hashicorp/consul/issues/14212

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/agent/session_endpoint.go
+++ b/agent/session_endpoint.go
@@ -47,6 +47,10 @@ func (s *HTTPHandlers) SessionCreate(resp http.ResponseWriter, req *http.Request
 
 	fixupEmptySessionChecks(&args.Session)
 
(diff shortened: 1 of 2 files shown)


trujillo-adam (commented):
Nice improvement. I made a few suggestions.

trujillo-adam (approved):

PR #15426, opened 2022-11-17
docs: Correct a typo in checks.mdx

am-ak:
Correcting a typo under  `UDP + Interval`

### Description
The default interval is discussed earlier in the paragraph, and the context of the line with the typo is the timeout not interval.

### PR Checklist
* [ ] updated test coverage
* [ ] not a security concern


website/content/docs/discovery/checks.mdx
```diff
@@ -86,7 +86,7 @@ There are several different kinds of checks:
 - `UDP + Interval` - These checks direct the client to periodically send UDP datagrams
   to the specified IP/hostname and port. The duration specified in the `interval` field sets the amount of time 
   between attempts, such as `30s` to indicate 30 seconds. The check is logged as healthy if any response from the UDP server is received. Any other result sets the status to `critical`.
-  The default interval for, UDP checks is `10s`, but you can configure a custom UDP check timeout value by specifying the
+  The default timeout for, UDP checks is `10s`, but you can configure a custom UDP check timeout value by specifying the
```
trujillo-adam, line 89:
```suggestion
   For UDP checks, the default value for the `timeout` field is `10s`, but you can configure a custom timeout by specifying the
```
The first comma in the sentence was also incorrect. We should use this opportunity to improve the language

trujillo-adam (commented):
Thanks for working on this. We should use this opportunity to improve the wording.

website/content/docs/discovery/checks.mdx
am-ak: Thank you for reviewing. 

In this case, this whole section under "There are several different kinds of checks:"-I will refer to it as **Summary-Section**- is an incomplete duplicate of the section ["Types of checks"](https://developer.hashicorp.com/consul/docs/discovery/checks#types-of-checks), as the later includes "OSService check" that is missing in the **Summary-Section** and also includes the examples.

I would suggest updating the **Summary-Section** to be an Anchored list with one liner description, for example:

```
There are several different kinds of checks:

- Script + Interval - These checks depend on invoking an external application that performs the health check
- HTTP + Interval - These checks make an HTTP GET request to the specified URL
- TCP + Interval - These checks make a TCP connection attempt to the specified IP/hostname and port
- UDP + Interval - These checks direct the client to periodically send UDP datagrams to the specified IP/hostname and port
- OSService + Interval - These checks periodically direct the Consul agent to monitor the health of a service running on the host operating system
- Time to Live (TTL) - These checks retain their last known state for a given TTL
- Docker + Interval - These checks depend on invoking an external application which is packaged within a Docker Container
- gRPC + Interval - These checks are intended for applications that support the standard [gRPC health checking protocol](https://github.com/grpc/grpc/blob/master/doc/health-checking.md).
- H2ping + Interval - These checks test an endpoint that uses http2 by connecting to the endpoint and sending a ping frame.
- Alias - These checks alias the health state of another registered node or service.
```
With each check name i.e. "Script+Interval" being a hyperlink to the corresponding detailed section

trujillo-adam:
@am-ak - Thanks for the feedback. Let's fix this specific line for now and create a task to address the larger issues separately.

trujillo-adam (approved):

PR #15467, opened 2022-11-18
add multilimiter and tests

dhiaayachi:
### Description
This PR adds an implementation of a `multilimiter` package that will be used to perform RPC server side rate limiting.

The `MultiLimiter` struct implements the `RateLimiter` interface which look as follow:

```go
type RateLimiter interface {
	Start()
	Stop()
	Allow(entity LimitedEntity) bool
	UpdateConfig(c Config)
}
```
The `LimitedEntity` interface look as follow:

```go
type LimitedEntity interface {
	Key() []byte
}
``` 
The `MultiLimiter` is using an immutable radix tree as an underlying structure, this is suitable specifically for IP as the data is optimal for a radix tree, specially when a consecutive range of IP is used. Also using immutable radix tree help optimizing read use cases (when it's a known IP) vs write case (when a new IP is introduced).

* unit tests with 100% coverage
* benchmark tests with different type of data

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/consul/multilimiter/multilimiter.go
+++ b/agent/consul/multilimiter/multilimiter.go
@@ -8,14 +8,14 @@ import (
 	"time"
 )
 
-type limitedEntity interface {
+type LimitedEntity interface {
 	Key() []byte
 }
 
 type RateLimiter interface {
 	Start()
 	Stop()
-	Allow(entity limitedEntity) bool
+	Allow(entity LimitedEntity) bool
 	UpdateConfig(c Config)
 }
 
@@ -72,7 +72,7 @@ func (m *MultiLimiter) Stop() {
 	}
 }
 
-func (m *MultiLimiter) Allow(e limitedEntity) bool {
+func (m *MultiLimiter) Allow(e LimitedEntity) bool {
 	limiters := m.limiters.Load()
 	l, ok := limiters.Get(e.Key())
 	now := time.Now().Unix()
--- a/agent/consul/multilimiter/multilimiter.go
+++ b/agent/consul/multilimiter/multilimiter.go
@@ -1,0 +1,128 @@
+package multilimiter
+
+import (
+	"context"
+	radix "github.com/hashicorp/go-immutable-radix"
+	"golang.org/x/time/rate"
+	"sync/atomic"
+	"time"
+)
+
+type limitedEntity interface {
+	Key() []byte
+}
+
+type RateLimiter interface {
+	Start()
+	Stop()
+	Allow(entity limitedEntity) bool
+	UpdateConfig(c Config)
+}
+
+type Limiter struct {
+	limiter    *rate.Limiter
+	lastAccess atomic.Int64
+}
+
+type Config struct {
+	Rate         rate.Limit
+	Burst        int
+	CleanupLimit time.Duration
+	CleanupTick  time.Duration
+}
+
+type MultiLimiter struct {
+	limiters *atomic.Pointer[radix.Tree]
+	config   *atomic.Pointer[Config]
+	cancel   context.CancelFunc
+}
+
+func (m *MultiLimiter) UpdateConfig(c Config) {
+	m.config.CompareAndSwap(m.config.Load(), &c)
+}
+
+func NewMultiLimiter(c Config) *MultiLimiter {
+	limiters := atomic.Pointer[radix.Tree]{}
+	config := atomic.Pointer[Config]{}
+	config.Store(&c)
+	limiters.Store(radix.New())
+	if c.CleanupLimit == 0 {
+		c.CleanupLimit = 30 * time.Millisecond
+	}
+	if c.CleanupTick == 0 {
+		c.CleanupLimit = 1 * time.Second
+	}
+	m := &MultiLimiter{limiters: &limiters, config: &config}
+	return m
+}
+
+func (m *MultiLimiter) Start() {
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	m.cancel = cancelFunc
+	go func() {
+		for {
+			m.cleanupLimited(ctx)
+		}
+	}()
+}
+
+func (m *MultiLimiter) Stop() {
+	if m.cancel != nil {
+		m.cancel()
+	}
+}
+
+func (m *MultiLimiter) Allow(e limitedEntity) bool {
+	limiters := m.limiters.Load()
+	l, ok := limiters.Get(e.Key())
+	now := time.Now().Unix()
+	if ok {
+		limiter := l.(*Limiter)
+		limiter.lastAccess.Store(now)
+		return limiter.limiter.Allow()
+	}
+	c := m.config.Load()
+	limiter := &Limiter{limiter: rate.NewLimiter(c.Rate, c.Burst)}
+	limiter.lastAccess.Store(now)
+	tree, _, _ := limiters.Insert(e.Key(), limiter)
+	m.limiters.Store(tree)
+
+	return limiter.limiter.Allow()
+}
+
+// Every minute check the map for visitors that haven't been seen for
+// more than 3 minutes and delete the entries.
+func (m *MultiLimiter) cleanupLimited(ctx context.Context) {
+	c := m.config.Load()
+	waiter := time.After(c.CleanupTick)
+
+	select {
+	case <-ctx.Done():
+		return
+	case now := <-waiter:
+		limiters := m.limiters.Load()
+		storedLimiters := limiters
+		iter := limiters.Root().Iterator()
+		k, v, ok := iter.Next()
+		var txn *radix.Txn
+		for ok {
+			limiter := v.(*Limiter)
+			lastAccess := limiter.lastAccess.Load()
+			lastAccessT := time.Unix(lastAccess, 0)
+			diff := now.Sub(lastAccessT)
+
+			if diff > c.CleanupLimit {
+				if txn == nil {
+					txn = limiters.Txn()
+				}
+				txn.Delete(k)
+			}
+			k, v, ok = iter.Next()
+		}
+		if txn != nil {
+			limiters = txn.Commit()
+
+			m.limiters.CompareAndSwap(storedLimiters, limiters)
+		}
+	}
+}
--- a/agent/consul/multilimiter/multilimiter_test.go
+++ b/agent/consul/multilimiter/multilimiter_test.go
@@ -1,0 +1,135 @@
+package multilimiter
+
+import (
+	"encoding/binary"
+	"github.com/stretchr/testify/require"
+	"math/rand"
+	"testing"
+	"time"
+)
+
+type Limited struct {
+	key string
+}
+
+func (l Limited) Key() []byte {
+	return []byte(l.key)
+}
+
+func TestNewMultiLimiter(t *testing.T) {
+	c := Config{Rate: 0.1}
+	m := NewMultiLimiter(c)
+	require.NotNil(t, m)
+	require.NotNil(t, m.limiters)
+}
+
+func TestNewMultiLimiterStop(t *testing.T) {
+	c := Config{Rate: 0.1}
+	m := NewMultiLimiter(c)
+	require.NotNil(t, m)
+	require.NotNil(t, m.limiters)
+	require.Nil(t, m.cancel)
+	m.Stop()
+	require.Nil(t, m.cancel)
+	m.Start()
+	require.NotNil(t, m.cancel)
+	m.Stop()
+	m.Stop()
+
+}
+
+func TestRateLimiterUpdate(t *testing.T) {
+	c := Config{Rate: 0.1, CleanupLimit: 1 * time.Millisecond, CleanupTick: 10 * time.Millisecond}
+	m := NewMultiLimiter(c)
+	m.Allow(Limited{key: "test"})
+	limiters := m.limiters.Load()
+	l1, ok1 := limiters.Get([]byte("test"))
+	require.True(t, ok1)
+	require.NotNil(t, l1)
+	la1 := l1.(*Limiter).lastAccess.Load()
+	m.Allow(Limited{key: "test"})
+	limiters = m.limiters.Load()
+	l2, ok2 := limiters.Get([]byte("test"))
+	require.True(t, ok2)
+	require.NotNil(t, l2)
+	require.Equal(t, l1, l2)
+	la2 := l1.(*Limiter).lastAccess.Load()
+	require.Equal(t, la1, la2)
+
+}
+
+func TestRateLimiterCleanup(t *testing.T) {
+	c := Config{Rate: 0.1, CleanupLimit: 1 * time.Millisecond, CleanupTick: 10 * time.Millisecond}
+	m := NewMultiLimiter(c)
+	m.Start()
+	m.Allow(Limited{key: "test"})
+	limiters := m.limiters.Load()
+	l, ok := limiters.Get([]byte("test"))
+	require.True(t, ok)
+	require.NotNil(t, l)
+	time.Sleep(20 * time.Millisecond)
+	limiters = m.limiters.Load()
+	l, ok = limiters.Get([]byte("test"))
+	require.False(t, ok)
+	require.Nil(t, l)
+	m.Stop()
+	m.Allow(Limited{key: "test"})
+	time.Sleep(20 * time.Millisecond)
+	limiters = m.limiters.Load()
+	l, ok = limiters.Get([]byte("test"))
+	require.True(t, ok)
+	require.NotNil(t, l)
+}
+
+func TestRateLimiterUpdateConfig(t *testing.T) {
+	c := Config{Rate: 0.1, CleanupLimit: 1 * time.Millisecond, CleanupTick: 10 * time.Millisecond}
+	m := NewMultiLimiter(c)
+	require.Equal(t, *m.config.Load(), c)
+	c2 := Config{Rate: 1, CleanupLimit: 10 * time.Millisecond, CleanupTick: 100 * time.Millisecond}
+	m.UpdateConfig(c2)
+	require.Equal(t, *m.config.Load(), c2)
+}
+
+type ipLimited struct {
+	key []byte
+}
+
+func (i ipLimited) Key() []byte {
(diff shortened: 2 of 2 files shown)


agent/consul/multilimiter/multilimiter.go
```diff
@@ -0,0 +1,128 @@
...
+	limiter := &Limiter{limiter: rate.NewLimiter(c.Rate, c.Burst)}
+	limiter.lastAccess.Store(now)
+	tree, _, _ := limiters.Insert(e.Key(), limiter)
+	m.limiters.Store(tree)
+
+	return limiter.limiter.Allow()
+}
+
+// Every minute check the map for visitors that haven't been seen for
+// more than 3 minutes and delete the entries.
```
jmurret, line 94:
```suggestion
// more than the CleanupLimit and delete the entries.
```

agent/consul/multilimiter/multilimiter.go
```diff
@@ -0,0 +1,128 @@
...
+type Limiter struct {
+	limiter    *rate.Limiter
+	lastAccess atomic.Int64
+}
+
+type Config struct {
+	Rate         rate.Limit
+	Burst        int
+	CleanupLimit time.Duration
+	CleanupTick  time.Duration
```
jmurret, line 31: Is this the the Interval which the routine checks to see if anything needs to be cleaned up?  If so, can we call it something like `CleanupCheckInterval`?

agent/consul/multilimiter/multilimiter.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+type MultiLimiter struct {
+	limiters *atomic.Pointer[radix.Tree]
+	config   *atomic.Pointer[Config]
+	cancel   context.CancelFunc
+}
+
+func (m *MultiLimiter) UpdateConfig(c Config) {
```
jmurret, line 40: I thought you mentioned this was going away?

agent/consul/multilimiter/multilimiter.go
```diff
@@ -0,0 +1,128 @@
...
+
+import (
+	"context"
+	radix "github.com/hashicorp/go-immutable-radix"
+	"golang.org/x/time/rate"
+	"sync/atomic"
+	"time"
+)
+
+type LimitedEntity interface {
```
jmurret, line 11: Nice. We can iterate/add properties to this in later phases.  Then in the case of limiting on IP, tenant, or service, we could add properties for things like LimitType (IP limiting vs tenant limiting vs service limiting vs global limiting) and Scope (IP1 vs IP2 or Tenant1 vs Tenant2).  So, basically we don't need to figure out that structure entirely yet and we can iterate this interface without changing the signature of MultiLimiter's `Allow()`?

agent/consul/multilimiter/multilimiter.go
```diff
@@ -0,0 +1,128 @@
+package multilimiter
```
jmurret, line 1: docstrings would be great when you get a moment.

agent/consul/multilimiter/multilimiter.go
```diff
@@ -0,0 +1,128 @@
...
+	l, ok := limiters.Get(e.Key())
+	now := time.Now().Unix()
+	if ok {
+		limiter := l.(*Limiter)
+		limiter.lastAccess.Store(now)
+		return limiter.limiter.Allow()
+	}
+	c := m.config.Load()
+	limiter := &Limiter{limiter: rate.NewLimiter(c.Rate, c.Burst)}
+	limiter.lastAccess.Store(now)
```
jmurret, line 86: I'm a little fuzzy on how this works, but I think we have to coordinate concurrency for writes but not reads to the immutable radix tree?  Do we need to implement a locking technique to ensure there is only one write at a time?
- multiple concurrent `Allow()` calls updating lastAccess from concurrent requests
- go routine to clean up
- Update config on a hot reload that will have to update many and possible all limiters in the radix tree?

jmurret (commented):
This is really cool.  Thank you for posting this early so we can see where this is headed.  

I left some comments and questions.  Let me know if there is anything I can help research if I brought up something (such as I am not sure if we need to ensure coordinate writes to the immutable-radix tree) ourselves or we get that for free.

agent/consul/multilimiter/multilimiter.go
jmurret: I think, we will have to deal with transaction management.  [MemDB here ](https://github.com/hashicorp/go-memdb/blob/main/txn.go#L148-L166)copies the root tree, updates and swaps out the root all the while it has a Lock and then releases the lock.  It guarantees one writer.  We'll have to do the same...or possible horrible idea - use memdb to store limiters?  🤔

jmurret: I did a quick POC of using memdb [here](https://github.com/hashicorp/consul/pull/15474) that is at least knowing what the benchmarking between the two approaches looks like.

dhiaayachi: So that's the beauty of using immutable data structures every write (or commited txn)create a new copy of the data structure without altering the old one. Which make synchronization unnecessary until you need to swap the pointer .
This is actually the technique I was hinting to when we discussed this last week.

dhiaayachi: atomic.Pointer was introduced in go 1.19 and it does pointer swapping atomically, so no need to lock here.

agent/consul/multilimiter/multilimiter.go
jmurret: Would that be problematic when configuration has changed but the server/ip/tenant/service is read more often than the CleanupLimit?   It seems like it would never update while this is true.

agent/consul/multilimiter/multilimiter.go
jmurret: 👍 that's great.  Thank you for the context!  This was a helpful thought exercise for me. I'll leave this here for other newbies, but we should consider this resolved.

agent/consul/multilimiter/multilimiter.go
dhiaayachi: yes I thought about it more and I think the 2 options we have are:
- update when allow is called
- update out of band in the clean up loop
I implemented the first option but I think we need to rework it as this mean that on config update all the new requests will go the slow path which is not ideal. 
Ideally we will do it out of band to avoid having rate limited requests go through  the slow path.
I think this can be done later, as for global limiter it's a no issue (we will only have 1 limiter for all the connections).

boxofrad (changes requested):
This is amazing, nice job! I really appreciate the careful thought (and benchmarking) you've put into the performance characteristics of this - using an atomically-updated immutable radix tree is inspired 🙌🏻 

I've left a couple of comments for your consideration. The lifecycle management one is the main thing I'd appreciate your thoughts on.

agent/consul/multilimiter/multilimiter.go
```diff
@@ -0,0 +1,158 @@
...
+
+	return limiter.limiter.Allow()
+}
+
+// cleanupLimitedOnce is called by the MultiLimiter clean up routine to remove old Limited entries
+// it will wait for CleanupCheckInterval before traversing the radix tree and removing all entries
+// with lastAccess > CleanupCheckLimit
+func (m *MultiLimiter) cleanupLimitedOnce(ctx context.Context) {
+	c := m.config.Load()
+	waiter := time.After(c.CleanupCheckInterval)
```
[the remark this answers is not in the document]
dhiaayachi: good to know! Thank you

agent/consul/multilimiter/multilimiter.go
```diff
@@ -0,0 +1,148 @@
...
+// which rate limiter to use to allow the request
+type LimitedEntity interface {
+	Key() []byte
+}
+
+// Limiter define a limiter to be part of the MultiLimiter structure
+type Limiter struct {
+	limiter    *rate.Limiter
+	lastAccess atomic.Int64
+	config     atomic.Pointer[LimiterConfig]
```
boxofrad, line 37: I've just realised that `config` applies to **all** limiters/keys. Does that mean we'll have separate `MultiLimiter`s for global read, global write, ip-based read and ip-based write?

This seems counter to the following diagram from the RFC:

image

boxofrad (approved):
Thanks for incorporating my feedback ⭐ 

I've left one last question, as I've realised that I don't fully understand how the configuration works.

agent/consul/multilimiter/multilimiter.go
dhiaayachi: Yes, That part of the design need to change IMO. I will update the RFC. Because the first Key (operation.read/write) is a finite number I think it's overkilled to embedded into the `MultiLimiter` and we can instead have a `map[string]MultiLimiter` created when bootstrapping or updating the config. What you think about that?

jmurret: I guess I was imaging that the radix tree was storing all limiters and that the keys would be based on the properties defined in the diagram.  Those properties being (rough names used here):
- Operation Category - GLOBAL, KVS, CATALOG, etc. (transaction type currently part of brown "column" in the diagram)
- Operation Type - WRITE,READ (transaction type currently part of brown "column" in the diagram)
- Limit Type - server/global/all, IP, tenant, service (this prevents name/key collisions.   corresponds to the phases of the implementation)
- Scope - global, 10.1.1.11, tenant-a, checkout-service, etc (this is all the possible values across all phases for the yellow column in the diagram)

So naively when I looked at LimitedEntity I imagined keys would look like this:
 ```
/<limit type>/<scope>/<category>/<operation type>
EX:
/server/all/global/read- phase 1
/ip/10.1.1.11/global/write/ - phase 2
/ip/10.1.1.11/catalog/write/ - phase 2
/tenant/tenant-a/global/write/ - phase 3
/service/checkout-service/global/write/ - phase 4
```

The above was really based on the LimitedEntity and how might iterate on it through phases, but possibly the above has storage concerns or other things.

Comment (jmurret) [reply agent/consul/multilimiter/multilimiter.go:37]: I also wonder if having `map[string]MultiLimiter` would cause us to use locking and atomic operations for this as well?

Comment (boxofrad) [reply agent/consul/multilimiter/multilimiter.go:37]: Yeah, I had assumed the same 😅

Without storing configuration on the tree leaf nodes, we'd need to track the identity/tenant-specific configuration in some other structure - which seems like a shame when you've put so much effort into making this one efficient!

File: agent/consul/multilimiter/multilimiter.go [hunk 9]
```diff
@@ -0,0 +1,188 @@
...
+	var config LimiterConfig
+	if okP {
+		config = c.(LimiterConfig)
+	} else {
+		config = m.config.Load().LimiterConfig
+	}
+	limiter := &Limiter{limiter: rate.NewLimiter(config.Rate, config.Burst), config: &atomic.Pointer[LimiterConfig]{}}
+	limiter.config.Store(&config)
+	limiter.lastAccess.Store(now)
+	tree, _, _ := limiters.Insert(e.Key(), limiter)
```
boxofrad, line 138: As an aside/not something we need to change now: I think the structure of the tree will need to change slightly in the future to accommodate setting config directly on a leaf-node (e.g. for a specific identity/tenant).

I guess we'd store a leaf struct containing the config and limiter? We can figure that out when we get to it, though 😄

dhiaayachi: yes that was one of the option I thought about, I think it make sense to make the tree store:
```go
type not_sure_about_the_name struct {
    config *LimiterConfig
    limiter *Limiter
} 
```
and in the cleanup routine (I think at this point we need to renamed to reconcile), we check if any is nil, instead of the type check.

agent/consul/multilimiter/multilimiter.go
```diff
@@ -0,0 +1,188 @@
...
+				if diff > c.CleanupCheckLimit {
+					if txn == nil {
+						txn = limiters.Txn()
+					}
+					txn.Delete(k)
+				}
+				if *limiter.config.Load() != config {
+					// update the limiter config
+					limiter.config.Store(&config)
+				}
```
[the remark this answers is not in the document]
dhiaayachi: yes but than what config should I set to XXX? I'm trying to avoid recursivity but it seems there is no way around it.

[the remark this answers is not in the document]
boxofrad: Yeah, I guess you could default to an unlimited rate-limiter in that situation (`rate.Inf`)?

dhiaayachi:
@boxofrad I changed the approach to update the config, please take a look and let me know if you would like to chat about it?

dhiaayachi:
@boxofrad I worked on the 2 versions we discussed:

- lock on write in `Allow` to prevent concurrent writes to the Tree #15642 
- add an extra go routine to perform writes #15643 

I also added an extra bit in #15643 that would buffer multiple writes into a transaction and commit them once after 10ms. That improve considerably the performances. I also added an update if we find an existing Limiter in the transaction, that way if multiple limiters are sent in the channel for the same key we will update the first one.

Let me know what you think about that, maybe we could pair on it again! 😃

agent/consul/multilimiter/multilimiter.go
```diff
@@ -0,0 +1,217 @@
...
+
+// UpdateConfig will update the MultiLimiter Config
+// which will cascade to all the Limiter(s) LimiterConfig
+func (m *MultiLimiter) UpdateConfig(c LimiterConfig, prefix []byte) {
+	if prefix == nil {
+		prefix = []byte("")
+	}
+	configs := m.limitersConfigs.Load()
+	newConfigs, _, _ := configs.Insert(prefix, &c)
+	m.limitersConfigs.Store(newConfigs)
```
boxofrad, line 74: Just leaving a note so we don't forget to add a lock here, to protect against concurrent calls to `UpdateConfig` overwriting each other's changes 🔒

dhiaayachi:
> Awesome job @dhiaayachi, thanks 🙌🏻 
> 
> It looks like the lock-based approach performs better overall, which I think makes sense because:
> 1. Sending on a channel [takes a lock](https://github.com/golang/go/blob/6a70292d1cb3464e5b2c2c03341e5148730a1889/src/runtime/chan.go#L202) anyway and if the channel is full the sender (currently) blocks.
> 2.  #15643 asynchronously updates the tree, so we'll fall into the slow-path (and wastefully re-create the limiter) many times.
> 
> I really like the channel-based approach, but I think the lock-based approach is possibly easier to reason about because its writes are applied to the tree immediately.

@boxofrad I'm not sure I understand right. The channel based approach is faster so far based on my benchmarks (I need to rerun those after fixing the waiter issue). The reason behind that is that it give us the opportunity to buffer writes in a transaction before committing which make the insert much more performant. Also I added some code in that approach that when we already have another limiter on the key we update it instead of rewriting it.
About blocking, my understanding is that the channel based approach would block the same as a lock based approach in the worst case so I don't see why we need to drop limiters specifically for that approach and not do it for the lock based one.

agent/consul/multilimiter/multilimiter.go
```diff
@@ -0,0 +1,277 @@
...
+
+func (t tickerWrapper) Ticker() <-chan time.Time {
+	return t.ticker.C
+}
+
+func (m *MultiLimiter) runStoreOnce(ctx context.Context, waiter ticker, txn *radix.Txn) *radix.Txn {
+	select {
+	case <-waiter.Ticker():
+		tree := txn.Commit()
+		m.limiters.Store(tree)
```
boxofrad, line 200: I think we need to make this a `CompareAndSwap` operation (like we do in `reconcileLimitedOnce`).

My thinking is that if the reconciliation loop takes a long time, there's a pretty good likelihood of the store/commit loop undoing its changes.

```
R = Reconciliation loop
S = Store loop
# = Transaction in-flight

      0ms   10ms   20ms   30ms   40ms   50ms
+---+----------------------------------------+
| R |        |###################|           |
+---+----------------------------------------+
| S |                      |########|        |
+---+----------------------------------------+
```

In the pathological case, the right pattern of load could prevent us from _ever_ re-configuring or cleaning up limiters.

What do you think @dhiaayachi?

boxofrad: Actually thinking about it a little more, `reconcileLimitedOnce` taking a long time means it'll probably end up failing at the `CompareAndSwap` frequently anyway because the tree would've been overwritten 🤔 

I wonder if we should accept the performance overhead of just not using transactions. We don't need them for atomicity, for example.

jmurret (approved):
Looks great. Thank you for all of this hard work and discussion.  I sent a separate slack about cleaning up the comments with two of them being easy wins and the rest most likely moved to additional follow up tasks.  Let's move this PR forward and move off the conservations to follow up tasks.

agent/consul/multilimiter/multilimiter.go
dhiaayachi: This is not needed anymore as config and limiters are now separate trees which make that case natively supported

agent/consul/multilimiter/multilimiter.go
```diff
@@ -0,0 +1,188 @@
...
+			m.cleanupLimitedOnce(ctx)
+		}
+	}()
+}
+
+// todo: split without converting to a string
+func splitKey(key []byte) ([]byte, []byte) {
+
+	s := strings.Split(string(key), string(separator))
+	if len(s) >= 2 {
```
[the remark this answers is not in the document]
dhiaayachi: This is done I added `Key` as an exported function for this

agent/consul/multilimiter/multilimiter.go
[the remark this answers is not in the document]
dhiaayachi: This is also fixed know that we don't loop anymore and go directly to the prefix to find a config. If a config was not found in the prefix we default to the default config, which could be Infinite.

agent/consul/multilimiter/multilimiter.go
dhiaayachi: We discussed this with @boxofrad and we agreed that we don't need to fully solve this but try to reduce the surface/impact of this happening. As config updates and clean-up would happen periodically and would end up converging under the right circonstances. I will start another PR with a way to do that after this is merged.

PR #15478, opened 2022-11-18
docs: clarify Vault CA provider permissions needed

jkirschner-hashicorp:
After this PR is merged, the associated tutorials that include Vault policies for use with the Vault as a Connect CA provider must be updated as well.

[Preview link to the Vault policies changes](https://consul-n4pht8hnd-hashicorp.vercel.app/consul/docs/connect/ca/vault#vault-acl-policies)

### DIFF
--- a/website/content/api-docs/connect/ca.mdx
+++ b/website/content/api-docs/connect/ca.mdx
@@ -169,6 +169,18 @@ The table below shows this endpoint's support for
 
 The corresponding CLI command is [`consul connect ca set-config`](/commands/connect/ca#set-config).
 
+~> **If currently using Vault CA provider:**
+   If you intend to change the CA provider from Vault to another,
+   or to change the Vault provider's [`RootPKIPath`](/docs/connect/ca/vault#rootpkipath),
+   you must temporarily provide the Vault
+   [auth method](/docs/connect/ca/vault#authmethod) or [token](/docs/connect/ca/vault#token)
+   with `sudo + update` permissions on the extremely privileged
+   [`<insert-RootPKIPath>/root/sign-self-issued` Vault endpoint](https://developer.hashicorp.com/vault/api-docs/secret/pki#sign-self-issued)
+   to enable the root cross-signing operation.
+   Refer to the [Vault CA provider policy examples](/consul/docs/connect/ca/vault#vault-acl-policies)
+   for more information. We recommend removing these elevated permissions
+   immediately after this HTTP API call completes.
+
 ### JSON Request Body Schema
 
 - `Provider` `(string: <required>)` - Specifies the CA provider type to use.
@@ -177,11 +189,14 @@ The corresponding CLI command is [`consul connect ca set-config`](/commands/conn
   for the chosen provider. For more information on configuring the Connect CA
   providers, see [Provider Config](/docs/connect/ca).
 
-- `ForceWithoutCrossSigning` `(bool: <optional>)` - Indicates that the CA change
-  should be forced to complete even if the current CA doesn't support cross
-  signing. Changing root without cross-signing may cause temporary connection
-  failures until the rollout completes. See [Forced Rotation Without
-  Cross-Signing](/docs/connect/ca#forced-rotation-without-cross-signing)
+- `ForceWithoutCrossSigning` `(bool: false)` - Indicates that the CA change
+  should be forced to complete even if the current CA doesn't support root cross-signing.
+
+  ~> **Caution:** Setting this field to `true` will cause temporary connection failures
+  until service mesh proxies and/or Consul client agents receive a new certificate
+  that establishes trust with the new root.
+  Do not use this field unless you are sure you need it.
+  Refer to [Forced Rotation Without Cross-Signing](/docs/connect/ca#forced-rotation-without-cross-signing)
   for more detail.
 
 ### Sample Payload
--- a/website/content/api-docs/connect/ca.mdx
+++ b/website/content/api-docs/connect/ca.mdx
@@ -172,14 +172,9 @@ The corresponding CLI command is [`consul connect ca set-config`](/commands/conn
 ~> **If currently using Vault CA provider:**
    If you intend to change the CA provider from Vault to another,
    or to change the Vault provider's [`RootPKIPath`](/docs/connect/ca/vault#rootpkipath),
-   you must temporarily provide the Vault
-   [auth method](/docs/connect/ca/vault#authmethod) or [token](/docs/connect/ca/vault#token)
-   with `sudo + update` permissions on the extremely privileged
-   [`<root_pki_path>/root/sign-self-issued` Vault endpoint](/vault/api-docs/secret/pki#sign-self-issued)
-   to enable the root cross-signing operation.
-   Refer to the [Vault CA provider policy examples](/consul/docs/connect/ca/vault#vault-acl-policies)
-   for more information. We recommend removing these elevated permissions
-   immediately after this HTTP API call completes.
+   you must temporarily elevate the privileges of the Vault token
+   or auth method in use as described in the
+   [Vault CA provider documentation](/docs/connect/ca/vault#additional-vault-acl-policies-for-sensitive-operations).
 
 ### JSON Request Body Schema
 
--- a/website/content/api-docs/connect/ca.mdx
+++ b/website/content/api-docs/connect/ca.mdx
@@ -175,7 +175,7 @@ The corresponding CLI command is [`consul connect ca set-config`](/commands/conn
    you must temporarily provide the Vault
    [auth method](/docs/connect/ca/vault#authmethod) or [token](/docs/connect/ca/vault#token)
    with `sudo + update` permissions on the extremely privileged
-   [`<insert-RootPKIPath>/root/sign-self-issued` Vault endpoint](https://developer.hashicorp.com/vault/api-docs/secret/pki#sign-self-issued)
+   [`<insert-RootPKIPath>/root/sign-self-issued` Vault endpoint](/vault/api-docs/secret/pki#sign-self-issued)
    to enable the root cross-signing operation.
    Refer to the [Vault CA provider policy examples](/consul/docs/connect/ca/vault#vault-acl-policies)
    for more information. We recommend removing these elevated permissions
--- a/website/content/api-docs/connect/ca.mdx
+++ b/website/content/api-docs/connect/ca.mdx
@@ -175,7 +175,7 @@ The corresponding CLI command is [`consul connect ca set-config`](/commands/conn
    you must temporarily provide the Vault
    [auth method](/docs/connect/ca/vault#authmethod) or [token](/docs/connect/ca/vault#token)
    with `sudo + update` permissions on the extremely privileged
-   [`<insert-RootPKIPath>/root/sign-self-issued` Vault endpoint](/vault/api-docs/secret/pki#sign-self-issued)
+   [`<root_pki_path>/root/sign-self-issued` Vault endpoint](/vault/api-docs/secret/pki#sign-self-issued)
    to enable the root cross-signing operation.
    Refer to the [Vault CA provider policy examples](/consul/docs/connect/ca/vault#vault-acl-policies)
    for more information. We recommend removing these elevated permissions
--- a/website/content/commands/connect/ca.mdx
+++ b/website/content/commands/connect/ca.mdx
@@ -97,6 +97,18 @@ Configuration updated!
 
 The return code will indicate success or failure.
 
+~> **If currently using Vault CA provider:**
+   If you intend to change the CA provider from Vault to another,
+   or to change the Vault provider's [`RootPKIPath`](/docs/connect/ca/vault#rootpkipath),
+   you must temporarily provide the Vault
+   [auth method](/docs/connect/ca/vault#authmethod) or [token](/docs/connect/ca/vault#token)
+   with `sudo + update` permissions on the extremely privileged
+   [`<insert-RootPKIPath>/root/sign-self-issued` Vault endpoint](https://developer.hashicorp.com/vault/api-docs/secret/pki#sign-self-issued)
+   to enable the root cross-signing operation.
+   Refer to the [Vault CA provider policy examples](/consul/docs/connect/ca/vault#vault-acl-policies)
+   for more information. We recommend removing these elevated permissions
+   immediately after this command exits.
+
 #### Command Options
 
 - `-config-file` - (required) Specifies a JSON-formatted file to use for the new configuration.
@@ -104,10 +116,13 @@ The return code will indicate success or failure.
   [Update CA Configuration API](/api-docs/connect/ca#update-ca-configuration).
 
 - `-force-without-cross-signing` `(bool: <optional>)` - Indicates that the CA change
-  should be forced to complete even if the current CA doesn't support cross
-  signing. Changing root without cross-signing may cause temporary connection
-  failures until the rollout completes. See [Forced Rotation Without
-  Cross-Signing](/docs/connect/ca#forced-rotation-without-cross-signing)
+  should be forced to complete even if the current CA doesn't support root cross-signing.
+
+  ~> **Caution:** Use of this flag will cause temporary connection failures
+  until service mesh proxies and/or Consul client agents receive a new certificate
+  that establishes trust with the new root.
+  Do not use this flag unless you are sure you need it.
+  Refer to [Forced Rotation Without Cross-Signing](/docs/connect/ca#forced-rotation-without-cross-signing)
   for more detail.
 
 #### API Options
--- a/website/content/commands/connect/ca.mdx
+++ b/website/content/commands/connect/ca.mdx
@@ -103,7 +103,7 @@ The return code will indicate success or failure.
    you must temporarily provide the Vault
    [auth method](/docs/connect/ca/vault#authmethod) or [token](/docs/connect/ca/vault#token)
    with `sudo + update` permissions on the extremely privileged
-   [`<insert-RootPKIPath>/root/sign-self-issued` Vault endpoint](https://developer.hashicorp.com/vault/api-docs/secret/pki#sign-self-issued)
+   [`<insert-RootPKIPath>/root/sign-self-issued` Vault endpoint](/vault/api-docs/secret/pki#sign-self-issued)
    to enable the root cross-signing operation.
    Refer to the [Vault CA provider policy examples](/consul/docs/connect/ca/vault#vault-acl-policies)
    for more information. We recommend removing these elevated permissions
--- a/website/content/commands/connect/ca.mdx
+++ b/website/content/commands/connect/ca.mdx
@@ -100,14 +100,9 @@ The return code will indicate success or failure.
 ~> **If currently using Vault CA provider:**
    If you intend to change the CA provider from Vault to another,
    or to change the Vault provider's [`RootPKIPath`](/docs/connect/ca/vault#rootpkipath),
-   you must temporarily provide the Vault
-   [auth method](/docs/connect/ca/vault#authmethod) or [token](/docs/connect/ca/vault#token)
-   with `sudo + update` permissions on the extremely privileged
-   [`<root_pki_path>/root/sign-self-issued` Vault endpoint](/vault/api-docs/secret/pki#sign-self-issued)
-   to enable the root cross-signing operation.
-   Refer to the [Vault CA provider policy examples](/consul/docs/connect/ca/vault#vault-acl-policies)
-   for more information. We recommend removing these elevated permissions
-   immediately after this command exits.
+   you must temporarily elevate the privileges of the Vault token
+   or auth method in use as described in the
+   [Vault CA provider documentation](/docs/connect/ca/vault#additional-vault-acl-policies-for-sensitive-operations).
 
 #### Command Options
 
--- a/website/content/commands/connect/ca.mdx
+++ b/website/content/commands/connect/ca.mdx
@@ -103,7 +103,7 @@ The return code will indicate success or failure.
    you must temporarily provide the Vault
    [auth method](/docs/connect/ca/vault#authmethod) or [token](/docs/connect/ca/vault#token)
    with `sudo + update` permissions on the extremely privileged
-   [`<insert-RootPKIPath>/root/sign-self-issued` Vault endpoint](/vault/api-docs/secret/pki#sign-self-issued)
+   [`<root_pki_path>/root/sign-self-issued` Vault endpoint](/vault/api-docs/secret/pki#sign-self-issued)
    to enable the root cross-signing operation.
    Refer to the [Vault CA provider policy examples](/consul/docs/connect/ca/vault#vault-acl-policies)
    for more information. We recommend removing these elevated permissions
--- a/website/content/docs/connect/ca/vault.mdx
+++ b/website/content/docs/connect/ca/vault.mdx
@@ -210,48 +210,79 @@ The following Vault policy allows Consul to use pre-existing PKI paths in Vault.
 Consul is granted read-only access to the PKI mount points and the Root CA, but is
 granted full control of the Intermediate or Leaf CA for Connect clients.
 
-In this example the `RootPKIPath` is `connect_root` and the `IntermediatePKIPath`
-is `connect_inter`. These values should be updated for your environment.
+This example uses placeholder values for `RootPKIPath` and `IntermediatePKIPath`.
+Please update those placeholders to match the path values in the CA provider configuration.
 
 <CodeBlockConfig filename="vault-managed-pki-policy.hcl">
 
 ```hcl
-# Existing PKI Mounts
-path "/sys/mounts" {
+# -----------------------------------------------------------------------------
+# Vault Managed PKI Mounts
+# -----------------------------------------------------------------------------
+
+# -----------------------------------------------------------------------------
+# Needed for Consul to read the PKI mounts already created in Vault
+
+path "/sys/mounts/<insert-RootPKIPath>" {
   capabilities = [ "read" ]
 }
 
-path "/sys/mounts/connect_root" {
+path "/sys/mounts/<insert-IntermediatePKIPath>" {
   capabilities = [ "read" ]
 }
 
-path "/sys/mounts/connect_inter" {
-  capabilities = [ "read" ]
-}
-
-# Needed for Consul 1.11+
-path "/sys/mounts/connect_inter/tune" {
+# Needed in Consul 1.11+ to update intermediate mount configuration
+path "/sys/mounts/<insert-IntermediatePKIPath>/tune" {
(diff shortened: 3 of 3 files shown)


website/content/docs/connect/ca/vault.mdx
```diff
@@ -210,30 +210,51 @@ The following Vault policy allows Consul to use pre-existing PKI paths in Vault.
...
 path "/sys/mounts" {
   capabilities = [ "read" ]
 }
 
-path "/sys/mounts/connect_root" {
+# -----------------------------------------------------------------------------
+# Needed for Consul to renew its Vault token (if renewable) whether provided
+# directly in the CA provider config or using an auth method.
+
+path "auth/token/renew-self" {
```
jkirschner-hashicorp, line 231: Question for reviewer: do we think we should move the auth token renewal stuff to the bottom of the list of permissions? Maybe the most important stuff is the CA stuff itself.

(I think I agree with this. Will probably move to bottom unless told otherwise.)

website/content/docs/connect/ca/vault.mdx
```diff
@@ -210,30 +210,51 @@ The following Vault policy allows Consul to use pre-existing PKI paths in Vault.
...
-In this example the `RootPKIPath` is `connect_root` and the `IntermediatePKIPath`
-is `connect_inter`. These values should be updated for your environment.
+This example uses placeholder values for `RootPKIPath` and `IntermediatePKIPath`.
+Please update those placeholders to match the path values in the CA provider configuration.
 
 <CodeBlockConfig filename="vault-managed-pki-policy.hcl">
 
 ```hcl
-# Existing PKI Mounts
+# -----------------------------------------------------------------------------
```
jkirschner-hashicorp, line 219: Note to self: I'm still missing the guidance on what permissions are needed to change the provider away from the Vault CA provider.

That should probably be mentioned in the CLI/HTTP API docs as well... I'll address that in a future commit.

website/content/docs/connect/ca/vault.mdx
```diff
@@ -210,30 +210,51 @@ The following Vault policy allows Consul to use pre-existing PKI paths in Vault.
...
+# Vault Managed PKI Mounts
+# -----------------------------------------------------------------------------
+
 path "/sys/mounts" {
```
jkirschner-hashicorp, line 223: As of [this PR](https://github.com/hashicorp/consul/pull/12655/files#diff-7bf30611a760296e2e6ffcd7fe955b0eb8317fe0a43c470b8b52d311af4a2515L359), I don't think we actually need read capabilities on `sys/mounts` anymore... There's no other call to `Sys().ListMounts()` as far as I can tell. (We should actually test this first.)

Can we remove that from the docs? Or at least mark the version upon which it is no longer necessary?

jkirschner-hashicorp: It looks like this change was made in Consul 1.12.0: https://github.com/hashicorp/consul/blob/main/CHANGELOG.md#1120-april-20-2022

website/content/docs/connect/ca/vault.mdx
```diff
@@ -210,48 +210,79 @@ The following Vault policy allows Consul to use pre-existing PKI paths in Vault.
...
 
 path "auth/token/lookup-self" {
-	capabilities = [ "read" ]
+  capabilities = [ "read" ]
+}
+
+# -----------------------------------------------------------------------------
+# Deprecated
+
+# Only needed before Consul 1.12
```
jkirschner-hashicorp, line 283: As of [this PR](https://github.com/hashicorp/consul/pull/12655/files#diff-7bf30611a760296e2e6ffcd7fe955b0eb8317fe0a43c470b8b52d311af4a2515L359), I don't think we actually need read capabilities on sys/mounts anymore... There's no other call to Sys().ListMounts() as far as I can tell. (We should actually test this first.)


It looks like this change was made in Consul 1.12.0: https://github.com/hashicorp/consul/blob/main/CHANGELOG.md#1120-april-20-2022

website/content/docs/connect/ca/vault.mdx
```diff
@@ -210,48 +210,79 @@ The following Vault policy allows Consul to use pre-existing PKI paths in Vault.
...
+# to either:
+# 1. A different CA provider (such as Consul's built-in provider)
+# 2. Changing the RootPKIPath used by the Vault CA provider
+#
+# SECURITY NOTE: Because this is an extremely privileged endpoint,
+# We recommend only temporarily granting this permission if there's a need
+# to modify Consul's CA provider configuration as in (1) or (2).
+# That's why this permission is currently commented out.
+#
+# path "<insert-RootPKIPath>/root/sign-self-issued" {
```
jkirschner-hashicorp, line 264: @rboyer : This is my attempt to strike a balance between documenting this necessary permission while optimizing for security.

I also reference this from the "modify CA" CLI and HTTP API docs.

website/content/commands/connect/ca.mdx
```diff
@@ -97,17 +97,32 @@ Configuration updated!
 
 The return code will indicate success or failure.
 
+~> **If currently using Vault CA provider:**
+   If you intend to change the CA provider from Vault to another,
+   or to change the Vault provider's [`RootPKIPath`](/docs/connect/ca/vault#rootpkipath),
+   you must temporarily provide the Vault
+   [auth method](/docs/connect/ca/vault#authmethod) or [token](/docs/connect/ca/vault#token)
+   with `sudo + update` permissions on the extremely privileged
+   [`<insert-RootPKIPath>/root/sign-self-issued` Vault endpoint](https://developer.hashicorp.com/vault/api-docs/secret/pki#sign-self-issued)
```
jkirschner-hashicorp, line 106:
```suggestion
   [`<insert-RootPKIPath>/root/sign-self-issued` Vault endpoint](/vault/api-docs/secret/pki#sign-self-issued)
```

rboyer (approved):

website/content/docs/connect/ca/vault.mdx
```diff
@@ -210,48 +210,79 @@ The following Vault policy allows Consul to use pre-existing PKI paths in Vault.
...
-}
+# -----------------------------------------------------------------------------
+# Vault Managed PKI Mounts
+# -----------------------------------------------------------------------------
+
+# -----------------------------------------------------------------------------
+# Needed for Consul to read the PKI mounts already created in Vault
 
-path "/sys/mounts/connect_root" {
+path "/sys/mounts/<insert-RootPKIPath>" {
```
kisunji, line 226: Since the `<>` already suggests "replace me" can we use the documented config field names:
```suggestion
path "/sys/mounts/<root_pki_path>" {
```

website/content/docs/connect/ca/vault.mdx
kisunji: Can this be handled with versioned docs? I found the commented versioning guidelines for `/tune` confusing.

kisunji (approved):

website/content/docs/connect/ca/vault.mdx
kisunji: I support removing version references from config code blocks and calling out 1.12+ required 👍

website/content/docs/connect/ca/vault.mdx
```diff
@@ -210,48 +210,79 @@ The following Vault policy allows Consul to use pre-existing PKI paths in Vault.
 Consul is granted read-only access to the PKI mount points and the Root CA, but is
 granted full control of the Intermediate or Leaf CA for Connect clients.
 
-In this example the `RootPKIPath` is `connect_root` and the `IntermediatePKIPath`
-is `connect_inter`. These values should be updated for your environment.
+This example uses placeholder values for `RootPKIPath` and `IntermediatePKIPath`.
+Please update those placeholders to match the path values in the CA provider configuration.
```
trujillo-adam, line 214:
```suggestion
The `RootPKIPath` and `IntermediatePKIPath` paths have placeholder values. Replace them with real path values in your CA provider configuration.
```
In docs, fewer words outweighs politeness.

website/content/docs/connect/ca/vault.mdx
```diff
@@ -210,48 +210,79 @@ The following Vault policy allows Consul to use pre-existing PKI paths in Vault.
...
 
 <CodeBlockConfig filename="vault-managed-pki-policy.hcl">
 
 ```hcl
-# Existing PKI Mounts
-path "/sys/mounts" {
-  capabilities = [ "read" ]
-}
+# -----------------------------------------------------------------------------
```
trujillo-adam, line 219: Curious about these in-file explanations. Do they mirror a template file included in some package?

trujillo-adam (changes requested):
Can we discuss the docs-within-the-docs approach here? Are we mirroring a template file shipped in some package or repository? If not, I don't think we should suddenly start providing documentation inside code snippets. Tagging @boruszak @im2nguyen for awarenss

website/content/docs/connect/ca/vault.mdx
jkirschner-hashicorp: I didn't check what established practices we have for including comments in HCL file examples. JSON doesn't support comments, so it's not even possible in that context.

I included these comments so that, if someone copies and pastes this policy for their own usage, someone looking at it now or later can understand the purpose when it might not otherwise be obvious. In my experience, I've seen configuration file templates included with software include inline docs comments like this, including commented out things with an explanation of when it should be uncommented (like I've done here for the root rotation permission).

I'm open to removing the comments. I personally found the policy a lot more understandable with them than without them. It also helps with playing "spot the difference" between the Consul vs Vault managed mounts approaches.

Before proceeding, I'm curious what @boruszak and @im2nguyen think.

trujillo-adam: Here is more context about how I'm looking at the comment-based docs.

First, I do see the value. I have documented systems where the product ships with a commented up configuration file that the user duplicates and modifies to their implementation. It can be a useful way to present the information. 

For me, there are a few issues. As you pointed out, you can only provide the comments in HCL (and YAML), so we can't provide the examples for JSON configurations. We lose the rich text features, such as hyperlinks and text formatting, which might be useful in the explanations. Third, it's a deviation from our convention. Are users now expecting documentation within other example configuration files? 

I think the big thing, though, is that the information in this section is presented as a hybrid of conceptual and reference when we should present it as usage. There appears to be logical splits in the policy file where we could create "Define Vault managed PKI mount policies" section, for example. The section would include an explanation with links to the configuration reference items and an example. Or maybe it's a step-by-step process describing how to configure blocks of rules. When the information is presented as conceptual or reference, I think it reads as less actionable, so readers don't know if they should or have to do something.

I'll let Jeff and Tu chime in, though.

jkirschner-hashicorp: If we went with the framing of "leave in comments where something may require thought/action", the only comment we'd want to leave in as the one about enabling root rotation, since best practices would suggest granting that permission for the minimum amount of time necessary when rotation is needed.

website/content/docs/connect/ca/vault.mdx
jkirschner-hashicorp: Modified - check the latest wording

website/content/docs/connect/ca/vault.mdx
```diff
@@ -181,174 +163,171 @@ The configuration options are listed below.
...
+If the paths already exist, Consul uses them as configured.
 
-## Vault ACL Policies
+### Configure Vault ACL policies
 
-Vault PKI can be managed by either Consul or by Vault. If you want to manually create and tune the PKI secret engines used to store the root and intermediate certificates, use Vault Managed PKI Paths. If you want to have the PKI automatically managed for you, use Consul Managed PKI Paths.
+You can define an access control list (ACL) policy that assigns Vault PKI management to Consul or Vault. If you want to manually create and tune the PKI secret engines used to store the root and intermediate certificates, define a policy that uses Vault-managed PKI paths. If you want to automate PKI management, define a policy that uses Consul-managed PKI paths.
 
-### Vault Managed PKI Paths
+#### Define a policy for Vault-managed PKI paths
```
jkirschner-hashicorp, line 187: This headings are often linked to. We need to create a link alias to avoid breaking existing links.

website/content/docs/connect/ca/vault.mdx
```diff
@@ -181,174 +163,171 @@ The configuration options are listed below.
...
+In the following examples, the path configurations use placeholder values for `RootPKIPath` and `IntermediatePKIPath`. Replace them to match the path values in the CA provider configuration.
 
-<CodeBlockConfig filename="vault-managed-pki-policy.hcl">
+1. Add the following rules to you policy file so that Consul can read the PKI mounts already created in Vault:
 
-```hcl
-# -----------------------------------------------------------------------------
-# Vault Managed PKI Mounts
-# -----------------------------------------------------------------------------
+  <CodeBlockConfig filename="vault-managed-pki-policy.hcl">
```
jkirschner-hashicorp, line 195: Note to self: go back through all of these and make sure root vs intermediate is used in all the right places.

website/content/docs/connect/ca/vault.mdx
```diff
@@ -181,174 +163,171 @@ The configuration options are listed below.
...
+  ```hcl
+  path "<root_pki_path>/root/sign-self-issued" {
+    capabilities = [ "sudo", "update" ]
+  }
+  ```
+  </CodeBlockConfig>
+  
+  Note that because this is an extremely privileged endpoint, we recommend only temporarily granting this permission if you need to modify Consul's CA provider configuration.
 
-### Consul Managed PKI Paths
```
jkirschner-hashicorp, line 260: This headings are often linked to. We need to create a link alias to avoid breaking existing links.

website/content/docs/connect/ca/vault.mdx
```diff
@@ -181,122 +163,171 @@ The configuration options are listed below.
...
-signed with the root certificate. The intermediate is used to sign all leaf
-certificates and Consul may periodically generate new intermediates for
+signed with the root certificate. The intermediate signs all leaf
+certificates, and Consul may periodically generate new intermediates for
 automatic rotation. Therefore, Consul requires write access to this path.
 
-If either path does not exist, then Consul will attempt to mount and
-initialize it. This requires additional privileges by the Vault token in use.
-If the paths already exist, Consul will use them as configured.
+If neither path exists, then Consul attempts to mount and
```
jkirschner-hashicorp, line 179: The original statement wasn't clear enough. The suggestion replacement isn't logically equivalent, though. I'll need to update.

website/content/docs/connect/ca/vault.mdx
jkirschner-hashicorp: After checking, I fixed one mistake where I mixed up root and intermediate.

jkirschner-hashicorp:
@trujillo-adam : ready for re-review

website/content/docs/connect/ca/vault.mdx
```diff
@@ -181,124 +170,207 @@ The configuration options are listed below.
...
+`RootPKIPath` and `IntermediatePKIPath`.
 
-```hcl
-# Existing PKI Mounts
-path "/sys/mounts" {
-  capabilities = [ "read" ]
-}
+Next, define the following Vault ACL policy that allows Consul to use those pre-existing PKI mounts in Vault.
+The policy grants Consul limited use of the root PKI engine
+and full control of the intermediate PKI engine to issue service mesh certificates.
```
trujillo-adam, line 213:
```suggestion
Next, define a Vault ACL policy that allows Consul to use the pre-existing PKI mounts in Vault.
The policy grants Consul limited use of the root PKI engine
and full control of the intermediate PKI engine to issue service mesh certificates.
```
Aren't we repeating the same information in the steps below? I think we can delete this paragraph and the one above it. The next paragraph is a good introduction to the procedure for building a suitable policy.

website/content/docs/connect/ca/vault.mdx
```diff
@@ -181,124 +170,207 @@ The configuration options are listed below.
...
+  ```
+  
+  </CodeBlockConfig>
+
+1. Conditional and temporary: Only include the following rules in your Vault policy file
+   while you are changing the CA provider from Vault to a different CA provider (such as Consul's built-in provider)
+   or changing the `RootPKIPath` used by the Vault CA provider.
+   Those CA provider modifications trigger a root CA change that requires an
+   extremely privileged root cross-sign operation.
+   We recommend removing this elevated privilege immediately after the CA provider modification is finished.
```
trujillo-adam, line 284:
```suggestion
1. You should only include the following rules in your Vault policy file
   while changing the CA provider from Vault to a different CA provider, such as Consul's built-in provider,
   or changing the `RootPKIPath` used by the Vault CA provider.
   The CA provider modifications trigger a root CA change that requires a
   highly-privileged root cross-sign operation.
   Remove the elevated privilege immediately after the CA provider modification is finished.
```
I get why you wanted to label the step, but it adds too many words without as much benefit. As far as the recommendation to remove the elevated privileges, should that be a final step at the end of this procedure? It would make this step shorter and provide a dedicated action for readers following the process.

website/content/docs/connect/ca/vault.mdx
```diff
@@ -181,124 +170,207 @@ The configuration options are listed below.
...
-path "/sys/mounts/connect_inter" {
-  capabilities = [ "create", "read", "update", "delete", "list" ]
-}
+#### Define a policy for Consul-managed PKI paths ((#consul-managed-pki-paths))
 
-# Needed for Consul 1.11+
-path "/sys/mounts/connect_inter/tune" {
-  capabilities = [ "update" ]
-}
+To use Consul-managed PKI paths, ensure no PKI secrets engines are mounted at
```
trujillo-adam, line 297: Same comment as the previous procedure - we already describe what each rule in the policy does in the steps, so I think we should remove the first two paragraphs in this introduction.

website/content/docs/connect/ca/vault.mdx
```diff
@@ -181,124 +170,207 @@ The configuration options are listed below.
...
+  ```
+  
+  </CodeBlockConfig>
+
+1. Conditional and temporary: Only include the following rules in your Vault policy file
+   while you are changing the CA provider from Vault to a different CA provider (such as Consul's built-in provider)
+   or changing the `RootPKIPath` used by the Vault CA provider.
+   Those CA provider modifications trigger a root CA change that requires an
+   extremely privileged root cross-sign operation.
+   We recommend removing this elevated privilege immediately after the CA provider modification is finished.
```
trujillo-adam, line 367:
```suggestion
1. You should only include the following rules in your Vault policy file
   while you are changing the CA provider from Vault to a different CA provider, such as Consul's built-in provider,
   or changing the `RootPKIPath` used by the Vault CA provider.
   The CA provider modifications trigger a root CA change that requires a
  highly-privileged root cross-sign operation.
   Remove the elevated privilege immediately after the CA provider modification is finished.
```
Same comment as above

trujillo-adam (commented):
I added a few more suggestions, mostly to add more precise language where necessary and to streamline areas where I think we're over-explaining.

website/content/api-docs/connect/ca.mdx
```diff
@@ -177,11 +189,14 @@ The corresponding CLI command is [`consul connect ca set-config`](/commands/conn
...
-  failures until the rollout completes. See [Forced Rotation Without
-  Cross-Signing](/docs/connect/ca#forced-rotation-without-cross-signing)
+- `ForceWithoutCrossSigning` `(bool: false)` - Indicates that the CA change
+  should be forced to complete even if the current CA doesn't support root cross-signing.
+
+  ~> **Caution:** Setting this field to `true` will cause temporary connection failures
+  until service mesh proxies and/or Consul client agents receive a new certificate
+  that establishes trust with the new root.
+  Do not use this field unless you are sure you need it.
+  Refer to [Forced Rotation Without Cross-Signing](/docs/connect/ca#forced-rotation-without-cross-signing)
```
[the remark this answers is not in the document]
jkirschner-hashicorp: @im2nguyen : On a separate PR, I received feedback that we want to standardize on generic callouts like "Note / Caution / Warning" for the different callout types. It seems like we may not be internally aligned on which direction we want to go between specific callouts vs general callouts.

website/content/docs/connect/ca/vault.mdx
```diff
@@ -7,40 +7,29 @@ description: >-
...
 
-~> **Note:** Do not use Vault v1.11.0+ as Consul's Connect CA provider — the intermediate CA will become unable to issue the leaf nodes required by service mesh, and by Consul client agents if using auto-encrypt or auto-config and using TLS for agent communication. If you are already using Vault 1.11+ as a Connect CA, refer to this [Knowledge Base article](https://support.hashicorp.com/hc/en-us/articles/11308460105491) for more information about the underlying cause and recommended workaround.
+- Use Vault 0.10.3 to latest 1.10.x.
 
-## Configuration
+~> **Note:** Do not use Vault v1.11.0+ as Consul's Connect CA provider — the intermediate CA becomes unable to issue the leaf nodes required by service mesh, and by Consul client agents if using auto-encrypt or auto-config and using TLS for agent communication. If you are already using Vault 1.11+ as a Connect CA, refer to this [Knowledge Base article](https://support.hashicorp.com/hc/en-us/articles/11308460105491) for more information about the underlying cause and recommended workaround.
```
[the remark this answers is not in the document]
jkirschner-hashicorp: They are both Consul options/configurations. I also find the correction to be a bit confusing. I'd prefer to keep the additional wording.

Regarding the call-out text, same comment as above: I'll go with whatever you and @im2nguyen agree upon on generic vs specific.

website/content/docs/connect/ca/vault.mdx
```diff
@@ -7,40 +7,29 @@ description: >-
...
-The Vault CA is enabled by setting the CA provider to `"vault"` and
-setting the required configuration values.
+## Enable Vault as the CA
 
-The configuration may either be provided in the agent's configuration file using
-the [`ca_provider`] and [`ca_config`] options, or configured using the
-[`/connect/ca/configuration`] API endpoint.
+You can enable Vault as the CA by configuring Consul to use `"vault"` as the CA provider
+and including the required provider configuration options.
+You can provide the CA configuration in the agent configuration file of server agents,
```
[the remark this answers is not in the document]
jkirschner-hashicorp: I hesitated to write it this way to begin with because their are multiple types of files that can contain configuration ... "configuration entries" are different from "agent configuration".

That said, perhaps it's better overall to make this suggested change, even though there may be a possibility of misinterpretation.

website/content/docs/connect/ca/vault.mdx
```diff
@@ -80,11 +69,12 @@ connect {
 
 </CodeTabs>
 
-The configuration options are listed below.
+## Configuration Reference
```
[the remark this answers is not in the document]
jkirschner-hashicorp: At some point, we should just make a pass over all the pages to change the capitalization.

For any _new_ headers I create, I'm using the style guide. But for content I'm not editing (or only shuffling on the page), I haven't been touching the existing header to make them compliant (because I'm focused on a particular change).

This probably creates a lot of back and forth for both reviewers and authors. If we update all the content at once, it should be easier for everyone going forward.

website/content/docs/connect/ca/vault.mdx
jkirschner-hashicorp: Yeah. I think I wrote this before I had the summary paragraph above that compares the two options.

website/content/docs/connect/ca/vault.mdx
jkirschner-hashicorp: The policy snippets in steps 1-3 must **always** be in place for Consul to work.

This policy snippet is ideally only in place for a very brief period of time while performing a specific operation.

These four "steps" aren't really a _procedure_ to follow. I don't think it makes sense to add a "step 5" for removing the elevated privileges.

Maybe the right solution is to move this temporarily privilege into its own section? It's shared between Vault and Consul managed paths, so I don't need to duplicate it in both sections... I'll try that out so we can take a look.

website/content/api-docs/connect/ca.mdx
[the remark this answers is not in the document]
jkirschner-hashicorp: Use "caution"

website/content/docs/connect/ca/vault.mdx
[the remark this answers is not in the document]
jkirschner-hashicorp: "Compatibility warning"

website/content/api-docs/connect/ca.mdx
```diff
@@ -169,6 +169,18 @@ The table below shows this endpoint's support for
 
 The corresponding CLI command is [`consul connect ca set-config`](/commands/connect/ca#set-config).
 
+~> **If currently using Vault CA provider:**
```
jkirschner-hashicorp, line 172: Note to self: revisit this before submitting for final review

website/content/commands/connect/ca.mdx
```diff
@@ -97,17 +97,32 @@ Configuration updated!
 
 The return code will indicate success or failure.
 
+~> **If currently using Vault CA provider:**
```
jkirschner-hashicorp, line 100: Note to self: revisit before final review

website/content/docs/connect/ca/vault.mdx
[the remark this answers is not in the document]
jkirschner-hashicorp: This paragraph will also be touched again in about ~1 week, as we're actively working to remediate this incompatibility.

website/content/docs/connect/ca/vault.mdx
```diff
@@ -181,124 +171,190 @@ The configuration options are listed below.
...
 }
 ```
-
 </CodeBlockConfig>
 
-<!-- Reference style links -->
-[`ca_config`]: /docs/agent/config/config-files#connect_ca_config
-[`ca_provider`]: /docs/agent/config/config-files#connect_ca_provider
-[`/connect/ca/configuration`]: /api-docs/connect/ca#update-ca-configuration
+We recommend adding this elevated privilege just before the CA provider
```
jkirschner-hashicorp, line 359: @rboyer : In practice... how would a Vault user actually accomplish this?

If I understand correctly, a Vault token's policies are immutable. That would mean you'd need to:
- Create a new Vault token with the elevated policy
- Apply the new Vault token to the Consul Connect CA configuration
- Perform the operation
- Reapply the previous Vault token with non-elevated permissions to the Consul Connect CA configuration

Alternatively, you could use a Vault auth method, and modify the policy associated with the auth method... But if I understand correctly, that wouldn't actually take effect unless Consul re-logs in using the auth method (after the modification) to get a new token (which contains the modified rules). And I don't know how the user would force Consul to do that... So maybe this is only workable when using a Vault token directly rather than an auth method, even though an auth method is generally preferable for enhanced security?

trujillo-adam (approved):
There are still some language and style guide issues, but I want to unblock you from getting the correct information published. I can open a separate PR to address the writing issues.

PR #15487, opened 2022-11-19
docs: add retry_max agent config option

jkirschner-hashicorp:
This agent config option is already supported (per @pglass), but not documented.


trujillo-adam (approved):
Acceptable (for now 🤔)

jkirschner-hashicorp:
@trujillo-adam : I also don't love the way these options are documented. But it was the quickest way to add missing info in the interim, in a manner consistent with what's less-than-ideal-but-already-present. Thanks :)

PR #15498, opened 2022-11-20
docs: add peering control plane diagrams

jkirschner-hashicorp:
Diagrams were created in the same style as the WAN federation control plane connectivity diagrams.

[Preview link](https://consul-63nv8sa4u-hashicorp.vercel.app/consul/docs/connect/gateways/mesh-gateway/peering-via-mesh-gateways)


jkirschner-hashicorp:
@freddygv : Can you confirm to accuracy of the diagrams in this PR?

[Preview link](https://consul-o3ha8mtmt-hashicorp.vercel.app/consul/docs/connect/gateways/mesh-gateway/peering-via-mesh-gateways)

trujillo-adam (approved):
This is fine from a docs perspective, but in the future we should avoid placing titles and explanatory text directly on images. If explanations of the diagram are necessary, we should provide them as text on the page. This makes it easier to edit descriptions and makes the image more portable if we need to make a change.

jkirschner-hashicorp:
Including legends and titles in the text make the images more portable outside of the docs, such as in workshops or presentations. That said, I agree that it does make modification slower, in case there's a correction or clarification to be made. To partially mitigate that, I've included the image source SVG files in addition to the rendered PNG, so that someone could just modify the text in the SVG and re-render a new PNG.

I agree that if we start having more diagrams, it would be good to agree on what should (and should not) be included!

freddygv:
@jkirschner-hashicorp Just one comment: the server port should be `grpc_tls`, which defaults to 8503:
https://developer.hashicorp.com/consul/docs/agent/config/config-files#grpc_tls_port

We always enable TLS for peering.

jkirschner-hashicorp:
Nice catch :) I'll update the diagrams

jkirschner-hashicorp:
@freddygv : I spotted another small mistake in the diagram: it isn't enabled by a server agent config, but instead by `Requires config entry of type "mesh" with Peering.PeerThroughMeshGateways=true`

PR #15503, opened 2022-11-21
fix(peering): increase the gRPC limit to 8MB

huikang:
### Description
Increase the gRPC limit to 50MB in both peering grpc client and health rpc client

The rpcclient limit is also updated since the health streaming client needs to pull the service from the catalog.

```
2022-11-21T11:08:33.477-0500 [DEBUG] agent.http: Request finished: method=GET url="/v1/internal/ui/services?dc=dc3&index=26" from=127.0.0.1:51829 latency=5.240129042s
2022-11-21T11:08:33.502-0500 [DEBUG] agent.server.peering-syncer: stream disconnected due to 'resource exhausted' error; reconnecting: peer_id=f7e2b2a5-2453-de22-823f-69d898971567 peer_name=acceptor error="unexpected error receiving from the stream: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (5245223 vs. 4194304)"


2022-11-21T11:32:45.270-0500 [ERROR] agent.rpcclient.health: subscribe call failed: err="rpc error: code = ResourceExhausted desc = grpc: received message larger than max (8389107 vs. 4194304)" failure_count=11 key=service-burst-0 topic=ServiceHealth
```

`/v1/health/service/large-service` return the error message

```
Error: Maximum response size reached
```

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/agent/grpc-internal/client.go
+++ b/agent/grpc-internal/client.go
@@ -148,7 +148,9 @@ func (c *ClientConnPool) dial(datacenter string, serverType string) (*grpc.Clien
 		grpc.WithKeepaliveParams(keepalive.ClientParameters{
 			Time:    30 * time.Second,
 			Timeout: 10 * time.Second,
-		}))
+		}),
+		grpc.WithDefaultCallOptions(grpc.MaxCallSendMsgSize(50*1024*1024), grpc.MaxCallRecvMsgSize(50*1024*1024)),
+	)
 	if err != nil {
 		return nil, err
 	}
--- a/agent/grpc-internal/client.go
+++ b/agent/grpc-internal/client.go
@@ -149,7 +149,7 @@ func (c *ClientConnPool) dial(datacenter string, serverType string) (*grpc.Clien
 			Time:    30 * time.Second,
 			Timeout: 10 * time.Second,
 		}),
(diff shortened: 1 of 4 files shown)


agent/grpc-internal/client.go
```diff
@@ -148,7 +148,9 @@ func (c *ClientConnPool) dial(datacenter string, serverType string) (*grpc.Clien
 		grpc.WithKeepaliveParams(keepalive.ClientParameters{
 			Time:    30 * time.Second,
 			Timeout: 10 * time.Second,
-		}))
+		}),
+		grpc.WithDefaultCallOptions(grpc.MaxCallSendMsgSize(50*1024*1024), grpc.MaxCallRecvMsgSize(50*1024*1024)),
```
hashi-derek, line 152: This gRPC client isn't used for peerstream traffic from what I remember, so I'm not sure if we need to increase its value. I suppose it would only ever prevent an issue from occurring, so it's probably safe.

huikang: Yeah, I  noticed this error message after I fixed the peering gRPC client. I think this is from the health rpcclient to query for the health of services in the cluster.

huikang: I think we might need this; otherwise the client agent in the peering client can't get the status of the very large services imported from peered cluster.

hashi-derek: Ah, okay. That makes sense.

hashi-derek (approved):

PR #15525, opened 2022-11-21
Fix Vault managed intermediate PKI bug

kisunji:
### Description
If you are 1) using Vault as a Consul Connect CA, 2) using Vault-managed policies (i.e. you are bringing external existing Vault mounts to use as CA instead of giving Consul privileges to make mounts itself), and 3) have an empty intermediate PKI mount, Consul will fail to startup the CA Manager.

People have gotten around this issue by adding a dummy cert in the intermediate PKI mount but this PR aims to fix the root cause.

* Added test that fails without the code changes

### Links
Vault as CA reference: https://developer.hashicorp.com/consul/docs/connect/ca/vault#vault-acl-policies

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/connect/ca/provider_vault.go
+++ b/agent/connect/ca/provider_vault.go
@@ -71,6 +71,11 @@ type VaultProvider struct {
 	spiffeID  *connect.SpiffeIDSigning
 	logger    hclog.Logger
 
+	// isConsulMountedIntermediate is used to determine if we should tune the
+	// mount if the VaultProvider is ever reconfigured. This is at most a
+	// "best guess" to determine whether this instance of Consul created the
+	// intermediate mount but will not be able to tell if an existing mount
+	// was created by Consul (in a previous running instance) or was external.
 	isConsulMountedIntermediate bool
 }
 
@@ -398,29 +403,33 @@ func (v *VaultProvider) setupIntermediatePKIPath() error {
 			return fmt.Errorf("unexpected error while fetching intermediate CA: %w", err)
 		}
 	} else {
-		// If Consul was responsible for mounting the intermediate PKI path
-		// we should update the mount with any new config.
-		if v.isConsulMountedIntermediate {
-			// This codepath requires the Vault policy:
-			//
-			//   path "/sys/mounts/<intermediate_pki_path>/tune" {
-			//     capabilities = [ "update" ]
-			//   }
-			//
-			err := v.tuneMountNamespaced(v.config.IntermediatePKINamespace, v.config.IntermediatePKIPath, &mountConfig)
-			if err != nil {
+		v.logger.Info("Found existing Intermediate PKI path mount",
+			"namespace", v.config.IntermediatePKINamespace,
+			"path", v.config.IntermediatePKIPath,
+		)
+
+		// This codepath requires the Vault policy:
+		//
+		//   path "/sys/mounts/<intermediate_pki_path>/tune" {
+		//     capabilities = [ "update" ]
+		//   }
+		//
+		err := v.tuneMountNamespaced(v.config.IntermediatePKINamespace, v.config.IntermediatePKIPath, &mountConfig)
+		if err != nil {
+			if v.isConsulMountedIntermediate {
 				v.logger.Warn("Intermediate PKI path was mounted by Consul but could not be tuned",
+					"namespace", v.config.IntermediatePKINamespace,
+					"path", v.config.IntermediatePKIPath,
+					"error", err,
+				)
+			} else {
+				v.logger.Debug("Failed to tune Intermediate PKI mount. 403 Forbidden is expected if Consul does not have tune capabilities for the Intermediate PKI mount (i.e. using Vault-managed policies)",
 					"namespace", v.config.IntermediatePKINamespace,
 					"path", v.config.IntermediatePKIPath,
 					"error", err,
 				)
 			}
 
-		} else {
-			v.logger.Info("Found existing Intermediate PKI path mount",
-				"namespace", v.config.IntermediatePKINamespace,
-				"path", v.config.IntermediatePKIPath,
-			)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


agent/connect/ca/provider_vault.go
```diff
@@ -381,19 +384,47 @@ func (v *VaultProvider) setupIntermediatePKIPath() error {
...
+				return fmt.Errorf("failed to mount intermediate PKI backend: %w", err)
 			}
+			// Required to determine if we should tune the mount
+			// if the VaultProvider is ever reconfigured.
+			v.isConsulMountedIntermediate = true
+
+		} else if err == ErrBackendNotInitialized {
+			// If this is the first time calling setupIntermediatePKIPath, the backend
+			// will not have been initialized. Since the mount is ready we can suppress
+			// this error.
```
kisunji, line 396: This `else if` block is the bugfix

agent/connect/ca/provider_vault.go
```diff
@@ -381,19 +384,47 @@ func (v *VaultProvider) setupIntermediatePKIPath() error {
...
+					"error", err,
+				)
+			}
+
+		} else {
+			v.logger.Info("Found existing Intermediate PKI path mount",
+				"namespace", v.config.IntermediatePKINamespace,
+				"path", v.config.IntermediatePKIPath,
+			)
 		}
```
kisunji, line 424: Since tuning the mount is only relevant for Consul-managed, I added a conditional to skip since otherwise you'd see an ominous WARN in the logs on startup.

This isn't the cleanest way to determine if the provider is "Consul-managed" or "Vault-managed" but introducing explicit configuration requires some more consideration around default/missing values.

agent/connect/ca/provider_vault.go
```diff
@@ -381,19 +384,47 @@ func (v *VaultProvider) setupIntermediatePKIPath() error {
...
 	}
 
-	// Create the role for issuing leaf certs if it doesn't exist yet
+	// Create the role for issuing leaf certs
```
kisunji, line 427: Updating comment to match code, since it no longer checks for existence

agent/connect/ca/provider_vault.go
jkirschner-hashicorp: Tuning this mount really only matters if the mount's `MaxLeaseTTL` is lower than the `IntermediateCertTTL`. Otherwise, tuning the config has no effect, as the [intermediate CA cert will still be issued with the requested TTL](https://github.com/hashicorp/consul/blob/b1013e06b56f522cbb96d24eb08a7dbd9d2c2d05/agent/connect/ca/provider_vault.go#LL587C39-L587C39).

I think I heard, though haven't tested, that if you ask Vault to generate a cert (i.e., intermediate CA cert) with a TTL greater than `MaxLeaseTTL`, the operation will still succeed, the TTL will just be truncated to `MaxLeaseTTL`. So everything will still *work*, it's just that the intermediate CA cert will expire sooner than configured. (Unless there's something about our "attempt to renew at 50% of lifetime" math that breaks because it depends on `IntermediateCertTTL` rather than the TTL of the actual issuer cert.)

cipherboy: I think its somewhat short: https://developer.hashicorp.com/vault/docs/concepts/tokens#the-general-case -- 32 days.

You'll definitely want to read the cert's actual validity period rather than assuming the TTL is OK, not sure if that's useful in this PR or a follow-up though.

Also, I think in 1.11+, we've started adding a warning in this event, but obviously that doesn't help Vault <= 1.10...

cipherboy (approved):
Another ACK from the Vault side of things, looks nice @kisunji!

agent/connect/ca/provider_vault.go
```diff
@@ -381,19 +384,47 @@ func (v *VaultProvider) setupIntermediatePKIPath() error {
 				Config:      mountConfig,
 			})
 			if err != nil {
-				return err
+				return fmt.Errorf("failed to mount intermediate PKI backend: %w", err)
 			}
+			// Required to determine if we should tune the mount
+			// if the VaultProvider is ever reconfigured.
+			v.isConsulMountedIntermediate = true
+
```
rboyer, line 392: On an earlier meeting did we decide that this needed a `fallthrough` like the root-generate switch had?

kisunji: When we talked I thought I should be initializing the mount from within `setupIntermediatePKIPath` but realized it wasn't necessary to do that here. So all we need to do is prevent error-ing out here and other codepaths like `generateIntermediateCSR` below will initialize the mount as necessary.

kisunji: So if we do end up mounting the intermediate PKI, there's no other action to be done (unlike `GenerateRoot` which is actually responsible for mounting+initializing+generating).

I thought about making the code paths for root and intermediate more parallel to each other but deferred it to future Chris

agent/connect/ca/provider_vault.go
```diff
@@ -381,19 +384,47 @@ func (v *VaultProvider) setupIntermediatePKIPath() error {
 				Config:      mountConfig,
 			})
 			if err != nil {
-				return err
+				return fmt.Errorf("failed to mount intermediate PKI backend: %w", err)
 			}
+			// Required to determine if we should tune the mount
+			// if the VaultProvider is ever reconfigured.
+			v.isConsulMountedIntermediate = true
```
rboyer, line 391: Note: this would only get set to true the FIRST time a consul instance ran through this code. After the mount existed it wouldn't set this if it matters.

kisunji: Yeah I intended it to be a "best guess" field that only sets if the `VaultProvider` EVER creates the intermediate mount. If the mount ever existed then we'll never attempt to tune the mount but output a log to help operators diagnose it if they're ever confused by the behavior.

rboyer (approved):

PR #15531, opened 2022-11-22
adding config for request_limits

jmurret:
### Description
Add new configuration to configure global rate limiter.
This should ensure:

- Config is parsed correctly
- Config generates a new rate limiter handler
- Config need to be reloadable using “consul reload”
- Config need to be reloadable using consul config auto reload feature

* Currently the main unit tests are:
  * [TestAgent_consulConfig_RequestLimits](https://github.com/hashicorp/consul/pull/15531/files#diff-e12d66df9a328a0b4b4e61abbde9a65eb60a4c8a6a8ca9249a455f173007ac89R4246) - ensures config is loaded
  * [TestServer_ReloadConfig](https://github.com/hashicorp/consul/pull/15531/files#diff-d8fc6b5ebbb6ec0cccca872b41ab0d6267bb2ed07982fe15cb566cc7ca95236cR1880-R1892) - ensures the rate limit handler gets called on reload

### Links
[NET-1379](https://hashicorp.atlassian.net/browse/NET-1379)

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### Adding a field checklist
There are four specific cases covered with increasing complexity:
 1. adding a simple config field only used by client agents
 1. adding a CLI flag to mirror that config field
 1. adding a config field that needs to be used in Consul servers
 1. adding a field to the Service Definition

#### Adding a Simple Config Field for Client Agents

   `agent/config/config.go`.
   `agent/config/runtime.go`.
   translate.
      - [x] Add validation to Validate in `agent/config/builder.go`.
      - [x] If the config needs to be defaulted for the test server used in unit tests,
            also add it to `DefaultConfig()` in `agent/consul/defaults.go`.
 - [x] **If** your config should take effect on a reload/HUP.
      - [x] Add necessary code to to trigger a safe (locked or atomic) update to
        more of the following places:
      - [x] Add a test to `agent/agent_test.go` similar to others with prefix
        `TestAgent_reloadConfig*`.

Done! You can now use your new field in a client agent by accessing
`s.agent.Config.<FieldName>`.

If you need a CLI flag, access to the variable in a Server context, or touched
the Service Definition, make sure you continue on to follow the appropriate
checklists below.

#### Adding a Simple Config Field for Servers
Consul servers have a separate Config struct for reasons. Note that Consul
server agents are actually also client agents, so in some cases config that is
only destined for servers doesn't need to follow this checklist provided it's
only needed during the bootstrapping of the server (which is done in code shared
by both server and client components in `agent.go`). For example WAN Gossip
configs are only valid on server agents but since WAN Gossip is setup in
`agent.go` they don't need to follow this checklist. The simplest (and mostly
accurate) rule is:

> If you need to access the config field from code in  `agent/consul` (e.g. RPC
> endpoints), then you need to follow this. If it's only in `agent` (e.g. HTTP
> endpoints or agent startup) you don't.

A final word of warning - **you should never need to pass config into the FSM
(`agent/consul/fsm`) or state store (`agent/consul/state`)**. Doing so is **_very
dangerous_** and can violate consistency guarantees and corrupt databases. If
you think you need this then please discuss the design with the Consul team
before writing code!

Consul's server components for historical reasons don't use the `RuntimeConfig`
struct they have their own struct called `Config` in `agent/consul/config.go`.

 - [x] Do all of the steps in [Adding a Simple Config
 - [x] Add code to set the values from the `RuntimeConfig` in the confusingly
   named `consulConfig` method in `agent/agent.go`
 - [x] **If needed**, add a test to `agent_test.go` if there is some non-trivial
 - [x] **If** your config should take effect on a reload/HUP
      - [x] Add necessary code to `ReloadConfig` in `agent/consul/server.go` this
        updated.
      - [x] Add a new test or a new assertion to `TestServer_ReloadConfig`

You can now access that field from `s.srv.config.<FieldName>` inside an RPC
handler.


agent/consul/rate/handler.go
```diff
@@ -134,6 +168,18 @@ func (h *Handler) UpdateConfig(cfg HandlerConfig) {
 	h.limiter.UpdateConfig(cfg.GlobalReadConfig, globalRead.ConfigKey())
 }
 
+func (h *Handler) GetGlobalReadLimiterConfig() (*multilimiter.LimiterConfig, bool) {
```
jmurret, line 171: Added this to be able to assert reload worked...is there a better way to assert this and not add this additional API surface area to handler and multilimiter?

agent/consul/rate/multilimiter/multilimiter.go
```diff
@@ -26,6 +26,7 @@ func Key(prefix, key []byte) KeyType {
 type RateLimiter interface {
 	Run(ctx context.Context)
 	Allow(entity LimitedEntity) bool
+	GetConfig(prefix []byte) (*LimiterConfig, bool)
```
jmurret, line 29: see above.

agent/agent_test.go
```diff
@@ -4179,6 +4180,39 @@ func TestAgent_ReloadConfig_XDSUpdateRateLimit(t *testing.T) {
...
+					read_rate = 8888
+					write_rate = 9999
+				}
+			}`,
+		},
+	)
+
+	require.Equal(t, consulrate.Mode("enforcing"), c.RequestLimitsMode)
+	require.NoError(t, a.reloadConfigInternal(c))
+	// JM TODO: assert on something else that shows it was reloaded.
```
jmurret, line 4210: Is this possible?

agent/config/runtime.go
```diff
@@ -1015,6 +1016,35 @@ type RuntimeConfig struct {
 	// flag: -rejoin
 	RejoinAfterLeave bool
 
+	// RequestLimitsMode enforces the action that will occur when RequestLimitsReadRate
+	// or RequestLimitsWriteRate is exceeded.  A value of "enforce" will block
```
jmurret, line 1020: Should we also include a disabled mode?  If not, do we only turn this on if `mode` is specified?  If they have read_rate or write rate should it be ignored if mode is not specified?

agent/consul/server_test.go
```diff
@@ -1864,6 +1870,20 @@ func TestServer_ReloadConfig(t *testing.T) {
 	require.Equal(t, rate.Limit(1000), limiter.Limit())
 	require.Equal(t, 10000, limiter.Burst())
 
+	// Check the incoming RPC rate limiter got updated
+	handlerConfig := s.incomingRPCLimiter.GetConfig()
+	require.Equal(t, rc.RequestLimitsMode, handlerConfig.GlobalMode)
+	require.Equal(t, 99, handlerConfig.Config.Rate)
```
jmurret, line 1876: Should HandlerCong.Config be asserted?

agent/consul/rate/handler.go
```diff
@@ -27,17 +27,53 @@ var (
...
-	ModePermissive Mode = iota
+	ModePermissive Mode = "permissive"
 
-	// ModeEnforcing causes the handler to reject the rate-limted operation.
-	ModeEnforcing
+	// ModeEnforcing causes the handler to reject the rate-limited operation.
+	ModeEnforcing Mode = "enforcing"
 )
 
+var modeToName = map[Mode]string{
```
jmurret, line 44: might move this out to a mode.go file

agent/config/default.go
```diff
@@ -97,6 +97,11 @@ func DefaultSource() Source {
 		limits = {
 			http_max_conns_per_client = 200
 			https_handshake_timeout = "5s"
+			request_limits = {
```
jmurret, line 100: Should this be omitted?  Configured to use disabled mode?

agent/config/builder.go
```diff
@@ -1047,6 +1048,9 @@ func (b *builder) build() (rt RuntimeConfig, err error) {
 		ReconnectTimeoutLAN:               b.durationVal("reconnect_timeout", c.ReconnectTimeoutLAN),
 		ReconnectTimeoutWAN:               b.durationVal("reconnect_timeout_wan", c.ReconnectTimeoutWAN),
 		RejoinAfterLeave:                  boolVal(c.RejoinAfterLeave),
+		RequestLimitsMode:                 b.requestsLimitsModeVal(stringVal(c.Limits.RequestLimits.Mode)),
```
jmurret, line 1051: Do we need validation rules for these to prevent things like setting one of the rates, but not the mode?  Or would the mode just pick up the default?

agent/consul/server.go
```diff
@@ -468,8 +472,17 @@ func NewServer(config *Config, flat Deps, externalGRPCServer *grpc.Server) (*Ser
 
 	// TODO(NET-1380, NET-1381): thread this into the net/rpc and gRPC interceptors.
 	s.incomingRPCLimiter = rpcRate.NewHandler(rpcRate.HandlerConfig{
-		// TODO(NET-1379): pass in _real_ configuration.
-		GlobalMode: rpcRate.ModePermissive,
+		// TODO(NET-1379):
+		// - should disabled mode not register this handler?  Or does the handler have logic for disabled?
+		GlobalMode: rpcRate.Mode(config.RequestLimitsMode),
+		GlobalReadConfig: multilimiter.LimiterConfig{
+			Rate:  config.RequestLimitsReadRate,
```
jmurret, line 479: Handler is always configured.  Here we detect the user did not set the mode and we pass rate.inf.

agent/agent_test.go
jmurret: Remove this test.

agent/consul/rate/handler.go
```diff
@@ -27,17 +27,53 @@ var (
 
 // Mode determines the action that will be taken when a rate limit has been
 // exhausted (e.g. log and allow, or reject).
-type Mode int
+type Mode string
 
 const (
+	// ModeDisabled causes the handler to not register itself.
+	ModeDisabled Mode = "disabled"
```
jmurret, line 34: remove this.

agent/consul/rate/handler.go
```diff
@@ -27,17 +27,53 @@ var (
 
 // Mode determines the action that will be taken when a rate limit has been
 // exhausted (e.g. log and allow, or reject).
-type Mode int
+type Mode string
```
dhiaayachi, line 30: See my comment above about converting this to int maybe

agent/consul/server.go
```diff
@@ -467,10 +471,19 @@ func NewServer(config *Config, flat Deps, externalGRPCServer *grpc.Server) (*Ser
...
+		s.incomingRPCLimiter = rpcRate.NewHandler(rpcRate.HandlerConfig{
+			GlobalMode: rpcRate.Mode(config.RequestLimitsMode),
+			GlobalReadConfig: multilimiter.LimiterConfig{
+				Rate:  config.RequestLimitsReadRate,
+				Burst: int(config.RequestLimitsReadRate) * requestLimitsBurstMultiplier,
+			},
+			GlobalWriteConfig: multilimiter.LimiterConfig{
+				Rate:  config.RequestLimitsWriteRate,
+				Burst: int(config.RequestLimitsWriteRate) * requestLimitsBurstMultiplier,
+			},
```
dhiaayachi, line 484: Maybe later when we get to IP rate limiting, but I think this should be a substruct so we can use it for other kind of limiting.

agent/consul/server.go
```diff
@@ -1676,6 +1689,19 @@ func (s *Server) ReloadConfig(config ReloadableConfig) error {
...
+		GlobalMode: config.RequestLimitsMode,
+		GlobalReadConfig: multilimiter.LimiterConfig{
+			Rate:  config.RequestLimitsReadRate,
+			Burst: int(config.RequestLimitsReadRate) * requestLimitsBurstMultiplier,
+		},
+		GlobalWriteConfig: multilimiter.LimiterConfig{
+			Rate:  config.RequestLimitsWriteRate,
+			Burst: int(config.RequestLimitsWriteRate) * requestLimitsBurstMultiplier,
+		},
+	})
```
dhiaayachi, line 1703: This is fine for now but I'm wondering if we need to check if the config changes before updating it. This could be done in the multi-limiter to make an update config a no-op if the config don't change.

dhiaayachi (commented):
Great work @jmurret. Adding a config in Consul is challenging and I think you covered all of the different ways that Consul config could change 😄 

I left few comments for consideration.

agent/config/runtime.go
```diff
@@ -1015,6 +1016,37 @@ type RuntimeConfig struct {
...
+
+	// RequestLimitsReadRate controls how frequently RPC, gRPC, and HTTP
+	// queries are allowed to happen. In any large enough time interval, rate
+	// limiter limits the rate to RequestLimitsReadRate tokens per second.
+	//
+	// See https://en.wikipedia.org/wiki/Token_bucket for more about token
+	// buckets.
+	//
+	// hcl: limits { request_limits { read_rate = (float64|MaxFloat64) } }
+	RequestLimitsReadRate rate.Limit
```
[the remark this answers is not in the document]
jmurret: We currently use it for `RPCRateLimit` and `XDSUpdateRateLimit`.  So I am following that convention.  I think if we consider this for our current project, we should consider making it consistent.  I also don't feel like the code is that invasive that the cost of changing this later, should we need to, makes it worth this refactor now.

agent/consul/server.go
jmurret: Good point. Would this be a comparison with the handler's config?

agent/agent.go
```diff
@@ -4027,17 +4032,18 @@ func (a *Agent) reloadConfig(autoReload bool) error {
...
 				}
 			}
 			if f.oldCfg.CertFile != f.newCfg.CertFile {
-				a.configFileWatcher.Replace(f.oldCfg.CertFile, f.newCfg.CertFile)
+				err = a.configFileWatcher.Replace(f.oldCfg.CertFile, f.newCfg.CertFile)
 				if err != nil {
 					return err
 				}
 			}
+
```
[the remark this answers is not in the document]
jmurret: I added this because currently the statements below will not return errors if the configFileWatcher swap returns an error.  I can move this to an outside PR, but either this change needs to occur or we just remove the error returns below them because they are not actually doing the way this is currently written.

agent/config/builder.go
```diff
@@ -1780,6 +1784,19 @@ func (b *builder) dnsRecursorStrategyVal(v string) dns.RecursorStrategy {
 	return out
 }
 
+func (b *builder) requestsLimitsModeVal(v string) consulrate.Mode {
+	var out consulrate.Mode
+
+	mode, ok := consulrate.RequestLimitsModeFromName(v)
```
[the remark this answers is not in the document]
jmurret: Good call!  That has been changed.

agent/config/default.go
```diff
@@ -97,6 +97,11 @@ func DefaultSource() Source {
...
+				mode = "disabled"
+				read_rate = -1
+				write_rate = -1
```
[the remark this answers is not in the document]
jmurret: Should I make these 0 and add validation that this is greater than 0?

agent/config/config.go
```diff
@@ -713,16 +713,23 @@ type UnixSocket struct {
...
-	RPCClientTimeout      *string  `mapstructure:"rpc_client_timeout"`
-	RPCHandshakeTimeout   *string  `mapstructure:"rpc_handshake_timeout"`
-	RPCMaxBurst           *int     `mapstructure:"rpc_max_burst"`
-	RPCMaxConnsPerClient  *int     `mapstructure:"rpc_max_conns_per_client"`
-	RPCRate               *float64 `mapstructure:"rpc_rate"`
-	KVMaxValueSize        *uint64  `mapstructure:"kv_max_value_size"`
-	TxnMaxReqLen          *uint64  `mapstructure:"txn_max_req_len"`
+	HTTPMaxConnsPerClient *int          `mapstructure:"http_max_conns_per_client"`
+	HTTPSHandshakeTimeout *string       `mapstructure:"https_handshake_timeout"`
+	RequestLimits         RequestLimits `mapstructure:"request_limits" json:"-"`
```
boxofrad, line 725: Is there a specific reason to omit this in JSON encoding? (which I think is what the `-` tag is doing)

agent/config/default.go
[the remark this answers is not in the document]
boxofrad: IMO the config builder should be responsible for turning `-1` into `rate.Inf`.

agent/config/testdata/full-config.hcl
```diff
@@ -306,6 +306,11 @@ limits {
     rpc_max_conns_per_client = 2954
     kv_max_value_size = 1234567800
     txn_max_req_len = 567800000
+    request_limits {
+        mode = "permissive"
+        read_rate = 100.0
+        write_rate = 100.0
```
boxofrad, line 312: Maybe worth making these different values to catch any conversion typos?

agent/consul/config.go
```diff
@@ -497,6 +517,10 @@ func DefaultConfig() *Config {
 
 		CheckOutputMaxSize: checks.DefaultBufSize,
 
+		RequestLimitsMode:      "permissive",
+		RequestLimitsReadRate:  -1, // ops / sec
+		RequestLimitsWriteRate: -1, // ops / sec
```
boxofrad, line 522:
```suggestion
		RequestLimitsReadRate:  rate.Inf, // ops / sec
		RequestLimitsWriteRate: rate.Inf, // ops / sec
```

agent/consul/rate/handler.go
```diff
@@ -30,14 +30,50 @@ var (
...
+	ModePermissive: "permissive",
+}
+var modeFromName = map[string]Mode{
+	"disabled":   ModeDisabled,
+	"enforcing":  ModeEnforcing,
+	"permissive": ModePermissive,
+	// If the value is not found in the persisted config file, then use the
+	// disabled default.
+	"": ModeDisabled,
+}
```
boxofrad, line 56: Here's that trick I mentioned so we don't have to maintain 2 maps 😄 

```suggestion
var modeFromName = func() map[string]Mode {
	vals := map[string]Mode{
		"": ModeDisabled,
	}
	for k, v := range modeToName {
		vals[v] = k
	}
	return vals
}()
```

agent/consul/server.go
```diff
@@ -467,10 +471,19 @@ func NewServer(config *Config, flat Deps, externalGRPCServer *grpc.Server) (*Ser
...
+			GlobalMode: rpcRate.RequestLimitsModeFromNameWithDefault(config.RequestLimitsMode),
+			GlobalReadConfig: multilimiter.LimiterConfig{
+				Rate:  config.RequestLimitsReadRate,
+				Burst: int(config.RequestLimitsReadRate) * requestLimitsBurstMultiplier,
+			},
+			GlobalWriteConfig: multilimiter.LimiterConfig{
+				Rate:  config.RequestLimitsWriteRate,
+				Burst: int(config.RequestLimitsWriteRate) * requestLimitsBurstMultiplier,
+			},
+		}, s)
```
boxofrad, line 485: Rather than duplicating this in the reload path, could we extract a conversion function?

website/content/docs/agent/config/config-files.mdx
```diff
@@ -541,6 +541,10 @@ Valid time units are 'ns', 'us' (or 'µs'), 'ms', 's', 'm', 'h'."
 
   - `http_max_conns_per_client` - Configures a limit of how many concurrent TCP connections a single client IP address is allowed to open to the agent's HTTP(S) server. This affects the HTTP(S) servers in both client and server agents. Default value is `200`.
   - `https_handshake_timeout` - Configures the limit for how long the HTTPS server in both client and server agents will wait for a client to complete a TLS handshake. This should be kept conservative as it limits how many connections an unauthenticated attacker can open if `verify_incoming` is being using to authenticate clients (strongly recommended in production). Default value is `5s`.
+  - `request_limits` - This object povides configuration for rate limiting RPC, gRPC and HTTP requests on the consul server.
```
boxofrad, line 544: I wonder if we should make it clear that it's actually the RPC/gRPC request that is limited, and HTTP requests are limited as a consequence of that.

boxofrad (approved):
This looks great! Thanks for running the gauntlet, @jmurret 🙌🏻 

I've left a few comments but please feel free to take them or leave them.

agent/agent.go
```diff
@@ -4138,16 +4144,19 @@ func (a *Agent) reloadConfigInternal(newCfg *config.RuntimeConfig) error {
...
-		RPCMaxConnsPerClient:  newCfg.RPCMaxConnsPerClient,
-		ConfigEntryBootstrap:  newCfg.ConfigEntryBootstrap,
-		RaftSnapshotThreshold: newCfg.RaftSnapshotThreshold,
-		RaftSnapshotInterval:  newCfg.RaftSnapshotInterval,
-		HeartbeatTimeout:      newCfg.ConsulRaftHeartbeatTimeout,
-		ElectionTimeout:       newCfg.ConsulRaftElectionTimeout,
-		RaftTrailingLogs:      newCfg.RaftTrailingLogs,
+		RequestLimitsMode:      newCfg.RequestLimitsMode,
+		RequestLimitsReadRate:  newCfg.RequestLimitsReadRate,
+		RequestLimitsWriteRate: newCfg.RequestLimitsWriteRate,
```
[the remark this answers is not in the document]
jmurret: This is done.

agent/config/default.go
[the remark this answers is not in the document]
jmurret: I like this!  Thank you

agent/config/config.go
jmurret: no.  good catch.  I think this is remnant of imitating another config section!  🙏

jmurret: this is fixed. 🙏

agent/config/default.go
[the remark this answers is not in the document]
jmurret: this is fixed.  thank you both!

agent/consul/server.go
jmurret: this is fixed.  🙏

website/content/docs/agent/config/config-files.mdx
jmurret: This has been fixed. Great suggestion.

agent/consul/server.go
jmurret: This has been fixed.  Thank you for the suggestion!

agent/config/testdata/full-config.hcl
jmurret: Done.  🙏

dhiaayachi (approved):
Thank you for taking into account my suggestions @jmurret! 
I recommend extracting the fix in a separate PR but other than that it LGTM!

agent/agent.go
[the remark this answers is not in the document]
jmurret: agree. will do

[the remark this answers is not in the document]
jmurret: This is now in https://github.com/hashicorp/consul/pull/15769.  It's currently in draft mode.  I'll fix up the description and changelog.

PR #15533, opened 2022-11-22
chore(ci): add auto-approve workflow for consul bot

DanStough:
### Description
The required approval is blocking using GH auto-merge with backport assistant. Rather than finding a similar solution of embedding this in the code, I thought we could globally approve PRs for the bot user.

My assumption is that giving the github-bot write permission should satisfy the requirement of having a non-author approver. If this is the case, I can remove the previous secret/user from the repo if and when this is running.

If the GitHub checks finish before this workflow runs, approval will cause the PR to merge immediately.


dhiaayachi (approved):
LGTM!

PR #15534, opened 2022-11-22
Add async reconciliation controller subpackage

andrewstucki:
### Description

This adds a port of the Kubernetes controller runtime that leverages our internal `state.EventPublisher` and `stream.SubscribeRequest` event subscription model to trigger retryable reconciliation invocations. Similar to the controller runtime, it allows for doing parallel processing of multiple requests, requeueing for later consumption after a reconciliation failure, and exponential backoff in the case of unexpected errors.

Much of the need for this is described in this [RFC](https://docs.google.com/document/d/1LRrtUpLcG2ImI_rYpdUrx9FH_63nbX25aRdzwoCwwzU/edit#).

* Unit tests were added for each component, since this package isn't currently used anywhere, it's all self-contained.

### Links

- [RFC](https://docs.google.com/document/d/1LRrtUpLcG2ImI_rYpdUrx9FH_63nbX25aRdzwoCwwzU/edit#)
- [Issue](https://hashicorp.atlassian.net/browse/NET-1673)

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/consul/controller/controller_test.go
```diff
@@ -0,0 +1,228 @@
...
+		require.NoError(t, store.EnsureConfigEntry(entryIndex, &structs.IngressGatewayConfigEntry{
+			Kind: structs.IngressGateway,
+			Name: name,
+		}))
+	}
+
+	go New(publisher, reconciler).Subscribe(&stream.SubscribeRequest{
+		Topic:   state.EventTopicIngressGateway,
+		Subject: stream.SubjectWildcard,
+	}).WithWorkers(10).Start(ctx)
```
nathancoleman, line 50: @sarahalsmiller @t-eckert I believe this is essentially what we'll need to do on startup for each type we want to control, `EventTopicAPIGateway`, `EventTopicHTTPRoute`, etc.

The more interesting part is just wiring up an implementation of `type Reconciler interface` defined elsewhere in this PR that handles reconciling each type.

t-eckert: Exactly, that's what I was fussing with as a "baby controller" for testing.

nathancoleman: BabyController™

agent/consul/controller/controller.go
```diff
@@ -0,0 +1,258 @@
...
+// queue, this is predominantly just used for testing. This should only ever be called
+// prior to running Start.
+func (c *controller) WithQueueFactory(fn func(ctx context.Context, baseBackoff time.Duration, maxBackoff time.Duration) WorkQueue) Controller {
+	c.makeQueue = fn
+	return c
+}
+
+// Start begins the Controller's main processing loop. When the given
+// context is canceled, the Controller stops processing any remaining work.
+// The Start function should only ever be called once.
```
nathancoleman, line 129: Would use `sync.Once` inside `Start( ... )` so that we don't rely on the caller to ensure they're only calling once

agent/consul/controller/controller.go
```diff
@@ -0,0 +1,258 @@
...
+				}
+				c.reconcileHandler(groupCtx, request)
+				// Done is called here because it is required to be called
+				// when we've finished processing each request
+				c.work.Done(request)
+			}
+		})
+	}
+
+	<-groupCtx.Done()
```
nathancoleman, line 190: #nit I find it odd that a func named `Start()` is blocking. I'd name it `Run()` or make it non-blocking

agent/consul/controller/controller.go
```diff
@@ -0,0 +1,258 @@
...
+				return err
+			}
+		}
+		return nil
+	default:
+		return fmt.Errorf("unhandled event type: %T", payload)
+	}
+}
+
+// enqueueEntry adds all of the given entry into the work queue
```
nathancoleman, line 211:
```suggestion
// enqueueEntry adds the given entry into the work queue. If one or more transformers
// are provided, then one transformed entry will be enqueued per transformer.
```

agent/consul/controller/controller.go
```diff
@@ -0,0 +1,258 @@
...
+func (c *controller) enqueueEntry(transformers []func(entry structs.ConfigEntry) []Request, entry structs.ConfigEntry) {
```
nathancoleman, line 212: #nit Would alias this out as `type transformer func(structs.ConfigEntry) []Request`

nathancoleman: I'd usually expect to see the args flipped here to simplify the caller

```go
func (c *controller) enqueueEntry(entry structs.ConfigEntry, transformers ...transformer)
```

agent/consul/controller/controller.go
```diff
@@ -0,0 +1,258 @@
...
+func (c *controller) WithQueueFactory(fn func(ctx context.Context, baseBackoff time.Duration, maxBackoff time.Duration) WorkQueue) Controller {
+	c.makeQueue = fn
+	return c
+}
+
+// Start begins the Controller's main processing loop. When the given
+// context is canceled, the Controller stops processing any remaining work.
+// The Start function should only ever be called once.
+func (c *controller) Start(ctx context.Context) error {
+	group, groupCtx := errgroup.WithContext(ctx)
```
nathancoleman, line 131: Could store `groupCtx` on the `controller` instead of passing it around everywhere internally

agent/consul/controller/controller.go
```diff
@@ -0,0 +1,258 @@
...
+			}
+		}
+	}
+}
+
+// reconcile wraps the reconciler in a panic handler
+func (c *controller) reconcile(ctx context.Context, req Request) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("panic: %v [recovered]", r)
```
nathancoleman, line 232: #nit I assume `[recovered]` is intended to make it clear that we dealt with it gracefully, so I'd reorder to indicate that as soon as possible instead of on the other end of the (potentially-lengthy) `r`

`panic [recovered]: %v`

agent/consul/controller/reconciler.go
```diff
@@ -0,0 +1,51 @@
...
+	Kind string
+	Name string
+	Meta *acl.EnterpriseMeta
+}
+
+// RequeueAfterError is an error that allows a Reconciler to override the
+// exponential backoff behavior of the Controller, rather than applying
+// the backoff algorithm, returning a RequeueAfterError will cause the
+// Controller to reschedule the Request at a given time in the future.
+type RequeueAfterError time.Duration
```
nathancoleman, line 24: Does it make sense to include a cause on this error?

agent/consul/controller/rate_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	for i := 0; i < 2; i++ {
+		limiter.NextRetry(request)
+	}
+	require.Equal(t, 4*time.Minute, limiter.NextRetry(request))
+
+	for i := 0; i < 1000; i++ {
+		limiter.NextRetry(overflow)
+	}
+	require.Equal(t, 1000*time.Hour, limiter.NextRetry(overflow))
```
nathancoleman, line 57: It takes a lot of thought to determine what a set of assertions like this is intended to test.

I usually use `t.Run` or comments to lower the bar for readers

nathancoleman (changes requested):
Nice work! Excited to see this

agent/consul/controller/controller.go
andrewstucki: 👍  I'm fine with renaming `Run`, that's used elsewhere in our structs that are meant to have entrypoints invoked in long-running goroutines.

agent/consul/controller/reconciler.go
andrewstucki: I don't think so, in this particular case this is just meant as a mechanism for controlling the explicit retry semantics of the controller -- unless we add in a bunch of internal logging around queueing/retries then a cause won't really ever get used. I personally am ok with just delegating those sorts of logging bits to someone implementing a reconciler -- just throw a logging statement before explicitly returning a `RequeueAfterError`.

agent/consul/controller/rate_test.go
andrewstucki: I can add some comments to the second assertions here for clarity, but it's testing that the rate limiter is incrementing the backoff time, but doesn't overflow the max backoff that it's given when initialized.

agent/consul/controller/controller.go
nathancoleman: 👍 Either approach is an improvement IMO

andrewstucki: Cool, added some panic-based assertions (with tests) around `Run` and the configuration functions to enforce this requirement in the concrete implementation.

agent/consul/controller/controller.go
andrewstucki: Added a publically exposed alias since this gets propagated down from `Subscribe`.

agent/consul/controller/controller.go
andrewstucki: So, the only place this really gets handed down to any struct functions is in the thin wrappers around the call to the reconciler's `Reconcile` function (`reconcileHandler`/`reconcile`). Generally I'd prefer not storing the context unless the context is passed in as part of the initialization of the controller itself.

agent/consul/controller/controller.go
```diff
@@ -0,0 +1,289 @@
...
+	"github.com/hashicorp/consul/agent/consul/state"
+	"github.com/hashicorp/consul/agent/consul/stream"
+	"github.com/hashicorp/consul/agent/structs"
+	"golang.org/x/sync/errgroup"
+)
+
+// much of this is a re-implementation of
+// https://github.com/kubernetes-sigs/controller-runtime/blob/release-0.13/pkg/internal/controller/controller.go
+
+// Transformer is a function that takes on type of config entry that has changed
```
nathancoleman, line 19:
```suggestion
// Transformer is a function that takes one type of config entry that has changed
```

nathancoleman (approved):

PR #15535, opened 2022-11-23
Add sections for different upstream annotation formats

Ranjandas:
### Description
Add subsections for different upstream annotation formats for better readability and also to make it easier to link from other documentation (eg: Learn Guide on cluster peering)

NOTE: The GitHub markdown preview seems to be different from the local preview 

### PR Checklist

* [ ] updated test coverage
* [x] external-facing docs updated
* [x] not a security concern

### DIFF
--- a/website/content/docs/k8s/annotations-and-labels.mdx
+++ b/website/content/docs/k8s/annotations-and-labels.mdx
@@ -64,25 +64,27 @@ The following Kubernetes resource annotations could be used on a pod to control 
   port value such as "8080". This is the port of the _service_, the proxy
   public listener will listen on a dynamic port.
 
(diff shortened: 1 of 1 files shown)


david-yu:
Hi @Ranjandas I like where you are headed here. Perhaps we need to re-org the annotations to make it easier to link to per use case? Is that the intent of the PR?

Ranjandas:
@david-yu, Yes, I planned to categorize/re-org the annotations but ended up only doing the `upstream` annotations. The main reason for cutting down the effort was that my local previews of the new website were extremely slow, and in addition, I wasn't sure which category some of them should fall into.

I can work on a draft PR with full re-org and send it for review if you are happy with the plan.

PR #15539, opened 2022-11-23
server: add placeholder glue for rate limit handler

boxofrad:
### Description
This PR adds a no-op implementation of the rate-limit handler and exposes it on the `consul.Server` struct.

It allows us to start working on the net/rpc and gRPC interceptors and config (re)loading logic, without having to implement the full handler up-front.


agent/consul/rate/handler.go
```diff
@@ -0,0 +1,103 @@
...
+// Operation the client is attempting to perform.
+type Operation struct {
+	// Name of the RPC endpoint (e.g. "Foo.Bar" for net/rpc and "/foo.service/Bar" for gRPC).
+	Name string
+
+	// SourceAddr is the client's (or forwarding server's) IP address.
+	SourceAddr net.Addr
+
+	// Type of operation to be performed (e.g. read or write).
+	Type OperationType
```
boxofrad, line 45: The RFC mentions keeping a mapping of RPC names to operation type (i.e. read vs write).

I wasn't sure whether we wanted to implement this in the generic handler or in the interceptors themselves, but as the RFC also mentions possibly automating this using protobuf annotations I exposed this on the `Operation` struct incase we handle net/rpc and gRPC methods separately.

We won't be adding any more net/rpc handlers, so I think it's ok to have a static list for these.

Also, with the upcoming controller-based architecture we can implement rate-limiting at the state store which removes this problem entirely (as it's very clear what is a read vs write).

agent/consul/rate/handler.go
```diff
@@ -0,0 +1,103 @@
...
+const (
+	// OperationTypeRead represents a read operation.
+	OperationTypeRead OperationType = iota
+
+	// OperationTypeWrite represents a write operation.
+	OperationTypeWrite
+)
+
+// Operation the client is attempting to perform.
+type Operation struct {
```
boxofrad, line 37: We might also want to add a `Category` field to this struct (e.g. `CategoryKV`, `CategoryCatalog`).

agent/consul/server.go
```diff
@@ -462,6 +466,13 @@ func NewServer(config *Config, flat Deps, externalGRPCServer *grpc.Server) (*Ser
 		Logger:   logger.Named("hcp_manager"),
 	})
 
+	// TODO(NET-1380, NET-1381): thread this into the net/rpc and gRPC interceptors.
+	s.incomingRPCLimiter = rpcRate.NewHandler(rpcRate.HandlerConfig{
```
boxofrad, line 470: We may actually need to do this in agent/agent.go because the external gRPC server is shared between client and server agents 🤔

agent/consul/rate/handler.go
```diff
@@ -0,0 +1,103 @@
...
+// exhausted (e.g. log and allow, or reject).
+type Mode int
+
+const (
+	// ModePermissive causes the handler to log the rate-limited operation but
+	// still allow it to proceed.
+	ModePermissive Mode = iota
+
+	// ModeEnforce causes the handler to reject the rate-limted operation.
+	ModeEnforce
```
jmurret, line 22: This comment is not particular to this PR but is a small nit that spans the RFCs that is hopefully a quick discussion and does not devolve into "Naming Things Is Hard!".

TLDR: I'd like to change "enforce" to "enforced" so both modes are adjectives.

LONGER CONTEXT:
The meat RFC refers to Mode and Action but eventually settles on Mode. To cut to the chase, Mode should have adjectives - permissive, strict, enforced, etc. If we named this action, the values would be verbs- enforce, permit, allow, etc.

I think the simple enforce -> enforced is easy and does not open the door too much.

agent/consul/rate/handler.go
```diff
@@ -0,0 +1,158 @@
...
+func (gl globalLimit) Key() []byte {
+	return multilimiter.Key(gl.ConfigKey(), []byte("limiter"))
+}
+
+// ConfigKey is the key of the multilimiter tree node in which the config lives.
+//
+// TODO: we have to do this beacause the multilimiter doesn't currently support
+// setting config directly on a leaf node, which we were eventually going to do
+// for per-identity/per-tenant limits. Maybe we should do that sooner?
+func (gl globalLimit) ConfigKey() []byte { return []byte(gl) }
```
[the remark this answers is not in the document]
dhiaayachi: See my last commit, it should be available now 😆 !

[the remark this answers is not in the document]
boxofrad: Wow! Like lightning ⚡ 😆

agent/consul/rate/handler.go
dhiaayachi: The lexicologie in the RFC was based on the selinux terms. I double checked what selinux use for the config and found out it actually use:
- disabled
- permissive
- enforcing

I'm not sure what is the reasoning behind that, If any, but I thought I would mention it. 

That said I'm good with naming this enforced too.

agent/consul/server.go
dhiaayachi, line 470: I'm not sure this is the right place for grpc, as at that level the grpc server is already created and we need to add the interceptor at the creation.

agent/consul/rate/handler.go
```diff
@@ -0,0 +1,158 @@
...
+	// GlobalReadConfig configures the global rate limiter for read operations.
+	GlobalReadConfig multilimiter.LimiterConfig
+}
+
+type HandlerDelegate interface {
+	// IsLeader is used to determine whether the operation is being performed
+	// against the cluster leader, such that if it can _only_ be performed by
+	// the leader (e.g. write operations) we don't tell clients to retry against
+	// a different server.
+	IsLeader() bool
```
dhiaayachi, line 93: I guess this will in the future also have isForwarded(), to check if a request is forwarded from another server?

agent/consul/rate/handler.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// SourceAddr is the client's (or forwarding server's) IP address.
+	SourceAddr net.Addr
+
+	// Type of operation to be performed (e.g. read or write).
+	Type OperationType
+}
+
+// Handler enforces rate limits for incoming RPCs.
+type Handler struct {
```
dhiaayachi, line 65: We will need to have a logger and be able to emit metrics in the handler.

dhiaayachi (commented):
Thank you for putting this together @boxofrad. It's cool to see things fit together 😄 !

I added few thoughts but I think this is definitely going the right direction.

agent/consul/rate/handler.go
boxofrad: Yep! I'm aware that we're planning to do some buffering and wasn't sure whether the handler would do this, or we'd encapsulate it in a wrapper around the logger (which would be my preference).

Anyway, we can tackle that in a future PR 🙇🏻

boxofrad: Yep! I'm aware that we're planning to do some buffering and wasn't sure whether the handler would do this, or we'd encapsulate it in a wrapper around the logger (which would be my preference).

Anyway, we can tackle that in a future PR 🙇🏻

agent/consul/rate/handler.go
boxofrad: Good question! I guess we'd need to compare the source IP to the known list of server addresses (for net/rpc at least) which would be something we'd put on the delegate, yeah.

agent/consul/server.go
boxofrad: It's a little awkward for sure. An alternative would be to put it on `BaseDeps` which doesn't feel right either, as we'll never actually use it on client agents.

Here's how I'm using it in my draft gRPC PR:

https://github.com/hashicorp/consul/blob/c6c6ec999d06263e6c997afa14d8b955add9297a/agent/agent.go#L565-L578

https://github.com/hashicorp/consul/blob/c6c6ec999d06263e6c997afa14d8b955add9297a/agent/consul/server.go#L1830-L1833

agent/consul/rate/handler.go
boxofrad: Nice! Have changed to `enforcing`.

agent/consul/rate/handler.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// Run the limiter cleanup routine until the given context is canceled.
+//
+// Note: this starts a goroutine.
+func (h *Handler) Run(ctx context.Context) {
+	h.limiter.Run(ctx)
+}
+
+// Allow returns an error if the given operation is now allowed to proceed
```
analogue, line 119: s/now/not

jmurret (approved):
Nice work @boxofrad.  Thank you for all of the hard work and thought you have put into this rate limiting project!

Issue #15246, opened 2022-11-03
Large number of error message RPC deadlines after the introduction of rpc_client_timeout


blockmar:
#### Overview of the Issue

Large number of error message regarding RPC deadlines after the introduction of https://github.com/hashicorp/consul/pull/14965

Both from Consul agents and from agents in other languages such as Spring Cloud Consul (Java)

Our configuration remains unchanged since 1.13.2 since this is a patch update.

```
Nov  3 15:48:09 ip-10-X-X-X.ec2.internal consul[2124]: agent.client: RPC failed to server: method=Catalog.ListServices server=10.X.Y.Y:8300 error="rpc error making call: i/o deadline reached"
```

```
com.ecwid.consul.v1.OperationException: OperationException(statusCode=500, statusMessage='Internal Server Error', statusContent='rpc error making call: i/o deadline reached')
```

Our systems still seem to be working as expected but our logs are getting filled with messages like the two examples above. This risks hiding real issues and triggering false alarms and causing alarm fatigue.

#### Reproduction Steps

Upgrade from Consul 1.13.2 to 1.13.3

### Consul info for both Client and Server

_No access to a token allowed to perform `agent:read`_

### Operating system and Environment details

Ubuntu Linux 20.04

### Log Fragments

```
Nov  3 15:48:09 ip-10-X-X-X.ec2.internal consul[2124]: agent.client: RPC failed to server: method=Catalog.ListServices server=10.X.Y.Y:8300 error="rpc error making call: i/o deadline reached"
```

```
com.ecwid.consul.v1.OperationException: OperationException(statusCode=500, statusMessage='Internal Server Error', statusContent='rpc error making call: i/o deadline reached')
```

huikang:
Hi, @blockmar , thanks for reporting the issue.

> Both from Consul agents and from agents in other languages such as Spring Cloud Consul (Java)

Just want to clarify that if all agents are upgraded to 1.13.3? Could you elaborate a bit on "agents in other languages such as Spring Cloud Consul (Java)"?

blockmar:
No problem, all participating Consul instances are running 1.13.3 and where updated at the same time by rotating servers, replacing them one by one. All where previously running 1.13.2.

My reference to agents in other languages was that we use Spring Cloud Consul in our services.
https://github.com/spring-cloud/spring-cloud-consul

They in turn use this API library https://github.com/Ecwid/consul-api

[...]

PR #15541, opened 2022-11-23
Use rpcHoldTimeout to calculate blocking timeout

kisunji:
### Description
Fixes https://github.com/hashicorp/consul/issues/15246

Reverts a bug I introduced while refactoring:
https://github.com/hashicorp/consul/pull/14965/files#diff-8f12aa7f72e8647fddfac936ab541982aae517babd3064a581bbcdcc2595c2dcL347-R360

Note that Timeout included `rpcHoldTimeout` but I extracted it out of `Timeout` into `HasTimedOut`. But we still need to consider it when calculating `BlockingTimeout`.

It's hard to write a test for this due to timeouts being very timing + random jitter dependent. It should not break any existing tests and at most it adds 7s to existing timeouts.

The problem I'm trying to solve is:
```
/v1/catalog/services?wait=2s&index=16165

client uses BlockingTimeout to calculate read timeout
2s + 2s/16 = 2.125s

server adds max possible jitter
2s + 2s/16 = 2.125s
```

if timeouts are nearly identical it is likelier for the client to timeout the connection before the server has had a chance to respond (or time out the blocking query).

With this PR:
```
GET /v1/catalog/services?wait=2s&index=16165

client uses (*QueryOptions).BlockingTimeout to calculate read timeout
2s + 2s/16 + 7s = 9.125s
             ^ RPCHoldTimeout buffer

server adds max possible jitter in (*Server).rpcQueryTimeout
2s + 2s/16 = 2.125s
```

Server will always be the one to apply timeout and return from a blocking query early

### DIFF
--- a/agent/consul/client_test.go
+++ b/agent/consul/client_test.go
@@ -532,6 +532,7 @@ func newDefaultDeps(t *testing.T, c *Config) Deps {
 		Datacenter:       c.Datacenter,
 		DefaultQueryTime: c.DefaultQueryTime,
 		MaxQueryTime:     c.MaxQueryTime,
+		RPCHoldTimeout:   c.RPCHoldTimeout,
 	}
 	connPool.SetRPCClientTimeout(c.RPCClientTimeout)
 	return Deps{
@@ -881,8 +882,9 @@ func TestClient_RPC_Timeout(t *testing.T) {
 		c.Datacenter = "dc1"
 		c.NodeName = uniqueNodeName(t.Name())
 		c.RPCClientTimeout = 10 * time.Millisecond
-		c.DefaultQueryTime = 100 * time.Millisecond
+		c.DefaultQueryTime = 150 * time.Millisecond
 		c.MaxQueryTime = 200 * time.Millisecond
+		c.RPCHoldTimeout = 50 * time.Millisecond
 	})
 	joinLAN(t, c1, s1)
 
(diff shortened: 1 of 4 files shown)


agent/pool/pool.go
```diff
@@ -630,8 +633,9 @@ func (p *ConnPool) rpc(dc string, nodeName string, addr net.Addr, method string,
 	if bq, ok := args.(BlockableQuery); ok {
 		blockingTimeout := bq.BlockingTimeout(p.MaxQueryTime, p.DefaultQueryTime)
 		if blockingTimeout > 0 {
-			// override the default client timeout
-			timeout = blockingTimeout
+			// Override the default client timeout but add RPCHoldTimeout
+			// as a buffer for retries during leadership changes.
+			timeout = blockingTimeout + p.RPCHoldTimeout
```
kisunji, line 638: If all this timeout calculation is confusing, in a nutshell I am re-adding RPCHoldTimeout that I removed in https://github.com/hashicorp/consul/pull/14965/files#diff-8f12aa7f72e8647fddfac936ab541982aae517babd3064a581bbcdcc2595c2dcL347-R360

kisunji: EXCEPT in the non-blocking case, which is controlled by a different config `rpc_client_timeout` and should not have `rpc_hold_timeout` included (since it's only relevant for blocking queries)

agent/consul/client_test.go
```diff
@@ -931,15 +933,37 @@ func TestClient_RPC_Timeout(t *testing.T) {
...
+		// adds maximum jitter) so the client will error first.
 		err := c1.RPC("Long.Wait", &structs.NodeSpecificRequest{
 			QueryOptions: structs.QueryOptions{
 				MinQueryIndex: 1,
 				MaxQueryTime:  20 * time.Millisecond,
 			},
 		}, &out)
 		require.Error(t, err)
-		require.Contains(t, err.Error(), "rpc error making call: i/o deadline reached")
+		require.ErrorContains(t, err, "rpc error making call: i/o deadline reached")
```
wilkermichael, line 967: oh that's nice, I didn't know about `ErrorContains`

agent/pool/pool.go
wilkermichael: This also guarantees that the server timeout is longer than the clients in case of leader rotation, which I guess is the whole point?

wilkermichael (approved):
LGTM, great job

agent/pool/pool.go
kisunji: Other way around: assuming client and server calculate timeout the same way, it guarantees clients will always have the longer timeout (servers will timeout first, which is desired)

wilkermichael: oh right, yep, reversed that in my head

PR #15542, opened 2022-11-24
Add Config example for Config Entry Bootstrap

Ranjandas:
### Description
Add an example configuration on bootstrapping Config Entry using the agent configuration.

This was found from an old GH issue created by @mkeeler.

I have tested this using the following steps:

1. Create the config entry
```
cat <<EOF > agent.hcl
config_entries {
  bootstrap {
    kind = "proxy-defaults"
    name = "global"
    config {
       envoy_dogstatsd_url = "udp://127.0.0.1:9125"
    }
  }
  bootstrap {
    kind = "service-defaults"
    name = "dashboard"
    protocol = "http"
  }
}
EOF
```

2. Launch a dev agent

```
$ consul agent -dev -config-file agent.hcl
```

3. Verify both `service-defaults` and `proxy-defaults`.

```
consul config read -kind proxy-defaults -name global
{
    "Kind": "proxy-defaults",
    "Name": "global",
    "Partition": "default",
    "Namespace": "default",
    "TransparentProxy": {},
    "Config": {
        "envoy_dogstatsd_url": "udp://127.0.0.1:9125"
    },
    "MeshGateway": {},
    "Expose": {},
    "CreateIndex": 15,
    "ModifyIndex": 15
}
```

```
consul config read -kind service-defaults -name dashboard
{
    "Kind": "service-defaults",
    "Name": "dashboard",
    "Partition": "default",
    "Namespace": "default",
    "Protocol": "http",
    "TransparentProxy": {},
    "MeshGateway": {},
    "Expose": {},
    "CreateIndex": 17,
    "ModifyIndex": 17
}
```

### PR Checklist

* [x] not a security concern

### DIFF
--- a/website/content/docs/agent/config/config-files.mdx
+++ b/website/content/docs/agent/config/config-files.mdx
@@ -434,6 +434,52 @@ Valid time units are 'ns', 'us' (or 'µs'), 'ms', 's', 'm', 'h'."
     See the [configuration entry docs](/docs/agent/config-entries) for more
     details about the contents of each entry.
 
+  <CodeTabs>
+
+  ```hcl
+  config_entries {
+    bootstrap {
+      kind = "proxy-defaults"
+      name = "global"
+      config {
+         envoy_dogstatsd_url = "udp://127.0.0.1:9125"
+      }
+    }
+    bootstrap {
+      kind = "service-defaults"
+      name = "dashboard"
+      protocol = "http"
+    }
+  }
+  ```
+
+  ```json
+  {
+    "config_entries": [
+      {
+        "bootstrap": [
+          {
+            "kind": "proxy-defaults",
+            "name": "global",
+            "config": [
+              {
+                "envoy_dogstatsd_url": "udp://127.0.0.1:9125"
+              }
+            ]
+          },
+          {
+            "kind": "service-defaults",
+            "name": "dashboard",
+            "protocol": "http"
+          }
+        ]
+      }
+    ]
+  }
+  ```
+
+  </CodeTabs>
+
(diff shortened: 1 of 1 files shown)


trujillo-adam (changes requested):
Thanks for adding the example, but I think the content makes more sense in this page: https://developer.hashicorp.com/consul/docs/agent/config-entries

There is already a subtopic on how to bootstrap where this example could go. It would need an introduction explaining what the example accomplishes. Please let me know how I can help draft the content.

The agent configuration page is already overwhelming, so it would be easy for example configurations to be overlooked. I know we already have a some examples in this page, but we have a long term initiative to organize content by type(reference, usage, etc.).

jkirschner-hashicorp:
I agree with @trujillo-adam that something like this, if we add it, belongs on the config entries page. We could cross-reference is from the agent config option ("For more details, refer to ___").

However, there's also a separate conversation happening internally on the conditions under which we'd even recommend using the `config_entries.bootstrap` agent config option. My current understanding is that this config option isn't the way most users should be managing config entries. We need provide that context so we don't accidentally guide (or implicitly suggest) users down a path that's unsuitable for them.

jkirschner-hashicorp:
@Ranjandas : I'm super appreciative of you documenting hidden knowledge as you go :) Keep up the great work. I'll follow up internally to see how we should contextualize this example.

brucelok:
it would be helpful if the [documentation](https://developer.hashicorp.com/consul/docs/agent/config/config-files#config_entries) can show some examples. It wasn't working for me, it took times for me to figure out

PR #15546, opened 2022-11-24
Use backport-compatible assertion

kisunji:


agent/consul/client_test.go
```diff
@@ -964,6 +964,6 @@ func TestClient_RPC_Timeout(t *testing.T) {
 			},
 		}, &out)
 		require.Error(t, err)
-		require.ErrorContains(t, err, "rpc error making call: i/o deadline reached")
+		require.Contains(t, err.Error(), "rpc error making call: i/o deadline reached")
```
kisunji, line 967: Older versions of `testify/require` do not have this function so it's hard to backport

curtbushko (approved):

PR #15550, opened 2022-11-24
grpc: add rate-limiting middleware

boxofrad:
### Description

This PR implements the gRPC middleware for rate-limiting as a `tap.ServerInHandle` function (executed before the request is unmarshaled).

Mappings between gRPC methods and their operation type is handled by a `protoc` plugin introduced by #15564.


agent/grpc-middleware/rate_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/health"
+	healthpb "google.golang.org/grpc/health/grpc_health_v1"
+	"google.golang.org/grpc/status"
+
+	"github.com/hashicorp/consul/agent/consul/rate"
+)
+
+func TestServerRateLimiterMiddleware_Integration(t *testing.T) {
+	limiter := NewMockRateLimiter(t)
```
jmurret, line 20: 👏

jmurret (commented):
👏 thank you for all of the great work!  this looks really good. I had one meta question that I left about handling errors and not blocking requests for errors in consul rate limiting code.

agent/grpc-middleware/rate.go
```diff
@@ -0,0 +1,55 @@
...
+// been decoded).
+func ServerRateLimiterMiddleware(limiter RateLimiter) tap.ServerInHandle {
+	return func(ctx context.Context, info *tap.Info) (context.Context, error) {
+		peer, ok := peer.FromContext(ctx)
+		if !ok {
+			// This should never happen!
+			return ctx, status.Error(codes.Internal, "gRPC rate limit middleware unable to read peer")
+		}
+
+		err := limiter.Allow(rate.Operation{
```
boxofrad, line 26: Just realised that we should probably special-case the xDS endpoint because it implements its own concurrency limiting.

agent/grpc-middleware/rate.go
```diff
@@ -0,0 +1,55 @@
...
+			SourceAddr: peer.Addr,
+			// TODO: operation type.
+		})
+
+		switch {
+		case err == nil:
+			return ctx, nil
+		case errors.Is(err, rate.ErrRetryElsewhere):
+			return ctx, status.Error(codes.ResourceExhausted, err.Error())
+		default:
```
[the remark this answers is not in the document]
boxofrad: Great call! I think it'd be ok to fail with `codes.Internal` for any unknown error, as its a bug in Consul that should be surfaced during our testing.

You raise a good point about `nil` pointers though, as this will presumably be called before/in a different code-path to our recovery interceptor so we probably need to handle panics itself.

agent/grpc-middleware/rate.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// NullRateLimiter returns a RateLimiter that allows every operation.
+func NullRateLimiter() RateLimiter {
+	return nullRateLimiter{}
+}
+
+type nullRateLimiter struct{}
+
+func (nullRateLimiter) Allow(rate.Operation) error { return nil }
```
dhiaayachi, line 65: Can we add a static check to make sure the nullRateLimiter Satisfy the RateLimiter interface?

agent/grpc-middleware/rate.go
```diff
@@ -0,0 +1,65 @@
...
+		default:
+			return ctx, status.Error(codes.Unavailable, err.Error())
+		}
+	}
+}
+
+//go:generate mockery --name RateLimiter --inpackage
+type RateLimiter interface {
+	Allow(rate.Operation) error
+}
```
dhiaayachi, line 56: Should we move this in the `multilimiter` package so we can use it in other tests?

agent/grpc-middleware/rate.go
```diff
@@ -0,0 +1,65 @@
...
+// been decoded).
+func ServerRateLimiterMiddleware(limiter RateLimiter, panicHandler recovery.RecoveryHandlerFunc) tap.ServerInHandle {
+	return func(ctx context.Context, info *tap.Info) (_ context.Context, retErr error) {
+		// This function is called before unary and stream RPC interceptors, so we
+		// must handle our own panics here.
+		defer func() {
+			if r := recover(); r != nil {
+				retErr = panicHandler(r)
+			}
+		}()
```
dhiaayachi, line 28: 🎉

agent/grpc-middleware/rate.go
```diff
@@ -0,0 +1,65 @@
...
+
+		peer, ok := peer.FromContext(ctx)
+		if !ok {
+			// This should never happen!
+			return ctx, status.Error(codes.Internal, "gRPC rate limit middleware unable to read peer")
+		}
```
dhiaayachi, line 34: What about the case where we do internal gRPC (http to gRPC inside the same server), does it have the correct peer info?

agent/grpc-middleware/rate.go
dhiaayachi: Not sure I understand, the xDS concurrency limiter is quiet different,isn'it? 🤔

dhiaayachi (changes requested):
Thank you for working on this @boxofrad I added some questions/suggestions.

agent/grpc-middleware/rate.go
boxofrad: We kinda have one in that the `NullRateLimiter` method's return-type is the interface, so if we change the interface without updating the struct it'll fail to compile 😄

agent/grpc-middleware/rate.go
boxofrad: No strong opinions, but I generally like to define interfaces at the point-of-use ([see here](https://blog.chewxy.com/2018/03/18/golang-interfaces/)).

agent/grpc-middleware/rate.go
boxofrad: Yeah, I'd say we should just _not_ rate-limit the xDS endpoint here, and trust the xDS server to do its own limiting.

Particularly because xDS sessions are long-lived and touch many different resources (so what kind of limit would we even apply here?)

agent/grpc-middleware/rate.go
boxofrad: From what I can tell, there's no equivalent of the net/rpc in-memory codec in our gRPC stack (we dial a real TCP connection) so the peer here will be whichever server agent is handling the HTTP request.

I guess that's probably ok in this phase?

jmurret: If we do need this, can we avoid blocking the PR on it?  So can we add either a Phase 1 or Phase 2 follow up task if this is needed?

agent/grpc-middleware/rate.go
```diff
@@ -0,0 +1,72 @@
...
+
+		err := limiter.Allow(rate.Operation{
+			Name:       info.FullMethodName,
+			SourceAddr: peer.Addr,
+			// TODO: operation type.
```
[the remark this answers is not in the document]
boxofrad: This is implemented in #15564 🙇🏻‍♂️

agent/grpc-middleware/rate.go
dhiaayachi: +1 to not blocking the PR over this in all cases!

agent/grpc-middleware/rate.go
dhiaayachi: I think this is reasonable. In all cases the xDS endpoint would end up using data from watches which perform RPC calls, so I guess that would be enough to protect the server.

dhiaayachi (approved):
🚀

PR #15554, opened 2022-11-25
[OSS] Add boilerplate for proto files implementing BlockableQuery

kisunji:


.changelog/_3729.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+namespace: **(Enterprise Only)** Fixed a bug that caused blocking queries during namespace replication to timeout
```
cthain, line 2: nit: Should use present tense for changelog entries.
```suggestion
namespace: **(Enterprise Only)** Fix a bug that caused blocking queries during namespace replication to timeout
```

cthain (approved):
Looks good. Just a minor nitpick on the changelog.

PR #15564, opened 2022-11-28
WIP: protoc plugin for generating gRPC rate limit

boxofrad:


proto-public/annotations/ratelimit/ratelimit.proto
```diff
@@ -0,0 +1,22 @@
...
+  WRITE = 2;
+}
+
+// Spec describes the kind of rate limit that will be applied to this RPC.
+message Spec {
+  OperationType operation_type = 1;
+}
+
+extend google.protobuf.MethodOptions {
+  Spec spec = 8300;
```
boxofrad, line 21: This number is arbitrary, it just needs to not conflict with other extensions. I chose it because it's our server port 😄

proto-public/pbacl/acl.proto
```diff
@@ -2,13 +2,23 @@ syntax = "proto3";
 
 package hashicorp.consul.acl;
 
+import "proto-public/annotations/ratelimit/ratelimit.proto";
```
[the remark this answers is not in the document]
mkeeler: It does seem to break it with our current proto file layout. However I have a branch going where I don't think it would be broken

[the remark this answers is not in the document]
mkeeler: Basically moving towards more independent modules that we generate at the root of the tree. So you would reference other local proto files without the proto-public prefix.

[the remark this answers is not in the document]
dhiaayachi: If we are allowing external consumers of our API wouldn't it be a use case for buf schema registery?

.gitignore
```diff
@@ -63,3 +63,6 @@ override.tf.json
 # Ignore CLI configuration files
 .terraformrc
 terraform.rc
+
+proto-public/*/.ratelimit
+proto/*/.ratelimit
```
dhiaayachi, line 68: I'm ok with using `.ratelimit` as a suffix but I would prefer something more generic like `.tmp` or `.temp` so we don't end up with a list of suffixes in the future.

agent/grpc-middleware/rate.go
```diff
@@ -33,10 +33,15 @@ func ServerRateLimiterMiddleware(limiter RateLimiter, panicHandler recovery.Reco
 			return ctx, status.Error(codes.Internal, "gRPC rate limit middleware unable to read peer")
 		}
 
+		operationType, ok := rpcRateLimitSpecs[info.FullMethodName]
+		if !ok {
+			logger.Warn("failed to determine which rate limit to apply to RPC", "rpc", info.FullMethodName)
+		}
```
dhiaayachi, line 39: Should we return here to avoid having to handle the empty string case in Allow?

build-support/scripts/protobuf.sh
```diff
@@ -123,7 +128,7 @@ function postprocess_protobuf_code {
 function generate_mog_code {
     local mog_order
 
-    mog_order="$(go list -tags "${GOTAGS}" -deps ./proto/pb... | grep "consul/proto")"
+    mog_order="$(go list -tags "${GOTAGS}" -deps ./proto/pb... | grep "consul/proto/")"
```
dhiaayachi, line 131: is adding a `/` here intensional?

internal/tools/protoc-gen-consul-rate-limit/postprocess/main.go
```diff
@@ -0,0 +1,190 @@
...
+	if err != nil {
+		return err
+	}
+	if err := os.WriteFile(outputPath, ossSource, 0666); err != nil {
+		return fmt.Errorf("failed to write output file: %s - %w", outputPath, err)
+	}
+
+	// ent should only be non-zero in the enterprise repository.
+	if len(ent) > 0 {
+		entSource, err := generateENT(ent)
```
dhiaayachi, line 69: Are the ent file will be an addition to the same map?

internal/tools/protoc-gen-consul-rate-limit/postprocess/main.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func generateENT(specs []spec) ([]byte, error) {
+	var output bytes.Buffer
+	output.WriteString(entTags)
+	output.WriteString(fileHeader)
+
+	output.WriteString("func init() {\n")
+	for _, spec := range specs {
+		fmt.Fprintf(&output, `rpcRateLimitSpecs["%s"] = %s`, spec.MethodName, spec.GoOperationType())
```
dhiaayachi, line 171: This answer my question about addition to the same map 😆

dhiaayachi (commented):
I have few question but nothing blocking. I need to double check the list of READ/WRITE calls before giving a final ✅

build-support/scripts/protobuf.sh
boxofrad: Yep! It causes us to only match `proto` and not `proto-public`.

agent/grpc-middleware/rate.go
boxofrad: Sounds good 👍🏻

agent/grpc-middleware/rate_limit_mappings.gen.go
```diff
@@ -0,0 +1,27 @@
+// generated by protoc-gen-consul-rate-limit; DO NOT EDIT.
+package middleware
+
+import "github.com/hashicorp/consul/agent/consul/rate"
+
+var rpcRateLimitSpecs = map[string]rate.OperationType{
```
jmurret, line 6: 👍 👏

jmurret (approved):
Nice.  I don't have any blocking comments.  We should resolve the conversation with Matt and Dhia before merging.

internal/tools/protoc-gen-consul-rate-limit/main.go
```diff
@@ -0,0 +1,139 @@
...
+
+	"google.golang.org/protobuf/compiler/protogen"
+	"google.golang.org/protobuf/proto"
+
+	"github.com/hashicorp/consul/proto-public/annotations/ratelimit"
+)
+
+const (
+	outputFileName = ".ratelimit.tmp"
+
```
analogue, line 48: I'm curious why `.ratelimit.tmp` is a hidden file and is it intentional that they never get deleted? I'd usually tie this sort of thing back to a `make clean` but that doesn't exist :laugh:

agent/grpc-middleware/rate_limit_mappings.gen.go
```diff
@@ -0,0 +1,27 @@
...
+	"/hashicorp.consul.acl.ACLService/Login":                                     rate.OperationTypeWrite,
+	"/hashicorp.consul.acl.ACLService/Logout":                                    rate.OperationTypeWrite,
+	"/hashicorp.consul.connectca.ConnectCAService/Sign":                          rate.OperationTypeWrite,
+	"/hashicorp.consul.connectca.ConnectCAService/WatchRoots":                    rate.OperationTypeRead,
+	"/hashicorp.consul.dataplane.DataplaneService/GetEnvoyBootstrapParams":       rate.OperationTypeRead,
+	"/hashicorp.consul.dataplane.DataplaneService/GetSupportedDataplaneFeatures": rate.OperationTypeRead,
+	"/hashicorp.consul.dns.DNSService/Query":                                     rate.OperationTypeRead,
+	"/hashicorp.consul.internal.operator.OperatorService/TransferLeader":         rate.OperationTypeWrite,
```
dhiaayachi, line 14: I think this should not be rate limited. We have 2 choices here:

- add another Type (eg: bypass) 
- remove it from the list and assume if not in the list it means bypass

I vote for option 1 if it's not too much work

dhiaayachi (changes requested):
I have 1 blocking comment around not rate limiting operator endpoint as we can imagine a scenario where someone would need it while the server is heavily rate limited (outage situation).

Other than that, I think this is good to go and we should iterate on the proto dependency in a separate PR.

agent/grpc-middleware/rate_limit_mappings.gen.go
boxofrad: Yep makes sense. It looks like we added `OperationTypeExempt` in the net/rpc mapping, so I'll adopt that here.

dhiaayachi (approved):
Great work @boxofrad !!
🚢

boxofrad:
Here's [a patch](https://go.hashi.co/consul-pr-15564-enterprise) with the enterprise mappings, for when this PR is merged and the automatic merge fails 🔗

PR #15595, opened 2022-11-29
integ-test: test consul upgrade from the snapshot of a running cluster

huikang:
### Description
This PR adds an upgrade test case to the integration test suite. This particular test case verifies consul can be upgraded from the snapshot of an supported version, i.e., _n-1_ and _n-2_.

- The snapshot includes service mesh
- The test case is `func TestStandardUpgradeToTarget_fromLatest(t *testing.T)`

This upgrade test can catch the data incompatibility type of upgrade bug, such as  https://github.com/hashicorp/consul/pull/14149
- [Link](https://github.com/hashicorp/consul/pull/15595/files#diff-fe79f36c4f2afab65044e1713c85af4d00f8d8d9bc6b18d85b5ec8187e72a142R28-R32) to the test case in `TestStandardUpgradeToTarget_fromLatest`

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/test/integration/consul-container/libs/agent/builder.go
+++ b/test/integration/consul-container/libs/agent/builder.go
@@ -5,6 +5,7 @@ import (
 	"path/filepath"
 
 	"github.com/pkg/errors"
+	"golang.org/x/mod/semver"
 
 	agentconfig "github.com/hashicorp/consul/agent/config"
 	"github.com/hashicorp/consul/test/integration/consul-container/libs/utils"
@@ -26,6 +27,7 @@ type BuildContext struct {
 	injectAutoEncryption   bool // initialize the built-in CA and set up agents to use auto-encrpt
 	injectCerts            bool // initializes the built-in CA and distributes client certificates to agents
 	injectGossipEncryption bool // setup the agents to use a gossip encryption key
+	consulVersion          string
 }
 
 // BuildOptions define the desired automated test setup overrides that are
@@ -35,6 +37,7 @@ type BuildOptions struct {
 	InjectCerts            bool   // Provides a CA for all agents and (future) agent certs.
 	InjectAutoEncryption   bool   // Configures auto-encrypt for TLS and sets up certs. Overrides InjectCerts.
 	InjectGossipEncryption bool   // Provides a gossip encryption key for all agents.
+	ConsulVersion          string
 }
 
 func NewBuildContext(opts BuildOptions) (*BuildContext, error) {
@@ -43,6 +46,7 @@ func NewBuildContext(opts BuildOptions) (*BuildContext, error) {
 		injectAutoEncryption:   opts.InjectAutoEncryption,
 		injectCerts:            opts.InjectCerts,
 		injectGossipEncryption: opts.InjectGossipEncryption,
+		consulVersion:          opts.ConsulVersion,
 	}
 
 	if opts.InjectGossipEncryption {
@@ -105,10 +109,14 @@ func NewConfigBuilder(ctx *BuildContext) *Builder {
 		HTTP:    nil,
 		HTTPS:   utils.IntToPointer(8501),
 		GRPC:    utils.IntToPointer(8502),
-		GRPCTLS: utils.IntToPointer(8503),
 		SerfLAN: utils.IntToPointer(8301),
 		SerfWAN: utils.IntToPointer(8302),
 		Server:  utils.IntToPointer(8300),
+	}
+
+	if ctx != nil && semver.Compare("v"+ctx.consulVersion, "v1.14.0") >= 0 {
+		// Enable GRPCTLS for version after v1.14.0
+		b.conf.Ports.GRPCTLS = utils.IntToPointer(8503)
 	}
 
 	return b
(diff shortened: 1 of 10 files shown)


test/integration/consul-container/libs/cluster/cluster.go
```diff
@@ -121,6 +123,60 @@ func (c *Cluster) Remove(n libagent.Agent) error {
...
+	}
+
+	if snapshotCount != 1 {
+		return fmt.Errorf("only leader agent can have a snapshot file, got %d", snapshotCount)
+	}
+
+	// Upgrade individual agent to the target version
+	client := c.Agents[0].GetClient()
+	for _, agent := range c.Agents {
+		agent.Terminate()
```
huikang, line 160: Do we need to run `agent.Exec([]string{"consul", "leave"})` before terminate the container?

test/integration/consul-container/libs/cluster/cluster.go
```diff
@@ -121,6 +123,60 @@ func (c *Cluster) Remove(n libagent.Agent) error {
...
+		files, err := ioutil.ReadDir(filepath.Join(agent.DataDir(), "raft", "snapshots"))
+		if err != nil {
+			return err
+		}
+		if len(files) >= 1 {
+			snapshotCount++
+		}
+	}
+
+	if snapshotCount != 1 {
```
NiniOak, line 153: Can a leader have more than 1 snapshot file? I assumed `snapshotCount` is the total number of snapshots found in the directory

test/integration/consul-container/test/upgrade/fullstopupgrade_test.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	for _, tc := range tcs {
+		t.Run(fmt.Sprintf("upgrade from %s to %s", tc.oldversion, tc.targetVersion),
+			func(t *testing.T) {
+				run(t, tc)
+			})
+		time.Sleep(5 * time.Second)
+	}
+}
```
NiniOak, line 120: Where do we assert that the upgrade was successful by checking the new version?

test/integration/consul-container/test/upgrade/healthcheck_test.go
```diff
@@ -256,7 +256,15 @@ func clientsCreate(t *testing.T, numClients int, image string, version string, c
...
 
 func serviceCreate(t *testing.T, client *api.Client, serviceName string) uint64 {
-	err := client.Agent().ServiceRegister(&api.AgentServiceRegistration{Name: serviceName, Port: 9999})
+	err := client.Agent().ServiceRegister(&api.AgentServiceRegistration{
+		Name: serviceName,
+		Port: 9999,
+		Connect: &api.AgentServiceConnect{
+			SidecarService: &api.AgentServiceRegistration{
+				Port: 22005,
+			},
```
DanStough, line 265: No objection to adding this but is there something we wanted to validate by adding it?

test/integration/consul-container/libs/cluster/cluster.go
DanStough: Honestly, I'm not 100% sure here, but I know this will de-register services associated with agents. You might end up with an empty catalog. 

Maybe a more realistic scenario is that we should use `leave` and agents should have some service registrations they bootstrap in their config?

test/integration/consul-container/test/upgrade/fullstopupgrade_test.go
DanStough: Similarly, do we want to check on a delta with the snapshot or verify that the service we created is still registered?

test/integration/consul-container/libs/agent/builder.go
```diff
@@ -35,6 +37,7 @@ type BuildOptions struct {
 	InjectCerts            bool   // Provides a CA for all agents and (future) agent certs.
 	InjectAutoEncryption   bool   // Configures auto-encrypt for TLS and sets up certs. Overrides InjectCerts.
 	InjectGossipEncryption bool   // Provides a gossip encryption key for all agents.
+	ConsulVersion          string
```
DanStough, line 40:
```suggestion
	ConsulVersion          string // The default Consul version for agents in the cluster when none is specified.
```

DanStough (approved):
Super cool to see you building on this! Nothing blocking me on this but some comments on potential improvements.

test/integration/consul-container/libs/cluster/cluster.go
huikang: Yes, consul will take snapshot periodically or on the snapshot command. But since the test takes snapshot in a short amount of time after agent is started, there must be only one snapshot file.

test/integration/consul-container/test/upgrade/healthcheck_test.go
huikang: @DanStough , thanks for your comments. This is added to break the upgrade from 1.12.x to 1.13.0. In the future, we plan to add more struct to the snapshot to avoid any upgrade failure.

huikang: link to the reported bug: https://github.com/hashicorp/consul/issues/14107

test/integration/consul-container/test/upgrade/fullstopupgrade_test.go
huikang: @NiniOak , @DanStough , thanks for your suggestions. I will add the following 2 checks:

1. Verify the version to avoid the situation where we package the wrong binary in the docker image
2. Verify the service is restored from the snapshot

PR #15610, opened 2022-11-29
Remove log line about server mgmt token init

freddygv:
### Description
Currently the server management token is only being bootstrapped in the primary datacenter. That means that servers on the secondary datacenter will never have this token available, and would log this line any time a token is resolved.

Bootstrapping the token in secondary datacenters will be done in a follow-up.


DanStough (approved):
LGTM, but you don't think we should do a bugfix changelog?

PR #15614, opened 2022-11-30
feat(ingress-gateway): support outlier detection of upstream service for ingress gateway

### LINKED ISSUE #15401: Add support to set circuit breaker config and outlier detection config for Ingress Gateway
#### Feature Description

I would like to set outlier detection and circuit breaker config in the Ingress Gateway. By default ingress gateway is using the default envoy configuration (because it's not provided explicitly). 

#### Use Case(s)

Ingress gateway is used to accept requests from the outside of service mesh and forwarding the requests to the services inside service mesh.
Consider the following scenario:
--> ingress gateway --> cart-service

Currently after 5 consecutive 5xx (default) the cart-service is ejected by the ingress gateway. It would be useful to provide an option to change this default configuration.

I couldn't find an option to do that.. the escape hatch override options is also not availabel for ingress gateway.. there is an issue about it. (https://github.com/hashicorp/consul/issues/8722)

Is there any other to set these configurations?

huikang:
### Description
Support outlier detection of upstream service for ingress gateway.
Consul has supported adding outlier detection to envoy proxy for the [connect-proxy services](https://developer.hashicorp.com/consul/docs/connect/proxies/envoy#passive_health_check). This PR adds the similar function to ingress gateway proxy.

Example:
```
Kind = "ingress-gateway"
Name = "ingress-gateway"
Defaults {
  PassiveHealthCheck {
    max_failures = 10   // defaults for all upstream services
  }
Listeners = [
 {
   Port = 18080
   Services = [
     {
        Name = "backend"
        PassiveHealthCheck {
          max_failures = 20   // override the defaults
        }
     },
   ]
 },
}
```

generated envoy config

```
      "outlier_detection": {
       "consecutive_5xx": 20,
      },
```

- An example is added to the integration test

### Links

Fix #15401 

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


website/content/docs/connect/config-entries/ingress-gateway.mdx
```diff
@@ -1076,6 +1076,32 @@ You can specify the following parameters to configure ingress gateway configurat
...
+          name: 'PassiveHealthCheck',
+          type: 'PassiveHealthCheck: <optional>',
+          description:
+            'Passive health checks are used to remove hosts from the upstream cluster which are unreachable or are returning errors.',
+          children: [
+            {
+              name: 'interval',
+              type: 'int: <optional>',
+              description:
+                "The time in nanosecond between checks. Each check will cause hosts which have exceeded `max_failures` to be removed from the load balancer, and any hosts which have passed their ejection time to be returned to the load balancer. If not specified, it uses the default value. For example, 10s for Envoy proxy.",
```
trujillo-adam, line 1089:
```suggestion
                "The time in nanosecond between checks. If the number of failed checks exceed the `max_failures`, Consul removes the failing host from the load balancer. Additionally, hosts that have passed their ejection time are returned to the load balancer. If not specified, Consul uses the proxy's default value. For example, `10s` for Envoy proxy.",
```
Is that clearer without changing the meaning? Also, what is the "ejection time"? I didn't see that phrase anywhere else on the page.

website/content/docs/connect/config-entries/ingress-gateway.mdx
```diff
@@ -1076,6 +1076,32 @@ You can specify the following parameters to configure ingress gateway configurat
...
+            },
+            {
+              name: 'max_failures',
+              type: 'int: <optional>',
+              description:
+                'The number of consecutive failures which cause a host to be removed from the load balancer. If not specified, it uses the default value. For example, 5 for Envoy proxy.',
```
trujillo-adam, line 1095:
```suggestion
                'The number of consecutive failures that cause a host to be removed from the load balancer. If not specified, Consul uses the proxy's default value. For example, `5` for Envoy proxy.',
```
Just FYI, use "that" to add essential information and "which" preceded by a comma to add a non-essential phrase. Also, I replaced "it" to make it clearer that Consul is the thing doing the action. If that is incorrect, we should be clear about _what_ "uses the proxy's default value".

website/content/docs/connect/config-entries/ingress-gateway.mdx
```diff
@@ -1076,6 +1076,32 @@ You can specify the following parameters to configure ingress gateway configurat
...
+            },
+            {
+              name: 'enforcing_consecutive_5xx',
+              type: 'int: <optional>',
+              description:
+                'The chance in percentage that a host will be actually ejected when an outlier status is detected through consecutive 5xx. If not specified, it uses the default value. For example, 100 for Envoy proxy.',
```
trujillo-adam, line 1101:
```suggestion
                'A percentage representing how often a host should be ejected when Consul detects an outlier status through consecutive errors in the 500 code range. If not specified, Consul uses the proxy's default value. For example, `100` for Envoy proxy.',
```
The target audience probably understands "5xx" to mean error codes from 500 - 599, but it might be confusing for beginners (like me :|). If that suggestion is too wordy, I still think something like "500 error" is better than "5xx". Also, I think we should be more concrete about describing what the percentage represents. Describing the value in terms of "chance" might confuse people.

trujillo-adam (commented):
Added some suggestions for the docs

website/content/docs/connect/config-entries/ingress-gateway.mdx
huikang: @trujillo-adam , thanks for helping editing the text. Actually I copied these sentences from [other places](https://developer.hashicorp.com/consul/docs/connect/proxies/envoy#passive_health_check) where passive health check is enabled. I like your edited version, so will overwrite the other place in the following commit.

eculver (approved):
Looks good! We should probably let someone from the core team chime in too just for good measure, but all good from my perspective.

Since this is technically a new feature, do we want to backport this?

agent/xds/clusters.go
```diff
@@ -849,6 +849,31 @@ func (s *ResourceGenerator) configIngressUpstreamCluster(c *envoy_cluster_v3.Clu
...
+	// override the default outlier detection value
+	if svc.PassiveHealthCheck != nil {
+		if svc.PassiveHealthCheck.Interval != 0 {
+			outlierDetection.Interval = durationpb.New(svc.PassiveHealthCheck.Interval)
+		}
+		if svc.PassiveHealthCheck.MaxFailures != 0 {
+			outlierDetection.Consecutive_5Xx = &wrappers.UInt32Value{Value: svc.PassiveHealthCheck.MaxFailures}
+		}
+
+		// if svc.PassiveHealthCheck.EnforcingConsecutive5xx != nil && *svc.PassiveHealthCheck.EnforcingConsecutive5xx != 0 {
```
erichaberkorn, line 865: Can we delete this?

agent/xds/clusters.go
```diff
@@ -849,6 +849,31 @@ func (s *ResourceGenerator) configIngressUpstreamCluster(c *envoy_cluster_v3.Clu
 	if threshold != nil {
 		c.CircuitBreakers.Thresholds = []*envoy_cluster_v3.CircuitBreakers_Thresholds{threshold}
 	}
+
+	// Configure the default outlier detector for upstream service
+	outlierDetection := ToOutlierDetection(cfgSnap.IngressGateway.Defaults.PassiveHealthCheck)
+
+	// override the default outlier detection value
+	if svc.PassiveHealthCheck != nil {
```
erichaberkorn, line 857: I wonder if it would simplify things to either update `ToOutlierDetection` or make a new function that takes a variable number of `*structs.PassiveHealthCheck`s and `peeredService`. If we do that, I believe we can share this logic everywhere in this file.

agent/xds/clusters.go
```diff
@@ -849,6 +849,31 @@ func (s *ResourceGenerator) configIngressUpstreamCluster(c *envoy_cluster_v3.Clu
...
+			outlierDetection.EnforcingConsecutive_5Xx = &wrappers.UInt32Value{Value: *svc.PassiveHealthCheck.EnforcingConsecutive5xx}
+		}
+	}
+
+	// Specail handling for failover peering service, which has set MaxEjectionPercent
+	if c.OutlierDetection != nil && c.OutlierDetection.MaxEjectionPercent != nil {
+		outlierDetection.MaxEjectionPercent = &wrappers.UInt32Value{Value: c.OutlierDetection.MaxEjectionPercent.Value}
+	}
+
+	c.OutlierDetection = outlierDetection
```
erichaberkorn, line 876: I think always setting outlier detection will end up being a breaking change and is causing golden tests like [this one](agent/xds/testdata/clusters/ingress-with-chain-and-failover-to-cluster-peer.latest.golden) to change. Is that intended?

If so, we should make this a breaking change in the changelog and we can't backport it.

agent/xds/clusters_test.go
```diff
@@ -540,12 +550,22 @@ func TestClustersFromSnapshot(t *testing.T) {
...
 							MaxPendingRequests: 512,
+							PassiveHealthCheck: &structs.PassiveHealthCheck{
+								Interval:                5000000000,
+								EnforcingConsecutive5xx: &defaultEnforcingConsecutive5xx,
+							},
 						}
+						enforcingConsecutive5xx := uint32(50)
 						entry.Listeners[0].Services[0].MaxConnections = 4096
 						entry.Listeners[0].Services[0].MaxPendingRequests = 2048
+						entry.Listeners[0].Services[0].PassiveHealthCheck = &structs.PassiveHealthCheck{
```
erichaberkorn, line 565: It might be safer to add the tests for outlier detection to new cases so it's easier to ensure the existing cases don't change.

agent/xds/clusters.go
huikang: Done. Add an `override` parameter to `ToOutlierDetection`.

agent/xds/clusters.go
huikang: After second thought, I think adding empty outlier_detection make it consistent with other places, like connect-proxy, mesh-gateway

https://github.com/hashicorp/consul/blob/4a32070210cdbff8c457334ee2e3e798b9f263b1/agent/xds/testdata/clusters/connect-proxy-with-tls-outgoing-cipher-suites.latest.golden#L21-L24

https://github.com/hashicorp/consul/blob/4a32070210cdbff8c457334ee2e3e798b9f263b1/agent/xds/testdata/clusters/mesh-gateway-with-exported-peered-services-http-with-router.latest.golden#L17-L19

So I added a note of this breaking change to the changelog.

agent/xds/clusters.go
huikang: Done

agent/xds/clusters.go
erichaberkorn: Since this is going to be a breaking change, could you remove the backport labels from this PR?

huikang: Sure, backport labels are removed.

proto/pbconfigentry/config_entry.proto
```diff
@@ -544,9 +546,9 @@ message UpstreamLimits {
 message PassiveHealthCheck {
   // mog: func-to=structs.DurationFromProto func-from=structs.DurationToProto
   google.protobuf.Duration Interval = 1;
-  uint32 MaxFailures = 2;
+  optional uint32 MaxFailures = 2;
```
erichaberkorn, line 549: Why did these become `optional`? https://github.com/hashicorp/consul/pull/15614/files#diff-9d2e9c8b34664ef85c83b6c6f4ec7f939cce7a64f41c267e5db248c1f95b0571R549-R551

Do you know if this is a breaking change?

huikang: IMO, this is not a breaking change.

Optional makes `mog` know if `MaxFailures` is defined or not. Otherwise, unspecified `MaxFailures` will become "0", which is a wrong value. In the PR, I added a test for unspecified `EnforcingConsecutive5xx`, which is the same type as `MaxFailures`, to verify the default value is "null" for envoy.

https://github.com/hashicorp/consul/pull/15614/files#diff-a44e2ceb5ced90fcc9be4d9518bb62626d62438751d9cc67286de9df15d15b54R72

huikang: @erichaberkorn , there is a test that asserts the `maxFailure` in upstream config

https://github.com/hashicorp/consul/blob/4a32070210cdbff8c457334ee2e3e798b9f263b1/test/integration/connect/envoy/case-upstream-config/verify.bats#L49-L52

huikang: Another test is added to verify unspecified `MaxFailure` in `upstreams.config.passive_health_check`

https://github.com/hashicorp/consul/pull/15614/commits/30af125d96bb7b6665659c428b7f62b59df193df

erichaberkorn: I'm talking about a protobuf compatibility breaking change. I'm pretty sure we would need to introduce a completely different field on the protobuf and deprecate the current field to get this working. Then we could remove the current field in a follow-up release.

huikang: @erichaberkorn , reverted the change to `message PassiveHealthCheck` by adding the constraint of `EnforcingConsecutive5xx > 0`. WDYT?

agent/xds/clusters.go
```diff
@@ -849,6 +849,20 @@ func (s *ResourceGenerator) configIngressUpstreamCluster(c *envoy_cluster_v3.Clu
...
+
+	// Configure the outlier detector for upstream service
+	var override *structs.PassiveHealthCheck
+	if svc != nil {
+		override = svc.PassiveHealthCheck
+	}
+	outlierDetection := ToOutlierDetection(cfgSnap.IngressGateway.Defaults.PassiveHealthCheck, override)
+
+	// Specail handling for failover peering service, which has set MaxEjectionPercent
+	if c.OutlierDetection != nil && c.OutlierDetection.MaxEjectionPercent != nil {
```
erichaberkorn, line 861: Sorry for yet another comment. I just noticed that we aren't setting panic threshold here [similar to other clusters](https://github.com/hashicorp/consul/blob/main/agent/xds/clusters.go#L264-L266). For the sake of consistency, I'm guessing we want to set it like we do on other cluster types (because panic threshold interacts with outlier detection).

agent/xds/config.go
```diff
@@ -175,24 +175,40 @@ func ParseGatewayConfig(m map[string]interface{}) (GatewayConfig, error) {
...
+
+		if p.Interval != 0 {
+			od.Interval = durationpb.New(p.Interval)
+		}
+		if p.MaxFailures != 0 {
+			od.Consecutive_5Xx = &wrappers.UInt32Value{Value: p.MaxFailures}
+		}
+
+		// NOTE: EnforcingConsecutive5xx must be great than 0
+		if p.EnforcingConsecutive5xx != nil && *p.EnforcingConsecutive5xx > 0 {
```
erichaberkorn, line 194: I believe this ends up getting rid of the ability for users to set `EnforcingConsecutive5xx` to zero which is a breaking change for other proxy types.

agent/xds/clusters.go
huikang: Sorry, I am little confused since this has been set for the upstream clusters of the ingress gateway in https://github.com/hashicorp/consul/blob/cd53120cd7496c21892494c27f234ef1d3623081/agent/xds/clusters.go#L1147, which is called in making upstream clusters for the ingress-gateway service https://github.com/hashicorp/consul/blob/cd53120cd7496c21892494c27f234ef1d3623081/agent/xds/clusters.go#L789-L796

agent/xds/config.go
huikang: Added some logic to handle the legacy scenarios. Please take another look. Thanks.

agent/xds/config.go
```diff
@@ -175,24 +175,53 @@ func ParseGatewayConfig(m map[string]interface{}) (GatewayConfig, error) {
 	return cfg, err
 }
 
-// Return an envoy.OutlierDetection populated by the values from this struct.
+// Return an envoy.OutlierDetection populated by the values from structs.PassiveHealthChec.
 // If all values are zero a default empty OutlierDetection will be returned to
 // enable outlier detection with default values.
-func ToOutlierDetection(p *structs.PassiveHealthCheck) *envoy_cluster_v3.OutlierDetection {
+//   - If override is not nil, it will overwrite the values from p, e.g., ingress gateway defaults
+//   - allowZero is added to handle the legacy case where connec-proxy and mesh gateway can set 0
```
erichaberkorn, line 182:
```suggestion
//   - allowZero is added to handle the legacy case where connect-proxy and mesh gateway can set 0
```

agent/xds/config.go
```diff
@@ -175,24 +175,53 @@ func ParseGatewayConfig(m map[string]interface{}) (GatewayConfig, error) {
...
+//     for EnforcingConsecutive5xx. Due to the definitino of proto of PassiveHealthCheck, ingress
```
erichaberkorn, line 183:
```suggestion
//     for EnforcingConsecutive5xx. Due to the definition of proto of PassiveHealthCheck, ingress
```

erichaberkorn (approved):

website/content/docs/connect/config-entries/ingress-gateway.mdx
```diff
@@ -1076,6 +1076,32 @@ You can specify the following parameters to configure ingress gateway configurat
...
               If not specified, it uses the default value. For example, 1024 for Envoy proxy.`,
         },
+        {
+          name: 'PassiveHealthCheck',
+          type: 'PassiveHealthCheck: <optional>',
+          description:
+            'Passive health checks remove hosts from the upstream cluster that are unreachable or that return errors.',
+          children: [
+            {
+              name: 'interval',
```
shweshi, line 1086: This should be Interval and not interval.

website/content/docs/connect/config-entries/ingress-gateway.mdx
```diff
@@ -1076,6 +1076,32 @@ You can specify the following parameters to configure ingress gateway configurat
...
+              type: 'int: <optional>',
+              description:
+                "The time in nanosecond between checks. Each check will cause hosts which have exceeded `max_failures` to be removed from the load balancer, and any hosts which have passed their ejection time to be returned to the load balancer. If not specified, it uses the default value. For example, 10s for Envoy proxy.",
+            },
+            {
+              name: 'max_failures',
```
shweshi, line 1092: The name should be MaxFailures

website/content/docs/connect/config-entries/ingress-gateway.mdx
```diff
@@ -1076,6 +1076,32 @@ You can specify the following parameters to configure ingress gateway configurat
...
+              type: 'int: <optional>',
+              description:
+                'The number of consecutive failures that cause a host to be removed from the upstream cluster. If not specified, Consul uses the proxy\'s default value. For example, `5` for Envoy proxy.',
+            },
+            {
+              name: 'enforcing_consecutive_5xx',
```
shweshi, line 1098: @huikang The name field should be EnforcingConsecutive5xx. Same for the Line 1251.

PR #15615, opened 2022-11-30
peering: better represent non-passing states during peer check flattening

rboyer:
### Description

During peer stream replication we flatten checks from the source cluster and build one thin overall check to hide the irrelevant details from the consuming cluster. This flattening logic did correctly flip to non-passing if there were any non-passing checks, but WHICH status it got during that was random (warn/error).

Also it didn't represent "maintenance" operations. There is an `api` package call [`AggregatedStatus`](https://github.com/hashicorp/consul/blob/api/v1.17.0/api/health.go#L183) which more correctly flattened check statuses.

This PR replicated the more complete logic into the peer stream package.

### DIFF
--- a/agent/grpc-external/services/peerstream/subscription_manager.go
+++ b/agent/grpc-external/services/peerstream/subscription_manager.go
@@ -664,6 +664,18 @@ func createDiscoChainHealth(
 	}
 }
 
+var statusScores = map[string]int{
+	// 0 is reserved for unknown
+	api.HealthMaint:    1,
+	api.HealthCritical: 2,
+	api.HealthWarning:  3,
+	api.HealthPassing:  4,
+}
+
+func isStatusBetter(curr, next string) bool {
+	return statusScores[next] < statusScores[curr]
+}
+
 func flattenChecks(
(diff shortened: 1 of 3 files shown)


agent/grpc-external/services/peerstream/subscription_manager.go
```diff
@@ -664,6 +664,18 @@ func createDiscoChainHealth(
...
 
+var statusScores = map[string]int{
+	// 0 is reserved for unknown
+	api.HealthMaint:    1,
+	api.HealthCritical: 2,
+	api.HealthWarning:  3,
+	api.HealthPassing:  4,
+}
+
+func isStatusBetter(curr, next string) bool {
```
hashi-derek, line 675: Nit. To simplify mental-effort, I would have this function return the winning string instead of having it return a bool. Also, the name "better" makes me think that it's returning the healthier status, which is a little confusing. I'd probably name it `getMostImportantStatus()` or something similar.

hashi-derek (approved):

hashi-derek (approved):

PR #14854, opened 2022-10-04
Rename `PeerName` to `Peer` on prepared queries and exported services

erichaberkorn:
### PR Checklist

* [X] updated test coverage
* [ ] external facing docs updated
* [X] not a security concern

PR #15616, opened 2022-11-30
[OSS] refactor(peering) make peer name uniform

### LINKED ISSUE #15509: Make Peering Name API Uniform
#### Feature Description

The use of a peer names is not consistent in the API. For config entries, we updated this field to be `Peer` in exported-services and prepared queries (#14854), but the [HTTP API](https://developer.hashicorp.com/consul/api-docs/peering#peername) still uses `PeerName`.  It would be ideal if these were uniform. 

Because this field is released, we should deprecate `PeerName` and add support for `Peer`.

#### Acceptance Criteria
1. HTTP API is updated by changing the protocol buffer messages ([example](https://github.com/hashicorp/consul/blob/22d216c1069e7823798034a30c64abe5b13700f4/proto/pbpeering/peering.proto#L365-L365)).
2. `PeerName` usages are updated to also use `Peer`
1. Docs are updated
  1. 1.14 API field is marked deprecated
  3. 1.XX API field is added.

DanStough:
### Description
Closes #15509 

In some places in the public API we are still using "PeerName" instead of "Peer", which we modified in other places as part of #14854.

### PR Checklist

* [X] updated test coverage
* [ ] not a security concern


DanStough:
Closing due to lack of interest

PR #15631, opened 2022-11-30
Onboard consul to use new version/VERSION file and reproducible actions

claire-labry:
# PR Description
This PR introduces the new [actions-set-product-version](https://github.com/hashicorp/actions-set-product-version), a tiny, but mighty, github action that acts as a bridge between the product repo and our new CRT feature: [automated version bumping](https://github.com/hashicorp/bob/commit/6813d9757c644679193a0af317e99570ac8cc848). This PR also introduces automatic version patch bumping via the bump-version {todo: enter link when merged} workflow.

tl;dr automated version bumping has a new command (bob update version) in the [bob CLI](https://github.com/hashicorp/bob) that automatically bumps the version to a new patch. This automation has been introduced to crt-workflows-common as a [new workflow](https://github.com/hashicorp/crt-workflows-common/pull/233) (with the new bob command)  and handles version bumping at the end of the release pipeline (after being released to production); for example, `1.0.0→1.0.1` and `1.0.0-dev→1.0.0`. Bumping the minor version (ie `1.0.x→1.1.0`) is only supported manually via `bob update version -bump minor`, but not supported in CRT (this work is upcoming).  This is made possible by adding the new event “bump-version-patch” in the ci.hcl file in this PR. 

What this small action does: 

-  Allows for the static version string from the `.release/VERSION` file to be read by the new CRT workflow and automagically be bumped to the next version (whether it be a minor, or patch, or major version bump). 
- Outputs an error if there’s no `VERSION` file in the `.release/ dir`
- Outputs an error if there’s no version string in the VERSION file
- Is able to parse `product_version` if it is `1.3.0-alpha1` as `1.3.0` (example: when `product_version = 1.3.0-alpha1`, `base_version = 1.3.0`)
- Is able to parse prerelease product versions such as `alpha1` (example `prerelease_product_version = alpha1`) in the statement above.  

In other words, the action grabs the version from the version file at the beginning of the build, then passes those versions (along with metadata, where necessary) to the new version bump automation CRT workflow at the end of the release pipeline to bump the product’s version. One thing to note is that the `version.go` and `base_version.go` files are used for local development and will be now overwritten by the ld-flags with the actual version that are being called in the go builds. 

This PR also introduces [actions-go-build](https://github.com/hashicorp/actions-go-build), an action that dynamically takes care of building and packaging go binaries. Through this action, you can define the go build, assert that it’s reproducible (optional*) and then use the artifacts in the workflow. (Once the build action run is finished, you can check out it the builds are reproducible by scrolling down on the summary page, [example](https://github.com/hashicorp/packer/actions/runs/3587803756#summary-9808859635)) 

*NOTE:  if [assert is enabled](https://github.com/hashicorp/actions-go-build#reproducibility-assertions), this could potentially create some disruptions in workflows as the action tries to verify two builds (primary build and ui builds) are the same (size, shasums) as ui builds might have non-deterministic compilation.

Passing tests ✔️  Check out these examples:

- https://github.com/hashicorp/consul-enterprise/actions/runs/3623272718

This PR also introduces a bump to the version tags for `actions/checkout`, `actions/setup-go`, and `actions/upload-artifact` (setup-go version tag update was required by the introduction of actions-go-build)-- however, if the Packer team feels like these should be isolated in their own PRs for easier/broader visibility, super happy to raise separate PRs! 

Finally, as a part of this effort to onboard new CRT tooling, we have implemented the new `prepare` workflow in the `ci.hcl` config file. This workflow simply wraps all workflows in the pre-stable channel in one workflow, to make your ci.hcl config file more readable for your team. The `prepare` runs `upload-dev`, `security-scan`, `notarize`, `sign`, `verify`, `promote-dev-docker`, and `fossa-scan` (depending on your team’s current needs) under the hood - no need to worry about these workflows anymore!

# TODO Before Merging
 - [ ] Update version/VERSION to latest version
 - [ ] Remove `set-product-version` branch from `build.yml` workflow dispatch list
 - [ ] Remove `set-product-version` branch from `ci.hcl` allowlist


.release/VERSION
```diff
@@ -0,0 +1 @@
+0.0.0
```
claire-labry, line 1: TODO: Update to match current version of Consul when PR is ready for full review.

.github/workflows/build.yml
```diff
@@ -7,46 +7,60 @@ on:
...
       - name: Set shared -ldflags
         id: shared-ldflags
         run: |
           T="github.com/hashicorp/consul/version"
-          echo "::set-output name=shared-ldflags::-X ${T}.GitCommit=${GITHUB_SHA::8} -X ${T}.GitDescribe=${{ steps.get-product-version.outputs.product-version }} -X ${T}.BuildDate=${{ steps.get-product-version.outputs.product-date }}"
-
+          echo "::set-output name=shared-ldflags::-X ${T}.GitCommit=${GITHUB_SHA::8} \
+          -X ${T}.GitDescribe=${{ steps.set-product-version.outputs.product-version }} \
+          -X ${T}.BuildDate=${{ steps.get-product-version.outputs.product-date }} \
+          -X ${T}.Version=${{ steps.set-product-version.outputs.base-product-version }} \
```
sarahethompson, line 46: we can drop setting the Version, VersionPrerelease, VersionMetadata here I think since we're using go:embed - like we did in the consul-enterprise PR -

emilymianeil:
### Manual Test for Prepare Workflow: PASSED ✅ 
We need to verify the most critical parts of the prepare workflow: notarization, codesigning (running in circle-codesign CI, and verified through verifying correct notarization jobs were run), and checksums file.
1. Verify prepare workflow has passed:
      - [ ] Prepare: [passed](https://github.com/hashicorp/crt-workflows-common/actions/runs/3643246148/jobs/6151458608) 
2. Verify correct notarization has run + passed:
      - [ ] Darwin AMD64: [passed](https://github.com/hashicorp/crt-workflows-common/actions/runs/3643246148/jobs/6151460741) 
      - [ ] Darwin ARM64: [passed](https://github.com/hashicorp/crt-workflows-common/actions/runs/3643246148/jobs/6151460518)
      - [ ] Windows 386: [passed](https://github.com/hashicorp/crt-workflows-common/actions/runs/3643246148/jobs/6151458847) 
      - [ ] Windows AMD64: [passed](https://github.com/hashicorp/crt-workflows-common/actions/runs/3643246148/jobs/6151458608) 
3. Verify the checksums file to ensure all binaries are present
      - [ ] Download SHASUMS from stable bucket: [passed](https://artifactory.hashicorp.engineering/artifactory/hashicorp-crt-stable-local/consul/1.15.0-dev/26f9008808cd281129eb1b2feb367cb0f767ce20/consul_1.15.0-dev_SHA256SUMS)

.github/workflows/build.yml
```diff
@@ -7,57 +7,68 @@ on:
       # Push events on the main branch
       - main
       - release/**
+      - set-product-version
```
DanStough, line 10: I'm guessing this can be removed now that dev is done.

.github/workflows/build.yml
```diff
@@ -7,57 +7,68 @@ on:
...
 
 env:
   PKG_NAME: consul
+  VERSION: oss
```
DanStough, line 14: This is `METADATA` in ENT.  IMHO I like that terminology better since it's competing with a lot of other "versions" here.

.release/ci.hcl
```diff
@@ -12,213 +12,33 @@ project "consul" {
     release_branches = [
       "main",
       "release/**",
+      "set-product-version",
```
DanStough, line 15: Same comment here about dev artifact.

DanStough (commented):
Awesome, thank you! 🙇. Just a few minor things. Holding off approving until after next week's release.

DanStough (approved):
🚀

PR #15632, opened 2022-11-30
Add warn log when all ACL policies are filtered out

kisunji:
### Description
UX improvement for operators to help identify when they use a token meant for another datacenter.

### DIFF
--- a/agent/consul/acl.go
+++ b/agent/consul/acl.go
@@ -633,7 +633,7 @@ func (r *ACLResolver) resolvePoliciesForIdentity(identity structs.ACLIdentity) (
 	policies = append(policies, syntheticPolicies...)
 	filtered := r.filterPoliciesByScope(policies)
 	if len(policies) > 0 && len(filtered) == 0 {
-		r.logger.Warn("all ACL policies have been filtered out based on datacenter", "accessor_id", identity.ID(), "datacenter", r.config.Datacenter)
+		r.logger.Warn("ACL token used lacks permissions in this datacenter: its associated ACL policies, service identities, and/or node identities are scoped to other datacenters", "accessor_id", identity.ID(), "datacenter", r.config.Datacenter)
 	}
 
 	return filtered, nil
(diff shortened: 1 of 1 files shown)


agent/consul/acl.go
```diff
@@ -632,6 +632,10 @@ func (r *ACLResolver) resolvePoliciesForIdentity(identity structs.ACLIdentity) (
 
 	policies = append(policies, syntheticPolicies...)
 	filtered := r.filterPoliciesByScope(policies)
+	if len(policies) > 0 && len(filtered) == 0 {
+		r.logger.Warn("all ACL policies have been filtered out based on datacenter", "accessor_id", identity.ID(), "datacenter", r.config.Datacenter)
```
jkirschner-hashicorp, line 636: I feel like there's some context missing from the message that might make it more understandable.

Perhaps something like:
```suggestion
		r.logger.Warn("Potential ACL token configuration issue: this token has no permissions in this datacenter because its associated ACL policies, service identities, and/or node identities are scoped to other datacenters", "accessor_id", identity.ID(), "datacenter", r.config.Datacenter)
```

jkirschner-hashicorp (commented):
Small suggestion on the message copy. We can workshop it back and forth a bit if needed.

It's nice when a code change is really small ;)

agent/consul/acl.go
cizh: Love the context! I wonder if we can also make the message concise to save user's reading effort, bear with me for I don't have the technical proficiency, here's my attempt: "`token has no permission in this datacenter: check the scope of its associated ACL policies, service identities and/or node identities`". or straight forward " `token meant for another datacenter: check the scope of its associated ACL policies, service identities and/or node identities`"

So we "[directly describe what happened]: [give an actionable opinionated way out]". 
**QQ:** Do we suggest user to check where this token belongs to(what we have above), or suggest them to find the right token in this case scenario? 
Happy to workshop it back-and-forth a bit.+1

kisunji: Updated wording from slack discussion

PR #15638, opened 2022-12-01
clean up go.mod

kisunji:
Removes unused `replace` directive, groups the consul sub-modules into one block, and adds an `exclude` directive for msgpack

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,8 +8,10 @@ replace (
 	github.com/hashicorp/consul/sdk => ./sdk
 )
 
-// go-msgpack v1.1.5 has breaking changes and must be avoided
-exclude github.com/hashicorp/go-msgpack v1.1.5
+exclude (
+	github.com/hashicorp/go-msgpack v1.1.5 // has breaking changes and must be avoided
+	github.com/hashicorp/go-msgpack v1.1.6 // contains retractions but same as v1.1.5
+)
 
 require (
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -2,13 +2,14 @@ module github.com/hashicorp/consul
...
-replace github.com/hashicorp/consul/sdk => ./sdk
-
-replace github.com/hashicorp/consul/proto-public => ./proto-public
+replace (
+	github.com/hashicorp/consul/api => ./api
+	github.com/hashicorp/consul/proto-public => ./proto-public
+	github.com/hashicorp/consul/sdk => ./sdk
+)
 
-replace launchpad.net/gocheck => github.com/go-check/check v0.0.0-20140225173054-eb6ee6f84d0a
```
kisunji, line 11: I couldn't find any reference of this (even in our enterprise repo) so it feels safe to remove

go.mod
```diff
@@ -2,13 +2,14 @@ module github.com/hashicorp/consul
...
+// go-msgpack v1.1.5 has breaking changes and must be avoided
+exclude github.com/hashicorp/go-msgpack v1.1.5
```
banks, line 12: I also added 1.1.6 here which is also retracted and broken just in case anyone ever did pin to that!

banks (approved):
I think it's worth adding 1.1.6 too, but otherwise LGTM.

PR #15643, opened 2022-12-01
use a separate go routine to write limiters

dhiaayachi:
### Description
This is an alternative to #15642 
Use a go buffered go routine to accumulate Limiters in a transaction and write them to the immutable-radix-tree every 10Ms.

 

#### Without any synchronization code in #15467 
``` bash
❯ go test -bench=BenchmarkTestRateLimiterAllowC -count 5 -benchtime=1s  -timeout 6000s -run=XXX ./agent/consul/multilimiter
goos: darwin
goarch: amd64
pkg: github.com/hashicorp/consul/agent/consul/multilimiter
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
AC/no_prefill-16         	  556688	      2060 ns/op
AC/no_prefill-16         	  672924	      1964 ns/op
AC/no_prefill-16         	  627190	      1932 ns/op
AC/no_prefill-16         	  615247	      1952 ns/op
AC/no_prefill-16         	  661696	      1907 ns/op
AC/prefill_with_1K_keys-16         	  727677	      1894 ns/op
AC/prefill_with_1K_keys-16         	  611496	      2310 ns/op
AC/prefill_with_1K_keys-16         	  642594	      2043 ns/op
AC/prefill_with_1K_keys-16         	  590719	      1888 ns/op
AC/prefill_with_1K_keys-16         	  625422	      1937 ns/op
AC/prefill_with_10K_keys-16        	  494367	      2436 ns/op
AC/prefill_with_10K_keys-16        	  522981	      2491 ns/op
AC/prefill_with_10K_keys-16        	  529140	      2291 ns/op
AC/prefill_with_10K_keys-16        	  612369	      2356 ns/op
AC/prefill_with_10K_keys-16        	 1000000	      2607 ns/op
AC/prefill_with_100K_keys-16       	 1000000	      3285 ns/op
AC/prefill_with_100K_keys-16       	 1000000	      3626 ns/op
AC/prefill_with_100K_keys-16       	 1000000	      3891 ns/op
AC/prefill_with_100K_keys-16       	 1000000	      4479 ns/op
AC/prefill_with_100K_keys-16       	 1000000	      4352 ns/op
PASS
ok  	github.com/hashicorp/consul/agent/consul/multilimiter	66.699s
```

#### This branch (channel with transaction)

```bash
❯ go test -bench=BenchmarkTestRateLimiterAllowC -count 5 -benchtime=1s  -timeout 6000s -run=XXX ./agent/consul/multilimiter
goos: darwin
goarch: amd64
pkg: github.com/hashicorp/consul/agent/consul/multilimiter
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
AC/no_prefill-16         	  465397	      3929 ns/op
AC/no_prefill-16         	  736290	      3168 ns/op
AC/no_prefill-16         	  770343	      2957 ns/op
AC/no_prefill-16         	  669290	      3759 ns/op
AC/no_prefill-16         	  823153	      3547 ns/op
AC/prefill_with_1K_keys-16         	  807282	      4839 ns/op
AC/prefill_with_1K_keys-16         	  718530	      2788 ns/op
AC/prefill_with_1K_keys-16         	  786652	      4330 ns/op
AC/prefill_with_1K_keys-16         	  465585	      2994 ns/op
AC/prefill_with_1K_keys-16         	  715992	      2619 ns/op
AC/prefill_with_10K_keys-16        	  699630	      2876 ns/op
AC/prefill_with_10K_keys-16        	  703555	      3843 ns/op
AC/prefill_with_10K_keys-16        	  347377	      3426 ns/op
AC/prefill_with_10K_keys-16        	  659928	      3561 ns/op
AC/prefill_with_10K_keys-16        	  575187	      2929 ns/op
AC/prefill_with_100K_keys-16       	  552110	      2893 ns/op
AC/prefill_with_100K_keys-16       	  538591	      3101 ns/op
AC/prefill_with_100K_keys-16       	  542137	      2986 ns/op
AC/prefill_with_100K_keys-16       	  501328	      3057 ns/op
AC/prefill_with_100K_keys-16       	  493551	      3292 ns/op
PASS
ok  	github.com/hashicorp/consul/agent/consul/multilimiter	176.243s
```

### DIFF
--- a/agent/consul/multilimiter/multilimiter.go
+++ b/agent/consul/multilimiter/multilimiter.go
@@ -137,7 +137,7 @@ func (m *MultiLimiter) Allow(e LimitedEntity) bool {
 	limiters := m.limiters.Load()
 	l, ok := limiters.Get(e.Key())
 	now := time.Now()
-	unixNow := now.Unix()
+	unixNow := time.Now().UnixMilli()
 	if ok {
 		limiter := l.(*Limiter)
 		if limiter.limiter != nil {
@@ -162,37 +162,43 @@ func (m *MultiLimiter) Allow(e LimitedEntity) bool {
 }
 
 func (m *MultiLimiter) runStore(ctx context.Context) {
-	var txn *radix.Txn
 	commitInterval := 10 * time.Millisecond
+	limiters := m.limiters.Load()
+	txn := limiters.Txn()
 	waiter := time.NewTicker(commitInterval)
 	defer waiter.Stop()
 	for {
-		select {
-		case <-waiter.C:
-			if txn != nil {
-				tree := txn.Commit()
-				m.limiters.Store(tree)
-			}
-		case lk := <-m.LimiterCh:
-			limiters := m.limiters.Load()
-			if txn == nil {
-				txn = limiters.Txn()
-			}
-			v, ok := txn.Get(lk.k)
-			if !ok {
-				txn.Insert(lk.k, lk.l)
-			} else {
-				switch l := v.(type) {
-				case *Limiter:
-					l.lastAccess.Store(lk.t.Unix())
-					l.limiter.AllowN(lk.t, 1)
-				}
-			}
-		case <-ctx.Done():
+		if m.runStoreOnce(ctx, waiter, txn) {
 			return
 		}
 
 	}
+}
+
+func (m *MultiLimiter) runStoreOnce(ctx context.Context, waiter *time.Ticker, txn *radix.Txn) bool {
+	select {
+	case <-waiter.C:
+		if txn != nil {
+			tree := txn.Commit()
+			m.limiters.Store(tree)
+			limiters := m.limiters.Load()
+			txn = limiters.Txn()
+		}
+	case lk := <-m.LimiterCh:
+		v, ok := txn.Get(lk.k)
+		if !ok {
+			txn.Insert(lk.k, lk.l)
+		} else {
+			switch l := v.(type) {
+			case *Limiter:
+				l.lastAccess.Store(lk.t.UnixMilli())
+				l.limiter.AllowN(lk.t, 1)
+			}
+		}
+	case <-ctx.Done():
+		return true
+	}
+	return false
 }
 
 // reconcileLimitedOnce is called by the MultiLimiter clean up routine to remove old Limited entries
@@ -211,7 +217,7 @@ func (m *MultiLimiter) reconcileLimitedOnce(now time.Time, reconcileCheckLimit t
 		case *Limiter:
 			if t.limiter != nil {
 				lastAccess := t.lastAccess.Load()
-				lastAccessT := time.Unix(lastAccess, 0)
+				lastAccessT := time.UnixMilli(lastAccess)
 				diff := now.Sub(lastAccessT)
 
 				if diff > reconcileCheckLimit {
--- a/agent/consul/multilimiter/multilimiter.go
+++ b/agent/consul/multilimiter/multilimiter.go
@@ -28,12 +28,18 @@ type RateLimiter interface {
 	UpdateConfig(c LimiterConfig, prefix []byte)
 }
 
+type LimiterWithKey struct {
+	l *Limiter
+	k []byte
+}
+
 // MultiLimiter implement RateLimiter interface and represent a set of rate limiters
 // specific to different LimitedEntities and queried by a LimitedEntities.Key()
 type MultiLimiter struct {
 	limiters        *atomic.Pointer[radix.Tree]
 	limitersConfigs *atomic.Pointer[radix.Tree]
 	defaultConfig   *atomic.Pointer[Config]
+	LimiterCh       chan *LimiterWithKey
 }
 
 type KeyType = []byte
@@ -88,12 +94,15 @@ func NewMultiLimiter(c Config) *MultiLimiter {
 	if c.ReconcileCheckInterval == 0 {
 		c.ReconcileCheckLimit = 1 * time.Second
 	}
-	m := &MultiLimiter{limiters: &limiters, defaultConfig: &config, limitersConfigs: &configs}
+	chLimiter := make(chan *LimiterWithKey, 100)
+	m := &MultiLimiter{limiters: &limiters, defaultConfig: &config, limitersConfigs: &configs, LimiterCh: chLimiter}
+
 	return m
 }
 
 // Run the cleanup routine to remove old entries of Limiters based on ReconcileCheckLimit and ReconcileCheckInterval.
 func (m *MultiLimiter) Run(ctx context.Context) {
+	go m.runStore(ctx)
 	go func() {
 		waiter := time.NewTimer(0)
 		for {
@@ -145,12 +154,38 @@ func (m *MultiLimiter) Allow(e LimitedEntity) bool {
 			config = prefixConfig
 		}
 	}
-
 	limiter := &Limiter{limiter: rate.NewLimiter(config.Rate, config.Burst)}
 	limiter.lastAccess.Store(now)
-	tree, _, _ := limiters.Insert(e.Key(), limiter)
-	m.limiters.Store(tree)
+	m.LimiterCh <- &LimiterWithKey{l: limiter, k: e.Key()}
 	return limiter.limiter.Allow()
+}
+
+func (m *MultiLimiter) runStore(ctx context.Context) {
+	var txn *radix.Txn
+	waiter := time.NewTimer(10 * time.Millisecond)
+	defer waiter.Stop()
+	for {
+		select {
+		case <-waiter.C:
+			if txn != nil {
+				tree := txn.Commit()
+				m.limiters.Store(tree)
+			}
+		case lk := <-m.LimiterCh:
+			limiters := m.limiters.Load()
+			_, ok := limiters.Get(lk.k)
+			if !ok {
+				if txn == nil {
+					txn = limiters.Txn()
+				}
+				txn.Insert(lk.k, lk.l)
+
+			}
+		case <-ctx.Done():
+			return
+		}
+
+	}
 }
 
 // reconcileLimitedOnce is called by the MultiLimiter clean up routine to remove old Limited entries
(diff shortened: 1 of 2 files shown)


agent/consul/multilimiter/multilimiter.go
```diff
@@ -145,14 +155,45 @@ func (m *MultiLimiter) Allow(e LimitedEntity) bool {
 			config = prefixConfig
 		}
 	}
-
 	limiter := &Limiter{limiter: rate.NewLimiter(config.Rate, config.Burst)}
-	limiter.lastAccess.Store(now)
-	tree, _, _ := limiters.Insert(e.Key(), limiter)
-	m.limiters.Store(tree)
+	limiter.lastAccess.Store(unixNow)
+	m.LimiterCh <- &LimiterWithKey{l: limiter, k: e.Key(), t: now}
```
boxofrad, line 160: I think we mentioned it's preferable to drop limiters when the buffer is full rather than blocking. Shall we do a non-blocking send here?
```suggestion
	select {
	case m.LimiterCh <- &LimiterWithKey{l: limiter, k: e.Key(), t: now}:
	default:
		// Write buffer is full. We choose to drop limiters rather than block to prioritize throughput.
	}
```

agent/consul/multilimiter/multilimiter.go
```diff
@@ -28,12 +28,19 @@ type RateLimiter interface {
...
+	t time.Time
+}
+
 // MultiLimiter implement RateLimiter interface and represent a set of rate limiters
 // specific to different LimitedEntities and queried by a LimitedEntities.Key()
 type MultiLimiter struct {
 	limiters        *atomic.Pointer[radix.Tree]
 	limitersConfigs *atomic.Pointer[radix.Tree]
 	defaultConfig   *atomic.Pointer[Config]
+	LimiterCh       chan *LimiterWithKey
```
boxofrad, line 43: nit: could we make this unexported please? 🙇🏻‍♂️

agent/consul/multilimiter/multilimiter.go
```diff
@@ -145,14 +155,45 @@ func (m *MultiLimiter) Allow(e LimitedEntity) bool {
...
+			}
+			v, ok := txn.Get(lk.k)
+			if !ok {
+				txn.Insert(lk.k, lk.l)
+			} else {
+				switch l := v.(type) {
+				case *Limiter:
+					l.lastAccess.Store(lk.t.Unix())
+					l.limiter.AllowN(lk.t, 1)
+				}
```
boxofrad, line 188: No biggie, but if `v` can only ever be a `*Limiter` we can use a more straightforward type-assertion:

```suggestion
				if l, ok := v.(*Limiter); ok {
					l.lastAccess.Store(lk.t.Unix())
					l.limiter.AllowN(lk.t, 1)
				}
```

boxofrad (commented):
Nice job! I've left a few comments - but obviously, feel free to ignore them if we go with #15642.

It looks like the lock-based approach performs better, but I wonder whether that might be partly because we're blocking when `LimiterCh` is full?

agent/consul/multilimiter/multilimiter.go
```diff
@@ -145,14 +155,45 @@ func (m *MultiLimiter) Allow(e LimitedEntity) bool {
 			config = prefixConfig
 		}
 	}
-
 	limiter := &Limiter{limiter: rate.NewLimiter(config.Rate, config.Burst)}
```
boxofrad, line 158: I wonder if constructing new limiters in `runStore` would result in fewer allocations under heavy load?

It might be a micro-optimization, but because we flush new writers every 10ms we'll likely end up creating limiters for the same key over and over.

agent/consul/multilimiter/multilimiter.go
```diff
@@ -145,14 +155,45 @@ func (m *MultiLimiter) Allow(e LimitedEntity) bool {
...
-	tree, _, _ := limiters.Insert(e.Key(), limiter)
-	m.limiters.Store(tree)
+	limiter.lastAccess.Store(unixNow)
+	m.LimiterCh <- &LimiterWithKey{l: limiter, k: e.Key(), t: now}
 	return limiter.limiter.Allow()
 }
 
+func (m *MultiLimiter) runStore(ctx context.Context) {
+	var txn *radix.Txn
+	waiter := time.NewTimer(10 * time.Millisecond)
```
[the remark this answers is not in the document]
dhiaayachi: Fixed that and used a ticker

agent/consul/multilimiter/multilimiter.go
dhiaayachi: After discussion we decided that this no different from locking, so no need to drop the limiter

agent/consul/multilimiter/multilimiter.go
dhiaayachi: We need the limiter to allow the current request, We can still return true but  I don't think that this optimization is necessary.

dhiaayachi:
@boxofrad This should be ready for another round of review before I merge it in the main PR.

agent/consul/multilimiter/multilimiter.go
```diff
@@ -88,13 +100,35 @@ func NewMultiLimiter(c Config) *MultiLimiter {
...
 	return m
 }
 
 // Run the cleanup routine to remove old entries of Limiters based on ReconcileCheckLimit and ReconcileCheckInterval.
 func (m *MultiLimiter) Run(ctx context.Context) {
+	isRunning := m.isRunning.Load()
+	if isRunning {
+		return
+	}
+	m.isRunning.Store(true)
```
boxofrad, line 115: This is thread-unsafe because `isRunning` could change after the `Load` — so we need to do an atomic CAS:

```suggestion
	if swapped := m.isRunning.CompareAndSwap(false, true); !swapped {
		return
	}
```

Alternatively, if we don't need to re-run once it has stopped, the whole thing could be replaced with a `sync.Once` 1️⃣

agent/consul/multilimiter/multilimiter.go
```diff
@@ -88,13 +100,35 @@ func NewMultiLimiter(c Config) *MultiLimiter {
...
 	go func() {
+		defer m.isRunning.Store(false)
```
boxofrad, line 117: We do this in _both_ goroutines, which seems like it could lead to some unexpected behavior. I'd argue we probably don't need to re-run once stopped, but if we do we should use something like a `sync.WaitGroup` to wait for both goroutines to exit.

agent/consul/multilimiter/multilimiter.go
```diff
@@ -145,14 +179,37 @@ func (m *MultiLimiter) Allow(e LimitedEntity) bool {
...
-	m.limiters.Store(tree)
+	limiter.lastAccess.Store(unixNow)
+	m.limiterCh <- &limiterWithKey{l: limiter, k: e.Key(), t: now}
 	return limiter.limiter.Allow()
 }
 
+func (m *MultiLimiter) runStoreOnce(ctx context.Context, waiter *time.Ticker, txn *radix.Txn) *radix.Txn {
+	select {
+	case <-waiter.C:
+		if txn != nil {
```
boxofrad, line 191: Can `txn` ever be nil here? It looks like that would only happen once the context is canceled which would terminate the outer loop.

agent/consul/multilimiter/multilimiter.go
```diff
@@ -145,14 +179,37 @@ func (m *MultiLimiter) Allow(e LimitedEntity) bool {
...
+			tree := txn.Commit()
+			m.limiters.Store(tree)
+			limiters := m.limiters.Load()
+			txn = limiters.Txn()
```
boxofrad, line 195: As `limiters` is only modified by this goroutine, I think we can omit the `Load`:

```suggestion
			txn = tree.Txn()
```

agent/consul/multilimiter/multilimiter_test.go
```diff
@@ -30,58 +31,168 @@ func TestNewMultiLimiter(t *testing.T) {
...
 	m.Run(ctx)
 	key := makeKey([]byte("test"))
 	m.Allow(Limited{key: key})
-	limiters := m.limiters.Load()
+	retry.Run(t, func(r *retry.R) {
+		time.Sleep(100 * time.Millisecond)
+		l := m.limiters.Load()
+		require.NotEqual(r, limiters, l)
+		limiters = l
+	})
```
boxofrad, line 84: The `retry` package has a method that lets you configure the interval between each check:

```suggestion
	retry.RunWith(&retry.Timer{Wait: 100 * time.Millisecond, Timeout: 2 * time.Second}, t, func(r *retry.R) {
		time.Sleep(100 * time.Millisecond)
		l := m.limiters.Load()
		require.NotEqual(r, limiters, l)
		limiters = l
	})
```

agent/consul/multilimiter/multilimiter_test.go
```diff
@@ -30,58 +31,168 @@ func TestNewMultiLimiter(t *testing.T) {
...
 	l, ok = limiters.Get(key)
 	require.True(t, ok)
 	require.NotNil(t, l)
 }
 
+func TestRateLimiterStore(t *testing.T) {
```
boxofrad, line 119: Totally optional/take with a pinch of salt 🧂 

I wonder if it'd be helpful to make the `Ticker` an interface that can be stubbed out in tests. That way you could call the (exported) `Run` method and have the tests trigger "ticks" rather than calling the unexported method directly. It would also mean you wouldn't need to sleep in tests (assuming you had some kind of commit callback as in my previous comment).

dhiaayachi: I tried this and I was partially successful. I was not able to create a direct interface to Ticker because it expose a channel directly so I needed to create a wrapper around it. I also don't like to expose that wrapper outside the package as its only purpose is testing. So I ended up with an hybrid solution where I can mock the ticker in some tests but need to use a real ticker in others, which needed to test the `Run(ctx)` API call. For those I used a retry instead of a sleep.

dhiaayachi:
@boxofrad Thanks again for the review. I went through it and decided to use a sync.Once for the Run call. Please take another look when you have time.

PR #15654, opened 2022-12-02
Export peering cli

nathancoleman:
### Description

### Links

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


command/peering/export/export.go
```diff
@@ -0,0 +1,187 @@
...
+	peerNames        string
+}
+
+func (c *cmd) init() {
+	//This function defines the flags
+
+	c.flags = flag.NewFlagSet("", flag.ContinueOnError)
+
+	c.flags.StringVar(&c.serviceName, "service", "", "(Required) Specify the name of the service you want to export.")
+	c.flags.StringVar(&c.serviceNamespace, "serviceNamespace", "", "(Optional) Specify the Namespace of the service, you want to export. Enterprise only feature")
```
nathancoleman, line 38: @DanStough we were thinking it might make sense to use the `namespace` flag inherited via `flags.Merge(c.flags, c.http.ClientFlags())` (and maybe the partition flag inherited the same way) instead of defining a new flag here. Thoughts?

command/peering/export/export.go
```diff
@@ -0,0 +1,187 @@
...
+						cfg.Services[i].Consumers = append(cfg.Services[i].Consumers, api.ServiceConsumer{Peer: peerName})
+					}
+				}
+			}
+		}
+
+		if !serviceExists {
+			cfg.Services = append(cfg.Services, api.ExportedService{
+				Name:      c.serviceName,
+				Namespace: c.serviceNamespace,
```
nathancoleman, line 135: Notably, we can assign a namespace here even when using OSS. Discussed in the issue w/ @DanStough that we will want to update the config-entry validation on the server to return an error such as `enterprise-only feature`.

command/peering/export/export.go
```diff
@@ -0,0 +1,187 @@
...
+	synopsis = "Export a service"
+	help     = `
+Usage: consul peering export [options] -service <service name> -peers <other cluster name>
+
+  Export a service. The peers provided will be used locally by
+  this cluster to refer to the other cluster where the services will be exported. 
+
+  Example:
+
+	$ consul peering export -service=web -peers=other-cluster
```
nathancoleman, line 185: Might be worth including multiple peers in the example, `-peers=cluster-1,cluster-2`

nathancoleman (commented):
One more question for you @DanStough: should we be making API calls to query for the service being exported to ensure that it exists, or should we depend on the server to validate that, or is that up to the user?

command/peering/export/export.go
DanStough: It doesn't look like we use that flag for anything else relevant to storing the `exported-services`, but it will automatically get added as a query parameter in the request: https://github.com/hashicorp/consul/blob/c046d1a4d870639227baff629ff304a1b72deede/api/api.go#L1019-L1021

No idea how Consul will handle that. You could test it out or use a dedicated field to be safe.

DanStough:
@nathancoleman , w.r.t. checking if the service exists, I would say no. Generally, design things in the CLI to support eventual consistency, so you don't need to worry about the order in which these things are created.

DanStough:
This looks awesome! One note is this would look almost identical if we made a command for exporting to partitions. I wonder if this should be a new top-level command, and we can create a followup tasks to support both `-peers` and `-partitions`?

nathancoleman:
@DanStough the question of whether this single command supports both peers and partitions or whether we have this command and another parition-specific command elsewhere in the CLI is something I'm expecting an answer from Jared on before we merge

nathancoleman:
@DanStough if you could please review this, it'd be super appreciated!

We've split things out such that OSS here only has the `-name` and `-consumer-peers` flags.

The `-namespace`, `-partition` and `-consumer-partitions` flags will be added in followup work based on our discussion yesterday.

command/services/export/export.go
```diff
@@ -0,0 +1,197 @@
...
+	return flags.Usage(c.help, nil)
+}
+
+const (
+	synopsis = "Export a service"
+	help     = `
+Usage: consul services export [options] -name <service name> -consumer-peers <other cluster name>
+
+  Export a service. The peers provided will be used locally by
+  this cluster to refer to the other cluster where the services will be exported.
```
DanStough, line 191: nit: I think there some naunce between `cluster name`, which confused with datacenter and `peer name`, which is a local idea.
```suggestion
Usage: consul services export [options] -name <service name> -consumer-peers <peer names>

  Export a service. The peer names provided will be used locally by
  this cluster to refer to the peer where the services will be exported. 
```

command/services/export/export_oss.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func (c *cmd) init() {
+	//This function defines the flags for OSS.
+	// Flags related to namespaces and partitions are excluded.
+
+	c.flags = flag.NewFlagSet("", flag.ContinueOnError)
+
+	c.flags.StringVar(&c.serviceName, "name", "", "(Required) Specify the name of the service you want to export.")
+	c.flags.StringVar(&c.peerNames, "consumer-peers", "", "A list of peers to export the service to, formatted as a comma-separated list.")
```
DanStough, line 20:
```suggestion
	c.flags.StringVar(&c.peerNames, "consumer-peers", "", "(Required) A list of peers to export the service to, formatted as a comma-separated list.")
```

command/services/export/export.go
```diff
@@ -0,0 +1,197 @@
...
+		cfg, ok := entry.(*api.ExportedServicesConfigEntry)
+		if !ok {
+			c.UI.Error(fmt.Sprintf("Existing config entry has incorrect type: %t", entry))
+			return 1
+		}
+
+		serviceExists := false
+
+		for i, service := range cfg.Services {
+			if service.Name == c.serviceName {
```
DanStough, line 97: Maybe this comparison should move to its own function in `export_oss.go`? That way for enterprise, we can also check the namespace.

command/services/export/export.go
```diff
@@ -0,0 +1,197 @@
...
+					Consumers: buildConsumers(peerNames, partitionNames),
+				},
+			},
+		}
+		c.UI.Info(fmt.Sprintf("%#v", cfg))
+		_, _, err = client.ConfigEntries().Set(&cfg, &api.WriteOptions{})
+		if err != nil {
+			c.UI.Error(fmt.Sprintf("Error creating config entry: %s", err))
+			return 1
+		}
```
DanStough, line 85: Optional: instead of the many levels of nesting here, I might move the logic in the if statement out into helper functions like `CreateExportConfigEntry` and `UpdateExportConfigEntry`.

command/services/export/export_test.go
```diff
@@ -0,0 +1,116 @@
...
+		cmd := New(ui)
+
+		args := []string{
+			"-name=testservice",
+		}
+
+		code := cmd.Run(args)
+		require.Equal(t, 1, code, "err: %s", ui.ErrorWriter.String())
+		require.Contains(t, ui.ErrorWriter.String(), "Missing the required -consumer-peers flag")
+	})
```
DanStough, line 58: I assume this is only true for OSS, so it would need to go in a file `export_test_oss.go`.

DanStough (commented):
Overall looks great! A few small things and one big question: Do we need to have the OSS/ENT functionality separated?  I'm looking at [service register](https://github.com/hashicorp/consul/blob/main/command/services/register/register.go#L64), and you can see we include all of the flags there. Usually, we rely on the server endpoints to correctly reject enterprise-y things if it is open source.

command/services/export/export_oss.go
```diff
@@ -0,0 +1,44 @@
...
+	help     = `
+Usage: consul services export [options] -name <service name> -consumer-peers <other cluster name>
+
+  Export a service. The peers provided will be used locally by
+  this cluster to refer to the other cluster where the services will be exported. 
+
+  Example:
+
+  $ consul services export -name=web -consumer-peers=other-cluster
+`
```
DanStough, line 43: I don't think this split makes sense anymore that all the flags and much of the logic are now in `export.go`. I would consolidate everything into that file.

DanStough (changes requested):
I think we can finish getting rid of the `export_oss.go` go logic since all the flags and some of the logic moved into the shared file.

Did you want to include docs in this PR? There is an empty `export.mdx` file. Your call if you want to add that here or in a follow-up.

command/services/export/export_test.go
```diff
@@ -0,0 +1,152 @@
...
+		args := []string{
+			"-http-addr=" + a.HTTPAddr(),
+			"-name=testservice",
+			"-consumer-partitions=a,b",
+		}
+
+		code := cmd.Run(args)
+		require.Equal(t, 0, code)
+		require.Contains(t, ui.OutputWriter.String(), "Successfully exported service")
+	})
```
DanStough, line 100: In a perfect world, we would move the partition test cases to another file `export_test_ent.go`, but since it looks like we don't have any server-side validation for this, I'm fine leaving it as it is.

DanStough (approved):
🥳 🍾 🎆 :shipit:

command/services/export/export.go
```diff
@@ -0,0 +1,260 @@
...
+		return 1
+	}
+
+	peerNames, err := c.getPeerNames()
+	if err != nil {
+		c.UI.Error(err.Error())
+		return 1
+	}
+
+	partitionNames, err := c.getPartitionNames()
```
jkirschner-hashicorp, line 62: What happens if partitions are specified in OSS?

command/services/export/export.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	ok, _, err := client.ConfigEntries().CAS(cfg, cfg.GetModifyIndex(), nil)
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("Error writing config entry: %s", err))
+		return 1
+	} else if !ok {
+		c.UI.Error(fmt.Sprintf("Config entry was changed during update. Please try again"))
+		return 1
+	}
```
jkirschner-hashicorp, line 106: Nice due diligence in handling this case :)

command/services/export/export.go
```diff
@@ -0,0 +1,260 @@
...
+
+	switch {
+	case len(c.peerNames) > 0 && len(c.partitionNames) > 0:
+		c.UI.Info(fmt.Sprintf("Successfully exported service %q to peers %q and to partitions %q", c.serviceName, c.peerNames, c.partitionNames))
```
jkirschner-hashicorp, line 110: Should we write "cluster peers" instead of "peers"?

I ask because, in the context of HCP, there's also "HVN peering/peers".

Same comment applies to the case immediately below.

(I could go either way)

command/services/export/export.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+	return peerNames, nil
+}
+
+func (c *cmd) getPartitionNames() ([]string, error) {
+	var partitionNames []string
+	if c.partitionNames != "" {
+		partitionNames = strings.Split(c.partitionNames, ",")
+		for _, partitionName := range partitionNames {
+			if partitionName == "" {
```
jkirschner-hashicorp, line 150: Is there any other validation of peer or partition names beyond "is it empty" that the backend does?

I guess the possible harm is that someone could write a config entry for a non-existent partition, or one with a mistyped/garbled name?

That said, we output a string like `Successfully exported service %q to peers %q` afterwards, so that would hopefully show whether something was wrong.

So maybe additional validation isn't important (unless it was really easy to do).

jkirschner-hashicorp (commented):
Left a few optional suggestions for wording clarifications and some questions. Looks good - nice work team!

command/services/export/export.go
nathancoleman: @DanStough and I talked at one point about adding validation in the backend for this as a separate PR.

Today, you can write a config entry that exports to a non-existent peer or partition using `consul config write` - this command is no different since it uses the same config entry mechanism. Based on that, I consider adding more validation on the server side to be outside the scope of this PR. It would be good to get on someone's radar though. Thoughts?

command/services/export/export.go
nathancoleman: I like the additional clarity. Will update

.changelog/15654.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+cli: Adds support for exporting a service to a peer or partition
```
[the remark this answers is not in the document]
nathancoleman: Reworded to talk about "adding a command for" instead of "adding support for"

command/services/export/export.go
nathancoleman: This follows the same theme as other enterprise features in this PR. The command with partitions specified will go through even for OSS users, same as if you use `$ consul config write` today (since this is really all that's happening behind the scenes).

Since the established pattern is to allow Enterprise features in the OSS CLI, the correct thing to do in the longer term IMO is to add server validation for this config entry that returns an error in this case.

jkirschner-hashicorp (approved):
LGTM :)

david-yu:
@nathancoleman Out of curiosity are there docs changes for this features as well? I currently do not see docs for this feature here: https://developer.hashicorp.com/consul/commands/services

nathancoleman:
@david-yu good catch - meant to get docs done before the 1.16 release and completely forgot. I'll get a PR up today and backport in 1.16.x.

PR #15659, opened 2022-12-02
connect: Add support for ConsulResolver to specifies a filter expression

pierluc-codes:
### Description
Add `Filter` to `ConsulResolver` to be able to specifies an expression for filtering the query results.

### PR Checklist
* [x] updated test coverage
* [ ] not a security concern

I don't know how secure is [hashicorp/go-bexpr](https://github.com/hashicorp/go-bexpr)


connect/resolver_test.go
```diff
@@ -68,13 +68,17 @@ func TestConsulResolver_Resolve(t *testing.T) {
 		Proxy: &api.AgentServiceConnectProxyConfig{
 			DestinationServiceName: "web",
 		},
+		Tags: []string{
+			"test-tag",
+		},
```
banks, line 73: nit: we view tags as a semi-deprecated construct in Consul registrations - everything that can be done with tag can also be done with `Meta` and `Filter`. Long term we'd like to consolidate on just `Meta` but have never been able to prioritize the one or two places (e.g. DNS) where tags still do something that Meta doesn't currently.

So I'd accept this - it's not _wrong_ and we've not officially announced or documented that "tags" are deprecated, but it is for example the reason we don't support tag filters in connect endpoints etc.

Would you be open to writing this test using Meta instead? It's not a huge deal either way, but internally we'd tend towards supporting/encouraging use of meta of tags as we _might_ one day get around to deprecating and removing tags entirely!

banks (approved):
This LGTM, Thanks for contributing!

I have one minor optional suggestion inline but fine to merge as is!

curtbushko (approved):
Thanks for doing this!

PR #15660, opened 2022-12-02
Fix broken link to Consul Dataplane index

eculver:
### Description
The `/index` appears to result in a 404.

Removing the `/index` in the URL results in the correct path loading.

### Links
- [Consul on Kubernetes: Architecture](https://developer.hashicorp.com/consul/docs/k8s/architecture)


nathancoleman (approved):
Was this link valid at one point? Wondering if it should forward to the URL you replaced it with

david-yu:
I don't think it was valid and is mainly a typo. Will merge.

PR #15661, opened 2022-12-02
Detect Vault 1.11+ import in secondary datacenters and update default issuer

rboyer:
### Description
The fix outlined and merged in https://github.com/hashicorp/consul/pull/15253 fixed the issue as it occurs in the primary DC. There is a similar issue that arises when vault is used as the Connect CA in a secondary datacenter that is fixed by this PR.

Additionally: this PR adds support to run the existing suite of vault related integration tests against the last 4 versions of vualt (1.9, 1.10, 1.11, 1.12)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,6 @@ references:
     GIT_COMMITTER_NAME: circleci-consul
     S3_ARTIFACT_BUCKET: consul-dev-artifacts-v2
     BASH_ENV: .circleci/bash_env.sh
-    VAULT_BINARY_VERSION: 1.9.4
     GO_VERSION: 1.19.2
   envoy-versions: &supported_envoy_versions
     - &default_envoy_version "1.21.5"
@@ -32,6 +31,11 @@ references:
     - &default_nomad_version "1.3.3"
     - "1.2.10"
     - "1.1.16"
+  vault-versions: &supported_vault_versions
+    - &default_vault_version "1.12.2"
+    - "1.11.6"
+    - "1.10.9"
+    - "1.9.10"
   images:
     # When updating the Go version, remember to also update the versions in the
     # workflows section for go-test-lib jobs.
@@ -587,7 +591,6 @@ jobs:
       - setup_remote_docker
       - run: make ci.dev-docker
       - run: *notify-slack-failure
-
   nomad-integration-test: &NOMAD_TESTS
     docker:
       - image: docker.mirror.hashicorp.services/cimg/go:1.19
@@ -939,9 +942,15 @@ jobs:
   test-connect-ca-providers:
     docker:
       - image: *GOLANG_IMAGE
+    parameters:
+      vault-version:
+        type: enum
+        enum: *supported_vault_versions
+        default: *default_vault_version
     environment:
       <<: *ENVIRONMENT
-    steps:
+      VAULT_BINARY_VERSION: << parameters.vault-version >>
+    steps: &VAULT_INTEGRATION_TEST_STEPS
       - run:
           name: Install vault
           command: |
@@ -1067,8 +1076,12 @@ workflows:
           name: "lint-32bit"
           go-arch: "386"
           <<: *filter-ignore-non-go-branches
-      - test-connect-ca-providers: *filter-ignore-non-go-branches
       - go-test-arm64: *filter-ignore-non-go-branches
+      - test-connect-ca-providers:
+          matrix:
+            parameters:
+              vault-version: *supported_vault_versions
+          <<: *filter-ignore-non-go-branches
       - dev-build: *filter-ignore-non-go-branches
       - go-test:
           requires: [dev-build]
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -938,8 +938,8 @@ jobs:
           path: *TEST_RESULTS_DIR
       - run: *notify-slack-failure
 
-  # run integration tests for the connect ca providers
-  test-connect-ca-providers:
+  # run integration tests for the connect ca providers with vault
+  vault-integration-test:
     docker:
       - image: *GOLANG_IMAGE
     parameters:
@@ -1077,12 +1077,6 @@ workflows:
           go-arch: "386"
           <<: *filter-ignore-non-go-branches
       - go-test-arm64: *filter-ignore-non-go-branches
-      - test-connect-ca-providers:
-          name: "go-test-connect-ca with vault"
-          matrix:
-            parameters:
-              vault-version: *supported_vault_versions
-          <<: *filter-ignore-non-go-branches
       - dev-build: *filter-ignore-non-go-branches
       - go-test:
           requires: [dev-build]
@@ -1162,6 +1156,11 @@ workflows:
           matrix:
             parameters:
               nomad-version: *supported_nomad_versions
+      - vault-integration-test:
+          matrix:
+            parameters:
+              vault-version: *supported_vault_versions
+          <<: *filter-ignore-non-go-branches
       - envoy-integration-test:
           requires:
             - dev-build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1078,6 +1078,7 @@ workflows:
           <<: *filter-ignore-non-go-branches
       - go-test-arm64: *filter-ignore-non-go-branches
       - test-connect-ca-providers:
+          name: "go-test-connect-ca with vault"
           matrix:
             parameters:
               vault-version: *supported_vault_versions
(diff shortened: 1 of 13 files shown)


agent/connect/ca/provider.go
```diff
@@ -179,14 +179,17 @@ type SecondaryProvider interface {
 	//
 	// After the certificate is signed, SecondaryProvider.SetIntermediate will
 	// be called to store the intermediate CA.
-	GenerateIntermediateCSR() (string, error)
+	//
+	// The second return value is an opaque string meant to be passed back to
+	// the subsequent call to SetIntermediate.
+	GenerateIntermediateCSR() (string, string, error)
```
rboyer, line 185: For the vault provider we needed a way to pass extra data from the `GenerateIntermediateCSR` step to the `SetIntermediate` step so an extra output/input pair was setup to thread an opaque string through.

The vault provider will pass a `keyId`.

agent/connect/ca/provider_vault_test.go
```diff
@@ -924,6 +924,76 @@ func TestVaultCAProvider_GenerateIntermediate(t *testing.T) {
 	require.NotEqual(t, orig, new)
 }
 
+func TestVaultCAProvider_GenerateIntermediate_inSecondary(t *testing.T) {
```
rboyer, line 927: This is a mirror of the test added in https://github.com/hashicorp/consul/pull/15253 to validate the primary side of the bug.

agent/consul/leader_connect_ca_test.go
```diff
@@ -1033,9 +1033,14 @@ func setupPrimaryCA(t *testing.T, client *vaultapi.Client, path string, rootPEM
...
 	require.NoError(t, err, "failed to sign intermediate")
 
+	cert := intermediate.Data["certificate"].(string)
+
 	var buf strings.Builder
-	buf.WriteString(lib.EnsureTrailingNewline(intermediate.Data["certificate"].(string)))
-	buf.WriteString(lib.EnsureTrailingNewline(rootPEM))
+	buf.WriteString(lib.EnsureTrailingNewline(cert))
+	if !strings.Contains(strings.TrimSpace(cert), strings.TrimSpace(rootPEM)) {
+		// Vault < v1.11 included the root in the output of sign-intermediate.
```
rboyer, line 1041: NOTE: this needed to be done to get the tests that depend on this to properly pass on vault 1.11+

agent/consul/leader_connect_ca.go
```diff
@@ -1052,7 +1052,7 @@ func (c *CAManager) primaryRenewIntermediate(provider ca.Provider, newActiveRoot
 // provider.
 // Should only be called while the state lock is held by setting the state to non-ready.
 func (c *CAManager) secondaryRequestNewSigningCert(provider ca.Provider, newActiveRoot *structs.CARoot) error {
-	csr, err := provider.GenerateIntermediateCSR()
+	csr, opaque, err := provider.GenerateIntermediateCSR()
```
rboyer, line 1055: Here we see the actual data ferrying from one to the other.

agent/connect/ca/provider_vault.go
```diff
@@ -496,13 +497,21 @@ func (v *VaultProvider) SetIntermediate(intermediatePEM, rootPEM string) error {
...
 
-	_, err = v.writeNamespaced(v.config.IntermediatePKINamespace, v.config.IntermediatePKIPath+"intermediate/set-signed", map[string]interface{}{
+	importResp, err := v.writeNamespaced(v.config.IntermediatePKINamespace, v.config.IntermediatePKIPath+"intermediate/set-signed", map[string]interface{}{
 		"certificate": intermediatePEM,
 	})
 	if err != nil {
 		return err
 	}
 
+	// Vault 1.11+ will return a non-nil response from intermediate/set-signed
```
rboyer, line 507: This is lifted verbatim from the former fix.

.circleci/config.yml
```diff
@@ -935,19 +938,26 @@ jobs:
           path: *TEST_RESULTS_DIR
       - run: *notify-slack-failure
 
-  # run integration tests for the connect ca providers
-  test-connect-ca-providers:
+  # run integration tests for the connect ca providers with vault
+  vault-integration-test:
```
rboyer, line 942: The new name is more right than the former name given what the make target does:
```
test-connect-ca-providers:
ifeq ("$(CIRCLECI)","true")
# Run in CI
	gotestsum --format=short-verbose --junitfile "$(TEST_RESULTS_DIR)/gotestsum-report.xml" -- -cover -coverprofile=coverage.txt ./agent/connect/ca
# Run leader tests that require Vault
	gotestsum --format=short-verbose --junitfile "$(TEST_RESULTS_DIR)/gotestsum-report-leader.xml" -- -cover -coverprofile=coverage-leader.txt -run Vault ./agent/consul
# Run agent tests that require Vault
	gotestsum --format=short-verbose --junitfile "$(TEST_RESULTS_DIR)/gotestsum-report-agent.xml" -- -cover -coverprofile=coverage-agent.txt -run Vault ./agent
else
# Run locally
	@echo "Running /agent/connect/ca tests in verbose mode"
	@go test -v ./agent/connect/ca
	@go test -v ./agent/consul -run Vault
	@go test -v ./agent -run Vault
endif
```

.circleci/config.yml
```diff
@@ -1067,7 +1077,6 @@ workflows:
           name: "lint-32bit"
           go-arch: "386"
           <<: *filter-ignore-non-go-branches
-      - test-connect-ca-providers: *filter-ignore-non-go-branches
```
rboyer, line 1070: I moved it to live with the other integrations.

agent/connect/ca/provider_aws.go
```diff
@@ -75,6 +75,8 @@ type AWSProvider struct {
 	logger          hclog.Logger
 }
 
+var _ Provider = (*AWSProvider)(nil)
```
jkirschner-hashicorp, line 78: I noticed this change was added to all 3 providers. What is the consequence of this change?

rboyer: It's a compile-time no-op type assertion so a failure to make the provider match the `Provider` interface fails in the definition of the struct not when you try to use it.

agent/connect/ca/provider_vault.go
jkirschner-hashicorp: Should we wrap this shared bit of code in a common for DRYness?

agent/connect/ca/provider_aws.go
jkirschner-hashicorp: Ah - that's really cool!

agent/connect/ca/provider_vault.go
rboyer: That already is in `setDefaultIntermediateIssuer`

agent/connect/ca/provider_vault_test.go
```diff
@@ -924,6 +924,76 @@ func TestVaultCAProvider_GenerateIntermediate(t *testing.T) {
...
+	_, err := primaryProvider.GenerateRoot()
+	require.NoError(t, err)
+
+	// Ensure that we don't configure vault to try and mint leafs that
+	// outlive their CA during the test (which hard fails in vault).
+	intermediateCertTTL := getIntermediateCertTTL(t, conf)
+	leafCertTTL := intermediateCertTTL - 4*time.Hour
+
+	provider, _ := testVaultProviderWithConfig(t, false, map[string]any{
+		"LeafCertTTL": []uint8(leafCertTTL.String()),
```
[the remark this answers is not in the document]
kisunji: I also don't think the conversion is necessary 🤔 Assuming go decodes it to a string anyways

agent/connect/ca/provider_vault_test.go
```diff
@@ -924,6 +924,76 @@ func TestVaultCAProvider_GenerateIntermediate(t *testing.T) {
...
+	})
+
+	var origIntermediate string
+	testutil.RunStep(t, "initialize secondary provider", func(t *testing.T) {
+		// Get the intermediate CSR from provider.
+		csrPEM, opaque, err := provider.GenerateIntermediateCSR()
```
kisunji, line 950: Since this is a VaultProvider test, can we make this specific for those reading?
```suggestion
		csrPEM, issuerId, err := provider.GenerateIntermediateCSR()
```

kisunji (approved):
I hope the Vault tests backport well 🤞

agent/connect/ca/provider_vault.go
kisunji: Yeah when I was refactoring `setDefaultIntermediateIssuer` I thought about bringing the nil check inside the function. I concluded that it's more useful to be explicitly loud about some behavioural assumptions than to call `setDefaultIntermediateIssuer` for all versions but secretly have it be a no-op for some versions.

agent/connect/ca/provider_vault_test.go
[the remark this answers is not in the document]
rboyer: IIRC `msgpack` round tripping will turn `[]byte` into `[]uint8` and then if you do a deep equality on the resulting things stuffed into `any` they would fail.

I did just move the type around, but I suspect ^^ is why it might be like this in the test, or at least inspired it.

PR #15669, opened 2022-12-05
connect: ensure all vault connect CA tests use limited privilege tokens

rboyer:
### Description

All of the current integration tests where Vault is the Connect CA now use non-root tokens for the test. This helps us detect privilege changes in the vault model so we can keep our guides up to date.

One larger change was that the `RenewIntermediate` function got refactored slightly so it could be used from a test, rather than the large duplicated function we were testing in a test which seemed error prone.

### DIFF
--- a/agent/connect/ca/testing.go
+++ b/agent/connect/ca/testing.go
@@ -315,8 +315,6 @@ path "%[2]s/*" {
 		if a.WithSudo {
 			rules += fmt.Sprintf(`
 
-# TODO(rb): not present in docs; needed for provider swap FROM vault
-# /v1/%[1]s/root/sign-self-issued
 path "%[1]s/root/sign-self-issued" {
   capabilities = [ "sudo", "update" ]
 }


agent/connect/ca/provider_vault_test.go
```diff
@@ -193,24 +212,35 @@ func TestVaultCAProvider_Configure(t *testing.T) {
 }
 
 func TestVaultCAProvider_SecondaryActiveIntermediate(t *testing.T) {
-
 	SkipIfVaultNotPresent(t)
 
-	provider, testVault := testVaultProviderWithConfig(t, false, nil)
-	defer testVault.Stop()
+	t.Parallel()
```
kisunji, line 217: Are these safe to run parallel now?

agent/connect/ca/provider_vault_test.go
```diff
@@ -464,94 +530,168 @@ func TestVaultCAProvider_SignLeaf(t *testing.T) {
...
 func TestVaultProvider_SignIntermediate(t *testing.T) {
-
 	SkipIfVaultNotPresent(t)
 
+	t.Parallel()
+
 	tests := CASigningKeyTypeCases()
 
 	for _, tc := range tests {
-		tc := tc
```
kisunji, line 526: We actually need this now that we're running `t.Parallel()`, right?

agent/consul/leader_connect_ca.go
```diff
@@ -1139,81 +1136,113 @@ func (c *CAManager) runRenewIntermediate(ctx context.Context) error {
...
+}
+
+func (c *CAManager) renewIntermediateNow(ctx context.Context) error {
+	renewed, err := c.renewIntermediate(ctx, true, true)
+	if err != nil {
+		return err
+	}
+	if !renewed {
+		return errors.New("expected to force renew and did not")
+	}
```
kisunji, line 1151: It feels awkward to add a `bool` return variable for this case which currently isn't reachable. Is there a need to be defensive about it? I can't see `forceRenewAheadOfSchedule` being misused for anything else other than a guard for the lifetime check.

agent/consul/leader_connect_ca.go
```diff
@@ -1139,81 +1136,113 @@ func (c *CAManager) runRenewIntermediate(ctx context.Context) error {
...
 	if err != nil {
-		return fmt.Errorf("error parsing active intermediate cert: %v", err)
+		return false, fmt.Errorf("error parsing active intermediate cert: %v", err)
 	}
 
-	if lessThanHalfTimePassed(c.timeNow(), intermediateCert.NotBefore, intermediateCert.NotAfter) {
-		return nil
+	if !forceRenewAheadOfSchedule {
+		if lessThanHalfTimePassed(c.timeNow(), intermediateCert.NotBefore, intermediateCert.NotAfter) {
+			return false, nil
```
kisunji, line 1208: Is this the only time we return `(false, nil)`? Wondering if we can alternatively return a new err

agent/connect/ca/provider_vault_test.go
```diff
@@ -464,94 +530,168 @@ func TestVaultCAProvider_SignLeaf(t *testing.T) {
...
 func TestVaultCAProvider_CrossSignCA(t *testing.T) {
-
 	SkipIfVaultNotPresent(t)
 
+	t.Parallel()
+
 	tests := CASigningKeyTypeCases()
 
-	for _, tc := range tests {
-		tc := tc
```
kisunji, line 473: Should this be preserved

kisunji: Oh I guess in this particular case it's captured by `run`

agent/connect/ca/provider_vault_test.go
rboyer: They always were I believe. Each test gets its own vault process.

I added this b/c it shaved the runtime down by a factor of ~5. Locally it was ~28s -> ~6s.

agent/connect/ca/provider_vault_test.go
rboyer: I did the `run(t, tc)` thing here instead.

agent/connect/ca/testing.go
```diff
@@ -238,3 +232,129 @@ func requireTrailingNewline(t testing.T, leafPEM string) {
...
+		t.Fatal("TODO: implement this and use it in tests")
+		return ""
+
+	default:
+		// zero value
+		return ""
+	}
+}
+
+func CreateVaultTokenWithAttrs(t testing.T, client *vaultapi.Client, attr *VaultTokenAttributes) string {
```
rboyer, line 339: New test helper for use outside of the package too.

agent/consul/acl_server.go
```diff
@@ -87,7 +87,7 @@ func (s *Server) checkBindingRuleUUID(id string) (bool, error) {
 }
 
 func (s *Server) InPrimaryDatacenter() bool {
-	return s.config.PrimaryDatacenter == "" || s.config.Datacenter == s.config.PrimaryDatacenter
+	return s.config.InPrimaryDatacenter()
```
rboyer, line 90: Convenience for a test in `leader_connect_ca_test.go`

agent/consul/config.go
```diff
@@ -415,6 +415,10 @@ type Config struct {
 	*EnterpriseConfig
 }
 
+func (c *Config) InPrimaryDatacenter() bool {
```
rboyer, line 418: Convenience for a test in `leader_connect_ca_test.go`

agent/consul/leader_connect_ca_test.go
```diff
@@ -778,37 +826,62 @@ func TestCAManager_Initialize_Vault_WithExternalTrustedCA(t *testing.T) {
```
rboyer, line 826: For the record `TestCAManager_Initialize_Vault_WithExternalTrustedCA` was a test I invested a lot in because it surfaced a bunch of strange things such as https://github.com/hashicorp/consul/pull/15661

agent/connect/ca/provider_vault_test.go
```diff
@@ -157,12 +161,15 @@ func TestVaultCAProvider_Configure(t *testing.T) {
...
-		rawConfig     map[string]interface{}
+		rawConfig     map[string]any
 		expectedValue func(t *testing.T, v *VaultProvider)
 	}{
 		{
-			name:      "DefaultConfig",
-			rawConfig: map[string]interface{}{},
+			name: "DefaultConfig",
+			rawConfig: map[string]any{
+				"RootPKIPath":         "pki-root/",
```
rboyer, line 170: To avoid accidentally triggering two providers to share some vault mount paths in tests, I changed it so that test authors need to explicitly set the paths all the time.

agent/connect/ca/provider_vault_test.go
```diff
@@ -1118,23 +1326,8 @@ func getIntermediateCertTTL(t *testing.T, caConf *structs.CAConfiguration) time.
 	return dur
 }
 
-func testVaultProviderWithConfig(t *testing.T, isPrimary bool, rawConf map[string]interface{}) (*VaultProvider, *TestVaultServer) {
```
rboyer, line 1121: I inlined `testVaultProviderWithConfig` so that we can create the token in between the two steps.

agent/connect/ca/provider_vault_test.go
```diff
@@ -380,178 +434,272 @@ func assertCorrectKeyType(t *testing.T, want, certPEM string) {
...
 
-				intPEM, err := provider1.ActiveIntermediate()
-				require.NoError(t, err)
-				assertCorrectKeyType(t, tc.SigningKeyType, intPEM)
-			}
+		attr1 := &VaultTokenAttributes{
+			RootPath:         "pki-root",
+			IntermediatePath: "pki-intermediate",
+			ConsulManaged:    true,
+			WithSudo:         withSudo,
```
rboyer, line 558: This validates the continued need for the `sudo` capability now.

agent/connect/ca/provider_vault_test.go
```diff
@@ -380,178 +434,272 @@ func assertCorrectKeyType(t *testing.T, want, certPEM string) {
...
+			root, err := provider2.GenerateRoot()
+			require.NoError(t, err)
+			assertCorrectKeyType(t, tc.CSRKeyType, root.PEM)
+
+			intPEM, err := provider2.ActiveIntermediate()
+			require.NoError(t, err)
+			assertCorrectKeyType(t, tc.CSRKeyType, intPEM)
+
+			if expectFailure {
+				testCrossSignProvidersShouldFail(t, provider1, provider2)
```
rboyer, line 608: When we repeat the test without `sudo` we expect it to fail.

agent/connect/ca/testing.go
```diff
@@ -238,3 +232,129 @@ func requireTrailingNewline(t testing.T, leafPEM string) {
...
+  capabilities = [ "sudo", "update" ]
+}
+`, a.RootPath)
+		}
+
+		return rules
+
+	case a.VaultManaged:
+		// Vault-managed PKI root.
+		t.Fatal("TODO: implement this and use it in tests")
```
rboyer, line 330: This stub is here as a reminder to add more tests for the other policy in our docs.

agent/consul/leader_connect_ca.go
rboyer: That's fair. I guess the test would fail if it didn't renew. I'll remove it.

kisunji (approved):

agent/connect/ca/testing.go
```diff
@@ -238,3 +232,129 @@ func requireTrailingNewline(t testing.T, leafPEM string) {
...
+
+path "%[2]s/*" {
+  capabilities = [ "create", "read", "update", "delete", "list" ]
+}
+`, a.RootPath, a.IntermediatePath)
+
+		if a.WithSudo {
+			rules += fmt.Sprintf(`
+
+# TODO(rb): not present in docs; needed for provider swap FROM vault
```
jkirschner-hashicorp, line 318: The docs were just updated: https://developer.hashicorp.com/consul/docs/connect/ca/vault#additional-vault-acl-policies-for-sensitive-operations

PR #15670, opened 2022-12-05
add log-drop package

dhiaayachi:
### Description
This add a new logger package that wrap another logger. The logger would start a go routine to consume logs and write any log into a buffered channel if there is room, otherwise drop the log and increment a counter.
The go routine would read the logs from the channel and log them into the underlying logger.

* Unit tests using a mock logger as an underlying logger

* [X] updated test coverage
* [x] ~external facing docs updated~
* [X] not a security concern


agent/log-drop/log-drop.go
```diff
@@ -0,0 +1,70 @@
...
+	DEBUG
+	INFO
+	WARN
+	ERROR
+)
+
+//go:generate mockery --name Logger --inpackage
+type Logger interface {
+	Info(string, ...interface{})
+}
```
boxofrad, line 20: What do you think of implementing the [`hclog.Logger` interface](https://pkg.go.dev/github.com/hashicorp/go-hclog#Logger) to make it easier to slot into places where we already depend on it?

agent/log-drop/log-drop.go
```diff
@@ -0,0 +1,70 @@
...
+
+type Log struct {
+	s string
+	i []interface{}
+	l Level
+}
+
+type logDrop struct {
+	logger Logger
+	logCH  chan Log
```
boxofrad, line 30: nit 😄 

```suggestion
	logCh  chan Log
```

agent/log-drop/log-drop.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+type Level int
+
+const (
+	TRACE Level = iota
+	DEBUG
+	INFO
+	WARN
+	ERROR
```
boxofrad, line 14: Similarly, we could use [`hclog.Level`](https://pkg.go.dev/github.com/hashicorp/go-hclog#Level) here.

agent/log-drop/log-drop.go
```diff
@@ -0,0 +1,70 @@
+package log_drop
```
boxofrad, line 1: Could we follow [Effective Go's guidance](https://go.dev/doc/effective_go#package-names) and remove the underscore (hyphen in the directory name)?

```suggestion
package logdrop
```

agent/log-drop/log-drop_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+func TestNewLogDrop(t *testing.T) {
+	mockLogger := NewMockLogger(t)
+	mockLogger.On("Info", "test Log", []interface{}{"val", 0}).Return()
+	ld := NewLogDrop(context.Background(), "test", 10, mockLogger, func(_ Log) {})
+	require.NotNil(t, ld)
+	ld.Info("test Log", "val", 0)
+	time.Sleep(10 * time.Millisecond)
+	mockLogger.AssertNumberOfCalls(t, "Info", 1)
+}
```
boxofrad, line 19: I think we can remove the sleep here with some clever mock/channel tricks 🤓

```suggestion
func TestNewLogDrop(t *testing.T) {
	callCh := make(chan struct{})
	mockLogger := NewMockLogger(t)
	mockLogger.On("Info", "test Log", []interface{}{"val", 0}).
		Run(func(mock.Arguments) { close(callCh) }).
		Return()
	
	ld := NewLogDrop(context.Background(), "test", 10, mockLogger, func(_ Log) {})
	require.NotNil(t, ld)
	
	ld.Info("test Log", "val", 0)
	
	select {
	case <-callCh:
	case <-time.After(100 * time.Millisecond):
		t.Fatal("timeout waiting for logger to be called")
	}
}
```

agent/log-drop/log-drop_test.go
```diff
@@ -0,0 +1,35 @@
...
+	ld := NewLogDrop(ctx, "test", 1, mockLogger, func(l Log) {
+		called = true
+	})
+	cancelFunc()
+	time.Sleep(1 * time.Second)
+	for i := 0; i < 2; i++ {
+		ld.Info("test", "test", "hello")
+	}
+	mockLogger.AssertNotCalled(t, "Info", mock.Anything)
+	require.True(t, called)
```
boxofrad, line 34: This one too 🕐 🔥 

```suggestion
	mockLogger := NewMockLogger(t)
	
	ctx, cancelFunc := context.WithCancel(context.Background())
	cancelFunc()

	calls := make(chan Log, 2)
	ld := NewLogDrop(ctx, "test", 1, mockLogger, func(l Log) {
		calls <- l
	})

	const numCalls = 2
	for i := 0; i < numCalls; i++ {
		ld.Info("test", "test", "hello")
	}
	
	for i := 0; i < numCalls; i++ {
		select {
		case <-callCh:
		case <-time.After(100 * time.Millisecond):
			t.Fatal("timeout waiting for drop func to be called")
		}
	}
	
	mockLogger.AssertNotCalled(t, "Info", mock.Anything)
```

agent/log-drop/log-drop.go
```diff
@@ -0,0 +1,70 @@
...
+			case INFO:
+				r.logger.Info(l.s, l.i)
+			}
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+func NewLogDrop(ctx context.Context, name string, depth int, logger Logger, dropFn func(l Log)) Logger {
```
boxofrad, line 61: Could we add some godoc comments here too please? 🙇🏻‍♂️ I think it's particularly important to point out that it starts a goroutine which needs to be stopped by canceling the given context.

boxofrad (commented):
Looking great! Left some comments and questions for your consideration 🙇🏻‍♂️

agent/log-drop/log-drop.go
dhiaayachi: I thought about that but I think implementing the full interface is a bit overkilled for what we need for now. At the same time I see a use for this kind of logger in places where we would like to add some debug logging without adding too much performance hit 🤷

boxofrad: Yeah, I think it'd be ok to embed the wrapped logger and only implement the methods we need:

```Go
var _ hclog.Logger = (*Logger)(nil)

type Logger struct {
  hclog.Logger
}

// We just implement these methods.
func (l *Logger) Info(s string, i ...interface{}) {
  r.pushLog(Log{l: INFO, s: s, i: i})
}

func (l *Logger) Debug(s string, i ...interface{}) {
  r.pushLog(Log{l: DEBUG, s: s, i: i})
}

// Everything else gets called on the base logger.
```

dhiaayachi: I think I have an idea. I can implement a log Sink instead. Which would give us the opportunity to use the same log interceptor we usually use and make it apply to all hclog.Logger interface. 

I should have something around that in the PR shortly 😄

boxofrad:
I really like the sink approach @dhiaayachi, nice job 👏🏻 

I notice that we're still wrapping the logger directly, and wonder if we should be wrapping another sink?

```
+--------+    +---------------+    +----------------------------------+
| Logger | => | Dropping Sink | => | Real Sink (STDOUT, File, Syslog) |
+--------+    +---------------+    +----------------------------------+
```

I may be wrong though! I'm not super familiar with hclog internals.

agent/log-drop/log-drop_test.go
```diff
@@ -0,0 +1,37 @@
...
+		mockLogger.AssertNumberOfCalls(r, "Log", 1)
+	})
+
+}
+
+func TestLogDroppedWhenChannelFilled(t *testing.T) {
+	mockLogger := NewMockLogger(t)
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	called := false
+	cancelFunc()
```
boxofrad, line 28: I think we're canceling the context here to cause the channel to get backed-up. I wonder if a more realistic test would be to have the underlying logger/sink we're wrapping block (e.g. have it receive on a nil channel).

boxofrad (approved):
Awesome work! 👏🏻

PR #15690, opened 2022-12-06
Fix local mesh gateway with peering discovery chains.

hashi-derek:
Prior to this patch, discovery chains with peer-targets would not properly honor the mesh gateway mode for two reasons.

1. An incorrect target upstream ID was used to lookup the mesh gateway mode. To fix this, the parent upstream uid is now used instead of the discovery-chain-target-uid to find the intended mesh gateway mode.

2. The watch for local mesh gateways was never initialized for discovery chains. To fix this, the discovery chains are now scanned, and a local GW watch is spawned if: the mesh gateway mode is local and the target is a peering connection.

### DIFF
--- a/agent/proxycfg/upstreams.go
+++ b/agent/proxycfg/upstreams.go
@@ -82,6 +82,15 @@ func (s *handlerUpstreams) handleUpdateUpstreams(ctx context.Context, u UpdateEv
 				s.logger.Trace("discovery-chain watch fired for unknown upstream", "upstream", uid)
 				return nil
 			}
+
+			// We have to inspect the discovery chains to see if local mesh gateways need to be watched.
+			for _, target := range resp.Chain.Targets {
+				// Register a gateway watch if any targets are pointing to a peer and require a mode of local.
+				if target.Peer != "" && target.MeshGateway.Mode == structs.MeshGatewayModeLocal {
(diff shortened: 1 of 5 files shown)


agent/proxycfg/upstreams.go
```diff
@@ -82,6 +82,15 @@ func (s *handlerUpstreams) handleUpdateUpstreams(ctx context.Context, u UpdateEv
 				s.logger.Trace("discovery-chain watch fired for unknown upstream", "upstream", uid)
 				return nil
 			}
+
+			// We have to inspect the discovery chains to see if local mesh gateways need to be watched.
```
hashi-derek, line 86: Should this also be done for ingress gateways above?

freddygv (commented):
Nice work figuring this out! I left a comment below about where the discovery chain inspection is happening.

By the way, would be good to update `state_test.go` to verify this is working. Could add something like this to `verifySnapshot` of `stage0` and `stage1` of `newConnectProxyCase`. There is an upstream there called `api-failover-to-peer` which should be impacted by this issue.

```go
if meshGatewayProxyConfigValue == structs.MeshGatewayModeLocal {
	require.True(t, snap.ConnectProxy.WatchedLocalGWEndpoints.IsWatched("dc1"))
	_, ok := snap.ConnectProxy.WatchedLocalGWEndpoints.Get("dc1")
	require.False(t, ok)
}
```

PR #15700, opened 2022-12-06
Pass remote addr of incoming HTTP requests through to RPC(..) calls

analogue:
### Description
The consul server agent http api handlers make local RPC calls. We want to make sure that these calls are subject to the preBodyInterceptor introduced in `consul-net-rpc` as part of [PR 17](https://github.com/hashicorp/consul-net-rpc/pull/17). 

This entails:
- Adding a middleware HTTP handler to inject the remote addr into the request's `context.Context`
- Refactoring all `RPC(method, args, reply)` to accept `context.Context` as their first argument
- Updating all HTTP handlers to pass the incoming request's context on to any `RPC(ctx, ...)` calls
- The `consul.Server` implementation of `RPC(ctx, ...)` has been updated to extract the remote addr from the context and pass it on to the `inmemCodec`.
- Finally, `inmemCodec.SourceAddr()` returns the remote addr which is used by the `PreBodyInterceptor`.

This does not include wiring up of the ratelimiter for net/rpc calls. That will be in a subsequent PR since this one already has 108 files with changes.

I tried to isolate the meaningful changes into a single commit so you don't have to bother with the noise in the remaining diffs which are all changes to the RPC method signature: https://github.com/hashicorp/consul/pull/15700/commits/d29923255b49e646d2312d7ad3949ff7dcbfc354   

### Testing
* Manually testing with `curl "http://localhost:8500/v1/catalog/services"` and observed remote addr getting threaded through all the way to RPC call.
* More meaningful end-to-end test planned for the next PR which will wire up the rate limiter and exercise it.

### Assumptions
* If the http handler middleware has any issues converting the string form of the remoteAddr into a net.Addr, it just doesn't inject it into the context.
* Accordingly, implementations of `preBodyInterceptor` are expected to deal with a nil `sourceAddr`

### PR Checklist

* [x] updated test coverage
* [ ] not a security concern

### DIFF
--- a/agent/http.go
+++ b/agent/http.go
@@ -1,7 +1,6 @@
 package agent
 
 import (
-	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -105,24 +104,6 @@ var endpoints map[string]unboundEndpoint
 // allowedMethods is a map from endpoint prefix to supported HTTP methods.
 // An empty slice means an endpoint handles OPTIONS requests and MethodNotFound errors itself.
 var allowedMethods map[string][]string = make(map[string][]string)
-
-// contextKey is a value for use with context.WithValue. It's used as
-// a pointer so it fits in an interface{} without allocation.
-
-// TODO: (spatel) Is there a way to continue using a contextKey without a circular import
-//                when referenced by consul/agent/server.go?
-
-// type contextKey struct {
-// 	name string
-// }
-
-// func (k *contextKey) String() string { return "agent/http context value " + k.name }
-
-// // RemoteAddrContextKey is a context key. It can be used in
-// // HTTP handlers with Context.Value to access the remote
-// // address the connection arrived on.
-// // The associated value will be of type net.Addr.
-// var RemoteAddrContextKey = &contextKey{"remote-addr"}
 
 // registerEndpoint registers a new endpoint, which should be done at package
 // init() time.
@@ -318,13 +299,13 @@ func (s *HTTPHandlers) handler(enableDebug bool) http.Handler {
 	return s.h
 }
 
-// Use key "remote-addr" to grab the net.Addr from the request's context.
+// Injects remote addr into the request's context
 func withRemoteAddrHandler(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(resp http.ResponseWriter, req *http.Request) {
 		addrPort, err := netip.ParseAddrPort(req.RemoteAddr)
 		if err == nil {
 			remoteAddr := net.TCPAddrFromAddrPort(addrPort)
-			ctx := context.WithValue(req.Context(), "remote-addr", remoteAddr)
+			ctx := consul.ContextWithRemoteAddr(req.Context(), remoteAddr)
 			req = req.WithContext(ctx)
 		}
 		next.ServeHTTP(resp, req)
--- a/agent/http.go
+++ b/agent/http.go
@@ -1,12 +1,14 @@
 package agent
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 	"net"
 	"net/http"
 	"net/http/pprof"
+	"net/netip"
 	"net/url"
 	"reflect"
 	"regexp"
@@ -103,6 +105,24 @@ var endpoints map[string]unboundEndpoint
 // allowedMethods is a map from endpoint prefix to supported HTTP methods.
 // An empty slice means an endpoint handles OPTIONS requests and MethodNotFound errors itself.
 var allowedMethods map[string][]string = make(map[string][]string)
+
+// contextKey is a value for use with context.WithValue. It's used as
+// a pointer so it fits in an interface{} without allocation.
+
+// TODO: (spatel) Is there a way to continue using a contextKey without a circular import
+//                when referenced by consul/agent/server.go?
+
+// type contextKey struct {
+// 	name string
+// }
+
+// func (k *contextKey) String() string { return "agent/http context value " + k.name }
+
+// // RemoteAddrContextKey is a context key. It can be used in
+// // HTTP handlers with Context.Value to access the remote
+// // address the connection arrived on.
+// // The associated value will be of type net.Addr.
+// var RemoteAddrContextKey = &contextKey{"remote-addr"}
 
 // registerEndpoint registers a new endpoint, which should be done at package
 // init() time.
@@ -288,12 +308,27 @@ func (s *HTTPHandlers) handler(enableDebug bool) http.Handler {
 	if s.agent.config.DisableHTTPUnprintableCharFilter {
 		h = mux
 	}
+
 	h = s.enterpriseHandler(h)
+	h = withRemoteAddrHandler(h)
 	s.h = &wrappedMux{
 		mux:     mux,
 		handler: h,
 	}
 	return s.h
+}
+
+// Use key "remote-addr" to grab the net.Addr from the request's context.
+func withRemoteAddrHandler(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(resp http.ResponseWriter, req *http.Request) {
+		addrPort, err := netip.ParseAddrPort(req.RemoteAddr)
+		if err == nil {
+			remoteAddr := net.TCPAddrFromAddrPort(addrPort)
+			ctx := context.WithValue(req.Context(), "remote-addr", remoteAddr)
+			req = req.WithContext(ctx)
+		}
+		next.ServeHTTP(resp, req)
+	})
 }
 
 // nodeName returns the node name of the agent
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ exit-code
 Thumbs.db
 .idea
 .vscode
+__debug_bin
 
 # MacOS
 .DS_Store
--- a/agent/acl_endpoint.go
+++ b/agent/acl_endpoint.go
@@ -38,7 +38,7 @@ func (s *HTTPHandlers) ACLBootstrap(resp http.ResponseWriter, req *http.Request)
 		Datacenter: s.agent.config.Datacenter,
 	}
 	var out structs.ACLToken
-	err := s.agent.RPC("ACL.BootstrapTokens", &args, &out)
+	err := s.agent.RPC(req.Context(), "ACL.BootstrapTokens", &args, &out)
 	if err != nil {
 		if strings.Contains(err.Error(), structs.ACLBootstrapNotAllowedErr.Error()) {
 			return nil, acl.PermissionDeniedError{Cause: err.Error()}
@@ -64,7 +64,7 @@ func (s *HTTPHandlers) ACLReplicationStatus(resp http.ResponseWriter, req *http.
 
 	// Make the request.
 	var out structs.ACLReplicationStatus
-	if err := s.agent.RPC("ACL.ReplicationStatus", &args, &out); err != nil {
+	if err := s.agent.RPC(req.Context(), "ACL.ReplicationStatus", &args, &out); err != nil {
 		return nil, err
 	}
 	return out, nil
@@ -89,7 +89,7 @@ func (s *HTTPHandlers) ACLPolicyList(resp http.ResponseWriter, req *http.Request
 
 	var out structs.ACLPolicyListResponse
 	defer setMeta(resp, &out.QueryMeta)
-	if err := s.agent.RPC("ACL.PolicyList", &args, &out); err != nil {
+	if err := s.agent.RPC(req.Context(), "ACL.PolicyList", &args, &out); err != nil {
 		return nil, err
 	}
 
@@ -150,7 +150,7 @@ func (s *HTTPHandlers) ACLPolicyRead(resp http.ResponseWriter, req *http.Request
 
 	var out structs.ACLPolicyResponse
 	defer setMeta(resp, &out.QueryMeta)
-	if err := s.agent.RPC("ACL.PolicyRead", &args, &out); err != nil {
+	if err := s.agent.RPC(req.Context(), "ACL.PolicyRead", &args, &out); err != nil {
 		return nil, err
 	}
 
@@ -219,7 +219,7 @@ func (s *HTTPHandlers) aclPolicyWriteInternal(_resp http.ResponseWriter, req *ht
 	}
 
 	var out structs.ACLPolicy
-	if err := s.agent.RPC("ACL.PolicySet", args, &out); err != nil {
+	if err := s.agent.RPC(req.Context(), "ACL.PolicySet", args, &out); err != nil {
 		return nil, err
 	}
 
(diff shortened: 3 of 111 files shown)


boxofrad:
Thanks for working on this @analogue 🙌🏻 

My only concern is that we'll end up forgetting to use the HTTP variant of the RPC method in places. I wonder if we should bite the bullet and change the `RPC` method's signature (and all call sites) in a way that makes this impossible.

For example, if we were to add a `context.Context` parameter we could have an HTTP middleware inject the IP address into the request's context. And in the non-HTTP paths, callers could pass `context.Background` or `context.TODO` as appropriate.

**Example:**

```Go
// HTTP Handler:
s.agent.RPC(req.Context(), &args, &out)

// Other code paths:
s.agent.RPC(context.Background(), &args, &out)
```

This is a fairly common pattern for threading request-scoped metadata like this.

analogue:
> My only concern is that we'll end up forgetting to use the HTTP variant of the RPC method in places. I wonder if we should bite the bullet and change the `RPC` method's signature (and all call sites) in a way that makes this impossible.
> 
> For example, if we were to add a `context.Context` parameter we could have an HTTP middleware inject the IP address into the request's context. And in the non-HTTP paths, callers could pass `context.Background` or `context.TODO` as appropriate.
> 
> **Example:**
> 
> ```go
> // HTTP Handler:
> s.agent.RPC(req.Context(), &args, &out)
> 
> // Other code paths:
> s.agent.RPC(context.Background(), &args, &out)
> ```
> 
> This is a fairly common pattern for threading request-scoped metadata like this.

Great call out! I'm in support of doing this (it came up in my discussion with @banks) but just want to make sure I surface the extent of the changes across the code base. I believe there are multiple non-test interfaces that include the `RPC(..)` method under consideration with the exact same signature:

```Go
Search: ^\sRPC\(method

7 results - 7 files

agent/agent.go:
  141 type delegate interface {
  190  
  191: 	RPC(method string, args interface{}, reply interface{}) error
  192  

agent/cache-types/rpc.go:
  8  type RPC interface {
  9: 	RPC(method string, args interface{}, reply interface{}) error
  10  }

agent/checks/check.go:
  50  type RPC interface {
  51: 	RPC(method string, args interface{}, reply interface{}) error
  52  }

agent/connect/testing_ca.go:
  416  type TestAgentRPC interface {
  417: 	RPC(method string, args interface{}, reply interface{}) error
  418  }

agent/consul/acl.go:
  136  	// TODO: separate methods for each RPC call (there are 4)
  137: 	RPC(method string, args interface{}, reply interface{}) error
  138  	EnterpriseACLResolverDelegate

agent/local/state.go:
  151  type rpc interface {
  152: 	RPC(method string, args interface{}, reply interface{}) error
  153  	ResolveTokenAndDefaultMeta(token string, entMeta *acl.EnterpriseMeta, authzContext *acl.AuthorizerContext) (resolver.Result, error)

agent/rpcclient/health/health.go:
  25  type NetRPC interface {
  26: 	RPC(method string, args interface{}, reply interface{}) error
  27  }

```

This PR is targeted at the first occurrence above in `agent.go` which is part of the `delegate` interface. I got the impression in my chat with @banks that you can't only change the single interface, but all of them need to be updated (IIRC, something related to circular dependencies which resulted in the interface method being duplicated).  Given that understanding, we're looking at four separate non-test implementations:

```Go
^func .+ RPC\(method

4 results - 4 files

agent/agent.go:
  1562  // This allows the agent to implement the Consul.Interface
  1563: func (a *Agent) RPC(method string, args interface{}, reply interface{}) error {
  1564  	method = a.translateMethodWhenOverrides(method)

agent/cache-types/mock_RPC.go:
  16  // RPC provides a mock function with given fields: method, args, reply
  17: func (_m *MockRPC) RPC(method string, args interface{}, reply interface{}) error {
  18  	ret := _m.Called(method, args, reply)

agent/consul/client.go:
  264  // RPC is used to forward an RPC call to a consul server, or fail if no servers
  265: func (c *Client) RPC(method string, args interface{}, reply interface{}) error {
  266  	// This is subtle but we start measuring the time on the client side

agent/consul/server.go:
  1537  // RPC is used to make a local RPC call
  1538: func (s *Server) RPC(method string, args interface{}, reply interface{}) error {
  1539  	codec := &inmemCodec{

```

and **166** call sites in non-test code and **459** call sites in test code (`*_test.go`).

Does this make sense? I'd like to build consensus on the approach here given the invasive nature of the change and possible unknown/unintended consequences. @dhiaayachi @banks @jmurret Thoughts?

dhiaayachi:
Thank you for putting this together and starting the discussion @analogue.

My understanding is that the inMemCodec is only used for internal calls. Nothing will actually go on the wire?
Would it be sufficient to keep the local IP in the [Server struct](http://github.com/hashicorp/consul/blob/7a759be2c578dfc81849e9ee1aaecc8a4f546c07/agent/consul/server.go#L146-L146) and pass it to the InMemCodec [at the call site](http://github.com/hashicorp/consul/blob/7a759be2c578dfc81849e9ee1aaecc8a4f546c07/agent/consul/server.go#L1510-L1510)?

boxofrad:
Thanks for raising that @analogue.

In my opinion, it's worth making the change because keeping the method signature the same in all contexts reduces the likelihood of programmer errors. In other words, passing contexts around is so commonplace in Go code and means you don't have to think about passing the IP at each call-site, you can just rely on the HTTP middleware (or whatever) to do the right thing.

We should automate the changes if possible (some `sed` magic should do the trick 🪄).

analogue:
> My understanding is that the inMemCodec is only used for internal calls. Nothing will actually go on the wire? Would it be sufficient to keep the local IP in the [Server struct](https://github.com/hashicorp/consul/blob/7a759be2c578dfc81849e9ee1aaecc8a4f546c07/agent/consul/server.go#L146-L146) and pass it to the InMemCodec [at the call site](https://github.com/hashicorp/consul/blob/7a759be2c578dfc81849e9ee1aaecc8a4f546c07/agent/consul/server.go#L1510-L1510)?

From an end-to-end perspective given my understanding of this scenario, the RPC call to the server agent satisfies a request from an external entity. In this case, the entrypoint into the server agent is via an HTTP request. The HTTP handler translates this request into a local RPC call with the help of `inmemCodec` in order to bypass a trip through the networking layer. Therefore, we're tracking the `sourceAddr` (aka `remoteAddr`) of the original HTTP request which is a per-request bit of information and not something that is tied to the `Server` struct whose scope spans any [number of requests](https://github.com/hashicorp/consul/blob/7a759be2c578dfc81849e9ee1aaecc8a4f546c07/agent/consul/server.go#L276). This seems to indicate that `localIP` is not relevant here.

agent/http.go
```diff
@@ -104,6 +106,24 @@ var endpoints map[string]unboundEndpoint
...
+// }
+
+// func (k *contextKey) String() string { return "agent/http context value " + k.name }
+
+// // RemoteAddrContextKey is a context key. It can be used in
+// // HTTP handlers with Context.Value to access the remote
+// // address the connection arrived on.
+// // The associated value will be of type net.Addr.
+// var RemoteAddrContextKey = &contextKey{"remote-addr"}
+
```
analogue, line 126: The context key is currently a string `"remote-addr"`. Golang docs recommend against used a string to prevent name collisions. However, using a type results in a circular dep (See comment on lines 112-113). Anything obvious I'm missing here or will a string suffice for now?

analogue:
Folks, I've updated the PR based on refactoring the RPC method signature to include a context. I revised the original description and would appreciate if you could take another look.

dhiaayachi:
> > My understanding is that the inMemCodec is only used for internal calls. Nothing will actually go on the wire? Would it be sufficient to keep the local IP in the [Server struct](https://github.com/hashicorp/consul/blob/7a759be2c578dfc81849e9ee1aaecc8a4f546c07/agent/consul/server.go#L146-L146) and pass it to the InMemCodec [at the call site](https://github.com/hashicorp/consul/blob/7a759be2c578dfc81849e9ee1aaecc8a4f546c07/agent/consul/server.go#L1510-L1510)?
> 
> From an end-to-end perspective given my understanding of this scenario, the RPC call to the server agent satisfies a request from an external entity. In this case, the entrypoint into the server agent is via an HTTP request. The HTTP handler translates this request into a local RPC call with the help of `inmemCodec` in order to bypass a trip through the networking layer. Therefore, we're tracking the `sourceAddr` (aka `remoteAddr`) of the original HTTP request which is a per-request bit of information and not something that is tied to the `Server` struct whose scope spans any [number of requests](https://github.com/hashicorp/consul/blob/7a759be2c578dfc81849e9ee1aaecc8a4f546c07/agent/consul/server.go#L276). This seems to indicate that `localIP` is not relevant here.

Thank you for the context @analogue.

I think something was lost in translation between the RFC, task creation...but this is not relevant for phase 1, which is we are focusing on right now, as for phase 1 we don't require the source IP and we rate limit all requests independently from the origin of it. What is important for now is to make sure that the InMemCodec Path is still going through the interceptors, which I think we confirmed already.

I suggest to let this PR aside to when we get to phase 2.

agent/http.go
boxofrad: I'd recommend creating helper methods in a package that both `agent` and `agent/consul` can import without cycles. Maybe even in `agent/consul` itself:

```Go
package consul

type contextKeyRemoteAddr struct{}

func ContextWithRemoteAddr(ctx context.Context, addr net.Addr) context.Context {
	return context.WithValue(ctx, contextKeyRemoteAddr{}, addr)
}

func RemoteAddrFromContext(ctx context.Context) (net.Addr, bool) {
	v, ok := ctx.Value(contextKeyRemoteAddr{})
	if !ok {
		return nil, false
	}
	return v.(net.Addr), true
}
```

```Go
func withRemoteAddrHandler(next http.Handler) http.Handler {
	return http.HandlerFunc(func(resp http.ResponseWriter, req *http.Request) {
		addrPort, err := netip.ParseAddrPort(req.RemoteAddr)
		if err == nil {
			remoteAddr := net.TCPAddrFromAddrPort(addrPort)
			ctx := consul.ContextWithRemoteAddr(req.Context(), remoteAddr)
			req = req.WithContext(ctx)
		}
		next.ServeHTTP(resp, req)
	})
}
```

banks:
@dhiaayachi that was my first response when I spoke to Semir about this task too, but the JIRA (not sure who added that) seemed to be talking more specifically about this being desired. I eventually convinced myself that it probably _is_ the right thing to do (even for phase1) to not be surprising to end-users who configured IP-based limits but would love to hear other's thoughts on that - I was also assuming the JIRA author had already decided to do this!

This isn't so much about generically handling HTTP client IPs but rather making sure the (very common) case of some workloads hitting the server's HTTP API directly instead of using a local client agent is also covered in the basic rate limiting. In cases where that pattern is used, it's kind of an implementation detail to operators that the servers are seeing that as HTTP rather than RPC load, and I guess they'd assume that rate limiting applied to all traffic regardless of whether it's going via a client agent or not.

We already discourage that practice in general and in our documentation, but I think many if not most of our largest users who would be likely to benefit from rate limiting like this do actually still have workloads that hit the server's HTTP API directly. It may be surprising to them that server-side rate limiting (by IP) doesn't apply to those requests.

In the end I did assume though that whoever write the JIRA had made a conscious decision that that use-case was important to support and I could understand why.

So I guess I'm interested in whether it's worth including this now it seems mostly done anyway for a more complete solution in phase 1, or whether there are concerns that about this approach or the UX that weren't covered in the RFC that make it more risky to merge this? If it's going to take a lot longer to resolve those things I understand deferring until the stuff the RFC did commit to is done, but if this is nearly done anyway it _does_ seem like something we'd need for a "complete" solution doesn't it?

dhiaayachi:
> @dhiaayachi that was my first response when I spoke to Semir about this task too, but the JIRA (not sure who added that) seemed to be talking more specifically about this being desired. I eventually convinced myself that it probably _is_ the right thing to do (even for phase1) to not be surprising to end-users who configured IP-based limits but would love to hear other's thoughts on that - I was also assuming the JIRA author had already decided to do this!
> 
> This isn't so much about generically handling HTTP client IPs but rather making sure the (very common) case of some workloads hitting the server's HTTP API directly instead of using a local client agent is also covered in the basic rate limiting. In cases where that pattern is used, it's kind of an implementation detail to operators that the servers are seeing that as HTTP rather than RPC load, and I guess they'd assume that rate limiting applied to all traffic regardless of whether it's going via a client agent or not.
> 
> We already discourage that practice in general and in our documentation, but I think many if not most of our largest users who would be likely to benefit from rate limiting like this do actually still have workloads that hit the server's HTTP API directly. It may be surprising to them that server-side rate limiting (by IP) doesn't apply to those requests.
> 
> In the end I did assume though that whoever write the JIRA had made a conscious decision that that use-case was important to support and I could understand why.
> 
> So I guess I'm interested in whether it's worth including this now it seems mostly done anyway for a more complete solution in phase 1, or whether there are concerns that about this approach or the UX that weren't covered in the RFC that make it more risky to merge this? If it's going to take a lot longer to resolve those things I understand deferring until the stuff the RFC did commit to is done, but if this is nearly done anyway it _does_ seem like something we'd need for a "complete" solution doesn't it?

> 

@banks I was not trying to make a case for not including the rate limiting of local http requests in phase 1. It's part of phase 1. But because phase 1 focus on global rate limiting we don't really need the source IP to rate limit requests as all the requests to a server are rate limited using the same rate limiter (1 for read and 1 for write). That said, the IP would be good to have to add to the log but was not mandatory for that phase.

The JIRA ticket originated from a discussion we had around the possibility  that when we use inMemCodec we could bypass the RPC interceptor for rate limiting that was recently added by @analogue and I think when the Jira ticket was added the scope was assumed to include the source IP.

That said, I agree, looking to the current state of the PR I think it's good to get it merged as tabling it would add more work to update it at the end.

dhiaayachi (approved):
Thank you @analogue for getting this through!

I second @boxofrad comment around using helper functions to be able to define a custom type specially that we are passing the http request context in some parts of the code.

jmurret:
@banks @dhiaayachi I added the task based on RB's mention of this scenario.  I can't say that I fully understood it to the point of "Should this be ready in Phase 1 or Phase 2?".  I viewed RB's description as "Here is a case where any rate limiting may not be processed properly."  So, I added it to the current wiring up of interceptors with the idea that we want to ensure "Everything flows through rate limiting".  

If this is really only seen in the source ip rate limiting, then we can move it to that phase.  We also need to consider that @analogue is also doing support and we are trying to get him tasks that are not blocking and not on the critical path, so we have to consider that some of the work that he is picking up may not be the next item in priority.

agent/http.go
analogue: I like it! Done.

analogue:
Rebased from main to fix some merge conflicts. New commits from review feedback and unit tests worth a 👀 :
- [Helper methods for getting/putting remoteAddr in request context](https://github.com/hashicorp/consul/pull/15700/commits/4a0e7535a4ad9d49b4b16112577f2773cbd0d0e0)
[4a0e753](https://github.com/hashicorp/consul/pull/15700/commits/4a0e7535a4ad9d49b4b16112577f2773cbd0d0e0)
- [Unit test for withRemoteAddrHandler(..)](https://github.com/hashicorp/consul/pull/15700/commits/b495689d8a4145636a9815e0fe973368f66d26a6)
[b495689](https://github.com/hashicorp/consul/pull/15700/commits/b495689d8a4145636a9815e0fe973368f66d26a6)

jmurret:
Nice refactor to get it under tests.  🔥 🙏

PR #15706, opened 2022-12-07
update go version to 1.18.9

dhiaayachi:
### Description
update go version to `1.18.9`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -134,8 +134,6 @@ commands:
     steps:
       - attach_workspace:
           at: /home/circleci/go/bin
-      - run: sudo apt-get update --allow-releaseinfo-change-suite --allow-releaseinfo-change-version && sudo apt-get install -y rsyslog
-      - run: sudo service rsyslog start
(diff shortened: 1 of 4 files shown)


eculver:
We don't need the `backport/1.13.x` label since this is going directly against `release/1.13.x`.

dhiaayachi:
> We don't need the `backport/1.13.x` label since this is going directly against `release/1.13.x`.

right 😆 !
I'm hitting a bump with this one. Something wrong with the ci image?

.circleci/config.yml
```diff
@@ -208,6 +208,8 @@ jobs:
       - checkout
       - run: go env
       - run:
+          environment: 
+           GOARCH: ""
           name: Install golangci-lint
```
rboyer, line 213: After this targeted fix we should get this on main etc.

rboyer (approved):

.circleci/config.yml
dhiaayachi: yes

PR #15710, opened 2022-12-07
Remove the `connect.enable_serverless_plugin` agent configuration option

erichaberkorn:
### Description
We are moving Lambda support to an Envoy extension so this flag is no longer needed.

### PR Checklist

* [X] updated test coverage
* [X] not a security concern


agent/config/config.go
```diff
@@ -613,7 +613,6 @@ type Connect struct {
 	CAProvider                      *string                `mapstructure:"ca_provider" json:"ca_provider,omitempty"`
 	CAConfig                        map[string]interface{} `mapstructure:"ca_config" json:"ca_config,omitempty"`
 	MeshGatewayWANFederationEnabled *bool                  `mapstructure:"enable_mesh_gateway_wan_federation" json:"enable_mesh_gateway_wan_federation,omitempty"`
-	EnableServerlessPlugin          *bool                  `mapstructure:"enable_serverless_plugin" json:"enable_serverless_plugin,omitempty"`
```
rboyer, line 616: These should move into https://github.com/hashicorp/consul/blob/main/agent/config/deprecated.go so it can emit warnings.

rboyer (approved):

PR #15727, opened 2022-12-08
upgrade go version to 1.18 and build version to 1.18.9

dhiaayachi:
### Description
This attempt to upgrade go version from 1.17, which is out of support to 1.18


.changelog/15727.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:note
+dependency: Upgrade to use Go 1.18.9.
+```
```
DanStough, line 3: I don't really have a preference but the last one I remember doing was formatted like this: https://github.com/hashicorp/consul/blob/main/.changelog/12808.txt#L1-L4

DanStough (approved):

PR #15737, opened 2022-12-08
[OSS] security: update x/net module

DanStough:
### Description
Updating x/net package in Consul to resolve [CVE-2022-41717](https://github.com/golang/go/issues/56350).

The api module has an indirect dependency on x/net, but using go mod why -m golang.org/x/net leads me to believe that the downstream package github.com/miekg/dns doesn't use x/net/http2.

No manual tests were performed.


dhiaayachi (approved):
LGTM!

PR #15742, opened 2022-12-08
test: fix flaky TestStreamResources_Server_DisconnectsOnHeartbeatTimeout test

rboyer:
### Description

This test previously was destined for flakiness given how time was used in the comparison. Now it is a little looser in the expectation so it should be far less likely to flake.

Additional changes:
- fixed typo in logging key
- (controversial?) when clearing a `*time.Time` value `nil` is now used instead of `&time.Time{}`. The code and test logic seemed to already be handling `nil` times specially so I switched it back which simplifies some things.

### DIFF
--- a/agent/grpc-external/services/peerstream/stream_test.go
+++ b/agent/grpc-external/services/peerstream/stream_test.go
@@ -1215,8 +1215,10 @@ func TestStreamResources_Server_DisconnectsOnHeartbeatTimeout(t *testing.T) {
 		base: time.Date(2000, time.January, 1, 0, 0, 0, 0, time.UTC),
 	}
 
+	const heartbeatTimeout = 50 * time.Millisecond
+
 	srv, store := newTestServer(t, func(c *Config) {
-		c.incomingHeartbeatTimeout = 50 * time.Millisecond
+		c.incomingHeartbeatTimeout = heartbeatTimeout
 	})
 	srv.Tracker.setClock(it.Now)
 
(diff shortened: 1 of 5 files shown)


rboyer:
Would this need to go to `1.13` too?

agent/grpc-external/services/peerstream/stream_test.go
```diff
@@ -1230,22 +1232,32 @@ func TestStreamResources_Server_DisconnectsOnHeartbeatTimeout(t *testing.T) {
...
+	var lastAck time.Time
 	testutil.RunStep(t, "new stream gets tracked", func(t *testing.T) {
 		retry.Run(t, func(r *retry.R) {
 			status, ok := srv.StreamStatus(testPeerID)
 			require.True(r, ok)
 			require.True(r, status.Connected)
+			require.NotNil(r, status.LastAck)
+			lastAck = *status.LastAck
 		})
 	})
```
rboyer, line 1244: A mysterious 50ms after the last ACK the disconnect time is captured by advancing the lockstep clock abstraction. If the runner was just slow enough then the disconnect time is advanced and recorded BEFORE the `stream is disconnected` subtest case is entered and it captures the wrong clock value for comparison (the NEXT one).

Now we capture the ACK value and assert that the disconnect time is just some time after it.

agent/grpc-external/services/peerstream/stream_tracker.go
```diff
@@ -156,7 +156,7 @@ func (t *Tracker) IsHealthy(s Status) bool {
 	// If last Nack is after last Ack, it means the peer is unable to
 	// handle our replication message
 	if s.LastAck == nil {
```
hashi-derek, line 158: Are these if-statements even required now that we set the value to nil? Also, I think there will be a potential nil-pointer panic just a couple lines below due to this logic change.

PR #15757, opened 2022-12-09
EventPublisher subscriptions for Consul Native API Gateway

t-eckert:
### Description

As part of Consul-Native API Gateway, 5 new config entries are being added to Consul:
- API Gateway
- Inline Certificate
- HTTP Route
- TCP Route
- Bound API Gateway (read-only)

These configs will be handled by the new eventually-consistent model of state storage for Consul using `EventPublisher`. This PR adds the structs and related behavior that will allow `EventPublisher` to subscribe to changes in these config entries.

* Added unit tests

### Links

* [Consul-Native API Gateway](https://hashidocs.hashicorp.services/document/1LRrtUpLcG2ImI_rYpdUrx9FH_63nbX25aRdzwoCwwzU)

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/consul/state/config_entry_events.go
```diff
@@ -117,6 +122,36 @@ func (s *Store) ServiceDefaultsSnapshot(req stream.SubscribeRequest, buf stream.
...
+}
+
+// InlineCertificateSnapshot is a stream.SnapshotFunc that returns a snapshot of
+// inline-certificate config entries.
+func (s *Store) InlineCertificateSnapshot(req stream.SubscribeRequest, buf stream.SnapshotAppender) (uint64, error) {
+	return s.configEntrySnapshot(structs.InlineCertificate, req, buf)
+}
+
+// BoundAPIGatewaySnapshot is a stream.SnapshotFunc that returns a snapshot of
+// bound-api-gateways config entries.
```
nathancoleman, line 150:
```suggestion
// bound-api-gateway config entries.
```

agent/consul/state/config_entry_events_test.go
```diff
@@ -547,3 +547,348 @@ func TestServiceDefaultsSnapshot(t *testing.T) {
 		})
 	}
 }
+
+func TestAPIGatewaySnapshot(t *testing.T) {
```
nathancoleman, line 551: 🧠  Seems like most of the test code that isn't directly defining `testCases` is duplicated. I'd probably personally just do one big table test with all of the snapshot types, but I see that this code is matching patterns established in this file already.

agent/consul/state/events_test.go
```diff
@@ -0,0 +1,148 @@
...
+			err:                      fmt.Errorf("cannot construct subject for topic 99999"),
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			actual, err := PBToStreamSubscribeRequest(tc.req, tc.entMeta)
+
+			if tc.err != nil {
+				require.Error(t, err)
```
nathancoleman, line 140: Is the intention here to see if the error matches? If so, we could
```go
require.EqualError(t, err, tc.err.Error())
```

nathancoleman (approved):
Just a few nitpicks. Approving assuming CI passes.

Nice work on this! 🎉

agent/consul/state/config_entry_events_test.go
t-eckert: I see what you're saying. I think I'll keep it as-is because it follows the existing pattern. But I would see the value in refactoring this file down the road.

agent/consul/state/events_test.go
t-eckert: Awesome! I even caught a bug in my test with this change. Thanks!

PR #15760, opened 2022-12-09
Fix DialedDirectly configuration for Consul dataplane.

hashi-derek:
The `DialedDirectly` configuration on upstreams is ignored by Consul Dataplane, because it selects data through the streaming backend which does not support the `MergeCentralConfig` flag. This means that any service-defaults or proxy-defaults configurations on upstreams were ignored.

This PR adds the `DialedDirectly` config into the compiled discovery chain targets, so that the "merged" version is available to the proxy config.

This PR also fixes an existing issue where passthrough clusters were not properly cleaned up whenever the `DialedDirectly` field is toggled on and off during runtime.

### DIFF
--- a/agent/proxycfg/upstreams.go
+++ b/agent/proxycfg/upstreams.go
@@ -190,10 +190,12 @@ func (s *handlerUpstreams) handleUpdateUpstreams(ctx context.Context, u UpdateEv
 			upstreamsSnapshot.PassthroughIndices[addr] = indexedTarget{idx: csnIdx, upstreamID: uid, targetID: targetID}
 			passthroughs[addr] = struct{}{}
 		}
+		// Always clear out the existing target passthroughs list so that clusters are cleaned up
+		// correctly if no entries are populated.
+		upstreamsSnapshot.PassthroughUpstreams[uid] = make(map[string]map[string]struct{})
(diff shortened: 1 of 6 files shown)


agent/proxycfg/upstreams.go
```diff
@@ -148,7 +149,17 @@ func (s *handlerUpstreams) handleUpdateUpstreams(ctx context.Context, u UpdateEv
 		passthroughs := make(map[string]struct{})
 
 		for _, node := range resp.Nodes {
-			if !node.Service.Proxy.TransparentProxy.DialedDirectly {
+			dialedDirectly := node.Service.Proxy.TransparentProxy.DialedDirectly
```
freddygv, line 152: Can you update this proxycfg test case to validate the change?
https://github.com/hashicorp/consul/blob/main/agent/proxycfg/state_test.go#LL2219

Could update the call to `TestCompileConfigEntries` to include a service-defaults with `DialedDirectly` set to true, and flip one of the CheckServiceNodes to set it to false. That way we check that we still get both addresses as passthrough upstreams.

freddygv (commented):
This is looking good, left a few asks below

erichaberkorn (approved):

PR #15767, opened 2022-12-12
Server side rate limiter: handle the race condition for limiters tree write in multilimiter

dhiaayachi:
### Description
The first version of MultiLimiter have a race related to writing the limiter tree in multiple go routine. That would have the following side effects:

- A written limiter can be overridden by the limiters reconcile loop
- An updated limiter config could be overridden by the limiters writing loop

To solve those issues this PR unify all the limiters writing into a single loop that would write new limiters, reconcile configs and remove old limiters.


agent/consul/multilimiter/multilimiter.go
```diff
@@ -267,8 +234,27 @@ func (m *MultiLimiter) reconcileLimitedOnce(now time.Time, reconcileCheckLimit t
...
+				lastAccessT := time.UnixMilli(lastAccess)
+				diff := now.Sub(lastAccessT)
+
+				if diff > reconcileCheckLimit {
+					txn.Delete(k)
+				}
+			}
+		}
+		k, v, ok = iter.Next()
+	}
```
boxofrad, line 257: Personal style (please feel free to ignore):

I tend to find that handling all of the type-cast failures and `nil` checking up-front (and reducing nesting) makes this kind of thing easier to read. 

```suggestion
	// remove all expired limiters
	for k, v, ok := iter.Next(); ok; k, v, ok = iter.Next() {
		t, isLimiter := v.(*Limiter)
		if !isLimiter || t.limiter == nil {
			continue
		}

		lastAccess := time.UnixMilli(t.lastAccess.Load())
		if now.Sub(lastAccess) > reconcileCheckLimit {
			txn.Delete(k)
		}
	}

```

boxofrad: Similarly, we could reduce the nesting in `reconcileConfig` too:

```Go
for k, v, ok := iter.Next(); ok; k, v, ok = iter.Next() {
	lim, isLimiter := v.(*Limiter)
	if !isLimiter || lim.limiter == nil {
		continue
	}

	// Grab the config for the limiter's prefix.
	prefix, _ := splitKey(k)
	val, exists := m.limitersConfigs.Load().Get(prefix)
	if !exists {
		continue
	}
	config, isConfig := val.(*LimiterConfig)
	if !isConfig || config == nil {
		continue
	}

	// Apply it to the limiter.
	if config.isApplied(lim.limiter) {
		continue
	}
	limiter := Limiter{limiter: rate.NewLimiter(config.Rate, config.Burst)}
	limiter.lastAccess.Store(lim.lastAccess.Load())
	txn.Insert(k, &limiter)
}
```

boxofrad (approved):
Looks great! ⭐ 

I've left a couple of personal style preference notes, but please feel free to ignore them.

Also, I wonder if we need to worry about `limiterCh` filling up now that it'll be blocked while we reconcile config and delete old limiters?

PR #15769, opened 2022-12-12
agent: Fix assignment of error when auto-reloading cert and key file changes.

jmurret:
### Description
For both the key and cert file watcher code, a call to `a.configFileWatcher.Replace()` is made but the error returned from this called is not used...but a check is right below it like this:
```
				a.configFileWatcher.Replace(f.oldCfg.KeyFile, f.newCfg.KeyFile)
				if err != nil {
					return err
				}
```
This seems to be an omission rather than strategic.  In its current case, this will either check on a previous error, or will never get triggered because a previous check would have exited.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/.changelog/15769.txt
+++ b/.changelog/15769.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+agent: Fix assignment of error when auto-reloading cert and key file changes.
+```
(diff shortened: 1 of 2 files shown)


dhiaayachi (approved):
Thank you for extracting this @jmurret, LGTM!

Do you think we should backport this to 1.14,1.13 and 1.12? I think it's self contained enough to be considered.

jmurret:
@dhiaayachi Agree.  Will backport.

Issue #15753, opened 2022-12-09
connect: ResourceExhausted desc = this server has too many xDS streams open, please try another


Poweranimal:
#### Overview of the Issue

Running a consul cluster (no consul clients) with approx. 20-25 connect services may cause the following error in consul servers:
```text
2022-12-09T15:54:57.782Z [ERROR] agent.envoy: Error handling ADS delta stream: xdsVersion=v3 error="rpc error: code = ResourceExhausted desc = this server has too many xDS streams open, please try another"
```
In this error state, none of my consul connect proxies can retrieve its envoy route configurations from consul server.

Is there a way to increase the limit of open xDS streams in consul server?

### Consul info for Server

<details>
  <summary>Client info</summary>

```
agent:
        check_monitors = 0
        check_ttls = 1
        checks = 1
        services = 1
build:
        prerelease = 
        revision = 0ba7a401
        version = 1.14.2
        version_metadata = 
consul:
        acl = enabled
        bootstrap = false
        known_datacenters = 1
        leader = false
        leader_addr = 10.125.75.23:8300
        server = true
raft:
        applied_index = 20352106
        commit_index = 20352106
        fsm_pending = 0
        last_contact = 45.516941ms
        last_log_index = 20352106
        last_log_term = 24822
        last_snapshot_index = 20336168
        last_snapshot_term = 24822
        latest_configuration = [{Suffrage:Voter ID:eba86246-584b-e10e-5e8e-ebef14fabb65 Address:10.125.75.23:8300} {Suffrage:Voter ID:d7068e33-24b8-cd37-2904-ea853fec1ce2 Address:10.125.52.24:8300} {Suffrage:Voter ID:c9f29f3d-1999-f570-68dc-7cbd0dee050c Address:10.125.88.35:8300}]
        latest_configuration_index = 0
        num_peers = 2
        protocol_version = 3
        protocol_version_max = 3
        protocol_version_min = 0
        snapshot_version_max = 1
        snapshot_version_min = 0
        state = Follower
        term = 24822
runtime:
        arch = amd64
        cpu_count = 8
        goroutines = 1227
        max_procs = 8
        os = linux
        version = go1.19.2
serf_lan:
        coordinate_resets = 0
        encrypted = true
        event_queue = 0
        event_time = 6948
        failed = 0
        health_score = 0
        intent_queue = 0
        left = 0
        member_time = 39678
        members = 3
        query_queue = 0
        query_time = 1
serf_wan:
[...]

PR #15789, opened 2022-12-14
xds: don't attempt to load-balance sessions for local proxies

boxofrad:
### Description
Previously, we'd begin a session with the xDS concurrency limiter regardless of whether the proxy was registered in the catalog or in the server's local agent state.

This caused problems for users who run `consul connect envoy` directly against a server rather than a client agent, as the server's locally registered proxies wouldn't be included in the limiter's capacity.

Now, the `ConfigSource` is responsible for beginning the session and we only do so for services in the catalog (bypassing the limiter for locally-registered proxies).

Fixes: https://github.com/hashicorp/consul/issues/15753

Also upgrades mockery to v2.15.0, as this was required to regenerate the mocks on Go 1.19.

* Run a Consul cluster with >1 server.
* Register the majority of the proxies to 1 server.
* Run `consul connect envoy` for each proxy.
* Previously you'd see "this server has too many xDS streams open" in the Envoy logs.
* Now it should work as expected.


agent/proxycfg-sources/catalog/config_source.go
```diff
@@ -82,11 +95,11 @@ func (m *ConfigSource) Watch(serviceID structs.ServiceID, nodeName string, token
 		if err := m.startSync(w.closeCh, proxyID); err != nil {
 			delete(m.watches, proxyID)
 			cancelWatch()
```
rboyer, line 97: Do you need to `session.End()` here?

boxofrad: Good catch!

boxofrad:
Hey @rboyer 👋🏻 

I've fixed CI and the missing `End` call you spotted. Would you mind giving this another 👀 please?

PR #15794, opened 2022-12-14
Error out `consul connect envoy` if agent explicitly disabled grpc

kisunji:
### Description

### Links

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


command/connect/envoy/envoy.go
```diff
@@ -659,7 +659,7 @@ func (c *cmd) xdsAddress() (GRPC, error) {
 	if addr == "" {
 		port, protocol, err := c.lookupXDSPort()
 		if err != nil {
-			c.UI.Error(fmt.Sprintf("Error connecting to Consul agent: %s", err))
+			return g, fmt.Errorf("Could not look up xDS port: %w", err)
```
kisunji, line 662: I'm unsure if the return was missed here or we intentionally went on to default to `8502`. It looks safe to error here since if the call to `Agent.Self` doesn't yield any port, we probably don't want the command to continue.

command/connect/envoy/envoy_test.go
```diff
@@ -835,32 +859,8 @@ func TestGenerateConfig(t *testing.T) {
...
-				AdminBindPort:         "19000",
-				LocalAgentClusterName: xds.LocalAgentClusterName,
-				PrometheusScrapePath:  "/metrics",
-			},
-		},
-		{
-			Name:  "CONSUL_HTTP_ADDR-with-https-scheme-does-not-affect-grpc-tls",
-			Flags: []string{"-proxy-id", "test-proxy", "-ca-file", "../../../test/ca/root.cer"},
-			Env:   []string{"CONSUL_HTTP_ADDR=https://127.0.0.1:8500"},
-			WantArgs: BootstrapTplArgs{
```
kisunji, line 851: This diff looks bizzare because I removed an unnecessary testcase and renamed some tests at the same time.

command/connect/envoy/envoy.go
rboyer: It can't error out because typically `agent:read` is not granted to sidecars, which is required for `/v1/agent/self`.  The existing code did the agent self lookup as an _UX optimization_.

Instead you could pick whichever port we already calculated and do a super simple `net.Dial` with a short timeout and see if the port is even open instead, which wouldn't even need ACL tokens.

hashi-derek: Yeah, I'm not sure why it never errored-out before. I assume it was to enable some backwards compatibility with old APIs that didn't include the port. I don't honestly know the history on this particular area of code, aside from the recent changes I made. I'll have to think about it quite a bit, because there might be some situation where the API is not available, but gRPC is.

hashi-derek: Nevermind. RB already answered the question.

kisunji: @rboyer I ended up logging and defaulting only if it's an ACL perm denied and otherwise erroring out.

pglass (approved):
👍

PR #15801, opened 2022-12-14
1.12.8 updates

malizz:
Update Changelog and bump version


david-yu:
@malizz I believe we should add this in 1.12, 1.13, and 1.14 as this was a missing changelog: https://github.com/hashicorp/consul/pull/15777/files

DanStough (approved):

david-yu (approved):

PR #15805, opened 2022-12-15
Rate limiting handler - ensure configuration has changed before modifying limiters

jmurret:
### Description
We want to ensure that rate limiting config has changed before we make possibly expensive updates to existing rate limiters.  If nothing has changed, then it is a noop.  If something has changed the multilimiter config and radix tree will get updated.

This change was made in the handler so that we could a mock rate limiter to ensure that the calls were being made or not being made.

### Not in this PR
Currently this PR does not call UpdateConfig on the rate limiter if the Mode has changed.  There is another task that deals entirely with ensuring Disabled mode means that the rate limiters are not on the code path.  That task will handle changes to Mode.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/consul/rate/handler_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"testing"
+
+	"github.com/hashicorp/consul/agent/consul/multilimiter"
+	mock "github.com/stretchr/testify/mock"
+)
+
+func TestNewHandlerWithLimiter_CallsUpdateConfig(t *testing.T) {
+	mockRateLimiter := multilimiter.NewMockRateLimiter(t)
+	mockRateLimiter.On("UpdateConfig", mock.Anything, mock.Anything).Return(
+		func(c multilimiter.LimiterConfig, prefix []byte) {})
```
boxofrad, line 13: I don't think `UpdateConfig` returns anything. Would this do the trick?

```suggestion
	mockRateLimiter.On("UpdateConfig", mock.Anything, mock.Anything).Return()
```

boxofrad (approved):
LGTM ⭐ 

For what it's worth, the multilimiter [compares the config](https://github.com/hashicorp/consul/blob/20f9e606b2c73feadf501a3af24a0dde72dc2f09/agent/consul/multilimiter/multilimiter.go#L232-L233) before re-creating the limiter. But it can't hurt to be belt and braces 😄

jmurret:
Thank you @boxofrad for the review!  This task came out of [this PR conversation](https://github.com/hashicorp/consul/pull/15531#discussion_r1041098852).  So it looks like in between the limiter change got added.  Is there any value to this PR then?  I could see it potentially saving some tree traversal and also it is explicitly called in a test.  Also, it will set the tone for future rate limiting phases to have implement the same logic here...I'm not sure if that's a drag or possibly a good thing.  🤷 Thoughts?

agent/consul/rate/handler_test.go
jmurret: Thank you. I updated this as well as another mock instance in the file.

jmurret:
@dhiaayachi do you have thoughts on whether this PR has value given change checking already does occur in the multilimiter?  (If this PR is overkill and unnecessary, I have no problems closing and moving on to the next thing.)

agent/consul/rate/handler.go
```diff
@@ -172,9 +178,15 @@ func (h *Handler) Allow(op Operation) error {
...
 	h.cfg.Store(&cfg)
-	h.limiter.UpdateConfig(cfg.GlobalWriteConfig, globalWrite)
-	h.limiter.UpdateConfig(cfg.GlobalReadConfig, globalRead)
+	if !reflect.DeepEqual(existingCfg.GlobalWriteConfig, cfg.GlobalWriteConfig) {
+		h.limiter.UpdateConfig(cfg.GlobalWriteConfig, globalWrite)
+	}
+	if !reflect.DeepEqual(existingCfg.GlobalReadConfig, cfg.GlobalReadConfig) {
+		h.limiter.UpdateConfig(cfg.GlobalReadConfig, globalRead)
+
+	}
```
dhiaayachi, line 189: I think we should add a TODO to log some message when we don't apply the config here.

dhiaayachi (approved):
@jmurret Thank you for putting this together!

I think it's good to have this in all cases. I prefer that we stop the processing as soon as possible as a general case. Also it give us an opportunity to log a message about it.

agent/consul/rate/handler.go
```diff
@@ -172,9 +181,18 @@ func (h *Handler) Allow(op Operation) error {
 }
 
 func (h *Handler) UpdateConfig(cfg HandlerConfig) {
+	existingCfg := h.cfg.Load()
 	h.cfg.Store(&cfg)
-	h.limiter.UpdateConfig(cfg.GlobalWriteConfig, globalWrite)
-	h.limiter.UpdateConfig(cfg.GlobalReadConfig, globalRead)
+	if reflect.DeepEqual(existingCfg, cfg) {
+		h.logger.Warn("UpdateConfig called but configuration has not changed.  Skipping updating the multilimiter.")
```
jmurret, line 187: Logging and exiting here because future checks / update calls will be added for server ip limiting in phase 2.

agent/consul/rate/handler.go
jmurret: @dhiaayachi Thank you for calling this out.  What do you think of something like this?  (It's my first foray into adding logging and wasn't sure if there was another way todo this.  Also, is warning the right level?)

https://github.com/hashicorp/consul/pull/15805/commits/33c36d25241b368f64883d4b784a2b70dab9c502

agent/consul/rate/handler.go
dhiaayachi, line 187:
```suggestion
		h.logger.Warn("UpdateConfig called but configuration has not changed.  Skipping updating the rate limiter configuration.")
```

dhiaayachi (approved):
I have a suggestion about the log wording but feel free to ignore.

LGTM!

PR #15806, opened 2022-12-15
Major updates and reorganizing of checks.mdx

am-ak:
### Description
- Removing Duplication at the top of the document
- Correcting links to go from less detailed section to more detailed section
- Removing misplaced Headlines

### PR Checklist
* [ ] updated test coverage
* [x] not a security concern


website/content/docs/discovery/checks.mdx
```diff
@@ -181,7 +77,7 @@ automatically monitor the health of a service instance or node.
   to temporarily remove one or all service instances on a node
   from service discovery DNS and HTTP API query results.
 
-### Script check ((#script-interval))
+### Script check
```
jkirschner-hashicorp, line 80: The `((#script-interval))` is an anchor tag alias that preserves backwards compatibility for other content that links to information about this kind of health check. Do we have a need to break that backwards compatibility? If not, I recommend preserving the existing link aliases (for all check types, not just this one).

jkirschner-hashicorp (changes requested):
Thanks for taking this on! I added a blocking review comment about link backwards compatibility that I want to make sure we align on before merging (separate from the education team review).

website/content/docs/discovery/checks.mdx
jkirschner-hashicorp: CC @boruszak

am-ak: the reason is that in the original doc, the detailed section at the bottom with the examples and all, was a link to the redundant section at the top so for example http check pointed to http+interval. with the update and shortening of the top section it would be even more ridiculous

jkirschner-hashicorp: I had previously restructured this page and was surprised to find the redundant info at the top. I'm not sure if I accidentally left that in, or if something weird happened in that merge or since. Either way, you're right that the bulleted items at the top were being linked to, not the details below, so we're effectively still preserving the old links with your change.

Cool - I'll dismiss my blocking review!

jkirschner-hashicorp (commented):
Dismissing my blocking review. LGTM

jkirschner-hashicorp (approved):
LGTM - pending the addressing of feedback from @boruszak

PR #15818, opened 2022-12-15
Enhancement: Consul Compatibility Checking

### LINKED ISSUE #13131: dropping support for envoy 1.18 in consul 1.12 should be listed as a breaking change
#### Overview of the Issue

Dropping support for a version of Envoy needs to be listed as a breaking change, not under "Improvements". 

We're setting up a new Consul cluster with Vault and using Vault as the CA for Consul Connect. We built a fully functional environment in our staging environment, but ran into issues in production because Envoy couldn't get the leaf certs through Consul. After over a day of investigation and research, we discovered that when we revved from Consul 1.11 to 1.12, we broke Consul Connect because support for Envoy 1.18 was dropped. Despite using debug logging, we didn't see any errors about this incompatibility.

#### Reproduction Steps

1. Install Consul 1.12
2. Install Envoy 1.18
3. Try to use Envoy as a sidecar proxy
4. See Envoy fail to retrieve the leaf certificate with an error that is unrelated to the version mismatch
5. Install Consul 1.11, and making no other changes, allowed Envoy to pull leaf certs and Consul Connect performed as expected

Amazon LInux 2

### Log Fragments

```
May 18 13:33:13 HOSTNAME consul: 2022-05-18T13:33:13.372-0400 [WARN]  agent: [core]grpc: Server.Serve failed to complete security handshake from "127.0.0.1:59926": tls: client didn't provide a certificate
May 18 13:33:13 HOSTNAME consul: [2022-05-18 13:33:13.372][7514][warning][config] [bazel-out/k8-opt/bin/external/envoy/source/common/config/_virtual_includes/grpc_stream_lib/common/config/grpc_stream.h:101] DeltaAggregatedResources gRPC config stream closed: 14, upstream connect error or disconnect/reset before headers. reset reason: connection failure, transport failure reason: TLS error: 268436498:SSL routines:OPENSSL_internal:SSLV3_ALERT_BAD_CERTIFICATE
```

wilkermichael:
### Description
Note: **docs team see last commit for docs change**

This PR introduces an enhancement to enable users to detect incompatible versions of Envoy early. This PR was motivated by issue [#13131](https://github.com/hashicorp/consul/issues/13131).

With the addition of these code changes, when using the `consul connect envoy` command the version of envoy being connected to is compared against the supported versions coded into the Consul binary. If the version of Envoy is deemed to be incompatible, then an error will be logged and the service will exit. 

Since this is a potentially breaking change, and there may be reasons that a user is using a version of Consul that we don't recognize as compatible (such as custom builds of Envoy), a flag has also been added to disable the check: `--ignore-envoy-compatibility`. By default this flag will be set to `false` and we will perform the check and error out as necessary.

Example: `consul connect envoy -admin-bind=0.0.0.0:19000 -sidecar-for=svc1 --ignore-envoy-compatibility`

#### How do we determine compatibility?
Compatibility is defined in [proxysupport.go](https://github.com/hashicorp/consul/blob/1406d428ddc83215a13a1fa0a3b3a7112b00e598/agent/xds/proxysupport/proxysupport.go#LL9C5-L9C5)
```Go
var EnvoyVersions = []string{
	"1.24.0",
	"1.23.2",
	"1.22.5",
	"1.21.5",
}
```
For the above list, Consul is compatible with any version `>= 1.21.0` and `<1.25`. For example, we are compatible with `1.21.1`, `1.21.2`, etc. and anything in between up to any future patches to `1.24`.

A list of explicitly unsupported Envoy versions has also been added to `proxysupport.go` in case there is a requirement to explicitly callout patches for breaking compatibility (e.g. if 1.21.2 had a known compatibility breaking change).

#### Log Message Example
```console
dc1svc1    | Envoy version 1.18.3 is not supported. If there is a reason you need to use this version of envoy use the ignore-envoy-compatibility flag. Using an unsupported version of Envoy is not recommended and your experience may vary. For more information on compatibility see https://developer.hashicorp.com/consul/docs/connect/proxies/envoy#envoy-and-consul-client-agent
```

This resolves #13131

As always, I try to add some extra context to my commits, so that's usually the best way to go through my PRs.

#### Manual Testing Scenarios
- **Lower Bound Test**: Tested a patch version below `1.21.5`, confirmed `no error`
- **In Between Test**: Tested a patch version between `1.21` and `1.24`, confirmed `no error`
- **Upper Bound Test**: Modified the list in `proxysupport.go` so that `1.23.1` was the top of the list, confirmed that `1.23.2` resulted in `no error`
- **Upper Out Of Bounds Test**: Modified the list in `proxysupport.go` so that `1.23.1` was the top of the list, confirmed that `1.24.0` resulted in `error` and exited
- **Lower Out Of Bounds Test**: Tested with Envoy `1.18`, confirmed that this resulted in an `error` and exited
- **Flag Testing**: Tested with `1.18` using `--ignore-envoy-compatibility` confirmed that there was `no error` for compatibility
- **Incompatibility List Testing**: added `1.21.1` to the incompatibility list, confirmed connecting to this version resulted in an `error` and exit

### Links
N/A

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/.changelog/15818.txt
+++ b/.changelog/15818.txt
@@ -1,0 +1,6 @@
+```release-note:enhancement
+connect: for early awareness of Envoy incompatibilities, when using the `consul connect envoy` command the Envoy version will now be checked for compatibility. If incompatible Consul will error and exit.
+```
+```release-note:breaking-change
+connect: Consul will now error and exit when using the `consul connect envoy` command if the Envoy version is incompatible. To ignore this check use flag `--ignore-envoy-compatibility`
+```
--- a/agent/xds/proxysupport/proxysupport.go
+++ b/agent/xds/proxysupport/proxysupport.go
@@ -1,5 +1,7 @@
 package proxysupport
 
+import "strings"
+
 // EnvoyVersions lists the latest officially supported versions of envoy.
 //
 // This list must be sorted by semver descending. Only one point release for
@@ -11,4 +13,29 @@ var EnvoyVersions = []string{
 	"1.23.2",
 	"1.22.5",
 	"1.21.5",
+}
+
+// UnsupportedEnvoyVersions lists any unsupported Envoy versions (mainly minor versions) that fall
+// within the range of EnvoyVersions above.
+// For example, if patch 1.21.3 (patch 3) had a breaking change, and was not supported
+// even though 1.21 is a supported major release, you would then add 1.21.3 to this list.
+// This list will be empty in most cases.
+//
+// see: https://www.consul.io/docs/connect/proxies/envoy#supported-versions
+var UnsupportedEnvoyVersions = []string{}
+
+// GetMaxEnvoyMinorVersion grabs the first value in EnvoyVersions and strips the patch number off in order
+// to return the maximum supported Envoy minor version
+// For example, if the input string is "1.14.1", the function would return "1.14".
+func GetMaxEnvoyMinorVersion() string {
+	s := strings.Split(EnvoyVersions[0], ".")
+	return s[0] + "." + s[1]
+}
+
+// GetMinEnvoyMinorVersion grabs the last value in EnvoyVersions and strips the patch number off in order
+// to return the minimum supported Envoy minor version
+// For example, if the input string is "1.12.1", the function would return "1.12".
+func GetMinEnvoyMinorVersion() string {
+	s := strings.Split(EnvoyVersions[len(EnvoyVersions)-1], ".")
+	return s[0] + "." + s[1]
 }
--- a/command/connect/envoy/envoy.go
+++ b/command/connect/envoy/envoy.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"strconv"
 	"strings"
 	"time"
 
@@ -21,6 +22,7 @@ import (
 	"github.com/hashicorp/consul/command/flags"
 	"github.com/hashicorp/consul/ipaddr"
 	"github.com/hashicorp/consul/tlsutil"
+	"github.com/hashicorp/go-version"
 )
 
 func New(ui cli.Ui) *cmd {
@@ -39,26 +41,27 @@ type cmd struct {
 	client *api.Client
 
 	// flags
-	meshGateway           bool
-	gateway               string
-	proxyID               string
-	nodeName              string
-	sidecarFor            string
-	adminAccessLogPath    string
-	adminBind             string
-	envoyBin              string
-	bootstrap             bool
-	disableCentralConfig  bool
-	grpcAddr              string
-	grpcCAFile            string
-	grpcCAPath            string
-	envoyVersion          string
-	prometheusBackendPort string
-	prometheusScrapePath  string
-	prometheusCAFile      string
-	prometheusCAPath      string
-	prometheusCertFile    string
-	prometheusKeyFile     string
+	meshGateway              bool
+	gateway                  string
+	proxyID                  string
+	nodeName                 string
+	sidecarFor               string
+	adminAccessLogPath       string
+	adminBind                string
+	envoyBin                 string
+	bootstrap                bool
+	disableCentralConfig     bool
+	grpcAddr                 string
+	grpcCAFile               string
+	grpcCAPath               string
+	envoyVersion             string
+	prometheusBackendPort    string
+	prometheusScrapePath     string
+	prometheusCAFile         string
+	prometheusCAPath         string
+	prometheusCertFile       string
+	prometheusKeyFile        string
+	ignoreEnvoyCompatibility bool
 
 	// mesh gateway registration information
 	register           bool
@@ -204,6 +207,10 @@ func (c *cmd) init() {
 	c.flags.StringVar(&c.prometheusKeyFile, "prometheus-key-file", "",
 		"Path to a private key file for Envoy to use when serving TLS on the Prometheus metrics endpoint. "+
 			"Only applicable when envoy_prometheus_bind_addr is set in proxy config.")
+	c.flags.BoolVar(&c.ignoreEnvoyCompatibility, "ignore-envoy-compatibility", false,
+		"If set to `true`, this flag ignores the Envoy version compatibility check. We recommend setting this "+
+			"flag to `false` to ensure compatibility with Envoy and prevent potential issues. "+
+			"Default is `false`.")
 
 	c.http = &flags.HTTPFlags{}
 	flags.Merge(c.flags, c.http.ClientFlags())
@@ -453,6 +460,27 @@ func (c *cmd) run(args []string) int {
 	if err != nil {
 		c.UI.Error("Couldn't find envoy binary: " + err.Error())
 		return 1
+	}
+
+	// Check if envoy version is supported
+	if !c.ignoreEnvoyCompatibility {
+		v, err := execEnvoyVersion(binary)
+		if err != nil {
+			c.UI.Warn("Couldn't get envoy version for compatibility check: " + err.Error())
+			return 1
+		}
+
+		ok, err := checkEnvoyVersionCompatibility(v, proxysupport.UnsupportedEnvoyVersions)
+
+		if err != nil {
+			c.UI.Warn("There was an error checking the compatibility of the envoy version: " + err.Error())
+		} else if !ok {
+			c.UI.Error(fmt.Sprintf("Envoy version %s is not supported. If there is a reason you need to use "+
+				"this version of envoy use the ignore-envoy-compatibility flag. Using an unsupported version of Envoy "+
+				"is not recommended and your experience may vary. For more information on compatibility "+
+				"see https://developer.hashicorp.com/consul/docs/connect/proxies/envoy#envoy-and-consul-client-agent", v))
+			return 1
+		}
 	}
 
 	err = execEnvoy(binary, nil, args, bootstrapJson)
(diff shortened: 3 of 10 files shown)


command/connect/envoy/envoy_test.go
```diff
@@ -125,7 +128,7 @@ type generateConfigTestCase struct {
 // This tests the args we use to generate the template directly because they
 // encapsulate all the argument and default handling code which is where most of
 // the logic is. We also allow generating golden files but only for cases that
-// pass the test of having their template args generated as expected.
+// pass the test of having their template args generated as expectedSupport.
```
wilkermichael, line 131: Accidental refactor artifact. I'll revert this

agent/xds/proxysupport/proxysupport.go
```diff
@@ -12,3 +14,28 @@ var EnvoyVersions = []string{
...
+func GetMaxEnvoyMajorVersion() string {
+	s := strings.Split(EnvoyVersions[0], ".")
+	return s[0] + "." + s[1]
+}
+
+// GetMinEnvoyMajorVersion grabs the last value in EnvoyVersions and strips the patch number off in order
+// to return the minimum supported Envoy Major version
+// For example, if the input string is "1.12.1", the function would return "1.12".
+func GetMinEnvoyMajorVersion() string {
+	s := strings.Split(EnvoyVersions[len(EnvoyVersions)-1], ".")
```
jkirschner-hashicorp, line 39: Do we need to handle the unlikely case that the `EnvoyVersions` array is misconfigured (e.g., empty list, or a value that doesn't contain a `.`).? Maybe not since this is a hard-coded internal array, but my brain spotted this as a hypothetical, potential out-of-bounds access issue (both on array `s` and array `EnvoyVersions`).

jkirschner-hashicorp: Same applies to the get max version func

agent/xds/proxysupport/proxysupport.go
```diff
@@ -12,3 +14,28 @@ var EnvoyVersions = []string{
...
+// This list will be empty in most cases.
+//
+// see: https://www.consul.io/docs/connect/proxies/envoy#supported-versions
+var UnsupportedEnvoyVersions = []string{}
+
+// GetMaxEnvoyMajorVersion grabs the first value in EnvoyVersions and strips the patch number off in order
+// to return the maximum supported Envoy Major version
+// For example, if the input string is "1.14.1", the function would return "1.14".
+func GetMaxEnvoyMajorVersion() string {
+	s := strings.Split(EnvoyVersions[0], ".")
```
jkirschner-hashicorp, line 31: Is it reasonable for us to assume that the 0th element of the list is the max major version?

I don't see a comment above `EnvoyVersions` stating that the array is expected to be sorted from newest to oldest major version. I think we should either document that expectation, or have the get min/max code actually iterate through the array to pull out the actual min/max (not assuming that `EnvoyVersions` is in order).

command/connect/envoy/envoy.go
```diff
@@ -448,6 +456,27 @@ func (c *cmd) run(args []string) int {
...
+
+		ok, err := checkEnvoyVersionCompatibility(v, proxysupport.UnsupportedEnvoyVersions)
+
+		if err != nil {
+			c.UI.Warn("There was an error checking the compatibility of the envoy version: " + err.Error())
+		} else if !ok {
+			c.UI.Error(fmt.Sprintf("Envoy version %s is not supported. If there is a reason you need to use "+
+				"this version of envoy use the ignore-envoy-compatibility flag. Using an unsupported version of Envoy "+
+				"is not recommended and your experience may vary. For more information on compatibility "+
+				"see https://developer.hashicorp.com/consul/docs/connect/proxies/envoy#envoy-and-consul-client-agent", v))
```
jkirschner-hashicorp, line 475: Note to self: if anyone moves this page or renames the subsection with this anchor tag, this link will no longer work. When I write the docs PR, I should make sure the docs page has an internal comment that calls out this dependency from the code that we should aim to preserve.

.changelog/15818.txt
```diff
@@ -0,0 +1,6 @@
+```release-note:enhancement
+connect: for early awareness of Envoy incompatibilities, when using the `consul connect envoy` command the Envoy version will now be checked for compatibility. If incompatible Consul will error and exit.
+```
+```release-note:breaking-change
+connect: Consul will now error and exit when using the `consul connect envoy` command if the Envoy version is incompatible. To ignore this check use flag `--ignore-envoy-compatibility`
```
jkirschner-hashicorp, line 5: I love the clarity of these changelog entries

command/connect/envoy/envoy.go
```diff
@@ -448,6 +456,27 @@ func (c *cmd) run(args []string) int {
...
+			c.UI.Warn("Couldn't get envoy version for compatibility check: " + err.Error())
+			return 1
+		}
+
+		ok, err := checkEnvoyVersionCompatibility(v, proxysupport.UnsupportedEnvoyVersions)
+
+		if err != nil {
+			c.UI.Warn("There was an error checking the compatibility of the envoy version: " + err.Error())
+		} else if !ok {
+			c.UI.Error(fmt.Sprintf("Envoy version %s is not supported. If there is a reason you need to use "+
```
jkirschner-hashicorp, line 472: Great error message

jkirschner-hashicorp (commented):
Added some comments to consider. I defer to the eng team for a formal review.

agent/xds/proxysupport/proxysupport.go
wilkermichael: Yeah, the array has this comment:
```
// EnvoyVersions lists the latest officially supported versions of envoy.
//
// This list must be sorted by semver descending. Only one point release for
// each major release should be present.
//
// see: https://www.consul.io/docs/connect/proxies/envoy#supported-versions
```

Some other parts of the code were assuming order, so I did the same

wilkermichael: I added a test to make sure it is in descending order as per the instructions

agent/xds/proxysupport/proxysupport.go
wilkermichael: I guess if we don't have `.` we have bigger problems, we wouldn't be able to parse the string out of `envoy --version` as well.

wilkermichael:
@hashi-derek This is ready for review

There's currently a failing integration test, but it's not related to these changes (It's failing in other PRs)

hashi-derek (approved):

website/content/commands/connect/envoy.mdx
```diff
@@ -104,6 +104,11 @@ Usage: `consul connect envoy [options] [-- pass-through options]`
   TLS on the Prometheus metrics endpoint. Only applicable when `envoy_prometheus_bind_addr`
   is set in proxy config.
 
+- `-ignore-envoy-compatibility` - Specifies whether or not the program should check the envoy version for compatibility.
+If this flag is set to true, the program will not check the envoy version for compatibility.
+It is recommended in most situations to leave this flag set to false, as ensuring compatibility with the
+Envoy version can prevent potential issues. By default this is set to `false`.
```
trujillo-adam, line 110:
```suggestion
- `-ignore-envoy-compatibility` - If set to `true`, this flag ignores the Envoy version compatibility check. We recommend setting this flag set to `false` to ensure compatibility with Envoy and prevent potential issues. Default is `false`.
```
Just a few tweaks to simplify the description and consistently use monospace. Did we update the documentation under /connect/proxies/envoy or wherever we describe the Envoy startup process? I was going to add a link, but I'm not sure where it would go. Maybe this would be a better place to document this behavior? /connect/registration/service-registration

trujillo-adam (commented):
I made a few adjustments to the flag description. Also, it seems that there should be an overview section somewhere that describes the default behavior when registering a sidecar proxy. If we don't already have this content, we should make a note to consider creating it.

website/content/commands/connect/envoy.mdx
wilkermichael: Sorry adam, I'm not sure which paths you're referring to? Where exactly is /connect/registration/service-registration or /connect/proxies/envoy?

This is described here because this is a flag for and specific to the `consul connect envoy` command

command/connect/envoy/envoy_test.go
```diff
@@ -1518,3 +1521,83 @@ func testMockAgentSelf(wantXDSPorts agent.GRPCPorts, agentSelf110 bool) http.Han
...
+			expectedSupport: true,
+		},
+		{
+			name:            "supported-minor-higher",
+			envoyVersion:    addNPatchVersion(proxysupport.EnvoyVersions[0], 1),
+			unsupportedList: proxysupport.UnsupportedEnvoyVersions,
+			expectedSupport: true,
+		},
+		{
+			name:            "not-supported-major-higher",
```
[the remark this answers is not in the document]
wilkermichael: yep, good catch... I realised at the end I had mixed that up everywhere, so must have missed these ones.

website/content/commands/connect/envoy.mdx
trujillo-adam: I was thinking one of these pages would be the place to describe the default bootstrap process: 
https://developer.hashicorp.com/consul/docs/connect/proxies/envoy
https://developer.hashicorp.com/consul/docs/connect/registration/service-registration

Maybe we don't need it, though. The beginning of this page says "The default behavior is to generate the necessary bootstrap configuration for Envoy based on the environment variables and options provided and by talking to the local Consul agent." Maybe we should add a sentence there letting people know about the version check.

jkirschner-hashicorp: Do we need to add a sentence about the version check at an overview level?

It seems like a thing that you don't need to be told about. If you run into it (because you're using an incompatible version), the error message you get will tell you what went wrong and why. It seems like something that's already very discoverable in-context in Consul itself, so doesn't need to be present at an overview level in the docs... Freeing us up to focus on other important, high-level info in the docs.

wilkermichael: I think that makes sense to me

wilkermichael: @jkirschner-hashicorp and @trujillo-adam are you okay with me resolving this discussing and merging this PR?

wilkermichael: I'm going to close this PR for now, we can always create another one if we want to address the documentation updates

PR #15819, opened 2022-12-15
Rate Limit Handler - ensure rate limiting is not in the code path when not configured

jmurret:
### Description
Consul should ensure that rate limiters are by passed when the rate limiting mode is set to disabled (which is the default).

### Considerations
We described two approaches:
- leaving the rate limiter instantiated
  - it does not create limiters in memory until allow for that key is called.
- setting limiter to a noop limiter that always returns true when allow is called

This PR demonstrates the first option.  

One thing to consider about this option is the assumption of what happens during a reload where the mode changes from permissive or enforcing to disabled where limiters are already loaded into memory.  These things are assumed to occur:
- Allow will be by passed on reload with this config change
- reload will not directly cleanup limiters that are currently in memory
- cleanup will occur out of process by the cleanup routine based on the interval to look for limiters that have not been called in a while.
  - this is process is how we have described handling config changes of any sort not just mode.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/agent.go
```diff
@@ -567,7 +567,7 @@ func (a *Agent) Start(ctx context.Context) error {
 
 	// gRPC calls are only rate-limited on server, not client agents.
 	var grpcRateLimiter rpcRate.RequestLimitsHandler
-	grpcRateLimiter = rpcRate.NullRateLimiter()
+	grpcRateLimiter = rpcRate.NullRequestLimitsHandler()
```
jmurret, line 570: I made this change because I originally was going to create a noop Rate Limiter and the fact that this is named NullRateLimiter when it is based on RequestLimitsHandler interface and no RateLimiter interface was going to cause confusion.
Anyway, this is now more accurate, but is not required for this PR.

This PR is essentially [this commit](https://github.com/hashicorp/consul/pull/15819/commits/2c24f08bda96a9cce31ba3942b471ec29fe0e37b)

agent/consul/rate/handler.go
```diff
@@ -169,11 +169,13 @@ func (h *Handler) Run(ctx context.Context) {
...
-	// Example:
-	//	if !h.limiter.Allow(globalWrite) {
-	//	}
+	existingCfg := h.cfg.Load()
+	if existingCfg.GlobalMode == ModeDisabled {
+		return nil
+	} else if !h.limiter.Allow(globalWrite) {
+		// TODO(NET-1383): actually implement the rate limiting logic and replace this returned nil.
+		return nil
+	}
```
dhiaayachi, line 178:
```suggestion
	if existingCfg.GlobalMode == ModeDisabled {
		return nil
	} 
	
	if !h.limiter.Allow(globalWrite) {
		// TODO(NET-1383): actually implement the rate limiting logic and replace this returned nil.
		return nil
	}
```

agent/consul/rate/handler.go
```diff
@@ -169,11 +169,15 @@ func (h *Handler) Run(ctx context.Context) {
...
 func (h *Handler) Allow(op Operation) error {
-	// TODO(NET-1383): actually implement the rate limiting logic.
-	//
-	// Example:
-	//	if !h.limiter.Allow(globalWrite) {
-	//	}
+	existingCfg := h.cfg.Load()
+	if existingCfg.GlobalMode == ModeDisabled {
+		return nil
+	}
```
boxofrad, line 175: nit: What's the reason for the `existing` prefix here? I think `cfg` would probably be clear enough?

boxofrad, line 175: nit: What's the reason for the `existing` prefix here? I think `cfg` would probably be clear enough.

agent/consul/rate/handler_test.go
```diff
@@ -85,3 +88,61 @@ func TestUpdateConfig(t *testing.T) {
...
+			}
+			mockRateLimiter.On("UpdateConfig", mock.Anything, mock.Anything).Return()
+			handler := NewHandlerWithLimiter(*tc.cfg, nil, mockRateLimiter)
+			h, p, err := net.SplitHostPort("127.0.0.1:1234")
+			require.NoError(t, err)
+			port, err := strconv.Atoi(p)
+			require.NoError(t, err)
+			mockRateLimiter.Calls = nil
+			handler.Allow(Operation{Name: "test", SourceAddr: &net.TCPAddr{IP: net.ParseIP(h), Port: port}})
+			mockRateLimiter.AssertNumberOfCalls(t, "Allow", tc.expectedAllowCalls)
```
boxofrad, line 145:
```suggestion
			addr := net.TCPAddrFromAddrPort(netip.MustParseAddrPort("127.0.0.1:1234"))
			mockRateLimiter.Calls = nil
			handler.Allow(Operation{Name: "test", SourceAddr: addr})
			mockRateLimiter.AssertNumberOfCalls(t, "Allow", tc.expectedAllowCalls)
```

boxofrad (approved):
Looks great, nice job @jmurret 🙌🏻

agent/consul/rate/handler.go
jmurret: good call.  fixed.

PR #15822, opened 2022-12-16
inject logger and create logdrop sink

dhiaayachi:
### Description
This add the integration of `logdrop` `SinkAdapter` implementation into the rate-limiter handler.
### Links
#15670 

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


agent/consul/rate/handler.go
```diff
@@ -150,6 +155,7 @@ func NewHandler(cfg HandlerConfig, delegate HandlerDelegate) *Handler {
 		cfg:      new(atomic.Pointer[HandlerConfig]),
 		delegate: delegate,
 		limiter:  limiter,
+		logger:   logger.NamedIntercept("rate-limit"),
```
boxofrad, line 158: I think sub-loggers will forward logs to the parent logger, not just their own registered sinks.

See here:
- https://github.com/hashicorp/go-hclog/blob/8b7499ad6ad46ce583e825d324d008032b644703/interceptlogger.go#L116
- https://github.com/hashicorp/go-hclog/blob/8b7499ad6ad46ce583e825d324d008032b644703/interceptlogger.go#L47

Which I think means we'd end up sending logs to STDOUT twice (and the dropper wouldn't be effective).

Maybe we need to create a totally new logger (perhaps with `Output: io.Discard`) and pass the parent logger as the sink argument in `NewLogDropSink`?

dhiaayachi: Yes and that mean we need to change the drop log to accept a logger instead of a sink. I made the change, let me know if that what you had in mind.

agent/consul/rate/handler.go
```diff
@@ -170,6 +175,10 @@ func NewHandler(cfg HandlerConfig, delegate HandlerDelegate, logger hclog.Logger
 // Note: this starts a goroutine.
 func (h *Handler) Run(ctx context.Context) {
 	h.limiter.Run(ctx)
+	discardLogger := hclog.NewInterceptLogger(&hclog.LoggerOptions{Output: io.Discard})
+	discardLogger.RegisterSink(logdrop.NewLogDropSink(ctx, 100, h.logger, func(l logdrop.Log) {
```
jmurret, line 179: just curious how do we select 100 as the depth?

jmurret:
@boxofrad @analogue Can you please just give this a review so we can get bearings on what else needs to be added.  I think Dhia covered the sub-logger/parent-logger duplication.  I added warning logging for rate limit blocking.  Please give any thoughts on outstanding work as well as if the logging is at the right levels for each message.  Thanks!

boxofrad (approved):
This is looking great 🙌🏻 

Heads up @jmurret, I pushed a couple of small changes as it felt a little easier than explaining via comments. Please feel free to revert/drop them!

Specifically:

- Added a Prometheus definition (and docs) for the log drop counter.
- Fixed a bug where log arguments weren't getting correctly printed.
- Moved the sink construction out of the handler and into the server setup code. This means the handler can just use whatever logger it's given (and we can easily swap it out for syslog or a file).

jmurret (approved):
👍 thank you for just going ahead and making these changes!

agent/consul/rate/handler.go
dhiaayachi: Circling back on this. The 100 is just a number that seemed fine in my head. I think we should refine that with more testing.
My logic is that 100 is enough to smoothen any peak and over than that it's fine to start dropping as most of those logs have duplicate information anyway.

PR #15825, opened 2022-12-16
Map net/rpc endpoints to a read or write op for rate-limiting

analogue:
### Description
Maps each net/rpc endpoint to a read or write operation type. This mapping will be used by server agent net/rpc middleware as required input to the rate limiter. 

Reviewers: please pay special attention to the method names which are *not* easily classified (those not matching `Set`, `Get`, `Apply`, `Delete`, `List`, etc).

A new `rate.OperationTypeExempt` was added to track endpoints that should be exempt from rate-limiting.

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


agent/rpc/middleware/rate_limit_mappings.go
```diff
@@ -0,0 +1,144 @@
...
+	"KVS.List":     rate.OperationTypeRead,
+	"KVS.ListKeys": rate.OperationTypeRead,
+
+	"Operator.AutopilotGetConfiguration": rate.OperationTypeRead,
+	"Operator.AutopilotSetConfiguration": rate.OperationTypeWrite,
+	"Operator.AutopilotState":            rate.OperationTypeRead,
+	"Operator.RaftGetConfiguration":      rate.OperationTypeRead,
+	"Operator.RaftRemovePeerByAddress":   rate.OperationTypeWrite,
+	"Operator.RaftRemovePeerByID":        rate.OperationTypeWrite,
+	"Operator.ServerHealth":              rate.OperationTypeRead,
```
dhiaayachi, line 121: I commented about this in the grpc [PR](https://github.com/hashicorp/consul/pull/15564).

I think we need to add a third type, (eg: bypass). All operator endpoints should not be rate limited as they can be used by an operator in an outage situation.

agent/rpc/middleware/rate_limit_mappings.go
```diff
@@ -0,0 +1,144 @@
...
+	"Session.Check":        rate.OperationTypeRead,
+	"Session.Get":          rate.OperationTypeRead,
+	"Session.List":         rate.OperationTypeRead,
+	"Session.NodeSessions": rate.OperationTypeRead,
+	"Session.Renew":        rate.OperationTypeWrite,
+
+	"Status.Leader":    rate.OperationTypeRead,
+	"Status.Peers":     rate.OperationTypeRead,
+	"Status.Ping":      rate.OperationTypeRead,
+	"Status.RaftStats": rate.OperationTypeRead,
```
dhiaayachi, line 140: Same here we need to bypass those, those are used by autopilot, which is considered as vital to the cluster.

agent/rpc/middleware/rate_limit_mappings.go
```diff
@@ -0,0 +1,144 @@
...
+	"ConfigEntry.Get":                  rate.OperationTypeRead,
+	"ConfigEntry.List":                 rate.OperationTypeRead,
+	"ConfigEntry.ListAll":              rate.OperationTypeRead,
+	"ConfigEntry.ResolveServiceConfig": rate.OperationTypeRead,
+
+	"ConnectCA.ConfigurationGet": rate.OperationTypeRead,
+	"ConnectCA.ConfigurationSet": rate.OperationTypeWrite,
+	"ConnectCA.Roots":            rate.OperationTypeRead,
+	"ConnectCA.Sign":             rate.OperationTypeWrite,
+	"ConnectCA.SignIntermediate": rate.OperationTypeWrite,
```
dhiaayachi, line 68: Not for the scope of this PR, but the `ConnectCA.Roots`, `ConnectCA.Sign`, `ConnectCA.SignIntermediate` are used both to auto generate certs for the servers and to sign leaf for clients. 

I'm wondering if we need to find a way to bypass the call from the server and rate limit the calls to sign client certs somehow 🤔

dhiaayachi (changes requested):
Thank you @analogue for working on this. I had a few comments about endpoints that we should bypass.

agent/rpc/middleware/rate_limit_mappings.go
analogue: Rather than assign additional meaning to the `OperationType` which seems to be doing a good job of capturing read/write intent, how about capturing "exempt from rate limiting" status separately?
```Go
// endpoints that are exempt from rate limiting
var rpcExemptOperations = []string {
    "Operator.AutopilotState",
    ...
}
```

Code that will probably end up depending on `OperationType` to have a read/write value will be more straightforward and not have to be concerned about non-read/write contexts. e.g. emitting metrics with a read/write label.

agent/rpc/middleware/rate_limit_mappings.go
analogue: Not actionable in this PR.

agent/rpc/middleware/rate_limit_mappings.go
analogue: Updated

agent/rpc/middleware/rate_limit_mappings.go
dhiaayachi: Added a task (NET-1893) to track this

dhiaayachi (approved):
Thank you for taking my comments into account!

LGTM!

PR #15827, opened 2022-12-16
test: fix compat integration error from new access log api

DanStough:
### Description
This fixes failures in the integration test suite for upgrades. The problem was caused by using the latest version of `api` module, with new fields, against the upgrade source of Consul, resulting in a 400 error. [This causes the tests to timeout](https://app.circleci.com/pipelines/github/hashicorp/consul/48423/workflows/9e8c0326-6824-4888-98d4-e01ef42f45aa/jobs/1099014). 

This isn't a fix for the problem so much as removing a nice-to-have step. All the values in the proxy-default are assumed to be the fault, so the effect is negligible.

PR #15830, opened 2022-12-17
tests: Pin api submodule to compatibile version for upgrade tests

eculver:
### Description
This is an alternative to #15827. The details are in #15827 but the gist is that because the upgrade tests use the latest version of the `api` submodule, the test code will fail when running against a Consul 1.14 container instance because it is sending a config field that is not known. I was able to pull in a compatible api submodule using a `replace` directive. It seems to work, so I'm wondering if we should use this instead of removing the test code that triggers the compatibility issue. 

```
cd test/integration/consul-container
go test -run TestPeering_UpgradeToTarget_fromLatest ./... --target-version local --latest-version latest
```

### DIFF
--- a/test/integration/consul-container/go.mod
+++ b/test/integration/consul-container/go.mod
@@ -2,11 +2,14 @@ module github.com/hashicorp/consul/test/integration/consul-container
 
 go 1.19
 
+replace github.com/hashicorp/consul/api/v118 => github.com/hashicorp/consul/api v1.18.0
+
 require (
 	github.com/docker/docker v20.10.11+incompatible
 	github.com/docker/go-connections v0.4.0
 	github.com/hashicorp/consul v1.14.1
 	github.com/hashicorp/consul/api v1.18.0
+	github.com/hashicorp/consul/api/v118 v118.0.0-00010101000000-000000000000
 	github.com/hashicorp/consul/sdk v0.13.0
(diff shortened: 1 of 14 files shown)


test/integration/consul-container/go.mod
```diff
@@ -2,11 +2,14 @@ module github.com/hashicorp/consul/test/integration/consul-container
...
 go 1.19
 
+replace github.com/hashicorp/consul/api/v118 => github.com/hashicorp/consul/api v1.18.0
+
 require (
 	github.com/docker/docker v20.10.11+incompatible
 	github.com/docker/go-connections v0.4.0
 	github.com/hashicorp/consul v1.14.1
 	github.com/hashicorp/consul/api v1.18.0
+	github.com/hashicorp/consul/api/v118 v118.0.0-00010101000000-000000000000
```
eculver, line 12: I don't know why the line above does not result in `api/v1.18.0` from being used, but it doesn't, the `replace` directive appears to be required.

huikang:
@eculver , there is another alternative: I think in Go the json tag will ignore `",omitempty"` if it is a struct. So we may need to change all of them to pointer like `*TransparentProxyConfig`

https://github.com/hashicorp/consul/blob/ab199a11b0d37f3f943936c8f26eaa43918e0b6f/api/config_entry.go#L294-L298

Please see #15831

PR #15833, opened 2022-12-19
Fix incorrect protocol check on discovery chains with peer targets.

hashi-derek:
The following message is incorrectly printed out whenever saving a service-resolver with a peer failover target.

`discovery chain "my-chain" uses inconsistent protocols; service "my-service" has "tcp" which is not "http"`

This PR excludes peer targets from the protocol validation logic when building discovery chains, which should prevent the error from occurring.

Steps to reproduce:

1. Set up a service-defaults with “http” protocol on a service “mylocal”
2. Set up a peering connection and import a remote service “myremote” (this remote's protocol shouldn't matter to reproduce the issue)
3. Attempt to save a resolver for “mylocal” that has a failover to “myremote” configured.

Be careful to ensure that no "myremote" service-default exists in the local cluster, as it will actually be looked up, and the error will not surface.

### DIFF
--- a/agent/consul/discoverychain/compile.go
+++ b/agent/consul/discoverychain/compile.go
@@ -874,8 +874,14 @@ RESOLVE_AGAIN:
 
 	targetID := target.ServiceID()
 
-	if err := c.recordServiceProtocol(targetID); err != nil {
-		return nil, err
+	// Only validate protocol if it is not a peered service.
+	// TODO: Add in remote peer protocol validation when building a chain.
+	// This likely would require querying the imported services, which
+	// shouldn't belong in the discovery chain compilation here.
+	if target.Peer == "" {
+		if err := c.recordServiceProtocol(targetID); err != nil {
+			return nil, err
+		}
 	}
 
 	// Fetch the config entry.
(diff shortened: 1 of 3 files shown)


agent/consul/discoverychain/compile.go
```diff
@@ -874,8 +874,14 @@ RESOLVE_AGAIN:
 
 	targetID := target.ServiceID()
 
-	if err := c.recordServiceProtocol(targetID); err != nil {
-		return nil, err
+	// Only validate protocol if it is not a peered service.
+	// TODO: Add in remote peer protocol validation when building a chain.
+	// This likely would require querying the imported services, which
+	// shouldn't belong in the discovery chain compilation here.
+	if target.Peer == "" {
```
rboyer, line 881: Ah so this boils down to "there is not a `service-defaults` for imported services to store the protocol" if I'm understanding.

hashi-derek: Yup. I was still writing the description up for the ticket, haha. I guess the system automatically assigned you to this PR.

erichaberkorn (approved):

PR #15837, opened 2022-12-19
Add controller stubs for API Gateway

sarahalsmiller:
### Description
3 empty controller stubs are being added to Consul as part of the work porting the Consul API Gateway

### Links

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


agent/consul/controller/controller_gateways.go
```diff
@@ -0,0 +1,58 @@
...
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+	return nil
+}
+
+func newAPIGatewayReconciler() *apiGatewayReconciler {
+	return &apiGatewayReconciler{}
+}
+
+func NewAPIGatewayController(ctx context.Context) *APIGatewayController {
```
andrewstucki, line 30: So, we shouldn't need any of these structs -- it should just be a matter of creating a struct that implements "Reconciler" and configuring a controller that leverages it accordingly, very similar to any of the existing k8s gateway stuff:

https://github.com/hashicorp/consul-api-gateway/blob/main/internal/k8s/controllers/gateway_class_controller.go

Really all you'll need is a method similar to the bits in `SetupWithManager` that does something like what's already in the unit tests:

https://github.com/hashicorp/consul/blob/831680d2c5b3c34caffee89168025b16f897f151/agent/consul/controller/controller_test.go#L47-L50

andrewstucki (changes requested):
I think this should likely take a slightly different approach based off of the comment below, let me know if it makes sense.

agent/consul/gateways/controller_gateways.go
```diff
@@ -0,0 +1,39 @@
...
+	"context"
+	"github.com/hashicorp/consul/agent/consul/controller"
+	"github.com/hashicorp/consul/agent/consul/fsm"
+	"github.com/hashicorp/consul/agent/consul/state"
+	"github.com/hashicorp/consul/agent/consul/stream"
+	"github.com/hashicorp/go-hclog"
+)
+
+//baseReconciler contains fields shared across multiple controller.Reconciler implementations in the gateways package
+type baseReconciler struct {
```
andrewstucki, line 13: Main thing here is that I'm not really sure we need this -- I'd imagine that the various reconcilers are going to leverage different fields at some point and that aside from maybe the route reconcilers, the logic is going to be fairly different, so there's not going to be much else on this struct...

So, personally, I'd prefer if we just dropped the struct from this and declared the fields directly on the other reconciler structs rather than embedding this. Aside from that, this looks fine.

nathancoleman: done

andrewstucki (approved):

nathancoleman:
Waiting until @t-eckert merges #15757 - which this PR is based on top of - to merge

nathancoleman (approved):

t-eckert (approved):

PR #15834, opened 2022-12-19
Backport of Add custom balancer to always remove subConns into release/1.12.x

hc-github-team-consul-core:
## Backport

This PR is auto-generated from #15701 to be assessed for backporting due to the inclusion of the label backport/1.12.

**WARNING** automatic cherry-pick of commits failed. Commits will require human attention.

> merge conflict error: POST https://api.github.com/repos/hashicorp/consul/merges: 409 Merge conflict []

The below text is copied from the body of the original PR.

---

Fixes https://github.com/hashicorp/consul/issues/10603

### Description
Remove the need to call `UpdateState` twice by changing the `pick_first` balancer's behavior to close existing connections (almost) every time.

### Testing & Reproduction steps
No net-new behavior; existing test `TestClientConnPool_IntegrationWithGRPCResolver_Rebalance` is still passing.

Manually tested with frequent rebalancing

### PR Checklist

* [x] updated test coverage
* [x] external facing docs updated
* [x] not a security concern

---

<details>
<summary> Overview of commits </summary>

  - c37343dfcc9e86458cfbbe1d32183546b11126b0  - 27a0ffb6e8ed1695de177278d00fddc7a758d5f0  - ee5841db5438740e86859c48b662ecb4b17fd441  - 35f72880ce65311ed82e0f04e6b2e06a8c1ad857  - e8be8e4d55babae4dce7da6308d855ab606d7fe0  - 44d5c20f1661cc737201d189248a2142d8dab339  - 5420eddfb01c114a67fbb14ab707a47f9c9ee672  - ec1b87a98e2df0180c368e390fe6891f19a2a626 

</details>

PR #15844, opened 2022-12-19
grpc custom balancer - UpdateClientConnState Break - instead of returning early

kisunji:
Small followup to https://github.com/hashicorp/consul/pull/15701

### DIFF
--- a/agent/grpc-internal/balancer/custombalancer.go
+++ b/agent/grpc-internal/balancer/custombalancer.go
@@ -45,7 +45,7 @@ func (b *customPickfirstBalancer) UpdateClientConnState(state balancer.ClientCon
 		// of addresses matched the currently connected address, it would
 		// be an effective no-op.
 		if a.Equal(b.activeAddr) {
-			return nil
+			break
 		}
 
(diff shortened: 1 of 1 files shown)


agent/grpc-internal/balancer/custombalancer.go
```diff
@@ -45,7 +45,7 @@ func (b *customPickfirstBalancer) UpdateClientConnState(state balancer.ClientCon
 		// of addresses matched the currently connected address, it would
 		// be an effective no-op.
 		if a.Equal(b.activeAddr) {
-			return nil
+			break
```
kisunji, line 48: Returning early would mean the rest of the addresses never get stored. Break instead.
This patch has been applied during the backports to 1.12.x (https://github.com/hashicorp/consul/pull/15834/commits/007a0f5b90d7f1c45c24ca61228288c3241f7f4d) and 1.13.x (https://github.com/hashicorp/consul/pull/15835/commits/5028b3f0bac8ab732e34b063210e8c2e7d555d87)

jmurret (approved):
👍 THank you!

PR #15855, opened 2022-12-20
APIGateway TCPRoute validation

andrewstucki:
### Description
This adds the basic fields we need for an TCPRoutes along with internal validation logic.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


t-eckert (approved):

t-eckert (approved):

agent/structs/config_entry_routes.go
```diff
@@ -79,6 +83,12 @@ type TCPRouteConfigEntry struct {
 	// of resources.
 	Name string
 
+	// Parents is a list of gateways that this route should be bound to
+	Parents []ResourceReference
+	// Services are a list of TCP-based services that this should route to.
+	// Currently only one service must be specified.
+	Services []TCPService
```
nathancoleman, line 90:
```suggestion
	// Services is a list of TCP-based services that this should route to.
	// Currently, this must specify exactly one service.
	Services []TCPService
```

agent/structs/config_entry_routes.go
```diff
@@ -103,10 +113,28 @@ func (e *TCPRouteConfigEntry) GetMeta() map[string]string {
...
 	return nil
 }
 
 func (e *TCPRouteConfigEntry) Validate() error {
+	validParentKinds := map[string]bool{
+		APIGateway: true,
+	}
+
+	if len(e.Services) > 1 {
+		return fmt.Errorf("tcp-based routes currently only support one service")
```
nathancoleman, line 131: I believe this would be clearer to just explicitly specify the type
```suggestion
		return fmt.Errorf("tcp-route currently only supports one service")
```

agent/structs/config_entry_routes.go
```diff
@@ -103,10 +113,28 @@ func (e *TCPRouteConfigEntry) GetMeta() map[string]string {
...
+	}
+	for _, parent := range e.Parents {
+		if !validParentKinds[parent.Kind] {
+			return fmt.Errorf("unsupported parent kind: %q", parent.Kind)
```
nathancoleman, line 135: I think we should consider including the list of acceptable parent kinds in the error message here, assuming that it will appear on the resource's status post-validation and will only ever be a handful of acceptable kinds.

agent/structs/config_entry_routes.go
```diff
@@ -135,3 +163,13 @@ func (e *TCPRouteConfigEntry) GetEnterpriseMeta() *acl.EnterpriseMeta {
 	}
 	return &e.EnterpriseMeta
 }
+
+// TCPService is a service reference for a TCPRoute
+type TCPService struct {
+	Name string
+	// Weight is an arbitrary integer used in calculating how much
+	// traffic should be sent to the given service.
```
nathancoleman, line 171: I think it'd be helpful to explain how `Weight` works here. The Gateway API [docs](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.BackendRef) say
> Weight specifies the proportion of requests forwarded to the referenced backend. This is computed as weight/(sum of all weights in this BackendRefs list)

Since `tcp-route` can only route to a single service right now, this is just a placeholder for future functionality, right?

api/config_entry_routes.go
```diff
@@ -69,6 +72,20 @@ func (a *TCPRouteConfigEntry) GetModifyIndex() uint64 {
 	return a.ModifyIndex
 }
 
+// TCPService is a service reference for a TCPRoute
+type TCPService struct {
+	Name   string
+	Weight int
```
nathancoleman, line 78: Feels important to document the fields on this struct since these are the ones external API consumers will be interacting with I believe

nathancoleman (commented):
Some feedback here but no blockers. Nice work!

PR #15857, opened 2022-12-20
Wire in rate limiter to  handle internal and external gRPC calls

boxofrad:
(This is @analogue taking dan's branch to completion).  fyi @boxofrad since I can't add you to the list of reviewers. 

### Description
Wires up the gRPC rate limiter for server agents. Both external and internal gRPC servers are subject to the rate limits. 

Dan did the majority of the work here, I (@analogue) just moved creation of the `rpcRate.RequestLimitsHandler` out of `consul.Server` and into `agent.Agent.Start(...)` to work around some circular dependency issues.  `RequestLimitsHandler.RegisterDelegate()` was introduced so that the object, `consul.Server`, fulfilling the `HandlerDelegate.IsLeader()` interface can be "fixed up" and make accessible to the request limits handler as a post-creation step (the workaround for circ deps). 

I did the least amount possible to merge in the existing changes from https://github.com/hashicorp/consul/pull/15805 and https://github.com/hashicorp/consul/pull/15819 hence didn't refactor any of the tests for possible overlap.

### PR Checklist

* [x] updated test coverage
* [ ] not a security concern


agent/consul/rate/handler.go
```diff
@@ -166,11 +183,34 @@ func (h *Handler) Run(ctx context.Context) {
...
+		enforced := l.mode == ModeEnforcing
+		h.logger.Trace("RPC exceeded allowed rate limit",
+			"rpc", op.Name,
+			"source_addr", op.SourceAddr.String(),
+			"limit_type", l.desc,
+			"limit_enforced", enforced,
+		)
+
+		if enforced {
+			if h.delegate.IsLeader() {
```
dhiaayachi, line 206: Because we register a delegate after the server creation, I think we should have a safe guard in Allow in case delegate is nil. Also, I think we should make delegate an atomic to avoid races.

agent/consul/rate/handler.go
```diff
@@ -130,8 +137,13 @@ type HandlerConfig struct {
...
 	// GlobalReadConfig configures the global rate limiter for read operations.
 	GlobalReadConfig multilimiter.LimiterConfig
+
+	// makeLimiter is a function that can be passed by tests to inject a mock
+	// multilimiter (it is unexported so it cannot be used in production code).
+	makeLimiter func(multilimiter.Config) multilimiter.RateLimiter
 }
 
+//go:generate mockery --name HandlerDelegate --inpackage --filename mock_HandlerDelegate_test.go
 type HandlerDelegate interface {
```
dhiaayachi, line 147: Personal preference so feel free to ignore but the name is a bit confusing

dhiaayachi (commented):
Thank you for working on this @analogue 
I left few comments for your consideration.

agent/consul/rate/handler.go
analogue: Good idea, `handler.Run(...)` now panics if `handler.RegisterDelegate(...)` hasn't been set.

I'm not sure I understand why `delegate` should be an atomic. `delegate.IsLeader()` is read-only and the underlying implementation ends up calling `rate.GetState(...)`.

```Go
func (r *raftState) getState() RaftState {
	stateAddr := (*uint32)(&r.state)
	return RaftState(atomic.LoadUint32(stateAddr))
}
```

agent/consul/rate/handler.go
analogue: Agreed, suggestions for better names welcome 😄 .

analogue:
@boxofrad @dhiaayachi @jmurret I think this is a lot closer to some sort of final state and have updated the description accordingly. Please review as time allows.

agent/consul/rate/handler.go
```diff
@@ -175,14 +182,42 @@ func (h *Handler) Run(ctx context.Context) {
 // Allow returns an error if the given operation is not allowed to proceed
 // because of an exhausted rate-limit.
 func (h *Handler) Allow(op Operation) error {
+
+	if h.delegate == nil {
+		panic("delegate not set on handler via RegisterDelegate(..)")
```
dhiaayachi, line 187: Maybe ass some logs around here to point that this should never happen and if it happen, it should not panic the whole program but should recover because we have a panic recovery handler added to the server.

agent/consul/rate/handler.go
dhiaayachi: Yeah I'm really bad at naming things 😆 

What about `ServerRoleDelegate` or `LeadershipCheckDelegate`?

agent/consul/rate/handler.go
dhiaayachi: I think we are also missing a check here about the request being a write, a snippet from the RFC:

```
The only exception to this will be write requests rate-limited in the leader node 
by a global limit which will return a UNAVAILABLE response.
The reason for this is that if a write request hit a global limit in the leader node:
It’s not a retriable error, as only the leader is able to handle this request
The leader reached a critical limit
```

If that not possible right now, because of dependencies in other PRs, maybe we can add a TODO around it.

dhiaayachi (approved):
@analogue I added few comments to be considered but nothing blocking for the scope of this PR.

agent/consul/rate/handler.go
analogue: Good catch! Fixed in https://github.com/hashicorp/consul/pull/15857/commits/2a936ab9202fde18b13341bf00970b693ac77b2a

agent/consul/rate/handler.go
analogue: Added a TODO to follow through on the panic handler stuff (I'm not too familiar with recovery handlers atm) so as not to block this PR.  Updates in https://github.com/hashicorp/consul/pull/15857/commits/d86244e8f6f03a5ea5decd111a978bc59c31b4dc

analogue:
@dhia Renamed `HandlerDelegate` -> `LeaderStatusProvider` in https://github.com/hashicorp/consul/pull/15857/commits/02ca05b195697814bce41cfad10ace4322711eb8

agent/consul/rate/handler.go
analogue: Renamed HandlerDelegate -> LeaderStatusProvider in https://github.com/hashicorp/consul/commit/02ca05b195697814bce41cfad10ace4322711eb8

dhiaayachi (approved):
🚀

PR #15859, opened 2022-12-20
APIGateway HTTPRoute scaffolding

andrewstucki:
### Description
This adds the basic fields we need for an HTTPRoute. Validation will come in a separate PR.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/structs/config_entry_routes.go
```diff
@@ -73,6 +81,129 @@ func (e *HTTPRouteConfigEntry) GetRaftIndex() *RaftIndex {
...
+	Path    HTTPPathMatch
+	Query   []HTTPQueryMatch
+}
+
+// HTTPMatchMethod specifies which type of HTTP verb should
+// be used for matching a given request.
+type HTTPMatchMethod string
+
+const (
+	HTTPMatchMethodAll     HTTPMatchMethod = ""
```
t-eckert, line 99: Is this intention here that if the value is unset, then the match works for all methods? I'm picturing the scenario where a `string` is initialized with the value of `""` and matches all methods. If this is the desired implementation, the `""` makes sense, otherwise it might make more sense to have this set at `"*"`.

andrewstucki: My preference is to keep this as just the default. Basically, not specifying results in all HTTP methods being matched, so rather than trying to disambiguate between not specified and zero-value behavior via something like a pointer, I'd rather just use the zero-value to specify the default behavior.

proto/pbconfigentry/config_entry.go
```diff
@@ -267,6 +267,168 @@ func meshGatewayModeToStructs(a MeshGatewayMode) structs.MeshGatewayMode {
...
 }
 
+func apiGatewayProtocolFromStructs(a structs.APIGatewayListenerProtocol) APIGatewayListenerProtocol {
+	switch a {
+	case structs.ListenerProtocolHTTP:
+		return APIGatewayListenerProtocol_ListenerProtocolHTTP
+	case structs.ListenerProtocolTCP:
+		return APIGatewayListenerProtocol_ListenerProtocolTCP
+	default:
+		return APIGatewayListenerProtocol_ListenerProtocolHTTP
```
t-eckert, line 277: Is this safe? Would it be better to panic if there is no protocol?

t-eckert (approved):
Fantastic! I'm going to see switch statements whenever I close my eyes for a few days, but it's really clean and straightforward implementation.

PR #15864, opened 2022-12-21
[OSS] feat: access logs for listeners and listener filters

DanStough:
### Description
This is the meat of the access logging epic. It adds an Envoy access log configuration to all listeners and listener filters according to the new `proxy-defaults` API (#15780 ).  Additional notes inline.

* Goldenfile Tests
* Manual E2E tests. [See Dan's test files](https://github.com/DanStough/consul-scenarios/blob/c3539b23314440df77dcc8418d2172e7942f700e/local/access-logs).

### PR Checklist

* [X] updated test coverage
* [ ] ~external facing docs updated~ TBD
* [X] not a security concern


hashi-derek (approved):

agent/structs/connect_proxy_config.go
```diff
@@ -174,8 +174,16 @@ type AccessLogsConfig struct {
 	TextFormat string `json:",omitempty" alias:"text_format"`
 }
 
-func (c AccessLogsConfig) ToAPI() api.AccessLogsConfig {
-	return api.AccessLogsConfig{
+func (c *AccessLogsConfig) IsZero() bool {
+	zeroVal := AccessLogsConfig{}
```
hashi-derek, line 178: Should we worry about `c` being nil here? Maybe a non-pointer receiver or return true on nil?

command/services/config.go
```diff
@@ -71,6 +71,16 @@ func serviceToAgentService(svc *structs.ServiceDefinition) (*api.AgentServiceReg
 		result.Check = nil
 	}
 
+	// The structs version has non-pointer Proxy.TransparentProxy and Proxy.AccessLogs
+	// The destination has pointers, so we need to set the destination to nil if there
+	// is a zero-value field.
+	if result.Proxy != nil && result.Proxy.TransparentProxy != nil && reflect.DeepEqual(*result.Proxy.TransparentProxy, api.TransparentProxyConfig{}) {
```
hashi-derek, line 77: I don't know if deep-equal is necessary, since there's no pointers in this struct, but I suppose it's better to be safe for future changes.

agent/structs/connect_proxy_config.go
DanStough: Good call. I'll return `true` on `nil`.

PR #15865, opened 2022-12-21
Fix issue with incorrect proxycfg watch on upstream peer-targets.

hashi-derek:
Sometimes when envoy was configured to do failovers to peered clusters, the SNI / SAN information was missing from the xDS configuration, which means that it was impossible to connect to the mesh gateway correctly.

This issue occurs because the incorrect partition / datacenter was used when creating watches from upstream targets. The datacenter / partition from discovery-chain peer-targets should NOT be used to perform lookups, since the fields are intentionally cleared during compilation (disco chain target fields indicate the destination's data, not the source's data).

Instead, the local datacenter / partition should be provided to the query so that it can properly lookup services.

### DIFF
--- a/.changelog/15865.txt
+++ b/.changelog/15865.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+connect: Fix issue where watches on upstream failover peer targets did not always query the correct data.
+```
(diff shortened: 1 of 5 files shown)


ndhanushkodi (approved):
Looks good to me, I just decided to review for my learning. Thanks for the super helpful comments in the PR!

PR #15866, opened 2022-12-21
Fix agent cache incorrectly notifying unchanged protobufs.

hashi-derek:
This change fixes a situation where the protobuf private fields would be read by reflect.DeepEqual() and indicate data was modified. This resulted in change notifications being fired every time, which could cause performance problems in proxycfg.


DanStough (approved):
Awesome find 🥇 !

erichaberkorn (approved):

erichaberkorn (commented):
It might be helpful to add some tests for `isEqual`

PR #15884, opened 2023-01-03
Output user-friendly name for anonymous token

kisunji:
### Description
When ACLs are enabled but an operator forgets to pass a token, a potential ACL error may refer to the [anonymous token](https://developer.hashicorp.com/consul/docs/security/acl/acl-tokens#anonymous-token) which appears in the logs as `00000000-0000-0000-0000-000000000002`.

We should avoid the accessorID being a source of confusion for users by simply aliasing it to `anonymous token` whenever we output it in logs.

`structs.ACLTokenAnonymousID -> acl.AnonymousTokenID` refactor was done partly to resolve an import cycle but a broader objective is to fold more ACL-related domain objects into `acl` package.

* No behavioral changes

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/structs/acl.go
```diff
@@ -72,10 +72,6 @@ session_prefix "" {
 	policy = "write"
 }` + EnterpriseACLPolicyGlobalManagement
 
-	// This is the policy ID for anonymous access. This is configurable by the
-	// user.
-	ACLTokenAnonymousID = "00000000-0000-0000-0000-000000000002"
```
cthain, line 77: This PR only migrates the anonymous token definition for ACLs.. I don't disagree with this at all but just want to point out that now we have ACL definitions in two places, `acl/` and `structs/`.. Is it worth adding a note in the code indicating that we intend to move the rest of the ACL defs to the `acl/` package at some point?

agent/consul/acl_endpoint_test.go
```diff
@@ -161,7 +161,7 @@ func TestACLEndpoint_TokenRead(t *testing.T) {
 
 	waitForLeaderEstablishment(t, srv)
 
-	acl := ACL{srv: srv}
+	aclEp := ACL{srv: srv}
```
cthain, line 164: This change creates a lot of noise in this file, but I think this is the right thing to do given that the `consul/acl` package was already imported and these `acl` vars where hiding that package in the tests. I really like the disambiguation here.

cthain (commented):
This is looking good Chris. I just had a couple of minor comments. I like that you also did some housekeeping around comments and formatting :+1:

acl/acl.go
```diff
@@ -43,3 +47,13 @@ func (c *Config) Close() {
 		c.EnterpriseConfig.Close()
 	}
 }
+
+// CheckAnonymous returns the string "anonymous token" if
+// accessorID is acl.AnonymousTokenID. Used for better
+// UX when logging the accessorID.
+func CheckAnonymous(accessorID string) string {
```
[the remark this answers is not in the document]
kisunji: I agree that this UX improvement would be useful for other well-known tokens but my understanding based on this [doc](https://developer.hashicorp.com/consul/docs/security/acl/acl-tokens#built-in-tokens) is that the anonymous token is the only static one in our system. 

Every other "well-known" token is configured and generated at runtime, making it harder to maintain a lookup table.

Not impossible, but I'd like to consider it out of scope for this PR.

jkirschner-hashicorp:
@kisunji : My understanding is that most ACL permission denied error messages are centrally defined here: https://github.com/hashicorp/consul/blob/ee2d47da83a536f911f816b197ffb1538ddf895c/acl/errors.go#L85-L111

I don't see that error message currently modified in this PR.

And when we do modify that error message, we might want to take a different approach than is done here.

Currently, the error message is:
> Permission denied: token with AccessorID 00000000-0000-0000-0000-000000000002 lacks permission node:write on node-1

I'm not sure that replacing the accessor ID with "accessor ID" is the clearest way to communicate what's happening in this case (though I agree it's a good solution for structured logging entries like those currently included in this PR). That would look like:
> Permission denied: token with AccessorID 'anonymous token' lacks permission node:write on node-1

Perhaps it could instead be something like...
> Permission denied: the anonymous token with AccessorID 00000000-0000-0000-0000-000000000002 lacks permission node:write on node-1. The anonymous token is used implicitly when a request does not specify a token.

Or ...
> Permission denied: token with AccessorID "anonymous token" lacks permission node:write on node-1. The anonymous token is used implicitly when a request does not specify a token.

---

... I feel a bit weird about saying that the accessor ID is "anonymous token" when that isn't actually accurate, though I think it's preferable to the alternative in the case of structure logging (leaving as ID ....002). It's not like we want to encourage the lookup and modification of the anonymous token to have broader privileges ... What we want to communicate is: you didn't specify a token, but probably should. And I think making the accessor ID display as "anonymous token" is a reasonable way to accomplish that.

</rambling>

acl/acl.go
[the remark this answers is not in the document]
jkirschner-hashicorp: I agree with deferring that generalization (YAGNI principle). Once we have >1 such token (if ever), we can make it more general as suggested.

agent/structs/acl.go
kisunji: Yeah it was originally an initiative started in core before Mark A left; the intent had always been to migrate things over. I'll add a todo

kisunji: As I was looking over the file I realized that there are some import cycles that might make a general migration difficult. Unsure if a general TODO would be helpful (or if it will grow stale and unaddressed). I think just an ad-hoc refactoring over time might be better?

cthain: Ya, I'm fine with whatever you think is best. I suspect that we will be thinking about and revisiting the ACLs system over the coming months and this is not likely to fall off our radar.

skpratt (approved):

PR #15895, opened 2023-01-04
Fix missing comment in generated enterprise rate-limit file

boxofrad:
### Description
Fixes bug in #15564 where gofmt would strip out the generated code warning comment because it was on the same line as the build tag.


jmurret (approved):
🎉

PR #15897, opened 2023-01-04
Native API Gateway Config Entries

t-eckert:
### Description

This work forms the foundation of Consul-Native API Gateways. It adds the following new config entries to support these gateways:
- APIGateway
- TCPRoute
- HTTPRoute
- InlineCertificate
- BoundAPIGateway (Read only)

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/consul/gateways/controller_gateways.go
```diff
@@ -0,0 +1,32 @@
+package gateways
+
+import (
+	"context"
+	"github.com/hashicorp/consul/agent/consul/controller"
```
nathancoleman, line 5: Looks like this needs a `goimports` run. I can commit that in a moment

t-eckert: Thank you! I'll make sure I check on that before opening the PR next time.

nathancoleman (commented):
LGTM, changes have been previously reviewed in a set of incremental PRs.

Would like to confirm that @andrewstucki and @mkeeler are good with merging as the leads.

mkeeler:
@nathancoleman I have reviewing this on my TODO list for the afternoon.

agent/structs/config_entry_gateways.go
```diff
@@ -687,3 +690,300 @@ func (g *GatewayService) Clone() *GatewayService {
...
+			return fmt.Errorf("listener port %d not in the range 1-65535", listener.Port)
+		}
+		if strings.ContainsRune(strings.TrimPrefix(listener.Hostname, wildcardPrefix), '*') {
+			return fmt.Errorf("host %q is not valid, a wildcard specifier is only allowed as the left-most label", listener.Hostname)
+		}
+		for _, certificate := range listener.TLS.Certificates {
+			if !allowedCertificateKinds[certificate.Kind] {
+				return fmt.Errorf("unsupported certificate kind: %q, must be 'inline-certificate'", certificate.Kind)
+			}
+			if certificate.Name == "" {
```
mkeeler, line 803: In enterprise, will we need to validate the namespace/partition too or will we restrict to referencing certificates in the same ns/partition. 

Having give this a whole 5 seconds of consideration it seems fine to restrict to the same ns/partition as the API GW itself but maybe you know of a reason not to.

I guess for this bit of code we won't need to validate it regardless and that would be handled by the controller when binding a cert to the gw right?

agent/structs/config_entry_routes.go
```diff
@@ -0,0 +1,177 @@
...
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/consul/acl"
+)
+
+// HTTPRouteConfigEntry manages the configuration for a HTTP route
+// with the given name.
+type HTTPRouteConfigEntry struct {
```
mkeeler, line 11: Is this just a stub for where future http route details will get added?  Unlike the TCP route below there seems to be no route information in this struct currently.

Thats fine if the intention is to come in a future PR, I just wanted to ensure I wasn't missing some detail.

agent/structs/config_entry_routes.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	return nil
+}
+
+func (e *TCPRouteConfigEntry) Validate() error {
+	validParentKinds := map[string]bool{
+		APIGateway: true,
+	}
+
+	if len(e.Services) > 1 {
```
mkeeler, line 132: Is this temporary. It looks like the TCPService struct has a weight which would give the impression of allowing multiple services and traffic splitting.

agent/consul/state/config_entry_events_test.go
```diff
@@ -547,3 +547,348 @@ func TestServiceDefaultsSnapshot(t *testing.T) {
 		})
 	}
 }
+
+func TestAPIGatewaySnapshot(t *testing.T) {
+	const index uint64 = 123
+
+	ixn1 := &structs.APIGatewayConfigEntry{
```
mkeeler, line 554: Could this be refactored to maybe `gw1` for the var name (or something similar).

When looking a little further down I thought that maybe we were reusing an intention. Upon closer inspection I suspect this was copy/pasted from an intention test and its just some var names that were untouched.

The same thing goes for the test variables in the other new tests within this file.

mkeeler (commented):
Generally everything in here LGTM. I had a few questions (and other comments about potential future functionality). I would say the biggest ones are about whether we should allow linking gateways to certs in other namespaces (I don't think we should) and some around what looks like is temporary intermediate state to get this code in and iterate on it in the very near future.

agent/consul/gateways/controller_gateways.go
```diff
@@ -0,0 +1,34 @@
+package gateways
```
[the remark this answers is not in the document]
t-eckert: That makes sense as an organizing principle. We can move this once we get closer to the finished product. Thank you for pointing this out so we can move towards it.

agent/structs/config_entry_routes.go
t-eckert: Yes, Andrew is working on it. It's more complex than TCP routing because there are more configurations for filtering.

agent/structs/config_entry_gateways.go
```diff
@@ -687,3 +690,300 @@ func (g *GatewayService) Clone() *GatewayService {
...
+		if _, found := listeners[listener.Name]; found {
+			return fmt.Errorf("found multiple listeners with the name %q", listener.Name)
+		}
+		listeners[listener.Name] = struct{}{}
+	}
+	return nil
+}
+
+func (e *APIGatewayConfigEntry) validateMergedListeners() error {
+	listeners := make(map[int]APIGatewayListener)
```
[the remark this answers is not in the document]
t-eckert: Good thought, I have compiled your "future notes" in [this doc](https://docs.google.com/document/d/1SmsakmGnrmd3uI3mthefHQkDrz6GvzMrQvaHoiKu15Y/edit?usp=sharing)

agent/structs/config_entry_gateways.go
```diff
@@ -687,3 +690,300 @@ func (g *GatewayService) Clone() *GatewayService {
...
+			return fmt.Errorf("listeners %q and %q cannot be merged", merged.Name, listener.Name)
+		}
+		listeners[listener.Port] = listener
+	}
+	return nil
+}
+
+func (e *APIGatewayConfigEntry) validateListeners() error {
+	validProtocols := map[APIGatewayListenerProtocol]bool{
+		ListenerProtocolHTTP: true,
```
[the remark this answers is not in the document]
t-eckert: Added to [Matt's Thoughts About the Future](https://docs.google.com/document/d/1SmsakmGnrmd3uI3mthefHQkDrz6GvzMrQvaHoiKu15Y/edit?usp=sharing)

agent/structs/config_entry_routes.go
t-eckert: Yes, we don't currently support multiple TCP services bound to a route or multiple routes bound to a gateway listener. The spec allows for it. There was some discussion about changing this approach in the future, but currently the thought is that load balancing between TCP connections might not make sense to users because the connections are persistent.

agent/structs/config_entry_gateways.go
nathancoleman: I do believe that we have to support cross-namespace/partition certificate references since K8s Gateway API allows them today, and an Enterprise user with namespace mirroring turned on would want to be able to use this.

The pattern that the K8s Gateway API uses repeatedly is: if the namespace is unset, then assume the referenced object lives in the same namespace as the referencing object. Thoughts on using that same approach here @mkeeler ? The implication for this particular bit of code is that we wouldn't need to validate that the namespace is set, we would resolve everything at binding time as you mentioned.

mkeeler: Defaulting to the same ns as the existing object is how our other config entries work (splitters, routers, resolvers) so it seems good to continue that pattern.

In general, I think cross-ns references are probably okay. There are limitations with consul ACLs that could make that a little tricky if the ACL token used for the API GW itself is not in the default ns but we could go over that more outside of github comments.

I still think it could make sense to not allow cross-partition references to certificates. In general in order to do that we would need some new ACL enforcement mechanism. The only thing we allow cross-partition access is exported services. In that case an operator has explicitly exported a service to another partition. Unless we do something similar or bake in some support in the ACL rules themselves, cross-partition references should be prevented.

andrewstucki (approved):

agent/structs/config_entry_gateways.go
t-eckert: Added this conversation to "Matt's Thoughts About the Future": https://docs.google.com/document/d/1SmsakmGnrmd3uI3mthefHQkDrz6GvzMrQvaHoiKu15Y/edit#

I want to make sure we capture these and implement them while moving forward with this existing work.

nathancoleman: Thanks guys!

agent/structs/config_entry_gateways.go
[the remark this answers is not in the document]
mikemorris: This is fine for now, but in the Gateway API spec listeners don't have to be unique by port, but instead by the tuple of `{ hostname, port, protocol }` (ALPN isn't explicitly referenced in the spec yet, but I have heard some questions about how it might be used).

https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.Gateway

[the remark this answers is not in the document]
mikemorris: This is fine for now, but the Gateway API spec doesn't require unique ports per listener, only uniqueness on the tuple of `{ hostname, port, protocol }`, so the upstream flexibility should mostly exist for the switching Matt is describing as a future possibility.

https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.Gateway

agent/structs/config_entry_routes.go
mikemorris: I thought this was just a limitation of the current ingress gateway implementation under the hood, and something that we likely _will_ want to support once we're generating our own xDS directly.

PR #15903, opened 2023-01-04
Fix TLS_BadVerify test assertions on macOS

pglass:
### Description

On macOS, we get a different error message for an untrusted TLS certificate which causes a couple of check tests cases to fail. This fixes the assertion so that these tests pass on macOS.

```
go test ./agent/checks -run TLS_BadVerify
```

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/checks/check_test.go
+++ b/agent/checks/check_test.go
@@ -844,6 +844,9 @@ func TestCheckHTTP_TLS_BadVerify(t *testing.T) {
 
 // isInvalidCertificateError checks the error string for an untrusted certificate error.
 // The specific error message is different on Linux and macOS.
+//
+// TODO: Revisit this when https://github.com/golang/go/issues/52010 is resolved.
+// We may be able to simplify this to check only one error string.
 func isInvalidCertificateError(err string) bool {
 	return strings.Contains(err, "certificate signed by unknown authority") ||
 		strings.Contains(err, "certificate is not trusted")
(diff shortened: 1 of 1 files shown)


cthain (approved):
LGTM!

agent/checks/check_test.go
```diff
@@ -836,12 +836,19 @@ func TestCheckHTTP_TLS_BadVerify(t *testing.T) {
...
 		}
-		if !strings.Contains(notif.Output(cid), "certificate signed by unknown authority") {
+		if !isInvalidCertificateError(notif.Output(cid)) {
 			r.Fatalf("should fail with certificate error %v", notif.OutputMap())
 		}
 	})
 }
 
+// isInvalidCertificateError checks the error string for an untrusted certificate error.
+// The specific error message is different on Linux and macOS.
```
kisunji, line 846: Can we add a "this is safe to remove when https://github.com/golang/go/issues/52010 is addressed"

kisunji:
Cause: https://github.com/golang/go/issues/52010

agent/checks/check_test.go
pglass: 👍 https://github.com/hashicorp/consul/pull/15903/commits/f7c028a4c0ee27a6ad79d92ea01413c893c7acb2

kisunji (approved):

PR #15913, opened 2023-01-05
Fix issue where TLS configuration was ignored for unix sockets in consul connect envoy.

hashi-derek:
When using `consul connect envoy`, TLS configuration was not applied to local unix sockets, even when flags were specified that provided certificates.

Notably, this affects Nomad deployments, which rely on unix sockets + encryption for certain configurations.
https://github.com/hashicorp/nomad/issues/15360

Before this change, the only way TLS was enabled was if `https://` was a prefix for the `-grpc-addr` flag. This allowed users to configure the certificates via environment variables / flags on clients and slowly adopt grpc+tls over time (the `-grpc-addr` value is also discoverable via an API call to `/v1/agent/self`, but the certificate is not).

After this change, TLS will be enabled only if either of the following are true:
* `-grpc-addr` starts with `https://`
* `-grpc-addr` starts with `unix://` AND grpc certificates are configured

This ensures that the existing ability for users to roll out HTTP -> HTTPS conversions is not affected, while users of unix sockets can encrypt traffic, if desired.

Finally, there was some logic added recently that attempted to dial the gRPC connection prior to launching envoy to help users debug issues. However, this would be run during bootstrap mode, which would cause an undesired early return (Nomad generates the bootstrap on one host with Consul CLI and then passes that into a container). To fix this, the check has been swapped to a warning only (instead of returning) and the check also no longer runs during bootstrap mode.

### DIFF
--- a/.changelog/15913.txt
+++ b/.changelog/15913.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cli: Fix issue where `consul connect envoy` was unable to configure TLS over unix-sockets to gRPC.
+```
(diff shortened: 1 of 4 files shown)


wilkermichael (approved):
LGTM!

PR #15917, opened 2023-01-06
Rate limit improvements and fixes

boxofrad:
### Description
Addresses a couple of issues found while manually testing server-side rate limiting:

- Fixes a panic when `Operation.SourceAddr` is nil (internal net/rpc calls)
- Adds proper HTTP response codes (429 and 503) for rate limit errors
- Makes the error messages clearer

### DIFF
--- a/agent/http.go
+++ b/agent/http.go
@@ -29,6 +29,7 @@ import (
 	"github.com/hashicorp/consul/agent/cache"
 	"github.com/hashicorp/consul/agent/config"
 	"github.com/hashicorp/consul/agent/consul"
+	"github.com/hashicorp/consul/agent/consul/rate"
 	"github.com/hashicorp/consul/agent/structs"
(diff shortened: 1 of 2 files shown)


agent/http.go
```diff
@@ -427,39 +461,41 @@ func (s *HTTPHandlers) wrap(handler endpoint, methods []string) http.HandlerFunc
 					"error", err)
 			}
 
+			// If the error came from gRPC, unpack it to get the real message.
+			msg := err.Error()
+			if s, ok := status.FromError(err); ok {
+				msg = s.Message()
+			}
+
 			switch {
```
boxofrad, line 470: I couldn't find unit tests for the error/status code conversion, but we should probably cover it with integration tests anyway.

jmurret (approved):
@boxofrad Awesome work.  thank you for finding and fixing this stuff!

PR #15937, opened 2023-01-06
Format Raft TimeSinceLastContact as Seconds

clly:
### Description
When we format the TimeSinceLastContact as a string duration and send it to
HCP we regularly receive a 400 error when the duration is over 1 minute.
This is because the parser on HCP is parsing the value as a durationpb.
The JSON string representation of the durationpb is
"<seconds>.<nanoseconds>s". Use the durationpb json representation so
that we parse it correctly on the other end.

### Links

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/.changelog/15937.txt
+++ b/.changelog/15937.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+agent/hcp: HCP Management Plane is expecting the json representation of a durationpb, instead of a Go Duration string
+```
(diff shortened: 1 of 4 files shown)


nickethier:
Thoughts on using `protojson` to get the correct coding here instead of string manipulation?
[Example](https://go.dev/play/p/CaMYD3br1cJ)

clly:
I thought the json encoding and the possibility for an error would make it awkward to extract the value but it looks like it's just wrapped in quotations.

Do you think it'd be better to return `0s` or a negative value if we get an error for some reason?

PR #15942, opened 2023-01-09
Add ratelimit Envoy builtin extension

huikang:
### Description
- Create the built-in http ratelimit Envoy extension ([link](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/local_ratelimit/v3/local_rate_limit.proto#envoy-v3-api-msg-extensions-filters-http-local-ratelimit-v3-localratelimit) to envoy api)
- The ratelimit extension will insert an `envoy.filters.http.local_ratelimit` to the head of the inbound listener's filter chain.

Example

```
Kind = "service-defaults"
Name = "s2"
Protocol = "http"
EnvoyExtensions = [
  {
    Name = "builtin/ratelimit",
    Arguments = {
      ProxyType = "connect-proxy"
      Listener = "inbound"
      maxTokens = 20,
      tokensPerFill = 5,
      fillInterval = 30,
    }
  }
]
```

Integration test

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/structs/config_entry.go
+++ b/agent/structs/config_entry.go
@@ -59,6 +59,7 @@ var AllConfigEntryKinds = []string{
 const (
 	BuiltinAWSLambdaExtension string = "builtin/aws/lambda"
 	BuiltinLuaExtension       string = "builtin/lua"
+	BuiltinRatelimitExtension string = "builtin/ratelimit"
 )
 
 // ConfigEntry is the interface for centralized configuration stored in Raft.
@@ -316,6 +317,7 @@ func builtInExtension(name string) bool {
 	extensions := map[string]struct{}{
 		BuiltinAWSLambdaExtension: {},
 		BuiltinLuaExtension:       {},
+		BuiltinRatelimitExtension: {},
 	}
 
 	_, ok := extensions[name]
--- a/agent/xds/builtinextensions/ratelimit/copied.go
+++ b/agent/xds/builtinextensions/ratelimit/copied.go
@@ -1,0 +1,59 @@
+package ratelimit
+
+import (
+	envoy_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	envoy_listener_v3 "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
+	envoy_http_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
+	envoy_tls_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3"
+
+	"github.com/golang/protobuf/ptypes"
+
+	"github.com/golang/protobuf/proto"
+)
+
+// This is copied from xds and not put into the shared package because I'm not
+// convinced it should be shared.
+
+func makeUpstreamTLSTransportSocket(tlsContext *envoy_tls_v3.UpstreamTlsContext) (*envoy_core_v3.TransportSocket, error) {
+	if tlsContext == nil {
+		return nil, nil
+	}
+	return makeTransportSocket("tls", tlsContext)
+}
+
+func makeTransportSocket(name string, config proto.Message) (*envoy_core_v3.TransportSocket, error) {
+	any, err := ptypes.MarshalAny(config)
+	if err != nil {
+		return nil, err
+	}
+	return &envoy_core_v3.TransportSocket{
+		Name: name,
+		ConfigType: &envoy_core_v3.TransportSocket_TypedConfig{
+			TypedConfig: any,
+		},
+	}, nil
+}
+
+func makeEnvoyHTTPFilter(name string, cfg proto.Message) (*envoy_http_v3.HttpFilter, error) {
+	any, err := ptypes.MarshalAny(cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	return &envoy_http_v3.HttpFilter{
+		Name:       name,
+		ConfigType: &envoy_http_v3.HttpFilter_TypedConfig{TypedConfig: any},
+	}, nil
+}
+
+func makeFilter(name string, cfg proto.Message) (*envoy_listener_v3.Filter, error) {
+	any, err := ptypes.MarshalAny(cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	return &envoy_listener_v3.Filter{
+		Name:       name,
+		ConfigType: &envoy_listener_v3.Filter_TypedConfig{TypedConfig: any},
+	}, nil
+}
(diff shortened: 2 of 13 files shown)


agent/xds/builtinextensions/ratelimit/copied.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	envoy_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	envoy_listener_v3 "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
+	envoy_http_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
+	envoy_tls_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3"
+
+	"github.com/golang/protobuf/ptypes"
+
+	"github.com/golang/protobuf/proto"
```
kisunji, line 11: Could we use `google.golang.org/protobuf` instead? I believe @mkeeler is working on updating `github.com/goland/protobuf` references in our codebase.

agent/xds/builtinextensions/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/hashicorp/consul/agent/xds/xdscommon"
+	"github.com/hashicorp/consul/api"
+)
+
+type ratelimit struct {
+	ProxyType     string
+	Listener      string
+	MaxTokens     int
+	TokensPerFill int
+	FillInterval  int
```
kisunji, line 31: nit: since `ratelimit` is not exported, I don't think we need the fields capitalized

agent/xds/builtinextensions/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,167 @@
...
+		resultErr = multierror.Append(resultErr, fmt.Errorf("fillInterval is required"))
+	}
+
+	if plugin.MaxTokens == 0 {
+		resultErr = multierror.Append(resultErr, fmt.Errorf("MaxTokens must be greater than 0"))
+	}
+
+	if plugin.TokensPerFill == 0 {
+		resultErr = multierror.Append(resultErr, fmt.Errorf("TokensPerFill must be greater than 0"))
+	}
```
kisunji, line 59: According to envoy's docs `TokensPerFill` defaults to 1 when not specified. Do we want this field explicit in Consul anyways?

agent/xds/builtinextensions/ratelimit/copied.go
```diff
@@ -0,0 +1,59 @@
+package ratelimit
```
erichaberkorn, line 1: We've reached the magical number 3 where we should probably just move these into xdscommon.

agent/xds/builtinextensiontemplate/template.go
```diff
@@ -166,6 +166,8 @@ func (envoyExtension EnvoyExtension) patchTerminatingGatewayListener(config xdsc
 			if ok {
 				filters = append(filters, newFilter)
 				patched = true
+			} else {
+				filters = append(filters, filter)
```
kisunji, line 170: If `err != nil` then we'll be calling `filters = append(filters, filter)` twice, right?
I think the `err`+`ok` check needs to be more robust for all cases here to prevent possible bugs

agent/xds/builtinextensions/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,167 @@
...
+
+	if err := validateProxyType(plugin.ProxyType); err != nil {
+		resultErr = multierror.Append(resultErr, err)
+	}
+
+	if err := validateListener(plugin.Listener); err != nil {
+		resultErr = multierror.Append(resultErr, err)
+	}
```
erichaberkorn, line 67: It looks like the local rate limit filter supports a bunch of additional fields. Could you give some info on why we aren't supporting those fields?

agent/structs/config_entry.go
```diff
@@ -59,6 +59,7 @@ var AllConfigEntryKinds = []string{
 const (
 	BuiltinAWSLambdaExtension string = "builtin/aws/lambda"
 	BuiltinLuaExtension       string = "builtin/lua"
+	BuiltinRatelimitExtension string = "builtin/ratelimit"
```
erichaberkorn, line 62: Should we add local to this name to distinguish it from global rate limiting?

agent/xds/builtinextensions/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// PatchCluster does nothing.
+func (p ratelimit) PatchCluster(c *envoy_cluster_v3.Cluster) (*envoy_cluster_v3.Cluster, bool, error) {
+	return c, false, nil
+}
+
+// PatchFilter inserts a http local rate_limit filter at the head of
+// envoy.filters.network.http_connection_manager filters
+func (p ratelimit) PatchFilter(filter *envoy_listener_v3.Filter) (*envoy_listener_v3.Filter, bool, error) {
```
kisunji, line 93: General question (maybe more directed for @erichaberkorn). Is it the programmer's responsibility to order extensions so they insert themselves in the right place in the chain?

What happens if two filters conflict when trying to be first in a chain?

agent/xds/builtinextensions/ratelimit/copied.go
mkeeler: A PR has been merged to remove any usage of golang/protobuf and the linter will now complain about importing them.

agent/xds/builtinextensions/ratelimit/ratelimit.go
erichaberkorn: We handle these issues in a couple of ways:
* Extensions should error out when patching resources that don't look as expected.
* We will add support for post validations to extensions. This would handle cases where an extension applies correctly and a later extension messes up the Envoy resources.

agent/xds/builtinextensions/ratelimit/ratelimit.go
cthain: :thinking: I believe that the fields must be exported in order to use `mapstructure.Decode` on this struct: https://github.com/hashicorp/consul/blob/33f6edef2a1fbd40bae2026f18e68f4071985c3f/agent/xds/builtinextensions/ratelimit/ratelimit.go#L45

kisunji: Ah yes, thats for pointing that out!

huikang:
Moved to https://github.com/hashicorp/consul-enterprise/pull/4206

PR #15943, opened 2023-01-09
upgrade test: test service-intention in single cluster scenario

huikang:
### Description
This PR adds service-intention upgrade test in single cluster scenario: `TestBadauthz_UpgradeToTarget_fromLatest` verifies service-intention continue functioning after upgrade.

Additional changes include:
- A single cluster topology creation function: `BasicSingleClusterTopology` that can be used by single cluster test
- Restart connect sidecar during upgrade (`consul connect` will exit when the consul agent is restarted. So we need to restart the sidecar and upgrade the mapped app port)
- combine agent and cluster package to reduce import cycle (This aligns with RB's test PR in ent)
- Move `meshgateway` out of creating cluster helper function to avoid import cycle

### PR Checklist

* [x] updated test coverage
* [ ] not a security concern

### DIFF
--- a/test/integration/consul-container/libs/cluster/cluster.go
+++ b/test/integration/consul-container/libs/cluster/cluster.go
@@ -16,8 +16,8 @@ import (
 	"github.com/testcontainers/testcontainers-go"
 
 	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/consul/command/helpers"
 	"github.com/hashicorp/consul/sdk/testutil/retry"
-	libagent "github.com/hashicorp/consul/test/integration/consul-container/libs/agent"
 )
 
 // Cluster provides an interface for creating and controlling a Consul cluster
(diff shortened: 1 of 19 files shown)


test/integration/consul-container/libs/cluster/cluster.go
```diff
@@ -341,6 +341,27 @@ func (c *Cluster) PeerWithCluster(acceptingClient *api.Client, acceptingPeerName
...
+	entries := client.ConfigEntries()
+	written := false
+	written, _, err = entries.Set(entry, nil)
+	if err != nil {
+		return fmt.Errorf("error set config entry: %v", err)
+	}
+	if !written {
+		return fmt.Errorf("config entry not updated: %s/%s", entry.GetKind(), entry.GetName())
+	}
+	return err
```
NiniOak, line 362: I think you should return `nil` here instead of err. Potential errors has been handled above in L350, 357 and 360

test/integration/consul-container/libs/cluster/container.go
```diff
@@ -267,6 +273,16 @@ func (c *consulContainerNode) Upgrade(ctx context.Context, config Config) error
 		})
 	}
 
+	// make sure all affiliated envoy sidecar are running
+	for _, sidecar := range c.connectSidecars {
+		// TODO: remove the hardcoded client's upstream port number
+		//   the number is the upstream LocalBindPort of the static-client
+		//   registered to consul.
```
NiniOak, line 280: Did you forget this TODO?

PR #15945, opened 2023-01-09
rate: add prometheus definitions, docs, and clearer names

boxofrad:
### Description
While testing the server-side rate limiting work, I spotted that we were missing docs and a Prometheus definition for the counter that is incremented whenever a limit is exceeded. Also renamed the counter for better clarity/consistency.


website/content/docs/agent/telemetry.mdx
```diff
@@ -477,6 +477,7 @@ These metrics are used to monitor the health of the Consul servers.
...
+| `consul.rpc.rate_limit.exceeded`                    | Increments whenever an RPC is over a configured rate limit. Note: in permissive mode, the RPC will have still been allowed to proceed.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             | RPCs                              | counter |
```
trujillo-adam, line 480:
```suggestion
| `consul.rpc.rate_limit.exceeded`                    | Increments whenever an RPC is over a configured rate limit. When Consul is in `permissive` mode, the RPC is allowed to proceed.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             | RPCs                              | counter |
```
Actually, what is in permissive mode in this context? I assumed that `permissive` is a value that you you can define in some configuration, but I'm not sure. Also, we should just use simple present tense per the style guide.

trujillo-adam (commented):
A few questions and suggestions for consistency with the style guide.

website/content/docs/agent/telemetry.mdx
boxofrad: Thanks for fixing my terrible writing! 😆 ❤️  

In this context, the rate limiter is in permissive mode, and there can be multiple limiters involved in a single RPC (e.g. global, ip-based, tenant-based) which are configured independently.

jmurret (approved):
👍 🎉

PR #15949, opened 2023-01-09
Protobuf Modernization

mkeeler:
### Description

golang/protobuf is deprecated. While the main code generator hasn't changed much, the libraries for dealing with well known types  (Any, Struct, Duration and Timestamp) have greatly improved. This PR intends to remove the older versioned libraries usage and make it clear to developers which packages and facilities should be used (as opposed to having our code with examples of both ways)

* Remove direct usage of golang/protobuf in favor of google.golang.org/protobuf
* Marshallers (protobuf and json) needed some changes to account for different APIs.
* Moved to using the google.golang.org/protobuf/types/known/* for the well known types including replacing some custom Struct manipulation with whats available in the structpb well known type package.
* This also updates our devtools script to install protoc-gen-go from the right location so that files it generates conform to the correct interfaces.

### Links

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


dhiaayachi (approved):
LGTM!

PR #15950, opened 2023-01-09
Implement BindRoutesToGateways

t-eckert:
### Description

When gateways or routes updates are reconciled, the routes which should be bound to particular "BoundAPIGateways" must be updated. This PR adds `BindRoutesToGateways` does this work.

* This PR adds unit tests in the `agent/structs` and `agent/consul/gateways`

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/consul/gateways/bind.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+	}
+
+	return modified, errored, nil
+}
+
+// BindRouteToGateways takes a slice of bound API gateways and a route.
+// It iterates over the parent references for the given route. These parents are gateways the
+// route should be bound to. If the parent matches a bound gateway, the route is bound to the
+// gateway. Otherwise, the route is unbound from the gateway if it was bound.
```
nathancoleman, line 46: I think this helps clarify the meaning a bit since we would never bind and then immediately unbind in the same function call
```suggestion
// gateway. Otherwise, the route is unbound from the gateway if it was previously bound.
```

agent/consul/gateways/bind.go
```diff
@@ -0,0 +1,98 @@
...
+func BindRouteToGateways(gateways []*structs.BoundAPIGatewayConfigEntry, route structs.BoundRoute) ([]*structs.BoundAPIGatewayConfigEntry, map[structs.ResourceReference]error, error) {
+	// errored stores the errors from events where a resource reference failed to bind to a gateway.
+	errored := make(map[structs.ResourceReference]error)
+
+	parentRefs, gatewayRefs := getReferences(route)
+
+	// Iterate over all BoundAPIGateway config entries and try to bind them to the route if they are a parent.
+	modified := make([]*structs.BoundAPIGatewayConfigEntry, 0, len(gateways))
+	for _, gateway := range gateways {
+		if references, routeReferencesGateway := gatewayRefs[configentry.NewKindNameForEntry(gateway)]; routeReferencesGateway {
```
nathancoleman, line 60: #nit this is about the length where this format for the conditional gets hard to read for me

```suggestion
		references, routeReferencesGateway := gatewayRefs[configentry.NewKindNameForEntry(gateway)]
		if routeReferencesGateway {
```

agent/consul/gateways/bind.go
```diff
@@ -0,0 +1,98 @@
...
+type gatewayRefs = map[configentry.KindName][]structs.ResourceReference
+
+// BindRoutesToGateways takes a slice of bound API gateways and a slice of routes.
+// It iterates over the parent references for each route. These parents are gateways the
+// route should be bound to. If the parent matches a bound gateway, the route is bound to
+// the gateway. Otherwise, the route is unbound from the gateway if it was bound.
+//
+// The function returns a list of references to the modified BoundAPIGatewayConfigEntry objects,
+// a map of resource references to errors that occurred when they were attempted to be
+// bound to a gateway, and an error if the overall process was unsucessful.
```
nathancoleman, line 23: #nit I would probably just comment that this is a convenience function for calling `BindRouteToGateways` with multiple routes; otherwise, I typically see that anyone modifying the other function will forget to update this doc.

nathancoleman: One other thought: is there value in having two separate functions - `BindRoute` and `BindRoutes` - vs. just one function that accepts any number of routes?
```go
func BindRoutesToGateways([]*structs.BoundAPIGatewayConfigEntry, routes ...structs.BoundRoute)
```

agent/consul/gateways/bind.go
```diff
@@ -0,0 +1,98 @@
...
+func BindRouteToGateways(gateways []*structs.BoundAPIGatewayConfigEntry, route structs.BoundRoute) ([]*structs.BoundAPIGatewayConfigEntry, map[structs.ResourceReference]error, error) {
+	// errored stores the errors from events where a resource reference failed to bind to a gateway.
+	errored := make(map[structs.ResourceReference]error)
+
+	parentRefs, gatewayRefs := getReferences(route)
+
+	// Iterate over all BoundAPIGateway config entries and try to bind them to the route if they are a parent.
+	modified := make([]*structs.BoundAPIGatewayConfigEntry, 0, len(gateways))
```
nathancoleman, line 58: Love seeing this initialized to the correct size! This is something that frequently gets overlooked 🎉

agent/consul/gateways/bind.go
```diff
@@ -0,0 +1,98 @@
...
+
+// BindRouteToGateways takes a slice of bound API gateways and a route.
+// It iterates over the parent references for the given route. These parents are gateways the
+// route should be bound to. If the parent matches a bound gateway, the route is bound to the
+// gateway. Otherwise, the route is unbound from the gateway if it was bound.
+//
+// The function returns a list of references to the modified BoundAPIGatewayConfigEntry objects,
+// a map of resource references to errors that occurred when they were attempted to be
+// bound to a gateway, and an error if the overall process was unsucessful.
+func BindRouteToGateways(gateways []*structs.BoundAPIGatewayConfigEntry, route structs.BoundRoute) ([]*structs.BoundAPIGatewayConfigEntry, map[structs.ResourceReference]error, error) {
```
nathancoleman, line 51: This is sorta hard to understand IMHO, trying to differentiate between a route-specific "error" - which really just maps to a status update saying why a given route could not bind - and an overall process error that means we couldn't do our job for one reason or another.

When I dug deeper to try to rephrase a bit, I realized that we never actually return that third `error` value. Could we / should we just remove it? I think it would help simplify/clarify this function.

```suggestion
// The function returns a list of references to the modified BoundAPIGatewayConfigEntry objects,
// a map of resource references to errors that occurred when they were attempted to be
// bound to a gateway, and an error if the overall process was unsucessful.
func BindRouteToGateways(gateways []*structs.BoundAPIGatewayConfigEntry, route structs.BoundRoute) ([]*structs.BoundAPIGatewayConfigEntry, map[structs.ResourceReference]error) {
```

nathancoleman: If you do wind up removing, we can simplify `BindRoutesToGateways` as well

nathancoleman (commented):
First pass. Nice work! Just some nit-picky interface type stuff

agent/structs/config_entry_gateways.go
```diff
@@ -980,10 +980,130 @@ func (e *BoundAPIGatewayConfigEntry) GetEnterpriseMeta() *acl.EnterpriseMeta {
...
+	}
+
+	didBind := false
+	for i, listener := range e.Listeners {
+		if listener.Name == ref.SectionName || ref.SectionName == "" {
+			if listener.BindRoute(route) {
+				didBind = true
+				e.Listeners[i] = listener
+			}
+		}
```
andrewstucki, line 1032: Somewhere in this conditional or in the `listener.BindRoute` implementation we should be checking that the protocol of the route aligns with the protocol type of the listener itself, i.e. `ListenerProtocolHTTP`-based listeners should only accept `HTTPRoute`s or error if binding, `ListenerProtocolTCP` should only accept `TCPRoute`.

andrewstucki (changes requested):
Agree with @nathancoleman that if we can't think of a use-case for returning an error to force re-reconciliation from `BindRoutesToGateways` we should probably delete it.

One more comment about checking the alignment of Route types with the listener protocol that will require a change. Apart from that this looks good to me.

agent/structs/config_entry_gateways.go
nathancoleman: @t-eckert and I discussed this outside of this PR last week and this morning. It sounds like it would require some additional modifications to have the protocol readily accessible

> **Thomas Eckert**
> There will be some things to double back on that where we do the binding currently regardless of the protocol of the route matching the protocol of the listener, but I wanted to unblock everyone
> 
> And I want to talk to Andrew about adding the protocol to the BoundListener struct because without it we have to pause and go check the store every time we want to bind.

andrewstucki: Sounds good, I chatted a bit with Thomas already and think we're good with just throwing a TODO here and circling back on a subsequent lightweight refactor to add the protocol checking code.

My main thing is that I don't want to bloat the object we're actually persisting for state to avoid some complicated deprecation/refactoring/migrations later, so I'd prefer not to store the protocol on the `BoundListener`. Instead, I'm thinking we might want to do some sort of tuple wrapper similar to what we have in our k8s controller code, where we have an intermediate `K8sGateway` that wraps both our computed state and the actual Gateway configuration and abstracts away a bit of the interaction between the two. But, that can be left for the follow-up.

agent/consul/gateways/bind.go
t-eckert: 😊 Thanks

agent/consul/gateways/bind.go
t-eckert: This was an AWESOME suggestion. It is now variadic

agent/consul/gateways/bind.go
t-eckert: Awesome! I removed it

agent/structs/config_entry_gateways.go
t-eckert: Added the TODO

andrewstucki (approved):

PR #15956, opened 2023-01-11
[OSS] Add Peer field to service-defaults upstream overrides

hashi-derek:
The [upstream overrides in service-defaults config entries](https://developer.hashicorp.com/consul/docs/connect/config-entries/service-defaults#overrides) do not have a Peer field, so it is impossible to distinguish between configuration that should apply for a local-cluster upstream and a peer-upstream. 

For example, an upstream override for svc1 would apply to both of the following upstreams:

```
connect {
    sidecar_service {
      proxy {
        upstreams = [
          {
            destination_name = "svc1"
            destination_peer = "peer-dc2"
            local_bind_port  = 5000
          },
          {
            destination_name = "svc1"
            local_bind_port  = 5001
          }
        ]
      }
    }
  }
```

This is not ideal, because users may not have control over what services are named in the remote peer cluster (they could overlap), which means we need to provide a way for them to distinguish between the two and provide different configurations for each.

Because this is existing behavior in 1.14 that users may already be relying on for configuring peer upstreams, this fix should only be applied in 1.15, where it is acceptable to make breaking-changes.


DanStough:
@mkeeler OSS version of the PR you've already looked at if I can get your 👍 .

mkeeler (approved):

PR #15958, opened 2023-01-11
removed Consul requirements from documentation as it may be confusing

wilkermichael:
### Description
The config entry documentation currently has Consul requirements listed that are much greater than n-2 releases back, and having the Consul requirements in multiple locations may be confusing for users. As such I've removed them from the config entry documentation.

I've also fixed some formatting issues.

N/A

### Links
<I'll add the vercel link once it's built>

### PR Checklist

* [n/a] updated test coverage
* [x] not a security concern


trujillo-adam:
_Is_ it confusing to state the minimum version of Consul required to use the configuration entry? I'm in favor of removing temporal language sprinkled throughout the docs that might confuse users, but I think this is one context where it makes sense to refer to previous versions. IMO, we should standardize on specifying minimum versions necessary for using the functionality in Requirements sections . . . and only Requirements sections. 

@im2nguyen @boruszak - care to chime in? Did we adopt the same policy as support perspective and decide to only refer to the two most recent versions?

wilkermichael:
> _Is_ it confusing to state the minimum version of Consul required to use the configuration entry? I'm in favor of removing temporal language sprinkled throughout the docs that might confuse users, but I think this is one context where it makes sense to refer to previous versions. IMO, we should standardize on specifying minimum versions necessary for using the functionality in Requirements sections . . . and only Requirements sections.
> 
> @im2nguyen @boruszak - care to chime in? Did we adopt the same policy as support perspective and decide to only refer to the two most recent versions?

I think the confusing part, is we don't version the fields in the config entry, so although a given version is the minimum to use the config entry, it isn't the minimum version to use the fields as they're written.

trujillo-adam:
Ugh, I had a good response and lost when I clicked away from stupid GH.
I think we're talking about two different things. What I think you are saying is that the configuration entry itself--the fields and values--doesn't increment versions as a Helm chart would, for instance. As a result, the version number might be confusing to users. Is that correct?
What I'm saying is that the configuration entry isn't even available in Consul 1.10 or older, so trying to define and apply the configuration entry won't work. But I guess we have versioned docs, so maybe stating the min version is superfluous. I guess I'm thinking about the use cases where someone with an older version of Consul is looking for information and could be confused because we always default to the latest version of the docs. I suppose that's kind of a corner case, though.

wilkermichael:
> Ugh, I had a good response and lost when I clicked away from stupid GH. I think we're talking about two different things. What I think you are saying is that the configuration entry itself--the fields and values--doesn't increment versions as a Helm chart would, for instance. As a result, the version number might be confusing to users. Is that correct? What I'm saying is that the configuration entry isn't even available in Consul 1.10 or older, so trying to define and apply the configuration entry won't work. But I guess we have versioned docs, so maybe stating the min version is superfluous. I guess I'm thinking about the use cases where someone with an older version of Consul is looking for information and could be confused because we always default to the latest version of the docs. I suppose that's kind of a corner case, though.

Yeah, I was also thinking that the versioned docs would cover this case, you wouldn't necessarily expect the latest docs to work with Consul 1.11 for example, which wouldn't understand the notion of say a `envoyExtension` which is a new field we're adding to ServiceDefaults in 1.15

david-yu:
I'm personally fine with removing version requirements since we now have versioned docs, especially if the versions are extremely outdated. This should only go in the 1.14.x branch and not back ported to other branches.

trujillo-adam:
I'm good with leaning into the versioned docs and removing the min version requirement. It puts the onus on us to make sure we are tagging changes and back porting correctly.

trujillo-adam (approved):

david-yu (approved):

PR #15960, opened 2023-01-12
Integration test for server rate limiting

jmurret:
### Description
Adding in an integration test to assert that the following effects occur:
- error messages returned to clients
- trace logs when limits exceed
- warning logs when limits blocked
- metrics when limits exceeded

This is currently only for HTTP/ net-RPC.

### DIFF
--- a/agent/consul/rate/handler.go
+++ b/agent/consul/rate/handler.go
@@ -230,7 +230,13 @@ func (h *Handler) Allow(op Operation) error {
 		})
 
 		if enforced {
-			// TODO(NET-1382) - use the logger to print rate limiter logs.
+			h.logger.Warn("RPC blocked due to exceeded allowed rate limit",
+				"rpc", op.Name,
+				"source_addr", op.SourceAddr,
+				"limit_type", l.desc,
+				"limit_enforced", enforced,
+			)
+
 			if h.leaderStatusProvider.IsLeader() && op.Type == OperationTypeWrite {
 				return ErrRetryLater
 			}
--- a/agent/consul/rate/handler.go
+++ b/agent/consul/rate/handler.go
@@ -207,7 +207,7 @@ func (h *Handler) Allow(op Operation) error {
 		// TODO(NET-1382): is this the correct log-level?
 
 		enforced := l.mode == ModeEnforcing
-		h.logger.Trace("RPC exceeded allowed rate limit",
+		h.logger.Warn("RPC exceeded allowed rate limit",
 			"rpc", op.Name,
 			"source_addr", op.SourceAddr,
 			"limit_type", l.desc,
@@ -230,13 +230,6 @@ func (h *Handler) Allow(op Operation) error {
 		})
 
 		if enforced {
-			h.logger.Warn("RPC blocked due to exceeded allowed rate limit",
-				"rpc", op.Name,
-				"source_addr", op.SourceAddr,
-				"limit_type", l.desc,
-				"limit_enforced", enforced,
-			)
-
 			if h.leaderStatusProvider.IsLeader() && op.Type == OperationTypeWrite {
 				return ErrRetryLater
 			}
--- a/test/integration/consul-container/test/ratelimit/ratelimit_test.go
+++ b/test/integration/consul-container/test/ratelimit/ratelimit_test.go
@@ -37,7 +37,47 @@ func TestRateLimit(t *testing.T) {
 
 	testCases := []testCase{
 		{
-			dsc: "Mode: enforcing",
+			dsc: "Mode: disabled - does not return errors / does not emit metrics",
+			cmd: "-hcl=limits { request_limits { mode = \"disabled\" read_rate = 0 write_rate = 0 }}",
+			operations: []testOperation{
+				{
+					action: func(client *api.Client) error {
+						_, _, err := client.Catalog().Nodes(&api.QueryOptions{})
+						return err
+					},
+					expectedErrorMsg: "",
+				},
+				{
+					action: func(client *api.Client) error {
+						_, err := utils.ApplyDefaultProxySettings(client)
+						return err
+					},
+					expectedErrorMsg: "",
+				},
+			},
+		},
+		{
+			dsc: "Mode: permissive - does not return error / emits metrics",
+			cmd: "-hcl=limits { request_limits { mode = \"permissive\" read_rate = 0 write_rate = 0 }}",
+			operations: []testOperation{
+				{
+					action: func(client *api.Client) error {
+						_, _, err := client.Catalog().Nodes(&api.QueryOptions{})
+						return err
+					},
+					expectedErrorMsg: "",
+				},
+				{
+					action: func(client *api.Client) error {
+						_, err := utils.ApplyDefaultProxySettings(client)
+						return err
+					},
+					expectedErrorMsg: "",
+				},
+			},
+		},
+		{
+			dsc: "Mode: enforcing - returns errors / emits metrics",
 			cmd: "-hcl=limits { request_limits { mode = \"enforcing\" read_rate = 0 write_rate = 0 }}",
 			operations: []testOperation{
 				{
@@ -70,13 +110,17 @@ func TestRateLimit(t *testing.T) {
 				if len(op.expectedErrorMsg) > 0 {
 					require.Error(t, err)
 					require.Equal(t, op.expectedErrorMsg, err.Error())
+				} else {
+					require.NoError(t, err)
 				}
 			}
-			metrics, err := client.Agent().Metrics()
-			require.NoError(t, err)
-			for counter := range metrics.Counters {
-				t.Logf("Counter: %+v", counter)
-			}
+
+			// currently returns NaN error
+			//			metrics, err := client.Agent().Metrics()
+			//			require.NoError(t, err)
+			//			for counter := range metrics.Counters {
+			//				t.Logf("Counter: %+v", counter)
+			//			}
 		})
 	}
 }
--- a/test/integration/consul-container/test/ratelimit/ratelimit_test.go
+++ b/test/integration/consul-container/test/ratelimit/ratelimit_test.go
@@ -1,9 +1,12 @@
 package ratelimit
 
 import (
+	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
+	"github.com/testcontainers/testcontainers-go"
 
 	"github.com/hashicorp/consul/api"
 	libcluster "github.com/hashicorp/consul/test/integration/consul-container/libs/cluster"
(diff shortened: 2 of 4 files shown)


test/integration/consul-container/test/ratelimit/ratelimit_test.go
```diff
@@ -0,0 +1,232 @@
...
+					if strings.Contains(msg, httpRequest) {
+						found = true
+					}
+				}
+				require.True(t, found, fmt.Sprintf("Log not found for: %s", httpRequest))
+			}
+
+			metricInfo, err := client.Agent().Metrics()
+			// TODO(NET-1978): currently returns NaN error
+			//			require.NoError(t, err)
```
jmurret, line 163: Some times this returns a 500 (NaN error).  Is there any context about calling metrics endpoints that would be helpful during troubleshooting this error?

test/integration/consul-container/test/ratelimit/ratelimit_test.go
```diff
@@ -0,0 +1,232 @@
...
+
+			// validate logs
+			for _, httpRequest := range urlsExpectingLogging {
+				found := false
+				for _, msg := range logConsumer.Msgs {
+					if strings.Contains(msg, httpRequest) {
+						found = true
+					}
+				}
+				require.True(t, found, fmt.Sprintf("Log not found for: %s", httpRequest))
```
jmurret, line 157: This sometimes occurs with `/v1/config PUT` where there is no logged messaged for it. Is there anything about `GetClient()` that would return a connection to a leader vs a follower that might cause this error to sometimes appear vs not?

agent/consul/rate/handler.go
```diff
@@ -230,7 +230,13 @@ func (h *Handler) Allow(op Operation) error {
 		})
 
 		if enforced {
-			// TODO(NET-1382) - use the logger to print rate limiter logs.
+			h.logger.Warn("RPC blocked due to exceeded allowed rate limit",
```
jmurret, line 233: I did this work logging previously but I must have lost it in a merge conflict somewhere.

test/integration/consul-container/test/ratelimit/ratelimit_test.go
```diff
@@ -0,0 +1,261 @@
...
+		},
+		rateLimitOperation: "KVS.Apply",
+		rateLimitType:      "global/write",
+	}
+
+	testCases := []testCase{
+		// HTTP & net/RPC
+		{
+			description: "HTTP & net/RPC / Mode: disabled - errors: no / exceeded logs: no / blocked logs: no / metrics: no",
+			cmd:         "-hcl=limits { request_limits { mode = \"disabled\" read_rate = 0 write_rate = 0 }}",
```
boxofrad, line 69: nit: we can avoid needing to escape those quotes by using backticks:

```suggestion
			cmd:         `-hcl=limits { request_limits { mode = "disabled" read_rate = 0 write_rate = 0 }}`,
```

test/integration/consul-container/test/ratelimit/ratelimit_test.go
```diff
@@ -0,0 +1,261 @@
...
+					expectedErrorMsg:  "",
+					expectExceededLog: false,
+					expectBlockedLog:  false,
+					expectMetric:      false,
+				},
+			},
+		},
+		{
+			description: "HTTP & net/RPC / Mode: permissive - errors: no / exceeded logs: yes / blocked logs: no / metrics: yes",
+			cmd:         "-hcl=limits { request_limits { mode = \"permissive\" read_rate = 0 write_rate = 0 }}",
```
boxofrad, line 89:
```suggestion
			cmd:         `-hcl=limits { request_limits { mode = "permissive" read_rate = 0 write_rate = 0 }}`,
```

test/integration/consul-container/test/ratelimit/ratelimit_test.go
```diff
@@ -0,0 +1,261 @@
...
+					expectedErrorMsg:  "",
+					expectExceededLog: true,
+					expectBlockedLog:  false,
+					expectMetric:      false,
+				},
+			},
+		},
+		{
+			description: "HTTP & net/RPC / Mode: enforcing - errors: yes / exceeded logs: yes / blocked logs: yes / metrics: yes",
+			cmd:         "-hcl=limits { request_limits { mode = \"enforcing\" read_rate = 0 write_rate = 0 }}",
```
boxofrad, line 109:
```suggestion
			cmd:         `-hcl=limits { request_limits { mode = "enforcing" read_rate = 0 write_rate = 0 }}`,
```

agent/consul/rate/handler.go
boxofrad: Do we need this log line? We already have this one, which includes whether or not the limit was enforced:
https://github.com/hashicorp/consul/blob/b8a850c2a658c32666a5f42d3fef045feab57df8/agent/consul/rate/handler.go#L210-L215

Logging twice feels like it'll bloat the log (and drop buffer) unnecessarily. Although, we may want to increase the log-level of the other one.

test/integration/consul-container/test/ratelimit/ratelimit_test.go
```diff
@@ -0,0 +1,261 @@
...
+						op.action.rateLimitOperation, "blocked")
+				}
+			}
+		})
+	}
+}
+
+func checkForMetric(t *testing.T, metricsInfo *api.MetricsInfo, operationName string, expectedLimitType string) {
+	for _, counter := range metricsInfo.Counters {
+		if counter.Name == "consul.rate.limit" {
```
boxofrad, line 184: Note: this will change when #15945 is merged.

boxofrad (commented):
This looks great, amazing work @jmurret 🙌🏻 

Can totally happen in a follow-up PR (just wanted to make sure it's captured) but it'd be good to have a test that spins up 2 servers, makes an HTTP request to the follower, and checks the limit is enforced on the leader correctly.

agent/consul/rate/handler.go
jmurret: I added this because the comment was there to do so, but when I was thinking about this the questions/thoughts that I was thinking of (which I would love input on the answers are)?
- is trace the right level?
- if i am exceeding levels but its permissive, trace seems ok.  
- if I am dropping requests, I probably want info at a minimum and possibly warnings (IE, should I have to turn on tracing to know that my servers are blocking requests?)
- issuing warnings on the other hand for permissive might seem excessive?  may be not?
- should this be an
  - if enforced, issue a warning
  - else if permissive issue trace

jmurret:
Thanks @boxofrad for the suggestions and the question.  Related to the follow up PR, we already have a different backlogged ticket for that test.  👍

test/integration/consul-container/test/ratelimit/ratelimit_test.go
```diff
@@ -0,0 +1,261 @@
...
+type TestLogConsumer struct {
+	Msgs []string
+}
+
+func (g *TestLogConsumer) Accept(l testcontainers.Log) {
+	g.Msgs = append(g.Msgs, string(l.Content))
+}
+
+// createCluster
+func createCluster(t *testing.T, cmd string, logConsumer *TestLogConsumer, logLevel string) *libcluster.Cluster {
```
dhiaayachi, line 228: We have a `createCluster` func in `test/integration/consul-container/test/basic/connect_service_test.go`, I'm wondering if we can unify both and extract it from the test file?

dhiaayachi (commented):
Great work @jmurret!! I left few comments for your consideration.

test/integration/consul-container/test/ratelimit/ratelimit_test.go
```diff
@@ -0,0 +1,261 @@
...
+				if op.expectExceededLog {
+					checkLogsForMessage(t, logConsumer.Msgs,
+						fmt.Sprintf("[TRACE] agent.server.rpc-rate-limit: RPC exceeded allowed rate limit: rpc=%s", op.action.rateLimitOperation),
+						op.action.rateLimitOperation, "exceeded")
+				}
+				if op.expectBlockedLog {
+					checkLogsForMessage(t, logConsumer.Msgs,
+						fmt.Sprintf("[WARN]  agent.server.rpc-rate-limit: RPC blocked due to exceeded allowed rate limit: rpc=%s", op.action.rateLimitOperation),
+						op.action.rateLimitOperation, "blocked")
+				}
```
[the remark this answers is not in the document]
jmurret: Implemented in this commit: https://github.com/hashicorp/consul/pull/15960/commits/071f1093d06e2e38439608206b91c490ff2e8943

test/integration/consul-container/test/ratelimit/ratelimit_test.go
```diff
@@ -0,0 +1,261 @@
...
+					require.Equal(t, op.expectedErrorMsg, err.Error())
+				} else {
+					require.NoError(t, err)
+				}
+			}
+
+			// validate logs and metrics
+			// doing this in a separate loop so we can perform actions, allow metrics
+			// and logs to collect and then assert on each.
+			for _, op := range tc.operations {
```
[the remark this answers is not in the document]
jmurret: great suggestion.  thank you.  

added here: https://github.com/hashicorp/consul/pull/15960/commits/23de54f3b1f860e876de9c727f0876b35323c1c2

test/integration/consul-container/test/ratelimit/ratelimit_test.go
jmurret: I think I am going to backlog a ticket so we can get Xinyi or Poonam to work on this.  It would be helpful to get them oriented to these integration tests.

agent/consul/rate/handler.go
dhiaayachi: I think a warning log in both cases will be needed. The reasoning is that permissive mode will be used by operators as a mean to configure rate-limiting and having to activate the trace logs to do so would:
- Add to the operator burden to configure rate limiting
- Tracing would introduce an extra performance hit that could lead to requests being rate limited.

jmurret: Thanks @dhiaayachi and @boxofrad.  It seems like just warning once regardless of permissive vs enforcing is the way to go and the message contains limited=true/false?  This is what I did in https://github.com/hashicorp/consul/pull/15960/commits/0a98872947fa77bcb6ff443d97f92e25d720be20

Let me know what you think.

dhiaayachi (approved):
LGTM! 🚀

test/integration/consul-container/test/ratelimit/ratelimit_test.go
```diff
@@ -0,0 +1,248 @@
...
+		}
+	}
+
+}
+
+func checkLogsForMessage(t *retry.R, logs []string, msg string, operationName string, logType string, logShouldExist bool) {
+	found := false
+	for _, log := range logs {
+		if strings.Contains(log, msg) {
+			found = true
```
boxofrad, line 198: Optional: it doesn't really matter because `logs` isn't huge, but once you've found `msg` you can break the loop.
 
```suggestion
			found = true
			break
```

test/integration/consul-container/test/ratelimit/ratelimit_test.go
```diff
@@ -0,0 +1,248 @@
...
+		}
+	}
+	require.Equal(t, logShouldExist, found, fmt.Sprintf("%s log check failed for: %s. Log expected: %t", logType, operationName, logShouldExist))
```
boxofrad, line 201: I'm guessing you didn't use `require.Contains` here because it'd output too much noise?

boxofrad (approved):
Looks great, nice work! 👏🏻

test/integration/consul-container/test/ratelimit/ratelimit_test.go
jmurret: I did not use it because each log in the collection of messages contains a timestamp and need to match on just a substring of an individual message.  At least that was my understanding of the limitiations.  if there is an easier way, I'm all for it.

PR #15967, opened 2023-01-12
Warn if ACL is enabled but no token is provided to Envoy

kisunji:
### Description
When ACLs are enabled, the CLI command `consul connect envoy` requires a valid ACL token for the [sidecar proxy](https://developer.hashicorp.com/consul/tutorials/security/access-control-setup-production#for-service-mesh-add-the-token-to-the-sidecar-proxy) to function. However, the command may not error in the absence of a token since the [anonymous token](https://developer.hashicorp.com/consul/docs/security/acl/acl-tokens#anonymous-token) fallback will usually be sufficient for `consul connect envoy` to complete execution.

This PR adds a call to check if ACLs are enabled then outputs a warning if ACLs are enabled but no token was provided.

This PR also adds a metric to count the number of unauthenticated streams in xDS.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/xds/server.go
```diff
@@ -195,6 +199,9 @@ func (s *Server) authenticate(ctx context.Context) (acl.Authorizer, error) {
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "error fetching options from context: %v", err)
 	}
+	if options.Token == "" {
+		metrics.IncrCounter([]string{"xds", "server", "unauthenticated"}, 1)
```
rboyer, line 203: Would it make more sense to handle this in the `activeStreamCounter` as a gauge since an individual stream may call `authenticate` or `authorize` multiple times over its lifetime and would likely be double/triple/etc counted?

website/content/docs/agent/telemetry.mdx
```diff
@@ -544,6 +544,7 @@ These metrics are used to monitor the health of the Consul servers.
...
+| `consul.xds.server.unauthenticated`                 | Counts the number of xDS streams that are unauthenticated because ACLs are not enabled or ACL tokens were missing.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           | ms                                | timer   |
```
rboyer, line 547: `ms/timer` is not correct as implemented (with the counter) and would be more correct as `streams?/counter`

if you address the gauge suggestion then this should be `streams/gauge`

kisunji: Woops good catch! Copy paste error. Will probably set it to `streams/gauge` with your prior suggestion.

agent/xds/server.go
kisunji: Yeah that makes sense to me. Will figure out how to carry over the logic

command/connect/envoy/envoy.go
```diff
@@ -612,6 +613,16 @@ func (c *cmd) generateConfig() ([]byte, error) {
 
 	var bsCfg BootstrapConfig
 
+	// Make a call to an arbitrary ACL endpoint. If we get back an ErrNotFound
+	// (meaning ACLs are enabled) check that the token is not empty.
+	if _, _, err := c.client.ACL().TokenReadSelf(
+		&api.QueryOptions{Token: args.Token},
+	); acl.IsErrNotFound(err) {
+		if args.Token == "" {
+			c.UI.Warn("ACL is enabled but no token was provided.")
```
cizh, line 622: "ACL is enabled but no token was provided." sounds good and describes the situation. I wonder if we can describe the potential cost& suggest fix as well. Maybe something like **"ACL is enabled but no token was provided. Please pass the correct token to Envoy to avoid potential access denial and communication failure."**? 
*Should it be "pass **a** token" or "pass **the** token"? I wish to be clear that they need the right token for the ACL they set up and make them think "a token? any token's fine?" or "the token? which token?". At the same time, I don't want to stress user with too many details like another doc to read&learn @jkirschner-hashicorp wdyt?
*Not sure if I described the consequences precisely, please feel free to edit. If possible, shorten "access denial and communication failure" this part would be cool.

cizh: Maybe "ACL is enabled but no token was provided. To avoid potential access denial and communication failure, please pass the authorized token to Envoy."

agent/grpc-external/metadata.go
```diff
@@ -4,11 +4,28 @@ import (
...
 	"google.golang.org/grpc/metadata"
+
+	"github.com/hashicorp/consul/agent/structs"
 )
 
+const metadataKeyToken = "x-consul-token"
+
+// TokenFromContext returns the ACL token in the gRPC metadata attached to the
+// given context.
+func TokenFromContext(ctx context.Context) string {
```
kisunji, line 17: Reverting a deletion from https://github.com/hashicorp/consul/pull/14724/files#diff-b4bf50db93d52f639b211a36e1e38f0e6d47a166ba03842d84c0a0251811e42a

Using `mapstructure` to extract the token seemed a bit inefficient for xDS

agent/xds/server.go
```diff
@@ -23,26 +23,30 @@ import (
...
-	},
-}
+var (
+	StatsGauges = []prometheus.GaugeDefinition{
+		{
+			Name: []string{"xds", "server", "streams"},
+			Help: "Measures the number of active xDS streams handled by the server split by protocol version.",
+		},
+		{
+			Name: []string{"xds", "server", "streamsUnauthenticated"},
```
kisunji, line 33: Does this read better as `unauthenticatedStreams`?

I thought it would be nicer to close to the related `streams` gauge when sorted alphabetically.

agent/xds/server.go
```diff
@@ -126,32 +130,34 @@ type Server struct {
...
-// ensure alignment is correct. This further requires that activeStreamCounters
-// be a pointer field.
-// TODO(eculver): refactor to remove xDSv2 refs
+// activeStreamCounters tracks various stream-related metrics.
+// Requires that activeStreamCounters be a pointer field.
 type activeStreamCounters struct {
-	xDSv3 uint64
-	xDSv2 uint64
+	xDSv3           atomic.Uint64
+	unauthenticated atomic.Uint64
```
kisunji, line 137: no need for manual alignment with these types:
>[Int64](https://tip.golang.org/pkg/sync/atomic/#Int64) and [Uint64](https://tip.golang.org/pkg/sync/atomic/#Uint64) are automatically aligned to 64-bit boundaries in structs and allocated data, even on 32-bit systems.

src: https://tip.golang.org/doc/go1.19

website/content/docs/agent/telemetry.mdx
```diff
@@ -541,6 +541,7 @@ These metrics are used to monitor the health of the Consul servers.
...
+| `consul.xds.server.streamsUnauthenticated`          | Measures the number of active xDS streams handled by the server that are unauthenticated because ACLs are not enabled or ACL tokens were missing.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  | streams                           | gauge   |
```
pglass, line 544: What does this gauge look like if ACLs are enabled. Will we have any (long-lived) xDS streams that are unauthenticated? Or would this basically be a gauge of the attempts to open an xDS stream that are unauthenticated?

agent/grpc-external/metadata.go
kisunji: Relatively speaking mapstructure is expensive but it's still in the realm of nanoseconds so certainly not too expensive.
```go
func BenchmarkTokenFromContext(b *testing.B) {
	md := metadata.MD{}
	md.Set(metadataKeyToken, "9df2d1a4-2d07-414e-8ead-6053f56ed2eb")
	ctx := metadata.NewOutgoingContext(context.Background(), md)

	var token string
	for i := 0; i < b.N; i++ {
		token = TokenFromContext(ctx)
	}
	_ = token
}

func BenchmarkQueryOptionsFromContext(b *testing.B) {
	md := metadata.MD{}
	md.Set(metadataKeyToken, "9df2d1a4-2d07-414e-8ead-6053f56ed2eb")
	ctx := metadata.NewOutgoingContext(context.Background(), md)

	var token string
	for i := 0; i < b.N; i++ {
		o, _ := QueryOptionsFromContext(ctx)
		token = o.Token
	}
	_ = token
}
```

```
BenchmarkQueryOptionsFromContext-16    	18966411	        61.19 ns/op	      96 B/op	       1 allocs/op
BenchmarkTokenFromContext-16    	165877292	         7.099 ns/op	       0 B/op	       0 allocs/op
```

I don't mind reusing QueryOptionsFromContext

website/content/docs/agent/telemetry.mdx
kisunji: Whether ACLs are on or off this gauge will track streams that do not have tokens associated with them. I'm not entirely sure if any unauthenticated streams will be long-lived, unless the anonymous token is privileged enough to continue operating.

pglass (approved):

PR #15968, opened 2023-01-12
avoid logging RPC errors when it's specific rate limiter errors

dhiaayachi:
### Description
By default every failed RPC operation will be logged in error level. RPC rate limiter have already a log implementation for rate limited requests (debug level). In this PR we exclude the rate limiting errors from the logs to avoid overloading the system by adding too much log.


agent/consul/rpc.go
```diff
@@ -419,10 +419,12 @@ func (s *Server) handleConsulConn(conn net.Conn) {
 
 		if err := s.rpcServer.ServeRequest(rpcCodec); err != nil {
 			if err != io.EOF && !strings.Contains(err.Error(), "closed") {
-				s.rpcLogger().Error("RPC error",
-					"conn", logConn(conn),
-					"error", err,
-				)
+				if !errors.Is(err, rate.ErrRetryLater) && !errors.Is(err, rate.ErrRetryElsewhere) {
```
boxofrad, line 422: It looks like we already have an `if` to filter out noisy error types on the line above, should we collapse it into one?

```Go
skipLog := errors.Is(err, io.EOF) ||
	errors.Is(err, rate.ErrRetryLater) ||
	errors.Is(err, rate.ErrRetryElsewhere) ||
	strings.Contains(err.Error(), "closed")

if !skipLog {
  // ...
}
```

boxofrad (approved):
Looks good! I've left one small suggestion though 🙇🏻

jmurret (approved):

agent/consul/rpc.go
dhiaayachi: @boxofrad, @jmurret 
I changed that to simplify if while keeping the same behaviour. Please give it another 👀

jmurret (approved):

PR #15969, opened 2023-01-12
 Add envoy assertion in Golang

NiniOak:
### Description
This PR adds support to validate Envoy Listeners in Golang. This functionality already exists in bash, it has been refactored and the Envoy listener assertion has been added to the existing `BasicConnectService` test. 

### PR Checklist

* [x] updated test coverage
* [ ] not a security concern

### DIFF
--- a/test/integration/consul-container/libs/assert/envoy.go
+++ b/test/integration/consul-container/libs/assert/envoy.go
@@ -1,0 +1,96 @@
+package assert
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/hashicorp/consul/sdk/testutil/retry"
+	libservice "github.com/hashicorp/consul/test/integration/consul-container/libs/service"
+	"github.com/hashicorp/consul/test/integration/consul-container/libs/utils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// GetEnvoyListenerTCPFilters validates that proxy was configured with tcp protocol and one rbac listener filter
+func GetEnvoyListenerTCPFilters(t *testing.T, adminPort int) {
+	var (
+		dump string
+		err  error
+	)
+	failer := func() *retry.Timer {
+		return &retry.Timer{Timeout: 30 * time.Second, Wait: 1 * time.Second}
+	}
+
+	retry.RunWith(failer(), t, func(r *retry.R) {
+		dump, err = libservice.GetEnvoyConfigDump(adminPort, "")
+		if err != nil {
+			r.Fatal("could not fetch envoy configuration")
+		}
+	})
+
+	// The services configured for the tests have proxy tcp protocol configured, therefore the HTTP request is on tcp protocol
+	// the steps below validate that the json result from envoy config dump returns active listener with rbac and tcp_proxy configured
+	filter := `.configs[2].dynamic_listeners[].active_state.listener | "\(.name) \( .filter_chains[0].filters | map(.name) | join(","))"`
+	results, err := utils.JQFilter(dump, filter)
+	require.NoError(t, err, "could not parse envoy configuration")
+
+	if len(results) != 2 {
+		require.Error(t, fmt.Errorf("s1 proxy should have been configured with one rbac listener filter. Got %d listener(s)", len(results)))
+	}
+
+	var filteredResult []string
+	for _, result := range results {
+		santizedResult := sanitizeResult(result)
+		filteredResult = append(filteredResult, santizedResult...)
+	}
+
+	require.Contains(t, filteredResult, "envoy.filters.network.rbac")
+	require.Contains(t, filteredResult, "envoy.filters.network.tcp_proxy")
+}
+
+// GetEnvoyHTTPrbacFilters validates that proxy was configured with an http connection manager
+// this assertion is currently unused current tests use http protocol
+func GetEnvoyHTTPrbacFilters(t *testing.T, port int) {
+	var (
+		dump string
+		err  error
+	)
+	failer := func() *retry.Timer {
+		return &retry.Timer{Timeout: 30 * time.Second, Wait: 1 * time.Second}
+	}
+
+	retry.RunWith(failer(), t, func(r *retry.R) {
+		dump, err = libservice.GetEnvoyConfigDump(port, "")
+		if err != nil {
+			r.Fatal("could not fetch envoy configuration")
+		}
+	})
+
+	// the steps below validate that the json result from envoy config dump configured active listeners with rbac and http filters
+	filter := `.configs[2].dynamic_listeners[].active_state.listener | "\(.name) \( .filter_chains[0].filters[] | select(.name == "envoy.filters.network.http_connection_manager") | .typed_config.http_filters | map(.name) | join(","))"`
+	results, err := utils.JQFilter(dump, filter)
+	require.NoError(t, err, "could not parse envoy configuration")
+
+	if len(results) != 2 {
+		require.Error(t, fmt.Errorf("s1 proxy should have been configured with one rbac listener filter. Got %d listener(s)", len(results)))
+	}
+
+	var filteredResult []string
+	for _, result := range results {
+		sanitizedResult := sanitizeResult(result)
+		filteredResult = append(filteredResult, sanitizedResult...)
+	}
+	require.Contains(t, filteredResult, "envoy.filters.http.rbac")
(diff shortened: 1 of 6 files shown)


test/integration/consul-container/libs/assert/envoy.go
```diff
@@ -0,0 +1,96 @@
...
+
+	retry.RunWith(failer(), t, func(r *retry.R) {
+		dump, err := libservice.GetEnvoyConfigDump(port, "")
+		if err != nil {
+			r.Fatal("could not curl envoy configuration")
+		}
+
+		// Make sure s1 proxy has been been configured with http rbac filters
+		filter := `.configs[2].dynamic_listeners[].active_state.listener | "\(.name) \( .filter_chains[0].filters[] | select(.name == "envoy.filters.network.http_connection_manager") | .typed_config.http_filters | map(.name) | join(","))"`
+		results, err := utils.JQFilter(dump, filter)
```
nfi-hashicorp, line 67: Tangent, but I find this function deeply weird :P

test/integration/consul-container/libs/assert/envoy.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+
+		// Make sure there is one rbac listener listener configured
+		filter := `.configs[2].dynamic_listeners[].active_state.listener | "\(.name) \( .filter_chains[0].filters | map(.name))"`
+		results, err := utils.JQFilter(dump, filter)
+		if err != nil {
+			r.Fatal("could not parse envoy configuration")
+		}
+
+		if len(results) != 2 {
```
nfi-hashicorp, line 35: Just double-checking: we really want to do the JSON querying inside the retry?

test/integration/consul-container/libs/assert/envoy.go
```diff
@@ -0,0 +1,96 @@
...
+// GetEnvoyListenerFilters validates that proxy was configured with one rbac listener filter
+func GetEnvoyListenerFilters(t *testing.T, adminPort int) {
+	failer := func() *retry.Timer {
+		return &retry.Timer{Timeout: 30 * time.Second, Wait: 1 * time.Second}
+	}
+
+	retry.RunWith(failer(), t, func(r *retry.R) {
+		dump, err := libservice.GetEnvoyConfigDump(adminPort, "")
+		if err != nil {
+			r.Fatal("could not curl envoy configuration")
```
nfi-hashicorp, line 25: nit:
```suggestion
			r.Fatal("could not fetch envoy configuration")
```

test/integration/consul-container/libs/assert/envoy.go
```diff
@@ -0,0 +1,96 @@
...
+			r.Fatal("%s", "could not parse envoy configuration")
+		}
+
+		if len(results) != 2 {
+			r.Fatal("s1 proxy should have been configured with one rbac listener filter and %d present", len(results))
+		}
+
+		// validate public listeners value
+		var filteredResult []string
+		rgx := regexp.MustCompile(`\[(.*?)\]`)
```
nfi-hashicorp, line 78: I guess this is why I find `utils.JQFilter` weird. I'm not super familiar with `gojq`, but it seems like it can give you real, typed results, not just strings, so you could use that to be a little more precise.

But again, just matching the style of the existing code, so 🤷

nfi-hashicorp (approved):
Definitely an improvement over BATS

test/integration/consul-container/libs/assert/envoy.go
NiniOak: Totally agree with this sentiment, We should definitely look into using a different library to unmarshall the api body

test/integration/consul-container/go.mod
```diff
@@ -61,7 +61,7 @@ require (
 	github.com/opencontainers/go-digest v1.0.0 // indirect
 	github.com/opencontainers/image-spec v1.0.3-0.20211202183452-c5a74bcca799 // indirect
 	github.com/opencontainers/runc v1.1.4 // indirect
-	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
+	github.com/pmezard/go-difflib v1.0.0 // indirect
```
huikang, line 64: is this a downgrade?

NiniOak: Seems like it. fixed by running go mod tidy. Good catch :)

test/integration/consul-container/libs/assert/envoy.go
```diff
@@ -0,0 +1,102 @@
...
+	retry.RunWith(failer(), t, func(r *retry.R) {
+		dump, err = libservice.GetEnvoyConfigDump(adminPort, "")
+		if err != nil {
+			r.Fatal("could not fetch envoy configuration")
+		}
+	})
+	// Make sure there is one rbac listener listener configured
+	filter := `.configs[2].dynamic_listeners[].active_state.listener | "\(.name) \( .filter_chains[0].filters | map(.name))"`
+	results, err := utils.JQFilter(dump, filter)
+	if err != nil {
```
huikang, line 35: nit: can we replace this with `require.NoError`?

test/integration/consul-container/libs/assert/envoy.go
```diff
@@ -0,0 +1,102 @@
...
+		t.Fatal("could not parse envoy configuration")
+	}
+
+	if len(results) != 2 {
```
huikang, line 39: same here, `require.Equal`

test/integration/consul-container/libs/assert/envoy.go
```diff
@@ -0,0 +1,102 @@
...
+		t.Fatalf("s1 proxy should have been configured with one rbac listener filter. Got %d listener(s)", len(results))
+	}
+
+	// validate public listeners value
+	var filteredResult []string
+	rgx := regexp.MustCompile(`\[(.*?)\]`)
+	for _, result := range results {
+		foundString := rgx.FindStringSubmatch(result)
+		sanitizedResult := sanitizeResult(foundString[1])
```
huikang, line 48: To avoid panic, add require.True(t, len(foundString)>=2)

test/integration/consul-container/libs/assert/envoy.go
```diff
@@ -0,0 +1,102 @@
...
+	results, err := utils.JQFilter(dump, filter)
+	if err != nil {
+		t.Fatal("could not parse envoy configuration")
+	}
+
+	if len(results) != 2 {
+		t.Fatalf("s1 proxy should have been configured with one rbac listener filter. Got %d listener(s)", len(results))
+	}
+
+	// validate public listeners value
```
huikang, line 43: Could you elaborate a bit on why the following code  validates public listeners value? Thanks.

test/integration/consul-container/libs/assert/envoy.go
```diff
@@ -0,0 +1,102 @@
...
+		sanitizedResult := sanitizeResult(foundString[1])
+		filteredResult = append(filteredResult, sanitizedResult...)
+	}
+
+	require.Contains(t, filteredResult, "envoy.filters.network.rbac")
+	assert.Contains(t, filteredResult, "envoy.filters.network.tcp_proxy")
+}
+
+// GetEnvoyHTTPrbacFilters validates that proxy was configured with http rbac filters
+func GetEnvoyHTTPrbacFilters(t *testing.T, port int) {
```
huikang, line 57: Looks like this function isn't used. Shall we call it somewhere?

test/integration/consul-container/test/peering/rotate_server_and_ca_then_fail_test.go
```diff
@@ -192,8 +192,10 @@ func verifySidecarHasTwoRootCAs(t *testing.T, sidecar libservice.Service) {
 	}
 
 	retry.RunWith(failer(), t, func(r *retry.R) {
-		dump, err := libservice.GetEnvoyConfigDump(adminPort)
-		require.NoError(r, err, "could not curl envoy configuration")
+		dump, err := libservice.GetEnvoyConfigDump(adminPort, "include_eds")
+		if err != nil {
+			r.Fatal("could not curl envoy configuration")
+		}
```
huikang, line 198: is there any reason to use `if err != nil {` rather than `equire.NoError(r, err, "could not curl envoy configuration")`

test/integration/consul-container/libs/assert/envoy.go
NiniOak: Yes this function isn't used because I couldn't find the right test to add it. Perhaps we can take a look at the test suite together to find where envoy http assertion is required

huikang (approved):

PR #15970, opened 2023-01-12
Support Vault agent auth config for AWS/GCP CA provider auth

cthain:
### Description
This PR addresses #11921 and adds support for supplying an existing Vault agent auto-auth configuration as the provider configuration when using the Vault CA provider. Support has been added for the AWS and GCP auth methods.

The changes have been made such that backwards compatibility is maintained for existing AWS/GCP CA auth configurations.

* Unit tests have been added and updated.
 
### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/connect/ca/provider_vault_auth.go
```diff
@@ -0,0 +1,239 @@
...
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	return client.Logical().WriteWithContext(ctx, c.LoginPath, loginData)
+}
+
+var _ VaultAuthenticator = (*gcp.GCPAuth)(nil)
+
+// NewGCPAuthClient returns a VaultAuthenticator that can log into Vault using the GCP auth method.
+func NewGCPAuthClient(authMethod *structs.VaultAuthMethod) (VaultAuthenticator, error) {
```
pglass, line 84: nit: It would be nice to put the GCP stuff here in its own file, and maybe the AWS stuff in its own file. I guess, `provider_vault_auth_gcp.go`, `provider_vault_auth_aws.go`? What do you think?

agent/connect/ca/provider_vault.go
```diff
@@ -976,57 +975,60 @@ func vaultLogin(client *vaultapi.Client, authMethod *structs.VaultAuthMethod) (*
...
 			serviceAccountToken, err := os.ReadFile(defaultK8SServiceAccountTokenPath)
 			if err != nil {
-				return "", err
+				return nil, err
 			}
 
 			authMethod.Params["jwt"] = string(serviceAccountToken)
 		}
-		loginPath = fmt.Sprintf("auth/%s/login", authMethod.MountPath)
+		return NewVaultAPIAuthClient(authMethod, loginPath), nil
```
pglass, line 1000: non-blocking: You don't have to do this here, but for organizational purposes, I wouldn't mind if we added functions for the other auth-method types that have non-default config: `NewK8sAuthClient`, `NewOCIAuthClient`, `NewAuthClientWithUsername`(?) - where those could check/set params and call `NewVaultAPIAuthClient`.

agent/connect/ca/provider_vault_auth_test.go
```diff
@@ -0,0 +1,301 @@
...
+				require.Contains(t, err.Error(), c.expErr.Error())
+				return
+			}
+
+			require.NoError(t, err)
+			require.NotNil(t, auth)
+
+			if c.isExplicit {
+				// in this case a JWT is provided so we'll call the login API directly using a VaultAuthClient.
+				_ = auth.(*VaultAuthClient)
```
pglass, line 85: nit: We could use [`require.IsType`](https://pkg.go.dev/github.com/stretchr/testify/require#IsType) for these type assertions

pglass (approved):

agent/connect/ca/provider_vault_auth.go
```diff
@@ -0,0 +1,239 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("aws auth failed to generate login data: %w", err)
+	}
+	return loginData, nil
+}
+
+// NewCredentialsConfig creates an awsutil.CredentialsConfig from the given set of parameters.
+// It is mostly copied from the awsutil package because that package does not currently provide
+// an easy way to configure all the available options.
+func NewCredentialsConfig(config map[string]interface{}) (*awsutil.CredentialsConfig, string, error) {
```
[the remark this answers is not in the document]
kisunji: Or if this is an implementation detail, we could simply unexport it?

agent/connect/ca/provider_vault_auth.go
```diff
@@ -0,0 +1,239 @@
...
+type VaultAuthClient struct {
+	// AuthMethod holds the configuration for the Vault auth method login.
+	AuthMethod *structs.VaultAuthMethod
+	// LoginPath is optional and can be used to explicitly set the API path that the client
+	// will use for a login request. If it is empty the path will be derived from AuthMethod.MountPath.
+	LoginPath string
+	// LoginDataGen derives the parameter map containing the data for the login API request.
+	// For some auth methods this is needed to transform the AuthMethod.Params into the login data
+	// needed for the API request.
+	LoginDataGen LoginDataGenerator
```
kisunji, line 43: My understanding is that we either use `AuthMethod.Params` directly or transform it with `LoginDataGen` to ultimately produce a `map[string]interface{}`.

Could this simply be `func(*structs.VaultAuthMethod) map[string]any` with a default that returns `AuthMethod.Params`? The `LoginDataGenerator` feels like a level of indirection which isn't entirely necessary.

agent/connect/ca/provider_vault_auth.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+// NewVaultAPIAuthClient creates a VaultAuthClient that uses the Vault API to perform a login.
+func NewVaultAPIAuthClient(authMethod *structs.VaultAuthMethod, loginPath string) *VaultAuthClient {
+	if loginPath == "" {
+		mount := authMethod.Type
+		if authMethod.MountPath != "" {
+			mount = authMethod.MountPath
+		}
```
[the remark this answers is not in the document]
cthain: Yes, good call. I think it is fair to assume that the caller sets the `MountPath` correctly.

agent/connect/ca/provider_vault.go
cthain: Yes I fully agree. I've added a new issue to our backlog for this.

agent/connect/ca/provider_vault_auth_test.go
cthain: Great suggestion!

agent/connect/ca/provider_vault_auth.go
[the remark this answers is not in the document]
cthain: Good call. I've added AWS and unexported it.

agent/connect/ca/provider_vault_auth.go
cthain: Yes, I like this suggestion. We don't really need an `interface` for this. I've updated the type to a func definition as suggested.

PR #15976, opened 2023-01-13
docs: Migrate link formats

ashleemboyer:
## What

- Migrates the formats of links in Markdown content as part of the [Docs Content Link Rewrites project](https://docs.google.com/document/d/1WaSyvoVPS-YCNiSPX0ynGpvc1gySEcpbrRoQVhimQCA/edit)
- Adds two new GitHub workflow files:
  - `.github/workflows/check-legacy-links-format.yml` - This is a workflow that checks if links in markdown content are in the new format. It does this by running the link rewrite script from `dev-portal`, and then asserting that no "rewriteable" links were found. This workflow will be used on subsequent PRs to maintain the new link format.
  - `.github/workflows/test-link-rewrites.yml` - This is a workflow that executes e2e tests of this PR's changes against the `main` branch's deploy preview. The tests only run for branches named `docs/amb.migrate-link-formats`, so future PRs will not have the tests run against them. This is a temporary workflow.
  
## PRs that came before this as pre-work

- https://github.com/hashicorp/consul/pull/15999
- https://github.com/hashicorp/consul/pull/16038

## Reviewing

With the two new workflows in place, this PR should only need a spot-check.

## Testing

- [ ] The latest run of the "Legacy Link Format Checker" workflow for this branch should pass ([check here](https://github.com/hashicorp/consul/actions/workflows/check-legacy-links-format.yml))
- [ ] The latest run of the "Test Link Rewrites" workflow for this branch should pass ([check here](https://github.com/hashicorp/consul/actions/workflows/test-link-rewrites.yml))

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203453044319621


website/content/api-docs/catalog.mdx
```diff
@@ -59,17 +59,17 @@ The table below shows this endpoint's support for
...
-  see the [Service - Agent API](/api-docs/agent/service) page
+  see the [Service - Agent API](/consul/api-docs/agent/service) page
   as registering services differs between using this or the Services Agent endpoint.
 
 - `Check` `(Check: nil)` - Specifies to register a check. The register API
   manipulates the health check entry in the Catalog, but it does not setup the
   script, TTL, or HTTP check to monitor the node's health. To truly enable a new
   health check, the check must either be provided in agent configuration or set
-  via the [agent endpoint](agent).
+  via the [agent endpoint](https://consul.io/agent).
```
kendallstrautman, line 72: 🎏  Need to look into this one

kendallstrautman: Looks like it should be relative

kendallstrautman: Looks like on prod its the same behavior, a broken link. Will note as something to fix as a follow up

ashleemboyer: This is an instance of a "no-dots" link. The way those are handled, is that it looks for the URL (`agent` in this case) within the page path (`/api-docs/cataglog` in this case). So here, because `agent` is not in the page/file path, the link is getting prefixed with the .io hostname. This can be categorized as a content bug and since the bug is present in prod as well, it can be fixed in a PR before or after this one.

topfunky (approved):
👍🏻

im2nguyen (approved):
Just did a spot check and this looks great. Thanks! 🚀

PR #15988, opened 2023-01-17
Add a flag for enabling debug logs to the `connect envoy` command

kyhavlov:
This PR adds a flag, `-enable-config-gen-logging`, to the `connect envoy` command to display some log messages when generating the bootstrap config. It can be hard to troubleshoot slowness that happens during the multiple api requests during this command due to the current lack of logs, so this flag enables outputting some messages (on stderr) to help with that.

### DIFF
--- a/website/content/docs/connect/proxies/envoy.mdx
+++ b/website/content/docs/connect/proxies/envoy.mdx
@@ -124,6 +124,12 @@ Envoy requires an initial bootstrap configuration file. You can either create th
 
 Connect to a local Consul client agent and run the [`consul connect envoy` command](/consul/commands/connect/envoy) to create the Envoy bootstrap configuration. The command either outputs the bootstrap configuration directly to stdout or generates the configuration and issues an `exec` command to the Envoy binary as a convenience wrapper. For more information about using `exec` to bootstrap Envoy, refer to [Exec Security Details](/consul/commands/connect/envoy#exec-security-details).
 
+If you experience issues when bootstrapping Envoy proxies from the CLI, use the
+`-enable-config-gen-logging` flag to enable debug message logging. These logs can
+help you troubleshoot issues during Envoy bootstrapping. For more information about
+available flags and parameters, refer to the
+[`consul connect envoy CLI` reference](/commands/connect/envoy).
+
 ### Generate the bootstrap file from Consul Dataplane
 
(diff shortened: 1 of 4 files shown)


david-yu:
@kyhavlov Is there a way to somehow expose this to Consul K8s via Helm? Also I assume for `consul-dataplane` things will work differently or is this also used but the `consul-dataplane` code and need to be registered as well there?

kyhavlov:
@david-yu I'm not sure what the best way to expose this in k8s is - is there any way to set cli flags or env vars in the sidecar via helm? Consul-dataplane is totally separate so this wouldn't affect the process there.

@boruszak The log messages are basically just checkpoints to show how long things are taking and what's happened so far. The impetus for adding this was a case where the sidecar container was taking several minutes to start up and generate the config and there wasn't a way to see where the slowness was coming from - we suspected it might be a specific API call or something like that and used these logs to track down the source. 

I'm not sure where it belongs on the '[Envoy Proxy Configuration for Service Mesh](https://developer.hashicorp.com/consul/docs/connect/proxies/envoy)' page, since that page already links to the CLI and it's a pretty CLI-specific thing.

thisisnotashwin (approved):
From a `consul-k8s` perspective, we can pass this flag on to the `consul connect envoy` command if a user enables a loglevel of debug in helm. Would that potentially help this situation @david-yu ? i can take a stab at it.

david-yu:
@thisisnotashwin That would be very helpful, thank you!

website/content/docs/connect/proxies/envoy.mdx
```diff
@@ -124,6 +124,12 @@ Envoy requires an initial bootstrap configuration file. You can either create th
 
 Connect to a local Consul client agent and run the [`consul connect envoy` command](/consul/commands/connect/envoy) to create the Envoy bootstrap configuration. The command either outputs the bootstrap configuration directly to stdout or generates the configuration and issues an `exec` command to the Envoy binary as a convenience wrapper. For more information about using `exec` to bootstrap Envoy, refer to [Exec Security Details](/consul/commands/connect/envoy#exec-security-details).
 
+If you experience issues when bootstrapping Envoy proxies from the CLI, use the
+`-enable-config-gen-logging` flag to enable debug message logging. These logs can
+help you troubleshoot issues during Envoy bootstrapping. For more information about
+available flags and parameters, refer to the
+[`consul connect envoy CLI` reference](/commands/connect/envoy).
```
kendallstrautman, line 131:
```suggestion
[`consul connect envoy CLI` reference](/consul/commands/connect/envoy).
```
Hi 👋🏻 Now that the [link migration](https://github.com/hashicorp/consul/pull/15976) has been released, I'm going through open PRs to resolve any issues. This is the new format that should be used for inline docs links. Let me know if you have questions!

PR #15997, opened 2023-01-17
[OSS] chore(ci): fix compat ent compat tests for sidecars and gateways

DanStough:
### Description
Should fix issues with missing `consul:local` image used as the base for sidecars/gateways. This image is not available in ENT pipelines. Use the same image that was used to create the cluster.


thisisnotashwin (approved):
TYSM

PR #16000, opened 2023-01-18
Fix configuration merging for implicit tproxy upstreams.

hashi-derek:
This PR fixes an issue where tproxy would not correctly use service-defaults and proxy-defaults for implicit upstreams. Normally, we handle this by taking configuration from an auto-generated wildcard `*` upstream, but that was not done in all cases -- notably peer upstreams.

The changes here are essentially:

1. Change the merging logic so that the wildcard upstream has correct proxy-defaults and service-defaults values combined into it. It did not previously merge all fields, and the wildcard upstream did not exist unless service-defaults existed (it ignored proxy-defaults, essentially).
2. Change the way we fetch upstream configuration in the xDS layer so that it falls back to the wildcard when no matching upstream is found. This is what allows implicit peer upstreams to have the correct "merged" config.
3. Change proxycfg to always watch local mesh gateway endpoints whenever a peer upstream is found. This simplifies the logic so that we do not have to inspect the "merged" configuration on peer upstreams to extract the mesh gateway mode.

It's worth noting that another large issue still needs to be resolved in a different PR (https://github.com/hashicorp/consul/pull/15956), where the `ServiceDefaults.UpstreamConfig.Overrides[].Peer` field does NOT exist. This manifests in two ways after these changes:
1. Explicit (non-tproxy) peer upstreams incorrectly have overrides applied to them, since the functions in `agent/configentry/merge_service_config.go` and `agent/configentry/resolve.go` are oblivious to the concept of a peer.
2. Implicit (tproxy) peer upstreams do not have overrides applied to them, because the UID that proxycfg expects is `default/default/mysvc?peer=cluster-02`, but the overrides UID is `default/default/mysvc`. This is technically correct behavior, but it's noted here, because overrides will NOT work for peer upstreams in tproxy mode until the `agent/configentry` package is updated.

### DIFF
--- a/.changelog/16000.txt
+++ b/.changelog/16000.txt
@@ -1,3 +1,3 @@
 ```release-note:breaking-change
-connect: Fix configuration merging for transparent proxy upstreams. Proxy-defaults and service-defaults config entries were not correctly merged for implicit upstreams in transparent proxy mode and would result in some configuration not being applied. To avoid issues when upgrading, ensure that any proxy-defaults or service-defaults have the correct information for upstreams, as this configuration will become active.
+connect: Fix configuration merging for transparent proxy upstreams. Proxy-defaults and service-defaults config entries were not correctly merged for implicit upstreams in transparent proxy mode and would result in some configuration not being applied. To avoid issues when upgrading, ensure that any proxy-defaults or service-defaults have correct configuration for upstreams, since all fields will now be properly used to configure proxies.
 ```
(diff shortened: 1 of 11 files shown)


.changelog/16000.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:breaking-change
+connect: Fix configuration merging for transparent proxy upstreams. Proxy-defaults and service-defaults config entries were not correctly merged for implicit upstreams in transparent proxy mode and would result in some configuration not being applied. To avoid issues when upgrading, ensure that any proxy-defaults or service-defaults have the correct information for upstreams, as this configuration will become active.
+```
```
ndhanushkodi, line 3: > To avoid issues when upgrading, ensure that any proxy-defaults or service-defaults have the correct information for upstreams, as this configuration will become active.

Is this part referencing the wildcard upstream config that you're ensuring gets created? I guess I'm not sure what "as this configuration will become active" means so just wondering if we can reword a bit?

PR #16001, opened 2023-01-18
[NET-1728] Updating docs to remove use of token query parameter

roncodingenthusiast:
### Description
- Updating docs to remove the use of token query parameters

### PR Checklist

* [x] not a security concern

### DIFF
--- a/website/content/api-docs/api-structure.mdx
+++ b/website/content/api-docs/api-structure.mdx
@@ -40,9 +40,9 @@ $ curl \
     http://127.0.0.1:8500/v1/agent/members
 ```
 
-Previously this was provided via a `?token=` query parameter. This functionality
-exists on many endpoints for backwards compatibility, but its use is **highly
+Previously this was provided via a `?token=` query parameter but its use is **highly
 discouraged**, since it can show up in access logs as part of the URL.
+This functionality is deprecated and will be removed in future versions.
 
 To learn more about the ACL system read the [documentation](/docs/security/acl).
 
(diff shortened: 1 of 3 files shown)


website/content/api-docs/api-structure.mdx
```diff
@@ -40,9 +40,9 @@ $ curl \
     http://127.0.0.1:8500/v1/agent/members
 ```
 
-Previously this was provided via a `?token=` query parameter. This functionality
-exists on many endpoints for backwards compatibility, but its use is **highly
+Previously this was provided via a `?token=` query parameter but its use is **highly
 discouraged**, since it can show up in access logs as part of the URL.
+This functionality is deprecated and will be removed in future versions.
```
jkirschner-hashicorp, line 45: Questions for author and reviewers:
- Should we issue a warning in the Consul agent logs if someone attempts to use this deprecated parameter, stating that it will be removed in a future version?
- Should we call out the deprecation in the version-specific upgrade docs for 1.15? https://developer.hashicorp.com/consul/docs/upgrading/upgrade-specific

I'm leaning towards "yes" to both questions.

jkirschner-hashicorp (commented):
Added a few questions on further places to call out the deprecation.

cthain (approved):
LGTM. Just had a minor suggestion on deprecation wording for your consideration.

website/content/api-docs/api-structure.mdx
cthain: @jkirschner-hashicorp I believe @roncodingenthusiast 's plan is to create a separate PR for the code changes addressing bullets 1 and 3.

Great catch on bullet 2, I totally agree!

kisunji (approved):

roncodingenthusiast:
I am closing this PR in favour of: https://github.com/hashicorp/consul/pull/16009 as it makes sense for the change and the logs to go together

Issue #11917, opened 2021-12-23
Deprecate the token API query parameter 


dnephin:
The API token should be included in a request using the `X-Consul-Token` header, but we still accept the token from a query parameter.

Query parameters can easily end up in logs, so generally it is considered a bad idea to accept authentication via a query parameter.

We should deprecate support for the `token` API query parameter:

1. in the first release introduce a config option to disable support, default it to allow the query parameter, and add a warning when it's set to allow. The warning should indicate the default value will change in a future version.
2. in a following release change the default value to not allow
3. some time in the future remove support entirely

PR #16009, opened 2023-01-18
Warn when the token query param is used for auth

roncodingenthusiast:
### Description
- Emit a warning log message every time the token query param is used 
- Update docs to reflect the deprecation of the token query param 

Makes progress on #11917. Remaining work: fully remove in a future release.

#### Manual testing steps 
* Ran consul locally (make dev...)
* Tried a request with header
```
curl \
    --header "X-Consul-Token: <my token>" \
    "http://127.0.0.1:8500/v1/agent/members"

## No logs
```
* Tried a request with token without the header
```
curl \
    "http://127.0.0.1:8500/v1/agent/members?token=<my token>"

## log message
2023-01-18T15:37:36.989-0500 [WARN]  agent.http: This request is using a token query parameter; This is deprecated and will be removed in future Consul versions.: logUrl=/v1/agent/members?token=<hidden>
```
* Tried a request with token and the header present
```
curl \
    --header "X-Consul-Token: <my token>" \
    "http://127.0.0.1:8500/v1/agent/members?token=<my token>"

## log message
2023-01-18T15:48:18.764-0500 [WARN]  agent.http: This request is using a token query parameter; This is deprecated and will be removed in future Consul versions.: logUrl=/v1/agent/members?token=<hidden>
```

### Still todo
* [x] update tests that still use token query params to use header instead
* [x] Generate change-log

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/.changelog/16009.txt
+++ b/.changelog/16009.txt
@@ -1,3 +1,3 @@
-```release-note:improvement
-acl: Warn when query token parameter is used for authentication.
+```release-note:deprecation
+acl: Deprecate the `token` query parameter and warn when it is used for authentication.
 ```
--- a/.changelog/16009.txt
+++ b/.changelog/16009.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+Authentication: Warn when query token parameter is used for authentication.
+```
--- a/.changelog/16009.txt
+++ b/.changelog/16009.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-Authentication: Warn when query token parameter is used for authentication.
+acl: Warn when query token parameter is used for authentication.
 ```
--- a/agent/agent_endpoint_test.go
+++ b/agent/agent_endpoint_test.go
@@ -52,7 +52,8 @@ func createACLTokenWithAgentReadPolicy(t *testing.T, srv *HTTPHandlers) string {
 		Rules: `agent_prefix "" { policy = "read" }`,
 	}
 
-	req, _ := http.NewRequest("PUT", "/v1/acl/policy?token=root", jsonReader(policyReq))
+	req, _ := http.NewRequest("PUT", "/v1/acl/policy", jsonReader(policyReq))
+	req.Header.Add("X-Consul-Token", "root")
 	resp := httptest.NewRecorder()
 	srv.h.ServeHTTP(resp, req)
 	require.Equal(t, http.StatusOK, resp.Code)
@@ -62,7 +63,8 @@ func createACLTokenWithAgentReadPolicy(t *testing.T, srv *HTTPHandlers) string {
 		Policies:    []structs.ACLTokenPolicyLink{{Name: "agent-read"}},
 	}
 
-	req, _ = http.NewRequest("PUT", "/v1/acl/token?token=root", jsonReader(tokenReq))
+	req, _ = http.NewRequest("PUT", "/v1/acl/token", jsonReader(tokenReq))
+	req.Header.Add("X-Consul-Token", "root")
 	resp = httptest.NewRecorder()
 	srv.h.ServeHTTP(resp, req)
 	require.Equal(t, http.StatusOK, resp.Code)
@@ -398,7 +400,8 @@ func TestAgent_Services_ACLFilter(t *testing.T) {
 			}
 		`)
 
-		req := httptest.NewRequest("GET", fmt.Sprintf("/v1/agent/services?token=%s", token), nil)
+		req := httptest.NewRequest("GET", "/v1/agent/services", nil)
+		req.Header.Add("X-Consul-Token", token)
 		resp := httptest.NewRecorder()
 		a.srv.h.ServeHTTP(resp, req)
 
@@ -412,7 +415,8 @@ func TestAgent_Services_ACLFilter(t *testing.T) {
 	})
 
 	t.Run("root token", func(t *testing.T) {
-		req, _ := http.NewRequest("GET", "/v1/agent/services?token=root", nil)
+		req, _ := http.NewRequest("GET", "/v1/agent/services", nil)
+		req.Header.Add("X-Consul-Token", "root")
 		resp := httptest.NewRecorder()
 		a.srv.h.ServeHTTP(resp, req)
 		dec := json.NewDecoder(resp.Body)
@@ -571,7 +575,8 @@ func TestAgent_Service(t *testing.T) {
 				time.Sleep(100 * time.Millisecond)
 				// Re-register with new proxy config, make sure we copy the struct so we
 				// don't alter it and affect later test cases.
-				req, _ := http.NewRequest("PUT", "/v1/agent/service/register?token=root", jsonReader(updatedProxy))
+				req, _ := http.NewRequest("PUT", "/v1/agent/service/register", jsonReader(updatedProxy))
+				req.Header.Add("X-Consul-Token", "root")
 				resp := httptest.NewRecorder()
 				a.srv.h.ServeHTTP(resp, req)
 				require.Equal(t, 200, resp.Code, "body: %s", resp.Body.String())
@@ -604,7 +609,8 @@ func TestAgent_Service(t *testing.T) {
 			updateFunc: func() {
 				time.Sleep(100 * time.Millisecond)
 				// Re-register with _same_ proxy config
-				req, _ := http.NewRequest("PUT", "/v1/agent/service/register?token=root", jsonReader(sidecarProxy))
+				req, _ := http.NewRequest("PUT", "/v1/agent/service/register", jsonReader(sidecarProxy))
+				req.Header.Add("X-Consul-Token", "root")
 				resp := httptest.NewRecorder()
 				a.srv.h.ServeHTTP(resp, req)
 				require.Equal(t, 200, resp.Code, "body: %s", resp.Body.String())
@@ -694,7 +700,8 @@ func TestAgent_Service(t *testing.T) {
 
 			// Register the basic service to ensure it's in a known state to start.
 			{
-				req, _ := http.NewRequest("PUT", "/v1/agent/service/register?token=root", jsonReader(sidecarProxy))
+				req, _ := http.NewRequest("PUT", "/v1/agent/service/register", jsonReader(sidecarProxy))
+				req.Header.Add("X-Consul-Token", "root")
 				resp := httptest.NewRecorder()
 				a.srv.h.ServeHTTP(resp, req)
 				require.Equal(t, 200, resp.Code, "body: %s", resp.Body.String())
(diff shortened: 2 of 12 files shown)


website/content/api-docs/api-structure.mdx
```diff
@@ -40,9 +40,9 @@ $ curl \
     http://127.0.0.1:8500/v1/agent/members
 ```
 
-Previously this was provided via a `?token=` query parameter. This functionality
-exists on many endpoints for backwards compatibility, but its use is **highly
+Previously this was provided via a `?token=` query parameter but its use is **highly
 discouraged**, since it can show up in access logs as part of the URL.
+This functionality is deprecated and will be removed in future versions.
```
jkirschner-hashicorp, line 45: Following up on @cthain 's comment from a previous PR...

Perhaps we should give explicit indication of when this will be removed. I don't expect all users to be familiar with our deprecation / removal policy.

Should we say something like:
```suggestion
Though the token could also be provided via a `?token=` query parameter,
that method is **highly discouraged** because the token can show up in access logs as part of the URL.
The `?token=` query parameter is deprecated and will be removed in Consul 1.17.
```

agent/http.go
```diff
@@ -367,6 +367,9 @@ func (s *HTTPHandlers) wrap(handler endpoint, methods []string) http.HandlerFunc
 					continue
 				}
 				logURL = strings.Replace(logURL, token, "<hidden>", -1)
+				httpLogger.Warn("This request is using a token query parameter "+
+					"which is deprecated and will be removed in future Consul versions.",
```
jkirschner-hashicorp, line 371: If we change the comment below, we might also say "will be removed in Consul 1.17" here.

website/content/docs/upgrading/upgrade-specific.mdx
```diff
@@ -20,6 +20,26 @@ upgrade flow.
...
+    http://127.0.0.1:8500/v1/agent/members?token=<consul token>
+```
+
+In Consul v1.15 and higher:
+
+```shell-session
+$ curl \
+    --header "X-Consul-Token: <consul token>" \
+    http://127.0.0.1:8500/v1/agent/members
+```
```
jkirschner-hashicorp, line 41: This may imply that passing via the `token` query parameter was expected before Consul v1.15, which isn't accurate.

The `X-Consul-Token` header has been supported for a while and has been the recommended way. Now we're explicitly deprecating the token query param.

We should also mention the log message that indicates if the token query param is in use.

Perhaps change to something like:

````suggestion
Providing a Consul ACL token in API requests using the `token` query parameter is deprecated and will be removed in Consul 1.17.
Instead, the token should be provided using the `X-Consul-Token` header or with the Bearer scheme in the authorization header as described in the [API authentication documentation](/consul/api-docs/api-structure#authentication).

To check whether the `token` query parameter is in use, search your Consul agent logs for the message:
```shell-session
<INSERT WHAT THE MESSAGE IS>
```

Deprecated authentication using the `token` query parameter:

```shell-session
$ curl \
    http://127.0.0.1:8500/v1/agent/members?token=<consul token>
```

Recommended usage:

```shell-session
$ curl \
    --header "X-Consul-Token: <consul token>" \
    http://127.0.0.1:8500/v1/agent/members
```
````

File: .changelog/16009.txt [hunk 4]
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+Authentication: Warn when query token parameter is used for authentication.
```
kisunji, line 2: It's a little hard to discover but we have a set of "subsystems" we use in our changelogs. Adding something new won't break anything but it likely won't be grouped with similar PRs

```suggestion
acl: Warn when query token parameter is used for authentication.
```

agent/http.go
```diff
@@ -367,6 +367,9 @@ func (s *HTTPHandlers) wrap(handler endpoint, methods []string) http.HandlerFunc
 					continue
 				}
 				logURL = strings.Replace(logURL, token, "<hidden>", -1)
+				httpLogger.Warn("This request is using a token query parameter "+
```
kisunji, line 370: I don't foresee there ever being more than one `"token"` value but we should have this warn outside of the for loop since we wouldn't need to log more than once.

kisunji (approved):
:shipit:

website/content/api-docs/api-structure.mdx
```diff
@@ -40,9 +40,9 @@ $ curl \
     http://127.0.0.1:8500/v1/agent/members
 ```
 
-Previously this was provided via a `?token=` query parameter. This functionality
-exists on many endpoints for backwards compatibility, but its use is **highly
-discouraged**, since it can show up in access logs as part of the URL.
+Though the token could also be provided via a `?token=` query parameter,
+that method is **highly discouraged** because the token can show up in access logs as part of the URL.
+The `?token=` query parameter is deprecated and will be removed in Consul 1.17.
```
im2nguyen, line 45:
```suggestion
~> **Security Note:** Though you could pass the token through a `?token=` query parameter, this method is highly discouraged because the token can show up in access logs as part of the URL. The `?token=` query parameter is deprecated and will be removed in Consul 1.17.
```

website/content/docs/upgrading/upgrade-specific.mdx
```diff
@@ -20,6 +20,32 @@ upgrade flow.
...
+#### Deprecating authentication via token query parameter
+
+Providing a Consul ACL token in API requests using the `token` query parameter is deprecated and will be removed in Consul 1.17.
+Instead, the token should be provided using the `X-Consul-Token` header or with the Bearer scheme in the authorization header as described in the [API authentication documentation](/consul/api-docs/api-structure#authentication).
+	
+To check whether the `token` query parameter is in use, search your Consul agent logs for the message:
+
+```shell-session
+$ This request is using a token query parameter which is deprecated and will be removed in Consul 1.17
+```
```
im2nguyen, line 32:
````suggestion
Instead, you should provide the token through the `X-Consul-Token` header or with the Bearer scheme in the authorization header as described in the [API authentication documentation](/consul/api-docs/api-structure#authentication).
	
Check whether you are using the `token` query parameter by searching your Consul agent logs for the message:

```shell-session hideClipboard
$ This request is using a token query parameter which is deprecated and will be removed in Consul 1.17
````
```

im2nguyen (approved):
a couple nits, but pre-approving so you're not blocked!

agent/http.go
```diff
@@ -368,6 +368,9 @@ func (s *HTTPHandlers) wrap(handler endpoint, methods []string) http.HandlerFunc
 				}
 				logURL = strings.Replace(logURL, token, "<hidden>", -1)
 			}
+			httpLogger.Warn("This request is using a token query parameter "+
```
jkirschner-hashicorp, line 371: There's only one, so we should say "the" token query param. "a" suggests there might be multiple query parameters that could include a token, rather than that the query param is named literally "token". (At least by my read)

```suggestion
			httpLogger.Warn("This request is using the token query parameter "+
```

website/content/docs/upgrading/upgrade-specific.mdx
```diff
@@ -20,6 +20,32 @@ upgrade flow.
...
 
+#### Deprecating authentication via token query parameter
+
+Providing a Consul ACL token in API requests using the `token` query parameter is deprecated and will be removed in Consul 1.17.
+Instead, the token should be provided using the `X-Consul-Token` header or with the Bearer scheme in the authorization header as described in the [API authentication documentation](/consul/api-docs/api-structure#authentication).
+	
+To check whether the `token` query parameter is in use, search your Consul agent logs for the message:
+
+```shell-session
+$ This request is using a token query parameter which is deprecated and will be removed in Consul 1.17
```
jkirschner-hashicorp, line 31:
```suggestion
$ This request is using the token query parameter which is deprecated and will be removed in Consul 1.17
```

website/content/docs/upgrading/upgrade-specific.mdx
jkirschner-hashicorp: "the token query param" rather than "a token query param"

.changelog/16009.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+acl: Warn when query token parameter is used for authentication.
```
cthain, line 2: What do you think about including the deprecation notice in the entry? I feel like it makes this more clear. Something like:

```suggestion
acl: Deprecate the `token` query parameter and warn when it is used for authentication.
```

agent/agent_endpoint_test.go
```diff
@@ -2009,9 +2023,11 @@ func TestAgent_Members_ACLFilter(t *testing.T) {
 	testrpc.WaitForLeader(t, a.RPC, "dc1")
 	testrpc.WaitForLeader(t, b.RPC, "dc1")
 
-	joinPath := fmt.Sprintf("/v1/agent/join/127.0.0.1:%d?token=root", b.Config.SerfPortLAN)
+	joinPath := fmt.Sprintf("/v1/agent/join/127.0.0.1:%d", b.Config.SerfPortLAN)
+	req, _ := http.NewRequest("PUT", joinPath, nil)
```
cthain, line 2027: nit: The original request was using `httptest.NewRequest` instead of `http.NewRequest` as is done now. Just wondering if there is a reason to use the latter in this case instead?

website/content/docs/upgrading/upgrade-specific.mdx
cthain: nit: The suggestion in http.go says "This request **used** the token..." I think "used" is the word that we want here, not "uses".. Just pointing this out for consistency.

cthain (commented):
Looking really good. I had a few minor comments/suggestions.

agent/agent_endpoint_test.go
roncodingenthusiast: no particular reason, I actually will reverse to `httptest` so I don't accidentally modify the behaviour. Good catch!

cthain (approved):
Great work on this PR Ronald :+1:

jkirschner-hashicorp:
Awesome work :) Looks like engineering and education have both approved. Merge at will!

(Well, I guess the branch needs to be re-updated from main first / the CI pipeline has to finish first)

agent/http.go
```diff
@@ -368,6 +368,9 @@ func (s *HTTPHandlers) wrap(handler endpoint, methods []string) http.HandlerFunc
 				}
 				logURL = strings.Replace(logURL, token, "<hidden>", -1)
 			}
+			httpLogger.Warn("This request used the token query parameter "+
+				"which is deprecated and will be removed in Consul 1.17",
+				"logUrl", logURL)
```
jkirschner-hashicorp, line 373: @kisunji , @cthain , @roncodingenthusiast : Should we strip the token from the `logUrl`?

jkirschner-hashicorp (changes requested):
Placing a block on this until we discuss something I just left a comment on. (Still writing the comment.)

agent/http.go
jkirschner-hashicorp: Does the token secret get stripped from the `logUrl` in the log output? Or does this log message actually include the token secret?

jkirschner-hashicorp: Edit: I've been informed that the `logUrl` already obfuscates the token before the log is emitted

jkirschner-hashicorp (approved):
My previous review comment has been addressed 👍

kisunji:
@roncodingenthusiast You'll have to OAuth into CircleCI with Github for the CI to run in this PR

PR #16013, opened 2023-01-19
Decrease polling frequency on peer proxy config data.

hashi-derek:
This is a temporary fix to reduce the performance cost of fetching peering data in proxycfg. A better long-term fix should be added that uses streaming or blocking-queries to ensure updates are triggered only when changes are made, rather than polling.


DanStough:
After talking with @hashi-derek, we decided this could have unintended consequences when used with `PeerThroughMeshGateways`, which has only a 4-second window to get an update. We'll be addressing the performance issue with blocking or streaming queries in a follow-up PR.

PR #16020, opened 2023-01-20
docs: Consul K8s upgrade leave on terminate

david-yu:
### Description

Add more instructions to guide users to use "leave_on_terminate" when upgrading via Helm. See https://github.com/hashicorp/consul-k8s/issues/1516#issuecomment-1253923057 

### Links

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


david-yu:
Closing as although this introduces less errors in log files this could result in a split brain scenario, i.e. when server 1/3 leaves, now quorum is 2/2, if there's a network partition at that point, then you can get a split brain scenario

PR #16022, opened 2023-01-20
feat: panic handler in rpc rate limit interceptor

JadhavPoonam:
### Description
Add a panic handler that logs the error which is passed into the RPC rate limit interceptor (similar gRPC codepath)

### PR Checklist

* [x] updated test coverage
* [x] not a security

### DIFF
--- a/agent/rpc/middleware/interceptors_test.go
+++ b/agent/rpc/middleware/interceptors_test.go
@@ -3,6 +3,7 @@ package middleware
 import (
 	"errors"
 	"net"
+	"net/netip"
 	"strings"
 	"sync"
 	"testing"
@@ -275,16 +276,16 @@ func TestGetNetRPCRateLimitingInterceptor(t *testing.T) {
 	limiter := rate.NewMockRequestLimitsHandler(t)
 
 	logger := hclog.NewNullLogger()
-	var rateLimitInterceptor = GetNetRPCRateLimitingInterceptor(limiter, NewPanicHandler(logger))
+	rateLimitInterceptor := GetNetRPCRateLimitingInterceptor(limiter, NewPanicHandler(logger))
 
-	listener, _ := net.Listen("tcp", "127.0.0.1:0")
+	addr := net.TCPAddrFromAddrPort(netip.MustParseAddrPort("1.2.3.4:5678"))
 
 	t.Run("allow operation", func(t *testing.T) {
 		limiter.On("Allow", mock.Anything).
 			Return(nil).
 			Once()
 
-		err := rateLimitInterceptor("Status.Leader", listener.Addr())
+		err := rateLimitInterceptor("Status.Leader", addr)
 		require.NoError(t, err)
 	})
 
@@ -293,7 +294,7 @@ func TestGetNetRPCRateLimitingInterceptor(t *testing.T) {
 			Return(errors.New("uh oh")).
 			Once()
 
-		err := rateLimitInterceptor("Status.Leader", listener.Addr())
+		err := rateLimitInterceptor("Status.Leader", addr)
 		require.Error(t, err)
 		require.Equal(t, "uh oh", err.Error())
 	})
@@ -303,7 +304,7 @@ func TestGetNetRPCRateLimitingInterceptor(t *testing.T) {
 			Panic("uh oh").
 			Once()
 
-		err := rateLimitInterceptor("Status.Leader", listener.Addr())
+		err := rateLimitInterceptor("Status.Leader", addr)
 
 		require.Error(t, err)
 		require.Equal(t, "rpc: panic serving request", err.Error())
--- a/agent/rpc/middleware/interceptors_test.go
+++ b/agent/rpc/middleware/interceptors_test.go
@@ -1,6 +1,7 @@
 package middleware
 
 import (
+	"errors"
 	"net"
 	"strings"
 	"sync"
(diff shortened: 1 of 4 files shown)


agent/rpc/middleware/interceptors_test.go
```diff
@@ -266,3 +269,23 @@ func TestRequestRecorder(t *testing.T) {
 		})
 	}
 }
+
+func TestGetNetRPCRateLimitingInterceptor(t *testing.T) {
```
dhiaayachi, line 273: I think as we are adding tests in here we could add some tests around the interceptor, I'm thinking:
- That `Allow` is called
- That the `Allow` error is returned through the interceptor.
- What would happen if `reqServiceMethod` is not defined in `rpcRateLimitSpecs` (I think that use case is a bug actually) 😃

dhiaayachi: Could be deferred to a separate PR if you think it's better though

JadhavPoonam: When the service method is not listed in the rate limit spec then it defaults to "OperationType.READ"

agent/rpc/middleware/recovery.go
```diff
@@ -0,0 +1,24 @@
+package middleware
```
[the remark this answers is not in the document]
JadhavPoonam: If we want to keep the grpc vs rpc visibility in logs then we should keep them separate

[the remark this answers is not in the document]
dhiaayachi: We are printing the call stack as part of the error log  we should be able to differentiate gRPC from RPC at that level.

agent/rpc/middleware/interceptors_test.go
```diff
@@ -266,3 +269,23 @@ func TestRequestRecorder(t *testing.T) {
...
+
+	t.Run("Allow panics", func(t *testing.T) {
+		limiter.On("Allow", mock.Anything).
+			Panic("uh oh").
+			Once()
+
+		err := rateLimitInterceptor("Status.Leader", listener.Addr())
+
+		require.Error(t, err)
+		require.Equal(t, "rpc: panic serving request", err.Error())
```
boxofrad, line 309: Doesn't look like we actually use the listener, other than to get an address. Let's hard-code it!

```suggestion
	addr := net.TCPAddrFromAddrPort(netip.MustParseAddrPort("1.2.3.4:5678"))

	t.Run("Allow panics", func(t *testing.T) {
		limiter.On("Allow", mock.Anything).
			Panic("uh oh").
			Once()

		err := rateLimitInterceptor("Status.Leader", addr)

		require.Error(t, err)
		require.Equal(t, "rpc: panic serving request", err.Error())
```

boxofrad (approved):
This looks great! Nice job @JadhavPoonam 🙌🏻 

I've left a couple of small comments for your consideration. The listener one is more important because we don't currently `Close` it, so we're kinda leaking a listener there.

agent/rpc/middleware/interceptors_test.go
JadhavPoonam: Addressed here https://github.com/hashicorp/consul/pull/16022/commits/218ad46ce73cb2b19649c7a264ed66102dba83af

agent/rpc/middleware/interceptors_test.go
```diff
@@ -266,3 +269,23 @@ func TestRequestRecorder(t *testing.T) {
 		})
 	}
 }
+
+func TestGetNetRPCRateLimitingInterceptor(t *testing.T) {
+	limiter := rate.NewMockRequestLimitsHandler(t)
+
+	logger := hclog.NewNullLogger()
+	var rateLimitInterceptor = GetNetRPCRateLimitingInterceptor(limiter, NewPanicHandler(logger))
```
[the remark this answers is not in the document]
JadhavPoonam: Addressed here https://github.com/hashicorp/consul/pull/16022/commits/218ad46ce73cb2b19649c7a264ed66102dba83af

agent/rpc/middleware/recovery.go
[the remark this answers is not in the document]
JadhavPoonam: Updated here https://github.com/hashicorp/consul/pull/16022/commits/8c3c7dca87a907da21a54fd88e8e855a5ef2af1e
@boxofrad can I also get your 👀 on this since it makes some changes to gRPC middleware 🙏

agent/grpc-middleware/recovery.go
```diff
@@ -2,35 +2,22 @@ package middleware
...
 	return func(p interface{}) (err error) {
-		// Log the panic and the stack trace of the Goroutine that caused the panic.
-		stacktrace := hclog.Stacktrace()
-		logger.Error("panic serving grpc request",
-			"panic", p,
-			"stack", stacktrace,
-		)
-
-		return status.Errorf(codes.Internal, "grpc: panic serving request")
+		panicErr := rate.NewPanicHandler(logger)
```
boxofrad, line 20: It feels a little odd to have the panic handler live in the `rate` package 🤔 

Personally, as it's just a couple of lines of code, I'd follow the [Go proverb](https://go-proverbs.github.io/) and just duplicate it:

> A little copying is better than a little dependency.

That said, if we do want to share it, maybe we could put it somewhere more generic like `agent/panic`?

boxofrad (approved):

agent/grpc-middleware/recovery.go
JadhavPoonam: Hmm... `agent/panic` might be too broad of a scope for the limited use case it currently serves 🤔  WDYT?
I can certainly move it out if you feel strongly about it 🙂 

From my perspective, the rate package made sense because this is where `Allow` is implemented and the panic handler is limited to the rate-limiting atm.

JadhavPoonam: Ahh yes...! Thank you for keeping us honest @boxofrad 🙏 
I synced with @dhiaayachi and we decided to keep these separate.

dhiaayachi (approved):
Great work @JadhavPoonam! 👏 👏 👏 
LGTM

PR #16044, opened 2023-01-23
Update token language to distinguish Accessor and Secret ID usage

skpratt:
### Description
Clarify in CLI and code when we are using a token's AccessorID vs SecretID.

Tested locally - bootstrapped ACLs for 3 server cluster with 2 clients, tested normal UI and token-related CLI and API commands.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


agent/acl_test.go
```diff
@@ -91,7 +91,7 @@ func NewTestACLAgent(t *testing.T, name string, hcl string, resolveAuthz authzRe
 
 func (a *TestACLAgent) ResolveToken(secretID string) (acl.Authorizer, error) {
```
kisunji, line 92: Does this need to be renamed to `ResolveTokenSecret` as well?

agent/structs/acl.go
```diff
@@ -1337,7 +1329,7 @@ type ACLTokenBatchSetRequest struct {
 // This is particularly useful during token replication when
 // multiple tokens need to be removed from the local DCs state.
 type ACLTokenBatchDeleteRequest struct {
-	TokenIDs []string // Tokens to delete
+	TokenAccessorIDs []string // Tokens to delete
```
kisunji, line 1332: I suspect that this will be a breaking change because this type is used for Raft Apply operations and the field names are serialized, so this cause older versions of Consul servers to fail to decode the newer version of this message.
cc @rboyer

api/acl.go
```diff
@@ -750,7 +750,7 @@ func (a *ACL) TokenUpdate(token *ACLToken, q *WriteOptions) (*ACLToken, *WriteMe
 // of an existing token.
 func (a *ACL) TokenClone(tokenID string, description string, q *WriteOptions) (*ACLToken, *WriteMeta, error) {
```
kisunji, line 751: Should we rename `tokenID` to `accessorID` here?

command/acl/token/clone/token_clone.go
```diff
@@ -47,16 +49,26 @@ func (c *cmd) init() {
...
 		return 1
 	}
 
-	if c.tokenID == "" {
-		c.UI.Error(fmt.Sprintf("Cannot update a token without specifying the -id parameter"))
-		return 1
+	tokenAccessor := c.tokenAccessorID
+	if tokenAccessor == "" {
+		if c.tokenID == "" {
+			c.UI.Error(fmt.Sprintf("Cannot update a token without specifying the -accessor-id parameter"))
```
kisunji, line 66: Nit: I think the `fmt.Sprintf` is unnecessary here and below

kisunji (changes requested):
LGTM but had one blocking point about msgpack backwards compatibility

agent/structs/acl.go
rboyer: A quick search seems to imply the field names in msgpack are required to not change: https://msgpack.org/index.html

agent/acl_test.go
skpratt: This ties into the server struct in xds code, which would make more sweeping changes that I'd like to consider outside the scope of this PR
https://github.com/hashicorp/consul/blob/674c5570b6c2d8f0e00972bc130efe4626bed2d7/agent/xds/server.go#L108

skpratt: done!

agent/structs/acl.go
skpratt: Thanks for flagging, reverted

api/acl.go
skpratt: done

command/acl/token/clone/token_clone.go
skpratt: thank you for catching, done

command/acl/token/clone/token_clone.go
```diff
@@ -47,16 +49,26 @@ func (c *cmd) init() {
...
-		c.UI.Error(fmt.Sprintf("Cannot update a token without specifying the -id parameter"))
-		return 1
+	tokenAccessor := c.tokenAccessorID
+	if tokenAccessor == "" {
+		if c.tokenID == "" {
+			c.UI.Error("Cannot update a token without specifying the -accessor-id parameter")
+			return 1
+		} else {
+			tokenAccessor = c.tokenID
+			c.UI.Warn("Use the -accessor-id parameter to specify token by Accessor ID.")
```
pglass, line 70: minor: I feel like the warning should include what they've done "wrong" (i.e. they used the deprecated flag).

```suggestion
			c.UI.Warn("The -id parameter is deprecated. Use the -accessor-id parameter instead.")
```

(commenting once for all locations with the same warning)

pglass (approved):
Looks good!

command/acl/token/clone/token_clone.go
jkirschner-hashicorp: Agree with Paul. The current wording also suggests that the operation didn't actually go through (because we're just saying "Do this differently"). Stating that a param is deprecated implies that the operation still went through.

PR #16045, opened 2023-01-23
test: container tests wait for available networks

rboyer:
### Description

If multiple container tests are run in parallel it is highly likely that one will randomly fail due to an inability to get a docker network for the test case. This change detects that scenario and sleep retries until it succeeds. Hopefully this means it will be safe to mark some tests as safe to run in parallel without flaking them out.


DanStough (approved):
LGTM

PR #16049, opened 2023-01-24
chore: document and unit test sdk/testutil/retry

nfi-hashicorp:
I was working with this test library in https://github.com/hashicorp/consul-enterprise/pull/4115 and the behavior is quite surprising, undocumented, and lacking unit tests showing the intended behavior. So I did some guessing at intent and documented it.

Note: Do not take this as my endorsement of this library being good. It is weird.

Bonus: there's a small bug that I've fixed. In [`ac2ac79` (#16049)](https://github.com/hashicorp/consul/pull/16049/commits/ac2ac7912143b77fb30c4cf18cc1a85779d9b30d) I add a test to show the bug (see [test run](https://app.circleci.com/pipelines/github/hashicorp/consul/49803/workflows/3a6a03d2-388b-471d-b84a-65a66b01422b/jobs/1141808)). In [`7e9af2c` (#16049)](https://github.com/hashicorp/consul/pull/16049/commits/7e9af2cf4c7afab648809528f40f83b00b62996d) I fix it. Doing `panic(struct{}{})` regularly would be a very weird thing to do, so I highly doubt this would affect anything. Even so, it should only be test code.

### PR Checklist

* [x] updated test coverage
* [x] external facing docs updated -- N/A
* [x] not a security concern

### DIFF
--- a/sdk/testutil/retry/retry_test.go
+++ b/sdk/testutil/retry/retry_test.go
@@ -42,7 +42,7 @@ func TestRetryer(t *testing.T) {
 	}
 }
 
-func TestBasic(t *testing.T) {
+func TestBasics(t *testing.T) {
 	t.Run("Error allows retry", func(t *testing.T) {
 		i := 0
 		Run(t, func(r *R) {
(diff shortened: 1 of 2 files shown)


sdk/testutil/retry/retry_test.go
```diff
@@ -65,6 +116,7 @@ func TestRunWith(t *testing.T) {
 			r.Fatalf("not yet")
 		})
 
+		// TODO: these should all be assert
```
NiniOak, line 119: Why assert? Do you want the test to keep running if there's a failure in the test suite?

NiniOak (approved):

sdk/testutil/retry/retry_test.go
nfi-hashicorp: The way I see it, you should use `assert` when those assertions are independent, but `require` if they are dependent. Or put another way, use `require` when the following assertions couldn't possibly pass, or are impossible to test.

Overusing `require` means that if all the following assertions fail for independent reasons, we'd need to do multiple fix-and-test passes to identify that that is the case.

NiniOak: Gotcha. Thanks for clarifying

PR #16052, opened 2023-01-24
Add trigger for doing reconciliation based on watch sets

andrewstucki:
### Description
This adds support for triggering reconciliation based on a function trigger. This is a workaround that we'll ideally remove later on, once all of our primitives are moved to "resources." The use-case is that we need to re-trigger reconciliation of routes when their referenced service discovery chains change.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/consul/controller/controller.go
```diff
@@ -130,6 +163,14 @@ func (c *controller) WithWorkers(i int) Controller {
 	return c
 }
 
+// WithLogger set the the internal logger for the controller.
```
t-eckert, line 166:
```suggestion
// WithLogger sets the the internal logger for the controller.
```

t-eckert (approved):
Excellent! This looks great and it's so clear!

sarahalsmiller (approved):

PR #16053, opened 2023-01-24
Add a server-only method for updating ConfigEntry Statuses

andrewstucki:
### Description
This adds functionality to our internal FSM operations in order to make ConfigEntry statuses that are set by our new controller infrastructure immutable to users but able to be set by directly initializing an FSM operation.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/consul/state/config_entry.go
```diff
@@ -281,7 +292,42 @@ func (s *Store) EnsureConfigEntryCAS(idx, cidx uint64, conf structs.ConfigEntry)
...
+	tx := s.db.WriteTxn(idx)
+	defer tx.Abort()
+
+	// Check for existing configuration.
+	existing, err := tx.First(tableConfigEntries, indexID, newConfigEntryQuery(conf))
+	if err != nil {
+		return false, fmt.Errorf("failed configuration lookup: %s", err)
+	}
+
+	// Check if the we should do the set. A ModifyIndex of 0 means that
```
t-eckert, line 314:
```suggestion
	// Check if we should do the set. A ModifyIndex of 0 means that
```

t-eckert (approved):
Excellent!

agent/consul/fsm/commands_oss.go
```diff
@@ -562,6 +562,14 @@ func (c *FSM) applyConfigEntryOperation(buf []byte, index uint64) interface{} {
 			return err
 		}
 		return true
+	case structs.ConfigEntryUpsertStatusCAS:
+		defer metrics.MeasureSinceWithLabels([]string{"fsm", "config_entry", req.Entry.GetKind()}, time.Now(),
+			[]metrics.Label{{Name: "op", Value: "upsert_status"}})
```
nathancoleman, line 567: Wondering if this should be rolled in with the `"upsert"`s above for the purposes of metrics. Thoughts?

agent/consul/fsm/commands_oss_test.go
```diff
@@ -1347,6 +1347,111 @@ func TestFSM_ConfigEntry(t *testing.T) {
...
 }
 
+func TestFSM_ConfigEntry_StatusCAS(t *testing.T) {
+	t.Parallel()
+
+	logger := testutil.Logger(t)
+	fsm, err := New(nil, logger)
+	require.NoError(t, err)
+
+	// Create a simple config entry
```
nathancoleman, line 1357: I appreciate the comments throughout this test - thanks!

agent/consul/fsm/commands_oss_test.go
```diff
@@ -1347,6 +1347,111 @@ func TestFSM_ConfigEntry(t *testing.T) {
...
+		_, config, err := fsm.state.ConfigEntry(nil, structs.APIGateway, "global", nil)
+		require.NoError(t, err)
+		entry.RaftIndex.CreateIndex = 2
+		entry.RaftIndex.ModifyIndex = 2
+		conditions := config.(*structs.APIGatewayConfigEntry).Status.Conditions
+		require.Len(t, conditions, 1)
+		require.Equal(t, "Foo", conditions[0].Status)
+	}
+
+	// do a regular update and make sure the status gets reset
```
nathancoleman, line 1423: I think this is the intention of this test, makes it a little clearer to me

```suggestion
	// attempt to change the status w/ a regular update and make sure it's ignored
```

agent/consul/fsm/commands_oss_test.go
```diff
@@ -1347,6 +1347,111 @@ func TestFSM_ConfigEntry(t *testing.T) {
...
+			t.Fatalf("bad: %v", resp)
+		}
+	}
+
+	// Verify it's in the state store.
+	{
+		_, config, err := fsm.state.ConfigEntry(nil, structs.APIGateway, "global", nil)
+		require.NoError(t, err)
+		entry.RaftIndex.CreateIndex = 2
+		entry.RaftIndex.ModifyIndex = 2
```
nathancoleman, line 1448: Do these need to be assigned here?

agent/consul/state/config_entry.go
```diff
@@ -281,7 +292,42 @@ func (s *Store) EnsureConfigEntryCAS(idx, cidx uint64, conf structs.ConfigEntry)
...
+	if err := ensureConfigEntryTxn(tx, idx, false, conf); err != nil {
+		return false, err
+	}
+
+	err = tx.Commit()
+	return err == nil, err
+}
+
+// EnsureConfigEntryStatusCAS is called to do a check-and-set upsert of a given config entry status.
+func (s *Store) EnsureConfigEntryStatusCAS(idx, cidx uint64, conf structs.ConfigEntry) (bool, error) {
```
nathancoleman, line 304: The name and godoc here imply to me that we're _only_ setting the status; however, I believe we're setting the whole config entry _and the status_ where we would usually just ignore any status on a normal `EnsureConfigEntryCAS`. Is that accurate?

If so, I might change the name a bit and update the godoc to reflect
```suggestion
// EnsureConfigEntryWithStatusCAS is called to do a check-and-set upsert of a given config entry and its status.
func (s *Store) EnsureConfigEntryWithStatusCAS(idx, cidx uint64, conf structs.ConfigEntry) (bool, error) {
```

nathancoleman: It might also be helpful to note in the godocs for other `EnsureX` that the status is ignored

nathancoleman (changes requested):
Nice work! Thanks for knocking this out so quickly

agent/consul/state/config_entry.go
```diff
@@ -237,7 +237,18 @@ func ensureConfigEntryTxn(tx WriteTxn, idx uint64, conf structs.ConfigEntry) err
...
+			if controlledConf, ok := conf.(structs.ControlledConfigEntry); ok {
+				controlledConf.SetStatus(existing.(structs.ControlledConfigEntry).GetStatus())
+			}
+		}
 	} else {
+		if !statusUpdate {
+			if controlledConf, ok := conf.(structs.ControlledConfigEntry); ok {
+				controlledConf.SetStatus(controlledConf.DefaultStatus())
+			}
+		}
```
nathancoleman, line 251: Thinking out loud here, it seems that in **most** cases, `conf` is _not_ a `ControlledConfigEntry` whenever `statusUpdate == false`; however, when `conf` _is_ a `ControlledConfigEntry`, we want to reset the status to whatever already exists in committed state or the default status if no committed state. Accurate?

agent/consul/fsm/commands_oss_test.go
andrewstucki: In this particular case, no, since this is the last expectation and doesn't involve a subsequent CAS operation

agent/consul/state/config_entry.go
andrewstucki: That makes sense RE updating the other godoc code. The name reflects my own indecision about whether or not this method should be used as a generic internal update method mechanism that can update both spec and status all at once or if it should solely be scoped to the status. I had followed the naming similar to the kubernetes client with `UpdateStatus`, but then went ahead and just implemented it as an update with the `Status` resetting bypassed.

Though Kubernetes does scope things solely to updating the status, I'm thinking that `WithStatus` semantics may be a bit more desirable, thoughts?

nathancoleman: That makes sense, and I agree 💯

agent/consul/state/config_entry.go
andrewstucki: Yep, so if it's a create operation for a `ControlledConfigEntry` and you're using one of the user-facing operations that shouldn't touch status, then set the status to `DefaultStatus`; if it's an update, then just ensure that the status is not changed by merging in the currently stored status.

agent/consul/fsm/commands_oss.go
andrewstucki: I'm thinking we probably want to split this out just because this is more of an internal call than the others, I'm ok with changing the metric from `upsert_status` to like `internal_upsert` or `controller_upsert` or something of the sort, but I personally think you'd get more utility out of keeping the stats separate for what is user-controlled v. internal.

agent/consul/fsm/commands_oss_test.go
```diff
@@ -1347,6 +1347,111 @@ func TestFSM_ConfigEntry(t *testing.T) {
...
+		if _, ok := resp.(error); ok {
+			t.Fatalf("bad: %v", resp)
+		}
+	}
+
+	// Verify it's in the state store.
+	{
+		_, config, err := fsm.state.ConfigEntry(nil, structs.APIGateway, "global", nil)
+		require.NoError(t, err)
+		entry.RaftIndex.CreateIndex = 1
```
jm96441n, line 1387: what does the `1` here mean?

andrewstucki: It's the operation number in the raft log that's used as part of the CAS operation and resource version tracking. It gets autoincremented as ops are applied via a call to `fsm.Apply`. It ends up getting incremented here to make it match the entry retrieved from the store so that the subsequent CAS operations can succeed.

nathancoleman (approved):

PR #16054, opened 2023-01-24
Run config entry controller routines on leader

nathancoleman:
### Description
For new Consul-native API Gateway config entries, a set of "controllers" - one per config entry kind - needs to be run on the leader which react to changes in the corresponding config entry kind.

The controllers don't actually _do_ anything yet in response to a config entry change, so this code change is starting up goroutines that just subscribe to topics. Passing 🤖 checks should be sufficient.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


t-eckert (approved):
Great work! Will we need a follow-up to add a controller for inline certs?

nathancoleman:
Discussed elsewhere that we don't expect to need a full-blown controller for inline certs since they can really only be valid or invalid - we aren't deriving state from them or anything like we do with xRoutes and Gateways

PR #16055, opened 2023-01-24
Adding a sentence to explain the interchangeable terms

luoxuan00733:
Failure tolerance and fault tolerance are often used interchangably.

### Description

### Links

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/website/content/docs/architecture/improving-consul-resilience.mdx
+++ b/website/content/docs/architecture/improving-consul-resilience.mdx
@@ -5,9 +5,9 @@ description: >-
   Fault tolerance is a system's ability to operate without interruption despite component failure. Learn how a set of Consul servers provide fault tolerance through use of a quorum, and how to further improve control plane resilience through use of infrastructure zones and Enterprise redundancy zones.
 ---
 
-# Fault Tolerance
+# Fault (Failure) Tolerance
 
-Fault tolerance is the ability of a system to continue operating without interruption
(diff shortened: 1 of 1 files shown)


website/content/docs/architecture/improving-consul-resilience.mdx
```diff
@@ -5,9 +5,9 @@ description: >-
   Fault tolerance is a system's ability to operate without interruption despite component failure. Learn how a set of Consul servers provide fault tolerance through use of a quorum, and how to further improve control plane resilience through use of infrastructure zones and Enterprise redundancy zones.
 ---
 
-# Fault Tolerance
+# Fault (Failure) Tolerance
 
-Fault tolerance is the ability of a system to continue operating without interruption
+Fault tolerance and failure tolerance are often used interchangably. Fault tolerance is the ability of a system to continue operating without interruption
```
jkirschner-hashicorp, line 10: I agree with the need to state that fault tolerance and failure tolerance are interchangeable terms. However, this is an auxiliary detail, not the most important thing about this topic, so it shouldn't be the only thing contained in the first sentence.

Maybe something like this?
```suggestion
Fault tolerance, also known as failure tolerance, is the ability of a system to continue operating without interruption
```

@im2nguyen : What do you think?

jkirschner-hashicorp (changes requested):
Small comment. Thanks for this clarification, Emma :)

PR #16057, opened 2023-01-24
Match route and listener protocols when binding

t-eckert:
### Description

When a route is bound to a gateway, the listener it gets bound to must match the protocol of the route. An HTTPRoute can only bind to a listener with the HTTP protocol.

This PR adds a check when binding which prevents binding when the route/listener protocols do not match and returns an error. One exception to this rule occurs when a user tries to bind a route to all listeners on a gateway using a wildcard `""`. As per the API Gateway spec, so long as this is able to match the protocol and bind to at least one listener on the gateway, protocol mismatches do not match and do not return an error.

* Unit tests!

### PR Checklist

* [x] updated test coverage
* ~[ ] external facing docs updated~
* [x] not a security concern


agent/consul/gateways/gateway_meta.go
```diff
@@ -0,0 +1,143 @@
...
+import (
+	"fmt"
+
+	"github.com/hashicorp/consul/acl"
+	"github.com/hashicorp/consul/agent/consul/state"
+	"github.com/hashicorp/consul/agent/structs"
+)
+
+// GatewayMeta embeds both a BoundAPIGateway and its corresponding APIGateway.
+type GatewayMeta struct {
```
t-eckert, line 12: I am open to a better name...

agent/consul/gateways/gateway_meta.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// UpdateGatewayMeta takes a parent resource reference and a BoundRoute and
+// modifies the listeners on the BoundAPIGateway config entry in GatewayMeta
+// to reflect the binding of the route to the gateway.
+//
+// If the reference is not valid or the route's protocol does not match the
+// targeted listener's protocol, a mapping of parent references to associated
+// errors is returned.
+func (g *GatewayMeta) UpdateRouteBinding(refs []structs.ResourceReference, route structs.BoundRoute) (bool, map[structs.ResourceReference]error) {
```
t-eckert, line 46: I didn't add tests for this directly because they would essentially repeat the tests for the `BindRoutesToGateways` function... That function is a kind of wrapper around this that filters out irrelevant binding cases before going through this process.

agent/structs/config_entry_gateways.go
```diff
@@ -980,77 +980,6 @@ func (e *BoundAPIGatewayConfigEntry) GetEnterpriseMeta() *acl.EnterpriseMeta {
 	return &e.EnterpriseMeta
 }
 
-func (e *BoundAPIGatewayConfigEntry) UpdateRouteBinding(refs []ResourceReference, route BoundRoute) (bool, map[ResourceReference]error) {
```
t-eckert, line 983: This functionality was moved to the `GatewayMeta` struct where it has the necessary context to match listener/route protocols.

agent/consul/gateways/gateway_meta.go
```diff
@@ -0,0 +1,143 @@
...
+			if didBind {
+				didUpdate = true
+			}
+		}
+	}
+
+	return didUpdate, errors
+}
+
+func (g *GatewayMeta) BindRoute(ref structs.ResourceReference, route structs.BoundRoute) (bool, error) {
```
t-eckert, line 83: Should this and `UnbindRoute` be made private? 🤷🏻‍♂️

t-eckert:
I have been seeing some flaky tests fail in CircleCI. They don't seem to be related to what I've changed.

agent/consul/gateways/gateway_meta.go
```diff
@@ -0,0 +1,143 @@
...
+
+	"github.com/hashicorp/consul/acl"
+	"github.com/hashicorp/consul/agent/consul/state"
+	"github.com/hashicorp/consul/agent/structs"
+)
+
+// GatewayMeta embeds both a BoundAPIGateway and its corresponding APIGateway.
+type GatewayMeta struct {
+	// Bound is the bound config entry for a given gateway.
+	Bound *structs.BoundAPIGatewayConfigEntry
```
nathancoleman, line 14: I would probably use `BoundGateway` here just to consistently have a noun

agent/consul/gateways/gateway_meta.go
```diff
@@ -0,0 +1,143 @@
...
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/consul/acl"
+	"github.com/hashicorp/consul/agent/consul/state"
+	"github.com/hashicorp/consul/agent/structs"
+)
+
+// GatewayMeta embeds both a BoundAPIGateway and its corresponding APIGateway.
```
nathancoleman, line 11: Might be helpful to expand this godoc with a sentence explaining why this is useful

agent/consul/gateways/gateway_meta.go
```diff
@@ -0,0 +1,143 @@
...
+type GatewayMeta struct {
+	// Bound is the bound config entry for a given gateway.
```
nathancoleman, line 13: Just for consistency
```suggestion
	// Bound is the bound-api-gateway config entry for a given gateway.
```

agent/consul/gateways/gateway_meta.go
```diff
@@ -0,0 +1,143 @@
...
+	Bound *structs.BoundAPIGatewayConfigEntry
+	// Gateway is the APIGateway config entry for the gateway.
```
nathancoleman, line 15: Just for consistency
```suggestion
	// Gateway is the api-gateway config entry for the gateway.
```

agent/consul/gateways/gateway_meta.go
```diff
@@ -0,0 +1,143 @@
...
+	Gateway *structs.APIGatewayConfigEntry
+}
+
+// GetGatewayMeta queries the state store for an API Gateway and a Bound API
+// Gateway matching the given name and enterprise meta and returns a GatewayMeta
+// struct containing both.
+func GetGatewayMeta(store *state.Store, name string, entMeta *acl.EnterpriseMeta) (*GatewayMeta, error) {
+	_, bound, err := store.ConfigEntry(nil, structs.BoundAPIGateway, name, entMeta)
```
nathancoleman, line 23: ~~Is it possible that `bound` is `nil` here? If so, is that okay?~~

Edit: Seeing below that they can be `nil` and we handle that in various places

agent/consul/gateways/gateway_meta.go
```diff
@@ -0,0 +1,143 @@
...
+		return nil, err
+	}
+
+	return &GatewayMeta{
+		Bound:   bound.(*structs.BoundAPIGatewayConfigEntry),
+		Gateway: gateway.(*structs.APIGatewayConfigEntry),
+	}, nil
+}
+
+// UpdateGatewayMeta takes a parent resource reference and a BoundRoute and
```
nathancoleman, line 39:
```suggestion
// UpdateRouteBinding takes a parent resource reference and a BoundRoute and
```

agent/consul/gateways/gateway_meta.go
nathancoleman: If not consumed outside of the package, I always prefer to start private and limit the API surface area then export them later if need be

nathancoleman, line 83: I think a high level doc would be super helpful here with the amount of logic gates and such in the function below

nathancoleman (changes requested):
Nice work! We discussed yesterday, I'm not a huge fan of `GatewayMeta` but don't have any other ideas myself. It is internal, so we can change it later if we come up with something better

agent/consul/gateways/gateway_meta.go
```diff
@@ -0,0 +1,143 @@
...
+		return nil, err
+	}
+
+	_, gateway, err := store.ConfigEntry(nil, structs.APIGateway, name, entMeta)
+	if err != nil {
+		return nil, err
+	}
+
+	return &gatewayMeta{
+		Bound:   bound.(*structs.BoundAPIGatewayConfigEntry),
```
jm96441n, line 34: is it possible that `bound` is not actually a `*structs.BoundAPIGatewayConfigEntry` and this panics?

t-eckert: I suppose it is possible. The `ConfigEntry` function is responsible for returning a value that should match the `kind` that was passed in.

andrewstucki: In this particular case, if it was something other than `*structs.BoundAPIGatewayConfigEntry` there would be a programming error deep within the bowels of our memdb storage layer, so I'm fine with the unchecked conversion here. It's pretty common in most of the read layers of different subsystems in Consul that are interacting with memdb, so if something were that wrong in our storage layer, arguably this should panic.

agent/consul/gateways/gateway_meta.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	if len(g.BoundGateway.Listeners) == 0 {
+		return false, fmt.Errorf("route cannot bind because gateway has no listeners")
+	}
+
+	didBind := false
+	for _, listener := range g.Gateway.Listeners {
+		// A route with a section name of "" is bound to all listeners on the gateway.
+		if listener.Name == ref.SectionName || ref.SectionName == "" {
```
jm96441n, line 99: can we flatten this out a bit:
```
if listener.Name != ref.SectionName && ref.SectionName != "") {
    continue
}

if listener.Protocol == route.GetProtocol() {
....
}

if ref.SectionName != "" {
     return false, fmt.Errorf(....)
}
```

agent/consul/gateways/gateway_meta.go
```diff
@@ -0,0 +1,143 @@
...
+				return false, fmt.Errorf("failed to bind route %s to gateway %s: listener %s is not a %s listener", route.GetName(), g.Gateway.Name, listener.Name, route.GetProtocol())
+			}
+		}
+	}
+
+	if !didBind {
+		return false, fmt.Errorf("failed to bind route %s to gateway %s: no valid listener has name '%s' and uses %s protocol", route.GetName(), g.Gateway.Name, ref.SectionName, route.GetProtocol())
+	}
+
+	return true, nil
```
jm96441n, line 117: should we return `didBind` here to make it a bit clearer what we're returning?

agent/consul/gateways/gateway_meta_test.go
```diff
@@ -0,0 +1,409 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/consul/agent/consul/state"
+	"github.com/hashicorp/consul/agent/structs"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetGatewayMeta(t *testing.T) {
```
jm96441n, line 12: are these tests safe to run in parallel? if so can we add a `t.ParallelI()` here (and in the `t.Run`, we just need to make sure to capture the `tc` loop variable) https://pkg.go.dev/testing#T.Parallel

agent/consul/gateways/gateway_meta.go
jm96441n: gotcha, that's good to know

agent/consul/gateways/bind.go
```diff
@@ -45,8 +45,8 @@ func BindRoutesToGateways(gateways []*structs.BoundAPIGatewayConfigEntry, routes
 					delete(parentRefs, ref)
 				}
 			} else {
-				if gateway.UnbindRoute(route) {
-					modified = append(modified, gateway)
+				if gateway.unbindRoute(route) {
```
jm96441n, line 48: can we move this `if` to an `else if` to cut down on the nesting here since there's no other conditions here?

nathancoleman (approved):

agent/consul/gateways/gateway_meta_test.go
```diff
@@ -0,0 +1,413 @@
...
```
andrewstucki, line 12: Since this is the only thing that invokes `getGatewayMeta` think we can 💧 it too.

andrewstucki (changes requested):
Quick change about dropping something that isn't currently used anywhere in the binding code and is a bit different than we'll actually need in our controllers. Aside from that LGTM.

andrewstucki (approved):
Approving if you can get the tests to pass (they look like they're failing on some sort of unrelated panic in the arm64 tests, so hopefully this'll just be remerging from main -- if not, might want to raise it the rest of Consul eng).

PR #16058, opened 2023-01-25
API Gateway Controller Logic

sarahalsmiller:
### Description
Adding logic into API gateway controller stubs and added unit tests for testing the functionality. 

* Added unit tests, code is not currently actively hooked up to anything. 

### Links

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


agent/consul/gateways/controller_gateways.go
```diff
@@ -2,27 +2,166 @@ package gateways
...
 
 func (r apiGatewayReconciler) Reconcile(ctx context.Context, req controller.Request) error {
+	//TODO thomases wrapper function
+	entry, err := r.store.GetConfigEntry(req.Kind, req.Name, req.Meta)
+	if err != nil {
+		return err
+	}
+
+	if entry == nil {
+		r.logger.Error("cleaning up deleted gateway object", "request", req)
```
t-eckert, line 56: I'm not so sure this should be an error. It might be better as an info log. This cleanup is a normal process when the gateway is deleted.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,27 +2,166 @@ package gateways
...
+		if err := r.store.Delete(&structs.BoundAPIGatewayConfigEntry{
+			Kind:           structs.BoundAPIGateway,
+			Name:           req.Name,
+			EnterpriseMeta: *req.Meta,
+		}); err != nil {
```
t-eckert, line 61: Might be good to log the error out here before returning.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,27 +2,166 @@ package gateways
...
+
+	gatewayEntry := entry.(*structs.APIGatewayConfigEntry)
+	//TODO is this what needs to happen for the validation step
+	err = gatewayEntry.Validate()
+	if err != nil {
+		r.logger.Debug("persisting gateway status", "gateway", gatewayEntry)
+		if err := r.store.UpdateStatus(gatewayEntry); err != nil {
+			return err
+		}
+		return err
```
t-eckert, line 75: If the gateway entry fails to validate and returns an error, then the status update fails too, you will lose the context of the validation error when you return.

I might suggest doing this instead:

```go
if updateErr := r.store.UpdateStatus(gatewayEntry); updateErr != nil {
    return fmt.Errorf("%v: %v", err, updateErr)
}
return err
```

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,27 +2,166 @@ package gateways
...
+	}
+	boundGateways, routeErrors := BindRoutesToGateways(wrapGatewaysInSlice(boundGatewayEntry), routes...)
+
+	if len(boundGateways) > 1 {
+		r.logger.Warn("imlpementation error, state should be impossible in gateway reconciler")
+		return errors.New("incorrect number of gateways bound")
+	} else if len(boundGateways) == 0 && len(routeErrors) == 0 {
+		r.logger.Debug("gateway had no updates to make ")
+		return nil
+	}
```
t-eckert, line 108: In the context of debugging, I think it would be helpful to give more metadata about the gateway when creating the errors and debug logging. 

That first state should be impossible here, but users do sometimes make the impossible possible.

I might try this instead:

```suggestion
	if len(boundGateways) > 1 {
	    err := fmt.Errorf("bind returned more gateways (%d) than it was given (1)", len(boundGateways))
	    r.logger.Errorf("API Gateway Reconciler failed to reconcile: %v", err)
		return err
	} 
	
	if len(boundGateways) == 0 && len(routeErrors) == 0 {
		r.logger.Debugf("API Gateway Reconciler: gateway %s reconciled without updates.")
		return nil
	}
```

t-eckert (commented):
Did a first look through. Once my PR goes in, let's pair on integrating that here.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,27 +2,208 @@ package gateways
...
+			return err
+		}
+		return nil
+	}
+
+	gatewayEntry := entry.(*structs.APIGatewayConfigEntry)
+	err = gatewayEntry.Validate()
+	if err != nil {
+		r.logger.Debug("persisting gateway status", "gateway", gatewayEntry)
+		if updateErr := r.store.UpdateStatus(gatewayEntry); err != nil {
```
sarahalsmiller, line 69: I'm not sure how the error is currently being propagated to the user at the moment

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,27 +2,208 @@ package gateways
...
+
+	if entry == nil {
+		r.logger.Info("cleaning up deleted gateway object", "request", req)
+		if err := r.store.Delete(&structs.BoundAPIGatewayConfigEntry{
+			Kind:           structs.BoundAPIGateway,
+			Name:           req.Name,
+			EnterpriseMeta: *req.Meta,
+		}); err != nil {
+			r.logger.Error("error cleaning up deleted gateway object", err)
+			return err
```
jm96441n, line 60: should we also wrap this error with that additional context  that's in the log message?

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,27 +2,208 @@ package gateways
...
+	}
+
+	gatewayEntry := entry.(*structs.APIGatewayConfigEntry)
+	err = gatewayEntry.Validate()
+	if err != nil {
+		r.logger.Debug("persisting gateway status", "gateway", gatewayEntry)
+		if updateErr := r.store.UpdateStatus(gatewayEntry); err != nil {
+			return fmt.Errorf("%v: %v", err, updateErr)
+		}
+		return err
```
jm96441n, line 72: should we just `return nil` here so it's explicit that this is a nil value in the case of no error? I always have to double take and re-read the implementation to see if it can be an actual error or not

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,27 +2,208 @@ package gateways
...
+	}
+	//TODO not implemented
+	//for _, r := range httpRoutes {
+	//	routes = append(routes, r.(*structs.HTTPRouteConfigEntry))
+	//}
+	return routes, nil
 }
 
 func (r apiGatewayReconciler) Reconcile(ctx context.Context, req controller.Request) error {
+	entry, err := r.store.GetConfigEntry(req.Kind, req.Name, req.Meta)
```
jm96441n, line 47: can we break this function up a bit?

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,27 +2,208 @@ package gateways
...
+			EnterpriseMeta: *req.Meta,
+		}); err != nil {
+			r.logger.Error("error cleaning up deleted gateway object", err)
+			return err
+		}
+		return nil
+	}
+
+	gatewayEntry := entry.(*structs.APIGatewayConfigEntry)
+	err = gatewayEntry.Validate()
```
andrewstucki, line 66: So, I think we can actually drop this and the block below. `Validate` is called on writing the entry and blocks the actual config entry being written to the store. At this point in the controller we should be able to assume that the entry has successfully passed validation, and, if not, we definitely shouldn't return an error in the reconciliation loop, as a failed `Validate` will mean that the entry is invalid on disk, so we'll just re-queue it until kingdom come.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,27 +2,208 @@ package gateways
...
+
+	metaGateway := &gatewayMeta{
+		BoundGateway: boundGatewayEntry,
+		Gateway:      gatewayEntry,
+	}
+
+	//TODO is this needed? It seems to be for my tests
+	r.reconcileListeners(metaGateway)
+	fmt.Println(metaGateway.Gateway.Listeners)
+	fmt.Println(metaGateway.BoundGateway.Listeners)
```
jm96441n, line 106: should these be log statements?

agent/consul/gateways/controller_gateways_test.go
```diff
@@ -0,0 +1,123 @@
...
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			r := apiGatewayReconciler{
+				logger: tt.fields.logger,
+				store:  tt.fields.store,
+			}
+			if err := r.Reconcile(tt.args.ctx, tt.args.req); (err != nil) != tt.wantErr {
```
jm96441n, line 68: are we planning on adding more tests around the sad path? if not can we simplify the error check to just  be `if err != nil`, if not would it make sense to use any sort of sentinel errors and use `errors.Is` around the type?

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,27 +2,208 @@ package gateways
...
+		return err
+	}
+
+	metaGateway := &gatewayMeta{
+		BoundGateway: boundGatewayEntry,
+		Gateway:      gatewayEntry,
+	}
+
+	//TODO is this needed? It seems to be for my tests
+	r.reconcileListeners(metaGateway)
```
andrewstucki, line 104: So, I would potentially merge this and the logic from lines 80-88 in something like an `ensureBoundGateway` method, because from the implementation what this is doing is just ensuring that the bound gateway entry initialized or read previously has the proper listener state assigned to it.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,27 +2,208 @@ package gateways
...
+	fmt.Println(metaGateway.Gateway.Listeners)
+	fmt.Println(metaGateway.BoundGateway.Listeners)
+
+	boundGateways, routeErrors := BindRoutesToGateways(wrapSlice(metaGateway), routes...)
+
+	if len(boundGateways) > 1 {
```
andrewstucki, line 110: If we really want this check, we should probably put it as an assertion within the binding helper itself, I don't think we need it though.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,27 +2,208 @@ package gateways
...
+		err := fmt.Errorf("bind returned more gateways (%d) than it was given (1)", len(boundGateways))
+		r.logger.Error("API Gateway Reconciler failed to reconcile: %v", err)
+
+		return err
+	} else if len(routeErrors) > 0 && len(boundGateways) > 0 {
+		err := fmt.Errorf("bind returned no gateways and (%d) errors: %v", len(routeErrors), routeErrors)
```
andrewstucki, line 116: So this is currently saying, "if I have any route binding errors and successfully bound gateways", it shouldn't error out in this case. Really, the only thing that you'll want to do is a bounds check on `boundGateways` prior to indexing it at line 127. As the code is currently, if `routeErrors` > 0 and `boundGateways` == 0, then your code will panic.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,27 +2,208 @@ package gateways
...
+		r.logger.Error("API Gateway Reconciler failed to reconcile: %v", err)
+		return err
+
+	}
+
+	if len(boundGateways) == 0 && len(routeErrors) == 0 {
```
andrewstucki, line 122: Not sure we need this shortcut, especially since we'll want to persist any changes to the gateway status regardless to show it was reconciled.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,27 +2,208 @@ package gateways
...
+		r.logger.Debug("API Gateway Reconciler: gateway %s reconciled without updates.")
+		return nil
+	}
+
+	boundGateway := boundGateways[0]
```
andrewstucki, line 127: You'll want to bounds check this, or just put it in an iteration loop for updating.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,27 +2,208 @@ package gateways
...
+	}
+
+	// then update the gateway status
+	r.logger.Debug("persisting gateway status", "gateway", gatewayEntry)
+	if err := r.store.UpdateStatus(gatewayEntry); err != nil {
+		return err
+	}
+
+	//// and update the route statuses
+	for _, listener := range boundGateway.Listeners {
```
andrewstucki, line 143: I believe it should be enough to just iterate over `routeErrors` here.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,27 +2,208 @@ package gateways
...
+	if err != nil {
+		return err
+	}
+
+	metaGateway := &gatewayMeta{
+		BoundGateway: boundGatewayEntry,
+		Gateway:      gatewayEntry,
+	}
+
+	//TODO is this needed? It seems to be for my tests
```
t-eckert, line 103: This might be an issue related to using a value receiver.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,27 +2,208 @@ package gateways
...
+	r.reconcileListeners(metaGateway)
+	fmt.Println(metaGateway.Gateway.Listeners)
+	fmt.Println(metaGateway.BoundGateway.Listeners)
+
+	boundGateways, routeErrors := BindRoutesToGateways(wrapSlice(metaGateway), routes...)
```
t-eckert, line 108: I might do the wrapping in line to reduce indirection, just `[]gatewayMeta{metaGateway}`.

t-eckert (commented):
Did another look through. I would take a look at using value/pointer receivers as potentially being the root of some of the problems you were seeing.

agent/consul/gateways/controller_gateways.go
t-eckert: In this case, the `err` will never be `nil`. This block is entered by that condition and additional error context is added if there is an error updating the status.

jm96441n: ah good catch

agent/consul/gateways/controller_gateways.go
jm96441n: in this case should this be `updateErr != nil` since we already know `err` is non nil?

agent/consul/gateways/controller_gateways.go
sarahalsmiller: No, just some lines that needed to be cleaned up, good catch

agent/consul/gateways/controller_gateways.go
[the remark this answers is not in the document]
sarahalsmiller: The BoundAPIGateway actually did get cast later, it was just hidden in the else statement

agent/consul/fsm_data_store.go
```diff
@@ -0,0 +1,53 @@
...
+	return err
+}
+
+func (f FSMDataStore) Delete(entry structs.ConfigEntry) error {
+	_, err := f.s.leaderRaftApply("ConfigEntry.Delete", structs.ConfigEntryRequestType, &structs.ConfigEntryRequest{
+		Op:    structs.ConfigEntryDelete,
+		Entry: entry,
+	})
+	return err
+}
```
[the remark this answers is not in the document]
sarahalsmiller: Updated everywhere that issue happens, I believe.

agent/consul/gateways/controller_gateways_test.go
sarahalsmiller: Definitely planning on adding more tests in the future

sarahalsmiller: Probably would make sense, pretty sure this was just the autogenerated boilerplate test structure

agent/consul/fsm_data_store.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/hashicorp/consul/agent/structs"
+)
+
+// implementation of consul/gateways/datastore
+type FSMDataStore struct {
+	s   *Server
+	fsm *fsm.FSM
+}
+
+func (f FSMDataStore) GetConfigEntry(kind string, name string, meta *acl.EnterpriseMeta) (*structs.ConfigEntry, error) {
```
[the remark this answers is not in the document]
sarahalsmiller: Done.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,27 +2,208 @@ package gateways
...
+	}
+	if boundEntry == nil {
+		boundGatewayEntry = &structs.BoundAPIGatewayConfigEntry{
+			Kind:           structs.BoundAPIGateway,
+			Name:           gatewayEntry.Name,
+			EnterpriseMeta: gatewayEntry.EnterpriseMeta,
+		}
+	} else {
+		boundGatewayEntry = boundEntry.(*structs.BoundAPIGatewayConfigEntry)
+	}
```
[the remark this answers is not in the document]
sarahalsmiller: Same issue I noted above "cannot assign in multiple assignment" error

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,33 +2,202 @@ package gateways
...
 	)
 }
+
+// Reconcile takes in a controller request and ensures this api gateways corresponding BoundAPIGateway exists and is
+// up to date
+func (r *apiGatewayReconciler) Reconcile(ctx context.Context, req controller.Request) error {
+	metaGateway, err := r.initGatewayMeta(req)
+	if err != nil {
+		return err
+	} else if metaGateway == nil && err == nil {
```
jm96441n, line 39: I don't think we need the `&& err == nil` here because it can't be non-nil here from the previous condition, we can also probably just make this `if metaGateway == nil` and not need the `else if` because we're exiting early if `err != nil`

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,33 +2,202 @@ package gateways
...
+
+	routes, err := r.retrieveAllRoutesFromStore()
+
+	if err != nil {
+		return err
+	}
+
+	boundGateways, routeErrors := BindRoutesToGateways([]*gatewayMeta{metaGateway}, routes...)
+
+	//In this loop there should only be 1 bound gateway returned, but looping over all returned gateways
```
jm96441n, line 66: is there ever a case where more than one could be returned?

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,33 +2,202 @@ package gateways
...
+	metaGateway := &gatewayMeta{}
+
+	apiGateway, err := r.store.GetConfigEntry(req.Kind, req.Name, req.Meta)
+	if err != nil {
+		return nil, err
+	}
+
+	if apiGateway == nil {
+		//gateway needs to be deleted, don't init
+		return nil, nil
```
jm96441n, line 132: minor: would this technically be an error condition because we're trying to init a gateway that should be deleted?

agent/consul/gateways/controller_gateways.go
sarahalsmiller, line 39:
```suggestion
	} else if metaGateway == nil {
```

agent/consul/gateways/controller_gateways.go
sarahalsmiller: My logic was that it being nil is a valid state for it to be in the store, but if it would be clearer to return an error I'm happy to do it that way.

agent/consul/gateways/controller_gateways.go
sarahalsmiller: I don't believe there should be in this particular controller since only 1 gateway is being passed in. More likely what will happen is 0 boundGateways and multiple errors.

t-eckert (approved):
Sarah, this is looking great! You've done a fantastic job handling all the complex cases and responding to feedback. I love it!

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,33 +2,202 @@ package gateways
...
+		return nil
+	}
+
+	r.ensureBoundGateway(metaGateway)
+
+	r.logger.Debug("started reconciling gateway")
+
+	routes, err := r.retrieveAllRoutesFromStore()
+
+	if err != nil {
```
nathancoleman, line 60: #nit I like to keep these grouped together for readability
```suggestion
	routes, err := r.retrieveAllRoutesFromStore()
	if err != nil {
```

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,33 +2,202 @@ package gateways
...
+			msg := "error cleaning up deleted gateway object"
+			r.logger.Error(msg, err)
+			return errors.Wrap(err, msg)
+		}
+		return nil
+	}
+
+	r.ensureBoundGateway(metaGateway)
+
+	r.logger.Debug("started reconciling gateway")
```
nathancoleman, line 56: Would it be helpful in a debug scenario to have this logged earlier, before we potentially clean up a deleted gateway?

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,33 +2,202 @@ package gateways
...
+		}
+
+		// then update the gateway status
+		r.logger.Debug("persisting gateway status", "gateway", metaGateway.Gateway)
+		if err := r.store.UpdateStatus(metaGateway.Gateway, err); err != nil {
+			return err
+		}
+	}
+
+	//// and update the route statuses
```
nathancoleman, line 84:
```suggestion
	// and update the route statuses
```

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,33 +2,202 @@ package gateways
...
+		if err := r.store.Update(boundGateway); err != nil {
+			msg := "error persisting state"
+			r.logger.Error(msg, "error", err)
+			return errors.Wrap(err, msg)
+		}
+
+		// then update the gateway status
+		r.logger.Debug("persisting gateway status", "gateway", metaGateway.Gateway)
+		if err := r.store.UpdateStatus(metaGateway.Gateway, err); err != nil {
+			return err
```
nathancoleman, line 80: Is this case worth logging?

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,33 +2,202 @@ package gateways
...
+		}
+	}
+
+	//// and update the route statuses
+	for route, routeError := range routeErrors {
+		configEntry := r.resourceReferenceToBoundRoute(route)
+		r.logger.Error("route binding error:", routeError)
+		if err := r.store.UpdateStatus(configEntry, routeError); err != nil {
+			return err
```
nathancoleman, line 89: Does it make sense to return here vs. continuing to process the other routes and their errors?

agent/consul/gateways/controller_gateways.go
nathancoleman: I think "return `nil` if the gateway doesn't exist" is good logic. The comment here is what's throwing me off. I don't feel like this function should be prescribing what the caller should do if the api-gateway doesn't exist

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,33 +2,202 @@ package gateways
...
+		return nil, err
+	}
+
+	//initialize object, values get copied over in ensureBoundGateway if they don't exist
+	metaGateway.BoundGateway = boundGateway.(*structs.BoundAPIGatewayConfigEntry)
+	return metaGateway, nil
+}
+
+func (r *apiGatewayReconciler) resourceReferenceToBoundRoute(ref structs.ResourceReference) structs.ControlledConfigEntry {
+	//TODO currently have to retrieve from the store to persist parent field on update call, is there a better way to do this?
```
nathancoleman, line 148: Not quite sure what you mean here. I'll check in with you on a video chat today

agent/consul/gateways/datastore/datastore.go
```diff
@@ -0,0 +1,15 @@
+package datastore
```
nathancoleman, line 1: This feels like we're creating a package for a single interface, and then we wind up stuttering `datastore.DataStore`. Is this beneficial over just putting the interface in the parent `gateways` package?

nathancoleman (commented):
Some thoughts and questions here. Nothing I would block merge over

agent/consul/gateways/datastore/datastore.go
sarahalsmiller: I think at the time I had been worried about potential circular dependencies, but it ended up not being an issue. I can pull it into the gateways package.

agent/consul/gateways/controller_gateways.go
sarahalsmiller: I suppose it depends on how often the reconciliation loop will be run? I'd imagine if theres an error updating the status it will likely apply to the other routes as well since I'd imagine that it would be an issue with the store itself.

agent/consul/gateways/controller_gateways.go
sarahalsmiller: Available to talk through this whenever, but my concern was if I were to update the status with a niled out parent field it would nil out the parent field in the store as well.

nathancoleman: We chatted and the gist is:
1. Receive reference to config entry, we need to update the status
2. Update the status
3. Write config entry w/ status to the store

The reason we fetch the config entry here is so as not to wipe out the config entry contents when trying to write the status

PR #16070, opened 2023-01-26
add troubleshoot cli

malizz:
### Description
Enterprise PR: https://github.com/hashicorp/consul-enterprise/pull/4079

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


ndhanushkodi (approved):

command/registry.go
```diff
@@ -240,6 +242,8 @@ func RegisteredCommands(ui cli.Ui) map[string]mcli.CommandFactory {
 		entry{"tls ca create", func(ui cli.Ui) (cli.Command, error) { return tlscacreate.New(ui), nil }},
 		entry{"tls cert", func(ui cli.Ui) (cli.Command, error) { return tlscert.New(), nil }},
 		entry{"tls cert create", func(ui cli.Ui) (cli.Command, error) { return tlscertcreate.New(ui), nil }},
+		entry{"troubleshoot connect", func(ui cli.Ui) (cli.Command, error) { return troubleshoot.New(ui), nil }},
```
DanStough, line 245: If you follow the pattern of the other commands, there is usually an entry at the subcommand level to provide intermediate help and usage guideline, e.g. [tls.go](https://github.com/hashicorp/consul-enterprise/blob/eaca8f33f71da474d45dccddcbf1ce3f027b1ee4/command/tls/tls.go#L10-L10)

command/troubleshoot/connect/troubleshoot_connect.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (c *cmd) Help() string {
+	return c.help
+}
+
+const (
+	synopsis = "Troubleshoots service mesh issues"
+	help     = `
+Usage: consul troubleshoot proxy [options]
```
DanStough, line 99:
```suggestion
Usage: consul troubleshoot connect [options]
```
Not sure if this line needs to change or if it's the package name. TBH if we are trying to get rid of `connect`, maybe the command should be `troubleshoot proxy`.

command/troubleshoot/connect/troubleshoot_connect.go
```diff
@@ -0,0 +1,109 @@
...
+  Connects to local envoy proxy and troubleshoots service mesh communication issues.
+  Requires an upstream service SNI.
```
DanStough, line 101: I think the term SNI is probably confusing for most practitioners. Can we provide an example here and as part of the flag docs, or maybe we just get rid of the term?

command/troubleshoot/upstreams/troubleshoot_upstreams.go
```diff
@@ -0,0 +1,102 @@
...
+
+	defaultAdminBind := "localhost:19000"
+	if adminBind := os.Getenv("ADMIN_BIND"); adminBind != "" {
+		defaultAdminBind = adminBind
+	}
+	c.flags.StringVar(&c.adminBind, "admin-bind", defaultAdminBind, "The address:port that envoy's admin endpoint is on.")
+
+	c.http = &flags.HTTPFlags{}
+	flags.Merge(c.flags, c.http.ClientFlags())
+	flags.Merge(c.flags, c.http.ServerFlags())
```
DanStough, line 41: Will this need to support Enterprise flags, e.g. Partitions? Sometimes these still go into OSS, e.g. [peering list](https://github.com/hashicorp/consul-enterprise/blob/eaca8f33f71da474d45dccddcbf1ce3f027b1ee4/command/peering/list/list.go#L46-L47).

command/troubleshoot/connect/troubleshoot_connect.go
```diff
@@ -0,0 +1,109 @@
...
+
+	defaultAdminBind := "localhost:19000"
+	if adminBind := os.Getenv("ADMIN_BIND"); adminBind != "" {
+		defaultAdminBind = adminBind
+	}
+	c.flags.StringVar(&c.adminBind, "admin-bind", defaultAdminBind, "The address:port that envoy's admin endpoint is on.")
+
+	c.http = &flags.HTTPFlags{}
+	flags.Merge(c.flags, c.http.ClientFlags())
+	flags.Merge(c.flags, c.http.ServerFlags())
```
DanStough, line 44: Same comment about Enterprise flags.

command/troubleshoot/upstreams/troubleshoot_upstreams.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+const (
+	synopsis = "Troubleshoots service mesh issues"
+	help     = `
+Usage: consul troubleshoot upstreams [options]
+  
+  Connects to local Envoy and lists upstream service envoy IDs.
+  This command is used in combination with 
+  'consul troubleshoot connect' to diagnose issues in Consul service mesh.
```
DanStough, line 98: Consistentcy with whatever you device in the previous case.

DanStough (changes requested):
Some minor stuff, but I would feel better if we resolved some comments before merging.

malizz:
Thank you @DanStough for the review, I'm going to address them in a second set or PRs, to keep this commit consistent with ENT.

DanStough (approved):
Approving for now since this is a placeholder.

command/troubleshoot/upstreams/troubleshoot_upstreams.go
malizz: We don't have any enterprise-specific flags atm.

DanStough: OK. I imagine at some point you'll need them to query for services/catalog information in the correct partition/namespace. It might work on agent-ful consul because the client agent belongs to a partition.

PR #16085, opened 2023-01-26
feat: apply retry policy to read only grpc endpoints

JadhavPoonam:
### Description
The gRPC service config is extended to include all read-only endpoints with the retry policy same as the RESOURCE_EXHAUSTED error.

### PR Checklist

* [ ] ~updated test coverage~


boxofrad (approved):
LGTM ⭐ 

I wonder if we should replace my `TODO` comment above with something that says these are hard-coded for now, but we're working on generating them automatically from the protobuf files?

agent/grpc-internal/client.go
```diff
@@ -52,6 +52,79 @@ const grpcServiceConfig = `
...
+				{
+					"service": "partition.PartitionService",
+					"method": "Read"
+				}
+			],
+			"retryPolicy": {
+				"MaxAttempts": 5,
+				"BackoffMultiplier": 2,
+				"InitialBackoff": "1s",
+				"MaxBackoff": "5s",
```
dhiaayachi, line 115: If my understanding is right, this would lead to the following schedule: 1s, 2s,4s. 
IS that what we intend? I thought we wanted a 5 time retry.

agent/grpc-internal/client.go
```diff
@@ -52,6 +52,79 @@ const grpcServiceConfig = `
 				"MaxBackoff": "5s",
 				"RetryableStatusCodes": ["RESOURCE_EXHAUSTED"]
```
dhiaayachi, line 53: What's the difference between the policy defined in here and the one below?

dhiaayachi (commented):
Thank you @JadhavPoonam for putting this together!
I have few questions around the intended policy.

agent/grpc-internal/client.go
JadhavPoonam: The first one makes read-and-write calls to retry only when rate-limited while the second one is defined only for read requests that are errored due to a server error.

agent/grpc-internal/client.go
JadhavPoonam: Actually it's a random number between 0 and 5 as per https://github.com/grpc/grpc-proto/blob/master/grpc/service_config/service_config.proto#L145

dhiaayachi (approved):
Thank you for answering my questions @JadhavPoonam 

LGTM!!

PR #16099, opened 2023-01-27
feat: update rpc retry criteria

JadhavPoonam:
### Description
This change updates the the retry criteria to accommodate following use cases:
```
- Client RPC to Server -> Retry should be in the client, only the "ErrRetryElsewhere" error
- Client RPC to Follower Forwarded to Leader -> Retry should be implemented in the client, only the "ErrRetryElsewhere" error
- Server RPC executed locally -> Should NOT retry
- Server RPC forwarded to Leader -> Should retry, only "ErrRetryLater" error

```
### PR Checklist

* [ ] ~updated test coverage~
* [x] not a security concern

### DIFF
--- a/agent/consul/client.go
+++ b/agent/consul/client.go
@@ -16,6 +16,7 @@ import (
 	"golang.org/x/time/rate"
 
 	"github.com/hashicorp/consul/acl"
+	rpcRate "github.com/hashicorp/consul/agent/consul/rate"
 	"github.com/hashicorp/consul/agent/pool"
 	"github.com/hashicorp/consul/agent/router"
 	"github.com/hashicorp/consul/agent/structs"
@@ -296,7 +297,16 @@ TRY:
 
 	// Use the zero value for RPCInfo if the request doesn't implement RPCInfo
 	info, _ := args.(structs.RPCInfo)
-	if retry := canRetry(info, rpcErr, firstCheck, c.config); !retry {
+	retryableMessages := []error{
+		// If we are chunking and it doesn't seem to have completed, try again.
+		ErrChunkingResubmit,
+
+		// These rate limit errors are returned before the handler is called, so are
+		// safe to retry.
+		rpcRate.ErrRetryElsewhere,
+	}
+
+	if retry := canRetry(info, rpcErr, firstCheck, c.config, retryableMessages); !retry {
 		c.logger.Error("RPC failed to server",
 			"method", method,
 			"server", server.Addr,
--- a/agent/consul/rpc.go
+++ b/agent/consul/rpc.go
@@ -557,7 +557,7 @@ func (c *limitedConn) Read(b []byte) (n int, err error) {
 }
 
 // canRetry returns true if the request and error indicate that a retry is safe.
-func canRetry(info structs.RPCInfo, err error, start time.Time, config *Config) bool {
+func canRetry(info structs.RPCInfo, err error, start time.Time, config *Config, retryableMessages []error) bool {
 	if info != nil {
 		timedOut, timeoutError := info.HasTimedOut(start, config.RPCHoldTimeout, config.MaxQueryTime, config.DefaultQueryTime)
 		if timeoutError != nil {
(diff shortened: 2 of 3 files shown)


dhiaayachi (changes requested):
Thank you @JadhavPoonam for working on this.

Looking to the change in here, I think it make the canRetry logic a bit confusing. 

Another approach that I think would be better, is to extract retriableErrors as a func parameter and pass it in both cases. What you think about that?

agent/consul/rpc.go
```diff
@@ -746,6 +745,10 @@ CHECK_LEADER:
 			return true, nil
 		}
 	}
+	// defer to caller to retry
+	if rpcErr != rate.ErrRetryElsewhere {
```
boxofrad, line 749: tip: use `errors.Is` to handle cases where the error has been [wrapped](https://rollbar.com/blog/golang-wrap-and-unwrap-error).
```suggestion
	if !errors.Is(rpcErr, rate.ErrRetryElsewhere) {
```

boxofrad (approved):

JadhavPoonam:
> Thank you @JadhavPoonam for working on this.
> 
> Looking to the change in here, I think it make the canRetry logic a bit confusing.
> 
> Another approach that I think would be better, is to extract retriableErrors as a func parameter and pass it in both cases. What you think about that?

I have passed the retryable messages separately now based on the following criteria:

1. On the client side we want to retry for `ErrChunkingResubmit` and rate limiting type error of `ErrRetryElsewhere`
2. In the `forwardRequestToLeader` method we do not want to retry for any rate limit errors so I only left `ErrChunkingResubmit` in there

dhiaayachi:
@JadhavPoonam we need to add retries in the server RPC for the rate limiter errors, this is to cover the case of direct HTTP to a server.

agent/consul/client.go
```diff
@@ -306,12 +316,6 @@ TRY:
 		return rpcErr
 	}
 
-	c.logger.Warn("Retrying RPC to server",
-		"method", method,
-		"server", server.Addr,
-		"error", rpcErr,
-	)
-
```
dhiaayachi, line 314: Sorry for the confusion around this. We should put this back as this log would happen client side while the rate limiting log happen server side.

agent/consul/server.go
```diff
@@ -1552,8 +1553,29 @@ func (s *Server) RPC(ctx context.Context, method string, args interface{}, reply
 		metrics.IncrCounter([]string{"client", "rpc", "exceeded"}, 1)
 		return structs.ErrRPCRateExceeded
 	}
+
+TRY:
```
dhiaayachi, line 1557: Thinking more about this, retrying here does not make any sense. As the error is retry elsewhere and we can't target another server from this server (local request). The error should be returned to the caller to retry elsewhere. sorry for the confusion again! 😓

agent/consul/server.go
```diff
@@ -1552,8 +1553,29 @@ func (s *Server) RPC(ctx context.Context, method string, args interface{}, reply
...
+		info, _ := args.(structs.RPCInfo)
+
+		retryableMessages := []error{
+			// These rate limit errors are returned before the handler is called, so are
+			// safe to retry.
+			rpcRate.ErrRetryElsewhere,
+		}
+
+		if retry := canRetry(info, err, firstCheck, s.config, retryableMessages); !retry {
+			metrics.IncrCounterWithLabels([]string{"client", "rpc", "failed"}, 1, []metrics.Label{{Name: "server", Value: s.config.NodeName}})
```
dhiaayachi, line 1568: The metrics error we are getting is because of this added metric call which is not tested, but I think with the comment above around no need to retry it should go away.

dhiaayachi (changes requested):
@JadhavPoonam I left some comments, mostly contradicting what I said last week. Sorry for the confusion around this!

dhiaayachi (approved):
LGTM! Thank you @JadhavPoonam!!

PR #16105, opened 2023-01-30
ACL error improvements: incomplete bootstrapping and non-existent token

skpratt:
### Description
These errors provide helpful information in the case where the ACL system has not been bootstrapped, when a token does not exist, or when it does not exist in a particular namespace.

Functionality changes
Useful errors are now returned when:
- trying to get a specific resource (specified by ID) and failing because it does not exist
- trying to delete a resource that does not exist

Table of new API error messages:
| Message | Appears when | Previously |
|----|----|----|
| Requested * not found in namespace *: ACL not found |  ACL * Read / Get (namespace) | No error, return nothing |
| Requested * does not exist: ACL not found | ACL * Read / Get | No error, return nothing |
| Cannot find * to delete | ACL * Delete | Policy / Token: “ACL Not Found”, Role / Rule / AuthMethod: No error, return ok |
| Supplied token does not exist | ACLTokenSelf, ACLLogout | “ACL Not Found” |
| ACL system must be bootstrapped before making any requests that require authorization: ACL not found | ACL Bootstrap | “ACL not found” |

Locally: spin up consul server cluster with ACLs enabled. 

- Try to interact with ACL system before bootstrapping has completed:
  - sample error, logs: `"leaf cert watch returned an error: rpc error making call: ACL system must be bootstrapped before making any requests that require authorization: ACL not found"`
  - sample error, CLI: `Failed to create new token: Unexpected response code: 403 (ACL system must be bootstrapped before making any requests that require authorization: ACL not found)`

- Delete a token and then try to read/update/clone it:
  - sample error, logs: `[ERROR] agent.http: Request error: method=GET url=/v1/acl/token/XXXX from=address:port error="Permission denied: token does not exist: ACL not found"`
  - sample error, CLI: `Error reading token "XXXXX": Unexpected response code: 401 (Requested token does not exist: ACL not found)`

### PR Checklist

* [x] updated test coverage
* [ ] not a security concern

### DIFF
--- a/.changelog/16105.txt
+++ b/.changelog/16105.txt
@@ -1,0 +1,3 @@
+```release-note:breaking-change
+acl errors: Delete and get requests now return descriptive errors when the specified resource cannot be found. Other ACL request errors provide more information about when a resource is missing.
+```
--- a/.changelog/16105.txt
+++ b/.changelog/16105.txt
@@ -5,7 +5,7 @@ acl errors: Delete and get requests now return descriptive errors when the speci
 3. Read Token/Policy/Role endpoints now return 404 when the resource cannot be found.
   - New error format: "Cannot find * to delete"
 4. Logout now returns a 401 error when the supplied token cannot be found
-  - New error format: "Token supplied with request not found"
+  - New error format: "Supplied token does not exist"
 5. Token Self endpoint now returns 404 when the token cannot be found.
   - New error format: "Supplied token does not exist"
 ```
--- a/.changelog/16105.txt
+++ b/.changelog/16105.txt
@@ -1,3 +1,11 @@
 ```release-note:breaking-change
-acl errors: Delete and get requests now return descriptive errors when the specified resource cannot be found. Other ACL request errors provide more information about when a resource is missing.
+acl errors: Delete and get requests now return descriptive errors when the specified resource cannot be found. Other ACL request errors provide more information about when a resource is missing. Add error for when the ACL system has not been bootstrapped.
+1. Delete Token/Policy/AuthMethod/Role/BindingRule endpoints now return 404 when the resource cannot be found.
+  - New error formats: "Requested * does not exist: ACL not found", "* not found in namespace $NAMESPACE: ACL not found"
+3. Read Token/Policy/Role endpoints now return 404 when the resource cannot be found.
+  - New error format: "Cannot find * to delete"
+4. Logout now returns a 401 error when the supplied token cannot be found
+  - New error format: "Token supplied with request not found"
+5. Token Self endpoint now returns 404 when the token cannot be found.
+  - New error format: "Supplied token does not exist"
 ```
--- a/agent/acl_endpoint.go
+++ b/agent/acl_endpoint.go
@@ -393,7 +393,7 @@ func (s *HTTPHandlers) ACLTokenGet(resp http.ResponseWriter, req *http.Request, 
 	}
 
 	if out.Token == nil {
-		return nil, acl.ErrNotFound
+		return nil, fmt.Errorf("token does not exist: %w", acl.ErrNotFound)
 	}
 
 	if args.Expanded {
--- a/agent/acl_endpoint.go
+++ b/agent/acl_endpoint.go
@@ -167,7 +167,11 @@ func (s *HTTPHandlers) ACLPolicyRead(resp http.ResponseWriter, req *http.Request
 	}
 
 	if out.Policy == nil {
-		// if no error was returned, the policy does not exist
+		// if no error was returned above, the policy does not exist
+		if ns := args.EnterpriseMeta.NamespaceOrEmpty(); ns != "" {
+			msg := fmt.Sprintf("Requested policy not found in namespace %s", ns)
+			return nil, HTTPError{StatusCode: http.StatusNotFound, Reason: msg}
+		}
 		return nil, HTTPError{StatusCode: http.StatusNotFound, Reason: "Requested policy does not exist"}
 	}
 
@@ -352,8 +356,8 @@ func (s *HTTPHandlers) ACLTokenSelf(resp http.ResponseWriter, req *http.Request)
 	}
 
 	if out.Token == nil {
-		// if no error was returned, the token does not exist
-		return nil, HTTPError{StatusCode: http.StatusNotFound, Reason: "Token does not exist"}
+		// if no error was returned above, the token does not exist
+		return nil, HTTPError{StatusCode: http.StatusNotFound, Reason: "Supplied token does not exist"}
 	}
 
 	return out.Token, nil
@@ -396,8 +400,12 @@ func (s *HTTPHandlers) ACLTokenGet(resp http.ResponseWriter, req *http.Request, 
 	}
 
 	if out.Token == nil {
-		// if no error was returned, token does not exist
-		return nil, HTTPError{StatusCode: http.StatusBadRequest, Reason: "Requested token does not exist"}
+		// if no error was returned above, the token does not exist
+		if ns := args.EnterpriseMeta.NamespaceOrEmpty(); ns != "" {
+			msg := fmt.Sprintf("Requested token not found in namespace %s", ns)
+			return nil, HTTPError{StatusCode: http.StatusNotFound, Reason: msg}
+		}
+		return nil, HTTPError{StatusCode: http.StatusNotFound, Reason: "Requested token does not exist"}
 	}
 
 	if args.Expanded {
@@ -454,8 +462,7 @@ func (s *HTTPHandlers) ACLTokenDelete(resp http.ResponseWriter, req *http.Reques
 	var ignored string
 	if err := s.agent.RPC(req.Context(), "ACL.TokenDelete", args, &ignored); err != nil {
 		if err == acl.ErrNotFound {
-			return nil, HTTPError{StatusCode: http.StatusNotFound, Reason: fmt.Sprintf("Token does not exist")}
-
+			return nil, HTTPError{StatusCode: http.StatusNotFound, Reason: fmt.Sprintf("Cannot find token to delete")}
 		}
 		return nil, err
 	}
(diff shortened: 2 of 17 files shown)


agent/consul/acl_endpoint.go
```diff
@@ -458,8 +461,13 @@ func (a *ACL) TokenClone(args *structs.ACLTokenSetRequest, reply *structs.ACLTok
 	_, token, err := a.srv.fsm.State().ACLTokenGetByAccessor(nil, args.ACLToken.AccessorID, &args.ACLToken.EnterpriseMeta)
 	if err != nil {
 		return err
-	} else if token == nil || token.IsExpired(time.Now()) {
-		return acl.ErrNotFound
+	} else if token == nil {
+		if ns := args.ACLToken.EnterpriseMeta.NamespaceOrEmpty(); ns != "" {
+			return fmt.Errorf("token not found in namespace %s: %w", ns, acl.ErrNotFound)
+		}
+		return fmt.Errorf("token does not exist: %w", acl.ErrNotFound)
```
pglass, line 468: minor / non-blocking: We do a sort of similar thing in multiple places. Is there a reasonable way to extract this into a re-usable function?

Maybe something simple like this.

```go
func aclResourceNotExistError(resourceType string, entMeta acl.EnterpriseMeta) error {
	if ns := entMeta.NamespaceOrEmpty(); ns != "" {
		return fmt.Errorf("%s not found in namespace %s: %w", resourceType, ns, acl.ErrNotFound)
	}
	return fmt.Errorf("%s does not exist: %w", resourceType, acl.ErrNotFound)
}

...
// And then call it like
return aclResourceNotExistError("token", args.ACLToken.EnterpriseMeta)
```

agent/consul/acl_server.go
```diff
@@ -150,7 +150,12 @@ func (s *Server) ResolveIdentityFromToken(token string) (bool, structs.ACLIdenti
 		return true, aclToken, nil
 	}
 
-	return s.InPrimaryDatacenter() || index > 0, nil, acl.ErrNotFound
+	defaultErr := acl.ErrNotFound
+	canBootstrap, _, _ := s.fsm.State().CanBootstrapACLToken()
```
pglass, line 154: Does this happen on every request where the token is not found? Does that end up being expensive?

agent/consul/acl_server.go
```diff
@@ -150,7 +150,12 @@ func (s *Server) ResolveIdentityFromToken(token string) (bool, structs.ACLIdenti
...
+	if canBootstrap {
+		defaultErr = fmt.Errorf("ACL system must be bootstrapped before making any requests that require authorization: %w", defaultErr)
```
pglass, line 156: non-blocking: When I tested this locally, I only saw this error when I provide a token (i.e. if I do not use the anonymous token):

```
$ CONSUL_HTTP_TOKEN=root ./bin/consul acl policy list
Failed to retrieve the policy list: Unexpected response code: 403 (ACL system must be bootstrapped before making any requests that require authorization: ACL not found)
$ ./bin/consul acl policy list
Failed to retrieve the policy list: Unexpected response code: 403 (Permission denied: anonymous token lacks permission 'acl:read'. The anonymous token is used implicitly when a request does not specify a token.)
```

If they haven't yet ACL bootstrapped, then in order to see the `ACL system must be bootstrapped` message, would they have to pass some token that has not yet been created (maybe a random value)? If so, it seems unlikely someone would pass a random token value. So I feel like this would be more valuable if the `ACL system must be bootstrapped` error messages is returned in the case of the anonymous token as well.

agent/acl_endpoint.go
```diff
@@ -978,7 +1031,7 @@ func (s *HTTPHandlers) ACLLogout(resp http.ResponseWriter, req *http.Request) (i
 	s.parseToken(req, &args.Token)
 
 	if args.Token == "" {
-		return nil, acl.ErrNotFound
+		return nil, HTTPError{StatusCode: http.StatusUnauthorized, Reason: "Token supplied with request not found"}
```
cizh, line 1034: I wonder if it makes a difference to say "ACL system don't recognize the token supplied in the request".
For the following reasons:
1. "not found" could imply it never existed, but "recognize" implies, it might exist before, but we don't acknowledge it know, IMO.
2. Active tense, with the subject. Maybe they've used token from elsewhere, we don't know, but we identify it's our ACL doesn't know this one.
3. I assume the request is "log out with deleting this token" and the token is not found, then the token is included ("in") the request, not "with" the request? 
I could be wrong. LMK how you think.

agent/consul/acl_server.go
skpratt: Did you test this on a single server with the -bootstrap flag?

The anonymous token gets created when a server acquires leadership (https://github.com/hashicorp/consul/blob/e91bc9c0586c5cd11b933f854165189720969340/agent/consul/leader.go#L501), which is intended so to avoid downtime / loss of functionality during the ACL enablement process. It's a bit of a tradeoff between providing a useful error and allowing some operations to continue. In fact, there is a related task for allowing anonymous token synthesis for the case where default_policy=allow to expand the ability to use the anonymous token for basic operations before leadership has been established: https://github.com/hashicorp/consul/pull/16200

In multi-server clusters, you will get this bootstrap error before the servers can be joined when you have a default deny policy. The bootstrap error generally appears a lot more with default_policy=deny, because the anonymous token becomes almost useless in that case.

It isn't great UX, but when users get the permission denied error for the anonymous token, if they try to use any non-nil / non-anonymous value as a next step, they will then get this bootstrap error which should point them in the right direction if they want to be able to create more tokens.

agent/consul/acl_server.go
skpratt: It is another call to memdb in the case where we are doing a token lookup by secret (i.e. validating the token header used), and failures in resolution, expiration, etc. don't get caught above. But significantly less expensive than than the token lookup itself 8 lines above. It'll get called when using a nonsense token value of the proper format or a token that has been completely deleted. Given that the ACLTokenGetBySecret is already the most expensive piece here, an additional fast lookup to provide clarification in these two scenarios seems worth it.

command/logout/logout_test.go
```diff
@@ -55,7 +55,7 @@ func TestLogoutCommand(t *testing.T) {
 
 		code := cmd.Run(args)
 		require.Equal(t, code, 1, "err: %s", ui.ErrorWriter.String())
-		require.Contains(t, ui.ErrorWriter.String(), "403 (ACL not found)")
+		require.Contains(t, ui.ErrorWriter.String(), "401 (Token supplied with request not found)")
```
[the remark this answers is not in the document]
skpratt: I think a case could be made for either. The way I'm thinking about it is that your authorization to destroy this token comes from the token itself - so conceptually, the completing the "delete X resource" request is secondary to having the correct token to do that. Wdyt?

agent/consul/acl_server.go
pglass: Ah.

> Did you test this on a single server with the -bootstrap flag?

Yes, though the behavior is the same with a multi-server cluster with -bootstrap-expect.

> In multi-server clusters, you will get this bootstrap error before the servers can be joined when you have a default deny policy.

Except if the requests get stuck in RPC forwarding, they hang trying to forward to the leader. But the Agent API requests that don't do RPC forwarding do show the error.

* `consul acl policy list` (for example) will hang during RPC forwarding to the leader when there's no cluster leader.
 
  ```
  $ CONSUL_HTTP_TOKEN=root consul acl policy list
  Failed to retrieve the policy list: Unexpected response code: 500 (No cluster leader)
  ```

* On the other hand, `consul members` and `consul join` will show the `ACL system must be bootstrapped` error when there is no cluster leader.

  ```
  $ consul members
  Error retrieving members: Unexpected response code: 403 (ACL system must be bootstrapped before making any 
  requests that require authorization: ACL not found)
  $ consul join 127.0.0.1:9300
  Error joining address '127.0.0.1:9300': Unexpected response code: 403 (ACL system must be bootstrapped before 
  making any requests that require authorization: ACL not found)
  Failed to join any nodes.
  ```

pglass: In any case, this is better than before 🙂 

I do think it's not likely people will try to use random token values that they haven't created yet, but seeing these errors on the `consul members` and `consul join` commands is super useful.

pglass: > Yes, though the behavior is the same with a multi-server cluster with -bootstrap-expect.

Oh, no. I see now. Ignore this! The multi-server case never elects a leader, so no anonymous token versus the single server case (pebkac problem where I was still using -bootstrap).

pglass (approved):
Looks good! 

The changelog message is my only blocking comment, but approved so you can get this in prior to code freeze.

agent/acl_endpoint.go
skpratt: updated to "Supplied token does not exist"

agent/consul/acl_endpoint.go
skpratt: done

.changelog/16105.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:breaking-change
+acl errors: Delete and get requests now return descriptive errors when the specified resource cannot be found. Other ACL request errors provide more information about when a resource is missing.
```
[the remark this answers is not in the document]
skpratt: done

agent/acl_endpoint.go
cizh: After discussion around the difference between the two token not found error messages:
1. "supplied token does not exist" from user scenario: User wants to read their own token.
2. "token supplied with request not found" from user scenario: user wants to destroy the token using itself.

We want to **keep both the same as "Supplied token does not exist"** for the following reasons:
1. Consul does not know any other information other than `we can't find the token` and `users are giving this particular request` therefore we can't provide more information for user to access the problem.
2. We cannot provide directions for the users in these two scenarios without making incorrect assumptions and potentially mislead the user.
4. "does not exist" should be sufficient information for the users to understand all the info we can provide.
5. Consistency in error messaging- user should realize both messages are coming from the same cause: Consul did not find this ACL token.

cizh (approved):

PR #16112, opened 2023-01-31
improvement: prevent filter being added twice from any enovy extension

huikang:
### Description
This is to prevent any potential bug from any extension 

(1) may add filter twice, e.g., the Patch method of an extension both return some error and `ok = false`. In this case, filter will be added twice.

(2) doesn't change the filter and will result in err == nil and ok == false. In this case, original filter should be added back to the filter chain

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/envoyextensions/extensioncommon/basic_envoy_extender.go
+++ b/agent/envoyextensions/extensioncommon/basic_envoy_extender.go
@@ -169,11 +169,12 @@ func (b BasicEnvoyExtender) patchTerminatingGatewayListener(config *RuntimeConfi
 			if err != nil {
 				resultErr = multierror.Append(resultErr, fmt.Errorf("error patching listener filter: %w", err))
 				filters = append(filters, filter)
+				continue
 			}
 			if ok {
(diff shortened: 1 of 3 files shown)


agent/envoyextensions/extensioncommon/basic_envoy_extender.go
```diff
@@ -173,6 +173,8 @@ func (b BasicEnvoyExtender) patchTerminatingGatewayListener(config *RuntimeConfi
 			if ok {
 				filters = append(filters, newFilter)
 				patched = true
+			} else if err == nil { // NOTE: this is to prevent filter being added twice
```
hashi-derek, line 176: Would it be safer long-term to just put a `continue` inside the `if err !=nil` block for each one of these loops?

huikang: Fixed and the PR description is updated.

hashi-derek (approved):

PR #16114, opened 2023-01-31
Add Envoy extension metrics.

hashi-derek:
This PR adds metrics for Envoy extensions to track how long it takes for each of three steps to execute:

- validate_arguments (constructing the extension)
- validate (performing runtime validation)
- extend (executing the extension)

Each one of these metrics also includes some metadata as labels to better identify the service affected.

### DIFF
--- a/agent/xds/delta.go
+++ b/agent/xds/delta.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"strconv"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -31,6 +32,7 @@ import (
 	"github.com/hashicorp/consul/agent/xds/extensionruntime"
 	"github.com/hashicorp/consul/agent/xds/xdscommon"
 	"github.com/hashicorp/consul/logging"
+	"github.com/hashicorp/consul/version"
 )
 
 var errOverwhelmed = status.Error(codes.ResourceExhausted, "this server has too many xDS streams open, please try another")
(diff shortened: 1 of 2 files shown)


agent/xds/delta.go
```diff
@@ -409,7 +411,18 @@ func (s *Server) applyEnvoyExtensions(resources *xdscommon.IndexedResources, cfg
 				"partition", cfg.ServiceName.Partition,
 			}
 
+			getMetricLabels := func(err error) []metrics.Label {
```
erichaberkorn, line 414: Do you think it makes sense to include `namespace` and `partition` here too?

erichaberkorn (approved):

agent/xds/delta.go
hashi-derek: When running in k8s, it generates output like this usually, which I figured was sufficient because it includes the namespace. I should add those extra fields just to be safe, though.

```
consul_envoy_extension_extend{app="consul", chart="consul-helm", component="server", controller_revision_hash="consul-server-59b6f98bb4", error="false", extension="builtin/lua", hasDNS="true", instance="10.42.0.13:8500", job="kubernetes-pods", kubernetes_namespace="consul", kubernetes_pod_name="consul-server-0", quantile="0.5", release="consul", service="forward1-sidecar-proxy", statefulset_kubernetes_io_pod_name="consul-server-0", version="builtin/1.15.0"}
```

erichaberkorn (approved):

PR #16115, opened 2023-01-31
Add unit test and update golden files.

hashi-derek:
This adds a test where both Lua and Lambda extensions are loaded at the same time. **Note to PR reviewers -- most of this is just whitespace changes whenever the golden file update was run. Try viewing with split-mode and "Hide whitespace" enabled.**


thisisnotashwin (approved):
heh! you werent kidding about the number of whitespace changes! nicely done

PR #16122, opened 2023-01-31
docs(service-defaults): upstream overrides for peered services

DanStough:
### Description
Docs for ##15956. Upstream overrides will require the `peer` field to work for peered services in 1.16.

### DIFF
--- a/website/content/docs/connect/config-entries/mesh.mdx
+++ b/website/content/docs/connect/config-entries/mesh.mdx
@@ -317,7 +317,7 @@ Note that the Kubernetes example does not include a `partition` field. Configura
           name: 'namespace',
           enterprise: true,
           description:
-            'Must be set to `default`. If running Consul Open Source, the namespace is ignored (see [Kubernetes Namespaces in Consul OSS](/docs/k8s/crds#consul-oss)). If running Consul Enterprise see [Kubernetes Namespaces in Consul Enterprise](/consul/docs/k8s/crds#consul-enterprise) for additional information.',
+            'Must be set to `default`. If running Consul Open Source, the namespace is ignored (see [Kubernetes Namespaces in Consul OSS](/consul/docs/k8s/crds#consul-oss)). If running Consul Enterprise see [Kubernetes Namespaces in Consul Enterprise](/consul/docs/k8s/crds#consul-enterprise) for additional information.',
         },
       ],
       hcl: false,
@@ -326,7 +326,7 @@ Note that the Kubernetes example does not include a `partition` field. Configura
       name: 'TransparentProxy',
       type: 'TransparentProxyConfig: <optional>',
       description:
-        'Controls configuration specific to proxies in `transparent` [mode](/docs/connect/config-entries/service-defaults#mode). Added in v1.10.0.',
+        'Controls configuration specific to proxies in `transparent` [mode](/consul/docs/connect/config-entries/service-defaults#mode). Added in v1.10.0.',
       children: [
         {
           name: 'MeshDestinationsOnly',
--- a/website/content/docs/connect/config-entries/mesh.mdx
+++ b/website/content/docs/connect/config-entries/mesh.mdx
@@ -317,7 +317,7 @@ Note that the Kubernetes example does not include a `partition` field. Configura
           name: 'namespace',
           enterprise: true,
           description:
-            'Must be set to `default`. If running Consul Open Source, the namespace is ignored (see [Kubernetes Namespaces in Consul OSS](/consul/docs/k8s/crds#consul-oss)). If running Consul Enterprise see [Kubernetes Namespaces in Consul Enterprise](/consul/docs/k8s/crds#consul-enterprise) for additional information.',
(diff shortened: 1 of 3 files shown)


website/content/docs/connect/config-entries/mesh.mdx
```diff
@@ -326,7 +326,7 @@ Note that the Kubernetes example does not include a `partition` field. Configura
       name: 'TransparentProxy',
       type: 'TransparentProxyConfig: <optional>',
       description:
-        'Controls configuration specific to proxies in `transparent` [mode](/consul/docs/connect/config-entries/service-defaults#mode). Added in v1.10.0.',
+        'Controls configuration specific to proxies in `transparent` [mode](/docs/connect/config-entries/service-defaults#mode). Added in v1.10.0.',
```
DanStough, line 329: No idea why, but these links are messed up in prod. Though maybe removing the prefix would help.

hashi-derek (approved):
This looks really good. Thanks for taking the time to polish this. I'm approving this, but we might wanna wait for someone from the docs team to also approve before merging.

website/content/docs/connect/config-entries/service-defaults.mdx
```diff
@@ -421,6 +421,16 @@ represents a location outside the Consul cluster. They can be dialed directly wh
...
             },
+            {
+              name: 'Peer',
+              type: 'string: ""',
+              description: 
+                `The name of the peer containing the upstream. This should not be set to the wildcard specifier \`*\`.<br><br>
+                    NOTE: If the \`peer\` field is not set in any of \`Overrides\`, an override applies to peered services with the same [\`name\`](/consul/docs/connect/config-entries/service-defaults#name) 
+                    regardless if this field is set. This maintains backward compatibility with Consul 1.14.0. 
+                    If any override has the \`peer\` field specified, this compatibility mode is disabled and each peered service must have it's own override.
+                    This compatibility mode will be removed in a future version of Consul.`,
```
trujillo-adam, line 432:
```suggestion
                `The name of the peer containing the upstream. Do not use a wildcard specifier ( \`*\`).<br><br>
                    If the \`peer\` field is not set in an \`Overrides\` configuration, then Consul applies overrides to peered services with the same [\`name\`](/consul/docs/connect/config-entries/service-defaults#name),
                    even if the \`peer\` field is set in the service configuration. This maintains backward compatibility with Consul 1.14.0. 
                    If the \`peer\` field is specified in any override, then backward compatibility with Consul 1.14.0 is disabled. As a result, each peered service must have an override configuration.` ,
```
A few suggestions to switch to active voice and to remove temporal language per the style guide. Also, I was a little confused by the "even if this field is set" part of the NOTE statement. It reads like contradictory statements. I tried to add something that made sense, but I don't think it's correct. Do we mean that the peer field must be set for all overrides, otherwise the override configuration is applied to any service name matching the configuration in any peered cluster?

website/content/docs/upgrading/upgrade-specific.mdx
```diff
@@ -87,6 +87,55 @@ In Consul v1.15 and higher:
 
    </CodeBlockConfig>
 
+#### `service-defaults` Overrides for Upstream Peered Services
+
+In Consul 1.14.x, `service-defaults` upstream [`overrides`](/consul/docs/connect/config-entries/service-defaults#overrides) would apply to a both local and peered services as long as the `name` field matched.
```
trujillo-adam, line 92:
```suggestion
In Consul 1.14.x, `service-defaults` upstream [`overrides`](/consul/docs/connect/config-entries/service-defaults#overrides) applies to both local and peered services as long as the `name` field matched.
```

website/content/docs/upgrading/upgrade-specific.mdx
```diff
@@ -87,6 +87,55 @@ In Consul v1.15 and higher:
...
+Consul 1.15.0 introduces a compatibility mode to maintain this behavior as long as [`peer`](/consul/docs/connect/config-entries/service-defaults#peer) is not set in any override.
```
trujillo-adam, line 93:
```suggestion
Consul 1.15.0 is backward compatible with 1.14 if the [`peer`](/consul/docs/connect/config-entries/service-defaults#peer) is not set in any override.
```
Just to make this simpler and remove the suggestion that we're implementing a new mode that the user has to keep track of. I understand this just to be more of a behavior

website/content/docs/upgrading/upgrade-specific.mdx
```diff
@@ -87,6 +87,55 @@ In Consul v1.15 and higher:
...
+However, starting in Consul 1.16.0, peered services will require explicit overrides in `service-defaults` by specifying the [`peer`](/consul/docs/connect/config-entries/service-defaults#peer) field.
```
trujillo-adam, line 94:
```suggestion
```
I recommend removing future-oriented language. I would reinsert this or a similar statement when 1.16 is available.

trujillo-adam (commented):
Added some suggestions for style consistency and a few questions to help improve the clarity.

website/content/docs/connect/config-entries/service-defaults.mdx
DanStough: Actually I think this is pretty close to what I was trying to say, I think we just need to remove `even if the \`peer\` field is set in the service configuration` for clarity.

trujillo-adam (approved):
looks great 👍

Issue #11816, opened 2021-12-13
ServiceRouter NumRetries says it defaults to 0 but the Envoy default is 1


lkysow:
Our doc here: https://www.consul.io/docs/connect/config-entries/service-router#numretries says the NumRetries is 0 by default, but Envoy defaults it to 1:

![image](https://user-images.githubusercontent.com/1034429/145765325-616e57ea-da83-4e97-9a15-14ba69f5e717.png)

(from https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-retrypolicy)

I'm not sure the best way to tackle this since the default of our config entry _is_ 0, but once you have any retry config in place then Envoy's default takes effect.

Issue #8516, opened 2020-08-14
Setting service-router NumRetries to 0 has no effect


lkysow:
#### Overview of the Issue

Setting `num_retries` to 0 results in envoy still retrying.

```
kind = "service-router"
name = "api"
routes = [
  {
    destination = {
      num_retries = 0,
      retry_on_status_codes = [500]
    }
  }
]
```

Results in:
```
          "route": {
           "cluster": "api.default.gke.internal.a8e6e9ad-c56f-3df4-f08d-f923570c4ca3.consul",
           "retry_policy": {
            "retry_on": "retriable-status-codes",
            "retriable_status_codes": [
             500
            ]
           }
          }
```

If `num_retries = 3` you get:
```
"route": {
           "cluster": "api.default.gke.internal.a8e6e9ad-c56f-3df4-f08d-f923570c4ca3.consul",
           "retry_policy": {
            "retry_on": "retriable-status-codes",
            "num_retries": 3,
            "retriable_status_codes": [
             500
            ]
           }
          }
```

I think because we never set `num_retries: 0` then envoy uses its default of 1 (https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-retrypolicy).

If we set `retry_on_status_codes = []` then we get no retries because the retry_policy block is omitted from the envoy config:
```
 "routes": [
         {
          "match": {
           "prefix": "/"
          },
          "route": {
           "cluster": "api.default.gke.internal.a8e6e9ad-c56f-3df4-f08d-f923570c4ca3.consul"
          }
         },
```

### Consul version
`1.8.2`

Mouni919:
Hi @lkysow

I would like to update Envoy to disable its automatic retry on request failure (for every back-end services). Im updating this  in values.yaml file 

```
routesOnly:
  # Launcher Mgmt Streaming status
  - name: launcher-mgmt-ms
    ports: 
      servicePort: 9100
      containerPort: 9100
    rewrite:
      uri: /
    route:
      prefix: /launcher/streaming/
      retry_on_status_codes: []
      timeoutSeconds: 300
     
  # Lesson Mgmt Health
  - name: lesson-mgmt-ms
    ports: 
      servicePort: 8300
      containerPort: 8300
    rewrite:
      uri: /status/
    route:
      prefix: /rest/api/latest/lesson/status/
      retry_on_status_codes: []
      timeoutSeconds: 300
 ```

[...]

PR #16123, opened 2023-01-31
Document how numRetries can't be set to 0

lkysow:
Fixes https://github.com/hashicorp/consul/issues/11816 and fixes https://github.com/hashicorp/consul/issues/8516. See those issues for more details but tl;dr: envoy defaults num_retries to 1 and so if any other retry configs are set, then you'll get 1 retry. If you try setting our NumRetries to 0 we treat it as the zero value (because our struct doesn't have a pointer for that field) and so we leave it unset in the Envoy config which then results in it defaulting to 1.

### DIFF
--- a/website/content/docs/connect/config-entries/service-router.mdx
+++ b/website/content/docs/connect/config-entries/service-router.mdx
@@ -701,8 +701,8 @@ spec:
       name: 'NumRetries',
       type: 'int: 1',
       description: {
-        hcl: 'The number of times to retry the request when a retryable result occurs. This cannot be set to 0 so if you wish to disable retries, unset all other retry settings (`RetryOnConnectFailure`, `RetryOn`, `RetryOnStatusCodes`)',
-        yaml: 'The number of times to retry the request when a retryable result occurs. This cannot be set to 0 so if you wish to disable retries, unset all other retry settings (`retryOnConnectFailure`, `retryOn`, `retryOnStatusCodes`)',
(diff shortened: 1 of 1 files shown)


website/content/docs/connect/config-entries/service-router.mdx
```diff
@@ -699,9 +699,11 @@ spec:
     },
     {
       name: 'NumRetries',
-      type: 'int: 0',
-      description:
-        'The number of times to retry the request when a retryable result occurs.',
+      type: 'int: 1',
+      description: {
+        hcl: 'The number of times to retry the request when a retryable result occurs. This cannot be set to 0 so if you wish to disable retries, unset all other retry settings (`RetryOnConnectFailure`, `RetryOn`, `RetryOnStatusCodes`)',
+        yaml: 'The number of times to retry the request when a retryable result occurs. This cannot be set to 0 so if you wish to disable retries, unset all other retry settings (`retryOnConnectFailure`, `retryOn`, `retryOnStatusCodes`)',
```
trujillo-adam, line 705:
```suggestion
        hcl: 'The number of times to retry the request when a retryable result occurs. You cannot set the value to `0`. To disable retries, unset all other retry settings (`RetryOnConnectFailure`, `RetryOn`, `RetryOnStatusCodes`)',
        yaml: 'The number of times to retry the request when a retryable result occurs. You cannot set the value to `0`. To disable retries, unset all other retry settings (`retryOnConnectFailure`, `retryOn`, `retryOnStatusCodes`)',
```
Switch to active voice

trujillo-adam (approved):
Looks good. I added a minor suggestion but approved the PR so you're not blocked.

PR #16124, opened 2023-01-31
Set `codegen-tools` to be a dependency of `deep-copy`

t-eckert:
### Description

I just ran into this when running `deep-copy` for the first time. While it didn't take me too long to figure out, I did spend a minute figuring out what what wrong when the error I got was originally a bit esoteric (raised from the `./agent/structs/deep-copy.sh` script).

I thought I would save future devs the moments of confusion with a helpful error.

Have `deep-copy` not installed (`rm $(which deep-copy)` works) then run `make deep-copy`.
You will get the error `deep-copy not found. Please run 'make codegen-tools' to install it.`

### PR Checklist

* ~[ ] updated test coverage~
* ~[ ] external facing docs updated~
* ~[ ] not a security concern~

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -366,11 +366,7 @@ codegen-tools:
 	@$(SHELL) $(CURDIR)/build-support/scripts/devtools.sh -codegen
 
 .PHONY: deep-copy
-deep-copy:
-	@if [ -z "$(shell which deep-copy)" ]; then \
-		echo "deep-copy not found. Please run 'make codegen-tools' to install it." ; \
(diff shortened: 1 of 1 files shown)


mkeeler (commented):
This seems fine but alternatively this could just depend on the tools make target which should install the tool.

t-eckert:
> This seems fine but alternatively this could just depend on the tools make target which should install the tool.

That is a much better idea!

boxofrad (approved):
Nice!

mkeeler (approved):

PR #16127, opened 2023-02-01
WIP Snapshot Translation

t-eckert:
- Begin stubbing for SDS
- Start adding tests
- Remove second BoundAPIGateway case in glue
- TO BE PICKED: fix formatting of str
- WIP
- Fix merge conflict
- Implement HTTP Route to Discovery Chain config entries

### Description

### Links

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/agent/proxycfg/snapshot.go
+++ b/agent/proxycfg/snapshot.go
@@ -688,12 +688,12 @@ type configSnapshotAPIGateway struct {
 //
 // FUTURE: Remove when API gateways have custom snapshot generation
 func (c *configSnapshotAPIGateway) ToIngress() configSnapshotIngressGateway {
+	ingressListeners := make(map[IngressListenerKey]structs.IngressListener, len(c.Listeners))
 
 	return configSnapshotIngressGateway{
 		ConfigSnapshotUpstreams: c.ConfigSnapshotUpstreams,
-		// TODO Build from c.Listeners
-		// Listeners:
-		Defaults: structs.IngressServiceConfig{},
+		Listeners:               ingressListeners,
+		Defaults:                structs.IngressServiceConfig{},
 	}
 }
 
--- a/agent/proxycfg/snapshot.go
+++ b/agent/proxycfg/snapshot.go
@@ -3,7 +3,9 @@ package proxycfg
 import (
 	"context"
 	"fmt"
+	"hash/crc32"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/hashicorp/consul/acl"
@@ -644,6 +646,8 @@ type configSnapshotAPIGateway struct {
 
 	TLSConfig structs.GatewayTLSConfig
 
+	APIGatewayConfigEntry *structs.APIGatewayConfigEntry
+
 	// GatewayConfigLoaded is used to determine if we have received the initial
 	// api-gateway config entry yet.
 	GatewayConfigLoaded bool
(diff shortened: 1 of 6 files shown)


agent/proxycfg/snapshot.go
```diff
@@ -691,14 +695,85 @@ type configSnapshotAPIGateway struct {
...
-	return configSnapshotIngressGateway{
-		ConfigSnapshotUpstreams: c.ConfigSnapshotUpstreams,
-		// TODO Build from c.Listeners
-		// Listeners:
-		Defaults: structs.IngressServiceConfig{},
+		ingressListener := structs.IngressListener{
+			Port:     listener.Port,
+			Protocol: string(listener.Protocol),
+			// TODO TLS determine what cluster name and cert resource should be.
+			TLS: nil,
```
t-eckert, line 713: This TLS work is dependent on SDS work that will be completed after this merge. Just set to `nil` for now.

agent/proxycfg/snapshot.go
```diff
@@ -691,14 +695,85 @@ type configSnapshotAPIGateway struct {
...
 // FUTURE: Remove when API gateways have custom snapshot generation
-func (c *configSnapshotAPIGateway) ToIngress() configSnapshotIngressGateway {
+func (c *configSnapshotAPIGateway) ToIngress(datacenter string) (configSnapshotIngressGateway, error) {
+	// Convert API Gateway Listeners to Ingress Listeners.
+	ingressListeners := make(map[IngressListenerKey]structs.IngressListener, len(c.Listeners))
+	synthesizedChains := map[UpstreamID]*structs.CompiledDiscoveryChain{}
+	for name, listener := range c.Listeners {
+		boundListener, ok := c.BoundListeners[name]
+		if !ok {
+			// Skip any listeners that don't have a bound listener. Once the bound listener is created, this will be run again.
```
t-eckert, line 705: @andrewstucki, does this track for you? I was thinking about the case where an API Gateway Listener exists, the reconciler is run, but the bound listener doesn't exist yet. I'm not sure if this case is even possible. But I figure when the bound listener is created, the reconciler will run again and this will be updated.

andrewstucki: So, this could be the case if something happened like we had a stale `BoundAPIGateway` -- i.e. a Gateway gets reconciled, has stuff bound to it, then renames one of its listeners. You're right that it should get re-reconciled fairly quickly, but we could potentially have a small window in which the bound gateway listeners don't match the gateway listeners themselves.

Ideally we'd also do some cursory checks like skipping routes if their protocols don't match the listener as well, just in case someone did something like changed the protocol and the resources haven't had time to reconcile yet.

Once we have a second pass at all of this with the discoverychain to GAMMA style work and potentially a bit more stable structs for our `BoundAPIGateway` data, I'd love to close these sorts of drift issues by just materializing everything onto the `BoundAPIGateway` itself so we don't have to glom multiple config entries together on the xDS side.

andrewstucki: Just to clarify, the reason this shouldn't ever get called when there are no bound listeners is that the `ToIngress` code only ever gets called after we have the filled-in snapshot, which I believe @nathancoleman gated on the presence of both `BoundAPIGateway` and `APIGateway` config entries.

t-eckert: Ok, good. I think that makes sense. Should we wait on those cursory checks until after feature freeze?

andrewstucki: We can handle it in a follow-up PR, but can you `TODO` it?

agent/proxycfg/snapshot_test.go
```diff
@@ -50,3 +53,36 @@ func TestConfigSnapshot_Clone(t *testing.T) {
 		t.FailNow()
 	}
 }
+
+func TestAPIGatewaySnapshotToIngressGatewaySnapshot(t *testing.T) {
```
t-eckert, line 57: This will need to be expanded as we further flesh out TSL work.

agent/proxycfg/snapshot.go
t-eckert: Certainly!

t-eckert: > Ideally we'd also do some cursory checks like skipping routes if their protocols don't match the listener as well, just in case someone did something like changed the protocol and the resources haven't had time to reconcile yet.

I think we might not have to worry about this case because the creation of the chain with the routes will only grab routes that match the protocol.

Is this right, or am I missing something?

```go
	for _, routeRef := range boundListener.Routes {
		switch routeRef.Kind {
		case structs.HTTPRoute:
			route, ok := c.HTTPRoutes.Get(routeRef)
			if !ok {
				continue
			}
			synthesizer.AddHTTPRoute(*route)
			for _, service := range route.GetTargets() {
				id := NewUpstreamIDFromServiceName(structs.NewServiceName(service.Name, &service.EnterpriseMeta))
				if chain := c.DiscoveryChain[id]; chain != nil {
					chains = append(chains, chain)
				}
			}
		case structs.TCPRoute:
			route, ok := c.TCPRoutes.Get(routeRef)
			if !ok {
				continue
			}
			synthesizer.AddTCPRoute(*route)
			for _, service := range route.GetTargets() {
				id := NewUpstreamIDFromServiceName(structs.NewServiceName(service.Name, &service.EnterpriseMeta))
				if chain := c.DiscoveryChain[id]; chain != nil {
					chains = append(chains, chain)
				}
			}
		default:
			continue
		}
	}
```

andrewstucki: So, it's not so much a matter of the routes drifting, but of the fact that the listener protocol may have changed and the given routes are no longer validly bound to the gateway. Basically if we're now an HTTP listener and we generate a bunch of TCP-based discovery chains it'd be no good -- I'm not entirely sure how the ingress gateway xDS generation code handles all of this under the hood and whether it'd error out or cause some odd issues, but my thought is that we should probably throw some `listener.Protocol == route.Protocol` checks as we iterate over the listener, doing a `continue` if they don't match.

agent/consul/discoverychain/gateway.go
```diff
@@ -0,0 +1,190 @@
...
+
+		l.matchesByHostname[host] = matches
+	}
+}
+
+// Synthesize assembles a synthetic discovery chain from multiple other discovery chains
+// that have StartNodes that are referenced by routers or splitters in the entries for the
+// given CompileRequest.
+//
+// This is currently used to help API gateways masquarade as ingress gateways
```
sarahalsmiller, line 86: Just for my own understanding, is this stuff only here to bridge the gap between the prexisting gateway implementation and the new stuff we're making? When does it get called as part of the reconciliation loop.

t-eckert: Yes! It allows us to avoid fully implementing xDS for API Gateway. Instead, we smoosh our configuration into an Ingress Gateway config and use the generated xDS from that.

In the reconciliation loop, there is the kick-off as any of the API Gateway configuration entries sees a change. The API Gateway as a whole is validated. Then a snapshot of its state is made.

That snapshot goes through this function which turns it into a snapshot of an Ingress Gateway. Then this sneaky Ingress Gateway snapshot is used to produce a discovery chain mapping the whole configuration out. This chain is then used to update xDS appropriately.

In the future, we'll get rid of discovery chains and have a different API for modifying xDS. So it didn't make sense to fully do that implementation in this cycle of work.

andrewstucki (approved):
Approving for merge after Nate's PR, assuming tests pass and no conflicts.

agent/consul/discoverychain/gateway_httproute.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	var splitters []*structs.ServiceSplitterConfigEntry
+
+	for idx, rule := range route.Rules {
+		modifier := httpRouteFiltersToServiceRouteHeaderModifier(rule.Filters.Headers)
+		prefixRewrite := httpRouteFiltersToDestinationPrefixRewrite(rule.Filters.URLRewrites)
+
+		var destination structs.ServiceRouteDestination
+		if len(rule.Services) == 1 {
+			// TODO open question: is there a use case where someone might want to set the rewrite to ""?
```
nathancoleman, line 82: Doesn't have to be resolved in this PR, but I believe the answer to this is "yes".

In previous places I've worked, you would do something like target a specific microservice in the mesh using `GET https://domain.com/service/myservice/v1/resources`.

`myservice` would serve the route `/v1/resources` but had no need to know how your request got to that service, so you would match on `/service/myservice` and send it to the `myservice` backend while rewriting the prefix to `""`.

This is just one use case that's probably pretty edge-casey, but it demonstrates why you might do such a thing. IIRC we also need to support this to fulfill the K8s Gateway API.

agent/consul/discoverychain/gateway_tcproute.go
```diff
@@ -0,0 +1,17 @@
+package discoverychain
+
+import "github.com/hashicorp/consul/agent/structs"
+
+func synthesizeTCPRouteDiscoveryChain(route structs.TCPRouteConfigEntry) []structs.IngressService {
+	services := []structs.IngressService{}
```
nathancoleman, line 6: #nit we know what size this needs to be
```suggestion
	services := make([]structs.IngressService, 0, len(route.Services))
```

agent/proxycfg/snapshot.go
```diff
@@ -644,6 +646,8 @@ type configSnapshotAPIGateway struct {
 
 	TLSConfig structs.GatewayTLSConfig
 
+	APIGatewayConfigEntry *structs.APIGatewayConfigEntry
```
nathancoleman, line 649: This is `GatewayConfig` below. Probably added in my branch after you had it here

agent/proxycfg/snapshot.go
```diff
@@ -688,14 +692,99 @@ type configSnapshotAPIGateway struct {
...
+			}
+			synthesizer.AddTCPRoute(*route)
+			for _, service := range route.GetServices() {
+				id := NewUpstreamIDFromServiceName(structs.NewServiceName(service.Name, &service.EnterpriseMeta))
+				if chain := c.DiscoveryChain[id]; chain != nil {
+					chains = append(chains, chain)
+				}
+			}
+		default:
+			continue
```
nathancoleman, line 774: Should we error here? It seems like something has gone horribly wrong if we encounter an unknown route kind at this point

agent/structs/config_entry_routes.go
```diff
@@ -216,6 +217,10 @@ type HTTPHeaderFilter struct {
 	Set    map[string]string
 }
 
+type URLRewrite struct {
+	Path string
```
nathancoleman, line 221: I suspect this will need to become an optional `*string` in the future, but not something that has to be resolved in this PR

nathancoleman (changes requested):
Nice work! Lots of familiar-looking code from consul-api-gateway up in here 🎉

agent/consul/discoverychain/gateway_httproute.go
t-eckert: Sounds good, I'll loop around and handle this as we do cleanup.

agent/proxycfg/snapshot.go
t-eckert: Ah, merges.

agent/structs/config_entry_routes.go
t-eckert: I will do a wraparound

nathancoleman (approved):

PR #16131, opened 2023-02-02
Discovery Chain Merging/Synthesizing for HTTP/TCPRoutes

andrewstucki:
Moves most of the route translation to the discoverychain package and adds the logic for merging in other discovery chains to the one we synthesize for a listener's routes.


t-eckert (approved):
This is awesome, let's merge it in and I'll write tests on it

PR #16141, opened 2023-02-02
gateways: add Gateway and Route config entry status condition types and helpers

mikemorris:
### Links
- https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#condition-v1-meta
- https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/shared_types.go

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/consul/fsm_data_store.go
+++ b/agent/consul/fsm_data_store.go
@@ -52,18 +52,30 @@ func (f *FSMDataStore) Update(entry structs.ConfigEntry) error {
 
 // UpdateStatus takes a config entry, an error, and updates the status field as needed in the FSM state
 func (f *FSMDataStore) UpdateStatus(entry structs.ControlledConfigEntry, err error) error {
-	if err == nil {
-		//TODO additional status messages for success?
-		return nil
-	}
-	status := structs.Status{
-		Conditions: []structs.Condition{{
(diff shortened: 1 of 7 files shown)


agent/consul/fsm_data_store.go
```diff
@@ -0,0 +1,77 @@
+package consul
```
sarahalsmiller, line 1: You'll probably want to rebase here? I don't think you meant to touch these files.

proto/pbconfigentry/config_entry.proto
```diff
@@ -602,6 +602,7 @@ message Status {
 // name=Structs
 message Condition {
   string Type = 1;
+  // mog: func-from=conditionStatusFromStructs func-to=conditionStatusToStructs
```
mikemorris, line 605: Why does this not seem to be working, do I need to regenerate something?

mikemorris: Needed to reinstall Xcode developer tools to get `make tools` and then `make proto` to run successfully.

mikemorris:
Superceded by https://github.com/hashicorp/consul/pull/16994

PR #16142, opened 2023-02-02
fix goroutine leak in renew testing

eikenb:
Test overwrote the stopWatcher() function variable for the test without keeping and calling the original value. The original value is the function that stops the goroutine... so it needs to be called.


agent/connect/ca/provider_vault_test.go
```diff
@@ -306,11 +305,12 @@ func TestVaultCAProvider_RenewTokenStopWatcherOnConfigure(t *testing.T) {
 		"IntermediatePKIPath": "pki-intermediate/",
 	})
 
-	var gotStopped = uint32(0)
+	gotStopped := uint32(0)
```
lkysow, line 308: Can you add a comment here explaining this dance?

lkysow (approved):

PR #16143, opened 2023-02-02
remove redundant vault api retry logic

### LINKED ISSUE #12613: Remove unnecessary custom retry logic associated with Consul's usage of the Vault API
#### Feature Description

Once we've upgraded to a later version of protobuf we can also upgrade to a newer version of vault/api. This will pull in improvements but the biggest known win is that the vault LifetimeWatcher will automatically backoff on Vault errors and allow us to remove out our own backoff logic. The backoff logic was added in https://github.com/hashicorp/consul/pull/12607 to prevent thousands of requests to Vault if the authentication method becomes invalid.

The protobuf changes are completed: https://github.com/hashicorp/consul/issues/10471

eikenb:
We upgraded Vault API module version to a version that has built-in retry logic. So this code is no longer necessary.

Fixes #12613

### DIFF
--- a/agent/connect/ca/provider_vault.go
+++ b/agent/connect/ca/provider_vault.go
@@ -21,7 +21,6 @@ import (
 	"github.com/hashicorp/consul/agent/structs"
 	"github.com/hashicorp/consul/lib"
 	"github.com/hashicorp/consul/lib/decode"
-	"github.com/hashicorp/consul/lib/retry"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


agent/connect/ca/provider_vault.go
```diff
@@ -225,41 +224,25 @@ func (v *VaultProvider) renewToken(ctx context.Context, watcher *vaultapi.Lifeti
...
-			retrier.Wait(ctx)
-
 			// If the watcher has exited and auth method is enabled,
 			// re-authenticate using the auth method and set up a new watcher.
 			if v.config.AuthMethod != nil {
 				// Login to Vault using the auth method.
 				loginResp, err := vaultLogin(v.client, v.config.AuthMethod)
 				if err != nil {
 					v.logger.Error("Error login in to Vault with %q auth method", v.config.AuthMethod.Type)
-					// Restart the watcher
```
skpratt, line 262: most inconsequential nit: these comment still help draw attention to the watcher.Start() lines below

skpratt (approved):

agent/connect/ca/provider_vault.go
eikenb: I've always thought of comments as useful to say *why* something is done. That if  *what* or *how* isn't clear it would indicate a code refactor is needed. That is, IMO, that it reads cleaner/clearer this way as comment doesn't break the flow of the code.

eikenb:
Thanks for the review @skpratt! :heart:

PR #16145, opened 2023-02-03
WIP: Add routes controllers

andrewstucki:
### Description
This adds the logic for TCP and HTTP route controllers, and depends on a couple of changes I made in the current snapshot branch. I can rebase and make the changes standalone if we want, but otherwise this can probably just be merged after the snapshot --> ingress code is merged.

Going to flesh out the testing of the controllers and then we should be good to go. Would love to get some eyes on it early though.

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


agent/consul/gateways/bind_test.go
```diff
@@ -891,7 +891,7 @@ func TestBindRoutesToGateways(t *testing.T) {
 
 	for name, tc := range cases {
 		t.Run(name, func(t *testing.T) {
-			actualBoundAPIGateways, referenceErrors := BindRoutesToGateways(tc.gateways, tc.routes...)
+			actualBoundAPIGateways, _, referenceErrors := BindRoutesToGateways(tc.gateways, tc.routes...)
```
t-eckert, line 894: Should I circle back and update tests to look for the bound references here?

agent/consul/gateways/bind.go
```diff
@@ -21,14 +22,20 @@ type gatewayRefs = map[configentry.KindName][]structs.ResourceReference
 // The function returns a list of references to the modified BoundAPIGatewayConfigEntry objects,
 // a map of resource references to errors that occurred when they were attempted to be
```
t-eckert, line 23: Let's update the doc comment too

Issue #16147, opened 2023-02-03
API Accepts Invalid Type for "ConnectTimeout" in service-resolver Configuration


reskin89:
#### Overview of the Issue

When applying a service resolver configuration, if ConnectTimeout is provided as an `int` instead of a `string` with a time suffix, the api will accept and apply the configuration, however this results in a failure to then speak to that service as the resolver config is invalid, and you are met with connection resets attempting to speak to it via connect.

#### Reproduction Steps

Steps to reproduce this issue, eg:

1. Create or have a cluster ready
1. Register a service
1. Create a service-resolver configuration for that service, using an `int` for the time value in `ConnectTimeout`
2. attempt to communicate with that service via connect.

### Consul info for both Client and Server

<details>
  <summary>Client info</summary>

```
agent:
	check_monitors = 0
	check_ttls = 0
	checks = 3
	services = 2
build:
	prerelease =
	revision = bd257019
	version = 1.14.3
	version_metadata =
consul:
	acl = enabled
	known_servers = 3
	server = false
runtime:
	arch = amd64
	cpu_count = 2
	goroutines = 93
	max_procs = 2
	os = linux
	version = go1.19.4
serf_lan:
	coordinate_resets = 0
	encrypted = true
	event_queue = 0
	event_time = 111
	failed = 0
	health_score = 0
	intent_queue = 0
	left = 0
	member_time = 1158115
	members = 121
	query_queue = 0
	query_time = 3323
```

</details>

<details>
  <summary>Server info</summary>

```
agent:
	check_monitors = 0
	check_ttls = 0
	checks = 0
	services = 0
build:
	prerelease =
	revision = 12dd8aa9
	version = 1.12.8
	version_metadata =
consul:
	acl = enabled
	bootstrap = false
	known_datacenters = 3
	leader = true
	leader_addr = REDACTED:8300
	server = true
raft:
	applied_index = 54345301
	commit_index = 54345301
	fsm_pending = 0
	last_contact = 0
	last_log_index = 54345301
	last_log_term = 641
	last_snapshot_index = 54330691
	last_snapshot_term = 641
	latest_configuration = [{REDACTED}]
	latest_configuration_index = 0
	num_peers = 2
	protocol_version = 3
	protocol_version_max = 3
	protocol_version_min = 0
	snapshot_version_max = 1
	snapshot_version_min = 0
	state = Leader
	term = 641
runtime:
	arch = amd64
	cpu_count = 2
	goroutines = 3917
[...]

PR #16149, opened 2023-02-03
docs(service-resolver): clarify the default time unit in service-resolver.ConnectTimeout

huikang:
### Description
When timeunit is not provided in `service-resolver.ConnectTimeout`, the default to be used in nanosecond.

* Use a config entry without time unit

```
{
  "Kind":"service-resolver",
  "Name":"REDACTED",
  "ConnectTimeout":10, 
  "Failover":{
    "*":{
      "Datacenters":["us-east-1","us-west-2"]
      }
  }
}
```

```bash
consul config write resolver.json


consul config read -kind service-resolver -name REDACTED  

{
    "ConnectTimeout": "10ns",
    "Kind": "service-resolver",
    "Name": "REDACTED",
    "Failover": {
        "*": {
            "Datacenters": [
                "us-east-1",
                "us-west-2"
            ]
        }
    },
    "CreateIndex": 23,
    "ModifyIndex": 137
}


```

### Links
https://github.com/hashicorp/consul/issues/16147

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/website/content/docs/connect/config-entries/service-resolver.mdx
+++ b/website/content/docs/connect/config-entries/service-resolver.mdx
@@ -438,7 +438,7 @@ spec:
       name: 'ConnectTimeout',
       type: 'duration: 0s',
       description:
-        'The timeout for establishing new network connections to this service.',
+        'The timeout for establishing new network connections to this service. If time unit isn't provided, the default unit is nanosecond',
     },
     {
       name: 'DefaultSubset',
--- a/website/content/docs/connect/config-entries/service-resolver.mdx
+++ b/website/content/docs/connect/config-entries/service-resolver.mdx
@@ -438,7 +438,7 @@ spec:
       name: 'ConnectTimeout',
       type: 'duration: 0s',
       description:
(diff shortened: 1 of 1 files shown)


website/content/docs/connect/config-entries/service-resolver.mdx
```diff
@@ -438,7 +438,7 @@ spec:
       name: 'ConnectTimeout',
       type: 'duration: 0s',
       description:
-        'The timeout for establishing new network connections to this service.',
+        'The timeout for establishing new network connections to this service. If time unit isn't provided, the default unit is nanosecond',
```
trujillo-adam, line 441:
```suggestion
        'The timeout for establishing new network connections to this service. The default unit of time is `ns`.',
```
Per the style guide, we want to avoid using contractions. And labeling the behavior as "default" makes the statement about not providing a unit unnecessary. Also, it might be better to use the actual unit value instead of writing out nanosecond just because it's closer to the actual usage, but I defer to you on that point.

trujillo-adam (approved):
I made a suggestion, but approved the PR so you're not blocked.

PR #16163, opened 2023-02-06
fix proxy config of service-defaults doesn't override the defaults

huikang:
### Description
Fix proxy config provided in `service-defaults doesn't override the values if presented in service definition`.

As a user wants to change some proxy config value like `MaxInBoundConnection` or `localRequestTimeoutMs` of all service instances, I want to change the behavior of writing the value to config-entry `service-defaults`. However, the current implementation doesn't allow service-defaults override the defaults if the field name presents.

1. Register a service with following values in its `sidecar.proxy.config`: `max_inbound_connections = 1024`, `local_idle_timeout_ms `.

2. Submit a `service-defaults` with values to override the defaults, like
```
MaxInboundConnections = 4096 // value different than in the service definition.
```

The observed value will be from the service definitions.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/configentry/merge_service_config.go
+++ b/agent/configentry/merge_service_config.go
@@ -109,7 +109,7 @@ func MergeServiceConfig(defaults *structs.ServiceConfigResponse, service *struct
 	// Merge proxy defaults
 	ns := nsRaw.(*structs.NodeService)
 
-	if err := mergo.Merge(&ns.Proxy.Config, defaults.ProxyConfig); err != nil {
+	if err := mergo.Merge(&ns.Proxy.Config, defaults.ProxyConfig, mergo.WithOverride); err != nil {
 		return nil, err
 	}
 	if err := mergo.Merge(&ns.Proxy.Expose, defaults.Expose); err != nil {
--- a/agent/configentry/merge_service_config_test.go
+++ b/agent/configentry/merge_service_config_test.go
@@ -758,3 +758,122 @@ func Test_MergeServiceConfig_UpstreamOverrides(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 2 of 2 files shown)


agent/configentry/merge_service_config.go
```diff
@@ -109,7 +109,7 @@ func MergeServiceConfig(defaults *structs.ServiceConfigResponse, service *struct
 	// Merge proxy defaults
 	ns := nsRaw.(*structs.NodeService)
 
-	if err := mergo.Merge(&ns.Proxy.Config, defaults.ProxyConfig); err != nil {
+	if err := mergo.Merge(&ns.Proxy.Config, defaults.ProxyConfig, mergo.WithOverride); err != nil {
```
hashi-derek, line 112: I'm under the impression that local service configuration is supposed to take precedence over the `service-defaults`, which would allow operators to set unique config on individual instances, if needed.

huikang: I see; that makes sense. However, does service-defaults should precede the values from proxy-defaults? 
Values from `proxy-default` overrides `service-defaults` is also what I observed in a test.

hashi-derek: Typically, service-defaults should win over proxy-defaults. Are you sure that was the situation? We have lots of unit tests ensuring that service-defaults takes precedence over proxy-defaults, and I've done quite a bit of testing with it recently. If you have a specific example where it occurs, we can definitely look into it.

huikang: I will reopen if I found any issue with the test example. Thanks.

PR #16165, opened 2023-02-06
Net 2229/rpc reduce max retries 2

dhiaayachi:
### Description
This is a remake of #16146 

> The retries were too aggressive which compounded rate limiting errors. Hence, implementing an exponential back-off strategy which gives enough time for the server to recover.

### DIFF
--- a/lib/cluster.go
+++ b/lib/cluster.go
@@ -45,6 +45,11 @@ func RandomStagger(intv time.Duration) time.Duration {
 	return time.Duration(uint64(rand.Int63()) % uint64(intv))
 }
 
+// RandomStagger returns an interval between min and the max duration
+func RandomStaggerWithRange(min time.Duration, max time.Duration) time.Duration {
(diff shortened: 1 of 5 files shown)


JadhavPoonam (commented):
Thank you for taking this forward @dhiaayachi 🙌 
If the bench tests look good then we can 🚀

lib/cluster.go
```diff
@@ -45,6 +45,11 @@ func RandomStagger(intv time.Duration) time.Duration {
 	return time.Duration(uint64(rand.Int63()) % uint64(intv))
 }
 
+// RandomStagger returns an interval between min and the max duration
+func RandomStaggerWithRange(min time.Duration, max time.Duration) time.Duration {
```
JadhavPoonam, line 49: ~ Missing test for this

JadhavPoonam (approved):

PR #16176, opened 2023-02-06
Adding experimental support for a more efficient LogStore implementation

banks:
### Description
This PR adds support for an experimental (and optional) alternative storage backend for raft logs that is more efficient. It's been reviewed already internally.

The aim is to include this as an opt-in option in Consul 1.15. Documentation of the feature and migration/testing steps will be added later after this is merged. We hope to solicit feedback from users who have tried it in controlled environments and eventually production before we switch it to be the default.

### PR Checklist

* [x] updated test coverage
* [ ] external facing docs updated (this is TODO post code freeze)
* [x] not a security concern


agent/config/testdata/TestRuntimeConfig_Sanitize.golden
```diff
@@ -16,9 +16,9 @@
     "ACLTokens": {
         "ACLAgentRecoveryToken": "hidden",
         "ACLAgentToken": "hidden",
+        "ACLConfigFileRegistrationToken": "hidden",
```
banks, line 19: Not sure why this is showing up in this PR to be honest. But the test passes so it must need to be there now. Does that mean it was merged in another PR that never passed config tests because this was missing or is there some strange magic going on here?

pglass: Hey, I was poking through PRs for an entirely unrelated reason, and saw this comment. When I added this new token type, it looks like I manually added this field to the test file, but added it in non-sorted key order (it was present on [line 21 prior](https://github.com/hashicorp/consul/pull/16176/files#diff-7639a07e925eafb265d9856ac83fca5947f4782c0864acd1aa85a698bca48324L21)), and the tests still passed because I guess they ignore sort order.

Sorry for the confusion 🙂

banks: np. thanks for clarifying Paul! Yeah we match with json equivalence not byte comparison so that explains it! I missed the deletion too.

jmurret (approved):
🎉 LGTM to me.  I verified this by auditing each file side by side with the prior PR.  Everything checked out.  :shipit:

PR #16188, opened 2023-02-07
Add additional controller implementations

andrewstucki:
### Description
This supersedes https://github.com/hashicorp/consul/pull/16145 and fixes up some of the binding/cleanup logic from the initial gateway controller implementation (for example, we need to cleanup any referenced route statuses corresponding to a gateway when a gateway is removed).

Additionally, it adds a bunch of dirty-checking logic so we don't wind up infinitely triggering reconciliation for a given resource.

One major design change is that I actually collapsed the controllers into a single monolithic controller with different entrypoints depending on which resource they are reconciling. This has the benefit of making it a bit easier to control for race-conditions between reconciliation paths since we're working with a single go-routine that is handling reconciliation for all of our resources and deriving state monolithically. We avoided the race conditions in the k8s implementation by implementing a global store mutex and handling most of the reconciliation logic with a lock held on the mutex as you can see [here](https://github.com/hashicorp/consul-api-gateway/blob/632e8fb6718e5084a4cc77f213a4ba599cdbf919/internal/store/store.go#L74-L75). Merging the controllers instead makes it so we don't have to introduce some sort of global lock and is a pattern found fairly often in the internals of the kubernetes code-base (for example in their GC controller).

I'll see if we can schedule some time to go over the logic here and I'll try and add some more comments to the code while y'all are reviewing. Thanks!

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/consul/controller/controller.go
```diff
@@ -347,6 +349,13 @@ func (c *controller) enqueueEntry(entry structs.ConfigEntry, transformers ...Tra
 	}
 }
 
+// Enqueue adds all of the given entry into the work queue.
+func (c *controller) Enqueue(requests ...Request) {
```
andrewstucki, line 353: This allows us to be super flexible with the way we handle triggering reconciliation on dependent resources. It's used in the inline-certificate reconciliation code path, allowing us to retrieve all gateways that reference the given cert and then enqueue them for reconciliation.

agent/consul/gateways/bind.go
```diff
@@ -43,8 +51,13 @@ func BindRoutesToGateways(gateways []*gatewayMeta, routes ...structs.BoundRoute)
...
 				for _, ref := range references {
 					delete(parentRefs, ref)
+					// this ref successfully bound, add it to the set that we'll update the
+					// status for
+					if _, found := errored[ref]; !found {
+						boundRefs = append(boundRefs, references...)
+					}
 				}
-			} else if gateway.unbindRoute(route) {
+			} else if gateway.unbindRoute(routeRef) {
```
andrewstucki, line 60: Switched to using refs directly for binding/unbinding so that we can unbind even when we no longer have a full object because of something like a route being deleted that we need to remove from our bound gateway.

agent/consul/gateways/bind.go
```diff
@@ -71,3 +84,52 @@ func getReferences(route structs.BoundRoute) (referenceSet, gatewayRefs) {
...
+		Name: req.Name,
+	}
+	if req.Meta != nil {
+		ref.EnterpriseMeta = *req.Meta
+	}
+	return ref
+}
+
+// RemoveGateway sets the route status for the given gateway to be unbound if it should be bound
+func RemoveGateway(gateway structs.ResourceReference, entries ...structs.BoundRoute) []structs.ControlledConfigEntry {
```
andrewstucki, line 100: This is triggered when we're cleaning up resources after a gateway has been deleted. We need to be able to set any route that references a gateway to now say, "I'm not bound to this gateway anymore since it's been deleted".

agent/consul/gateways/bind.go
```diff
@@ -71,3 +84,52 @@ func getReferences(route structs.BoundRoute) (referenceSet, gatewayRefs) {
...
+		}
+		if toUpdate, shouldUpdate := updater.UpdateEntry(); shouldUpdate {
+			modified = append(modified, toUpdate)
+		}
+	}
+	return modified
+}
+
+// RemoveRoute unbinds the route from the given gateways, returning the list of gateways that were modified.
+func RemoveRoute(route structs.ResourceReference, entries ...*gatewayMeta) []*gatewayMeta {
```
andrewstucki, line 125: This is triggered when a route gets deleted by using the route itself as a resource ref -- see the comment above on the switch of `unbindRoute` to use refs instead of `structs.BoundRoute`.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,201 +2,563 @@ package gateways
...
+			return r.reconcileRoute(ctx, req, store, route)
+		}, r.cleanupRoute)
+	case structs.TCPRoute:
+		return reconcileEntry(r.fsm.State(), ctx, req, func(ctx context.Context, req controller.Request, store *state.Store, route *structs.TCPRouteConfigEntry) error {
+			return r.reconcileRoute(ctx, req, store, route)
+		}, r.cleanupRoute)
+	case structs.InlineCertificate:
+		return r.enqueueCertificateReferencedGateways(r.fsm.State(), ctx, req)
+	default:
+		return nil
```
andrewstucki, line 54: This is the aforementioned controller collapsing code.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,201 +2,563 @@ package gateways
...
-			Topic:   state.EventTopicAPIGateway,
-			Subject: stream.SubjectWildcard,
-		},
-	)
 }
 
-// Reconcile takes in a controller request and ensures this api gateways corresponding BoundAPIGateway exists and is
-// up to date
-func (r *apiGatewayReconciler) Reconcile(ctx context.Context, req controller.Request) error {
+func reconcileEntry[T structs.ControlledConfigEntry](store *state.Store, ctx context.Context, req controller.Request, reconciler func(ctx context.Context, req controller.Request, store *state.Store, entry T) error, cleaner func(ctx context.Context, req controller.Request, store *state.Store) error) error {
```
andrewstucki, line 58: This does the initial fetch operation for a config entry and then hands off to either a cleanup function or a reconciliation function in the case of this being an update or create operation.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,201 +2,563 @@ package gateways
...
 		}
 	}
-
 	return nil
 }
 
-func (r *apiGatewayReconciler) retrieveAllRoutesFromStore() ([]structs.BoundRoute, error) {
-	tcpRoutes, err := r.store.GetConfigEntriesByKind(structs.TCPRoute)
+func (r apiGatewayReconciler) reconcileBoundGateway(ctx context.Context, req controller.Request, store *state.Store, bound *structs.BoundAPIGatewayConfigEntry) error {
+	// this reconciler handles orphaned bound gateways at startup, it just checks to make sure there's still an existing gateway, and if not, it deletes the bound gateway
```
andrewstucki, line 123: As this comment mentions, the main use-case for adding this as a reconciliation loop is if we have orphaned resources due to a leader election switch or controller crash after a gateway has been deleted. If a gateway has been deleted but we don't detect it due to a crash/leader switch, then we'd leave around bound gateway resources and never mark the routes as now unbound from a gateway.

Due to the startup characteristics of the controller we do a list + watch operation when subscribing to a given resource type, so when the controller starts back up this will detect any `BoundAPIGatewayConfigEntry` objects that don't have corresponding APIGateways and garbage collect them.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,201 +2,563 @@ package gateways
...
-	}
+	resource := requestToResourceRef(req)
+	resource.Kind = structs.APIGateway
 
-	// and update the route statuses
-	for route, routeError := range routeErrors {
-		configEntry := r.resourceReferenceToBoundRoute(route)
-		r.logger.Error("route binding error:", routeError)
-		if err := r.store.UpdateStatus(configEntry, routeError); err != nil {
+	for _, toUpdate := range RemoveGateway(resource, routes...) {
```
andrewstucki, line 114: We handle the cleanup of route statuses in the cleanup of `BoundAPIGatewayConfigEntry` objects so that this gets triggered either on a normal detection of a gateway delete or when we GC an orphaned bound object.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,201 +2,563 @@ package gateways
...
+			Message:            "listener has no route conflicts",
+			LastTransitionTime: &now,
+		})
 	}
 
-	//initialize object, values get copied over in ensureBoundGateway if they don't exist
-	metaGateway.BoundGateway = boundGateway.(*structs.BoundAPIGatewayConfigEntry)
-	return metaGateway, nil
+	// now check if we need to update the gateway status
+	if toUpdate, shouldUpdate := updater.UpdateEntry(); shouldUpdate {
```
andrewstucki, line 284: We wrap all of the status update operations with `StatusUpdater`s so that we can detect whether or not a status update is actually needed to avoid triggering infinite loops in our reconciliation code since an update of a resource's status will trigger a write operation that our controller subscription will detect.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,201 +2,563 @@ package gateways
...
+		r.logger.Debug("persisting gateway status", "gateway", gateway)
+		if err := r.updater.UpdateWithStatus(toUpdate); err != nil {
+			r.logger.Error("error persisting gateway status", "error", err)
+			return err
+		}
+	}
+
+	// next update route statuses
+	for _, toUpdate := range updatedRoutes {
```
andrewstucki, line 293: We only ever return routes that are newly bound/unbound from the gateway, so this should effectively act like the dirty checking mentioned above.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,201 +2,563 @@ package gateways
...
+		r.logger.Debug("persisting route status", "route", toUpdate)
+		if err := r.updater.UpdateWithStatus(toUpdate); err != nil {
+			r.logger.Error("error persisting route status", "error", err)
+			return err
+		}
+	}
+
+	// now update the bound state if it changed
+	if bound == nil || stateIsDirty(bound.(*structs.BoundAPIGatewayConfigEntry), meta.BoundGateway) {
```
andrewstucki, line 302: Unlike the other config entries which have status values, we need to check to see if anything has actually changed in the BoundAPIGatewayConfigEntry state. This check isn't strictly necessary since we don't re-write `BoundAPIGatewayConfigEntry` objects in their handler, but it avoids any unneeded extra raft + reconciliation operation.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,201 +2,563 @@ package gateways
...
+	}
+
+	r.logger.Debug("got route reconcile call", "request", req)
+
+	updater := structs.NewStatusUpdater(route)
+	// we clear out the initial status conditions since we're doing a full update
+	// of this route's status
+	updater.ClearConditions()
+
+	ws := memdb.NewWatchSet()
```
andrewstucki, line 345: This watchset acts as a trigger for us to re-reconcile any time some upstream service has a change in its discovery chain so that we can revalidate the route (i.e. check that it still has the proper protocol alignment).

agent/structs/config_entry_status.go
```diff
@@ -31,11 +45,66 @@ type Status struct {
 	Conditions []Condition
 }
 
+func (s Status) SameConditions(other Status) bool {
```
andrewstucki, line 48: This is pretty much only used in our testing code for equality checking the conditions that we're expecting to be set on an object and the actual conditions set, so I may move it there.

andrewstucki (commented):
Self review/comments.

agent/consul/gateways/bind.go
```diff
@@ -21,14 +23,20 @@ type gatewayRefs = map[configentry.KindName][]structs.ResourceReference
 // The function returns a list of references to the modified BoundAPIGatewayConfigEntry objects,
 // a map of resource references to errors that occurred when they were attempted to be
 // bound to a gateway.
```
t-eckert, line 25: Let's also be sure to update this comment.

andrewstucki: 👍 prior to merge I'll be going over each of the docstrings here and adjusting a bunch of comments.

t-eckert: Great, just wanted to check

agent/consul/gateways/bind.go
```diff
@@ -71,3 +84,52 @@ func getReferences(route structs.BoundRoute) (referenceSet, gatewayRefs) {
...
+	}
+	if req.Meta != nil {
+		ref.EnterpriseMeta = *req.Meta
+	}
+	return ref
+}
+
+// RemoveGateway sets the route status for the given gateway to be unbound if it should be bound
+func RemoveGateway(gateway structs.ResourceReference, entries ...structs.BoundRoute) []structs.ControlledConfigEntry {
+	now := time.Now()
```
t-eckert, line 101: Do we want this to be `time.Now().UTC()`? `Now()` will return local time. So long as all servers are in the same region that would be fine. I just wonder if there is a situation where a comparison is done between servers in different regions and something doesn't work.

agent/consul/gateways/gateway_meta.go
```diff
@@ -17,6 +21,36 @@ type gatewayMeta struct {
...
+		for _, g := range gateways {
+			gateway := g.(*structs.APIGatewayConfigEntry)
+			if bound.References(gateway) {
+				meta = append(meta, &gatewayMeta{
+					BoundGateway: bound,
+					Gateway:      gateway,
+				})
+				break
+			}
+		}
```
t-eckert, line 49: Might cause more headaches than it's worth, but I think we could pop the gateway at the index after it is matched to make successive checks faster.

agent/consul/gateways/gateway_meta.go
```diff
@@ -137,3 +181,118 @@ func (g *gatewayMeta) boundListenerByName(name string) (int, *structs.BoundAPIGa
...
+			} else {
+				bound.Certificates = append(bound.Certificates, ref)
+			}
+		}
+	}
+	return certificateErrors, nil
+}
+
+func (g *gatewayMeta) checkConflicts() (structs.ControlledConfigEntry, bool) {
+	now := time.Now()
```
t-eckert, line 205: Similar with the UTC question above. I think if we always parse back into a Go code `Time` object, it should be fine. It's a bit of an edge case. 🤷🏻‍♂️

andrewstucki: I'll swap all of these to use UTC -- which was a good catch. This is only used for users to see when a status condition was set though, so it shouldn't be problematic for any other code.

agent/consul/gateways/gateway_meta.go
andrewstucki: So, the problem is that the BoundAPIGateway and APIGateway config entries might not align entirely 1-1 or in the same order, for example, maybe a gateway has been deleted, but its bound gateway is still around, or maybe it hasn't been reconciled, or maybe we ended up reconciling gateways in the opposite order that they were created due to a transient reconciliation failure. Basically, this code is trying to handle any number of gateways in any order that may or may not have corresponding bound gateways, so I don't think we can just use the same index.

andrewstucki: Ah, nvm, I got what you're saying, reslice gateways so we don't have to iterate over the whole length of the array. I'd be alright with that, but I think it's probably more trouble than it's worth as an initial optimization.

agent/consul/gateways/bind.go
```diff
@@ -71,3 +84,52 @@ func getReferences(route structs.BoundRoute) (referenceSet, gatewayRefs) {
...
+	for _, route := range entries {
+		updater := structs.NewStatusUpdater(route)
+		for _, parent := range route.GetParents() {
+			if parent.Kind == gateway.Kind && parent.Name == gateway.Name && parent.EnterpriseMeta.IsSame(&gateway.EnterpriseMeta) {
+				updater.SetCondition(structs.Condition{
+					Type:               "Bound",
+					Status:             "False",
+					Reason:             "GatewayNotFound",
+					Message:            "gateway was not found",
+					Resource:           &parent,
```
jm96441n, line 112: will this run into the issue of capturing a pointer to the loop variable `parent` and not the actual value that is there?

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,201 +2,563 @@ package gateways
...
+	// shared state. In our current out-of-repo code, this is handled via a global
+	// lock on our shared store, but this makes it so we don't have to deal with lock
+	// contention, and instead just work with a single control loop.
+	switch req.Kind {
+	case structs.APIGateway:
+		return reconcileEntry(r.fsm.State(), ctx, req, r.reconcileGateway, r.cleanupGateway)
+	case structs.BoundAPIGateway:
+		return reconcileEntry(r.fsm.State(), ctx, req, r.reconcileBoundGateway, r.cleanupBoundGateway)
+	case structs.HTTPRoute:
+		return reconcileEntry(r.fsm.State(), ctx, req, func(ctx context.Context, req controller.Request, store *state.Store, route *structs.HTTPRouteConfigEntry) error {
```
jm96441n, line 44: can we define this as a named function rather than an inline anonymous function to help make this a bit more readable?

agent/consul/gateways/bind.go
andrewstucki: 👍 good catch -- I'll check for other captured range vars as well.

agent/consul/gateways/controller_gateways.go
jm96441n, line 58: oooooo some generics in the wild

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,201 +2,563 @@ package gateways
...
 	}
+	r.controller.Enqueue(requests...)
+	return nil
+}
 
-	r.ensureBoundGateway(metaGateway)
+func (r apiGatewayReconciler) cleanupBoundGateway(_ context.Context, req controller.Request, store *state.Store) error {
+	logger := r.logger.With("bound-gateway", req.Name, "partition", req.Meta.PartitionOrDefault(), "namespace", req.Meta.NamespaceOrDefault())
+	logger.Debug("cleaning up bound gateway")
+	defer logger.Debug("finished cleaning up bound gateway")
```
jm96441n, line 104: do we always want to log this? could this be misleading in the case where we bail early and return an error?

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,201 +2,563 @@ package gateways
...
+		if entry, updated := routeUpdater.UpdateEntry(); updated {
+			updatedRoutes = append(updatedRoutes, entry)
+		}
+	}
+
+	// first check for gateway conflicts
+	for i, listener := range meta.BoundGateway.Listeners {
+		protocol := meta.Gateway.Listeners[i].Protocol
+		switch protocol {
+		case structs.ListenerProtocolTCP:
```
jm96441n, line 250: should we use a regular if statement here since the switch statement only has a single case or do we plan on adding more cases in the future?

andrewstucki: My thought is that we'll add more in the future, we currently don't have support for `UDPRoute` or `GRPCRoute` as defined in the upstream spec.

agent/consul/gateways/controller_gateways.go
andrewstucki: So, I'm thinking this is good to have for timing more than anything, mainly since the loggers inject timestamps at each case. I think what I'd probably prefer to do is have this, but just ensure that each error code path actually logs the error as well.

agent/consul/gateways/bind.go
```diff
@@ -65,9 +78,58 @@ func getReferences(route structs.BoundRoute) (referenceSet, gatewayRefs) {
...
+		Kind: req.Kind,
+		Name: req.Name,
+	}
+	if req.Meta != nil {
+		ref.EnterpriseMeta = *req.Meta
+	}
+	return ref
+}
+
+// RemoveGateway sets the route status for the given gateway to be unbound if it should be bound
```
nathancoleman, line 99:
```suggestion
// RemoveGateway sets the route's status appropriately when the gateway that it's
// attempting to bind to does not exist
```

agent/consul/gateways/bind.go
```diff
@@ -65,9 +78,58 @@ func getReferences(route structs.BoundRoute) (referenceSet, gatewayRefs) {
...
+	for _, route := range entries {
+		updater := structs.NewStatusUpdater(route)
+		for _, parent := range route.GetParents() {
+			if parent.Kind == gateway.Kind && parent.Name == gateway.Name && parent.EnterpriseMeta.IsSame(&gateway.EnterpriseMeta) {
+				updater.SetCondition(structs.Condition{
+					Type:               "Bound",
+					Status:             "False",
+					Reason:             "GatewayNotFound",
+					Message:            "gateway was not found",
+					Resource:           pointerTo(parent),
```
nathancoleman, line 112: Is this returning a `ResourceReference` or a literal pointer? If the former, the function name could be improved

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,201 +2,571 @@ package gateways
...
-		if err := r.store.Delete(&structs.BoundAPIGatewayConfigEntry{
-			Kind:           structs.BoundAPIGateway,
-			Name:           req.Name,
-			EnterpriseMeta: *req.Meta,
-		}); err != nil {
-			msg := "error cleaning up deleted gateway object"
-			r.logger.Error(msg, err)
-			return errors.Wrap(err, msg)
+	}
+	requests := []controller.Request{}
```
nathancoleman, line 74: #nit for readability

```suggestion
	}

	requests := []controller.Request{}
```

agent/consul/gateways/gateway_meta.go
t-eckert: I agree. We can come back to it if this becomes a pain point, but it's not likely.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,201 +2,571 @@ package gateways
...
+		return err
+	}
+	if entry == nil {
+		return cleaner(ctx, req, store)
+	}
+	return reconciler(ctx, req, store, entry.(T))
+}
 
-	r.logger.Debug("started reconciling gateway", "gateway", req.Name)
+func (r apiGatewayReconciler) enqueueCertificateReferencedGateways(store *state.Store, _ context.Context, req controller.Request) error {
```
nathancoleman, line 65:
```suggestion
// enqueueCertificateReferencedGateways retrieves all gateway objects, filters to those referencing
// the provided certificate, and enqueues the gateways for reconciliation
func (r apiGatewayReconciler) enqueueCertificateReferencedGateways(store *state.Store, _ context.Context, req controller.Request) error {
```

nathancoleman (changes requested):
This is my review so far - going to post a few small reviews as I go today rather than one giant one at the end of the day

agent/consul/gateways/bind.go
andrewstucki: It's a generic function, it returns a literal pointer to whatever you passed in so that you can do things like chain calls ala `pointerTo(time.Now().UTC())`, in this case it has the added benefit of acting like shadowing the `parent` loop variable so that we don't have to put something like `parent := parent` as the first line of this loop.

agent/consul/gateways/controller_gateways.go
andrewstucki: Going to resolve this @jm96441n since I cleaned up the logging and added error messages wherever we return an error from the reconciliation loop. Let me know if there are any other logging changes you want met to make.

agent/structs/config_entry_routes.go
```diff
@@ -54,8 +73,7 @@ func (e *HTTPRouteConfigEntry) GetParents() []ResourceReference {
 	if e == nil {
 		return []ResourceReference{}
 	}
-	// TODO HTTP Route should have "parents". Andrew will implement this in his work.
-	return []ResourceReference{}
+	return e.Parents
```
t-eckert, line 76: Does it matter that `Parents` can be `nil` here? Should we protect against that above?

andrewstucki: It shouldn't matter since it's a slice and we're not doing any sort of JSON marshaling/unmarshaling where a `nil` slice v. zero-valued slice actually makes a difference.

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,201 +2,677 @@ package gateways
...
+}
+
+func NewAPIGatewayController(fsm *fsm.FSM, publisher state.EventPublisher, updater *Updater, logger hclog.Logger) controller.Controller {
+	reconciler := apiGatewayReconciler{
+		fsm:     fsm,
+		logger:  logger,
+		updater: updater,
+	}
+	reconciler.controller = controller.New(publisher, reconciler)
+	return reconciler.controller.Subscribe(
```
t-eckert, line 595: This is pretty

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,201 +2,677 @@ package gateways
...
 	}
 
-	r.ensureBoundGateway(metaGateway)
+	r.controller.Enqueue(requests...)
 
-	routes, err := r.retrieveAllRoutesFromStore()
+	return nil
+}
+
+func (r apiGatewayReconciler) cleanupBoundGateway(_ context.Context, req controller.Request, store *state.Store) error {
```
nathancoleman, line 108:
```suggestion

// cleanupBoundGateway retrieves all routes from the store and removes the gateway from any
// routes that are bound to it, updating their status appropriately
func (r apiGatewayReconciler) cleanupBoundGateway(_ context.Context, req controller.Request, store *state.Store) error {
```

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,201 +2,677 @@ package gateways
...
-	// and update the route statuses
-	for route, routeError := range routeErrors {
-		configEntry := r.resourceReferenceToBoundRoute(route)
-		r.logger.Error("route binding error:", routeError)
-		if err := r.store.UpdateStatus(configEntry, routeError); err != nil {
+	return nil
+}
+
+func (r apiGatewayReconciler) reconcileBoundGateway(_ context.Context, req controller.Request, store *state.Store, bound *structs.BoundAPIGatewayConfigEntry) error {
+	// this reconciler handles orphaned bound gateways at startup, it just checks to make sure there's still an existing gateway, and if not, it deletes the bound gateway
```
nathancoleman, line 136:
```suggestion

// reconcileBoundGateway deletes any bound-api-gateway for which a corresponding api-gateway
// config entry does not exist
func (r apiGatewayReconciler) reconcileBoundGateway(_ context.Context, req controller.Request, store *state.Store, bound *structs.BoundAPIGatewayConfigEntry) error {
```

agent/consul/gateways/controller_gateways.go
```diff
@@ -2,201 +2,677 @@ package gateways
...
+
+		// if we've updated any statuses, then store them as needing
+		// to be updated
+		if entry, updated := routeUpdater.UpdateEntry(); updated {
+			updatedRoutes = append(updatedRoutes, entry)
+		}
+	}
+
+	// first check for gateway conflicts
+	for i, listener := range meta.BoundGateway.Listeners {
```
nathancoleman, line 293: Can we use `meta.checkConflicts()` here?

nathancoleman (changes requested):
Lots of small change requests, but nothing systemic that gives me heartburn. I think it'd be good to take a cleanup pass before release, adding godocs, cleaning up some naming, etc.

agent/consul/gateways/controller_gateways.go
andrewstucki: Unfortunately not in its current form because of the different scope of the status updater. I'm sure we can adjust things slightly in a refactor though. WDYT about me adding a TODO to clean this up in a refactor subsequently?

agent/structs/config_entry_status.go
```diff
@@ -23,6 +24,19 @@ type ResourceReference struct {
...
+func (r *ResourceReference) IsSame(other *ResourceReference) bool {
+	if r == nil && other == nil {
+		return true
+	}
+	if r == nil || other == nil {
+		return false
+	}
+	return r.Kind == other.Kind &&
+		r.Name == other.Name &&
+		r.SectionName == other.SectionName &&
```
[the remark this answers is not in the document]
andrewstucki: It shouldn't need to. This code is only called when filtering  conditions to determine whether or not a condition is for the same `Type` + `Resource` and should be overwritten when we're setting a condition.

nathancoleman:
TODO is fine w/ me

agent/consul/gateways/controller_gateways.go
nathancoleman: TODO is fine w/ me

nathancoleman (approved):

PR #16194, opened 2023-02-07
Implement APIGateway proxycfg snapshot

nathancoleman:
### Description
As a prerequisite for generating xDS for API Gateway, we need to subscribe to an `api-gateway` and all related resources. As updates come in, we need to build the `ConfigSnapshot` for the gateway.

### Links

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


agent/proxycfg/state.go
```diff
@@ -32,6 +32,8 @@ const (
 	serviceResolversWatchID            = "service-resolvers"
 	gatewayServicesWatchID             = "gateway-services"
 	gatewayConfigWatchID               = "gateway-config"
+	inlineCertificateConfigWatchID     = "inline-certificate-config"
+	routeConfigWatchID                 = "route-config"
```
nathancoleman, line 36: These prefixes are used when subscribing to config entry changes and then passed along to the change handler. In the change handler, this prefix indicates what sort of change happened so that we can respond to it

agent/structs/config_entry_routes.go
```diff
@@ -226,6 +226,10 @@ type HTTPService struct {
 	acl.EnterpriseMeta
 }
 
+func (s HTTPService) ServiceName() ServiceName {
```
nathancoleman, line 229: This is purely for convenience since we were constructing these `ServiceName`s all over the place

agent/structs/deep-copy.sh
```diff
@@ -21,10 +21,12 @@ deep-copy \
   -type GatewayService \
   -type GatewayServiceTLSConfig \
   -type HTTPHeaderModifiers \
+  -type HTTPRouteConfigEntry \
```
nathancoleman, line 24: Same thing as the generator code above - including the types here means they get `DeepCopy()` functions generated for them.

The generated code can be seen below in `agent/structs/structs.deepcopy.go`

agent/xds/delta.go
```diff
@@ -476,9 +483,10 @@ var xDSUpdateOrder = []xDSUpdateOperation{
...
 	// 5. (NOT IMPLEMENTED YET IN CONSUL) VHDS updates (if any) related to the newly added RouteConfigurations must arrive after RDS updates.
 	// {},
-	// 6. Stale CDS clusters and related EDS endpoints (ones no longer being referenced) can then be removed.
+	// 6. Stale CDS clusters, related EDS endpoints (ones no longer being referenced) and SDS secrets can then be removed.
 	{TypeUrl: xdscommon.ClusterType, Remove: true},
 	{TypeUrl: xdscommon.EndpointType, Remove: true},
+	{TypeUrl: xdscommon.SecretType, Remove: true},
 	// xDS updates can be pushed independently if no new
 	// clusters/routes/listeners are added or if it’s acceptable to
 	// temporarily drop traffic during updates. Note that in case of
```
nathancoleman, line 492: One thing that R.B. and I discussed is how important the order of operations is here. We believe this is the right order at the time but will find out for sure as I work through the SDS code in a future PR.

agent/xds/secrets.go
```diff
@@ -0,0 +1,38 @@
...
+	case structs.ServiceKindAPIGateway:
+		return s.secretsFromSnapshotAPIGateway(cfgSnap)
+	default:
+		return nil, fmt.Errorf("Invalid service kind: %v", cfgSnap.Kind)
+	}
+}
+
+func (s *ResourceGenerator) secretsFromSnapshotAPIGateway(cfgSnap *proxycfg.ConfigSnapshot) ([]proto.Message, error) {
+	var res []proto.Message
+	// TODO
```
nathancoleman, line 36: This is the big question mark right now. This function will be implemented in a followup PR

nathancoleman (commented):
Personal review

agent/proxycfg/api_gateway.go
```diff
@@ -0,0 +1,502 @@
...
+				seenRefs[ref] = struct{}{}
+
+				ctx, cancel := context.WithCancel(ctx)
+				switch ref.Kind {
+				case structs.HTTPRoute:
+					snap.APIGateway.HTTPRoutes.InitWatch(ref, cancel)
+				case structs.TCPRoute:
+					snap.APIGateway.TCPRoutes.InitWatch(ref, cancel)
+				default:
+					return fmt.Errorf("unexpected route kind on gateway: %s", ref.Kind)
```
andrewstucki, line 173: think you'll want to call `cancel` before this to not leak contexts

agent/proxycfg/api_gateway.go
```diff
@@ -0,0 +1,492 @@
...
+	snap.APIGateway.HTTPRoutes = watch.NewMap[structs.ResourceReference, *structs.HTTPRouteConfigEntry]()
+	snap.APIGateway.TCPRoutes = watch.NewMap[structs.ResourceReference, *structs.TCPRouteConfigEntry]()
+	snap.APIGateway.Certificates = watch.NewMap[structs.ResourceReference, *structs.InlineCertificateConfigEntry]()
+
+	snap.APIGateway.WatchedDiscoveryChains = make(map[UpstreamID]context.CancelFunc)
+	snap.APIGateway.DiscoveryChain = make(map[UpstreamID]*structs.CompiledDiscoveryChain)
+	snap.APIGateway.WatchedUpstreams = make(map[UpstreamID]map[string]context.CancelFunc)
+	snap.APIGateway.WatchedUpstreamEndpoints = make(map[UpstreamID]map[string]structs.CheckServiceNodes)
+	snap.APIGateway.WatchedGateways = make(map[UpstreamID]map[string]context.CancelFunc)
+	snap.APIGateway.WatchedGatewayEndpoints = make(map[UpstreamID]map[string]structs.CheckServiceNodes)
```
andrewstucki, line 68: Can you separate these from the rest, they appear to be handled via the delegated `handleUpdateUpstreams` call, and hence need to be initialized, but I'm not entirely sure what all of them are used for. Just so we don't drop them when they need to be set can you also add a comment that these must be set for the upstream handling?

agent/proxycfg/api_gateway.go
```diff
@@ -0,0 +1,492 @@
...
+	snap.APIGateway.Listeners = make(map[string]structs.APIGatewayListener)
+	snap.APIGateway.BoundListeners = make(map[string]structs.BoundAPIGatewayListener)
+	snap.APIGateway.UpstreamPeerTrustBundles = watch.NewMap[string, *pbpeering.PeeringTrustBundle]()
+	snap.APIGateway.PeerUpstreamEndpoints = watch.NewMap[UpstreamID, structs.CheckServiceNodes]()
+	snap.APIGateway.PeerUpstreamEndpointsUseHostnames = make(map[UpstreamID]struct{})
```
andrewstucki, line 73: I believe these are also part of the upstream handling?

agent/proxycfg/snapshot.go
```diff
@@ -639,6 +639,68 @@ func (c *configSnapshotMeshGateway) isEmptyPeering() bool {
...
+	GatewayConfig *structs.APIGatewayConfigEntry
+
+	// BoundGatewayConfigLoaded is used to determine if we have received the initial
+	// bound-api-gateway config entry yet.
+	BoundGatewayConfigLoaded bool
+	BoundGatewayConfig *structs.BoundAPIGatewayConfigEntry
+
+	// Hosts is the list of extra host entries to add to our leaf cert's DNS SANs
+	Hosts    []string
+	HostsSet bool
```
t-eckert, line 659: Maybe `AreHostsSet`? I read this and sometimes think this is a `set` of Hosts.

agent/proxycfg/snapshot.go
```diff
@@ -639,6 +639,68 @@ func (c *configSnapshotMeshGateway) isEmptyPeering() bool {
...
+
+	// Upstreams is a list of upstreams this ingress gateway should serve traffic
+	// to. This is constructed from the ingress-gateway config entry, and uses
+	// the GatewayServices RPC to retrieve them.
+	// TODO Determine if this is updated "for free" or not. If not, we might need
+	//   to do some work to populate it in handlerAPIGateway
+	Upstreams map[IngressListenerKey]structs.Upstreams
+
+	// UpstreamsSet is the unique set of UpstreamID the gateway routes to.
+	UpstreamsSet map[UpstreamID]struct{}
```
t-eckert, line 673: Yeah, given that you have `xSet` meaning both a group and a check that something has happened, I would rename the other.

t-eckert: You can replace `struct{}` with `any` here.

agent/proxycfg/api_gateway.go
```diff
@@ -0,0 +1,492 @@
...
+					snap.APIGateway.TCPRoutes.InitWatch(ref, cancel)
+				default:
+					cancel()
+					return fmt.Errorf("unexpected route kind on gateway: %s", ref.Kind)
+				}
+
+				err := h.subscribeToConfigEntry(ctx, ref.Kind, ref.Name, routeConfigWatchID)
+				if err != nil {
+					// TODO May want to continue
+					return err
```
andrewstucki, line 172: Just wondering for familiarity sake, what happens when we error here? Do we just retry the event handler execution?

agent/proxycfg/snapshot.go
```diff
@@ -639,6 +639,68 @@ func (c *configSnapshotMeshGateway) isEmptyPeering() bool {
...
+	TCPRoutes    watch.Map[structs.ResourceReference, *structs.TCPRouteConfigEntry]
+	Certificates watch.Map[structs.ResourceReference, *structs.InlineCertificateConfigEntry]
+
+	// Listeners is the original listener config from the api-gateway config
+	// entry to save us trying to pass fields through Upstreams
+	Listeners map[string]structs.APIGatewayListener
+
+	BoundListeners map[string]structs.BoundAPIGatewayListener
+
+	// Defaults is the default configuration for upstream service instances
```
t-eckert, line 685: Maybe add a TODO that we aren't sure what this should be yet.

agent/proxycfg/api_gateway.go
```diff
@@ -0,0 +1,492 @@
...
+
+	seenUpstreamIDs := make(map[UpstreamID]struct{})
+	upstreams := make(map[APIGatewayListenerKey]structs.Upstreams)
+
+	switch route := resp.Entry.(type) {
+	case *structs.HTTPRouteConfigEntry:
+		snap.APIGateway.HTTPRoutes.Set(ref, route)
+
+		for _, rule := range route.Rules {
+			for _, service := range rule.Services {
```
andrewstucki, line 280: One my PR merges, think you can replace these nested loops with `for _, service := range route.GetTargetedServices() {`

agent/proxycfg/snapshot.go
nathancoleman: Will update this TODO. We know what it should be but haven't exposed it on our config entry model yet

agent/proxycfg/api_gateway.go
```diff
@@ -0,0 +1,492 @@
...
+				for _, listener := range snap.APIGateway.Listeners {
+					shouldBind := false
+					for _, parent := range route.Parents {
+						if h.referenceIsForListener(parent, listener, snap) {
```
andrewstucki, line 284: Can you add a TODO somewhere in here about extracting this into a shared helper in the structs package? I think we have the same logic that's in this function implemented 2-3 ways across the controller and xDS code.

agent/proxycfg/api_gateway.go
```diff
@@ -0,0 +1,492 @@
...
+					if !shouldBind {
+						continue
+					}
+
+					upstream := structs.Upstream{
+						DestinationName:      service.Name,
+						DestinationNamespace: service.NamespaceOrDefault(),
+						DestinationPartition: service.PartitionOrDefault(),
+						LocalBindPort:        listener.Port,
+						//IngressHosts:         g.Hosts,
```
andrewstucki, line 298: This is used in the xDS generation code, no?

agent/proxycfg/api_gateway.go
```diff
@@ -0,0 +1,492 @@
...
+		return fmt.Errorf("invalid type for config entry: %T", resp.Entry)
+	}
+
+	snap.APIGateway.Upstreams = upstreams
+	snap.APIGateway.UpstreamsSet = seenUpstreamIDs
+	//snap.APIGateway.Hosts = TODO
+	snap.APIGateway.HostsSet = true
+
+	// Stop watching any upstreams and discovery chains that have become irrelevant
+	for upstreamID, cancelDiscoChain := range snap.APIGateway.WatchedDiscoveryChains {
```
andrewstucki, line 388: does this code not happen in the upstream handler?

agent/proxycfg/api_gateway.go
```diff
@@ -0,0 +1,492 @@
...
+	if ref.Name != snap.APIGateway.GatewayConfig.Name {
+		return false
+	}
+	return ref.SectionName == "" || ref.SectionName == listener.Name
+}
+
+func (h *handlerAPIGateway) watchIngressLeafCert(ctx context.Context, snap *ConfigSnapshot) error {
+	// Note that we DON'T test for TLS.enabled because we need a leaf cert for the
+	// gateway even without TLS to use as a client cert.
+	if !snap.IngressGateway.GatewayConfigLoaded || !snap.IngressGateway.HostsSet {
```
andrewstucki, line 429: I'm assuming this should be `APIGateway` v. `IngressGateway`? Also, seeing as serving over TLS requires specifying an inline certificate at the moment, I think we should probably drop this entire handler -- it seems to be for pulling the terminating TLS cert signed by the connect CA.

agent/proxycfg/api_gateway.go
```diff
@@ -0,0 +1,492 @@
...
+	if err != nil {
+		cancel()
+		return err
+	}
+	snap.IngressGateway.LeafCertWatchCancel = cancel
+
+	return nil
+}
+
+func (h *handlerAPIGateway) generateIngressDNSSANs(snap *ConfigSnapshot) []string {
```
andrewstucki, line 454: same as above, I think we can probably just drop this entire function.

agent/structs/config_entry_routes.go
t-eckert: I would add a comment for future generations about how this is different from `Name`.

andrewstucki (commented):
Initial pass

agent/proxycfg/api_gateway.go
nathancoleman: Yes it is, leaving as a TODO (will add TODO since I apparently forgot)

agent/structs/structs.go
```diff
@@ -1217,6 +1218,11 @@ const (
 	// centralized configuration.
 	ServiceKindIngressGateway ServiceKind = "ingress-gateway"
 
+	// ServiceKindAPIGateway is an API Gateway for the Connect feature.
```
t-eckert, line 1221: Should we be using "Connect" anymore? Just say "service mesh"?

nathancoleman: This is me following the existing patterns, but I wouldn't mind updating the other comments nearby in this PR

t-eckert (approved):

agent/proxycfg/snapshot.go
sarahalsmiller: +1, I prefer bools in the form of a question, but not a blocking thing at all.

agent/proxycfg/api_gateway.go
nathancoleman: Looks like we just log and wait for the next watch trigger [here](https://github.com/hashicorp/consul/blob/ba2f44ea0de39551c37728f995832c0b4a2ba2e2/agent/proxycfg/state.go#L344-L349)

sarahalsmiller (approved):
Looks really good, looking forward to getting this stuff all stitched together.

agent/proxycfg/api_gateway.go
nathancoleman: Added to the godoc for `referenceIsForListener`

agent/proxycfg/api_gateway.go
nathancoleman: Not today. You can see that we have to do the same thing in `handlerIngressGateway` [here](https://github.com/hashicorp/consul/blob/d3aa2bd9c563dfbdf98610ff577bb753d4dba506/agent/proxycfg/ingress_gateway.go#L160-L177). I just changed the variable names a bit to make it easier to understand what's happening

agent/proxycfg/snapshot.go
nathancoleman: > You can replace `struct{}` with `any` here.

The deep-copy codegen tool doesn't handle this correctly, so sticking with `struct{}` for now

agent/proxycfg/snapshot.go
nathancoleman: done

PR #16197, opened 2023-02-07
get upstream IPs

malizz:
### Description
Get Upstream IPs and clusters for troubleshooting tool 

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/troubleshoot/proxy/upstreams.go
+++ b/troubleshoot/proxy/upstreams.go
@@ -1,0 +1,120 @@
+package troubleshoot
+
+import (
+	envoy_admin_v3 "github.com/envoyproxy/go-control-plane/envoy/admin/v3"
+	envoy_listener_v3 "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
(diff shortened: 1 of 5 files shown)


troubleshoot/proxy/upstreams.go
```diff
@@ -0,0 +1,120 @@
+package troubleshoot
```
ndhanushkodi, line 1: Should these upstreams files be in troubleshoot/upstreams? I figured the structure would be something like 

```
troubleshoot
  proxy
    *files related to the troubleshoot proxy command*
  upstreams
    *files related to the troubleshoot upstreams command* 
```

ndhanushkodi (approved):
Looks great Mali, just had a question about the dir/package structure!

troubleshoot/proxy/upstreams.go
malizz: We use the same receiver in those files, separating them needs a refactor, will create a separate PR for that.

curtbushko (approved):
LGTM! Thanks for putting this together!

PR #16201, opened 2023-02-08
Add test jobs for arm64 in CircleCI

odidev:
Added test jobs for arm64 in CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,6 +46,13 @@ references:
     yarn: &YARN_CACHE_KEY consul-ui-v9-{{ checksum "ui/yarn.lock" }}
 
 steps:
+  install-go-arm64: &install-go-arm64
+    name: install go
+    command: |
+      sudo rm -rf /usr/local/go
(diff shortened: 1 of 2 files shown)


odidev:
Please review the PR and let me know if any changes required.

jmurret:
Hi @odidev, thank you for your contribution!  We are currently right in the process of migrating off of CircleCI and onto GitHub actions, so we cannot include this change at this time.  We will need to stabilize on GitHub actions and also assess our compute spend for CI and jobs before we add in lots of more processing.  That is not to say we won't include such changes, but CI costs can increase dramatically without intention and we need to stabilize on GitHub actions and assess the cost of the existing amd64 tests across all pushes and calculate the cost of adding all of these arm64 tests as a part of increasing the CI footprint.  I imagine this will take 2-3 months before we are in a position to propose these changes for GitHub actions, but they are valuable and appreciated!  We can reach out then to see if you are interested in contributing before we make such changes.  Thank you again!

PR #16210, opened 2023-02-08
troubleshoot: handle tproxy dialed directly case

ndhanushkodi:
The troubleshoot proxy command validates Envoy config by checking for listeners, routes, clusters, and endpoints. However, if using Tproxy with DialedDirectly, there would be a passthrough cluster for that upstream without any endpoints. We shouldn't error in that case, and should log that it's a passthrough cluster.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/troubleshoot/validate/validate.go
+++ b/troubleshoot/validate/validate.go
@@ -2,6 +2,7 @@ package validate
 
 import (
 	"fmt"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


curtbushko (approved):
This looks familiar!  LGTM!

PR #16211, opened 2023-02-08
update troubleshoot CLI, update flags and upstreams output

malizz:
### Description
Update troubleshoot CLI:
- flag names
- required flags 
- upstreams output

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/command/troubleshoot/upstreams/troubleshoot_upstreams.go
+++ b/command/troubleshoot/upstreams/troubleshoot_upstreams.go
@@ -24,17 +24,17 @@ type cmd struct {
 	help  string
 
 	// flags
-	adminBind string
+	envoyAdminEndpoint string
 }
 
 func (c *cmd) init() {
 	c.flags = flag.NewFlagSet("", flag.ContinueOnError)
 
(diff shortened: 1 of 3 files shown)


command/troubleshoot/upstreams/troubleshoot_upstreams.go
```diff
@@ -74,14 +75,21 @@ func (c *cmd) Run(args []string) int {
 		return 1
 	}
 
+	c.UI.Output("==== Upstreams (explicit upstreams only)")
```
curtbushko, line 78: I think with this output, to make sure it matches consul-k8s should be:

`==> Upstreams (explicit upstreams only) (count)` 

The same goes with Upstream IPs, but the count in the header.

curtbushko (approved):
I have one request/suggestion below but I won't block on the merge.

command/troubleshoot/upstreams/troubleshoot_upstreams.go
malizz: Sounds good,  will change it (I was following the format in RFC).

PR #16214, opened 2023-02-09
Clean-up Gateway Controller Binding Logic

andrewstucki:
### Description
This is a somewhat lightweight refactor to accomplish a few goals:

1. Cleaning up the logic from https://github.com/hashicorp/consul/pull/16213 needed to fix what was causing infinitely rescheduled reconciliations
2. To collapse all of the various files into a single gateway controller file so we can move this to a different shared subpackage
3. To clean up and add some docstrings

If we want to sign off on this directly, it incorporates the fixes found in #16213.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


t-eckert (approved):
This is great! I like the approach with the `setStatus` function and passing the function to process each listener. Very clean!

PR #16215, opened 2023-02-09
troubleshoot basic envoy stats for an upstream

malizz:
### Description
troubleshoot basic envoy stats for an upstream

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/troubleshoot/proxy/troubleshoot_proxy.go
+++ b/troubleshoot/proxy/troubleshoot_proxy.go
@@ -85,10 +85,11 @@ func (t *Troubleshoot) RunAllTests(upstreamEnvoyID, upstreamIP string) (validate
 	}
 
 	// getStats usage example
-	// rejectionStats, err := t.getEnvoyStats("update_rejected")
-	// if err != nil {
-	// 	resultErr = multierror.Append(resultErr, err)
-	// }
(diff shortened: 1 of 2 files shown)


troubleshoot/proxy/troubleshoot_proxy.go
```diff
@@ -85,10 +85,11 @@ func (t *Troubleshoot) RunAllTests(upstreamEnvoyID, upstreamIP string) (validate
 	}
 
 	// getStats usage example
-	// rejectionStats, err := t.getEnvoyStats("update_rejected")
-	// if err != nil {
-	// 	resultErr = multierror.Append(resultErr, err)
-	// }
+	messages, err = t.troubleshootStats(upstreamEnvoyID)
```
curtbushko, line 88: Would this only work when an EnvoyID is passed in? What about with vips?

curtbushko (changes requested):

troubleshoot/proxy/troubleshoot_proxy.go
malizz: I had a conversation with Nitya, and similar to before, we do not have a reliable way of constructing the SNI from envoyID, so I removed it.

curtbushko (approved):

PR #16217, opened 2023-02-09
Add basic smoke test to make sure an APIGateway runs

andrewstucki:
### Description
This adds a smoke test to make sure that an API gateway can actually run and register itself in Consul. As we merge the ingress translation code we can add more behavioral tests.

It found a couple of things:

1. We needed to add a conditional so that an api gateway instance doesn't need to specify the ports it binds (since it specifies them in its config entry)
2. We missed a few places to convert to/from structs/protobufs

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


sarahalsmiller (approved):
Looks good as a leaping off point!

PR #16218, opened 2023-02-09
Immediately (mostly) useable ACL tokens: add wait and check call to allow for ACL token replication

skpratt:
### Description
This is a borrowed workaround to allow time for ACL tokens to go through raft so that they can be immediately used after a login / creation request. See code comments for full description.

Original implementation in consul-k8s: https://github.com/hashicorp/consul-k8s/blob/dc7f08965c01d2180813d0d83539a49bcc60a7d3/control-plane/subcommand/common/common.go#L156
Once this PR is merged, a follow up PR will remove this wait from consul-k8s, since this will be moving the retry loop upstream into the consul client API itself. This should address all calls to the underlying token API, removing the need for dependent projects like consul-k8s to determine which calls should be handled with a retry loop.

Tested with removed consul k8s retry loop.

No k8s retry loop + these changes: 
No k8s retry loop + without these changes:
No k8s retry loop + forced leader change + changes to delay raft write:
VM create at follower + immediate access request  + delay raft write + without these changes:
VM create at follower + immediate access request  + delay raft write + with these changes:

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/agent/acl_endpoint.go
+++ b/agent/acl_endpoint.go
@@ -4,11 +4,20 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"time"
+
+	"github.com/cenkalti/backoff/v3"
+	"golang.org/x/net/context"
 
 	"github.com/hashicorp/consul/acl"
 	"github.com/hashicorp/consul/agent/structs"
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/consul/lib"
+)
+
+const (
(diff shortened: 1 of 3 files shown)


jkirschner-hashicorp:
@skpratt : What user-facing operations does this change apply to? Does it ensure that the token returned by `consul acl token create` or `consul login` is (usually) immediately usable? Or something else? I can't tell for sure because it seems like the change is in HTTP API endpoint files (?) rather than command files.

PR #16221, opened 2023-02-09
API Gateway Envoy Golden Listener Tests

t-eckert:
### Description

This PR adds tests that the correct Envoy configuration is produced for Envoy LDS by generating `.golden` files to test against.

* Run the `make envoy-regen` to regenerate Envoy golden files. 

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/proxycfg/snapshot.go
```diff
@@ -831,6 +831,10 @@ func (c *configSnapshotAPIGateway) ToIngress(datacenter string) (configSnapshotI
 }
 
 func (c *configSnapshotAPIGateway) synthesizeChains(datacenter string, protocol structs.APIGatewayListenerProtocol, port int, name string, boundListener structs.BoundAPIGatewayListener) ([]structs.IngressService, structs.Upstreams, []*structs.CompiledDiscoveryChain, error) {
+	if len(boundListener.Routes) == 0 {
```
t-eckert, line 834: Just moved this to the top of the function for clarity in circuit breaking.

agent/proxycfg/api_gateway.go
```diff
@@ -461,3 +462,49 @@ func (h *handlerAPIGateway) watchIngressLeafCert(ctx context.Context, snap *Conf
 
 	return nil
 }
+
+func (h *handlerAPIGateway) generateIngressDNSSANs(snap *ConfigSnapshot) []string {
```
andrewstucki, line 466: So, this is only used for generating additional SANs on the leaf certificate used for a TLS-based listener. Since we provision TLS independently of the leaf cert and only use the cert with its spiffe id for encrypting/verifying mesh traffic, we shouldn't need any of these SANs, so you should be able to drop this entire function and its call above.

agent/xds/listeners_test.go
```diff
@@ -527,6 +527,211 @@ func TestListenersFromSnapshot(t *testing.T) {
...
+									{Name: "http-service"},
+								},
+							},
+						},
+					},
+				}, nil)
+			},
+		},
+		{
+			name: "api-gateway-tcp-listener-with-tcp-and-http-route",
```
andrewstucki, line 665: These are good for this PR, but I'd like to see about adding some tests for things like multiple HTTP routes and/or services with the same matching criteria so that we can have something that verifies the envoy router gets generated with our proper route prioritization. Do you mind taking a stab at that in a follow-up?

agent/proxycfg/snapshot.go
andrewstucki: So, we still need to check the length of `chains` because of the conditional addition to the array. Imagine you have route refs, but you haven't triggered any of the route watches yet,  you'll wind up with an empty chain that you pass into the `Synthesize` invocation at the bottom. While it's not that big of a deal, it basically means you may be spamming logs for a bit with errors about being unable to compile empty discovery chains.

I think it may be easier just to keep the `chains` check at the bottom here.

andrewstucki (changes requested):
Definitely think you should drop the DNSSans helper, also WDYT about a follow-up PR with a bit more "unhappy" path tests?

agent/proxycfg/api_gateway.go
t-eckert: Cool, I actually wasn't 100% sure where this came in from. I saw you added it and didn't want to nuke something accidentally.

agent/proxycfg/snapshot.go
t-eckert: That's a great point! Thanks for catching that.

agent/xds/listeners_test.go
t-eckert: Certainly, I'll ping you tomorrow and we can make a list of which cases would be relevant.

andrewstucki (approved):

PR #16223, opened 2023-02-09
add integration tests for troubleshoot

malizz:
### Description
- add integration tests for troubleshoot command. (happy paths)
- remove adminAddr from interface: only envoy proxies use this function 

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


test/integration/consul-container/libs/service/service.go
```diff
@@ -1,14 +1,18 @@
...
+	"io"
+)
 
 // Service represents a process that will be registered with the
 // Consul catalog, including Consul components such as sidecars and gateways
 type Service interface {
 	// Export a service to the peering cluster
+	Exec(ctx context.Context, cmd []string) (int, io.Reader, error)
 	Export(partition, peer string, client *api.Client) error
 	GetAddr() (string, int)
```
DanStough, line 15: Love the refactoring here 🙇

DanStough (changes requested):
No problems with the PR content, but looking at the linting errors in CircleCI, some other integration tests need to `GetAdminPort()` to be refactored. 

Will there be more verbose assertions added in the future? Like one case of failing and one of passing?

test/integration/consul-container/libs/service/gateway.go
```diff
@@ -18,8 +18,8 @@ import (
 	"github.com/hashicorp/consul/test/integration/consul-container/libs/utils"
 )
 
-// gatewayContainer
-type gatewayContainer struct {
+// GatewayContainer
+type GatewayContainer struct {
```
huikang, line 22: Why are we exposing `GatewayContainer`, given it satisfies the `Service` interface?

test/integration/consul-container/test/troubleshoot/troubleshoot_proxy_test.go
```diff
@@ -0,0 +1,73 @@
...
+		InjectAutoEncryption: true,
+	})
+
+	_, clientService := observability.CreateServices(t, cluster)
+
+	clientSidecar, ok := clientService.(*libservice.ConnectContainer)
+	require.True(t, ok)
+	_, port := clientSidecar.GetInternalAdminAddr()
+	// wait for envoy
+	time.Sleep(10 * time.Second)
```
huikang, line 29: can we use retry instead of sleep?

test/integration/consul-container/test/upgrade/peers_http_test.go
```diff
@@ -49,7 +49,9 @@ func TestPeering_UpgradeToTarget_fromLatest(t *testing.T) {
 		acceptingClient, err := acceptingCluster.GetClient(nil, false)
 		require.NoError(t, err)
 
-		_, gatewayAdminPort := dialing.Gateway.GetAdminAddr()
+		gatewayProxy, ok := dialing.Gateway.(*libservice.GatewayContainer)
```
huikang, line 52: I think dialing.Gateway is a `libservice.Service` interface so there's no need to cast the type. Is there anything changed in this PR.

huikang (commented):
Some non-blocking questions. Thanks.

test/integration/consul-container/libs/service/gateway.go
malizz: Only `GatewayContainer` and `ConnectContainer `should have functions `getAdminAddr` given that it means envoy admin address and non-connect containers should not call this function. Removing `GetAdminAddr` from Service interface, I then have to do a type assertion before calling `GetAdminAddr`. That's why I exposed `GatewayContainer`

test/integration/consul-container/test/upgrade/peers_http_test.go
malizz: explained above, `GetAdminAddr` should only be called from GatewayContainer and ConnectContainer types.

test/integration/consul-container/test/troubleshoot/troubleshoot_proxy_test.go
malizz: This one should be retriable, but for the failure case I cannot retry.

test/integration/consul-container/libs/service/gateway.go
huikang: Got it. But we can add empty function for the other types, which I remember is `exampleContainer` so that we can reduce the size of this PR. Additionally, I think the main purpose of using the `type Service interface ` is to not exposing these internal structs.

DanStough: My original intention with the `Service` interface was to be the lowest common denominator of anything that could be part of the consul catalog.  Concrete types would have additional methods applicable to that type as needed. We can certainly change it based on consensus, but I'm +1 smaller interfaces where everyone implements the behavior.

test/integration/consul-container/test/troubleshoot/troubleshoot_proxy_test.go
```diff
@@ -0,0 +1,73 @@
...
+	_, outputReader, err := clientSidecar.Exec(context.Background(), []string{"consul", "troubleshoot", "proxy",
+		"-envoy-admin-endpoint", fmt.Sprintf("localhost:%v", clientAdminPort),
+		"-upstream-envoy-id", libservice.StaticServerServiceName})
+	require.NoError(t, err)
+	buf, err := io.ReadAll(outputReader)
+	require.NoError(t, err)
+	require.Contains(t, string(buf), "certificates are valid")
+	require.Contains(t, string(buf), fmt.Sprintf("listener for upstream \"%s\" found", libservice.StaticServerServiceName))
+	require.Contains(t, string(buf), fmt.Sprintf("route for upstream \"%s\" found", libservice.StaticServerServiceName))
+	require.Contains(t, string(buf), "no healthy endpoints for cluster")
```
DanStough, line 72: Can we chain this with the previous test and segment with `t.Run`? Every time we create a cluster, we create a new docker network. Not only will it be more resource intensive to run them separately, but allocating more network blocks will prevent other tests from running in parallel.

test/integration/consul-container/test/troubleshoot/troubleshoot_upstream_test.go
```diff
@@ -0,0 +1,35 @@
...
+	clientSidecar, ok := clientService.(*libservice.ConnectContainer)
+	require.True(t, ok)
+	_, port := clientSidecar.GetInternalAdminAddr()
+	// wait for envoy
+	time.Sleep(5 * time.Second)
+	_, outputReader, err := clientSidecar.Exec(context.Background(), []string{"consul", "troubleshoot", "upstreams", "-envoy-admin-endpoint", fmt.Sprintf("localhost:%v", port)})
+	require.NoError(t, err)
+	buf, err := io.ReadAll(outputReader)
+	require.NoError(t, err)
+	require.Contains(t, string(buf), libservice.StaticServerServiceName)
```
DanStough, line 34: Same comment about consolidation - maybe we just have one test case that calls this first with `t.Run`.

test/integration/consul-container/test/observability/util.go
```diff
@@ -0,0 +1,50 @@
+package observability
```
DanStough, line 1: I think a better home for this would be in `lib/service/helpers.go`

DanStough (commented):
Small stuff maybe we can negotiate

test/integration/consul-container/test/troubleshoot/troubleshoot_proxy_test.go
```diff
@@ -0,0 +1,73 @@
...
+	libservice "github.com/hashicorp/consul/test/integration/consul-container/libs/service"
+	"github.com/hashicorp/consul/test/integration/consul-container/libs/topology"
+	"github.com/hashicorp/consul/test/integration/consul-container/test/observability"
+	"github.com/stretchr/testify/require"
+	"io"
+	"testing"
+	"time"
+)
+
+func TestTroubleshootProxy_Success(t *testing.T) {
```
DanStough, line 16: You can add `t.Parrallel` on all of these. It's just missing from the AccessLog test by accident.

test/integration/consul-container/test/troubleshoot/troubleshoot_proxy_test.go
malizz: That's a great idea, thanks!

test/integration/consul-container/test/troubleshoot/troubleshoot_proxy_test.go
malizz: with parallel I had some tests failing. I think there's a race condition in createCluster. But I will address your previous comment and look into the performance.

huikang: @DanStough , given more tests have been added to consul-container, shall we increase the number of parallel jobs (i think the current value is `p=4`)?

test/integration/consul-container/test/troubleshoot/troubleshoot_proxy_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"context"
+	"fmt"
+	libcluster "github.com/hashicorp/consul/test/integration/consul-container/libs/cluster"
+	libservice "github.com/hashicorp/consul/test/integration/consul-container/libs/service"
+	"github.com/hashicorp/consul/test/integration/consul-container/libs/topology"
+	"github.com/hashicorp/consul/test/integration/consul-container/test/observability"
+	"github.com/stretchr/testify/require"
+	"io"
+	"testing"
+	"time"
```
huikang, line 13: nit: we can re-gourp the import path.

test/integration/consul-container/test/troubleshoot/troubleshoot_proxy_test.go
```diff
@@ -0,0 +1,73 @@
...
+	clientSidecar, ok := clientService.(*libservice.ConnectContainer)
+	require.True(t, ok)
+	_, port := clientSidecar.GetInternalAdminAddr()
+	// wait for envoy
+	time.Sleep(10 * time.Second)
+	_, outputReader, err := clientSidecar.Exec(context.Background(), []string{"consul", "troubleshoot", "proxy",
+		"-envoy-admin-endpoint", fmt.Sprintf("localhost:%v", port),
+		"-upstream-envoy-id", libservice.StaticServerServiceName})
+	require.NoError(t, err)
+	buf, err := io.ReadAll(outputReader)
```
huikang, line 34: I think you can make the Exec returns the string so that we don't need line 34 in other places. Please refer to https://github.com/hashicorp/consul/blob/78a4b5fd6f812c1ba288aa0ad84de62c6f4a88d3/test/integration/consul-container/libs/cluster/container.go#L364-L366

test/integration/consul-container/test/troubleshoot/troubleshoot_proxy_test.go
DanStough: @huikang Sure. RB wrote some code similar to `freeport` which will throttle the tests when docker networks aren't available, so it may not increase the speed of the tests at some point.

DanStough (commented):
@malizz thoughts about these two previous comments?
1. Moving `CreateServices` to `lib/service/helpers.go`
1. Using `t.Run()` to put all the tests into one case.

malizz:
> @malizz thoughts about these two previous comments?
> 
> 1. Moving `CreateServices` to `lib/service/helpers.go`
> 2. Using `t.Run()` to put all the tests into one case.

Right, I will address them today. Thanks.

DanStough (approved):
Awesome! Thank you for all the great improvements on the test framework 🍾 🥳

PR #16224, opened 2023-02-09
support vault auth config for alicloud ca provider

eikenb:
Add support for using existing vault auto-auth configurations as the provider configuration when using Vault's CA provider with AliCloud.

AliCloud requires 2 extra fields to enable it to use STS (it's preferred auth setup). Our vault-plugin-auth-alicloud package contained a method to help generate them as they require you to make an http call to a faked endpoint proxy to get them (url and headers base64 encoded).

### PR Checklist

* [X] updated test coverage
* [x] ~external facing docs updated~ (not doing on purpose)
* [X] not a security concern


agent/connect/ca/provider_vault_auth_alicloud.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"fmt"
+
+	"github.com/aliyun/alibaba-cloud-sdk-go/sdk/auth/credentials/providers"
+	"github.com/hashicorp/consul/agent/structs"
+	"github.com/hashicorp/vault-plugin-auth-alicloud/tools"
+)
+
+func NewAliCloudAuthClient(authMethod *structs.VaultAuthMethod) (*VaultAuthClient, error) {
```
cthain, line 11: There is no `legacyCheck` here like there is for the other auth methods. Will this break backwards compatibility for configurations using the `/login` API directly?

cthain (commented):
Looking good. I left a couple of minor suggestions/comments.

agent/connect/ca/provider_vault_auth_alicloud.go
```diff
@@ -0,0 +1,44 @@
...
+	if _, ok := authMethod.Params["role"].(string); !ok {
```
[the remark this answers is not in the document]
eikenb: Good idea. Thanks!

agent/connect/ca/provider_vault_auth_alicloud.go
eikenb: The PMs don't think anyone is using Alicloud with mesh ATM and weren't even sure Alicloud was supported for Consul Mesh, but we had decided to go ahead and merge it as it seemed possible and the code was self contained.

I took the above to mean that backwards compatibility wasn't necessary for it, that we more wanted it there as a way forward. Does that make sense or do you think it'd be better to keep it in line with the others?

cthain: Ok, thanks for the explanation. I feel like if it's not too much work to include the `legacyCheck` that we should do it, even if no one's using this and it's just for consistency across all of our auth method implementations. I'll leave it up to you :+1:

agent/connect/ca/provider_vault_auth_test.go
```diff
@@ -428,3 +430,89 @@ func TestVaultCAProvider_AzureAuthClient(t *testing.T) {
...
+		"no-role": {
+			authMethod: &structs.VaultAuthMethod{
+				Type: VaultAuthMethodTypeAliCloud,
+				Params: map[string]interface{}{
+					"region": "test-region",
+				},
+			},
+			expErr: fmt.Errorf("role is required for AliCloud login"),
+		},
+		"no-region": {
```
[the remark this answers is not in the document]
eikenb: I think that level of detail in testing is not worth it without a specific case. Once the tests exercise the general path and communicate the type of errors the code might produce, it is enough.

cthain (approved):
:+1:

agent/connect/ca/provider_vault_auth_alicloud.go
eikenb: I'll go ahead and add it... I had almost talked myself into it when writing about keeping it in line with the others. Seems like it should be the same (with legacyCheck) or it would need a comment why it didn't. I think just adding it will be easier. :)

eikenb: Added legacyCheck + a test for it.

PR #16228, opened 2023-02-10
[API Gateway] Update simple test to leverage intentions and multiple listeners

andrewstucki:
### Description
This updates our TCP listener test to leverage intentions and multiple listeners with different routes bound to them. One service has its intentions from the gateway denied, the other has them allowed. The net result should be that we create listeners and upstreams for both routes and envoy knows about them, but only one will actually work, the other will reject our attempted connection.

In the process of testing this I ran across a bug in the way the upstreams were being tracked for the proxycfg generation. Essentially each time we were getting a response for a route we were newly subscribed to, it was overwriting all of the upstreams that we were tracking. When you have multiple routes bound this means that only a single route will ever have its upstreams tracked. I swapped the upstream tracking code to leverage some maps that are keyed by the route ref we're using elsewhere for tracking the route to fix this.

Currently this is a stacked PR on https://github.com/hashicorp/consul/pull/16225, so that should get merged first.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/proxycfg/snapshot.go
```diff
@@ -641,6 +640,55 @@ func (c *configSnapshotMeshGateway) isEmptyPeering() bool {
...
 
+type upstreamIDSet map[UpstreamID]struct{}
+
+func (u upstreamIDSet) add(uid UpstreamID) {
+	u[uid] = struct{}{}
+}
+
+type routeUpstreamSet map[structs.ResourceReference]upstreamIDSet
+
+func (r routeUpstreamSet) has(uid UpstreamID) bool {
```
nathancoleman, line 651: I would rename this since `set.has()` reads as though it's checking for a key in a map instead of looping through all the values
```suggestion
func (r routeUpstreamSet) hasUpstream(uid UpstreamID) bool {
```

nathancoleman (commented):
LGTM, just waiting for the rebase/merge since #16225 merged

nathancoleman (approved):

PR #16229, opened 2023-02-10
grpc: fix data race in balancer registration

boxofrad:
### Description
Registering gRPC balancers is thread-unsafe because they are stored in a [global map](https://github.com/grpc/grpc-go/blob/dba26e15a07f43875ccf806a2dd6cbcbc1c12eab/balancer/balancer.go#LL41) variable that is accessed without holding a lock. Therefore, it's expected that balancers are registered _once_ at the beginning of your program (e.g. in a package `init` function) and certainly not after you've started dialing connections, etc.

> NOTE: this function must only be called during initialization time (i.e. in an init() function), and is not thread-safe.

While this is fine for us in production, it's challenging for tests that spin up multiple agents in-memory. We currently register a balancer per-agent which holds agent-specific state that cannot safely be shared.

This commit introduces our own registry that _is_ thread-safe, and implements the [`Builder` interface](https://pkg.go.dev/google.golang.org/grpc/balancer#Builder) such that we can call gRPC's `Register` method once, on start-up. It uses the same pattern as our [resolver registry](https://github.com/hashicorp/consul/blob/78a4b5fd6f812c1ba288aa0ad84de62c6f4a88d3/agent/grpc-internal/resolver/registry.go) where we use the dial target's host (aka "authority"), which is unique per-agent, to determine which builder to use.

### DIFF
--- a/agent/setup.go
+++ b/agent/setup.go
@@ -55,11 +55,18 @@ type BaseDeps struct {
 
 type ConfigLoader func(source config.Source) (config.LoadResult, error)
 
-func NewBaseDeps(configLoader ConfigLoader, logOut io.Writer, providedLogger hclog.InterceptLogger) (BaseDeps, error) {
+func NewBaseDeps(configLoader ConfigLoader, logOut io.Writer, providedLogger hclog.InterceptLogger) (BaseDeps, func(), error) {
+	var balancerBuilder *balancer.Builder
+	cleanup := func() {
+		if balancerBuilder != nil {
+			balancerBuilder.Deregister()
+		}
+	}
+
 	d := BaseDeps{}
 	result, err := configLoader(nil)
 	if err != nil {
-		return d, err
+		return d, cleanup, err
 	}
 	d.WatchedFiles = result.WatchedFiles
 	cfg := result.RuntimeConfig
@@ -71,7 +78,7 @@ func NewBaseDeps(configLoader ConfigLoader, logOut io.Writer, providedLogger hcl
 	} else {
 		d.Logger, err = logging.Setup(logConf, logOut)
 		if err != nil {
-			return d, err
+			return d, cleanup, err
 		}
 	}
 
@@ -85,7 +92,7 @@ func NewBaseDeps(configLoader ConfigLoader, logOut io.Writer, providedLogger hcl
 
 	cfg.NodeID, err = newNodeIDFromConfig(cfg, d.Logger)
 	if err != nil {
-		return d, fmt.Errorf("failed to setup node ID: %w", err)
+		return d, cleanup, fmt.Errorf("failed to setup node ID: %w", err)
 	}
 
 	isServer := result.RuntimeConfig.ServerMode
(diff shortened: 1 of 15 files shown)


agent/setup.go
```diff
@@ -55,11 +55,18 @@ type BaseDeps struct {
...
 type ConfigLoader func(source config.Source) (config.LoadResult, error)
 
-func NewBaseDeps(configLoader ConfigLoader, logOut io.Writer, providedLogger hclog.InterceptLogger) (BaseDeps, error) {
+func NewBaseDeps(configLoader ConfigLoader, logOut io.Writer, providedLogger hclog.InterceptLogger) (BaseDeps, func(), error) {
+	var balancerBuilder *balancer.Builder
+	cleanup := func() {
+		if balancerBuilder != nil {
+			balancerBuilder.Deregister()
+		}
+	}
```
boxofrad, line 64: This feels a bit gross. But so is mutating global state here! We probably need to rethink `BaseDeps` in general.

banks: I can think of the other alternatives:
 1. Pass in a context that lives for the live of the agent. I think we already have ones created from stopChans in a few other places and just run deregistration when the context is closed.
 2. Since `BaseDeps` is a type, make it an `io.Closer` and have the agent call `Close` during shutdown, then hook deregistration etc. up to it here before returning.

Got to be honest though, I don't think either of them is clearly better than what you've done here so I'd be inclined to keep it this way unless/until this becomes really unpleasant to maintain for some reason 🤷 .

command/agent/agent.go
```diff
@@ -165,7 +165,7 @@ func (c *cmd) run(args []string) int {
 		return 1
 	}
 
-	bd, err := agent.NewBaseDeps(loader, logGate, nil)
+	bd, _, err := agent.NewBaseDeps(loader, logGate, nil)
```
banks, line 168: Should we store the `cleanup` func and execute it on agent shutdown? Then it wouldn't be a special case for tests? I realise our general assumption is that agents outside of tests always live for the whole process lifetime but still I feel like we should aim to move towards a world where agents don't just leak by default and need special cases to clean up in tests!

If we did that both here and in `OverrideDeps` which is called by `TestAgent` then would be a small step towards agents actually cleaning themselves up properly on shutdown all the time rather than needing testing special cases?

No strong opinon though, there may be other downsided to that approach I'm not thinking of.

banks:
Amazing Dan 🎉 👏 

Overall this looks way better than the racey approach and thankfully is "only" a plumbing change rather than a reworking of how we use Balancers etc. Some thoughts inline but mostly thinking-out-loud rather than blockers for this!

command/agent/agent.go
boxofrad: Agreed! I almost added `defer cleanup()` here, but wasn't sure if there was much point as the process exits once the agent has shut down anyway.

I like the idea in your other comment of adding a `Close` method to `BaseDeps`, which we could call from `ShutdownAgent` to solve both cases. We could even move the following 2 lines there too:

https://github.com/hashicorp/consul/blob/9108aace2e05f16179ea4d6471fdc8dd067c67bc/agent/agent.go#L1601-L1604

rboyer (approved):

PR #16230, opened 2023-02-10
Fix peering acceptors in secondary datacenters.

hashi-derek:
Prior to this commit, secondary datacenters could not be initialized as peering acceptors if ACLs were enabled. This is due to the fact that internal server-to-server API calls would fail because the management token was not generated. This PR makes it so that both primary and secondary datacenters generate their own management token whenever a leader is elected in their respective clusters.

### DIFF
--- a/.changelog/16230.txt
+++ b/.changelog/16230.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+peering: Fix issue where secondary wan-federated datacenters could not be used as peering acceptors.
+```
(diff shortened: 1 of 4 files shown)


.changelog/16230.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug-fix
```
DanStough, line 1: Won't show up in the release notes with this.
```suggestion
```release-note:bug
```

DanStough (approved):
LGTM, just need to fix the release note

PR #16236, opened 2023-02-10
[API Gateway] Add integration test for HTTP routes

andrewstucki:
### Description
This adds an integration test for basic http route functionality. In the process it fixes a number of things:

1. The`api` subpackage was missing HTTP route definitions for a number of fields
2. To match the behavior in the proxycfg construction, I made the `kind` specifier for parents on routes optional in the controller, an empty `kind` will be treated as `api-gateway`
3. The http synthetic discovery chain code was returning a single discovery chain per-listener, it needed to do a discovery chain per-service on a listener instead as that is what the xDS code is expecting
4. The chain synthesis code requires a trust domain, this goes into constructing the SNI values for the discovery chain. Since the chain is based off of other chains, we can pull the trust domain off of the SNI for an existing target in one of our upstream chains
5. We needed to add a per-listener suffix on the service names that we generate for our synthetic http config entries to avoid collisions and the generated discovery chains overwriting each other because of identical upstream ids based off of the synthesized service name
6. Because the endpoints which resolve the IP/port of our actual upstream services rely on `WatchedUpstreamEndpoints`, we need to merge whatever was being watched onto the generated snapshot, indexed by the synthetic upstream ids

This builds on https://github.com/hashicorp/consul/pull/16228 and should be merged after it.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/connect/sni.go
```diff
@@ -65,6 +65,39 @@ func ServiceSNI(service string, subset string, namespace string, partition strin
...
+			// subset, service, namespace, datacenter, internal, trustDomain
+			return dotSplitLast(target.SNI, 6)
+		}
+	default:
+		if target.ServiceSubset == "" {
+			// service, namespace, partition, datacenter, internalVersion, trustDomain
+			return dotSplitLast(target.SNI, 6)
+		} else {
+			// subset, service, namespace, partition, datacenter, internalVersion, trustDomain
+			return dotSplitLast(target.SNI, 7)
```
jm96441n, line 96: where does the `5` `6` and `7` numbers come from? should we define those as constants?

t-eckert (approved):

PR #16237, opened 2023-02-10
[API Gateway] Fix missing references to enterprise metadata

andrewstucki:
### Description
Leaving out enterprise metadata here causes our later route lookups to fail due to having enterprise metadata in one place and not another.

Verified the fix on an enterprise build.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/proxycfg/api_gateway.go
+++ b/agent/proxycfg/api_gateway.go
@@ -243,10 +243,10 @@ func (h *handlerAPIGateway) handleInlineCertConfigUpdate(_ context.Context, u Up
 		return fmt.Errorf("invalid type for config entry: %T", resp.Entry)
 	}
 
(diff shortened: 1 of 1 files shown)


agent/proxycfg/api_gateway.go
```diff
@@ -243,10 +243,10 @@ func (h *handlerAPIGateway) handleInlineCertConfigUpdate(_ context.Context, u Up
 		return fmt.Errorf("invalid type for config entry: %T", resp.Entry)
 	}
 
-	// TODO Consider if unset SectionName and acl.EnterpriseMeta could trip us up
```
nathancoleman, line 246: Look, it did trip us up! 😆 Thanks for addressing

nathancoleman (approved):

PR #16241, opened 2023-02-10
Bump x/time to 0.3.0 and fix related breakage linked to RPCRateLimit

analogue:
### Description
The old version of `x/time` we were using was not affected by setting the `limit` to `-1` and thus `Allow()`ed calls even when configured incorrectly. Newer versions ( > `0.1.0`) are not as lenient and block calls when configured with `limit=-1`.  This is why we saw unusual breakage in the consul-k8s integration tests that use vault enterprise since the configs are reloaded a number of times on consul server startup.

This PR fixes the runtime configuration for the `RPCRateLimit` to default to `rate.Inf` when configuration is (re)loaded. This brings the behavior inline with other limit based default configurations put in place by @jmurret when he implemented this functionality for request limits.  

Shout outs to @dhia and @JadhavPoonam for debugging the issue with me which finally resulted in uncovering this particularly sneaky bug. 

* `TestVault` integration test in consul-k8s now passes when this fix is applied to consul enterprise. 
* Unit test updated to verify default value of `rate.Inf` 

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

primary: @dhiaayachi


JadhavPoonam (approved):
Thank you for the detailed PR description. 🚀

PR #16242, opened 2023-02-11
provider ca support vault auth clouf foundry conf

eikenb:
Add support for using existing vault auto-auth cofiguration as CA provider auth configuration for Cloud Foundry.

CF requires 2 environment variables that contain paths to certs and key files that are required for login/auth.

### PR Checklist

* [X] updated test coverage
* [X] not a security concern


eikenb:
Turns out CF doesn't need this as it isn't compatible.

PR #16245, opened 2023-02-13
Watch inline-certificates on api-gateway before attempting to stash

nathancoleman:
### Description
`handlerAPIGateway` collects and stores config entries required for xDS generation on the config snapshot. You must initialize a watch on the config entry before attempting to store it in the `watchMap`, otherwise storing is a no-op. This watching wasn't done for the `inline-certificates` on the `bound-api-gateway`, so they were missing when it came time for xDS generation.

Working on a full automated test in followup PR.

### Links
https://github.com/hashicorp/consul/blob/e6b55d1d81c6e90dd5d09e7dfb24d1db7604b7b5/agent/proxycfg/internal/watch/watchmap.go#L87-L97

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


nathancoleman:
I don't believe this PR is needed

PR #16254, opened 2023-02-14
Support inline-certificate as payload of config-entry wrapper

nathancoleman:
Co-Authored-By: Andrew Stucki <3577250+andrewstucki@users.noreply.github.com>

### Description
We missed the `inline-certificate` when adding to the ConfigEntry proto model. This adds it, same as we already did with `http-route`, `api-gateway`, etc.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


t-eckert (approved):
Excellent

PR #16257, opened 2023-02-14
Fix mesh gateways incorrectly matching peer locality.

hashi-derek:
This fixes an issue where local mesh gateways use an incorrect address when attempting to forward traffic to a peered datacenter. Prior to this change it would use the lan address instead of the wan if the locality matched. This should never be done for peering, since we must route all traffic through the remote mesh gateway.


ndhanushkodi (approved):
LGTM!

PR #16260, opened 2023-02-14
Fix a couple inconsistencies in `operator usage instances` command

kyhavlov:
Update one of the help texts that got missed when moving the command under `operator` and prevent `-billable` and `-connect` from both being set at the same time.


david-yu:
@kyhavlov I believe we need to merge this for 1.15.0? Could we get someone to review?

rboyer (approved):

PR #16262, opened 2023-02-14
add provider ca auth support for kubernetes

eikenb:
Adds support for Kubernetes jwt/token file based auth. Only needs to read the file and save the contents as the jwt/token.

### PR Checklist

* [X] updated test coverage
* [X] not a security concern

### DIFF
--- a/agent/connect/ca/provider_vault.go
+++ b/agent/connect/ca/provider_vault.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"os"
 	"strings"
 	"sync"
 	"time"
@@ -922,6 +921,14 @@ func vaultLogin(client *vaultapi.Client, authMethod *structs.VaultAuthMethod) (*
 	return resp, nil
 }
 
+// Note the authMethod's parameters (Params) is populated from a freeform map
+// in the configuration where they could hardcode values to be passed directly
+// to the `auth/*/login` endpoint. Each auth method's authentication code
+// needs to handle two cases:
+// - The legacy case (which should be deprecated) where the user has
+// hardcoded login values directly (eg. a `jwt` string)
+// - The case where they use the configuration option used in the
+// vault agent's auth methods.
 func configureVaultAuthMethod(authMethod *structs.VaultAuthMethod) (VaultAuthenticator, error) {
 	if authMethod.MountPath == "" {
 		authMethod.MountPath = authMethod.Type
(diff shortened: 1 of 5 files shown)


agent/connect/ca/provider_vault.go
```diff
@@ -922,6 +921,13 @@ func vaultLogin(client *vaultapi.Client, authMethod *structs.VaultAuthMethod) (*
 	return resp, nil
 }
 
+// Note the authMethod's parameters (Params) is populated from a freeform map
+// in the configuration where they could hardcode values to be passed directly
+// to the `auth/*/login` endpoint. So each auth method's authentication code
+// needs to handle both these cases. The legacy case (which should be
+// deprecated) where the user has hardcoded login values directly (eg. a `jwt`
```
cthain, line 928: Do you think that we should officially deprecate the direct `auth/<method>/login` configuration support? If so we should probably add an issue to our backlog.

cthain (approved):

agent/connect/ca/provider_vault_auth_test.go
```diff
@@ -428,3 +428,59 @@ func TestVaultCAProvider_AzureAuthClient(t *testing.T) {
...
+	tokenF, err := os.CreateTemp("", "token-path")
+	require.NoError(t, err)
+	_, err = tokenF.WriteString("test-token")
+	require.NoError(t, err)
+	err = tokenF.Close()
+	require.NoError(t, err)
+
+	defer func() {
+		os.Remove(tokenF.Name())
+	}()
```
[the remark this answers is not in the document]
eikenb: Probably not a big deal to leak a few files to /tmp in weird testing conditions but better to not have it at all... I'll bump that defer up a few lines as suggested.

agent/connect/ca/provider_vault.go
```diff
@@ -922,6 +921,13 @@ func vaultLogin(client *vaultapi.Client, authMethod *structs.VaultAuthMethod) (*
...
+// string) and the case where they use  the configuration option used in the
+// vault agent's auth methods.
```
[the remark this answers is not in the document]
eikenb: Nice. Thanks!

agent/connect/ca/provider_vault.go
eikenb: Added to our backlog.

agent/connect/ca/provider_vault_auth_k8s.go
```diff
@@ -0,0 +1,47 @@
...
+func NewK8sAuthClient(authMethod *structs.VaultAuthMethod) (*VaultAuthClient, error) {
+	params := authMethod.Params
+	role, ok := params["role"].(string)
+	if !ok || strings.TrimSpace(role) == "" {
+		return nil, fmt.Errorf("missing 'role' value")
+	}
+	// don't check for `token_path` as it is optional
+
+	authClient := NewVaultAPIAuthClient(authMethod, "")
+	// Note the `jwt` can be passed directly in the authMethod as the it's Params
```
skpratt, line 20:
```suggestion
	// Note the `jwt` can be passed directly in the authMethod as a Param value
```

agent/connect/ca/provider_vault_test.go
```diff
@@ -113,7 +113,7 @@ func TestVaultCAProvider_configureVaultAuthMethod(t *testing.T) {
 		"gcp":         {expLoginPath: "auth/gcp/login", params: map[string]interface{}{"type": "iam", "role": "test-role"}},
 		"jwt":         {expLoginPath: "auth/jwt/login"},
 		"kerberos":    {expLoginPath: "auth/kerberos/login"},
-		"kubernetes":  {expLoginPath: "auth/kubernetes/login", params: map[string]interface{}{"jwt": "fake"}},
```
skpratt, line 116: Until we officially deprecate support for passing the jwt directly, should we keep test coverage of that functionality?
Would this test case now break, since there is no role specified in the params?

eikenb: TLDR; I'm adding a legacy test case to the k8s specific tests in provider_vault_auth_test.go.

--

That test really should have failed previously as the role field is required along with the jwt. K8s requires both for login. IE. the code should have tested for the required role field but didn't. I don't check for that here as this test is more just to exercise the configureVaultAuthMethod and make sure it returns the correct auth method. I added the expanded tests for these sorts of things to the provider_vault_auth_test.go file... though looking at it I see it doesn't have a specific legacy-case check I have in a few others. I'll add that.

skpratt (approved):

PR #16266, opened 2023-02-15
add provider ca support for jwt file base auth

eikenb:
Adds support for a jwt token in a file. Simply reads the file and sends the read in jwt token along to the vault login.

### PR Checklist

* [X] updated test coverage
* [X] not a security concern

### DIFF
--- a/agent/connect/ca/provider_vault_auth_jwt.go
+++ b/agent/connect/ca/provider_vault_auth_jwt.go
@@ -1,0 +1,68 @@
+package ca
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/hashicorp/consul/agent/structs"
+)
+
+func NewJwtAuthClient(authMethod *structs.VaultAuthMethod) (*VaultAuthClient, error) {
+	params := authMethod.Params
+
+	role, ok := params["role"].(string)
+	if !ok || strings.TrimSpace(role) == "" {
+		return nil, fmt.Errorf("missing 'role' value")
+	}
+
+	// The path is required for the auto-auth config, but this auth provider
+	// seems to be used for jwt based auth by directly passing the jwt token.
+	// So we only require the token file path if the token string isn't
+	// present.
+	needTokenPath := true
+	if _, ok := hasJWT(params); ok {
+		needTokenPath = false
+	}
+	if needTokenPath {
+		tokenPath, ok := params["path"].(string)
+		if !ok || strings.TrimSpace(tokenPath) == "" {
+			return nil, fmt.Errorf("missing 'path' value")
+		}
+	}
+
+	authClient := NewVaultAPIAuthClient(authMethod, "")
+	authClient.LoginDataGen = JwtLoginDataGen
+	return authClient, nil
+}
+
+func JwtLoginDataGen(authMethod *structs.VaultAuthMethod) (map[string]any, error) {
+	params := authMethod.Params
+	role := params["role"].(string)
+
+	if jwt, ok := hasJWT(params); ok {
+		return map[string]any{
+			"role": role,
+			"jwt":  jwt,
+		}, nil
+	}
+
+	tokenPath := params["path"].(string)
+
+	rawToken, err := os.ReadFile(tokenPath)
+	if err != nil {
+		return nil, err
+	}
+
+	return map[string]any{
+		"role": role,
+		"jwt":  strings.TrimSpace(string(rawToken)),
+	}, nil
+}
+
+func hasJWT(params map[string]any) (string, bool) {
+	if jwt, ok := params["jwt"].(string); ok && strings.TrimSpace(jwt) != "" {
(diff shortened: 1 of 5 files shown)


agent/connect/ca/provider_vault_auth_jwt.go
```diff
@@ -0,0 +1,73 @@
...
+	// The path is required for the auto-auth config, but this auth provider
+	// seems to be used for jwt based auth by directly passing the jwt token.
+	// So we only require the token file path if the token string isn't
+	// present.
+	needTokenPath := true
+	// support legacy setup that allows directly passing the `jwt`
+	if _, ok := hasJWT(params); ok {
+		needTokenPath = false
+	}
+	if needTokenPath {
```
pglass, line 28: nit / not-blocking: the `needTokenPath` could be removed, I think?

```suggestion
	// support legacy setup that allows directly passing the `jwt`
	if _, ok := hasJWT(params); !ok {
```

agent/connect/ca/provider_vault_auth_jwt.go
```diff
@@ -0,0 +1,73 @@
...
+		return nil, err
+	}
+
+	return map[string]any{
+		"role": role,
+		"jwt":  strings.TrimSpace(string(rawToken)),
+	}, nil
+}
+
+// Note the `jwt` can be passed directly in the authMethod as the it's Params
```
pglass, line 65: wording: `the it's` 

```suggestion
// Note: the `jwt` can be passed directly in the authMethod as Params
```

agent/connect/ca/provider_vault_auth_test.go
```diff
@@ -301,3 +301,78 @@ func TestVaultCAProvider_AWSLoginDataGenerator(t *testing.T) {
...
+	tokenF, err := os.CreateTemp("", "token-path")
+	require.NoError(t, err)
+	_, err = tokenF.WriteString("test-token")
+	require.NoError(t, err)
+	err = tokenF.Close()
+	require.NoError(t, err)
+
+	defer func() {
+		os.Remove(tokenF.Name())
+	}()
```
pglass, line 315: nit: use `t.Cleanup`, although doesn't make a functional difference in this case. But `t.Cleanup` is useful when moving test cleanups into helper functions, so I try to be in the habit of using it for all test cleanups in case of refactoring 🙂 

```suggestion
	t.Cleanup(func() {
		os.Remove(tokenF.Name())
	})
```

pglass (approved):
👍 Looks good!

agent/connect/ca/provider_vault_auth_jwt.go
eikenb: Opps... looks like I rewrote that sentence once to many times. Thanks!

agent/connect/ca/provider_vault_auth_jwt.go
eikenb: I was seriously thinking about reworking this completely anyways as it could use the same legacyCheck() used in the Azure PR to do something similar. Having them all use the same method would be good even if it were very simple in this case. With the review for that (azure) PR completed I think I'll get it merged so I can rework this code to use it and clean this up. Thanks.

eikenb: Rework of this pushed up. Going to resolve this thread as the code in question is gone.

agent/connect/ca/provider_vault_auth_jwt.go
```diff
@@ -0,0 +1,50 @@
...
+	role, ok := params["role"].(string)
+	if !ok || strings.TrimSpace(role) == "" {
+		return nil, fmt.Errorf("missing 'role' value")
+	}
+
+	authClient := NewVaultAPIAuthClient(authMethod, "")
+	// The path is required for the auto-auth config, but this auth provider
+	// seems to be used for jwt based auth by directly passing the jwt token.
+	// So we only require the token file path if the token string isn't
+	// present.
```
kisunji, line 23: This comment is a little confusing to read - is it referring to L28 below?

kisunji (approved):

agent/connect/ca/provider_vault_auth_test.go
kisunji: Mentioned this in another PR but should this defer/cleanup be moved to before we try calling `WriteString` or `Close`? Otherwise if either error out we'll have a stray file that is not cleaned up.

eikenb: Moved that defer up a few line like the other spot.

agent/connect/ca/provider_vault_auth_jwt.go
eikenb: Opps.. forgot to move that down when I switched the code over to use legacyCheck(). 
IE. you are right that this is about line 28.

PR #16270, opened 2023-02-15
Add WAL documentation. Also fix some minor metrics registration details

banks:
This PR includes the public documentation for the experimental WAL backend merged in #16176.

It also includes a couple of minor non-functional fixes discovered:
 * Verification failures were being logged as INFO not ERROR
 * While WAL and verifier metrics were integrated and working, we weren't registering their definitions for Prometheus.
 * Added tests for the registration to ensure we only add new metrics when the functionality is enabled

## Feedback Requested

* I'd love feedback on the overall clarity and structure of the testing guide.
* Does the location in the information hierarchy make sense? For now it seems like an upgrading thing as most of that content is specific to this version where the backend is experimental. It might be once it's GA we need to find a new home for some of the content about the difference between backends etc?

### PR Checklist

* ~[ ] updated test coverage~
* [x] not a security concern


agent/metrics_test.go
```diff
@@ -432,3 +432,193 @@ func TestHTTPHandlers_AgentMetrics_CACertExpiry_Prometheus(t *testing.T) {
...
+		bootstrap = false
+		`
+
+		a := StartTestAgent(t, TestAgent{HCL: hcl})
+		defer a.Shutdown()
+
+		respRec := httptest.NewRecorder()
+		recordPromMetrics(t, a, respRec)
+
+		require.NotContains(t, respRec.Body.String(), "agent_4_raft_wal")
```
loshz, line 460: I think we might still need to defer/call `respRec.Body.Close()` to free resources?

agent/metrics_test.go
```diff
@@ -432,3 +432,193 @@ func TestHTTPHandlers_AgentMetrics_CACertExpiry_Prometheus(t *testing.T) {
 	})
 
 }
+
+func TestHTTPHandlers_AgentMetrics_WAL_Prometheus(t *testing.T) {
+	skipIfShortTesting(t)
+	// This test cannot use t.Parallel() since we modify global state, ie the global metrics instance
```
loshz, line 438: _Leaving this comment with some gaps in context, so apologies in advance._

I understand why we're doing this, and it seems to be an existing pattern in this file, but modifying global state is usually bad. Not necessarily one for this PR, but could we just clone/create new metrics instances and operate on these instead?

agent/setup.go
```diff
@@ -316,14 +341,40 @@ func getPrometheusDefs(cfg lib.TelemetryConfig, isServer bool) ([]prometheus.Gau
 		raftCounters,
 		rate.Counters,
 	}
+
+	// For some unknown reason, we seem to add the raft counters above without
+	// checking if this is a server like we do above for some of the summaries
+	// above. We should probably fix that but I want to not change behavior right
+	// now. If we are a server, add summaries for WAL and verifier metrics.
+	if isServer && cfg.RaftLogStoreConfig.Verification.Enabled {
+		verifierCounters := make([]prometheus.CounterDefinition, 0)
```
loshz, line 350: Might be a simple optimization to pass the cap here?
```suggestion
		verifierCounters := make([]prometheus.CounterDefinition, 0, len(verifier.MetricDefinitions.Counters))
```

agent/setup.go
```diff
@@ -316,14 +341,40 @@ func getPrometheusDefs(cfg lib.TelemetryConfig, isServer bool) ([]prometheus.Gau
...
+		for _, d := range verifier.MetricDefinitions.Counters {
+			verifierCounters = append(verifierCounters, prometheus.CounterDefinition{
+				Name: []string{"raft", "logstore", "verifier", d.Name},
+				Help: d.Desc,
+			})
+		}
+		counters = append(counters, verifierCounters)
+	}
+	if isServer && cfg.RaftLogStoreConfig.Backend == consul.LogStoreBackendWAL {
+		walCounters := make([]prometheus.CounterDefinition, 0)
```
loshz, line 360: Might be a simple optimization to pass the cap here?
```suggestion
		walCounters := make([]prometheus.CounterDefinition, 0, len(wal.MetricDefinitions.Counters))
```

agent/metrics_test.go
banks: Great callout. This is a shortcoming of go-metrics in general. I'd _love_ to move away from go-metrics to something more modern but that comes with a lot of compatibility issues since all our metrics config and output depends on quirks of how go-metrics has always done things. All HashiCorp products are also built on it so it is a big change to go changing how go-metrics works.

If we wanted to fix just the global state part it would at least mean updating every instrumentation point in our code that calls `metrics.*()` functions to instead call a method on a specific instance which is part of the agent dependencies and plumbed though all the other modules and layers. I'd _love_ to get there one day but it's not been something we've chosen to bite off just yet!

website/content/docs/agent/telemetry.mdx
```diff
@@ -337,11 +337,15 @@ indicator of an actual issue, this metric can be used to diagnose why the `consu
 is high.
 
 If Bolt DB log storage performance becomes an issue and is caused by free list management then setting
-[`raft_boltdb.NoFreelistSync`](/consul/docs/agent/config/config-files#NoFreelistSync) to `true` in the server's configuration
+[`raft_logstore.boltdb.no_freelist_sync`](/consul/docs/agent/config/config-files#raft_logstore_boltdb_no_freelist_sync) to `true` in the server's configuration
 may help to reduce disk IO and log storage operation times. Disabling free list syncing will however increase
 the startup time for a server as it must scan the raft.db file for free space instead of loading the already
 populated free list structure.
 
+~> Note that as of Consul 1.15 there is a [new experimental storage
```
krastin, line 345: Should we call out what the storage backend is?

website/content/docs/agent/telemetry.mdx
```diff
@@ -452,6 +456,11 @@ These metrics are used to monitor the health of the Consul servers.
...
+| `consul.raft.logstore.verifier.checkpoints_written`     | Counts the number of checkpoint entries written to the LogStore. | checkpoints | counter |
+| `consul.raft.logstore.verifier.dropped_reports`         | Counts how many times the verifier routine was still busy when the next checksum came in and so verification for a range was skipped. If you see this happen, consider increasing the interval between checkpoints with [`raft_logstore.verification.interval`](http://localhost:3000/consul/docs/agent/config/config-files#raft_logstore_verification) | reports dropped | counter |
```
krastin, line 460:
```suggestion
| `consul.raft.logstore.verifier.dropped_reports`         | Counts how many times the verifier routine was still busy when the next checksum came in and so verification for a range was skipped. If you see this happen, consider increasing the interval between checkpoints with [`raft_logstore.verification.interval`](/consul/docs/agent/config/config-files#raft_logstore_verification) | reports dropped | counter |
```

website/content/docs/agent/telemetry.mdx
```diff
@@ -452,6 +456,11 @@ These metrics are used to monitor the health of the Consul servers.
...
+| `consul.raft.logstore.verifier.ranges_verified`         | Counts the number of log ranges for which a verification report has been completed. See [monitoring experimental backends](http://localhost:3000/consul/docs/upgrading/instructions/testing-experimental-raft-backend#monitoring) for more information. | log ranges verifications | counter |
+| `consul.raft.logstore.verifier.read_checksum_failures`  | Counts the number of times a range of logs between two check points contained at least one disk corruption. See [monitoring experimental backends](http://localhost:3000/consul/docs/upgrading/instructions/testing-experimental-raft-backend#monitoring) for more information. | disk corruptions | counter |
```
krastin, line 462:
```suggestion
| `consul.raft.logstore.verifier.read_checksum_failures`  | Counts the number of times a range of logs between two check points contained at least one disk corruption. See [monitoring experimental backends](/consul/docs/upgrading/instructions/testing-experimental-raft-backend#monitoring) for more information. | disk corruptions | counter |
```

krastin (commented):
Just a few suggestions here and there.

website/content/docs/upgrading/instructions/testing-experimental-raft-backend.mdx
```diff
@@ -0,0 +1,420 @@
...
+}
+```
+
+**6. Start the target server.**
+
+For example with `systemctl start consul` if using systemd.
+
+**7. Monitor target server raft metrics and logs.**
+
+See the section below on [monitoring WAL tests](#monitoring-wal-tests).
```
danielehc, line 239: The link does not work.

Maybe `(#monitoring)` is the one you meant.

website/content/docs/upgrading/instructions/testing-experimental-raft-backend.mdx
```diff
@@ -0,0 +1,420 @@
...
+was an expedient option when our raft library was first written and always
+assumed we'd replace it with something more purpose-built.
+
+Importantly, a BoltDB database is a single file that only ever grows. Deleting
+the oldest logs which we do regularly when we've made a new snapshots of the
+state, leaves free space in the file that needs to be tracked to be re-used on
+future writes. By contrast a simple segmented log can just delete the oldest log
+files from disk. With BoltDB uses as a log, sudden burst of writes say three
+times larger than the normal volume can suddenly cause the file to grow to about
+several times it's steady-state size. After the next snapshot is taken, and the
```
danielehc, line 41: This, until the end of the paragraph might be a bit confusing to digest.

```suggestion
files from disk. When BoltDB is used as a log backend, sudden burst of writes at a rate 2-3x higher than the normal volume, can suddenly cause the log file to grow to several times it's steady-state size. After the next snapshot is taken, and the
```

website/content/docs/upgrading/instructions/testing-experimental-raft-backend.mdx
```diff
@@ -0,0 +1,420 @@
...
+oldest logs truncated again, the file is left as mostly empty space. Tracking
```
danielehc, line 42: > the file is left as mostly empty space

shall we clarify that it keeps the same size?

website/content/docs/upgrading/instructions/testing-experimental-raft-backend.mdx
```diff
@@ -0,0 +1,420 @@
...
+Even if this has never happened to a catastrophic degree in a cluster, the fact
+that it's a risk has meant that Consul has erred on the side of never letting
+too many logs accumulate in the LogStore. Significantly larger BoltDB files are
+somewhat slower in general because it's a tree and so still has log(N) work to
+do n every write. But out user's experience showed that the larger the file, the
+more likely it is to have a large freelist or suddenly form one after a burst of
+writes. For this reason, the default options for how frequently we make a full
+snapshot and truncate the logs, and for how many logs we keep around have always
+been aggressively set towards keeping BoltDB small rather than using disk IO the
+most efficiently.
```
danielehc, line 57:
```suggestion
Even if this has never happened to a catastrophic degree in a cluster, Consul was tuned to avoid for too many logs to accumulate in the LogStore, in order to reduce/mitigate the risk. Significantly larger BoltDB files are
somewhat slower in general because it's a tree and so still has log(N) work to
do n every write. But out user's experience showed that the larger the file, the
more likely it is to have a large freelist or suddenly form one after a burst of
writes. For this reason, the default options for how frequently we make a full
snapshot and truncate the logs, and for how many logs we keep around have always
been aggressively set towards keeping BoltDB small rather than using disk IO the
most efficiently.
```

danielehc: Also:

> Significantly larger BoltDB files are
somewhat slower in general because it's a tree and so still has log(N) work to
do n every write. But out user's experience showed that the larger the file, the
more likely it is to have a large freelist or suddenly form one after a burst of
writes.

Do we mean that both (the fact it is a tree and the higher probability of freelist) are a reason of bad performances? Or that despite the good performances offered by the tree, freelists drags them down?

> because it's a tree and so still has log(N) work to do n every write
Did we mean `on every write`?

danielehc: Also I am assuming that by `freelist` we refer either to chunks of free space in the file or to the data structures we use to track the free space in the file.

It is not mentioned anywhere else in the file so it might be worth a clarification.

website/content/docs/upgrading/instructions/testing-experimental-raft-backend.mdx
```diff
@@ -0,0 +1,420 @@
...
+
+Other reliability issues such as [followers being unable to catch
+up](/consul/docs/agent/telemetry#raft-replication-capacity-issues) also stem
+from this need to carefully balance the size of the BoltDB log store against how
+long snapshots take to restore - there is a simple solution to that issue if
```
danielehc, line 62: We do not state the solution and the jump to WAL is not fully obvios. We could state the solution, clarify the technical issues to be solved before being able to use the approach (let file grow), something like:

> A simple solution for that issue is ... but could not get implemented unless/until letting logs grow ...

And then specify that WAL follows this approach.

website/content/docs/upgrading/instructions/testing-experimental-raft-backend.mdx
```diff
@@ -0,0 +1,420 @@
...
+letting logs grow much larger to ensure recovery didn't have a potentially
+catastrophic impact on the cluster's write performance.
+
+While not every user will experience a huge difference in performance, the WAL
+backend avoids these performance concerns entirely. It is more performant when
+directly measured due to solving a simpler storage problem than BoltDB was
+designed for. For example it can commit a single log entry with on fsync instead
```
danielehc, line 69:
```suggestion
designed for. For example it can commit a single log entry with one fsync instead
```

Shall we use monospace? >  `fsync`

website/content/docs/upgrading/instructions/testing-experimental-raft-backend.mdx
```diff
@@ -0,0 +1,420 @@
...
+of two, and tends to write 2-3x fewer bytes to the disk to do it. The real
+benefit though is that retaining more logs won't impact write performance at all
+and so reducing disk IO with slower snapshots or keeping logs around to catch up
+slower followers are all possible.
```
danielehc, line 73:
```suggestion
and so strategies for reducing disk IO with slower snapshots or for keeping logs around to permit slower followers to catch up with cluster state, are all possible.
```

website/content/docs/upgrading/instructions/testing-experimental-raft-backend.mdx
```diff
@@ -0,0 +1,420 @@
...
+This guide describes how to safely try and verify it without risking the
+availability of your cluster should there be a latent data-loss issue
+discovered.
+
+## Requirements
+
+- All servers in the Datacenter should be upgraded to Consul 1.15 using the
+  [standard upgrade procedure](/consul/docs/upgrading/general-process) and
+  the [1.15 upgrade notes](/consul/docs/upgrading/upgrade-specific#consul-1-15-x).
+- You need a Consul cluster with at least 3 nodes to safely test the new
```
danielehc, line 111: Is this enough?

We mention below that enabling WAL on a voting server might itself generate an issue:

> If you enable a server to use WAL using Consul OSS or on a voting server with
   Consul Enterprise, it's _possible_ that the WAL could cause corruption of
   that server's state (with the caveats above) _and then_ become the leader and
   replicate that corruption to all other servers.

Or is this more related to the fact then even in case it goes all good users with one server will have downtime?

To be safe the real requirement is to be able to use non-voting servers aka to use Consul Enterprise

website/content/docs/upgrading/instructions/testing-experimental-raft-backend.mdx
```diff
@@ -0,0 +1,420 @@
...
+```
+2023-01-31T14:44:31.174Z [INFO]  agent.server.raft.logstore.verifier: verification checksum OK: elapsed=488.463268ms leaderChecksum=f15db83976f2328c rangeEnd=357802 rangeStart=298132 readChecksum=f15db83976f2328c
+```
+
+
+**2. Select a server to enable WAL on.**
+
+As noted in [Risks](#risks), Consul Enterprise users should select a non-voting
+server at first. For Consul OSS users, or Enterprise users who don't have
+non-voting servers, select one of the follower servers.
```
danielehc, line 185: Shall we hint on the fact that ENT users could test it by adding a new non-voting server in case they don't have one deployed yet? This will reduce their risks.

website/content/docs/upgrading/instructions/testing-experimental-raft-backend.mdx
```diff
@@ -0,0 +1,420 @@
...
+configuration.
+
+WAL is a acronym for "Write-Ahead Log". We called it this because it implements
+a traditional log with rotating, append-only log files. The current `LogStore`
+uses BoltDB which is a copy-on-write BTree which is less optimized for
+append-only workloads.
+
+~> The `wal` backend is considered **experimental** in Consul 1.15. Please test
+it safely in pre-production first and verify on a subset of servers in
+production using this guide before fully enabling it on all servers.
```
trujillo-adam, line 24:
```suggestion
WAL implements a traditional log with rotating, append-only log files. The current `LogStore`
uses BoltDB, which is a copy-on-write BTree. BoltDB is not optimized for append-only workloads.

!>**Upgrade warning**: The WAL backend is experimental. You should test and verify the WAL backend on a subset of servers in pre-production before fully enabling it on all servers.
```
This is one of the designated uses for callouts per the style guide.

website/content/docs/upgrading/instructions/testing-experimental-raft-backend.mdx
```diff
@@ -0,0 +1,420 @@
...
+
+## Why build a new backend?
```
trujillo-adam, line 26:
```suggestion
### WAL versus BoldDB
```
I think this is still part of the introduction. In fact, most of the information in this section, I would argue, isn't documentation and would be better suited to a blog article.

website/content/docs/upgrading/instructions/testing-experimental-raft-backend.mdx
```diff
@@ -0,0 +1,420 @@
...
+`consul.raft.logstore.verifier.{read|write}_checksum_failures` metrics. The
+sooner a corrupt server is handled, the lower the chance of any of the
+[potential risks](#risks) causing problems in your cluster.
+
+### Performance Metrics
+
+The key performance metrics to watch are:
+
+* `consul.raft.commitTime` measures the time to commit new writes on a quorum of
+  servers. It should be the same or lower after deploying WAL. Even if WAL is
```
danielehc, line 396:
```suggestion
* `consul.raft.commitTime` measures the time to commit new writes on a quorum of
  servers. It should be the same or lower after deploying WAL. Even if WAL is
```

In [telemetry.mdx](https://github.com/hashicorp/consul/pull/16270/files#diff-8d2b6d7521b66ab4359f6ccd7e79e955c168d5f3b9d9066b76236ebc782377ab) we use a different definition:

> Measures the time it takes to commit a new entry to the Raft log on the leader.

It might be better to reuse the same definition.

website/content/docs/upgrading/instructions/testing-experimental-raft-backend.mdx
```diff
@@ -0,0 +1,420 @@
...
+  faster for your workload and hardware, it may not be reflected in commitTime
+  until enough followers are using it that the leader doesn't have to wait for a
+  slower one (one in a cluster of three, two in a cluster of five etc.).
+
+* `consul.raft.rpc.appendEntries.storeLogs` measures the time spent persisting
+  logs to disk on each _follower_. It should be the same or lower for
+  WAL-enabled followers.
```
danielehc, line 403: In [telemetry.mdx](https://github.com/hashicorp/consul/pull/16270/files#diff-8d2b6d7521b66ab4359f6ccd7e79e955c168d5f3b9d9066b76236ebc782377ab) we use a different definition:

> Measures the time taken to add any outstanding logs for an agent, since the last appendEntries was invoked

website/content/docs/upgrading/instructions/testing-experimental-raft-backend.mdx
```diff
@@ -0,0 +1,420 @@
...
+
+* `consul.raft.replication.appendEntries.rpc` measures the time taken for each
+  `AppendEntries` RPC from the leader's perspective. If this is significantly
```
danielehc, line 406: In [telemetry.mdx](https://github.com/hashicorp/consul/pull/16270/files#diff-8d2b6d7521b66ab4359f6ccd7e79e955c168d5f3b9d9066b76236ebc782377ab) we use a different definition:

> Measures the time taken by the append entries RFC, to replicate the log entries of a leader agent onto its follower agent(s)

website/content/docs/upgrading/instructions/testing-experimental-raft-backend.mdx
```diff
@@ -0,0 +1,420 @@
...
+  essentially unrelated to the backend. The follower(s) with WAL enabled should
+  not be slower than the others. You can work out which follower each metric is
+  for by matching the `peer_id` label value to the server IDs listed by `consul
+  operator raft list-peers`.
+
+* `consul.raft.compactLogs` measures the time take to truncate the logs after a
+  snapshot. WAL-enabled servers should not be slower than BoltDB ones.
+
+* `consul.raft.leader.dispatchLog` measures the time spent persisting logs to
+  disk on the _leader_. It is only relevant if a WAL-enabled server becomes a
```
danielehc, line 418: In [telemetry.mdx](https://github.com/hashicorp/consul/pull/16270/files#diff-8d2b6d7521b66ab4359f6ccd7e79e955c168d5f3b9d9066b76236ebc782377ab) we use a different definition:

> Measures the time it takes for the leader to write log entries to disk.

website/content/docs/upgrading/instructions/testing-experimental-raft-backend.mdx
```diff
@@ -0,0 +1,420 @@
...
+* `consul.raft.replication.appendEntries.rpc` measures the time taken for each
+  `AppendEntries` RPC from the leader's perspective. If this is significantly
+  higher than `consul.raft.rpc.appendEntries` on the follower, it indicates a
+  known queuing issue in our raft library that will be fixed soon and is
+  essentially unrelated to the backend. The follower(s) with WAL enabled should
+  not be slower than the others. You can work out which follower each metric is
+  for by matching the `peer_id` label value to the server IDs listed by `consul
+  operator raft list-peers`.
+
+* `consul.raft.compactLogs` measures the time take to truncate the logs after a
```
danielehc, line 414: I could not find this in [telemetry.mdx](https://github.com/hashicorp/consul/pull/16270/files#diff-8d2b6d7521b66ab4359f6ccd7e79e955c168d5f3b9d9066b76236ebc782377ab).

Should it be there too?

website/content/docs/upgrading/instructions/testing-experimental-raft-backend.mdx
trujillo-adam, line 57:
```suggestion
The WAL backend stores append-only logs to disk more efficiently than BoltDB because it is a full key-value database. A BoltDB database is a single file that continuously grows and deleting
the oldest logs leaves free space in the file that must be tracked to be re-used on
future writes. WAL is a simple segmented log that can delete the oldest log
files from disk. 

With BoltDB, a burst of writes can cause the file to grow exponentially. After taking the next snapshot and the oldest logs are truncated, the file is mostly empty space. Tracking
the free space requires Consul to write extra metadata proportional to the amount of
free pages to disk with every write. After a burst of writes, write latencies
tend to increase that cause serious performance degradation to the cluster.

Consul never lets too many logs accumulate in the LogStore. Large BoltDB files are
slow as a result of its tree structure. It requires additional work for every write. But large files tend to either have a large freelist or suddenly form one after a burst of
writes. For this reason, the default options for how frequently we make a full
snapshot and truncate the logs, and for how many logs we keep around have always
been aggressively set toward keeping BoltDB small rather than using disk IO the
most efficiently.
```
I think this is too much detail for a usage doc. Can't we just say that the WAL backend is more efficient for writing logs than BoltDB because it has a truncating mechanism more suitable to rolling logs than BoltDB? We can move the details to either a blog about the change or to the release notes.

trujillo-adam (changes requested):
I've spent a few hours on this and still haven't made it all the way through. I have a few big things. First, there is too much information on the usage page; and maybe too much information in general. I wonder if you can offload a lot of this background information about our decision to create a new backend to a blog or white paper or some other channel that isn't core documentation. The second point is related to the first in that there are two content types in one page -- the conceptual information and the usage information. If we want to keep the conceptual information (first 100 lines) in the docs, we should put that into an overview page. To that end, placing this type of usage information in the Upgrade documentation seems like an odd choice. As far as I can tell, the configuration is on the Consul server agent, so I would probably put this with the Agent topics.

agent/consul/server_log_verification.go
```diff
@@ -62,12 +62,12 @@ func makeLogVerifyReportFn(logger hclog.Logger) verifier.ReportFn {
 			if r.WrittenSum > 0 && r.WrittenSum != r.ExpectedSum {
 				// The failure occurred before the follower wrote to the log so it
 				// must be corrupted in flight from the leader!
-				l2.Info("verification checksum FAILED: in-flight corruption",
```
jmurret, line 65: Not for now...but....Would it be helpful to have metrics here for failures?  Or is discovering these via aggregrated logs be enough?

jmurret (approved):
👍 Nothing blocking.  Had a minor suggestion/question about whether we want to emit metrics for failed verifications in addition to logs.

david-yu (approved):
I think the documentation for a feature like this should include as technical information about the conceptual design as possible. I'd have to respectfully disagree with @trujillo-adam about putting this in a white paper or blog, since this is a core feature that underpins the overall Consul system.

website/content/docs/upgrading/instructions/testing-experimental-raft-backend.mdx
banks: I feel like the extra specificity here is useful in terms of what it means for WAL. The one in telemetry is arguably misleading even - it's only emitted by a follower, it's not "any outstanding logs" but just whatever the next batch the leader choose to send is (and there may be multiple in flight at once) and "since the last appendEntries..." is just wrong. it's the time from recieving an append entries RPC to responding to it. If the previous one was 30s ago it won't suddenly jump up to 30s... Maybe we should just fix the `telemetry.mdx` definition?

website/content/docs/agent/config/config-files.mdx
```diff
@@ -1586,15 +1586,98 @@ Valid time units are 'ns', 'us' (or 'µs'), 'ms', 's', 'm', 'h'."
...
+    guide](/consul/docs/upgrading/instructions/testing-experimental-raft-backend)
+    to learn how to safely evaluate it for your workload.
+
+  - `disable_log_cache` ((#raft_logstore_disable_log_cache)) This allows
+    disabling of the in-memory cache of recent logs. This exists mostly for
+    performance testing purposes. In theory the log cache prevents disk reads
+    for recent logs. In practice recent logs are still in OS page cache so tend
+    not to be slow to read using either backend. We recommend leaving it enabled
+    for now as we've not measured a significant improvement in any metric by
+    disabling.
```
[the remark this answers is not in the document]
banks: Honestly I was considering not documenting this at all but I feel equally icky about having undocumented settings. It's really there just so we can actually evaluate whether it even makes a difference or not. So far the answer seems "probably not" but it's less risky to leave it alone than change the behaviour since it doesn't help to remove it either.

Do you think this would be better:
 1. Undocumented (but still present for when we want to do perf testing)
 2. Documented as "don't use this it's just for internal testing"
 3. Documented as you proposed above, less verbose but with a general "you probably don' need to touch this ever" vibe?

website/content/docs/agent/telemetry.mdx
```diff
@@ -476,7 +485,18 @@ These metrics are used to monitor the health of the Consul servers.
...
+| `consul.raft.wal.log_entry_bytes_read`        |  Counts the bytes of log entry read from segments before decoding. actual bytes read from disk might be higher as it includes headers and index entries and possible secondary reads for large entries that don't fit in buffers. | bytes | counter |
+| `consul.raft.wal.log_entry_bytes_written`     |  Counts the bytes of log entry after encoding with Codec. Actual bytes written to disk might be slightly higher as it includes headers and index entries. | bytes | counter |
+| `consul.raft.wal.segment_rotations`           |  Counts how many times we move to a new segment file. | rotations  | counter |
+| `consul.raft.wal.stable_gets`                 |  Counts how many calls to StableStore.Get or GetUint64. | calls  | counter |
+| `consul.raft.wal.stable_sets`                 |  Counts how many calls to StableStore.Set or SetUint64. | calls  | counter |
+| `consul.raft.wal.tail_truncations`            |  Counts how many log entries have been truncated from the head - i.e. the newest entries. by graphing the rate of change over time you can see individual truncate calls as spikes. | logs entries truncated | counter |
```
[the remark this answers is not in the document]
banks: heh this (in my version to) is actually wrong and opposite!

website/content/docs/upgrading/instructions/testing-experimental-raft-backend.mdx
banks: I actually had `// TODO should this be in a blog?` right up until I posted the PR on this section. My main reasoning not to was just that I think this needs to live _somewhere_ and don't think a blog will be ready on release day so wanted to land it somewhere at least initially so the content is available.

I'll go back and look at timing to see if we can make a blog happen in time.

agent/consul/server_log_verification.go
banks: We do already - they are emitted by the verifier directly - see the telemetry.mdx additions.

trujillo-adam (approved):

banks:
The functional parts of this PR were merged in https://github.com/hashicorp/consul/pull/16388

The documentation was refactored by @im2nguyen and @trujillo-adam in this new PR: https://github.com/hashicorp/consul/pull/16387 so I'm going to close this one and continue with their suggestions there.

Thanks everyone for the reviews which have helped shape one or other of those PRs!

PR #16276, opened 2023-02-15
Fix infinite recursion in inline-certificate config entry

nathancoleman:
Co-Authored-By: Andrew Stucki <3577250+andrewstucki@users.noreply.github.com>

### Description
`GetNamespace()` + `GetMeta()` were calling themselves. This PR also formats similar functions to match our config entries.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/structs/config_entry_inline_certificate.go
+++ b/agent/structs/config_entry_inline_certificate.go
@@ -29,17 +29,14 @@ type InlineCertificateConfigEntry struct {
 	RaftIndex
 }
 
-func (e *InlineCertificateConfigEntry) GetKind() string {
-	return InlineCertificate
+func (e *InlineCertificateConfigEntry) GetKind() string            { return InlineCertificate }
+func (e *InlineCertificateConfigEntry) GetName() string            { return e.Name }
+func (e *InlineCertificateConfigEntry) Normalize() error           { return nil }
+func (e *InlineCertificateConfigEntry) GetMeta() map[string]string { return e.Meta }
+func (e *InlineCertificateConfigEntry) GetEnterpriseMeta() *acl.EnterpriseMeta {
+	return &e.EnterpriseMeta
 }
-
-func (e *InlineCertificateConfigEntry) GetName() string {
-	return e.Name
-}
-
-func (e *InlineCertificateConfigEntry) Normalize() error {
-	return nil
-}
+func (e *InlineCertificateConfigEntry) GetRaftIndex() *RaftIndex { return &e.RaftIndex }
 
 func (e *InlineCertificateConfigEntry) Validate() error {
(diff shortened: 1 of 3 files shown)


nathancoleman (commented):
Personal review

andrewstucki (approved):

agent/structs/config_entry_inline_certificate.go
```diff
@@ -29,17 +29,14 @@ type InlineCertificateConfigEntry struct {
...
-	return e.Name
-}
-
-func (e *InlineCertificateConfigEntry) Normalize() error {
-	return nil
+func (e *InlineCertificateConfigEntry) GetKind() string            { return InlineCertificate }
+func (e *InlineCertificateConfigEntry) GetName() string            { return e.Name }
+func (e *InlineCertificateConfigEntry) Normalize() error           { return nil }
+func (e *InlineCertificateConfigEntry) GetMeta() map[string]string { return e.Meta }
+func (e *InlineCertificateConfigEntry) GetEnterpriseMeta() *acl.EnterpriseMeta {
```
jm96441n, line 36: this is the most minor thing, but could we put a newline above this function or move the above opening braces to be inline with this one? totally not necessary to do and purely aesthetic :joy_cat:

api/config_entry_inline_certificate.go
```diff
@@ -12,7 +12,7 @@ type InlineCertificateConfigEntry struct {
 	// Certificate is the public certificate component of an x509 key pair encoded in raw PEM format.
 	Certificate string
 	// PrivateKey is the private key component of an x509 key pair encoded in raw PEM format.
-	PrivateKey string
+	PrivateKey string `alias:"private_key"`
```
jm96441n, line 15: where does `alias` come from?

jm96441n:
LGTM! just had a minor question

api/config_entry_inline_certificate.go
andrewstucki: `alias` is used in the HCL parser we leverage for making it so you can pass the value snake-cased rather than camelized, we leverage it all over the API package, for example:

https://github.com/hashicorp/consul/blob/6599a9be1df8749431ed0f5af7a7a8772ce4281e/api/config_entry.go#L98

It's pretty typical that we prefer snake-casing to camelized values in the ConfigEntry HCL definitions

jm96441n: TIL! that's pretty nifty

t-eckert (approved):

agent/structs/config_entry_inline_certificate.go
nathancoleman: I tried moving the braces to be inline, but `gofmt` wants it this way. Given the flakiness of tests and that this is how our other config entries are today, Imma leave it like this for now

PR #16278, opened 2023-02-15
Basic gobased API gateway spinup test

sarahalsmiller:
### Description
Basic proof of concept for how to spin up an API gateway using consul test helpers + a translation TestHTTPRouteFlattening in consul-api-gateway

* go test ./...

### Links
[https://github.com/hashicorp/consul-api-gateway/blob/f78ec587946fe18eae7d4fbb297e2527cdea2299/internal/commands/server/k8s_e2e_test.go#L530] Original end to end test

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


test/integration/consul-container/libs/service/connect.go
```diff
@@ -55,6 +55,10 @@ func (g ConnectContainer) GetAddr() (string, int) {
 	return g.ip, g.appPort
 }
 
+func (g ConnectContainer) GetPort(port int) int {
+	return 0
+}
+
```
t-eckert, line 61: Is this a placeholder value?

test/integration/consul-container/libs/service/gateway.go
```diff
@@ -101,6 +102,10 @@ func (g gatewayContainer) GetAdminAddr() (string, int) {
 	return "localhost", g.adminPort
 }
 
+func (g gatewayContainer) GetPort(port int) int {
+	return g.portMappings[port]
+}
```
t-eckert, line 107: Would it make sense to modify this function to return the boolean representing if the port key exists?

```suggestion
func (g gatewayContainer) GetPort(port int) (int, bool) {
	return g.portMappings[port]
}
```

Or if a port missing is an error, 

```suggestion
func (g gatewayContainer) GetPort(port int) (int, error) {
	p, ok := g.portMappings[port]
	if !ok {
	    return 0, fmt.Errorf("port does not exist")
	}
	return p, nil
}
```

test/integration/consul-container/test/gateways/gateway_endpoint_test.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	_, _, err = client.ConfigEntries().Set(tcpRoute, nil)
+	assert.NoError(t, err)
+
+	// Create a client proxy instance with the server as an upstream
+	_, gatewayService := createServices(t, cluster, listenerPortOne)
+
+	//TODO this can and should be broken up more effectively, this is just proof of concept
+	//check statuses
```
t-eckert, line 79: I think this implementation works just fine. It's super readable and clear. If you wanted to include a circuitbreaker to avoid and infinite loop if the gateway and route are never ready, you can look at what we do in the Consul K8s Repo with the `retry` package: https://github.com/hashicorp/consul-k8s/blob/74af05612c0f53c292d3aa1727106d4e8ab2cce3/acceptance/tests/peering/peering_connect_test.go#L136

test/integration/consul-container/libs/service/connect.go
andrewstucki: I can address that since I added it 😄 -- it's to make the `ConnectContainer` adhere to the `Service` interface in this package since I added a `GetPort` method in order to retrieve the ports that we dynamically added to the punched port mappings in the proxy container that acts as the entrypoint to all the docker containers that get run (you need to map from the port you asked for it to forward to your container to the port that it's actually mapped to on the host in order to route to it).

test/integration/consul-container/libs/service/gateway.go
sarahalsmiller: Possibly, but considering that this would involve modifying other segments of the preexisting tests, I'd rather not on this PR

test/integration/consul-container/test/gateways/gateway_endpoint_test.go
sarahalsmiller: Updated to use require.Eventually instead

andrewstucki:
We'll finish off this PR after release since the test cases here have been manually vetted.

test/integration/consul-container/test/gateways/gateway_endpoint_test.go
```diff
@@ -0,0 +1,253 @@
...
+
+	return cluster
+}
+
+func createService(t *testing.T, cluster *libcluster.Cluster, serviceOpts *libservice.ServiceOpts, containerArgs []string) libservice.Service {
+	node := cluster.Agents[0]
+	client := node.GetClient()
+	// Create a service and proxy instance
+
+	// Create a service and proxy instance
```
missylbytes, line 159: This is not important but this comment is here twice

test/integration/consul-container/test/gateways/gateway_endpoint_test.go
```diff
@@ -0,0 +1,251 @@
...
+		},
+	}
+
+	_, _, err = client.ConfigEntries().Set(tcpRoute, nil)
+	assert.NoError(t, err)
+
+	// Create a client proxy instance with the server as an upstream
+	_, gatewayService := createServices(t, cluster, listenerPortOne)
+
+	//TODO this can and should be broken up more effectively, this is just proof of concept
```
sarahalsmiller, line 78:
```suggestion
```

test/integration/consul-container/libs/cluster/container.go
```diff
@@ -501,7 +505,7 @@ type containerOpts struct {
 	addtionalNetworks []string
 }
 
-func newContainerRequest(config Config, opts containerOpts) (podRequest, consulRequest testcontainers.ContainerRequest) {
+func newContainerRequest(config Config, opts containerOpts, ports ...int) (podRequest, consulRequest testcontainers.ContainerRequest) {
```
huikang, line 508: I think we can per-allocated the exposed port at https://github.com/hashicorp/consul/blob/7c9801ecf6f9e1a2e9f71ddc8b655a7fd6704fc2/test/integration/consul-container/libs/cluster/container.go#L512-L526

Therefore, we don't need to change the function and group all used ports in one place

test/integration/consul-container/libs/service/helpers.go
```diff
@@ -35,7 +35,7 @@ type ServiceOpts struct {
 }
 
 // createAndRegisterStaticServerAndSidecar register the services and launch static-server containers
-func createAndRegisterStaticServerAndSidecar(node libcluster.Agent, grpcPort int, svc *api.AgentServiceRegistration) (Service, Service, error) {
+func createAndRegisterStaticServerAndSidecar(node libcluster.Agent, grpcPort int, svc *api.AgentServiceRegistration, containerArgs ...string) (Service, Service, error) {
```
huikang, line 38: looks like containerArgs is added, but it's used nowhere (I might be missing something.)

huikang (commented):
Overall looks good. Just some non-blocking questions.

test/integration/consul-container/libs/cluster/container.go
andrewstucki: Just as an ergonomics thing, having to explicitly add in punched ports to a global map in a separate subpackage makes things quite a bit harder to track down then to have a test specify the ports it wants punched and just do something like panic if the ports have already been allocated. Especially if you're doing some more complex setups that might involve something like 3-4 listeners on a gateway, and doing that on multiple tests this quickly grows into having to explicitly add a big chunk of forwarded, non-overlapping ports to this list.

test/integration/consul-container/libs/service/helpers.go
andrewstucki: it's used in creating the services that the api gateway is going to route to in order to disambiguate between which service is getting hit when issuing an HTTP request:

```go
	serviceTwo := createService(t, cluster, &libservice.ServiceOpts{
		Name:     "service2",
		ID:       "service2",
		HTTPPort: 8081,
		GRPCPort: 8082,
	}, []string{
		"-echo-server-default-params", fmt.Sprintf("status=%d", service2ResponseCode),
	},
```

and then later on:

```go
	checkRoute(t, ip, gatewayPort, "v2", map[string]string{
		"Host": "test.foo",
		"x-v2": "v2",
	}, checkOptions{statusCode: service2ResponseCode, testName: "service2 header and path"})
```

basically setting the status code the server is going to return and then verifying that it does indeed return that so we know we're hitting that service and not some other.

huikang: Thanks for clarifying this.

test/integration/consul-container/test/gateways/http_route_test.go
```diff
@@ -0,0 +1,260 @@
...
+	p := make([]byte, n)
+	rand.Read(p)
+	return fmt.Sprintf("%s-%s", prefix, hex.EncodeToString(p))[:n]
+}
+
+func TestHTTPRouteFlattening(t *testing.T) {
+	if testing.Short() {
+		t.Skip("too slow for testing.Short")
+	}
+
```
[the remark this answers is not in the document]
sarahalsmiller: Added!

test/integration/consul-container/test/gateways/gateway_endpoint_test.go
```diff
@@ -0,0 +1,250 @@
...
+}
+
+func conditionStatusIsValue(typeName string, statusValue string, conditions []api.Condition) bool {
+	for _, c := range conditions {
+		if c.Type == typeName && c.Status == statusValue {
+			return true
+		}
+	}
+	return false
+}
```
mikemorris, line 117: We'll probably want to allow this to eventually target a specific `ResourceRef` instead of the first condition of `typeName`, but fine for now.

mikemorris (approved):
LGTM, assertions seem solid and I'm building on top of these patterns and helpers successfully in #16408

PR #16294, opened 2023-02-16
troubleshoot: fixes and updated messages

ndhanushkodi:
This PR makes a couple fixes, and updates all the output to be consistent across what's been discussed in the RFC. Note that this needs to be backported into 1.15.

Previously, it always printed `route ___ found`, even when there isn't a route. Now it only says that if it actually is a case where a route should be found.

In the below case, there are no routes and backendbloo does not have any instances. You can see there are no routes found in the below case (which is correct behavior) and the actions are displayed.

Screen Shot 2023-02-15 at 2 26 23 PM

This also capitalizes the output, and makes the error messages more consistent and adds possible actions to most error cases.

Error case (deleted deployment for a peered upstream):
Screen Shot 2023-02-16 at 12 32 33 PM
Success case (healthy upstream):
Screen Shot 2023-02-16 at 12 32 46 PM
Output for `upstreams command`:
Screen Shot 2023-02-16 at 12 33 06 PM

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/xds/validateupstream-test/validateupstream_test.go
+++ b/agent/xds/validateupstream-test/validateupstream_test.go
@@ -55,7 +55,7 @@ func TestValidateUpstreams(t *testing.T) {
 				delete(ir.Index[xdscommon.ListenerType], listenerName)
 				return ir
 			},
-			err: "no listener for upstream \"db\"",
+			err: "No listener for upstream \"db\"",
 		},
 		{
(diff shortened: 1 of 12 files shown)


hashi-derek (approved):

curtbushko (approved):
Thanks Nitya!

PR #16295, opened 2023-02-16
Inline API Gateway TLS cert code

andrewstucki:
### Description
We initially attempted to ship API Gateway certificates via SDS secrets over ADS, however, currently certificates pushed down via ADS never come out of a warming state. This acts as a stop-gap to try and ship certificates to API Gateways without requiring having a fully functioning ADS secret pushdown.

In order to implement binding multiple certificates to a gateway listener it adds some SNI behavior to the gateways, specifically the xDS generation code, when applying a single certificate leverages it for all traffic. For multiple certificates it:

1. parses all of the possible server names that a cert responds to out of the CN, DNSSans, and IPSans fields of the certificate
2. finds any overlapping server names for the certificates (envoy starts spitting crazy amounts of errors if you have overlapping server names in two different filter chains)
3. only adds filters for the server names that don't overlap and associates them with the corresponding certificate
4. constructs a default listener using the connect leaf certificate for the gateway and uses it for any other requests (including for requests for the server name that the certificates overlap on)

The above overlapping check to a default listener was done to make things deterministic rather than based off of something like which cert was created first. 

In the future I think we should talk to the upstream community about the possibility of either adding an optional flag for a certificate to be the `default` certificate, or potentially adding an optional `hostnames` field to the certificate references so that an implementation can just leverage that rather than trying to construct a non-overlapping hostnames set themselves, or both (this is a generic multi-cert problem).

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/xds/listeners_ingress.go
```diff
@@ -379,10 +412,142 @@ func makeSDSOverrideFilterChains(cfgSnap *proxycfg.ConfigSnapshot,
...
+			}
+		}
+
+		tlsContext := makeInlineTLSContextFromGatewayTLSConfig(tlsCfg, cert)
+		if multipleCerts {
+			// Configure alpn protocols on TLSContext
+			tlsContext.AlpnProtocols = getAlpnProtocols(listenerCfg.Protocol)
+		}
+
+		if err := constructChain(cert.Name, hosts, tlsContext); err != nil {
```
andrewstucki, line 514: when `hosts` is uninitialized as in the case of only a single cert above, the call to `makeSNIFilterChainMatch` in `constructChain` winds up creating no additional server_name filter matches, so the chain ends up handling all traffic for the listener.

agent/xds/listeners_ingress.go
```diff
@@ -379,10 +412,142 @@ func makeSDSOverrideFilterChains(cfgSnap *proxycfg.ConfigSnapshot,
...
+
+		// if we only have one cert, we just use it for all ingress
+		if multipleCerts {
+			// otherwise, we need an SNI per cert and to fallback to our ingress
+			// gateway certificate signed by our Consul CA
+			certHosts, err := cert.Hosts()
+			if err != nil {
+				return nil, fmt.Errorf("unable to parse hosts from x509 certificate: %v", hosts)
+			}
+			// filter out any overlapping hosts so we don't have collisions in our filter chains
```
andrewstucki, line 494: without making the hosts values unique when we have multiple filter chains, envoy and consul wind up just spewing errors about failure to apply xDS.

agent/xds/listeners_ingress.go
```diff
@@ -379,10 +412,142 @@ func makeSDSOverrideFilterChains(cfgSnap *proxycfg.ConfigSnapshot,
...
+
+		return nil
+	}
+
+	multipleCerts := len(certs) > 1
+
+	allCertHosts := map[string]struct{}{}
+	overlappingHosts := map[string]struct{}{}
+
+	if multipleCerts {
```
andrewstucki, line 461: we only need to check the collision of certificate hosts if we wind up having multiple certs, because that's the only time we have multiple filter chains

agent/xds/listeners_ingress.go
```diff
@@ -134,7 +167,7 @@ func (s *ResourceGenerator) makeIngressGatewayListeners(address string, cfgSnap
 
 			// See if there are other services that didn't have specific SNI-matching
 			// filter chains. If so add a default filterchain to serve them.
-			if len(sniFilterChains) < len(upstreams) {
+			if len(sniFilterChains) < len(upstreams) && !isAPIGatewayWithTLS {
```
andrewstucki, line 170: we bypass this as it's just creating a default filter chain which we already did inside of the new override function

agent/xds/listeners_ingress.go
```diff
@@ -87,8 +98,23 @@ func (s *ResourceGenerator) makeIngressGatewayListeners(address string, cfgSnap
 			l.FilterChains = []*envoy_listener_v3.FilterChain{
 				filterChain,
 			}
-			resources = append(resources, l)
 
+			if isAPIGatewayWithTLS {
+				l.FilterChains, err = makeInlineOverrideFilterChains(cfgSnap, cfgSnap.IngressGateway.TLSConfig, listenerKey, listenerFilterOpts{
```
andrewstucki, line 103: I should probably write a second test for TLS on tcp listeners.

andrewstucki (commented):
Self review.

Also, I'm thinking that in the future (and potentially with the upstream spec), we should talk about doing something like being able to flag a certificate as the "default" certificate that gets used if no other SNI matches, that way we don't have to use the leaf cert as the default cert.

agent/proxycfg/snapshot.go
```diff
@@ -905,9 +911,25 @@ DOMAIN_LOOP:
...
+	for _, certRef := range bound.Certificates {
+		cert, ok := c.Certificates.Get(certRef)
+		if !ok {
+			continue
+		}
+		c.ListenerCertificates[key] = append(c.ListenerCertificates[key], *cert)
+	}
+
+	return &structs.GatewayTLSConfig{
+		Enabled:       true,
```
t-eckert, line 928: Is there an edge case where the `listener.TLS.Certificates` is not empty but the `bound.Certificates` does not have a cert ref that matches in `c.Certificates`? Could we be enabling TLS without a legit cert?

agent/xds/listeners_ingress.go
```diff
@@ -25,6 +26,15 @@ func (s *ResourceGenerator) makeIngressGatewayListeners(address string, cfgSnap
...
 		}
 
+		var isAPIGatewayWithTLS bool
+		var certs []structs.InlineCertificateConfigEntry
+		if cfgSnap.APIGateway.ListenerCertificates != nil {
+			certs = cfgSnap.APIGateway.ListenerCertificates[listenerKey]
+		}
+		if certs != nil {
+			isAPIGatewayWithTLS = true
+		}
```
t-eckert, line 36: 🤷🏻 
```suggestion
		isAPIGatewayWithTLS = certs != nil
```

agent/proxycfg/snapshot.go
andrewstucki: I think that should be alright. If we wind up enabling TLS and have no certs that match in our snapshot, what should wind up happening is that we'll see no certificates in the listener generation code but will still see the `enabled` flag and accordingly fallback to the basic ingress behavior which is "use the leaf cert".

agent/xds/listeners_ingress.go
t-eckert: I think if it's worth calling out this way in a PR, you could probably safely add a comment in the code explaining the same.

t-eckert (approved):
Adding this PR to my Goodreads for the year.

agent/structs/config_entry_inline_certificate.go
```diff
@@ -64,6 +64,30 @@ func (e *InlineCertificateConfigEntry) Validate() error {
 	return nil
 }
 
+func (e *InlineCertificateConfigEntry) Hosts() ([]string, error) {
+	certificateBlock, _ := pem.Decode([]byte(e.Certificate))
+	if certificateBlock == nil {
+		return nil, errors.New("failed to parse certificate PEM")
```
jm96441n, line 70: would it make sense to use a sentinel error here so callers can use `errors.Is` to take action depending on the error type?

andrewstucki: I can do that in a follow-up. These are generic enough errors where we're pretty much only propagating this error back to the user or logging it and moving on, so I don't envision the need for ever trying to reflect on this or some wrapped version of it, but seeing as it's a static error string, I should probably make it into a variable that can be reused.

andrewstucki:
Going to merge as-is for now and circle back with the suggested changes in a follow-up so we don't block getting this in.

PR #16297, opened 2023-02-16
Fix up API Gateway go tests

andrewstucki:
### Description
This adds some optional ports you can pass in to the consul container used for initializing tests so that the ports get forwarded correctly to an upstream gateway so you can check the routing properly. It shouldn't break any existing tests, but adding this functionality to the testing libraries allows us to specify which ports to forward from our testing code.

### PR Checklist

* [x] updated test coverage
* [ ] not a security concern


sarahalsmiller (approved):
Thanks again for your eyes on this

PR #16298, opened 2023-02-17
add provider ca auth-method support for azure

eikenb:
Does the required dance with the local HTTP endpoint to get the required data for the jwt based auth setup in Azure. Keeps support for 'legacy' mode where all login data is passed on via the auth methods parameters.

### PR Checklist

* [X] updated test coverage
* [X] not a security concern

### DIFF
--- a/.changelog/16298.txt
+++ b/.changelog/16298.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ca: support Vault agent auto-auth config for Vault CA provider using Azure MSI authentication.
+```
--- a/agent/connect/ca/provider_vault_auth_azure.go
+++ b/agent/connect/ca/provider_vault_auth_azure.go
@@ -1,0 +1,153 @@
+package ca
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	"github.com/hashicorp/consul/agent/structs"
+	"github.com/hashicorp/go-cleanhttp"
+	"github.com/hashicorp/vault/sdk/helper/jsonutil"
+)
+
+func NewAzureAuthClient(authMethod *structs.VaultAuthMethod) (*VaultAuthClient, error) {
+	params := authMethod.Params
+	authClient := NewVaultAPIAuthClient(authMethod, "")
+	// check for login data already in params (for backwards compability)
+	if legacyCheck(params) {
+		return authClient, nil
+	}
+
+	role, ok := params["role"].(string)
+	if !ok || strings.TrimSpace(role) == "" {
+		return nil, fmt.Errorf("missing 'role' value")
+	}
+	resource, ok := params["resource"].(string)
+	if !ok || strings.TrimSpace(resource) == "" {
+		return nil, fmt.Errorf("missing 'resource' value")
+	}
+
+	authClient.LoginDataGen = AzureLoginDataGen
+	return authClient, nil
+}
+
+var ( // use variables so we can change these in tests
+	instanceEndpoint = "http://169.254.169.254/metadata/instance"
+	identityEndpoint = "http://169.254.169.254/metadata/identity/oauth2/token"
+	// minimum version 2018-02-01 needed for identity metadata
+	apiVersion = "2018-02-01"
+)
+
+type instanceData struct {
+	Compute
+}
+type Compute struct {
+	Name              string
+	ResourceGroupName string
+	SubscriptionID    string
+	VMScaleSetName    string
+}
+type identityData struct {
+	AccessToken string `json:"access_token"`
+}
+
+func AzureLoginDataGen(authMethod *structs.VaultAuthMethod) (map[string]any, error) {
+	params := authMethod.Params
+	role := params["role"].(string)
+	metaConf := map[string]string{
+		"role":     role,
+		"resource": params["resource"].(string),
+	}
+	if objectID, ok := params["object_id"].(string); ok {
+		metaConf["object_id"] = objectID
+	}
+	if clientID, ok := params["client_id"].(string); ok {
+		metaConf["client_id"] = clientID
+	}
+
+	// Fetch instance data
+	var instance instanceData
+	body, err := getMetadataInfo(instanceEndpoint, nil)
+	if err != nil {
+		return nil, err
+	}
+	err = jsonutil.DecodeJSON(body, &instance)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing instance metadata response: %w", err)
+	}
+
+	// Fetch JWT
+	var identity identityData
+	body, err = getMetadataInfo(identityEndpoint, metaConf)
+	if err != nil {
+		return nil, err
+	}
+	err = jsonutil.DecodeJSON(body, &identity)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing instance metadata response: %w", err)
+	}
+
+	data := map[string]interface{}{
+		"role":                role,
+		"vm_name":             instance.Compute.Name,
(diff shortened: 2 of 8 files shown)


agent/connect/ca/provider_vault_auth_azure.go
```diff
@@ -0,0 +1,153 @@
+package ca
+
+import (
+	"fmt"
+	"io/ioutil"
```
kisunji, line 5: `ioutil` has been deprecated so you'll probably want `io` here

agent/connect/ca/provider_vault_auth_test.go
```diff
@@ -301,3 +305,126 @@ func TestVaultCAProvider_AWSLoginDataGenerator(t *testing.T) {
...
+			switch url {
+			case "/metadata/instance":
+				w.Write(instanceJSON)
+			case "/metadata/identity/oauth2/token":
+				w.Write(identityJSON)
+			default:
+				t.Errorf("unexpected testing URL: %s", url)
+			}
+		}))
+	_ = msi
```
kisunji, line 331: Should this be removed?

agent/connect/ca/provider_vault_auth_azure.go
```diff
@@ -0,0 +1,153 @@
...
+var ( // use variables so we can change these in tests
+	instanceEndpoint = "http://169.254.169.254/metadata/instance"
+	identityEndpoint = "http://169.254.169.254/metadata/identity/oauth2/token"
+	// minimum version 2018-02-01 needed for identity metadata
+	apiVersion = "2018-02-01"
+)
+
+type instanceData struct {
+	Compute
+}
```
kisunji, line 44: What is the reason for embedding this instead of using `Compute` type directly?

agent/connect/ca/provider_vault_auth_azure.go
```diff
@@ -0,0 +1,153 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	"github.com/hashicorp/consul/agent/structs"
+	"github.com/hashicorp/go-cleanhttp"
+	"github.com/hashicorp/vault/sdk/helper/jsonutil"
```
kisunji, line 11: Any particular reason for using vault's `jsonutil`? Consul has its own `DecodeJSON` too I believe which is a small wrapper around `json` from stdlib.

agent/connect/ca/provider_vault_auth_azure.go
```diff
@@ -0,0 +1,153 @@
...
+		return nil, fmt.Errorf("missing 'resource' value")
+	}
+
+	authClient.LoginDataGen = AzureLoginDataGen
+	return authClient, nil
+}
+
+var ( // use variables so we can change these in tests
+	instanceEndpoint = "http://169.254.169.254/metadata/instance"
+	identityEndpoint = "http://169.254.169.254/metadata/identity/oauth2/token"
```
kisunji, line 37: Where are these IP addresses from?

.changelog/16298.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
```
kisunji, line 1: Should this be a feature instead?

agent/connect/ca/provider_vault_auth_azure.go
eikenb: They are Azure's standard metadata endpoints. AWS uses the same IP addresses. The 169.254.0.0/16 IP block is reserved for private, "link-local" addresses. I'm not 100% on why they standardized on 169.254.169.254.. best guess is that it's easy to remember.

.changelog/16298.txt
eikenb: This was what @cthain used in #15970, adding the same functionality for AWS/GCP. I'm guessing the idea is that it is improving the functionality from the hack support (that they could hardcode the /login data) to real support.

agent/connect/ca/provider_vault_auth_azure.go
rboyer: IIRC the big 3 major clouds all use `169.254.169.254` as the ip address of the metadata server.

agent/connect/ca/provider_vault_auth_test.go
eikenb: Missed that, thanks.

agent/connect/ca/provider_vault_auth_azure.go
eikenb: First I've seen of these that it wasn't os.ReadFile. Thanks.

agent/connect/ca/provider_vault_auth_azure.go
eikenb: The JSON data we're decoding has the wrapping structure with other data, we just only care about the Compute data part.

agent/connect/ca/provider_vault_auth_azure.go
eikenb: I had based this off code from the vault agent which used that. We already included that dependency and I figured, as we were hitting a Vault endpoint, it made sense to keep using it.

.changelog/16298.txt
cthain: Yes, that was my thought process in choosing `improvement` instead of `feature`.

agent/connect/ca/provider_vault_auth_azure.go
```diff
@@ -0,0 +1,153 @@
...
+		"subscription_id":     instance.Compute.SubscriptionID,
+		"jwt":                 identity.AccessToken,
+	}
+
+	return data, nil
+}
+
+// Checks to see if all the parameters needed to login have been hardcoded in the
+// auth-method's config Parameters field.
+func legacyCheck(params map[string]any) bool {
```
cthain, line 105: I really like the func name `legacyCheck`.. there is already a function called `containsVaultLoginParams` that has some overlap with this method.. What do you think about consolidating these two functions?

cthain (approved):
Good stuff John :+1:

kisunji (approved):

agent/connect/ca/provider_vault_auth_azure.go
eikenb: Right, it is basically the same function just passing in the list of keys. I'll replace containsVaultLoginParams with legacyCheck (adding the expectedKeys as an arg) and update the AWS plugin to us it. Thanks.

eikenb: Noticed they have a key difference in testing... the containsVaultLoginParams returns true when it finds any of the hardcoded /login values where legacyCheck only returns true if all of them are present. In the case of Azure they are all required for it to work, were AWS/GCP different or did you go with the idea that if they had any of the /login fields it meant they were using that path and can return early?

cthain: Yeah, exactly.. My thought was that if any of the `/login` args are present, then the user has most likely configured the CA provider for a direct login.. I think that it's reasonable for the func to check for all required args before returning true since most folks will probably want to use new auto auth config.

eikenb: Thanks @cthain, I'll go with all fields required then and make those changes.

eikenb: Started looking at updating AWS to this and it has multiple different ways of calling /login with different sets of parameters, thus there is no 1 set of required parameters to check.

So I'm flip-flopping on this and will make it use 'legacy mode' if it detects **any** of the 'legacy' or hard coded /login fields. Going to re-review my uses to make sure none of them have fields that overlap with non-/login-hack uses.

PR #16300, opened 2023-02-17
Fix hostname alignment checks for HTTPRoutes

andrewstucki:
### Description
This adds a missing hostname check between listeners and HTTPRoutes. Without it we do no verification of the alignment between route and Listener hostnames and the listener hostnames are essentially unused. In the upstream spec, only http routes that have hostnames matching the listener hostname (or no hostnames specified) can be bound to a listener. We accomplish this by filtering out routes (both in the xDS code and in our controller) based on the listener hostname and then checking to see if there are any valid hosts to bind to.

It's a stacked PR on top of the inline certs PR (#16295), so that should get merged first.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


missylbytes (approved):
LGTM

sarahalsmiller (approved):
Looks good to me. Thanks for knocking this out

PR #16302, opened 2023-02-17
Protobuf Refactoring for Multi-Module Cleanliness

mkeeler:
## Description

This commit includes the following:

* Moves all packages that were within proto/ to proto/private 
* Rewrites imports to account for the packages being moved 
* Adds in buf.work.yaml to enable buf workspaces
* Names the proto-public buf module so that we can override the Go package imports within proto/buf.yaml 
* Bumps the buf version dependency to 1.14.0 (I was trying out the version to see if it would get around an issue - it didn't but it also doesn't break things and it seemed best to keep up with the toolchain changes) 

Why:

* In the future we will need to consume other protobuf dependencies such as the Google HTTP annotations for openapi generation or grpc-gateway usage. 
* There were some recent changes to have our own ratelimiting annotations. 
* The two combined were not working when I was trying to use them together (attempting to rebase another branch) 
* Buf workspaces should be the solution to the problem 
* Buf workspaces means that each module will have generated Go code that embeds proto file names relative to the proto dir and not the top level repo root. 
* This resulted in proto file name conflicts in the Go global protobuf type registry. 
* The solution to that was to add in a private/ directory into the path within the proto/ directory. 
* That then required rewriting all the imports.

Is this safe?

* AFAICT yes
* The gRPC wire protocol doesn't seem to care about the proto file names (although the Go grpc code does tack on the proto file name as Metadata in the ServiceDesc) Other than imports, there were no changes to any generated code as a result of this.

* This PR is about preparedness and protobuf organization. It is intended to produce no functional changes so no additional tests have been added. If all tests pass then things should be fine.


agent/grpc-middleware/rate_limit_mappings.gen.go
```diff
@@ -11,17 +11,5 @@ var rpcRateLimitSpecs = map[string]rate.OperationType{
 	"/hashicorp.consul.dataplane.DataplaneService/GetEnvoyBootstrapParams":       rate.OperationTypeRead,
 	"/hashicorp.consul.dataplane.DataplaneService/GetSupportedDataplaneFeatures": rate.OperationTypeRead,
 	"/hashicorp.consul.dns.DNSService/Query":                                     rate.OperationTypeRead,
-	"/hashicorp.consul.internal.operator.OperatorService/TransferLeader":         rate.OperationTypeExempt,
```
rboyer, line 14: Should these be here?

mkeeler: Yes, I had fixed that but somehow the fix didn't make it over to oss

rboyer (approved):

PR #16316, opened 2023-02-17
Normalize all API Gateway references

andrewstucki:
### Description
This makes sure we normalize all enterprise metadata on write across our config entries. Enterprise tests were failing with the new TLS tests due to having a mismatch between what was in a `ResourceReference` and what was on the ConfigEntry itself as returned from the gRPC stream, as it got normalized elsewhere.

Tested and validated on enterprise.

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/agent/structs/config_entry_routes.go
+++ b/agent/structs/config_entry_routes.go
@@ -79,6 +79,7 @@ func (e *HTTPRouteConfigEntry) Normalize() error {
 	for i, parent := range e.Parents {
 		if parent.Kind == "" {
 			parent.Kind = APIGateway
+			parent.EnterpriseMeta.Normalize()
 			e.Parents[i] = parent
 		}
 	}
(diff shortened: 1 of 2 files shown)


agent/structs/config_entry_routes.go
```diff
@@ -87,12 +88,22 @@ func (e *HTTPRouteConfigEntry) Normalize() error {
...
 	}
 
 	return nil
 }
 
+func normalizeHTTPService(service HTTPService) HTTPService {
+	service.EnterpriseMeta.Normalize()
+
+	return service
+}
```
mikemorris, line 105: Any reason this follows a different pattern than everywhere else? Aside from that nit, LGTM assuming that `EnterpriseMeta.Normalize()` mutates the struct.

mikemorris (approved):

agent/structs/config_entry_routes.go
andrewstucki: It does mutate the struct -- verified it and double checked that normalize takes a pointer receiver 😅.

I just did this pattern since this is how the rest of the HTTPRoute normalization/validation code is done since there are substantially more subsections to cover than the other resources.

t-eckert (approved):

PR #16263, opened 2023-02-14
[OSS] security: update go to 1.20.1

DanStough:
### Description
This resolves a vulnerabilities [CVE-2022-41724](https://go.dev/issue/58001) in `crypto/tls` and [CVE-2022-41723](https://go.dev/issue/57855) in `net/http`.

### Testing & Reproduction steps
* CI and `make dev`

PR #16323, opened 2023-02-17
fix: revert go mod compat for sdk,api to 1.19

DanStough:
### Description
This reverts changes from #16263 so that consumers of API/SDK modules can still use older, supported versions of Go with those modules.

### DIFF
--- a/sdk/freeport/freeport.go
+++ b/sdk/freeport/freeport.go
@@ -76,6 +76,9 @@ var (
 	// total is the total number of available ports in the block for use.
 	total int
 
+	// seededRand is a random generator that is pre-seeded from the current time.
+	seededRand *rand.Rand
+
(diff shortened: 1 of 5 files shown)


sdk/freeport/freeport.go
```diff
@@ -114,6 +117,7 @@ func initialize() {
 		panic("freeport: block size too big or too many blocks requested")
 	}
 
+	seededRand = rand.New(rand.NewSource(time.Now().UnixNano())) // This is compatible with go 1.19 but unnecessary in >= go1.20
```
DanStough, line 120: Used to be `rand.Seed(time.Now().UnixNano())`, seeded globally

hashi-derek (approved):

rboyer (approved):

PR #16340, opened 2023-02-21
[API Gateway] Validate listener name is not empty

andrewstucki:
### Description
This adds the same validation rules that the API Gateway spec has for listener names. Previously we were allowing for an empty listener name which made it impossible to distinguish between wanting to bind to all listeners on a gateway or just the listener without a name.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/structs/config_entry_gateways.go
+++ b/agent/structs/config_entry_gateways.go
@@ -867,9 +867,8 @@ const (
 
 // APIGatewayListener represents an individual listener for an APIGateway
 type APIGatewayListener struct {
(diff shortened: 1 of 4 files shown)


t-eckert (approved):

nathancoleman:
LGTM but will want to update the docstrings in some places such as

https://github.com/hashicorp/consul/blob/13da1a52853038902c84f387e350209b8709cf38/api/config_entry_gateways.go#L271-L274

nathancoleman (approved):

PR #16342, opened 2023-02-21
Rate limiter/add ip prefix

dhiaayachi:
### Description

This change add the ability in multilimiter to associate a limiter with a config in the longest prefix tree. That way if we have a config under `127.0.0` the limiter for ip range `127.0.0.0/24` will use that config. 

Added a test to cover the use case above and another where we have 2 different prefixes (`127.0`(c1) and `127.0.0`(c2)) and we expect `127.0.0.1` to use c2 while `127.0.1.1` to use c2.

### PR Checklist

* [x] updated test coverage
* [ ] not a security concern


wangxinyi7 (approved):
LGTM!

PR #16347, opened 2023-02-21
[API Gateway] Various fixes for Config Entry fields

andrewstucki:
### Description

Dropping Weight from TCPRoute as it's currently unused and unsupported. Swapping `URLRewrite` in `HTTPFilters` to be a single field rather than an array to go along with the upstream spec specifying:

> At most one of these filters may be used on a Route rule

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


agent/consul/discoverychain/gateway_httproute.go
```diff
@@ -77,15 +77,14 @@ func httpRouteToDiscoveryChain(route structs.HTTPRouteConfigEntry) (*structs.Ser
 
 	for idx, rule := range route.Rules {
 		modifier := httpRouteFiltersToServiceRouteHeaderModifier(rule.Filters.Headers)
-		prefixRewrite := httpRouteFiltersToDestinationPrefixRewrite(rule.Filters.URLRewrites)
+		prefixRewrite := httpRouteFiltersToDestinationPrefixRewrite(rule.Filters.URLRewrite)
 
 		var destination structs.ServiceRouteDestination
 		if len(rule.Services) == 1 {
-			// TODO open question: is there a use case where someone might want to set the rewrite to ""?
```
nathancoleman, line 84: 🎉

t-eckert (approved):
Super fast turnaround. I'll update the docs to make sure they match. This will be much better for users.

agent/consul/discoverychain/gateway_httproute.go
```diff
@@ -176,11 +175,9 @@ func httpRouteToDiscoveryChain(route structs.HTTPRouteConfigEntry) (*structs.Ser
...
-func httpRouteFiltersToDestinationPrefixRewrite(rewrites []structs.URLRewrite) string {
-	for _, rewrite := range rewrites {
-		if rewrite.Path != "" {
-			return rewrite.Path
-		}
+func httpRouteFiltersToDestinationPrefixRewrite(rewrite *structs.URLRewrite) string {
+	if rewrite != nil && rewrite.Path != "" {
+		return rewrite.Path
 	}
 	return ""
```
nathancoleman, line 182:
```suggestion
	if rewrite == nil {
		return ""
	}
	return rewrite.Path
```

agent/structs/config_entry_routes.go
```diff
@@ -554,9 +552,6 @@ func (e *TCPRouteConfigEntry) CanWrite(authz acl.Authorizer) error {
 // TCPService is a service reference for a TCPRoute
 type TCPService struct {
 	Name string
-	// Weight specifies the proportion of requests forwarded to the referenced service.
-	// This is computed as weight/(sum of all weights in the list of services).
-	Weight int
```
nathancoleman, line 559: 🎉

nathancoleman (approved):

mikemorris:
Did we want to swap the `Redirect` HTTP request filter to a string pointer too? https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRequestRedirectFilter

PR #16350, opened 2023-02-22
[API Gateway] Fix targeting service splitters in HTTPRoutes

andrewstucki:
### Description

Previously when synthesizing an ingress discovery chain from API Gateway primitives, we weren't properly fixing up the discovery chain to allow for targeting service splitters, which we support today in our Kubernetes solution. This fixes this, allowing the use of virtual splitters in a service target inside of an HTTPRoute.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/consul/discoverychain/gateway.go
+++ b/agent/consul/discoverychain/gateway.go
@@ -5,6 +5,7 @@ import (
 	"hash/crc32"
 	"sort"
 	"strconv"
+	"strings"
 
 	"github.com/hashicorp/consul/agent/configentry"
 	"github.com/hashicorp/consul/agent/structs"
@@ -126,6 +127,23 @@ func (l *GatewayChainSynthesizer) Synthesize(chains ...*structs.CompiledDiscover
 		if err != nil {
 			return nil, nil, err
 		}
+
+		// fix up the nodes for the terminal targets to either be a splitter or resolver if there is no splitter present
+		for name, node := range compiled.Nodes {
+			switch node.Type {
+			// we should only have these two types
+			case structs.DiscoveryGraphNodeTypeRouter:
+				for i, route := range node.Routes {
+					node.Routes[i].NextNode = targetForResolverNode(route.NextNode, chains)
+				}
+			case structs.DiscoveryGraphNodeTypeSplitter:
+				for i, split := range node.Splits {
+					node.Splits[i].NextNode = targetForResolverNode(split.NextNode, chains)
+				}
+			}
+			compiled.Nodes[name] = node
+		}
+
 		for _, c := range chains {
 			for id, target := range c.Targets {
 				compiled.Targets[id] = target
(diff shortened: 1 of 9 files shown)


agent/consul/discoverychain/gateway.go
```diff
@@ -177,6 +195,27 @@ func (l *GatewayChainSynthesizer) consolidateHTTPRoutes() []structs.HTTPRouteCon
 	return routes
 }
 
+func targetForResolverNode(nodeName string, chains []*structs.CompiledDiscoveryChain) string {
+	resolverPrefix := structs.DiscoveryGraphNodeTypeResolver + ":"
+	splitterPrefix := structs.DiscoveryGraphNodeTypeSplitter + ":"
```
t-eckert, line 200: Not blocking: Would these make more sense as constants defined at the top of the file?

t-eckert (approved):
Excellent

test/integration/connect/envoy/case-api-gateway-http-splitter-targets/verify.bats
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bats
+
+load helpers
+
+@test "api gateway proxy admin is up on :20000" {
+  retry_default curl -f -s localhost:20000/stats -o /dev/null
+}
+
+@test "api gateway should have be accepted and not conflicted" {
```
nathancoleman, line 9:
```suggestion
@test "api gateway should be accepted and not conflicted" {
```

test/integration/connect/envoy/case-api-gateway-http-splitter-targets/setup.sh
```diff
@@ -0,0 +1,80 @@
...
+        name = "splitter-one"
+      }
+    ]
+  }
+]
+parents = [
+  {
+    name = "api-gateway"
+    sectionName = "listener-one"
+  }
```
nathancoleman, line 41: Does `kind` default to `"api-gateway"` here?

My belief was "no" coming into this PR, based on
https://github.com/hashicorp/consul/blob/15d2684ecc5c97036d946c8e8f4578446a56b3e8/agent/structs/config_entry_routes.go#L134-L142

andrewstucki: It get's normalized here if unset: https://github.com/hashicorp/consul/blob/15d2684ecc5c97036d946c8e8f4578446a56b3e8/agent/structs/config_entry_routes.go#L80-L84

agent/consul/discoverychain/gateway.go
andrewstucki: I'm ok with this for now -- most of this code is a temporary holdover as we work towards native xDS for API Gateways anyway, so it'll be dropped soon.

nathancoleman (approved):

nathancoleman:
Having some trouble getting the tests to pass, though they seem to be failing for unrelated reasons

andrewstucki:
Yep, the unit tests have been extremely flaky for the past week and it's pretty much entirely unrelated to any code we've been committing.

PR #16396, opened 2023-02-23
Fix various flaky tests

kisunji:
Fixes tests that have high rates of failure in our CI pipeline:
* `TestStreamResources_Server_DisconnectsOnHeartbeatTimeout/stream_is_disconnected_due_to_heartbeat_timeout`
* `TestDebugCommand_DebugDisabled`
* `TestServer_DeltaAggregatedResources_v3_ACLEnforcement` (added retries but not 100% confident in fix)

### DIFF
--- a/agent/grpc-external/services/peerstream/stream_test.go
+++ b/agent/grpc-external/services/peerstream/stream_test.go
@@ -1252,8 +1252,8 @@ func TestStreamResources_Server_DisconnectsOnHeartbeatTimeout(t *testing.T) {
 	})
 
 	testutil.RunStep(t, "stream is disconnected due to heartbeat timeout", func(t *testing.T) {
-		disconnectTime := ptr(it.FutureNow(1))
 		retry.Run(t, func(r *retry.R) {
+			disconnectTime := ptr(it.StaticNow())
 			status, ok := srv.StreamStatus(testPeerID)
 			require.True(r, ok)
 			require.False(r, status.Connected)
@@ -1423,7 +1423,7 @@ func makeClient(t *testing.T, srv *testServer, peerID string) *MockClient {
 		},
 	}))
 
-	// Receive a services and roots subscription request pair from server
+	// Receive ExportedService, ExportedServiceList, and PeeringTrustBundle subscription requests from server
 	receivedSub1, err := client.Recv()
 	require.NoError(t, err)
 	receivedSub2, err := client.Recv()
--- a/agent/xds/delta_test.go
+++ b/agent/xds/delta_test.go
@@ -10,7 +10,7 @@ import (
 
 	"github.com/armon/go-metrics"
 	envoy_discovery_v3 "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
-	"github.com/hashicorp/consul/api"
+
 	"github.com/stretchr/testify/require"
 	rpcstatus "google.golang.org/genproto/googleapis/rpc/status"
 	"google.golang.org/grpc/codes"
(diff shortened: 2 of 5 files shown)


agent/grpc-external/services/peerstream/stream_test.go
```diff
@@ -1252,8 +1252,8 @@ func TestStreamResources_Server_DisconnectsOnHeartbeatTimeout(t *testing.T) {
 	})
 
 	testutil.RunStep(t, "stream is disconnected due to heartbeat timeout", func(t *testing.T) {
-		disconnectTime := ptr(it.FutureNow(1))
```
kisunji, line 1255: incrementalTime `it` was being modified in a goroutine so there was a data race, leading `disconnectTime` to be incorrect. I moved the call to be local to each retry loop so that it would fetch the latest "now" value instead of predicting it would be 1 second in the future.

agent/xds/delta_test.go
```diff
@@ -1095,13 +1097,15 @@ func TestServer_DeltaAggregatedResources_v3_ACLEnforcement(t *testing.T) {
...
-				val, ok := item.Gauges["consul.xds.test.xds.server.streamsUnauthenticated"]
-				require.True(t, ok)
-				require.Equal(t, float32(1), val.Value)
+				retry.Run(t, func(r *retry.R) {
+					data := scenario.sink.Data()
+					require.Len(r, data, 1)
+
+					item := data[0]
+					val, ok := item.Gauges["consul.xds.test.xds.server.streamsUnauthenticated"]
+					require.True(r, ok)
```
kisunji, line 1106: This assertion was failing randomly; I'm not 100% confident this will fix the issue but since `scenario.sink.Data` is timing-based I'm hoping retries will help here.

command/debug/debug.go
```diff
@@ -270,7 +270,8 @@ func (c *cmd) prepare() (version string, err error) {
 	// If none are specified we will collect information from
 	// all by default
 	if len(c.capture) == 0 {
-		c.capture = defaultTargets
+		c.capture = make([]string, len(defaultTargets))
```
kisunji, line 273: very fun slice debugging exercise where the code block below was mutating `defaultTargets` which affected tests running in parallel

skpratt (approved):
Looks good! Thanks for doing this! 

Will be curious to see whether a deeper issue surfaces for TestServer_DeltaAggregatedResources_v3_ACLEnforcement. But retries will probably help for now.

agent/xds/delta_test.go
NiniOak: When reading this code earlier, I assumed this assertion was failing randomly because this key `consul.xds.test.xds.server.streamsUnauthenticated` isn't always found in the data from #L1101. 
- #L1102 validates that data exists and the value isn't empty; which leads me to assume that the key requested isn't found in #L1104
- Is there a precise set of data that we use for testing that will always return this key/value?

NiniOak (approved):

agent/xds/delta_test.go
kisunji: The gauge should be incremented in a normal codepath that is taken during the test: [1](https://github.com/hashicorp/consul/blob/72a73661c95800c1e3641af5d88555b4b019b1c9/agent/xds/server.go#L136), [2](https://github.com/hashicorp/consul/blob/b3ddd4d24efd42cc10db34e99400f44c30d7ec37/agent/xds/delta.go#L50)

I think the tests could be flaking because either:
1. There is some timing issue with the in-memory metrics sink
2. There is a data race between goroutines within the test

I didn't spend too much time looking into it but added a bunch of fixes for possible culprits

agent/xds/xds_protocol_helpers_test.go
```diff
@@ -166,9 +166,6 @@ func newTestServerDeltaScenario(
 ) *testServerScenario {
 	mgr := newTestManager(t)
 	envoy := NewTestEnvoy(t, proxyID, token)
-	t.Cleanup(func() {
-		envoy.Close()
-	})
```
kisunji, line 171: cleanups and defers are run in LIFO order. I wanted to ensure `envoy.Close()` got called before the global metrics sink was reset

PR #16408, opened 2023-02-23
gateways: add e2e test for API Gateway HTTPRoute ParentRef change

mikemorris:
### Description

Adds an e2e test for switching the `ParentRef` of an `HTTPRoute` from one API Gateway to another.

### Links

- [x] Depends on #16278, should be rebased after that merges (not targeting that branch and relying on the auto-retarget because CI tests only seem to run for PRs targeting `main`)

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/test/integration/consul-container/test/gateways/gateway_endpoint_test.go
+++ b/test/integration/consul-container/test/gateways/gateway_endpoint_test.go
@@ -248,3 +248,36 @@ func checkRoute(t *testing.T, ip string, port int, path string, headers map[stri
 
 	})
 }
+
+func checkRouteError(t *testing.T, ip string, port int, path string, headers map[string]string, expected string) {
+	failer := func() *retry.Timer {
+		return &retry.Timer{Timeout: time.Second * 60, Wait: time.Second * 60}
+	}
+
+	client := cleanhttp.DefaultClient()
+	url := fmt.Sprintf("http://%s:%d", ip, port)
+
+	if path != "" {
+		url += "/" + path
+	}
+
+	retry.RunWith(failer(), t, func(r *retry.R) {
+		t.Logf("making call to %s", url)
+		req, err := http.NewRequest("GET", url, nil)
+		assert.NoError(t, err)
+
+		for k, v := range headers {
+			req.Header.Set(k, v)
+
+			if k == "Host" {
+				req.Host = v
+			}
+		}
+		_, err = client.Do(req)
+		assert.Error(t, err)
+
+		if expected != "" {
+			assert.ErrorContains(t, err, expected)
+		}
+	})
+}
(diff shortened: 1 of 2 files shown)


test/integration/consul-container/test/gateways/http_route_test.go
```diff
@@ -256,3 +256,213 @@ func TestHTTPRouteFlattening(t *testing.T) {
 	}, checkOptions{debug: false, statusCode: service1ResponseCode, testName: "service1, v2 path with v2 hostname"})
 
 }
+
+func TestHTTPRouteParentRefChange(t *testing.T) {
+	if testing.Short() {
+		t.Skip("too slow for testing.Short")
+	}
+
```
sarahalsmiller, line 264: you'll probably want to add a t.Parallel() in here.

mikemorris:
```suggestion
	t.Parallel()
```

test/integration/consul-container/test/gateways/http_route_test.go
```diff
@@ -256,3 +256,214 @@ func TestHTTPRouteFlattening(t *testing.T) {
...
+func TestHTTPRouteParentRefChange(t *testing.T) {
+	if testing.Short() {
+		t.Skip("too slow for testing.Short")
+	}
+	t.Parallel()
+
+	// infrastructure set up
+	address := "localhost"
+
+	// TODO: could/should we somehow pick these randomly?
```
t-eckert, line 269: I don't know if it's worth it. I think just keeping these hard-coded is fine.

test/integration/consul-container/test/gateways/gateway_endpoint_test.go
```diff
@@ -248,3 +248,36 @@ func checkRoute(t *testing.T, ip string, port int, path string, headers map[stri
...
+			if k == "Host" {
+				req.Host = v
+			}
+		}
+		_, err = client.Do(req)
+		assert.Error(t, err)
+
+		if expected != "" {
+			assert.ErrorContains(t, err, expected)
+		}
```
andrewstucki, line 281: both expected errors are empty in the functions that call this, I'm assuming that's because the error value is changing depending on how quickly the connection is terminating or something?

t-eckert (approved):

test/integration/consul-container/test/gateways/gateway_endpoint_test.go
mikemorris: Yeaaaa, I think keeping this capability in the helper function makes sense, but I had to move away from checking for EOF due to differences in CI in https://github.com/hashicorp/consul/pull/16408/commits/a596dbf175d7446606b188c1531f3b8801770449

test/integration/consul-container/test/gateways/http_route_test.go
mikemorris:
```suggestion
```

test/integration/consul-container/test/gateways/gateway_endpoint_test.go
andrewstucki: That's fine then, I think so long as we keep the positive checks first (i.e. the gateway route is routable), then we should be fine here, mainly just don't want this to be a false positive because of something like the gateway not having been configured yet.

Issue #16431, opened 2023-02-26
consul acl token read -self broken in 1.15.0


dani:
#### Overview of the Issue

In my scripts, I check if I already have a valid consul token, to ask vault for a new one when it's not the case. For this, I used to run
```
consul acl token read -self
```
Since I've upgraded to 1.15.0, this command fails and ask for an accessor-id. The help output (consul acl token read help) still mention -self as a valid option, to use our current http token instead of providing an accessor-id

david-yu:
Thanks for reporting @dani, this should come in the Consul 1.15.1 patch release in a few weeks.

PR #16445, opened 2023-02-27
cli: ensure acl token read -self works

rboyer:
### Description

Fixes a regression in #16044

The `consul acl token read -self` cli command should not require an `-accessor-id` because typically the persona invoking this would not already know the accessor id of their own token.

### DIFF
--- a/.changelog/16445.txt
+++ b/.changelog/16445.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cli: ensure acl token read -self works
+```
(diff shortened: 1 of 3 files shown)


skpratt (approved):
Thank you for adding the test for that flag! As soon as I saw the regression report, I figured it was something like me [dropping a flag check in #16044](https://github.com/hashicorp/consul/pull/16044/files#diff-cef6aa018262ee2d7cf82cb6bd1931a6c7bec9b6e6d770dd562a1db42f616a83L64).  :(

david-yu:
Linked this PR with https://github.com/hashicorp/consul/issues/16431

cthain (approved):
:+1:

PR #16498, opened 2023-03-02
Fix issue where terminating gateway service resolvers weren't properly cleaned up

andrewstucki:
### Description

This fixes an issue where terminating gateways weren't properly cleaning up service resolvers attached to their external services. When a resolver was deleted, the casting guard was keeping around the old `ServiceResolver` value due to a failed `nil` cast.

However, there are some questions that I do have about whether or not `ServiceResolver` subsets are ever supposed to work with an external service accessed through a terminating gateway, as they currently do not -- when a local proxy attempts to resolve the endpoint for an upstream that is an external service that has a `ServiceResolver` with subsets it winds up attempting to watch the subset upstream directly which is not returning any address (verses without a subset, returning the terminating gateway address). Either way though, that would be a different bug.

Run the following bash script:
```bash
#!/bin/bash

cleanup() {
  echo "shutting down upstreams"
}

trap 'trap " " SIGTERM; kill 0; wait; cleanup' SIGINT SIGTERM

cat << EOF | ./consul config write -
Kind      = "proxy-defaults"
Name      = "global"
Config {
  protocol = "http"
}
EOF

echo "Writing terminating gateway config entry"
cat << EOF | ./consul config write -
Kind = "terminating-gateway"
Name = "gateway"

Services = [
  {
    Name = "external"
  }
]
EOF

cat << EOF > /tmp/external.json
{
  "Node": "hashicorp",
  "Address": "127.0.0.1",
  "NodeMeta": {
    "external-node": "true",
    "external-probe": "true"
  },
  "Service": {
    "ID": "external-v1",
    "Service": "external",
    "Port": 9877,
    "Meta": {
      "version": "v1"
    }
  }
}
EOF
curl --request PUT --data @/tmp/external.json localhost:8500/v1/catalog/register

cat << EOF > /tmp/external.json
{
  "Node": "hashicorp",
  "Address": "127.0.0.1",
  "NodeMeta": {
    "external-node": "true",
    "external-probe": "true"
  },
  "Service": {
    "ID": "external-v2",
    "Service": "external",
    "Port": 9878,
    "Meta": {
      "version": "v2"
    }
  }
}
EOF
curl --request PUT --data @/tmp/external.json localhost:8500/v1/catalog/register

echo "Writing resolver config entry"
cat << EOF | ./consul config write -
Kind = "service-resolver"
Name = "external"
DefaultSubset = "v1"
Subsets = {  
  v1 = {    
    Filter = "Service.Meta.version == v1"  
  } 
  v2 = {    
    Filter = "Service.Meta.version == v2"  
  }
}
EOF

echo "Running terminating gateway"
./consul connect envoy -gateway terminating -register -service gateway -proxy-id gateway -- -l trace  &

wait
```

Before the fix:

```zsh
➜  consul git:(terminating-gateway-resolvers) ✗ curl -s http://localhost:19000/clusters | grep external | cut -d':' -f1 | uniq | sort
external.default.dc1.internal.4e38b938-3a3e-7b67-6624-a8dae1058918.consul
v1.external.default.dc1.internal.4e38b938-3a3e-7b67-6624-a8dae1058918.consul
v2.external.default.dc1.internal.4e38b938-3a3e-7b67-6624-a8dae1058918.consul
➜  consul git:(terminating-gateway-resolvers) ✗ ./consul config delete -kind service-resolver -name external
Config entry deleted: service-resolver/external
➜  consul git:(terminating-gateway-resolvers) ✗ curl -s http://localhost:19000/clusters | grep external | cut -d':' -f1 | uniq | sort
external.default.dc1.internal.4e38b938-3a3e-7b67-6624-a8dae1058918.consul
v1.external.default.dc1.internal.4e38b938-3a3e-7b67-6624-a8dae1058918.consul
v2.external.default.dc1.internal.4e38b938-3a3e-7b67-6624-a8dae1058918.consul
```

After the fix:

```zsh
➜  consul git:(terminating-gateway-resolvers) ✗ curl -s http://localhost:19000/clusters | grep external | cut -d':' -f1 | uniq | sort
external.default.dc1.internal.d27ec462-6e9d-edfa-bf3a-8d603e174492.consul
v1.external.default.dc1.internal.d27ec462-6e9d-edfa-bf3a-8d603e174492.consul
v2.external.default.dc1.internal.d27ec462-6e9d-edfa-bf3a-8d603e174492.consul
➜  consul git:(terminating-gateway-resolvers) ✗ ./consul config delete -kind service-resolver -name external
Config entry deleted: service-resolver/external
➜  consul git:(terminating-gateway-resolvers) ✗ curl -s http://localhost:19000/clusters | grep external | cut -d':' -f1 | uniq | sort
external.default.dc1.internal.d27ec462-6e9d-edfa-bf3a-8d603e174492.consul
```

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


nathancoleman:
Are there applicable tests anywhere for this code path?

andrewstucki:
@nathancoleman No, there aren't any tests that I can see which currently delete a `ServiceResolver` attached to an external service referenced by a `TerminatingGateway`, I'm planning on adding some after chatting with someone who has worked on terminating gateways a bit so I can ask some questions to make sure I'm doing something sane here and in the other PRs.

agent/proxycfg/terminating_gateway.go
```diff
@@ -354,8 +354,13 @@ func (s *handlerTerminatingGateway) handleUpdate(ctx context.Context, u UpdateEv
 		// There should only ever be one entry for a service resolver within a namespace
 		if resolver, ok := resp.Entry.(*structs.ServiceResolverConfigEntry); ok {
 			snap.TerminatingGateway.ServiceResolvers[sn] = resolver
+			snap.TerminatingGateway.ServiceResolversSet[sn] = true
+		} else {
+			// we likely have a deleted service resolver, and our cast is a nil
+			// cast, so clear this out
+			delete(snap.TerminatingGateway.ServiceResolvers, sn)
```
erichaberkorn, line 361: Is it possible to handle this similar to [here](https://github.com/hashicorp/consul/blob/21c30958cc89ba4a14a099d758002bc0a0ecc454/agent/proxycfg/terminating_gateway.go#L91-L99)? This allows us to differentiate between `nil` and a bad config entry.

andrewstucki: Sure, that should be fine, in the case of a bad cast though, we're likely going to want to clean this up, no? That or return an error about casting?

Though, not sure what use returning an error would do if when we're requesting a service resolver we get something else.

erichaberkorn: Yeah, that seems reasonable to me. This isn't a huge deal either way.

test/integration/connect/envoy/case-terminating-gateway-deleted-resolver/capture.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
erichaberkorn, line 1: It should be easier to test this [here](https://github.com/hashicorp/consul/blob/main/agent/proxycfg/state_test.go) rather than making an entirely new integration test.

andrewstucki: ah, cool, I was trying to figure out where I could do something that was a little less heavy-handed, but wasn't sure where we were doing unit-test like stuff for the proxycfg structs.

andrewstucki: @erichaberkorn removed the integration test and added a state-based unit test that does the job. Thanks for the tip!

erichaberkorn (approved):

PR #16499, opened 2023-03-02
Fix resolution of service resolvers with subsets for external upstreams

andrewstucki:
### Description

While fixing https://github.com/hashicorp/consul/pull/16498 I noticed that applying a `ServiceResolver` with subsets wasn't functional when referencing an external service proxied through a `TerminatingGateway` as an upstream. I'm not too familiar with the way we return service health for external services, but the problem appears to be that in our health check materializer we:

1. Grab the `CheckServiceNode` values from our subscription, and then
2. Apply any filters that were in our initial subscription request

Since we return the gateway associated with the service when we're using external services in conjuction with a `TerminatingGateway`:

https://github.com/hashicorp/consul/blob/21c30958cc89ba4a14a099d758002bc0a0ecc454/agent/consul/state/catalog.go#L2770-L2782

The filter never passes and we are never able to resolve the upstream endpoint properly.

**_I'm not entirely sure whether this is the only change needed, but from what I could tell all of the health checks initiated via `proxycfg` go through this code path since they leverage either the gRPC endpoints or a direct subscription to the in-memory store._**

Create a set of external services that has a `ServiceResolver` with subsets as in #16498 and a local service that leverages those services as an upstream. Hit the local proxy's admin cluster listing endpoint.

Without the fix (no ip address ever associates with the endpoint):

```bash
➜  ~ curl -s localhost:9092/clusters | grep v1.external | sort | head -n 1
v1.external.default.dc1.internal.cba29ba8-8796-2c26-cacd-0ee5dee70b82.consul::added_via_api::true
```

With the fix (contains terminating gateway ip for its endpoint):

```bash
➜  ~ curl -s localhost:9092/clusters | grep v1.external | sort | head -n 1
v1.external.default.dc1.internal.ea87fe29-2a6c-bd80-e248-5ebdbfed0a7a.consul::127.0.0.1:8443::canary::false
```

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/rpcclient/health/view.go
+++ b/agent/rpcclient/health/view.go
@@ -52,6 +52,8 @@ func NewHealthView(req structs.ServiceSpecificRequest) (*HealthView, error) {
 	return &HealthView{
 		state:  make(map[string]structs.CheckServiceNode),
 		filter: fe,
+		name:   req.ServiceName,
+		kind:   req.ServiceKind,
 	}, nil
 }
 
@@ -61,6 +63,8 @@ func NewHealthView(req structs.ServiceSpecificRequest) (*HealthView, error) {
 // (IndexedCheckServiceNodes) and update it in place for each event - that
 // involves re-sorting each time etc. though.
 type HealthView struct {
+	name   string
+	kind   structs.ServiceKind
 	state  map[string]structs.CheckServiceNode
 	filter filterEvaluator
 }
@@ -84,6 +88,13 @@ func (s *HealthView) Update(events []*pbsubscribe.Event) error {
 			if csn == nil {
 				return errors.New("check service node was unexpectedly nil")
 			}
+
+			// check if we intentionally need to skip the filter
+			if s.skipFilter(csn) {
+				s.state[id] = *csn
+				continue
+			}
+
 			passed, err := s.filter.Evaluate(*csn)
 			if err != nil {
 				return err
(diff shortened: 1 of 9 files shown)


andrewstucki:
@mkeeler could you take a look at this since you have the context from our discussion yesterday? Thanks!

agent/rpcclient/health/view.go
```diff
@@ -100,6 +111,17 @@ func (s *HealthView) Update(events []*pbsubscribe.Event) error {
 	return nil
 }
 
+func (s *HealthView) skipFilter(csn *structs.CheckServiceNode) bool {
+	// we only do this for services that need to be routed through a gateway
+	if s.kind != "" {
+		return false
+	}
+	if s.name != csn.Service.Service && csn.Service.Kind == structs.ServiceKindTerminatingGateway {
```
mkeeler, line 119: I wonder if `s.name != csn.Service.Service` is guaranteed to be the case.

In theory you could name the terminating gateway the same thing as some external service (I don't think this would be good for users to do but it probably would work.

A possibly better indication would be storing the `Connect` parameter from the `ServiceSpecificRequest` in the `NewHealthView` method. Then the heuristic for skipping here would be: `s.connect && csn.Service.Kind  == structs.ServiceKindTerminatingGateway`. I think that would make more sense because you are requesting connect endpoints for a service and our indexing wouldn't return a tgw of the same name for that query. https://github.com/hashicorp/consul/blob/5deffbd95bca0bd57187aceacffbd0a7cc0f972b/agent/consul/state/catalog_schema.go#L239-L253

So the only way a connect query will return a tgw is if there are external services and in that case we cannot expect the tgw metadata to match.

andrewstucki: That makes sense to me -- I can update the logic, but in general it seems like we need to check:

1. We are requesting health for a service,
2. We are setting the `Connect` flag, and
3. Instead of getting a service back, we get a terminating gateway

so something like `s.kind == "" && s.connect && csn.Service.Kind == structs.ServiceKindTerminatingGateway`

mkeeler (approved):

PR #16501, opened 2023-03-02
Suppress AlreadyRegisteredError to fix test retries

kisunji:
### Description

There is a new set of tests that have high rates of failure [(example)](https://app.circleci.com/pipelines/github/hashicorp/consul/52978/workflows/c3f09d66-2e45-4fb8-a56d-efca6223b88c/jobs/1235632).

They are labelled "flaky" but in reality are failing deterministically whenever there is a test retry which causes TestAgents to initialize the prometheus collectors more than once in a single runtime.

This PR suppresses [`prometheus.AlreadyRegisteredError`](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#AlreadyRegisteredError) so retries can run properly. This should have no impact on production usage as the metrics initialization happens only once on consul agent startup.

* Run a test which configures prometheus metrics from `metrics_test.go` like `TestHTTPHandlers_AgentMetrics_WAL_Prometheus` with `-count 2`. Without this patch, the second run should always fail.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/lib/telemetry.go
+++ b/lib/telemetry.go
@@ -345,7 +345,6 @@ func configureSinks(cfg TelemetryConfig, memSink metrics.MetricSink) (metrics.Fa
 	addSink(statsdSink)
 	addSink(dogstatdSink)
 	addSink(circonusSink)
-	addSink(circonusSink)
 	addSink(prometheusSink)
 
 	if len(sinks) > 0 {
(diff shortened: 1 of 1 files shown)


dhiaayachi (approved):
🚀

PR #16288, opened 2023-02-16
Improve ux to help users avoid overwriting fields of ACL tokens, roles and policies

roncodingenthusiast:
### Description
- Marking `mergePolicies` for deprecation and encouraging the use of `-append-policy-id` and `-append-policy-name` to reduce the ambiguity when overwriting fields on update 
- Updating docs to clarify that using: `-policy-id` and `-policy-name` will overwrite existing policies
- Marking `mergeRoles` for deprecation and encouraging the use of `-append-role-id` and `-append-role-name` to reduce the ambiguity when overwriting fields on update 
- Updating docs to clarify that using: `-append-role-id` and `-append-role-name` will overwrite existing policies

## TODO

- [x] add tests cases
- [x] update description
- [x] Do exact thing for `-merge-roles` as we did for `-merge-policies`
- [x] add change-log
- [x] de-duplicate all the logics 

### Testing & Reproduction steps
* In the case of bugs, describe how to replicate
* If any manual tests were done, document the steps and the conditions to replicate
* Call out any important/ relevant unit tests, e2e tests or integration tests you have added or are adding

### Links
- [current docs](https://developer.hashicorp.com/consul/commands/acl/token/update)

### PR Checklist

* [x] updated test coverage
* [x] external facing docs updated
* [x] not a security concern

PR #16506, opened 2023-03-02
Improve ux around ACL token to help users avoid overwriting node/service identities

roncodingenthusiast:
### Description

Follow up to https://github.com/hashicorp/consul/pull/16288 
TLDR: `node-identity` and `service-identity` flags when specified without the `merge-*` flags overwrites service identities and node identities of a token. This is ambiguous to the user because we have a `no-merge` flag which seems to imply that on update we are merging. So we have been doing work to improve the ux by renaming/adding/sunsetting flags and adding more docs to help reduce the ambiguity

- Marking `merge-node-identities` and `merge-service-identities` for deprecation and encouraging the use of their replacement `append-node-policy` and `append-service-identity`
- Updating docs to clarify that using `-rules` overwrites existing rules 
- Adding docs for the new flags: `append-node-policy` and `append-service-identity`

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/command/acl/token/update/token_update.go
+++ b/command/acl/token/update/token_update.go
@@ -25,37 +25,35 @@ type cmd struct {
 	http  *flags.HTTPFlags
 	help  string
 
-	tokenAccessorID    string
-	policyIDs          []string
-	appendPolicyIDs    []string
-	policyNames        []string
-	appendPolicyNames  []string
-	roleIDs            []string
-	appendRoleIDs      []string
-	roleNames          []string
-	appendRoleNames    []string
-	serviceIdents      []string
-	nodeIdents         []string
-	description        string
-	mergeServiceIdents bool
-	mergeNodeIdents    bool
-	showMeta           bool
-	format             string
+	tokenAccessorID     string
+	policyIDs           []string
+	appendPolicyIDs     []string
+	policyNames         []string
+	appendPolicyNames   []string
+	roleIDs             []string
+	appendRoleIDs       []string
+	roleNames           []string
+	appendRoleNames     []string
+	serviceIdents       []string
+	nodeIdents          []string
+	appendNodeIdents    []string
+	appendServiceIdents []string
+	description         string
+	showMeta            bool
+	format              string
 
 	// DEPRECATED
-	mergeRoles    bool
-	mergePolicies bool
-	tokenID       string
+	mergeServiceIdents bool
+	mergeNodeIdents    bool
+	mergeRoles         bool
+	mergePolicies      bool
+	tokenID            string
 }
 
 func (c *cmd) init() {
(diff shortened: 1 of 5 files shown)


command/acl/token/update/token_update.go
```diff
@@ -79,9 +77,15 @@ func (c *cmd) init() {
 	c.flags.Var((*flags.AppendSliceValue)(&c.serviceIdents), "service-identity", "Name of a "+
 		"service identity to use for this token. May be specified multiple times. Format is "+
 		"the SERVICENAME or SERVICENAME:DATACENTER1,DATACENTER2,...")
+	c.flags.Var((*flags.AppendSliceValue)(&c.appendServiceIdents), "append-service-identity", "Name of a "+
```
kisunji, line 80: can we add the "Overwrites existing X" comment on the non-append flags?

command/acl/token/update/token_update.go
```diff
@@ -147,13 +155,39 @@ func (c *cmd) Run(args []string) int {
...
+	if hasAppendServiceFields {
+		parsedServiceIdents, err = acl.ExtractServiceIdentities(c.appendServiceIdents)
+	}
+
+	if hasAppendServiceFields && hasServiceFields {
+		c.UI.Error("Cannot combine the use of service-identity flag with append-service-identity. " +
+			"To set or overwrite existing service identities, use -service-identity. " +
+			"To append to existing service identities, use -append-service-identity.")
+		return 1
+	}
```
kisunji, line 170: Let's move this up before we call `acl.ExtractServiceIdentities(c.serviceIdents)` to keep the validation ordering consistent with the NodeIdents below

command/acl/token/update/token_update_test.go
```diff
@@ -109,6 +109,22 @@ func TestTokenUpdateCommand(t *testing.T) {
...
+			"-accessor-id=" + token.AccessorID,
+			"-token=root",
+			"-append-node-identity=third:node",
+			"-description=test token",
+		})
+
+		require.Len(t, token.NodeIdentities, 3)
+		require.Equal(t, "third", token.NodeIdentities[2].NodeName)
+		require.Equal(t, "node", token.NodeIdentities[2].Datacenter)
+	})
```
kisunji, line 126: FYI subtests using `t.Run` should be independently runnable and pass. These append tests will not pass unless the previous subtests pass (but the previous test is also dependent on the previous-previous one).

As a team convention we tend to use `testutil.RunStep` to indicate a sequence of tests that must be run in a specific order.

Not a blocker, but just something to keep in mind as you write tests.

kisunji (approved):
Minor feedback but LGTM

cthain (approved):
No additional comments from me :+1:

PR #16511, opened 2023-03-02
Allow HCP metrics collection for Envoy proxies

Achooo:
### Description
Add a new envoy flag: `envoy_hcp_metrics_bind_socket_dir`, a directory where a unix socket will be created with the name `<namespace>_<proxy_id>.sock` to forward Envoy metrics.
If set, this will configure a local `stats_sink` and `STATIC` cluster to forward Envoy metrics to an HCP collector.

The HCP Cloud team is working on enabling observability metrics in HCP. 

There will be a follow-up PR to generate a listener and cluster dynamically that will be listening on this unix socket. 
That listener and cluster will be configured to route to collector instances through the service mesh.

Caveat: we needed to add this local listener indirection as a workaround due to this [issue](https://github.com/envoyproxy/envoy/issues/25700)

- Added unit tests
- Tested locally by running consul, registering a proxy with the flag configured and starting Envoy with consul connect command.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/command/connect/envoy/bootstrap_config.go
+++ b/command/connect/envoy/bootstrap_config.go
@@ -49,6 +49,11 @@ type BootstrapConfig struct {
 	// stats_config.stats_tags can be made by overriding envoy_stats_config_json.
 	StatsTags []string `mapstructure:"envoy_stats_tags"`
 
+	// HCPMetricsBindPort is an int that configures a local listener port
+	// where Envoy will forward metrics. These metrics get pushed to the HCP Metrics
+	// collector to show service mesh metrics on HCP.
+	HCPMetricsBindPort int `mapstructure:"envoy_hcp_metrics_bind_port"`
+
 	// PrometheusBindAddr configures an <ip>:<port> on which the Envoy will listen
 	// and expose a single /metrics HTTP endpoint for Prometheus to scrape. It
 	// does this by proxying that URL to the internal admin server's prometheus
@@ -238,6 +243,11 @@ func (c *BootstrapConfig) ConfigureArgs(args *BootstrapTplArgs, omitDeprecatedTa
 		args.StatsFlushInterval = c.StatsFlushInterval
 	}
 
+	// Setup HCP Metrics if needed. This MUST happen after the Static*JSON is set above
+	if c.HCPMetricsBindPort != 0 {
+		appendHCPMetricsConfig(args, c.HCPMetricsBindPort)
+	}
+
 	return nil
 }
 
@@ -271,7 +281,7 @@ func (c *BootstrapConfig) generateStatsSinks(args *BootstrapTplArgs) error {
 	}
 
 	if len(stats_sinks) > 0 {
-		args.StatsSinksJSON = "[\n" + strings.Join(stats_sinks, ",\n") + "\n]"
+		args.StatsSinksJSON = strings.Join(stats_sinks, ",\n")
 	}
 	return nil
 }
@@ -794,6 +804,52 @@ func (c *BootstrapConfig) generateListenerConfig(args *BootstrapTplArgs, bindAdd
 	args.StaticSecretsJSON += secretsJSON
 
 	return nil
+}
+
+func appendHCPMetricsConfig(args *BootstrapTplArgs, hcpMetricsBindPort int) {
+	if args.StatsSinksJSON != "" {
+		args.StatsSinksJSON += ",\n"
+	}
+	args.StatsSinksJSON += `{
+		"name": "envoy.stat_sinks.metrics_service",
+		"typed_config": {
+		  "@type": "type.googleapis.com/envoy.config.metrics.v3.MetricsServiceConfig",
+		  "transport_api_version": "V3",
+		  "grpc_service": {
+			"envoy_grpc": {
+			  "cluster_name": "hcp_metrics_collector"
+			}
+		  }
+		}
+	  }`
+
+	if args.StaticClustersJSON != "" {
+		args.StaticClustersJSON += ",\n"
+	}
+	args.StaticClustersJSON += fmt.Sprintf(`{
+		"name": "hcp_metrics_collector",
+		"type": "STATIC",
+		"http2_protocol_options": {},
+		"loadAssignment": {
+		  "clusterName": "hcp_metrics_collector",
+		  "endpoints": [
+			{
+			  "lbEndpoints": [
+				{
+				  "endpoint": {
+					"address": {
+					  "socket_address": {
+						"address": "127.0.0.1",
+						"port_value": %d
+					  }
+					}
+				  }
+				}
+			  ]
+			}
+		  ]
+		}
+	  }`, hcpMetricsBindPort)
 }
 
 func containsSelfAdminCluster(clustersJSON string) (bool, error) {
--- a/command/connect/envoy/bootstrap_config.go
+++ b/command/connect/envoy/bootstrap_config.go
@@ -49,10 +49,10 @@ type BootstrapConfig struct {
 	// stats_config.stats_tags can be made by overriding envoy_stats_config_json.
 	StatsTags []string `mapstructure:"envoy_stats_tags"`
 
-	// HCPMetricsBindPort is an int that configures a local listener port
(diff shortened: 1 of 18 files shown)


freddygv:
FYI @DanStough @wilkermichael we're thinking of holding off on merging this until after next week's patch releases.

The follow-up PR will be chained off of this one and will have the changelog entry.

command/connect/envoy/bootstrap_config_test.go
```diff
@@ -572,7 +618,7 @@ func TestBootstrapConfig_ConfigureArgs(t *testing.T) {
 			},
 			wantArgs: BootstrapTplArgs{
 				StatsConfigJSON: defaultStatsConfigJSON,
-				StatsSinksJSON: `[{
+				StatsSinksJSON: `{
```
Achooo, line 621: The tests below just remove the `[` and `]` which are now in the template

command/connect/envoy/bootstrap_tpl.go
```diff
@@ -262,7 +262,9 @@ const bootstrapTemplate = `{
     {{- end }}
   },
   {{- if .StatsSinksJSON }}
-  "stats_sinks": {{ .StatsSinksJSON }},
+  "stats_sinks": [
```
Achooo, line 265: https://github.com/hashicorp/consul/pull/16511/files#r1123725950

Move the brackets into the template file

command/connect/envoy/envoy.go
```diff
@@ -744,6 +744,10 @@ func (c *cmd) generateConfig() ([]byte, error) {
 		if err := mapstructure.WeakDecode(svcProxyConfig.Config, &bsCfg); err != nil {
 			return nil, fmt.Errorf("failed parsing Proxy.Config: %s", err)
 		}
+
+		if bsCfg.HCPMetricsBindPort < 0 || bsCfg.HCPMetricsBindPort > 65535 {
```
Achooo, line 748: Validate the port

DanStough (approved):
LGTM

website/content/commands/connect/envoy.mdx
```diff
@@ -75,6 +75,9 @@ Usage: `consul connect envoy [options] [-- pass-through options]`
   In cases where either assumption is violated this flag will prevent the
   command attempting to resolve config from the local agent.
 
+- `envoy_hcp_metrics_bind_port` - By setting this port, Envoy will be configured to
+  send metrics to an HCP collector. By default, this is disabled.
```
trujillo-adam, line 79:
```suggestion
- `envoy_hcp_metrics_bind_port` - Specifies the port number for Envoy to send metrics to. HCP collectors can collect the metrics at this port. The port is not configured by default. 
```
Attempted to reword this so that we are more concrete about how to configure the param. Better? Worse? No different?

trujillo-adam (commented):
I left a suggestion on the docs, but I'm not convinced that it's the perfect way to describe the param.

command/connect/envoy/bootstrap_config.go
```diff
@@ -796,6 +806,63 @@ func (c *BootstrapConfig) generateListenerConfig(args *BootstrapTplArgs, bindAdd
 	return nil
 }
 
+// appendHCPMetricsConfig generates config to enable a socket at path: <hcpMetricsBindSocketDir>/<namespace>_<proxy_id>.sock
+// or <hcpMetricsBindSocketDir>/<proxy_id>.sock, if namespace is empty.
+func appendHCPMetricsConfig(args *BootstrapTplArgs, hcpMetricsBindSocketDir string) {
+	dir := hcpMetricsBindSocketDir
```
Achooo, line 812: Handle `/` addition if needed from user input.

website/content/commands/connect/envoy.mdx
Achooo: Thank you so much for the review , and so sorry for the delay. We made some changes to our approach, see [here](https://github.com/hashicorp/consul/pull/16511/commits/3896dfe280c85865eb6476b505d63527d6f0a375#diff-3944e3639e03d55f8f432002a46d697c061b257ae0fc79e62815fcfd4ce4b47fR78). I tried to reword it based on your suggestions - which I liked. 

I'm not 💯 on the new changes. Would love to get your feedback on the new version, and maybe @freddygv has thoughts here.

website/content/commands/connect/envoy.mdx
```diff
@@ -75,6 +75,10 @@ Usage: `consul connect envoy [options] [-- pass-through options]`
   In cases where either assumption is violated this flag will prevent the
   command attempting to resolve config from the local agent.
 
+- `envoy_hcp_metrics_bind_socket_dir` - Specifies the directory of a unix socket
```
Achooo, line 78: Would `Specifies the directory where a unix socket is created` be better here?

Achooo:
See discussion on Windows support  concern here: https://github.com/hashicorp/consul-dataplane/pull/90#issuecomment-1460389995

TLDR: We'd like to move forward with this approach, and revisit at a later stage if needed.

command/connect/envoy/bootstrap_config.go
```diff
@@ -814,10 +814,13 @@ func appendHCPMetricsConfig(args *BootstrapTplArgs, hcpMetricsBindSocketDir stri
...
-	path := fmt.Sprintf("%s%s.sock", dir, args.ProxyID)
-	if args.Namespace != "" {
-		path = fmt.Sprintf("%s%s_%s.sock", dir, args.Namespace, args.ProxyID)
+	// Normalize namespace to "default". This ensures we match the namespace behaviour in proxycfg package,
+	// where a dynamic listener will be created at the same socket path via xDS.
+	namespace := args.Namespace
+	if namespace == "" {
+		namespace = "default"
 	}
+	path := fmt.Sprintf("%s%s_%s.sock", dir, namespace, args.ProxyID)
```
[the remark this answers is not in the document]
Achooo: Using this function fails the tests for a set namespace, as they all get normalized to `default`. I will add a test in enterprise to cover the other namespace case.

hashi-derek (approved):

website/content/commands/connect/envoy.mdx
trujillo-adam: What creates the directory? The passive voice sounds as if either Envoy or Consul creates the directory. If that's true, then we should say something like "Specifies the directory where <Envoy/Consul> creates a unix socket. Envoy sends metrics to the socket so that HCP collectors can connect collect them."

command/connect/envoy/bootstrap_config.go
```diff
@@ -796,6 +807,62 @@ func (c *BootstrapConfig) generateListenerConfig(args *BootstrapTplArgs, bindAdd
...
+// or <hcpMetricsBindSocketDir>/<proxy_id>.sock, if namespace is empty.
+func appendHCPMetricsConfig(args *BootstrapTplArgs, hcpMetricsBindSocketDir string) {
+	dir := hcpMetricsBindSocketDir
+	if !strings.HasSuffix(dir, "/") {
+		dir += "/"
+	}
+
+	// Normalize namespace to "default". This ensures we match the namespace behaviour in proxycfg package,
+	// where a dynamic listener will be created at the same socket path via xDS.
+	path := fmt.Sprintf("%s%s_%s.sock", dir, acl.NamespaceOrDefault(args.Namespace), args.ProxyID)
```
loshz, line 820: I would use the [path](https://pkg.go.dev/path) package here:
```suggestion
	sock := fmt.Sprintf("%s.sock", args.ProxyID)

	// Normalize namespace to "default". This ensures we match the namespace behaviour in proxycfg package,
	// where a dynamic listener will be created at the same socket path via xDS.
	path := path.Join(hcpMetricsBindSocketDir, acl.NamespaceOrDefault(args.Namespace), sock)
```

website/content/commands/connect/envoy.mdx
Achooo: Fixed that's much better, thank you.

command/connect/envoy/bootstrap_config.go
Achooo: thanks for the suggestion, fixed! That looks prettier 💯

Issue #16454, opened 2023-02-28
[API Gateway] HTTPRoute - when all services have 0 weight, we get protocol mismatches


andrewstucki:
#### Overview of the Issue

Two issues we have with HTTPRoute services is that we:

1. Seem to have failed to normalize  the weight of a given service to 1 by default (it stays 0 if unspecified)
2. Additionally, when a service is skipped in our discoverychain construction code:

https://github.com/hashicorp/consul/blob/73b9b407ba357f99cd02b5387fa15e159ab24327/agent/consul/discoverychain/gateway_httproute.go#L131-L133

we can potentially wind up with no service splits (if we have no services with non-zero weight), and we fail to append a corresponding splitter here:

https://github.com/hashicorp/consul/blob/73b9b407ba357f99cd02b5387fa15e159ab24327/agent/consul/discoverychain/gateway_httproute.go#L156-L158

that winds up with our synthetic router referencing a service route for a splitter that doesn't exist:

https://github.com/hashicorp/consul/blob/73b9b407ba357f99cd02b5387fa15e159ab24327/agent/consul/discoverychain/gateway_httproute.go#L168-L171

and we wind up with envoy spitting out tcp/http mismatch protocol errors (due to the compiled discovery chain thinking that the referenced, non-existent splitter should be treated as a TCP-resolver discovery node).

#### Reproduction Steps

Create and attach an HTTPRoute with referencing two services in a routing rule where each has 0 weight, such as:

```hcl
Rules = [
  {
    Services = [
      {
        Name = "service-one"
      },
      {
        Name = "service-two"
      }
    ]
  }
]
```

Attach to a gateway and watch Consul logs for protocol mismatch errors.

This should be fixed in two ways:

1. We should normalize to, `Weight = 1`, and 
2. We shouldn't add a route to the router if no corresponding splitter was added.

missylbytes:
I think we will also need to update the [docs](https://developer.hashicorp.com/consul/docs/connect/gateways/api-gateway/configuration/http-route#rules-services-weighthttps://developer.hashicorp.com/consul/docs/connect/gateways/api-gateway/configuration/http-route#rules-services-weight), we don't really talk about it being 1.

wahyugnc:
Can API Gateway with http route expose GRPC service?

PR #16512, opened 2023-03-02
NET-2903 Normalize weight for http routes

missylbytes:
### Description
Currently, if the weight of an HTTPRoute Service is unspecified or set to 0, the service will not be properly setup to have traffic routed to it.
These changes normalize the weight of a service on an HTTPRoute. If the weight is not specified or is set to 0, it will be normalized to 1.

Start consul with changes from branch
Use [this gist](https://gist.github.com/missylbytes/543537fc3af912b60d3319f983f7a5e4) to setup HTTPRoute services that have no weight defined.
You should not see an error. 
 
### Links
This prevents the bug seen in [this issue](https://github.com/hashicorp/consul/issues/16454)

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


andrewstucki (approved):
Overall this looks fine to me! If you want to gain some familiarity this _could_ also use an integration-level test, but I think that it's a clear enough change where that won't be necessary, so if you're having problems setting one up, no need.

website/content/docs/connect/gateways/api-gateway/configuration/http-route.mdx
```diff
@@ -630,7 +630,8 @@ Specifies the Enterprise [admin partition](/consul/docs/enterprise/admin-partiti
 
 ### `Rules[].Services[].Weight`
 
-Specifies the proportion of requests forwarded to the specified service. The
+Specifies the proportion of requests forwarded to the specified service. If no weight is specified, or if the specified
+weight is set to less than or equal to 0, it will be normalized to 1. The
```
trujillo-adam, line 634:
```suggestion
weight is set to less than or equal to `0`, the weight is normalized to `1`. The
```
Present tense code font for input values per the style guide.

trujillo-adam (approved):
Thanks for improving the description. I added some suggestions for style guide consistency. Pre-approving so you're not blocked.

PR #16518, opened 2023-03-03
modified unsupported envoy version error

wilkermichael:
### Description
- When an envoy version is out of a supported range, we now return the envoy version being used as `major.minor.x` to indicate that it is the minor version at most that is incompatible
- When an envoy version is in the list of unsupported envoy versions we return back the envoy version in the error message as `major.minor.patch` as now the exact version matters.

Note: Currently we do not have any envoy versions in the list of `unsupported` versions but we can support this in the future should the need arise to invalidate a particular major, minor or patch version

Verified error message with envoy out of bounds:
```shell
dc1svc2    | Envoy version 1.18.x is not supported. If there is a reason you need to use this version of envoy use the ignore-envoy-compatibility flag. Using an unsupported version of Envoy is not recommended and your experience may vary. For more information on compatibility see https://developer.hashicorp.com/consul/docs/connect/proxies/envoy#envoy-and-consul-client-agent
```

Verified error message with envoy in unsupported list:
```shell
dc1svc2    | Envoy version 1.18.3 is not supported. If there is a reason you need to use this version of envoy use the ignore-envoy-compatibility flag. Using an unsupported version of Envoy is not recommended and your experience may vary. For more information on compatibility see https://developer.hashicorp.com/consul/docs/connect/proxies/envoy#envoy-and-consul-client-agent

```

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/command/connect/envoy/envoy.go
+++ b/command/connect/envoy/envoy.go
@@ -503,15 +503,15 @@ func (c *cmd) run(args []string) int {
 			return 1
 		}
 
-		ok, err := checkEnvoyVersionCompatibility(v, xdscommon.UnsupportedEnvoyVersions)
+		ec, err := checkEnvoyVersionCompatibility(v, xdscommon.UnsupportedEnvoyVersions)
 
 		if err != nil {
 			c.UI.Warn("There was an error checking the compatibility of the envoy version: " + err.Error())
-		} else if !ok {
+		} else if !ec.isCompatible {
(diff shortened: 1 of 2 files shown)


david-yu (approved):
Thank you!

david-yu:
Added `backport/1.15` label.

thisisnotashwin (approved):
nice!

PR #16530, opened 2023-03-03
Follow-up fixes to consul connect envoy command

kisunji:
### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


command/connect/envoy/envoy.go
```diff
@@ -822,7 +821,7 @@ func (c *cmd) xdsAddress() (GRPC, error) {
 			// This is the dev mode default and recommended production setting if
 			// enabled.
 			port = 8502
-			c.UI.Info("-grpc-addr not provided and unable to discover a gRPC address for xDS. Defaulting to localhost:8502")
+			c.UI.Warn("-grpc-addr not provided and unable to discover a gRPC address for xDS. Defaulting to localhost:8502")
```
kisunji, line 824: Some workflows using the `-bootstrap` flag require stdout to be purely envoy config. This line is changed to output to stdout instead.

command/connect/envoy/envoy.go
```diff
@@ -887,7 +886,7 @@ func (c *cmd) lookupXDSPort() (int, string, error) {
 
 	var resp response
 	if err := mapstructure.Decode(self, &resp); err == nil {
-		if resp.XDS.Ports.TLS < 0 && resp.XDS.Ports.Plaintext < 0 {
+		if resp.XDS.Ports.TLS <= 0 && resp.XDS.Ports.Plaintext <= 0 {
```
kisunji, line 889: I had initially assumed `0` meant "not set by user" whereas `-1` meant "explicitly disabled by user". While this is true during consul agent startup, the response from `/v1/agent/self` actually has an empty XDS field, which decode the port fields to `0`.

This means that `0` effectively means "gRPC is disabled on the agent"

pglass (approved):

Issue #16455, opened 2023-02-28
[API Gateway] HTTPRoutes with weights that don't divide evenly into 10000


andrewstucki:
#### Overview of the Issue

Under the hood we do calculations based on the passed in weight for HTTPRoute services to normalize them into a percentage, which the virtual service splitters end up translating internally to envoy xDS, scaled to its required total weight of 10000 per-route.

https://github.com/hashicorp/consul/blob/73b9b407ba357f99cd02b5387fa15e159ab24327/agent/consul/discoverychain/gateway_httproute.go#L137-L141

When we provide weights that, when summed, don't divide evenly into 10000, we get errors due to errors in float rounding and envoy complains that it requires a total weight of 10000 when doing weighted load-balancing.

#### Reproduction Steps

Create and attach an HTTPRoute with weights for a set of services that don't divide evenly into 10000, such as:

```hcl
Rules = [
  {
    Services = [
      {
        Name = "service-one"
        Weight = 2
      },
      {
        Name = "service-two"
        Weight = 1
      }
    ]
  }
]
```

Here's a simple go snippet of the [rounding behavior we're hitting](https://go.dev/play/p/ipEce6JPgVl). We should ensure that we handle rounding errors by doing something like adding whatever offset necessary to the highest weighted service in order to normalize all weight percentages to exactly equal 100 (or 10000 when converted to xDS).

PR #16531, opened 2023-03-03
NET-2904 Fixes API Gateway Route Service Weight Division Error

missylbytes:
### Description

Fixes HTTPRoutes with weights that don't divide evenly into 10000. 
Previously, if a user created an HTTPRoute with services that had weights that did not divide evenly into 10000, (for example, two services, one with `weight=1` and one with `weight=2`), they would never be registered properly so no traffic would go to them.
Now the weights will be normalized.
Start consul with changes from branch
Use [this gist](https://gist.github.com/missylbytes/91b603ec81012bc42e490f1207afe160) to setup HTTPRoute services that have weights that do not divide evenly into 10000
Go to http://localhost:19000 , click on config-dump, and look at the `"dynamic_route_configs"` to see the route weights adding up to 10000.

### Links

Fixes https://github.com/hashicorp/consul/issues/16455

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


.changelog/16531.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+gateways: Fix HTTPRoute bug where services with a weight not divisible by 10000 are never registered properly
```
missylbytes, line 2:
```suggestion
gateways: fix HTTPRoute bug where services with a weight not divisible by 10000 are never registered properly
```

test/integration/connect/envoy/case-api-gateway-http-simple/setup.sh
```diff
@@ -35,6 +35,10 @@ rules = [
     services = [
       {
         name = "s1"
+      },
+      {
+        name = "s2"
+        weight = 2
```
andrewstucki, line 41: We'll need to remove the intentions checks in the verification step (which is fine IMO since they were only used to distinguish between services initially and we do that elsewhere), otherwise this test will fail 2/3 of the time since there's an explicit check for the listener this route binds to of "I can hit this service without an intentions error" and s2 will get intentions errors (and get routed to 2/3 of the time).

missylbytes: Updated the test

agent/xds/routes.go
```diff
@@ -893,6 +896,12 @@ func (s *ResourceGenerator) makeRouteActionForSplitter(
 		return nil, fmt.Errorf("number of clusters in splitter must be > 0; got %d", len(clusters))
 	}
 
+	if 10000 < totalWeight {
```
sarahalsmiller, line 899: Nit: I'd suggest moving 10000 into a variable so its easier to see what it means/why its that value

sarahalsmiller (approved):
Looks good to me, small nitpick, thanks for knocking this out so quickly

agent/xds/routes.go
missylbytes: fixed

andrewstucki (approved):

PR #16538, opened 2023-03-06
Add storage backend interface and in-memory implementation

boxofrad:
### Description

This PR introduces `storage.Backend`, which will serve as the interface between the [resource service](https://github.com/hashicorp/consul/pull/16528) and the underlying storage system (Raft today, but in the future, who knows! 😉).

Architecture Overview

The primary design goal of this interface is to keep its surface area small, and push as much functionality as possible into the layers above, so that new implementations can be added with little effort, and easily proven to be correct. To that end, we also provide a suite of "conformance" tests that can be run against a backend implementation to check it behaves correctly.

In this PR, we introduce an initial in-memory storage backend, which is suitable for tests and when running Consul in development mode. This backend is a thin wrapper around the `Store` type, which implements a resource database using [go-memdb](https://github.com/hashicorp/go-memdb) and our internal [pub/sub system](https://github.com/hashicorp/consul/tree/main/docs/rpc/streaming#event-publisher). Store will also be used to handle reads in our Raft backend, and in the future, used as a local cache for external storage systems.

```bash
$ go test -v ./internal/storage/inmem/...
```


internal/storage/inmem/store.go
```diff
@@ -0,0 +1,328 @@
...
+// local cache when storing data in external systems (e.g. RDBMS, K/V stores).
+type Store struct {
+	db  *memdb.MemDB
+	pub *stream.EventPublisher
+}
+
+// NewStore creates a Store.
+//
+// You must call Run before using the store.
+func NewStore() (*Store, error) {
```
boxofrad, line 27: With most of the 🧠s in this type, adding the Raft backend should mostly just be an exercise in plumbing it into the FSM + and pushing writes through [`raftApply`](https://github.com/hashicorp/consul/blob/9e93a30f4d3ca16de7468197842e6227b9cfec46/agent/consul/rpc.go#L898-L970).

internal/storage/inmem/query.go
```diff
@@ -0,0 +1,187 @@
...
+	b.String(t.Group)
+	b.String(t.Kind)
+	return b.Bytes()
+}
+
+func indexFromTenancy(t *pbresource.Tenancy) []byte {
+	var b indexBuilder
+	b.String(t.Partition)
+	b.String(t.PeerName)
+	b.String(t.Namespace)
```
boxofrad, line 86: To avoid having to implement this twice (and handle the ensuing OSS→ENT merge hell) I was thinking of keeping the partition and namespace plumbing in OSS, but having the brains/controllers live in ENT only, and rejecting non-default values at the RPC handle level.

Does that sound ok, @mkeeler?

internal/storage/inmem/query.go
```diff
@@ -0,0 +1,187 @@
...
+	q, ok := args[0].(query)
+	if !ok {
+		return nil, fmt.Errorf("expected query, got: %T", args[0])
+	}
+	return q.indexPrefix(), nil
+}
+
+// ownerIndexer implements the memdb.SingleIndexer interface. It is used for
+// indexing resources by their owners.
+type ownerIndexer struct{ readIndexer }
```
andrewstucki, line 61: Just adding this here as an eventual idea, but if we're going to use this eventually as a local cache for remote backends, we may consider exposing these indexers to be extended on a per-type basis. The corresponding local indexers for Kubernetes allow you to do something similar, basically letting you index some arbitrary fields via a call to [IndexField](https://github.com/kubernetes-sigs/controller-runtime/blob/e2d8821b277fb57f3a951f430bb1746d8297b853/pkg/cache/informer_cache.go#L161-L166) in your setup code and then query them via an explicit `FieldSelector` parameter on your query.

See envoy's gateway controller implementation for usage around calling [IndexField](https://github.com/envoyproxy/gateway/blob/18890ecc7e526cfb5d1ee5958237badff215a12b/internal/provider/kubernetes/controller.go#L486-L488), an [example indexer](https://github.com/envoyproxy/gateway/blob/18890ecc7e526cfb5d1ee5958237badff215a12b/internal/provider/kubernetes/controller.go#L504-L523), and how the index is [being queried](https://github.com/envoyproxy/gateway/blob/main/internal/provider/kubernetes/routes.go#L178-L180). I'd imagine we'd eventually greatly benefit if we were able to do something like be able to add additional resource-based indices at resource registration time.

internal/storage/conformance/conformance.go
```diff
@@ -0,0 +1,611 @@
...
+				require.NoError(t, err)
+
+				require.Equal(t, pbresource.WatchEvent_OPERATION_DELETE, event.Operation)
+				require.Equal(t, del, event.Resource)
+			})
+		}
+	})
+}
+
+func testWatch(t *testing.T, opts TestOptions) {
```
analogue, line 453: I'm not seeing where this function is getting called.

analogue (approved):
Overall, looks good to me. Nice work!

I'm guessing the tests for `ReadConsistent(..)` don't make sense until the raft backed storage backend.

internal/storage/conformance/conformance.go
boxofrad: Whoops! Good catch. Got left behind in a refactoring 🤦🏻

internal/storage/inmem/query.go
boxofrad: Thanks for the idea, @andrewstucki! I've wondered about something similar too. I wonder if it'd be better to maintain a clean separation and have controllers implement their own MemDB stores (with their own custom indices) though?

internal/storage/inmem/store.go
```diff
@@ -0,0 +1,328 @@
...
+func (s *Store) WriteCAS(res *pbresource.Resource, vsn string) error {
+	tx := s.db.Txn(true)
+	defer tx.Abort()
+
+	existing, err := tx.First(tableNameResources, indexNameID, res.Id)
+	if err != nil {
+		return err
+	}
+
+	if existing == nil && vsn != "" {
```
mkeeler, line 122: nit: Maybe add a comment about all creations using the WriteCAS method are expected to be issued with an empty version. It took me a moment to figure out that this was correct in order to prevent two creations of the same resource racing in odd ways.

internal/storage/inmem/store.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+	return list, nil
+}
+
+type meta struct {
+	Key   string
+	Value any
+}
+
+func incrementEventIndex(tx *memdb.Txn) (uint64, error) {
```
mkeeler, line 306: Is the event index generally just a way of having a monotonically increasing value for all writes that wont be coupled with the Raft index?

internal/storage/inmem/store.go
```diff
@@ -0,0 +1,328 @@
+package inmem
```
mkeeler, line 1: nit:

Do you think it would be worthwhile to break apart this file into a few:

store_schema.go - for defining the memdb schema and constants (maybe this should be combined with query.go?)
store_resources.go - for methods pertaining to the core resource centric method
store_monotonic_index.go - for methods pertaining to maintaining the monotonically increasing index counter.

internal/storage/inmem/store.go
```diff
@@ -0,0 +1,328 @@
...
+		return err
+	}
+
+	idx, err := incrementEventIndex(tx)
+	if err != nil {
+		return nil
+	}
+	tx.Commit()
+
+	s.publishEvent(idx, &pbresource.WatchEvent{
```
mkeeler, line 197: ❗ (probably a blocker)

With event publishing after the `tx.Commit` isn't it theoretically possible to see events out-of-order. 

If using this Store in a single-threaded fashion like we currently do with Raft + MemDB then concurrent writes wont be happening anyways so its *extremely* unlikely it happens as it would require the go routine to be paused for long enough between the `tx.Commit` and the `publishEvent` calls happen for another write + publish to occur. 

As this is the in-mem store and it appears like there are no guarantees of consistency or write serialization, it feels like this is more likely to happen. That in turn could be a source of flakey tests.

Do you think it would be too difficult to integrate the event publishing bits in here with memdbs changeset tracking to ensure that events are published in the same order that the txns occur.

internal/storage/inmem/query.go
andrewstucki: Yeah, I'm fine either way, and again, this is definitely not an "initial pass" sort of thing. It makes sense to me to have controller-specific stores/indices in the sense that you don't wind up having to worry as much about having potentially conflicting indexed fields -- that said, it would mean having more layers of caching/storage, alternatively we could eventually expose doing something like prefixing controller names into the indexed field and implicitly appending those prefixes to a lookup from the controller, that way they're essentially still acting on the monolithic cache and we have less memory usage -- but 🤷, we can figure that out in subsequent PRs/layers of the architecture.

internal/storage/inmem/query.go
jmurret: starting a slack convo...

internal/storage/inmem/store.go
```diff
@@ -0,0 +1,328 @@
...
+			},
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	s := &Store{
+		db:  db,
+		pub: stream.NewEventPublisher(10 * time.Second),
```
jmurret, line 66: ❓ How do we arrive at this frequency?  Is this what is used in the server agent today? If so, may be adding comment for context would be helpful.

internal/storage/inmem/watch.go
```diff
@@ -0,0 +1,188 @@
...
+func (s tenancySubject) String() string {
+	return s.resourceType.Group + indexSeparator +
+		s.resourceType.Kind + indexSeparator +
+		s.tenancy.Partition + indexSeparator +
+		s.tenancy.PeerName + indexSeparator +
+		s.tenancy.Namespace
+}
+
+// publishEvent sends the event to the relevant Watches.
+func (s *Store) publishEvent(idx uint64, event *pbresource.WatchEvent) {
```
jmurret, line 108: [super nit :) ] Does the caller need to construct a WatchEvent or could they do:
```
s.publishEvent(idx, pbresource.WatchEvent_OPERATION_UPSERT, res)
```

instead of 
```
	s.publishEvent(idx, &pbresource.WatchEvent{
		Operation: pbresource.WatchEvent_OPERATION_UPSERT,
		Resource:  res,
	})
```

internal/storage/storage.go
```diff
@@ -0,0 +1,220 @@
...
+	// single-resource sequential consistency, typically by bypassing any caches
+	// and proxying the request directly to the underlying storage system.
+	//
+	// If a consistent read cannot be achieved (e.g. when interacting with a Raft
+	// follower, or quorum is lost) ErrInconsistent will be returned.
+	//
+	// Use ReadConsistent sparingly, and prefer Read when possible.
+	ReadConsistent(ctx context.Context, id *pbresource.ID) (*pbresource.Resource, error)
+
+	// WriteCAS performs an atomic CAS (Check-And-Set) write of a resource based
```
jmurret, line 89: [nit/question] I seems like [consul docs for kv put](https://developer.hashicorp.com/consul/commands/kv/put) refer to `-cas` as a Check-And-Set operation, but I think everyone else refers to CAS as Compare And Swap.  Our implementation is consistent with Compare and Swap.  I think it might be confusing since Test-And-Set is another locking operation type.  If this is all inadvertent, I'd like to change this here and in another PR, we can update the consul docs.  I'm just not sure if I am missing a broader use of `Check-And-Set`.

jmurret (commented):
Super cool!  I left a few nits and questions, but love this progress you made.  Awesome job!

internal/storage/storage.go
```diff
@@ -0,0 +1,220 @@
...
+
+	// ReadConsistent provides the same functionality as Read, but guarantees
+	// single-resource sequential consistency, typically by bypassing any caches
+	// and proxying the request directly to the underlying storage system.
+	//
+	// If a consistent read cannot be achieved (e.g. when interacting with a Raft
+	// follower, or quorum is lost) ErrInconsistent will be returned.
+	//
+	// Use ReadConsistent sparingly, and prefer Read when possible.
+	ReadConsistent(ctx context.Context, id *pbresource.ID) (*pbresource.Resource, error)
```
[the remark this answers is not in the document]
boxofrad: I like that a lot! I'm imagining some helpers methods like this:

```Go
resource, err := backend.Read(ctx, id, storage.MaxStaleness(30 * time.Second))

resource, err := backend.Read(ctx, id, storage.StronglyConsistent())
```

Couple of follow up questions:
1. Am I right that we don't actually have a way to measure the staleness of a Raft follower as a wall clock time? I guess the closest we could do would be to compare the applied index and make inferences from there.
2. Other than strong/eventual consistency and max staleness, can you think of any other parameters we'd want to support?
3. Do we want to maintain the difference between our "default" and "consistent" consistency modes? In https://github.com/hashicorp/consul/pull/16619 I've more-or-less abandoned the "default" mode, with the rationale being that we're going to assume results are stale and use expensive consistent reads sparingly (i.e. just for ACL enforcement).

internal/storage/storage.go
```diff
@@ -0,0 +1,220 @@
...
+	//
+	// See Backend docs for more details.
+	//
+	// # GroupVersion
+	//
+	// Write does not validate the GroupVersion and allows you to overwrite a
+	// resource stored in an older form with a newer, and vice versa.
+	//
+	// See Backend docs for more details.
+	WriteCAS(ctx context.Context, res *pbresource.Resource, version string) (*pbresource.Resource, error)
```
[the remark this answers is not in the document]
boxofrad: Yep, you got it. Carrying the status forward, preventing raciness around the `Uid` on-creation, that sort of thing. Will add to the docs 👍🏻 

Agreed, we should auto-retry CASified user writes.

internal/storage/storage.go
```diff
@@ -0,0 +1,220 @@
...
+
+	// List resources of the given type, tenancy, and optionally matching the given
+	// name prefix.
+	//
+	// # Tenancy Wildcard
+	//
+	// In order to list resources across multiple tenancy units (e.g. partitions)
+	// pass the Wildcard sentinel value in tenancy fields.
+	//
+	// # GroupVersion
```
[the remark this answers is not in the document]
boxofrad: Good question! I'd vote for keeping things simple at the storage backend layer, and optimizing if we find it's a problem in the future.

internal/storage/storage.go
```diff
@@ -0,0 +1,220 @@
...
+	// See List docs for details about Tenancy Wildcard, GroupVersion, and
+	// Consistency.
+	WatchList(ctx context.Context, resType UnversionedType, tenancy *pbresource.Tenancy, namePrefix string) (Watch, error)
+
+	// OwnerReferences returns the IDs of resources owned by the resource with the
+	// given ID. It is typically used to implement cascading deletion.
+	//
+	// # Consistency
+	//
+	// OwnerReferences makes no guarantees about consistency, and may return stale
```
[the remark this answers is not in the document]
boxofrad: Thanks for raising this! I'd originally convinced myself that stale results were fine because by the point the reaper controller would call this method, the owner had been deleted, so we wouldn't allow the creation of any more owned resources - but that's still racy.

I think this might need to be strongly consistent after all.

internal/storage/inmem/store.go
```diff
@@ -0,0 +1,328 @@
...
+		existingRes := existing.(*pbresource.Resource)
+
+		// Ensure CAS semantics.
+		if existingRes.Version != vsn {
+			return storage.ErrConflict
+		}
+
+		// Uid is immutable.
+		if existingRes.Id.Uid != res.Id.Uid {
+			return storage.ErrConflict
```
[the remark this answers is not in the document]
boxofrad: Sold! That's much better ⭐

internal/storage/inmem/store.go
boxofrad: Ah yes, good call. It's non-obvious for sure!

internal/storage/inmem/store.go
boxofrad: Yep, exactly. It's purely for the event publishing machinery (and splicing the snapshot together with subsequent events).

internal/storage/inmem/store.go
boxofrad: Good catch. Just moving the call to `publishEvent` before the transaction is committed should do the trick (which is what the changeset tracking does anyway).

https://github.com/hashicorp/consul/blob/d5677e5680050a9b22403c160f5edde59613a4fc/agent/consul/state/memdb.go#L164-L170

internal/storage/inmem/store.go
boxofrad: Copied from here:

https://github.com/hashicorp/consul/blob/d5677e5680050a9b22403c160f5edde59613a4fc/agent/setup.go#L182

No idea how it was originally chosen 😄

internal/storage/inmem/watch.go
boxofrad: Good call! 🙌🏻

internal/storage/storage.go
boxofrad: Ah yes, I did copy it from the KV definition of CAS. Let's go with Compare-And-Swap, as it's more accurately descriptive.

boxofrad:
Thanks for the thorough review, @mkeeler 🙇🏻 

I've replied to your comments inline; lots of really great suggestions there, which I'll work through next. In answer to your other questions:

> First, should we have some support for blocking query like functionality. Mainly this would be within the `Read`/`List` methods on the `Backend`/`Store` types (and I guess on the `StorageBackend` interface). I think supporting blocking would likely require some change in the backends themselves as that is where MemDB lives so the resource API itself cannot necessarily handle the blocking.

I can't immediately see a use case for blocking queries in the controller architecture that isn't already solved (or at least could be approximated) by watches.

If we did support them, would we expose them via the resources API? I think that means we'd need to expose an index, or maybe block on the resource's version instead.

> Secondly, should we remove all partition knowledge from indexing? I was envisioning a scenario where the OSS storage backend would be peer/ns aware but not partition aware. Then there would be a Enterprise storage backend implementation that keeps 1 Store per partition and for each operation selects the correct store to use based on the partition. 

Wow, that's a really smart solution that would solve a bunch of the pain associated with keeping the indexing in enterprise. I guess one downside (or potentially upside) would be that we'd need to run an individual instance of each controller for every partition. Let's keep chatting about this on Slack/in our weekly sync.

internal/storage/inmem/store.go
mkeeler: I don't know either but it seems reasonable. 

The snapshot points to the event buffer head at the time of its creation. As time moves forward new events get appended to the buffer and any active streams will go ahead and read those events from the stream. Even once all active streams have processed the events, the events are prevented from being garbage collected due to the snapshot pointing at them. 

Setting the snapshot cache ttl to 10s allows us to not be constantly recalculating snapshots for every new streaming request but isn't so high as to cause memory usage to get too high.

internal/storage/inmem/store.go
mkeeler: We discussed this a bunch offline and having the publish after memdb txn commit should be fine iff we also ensure that no more than one write operation against the store is happening.

Basically we need to add a `sync.Mutex` to the `Store` and lock it for the duration of the `WriteCAS` and `DeleteCAS` methods. One optimization to that could be to lock it immediately before the `txn.Commit` calls and relinquish the lock once the events have been published. MemDB will already prevent two write txns from progressing at the same time and and not gaining the lock until the end will allow the memdb txn preparation concurrently with the previous writers event publishing. However that seems like it may be premature optimization so just locking the whole method also seems fine.

internal/storage/inmem/store.go
```diff
@@ -0,0 +1,328 @@
...
+		}
+	}
+
+	return res, nil
+}
+
+// WriteCAS performs an atomic Check-And-Set (CAS) write of a resource.
+//
+// For more information, see the storage.Backend documentation.
+func (s *Store) WriteCAS(res *pbresource.Resource, vsn string) error {
```
mkeeler, line 113: Is it expected that the version of the resource is set on the resource prior to invoking `WriteCAS` and `DeleteCAS`?

Should these two methods additionally take in the new version and update the resource accordingly?

internal/storage/storage.go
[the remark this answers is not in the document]
mkeeler: 1. Some RPCs respect the `max_stale` query param today. They issue the query once with `AllowStale` being true, the response tacks on the time since the follower was last contacted by the leader. If the max stale duration exceeds that last contact time, the RPC will be retried against the leader instead: https://github.com/hashicorp/consul/blob/9d21736e9f5828920cb8302397e0f11c95cae88c/agent/catalog_endpoint.go#L250-L254

Comparing against the index would require knowing what index the leader was at. I _think_ we could glean that information from Raft to detect that its having a hard time catching up even if the leader is contacting the follower regularly.

2. I cannot think of anything else we would need right now. I also don't see adding more options as too difficult a task either so I would focus on what we know we need now.
3. In general I think the consistency modes will be up for interpretation by backends. When using Raft consistent means going to the leader whereas for leaderless backends consistent could mean retrieving from the underlying storage (k8s, dynamo etc) in a consistent manner. I need to review your other PR but my initial inclination would be to keep default mode as an option. Maybe there is a good reason why it doesn't work out cleanly and we need to drop it. I think your assumptions about our internal usage of the APIs being predominantly stale is correct. Whether that would hold for other users of the API I am less certain. I will come back to this after taking a look at your other PR.

internal/storage/storage.go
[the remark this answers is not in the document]
mkeeler: I _think_ the assumptions will still hold.

If you pass in a resource id that contains a uid then only resources owned by that incarnation of the parent resource should be returned. As the original parent has been deleted, no more resources owned by that incarnation will be created. Others owned by a new incarnation may but we shouldn't care.

internal/storage/conformance/conformance.go
```diff
@@ -0,0 +1,592 @@
...
+
+				for i := 0; i < len(tc.results); i++ {
+					ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+					t.Cleanup(cancel)
+
+					event, err := watch.Next(ctx)
+					require.NoError(t, err)
+
+					require.Equal(t, pbresource.WatchEvent_OPERATION_UPSERT, event.Operation)
+					require.Containsf(t, tc.results, event.Resource, "resource not in expected results: %s", event.Resource.Id)
```
boxofrad, line 433: Reminder to myself after [our discussion](https://github.com/hashicorp/consul/pull/16538#discussion_r1134443895): we should try to `Read` the resource here, to make sure reads are sequentially consistent with watches (which isn't currently the case). We should also document the expectation too.

internal/storage/inmem/store.go
boxofrad: Yep, exactly. The expectation is that the Resource Service will generate the `Version`, `Generation`, and `Uid` so we can use a consistent scheme ([ULIDs](https://github.com/ulid/spec)) across the board.

The storage backend doesn't need to do anything special to update the version (or even validate that it has been changed) so I don't think we need a separate argument.

internal/storage/storage.go
boxofrad: Fixed in https://github.com/hashicorp/consul/pull/16538/commits/f54bcb8024ab0b3176850287120799df004cbaf3

internal/storage/inmem/watch.go
boxofrad: Fixed in https://github.com/hashicorp/consul/pull/16538/commits/0167a5bc1eedd1d7ba6badae9a12ec366e7bb7dc

internal/storage/storage.go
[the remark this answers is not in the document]
boxofrad: Documented in https://github.com/hashicorp/consul/pull/16538/commits/048712d39e062ea4f742965ed7655e7e626248ea

internal/storage/storage.go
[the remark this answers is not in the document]
boxofrad: Thanks for the answers, Matt!

### Bounding Staleness
I remember we spoke about bounding staleness before in the context of the work @rboyer was doing to prevent old servers re-joining the cluster and handling RPCs. There are a couple of approaches we could take:

#### Stamp Raft log entries with the wall-clock time, and track the highest timestamp seen
Clock skew aside, this would be imprecise in clusters with light/infrequent write load (unless we pushed the timestamp out with the Raft heartbeat).

#### Estimate the number of writes per second
By estimating the leader is handling 100 writes/s, for example, a follower could infer that it is lagging `(<leader commit index> - <follower applied index>) / 100` seconds behind.

### Consistency Modes

I'd be in favor of supporting a smaller number of modes with very clearly defined guarantees that can be supported by all backends.

The Raft backend introduced in #16619 essentially supports two:
- **Eventually Consistent (default):** Equivalent to our current "stale" mode. Reads are [sequentially consistent](https://jepsen.io/consistency/models/sequential) with each other, but may not reflect writes made to the same server because of leader forwarding.
- **Strongly Consistent:** Equivalent to our current "consistent" mode. Full [linearizability](https://jepsen.io/consistency/models/linearizable).

The main difference between our current "default" and "consistent" modes is whether or not we do a [round of heartbeats](https://pkg.go.dev/github.com/hashicorp/raft#Raft.VerifyLeader) to verify leadership. In my opinion, the "default" mode only really makes sense if consistent reads are common enough to justify the reduction in safety for the sake of performance or scalability.

In the new controller architecture, we're assuming eventual consistency for the majority of our internal uses. Which I think makes it a reasonable assumption that if we opt into strong consistency, correctness really matters, so we should verify leadership.

You do raise a good point about external users of the API, though.

internal/storage/inmem/store.go
mkeeler: I am not sure ULIDs are appropriate for the resource version (generation and uid should be fine).

For a single resource I think ULIDs can provide write linearizability as we generally expect in a few places.

Scenario (two concurrent writes to a single resource):
* Some resource exists with ULID version of `01GVJVKJTMXSD3FT9ZSEGJEAVM`
* Write A received
   * ULID generated is `01GVJVMFHBT9NK7MYEXCMCNFHG`
   * CAS version to be used will be `01GVJVKJTMXSD3FT9ZSEGJEAVM`
* Write B received 
   * ULID generated is `01GVJVMFHBZC4SWHT5XPCZJNWW`
   * CAS version to be used will be `01GVJVKJTMXSD3FT9ZSEGJEAVM`
* _Note that writes A an B share the same 10 byte timestamp prefix due to being generated in the same millisecond. Also note that write Bs ULID is lexicographically after A_
* _For whatever reason the validation callbacks for write A take longer than write B_
* API sends write B to the storage backend
* Storage backend applies write B as the CAS index matches the resources current version.
* API sends write A to the storage backend
* Storage backend rejects write A due to the CAS index mismatch since write B changed the resource version
* API retries write A
   * New ULID is generated (_I think this must happen in order for the ULID to be lexicographically after write B, however that will only be true if the process to attempt writing to the storage backend and be rejected would take more than 1ms_)
   * CAS version is updated to write Bs version
* Storage Backend applies write A

I think this is okay because:

1. We are CASifying all writes.
2. It is practically impossible today for the time between ULID generation and ULID regeneration after a rejected CAS write to take less than 1ms.

The second constraint there I think will be valid for the Raft backend or any generally any backend that does disk or network IO (like k8s or dynamo or even some flat file backend). However for the inmem storage backend for testing I am not quite as convinced that it will always be the case.

Things get even gnarlier when you consider operations against multiple resources of some kind such as a resource listing. Since concurrent writes to multiple distinct resources will not have CAS operations to help enforce version linearizability it is entirely possible and even expected for resources to be written and events seen for resources out of lexicographic order based on the ULID version.

Is this a problem for watches, particularly if we want to allow watch resumption? What about if we wanted to support [functionality similar to k8s](https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions) where you can provide a resource version to list operations and only get data newer than that version. 

In some of those cases strict linearizability of all resource writes of a particular type is necessary. Do we care and are we willing to rule out implementing those features to avoid having the storage backends more involved in setting resource versions?

boxofrad: I probably should've clarified, I'm not planning to depend on the lexicographical sortability of ULIDs for determining the order of events or anything else related to correctness. The timestamp component is inherently susceptible to clock skew across servers, and other anomalies on the same server, so this wouldn't be safe, even if the "entropy" component was a monotonic index.

Essentially, we'll be treating ULIDs as opaque random strings, but with the benefit of being able to tell if one version came **roughly** before or after another (think day/week/month level of granularity) 😄 but if this leads to folks making greater assumptions about ordering, I'd be keen to replace it with a truly random string.

I haven't accounted for watch resumption, or listing resources that are "newer" than a given version, because they require such strict consistency guarantees ([serializability](https://jepsen.io/consistency/models/serializable) I think) to implement correctly, which would rule out a lot of potential external storage systems.

I guess what you're saying is that Raft *does* provide serializability, so why not use the log index as a version, so that we at least have the option of supporting these features in the future? I'm not strongly against this, but would argue that we should think carefully before adding features that can only be satisfied with strong consistency, because it closes the door on lots of potentially great storage integrations.

internal/storage/storage.go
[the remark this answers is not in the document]
boxofrad: Following up from our call earlier:

We'll need to support the old "default" mode for backward-compatibility with the v1 API. To that end, I think we'll have a single `Read` method with a consistency param like you suggested, and a "legacy" mode that's only supported by the Raft backend (and isn't available in the v2 API at all).

I think we can leave bounded staleness as a potential future improvement. As we discussed, it's difficult to implement something better than just checking the last leader contact timestamp (and probably even harder in external systems we have less control over).

internal/storage/inmem/store.go
boxofrad: Following up from our call earlier:

We're going to make storage backends responsible for managing the version, and have the Raft backend use the log index so that we can build a blocking query layer on top (again, for backward-compatibility with the v1 API). Other backends will probably continue to treat versions as opaque strings, and make them ULIDs or something.

internal/storage/storage.go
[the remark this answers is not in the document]
boxofrad: Sorry for all the back-and-forth on this topic 😅 

I realized shortly after writing my last comment, that if the v1 handlers:
- Continue to do their own leader forwarding
- Call the resource service in the same process (via something similar to our net/rpc in-mem codec)

We don't actually need to support a "legacy" mode at the storage backend level, because a leader handling an "eventually consistent" read provides the same guarantees as our current "default" mode 🎉

I also realized that the (intentional) omission of a strongly consistent version of `List` breaks backward compatibility. I don't think we want to support strongly consistent list operations going forward, because it requires serializability, but we probably need to support it in the Raft backend until the v1 APIs are gone.

internal/storage/inmem/query.go
boxofrad: Keeping this as-is for the moment 🧑‍⚖️ 

I really like @mkeeler's idea of having separate stores (or even backends) for each partition, and think this is ultimately the way to go, but it comes with some complexity that I'd like to defer for now so we can make progress on the other parts of this project.

As separate stores makes cross-partition watches impossible, or at least more difficult, need to either run a controller per-partition or have controllers watch resources in multiple partitions, and then reconfigure all of this on the fly as partitions are created and deleted.

mkeeler (approved):

PR #16545, opened 2023-03-06
Fix flakey tests related to ACL token updates

roncodingenthusiast:
### Description

The idea is to make sure each test can independently run so this PR reduce the dependency of tests on each other which causes a degree of flakiness. 

- Each tests depended on the same token to run which sometimes meant one test will overwrite changes from the other. So fixing this by allowing each test to create its own individual token that it could update and read to without any interference from other tests. 

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/command/acl/token/update/token_update_test.go
+++ b/command/acl/token/update/token_update_test.go
@@ -50,13 +50,6 @@ func TestTokenUpdateCommand(t *testing.T) {
 	)
 	require.NoError(t, err)
 
-	// create a token
-	token, _, err := client.ACL().TokenCreate(
-		&api.ACLToken{Description: "test"},
-		&api.WriteOptions{Token: "root"},
-	)
-	require.NoError(t, err)
-
 	run := func(t *testing.T, args []string) *api.ACLToken {
 		ui := cli.NewMockUi()
 		cmd := New(ui)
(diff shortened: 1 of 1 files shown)


command/acl/token/update/token_update_test.go
```diff
@@ -72,21 +65,38 @@ func TestTokenUpdateCommand(t *testing.T) {
 
 	// update with node identity
 	t.Run("node-identity", func(t *testing.T) {
-		token := run(t, []string{
+		// create a token
+		token, _, err := client.ACL().TokenCreate(
```
roncodingenthusiast, line 69: i wonder if i could move this token creation to a helper that just returns the token 🤔  maybe i am overcomplicating this

command/acl/token/update/token_update_test.go
```diff
@@ -219,27 +228,13 @@ func TestTokenUpdateCommandWithAppend(t *testing.T) {
...
-		&api.WriteOptions{Token: "root"},
-	)
-	require.NoError(t, err)
-
 	//secondary policy
 	secondPolicy, _, policyErr := client.ACL().PolicyCreate(
 		&api.ACLPolicy{Name: "secondary-policy"},
 		&api.WriteOptions{Token: "root"},
 	)
 	require.NoError(t, policyErr)
```
kisunji, line 236: Can this be removed as well?

kisunji (approved):
LGTM! Thanks for cleaning this up

command/acl/token/update/token_update_test.go
cthain: It's certainly a possibility given how many times this construct appears in the tests.. but I think it's completely up to you to decide if you think it is worth factoring this out of the tests. It might be nice to be able to replace these occurrences with something like:

```go
token := createToken(t, aclToken, writeOpts)
```
And inside `createToken`, you could use `t` to make the necessary assertions. You still have to pass the token and write options from each test so it probably won't save a bunch of space.

cthain (approved):
This is awesome, thanks @roncodingenthusiast!

command/acl/token/update/token_update_test.go
roncodingenthusiast: this second policy is used by some tests

PR #16552, opened 2023-03-07
perf: Remove expensive reflection from raft/mesh hot path

### LINKED ISSUE #16551: Performance issues/raft thread saturation on mesh updates
#### Overview of the Issue

We were seeing performance issues in one of our mesh clusters that has about 5k participants, when doing deployments. The registration time would spike to 10+ seconds, which would cause errors and delays downstream. This corresponded to the raft/FSM thread being saturated for potentially minutes at a time. We seemed to have pretty low update rates overall, so this was surprising.

Screenshot 2023-03-07 at 11 44 06
Screenshot 2023-03-07 at 11 27 44
Screenshot 2023-03-07 at 11 27 49
Screenshot 2023-03-07 at 11 27 10

I grabbed a pprof to look in to it a bit more, but unfortunately a single routine on a busy server didn't give the greatest information. Still, reflection stood out here and it gave us somewhere to look.

Screenshot 2023-03-07 at 11 34 10

#### Reproduction Steps

To confirm what we were seeing here, I wrote a quick test case to run a snapshot of our data through this code path.

```go
func BenchmarkRealData(b *testing.B) {
	sfpath := os.Getenv("STATE_BIN_PATH")
	if sfpath == "" {
		b.Skipf("neet to set STATE_BIN_PATH env var")
	}

	snapfile, err := os.Open(sfpath)
	if err != nil {
		b.Fatalf("failed opening snap file %s: %v", sfpath, err)
	}

	store := state.NewStateStore(nil)
	var idx uint64 = 1

	handler := func(header *fsm.SnapshotHeader, msg structs.MessageType, dec *codec.Decoder) error {
		switch {
		case msg == structs.RegisterRequestType:
			var req structs.RegisterRequest
			if err := dec.Decode(&req); err != nil {
				return err
			}
			if err := store.EnsureRegistration(idx, &req); err != nil {
				return fmt.Errorf("ensuring registration: %v", err)
			}
		default:
			// need to read the message, to advance the decoder past it
			var val interface{}
			if err := dec.Decode(&val); err != nil {
				return fmt.Errorf("decoding message: %w", err)
			}
		}
		return nil
	}
	if err := fsm.ReadSnapshot(snapfile, handler); err != nil {
		b.Fatalf("Error reading snapshot %s: %v", sfpath, err)
	}
}
```

Which confirmed that most of the registration time is spent in reflection.

image

Consul 1.14

Linux, amd64.

lstoll:
### Description

Replaces a reflection-based copy of a struct in the mesh topology with a deep-copy generated implementation.

This is in the hot-path of raft FSM updates, and the reflection overhead was a substantial part of mesh registration times (~90%). This could manifest as raft thread saturation, and resulting instability.

This is a similar problem/fix as #14934, it might be worth seeing if there are any other uses of the reflection copy library and replacing it with (generated) copies.

Fixes #16551

Details in issue #16551 . The before/after from the test case in that issue:

Screenshot 2023-03-07 at 11 35 23

After the change was rolled out, we no longer saw any registration delays and the general raft/FSM thread saturation dropped off substantially

image

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


karlhungus:
@rboyer / @boxofrad  would it be possible to take a look at this? It is an identical improvement to the one made in https://github.com/hashicorp/consul/pull/14934, we've been running this for a few months now, and it'd be great if we could avoid patching consul ourselves.

agent/consul/state/deep-copy.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+readonly PACKAGE_DIR="$(dirname "${BASH_SOURCE[0]}")"
+cd $PACKAGE_DIR
+
+# Uses: https://github.com/globusdigital/deep-copy
+deep-copy \
+  -pointer-receiver \
+  -o ./catalog_schema.deepcopy.go \
```
boxofrad, line 9: nit: Is there a particular reason for this to be `catalog_schema`?
```suggestion
  -o ./catalog.deepcopy.go \
```

boxofrad (approved):
This is awesome, thanks @lstoll 🙌🏻  🔥

agent/consul/state/deep-copy.sh
boxofrad: Actually, ignore me! Just spotted the type is defined in `catalog_schema.go` 🙈

jmurret:
thank you @lstoll !  🎉

Issue #16580, opened 2023-03-09
Modifying Connect CA config attribute results in Envoy TLS connection failures


jbrandhorst:
### Overview of the issue

I am attempting to modify the connect CA attribute "CSRMaxPerSecond" from the default value of 50. Our clusters use vault as the connect CA, and Envoy as the proxy. Attempts to modify this setting have left the system in a state where any new registrants to the mesh after the config change are unable to establish TLS connections with peers. 

It is possible that I have the format/syntax incorrect for that settting, but I'll also note that re-applying the original ca config without the CSRMaxPerSecond line does not fix the condition. 

### Consul Info

consul version: 1.14.4
envoy version: 1.23.1

### Reproduction Steps

This is the process I have followed:

1. `consul connect ca get-config > config.json` to get the existing configuration
2. Remove the `"CreateIndex"` and `"ModifyIndex`" fields from the json file.
3. Insert a line into the `"Config"` block of the json file with `"CSRMaxPerSecond": 100`
4. Apply the new configuration with `consul connect ca set-config -config-file config.json`

This is the config.json file being set with the updated CSRMaxPerSecond setting:

```
{
	"Provider": "vault",
	"Config": {
            "Address": "https://vault.hostname.goes.here:8200",
            "CSRMaxPerSecond": 100.0,
            "IntermediateCertTTL": "8760h",
            "IntermediatePKIPath": "intermediate-path-goes-here",
            "LeafCertTTL": "72h",
            "RootPKIPath": "root-path-goes-here",
            "RotationPeriod": "2160h",
            "Token": "token-goes-here"
	},
	"State": null,
	"ForceWithoutCrossSigning": false
}
```

The only logs generated from the leader server when this is applied are:

`connect.ca: CA provider config updated` 
and 
`connect.ca.vault: Successfully renewed token for Vault provider`

From that point forward, any new registrants to the connect service mesh cannot establish TLS connections with other mesh participants, all connection attempts result in Envoy generating an error: 
`SSL routines:OPENSSL_internal:TLSV1_ALERT_UNKNOWN_CA`.  

I have pulled the Envoy configs from a working instance and a broken instance, and they have the same certificate chain info in the `common_tls_context -> validation_context -> certificate_chain -> inline_string` field.  Consul server and client agents don't generate any error logs from what I can see. 

This was attempted on two distinct consul/mesh clusters with the same results. All consul clusters are operated as a single datacenter.

### Rollback Steps

[...]

Issue #14564, opened 2022-09-12
Updating connect Vault CA configuration drops intermediate certificates


mr-miles:
When filing a bug, please include the following headings if possible. Any example text in this template can be deleted.

#### Overview of the Issue

Upgrading from 1.11 to 1.12 we observed that clients no longer had valid certificate chains after an auto-encrypt startup.  They were missing the intermediate certificate.

#### Reproduction Steps

Steps to reproduce this issue, eg:

1. Create a cluster with 1 client node and 1 server nodes
2. Configure the client with autoencrypt
3. Configure the server with the vault CA provider.  Set up the root and intermediate certificate paths.
4. Start the client and server
5. Observe that the TLS certificate chain is valid - e.g. visit https://<client>:8501/v1/agent/connect/ca/roots in the browser
6. PUT the identical vault ca config to https://<server>:8501/v1/connect/ca/configuration
7. Restart the client
8. Observe that the TLS certificate chain is now invalid and is missing the intermediate certificates - e.g. visit https://<client>:8501/v1/agent/connect/ca/roots in the browser (the text response still contains the intermediate but the cert chain for https does not)

### Consul info for both Client and Server
1.12,4 for both
Vault Connect CA with intermediate certificate

### Root cause
In leader_connect_ca.go:891 - primaryUpdateRootCA - the root is updated but the intermediates aren't, before a short-circuit check on line 921.  Comparing this code to that of leader_connect_ca.go:471 - primaryInitialize - indicates that GenerateIntermediate should be called but isn't.  It is currently called much later (line 992) which is after the early exit.  So effectively updating an existing vault CA config strips the intermediate certificates.

If I alter leader_connect_ca.go:897 to add intermediate handling before the early exit, the error I experience goes away:

```	newRootPEM := providerRoot.PEM
	newActiveRoot, err := newCARoot(newRootPEM, args.Config.Provider, args.Config.ClusterID)
	if err != nil {
		return err
	}

        # new code starts here
	interPEM, err := newProvider.GenerateIntermediate()
	if err != nil {
		return fmt.Errorf("error generating intermediate cert: %v", err)
	}

	// Add the local leaf signing cert to the rootCA struct. This handles both
	// upgrades of existing state, and new rootCA.
	if c.getLeafSigningCertFromRoot(newActiveRoot) != interPEM {
		newActiveRoot.IntermediateCerts = append(newActiveRoot.IntermediateCerts, interPEM)
	}
```

[...]

PR #16592, opened 2023-03-09
Preserve CARoots when updating Vault CA configuration

kisunji:
### Description

If a CA config update did not cause a root change, the codepath would return early and skip some steps which preserve its intermediate certificates and signing key ID. This commit re-orders some code and prevents updates from unnecessarily generating new intermediate certificates if one exists.

* Added new testcase which should fail without this PR
* Existing integration tests are passing

### Links
Fixes https://github.com/hashicorp/consul/issues/16580
Fixes https://github.com/hashicorp/consul/issues/14564

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/consul/leader_connect_ca.go
+++ b/agent/consul/leader_connect_ca.go
@@ -12,7 +12,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-hclog"
-	uuid "github.com/hashicorp/go-uuid"
+	"github.com/hashicorp/go-uuid"
 	"golang.org/x/time/rate"
 
 	"github.com/hashicorp/consul/acl"
@@ -887,6 +887,20 @@ func (c *CAManager) primaryUpdateRootCA(newProvider ca.Provider, args *structs.C
 		return err
 	}
 
+	// TODO: https://github.com/hashicorp/consul/issues/12386
+	intermediate, err := newProvider.ActiveIntermediate()
+	if err != nil {
+		return err
+	}
+	if intermediate == "" {
+		return fmt.Errorf("no active intermediate")
+	}
+	if intermediate != newRootPEM {
+		if err := setLeafSigningCert(newActiveRoot, intermediate); err != nil {
+			return err
+		}
+	}
+
 	// See if the provider needs to persist any state along with the config
 	pState, err := newProvider.State()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


agent/consul/leader_connect_ca.go
```diff
@@ -272,7 +272,7 @@ func newCARoot(pemValue, provider, clusterID string) (*structs.CARoot, error) {
 		ExternalTrustDomain: clusterID,
 		NotBefore:           primaryCert.NotBefore,
 		NotAfter:            primaryCert.NotAfter,
-		RootCert:            pemValue,
+		RootCert:            lib.EnsureTrailingNewline(pemValue),
```
kisunji, line 275: Older versions of Vault did not add a newline to the PEM file. `lib.EnsureTrailingNewline` is used in a few places in CA code to patch `\n` if missing.

My new test case revealed that the CARoot stored in CAManager did not have a trailing `\n`

agent/consul/leader_connect_ca.go
```diff
@@ -887,6 +887,23 @@ func (c *CAManager) primaryUpdateRootCA(newProvider ca.Provider, args *structs.C
 		return err
 	}
 
+	// TODO: https://github.com/hashicorp/consul/issues/12386
+	intermediate, err := newProvider.ActiveIntermediate()
```
kisunji, line 891: Should address https://github.com/hashicorp/consul/issues/11811

agent/consul/leader_connect_ca_test.go
```diff
@@ -612,39 +612,72 @@ func TestCAManager_UpdateConfiguration_Vault_Primary(t *testing.T) {
...
+		require.NoError(t, err)
 
-	_, newRoot, err := s1.fsm.State().CARootActive(nil)
-	require.NoError(t, err)
-	require.Len(t, newRoot.IntermediateCerts, 2,
-		"expected one cross-sign cert and one local leaf sign cert")
-	require.NotEqual(t, origRoot.ID, newRoot.ID)
+		_, newRoot, err := s1.fsm.State().CARootActive(nil)
+		require.NoError(t, err)
+		require.Len(t, newRoot.IntermediateCerts, 2,
```
eikenb, line 673: Does this fail if the #11811 fix isn't there? (ie. does this verify that fix?)

kisunji: This test (`update config and change root`) passes regardless of the fix because when the root and intermediate PKI paths are changed, the provider no longer has an `provider.ActiveIntermediate()` so it will always call `provider.GenerateIntermediate()`

kisunji: I just realized this PR doesn't resolve #11811 entirely, but it does prevent unnecessary calls to `GenerateIntermediate`: https://github.com/hashicorp/consul/pull/16592#discussion_r1131714901

rboyer (approved):

eikenb (approved):
I'm still not 100% on this code but I see no red flag and will give this my :+1: to move things along in case RB doesn't have time.

PR #16600, opened 2023-03-10
Add known issues to Raft WAL docs.

banks:
Adding this since we discovered bugs that are currently being fixed.


website/content/docs/agent/wal-logstore/index.mdx
```diff
@@ -45,4 +47,12 @@ The WAL backend has been tested thoroughly during development:
 
 We are aware of how complex and critical disk-persistence is for your data.
 
-We hope that many users at different scales will try WAL in their environments after upgrading to 1.15 or later and report success or failure so that we can confidently replace BoltDB as the default for new clusters in a future release.
\ No newline at end of file
+We hope that many users at different scales will try WAL in their environments after upgrading to 1.15 or later and report success or failure so that we can confidently replace BoltDB as the default for new clusters in a future release.
+
+## Known Issues
+
+The following issues were discovered after release of Consul 1.15.1. Fixes should be available in the next patch release.
```
david-yu, line 54:
```suggestion
The following issues were discovered after release of Consul 1.15.1. Fixes should be available in a future patch release.
```

david-yu: Will love to see this come in the next patch release, but I believe this depends on timing!

david-yu (approved):
Approved, with a comment about patch release version.

david-yu:
You may know this already, but if you create docs branches with a prefix  `docs/` for example `docs/insert-your-branch-name` it should skip go and integration tests.

website/content/docs/agent/wal-logstore/index.mdx
```diff
@@ -11,6 +11,8 @@ This topic provides an overview of the experimental WAL (write-ahead log) LogSto
 
 !> **Experimental feature:** The WAL LogStore backend is experimental.
 
+!> **Known Issues:** Consul 1.15.0 and 1.15.1 have [known issues](#known-issues) so should not have WAL enabled in production.
```
trujillo-adam, line 14:
```suggestion
!> **Known Issues:** Consul 1.15.0 and 1.15.1 have [known issues](#known-issues) so should not have WAL enabled in production.
```
We should move this to the usage pages associated with this functionality: https://developer.hashicorp.com/consul/docs/agent/wal-logstore/enable#requirements

Also, this callout and the *Experimental features* callout are inconsistent with our style guidance on usage: https://github.com/hashicorp/engineering-docs/blob/main/writing/style-guide.md#callout-boxes

website/content/docs/agent/wal-logstore/index.mdx
```diff
@@ -45,4 +47,12 @@ The WAL backend has been tested thoroughly during development:
...
\ No newline at end of file
+We hope that many users at different scales will try WAL in their environments after upgrading to 1.15 or later and report success or failure so that we can confidently replace BoltDB as the default for new clusters in a future release.
+
+## Known Issues
+
+The following issues were discovered after release of Consul 1.15.1. Fixes should be available in the next patch release.
+
+ * A follower that is disconnected for a while may be unable to catch up if it is using the WAL backend.
+ * Restoring user snapshots can break replication to WAL-enabled followers.
+ * Restoring user snapshots can cause a WAL-enabled leader to panic.
```
trujillo-adam, line 58:
```suggestion
## Known issues

The following issues were discovered after releasing Consul 1.15.1 and will be fixed in the next patch release:

 * A follower that is disconnected for a while may be unable to catch up if it is using the WAL backend.
 * Restoring user snapshots can break replication to WAL-enabled followers.
 * Restoring user snapshots can cause a WAL-enabled leader to panic.
```
The style guidance on headings recently changed. Use sentence-style capitalization for all titles and headings. 
We should also use concrete language, which is why I suggested changing "should" to "will" in the opening paragraph, even though the style guidance prohibits describing future functionality. As an experimental feature, this is probably good enough exception. Can we be more concrete about what we mean by "disconnected for a while"? What is a while? 10 seconds? 10 minutes? 10 days?

trujillo-adam (commented):
Made a few suggestions that are mostly style-related.

website/content/docs/agent/wal-logstore/index.mdx
banks: I agree that consecutive callouts isn't ideal but conveying both important notes in one didn't seem ideal either. I'm super open to alternative suggestions though!

I also considered putting this callout on the enable page too. I settled on here because this is the page we be most likely to link people to and until we have a fix I don't want to "bury the lead" that they shouldn't be using this yet several pages into their investigation.

trujillo-adam: The topic description sentence says that the feature is experimental, so I think the first callout is unnecessary and actually distracts the reader--especially right at the top of the page. The same goes for the second callout. 

My suggestion is that we get rid of that callouts and revise the introductory statement:

"This topic provides an overview of the WAL (write-ahead log) LogStore backend. The WAL backend is an experimental feature. Refer to [Requirements]([/consul/docs/agent/wal-logstore/enable#requirements](https://developer.hashicorp.com/consul/docs/agent/wal-logstore/enable#requirements)) for supported environments and known issues. "

We just need to be super judicious with how and when we use callouts. Most style guides consider them distractions but allow them only when absolutely necessary.

banks: Thanks I'll update as you proposed with some additional clarity on the level of risk being taken.

Personally, I'd always err on the side of being as obvious as possible when describing a feature that is not ready for production usage yet so there is no room for missing that if people are skimming. It's _meant_ to be a distraction in the sense that no reader should be able to look at the page and be in doubt about whether this is ready to use in production or not 😄 But I also get why two callouts together reduces the impact so I'll change it to match your suggestion. "Experimental" in the title is a clue, but there is "experimental" as in "the API might change one day" and "experimental" as in "you may loose all your data and be unable to recover" which is worth being clear about IMO!

banks:
@trujillo-adam thanks for your review. I think the latest push implements the changes you proposed. Please let me know if there's anything I missed so we can get the site updates ASAP.

If we know for sure when the fixes will be in I'll PR more specific language but I hope we can get this live today ahead of that.

trujillo-adam (approved):
👍

PR #16608, opened 2023-03-10
Add sameness-group configuration entry.

hashi-derek:
This PR introduces a new sameness-group configuration entry. Sameness-groups will allow for multiple partitions / peers to be grouped together for failover, service-export, and intention configuration.

Note that while the data structure exists in the open-source version of the software, all functionality is currently planned to be enterprise-only.


wilkermichael (approved):
Just compared this to your other PR, looks the same to me!

PR #16609, opened 2023-03-10
Backport of Allow HCP metrics collection for Envoy proxies into release/1.15.x

hc-github-team-consul-core:
## Backport

This PR is auto-generated from #16511 to be assessed for backporting due to the inclusion of the label backport/1.15.

**WARNING** automatic cherry-pick of commits failed. Commits will require human attention.

> merge conflict error: POST https://api.github.com/repos/hashicorp/consul/merges: 409 Merge conflict []

The below text is copied from the body of the original PR.

---

### Description
Add a new envoy flag: `envoy_hcp_metrics_bind_socket_dir`, a directory where a unix socket will be created with the name `<namespace>_<proxy_id>.sock` to forward Envoy metrics.
If set, this will configure a local `stats_sink` and `STATIC` cluster to forward Envoy metrics to an HCP collector.

The HCP Cloud team is working on enabling observability metrics in HCP. 

There will be a follow-up PR to generate a listener and cluster dynamically that will be listening on this unix socket. 
That listener and cluster will be configured to route to collector instances through the service mesh.

Caveat: we needed to add this local listener indirection as a workaround due to this [issue](https://github.com/envoyproxy/envoy/issues/25700)

### Testing & Reproduction steps

- Added unit tests
- Tested locally by running consul, registering a proxy with the flag configured and starting Envoy with consul connect command.

### PR Checklist

* [x] updated test coverage
* [x] external facing docs updated
* [x] not a security concern

---

<details>
<summary> Overview of commits </summary>

  - 8ff4b6fa03bb584f4bf638624bda31a58d4fb3be  - a5498b95a995cd65bec08f6ebc2a0e69725fa59a  - f0f7a3b17ff52cba7b699f2c30d3bcc7332297e0  - 28442ae9faf89d7b53ac035243f4a3b553f56ae6  - 4eb82e1994ef1b458a293a31fe7f6abf20d67cd0  - 59d43784a4a789c63fc3b131bac72591e0f79ffb 

</details>

PR #16611, opened 2023-03-10
Backport of Allow HCP metrics collection for Envoy proxies into release/1.15.x

freddygv:
Replaces #16609 since it did not pull the entire change set from #16511


hashi-derek (approved):

nathancoleman (approved):
Difference in diff size is due to some import changes that are not applicable to this base branch

PR #16622, opened 2023-03-13
Basic resource type registry

analogue:
### Description

See NET-2686

`gotestsum --format=testname ./internal/resource`

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/internal/resource/registry.go
+++ b/internal/resource/registry.go
@@ -2,6 +2,7 @@ package resource
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/hashicorp/consul/proto-public/pbresource"
 )
@@ -24,6 +25,7 @@ type Registration struct {
 // Resource type registry
 type TypeRegistry struct {
 	registrations map[*pbresource.Type]Registration
+	lock          sync.RWMutex
 }
 
 func NewRegistry() Registry {
@@ -33,6 +35,9 @@ func NewRegistry() Registry {
 }
 
 func (r *TypeRegistry) Register(registration Registration) {
+	r.lock.Lock()
+	defer r.lock.Unlock()
+
 	typ := registration.Type
 	if typ.Group == "" || typ.GroupVersion == "" || typ.Kind == "" {
 		panic("type field(s) cannot be empty")
@@ -46,6 +51,9 @@ func (r *TypeRegistry) Register(registration Registration) {
 }
 
 func (r *TypeRegistry) Resolve(typ *pbresource.Type) (reg Registration, ok bool) {
+	r.lock.RLock()
+	defer r.lock.RUnlock()
+
 	if registration, ok := r.registrations[typ]; ok {
 		return registration, true
 	}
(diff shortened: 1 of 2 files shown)


internal/resource/registry.go
```diff
@@ -0,0 +1,58 @@
...
+type Registration struct {
+	// Type is the GVK of the resource type.
+	Type *pbresource.Type
+
+	// In the future, we'll add hooks, the controller etc. here.
+}
+
+// Resource type registry
+type TypeRegistry struct {
+	registrations map[*pbresource.Type]Registration
```
boxofrad, line 26: Could we guard this with a `sync.RWMutex`, please? Theoretically, everything should be registered on-startup (via package init functions) but we've been burned by gRPC's thread-unsafe resolver and balancer registries.

internal/resource/registry.go
```diff
@@ -0,0 +1,58 @@
...
+	typ := registration.Type
+	if typ.Group == "" || typ.GroupVersion == "" || typ.Kind == "" {
+		panic("type field(s) cannot be empty")
+	}
+
+	if _, ok := r.registrations[registration.Type]; ok {
+		panic(fmt.Sprintf("resource type %s already registered", ToGVK(registration.Type)))
+	}
+
+	r.registrations[registration.Type] = registration
```
boxofrad, line 45: I think it might be unsafe to use `Type` as a map key, because it has a bunch of internal protobuf fields which would presumably change the hash value. Could we use the `ToGVK` string instead?

internal/resource/registry.go
```diff
@@ -0,0 +1,58 @@
...
+
+type Registry interface {
+	// Register the given resource type and its hooks.
+	Register(reg Registration)
+
+	// Resolve the given resource type and its hooks.
+	Resolve(typ *pbresource.Type) (reg Registration, ok bool)
+}
+
+type Registration struct {
```
boxofrad, line 17: We can definitely tackle this in a follow-up, but I've just been reflecting that we probably want to keep a reference to the protobuf message type on here (so that we can validate it on writes).

Protobufs has some prior art for this in its own [global type registry](https://github.com/golang/protobuf/blob/5d5e8c018a13017f9d5b8bf4fad64aaa42a87308/proto/registry.go).

boxofrad (changes requested):
This is looking great, nice work! I've left a handful of comments. Please feel free to ignore the one about protobuf message types for now.

internal/resource/registry.go
analogue: 👍 Fixed in https://github.com/hashicorp/consul/pull/16622/commits/66818c72de789b3305fbcec596602fdcb56d0a52

internal/resource/registry.go
analogue: 👍 Fixed in https://github.com/hashicorp/consul/pull/16622/commits/c5b3faa0c528d2154e42cd3f7ce88d88aa95535a

internal/resource/registry.go
analogue: Added TODO in https://github.com/hashicorp/consul/pull/16622/commits/83ff27e7618383ccfaf1c0b54fb3d5d52bf1eeb8

boxofrad (approved):

PR #16627, opened 2023-03-13
Update e2e tests for namespaces

jm96441n:
### Description

This updates the http_route e2e tests for namespaces and cleans up some previous errors from the last PR. Also patches a small bug with namespaces and `ResourceRefs` not using the mapstructure squash tag for the embedded `EnterpriseMeta` struct.

Run the tests

### Links

N/A

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


agent/structs/config_entry_status.go
```diff
@@ -22,7 +22,7 @@ type ResourceReference struct {
 	// unused, this should be blank.
 	SectionName string
 
-	acl.EnterpriseMeta
+	acl.EnterpriseMeta `mapstructure:",squash"`
```
nathancoleman, line 25: Should this also have the `hacl:",squash"` tag that's present on this field for `APIGatewayConfigEntry`, for example?

https://github.com/hashicorp/consul/blob/d5677e5680050a9b22403c160f5edde59613a4fc/agent/structs/config_entry_gateways.go#L713

test/integration/consul-container/test/gateways/http_route_test.go
```diff
@@ -79,17 +79,22 @@ func TestHTTPRouteFlattening(t *testing.T) {
 	)
 
 	namespace := getNamespace()
+	if namespace != "" {
+		ns := &api.Namespace{Name: namespace}
+		_, _, err := client.Namespaces().Create(ns, nil)
+		assert.NoError(t, err)
```
nathancoleman, line 85: I would use `require` here since the rest of the test is invalid if this assertion doesn't pass

```suggestion
		require.NoError(t, err)
```

test/integration/consul-container/test/gateways/http_route_test.go
```diff
@@ -189,23 +193,21 @@ func TestHTTPRouteFlattening(t *testing.T) {
 	require.NoError(t, cluster.ConfigEntryWrite(routeOne))
 	require.NoError(t, cluster.ConfigEntryWrite(routeTwo))
 
-	//create gateway service
-	gatewayService, err := libservice.NewGatewayService(context.Background(), gatewayName, "api", cluster.Agents[0], listenerPort)
+	// create gateway service
+	gatewayService, err := libservice.NewGatewayService(context.Background(), gatewayName, "api", namespace, cluster.Agents[0], listenerPort)
 	require.NoError(t, err)
-	libassert.CatalogServiceExists(t, client, gatewayName, nil)
+	libassert.CatalogServiceExists(t, client, gatewayName, &api.QueryOptions{Namespace: namespace})
```
nathancoleman, line 199: I _think_ this works okay if `namespace == ""`

test/integration/consul-container/test/gateways/http_route_test.go
```diff
@@ -277,25 +278,30 @@ func TestHTTPRoutePathRewrite(t *testing.T) {
 	)
 
 	namespace := getNamespace()
+	if namespace != "" {
+		ns := &api.Namespace{Name: namespace}
+		_, _, err := client.Namespaces().Create(ns, nil)
+		assert.NoError(t, err)
```
nathancoleman, line 284:
```suggestion
		require.NoError(t, err)
```

test/integration/consul-container/test/gateways/http_route_test.go
```diff
@@ -442,6 +445,11 @@ func TestHTTPRouteParentRefChange(t *testing.T) {
 
 	// getNamespace() should always return an empty string in Consul OSS
 	namespace := getNamespace()
+	if namespace != "" {
+		ns := &api.Namespace{Name: namespace}
+		_, _, err := client.Namespaces().Create(ns, nil)
+		assert.NoError(t, err)
```
nathancoleman, line 451:
```suggestion
		require.NoError(t, err)
```

test/integration/consul-container/test/gateways/http_route_test.go
```diff
@@ -540,8 +550,7 @@ func TestHTTPRouteParentRefChange(t *testing.T) {
 			{
 				Services: []api.HTTPService{
 					{
-						Name:      service.GetServiceName(),
-						Namespace: namespace,
```
nathancoleman, line 544: Was this causing issues? It looks like the gateway, route and service will all reside in the same namespace so this would be optional; however, I wouldn't expect it to cause a problem if set.

nathancoleman (commented):
Nice work! Some questions here for me

agent/structs/config_entry_status.go
jm96441n: good call, I'll add that

test/integration/consul-container/test/gateways/http_route_test.go
jm96441n: oh TIL that `assert` doesn't exit the test

test/integration/consul-container/test/gateways/http_route_test.go
jm96441n: yup, in OSS the `getNamespace` function always returns an empty string

test/integration/consul-container/test/gateways/http_route_test.go
jm96441n: so in the ENT tests I'm putting the gateway in a different namespace than the route/service (though talking about it, it would probably be better to have these tests always exercise the flows in the same namespace and add additional tests in ent for resources in different namespaces)

test/integration/consul-container/libs/service/connect.go
```diff
@@ -142,18 +142,24 @@ func (g ConnectContainer) GetStatus() (string, error) {
 	return state.Status, err
 }
 
+type SidecarConfig struct {
+	SidecarServiceName string
```
nathancoleman, line 146: Given the name of the struct, I would go with just `ServiceName` here to avoid stutter

nathancoleman (approved):

test/integration/consul-container/libs/service/connect.go
jm96441n: I was on the fence, I didn't want this to be confused for the name of the service that this is a sidecar for, maybe just `Name`?

nathancoleman: I don't have a very strong opinion here, just thought I'd call it out while also approving if you wanted to leave as-is

PR #16642, opened 2023-03-15
net 2731 ip config entry OSS version

wangxinyi7:
### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


jmurret (approved):
Double checked contents are the same.  Looks good.  👍

PR #16645, opened 2023-03-15
Resource service read endpoint

analogue:
### Description

See NET-2687. 

I branched off main and then merged in `boxofrad/storage-backend` so I could use the inmem storage backend for the tests.  Changes for this PR are entirely in `server.go` and `server_test.go`. 

`gotestsum --format testname ./agent/grpc-external/services/resource`

Let me know if I'm the right path with the tests regarding using the in-mem backend (more integration like) or if just mocking out the `registry` and `backend` is preferred which will allow testing of the "read consistent" variations.
 
### PR Checklist

* [x] updated test coverage
* [x] not a security concern


analogue:
Was looking into the testing part a bit more and noticed the pattern used by some of the other gRPC services. Specifically, making local interfaces in the same package that only define the subset of methods used by the actual "service under test". Then using mockery to generate a mock for those interface subsets and inject them through the server's config struct.  

Let me switch this back to a draft and I'll re-submit with the changes.

PR #16647, opened 2023-03-15
Snapshot restore tests

dhiaayachi:
### Description
This adds an integration test to cover the use case of taking and restoring a snapshot on a consul cluster, it also run this test case using both raft log store backends (`boltdb` and `wal`).

### DIFF
--- a/.changelog/16647.txt
+++ b/.changelog/16647.txt
@@ -1,4 +1,3 @@
-integration test to cover the use case of taking and restoring a snapshot on a consul cluster
 ```release-note:bug
-raft_logstore: Fixes a bug where restoring a snapshot with a WAL backend causes panic because raf-wal expects a monotonically increasing log index and the restore inserts a gap where no files exist.
+raft_logstore: Fixes a bug where restoring a snapshot when using the experimental WAL storage backend causes a panic.
 ```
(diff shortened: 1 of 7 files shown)


banks:
With this change I pushed, this test passes for me now if I use these two branches locally:

https://github.com/hashicorp/raft-wal/pull/24
https://github.com/hashicorp/raft/pull/545

And this in go.mod:

```
	github.com/hashicorp/raft => ../raft
	github.com/hashicorp/raft-wal => ../raft-wal
```

.changelog/16647.txt
```diff
@@ -0,0 +1,4 @@
+integration test to cover the use case of taking and restoring a snapshot on a consul cluster
```
banks, line 1: Do we normally add testing items to the changelog? I'm not even quite sure what happens to text outside of the release note.

.changelog/16647.txt
```diff
@@ -0,0 +1,4 @@
...
+```release-note:bug
+raft_logstore: Fixes a bug where restoring a snapshot with a WAL backend causes panic because raf-wal expects a monotonically increasing log index and the restore inserts a gap where no files exist.
```
banks, line 3:
```suggestion
raft_logstore: Fixes a bug where restoring a snapshot when using the experimental WAL storage backend causes a panic.
```
nits really, but see what you think

banks (approved):
I ran this new integration test locally as is and also adding 

```
	b.conf.Set("raft_logstore.verification.enabled", true)
	b.conf.Set("raft_logstore.verification.interval", "5s")
```

As a one-time smoke test that it still works with the verifier too (don't think it's worth having that extra test case permenently though).

So I say good to go!

.changelog/16647.txt
jmurret: yea, that was a remnant that was not meant to be there.  I removed it.

.changelog/16647.txt
jmurret: I've updated this.  Thanks for the suggestion.

PR #16649, opened 2023-03-16
Add validation for apigw creation with no routes

missylbytes:
### Description

To meet the standard ([Gateways require at least one listener defined](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.Gateway)) we need to add checks in to verify that at least one listener is defined.

```
make dev-build 
consul agent -dev
```
Open another terminal, create a file called gateway.hcl with the following content:
```
Kind = "api-gateway"
Name = "my-gateway"
```
After creating the file, run the following command in the same terminal:
```
consul config write gateway.hcl
```
This should error out, saying "api gateway must have at least one listener"

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/agent/structs/config_entry_gateways.go
+++ b/agent/structs/config_entry_gateways.go
@@ -1,6 +1,7 @@
 package structs
 
 import (
+	"errors"
 	"fmt"
 	"regexp"
 	"sort"
@@ -769,6 +770,9 @@ func (e *APIGatewayConfigEntry) Validate() error {
 		return err
 	}
 
+	if len(e.Listeners) == 0 {
+		return errors.New("api gateway must have at least one listener")
+	}
 	if err := e.validateListenerNames(); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


agent/structs/config_entry_gateways.go
```diff
@@ -769,6 +770,9 @@ func (e *APIGatewayConfigEntry) Validate() error {
 		return err
 	}
 
+	if len(e.Listeners) == 0 {
+		return errors.New("api gateway must have at least one listener")
```
jm96441n, line 774: minor (and more stylistic than anything and not blocking), would we want to use a sentinel error here so callers could act on the error type in the future?

missylbytes: So I don't feel like that matches the way the other validations are handled, but I'm also realizing I didn't follow that way either. Ex: https://github.com/hashicorp/consul/blob/main/agent/structs/config_entry_gateways.go#L803

missylbytes: Especially because I think it sends a 500 error back right now

t-eckert (approved):
This is great! Would you mind just formatting the `.mdx` tables you added so the vertical bars align?

mikemorris (approved):

PR #16655, opened 2023-03-16
Read(...) endpoint for the resource service

analogue:
### Description

See NET-2687.  I swiped `internal/storage/storage.go` from the latest `boxofrad/storage-backend` branch since it has necessary dependencies.

`gotestsum --format testname ./agent/grpc-external/services/resource` 

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/grpc-external/services/resource/read.go
```diff
@@ -0,0 +1,54 @@
...
+	if !ok {
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("resource type %s not registered", resource.ToGVK(req.Id.Type)))
+	}
+
+	readFn := s.backend.Read
+	if isConsistentRead(ctx) {
+		readFn = s.backend.ReadConsistent
+	}
+
+	resource, err := readFn(ctx, req.Id)
```
boxofrad, line 29: So sorry, but I've just broken this API in #16538 😅 

It should now be:
```suggestion
	consistency := storage.EventualConsistency
	if isConsistentRead(ctx) {
		consistency = storage.StrongConsistency
	}

	resource, err := s.backend.Read(ctx, consistency, req.Id)
```

agent/grpc-external/services/resource/read.go
```diff
@@ -0,0 +1,54 @@
...
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+
+	"github.com/hashicorp/consul/internal/resource"
+	"github.com/hashicorp/consul/internal/storage"
+	"github.com/hashicorp/consul/proto-public/pbresource"
+)
+
+func (s *Server) Read(ctx context.Context, req *pbresource.ReadRequest) (*pbresource.ReadResponse, error) {
+
```
boxofrad, line 17: Super nit-pick 🐜 would you mind nixing this trailing newline? We don't tend to have them at the beginning of function bodies.

agent/grpc-external/services/resource/read_test.go
```diff
@@ -0,0 +1,161 @@
...
+	mock "github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+func TestRead_TypeNotFound(t *testing.T) {
+	mockRegistry := &MockRegistry{}
+	mockRegistry.On("Resolve", mock.Anything).Return(resource.Registration{}, false)
```
boxofrad, line 19: Is there a reason we can't just use the real registry here? (and in the below test cases too)

agent/grpc-external/services/resource/read.go
boxofrad, line 1: Thanks for pulling this out into its own file. That's much better!

boxofrad (commented):
This looks great! 👏🏻  As there's still some churn going on in the storage backend interface, let's maybe hold off on merging until #16538 is merged.

agent/grpc-external/services/resource/read.go
analogue: Fixed

agent/grpc-external/services/resource/read.go
analogue: 👍

agent/grpc-external/services/resource/read_test.go
analogue: Updated all tests to not use a mock Registry or Backend since `consistency` can now be asserted against.

analogue:
Fixed all outstanding issues.

agent/grpc-external/services/resource/read.go
```diff
@@ -0,0 +1,53 @@
...
+	resource, err := s.backend.Read(ctx, consistency, req.Id)
+	if err != nil {
+		if err == storage.ErrNotFound {
+			return nil, status.Error(codes.NotFound, err.Error())
+		}
+		if _, ok := err.(storage.GroupVersionMismatchError); ok {
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+		return nil, err
```
boxofrad, line 36: tip: it's probably worth using [`errors.Is`](https://pkg.go.dev/errors#Is) and [`errors.As`](https://pkg.go.dev/errors#As) here in case backends [wrap the error](https://go.dev/blog/go1.13-errors).

agent/grpc-external/services/resource/read_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/hashicorp/consul/internal/resource"
+	storage "github.com/hashicorp/consul/internal/storage"
+	"github.com/hashicorp/consul/internal/storage/inmem"
+	pbresource "github.com/hashicorp/consul/proto-public/pbresource"
+	"github.com/hashicorp/consul/proto/private/prototest"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
```
boxofrad, line 16: Lots of folks on the team have their editor set up to split imports into stdlib, project, and 3rd party groups. It's what you get if you run `goimports -local github.com/hashicorp/consul` but I think there's an equivalent [snippet](https://stackoverflow.com/questions/38704769/golang-import-grouping-by-package) for VS Code and GoLand.

I don't think it's a hard rule, but it'll reduce some diff noise for folks that work on this file later. 😄 

```suggestion
import (
	"context"
	"testing"

	"github.com/hashicorp/consul/internal/resource"
	"github.com/hashicorp/consul/internal/storage"
	"github.com/hashicorp/consul/internal/storage/inmem"
	"github.com/hashicorp/consul/proto-public/pbresource"
	"github.com/hashicorp/consul/proto/private/prototest"
	
	"github.com/stretchr/testify/require"
	"google.golang.org/grpc/codes"
	"google.golang.org/grpc/metadata"
	"google.golang.org/grpc/status"
)
```

boxofrad (commented):
Awesome work! I've left a couple more small optional comments, but this is good to go as soon as the storage backend is merged 🎉

agent/grpc-external/services/resource/read_test.go
analogue: I ran `goimports` as described and got
```
import (
        context "context"
        "testing"

        mock "github.com/stretchr/testify/mock"
        "github.com/stretchr/testify/require"
        "google.golang.org/grpc/codes"
        "google.golang.org/grpc/metadata"
        "google.golang.org/grpc/status"

        "github.com/hashicorp/consul/internal/resource"
        storage "github.com/hashicorp/consul/internal/storage"
        "github.com/hashicorp/consul/internal/storage/inmem"
        pbresource "github.com/hashicorp/consul/proto-public/pbresource"
        "github.com/hashicorp/consul/proto/private/prototest"
)
```

Ignoring the duplicate aliases, it put `github.com/hashicorp/consul` imports at the end instead of the middle. I don't see any additional flags on `goimports` to control this behavior.

agent/grpc-external/services/resource/read.go
analogue: TIL! Fixed in https://github.com/hashicorp/consul/pull/16655/commits/d5cd617e19421a2097ed1d404cf497acbb790583

agent/grpc-external/services/resource/read_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+			rsp, err := client.Read(tc.ctx, &pbresource.ReadRequest{Id: id1})
+			require.NoError(t, err)
+			prototest.AssertDeepEqual(t, resource1, rsp.Resource)
+		})
+	}
+}
+
+type readTestCase struct {
+	consistency storage.ReadConsistency
```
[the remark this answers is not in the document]
analogue: Ah! This is leftover from the hybrid use of the non-mock registry and the mock backend. I added a test as suggested in https://github.com/hashicorp/consul/pull/16655/commits/5d8dffb57d961af73a68f90803d0a14267d94187.

agent/grpc-external/services/resource/server.go
```diff
@@ -19,6 +19,11 @@ type Config struct {
 	backend  storage.Backend
 }
 
+//go:generate mockery --name Backend --inpackage
+type Backend interface {
+	storage.Backend
+}
+
```
boxofrad: This is really clever btw! I've always struggled to figure out how to generate mocks for interfaces in other packages 😍

boxofrad (approved):

PR #16661, opened 2023-03-17
[API Gateway] Fix invalid cluster causing gateway programming delay

andrewstucki:
### Description

We were seeing that HTTP listeners on API Gateways were taking ~15 seconds or more to get programmed onto the envoy instances (this also happens on our Kubernetes offering). I finally tracked down what was the root cause. Because we are currently masquerading as ingress gateways for the purposes of xDS generation, the main entrypoint for each HTTP listener is a "virtual" service router, basically one that doesn't correspond to an actual service, but that contains our routing logic based on the rules in an HTTPRoute. What was happening is that when we use such a router, the discovery chain compilation process adds a "fallback target" for when no rules match that targets a service resolver with the same name as the router.

Thus, if our router was named `gateway-listener-12345` -- we were generating a route and resolver node for `gateway-listener-12345`. The problem is that our xDS code then generates cluster targets for those services, attempting to resolve `gateway-listener-12345.default.default...` or some such hostname. Because we specify that we'll ship the resolved endpoints via EDS over ADS, envoy then blocks on applying its configuration, keeping the `gateway-listener-12345` cluster in a warming state. This, in turn, keeps the whole listener in a warming state until the time at which the EDS config source's `initial_fetch_timeout` value is hit. Since we weren't setting that, this defaults to 15 seconds. So, any use of a virtual router causes an ~15 second delay in applying envoy configuration.

The temporary fix with our native API gateway until we move to fully native xDS generation is to munge the compiled discovery chain that we synthesize for the API gateway and filter out both the router's default route and the resolver node for the router.

Running [this script](https://gist.github.com/andrewstucki/fce7d28e6544a727970a173fca37ecd3) shows these timings.

Before the fix:

```bash
➜  ./timer.sh
Config entry written: proxy-defaults/global
Config entry written: api-gateway/gateway
Config entry written: http-route/router
Registered service: service
gateway not listening yet
...
gateway listening

real	0m20.379s
user	0m3.641s
sys	0m15.512s
```

And after:

```bash
➜  ./timer.sh
Config entry written: proxy-defaults/global
Config entry written: api-gateway/gateway
Config entry written: http-route/router
Registered service: service
gateway not listening yet
...
gateway listening

real	0m0.734s
user	0m0.137s
sys	0m0.561s
```

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/agent/consul/discoverychain/gateway.go
+++ b/agent/consul/discoverychain/gateway.go
@@ -128,6 +128,29 @@ func (l *GatewayChainSynthesizer) Synthesize(chains ...*structs.CompiledDiscover
 			return nil, nil, err
 		}
 
+		node := compiled.Nodes[compiled.StartNode]
+		if node.IsRouter() {
+			resolverPrefix := structs.DiscoveryGraphNodeTypeResolver + ":" + node.Name
+
+			// clean out the clusters that will get added for the router
+			for name := range compiled.Nodes {
+				if strings.HasPrefix(name, resolverPrefix) {
+					delete(compiled.Nodes, name)
+				}
+			}
+
+			// clean out the route rules that'll get added for the router
+			filtered := []*structs.DiscoveryRoute{}
+			for _, route := range node.Routes {
+				if strings.HasPrefix(route.NextNode, resolverPrefix) {
+					continue
+				}
+				filtered = append(filtered, route)
+			}
+			node.Routes = filtered
+		}
+		compiled.Nodes[compiled.StartNode] = node
+
 		// fix up the nodes for the terminal targets to either be a splitter or resolver if there is no splitter present
 		for name, node := range compiled.Nodes {
 			switch node.Type {
--- a/agent/xds/resources_test.go
+++ b/agent/xds/resources_test.go
@@ -11,6 +11,8 @@ import (
 	envoy_route_v3 "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
 	envoy_tls_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3"
 
+	"github.com/hashicorp/consul/agent/connect"
+	"github.com/hashicorp/consul/agent/consul/discoverychain"
 	"github.com/hashicorp/consul/agent/xds/testcommon"
 	"github.com/hashicorp/consul/envoyextensions/xdscommon"
 
@@ -175,7 +177,7 @@ func TestAllResourcesFromSnapshot(t *testing.T) {
 	tests = append(tests, getMeshGatewayPeeringGoldenTestCases()...)
 	tests = append(tests, getTrafficControlPeeringGoldenTestCases()...)
 	tests = append(tests, getEnterpriseGoldenTestCases()...)
-	tests = append(tests, getAPIGatewayGoldenTestCases()...)
+	tests = append(tests, getAPIGatewayGoldenTestCases(t)...)
 
 	latestEnvoyVersion := xdscommon.EnvoyVersions[0]
 	for _, envoyVersion := range xdscommon.EnvoyVersions {
@@ -314,7 +316,13 @@ AAJAMaoXmoYVdgXV+CPuBb2M4XCpuzLu3bcA2PXm5ipSyIgntMKwXV7r
 -----END CERTIFICATE-----`
 )
 
-func getAPIGatewayGoldenTestCases() []goldenTestCase {
+func getAPIGatewayGoldenTestCases(t *testing.T) []goldenTestCase {
+	t.Helper()
+
+	service := structs.NewServiceName("service", nil)
+	serviceUID := proxycfg.NewUpstreamIDFromServiceName(service)
+	serviceChain := discoverychain.TestCompileConfigEntries(t, "service", "default", "default", "dc1", connect.TestClusterID+".consul", nil)
+
 	return []goldenTestCase{
 		{
 			name: "api-gateway-with-tcp-route-and-inline-certificate",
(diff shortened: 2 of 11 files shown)


andrewstucki:
@jm96441n / @sarahalsmiller / @mikemorris -- the shortened time for propagating configuration wound up unmasking that the go-based integration tests were quite flaky to the point of almost never passing, likely due to a combination of port collisions and `r.Fail`/`require` statements inside of our `Eventually` loops.

I refactored a good chunk of them to ensure consistency since they were using a whole host of different ways to construct clusters, services, and gateways, a few different ways of doing retry assertions, and at least one test was being fully skipped.

Can y'all take a look to sign off on this? Thanks!

agent/consul/discoverychain/gateway.go
```diff
@@ -128,6 +128,29 @@ func (l *GatewayChainSynthesizer) Synthesize(chains ...*structs.CompiledDiscover
 			return nil, nil, err
 		}
 
+		node := compiled.Nodes[compiled.StartNode]
```
mikemorris, line 131: The problematic router is always/only the first entry in the chain?

agent/consul/discoverychain/gateway.go
```diff
@@ -128,6 +128,29 @@ func (l *GatewayChainSynthesizer) Synthesize(chains ...*structs.CompiledDiscover
...
+
+			// clean out the route rules that'll get added for the router
+			filtered := []*structs.DiscoveryRoute{}
+			for _, route := range node.Routes {
+				if strings.HasPrefix(route.NextNode, resolverPrefix) {
+					continue
+				}
+				filtered = append(filtered, route)
+			}
+			node.Routes = filtered
```
mikemorris, line 150: So what this is doing is dropping any route (specifically the default "catch-all" mentioned in https://developer.hashicorp.com/consul/docs/connect/config-entries/service-router#path-prefix-matching) that would route to the problematic generated resolver?

In "clean out the clusters" above, is what's happening removing the generated resolver node (with it's corresponding compiled clusters) from the discovery chain?

agent/xds/resources_test.go
```diff
@@ -314,7 +316,13 @@ AAJAMaoXmoYVdgXV+CPuBb2M4XCpuzLu3bcA2PXm5ipSyIgntMKwXV7r
 -----END CERTIFICATE-----`
 )
 
-func getAPIGatewayGoldenTestCases() []goldenTestCase {
+func getAPIGatewayGoldenTestCases(t *testing.T) []goldenTestCase {
+	t.Helper()
+
+	service := structs.NewServiceName("service", nil)
+	serviceUID := proxycfg.NewUpstreamIDFromServiceName(service)
+	serviceChain := discoverychain.TestCompileConfigEntries(t, "service", "default", "default", "dc1", connect.TestClusterID+".consul", nil)
```
mikemorris, line 324: The argument list for this feels unfortunately confusing and could be more clear as fields on a single struct arg, but assuming that's out of scope for this change.

agent/xds/testdata/routes/api-gateway-with-http-route-and-inline-certificate.latest.golden
```diff
@@ -0,0 +1,31 @@
...
+        {
+          "name":  "api-gateway-listener-9b9265b",
+          "domains":  [
+            "*",
+            "*:8080"
+          ],
+          "routes":  [
+            {
+              "match":  {
+                "prefix":  "/"
```
mikemorris, line 17: Had to double-check, but this is the expected behavior for defining a rule with only a Service backend (no matches or filters) 👍

test/integration/consul-container/test/gateways/gateway_endpoint_test.go
```diff
@@ -63,32 +76,48 @@ func TestAPIGatewayCreate(t *testing.T) {
 
 	require.NoError(t, cluster.ConfigEntryWrite(apiGateway))
 
+	_, _, err := libservice.CreateAndRegisterStaticServerAndSidecar(cluster.Agents[0], &libservice.ServiceOpts{
+		ID:        serviceName,
+		Name:      serviceName,
+		Namespace: namespace,
+		HTTPPort:  serviceHTTPPort,
+		GRPCPort:  serviceGRPCPort,
```
mikemorris, line 84: Are these actually needed/being checked in this test? It looks like it was originally simpler, just verifying gateway creation?

test/integration/consul-container/test/gateways/gateway_endpoint_test.go
mikemorris, line 1: Most of the changes in here are to ensure that a boolean is always returned in `require.Eventually` calls rather than having any assertions that could fail the test.

test/integration/consul-container/test/gateways/http_route_test.go
```diff
@@ -445,17 +503,43 @@ func TestHTTPRouteParentRefChange(t *testing.T) {
...
 
 	// infrastructure set up
 	address := "localhost"
 
-	listenerOnePort := 6000
-	listenerTwoPort := 6001
+	listenerOnePort := 6014
+	listenerTwoPort := 6015
+	serviceHTTPPort := 6016
+	serviceGRPCPort := 6017
```
mikemorris, line 515: Could we use the magic behavior of `0` (preferably, but guessing maybe not possible) or perhaps https://github.com/hashicorp/consul/tree/main/sdk/freeport to automatically select a port for these, rather than manually avoiding overlap between tests?

mikemorris (commented):
Left a few post-merge minor notes, but thanks for tracking this down, this is a big improvement!

From discussing this, we don't want to make a orresponding change to ingress gateways because it would be a breaking change to a default behavior of `service-router`. Instead, our Kubernetes implementation of API Gateway will pick up these changes when we eventually migrate from the Ingress gateway backend in Consul to the native API Gateway backend here.

agent/consul/discoverychain/gateway.go
andrewstucki: Yes, in fact you can only have a single router in the entire chain since routers can only route to splitters/resolvers, and splitters can only route to splitters/resolvers -- so the only place you can use a router is in the very first part of a discovery chain.

Regardless though, we only ever create a single router as the first node in our synthetic chain.

agent/consul/discoverychain/gateway.go
andrewstucki: Yeah, so this is removing a route that goes to a resolver matching the name of the router -- which is the catch-all route.

The other step is removing the same resolver from the list of nodes in the full discovery chain, which is what is used to generate the cluster.

The node removal is necessary to keep this from blocking configuration push down, the route removal is to keep envoy from trying to resolve some random hostname and having a bogus routing rule.

agent/xds/resources_test.go
andrewstucki: Yeah, this is primarily just copying from what other tests are doing -- many of these xDS golden file generation tests are like this and are fairly opaque until you've poked around a good bit.

test/integration/consul-container/test/gateways/gateway_endpoint_test.go
andrewstucki: So, I moved these out to ensure nothing was causing port collisions somewhere. Previously it had pinned the ports to 8080 and 8081 which made running multiple services (i.e. the flatenning code) problematic. With the way that the services get run, the port collision actually was causing the services to get routed improperly because they're competing for the same binding port. I'm assuming the tests were passing in a flaky way before due to sometimes one proxy starting up quicker than another.

test/integration/consul-container/test/gateways/http_route_test.go
andrewstucki: Unfortunately not, these aren't actually ports we're binding on the host, they're ports that are bound inside the context of a container and then hole-punched to. We might be able to work with the assumption that the ports are random enough when using `freeport` (also since they're in the ephemeral range) that they're unlikely to be bound inside the container context as well, but I'd prefer to test that out some other time.

PR #16676, opened 2023-03-20
Update WAL Known issues

banks:
Update for the known issues for WAL to go out as they will be fixed in 1.15.2 this week.


website/content/docs/agent/wal-logstore/enable.mdx
```diff
@@ -34,8 +34,9 @@ We recommend taking the following additional measures:
 
 ## Known issues
 
-The following issues were discovered after release of Consul 1.15.1 and will be
-fixed in a future patch release.
+There are no known issues for the WAL backend in Consul 1.15.2 and higher.
```
trujillo-adam, line 37:
```suggestion
There are no known issues for the WAL backend in Consul 1.15.2 and higher.
```
I think my preference would be to remove this line and just note the versions that _do have_ an issue--call out the exceptions and not the rule. Also, there's no guarantee that there won't be issues later on, so "and higher" might not be future proof.

trujillo-adam (approved):

PR #16677, opened 2023-03-20
Fix route subscription when using namespaces

jm96441n:
### Description

We’re improperly subscribing to our routes in the proxycfg package, leveraging the proxy enterprise metadata rather than the route metadata. This causes the gateway to never get configured when it binds a route in another namespace. This PR fixes that by ensuring the correct EnterpriseMeta is used to reference the namespace

In consul enterprise set up a two namespaces, a gateway in one namespace, a service in the other, and attempt to setup a httproute in the service namespace. The route will never register without this fix (also automated tests showing this are coming up in a follow up PR).

### Links

N/A

### PR Checklist

* [ ] updated test coverage
* [X] not a security concern


.changelog/_16677.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+gateway: Fix bug where routes defined in a different namespace than a gateway would fail to register. [[GH-16677](https://github.com/hashicorp/consul/pull/16677)].
```
missylbytes, line 2: Does this need an enterprise only tag? (I don't actually know if that is a thing in the changelog, just wanted to point it out)

mikemorris: 👍 We do commonly use a **`(Enterprise only)`** tag in the changelog when appropriate, see https://github.com/hashicorp/consul/blob/main/CHANGELOG.md#1135-december-13-2022 for an example.

mikemorris:
```suggestion
gateway: **(Enterprise only)** Fix bug where routes defined in a different namespace than a gateway would fail to register. [[GH-16677](https://github.com/hashicorp/consul/pull/16677)].
```

mikemorris (approved):
Would be preferable to include tests alongside this fix, but understand that it would complicate the backport because of drift, and the actual scope of the testing update is enterprise-only.

agent/proxycfg/api_gateway.go
```diff
@@ -391,7 +392,7 @@ func (h *handlerAPIGateway) handleRouteConfigUpdate(ctx context.Context, u Updat
 		snap.APIGateway.Upstreams.set(ref, listener, set)
 	}
 	snap.APIGateway.UpstreamsSet.set(ref, seenUpstreamIDs)
-	//snap.APIGateway.Hosts = TODO
+	// snap.APIGateway.Hosts = TODO
 	snap.APIGateway.AreHostsSet = true
```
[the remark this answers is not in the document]
jm96441n: https://github.com/hashicorp/consul/pull/16691

PR #16690, opened 2023-03-20
delete config when nil

dhiaayachi:
### Description

This add the ability for multi-limiter to delete a configuration.

For example the following sequence:
- `m.UpdateConfig(c1,"test.ip")`
- `m.DeleteConfig("test.ip")`
Will result to "test.ip" not having any configuration attached to it. The reconciler will delete any limiter associated with that configuration. 

Also this PR remove the concept of defaultConfig as this is not useful and add extra complexity related to deleting config.

The need for this is to be able to delete an IP based config entry for IP rate limiting and subsequently delete all the rate limiter associated with that configuration.

Added test case to cover `m.DeleteConfig("test.ip")` use case

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/consul/multilimiter/multilimiter_test.go
```diff
@@ -279,8 +298,29 @@ func TestRateLimiterUpdateConfig(t *testing.T) {
...
+		l, ok := m.limiters.Load().Get(ip)
+		require.True(t, ok)
+		require.NotNil(t, l)
+		limiter := l.(*Limiter)
+		require.True(t, c1.isApplied(limiter.limiter))
+		m.DeleteConfig(prefix)
+		reconcile(m)
+		_, ok = m.limiters.Load().Get(ip)
+		require.False(t, ok)
+	})
```
boxofrad, line 321: nit: would you mind adding separating out the blocks with newlines to make this easier to follow? 🙇🏻‍♂️

boxofrad (approved):
LGTM

wangxinyi7:
Sorry @dhiaayachi for my lateness, but I just noticed that the `defaultConfg` at here 
https://github.com/hashicorp/consul/blob/4845816c602a883074537330ccdcebf99531cf96/agent/consul/multilimiter/multilimiter.go#L43
was not deleted, which is supposed to be deleted right?

PR #16691, opened 2023-03-20
Remove unused are hosts set check

jm96441n:
### Description

Follow up PR from https://github.com/hashicorp/consul/pull/16677#discussion_r1142437422, this handles removing the check for `AreHostsSet`

N/A

### Links

https://github.com/hashicorp/consul/pull/16677#discussion_r1142437422

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


andrewstucki (commented):
@jm96441n can you also just go ahead and remove the field from the struct itself since this makes it completely unused? Thanks!

agent/proxycfg/snapshot.go
```diff
@@ -706,12 +708,11 @@ type configSnapshotAPIGateway struct {
 	BoundGatewayConfig       *structs.BoundAPIGatewayConfigEntry
 
 	// Hosts is the list of extra host entries to add to our leaf cert's DNS SANs
-	Hosts       []string
-	AreHostsSet bool
+	Hosts []string
```
andrewstucki, line 711: Given the above docstring and the fact that we don't actually add any SANs to the leaf cert, I think this is unused as well, no? Can you confirm and drop as well?

jm96441n: can do

agent/proxycfg/api_gateway.go
```diff
@@ -393,7 +393,6 @@ func (h *handlerAPIGateway) handleRouteConfigUpdate(ctx context.Context, u Updat
 	}
 	snap.APIGateway.UpstreamsSet.set(ref, seenUpstreamIDs)
 	// snap.APIGateway.Hosts = TODO
```
nathancoleman, line 395:
```suggestion
```

nathancoleman (approved):

PR #16701, opened 2023-03-20
Regen helm docs

lkysow:
Generated by running `make gen-helm-docs`


david-yu (approved):

david-yu:
Does this need to be backported to 1.15. at least or would that be a separate PR?

PR #16702, opened 2023-03-20
Fix 'device' typo in intro doc

timsutton:
### Description

One-character typo fix in the docs I noticed when reading the intro.


trujillo-adam (approved):
Thanks for fixing this!

PR #16712, opened 2023-03-21
add extra resiliency to snapshot restore test

dhiaayachi:
### Description

This add a check before reading the KV entries from the new cluster (restored from a snapshot) in the snapshot restore test. The extra check would verify that the follower is caught up with the leader before reading logs as we would like to read those KV entry performing a stale read from the follower to ensure that the logs were replicated to the follower correctly.

This is to avoid some flakes observed in CI as [this](https://app.circleci.com/pipelines/github/hashicorp/consul/54329/workflows/b15eff84-ef76-48db-ace3-5cbf5ab14c79/jobs/1284077/tests#failed-test-1).


jmurret (approved):
👍 thank you!

PR #16726, opened 2023-03-21
WatchList(..) endpoint for the resource service

analogue:
### Description

See NET-2692. Branched off of the latest `boxofrad/storage-backend` and made only the necessary changes to implement the `WatchList` endpoint. 

`gotestsum --format testname ./agent/grpc-external/services/resource `

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/grpc-external/services/resource/watch.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (s *Server) WatchList(req *pbresource.WatchListRequest, stream pbresource.ResourceService_WatchListServer) error {
+	// check type exists
+	_, ok := s.registry.Resolve(req.Type)
+	if !ok {
+		return status.Error(
+			codes.InvalidArgument,
+			fmt.Sprintf("resource type %s not registered", resource.ToGVK(req.Type)),
+		)
+	}
```
boxofrad, line 32: As we'll probably end up doing this check in all endpoints, it might be worth adding a helper:

```Go
if _, err := s.resolveType(req.Type); err != nil {
	return err
}
```

```Go
func (s *Server) resolveType(typ *pbresource.Type) (*resource.Registration, error) {
	v, ok := s.registry.Resolve(typ)
	if ok {
		return v, nil
	}
	return nil, status.Errorf(
		codes.InvalidArgument,
		"resource type %s not registered", resource.ToGVK(typ),
	)
}
```

agent/grpc-external/services/resource/watch.go
```diff
@@ -0,0 +1,64 @@
...
+	unversionedType := storage.UnversionedTypeFrom(req.Type)
+	watch, err := s.backend.WatchList(
+		stream.Context(),
+		unversionedType,
+		req.Tenancy,
+		req.NamePrefix,
+	)
+	if err != nil {
+		return err
+	}
```
boxofrad, line 43: Reminder to myself 😄 

When #16619 is merged, we'll need to update this to handle `ErrWatchClosed` (which I've just noticed is in the wrong package too).

agent/grpc-external/services/resource/watch_test.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+func TestWatchList_GroupVersionMatches(t *testing.T) {
+	ctx := testContext(t)
+	registry := resource.NewRegistry()
+	registry.Register(resource.Registration{Type: typev1})
+	backend, err := inmem.NewBackend()
+	require.NoError(t, err)
+	go backend.Run(ctx)
+	server := NewServer(Config{registry: registry, backend: backend})
```
boxofrad, line 44: What do you think of moving the server construction into a helper method?

```Go
func testServer(t *testing.T) *Server {
	t.Helper()

	backend, err := inmem.NewBackend()
	require.NoError(t, err)
	go backend.Run(testContext(t))

	registry := resource.NewRegistry()
	return NewServer(Config{registry: registry, backend: backend})
}
```

agent/grpc-external/services/resource/watch_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+func mustGetNoResource(t *testing.T, ch <-chan resourceOrError) {
+	t.Helper()
+
+	select {
+	case rsp := <-ch:
+		require.NoError(t, rsp.err)
+		require.Nil(t, rsp.rsp, "expected nil response with no error")
+	case <-time.After(1 * time.Second):
+		// Dan: Any way to do this without the annoying delay?
```
boxofrad, line 126: Nope, sorry! 🙈

We can probably make this delay shorter (like 250ms maybe) or make it configurable using [`-short` flag](https://pkg.go.dev/testing#Short).

It might also help to mark all of these tests as [parallel](https://pkg.go.dev/testing#T.Parallel) to avoid it holding up the suite.

internal/resource/registry.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/hashicorp/consul/proto-public/pbresource"
+)
+
+type Registry interface {
+	// Register the given resource type and its hooks.
+	Register(reg Registration)
+
+	// Resolve the given resource type and its hooks.
+	Resolve(typ *pbresource.Type) (reg Registration, ok bool)
+}
```
boxofrad, line 16: Ah, guessing you needed to include these files because they're not present in my branch? Sorry!

boxofrad (commented):
Looking awesome as usual. Nice job 👏🏻

agent/grpc-external/services/resource/watch.go
```diff
@@ -0,0 +1,64 @@
...
+func watchListOpFrom(eventOp pbresource.WatchEvent_Operation) pbresource.WatchListResponse_Operation {
+	switch eventOp {
+	case pbresource.WatchEvent_OPERATION_UPSERT:
+		return pbresource.WatchListResponse_OPERATION_UPSERT
+	case pbresource.WatchEvent_OPERATION_DELETE:
+		return pbresource.WatchListResponse_OPERATION_DELETE
+	default:
+		panic(fmt.Sprintf("unhandled op %s", eventOp.String()))
+	}
+}
```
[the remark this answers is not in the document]
analogue: 👍 Fixed in https://github.com/hashicorp/consul/pull/16726/commits/dc0f42b389483c356ec070201f50457c3a62c15a

agent/grpc-external/services/resource/watch.go
analogue: Good call! Was waiting for some of the other endpoints to come together but we can do it now. Fixed in https://github.com/hashicorp/consul/pull/16726/commits/7aaffecd6cf378fc95d6797a1ed9afa1f3d6601f

agent/grpc-external/services/resource/watch_test.go
analogue: 👍 Fixed in https://github.com/hashicorp/consul/pull/16726/commits/ed4564807dbcb090eed4b2ddf3db9768ac698171

agent/grpc-external/services/resource/watch.go
```diff
@@ -0,0 +1,43 @@
...
+
+		// drop versions that don't match
+		if event.Resource.Id.Type.GroupVersion != req.Type.GroupVersion {
+			continue
+		}
+
+		if err = stream.Send(&pbresource.WatchEvent{
+			Operation: event.Operation,
+			Resource:  event.Resource,
+		}); err != nil {
```
boxofrad, line 39: super nit 🐜 I think you could simplify this to just:
```suggestion
		if err = stream.Send(event); err != nil {
```

boxofrad (commented):
:shipit: ⭐

agent/grpc-external/services/resource/watch.go
analogue: 👍 Fixed in https://github.com/hashicorp/consul/pull/16726/commits/28caf485d629db1c91c4654e9d16a2d01b12813d

analogue:
Ok, moved some test helper methods and variables to `server_test.go` so they're shared by the resource service endpoint tests.

boxofrad (approved):

PR #16731, opened 2023-03-21
[NET-3029] Migrate frontend to GHA

loshz:
Signed-off-by: Dan Bond <danbond@protonmail.com>

## Description
This PR migrates the CircleCI `frontend` jobs into GitHub Actions. It also simplifies some of the individual steps:
- Don't upload artifacts and share between builds (this is a waste of resources in GHA)
- Don't split build/test jobs - do them together to save resources.
- Don't split OSS/ENT jobs - set an env var specifically in ENT to enable feature tests.
- Install Node via actions instead of using Docker.

| CircleCI job | GitHub Actions Job | Notes |
| --- | --- | --- |
| frontend-cache | N/A | No longer needed as we don't share artifacts across jobs. |
| workspace-tests | workspace-tests | Migrated as is. |
| node-tests | node-tests | Migrated as is. |
| ember-build-oss | ember-build-test | Merged into single job as commands are identical across OSS/ENT. |
| ember-build-ent | ember-build-test | Merged into single job as commands are identical across OSS/ENT. |
| ember-test-oss | ember-build-test | Merged into single job as commands are identical across OSS/ENT. |
| ember-test-ent | ember-build-test | Merged into single job as commands are identical across OSS/ENT. |
| ember-coverage | ember-build-test | Job added as a single command in ember-build-test. |


.github/workflows/frontend.yml
```diff
@@ -0,0 +1,72 @@
...
+
+name: frontend
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+      - ui/**
+      - backport/**
```
jmurret, line 12:
```suggestion
      - backport/ui/**
```

.github/workflows/frontend.yml
```diff
@@ -0,0 +1,72 @@
...
+
+    # Install dependencies.
+    - name: install yarn packages
+      working-directory: ui
+      run: make deps
+
+    - run: make test-node
+      working-directory: ui/packages/consul-ui
+
+  ember-build-test:
```
jmurret, line 49: Have we confirmed the following?
- we don't need to run OSS and ENT versions of these tests?
- just running the yarn test coverage out of [these existing CI tests](https://github.com/hashicorp/consul/blob/main/.circleci/config.yml#L800-L857) is ok?  IE, we are no long running the ember exam command.

If this is ok with the UI team, can you update the PR description to call this out, please?

jmurret (changes requested):
Nice!  branch filter just needs to be changed. also left a question about test consolidation.

.github/workflows/frontend.yml
loshz: >we don't need to run OSS and ENT versions of these tests?

These jobs were duplicated and the only difference was the value of `CONSUL_NSPACES_ENABLED`, which can be controlled independently in the ENT repo

>IE, we are no long running the ember exam command

We are still running the exam command here: https://github.com/hashicorp/consul/pull/16731/files#diff-ffe3a297ed15e9a79846161a3731a580511bf13c6b5cc5a9e03ddcbd8515fdeeR69

jmurret (approved):

jmurret (approved):
👍 nice work!

PR #16736, opened 2023-03-22
docs: Usage for exported services configuration entry

boruszak:
### Description

The upcoming HCP Consul Cluster Peering updates include a number of UI updates. Some of these updates require links to the regular Consul documentation.

When both clusters are HCP-managed, the HCP platform can detect whether the cluster runs Consul or Consul on Kubernetes and provide a link to the required documentation for that runtime. However, in situations where users connect HCP-managed and self-managed clusters, situations may occur where both runtimes are used and the HCP platform will not be able to detect which runtime's documentation to link to.

This PR makes several edits to provide a landing point for users to receive runtime-agnostic guidance. 

### Links

- [Direct link to deployment preview](https://consul-bjm50nu1n-hashicorp.vercel.app/consul/docs/connect/config-entries/exported-services#export-services-between-partitions-or-clusters)

### PR Checklist

* [ ] updated test coverage
* [X] not a security concern

### DIFF
--- a/website/content/docs/connect/config-entries/exported-services.mdx
+++ b/website/content/docs/connect/config-entries/exported-services.mdx
@@ -18,15 +18,8 @@ To configure Consul to export services contained in a Consul Enterprise admin pa
 You can configure the settings defined in the `exported-services` configuration entry to apply to all namespaces in a Consul Enterprise admin partition.
 
 ## Requirements
+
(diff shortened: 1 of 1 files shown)


trujillo-adam (approved):
Just a few minor suggestions that we can address later as we refactor content according to information types. Approving so you're not blocked.

PR #16739, opened 2023-03-22
Adding key words for better google results for index.mdx

ramramhariram:
currently these are hidden in our Consul enterprise page which is not necessarily what everyone is searching for, i see customers/users searching for things like 'compatibiity matrix' or 'feature matrix' to pull this kind of info for consul and I think adding in 'matrix' will help with google searches.

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/website/content/docs/enterprise/index.mdx
+++ b/website/content/docs/enterprise/index.mdx
@@ -61,7 +61,7 @@ to the Consul Enterprise binary that grants access to the desired features.
 
 Contact your [HashiCorp Support contact](https://support.hashicorp.com/) for a development license.
 
-## Consul Enterprise Feature Availability matrix 
+## Consul Enterprise feature availability matrix ((#consul-enterprise-feature-availability))
 
 The Consul Enterprise features that are available depend on your license and the runtimes you use in your deployment.
 
@@ -88,7 +88,7 @@ Available Enterprise features per Consul form and license include:
 [HashiCorp Cloud Platform (HCP) Consul]: https://cloud.hashicorp.com/products/consul
 [Consul Enterprise]: https://www.hashicorp.com/products/consul/
 
-### Feature availability matrix by runtime
+### Feature availability matrix by runtime ((#feature-availability-by-runtime))
 
 Consul Enterprise feature availability can change depending on your server and client agent runtimes.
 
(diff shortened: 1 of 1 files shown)


im2nguyen (approved):
thank you!

jkirschner-hashicorp:
@boruszak : If we want to improve search results that include "matrix" in the query, what's the right way to accomplish that? Would it be altering the page metadata rather than specific header titles?

Note: If we did change the header titles, we should consider adding anchor aliases to we don't break existing links to these sections.

im2nguyen:
Good call, we could definitely add "Evaluate Enteprise features with the feature availability and compatibility matrix." to the metadata (description) so it appears in the the devdot search. Updating the header only will let folks do a "control + f" on the page.

Google doesn't treat headers differently for SEO [source](https://www.searchenginejournal.com/heading-tags-for-seo/341817/)... but if "compatibility matrix" is what folks are searching for, if it appears on the page, it would rank higher.

website/content/docs/enterprise/index.mdx
```diff
@@ -88,7 +88,7 @@ Available Enterprise features per Consul form and license include:
 [HashiCorp Cloud Platform (HCP) Consul]: https://cloud.hashicorp.com/products/consul
 [Consul Enterprise]: https://www.hashicorp.com/products/consul/
 
-### Feature availability by runtime
+### Feature availability matrix by runtime
```
im2nguyen, line 91:
```suggestion
### Feature availability matrix by runtime ((#feature-availability-by-runtime))
```

ramramhariram:
hey @boruszak -  my initial comments are about the specific content that i proposed changes for. The feature matrix that i updated specifically addresses questions like  - "Do Consul EKS servers support clients on ECS"? That is not an enterprise feature question, in fact it is not even about any specific feature, but rather a question about compatibility. How do we drive answers to those kind of questions to this page? Currently any search around 'compatibility' etc goes to our k8s version compatibility or envoy/consul tables because those pages have those keywords in them.

ramramhariram:
yes, and maybe the best option is to have one page called compatibility matrix that houses ALL of compatibility matrices together.

PR #16740, opened 2023-03-22
fix: remove unused tenancy category from rate limit spec

JadhavPoonam:
### Description

@dhiaayachi  noticed `Tenancy` category is not referred anywhere which prompted me to look into this. 

The only place I found tenancy referred is in gRPC `ResourceService`. Resource category is already added to the relevant proto file and is included in the `OperationCategory` struct hence we don't need Tenancy.

### Links

Resource proto file: https://github.com/hashicorp/consul-enterprise/blob/main/proto-public/pbresource/resource.proto

### PR Checklist

* [ ] ~updated test coverage~


dhiaayachi (approved):
LGTM!
Thank you for adding this @JadhavPoonam

PR #16745, opened 2023-03-22
Expand route flattening test for multiple namespaces

jm96441n:
### Description
 Expands the route flattening test to encompass multiple namespaces.
 
 Also adds new helper to just check that a route config entry exists without checking it's bound status.

N/A

### Links

N/A

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


test/integration/consul-container/test/gateways/gateway_endpoint_test.go
```diff
@@ -179,6 +179,25 @@ func checkHTTPRouteConfigEntry(t *testing.T, client *api.Client, routeName strin
 	}, time.Second*10, time.Second*1)
 }
 
+func checkHTTPRouteConfigEntryWithoutBindStatus(t *testing.T, client *api.Client, routeName string, namespace string) {
```
missylbytes, line 182: I don't think this actually gets called in the tests, is that expected/is this going to help another PR?

jm96441n: it doesn't here, it's used in an ENT PR for cross namespace tests, but figured it should go into OSS first so it's ported to ENT

missylbytes (approved):

test/integration/consul-container/test/gateways/gateway_endpoint_test.go
sarahalsmiller, line 182: I'd probably update the name to something like, checkConfigEntryExists, but otherwise looks good to me.

sarahalsmiller (approved):

test/integration/consul-container/test/gateways/gateway_endpoint_test.go
jm96441n: I like that name a whoooole lot better

PR #16746, opened 2023-03-22
log warning about CA expiring sooner

eikenb:
The old setting of 24 hours was not enough time to deal with an expiring CA. Changing it to 30 days to give a reasonable amount of time to do the rotation.

### DIFF
--- a/agent/consul/leader_metrics.go
+++ b/agent/consul/leader_metrics.go
@@ -125,21 +125,25 @@ func (m CertExpirationMonitor) Monitor(ctx context.Context) error {
 		}
 
 		if expiresSoon(lifetime, untilAfter) {
-			key := strings.Join(m.Key, "")
-			rootKey := strings.Join(metricsKeyMeshRootCAExpiry, "")
-			signingKey := strings.Join(metricsKeyMeshActiveSigningCAExpiry, "")
+			key := strings.Join(m.Key, ":")
 			switch key {
-			case rootKey:
+			case "mesh:active-root-ca:expiry":
 				logger.Warn("root certificate will expire soon",
 					"time_to_expiry", untilAfter,
 					"expiration", time.Now().Add(untilAfter),
-					"suggested_action", "rotate the root certificate",
+					"suggested_action", "manually rotate the root certificate",
 				)
-			case signingKey:
+			case "mesh:active-signing-ca:expiry":
 				logger.Warn("signing (intermediate) certificate will expire soon",
 					"time_to_expiry", untilAfter,
 					"expiration", time.Now().Add(untilAfter),
 					"suggested_action", "check consul logs for rotation issues",
+				)
+			case "agent:tls:cert:expiry":
+				logger.Warn("agent TLS certificate will expire soon",
+					"time_to_expiry", untilAfter,
+					"expiration", time.Now().Add(untilAfter),
+					"suggested_action", "manually rotate this agent's certificate",
 				)
 			}
 		}
--- a/agent/consul/leader_metrics.go
+++ b/agent/consul/leader_metrics.go
@@ -118,29 +118,29 @@ func (m CertExpirationMonitor) Monitor(ctx context.Context) error {
 	logger := m.Logger.With("metric", strings.Join(m.Key, "."))
 
 	emitMetric := func() {
-		lifetime, notAfter, err := m.Query()
+		lifetime, untilAfter, err := m.Query()
 		if err != nil {
 			logger.Warn("failed to emit certificate expiry metric", "error", err)
 			return
 		}
 
-		if expiresSoon(lifetime, notAfter) {
+		if expiresSoon(lifetime, untilAfter) {
 			key := strings.Join(m.Key, "")
 			rootKey := strings.Join(metricsKeyMeshRootCAExpiry, "")
 			signingKey := strings.Join(metricsKeyMeshActiveSigningCAExpiry, "")
 			switch key {
 			case rootKey:
 				logger.Warn("root certificate will expire soon",
-					"time_to_expiry", notAfter,
-					"expiration", time.Now().Add(notAfter))
+					"time_to_expiry", untilAfter,
+					"expiration", time.Now().Add(untilAfter))
 			case signingKey:
 				logger.Warn("signing (intermediate) certificate will expire soon",
-					"time_to_expiry", notAfter,
-					"expiration", time.Now().Add(notAfter))
+					"time_to_expiry", untilAfter,
+					"expiration", time.Now().Add(untilAfter))
 			}
 		}
 
-		expiry := notAfter / time.Second
+		expiry := untilAfter / time.Second
 		metrics.SetGaugeWithLabels(m.Key, float32(expiry), m.Labels)
 	}
 
@@ -174,7 +174,7 @@ func initLeaderMetrics() {
 // we should send out a WARN log message.
 // It defaults to returning true if the cert will expire within 28 days or 40%
 // of the certs total duration, whichever is shorter.
-func expiresSoon(lifetime, notAfter time.Duration) bool {
+func expiresSoon(lifetime, untilAfter time.Duration) bool {
 	defaultPeriod := 28 * (24 * time.Hour) // 28 days
 	fortyPercent := (lifetime / 10) * 4    // 40% of total duration
 
@@ -183,5 +183,5 @@ func expiresSoon(lifetime, notAfter time.Duration) bool {
 		warningPeriod = fortyPercent
 	}
 
-	return notAfter < warningPeriod
+	return untilAfter < warningPeriod
 }
--- a/agent/consul/leader_metrics.go
+++ b/agent/consul/leader_metrics.go
@@ -104,8 +104,10 @@ type CertExpirationMonitor struct {
 	// then the metrics will expire before they are emitted again.
 	Labels []metrics.Label
 	Logger hclog.Logger
-	// Query is called at each interval. It should return the duration until the
-	// certificate expires, or an error if the query failed.
+	// Query is called at each interval. It should return 2 durations, the full
+	// lifespan of the certificate (NotBefore -> NotAfter) and the duration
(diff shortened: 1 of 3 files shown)


eikenb:
Chatted with @jkirschner-hashicorp and we decided to look into some  additions if possible..
- can we make it warn based on % of remaining cert time (to handle cases of CA's with 30 days or less duration)?
- can we differentiate the log message between the root cert and the intermediate cert?

eikenb:
Updated with the time calculations for when to send out the log messages. Still need to differentiate between the 2 certs log messages.

agent/consul/leader_metrics.go
```diff
@@ -151,3 +152,17 @@ func initLeaderMetrics() {
...
 }
+
+// shouldWarn checks to see if we are close enough to the cert expiring that we
+// should send out a WARN log message.
+// It defaults to returning true if the cert will expire within 28 days or 40%
+// of the certs total duration, whichever is shorter.
+func expiresSoon(before, after time.Duration) bool {
+	warningPeriod := 28 * (24 * time.Hour) // 28 days (default)
+	seventyDays := 70 * (24 * time.Hour)   // 28 days is 40% of 70 days
+	totalDuration := time.Duration(before + after)
```
jkirschner-hashicorp, line 163: I had a hard time following this logic because I didn't realize that `before` was actually `timeSinceLifetimeStart` and `after` is actually `timeUntilLifetimeEnd`, which together are equal to `totalLifetime`. Renaming the arguments might help.

agent/consul/leader_metrics.go
```diff
@@ -151,3 +152,17 @@ func initLeaderMetrics() {
...
+	if totalDuration < seventyDays {
+		warningPeriod = (totalDuration * 4) / 10 // 40% of total duration
+	}
+	return after < warningPeriod
+}
```
jkirschner-hashicorp, line 168: In the current implementation, the 40% assumption is baked into two lines of code rather than defined in one place. Here's my attempt at reworking it slightly to define in one place (including naming changes from my other comment).

```suggestion
func expiresSoon(timeSinceLifetimeStart, timeUntilLifetimeEnd time.Duration) bool {
	warningPeriod = 28 * (24 * time.Hour) // 28 days (default) 
	
	totalLifetime := time.Duration(timeSinceLifetimeStart + timeUntilLifetimeEnd)
	warningPeriodByTotalLifetime = (totalLifetime * 4) / 10 // 40% of total duration
	if warningPeriodByTotalLifetime < warningPeriod {
		warningPeriod = warningPeriodByTotalLifetime 
	}
	
	return timeUntilLifetimeEnd < warningPeriod
```

I defer to you/eng on whether to make sure a change / whether it helps with clarity!

jkirschner-hashicorp (commented):
Thanks John :) added a few comments for your + eng reviewer's consideration

agent/consul/leader_metrics.go
eikenb: They are named NotBefore and NotAfter in the Cert itself, which is where I derived the name from. I could expand the names to add the not to make that more clear. That naming should be very familiar to anyone who works with Certs.

agent/consul/leader_metrics.go
eikenb: I see what you're shooting for, trying to make the 2 periods more explicit. I'll think about this one.

agent/consul/leader_metrics.go
eikenb: To expand a bit.. The names previously used are not good. They are too long and, importantly, don't reflect the field name in the Cert that they represent. I do think adding the not will help though and will do that.

jkirschner-hashicorp: I agree that `NotBefore` and `NotAfter` are clear names. That said, they don't reflect the value of what's being passed in. Both of those fields are timestamps, whereas what's being passed into this function are durations relative to the current time.

For example, when I read the test case that set `before: (day * 50), after: (day * 20),` I thought it was conveying a cert that was always expired (because `NotBefore > NotAfter`). But in actuality, it's intending to express a cert that is 50/70 through its lifetime.

eikenb: TLDR; How about NotAfter and Lifetime as the 2 variables instead?

I get what you mean but IMO the importance of maintaining coherence with the Certificate names is more important. An alternative would be to change the Query's signature to carry the total time and the NotAfter time. NotAfter is required for the checks but NotBefore isn't, it is just a convenient way to get the missing data to calculate the total lifetime (as we already have NotAfter) but the relevant bit is the total duration not the start time.

eikenb: Trying out the new names.. it does make the tests more readable for sure.

eikenb: I like the new names and have pushed them up. Let me know if you think this helps with the readability. Thanks!

eikenb:
I've pushed up the log message changes. They aren't much at this point but I'm not sure what would be any better. 

@cizh.. any thoughts?

agent/consul/leader_metrics.go
```diff
@@ -116,7 +116,8 @@ func (m CertExpirationMonitor) Monitor(ctx context.Context) error {
 			return
 		}
 
-		if d < 24*time.Hour {
+		thirtyDays := 30 * (24 * time.Hour)
+		if d < thirtyDays {
 			logger.Warn("certificate will expire soon",
```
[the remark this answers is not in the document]
eikenb: Spoke with Cindy and I'm going to tweak the expiresSoon() function to return the number of days remaining along with the boolean so it can be used in the log message.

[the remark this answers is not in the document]
eikenb: After looking into this I remembered the days remaining was already there. It is part of the structured logging at the end, it has both "time_to_expiry" (how long the cert has left) and "expiration" (the date it will expire). So this is covered.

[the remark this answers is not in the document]
jkirschner-hashicorp: @eikenb : Have you considered @cizh 's point 3 about suggesting a course of action?

For the root cert, the operator must handle the rotation themselves.

If the intermediate (signing) cert is emitting that message, something is wrong with the automated rotation and they should inspect their logs for other messages that may be related.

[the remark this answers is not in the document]
eikenb: Took a shot at those messages. Added them as an additional 'suggested_action' entry in the structured log seection.

agent/consul/leader_metrics.go
```diff
@@ -153,3 +177,19 @@ func initLeaderMetrics() {
...
+func expiresSoon(lifetime, untilAfter time.Duration) bool {
+	defaultPeriod := 28 * (24 * time.Hour) // 28 days
+	fortyPercent := (lifetime / 10) * 4    // 40% of total duration
+
+	warningPeriod := defaultPeriod
+	if fortyPercent < defaultPeriod {
+		warningPeriod = fortyPercent
+	}
+
+	return untilAfter < warningPeriod
```
cthain, line 194: I think we could avoid the conditional assignment by returning:

```suggestion
	return untilAfter < defaultPeriod || untilAfter < fortyPercent
```

cthain (approved):
LGTM. A couple of non-blocking comments.

agent/consul/leader_metrics.go
eikenb: Interesting. I haven't seen this pattern used much in Go code, where it seems more at home in something like Python. The usual concern with these things is that it adds to the cognitive load as you have more to parse on the 1 line. But here I think it lines up pretty well with the OR rule in the comment and the function is short enough that the small bump in complexity isn't an issue.

TLDR; :+1: I like it.

agent/consul/leader_metrics.go
```diff
@@ -153,3 +179,14 @@ func initLeaderMetrics() {
...
+
+// expiresSoon checks to see if we are close enough to the cert expiring that
+// we should send out a WARN log message.
+// It returns true if the cert will expire within 28 days or 40% of the
+// certificate's total duration (whichever is shorter).
+func expiresSoon(lifetime, untilAfter time.Duration) bool {
+	defaultPeriod := 28 * (24 * time.Hour) // 28 days
+	fortyPercent := (lifetime / 10) * 4    // 40% of total duration
+
+	return untilAfter < defaultPeriod || untilAfter < fortyPercent
```
jkirschner-hashicorp, line 191: Is this logically equivalent to the intended behavior, which is `untilAfter < min(defaultPeriod, fortyPercent)`?

Let's say this is for an intermediate CA with a 30 day TTL. Automatic renewal will be first attempted at 50% lifetime (`untilAfter = 15 days`).

Let's say that `untilAfter` is currently 16 days. `fortyPercent` should be 12 days.

`untilAfter < defaultPeriod` -> `16 days < 28 days` -> `true`. The function returns `true`.

The intended behavior is:
`untilAfter < min(defaultPeriod, fortyPercent)` -> `16 days < min(28 days, 12 days)` -> `16 days < 12 days` -> `false`

jkirschner-hashicorp (changes requested):
I'm concerned that recent refactors may no longer reflect the originally intended logic (refer to my comment).

I may be misreading things, feel free to tell me otherwise and I'll dismiss my review!

agent/consul/leader_metrics.go
jkirschner-hashicorp: I haven't examined the test cases to check whether it would have caught this. (Maybe there already is such a test case, and there's just something off in my reasoning above.)

eikenb: You are 100% right and this is why I :heart: tests. This made sense on the surface but was slightly different and off... and the tests caught it.

eikenb:
> I'm concerned that recent refactors may no longer reflect the originally intended logic (refer to my comment).
> 
> I may be misreading things, feel free to tell me otherwise and I'll dismiss my review!

FYI I reverted that change as it was failing the tests. That undid the change but didn't clear your request change. I'm going to dismiss it but wanted you to know it is fixed.

PR #16748, opened 2023-03-22
refactor build workflows to contain universal logic for oss v. ent builds

modrake:
### Description

As part of our migration effort off of circleci, this adds in workflow logic that detects whether we are working in our OSS or enterprise repos in order to select from the appropriate github runner pools. This allows us to maintain the same workflow code in both repositories, easing automerging between OSS and ent.

### DIFF
--- a/.github/workflows/build-distros.yml
+++ b/.github/workflows/build-distros.yml
@@ -8,8 +8,25 @@ permissions:
   contents: read
 
 jobs:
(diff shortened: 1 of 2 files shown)


jmurret:
Thank you for your help on this.  I've moved this into [my existing PR](https://github.com/hashicorp/consul/pull/16761) since it is needed and also minimize the syncing and backporting.  Going to close this PR.

PR #16752, opened 2023-03-23
storage/raft: transparently forward writes and consistent reads to the leader

boxofrad:
### Description

Builds on #16619 to implement transparent leader forwarding for writes and consistent reads, authenticated using the same TLS certificate method as our regular Raft traffic.

As discussed in our sync call today, this uses the old pattern of terminating TLS in `Server.handleConn` rather than having the gRPC server do it, which means adding a new magic type-byte (`pool.RPCRaftForwarding`) 🪄.

It would've been nicer to move the TLS termination to the gRPC layer, put the forwarding service on the existing internal gRPC server, and checked the certificate in the handler. Unfortunately, it isn't possible to do this in a backwards-compatible way because it means moving the gRPC magic byte out of the TLS envelope. As @rboyer pointed out, we could use ALPN for this instead, but it'd mean handling non-TLS traffic separately 🤷🏻.


proto/private/pbstorage/raft.proto
```diff
@@ -1,25 +1,103 @@
...
-  enum Op {
-    OP_UNSPECIFIED = 0;
-    OP_WRITE = 1;
-    OP_DELETE = 2;
+// Forwarding service is used for forwarding write and consistent read
+// operations to the Raft leader. It is served on Consul's multiplexed
+// server port, which is the same port used for regular Raft traffic.
+service Forwarding {
+  // RoundTrip handles the forwarded operation.
+  rpc RoundTrip(Request) returns (Response) {
```
mkeeler, line 16: What are the tradeoffs of the single `RoundTrip` RPC approach vs something like:

```
service Forwarding {
   rpc ForwardRead(ReadRequest) returns (ReadResponse) ;
   rpc ForwardList(ListRequest) returns (ListResponse) ;
   rpc ForwardWrite(WriteRequest) returns (WriteResponse) ;
   rpc ForwardDelete(DeleteRequest) returns (DeleteResponse) ;
}
```

I can see that the expanded RPC form will have more methods to implement, however it looks like in internal/storage/raft/forwarding.go you ended up with a big switch statement and all the individual methods anyways. 

The big advantage I see to the expanded form is that there is no implicit contract regarding sending one type of request and getting back a specific type of response. With the expanded RPC approach the valid request/response type are a part of the RPC definition and the contents have no variability to deal with and ensure the consistency of within the code.

agent/grpc-internal/listener.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"fmt"
+	"net"
+)
+
+// PipeListener implements the net.Listener interface and allows you to manually
+// pass connections to it. This is useful when you need to accept connections
+// and do something with them yourself first (e.g. handling our multiplexing
+// scheme) before giving them to the gRPC server.
+type PipeListener struct {
```
mkeeler, line 12: bikeshed:

Would `ConnChannelListener` be more descriptive here?

Seeing `Pipe` in the name gave me the impression this would be using `net.Pipe` similar to what I had done previously here: https://github.com/mkeeler/resource-api-poc/blob/main/pkg/grpc-pipe/listener.go.

agent/grpc-internal/listener.go
```diff
@@ -0,0 +1,60 @@
...
+
+}
+
+// Handle makes the given connection available to Accept.
+func (l *PipeListener) Handle(conn net.Conn) {
+	select {
+	case l.conns <- conn:
+	case <-l.done:
+	}
+}
```
mkeeler, line 36: As this could (at least in theory) block indefinitely if nothing is calling `Accept` on the listener, I think it would be advantageous to add a `context.Context` in the signature and to allow for this to return an error indicating the conn was not successfully processed.

I suspect this would trickle down to other RPC multiplex handling code passing in a new context with some timeout and handling the error by logging the timeout.

agent/grpc-internal/listener.go
mkeeler: Looks like in the previous iteration this was called chanListener in the other file so maybe there is some nuance here I am missing.

mkeeler (approved):
The few comments I left aren't really blocking if you feel that things should remain the same. So I will go ahead and approve.

If you do make changes and want another review feel free to re-request it.

proto/private/pbstorage/raft.proto
boxofrad: I'd originally gone with the single endpoint for symmetry with the `Backend.raftApply` method (previously called `roundTrip`) which needs a wrapping message and type enum for the Raft log entries.

I think you're right that the expanded form makes more sense for forwarding, though, as without it we're sort of implementing an RPC system on top of an RPC system 😅

analogue (approved):
Again, nice work! 🚀

agent/grpc-internal/listener.go
boxofrad: By the way, the code you linked to is **super** helpful 🙌🏻  thank you!

agent/consul/raft_handle.go
```diff
@@ -26,4 +31,49 @@ func (h *raftHandle) Apply(msg []byte) (any, error) {
...
+	if leaderAddr == "" {
+		return nil, errors.New("leader unknown")
+	}
+
+	dc := h.s.config.Datacenter
+	tlsCfg := h.s.tlsConfigurator
+
+	return grpc.Dial(string(leaderAddr),
+		// TLS is handled in the dialer below.
+		grpc.WithInsecure(),
```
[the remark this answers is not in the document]
boxofrad: Good catch, thanks!

agent/grpc-internal/listener.go
boxofrad: As this is already in the `grpc-internal` package, I've gone with just `Listener` as we don't really need a more descriptive name 😄

mkeeler (approved):

PR #16756, opened 2023-03-23
Delete(...) endpoint for the resource service

analogue:
### Description

See NET-2690.

`gotestsum --format=testname ./agent/grpc-external/services/resource`

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


boxofrad (changes requested):
Great work as usual! I've left one minor comment inline, but need to clarify this point:

> Acceptance criteria mentioned having to re-read the resource to get the version, but the version is already passed in via the `DeleteRequest` and seem like punting that off to the storage backend is a better place for that responsibility.

We need to read the resource for cases where `req.Version` is blank — i.e. the majority of user-issued writes on the CLI. In these cases, the user isn't _intentionally_ doing a CAS operation (they just want to delete whatever version of the resource exists) but the storage backend, by design, only supports CAS write and delete operations, partly to keep the interface as small as possible, and to allow for validation hooks in the future.

With this approach, we'll also need to handle `ErrCASFailure` by retrying the whole read-delete loop, but only when the user didn't specify a version. I'm currently implementing this for the `Write` endpoint though, so maybe hold off until we can reuse it here?

analogue:
@boxofrad I merged in `boxofrad/resource-write-endpoint` to get at the demo apis for testing and took another stab at the endpoint. I think there are some gaps in the acceptance criteria so I left a couple of TODOs in the tests. Let's chat when you get a chance.

agent/grpc-external/services/resource/delete.go
```diff
@@ -0,0 +1,48 @@
...
+
+	if req.Version == "" {
+		// Delete resource regardless of the persisted Version. Hence, strong read
+		// necessary to get latest Version
+		existing, err := s.Backend.Read(ctx, storage.StrongConsistency, req.Id)
+		if err != nil {
+			return nil, err
+		}
+		if err = s.Backend.DeleteCAS(ctx, existing.Id, existing.Version); err != nil {
+			return nil, err
```
boxofrad, line 35: Remember to check for `ErrCASFailure` here too 😉

agent/grpc-external/services/resource/delete.go
```diff
@@ -0,0 +1,48 @@
...
+		return nil, err
+	}
+
+	// TODO(spatel): reg will be used for ACL hooks
+	_ = reg
+
+	if req.Version == "" {
+		// Delete resource regardless of the persisted Version. Hence, strong read
+		// necessary to get latest Version
+		existing, err := s.Backend.Read(ctx, storage.StrongConsistency, req.Id)
```
boxofrad, line 30: FWIW, there's a chance that the resource will be updated between the `Read` and `DeleteCAS`, in which case the user would see `ErrCASFailure`.

It's pretty unlikely though, so I don't think we have to handle it with retries like we do in the `Write` endpoint. I guess it'd happen if a user was trying to delete a resource at the same time a controller was modifying it.

boxofrad (commented):
Nice job 🏆 🎉

analogue:
@boxofrad Ok, changed things up a bit based on a better understanding of the behavior here.

agent/grpc-external/services/resource/delete.go
analogue: 👍 Fixed in https://github.com/hashicorp/consul/pull/16756/commits/d038a2538accb28fdad204653df3305e38058fda

agent/grpc-external/services/resource/delete.go
analogue: Fixed in https://github.com/hashicorp/consul/pull/16756/commits/d038a2538accb28fdad204653df3305e38058fda

agent/grpc-external/services/resource/delete_test.go
```diff
@@ -0,0 +1,103 @@
...
+	_, err = client.Delete(ctx, &pbresource.DeleteRequest{Id: rsp.Resource.Id, Version: ""})
+	require.NoError(t, err)
+
+	// verify
+	_, err = server.Backend.Read(ctx, storage.StrongConsistency, rsp.Resource.Id)
+	require.ErrorIs(t, err, storage.ErrNotFound)
+}
+
+func TestDelete_ByVersion_NonExistantIsNoOp(t *testing.T) {
+	// TODO: Not sure what is expected bahavior here
```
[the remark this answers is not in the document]
analogue: 👍 Fixed in https://github.com/hashicorp/consul/pull/16756/commits/d038a2538accb28fdad204653df3305e38058fda

agent/grpc-external/services/resource/delete.go
```diff
@@ -24,25 +27,26 @@ func (s *Server) Delete(ctx context.Context, req *pbresource.DeleteRequest) (*pb
...
 		}
-		return &pbresource.DeleteResponse{}, nil
+		versionToDelete = existing.Version
 	}
 
-	// non-empty Version so let backend enforce Versions match for a CAS delete
-	if err = s.Backend.DeleteCAS(ctx, req.Id, req.Version); err != nil {
+	if err = s.Backend.DeleteCAS(ctx, req.Id, versionToDelete); err != nil {
 		if errors.Is(err, storage.ErrCASFailure) {
 			return nil, status.Error(codes.FailedPrecondition, err.Error())
```
boxofrad, line 47: I've realized while reading the gRPC status [docs](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) that `Aborted` is a more appropriate code for CAS failures. 
```suggestion
			return nil, status.Error(codes.Aborted, err.Error())
```

agent/grpc-external/services/resource/delete.go
```diff
@@ -24,25 +27,26 @@ func (s *Server) Delete(ctx context.Context, req *pbresource.DeleteRequest) (*pb
...
 		if err != nil {
-			return nil, err
-		}
-		if err = s.Backend.DeleteCAS(ctx, existing.Id, existing.Version); err != nil {
-			return nil, err
+			if errors.Is(err, storage.ErrNotFound) {
+				// deletes are idempotent so no-op if resource not found
+				return &pbresource.DeleteResponse{}, nil
+			}
+			return nil, fmt.Errorf("failed read: %v", err)
```
boxofrad, line 40: Sorry to be a pain, but would you mind making the status code explicit here? 🙇🏻‍♂️  (see: https://github.com/hashicorp/consul/pull/16842#discussion_r1157085813)

boxofrad (approved):
Couple of very small comments, but this is basically good to go (after a rebase). Awesome work! 👏🏻

agent/grpc-external/services/resource/delete.go
analogue: 👍 Fixed in https://github.com/hashicorp/consul/pull/16756/commits/24714ca30e26d6ac720f02bdfe4232159f549c2c

agent/grpc-external/services/resource/delete.go
analogue: 👍 Fixed in https://github.com/hashicorp/consul/pull/16756/commits/24714ca30e26d6ac720f02bdfe4232159f549c2c

PR #16760, opened 2023-03-24
state-store: allocate / deallocate VIP on config-entry writes

kyhavlov:
### Description
Allocate virtual ips for resolver/router/splitter config entries if there isn't already one for the service. This is part of upcoming enhancements for transparent proxy. A virtual IP also now won't be deallocated unless there are no instances left and no resolver/router/splitter config entry for a service.

### PR Checklist

* [X] updated test coverage
* [X] not a security concern

### DIFF
--- a/agent/consul/state/config_entry.go
+++ b/agent/consul/state/config_entry.go
@@ -455,6 +455,15 @@ func deleteConfigEntryTxn(tx WriteTxn, idx uint64, kind, name string, entMeta *a
 		return fmt.Errorf("failed updating index: %s", err)
 	}
 
+	// If this is a resolver/router/splitter, attempt to delete the virtual IP associated
+	// with this service.
+	if kind == structs.ServiceResolver || kind == structs.ServiceRouter || kind == structs.ServiceSplitter {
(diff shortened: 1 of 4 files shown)


agent/consul/state/config_entry.go
```diff
@@ -465,14 +474,23 @@ func insertConfigEntryWithTxn(tx WriteTxn, idx uint64, conf structs.ConfigEntry)
...
+	if kind == structs.TerminatingGateway || kind == structs.IngressGateway {
 		err := updateGatewayServices(tx, idx, conf, conf.GetEnterpriseMeta())
 		if err != nil {
 			return fmt.Errorf("failed to associate services to gateway: %v", err)
 		}
 	}
 
-	switch conf.GetKind() {
+	// If this is a resolver/router/splitter, attempt to assign a virtual IP for the service.
+	if kind == structs.ServiceResolver || kind == structs.ServiceRouter || kind == structs.ServiceSplitter {
```
hashi-derek, line 486: This is a nit, and you don't have to address it necessarily. It might be cleaner to put this logic in the switch statement below with some fallthroughs.

hashi-derek (approved):

PR #16771, opened 2023-03-24
Start using pre-commit to automate things on git commit: goimports

analogue:
### Description

See [NET-3409](https://hashicorp.atlassian.net/browse/NET-3409) for motivation and acceptance criteria.

[pre-commit](https://pre-commit.com/) is a pretty popular tool in this space and has 10k+ stars on github. 

I'm specifically looking for `git commit` workflows which can possibly fail by introducing this change. The workaround to bypass pre-commit is `git commit --no-verify`.

There is a wrinkle with bootstrapping this functionality. The git hook can't be stored in source control because it is generated and dependent on the local dev environment (where is python installed? etc). So, after a fresh `git clone` it is necessary to run `pre-commit install` to install the git hook. Currently, I have piggybacked this behavior onto `make tools` target to not opt-everyone into this as the experience probably needs refining and some tweaks. One of the things we can do once confidence in this workflow has increased is to just have a raw make target on the `./.git/hooks/pre-commit` file and make it a dependency of the default make target.

### Highlights
- Removes the need for developers to configure `goimports` specifically for this repo 
- Removes the need for developers to have to remember to run `goimports` before they commit code locally
- Provides an avenue to add even more automation around coding standards, conventions, and policies. e.g. linting, protobufs, yaml, hcl, ui stuff, etc.
- Catches automatable fixes locally rather than waiting for CI to catch them post-commit.
 
- verified on an M1 mac with homebrew
- verified on an M1 mac with python + pip
- verified on a linux VM with python + pip

### Developer experience
- Dev gets a feature branch in shape for commit and stages files
- Dev runs `git commit -m ...`
- All pre-commit hooks (go-imports in this case) are run against staged files
- Changes are written to files by goimports and dev is notified of the failure + changes
- Dev inspects changes which were made
- Dev likes the changes, re-stages files and runs `git commit` again
- All pre-commit hooks are run and this time they pass with a visual green
- The commit operation is successful

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

[NET-3409]: https://hashicorp.atlassian.net/browse/NET-3409?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -366,6 +366,10 @@ proto-tools:
 codegen-tools:
 	@$(SHELL) $(CURDIR)/build-support/scripts/devtools.sh -codegen
 
+.PHONY: pre-commit-tools
+pre-commit-tools:
+	@$(SHELL) $(CURDIR)/build-support/scripts/devtools.sh -pre-commit
+
 .PHONY: deep-copy
 deep-copy: codegen-tools
 	@$(SHELL) $(CURDIR)/agent/structs/deep-copy.sh
@@ -376,8 +380,8 @@ deep-copy: codegen-tools
 precommit: pre-commit
 
 .PHONY: pre-commit
-pre-commit:
-	@$(SHELL) $(CURDIR)/build-support/scripts/devtools.sh -pre-commit
+pre-commit: pre-commit-tools
+	pre-commit run
 
 version:
 	@echo -n "Version:                    "
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -188,7 +188,7 @@ remote-docker: check-remote-dev-image-env
 	@echo "Building and Pushing Consul Development container - $(REMOTE_DEV_IMAGE)"
 	@if ! docker buildx inspect consul-builder; then \
 		docker buildx create --name consul-builder --driver docker-container --bootstrap; \
-	fi; 
+	fi;
 	@docker buildx use consul-builder && docker buildx build -t '$(REMOTE_DEV_IMAGE)' \
        --platform linux/amd64,linux/arm64 \
 	   --build-arg CONSUL_IMAGE_VERSION=$(CONSUL_IMAGE_VERSION) \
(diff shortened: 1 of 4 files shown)


GNUmakefile
```diff
@@ -188,7 +188,7 @@ remote-docker: check-remote-dev-image-env
 	@echo "Building and Pushing Consul Development container - $(REMOTE_DEV_IMAGE)"
 	@if ! docker buildx inspect consul-builder; then \
 		docker buildx create --name consul-builder --driver docker-container --bootstrap; \
-	fi;
```
analogue, line 191: oh look! the trailing-whitespace hook is already working 😄

GNUmakefile
```diff
@@ -188,7 +188,7 @@ remote-docker: check-remote-dev-image-env
 	@echo "Building and Pushing Consul Development container - $(REMOTE_DEV_IMAGE)"
 	@if ! docker buildx inspect consul-builder; then \
 		docker buildx create --name consul-builder --driver docker-container --bootstrap; \
-	fi; 
+	fi;
```
analogue, line 191: oh look! the trailing whitespace hook is already working :smile:

jmurret:
I've used pre-commit a lot and like it myself but this is really an entire group change and my proposal to use this on my team's repos at my last job turned into lots of debate and ultimately punting.  

Since we have so many contributors, I think would be worthwhile to get some other team leads and so staff/principal input because I would not be surprised if this has been suggested before or tried and removed.

Having said all that, I would be in favor of this.

analogue:
> 

Adding @banks, @rboyer, and @erichaberkorn for wider discussion.

GNUmakefile
```diff
@@ -366,11 +366,23 @@ proto-tools:
...
+
 .PHONY: deep-copy
 deep-copy: codegen-tools
 	@$(SHELL) $(CURDIR)/agent/structs/deep-copy.sh
 	@$(SHELL) $(CURDIR)/agent/proxycfg/deep-copy.sh
 
+# Devs shoudln't have to think about whether there is a hypen or not
+.PHONY: precommit
+precommit: pre-commit
+
```
dhiaayachi, line 381: I like the fact that this is voluntary, but I guess once merged it need to be socialized so people install it locally.

dhiaayachi (approved):
Great Idea @analogue !! I fully support introducing this.

jmurret (approved):

PR #16773, opened 2023-03-24
Modify build-distros to use medium runners

jmurret:
### Description
The current build-distro times in GHA are below compared to CircleCI:
Screenshot 2023-03-24 at 2 31 38 PM

This change brings them to:
Screenshot 2023-03-24 at 2 32 24 PM

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


loshz:
How does this impact the other work we're doing to use custom runners?

.github/workflows/build-distros.yml
```diff
@@ -2,7 +2,7 @@
 # It is aimed at checking new commits don't introduce any breaking build changes.
 name: build-distros
 
-on: [pull_request]
+on: [push]
```
loshz, line 5: Was this just for testing?

jmurret: this was based on the discussion about keeping this the same as CircleCI until we migrate and get everyone on board that we are changing how we trigger workflows.

loshz: Ah, I thought we were just going to do it straight out of the gate.

jmurret: I'll post something asking the question so its a fast follow, but I think it would be good to get full awareness before we change this.

loshz (approved):

PR #16776, opened 2023-03-24
Allow dialer to re-establish terminated peering

freddygv:
### Description

Currently, if an acceptor peer deletes a peering the dialer's peering will eventually get to a "terminated" state. If the two clusters need to be re-peered the acceptor will re-generate the token but the dialer will encounter this error on the call to establish:

>failed to get addresses to dial peer: failed to refresh peer server addresses, will continue to use initial addresses: there is no active peering for "<<<ID>>>"

This is because in `exchangeSecret().GetDialAddresses()` we will get an error if fetching addresses for an inactive peering. The peering shows up as inactive at this point because of the existing terminated state.

Rather than checking whether a peering is active we can instead check whether it was deleted. This way users do not need to delete terminated peerings in the dialing cluster before re-establishing them.

* The `TestLeader_PeeringSync_Lifecycle_ServerDeletion` test updated below shows how to reproduce the issue. It fails without this update.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/.changelog/16776.txt
+++ b/.changelog/16776.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+peering: allow re-establishing terminated peering from new token without deleting existing peering first.
+```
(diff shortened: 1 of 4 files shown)


wilkermichael (approved):
Looks good to me, I think this makes more sense

PR #16777, opened 2023-03-25
Github Actions Migration - move verify-ci workflows to GHA

jmurret:
### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/.github/workflows/verify-ci.yml
+++ b/.github/workflows/verify-ci.yml
@@ -8,8 +8,7 @@ name: verify-ci
 permissions:
   contents: read
 
-on:
-  pull_requests:
+on: [pull_request]
 
 jobs:
   noop:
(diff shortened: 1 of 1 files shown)


loshz (approved):
We should probably add a TODO or note somewhere to investigate if we still need this going forward so we can potentially remove this.

jmurret:
@loshz yea, I am going to close for now and when we change the required GH Checks for PRs, let's see if we can't just remove this.  if we need it, we can re-open this.

jmurret:
@loshz actually re-opening.  This is used as a `fail fast` check so that we detect when CI is down as quickly as possible.

loshz:
> @loshz actually re-opening. This is used as a `fail fast` check so that we detect when CI is down as quickly as possible.

Sounds good. Do we need to run on push as well as pr?

PR #16786, opened 2023-03-27
Resource `Write` endpoint

boxofrad:
### Description

This PR implements the `Write` endpoint for the resource service, including handling for both CAS and non-CAS writes (they're both actually CAS under the hood 🤫).

It adds the protobuf message to the type registration so that we can validate the `data` field on-write.

It also introduces a set of fake resource types (`Artist` and `Album`) that can be used in tests and `-dev` mode, so we can work on generic resource storage stuff without referring to specific Consul features.

Finally, it hooks up the storage backend and registry in `agent/consul/server.go` so we can poke the service using grpcurl.

### Unit Tests

```shell
$ go test -v ./agent/grpc-external/services/resource -run="TestWrite_/"
```

#### Terminal 1: Write Command
```shell
$ grpcurl -d @ \
  -plaintext \
  -import-path proto-public \
  -import-path proto/private \
  -proto proto/private/pbdemo/v1/demo.proto \
  -proto proto-public/pbresource/resource.proto \
  127.0.0.1:8502 \
  hashicorp.consul.resource.ResourceService.Write \
<<EOF
  {
    "resource": {
      "id": {
        "type": {
          "group": "demo",
          "group_version": "v1",
          "kind": "artist"
        },
        "tenancy": {
          "partition": "default",
          "peer_name": "local",
          "namespace": "default"
        },
        "name": "fightstar"
      },
      "data": {
        "@type": "types.googleapis.com/hashicorp.consul.internal.demo.v1.Artist",
        "name": "Fightstar",
        "genre": "GENRE_METAL"
      }
    }
  }
EOF
```
```json
{
  "resource": {
    "id": {
      "uid": "01GWMK7WEN6QP4PESWSPNBJY5J",
      "name": "fightstar",
      "type": {
        "group": "demo",
        "groupVersion": "v1",
        "kind": "artist"
      },
      "tenancy": {
        "partition": "default",
        "namespace": "default",
        "peerName": "local"
      }
    },
    "version": "2",
    "generation": "01GWMK7WEN6QP4PESWSRHCC7AD",
    "data": {"@type":"types.googleapis.com/hashicorp.consul.internal.demo.v1.Artist","genre":"GENRE_METAL","name":"Fightstar"}
  }
}
```

#### Terminal 2: Watch Command

```shell
$ grpcurl -d @ \
  -plaintext \
  -import-path proto-public \
  -import-path proto/private \
  -proto proto/private/pbdemo/v1/demo.proto \
  -proto proto-public/pbresource/resource.proto \
  127.0.0.1:8502 \
  hashicorp.consul.resource.ResourceService.WatchList \
<<EOF
  {
      "type": {
        "group": "demo",
        "group_version": "v1",
        "kind": "artist"
      },
      "tenancy": {
        "partition": "default",
        "peer_name": "local",
        "namespace": "default"
      }
  }
EOF
```
```json
{
  "operation": "OPERATION_UPSERT",
  "resource": {
    "id": {
      "uid": "01GWMK7WEN6QP4PESWSPNBJY5J",
      "name": "fightstar",
      "type": {
        "group": "demo",
        "groupVersion": "v1",
        "kind": "artist"
      },
      "tenancy": {
        "partition": "default",
        "namespace": "default",
        "peerName": "local"
      }
    },
    "version": "2",
    "generation": "01GWMK7WEN6QP4PESWSRHCC7AD",
    "data": {"@type":"types.googleapis.com/hashicorp.consul.internal.demo.v1.Artist","genre":"GENRE_METAL","name":"Fightstar"}
  }
}

```


internal/storage/inmem/store.go
```diff
@@ -114,15 +114,15 @@ func (s *Store) WriteCAS(res *pbresource.Resource, vsn string) error {
...
-
 		// Uid is immutable.
 		if existingRes.Id.Uid != res.Id.Uid {
 			return storage.ErrWrongUid
 		}
+
+		// Ensure CAS semantics.
+		if existingRes.Version != vsn {
+			return storage.ErrCASFailure
+		}
```
boxofrad, line 125: Moving the `Uid` check first makes the error message clearer when you're trying to do a non-CAS write against an old incarnation of the resource. It's because if the resource exists but with a different `Uid`, the read will return `ErrNotFound`, so we'll leave the `Version` blank which will return `ErrCASFailure`.

agent/grpc-external/services/resource/write.go
```diff
@@ -0,0 +1,144 @@
...
+				// it here too to make sure we don't carry over statuses from the wrong
+				// version.
+				return storage.ErrCASFailure
+			}
+			// TODO: Carry over the statuses here.
+
+		case errors.Is(err, storage.ErrNotFound):
+			input.Id.Uid = ulid.Make().String()
+			// TODO: Prevent setting statuses in this endpoint.
+
```
analogue, line 66: Maybe add a comment here that this is a "create" hence generating and assigning a Uid?

agent/grpc-external/services/resource/write.go
```diff
@@ -0,0 +1,144 @@
...
+		// CAS failure, so it's not worth the latency penalty of StrongConsistency.
+		existing, err := s.Backend.Read(ctx, storage.EventualConsistency, input.Id)
+		switch {
+		case err == nil:
+			if input.Id.Uid == "" {
+				input.Id.Uid = existing.Id.Uid
+			}
+
+			if input.Version == "" {
+				input.Version = existing.Version
```
analogue, line 54: IIUC, when the resource is found and the `input.Version` is empty, the caller's intent is to do an update regardless of the version of the object (a non-CAS write). Adding a comment here to that effect would be useful.

analogue (commented):
Great progress here! 😄   

I had to lean heavily on the tests to decode what was going on in the "Write" endpoint but I think that can be remedied easily with some additional comments.

agent/grpc-external/services/resource/write.go
```diff
@@ -0,0 +1,144 @@
...
+		input := clone(req.Resource)
+
+		// Read with EventualConsistency because we'll already automatically retry on
+		// CAS failure, so it's not worth the latency penalty of StrongConsistency.
+		existing, err := s.Backend.Read(ctx, storage.EventualConsistency, input.Id)
+		switch {
+		case err == nil:
+			if input.Id.Uid == "" {
+				input.Id.Uid = existing.Id.Uid
+			}
```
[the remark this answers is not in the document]
boxofrad: Yep! Probably also worth pointing out that users generally won't provide the `Uid` (or even know it's a thing) but controllers **always** will.

agent/grpc-external/services/resource/write.go
boxofrad: Sure thing! 👍🏻

agent/grpc-external/services/resource/write.go
boxofrad: Exactly, yeah. Good idea!

agent/grpc-external/services/resource/write.go
```diff
@@ -0,0 +1,144 @@
...
+		)
+	}
+
+	var result *pbresource.Resource
+	err = s.retryCAS(ctx, req.Resource.Version, func() error {
+		input := clone(req.Resource)
+
+		// Read with EventualConsistency because we'll already automatically retry on
+		// CAS failure, so it's not worth the latency penalty of StrongConsistency.
+		existing, err := s.Backend.Read(ctx, storage.EventualConsistency, input.Id)
```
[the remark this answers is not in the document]
boxofrad: Thanks for the call out! To clarify, the whole retry-loop thing is less of an optimization, and more to hide the fact we're doing a CAS operation from the user.

For context, all writes have to be CAS operations if we want to (safely) do things like:
- Ensure the `Uid` doesn't change between writes.
- Copy the resource `Status` between writes.
- Pass the current version of a resource to pre-write validation/mutation hooks.

That is, unless we only support more feature-rich storage backends that allow you to do "patch"-style updates.

Generally, people don't need CAS semantics in the same way that scripts and controllers do, they just want to overwrite whatever's currently there. So it'd be confusing for them to see a message about a CAS operation they didn't know was happening having failed because of a concurrent write 😄 hence the retry-loop.

My thinking regarding reading with `EventualConsistency` was that because of our fairly conservative network latency recommendations, Raft followers are usually less than a few hundred milliseconds behind the leader in a healthy cluster. In the majority of cases, the first read will return the latest version, so it seems like a reasonable trade-off to avoid having to do 2 round trips to the followers (one to verify leadership for the consistent read, and one to append the write log).

Does that sound ok? If so I'll update the comments to be clearer.

agent/grpc-external/services/resource/write.go
```diff
@@ -0,0 +1,168 @@
...
+		//
+		//	- In the common case, individual resources are written infrequently, and
+		//	  when using the Raft backend followers are generally within a few hundred
+		//	  milliseconds of the leader, so the first read will probably return the
+		//	  current version.
+		//
+		//	- StrongConsistency is expensive. In the Raft backend, it involves a round
+		//	  of heartbeats to verify cluster leadership (on top of log replication).
+		//
+		//	- CAS failures will be retried anyway. So the read-modify-write cycle
```
analogue, line 64: minor doc: Just so the reader doesn't get confused with the "retried" mentioned here and the s.retryCAS(..) function, a little tweak might be useful:

- CAS failures will be retried **by the caller** anyway.

agent/grpc-external/services/resource/write.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/oklog/ulid/v2"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/hashicorp/consul/internal/storage"
+	"github.com/hashicorp/consul/lib/retry"
+	"github.com/hashicorp/consul/proto-public/pbresource"
+)
+
+func (s *Server) Write(ctx context.Context, req *pbresource.WriteRequest) (*pbresource.WriteResponse, error) {
```
analogue, line 18: Can you add a comprehensive doc string for the function or is that better done in the proto file? I imagine quite a few controller authors will be calling/relying on this function and will want to know expectations for inputs/outputs/errors.

analogue, line 18: Can you add a short doc string for the function or is that better done in the proto file? I imagine quite a few controller authors will be calling/relying on this function and will want to know expectations for inputs/outputs/errors.

analogue (approved):
Looks good! Thanks for adding comprehensive comments - they really helped :)  I just had a couple more non-blocking doc nits but this looks good to go. 🚀

agent/grpc-external/services/resource/write.go
boxofrad: Ah yes, I've been meaning to loop back and write some proper API documentation in the proto file. I think that's the best place because it'll be used as the godoc comments for the interface.

agent/grpc-external/services/resource/write.go
boxofrad: Thanks for the call out! Retried here does actually refer to `retryCAS` 😅 so this could definitely use a little clarifying.

PR #16788, opened 2023-03-27
Merge Bootstrap Config from CCM

jjacobson93:
### Description

This PR adds an additional field to the internal runtime config which allows for an additional initial management token to be sent from HCP, such that it does override the user-provided initial management token. We also add implementation to merge HTTP response headers into the runtime config from the HCP bootstrap.

### Links

[RFC for CCM Linking Existing Clusters](https://docs.google.com/document/d/1q-cGem9JF4LIuSoD2n_KwRobL0Uf2QWbBqnFj5YcWPM/edit#heading=h.cxaty5xfhqiw)

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


freddygv:
Thanks for getting this started! I opened up a new PR to make things easier to review: #16916

PR #16789, opened 2023-03-27
Update normalization of route refs

jm96441n:
### Description

When normalizing refs to routes, if the ref does not specify a namespace we should default to the routes namespace. Previously this would always default back to the default namespace.

In consul enterprise create an api gateway and a service in a namespace, then create a route in the same namespace from the gateway to the service and in the stanzas for the gw and the service do not specify the namespace. Before this change the route will fail to bind because it assumes the gw and service are in the default ns, after this change it will bind correctly.

### Links

N/A

### PR Checklist

* [X] updated test coverage
* [X] not a security concern


agent/structs/config_entry_gateways.go
```diff
@@ -173,7 +173,7 @@ func (e *IngressGatewayConfigEntry) Normalize() error {
 	}
 
 	e.Kind = IngressGateway
-	e.EnterpriseMeta.Normalize()
+	e.Normalize()
```
andrewstucki, line 176: So, pretty sure this needs to switch back, otherwise you're in an infinite call loop:

```suggestion
	e.EnterpriseMeta.Normalize()
```

andrewstucki (commented):
Aside from the one issue, generally looks fine to me.

agent/structs/config_entry_gateways.go
jm96441n: oh yup, good catch

andrewstucki (approved):
I'm ok with this overall. Stylistically I'd probably prefer not to elide the calls to `EnterpriseMeta.Normalize` on the embedded `EnterpriseMeta` on the `ResourceReference` objects, just because it's a bit more explicit what we're actually trying to normalize, but not going to nitpick too much on that.

jm96441n:
I can pretty easily undo the removal of the `EnterpriseMeta` in the calls if that's preferred

PR #16797, opened 2023-03-28
[CC-4374]Scaffold HCP Telemetry

Achooo:
#### HCP Telemetry
TLDR: new feature to ship Server metrics directly to HCP.
- That branch will be merged at the end into `main` and backported

PRs to look at before this one:
- #16798 
- #16800 

⚠️ Merge those first into feature branch, as this is rebased on top of it.

#### Changes
This PR is a first to scaffold HCP telemetry, namely to:
- Create a **reporter** that obtains metrics from go metrics at a configurable interval
  - Reporter starts if HCP creds are configured.
  - Server is registered with CCM, and metrics endpoint configuration is obtained.
- Create a **metrics exporter** that exports the metrics:
  - In OTLP format
  - If they are in the allowed list of filters (regex based)
  - Exports to the configured endpoint to the telemetry gateway
  - Uses a client capable of HCP auth (parts of it will be done later)

#### Next steps
- Batching strategy improvements
- Complete HCP Client auth in OTLP Exporter

I tested this locally end to end with:
- Configuring a `server.json` that contains HCP creds configuration to run the reporter
- Modyfing the OTLP client to allow HTTP requests
```golang
unc (c *hcpClient) InitMetricsClient(ctx context.Context, endpoint string) error {
	// hcpConfig, err := c.cfg.HCPConfig()
	// if err != nil {
	// 	return err
	// }

	// TODO:In follow up PR (CC-4635) : Need to use oauth2 transport in the otlpmetrichttp client to add token.
	// We likely need to fork and update the otlpmetrichttp client, as there is no interface to do this currently.
	exp, err := otlpmetrichttp.New(ctx, otlpmetrichttp.WithEndpoint(endpoint), otlpmetrichttp.WithInsecure())
	if err != nil {
		return err
	}

	c.exporter = exp
	return nil
}
``` 
- Setting a mock OTLP receiver at `localhost:9090` since the telemetry config returns that hardcoded value for now
- Run `bin/consul agent -dev -config-file=config.json`

In my OTLP receiver, I get metrics 🥳 :
```
~/Github/HashiCorp/mock-receiver 6m 24s
❯ go run main.go
{"resourceMetrics":[{"resource":{"attributes":[{"key":"service.instance.id","value":{"stringValue":"04b1534a-fa03-dc66-e127-f829314ef448"}},{"key":"service.name","value":{"stringValue":"consul-server"}},{"key":"service.version","value":{"stringValue":"1.16.0-dev"}}]},"scopeMetrics":[{"scope":{"name":"github.com/hashicorp/consul/agent/hcp/client/telemetry","version":"v1"},"metrics":[{"name":"consul.raft.leader.oldestLogAge","gauge":{"dataPoints":[{"startTimeUnixNano":"11651379494838206464","timeUnixNano":"1680028130000000000","asDouble":0}]}},{"name":"consul.raft.last_index","gauge":{"dataPoints":[{"startTimeUnixNano":"11651379494838206464","timeUnixNano":"1680028130000000000","asDouble":17}]}},{"name":"consul.raft.applied_index","gauge":{"dataPoints":[{"startTimeUnixNano":"11651379494838206464","timeUnixNano":"1680028130000000000","asDouble":17}]}},{"name":"consul.raft.thread.main.saturation","histogram":{"dataPoints":[{"startTimeUnixNano":"11651379494838206464","timeUnixNano":"1680028130000000000","count":"10","sum":0,"min":0,"max":0}],"aggregationTemporality":2}}]}]}]}
{"resourceMetrics":[{"resource":{"attributes":[{"key":"service.instance.id","value":{"stringValue":"04b1534a-fa03-dc66-e127-f829314ef448"}},{"key":"service.name","value":{"stringValue":"consul-server"}},{"key":"service.version","value":{"stringValue":"1.16.0-dev"}}]},"scopeMetrics":[{"scope":{"name":"github.com/hashicorp/consul/agent/hcp/client/telemetry","version":"v1"},"metrics":[{"name":"consul.raft.last_index","gauge":{"dataPoints":[{"startTimeUnixNano":"11651379494838206464","timeUnixNano":"1680028140000000000","asDouble":17}]}},{"name":"consul.raft.leader.oldestLogAge","gauge":{"dataPoints":[{"startTimeUnixNano":"11651379494838206464","timeUnixNano":"1680028140000000000","asDouble":0}]}},{"name":"consul.raft.applied_index","gauge":{"dataPoints":[{"startTimeUnixNano":"11651379494838206464","timeUnixNano":"1680028140000000000","asDouble":17}]}},{"name":"consul.raft.thread.main.saturation","histogram":{"dataPoints":[{"startTimeUnixNano":"11651379494838206464","timeUnixNano":"1680028140000000000","count":"10","sum":0,"min":0,"max":0}],"aggregationTemporality":2}}]}]}]}
```
Shortened for brevity ^
Screenshot 2023-03-28 at 3 51 03 PM

### PR Checklist

* [x] updated test coverage
* [ ] not a security concern


agent/hcp/manager.go
```diff
@@ -71,15 +77,80 @@ func NewManager(cfg ManagerConfig) *Manager {
...
+	if telemetryCfg == nil {
+		// Not an error, as the server is not registered with CCM, but early return.
+		return nil
+	}
+
+	// Step 2: Init telemetry.MetricsExporter which sends metrics to HCP Metrics Gateway in OTLP format.
+	// It uses a an OTLP exporter client wrapped by the HCP client.
+	// This enables us to perform HCP auth and easily mock the client interface for tests.
+	// It must first be initialized within the HCP Client with the configured endpoint.
+	if err := m.cfg.Client.InitMetricsClient(ctx, telemetryCfg.Endpoint); err != nil {
```
Achooo, line 98: @nickethier Where the Client gets initialized and injected into the exporter. We could even do this within the Exporter init, by passing the HCP client. I liked that it was more obvious what is happening this way.

agent/hcp/manager.go
```diff
@@ -71,15 +77,80 @@ func NewManager(cfg ManagerConfig) *Manager {
...
+// runReporter initializes, and if successful, then runs the metrics reporter.
+func (m *Manager) runReporter(ctx context.Context) error {
+	// Step 1: Obtain CCM telemetry configuration
+	// Only enable HCP metrics reporting if server is registered with management plane.
+	telemetryCfg, err := m.cfg.Client.FetchTelemetryConfig(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to obtain CCM telemetry config: %v", err)
+	}
+
+	if telemetryCfg == nil {
```
Achooo, line 89: Thought I should handle the case where we don't have an error per say, but the server isn't registered with CCM 🤔 This might not be nil, more empty (no endpoint) but for now since the CCM protos are not yet available I did it this way. Might change later.

agent/hcp/manager.go
```diff
@@ -71,15 +77,80 @@ func NewManager(cfg ManagerConfig) *Manager {
 	}
 }
 
+// runReporter initializes, and if successful, then runs the metrics reporter.
+func (m *Manager) runReporter(ctx context.Context) error {
```
Achooo, line 81: Returns an error so we can unit test easily!

agent/hcp/manager.go
```diff
@@ -71,15 +77,80 @@ func NewManager(cfg ManagerConfig) *Manager {
...
+
+	m.reporter, err = telemetry.NewReporter(cfg)
+	if err != nil {
+		return fmt.Errorf("failed to create exporter: %v", err)
+	}
+
+	// If setup is successful, run the reporter, which is a blocking operation.
+	m.reporter.Run(ctx)
+
+	return nil
```
Achooo, line 134: Could also make the `m.reporter.Run(ctx)` return an error to make this `return m.reporter.Run(ctx)`, but the reporter can be run outside of this package, and so it felt forced to do that.

agent/hcp/manager.go
```diff
@@ -71,15 +77,80 @@ func NewManager(cfg ManagerConfig) *Manager {
...
+}
+
 // Run executes the Manager it's designed to be run in its own goroutine for
 // the life of a server agent. It should be run even if HCP is not configured
 // yet for servers since a config update might configure it later and
 // UpdateConfig called. It will effectively do nothing if there are no HCP
 // credentials set other than wait for some to be added.
 func (m *Manager) Run(ctx context.Context) {
-	var err error
 	m.logger.Debug("HCP manager starting")
```
Achooo, line 144: One thing that worries me is that we are not handling goroutine exits, in the case that the if the shared context is cancelled both the manager `Run` and this inner goroutine will return with a parent that exited. 

Some thoughts:
- Maybe this is this fine as is? 
- If not:
  - Should I use a separate context for the reporter goroutine, and cancel it when the manager `Run` function cancels?
  - Should I convert the contents of the manager `Run` function into a goroutine `run` and wait on both `runReporter` and `run` ?

agent/hcp/telemetry/exporter.go
Achooo, line 1: The exporter converts batched metrics from the Reporter into OTLP format and exports these.

agent/hcp/telemetry/exporter.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+
+		for _, v := range interval.Counters {
+			if !m.filter.Match(v.Name) {
+				continue
+			}
+			otlpMetrics = append(otlpMetrics, metricdata.Metrics{
+				Name: v.Name,
+				Data: metricdata.Sum[float64]{
+					Temporality: metric.DefaultTemporalitySelector(metric.InstrumentKindCounter),
```
Achooo, line 121: I _think_ these are the right Temporality values...

agent/hcp/manager.go
```diff
@@ -71,15 +77,80 @@ func NewManager(cfg ManagerConfig) *Manager {
...
+	// It must first be initialized within the HCP Client with the configured endpoint.
+	if err := m.cfg.Client.InitMetricsClient(ctx, telemetryCfg.Endpoint); err != nil {
+		return fmt.Errorf("failed to init metrics HCP client: %v", err)
+	}
+
+	expCfg := &telemetry.MetricsExporterConfig{
+		Labels: map[string]string{
+			"service.name":        "consul-server",
+			"service.version":     version.GetHumanVersion(),
+			"service.instance.id": m.cfg.NodeID,
```
Achooo, line 106: Labels are attributes that the metrics will be tagged with, for example the service instance id (`service.instance.id`)

agent/hcp/telemetry/filter.go
Achooo, line 1: #### Functions of the reporter

Data structure that holds the metrics filters (what we want to send)/ allowlist

agent/hcp/client/client.go
```diff
@@ -120,6 +143,31 @@ func (c *hcpClient) PushServerStatus(ctx context.Context, s *ServerStatus) error
...
 
+func (c *hcpClient) InitMetricsClient(ctx context.Context, endpoint string) error {
+	hcpConfig, err := c.cfg.HCPConfig()
+	if err != nil {
+		return err
+	}
+
+	// TODO:In follow up PR (CC-4635) : Need to use oauth2 transport in the otlpmetrichttp client to add token.
+	// We likely need to fork and update the otlpmetrichttp client, as there is no interface to do this currently.
+	exp, err := otlpmetrichttp.New(ctx, otlpmetrichttp.WithEndpoint(endpoint), otlpmetrichttp.WithTLSClientConfig(hcpConfig.APITLSConfig()))
```
Achooo, line 154: Note for reviewers: I initially wanted to have this directly in the Exporter (telemetry/exporter.go), however, to do HCP auth, we need the `hcpConfig`.

After looking at different options, this seemed like the best as the HCP client.... is a client.
The exporter is depending on a client to do the actual export. 

With this design, it made mocking very easy, and so testing extremely easy as well.

Achooo, line 154: Note for reviewers: I initially wanted to have this `otlpmetrichttp` client directly in the Exporter (telemetry/exporter.go), however, to do HCP auth, we need the `hcpConfig` which is injected during bootstrap into the HCP client.

With this design, it made mocking very easy, and so testing extremely easy as well. 

Other options I considered:
1. Return the `hcpConfig ` with a new `HCPConfig` method on the HCP Client and keep the OTLP specific client in Exporter. Although this works fine, I didn't get the mocking benefits, and it became hard to test the Exporter. A
2. Remove Exporter and use the Reporter + HCP Client only. This made the Reporter really complex. From a code maintainability standpoint, this felt much cleaner and the concerns are well separated.

agent/hcp/telemetry/exporter_test.go
```diff
@@ -0,0 +1,245 @@
...
+					Name:  "consul.raft.peers",
+					Value: 2.0,
+					Labels: []metrics.Label{
+						{
+							Name:  "peer_type",
+							Value: "test",
+						},
+					},
+				},
+				"filtered_gauges": {
```
Achooo, line 118: TODO: should probably refactor out this filter case :p tried to do it all in one test case...

agent/hcp/telemetry/exporter.go
```diff
@@ -0,0 +1,169 @@
...
+	for key, val := range cfg.Labels {
+		attr = append(attr, attribute.KeyValue{
+			Key:   attribute.Key(key),
+			Value: attribute.StringValue(val),
+		})
+	}
+
+	f := &FilterList{map[string]*regexp.Regexp{}}
+	if err := f.Set(cfg.Filters); err != nil {
+		return nil, fmt.Errorf("invalid regex: %v", err.Error())
```
Achooo, line 54: TODO: When we have bad filters, I think we should just log the error instead of failing the metrics collection. Will update this but if anyone has any thoughts please let me know!

agent/hcp/telemetry/exporter.go
```diff
@@ -0,0 +1,169 @@
...
+	client   hcpclient.Client
+	resource *resource.Resource
+	logger   hclog.Logger
+	filter   *FilterList
+}
+
+// NewMetricsExporter returns a Metrics exporter with labels to tag metrics
+// and filters to only export allowed metrics.
+func NewMetricsExporter(cfg *MetricsExporterConfig) (*MetricsExporter, error) {
+	if cfg.Client == nil || cfg.Logger == nil {
```
Achooo, line 40: Safety guard, a client / logger are needed. If this package is used outside of our use case, this ensures the config object is configured well.

agent/hcp/telemetry/reporter.go
```diff
@@ -0,0 +1,150 @@
...
+	batchedMetrics       map[time.Time]*metrics.IntervalMetrics
+	flushCh              chan struct{}
+	lastIntervalExported time.Time
+
+	// Only used for test purposes to know when flush has occurred.
+	testFlushCh chan struct{}
+}
+
+func NewReporter(cfg *ReporterConfig) (*Reporter, error) {
+	if cfg.Exporter == nil || cfg.Gatherer == nil || cfg.Logger == nil {
```
Achooo, line 50: Safety guard, a client / logger are needed. If this package is used outside of our use case, this ensures the config object is configured well.

agent/hcp/manager.go
```diff
@@ -71,15 +77,80 @@ func NewManager(cfg ManagerConfig) *Manager {
...
+	// Step 2: Init telemetry.MetricsExporter which sends metrics to HCP Metrics Gateway in OTLP format.
+	// It uses a an OTLP exporter client wrapped by the HCP client.
+	// This enables us to perform HCP auth and easily mock the client interface for tests.
+	// It must first be initialized within the HCP Client with the configured endpoint.
+	if err := m.cfg.Client.InitMetricsClient(ctx, telemetryCfg.Endpoint); err != nil {
+		return fmt.Errorf("failed to init metrics HCP client: %v", err)
+	}
+
+	expCfg := &telemetry.MetricsExporterConfig{
+		Labels: map[string]string{
```
jjtimmons, line 103: Is there a future PR where we get Labels out of `telemetryCfg`? I ask because I think we'll want to tag all metrics with `hcp_internal_id` (possibly organization/project ID as well) so the frontend dashboards can delineate between multiple clusters' metrics in an organization's workspace

agent/hcp/telemetry/reporter_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+	expectedOTLPMetrics := metricdata.ResourceMetrics{
+		Resource: resource.NewWithAttributes(""),
+		ScopeMetrics: []metricdata.ScopeMetrics{
+			{
+				Scope: instrumentation.Scope{
+					Name:    "github.com/hashicorp/consul/agent/hcp/client/telemetry",
+					Version: "v1",
+				},
+				Metrics: r.cfg.Exporter.ConvertToOTLP([]*metrics.IntervalMetrics{interval}),
```
Achooo, line 106: Ash TODO: move the metricdtat.ResourceMetrics into the `ConvertToOTLP` so that these lines of code reduce even further

agent/hcp/manager.go
jjtimmons: Related to metric tagging, Cortex/AMP has a feature called "high-availability" that deduplicates metrics given a couple of metric tags: https://docs.aws.amazon.com/prometheus/latest/userguide/AMP-ingest-dedupe.html#Send-high-availability-data. The benefit of this is that if Consul servers and our OTEL collector are both pushing the same metrics, AMP will deduplicate them (saving us $ and avoiding inaccurate metrics).

I think we may want to use something like the below to take advantage of that:

```
"cluster": telemetryCfg.GetLabels()["internal_id"],
"__replica": m.cfg.NodeID,
```

`telemetryCfg.GetLabels()["internal_id"]` is just an example, we could actually set `cluster` in `telemetryCfg.Labels` (the server response)

cc @clly

Achooo: I'll update this in a follow up PR, just to keep the PRs small, with the suggestions made for Cortex, ticket CC-4742! Thanks for catching this and the context with pairing.

agent/hcp/client/client.go
```diff
@@ -21,9 +24,19 @@ import (
 //
 //go:generate mockery --name Client --with-expecter --inpackage
 type Client interface {
+	FetchTelemetryConfig(ctx context.Context) (*TelemetryConfig, error)
 	FetchBootstrap(ctx context.Context) (*BootstrapConfig, error)
 	PushServerStatus(ctx context.Context, status *ServerStatus) error
 	DiscoverServers(ctx context.Context) ([]string, error)
+	InitMetricsClient(ctx context.Context, endpoint string) error
```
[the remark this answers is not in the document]
nickethier: What if we define two interfaces in this package. If we had a `client.MetricsClient` interface with `ExportMetrics` then we could create it after calling `Client.FetchTelemetryConfig` and avoid needing the `InitMetricsClient` func

PR #16798, opened 2023-03-28
[CC-4374] Expose MetricsConfig in Deps for HCP Manager

Achooo:
### Description

Change originally authored by @freddygv 

#### Background: HCP Telemetry
TLDR: new feature to ship Server metrics directly to HCP.
- That branch will be merged at the end into `main` and backported

This allows us to iteratively improve on this feature over time before 🚢 

In #16797, we use a go metrics sink to fetch data for a new HCP Metrics Reporter.
This PR is a pre-requisite to expose the backend within the HCP manager via `Deps`:
- See this [line of code](https://github.com/hashicorp/consul/pull/16797/files#diff-2cc91c6c9b4a1393e287b185f5e20cde300cbccdddafb199953094e2308d4eb9R501) for initialization in Manager
- See this [line of code](https://github.com/hashicorp/consul/pull/16797/files#diff-52d267da081e83b27274393deb6de37983af468d2131d82e63103983b2e277baR104) for usage of the `Data()` method 

#### Changes

1. Expose the `Data()` method in the `MetricsHandler` interface
2. Expose the Backend in `Deps` instead of BaseDeps (cleanup)

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/agent/consul/options.go
+++ b/agent/consul/options.go
@@ -16,10 +16,12 @@ import (
 	"github.com/hashicorp/consul/agent/router"
 	"github.com/hashicorp/consul/agent/rpc/middleware"
 	"github.com/hashicorp/consul/agent/token"
+	"github.com/hashicorp/consul/lib"
 	"github.com/hashicorp/consul/tlsutil"
 )
 
 type Deps struct {
+	MetricsConfig    *lib.MetricsConfig
 	EventPublisher   *stream.EventPublisher
 	Logger           hclog.InterceptLogger
 	TLSConfigurator  *tlsutil.Configurator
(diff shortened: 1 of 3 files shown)


agent/consul/options.go
```diff
@@ -1,6 +1,7 @@
 package consul
 
 import (
+	"github.com/hashicorp/consul/lib"
```
loshz, line 4: NIT: needs to be moved down and grouped with the other `github.com/hashicorp/consul` imports.

Achooo: thanks for catching! fixed

PR #16800, opened 2023-03-28
[HCP Observability] Move HCP client to subpackage hcp/client

Achooo:
##### Overall Feature background
This PR is part of the HCP Observability feature to send Server metrics to HCP's new Telemetry Gateway. The idea is to create a new OpenTelemetry go-metrics sink that will be initialized automatically when a Consul server is registered with the HCP management plane.
- That branch will be merged at the end into `main` and back ported

#### Changes

Refactor: move the hcp client and mock into a sub package `client`

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


agent/hcp/client/client.go
```diff
@@ -1,4 +1,4 @@
-package hcp
+package client
```
Achooo, line 1: I considered using `hcpclient` as the package name, but the imports end up being named imports anyways i.e. : `import hcpclient "github.com/hashicorp/consul/agent/hcp/client"`

We'd ideally change the path itself to `/hcp/hcpclient` or something like that, but less repetitive to make this better! I was lacking imagination here so left it as client.

If anyone has any preferences, please let me know :)

loshz: `client` is fine :)

loshz (approved):

PR #16815, opened 2023-03-30
Add manual virtual IP support to state store

kyhavlov:
Follow-up to #16760 - add a manual virtual IPs field that can be used to manually assign multiple virtual IPs to a service, which will be returned through discovery chain lookups by transparent proxies.

### DIFF
--- a/agent/consul/state/catalog.go
+++ b/agent/consul/state/catalog.go
@@ -1090,6 +1090,69 @@ func assignServiceVirtualIP(tx WriteTxn, idx uint64, psn structs.PeeredServiceNa
 	return result.String(), nil
 }
 
+func (s *Store) AssignManualVirtualIPs(idx uint64, psn structs.PeeredServiceName, ips []string) error {
+	tx := s.db.WriteTxn(idx)
+	defer tx.Abort()
+
+	// First remove the given IPs from any existing services, to avoid duplicate assignments.
+	assignedIPs := map[string]struct{}{}
+	for _, ip := range ips {
+		assignedIPs[ip] = struct{}{}
+	}
+	for ip := range assignedIPs {
+		entry, err := tx.First(tableServiceVirtualIPs, indexManualVIPs, psn.ServiceName.PartitionOrDefault(), ip)
+		if err != nil {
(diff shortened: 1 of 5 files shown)


agent/consul/state/catalog.go
```diff
@@ -1090,6 +1090,33 @@ func assignServiceVirtualIP(tx WriteTxn, idx uint64, psn structs.PeeredServiceNa
 	return result.String(), nil
 }
 
+func (s *Store) AssignManualVirtualIPs(idx uint64, psn structs.PeeredServiceName, ips []string) error {
```
hashi-derek, line 1093: This looks good, but I'm starting to wonder if we should be extra-careful and remove old VIPs if there happens to be a conflict when performing this upsert. Do you have any thoughts about this? It's extra effort, but it might save us from a nasty bug where consul-k8s doesn't clean up old VIPs.

kyhavlov: Updated the PR to add this behavior - I think k8s not cleaning up old VIPs is a real concern, though allowing services to "steal" manual VIP entries from others feels a bit off, especially when considering ACLs or namespaces/partitions. Not sure if there's a cleaner solution though - we really want to avoid the case where multiple upstreams with shared manual VIPs end up on the same transparent proxy setup, and checking for duplicates globally seems like the only way.

agent/consul/state/catalog.go
```diff
@@ -1090,6 +1090,65 @@ func assignServiceVirtualIP(tx WriteTxn, idx uint64, psn structs.PeeredServiceNa
...
+	for _, ip := range ips {
+		assignedIPs[ip] = struct{}{}
+	}
+	for ip := range assignedIPs {
+		entry, err := tx.First(tableServiceVirtualIPs, indexManualVIPs, ip)
+		if err != nil {
+			return fmt.Errorf("failed service virtual IP lookup: %s", err)
+		}
+
+		if entry == nil {
```
hashi-derek, line 1108: Should we skip our own service name if it's found by this query?

hashi-derek (approved):

PR #16817, opened 2023-03-30
storage: fix resource leak in `Watch`

boxofrad:
### Description

While reviewing the existing controller code, I realized that when implementing watches, I'd forgotten that we need to call `Unsubscribe` on the event publisher subscription to free its resources when we're done.


analogue (approved):
🚀

Issue #16779, opened 2023-03-26
Consul connect service mesh connections may fail after 72 hours


mmeier86:
#### Overview of the Issue

I'm running a Nomad cluster using Consul Connect service mesh for almost all jobs. Since the update of Consul to 1.15.1 from 13.5, every three days, the Consul Connect mesh stops working. This seems to happen reliably 72 hours after a job (and its accompanying Envoy sidecar proxy) have been started.

I have also tried out the "consul troubleshoot proxy" command, and finally found a potential error source:

```
nsenter -t 691115 -n consul troubleshoot proxy -envoy-admin-endpoint=127.0.0.2:19001 -upstream-ip=127.0.0.1
==> Validation                                             
 ! Certificate chain is expired
  -> Check the logs of the Consul agent configuring the local proxy and ensure XDS updates are being sent to the proxy
 ! Certificate chain is expired
  -> Check the logs of the Consul agent configuring the local proxy and ensure XDS updates are being sent to the proxy
 ! Certificate chain is expired
  -> Check the logs of the Consul agent configuring the local proxy and ensure XDS updates are being sent to the proxy
 ✓ Envoy has 0 rejected configurations
 ✓ Envoy has detected 1464 connection failure(s)
 ! No listener for upstream "127.0.0.1" 
  -> Check that your upstream service is registered with Consul
  -> Make sure your upstream exists by running the `consul[-k8s] troubleshoot upstreams` command
  -> If you are using transparent proxy for this upstream, ensure you have set up allow intentions to the upstream
  -> Check the logs of the Consul agent configuring the local proxy to ensure XDS resources were sent by Consul
 ! No clusters found on route or listener
```

This seems to fit the "failing 72 hours after job start" symptoms, if indeed the problem is that the mTLS certs are not getting renewed. 

Also note that I upgraded directly from 1.3.5 to 1.5.1, including the necessary changes for the breaking grpc_tls change. I did this relatively late update due to experiencing the Nomad config issues with gRPC and the Consul connect envoy proxy setup. Is it possible that I got the grpc_tls config wrong somewhere, and Consul is not able to update the Envoy proxy with new certificates?

One last note: Taking down the downstream job and restarting it (`nomad job stop` and `nomad job start`) seems to completely fix the issue.
---

#### Reproduction Steps

Steps to reproduce this issue, eg:

1. Setup a Consul and Nomad cluster with three servers and two nodes
2. Enable Consul Service mesh, using Consul's internal CA as the Connect CA
[...]

PR #16818, opened 2023-03-30
Revert "cache: refactor agent cache fetching to prevent unnecessary f…

hashi-derek:
This reverts commit 3c44116a8f199b4003c68cb52a2579fc3a7746bb.

The revert was performed due to user reports of leaf certificates not renewing properly. https://github.com/hashicorp/consul/issues/16779

Since users are reporting that services worked appropriately in 1.14 and not 1.15, this revert attempts to mitigate those problems by swapping back to the original 1.14 behavior. Further investigation will need to be performed to determine why the new 1.15 changes result in this behavior.

Manual testing efforts on a modified version of the code show that the calls to the agent cache may occasionally never return when attempting to wait on a channel that is returned from this path: https://github.com/hashicorp/consul/blob/release/1.15.1/agent/cache/cache.go#L608

At this time, it is believed that a race condition prevents the channel from being correctly closed, which results in the stale leaf certificates.

### DIFF
--- a/.changelog/14956.txt
+++ b/.changelog/14956.txt
@@ -1,3 +1,0 @@
-```release-note:bug
-cache: refactor agent cache fetching to prevent unnecessary fetches on error
-```
--- a/.changelog/14956.txt
+++ b/.changelog/14956.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cache: refactor agent cache fetching to prevent unnecessary fetches on error
+```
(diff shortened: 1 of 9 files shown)


erichaberkorn (approved):

kisunji (approved):
I think we should keep the changelog and make a new one calling out the revert

kisunji:
Maybe something like: 
```
```release-note:bug
cache: revert cache refactor which could cause blocking queries to never return
``````

PR #16827, opened 2023-03-30
ci: always test oss and conditionally test enterprise

jmurret:
### Description

This has a passing CI run in enterprise: https://github.com/hashicorp/consul-enterprise/actions/runs/4568643952

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


.github/workflows/go-tests.yml
```diff
@@ -215,6 +215,22 @@ jobs:
       runner-count: 12
       runs-on: ${{ needs.setup.outputs.compute-xl }}
       repository-name: ${{ github.repository }}
+      go-tags: ""
+    secrets:
+      elevated-github-token: ${{ secrets.ELEVATED_GITHUB_TOKEN }}
```
loshz, line 220: Do we need this for OSS?

loshz: Yeah, I think passing an empty string for OSS makes sense.

loshz (approved):

PR #16842, opened 2023-03-31
Check acls on resource `Read`, `List`, and `WatchList`

analogue:
### Description

Apply ACLs on `Read`, `List`, and `WatchList` endpoints for the resource service.  ACLs for `Write` and `Delete` will be done in a separate PR.

### Testing
- `gotestsum --format=testname ./agent/grpc-external/services/resource -v`

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/consul/server.go
```diff
@@ -1250,7 +1250,7 @@ func (s *Server) setupExternalGRPC(config *Config, logger hclog.Logger, backend
 	registry := resource.NewRegistry()
 
 	if s.config.DevMode {
-		demo.Register(registry)
+		demo.Register(registry, nil)
```
analogue, line 1253: I didn't want to change the signature to `Register` to pass in ACLs, but the `Registration` object is effectively immutable post-registration since it passed around by value. Open to better suggestions for this as our tests scale.

agent/grpc-external/services/resource/read.go
```diff
@@ -13,9 +13,21 @@ import (
...
 func (s *Server) Read(ctx context.Context, req *pbresource.ReadRequest) (*pbresource.ReadResponse, error) {
 	// check type exists
-	if _, err := s.resolveType(req.Id.Type); err != nil {
+	reg, err := s.resolveType(req.Id.Type)
+	if err != nil {
+		return nil, err
+	}
+
+	// check acls
+	authz, err := s.ACLResolver.ResolveTokenAndDefaultMeta("??", nil, nil)
```
analogue, line 22: Should I expect a token id in the context and extract it?

internal/resource/registry.go
```diff
@@ -58,6 +77,19 @@ func (r *TypeRegistry) Register(registration Registration) {
...
+	}
+	if registration.ACLs.Read == nil {
+		registration.ACLs.Read = func(authz acl.Authorizer, id *pbresource.ID) error {
+			if err := authz.ToAllowAuthorizer().OperatorReadAllowed(&acl.AuthorizerContext{}); err != nil {
+				return err
+			}
+			return nil
+		}
+	}
+	// TODO: write & list hooks
```
analogue, line 92: Maybe this should be refactored into a `NewRegistration(type, proto, acls, etc)`?

analogue (commented):
@boxofrad Just looking for some feedback before I rubberstamp this general pattern for other endpoints.

internal/resource/registry.go
```diff
@@ -58,6 +77,25 @@ func (r *TypeRegistry) Register(registration Registration) {
 		panic(fmt.Sprintf("resource type %s already registered", key))
 	}
 
+	// set default acl hooks for those not provided
```
boxofrad, line 80: I really like that you defaulted these instead of adding `nil` checks in the handlers 👏🏻

agent/grpc-external/services/resource/watch.go
```diff
@@ -1,16 +1,40 @@
...
-	if _, err := s.resolveType(req.Type); err != nil {
+	reg, err := s.resolveType(req.Type)
+	if err != nil {
 		return err
 	}
 
+	// check acls
+	// TODO(spatel): FIXME Where do I get the ctx in order to extract the token?
+	ctx := context.Background()
+	authz, err := s.ACLResolver.ResolveTokenAndDefaultMeta(tokenFromContext(ctx), nil, nil)
```
boxofrad, line 25: Using `stream.Context()` here should do the trick.

agent/grpc-external/services/resource/watch.go
```diff
@@ -1,16 +1,40 @@
...
+	if err != nil {
+		return fmt.Errorf("getting authorizer: %w", err)
```
boxofrad, line 27: Although gRPC will automatically treat this as an `INTERNAL` error, it might be clearer to be explicit about the status code here (and in a handful of other places I see).

```suggestion
		return status.Errorf(codes.Internal, "failed to get authorizer: %v", err)
```

agent/grpc-external/services/resource/watch.go
```diff
@@ -33,6 +57,16 @@ func (s *Server) WatchList(req *pbresource.WatchListRequest, stream pbresource.R
...
+		// filter out items that don't pass read ACLs
+		if err = reg.ACLs.Read(authz, event.Resource.Id); err != nil {
+			switch {
+			case acl.IsErrPermissionDenied(err):
+				continue
+			default:
+				return fmt.Errorf("authorizing read: %w", err)
+			}
+		}
+
```
boxofrad, line 69: nit: you could reduce the level of nesting here by combining the `if` and `switch`.

```suggestion
		err = reg.ACLs.Read(authz, event.Resource.Id)
		switch {
		case acl.IsErrPermissionDenied(err):
			continue
		case err != nil:
			return fmt.Errorf("authorizing read: %w", err)
		}
```

agent/grpc-external/services/resource/list_test.go
```diff
@@ -131,6 +131,92 @@ func TestList_VerifyReadConsistencyArg(t *testing.T) {
 	}
 }
 
+func TestList_ACL_RegisteredHooks(t *testing.T) {
```
boxofrad, line 134: TODO: @boxofrad 

Write a comment about testing these generally.

agent/grpc-external/services/resource/server.go
```diff
@@ -55,7 +63,17 @@ func (s *Server) Delete(ctx context.Context, req *pbresource.DeleteRequest) (*pb
 	return &pbresource.DeleteResponse{}, nil
 }
 
-//nolint:unparam
+// Extracts ACL token ID from incoming context
+func tokenFromContext(ctx context.Context) string {
+	// TODO(spatel): Figure out where to get the token from correctly. This is just a stand in for the PR
+	//               Should I be using QueryOptions like some of the other gRPC servers?
+	token := ctx.Value("x-token")
```
boxofrad, line 70: Generally, the ACL token is passed around via a gRPC metadata field called `x-consul-token`. I don't think we should carry over the `QueryOptions` stuff here because it contains a lot of stuff that doesn't apply to this API.

agent/consul/server.go
boxofrad: That's a really good point! Ideally, I think the `demo` package should have a realistic set of ACL hooks (e.g. `demo.v1.artist:write`, `demo.v1.album:read`), and rather than mocking out the hooks in our tests, we should construct authorizers using real policy HCL ([example](https://github.com/hashicorp/consul/blob/0582f137c53f6200575eea5a2aef8257e0f0e805/agent/grpc-external/testutils/acl.go#L46-L63)).

Unfortunately, our current ACL system (like most of our stack) requires _manually_ adding new resource types and stanzas - something we'll need to address in the future, but out of scope for our current work. 😞 

I wonder if, for now, we could fake it using K/V permissions? For example, `demo.v1.artist:write` could be implemented as:

```Go
func(authz acl.Authorizer, res *pbresource.Resource) error {
	key := fmt.Sprintf("resource/%s/%s", resource.ToGVK(res.ID.Type), res.ID.Name)
	return authz.ToAllowAuthorizer().KeyWriteAllowed(key, &acl.AuthorizerContext{})
}
```

Then, for tests, the `demo` package could expose a policy like this:

```Go
const ArtistV1WritePolicy = `key_prefix "resource/demo.v1.artist/" { policy = "write" }`
```

Which we could use to construct a policy authorizer in our endpoint tests, and we'd no longer need that second argument to `demo.Register`.

Slightly whacky idea. What do you think? 😄

boxofrad (commented):
This is looking great! I've left a handful of small comments, and a more general thought on testing/that second argument to `demo.Register`. Let me know if you'd like to chat about it synchronously 📞

agent/grpc-external/services/resource/watch.go
analogue: Fixed in https://github.com/hashicorp/consul/pull/16842/commits/b353ea51da4a42e8af8b3ceca67bfc312b0934ef

agent/grpc-external/services/resource/watch.go
analogue: 👍 Fixed in 
- https://github.com/hashicorp/consul/pull/16842/commits/ee5b8bdc94022489faf3794aaff811869999a0c0
- https://github.com/hashicorp/consul/pull/16842/commits/d998f82794d593ade6f8dc700976298d9694ee47
- https://github.com/hashicorp/consul/pull/16842/commits/add7f76e42e5b4bf951c90ab9705121be7edd2ba

agent/grpc-external/services/resource/watch.go
analogue: 👍 Fixed

agent/grpc-external/services/resource/server.go
analogue: 👍 Fixed

agent/consul/server.go
analogue: Sounds worth trying. I did this for the `Read` tests only in https://github.com/hashicorp/consul/pull/16842/commits/45352a443123a65c783a9d4d17298e79cfaadec1. Let me know what you think.

analogue:
@boxofrad Ready for another pass when you get a chance.

agent/consul/server.go
boxofrad: It's a little odd, for sure, but I think it actually works quite well as a stand-in until our ACL system can be more modularly extended (or support generic resource permissions). What do you think?

If we keep it, we should probably add some comments explaining why we're using K/V rules in this way. I also notice that the `Read` tests exercise the cases where `Registration.ACLs` is nil or missing hooks, IMO we should cover this in the registry's unit tests instead (so we can go ahead and remove that second argument to `demo.Register`).

boxofrad (commented):
Thanks for incorporating my feedback (and entertaining my whacky ACL policy idea!). This looks great ⭐

agent/consul/server.go
```diff
@@ -1237,7 +1247,17 @@ func (s *Server) setupExternalGRPC(config *Config, logger hclog.Logger) {
...
+	registry := resource.NewRegistry()
+
+	if s.config.DevMode {
+		demo.Register(registry, nil)
+	}
+
+	resourcegrpc.NewServer(resourcegrpc.Config{
+		Registry: registry,
+		Backend:  backend,
+		Logger:   logger.Named("grpc-api.resource"),
```
boxofrad, line 1259: Remember to pass the `ACLResolver` in here 🙇🏻

agent/consul/server.go
analogue: Yeah, I think it is worth keeping as a workaround until we're further down the road and it does the job for now. I'll go ahead and apply this to the other endpoints.

agent/consul/server.go
analogue: Nice catch! Fixed in https://github.com/hashicorp/consul/pull/16842/commits/04818a63a4e1e069bba34b6a9aa7a99b9b0afa4b

analogue:
@boxofrad Ok, I've completed the remaining ACL tests for the `WatchList` and `List` endpoints and added tests for the default acl hooks to the `Registry` tests.

boxofrad (approved):
Fantastic work! Thanks again 🙌🏻 

I've left one last comment about possibly moving the ACL policy strings, but it's definitely not a blocker.

PR #16862, opened 2023-04-03
Include virtual services from discovery chain in intention topology

kyhavlov:
This PR causes 'virtual services' represented by service resolver/router/splitter config entries but without any existing service instances to be returned in intention topology lookups. This is to enable creating the config entry for the service ahead of time so that there will be a virtual IP for transparent proxies to use for these upstreams.

### DIFF
--- a/agent/consul/state/intention_test.go
+++ b/agent/consul/state/intention_test.go
@@ -2161,6 +2161,7 @@ func TestStore_IntentionTopology(t *testing.T) {
 		name            string
 		defaultDecision acl.EnforcementDecision
 		intentions      []structs.ServiceIntentionsConfigEntry
+		discoveryChains []structs.ConfigEntry
 		target          structs.ServiceName
 		downstreams     bool
 		expect          expect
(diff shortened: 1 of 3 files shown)


agent/consul/state/intention_test.go
```diff
@@ -2192,6 +2193,68 @@ func TestStore_IntentionTopology(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:            "(upstream) acl allow includes virtual service",
+			defaultDecision: acl.Allow,
+			discoveryChains: []structs.ConfigEntry{
+				&structs.ServiceResolverConfigEntry{
+					Kind: structs.ServiceIntentions,
```
hashi-derek, line 2201: Shouldn't the `Kind` here be a service resolver?

agent/consul/state/intention_test.go
```diff
@@ -2192,6 +2193,68 @@ func TestStore_IntentionTopology(t *testing.T) {
...
+					},
+				},
+			},
+		},
+		{
+			name:            "(upstream) acl deny all intentions allow virtual service",
+			defaultDecision: acl.Deny,
+			discoveryChains: []structs.ConfigEntry{
+				&structs.ServiceResolverConfigEntry{
+					Kind: structs.ServiceIntentions,
```
hashi-derek, line 2230: Same here.

hashi-derek (approved):

PR #16871, opened 2023-04-04
Memdb Txn Commit race condition fix

dhiaayachi:
### Description

a hook to publish events when a commit happen to memdb was introduced in [here](https://github.com/hashicorp/consul/commit/2a8a8f7b8d31aeb83c24cfe67c78acfae1f154bd). The event publishing would happen before the commit is done to prevent out of order event publishing (as long as the commit is not done the transaction lock will be held in memdb and no other transaction will be able to perform a write). This create a race condition, where an event could be issued to a consumer before the data is committed into memdb. The consumer risk to read a stale data.

Until recently and before introducing the controller pattern in https://github.com/hashicorp/consul/pull/16058, this was not an issue as the event will have all the relevant data attached to it and there is no need to read the data from memdb again after receiving it.

The controller pattern introduced in https://github.com/hashicorp/consul/pull/16058 change that and, by design, would make the event consumer read the data from memdb after consuming the event. Furthermore the issued event will not have any version identifying data that the consumer can use to identify if it's reading stale data.

This PR adds a test that reproduce the race by artificially introducing a delay into the publisher hook and fix the race condition by making the publish hook called after the commit. To avoid out of order published event, a lock is introduced to guarantee that event order is respected.

A test was added to reproduce the issue and verify that the fix pass it.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


mkeeler (approved):

.changelog/16871.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+Fix a race condition where an event is published before the data associated is commited to memdb..
```
boxofrad, line 2: nit (sorry!)

```suggestion
Fix a race condition where an event is published before the data associated is commited to memdb.
```

agent/consul/state/memdb_test.go
```diff
@@ -0,0 +1,95 @@
...
+		},
+	}
+}
+
+type TestObject struct {
+	ID  string
+	Foo string
+}
+
+func Test_txn_Commit(t *testing.T) {
```
boxofrad, line 39: Would you mind adding a comment here to explain what this test is exercising? If I understand correctly:

- `ch1` is used to trigger a read when the writer publishes its events.
- `ch2` is used to block the publish method from returning until the read has happened.

Before your fix, the test would fail because `ch2` being blocked would ensure the read happened before the transaction was committed. Now, we commit the transaction first. Is that about right?

agent/consul/state/memdb.go
```diff
@@ -140,7 +137,11 @@ type txn struct {
 	// Index is stored so that it may be passed along to any subscribers as part
 	// of a change event.
 	Index   uint64
-	publish func(tx ReadTxn, changes Changes) error
+	publish publishFuncType
+
+	prePublish prePublishFuncType
```
boxofrad, line 142: Do we still need this to be a struct member, rather than just calling `processChanges` directly in `Commit`?

agent/consul/state/memdb.go
```diff
@@ -161,16 +162,30 @@ func (tx *txn) Commit() error {
 		}
 	}
 
-	// publish may be nil if this is a read-only or WriteTxnRestore transaction.
-	// In those cases changes should also be empty, and there will be nothing
-	// to publish.
-	if tx.publish != nil {
-		if err := tx.publish(tx.Txn, changes); err != nil {
+	// This lock prevent concurrent commits to get published out of order.
```
boxofrad, line 165: nit:

```suggestion
	// This lock prevents events from concurrent transactions getting published out of order.
```

boxofrad (approved):
Looks great, nice job! 👏🏻  I've left a couple of stylistic comments inline, but please feel free to go ahead without them.

agent/consul/state/memdb.go
dhiaayachi: Not sure that's feasible without major changes to the data struct, maybe I'm not understanding 🤔. Also, it could be handy to have the prePublish mockeable for future tests.

PR #16747, opened 2023-03-22
CA mesh CA expiration to it's own section

eikenb:
This is part of an effort to raise awareness that you need to monitor your mesh CA if coming from an external source as you'll need to manage the rotation.

PR #16874, opened 2023-04-04
highlight the agent.tls cert metric with CA ones

eikenb:
Related to #16747, this adds the agent.tls certificate to the section on important cert's that need monitoring.

### DIFF
--- a/website/content/docs/agent/telemetry.mdx
+++ b/website/content/docs/agent/telemetry.mdx
@@ -100,7 +100,7 @@ These are some metrics emitted that can help you understand the health of your c
 ** Why they're important:** Consul Mesh requires a CA to sign all certificates
 used to connect the mesh and the mesh network ceases to work if they expire and
 become invalid. The Root is particularly important to monitor as Consul does
-not automatically rotate it. The third metric is to monitor the certificate
+not automatically rotate it. The TLS certificate metric monitors the certificate
 that the server's agent uses to connect with the other agents in the cluster.
 
 ** What to look for:** The Root CA should be monitored for an approaching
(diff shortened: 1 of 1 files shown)


website/content/docs/agent/telemetry.mdx
```diff
@@ -93,19 +93,23 @@ These are some metrics emitted that can help you understand the health of your c
...
-expiration, to indicate it is time for you to rotate the "root" CA either manually or with external automation.
-The signing (intermediate) certificate should be
-rotated automatically by Consul, but should be monitored in case of rotation isn't working;
-in this scenario, check the server agent logs for messages related to the CA system.
+expiration, to indicate it is time for you to rotate the "root" CA either
+manually or with external automation. The signing (intermediate) certificate
+should be rotated automatically by Consul, but should be monitored in case of
+rotation isn't working; in this scenario, check the server agent logs for
+messages related to the CA system. The agent certificate's rotation handling
+varies based on the configuration.
```
eikenb, line 112: I was kind of hoping to have this be a link to some docs on server agent configuration, but I can't find anything. If you know of a good spot in the docs that covers this please let me know. Thanks.

jkirschner-hashicorp: Unfortunately, I don't think we have a good reference in the docs today for talking about when rotation is manual vs automated by Consul for both server and client agent certs.

It's automated for client agent certs when using auto-encrypt or auto-config, and automated for server agent certs when using Consul on Kubernetes (I _think_ ... I'm not sure if it depends on whether the Vault secrets backend is in use). We might want to note that as something to follow up on at some point in the Jira ticket, even if we don't do it now.

I'll have to ask someone familiar with the k8s integration about the server agent cert rotation stuff.

eikenb: We resolved these topics in chat.

PR #16886, opened 2023-04-05
boxofrad/resource write status endpoint

boxofrad:
- storage: check Uid before Version
- resource: add demo types
- resource: accept protobuf message type in registration
- resource: `Write` endpoint
- resource: wire up Registry, Backend, and demo types
- resource: use demo types in List, WatchList and Read endpoints
- resource: clarify comments and simplify handler code
- resource/write: fix potential bug where statuses are overwritten
- resource: finalize status format
- resource: WriteStatus endpoint

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


analogue (approved):
Looking good! I think docs for the proto file in this PR would be helpful. For example, it is hard to evaluate the need for `Reference.Section` without the additional context of the GW api use cases.

PR #16888, opened 2023-04-05
Update list of Envoy versions

nathancoleman:
### Description

Several new patch versions of Envoy have been released. This updates the list of supported Envoy versions to include those patch releases.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


david-yu (approved):

.circleci/config.yml
```diff
@@ -19,10 +19,10 @@ references:
     BASH_ENV: .circleci/bash_env.sh
     GO_VERSION: 1.20.1
   envoy-versions: &supported_envoy_versions
-    - &default_envoy_version "1.21.5"
-    - "1.22.5"
-    - "1.23.2"
-    - "1.24.0"
+    - &default_envoy_version "1.21.6"
```
curtbushko, line 22: Doing this here might not be the correct place because of the migration to Github Actions?

curtbushko (approved):

.circleci/config.yml
nathancoleman: I chatted w/ @jmurret and am going to commit envoy version updates to https://github.com/hashicorp/consul/pull/16915

nathancoleman: Also chatted about a backport that incorrectly updated these versions. Temporarily reverting them here, but there's no real impact since this workflow will be removed shortly

david-yu:
@nathancoleman Envoy just released 1.24.6 and 1.23.8 today. Perhaps makes sense to update this list again in the same set of PRs? https://www.envoyproxy.io/docs/envoy/v1.24.6/version_history/v1.24/v1.24.6

PR #16915, opened 2023-04-07
ci: add test-integrations

jmurret:
### Description
This configuration currently works in ENT but fails in OSS.  GitHub is helping us look at it.  The docker based tests in `compatability-tests` and `upgrade-test` fail with intermittent communication failures when docker containers need to talk to each other as is the case for 3 servers joining in a cluster.

As part of troubleshooting that issue, most iteration has been done on [this](https://github.com/hashicorp/consul/tree/jm/hack-int) which comments out all of the envoy, nomad, and vault stuff so that it only runs the two jobs with docker based tests.

### Remaining items Checklist

- [x] get docker based compatibility tests and upgrade tests to pass in OSS.  They currently pass in ENT.
- [x] revisit testsplitting for envoy
- [x] audit commands for GHA vs CircleCI in OSS and ENT
- [x] ensure envoy, nomad, vault versions match CircleCI for OSS and ENT

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -346,6 +346,7 @@ jobs:
             sudo tar -C /usr/local -xzvf go${GO_VERSION}.linux-amd64.tar.gz
           environment:
             <<: *ENVIRONMENT
+      - run: go env
       - run: *install-gotestsum
       - run: docker build -t consul:local -f ./build-support/docker/Consul-Dev.dockerfile .
       - run:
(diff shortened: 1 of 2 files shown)


nathancoleman:
Noting that 1.13.x and 1.14.x do not support the same versions of Envoy, so the backports will need a little manual intervention RE: https://github.com/hashicorp/consul/pull/16915/commits/9e7728077f855422bf31eb0082fdbc6d1ba60c68

loshz (approved):
LGTM :shipit:

PR #16916, opened 2023-04-07
Update HCP bootstrapping to support existing clusters

freddygv:
Replaces #16788

### Description
We want to allow users to link Consul clusters that already exist to
HCP. Existing clusters need care when bootstrapped by HCP, since we do
not want to do things like change ACL/TLS settings for a running
cluster.

The plan is to only send down a management token to existing clusters
and nothing else (at least initially). 

This PR adds a few commits to that end:

- Add an HCP management token to server config so that it can be persisted
  through Raft as we do for the initial management token.
- Update the HCP SDK to be able to receive this new management token.
- Update the bootstrapping logic to account for the different types of
  clusters that are going to be supported.

**Best reviewed by commit**

- [x] Unit tests

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### TODO
* [ ] Manual testing


agent/config/runtime.go
```diff
@@ -1739,6 +1739,9 @@ func (c *RuntimeConfig) Sanitized() map[string]interface{} {
 
 // IsCloudEnabled returns true if a cloud.resource_id is set and the server mode is enabled
 func (c *RuntimeConfig) IsCloudEnabled() bool {
+	if c == nil {
+		return false
```
jjacobson93, line 1743: What's the case where `RuntimeConfig` is `nil` here?

freddygv: There are some tests in `command/agent` that hit that and panic'd: https://app.circleci.com/pipelines/github/hashicorp/consul/56048/workflows/a6e357e8-4c0f-49a7-a22b-6872e2f391f8/jobs/1349145/parallel-runs/3

agent/config/builder.go
```diff
@@ -2528,17 +2528,20 @@ func validateAutoConfigAuthorizer(rt RuntimeConfig) error {
 }
 
 func (b *builder) cloudConfigVal(v *CloudConfigRaw) (val hcpconfig.CloudConfig) {
+	val.ResourceID = os.Getenv("HCP_RESOURCE_ID")
```
jmurret, line 2531: I think you will need to use `LookupEnv` here similar to [what the SDK uses](https://github.com/hashicorp/hcp-sdk-go/blob/dbc8bfc75af83d596ca8d33705c466f1ef43708f/config/env.go#L67) or else this will set an empty string if it is null.  Previous iterations of the SDK also checked for null rather than empty string and would fail.  So, probably best to only set this when the env var is actually configured.

agent/consul/leader.go
```diff
@@ -444,72 +444,22 @@ func (s *Server) initializeACLs(ctx context.Context) error {
...
-					if _, err := s.raftApply(structs.ACLTokenSetRequestType, &req); err != nil {
-						return fmt.Errorf("failed to create initial management token: %v", err)
-					}
+		}
 
-					s.logger.Info("Created ACL initial management token from configuration")
-				}
+		// Check for configured management token from HCP
+		if hcpManagement := s.config.Cloud.ManagementToken; len(hcpManagement) > 0 {
+			err := s.initializeManagementToken("HCP Management Token", hcpManagement)
```
jmurret, line 455: Can we add a comment here for "do not override an initial management token provided by the operator in the configuration."?

~If it is easy enough to test `initializeManagementToken()` in isolation, ca we also add a test that ensures this behavior that the additional CCM management token is persisted but does not replace the Initial Management Token provided by the operator?~. NVM, it looks like `bootstrap with both` from `TestLeader_ACL_Initialization()` does this.

jmurret (changes requested):
👏 nice work.  just had two small things, but one could possibly cause a subtle bug, so asking for changes/discussion.

agent/config/builder.go
freddygv: The `val` variable here is already a zero-value of `hcp.CloudConfig` because of the named return, so if `HCP_RESOURCE_ID` is not set then it stays empty. The SDK is checking that it's set first because that function is mutating potentially non-empty config. 

I updated this to make the `val` variable declaration clearer. This isn't really a case to use named returns: [style guide](https://eng-handbook.hashicorp.services/language-style/go/#return-values).

agent/consul/leader.go
freddygv: Yep

hanshasselberg (approved):
Thanks, this looks great! I like the combined flow and the tests!

I left a couple nits and I have one question: When a cluster with consul 1.14 with CCM integration is upgraded to Consul 1.15, will it rebootstrap because of the missing sucess marker?

agent/hcp/bootstrap/bootstrap.go
```diff
@@ -356,23 +338,31 @@ func persistTLSCerts(dataDir string, serverCert, serverKey string, caCerts []str
...
+}
+
+func persistManagementToken(dir, token string) error {
+	name := filepath.Join(dir, tokenFileName)
 	return os.WriteFile(name, []byte(token), 0600)
 }
 
-func persistBootstrapConfig(dataDir, cfgJSON string) error {
+func persistBootstrapConfig(dir, cfgJSON string) error {
+	if cfgJSON == "" {
```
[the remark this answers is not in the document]
freddygv: What if we send that empty JSON string down from HCP? It would simplify some of the conditional logic in `persistAndProcessConfig`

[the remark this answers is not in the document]
freddygv: Take a look at the last commit, I updated that bit to expect `{}` from CCM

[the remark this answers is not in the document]
hanshasselberg: I like the idea to always send valid json 👍

jmurret (approved):
Nice work! 👍

PR #16919, opened 2023-04-07
Test: add noCleanup to TestServer stop

huikang:
### Description

`noCleanup` exists in the test code to preserve the data after unit test run.

https://github.com/hashicorp/consul/blob/430df05e61a017627f4e5eb5e38513840f6172d1/sdk/testutil/io.go#L12

`noCleanup` has been used in someplace like 

https://github.com/hashicorp/consul/blob/430df05e61a017627f4e5eb5e38513840f6172d1/sdk/testutil/io.go#L18

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


eculver (approved):
Yep, we need this.

PR #16931, opened 2023-04-08
upgrade test: config nodeName, nodeid, and inherited persistent data for consul container

huikang:
### Description
This PR allows configuring Node name, node Id, and external consul data to copy from for the consul container.
With this change, we can do upgrade test if a newer version of consul can use the data generated from an older version of consul.

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/test/integration/consul-container/libs/cluster/container.go
+++ b/test/integration/consul-container/libs/cluster/container.go
@@ -16,6 +16,7 @@ import (
 	goretry "github.com/avast/retry-go"
 	dockercontainer "github.com/docker/docker/api/types/container"
 	"github.com/hashicorp/go-multierror"
+	"github.com/otiai10/copy"
 	"github.com/pkg/errors"
 	"github.com/testcontainers/testcontainers-go"
 	"github.com/testcontainers/testcontainers-go/wait"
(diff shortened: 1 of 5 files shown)


test/integration/consul-container/libs/cluster/container.go
```diff
@@ -111,6 +116,14 @@ func NewConsulContainer(ctx context.Context, config Config, cluster *Cluster, po
 		return nil, fmt.Errorf("error chowning data directory %s: %w", tmpDirData, err)
 	}
 
+	if config.ExternalDataDir != "" {
+		// copy consul persistent state from an external dir
+		err := copy.Copy(config.ExternalDataDir, tmpDirData)
```
eculver, line 121: This seems fine since we're making the temporary directory on the fly, but I'm curious if it makes sense to just use the given `config.ExternalDataDir` below as the `dataDir` value and skip creating a temporary directory and having to copy the contents to it here? If we didn't need to copy the data, we wouldn't need the new dependency and this code could be skipped.

Just a thought.

eculver (approved):
This looks good. I am curious about needing to do the copy at all though.

test/integration/consul-container/libs/cluster/container.go
huikang: Thanks for the comments. I was thinking using the ExternalDataDir as well. however, I think it's better copying the data because

1. The container test framework will create another files in `tmpDirData` such as the tls files, so generating these file might not be possible in `ExternalDataDir`.
2. `ExternalDataDir` could be reused for the test where it comes from, so keeping a copy might be easier for separating the state between these tests.

PR #16933, opened 2023-04-08
Add support for LocalIdleTimeoutMs on service-defaults

ilpianista:
### Description

In current implementation, local_idle_timeout_ms can only be configured either in proxy.config of individual service instances or the global proxy-defaults, but we need to use the service-defaults config-entry to propagate the values only to the proxies of the selected service.

Following is an example of the service-defaults entry with the added field.

```
Kind      = "service-defaults"
Name      = "backend"
Protocol  = "grpc"

LocalRequestTimeoutMs = 11000
LocalIdleTimeoutMs = 11000
```

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


trujillo-adam (approved):
The docs look good - approving on behalf of consul-docs

ilpianista:
Up? Who should review this?

ilpianista:
@hashi-derek maybe you could have a review at this? :pray:

PR #16950, opened 2023-04-10
Resource validation hook for `Write` endpoint

analogue:
### Description

See NET-2694. 

Existing and new tests pass

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/grpc-external/services/resource/write.go
```diff
@@ -37,6 +37,10 @@ func (s *Server) Write(ctx context.Context, req *pbresource.WriteRequest) (*pbre
 		)
 	}
 
+	if err = reg.Validate(req.Resource); err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "failed validation: %v", err)
```
boxofrad, line 41: I wonder if it might be better to remove the `failed validation: ` prefix, as our own structural validation (e.g. checking `resource.id` is non-nil) doesn't have it?

internal/resource/demo/demo.go
```diff
@@ -84,6 +85,32 @@ func Register(r resource.Registry) {
...
+				return err
+			}
+			if artist.Name == "" {
+				return fmt.Errorf("artist.name required")
+			}
+			return nil
+		}
+
+		return fmt.Errorf("unsupported artist type %v", resource.ToGVK(res.Id.Type))
+	}
```
boxofrad, line 112: Is there any reason not to split this into two separate functions?

internal/resource/demo/demo.go
```diff
@@ -84,6 +85,32 @@ func Register(r resource.Registry) {
...
 	}
 
+	validateArtistFn := func(res *pbresource.Resource) error {
+		if proto.Equal(res.Id.Type, TypeV1Artist) {
+			artist := &pbdemov1.Artist{}
+			if err := anypb.UnmarshalTo(res.Data, artist, proto.UnmarshalOptions{}); err != nil {
+				return err
+			}
+			if artist.Name == "" {
+				return fmt.Errorf("artist.name required")
```
boxofrad, line 95: Out of scope for this PR, but seeing this makes me wonder whether we should have some basic helpers for constructing validation errors.

For example:

```Go
return resource.RequiredFieldError("name")
```

boxofrad (approved):
Nice job 👏🏻

internal/resource/demo/demo.go
analogue: 🤦 Fixed in https://github.com/hashicorp/consul/pull/16950/commits/f47554ba639d56e060827dec87b55d89201b6fa9

PR #16978, opened 2023-04-12
server: wire up in-process Resource Service

boxofrad:
### Description

This PR adds an instance of the Resource Service that can be interacted with in-process, without going over the network, and without authentication/authorization. It will be used for purely internal operations (e.g. by controllers).

### Testing

Manually tested by adding an RPC to the server setup code, and logging the response.

I think this is the kind of thing that's best exercised by a full-stack integration test, once the rest of the controller runtime pieces are in-place.

### DIFF
--- a/acl/resolver/danger.go
+++ b/acl/resolver/danger.go
@@ -1,0 +1,15 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package resolver
+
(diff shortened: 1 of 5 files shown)


analogue (approved):
Looks good to me! 👏  
FYI, this [in-process transport issue](https://github.com/grpc/grpc-go/issues/906) (which is still open) was an interesting read while doing a bit of research on the topic.

PR #16982, opened 2023-04-12
controller: make the `WorkQueue` generic

boxofrad:
### Description

Makes the `WorkQueue` generic (and moves into a `queue` sub-package) so that it can be reused in the new resource-based controller implementation.

### DIFF
--- a/agent/consul/controller/controller.go
+++ b/agent/consul/controller/controller.go
@@ -11,11 +11,13 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/hashicorp/go-hclog"
+	"golang.org/x/sync/errgroup"
+
+	"github.com/hashicorp/consul/agent/consul/controller/queue"
 	"github.com/hashicorp/consul/agent/consul/state"
 	"github.com/hashicorp/consul/agent/consul/stream"
 	"github.com/hashicorp/consul/agent/structs"
-	"github.com/hashicorp/go-hclog"
-	"golang.org/x/sync/errgroup"
 )
 
 // much of this is a re-implementation of
@@ -53,7 +55,7 @@ type Controller interface {
 	// WithQueueFactory allows a Controller to replace its underlying work queue
 	// implementation. This is most useful for testing. This should only ever be called
 	// prior to running Run.
-	WithQueueFactory(fn func(ctx context.Context, baseBackoff time.Duration, maxBackoff time.Duration) WorkQueue) Controller
+	WithQueueFactory(fn func(ctx context.Context, baseBackoff time.Duration, maxBackoff time.Duration) queue.WorkQueue[Request]) Controller
 	// AddTrigger allows for triggering a reconciliation request when a
 	// triggering function returns, when the passed in context is canceled
 	// the trigger must return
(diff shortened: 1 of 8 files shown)


agent/consul/controller/controller.go
```diff
@@ -53,7 +55,7 @@ type Controller interface {
 	// WithQueueFactory allows a Controller to replace its underlying work queue
 	// implementation. This is most useful for testing. This should only ever be called
 	// prior to running Run.
-	WithQueueFactory(fn func(ctx context.Context, baseBackoff time.Duration, maxBackoff time.Duration) WorkQueue) Controller
+	WithQueueFactory(fn func(ctx context.Context, baseBackoff time.Duration, maxBackoff time.Duration) queue.WorkQueue[Request]) Controller
```
andrewstucki, line 58: Did we want to actually parameterize the controller interface itself too? I'm envisioning we could probably keep this interface for `ConfigEntry` v. `Resource` and then just have the two concrete implementations, one using this current `Request` and the other doing like `ResourceRequest` or something?

andrewstucki (commented):
Just a quick question as to how you want to re-use this.

If you just want the queueing stuff that's totally cool, but if this is for the `Resource` adoption, wondering if it makes sense to also make the `Controller` interface itself generic and swap out the `Request` and/or subscription parameters passed in the functions that they implement? That way there'd be consistency in the setup code and internal calling conventions whether you were using a Resource-based controller or one for ConfigEntries.

If you think there's going to be big enough drift between the interfaces, or if you wanted to keep the Resource stuff more concrete without adhering to this interface, that's fine too though.

agent/consul/controller/queue/queue.go
```diff
@@ -119,7 +123,7 @@ func (q *queue) Get() (item Request, shutdown bool) {
 
 // Add puts the given Request in the queue. If the Request is already in
 // the queue or the queue is stopping, then this is a no-op.
-func (q *queue) Add(item Request) {
+func (q *queue[T]) Add(item T) {
```
boxofrad, line 126: Something I've been reflecting on is that `queue[Request].Add` has a natural de-dupliation behavior in OSS because two `Request`s with the same `Kind` and `Name` will hash to the same value. 

In contrast, the new controller request type contains a pointer to `*pbresource.ID` so requests for the same resource will hash to different values.

I'm wondering if we should add a `HashKey` method to `T` to fix this?

boxofrad: Fixed in https://github.com/hashicorp/consul/pull/17168

agent/consul/controller/queue_test.go
```diff
@@ -6,28 +6,30 @@ package controller
...
 	"time"
+
+	"github.com/hashicorp/consul/agent/consul/controller/queue"
 )
 
-var _ WorkQueue = &countingWorkQueue{}
+var _ queue.WorkQueue[string] = &countingWorkQueue[string]{}
 
-type countingWorkQueue struct {
+type countingWorkQueue[T queue.ItemType] struct {
```
andrewstucki, line 15: I'm wondering if it makes more sense to move this into the new subpackage and then just expose it publicly for use in tests... I think if that ended up happening though it can probably wait for now and this is fine for the time being.

andrewstucki (approved):
Overall looks good to me, thanks Dan!

PR #16983, opened 2023-04-12
Enforce Owner rules in `Write` endpoint

analogue:
### Description

Rules:
- Owners can only be set on creation
- If a resource has an owner, their tenancies must be the same 

- Existing and new tests are green

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/grpc-external/services/resource/write.go
```diff
@@ -142,10 +151,11 @@ func (s *Server) Write(ctx context.Context, req *pbresource.WriteRequest) (*pbre
 		return nil, status.Error(codes.Aborted, err.Error())
 	case errors.Is(err, storage.ErrWrongUid):
 		return nil, status.Error(codes.FailedPrecondition, err.Error())
+	case isGRPCStatusError(err):
+		return nil, err
```
analogue, line 155: Not a big fan of this for passing through already constructed GRPC status errors. Time to introduce private resource service level errors and do conversion in this switch maybe?

boxofrad (approved):
Great job as always :shipit:

PR #16994, opened 2023-04-13
APIGW Normalize Status Conditions

jm96441n:
### Description

We want to normalize status conditions for gateways and routes to more closely match the [k8s spec](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.GatewayConditionReason). This PR also includes validations for combinations of status/type/reason. We deviate from the spec in a few places on routes and this is called out in the code.

Run the tests

### Links

[k8s spec](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.GatewayConditionReason)
[k8s shared types](https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/shared_types.go)

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/consul/fsm_data_store.go
```diff
@@ -56,15 +56,11 @@ func (f *FSMDataStore) Update(entry structs.ConfigEntry) error {
 // UpdateStatus takes a config entry, an error, and updates the status field as needed in the FSM state
 func (f *FSMDataStore) UpdateStatus(entry structs.ControlledConfigEntry, err error) error {
```
jm96441n, line 57: not entirely sure what needs to go here as this function appears to be unused

agent/consul/gateways/controller_gateways.go
```diff
@@ -840,151 +829,132 @@ func newGatewayMeta(gateway *structs.APIGatewayConfigEntry, bound structs.Config
 	}).initialize()
 }
 
-// gatewayConditionGenerator is a simple struct used for isolating
-// the status conditions that we generate for our components
-type gatewayConditionGenerator struct {
```
jm96441n, line 845: this struct was no longer needed because we encapsulated creating the conditions in the structs package along with validations

agent/structs/config_entry_status.go
```diff
@@ -195,3 +210,466 @@ func (u *StatusUpdater) UpdateEntry() (ControlledConfigEntry, bool) {
...
+	// to find the gateway
+	RouteReasonGatewayNotFound RouteConditionReason = "GatewayNotFound"
+)
+
+// NewRouteCondition is a helper to build allowable Conditions for a Route config entry
+func NewRouteCondition(name RouteConditionType, status ConditionStatus, reason RouteConditionReason, message string, ref ResourceReference) Condition {
+	if err := checkRouteConditionReason(name, status, reason); err != nil {
+		// note we panic here because an invalid combination is a programmer error
+		// this  should never actually be hit
+		panic(err)
```
jm96441n, line 588: we panic here and in the gateway constructor because an error here is 100% programmer error

agent/consul/fsm_data_store.go
jm96441n: this can just go!

agent/consul/gateways/controller_gateways.go
```diff
@@ -840,151 +828,132 @@ func newGatewayMeta(gateway *structs.APIGatewayConfigEntry, bound structs.Config
...
-		Status:             "True",
-		Reason:             "Accepted",
-		Message:            "gateway is valid",
-		LastTransitionTime: g.now,
-	}
+// gatewayListenerNoConflicts marks an APIGateway listener as having no conflicts within its
+// bound routes
+func gatewayListenerNoConflicts(ref structs.ResourceReference) structs.Condition {
+	return structs.NewGatewayCondition(
+		structs.GatewayConditionListenersConfigured,
```
jm96441n, line 871: I changed out our custom "Conflicted" type for "ListenersConfigured" for two reasons:
1. to be closer to the spec
2. It was the only condition where `ConditionStatusFalse` meant there were no issues and that just seemed a bit counter intuitive

agent/consul/gateways/controller_gateways.go
```diff
@@ -840,151 +828,132 @@ func newGatewayMeta(gateway *structs.APIGatewayConfigEntry, bound structs.Config
...
 // to invalid due to missing certificates that it references.
-func (g *gatewayConditionGenerator) invalidCertificates() structs.Condition {
-	return structs.Condition{
-		Type:               "Accepted",
-		Status:             "False",
-		Reason:             "InvalidCertificates",
-		Message:            "gateway references invalid certificates",
-		LastTransitionTime: g.now,
-	}
+func invalidCertificates() structs.Condition {
```
jm96441n, line 857: now this and `invalidCertificate` fail for the same Condition/Reason but have different messages

agent/structs/config_entry_status.go
```diff
@@ -195,3 +210,467 @@ func (u *StatusUpdater) UpdateEntry() (ControlledConfigEntry, bool) {
...
+	// This reason is used with the "ListenersConfigured" condition when
+	// the Listener conflicts with hostnames in other Listeners. For
+	// example, this reason would be used when multiple Listeners on
+	// the same port use `example.com` in the hostname field.
+	GatewayListenerReasonHostnameConflict GatewayConditionReason = "ListenerHostnameConflict"
+
+	// This reason is used with the "ListenersConfigured" condition when
+	// multiple Listeners are specified with the same Listener port
+	// number, but have conflicting protocol specifications.
+	GatewayListenerReasonProtocolConflict GatewayConditionReason = "ListenerProtocolConflict"
```
andrewstucki, line 296: So, for a number of these things, including this, we don't actually ever use these `GatewayConditionReason` values, namely because this is something that we can validate synchronously on an attempted write and just reject with an error. Do you mind removing all of the unused `Reason`/`Type` values that we don't actually use? Otherwise I think we're just accumulating enums that aren't actually valid for our implementation.

agent/structs/config_entry_status.go
```diff
@@ -195,3 +210,467 @@ func (u *StatusUpdater) UpdateEntry() (ControlledConfigEntry, bool) {
...
+	// used in a number of instances, including:
+	//
+	// * The port is already in use.
+	// * The port is not supported by the implementation.
+	GatewayListenerReasonPortUnavailable GatewayConditionReason = "ListenerPortUnavailable"
+
+	// This reason is used with the "ListenersConfigured" condition when the
+	// Listener could not be attached to be Gateway because its
+	// protocol type is not supported.
+	GatewayListenerReasonUnsupportedProtocol GatewayConditionReason = "ListenerUnsupportedProtocol"
```
andrewstucki, line 309: As is this.

andrewstucki (changes requested):
So, I like the better strong typing, but I think we need to clean up and have only a subset of these specified. We don't need to dupe the entire upstream spec because a number of these conditions/reasons are things that we shouldn't ever allow to be written in the first place.

agent/consul/gateways/controller_gateways.go
```diff
@@ -840,151 +828,132 @@ func newGatewayMeta(gateway *structs.APIGatewayConfigEntry, bound structs.Config
...
+	)
+}
+
+// gatewayListenerConflicts marks an APIGateway listener as having bound routes that conflict with each other
+// and make the listener, therefore invalid
+func gatewayListenerConflicts(ref structs.ResourceReference) structs.Condition {
+	return structs.NewGatewayCondition(
+		structs.GatewayConditionListenersConfigured,
+		structs.ConditionStatusFalse,
+		structs.GatewayListenerReasonProtocolConflict,
```
andrewstucki, line 885: Also, this appears to change things to say `ProtocolConflict` which actually isn't accurate here -- there's no "protocol conflict" that we can ever have as far as a listener is concerned. In particular this helper is used due to the fact that we don't support TCP load-balancing, so if you bind multiple routes to a TCP-based listener we say "this thing is invalid". So I think we'll want to stick with the custom reason here.

agent/consul/gateways/controller_gateways.go
jm96441n: I missed that one when I looked at the spec! I'll add that in and move it back to the condition being conflicted and use a custom `RouteConflict` as the reason

agent/structs/config_entry_status.go
```diff
@@ -141,6 +155,7 @@ func (c *Condition) IsCondition(other *Condition) bool {
 
 func (c *Condition) IsSame(other *Condition) bool {
 	return c.IsCondition(other) &&
+		c.Type == other.Type &&
```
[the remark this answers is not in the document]
jm96441n: oh good catch

agent/consul/gateways/controller_gateways.go
jm96441n: so @andrewstucki this is now the only one (along with `invalidCertificate` ) to have a different type/reason (so `ResolvedRefs` type and `InvalidCertifcateRef` for the reason vs `Accepted` type and `InvalidCertificate` for the reason). Do you have an opinion on whether this makes sense or should be reverted?

andrewstucki: So, I'm thinking that it makes sense to move `invalidCertificate` to a `ResolvedRefs` status along with keeping it scoped to an individual listener. I'd be _ok_ potentially dropping this entire status, but maybe in a separate PR after discussing a bit more. Specifically here's the issue:

1. I believe we want `Accepted` to be a rolled up condition as to whether a Gateway/Route is invalid or not
2. We actually have a check that a Gateway has `Accepted` == `True` in order for it to actually bind any routes (meaning changing this will change this behavior):
https://github.com/hashicorp/consul/blob/a3bf847208797462fc44c333dadc95b34163d7d2/agent/consul/gateways/controller_gateways.go#L673-L682

What this currently means is that if any of our certificate refs are invalid across any listener, we actually say that the Gateway can't bind a route, that's probably wrong and should probably change and be scoped to a given listener (which the method that this check is contained in actually has access to).

If we want to fix it, rather than mark the entire Gateway as not `Accepted`, we could potentially mark each Listener individually as `Accepted` and scope the above binding check to a listener. That said, that would potentially defeat the purpose of considering `Accepted` as a rolled-up status since I can't see how else an _entire_ Gateway would ever be `Accepted == False`.

So, the TLDR; on my initial thoughts are:

**keep this `Accepted`, switch `invalidCertificate` to `ResolvedRefs`, at least for this PR**

but I'm definitely open to discussion.

@mikemorris any thoughts?

jm96441n: that makes sense to me, if anything we can make changes in a future PR once we get a better feel for how this should look

agent/structs/config_entry_status.go
```diff
@@ -195,3 +209,308 @@ func (u *StatusUpdater) UpdateEntry() (ControlledConfigEntry, bool) {
 	u.entry.SetStatus(u.status)
 	return u.entry, true
 }
+
+// GatewayConditionType is a type of condition associated with a
+// Gateway. This type should be used with the GatewayStatus.Conditions
+// field.
+type GatewayConditionType string
```
jm96441n, line 216: so I'm also thinking that either these are in the wrong package (should be in `agent/api/config_entry_status`) or we also need to export these from the API package so we can use them from other places programatically

api/go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/hashicorp/serf v0.10.1
 	github.com/mitchellh/mapstructure v1.4.1
 	github.com/stretchr/testify v1.7.0
+	golang.org/x/exp v0.0.0-20230321023759-10a507213a29
```
jm96441n, line 17: so running `go mod tidy` here to get exp slices package caused a cascading need of changes to other `go.mod` files that import this module which is why this change seems larger than it is

PR #16997, opened 2023-04-13
add IP rate limiting config update

dhiaayachi:
### Description
This adds the needed bits to build IP based rate limiting as an enterprise feature.


JadhavPoonam (approved):
Looks fine to me!

agent/consul/rate/handler.go
```diff
@@ -8,6 +8,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"golang.org/x/time/rate"
```
loshz, line 11: Nit: should be grouped with non stdlib imports.

agent/consul/rate/handler.go
```diff
@@ -189,20 +195,29 @@ type LeaderStatusProvider interface {
 	IsLeader() bool
 }
 
+func isInfRate(cfg multilimiter.LimiterConfig) bool {
+	if cfg.Rate == rate.Inf {
+		return true
+	}
+	return false
```
loshz, line 202: Nit:
```suggestion
	return cfg.Rate == rate.Inf
```

Issue #15338, opened 2022-11-11
Inbound Peering Control Plane Traffic includes Follower Nodes in Cluster Endpoints


DanStough:
#### Overview of the Issue

Peering control plane traffic is established between the respective Leader Servers in peered clusters. When `PeerThroughMeshGateways` is enabled in the Mesh config entry, this traffic can be mistakenly routed to Follower servers on the accepting side because that mesh gateway has all the server addresses in the Envoy endpoint configuration. This doesn't cause an error with peering over mesh gateways because peering is designed so that Follower servers will hang up on dialing requests, and the dialing Leader will retry. 

This is less of a bug and more issues with sub-optimal performance and [dead-code that is not performing as designed](https://github.com/hashicorp/consul/blob/bf0f61a87884d444d472b981099163856658d3ea/agent/xds/endpoints.go#L299-L303). A successful PR will refactor the code to allow only the leader node in the mesh gateway inbound cluster.

#### Reproduction Steps

1. Run ./run.sh to set up a peering between two clusters through mesh gateways ([reference repo](https://github.com/DanStough/consul-scenarios/tree/main/scenario-peering-mesh-gw)). You will also need to manually run the commands after the `exit 0` in the script to complete the peering connection. 
2. Add two servers:
   1. `consul agent -log-level trace -config-file config_dc1-server2.hcl 1>./logs/dc1-server2.log &.`
   1. `consul agent -log-level trace -config-file config_dc1-server3.hcl 1>./logs/dc1-server3.log &.`
1. Observer the mesh gateway has all server gRPC addresses in its cluster endpoints.
  1. `http://localhost:19004/clusters?format=json`

DataDavD:
hey @DanStough I'm gonna try and work on this issue. 

Note, if anyone comes along wanting to work on this/submit PR/etc. feel free. I'm a bit slow right now after moving across the country.

DanStough:
That would be awesome, @DataDavD!  This is not a priority right now, so it's all yours. 

I realized I forgot to include the setup script - I am adding a link back in the original issue.

DataDavD:
@DanStough thank you! Just forewarning you I'll probably be back with some questions as I'm not very familiar with consul....trying to improve my Go while helping the community and getting familiar with consul/HashiCorp software 👍 

thanks again!

DataDavD:
[...]

PR #17000, opened 2023-04-14
Add a check for local leader in cluter peering

sladyn98:
### Description

This change adds a check for the leader referring to this issue https://github.com/hashicorp/consul/issues/15338

Run ./run.sh to set up a peering between two clusters through mesh gateways ([reference repo](https://github.com/DanStough/consul-scenarios/tree/main/scenario-peering-mesh-gw)). You will also need to manually run the commands after the exit 0 in the script to complete the peering connection.
Add two servers:
consul agent -log-level trace -config-file config_dc1-server2.hcl 1>./logs/dc1-server2.log &.
consul agent -log-level trace -config-file config_dc1-server3.hcl 1>./logs/dc1-server3.log &.
Observer the mesh gateway has all server gRPC addresses in its cluster endpoints.
http://localhost:19004/clusters?format=json

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern

### DIFF
--- a/agent/xds/endpoints.go
+++ b/agent/xds/endpoints.go
@@ -304,9 +304,8 @@ func (s *ResourceGenerator) endpointsFromSnapshotMeshGateway(cfgSnap *proxycfg.C
 
 		servers, _ := cfgSnap.MeshGateway.WatchedLocalServers.Get(structs.ConsulServiceName)
 		for _, srv := range servers {
-			if isReplica := srv.Service.Meta["read_replica"]; isReplica == "true" {
-				// Peering control-plane traffic can only ever be handled by the local leader.
-				// We avoid routing to read replicas since they will never be Raft voters.
+			isLeader := srv.Service.Meta["leader"] == "true"
+			if !isLeader {
 				continue
 			}
 


agent/xds/endpoints.go
```diff
@@ -304,9 +304,8 @@ func (s *ResourceGenerator) endpointsFromSnapshotMeshGateway(cfgSnap *proxycfg.C
 
 		servers, _ := cfgSnap.MeshGateway.WatchedLocalServers.Get(structs.ConsulServiceName)
 		for _, srv := range servers {
-			if isReplica := srv.Service.Meta["read_replica"]; isReplica == "true" {
-				// Peering control-plane traffic can only ever be handled by the local leader.
-				// We avoid routing to read replicas since they will never be Raft voters.
+			isLeader := srv.Service.Meta["leader"] == "true"
```
sladyn98, line 307: I am not sure if this exists in the metadata

sladyn98:
@DanStough

agent/xds/endpoints.go
DanStough: Unfortunately, it does not, and adding it would be a chore. The current metadata is constant at startup time so it is propagated with serf, but the raft leader is not known until the leader election completes.
```json
"ServiceMeta": {
  "grpc_port": "8522",
  "grpc_tls_port": "8523",
  "non_voter": "true",
  "raft_version": "3",
  "read_replica": "true",
  "serf_protocol_current": "2",
  "serf_protocol_max": "5",
  "serf_protocol_min": "1",
  "version": "1.16.0"
},
```

DanStough:
Firstly, thank you, @sladyn98, for the PR 🥳 !

The original issue is still valid, but thinking through a solution with some other Consul folks, I vastly underestimated the amount of work needed. It's my fault for tagging the issue with the `good first issue` label, and I apologize if you sunk time in this under that assumption.

For a solution, we need to address two concerns:
1. Which IP addresses represent a leader? As mentioned in the comment above, this isn't part of the service instance metadata.
1. Triggering an xDS update when the leader changes position. This requires a new watch in the `proxyconfg` package.

That said, I'll say it's your call if you'd like to keep working on this issue, but IMHO it isn't worth the effort for the small performance gain. If you're out, I'll close the original issue with a comment to the effect "won't fix right now".

sladyn98:
@DanStough Thanks for the heads up, yeah you can close the issue and ill pick up a new one

PR #17002, opened 2023-04-14
API Gateway XDS Primitives, endpoints and clusters

sarahalsmiller:
### Description
Work to move APIGateway away from using prexisting ingress gateway functionality. 

Existing gateway behavior should be unchanged with this change, so existing CI tests continuing to pass should be sufficient. 

### Links

Relevant RFC: https://docs.google.com/document/d/15kzB5wqhzzPOMWCD8SphpcYUAigUmJvV6R6y2MZUnJU/edit?usp=sharing

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


agent/xds/clusters.go
```diff
@@ -816,6 +810,54 @@ func (s *ResourceGenerator) clustersFromSnapshotIngressGateway(cfgSnap *proxycfg
...
+
+				uid := proxycfg.NewUpstreamID(&u)
+
+				// If we've already created a cluster for this upstream, skip it. Multiple listeners may
+				// reference the same upstream, so we don't need to create duplicate clusters in that case.
+				if createdClusters[uid] {
+					continue
+				}
+
+				chain, ok := cfgSnap.IngressGateway.DiscoveryChain[uid]
```
andrewstucki, line 833: Hey Sarah, so I'll need to review some of the RFC to see the full approach suggested, but part of our desire is to get fully rid of the translation that happens from the `APIGateway` portion of a proxycfg snapshot into the `IngressGateway` portion, which means that ideally none of our code would use any of the `IngressGateway` fields. That would allow us to drop all of the transformation logic found here:

https://github.com/hashicorp/consul/blob/0dd4ea20337478174fb56538b42f7ca1cd99e10c/agent/proxycfg/snapshot.go#L749 and in the corresponding discovery chain "synthesizing" code that exists in the `gateway*.go` files in [this](https://github.com/hashicorp/consul/tree/main/agent/consul/discoverychain) subpackage.

But it does mean that we'll need to come up with the direct HTTPRoute --> xDS route synthesis code (which may or may not also leak to listeners, and seemingly here into clusters as well).

Maybe we can set up some time next week to chat about this synchronously to get on the same page?

sarahalsmiller: Hi Andrew, I would be happy to hop on a call with you sometime this week, but I just accidentally missed that IngressGateway field in my initial pass. Part of the trouble I'm having with this is how much of the preexisting Xds generation/helper codes seem to rely heavily on the preexisting IngressGateway struct, so I do think it'd be helpful to talk through how our architecture is set up and what we want this to look like in the end.

PR #17004, opened 2023-04-14
Tenancy wildcard validaton for `Write`, `Read`, and `Delete` endpoints

analogue:
### Description

Make sure tenancy fields are not wildcards on write, read, or delete. See questions inline.

Existing and new tests and green.

### PR Checklist

* [x] updated test coverage
* [x] not a security concern


agent/grpc-external/services/resource/write.go
```diff
@@ -226,18 +226,52 @@ func validateWriteRequest(req *pbresource.WriteRequest) error {
...
+		return status.Errorf(codes.InvalidArgument, "%s.%s is required", errorPrefix, field)
+	}
+
+	switch {
+	case id.Tenancy.Namespace == storage.Wildcard:
+		field = "tenancy.namespace"
+	case id.Tenancy.Partition == storage.Wildcard:
+		field = "tenancy.partition"
+	case id.Tenancy.PeerName == storage.Wildcard:
+		field = "tenancy.peername"
```
analogue, line 270: I'm guessing `Namespace` and `Partition` can't be empty and `default` should be understood as the default, but I don't think I've seen that called out anywhere. Not too sure about `PeerName`.

analogue: To close the loop here, we'll be handling `Namespace` and `Partition` defaults in a separate PR.

PR #17055, opened 2023-04-19
[API Gateway] Fix use of virtual resolvers in HTTPRoutes

andrewstucki:
### Description

I realized that we were actually improperly munging http route references in synthesized discovery chains for API gateways. What this meant was that when using an service resolver to create a virtual service referenced by an HTTPRoute we wind up writing in the wrong target cluster for a routing rule. This results in the gateway returning "no healthy upstreams".

This change takes into account when the target is just a virtual service created by a service-resolver.

Unit tested and manually verified with the following configuration entries and a mesh-federated Consul setup:

1. TCP Routes

```hcl
Kind = "tcp-route"
Name = "api-gateway-route"
Services = [
  {
    Name = "tcp-dc-2"
  }
]

Parents = [
  {
    Name = "api"
  }
]
```

```hcl
Kind = "api-gateway"
Name = "api"
Listeners = [
  {
    Name     = "listener-one"
    Port     = ...
    Protocol = "tcp"
  }
]
```

```hcl
Kind = "service-resolver"
Name = "tcp-dc-2"
Redirect {
  Service    = "tcp-2"
  Datacenter = "dc2"
}
```

2. HTTPRoutes

```hcl
Kind = "http-route"
Name = "api-gateway-route"
Services = [
  {
    Name = "http-dc-2"
  }
]

Parents = [
  {
    Name = "api"
  }
]
```

```hcl
Kind = "api-gateway"
Name = "api"
Listeners = [
  {
    Name     = "listener-one"
    Port     = ...
    Protocol = "http"
  }
]
```

```hcl
Kind = "service-resolver"
Name = "http-dc-2"
Redirect {
  Service    = "http-2"
  Datacenter = "dc2"
}
```

### PR Checklist

* [x] updated test coverage
* [x] not a security concern

### DIFF
--- a/.changelog/17055.txt
+++ b/.changelog/17055.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+gateways: Fix an bug where targeting a virtual service defined by a service-resolver was broken for HTTPRoutes.
+```
--- a/agent/consul/discoverychain/gateway.go
+++ b/agent/consul/discoverychain/gateway.go
@@ -232,6 +232,13 @@ func targetForResolverNode(nodeName string, chains []*structs.CompiledDiscoveryC
 	splitterName := splitterPrefix + strings.TrimPrefix(nodeName, resolverPrefix)
 
 	for _, c := range chains {
+		targetChainPrefix := resolverPrefix + c.ServiceName + "."
+		if strings.HasPrefix(nodeName, targetChainPrefix) && len(c.Nodes) == 1 {
+			// we have a virtual resolver that just maps to another resolver, return
+			// the given node name
+			return c.StartNode
+		}
+
 		for name, node := range c.Nodes {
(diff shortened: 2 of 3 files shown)


.changelog/17055.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+gateways: Fix an bug where targeting a virtual service defined by a service-resolver was broken for HTTPRoutes.
```
nathancoleman, line 2:
```suggestion
gateways: Fix a bug where targeting a virtual service defined by a service-resolver was broken for HTTPRoutes.
```

nathancoleman (approved):

andrewstucki:
merging this as-is for now so it can get backported for the 1.15.3 release, but will circle back to add additional integration tests later

PR #17066, opened 2023-04-20
Remove artificial ACLTokenMaxTTL limit for configuring acl token expiry

johnlanda:
### Description

The default expiry for an ACL token is never. Currently, when supplying an acl token ttl value the max allowed via the `-expires-ttl` flag is 24 hours. 

This change sets the default max ttl value for ACL tokens to the max duration value. 

* `command/acl/token/create/token_create_test.go`
  * `expires-ttl_short`
  * `expires-ttl_long`

### Links

[-expires-ttl flag docs](https://developer.hashicorp.com/consul/commands/acl/token/create#expires-ttl)

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern


test/integration/consul-container/test/consul_envoy_version/consul_envoy_version.go
```diff
@@ -34,7 +34,7 @@ func main() {
 	sort.Sort(sort.Reverse(sort.StringSlice(cev.EnvoyVersions)))
 
 	ceVersions := consulEnvoyVersions{
-		ConsulVersion: string(cVersion),
+		ConsulVersion: cVersion,
```
johnlanda, line 37: This was causing issues with the linter.

agent/consul/config.go
```diff
@@ -522,11 +522,13 @@ func DefaultConfig() *Config {
 		TombstoneTTLGranularity:              30 * time.Second,
 		SessionTTLMin:                        10 * time.Second,
 		ACLTokenMinExpirationTTL:             1 * time.Minute,
-		ACLTokenMaxExpirationTTL:             24 * time.Hour,
+		// Duration is stored as an int64. Setting the default max
+		// to the max possible duration (approx 290 years).
+		ACLTokenMaxExpirationTTL: 1<<63 - 1,
```
johnlanda, line 527: Alternatively, we could rework things so that this value is completely removed but I figured it makes sense to allow people to still configure their own max if they want. So just setting the default to the max `time.Duration` value.

.changelog/17066.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+command: Allows creating ACL Token TTL with greater than 24 hours with the -expires-ttl flag.
```
cthain, line 2: nit: per the style guide in the eng doc we should be using the imperitive for changelog entries for consistency.
```suggestion
command: Allow creating ACL Token TTL with greater than 24 hours with the -expires-ttl flag.
```

test/integration/consul-container/test/consul_envoy_version/consul_envoy_version.go
cthain: It looks like this change has [already been applied to `main`](https://github.com/hashicorp/consul/blob/main/test/integration/consul-container/test/consul_envoy_version/consul_envoy_version.go#L37). You may need to merge/rebase.

cthain (approved):
This is awesome @johnlanda. Congrats on your first PR! I just had one non-blocking nitpick.

agent/consul/config.go
jkirschner-hashicorp, line 527: @johnlanda : Do we need to update any docs associated with this parameter?

Strangely, it seems like the relevant docs I could find never mention the current 24 hour limit ... so maybe there is nothing to update?
- https://developer.hashicorp.com/consul/commands/acl/token/create#expires-ttl
- https://developer.hashicorp.com/consul/commands/acl/auth-method/create#max-token-ttl

johnlanda: I could not find any mention of the limit either. I will take another look though to be sure.

johnlanda: I do not believe there is anything to update. It looks like that was an undocumented limit.

PR #17076, opened 2023-04-20
matrix integration testing POC

eikenb:
A proof of concept implementation of a testing matrix which runs a simple test walk through of the tutorial covering using Vault as the CA for Consul service mesh. It runs the matrix test using the latest three releases of Consul and Vault against test.


test/integration/matrix/download.go
```diff
@@ -0,0 +1,122 @@
...
+	type releasesJSON struct {
+		Builds []build
+	}
+	var rel releasesJSON
+	if err := json.Unmarshal(jsonBody, &rel); err != nil {
+		panic(err)
+	}
+
+	for _, b := range rel.Builds {
+		if b.Arch == "amd64" && b.OS == "linux" {
```
nfi-hashicorp, line 70: Do you think making this work for other platforms would be as simple as swapping this with something like this?

```
		if b.Arch == runtime.GOARCH && b.OS == runtime.GOOS {
```

(and similar in the runner code)

eikenb: Sure.. something like that should work. Though multi-platform testing is not in scope at this time.

PR #17080, opened 2023-04-20
ci: fix test splits that have less test packages than runner count from hanging

jmurret:
### Description
In 1.13, there are only two compatibility test packages.  So when a generate matrix job has a TOTAL_RUNNERS of 5 but there are only 2 packages, the process hangs.  This PR changes the runner count to 2 in this case.  

Another issue that occurs that is fixed in this PR is that when you specify TOTAL_RUNNERS = 4, you will actually split across 5 runners.  (The same occurs with any number you set TOTAL_RUNNERS to.  You will actually get 1 greater.)

### DIFF
--- a/.github/workflows/test-integrations.yml
+++ b/.github/workflows/test-integrations.yml
@@ -4,7 +4,6 @@
 name: test-integrations
 
 on:
-  push:
   pull_request:
     branches-ignore:
       - stable-website
--- a/.github/workflows/test-integrations.yml
+++ b/.github/workflows/test-integrations.yml
@@ -168,6 +168,7 @@ jobs:
           fi
           # fix issue where test splitting calculation generates 1 more split than TOTAL_RUNNERS.
           NUM_RUNNERS=$(($NUM_RUNNERS))-1
+          echo "NUM_RUNNERS: $NUM_RUNNERS"
           {
             echo -n "envoy-matrix="
             find ./test/integration/connect/envoy -maxdepth 1 -type d -print0 \
@@ -284,6 +285,7 @@ jobs:
           fi
           # fix issue where test splitting calculation generates 1 more split than TOTAL_RUNNERS.
           NUM_RUNNERS=$(($NUM_RUNNERS))-1
+          echo "NUM_RUNNERS: $NUM_RUNNERS"
           {
             echo -n "compatibility-matrix="
             find ./test -maxdepth 2 -type d -print0 | xargs -0 -n 1 \
@@ -391,6 +393,7 @@ jobs:
           fi
           # fix issue where test splitting calculation generates 1 more split than TOTAL_RUNNERS.
           NUM_RUNNERS=$(($NUM_RUNNERS))-1
+          echo "NUM_RUNNERS: $NUM_RUNNERS"
           {
             echo -n "upgrade-matrix="
             go test ./...  -list=. -json | jq -r '.Output | select (. !=null) |  select(. | startswith("Test")) | gsub("[\\n\\t]"; "")' \
--- a/.github/workflows/test-integrations.yml
+++ b/.github/workflows/test-integrations.yml
@@ -156,7 +156,7 @@ jobs:
           # multiplied by 8 based on these values:
           # envoy-version: ["1.22.11", "1.23.8", "1.24.6", "1.25.4"]
           # xds-target: ["server", "client"]
-          TOTAL_RUNNERS: 700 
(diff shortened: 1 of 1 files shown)


.github/workflows/test-integrations.yml
```diff
@@ -4,6 +4,7 @@
 name: test-integrations
 
 on:
+  push:
```
loshz, line 7: Was this added for debugging?

.github/workflows/test-integrations.yml
```diff
@@ -155,17 +156,25 @@ jobs:
...
         run: |
+          NUM_RUNNERS=$TOTAL_RUNNERS
+          NUM_DIRS=$(find ./test/integration/connect/envoy -maxdepth 1 -type d | wc -l)
+          if [ $NUM_DIRS -lt $NUM_RUNNERS ]
+          then
+            echo "TOTAL_RUNNERS is larger than the number of tests/packages to split."
+            NUM_RUNNERS=$NUM_DIRS
+          fi
+          # fix issue where test splitting calculation generates 1 more split than TOTAL_RUNNERS.
+          NUM_RUNNERS=$(($NUM_RUNNERS-1))
```
loshz, line 170:
```suggestion
          if [ "$NUM_DIRS" -lt "$NUM_RUNNERS" ]; then
            echo "TOTAL_RUNNERS is larger than the number of tests/packages to split."
            NUM_RUNNERS=$NUM_DIRS
          fi
          # fix issue where test splitting calculation generates 1 more split than TOTAL_RUNNERS.
          NUM_RUNNERS=$((NUM_RUNNERS-1))
```

.github/workflows/test-integrations.yml
```diff
@@ -261,18 +270,26 @@ jobs:
...
           cd ./test/integration/consul-container
+          NUM_RUNNERS=$TOTAL_RUNNERS
+          NUM_DIRS=$(find ./test -maxdepth 2 -type d | wc -l)
+          if [ $NUM_DIRS -lt $NUM_RUNNERS ]
+          then
+            echo "TOTAL_RUNNERS is larger than the number of tests/packages to split."
+            NUM_RUNNERS=$NUM_DIRS
+          fi
+          # fix issue where test splitting calculation generates 1 more split than TOTAL_RUNNERS.
+          NUM_RUNNERS=$(($NUM_RUNNERS-1))
```
loshz, line 285:
```suggestion
          if [ "$NUM_DIRS" -lt "$NUM_RUNNERS" ]; then
            echo "TOTAL_RUNNERS is larger than the number of tests/packages to split."
            NUM_RUNNERS=$NUM_DIRS
          fi
          # fix issue where test splitting calculation generates 1 more split than TOTAL_RUNNERS.
          NUM_RUNNERS=$((NUM_RUNNERS-1))
```

.github/workflows/test-integrations.yml
```diff
@@ -359,17 +376,25 @@ jobs:
...
           cd ./test/integration/consul-container/test/upgrade
+          NUM_RUNNERS=$TOTAL_RUNNERS
+          NUM_DIRS=$(go test ./...  -list=. -json | jq -r '.Output | select (. !=null) |  select(. | startswith("Test")) | gsub("[\\n\\t]"; "")' | wc -l)
+          if [ $NUM_DIRS -lt $NUM_RUNNERS ]
+          then
+            echo "TOTAL_RUNNERS is larger than the number of tests/packages to split."
+            NUM_RUNNERS=$NUM_DIRS
+          fi
+          # fix issue where test splitting calculation generates 1 more split than TOTAL_RUNNERS.
+          NUM_RUNNERS=$(($NUM_RUNNERS-1))
```
loshz, line 391:
```suggestion
          if [ "$NUM_DIRS" -lt "$NUM_RUNNERS" ]; then
            echo "TOTAL_RUNNERS is larger than the number of tests/packages to split."
            NUM_RUNNERS=$NUM_DIRS
          fi
          # fix issue where test splitting calculation generates 1 more split than TOTAL_RUNNERS.
          NUM_RUNNERS=$((NUM_RUNNERS-1))
```

loshz (commented):
Just the one comment about the `push` trigger. The rest are non-blocking syntactic bash NITs.

.github/workflows/test-integrations.yml
jmurret: Fixed.  ty!

loshz (approved):

PR #17086, opened 2023-04-21
Add ACLs Enabled field to consul agent startup status message

johnlanda:
### Description

When starting the consul agent, we output a status message which shows various parts of the configuration including `Gossip Encryption` and `Auto-Encrypt-TLS` configurations. 

This adds another field to the status showing whether or not ACLs are enabled.

```
make dev
./bin/consul agent -dev
```

### PR Checklist

* [ ] updated test coverage
* [x] not a security concern

### DIFF
--- a/command/agent/agent.go
+++ b/command/agent/agent.go
@@ -230,6 +230,7 @@ func (c *cmd) run(args []string) int {
 		config.SerfPortLAN, config.SerfPortWAN))
 	ui.Info(fmt.Sprintf("Gossip Encryption: %t", config.EncryptKey != ""))
 	ui.Info(fmt.Sprintf(" Auto-Encrypt-TLS: %t", config.AutoEncryptTLS || config.AutoEncryptAllowTLS))
+	ui.Info(fmt.Sprintf("     ACLs Enabled: %t", config.ACLsEnabled))
 	ui.Info(fmt.Sprintf("        HTTPS TLS: Verify Incoming: %t, Verify Outgoing: %t, Min Version: %s",
 		config.TLS.HTTPS.VerifyIncoming, config.TLS.HTTPS.VerifyOutgoing, config.TLS.HTTPS.TLSMinVersion))
 	ui.Info(fmt.Sprintf("         gRPC TLS: Verify Incoming: %t, Min Version: %s", config.TLS.GRPC.VerifyIncoming, config.TLS.GRPC.TLSMinVersion))
--- a/command/agent/agent.go
+++ b/command/agent/agent.go
@@ -19,10 +19,9 @@ import (
 	"github.com/hashicorp/go-hclog"
 	mcli "github.com/mitchellh/cli"
 
-	"github.com/hashicorp/consul/agent/hcp"
-
 	"github.com/hashicorp/consul/agent"
 	"github.com/hashicorp/consul/agent/config"
+	"github.com/hashicorp/consul/agent/hcp"
 	hcpbootstrap "github.com/hashicorp/consul/agent/hcp/bootstrap"
 	"github.com/hashicorp/consul/command/cli"
 	"github.com/hashicorp/consul/command/flags"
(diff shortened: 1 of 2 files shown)


command/agent/agent.go
```diff
@@ -215,6 +215,7 @@ func (c *cmd) run(args []string) int {
 		config.SerfPortLAN, config.SerfPortWAN))
 	ui.Info(fmt.Sprintf("Gossip Encryption: %t", config.EncryptKey != ""))
 	ui.Info(fmt.Sprintf(" Auto-Encrypt-TLS: %t", config.AutoEncryptTLS || config.AutoEncryptAllowTLS))
+	ui.Info(fmt.Sprintf("     ACLs Enabled: %t", config.ACLsEnabled))
```
jkirschner-hashicorp, line 218: It's critical to also include what the ACL default policy is, not just that the ACL system is enabled.

If the ACL system is enabled, but the default policy was never set to "deny" (the default is "allow"), then that's effectively the same as having the ACL system disabled.

@cizh : Copying you to help think through how to communicate this best to the user.

For example, I'm worried that if the display shows:
```
ACLs Enabled: true
ACL Default Policy: allow
```

... they might think that things are set up securely after viewing the "ACLs Enabled" line, and maybe glaze over the default policy line that follows.

jkirschner-hashicorp (changes requested):
Thanks for taking this on!

I have one important suggestion for improvement in the comments.

command/agent/agent.go
jkirschner-hashicorp: Maybe it could be something like:

```
Secure ACLs: false (default policy is allow)
Secure ACLs: false (ACL system disabled; default policy is allow)
Secure ACLs: true (ACL system enabled with default policy of deny)
```

johnlanda: Good call. I will mark this as a draft until we know what the message should look like.

johnlanda: I also noticed that the startup output in the consul agent docs is incomplete and can update that as part of this PR as well.

https://developer.hashicorp.com/consul/docs/agent#understanding-the-agent-startup-output

Missing items:
* `Version`
* `Build Date`
* `Node ID`
* `Gossip Encryption`
* `Auto-Encrypt-TLS`
* `ACLs Enabled` <-- whatever fields we add for this one
* `HTTPS TLS`
* `gRPC TLS`
* `Internal RPC TLS`

jkirschner-hashicorp: After discussing, @cizh and I recommend moving forward with:
```
ACLs Enabled: true
ACL Default Policy: allow
```

johnlanda: Updated in [ccc8de7](https://github.com/hashicorp/consul/pull/17086/commits/ccc8de7c7e7d5a18492896f4ac64329ebe42df75)

jkirschner-hashicorp (commented):
LGTM from a PM/UX perspective. Someone from eng needs to give an eng review though (which shouldn't take long). I recommend selecting someone from the team or posting on our Slack channel asking for a review.

jkirschner-hashicorp:
Looks like adding the ZTS reviewer auto-assigned @skpratt :)

command/agent/agent.go
```diff
@@ -15,11 +15,12 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/hashicorp/consul/agent/hcp"
 	"github.com/hashicorp/go-checkpoint"
 	"github.com/hashicorp/go-hclog"
 	mcli "github.com/mitchellh/cli"
 
+	"github.com/hashicorp/consul/agent/hcp"
```
pglass, line 22: nit: I think this should get grouped with the other "hashicorp/consul" imports below.

pglass (approved):

command/agent/agent.go
johnlanda: oops looks like I didn't have my commit hooks set up to fix that. Good catch.

johnlanda: Kind of surprised we don't have a linter in CI for formatting

cthain (approved):
:+1:

PR #17090, opened 2023-04-21
ci: fix runner calculation to exclude the top level directory as part of the calculation

jmurret:
When calculating number of runners, `find` reports back the top level current directory in addition to the child directories.  The actual splitting does not include this top level, so the process hangs.  This fix excludes the top level.

### PR Checklist

* [ ] updated test coverage
* [ ] not a security concern


.github/workflows/test-integrations.yml
```diff
@@ -378,10 +386,14 @@ jobs:
         run: |
           cd ./test/integration/consul-container/test/upgrade
           NUM_RUNNERS=$TOTAL_RUNNERS
-          NUM_DIRS=$(go test ./...  -list=. -json | jq -r '.Output | select (. !=null) |  select(. | startswith("Test")) | gsub("[\\n\\t]"; "")' | wc -l)
+          NUM_DIRS=$(find ./test -maxdepth 2 -type d | wc -l)
```
loshz, line 389: This one looks like a completely different command?

jmurret: ahh, yikes!  ty!

loshz (approved):

PR #17100, opened 2023-04-24
[HCP Observability] New MetricsClient

Achooo:
##### Overall Feature background
This PR is part of the HCP Observability feature to send Server metrics to HCP's new Telemetry Gateway. The idea is to create a new OpenTelemetry go-metrics sink that will be initialized automatically when a Consul server is registered with the HCP management plane.
- That branch will be merged at the end into `main` and back ported

##### This PR: MetricsClient

In #16800 , we refactored the HCP client within the `hcp/client` package.
This PR adds a new `MetricsClient` to this package. 

The key responsibility of the MetricsClient is to send OTLP metrics  to a configured endpoint. To achieve this, it:
- Has a method `ExportMetrics` which sends OTLP metrics to the desired HCP endpoint passed as a parameter.
- Performs HCP Authentication (using a Cloud config object)
- The HTTP client is wrapped with github.com/hashicorp/go-retryablehttp for retry/throttle

##### Next steps
The MetricsClient will be used by the OTEL Exporter to periodically send metrics, coming in a follow up PR.
It will be initialized when the Consul server starts up with the Cloud configuration to provide authentication tokens. 

##### Reference image

minirfc_optionA_metricsclient

### PR Checklist

* [x] updated test coverage
* [ ] not a security concern


agent/hcp/client/metrics_client.go
```diff
@@ -0,0 +1,165 @@
...
+		return nil, fmt.Errorf("failed to init telemetry client: provide valid TelemetryClientCfg")
+	}
+
+	c, err := newHTTPClient(cfg.CloudCfg, cfg.Logger)
+	if err != nil {
+		return nil, fmt.Errorf("failed to init telemetry client: %v", err)
+	}
+
+	headers := map[string]string{
+		"X-HCP-Source-Channel": fmt.Sprintf("consul %s hcp-go-sdk/%s", version.GetHumanVersion(), version.Version),
```
Achooo, line 77: These are some default headers that are needed, let me know if I'm missing any others.  

 I looked at this [client](https://github.com/hashicorp/hcp-sdk-go/blob/main/httpclient/httpclient.go) for inspiration (used in the other hcp client)

Achooo: @nickethier mentioned missing `X-Resource-ID`? Do you have an example of how this is set?

agent/hcp/client/metrics_client.go
```diff
@@ -0,0 +1,165 @@
...
+)
+
+const (
+	// HTTP Client config
+	defaultStreamTimeout = 15 * time.Second
+
+	// Retry config
+	defaultRetryWaitMin = 1 * time.Second
+	defaultRetryWaitMax = 15 * time.Second
+	defaultRetryMax     = 4
```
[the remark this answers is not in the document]
chapmanc: Can we change the config dynamically if we see problems?

agent/hcp/client/metrics_client_test.go
```diff
@@ -0,0 +1,162 @@
...
+func TestExportMetrics(t *testing.T) {
+	for name, test := range map[string]struct {
+		wantErr string
+		status  int
+	}{
+		"success": {
+			status: http.StatusOK,
+		},
+		"failsWithNonRetryableError": {
+			status:  http.StatusBadRequest,
```
chapmanc, line 112: Will partial success always return badrequest?

agent/hcp/client/metrics_client.go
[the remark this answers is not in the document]
Achooo: I'd be happy to create a JIRA issue for this, and have discussion. We could address this in a follow-up PR.

agent/hcp/client/metrics_client.go
Achooo: The go sdk isn't really used, so consider removing that.

loshz: As mentioned on the call earlier, if the telemetry gateway isn't reading and acting on any of these headers then maybe we can just remove them?

agent/hcp/client/metrics_client.go
```diff
@@ -0,0 +1,158 @@
...
+// cloudConfig represents cloud config for TLS abstracted in an interface for easy testing.
+type cloudConfig interface {
+	HCPConfig(opts ...hcpcfg.HCPConfigOption) (hcpcfg.HCPConfig, error)
+}
+
+// otlpClient is an implementation of MetricsClient with a retryable http client for retries and to honor throttle.
+// It also holds default HTTP headers to add to export requests.
+type otlpClient struct {
+	client  *retryablehttp.Client
+	headers map[string]string
```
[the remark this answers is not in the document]
loshz: Go defines headers as `map[string][]string` https://pkg.go.dev/net/http#Header

agent/hcp/client/metrics_client.go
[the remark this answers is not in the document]
loshz: I think having these hard coded for now is ok. I would definitely like to see us make these configurable down the road. Maybe we can add a TODO comment as well as a Jira/backlog item.

agent/hcp/client/metrics_client.go
```diff
@@ -0,0 +1,158 @@
...
+	"google.golang.org/protobuf/proto"
+
+	colmetricpb "go.opentelemetry.io/proto/otlp/collector/metrics/v1"
+	metricpb "go.opentelemetry.io/proto/otlp/metrics/v1"
+
+	"github.com/hashicorp/consul/version"
+	"github.com/hashicorp/go-cleanhttp"
+	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-retryablehttp"
+	hcpcfg "github.com/hashicorp/hcp-sdk-go/config"
```
loshz, line 21: NIT: goimports
```suggestion
	"github.com/hashicorp/go-cleanhttp"
	"github.com/hashicorp/go-hclog"
	"github.com/hashicorp/go-retryablehttp"
	hcpcfg "github.com/hashicorp/hcp-sdk-go/config"
	colmetricpb "go.opentelemetry.io/proto/otlp/collector/metrics/v1"
	metricpb "go.opentelemetry.io/proto/otlp/metrics/v1"
	"golang.org/x/oauth2"
	"google.golang.org/protobuf/proto"

	"github.com/hashicorp/consul/version"

```

agent/hcp/client/metrics_client.go
```diff
@@ -0,0 +1,158 @@
...
+
+// otlpClient is an implementation of MetricsClient with a retryable http client for retries and to honor throttle.
+// It also holds default HTTP headers to add to export requests.
+type otlpClient struct {
+	client  *retryablehttp.Client
+	headers map[string]string
+}
+
+// TelemetryClientCfg is used to configure the MetricsClient.
+type TelemetryClientCfg struct {
```
loshz, line 52: Wondering if we need this struct as we only pass it to `NewMetricsClient` and we could just pass the fields directly? `func NewMetricsClient(cloudConfig, hclog.Logger) (MetricsClient, error)`

agent/hcp/client/metrics_client.go
```diff
@@ -0,0 +1,158 @@
...
+	CloudCfg cloudConfig
+	Logger   hclog.Logger
+}
+
+// NewMetricsClient returns a configured MetricsClient.
+// The current implementation uses otlpClient to provide retry functionality.
+func NewMetricsClient(cfg *TelemetryClientCfg) (MetricsClient, error) {
+	if cfg.CloudCfg == nil || cfg.Logger == nil {
+		return nil, fmt.Errorf("failed to init telemetry client: provide valid TelemetryClientCfg")
```
loshz, line 61: We should be a little more descriptive in the error returned here. Maybe specify the exact fields that are missing?

agent/hcp/client/metrics_client.go
```diff
@@ -0,0 +1,158 @@
...
+		req.Header.Set(k, v)
+	}
+
+	resp, err := o.client.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to export metrics: %v", err)
+	}
+
+	var respData bytes.Buffer
+	if _, err := io.Copy(&respData, resp.Body); err != nil {
```
loshz, line 141: I can't remember if `io.Copy` does this for us (I don't think so), but we should probably call `defer resp.Body.Close()` beforehand.

agent/hcp/client/metrics_client_test.go
```diff
@@ -0,0 +1,162 @@
...
+	"golang.org/x/oauth2"
+	"google.golang.org/protobuf/proto"
+
+	colpb "go.opentelemetry.io/proto/otlp/collector/metrics/v1"
+	metricpb "go.opentelemetry.io/proto/otlp/metrics/v1"
+
+	"github.com/hashicorp/consul/version"
+	"github.com/hashicorp/go-hclog"
+	hcpcfg "github.com/hashicorp/hcp-sdk-go/config"
+	"github.com/stretchr/testify/require"
```
loshz, line 23: NIT: goimports
```suggestion
	"github.com/hashicorp/go-hclog"
	hcpcfg "github.com/hashicorp/hcp-sdk-go/config"
	"github.com/stretchr/testify/require"
	colpb "go.opentelemetry.io/proto/otlp/collector/metrics/v1"
	metricpb "go.opentelemetry.io/proto/otlp/metrics/v1"
	"golang.org/x/oauth2"
	"google.golang.org/protobuf/proto"

	"github.com/hashicorp/consul/version"
```

loshz (changes requested):
Couple of small changes, but nothing serious.

agent/hcp/client/metrics_client.go
```diff
@@ -0,0 +1,158 @@
...
+// It also holds default HTTP headers to add to export requests.
+type otlpClient struct {
+	client  *retryablehttp.Client
+	headers map[string]string
+}
+
+// TelemetryClientCfg is used to configure the MetricsClient.
+type TelemetryClientCfg struct {
+	CloudCfg cloudConfig
+	Logger   hclog.Logger
```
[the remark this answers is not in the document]
Achooo: yep I can remove it, I passed it in in order to convert it into a named logger. I see that I don't do this. I think we're fine with a custom logger, I'll take it out! Thanks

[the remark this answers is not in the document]
Achooo: Originally, it was used as so see here:
https://github.com/hashicorp/consul/pull/17100/files#diff-0bb7f37c7c0bb6c638aa04cb39fc360152474f06f7061bf92a78a72a81e03455R102

The names get appended `logger.Named("hcp_telemetry_client")`

I do feel like it sucks to fail init for not passing a logger, so I'll remove it.

agent/hcp/client/metrics_client.go
Achooo: weird.. seems like that got deleted in one of my rebases. Was in a previous commit. Good catch, we defo need to close it.

https://github.com/hashicorp/consul/pull/17100/commits/b2b44f5db957e3c7bf8be7f7fc984b445b3bdc09

agent/hcp/client/metrics_client.go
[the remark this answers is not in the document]
Achooo: Hm I was just holding a constant in the MetricsClient for default headers.

 If we want to reuse the header object directly, we'll have to make a copy when we create the request to be safe... which is also fine. I'll make those updates.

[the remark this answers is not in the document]
Achooo: I _think_ we're supposed to use a clone, maybe that's changed now

agent/hcp/client/metrics_client.go
Achooo: will remove in upcoming commit - definitely overkill!

agent/hcp/client/metrics_client.go
Achooo: Do we want to completely remove it or just the `hcp-go-sdk`? I thought that it was a standard and if not already, eventually useful for internal purposes (that header acts like a user agent from what I remember in our discussion). @nickethier any thoughts on how the Gateway uses these?

Happy to remove if useless

agent/hcp/client/metrics_client.go
```diff
@@ -0,0 +1,158 @@
...
+	req, err := retryablehttp.NewRequest(http.MethodPost, endpoint, bytes.NewBuffer(body))
+	if err != nil {
+		return fmt.Errorf("failed to export metrics: %v", err)
+	}
+
+	for k, v := range o.headers {
+		req.Header.Set(k, v)
+	}
+
+	resp, err := o.client.Do(req)
```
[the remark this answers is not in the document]
Achooo: yep! thanks - will fix in upcoming commit

agent/hcp/client/metrics_client.go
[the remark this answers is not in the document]
Achooo: hmm actually seems like the logger benefits from the default options with the previous fashion. I think let's keep it to be consistent with the rest of how Consul does logging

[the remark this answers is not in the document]
Achooo: Oh think I just understood what you meant..... Just not add the logger to the MetricsClient 😓 sure I can remove it.

agent/hcp/client/metrics_client.go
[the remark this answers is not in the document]
Achooo: Created JIRA ticket for GA and added todo comment.

agent/hcp/client/metrics_client.go
```diff
@@ -0,0 +1,158 @@
...
+
+	client := &http.Client{
+		Transport: transport,
+		Timeout:   defaultStreamTimeout,
+	}
+
+	retryClient := &retryablehttp.Client{
+		HTTPClient:   client,
+		Logger:       logger.Named("hcp_telemetry_client"),
+		RetryWaitMin: defaultRetryWaitMin,
```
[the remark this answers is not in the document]
Achooo: yep, created JIRA ticket and added TODO comment!

agent/hcp/client/metrics_client.go
```diff
@@ -0,0 +1,150 @@
...
+	body, err := proto.Marshal(pbRequest)
+	if err != nil {
+		return fmt.Errorf("failed to export metrics: %v", err)
+	}
+
+	req, err := retryablehttp.NewRequest(http.MethodPost, endpoint, bytes.NewBuffer(body))
+	if err != nil {
+		return fmt.Errorf("failed to export metrics: %v", err)
+	}
+	req.Header = (*o.header).Clone()
```
Achooo, line 124: @loshz @chapmanc I think it's best to clone this. Usually, clients will hold a copy of the request and copy the whole request.

agent/hcp/client/metrics_client.go
```diff
@@ -0,0 +1,158 @@
...
+		return fmt.Errorf("failed to export metrics: %v", err)
+	}
+
+	if respData.Len() != 0 {
+		var respProto colmetricpb.ExportMetricsServiceResponse
+		if err := proto.Unmarshal(respData.Bytes(), &respProto); err != nil {
+			return fmt.Errorf("failed to export metrics: %v", err)
+		}
+
+		if respProto.PartialSuccess != nil {
```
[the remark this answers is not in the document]
Achooo: I followed suit of this exporter if it helps: https://github.com/open-telemetry/opentelemetry-go/blob/8e76ab23b495ce2411d936b497fb8f1a9b35f974/exporters/otlp/otlpmetric/otlpmetrichttp/client.go#L200

agent/hcp/client/metrics_client.go
Achooo: Done ! split those up.

agent/hcp/client/metrics_client_test.go
Achooo: Not necessarily actually, I just added it to test out that part of my code which handles the partial success

agent/hcp/client/mock_CloudConfig.go
Achooo, line 1: Moved the mocks to a common file instead of keeping in `metrics_client_test.go`. This allows its usage in external places (e.g. Deps) where the MetricsClient is initialized for easy testing, see this PR : https://github.com/hashicorp/consul/pull/17162

loshz (approved):
LGTM

agent/hcp/client/metrics_client.go
```diff
@@ -37,48 +35,44 @@ type MetricsClient interface {
...
 // The current implementation uses otlpClient to provide retry functionality.
-func NewMetricsClient(cfg *TelemetryClientCfg) (MetricsClient, error) {
-	if cfg.CloudCfg == nil || cfg.Logger == nil {
-		return nil, fmt.Errorf("failed to init telemetry client: provide valid TelemetryClientCfg")
+func NewMetricsClient(cfg CloudConfig, logger hclog.Logger) (MetricsClient, error) {
+	if cfg == nil {
+		return nil, fmt.Errorf("failed to init telemetry client: provide valid cloudCfg (Cloud Configuration for TLS)")
+	}
+
+	if logger == nil {
```
chapmanc, line 56: If all good with you I think it's acceptable to use the `hclog.FromContext(ctx)` unless we don't have a context? Don't let it block you but seems like we should just pass the context instead of the logger.

chapmanc (approved):
LGTM with a small question.

agent/hcp/client/metrics_client.go
loshz: Agreed. I think this could be useful in the future.

Achooo: Fixed!

PR #17108, opened 2023-04-24
Create tombstone on resource `Delete`

analogue:
### Description

See NET-2698. There are some TODOs inline to address.

- New and existing tests are green

* [x] updated test coverage
* [x] not a security concern


agent/grpc-external/services/resource/delete_test.go
```diff
@@ -140,10 +142,64 @@ func TestDelete_Success(t *testing.T) {
...
+				NamePrefix: "",
+			})
+			require.NoError(t, err)
+			rspCh := handleResourceStream(t, stream)
+			event := mustGetResource(t, rspCh)
+			require.Equal(t, event.Operation, pbresource.WatchEvent_OPERATION_UPSERT)
+			tombstone := &pbresource.Tombstone{}
+			require.NoError(t, event.Resource.Data.UnmarshalTo(tombstone))
+			prototest.AssertDeepEqual(t, artistId, tombstone.OwnerId)
+			require.Equal(t, artist.Version, tombstone.OwnerVersion)
```
boxofrad, line 159: Is there a particular reason to use `WatchList` here, rather than `Read`-ing the tombstone?

internal/resource/registry_test.go
```diff
@@ -123,3 +115,20 @@ func TestResolve(t *testing.T) {
...
+		} else {
+			errstr, ok := r.(string)
+			if !ok {
+				t.Errorf("unexpected error type returned from panic")
+			} else if errstr != panicString {
+				t.Errorf("expected %s error message but got: %s", panicString, errstr)
+			}
+		}
+	}()
+
```
boxofrad, line 132: I think this could be replaced with [`require.PanicsWithValue`](https://pkg.go.dev/github.com/stretchr/testify/require#PanicsWithValue).

boxofrad (commented):
This looks great! 🏆 Couple of minor points of feedback inline...

agent/grpc-external/services/resource/delete.go
```diff
@@ -80,6 +89,44 @@ func (s *Server) Delete(ctx context.Context, req *pbresource.DeleteRequest) (*pb
...
+		Metadata: map[string]string{
+			"generated_at": time.Now().Format(time.RFC3339),
+		},
+	}
+
+	if _, err := s.Write(ctx, &pbresource.WriteRequest{Resource: tombstone}); err != nil {
+		return fmt.Errorf("failed writing tombstone: %w", err)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
analogue: I'm not seeing where the `ErrCASFailure` occurs in the following scenario. 

Given an existing resource "foo":

Scenario:
1. User instructs ResourceService to Delete("foo")
2. ResourceService creates tombstone via WriteCAS("tombstone-foo-uid1")
3. ResourceService fails to DeleteCAS("foo") and returns error to User
4. Time passes and User retries Delete("foo")
6a. Branch a: tombstone has not been reconciled by the reaper controller and still exists
    - ResourceService creates tombstone via WriteCAS("tombstone-foo-uid1")
    - WriteCAS() is identical to the write in Step 2. and thus a no-op.
    - Delete("foo") successful and scenario ends. 
6b. Branch b: tombstone has been reconciled by the reaper controller and has been deleted
    - ResourceService creates a tombstone via WriteCAS("tombstone-foo-uid1")
    - WriteCAS() creates a new "tombstone-foo-uid1".
    - Delete("foo") is successful and the scenario ends.

agent/grpc-external/services/resource/delete.go
```diff
@@ -80,6 +89,44 @@ func (s *Server) Delete(ctx context.Context, req *pbresource.DeleteRequest) (*pb
...
+	if err != nil {
+		return status.Errorf(codes.Internal, "failed creating tombstone: %v", err)
+	}
+
+	tombstone := &pbresource.Resource{
+		Id: &pbresource.ID{
+			Type:    resource.TypeV1Tombstone,
+			Tenancy: deleteId.Tenancy,
+			// TODO(spatel): Does this need to be more unique (embed deleteId.Uid)?
+			Name: fmt.Sprintf("%s-tombstone", deleteId.Name),
```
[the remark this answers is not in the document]
analogue: Updated in https://github.com/hashicorp/consul/pull/17108/commits/852bba518b21890037b808760b7612ea6e451302

internal/resource/registry_test.go
analogue: 👍 Fixed in https://github.com/hashicorp/consul/pull/17108/commits/1c66e1f3551d8b11a6e9c3a82cf5ebad60bdea23

agent/grpc-external/services/resource/delete.go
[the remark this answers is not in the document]
boxofrad: > WriteCAS() is identical to the write in Step 2. and thus a no-op.

This is where we'll get an `ErrCASFailure` ☝🏻 because we're providing a blank version.

agent/grpc-external/services/resource/delete_test.go
analogue: Yup, way simpler! https://github.com/hashicorp/consul/pull/17108/commits/ea8dc796d36d5f1b9aef7eddac7564573437d798

proto-public/pbresource/resource.proto
```diff
@@ -171,3 +171,10 @@ message WatchListRequest {
   Tenancy tenancy = 2;
   string name_prefix = 3;
 }
+
+// Given an owner resource, a tombstone represents a promise to delete all
+// the immediately owned (child) resources.
+message Tombstone {
+  ID owner_id = 1;
+  string owner_version = 2;
```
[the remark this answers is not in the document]
analogue: 👍 Fixed in https://github.com/hashicorp/consul/pull/17108/commits/b61abdab605f7bb8905e2ee1dc784f3f017368b0

internal/resource/registry.go
```diff
@@ -68,9 +68,21 @@ type TypeRegistry struct {
...
+	registry.Register(Registration{
+		Type:  TypeV1Tombstone,
+		Proto: &pbresource.Tombstone{},
+		ACLs: &ACLHooks{
+			// TODO(spatel): Figure out ACLs. Delete(ctx, resource) doesn't have acls to Write(ctx, resource's tombstone)
+			//               How to pull a tombstone only ACL out of nowhere to inline in Delete(..)?
+			Read:  func(a acl.Authorizer, i *pbresource.ID) error { return nil },
+			Write: func(a acl.Authorizer, i *pbresource.ID) error { return nil },
+			List:  func(a acl.Authorizer, t *pbresource.Tenancy) error { return nil },
+		},
```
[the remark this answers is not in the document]
analogue: Let me see if I can elaborate.

1. A user wants to delete artist "foo"
2. User calls `Delete("foo")` on the resource service passing a token which only allows it to read/write/list on resources of type "artist".
3. The resource service in turns calls `Write("tombstone-foo-uid1")` with the same context (hence user token) that was passed in to `Delete("foo")`
4.  We're now requiring that the user's token also has privileges to write resources of type tombstone which it does not have.

To remedy this, the context at the call site to `Write("tombstone-foo-uid1")` in the `Delete` endpoint needs a token that allows creating a tombstone.  I'm assuming users should not have to be concerned with ACLs related to the tombstone resource so the request needs to switch "roles" transparently for the writes related to tombstones.

boxofrad:
Thanks for incorporating my feedback, @analogue 🙇🏻 

I've just realised where my confusion has been coming from: I've been reading `maybeCreateTombstone` wrong this whole time, my bad entirely! 🙈 

Rather than creating the tombstone with `Server.Write`, I'd assumed that we were calling `Backend.WriteCAS` directly. I think this makes more sense because it removes the ACL problem you [describe](https://github.com/hashicorp/consul/pull/17108#discussion_r1178356637) and also avoids calling validation/mutation hooks which aren't relevant for tombstones.

I think this should do the trick:

```Go
_, err = s.Backend.WriteCAS(ctx, &pbresource.Resource{
	Id: &pbresource.ID{
		Type:    resource.TypeV1Tombstone,
		Tenancy: deleteId.Tenancy,
		Name:    fmt.Sprintf("%s-%s-tombstone", deleteId.Name, deletedId.Uid),
		Uid:     ulid.Make().String(),
	},
	Generation: ulid.Make().String(),
	Data:       data,
	Metadata: map[string]string{
		"generated_at": time.Now().Format(time.RFC3339),
	},
})
```

internal/resource/registry.go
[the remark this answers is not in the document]
boxofrad: Just spotted where my confusion was coming from, see: https://github.com/hashicorp/consul/pull/17108#issuecomment-1525436560

agent/grpc-external/services/resource/delete.go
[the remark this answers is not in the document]
boxofrad: Just spotted where my confusion was coming from, see: https://github.com/hashicorp/consul/pull/17108#issuecomment-1525436560

analogue:
@boxofrad ok, think we're in much better shape now with https://github.com/hashicorp/consul/pull/17108/commits/aa01e42a6ec0a3f17574d55a8b6939527b0e3433 and https://github.com/hashicorp/consul/pull/17108/commits/6eb1bb02fa02882aaba126a559f05adfedda8073. I can't claim to be a fan of creating the tombstone resource via the backend instead of going through the resource service's write endpoint. Not so much for the things which don't apply, but more so for the duplication of things that may change however unlikely, but 🤷.

proto-public/pbresource/resource.proto
```diff
@@ -230,6 +237,11 @@ service ResourceService {
   // version in the Resource.Version field. If the given version doesn't match
   // what is currently stored, an Aborted error code will be returned.
   //
+  // To perform a blanket write (update regardless of the stored version), 
+  // provide an empty Version in the Resource.Version field. Note that the
+  // write may still fail due to not being able to internally do a CAS write
+  // and return an Aborted error code.
+  //
```
boxofrad, line 244: Thanks for updating the docs 🙌🏻

agent/grpc-external/services/resource/server.go
```diff
@@ -149,3 +150,8 @@ func validateId(id *pbresource.ID, errorPrefix string) error {
 }
 
 func clone[T proto.Message](v T) T { return proto.Clone(v).(T) }
+
+func pprint(pb proto.Message) string {
+	marshaler := &protojson.MarshalOptions{Indent: "    "}
+	return marshaler.Format(pb)
+}
```
boxofrad, line 157: Can't see where this is used. Was it left by accident? 🙇🏻

proto-public/pbresource/resource.proto
```diff
@@ -181,6 +181,13 @@ message Reference {
   string section = 4;
 }
 
+// Given a resource, a tombstone represents a promise to delete all immediately
+// owned (child) resources, if any.
+message Tombstone {
+  // Owning resource identifier.
+  ID owner = 1;
+}
+
```
boxofrad, line 190: nit (sorry!): Would you mind starting these comments with the type/field name, so that they're valid godoc comments?

```suggestion
// Tombstone represents a promise to delete all of a resource's immediately
// owned (child) resources, if any.
message Tombstone {
  // Owner identifies the deleted resource.
  ID owner = 1;
}

```

boxofrad (approved):
Looks great! Thanks @analogue 🙌🏻 

Agreed that the duplication doesn't feel great, but I think it's the right trade-off.

proto-public/pbresource/resource.proto
analogue: TIL 😄

PR #17115, opened 2023-04-24
APIGW: Update how status conditions for certificates are handled

jm96441n:
### Description

This changes how status conditions for invalid certifcates on listeners are reported so that the status condition is on the listener and mentions which certificate is invalid in the message.

Run unit tests

### Links

N/A

### PR Checklist

* [X] updated test coverage
* [X] appropriate backport labels added
* [ ] not a security concern


agent/consul/gateways/controller_gateways.go
```diff
@@ -248,6 +248,16 @@ func (r *apiGatewayReconciler) reconcileGateway(_ context.Context, req controlle
...
+		meta.eachListener(func(listener *structs.APIGatewayListener, bound *structs.BoundAPIGatewayListener) error {
+			listenerRef := structs.ResourceReference{
+				Kind:           structs.APIGateway,
+				Name:           meta.BoundGateway.Name,
+				SectionName:    bound.Name,
+				EnterpriseMeta: meta.BoundGateway.EnterpriseMeta,
+			}
+			updater.SetCondition(validCertificate(listenerRef))
+			return nil
+		})
```
andrewstucki, line 260: So, I think what you'd actually need to do here is move this call above the iteration on line 244. Take for instance if one listener has valid certificate refs and the other does not, what will happen?

1. The first, valid listener, will get no `ResolvedRefs` status condition
2. The second, invalid listener, will get `ResolvedRefs == False`

What we want is for:

1. The first, valid listener, to get `ResolvedRefs == True`
2. The second, invalid listener, to get `ResolvedRefs == False`

Since calling `SetCondition` will override the condition with the same `Type` and `Ref`, then we should be able to just set both of them to `ResolvedRefs == True` initially and use the range in line 244 to override them to `ResolvedRefs == False` when we encounter an error. You should be able to add a test that verifies the whole matrix of:

1. all listeners valid refs
2. all listeners invalid refs
3. mixed valid/invalid refs

checking the `ResolvedRefs` and `Accepted` statuses (with `InvalidCertificates`) of the gateway after.

andrewstucki (changes requested):

agent/consul/gateways/controller_gateways.go
jm96441n: ahhh gotcha, that makes sense, lemme update this

agent/consul/gateways/controller_gateways_test.go
```diff
@@ -3625,7 +4020,7 @@ func TestAPIGatewayController(t *testing.T) {
 						require.NoError(t, err)
 						ppExpected, err := json.MarshalIndent(expectedStatus, "", "  ")
 						require.NoError(t, err)
-						require.True(t, statusEqual, fmt.Sprintf("statuses are unequal: %+v != %+v", string(ppActual), string(ppExpected)))
+						require.True(t, statusEqual, fmt.Sprintf("statuses are unequal (actual != expected): %+v != %+v", string(ppActual), string(ppExpected)))
```
jm96441n, line 4023: this was a small QoL change, I can remove it if we think it's unnecessary

andrewstucki (approved):
LGTM -- I do eventually want to circle back and think about whether we want to handle the `Accepted` status differently than we do now (since a single `ResolvedRefs == False` listener makes `Accepted == False`), but as that likely will require a decent amount of additional internal plumbing to handle appropriately, I'm cool doing it later.

